PR #310, opened 2016-03-29
(PSI) add Rust logo to all Rust icons

### LINKED ISSUE #306: Add Rust logo to all Rust icons
Take a look at his screenshot:

icons

The lower two icons are for Kotlin classes and they are easy distinguishable because of a small Kotlin logo. The upper two icons are for Rust structs, but it is not that easy to guess.

zjhmale:
Fix #306

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
+++ b/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
@@ -336,6 +336,7 @@ fake item_ ::= attrs_and_vis {
                   "org.rust.lang.core.psi.RustAccessControlElement" ]
   mixin       = "org.rust.lang.core.psi.impl.mixin.RustItemImplMixin"
   stubClass   = "org.rust.lang.core.stubs.RustItemStub"
+  methods     = [getPresentation]
 }
 
 upper item ::= << injectInto attrs_and_vis item_inner >> {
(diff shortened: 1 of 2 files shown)


matklad:
This is not exactly what is needed here. We still want different icons for different Rust items:

icons


I think that adding a small R (maybe without circle?) to the lower right corner should do the trick.

zjhmale:
just like kotlin right?  i will get these icons done.

matklad:
Yes, like Kotlin. I think there should be a way of composing Rust icon over existing icons, and that we don't need to duplicate actual icon pictures.

zjhmale:
kotlin maintains the icons itself, so maybe there is not possible of composing customize icon over existing icons? just almost 5 or 6 icons, not a big deal to duplicate exist icon pictures here i think.

https://github.com/JetBrains/kotlin/blob/master/idea/idea-analysis/src/org/jetbrains/kotlin/idea/KotlinIcons.java

https://github.com/JetBrains/kotlin/blob/master/idea/resources/org/jetbrains/kotlin/idea/icons/classKotlin.png

matklad:
Hm, on the other hand, do we really need this feature? 

Kotlin surely needs it, because it is used together with Java and it is important to distinguish Java and Kotlin items. 

Rust, on the other hand, is usually used by itself, so there is no information added by Rust logo. It only would clutter the icon. 

I would like to have this feature personally, because intelij-rust is a mixed Kotlin/Rust/Java codebase, but I am a minority here :)

Sorry for a won't fix feature request :)

zjhmale:
it's ok :)

PR #253, opened 2016-02-20
(CARGO): use sdk when importing project

matklad:
Finally it is working.

The aim of this change is twofold:

* replace custom `cargoHomePath` field and [associated logic](https://github.com/intellij-rust/intellij-rust/blob/ffde5f02d182b267547f9b9c6eca8f68eb478814/src/main/kotlin/org/rust/cargo/project/settings/controls/builders/CargoProjectSettingsControlBuilderImpl.kt#L63-L66) with a build in `SdkComboBox`

* automatically setup project SDK on import (inspired by [Java](https://upsource.jetbrains.com/idea-ce/file/idea-ce-5fec12b146609338bee5dee5208dd1a011a1fabd/java/java-impl/src/com/intellij/externalSystem/JavaProjectDataService.java?nav=1843:1865:focused&line=0)).

So, before this change the import dialog looked like this

before

and the user had to setup SDK manually even though she has already specified path to cargo.

Now it looks like this: 

after

@alexeykudinkin please review

PR #313, opened 2016-03-31
(INSPECTIONS): warn if SDK is not configured

matklad:
This warns user is SDK is not configured:

sdk

By itself this is pretty useless after #253 has been merged. But I want to expand this and add a warning/quick fix for attaching SDK sources (in this or in a separate PR).

@alexeykudinkin please review.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/RustSdkType.kt
+++ b/src/main/kotlin/org/rust/cargo/project/RustSdkType.kt
@@ -15,7 +15,6 @@ import org.rust.cargo.util.PlatformUtil
 import org.rust.ide.icons.RustIcons
 import java.io.File
 import java.util.concurrent.ExecutionException
-import java.util.regex.Pattern
 
 class RustSdkType : SdkType("Rust SDK") {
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt
```diff
@@ -25,6 +25,8 @@ class RustFile(fileViewProvider: FileViewProvider) : PsiFileBase(fileViewProvide
 val PsiFile.modulePath: RustModulePath?
     get() = RustModulePath.devise(this)
 
+val VirtualFile.isRustFile: Boolean get() = fileType == RustFileType
+val VirtualFile.isNotRustFile: Boolean get() = !isRustFile
```
matklad: Dubious API decision :) But I think that most uses will be `if (!file.isRustFile) return defualt`, so it makes sense to add a bit of readability here.

matklad:
Ok, now it downloads sources of standard library. There are no tests for this thing:

* we generally don't test GUI
* the implementation will change in a mid term, when Rust guys find a way to package std library sources with the compiler (rustup issue: https://github.com/rust-lang-nursery/multirust-rs/issues/37)

And hey, here is a gif:

sdk

src/main/kotlin/org/rust/cargo/project/RustSdkType.kt
```diff
@@ -205,5 +165,86 @@ val Module.rustSdk: Sdk? get() {
...
+            )
+        }
+
+        private val log = Logger.getInstance(RustcVersion::class.java)
+    }
+}
+
 private val Sdk?.isRustSdk: Boolean get() =
-    this != null && sdkType == RustSdkType.INSTANCE
+this != null && sdkType == RustSdkType.INSTANCE
```
jajakobyly: This indentation was probably removed by accident

matklad: Well, I guess you can call the current Kotlin formatter an accident =) Thanks!

matklad:
Urgh, this is yet one more thing which is completely broken in non IDEA IDEs =(

matklad:
superseded by #323

PR #317, opened 2016-04-04
(RUNCONFIG) Show env vars dialog, and run cargo with env vars

ntakimura:
This PR is to run cargo with environment variables, ex. `RUST_LOG` for `env_logger`.

It adds two ui components:

 - `Environment variables` field on the run config dialog
 - `Environment Variables` dialog on clicking a `...` button

screen shot 2016-04-04 at 1 16 51 pm

screen shot 2016-04-04 at 1 17 00 pm

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt
@@ -25,6 +25,7 @@ class CargoCommandConfiguration(project: Project,
 
     var command: String = "run"
     var additionalArguments: String = ""
+    var environmentVariables: Map<String, String>? = null
 
     init {
         configurationModule.module = project.getModules().firstOrNull()
@@ -49,7 +50,7 @@ class CargoCommandConfiguration(project: Project,
         val pathToCargo = module.pathToCargo ?: return null
         val moduleDirectory = PathUtil.getParentPath(module.moduleFilePath)
         val args = ParametersListUtil.parse(additionalArguments)
-        return CargoRunState(environment, pathToCargo, moduleDirectory, command, args)
+        return CargoRunState(environment, pathToCargo, moduleDirectory, command, args, environmentVariables)
     }
 
     override fun writeExternal(element: Element) {
--- a/src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt
@@ -25,7 +25,7 @@ class CargoCommandConfiguration(project: Project,
 
     var command: String = "run"
     var additionalArguments: String = ""
-    var environmentVariables: Map<String, String>? = null
+    var environmentVariables: Map<String, String> = mutableMapOf()
 
     init {
         configurationModule.module = project.getModules().firstOrNull()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt
```diff
@@ -25,6 +25,7 @@ class CargoCommandConfiguration(project: Project,
 
     var command: String = "run"
     var additionalArguments: String = ""
+    var environmentVariables: Map<String, String>? = null
```
matklad: Can we use `var environmentVariables: Map<String, String> = emptyMap()` here and elsewhere to avoid dealing with `null`?

matklad:
Ouch, this is a big omission from the Run Configuration :)

Thanks!

src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt
ntakimura: Thank you so much for your review.

I will fix it, but IDE tells me a few errors. So, if you’d indulge, would you accept that syntax?

I rewrote to `var environmentVariables: Map<String, String> = emptyMap()`, and 
I upgraded to snapshot from stable. Then, IntelliJ tells an error: 

> XmlSerializationException: java.lang.UnsupportedOperationException: Mutating immutable collection

Current stable plugin does not have that configuration, so IntelliJ cannot generate the CargoCommandConfiguration object.

After that, I rewrote to `var environmentVariables: Map<String, String>?`, and I rebuilt plugin. Then Kotlin Compiler tells and error: 

> Error:(28, 5) Kotlin: Property must be initialized or be abstract

That’s not allowed in kotlin syntax unlike java syntax.

I have no idea to fix it...

matklad: Ah, that's a tough bug!

TLDR: to fix it, just use `var environmentVariables: Map<String, String> = mutableMapOf()`


To understand what is happening here, we need to understand Kotlin collection types. The official docs are here: https://kotlinlang.org/docs/reference/collections.html, what follows is my rough interpretation.

There are mutable collections, read only collections and immutable collections. If you have an immutable collection, you can be sure that it will never change. If you have a read only collection, you can't change it yourself, but somebody else can! A view into mutable collection is a readonly collection. Immutable implies read only, but vice verse is not true.

In Java, there is a single interface for all collection sorts, and readonly/immutable flavors just throw an exception in update operations.

In Kotlin, `MutableCollection` interface corresponds to Java `Collection`. However Kotlin also provides `Collection` interface, which is a restricted readonly version of `MutableCollection`. Kotlin does not have a special interface for truly immutable collections.

So, `Map<String, String>` is a readonly map, which can be backed up by either mutable or immutable map.

But `emptyMap()` is a singleton **immutable** map.


The catch is, `CommandConfiguration` is deserialized from XML, and for collections, IDEAs serializer would just `put` elements into them, so the serializer requires that the underlying collection is mutable.

By using `var environmentVariables: Map<String, String> = mutableMapOf()` we achieve most benefits:

* we use an empty map instead of null (Null Object pattern) 
* we restrict ourselves to readonly portion of the interface
* the serializer is able to read the class from XML, because the underlying map is mutable.

ntakimura: @matklad Thank you so detailed description.

I fix it at 9169624 , and remove verbose optional `?` keyword from a few lines to apply common syntax. Would you review last PR?

PR #318, opened 2016-04-04
(INTENTION) Added "Add derive clause" intention

### LINKED ISSUE #309: Intention: add derive clause
Type

```
struct S {
    field: i32,
}
```

Hit `alt+enter` and get

```
#[derive(<caret>)]
struct S {
    field: i32,
}
```

This should also work for `enum`s. It may also work for adding a trait to the existing `derive` clause.

netvl:
Fixes #309.

I will add tests and code examples soon. For now I believe the code is ready for review.

There is one problem, I think. Adding new items to existing code usually ruins formatting. In example intentions in IDEA SDK docs auto-reformatting is used to overcome this, but apparently it does not work for Rust files. Therefore adding a new attribute now looks somewhat messy. It may make sense to add some code to insert the necessary indentation manually, but I'm not sure whether it is really needed.


src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.rust.lang.core.psi.util.parentOfType
+
+class AddDerive : PsiElementBaseIntentionAction()
+                , IntentionAction {
+    override fun getFamilyName() = "Add derive clause"
+    override fun getText() = "Add Derive Clause"
+    override fun startInWriteAction() = true
+
+    override fun invoke(project: Project, editor: Editor?, element: PsiElement) {
+        val item = element.parentOfType<RustItem>() ?: return
```
matklad: I think we want only `struct` or `enum` items here.

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
```diff
@@ -0,0 +1,62 @@
...
+        val existingDeriveAttr = item.outerAttrList.find { it.metaItem?.identifier?.textMatches("derive") ?: false }
```
matklad: There is an API for querying item attributes: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustItemImplMixin.kt#L35-L53.

It could be extended to have a function like `findAttribute(name: String) : RustMetaItem`. I'm not 100% sure that it is better to go through `queryAttributes` here, because you only need to deal with outer attributes.

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        val deriveAttr = if (existingDeriveAttr != null) existingDeriveAttr
```
matklad: This can be replaced with elvis operator

`val deriveAttr = exisintDeriveAttr ?: ...`

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
matklad: Ah, I see this in `isAvailable` now. Hm, what about introducing `findStructOrEnum` helper function and using it here and in `isAvailable`? 

It should also allow to get rid of that `when`.

matklad:
>dding new items to existing code usually ruins formatting. In example intentions in IDEA SDK docs auto-reformatting is used to overcome this, but apparently it does not work for Rust files.

Yep, reformatting code is necessary after refactorings. Our formatter is pretty useless at the moment, but @jajakobyly is working on it :) There is no need to write custom hacks for formatting for each particular intentions, we'd better fix our formatter. It's ok if currently intention results in ungly code.

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+ * #[attr(something]    ->  #[attr(something, new]
+ * #[attr()]            ->  #[attr(new)]
+ * #[attr(]             ->  #[attr(new]
+ * ```
+ *
+ * @return `meta` argument or its copy if insertion was successful, null otherwise.
+ */
+fun RustAttr.appendMetaToList(meta: RustMetaItem): RustMetaItem? {
+    // extract attribute internals
+    val attrMeta = when (this) {
```
matklad: Ah, this is sad :(

It is a problem with our grammar, I'll fix https://github.com/intellij-rust/intellij-rust/issues/319. Can't give you an ETA though, so it's better to add a `TODO: #319` comment here.

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+            attrMeta.deleteChildRange(attrMeta.identifier.nextSibling, attrMeta.lastChild)
+        }
+
+        val tempMeta = RustElementFactory.createMeta(this.project, "whatever(item)") ?: return null
+        tempMeta.metaItem!!.replace(meta)
+
+        val newLeftParen = attrMeta.addRangeAfter(tempMeta.lparen, tempMeta.rparen, attrMeta.identifier)
+        return newLeftParen.nextSibling as RustMetaItem
+
+    } else if (attrMeta.lparen != null) {
```
matklad: First of all, this `else` can be removed because the previous branch ends with a `retrun`.

Also, in Kotlin you don't usually write `if else if` chains, and use 

```Kotlin
when {
    cond1 -> expr1
    cond2  -> expr2
    else -> default
}
```

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+        if (leftParen.nextSibling != null && leftParen.nextSibling != attrMeta.rparen) {
+            // if there are other meta items in parens, add `meta` after the last one of them
+            val elementToInsertAfter = if (attrMeta.rparen != null)
+                attrMeta.rparen!!.prevSibling!!
+            else
+                attrMeta.lastChild!!
+
+            // we also need a comma, so first we insert a comma and a temporary meta item from another well-formed meta
+            val tempMeta = RustElementFactory.createMeta(this.project, "whatever(item,item)") ?: return null
+            val start = tempMeta.lparen!!.nextSibling!!.nextSibling!!
```
matklad: `!!` almost always leads to `NPE` eventually, when code is refactored and moved around. It's better to save intermediate results in locals to benefit from flow sensitive typing:

```
val leftParen =  attrMeta.lparen

if (leftParen != null) {
    // No need for `leftParen!!` here, because Kotlin knows it can't be `null`
}

```

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
matklad: Fixed in https://github.com/intellij-rust/intellij-rust/commit/1c609f91f37fbd0d8ab39c1066ddd3c7b55145a6

netvl: That was quick! Thanks :+1: will rebase.

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
netvl: Nope, it cannot. I attempted that at first, but apparently IDEA sees the right hand side as a closure, not as a block.

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
netvl: Well, in this particular case I think that `if`s (without elses, as you suggest) look better. And also `when` introduces an additional level of nesting.

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
netvl: Well, in this particular case I view these `!!`s as assertions - here neither `lparen`, nor both `nextSibling`s cannot be null, it is a programming error otherwise. This is much like `Objects.requireNonNull()` from Java SE. So I personally think that it is fine. Also, multiple ifs here will clutter the intention completely unnecessarily. I guess `?.` would also be fine here, but because these variables can never be null, I think it is really redundant.

Moreover, when refactoring, I will definitely be alerted if I see a `!!`, and I will certainly try to understand the logic and make sure that it is still correct after refactoring. I hope this is true for others as well.

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
netvl: I've added a method there solely for searching for outer attributes. It does look nicer this way.

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
matklad: >and I will certainly try to understand the logic and make sure that it is still correct after refactoring.  I hope this is true for others as well.

Humans (well, at least me :) ) tend to make mistakes (especially when it comes to logic), so imo it is much better to turn assertions into compile time errors. I think that in this particular case introducing two variables, `leftParen` and `rightParen` would substantially reduce the need in `!!`. And this does not need any additional ifs, they are already there (but Kotlin can't deduce non nulability for `foo.bar` when `bar` is a getter, hence the need in local `val`s). 

The right way to turn this into compile time errors would be to make the following PSI (which I might do some day)

```
interface RustMetaItem {
    @Nullable MetaItemArgList getArgList()
}

interface MetaItemArgList {
    @NotNull PsiElement getLparen()
    @NotNull List<RustMetaItem> getArgs()
    @NotNull PsiElement getRparen()
}
```

However, I think we can completely sidestep the issue if instead of custom logic for adding meta item to arbitrary attr, we will just do the following: 

```
        val deriveAttr = if (existingDeriveAttr != null) existingDeriveAttr
        else {
            val attr = RustElementFactory.createOuterAttr(project, "derive()") ?: return
            when (item) {
                is RustStructItem -> item.addBefore(attr, item.struct)
                is RustEnumItem   -> item.addBefore(attr, item.enum)
                else              -> return
            } as RustOuterAttr
        }
        // Just move the caret and don't insert anything.
        val offset = deriveAttr.metaItem?.rparen?.textOffset ?: deriveAttr.textOffset
        editor?.caretModel?.moveToOffset(offset)
        return
```

It is not as fancy as with adding a coma (btw, have you handled the case when there is already a trailing coma in the arg list?), but hopefully much more robust. What do you think?

src/main/resources/intentionDescriptions/AddDerive/after.rs.template
```diff
@@ -0,0 +1,4 @@
+#[derive()]
```
matklad: You can do `#[derive(<caret>)]` to assert caret position here.

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
netvl: When I was writing the initial implementation, something like `derive()` wasn't a valid `RustMetaItem` :) But since #319 is closed I guess it could work.

And no, I didn't handle adding a comma :( I was under impression that it was forbidden by the grammar as well but it seems it is not so. Yes, I guess your approach does look better and more robust.

Actually, I was thinking of adding an intention like "Add derive clauses for common traits" which should add well-known traits like `Eq`, `PartialEq` and such to the `derive` clause, possibly presenting a list to the user to select the necessary ones and/or deducing possible ones automatically. But this, I guess, would require more thorough analysis of struct contents, like which traits inner types implement. Anyway, this is a matter for another pull request, I believe :) Just wanted to share an idea.

matklad: >possibly presenting a list to the user to select the necessary ones and/or deducing possible ones automatically

We've discussed this on gitter: https://gitter.im/intellij-rust/intellij-rust?at=56fa76be11ea211749c2c4ee

TLDR: I hate non text based UIs, and I think that for this use case the ideal functionality should come in two independent pieces:

* `add derive` intention, which adds a `#[derive()]` if needed and places the caret in the attribute (this PR)
* a completion contributor which suggests common traits and trait combinations (`CC` -> `Clone, Copy`) in `derive`

It is basically the same functionality that you suggest, built with a combination of two smaller features.

netvl: Yes, completion does make sense. But I guess it will have to wait until there is a proper trait analysis in place.

netvl:
I've pushed a commit to my branch but apparently github didn't register it :( it is possibly related to problems on status.github.com. Hopefully it will appear here soon.

netvl:
The commit is here, ready for review :)

src/main/kotlin/org/rust/lang/core/psi/RustElementFactory.kt
```diff
@@ -19,4 +20,14 @@ object RustElementFactory {
         val file = createFileFromText(project, "mod $modName;")
         return file?.childOfType<RustModDeclItem>()
     }
+
+    fun createOuterAttr(project: Project, attrContents: String): RustOuterAttr? {
+        val file = createFileFromText(project, "#[$attrContents] struct Dummy;")
+        return file?.childOfType<RustOuterAttr>()
+    }
+
+    fun createMeta(project: Project, metaContents: String): RustMetaItem? {
```
matklad: This function is unused. Let's remove it now. We'll reintroduce it when we need this functionality.

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
```diff
@@ -0,0 +1,56 @@
...
+        PsiTreeUtil.mark(deriveAttr, marker)
+
+        // XXX: currently reformat does not do anything
+        val reformattedItem = CodeStyleManager.getInstance(project).reformat(item) as RustItem
+        val reformattedDeriveAttr = PsiTreeUtil.releaseMark(reformattedItem, marker) as RustOuterAttr
+
+        val offset = reformattedDeriveAttr.metaItem?.rparen?.textOffset ?:
+            reformattedDeriveAttr.rbrack?.textOffset ?:
+            reformattedDeriveAttr.textOffset
+        editor?.caretModel?.moveToOffset(offset)
```
matklad: Looks good, but I have a couple of minor stylistic suggestions:

* you can move `getKeywordItem(item)` to the top of the function to make it clear that we operate on structs and enums
* `getStructOrEnumKeyword` is a bit more descriptive name than `getKeywordItem`.
* What do you think about introducing three helper functions, corresponding to the logical parts of the intention?

```Kotlin
override fun invoke(project: Project, editor: Editor?, element: PsiElement) {
    val item = element.parentOfType<RustItem>() ?: return
    val keyword = getStructOrEnumKeywork(item) ?: return

    var deriveAttr = locateOrCreateDeriveAttr(item, keyword)
    deriveAttr = reformat(deriveAttr)
    positionCaret(editor, deriveAttr)
}
```

src/main/kotlin/org/rust/lang/core/psi/util/RustAttrExtensions.kt
matklad: Actually, the set of derivable traits is hard coded in the rust compiler at the moment, so we should already be able to do completion here. 

The type system will allow us to quick fix type errors and flag impossible derives.

src/test/resources/org/rust/ide/intentions/fixtures/add_derive/add_derive_enum_after.rs
```diff
@@ -0,0 +1,3 @@
+    #[derive(<caret>)]enum Test {
+        Something
+    }
```
matklad: @jajakobyly here is another test for the formatter

jajakobyly: Actually this seems to be the same case as that one you posted:
```rust
use std::io::Error;

    pub enum RustfmtError {
    IoError(io::Error)
}
```

src/main/kotlin/org/rust/lang/core/psi/RustElementFactory.kt
netvl: Thanks, missed it.

src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
netvl: Done.

matklad:
Thanks!

src/test/resources/org/rust/ide/intentions/fixtures/add_derive/add_derive_enum_after.rs
matklad: There is one more thing in this test: a newline should be required between `#[dervive ]` and `enum`. This can be specified in `Spacing`.

PR #38, opened 2015-10-30
Basic support for cargo projects

winger:
Port of https://github.com/Vektah/idea-rust/pull/42.

Implements cargo project (auto)import, creates a module for each cargo dependency under project root and a library for each dependency outside of the project root.

Relies on cargo metadata subcommand, which is available at https://github.com/winger/cargo-metadata and (hopefully) soon will be installable with cargo install cargo-metadata (after https://github.com/rust-lang/cargo/pull/2091 will be checked in).

PR #323, opened 2016-04-07
Toolchain is the new sdk

matklad:
clion

Ok, so here is an attempt to make the plugin work with other IDEs besides the IDEA itself. The screenshot is from CLion. 

## RustSdkType

Is gone. Although the SDK class itself is present in IDEs, there is no any SDK associated UI. `chooseAndSetSdk()` for example is a no-op in CLion. So this PR introduces a bit of custom state and UI, called `RustToolchain. 

cargo-config

This is a per project thing which stores the path to Cargo

## RustModuleType

Is gone too. Again, there are module classes in other IDEs, but the UI for creating/configuring modules is missing. This PR allows to work with Rust in any module, so you'll be able to use Java and Rust in the same module (and some of our users [use Java and Rust together](http://nitschinger.at/First-Steps-with-Rust-and-JNI/) :) )

## ExternalSystem (import project)

Is also gone. Instead, `CargoMetadataService` is introduced. This service monitors `Cargo.toml` files in the projects and executes `cargo metadata` if necessary. So now you just `open` the directory with Rust sources, setup toolchain in the settings and that's it. 

## Unresolved questions

* tests don't yet compile :)

* some refactoring and code rearrangement may be necessary

* Should we provide extended functionality (SDK and module types) specifically for IDEA? I think in the long run this is a good idea, but currently it is better to stick with a single implementation for all platforms. 

@alexeykudinkin please review. This is not finished yet, but the big chunk of work is already done.


src/main/kotlin/org/rust/cargo/CargoProjectDescription.kt
```diff
@@ -1,40 +1,47 @@
...
 class CargoProjectDescription private constructor(
-    val packages: Collection<Package>,
-    private val rootModule: Package
+    private val idToPackage: BidirectionalMap<Int, Package>,
+    private val dependencies: MultiMap<Int, Int>
 ) {
+    val packages: Collection<Package> get() = idToPackage.values
+    val rawDependencies: MultiMap<Int, Int> get() = MultiMap(dependencies)
 
     class Package(
```
jajakobyly: This looks like a data class for me

matklad: Yep, you are right! I've struggled quite a bit with the design of this `ProjectDescription` class, but I am still not happy :(

The project is a DAG of packages. Serializing graphs is awkward, and we have **two** formats for serializing this thing. One is `cargo metadata` output, which has snake_case names and other peculiarities (`target_kind`). Another is IDEAs XML format, which requires all fields to be mutable and nullable.

matklad:
Hey, it even passes the tests!

:tada:

matklad:
@alexeykudinkin @atsky I think this is basically ready. 

To recap, the aim of this PR is to get rid of IDEA specific parts and to make plugin compatible with CLion and friends. The main change is the replacement of ExternalSystem (introduced in https://github.com/intellij-rust/intellij-rust/pull/38) by a custom `CargoMetadataService`. Sdks and modules also turned to be rather IDEA specific as well. 

There are a couple of related things which can be done in follow up PRs:

* `ModuleIndex` currently does not work really well, because it indexes files before we learn the structure of a project. 
* A notification about misconfigured toolchain would be really helpful.
* Attach standard library sources action.

The last two are basically a reincarnation of https://github.com/intellij-rust/intellij-rust/pull/313.

src/main/kotlin/org/rust/cargo/CargoProjectDescription.kt
```diff
@@ -1,119 +1,84 @@
...
+/**
+ * Rust project model represented roughly in the same way as in Cargo itself.
+ *
+ * [org.rust.cargo.toolchain.CargoMetadataService] is responsible for providing a [CargoProjectDescription] for
+ * an IDEA module.
+ */
 class CargoProjectDescription private constructor(
-    val packages: Collection<Package>,
-    private val rootModule: Package
+    private val idToPackage: BidirectionalMap<Int, Package>,
```
alexeykudinkin: What are the reasons to submit `idToPackage`? I can only spot just a single usage: `idToPackage.values`.

src/main/kotlin/org/rust/cargo/CargoProjectDescription.kt
```diff
@@ -1,119 +1,84 @@
...
         val isModule: Boolean get() = source == null
+        val libTarget: Target? get() = targets.find { it.isLib }
+
+        val virtualFile: VirtualFile? get() = VirtualFileManager.getInstance().findFileByUrl(contentRootUrl)
     }
 
     data class Target(
         /**
-         * Path to the crate root file, relative to the directory with Cargo.toml
+         * Absolute path to the crate root file
```
alexeykudinkin: Absolute paths are fishy. Them ban user from moving her project around.

src/main/kotlin/org/rust/cargo/toolchain/impl/LibraryUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+        ?: return
+
+    fillLibrary(library, roots)
+
+    ModuleRootModificationUtil.addDependency(module, library)
+}
+
+private fun fillLibrary(library: Library, roots: Collection<VirtualFile>) {
+    val model = library.modifiableModel
+    for (url in library.getUrls(OrderRootType.CLASSES)) {
```
alexeykudinkin: What's that for?

alexeykudinkin:
First of all, amount of work is drastic! Though, the work is prominent, it's much better to discuss such a drastic changes prior to implementing them, since _a lot_ of concerns may consequently arise. 

And fixing them post-factum may be much (much!) harder.

matklad:
I've just started to prototype some things and got a bit carried away :) It's OK if all of this has to be rewritten or implemented in a completely different way. I find it fruitful to prototype before discussing in depth.

src/main/kotlin/org/rust/cargo/RefreshCargoProjectAction.kt
```diff
@@ -0,0 +1,39 @@
...
+        if (modules.isEmpty()) return
+        ApplicationManager.getApplication().saveAll()
+        for ((module, toolchain) in modules) {
+            val service = module.getService<CargoMetadataService>()
+            service.updateNow(toolchain)
+        }
+    }
+}
+
+private val AnActionEvent.modulesWithToolchains: List<Pair<Module, RustToolchain>>
```
alexeykudinkin: Why `modulesWithToolchains` is attributed to the `AnActionEvent`? How do they even bind together?

src/main/kotlin/org/rust/cargo/commands/impl/CargoMetadata.kt
```diff
@@ -101,4 +107,61 @@ object CargoMetadata {
          */
         val dependencies: List<String>
     )
+
+    fun intoCargoProjectDescription(project: Project): CargoProjectDescription? {
+        val isRootPkg = { pkg: Package -> pkg.id == project.resolve.root }
+        val orderedPackages = project.packages.filter(isRootPkg) + project.packages.filter { !isRootPkg(it) }
```
alexeykudinkin: What's that for?

matklad: It was a awful idea not to remember the root package explicitly, and instead relay on the fact that the root package is first. Fixed.

src/main/kotlin/org/rust/cargo/CargoProjectDescription.kt
matklad: fixed.

src/main/kotlin/org/rust/cargo/toolchain/impl/LibraryUtils.kt
matklad: When refreshing cargo project, we need to remove old libraries and add new ones. Turns out that it is a bit easier to change an existing library than to delete it.

So this function takes a library (which is either be a brand new or an old one from the previous refresh) and resets its roots to the up to date state.

src/main/kotlin/org/rust/cargo/RefreshCargoProjectAction.kt
matklad: It uses `project` of the `AnActionEvent` to fetch the modules which need to be refreshed. I agree that the abstraction is probably wrong and that the return type is ugly. This `val` is `private` and exists solely to extract common code from `update` and `actionPerfomed` functions. I don't see a way to make this significantly better and suggest leaving it as is.

src/main/kotlin/org/rust/cargo/toolchain/impl/LibraryUtils.kt
alexeykudinkin: Got it.

src/main/kotlin/org/rust/cargo/RefreshCargoProjectAction.kt
alexeykudinkin: I see: attribute extension method to the `project` itself.

matklad: It's not that better imo because `AnActionEvent.project` is nullable. Updated nevertheless.

alexeykudinkin:
I believe we're good to go. 

Though we definitely do need to have this stamped in our docs, prior to merging this. Just copying the full description of the P/R seems to be sufficent for me.

matklad:
New docs: https://github.com/intellij-rust/intellij-rust/blob/new-docs/doc/QuickStart.md

matklad:
The latest two commits add some gui

* when importing project from source in idea

from-source

* when opening Rust file in a project without toolchain in any IDE

notification

src/main/kotlin/org/rust/cargo/projectSettings/RustProjectSettingsService.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.rust.cargo.toolchain.RustToolchain
+
+interface RustProjectSettingsService {
+    var toolchain: RustToolchain?
+    var autoUpdateEnabled: Boolean
+
+    /*
+     * Show a dialog for toolchain configuration
+     */
+    fun configureToolchain()
```
alexeykudinkin: That's an invalid Kotlin-doc. I'll fix those (in this P/R), though let's stick to the valid ones.

PR #287, opened 2016-03-12
(INTENSION) move inline mod to file

zjhmale:
Fix #270 

and test will be covered later.

PR #324, opened 2016-04-08
add fix call operator

zjhmale:
it is cool when [merged](https://github.com/intellij-rust/intellij-rust/pull/287/files#diff-1079d6cc509e9b2ef26a949b13000903R32), but the safe call operator is missing on master branch, really wired :joy: 

@matklad


matklad:
The code in master was updated after I've changed [this](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf#L489) pin.

Now the parser parse `mod_item` only if has seen `{`. I think it is a good idea, because when you parsed `mod foo`, you can't be sure if it will be `mod foo;` or `mod foo { ... }`.

zjhmale:
but `lbrace` of a mod item is still `Nullable`, why should we omit the safe call operator?

matklad:
Ah, you need to generate the parser again (`./gradlew generate`)

zjhmale:
make sense now :)

PR #325, opened 2016-04-08
(GRAM): Collapse LTEQ and similar productions into single tokens

### LINKED ISSUE #321: Collapse LTEQ and similar productions into single tokens
Currently [following productions](https://github.com/intellij-rust/intellij-rust/blob/dda22abaef6f26db0b4f8e1f457665759d345f14/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf#L1101-L1107) are implemented as composite PSI elements:
java_2016-04-05_15-36-41
This makes them pretty inconvenient to manipulate with, and can generate bugs and unexpected behaviours in random places. And also current parser thinks this is 100% ok:
```rust
if a < = b { ... }
```

Instead, we can use [`PsiBuilder.Marker#collapse`](https://upsource.jetbrains.com/idea-ce/file/idea-ce-1bc380f6fd5ae5ac3264332f896fe519dd464807/platform/core-api/src/com/intellij/lang/PsiBuilder.java?nav=5682:5690:focused&line=277) method here, to _merge_ them into single tokens.

jajakobyly:
Fixes #321


src/test/resources/org/rust/lang/core/parser/fixtures/ill-formed/shifts.txt
```diff
@@ -0,0 +1,99 @@
...
+      RustBlockImpl(BLOCK)
+        PsiElement({)('{')
+        PsiWhiteSpace('\n    ')
+        RustLitExprImpl(LIT_EXPR)
+          RustNumericLiteralImpl(<INTEGER>)('1')
+        PsiErrorElement:';' or '}' expected, got '>'
+          <empty list>
+  PsiWhiteSpace(' ')
+  PsiElement(>)('>')
+  PsiElement(DUMMY_BLOCK)
```
jajakobyly: I don't like the mess which starts here, but I don't know how to make recovering properly here (some tricks in `collapse` method?)

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -1098,13 +1098,12 @@ private stmt_ ::= expr_stmt
...
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 
-GTGTEQ ::= '>' '>' '='
-GTGT   ::= '>' '>'
-GTEQ   ::= '>' '='
-
-LTLTEQ ::= '<' '<' '='
-LTLT   ::= '<' '<'
-LTEQ   ::= '<' '='
+external gtgteq ::= collapse GTGTEQ GT GT EQ
```
matklad: What do you think about using `'>>='`  tokens here? I think they are a bit more readable then named counterparts (you  will need to add `GTGTEQ = ">>="` to the tokens array at the beginning of the file).

Overall it is more typing, but also more readability.

src/test/resources/org/rust/lang/core/parser/fixtures/ill-formed/shifts.txt
matklad: I think that `stmt_` rule is the one responsible for recovering here.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
jajakobyly: For me it doesn't matter

jajakobyly: Actually this cannot be done. When I change `GTGTEQ` to `'>>='`, `collapse` is provided with a `Parser` object instead of `IElementType`.

matklad:
Thanks!

jajakobyly:
Ouch, I actually wanted to improve recovering in this PR :sweat_smile:

matklad:
Ah, sorry. Nevertheless, my understanding is that recovering should happen not in collapse, but in the `stmt` and `block` rules, so it's ok to file a separate PR for it. Or event to postpone it: I think we don't need recovering as much as we need formatting =)

PR #326, opened 2016-04-09
Fix shebang line lexing

### LINKED ISSUE #320: Remove `#!` token and fix shebang line
This is valid rust:

```
mod m {
   #    !    [ cfg ( test ) ]
}
```

Not that currently `#! [foo]` is lexed as shebang line, because the symbol after `!` is not `[`. We should fix lexer to generate shebang token only if it is the very first token in the file.

jajakobyly:
Fixes #320


src/main/kotlin/org/rust/lang/core/lexer/RustLexer.flex
```diff
@@ -120,18 +122,17 @@ CHAR_LITERAL   = ( \' ( [^\\\'\r\n] | \\[^\r\n] | "\\x" [a-fA-F0-9]+ | "\\u{" [a
...
-
-SHEBANG_LINE=\#\![^\[].*
-
 %%
 
-<YYINITIAL> \'                    { yybegin(IN_LIFETIME_OR_CHAR); yypushback(1); }
+<YYINITIAL>                     {
+  "#!" [\r\n]                     { yybegin(INITIAL); yypushback(1); return SHEBANG_LINE; }
+  "#!" [^\[\r\n] [^\r\n]*         { yybegin(INITIAL); return SHEBANG_LINE; }
+  [^]                             { yybegin(INITIAL); yypushback(1); }
```
jajakobyly: There is one extreme case when this lexer behaves differently than rustc ([playground](http://is.gd/hU30nI)):
```rust
#!<<EOF>>
```
Though, I don't think handling this specifically is worth anything.

matklad:
:+1:

PR #302, opened 2016-03-27
(FMT) Improve code formatting: Spacing

jajakobyly:
~~This PR is going to introduce fully-fledged Rust code formatter using IDEA's formatting engine. It's not going to replace `rustfmt`, as that tool does more than just white space manipulation.~~

No it's going to implement some spacing fixing.

Creative test cases are very, very welcome (I have tried copying tests from `rustfmt`, but they expect performing too invasive changes).

Most inspiration was taken from [Python formatter](https://upsource.jetbrains.com/idea-ce/file/idea-ce-d93ab79ac5306c9a9880e16438147eafe137f039/python/src/com/jetbrains/python/formatter/PythonFormattingModelBuilder.java) and [Go formatter](https://upsource.jetbrains.com/go-lang-idea-plugin/file/ede4f339b4968989ea1a5bfcfefa0894874fef81/src/com/goide/formatter/GoFormattingModelBuilder.java).

Blocked by: too much stuff
Depends on: even more stuff

PR #328, opened 2016-04-10
(GRAM, PSI) Various PSI improvements

jajakobyly:
This PR is going to improve some parts of the PSI, mostly in order to push forward #302 development (as a side effect we will also get improved expand selection behaviour).

See commit messages for detailed list of changes (I'll put each change in separate commit).


src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -582,11 +582,13 @@ trait_ref ::= path_generic_args_without_colons
...
 
-impl_item ::= UNSAFE? IMPL generic_params type where_clause? '{' inner_attr* impl_member* '}'
-            | UNSAFE? IMPL generic_params '!'? trait_ref FOR (type where_clause? '{' inner_attr* impl_member* '}' | '..' '{' '}') {
-  pin(".*") = "'{'"
+impl_item ::= UNSAFE? IMPL generic_params type where_clause? impl_members_block
+            | UNSAFE? IMPL generic_params '!'? trait_ref FOR (type where_clause? impl_members_block | '..' '{' '}') {
+  pin(".*") = '{'
 }
 
+impl_members_block ::= '{' (inner_attr | impl_member)* '}'
```
matklad: This accepts more programs then the previous implementation and rust compiler. In particular,

```
impl S {
    fn foo() {}
    #![cfg(test)]
}
```

should be a syntax error.

Of course we can handle it on the semantic level, but I think that doing this in grammar is better.

matklad: This can be a nice ill-formed tests btw

matklad: And this should probably include `pin = '{'`

jajakobyly: > Of course we can handle it on the semantic level, but I think that doing this in grammar is better.

My intention was to introduce annotator, but anyway it's material for separate PR

matklad: Let me elaborate a bit why I think handling this at the grammar level is the right way. The main reason is that it is just dramatically less code. Of course we can introduce an annotator for this. And it will be able to provide better error messages and even a quick fix to put inner attributes on top. 

But this is a lot of code which has to be maintained. And anywhere else in the project you'll have to assume that attributes and methods can be in mixed order in `impl`s. 

On the other hand the gain from the custom annotator seems pretty small: inner attributes in impls are rarely used. So spending more than a couple lines of code on this thing seems wasteful.

jajakobyly: Ok, I understand

src/main/kotlin/org/rust/ide/inspections/SelfConventionInspection.kt
```diff
@@ -15,7 +15,7 @@ class SelfConventionInspection : RustLocalInspectionTool() {
             override fun visitImplMethodMember(m: RustImplMethodMember) {
                 val convention = SELF_CONVENTIONS.find { m.identifier.text.startsWith(it.prefix) } ?: return
                 if (m.selfType !in convention.selfTypes) {
-                    holder.registerProblem(m.selfArgument ?: m.identifier, convention)
+                    holder.registerProblem(m.fnParamsWithSelf.selfArgument ?: m.identifier, convention)
```
jajakobyly: Is it possible to rename this property to `params` or something? The only method I have found is to add `methods=[...]` directive in grammar, but this doesn't remove the obscure version.

matklad: As far as I understand it is impossible to remove particular grammar kit accessor.  For this particular case, I think we should use something like

```
fake argument ::= pat? type
fake params ::= Self? argument*

trait_fn_argument ::= pat? type {
  elementType = argument
}
trait_fn_params ::= ... {
   elementType = params
}

impl_fn_argument ::= pat type {
   elementType = argument
}
impl_fn_params ::= ... {
  elementType = params
}
```

This should generate `params` accessor everywhere. And I think we should not have different PSI nodes for params with and without self.

matklad: >The only method I have found is to add methods=[...] directive in grammar

I find it easier to write extension methods than custom accessors.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -887,8 +882,9 @@ cast_expr ::= restricted_expr AS type
 
 unary_expr ::= ('-' | '*' | '!' | ('&' | '&&') MUT?) restricted_expr
 
-lambda_expr ::= MOVE? ('||' | '|' [ <<comma_separated_list lambda_param>> ] '|') ret_type? any_expr
-lambda_param ::= pat type_ascription?
+lambda_expr   ::= MOVE? lambda_params ret_type? any_expr
+lambda_params ::= '||' | '|' [ <<comma_separated_list lambda_param>> ] '|' { elementType=params }
```
jajakobyly: What do you think about removing `||` token from lexer, and use `collapse` wherever it's needed?

matklad: :-1:

jajakobyly:
Done

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -582,11 +582,13 @@ trait_ref ::= path_generic_args_without_colons
...
 
-impl_item ::= UNSAFE? IMPL generic_params type where_clause? '{' inner_attr* impl_member* '}'
-            | UNSAFE? IMPL generic_params '!'? trait_ref FOR (type where_clause? '{' inner_attr* impl_member* '}' | '..' '{' '}') {
-  pin(".*") = "'{'"
+impl_item ::= UNSAFE? IMPL generic_params type where_clause? impl_members_block
+            | UNSAFE? IMPL generic_params '!'? trait_ref FOR (type where_clause? impl_members_block | '..' '{' '}') {
+  pin(".*") = '{'
 }
 
+impl_members_block ::= '{' inner_attr* impl_member* '}' {pin=1}
```
matklad: We tend to use more spaces in attribues: `{ pin = 1 }`

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -357,13 +351,16 @@ private block_item_group ::= &(CONST? UNSAFE? extern_abi? FN) fn_item
...
+fake params ::= self_argument? param*
+
 private fn_header ::= CONST? UNSAFE? extern_abi? FN IDENTIFIER
 
 extern_abi ::= EXTERN STRING_LITERAL?
 
-fn_params ::= '(' !',' params ','? ')' { pin(".*") = 1 }
-private params ::= [ param (',' param)* ]
-param ::= pat ':' type
+fn_params ::= '(' !',' fn_params_ ','? ')' { pin(".*")=1 elementType=params }
```
matklad: Again, I think for consistency it should be formatted as

```
... {
  pin(".*") = 1
  elementType = params
}
```

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
matklad: Well, it may be a wrong decision, because everybody else don't uses spaces here. But it is definitely not a good idea to change code style during functional changes.

matklad:
Have you verified that generated PSI classes have all the necessary accessors?

src/main/kotlin/org/rust/ide/structure/RustImplTreeElement.kt
```diff
@@ -10,10 +10,8 @@ import org.rust.lang.core.psi.visitors.RecursiveRustVisitor
 class RustImplTreeElement(element: RustImplItem) : PsiTreeElementBase<RustImplItem>(element) {
 
     override fun getPresentableText(): String? {
-        val type: RustType = element?.typeList?.firstOrNull() ?: return "<unknown>"
```
matklad: Oh my god, it was **that** bad...

matklad:
I'll take a closer look tomorrow, but it is definitely a marked improvement :)

However, I do think that this could have been split in several more easily reviewable PR ;)

jajakobyly:
> However, I do think that this could have been split in several more easily reviewable PRs ;)

My intention was not to spam with millions of PRs and put instead each distinguishable change into separate commit, so you could review it using that `Show changes from [...]` select box.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -254,6 +247,13 @@
     implements  ("use_glob") = "org.rust.lang.core.psi.RustNamedElement"
     mixin       ("use_glob") = "org.rust.lang.core.psi.impl.mixin.RustUseGlobImplMixin"
 
+    // HACK(jajakobyly): Grammar-Kit is unable to simply generate accesors for tokens
+    //                   created by external rules. We need to work around this limitation.
+    //                   Fortunately, all "external" tokens are only used inside binary
+    //                   expressions, so we can inject accessors for them using mixins.
+    implements  ("binary_expr") = "org.rust.lang.core.psi.impl.RustBinaryExprExternalTokens"
+    mixin       ("binary_expr") = "org.rust.lang.core.psi.impl.mixin.RustBinaryExprImplMixin"
```
jajakobyly: Ugh, this one is nasty, but I couldn't find any other way to accomplish this.

matklad:
> so you could review it using that Show changes from [...] select box.

While I can view commits separately, I can not test them one by one, because the latter commits fix the earlier ones. And I can not merge them one by one either. Also, amending the last commit is easier then fix-upping the one in the middle, and I love pretty history :)

I really love the changes here, but I just don't feel comfortable merging them all at once, without careful review. Can we possibly create several PRs? 

* impl members block
* unification of fn params
* oror work
* macro work

Sorry, I should have noticed that this PR is getting out of hands earlier ;(

jajakobyly:
Sure, no problem

PR #330, opened 2016-04-13
(INTENTION, FIX) Fixed add derive intention in presence of pub modifiers

### LINKED ISSUE #329: Fix `add derive clause` intention if visibility is present
```
pub struct S;
```

`Alt+Enter, add derive clause`

```
pub #[derive()]struct S;
```

@netvl you may be interested in this :)

netvl:
Fixes #329 

@matklad please review

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
@@ -19,7 +19,7 @@ class AddDerive : PsiElementBaseIntentionAction()
 
     override fun invoke(project: Project, editor: Editor?, element: PsiElement) {
         val item = element.parentOfType<RustItem>() ?: return
-        val keyword = getStructOrEnumKeyword(item) ?: return
+        val keyword = getLeftmostStructOrEnumKeyword(item) ?: return
 
         val deriveAttr = findOrCreateDeriveAttr(project, item, keyword) ?: return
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/intentions/AddDerive.kt
```diff
@@ -31,12 +31,12 @@ class AddDerive : PsiElementBaseIntentionAction()
...
         }
         val item = element.parentOfType<RustItem>() ?: return false
-        return getStructOrEnumKeyword(item) != null
+        return getLeftmostStructOrEnumKeyword(item) != null
     }
 
-    private fun getStructOrEnumKeyword(item: RustItem): PsiElement? = when (item) {
-        is RustStructItem -> item.struct
-        is RustEnumItem   -> item.enum
+    private fun getLeftmostStructOrEnumKeyword(item: RustItem): PsiElement? =when (item) {
```
matklad: A space is missing after `=`. 

Another way to write this is to return `firstChild`. I'm not sure which one is better.

netvl: Thanks, missed it.

As for `firstChild`, I'm not sure it would work. It would mean inserting the new attribute before all other attributes, for example. I'd personally prefer it to be the "inmost" one.

netvl: Fixed.

matklad:
Thanks!

PR #337, opened 2016-04-17
(GRAM, PSI) Remove type_ascription

jajakobyly:
This rule brings some inconsistency in our PSI structure. Let's ~~remove it~~ make it private.


matklad:
I think it's better to go in the opposite direction and add `type_ascription` everywhere. The benefit is that `[':' type]` does not enfore the invariant that `:` and `type` must be *simultaneously* absent or present.

jajakobyly:
I was inspired by the way how it's done in Kotlin's PSI (there is no PSI node for type ascriptions), but yeah; making it private production is reasonable :)

matklad:
>making it private production is reasonable :)

Yes! This will allow us to easily change decision about privacy.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -280,6 +280,8 @@
     generateTokens=false
     generateTokenAccessors=true
 
+    // Preserve information about missing ':'
+    consumeTokenMethod("type_ascription") = "consumeToken"
```
jajakobyly: As fair as I understand from comparing `consumeToken` and `consumeTokenFast`, this line makes Grammar-Kit push information about missing `':'` to `ErrorState`, when consuming it fails at `type_ascription`. This information is retrieved when error message is generated in users of this production. That's why `':'` popped out in [`struct_pat.txt`](https://github.com/intellij-rust/intellij-rust/pull/337/files#diff-b899c29d69c1bb354a1c4e69ac12b213R44).

Fast consuming doesn't do this and instead quickly exits `type_ascription` parser when there is no `':'` to eat.

PR #339, opened 2016-04-19
(LEX, GRAM, T): Refactor handling of `||` and `|`

jajakobyly:
Currently `||` token is handled by lexer, which yields pretty ugly situation for lambdas without arguments. This PR moves responsibility for parsing `||` to parser which can perform contextual token collapsing.


src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -1117,6 +1117,8 @@ external gteq   ::= collapse GTEQ GT EQ       { name = ">=" }
 external ltlteq ::= collapse LTLTEQ LT LT EQ  { name = "<<=" }
 external ltlt   ::= collapse LTLT LT LT       { name = "<<" }
 external lteq   ::= collapse LTEQ LT EQ       { name = "<=" }
+external oror   ::= collapse OROR OR OR       { name = "||" }
+external or     ::= onlySingle OR             { name = "|" }
```
matklad: Are you sure that we need this rule? Seems like `bool_or_bin_expr`  has a higher precedence then `bit_or_bin_expr` and so we don't need to check that `|` is single. I had not tested this and I always  forget which is high and which is low precedence, so I might be mistaking here :)

jajakobyly: This rule is specifically because of this. Without this check, the parser always matches `bit_or_bin_expr` and throws errors for `||`.

matklad: And if we do need this, then a test will be useful!

jajakobyly: `oror.rs` checks for this. It's a stripped down version of some file from `ctrs` which caught this.

matklad: And it would be a good idea to add a test which combines `|`  and `||` to make sure we actually handle precedence right!

jajakobyly: Ok

matklad: So as the order of rules has changed  maybe we don't need `or` any more?

And, if we still need it, I bet it can be defined simply as

```
bit_or_bin_expr     ::= restricted_expr '|' !'|' restricted_expr
```

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -805,6 +804,8 @@ left view_path_part ::= '::' (SUPER | IDENTIFIER) {
 //
 expr ::=
 
+// https://doc.rust-lang.org/reference.html#operator-precedence
+// https://github.com/JetBrains/Grammar-Kit/blob/master/HOWTO.md#24-compact-expression-parsing-with-priorities
```
matklad: The nice description of the algorithm is here: http://journal.stuffwithstuff.com/2011/03/19/pratt-parsers-expression-parsing-made-easy/

PR #340, opened 2016-04-19
(GRAM, PSI): refactor macros parsing & PSI

jajakobyly:
This PR tries to improve how we parse and represent macros. See commits for detailed changelog.


matklad:
Having both `macro_item` and `item_macro` is confusing. I'll try to come up with a better name for them.

jajakobyly:
I thought about sth like `item_like_macro` but I didn't like it.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -999,17 +1003,31 @@ lit_expr ::= lit
...
+macro_definition ::= macro_rules_invocation IDENTIFIER item_macro_arg {
+  pin = 1
+  extends = item_macro
 }
 
+private build_in_macro ::= try_macro
+                         | format_like_macro
+
+private macro_call_ ::= build_in_macro | macro_call { pin(".*") = macro_invocation }
+private macro_item_ ::= macro_definition | item_macro { pin(".*") = macro_invocation }
```
matklad: I think we'd better avoid trailing `_` in rule names: https://github.com/JetBrains/Grammar-Kit/issues/97

matklad:
What do you think about the following renames

`macro_call -> expr_like_macro`

`item_macro -> item_like_macro`

?

jajakobyly:
Looks fine

PR #342, opened 2016-04-20
(T): use rust sources to test parser performance

matklad:
This adds the first performance test. The test simply executes the parser on the compiler's source code.

### DIFF
--- a/src/test/kotlin/org/rust/Performance.kt
+++ b/src/test/kotlin/org/rust/Performance.kt
@@ -1,0 +1,6 @@
+package org.rust
+
+/**
+ * Marker interface for tests category
+ */
+interface Performance
--- a/src/test/kotlin/org/rust/lang/RustTestCaseBase.kt
+++ b/src/test/kotlin/org/rust/lang/RustTestCaseBase.kt
@@ -61,14 +61,6 @@ abstract class RustTestCaseBase : LightPlatformCodeInsightFixtureTestCase(), Rus
     protected fun getVirtualFileByName(path: String): VirtualFile? =
         LocalFileSystem.getInstance().findFileByPath(path)
 
-    companion object {
-        @JvmStatic
-        fun camelToSnake(camelCaseName: String): String =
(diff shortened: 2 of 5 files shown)


matklad:
@jajakobyly please take a look

src/test/kotlin/org/rust/Performance.kt
```diff
@@ -0,0 +1,7 @@
+package org.rust
+
+/**
+ * Marker interface for tests category
+ */
+interface Performance {
+}
```
jajakobyly: This braces can be omitted :smile:

src/test/kotlin/org/rust/lang/RustTestCaseBase.kt
```diff
@@ -68,6 +68,16 @@ abstract class RustTestCaseBase : LightPlatformCodeInsightFixtureTestCase(), Rus
...
                 .joinToString("_")
+
+        @JvmStatic
+        fun rustSources(): VirtualFile {
+            val sdkArchiveFile = LocalFileSystem.getInstance()
+                .findFileByPath("${RustTestCase.testResourcesPath}/rustc-src.zip")
+
+            return checkNotNull(sdkArchiveFile?.let { JarFileSystem.getInstance().getJarRootForLocalFile(it) }) {
+                "Rust sources archive not found. Run `./gradlew test` to download the archive."
+            }
```
jajakobyly: Pretty much the same thing is also done inside `WithStdlibRustProjectDescriptor#testCargoProject`

jajakobyly:
:+1:

jajakobyly:
It might be nice to run this, let's say, every 10th CI build.

matklad:
It would be run on every CI build. I'm not sure how robust are time measurements though, they are a bit high at the moment. I hope to run `./gradle performanceTest` manually sometimes.

matklad:
It's about time to stop waffling about it :) 

I've removed actual timing assertions because I don't believe they'll well with our current CI infra.

matklad:
Yay, down to single branch again :tada:

PR #343, opened 2016-04-21
(GRAM): force accessor generation

matklad:
This forces grammar kit to generate proper accessors for expresisons.

Before:

```Java

public interface RustIfExpr extends RustExpr {

  @NotNull
  List<RustBlock> getBlockList();

  @Nullable
  RustExpr getExpr(); // this is the `then` branch, condition is missing

  @Nullable
  PsiElement getElse(); 

  @NotNull
  PsiElement getIf();

}

public interface RustParenExpr extends RustExpr {

  @NotNull
  PsiElement getLparen();

  @NotNull
  PsiElement getRparen();

}
```

After:

```Java
public interface RustIfExpr extends RustExpr {

  @NotNull
  List<RustBlock> getBlockList();

  @NotNull
  List<RustExpr> getExprList();

  @Nullable
  PsiElement getElse();

  @NotNull
  PsiElement getIf();

}

public interface RustParenExpr extends RustExpr {

  @Nullable
  RustExpr getExpr();

  @NotNull
  PsiElement getLparen();

  @NotNull
  PsiElement getRparen();

}
```

@alexeykudinkin @jajakobyly what do you think?

It's a horrible hack, but the tests pass.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
+++ b/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
@@ -833,8 +833,9 @@ restricted_expr ::= ret_expr
 }
 
 // https://github.com/rust-lang/rfcs/blob/master/text/0092-struct-grammar.md
-private no_struct_lit_expr  ::= << withoutStructLiterals restricted_expr >>
-private any_expr            ::= << withStructLiterals restricted_expr >>
+private no_struct_lit_expr  ::= << withoutStructLiterals restricted_expr >> | never restricted_expr
+private any_expr            ::= << withStructLiterals restricted_expr >>    | never restricted_expr
+private never ::= !()
 
 
 block_expr ::= while_expr


src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -109,6 +109,8 @@ object RustParserUtil : GeneratedParserUtilBase() {
         }
     }
 
+    @JvmStatic fun never(b: PsiBuilder, level: Int) : Boolean = false
```
jajakobyly: nice

jajakobyly: Kotlin compiler screams about these arguments being unused.

jajakobyly:
Generated parser code looks "fine":
```java
  /* ********************************************************** */
  // << withStructLiterals restricted_expr >>    | << never >> restricted_expr
  static boolean any_expr(PsiBuilder builder_, int level_) {
    if (!recursion_guard_(builder_, level_, "any_expr")) return false;
    boolean result_;
    Marker marker_ = enter_section_(builder_);
    result_ = withStructLiterals(builder_, level_ + 1, restricted_expr_parser_);
    if (!result_) result_ = any_expr_1(builder_, level_ + 1);
    exit_section_(builder_, marker_, null, result_);
    return result_;
  }

  // << never >> restricted_expr
  private static boolean any_expr_1(PsiBuilder builder_, int level_) {
    if (!recursion_guard_(builder_, level_, "any_expr_1")) return false;
    boolean result_;
    Marker marker_ = enter_section_(builder_);
    result_ = never(builder_, level_ + 1);
    result_ = result_ && restricted_expr(builder_, level_ + 1, -1);
    exit_section_(builder_, marker_, null, result_);
    return result_;
  }
```

But it's worth notice that this method does not give us a control of nullability annotations. Though the only productions I can find where expressions are required (and so `@NotNull`s fit better) are `tuple_expr` and `paren_expr`

matklad:
Looks like I subconsciously stolen the trick form the Go plugin: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/grammars/go.bnf#L188. `!()` is better then an external rule

jajakobyly:
This trick also seems to work for `GTGTEQ` and the rest

matklad:
@jajakobyly will you submit a PR? I think it should be something like

```
private gtgtgt ::= <<collapse GT GT GT>> | never GTGTGT
```

but I have not checked this.

jajakobyly:
Yep exactly

PR #344, opened 2016-04-21
(GRAM, PSI, T): Unify params

jajakobyly:
This PR aims to improve how we represent various kinds of function parameters. They are now represented by single universal PSI classes: `RustParameter`, `RustParameters` with unified and expectable layout of tokens inside them.


src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -758,7 +755,7 @@ private path_generic_args_without_colons_part_inner ::= ident_or_self_or_super [
 //    <T>::c<V,W>
 //    <T as Foo>::c(V) -> W
 //
-private meta qualified_path ::= '<' type [ AS trait_ref ] '>' '::' << param >>
+private meta qualified_path ::= '<' type [ AS trait_ref ] '>' '::' << fn_param >>
```
matklad: `<< param >>` here is an argument of the meta rule and not the `param ::= ` rule

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -1124,6 +1122,6 @@ external lteq   ::= collapse LTEQ LT EQ       { name = "<=" }
 external oror   ::= collapse OROR OR OR       { name = "||" }
 
 // Trailing commas are allowed
-private meta comma_separated_list ::= <<param>> ( ',' <<param>> ) * ','?
+private meta comma_separated_list ::= <<fn_param>> ( ',' <<fn_param>> ) * ','?
```
matklad: ditto

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
matklad: Maybe we should rename `*param` to `*parameter`? I'd rather use long names for things which do not have corresponding Rust term.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -365,13 +361,17 @@ private block_item_group ::= &(CONST? UNSAFE? extern_abi? FN) fn_item
 // Functions
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 
+// TODO: Maybe enforce @NotNull's in specific cases
+fake param  ::= pat? type?
+fake params ::= self_argument? param*
+
```
matklad: Maybe we should move all descendants of `param` here?

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
jajakobyly: Or maybe `args` or `arguments`?

jajakobyly: I'd like to have some short here, as `parameters.parameterList` is pretty looong.

matklad: We already have `generic_args` and `call_args` =( 

Another (more obscure) option is `formals`.

jajakobyly: Actually I don't see any big naming collisions with them.

jajakobyly: Hmm and suddenly accessors for arguments of `comma_separated_list` have disappeared when I have tried to fix it.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
jajakobyly: Hmm and suddenly accessors for arguments of `comma_separated_list` have disappeared when I have tried to fix it.

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
jajakobyly: Ugh, I get the collision now: `generic_args` and `generic_params`

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
jajakobyly: Ok I get it. Grammar-Kit gets crazy when it's fed with `<< param >>` instead of `<<param>>` (notice whitespaces). Would you like me to fix all such occurrences in grammar in this PR?

matklad: >Grammar-Kit gets crazy when it's fed with << param >> instead of <<param>> (notice whitespaces)

I have no emoji for this.

Yes, can you change all occurrences of `<< foo >>` to `<<foo>>` to align with more common grammar kit style. I would have done it myself, but I don't want to interfere with your other PRs.

Thanks a lot for all this work!

PR #348, opened 2016-04-23
`CargoProjectWorkspace` refactoring

alexeykudinkin:
1. Made module-component instead of a service (as soon as it gets a state, it should not be service anymore).
2. Killed persistence and made it transparent one-directional proxy of the `Cargo.toml` project-workspace.
3. Refactored some concurrency issues (and hopefully didn't introduced new ones).

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -11,7 +11,7 @@ interface CargoProjectWorkspace {
     /**
      * Updates Rust libraries asynchronously. Consecutive requests are coalesced.
      */
-    fun requestUpdate(toolchain: RustToolchain, immediately: Boolean = false)
+    fun requestUpdateUsing(toolchain: RustToolchain, immediately: Boolean = false)
 
     /**
      * Latest version of the Cargo's project-description obtained
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -23,15 +23,7 @@ interface CargoProjectWorkspace {
 
     /**
      * Latest version of the Cargo's project-description obtained
-     *
-     * NOTA BENE: In the current implementation it's SYNCHRONOUS
      */
     val projectDescription: CargoProjectDescription?
-
-    /**
-     * Subscribes given listener to the supplied topic on the [CargoProjectWorkspace]'s private
-     * message-bus
-     */
-    fun <L: Any> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable? = null)
 
 }
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -1,5 +1,8 @@
 package org.rust.cargo.project.workspace
 
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.util.Disposer
+import com.intellij.util.messages.Topic
 import org.rust.cargo.project.CargoProjectDescription
 import org.rust.cargo.toolchain.RustToolchain
 
@@ -24,5 +27,10 @@ interface CargoProjectWorkspace {
      * NOTA BENE: In the current implementation it's SYNCHRONOUS
      */
     val projectDescription: CargoProjectDescription?
+
+    /**
+     * Subscribes given listener to the supplied topic
+     */
+    fun <L> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable? = null)
 
 }
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -1,5 +1,6 @@
 package org.rust.cargo.project.workspace
 
+import com.intellij.util.messages.Topic
 import org.rust.cargo.project.CargoProjectDescription
 import org.rust.cargo.toolchain.RustToolchain
 
@@ -7,17 +8,21 @@ import org.rust.cargo.toolchain.RustToolchain
  * Uses `cargo metadata` command to update IDEA libraries and Cargo project model.
  */
 interface CargoProjectWorkspace {
-    /**
-     * Updates Rust libraries asynchronously. Consecutive updates are coalesced.
-     */
-    fun scheduleUpdate(toolchain: RustToolchain)
 
     /**
-     * Immediately schedules an update. Shows balloon upon completion.
+     * Updates Rust libraries asynchronously. Consecutive requests are coalesced.
+     */
+    fun requestUpdate(toolchain: RustToolchain, immediately: Boolean = false)
+
+    /**
+     * Latest version of the Cargo's project-description obtained
      *
-     * Update is still asynchronous.
+     * NOTA BENE: In the current implementation it's SYNCHRONOUS
      */
-    fun updateNow(toolchain: RustToolchain)
+    val projectDescription: CargoProjectDescription
 
-    val projectDescription: CargoProjectDescription?
+    companion object {
+        val UPDATES = Topic("org.rust.CargoProjectUpdatesTopic", CargoProjectWorkspaceListener::class.java)
+    }
+
 }
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -4,7 +4,12 @@ import org.rust.cargo.project.CargoProjectDescription
 import org.rust.cargo.toolchain.RustToolchain
 
 /**
- * Uses `cargo metadata` command to update IDEA libraries and Cargo project model.
+ * Cargo based project's workspace abstraction insulating inter-op with the `cargo` & `Cargo.toml`
+ * itself. Uses `cargo metadata` sub-command to update IDEA module's & project's model.
+ *
+ * Quite low-level in its nature & follows a soft-fail policy, i.e. provides access
+ * for latest obtained instance of the [CargoProjectDescription], though doesn't assure that this
+ * one could be obtained (consider the case with invalid, or missing `Cargo.toml`)
  */
 interface CargoProjectWorkspace {
 
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -29,7 +29,8 @@ interface CargoProjectWorkspace {
     val projectDescription: CargoProjectDescription?
 
     /**
-     * Subscribes given listener to the supplied topic
+     * Subscribes given listener to the supplied topic on the [CargoProjectWorkspace]'s private
+     * message-bus
      */
     fun <L: Any> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable? = null)
 
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -31,6 +31,6 @@ interface CargoProjectWorkspace {
     /**
      * Subscribes given listener to the supplied topic
      */
-    fun <L> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable? = null)
+    fun <L: Any> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable? = null)
 
 }
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
@@ -1,6 +1,5 @@
 package org.rust.cargo.project.workspace
 
-import com.intellij.util.messages.Topic
 import org.rust.cargo.project.CargoProjectDescription
 import org.rust.cargo.toolchain.RustToolchain
 
@@ -19,10 +18,6 @@ interface CargoProjectWorkspace {
      *
      * NOTA BENE: In the current implementation it's SYNCHRONOUS
      */
-    val projectDescription: CargoProjectDescription
-
-    companion object {
-        val UPDATES = Topic("org.rust.CargoProjectUpdatesTopic", CargoProjectWorkspaceListener::class.java)
-    }
+    val projectDescription: CargoProjectDescription?
 
 }
--- a/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
@@ -76,8 +76,6 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
         messageBus = MessageBusFactory.newMessageBus(CargoProjectWorkspace::class.java.name, module.messageBus)
 
         subscribeTo(VirtualFileManager.VFS_CHANGES, FileChangesWatcher())
-
-        subscribeTo(CargoProjectWorkspaceListener.Topics.UPDATES, ProjectModelUpdater())
     }
 
     override fun disposeComponent() {
@@ -132,6 +130,24 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
                 .onWorkspaceUpdateCompleted(r)
     }
 
+    private fun updateModuleDependencies(r: UpdateResult) {
+        ApplicationManager.getApplication().assertIsDispatchThread()
+
+        when (r) {
+            is UpdateResult.Ok ->
+                ApplicationManager.getApplication().runWriteAction {
+                    if (!module.isDisposed) {
+                        val libraryRoots =
+                            r.projectDescription.packages
+                                .filter { !it.isModule }
+                                .mapNotNull { it.virtualFile }
+
+                        module.updateLibrary(module.cargoLibraryName, libraryRoots)
+                    }
+                }
+        }
+    }
+
     /**
      * Subscribes given listener to the supplied topic on the [CargoProjectWorkspace]'s private
      * message-bus
@@ -184,6 +200,8 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
             val r = requireNotNull(result)
 
             lock (lock) {
+                updateModuleDependencies(r);
+
                 cached = when (r) {
                     is UpdateResult.Ok -> r.projectDescription
                     else               -> null
@@ -195,28 +213,6 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
             when (r) {
                 is UpdateResult.Ok  -> LOG.info("Project '${module.project.name}' successfully updated")
                 is UpdateResult.Err -> LOG.info("Project '${module.project.name}' update failed", r.error)
-            }
-        }
-    }
-
-    /**
-     * IDEA's project model updater
-     */
-    inner class ProjectModelUpdater : CargoProjectWorkspaceListener {
-
-        override fun onWorkspaceUpdateCompleted(r: UpdateResult) {
-            when (r) {
-                is UpdateResult.Ok ->
-                    ApplicationManager.getApplication().runWriteAction {
-                        if (!module.isDisposed) {
-                            val libraryRoots =
-                                r.projectDescription.packages
-                                    .filter { !it.isModule }
-                                    .mapNotNull { it.virtualFile }
-
-                            module.updateLibrary(module.cargoLibraryName, libraryRoots)
-                        }
-                    }
             }
         }
     }
--- a/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
@@ -3,6 +3,7 @@ package org.rust.cargo.project.workspace.impl
 import com.intellij.execution.ExecutionException
 import com.intellij.execution.process.ProcessAdapter
 import com.intellij.execution.process.ProcessEvent
+import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.ModalityState
 import com.intellij.openapi.diagnostic.Logger
@@ -11,13 +12,14 @@ import com.intellij.openapi.module.ModuleComponent
 import com.intellij.openapi.module.ModuleUtilCore
 import com.intellij.openapi.progress.ProgressIndicator
 import com.intellij.openapi.progress.Task
-import com.intellij.openapi.ui.MessageType
-import com.intellij.openapi.ui.popup.util.PopupUtil
+import com.intellij.openapi.util.Disposer
 import com.intellij.openapi.util.Key
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapi.vfs.newvfs.BulkFileListener
 import com.intellij.openapi.vfs.newvfs.events.VFileEvent
 import com.intellij.util.Alarm
+import com.intellij.util.messages.MessageBus
+import com.intellij.util.messages.MessageBusFactory
 import com.intellij.util.messages.Topic
 import org.jetbrains.annotations.TestOnly
 import org.rust.cargo.project.CargoProjectDescription
@@ -29,7 +31,6 @@ import org.rust.cargo.project.workspace.CargoProjectWorkspaceListener.UpdateResu
 import org.rust.cargo.toolchain.RustToolchain
 import org.rust.cargo.util.cargoLibraryName
 import org.rust.cargo.util.cargoProjectRoot
-import org.rust.cargo.util.enqueue
 import org.rust.cargo.util.updateLibrary
 import org.rust.lang.utils.lock
 import java.util.concurrent.locks.Lock
@@ -61,19 +62,27 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
      */
     private var cached: CargoProjectDescription? = null
 
+    /**
+     * Isolated message-bus to insulate cargo-project-workspace related messaging
+     * infra from general-purpose listeners, and properly manage listeners' lifetimes
+     */
+    private var messageBus: MessageBus? = null
+
     /** Component hooks */
 
     override fun getComponentName(): String = "org.rust.cargo.CargoProjectWorkspace"
 
     override fun initComponent() {
+        messageBus = MessageBusFactory.newMessageBus(CargoProjectWorkspace::class.java.name, module.messageBus)
+
         subscribeTo(VirtualFileManager.VFS_CHANGES, FileChangesWatcher())
 
         subscribeTo(CargoProjectWorkspaceListener.Topics.UPDATES, ProjectModelUpdater())
-        //subscribeTo(CargoProjectWorkspaceListener.Topics.UPDATES, Notifier())
     }
 
     override fun disposeComponent() {
-        alarm.dispose()
+        alarm       .dispose()
+        messageBus!!.dispose()
     }
 
     override fun projectClosed() { /* NOP */ }
@@ -124,6 +133,20 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
             module.messageBus
                 .syncPublisher(CargoProjectWorkspaceListener.Topics.UPDATES)
                 .onWorkspaceUpdateCompleted(r)
+    }
+
+    /**
+     * Subscribes given listener to the supplied topic on the [CargoProjectWorkspace]'s private
+     * message-bus
+     */
+    override fun <L> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable?) {
+        val conn = messageBus!!
+            .connect().let {
+                it.subscribe(t, listener!!)
+                it
+            }
+
+        disposer?.let { Disposer.register(disposer, conn) }
     }
 
     private inner class UpdateTask(
@@ -179,23 +202,6 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
     }
 
     /**
-     * Project-update task's notifier
-     */
-    inner class Notifier : CargoProjectWorkspaceListener {
-
-        override fun onWorkspaceUpdateCompleted(r: UpdateResult) {
-            when (r) {
-                is UpdateResult.Ok  -> showBalloon("Project '${module.project.name}' successfully updated!", MessageType.INFO)
-                is UpdateResult.Err -> showBalloon("Project '${module.project.name}' update failed: ${r.error.message}", MessageType.ERROR)
-            }
-        }
-
-        private fun showBalloon(message: String, type: MessageType) {
-            PopupUtil.showBalloonForActiveComponent(message, type)
-        }
-    }
-
-    /**
      * IDEA's project model updater
      */
     inner class ProjectModelUpdater : CargoProjectWorkspaceListener {
@@ -237,10 +243,6 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
                 requestUpdateUsing(toolchain)
             }
         }
-    }
-
-    private inline fun <reified L> subscribeTo(t: Topic<L>, listener: L) {
-        module.messageBus.connect().subscribe(t, listener!!)
     }
 
     @TestOnly
--- a/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
@@ -101,9 +101,6 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
      */
     override fun requestUpdateUsing(toolchain: RustToolchain, immediately: Boolean) {
         val contentRoot = module.cargoProjectRoot ?: return
-
-        val delay = if (ApplicationManager.getApplication().isUnitTestMode) 0 else DELAY
-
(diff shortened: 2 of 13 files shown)


alexeykudinkin:
@matklad take a look, please.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -20,110 +18,128 @@ import com.intellij.openapi.vfs.VirtualFileManager
...
-    private val DELAY_MILLIS = 1000
+    private val DELAY = 1000 /* milliseconds */
+
+    private val LOG = Logger.getInstance(CargoProjectWorkspaceImpl::class.java)
 
     /**
      * Alarm used to coalesce consecutive update requests.
-     * It uses EDT thread, but the tasks are really tiny and
-     * only spawn background update.
+     * It uses pooled thread.
```
matklad: false

src/main/kotlin/org/rust/cargo/util/TasksUtil.kt
```diff
@@ -0,0 +1,59 @@
...
+
+object TasksUtil
+
+/**
+ * Schedules task to run asynchronously (on the application thread-pool)
+ *
+ * NOTA BENE: This would occupy 2 (sic!) threads: one carrying payload, and the second one to piggyback results of the
+ *            former right into `ListenableFuture`. We need to address that
+ */
+fun Task.Backgroundable.enqueue(): ListenableFuture<*> =
```
matklad: Looks like this method is not actually used in a meaningful way. The result of `.enqueue` is discarded, so it should be possible to use plain `.queue` instead.

src/main/kotlin/org/rust/cargo/util/TasksUtil.kt
```diff
@@ -0,0 +1,59 @@
...
+        (ProgressManager.getInstance() as CoreProgressManager).runProcessWithProgressAsynchronously(this),
+        AppExecutorUtil.getAppExecutorService()
+        )
+
+
+/**
+ * Allows to bind one future to the other one, 'piggybacking' results
+ * of the former into the latter one
+ */
+inline fun <reified T> ListenableFuture<T>.flowInto(f: SettableFuture<T>) {
```
matklad: Next two functions are unused

src/test/kotlin/org/rust/cargo/toolchain/CargoProjectResolveTestCase.kt
```diff
@@ -9,31 +9,50 @@ import org.assertj.core.api.Assertions.assertThat
...
+                CargoProjectWorkspaceListener.Topics.UPDATES,
+                object: CargoProjectWorkspaceListener {
+                    override fun onWorkspaceUpdateCompleted(r: UpdateResult) {
+                        assertThat(r is UpdateResult.Ok)
+
+                        val reference = extractReference("src/main.rs")
+                        assertThat(reference.resolve()).isNotNull()
+                    }
+                })
+
```
matklad: Async tests are tricky. If the message is never delivered, this code never gets executed and the test pass happily. 

In essence, what we want to test here is that the message is actually delivered. This is a hard problem.

src/main/kotlin/org/rust/lang/utils/Utils.kt
```diff
@@ -31,3 +32,43 @@ fun <T> using(d: Disposable, block: () -> T): T {
...
+
+/**
+ * Helper executing the block supplied under the lock being RELEASED,
+ * and ACQUIRED upon completion (successful or not), i.e. it performs in the
+ * way exactly opposite to [lock]
+ *
+ * @l       Lock to be release
+ * @block   Target block to be run 'under' the lock being released
+ */
+fun <T> release(l: Lock, block: () -> T): T =
```
matklad: unused

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -138,54 +154,102 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
...
+
+        private fun showBalloon(message: String, type: MessageType) {
+            PopupUtil.showBalloonForActiveComponent(message, type)
+        }
+    }
+
+    /**
+     * IDEA's project model updater
+     */
+    inner class ProjectModelUpdater : CargoProjectWorkspaceListener {
```
matklad: I believe it is a bad idea to use listeners if you can call the method directly. This listener only introduces complexity. Why not update the workspace before invoking listeners? 

Also this introduce a subtle race condition: if you are another listener, and receive an update with `projectDescripton`, you don't know if libraries were updated or not. 

Listeners should be invoked after all state changes has happened.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -138,54 +154,102 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
...
+                is UpdateResult.Ok  -> LOG.info("Project '${module.project.name}' successfully updated")
+                is UpdateResult.Err -> LOG.info("Project '${module.project.name}' update failed", r.error)
+            }
+        }
+    }
+
+    /**
+     * Project-update task's notifier
+     */
+    inner class Notifier : CargoProjectWorkspaceListener {
```
matklad: This one is tricky actually. I think that showing balloons after automatic updates is a bad idea. Positive result does not give any information, and negative result can legitimately happen if, say, the user saved `Cargo.toml` in the middle of editing. 

On the other hand, showing feedback after explicitly requested update is the right thing to do.

So I think a `Listener` not a right choice here, because, first of all, you can just use a direct method call and, second, the *caller* of the `requestUpdate` must decide if the balloon is necessary.

matklad: Original motivation for the previous design is in this commit message: https://github.com/intellij-rust/intellij-rust/commit/e7b177b129095ed302d08cad192bd96920b4e12d

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -20,110 +18,128 @@ import com.intellij.openapi.vfs.VirtualFileManager
...
     private val alarm = Alarm()
 
-    private var cargoProjectState: CargoProjectState by Delegates.observable(CargoProjectState()) {
-        prop, old, new ->
-            projectDescription = new.projectData?.let { CargoProjectDescription.deserialize(it) }
-    }
+    /**
+     * Lock to guard reads/updates to [cached]
+     */
+    private var lock: Lock = ReentrantLock()
```
matklad: Looks are hard. I don't feel confident using looks (unless in Rust :) ), so I think it is a good idea to avoid them. 

And I think that [AtomicReferece](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicReference.html) is a much simpler synchronization primitive which can be used here. 

Why is it sufficient?  Tasks are executed in serialized order and, moreover, `onSuccess` handler is invoked on `EDT` thread. And `onSuccess` is (or should be) the only place where we modify `cached`. So we have a single writer / multiple readers situation, so we don't need to enforce mutual exclusion and only need to care for liveness. So `AtomicReference` fits the bill, and, what is crusually important guarantees the absence of dead locks by construction. Another interesting property is that you can't forget to lock the lock if you use atomic.

matklad: And yes, the previous implementation which used non volatile non atomic variable was broken ;(

src/main/kotlin/org/rust/lang/utils/Utils.kt
```diff
@@ -31,3 +32,43 @@ fun <T> using(d: Disposable, block: () -> T): T {
...
     }
 }
+
+/**
+ * Helper disposing [d] upon completing the execution of the [block] (under the [d])
+ *
+ * @d       Target `Disposable` to be disposed upon completion of the @block
+ * @block   Target block to be run prior to disposal of @d
+ */
+fun <D: Disposable, T> usingWith(d: D, block: (D) -> T): T {
```
matklad: unused

src/main/kotlin/org/rust/cargo/util/TasksUtil.kt
alexeykudinkin: Those are preserved on purpose.

src/main/kotlin/org/rust/cargo/util/TasksUtil.kt
alexeykudinkin: Good catch. Though i (personally) think some IDEA's API is a pure crap, the overhead for this facade is too big.

src/test/kotlin/org/rust/cargo/toolchain/CargoProjectResolveTestCase.kt
alexeykudinkin: Good catch.

src/main/kotlin/org/rust/lang/utils/Utils.kt
alexeykudinkin: Same -- saved on purpose.

src/main/kotlin/org/rust/cargo/util/TasksUtil.kt
matklad: Ok, but http://nedbatchelder.com/text/deleting-code.html.

I would say YAGNI if I didn't hate this acronym :)

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: Because, single purpose -- single responsibility. `UpdateTask` should do just `cargo metadata` parsing and pushing the conveyor, the rest isn't its responsibility. It should not serve a kitchen combiner purpose. Having things small makes it easily pluggable & maintainable.

No, there is no race-condition. API is low-level & its users may not make any assumptions about the current project-model state.

matklad: Are you sure that this listeners are not leaked? 

>API is low-level 

Listening to a topic does not seem like a low-level API. This looks more like leaky abstraction. The "correct" solution would be to have two topics: one for private implementation, like notifications and libraries update, and another for public listeners. Then you'd invoke private listeners first, and then the public ones.

>Having things small makes it easily pluggable & maintainable.

Plugin an extra method call is the same amount of work as plugin an extra listener.

But this is not a major issue anyway.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: No, we can't `cached` update & listeners notification should go in concert.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: All of those usages are 'private' in a sense. The bad point here may be to pass the messages through the general-purpose bus.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
matklad: Calling arbitrary listeners while holding a lock is asking for trouble.

But I steel don't get it. If you listener does some async work then the lock has no effect. If it does some sync work, then, well, `onSucess` calls are serialized and you don't need the lock.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: > Then you'd invoke private listeners first, and then the public ones.
There should be no ordering assumed on listeners, if not provided with the means to manage that.

matklad: Yes, this is exactly the problem. Some private guys chat on the public topic which may affect other listeners in unexpected way.

alexeykudinkin: This topic isn't for the general (high-level) use.

matklad: >There should be no ordering assumed on listeners, if not provided with the means to manage that.

Yes, unless you use separate private and public topics :)

matklad: >This topic isn't for the general (high-level) use.

Let's make it private then?

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: > Calling arbitrary listeners while holding a lock is asking for trouble.

It's only in the case you can accidentally reverse the order of acquisition of those. That's not the case.

> But I steel don't get it. If you listener does some async work then the lock has no effect. If it does some sync work, then, well, onSucess calls are serialized and you don't need the lock.

w/o the lock you'd abuse your listeners: them would be able to see `projectDescription` prior to `module` instance being updated.

matklad: >w/o the lock you'd abuse your listeners: them would be able to see projectDescription prior to module instance being updated.

What `module`? The lock is reentrant, so the listener still can just call `module.cargoProject`.

alexeykudinkin: Current module. 
Don't get your second statement. Reentrancy mean, that you can re-acquire it from the same thread, not that it's no-op.

matklad: I am sure that if I call `module.cargoProject` from a listener it will just happily reaquire the lock ([The method will return immediately if the current thread already owns the lock](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html)).

So the lock does not protect synchronous listeners from accessing `cargoProjectDescription`.

src/main/kotlin/org/rust/lang/utils/Utils.kt
```diff
@@ -31,3 +32,43 @@ fun <T> using(d: Disposable, block: () -> T): T {
...
+}
+
+/**
+ * Helper executing the block supplied under the lock being ACQUIRED,
+ * and RELEASED upon completion (successful or not)
+ *
+ * @l       Lock to be acquired
+ * @block   Target block to be run 'under' the lock being acquired
+ */
+fun <T> lock(l: Lock, block: () -> T): T =
```
matklad: Also, what do you think about "protect the data, not the code", API, like this:

```Kotlin
class Locked<T>(
    private val data
) {
    private val lock = Lock()

    fun read(block: T -> R): R {
        lock.lock()
        try {
            return body(data)
        } finally { lock.unlock() }
    }

    fun modify(block: T -> T) {
        lock.lock()
        try {
           data = block(data)
        } finally { lock.unlock() }
    }
}
```

It would make "Lock to guard reads/updates to [cached]" comments unnecessary and would make it harder to forget to lock the lock.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: > I am sure that if I call module.cargoProject from a listener it will just happily reaquire the lock (The method will return immediately if the current thread already owns the lock). So the lock does not protect synchronous listeners from accessing cargoProjectDescription.

Once again: what's that may be for? You can get `projectDescription` from the result supplied. And there is (and moreover, _should be_) NO high-level listeners subscribing for the `Cargo.toml` updates that may query `cargoProject`.

src/main/kotlin/org/rust/lang/utils/Utils.kt
alexeykudinkin: > It would make "Lock to guard reads/updates to [cached]" comments unnecessary and would make it harder to forget to lock the lock.

It wouldn't make it any harder.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
matklad: I though that this:

>w/o the lock you'd abuse your listeners: them would be able to see projectDescription prior to module instance being updated.

means: "we need lock to restrict what the listeners can do", and my argument is that "no, this lock does not impose any restrictions on listeners". I must have misunderstood that comment, but I still don't get it :(

src/main/kotlin/org/rust/lang/utils/Utils.kt
alexeykudinkin: Though i'm hugely for (reasonable) usage of this approach.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: Readers of course, not listeners. Sorry for misleading.

matklad: Ah, now I see, thanks!

But why would you want to block readers? This does not provide any guarantees. The following still can happen:

```Kotlin
// I am a reader

// I've locked a lock, read the projectDesctiption
val project = module.cargoProject
// released the lock

// cargo workspace update happens on the other thread

// `project` is now obsolete
foo(project)
```

And it does not look any worse then

```Kotlin
// I am a reader

// workspace update has set `cargoProject`

// I've atomically read current `cargoProject`
val project = module.cargoProject
foo(project)
// workspace update changed the libraries roots.
```

matklad:
It's quite a discussion here (which is good!) so here is the summary of my main concerns:

* The test must be fixed. It should actually fail if you place a bug into `WorkspaceImpl`.
* If the `TOPIC` is supposed to be private, it should be private.
* I have convinced myself that the lock here does not deadlock, but I still believe that `AtomicReference` would provide the same concurrency guarantees (`module.cargoProject` is or was a valid project description) and would be correct by construction. 
* Dead code like `enqueue` should be removed.
* Balloon notifications should be shown only for non-automatic updates.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: It does: it assures that user will not see a case when `projectDescription` isn't null, but `module` has no dependencies set (for ex.).

matklad: The same can be achieved if you update the libraries before you update the project description.

matklad: >It does: it assures that user will not see a case when projectDescription isn't null, but module has no dependencies set (for ex.).

And it does not assures it: the human can just delete any libraries we set up. It's very hard to protect with the lock the data you do not own.

alexeykudinkin: That's out of the scope, we're not pursuing coherency, just atomicity.

matklad: Ok, convinced. Sorry for the long conversation, but I just want to really understand all possible concurrency issues.

alexeykudinkin: No problem. Conversation is what this P/R was made for.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -20,110 +18,127 @@ import com.intellij.openapi.vfs.VirtualFileManager
...
      */
-    override fun scheduleUpdate(toolchain: RustToolchain) {
+    override fun requestUpdateUsing(toolchain: RustToolchain, immediately: Boolean) {
         val contentRoot = module.cargoProjectRoot ?: return
 
-        val task = UpdateTask(toolchain, contentRoot.path, showFeedback = false)
-        alarm.cancelAllRequests()
-        alarm.addRequest({ task.queue() }, DELAY_MILLIS)
-    }
+        val delay = if (ApplicationManager.getApplication().isUnitTestMode) 0 else DELAY
```
matklad: We use `immediatelly` in tests, so this presumably can be removed

alexeykudinkin:
> The test must be fixed. It should actually fail if you place a bug into WorkspaceImpl.

Agree. Fixed.

> If the TOPIC is supposed to be private, it should be private.

Kotlin unfortunately doesn't allow that-flexible visibility settings.

> Dead code like enqueue should be removed.

It's common-purpose _utility_ code that enforces good practices to be applied instead of bike-shedding every time someone stumbled upon the case not handled by the standard library. Won't be removed.

> Balloon notifications should be shown only for non-automatic updates.

Agree. Fixed.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -61,19 +62,27 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
      */
     private var cached: CargoProjectDescription? = null
 
+    /**
+     * Isolated message-bus to insulate cargo-project-workspace related messaging
+     * infra from general-purpose listeners, and properly manage listeners' lifetimes
+     */
+    private var messageBus: MessageBus? = null
```
matklad: lateinit?

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
matklad: `val lock` instead of `var lock`?

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: Holy grail, they finally introduced it.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
alexeykudinkin: Fixed.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -126,6 +135,20 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
                 .onWorkspaceUpdateCompleted(r)
     }
 
+    /**
+     * Subscribes given listener to the supplied topic on the [CargoProjectWorkspace]'s private
+     * message-bus
+     */
+    override fun <L> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable?) {
+        val conn = messageBus!!
+            .connect().let {
```
matklad: You probably want to use `.apply` here.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
matklad: The commend is still lying. `Alarm()` creates a `EDT` alarm, doesn't it?

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
matklad: Well, it was around at least since I've started using Kotlin =)

matklad: Now there is unnecessary `messageBus!!`

src/main/kotlin/org/rust/ide/actions/RefreshCargoProjectAction.kt
```diff
@@ -35,8 +38,33 @@ class RefreshCargoProjectAction : AnAction() {
...
+                workspace.subscribeTo(CargoProjectWorkspaceListener.Topics.UPDATES, it, it.connectionDisposer)
+                workspace.requestUpdateUsing(toolchain, immediately = true)
+            }
+        }
+    }
+
+    /**
+     * Project-update task's notifier
+     */
+    inner class Notifier(val module: Module) : CargoProjectWorkspaceListener {
```
matklad: Should be private and non inner.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -61,19 +62,27 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
...
+     * infra from general-purpose listeners, and properly manage listeners' lifetimes
+     */
+    private var messageBus: MessageBus? = null
+
     /** Component hooks */
 
     override fun getComponentName(): String = "org.rust.cargo.CargoProjectWorkspace"
 
     override fun initComponent() {
+        messageBus = MessageBusFactory.newMessageBus(CargoProjectWorkspace::class.java.name, module.messageBus)
```
matklad: This message bus provides the same messages as `module.messageBus` (you still publish updates to the `module.messageBus` and not to this private one). is this intentional?

alexeykudinkin: It's not, of course.

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -3,127 +3,166 @@ package org.rust.cargo.project.workspace.impl
...
+    /**
+     * Subscribes given listener to the supplied topic on the [CargoProjectWorkspace]'s private
+     * message-bus
+     */
+    override fun <L> subscribeTo(t: Topic<L>, listener: L, disposer: Disposable?) {
+        val conn =
+            messageBus
+                .connect()
+                .apply {
+                    this.subscribe(t, listener!!)
```
matklad: This `!!` can be avoided: just add `<L: Any>` bound to the `subscribeTo` function (the default is `L: Any?`). Also, I'd rather ban `!!` in non-test code :)

matklad:
I am still not satisfied with the listener :)

First of all, it is definitely not a private API any more: `RefreshCargoProjectAction` which is from a completely different package uses this API.

Second, it may invoke listeners in an inconsistent state, when `projcetDescription` is updated but libraries state is not. This is exactly the problem we are trying to prevent by using a lock. 

I would propose the following fix:

* remove `ProjectModelUpdater` and instead invoke `updateLibraries` function just before you `.syncPublisher`

After this change the following falls out naturally,

* `CargoProjectWorkspace` becomes a completely solid public API without hidden catches,
* `private lateinit var messageBus` becomes unnecessary because you can just pub/sub to the `module.messageBus`, because it is a completely public API. The complexity of implementation is reduced.
* `subscribeTo` function becomes then unnecessary, and this *greatly* simplifies the interface, because if you look at the signature of `subscribeTo` now, it is impossible to realize *on what message bus* does it "subscribe given listener to the supplied topic". (and `subscribeTo` is a public API, even if it is not indented to be. It is used by `org.rust.ide.actions.RefreshCargoProjectAction`).

Looks like complexity is magically removed and the amount of code spent is reduced. 

If you don't agree with this analysis it's fine. Let's just use the current implementation then. 

RTM with https://github.com/intellij-rust/intellij-rust/pull/348#discussion_r60833526 addressed.

src/main/kotlin/org/rust/ide/actions/RefreshCargoProjectAction.kt
```diff
@@ -34,8 +38,33 @@ class RefreshCargoProjectAction : AnAction() {
...
+            }
+
+            Disposer.dispose(connectionDisposer)
+        }
+
+        private fun showBalloon(message: String, type: MessageType) {
+            PopupUtil.showBalloonForActiveComponent(message, type)
         }
     }
 }
```
matklad: Maybe the logic here could be extracted into a single generic function

```
fun subscribeForOneMessage<L>(bus: MessageBus, topic: Topic<L>, listener: L)
```

alexeykudinkin:
> First of all, it is definitely not a private API any more: RefreshCargoProjectAction which is from a completely different package uses this API.

It isn't private API, it's a _low-level_ API.

`RefreshCargoProjectAction` is an action tightly bound to the `CargoProjectWorkspace`, and it may not be considered high-level one.

> Second, it may invoke listeners in an inconsistent state, when projcetDescription is updated but libraries state is not. This is exactly the problem we are trying to prevent by using a lock.

Once again, it's a _low-level_ API and it assumes that listener may not be assured that they will see it in consistent state. That's by design: message-bus is used to notify that _only_ workspace-related project-layout has been read successfully, nothing more.

Though, convinced. `ProjectModelUpdater` feels unnatural. Removed.

We still need to have private message-bus to properly manage respective connections' lifetimes, since plugin's lifetime often may not be the same as the module one (consider user updating or removing the plugin)

src/main/kotlin/org/rust/cargo/project/workspace/CargoProjectWorkspace.kt
```diff
@@ -1,23 +1,37 @@
...
      */
-    fun scheduleUpdate(toolchain: RustToolchain)
+    fun requestUpdateUsing(toolchain: RustToolchain, immediately: Boolean = false)
 
     /**
-     * Immediately schedules an update. Shows balloon upon completion.
+     * Latest version of the Cargo's project-description obtained
      *
-     * Update is still asynchronous.
+     * NOTA BENE: In the current implementation it's SYNCHRONOUS
```
matklad: Looks like the "SYNCHRONOUS" part is now misleading. We do not do any drastic blocking in this call any more (non EDT thread can block on the `lock` for a short period of time, ofc)

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -138,54 +190,61 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
...
-                }
-                is Result.Ok  ->
-                    ApplicationManager.getApplication().runWriteAction {
-                        if (!module.isDisposed) {
-                            val libraryRoots =
-                                result.cargoProject.packages
-                                    .filter { !it.isModule }
-                                    .mapNotNull { it.virtualFile }
+            lock (lock) {
+                updateModuleDependencies(r);
```
matklad: Is this intentionally not called in `setState`? If it is not the case then it makes sense to extract `lock, updateLibs, notify listeners` piece from `onSuccess` and `setState`.

matklad:
>We still need to have private message-bus to properly manage respective connections' lifetimes, since plugin's lifetime often may not be the same as the module one

But `CargoProjectWorkspace` is a `moduleService`, hence it's lifetime should be more or less the same as that of the module. Hence, the lifetime of this private messageBus is the same as that of `module.messageBus`. Is this correct? Then why can't we just use the `module.messageBus`? 

>consider user updating or removing the plugin

But the whole IDE needs to be restarted when plugins are changed! So how can this affect us?

matklad:
@alexeykudinkin do you want to clean up the history here? If not, let's merge this as is and fix remaining issues (if any) separately.

alexeykudinkin:
Sigh, thought that you're bound to restart the whole IDEA only in the case of plugin being _installed_ not being _uninstalled_.

Yeah, in that case it doesn't make much sense. Will collapse it.

matklad:
>Made module-component instead of a service (as soon as it gets a state, it should not be service anymore).

Can you please clarify the difference between a service and a component? I'm not sure I fully understand the distinction. Looks like

* A component is created eagerly when the module is created.
* A service is created lazily, when it is accessed for the first time. 
* Components have lifecycle hooks, services do not. 
* Both services and components can have [persistent state](http://www.jetbrains.org/intellij/sdk/docs/basics/persisting_state_of_components.html)
* Only once instance of service/component is loaded, so both can have non-persistent state. 

Is this correct?

matklad:
>I have convinced myself that the lock here does not deadlock

As always happens with concurrency issues, I was wrong :) This particular code does not deadlock, however a very simple change in an unrelated part of the project causes a deadlock. 

Here is it: https://github.com/intellij-rust/intellij-rust/commit/8f27650ea09ff66f43bedc206d29b4c6392ada9a

I've tried to add a notification for missing `stdlib`, and got this instead:

deadlock

I don't know what is the precise sequence of events that causes this, but here is my hypothesis: we update libraries roots while holding a lock. This update dispatches `LIBRARY_ROOTS_CHANGED` message in the message bus and executes listeners. Looks like one of those listeners drains EDT event queue. And one of those events waits (on the EDT thread) for all notifications to update. And notifications update in background thread. ;( .


I think that the core of a problem lies in [this critical section](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt#L173-L181). Critical sections should be as small as possible, and they certainly need to be bounded. However, we update library roots here, which pushes a message to the messageBus. And this means that the unknown and unbounded amount of code can run while we hold a lock.

TLDR: I think we should not update libraries while we hold the lock :)

PR #350, opened 2016-04-23
(PSI): introduce trait_body

jajakobyly:


matklad:
Hm, looks like in the `go` plugin they do not introduce a special syntactic category for bodies of things. https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/58a85af0be6652f1882ba5725588d691af7cd27d/grammars/go.bnf#L168

I guess this will not work for us, because we have rather more involved headers.

jajakobyly:
I find multiple advantages of this solution:
1) I believe this will make creating indentation fixer easier
2) Expand selection <3
3) Kotlin & Java do this

Disadvantage: `**Body.foo` instead of just `foo`

PR #352, opened 2016-04-23
(GRAM, PSI): introduce fn_type_params & type_list

jajakobyly:
I thought about making these guys just an instances of `RustParameters`, but they do not fully match it both syntactically and semantically.


src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -662,8 +663,7 @@ type_qual_path ::= << qualified_path view_path_prefix >>
 
 private types ::= type !'=' (',' type !'=')*
 
-private fn_anon_params ::= '(' anon_parameter anon_params_allow_variadic_tail ')'
-                         | '(' ')'
+bare_fn_params ::= '(' [ anon_parameter anon_params_allow_variadic_tail ] ')' { pin = 1 }
```
matklad: I would suggest renaming this to `fn_type_params`. 

`bare_fn` is from the days of youre, when closure_type and proc_type were around: https://github.com/rust-lang/rust/commit/37afc528eacf4a0a874f8e97f6e7d4d7b63e8ac6

matklad: Oh no, there are `type_params` already =(

jajakobyly: So what do you suggest? I have no strong feelings about what's (`bare_fn_params` or `fn_type_params`) more appropriate

matklad:
I think that `fn_type_parameters` is OK. However I think this whole parameter mess can be made a lot more readable. What do you think about the following organization:

```
// TODO: Maybe enforce @NotNull's in specific cases
fake parameter  ::= pat? type?

fn_parameter ::= pat type_ascription
lambda_parameter ::= pat type_ascription?
anon_parameter ::= [ restricted_pattern ':' ] type

fake parameters ::= self_argument? parameter*

fn_parameters     ::= '(' [ <<comma_separated_list fn_parameter>> ] ')' { pin = 1 }
lambda_parameters ::= '|' [ <<comma_separated_list lambda_parameter>> ] '|'
method_parameters ::= '(' ( self_argument [ ',' <<comma_separated_list fn_parameter>> ]
                          | [ <<comma_separated_list fn_parameter>> ] ) ')' { pin = 1 }
method_decl_parameters ::= '(' ( self_argument [ ',' <<comma_separated_list anon_parameter>> ]
                               | <<comma_separated_list anon_parameter>> ) ')' { pin = 1 }

private meta variadic_parameters ::= '(' [ <<param>> (',' <<param>>)*  [',' | '...' ] ] ')' { pin = 1 }

foreign_fn_parameters ::= <<variadic_parameters fn_parameter>>
fn_type_parameters ::= <<variadic_parameters anon_parameter>>

private restricted_pattern ::= &( IDENTIFIER | '_' | '&' (IDENTIFIER | '_') | '&&' (IDENTIFIER | '_') | MUT IDENTIFIER ) pat
self_argument ::= [ '&' LIFETIME? ] MUT? SELF type_ascription?
```

does it seem better only to me, or is it genuinely more readable? :)

jajakobyly:
Much easier to navigate, should I implement this or do you want to do this on your own? :)

matklad:
It would be nice if you implemet something similar, this is a rough sketch only. I'll be busy later today unfortunatelly.

jajakobyly:
Done

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -365,18 +366,45 @@ private block_item_group ::= &(CONST? UNSAFE? extern_abi? FN) fn_item
...
+fn_parameter ::= pat type_ascription
+lambda_parameter ::= pat type_ascription?
+anon_parameter ::= [ restricted_pattern ':' ] type
+path_parameter ::= type !'='
+
+fake parameters          ::= self_argument? parameter*
+fake variadic_parameters ::= self_argument? parameter* '...'? { extends = parameters }
+
+fn_parameters          ::= '(' !',' [ <<comma_separated_list fn_parameter>> ] ')' { pin = 1 }
+lambda_parameters      ::= '|' !',' [ <<comma_separated_list lambda_parameter>> ] '|'
```
matklad: I believe there is no need for this `!','` any more. They were introduced in https://github.com/intellij-rust/intellij-rust/commit/7513af403162f656dee3fa7ea5174986a4e0733a and it was not a great idea I think :)

jajakobyly: They provide better error messages imho and break less partial tests

PR #356, opened 2016-04-26
(GRAM, PSI, FMT): enum improvements:

jajakobyly:
- introduce enum_body
 - introduce enum_discriminant
 - make enum_args private
 - better formatting


jajakobyly:
I'm sure not: formatter consumes a set of rules and computes minimum amount of white space to be modified in order to comply with these rules.

Issue #354, opened 2016-04-25
Attach doc comments to all items properly


matklad:
Follow up of #133 

Currently, we do not attach doc comments to the first item of the file and to the items declared inside functions.

matklad:
Docs are attached properly to the first item. 

One way to fix docs for inner items is to revise our grammar:

Now:
```
expr_stmt   ::= outer_attr? ( block_expr !('.' | '?') ';'? | any_expr ';' )
decl_stmt   ::= docs_and_item | outer_attr? let_decl
empty_stmt  ::= ';'

private stmt ::= expr_stmt
                | decl_stmt
                | empty_stmt
```

Better?
```
expr_stmt   ::= outer_attr? ( block_expr !('.' | '?') ';'? | any_expr ';' )
let_stmt   ::= outer_attr? LET ...
empty_stmt  ::= ';'

private stmt ::= expr_stmt
                | let_stmt
                | empty_stmt

private stmt_or_item ::= stmt | docs_and_item
```

PR #358, opened 2016-04-26
(GRAM, PSI, FMT): introduce foreign_mod_body and mod_body

jajakobyly:


src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -526,11 +526,17 @@ extern_crate_item ::= EXTERN CRATE IDENTIFIER alias? ';'
 
 mod_decl_item ::= MOD IDENTIFIER ';'
 
-mod_item ::= MOD IDENTIFIER '{' inner_attr* docs_and_item* '}' {
-  pin(".*") = 3
+mod_item ::= MOD IDENTIFIER mod_body {
+  methods = [
+    // FIXME(jajakobyly): This method is tagged as @Nullable while we want @NotNull
+    itemList = "/mod_body/item_"
+  ]
```
jajakobyly: The overall problem here is, that `file_mod_item` extends `mod_item` but the former doesn't have anything like a `mod_body`. I decided to unify their apis (so users do not have to differentiate between mod and file mod) by creating artificial `getItemList` method on `mod_item`.

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustModItemImplMixin.kt
```diff
@@ -18,7 +18,7 @@ abstract class RustModItemImplMixin : RustItemImpl
     constructor(stub: RustItemStub, nodeType: IStubElementType<*, *>) : super(stub, nodeType)
 
     override val declarations: Collection<RustDeclaringElement>
-        get() = itemList
+        get() = itemList.orEmpty()
```
jajakobyly: We do not want this

src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt
```diff
@@ -291,7 +291,7 @@ private class Resolver {
         override var matched: RustNamedElement? = null
 
         override fun visitModItem(mod: RustModItem) {
-            seek(mod.itemList)
+            seek(mod.itemList.orEmpty())
```
jajakobyly: Again

matklad:
I am not happy about this `fileModItem` situation at all... It also causes https://github.com/intellij-rust/intellij-rust/issues/354.

Maybe we can make `RustFile` and `RustModItem` to extend some kind of `RustMod`?

jajakobyly:
Seems reasonable

matklad:
>Implement this here?

Definitely not here, I predict a massive fallout after this change, especially in the resolve area. It would be cool if you implement this, but I am not sure about the design, so expect quite a bit of back and forth. If you don't want to mess with it, I think I can fix this tomorrow. 

I'll be glad to accept `foreign_mod` part right away!

jajakobyly:
I think you can take on this refactoring. Also the `mod_item` part works, this method will never return null; it's just GK which puts `@Nullable` there. I could replace `orEmpty()` with `!!` and there would be no harm.

matklad:
Ok, I really hope that I have not underestimate the amount of work the refactoring :)

alexeykudinkin:
Are there any reasons for the new PSI element (`mod_body`) introduction?

matklad:
@alexeykudinkin yep: https://github.com/intellij-rust/intellij-rust/pull/350#issuecomment-213802679.

alexeykudinkin:
@matklad no, it's bad idea to fix indenter and expand-selection by introduction of the new PSI node. 

Can hardly spot any other reason for it to be introduced.

matklad:
@alexeykudinkin and what are the benefits of not introducing the PSI node? I don't feel that either way is significantly better, but simpler formatter is a nice to have.

alexeykudinkin:
> @alexeykudinkin and what are the benefits of not introducing the PSI node?

Mitigating introduction of the useless abstraction is the greatest benefit. The problem is with formatter & expand-selection not with the node.

matklad:
Sounds reasonable, @jajakobyly what do you think?

matklad:
I've looked at Java and Kotlin implementations, and the interesting patterns is that `{ stuff }` has its own PsiNode iff it is optional. 

That is, in Java `PsiClass` always has `{ }`, so there is no `PsiClass`. In Kotlin `{}` is optional, so there are `KtClass` and `KtClassBody`. 

I think this is a reasonable approach, based on this, in Rust there should be separate body elements for structs, enums and methods. Mods should not have a separate `ModBody` because module without a body is a `mod_decl`.

jajakobyly:
Your arguments sound reasonable for me too, though I prefer introducing separate PSI nodes for such things as `mod_body` as I perceive them as hmm *distinguishable* entities.

Formatter can be easily adopted to the lack of `mod_body`, expand selection unfortunately is another story, as I cannot see any abilities to configure it ([ExpandSelectionAction](https://upsource.jetbrains.com/idea-ce/file/idea-ce-52c7416b07ef88943ab4cccebc34b4c3a304614a/plugins/ui-designer/src/com/intellij/uiDesigner/actions/ExpandSelectionAction.java?nav=0:0:focused)) (and it would be actually nice to have an ability to expand selection to the `{ ... }` range, because of already mentioned `mod_decl`. Though I understand that this is very minor feature.

So should I narrow this PR to `foreign_mod_body`, or close it completely?

matklad:
>So should I narrow this PR to foreign_mod_body, or close it completely?

`foreign_mod_body` is also mandatory, so let's close this.

PR #359, opened 2016-04-27
Rust mod

matklad:
This PR refactors the way we handle top-level module in file. Why do we need this?

Well, currently we have `mod_item` and `file_mod_item { extends = mod_item }`. This has several drawbacks: 

* this is one of the [fragile points](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf#L300-L301) of our parser.
* we have a semi invariant that Rust file has exactly one child node, and it's type is `RustFileModItem`. This is *semi* because it does not work for comments and docs (#354). It's interesting that if I try to fix this particular issue with the help of `TokenBinders`, I get `NoClassDefFound` error during `generate` phase (this reinforces the "fragile point" point)
* the invariant looks kind of strange: if file node has exactly one child, can we just remove it and attach all elements to the file instead?

Moreover the inheritance itself seems backwards. `mod_item` has more stuff in it then `file_mod_item`, so looks like `mod_item` should extend `file_mod_item`. Alas, `mod_item` already extends `item`, `file_mod_item` is definitely not an item, and we can't extend two parents simultaneously. 

Ok, if `A` can't extend `B` and `B` can't extend `C` then why can't we simply introduce an interface `I` and implement it for `A` and `B`? Well, this is exactly what happens here :) 

Here is [the interface](https://github.com/intellij-rust/intellij-rust/blob/rust-mod/src/main/kotlin/org/rust/lang/core/psi/RustMod.kt). It is implemented for `RustFile` and for `RustModItem`. That is, `RustFileModItem` is gone and `RustFile` is used directly (PSI becomes one level more shallow (this is the reason why **every** parsing test result is changed :( ) ) 

I am pretty sure that this implementation is much better then the previous one, but I am not absolutely happy with it either. Particularly, I don't like the amount of changes here, ~~and the introduction of `visitRustFile` visitor method~~(removed). 

@jajakobyly @alexeykudinkin what do you think about this?


src/main/kotlin/org/rust/lang/core/psi/visitors/RustVisitorEx.kt
```diff
@@ -1,10 +1,19 @@
...
     open fun visitLiteral(literal: RustLiteral) = visitElement(literal)
     open fun visitNumericLiteral(literal: RustLiteral.Number) = visitLiteral(literal)
     open fun visitTextLiteral(literal: RustLiteral.Text) = visitLiteral(literal)
+    open fun visitRustFile(file: RustFile) {
+        super.visitFile(file)
+    }
+
+    override fun visitFile(file: PsiFile) {
+        if (file is RustFile) visitRustFile(file) else super.visitFile(file)
+    }
```
jajakobyly: `visitRustFile` looks to be just a convenience method. It could be easily removed and that `if (file is RustFile) ...` pattern could be used instead. I don't see any disadvantages of this other than you have to be careful when override `visitFile`, which probably will never happen having in mind that this is `RustVisitorEx`

src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt
```diff
@@ -5,20 +5,54 @@ import com.intellij.openapi.fileTypes.FileType
...
+    override fun getReference(): RustReference? = null
+
+    override val nameElement: PsiElement? = null
 
     override fun getFileType(): FileType = RustFileType
 
-    val mod: RustFileModItem?
-        get() = findChildByClass(RustFileModItem::class.java)
+    override val items: List<RustItem>
+        get() = PsiTreeUtil.getChildrenOfTypeAsList(this, RustItem::class.java)
```
jajakobyly: This can be extracted to a separate extension method which uses reified parameters, like `parentOfType`

jajakobyly:
:+1: the amount of changes in such refactoring is inevitable imho

src/main/kotlin/org/rust/lang/core/psi/visitors/RustVisitorEx.kt
matklad: >it could be easily removed and that if (file is RustFile) ... pattern could be used instead. 

I don't think that requiring `as` casts is a good API. It is much better to encapsulate them behind the functions (if it is not possible to eliminate them all together).

But we already have such encapsulating function (`rustMod`) and this visitor extension is indeed unnecessary, thanks for pointing this out! Will remove it.

src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt
matklad: Yep, will do it separately.

alexeykudinkin:
I believe, we can now update this thing: 
https://github.com/intellij-rust/intellij-rust/blob/master/src/test/resources/org/rust/lang/core/parser/fixtures/well-formed/doc_comments.rs

matklad:
@alexeykudinkin yep, will do after the merge. I've tried to minimize the diff here (not very successfully)

src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt
```diff
@@ -5,20 +5,54 @@ import com.intellij.openapi.fileTypes.FileType
...
+import org.rust.lang.core.psi.util.module
 import org.rust.lang.core.resolve.indexes.RustModulePath
+import org.rust.lang.core.resolve.indexes.RustModulesIndex
+import org.rust.lang.core.resolve.ref.RustReference
 
-class RustFile(fileViewProvider: FileViewProvider) : PsiFileBase(fileViewProvider, RustLanguage) {
+class RustFile(fileViewProvider: FileViewProvider) : PsiFileBase(fileViewProvider, RustLanguage), RustMod {
+    override fun getReference(): RustReference? = null
+
+    override val nameElement: PsiElement? = null
```
alexeykudinkin: I don't like this stub -- module is actually named, though, since every use of the `nameElement` 'conventionally' followed by the `.text` invocation, we can't return file-instance itself here.

alexeykudinkin: I believe its the right time to fix all usages of the `.text`. Let's return the file-instance itself here. Or `PsiDirectory` in case of `mod.rs`.

matklad: Yep, I agree that it should be fixed, but I'd prefer to do this in a separate PR. It is an orthogonal issue (I believe returning `null` is what `FileModItem` does currently). I am opening the issue for this.

alexeykudinkin:
Then, merge when ready. And please fix `nameElement` for consistency.

src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt
matklad: Issue:  #360 

On the other hand, the top level module does not own any `identifier`. `PsiFile` and `PsiDirectory` are not exactly identifiers. Maybe it is possible to remove `RustNamedElement` bound from the `RustMod` with some work? Will investigate this.

alexeykudinkin: No, do not remove `RustNamedElement`. The fact `PsiFile` & `PsiDirectory` do not have identifiers doesn't mean they could not be named-elements (actually they are).

matklad: >No, do not remove RustNamedElement`

Agree. 

There is a difference between `NamedElement` and `NameIdentifierOwner`. Files and Directories are certainly named, but they don't have an identifier element.

I think it then should be possible to remove `val nameElement` from `RustNamedElement` and create a `RustIdentifierOwner` which extends `RustNamedElement` and aslo has `nameElement`

matklad: @alexeykudinkin ouch, there is yet one more nasty complication here. The name of the mod is actually not related to the name of the file. Consider, i.e, this

```
#[path="./bar.rs"]
mod foo;
```

Issue #360, opened 2016-04-28
Refactor `RustNamedElement`


matklad:
All usages of `nameElement` are followed by `.text`, which does not really work for top level modules. Besides, there is `com.intellij.psi.PsiNameIdentifierOwner`, which can be reused instead of providing our own `val nameElement` (the benefits are unclear however).

jajakobyly:
Whoa didn't know that PsiNameIdentifierOwner isn't used :O

alexeykudinkin:
This refactoring should be straightforward and assume no more than just a replacement of the `.text` with  the `.name` references of the respective named-elements.

matklad:
@alexeykudinkin I suspect we are not on the same page here, I must have caused a massive confusion by #361 :(

So, we have `RustNamedElement`:

```
interface RustNamedElement   : RustCompositeElement
                             , PsiNamedElement
                             , NavigatablePsiElement {
    val nameElement: PsiElement?
}
```

What is the semantics of this `nameElement`? To me, it looks like this should be the identifier (a single word), so that `RustNameElement.name` can be implemented as `this.nameElement.text`. I don't understand how it is possible to return `this` for `nameElement` and then use `.name` (I understand that it was your suggestion [here](https://github.com/intellij-rust/intellij-rust/pull/359#discussion_r61402455)). Using `.name` will require this recursion

```
interface RustNamedElement   : RustCompositeElement
                             , PsiNamedElement
                             , NavigatablePsiElement {
    val nameElement: RustNamedElement?
}
```

To me, looks like this `nameElement` is an implementation detail (and `name` is the interface). I've tried to implement it in this PR: https://github.com/intellij-rust/intellij-rust/pull/364

alexeykudinkin:
Fix also `nameElement` for the `RustFile`, please!

matklad:
Closed by https://github.com/intellij-rust/intellij-rust/pull/364

PR #361, opened 2016-04-28
(RES): relax type bounds in resolve

matklad:
Resolve to `RustCompositeElement` instead of `RustNamedElement`.

Resolving to `NamedElement` is both useless and wrong.

It is useless because an element can be renamed several times via `as` alias, so the name of the resolved element has no connection with the name of the referring element.

It is wrong because in fact it is possible to resolve to an anonymous element!

Example 1, tuple struct fields:

```Rust
#[derive(Default)]
struct Pair(String, i32);

fn main() {
    let p: Pair = Default::default();
    p.<caret>1; // this should resolve to the anonymous field of the Pair.
}
```

Example 2, the root module is always anonymous so `::` resolves to a nameless item. 

@alexeykudinkin what do you think about this? I think this is the right step in the direction of fixing #360.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
@@ -64,7 +64,7 @@ private class CompletionScopeVisitor(private val context: RustQualifiedReference
     }
 }
 
-private fun RustNamedElement?.completionsFromResolveScope(): Collection<RustNamedElement> =
+private fun RustCompositeElement?.completionsFromResolveScope(): Collection<RustNamedElement> =
     when (this) {
         is RustResolveScope -> boundElements
         else                -> emptyList()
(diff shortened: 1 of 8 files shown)


jajakobyly:
Hmm, why not to just resolve to `PsiElement`? Are there any strong reasons to use `RustCompositeElement`?

matklad:
>Hmm, why not to just resolve to PsiElement? Are there any strong reasons to use RustCompositeElement?

I don't see solid reasons to use `RustCompositeElement`, but I don't see them for `PsiElement` either. 

On the philosophical point of view, if we can guarantee that we resolve to `RustCompositeElement`, then it makes sense to express this in the types.

jajakobyly:
I think about possibilities to resolve to non-Rust symbols in the (distant) future.

alexeykudinkin:
No, they're not.

`RustNamedElement` in the signature of the `resolve` method mean that the target element being resolved to is a _named_ entity and doesn't bound it to any particular name out there.

Your example, with the pair isn't valid because its a _structural_ matching (which'd be accompanied by the specific reference). Example with the root module is valid even less, since it can be used only for addressing items inside of the root module.

PR #362, opened 2016-04-28
(FMT): Improve code formatting: Line Breaks

jajakobyly:
This is a continuation of #302, which concentrates on formatting line breaks between code blocks.

Creative test cases are very, very welcome.

Most inspiration was taken from [Python formatter](https://upsource.jetbrains.com/idea-ce/file/idea-ce-d93ab79ac5306c9a9880e16438147eafe137f039/python/src/com/jetbrains/python/formatter/PythonFormattingModelBuilder.java), [Go formatter](https://upsource.jetbrains.com/go-lang-idea-plugin/file/ede4f339b4968989ea1a5bfcfefa0894874fef81/src/com/goide/formatter/GoFormattingModelBuilder.java) and [Kotlin formatter](https://github.com/JetBrains/kotlin/tree/master/idea/formatter/src/org/jetbrains/kotlin/idea/formatter) [(2)](https://github.com/JetBrains/kotlin/tree/master/idea/src/org/jetbrains/kotlin/idea/formatter).

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/RustLanguageCodeStyleSettingsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/RustLanguageCodeStyleSettingsProvider.kt
@@ -27,7 +27,10 @@ class RustLanguageCodeStyleSettingsProvider : LanguageCodeStyleSettingsProvider(
     override fun customizeSettings(consumer: CodeStyleSettingsCustomizable, settingsType: SettingsType) {
         when (settingsType) {
             BLANK_LINES_SETTINGS         -> {
-                // TODO Implement this.
+                consumer.showStandardOptions(
+                    "KEEP_LINE_BREAKS",
+                    "KEEP_BLANK_LINES_IN_DECLARATIONS",
+                    "KEEP_BLANK_LINES_IN_CODE")
             }
 
             SPACING_SETTINGS             -> {
--- a/src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
@@ -1,20 +1,65 @@
 package org.rust.ide.formatter
 
+import com.intellij.formatting.ASTBlock
+import com.intellij.formatting.Block
+import com.intellij.formatting.Spacing
+import com.intellij.formatting.Spacing.createSpacing
 import com.intellij.formatting.SpacingBuilder
+import com.intellij.lang.ASTNode
+import com.intellij.psi.PsiComment
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.TokenType.WHITE_SPACE
 import com.intellij.psi.codeStyle.CommonCodeStyleSettings
+import com.intellij.psi.formatter.FormatterUtil
+import com.intellij.psi.impl.source.tree.TreeUtil
 import com.intellij.psi.tree.IElementType
 import com.intellij.psi.tree.TokenSet
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RustCompositeElementTypes.*
-import org.rust.lang.core.psi.RustKeywordTokenType
 import org.rust.lang.core.psi.RustTokenElementTypes.*
+import org.rust.lang.core.psi.impl.RustFile
+import org.rust.lang.core.psi.util.containsEOL
+import org.rust.lang.core.psi.util.getNextNonCommentSibling
+import org.rust.lang.core.psi.util.getPrevNonCommentSibling
 import com.intellij.psi.tree.TokenSet.create as ts
+
+private val KEYWORDS = ts(*IElementType.enumerate { it is RustKeywordTokenType })
+private val NO_SPACE_AROUND_OPS = ts(COLONCOLON, DOT, DOTDOT)
+private val SPACE_AROUND_OPS = ts(AND, ANDAND, ANDEQ, ARROW, FAT_ARROW, DIV, DIVEQ, EQ, EQEQ,
+    EXCLEQ, GT, LT, MINUSEQ, MUL, MULEQ, OR, OREQ, OROR, PLUSEQ, REM, REMEQ, XOR, XOREQ, MINUS, PLUS,
+    GTGTEQ, GTGT, GTEQ, LTLTEQ, LTLT, LTEQ)
+private val UNARY_OPS = ts(MINUS, MUL, EXCL, AND, ANDAND)
+// PATH_PART because `Fn(A) -> R`
+private val PAREN_LIST_HOLDERS = ts(PAREN_EXPR, TUPLE_EXPR, TUPLE_TYPE, PARAMETERS, VARIADIC_PARAMETERS, ARG_LIST,
+    IMPL_METHOD_MEMBER, BARE_FN_TYPE, PATH_PART, PAT_ENUM, PAT_TUP, ENUM_TUPLE_ARGS)
+private val BRACK_LIST_HOLDERS = ts(VEC_TYPE, ARRAY_EXPR, INDEX_EXPR)
+private val BRACE_LIST_HOLDERS = ts(USE_GLOB_LIST)
+private val ANGLE_LIST_HOLDERS = ts(GENERIC_PARAMS, GENERIC_ARGS, QUAL_PATH_EXPR)
+private val ATTRS = ts(OUTER_ATTR, INNER_ATTR)
+private val BLOCK_LIKE = ts(BLOCK, STRUCT_DECL_ARGS, STRUCT_EXPR_BODY, IMPL_BODY, MATCH_BODY, TRAIT_BODY, ENUM_BODY,
+    ENUM_STRUCT_ARGS)
+private val TYPES = ts(VEC_TYPE, PTR_TYPE, REF_TYPE, BARE_FN_TYPE, TUPLE_TYPE, PATH_TYPE,
+    TYPE_WITH_BOUNDS_TYPE, FOR_IN_TYPE, WILDCARD_TYPE)
+private val MACRO_ARGS = ts(MACRO_ARG, FORMAT_MACRO_ARGS, TRY_MACRO_ARGS)
+private val PARAMS_LIKE = ts(PARAMETERS, VARIADIC_PARAMETERS)
 
 fun createSpacingBuilder(commonSettings: CommonCodeStyleSettings,
                          rustSettings: RustCodeStyleSettings): SpacingBuilder =
     SpacingBuilder(commonSettings)
-        // rules defined earlier have higher priority
+        // Rules defined earlier have higher priority.
+        // Beware of comments between blocks!
+
         //== some special operators
-        .applyForEach(SPACE_AFTER_OPS) { after(it).spaceIf(true).before(it).spaceIf(false) }
+        // FIXME(jajakobyly): Doesn't work well with comments
+        .afterInside(COMMA, ts(STRUCT_DECL_ARGS, ENUM_BODY)).spacing(1, 1, 1, true, 1)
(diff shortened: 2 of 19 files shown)


src/main/kotlin/org/rust/ide/formatter/RustBlock.kt
```diff
@@ -21,6 +26,23 @@ class RustBlock(
...
+            }
+
+            if (psi1.isTopLevelItem && psi2.isTopLevelItem) {
+                return RustSpacing.lineBreak(
+                    keepLineBreaks = ctx.commonSettings.KEEP_LINE_BREAKS,
+                    keepBlankLines = ctx.commonSettings.KEEP_BLANK_LINES_IN_DECLARATIONS)
+            }
+        }
         return ctx.spacingBuilder.getSpacing(this, child1, child2)
     }
```
jajakobyly: I'm currently writing this the "classic" way - how it's implemented in Python, Go, Java, etc etc formatters. Though Kotlin formatter does this differently: it [greatly extends SpacingBuilder](https://github.com/JetBrains/kotlin/blob/master/idea/formatter/src/org/jetbrains/kotlin/idea/formatter/KotlinSpacingBuilder.kt) to empower creating complex spacing rules with a declarative way.

src/main/kotlin/org/rust/ide/formatter/RustBlock.kt
```diff
@@ -21,6 +26,23 @@ class RustBlock(
...
+            val parent = psi1.parent
+
+            if (psi1 is RustOuterAttr && (psi2 is RustOuterAttr || parent is RustItem)) {
+                return RustSpacing.lineBreak()
+            }
+
+            if (psi1.isTopLevelItem && psi2.isTopLevelItem) {
+                return RustSpacing.lineBreak(
+                    keepLineBreaks = ctx.commonSettings.KEEP_LINE_BREAKS,
+                    keepBlankLines = ctx.commonSettings.KEEP_BLANK_LINES_IN_DECLARATIONS)
```
jajakobyly: This rule uses IDEA's default which states +/-: "keep blank lines, but maximally 2 consecutive"

src/main/kotlin/org/rust/ide/formatter/RustBlock.kt
matklad: Judging by the end result, I am not sure that the Kotlin way is better :)

src/test/resources/org/rust/ide/intentions/fixtures/add_derive/add_derive_enum_after.rs
```diff
@@ -1,3 +1,4 @@
-#[derive(<caret>)] enum Test {
+#[derive(<caret>)]
+enum Test {
```
matklad: :+1:

matklad:
Looks good, but the tests are failing (so we have good tests, yay!)

jajakobyly:
I prefer to open PRs early and receive code review as fast as possible ;) It's not everything I plan to create here

src/main/kotlin/org/rust/ide/formatter/RustBlock.kt
```diff
@@ -58,33 +56,31 @@ class RustBlock(
...
-        private val BLOCK_START_TOKEN_SET = TokenSet.create(
-            PUB,
-            MOD,
-            STRUCT,
-            ENUM,
-            IMPL,
-            TRAIT,
-            MATCH
-        )
+private val BLOCK_START_TOKEN_SET = TokenSet.create(
```
matklad: It feels a bit inconsistent that we use either `TokenSet.create` or `ts`

src/main/kotlin/org/rust/ide/formatter/RustLanguageCodeStyleSettingsProvider.kt
```diff
@@ -27,7 +27,10 @@ class RustLanguageCodeStyleSettingsProvider : LanguageCodeStyleSettingsProvider(
     override fun customizeSettings(consumer: CodeStyleSettingsCustomizable, settingsType: SettingsType) {
         when (settingsType) {
             BLANK_LINES_SETTINGS         -> {
-                // TODO Implement this.
+                consumer.showStandardOptions(
+                    "KEEP_LINE_BREAKS",
+                    "KEEP_BLANK_LINES_IN_DECLARATIONS",
+                    "KEEP_BLANK_LINES_IN_CODE")
```
matklad: Looks like there is really no way to avoid string literals here... Am I correct that it will not throw an error if you make a typo here?

src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
```diff
@@ -1,124 +1,163 @@
...
-        TYPE_WITH_BOUNDS_TYPE, FOR_IN_TYPE, WILDCARD_TYPE)
-    private val MACRO_ARGS = ts(MACRO_ARG, FORMAT_MACRO_ARGS, TRY_MACRO_ARGS)
-    private val PARAMS_LIKE = ts(PARAMETERS, VARIADIC_PARAMETERS)
-
-    inline fun SpacingBuilder.applyForEach(tokenSet: TokenSet,
-                                           block: SpacingBuilder.(IElementType) -> SpacingBuilder): SpacingBuilder {
-        var self = this
-        for (tt in tokenSet.types) {
-            self = block(this, tt)
+fun lineBreak(minLineFeeds: Int = 1,
```
matklad: can be private

src/main/kotlin/org/rust/ide/formatter/RustBlock.kt
jajakobyly: I'm not sure, but it's possible that I will delete these guys when I will be working on indentation anyway.

src/main/kotlin/org/rust/ide/formatter/RustLanguageCodeStyleSettingsProvider.kt
jajakobyly: Yep it just do not display given option in code style settings panel; so it's pretty easy to verify this anyway.

src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
```diff
@@ -1,124 +1,198 @@
...
+    get() = (this is RustItem || this is RustAttr) && this.parent is RustFile
+
+private val PsiElement.isStmtOrExpr: Boolean
+    get() = this is RustStmt || this is RustExpr
+
+private val ASTNode.hasLineBreakAfterInSameParent: Boolean
+    get() = treeNext != null && isWhiteSpaceWithLineBreak(TreeUtil.findFirstLeaf(treeNext))
+
+private val ASTNode.hasLineBreakBeforeInSameParent: Boolean
+    get() = treePrev != null && isWhiteSpaceWithLineBreak(TreeUtil.findLastLeaf(treePrev))
```
jajakobyly: I'm not sure whether `TreeUtil` is available in non-idea ides; though these methods are almost copy-pasted from Python's formatter; so it has to be available at least in PyCharm.

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiExtensions.kt
```diff
@@ -22,16 +22,17 @@ inline fun <reified T : PsiElement> PsiElement.parentOfType(strict: Boolean = tr
 inline fun <reified T : PsiElement> PsiElement.childOfType(strict: Boolean = true): T? =
     PsiTreeUtil.findChildOfType(this, T::class.java, strict)
 
-fun PsiElement?.getNextNonPhantomSibling(): PsiElement? =
```
jajakobyly: This method was dead code (and so the `PHANTOM_TOKEN_SET` used only in it), so I have allowed myself to just replace it with two sleeker alternatives.

src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
```diff
@@ -1,124 +1,198 @@
...
+            origPsi1.getPrevNonCommentSibling() ?: origPsi1
+        } else {
+            origPsi1
+        },
+        if (origPsi2 is PsiComment && origNode2.hasLineBreakBeforeInSameParent) {
+            origPsi2.getNextNonCommentSibling() ?: origPsi2
+        } else {
+            origPsi2
+        }
+    )
```
jajakobyly: Ugh comments between formatting blocks are very burdensome. I hope the heuristics I have used here will work (almost) flawlessly.

matklad: Yep, comments are the hardest problem of every formatter

src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
```diff
@@ -1,124 +1,225 @@
...
+        val parentPsi = origPsi1.parent
+        val parentParentPsi = parentPsi.parent
+
+        // Here are rules which are comment-sensitive
+        when {
+            origPsi1 is RustOuterAttr && origPsi2 is PsiComment
+            -> return one()
+
+            origPsi1 is RustOuterAttr && (origPsi2 is RustOuterAttr || origPsi1.parent is RustItem)
+                || origPsi1 is PsiComment && (origPsi2 is RustOuterAttr || origPsi1.getPrevNonCommentSibling() is RustOuterAttr)
```
jajakobyly: The hardest thing in programming is naming things. The second hardest it beautifully laying out complex logical expressions :|

matklad: The best way to format complex logical expressions which I am aware of is just not writing them. Extracting and naming a couple of sub expressions usually helps both formatting and clarity (although sometimes it interferes badly with order of evaluation). But I am not sure how to apply the technique in this particular case :)

Formatting [is hard](http://journal.stuffwithstuff.com/2015/09/08/the-hardest-program-ive-ever-written/)

jajakobyly:
Ok, should work +/-. It doesn't play well with comments sometimes, but this formatter doesn't have to be the perfect one (rustfmt should be) and I can improve it later.

src/main/kotlin/org/rust/ide/formatter/RustSpacing.kt
```diff
@@ -1,124 +1,293 @@
...
+
+        //== empty parens
+        .between(LPAREN, RPAREN).spaceIf(false)
+        .between(LBRACK, RBRACK).spaceIf(false)
+        .between(LBRACE, RBRACE).spaceIf(false)
+        .betweenInside(OR, OR, LAMBDA_EXPR).spaceIf(false)
+
+        //== paren delimited lists
+        // withinPairInside does not accept TokenSet as parent node set :(
+        // and we cannot create our own, because RuleCondition stuff is private
```
matklad: You can write an extension method which uses cycle to simulate `TokenSet`, but it may not be a good idea.

matklad:
Works good!

The code is hard to follow, but I don't think it can be made significantly simper. One trick to make is easier to read is to place Rust code examples in the comments for the trickiest parts.

matklad:
Here are some examples of bad formatting: https://github.com/matklad/rustraytracer/commit/530c17b71dfddaae6345fa969af400a95a3199bf

They are mostly about bad indents, but there is also a spacing bug for `&`:

https://github.com/matklad/rustraytracer/commit/530c17b71dfddaae6345fa969af400a95a3199bf#diff-c14d5a69a77d82f4c014d3fd2fcb3361L40
https://github.com/matklad/rustraytracer/commit/530c17b71dfddaae6345fa969af400a95a3199bf#diff-73dec787512a58517803c1819029c229L112

matklad:
Hm, can you reproduce https://youtrack.jetbrains.com/issue/KT-12239 ?

jajakobyly:
> Hm, can you reproduce https://youtrack.jetbrains.com/issue/KT-12239 ?

Nope, works fine on my machine (I assume that this relates to Gradle build, so my spec shouldn't matter).

PR #363, opened 2016-04-28
(WORKSPACE): synchronize via read/write actions

matklad:
Looks like the simplest way to provide the necessary synchronization is to leverage existing read/write actions infra. This should prevent effects like https://github.com/intellij-rust/intellij-rust/pull/348#issuecomment-215506753.

@alexeykudinkin please review

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
@@ -5,6 +5,7 @@ import com.intellij.execution.process.ProcessAdapter
 import com.intellij.execution.process.ProcessEvent
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.ModalityState
+import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.module.Module
 import com.intellij.openapi.module.ModuleComponent
@@ -27,9 +28,6 @@ import org.rust.cargo.toolchain.RustToolchain
 import org.rust.cargo.util.cargoLibraryName
 import org.rust.cargo.util.cargoProjectRoot
 import org.rust.cargo.util.updateLibrary
-import org.rust.lang.utils.lock
-import java.util.concurrent.locks.Lock
-import java.util.concurrent.locks.ReentrantLock
 
 
 /**
@@ -46,11 +44,6 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
      * It uses dispatch-thread.
      */
     private val alarm = Alarm()
-
-    /**
-     * Lock to guard reads/updates to [cached]
-     */
-    private val lock: Lock = ReentrantLock()
 
     /**
      * Cached instance of the latest [CargoProjectDescription] instance synced with `Cargo.toml`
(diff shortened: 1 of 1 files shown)


alexeykudinkin:
First of all, let's make a convention of not fixing a bug, until it's established to be a bug. 
If we're uncertain about the reasons of the concrete deadlock case, let's (at first) investigate it further to become certain whether the actual problem matches our assumption.

Will look closer on the problem tomorrow.

matklad:
Here is the jstack dump of the described deadlock: https://gist.github.com/matklad/0fb650d7fb70e9ea1ac5801d1b1d3cda

src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt
```diff
@@ -107,30 +100,42 @@ class CargoProjectWorkspaceImpl(private val module: Module) : CargoProjectWorksp
...
+    private fun commitUpdate(r: UpdateResult) {
         ApplicationManager.getApplication().assertIsDispatchThread()
+        if (module.isDisposed) return
+
+        runWriteAction {
+            if (r is UpdateResult.Ok) {
+                cached = r.projectDescription
+                updateModuleDependencies(r.projectDescription);
+            }
+            notifyCargoProjectUpdate(r)
```
alexeykudinkin: If notifying under _local_ lock isn't such a great problem, notifying holding a _global write lock_ is seeking for a trouble for sure: action inside listeners may accidentally trigger read-actions, and then we'd get a deadlock on a singe-thread.

alexeykudinkin: Correction: it's not the case since we can re-enter RL holding WL.

matklad: But it still may be a good idea to notify listeners after modification (we are EDT thread, so this won't decrease synchronization guarantees)

alexeykudinkin: Yep, this is a mandatory: we do only _critical_ actions under WL. 
Notification isn't of that breed for sure.

alexeykudinkin: BTW, a short post-mortem: initial decision of taking _global WL_ holding local lock was smelly as hell, and should not be introduced in the future -- it's a source of the _guaranteed_ dead-lock, since (most of) the actions are performing under RL.

alexeykudinkin: BTW, your listener should run out of dispatch-thread -- all UI related activity should go through EDT _only_.

matklad: Fixed. 

There was even one more problem with this: a listener won't get an assertion error if it executed accidentally executes modification operation.

matklad: >BTW, your listener should run out of dispatch-thread -- all UI related activity should go through EDT only.

Not sure I have understood you correctly here, can you please elaborate? 

Looks like root modification messages [are dispatched](https://upsource.jetbrains.com/idea-ce/file/idea-ce-c93c327673bdca7d0f0d022e82842485cd3bad59/platform/projectModel-impl/src/com/intellij/openapi/module/impl/ModuleManagerImpl.java?nav=31266:31350:focused&line=1028) in EDT ( an even in write action ??? )

In any case, we have only one listener whose job is to show a popup.

alexeykudinkin:
Under 'your action' i've meant an action that caused deadlock originally.

Issue #207, opened 2016-01-10
Resolve of local items


matklad:
Items can be declared inside blocks, so we should complete/resolve them.

These  are some contrived examples:

```Rust
fn main() {
    foo();

    // No lexical scope?
    fn foo() {}
}

fn main() {
    fn foo() -> i32 { 92 }

    {
        println!("{}", foo());  // prints 62
        // Or is there?
        fn foo() -> i32 { 62 }
    }

}

fn main() {
    let foo = || { 92 };

    {
        // Or maybe not?
        fn foo() -> i32 { 62 }
        println!("{}", foo());  // prints 92
    }

}
```

So, there are different scoping rules for locals and items, and a local always takes precedence over an item.

matklad:
More context: https://github.com/rust-lang/rust/issues/33458

PR #365, opened 2016-04-30
(GRAM): make item not a statement

### LINKED ISSUE #354: Attach doc comments to all items properly

matklad:
Currently, we have `decl_stmt ::= let_decl | item`. However In Rust, local item declarations do not behave like statements. Most importantly, the order of declarations is not important and the item can
be used before it is declared:

```Rust
fn main() {
    foo(); // We can call foo here

    fn foo() { }
}
```

Does it make sense not to treat local items like statements? @jajakobyly @alexeykudinkin what do you think? This will fix #354 and will make #207 a tiny bit more natural.


jajakobyly:
:+1:

src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
```diff
@@ -269,7 +269,6 @@
     extends     ("(if|while)_let_expr") = scoped_let_expr
     extends     (".*_expr") = restricted_expr
     extends     ("(?!ret_type).*_type") = type
-    extends     (".*_stmt") = stmt
```
jajakobyly: Maybe I'm a bit picky, but this rule seemed fine. Why not just add `|let_decl` to it (`extends     (".*_stmt|let_decl") = stmt`)?

matklad: The more I work with our parser, the less I like that `extends` declarations are detached from the actual productions. They are nice if you want to save typing `extends` for a dozen of productions (like expessions). But here we have only three cases, and it is clearer just to specify them inline.

Issue #311, opened 2016-03-30
Add context aware run configuration


matklad:
`Ctrl+Shift+f10` should create a specific run configuration depending on context. 

If it is invoked in a file with `fn main()`, a `cargo run` configuration should be created with the correct `--bin`.

If it is invoked inside a test module or a test function, a `cargo test` configuration for the specific tests should be created. 

To implement this, we need to split our single `CargoCommandConfiguration` into `CargoBase` `CargoCommand`, `CargoRun` and `CargoTest` `Configuration`s. Then we should add context awareness to the configurations. As usual, the go plugin is a good example of this feature.

alexpana:
I can implement this if no one is working on it.

matklad:
Started implementing this in https://github.com/intellij-rust/intellij-rust/commit/283f8ac1b441147c1227ebc5992b7dc75fbbacb0.

My idea is simply to use single `CargoCommand` configuration and add a number of `ConfigurationProviders` for tests and binary files.

matklad:
Implemented running a single test in https://github.com/intellij-rust/intellij-rust/commit/b35451a6bb69bab91bce06a130a4bf29504a7467

PR #366, opened 2016-05-01
(RUNCFG) Separate configuration types.

alexpana:
PR for https://github.com/intellij-rust/intellij-rust/issues/311

It's only the first working prototype. I haven't added any tests, but I'd like a code review before continuing.


src/main/kotlin/org/rust/cargo/runconfig/run/CargoRun.kt
```diff
@@ -0,0 +1,25 @@
...
+: CargoRunConfigurationBase<CargoRunConfigurationModule>(name,
+                                                         CargoRunConfigurationModule(project),
+                                                         configurationType) {
+
+    override var command: String = "run"
+
+    override fun getConfigurationEditor(): SettingsEditor<out RunConfiguration> = CargoRunConfigurationEditForm()
+}
+
+class CargoRunConfigurationModule(project: Project) : RunConfigurationModule(project)
```
matklad: I think it should be possible to use one `RunConfigurationModule` for all `RunConfigurations`. That is, instead of making `CargoRunConfigurationBase` generic, you can specify one concrete type, as in the [go plugin](https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/src/com/goide/runconfig/GoRunConfigurationBase.java#L47)

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRun.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.intellij.execution.configurations.RunConfigurationModule
+import com.intellij.openapi.options.SettingsEditor
+import com.intellij.openapi.project.Project
+import org.rust.cargo.runconfig.CargoRunConfigurationBase
+import org.rust.cargo.runconfig.ui.CargoRunConfigurationEditForm
+
+/**
+ * [RunConfiguration] implementation that runs the main binary of a local package
+ */
+class CargoRun(project: Project,
```
matklad: I think it's better to suffix all different configurations with `RunConfiguration`, so you'll have `CargoTestRunConfiguration` and `CargoRunRunConfiguration`. This is verbose, but helps when you "GoTo Class".

src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandType.kt
```diff
@@ -9,20 +9,18 @@ import com.intellij.openapi.project.Project
...
-                CargoCommandConfiguration(project, "Cargo", this@CargoCommandRunConfigurationType)
+                CargoCommand(project, "Cargo", this@CargoCommandType)
 
             override fun configureBeforeRunTaskDefaults(providerID: Key<out BeforeRunTask<BeforeRunTask<*>>>,
                                                         task: BeforeRunTask<out BeforeRunTask<*>>) {
                 if (providerID == CompileStepBeforeRun.ID) {
-                    // We don't use jps, so we don't need to execute `Make` task
-                    // before run configuration is executed
                     task.isEnabled = false
                 }
```
matklad: This `ConfigurationFactory` is repeated in all three different configurations. I think this duplication can be avoided by introducing `CargoConfigurationFactory` class. It will also allow us to preserve the comment ;)

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRun.kt
alexpana: I consciously avoided suffixing with `RunConfiguration` because of `CargoRunRunConfiguration` :smile:

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRunProducer.kt
```diff
@@ -0,0 +1,47 @@
...
+        }
+        return false
+    }
+
+    private fun isMainFile(file: PsiFile?): Boolean {
+        if (file == null) {
+            return false
+        }
+
+        return file.name == "main.rs" && file.parent?.name == "src" && fileContainsMainFn(file)
```
matklad: This check can be made more precise. There is `module.cargoProject`, which stores the information about the project as cargo sees it. In particular, file is a `MainFile` if it is a binary target. No need to fix it in this PR, the proper check can be implemented later.

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRunProducer.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    private fun fileContainsMainFn(file: PsiFile?): Boolean {
+        if (file == null) {
+            return false
+        }
+
+        val declaredFunctions = PsiTreeUtil.findChildrenOfType(file, RustFnItemImpl::class.java)
```
matklad: You generally should not use `RustBlaImpl`. Instead, use `RustBla` interfaces. That is,  `PsiTreeUtil.findChildrenOfType(file, RustFnItem::class.java)`.

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRunProducer.kt
```diff
@@ -0,0 +1,47 @@
...
+        for (declaredFunction in declaredFunctions) {
+            if (declaredFunction.identifier.text == "main") {
+                return true
+            }
+        }
+        return false
```
matklad: This can be spelled more concisely as  `delcaredFunctions.any {it.name == "main"}`.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    override var command: String = "test"
+    override var arguments: String = ""
+
+    private var _testName: String = ""
+    var testName: String
+        get() = _testName
+        set(value) {
+            _testName = value
+            arguments = "--test " + testName
+        }
```
matklad: I am not super fluent with Kotlin properties, but I think the same can be written as 

```
var testName: String = ""
    get() = field
    set(value) {
        field = value
        arguments = "--test " + testName
    }
```

Also, mutating `arguments` when you change `testName` seems a bit not obvious... Maybe it is possible to override `arguments` getter in such a way that it inspects `testName` and appends `--test` if necessary?

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRun.kt
matklad: Yeah, this sounds a funny, but I think it is better than plain `CargoRun` :) Reminds me of C++ template template parameters

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestProducer.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.impl.RustFnItemImpl
+
+class CargoTestProducer : RunConfigurationProducer<CargoTest>(CargoTestType()) {
+    override fun isConfigurationFromContext(configuration: CargoTest?, context: ConfigurationContext?): Boolean {
+        return configuration?.name == configurationName(context)
+    }
+
+    override fun setupConfigurationFromContext(configuration: CargoTest?,
```
matklad: I think you can omit `?` for all arguments here. The base function does not have nullability annotations, so Kotlin conservatively uses `?` here. However the docstring does not mentions `null` at all, so I think that all these are in fact `NotNull`. 

This in an important gotcha when overriding a Java method in Kotlin. If you see that all parameters are nullable it usually means missing annotations. You need to figure out right nullability manually in such case.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestProducer.kt
```diff
@@ -0,0 +1,59 @@
...
+        for (declaredFunction in declaredFunctions) {
+            if (isTestFunction(declaredFunction)) {
+                return true
+            }
+        }
+        return false
+    }
+
+    private fun isTestFunction(function: RustFnItemImpl): Boolean {
+        if (!function.outerAttrList.isEmpty()) {
```
matklad: You can use [queryAttributes](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustItemImplMixin.kt#L35) to query attributes.

matklad:
Hm, looks like test configuration will be a bit more involved then I expected. 

You can run a specific test *crate* with

```
cargo test --test foo
```

This will work if you have directory layout like

```
src/
  lib.rs
tests/
  foo.rs
  bar.rs
```

You can also run a specific test with

```
cargo test foo
```

This will work if you have the following in your lib.rs

```
...

mod tests {
    use super::*;

    #[test]
    fn foo() {
       ....
    }
}
```

Basically, one use case is running a specific `test` target, and another use case is running a specific test by name. 

I think it would be a good idea to handle this complexity in a separate PR. That is, in this PR we split `RunConfig` into `RunConfigBase`, `CommandRunConfig`, `RunRunConfig` and add a `Producer` for the latter.

In the next PR, we properly add the `TestRunConfig`. What do you think about it?

alexpana:
> In the next PR, we properly add the TestRunConfig. What do you think about it?

I agree. I'm not very familiar with cargo and the conventions used for specifying tests, so more research is warranted on my part.

src/main/kotlin/org/rust/cargo/runconfig/CargoRunConfigurationBase.kt
```diff
@@ -46,14 +56,11 @@ class CargoCommandConfiguration(project: Project,
...
 
-    override fun getConfigurationEditor(): SettingsEditor<out RunConfiguration> =
-        CargoRunConfigurationEditorForm()
-
     override fun getState(executor: Executor, environment: ExecutionEnvironment): RunProfileState? {
         val module = configurationModule.module ?: return null
         val toolchain = module.toolchain ?: return null
         val moduleDirectory = module.cargoProjectRoot ?: return null
-        val args = ParametersListUtil.parse(additionalArguments)
+        val args = ParametersListUtil.parse(arguments + additionalArguments)
```
matklad: I think this `arguments + additionalArguments` is not quite correct. If `arguments` is `foo`, and `additionalArguments` is `bar`, then their concatenation is `foobar` and not `foo bar`. 

I think that the simplest thing to do now is to remove the `arguments` field and `CargoTestRunConfiguration` altogether. It would be easier to add them after we merge the rest of this PR. 

The end design could look like this:

```Kotlin
// in base run configuration
// This is the interface for inheritors, and we don't show this field in
// GUI directly
protected open val arguments: List<String>

// in derived configuration, i.e, Test
override val arguments = listOf("--test", testName)

// This is what we show in GUI. 
var testName: String = ""

```

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRunRunConfigurationProducer.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.RustFnItem
+
+class CargoRunRunConfigurationProducer
+: RunConfigurationProducer<CargoRunRunConfiguration>(CargoRunRunConfigurationType()) {
+    override fun isConfigurationFromContext(configuration: CargoRunRunConfiguration,
+                                            context: ConfigurationContext): Boolean {
+        return configuration.name == "Run " + context.project?.name
```
matklad: nit: you can `"Run ${context.project?.name}"`

matklad: Ah, and you don't actually need `?`: again, because of the missing annotations Kotlin does not know that `context.project` is not null, so it does not warn about unnecessary  `?` here.

src/main/kotlin/org/rust/cargo/runconfig/CargoRunConfigurationBase.kt
alexpana: > I think this arguments + additionalArguments is not quite correct. If arguments is foo, and additionalArguments is bar, then their concatenation is foobar and not foo bar.

Indeed you are correct, sorry for letting that slip.

> I think that the simplest thing to do now is to remove the arguments field and CargoTestRunConfiguration altogether. It would be easier to add them after we merge the rest of this PR.

Will do.

src/main/kotlin/org/rust/cargo/runconfig/run/CargoRunRunConfigurationProducer.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    private fun isMainFile(file: PsiFile?): Boolean {
+        if (file == null) {
+            return false
+        }
+
+        return file.name == "main.rs" && file.parent?.name == "src" && fileContainsMainFn(file)
+    }
+
+    private fun fileContainsMainFn(file: PsiFile?): Boolean {
```
matklad: No need for `?` here

src/test/kotlin/org/rust/cargo/runconfig/RunConfigurationTestCase.kt
```diff
@@ -1,49 +1,40 @@
...
+//        with(result.processHandler) {
+//            addProcessListener(listener)
+//            startNotify()
+//            waitFor()
+//        }
+//        Disposer.dispose(result.executionConsole)
+//        return listener.output
+//    }
+//}
+//
```
matklad: Hm, I think it would be a good idea to uncomment this test :)

matklad:
@alexpana this looks good!

src/main/kotlin/org/rust/cargo/runconfig/CargoRunConfigurationBase.kt
matklad: > sorry for letting that slip.

No need to say sorry, it is impossible to write correct code and everybody makes bugs :)

alexpana:
@matklad  thanks, I'll make the necessary changes when I get home from work today :)

matklad:
I'll close this because of inactivity, but the issue is still open ;)

Issue #367, opened 2016-05-02
Great performance degradation


e-oz:
Please watch [attached video](https://github.com/intellij-rust/intellij-rust/files/245344/slow.mov.zip) to see how currently plugin works with file of 1854 lines. Latest version of nightly update, latest version of IDEA.

IntelliJ IDEA 2016.1
Build #IU-145.970, built on April 18, 2016
JRE: 1.8.0_76-release-b162 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
OS X 10.11.4,
MacBookPro11,1;Intel Core i5;2.4 GHz;8 GB.

Lowest CPU usage when I type: 107%.
I'm ready to send you content of source file by email - just let me know where to send (and please don't publish it).

matklad:
You can send the file to `aleksey.kladov@gmail.com`

e-oz:
done

matklad:
fixed by https://github.com/intellij-rust/intellij-rust/commit/9970c82fb72d28b37ca434356cf4868880aec900

e-oz:
Works awesome, thank you **very** much!

PR #369, opened 2016-05-03
(MISC): avoid unnecessary `subSequence` calls

jajakobyly:
See discussion under 9970c82fb72d28b37ca434356cf4868880aec900

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/lexer/RustEscapesLexer.kt
+++ b/src/main/kotlin/org/rust/lang/core/lexer/RustEscapesLexer.kt
@@ -1,8 +1,9 @@
 package org.rust.lang.core.lexer
 
-import com.intellij.openapi.util.text.StringUtil
+import com.intellij.openapi.util.text.StringUtil.isHexDigit
 import com.intellij.psi.StringEscapesTokenTypes.*
 import com.intellij.psi.tree.IElementType
+import com.intellij.util.text.CharArrayUtil.indexOf
 import org.rust.lang.core.psi.RustTokenElementTypes.*
 
 private const val BYTE_ESCAPE_LENGTH = "\\x00".length
@@ -60,8 +61,8 @@ class RustEscapesLexer private constructor(val defaultToken: IElementType,
 
             when (bufferSequence[i]) {
                 'x'        ->
-                    if (bufferEnd - (i + 1) >= 1 && StringUtil.isHexDigit(bufferSequence[i + 1])) {
-                        if (bufferEnd - (i + 2) >= 1 && StringUtil.isHexDigit(bufferSequence[i + 2])) {
+                    if (bufferEnd - (i + 1) >= 1 && isHexDigit(bufferSequence[i + 1])) {
+                        if (bufferEnd - (i + 2) >= 1 && isHexDigit(bufferSequence[i + 2])) {
                             return i + 2 + 1
                         } else {
                             return i + 1 + 1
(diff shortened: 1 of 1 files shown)


jajakobyly:
@matklad @e-oz please check if it doesn't accidentally recreate #367

e-oz:
@jajakobyly last nightly build works fine

matklad:
@jajakobyly I think it should be possible to reuse `indexOf` from `com.intellij.util.text.CharArrayUtil` here.

matklad:
I mean, to reuse it *inside* the extension method.

matklad:
TIL:

It is unspecified if `CharSequence::subSequence` is O(1) or O(N). At least for `String` since Java 7 it is O(N).

If you want an explicit O(1) sub sequnece, use `CharSequenceSubSequence`. 

`ImmutableText`, which is used in IDEA, has O(1) subSequence.

`CharArrayUtil` has some useful stuff!

jajakobyly:
Ok pushed `CharArrayUtil`'s `indexOf` as an static import, because introducing special extension method looked like a triumph of form over content for me.

jajakobyly:
@matklad looks like the `com.intellij.util.text` package has lots of useful stuff :)

src/main/kotlin/org/rust/lang/core/lexer/RustEscapesLexer.kt
```diff
@@ -82,18 +84,11 @@ class RustEscapesLexer private constructor(val defaultToken: IElementType,
             }
             return i + 1
         } else {
-            val idx = indexOf('\\', start + 1)
-            return if (idx != -1) Math.min(idx, bufferEnd) else bufferEnd
+            val idx = indexOf(bufferSequence, "\\", start + 1, bufferEnd)
+            return if (idx != -1) min(idx, bufferEnd) else bufferEnd
```
matklad: I bet this min always returns `idx`.

src/main/kotlin/org/rust/lang/core/lexer/RustEscapesLexer.kt
```diff
@@ -60,17 +62,17 @@ class RustEscapesLexer private constructor(val defaultToken: IElementType,
...
                         } else {
                             return i + 1 + 1
                         }
                     }
                 'u'        ->
                     if (bufferEnd - (i + 1) >= 1 && bufferSequence[i + 1] == '{') {
-                        val idx = indexOf('}', i + 1)
-                        return if (idx != -1) Math.min(idx + 1, bufferEnd) else bufferEnd
+                        val idx = indexOf(bufferSequence, "}", i + 1, bufferEnd)
+                        return if (idx != -1) min(idx + 1, bufferEnd) else bufferEnd
```
matklad: and also this one.

matklad:
>looks like the com.intellij.util.text package has lots of useful stuff :)

Yep, will use `SevVer` for [this thing](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt#L22) one day

src/main/kotlin/org/rust/lang/core/lexer/RustEscapesLexer.kt
jajakobyly: Yeah, you're right

jajakobyly:
Done

matklad:
Thanks!

PR #377, opened 2016-05-05
Test function template

mitchhentges:
Use `tf` to inflate the following:
```
#[test]
fn $NAME$() {
    $END$
}


matklad:
Nice! I've written some tests recently and wanted this thing. 

But lets make it `tfn`, to be consistent with `tmod`. What do you think about it?

mitchhentges:
Sounds great! Updated.

matklad:
Thanks!

PR #388, opened 2016-05-12
Quick documentation support for resolved items.

### LINKED ISSUE #370: Show documentation in Quick Doc (`Ctrl+Q`)
We need to show quick doc for items. If the user has

```Rust
/// Super useful
fn foo(x: i32) -> u32 { 92 }

fn main() {
   <caret> foo()
}
```

and presses `Ctrl+Q`, she should see "Super useful" in the documentation popup. Here is [the code](https://github.com/intellij-rust/intellij-rust/blob/a16d71528949caff0709abc97edbbb996d3e7f38/src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt) which is responsible for this. `DocComment` nodes are now properly attached to the item PSI. 

@Turbo87 @alexpana you were interested in this :)

Furyhunter:
Implements quick documentation in the Documentation Provider by adding a read-only property `documentation: String?` on `RustNamedItem`. This is not _quite_ the right way to go about implementing this, but it is robust enough that it works over all resolved items.

The Markdown parser available in the IntelliJ API is not quite suitable for formatting the documentation string. It does not properly handle the HTML entity codes necessary to provide space indentation in code blocks, nor does it handle things like strikethrough.

A continued improvement would be to provide the full declaration at the top of the documentation text (i.e. instead of just the name, show `pub fn the_function(t: isize) -> isize`) for functions, and visibility for traits structs and enums.

Fixes #370

### DIFF
--- a/src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
@@ -1,12 +1,16 @@
 package org.rust.ide.documentation
 
 import com.intellij.lang.documentation.AbstractDocumentationProvider
+import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.util.text.MarkdownUtil
 import com.petebevin.markdown.MarkdownProcessor
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.RustImplMethodMemberImpl
 import org.rust.lang.core.psi.impl.mixin.isMut
+import java.util.*
 
 class RustDocumentationProvider : AbstractDocumentationProvider() {
 
@@ -17,12 +21,10 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
                 return if (doc != null) formatDoc(element.name!!, doc)
                 else null
             }
-            // THIS IS A HACK UNTIL RustImplMethodMember IS A RustItem
-            // I have a hunch fixing that will also fix struct member resolution...
+            // Hack until there is a trait interface for "documentable" elements.
             is RustImplMethodMemberImpl -> {
-                val doc = element.documentation
-                return if (doc != null) formatDoc(element.name!!, doc)
-                else null
+                val doc = (element.outerDocumentationLinesForElement + element.innerDocumentationLinesForElement).joinToString("\n")
+                return formatDoc(element.name.orEmpty(), doc)
             }
             else -> {
                 return null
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustItemImplMixin.kt
```diff
@@ -26,15 +27,28 @@ abstract class RustItemImplMixin : RustStubbedNamedElementImpl<RustItemStub>
 
         override fun getPresentableText(): String? = name
     }
+
+    override val documentation: String?
```
matklad: So, let's implement this as an extension method on `RustItem`:

```
val RustItem.documentation: String? get {
    ...
}
```

You can place it in the `RustItemImplMixin.kt` file

src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
```diff
@@ -2,12 +2,43 @@ package org.rust.ide.documentation
...
+import org.rust.lang.core.psi.impl.RustImplMethodMemberImpl
 import org.rust.lang.core.psi.impl.mixin.isMut
 
 class RustDocumentationProvider : AbstractDocumentationProvider() {
 
+    override fun generateDoc(element: PsiElement?, originalElement: PsiElement?): String? {
+        when (element) {
+            is RustItem -> {
+                val doc = element.documentation
+                return if (doc != null) formatDoc(element.name!!, doc)
```
matklad: It is a bad idea to use `!!`.

Name of an `item` can be null because the parser recovers aggressively (at least, it should).

If you have a syntax error in your code 

```
struct {

}
``` 

Then, as soon as the parser sees the `struct` keyword, it will generate a `RustStructItem` PSI node, but it's name will be `null`.

src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
```diff
@@ -1,13 +1,46 @@
...
+import com.petebevin.markdown.MarkdownProcessor
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.RustImplMethodMemberImpl
 import org.rust.lang.core.psi.impl.mixin.isMut
+import java.util.*
 
 class RustDocumentationProvider : AbstractDocumentationProvider() {
 
+    override fun generateDoc(element: PsiElement?, originalElement: PsiElement?): String? {
+        when (element) {
```
jajakobyly: This overall `when` construct ends with `return` in each possible branch, so it might be nice to rewrite this into return expression (`return when (element) {...}`), and because this is the only expression in the whole method as `fun generateDoc(...) = when (element) {...}`).

matklad: @jajakobyly I'll tweak this a bit myself before merging, so hold your comments for a couple of minutes :)

jajakobyly: ok ;)

matklad:
Thanks a lot, @Furyhunter. I've [tweaked this a bit](https://github.com/intellij-rust/intellij-rust/commit/7e42dc99d25a0c2ac77e75ec9735e252b8b50b23) and merged, but I believe you can do a couple of improvements here.

src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
matklad: Done!

matklad:
The main improvement is, of course, adding more tests! :)

src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
```diff
@@ -42,3 +75,66 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
...
+                    val s = (c.metaItem.litExpr)?.stringLiteral?.text?.removeSurrounding("\"")?.trim()
+                    if (s != null) add(s)
+                } else if (c is PsiComment && c.tokenType == RustTokenElementTypes.OUTER_DOC_COMMENT) {
+                    val s = c.text.substringAfter("///").trim()
+                    add(s)
+                }
+            }
+        }
+
+        return lines
```
matklad: This bit can be simplified by using

```
        childOuterIterator.asSequence().map { 
            ...
        }.toList()
```

src/main/kotlin/org/rust/ide/documentation/RustDocumentationProvider.kt
```diff
@@ -42,3 +75,66 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
...
+                    } else if (c is RustInnerAttr && c.metaItem.identifier.textMatches("doc")) {
+                        val s = (c.metaItem.litExpr)?.stringLiteral?.text?.removeSurrounding("\"")?.trim()
+                        if (s != null) add(s)
+                    } else if (c is PsiComment && c.tokenType == RustTokenElementTypes.INNER_DOC_COMMENT) {
+                        val s = c.text.substringAfter("//!").trim()
+                        add(s)
+                    }
+                }
+            }
+        }
```
matklad: same here

PR #281, opened 2016-03-09
(INSPECTION) Unnecessary parenthesis inspection

zjhmale:
Fix #274 

snapshot

PR #392, opened 2016-05-12
Add unnecessary parenthesis inspection.

### LINKED ISSUE #274: Remove unnecessary parenthesis around condition quick fix/inspection
Oftentimes I write

```Rust
if (cond) {
    // do stuff
}
```

and it's cumbersome to remove parenthesis manually. I wish there were an inspection for this.

Docs:
http://www.jetbrains.org/intellij/sdk/docs/tutorials/custom_language_support/annotator.html
http://www.jetbrains.org/intellij/sdk/docs/tutorials/custom_language_support/quick_fix.html

Example:
https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/inspections/UnresolvedModuleDeclarationInspection.kt

Test:
https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/inspections/RustInspectionsTest.kt

Furyhunter:
Derived partially from @zjhmale's implementation #281, but checks all expressions generally as well as if predicates.

There are currently some errors that occur when using the quick fixes; upon applying them, the inspection state is not updated, and it is possible to reapply the inspection, causing an internal error to occur.

Fixes #274

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/UnnecessaryParenthesisInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/UnnecessaryParenthesisInspection.kt
@@ -1,0 +1,67 @@
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.LocalQuickFixBase
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElementVisitor
+import org.rust.lang.core.psi.RustBinaryExpr
+import org.rust.lang.core.psi.RustIfExpr
+import org.rust.lang.core.psi.RustParenExpr
+import org.rust.lang.core.psi.RustVisitor
+import org.rust.lang.core.psi.impl.RustParenExprImpl
+import org.rust.lang.core.psi.util.parentRelativeRange
+
+class UnnecessaryParenthesisInspection : RustLocalInspectionTool() {
+    override fun getDisplayName(): String = "Unnecessary parenthesis"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : RustVisitor() {
+            override fun visitParenExpr(expr: RustParenExpr) {
+                val subExpr = expr.expr
+                when (subExpr) {
+                    !is RustBinaryExpr -> {
+                        if (expr.parent !is RustIfExpr)
+                            holder.registerProblem(expr, "Redundant parenthesis around unambiguous expression", RemoveParenthesis)
+                    }
+                }
+            }
+
+            override fun visitIfExpr(expr: RustIfExpr) {
+                for (e in expr.exprList) {
+                    // Why is this a... list of expressions? Under what circumstances can an if have more than one?
+                    if (e is RustParenExpr) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/inspections/UnnecessaryParenthesisInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+    override fun getDisplayName(): String = "Unnecessary parenthesis"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : RustVisitor() {
+            override fun visitParenExpr(expr: RustParenExpr) {
+                val subExpr = expr.expr
+                when (subExpr) {
+                    !is RustBinaryExpr -> {
+                        if (expr.parent !is RustIfExpr)
+                            holder.registerProblem(expr, "Redundant parenthesis around unambiguous expression", RemoveParenthesis)
```
matklad: I am not quite sure that this is entirely correct. Let's concentrate on the most important problem of redundant parenthesis in `if`. 

If we got it working, adding more cases would be easy.

src/main/kotlin/org/rust/ide/inspections/UnnecessaryParenthesisInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    object RemoveParenthesis : LocalQuickFixBase("Remove parenthesis in expression") {
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            val expr = descriptor.psiElement as? RustParenExprImpl
+            if (expr != null) {
```
matklad: I would try to use `expr.replace(expr.expr)` here, but I am not sure that it will work...

matklad:
We definitely need a test for this!

src/main/kotlin/org/rust/ide/inspections/UnnecessaryParenthesisInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+                    !is RustBinaryExpr -> {
+                        if (expr.parent !is RustIfExpr)
+                            holder.registerProblem(expr, "Redundant parenthesis around unambiguous expression", RemoveParenthesis)
+                    }
+                }
+            }
+
+            override fun visitIfExpr(expr: RustIfExpr) {
+                for (e in expr.exprList) {
+                    // Why is this a... list of expressions? Under what circumstances can an if have more than one?
```
matklad: This is because the `else` branch can be either a block, an `if` or an `if_else` expression. So the grammar uses the common ancestor `expr` for the else branch.

matklad:
>There are currently some errors that occur when using the quick fixes; upon applying them, the inspection state is not updated, and it is possible to reapply the inspection, causing an internal error to occur.

I am not super familiar with inner working of quick fixes, but sounds like some PSI modifications here confuse the IDE. You can use PSI viewer plugin to make sure that the PSI is correct after all modifications. And take a look at similar inspections in Java/Kotlin/Go.

Furyhunter:
Kotlin's plugin handles this with an Intention over expressions in general instead, and has no corresponding inspection for showing yellow squigglies. Perhaps it would be smarter to follow that route, and provide a separate inspection specifically for `if` that simply runs the intention over the if's predicate expression.

I will reimplement this in a separate branch.

PR #393, opened 2016-05-13
Add "remove parentheses from expression" intention and annotators for unnecessary parentheses.

### LINKED ISSUE #274: Remove unnecessary parenthesis around condition quick fix/inspection

Furyhunter:
This is a reimplementation of #392 that is built over a simple Intention instead of an Inspection. The Intention will be available wherever there is a parentheses expression, and the annotators provide:

* Unnecessary if expression parentheses ala `if (true) {} -> if true {}`
* Redundant parentheses ala `((true)) -> (true)`

These annotations will still appear in code analysis; annotators should be used in place of inspections for simpler checks according to the [SDK documentation](http://www.jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/code_inspections_and_intentions.html), for inspections are meant to be able to address issues across the whole project. It is also based loosely on the implementation in the [Kotlin plugin's own intention.](https://github.com/JetBrains/kotlin/blob/b7581b35c9bfee9e197c82ac7c8a8c69906012ea/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveUnnecessaryParenthesesIntention.kt#L24)

Test cases for all features have also been added.

Fixes #274


jajakobyly:
What about `while`, `match` and `return`? Also what do you think about immediately simplifying `((foo))` to `foo`? 

I think it might be also better to implement this generally for `RustParenExpr` and customise annotation message basing on the expr parent.

matklad:
>What about while, match and return?

The cost of implementing a feature is a (`fixed_cost + sum_{sub_features} additional_sub_feaute_cost` ). I find it more productive to pay only the `fixed_cost` first and then pay `sub_feature` fees incrementally. 

That is, I would prefer first to implement the inspection only for the `if` expression. This requires adding an inspection class, the test class, the tests folders, the entry in the `plugin.xml` and just a tiny bit of the actual logic. 

But, after this first piece of work is merged, expanding the inspection for more cases becomes much simpler, because we know that the base implementation is solid and well tested :) 

And yes, eventually we need to implement this for `while`, `match`, `return` and possible some other cases :)

src/main/kotlin/org/rust/ide/annotator/RustExpressionAnnotator.kt
```diff
@@ -0,0 +1,41 @@
...
+
+class RustExpressionAnnotator : Annotator {
+    override fun annotate(element: PsiElement, holder: AnnotationHolder) {
+        when (element) {
+            is RustIfExpr -> {
+                checkIfParentheses(element, holder)
+            }
+            is RustParenExpr -> {
+                checkImmediateChildIsParen(element, holder)
+            }
```
matklad: It is possible to omit `{}` in when:

```
            is RustIfExpr -> checkIfParentheses(element, holder)
            is RustParenExpr -> checkImmediateChildIsParen(element, holder)
```

src/main/kotlin/org/rust/ide/intentions/RemoveParenthesesFromExprIntention.kt
```diff
@@ -0,0 +1,26 @@
...
+    override fun isAvailable(project: Project, editor: Editor?, element: PsiElement): Boolean {
+        if (!element.isWritable) return false
+
+        element.parentOfType<RustParenExpr>() ?: return false
+        return true
+    }
+
+    override fun invoke(project: Project, editor: Editor?, element: PsiElement) {
+        val parentExpr = element.parentOfType<RustParenExpr>() ?: return
+        parentExpr.replace(parentExpr.expr!!)
```
matklad: Again, `(parentExpr.expr ?: return)` is better.

matklad:
Thanks!

matklad:
>What about while, match and return?

The issue https://github.com/intellij-rust/intellij-rust/pull/393

PR #397, opened 2016-05-16
(TYPE): add type inference for `self` argument

matklad:
Ok, so I've thought about it for a while, and I still don't know how type inference should be implemented :)

For one thing, Rust type system at the moment is basically implementation defined and, quoting [docs](infer) from the compiler:

>This is loosely based on standard HM-type inference, but with an extension to try and accommodate subtyping. There is nothing principled about this extension; it's sound---I hope!---but it's a heuristic, ultimately, and does not guarantee that it finds a valid typing even if one exists (in fact, there are known scenarios where it fails, some of which may eventually become problematic).

The relevant modules are [infer] and [typecheck]

[infer]: https://github.com/rust-lang/rust/tree/89430be0c1a45600d7e89ee4e5d3bb68ec8902bb/src/librustc/infer
[typecheck]: https://github.com/rust-lang/rust/tree/89430be0c1a45600d7e89ee4e5d3bb68ec8902bb/src/librustc_typeck

So I think that the right approach here is to start incrementally, one test a time,  build an understanding of the type system together with the test suite and then refactor the implementation guided by the acquired knowledge. 

So this PR implements the bare minimum needed for [this test](https://github.com/intellij-rust/intellij-rust/blob/d42fb1d21f4d972076575e18259f44fdce094943/src/test/resources/org/rust/lang/core/resolve/fixtures/type/method.rs) to pass. 

@alexeykudinkin @atsky please review


src/main/kotlin/org/rust/lang/core/type/TypeInference.kt
```diff
@@ -0,0 +1,35 @@
...
+            val target = path.reference.resolve()
+            when (target) {
+                is RustSelfArgument -> {
+                    val impl = target.parentOfType<RustImplItem>()
+                    impl?.type?.resolvedType ?: RustUnknownType(manager)
+                }
+                else -> RustUnknownType(manager)
+            }
+        }
+        else -> RustUnknownType(manager)
```
matklad: I've dropped alignment of `->` for this snippet because it is ugly. In fact, now I am in favor of dropping alignment rules from our code style: most of the time they make the code slightly prettier, but sometimes they make it absolutely ugly. In contrast, non aligning things always works consistently good.

alexeykudinkin: I'd elaborate on that rule: we're aligning only those `->` being placed on the adjacent lines.

src/main/kotlin/org/rust/lang/core/type/RustResolvedType.kt
```diff
@@ -0,0 +1,26 @@
+package org.rust.lang.core.type
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.SyntheticElement
+import org.rust.lang.core.psi.RustImplItem
+import org.rust.lang.core.psi.RustImplMethodMember
+import org.rust.lang.core.psi.impl.mixin.isStatic
+
+
+interface RustResolvedType: PsiElement, SyntheticElement {
```
alexeykudinkin: What is the motivation to make types part of the PSI hierarchy? Seems questionable to me.

Furyhunter:
Autocomplete of `self.<caret>` doesn't work for resolving methods unless there's a `()` after the `.` (i.e. `self.f<caret>()` works, `self.f<caret>` does not), the PSI tree has a `RustFieldExpr` otherwise which doesn't yield a reference. The ambiguity here is difficult; should a hypothetical `RustFieldReference` yield variant arrays that include _both methods and fields_, or should the grammar be made more explicit? How would one even go about doing the latter?

Also, reference and binding modes should be considered in resolved types, so that for example the variants array can filter methods which require a mutable receiver when the receiver is immutable.

Furyhunter:
I have further expanded upon this branch on my own to provide some basic support for type inference and resolution on several different kinds of expressions: https://github.com/Furyhunter/intellij-rust/tree/types

As of this writing it is missing support for resolving inherited impls for primitive types (i.e. from prelude/stdlib) and tuple types. It does not support pattern bind matching beyond the basic name pattern. It does not resolve type bounds for generic impls. There are also no tests written yet. It does, however, manage to provide variants for methods on many kinds of expressions when the type is resolved successfully.

alexeykudinkin:
@Furyhunter why not create P/R then? If it's based on the current branch we'd take a look at it right after this one get landed.

src/main/kotlin/org/rust/lang/core/type/RustResolvedType.kt
matklad: An excellent question! There it is not needed in this at all, fixed :) 

Looks like I originally implemented this because I was doing PSI_MODIFICATION_COUNT based caching "inside" the types themselves. But now caching happens on the real PSI elements, which is the right way I think.

src/main/kotlin/org/rust/lang/core/type/RustStructType.kt
```diff
@@ -0,0 +1,18 @@
+package org.rust.lang.core.type
+
+import org.rust.lang.core.psi.RustImplItem
+import org.rust.lang.core.psi.RustMod
+import org.rust.lang.core.psi.RustStructItem
+import org.rust.lang.core.psi.util.parentOfType
+
+data class RustStructType(val struct: RustStructItem) : RustResolvedType {
+    override val inheritedImpls: Collection<RustImplItem> by lazy {
+        val potentialImpls = struct.parentOfType<RustMod>()?.items.orEmpty().filterIsInstance<RustImplItem>()
```
alexeykudinkin: `potentialImpls` seems redundant here.

src/main/kotlin/org/rust/lang/core/type/TypeInference.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.rust.lang.core.psi.util.parentOfType
+import org.rust.lang.utils.psiCached
+
+val RustExpr.inferredType: RustResolvedType by psiCached {
+    when (this) {
+        is RustPathExpr -> {
+            val target = path.reference.resolve()
+            when (target) {
+                is RustSelfArgument -> {
+                    val impl = target.parentOfType<RustImplItem>()
```
alexeykudinkin: Same here: `impl` seems redundant

src/main/kotlin/org/rust/lang/core/type/RustStructType.kt
matklad: Yep, simplified.

Furyhunter:
Is there anything in particular blocking this PR? The basic API structure seems reasonably solid, but I am beginning to question certain parts in my own experimentation branch.

* Some expressions are coerced into other types by the compiler based on the context i.e. numeral literals: `let a: i64 = 4`. When inferring the type of `4`, it would need the surrounding context to be coerced. This leaves a question of responsibility. Resolved types probably need some interface of the kind `RustResolvedType.coercableTo(t: RustResolvedType): Boolean`. I think the best approach is to evaluate the expression's type first, then in a placement where the type is expected to be something else, check if it _can_ be coerced, then if it can simply set the coerced-to type.
* Type aliases introduce some issues for comparing resolved types. An interface `RustResolvedType.equivalentTo(t: RustResolvedType): Boolean` should be added, since `equals` doesn't really fit this well.
* Evaluating the type of a name binding requires evaluating the type of the bound expression and then destructuring it all the way down. Similar to the issue above, it requires more context than the element and its children to comprehend. I'm not even sure how to phrase this issue, but it seems a bit messy.
* I question the use of data class instances for each expression evaluated. This could balloon into a large number of unnecessary objects in memory for a large codebase. Letting items materialize their own types that get reused for type resolution might help a lot.
* Where do type parameters and trait bounds fall into this system?

Implementing my proof of concept is showing some deficiencies in the grammar and generated Psi as well. Anywhere there's a block or brace body (like fn bodies), `getExpr()` returns the last `expr` in a series of statement expressions in the body and ignores the rest. Grammatically this is a bit strange. It may also be inappropriate to make statement expressions chain like a linked sequence in the Psi tree, but I believe that is closer to what the compiler AST might represent. It would make the PSI debug viewer harder to use though.

alexeykudinkin:
@matklad, please, squash both `Simplify` commits, and we're ready to go.

matklad:
@alexeykudinkin no need to squash anything manually if you want to combine all the commits into a single one: merge pull requrest button now offers "squash" option which does exactly this (that's why I've switched from rebasing my pull requests to just adding fixup style commits on top)

Issue #73, opened 2015-11-05
Improve code formatting


matklad:
While we should leverage [rustfmt](https://github.com/nrc/rustfmt) for code formatting, it is still necessary to provide indentation support for enjoyable editing. 

* [code](https://github.com/alexeykudinkin/intellij-rust/blob/master/src/org/rust/lang/formatter/RustFormattingBlock.kt)
* [docs](http://www.jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/code_formatting.html)
* formatting in the [go plugin](https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/src/com/goide/formatter/GoFormattingModelBuilder.java)

matklad:
The major pain at the moment is the support for chain calls. That is, this

```Rust
foo().bar()
     .baz()
```

is not aligned.

matklad:
Current formatter is here: https://github.com/intellij-rust/intellij-rust/tree/master/src/main/kotlin/org/rust/ide/formatter

zjhmale:
is this can be closed after #289 merged?

matklad:
Not quite. There is at least three reasons to implement our own formatter in addition to `rustfmt`.

* refactorings and snippets use IDEAs formatter
* `rustfmt` is not quite done yet, so by itself it is not a 100% solution for all formatting problems
* Not everybody has `rustfmt` installed

zjhmale:
:+1: the second point matters most i think.

jajakobyly:
I'll try to do sth in this area

jajakobyly:
Hmm should anything be done here apart from providing similar configurability to rustfmt? Wrapping could be imho delegated to rustfmt, the same may apply to inserting commas etc.

matklad:
Yeah, I guess the issue can be closed because it is too vague. However, I think there are still some bugs with formatter which need fixing. For example, I found that it does not indents macro calls properly. I can't check the example, but it was somethign along the lines of

```
try!(write!(wherer, "str", 
     param))
```

getting reformatted to 

```
try!(write!(wherer, "str", 
param))
```

Will create an issue once I have time to properly reproduce this.

PR #403, opened 2016-05-19
(FMT): Improve code formatting: Indentation

jajakobyly:
Yay, a big boy now. This is a continuation of #362 that implements better indentation handling. Also, I have removed dependency on `AbstractBlock` (and rewritten most of it directly inside `AbstractRustFmtBlock`) because it has pretty inflexible API in terms of `ChildAttributes`.

Creative test cases are very, **very** welcome.

Most inspiration was taken from [Go formatter](https://upsource.jetbrains.com/go-lang-idea-plugin/file/ede4f339b4968989ea1a5bfcfefa0894874fef81/src/com/goide/formatter/GoFormattingModelBuilder.java) and [Erlang formatter](https://upsource.jetbrains.com/intellij-erlang/file/7984bf797e54ae436755c123c42d930f784125e4/src/org/intellij/erlang/formatter/ErlangFormattingModelBuilder.java) this time.

The overall structure (which is convenient to me) of implementation of this formatter starts to emerge. The whole logic will be implemented as plain functions inside small/medium-sized files called `spacing.kt`, `indent.kt`, `alignment.kt` and so, which will be wired to the formatter inside formatting blocks. This approach greatly improve readability of block classes imho (just compare [`GoFormattingBlock`](https://upsource.jetbrains.com/go-lang-idea-plugin/file/ede4f339b4968989ea1a5bfcfefa0894874fef81/src/com/goide/formatter/GoFormattingModelBuilder.java?nav=6390:6390:focused) with `RustFmtBlock`).

#73 

To do:
 - [x] Finish that indentation stuff
 - [x] Reorganise `formatter` package; remove `internal` modifiers
 - [x] **LOTS** of tests, especially for auto indent


src/main/kotlin/org/rust/ide/formatter/AbstractRustFmtBlock.kt
```diff
@@ -1,26 +1,48 @@
...
+) : ASTBlock {
 
+    override fun getNode(): ASTNode = myNode
+    override fun getTextRange(): TextRange = node.textRange
+    override fun getAlignment(): Alignment? = myAlignment
     override fun getIndent(): Indent? = myIndent
+    override fun getWrap(): Wrap? = myWrap
+
+    override fun getSubBlocks(): List<Block> = mySubBlocks
+    private val mySubBlocks: List<Block> by lazy { buildChildren() }
```
matklad: Do we really need laziness in the formatter? Why not simply `private val mySubBlocks = buildChildren()`?

matklad:
I like the reorganization changes here! Perhaps we can merge then right now (the tests need to be green of course) ?

src/main/kotlin/org/rust/ide/formatter/utils.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.tree.IElementType
+import com.intellij.psi.tree.TokenSet.orSet
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RustCompositeElementTypes.*
+import org.rust.lang.core.psi.RustTokenElementTypes.*
+import org.rust.lang.core.psi.impl.RustFile
+import com.intellij.psi.tree.TokenSet.create as ts
+
+internal val KEYWORDS = ts(*IElementType.enumerate { it is RustKeywordTokenType })
```
matklad: The `internal` modifier is pretty useless for us, because we have only one module. It's OK to make this just public.

jajakobyly: These files are going to land in separate package after indentation commit is done

jajakobyly:
> I like the reorganization changes here! Perhaps we can merge them right now (the tests need to be green of course) ?

Most of them are done inside indentation commit as wild ideas.

src/main/kotlin/org/rust/ide/formatter/AbstractRustFmtBlock.kt
jajakobyly: Everybody does this this way:
```kotlin
if (mySubBlocks == null) {
    mySubBlocks = buildChildren();
}
```
So did I :smile:, though I understand that this doesn't have to be synchronized, but I don't know any non-synchronized `lazy { ... }` alternative in Kotlin stdlib.

matklad: There is at least `LazyThreadSafetyMode`, though I'd rather avoid it and use synchronized lazy or eager evaluation.

src/main/kotlin/org/rust/ide/formatter/utils.kt
matklad: But `internal` is not about packages. Basically, `internal == public` for our use case.

jajakobyly: Ouch, you're right. I thought it was analogous to _no modifier_ in Java. Sorry

matklad: Yeah, "package private" is impossible to express in Kotlin, and it drives me nuts every time I want something shared across several files :)

src/main/kotlin/org/rust/ide/formatter/AbstractRustFmtBlock.kt
matklad: And yes, it is currently impossible to override Java function with Kotiln property :(

src/main/kotlin/org/rust/ide/formatter/RustFormattingModelBuilder.kt
```diff
@@ -16,6 +16,7 @@ class RustFormattingModelBuilder : FormattingModelBuilder {
     override fun createModel(element: PsiElement, settings: CodeStyleSettings): FormattingModel {
         val ctx = RustFmtBlockContext.create(settings)
         val block = AbstractRustFmtBlock.createBlock(element.node, null, Indent.getNoneIndent(), null, ctx)
+        /** / com.intellij.formatting.FormattingModelDumper.dumpFormattingModel(block, 2, System.err) // */
```
jajakobyly: I know this is hacky, but this line is pretty useful when debugging

matklad: You can also write a local helper function: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L456-L465

src/main/kotlin/org/rust/ide/formatter/indent.kt
```diff
@@ -0,0 +1,56 @@
...
+}
+
+fun computeIndent(block: RustFmtBlock, child: ASTNode): Indent? {
+    val parentType = block.node.elementType
+    val childType = child.elementType
+    val childPsi = child.psi
+    return when {
+        block.node.isDelimitedBlock -> getIndentIfNotDelim(child, block.node)
+
+        block.node.isFlatBlock && block.getUserData(INDENT_BETWEEN_FLAT_BLOCK_BRACES) == true ->
```
matklad: Hm, why do we need `UserData` here? We control `RustFmtBlock`, so it should be possible to implement necessary functionality directly.

src/test/kotlin/org/rust/ide/formatter/RustAutoIndentTestCase.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    fun testModItem() = doTest()
+    fun testModItem2() = doTest()
+    fun testForeignModItem() = doTest()
+    fun testForeignModItem2() = doTest()
+
+    private fun doTest() = checkByFile {
+        myFixture.type('\n')
+    }
+}
```
matklad: Neat! We definitely need such tests because autoindent is super imoprtant

src/main/kotlin/org/rust/ide/formatter/indent.kt
jajakobyly: We want to put additional data which will be necessary only in very specific, rare cases. I don't want to pollute `RustFmtBlock` interface with such properties.

matklad: Ok. And what about creating file-private extension methods?

jajakobyly: These ones will require us to do O(n) lookups (to find if there are any `{` before). Nope ;)

jajakobyly:
Ok let's say it's ready as for now. Eventual bugs may be fixed in separate PRs

matklad:
Thanks! Any plans for chain call support? :)

jajakobyly:
Of course, alignment is the next step

PR #404, opened 2016-05-19
(T, MISC): fix bad eol handling

jajakobyly:
These tests failed on Windows because `assertThat(actual).isEqualTo(expected)` doesn't differentiate between `\n` and `\r\n`


matklad:
Thanks!

PR #406, opened 2016-05-21
(FMT): small polish

jajakobyly:
Mostly readability improvements

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
@@ -77,7 +77,7 @@ fun createSpacingBuilder(commonSettings: CommonCodeStyleSettings,
         .between(PARAMS_LIKE, RET_TYPE).spacing(1, 1, 0, true, 0)
         .before(WHERE_CLAUSE).spacing(1, 1, 0, true, 0)
         .applyForEach(BLOCK_LIKE) { before(it).spaces(1) }
-        .beforeInside(LBRACE, FLAT_BLOCKS).spaces(1)
+        .beforeInside(LBRACE, FLAT_BRACE_BLOCKS).spaces(1)
 
         .between(ts(IDENTIFIER, FN), PARAMS_LIKE).spaceIf(false)
         .between(IDENTIFIER, GENERIC_PARAMS).spaceIf(false)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
```diff
@@ -254,8 +259,10 @@ private fun SpacingContext.blockMustBeMultiLine(): Boolean {
     }
 }
 
+/**
+ * Assumes that left is before right and they are siblings, otherwise bad things will happen.
+ */
 private fun countNonWhitespaceASTNodesBetween(left: ASTNode, right: ASTNode): Int {
-    // TODO(jajakobyly): assert that left is before right and they are siblings
```
matklad: I would use `check` here instead of the comment

jajakobyly: The problem is that checking condition is quite expensive (it will require linear scans of the AST tree)

jajakobyly: And I don't know how Kotlin compiles `assert(...)` calls

matklad: Hm, would `check(left.parent == right.parent && left.startOffset < right.startOffset)` do the necessary checks in O(1)?

Kotlin has `assert` which is the same as Javas `assert`. It is disabled or enabled depending on the runtime flag.

There are also design by contract style  `require` (preconditions) and `check` (invariant) functions which are always on. I usually use those.

jajakobyly: Uh, forgot about `startOffset` :sweat_smile: 

The problem with assert is that Java's `assert ...` does not evaluate conditions when assertions are disabled, while Kotlin's `assert(...)` do. Just compare these two guys with and without `-ea`:
```java
class Main {
    public static void main(String[] args) {
        // no -ea: prints nothing
        // -ea: prints Mooo
        assert boo();
    }
    static boolean boo() {
        System.out.println("Mooo");
        return false;
    }
}
```
```kotlin
fun main(args: Array<String>) {
    // no -ea: prints Mooo
    // -ea: prints Mooo
    assert(boo())
}

fun boo(): Boolean {
    System.out.println("Mooo")
    return false
}
```

matklad: TIL about Kotlin assert :) 

https://youtrack.jetbrains.com/issue/KT-12454
https://youtrack.jetbrains.com/issue/KT-7540

One more reason to prefer always on assertions :)

Issue #312, opened 2016-03-31
Syntax highlighting for types


netvl:
Currently the plugin highlights all types (except type parameters) as identifiers. Ideally all types should be highlighted as types, so they would be more visible.

I understand that the partial reason to treat them as just identifiers is that there is no real identifier resolution available, but:
* it should be possible to highlight at least primitive types
* it may be possible to inspect whether an identifier is in a type position (after `:` in function signatures, for example) and highlight it accordingly.

What do you think?

matklad:
>I understand that the partial reason to treat them as just identifiers is that there is no real identifier resolution available

Well, we do have identifier resolution available :) It is not complete yet, but it is constantly improving. We don't use it for highlighting at the moment, but this is definitely on the roadmap.

netvl:
@matklad I meant that, as far as I can see, it is still impossible to navigate, say, from type use site to its definition :) But of course, this is all just speculations on my part, I may be totally wrong :)

matklad:
Basic navigation within one file already works:

![test](https://cloud.githubusercontent.com/assets/1711539/14172981/e10bd8c2-f743-11e5-8c48-33f685efb16f.gif)

Navigation between files of a single project should work if you use beta version of Cargo and import the project properly.

Navigation to the standard library sources kinda works, but it is a bit complex to setup right now, I'm improving this at the moment. 

There are surely some bugs with resolve! If **Goto declaration** does not work for your particular use case than you can file an issue/ask about it on gitter.

netvl:
Are these changes published already? I'm asking because I've just installed the plugin from the nightly plugins repository and keyboard shortcuts for navigation do not work, including types in the single file.

But anyway, this is not what this issue is about :) I'll play with it more later and file a new issue if there are indeed problems with it.

matklad:
Yes, Goto definition was implemented several months ago.

matklad:
@netvl I believe this is fixed by @mfarrugi  in #503. The default colors for enum and struct are the same as for any identifier though, so you need to change them to see it:

![idents](https://cloud.githubusercontent.com/assets/1711539/16567353/7be02d6c-4226-11e6-8102-9525b1fcc92f.png)

Perhaps we should have different default colors for structs and enums?

mfarrugi:
The default is set to `Class`, so that'd be a PR for jetbrains. 
It'll play nicely with other color schemes, though.

PR #408, opened 2016-05-22
Add documenting for trait members.

Furyhunter:
Adds documentation and named element support to all possible trait members.

Also adds tests and adjusts parse for trait type members slightly. This is because type parameters are considered an independent named element, and it did not make grammatical sense to have that rule be used in the place where associated types are declared. The tests checking for this have been adjusted accordingly.

**Tests added**:

* documentation:
  * `trait_assoc_type`
  * `trait_const`
  * `trait_method`
  * `trait_method_provided`

**Tests changed**:

* complete parses _(due to aformentioned adjustments to trait type member declaration rule)_
  * `impls`
  * `traits`

**Notes**:

Currently, impl block const and type members do _not_ get named element support. This is something that should be considered since it affects documenting support.


matklad:
Awesome!

> This is because type parameters are considered an independent named element, and it did not make grammatical sense to have that rule be used in the place where associated types are declared. 

Most likely we will change the PSI around type parameters and traits bounds. It makes sense to abstract over all different ways to specify trait bounds (bounds in generics, associated types, traits and, of course, where clauses).

The related highlighting issue is #312. Looks like we don't need to differentiate between different kinds of types, and instead have on `type` color in our highlighter.

PR #409, opened 2016-05-22
Add more annotations for unnecessary parens.

### LINKED ISSUE #395: Highlight more redundant parenthesis
Follow up of https://github.com/intellij-rust/intellij-rust/pull/393

Highlight redundant parenthesis in `while (true)`, `match (1)`, `return (1)`.

I am not sure about the proper error level here. If `rustc` itself reports a warning then we probably should use `warning` instead of `weak warning`.

Furyhunter:
Fixes #395


matklad:
Thanks! An interesting task would be to make `it` and friends checks more precise. You do need parenthesis in this case, for example:

```Rust
struct S { f: bool }
fn main() {
    if (S {f: true}.f) {}
}
```

But I think this is a very rare corner case, so we don't need to fix this urgently. Or am I mistaking an there are some relatively frequent patterns which need parenthesis?

PR #412, opened 2016-05-22
Restore compatibility with the IDEA 145.* (IDEA 15.*) branch

alexeykudinkin:


matklad:
Ouch, I have a question actually :)

alexeykudinkin:
Go ahead. Its gonna land in trunk in some way anyway.

matklad:
Do we really need to support `15.0.4`? The latest minor version is `15.0.6`, and `15.0.4` does not include a security update (is this actually correct?).

With `15.0.6` (and I think 15.0.5 as well) plugin works without this changes.

matklad:
Yep, `15.0.5` works just fine.

alexeykudinkin:
We're not breaking compatibility with the previous versions until we're sure that no one (or almost no one) of our users are affected.

Especially in minor updates.

alexeykudinkin:
Will trash this after complete migration to `15.0.6`.

matklad:
Ok, then one other question: `15.0.3` in `build.gradle` looks kinda arbitrary. Why not `15.0.0` or `15.0.6` then? I would prefer to run local tests against `2016.1` and use CI to test for backwards compatibility .

alexeykudinkin:
Yeah, my bad. Reverted `2016.1.2` as a default option.

alexeykudinkin:
`15.0.3` has been chosen as the last one non-breaking version.

PR #413, opened 2016-05-22
(FMT): Improve code formatting: Alignment & chain calls

### LINKED ISSUE #355: Improve chain call formatting
Problem: chain calls should be formatted beautifully.

Looks like the rust community is [split](https://github.com/rust-lang/rust/pull/33020#discussion_r59975354) about what the proper formating is (rustfmt [issue](https://github.com/rust-lang-nursery/rustfmt/issues/439)).

Basically, there are two options:

a) simply indent

```
calculate().bar().iter()
    .map(|i| ... )
    .filter(...)
    .collect(...)
```

b) align to dot

```
calculate().bar()
           .iter()
           .map(|i| ... )
           .filter(...)
           .collect(...)
```

I think we should implement both styles and add an option "align method calls" (off by default, imo).  We should not insert newlines ourselves.

jajakobyly:
This is a continuation of #403, this time focuses on better alignment and chain calls (these guys are quite hard to handle properly because of their PSI structure) support. Also, again some refactorings.

Creative test cases are very, **very** welcome.

Most inspiration was taken from [Kotlin formatter](https://github.com/JetBrains/kotlin/tree/master/idea/formatter/src/org/jetbrains/kotlin/idea/formatter) [(2)](https://github.com/JetBrains/kotlin/tree/master/idea/src/org/jetbrains/kotlin/idea/formatter) this time.

Fixes #355 

#73


src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
```diff
@@ -0,0 +1,42 @@
+package org.rust.ide.formatter
+
+import com.intellij.formatting.Alignment
+import com.intellij.lang.ASTNode
+
+/**
+ * Mimics [com.intellij.formatting.alignment.AlignmentStrategy], but offers more flexible API
+ */
+interface RustAlignmentStrategy {
```
jajakobyly: I have allowed myself to "fork" `AlignmentStrategy`, the overall idea is very nice, but IntelliJ's implementation is terrible for our use.

src/main/kotlin/org/rust/ide/formatter/settings/RustLanguageCodeStyleSettingsProvider.kt
```diff
@@ -55,7 +55,11 @@ class RustLanguageCodeStyleSettingsProvider : LanguageCodeStyleSettingsProvider(
     override fun getDefaultCommonSettings(): CommonCodeStyleSettings =
         CommonCodeStyleSettings(language).apply {
             RIGHT_MARGIN = 99
-            initIndentOptions()
+            initIndentOptions().apply {
+                // FIXME(jajakobyly): It's a hack
+                // Nobody else does this and still somehow achieve similar effect
+                CONTINUATION_INDENT_SIZE = INDENT_SIZE
```
jajakobyly: Does anybody know how other plugins make continuation indents look like normal indents? I couldn't find how it's done in Go, Erlang, Java & Kotlin plugins (or maybe I just don't know where to search :sweat_smile: )

src/test/resources/org/rust/ide/formatter/fixtures/auto_indent/chain_call4_after.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let moo = foo().bar().baz()
+                   .moo()
+                   <caret>;
```
jajakobyly: I'm not sure whether it's desired behaviour. Having complete statement (already placed `;`) we may give us a hint, that user might want to continue chain call.

src/test/resources/org/rust/ide/formatter/fixtures/spacing_after.rs
```diff
@@ -207,8 +207,7 @@ impl<T: PartialEq> Vec<T> {
     }
 }
 
-pub fn from_elem<T: Clone>(elem: T, n: usize) -> Vec<T> {
-}
+pub fn from_elem<T: Clone>(elem: T, n: usize) -> Vec<T> {}
```
jajakobyly: Also I'm not 100% sure that this change is also desired. I have noticed that rustfmt always transforms `{<whitespace>}` to `{}`, so I have implemented similar behaviour here (previously it allowed 1 `\n` between braces).

src/main/kotlin/org/rust/ide/formatter/blocks/RustFmtBlock.kt
```diff
@@ -0,0 +1,70 @@
...
+    private fun buildChild(child: ASTNode, alignment: RustAlignmentStrategy): AbstractRustFmtBlock {
+        if (node.isFlatBlock && child.isBlockDelim(node)) {
+            replace(INDENT_MET_LBRACE, null, true)
+        }
+
+        val block = createBlock(child, alignment.getAlignment(child, node), computeIndent(child), null, ctx)
+
+        if (node.elementType in FN_DECLS && child.elementType in PARAMS_LIKE) {
+            // Pass alignment object to parameter list
+            block.putUserData(PARAMETERS_ALIGNMENT, getUserData(PARAMETERS_ALIGNMENT))
```
jajakobyly: Ugh, but I couldn't invent better and more readable way to do that alignment sharing. We are determining block alignment during creation, so we can't just navigate formatting tree.

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
```diff
@@ -0,0 +1,58 @@
...
+     */
+    fun getAlignment(child: ASTNode, parent: ASTNode?): Alignment?
+
+    companion object {
+        /**
+         * Always returns `null`.
+         */
+        val NULL_STRATEGY = object : RustAlignmentStrategy {
+            override fun getAlignment(child: ASTNode, parent: ASTNode?): Alignment? = null
+        }
```
matklad: Instead of

```
class Foo {
    companion object {
        val QUUX = object I {}
    }
}
```

you can simply

```
class Foo {
    object QUUX: I {}
}
```

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
matklad: Hm, and can a function type be used as an alignment strategy? like

```
val NULL_ALIGNMENT = {child: AstNode, parent: AstNode? -> null}
```

matklad: Typealiases:

```
typealias AlignmentStragey = (child:ASTNode, parent: ASTNode?) -> Alignment?
```

are a planned feature for Kotlin

src/main/kotlin/org/rust/ide/formatter/blocks/RustFmtBlock.kt
matklad: perhaps we can pass parent block to the `createBlock` function as a context?

src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.openapi.util.Key
+import com.intellij.psi.tree.TokenSet
+import org.rust.ide.formatter.RustAlignmentStrategy
+import org.rust.ide.formatter.blocks.RustFmtBlock
+import org.rust.lang.core.psi.RustCompositeElementTypes.*
+
+/**
+ * Stores common alignment object for function declarations's parameters, return type & where clause.
+ */
+val PARAMETERS_ALIGNMENT: Key<Alignment> = Key.create("PARAMETERS_ALIGNMENT")
```
matklad: Hm, I am still not sure that `UserDataHolder` is a superior approach to plain old fields :)

src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    in PARAMS_LIKE ->
+        RustAlignmentStrategy.lazyWrapCond(
+            { this.getUserData(PARAMETERS_ALIGNMENT) },
+            { c, p -> !c.isBlockDelim(p) })
+
+    in FN_DECLS ->
+        RustAlignmentStrategy.lazyWrapFiltered(
+            { this.getUserData(PARAMETERS_ALIGNMENT) },
+            TokenSet.create(RET_TYPE, WHERE_CLAUSE))
```
matklad: I would use something like 

```
RunstAlignmentStrategy.inheretedFiltered(
  this, TokenSet.create(RET_TYPE, WHERE_CALUSE)
)
```

This lazyness is a bit misleading, because it is in fact recomputed every time, and may cause strange mutable state related bugs. "Inheritance" is not better in this regard, but it is at least less general and easier to reason about because of this.

src/main/kotlin/org/rust/ide/formatter/settings/RustLanguageCodeStyleSettingsProvider.kt
matklad: I don't know from the top of my head.

matklad:
Looks like this supposed to fix one half of https://github.com/intellij-rust/intellij-rust/issues/355

jajakobyly:
The other half is incoming ;)

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
jajakobyly: Nice idea, though quite cumbersome without type aliases. I'll consider this

jajakobyly: Actually I can translate the interface (only it) to Java and exploit https://kotlinlang.org/docs/reference/java-interop.html#sam-conversions as a kind of workaround for now.

src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
jajakobyly: The naming will probably change. I think of renaming `wrap* -> shared*` and `lazyWrap* -> computed*`

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
matklad: I think that using a java interface is the worst of three options

src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
matklad: The point is, if every call to lasyWrap stats with `{ this.getUserData`, then it make sense to move this common part from caller to the callee.

jajakobyly: But such change has a cost of reducing versatility of `RustAlignmentStrategy`. Probably (but not 100% sure), `lazyWrap` will be also used to implement alignment for chain calls. Let's wait until all features of this PR will be implemented and then do proper cleanups

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
jajakobyly: What do you think about such API?
```kotlin
fun RustFmtBlock.getAlignmentStrategy(): RustAlignmentStrategy = when (node.elementType) {
    TUPLE_EXPR, TUPLE_TYPE, ARG_LIST, ENUM_TUPLE_ARGS ->
        RustAlignmentStrategy.shared()
            .cond { c, p -> !c.isBlockDelim(p) } 
            .cfg(...) // enable configuration

    in PARAMS_LIKE ->
        RustAlignmentStrategy.lazy { this.getUserData(PARAMETERS_ALIGNMENT) } // actually this can be implemented as a real lazy
            .cond { c, p -> !c.isBlockDelim(p) }
            .cfg(...)

    in FN_DECLS ->
        RustAlignmentStrategy.lazy { this.getUserData(PARAMETERS_ALIGNMENT) }
            .filter(TokenSet.create(RET_TYPE, WHERE_CLAUSE))
            .cfg(...)

    PAT_ENUM ->
        RustAlignmentStrategy.shared()
            .cond { c, p -> this.getUserData(INDENT_MET_LBRACE) == true && !c.isBlockDelim(p) }

    else -> RustAlignmentStrategy.NULL_STRATEGY
}
```

The problem with plain functions is that there is no clean way to store alignment object

src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
jajakobyly: Hmm, I think, that method calls, parameters and flat blocks may just get separate formatting block class. This way user data could be eliminated and`RustFmtBlock` would be less cluttered

jajakobyly: But this might be huuge over-engineering

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
matklad: Yep, builder DSL looks nice. 

Can you please elaborate why exactly you need lazyness? Is it because you set the alignment after the block is crated?  Will [this](https://github.com/intellij-rust/intellij-rust/pull/413#discussion_r64387669) help?

The laziness make me feel a little uneasy.

jajakobyly: > Can you please elaborate why exactly you need laziness? Is it because you set the alignment after the block is crated? Will this help?

I need to read user data when alignment for (first) child is created, not when `getAlignmentStrategy` is fired, because there is no guarantee that something won't change between `getAlignmentStrategy` and children building loop.

I'm aware that this design is ugly, and I'm working on replacing this whole mess with separate formatting blocks for _special_ PSI nodes. Unfortunately I don't have a lot of spare time currently (finals are coming), so this may wait for a bit.

matklad: Good luck with the finals!

jajakobyly: Thanks! :smile:

jajakobyly: >  Will [this](https://github.com/intellij-rust/intellij-rust/pull/413#discussion_r64387669) help?

Hmm I'm tinkering around it now and it turns out that formatting blocks are created BFS-style: block's grandchildren are created **after** it's children are fully created. This means that in order to pass alignment object we have to:
  - store it somewhere for the whole lifetime of fn decl block (property or user data)...
  - ...or do some reference-passing shenaningans (this means introducing lots of separate formatting blocks)

OMG this is going to yield horrible code :sob: Neither of these options is appealing to me.

jajakobyly: I am also thinking about ditching `alignment.kt` and `indent.kt` and integrating their logic directly inside `AbstractRustFmtBlock`. Just take a look at [this monster](https://github.com/intellij-rust/intellij-rust/pull/413/files#diff-253bc0944bbecd6130053f584e283569R35)

matklad:
`FmtContext`s looks nice! I've just use data classes `.copy` super function recently myself in the ResolveScope P

jajakobyly:
Done

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.intellij.formatting.Alignment
+import com.intellij.lang.ASTNode
+import com.intellij.psi.tree.IElementType
+import com.intellij.psi.tree.TokenSet
+
+interface RustAlignmentStrategy {
+    /**
+     * Requests current strategy for alignment to use for given child.
+     */
+    fun getAlignment(child: ASTNode, parent: ASTNode?, childCtx: RustFmtContext): Alignment?
```
matklad: Just a random design note, not something that should be changed in this case. When you design an API based on callbacks, it may make sense to use parameter object pattern (`getAlignmemnt(ctx: AlignmentContext)`), because it allows to evolve the API better. Adding a new parameter to the interface breaks all the implementations, adding a field to a context does not.

jajakobyly: Got it, but I don't think it is necessary here.

src/main/kotlin/org/rust/ide/formatter/RustAlignmentStrategy.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Apply this strategy only when child element is in [tt].
+     *
+     * @param isBlackList   denotes how filter set is interpreted. If set to `false`, the filter is a white list:
+     *                      only children which element types are in filter set will get wrapped alignment;
+     *                      otherwise, the filter works as a black list: these children will be ignored.
+     */
+    fun filter(vararg tt: IElementType, isBlackList: Boolean = false): RustAlignmentStrategy =
```
matklad: `isBlackList` is never used. What do you think about renaming this to `alignIf`, and the adding `notAlignIf` in the future when `balklisting` will be needed?

matklad:
Wow, this works great and is even pretty readable! A great job, @jajakobyly !

PR #423, opened 2016-05-26
(RES): change ModuleIndex algorithm

### LINKED ISSUE #421: Make RustModulesIndex independent of CargoProjectWorkspace.
Currently, `RustModulesIndex` depends on `CargoProjectWorkspace`, because key in the index is `RustCratePath`.This is problematic, because the index becomes invalid once workspace data is updated. Here is a [slight modification] of `CargoProjectResolveTestCase` which demonstrates the issue. It is also pretty visible when working with real projects. 

The core of the problem is that `ModuleIndex` violates

>The data returned by DataIndexer.map() must depend only on input data passed to the method, and must not depend on any external files. Otherwise your index will not be correctly updated when the external data changes, and you will have stale data in your index.

contract. This can be avoided by using `FileIncludeIndex` trick: store a name as a key and a referece offest as a value in the index and get the final result by resolving the reference at offeset. Also with some effort I can get rid of the JarFileSystem, so that we should be able to use `FileIncludeIndex` directly. 

[slight modification]: https://github.com/intellij-rust/intellij-rust/commit/b67bf90c05d9bd54bd449a174b7e0d777af266b0

matklad:
Module index stores a mapping from names to mod_decls. Each mod decl is
represented as an offset in the file.

Finding a super module for a given file is a two step process.

First, mod_decl candidates are found in the index based on their
names (this may give false positives).

Then, each candidate is resovled and if it resolves to the target
file (modulo usual complication with the `origianlFile` for completion),
then it is the answer.

closes #421 

Looks like I have not hit a roadblock. Tests pass and looks like it actually works :) 
@alexeykudinkin please review

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/indexes/RustModulesIndex.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/indexes/RustModulesIndex.kt
@@ -1,40 +1,46 @@
 package org.rust.lang.core.resolve.indexes
 
-import com.intellij.openapi.module.Module
+import com.intellij.openapi.util.io.FileUtil
 import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.util.indexing.FileBasedIndex
 import com.intellij.util.indexing.ID
-import org.rust.lang.core.names.RustQualifiedName
+import org.rust.cargo.util.getPsiFor
 import org.rust.lang.core.psi.RustMod
 import org.rust.lang.core.psi.impl.RustFile
-import org.rust.lang.core.psi.impl.cratePath
-import org.rust.lang.core.psi.util.module
-import org.rust.lang.core.resolve.RustResolveEngine
+import org.rust.lang.core.psi.util.parentOfType
 
-interface RustModulesIndex {
+object RustModulesIndex {
+    val ID: ID<RustModulesIndexExtension.Key, RustModulesIndexExtension.Value> =
+        com.intellij.util.indexing.ID.create("org.rust.lang.indexes.RustModulesIndex")
 
-    companion object {
+    fun getSuperFor(mod: RustFile): RustMod? {
+        val project = mod.project
+        val key = getKey(mod) ?: return null
 
-        val ID: ID<RustCratePath, RustQualifiedName> =
-            com.intellij.util.indexing.ID.create("org.rust.lang.indexes.RustModulesIndex")
+        var result: RustMod? = null
+        FileBasedIndex.getInstance().processValues(ID, key, null, { file, value ->
+            val reference = project.getPsiFor(file)?.findReferenceAt(value.referenceOffset)
+                ?: return@processValues true
 
-        fun getSuperFor(mod: RustFile): RustMod? =
-            mod.containingFile.originalFile.let { file ->
-                mod.module?.let { module ->
-                    file.cratePath?.let { path ->
-                        findByHeterogeneous(
-                            FileBasedIndex.getInstance()
-                                .getValues(ID, path, GlobalSearchScope.allScope(module.project))
-                                .firstOrNull(),
-                            module
-                        )
-                    }
-                }
+            if (reference.resolve() == mod.originalFile) {
+                result = reference.element.parentOfType<RustMod>()
+                false
+            } else {
+                true
             }
+        }, GlobalSearchScope.allScope(project) )
 
-        private fun findByHeterogeneous(path: RustQualifiedName?, module: Module): RustMod? =
-            path?.let {
-                RustResolveEngine.resolve(path, module).resolved as RustMod?
-            }
+        return result
+    }
+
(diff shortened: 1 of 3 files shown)


matklad:
@alexeykudinkin I would prefer to move with this PR a bit faster then usual, because it should greatly improve the quality of resolve/completion (and there were a number of complains about the suddenly breaking completion in the gitter, which I attribute to the non-determenistic nature of the current ModuleIndex).

alexeykudinkin:
I have few concerns about it, will deliver them on Sunday.

Best regards,
Alexey Kudinkin

alexeykudinkin:
First of all, it doesn't solve the problem of the stale-data residing in cache in any way.

matklad:
>First of all, it doesn't solve the problem of the stale-data residing in cache in any way.

Can you please elaborate this a bit? Tests [check] that the index is independent of the cargo project information (current implementation of ModulesIndex fails the check). 

Also the [Indexer] obviously does not touch anything except the data supplied via the `inputData` value, so there is simply no way for the index to become stale. 


[check]: https://github.com/intellij-rust/intellij-rust/blob/module-index-mkII/src/test/kotlin/org/rust/cargo/toolchain/CargoProjectResolveTestCase.kt#L39

[Indexer]: https://github.com/intellij-rust/intellij-rust/blob/module-index-mkII/src/main/kotlin/org/rust/lang/core/resolve/indexes/RustModulesIndexExtension.kt#L48-L61

matklad:
And sorry for being pedantic, but this is an index, and not a cache. The platform guarantees that it is always up to date and has complete information as long as the `Indexer` does not touch anything except the data derived from `inputData`.

alexeykudinkin:
> Also the Indexer obviously does not touch anything except the data supplied via the inputData value, so there is simply no way for the index to become stale.

Initially i may have dependency on the `rand` crate, and later remove that. All the module-inclusion info related to that external crate will still stay inside the index.

> And sorry for being pedantic, but this is an index, and not a cache. The platform guarantees that it is always up to date and has complete information as long as the Indexer does not touch anything except the data derived from inputData.

Index and cache being used interchangeably throughout the platform codebase.

alexeykudinkin:
Though, this P/R obviously solves the problem of storing _invalid_ data, not the _stale_ one.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustModulesIndex.kt
```diff
@@ -1,40 +1,46 @@
...
+
+        var result: RustMod? = null
+        FileBasedIndex.getInstance().processValues(ID, key, null, { file, value ->
+            val reference = project.getPsiFor(file)?.findReferenceAt(value.referenceOffset)
+                ?: return@processValues true
 
-        private fun findByHeterogeneous(path: RustQualifiedName?, module: Module): RustMod? =
-            path?.let {
-                RustResolveEngine.resolve(path, module).resolved as RustMod?
+            if (reference.resolve() == mod.originalFile) {
```
alexeykudinkin: This approach, doesn't seem to be reentrant-safe: how do we guarantee, we're not stuck in a resolve-loop?

matklad: The reference at offset is guaranteed to be a mod_decl (`mod foo;`). `mod_decl` resolving is [non recursive] and does not and should not call any other resolve routines. 

I think that that "reference is mod_decl" invariant can be made explicit by switching to stub based index instead of file based index. I believe it is possible to implement this index on top of stubs, but I'd rather use a simpler file based approach first. 

[non recursive]: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L92-L111

matklad:
>Initially i may have dependency on the rand crate, and later remove that. All the module-inclusion info related to that external crate will still stay inside the index.

Yep, but I think it should not harm in any way? The inclusion data for `rand` files would still be valid, we simply won't use it because there will be no more `rand` related files in the project or libraries.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustModulesIndex.kt
alexeykudinkin: Yeah, right. Followed wrong path, assuming, that we might trigger `getSuperFor` another time during resolve.

alexeykudinkin:
Unfortunately, it does, since there is no other way to clear the index from that stale data, other than full-rebuild.

alexeykudinkin:
Though, that is different kind of a problem (compounded with the fact that we need to resolve `mod_decl` every time), that may result into revisiting `RustModulesIndex` scheme in the nearest future.

matklad:
>compounded with the fact that we need to resolve mod_decl every time

Yeah, I've been looking into resolve caching. It should help not only with `mod_decl` resolve, but also with `super` resolve.

PR #426, opened 2016-05-28
(PSI): Fix assertion related to Self type in impls

himikof:
This adds a resolve test a a way to trigger the assertion, but it seems that actually implementing resolve for Self right now is hard, so the test just checks for Self not resolving. If this is not desirable, I can drop the test altogether.


matklad:
Thanks! `cself` should definitely be there. But I am curious, what kind of assertion was triggered? Was it something related to completion?

himikof:
Assertion was in `com.intellij.psi.impl.SharedPsiElementImplUtil.addReferences`, and stack trace did not contain Rust plugin code. Apparently, `RustQualifiedReferenceImpl.getRangeInElement` returned null, because `RustPathImplMixin.nameElement` returned null. Come to think of it, there is probably a `@NotNull` annotation missing on `PsiReference.getRangeInElement`, and return type of `RustQualifiedReferenceImpl.getRangeInElement` is wrong, too.

matklad:
This should make it a bit more robust: https://github.com/intellij-rust/intellij-rust/commit/5ee8e2561950f12ab97eb1d52d103a363187411d

PR #427, opened 2016-05-28
(RES): Resolve type parameters in generic type items

himikof:


matklad:
Thanks, I've indeed missed this one!

PR #432, opened 2016-05-31
Add block folding.

### LINKED ISSUE #429: Folding of definitions and block expressions
The text editor should allow the user to fold blocks.

Relevent IntelliJ SDK documentation: http://www.jetbrains.org/intellij/sdk/docs/tutorials/custom_language_support/folding_builder.html

Furyhunter:
Adds some basic block folding and unit tests for folding.

Fixes #429.

Something of note: for some reason, I can't wrap the folds in `<fold></fold>` in the unit test fixtures, only `<fold>` at the start of the fold works. The folds work correctly in the editor, but the unit tester doesn't generate the closing tag.


src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.intellij.lang.folding.FoldingDescriptor
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.RustFile
+import java.util.*
+
+class RustFoldingBuilder() : FoldingBuilderEx() {
```
matklad: No need for empty parenthesis in `RustFoldingBuilder()`

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -0,0 +1,85 @@
...
+    override fun getPlaceholderText(node: ASTNode): String? {
```
matklad: Can be more concisely expressed as `override fun getPlaceholderText(node: ASTNode): String = "{...}" `

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -0,0 +1,85 @@
...
+        return "{...}"
+    }
+
+    override fun buildFoldRegions(root: PsiElement, document: Document, quick: Boolean): Array<out FoldingDescriptor> {
+        if (root !is RustFile) return emptyArray()
+
+        val descriptors: MutableList<FoldingDescriptor> = ArrayList()
+
+        PsiTreeUtil.findChildrenOfType(root, RustBlockExpr::class.java).forEach {
```
matklad: Can you add

```Kotlin
inline fun <reified T : PsiElement> PsiElement.descendantsOfType(): Collection<T>
```

to [RustPsiExtensions]? 

Alternatively, I think this can be expressed shorter (and perhaps faster) if you use recursive `RustVisitor`. 

[RustPsiExtensions]: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/util/RustPsiExtensions.kt#L25

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -0,0 +1,85 @@
...
+        PsiTreeUtil.findChildrenOfType(root, RustModItem::class.java).forEach {
+            val rbrace = it.rbrace;
+            if (rbrace != null) {
+                descriptors += FoldingDescriptor(it.node, TextRange(it.lbrace.textOffset, rbrace.textOffset))
+            }
+        }
+        return descriptors.toTypedArray()
+    }
+
+    override fun isCollapsedByDefault(node: ASTNode): Boolean {
```
matklad: Again, expression body will look a bit cleanly here (there is `alt+Enter` intention to switch between block and expression body).

matklad:
I have no idea why tests fail with `</fold>`. Can you investigate this? Everything seems to work fine, but may be something is wrong nonetheless?

matklad:
Aha, I've figured it out, the test do not work at all  :)

The `text=''` part is actually mandatory for the opening `<fold>` tag, so, when you have a test case like 

```
fn main() <fold>{
}
```

Then the `<fold>` is not deleted from the document when you build folding model. And, because the document contains a syntax error, no folds are produced at all. However, because `<fold>` is present literally in the resulting text, the tests pass.

Can you change tests to

```
fn main() <fold text='{...}'>{
}</fold>
``` 
?

Also, you need to implement `DumbAware` for the folding builder to make the tests pass

```
class RustFoldingBuilder : FoldingBuilderEx(), DumbAware {
```

Also it would allow to build folding regions even when indices are updating.

PR #435, opened 2016-06-01
Add qualified ref completions from std prelude.

### LINKED ISSUE #405: Complete symbols from the prelude
```
fn main() {
    let x: Op<caret>
}
```

should complete to 

```
fn main() {
    let x: Option
}
```

if the user has set up the Rust standard library.

Furyhunter:
Fixes #405

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
@@ -2,15 +2,19 @@ package org.rust.lang.core.completion
 
 import com.intellij.psi.PsiFile
 import com.intellij.psi.util.PsiTreeUtil
+import org.rust.cargo.util.preludeModule
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.RustFile
 import org.rust.lang.core.psi.impl.mixin.basePath
 import org.rust.lang.core.psi.impl.mixin.letDeclarationsVisibleAt
 import org.rust.lang.core.psi.impl.rustMod
+import org.rust.lang.core.psi.util.module
 import org.rust.lang.core.psi.util.parentOfType
 import org.rust.lang.core.psi.util.visibleFields
 import org.rust.lang.core.resolve.enumerateScopesFor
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
```diff
@@ -28,7 +32,15 @@ object RustCompletionEngine {
...
 
         val visitor = CompletionScopeVisitor(ref)
-        for (scope in enumerateScopesFor(ref)) {
+        val scopes = enumerateScopesFor(ref).toMutableList()
+
+        val rootMod = RustResolveUtil.getCrateRootModFor(ref)
+        if (rootMod != null) {
+            val preludeMod = rootMod.module?.preludeModule
+            if (preludeMod != null && preludeMod is RustFile) scopes += preludeMod
+        }
```
matklad: I think that you can simply do `ref.module?.preludeModule` without resolving the crate root. When we implement proper handing of `#[no_std]` attributes and the like we will certainly need to consult with the crate root file. For now it's better not to guess how that API will look like and instead use the simplest solution.

src/test/kotlin/org/rust/lang/core/completion/RustStdlibCompletionTest.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.assertj.core.api.Assertions
+import org.rust.cargo.util.cargoProject
+import org.rust.lang.core.resolve.RustMultiFileResolveTestCaseBase
+
+class RustStdlibCompletionTest : RustMultiFileResolveTestCaseBase() {
+    override val dataPath = "org/rust/lang/core/completion/fixtures/stdlib"
+
+    override fun getProjectDescriptor(): LightProjectDescriptor = WithStdlibRustProjectDescriptor()
+
+    fun testHasStdlibSources() {
```
matklad: I think there is no need to duplicate this test both in completion and in resovle.

matklad:
Could you please rebase the branch?

matklad:
Was fixed with https://github.com/intellij-rust/intellij-rust/pull/444. Thanks for the test!

Issue #436, opened 2016-06-01
Create LookupElements for completions


HybridEidolon:
Right now the direct Named Elements are used in `getVariants` for the different reference types. The completions work, but it's hard to differentiate between functions, bindings, and modules because the icons flicker a lot in the editor.

Using LookupElements in the variants array allows IntelliJ to display rich information about each completion, and also attach rules for inserting the completion.

Some notes:
- `#[deprecated]` is now stable and can be used by everyone. LookupElements should use strikethrough for deprecated elements.
- Bold face could be used for field references and impl methods specific to the type, and non-bold for trait methods. This is not particularly useful without type system support.
- An InsertHandler should be attached for fn call and method completions to add the parentheses and set the cursor inside them. For method calls, type system support is needed. Macros should add the !, but as of this writing macro items are not included in completions nor are references made for them at all.
- Wrapping LookupElements in `PrioritizedLookupElement.withPriority` allows the completion list to be sorted. This allows elements of similar kinds and proximity to be sorted together (modules, fn items, consts, statics, bindings, ...).
- The type string could be added for certain kinds of elements, but this cannot be fully resolved without type system support.

SDK doc reference: http://www.jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/code_completion.html

Example of LookupElement creation in Go-lang plugin: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/3936626d8c3d1cda0273fb31c823c42f76ec1235/src/com/goide/completion/GoCompletionUtil.java#L314

matklad:
Closing: we already have some simplistic lookup elements and insertion handlers. There's a lot room for improvement of course!

PR #438, opened 2016-06-01
Create rich lookup element information for completions.

Furyhunter:
This is simply a starting point for which robust completion information can be shown to the user and does not currently have InsertHandlers, which are context dependent. Elements are not sorted either. Parameters for functions and enum variants are shown as well as some more context information, such as return types, if it is available. Type information is simply the type text since type resolution does not offer an interface for getting canonical type names.

In action:

image

image

Related to #436 but does not _completely_ implement yet.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -1,9 +1,14 @@
 package org.rust.lang.core.completion
 
+import com.intellij.codeInsight.completion.InsertHandler
+import com.intellij.codeInsight.completion.InsertionContext
 import com.intellij.codeInsight.completion.util.ParenthesesInsertHandler
 import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.codeInsight.lookup.LookupElementBuilder
+import org.rust.cargo.util.cargoProject
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.RustFile
+import org.rust.lang.core.psi.util.module
 import org.rust.lang.core.psi.util.parentOfType
 
 fun RustNamedElement.createLookupElement(): LookupElement {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -0,0 +1,101 @@
...
+        is RustUseGlob           -> createLookupElement()
+        is RustTraitItem         -> createLookupElement()
+        is RustStructItem        -> createLookupElement()
+        is RustEnumVariant       -> createLookupElement()
+        is RustFieldDecl         -> createLookupElement()
+        else                     -> LookupElementBuilder.createWithIcon(this).withLookupString(name ?: "")
+    }
+}
+
+fun RustFnItem.createLookupElement(): LookupElement {
```
matklad: This functions should be private. Alternatively, they all can be just inlined in that `when`. Overloading of extension methods can seem a bit confusing.

src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
```diff
@@ -84,5 +84,5 @@ private fun RustNamedElement?.completionsFromResolveScope(): Collection<RustName
         else                -> emptyList()
     }
 
-private fun Collection<RustNamedElement>.toVariantsArray(): Array<RustNamedElement> =
-    filter { it.name != null }.toTypedArray()
+private fun Collection<RustNamedElement>.toVariantsArray(): Array<out Any> =
```
matklad: Can this be changed to `out LookupElement`?

Furyhunter:
I've rebased this PR. There are some issues, possibly compiler related, that required some workarounds, in LookupElements.kt.

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -0,0 +1,67 @@
...
+            .withTypeText(type.text)
+        is RustPathElementImpl -> {
+            // re-export needs to be resolved
+            val resolved = reference.resolve()
+            resolved?.createLookupElement() ?: LookupElementBuilder.createWithIcon(this)
+        }
+        is RustUseGlobElement -> {
+            // There is a problem here with ambiguous getReference(): RustReference.
+            // Completely possible this is a compiler issue. Not sure why though.
+            val that = this as RustNamedElement
```
matklad: I think this is a pretty reasonable error. `RustUseGlob` is no longer a `RustNamedElement`, so this branch is a dead code. See https://github.com/intellij-rust/intellij-rust/commit/fe072310d335e7a26c80e090f51c3d7d41fe0c66 and https://github.com/intellij-rust/intellij-rust/commit/6be06285388ce2f884e243d3cb426298f048d9fe for motivation for the change with NamedElements. 

It makes sense to check if some other branches are dead code as well. The one with `RutsPathElementImpl` seems like it.

matklad: Maybe we should add a test for this function?

Something along the lines of 

```
myFixture.configureByText("""
fn foo() -> i32 { 92 }
""")

val element = myFixter.findElementByText(foo).parentOfType<RustNamedElement>()!!
val lookup = element.createLookup()

// assert that icon is not null
// assert proper text
```

src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt
```diff
@@ -32,7 +34,12 @@ class RustFile(
     override val `super`: RustMod?
         get() = RustModulesIndex.getSuperFor(this)
 
-    override val modName: String? = if (name != RustMod.MOD_RS) FileUtil.getNameWithoutExtension(name) else parent?.name
+    override val modName: String? = when {
```
matklad: I'd rather leave the `modName` of the root crate anonymous and handle extern crates specially in `createLookupElement`. 

In Rust, root modules are anonymous, and it is for a good reason. There is really no good way to identify a particular crate. For example, you can have two crates with the same "names", but different versions. In reality, what happens is that rustc is invoked with a mapping "external_crate_name -> location of the external crate". That is, mapping of a crate name to a particular root module file is *internal* to the crate being compiled, there is no single shared global namespace of crates.

matklad:
I've just tried this and it works great!  I'll push it over the finish line myself then.

matklad:
Merged manually in https://github.com/intellij-rust/intellij-rust/commit/fb6477b31e72e151db691548afe9d3e8fc4015df.

Thanks!

PR #440, opened 2016-06-02
Improve SemicolonsFixer in smart enter action

himikof:
Smart enter action did bogus things in `fix_match_in_let.rs` test case, so I simplified/fixed the implementation and added some tests.

The new implementation uncovers an interesting question: should a smart enter action in Rust always insert a semicolon when used for the last expression in a block? I think that it should, and the previously existing tests already require this behavior.


matklad:
>I think that it should, and the previously existing tests already require this behavior.

I agree. If you add a new line after the last expression you probably want to type something after it, so it better be a statement then.

PR #442, opened 2016-06-03
Renamed all the PSI-elements' classes to have `Element` suffix

alexeykudinkin:
This is to draw a solid line between the PSI-related stuff and all the other structures.


matklad:
:clap:

PR #443, opened 2016-06-04
Revisited initial efforts in typing facilities

alexeykudinkin:
Separated types into two _unrelated_ hierarchies: `RustUnresolvedType`-based and (resolved) `RustType`-based.

Suggesting to follow next rule-of-thumb: unresolved-types' names should go in the form `RustUnresolved{Particular}Type` and resolved-ones in the form `Rust{Particular}Type`.

@matklad take a look please.


src/main/kotlin/org/rust/lang/core/type/TypeInference.kt
```diff
@@ -1,32 +1,47 @@
...
-    when (this) {
-        is RustPathExprElement -> {
-            val target = path.reference.resolve()
-            when (target) {
-                is RustSelfArgumentElement -> target.parentOfType<RustImplItemElement>()?.type?.resolvedType ?: RustUnknownType
-                else -> RustUnknownType
-            }
-        }
-        else -> RustUnknownType
+val RustExprElement.type: RustUnresolvedType by psiCached {
```
alexeykudinkin: @matklad nota bene, that every `psiCached` delegate should have _its_ _own_ `CachedValueProvider`.

matklad: Ouch, it was indeed a terrible bug. But I wonder if we can use some Kotlin features to still get the simple API for this one. Perhaps an inline function can help?

```
/**
 * Cached value invalidated on any PSI modification
 */
inline fun<E : PsiElement, T> psiCached(noinline block: E.() -> T): PsiCacheDelegate<E, T> {
    val provider = {element: E -> CachedValueProvider {
        CachedValueProvider.Result.create(element.block(), PsiModificationTracker.MODIFICATION_COUNT) }
    }
    return PsiCacheDelegate(provider)
}


class PsiCacheDelegate<E : PsiElement, T>(val provider: (E) -> CachedValueProvider<T>) {
    operator fun getValue(element: E, property: KProperty<*>): T {
        return CachedValuesManager.getCachedValue(element, provider(element))
    }
}
```

alexeykudinkin: Unfortunately, it wouldn't help, since inlining works on the instruction level, not the source one.

matklad:
I have two questions:

What would be the purpose of `UnresovledType`s? Can we just use PSI classes instead of them (this is what I had in mind originally: some PSI elements *represent* types, and can be cast into a real type by `resolvedType`)

What is exactly `ImplType`? There are no impl types in Rust, `impl` is just the thing that binds together some concrete type and some trait or a thing that specifies methods on the concrete type.

src/main/kotlin/org/rust/lang/core/type/TypeInference.kt
matklad: Perheps we don't need that helper at all? It does not save us from much of the boilerplate, but introduces complexity in the form of Kotlin's delegates.

alexeykudinkin:
> What is exactly ImplType? There are no impl types in Rust, impl is just the thing that binds together some concrete type and some trait or a thing that specifies methods on the concrete type.

`Rust*ImplType`s have been conceived as an union-types, though `RustImplType` by itself doesn't have much sense.

matklad:
>Rust*ImplTypes have been conceived as an union-types, though RustImplType by itself doesn't have much sense.

Can you elaborate a bit? How would for example, `RustTraitImplType` work in this situation?

```Rust
trait Foo {
    fn foo(&self)
}

trait Bar {
    fn bar(&self)
}

struct S;

impl Foo for S {
    fn foo(&self) {}
}

impl Bar for S {
    fn bar(&self) {
        self.fo<caret> // need to complete `foo` here
    }
}
```

I would say that the type of `self` should be just `StructType(S)`, and that `foo` and `bar` should be looked up as methods on the `S` type. 

Because in this PR `self` type is inferred as `TraitImplType(Bar)`, seems like only `bar` method is naturally available at the point:

```
            is RustTraitImplType -> recType.trait.traitBody.traitMethodMemberList.find { it.name == referenceName }
```

matklad:
LGTM, though I am a bit worried that UnresolvedType hierarchy would simply more or less copy Psi hierarchy without introducing much benefits.

alexeykudinkin:
Whops, it seems, i've assumed (w/o proper checking) that those `impl`s are isolated, which now looks as a bizarre assumption. 

You're correct then, we don't need any of those.

alexeykudinkin:
> LGTM, though I am a bit worried that UnresolvedType hierarchy would simply more or less copy Psi hierarchy without introducing much benefits.

That's right, the sole purpose of UTs is to strip down verbosity of PSI.

matklad:
>Whops, it seems, i've assumed (w/o proper checking) that those impls are isolated, which now looks as a bizarre assumption.

Yeah, I think the API for this thing should live on `RustStructType` or maybe `RustType`. Currently there is `inheritedImpls` property which shold collect `impl S {}` impls. Something like `val traitImpls` should be added to handle `impl Foo for S {}` syntax. This actually will be quite hard, because you can have all sorts of complex bounds there, like:

```
impl Foo for S {}

impl<T: Foo> Bar for T { // this also implements `Bar` for `S`, because `S: T`

}
```

matklad:
Just some more info. `impl S {}` is called an inherited impl, and the methods it defines are visible everywhere where the `S` type is visible.

That is, you can have

```Rust
pub mod foo {
    pub struct S;
}

pub mod bar { // this may be in another file
    impl foo::S {
        pub method(&self) {}
    }
}

fn main() {
    let s = S;
    s.method(); // method available via inherited impl
}
```

`impl Trait for S {}` form is a non inherited (trait) impl, and you must have `Trait` in scope for the methods to work.

```Rust
pub mod foo {
    pub struct S;
}

pub mod bar {
    pub trait T {
        pub method(&self);
    }
    impl T for foo::S {
        pub method(&self) {}
    }
}

fn main() {
    use bar::T;

    let s = S;
    s.method(); // will not compile if you comment out the `use`
}
```

So the algorithm to resolve a method call should be about as follows:

Determine receiver type. Collect all the inherited impls in the whole trait (we'll definitely need an index for this one). For each trait currently in scope determine if it is implemented for the receiver type. Among all inherited and trait impls, find a method with a given name. If there is exactly one match, we are done. If there are zero matches, report unresolved method (ideally, try to find this method in some non-imported trait and suggest auto import). If there are more then one method, report ambiguity (ideally, suggest to use UFCS (`qualified_path_expr` in our rust.bnf)).

Nice points are that we don't need to look at the types of arguments at all (no overloading) and that traits must be explicitly imported.

Hard point is that it may be difficult to determine if the trait is implemented for the given type.

PR #444, opened 2016-06-07
(RES): refactor RustResolveScope

matklad:
This PR tries to refactor `RustResolveScope` to make it more general and useful.

Currently, `RustResolveScope` allows to get a list of `RustNamedElements` for a scope. Unfortunately, the interface is not flexible enough. In particular, it can't handle

* elements, which need further resolve (`use foo;`, `mod foo;`, `extern crate foo;`). This causes a ton of casts in the `assign` function of RRE. 
* aliases (special case of the previous bullet)
* reexports (`use foo::*;` and prelude items)
* context sensitivity (proper scoping for local variables)
* laziness

So currently the scoping logic is dispersed across a number of `RustResolveScope` PSI elements, a bunch of RRE functions and a couple of extension methods.

The idea of this PR is to consolidate all this logic within `RustResolveScope`. So, the new `declarations` signature looks like this:

```
fun RustResolveScope.declarations(context: RustResolveScope.Context): Sequence<ScopeEntry> {
```

It allows to lazily retrieve **all** items in the scope which are visible at a particular place together with their actual names.

@alexeykudinkin what do you think about this approach?


alexeykudinkin:
First of all, let's get rid of all this _private_ extension-methods, they make code almost utterly unreadable due to the that you're constantly urged to get back and forth to get a sense what the thing you're looking at right now. And particularly, let's get rid of those used essentially ones.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
```diff
@@ -1,9 +1,256 @@
...
+import org.rust.cargo.util.preludeModule
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.RustFile
+import org.rust.lang.core.psi.impl.mixin.isStarImport
+import org.rust.lang.core.psi.impl.rustMod
+import org.rust.lang.core.psi.util.module
 
 interface RustResolveScope : RustCompositeElement {
-    val declarations: Collection<RustNamedElement>
+    data class Context(
```
alexeykudinkin: Both `Context` & `Entry` aren't inherent part of the `RustResolveScope`, and they essentially have nothing to do with it at all -- both of them are implementation details of our resolve procedure.

Let's move it out of `RustResolveScope` back to `RustResolveEngine`

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
```diff
@@ -1,9 +1,256 @@
...
+    }
+}
+
+fun RustResolveScope.declarations(context: RustResolveScope.Context): Sequence<RustResolveScope.Entry> {
+    val visitor = RustScopeVisitor(context)
+    accept(visitor)
+    return visitor.result
+}
+
+private class ScopeEntryImpl private constructor(
```
alexeykudinkin: Seems like we don't need to split this into interface & implementation actually.

alexeykudinkin:
BTW, i like the idea of lazyfying all the bulky operations, and i believe that we could extend that onto retrieval of `boundElements` too.

alexeykudinkin:
Right now, it seems semi-complete just because, most of the stuff is done in a non-lazy fashion.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
```diff
@@ -1,9 +1,256 @@
...
+        val visitedStarImports: Set<RustUseItemElement> = emptySet()
+    )
+
+    interface Entry {
+        val name: String
+        val element: RustNamedElement?
+    }
+}
+
+fun RustResolveScope.declarations(context: RustResolveScope.Context): Sequence<RustResolveScope.Entry> {
```
alexeykudinkin: This extension point seems more like implementation detail, that should not be exposed outwards.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
```diff
@@ -1,9 +1,258 @@
...
+ *        let x = x;  // not visible
+ *                ^ element
+ *        let x = 62; // not visible
+ *    }
+ *    ```
+ */
+private fun RustBlockElement.letDeclarationsVisibleAt(element: RustCompositeElement): Sequence<RustLetDeclElement> =
+    letDeclarations
+        .dropWhile { PsiUtilCore.compareElementsByPosition(element, it) < 0 }
+        // Drops at most one element
```
himikof: Is this comment true even in the following case? 
```rust
fn main() {
    let x = 92;
    let x = { let x = <caret>x; 42};
}
```

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
matklad: O_o, this is THE api of this whole thing...

matklad: But a doc string is definitely missing!

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
matklad: This split allows to make construction of `ScopeEntries` private to this file. If we move `ScopeEntry` to the `RRE` as per you suggestion then this split will indeed become redundant as we'd have to expose constructors anyway.

matklad:
>has all essential things lazified

Excpt the initial visitor's visit, fixed in f09667e4dcc653f788f69188fd6dfad2a4b85e1a

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
matklad: Yeah. In the inner block there is only one `let`, and it will be dropped. In the outer block, `let x = 92` won't be dropped. This function looks only at one block at a time, and for both blocks it drops exactly one `let.`

However your example indeed demonstrates a pathological behavior: for a series of `n` nested lets, the total cost of `isAncestor` checks is `O(n^2)`. I hope this won't create any troubles because even two level deep nesting of `let`s is pretty rare.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
matklad: `Entry` is definitely an integral part of the `ResolveScope`.  `ResolveScope` is the thing which contains `Entries`, and `Entries` are more then just `NamedElements`, because they include aliasing info and provide laziness. 

As for the `Context`, it should be simply made `private` to this file (done). 

Please note that this implementation of `ResolveScope` knows nothing about `RRE` which is a nice property which I'd love to preserve.

matklad:
>First of all, let's get rid of all this private extension-methods, they make code almost utterly unreadable due to the that you're constantly urged to get back and forth to get a sense what the thing you're looking at right now. And particularly, let's get rid of those used essentially once.

Hm, I generally love to introduce a gazillion of small helper methods to keep my functions at most several lines long. I try to extract every piece of behavior that can be reasonably named into a separate method, and I don't mind navigating between usage and definition at all. 

But I've failed to chose the illuminating names this time, so the net effect is indeed bad :) Fixed in https://github.com/intellij-rust/intellij-rust/pull/444/commits/772b0b97250dbe56f4de5eec23fc577471640d6f, but some extensions are still necessary to avoid code duplication.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
alexeykudinkin: > Please note that this implementation of ResolveScope knows nothing about RRE which is a nice property which I'd love to preserve.

Pros?

> Entry is definitely an integral part of the ResolveScope. ResolveScope is the thing which contains Entries, and Entries are more then just NamedElements, because they include aliasing info and provide laziness. 

No, both of those two are integral parts of the _resolving_ procedure (and not the marking interface `RustResolveScope`) which should be encapsulated (or at least confined) by the single 'module' -- this module is historically `RustResolveEngine`.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
alexeykudinkin: No, all its usages are confined (and should be like this actually) inside the resolving-engine.

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
matklad: >No, both of those two are integral parts of the resolving procedure (and not the marking interface RustResolveScope) which should be encapsulated (or at least confined) by the single 'module' -- this module is historically RustResolveEngine.

Yes, this is part of the resolve, and that's why it's in the org.rust.lang.core.**resolve**.scope package. `RustResolveEngine` does a bunch of other stuff, besides scopes, so by extracting scoping rules into a 'submodule' we can separate concerns. And this is possible because the dependency between scopes and RRE is unidirectional.

Note that another consumer of scopes API is the [completion engine], and this PR actually fixes #405 because the same scoping logic is now used for both resolve and completion. Hm, interesting, it also fixes completion for the `*`-imported items, will need to add a test for it. 

[completion engine]: https://github.com/intellij-rust/intellij-rust/pull/444/files#diff-f6a8fb5e98ccbac0c27f3afcc887a52eR43

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
matklad: >No, all its usages are confined (and should be like this actually) inside the resolving-engine.

It is also used by completion engine. I expand on this a bit in [another comment].

[another comment]: https://github.com/intellij-rust/intellij-rust/pull/444#discussion_r66910366

src/main/kotlin/org/rust/lang/core/resolve/scope/RustResolveScope.kt
alexeykudinkin: > Yes, this is part of the resolve, and that's why it's in the org.rust.lang.core.resolve.scope package. RustResolveEngine does a bunch of other stuff, besides scopes, so by extracting scoping rules into a 'submodule' we can separate concerns. And this is possible because the dependency between scopes and RRE is unidirectional.

Primary goal of resolve engine -- is dealing with scoping rules. All the stuff related to `Entry`s is an _**implementation** detail_ and should not be exposed beyound the `RRE`.

Having the property of both completion & resolving engine using the same scoping infra is a nice thing, though it should be achived the other way -- completion engine should reference resolve-engine.

matklad: Ok, will move it inside the RRE, but the `Entry` will still be exposed: during completion we need to know both the name to complete, and the actual element the name is bound to.

alexeykudinkin: Why?

matklad:
```Rust
use foo::bar as quux;

fn main() {
    q<caret> // here we offer completion to `quux`, but in the completion popup we must show
             // the information (icon, type, parameter names, etc.) about the `bar`
}

```

matklad: And I think that the entry is the right place to grow things like visibility and priority, which we will need both in the completion and in the resolve.

alexeykudinkin: I know the use-case, i'm asking about why do we need the `Entry` instead of checking whether we're listing an aliased decl.

alexeykudinkin: > And I think that the entry is the right place to grow things like visibility and priority, which we will need both in the completion and in the resolve.
No, it isn't, since it can be directly derived from PSI.

matklad: >Entry instead of checking whether we're listing an aliased decl.

Checking for aliases specifically feels smelly. Just look at the [assign](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L314-L337) method: all this casts looks ugly to me because they try restore the information which was simply forgotten when we were processing scopes.

Not sure that visibility is derivable from PSI: you'll need to know the whole path which was used to resolve the item. 

The same thing with priority: you need to know not only the item itself, but the way which was used for importing: was it a start import or perhaps a direct import.

alexeykudinkin: > Checking for aliases specifically feels smelly. Just look at the assign method: all this casts looks ugly to me because they try restore the information which was simply forgotten when we were processing scopes.
Not sure that visibility is derivable from PSI: you'll need to know the whole path which was used to resolve the item.
The same thing with priority: you need to know not only the item itself, but the way which was used for importing: was it a start import or perhaps a direct import.

Then we'd split the resolve scope visitor, or will make it pluggable with callbacks. What we are not going to do is a converting `Entry` into a magic bin storing all the stuff usable somewhere else. If we need to separate concerns we'd separate at that level: whoever need some info, should get just it and not all the immediately available stuff.

matklad: Ok, will make entry private

matklad: fixed in https://github.com/intellij-rust/intellij-rust/pull/444/commits/72101c2f6e95555daec9c7b202e4c5982db12e65

PR #447, opened 2016-06-12
(GRAM): Implement simple block-level recovery

himikof:
This should always recover from errors in blocks if braces structure in file is intact.
Unfortunately, it will do bogus things if there are missing closing braces.

I have doubts about usefulness of the partial parsing test `fn.rs`. It specifically tests for missing closing braces, and I think it should not be done for two reasons:

1. I don't think it is possible to recover from missing closing brace, because items from the parent scope can appear in the scope of the current block just fine. So the closing brace is the only chance to see where the block ends.
2. During typing, matched braces are inserted automatically. So it is hard to mismatch them while typing, which should be the most common case for partial parsing to test for.


matklad:
Yeah, `fn.rs` should be revised, will do it. Thanks a lot for this change, it seems simple, but should greatly improve the user experience!

matklad:
https://github.com/intellij-rust/intellij-rust/commit/73f6eeb169f7efaca20e28f46cc24a1c2eeb1a30

PR #452, opened 2016-06-13
(FMT): fix #451

### LINKED ISSUE #451: Properly indent known macro calls
```Rust
fn foo() {
    write!(&mut self.destination, "{}",
           magic_number);
}
```

gets reformatted to

```Rust
fn foo() {
    write!(&mut self.destination, "{}",
    magic_number);
}
```

And it really should not. It is important that `write!` is a known format-like macro. Unknown macros are properly left intact. @jajakobyly you might want to look into this :)

jajakobyly:


src/test/kotlin/org/rust/ide/formatter/RustFormatterTestCase.kt
```diff
@@ -33,4 +33,6 @@ class RustFormatterTestCase : FormatterTestCase() {
     // FIXME: these two guys are way too big
     fun testSpacing() = doTest()
     fun testLineBreaks() = doTest()
+
+    fun testIssue451() = doTest()   // https://github.com/intellij-rust/intellij-rust/issues/451
```
matklad: Pretty scary name for the issue, I like it! :)

jajakobyly:
I'm not sure whether the behaviour for `write!{...}` is correct. rustfmt seems to not touch it

matklad:
I think write should be handled the same way as a normal function call (+ some handling of keyword arguments)

matklad:
Thanks!

matklad:
FYI: rustraytracer now is fully reformatted with `ctrl+alt+l` :)

PR #456, opened 2016-06-13
(FMT): Moar options

jajakobyly:
All alignment can be now turned off

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/settings/RustLanguageCodeStyleSettingsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/settings/RustLanguageCodeStyleSettingsProvider.kt
@@ -35,7 +35,9 @@ class RustLanguageCodeStyleSettingsProvider : LanguageCodeStyleSettingsProvider(
             WRAPPING_AND_BRACES_SETTINGS ->
                 consumer.showStandardOptions(
                     "RIGHT_MARGIN",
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/ide/formatter/settings/RustLanguageCodeStyleSettingsProvider.kt
```diff
@@ -32,18 +32,27 @@ class RustLanguageCodeStyleSettingsProvider : LanguageCodeStyleSettingsProvider(
...
+            }
         }
     }
 
     override fun getIndentOptionsEditor(): IndentOptionsEditor? = SmartIndentOptionsEditor()
 
     override fun getDefaultCommonSettings(): CommonCodeStyleSettings =
         CommonCodeStyleSettings(language).apply {
-            RIGHT_MARGIN = 99
+            RIGHT_MARGIN = 100
```
jajakobyly: This is _de facto_ standard in `rustc` and default value in `rustfmt`

matklad: Yep, looks like it. [Styleguide] says 99, but I think it is wrong. It makes sense to use 79 instead of 80 to save one column for the terminal text editor info, but this argument does not work for 99 vs 100

[Styleguide]: https://aturon.github.io/style/whitespace.html

PR #457, opened 2016-06-13
(MISC): Impl blocks without traits are inherent, not inherited

himikof:
For example, see [syntax index](https://doc.rust-lang.org/book/syntax-index.html) for `impl` keyword.


matklad:
Indeed they are, thanks!

PR #465, opened 2016-06-16
(IDE): fix type info hints provider

himikof:
With this change, Ctrl+Shift+P works nicely for me. Unfortunately, I'm not sure how to write a test.

Collecting parent expressions until an item boundary makes more sense than doing the opposite.
`escapeXml` call is needed due to `getInformationHint` result being interpreted as HTML, so `<unknown>` now displays properly.


matklad:
I think you should be able to test this with something like

```
    fun testFoo() {
        myFixture.configureByText("main.rs", """
        struct Quux;
        fn main() {
            let q = Quux;
            let _ = <caret>q;
        }
        """)

        val p = myFixture.testAction(ShowExpressionTypeAction())
        println(p)
    }
```

But it currently throws NPE somewhere near UI.

PR #469, opened 2016-06-18
(COM): Add derivable trait completion

bgourlie:
This PR adds autocompletion for derivable traits.  It's pretty naive right now, only omitting suggestions for traits that have already been specified in the `derive` attribute.  Eventually, I would like it to omit suggestions for traits that can't be derived (not suggesting `Copy` for a struct that has non-Copy fields, for example), and traits that have already been manually implemented.

### DIFF
--- a/src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
@@ -1,0 +1,63 @@
+package org.rust.ide.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionProvider
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.patterns.ElementPattern
+import com.intellij.patterns.PlatformPatterns.psiElement
+import com.intellij.psi.PsiElement
+import com.intellij.util.ProcessingContext
+import org.rust.lang.RustLanguage
+import org.rust.lang.core.parser.RustPsiTreeUtil
+import org.rust.lang.core.psi.RustCompositeElementTypes
+import org.rust.lang.core.psi.RustMetaItemElement
+import org.rust.lang.core.psi.RustTokenElementTypes
+import org.rust.lang.core.resolve.util.RustResolveUtil
+
+class DeriveCompletionProvider : CompletionProvider<CompletionParameters>() {
+
+    private val DERIVABLE_TRAITS = listOf("Eq", "PartialEq", "Ord", "PartialOrd", "Copy", "Clone", "Hash", "Default",
+        "Debug")
+
+    override fun addCompletions(parameters: CompletionParameters,
+                                context: ProcessingContext?,
+                                result: CompletionResultSet) {
+
+        val outerAttrElem = RustResolveUtil.getResolveScopeFor(parameters.position)?.firstChild
+        val deriveMetaItem = RustPsiTreeUtil.getChildOfType(outerAttrElem, RustMetaItemElement::class.java)
+        val alreadyDerived = RustPsiTreeUtil.getChildrenOfType(deriveMetaItem, RustMetaItemElement::class.java)
+            ?.mapNotNull { it.firstChild.text }.orEmpty()
+        val lookupElements = DERIVABLE_TRAITS.filter { !alreadyDerived.contains(it) }
+            .map { LookupElementBuilder.create(it) }
+        result.addAllElements(lookupElements)
+    }
+
+    companion object ElementPatternFactory {
+        val elementPattern: ElementPattern<PsiElement> get() {
+
+            val deriveIdentifier = psiElement()
+                .withElementType(RustTokenElementTypes.IDENTIFIER)
+                .withText("derive")
+
+            val outerAttr = psiElement()
+                .withElementType(RustCompositeElementTypes.OUTER_ATTR)
+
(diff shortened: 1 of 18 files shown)


src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.util.ProcessingContext
+import org.rust.lang.RustLanguage
+import org.rust.lang.core.parser.RustPsiTreeUtil
+import org.rust.lang.core.psi.RustCompositeElementTypes
+import org.rust.lang.core.psi.RustMetaItemElement
+import org.rust.lang.core.psi.RustTokenElementTypes
+import org.rust.lang.core.resolve.util.RustResolveUtil
+
+class DeriveCompletionProvider : CompletionProvider<CompletionParameters>() {
```
matklad: This can be `object DeriveCompletionProvider`

src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    private val DERIVABLE_TRAITS = listOf("Eq", "PartialEq", "Ord", "PartialOrd", "Copy", "Clone", "Hash", "Default",
+        "Debug")
+
+    override fun addCompletions(parameters: CompletionParameters,
+                                context: ProcessingContext?,
+                                result: CompletionResultSet) {
+
+        val outerAttrElem = RustResolveUtil.getResolveScopeFor(parameters.position)?.firstChild
```
matklad: Hm, can this be simplified to `parametes.position.parentOfType<RustOuterAttrElement>`?

src/test/kotlin/org/rust/completion/RustDeriveCompletionTest.kt
```diff
@@ -0,0 +1,43 @@
+package org.rust.completion
+
+import com.intellij.codeInsight.completion.CompletionType
+import org.rust.lang.RustTestCaseBase
+
+class RustDeriveCompletionTest : RustTestCaseBase() {
```
matklad: You should be able to inherit from `RustCompletionTestBase` and just write

```
fun testCompleteOnStruct = checkSoleCompletion()
```

Or is this not working because there several variants of completion? In any case it's better to remove this duplication in tests functions. See `checkSoloCompletion` for how it can be done (basically, you can get the name of the test file from the name of the test function).

src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+        val lookupElements = DERIVABLE_TRAITS.filter { !alreadyDerived.contains(it) }
+            .map { LookupElementBuilder.create(it) }
+        result.addAllElements(lookupElements)
+    }
+
+    companion object ElementPatternFactory {
+        val elementPattern: ElementPattern<PsiElement> get() {
+
+            val deriveIdentifier = psiElement()
+                .withElementType(RustTokenElementTypes.IDENTIFIER)
```
matklad: There is `psiElement` pattern constructor which accepts `ElementType` as a parameter:

```Java
  public static PsiElementPattern.Capture<PsiElement> psiElement(IElementType type) {
```

src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    override fun addCompletions(parameters: CompletionParameters,
+                                context: ProcessingContext?,
+                                result: CompletionResultSet) {
+
+        val outerAttrElem = RustResolveUtil.getResolveScopeFor(parameters.position)?.firstChild
+        val deriveMetaItem = RustPsiTreeUtil.getChildOfType(outerAttrElem, RustMetaItemElement::class.java)
+        val alreadyDerived = RustPsiTreeUtil.getChildrenOfType(deriveMetaItem, RustMetaItemElement::class.java)
+            ?.mapNotNull { it.firstChild.text }.orEmpty()
+        val lookupElements = DERIVABLE_TRAITS.filter { !alreadyDerived.contains(it) }
```
matklad: `contains` can be spelled as `in` in Kotlin: `{ it !in alreadyDerived}`

src/main/kotlin/org/rust/ide/completion/DeriveCompletionProvider.kt
bgourlie: If I do this, then I can't declare `companion object ElementPatternFactory`.  This is my first time writing Kotlin, so I don't know if using a companion object here is appropriate anyway, but it seemed to work well for what I was doing.

matklad: Yes, you can't declare a companion object of an object, but you don't need one :) This should to the trick: 

```Kotlin
object DeriveCompletionProvider : CompletionProvider<CompletionParameters>() {
    private val DERIVABLE_TRAITS = ...
    override fun addCompletions(
        parameters: CompletionParameters,
        context: ProcessingContext?,
        result: CompletionResultSet
    ) {
        ...
    }

    val elementPattern: ElementPattern<PsiElement> get() {
        ...
    }
}
```

jajakobyly: Remove `companion` from declaration and it'll work

matklad:
Wonderful, thanks!

PR #470, opened 2016-06-18
(FMT): Code samples

jajakobyly:
This PR adds better code samples to the code style settings page. Also I have tinkered around type params formatting.


src/main/resources/org/rust/ide/formatter/settings/blank_lines_sample.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        }
+
+
+
+
+        // Close the opened bracket and return a fmt::Result value
+        write!(f, "]")
+    }
+}
```
matklad: Why are these sample are misformated so drastically? 

I think we should leave one instance with too many lines, and one with too few. That way, the user can easily identify what exactly is changed when the option is toggled.

src/main/kotlin/org/rust/ide/formatter/impl/utils.kt
```diff
@@ -48,24 +47,22 @@ val MACRO_ARGS = ts(MACRO_ARG, FORMAT_MACRO_ARGS, TRY_MACRO_ARGS)
 
 val FN_DECLS = ts(FN_ITEM, FOREIGN_FN_DECL, TRAIT_METHOD_MEMBER, IMPL_METHOD_MEMBER, BARE_FN_TYPE, LAMBDA_EXPR)
 
-val FN_SHARED_ALIGN_OWNERS = orSet(PARAMS_LIKE, ts(RET_TYPE, WHERE_CLAUSE))
-
 
 val PsiElement.isTopLevelItem: Boolean
-    get() = (this is RustItemElement || this is RustAttrElement) && this.parent is RustFile
+    get() = (this is RustItemElement || this is RustAttrElement) && parent is RustMod
```
matklad: Not related to this PR, but the second case `this is RustAttrElement` seems suspicious. The parent of `RustAttr` is just another item. Perhaps this should be `RustInnerAttr`, to handle mod inner attributes?

src/test/kotlin/org/rust/ide/formatter/RustFormatterTestCase.kt
```diff
@@ -52,6 +52,11 @@ class RustFormatterTestCase : FormatterTestCase() {
         doTest()
     }
 
+    fun testAlignTypeParamsOn() {
+        custom().ALIGN_TYPE_PARAMS = true
+        doTest()
+    }
```
matklad: Again, not related to PR, but a more DSLy API would be

```Kotlin
   fun testAlignTypeParamsOn() = custom {
       ALIGN_TYPE_PARAMS = true
   }
```

where `custom` is defined as

```Kotlin
fun custom(block: RustCustomSettings.() -> Unit) {
    val settings = settings.getCustomSettings(RustCodeStyleSettings::class.java)
    settings.block() // <- this is pure awesomeness, because the extension function here is a parameter. 
    doTest()
}
```

src/main/kotlin/org/rust/ide/formatter/impl/utils.kt
jajakobyly: Yeah, I haven't noticed this. It was correct when parent could be only a file. Fix it here?

matklad: I'd rather make a separate PR, as usual, but if you find it more convenient to fix it here go ahead.

jajakobyly: I can submit separate PR with further small improvements again in few days, it's no problem for me.

PR #472, opened 2016-06-20
(HIGHLIGHTER): re-highlight literals in attributes

jajakobyly:
Before:
idea_2016-06-18_21-59-07

After:
java_2016-06-20_16-10-15

Small tweak but makes attributes more appealing to me :smile:

### DIFF
--- a/src/test/resources/org/rust/ide/annotator/fixtures/attributes.rs
+++ b/src/test/resources/org/rust/ide/annotator/fixtures/attributes.rs
@@ -1,4 +1,4 @@
 <info>#[cfg_attr(foo)]</info>
 fn <info>main</info>() {
-    <info>#![crate_type = "lib"]</info>
+    <info>#![crate_type = <info>"lib"</info>]</info>
 }
--- a/src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
@@ -2,16 +2,27 @@ package org.rust.ide.annotator
 
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.Annotator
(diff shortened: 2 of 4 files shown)


matklad:
Nice! Travis is sad.

jajakobyly:
Yeah, forgot about running tests in order to check if this feature is tested :sweat_smile:

src/test/resources/org/rust/ide/annotator/fixtures/attributes.rs
```diff
@@ -1,4 +1,4 @@
 <info>#[cfg_attr(foo)]</info>
 fn <info>main</info>() {
-    <info>#![crate_type = "lib"]</info>
+    <info>#![crate_type = <info>"lib"</info>]</info>
```
matklad: open and closing order is reversed?

jajakobyly: I don't know honestly, but I believe nested tags are supported; otherwise test should fail

matklad: In this test they are nested, but in the previous one they are disjoint. Looks like at least one case is wrong?

matklad: Nevermind, that was not a test

jajakobyly: We have such annotation layout:
```
|--------------|
         |--|
```

While if nested tags weren't supported, test framework should (as far as I understand) expect this:
```
|---------|
       |----------|
```

jajakobyly: > In this test they are nested, but in the previous one they are disjoint. Looks like at least one case is wrong?

It was colour scheme sample, and there for nesting I would have to declare custom tag and blablabla. Not worth it ;)

matklad:
Thanks!

PR #476, opened 2016-06-20
Fix #475

### LINKED ISSUE #475: Indentation: within match block when typing `(`
It's better to show on example:
```rust
fn foobar(n: u64) -> String {
    match (n % 3, n % 5) {
        // cursor is on this level of indentation (it's correct).
    }
}
```
Then I typed `(` and got this:
```rust
fn foobar(n: u64) -> String {
    match (n % 3, n % 5) {
    () // cursor was moved one level to the left.
    }
}
```
-
```
IntelliJ IDEA 2016.1.3
Build #IC-145.1617, built on June 3, 2016
JRE: 1.8.0_76-release-b198 x86_64
```
```
Plugin version: 0.0.1.675
```

jajakobyly:
~~IntelliJ does not interact with formatter code here, so there's probably some other way to control typing behaviour. I'll take a deeper look into this after finals, sorry~~

Nevermind, it went out to be a nasty parser recovery issue.


src/test/kotlin/org/rust/ide/formatter/RustAutoIndentTestCase.kt
```diff
@@ -21,7 +21,7 @@ class RustAutoIndentTestCase : RustTestCaseBase() {
     fun testExpr2() = doTest()
     fun testExpr3() = doTest()
 
-    private fun doTest() = checkByFile {
-        myFixture.type('\n')
+    private fun doTest(c: Char = '\n') = checkByFile {
```
matklad: Is it used anywher?

jajakobyly: Nope, but I thought it might be handy in the future

matklad: I think that adding an indent test in addition to the parser test is a good idea. I'd even say that this test would be more useful then the `PatialParse` one, but this is subjective (https://github.com/intellij-rust/intellij-rust/commit/16230202c2c8adcaeaf6de0aef8a6fae06515d03#commitcomment-17966181 for some opinions on this)

matklad:
Thanks!

PR #478, opened 2016-06-21
(DOC): Fix find symbol shortcut

friedroman:
I believe this is the default shortcut for symbol search on windows.


matklad:
Yes,  indeed, `Ctrl+Shift+N` is "find file", thanks!

PR #483, opened 2016-06-22
Min blank lines between items

matklad:
@jajakobyly please review. This adds an option to control the minimum number of blank lines between large items, like `structs` or `enums`. 1 is chosen as a reasonable default value, you can use 2 to get a PEP8 style. 

It currently does not pass the tests because of the comments. I think that this should be fixed by changing the comment binder on items to bind all preceding comments, and not only the doc ones.


src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
```diff
@@ -268,6 +272,11 @@ private fun SpacingContext.blockMustBeMultiLine(): Boolean {
     }
 }
 
+private fun SpacingContext.needsBlankLineBetweenItems(): Boolean {
+    // Allow to keep consecutive runs of `use`, `const` or `static` items without blank lines
+    return !(elementType1 == elementType2 && elementType1 in ONE_LINE_ITEMS)
```
jajakobyly: Have you tried `ncPsi1.elementType` instead of `elementType1` and similarly for 2 ? I believe this should do the work. `nc` prefix comes from `non-comment`, and they are determined by `SpacingContext#omitCommentBlocks`. Though I must admit that comment handling is pretty poor now.

jajakobyly:
There is no entry in code style settings pane :smile:

src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
matklad: Hm, `ncPsi` didn't help me, but just pluggin `if either one is comment return false` seems to do the job. 

Still, I think the right approach here is to properly use comment binders, so, for example

```
mod foo;

//this comment is a part of the PSI node for bar, 
// so no need for special handling in formatter
mod bar        
```

I've tried to use simple gready left binder, but it breaks in the following case

```
mod foo; // this should be attached to foo, but is attached to `bar` with gready binder

mod bar; 
```

jajakobyly: I believe comment binders approach won't work, here is counterexample:
```rust
mod foo;
//! Where is your God now?
mod bar;
```

This guy has to be unbound comment element. We need to handle comments in formatter. As far as I remember PyCharm does this (and the `ncPsi` approach is inspired by it) but it's quite hard reference to grasp into imho because it's formatter looks like spaghetti code for me in some places :(

jajakobyly: Hmm though I think that this counterexample might not be counterexample at all. How would you like to format this thing?

```rust
mod foo;
//! Where is your God now?
mod bar;
```

or

```rust
mod foo;

//! Where is your God now?

mod bar;
```

matklad: > How would you like to format this thing?

I believe this is a syntax error in rustc =) Inner doc comments and inner attributes are valid only at the beginning of the module/block. Although we don't enforce this yet.

jajakobyly: :sweat_smile:

matklad: Here is how Kotlin attaches comments:

```
class Foo

//bar
/*bar */ class Bar //bar
//self standing comment

//baz
class Baz
```

Looks like in the end we would wont *both* comment binders and special comment handling =/

matklad:
@jajakobyly this should be good to go now

jajakobyly:
:+1: You might want to change commit name ;)

matklad:
>You might want to change commit name ;)

Ah, indeed, thanks!

PR #486, opened 2016-06-24
[RDY] Add context-aware attribute completion

bgourlie:
Ready to merge, pending any feedback.

Remaining work:

- [x] Tests
- [x] Implement onStaticMut pattern
- [x] Don't suggest attributes that have already been defined on an item


src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+        val metaItemElem = psiElement(RustCompositeElementTypes.META_ITEM)
+            .andOr(psiElement().withParent(outerAttrElem), psiElement().withParent(innerAttrElem))
+        return psiElement().withParent(metaItemElem).withLanguage(RustLanguage);
+    }
+
+    private data class RustAttribute(val name: String, val appliesTo: ElementPattern<PsiElement>)
+
+    private val onStruct: ElementPattern<PsiElement> get() {
+        return psiElement().withSuperParent(3, RustStructItemElement::class.java)
+    }
```
matklad: I think you can avoid a getter here

```
val onStruct: ElementPattern<PsiElement> = psiElement().withSuperParent(3, RustStructItemElement::class.java)
```

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    private val onEnum: ElementPattern<PsiElement> get() {
+        return psiElement().withSuperParent(3, RustEnumItemElement::class.java)
```
matklad: Let's abstract away this `withSuperParent(3` thing. There is a high probability that PSI structure for attributes changes, and so this constant would had to be changed everywhere. I would suggest something like

```
inline fun <reified I: RustItemElement> itemAttribute: ElementPattern<PsiElement>
```

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+        return psiElement().withSuperParent(3, RustMacroItemElement::class.java)
+    }
+
+    private val onTupleStruct: ElementPattern<PsiElement> get() {
+        return psiElement().withSuperParent(2, psiElement().withChild(psiElement(RustStructTupleArgsElement::class.java)))
+    }
+
+    private val onCrate: ElementPattern<PsiElement> get() {
+        val inMain = virtualFile().withName("main.rs")
+        val inLib = virtualFile().withName("lib.rs")
```
matklad: This is a nice approximation of being a crate concept, but the precise information about crate root file is available via `CargoProjectDescription` class.

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+        RustAttribute("rustc_on_unimplemented", onTrait), // Feature gated as of 1.9
+        RustAttribute("cfg", onFn),
+        RustAttribute("cfg_attr", onAnyItem),
+        RustAttribute("allow", onAnyItem),
+        RustAttribute("warn", onAnyItem),
+        RustAttribute("forbid", onAnyItem),
+        RustAttribute("deny", onAnyItem),
+        RustAttribute("lang", onFn),
+        RustAttribute("inline", onFn),
+        RustAttribute("derive", onStruct or onEnum))
```
matklad: Neat `or`!

matklad:
Looks nice! But we definitely need tests for this, because this psi patterns are really tied to a particular PSI structure, and this binding is not enforced by the compiler. And at the moment PSI structure changes a lot.

bgourlie:
Yep, tests are on my todo list.  If we were to write tests for each individual attribute, exhaustively asserting `checkSoleCompletion` and `checkNoCompletion` on items where the attribute should/should not complete, we're talking hundreds of tests.  I'm wondering if there's a nice way to expose the element patterns (`onFn`, `onStruct`, etc) and test those instead.  Thoughts?

matklad:
Year, I testing the patterns themselves sounds like the right approach. Take a look at these two classes:

https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/lang/core/type/RustTypificationTestBase.kt
https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/lang/core/type/RustTypeResolvingTest.kt

They help to write "inline Rust code" style tests, which I think is better then fixtures for the task at hand. 

Note this `//language=Rust` thing, it will allow IntelliJ Idea to highlight the string literal as Rust code. Perhapse you'll need to use `//language=RUST` (note the case) instead: we are in the process of renaming this thing.

matklad:
Also, I think it's important to cover all the patterns, but it does not make much sense to cover all the completion variants.

matklad:
>You could rebase onto upstream master and this problem shouldn't occur

We have not decided yet which one, "RUST" or "Rust" will stay (the new one casues some troubles at plugins.jetbrains.com), so either one is probably incorrect :)

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
bgourlie: Good call.  I'm not sure what you're suggesting with the inline function though, which may be due the fact that I'm relatively new to Kotlin.

Would a standard function accomplish the same thing?

    private fun onItem(aClass: Class<out PsiElement>): ElementPattern<PsiElement> {
        return psiElement().withSuperParent(3, aClass)
    }

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
bgourlie: How do I get ahold of an instance of this class?

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
bgourlie: Thanks :) Being able to define arbitrary infix functions is pretty neat.  I'm really enjoying Kotlin.

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
matklad: Yep, but the call side would look a bit less idiomatic: 

```
itemAttribute<RustStructItem>()
```

vs

```
onItem(RustStructItem::class.java)
```

The `inline` and `reified` keywords allow you to use `I::class.java` inside the function. You can't do this with usual functions, because generics are erased and that `I` is just the `Object` (modulo bounds and variance). But in Kotlin, because the function is inline, you can freely use `T::class` (or `as T` casts). Here are the doc:
https://kotlinlang.org/docs/reference/inline-functions.html#reified-type-parameters

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
matklad: `psiElement.module.cargoProject`. `module` and `cargoProject` are extension functions.

bgourlie: We don't have an instance of the psiElement to call the extension methods on.  `psiElement()` returns an instance of `PsiElementPattern.Capture<PsiElement>`.

matklad: Yeah, you need to lookup the project on the element to which the pattern is applied. The pattern is created only once, but can be reused across several projects. There should be some API which accepts a predicate. I am also Ok with leaving it as is, and addressing the issue in the separate PR.

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
matklad: Hm, perhaps it makes sense to group attributes by pattern?

```Kotlin
val attributes = mapOf(
    onCrate to "crate_name crate_type feature ..."
    onFn to "lang linline ..."
).flatMap { entry ->
    entry.value.split().map { attrName -> RustAttribute(attrName, entry.key) }
}
```

bgourlie: 👍 I like this

src/test/kotlin/org/rust/lang/core/pattern/RustPatternTestBase.kt
```diff
@@ -0,0 +1,13 @@
+package org.rust.lang.core.pattern
+
+import com.intellij.patterns.ElementPattern
+import org.rust.lang.RustTestCaseBase
+
+abstract class RustPatternTestBase : RustTestCaseBase() {
+    override val dataPath: String get() = ""
+
+    protected fun <T>testPattern(code: String, pattern: ElementPattern<T>) {
```
matklad: With the latest Kotlin plugin (update via `Tools > Kotlin > Configure Kotlin plugin updates`) you should be able to get rid of `//language=Rust` comments and instead use an annotation:

```
protected fun <T>testPattern(@Language("Rust") code: String, pattern: ElementPattern<T>) {
```

See https://github.com/intellij-rust/intellij-rust/blob/069cd1d05c3f567d32e40010c80a4a2dee7ecc83/src/test/kotlin/org/rust/ide/search/RustFindUsagesTest.kt for an example.

src/main/kotlin/org/rust/ide/completion/AttributeCompletionProvider.kt
```diff
@@ -0,0 +1,129 @@
...
+    val onTupleStruct: PsiElementPattern.Capture<PsiElement> = psiElement()
+        .withSuperParent(3, psiElement().withChild(psiElement<RustStructTupleArgsElement>()))
+
+    val onCrate: PsiElementPattern.Capture<PsiElement> = psiElement().withSuperParent<PsiFile>(3).with(
+        object : PatternCondition<PsiElement>("onCrateCondition") {
+            override fun accepts(t: PsiElement, context: ProcessingContext?): Boolean {
+                val file = t.containingFile.originalFile.virtualFile
+                val crateSourceRoot = "${t.module?.cargoProjectRoot ?: "temp://"}/src"
+                val crateFile = listOf("$crateSourceRoot/lib.rs", "$crateSourceRoot/main.rs")
+                return crateFile.contains(file.toString())
```
matklad: Hm, looks like we already have a function to check if the file is a crate root, [here it is](isCrateRoot)! It should be ok to cast `PsiFile` to `RustFile` here. 

[isCrateRoot]: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/impl/RustFile.kt#L43

matklad:
Awesome, thanks!

PR #491, opened 2016-06-25
Documentation comments highlighting

jajakobyly:
This PR breathes some life into doc comments!

I use [CommonMark](http://spec.commonmark.org/0.25/) as a kind of Markdown spec, but I don't fully follow it. Unfortunately this code is kinda spaghetti, but I have doubts whether it is possible to organise this in a more readable way.

This is how it looks like:


matklad:
Hm, this looks cool, but the amount of code to achieve this is disturbing... 


What exactly is the problem that we are trying to solve here? Is it worth solving? Perhaps we can reformulate it in such a way that we don't need to create yet another not exactly spec conforming markdown parser?

jajakobyly:
Well, some complexity comes from the fact that I avoided to create new token types for doc comments (from `RustLexer` perspective) - I had to "fork" `MergingLexerAdapterBase`, because IntelliJ's one doesn't provide access to the token which fires merging operation.

As for Markdown itself, I don't think so. We need to deal with these decorations here and are only interested in a very small subset of Markdown. I considered borrowing some code from intellij-markdown, but a) handling these `///` etc. fences would be very, very complicated and b) we don't need whole table, lists, images etc. stuff. I have checked out how Kotlin handles this, and their approach is nearly identical.

Also most from these 1,125+23 lines are actually highlighter tests (about 740)

matklad:
But does this in-code highlighting of markdown syntax brings significant value? 

I think that a similar, but more important thing would be to render markdown properly in quick documentation provider. 

> I have checked out how Kotlin handles this, and their approach is nearly identical.

Can you share a link to code please? I'd like to take a look. Do they also have their own `.flex` specially for docs?

matklad:
An interesting thing would be to detect and resolve references in the doc comments, but there is no official syntax for it yet: https://internals.rust-lang.org/t/pre-rfc-standard-form-to-reference-things/2720/19

matklad:
>handling these /// etc. fences would be very, very complicated

Hm, what exactly is difficult with fences? Can we strip them, pass the resulting code to some existing mardown implementation and then fix the offsets of tokens? And can we directly depend on intellij-markdown? I think that it's better to call someone else's code to deal with markdown :)

jajakobyly:
> Can you share a link to code please?

https://github.com/JetBrains/kotlin/tree/master/compiler/frontend/src/org/jetbrains/kotlin/kdoc/lexer

Yes, I know it's compiler frontend, but [it's registered in highlighter](https://github.com/JetBrains/kotlin/blob/master/idea/idea-analysis/src/org/jetbrains/kotlin/idea/highlighter/KotlinHighlightingLexer.java) ([and it seems to be the only place where this is used](https://github.com/JetBrains/kotlin/search?utf8=%E2%9C%93&q=KDocLexer)).

> Do they also have their own `.flex` specially for docs?

[Yes](https://github.com/JetBrains/kotlin/blob/master/compiler/frontend/src/org/jetbrains/kotlin/kdoc/lexer/KDoc.flex)

> I think that a similar, but more important thing would be to render markdown properly in quick documentation provider. 

I think it's a totally unrelated thing - in that case we can freely rely on ready libraries.

> But does this in-code highlighting of markdown syntax brings significant value? 

Did you mean significant value over not highlighting at all, or finding other way to do this (annotators)? If at all then well, this is just a feature - it's not obligatory :smile:. If other way then only this way supports highlighting on-demand (when user is typing) afaik.

> Hm, what exactly is difficult with fences? Can we strip them, pass the resulting code to some existing mardown implementation and then fix the offsets of tokens?

I'm afraid that fixing these offsets might be not so easy. It's not only adding 4 to all token offsets, but also dealing with line breaks.

> And can we directly depend on intellij-markdown?

Well, the problem is that this plugin relies on some markdown parsing library ([more precisely: on this jar](https://github.com/JetBrains/intellij-plugins/blob/master/markdown/lib/intellij-markdown.jar), I don't know where it's from), so the only way to use it is to depend on this plugin (or detect it at runtime, but damn this is terribly overcomplicating simple problem). And again: fixing offsets.

> I think that it's better to call someone else's code to deal with markdown :)

I agree, though about this, but resigned from this idea :wink:

matklad:
Hm, ok, If Kotlin does the same then probably there is no significantly better way to handle this. Will take a more thorough look at this. 

But please do investigate if we can use intellij-markdown. It's bundled with intelliJ. You'll need some runtime detection of the plugin (`<depends optional=Ture>` in `plugin.xml`), but I suspect this to be simpler then creating another `.flex` file. 

>It's not only adding 4 to all token offsets, but also dealing with line breaks.

You don't need to deal with line breaks. You need to be able to answer the question "how many `^/// ` were there before offset `j`?", aka preffix sum. I think an algorithm based on calculating the cumulative total should handle this.

jajakobyly:
Ok, I'm closing this and will reimplement this in separate PR

matklad:
@jajakobyly does this mean that you've find a way to reuse intellij-markdown?

jajakobyly:
> @jajakobyly does this mean that you've found a way to reuse intellij-markdown? 

Have an idea how to generally integrate 3rd party parser, but I'm not sure whether it'll be intellij-markdown or [commonmark-java](https://github.com/atlassian/commonmark-java) yet

matklad:
Nice! I hope it wouldn't be to difficult, or otherwise we will had to revisit this roll your own .flex approach, or even postpone the issue.

PR #493, opened 2016-06-26
[WIP] (PSI): Make RustMacroItemElement extend RustOuterAttributeOwner

bgourlie:
`RustMacroItemElement` declares a `getOuterAttrList()`, it should really just extend `RustOuterAttributeOwner`.  This is my naive attempt to do this.  It's a WIP because I still need to figure out how to make `RustMacroItemElement` not declare its own `getOuterAttrList()` method.

This is a generalization I need for [this](https://github.com/intellij-rust/intellij-rust/pull/486) PR.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
+++ b/src/main/kotlin/org/rust/lang/core/grammar/rust.bnf
@@ -222,7 +222,6 @@
     extends     ("pat_(wild|ref|tup|vec|macro|struct|enum|ident|range|uniq|qual_path)")  = pat
     extends     (".*_attr") = attr
 
-    implements  ("macro_item") = [ "org.rust.lang.core.psi.RustOuterAttributeOwner" "org.rust.lang.core.psi.RustCompositeElement" ]
     elementType ("expr_like_macro|item_like_macro") = macro
     extends     ("(try|format_like)_macro") = macro
     elementType (".*_macro_invocation|macro_rules_invocation") = macro_invocation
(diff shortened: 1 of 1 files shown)


matklad:
> I still need to figure out how to make RustMacroItemElement not declare its own getOuterAttrList() method.

I believe that you can't do this and that you don't want to do this. The interface method and, most importantly, its implementation in the corresponding Impl class are generated by the Grammar Kit. 

Moreover, the `VisibilityOwner` interface was constructed to use the exact same method signature (modulo automatic conversion of Kotlin properties to Java getters) such that the GK generated method implements this interface. 

You don't need `CompositeElement`, it is automatically implemented for all elements. And can you please put `implements` block directly on the `macro_item` rule?

Also, you can add a test for this. See `RustCompleteParsingTestCase`/

bgourlie:
So for the test I have issue492.rs:

    macro_rules! vec {
    }

and then issue492.txt:

```
FILE
  RustMacroItemElementImpl(MACRO_ITEM)
    RustMacroDefinitionElementImpl(MACRO_DEFINITION)
      RustMacroInvocationElementImpl(MACRO_INVOCATION)
        PsiElement(<IDENTIFIER>)('macro_rules')
        PsiElement(!)('!')
      PsiWhiteSpace(' ')
      PsiElement(<IDENTIFIER>)('vec')
      PsiWhiteSpace(' ')
      RustMacroArgElementImpl(MACRO_ARG)
        PsiElement({)('{')
        PsiWhiteSpace('\n')
        PsiElement(})('}')
```

How do I make this assert that `RustMacroItemElementImpl` extends `RustOuterAttributeOwner`?

matklad:
Thanks!

I was thinking about the test that makes sure that attributes on macros are a valid syntax an are properly attached to macro nodes. I've added it in https://github.com/intellij-rust/intellij-rust/commit/865567e175d92b57f6caf41818dca4452c62892e

PR #497, opened 2016-06-27
(FMT): fix block expanding heuristics and allow one line match exprs

### LINKED ISSUE #496: Formatting - one-liners
Please add setting for formatting to allow one-line matches

```rust
let x = match func() { Ok(v) => v.unwrap_or(0), Err(_) => 0};
```

jajakobyly:
fixes #496


src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
```diff
@@ -178,20 +177,24 @@ private data class SpacingContext(val node1: ASTNode,
...
             val psi1 = node1.psi
             val psi2 = node2.psi
             val elementType1 = psi1.node.elementType
             val elementType2 = psi2.node.elementType
+            val parentType = node1.treeParent.elementType
             val parentPsi = psi1.parent
             val (ncPsi1, ncPsi2) = omitCommentBlocks(node1, psi1, node2, psi2)
-            return SpacingContext(node1, node2, psi1, psi2, elementType1, elementType2, parentPsi, ncPsi1, ncPsi2)
+            return SpacingContext(node1, node2, psi1, psi2, elementType1, elementType2,
+                parentType, parentPsi, ncPsi1, ncPsi2, ctx)
```
matklad: Just a thought: instead of a factory fucntion, we can do something like

```
class Ctx(child1: ASTBlock, child2: ASTBloc) {
    val node1 = child1.node
    val node2 = child2.node
    ...
}
```

This should cut down on the boilerplate (and should be adressed in a separate PR).

PR #499, opened 2016-06-28
Improve folding

Furyhunter:
Couple of improvements to folding.

* Now using a recursive visitor. This should have much higher performance than using a find-all operation for every kind of fold possible on every edit.
* A few more foldables.
* Block comments fold with appropriate fold indicator.

Future improvements could include folding up all attributes in a block (doc comments included), various argument-like elements, use blocks.

### DIFF
--- a/src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
+++ b/src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
@@ -94,6 +94,32 @@ class RustFoldingBuilder() : FoldingBuilderEx(), DumbAware {
                     descriptors += FoldingDescriptor(o.node, TextRange(o.lbrace.textOffset, rbrace.textOffset + 1))
                 }
             }
+
+            override fun visitMatchExpr(o: RustMatchExprElement) {
+                super.visitMatchExpr(o)
+
+                val body = o.matchBody
+                if (body != null) {
+                    descriptors += FoldingDescriptor(o.node, body.textRange)
+                }
+            }
+
+            override fun visitMacroArg(o: RustMacroArgElement) {
+                super.visitMacroArg(o)
+
+                val lbrace = o.lbrace
+                val rbrace = o.rbrace
+
+                if (lbrace != null && rbrace != null) {
+                    descriptors += FoldingDescriptor(o.node, TextRange(lbrace.textOffset, rbrace.textOffset + 1))
+                }
+            }
+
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -6,79 +6,139 @@ import com.intellij.lang.folding.FoldingDescriptor
...
-import org.rust.lang.core.psi.util.descendentsOfType
+import org.rust.lang.core.psi.visitors.RustRecursiveElementVisitor
 import java.util.*
 
 class RustFoldingBuilder() : FoldingBuilderEx(), DumbAware {
-    override fun getPlaceholderText(node: ASTNode): String = "{...}"
+    override fun getPlaceholderText(node: ASTNode): String {
+        when {
+            node.psi is PsiComment -> return "/* ... */"
+            else -> return "{...}"
```
matklad: Nit: small code economy can be achieved in similar cases by pulling `return` outside the conditional expression:

```
return when {
    node.psi is PsiComment -> "/* .. */"
    else -> "{ ... }"
}
```

Also I usually prefer `if` over `when`:

```
return if (node.psi is PsiComment)  "/* ... */" else "{ ... }"
```

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -6,79 +6,139 @@ import com.intellij.lang.folding.FoldingDescriptor
...
+                }
+            }
+
+            override fun visitUseGlobList(o: RustUseGlobListElement) {
+                super.visitUseGlobList(o)
+
+                descriptors += FoldingDescriptor(o.node, o.textRange)
+            }
+
+            override fun visitComment(comment: PsiComment?) {
```
matklad: You can remove `?` from the method signature. It is added by IDE just because the parent method is not annotated, not because it can really be null: https://youtrack.jetbrains.com/issue/KT-12213

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -6,79 +6,139 @@ import com.intellij.lang.folding.FoldingDescriptor
...
+                super.visitComment(comment)
+
+                if (comment != null) {
+                    when (comment.tokenType) {
```
matklad: This could be an `if`.

matklad:
Thanks!

PR #500, opened 2016-06-28
(RUN) "--explain Exxxx" now links to relevant Rust error index entry

### LINKED ISSUE #450: Add link in log that runs rustc --explain
When you have a message suggesting **rustc --explain**, that line should be hidden.
For example, hide this line:
> examples/test_video.rs:58:57: 58:58 help: run `rustc --explain E0308` to see a detailed explanation

Instead, the **[E0308]** error code above the removed line should be a link that runs `rustc --explain E0308` for you and displays the result in a popup or dialog.

This would make logs a bit shorter and easier to scan visually.

This is low priority, but I think it would be a cool feature and probably fairly easy. Maybe tag this up for grabs?

tov:
This should resolve #450. I hope you like it.

Cheers,
Jesse


matklad:
Thanks!

PR #501, opened 2016-06-28
(RUN): Adds "backtrace" checkbox to run configurations

### LINKED ISSUE #490: Add "backtrace" checkbox to run configurations
If enabled, this checkbox should pass `RUST_BACKTRACE=1` via environment variable

tov:
This should resolve #490.

Cheers,
Jesse


matklad:
The test results should be refreshed. You can just remove all the files from https://github.com/intellij-rust/intellij-rust/tree/master/src/test/resources/org/rust/cargo/runconfig/producers/fixtures and run the tests. This will generate xmls with updated results. 

It's a shame that these tests are than brittle, but I don't know a better way to check run configurations.

tov:
Oops, sorry. Yeah, I was looking at the tests for a while and couldn’t figure out how to test the new thing, but I didn’t know about regenerating them. Doing that now.

matklad:
Splendid! The  checkbox position in the UI looks a bit off, but I don't think this is really important.

PR #502, opened 2016-06-28
Fix cargo version regex to support missing stable or nightly.

oistein:
Eg: cargo 0.10.0 (10ddd7d 2016-04-08)

Defaults to stable if missing.


matklad:
Thanks!

PR #503, opened 2016-06-29
Additional Syntax Highlighting 

mfarrugi:
Added implementations for some of the simpler syntax highlighting, tests pass and it anecdotally works. 

As I started pulling out elements of expressions it started feeling wrong to do more. Is it possible that this sort of thing could be generated, perhaps by identifying leaf elements and coloring those a particular way? 

I'm aware the resolveColor code is unsightly (and I didn't _really_ add appropriate tests), but I figured I would throw this into a pull request before I had to do more merges and refactoring. Future work would involve moving closer to providing the same the elaborate syntax highlighting available for kotlin or java.

Screenshots http://imgur.com/a/VoAd4 with some placeholder colors :)

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
@@ -2,73 +2,89 @@ package org.rust.ide.annotator
 
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.Annotator
-import com.intellij.openapi.editor.colors.TextAttributesKey
 import com.intellij.psi.PsiElement
 import org.rust.ide.colors.RustColor
 import org.rust.ide.highlight.syntax.RustHighlighter
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.mixin.isMut
 import org.rust.lang.core.psi.util.elementType
+import org.rust.lang.core.psi.visitors.RustComputingVisitor
 
 // Highlighting logic here should be kept in sync with tags in RustColorSettingsPage
 class RustHighlightingAnnotator : Annotator {
-    override fun annotate(element: PsiElement, holder: AnnotationHolder) = element.accept(object : RustElementVisitor() {
-        override fun visitAttr(o: RustAttrElement) {
-            holder.highlight(o, RustColor.ATTRIBUTE)
+
+    override fun annotate(element: PsiElement, holder: AnnotationHolder) {
+        if (element is RustReferenceElement) {
+            if (element is RustPathElement && element.isPrimitive) {
+                holder.highlight(element, RustColor.PRIMITIVE_TYPE)
+                return
+            }
+            val text = when (element) {
+                is RustPathElement -> element.identifier
+                else -> element
+            }
+            val ref = element.reference.resolve() ?: return
+            // Highlight the element dependent on what it's referencing.
+            val color = highlightingVisitor().computeNullable(ref)?.second
+            holder.highlight(text, color)
+        } else {
+            val (text, color) = highlightingVisitor().computeNullable(element) ?: return
+            holder.highlight(text, color)
         }
+    }
+
+    fun AnnotationHolder.highlight(element: PsiElement?, color: RustColor?) {
+        val textAttributesKey = color?.textAttributesKey
+        if (element != null && textAttributesKey != null) {
+            createInfoAnnotation(element, null).textAttributes = textAttributesKey
+        }
+    }
+
+    fun highlightingVisitor() = object : RustComputingVisitor<Pair<PsiElement?, RustColor?>>() {
+
+        fun highlight(element: PsiElement?, color: RustColor?) = set { Pair(element, color) }
 
         override fun visitLitExpr(o: RustLitExprElement) {
             // Re-highlight literals in attributes
             if (o.parent is RustMetaItemElement) {
                 val literal = o.firstChild
-                holder.highlight(literal, RustHighlighter.map(literal.elementType))
+                highlight(literal, RustHighlighter.map(literal.elementType))
             }
         }
 
-        override fun visitMacroInvocation(m: RustMacroInvocationElement) {
-            holder.highlight(m, RustColor.MACRO)
-        }
+        override fun visitTypeParam(o: RustTypeParamElement) = highlight(o.identifier, RustColor.TYPE_PARAMETER)
 
-        override fun visitTypeParam(o: RustTypeParamElement) {
-            holder.highlight(o, RustColor.TYPE_PARAMETER)
-        }
+        override fun visitAttr(o: RustAttrElement) = highlight(o, RustColor.ATTRIBUTE)
+
+        override fun visitTraitRef(o: RustTraitRefElement) = highlight(o.path.identifier, RustColor.TRAIT)
 
         override fun visitPatBinding(o: RustPatBindingElement) {
             if (o.isMut) {
-                holder.highlight(o.identifier, RustColor.MUT_BINDING)
+                highlight(o.identifier, RustColor.MUT_BINDING)
             }
         }
 
-        override fun visitPath(o: RustPathElement) {
-            o.reference.resolve().let {
-                if (it is RustPatBindingElement && it.isMut) {
-                    holder.highlight(o.identifier, RustColor.MUT_BINDING)
-                }
-            }
-        }
+        override fun visitEnumItem(o: RustEnumItemElement)       = highlight(o.identifier, RustColor.ENUM)
+        override fun visitEnumVariant(o: RustEnumVariantElement) = highlight(o.identifier, RustColor.ENUM_VARIANT)
 
-        override fun visitFnItem(o: RustFnItemElement) {
-            holder.highlight(o.identifier, RustColor.FUNCTION_DECLARATION)
-        }
+        override fun visitStructItem(o: RustStructItemElement)   = highlight(o.identifier, RustColor.STRUCT)
+        override fun visitTraitItem(o: RustTraitItemElement)     = highlight(o.identifier, RustColor.TRAIT)
+        override fun visitModDeclItem(o: RustModDeclItemElement) = highlight(o.identifier, RustColor.MODULE)
+        override fun visitModItem(o: RustModItemElement)         = highlight(o.identifier, RustColor.MODULE)
 
-        override fun visitImplMethodMember(o: RustImplMethodMemberElement) {
-            val color = if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD
-            holder.highlight(o.identifier, color)
-        }
+        override fun visitFieldDecl(o: RustFieldDeclElement)     = highlight(o.identifier, RustColor.FIELD)
 
-        override fun visitTraitMethodMember(o: RustTraitMethodMemberElement) {
-            val color = if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD
-            holder.highlight(o.identifier, color)
-        }
-    })
+        override fun visitExternCrateItem(o: RustExternCrateItemElement) = highlight(o.identifier, RustColor.CRATE)
 
-    private fun AnnotationHolder.highlight(element: PsiElement?, color: RustColor?) {
-        highlight(element, color?.textAttributesKey)
-    }
+        //TODO: When are element and element.identifier different?
+        override fun visitMacroInvocation(m: RustMacroInvocationElement) = highlight(m, RustColor.MACRO)
+        override fun visitMethodCallExpr(o: RustMethodCallExprElement)   = highlight(o.identifier, RustColor.INSTANCE_METHOD)
+        override fun visitFnItem(o: RustFnItemElement)                   = highlight(o.identifier, RustColor.FUNCTION_DECLARATION)
 
-    private fun AnnotationHolder.highlight(element: PsiElement?, textAttributesKey: TextAttributesKey?) {
-        if (element != null && textAttributesKey != null) {
-            createInfoAnnotation(element, null).textAttributes = textAttributesKey
-        }
+        override fun visitImplMethodMember(o: RustImplMethodMemberElement) =
+            highlight(o.identifier, if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD)
+        override fun visitTraitMethodMember(o: RustTraitMethodMemberElement) =
+            highlight(o.identifier, if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD)
+
     }
 }
--- a/src/main/kotlin/org/rust/ide/colors/RustColorSettingsPage.kt
+++ b/src/main/kotlin/org/rust/ide/colors/RustColorSettingsPage.kt
@@ -10,17 +10,13 @@ class RustColorSettingsPage : ColorSettingsPage {
     private val ATTRS = RustColor.values().map { it.attributesDescriptor }.toTypedArray()
 
     // This tags should be kept in sync with RustHighlightingAnnotator highlighting logic
-    private val ANNOTATOR_TAGS = mapOf(
-        "attribute" to RustColor.ATTRIBUTE,
-        "macro" to RustColor.MACRO,
-        "type-parameter" to RustColor.TYPE_PARAMETER,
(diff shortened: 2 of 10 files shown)


src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -39,14 +52,30 @@ class RustHighlightingAnnotator : Annotator {
             }
         }
 
-        override fun visitPath(o: RustPathElement) {
-            o.reference.resolve().let {
-                if (it is RustPatBindingElement && it.isMut) {
-                    holder.highlight(o.identifier, RustColor.MUT_BINDING)
-                }
+        // @TODO ReferenceElement
```
mfarrugi: I had trouble replacing RustPathElement with RustReferenceElement, but didn't look into it

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -60,15 +89,46 @@ class RustHighlightingAnnotator : Annotator {
             val color = if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD
             holder.highlight(o.identifier, color)
         }
+
+            // Capture the color of the element w/o mutating anything.
+        private fun resolveColor(e: PsiElement): RustColor? {
```
mfarrugi: ¯\_(ツ)_/¯

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -31,6 +35,15 @@ class RustHighlightingAnnotator : Annotator {
 
         override fun visitTypeParam(o: RustTypeParamElement) {
             holder.highlight(o, RustColor.TYPE_PARAMETER)
+            o.typeParamBounds?.let {
+                holder.highlight(it.colon, RustColor.KEYWORD) // feels off
```
bgourlie: I'd favor explicit lambda params here over `it`.  A bit hard to grok with `it` changing context so many times.

matklad: Year, as a general rule, it's better to avoid nested lambdas with implicit  parameters, or you may hit some [nasty bugs]. Also, using early returns helps to avoid deep nesting, so I probably would write something more like this:

```
val bounds = o.typeParamBouds ?: return
holder.highligh(bounds.colon, RustColor.KEYWORD)
```

Highlighting trait ref can be done in `visitTraitRef`.


[nasty bugs]: https://youtrack.jetbrains.com/issue/KT-11849

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -12,7 +12,11 @@ import org.rust.lang.core.psi.util.elementType
 
 // Highlighting logic here should be kept in sync with tags in RustColorSettingsPage
 class RustHighlightingAnnotator : Annotator {
-    override fun annotate(element: PsiElement, holder: AnnotationHolder) = element.accept(object : RustElementVisitor() {
+
+    override fun annotate(element: PsiElement, _holder: AnnotationHolder) {
+        element.accept(object : RustElementVisitor() {
+        var holder = wrap(_holder)
```
matklad: I think that instead of creating `Highlighter` class we can more naturally declare a couple of local extension functions:

```Kotlin
    override fun annotate(element: PsiElement, holder: AnnotationHolder) {
        fun PsiElement?.highlight(textAttributesKey: TextAttributesKey?) {
            if (this != null && textAttributesKey != null) {
                holder.createInfoAnnotation(element, null).textAttributes = textAttributesKey
            }
        }

        fun PsiElement?.highlight(color: RustColor?) {
            highlight(color?.textAttributesKey)
        }


        element.accept(object : RustElementVisitor() {

            override fun visitAttr(o: RustAttrElement) {
                o.highlight(RustColor.ATTRIBUTE)
            }
```

matklad: And seems like the level of indentation is missing

matklad: Ah, I see know the `resolveColor` trick. Hm, I wonder if this can be simplified by using a more side effect free style...

What about something like this:

```Kotlin
data class HighlightingInfo (
    val textAttributesKey: TextAttributesKey
) {
    fun apply(element: PsiElement, holder: AnnotationHolder) {
        holder.createInfoAnnotation(element, null).textAttributes = textAttributesKey
    }
}

override fun annotate(element: PsiElement, holder: AnnotationHolder) {
    if (element is RustReference) {
        val resolve = element.reference.resolve() ?: return
        val highlightingInfo = calculateHighlightingInfo(resolve)
        highlightingInfo.apply(element, holder) 
    }
}

```

?

matklad:
These new colors look nice!

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -31,6 +35,15 @@ class RustHighlightingAnnotator : Annotator {
...
             holder.highlight(o, RustColor.TYPE_PARAMETER)
+            o.typeParamBounds?.let {
+                holder.highlight(it.colon, RustColor.KEYWORD) // feels off
+                it.polyboundList.forEach {
+                    it.bound.traitRef?.let {
+                        // Should this visit path instead?
+                        holder.highlight(it.path.identifier, RustColor.TRAIT)
+                    }
+                }
+            }
```
jajakobyly: Do I understand correctly that you are colouring whole type parameter and then fixing particular parts? 

Why not just then:
```kotlin
override fun visitTypeParam(o: RustTypeParamElement) {
    holder.highlight(o.identifier, RustColor.TYPE_PARAMETER)
    o.typeParamBounds?.polyboundList?.forEach { polybound ->
        polybound.bound.traitRef?.let {
            holder.highlight(it.path.identifier, RustColor.TRAIT)
        }
    }
}
```

src/main/kotlin/org/rust/ide/colors/RustColors.kt
```diff
@@ -19,6 +19,11 @@ enum class RustColor(val humanName: String, externalName: String, fallback: Text
     STRING                ("String", "org.rust.STRING", Default.STRING),
     NUMBER                ("Number", "org.rust.NUMBER", Default.NUMBER),
 
+    STRUCT                ("Struct", "org.rust.STRUCT", Default.CLASS_NAME),
+    TRAIT                 ("Trait",  "org.rust.TRAIT",  Default.INTERFACE_NAME),
+    ENUM                  ("Enum",   "org.rust.ENUM",   Default.CLASS_NAME),
+    MODULE                ("Module", "org.rust.MODULE",   Default.IDENTIFIER),
```
jajakobyly: Sorry for nitpicking, but spacing is inconsistent here :smile: Also I'd name these colours `STRUCT_NAME` etc. instead as such naming is more meaningful imho.

src/main/kotlin/org/rust/ide/colors/RustColorSettingsPage.kt
```diff
@@ -10,7 +10,13 @@ class RustColorSettingsPage : ColorSettingsPage {
     private val ATTRS = RustColor.values().map { it.attributesDescriptor }.toTypedArray()
 
     // This tags should be kept in sync with RustHighlightingAnnotator highlighting logic
+    // @TODO generate from RustColor
```
jajakobyly: Good idea :+1:

src/main/kotlin/org/rust/ide/colors/RustColorSettingsPage.kt
```diff
@@ -10,7 +10,13 @@ class RustColorSettingsPage : ColorSettingsPage {
...
+    // @TODO Figure out how to namespace menu elements a la
+    //       https://github.com/JetBrains/kotlin/blob/8b30e7ef4e48494ba245b441a5b23142f1d6ae33/idea/idea-analysis/src/org/jetbrains/kotlin/idea/KotlinBundle.properties
```
jajakobyly: I don't know whether this is available on all IDEs we support. @matklad haven't these groups been introduced in 2016.1, have they?

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -39,14 +52,30 @@ class RustHighlightingAnnotator : Annotator {
...
+            // This highlights the *path* dependent on the reference type.
+            resolveColor(ref)?.let { holder.highlight(path.identifier, it) }
         }
 
+        override fun visitMethodCallExpr(o: RustMethodCallExprElement) {
+            holder.highlight(o.identifier, RustColor.INSTANCE_METHOD)
+        }
+
+        // @TODO Bind types and paths similarly
+        //       grammar was tough to read, doesn't look like it can differentiate between bottom types.
```
jajakobyly: Grammar won't tell you what kind of item does the path point to. You have to resolve it, which should be already doable.

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
jajakobyly: Anyway, highlighting trait reference should be moved to dedicated visitor method.

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -60,15 +89,46 @@ class RustHighlightingAnnotator : Annotator {
...
+        private fun resolveColor(e: PsiElement): RustColor? {
+            var outColor: RustColor? = null
+            val prevHolder = holder
+            this.holder = object : Highlighter {
+                override fun highlight(element: PsiElement?, textAttributesKey: TextAttributesKey?) {
+                    outColor = null;
+                }
+
+                override fun highlight(element: PsiElement?, color: RustColor?) {
+                    assert(outColor == null, { e.toString() + " is not a leaf element, ambiguous color." })
```
jajakobyly:
```kotlin
assert(outColor == null) { "$e is not a leaf element, ambiguous color." }
```

is more _kotlinish_

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -12,7 +12,11 @@ import org.rust.lang.core.psi.util.elementType
 
 // Highlighting logic here should be kept in sync with tags in RustColorSettingsPage
 class RustHighlightingAnnotator : Annotator {
-    override fun annotate(element: PsiElement, holder: AnnotationHolder) = element.accept(object : RustElementVisitor() {
```
jajakobyly: Nitpicking: using expression syntax here allowed us to reduce level of indentation.

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
mfarrugi: How should that work for resolving compound elements, is it still fine to skip? 

It would be nice to instead be able to produce a map of leaf elements to colors, and then have a visitor resolve leaf elements.. but does that actually make sense to do?

matklad: I don't know what is the best way to handle this, but I see two possible solutions:

* we can add `val target: PsiElement` to `HighlightingInfo`, such that `apply` method would update the `target`'s coloring. 

* we can color the leaf elements directly, using `leaf.parent` to make decisions about what color should be applied. 

I think that these are orthogonal issues:

* using stateless color computation to be able to reuse code for coloring references without "holder context switches"

* deciding, which is best: color leaf elements while visiting them or their parent.

It's interesting that in Go Plugin they seem to simply duplicate some code for coloring references and non references: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/src/com/goide/highlighting/GoHighlightingAnnotator.java

matklad:
>but also in the ballpark of being generable from the grammar.

Can you explain this a bit more? I don't think it should be possible to generate something reasonable from the grammar. You already have access to PSI here, and PSI is a reflection of the grammar. 

I think that there should be two main code paths: highlighting a reference element to the color of the target element, and highlighting individual elements based on their element types and maybe some other attributes (underlined `mut` values, for example). I think this PR is pretty close to what the general structure should look like, modulo a bit overly complicated handling of references.

src/test/resources/org/rust/ide/annotator/fixtures/type_parameters.rs
```diff
@@ -3,6 +3,6 @@ trait <info>MyTrait</info> {
     fn <info>some_fn</info>(&self);
 }
 
-struct <info>MyStruct</info><<info>N<info>:</info> ?<info>Sized</info>+<info>Debug</info>+<info><info>MyTrait</info></info></info>> {
+struct <info>MyStruct</info><<info>N</info>: ?<info><info>Sized</info></info>+<info><info>Debug</info></info>+<info><info><info>MyTrait</info></info></info>> {
```
mfarrugi: Is this correct? It's suspicious to me, but I don't know what else is going on.

matklad: No, I believe `traitRefs` are highlighted twice here, while they shouldn't be.  I think that if you remove completely [this `forEach`](https://github.com/intellij-rust/intellij-rust/pull/503/files#diff-bc8f1651165bb2662ad415a152f3f9eaR61) then the result will be just right.

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -12,10 +11,40 @@ import org.rust.lang.core.psi.util.elementType
...
+            resolveColor(ref)?.let { highlighter.highlight(text, it) }
+        } else {
+            element.accept(highlightingVisitor(highlighter))
         }
+    }
+
+    // Capture the color of the element w/o mutating anything.
+    fun resolveColor(e: PsiElement): RustColor? {
+        var outColor: RustColor? = null;
+        val resolver = highlightingVisitor(object : Highlighter { // @TODO Could use RustComputingVisitor
```
matklad: Please do, I hope that data class for highlighting info would be much clearer then `Highlighter` interface.

mfarrugi:
Hmm, I have pull set to rebase by default so I ended up force pushing instead of googling my way into undoing it. Apologies for nuking the PR, though I'm not quite sure what standard practice is anyways.

matklad:
Yep, we should have added PR best practices to the `CONTRIBUTING.md`. In short, we do prefer `rebase` && `push --force-with-lease` approach, but looks like something has gone wrong with the history here. In particular, `	(INSP): clean up inspections` commit does not belong to this PR (I've actually broken the build with it :) ). 

Would you mind opening a fresh PR with this changes? I think that it's even better to open two pull requests: one with [Rust colors refactoring] piece which can be merged right away, and another one with actual highlighter changes.

[Rust colors refactoring]: https://github.com/intellij-rust/intellij-rust/pull/503/files#diff-644f2bb751ec92f0171fc8b70ab372bbR10

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -2,73 +2,89 @@ package org.rust.ide.annotator
...
-            val color = if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD
-            holder.highlight(o.identifier, color)
-        }
-    })
+        override fun visitExternCrateItem(o: RustExternCrateItemElement) = highlight(o.identifier, RustColor.CRATE)
 
-    private fun AnnotationHolder.highlight(element: PsiElement?, color: RustColor?) {
-        highlight(element, color?.textAttributesKey)
-    }
+        //@TODO When are element and element.identifier different?
```
matklad: We use `//TODO: bla-bla-bla` instead of `//@TODO `

matklad:
@mfarrugi I've written up some tips about synchronizing pull requests, you may find them helpful: https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#pull-requests-best-practices

PR #504, opened 2016-06-29
Documentation comments highlighting, try #2

### LINKED ISSUE #495: Add html entities escaping in quick doc view
Currently quick doc sometimes omits angle brackets, e.g. in doc for AsRef:

```
Both String and &str implement AsRef<str>:
fn is_hello  >(s: T) {
assert_eq!("hello", s.as_ref());
}
```

instead of `fn is_hello<T: AsRef<str>>(s: T) {`

jajakobyly:
This is much cleaned up reimplementation of #491. Doesn't reinvent the wheel with merging lexers, supports simpler link syntax, is cleaner, less buggy etc. Also fixes/improves documentation fetching.

It looks nearly the same as it's predecessor (now code blocks are a bit darker in _Default_):

~~Fixes #495~~ Ok I'll manage to fix this in separate PR


alexeykudinkin:
WOW! That looks awesome! :clap:

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
```diff
@@ -0,0 +1,179 @@
...
+    return zzMarkedPos == zzEndRead;
+  }
+
+  private void docHeadingTrimRight() {
+    int i = yylength() - 1;
+    char ch = yycharat(i);
+
+    // Trim trailing *****/ if we are at the end of block doc comment
+    if (i >= 1 && MAIN_STATE == IN_BLOCK && isLastToken() && ch == '/' && yycharat(i - 1) == '*') {
+      i -= 2;
```
matklad: Is `-=2`safe? `-=1` should be safe because of the `i >= 1` condition.

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
```diff
@@ -0,0 +1,179 @@
...
+      ch = yycharat(i);
+      while (ch == '*') {
+        ch = yycharat(--i);
```
matklad: Why this `--i` can't overflow over zero?

matklad: Ah, I see, this must start with a heading hash, which is at least one `#`

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
jajakobyly: Hmm, actually I can't answer this question :sweat_smile: Will have to revise this condition

jajakobyly:
Ready

jajakobyly:
Sorry for another huge PR, but it's quite monolithic :disappointed:

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
```diff
@@ -0,0 +1,81 @@
...
+
+// TODO Better tests?
+
+fun RustDocAndAttributeOwner.documentation(): String? =
+    (outerDocs() + innerDocs())
+        .flatMap { it.lines() }
+        .map { it.removeDecoration() }
+        .joinToString("\n")
+
+fun RustDocAndAttributeOwner.documentationFormatted(): String? {
```
matklad: A slightly better name would be `documentationAsHtml`.

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+}
+
+private fun Pair<RustDocKind, String>.lines(): Sequence<Pair<RustDocKind, String>> =
+    second.splitToSequence("\r\n", "\r", "\n").map { first to it }
+
+// FIXME: This trimming is devastating
+// FIXME: Asterisk handling in block comments is just wrong
+private fun Pair<RustDocKind, String>.removeDecoration(): String = when (first) {
```
matklad: I think this can be a method on `RustDocKind`.

src/main/kotlin/org/rust/lang/doc/psi/RustDocKind.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    OuterEol {
+        override val infix: String = "///"
+        override val prefix: String = infix
+    };
+
+    abstract val prefix: String
+    abstract val infix: String
+
+    val isEol: Boolean
```
matklad: unused

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
matklad: Also, this is a self contained function which can be unit tested on it's own ;)

src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java
```diff
@@ -0,0 +1,13 @@
+package org.rust.lang.doc.psi;
+
+public interface RustDocElementTypes {
+    RustDocTokenType DOC_TEXT = new RustDocTokenType("<DOC_TEXT>");
+    RustDocTokenType DOC_DECO = new RustDocTokenType("<DOC_DECO>");
```
matklad: What is this `DOC_DECO`? It's not obvious from the name.

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
```diff
@@ -0,0 +1,190 @@
...
+%{
+  private int MAIN_STATE = YYINITIAL;
+
+  public _RustDocHighlightingLexer(boolean isBlock) {
+    this((java.io.Reader)null);
+    MAIN_STATE = isBlock ? IN_BLOCK : IN_EOL;
+  }
+
+  // FIXME(jajakobyly): I think it is possible this will break, though I couldn't make it do so :-)
+  private char CODE_FENCE_DELIM = '\0';
```
matklad: I don't see any problem with this .

matklad: Ah, there actually can be something funny with `getCurrentPosition` / `restore` pair of methods. Perhaps you should override `getCurrentPosition` to also contain the `CODE_FENCE_DELIM`. 

And perhaps the same should be done for `rust.flex`? Are we [violating the contract](http://www.jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/implementing_lexer.html)

>A lexer that can be used incrementally may need to return its state, which means the context corresponding to each position in a file. For example, a Java lexer could have separate states for top level context, comment context and string literal context. An important requirement for a syntax highlighting lexer is that its state must be represented by a single integer number returned from Lexer.getState(). That state will be passed to the Lexer.start() method, along with the start offset of the fragment to process, when lexing is resumed from the middle of a file. Lexers used in other contexts can always return 0 from the getState() method.

? Looks like it, but I have never observed any problems...

matklad:
@jajakobyly have you seen `com.intellij.embedding.IndentEatingLexer`? Looks like it implements something similar to stripping prefixes.

src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java
jajakobyly: `DECO` comes from _decoration_. It refers to all these `///`, `//!, `/*!`, `*`, etc. Sincerely I don't know how they are properly called in English :sweat_smile:

matklad: A short example comment would help :)

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
jajakobyly: Yes, this is exactly the reason why I see problem here. The same case caused the bug fixed in  	687cb5a

jajakobyly:
> @jajakobyly have you seen com.intellij.embedding.IndentEatingLexer? Looks like it implements something similar to stripping prefixes.

Never heard about it. Will take a look :)

src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java
alexeykudinkin: Do we need actually a separate token-type for them?

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
jajakobyly: > Also, this is a self contained function which can be unit tested on its own ;)

Not here, in separate PR

src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java
jajakobyly: > Do we need actually a separate token-type for them?

Do you suggest merging it with `DOC_TEXT`?

jajakobyly: Have in mind, that these guys are not present in the actual PSI, they are only visible to the highlighter (currently)

jajakobyly:
> IndentEatingLexer

Nice! Good to know that something like this is present in the SDK (in pair with `com.intellij.embedding.EmbeddingUtil#calcBaseIndent`), though I don't think using this somehow here will simplify anything magically.

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
matklad: I wonder if the Rust lexer is broken then...

jajakobyly:
```jflex
%{}
  /**
    * '#+' stride demarking start/end of raw string/byte literal
    */
  private int zzShaStride = -1;

  /**
    * Dedicated storage for starting position of some previously successful
    * match
    */
  private int zzPostponedMarkedPos = -1;

  /**
    * Dedicated nested-comment level counter
    */
  private int zzNestedCommentLevel = 0;
%}
```

yeah, it is possible

src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java
alexeykudinkin: I just don't see any reason to introduct separate entity, though i may be wrong.

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
alexeykudinkin: Not critical in any way, though it's a good practice to make P/Rs as self-contained as possible.

src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java
jajakobyly: I have introduced it mostly for clarity, can freely get rid of it afaik

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
jajakobyly: But this function is broken anyway, so
a) writing unit tests for broken code is nonsense
b) fixing it here is just putting additional complexity to the already big monster

src/main/kotlin/org/rust/lang/doc/lexer/RustDocHighlightingLexer.flex
jajakobyly: Uh, I overrode `getCurrentPosition` / `restore` and just out of curiosity reverted 687cb5a (of course added handling them in mentioned methods) and... the bug haven't disappeared.

src/main/kotlin/org/rust/lang/doc/RustDocPipeline.kt
alexeykudinkin: Fixing it here would have much more sense, than committing broken stuff.

jajakobyly:
Ok

matklad:
@jajakobyly this caused a noticeable regressed performance. Try opening `libsyntax/src/parser/parser.rs` near the end and typing something. I am investigating this at the moment.

alexeykudinkin:
@matklad first of all, let's revert this until fixed unless you (or @jajakobyly) would be able to fix it quickly.

Second, it seems that time to introduce performance tests has come to our land.

matklad:
I have a fix, will push soon. 

:+1: for perf tests monitoring. Hm, we actually have a parser performance/smoke test right now, but without timing assertions, I'll check if it would have caught this regression.

matklad:
Ok, it was a bit more complicated than I expected. [The fix]. This was essentially the same issue as #367. Method which was invoked with `CharSequence, start, end` had `O(CharSequence.length)` complexity instead of `O(end - start)`. Here is [the offending line]. 

Arguably a better fix would be to use a patched jFlex jar from the [idea repository], but I am not very excited about vending one more jar, because we also have the TOML plugin... Perhaps we should publish magical jars to bintray? 


[The fix]: https://github.com/intellij-rust/intellij-rust/commit/69f2ed7bd44f29e3370ef8e360f9433a38839a2f
[the offending line]: https://github.com/intellij-rust/intellij-rust/blob/69f2ed7bd44f29e3370ef8e360f9433a38839a2f/src/main/kotlin/org/rust/lang/core/lexer/RustLexer.skeleton#L75
[idea repository]: https://github.com/JetBrains/intellij-community/tree/master/tools/lexer

matklad:
Issue for perf testing: https://github.com/intellij-rust/intellij-rust/issues/513

jajakobyly:
@matklad why not fixing skeleton file instead of introducing `FlexWrapper` ?

matklad:
@jajakobyly I believe it is impossible to fix a skeleton so that jFlex does not need `char[]`. 

These [changes] must be done in jFlex itself, not in the skeleton. 

[changes]: https://github.com/JetBrains/intellij-community/blob/master/tools/lexer/jflex-changes.patch#L180

Issue #498, opened 2016-06-27
Impl Index


matklad:
To do resolve and completion, we need to answer a question "what traits are implemented for a given type?" Moreover, we need to consider only the traits that are currently in scope , so the question reduces to "Is the trait T implemented for the type S?" However, keep in mind the following "advanced" example:

```Rust
mod a {
    pub trait H {}

    impl<T> H for T {}
}

mod b {
    pub trait V {
        fn foo(&self) {}
    }

    impl<T: super::a::H> V for T {}
}

struct S;

fn main() {
    use b::V;
    let s = S;

    // To understand that `S: V` we need to know that `S: H`,
    // although `H` is not in scope
    s.foo();
}
```

We do not handle generics in any way yet, so I think it's ok to focus on simpler cases for now.

The simplest implementation would be to create an index of all impls. Then in order to answer The Question you would process all entries and check that the trait and the type of the impl resolve to the trait and the type in question. Arguably this is not very efficient, because there are a lot of impls, and majority of them are irrelevant.

The more advanced implementation would index the impls by the name of the trait. You still need to process a number of irrelevant impls though.

It's not easy to index the impl by the type name (`impl<T> X for T where T: A + B {}` ?). However I suspect that the vast majority of impls have a simpler form of `impl X for S` or `impl<T> X for S<X>`, for which we can index by both the type and the trait. So my suggestion is to create *two* indexes:

* simple impl index, which stores impls keyed by the trait and the type name, if there is a type name.

* sophisticated (generic) impl index, keyed by the trait name (can omit this in the initial implementation).

The idea is that you first try to find a simple impl for a given trait and type. If you do not succeed, you try to find a sophisticated impl and then try to solve its constraints. Looks like this would work nicely with plain stub indexes, invalidated on file change (we only need to verify impl's trait and type resolve target)

However there are additional properties which we may want to exploit in indexing/caching.

* Crates form a DAG, and you usually work near the sink of the DAG. It may make sense to introduce `CRATE_MODIFICATION_COUNT` and use it for resolve caching. This count would get an update when the crate or any of it's dependencies is modified.

* Coherence rules say, roughly, that there are only two possible crates where an `impl X for S` can be defined. Namely, they are the crates where `X` is defined and where `S` is defined. In theory, this should allow us to cut down on the number of impls we investigate, because we'll need to consider only two crates. In practice, I think that **all** impls would be in these two crates anyway, so this will not cut down on anything.

matklad:
[...]

PR #508, opened 2016-06-30
Add inherent impl index

matklad:
Ok, here is my take on the impls index (#498). This tackles only inherent impls case (impls without traits). We try to solve the following problem: "for a given type, find all impls." Note that impls may be found anywhere in the project ([example]). 

The core of the problem is that we can't possibly use resolved types in the index, because resolve can be potentially invalidated by any PSI change. So instead we index impls by `baseTypeName`. For

```Rust
impl<U, V> ::foo::bar::Quux<U, V> { }
```

The `baseTypeName` would be `Quux`. This is not exact, so after we found a set of candidate impls, we resolve them to filter out false positives. The approach is inspired by [fast_reject] from `rustc`. Note that this approach may give *false negatives* in the presence of aliases (`use foo::S as T`, `type T = S`). I don't think we should try to handle aliases at the moment, but a possible future approach would be to create a separate index of aliased names and seek impls simultaneously for all potential aliases.  

@alexeykudinkin @jajakobyly does this look reasonable? 

[example]: https://github.com/intellij-rust/intellij-rust/blob/44b491d3e0a7986971686806aaae7f71bf215b29/src/test/resources/org/rust/lang/core/resolve/fixtures/type_aware/hidden_inherent_impl.rs 

[fast_reject]: https://github.com/rust-lang/rust/blob/c2b56fb7a0c24e04227318ca7e5950e9289ee3e4/src/librustc/ty/fast_reject.rs#L37

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/RustFileElementType.kt
+++ b/src/main/kotlin/org/rust/lang/core/RustFileElementType.kt
@@ -5,5 +5,5 @@ import org.rust.lang.RustLanguage
 import org.rust.lang.core.stubs.RustFileStub
 
 object RustFileElementType : IStubFileElementType<RustFileStub>(RustLanguage) {
-    override fun getStubVersion(): Int = 6
+    override fun getStubVersion(): Int = 7
 }
--- a/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplItemImplMixin.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplItemImplMixin.kt
@@ -4,9 +4,7 @@ import com.intellij.lang.ASTNode
 import com.intellij.psi.stubs.IStubElementType
 import org.rust.ide.icons.RustIcons
 import org.rust.lang.core.psi.RustImplItemElement
-import org.rust.lang.core.psi.RustPathTypeElement
 import org.rust.lang.core.psi.impl.RustStubbedElementImpl
-import org.rust.lang.core.psi.referenceName
 import org.rust.lang.core.stubs.elements.RustImplItemElementStub
 import javax.swing.Icon
 
@@ -18,10 +16,4 @@ abstract class RustImplItemImplMixin : RustStubbedElementImpl<RustImplItemElemen
     override fun getIcon(flags: Int): Icon = RustIcons.IMPL
 
     override val isPublic: Boolean get() = false // pub does not affect imls at all
-}
-
-
-val RustImplItemElement.baseTypeName: String? get() {
-    //TODO: move to `RustTypeElement`
-    return (type as? RustPathTypeElement)?.path?.referenceName
 }
--- a/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplItemImplMixin.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplItemImplMixin.kt
@@ -4,7 +4,9 @@ import com.intellij.lang.ASTNode
 import com.intellij.psi.stubs.IStubElementType
 import org.rust.ide.icons.RustIcons
 import org.rust.lang.core.psi.RustImplItemElement
+import org.rust.lang.core.psi.RustPathTypeElement
 import org.rust.lang.core.psi.impl.RustStubbedElementImpl
+import org.rust.lang.core.psi.referenceName
 import org.rust.lang.core.stubs.elements.RustImplItemElementStub
 import javax.swing.Icon
 
@@ -16,4 +18,10 @@ abstract class RustImplItemImplMixin : RustStubbedElementImpl<RustImplItemElemen
     override fun getIcon(flags: Int): Icon = RustIcons.IMPL
 
     override val isPublic: Boolean get() = false // pub does not affect imls at all
+}
+
+
+val RustImplItemElement.baseTypeName: String? get() {
+    //TODO: move to `RustTypeElement`
+    return (type as? RustPathTypeElement)?.path?.referenceName
 }
--- a/src/main/kotlin/org/rust/lang/core/stubs/elements/RustImplItemElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/elements/RustImplItemElement.kt
@@ -4,30 +4,41 @@ package org.rust.lang.core.stubs.elements
 import com.intellij.psi.stubs.*
 import org.rust.lang.core.psi.RustImplItemElement
 import org.rust.lang.core.psi.impl.RustImplItemElementImpl
+import org.rust.lang.core.psi.impl.mixin.baseTypeName
 import org.rust.lang.core.stubs.RustElementStub
 import org.rust.lang.core.stubs.RustStubElementType
+import org.rust.lang.core.stubs.index.RustInherentImplIndex
 
 
 object RustImplItemStubElementType : RustStubElementType<RustImplItemElementStub, RustImplItemElement>("IMPL_ITEM") {
     override fun createStub(psi: RustImplItemElement, parentStub: StubElement<*>?): RustImplItemElementStub =
-        RustImplItemElementStub(parentStub, this)
+        RustImplItemElementStub(parentStub, this, psi.baseTypeName)
 
     override fun createPsi(stub: RustImplItemElementStub): RustImplItemElement =
         RustImplItemElementImpl(stub, this)
 
     override fun deserialize(dataStream: StubInputStream, parentStub: StubElement<*>?): RustImplItemElementStub =
-        RustImplItemElementStub(parentStub, this)
+        RustImplItemElementStub(parentStub, this, dataStream.readName()?.string)
 
     override fun serialize(stub: RustImplItemElementStub, dataStream: StubOutputStream) = with(dataStream) {
+        writeName(stub.baseTypeName)
     }
 
     override fun indexStub(stub: RustImplItemElementStub, sink: IndexSink) {
+        stub.baseTypeName?.let {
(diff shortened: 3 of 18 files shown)


matklad:
Should be ready for a proper review now.

alexeykudinkin:
It looks fine, except that we'd not going to index type by names -- solution simply isn't scalable.

matklad:
>we'd not going to index type by names -- solution simply isn't scalable.

What are the alternatives?

alexeykudinkin:
Types of course.

matklad:
I believe that indexing using resolve info (and indexing by resolved types naturally requires resolve) is not scalable. The amount of work (both CPU cycles and lines of code) required for proper invalidation would be horrendous. At least I don't see any simple and efficient way to implement such indexing by types. 

At the moment `RustModulesIndex` uses essentially the same strategy of indexing by names + resolving to kill false positives.

We can of course add an additional layer of `PSI_MODIFICATION_COUNT` caching to the mix. In fact this is already done, because `inherentImpls` is calculated once per type, and types themselves are cached.

alexeykudinkin:
> I believe that indexing using resolve info (and indexing by resolved types naturally requires resolve) is not scalable. The amount of work (both CPU cycles and lines of code) required for proper invalidation would be horrendous. At least I don't see any simple and efficient way to implement such indexing by types.

Why is that? How's that different from indexing by types?

matklad:
Consider the following example:

```Rust
fn main() {
    let s = ::a::b::c::S;
    s.foo()
}

pub mod a {
    pub mod b {
        pub mod c {
            
            pub struct S;
            
        }
    }
}

pub mod d {
    pub mod e {
        pub mod f {
            
            impl ::a::b::c::S {
                pub fn foo(&self ) { println!("Hello"); }
            }
            
        }
    }
}
```

We need to somehow associate that impl with that `S` struct.

One approach is to index by resolved types. That is, during indexing `impl ::a::b::c::S` resolve the path to the particular PSI element `S` and use it as a key in the index. I don't think that this is a workable solution, because a change in any of the modules may invalidate the resolve. 

Another option is to index by unresolved types (that is, the `::a::b::c::S` path itself). It also does not seem  workable, because different unresolved types may resolve to the same type. For example, the path could have been spelled as `super::super::super::a::b::c::S` instead. Devising absolute path requires resolve, which brings us back to the problems of the first approach. Also note that impls like `impl &S`, `impl (S, S)` are invalid. 

What's left is indexing by the *name* `S`, and then verifying the result by doing the resolve. If you index by names, you don't need to do any invalidation, because you don't care about module structure at all. I agree that the lookup is more costly then we would ideally like, but it seems to be the necessary cost to avoid the prohibitively expensive invalidation.

We can mitigate lookup costs by employing additional `PSI_MODIFICATION_COUNT` tracking. For the libraries, it could be extended to `CRATE_MODIFICATION_COUNT` if we implement that. Another option is to leverage orphan rules and add a crate name to the index key.

alexeykudinkin:
Doing the resolve during every cache query is no less performance-prohibitive. 

We're not going index impls by type-names because a) employing the scheme with 'typename'-key and then filtering out invalid results is asymptotically equal (and common-sensely isomorphic too) to the resolved-type keys being used, b) it isn't scalable in a sense we couldn't generalise that on generics (which is an obvious next step).

matklad:
>a) employing the scheme with 'typename'-key and then filtering out invalid results is asymptotically equal (and common-sensely isomorphic too) to the resolved-type keys being used

No, I don't think so. In short, indexing by types requires inspecting *all* impls, while indexing by names allows to inspect only a much smaller subset of potentially relevant impls. Let's break down the costs.


|  | types |names|
| --- | --- |---|
| Initial indexing cost | N * R | N |
| Invalidation after PSI change| **N * R** | 0 |
| first lookup of impl by type | 1 | R * (1 + F) |
|subsequent lookups | 1 | 1 |complexity

where
  N -- total number of impls, I assume it to be large and proportional to the crate size
  R -- a cost of a single resolve procedure
  F -- the number of different types which share the same name, I assume it to be effectively constant, especially within a single crate. 

So, for the typical "resolve impls of k types in the file after an edit" workload, by names index will need roughly `k * R * (1 + F)` work, which is proportional to the size of file being edited. Indexing by types, on the other hand, will require `k + N * R` work, which is insane, because it is proportional to the total size of the project being edited. 

Or is there a clever way to avoid that **N * R** invalidation cost? I don't think there is, because any module in Rust can potentially depend on any other.


Also, I would argue that we already effectively have "by type" impls cache because the types themselves are cached, and `impls` are a lazy property of the type.

alexeykudinkin:
You solution is prone to one inevitable `false negative` -- it misses _aliases_ completely.

matklad:
Agree, in its current form it indeed misses aliases. The solution for the alias problem won't be to difficult though: we can have an index of aliased names (`Map<String, String>`), use it to collect all potential aliases, and query `impl-by-name` index using the original and aliased names. 

I don't think that we should implement this alias index right away as it seems to be a rare edge case (in comparison with a ton of stuff we don't handle yet). And maybe we'll come up with a more appropriate indexing scheme once we implement more typing around generics and traits.

alexeykudinkin:
And that index would be essentially a _resolved_ one, bringing all the hassle of validation/invalidation procedures inherent for the resolved indices.

matklad:
Nope, it would be an unresolved one. When visiting `foo as bar`, you would remember a mapping from `foo` string to `bar` string, without resolving anything. This will work because you don't need precise aliasing info, over approximation would do.

alexeykudinkin:
Then your time-bound approximations are simply incorrect.

matklad:
They were given for the index without aliases, because I don't think that handling of aliases is important at this stage :) 

But the presence of aliases should not change anything significantly. Aliases have the same significance as the `F` term: they introduce additional number of lookups for the minority of types.

matklad:
I'd rather merge this likely suboptimal implementation now if we don't have an obviously better solution. We'll need to implement the traits system to obtain the understanding of what indexes exactly do we need, so this index will probably change anyway.

alexeykudinkin:
What are the reasons to merge it for then?

matklad:
It correctly determines most of the inherent impls ([new tests]) and forms the basis for a more advanced non inherent impl index. 

[new tests]: https://github.com/intellij-rust/intellij-rust/pull/508/files#diff-413b89b900967e7ebd033311df9a2fc1R1

alexeykudinkin:
I don't see how it fits the basis for a _proper_ impl-index:

  > It isn't generalisable to traits and generics.
  > it exposed to the obvious aliasing problems.

This seems to me more of experiment-like endeavour and it isn't ready to be shipped yet. Let's establish rule of thumb that we're either merging _complete_ features or at least _partially_ complete ones (solving valuable partition of the problem and extensible in a clear way).

matklad:
Hm, looks like Kotlin uses similar by name index for extension functions: https://github.com/JetBrains/kotlin/blob/master/idea/idea-analysis/src/org/jetbrains/kotlin/idea/stubindex/KotlinTopLevelExtensionsByReceiverTypeIndex.kt

matklad:
And here's how they handle aliases: https://github.com/JetBrains/kotlin/blob/master/idea/idea-analysis/src/org/jetbrains/kotlin/idea/stubindex/IndexUtils.kt#L56

matklad:
Though it is used only for completion and auto imports, and not for the resolve. In Kotlin you must import extensions explicitly, so in general you know where to look for them. In contrast in Rust `impl`s may be situated anywhere, and you don't have any hints about their location.

alexeykudinkin:
> Hm, this is inherent impls index. It should not deal with traits. Inherent and non inherent impls are rather different beasts, and can't be handled with the same code.

Why is that? Don't quite see any obstacles (and any reasons actually to have the opposite) to have single index facility and let user discriminate for itself whether it looks for inherent or not.

> Which do not matter that much in practice in comparison with other outstanding issues. And they are perfectly solvable with the help of the alias index, which is the instance of "extensible in a clear way".

> I would say that this implementation is more complete than the current one :)

It may be more complete, though still poorly designed (to my taste).

matklad:
>Why is that? Don't quite see any obstacles (and any reasons actually to have the opposite) to have single index facility and let user discriminate for itself whether it looks for inherent or not.

No contradiction here: these indices can have the same interface, but they  should have different implementations. At the moment I think that for non-inherent impls we want to index by traits and not by types, to be able to handle cases like `impl<T> Foo for  T {}` and to be able to filter impls by trait (we want to keep only the impls for the traits which are currently in scope). Hm, maybe it makes sense to handle only generic impls by another index? That is, to have "generic" (by traits) and "non generic" ("by types") indices, instead of "inherent"(by types) and "non inherent"(by traits)


>It may be more complete, though still poorly designed (to my taste).

What would be a better design?

alexeykudinkin:
> No contradiction here: these indices can have the same interface, but they should have different implementations. At the moment I think that for non-inherent impls we want to index by traits and not by types, to be able to handle cases like impl<T> Foo for T {} and to be able to filter impls by trait (we want to keep only the impls for the traits which are currently in scope). Hm, maybe it makes sense to handle only generic impls by another index? That is, to have "generic" (by traits) and "non generic" ("by types") indices, instead of "inherent"(by types) and "non inherent"(by traits)

Still hardly see any reason to have more than but a single index. I'll scaffold my own vision to see how it goes.

matklad:
If we want to move in the direction of the resolve-depending indexing, incremental compilation [RFC] might be of interest. Though they start form parsing the whole crate as a first step of incremental compilation. 

[RFC]: https://github.com/rust-lang/rfcs/blob/master/text/1298-incremental-compilation.md

matklad:
I've repurposed the inherent impl index to be the geneneral impl index. Should be ok to merge now?

PR #516, opened 2016-07-07
[WIP] Register Module Source Roots To Create Project from IDEA

06needhamt:
Work In Progress Do Not Merge !!!

### DIFF
--- a/src/main/kotlin/org/rust/ide/idea/RunWriteActionRunnable.kt
+++ b/src/main/kotlin/org/rust/ide/idea/RunWriteActionRunnable.kt
@@ -1,0 +1,45 @@
+package org.rust.ide.idea
+
+import com.intellij.openapi.module.Module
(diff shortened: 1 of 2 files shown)


matklad:
Hm, we already have a thing for detecting project structure: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/idea/RustProjectStructureDetector.kt.

I wonder if we could somehow invoke `cargo new` when creating a new project, and then just reuse `RustProjectStructureDetector`.

matklad:
Turns out that the simplest way to create content roots was inheriting from JavaModuleBuilder: https://github.com/intellij-rust/intellij-rust/commit/968e29095428a092fb723a3ac7ed0420ee4b0eda

06needhamt:
@matklad Everything Works Fine Now

matklad:
Good to hear, thanks for conformation!

PR #517, opened 2016-07-07
Pass rustc location via RUSTC when invoking cargo

madisp:


matklad:
Will this potentially break installations where `rustc` and `cargo` are not in the same directory, but `rustc` is in `PATH`? Are there any such layouts? @jajakobyly is it correct that on windows `rustc` and `cargo` reside in the same directory?

src/main/kotlin/org/rust/cargo/commands/Cargo.kt
```diff
@@ -53,6 +55,7 @@ class Cargo(
             .withParameters(command)
             .withParameters(additionalArguments)
             .withEnvironment(environmentVariables)
+            .withEnvironment(CargoConstants.RUSTC_ENV_VAR, pathToRustExecutable)
```
matklad: I think this should come before `withEnvironment(environmentVariables)` to be able to override `RUSTC`.

madisp: Doesn't really matter, `withEnvironment(Map)` does a `putAll()` and `withEnvironment(String, String)` does a put.

madisp: Will change, how do you usually do this, with a separate commit or a force push to the branch?

matklad: Does not matter much, this is one commit length pull request, so "squash merge" would work anyway.

matklad:
I think that this should not break anything (will check windows story), but ideally we should provide an *optional* "rustc path" settings in the GUI. And teaching `cargo` not to fail if there is no `rustc` is a also good idea: https://github.com/rust-lang/cargo/issues/2833

matklad:
Thanks!

jajakobyly:
> @jajakobyly is it correct that on windows rustc and cargo reside in the same directory? 

yup

PR #528, opened 2016-07-14
(FMT): fix spacing in method calls

### LINKED ISSUE #526: Fix spacing in method calls
Test case:

```
fn main() {
    foo::<i32>       ();
    foo.foo::<i32>   ();
}
```

Goal:

```
fn main() {
    foo::<i32>();
    foo.foo::<i32>();
}
```

Current formatting: 

```
fn main() {
    foo::<i32>();
    foo.foo::<i32>   ();
}
```

cc @jajakobyly

jajakobyly:
Fixes #526


src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
```diff
@@ -83,6 +83,7 @@ fun createSpacingBuilder(commonSettings: CommonCodeStyleSettings, rustSettings:
         .between(IDENTIFIER, GENERIC_ARGS).spaceIf(false)
         .between(IDENTIFIER, ARG_LIST).spaceIf(false)
         .between(GENERIC_PARAMS, PARAMS_LIKE).spaceIf(false)
+        .between(GENERIC_ARGS, ARG_LIST).spaceIf(false)
         .beforeInside(ARG_LIST, CALL_EXPR).spaceIf(false)
```
matklad: Can both method_call and call_expr be handled by the same rule?

jajakobyly: Unfortunately, no. These guys from test have totally different PSI.

`SpacingBuilder`'s rules apply only to neighbouring siblings in the AST. That's why this didn't work for method call's.
idea_2016-07-14_21-01-02

matklad: Something like

```
      .before(ARG_LIST).spaceIf(false)
```

perhaps?

jajakobyly: yeah :sweat_smile:

matklad:
Thanks!

PR #546, opened 2016-07-23
(FMT): fix #543

### LINKED ISSUE #543: Formatting issues
Hey, the plugin looks fantastic so far!
I've tried auto format (ctrl+alt+l) on a couple of files and it got mostly everything correct but stumbled on some things:

```patch
-pub type TeraResult<T> = Result<T, TeraError>;
+pub type TeraResult<T>= Result<T, TeraError>;
```

The one below is a bit contention as both might be correct but the "formatted" version breath less, maybe don't change macro invocation  since they can take many aspects?
```patch
-impl_rdp! {
+impl_rdp!{
```

Now the last one is even more contentious and i see why the formatter did it but is there a way to disable that behaviour somehow? It drifts to the right too much for me
```patch
                     if previous_end < token.start {
                         space_tokens.push((
-                            insert_at,
-                            Token::new(Rule::text, previous_end, token.start)
-                        ));
+                                              insert_at,
+                                              Token::new(Rule::text, previous_end, token.start)
+                                          ));
                     }

```

jajakobyly:
fix #543

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
@@ -8,6 +8,29 @@ import org.rust.lang.core.psi.RustTokenElementTypes.DOT
 fun RustFmtBlock.getAlignmentStrategy(): RustAlignmentStrategy = when (node.elementType) {
     TUPLE_EXPR, ARG_LIST, FORMAT_MACRO_ARGS, TRY_MACRO_ARGS ->
         RustAlignmentStrategy.wrap()
+            .alignIf { child, parent, ctx ->
+                // Do not align if we have only one argument as this may lead to
+                // some quirks when that argument is tuple expr.
+                // Alignment do not allow "negative indentation" i.e.:
+                //     func((
+                //         happy tuple param
+                //     ))
+                // would be formatted to:
+                //     func((
+                //              happy tuple param
+                //          ))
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
```diff
@@ -8,6 +8,29 @@ import org.rust.lang.core.psi.RustTokenElementTypes.DOT
...
+                // Alignment do not allow "negative indentation" i.e.:
+                //     func((
+                //         happy tuple param
+                //     ))
+                // would be formatted to:
+                //     func((
+                //              happy tuple param
+                //          ))
+                // because due to applied alignment, closing paren has to be
+                // at least in the same column as the opening one.
```
jajakobyly: This one was tough. I hope that my reasoning what this has happened is correct :sweat_smile:

matklad: Looks reasonable, though maaybe there is something more general?

Like, how

```
foo(1, (
    1,
    2,
    3
))
```

Should be formatted?

matklad:
Thanks!

src/main/kotlin/org/rust/ide/formatter/impl/alignment.kt
jajakobyly: Ugh I don't know. Personally I would indent this one

PR #554, opened 2016-07-26
(FMT): allow #[macro_use] attr to be placed in line with extern crate item

### LINKED ISSUE #548: Allow #[macro_use] attr to be placed in line with extern crate item
subject

jajakobyly:
fixes #548

### DIFF
--- a/src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
+++ b/src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
@@ -1,0 +1,4 @@
+#[macro_use] extern crate foo;
+#[macro_use] extern crate bar;
+#[macro_use] // comment
+extern crate moo;
--- a/src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
+++ b/src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
@@ -2,3 +2,5 @@
 #[macro_use] extern crate bar;
 #[macro_use] // comment
 extern crate moo;
+
+#[macro_use] mod foo;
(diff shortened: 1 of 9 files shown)


src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
```diff
@@ -0,0 +1,4 @@
+#[macro_use] extern crate foo;
```
matklad: So we always place `macro_use` inline? I think this is a bit too opinionated: some do this, some don't:

https://github.com/search?utf8=%E2%9C%93&q=%23%5Bmacro_use%5D&type=Code&ref=searchresults

It's better to either have an option here, or to be able to preserve existing formatting.

matklad:
Nice test: https://github.com/cuviper/rust-libprobe/blob/c889fce367aabc78fcd7f347eea7fff569fb6268/examples/loop.rs

src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
jajakobyly: And what should be the default?

matklad:
Also, was there `#[macro_use] mod foo;` as well?

jajakobyly:
> Nice test: https://github.com/cuviper/rust-libprobe/blob/c889fce367aabc78fcd7f347eea7fff569fb6268/examples/loop.rs

I see this as bad formatting; the same thinks rustfmt: https://is.gd/Qh5FCz -> click `Format`

jajakobyly:
> Also, was there #[macro_use] mod foo; as well? 

ups

src/test/resources/org/rust/ide/formatter/fixtures/macro_use_after.rs
jajakobyly: Ok I set inlining `#[macro_use]` as the default.

matklad:
Thanks!

Issue #371, opened 2016-05-04
Resolve of primitive types


matklad:
We should do something about primitive types, like `i32`, in resolve. One way is to resolve them to the corresponding module in stdlib. Another way is to return `null` from `getReference` for primitives.

matklad:
>One way is to resolve them to the corresponding module in stdlib.

`bool` does not have the module

jajakobyly:
I think, if it's possible, we could resolve them to some virtual elements located in custom prelude (something like `struct i32;`). I'm not sure, but PyCharm may do something like this for builtins.

Resolving to modules may put us in a situation when we won't be able to differentiate primitive from module.

matklad:
PyCharm uses skeletons: https://github.com/JetBrains/python-skeletons/blob/master/__builtin__.py

GoPlugin resolve to https://golang.org/src/builtin/builtin.go

jajakobyly:
```go
type uint8 uint8
```
This is pure cool

kumbayo:
Is this Issue related to code completion for primitive types or to make Ctrl-Click on a primitive type work?

I ask because i am currently working on auto completion for primitive types and i am not sure if i should open a new issue or work with this Issue.

farodin91:
@kumbayo Any process on primitive type completion?

kumbayo:
Yes I have some changes for improved primitive type completion.
But I need to rebase them on the current version first and they also need to be reviewed.

Issue #357, opened 2016-04-26
Snippets from other editors


RustSnippets:
https://github.com/zargony/atom-language-rust/blob/master/snippets/rust.cson
https://github.com/abaez/ta-rust/blob/master/snippets.lua
https://github.com/freebroccolo/rust-snippets/blob/master/rust-snippets.el
https://github.com/darookee/vim-snippets/blob/master/UltiSnips/rust.snippets
https://github.com/rust-lang/sublime-rust/tree/master/snippets

Maybe you can derive some ideas from other editors and implement it in your project.

matklad:
Our snippets are here (warning, XML! ) https://github.com/intellij-rust/intellij-rust/blob/master/src/main/resources/org/rust/ide/liveTemplates/Rust.xml .

jajakobyly:
Ok, "taking" this.

PR #561, opened 2016-07-28
Surrounders

jajakobyly:
This PR reworks surrounders infrastructure. Rather than relying on live templates, they are based on `Surrounder` interface now - which means they are: more accurate, more capable, more intelligent etc etc. Also features some testing infra, which generalised, might be useful e.g. for postfix completion.

Planned surrounders for this PR:
 - Expression:
    - [x] `(expr)`
    - [x] `!(expr)` (not available everywhere where applicable, because of missing typing, see #371)
 - Statements:
    - [x] `{ }`

**EDIT:** Ok changed my mind about loops. Surrounders were bad idea for them, rather then rewriting these snippets they should be just polished.

Highly inspired on Java, Kotlin, Go, Python surrounders (sorted by level of influence)

cc #357 
cc @RustSnippets


src/main/kotlin/org/rust/ide/surroundWith/RustExpressionSurrounderBase.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustExprElement
+import java.util.*
+
+abstract class RustExpressionSurrounderBase : Surrounder {
+    abstract fun isApplicable(expression: RustExprElement): Boolean
+    abstract fun surroundExpression(project: Project, editor: Editor, expression: RustExprElement): TextRange?
+
+    final override fun isApplicable(elements: Array<out PsiElement>): Boolean {
+        if (elements.size != 1 || elements[0] !is RustExprElement) {
```
matklad: `elemente.singleOrNull() as? RustExprElement`

src/main/kotlin/org/rust/ide/surroundWith/RustExpressionSurrounderBase.kt
```diff
@@ -0,0 +1,34 @@
...
+
+        val expression = elements[0] as RustExprElement
+
+        // TODO: Some additional filtering may be required.
+
+        return isApplicable(expression)
+    }
+
+    final override fun surroundElements(project: Project, editor: Editor, elements: Array<out PsiElement>): TextRange? {
+        require(elements.size == 1 && elements[0] is RustExprElement) {
```
matklad: I prefer to extract checking and retrieving logic to a single function ( `fun find(Context): Element?` ). That way I won't forget to update code in two places.

src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurrounder.kt
```diff
@@ -0,0 +1,24 @@
...
+
+abstract class RustStatementsSurrounder : Surrounder {
+    abstract fun surroundStatements(
+        project: Project,
+        editor: Editor,
+        container: PsiElement,
+        statements: Array<out PsiElement>
+    ): TextRange?
+
+    override fun isApplicable(elements: Array<out PsiElement>): Boolean = elements.size > 0
```
matklad: femtonit: `elements.isNotEmpty()`

src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurrounder.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    override fun surroundElements(project: Project, editor: Editor, elements: Array<out PsiElement>): TextRange? {
+        require(elements.size > 0)
+        val container = elements[0].parent ?: throw IllegalStateException()
```
matklad: `requireNotNull`

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
```diff
@@ -0,0 +1,159 @@
...
+ * `&lt;&nbsp;&nbsp;&nbsp;expression&nbsp;&nbsp;&nbsp;&gt;`.
+ */
+fun findExpressionInRange(file: PsiFile, startOffset: Int, endOffset: Int): RustExprElement? {
+    val (element1, element2) = file.getElementRange(startOffset, endOffset) ?: return null
+
+    // Get common expression parent.
+    var parent = PsiTreeUtil.findCommonParent(element1, element2) ?: return null
+    if (parent !is RustExprElement) {
+        parent = parent.parentOfType<RustExprElement>() ?: return null
+    }
```
matklad: You can use `strict = false` instead of the `if`.

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
```diff
@@ -0,0 +1,159 @@
...
+    return null
+}
+
+// FIXME: Maybe items are ok?
+// FIXME: What about macros?
+/**
+ * Finds statements (mostly [RustStmtElement]s, [PsiComment]s and [RustExprElement]
+ * as return expression, doesn't allow attrs and items) within selected range.
+ */
+fun findStatementsInRange(file: PsiFile, startOffset: Int, endOffset: Int): Array<out PsiElement> {
```
matklad: Why not a `List<out PsiElement>`?

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
```diff
@@ -0,0 +1,159 @@
...
+    if (parent == this) return this
+
+    var element = this
+    while (parent != element.parent) {
+        element = element.parent
+    }
+    return element
+}
+
+private fun collectElements(start: PsiElement, stop: PsiElement?, pred: (PsiElement) -> Boolean): Array<out PsiElement> {
```
matklad: FYI: a similar piece of code is here: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt#L60

src/test/kotlin/org/rust/ide/surroundWith/RustSurrounderTestCaseBase.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.intellij.openapi.command.WriteCommandAction
+import com.intellij.util.readText
+import org.assertj.core.api.Assertions.assertThat
+import org.rust.lang.RustLanguage
+import org.rust.lang.RustTestCaseBase
+import java.nio.file.Paths
+import java.util.*
+
+abstract class RustSurrounderTestCaseBase(val surrounder: Surrounder) : RustTestCaseBase() {
+    override val dataPath: String get() = "org/rust/ide/surroundWith/fixtures/$testClassName"
```
matklad: :+1:

src/test/kotlin/org/rust/ide/surroundWith/RustSurrounderTestCaseBase.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    protected val testClassName: String =
+        surrounder.javaClass.simpleName.replace(Regex("^RustWith(\\w+)Surrounder$"), "$1").decapitalize()
```
matklad: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/to-regex.html

src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurrounder.kt
jajakobyly: Does it enable smart casts? In this case this will work, because `getParent()` isn't annotated

src/test/kotlin/org/rust/ide/surroundWith/RustSurrounderTestCaseBase.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    protected fun doTest(isApplicable: Boolean = true) {
```
matklad: Perhaps create two different functions instead of one with a boolean parameter, which is always a compile time constant?

src/main/kotlin/org/rust/ide/surroundWith/RustExpressionSurrounderBase.kt
jajakobyly: this guy is only assertion

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
jajakobyly: API requirement of `SurroundDescriptor#getElementsToSurround()`. This array must be created anyway

src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurrounder.kt
matklad: `val container = requireNotNull(elements[0].parent)`

matklad:
Awesome work! 


>(Question: will this close this issue?) #357 

I think not yet, things like `match`, `if`, `fn` are not covered.

src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurrounder.kt
jajakobyly: wow, didn't know it returns value :sweat_smile:

src/main/kotlin/org/rust/ide/surroundWith/RustExpressionSurrounderBase.kt
matklad: Which proves that `elements[0] as RustExprElement` on the next line is valid. This is exactly my point: if you don't separate "look" and "leap", they can't get out of sync. 

If you define

```
fun target(elements: Array<out PsiElement>): RustExprElement? =
    elements.singleOrNull() as? RustExprElement
```

you would be able to reuse it in `isApplicable` and in `surroundElements`.

jajakobyly:
travis wtf

matklad:
Idea 15 has failed. Perhaps some API has changed?

jajakobyly:
Yeah, using Kotlin extensions from SDK. Might be the cause

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
jajakobyly: similar, but different :smile: 

I tried to introduce some small abstraction over range of PSI elements, but they are not comparable, so we are not able to use Kotlin's ranges :(

jajakobyly:
rtm

matklad:
Thanks!

PR #564, opened 2016-07-28
Improve snippets

### LINKED ISSUE #357: Snippets from other editors

jajakobyly:
As title says, this PR is going to greatly extend and improve our snippets collection.

fix #357
cc @rustsnippets


src/main/kotlin/org/rust/ide/template/macros/utils.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.rust.lang.core.psi.RustPatBindingElement
+import org.rust.lang.core.resolve.RustResolveEngine
+
+fun getPatBindingsVisibleAt(place: PsiElement?): List<RustPatBindingElement> {
+    if (place == null) return emptyList()
+    val pivot = place as? RustCompositeElement
+    return RustResolveEngine.enumerateScopesFor(place)
+        .flatMap { RustResolveEngine.declarations(it, pivot) }
+        .mapNotNull { it as? RustPatBindingElement }
+        .toList()
```
jajakobyly: Ok, so this has two issues which need to be fixed before merging:
 - The requirement for pivot to be `RustCompositeElement` cannot be fulfilled in most situations, because place is mostly `PsiWhitespace`. Can this limitation be removed?
 - This is *incredibly* slow

matklad: > Can this limitation be removed?

Yes. 

> This is incredibly slow

Well, you are asking for all visible symbols (even for the private ones) :) I think that you can add a `takeWhile` after `enumerateScopesFor` to consider only local scopes.

src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt
```diff
@@ -532,7 +534,7 @@ private fun PsiDirectory.findFileByRelativePath(path: String): PsiFile? {
  * Helper to debug complex iterator pipelines
  */
 @Suppress("unused")
-private fun<T> Sequence<T>.inspect(f: (T) -> Unit = { println("inspecting $it") }): Sequence<T> {
+private fun <T> Sequence<T>.inspect(f: (T) -> Unit = { println("inspecting $it") }): Sequence<T> {
     return map { it ->
         f(it)
         it
```
jajakobyly: ouch, I have accidentally fired formatter on whole file

src/main/kotlin/org/rust/ide/template/macros/utils.kt
matklad: And why are you calling `toList`? Can you work with the `Sequence` instead?

jajakobyly: I will be iterating over it until suitable letter is found

src/main/kotlin/org/rust/ide/template/macros/RustSuggestIndexNameMacro.kt
```diff
@@ -0,0 +1,38 @@
...
+        if (params.isNotEmpty()) return null
+
+        val project = context.project
+        val offset = context.startOffset
+        val document = context.editor?.document ?: return null
+        val file = PsiDocumentManager.getInstance(project).getPsiFile(document) ?: return null
+        val place = file.findElementAt(offset)
+        val pats = getPatBindingsVisibleAt(place)
+
+        chooseLetterLoop@ for (letter in 'i'..'z') {
```
matklad: `('i' .. 'z').find`

src/main/kotlin/org/rust/ide/template/macros/RustSuggestIndexNameMacro.kt
```diff
@@ -0,0 +1,38 @@
...
+            for (pat in pats) {
```
matklad: This is `O(N)`. It can be `O(1)` if you create a set of names.

src/main/kotlin/org/rust/ide/template/macros/RustSuggestIndexNameMacro.kt
```diff
@@ -0,0 +1,38 @@
...
+                val name = pat.name ?: continue
+                if (name.singleOrNull() == letter) {
```
matklad: Won't plain `name == letter` work here?

src/main/kotlin/org/rust/ide/template/macros/RustSuggestIndexNameMacro.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import com.intellij.codeInsight.template.*
+import com.intellij.psi.PsiDocumentManager
+import org.rust.ide.template.RustContextType
+
+class RustSuggestIndexNameMacro : Macro() {
+    override fun getName(): String = "rustSuggestIndexName"
+    override fun getPresentableName(): String = "rustSuggestIndexName()"
+
+    override fun getDefaultValue(): String = "a"
```
matklad: `i` is more conventional

src/main/kotlin/org/rust/ide/template/macros/RustSuggestIndexNameMacro.kt
jajakobyly: yeah, planned to do manual checks if pattern is defined after place but it's not necessary

src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt
matklad: Yeah, let's not do this yet and wait for official kotlin styleguide

src/main/kotlin/org/rust/ide/template/macros/RustSuggestIndexNameMacro.kt
jajakobyly: it's never used from what I know

src/main/kotlin/org/rust/ide/template/macros/utils.kt
```diff
@@ -0,0 +1,27 @@
...
+            // we are only interested in local scopes
+            if (it is RustItemElement) {
+                // workaround diamond inheritance issue
+                // (ambiguity between RustItemElement.parent & RustResolveScope.parent)
+                val item: RustItemElement = it
+                item.parent is RustBlockElement
+            } else {
+                it !is RustFile
+            }
+        }
```
jajakobyly: Hmm, I think this `takeWhile` may be moved to `RustResolveEngine` as it might be useful generally

jajakobyly:
Ok, other interesting macros will require advanced typification, so that's it as for now :)

ereichert:
I'm the person who posted the snippets question on the user forum. (https://users.rust-lang.org/t/intellij-idea-live-templates/6727)

As I said on the forum, I'm not sure some of my templates belong in the plugin repo.

For instance, I have a few that I use constantly and probably would be great in the plugin repo.

<template name=".at" value=".and_then(|$PARAMS$| {&#10;    $END$&#10;})" description="and then with closure" toReformat="false" toShortenFQNames="true">
  <variable name="PARAMS" expression="" defaultValue="" alwaysStopAt="true" />
  <context>
    <option name="RUST_FILE" value="true" />
  </context>
</template>

<template name=".map" value=".map(|$PARAMS$| {&#10;    $END$&#10;})" description="map with closure" toReformat="false" toShortenFQNames="true">
  <variable name="PARAMS" expression="" defaultValue="" alwaysStopAt="true" />
  <context>
    <option name="RUST_FILE" value="true" />
  </context>
</template>

And then there are some that might only sooth my annoyance.

<template name="unimp" value="unimplemented!()$END$" description="unimplemented!()" toReformat="false" toShortenFQNames="true">
  <context>
    <option name="RUST_FILE" value="true" />
  </context>
</template>

<template name="dd" value="#[derive(Debug)]" description="derive debug" toReformat="false" toShortenFQNames="true">
  <context>
    <option name="RUST_FILE" value="true" />
  </context>
</template>

But I want to be a good open source citizen.  So...

How should I propose templates without having to do a full PR?

jajakobyly:
Well, honestly I believe all this templates should be "done" by code completion, not snippets. 
I understand that they save some typing, but they feel non-intellijish to me.

matklad:
@ereichert perfect, this is exactly the sort of feedback we need. As @jajakobyly said, we might be able to do something smarter then snippets here. 

And I think we have something better for `dd` case already? Do you you know about "add derive intention"? When your cursor is at some `struct` or `enum`, you can press <kbd>Alt+Enter</kbd> to invoke this intention and it will add `#[derive()]` clause (if there is any one yet) and place the cursor right between parenthesis (here is a gif from our docs: https://intellij-rust.github.io/features/#add-derive). Do you use this intention? 


As for `map` and `and_then`, I think they should already be suggested as completions? We need append `()` on completing them though. Not sure how we can handle lambdas though, any ideas?

jajakobyly:
> Not sure how we can handle lambdas though, any ideas?

I think they should be not handled at all (just like in any existing language plugin) , little added value imho

ereichert:
I did not know about the intention.  Was that on the site when the new site was announced?  If it was, I missed it and I haven't been back to the site since the day of the announcement.  I guess I need to check back with the site occasionally.

I agree that everything I've suggested should be done with code completion.  And I'm looking forward to the day code completion is working for all the nooks and crannies.  What you have built thus far is great.  I can only imagine what development is going to be like when the plugin is feature complete.

However, the reason I wrote most of the templates I use is because the code completion isn't feature complete at the moment.  I think the templates can fill in until it is.  It's an easy, quick way to fill in some gaps while the hard work is done.  Having to type out .to_string() at the end of every "afd;lkj" was driving me insane.

And `map` and `and_then` are not being suggested in the codebase I'm working on right now.  I'm calling BTreeMap::get which returns an Option and I'm not getting any suggestions for the Option.

jajakobyly:
I think your snippets might be placed in some `Pro tips` (or how to spell this correctly) section in our docs :)

jajakobyly:
> Was that on the site when the new site was announced?

Nope. Ouch, I'm starting to think that I have posted this site too early on reddit/twitter 😅

ereichert:
> I think your snippets might be placed in some Pro tips (or how to spell this correctly) section in our docs :)

Yah.  I get that.  That was kind of the point I was trying to make with the forum post.

I'll take a look at the docs and see if there's something I can do there.  Nice timing with the docs day coming up.

It could serve as a nice check list for completions that need to be written also.  As the completion gets more feature complete you can remove the templates from the docs.

ereichert:
PS - There are two (soon to be three) developers on my team writing Rust for a living.  We've all switched to the this plugin from VS Code because of how far it has come.

src/main/kotlin/org/rust/ide/template/RustContextType.kt
```diff
@@ -2,41 +2,74 @@ package org.rust.ide.template
...
+import kotlin.reflect.KClass
 
-
-sealed class RustContextType(id: String, presentableName: String, parentContext: Class<out TemplateContextType>)
-    : TemplateContextType(id, presentableName, parentContext) {
-
+sealed class RustContextType(
+    id: String,
+    presentableName: String,
+    baseContextType: KClass<out TemplateContextType>
```
matklad: `KClass` is neat!

src/main/resources/org/rust/ide/liveTemplates/other.xml
```diff
@@ -0,0 +1,59 @@
...
+        </context>
+    </template>
+
+    <template name="main" description="main() function declaration" toReformat="true" toShortenFQNames="true"
+              value="fn main() {&#10;    $END$&#10;}">
+        <context>
+            <option name="RUST_FILE" value="true"/>
+            <option name="RUST_STATEMENT" value="false"/>
+            <option name="RUST_ITEM" value="false"/>
+            <option name="RUST_MOD" value="true"/>
```
matklad: Should it really be applicable inside the inner module?

src/main/resources/org/rust/ide/liveTemplates/output.xml
```diff
@@ -0,0 +1,56 @@
...
+    <template name="pr" description="print!" toReformat="true" toShortenFQNames="true"
+              value="print!(&quot;$END$&quot;);">
+        <context>
+            <option name="RUST_STATEMENT" value="true"/>
+        </context>
+    </template>
+
+    <template name="pd" description="debug println!" toReformat="true" toShortenFQNames="true"
+              value="println!(&quot;$FORMAT$&quot;, $END$);">
+        <variable name="FORMAT" expression="" defaultValue="&quot;{:?}&quot;" alwaysStopAt="true"/>
```
matklad: Why stop a format string? We have `p`, `pd` and `ppd` to avoid this. 

Sidenote: it would be cool to implement "echoing" behaviour of Kotlin's `soutv` template.

src/main/resources/org/rust/ide/liveTemplates/other.xml
jajakobyly: Semantically this is correct from what I know, though nobody does this probably

src/main/resources/org/rust/ide/liveTemplates/output.xml
jajakobyly: User might want to attach some additional content (which for example I often do). My reasoning is that typing `tab` one more time is much cheaper than having to manually point cursor there

src/main/kotlin/org/rust/lang/utils/RustNames.kt
```diff
@@ -0,0 +1,35 @@
...
+fun String.isRustIdentifier(withPrimitives: Boolean = true): Boolean =
+    !isRustReserved(withPrimitives) && isRustIdentifierString()
+
+fun String.isRustReserved(withPrimitives: Boolean = true): Boolean =
+    this in RustKeywords || (withPrimitives && this in RustPrimitiveTypes)
+
+fun String.isRustIdentifierString(): Boolean =
+    !isEmpty()
+        && Character.isJavaIdentifierStart(this[0])
+        && drop(1).all { Character.isJavaIdentifierPart(it) }
```
jajakobyly: Nice, will use it

src/main/kotlin/org/rust/ide/template/macros/utils.kt
matklad: We'll move it once we have more use cases for it

src/main/resources/org/rust/ide/liveTemplates/other.xml
matklad: I've tried it, and `rustc` requires to place a `#[main]` attribute on it and then yells that it is unstable. Let's just remove it if  nobody uses it?

jajakobyly:
Hmm, actually Kotlin has `closure` snippet.

src/main/resources/org/rust/ide/liveTemplates/output.xml
```diff
@@ -31,6 +31,24 @@
         </context>
     </template>
 
+    <template name="pdv" description="debug println! value" toReformat="true" toShortenFQNames="true"
+              value="println!(&quot;$EXPRESSION_COPY$ = {:?}&quot;, $EXPRESSION$);">
```
matklad: Wow, it was that easy? 
 
What do you think about merging `pd` and `pdv` templates? We seem to have too may printing related templates now. 

I personally would prefer to have `p` as now and `pd` and `pdd` corresponding to `pdv` and `ppdv`. The reasoning is as follows:

* "Echoing" `pd` and `ppd` perfectly cover printf debuggin scenario.
* `p` covers the lowest common denominator of any other custom printing. 

Providing more specialized templates then `p` seems not very efficient, because there are a lot of ways how you can print something. Inspection to verify the format string ant intentions to switch between `{}`, `{:?}` and `{:#?}` seem a bit more general. 

Thoughts?

jajakobyly: I have no strong feelings on this

matklad:
Thanks @jajakobyly !

src/main/resources/org/rust/ide/liveTemplates/output.xml
matklad: So I've merged this templates, and I think it's kinda best of worlds: cursor is first placed on EXPRESSION, which is copied to EXPRESSION_COPY, but then it is placed over the copy to allow editing.  

https://github.com/intellij-rust/intellij-rust/commit/f85119a2469f0c9507c746887022388871651651

jajakobyly: Hmm, I believe some community feedback will be very useful here. Is this already in nightly? If so maybe I'll ask guys on reddit about this. What do you think?

matklad: It's not yet in nightly and it didn't get into the current alpha as well. 

As for the community feedback, I think we have a general problem that we receive little feedback :( And I suspect this is in part because not everyone understands what features are already available, and what should be possible. 


So yes, it would be nice to collect info about this particular feature (and to learn what part of users know about this snippets!)

PR #566, opened 2016-07-30
`RustImplIndex` revisited

alexeykudinkin:
RTM


src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
```diff
@@ -23,9 +24,12 @@ object RustCompletionEngine {
 
     fun completeFieldOrMethod(field: RustFieldExprElement): Array<out LookupElement> {
         val dispatchType = field.expr.resolvedType.stripAllRefsIfAny()
-        val methods = dispatchType.nonStaticMethods
+
         // Needs type ascription to please Kotlin's type checker, https://youtrack.jetbrains.com/issue/KT-12696.
         val fields: List<RustNamedElement> = (dispatchType as? RustStructType)?.struct?.fields.orEmpty()
+
+        val methods = RustImplIndex.findNonStaticMethodsFor(dispatchType, field.project)
```
matklad: Finding methods is an API of the `RustType`. The fact that we use an index for this is an implementation detail. And we want to use dynamic dynamic dispatch here to avoid `is RustTypeParameterType` cast.

src/main/kotlin/org/rust/lang/core/psi/RustItemElement.kt
```diff
@@ -1,3 +1,22 @@
 package org.rust.lang.core.psi
 
+import org.rust.lang.core.symbols.RustQualifiedPath
+import org.rust.lang.core.symbols.RustQualifiedPathPart
+
 interface RustItemElement : RustVisibilityOwner, RustOuterAttributeOwner
+
+/**
+ * Returns a fully-qualified [RustQualifiedPath] representing a 'path' to this [RustItemElement]
+ * from the crate root, like `foo::bar::baz`. Returns null for the crate's root.
```
matklad: Docstring is incorrect:

>Returns null for the crate's root.

matklad:
Ah, this is huge! I'll try to make only the essential comments yet, because of the WIP status.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustImplIndex.kt
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        return found.asSequence()
+                    .filter { impl ->
+                        impl.type?.let {
+                            it.resolvedType.let { ty ->
+                                ty is RustStructOrEnumTypeBase &&
+
+                                // TODO(XXX): Revisit
+                                ty.item.canonicalCratePath == target.canonicalCratePath
```
matklad: Why not `ty.item == target`?

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustImplIndex.kt
```diff
@@ -0,0 +1,221 @@
...
+                StubIndexKey.createIndexKey("org.rust.lang.core.stubs.index.RustImplIndex.ByName")
+        }
+
+        override fun getKey(): StubIndexKey<String, RustImplItemElement> = KEY
+
+    }
+}
+
+
+class RustAliasIndex : StringStubIndexExtension<RustUseItemElement>() {
```
matklad: I would suggest not to implement AliasIndex at the moment.

* it does not matter in practice in comparison to all other missing features
* current implementation is incomplete anyway, `use foo::{bar as baz, spam as eggs}` and `type Foo = Bar` aliases are not handled.

src/main/kotlin/org/rust/lang/core/symbols/impl/RustQualifiedPath.kt
```diff
@@ -0,0 +1,44 @@
...
+
+
+    override fun equals(other: Any?): Boolean {
+        if (other !is org.rust.lang.core.symbols.impl.RustQualifiedPath)
+            return false
+
+        val lop = unfold().toList()
+        val rop = other.unfold().toList()
+
+        if (lop.size != rop.size)
```
matklad: Why not simply `lop == rop`?

src/main/kotlin/org/rust/lang/core/types/util/RustTypeExtensions.kt
```diff
@@ -11,3 +12,25 @@ fun RustType.stripAllRefsIfAny(): RustType =
...
+    get() = accept(RustDecayTypeVisitor())
+
+
+/**
+ * Checks whether this particular type is a primitive one
+ */
+val RustUnresolvedType.isPrimitive: Boolean
+    get() =
+        when (this) {
+            is RustFloatType, is RustIntegerType, is RustCharacterType, is RustStringType -> true
```
matklad: Boolean is missing

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustEqualityTypeVisitor.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        return visitTypeList(lop.types, type.types)
+    }
+
+    override fun visitFunctionType(type: RustFunctionType): Boolean {
+        val lop = lop
+        if (lop !is RustFunctionType)
+            return false
+
+        return visit(lop.retType, type.retType) && visitTypeList(lop.paramTypes, type.paramTypes)
```
matklad: Bug, lengths of `paramTypes` are not considered.

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustEqualityTypeVisitor.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        return lop.path == type.path
+    }
+
+    override fun visitTupleType(type: RustUnresolvedTupleType): Boolean {
+        val lop = lop
+        if (lop !is RustUnresolvedTupleType)
+            return false
+
+        return visitTypeList(lop.types, type.types)
```
matklad: Bug?

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustEqualityTypeVisitor.kt
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    override fun visitFunctionType(type: RustUnresolvedFunctionType): Boolean {
+        val lop = lop
+        if (lop !is RustUnresolvedFunctionType)
+            return false
+
+        return visit(lop.retType, type.retType) && visitTypeList(lop.paramTypes, type.paramTypes)
```
matklad: Same

matklad:
Would be nice to see some new tests! On the first sight it looks like a ton of complexity with apparently little gain and 3x worse performance ([RustHighlightingPerformanceTest](https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/annotator/RustHighlightingPerformanceTest.kt)).

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -9,24 +9,36 @@ import org.rust.lang.core.psi.*
...
-            HighlightingVisitor().compute(element)
-        }
+        val highlightingInfo =
+                // TODO(XXX): Much better would be to incorporate
+                //            that behaviour into visitor
+                if (element is RustReferenceElement) {
+                    val parent = element.parent
+                    val isPrimitiveType =
+                        element is RustPathElement &&
+                        parent is RustPathTypeElement &&
```
matklad: This won't work for cases like

```
fn main() {
    let _ = f32::max(1.0, 2.0);
}
```

but it's not very important.

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustEqualityTypeVisitor.kt
alexeykudinkin: Good catch

alexeykudinkin:
Little gain? Previous approach was a dead-end with no way to extend it to the general case of non-inherent `impl` implementing the trait for the generic type.

alexeykudinkin:
How did you get 3x performance degradation? My measurements show <10% slowdown.

matklad:
>How did you get 3x performance degradation?

I've run this test: https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/annotator/RustHighlightingPerformanceTest.kt#L14

It highlights a not very big Rust file with the stdlib properly setup. So most of the time should be spend on the resolve inside stdlib. 

On master it reports about 2000 ms, on impls-index it is about 6000. Variance is pretty high in both cases though, but not high enough to hide the performance gap.

matklad:
>Previous approach was a dead-end with no way to extend it to the general case of non-inherent impl implementing the trait for the generic type.

I want to make sure that this approach is extensible. Will it be able to handle `impl<T> Foo for T where T: Bar`? What are other cases that it handles and previous approach does not?

alexeykudinkin:
It will. Generally speaking, it would be capable to handle any structure-matching required.

alexeykudinkin:
Will review the gap.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustImplIndex.kt
alexeykudinkin: Cause, we need to revisit all comparisons of that kind. Essentially two PSI files one of which stub-backed, the other being AST-backed, are far from being equal.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustImplIndex.kt
alexeykudinkin: > it does not matter in practice in comparison to all other missing features

Wrong path. Missing feature and _misbehaving_ feature is 2 different kinds of issues. 

These issues are utterly minor, and i don't see tremendous efforts required to fix them.

src/main/kotlin/org/rust/lang/core/symbols/impl/RustQualifiedPath.kt
alexeykudinkin: It's a good time to start trusting `List.equals`.

src/main/kotlin/org/rust/lang/core/types/util/RustTypeExtensions.kt
alexeykudinkin: image

src/main/kotlin/org/rust/lang/core/resolve/indexes/RustImplIndex.kt
matklad: > two PSI files one of which stub-backed, the other being AST-backed, are far from being equal.

?

A PsiFile is a singe object, which can be backed by either AST or Stubs. There is exactly one PsiFile for each file, so you can't have both stub and AST backed files. So, we should trust equals for PsiElements

The only gotcha is completion is invoked in a temporary copy of the original file (because `intellijIdeaRulezz` is inserted), but there is `getOriginalElement()` for that.

alexeykudinkin: That's not true obviously. You can verify that removing `.canonicalCratePath` part from that comparison.

matklad: I've tried removing `.canonicalPath` and I still believe that my analysis is correct. Only completion tests fail and this is precisely because completion is invoked on the **different** PsiFile, with literally different contents. This has nothing to to with Stubs. 

Tests regressed after the introduction of `val item`  to `RustStructOrEnumTypeBased`, which bypasses [proper initialization](https://github.com/intellij-rust/intellij-rust/blob/0be9632e4664382d654e5b467bffe3e5e2899c46/src/main/kotlin/org/rust/lang/core/types/RustStructType.kt#L8) which makes sure that types are always based on the original elements. We should avoid storing the same element twice withing a type, which currently happens with structs and enums. 

It should be fixed like this:

```Kotlin
abstract class RustStructOrEnumTypeBase : RustTypeBase() {
    abstract val item: RustStructOrEnumItemElement
}

class RustStructType(struct: RustStructItemElement) : RustStructOrEnumTypeBase() {

    override val item = CompletionUtil.getOriginalOrSelf(struct)
}
```

alexeykudinkin: Wow! Good point. Didn't noticed that the only tests failing were completion-related ones.

alexeykudinkin: Didn't give it much of attention due to the fact that, though after parse being triggered stub-based PSI nodes may be still there due to outstanding references being held still.

matklad: > stub-based PSI nodes may be still there due to outstanding references being held still.

?

Parsing a file will switch all outstanding PSI elements to AST (see `PsiFileImpl#switchFromStubToAst`).

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustEqualityTypeVisitor.kt
```diff
@@ -87,7 +87,7 @@ open class RustEqualityUnresolvedTypeVisitor(override var lop: RustUnresolvedTyp
     }
 
     fun visitTypeList(lop: Iterable<RustUnresolvedType>, rop: Iterable<RustUnresolvedType>): Boolean =
-        lop.zip(rop).fold(true, { r, p -> r && visit(p.first, p.second) })
+        lop.count() == rop.count() && lop.zip(rop).fold(true, { r, p -> r && visit(p.first, p.second) })
```
matklad: There's `all` function for this kind of `fold`.

src/main/kotlin/org/rust/lang/core/types/RustTraitType.kt
```diff
@@ -9,9 +9,9 @@ import org.rust.lang.core.types.visitors.RustTypeVisitor
  * Though you use the same path to denote both traits and trait objects,
  * only the latter are types.
  */
-class RustTraitObjectType(val trait: RustTraitItemElement) : RustTypeBase() {
+class RustTraitType(val trait: RustTraitItemElement) : RustTypeBase() {
```
matklad: No, trait object is the correct term here: https://doc.rust-lang.org/reference.html#trait-objects

src/main/kotlin/org/rust/lang/core/psi/RustItemElement.kt
matklad: Still there

src/main/kotlin/org/rust/lang/core/psi/RustMod.kt
```diff
@@ -24,22 +25,19 @@ interface RustMod : RustNamedElement, RustItemsOwner {
...
- *
- * This name is not guaranteed to be unique: modules from different crates
- * can have the same path within the crate.
- */
-val RustMod.canonicalCratePath: String? get() {
-    if (isCrateRoot) return ""
-    val parentPath = `super`?.canonicalCratePath ?: return null
-    return if (parentPath == "") modName else "$parentPath::$modName"
-}
+val RustMod.canonicalCratePath: RustQualifiedPath?
```
matklad: Potential footgun: invoking `canonicalCratePath` on `RustMod` will give different results depending on the static type of the receiver expression.

src/main/kotlin/org/rust/lang/core/stubs/elements/RustUseItem.kt
```diff
@@ -0,0 +1,43 @@
...
+
+class RustUseItemElementStub : RustElementStub<RustUseItemElement> {
+
+    val alias: String?
+
+    constructor(parent: StubElement<*>?, elementType: IStubElementType<*, *>, aliased: String?)
+        : super(parent, elementType) {
+
+        alias = aliased
+    }
```
matklad: The whole body can be turned into just one primary constructor

src/test/resources/org/rust/cargo/runconfig/producers/fixtures/meaningfulConfigurationName.xml
```diff
@@ -1,6 +1,6 @@
 <configurations>
   <configuration name="Test bar::tests" class="CargoCommandConfiguration" show_console_on_std_err="false" show_console_on_std_out="false">
-    <option name="additionalArguments" value="--lib bar::tests" />
+    <option name="additionalArguments" value="--lib ::bar::tests" />
```
matklad: Nope, the test harness requires that there is no leading double colon here.

matklad:
I still think that we should add some tests that demonstrate that this approach is more powerful than the current one. 

On the first glance it seems that we are still just comparing the names of the types, albeit in a very elaborate double dispatchy way...

src/main/kotlin/org/rust/lang/core/psi/RustMod.kt
alexeykudinkin: Elaborate?

alexeykudinkin:
Reasonable, will push some tests stressing this caps out.

alexeykudinkin:
BTW, to see the real impact of this change, better would be to open any real project.

src/main/kotlin/org/rust/lang/core/psi/RustMod.kt
matklad:
```Kotlin
fun quux(mod: RustModItemElement) {
    fun foo(mod: RustMod)         = mod.canonicalCratePath
    fun bar(mod: RustItemElement) = mod.canonicalCratePath

    foo(mod) // This will call a function from `RustMod.kt`
    bar(mod) // This will call a function from `RustItemElement.kt`
}
```

src/test/kotlin/org/rust/lang/core/resolve/RustTypeAwareResolveTestCase.kt
```diff
@@ -25,9 +25,12 @@ class RustTypeAwareResolveTestCase : RustResolveTestCaseBase() {
 
     fun testHiddenInherentImpl() = checkIsBound()
     fun testWrongInherentImpl() = checkIsUnbound()
-    fun testNonInherentImpl() = checkIsBound()
     fun testImplGenericsStripped() = checkIsBound()
 
+    fun testNonInherentImpl1() = checkIsBound()
+    fun testNonInherentImpl2() = checkIsBound()
+    fun testNonInherentImpl3() = checkIsBound()
```
matklad: `Impl2` and `Impl3` cases should resolve to the same method:

http://play.integer32.com/?gist=8eac4db6da8c8198687bda55139f8a36

alexeykudinkin: Yeah, the example turned out to be a little bit more sophisticated, than i've initially deemed it.

alexeykudinkin: Though it's hugely unrelated to the P/R itself and should be fixed separately.

src/main/kotlin/org/rust/lang/core/psi/RustMod.kt
alexeykudinkin: Fixed.

alexeykudinkin:
@matklad other comments/suggestions? Let's land it today.

matklad:
I'd love to see more tests because it still seems to me that this PR does basically the same "name comparison" and does not significantly improve state of the art besides aliases support.

>BTW, to see the real impact of this change, better would be to open any real project.

I've tried it with cargo and failed to see any obvious improvements. 

Also, let's revert `RustTraitObjectType -> RustTraitObject` rename (https://github.com/intellij-rust/intellij-rust/pull/566#discussion_r73792487) and fix or remove incorrect docstrings (https://github.com/intellij-rust/intellij-rust/pull/566#discussion_r72890617, maybe more)
 
If you are confident that this is the way to go with indexes (I am, as usual, not :) ), let's merge this.

alexeykudinkin:
> I'd love to see more tests because it still seems to me that this PR does basically the same "name comparison" and does not significantly improve state of the art besides aliases support.

I don't quite understand your position. What type of significance you're looking for? It does the same thing as the previous implementation, that's for sure, but it does it in the different manner, allowing to reduce amount of resolves being triggered during the match and speeding up very that resolve whatsoever.

alexeykudinkin:
> > BTW, to see the real impact of this change, better would be to open any real project.

> I've tried it with cargo and failed to see any obvious improvements.

My previous comment is irrelevant, seems like i've just didn't sync with `master` for a while.

matklad:
My point is that this PR adds a significant amount of complexity (~+2kloc), but apparently uses fundamentally the same algorithm as the current approach. 

So, why exactly is this more complex approach better? 

> allowing to reduce amount of resolves being triggered during the match and speeding up very that resolve whatsoever.

Hm, the performance of `RustHighlightingPerformanceTest` is the same for me in this branch and in the master. Where do these performance gains should come from? Loos like we still do two phase "look by name and resolve to verify" thing. Here is resolve being triggered form the impl index: https://github.com/intellij-rust/intellij-rust/blob/3dec3949b4ed9d37403f5c42953a3d01a8bc028b/src/main/kotlin/org/rust/lang/core/resolve/indexes/RustImplIndex.kt#L92

alexeykudinkin:
Most of this 2k has nothing to do with the index.

> So, why exactly is this more complex approach better?

Becuase it's just doesn't need to rely on the stuff like `implNominalTypeName`, at first. 
Second, it do relies on resolve also indeed, b/c it's still unresolved-index. It though able to cut the search space during the search since, all the `impl`s are indexed by their respective unresolved self-type, and therefore being matched structurally prior to kicking off the resolve.

matklad:
>Most of this 2k has nothing to do with the index.

But what improvements do these new lines bring? If they don't allow for more precise or faster analysis then perhaps we don't need them? It's easier to test and maintain less code.

There are certainly very useful aspects of this PR, like alias index, new stubs and `rust.bnf` improvements. But this whole equality/hash visitors and Qualified Paths stuff seems unnecessary to me.

>Becuase it's just doesn't need to rely on the stuff like implNominalTypeName, at first. 

And why is this beneficial? This equality/hash visitors seems massively more complicated to me than a simple string key function. It actually reminds me of the different APIs for the sort function, key based and comparator based. 

>It though able to cut the search space during the search since, 

The cut is negligible in practice. This is demonstrated by the performance test and is easy to explain empirically: almost all impls are of the form `impl Name for Name` or `impl Name`.

To sum up, here is what I think this change does and does not deliver:

* A number of useful improvements, not related to the impl index
* No significant improvement in resolve quality
* No significant performance improvement or degradation
* A new infra for the impls index, based on complicated `equals/hashCode` for types and `QualifiedPaths`.

The first bullet is a sure win, the next two do not matter, and the third one seems a lose to me (more code without new features).

Perhaps we can split this PR into two?

matklad:
> we can split this PR into two?

And if you don't want to, I'm Ok with merging this as is :)

alexeykudinkin:
> But what improvements do these new lines bring? If they don't allow for more precise or faster analysis then perhaps we don't need them? It's easier to test and maintain less code.

How current implementation is gonna tackle cases like this one: `impl<T> Foo for T {}`?

alexeykudinkin:
> > Becuase it's just doesn't need to rely on the stuff like implNominalTypeName, at first.

> And why is this beneficial? This equality/hash visitors seems massively more complicated to me than a simple string key function. It actually reminds me of the different APIs for the sort function, key based and comparator based.

That simple string key function is a crutch: it served its purpose, but its in-extensible in any imaginable way. What particular entity of the types this `nominalTypeName` thing belongs to? What does this mean for reference type to have `nominalTypeName`?

alexeykudinkin:
Equality/HC visitors is clear as day abstraction that has no more complexity than overriding `hashCode`/`equals`. Replacing the latter two with such kind of visitors is actually beneficial in a  way that this P/R hugely relies on -- it allows to implement custom comparison logic or override HC for particular types at no cost.

matklad:
And why we need QualifiedPaths if we ever look at only the last segment?

alexeykudinkin:
B/c we can't store PSI in the index and we need a way to store unresolved path-types in it.

matklad:
>How current implementation is gonna tackle cases like this one: impl<T> Foo for T {}?

And how this implementation would do it? :) 

I think it's possible to use two approaches here

* Completely dodge the issue by Indexing non-inherent impls by trait instead of the type
* This seems to be the only case without obvious "nominal type", so we can store it under a special `"<Generic>"` key. 

> What does this mean for reference type to have nominalTypeName?

As the `testNonInhetentImpl` 2 and 3 shows, reference types need some special handling. I would suggest that we just strip all the references when we index an impl (in either implementation of the impl index)

>What particular entity of the types this nominalTypeName thing belongs to?

Looks like any type in Rust can be thought of as `TypeConstrucor<TypeParam1, ..., TypeParamN>`. This nominal type thing is actually the name of `TypeConstructor`. It serves the same purpose as those custom visitors: custom equals for `t1` and `t2` returns `true` iff `t1.nominalType == t2.nominalType`. In essence, `nominalType` is just a fingerprint of the type which is oblivious of irrelevant details such as references or type parameters.

matklad:
>B/c we can't store PSI in the index and we need a way to store unresolved path-types in it.

But we actually don't need to store full unresolved path types because we only ever look at the last segment.

alexeykudinkin:
We store it not only as the key in the index but instead of stubs also: that allows to mitigate committing PSI altogether.

alexeykudinkin:
> And how this implementation would do it? :)

The same way it does now with concrete types: but instead of matching the tail name parts, it would collect the substitution-map during type traversal, and if all the entries of substitution map match required conditions, then we're done.

matklad:
Thinking of it more, we need to store some kind of the fingerprint, which should be more coarse grained thsn resolved types. It also should be more coarse grained then unresolved types (because in this PR we compare unresolved types in a "fuzzy" way anyway)

So perhaps instead of QualifiedPaths, we can have a simple *non recursive* 

```
data class FingerPrint(
    val name: String,
    val referenceDepth: Int,
    val anyAdditionalInfoWeWouldNeedToStore
)
```

And we then will be able to implement `RustType.fingerPrint` and `RustUnresolvedType.fingerPrint`, which would allow us to get rid of Eq, Hash and Decay visitors.

alexeykudinkin:
> Completely dodge the issue by Indexing non-inherent impls by trait instead of the type
This seems to be the only case without obvious "nominal type", so we can store it under a special "<Generic>" key.

First one, assumes that we'd be checking `impl`s for every trait in the crate. Second one is as worst as it could be.

alexeykudinkin:
> Thinking of it more, we need to store some kind of the fingerprint, which should be more coarse grained thsn resolved types. It also should be more coarse grained then unresolved types (because in this PR we compare unresolved types in a "fuzzy" way anyway) 
So perhaps instead of QualifiedPaths, we can have a simple non recursive

Thats fuzziness is there for a reason. Let not re-invent the wheel and call it _the wheel_: all the fuzziness required may be implemented as a custom comparison policy, and types are used not just b/c they can be used there but b/c them can be easily converted back and forth and store all the information required.

matklad:
>The same way it does now with concrete types: but instead of matching the tail name parts, it would collect the substitution-map during type traversal, and if all the entries of substitution map match required conditions, then we're done.

You won't be able to query the index based on the substitution map, so you'll need to invent some kind of 

```
object Generic: QualifiedPathParh {
    val name = "<Generic>"
}
```

to actually lookup this thing in the index. 

>all the fuzziness required may be implemented as a custom comparison policy, and types are used not just b/c they can be used there but b/c them can be easily converted back and forth and store all the information required

All the fuzziness can be implemented by using `nominalTypeName` thing as well, and you won't need to write custom code for serialization, hashing, decaying and comparison. 


Storing unresolved types in the stubs instead of stubbing `TypeElement`s themselves seems like an interesting idea, but maybe it's easier to just stub `TypeElements`?

matklad:
>First one, assumes that we'd be checking impls for every trait in the crate. 

Nope, you'll need to check only crates in scope. This actually an interesting tradeoff: if you index by trait, you'll need to scrutinize impls for unrelated types, but if you index by type, you'll get imls for unrelated (not visible) traits.

> Second one is as worst as it could be.

The same as in 
```
object RustAnonymousQualifiedPathPart : RustQualifiedPathPart {
    override val name: String
        get() = "<anonymous>"
}
```

matklad:
>Also, let's revert RustTraitObjectType -> RustTraitObject rename (#566 (comment)) and fix or remove incorrect docstrings (#566 (comment), maybe more)

Well, let's fix ^^^^^ and merge this already :)

alexeykudinkin:
> You won't be able to query the index based on the substitution map, so you'll need to invent some kind of
```
object Generic: QualifiedPathParh {
    val name = "<Generic>"
}
```
to actually lookup this thing in the index.

No, you would not. You just need to answer the question whether particular `impl`s self-type may satisfy (ie _looks_ like) the type being query's target.

alexeykudinkin:
> All the fuzziness can be implemented by using nominalTypeName thing as well, and you won't need to write custom code for serialization, hashing, decaying and comparison.

https://github.com/intellij-rust/intellij-rust/pull/566#issuecomment-238111551

alexeykudinkin:
> Storing unresolved types in the stubs instead of stubbing TypeElements themselves seems like an interesting idea, but maybe it's easier to just stub TypeElements?

Easier to implement the whole stub-tree instead of using what's done already? What's the point of this stubs, where they can be reused futher?

alexeykudinkin:
> Nope, you'll need to check only crates in scope. This actually an interesting tradeoff: if you index by trait, you'll need to scrutinize impls for unrelated types, but if you index by type, you'll get imls for unrelated (not visible) traits.

The former one seems to be hugely majoring the latter in any reasonable code-base.

alexeykudinkin:
> The same as in
```
object RustAnonymousQualifiedPathPart : RustQualifiedPathPart {
    override val name: String
        get() = "<anonymous>"
}
```

That's used primarily for graceful degradation purposes, and isn't sensible in any other way.

alexeykudinkin:
> > Also, let's revert RustTraitObjectType -> RustTraitObject rename (#566 (comment)) and fix or remove incorrect docstrings (#566 (comment), maybe more)

> Well, let's fix ^^^^^ and merge this already :)

The doc you're sending too is poorly worded, and contains a few self-contradictions:

> Calling a method _on a trait object_ results in virtual dispatch at runtime

What's actually meant by _trait objects_ in there are instances of types `&SomeTrait` or `Box<SomeTrait>`, which can be clearly followed [here](https://doc.rust-lang.org/book/trait-objects.html#dynamic-dispatch).

Therefore `RustTraitType` seems more appropriate than `RustTraitObjectType`.

matklad:
No, trait object is the official term:

https://doc.rust-lang.org/book/trait-objects.html

Traits are not types. You can't have a value whose type is a trait. You can have a value whose type is a trait object.

alexeykudinkin:
> Calling a method on a trait object results in virtual dispatch at runtime

Once again: this wording then suppose you're calling method on a _trait_ and not on object whose _projection_ is that trait (which is clearly what they're meant to be).

matklad:
>you're calling method on a trait 

You can't call a method on a trait because traits are not values, unlike structs, enums or trait objects.

alexeykudinkin:
That's what i'm talking about.

PR #571, opened 2016-08-01
(FMT): add spacing rules for lifetime lists in polybounds

### LINKED ISSUE #569: `for<'lifetime>` formatting issue
Currently (rust plugin v0.0.1.866-nightly) thing like `where F: for<'b> Fn(...)` are reformatted to `where F: for < 'b > Fn(...)` which is a bit ugly and inconsistent with type parameters formatting in other cases (in `impl`, `struct`/`trait`, `fn` they are condensed).

Would you change default formatting for this case and add an option in `Code style -> Rust` if it's really needed for someone.

jajakobyly:
fixes #569


matklad:
Thanks!

PR #576, opened 2016-08-05
(INTENTION) wrap and unwrap lambda expression body in braces

### LINKED ISSUE #538: Intention to wrap lambda expression body in braces
That is, <kbd>Alt+Enter</kbd> should allow to toggle between

```
xs.iter().map(|x| x * x)
```

and

```
xs.iter().map(|x| { x * x })
```

zjhmale:
Fixes #538 

snapshot


src/main/kotlin/org/rust/ide/intentions/UnWrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustBlockExprElement
+import org.rust.lang.core.psi.RustElementFactory
+import org.rust.lang.core.psi.RustLambdaExprElement
+import org.rust.lang.core.psi.util.parentOfType
+
+class UnwrapLambdaExprIntention : PsiElementBaseIntentionAction() {
+    override fun getText() = "UnWrap lambda expression body in braces"
```
mkaput: Unwrapis correct word

zjhmale: woops 😂

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustBlockExprElement
+import org.rust.lang.core.psi.RustElementFactory
+import org.rust.lang.core.psi.RustLambdaExprElement
+import org.rust.lang.core.psi.util.parentOfType
+
+class UnwrapLambdaExprIntention : PsiElementBaseIntentionAction() {
+    override fun getText() = "Unwrap lambda expression body in braces"
```
mkaput: What do you think about changing this to _Remove braces from lambda expression_? Just like in [Kotlin plugin](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveBracesIntention.kt)

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustBlockExprElement
+import org.rust.lang.core.psi.RustElementFactory
+import org.rust.lang.core.psi.RustLambdaExprElement
+import org.rust.lang.core.psi.util.parentOfType
+
+class WrapLambdaExprIntention : PsiElementBaseIntentionAction() {
+    override fun getText() = "Wrap lambda expression body in braces"
```
mkaput: What do you think about changing this to _Add braces to lambda expression_? Just like in [Kotlin plugin](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt)

src/test/resources/org/rust/ide/intentions/fixtures/unwrap_lambda_expr/unwrap_braces.rs
```diff
@@ -0,0 +1,3 @@
+fn main () {
+    |x| { x *<caret> x }
```
mkaput: I think enabling this inspection here may be too much invasive. What do you think about limiting it only to  top-level expr (in other words, enabling only on ends of expression)

zjhmale: @mkaput sorry, i don't get it 😂 , which part do you think is too invasive, i mean top-level lambda expression is illegal if it not wraped inside a function.

mkaput: I mean, I don't like that 
```rust
|x| { x *<caret> x }
```
works.

Imagine that our block is 100 lines long, possibly with multiple nested blocks. User will encounter similar situation to the one we had with expand module.

Instead I think this inspection should be only available here:
```rust
|x| <caret>{ x * x }
```
and eventually here:
```rust
|x| { x * x }<caret>
```

See linked implementation in Kotlin how they did it there for reference. The same applies to inverse intention. 

But before implementing this I would love to hear feedback from others, espacially @matklad but he's "afk" now :smile:

zjhmale: ok, when nested blocks it will indeed encounter a problem, i will refer the Kotlin implementation.

zjhmale: @mkaput `{ ... }` can only be `RustBlockExprElement` so i can not figure out a way to access lbrace and rbrace element just like [here](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveBracesIntention.kt#L31-L33) does.

mkaput: Hmm what about (pseudocode):
```kotlin
// find first parent which is either block or lambda
// if we get lambda, than return false as we are inline "unblocked" (hehe) lambda
val block = PsiTreeUtil.findFirstParent(element, blockOrLambda) as? RustBlockExprElement ?: return false

// check whether our block's parent is lambda to avoid false positives
// like nested blocks in lambda (we are not interested in this case as we can
// be sure that cursor isn't at brace; well, I don't know what about `|x| {<caret>{ x }}`,
// but it's possibly rare edge case), or just some random block in code
if (block.parent !is RustLambdaExprElement) return false

// ... in companion object
private val blockOrLambda = Condition<PsiElement> { element ->
    element is RustBlockExprElement || element is RustLambdaExprElement
}
```

I used this trick with `findFirstParent` in [RustContextType](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/template/RustContextType.kt) and nobody complained yet.

zjhmale: this is cool i think, to handle the edge case is a little bit tricky :joy:

src/test/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntentionTest.kt
```diff
@@ -0,0 +1,17 @@
...
+
+    fun testEnableUnwrapBraces() = checkByFile {
+        openFileInEditor("enable_unwrap_braces.rs")
+        myFixture.launchAction(UnwrapLambdaExprIntention())
+    }
+
+    fun testUnableUnwrapBraces() = checkByFile {
+        openFileInEditor("unable_unwrap_braces.rs")
+        myFixture.launchAction(UnwrapLambdaExprIntention())
+    }
```
mkaput: enabled/disabled

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,39 @@
...
+        val paramsStr = lambdaExpr.parameters.text
+        val bodyStr = block.block?.expr?.text ?: return
+        val unbraced = RustElementFactory.createUnbracedLambdaExpr(project, paramsStr, bodyStr) ?: return
+        lambdaExpr.replace(unbraced)
+    }
+
+    override fun isAvailable(project: Project, editor: Editor, element: PsiElement): Boolean {
+        if (!element.isWritable) return false
+        val block = PsiTreeUtil.findFirstParent(element, blockOrLambda) as? RustBlockExprElement ?: return false
+        if (block.parent !is RustLambdaExprElement) return false
```
mkaput: Please leave a comment about what's going on here. You can freely copy paste ones I provided.

zjhmale: ok 😄

src/test/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntentionTest.kt
zjhmale: i prefer to use unable, it means we can not trigger this intention. anyway, enabled/disabled also make sense.

mkaput: I think available / unavailable is even better

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
zjhmale: @mkaput this is a little bit confusing that if we want block, why should we find block or lambda first?
```
// find first parent which is either block or lambda
// if we get lambda, than return false as we are inside "unblocked" (hehe) lambda
```

src/test/resources/org/rust/ide/intentions/fixtures/unwrap_lambda_expr/unwrap_braces.rs
matklad: Oups, looks like I've accidentally edited @mkaput comment instead of creating a reply :sweat_smile: https://github.com/intellij-rust/intellij-rust/pull/576#discussion_r73699058

Sorry for that.

zjhmale:
poke @matklad

matklad:
>poke @matklad

Yeah, I hope I'll get the chance to review these PR's soon: have been participating in ICFP contest during the weekand, and know investigating some hugely mysterious build problems :sob:

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+        // if we get lambda, than return false as we are inside "unblocked" lambda
+        val block = PsiTreeUtil.findFirstParent(element, blockOrLambda) as? RustBlockExprElement ?: return false
+
+        // check whether our block's parent is lambda to avoid false positives
+        // like nested blocks in lambda (we are not interested in this case as we can
+        // be sure that cursor isn't at brace; well, I don't know what about `|x| {<caret>{ x }}`,
+        // but it's possibly rare edge case), or just some random block in code
+        if (block.parent !is RustLambdaExprElement) return false
+        val body = block.block?.expr ?: return false
+        return block.text.removePrefix("{").removeSuffix("}").trim() == body.text
```
matklad: I think a PSI level check is preferable to the text level check: 

```Kotlin
        val blockExpr = PsiTreeUtil.findFirstParent(element, blockOrLambda) as? RustBlockExprElement ?: return false
        val block = blockExpr.block ?: return false
        return blockExpr.parent is RustLambdaExprElement 
            && block.expr != null 
            && block.lbrace.getNextNonCommentSibling() == block.expr
```

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
```
matklad: As usual, `isAvailable` and `invoke` duplicate quite a bit of logic, which can be extracted into a separate method.

Something like this:

```Kotlin
    private data class Context(
        val lambda: RustLambdaExprElement,
        val expression: RustExprElement
    )

    private fun findContext(element: PsiElement): Context? {
        ...
    }

    override fun invoke(project: Project, editor: Editor, element: PsiElement) {
        val ctx = findContext(element) ?: return
        val paramsStr = ctx.lambda.parameters.text
        val bodyStr = ctx.expression.text
        val unbraced = RustElementFactory.createUnbracedLambdaExpr(project, paramsStr, bodyStr) ?: return
        ctx.lambda.replace(unbraced)
    }

    override fun isAvailable(project: Project, editor: Editor, element: PsiElement): Boolean {
        return findContext(element) != null
    }
```

Perhaps we should even add a base intention class for this pattern?

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+    override fun getFamilyName() = getText()
+    override fun startInWriteAction() = true
+
+    override fun invoke(project: Project, editor: Editor, element: PsiElement) {
+        val lambdaExpr = element.parentOfType<RustLambdaExprElement>() ?: return
+        val block = PsiTreeUtil.findFirstParent(element, blockOrLambda) as? RustBlockExprElement ?: return
+        val paramsStr = lambdaExpr.parameters.text
+        val bodyStr = block.block?.expr?.text ?: return
+        val unbraced = RustElementFactory.createUnbracedLambdaExpr(project, paramsStr, bodyStr) ?: return
+        lambdaExpr.replace(unbraced)
```
matklad: I think that instead of replacing the whole lambda we may replace only it's body:

```
lambdaExpr.replace(block.expr)
```

matklad:
Looks good! Don't forget to add a description to https://github.com/intellij-rust/intellij-rust/tree/master/src/main/resources/intentionDescriptions

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
zjhmale: cool! that's what i was looking for 😂  can not find a PSI level api like `lbrace` and `rbrace`, thx!

zjhmale: cool! that's what i was looking for 😂  can not find a PSI level api like `lbrace` and `rbrace`, thx!

zjhmale:
that should be a lot of fun! 😄 
>Yeah, I hope I'll get the chance to review these PR's soon: have been participating in ICFP contest during the weekand, and now investigating some hugely mysterious build problems 😭

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
zjhmale: ok, i will fix that.

yeah, maybe open an new issue and discuss further?
>Perhaps we should even add a base intention class for this pattern?

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
zjhmale: fair enough 😄

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
matklad: >yeah, maybe open an new issue and discuss further?

Let's not worry about it just yet, implementing new feature and refactoring old in one go is a sure way to mess something up :)

zjhmale:
@matklad ready to be reviewed 😃

src/main/kotlin/org/rust/lang/core/psi/RustElementFactory.kt
```diff
@@ -26,6 +26,12 @@ object RustElementFactory {
         return createFromText(project, "fn main() { S { $fields }; }")
     }
 
+    fun createBracedLambdaExpr(project: Project, params: String, body: String): RustLambdaExprElement? =
+        createFromText(project, "fn main() { $params { $body } }")
+
+    fun createUnbracedLambdaExpr(project: Project, params: String, body: String): RustLambdaExprElement? =
```
matklad: Unused function, it should be removed.

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+            return null
+        }
+    }
+
+    override fun invoke(project: Project, editor: Editor, element: PsiElement) {
+        val ctx = findContext(element) ?: return
+        val lambdaBody = ctx.lambdaExpr.expr ?: return
+        val paramsStr = ctx.lambdaExpr.parameters.text
+        val bodyStr = ctx.lambdaExpr.expr?.text ?: return
+        val braced = RustElementFactory.createBracedLambdaExpr(project, paramsStr, bodyStr) ?: return
```
matklad: I think it's better to create a block expression directly.

zjhmale: sounds reasonable 👍 , so `createBracedLambdaExpr` and `createUnBracedLambdaExpr` can be totally removed as your last comment.

zjhmale:
@matklad done! now create block expression directly.

matklad:
Thanks!

zjhmale:
my pleasure 😄

PR #579, opened 2016-08-06
(EDT): improve enter in line comment handler

### LINKED ISSUE #578: Newline is handled incorrectly in doc comments
In doc comment like this:
```
//! This crate does something useful
(cursor here)//! Description goes here
//! And some notes
```

When I try to insert a new blank like before description, I get this:
```
//! This crate does something useful

//! (cursor here)//! Description goes here
//! And some notes
```

However, I expect something like this:
```
//! This crate does something useful
//! 
(cursor here)//! Description goes here
//! And some notes
```

The same issue is true for `///` doc comment.

mkaput:
also sneaked minor improvements
fix #578


matklad:
Thanks!

PR #585, opened 2016-08-08
(EDT): Enter in string literal

### LINKED ISSUE #195: Support splitting and joining string literals
According to _Rust Reference: 3.5.1.2.2_ [string literals][string literals] can span multiple lines.

It would be nice to able to split the string literals with `Enter` and join with `Join Lines` action.

### Suggested fixtures

Before splitting:

```rust
fn main() {
    let lit = "Hello, <caret>World";
    println!("{}", lit);
}
```

After splitting:

```rust
fn main() {
    let lit = "Hello, \
               <caret>World";
    println!("{}", lit);
}
```

Before joining:

```rust
fn main() {
    let lit = "He<caret>llo, \
               World";
    println!("{}", lit);
}
```

After joining:

```rust
fn main() {
    let lit = "Hello, <caret>World";
    println!("{}", lit);
}
```

 [string literals]: http://doc.rust-lang.org/reference.html#string-literals

mkaput:
This already has been waiting to be finished for quite a while, so I eventually decided not to implement joining here. Also I think fixing indentation could be also done later.

cc #195


src/test/resources/org/rust/ide/typing/string/fixtures/after_opening_after.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let lit = "\
+    <caret>Hello, World";
```
matklad: Perhaps we want an indent level here

src/main/kotlin/org/rust/lang/core/psi/impl/RustRawStringLiteralImpl.kt
```diff
@@ -4,6 +4,15 @@ import com.intellij.psi.tree.IElementType
 import org.rust.lang.core.psi.RustLiteralTokenType
 
 class RustRawStringLiteralImpl(type: IElementType, text: CharSequence) : RustTextLiteralImplBase(type, text) {
+    private var _hashes: Int = 0
+
+    val hashes: Int
+        get() {
+            @Suppress("UNUSED_VARIABLE")
+            val o = offsets // ensure #computeOffsets() has been called
```
matklad: Very smelly. Let's just eagerly compute offsets in the constructor? I doubt that it will have any performance implications, but you can check using something like `RustHighlightingPerformanceTestCase`

src/main/kotlin/org/rust/ide/typing/RustQuoteHandler.kt
```diff
@@ -1,12 +1,63 @@
...
+        val offsets = literal.offsets ?: return true
+        val relativeOffset = offset - iterator.start
+        return offsets.value?.containsOffset(relativeOffset) ?: true
+    }
+
+    override fun getClosingQuote(iterator: HighlighterIterator, offset: Int): CharSequence? {
+        val literal = getLiteral(iterator) ?: return null
+
+        if (literal is RustRawStringLiteralImpl) {
+            return '"' + StringUtil.repeatSymbol('#', literal.hashes)
```
matklad: `"#".repace(literal.hashes)`

src/main/kotlin/org/rust/ide/typing/RustEnterInStringLiteralHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        // find the PsiElement at the caret
+        val elementAtCaret = file.findElementAt(caretOffset) ?: return Result.Continue
+
+        when (elementAtCaret.elementType) {
+            STRING_LITERAL,
+            BYTE_STRING_LITERAL -> {
+                // If we are inside string literal, add trailing '\' just before caret
+                var offset = caretOffset
+                editor.document.insertString(offset++, "\\")
```
matklad: This is always in bounds, right?

src/main/kotlin/org/rust/ide/typing/RustEnterInStringLiteralHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+                caretOffsetRef.set(offset)
+                return Result.DefaultForceIndent
+            }
+
+            RAW_STRING_LITERAL,
+            RAW_BYTE_STRING_LITERAL -> {
+                return Result.DefaultSkipIndent
+            }
+
+            else -> return Result.Continue
```
matklad: return can be pulled out of `when`

src/main/kotlin/org/rust/ide/typing/RustEnterInStringLiteralHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    private fun isInStringOrRawStringLiteral(editor: Editor, offset: Int): Boolean {
+        if (offset < 1) return false
+        val quotedHandler = RustQuoteHandler()
+        val highlighter = (editor as EditorEx).highlighter
+        val iterator = highlighter.createIterator(offset - 1)
+        return quotedHandler.isDeepInsideLiteral(iterator, offset)
```
matklad: This `isDeepInsideLiteral` is only used here. So perhaps instead of creating highlighting iterator here we ca just move this function inside this class? Or perhaps just in some utility module?

It does not feels right to create a bunch of statefull objects just to be able to call what should be a pure function :)

src/test/resources/org/rust/ide/typing/string/fixtures/after_opening_after.rs
mkaput: Yep, but I don't know yet how to implement this :3

src/main/kotlin/org/rust/lang/core/psi/impl/RustRawStringLiteralImpl.kt
mkaput: This is probably temporary hack, I'm thinking yet on how to implement this as computing offsets may fail in this case.

The problem is, we are fed with highlighter iterator output, which is polluted with escape sequence tokens, so for `b"foo\nba<caret>r"` we won't get `<BYTE_STRING_LITERAL>(b"foo\\nbar")`, but `<BYTE_STRING_LITERAL>(bar")`.

src/main/kotlin/org/rust/ide/typing/RustEnterInStringLiteralHandler.kt
mkaput: `offset` naturally, because we are inside some PSI element, so `insertString` will succeed.
If `insertString` succeeds, then `offset++; set(offset)` also succeeds, because we have added `\` just a moment ago.

src/main/kotlin/org/rust/lang/core/psi/impl/RustRawStringLiteralImpl.kt
matklad: One more thing to keep in mind: offsets for `""` (an empty string) have `null` as value, so the value of this string literal is `null` instead of `""`.

mkaput: 🐛

mkaput: > One more thing to keep in mind: offsets for "" (an empty string) have null as value, so the value of this string literal is null instead of "". 

Fixed, where was the issue spotted?

mkaput:
ready for review... finally

src/main/kotlin/org/rust/lang/core/psi/impl/RustRawStringLiteralImpl.kt
matklad: https://github.com/intellij-rust/intellij-rust/pull/621#r75926452, can add a test to https://github.com/intellij-rust/intellij-rust/blob/master/src/test/resources/org/rust/ide/annotator/fixtures/items/invalid_module_declarations.rs

matklad:
Wonderful!

matklad:
It'd be cool to automatically add or remove hashes to both sides of raw string literal if you edit either one (the same way as closing and opening xml tags are edited simultaneously).

mkaput:
> It'd be cool to automatically add or remove hashes to both sides of raw string literal if you edit either one (the same way as closing and opening xml tags are edited simultaneously).

Nice idea, will open issue

Issue #588, opened 2016-08-09
Add De Morgan's Laws intentions


mkaput:
Implement boolean expression manipulating intentions like one in IntelliJ

matklad:
:apple:

mkaput:
Yeah, typing on SwiftKey, so much fun

zjhmale:
https://blog.jetbrains.com/idea/2014/09/the-inspection-connection-issue-2/

seems a lot work to do here :scream:

mkaput:
You can copy paste this from IntelliJ, PyCharm, Groovy and Kotlin implementations :p

mkaput:
Also De Morgan laws are pretty simple and straightforward. Implementation should be recursive over rust expression. Typing infra to check whether expression is boolean already works and it's api is very very friendly.

zjhmale:
i will checkout the intellij implementations 😁

zjhmale:
@mkaput and maybe a intention instead of a inspection is better? i mean boolean sequence just have DeMorgan equivalent, but there not exists a best one.

mkaput:
> @mkaput and maybe a intention instead of a inspection is better?

yeah, wrong words

zjhmale:
@mkaput here is a problem that we can not easily get the operator for a `RustBinaryExprElement` just like [python-plugin](https://github.com/JetBrains/intellij-community/blob/master/python/src/com/jetbrains/python/codeInsight/intentions/PyDemorganIntention.java#L56), [groovy-plugin](https://github.com/JetBrains/intellij-community/blob/master/plugins/groovy/src/org/jetbrains/plugins/groovy/intentions/control/DemorgansLawIntention.java#L57) and [kotlin-plugin](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertNegatedBooleanSequenceIntention.kt#L60).

![2333333333333333](https://cloud.githubusercontent.com/assets/6234553/17561728/00817d00-5f5a-11e6-9b49-78c66437fcdd.png)

mkaput:
working on this

mkaput:
Here you are ^ @zjhmale  :smile:

zjhmale:
@mkaput thx! 😃

PR #591, opened 2016-08-10
(PSI): provide operator accessor for binary exprs

mkaput:
cc #588

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RustBinaryExprElementEx.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RustBinaryExprElementEx.kt
@@ -1,0 +1,12 @@
+package org.rust.lang.core.psi
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.tree.IElementType
+import org.rust.lang.core.psi.RustTokenElementTypes.BINARY_OPS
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/psi/RustBinaryExprElementEx.kt
```diff
@@ -0,0 +1,12 @@
+package org.rust.lang.core.psi
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.tree.IElementType
+import org.rust.lang.core.psi.RustTokenElementTypes.BINARY_OPS
+import org.rust.lang.core.psi.util.elementType
+
+val RustBinaryExprElement.operator: PsiElement
+    get() = node.findChildByType(BINARY_OPS)!!.psi // guaranteed to be not-null by parser; null == bug
```
matklad: It's better to use `reuireNotNull(thing) { "why is this impossible" }` instead of comments, because one day the impossible will happen and it would be convenient to see this justification in the stack trace :)

mkaput: fixed

matklad:
Thanks!

zjhmale:
@mkaput thx!

PR #594, opened 2016-08-11
Add De Morgan's Laws equivalent intention

### LINKED ISSUE #588: Add De Morgan's Laws intentions

zjhmale:
Fixes #588 

snapshot

based on [python-plugin](https://github.com/JetBrains/intellij-community/blob/master/python/src/com/jetbrains/python/codeInsight/intentions/PyDemorganIntention.java).

honestly, this intention is implemented use text level manipulate instead of PSI level, but it is quit robust i think.


src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.tree.IElementType
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.elementType
+import org.rust.lang.core.psi.util.parentOfType
+
+class DemorgansLawIntention : PsiElementBaseIntentionAction() {
+    override fun getText() = "DeMorgan Law"
```
mkaput: I would introduce base class [MutablyNamedIntention](https://upsource.jetbrains.com/idea-ce/file/idea-ce-cfc493fae3127d133589d446cc535d8ac51b72c5/plugins/IntentionPowerPak/src/com/siyeh/ipp/base/MutablyNamedIntention.java?nav=stub:6:focused) and implement intention's name like [here](https://upsource.jetbrains.com/idea-ce/file/idea-ce-cfc493fae3127d133589d446cc535d8ac51b72c5/plugins/IntentionPowerPak/src/com/siyeh/ipp/bool/DemorgansIntention.java?nav=1148:1607:focused&line=0), so we would get nice `Replace '&&' with '||'` and vice versa names.

Of course, family name should stay as it is (sans typos): `De Morgan's laws`

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,107 @@
...
+            convertConjunctionExpression(rhs as RustBinaryExprElement, opType)
+        } else {
+            convertLeafExpression(rhs)
+        }
+
+        val flippedConjunction = when (opType == RustTokenElementTypes.ANDAND) {
+            true -> " || "
+            else -> " && "
+        }
+        return lhsText + flippedConjunction + rhsText
```
mkaput:
```kotlin
val flippedConjunction = if (opType == RustTokenElementTypes.ANDAND) "||" else "&&"
return "$lhsText $flippedConjunction $rhsText"
```
looks better imho

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private fun getNegated(expression: PsiElement): RustExprElement? {
+        return (expression as RustUnaryExprElement).expr
+    }
+
+    private fun isNegation(expression: PsiElement): Boolean {
+        if (expression !is RustUnaryExprElement) return false
+        val op = expression.excl ?: return false
+        return op.elementType == RustTokenElementTypes.EXCL
```
mkaput: grammar should guarantee that `excl` is always `RustTokenElementTypes.EXCL`

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,107 @@
...
+        if (opType == RustTokenElementTypes.ANDAND || opType == RustTokenElementTypes.OROR) {
+            return Context(binExpr, opType)
+        }
+        return null
+    }
+
+    override fun invoke(project: Project, editor: Editor?, element: PsiElement) {
+        val ctx = findContext(element) ?: return
+        var binExpr = ctx.binaryExpr
+        val opType = ctx.binaryOpType
```
mkaput: `Context` is data class, so we can unpack it

```kotlin
val (binExpr, opType) = findContext(element) ?: return
```

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        while (binExpr.parent is RustBinaryExprElement) {
+            binExpr = binExpr.parent as RustBinaryExprElement
+        }
```
mkaput: Rather than mutating `binExpr` I would introduce new immutable variable with more meaningful name: `topBinExpr` or sth like that.

mkaput:
How does it work over more complicated expressions?
```rust
2 + 2 == 2 && (foo.bar() || !(78 < 90 || 90 > 78)) || (20 >= 50 && 40 != 20)
```

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: ah, right. i forgot to use string template 😂

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: ok, i will remove this line.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: thx, just like ML family language.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: fair enough.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: yeah, fair enough, i overlooked this design first time 😂 . and kotlin plugin introduces two intention to do the same thing, a little bit exaggrated.

src/main/kotlin/org/rust/ide/intentions/MutablyNamedIntention.kt
```diff
@@ -0,0 +1,14 @@
+package org.rust.ide.intentions
+
+import com.intellij.codeInsight.intention.PsiElementBaseIntentionAction
+import com.intellij.psi.PsiElement
+
+abstract class MutablyNamedIntention : PsiElementBaseIntentionAction() {
```
mkaput: Let's stick with our naming conventions -> `RustMutablyNamedIntention`

src/main/kotlin/org/rust/ide/intentions/MutablyNamedIntention.kt
```diff
@@ -0,0 +1,14 @@
...
+    private var m_text = ""
+
+    protected abstract fun getTextForElement(element: PsiElement): String
+
+    override fun getText() = m_text
+    override fun getFamilyName() = text
```
mkaput: Emm, it looks like this doesn't work.
1) `m_text` is never set
2) Family name depends on PSI element, it's nonsense.

src/main/kotlin/org/rust/ide/intentions/MutablyNamedIntention.kt
```diff
@@ -0,0 +1,14 @@
...
+    override fun startInWriteAction() = true
```
mkaput: Hmm, I don't think this should be placed here.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
mkaput: Not exactly; in Kotlin this is just a sugar:
```kotlin
val (a, b, c) = Foo()
```
=>
```kotlin
val __foo = Foo()
val a = __foo.component1()
val b = __foo.component2()
val c = __foo.component3()
```

Not so powerful as in ML/Rust/Haskell/Scala, but still pretty neat

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: @mkaput and i think there is no need to introduce a base class here, i mean to dynamic change the intention text by adding a base class is a little bit overstate.

mkaput: Currently, yes. But I think this pattern could be useful in the future. You can merge that class if you like, it's no problem for me :smile:

zjhmale:
@mkaput for the complicated expression

snapshot

src/main/kotlin/org/rust/ide/intentions/MutablyNamedIntention.kt
zjhmale: @mkaput just delete the base class 😂 , i think add a base class is a little bit overstate.

src/main/kotlin/org/rust/ide/intentions/MutablyNamedIntention.kt
zjhmale: other intentions all use text as Family name so i just "stick" with the convention, and there is no family for this intention i think.

mkaput:
So: 
1. the right side of `&&` expression isn't negated. I mean it shouldn't be wrapped in negation, but it should be negated recursively.
2. I think, relational expressions should be negated using mathematical counterparts, i.e. `==` -> `!=`, `<` -> `>=` etc.

src/main/kotlin/org/rust/ide/intentions/MutablyNamedIntention.kt
mkaput: The problem is that `m_text` is not set, so `getText()` (and `getFamilyName()`) return empty string.

zjhmale:
@mkaput (2) is done, you can checkout the test cases. (1) is a little bit tricky and hard to implement.

mkaput:
Never, user can nest parents indefinitely. You must fire negation procedure recursively over children exprs

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,138 @@
...
+    override fun startInWriteAction() = true
+
+    fun getTextForElement(element: PsiElement) {
+        val binaryExpression = element as RustBinaryExprElement
+        m_text = when (binaryExpression.operatorType) {
+            RustTokenElementTypes.ANDAND -> "DeMorgan's Law, Replace '&&' with '||'"
+            RustTokenElementTypes.OROR -> "DeMorgan's Law, Replace '||' with '&&'"
+            else -> ""
+        }
+    }
```
matklad: You don't need to override `getText`. This `m_text` thing is basically what is already done in the parent class. Instead, you can just assign to `text` directly:

```
text = when ...
```

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.parentOfType
+
+class DemorgansLawIntention : PsiElementBaseIntentionAction() {
+    private var m_text = ""
+    override fun getText() = m_text
+    override fun getFamilyName() = text
+    override fun startInWriteAction() = true
+
+    fun getTextForElement(element: PsiElement) {
```
matklad: `setTextForElement` would be a much better name here. Also, this `fun` can be private.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
```diff
@@ -0,0 +1,138 @@
...
+            return true
+        }
+        return false
+    }
+
+    private fun isConjunctionExpression(expression: RustExprElement, opType: IElementType): Boolean {
+        return expression is RustBinaryExprElement && expression.operatorType == opType
+    }
+
+    private fun getNegated(expression: PsiElement): RustExprElement? {
```
matklad: It's better to inline this function because it is used only once and it is not obvious what it does without the preceding check.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: ah, cool.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: cool.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: @matklad inline this function just get a warning

image

mkaput: @matklad meant to put function's code directly inside callee's code

zjhmale:
@mkaput (1) is done i think, i will do some code refactor to make this code neat tomorrow.

src/test/kotlin/org/rust/ide/intentions/DemorgansLawIntentionTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    fun testComplex7() = checkByFile {
+        openFileInEditor("complex7.rs")
+        myFixture.launchAction(DemorgansLawIntention())
+    }
+
+    fun testComplex8() = checkByFile {
+        openFileInEditor("complex8.rs")
+        myFixture.launchAction(DemorgansLawIntention())
+    }
```
mkaput: this is pretty repetitive, also what do you think about rerunning this again in order to check that the inspection is reversible?

mkaput:
I am thinking about introducing some abstraction called `ExpressionTransformer` which would allow to transform expressions using visitors, but this might be out-of-scope of this PR (probably would require total rewrite :P), also I have to polish this idea

src/test/kotlin/org/rust/ide/intentions/DemorgansLawIntentionTest.kt
zjhmale: i check reversible in seperated fixtures, and since Kotlin doesn't have a Macro system to generate boilerplate code i will find another way to eliminate the repetitivity.

src/main/kotlin/org/rust/ide/intentions/DemorgansLawIntention.kt
zjhmale: ah, fair enough, i misunderstood the purpose 😂

src/test/kotlin/org/rust/ide/intentions/DemorgansLawIntentionTest.kt
mkaput: Our common pattern is to introduce private doTest method

zjhmale: yeah, i added one 😄 
>Our common pattern is to introduce private doTest method
Add a line note

zjhmale:
poke @mkaput @matklad

matklad:
@zjhmale my vacation will end on Monday, so I'll able to give a thorough review then. Though I'd be happy to merge the PRs right away if @mkaput gives an r+ :)

mkaput:
Heh, I sincerely have currently no time to watch what's going on here :disappointed: , so it's up to you @matklad

zjhmale:
that's all right! i also busy with preparing for some tests, just make sure you guys do not forget about this PR 😄

mkaput:
@zjhmale good luck then!

zjhmale:
@mkaput thx!

matklad:
LGTM!

Issue #542, opened 2016-07-21
Create "Duplicate trait item inspection"


matklad:
This inspection should trigger if there are duplicate methods or types in a trait definition. 

See https://github.com/intellij-rust/intellij-rust/commit/b0bc03d126f1c165868bef4ef40ccb9cbc9ef3b4 for the example of how to implement an inspection.

mbStavola:
I'd like to attempt this.

PR #596, opened 2016-08-11
(INSP): Add duplicate trait item inspections

mbStavola:
Addresses #542 

Added `RustDuplicateInspectionTool` to facilitate the creation of inspections to find duplicate occurrences. 

Rewrote`RustDuplicateStructFieldInspection` to extend `RustDuplicateInspectionTool`.

Added inspections for duplicate trait methods, constants, and associated types (respectively `RustDuplicateTraitMethodInspection`, `RustDuplicateTraitConstantInspection`, and `RustDuplicateTraitConstantInspection`).

Feedback is very much appreciated.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/duplicates/RustDuplicateStructFieldInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/duplicates/RustDuplicateStructFieldInspection.kt
@@ -2,21 +2,16 @@ package org.rust.ide.inspections.duplicates
 
 import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.codeInspection.ProblemsHolder
-import com.intellij.psi.PsiElement
-import org.rust.ide.inspections.RustLocalInspectionTool
-import org.rust.lang.core.psi.RustElementVisitor
+import org.rust.lang.core.psi.RustFieldDeclElement
 import org.rust.lang.core.psi.RustFieldsOwner
 import org.rust.lang.core.psi.fields
 
-class RustDuplicateStructFieldInspection : RustLocalInspectionTool() {
-    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = object : RustElementVisitor() {
-        override fun visitElement(element: PsiElement) {
-            if (element is RustFieldsOwner) {
-                for (dupe in element.fields.findDuplicates()) {
-                    holder.registerProblem(dupe, "Duplicate field <code>#ref</code>", ProblemHighlightType.GENERIC_ERROR_OR_WARNING)
-                }
-            }
+class RustDuplicateStructFieldInspection : RustDuplicateInspectionTool() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        createInspection(RustFieldsOwner::getFields) {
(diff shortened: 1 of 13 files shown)


src/main/kotlin/org/rust/ide/inspections/duplicates/RustDuplicateItem.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustElementVisitor
+import org.rust.lang.core.psi.RustNamedElement
+
+class RustDuplicateItem {
+    companion object {
+        inline fun <reified T : Any> createInspection(
+                crossinline memberList: T.() -> List<RustNamedElement>,
+                crossinline holderFunc: (PsiElement) -> Unit
```
matklad: This can be made even more generic to remove some casts:

```
crossinline memberList: T.() -> List<E>,
crossinline holderFunc: (E) -> Unit
```

src/main/kotlin/org/rust/ide/inspections/duplicates/RustDuplicateStructFieldInspection.kt
```diff
@@ -2,21 +2,18 @@ package org.rust.ide.inspections.duplicates
...
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+            RustDuplicateItem.createInspection(RustFieldsOwner::getFields) {
+                holder.registerProblem(it, "Duplicate field <code>#ref</code>", ProblemHighlightType.GENERIC_ERROR_OR_WARNING)
             }
-        }
-    }
 }
 
+//TODO: Why do we need this extra layer of indirection? Why can't we rely on the property's getter via reference?
+private fun RustFieldsOwner.getFields(): List<RustFieldDeclElement> = fields
```
matklad: Not sure what's going on here either.

src/main/kotlin/org/rust/ide/inspections/duplicates/RustDuplicateItem.kt
```diff
@@ -0,0 +1,20 @@
+package org.rust.ide.inspections.duplicates
+
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustElementVisitor
+import org.rust.lang.core.psi.RustNamedElement
+
+class RustDuplicateItem {
```
matklad: Looks like you use this class simply as a wrapper for a function. In this case, you can use an object instead:

```Kotlin
object RustDuplicateItem {
    inline fun <reified T : Any> createInspection( ....
}
```

Though using objects purely for namespaces is somewhat discouraged: https://github.com/yole/kotlin-style-guide/issues/6

Alternatively, you can make this class an abstract inheritor of `RustLocalInspectionTool` and then make `createInspection` function protected. That way you'd be able to enforce privacy better.

Either solution is fine imo.

src/main/kotlin/org/rust/ide/inspections/duplicates/RustDuplicateStructFieldInspection.kt
mbStavola: Yeah, this was probably the single most confusing part of this PR.

IntelliJ seems to be okay with letting you pass `RustFieldsOwner::fields` to `createInspection()`, but the compiler doesn't like that. I really only figured it out after taking a look at the generated bytecode :(

matklad:
Thanks!

PR #605, opened 2016-08-17
lookup elements for multiline declaration

### LINKED ISSUE #604: Lookup elements should work for multiline declaration
For 

```
pub fn get_resolved_packages<'a>(resolve: &Resolve,
                                 registry: PackageRegistry<'a>)
                                 -> PackageSet<'a> {
```

lookup element looks like

nav_info

We should try to show all parameters on the single line! 

The code is here: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt

zjhmale:
Fixes #604

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -7,12 +7,13 @@ import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.RustFile
 import org.rust.lang.core.psi.util.module
 import org.rust.lang.core.psi.util.parentOfType
+import java.util.regex.Pattern
 
 fun RustNamedElement.createLookupElement(): LookupElement {
     return when (this) {
         is RustFnElement -> LookupElementBuilder.createWithIcon(this)
             .withLookupString(name ?: "")
-            .withTailText(parameters?.text ?: "()")
+            .withTailText(parameters?.text?.replace(Pattern.compile("\\s+").toRegex(), " ") ?: "()")
             .withTypeText(retType?.type?.text ?: "()")
         is RustConstItemElement -> LookupElementBuilder.createWithIcon(this)
             .withLookupString(name ?: "")


src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -7,12 +7,13 @@ import org.rust.lang.core.psi.*
...
 import org.rust.lang.core.psi.util.module
 import org.rust.lang.core.psi.util.parentOfType
+import java.util.regex.Pattern
 
 fun RustNamedElement.createLookupElement(): LookupElement {
     return when (this) {
         is RustFnElement -> LookupElementBuilder.createWithIcon(this)
             .withLookupString(name ?: "")
-            .withTailText(parameters?.text ?: "()")
+            .withTailText(parameters?.text?.replace(Pattern.compile("\\s+").toRegex(), " ") ?: "()")
```
mkaput: why regexps :cry:

zjhmale: that's the best way to eliminate extra whitespaces come into my mind 😂

matklad:
Thanks!

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
matklad: @mkaput do you have a better solution in mind?

mkaput:
```kotlin
parameters?.text?.let { StringUtil.tokenize(it, " \t\n\r\f").joinToString(" ") } ?: "()"
```
is first thing which comes to my mind. Personally I don't like to use regexps for trivial things

matklad: I don't like to use regexps for non-trivial things :) And I have to google up `\f`.

Either solution is good.

zjhmale: cool, also googled `\f` 😂 , first time to see this special character.

Issue #308, opened 2016-03-28
Support multirust/rustup


matklad:
http://www.rustup.rs/

matklad:
In particular, use `RUSTUP_TOOLCHAIN` to switch toolchains.

matklad:
Rough sketch of a design:

Toolchain is determined by a `toolchain_home` path. if `toolchain_path/rustup` or `toolchain_path/multirust` exist, toolchain is *rustup toolchain*. For a rustup toolchain and additional string field `toolchain_version` is stored. The field should be displayed in the UI. It can be either null (the default toolchain is used) or some variant from `rustup toolchain list` / `multirust list-toolchains`. If it is not null than `RUSTUP_TOOLCHAIN` environmental variable must be set for every cargo command.

matklad:
Persistence and UI parts are here: https://github.com/intellij-rust/intellij-rust/tree/master/src/main/kotlin/org/rust/cargo/project/settings

Cargo execution parts are here: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/commands/Cargo.kt

siler:
I've started poking around at this for fun despite the current state of the configuration folder RFC. Given that rustup will be the official Rust toolchain manager would it make sense to create an application component for configuring rustup and then change the project-level configurations to either select from currently installed toolchains or install a new one? 

I'm not sure that rustup is currently advanced enough for all the features we need (i.e., show me the directories in which everything resides in a convenient json/toml format) but for now we can infer based on the current standards.

matklad:
Nice!

I think a good and isolated first step in this direction would be just to figure out if we are running rustup and show toolchain name in the settings (*without* the ability to choose it).

Application component would make sense if we persisted some information about toolchains, but it seems like we don't need to yet? I think that we can just query `rustup` about available toolchains when we need the list, and store a project toolchain in per-project settings. 

>i.e., show me the directories in which everything resides in a convenient json/toml format

And what directories to we need? It looks like that using rustup amounts to setting `RUSTUP_TOOLCHAIN` environmental variable when we call `~/.cargo/bin/rustup`.

siler:
> I think a good and isolated first step in this direction would be just to figure out if we are running rustup and show toolchain name in the settings (without the ability to choose it).

Agreed, I think that's fair.

> Application component would make sense if we persisted some information about toolchains, but it seems like we don't need to yet?

[...]

PR #620, opened 2016-08-23
show rustup version in settings panel

zjhmale:
First step for #308 

image

seems to achieve the whole supporting is a long journey, so maybe tackle it with seperated PRs is better 😃

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
@@ -14,8 +14,11 @@ import java.io.File
 
 data class RustToolchain(val location: String) {
 
+    private fun hasExecuteable(exec: String): Boolean =
+        File(pathToExecutable(exec)).canExecute()
+
     fun looksLikeValidToolchain(): Boolean =
-        File(pathToExecutable(CARGO)).canExecute()
+        hasExecuteable(CARGO)
 
     fun containsMetadataCommand(): Boolean {
         check(!ApplicationManager.getApplication().isDispatchThread)
@@ -46,6 +49,17 @@ data class RustToolchain(val location: String) {
             .withParameters("--version", "--verbose")
 
         return runExecutableAndProcessStdout(cmd) { parseRustcVersion(it) }
+    }
+
+    fun queryRustupVersion(): Version? {
+        check(!ApplicationManager.getApplication().isDispatchThread)
+        if (!hasExecuteable(RUSTUP)) return null
+
+        val cmd = GeneralCommandLine()
+            .withExePath(pathToExecutable(RUSTUP))
+            .withParameters("--version")
+
+        return runExecutableAndProcessStdout(cmd) { parseRustupVersion(it) }
     }
 
     private fun <T> runExecutableAndProcessStdout(cmd: GeneralCommandLine, extractor: (List<String>) -> T): T? {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -176,6 +188,23 @@ private fun parseRustcVersion(lines: List<String>): Version? {
...
+    //  ```
+    //  rustup 0.5.0 (4be1012 2016-07-30)
+    //  ```
+    val releaseRe = """rustup (\d+)\.(\d+)\.(\d+) \(.*\)""".toRegex()
+
+    val match = lines.mapNotNull { releaseRe.matchEntire(it) }.firstOrNull() ?: return null
+
+    val major = match.groups[1]!!.value.toInt()
+    val minor = match.groups[2]!!.value.toInt()
+    val build = match.groups[3]!!.value.toInt()
```
mkaput: What if user has `rustup` binary which isn't rustup at all? I mean this will fail when output doesn't match regexp.

zjhmale: if the output doesn't match regexp, it will return null for `Version?` and show `N/A` on the panel just like `rustc` and `cargo`.

mkaput: Uh, must have to better get into regexp api then 😅

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -14,8 +14,8 @@ import java.io.File
 
 data class RustToolchain(val location: String) {
 
-    fun looksLikeValidToolchain(): Boolean =
-        File(pathToExecutable(CARGO)).canExecute()
+    fun looksLikeValidToolchain(exec: String = CARGO): Boolean =
```
matklad: Let's not add this `exec` parameter and instead create `hasExecuteable(exec: String)` *private* helper which can be used by both `looksLikeValidToolchain` and `queryRustupVersion`

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -14,8 +14,11 @@ import java.io.File
 
 data class RustToolchain(val location: String) {
 
+    private fun hasExecuteable(exec: String): Boolean =
```
zjhmale: hope not misunderstand your purpose here :sweat_smile:

matklad: Yes, this is exactly what I had in mind! 

And now that this is written this way, I see that the old code has probably had a bug, because it did check for the presence of `cargo` and not `rustc` in `queryRustcVersion`!

zjhmale: @matklad should i make another PR for this?

matklad: No, I am already auditing the usages of `looksLikeValidToolchains` myself :)

zjhmale: cool 😄

PR #621, opened 2016-08-23
Improve explicit path attributes on module declarations

kumbayo:
Improve handling of explicit path attributes on module declarations

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
@@ -29,14 +29,16 @@ class RustItemsAnnotator : Annotator {
     }
 
     private fun checkModDecl(holder: AnnotationHolder, modDecl: RustModDeclItemElement) {
-        if (modDecl.isPathAttributeRequired && modDecl.pathAttribute == null) {
+        val pathAttribute = modDecl.pathAttribute
+
+        if (modDecl.isPathAttributeRequired && pathAttribute == null) {
             val message = "Cannot declare a non-inline module inside a block unless it has a path attribute"
             holder.createErrorAnnotation(modDecl, message)
             return
         }
 
         val containingMod = modDecl.containingMod ?: return
-        if (!containingMod.ownsDirectory) {
+        if (!containingMod.ownsDirectory && pathAttribute == null) {
             // We don't want to show the warning if there is no cargo project
             // associated with the current module. Without it we can't know for
             // sure that a mod is not a directory owner.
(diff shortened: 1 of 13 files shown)


src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
```diff
@@ -29,14 +29,16 @@ class RustItemsAnnotator : Annotator {
...
             holder.createErrorAnnotation(modDecl, message)
             return
         }
 
         val containingMod = modDecl.containingMod ?: return
-        if (!containingMod.ownsDirectory) {
+        if (!containingMod.ownsDirectory && pathAttribute == null) {
             // We don't want to show the warning if there is no cargo project
             // associated with the current module. Without it we can't know for
             // sure that a mod is not a directory owner.
```
kumbayo: I do not know how to properly test for error annotations, so i did not create a test for this.

src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
```diff
@@ -29,14 +29,16 @@ class RustItemsAnnotator : Annotator {
     }
 
     private fun checkModDecl(holder: AnnotationHolder, modDecl: RustModDeclItemElement) {
-        if (modDecl.isPathAttributeRequired && modDecl.pathAttribute == null) {
+        val pathAttribute = modDecl.pathAttribute
+
+        if (modDecl.isPathAttributeRequired && pathAttribute == null) {
```
kumbayo: RustDocAndAttributeOwner.lookupStringValueForKey("path") returns null for this code.
So in this case we still output the old error message.
Maybe it should return "" instead since the attribute is there but empty?
```rust
#[path=""]
mod foo;
```

src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
matklad: There is `src/test/resources/org/rust/ide/annotator/fixtures/items/invalid_module_declarations.rs` which is checked by `RustItemAnnotatorTest`.

src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
matklad: Nice catch!

This will be fixed perhaps together with this PR: https://github.com/intellij-rust/intellij-rust/pull/585#discussion_r75961718

matklad:
Awesome! Would you mind adding an annotator test (just another case in `invalid_module_declarations.rs`)?

kumbayo:
I added an annotator test.
For this I had to slightly modify RustAnnotatorTestBase.
I hope thats OK.

matklad:
Thanks!

matklad:
Just a heads up: we've decided to drop authors.md file because it is a hassle to maintain, but we mention all contributors in the changelog ( https://intellij-rust.github.io/thisweek/ ).

src/main/kotlin/org/rust/ide/annotator/RustItemsAnnotator.kt
mkaput: 5f1a681

PR #624, opened 2016-08-24
(MISC): extend GotoSuper action to work for iml methods. Fix #541

### LINKED ISSUE #541: Extend GotoSuper action to work for iml methods
Currently Goto super forks for modules. I think it should work for methods as well. 

code:
https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/navigation/goto/RustGotoSuperHandler.kt

tests:
https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/navigation/RustNavigationTest.kt

atsky:
Fix for #541.
@matklad, please review

### DIFF
--- a/src/main/kotlin/org/rust/ide/navigation/goto/RustGotoSuperHandler.kt
+++ b/src/main/kotlin/org/rust/ide/navigation/goto/RustGotoSuperHandler.kt
@@ -4,9 +4,14 @@ import com.intellij.lang.LanguageCodeInsightActionHandler
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RustImplItemElement
+import org.rust.lang.core.psi.RustImplMethodMemberElement
 import org.rust.lang.core.psi.RustMod
+import org.rust.lang.core.psi.RustNamedElement
 import org.rust.lang.core.psi.impl.RustFile
+import org.rust.lang.core.psi.impl.mixin.superMethod
 import org.rust.lang.core.psi.util.parentOfType
+import org.rust.lang.core.psi.util.trait
 
 class RustGotoSuperHandler : LanguageCodeInsightActionHandler {
     override fun startInWriteAction() = false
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/ide/navigation/goto/RustGotoSuperHandler.kt
```diff
@@ -4,20 +4,36 @@ import com.intellij.lang.LanguageCodeInsightActionHandler
...
+            val focusedMod = focusedElement.parentOfType<RustMod>(strict = false) ?: return
+            val superMod = focusedMod.`super` ?: return
 
-        superMod.navigate(true)
+            superMod.navigate(true)
+        }
+    }
+
+    private fun getSuperMethod(focusedMethod: RustImplMethodMemberElement): RustNamedElement? {
+        val rustImplItem = focusedMethod.parentOfType<RustImplItemElement>(strict = false) ?: return null
```
matklad: No need in `strict = false` here, because method and impl can never be the same PSI element. 

Also perhaps it makes sense to introduce an extension methods to get particular ancestor elements in a typed manner? Something like `methodImpl.imp`, `methodDefinition.trait`. Currently we use ad-hoc `parentOfType` and chains of `.parent` everywhere, looks like this is not the best approach.

src/main/kotlin/org/rust/ide/navigation/goto/RustGotoSuperHandler.kt
```diff
@@ -4,20 +4,36 @@ import com.intellij.lang.LanguageCodeInsightActionHandler
...
 
 class RustGotoSuperHandler : LanguageCodeInsightActionHandler {
     override fun startInWriteAction() = false
 
     override fun invoke(project: Project, editor: Editor, file: PsiFile) {
         val focusedElement = file.findElementAt(editor.caretModel.offset) ?: file ?: return
 
-        val focusedMod = focusedElement.parentOfType<RustMod>(strict = false) ?: return
-        val superMod = focusedMod.`super` ?: return
+        val focusedMethod = focusedElement.parentOfType<RustImplMethodMemberElement>(strict = false)
```
matklad: Technically, we should search *simultaneously* for both methods and modules: 

```
impl Foo for Bar {
     fn foo() {
        mod m {  mod n { <caret> /* ideally, should find n and not foo */ } }
     }
}
```

Though I think this is an obscure corner case, so perhaps there is no need in handling it.

src/main/kotlin/org/rust/ide/navigation/goto/RustGotoSuperHandler.kt
atsky: IMO `parentOfType<RustImplItemElement>` is more readable here. When here will be more logic we can create extension method.

matklad:
:+1:

PR #628, opened 2016-08-25
Toolchain dedup

matklad:
This reduces duplication in the `RustToolchain` and also makes it easier to use, because there is only one method with `isDispatchThread` precondition. 

@zjhmale please review.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
+++ b/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
@@ -12,7 +12,6 @@ import com.intellij.ui.JBColor
 import com.intellij.util.Alarm
 import org.rust.cargo.project.settings.RustProjectSettingsService
 import org.rust.cargo.toolchain.RustToolchain
-import org.rust.cargo.toolchain.Version
 import javax.swing.JCheckBox
 import javax.swing.JLabel
 import javax.swing.JPanel
@@ -102,43 +101,33 @@ class RustProjectSettingsPanel : JPanel() {
     private fun scheduleVersionUpdate(toolchainLocation: String) {
         versionUpdateAlarm.cancelAllRequests()
         versionUpdateAlarm.addRequest({
-            val toolchain = RustToolchain(toolchainLocation)
-
-            val rustc = toolchain.queryRustcVersion()
-            val cargo = toolchain.queryCargoVersion()
-            val rustup = toolchain.queryRustupVersion()
-
-            updateVersion(rustc, cargo, rustup)
-
+            val versionInfo = RustToolchain(toolchainLocation).queryVersions()
+            updateVersion(versionInfo)
         }, versionUpdateDelayMillis)
     }
 
-    private fun updateVersion(rustc: Version?, cargo: Version?, rustup: Version?) {
+    private fun updateVersion(info: RustToolchain.VersionInfo) {
         ApplicationManager.getApplication().invokeLater({
-            if (!Disposer.isDisposed(disposable)) {
-                if (rustc != null) {
-                    rustVersion.text        = rustc.release
-                    rustVersion.foreground  = JBColor.foreground()
-                } else {
-                    rustVersion.text        = "N/A"
-                    rustVersion.foreground  = JBColor.RED
-                }
+            if (Disposer.isDisposed(disposable)) return@invokeLater
 
-                if (cargo != null) {
-                    cargoVersion.text       = cargo.release
-                    cargoVersion.foreground = if (cargo >= RustToolchain.CARGO_LEAST_COMPATIBLE_VERSION) JBColor.foreground() else JBColor.RED
-                } else {
-                    cargoVersion.text       = "N/A"
-                    cargoVersion.foreground = JBColor.RED
-                }
+            val labelToVersion = listOf(
+                rustVersion to info.rustc?.semver,
+                cargoVersion to info.cargo,
+                rustupVersion to info.rustup
+            )
 
-                if (rustup != null) {
-                    rustupVersion.text       = rustup.release
-                    rustupVersion.foreground = JBColor.foreground()
+            for ((label, version) in labelToVersion) {
+                if (version == null) {
+                    label.text = "N/A"
+                    label.foreground = JBColor.RED
                 } else {
-                    rustupVersion.text       = "N/A"
-                    rustupVersion.foreground = JBColor.RED
+                    label.text = version.parsedVersion
+                    label.foreground = JBColor.foreground()
                 }
+            }
+
+            if (!info.cargoHasMetadataCommand) {
+                cargoVersion.foreground = JBColor.RED
             }
         }, ModalityState.any())
     }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
```diff
@@ -102,43 +101,33 @@ class RustProjectSettingsPanel : JPanel() {
...
 
-                if (rustup != null) {
-                    rustupVersion.text       = rustup.release
-                    rustupVersion.foreground = JBColor.foreground()
-                } else {
-                    rustupVersion.text       = "N/A"
-                    rustupVersion.foreground = JBColor.RED
-                }
+            if (!info.cargoHasMetadataCommand) {
+                cargoVersion.foreground = JBColor.RED
```
mkaput: I think it might be nice to provide user explicit reason why we don't like he's cargo.

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -108,31 +100,16 @@ data class RustToolchain(val location: String) {
     }
 }
 
+open class Version(val major: Int, val minor: Int, val patch: Int) {
+    override fun toString(): String = "$major.$minor.$patch"
+}
```
mkaput: looks like a duplicate of [SemVer](https://upsource.jetbrains.com/idea-ce/file/idea-ce-7771e0e8bc3c66a436063345d56b46f2fc670cd4/platform/util/src/com/intellij/util/text/SemVer.java?nav=0:0:focused) class

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -14,57 +14,55 @@ import java.io.File
...
-        }
-        return procOut.exitCode == 0
-    }
+        hasExecutable(CARGO)
 
-    fun queryCargoVersion(): Version? {
+    fun queryVersions(): VersionInfo {
         check(!ApplicationManager.getApplication().isDispatchThread)
-        if (!looksLikeValidToolchain()) return null
+        if (!looksLikeValidToolchain()) return VersionInfo(null, null, null, false)
```
mkaput: What do you think about introducing singleton `VersionInfo.INVALID` instance?

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -181,14 +158,14 @@ private fun parseRustcVersion(lines: List<String>): Version? {
 
     val major = releaseMatch.groups[1]!!.value.toInt()
```
zjhmale: i wondered if we can have a safer way to get these numbers from regex match results instead use `!!` directly.

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -14,57 +14,55 @@ import java.io.File
...
-        return runExecutableAndProcessStdout(cmd) { parseRustcVersion(it) }
+        val rustup = if (hasExecutable(RUSTUP)) {
+            GeneralCommandLine(pathToExecutable(RUSTUP))
+                .withParameters("--version")
+                .runExecutable()?.let { parseRustupVersion(it) }
+        } else {
+            null
+        }
+
+        return VersionInfo(
```
zjhmale: this abstraction is pretty neat 😄

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -73,21 +71,15 @@ data class RustToolchain(val location: String) {
...
-    fun nonProjectCargo(): Cargo =
-        Cargo(pathToExecutable(CARGO), pathToExecutable(RUSTC), null)
-
-    val presentableLocation: String = PathUtil.toPresentableUrl(pathToExecutable(CARGO))
-
-    private fun pathToExecutable(toolName: String): String {
-        val exeName = if (SystemInfo.isWindows) "$toolName.exe" else toolName
-        return File(File(location), exeName).absolutePath
-    }
+    data class VersionInfo(
```
zjhmale: hope we can use [typealias](https://github.com/Kotlin/KEEP/issues/4) as quickly as possible so that it is only to create something like

```
typealias LabelVersion = (JLabel, Version?)

data class VersionInfo(
  val cargo: LabelVersion,
 ...
)
```

so that we can encode labels with relative version info directory with explicitly create a map to iterate over it.

mkaput: hmm, creating UI control in logic and passing it all around doesn't seem like a good idea imho

zjhmale: alright

src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
matklad: It would be nice, but I hope that almost everybody uses a cargo with `metadata` command this days, so I don't think that it's worth to do.

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
matklad: Thanks! I've completely forgotten that I wanted to use it the next time I refactor this!

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
matklad: That it is a nice idea!

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
matklad: Yes, I also wonder if it is possible to provide a "compile time safe" accessors for regexpes. But I can't think of a language where this is possible (D perhaps?), because it would need a fair amount of compile time work.

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -108,59 +101,22 @@ data class RustToolchain(val location: String) {
...
-    return if (isStable) {
-        Version(major, minor, build)
-    } else {
-        val commitHash = match.groups[5]!!.value
-        UnstableVersion(commitHash, major, minor, build)
-    }
+private fun findSemVer(lines: List<String>): SemVer {
+    val re = """\d+\.\d+\.\d+""".toRegex()
+    val match = lines.mapNotNull { re.find(it) }.firstOrNull()
+    return SemVer.parseFromTextNonNullize(match?.value)
```
zjhmale: pretty neat! nice job @matklad @mkaput

src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
matklad: Urgh, apparently old IDEA does not have `SemVer.UNKNOWN`, so it's simpler to resort to `null`s for the time being.

zjhmale:
@matklad is it ok to merge this PR even the CI servers were unhappy with it?

matklad:
Yep, if you know what you are doing :) 

I've tested locally and checked that CI for IDEA 15 passes :) Looks like app veyour failed because I had removed the branch :) 

This functionality is not really covered by unit tests anyway :( 

I hope that CI for master passes though :stuck_out_tongue_winking_eye:

zjhmale:
@matklad cool :smile:

PR #638, opened 2016-08-29
remove unnecessary git ignore

zjhmale:
lib file is definitely for lexing and parsing processes, no need to ignore changes about them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,5 @@
 build/
 out
 classes
-lib
 gen
 src/test/resources/rustc-src.zip


matklad:
Thanks!

I really hope that one day we will be able to git rid of this vendoring of binary blobs though!

zjhmale:
agree, and btw why can't we manage these dependencies using gradle at the moment, is that because grammar-kit is currently not upload to maven central repository?
>I really hope that one day we will be able to git rid of this vendoring of binary blobs though!

mkaput:
There were some issues with downloading them in Russia as far as I remember.

zjhmale:
alright
>There were some issues with downloading them in Russia as far as I remember.

matklad:
Yes, I think we can use `gradle` to simply download those jars to the `lib` directory as a part of the build. @alexeykudinkin has some objections https://github.com/intellij-rust/intellij-rust/commit/c6d5ed0c5bcff97ef533b68aa1bc8dd067fb1830, but I don't fully understand them.

Dependencies will be downloaded only once after the initial `.clone`. For all subsequent builds (both clean and incremental), the downloaded jars from the lib dir will be used. 

Because the download happens after the initial clone, you must have access to the GitHub anyway. And I personally haven't had any issues with GitHub availability in Staint Petersbrg (only when troubles are when GH is globally down).

zjhmale:
cool, seems we used to manage these jars using cloud service. anyway it is fine for now to keep tracking these external jars.

PR #650, opened 2016-09-02
(MISC) Add function signature to quick doc

### LINKED ISSUE #641: Show function signature in Ctrl-Q Quick Doc
When `Ctrl-Q` (Quick Doc) onto a function, it would be neat if it showed the function signature (parameters, parameter types, return type).

Here is an example of how it is done in PyCharm:
pycharm-function-signature

allengeorge:
Is something like this reasonable? I'll (obviously) fix up the tests, but I'd like to know if this is along the right track.

Generated documentation:

```
Foo
fn <b>foo</b>()
Foo's documentation
```

- Not sure what signature to return for `RustPatBindingElement` (what is it, anyways?)
- Not sure if the documentation formatting is ideal

Fixes #641


matklad:
Yep, looks reasonable! 

>Not sure what signature to return for RustPatBindingElement (what is it, anyways?)

PatBinding is a name bound in the pattern. For example for 

```
fn foo((x, y): (i32, String)) { }
```

`x` and `y` are PatBindings. I think it's OK to leve signature for them empty for now. In the future we should derive their types from the pattern and show it. 

>Not sure if the documentation formatting is ideal

It don't have to be ideal, some formatting is better than nothing at all :)

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -16,8 +16,14 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
 
         val name = if (element is RustMod) element.modName else element.name
         val header = if (name != null) "<pre>$name</pre>\n" else ""
+        val signature = getSignature(element)
```
matklad: I think this can be condensed to something like `(element as? RustFnElement)?.formatSignature() ?: ""`

mrwiggles: Oh. Very nice. Didn't think of that. Modifying.

mrwiggles:
@matklad I've updated the PR with your suggestion and fixed up all but one test. `RustQuickDocumentationTest.testTraitMethod` fails because it appears that the trailing `;` is being picked up as part of the function signature in `RustDocumentationProvider.formatSignature()`.

Seems like I have to fix that function as well?

matklad:
>Seems like I have to fix that function as well?

Yes, this line is wrong: https://github.com/allengeorge/intellij-rust/blob/4233244cac438c0fbb5ae5c8a35f30e63f5cc9a0/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt#L65

Not sure what is the best fix though. One option is to select the last offset among `)`, return type and where clause. Another option is to switch on the particular type of the function and select between `;` and `{`.

mrwiggles:
Wouldn't option 1 be better? Or, are there consequences that I'm not seeing. I'm happy to fix it - just want to know what the right approach is (since I'm very, very new to this code).

matklad:
Yes, I think option 1 would be slightly better, but neither of them is bullet proof, because PSI is very dynamic and changes rather frequently at the moment. So you'll need to write a fair amount of tests anyway to make sure that it works right, and with good test coverage the detail of implementation are not that important because you can easily change them.

allengeorge:
OK. Here's what I'll do: I'll create a separate PR with tests for the signature change (it might take me a little while to do), and once you're OK with that - we can revisit this PR. Does that make sense?

Thanks for guiding me along @matklad!

matklad:
Yes, this is a good plan!

allengeorge:
@matklad This should work now (tested locally)

matklad:
Thanks @allengeorge !

PR #663, opened 2016-09-10
(MISC) Improve quick doc signature generation

allengeorge:
This is the preliminary work required for #650. Since I'm a newbie at Kotlin and IDEA, I'd really appreciate any input on style and better ways to structure the checks I'm doing.

### DIFF
--- a/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
@@ -65,8 +65,7 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
         // if all else fails, drop down to the length of the current element
         val functionElements = listOf(whereClause, retType, parameters)
         val signatureEnd = functionElements
-                .filterNotNull()
-                .firstOrNull()
+                .filterNotNull().firstOrNull()
                 ?.let { it.startOffsetInParent + it.textLength }
                 ?: textLength
 
--- a/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
@@ -61,18 +61,14 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
 
         val signatureStart = signatureStartElement?.startOffsetInParent ?: 0
 
-        // pick last of: where clause, return type, or closing brace of the parameters
+        // pick (in order) where clause, return type, or closing brace of the parameters
         // if all else fails, drop down to the length of the current element
-        val signatureEnd = if (block != null)
-            block!!.startOffsetInParent
-        else if (whereClause != null)
-            whereClause!!.let { it.startOffsetInParent + it.textLength }
-        else if (retType != null)
-            retType!!.let { it.startOffsetInParent + it.textLength }
-        else if (parameters != null)
-            parameters!!.let { it.startOffsetInParent + it.textLength }
-        else
-            textLength
+        val functionElements = listOf(whereClause, retType, parameters)
+        val signatureEnd = functionElements
+                .filterNotNull()
+                .firstOrNull()
+                ?.let { it.startOffsetInParent + it.textLength }
+                ?: textLength
 
         val identStart = identifier.startOffsetInParent
         val identEnd = identStart + identifier.textLength
@@ -81,13 +77,11 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
 
         val beforeIdent = text.substring(signatureStart, identStart).escaped
         val afterIdent = text.substring(identEnd, signatureEnd)
-                .replace(Regex("""\n"""), "")    // collapse signature into a single line
-                .replace(Regex("""\s\s+"""), "") // collapse consecutive whitespace runs into a single whitespace
-                .replace(Regex("""\(\s"""), "(") // remove whitespace after a parens
-                .replace(Regex("""\s\)"""), ")") // remove whitespace before a parens
-                .replace(Regex(""",\s"""), ",")  // normalize all commas to ?,?
-                .replace(Regex(""","""), ", ")   // now that commas don't have following spaces, add one
-                .trimEnd()                       // finally, trim any ending whitespace
+                .replace("""\s+""".toRegex(), " ")
+                .replace("""( """, "(")
+                .replace(""" )""", ")")
+                .replace(Regex(""" ,"""), ",")
+                .trimEnd()
                 .escaped
 
         return "$beforeIdent<b>$name</b>$afterIdent"
--- a/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
@@ -60,7 +60,19 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
         }
 
         val signatureStart = signatureStartElement?.startOffsetInParent ?: 0
-        val signatureEnd = block?.startOffsetInParent ?: textLength
+
+        // pick last of: where clause, return type, or closing brace of the parameters
+        // if all else fails, drop down to the length of the current element
+        val signatureEnd = if (block != null)
+            block!!.startOffsetInParent
+        else if (whereClause != null)
+            whereClause!!.let { it.startOffsetInParent + it.textLength }
+        else if (retType != null)
+            retType!!.let { it.startOffsetInParent + it.textLength }
+        else if (parameters != null)
+            parameters!!.let { it.startOffsetInParent + it.textLength }
+        else
+            textLength
 
         val identStart = identifier.startOffsetInParent
         val identEnd = identStart + identifier.textLength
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -60,15 +60,32 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
         }
 
         val signatureStart = signatureStartElement?.startOffsetInParent ?: 0
-        val signatureEnd = block?.startOffsetInParent ?: textLength
+
+        // pick last of: where clause, return type, or closing brace of the parameters
+        // if all else fails, drop down to the length of the current element
+        val signatureEnd = if (block != null)
+            block!!.startOffsetInParent
```
allengeorge: Not sure what `block` is, and why we want its `startOffsetInParent`.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -60,15 +60,32 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
...
             identEnd <= signatureEnd && signatureEnd <= textLength)
 
         val beforeIdent = text.substring(signatureStart, identStart).escaped
-        val afterIdent = text.substring(identEnd, signatureEnd).trimEnd().escaped
+
+        // since where clauses and return types may be separated
+        // by significant whitespace, remove consecutive whitespace runs
+        // unfortunately this doesn't handle a single trailing
+        // whitespace, so call trimEnd() as well
+        val afterIdent = text.substring(identEnd, signatureEnd).replace(Regex("""\s\s+"""), " ").trimEnd().escaped
```
allengeorge: I do this so that a `where` clause split over multiple lines can be collapsed into a single line.

allengeorge: Oh. Maybe I should check that parameters and returns spread over multiple lines work too...

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -60,15 +60,35 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
...
+            textLength
 
         val identStart = identifier.startOffsetInParent
         val identEnd = identStart + identifier.textLength
         check(signatureStart <= identStart && identStart <= identEnd &&
             identEnd <= signatureEnd && signatureEnd <= textLength)
 
         val beforeIdent = text.substring(signatureStart, identStart).escaped
-        val afterIdent = text.substring(identEnd, signatureEnd).trimEnd().escaped
+        val afterIdent = text.substring(identEnd, signatureEnd)
```
allengeorge: @matklad This new change handles the case where the params, return types and where clause are all spread out. But, it's pretty ugly.

I think the right thing to do is normalize every signature element and then concatenate them with spaces. Unfortunately I'm not sure how best to do this.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
matklad: `block` is the body of the function inside (and including) `{` and `}`. So the `startOffsetInParent` is the offset `{`. 

We probably don't need this case, as we now explicitly handle all three possible endings of the function signature.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -60,15 +60,35 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
...
+        // if all else fails, drop down to the length of the current element
+        val signatureEnd = if (block != null)
+            block!!.startOffsetInParent
+        else if (whereClause != null)
+            whereClause!!.let { it.startOffsetInParent + it.textLength }
+        else if (retType != null)
+            retType!!.let { it.startOffsetInParent + it.textLength }
+        else if (parameters != null)
+            parameters!!.let { it.startOffsetInParent + it.textLength }
+        else
```
matklad: A chain of `else if` can be written more succinctly using a `when`:

```
when {
    cond1 -> ...
    cond2 -> ...
}
```

But in this case, I think we can do something even neater to avoid code duplication:

```
listOf(whereClause, retType, parameters).firstNotNull()?.let {it.startOffsetInParent + it.textLenght} ?: textLength
```

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -60,15 +60,35 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
...
         val beforeIdent = text.substring(signatureStart, identStart).escaped
-        val afterIdent = text.substring(identEnd, signatureEnd).trimEnd().escaped
+        val afterIdent = text.substring(identEnd, signatureEnd)
+                .replace(Regex("""\n"""), "")    // collapse signature into a single line
+                .replace(Regex("""\s\s+"""), "") // collapse consecutive whitespace runs into a single whitespace
+                .replace(Regex("""\(\s"""), "(") // remove whitespace after a parens
+                .replace(Regex("""\s\)"""), ")") // remove whitespace before a parens
+                .replace(Regex(""",\s"""), ",")  // normalize all commas to ?,?
+                .replace(Regex(""","""), ", ")   // now that commas don't have following spaces, add one
+                .trimEnd()                       // finally, trim any ending whitespace
```
matklad: Would just `replace("""\s+""".toRegex, " ")` work?

matklad: Ah, I see, it will handle only the fist two cases. Still, I think after we normalized all ws to a single space, we can get by without regexpes: `.replace("( ", "(").replace(" ,", ",")`

mkaput: not completely I think, `fn foo(a: i32   , b:i32)` would be transformed to `fn foo(a: i32 , b:i32)`

matklad: >would be transformed to

It would if we `\s+ -> \s` before

>Running formatter on this would be ideal

Yep, though we would need to collapse newlines anyway.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
allengeorge: Ah. Gotcha. I'll remove this branch.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
allengeorge: Oh wow. That's *really* nice. Thanks!

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
allengeorge: So, what should I do? I'm all for simplifying this drastically. I really don't like my hand-coded solution above.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
allengeorge: Unfortunately `firstNotNull` doesn't exist... Is there some import I have to have?

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
matklad: `.replace("""\s+""".toRegex, " ").replace("( ", "(").replace(" ,", ",")` should fix most of the things and is marginally better.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
matklad: Indeed, it should be `.filterNotNull().firstOrNull()` then.

allengeorge: Yup - figured. Change already made and updated PR submitted.

src/main/kotlin/org/rust/ide/docs/RustDocumentationProvider.kt
```diff
@@ -60,15 +60,29 @@ class RustDocumentationProvider : AbstractDocumentationProvider() {
         }
 
         val signatureStart = signatureStartElement?.startOffsetInParent ?: 0
-        val signatureEnd = block?.startOffsetInParent ?: textLength
+
+        // pick (in order) where clause, return type, or closing brace of the parameters
+        // if all else fails, drop down to the length of the current element
+        val functionElements = listOf(whereClause, retType, parameters)
+        val signatureEnd = functionElements
+                .filterNotNull()
```
allengeorge: Maybe I should collapse these two into one line to make the intent clear (i.e. pick first non-null element in list)

matklad: Yes, I think that would be a bit clearer, though it is not that important.

matklad:
Great!

PR #665, opened 2016-09-11
Fixes for inference of primitive types and literals

### LINKED ISSUE #662: Treat 'char' and 'str' as primitives
Currently [`char`](https://doc.rust-lang.org/std/primitive.char.html) and [`str`](https://doc.rust-lang.org/std/primitive.str.html) are not treated as [primitives](https://doc.rust-lang.org/std/#primitives), and this seems to cause incorrect highlighting in the editor as well as some false positive unresolved references.

char_str_primitives

kumbayo:
This fixes #662

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/RustCharacterType.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/RustCharacterType.kt
@@ -5,6 +5,9 @@ import org.rust.lang.core.types.visitors.RustUnresolvedTypeVisitor
 
 object RustCharacterType : RustPrimitiveTypeBase() {
 
+    fun deduce(text: String?): RustCharacterType? =
+        text?.let { if (it == "char") RustCharacterType else null }
+
     override fun <T> accept(visitor: RustTypeVisitor<T>): T = visitor.visitChar(this)
 
     override fun <T> accept(visitor: RustUnresolvedTypeVisitor<T>): T = visitor.visitChar(this)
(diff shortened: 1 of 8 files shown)


alexeykudinkin:
Awesome work @kumbayo! Instead of commenting out let's make it ignored: just rename it following way -- replace `testXXX` prefix with `ignoreXXX` prefix (yes, unfortunately, we're bound to JUnit 3).

kumbayo:
Hmm it seems this will completely ignore the test then :-/
Is it ok if i do something like:
```kotlin

 fun testStrRef() = expectComparisonFailure {
     testExpr("""
     fn main() {
         let a = "Hello";
                    //^ & str
     }
 """)
 }

protected fun expectComparisonFailure(f: () -> Unit ) {
    try {
        f()
    } catch (e: ComparisonFailure) {
        return
    }
    fail("No ComparisonFailure was thrown during the test")
}
```

This way we remember about it being broken and can know when the behavior changes.

alexeykudinkin:
Reasonable. But i'd rather go like this:

```
protected inline fun <reified X: Exception> ignore(f: () -> Unit) {
    try {
        f()
    } catch (e: X) {
        return
    }
    fail("No ${X::class.java} was thrown during the test")
}

 fun testStrRef() = ignore<ComparisonFailure> {
     testExpr("""
     fn main() {
         let a = "Hello";
                    //^ & str
     }
 """)
 }

```

kumbayo:
This looks nicer and more versatile but unfortunately i get
`"Reified type is forbidden for catch parameter"`
at the catch clause and i don't know enough Kotlin to fix it :-/

alexeykudinkin:
Argh! Kotlin make me wonder sometimes, how did they come up with this.

Nice workaround though! LGTM

PR #669, opened 2016-09-14
(PSI): refactor path handling

matklad:
`RustPath` is now fully decoupled from PSI: it does not concern itself
with grammar idiosyncrasies like `self` being an identifier only
sometimes or implicitly absolute paths in view items.

`RustPath` is always valid by construction, so converting invalid PSI
path (like `::foo::super::self`) into `RustPath` will fail.

`RustQualifedReferenceElement` is removed because `RustPath` now is the
interface for the resolve.

@alexeykudinkin @mkaput please review.


src/main/kotlin/org/rust/utils/io.kt
```diff
@@ -0,0 +1,29 @@
+package org.rust.utils
+
+import com.intellij.util.io.DataExternalizer
+import java.io.DataInput
+import java.io.DataOutput
+
+interface RustDataExternalizer<T> : DataExternalizer<T> {
+    override fun save(output: DataOutput, value: T)
+    override fun read(input: DataInput): T
```
mkaput: I don't get what these overrides do. Couldn't we just provide `nullable` as extension method for `DataExternalizer` itself?

src/main/kotlin/org/rust/lang/core/symbols/RustPath.kt
```diff
@@ -0,0 +1,103 @@
...
+        override fun equals(other: Any?): Boolean = other is Named && segment == other.segment
+
+        override fun hashCode(): Int = segment.hashCode()
+    }
+
+    companion object : RustDataExternalizer<RustPathHead> {
+        override fun save(output: DataOutput, value: RustPathHead) {
+            when (value) {
+                is RustPathHead.Absolute -> output.writeInt(-1)
+                is RustPathHead.Relative -> output.writeInt(value.level)
```
mkaput: Maybe I'm nitpicking, but do we really need 4 bytes of possible levels? From what I understand this may have some impact on cache size in bigger project so it might be worth to take into account even if it's microoptimization.

src/main/kotlin/org/rust/utils/io.kt
matklad: They enforce proper not null types. Without it, you could do

```
class Bug: DataExternalizer<String> {
    override fun read(input: DataInput): String? = null
}
```

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustPathImplMixin.kt
```diff
@@ -2,67 +2,84 @@ package org.rust.lang.core.psi.impl.mixin
...
 
-    override val fullyQualified: Boolean
-        get() = qualifier?.fullyQualified ?: separator != null || isViewPath && self == null && `super` == null
+    return when {
+        coloncolon != null -> {
+            if (self != null || `super` != null) {
+                return null // Forbid `::super` and `::self`.
+            }
+            RustPath(RustPathHead.Absolute, listOf(RustPathSegment(referenceName)))
+        }
```
mkaput: `return when if return` looks ugly imho

src/main/kotlin/org/rust/lang/core/psi/RustItemElement.kt
```diff
@@ -1,26 +1,15 @@
...
-        else
-            RustQualifiedPath.create(
-                RustQualifiedPathPart.from(name),
-                qualifier = containingMod?.canonicalCratePath,
-                fullyQualified = true
-            )
+val RustItemElement.canonicalCratePath: RustPath? get() {
+    if (this is RustMod) {
+        @Suppress("USELESS_CAST")
+        return (this as RustMod).canonicalCratePath
```
mkaput: Kotlin bug?

mkaput:
:+1:

src/main/kotlin/org/rust/lang/core/psi/RustItemElement.kt
matklad: While technically it is not a useless cast, I kind'a like this warning, because defining two extension methods for non disjoint receiver types is scary. 

I think we need some better abstraction here: `ForeignMod` is an item, but it can't be identified with a path, while `EnumVariant` is not an item, but it has a reasonable `path`. And there are also things like

```
fn foo() {
    { struct S; }
    { struct S; }
}
```

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustPathImplMixin.kt
matklad: Indeed it does!

src/main/kotlin/org/rust/lang/core/symbols/RustPath.kt
matklad: > this only refers to relative paths

Not really, `::foo::bar::baz` will also write a length here. I don't think this would be a problem, but it's easy to shave couple of bytes off, so why not? 

A bigger problem is that we use `writeUTF` instead of `writeName` when the underlying stream is `StubStream`. `writeName` is more efficient, it's like interning.

matklad: >Not really

Ah, I've confused it with `writeInt(segments.length)`. Let's than leave `int`s for know because of confusability!

PR #177, opened 2015-12-11
Logo

Turbo87:
I've combined the new IntelliJ logo with the official Rust logo to make a logo for this plugin (and the `intellij-rust` GitHub org)

PR #672, opened 2016-09-15
(ART): project icon

mkaput:
My reincarnation of #177, I'll prepare website change later

I though also about `IR` abbreviation, what do you think, which is better?

@matklad please update organization icon if this merges :)


matklad:
Yay! :joy_cat: 


>I though also about IR abbreviation, what do you think, which is better?

It took me more than a moment to decipher IR, and it looks suspiciously like the IR from MIR, so let's stick with RS!

PR #682, opened 2016-09-22
Intentions. UnwrapSingleExprIntention and WrapLambdaExprIntention

ujpv:


ujpv:
Sorry, I changed the comments. 
I hope it became better.

ujpv:
:) Open

ujpv:
Sorry, I changed the comments.
I hope it became better.

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
```diff
@@ -32,9 +35,11 @@ class WrapLambdaExprIntention : PsiElementBaseIntentionAction() {
     override fun invoke(project: Project, editor: Editor, element: PsiElement) {
         val ctx = findContext(element) ?: return
         val lambdaBody = ctx.lambdaExpr.expr ?: return
-        val bodyStr = ctx.lambdaExpr.expr?.text ?: return
+        var bodyStr = ctx.lambdaExpr.expr?.text ?: return
+        bodyStr = "\n$bodyStr\n"
```
mkaput: Why do you put newlines here? There's no need for them, formatter should do the work when needed.

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
```diff
@@ -32,9 +35,11 @@ class WrapLambdaExprIntention : PsiElementBaseIntentionAction() {
...
         val blockExpr = RustElementFactory.createBlockExpr(project, bodyStr) ?: return
-        lambdaBody.replace(blockExpr)
+        val offset = ((lambdaBody.replace(blockExpr)) as RustBlockExprElement).block?.expr?.textOffset ?: return
+        editor.caretModel.moveToOffset(offset)
```
mkaput: I don't know whether we should move cursor here. @matklad what's your opinion?

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
```diff
@@ -1,60 +0,0 @@
...
-        // if we get lambda, than return false as we are inside "unblocked" lambda
-        val blockExpr = PsiTreeUtil.findFirstParent(element, blockOrLambda) as? RustBlockExprElement ?: return null
-        val block = blockExpr.block ?: return null
-
-        val lambdaExpr = element.parentOfType<RustLambdaExprElement>() ?: return null
-
-        // check whether our block's parent is lambda to avoid false positives
-        // like nested blocks in lambda (we are not interested in this case as we can
-        // be sure that cursor isn't at brace; well, I don't know what about `|x| {<caret>{ x }}`,
-        // but it's possibly rare edge case), or just some random block in code
```
mkaput: These comments were fine :sob:

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
matklad: I think this is fine: formatter wont change `|x| { x }` into 

```
|x | {
    s
}
```

because it is Ok with both variants. But when you want to add braces to lambda, you almost always want the second formatting.

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
matklad: I think that we definitely should not live the cursor just after `{` and before the newline. placing it just before the expression seems fine, or are there better alternatives?

mkaput: Leave it where it was before applying inspection.

matklad: Yes, that would be less confusing perhaps. @ujpv let's then just leave the caret where it was!

src/test/kotlin/org/rust/ide/intentions/UnwrapSingleExprIntentionTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        """
+        ,
+        """
+        fn main() {
+            |x| x * x
+        }
+        """
+    )
+
+    fun testAvailableUnwrapBracesSingleExpressionIf() = doTest(
```
matklad: I think that for cases where the intention should not be available it should be possible to specify only a single input string.

src/main/kotlin/org/rust/ide/intentions/UnwrapLambdaExprIntention.kt
matklad: Now we do handle nested block, so the comment no longer describes the state of affairs.

src/main/kotlin/org/rust/ide/intentions/WrapLambdaExprIntention.kt
ujpv: Done.
Also I have given the same behavior to UnwrapSingleExprIntention.

ujpv:
I have corrected everything.

matklad:
Thanks!

PR #683, opened 2016-09-22
(EDT): Implement raw literal hashes balancing on typing

### LINKED ISSUE #673: Automagically handle raw str literal hashes when typing
Citing @matklad:
>It'd be cool to automatically add or remove hashes to both sides of raw string literal if you edit either one (the same way as closing and opening xml tags are edited simultaneously).

Rolling out custom typing handler delegate is the way to go here.

mkaput:
(the end of the gif seems not to be a bug, I cannot find how to handle <kbd>Delete</kbd>)

fix #673 

BTW: I love our test suite, I have implemented this without opening IntelliJ even once, and still know that this works!


src/test/kotlin/org/rust/ide/typing/RustRawLiteralHashesDeleterTest.kt
```diff
@@ -0,0 +1,27 @@
...
+    override val dataPath = "org/rust/ide/typing/rawHashes/fixtures"
+
+    // Swap file names, so we can reuse tests for inserter :-D
+    override fun beforeFileName(fileName: String): String {
+        return super.afterFileName(fileName)
+    }
+
+    override fun afterFileName(fileName: String): String {
+        return super.beforeFileName(fileName)
+    }
```
mkaput: Rather than doing naming shenanigans, I'm thinking on introducing `isInverse: Boolean` field

src/test/resources/org/rust/ide/typing/rawHashes/fixtures/at_total_end_after.rs
```diff
@@ -0,0 +1,3 @@
+// Caret and EOL is important here!
```
mkaput: typo :sob:

src/test/kotlin/org/rust/ide/typing/RustRawLiteralHashesDeleterTest.kt
matklad: What do you think about switching to "inline rust code" style test here? I find it rather annoying to work with two 3-line fixture files  :)

src/main/kotlin/org/rust/ide/typing/RustRawLiteralHashesInserter.kt
```diff
@@ -0,0 +1,46 @@
...
+        // We are only interested in valid raw literals
+        val literal = getLiteralDumb(iterator) as? RustRawStringLiteralImpl ?: return Result.CONTINUE
+        val openDelim = literal.offsets.openDelim ?: return Result.CONTINUE
+        val closeDelim = literal.offsets.closeDelim ?: return Result.CONTINUE
+
+        // Get rid of ugly quotes from out precious ranges (remember - we are operating on left-closed ranges)
+        val openHashes = openDelim
+        val closeHashes = closeDelim.shiftRight(1)
+
+        // No detect on which side of the literal we are, and insert hash on the other one
```
matklad: No -> Now

src/test/kotlin/org/rust/ide/typing/RustRawLiteralHashesDeleterTest.kt
mkaput: I don't how could reverse testing be implemented in that style.

src/main/kotlin/org/rust/ide/typing/RustRawLiteralHashesInserter.kt
```diff
@@ -0,0 +1,46 @@
...
+        // handle this case: r#""#<caret>, cases r<caret>#""# and <caret>r#""# will still work.
+        val highlighter = (editor as EditorEx).highlighter
+        val iterator = highlighter.createIterator(caretOffset - 1)
+
+        // We are only interested in valid raw literals
+        val literal = getLiteralDumb(iterator) as? RustRawStringLiteralImpl ?: return Result.CONTINUE
+        val openDelim = literal.offsets.openDelim ?: return Result.CONTINUE
+        val closeDelim = literal.offsets.closeDelim ?: return Result.CONTINUE
+
+        // Get rid of ugly quotes from out precious ranges (remember - we are operating on left-closed ranges)
```
mkaput: out -> out

src/test/kotlin/org/rust/ide/typing/RustRawLiteralHashesDeleterTest.kt
matklad: Like this:

```
fun doTest1(before: @Languge("Rust), String, after: @Languge("Rust), String) = inner(before, after)
fun doTest2(before: @Languge("Rust), String, after: @Languge("Rust), String) = inner(after, before)
```

(Github, why have you make "Start a review" a default button :sob: )

matklad: Hm, and what is the problem? Usual tests boil down to [checkResultByFile](https://github.com/intellij-rust/intellij-rust/blob/4d4cc2dd0bf524cdc27f06fd7a7b53989023b14f/src/test/kotlin/org/rust/lang/RustTestCaseBase.kt#L44) which does uses `fileName` to configure fixtures, while the the inline style tests use `InlineFile`(https://github.com/intellij-rust/intellij-rust/blob/4d4cc2dd0bf524cdc27f06fd7a7b53989023b14f/src/test/kotlin/org/rust/lang/RustTestCaseBase.kt#L88) and configure files by text.  `//^` is used instead of `<caret>`, maybe `/*|*/` should be added as an alternative.

mkaput: I decided to stick with `val inverse: Boolean` solution, as it looks much clearer for me. InlineFile is not fine imho as `/*|*/` solution still interrupts lexer.

mkaput:
ready

src/main/kotlin/org/rust/ide/typing/RustRawLiteralHashesBalancer.kt
```diff
@@ -0,0 +1,89 @@
...
+        val caretOffset = editor.caretModel.offset + 1
+        val (openHashes, closeHashes) = checkNotNull(offsets)
+
+        // Now detect on which side of the literal we are, and remove hash on the other one.
+        // Remember that offsets apply to literal before deletion!
+        when (caretOffset) {
+            in openHashes -> {
+                // -1 because left-closed ranges
+                // -1 because open hash was deleted some range is shifted do right by 1
+                // TODO: Why one more -1?
```
mkaput: any ideas?

mkaput:
Is anything to fix here I forgot about?

matklad:
Nope, I want to investigate https://github.com/intellij-rust/intellij-rust/pull/683/files#diff-ad1fbcaefac11819c695f1e9ece28781R70 :) Also I want to cut alpha today, and to merge this after so as to let it bake in in nighties for a week. Indexing strings always make me feel uneasy :) 

Sorry for the long review, was a bit busy during the weekend!

mkaput:
No problem 😄

src/main/kotlin/org/rust/ide/typing/RustRawLiteralHashesBalancer.kt
matklad: Looks like hash ranges are one greater than they should be. for

```
fn main() {
    r###"aaa"###
}
```

I got

```
openHashes = (17,21)
closeHashes = (25,29)
```

That is, the ranges are of length 4, while looks like thy should be of the length 3. Is this a bug? If it is, then perhaps a test is missing?

mkaput: Yeah, that's might be the cause and it's a hidden bug which I'm blindly workarounding here. I want to remove one offset from them (`"`), but for example for openhashes I don't do this as I see

mkaput:
fixed

src/main/kotlin/org/rust/ide/typing/RustEnableableBackspaceHandlerDelegate.kt
```diff
@@ -0,0 +1,31 @@
...
+     */
+    open fun deleting(c: Char, file: PsiFile, editor: Editor): Boolean = true
+
+    /**
+     * Perform action after char [c] was deleted.
+     *
+     * @return true whether this handler succeeded and the IDE should stop evaluating
+     *         remaining handlers; otherwise, false
+     */
+    open fun deleted(c: Char, file: PsiFile, editor: Editor): Boolean = false
```
matklad: In general, I prefer `open protected` and `public final` combinations, sort of like C++ NVI idiom.

matklad:
Fabulous!

PR #690, opened 2016-09-26
Surraunders test by text

ujpv:


matklad:
It does not make much sense to add a base test class without actually using it. It's better either to convert all tests to the new style, or to add this base together with the new tests for new surrenders.

PR #692, opened 2016-09-27
(TYPE): support for type parameters substitutions

matklad:
@alexeykudinkin @mkaput please review. The main interface changes are here: https://github.com/intellij-rust/intellij-rust/compare/type-params?expand=1#diff-231e8a807fef2b7083c5227a54b3aed0R23

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/RustEnumType.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/RustEnumType.kt
@@ -4,12 +4,21 @@ import com.intellij.codeInsight.completion.CompletionUtil
 import org.rust.lang.core.psi.RustEnumItemElement
 import org.rust.lang.core.types.visitors.RustTypeVisitor
 
-class RustEnumType(enum: RustEnumItemElement) : RustStructOrEnumTypeBase() {
+class RustEnumType(
+    enum: RustEnumItemElement,
+    typeArguments: List<RustType> = emptyList()
+) : RustStructOrEnumTypeBase(typeArguments) {
 
     override val item = CompletionUtil.getOriginalOrSelf(enum)
 
     override fun <T> accept(visitor: RustTypeVisitor<T>): T = visitor.visitEnum(this)
 
     override fun toString(): String = item.name ?: "<anonymous>"
+
+    override fun withTypeArguments(typeArguments: List<RustType>): RustEnumType =
+        RustEnumType(item, typeArguments)
+
+    override fun substitute(map: Map<RustTypeParameterType, RustType>): RustEnumType =
+        RustEnumType(item, typeArguments.map { it.substitute(map) })
 
 }
--- a/src/main/kotlin/org/rust/lang/core/types/RustTypeParameterType.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/RustTypeParameterType.kt
@@ -16,6 +16,8 @@ class RustTypeParameterType(val parameter: RustTypeParamElement) : RustTypeBase(
     override fun getNonStaticMethodsIn(project: Project): Sequence<RustFnElement> =
         getTraitsImplementedIn(project).flatMap { it.traitMethodMemberList.asSequence() }
 
+    override fun substitute(map: Map<RustTypeParameterType, RustType>): RustType = map[this] ?: this
+
     override fun <T> accept(visitor: RustTypeVisitor<T>): T = visitor.visitTypeParameter(this)
 
     override fun toString(): String = parameter.name ?: "<unknown>"
--- a/src/main/kotlin/org/rust/lang/core/types/RustFunctionType.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/RustFunctionType.kt
@@ -10,4 +10,7 @@ class RustFunctionType(val paramTypes: List<RustType>, val retType: RustType) : 
         val params = paramTypes.joinToString(", ", "fn(", ")")
         return if (retType === RustUnitType) params else "$params -> $retType"
     }
+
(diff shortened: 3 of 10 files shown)


src/main/kotlin/org/rust/lang/core/types/RustTypeParameterType.kt
```diff
@@ -16,6 +16,8 @@ class RustTypeParameterType(val parameter: RustTypeParamElement) : RustTypeBase(
     override fun getNonStaticMethodsIn(project: Project): Sequence<RustFnElement> =
         getTraitsImplementedIn(project).flatMap { it.traitMethodMemberList.asSequence() }
 
+    override fun substitute(map: Map<RustTypeParameterType, RustType>): RustType = map[this] ?: this
+
```
mkaput: If we are mapping over `RustTypeParameterType` then we should provide specialized `hashCode`,

or I don't know something about Kotlin/Java in this area

matklad: There are specialized equals/hashCode over there: https://github.com/intellij-rust/intellij-rust/blob/bd2d1019783eb7fc546f8d81bdf912534d857bd8/src/main/kotlin/org/rust/lang/core/types/RustTypeBase.kt#L19-L21

src/main/kotlin/org/rust/lang/core/types/RustEnumType.kt
```diff
@@ -4,12 +4,21 @@ import com.intellij.codeInsight.completion.CompletionUtil
 import org.rust.lang.core.psi.RustEnumItemElement
 import org.rust.lang.core.types.visitors.RustTypeVisitor
 
-class RustEnumType(enum: RustEnumItemElement) : RustStructOrEnumTypeBase() {
+class RustEnumType(
+    enum: RustEnumItemElement,
+    typeArguments: List<RustType> = emptyList()
+) : RustStructOrEnumTypeBase(typeArguments) {
```
alexeykudinkin: Don't spoil the interface of the type with the `typeArguments` -- make a decorator type `TypeWithSubstitution`.

matklad: I'll try the decorator solution, but I worry that it may break (for good or for bad) matching over types like

```
when (fooType) {
    is RustStructType -> ....
}
```

matklad: Yep, the decorator approach does not work here at all because sometime we want to know the precise type of the thing ([example](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L122)).

Also, it would make type visitors massively more complicated, because you'll need to deal with possibility of `TypeWithSubstitution(TypeWithSubstitution(TypeWithSubstitution(BaseType)))`.

alexeykudinkin: The decorator may be extended to set of _aggregators_: having particular aggregators like `SubstitutedStructType`, etc.

alexeykudinkin: Types that are not generics should not know anything about the type-arguments.

src/main/kotlin/org/rust/lang/core/types/RustTypeParameterType.kt
alexeykudinkin: We should not map over `RustTypeParameterType`, b/c we're misusing the abstraction -- we're mapping over the _name_ (and implicit position) of the parameter, not against its type.

Let's replaced `RustTypeParameterType` here with abstract `RustTypeParameterTag` of which every `RustTypeParameter` would've got a _unique_ one.

alexeykudinkin:
Awesome and long-awaited change!

src/main/kotlin/org/rust/lang/core/types/RustTypeParameterType.kt
matklad: And would be the direct benefits of `RustTypeParameterTag`? I fail to see any at the moment. Perhaps we'd better add it later, when/if we understand why exactly do we need it?

src/main/kotlin/org/rust/lang/core/types/RustEnumType.kt
matklad: Hm, `StructType` is basically `SubstitutedStructType` with an empty substitution. Again, I don't see any concrete benefits from agregators/decorators, but I do see an increase in code size.

src/main/kotlin/org/rust/lang/core/types/RustTypeParameterType.kt
alexeykudinkin: B/c you don't map _type_ into _type_, you map _parameter_ (which has nothing to do with types) into type.

src/main/kotlin/org/rust/lang/core/types/RustEnumType.kt
alexeykudinkin: Concrete benefits stem from the fact that we use statically-typed language and you suggest to fallback to duck-typing mode: `StructType` in general has nothing to do with substitution and should continue to be doing that way instead of effectively being `StructTypeWithSubstitution` with empty substitution list. Let's not accumulate technical debt.

matklad:
Hm, I am pretty sure that we'd better not to add more complexity that we actually need, and I fail to see any real need here. We've already had problems caused by needless complexity before, so I am taking the liberty of merging this pull request as is. 

Of course I may be wrong! In this case, it should be pretty easy to add necessary abstractions latter if they really do reduce the amount of code and fix tests.

PR #693, opened 2016-09-27
(EDT): fix quote completion in raw literals

### LINKED ISSUE #687: Typing quote (") inside a raw string literal should not close it
```
r###"Hello, <caret> World!"###
```

typing `"` should produce

```
r###"Hello, "" World!"###
```

now it gives

```
r###"Hello, ""### World!"###
```

mkaput:
I decided not to complete quotes inside, as it might be cumbersome sometimes, also checked out what does PyCharm do in `"""..."""` literals and it's also not completing quotes inside.
fix #687


matklad:
Thanks! Don't worry about failing CI: my mistake, IDEA 15 is missing some test function yet again :)

PR #695, opened 2016-09-28
(SURROUND) InlineCode Tests

ujpv:


matklad:
Much more readable now, great!

PR #699, opened 2016-09-30
(SURR): While and If Surrounder

ujpv:


src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurroundDescriptor.kt.orig
```diff
@@ -0,0 +1,32 @@
...
+            RustWithBlockSurrounder(),
+<<<<<<< HEAD
+            RustWithLoopSurrounder()
+=======
+            RustWithWhileSurrounder(),
+            RustWithIfSurrounder()
+>>>>>>> 41a8e8f... (SURR): While and If Surrounder
+        )
+    }
+}
```
mkaput: ups

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
```diff
@@ -0,0 +1,36 @@
...
+        assert(ifExpr.block != null)
+        ifExpr.block?.addStatements(statements) ?: return null
+
+        container.deleteChildRange(statements.first(), statements.last())
+
+        ifExpr = CodeInsightUtilBase.forcePsiPostprocessAndRestoreElement(ifExpr)
+
+        val conditionTextRange = ifExpr.expr.textRange
+        val range = TextRange.from(conditionTextRange.startOffset, 0)
+
```
mkaput: move this empty line one step down, so deletion will be contextually matched with used variables

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
```diff
@@ -0,0 +1,36 @@
...
+        editor.document.deleteString(conditionTextRange.startOffset, conditionTextRange.endOffset)
```
mkaput: I don't see placing cursor inside condition. Does this happen automatically?

src/main/kotlin/org/rust/ide/surroundWith/RustWithWhileSurrounder.kt
```diff
@@ -0,0 +1,35 @@
...
+        whileExpr.block?.addStatements(statements) ?: return null
+
+        container.deleteChildRange(statements.first(), statements.last())
+
+        whileExpr = CodeInsightUtilBase.forcePsiPostprocessAndRestoreElement(whileExpr)
+
+        val conditionTextRange = whileExpr.expr.textRange
+        val range = TextRange.from(conditionTextRange.startOffset, 0)
+
+        editor.document.deleteString(conditionTextRange.startOffset, conditionTextRange.endOffset)
```
mkaput: same comments as to `if` surrounder

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
```diff
@@ -0,0 +1,36 @@
...
+        project: Project,
+        editor: Editor,
+        container: PsiElement,
+        statements: Array<out PsiElement>
+    ): TextRange? {
+        var ifExpr = RustElementFactory.createExpression(project, "if true {\n}") as RustIfExprElement
+        ifExpr = container.addBefore(ifExpr, statements[0]) as RustIfExprElement
+
+        assert(ifExpr.block != null)
+        ifExpr.block?.addStatements(statements) ?: return null
```
mkaput: This one-liner is more kotlinish and yields better error message than these 2 lines:
```kotlin
checkNotNull(ifExpr.block).addStatements(statements)
```

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
ujpv: If I correctly understood the question.
In the `range` stores a void TextRange, refers to the first character of temporary condition expr. 
```Rust
if <selection></selection>true {...}
```
Then in the editor, this condition is removed.
```Rust
if <selection></selection> {...}
```
It `range` returns from `fun surroundStatements()`. And this `range` determines the cursor position.

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
ujpv: Or better to just remove the line 23?

src/main/kotlin/org/rust/ide/surroundWith/RustStatementsSurroundDescriptor.kt.orig
ujpv: Really ups.

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
mkaput: This will hide potential implementation bugs

src/main/kotlin/org/rust/ide/surroundWith/RustWithIfSurrounder.kt
mkaput: Ok, got it. Sorry for misunderstanding :)

ujpv:
@mkaput , thanks for the review.
If fixed PR.

About `if (expr)`

If you mean 
```Rust
a == b
```
=>
```Rust
if a == b {
}
```
I think it can do postfix-template.

mkaput:
Both ways should be implemented imho (not everybody is used to postfix templates)

matklad:
Huh, while and if looks suspiciously similar! I don't think we should extract a common something out of them right now, but this possibility is a good thing to keep in mind!

ujpv:
@matklad , I can say more...
All surrounders looks suspiciously similar. :) 
But for extract duplicated code, necessary to make some refactoring existing PSI classes.

matklad:
>But for extract duplicated code, necessary to make some refactoring existing PSI classes.

I don't think that we should add additional methods directly to PSI classes. First of all, it is extremely cumbersome to do because of the code generation. Second, if we add everything to PSI, it will become monstrously huge. 

I think the way to go here is perhaps some local extensions? There is a neat trick that if you define the extension method in the companion object, it will be effectively protected:

```
abstract class BaseSurrounder {
  companion object {
    fun RustExprElement.someExtension() {}
  }
}

class MySurrounder: BaseSurrounder {
    // someExtension is visible here
}
```

mkaput:
Take a look how it's implemented in Kotlin, they have common base class for this stuff.

PR #701, opened 2016-10-01
Project structure icons improvement

alygin:
IntelliJ Rust now uses icons from the IntelliJ built-in resources for methods and functions. There're two problems whith those icons and their usage:

First, for some reason, the `function.png` icon from the built-in IntellJ icon set contains the lambda letter instead of `f`. No other popular language plugins use the lambda symbol for functions (I've checked Java, Kotlin, Scala, C++) because it's obviously associated with lambdas, thus leading to confusion. The lambda-icon is sometimes used by other languages but only in connection with lambdas.

Second, The Book is [quite clear](https://doc.rust-lang.org/book/method-syntax.html) on using the terms *method* and *function* in the language. Formally, there're no *static methods* in Rust, there're *associated functions*.

The PR fixes these two problems:

1. Functions now have an icon with the `f` symbol, the same as in other languages. It's packaged with the plugin because IntelliJ has no such build-in icon and other plugins include it into their own resources too.

2. Associated functions now use the function icon with the static mark instead of the method icon.

Here's the before-after picture (the green arrows show the cases affected by the PR):

ps_before_after


src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplMethodMemberImplMixin.kt
```diff
@@ -25,10 +25,7 @@ abstract class RustImplMethodMemberImplMixin : RustFnImplMixin<RustImplMethodMem
     override fun getParent(): PsiElement? = parentByStub
 
     override fun getIcon(flags: Int): Icon? {
-        var icon = RustIcons.METHOD
-        if (isStatic) {
-            icon = icon.addStaticMark()
-        }
+        val icon = if (isStatic) RustIcons.FUNCTION.addStaticMark() else RustIcons.METHOD
```
mkaput: Maybe extract `RustIcons.FUNCTION.addStaticMark()` as separate `ASSOC_FUNCTION` field?

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustTraitMethodMemberImplMixin.kt
```diff
@@ -20,9 +20,11 @@ abstract class RustTraitMethodMemberImplMixin : RustFnImplMixin<RustTraitMethodM
...
-        var icon = if (isAbstract) RustIcons.ABSTRACT_METHOD else RustIcons.METHOD
+        val icon: Icon
         if (isStatic) {
-            icon = icon.addStaticMark()
+            icon = (if (isAbstract) RustIcons.ABSTRACT_FUNCTION else RustIcons.FUNCTION).addStaticMark()
+        } else {
+            icon = if (isAbstract) RustIcons.ABSTRACT_METHOD else RustIcons.METHOD
         }
         return icon
     }
```
mkaput: Maybe:
```kotlin
override fun getIcon(flags: Int): Icon = when {
        isStatic && isAbstract -> RustIcons.ABSTRACT_FUNCTION.addStaticMark()
        isStatic -> RustIcons.FUNCTION.addStaticMark()
        isAbstract -> RustIcons.ABSTRACT_METHOD
        else -> RustIcons.METHOD
}
```
?

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplMethodMemberImplMixin.kt
alygin: It totally makes sense. I didn't find any predefined icons like that in the existing code, so decided to implement it the same way as other similar classes do. I'll extract the icon.

alygin:
Green quarters are usually used for abstract classes. Abstract functions/methods have the top left quarter cut off. For example, here's how an abstract class with an abstract function is displayed in Java:

screen shot 2016-10-01 at 17 25 59

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustTraitMethodMemberImplMixin.kt
alygin: Agree. It looks more idiomatic. I'll do the refactoring.

matklad:
Thanks!

PR #702, opened 2016-10-01
Display consts and statics in the Project Structure window

alygin:
This fix adds information about constants and static values to the Project Structure window. They are displayed as bindings with the final and static markers respectively:

const-static

### DIFF
--- a/src/main/kotlin/org/rust/ide/icons/RustIcons.kt
+++ b/src/main/kotlin/org/rust/ide/icons/RustIcons.kt
@@ -25,6 +25,7 @@ object RustIcons {
 
     // Marks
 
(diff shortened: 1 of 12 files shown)


matklad:
Nice! Could you perhaps add a test over there: https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/structure/RustStructureViewTest.kt ?

alygin:
Sure, I'll get back to it shortly.

alygin:
Covered by tests and rebased

matklad:
Thanks!

alygin:
Thanks for merging!

PR #703, opened 2016-10-01
(SURR): for Surrounder

ujpv:


matklad:
Thanks!

Let's try to refactor out common code of surrounds though!

PR #705, opened 2016-10-02
(SURR): Refactoring + if expr + while expr

ujpv:
That you think about this?


src/test/kotlin/org/rust/ide/surroundWith/RustWithBlockSurrounderTest.kt
```diff
@@ -1,5 +1,7 @@
 package org.rust.ide.surroundWith
 
+import org.rust.ide.surroundWith.statement.RustWithBlockSurrounder
+
```
mkaput: Maybe also just move test to related packages?

src/main/kotlin/org/rust/ide/surroundWith/expression/RustExpressionSurrounderBase.kt
```diff
@@ -20,7 +24,15 @@ abstract class RustExpressionSurrounderBase : Surrounder {
...
-        return surroundExpression(project, editor, requireNotNull(targetExpr(elements)))
+        val expression = requireNotNull(targetExpr(elements))
+        val templateExpr = RustElementFactory.createExpression(project, getTemplate()) as RustExprElement
+
+        getLeafExpression(templateExpr).replace(expression)
+        val newExpression = expression.replace(templateExpr)
+
+        CodeInsightUtilCore.forcePsiPostprocessAndRestoreElement(newExpression)
+
+        return getSelectionRange(newExpression)
```
mkaput: I have mixed feelings on changing API semantics of `RustExpressionSurrounderBase`. Previous, low level, form might be useful for some complex surrounders in the future.

I'm curious what are @matklad's and @alexeykudinkin's opinions on this

matklad: Yep, I agree. And actually `if`, `while` and friends look more like `RustStatementSurrounder`, because they surround statements.

matklad: Oups, sorry, completely out of base, we do have a separate statements surrounder.

matklad: In this case, I think this refactoring is good. When (and if :) ) we need a more low level API, we can introduce another base class, or just use a plain old `Surrounder`. 

But I think that we can do better, if we make surrounder generic over the the resulting exression

```
class RustExpressionSurrounder<E: RustExprElement> {
      abstract fun createTemplate(): R 
      abstract fun getLeafExpression(expr: R)
}
```

ujpv: Also I can keep the existing RustExpressionSurrounderBase, and derive RustSimpleExprSurrounderBase.

I wanted to use the same strategy for statements surrounder. Is it good way?

matklad: >Also I can keep the existing RustExpressionSurrounderBase, and derive RustSimpleExprSurrounderBase.

Will there be any classes which inherent from `RustExpressionSurrounderBase`, but not from `RustSimpleExprSurrounderBase`? If the answer is no, it's better not to do this right now. 

>I wanted to use the same strategy for statements surrounder. Is it good way?

Yep, it looks reasonable!

ujpv: Ок.
Thanks.

ujpv:
I have moved tests to appropriate packages.
RustExpressionSurrounderBase now is generic-class.

src/main/kotlin/org/rust/ide/surroundWith/expression/RustExpressionSurrounderBase.kt
```diff
@@ -1,15 +1,19 @@
...
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustElementFactory
 import org.rust.lang.core.psi.RustExprElement
 
-abstract class RustExpressionSurrounderBase : Surrounder {
+abstract class RustExpressionSurrounderBase<E: RustExprElement> : Surrounder {
+    abstract fun createTemplate(project: Project): E
+    abstract fun getLeafExpression(expression: E) : RustExprElement
```
matklad: Whitespace around `:` is a bit off, and there is an unused import. **Reformat code** action should take care of these.

src/main/kotlin/org/rust/ide/surroundWith/expression/RustExpressionSurrounderBase.kt
```diff
@@ -1,15 +1,19 @@
...
     abstract fun isApplicable(expression: RustExprElement): Boolean
-    abstract fun surroundExpression(project: Project, editor: Editor, expression: RustExprElement): TextRange?
+    abstract fun getSelectionRange(expression: PsiElement) : TextRange?
```
matklad: Ah, It's a shame that we can't use `E` here because of type erasure :)

src/main/kotlin/org/rust/ide/surroundWith/expression/RustExpressionSurrounderBase.kt
ujpv: I'm sorry.

src/main/kotlin/org/rust/ide/surroundWith/expression/RustExpressionSurrounderBase.kt
ujpv: Yes, it is. :(

ujpv:
I reformatted a code and removed unused imports.

src/main/kotlin/org/rust/ide/surroundWith/expression/RustExpressionSurrounderBase.kt
matklad: No need to be sorry :), everybody makes mistakes, that's what code review is for :)

matklad:
Perfect, thanks!

Issue #696, opened 2016-09-28
Tab title for mod.rs files


scharris:
When a mod.rs file is open, the tab title is only "mod.rs" (unless multiple mod.rs files happen to be open), which does not distinguish the module being viewed.  If the user has a lot of modules of this form in their own directories, it is difficult to distinguish which module is being viewed.

Workarounds currently are to keep "scroll from source" on (which otherwise might not be wanted as for example I find it distracting), or to always keep an unwanted "mod.rs" open so all mod.rs tabs are labeled with their directories for disambiguation.

I'm proposing instead that mod.rs files always show the parent directory name at least.

mkaput:
Note: Gradle support does something like this for gradle scripts

matklad:
Should be an easy thing to add: com.intellij.openapi.fileEditor.impl.EditorTabTitleProvider

mbStavola:
I'd like to attempt this.

matklad:
Thanks @mbStavola for fixing this!

@scharris have you tried disabling tabs altogether and using GoTo class/symbol/file instead for navigation :) ? 

This sounds strange, but may actually be more convenient than tabs (I don't use them myself, and I am not alone: https://www.youtube.com/watch?v=eq3KiAH4IBI&feature=youtu.be&t=10m30s ).

PR #706, opened 2016-10-02
(MISC): Qualify mod.rs files with parent dir name

mbStavola:
Overview
===
In order to help figure users understand what module file they're working in, this PR forces the editor to _qualify_ the tab name.

Previously, a tab for a module file would simply display `mod.rs`. This has been changed to include the parent directory (e.g. `foo/mod.rs`).

Addresses #696 

Technical Details
===
Created an concrete implementation of `EditorTabTitleProvider` for module files and registered it with the plugin.

When providing a tab title, we check to see if we are using a rust file and if that file is a module. If that's the case, we just pull in the parent directory's name and separate them with a forward slash.

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/RustModTabTitleProvider.kt
+++ b/src/main/kotlin/org/rust/ide/utils/RustModTabTitleProvider.kt
@@ -1,16 +1,17 @@
 package org.rust.ide.utils
 
-import com.intellij.openapi.fileEditor.impl.EditorTabTitleProvider
+import com.intellij.openapi.fileEditor.impl.UniqueNameEditorTabTitleProvider
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import org.rust.lang.core.psi.impl.isRustFile
 
-class RustModTabTitleProvider : EditorTabTitleProvider {
+class RustModTabTitleProvider : UniqueNameEditorTabTitleProvider() {
     val MOD_FILE_NAME = "mod.rs"
 
     override fun getEditorTabTitle(project: Project, file: VirtualFile): String? {
         if (file.isRustFile && MOD_FILE_NAME == file.name) {
-            return "${file.parent.name}/${file.name}"
+            return  super.getEditorTabTitle(project, file) ?:
+                "${file.parent.name}/${file.name}"
         }
 
         return null
--- a/src/main/kotlin/org/rust/ide/utils/RustModTabTitleProvider.kt
+++ b/src/main/kotlin/org/rust/ide/utils/RustModTabTitleProvider.kt
@@ -1,0 +1,18 @@
+package org.rust.ide.utils
+
+import com.intellij.openapi.fileEditor.impl.EditorTabTitleProvider
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import org.rust.lang.core.psi.impl.isRustFile
+
+class RustModTabTitleProvider : EditorTabTitleProvider {
+    val MOD_FILE_NAME = "mod.rs"
+
+    override fun getEditorTabTitle(project: Project, file: VirtualFile): String? {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/utils/RustModTabTitleProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import org.rust.lang.core.psi.impl.isRustFile
+
+class RustModTabTitleProvider : EditorTabTitleProvider {
+    val MOD_FILE_NAME = "mod.rs"
+
+    override fun getEditorTabTitle(project: Project, file: VirtualFile): String? {
+        if (file.isRustFile && MOD_FILE_NAME == file.name) {
+            return "${file.parent.name}/${file.name}"
```
mkaput: This presentation may be still unambiguous for following situation (I tend to use such module structure, so this is real life example):

- /foo
  - /moo
  - /macros
    - **/mod.rs** => `macros/mod.rs`
    - /macro_a.rs
    - ...
- /bar
  - /boo
  - /macros
    - **/mod.rs** => `macros/mod.rs`
    - /macro_b.rs
    - ...

Though I don't know yet how to handle this. Showing module name may be ugly while trying to append parent dirs until there are no ambiguities can be overworking simple problem.

matklad: Can we perhaps just fallback to `UniqueNameEditorTabTitleProvider` implementation? Like, we inherit from it, use super method to get a candidate name, and if it is `mod.rs`, then we prepend the directory name to it?

mbStavola: @mkaput  Didn't think of that!  Good catch!

@matklad Adjusted in next commit. Is this what you meant?

matklad: @mbStavola yep, this is what I meant, but looks like I was wrong about the details :) 

TabTitleProvider should return `null` when it is not applicable, and the defalt name should be used. So, UniqueNameProvider will return `null` when a name will be unique, and this will cause this provider to return `null` as well. 

I *think* the proper logic should be along this lines:

```
if is not RustFile: return null
if name != "mod.rs" return null
val unique = super.getEditorTabTitle()
if (unique != null) return unique
return "${file.parent.name}/${file.name}"
```

But maybe I am missing something again? Better to verify it!

alygin: Aren't you trying to solve a problem that is already solved by IDEA itself? There's a setting: Editor -> General -> Editor Tabs -> Show directory in editor tabs for non-unique file names. If the flag is set, IDEA adds some intelligence into tabs naming. Given the structure provided by @mkaput, it uses `bar/.../mod.rs` and `foo/.../mod.rs` as the tab titles if both files are open.

Or is it something else you're talking about?

matklad: Yep, it is helpful to prepend a directory name even if a single `mod.rs` is opened. However, when several `mod.rs` files are open, we should fall back to the default IDEA behavior.

mkaput: Hmm, actually for resolving ambiguity of my example when we have only 1 opened tab, some kind of `mod.rs` file index should be used.

matklad: >Hmm, actually for resolving ambiguity of my example when we have only 1 opened tab, some kind of mod.rs file index should be used. Do I think correctly?

Yep. I think what you want is to select a unique name among all files, and not among all *open* files, as IDEA does by default. But I don't think that we should go that far :)

matklad:
Works nicely, thanks!

PR #707, opened 2016-10-02
(T): add fuzzy test for typing chars in empty file

mkaput:


src/test/kotlin/org/rust/ide/typing/RustTypingEmptyFileFuzzyTest.kt
```diff
@@ -0,0 +1,26 @@
+package org.rust.ide.typing
+
+import com.intellij.openapi.util.text.StringUtil
+
+class RustTypingEmptyFileFuzzyTest : RustTypingTestCaseBase() {
+    override val dataPath: String = ""
+
+    fun testFuzzy() {
+        // These chars cause problems or require special treating
+        val blacklist = setOf('\b', '\t', '\r', '"', '(', '[', '{')
```
matklad: Perhaps we can type backspace more than one to handle these?

matklad:
Splendid! Have you check that this test indeed catches that bug about typing `#`?

mkaput:
Of course, it even maintains stack trace !

src/test/kotlin/org/rust/ide/typing/RustTypingEmptyFileFuzzyTest.kt
mkaput: It's still special treating, you shouldn't expect `\b` character in file

mkaput: `\b` and `\r` are rather specific cases
`\t`'s behaviour depends on user configuration
and the rest are chars we handle with our own typing handlers, and are tested in applicable places

matklad:
Thanks! I llove fuzzy tests!

PR #708, opened 2016-10-02
Project structure visualization improvement

alygin:
The PR addresses several inaccuracies and inconsistencies in how elements are displayed in the Project Structure window and in the code completion dropdown list. The main purpose is to visualize nodes the same way as other popular IntelliJ's plugins do (Java, Kotlin, Scala and C++ in CLion were the main points of reference).

Changes overview:

1. Constants and statics have a special icon with the letter `g`.
2. Function/method arguments have an icon with the letter `p`.
3. Non-mutable bindings (consts, statics, arguments, local bindings) have the final mark.
4. Static elements (statics and enum variants) have the static mark.
5. Types are displayed for constants and statics.
6. Functions/methods are displayed along with their parameter types, but without parameter names.
7. Enum variants that have tuple-like data associated with them are displayed with the types of the data.

Here're before-after pics:

pr_ba_1

pr_ba_2


src/main/kotlin/org/rust/ide/structure/RustConstTreeElement.kt
```diff
@@ -6,7 +6,7 @@ import org.rust.lang.core.psi.RustConstItemElement
 
 class RustConstTreeElement(element: RustConstItemElement) : PsiTreeElementBase<RustConstItemElement>(element) {
 
-    override fun getPresentableText(): String? = element?.name
+    override fun getPresentableText() = "${element?.name}: ${element?.type?.text}"
```
mkaput: If we don't have type (is this possible?), then we will be left with ugly trailing colon

matklad: I believe that currently the type is always there, because there are no `pin` atribues associated with const items. However I think that we add `pin`s sooner or later, so the `type` filed would become optional.

It's indeed better to be careful with string formatting, because you can write `"$foo"` even when `foo` is nullable, so you'll get a `"null"` string then.

alygin: OK, I can fix that. I don't know what `pin` is, but I assume the same fix must be applied to all places where the type information is displayed in the Project Structure (consts, statics, function arguments, enums with tuple-like data attached). Is it correct?

alexeykudinkin:
Wow! Awesome efforts @alygin!

src/main/kotlin/org/rust/ide/structure/RustConstTreeElement.kt
matklad: >Is it correct?

Yep, when using `"$stringInterpolation"` it's better to assume that things may be null, because unlike in all other Kotlin code, you won't get a warning about possibly null values there. 


And let me explain a bit about this `pin` thing. Consider simplified grammar for `const` item:

```
const_item ::= CONST ident ':' type '=' expr ';'
```

What happens if a user types a part of const declaration, and invokes completion?

```
struct FooBar;

const C: Foo|
```

The parser sees `const C: Foo`, which does not match `const_item` rule, and hence we don't know that `Foo` is a part of a type, and that we should suggest `FooBar`. 

We can fix this by adding a pin attribute to the grammar:

```
const_item ::= CONST ident ':' type '=' expr ';' { pin = ident }
```

This way, we tell the parser that as soon as `const C` part is parsed, it should produce a `const_item`, even if the production as a whole does not match. But this means that the `type` field of the `RustConstItem` should be nullable: you can't omit the type in the valid Rust code, but it may be not present in an incomplete code. 

So at the moment we don't apply pin attribute to const items (and as a result completion does not work), and we have the following PSI:

```Java
public interface RustConstItemElement extends ... {
  @NotNull
  RustTypeElement getType();
  ...
}
```

And when we would add the pin, the PSI would look like this:

```Java
public interface RustConstItemElement extends ... {
  @Nullable // <- here is the change
  RustTypeElement getType();
  ...
}
```

matklad:
Great, thanks!

src/main/kotlin/org/rust/ide/structure/RustConstTreeElement.kt
```diff
@@ -6,7 +6,13 @@ import org.rust.lang.core.psi.RustConstItemElement
 
 class RustConstTreeElement(element: RustConstItemElement) : PsiTreeElementBase<RustConstItemElement>(element) {
 
-    override fun getPresentableText(): String? = element?.name
+    override fun getPresentableText(): String? {
+        var text = element?.name
+        if (element?.type != null) {
```
matklad: An alternative way to write this would be

```
element?.type?.let {
    text += ": ${it.text}"
}
```

src/main/kotlin/org/rust/ide/structure/RustConstTreeElement.kt
alygin: Thanks for the thorough explanation! It really clears things up.

src/main/kotlin/org/rust/ide/structure/RustConstTreeElement.kt
alygin: I'm still trying to accommodate myself to idiomatic Kotlin, so thanks for the advise, it's really appreciated!

PR #710, opened 2016-10-03
(ACTION): string literal joiner

### LINKED ISSUE #195: Support splitting and joining string literals

mkaput:
fix #195

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandler.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandler.kt
@@ -1,0 +1,38 @@
+package org.rust.ide.actions
+
+import com.intellij.codeInsight.editorActions.JoinLinesHandlerDelegate
+import com.intellij.codeInsight.editorActions.JoinLinesHandlerDelegate.CANNOT_JOIN
+import com.intellij.openapi.editor.Document
+import com.intellij.psi.PsiFile
+import com.intellij.util.text.CharSequenceSubSequence
+import org.rust.ide.typing.endsWithUnescapedBackslash
+import org.rust.lang.core.psi.RustTokenElementTypes
+import org.rust.lang.core.psi.impl.RustFile
+import org.rust.lang.core.psi.util.elementType
+
+class RustStringLiteralJoinLinesHandler : JoinLinesHandlerDelegate {
+    override fun tryJoinLines(document: Document, file: PsiFile, offsetNear: Int, end: Int): Int {
+        if (file !is RustFile) return CANNOT_JOIN
+
+        val text = document.charsSequence
+        val leftPsi = file.findElementAt(offsetNear) ?: return CANNOT_JOIN
+        val rightPsi = file.findElementAt(end) ?: return CANNOT_JOIN
+
+        if (leftPsi != rightPsi || leftPsi.elementType !in RustTokenElementTypes.STRING_LITERALS) return CANNOT_JOIN
+
+        var start = offsetNear
+
+        // Strip newline escape
+        if (CharSequenceSubSequence(text, 0, start + 1).endsWithUnescapedBackslash()) {
+            start--
+            while (start >= 0 && (text[start] == ' ' || text[start] == '\t')) {
+                start--
+            }
+        }
+
+        document.deleteString(start + 1, end)
+        document.insertString(start + 1, " ")
+
+        return start + 1
+    }
+}
--- a/src/test/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandlerTest.kt
+++ b/src/test/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandlerTest.kt
@@ -1,0 +1,116 @@
+package org.rust.ide.actions
+
+import com.intellij.openapi.actionSystem.IdeActions
+import org.intellij.lang.annotations.Language
+import org.rust.lang.RustTestCaseBase
+
+class RustStringLiteralJoinLinesHandlerTest : RustTestCaseBase() {
+    override val dataPath: String = ""
+
+    private fun doTest(
+        @Language("Rust") before: String,
+        @Language("Rust") after: String
+    ) {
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+
+        val text = document.charsSequence
+        val leftPsi = file.findElementAt(offsetNear) ?: return CANNOT_JOIN
+        val rightPsi = file.findElementAt(end) ?: return CANNOT_JOIN
+
+        if (leftPsi == rightPsi && leftPsi.elementType in RustTokenElementTypes.STRING_LITERALS) {
+            var start = offsetNear
+
+            // Strip newline escape
+            if (CharSequenceSubSequence(text, 0, start + 1).endsWithUnescapedForwardSlash()) {
```
mkaput: Omg IntelliJ does string copying of entire document here o_O [source](https://upsource.jetbrains.com/idea-ce/file/idea-ce-5418ee4421e14e3eeb4cb10eca6407ee9618379e/plugins/properties/src/com/intellij/lang/properties/PropertiesJoinLinesHandler.java?nav=1162:1200:focused&line=0)

src/main/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+
+class RustStringLiteralJoinLinesHandler : JoinLinesHandlerDelegate {
+    override fun tryJoinLines(document: Document, file: PsiFile, offsetNear: Int, end: Int): Int {
+        if (file !is RustFile) return CANNOT_JOIN
+
+        val text = document.charsSequence
+        val leftPsi = file.findElementAt(offsetNear) ?: return CANNOT_JOIN
+        val rightPsi = file.findElementAt(end) ?: return CANNOT_JOIN
+
+        if (leftPsi == rightPsi && leftPsi.elementType in RustTokenElementTypes.STRING_LITERALS) {
```
matklad: I would have inverted this `if` to save indentation.

src/test/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandlerTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    fun testNoEscape() = doTest(
+        """
+        fn main() {
+            "Hello<caret>,
+             World"
+        }
+        """,
+        """
+        fn main() {
+            "Hello,<caret> World"
```
matklad: It does not matter here, but you can use `/*caret*/` now: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/test/kotlin/org/rust/lang/RustTestCaseBase.kt#L91

src/test/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandlerTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    fun testEscapedButNotEscapedInFactNewlineEscape() = doTest(
+        """
+        fn main() {
+            "He<caret>llo, \\\
+             World"
+        }
+        """,
+        """
+        fn main() {
+            "Hello, \\<caret> World"
```
matklad: Hm, shouldn't this be `"Hello, \\<caret>\n                                 World"`?

src/main/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandler.kt
mkaput: I'm considering implementing joining this:
```rust
"aaa" +
    "bbb"
```

That's why this if has such form

src/test/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandlerTest.kt
mkaput: I don't know. I'm not using this feature on the daily basis, so I may have bad assumptions about it's working.

src/main/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandler.kt
matklad: But do we need this? `let s = "Hello" + "world";` is a compile time error in Rust.

src/test/kotlin/org/rust/ide/actions/RustStringLiteralJoinLinesHandlerTest.kt
matklad: Same here :) I think that adding a `\n` is the correct behavior here, but let's not do this for now and wait for the feedback from users?

mkaput: We already have fast feedback on such things (looking at you `#` hehe), so this is good idea :)

mkaput:
Looks like it's finised

matklad:
:+1:

PR #711, opened 2016-10-04
(MISC) Drop Reference inspection

alygin:
I've implemented an analogue of the Clippy's drop_ref inspection with a quick fix that converts reference argument into the owned value.

It works well when is used inside IDEA but I need advise to make its tests work. The problem is that `RustReference.resolve()` works differently when is called within unit tests. Particularly, [this line](https://github.com/alygin/intellij-rust/blob/3d0a41bd9ad842af18fcdf40942782be79123777/src/main/kotlin/org/rust/ide/inspections/RustDropRefInspection.kt#L30) is expected to return a `RustFnItemElement` (and does it when is running inside IDEA), but always return `null` in unit tests.

I'm tending to think that I need to implement `org.rust.lang.core.resolve.ref.RustFnCallReferenceImpl`. But I'm not sure and the fact that it only breaks when is running as a part of a unit test perplexes me. What am I missing here?


matklad:
>But I'm not sure and the fact that it only breaks when is running as a part of a unit test perplexes me. What am I missing here?

TLDR: look at the [stdlib aware tests].

There main difference between opening a project in the IDE and the unit test environment is the absence of Cargo and standard library in most tests. 

Cargo is used to determine crate roots, and you need crate roots to resolve `::absolute::paths`. In the test environment, `main.rs` and `lib.rs`. files are [hardcoded] as crate roots.  So, you need either to make sure the test file is names `main.rs` ([example]), or to use a more modern [inline code] tests. 

To make stdlib available in test, you need to setup special [project descriptor].


[stdlib aware tests]: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/test/kotlin/org/rust/lang/core/resolve/RustStdlibResolveTestCase.kt#L11

[hardcoded]: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/test/kotlin/org/rust/lang/RustTestCaseBase.kt#L175-L178

[example]: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/test/kotlin/org/rust/lang/core/resolve/RustStdlibResolveTestCase.kt#L24

[inline code]: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/test/kotlin/org/rust/lang/core/resolve/RustUseResolveTestCase.kt#L11

[project descriptor]: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/test/kotlin/org/rust/lang/core/resolve/RustStdlibResolveTestCase.kt#L11

src/main/kotlin/org/rust/ide/inspections/RustDropRefInspection.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    fun inspectExpr(expr: RustCallExprElement, holder: ProblemsHolder) {
+        val pathExpr = expr.expr
+        if (pathExpr is RustPathExprElement) {
+            val resEl = pathExpr.path.reference.resolve()
+            if (resEl is RustFnItemElementImpl && resEl.canonicalCratePath.toString() == "::mem::drop") {
+                val args = expr.argList.exprList
+                if (args.size == 1) {
+                    val arg = args[0]
+                    if (arg is RustUnaryExprElement && arg.expr is RustPathExprElement) {
```
matklad: Hey, this can be more precise and catch cases like

```
let x = &foo();
drop(x)
```

if you look at the type of the expression: `if (arg.resolvedTye is RustReferenceType)`. The entry to the type machinery is [here](https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/main/kotlin/org/rust/lang/core/types/util/RustPsiExtensions.kt#L12). It is very WIP at the moment, but it should be strictly better than checking for literal `&` (if it is not, it is a bug :) ).

src/main/kotlin/org/rust/ide/inspections/RustDropRefInspection.kt
```diff
@@ -0,0 +1,45 @@
...
+                        holder.registerProblem(
+                            expr,
+                            "Call to std::mem::drop with a reference argument. Dropping a reference does nothing",
+                            RemoveRefFix(arg, "Call with owned value"))
+                    }
+                }
```
matklad: Using `ElementPattern` API may, or may not be the easier way to achieve this. You can see an example of pattern usage here: https://github.com/intellij-rust/intellij-rust/blob/3dd1a3f922f8f59ccfe6b2705781cfb82f2784d4/src/main/kotlin/org/rust/lang/core/completion/AttributeCompletionProvider.kt

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveRefFix.kt
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Fix that converts the given reference to owned value.
+ * @param argEl An element, that represents a reference from which the first
+ * symbol '&' must be removed.
+ * @param fixName A name to use for the fix instead of the default one to better fit the inspection.
+ */
+class RemoveRefFix(
+    val argEl: RustExprElement,
+    val fixName: String = "Change reference to owned value") : LocalQuickFix {
```
matklad: We usually move `) :` on the new line. That way, the class body is more visible, and the extends list has a predictable position:

```
class RemoveRefFix(
    val argEl: RustExprElement,
    val fixName: String = "Change reference to owned value"
) : LocalQuickFix {
```

alygin:
@matklad, thanks for the feedback, it's very helpful! I'll fix the issues.

alygin:
Fixed and refactored:

- Tests are stdlib aware and implemented in a more-modern-inline-code way.
- Fixed broken folding tests.
- `if (arg.resolvedTye is RustReferenceType)` is now used instead of `RustUnaryExpElement`, so all reference types are caught by the inspection. But fix is only suggested when the argument starts with `&`.
- Fixed code style and did other minor refactoring.

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
```diff
@@ -72,6 +72,15 @@ class RustFoldingBuilder() : FoldingBuilderEx(), DumbAware {
                 }
             }
 
+            override fun visitFnItem(o: RustFnItemElement) {
```
matklad: Ah, am I correct that this is necessary, because visitor hierarchy changes after adding `RustPathNameElement` to `RustFnItem`?

matklad:
Wonderful!

src/main/kotlin/org/rust/ide/folding/RustFoldingBuilder.kt
alygin: Yes, the behaviour changed after adding `RustPathNameElement` to `RustFnItem`. So, it's covered by several (four to be exact) existing tests.

PR #712, opened 2016-10-04
(TYPE): Resolve ==,!=,<=,>=,<,>,||,&& to bool

ujpv:
Now expressions with these operators are resolving in undefined type.

I suggest to resolve it in bool.
If it's not too ugly :), it can greatly extend the scope for application surrounders (if exp, for example) and other plugin features.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustTypificationEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustTypificationEngine.kt
@@ -140,7 +140,28 @@ private class RustExprTypificationVisitor : RustComputingVisitor<RustType>() {
     override fun visitParenExpr(o: RustParenExprElement) = set { o.expr.resolvedType }
 
     override fun visitBinaryExpr(o: RustBinaryExprElement) {
-        super.visitBinaryExpr(o)
+        when (o.operatorType) {
+            RustTokenElementTypes.ANDAND,
+            RustTokenElementTypes.OROR -> {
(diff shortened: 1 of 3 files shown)


matklad:
>1 && 2 should yield unknown type IMHO, as it's typing error

Not really, we don't need to check types here, we just need to compute some plausible type here (that is, we can "type check" more programs that Rustc does).

Take a look at the rule for if: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustTypificationEngine.kt#L133

It does not check that the arms have the same type, it just returns the type of the first arm. I argue that it is a more useful behavior, because it may allow to typecheck more code down the line. 

But I agree that `true && false` is a better test than `1 && 2` :)

ujpv:
For && and || it very easy to fix, and 1 && 2 will resolved to undefined type.

ujpv:
I fixed it.
Now expressions && and || with type error resolving in unknown type.
Also I added argument `description` to `fun testExpr`, for custom fail description.

matklad:
Sorry, I was not clear enough. The previous algorithm for type inference for `&&` and `||` was correct. We should look only on the operator, and completely ignore the operands. This is helpful in the following case, for example:


```
foo.is_some() && (bar > some_unresolved_name)
```

We should be able to say that this is a boolean expression even if we don't know the type of inner expressions because of name resolution errors. 

Of course, if you were writing a compiler, you should have bail with error in this case, but the requirements of the IDE are different :)

ujpv:
I removed the type checking for arguments || and &&.
Please, review again.

PR #713, opened 2016-10-05
(MISC) Double Negation inspection

alygin:
Inspection that is an analogue of Clippy's [double_neg](https://github.com/Manishearth/rust-clippy/wiki#double_neg) lint. Catches expressions of the form `--x` which are no-op in Rust.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RustDoubleNegInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustDoubleNegInspection.kt
@@ -1,0 +1,30 @@
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElementVisitor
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/inspections/RustDoubleNegInspection.kt
```diff
@@ -0,0 +1,30 @@
...
+        return object : RustElementVisitor() {
+            override fun visitUnaryExpr(expr: RustUnaryExprElement) {
+                if (isNegation(expr) && isNegation(expr.expr)) {
+                    holder.registerProblem(expr, "--x could be misinterpreted as a pre-decrement, but effectively is a no-op")
+                }
+            }
+        }
+    }
+
+    private fun isNegation(el: RustExprElement?) =
```
alygin: You're right, that looks better

matklad:
Thanks!

PR #714, opened 2016-10-05
(MISC) Suspicious Assignment inspection

alygin:
An analogue of the Clippy's [suspicious_assignment_formatting](https://github.com/Manishearth/rust-clippy/wiki#suspicious_assignment_formatting) lint with a couple of quick fixes.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RustSuspiciousAssignmentInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustSuspiciousAssignmentInspection.kt
@@ -25,21 +25,29 @@ class RustSuspiciousAssignmentInspection : RustLocalInspectionTool() {
         return object : RustElementVisitor() {
             override fun visitBinaryExpr(expr: RustBinaryExprElement) {
                 if (expr.operator.text == "=") {
-                    val rightExpr = findUnaryExpr(expr.right)
-                    if (rightExpr is RustUnaryExprElement && rightExpr.expr != null) {
-                        val uExpr = rightExpr.expr
-                        val op = rightExpr.text[0]
-                        if (uExpr != null
+                    val unaryExpr = findUnaryExpr(expr.right)
+                    if (unaryExpr is RustUnaryExprElement && unaryExpr.expr != null) {
+                        val unaryBody = unaryExpr.expr
+                        val op = unaryExpr.text[0]
+                        if (unaryBody != null
                             && (op == '-' || op == '*' || op == '!')
-                            && expr.operator.distanceTo(rightExpr) == 1
-                            && expr.operator.distanceTo(uExpr) > 2) {
-                            val uExprOffset = uExpr.textRange.startOffset - expr.left.textRange.startOffset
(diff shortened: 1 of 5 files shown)


alygin:
There's something wrong with the `RustupTest.testDownloadStdlib()` test in one of the jobs. I don't think it's somehow connected to my changes.

Could someone, please, restart the checks?

src/main/kotlin/org/rust/ide/inspections/RustSuspiciousAssignmentInspection.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.rust.lang.core.psi.*
+
+/**
+ * Checks for use of the non-existent =*, =! and =- operators that are probably typos but can be compiled.
+ * Analogue of Clippy's suspicious_assignment_formatting.
+ * QuickFix 1: Change `a =? b` to `a ?= b`
+ * QuickFix 2: Change `a =? b` to `a = ?b`
+ */
+class RustSuspiciousAssignmentInspection : RustLocalInspectionTool() {
+    private object Constants {
```
matklad: No need to name this object, you can use the companion:

```Kotlin
companion object {
    private val LONG_TEXT_THRESHOLD = 10
    private val LONG_TEXT_SUBST = ".."
}
```

matklad: And we tend to place companions at the bottom.

src/main/kotlin/org/rust/ide/inspections/RustSuspiciousAssignmentInspection.kt
```diff
@@ -0,0 +1,79 @@
...
+        val LONG_TEXT_THRESHOLD = 10
+        val LONG_TEXT_SUBST = ".."
+    }
+
+    override fun getDisplayName(): String = "Suspicious assignment"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : RustElementVisitor() {
+            override fun visitBinaryExpr(expr: RustBinaryExprElement) {
+                if (expr.operator.text == "=") {
```
alygin: I'm totally ok with `if (!cond) return` in such cases, just usually try to follow a more conservative rule of having only one `return` per function in a new project until I get familiar with the adopted code style : )
I'll do the refactoring.

matklad: >having only one return per function

This is actually a rather interesting aspect of Kotlin code style. I really do prefer early returns for handling "unusual" conditions. However inside lambdas `return` must be qualified ( `xs.map { return@map 92 }` ) which is not pretty, so with lambdas I usually stick with a single expression body.

alygin:
@matklad, thanks for the review! The code looks cleaner now. I'll go over other inspections in a separate PR to bring them to the same code style.

matklad:
Thanks a lot!

matklad:
BTW, INSP tag is a bit more precise than MISC.

PR #716, opened 2016-10-06
(INSP) Missing Else inspection

alygin:
A partial analogue of the Clippy's [suspicious_else_formatting](https://github.com/Manishearth/rust-clippy/wiki#suspicious_else_formatting) inspection. Detects `if`-statements with potentially missing `else`s. A quick fix adds an `else` to the `if`.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
@@ -21,17 +21,17 @@ class RustMissingElseInspection : RustLocalInspectionTool() {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
         object : RustElementVisitor() {
             override fun visitExprStmt(expr: RustExprStmtElement) {
-                if (expr.extractIf() == null) return
-                val nextPair = expr.nextSibling.consumeSpaces() ?: return
-                val spaceLen = nextPair.second
-                val nextEl = nextPair.first
-                val nextIf = nextEl.extractIf() ?: return
-                val rangeStart = expr.startOffsetInParent + expr.textLength
-                val range = TextRange(rangeStart, rangeStart + spaceLen + nextIf.expr.startOffsetInParent)
+                val firstIf = expr.extractIf() ?: return
+                val nextIf = expr.rightSiblings
+                    .dropWhile { (it is PsiWhiteSpace || it is PsiComment) && '\n' !in it.text }
+                    .firstOrNull()
+                    .extractIf() ?: return
+                val rangeStart = expr.startOffsetInParent + firstIf.textLength
+                val rangeLen = nextIf.expr.textRange.startOffset - firstIf.textRange.startOffset - firstIf.textLength
                 val fixRange = TextRange(nextIf.textRange.startOffset, nextIf.textRange.startOffset)
                 holder.registerProblem(
                     expr.parent,
-                    range,
+                    TextRange(rangeStart, rangeStart + rangeLen),
                     "Suspicious if. Did you mean `else if`?",
                     SubstituteTextFix(expr.containingFile, fixRange, "else ", "Change to `else if`"))
             }
@@ -43,21 +43,5 @@ class RustMissingElseInspection : RustLocalInspectionTool() {
         else -> null
     }
 
-    /**
-     * Finds the first non-space/comment sibling starting from this element.
-     * Returns the found sibling along with the length of the consumed spaces/comments.
-     * If there's a space/comment with a line break, returns `null`.
-     */
-    private fun PsiElement?.consumeSpaces(): Pair<PsiElement?, Int>? {
-        var nextEl = this
-        var len = 0
-        while (nextEl is PsiWhiteSpace || nextEl is PsiComment) {
-            if (nextEl.textContains('\n')) {
-                return null
-            }
-            len += nextEl.textLength
-            nextEl = nextEl.nextSibling
-        }
-        return Pair(nextEl, len)
-    }
+    private val PsiElement.rightSiblings: Sequence<PsiElement> get() = generateSequence(this.nextSibling) { it.nextSibling }
 }
--- a/src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
@@ -1,0 +1,62 @@
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiComment
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import org.rust.ide.inspections.fixes.SubstituteTextFix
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+                    "Suspicious if. Did you mean `else if`?",
+                    SubstituteTextFix(expr.containingFile, fixRange, "else ", "Change to `else if`"))
+            }
+        }
+
+    private fun PsiElement?.extractIf(): RustIfExprElement? = when(this) {
+        is RustIfExprElement -> this
+        is RustExprStmtElement -> firstChild.extractIf()
+        else -> null
+    }
```
mkaput: I believe there is dedicated method for this in `PsiTreeUtil`

mkaput: Ugh, looks like there isn't any 😞

alygin: I'm not fully confident that `PsiTreeUtil.getChildOfType()` won't produce false positives. It iterates through all the children while I only need the first one, otherwise this if-block is not of interest for this inspection.

alygin: Oh, sorry, didn't notice your last comment.

mkaput: Yeah, I agree. Clicked "submit" before thinking ;)

matklad:
Am I correct that this would not produce a warning in this case:

```Rust
if (cond) {
    // Do some preparation stuff.
}

{
    let file = // open a file
    write!(file);
    // File is dropped. 
}
// Now the file is closed, the buffer is flushed and the data is safe in the OS cache.
```

? I think that we want some tests that check that `\n` does indeed prevent this inspection from being triggered.

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+ * QuickFix: Change to `else if`
+ */
+class RustMissingElseInspection : RustLocalInspectionTool() {
+    override fun getDisplayName(): String = "Missing else"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RustElementVisitor() {
+            override fun visitExprStmt(expr: RustExprStmtElement) {
+                if (expr.extractIf() == null) return
+                val nextPair = expr.nextSibling.consumeSpaces() ?: return
```
matklad: Can use unpacking here: `val (nextEl, spaceLen) = expr.nextSibling.consumeSpaces()`

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+        is RustExprStmtElement -> firstChild.extractIf()
+        else -> null
+    }
+
+    /**
+     * Finds the first non-space/comment sibling starting from this element.
+     * Returns the found sibling along with the length of the consumed spaces/comments.
+     * If there's a space/comment with a line break, returns `null`.
+     */
+    private fun PsiElement?.consumeSpaces(): Pair<PsiElement?, Int>? {
```
matklad: Perhaps this can be `Pair<PsiElement, Int>?` rather than `Pair<PsiElement?, Int>?`

alygin:
It surely won't emit a warning in such a case. It requires two consecutive `if`s with no line breaks between them. Something like

    if (cond) {
        // ...
    } if (another_cond) {
        // ...
    }

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
alygin: an eagle eye :)

matklad:
Yep, so I think two test are neccessary:

```
if {
    92;
}
{}
```

and

```
if true {}
if true {}
```

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
alygin: Though, it allows `null`s in cases when a space or comment is the last element in the block. The function will consume it and return (null, len).

matklad: It's an interesting phenomenon with Kotlin. It's so easy to propagate `null`s, that sometimes an API grows null although it is not strictly necessary. This happens, for example, when generating an override for a not annotated Java method: https://youtrack.jetbrains.com/issue/KT-12213

matklad: >The function will consume it and return (null, len).

And then we return anyway because of `val nextIf = nextEl.extractIf() ?: return`

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+        var nextEl = this
+        var len = 0
+        while (nextEl is PsiWhiteSpace || nextEl is PsiComment) {
+            if (nextEl.textContains('\n')) {
+                return null
+            }
+            len += nextEl.textLength
+            nextEl = nextEl.nextSibling
```
alygin: Actually, I don't see how it would help. In this inspection, we only want to jump over spaces/comments and only get what sibling element follows them. We also don't need those spaces/comments in a collection, we only need their length to properly highlight the inspection range.

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
alygin: I see your point and don't like this `null` too. My thought was that it's not the `consumeSpaces()` responsibility to decide on how to interpret its results. But given the fact that it already contains a logic of returning `null` when a line break is captured, you're right, it may also return `null` when nothing left after the consuming.

src/main/kotlin/org/rust/ide/inspections/RustMissingElseInspection.kt
matklad: I was thinking about something like 

```Kotlin
rightSiblings
  .dropWhile { (it is PsiWhitespace || it is PsiComment) && '\n' !in it.text}
  .firstOrNull()
  .extractIf()
```

Note that sequence is not a collection, so this is pretty efficient memory wise.

We won't be able to calculate length nicely then, but perhaps it can be derived somehow from the offsets of the existing elements?

alygin: Looks nice! I'll try to refactor the highlighting logic.

alygin:
Looks like a glitch on the build server. @matklad, could you, please, restart the checks?

matklad:
> @matklad, could you, please, restart the checks?

Looks like the easiest way to restart travis is to close and reopen the PR. But anyway, I am usually confident when appveyor is green, because we don't have that much platform dependent code.

PR #717, opened 2016-10-07
(SURR): Refactor statements surrounders

ujpv:
The major part of code moved to `RustStatementsSurrounderBase`.

In src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt 
Тype of `var current` is `PsiElement`. But current.nextSibling is Nullable.
For this reason, exceptions periodically appear.

### DIFF
--- a/src/main/kotlin/org/rust/ide/surroundWith/statement/RustStatementsSurrounderBase.kt
+++ b/src/main/kotlin/org/rust/ide/surroundWith/statement/RustStatementsSurrounderBase.kt
@@ -1,24 +1,44 @@
 package org.rust.ide.surroundWith.statement
 
+import com.intellij.codeInsight.CodeInsightUtilBase
 import com.intellij.lang.surroundWith.Surrounder
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
+import org.rust.ide.surroundWith.addStatements
+import org.rust.lang.core.psi.RustBlockElement
 
 abstract class RustStatementsSurrounderBase : Surrounder {
-    abstract fun surroundStatements(
-        project: Project,
-        editor: Editor,
-        container: PsiElement,
-        statements: Array<out PsiElement>
-    ): TextRange?
+    protected abstract fun createTemplate(project: Project): PsiElement
+    protected abstract fun getCodeBlock(expression: PsiElement): RustBlockElement
 
-    final override fun isApplicable(elements: Array<out PsiElement>): Boolean = elements.isNotEmpty()
+    protected open fun getExprForRemove(expression: PsiElement): PsiElement? = null
+    protected open val shouldRemoveExpr = false
+
+    final override fun isApplicable(elements: Array<out PsiElement>): Boolean =
+        elements.isNotEmpty()
 
     final override fun surroundElements(project: Project, editor: Editor, elements: Array<out PsiElement>): TextRange? {
         require(elements.isNotEmpty())
         val container = requireNotNull(elements[0].parent)
-        return surroundStatements(project, editor, container, elements)
+
+        var template = createTemplate(project)
+        template = container.addBefore(template, elements[0])
+
+        getCodeBlock(template).addStatements(elements)
+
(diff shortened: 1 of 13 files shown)


src/main/kotlin/org/rust/ide/surroundWith/statement/RustStatementsSurrounderBase.kt
```diff
@@ -1,24 +1,44 @@
...
-        editor: Editor,
-        container: PsiElement,
-        statements: Array<out PsiElement>
-    ): TextRange?
+    abstract fun createTemplate(project: Project): PsiElement
+    abstract fun getCodeBlock(expression: PsiElement): RustBlockElement
 
-    final override fun isApplicable(elements: Array<out PsiElement>): Boolean = elements.isNotEmpty()
+    open fun getExprForRemove(expression: PsiElement): PsiElement? = null
+    open fun isNeedRemoveExpr() = false
```
matklad: There functions can be `protected abstract` and `protected open`. `isNeedRemoveExpr` could be a `val`:

```
protected open val shouldRemoveExpr: Boolean
```

Even better, I think you can have only `getExprForRemove`, which would return `null` if nothing should be removed.

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
```diff
@@ -146,12 +146,12 @@ private fun PsiElement.getTopmostParentInside(parent: PsiElement): PsiElement {
 
 private fun collectElements(start: PsiElement, stop: PsiElement?, pred: (PsiElement) -> Boolean): Array<out PsiElement> {
     val list = mutableListOf<PsiElement>()
```
matklad: I think it's a good idea to add `check (stop == null || start.parent == stop.parent)` here.

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
```diff
@@ -146,12 +146,12 @@ private fun PsiElement.getTopmostParentInside(parent: PsiElement): PsiElement {
...
-    var current = start
+    var current: PsiElement? = start
     while (current != stop) {
-        if (pred(current)) {
-            list.add(current)
+        if (pred(checkNotNull(current))) {
+            list.add(checkNotNull(current))
         }
-        current = current.nextSibling
+        current = checkNotNull(current).nextSibling
```
matklad: I think a simple solution would be to add `current = current.nextSibling ?: break`. Also, I have a suspicion that this function can be neatly rewritten with the help of `generateSequence` (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/generate-sequence.html).

matklad:
>But current.nextSibling is Nullable. For this reason, exceptions periodically appear.

Good catch! Perhaps we can add a test for some surrounded which would trigger this behavior?

src/main/kotlin/org/rust/lang/core/psi/util/RustPsiNavigationExtensions.kt
ujpv: Thanks. 
Really is possible to make it more beautiful.

ujpv:
Yes, of course!

src/main/kotlin/org/rust/ide/surroundWith/statement/RustStatementsSurrounderBase.kt
ujpv: I tried to get rid of `shouldRemoveExpr`. But did not do it so briefly.
Because, in `fun surroundElements`  we should get `shouldRemoveExpr`. If `shouldRemoveExpr==true` we need to do some work. And only after this we can get valid `getExprForRemove`.

matklad: Ah, I see. I don't know the proper solution to this off the top of my head, so let's leave this as is. Perhaps we'll invent something better in the future :)

ujpv:
I fixed issues.
+
I added a test case with NPE in collectElements.
+
Test case for incorrect reformatting after apply "Surround with..."

PR #718, opened 2016-10-07
(INSP) Dangling Else inspection

alygin:
A partial analogue of the Clippy's [suspicious_else_formatting](https://github.com/Manishearth/rust-clippy/wiki#suspicious_else_formatting) inspection. Detects `else if`s that are separated by new lines:

    if (cond) {
    } else

    if (another_cond) {
    }

The inspection provides quick fixes for removing `else` and for joining `else if`.


matklad:
Thanks!

PR #719, opened 2016-10-07
Recognize assert!, assert_eq! and assert_ne! macros

### LINKED ISSUE #636: Special case assert! and assert_eq! as well as try!
Would it be possible to make Goto Declarations work inside assert/assert_eq as well as try!?

kumbayo:
Fixes #636


matklad:
>Is there a better way to test macros or is this OK?

I think this is fine. An alternative would be to add resolve tests for macros, [like these](https://github.com/intellij-rust/intellij-rust/blob/bc18de767ab32351a9459edfeb37413572571e80/src/test/kotlin/org/rust/lang/core/resolve/RustResolveTestCase.kt#L360-L377).

matklad:
Thanks!

PR #723, opened 2016-10-09
(INSP) Naming Convention inspections

alygin:
A set of inspections that match the `rustc`'s lints `non_snake_case`, `non_camel_case_types` and `non_upper_case_globals` with the following minor differences:
- Associated constants are not inspected (they are experimental in Rust and not supported by the plugin so far, if I understand correctly)
- Associated types are inspected while `rustc` ignores them when checking names
- The inspections sometimes offer better renaming options (`fooBar` &rarr; `FooBar` instead of `fooBar` &rarr; `Foobar` in `rustc`)

I've extracted separate inspections for every element type (the same way it's done in Java) because it will help to adjust inspections with higher level of granularity after improvements (see below).

Further improvements I'm thinking of:
- Analyze `#[allow(...)]` attributes to suppress inspections.
- Add a Rename fix. I have a working implementation that leverages IntelliJ's refactoring facilities, but it throws an exception somewhere in its internals that warns about possible deadlocks. So I'm not confident to include it into the PR so far.
- In the future, inspections may also impose reasonable name length limits (`rustc` doesn't care much about it) and RegEx to match to (as they do in Java).

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
@@ -47,16 +47,18 @@ open class RustCamelCaseNamingInspection(
         val result = StringBuilder(name.length)
         var wasUnderscore = true
         var startWord = true
-        name.forEach { char ->
-            if (char == '_') {
-                wasUnderscore = true
-            } else if (wasUnderscore || startWord && char.canStartWord) {
-                result.append(char.toUpperCase())
-                wasUnderscore = false
-                startWord = false
-            } else {
-                startWord = char.isLowerCase()
-                result.append(char.toLowerCase())
+        for (char in name) {
+            when {
+                char == '_' -> wasUnderscore = true
+                wasUnderscore || startWord && char.canStartWord -> {
+                    result.append(char.toUpperCase())
+                    wasUnderscore = false
+                    startWord = false
+                }
+                else -> {
+                    startWord = char.isLowerCase()
+                    result.append(char.toLowerCase())
+                }
             }
         }
         return if (result.isEmpty()) "CamelCase" else result.toString()
--- a/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
@@ -1,0 +1,253 @@
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.RustParameterElementImpl
+
+/**
+ * Base class for naming inspections. Implements the core logic of checking names
+ * and registering problems.
+ */
+abstract class RustNamingInspection(
+    val elementType: String,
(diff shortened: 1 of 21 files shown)


src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
```diff
@@ -0,0 +1,253 @@
...
+        }
+        return Pair(false, if (str.isEmpty()) "UPPER_CASE" else name.toSnakeCase(true))
+    }
+}
+
+private fun String.toSnakeCase(upper: Boolean): String {
+    val result = StringBuilder(length + 3)     // 3 is a reasonable margin for growth when `_`s are added
+    result.append(takeWhile { it == '_' || it == '\'' })    // Preserve prefix
+    var firstPart = true
+    drop(result.length).splitToSequence('_').forEach pit@ { part ->
```
matklad: TIL that you can `label@` lambdas, thanks a bunch! :)

matklad:
Nice!

What do you think about introducing a single "Naming convention" inspection with a number of options like "check lifetimes names", "check struct names", etc? Here is an example of inspection with parameters: [HasPlatformTypeInspection](https://github.com/JetBrains/kotlin/blob/1e59f96f6b099c550817e6e88189f4b7eb05fe18/idea/src/org/jetbrains/kotlin/idea/inspections/HasPlatformTypeInspection.kt).

I think that we do need fine grained configuration here (Niko Matsakis loves SCREAMING_SNAKE_CASE for lifetimes, for example), but at the same time we don't want bazillion similar inspections in the UI.

src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
alygin: Actually, I found it out when made a research after [your comment](https://github.com/intellij-rust/intellij-rust/pull/714#discussion_r82158230). So the thanks are mutual )

matklad:
Ah, I've missed the point about supplying a regex to validate a name. With this in mind, I think it's better to stick to separate inspections!

alygin:
In fact, I started with a single inspection, but then, after looking at some languages with mature support in IntelliJ (Java, JavaScript, Groovy), I divided them into pieces. The arguments were:

- It will allow to easily add additional parameters to inspections. For instance, length limits (for instance, to enforce one-character type parameters) or regexps (for Niko Matsakis to be able to enforce screaming cases for lifetimes in his projects).
- It will allow to redefine the severity of warnings for every element type separately which sometimes is quite handy.
- All these inspections are placed in a separate group and don't clutter the interface.

What do you think?

matklad:
Yep, separate inspections are indeed better!

matklad:
An alternative way to test these would be

1) to write tests for `checkName`, `suggestName` variants. These are pure functions independent of IDEA machinery, so they can be easily tested without using fixtures. It should be faster and a bit easier to add new test cases. 

2) For each inspection, write a single test that checks that it is indeed aplicable to a particular PSI node and that quick fix works.

alygin:
I agree on tests, will refactor them.

alygin:
Refactored

matklad:
Awesome, as usual!

PR #728, opened 2016-10-11
(INSP) Rename Fix for Naming Convention inspections

alygin:
I've implemented the Rename fix for the Naming Convention inspections. It renames the element to the suggested name. The next element types are not supported:

- Associated types
- Macros
- Lifetimes
- Trait methods

It seems there're no links between declarations of these elements and their usages. When such links are added, the fix will be easy to activate for them too.


matklad:
Thanks!

PR #732, opened 2016-10-12
(INSP): Fix for Naming Convention inspection false positives

### LINKED ISSUE #730: Naming convention inspection false positive
```
match Some(()) {
    None => ()
}
```

The `None` here triggers inspection, while it shouldn't. The underlying problem is that we don't distinguish between name bindings and nullary enum variants in patterns. That is, for the following code

```
match Some(()) {
    None => (),
    Nane => (),
}
```

`None` is a constant, so it should not trigger the warning, while `Nane` is a variable name, and should be highlighted with red. I think we should just ignore variable names in patterns until we are able to properly distinguish the two cases. cc @alygin

alygin:
The scope of Variable Naming inspection is narrowed to local `let ...` statements. Bindings within `match`es and `if let`s  are ignored.

Fixes #730

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
@@ -254,7 +254,8 @@ class RustVariableNamingInspection : RustSnakeCaseNamingInspection("Variable") {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
         object : RustElementVisitor() {
             override fun visitPatBinding(el: RustPatBindingElement) {
-                if (el.parent?.parent !is RustParameterElementImpl) {
+                val parent = el.parent?.parent
+                if (parent is RustLetDeclElement || parent is RustPatTupElement) {
                     inspect(el.identifier, holder)
                 }
             }
--- a/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
@@ -3,6 +3,7 @@ package org.rust.ide.inspections
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiNamedElement
+import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.inspections.fixes.RenameFix
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.RustParameterElementImpl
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
```diff
@@ -254,7 +254,8 @@ class RustVariableNamingInspection : RustSnakeCaseNamingInspection("Variable") {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
         object : RustElementVisitor() {
             override fun visitPatBinding(el: RustPatBindingElement) {
-                if (el.parent?.parent !is RustParameterElementImpl) {
+                val parent = el.parent?.parent
+                if (parent is RustLetDeclElement || parent is RustPatTupElement) {
```
matklad: I think this is not exactly what we want here, consider these two examples:

```Rust
struct S { f: i32 }

let S { f: FALSE_NEGATIVE } = S { f: 92 };

match (a, b) => {
    (None, Some(x)) => "false positive"
} 
```

Ideally we want something more like

```Rust
val pattern = PsiTreeUtil.getTopmostParentOfType(el, RustPatElement::class.java) ?: return
when (pattern.parent) {
    is RustLetDeclElement -> inspect(el.identifier, holder)
}
```

I think it's a high time to introduce 

```
val RustPatBindingElement.rootPattern: RustPatElement?
```

extension. We need something similar over here: https://github.com/intellij-rust/intellij-rust/blob/6a78cae5f629d45608bd9dc79af6ee1605c51a72/src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustTypificationEngine.kt#L229

alygin: The solution with `PsiTreeUtil.getTopmostParentOfType()` surely works better. Though, I haven't yet gain a good enough understanding of the PSI tree structure to grasp what's happening inside this function or what `rootPattern` should do.

matklad: >r what rootPattern should do.

rootPattern should return the topmost pattern this particular binding belongs to. 

If you have something like `let (S { x }, y) = unimplemented!()` then the binding `x` is inside the pattern `S { x }` which is inside the (root) pattern `( S { x }, y)`. Not sure that "root" is the best term here though. 

The idea is that by examining the parent of the root pattern you should be able to understand the context where this bindig occurs (is it a let statement, match expression, function parameter or something else).

matklad:
Perfect, thanks!

PR #739, opened 2016-10-14
(FMT): decouple where and ret type alignment

matklad:
@mkaput please review. 

This makes different options for alignment of ret type and where clause, and change the default. 
Returns are aligned, where clauses are not (real world example: https://github.com/rust-lang/cargo/blob/5f2cc15de7cf51a362fb8320d8e81e6da67f1aab/src/cargo/core/shell.rs#L217-L223)


matklad:
Hm, I wonder why the formatting and the autoindent disagree with each other?

With "align where clauses" on, "ret type" off, **Reformat code** gives

```
fn foo<T>(t: T)
    -> T
          where T: Clone {}
```

but if I **Auto indent line**, I get


```
fn foo<T>(t: T)
    -> T
    where T: Clone {}
```

mkaput:
They're different routines. You have to customise `getChildrenAttributes`

PR #742, opened 2016-10-15
(INSP): Deprecated API Usage inspection

alygin:
The inspection detects usage of deprecated modules, functions, structs, consts, traits etc. Corresponds to Rust's `deprecated` and `rustc_deprecated` lints with a couple of small enhancements:
- It tells what's the exact item that is deprecated (for instance, if a module is deprecated and one is calling a function from this module, the inspection displays "Module `foo` is deprecated", while `rustc` uses abstract "use of deprecated item")
- In addition to `note` or `reason`, it displays information from the `since` meta item (for instance, "Function `sleep_ms` is deprecated since 1.6.0: replaced by `std::thread::sleep`")

The implementation contains a couple of functions that could be generalized and placed in some utility class or spread between classes:
- `PsiElement.getTitle()` returns the name of an element along with its type. For instance, "Function `foo`", "Enum variant `Bar`" etc. It might be handy in inspections, intentions and maybe somewhere else.
- `PsiElement.findDeprecation()` is searching for the deprecation attributes. It also might be generalized. We surely need some utility function for searching for attributes in the PSI tree. For instance, many inspections need them to allow automatic suppression. Even this narrowly oriented function could be used to mark (strikethrough) elements as deprecated in the project tree and in the completion drop down list.

What do you think? What would be the right place for these methods?


matklad:
>We surely need some utility function for searching for attributes in the PSI tree.

We have one! It's pretty bare bones at the moment, but can be extended: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/RustDocAndAttributeOwner.kt#L26

src/main/kotlin/org/rust/ide/inspections/RustDeprecatedInspection.kt
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Represents information about deprecated element.
+ */
+private class DeprecationInfo(
+    val element: PsiElement,
+    attr: RustAttrElement
+) {
+    var since: String? = null
+    var note: String? = null
+    var reason: String? = null
```
matklad: Can we use `val`s here? 

You'll need three local variables for this in `init`, which is a bit more code, but immutability is a great thing!

alygin: Sure

alygin:
Great! I'll take a look at it.

matklad:
>PsiElement.getTitle()

There's `getPresentation` method which should handle this: https://github.com/intellij-rust/intellij-rust/blob/59f34a592859d7c217062e92cd1543d7898939ac/src/main/kotlin/org/rust/lang/core/psi/impl/RustStubbedNamedElementImpl.kt#L41. Not sure that `RustStubbedNamedElementImpl` is the right place for it though. Perhaps we should use `PresentationProvider` to avoid complicated inheritance mess in PSI.

src/main/kotlin/org/rust/ide/inspections/RustDeprecatedInspection.kt
```diff
@@ -0,0 +1,106 @@
...
+    private fun PsiElement.findDeprecation(): DeprecationInfo? {
+        var attr: RustAttrElement? = null
+        if (this is RustOuterAttributeOwner) {
+            attr = findOuterAttr("deprecated") ?: findOuterAttr("rustc_deprecated")
+        }
+        if (attr == null && this is RustInnerAttributeOwner) {
+            attr = findInnerAttr("deprecated") ?: findInnerAttr("rustc_deprecated")
+        }
+        if (attr == null && parent != null) {
+            return parent.findDeprecation()
```
alygin: Could you, please, clarify on this part? For me, the main purpose of this inspection is to warn the developer about deprecations in other modules and crates (including the standard library), not in those they are working on. Limiting this search by a module will ruin the whole idea.

If the traversal looks costly, I could try to find a way to make this inspection "slow". I think I saw something like that in other plugins.

matklad: >Could you, please, clarify on this part? 

Sure! If module `m` is deprecated, and you use a function `foo` from the module, then you either:

* use a full path to the item: `m::foo`,

* import an item from a module `use m::{foo, bar}`.

That is, if you have a reference to a function, you must have a reference to containing module as well, so it makes sense to warn about the module itself. 

Also keep in mind that the following code compiles without warnings in rutsc:

```
#[deprecated]
pub mod m {
    pub fn foo() {}  
    pub fn bar() { 
        foo() // no deprecation warning here from rustc
    }
}
```

alygin: The first part now works as you described, it strikes out all elements from the deprecated module and adds a hint that the module is deprecated. But I didn't notice that `rustc` doesn't show deprecations within the deprecated module, thanks for pointing that out.

I think I need to improve the inspection, till then I close the PR. Anyway, thanks for helping!

PR #745, opened 2016-10-16
(MISC): Line markers for trait method implementations

alygin:
Trait methods implementations are annotated with icons on the gutter. Icons are linked to the trait method declarations. Pure implementations and overridings feature different icons:

screen shot 2016-10-16 at 22 35 43

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustTraitMethodImplLineMarkerProvider.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustTraitMethodImplLineMarkerProvider.kt
@@ -4,11 +4,11 @@ import com.intellij.codeInsight.daemon.RelatedItemLineMarkerInfo
 import com.intellij.codeInsight.daemon.RelatedItemLineMarkerProvider
 import com.intellij.codeInsight.navigation.NavigationGutterIconBuilder
 import com.intellij.psi.PsiElement
-import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.icons.RustIcons
 import org.rust.lang.core.psi.RustImplItemElement
 import org.rust.lang.core.psi.RustImplMethodMemberElement
 import org.rust.lang.core.psi.RustTraitItemElement
+import org.rust.lang.core.psi.util.parentOfType
 import javax.swing.Icon
 
 /**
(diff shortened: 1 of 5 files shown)


alygin:
There's a small flaw in the implementation &ndash; when looking for the method declarations, it only checks their names and ignores parameters. I'm planning to get back to it a bit later. But since Rust doesn't allow method overloading, it shouldn't be a significant problem.

src/main/kotlin/org/rust/ide/annotator/RustTraitMethodImplLineMarkerProvider.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.rust.lang.core.psi.RustTraitItemElement
+import javax.swing.Icon
+
+/**
+ * Annotates the implementation of a trait method with an icon on the gutter.
+ */
+class RustTraitMethodImplLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun collectNavigationMarkers(el: PsiElement, result: MutableCollection<in RelatedItemLineMarkerInfo<PsiElement>>?) {
+        if (result == null || el !is RustImplMethodMemberElement) return
+        val implBlock = PsiTreeUtil.getParentOfType(el, RustImplItemElement::class.java) ?: return
```
matklad: Why not `el.parentOfType<RustImplItemElement>()`?

src/main/kotlin/org/rust/ide/annotator/RustTraitMethodImplLineMarkerProvider.kt
```diff
@@ -0,0 +1,43 @@
...
+        val trait = implBlock.traitRef?.path?.reference?.resolve() ?: return
+        if (trait !is RustTraitItemElement) return
+        for (traitMethod in trait.traitMethodMemberList) {
```
matklad: This can be written as `trait.traitMethodMemberList.find { it.name == el.name } ? return`

src/main/kotlin/org/rust/ide/annotator/RustTraitMethodImplLineMarkerProvider.kt
```diff
@@ -0,0 +1,43 @@
...
+            if (traitMethod.name == el.name) {
+                val action: String
+                val icon: Icon
+                if (traitMethod.block == null) {
```
matklad: I think there was `isAbstract` extension property somewhere.

alygin:
Thanks, it looks simpler after refactoring.

matklad:
This is awesome, thanks!

matklad:
Hey, turns out we already have a `superMethod` extension: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustImplMethodMemberImplMixin.kt#L33

I've used it here in https://github.com/intellij-rust/intellij-rust/commit/89119c5145f08cad628365c7e0bbbc11ea30b68f

alygin:
Oh, good to know. Thanks for the refactoring!

Issue #630, opened 2016-08-25
Support Go To and Quick Doc on stdlib macros


johnthagen:
It would be really helpful to those new to Rust if Go To and Quick Doc were supported for the standard macros (e.g. `println!`)

``` rust
println!("Hello, World.")  // Add Ctrl-Q and Ctrl-Click to println!
```

In this example the Quick Docs would show: https://doc.rust-lang.org/src/std/up/src/libstd/macros.rs.html#101-115

PR #749, opened 2016-10-17
add more macroses

HeyLey:
add vec! and panic! macroses + tests


matklad:
`RustParserPerformanceTest` is failing. This tests feeds the source code of the compiler itself to our parser and checks that there are no parsing errors. It is not run during `./gradlew test` because it is a longish test. Instead, you can run `./gradlew performanceTest`, or execute it from the IDE. 

The problem is with an empty vec macro: `vec![]`.

johnthagen:
Curious, will this support Go To and Quick Doc for stdlib macros? (#630)

matklad:
Thanks!

PR #750, opened 2016-10-19
(MISC): Inspections refactoring

alygin:
No functional changes involved. Just minor refactoring, simplification and bringing inspections to a a common style.


matklad:
Looks better now, thanks!

PR #753, opened 2016-10-20
(MISC): Go to trait impl gutter icon

alygin:
A `LineMarkerProvider` that adds an icon to a trait declaration that allows to jump to the trait implementations.

screen shot 2016-10-20 at 22 31 35

There's a problem that if a trait has a rustdoc comment and/or attributes, the icon is placed at the beginning of the comment/attribute, not at the trait declaration line. I can try to find a workaround in this marker provider, but I think it would be better (and maybe even easier) to solve the problem in the parser, though I don't know how to do that.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustTraitLineMarkerProvider.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustTraitLineMarkerProvider.kt
@@ -1,0 +1,34 @@
+package org.rust.ide.annotator
+
+import com.intellij.codeInsight.daemon.RelatedItemLineMarkerInfo
+import com.intellij.codeInsight.daemon.RelatedItemLineMarkerProvider
+import com.intellij.codeInsight.navigation.NavigationGutterIconBuilder
+import com.intellij.psi.PsiElement
+import com.intellij.psi.search.searches.ReferencesSearch
+import org.rust.ide.icons.RustIcons
+import org.rust.lang.core.psi.RustImplItemElement
+import org.rust.lang.core.psi.RustTraitItemElement
+
+/**
+ * Annotates trait declaration with an icon on the gutter that allows to jump to
+ * its implementations.
+ */
+class RustTraitLineMarkerProvider : RelatedItemLineMarkerProvider() {
+
+    override fun collectNavigationMarkers(el: PsiElement, result: MutableCollection<in RelatedItemLineMarkerInfo<PsiElement>>?) {
+        if (result == null || el !is RustTraitItemElement) return
+        val targets = ReferencesSearch.search(el, el.useScope)
+            .map { it.element.parent?.parent }
+            .filter { it is RustImplItemElement && it.type != null }
+            .toList()
+
+        if (!targets.isEmpty()) {
+            val builder = NavigationGutterIconBuilder
+                .create(RustIcons.IMPLEMENTED)
+                .setTargets(targets)
+                .setPopupTitle("Go to implementation")
+                .setTooltipText("Has implementations")
+            result.add(builder.createLineMarkerInfo(el))
+        }
+    }
+}
--- a/src/main/kotlin/org/rust/ide/annotator/RustTraitLineMarkerProvider.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustTraitLineMarkerProvider.kt
@@ -1,7 +1,7 @@
 package org.rust.ide.annotator
 
-import com.intellij.codeInsight.daemon.RelatedItemLineMarkerInfo
-import com.intellij.codeInsight.daemon.RelatedItemLineMarkerProvider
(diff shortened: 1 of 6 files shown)


matklad:
>There's a problem that if a trait has a rustdoc comment and/or attributes, the icon is placed at the beginning of the comment/attribute, not at the trait declaration line.

I believe this is correct that we include documentation comments and attributes into the PSI node for an item. Kotlin does the same:

psi

There should be some reasonable way to customize the icon's position.

src/main/kotlin/org/rust/ide/annotator/RustTraitLineMarkerProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.rust.lang.core.psi.RustImplItemElement
+import org.rust.lang.core.psi.RustTraitItemElement
+
+/**
+ * Annotates trait declaration with an icon on the gutter that allows to jump to
+ * its implementations.
+ */
+class RustTraitLineMarkerProvider : RelatedItemLineMarkerProvider() {
+
+    override fun collectNavigationMarkers(el: PsiElement, result: MutableCollection<in RelatedItemLineMarkerInfo<PsiElement>>?) {
```
alygin: Yeah, it isn't indeed. I'll remove the `?`.

alygin:
Sure, there's a way to do that. I'll replace the icon.

matklad:
I think this can also be plugged into `GotoImplementationHandler` to provide **Goto implementation** action.

matklad:
Kotlin does something like this:

```
    val anchor = element.nameIdentifier ?: element

    result.add(LineMarkerInfo(
            anchor,
            anchor.textOffset,
            if (isTrait) IMPLEMENTED_MARK else OVERRIDDEN_MARK,
            Pass.UPDATE_OVERRIDEN_MARKERS,
            SUBCLASSED_CLASS.tooltip,
            SUBCLASSED_CLASS.navigationHandler
    ))
```

matklad:
And perhaps we should use `collectSlowLineMarkers` because reference search is not exactly a fast operation

alygin:
Actually, I started working on this feature with the intention to provide trait declarations with both the gutter icon and "Go to implementation" action to close #570. But after studying the [tutorial](http://www.jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/references_and_resolve.html), the code behind the `GotoImplementationHandler` class and a couple of plugins (Go and Scala) I came to the conclusion that IntelliJ provides both "Go to declaration" and "Go to implementation" actions automatically if the PSI tree nodes properly implement references and the `PsiElement.processDeclarations()` method.

Unfortunately, I don't understand all the details of the current PSI tree design good enough to properly implement `processDeclarations()` where it's needed. So, I decided to go with the solution based on references analysis and then study other plugins and the IntelliJ Rust PSI tree utilities to be able to come up with the final implementation.

So, I can make the line marker use the `collectSlow...` version and fix the icon position quickly. But it will take me some time to come up with the final and full-scale implementation. Though, I'm not even sure that the idea about `processDeclarations()` is correct )

matklad:
>So, I can make the line marker use the collectSlow... version and fix the icon position quickly. 

Yep, let's do this, I want to merge this ASAP, because it is awesome :) We can always add a polish later.

matklad:
Travis is down, so tested locally, and it works like a charm!

alygin:
Thanks for merging!

matklad:
>IntelliJ provides both "Go to declaration" and "Go to implementation" actions automatically if the PSI tree nodes properly implement references and the PsiElement.processDeclarations() method.

Not exactly. "Go to declaration" will work as soon as `getReferece` is implemented on the PSI elements, and `resolve` is implemented on the references. This is a core functionality of every language, and it works similarly everywhere. 

Stuff like "Go to super", "Go to implementation", etc needs some language specific tweaking though, because, at least, you need to know on which refernces and named elements a particular search is available: you don't want to search for implementations of a local variable. 

I believe you need to implement [DefinitionsScopeSearcher](https://github.com/JetBrains/kotlin/blob/3471b3311f8358a14ab21d957c6137a84615484a/idea/src/org/jetbrains/kotlin/idea/search/ideaExtensions/KotlinDefinitionsSearcher.java) to search for implementations. 

Regarding `processDeclarations` method, it is just one of the possible way to implement `resolve` method on the references. Basically this is an internal iterator of the names available in scope. 

We use an external style iterator instead, based on the Kotlin `Sequecne` class: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L214

Though, I am not really happy with our current resolve infrastructure, because it can break for some particularly tricky combination of start imports. Rustc uses fixed point iteration for the import resolution, but we can't use this approach because it is not lazy.

alygin:
Thanks for the explanation and for the links! I'll have a look at how it works in Kotlin.

PR #755, opened 2016-10-21
(INT) Add else branch intention

ujpv:
What do you think, is it useful intention or it's too elementary action for intention?

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/AddElseIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/AddElseIntention.kt
@@ -1,0 +1,36 @@
+package org.rust.ide.intentions
+
+import com.intellij.codeInsight.intention.PsiElementBaseIntentionAction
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/intentions/AddElseIntention.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    private data class Context(
+        val blockExpr: RustIfExprElement
+    )
+
+    private fun findContext(element: PsiElement): Context? {
+        if (!element.isWritable) return null
+
+        val ifStmnt = element.parentOfType<RustIfExprElement>() as? RustIfExprElement ?: return null
+        return if (ifStmnt.elseBranch == null) {
```
matklad: I would remove the `{}` braces here and just put this on the single line

matklad:
I thing it's nice!

PR #756, opened 2016-10-21
(MISC): Recursive calls gutter marker

alygin:
An annotator that detects and marks recursive calls:

screen shot 2016-10-21 at 17 32 59

Works for pure functions, associated functions and methods.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustRecursiveCallLineMarkerProvider.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustRecursiveCallLineMarkerProvider.kt
@@ -39,7 +39,7 @@ class RustRecursiveCallLineMarkerProvider : LineMarkerProvider {
                         el,
                         el.textRange,
                         RustIcons.RECURSIVE_CALL,
-                        Pass.UPDATE_OVERRIDDEN_MARKERS,
+                        Pass.UPDATE_OVERRIDEN_MARKERS,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/RustRecursiveCallLineMarkerProvider.kt
```diff
@@ -0,0 +1,64 @@
...
+                        GutterIconRenderer.Alignment.RIGHT))
+                    lines.add(lineNumber)
+                }
+            }
+        }
+    }
+
+    private fun RustCallExprElement.pathExpr() : RustPathExprElement? {
+        val e = expr
+        return if (e is RustPathExprElement) e else null
```
matklad: Can be `expr as? RustPathExprElement`.

src/main/kotlin/org/rust/ide/annotator/RustRecursiveCallLineMarkerProvider.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    private fun RustExprElement.isRecursive(ref: RustReference?): Boolean {
+        if (ref == null) return false
+        val def = ref.resolve() ?: return false
+        return (Comparing.equal(parentOfType<RustImplMethodMemberElement>(), def))  // Methods and associative functions
```
matklad: You can use `==` here, it is null safe.

PR #757, opened 2016-10-21
(SURR): assert! and dubug_assert! surrounders

ujpv:


mkaput:
Personally I don't think this functionality will gain any usage

ujpv:
@mkaput 
I have thought a little. You are right.
I think this will be useful as a template postfix.

PR #760, opened 2016-10-26
(SURR): Correct postprocess

ujpv:
Fix syntax error, which appears immediately after surrounder use.

### DIFF
--- a/src/test/kotlin/org/rust/ide/template/postfix/IfPostfixTemplateTest.kt
+++ b/src/test/kotlin/org/rust/ide/template/postfix/IfPostfixTemplateTest.kt
@@ -32,7 +32,7 @@ class IfPostfixTemplateTest : PostfixTemplateTestCase(IfExpressionPostfixTemplat
                 a.if<caret>
             }
             """
-        ,
+            ,
             """
             fn main() {
                 let a = 4 == 2;
@@ -64,7 +64,7 @@ class IfPostfixTemplateTest : PostfixTemplateTestCase(IfExpressionPostfixTemplat
                 !a.if<caret>
             }
             """
-        ,
+            ,
             """
             fn main() {
                 let a = 4 == 2;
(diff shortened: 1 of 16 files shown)


matklad:
Nice! Let's add a test here, to make sure this does not regress and to make it clear in the git history what is the purpose of these changes!

The test should look like this:

```
fun testCorrectPostProcess() = doTest(
    """
    fn main() {
        true.if<caret>
        1;
    }
    """
    ,
    """
    fn main() {
        if true {}
        1;
    }
    """
)
```

To check for the absence of errors,  you'll need to modify

```
    protected fun doTest(@Language("Rust") before: String, @Language("Rust") after: String) {
        myFixture.configureByText(RustFileType, before)
        checkApplicability(before, true)
        myFixture.type('\t')
        myFixture.checkHighlighting(false, false, false) // add this line
        myFixture.checkResult(after)
    }
```

ujpv:
I added tests.
But it is a bit more complicated. Not always code after action is syntactically correct.
For example 

```Rust
    fn main() {
        <selection>println!("Hellio");<selection>
    }
```
Surround with while {} gives
```Rust
    fn main() {
        while <caret> {println!("Hellio");}
    }
```

matklad:
Indeed! But in most of the cases it is correct? Then perhaps we should use ` checkSyntaxErrors = true` as the default?

ujpv:
Of course.

ujpv:
Done.

src/test/kotlin/org/rust/ide/template/postfix/IfPostfixTemplateTest.kt
```diff
@@ -110,6 +110,8 @@ class IfPostfixTemplateTest : PostfixTemplateTestCase(IfExpressionPostfixTemplat
                 }
             }
             """
+            ,
+            checkSyntaxErrors = false
```
matklad: This should actually be `true` :) There is a `;` missing after let. The test should be

```
             fn main() {
                 let a = if (true) {
                     true == false.if<caret>
                 } else {
                     false == true
                  }; // <-
              }
```

ujpv:
Yes my fault.
Fixed.

matklad:
CI fails...

matklad:
Great!

PR #761, opened 2016-10-26
(MISC) Fix gutter action for test functions in the main module

mrobakowski:
When running test function in main module from the gutter icon, incorrect action was taken (the module was run instead of the test). This fixes that.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/producers/CargoExecutableRunConfigurationProducer.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/producers/CargoExecutableRunConfigurationProducer.kt
@@ -11,6 +11,7 @@ import org.rust.cargo.runconfig.CargoCommandConfiguration
 import org.rust.cargo.runconfig.CargoCommandRunConfigurationType
 import org.rust.cargo.util.cargoProject
 import org.rust.lang.core.psi.RustFnElement
+import org.rust.lang.core.psi.RustFnItemElement
 import org.rust.lang.core.psi.util.module
 import org.rust.lang.core.psi.util.parentOfType
 
@@ -20,6 +21,7 @@ class CargoExecutableRunConfigurationProducer : RunConfigurationProducer<CargoCo
         configuration: CargoCommandConfiguration,
         context: ConfigurationContext
     ): Boolean {
+        if (isTestFn(context.psiLocation)) return false
         val target = findBinaryTarget(context) ?: return false
 
         return configuration.command == CargoConstants.Commands.RUN &&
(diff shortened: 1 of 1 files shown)


matklad:
Hm, quite probably something's fishy does happen with gutter icons, but I can't reproduce the issue you are referring to...

For the following main.rs file

```
#[test]
fn bar() {}

fn main() {}

#[test]
fn fooo() {}
```

an appropriate test function is run when I click the icon near bar or foo, and `main` is executed if I click the middle icon. 

What else is needed to reproduce the issue? 

Also, I think you can add a test for this to `RunConfigurationProducerTestCase`:

```
    fun testGutterIcon() {
        testProject {
            bin("foo", "src/main.rs", """
                fn main() { }
                #[test]
                fn <caret>test_foo() { }
            """).open()
        }

        for (mark in myFixture.findGuttersAtCaret()) {
            val event = TestActionEvent()
            val gutterIconRenderer = mark as GutterIconRenderer
            println(gutterIconRenderer.tooltipText)
            val actions = gutterIconRenderer.popupMenuActions!!.getChildren(event).mapNotNull { action ->
                val actionEvent = TestActionEvent()
                action.update(actionEvent)
                actionEvent.presentation.text
            }
            println("actions = ${actions}")
        }
    }
```

mrobakowski:
I should probably mention that I am using IntelliJ IDEA EAP 2016.2.4, build IU-162-2032-8, on a Windows 10 machine

https://gfycat.com/QuaintRightAfricanclawedfrog

As you can see, not only the action taken is wrong, also the texts are'Run ...' instead of 'Test ...'.

I tested it on another machine (also Windows), with slightly different IDEA version (2016.3), and yesterday I could, but today I cannot reproduce the bug (the texts are still bugged, but at least the action taken is correct).

I will try to run the test you wrote on both machines, and will report back soon.

mrobakowski:
Okay, I've run the test you provided on the machine I could reproduce the bug on and...

```
[   3546]   WARN - impl.local.LocalFileSystemBase - Invalid path: light_idea_test_case.iml 
[   3651]   WARN - impl.local.LocalFileSystemBase - Invalid path: light_idea_test_case.iml 
[   3658]   WARN - impl.local.LocalFileSystemBase - Invalid path: light_idea_test_case.iml 
[   3751]   WARN - impl.local.LocalFileSystemBase - Invalid path: light_idea_test_case.iml 
[   5519]   WARN - impl.local.LocalFileSystemBase - Invalid path: light_idea_test_case.iml 
Run Test
actions = [Run 'Test test_foo', Debug 'Test test_foo']

```

Yet the bug still remains (both on my instance of IDEA, and the sandboxed one from gradle runIdea)

matklad:
Thanks for investigating this! I've pushed a one-character fix ( I hope :) ) other there: https://github.com/intellij-rust/intellij-rust/commit/52380a8cd6410a051ce38b7f3608c8fe9c765ac8

Issue #694, opened 2016-09-27
Intention: Move type constraint to the where clause


matklad:
``` Rust
fn foo<T: Bar>() { }
```

->

``` Rust
fn foo<T>() where T: Bar { }
```

mkaput:
Vice versa could also be nice

matklad:
Fixed by #762

PR #762, opened 2016-10-27
(INT): Add constraint to where clause intention

alobb:
This implements (part of) the functionality for #694.  It adds a new intention that moves type constraints to a where clause for functions (currently only works on functions).  I've never written Kotlin before or worked with custom language plugins for Intellij, so let me know if anything looks wrong/un-idiomatic!  Also, would it make sense to get this working with structs/traits/enums on this PR, or a separate one?


alobb:
Something I noticed in some (but not all) of the other inspections is that they reformat the code after inserting a PSI Element; should I do that here as well? What determines when this should be done?

src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToWhereClauseIntention.kt
```diff
@@ -0,0 +1,36 @@
...
+class MoveTypeConstraintToWhereClauseIntention : PsiElementBaseIntentionAction() {
+    override fun getText() = "Move type constraint to where clause"
+    override fun getFamilyName() = text
+    override fun startInWriteAction() = true
+
+    override fun isAvailable(project: Project, editor: Editor, element: PsiElement): Boolean {
+        if (!element.isWritable) return false
+
+        val function = element.parentOfType<RustFnItemElement>() ?: return false
+        val genericTypeBounds = function.genericParams?.typeParamList?.filter { it.typeParamBounds != null }.orEmpty()
```
matklad: There's `filterNotNull`.

src/main/kotlin/org/rust/lang/core/psi/RustElementFactory.kt
```diff
@@ -42,6 +42,17 @@ object RustElementFactory {
     fun createIdentifier(project: Project, name: String): PsiElement =
         createFromText<RustModDeclItemElement>(project, "mod $name;")!!.identifier
 
+    fun createWhereClause(project: Project, lifetimeBounds: List<RustLifetimeParamElement>, typeBounds: List<RustTypeParamElement>): RustWhereClauseElement? {
+        val lifetimes = lifetimeBounds
+            .filter { it.lifetimeParamBounds != null }
+            .mapNotNull { it.node.text }
```
matklad: I believe you can use a single `.mapNotNull { it.lifetimeParamBouds?.node?.text }` here.

matklad:
>What determines when this should be done?

It's always safe to reformat code after refactorings, although if changes are local (like in this case) you can get away without reformatting.

matklad:
Thanks!

Issue #734, opened 2016-10-12
Add function parameter color control


johnthagen:
It would be neat to be able to control the color of parameters in `intellij-rust` similar to how it is done in PyCharm.

![pycharm_parameter_color](https://cloud.githubusercontent.com/assets/10340167/19319714/24d14a7a-907c-11e6-94b1-c4927ed7c548.png)

johnthagen:
Fixed in #763

PR #763, opened 2016-10-29
(MISC): Add function parameter color control

alobb:
This adds parameters to the color settings as requested in #734. I split `self` params from the rest (similar to the screenshot in #734), but I'm not sure if that is what the project wants, so please give input if this PR seems misguided.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
@@ -96,5 +96,8 @@ class RustHighlightingAnnotator : Annotator {
         override fun visitTraitMethodMember(o: RustTraitMethodMemberElement) =
             highlight(o.identifier, if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD)
 
+        override fun visitParameters(o: RustParametersElement) = o.parameterList.forEach { highlight(it.pat, RustColor.PARAMETER) }
+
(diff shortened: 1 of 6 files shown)


matklad:
LGTM! Will get back to this after cutting this week's alpha!

src/main/kotlin/org/rust/ide/annotator/RustHighlightingAnnotator.kt
```diff
@@ -96,5 +96,8 @@ class RustHighlightingAnnotator : Annotator {
         override fun visitTraitMethodMember(o: RustTraitMethodMemberElement) =
             highlight(o.identifier, if (o.isStatic) RustColor.STATIC_METHOD else RustColor.INSTANCE_METHOD)
 
+        override fun visitParameters(o: RustParametersElement) = o.parameterList.forEach { highlight(it.pat, RustColor.PARAMETER) }
```
matklad: I think this is not exactly what we want here, because a single pattern can introduce multiple bindings, each of which should be colored separately: `fn foo((x, y): (i32, i32))`. Perhaps we can do something like

```
when {
    o.parentOfType<RustParameterElement>() != null ->
        highlight(o, RustColor.PARAMETER)

    o.isMut ->
        highlight(o.identifier, RustColor.MUT_BINDING)
}
```

?

That way, the usages of function parameters will be colored as well!

matklad:
I've pushed it over the finish line myself, thanks for the PR!

alobb:
My apologies for leaving this hanging, school was busy the past week.

PR #767, opened 2016-11-05
Add reference to TOML plugin

johnthagen:
Help new users find the related TOML plugin for `Cargo.toml` and `Cargo.lock` files.


matklad:
Thanks!

PR #770, opened 2016-11-10
(TEMPL) Postfix asserts, match, ()

ujpv:
Several templates postfix.


ujpv:
I'll make path resolving for match template.
For now I close request.

src/main/kotlin/org/rust/ide/template/postfix/utils.kt
```diff
@@ -41,3 +42,30 @@ class RustTopMostInScopeSelector(pred: (RustExprElement) -> Boolean) : RustExprP
...
+
+    override fun hasExpression(context: PsiElement, copyDocument: Document, newOffset: Int): Boolean =
+        context
+            .ancestors
+            .takeWhile { it !is RustBlockElement }
+            .any { it is RustExprElement && pred(it) }
+}
+
+
+internal fun RustExprElement.isBool() = resolvedType == RustBooleanType
```
matklad: Unfortunately `internal` modifier is not really useful for us, because the whole project consists of a single module.

PR #776, opened 2016-11-13
(TEMPL) Postfix asserts, match, ()

ujpv:
I have a little changed RustPsiImplUtil.
Please, review it.

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
@@ -1,0 +1,76 @@
+package org.rust.ide.template.postfix
+
+import com.intellij.codeInsight.template.Template
+import com.intellij.codeInsight.template.impl.TextExpression
+import com.intellij.codeInsight.template.postfix.templates.StringBasedPostfixTemplate
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustBinaryExprElement
+import org.rust.lang.core.psi.RustExprElement
+import org.rust.lang.core.psi.RustTupleFieldDeclElement
+import org.rust.lang.core.psi.impl.RustPsiImplUtil
+import org.rust.lang.core.psi.util.descendentsOfType
+import org.rust.lang.core.psi.util.parentOfType
+import org.rust.lang.core.resolve.RustResolveEngine
+import org.rust.lang.core.resolve.scope.RustResolveScope
+import org.rust.lang.core.types.RustEnumType
+import org.rust.lang.core.types.util.resolvedType
+
+class MatchPostfixTemplate : StringBasedPostfixTemplate(
+    "match",
+    "match expr {...}",
+    RustTopMostInScopeSelector(RustExprElement::isEnum)
+) {
+    override fun getTemplateString(element: PsiElement): String? {
+        val enumType = (element as RustExprElement).resolvedType as RustEnumType
+
+        val allAvailableNames = generateSequence(
+            element.parentOfType<RustResolveScope>(),
+            { it.parentOfType<RustResolveScope>() }
+        )
+            .flatMap { RustResolveEngine.declarations(it, element) }
+            .mapNotNull { it.element }
+            .mapNotNull { RustPsiImplUtil.canonicalCratePath(it) }
+            .map { it.toString() }
+            .toSet()
+
+        val stringBuilder = StringBuilder()
+        stringBuilder.append("match ${element.text} {\n")
+
+        for ((i, item) in enumType.item.enumBody.enumVariantList.withIndex()) {
+            val itemCanonicalName = RustPsiImplUtil.canonicalCratePath(item).toString()
+
+            val excessPath = allAvailableNames
+                .filter { it.commonPrefixWith(itemCanonicalName).length == it.length }
+                .maxBy { it.length }
+
+            val itemShortName = itemCanonicalName.substring(excessPath?.lastIndexOf("::")?.plus(2) ?: 2)
+
+            val tupleOrStructFields = when {
+                item.blockFields != null -> item.blockFields!!.fieldDeclList
+                    .map { it.identifier.text }
+                    .joinToString(prefix = "{", separator = ", ", postfix = "}")
+
+                item.tupleFields != null -> (0 until item.tupleFields!!.descendentsOfType<RustTupleFieldDeclElement>().size)
+                    .map { "v$it" }
+                    .joinToString(prefix = "(", separator = ", ", postfix = ")")
+
+                else -> ""
+            }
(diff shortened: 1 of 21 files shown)


src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -0,0 +1,136 @@
...
+    "assert!(expr);",
+    RustTopMostInScopeSelector(RustExprElement::isBool)
+) {
+    override fun getTemplateString(element: PsiElement): String? =
+        "assert!(\$expr$);\$END$"
+
+    override fun getElementToRemove(expr: PsiElement): PsiElement = expr
+}
+
+class AssertEqPostfixTemplate : StringBasedPostfixTemplate(
```
matklad: Can we make a single intelligent template for both `Assert` and `AssertEq`?

`foo && bar.assert` => `assert!(foo && bar)`

`foo == bar.assert` => `assert_eq!(foo, bar)`

?

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -0,0 +1,136 @@
...
+        val allAvailableNames = generateSequence(
+            element.parentOfType<RustResolveScope>(),
+            { it.parentOfType<RustResolveScope>() }
+        )
+            .flatMap { RustResolveEngine.declarations(it, element) }
+            .filter { it.element != null }
+            .map { RustPsiImplUtil.canonicalCratePath(it.element!!) }
+            .filterNotNull()
+            .map { it.toString() }
+            .toSet()
```
matklad: This can be simplified a bit.

`filter { it.element != null}` => `mapNotNull { it.element }`

`map { ... }. filterNotNull()` => `mapNotNull { ... }`

src/main/kotlin/org/rust/ide/template/postfix/utils.kt
```diff
@@ -41,3 +42,30 @@ class RustTopMostInScopeSelector(pred: (RustExprElement) -> Boolean) : RustExprP
...
+        context
+            .ancestors
+            .takeWhile { it !is RustBlockElement }
+            .any { it is RustExprElement && pred(it) }
+}
+
+
+internal fun RustExprElement.isBool() = resolvedType == RustBooleanType
+internal fun RustExprElement.isEnum() = resolvedType is RustEnumType
+internal fun RustExprElement.all() = true
```
matklad: Looks like an unused function?

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
ujpv: Yes, we can.
I thought before, it is worth making so or not.
Probably, yes.

ujpv:
I removed asserts and parenthesis templates from this PR and fixed issues.

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        val stringBuilder = StringBuilder()
+        stringBuilder.append("match ${element.text} {\n")
+
+        for ((i, item) in enumType.item.enumBody.enumVariantList.withIndex()) {
+            val itemCanonicalName = RustPsiImplUtil.canonicalCratePath(item).toString()
+
+            val excessPath = allAvailableNames
```
matklad: This is a costly computation which can be hoisted out of the loop. 

But I don't think that the logic here is 100% correct. `allAvailableNames` will give you absolute names, which are not necessary the name you want in local scope. I'd keep it simple first and just use the name of the enum itself.

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -0,0 +1,76 @@
...
+                .filter { it.commonPrefixWith(itemCanonicalName).length == it.length }
+                .maxBy { it.length }
+
+            val itemShortName = itemCanonicalName.substring(excessPath?.lastIndexOf("::")?.plus(2) ?: 2)
+
+            val tupleOrStructFields = when {
+                item.blockFields != null -> item.blockFields!!.fieldDeclList
```
matklad: I would've introduced variables for `item.blockFields` and `item.tupleFIelds` to avoid `!!`.

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
ujpv: Yes, it's costly. But every enum variant name checks for cases like this:
```Rust
        enum MyOption<T> {
            None,
            Some(T)
        }
        fn main() {
            let a = MyOption::Some(42);
            use MyOption::None;
            match a {
                None => {},
                MyOption::Some(v0) => {},
            };
        }
```
Could you specify where there may be a problem with correctness?
1. We have enum variant. We can get the full path to it.
2. We can get all declarations, which available from the current scope.
3. We can get get absolute path for every declaration.
4. If some declaration from enum-variant path present in the current scope, when not necessary to specify full path to it. And we can drop some prefix of the path.

ujpv: I can try to add more tests :)

ujpv: I checked, the most heavy action in this method is a set creation. 
The list works better.
It was a strange choice of container.

matklad: >This is a costly computation which can be hoisted out of the loop.

Sorry, of course it can not be moved out of the loop because it uses `itemCanonicalName`. 

The problem with correctness can arise due to aliases: 

```
mod m {
    pub enum E {
        X
    }
}

use self::m::E as S;

fn main() {
    match S::X {
        S::X => {}
    }
}
```

But yes, adding a prefix when there's an imported variant in scope seems bad. 


What do you think about the following rule: "either all variants should be qualified by enum name, or none should be qualified". 

That, is the following are possible expansions:

```
match expr {
  Expr::Add() => {}
  Expr::Sub() => {}
}

match expr {
  Add() => {}
  Sub() => {}
}
```

and mixing or further qualifying is not allowed

```
match expr {
  Expr::Add() => {}
  Sub() => {}
}
match expr {
  ast::Expr::Add() => {}
  ast::Expr::Sub() => {}
}
```

?

That way, I think you can lookup only the first variant in the scope, which should allow to avoid the creation of list or set of names altogether. Basically, `val useShortNames = declarations.find {it.element == firstEnumVariant} != null` ?

ujpv:
I created #793 instead this.

PR #777, opened 2016-11-13
(INT): Match to if let

ujpv:
New intention.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
@@ -1,0 +1,54 @@
+package org.rust.ide.intentions
+
+import com.intellij.codeInsight.intention.PsiElementBaseIntentionAction
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.getNextNonCommentSibling
+import org.rust.lang.core.psi.util.parentOfType
+
+class MatchToIfLetIntention : PsiElementBaseIntentionAction() {
+    override fun getText() = "Convert math statement to if let"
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
```diff
@@ -0,0 +1,57 @@
...
+        if (notVoidArms.size == 1 &&
+            notVoidArms[0].matchPat.patList.size == 1 &&
+            notVoidArms[0].matchPat.patList[0].text != "_") {
+            return matchExpr
+        } else {
+            return null
+        }
+    }
+
+    fun RustExprElement.notVoid(): Boolean =
```
matklad: I think we want to treat `{}` and `()` the same. I'd write this as

```
val RustExprElement.isVoid: Boolean get() =
   (this is RustBlockExpeElement && (block?.lbrace.getNextNonCOmmentSIbling()? == block.rbrace)) ||
    this is RustUnitExpr
```

Patterns like

```
if (cond) false else true
```

can be simplified to just `!cond`.

src/test/kotlin/org/rust/ide/intentions/MatchToIfLetIntentionTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    fun testUnavailable2() = doUnavailableTest(
+        """
+        enum MyOption {
+            Nothing,
+            Some(x),
+        }
+
+        fn main() {/*caret*/
+            let a = MyOption::Some(52);
```
matklad: I think the caret should be inside a match expr to really test the case of non-void arms.

src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
ujpv: Thanks!
I have simplified it.

matklad:
LGTM!

Issue #731, opened 2016-10-12
Support for clickable rust errors


lilianmoraru:
Let's say you make a mistake and this message appears:
`error[E0308]: mismatched types`

It would be nice to be able to click in the output(or if it is not possible, than maybe present the option inside the IDE Window somewhere) on the error.
So, when clicking on "E0308", the plugin would sent the user to `https://doc.rust-lang.org/error-index.html#` + the error.
In this case it would look like this: `https://doc.rust-lang.org/error-index.html#E0308`.

Even better would be if it could take that part from the web and present it inside the IDE.

matklad:
Ah, we kinda have this functionality already,  but it has bitrotten and no longer works with the new error format. 

Here it is: https://github.com/intellij-rust/intellij-rust/blob/aea198e1893dc73d7ecf0fafc6f8f714951681e8/src/main/kotlin/org/rust/cargo/runconfig/RustExplainFilter.kt

And I am a bit hesitant to fix it, because it will conflict with next/previous error shortcuts (see #611). As far as I know, it is impossible to make a link clickable with mouse, but exclude it from shortcuts. However, we can add a right click dropdown which would suggest to open a link in browser. But I am afraid nobody will understand that it is possible.

alygin:
There's obviously a demand for such a feature. Maybe we could combine the dropdown list with a error code highlight (by means of `com.intellij.execution.filters.Filter.Result.highlightAttributes`) to give the user a hint about the possibility of doing something with this error code. What do you think?

matklad:
Fixed by #788, thanks @d9n!

PR #788, opened 2016-11-23
(MISC): Console now recognizes new error message format

d9n:
The compiler used to output errors with an [--explain E####] message,
but recent changes now use a new error[E####] format instead.

This CL updates the relevant console filter to recongize both
formats. If it comes across either pattern, it will turn them into
links to the appropriate Rust Compiler Error Index entry.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/RustExplainFilter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/RustExplainFilter.kt
@@ -5,29 +5,30 @@ import com.intellij.execution.filters.Filter
 import com.intellij.execution.filters.Filter.Result
 import com.intellij.ide.browsers.OpenUrlHyperlinkInfo
 import com.intellij.openapi.project.DumbAware
+import java.util.regex.Matcher
 import java.util.regex.Pattern
 
 
 /**
- * Filters output for “[--explain Exxxx]” and links to the relevant
- * documentation.
+ * Filters output for “[--explain Exxxx]” (or other similar patterns) and links
+ * to the relevant documentation.
  */
 class RustExplainFilter : Filter, DumbAware {
-    private val link_length: Int = 15
-    private val pattern: Pattern = Pattern.compile("--explain E(\\d{4})")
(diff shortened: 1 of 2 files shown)


alygin:
There's [an issue](https://github.com/intellij-rust/intellij-rust/issues/731) about that. Such an implementation breaks navigation through code errors and it seems there's no easy way to make it work correctly in all scenarios.

d9n:
Thanks for the quick reply and I appreciate the added context. That's really unfortunate about the automatic clicking on links, since clicking on the error message to get to the docs is so useful, especially for new users. I'll play around with it a bit more and if I can think of anything I'll update this patch request.

alygin:
I agree that such a feature would be very handy I already tried to find a way to implement it without breaking things. Unfortunately, I failed to find a proper solution.

I've added a suggestion to the issue. Maybe it will help you somehow. If you have any thoughts on it, let's discuss there.

matklad:
Yep, I think a way to go here is to add a right click action to the existing error filter, which would suggest opening explain in the browser. You can see `com.intellij.ide.browsers.OpenUrlHyperlinkInfo` as an example of link with a popup. In the long run, we should be able to get errors from the RLS and perhaps make a gutter icons or something inline in the code.

d9n:
I may have gotten around this issue, but of course there may be additional concerns that I'm not addressing. Please see parent commit https://github.com/intellij-rust/intellij-rust/pull/788/commits/243bd601f05d4a693adcc392c87f5293b48402c6

P.S. IntelliJ and its overabundance of instanceof checks for special case logic in their core API makes me sad sometimes :/

matklad:
@d9n wow, I didn't know that `BrowserHyperlinkInfo` are skipped! Kudos!

matklad:
>IntelliJ and its overabundance of instanceof checks for special case logic in their core API makes me sad sometimes :/

I am permanently sad because of this :) Although sometimes they are justified (hard to add a method to interface prior Java8, working with AST in a fully typed manner is generally hard).

PR #792, opened 2016-11-26
(ART): update file icons

mkaput:
Unfortunately I'm not skilled enough to make nice looking logo-based file icons (Rust logo looks ugly and very clumsy in small sizes; and I did old icons as bitmaps), so I opted to use shortcut-based style.


alygin:
The current cargo file icons are way outdated. Maybe they should be revamped too?

mkaput:
> The current cargo file icons are way outdated. Maybe they should be revamped too?

@alygin Do you have any idea, how it could look like? Current ones are just resized Cargo logos.

alygin:
Given the trend, I think it should be something flat and resemble the official Cargo logo. I only have two ideas:

1) A flat full-face picture of the stack of boxes from the official logo. Sort of this:

cargo_ico_1

2) The standard IntelliJ file icon but with colors from the official logo. Something like this:

cargo_ico_2

A lock or a check-flag can be placed somewhere for the lock-files.

That's the limits of my design capabilities )

mkaput:
1) is awesome!!! Please submit a PR! 😄

alygin:
I will. But let's wait a bit, maybe someone will come up with something even more awesome )

ice1000:
The screenshots on the website are using old icons. Maybe we should update those screenshots too.

PR #793, opened 2016-11-27
(TEMPL): Postfix match

ujpv:


src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustEnumVariantImplMixin.kt
```diff
@@ -1,15 +1,21 @@
...
 import org.rust.lang.core.psi.impl.RustNamedElementImpl
+import org.rust.lang.core.symbols.RustPath
 import javax.swing.Icon
 
 
 abstract class RustEnumVariantImplMixin(node: ASTNode) : RustNamedElementImpl(node), RustEnumVariantElement {
     override fun getIcon(flags: Int): Icon = RustIcons.ENUM_VARIANT
+
+    override val canonicalCratePath: RustPath? get() =
+    RustPsiImplUtil.canonicalCratePath(this)
```
matklad: Indentation is off. Also I think instead of `if (element is RustEnumVarianeElement`) check in the `PsiImplUtils`, you can just implement enum variant specific logic here. That's precisely the idea behind virtual functions: instead of an explicit `instanceof` check, you just call a method and it gets dispatched to the appropriate implementation based on the run time type of an object.

ujpv:
I rewrote this template.

matklad:
LGTM!

PR #794, opened 2016-11-27
(INSP): Automatic inspections suppression

### LINKED ISSUE #775: I shouldn't see warnings in foo.rs for items I allowed in lib.rs (like #[allow(non_snake_case)])
if I have a crate, say, `rustc_llvm`, containing some files like `lib.rs` and `ffi.rs`, Intellij Rust shouldn't complain about non-snake case names in `ffi.rs` if `lib.rs` contains

```
...
#[allow(non_snake_case)]
...
```

Same applies to other `#[allow(...)]` and `#[deny(...)]` uses.

alygin:
I've implemented a function that detects the level of the given lint for the given PSI element (`PsiElement.lintLevel(RustLint): RustLintLevel`). 

The function analyses `allow`, `warn` and `deny` attributes and follows the same rules as `rustc` does. As the result, it allows to automatically suppress inspections based on the attributes in the source code. Potentially, it can also be used to automatically change the inspections warning levels. It's now used in the naming inspections and so fixes #775.

The scope of analysis is now limited by a file. This is the only known deviation from how `rustc` works. What would be the better way [than I use](https://github.com/alygin/intellij-rust/blob/f32d0628f96bd4448afdfe1ae26ba23943f504ba/src/main/kotlin/org/rust/lang/core/psi/RustCompositeElement.kt#L30) to iterate through the element parents without being limited by the current file?

I'm also not sure what's the best place for such a function. Would it be better to move it somewhere else from `RustCompositeElement`?

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
@@ -7,7 +7,6 @@ import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.inspections.fixes.RenameFix
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.RustParameterElementImpl
-import org.rust.lang.core.psi.util.parentOfType
 
 /**
  * Base class for naming inspections. Implements the core logic of checking names
@@ -18,13 +17,14 @@ abstract class RustNamingInspection(
     val styleName: String,
     elementTitle: String = elementType
 ) : RustLocalInspectionTool() {
+    abstract val lint: RustLint
     val dispName = elementTitle + " naming convention"
     override fun getDisplayName(): String = dispName
 
     fun inspect(id: PsiElement?, holder: ProblemsHolder, fix: Boolean = true) {
         if (id == null) return
         val (isOk, suggestedName) = checkName(id.text)
-        if (isOk || suggestedName == null || hasAllowAttribute(id)) return
+        if (isOk || suggestedName == null || id.lintLevel(lint) == RustLintLevel.ALLOW) return
 
         val fixEl = id.parent
         val fixes = if (fix && fixEl is PsiNamedElement) arrayOf(RenameFix(fixEl, suggestedName)) else emptyArray()
@@ -33,8 +33,6 @@ abstract class RustNamingInspection(
             "$elementType `${id.text}` should have $styleName case name such as `$suggestedName`",
             *fixes)
     }
-
-    open fun hasAllowAttribute(id: PsiElement): Boolean = false
 
     abstract fun checkName(name: String): Pair<Boolean, String?>
 }
@@ -46,6 +44,8 @@ open class RustCamelCaseNamingInspection(
     elementType: String,
     elementTitle: String = elementType
 ) : RustNamingInspection(elementType, "a camel", elementTitle) {
+    override val lint = RustLint.NonCamelCaseTypes
+
     override fun checkName(name: String): Pair<Boolean, String?> {
         val str = name.trim('_')
         if (!str.isEmpty() && str[0].canStartWord && '_' !in str) {
@@ -82,6 +82,8 @@ open class RustCamelCaseNamingInspection(
  * Checks if the name is snake_case.
  */
 open class RustSnakeCaseNamingInspection(elementType: String) : RustNamingInspection(elementType, "a snake") {
+    override val lint = RustLint.NonSnakeCase
+
     override fun checkName(name: String): Pair<Boolean, String?> {
         val str = name.trim('_')
         if (!str.isEmpty() && str.all { !it.isLetter() || it.isLowerCase() }) {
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/lang/core/psi/RustCompositeElement.kt
```diff
@@ -10,3 +13,21 @@ interface RustCompositeElement : PsiElement {
 
 val RustCompositeElement.containingMod: RustMod?
     get() = PsiTreeUtil.getStubOrPsiParentOfType(this, RustMod::class.java)
+
+/**
+ * Returns the level of the given lint for this element.
+ */
+fun PsiElement.lintLevel(lint: RustLint): RustLintLevel {
```
matklad: I think it's better to swap arguments: `RustLint.lintLevel(element: PsiElement)`. `PsiElement`s are used everywhere, and you rarely have a need for `lintLevel`.

src/main/kotlin/org/rust/lang/core/psi/RustCompositeElement.kt
```diff
@@ -10,3 +13,21 @@ interface RustCompositeElement : PsiElement {
...
+    var level: RustLintLevel? = null
+    if (this is RustDocAndAttributeOwner) {
```
matklad: You can write this as `ancestors.filterIsInstance<RustDocAndAttributeOwner>` I think.

src/main/kotlin/org/rust/lang/core/psi/RustCompositeElement.kt
```diff
@@ -10,3 +13,21 @@ interface RustCompositeElement : PsiElement {
...
+        level = queryAttributes.metaItems
+            .filter { it.metaItemList.any { it.text == lint.id } }
+            .map { RustLintLevel.valueForId(it.identifier.text) }
+            .filterNotNull()
+            .firstOrNull()
+    }
+    if (level == null) {
+        level = parentOfType<RustDocAndAttributeOwner>()?.lintLevel(lint)
```
matklad: No, there's no API for iterating both inside and outside a file. There is `RustMod.superMods` however which should help.

src/main/kotlin/org/rust/ide/inspections/RustNamingInspection.kt
```diff
@@ -18,13 +17,14 @@ abstract class RustNamingInspection(
...
 ) : RustLocalInspectionTool() {
+    abstract val lint: RustLint
     val dispName = elementTitle + " naming convention"
     override fun getDisplayName(): String = dispName
 
     fun inspect(id: PsiElement?, holder: ProblemsHolder, fix: Boolean = true) {
         if (id == null) return
         val (isOk, suggestedName) = checkName(id.text)
-        if (isOk || suggestedName == null || hasAllowAttribute(id)) return
+        if (isOk || suggestedName == null || id.lintLevel(lint) == RustLintLevel.ALLOW) return
```
matklad: Aha, so this will do tree traversal, but only when we have problems. Neat!

src/main/kotlin/org/rust/lang/core/psi/RustCompositeElement.kt
```diff
@@ -10,3 +13,21 @@ interface RustCompositeElement : PsiElement {
...
+
+/**
+ * Returns the level of the given lint for this element.
+ */
+fun PsiElement.lintLevel(lint: RustLint): RustLintLevel {
+    var level: RustLintLevel? = null
+    if (this is RustDocAndAttributeOwner) {
+        level = queryAttributes.metaItems
+            .filter { it.metaItemList.any { it.text == lint.id } }
+            .map { RustLintLevel.valueForId(it.identifier.text) }
```
matklad: `mapNotNull`

src/main/kotlin/org/rust/lang/core/psi/RustDocAndAttributeOwner.kt
```diff
@@ -35,16 +36,55 @@ class QueryAttributes(private val attributes: List<RustAttrElement>) {
...
         get() = attributes.mapNotNull { it.metaItem }
+}
+
+/**
+ * Iterator that walks through both inner and outer attributes without allocating extra collections.
+ * Inner attributes have priority, so they are iterated first.
+ */
+class RustAttributeIterator (
+    val owner: RustDocAndAttributeOwner
+) : Iterator<RustAttrElement> {
```
matklad: I think you can more shortly implement a lazy version with sequence API, without fallback to the `Iterator`:

```Kotlin
val RustDocAndAttributeOwner.allAttributes: Sequence<RustAttrElement>
    get() = Sequence { (this as? RustOuterAttributeOwner)?.outerAttrList.orEmpty().iterator() } +
        Sequence { (this as? RustInnerAttributeOwner)?.innerAttrList.orEmpty().iterator() }
```

:+1: for switching to sequences from lists! Unfortunately base Grammar Kit APIs allocate collections, but this is something we can work around if necessary.

alygin: Oh, silly me, that was so easy.

matklad:
Great! Thanks!

PR #796, opened 2016-11-28
WIP: Introduce local variable

### LINKED ISSUE #786: introduce local variable - refactor.
Intelij usually has cntrl - alt - v to introduce a local variable from an expression.
Are there any plans to support this? is this a feature you guys even want?

Not sure how hard this is to implement but i'd like to give it a shot.
Any pointers for implementing this?

P.S.
Sorry if it is actually there and I'm just to stupid to find it.

LiamClark:
fixes #786 

This is still a work in progress.

Very basic support for introducing local variables is implemented.
Choosing what expression you want to extract.
Choosing replace single or multiple occurrences.

The inline rename dialog is automatically opened, but the suggested variable name is always x.

There are a few basic test cases, but there should be many more.

Right now i have two big problems.
How to extract meaningful names to suggest in the rename dialog.

how should this interact with if-let constructs?


src/main/kotlin/org/rust/lang/refactoring/RustLocalVariableHandler.kt
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+}
+
+fun PsiElement.findBinding() = PsiTreeUtil.findChildOfType(this, RustPatBindingElement::class.java)
+
+class OccurrenceVisitor(val element: PsiElement) : PsiRecursiveElementVisitor() {
+    val foundOccurrences = ArrayList<PsiElement>()
+
+    override fun visitElement(element: PsiElement?) {
```
matklad: You can remove `?` here.

src/main/kotlin/org/rust/lang/refactoring/RustRefactoringSupportProvier.kt
```diff
@@ -2,10 +2,21 @@ package org.rust.lang.refactoring
...
 import com.intellij.psi.PsiElement
+import com.intellij.refactoring.RefactoringActionHandler
 import org.rust.lang.core.psi.RustPatBindingElement
 
 class RustRefactoringSupportProvider : RefactoringSupportProvider() {
     override fun isInplaceRenameAvailable(element: PsiElement, context: PsiElement?): Boolean {
         return element is RustPatBindingElement
     }
+
+    override fun getIntroduceVariableHandler(): RefactoringActionHandler? {
```
matklad: You can write ` override fun getIntroduceVariableHandler() = RustLocalVariableHandler()`.

src/test/kotlin/org/rust/lang/refactoring/IntroduceVariableTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+//    @Ignore
+//    fun testVariable() = checkByFile {
+//        val rustLocalVariableHandler = RustLocalVariableHandler()
+//        openFileInEditor("variable.rs")
+//        rustLocalVariableHandler.invoke(myFixture.project, myFixture.editor, myFixture.file, com.intellij.openapi.actionSystem.DataContext.EMPTY_CONTEXT)
+//    }
+
+
+    fun testMultipleOccurrences() = checkByFile {
```
matklad: Let's try avoid fixture files here. Take a look at these tests: https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/typing/RustEnterInStringLiteralHandlerTest.kt#L8

Here, we use multiline string literal to embed Rust code inside Kotlin, which is a bit more convenient.

LiamClark: @matklad Oh yes! that looks a lite nicer than the fixture files. I picked those up from the contributing, but this looks a lot nicer

src/main/kotlin/org/rust/lang/refactoring/RustLocalVariableHandler.kt
```diff
@@ -0,0 +1,204 @@
...
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.refactoring.IntroduceTargetChooser
+import com.intellij.refactoring.RefactoringActionHandler
+import com.intellij.refactoring.introduce.inplace.InplaceVariableIntroducer
+import com.intellij.refactoring.introduce.inplace.OccurrencesChooser
+import org.rust.lang.core.psi.*
+import java.util.*
+
+class RustLocalVariableHandler : RefactoringActionHandler {
+    override fun invoke(project: Project, editor: Editor?, file: PsiFile?, dataContext: DataContext?) {
```
LiamClark: Sweet, this is how intelij autogenerated the signature for me, but removing it makes it a lot simpler, awesome

src/main/kotlin/org/rust/lang/refactoring/RustLocalVariableHandler.kt
```diff
@@ -0,0 +1,206 @@
...
+
+fun <T> pass(pass: (T) -> Unit): Pass<T> {
+    return object : Pass<T>() {
+        override fun pass(t: T) {
+            pass.invoke(t)
+        }
+    }
+}
+
+fun PsiElement.findBinding() = PsiTreeUtil.findChildOfType(this, RustPatBindingElement::class.java)
```
matklad: Ah, can you please mark these helper functions `private`?

src/main/kotlin/org/rust/lang/refactoring/RustLocalVariableHandler.kt
```diff
@@ -0,0 +1,206 @@
...
+    fun moveEditorToNameElement(editor: Editor, element: PsiElement?): RustPatBindingElement? {
+        val newName = element?.findBinding()
+
+        editor.caretModel.moveToOffset(newName?.identifier?.textRange?.startOffset ?: 0)
+
+        return newName
+    }
+
+    override fun invoke(project: Project, elements: Array<out PsiElement>, dataContext: DataContext?) {
+        println("not from the editor.")
```
matklad: Let's use comment rather then print.

LiamClark:
I changed the tests to use multi line string literals. The test pass however they do throw this exception at the end. with problems during index shutdown. any ideas what might be causing this?

matklad:
Thanks a lot! I can't reproduce test failures, but there were some index shenanigans recently, they might be responsible. I'll merge it and do some cleanup not worthy another round of comments! Feel free to send additional pull requests if you find some features/tests missing!

matklad:
tip: if you have a class `Foo` and a test named `FooTest` or `FooTestCase`, then you can quickly navigate between the two using **go to test** action  (<kbd>Ctrl+Shift+T</kbd>).

PR #797, opened 2016-11-29
(INSP): Suppress Self Convention Inspection for copyable items

### LINKED ISSUE #515: Do not fire RustSelfConventionInspection for Copyable items
Sample ([source](https://github.com/rust-lang-nursery/regex/blob/master/src/input.rs#L242-L292))

``` rust
#[derive(Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
pub struct Char(u32);

impl Char {
    /// Returns true iff the character is absent.
    #[inline]
    pub fn is_none(self) -> bool { self.0 == u32::MAX }
}
```

triggers RustSelfConventionInspection, while that `self` is totally ok here

alygin:
Suppresses Self Convention Inspections for items that derive `Copy`.
It won't work if one uses `impl Copy for T` explicitly, but I think such cases are rare and can be ignored here.

Fixes #515


matklad:
Thanks!

PR #800, opened 2016-11-30
(INT) 3 Intentions : Set reference mutable/immutable, Implements struct

Litarvan:
## Added 3 new intentions

* Set reference mutable
* Set reference immutable
* Implements struct

#### Before
```java
&type
```

#### After
```java
&mut type
```

#### Before
```java
&mut type
```

#### After
```java
&type
```

#### Before
```java
struct MyStruct {
    ...
}
```

#### After

**Note: Cursor moved between the braces of the impl**

```java
struct MyStruct {
    ...
}

impl MyStruct {}
````

### CODE REVIEW

**Comment by alygin** [comment]
I think it would be great to automatically add a new line in the Implement Struct intention, so the result would be:

```rust
impl MyStruct {
    <caret>
}
```

Litarvan:
In some rare cases you want to let it empty, so i put the caret here : 
```rust
impl MyStruct {<caret>}
```

So you can just press enter ^^

mkaput:
Rare cases are rare cases, so I agree with @alygin, but there is problem that formatter enforces @Litarvan's style and probably some workarounds would be needed, wouldn't they (not sure)?

Litarvan:
I thought of the formatting problem but i saw in the other intentions that there was the same problem.
I think this is not important to add a line break in the impl if the user just has to press one key but, if you are sure, I (or you) can add it.

matklad:
Nice!

I think that adding a newline would be nice, but it is not super important. 

Would you mind changing tests to the new "inline code" style format, like here https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/intentions/MoveTypeConstraintToWhereClauseIntentionTest.kt ? I guess I should update Contributing.md section about the tests :)

Litarvan:
Wow these tests are much better ! Done :)

matklad:
Thanks! 

Some possible future directions: 

* Implement struct can move caret to an existing impl block if there is one,

* navigate to implementation can be implemented for structs.

Litarvan:
I'll think of that :)

PR #801, opened 2016-11-30
(MISC): Colored output

### LINKED ISSUE #646: No colors from rustc/cargo
The run window doesn't show colored compiler output, even with `--color always`.

I've changed the stderr color, tried disabling it as well.
I'm running 0.0.1.955 in PyCharm Community 2016.2.2, rustc 1.11.0 and cargo 0.12.0.

My program is outputting colored text to stdout which works fine, if I output to stderr, it doesn't though.

alygin:
Makes the `cargo` and `rustc` output colored on Linux and Mac.
Unfortunately, I don't know how to make it work on Windows and have no Windows machine for experiments. It uses Win API by default and can switch to `msys` under certain circumstances, but I think neither of these will work in IDEA. Though, I might be wrong about it.

Fixes #646 for Linux and Mac

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
@@ -76,8 +76,10 @@ class Cargo(
             .withEnvironment(CargoConstants.RUSTC_ENV_VAR, pathToRustExecutable)
             .withEnvironment(environmentVariables)
 
-        // Make output colorful
-        if ((SystemInfo.isLinux || SystemInfo.isMac) && additionalArguments.none { it.startsWith("--color") }) {
+        // Make output colored
+        if ((SystemInfo.isLinux || SystemInfo.isMac)
+            && command != "fmt"
+            && additionalArguments.none { it.startsWith("--color") }) {
             cmdLine
                 .withEnvironment("TERM", "linux")
                 .withRedirectErrorStream(true)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -67,14 +68,26 @@ class Cargo(
...
+        val cmdLine = GeneralCommandLine(pathToCargoExecutable)
             .withWorkDirectory(projectDirectory)
             .withParameters(command)
-            .withParameters(additionalArguments)
             .withEnvironment(CargoConstants.RUSTC_ENV_VAR, pathToRustExecutable)
             .withEnvironment(environmentVariables)
 
+        // Make output colored
+        if ((SystemInfo.isLinux || SystemInfo.isMac)
+            && command != "fmt"
```
matklad: I think we should use a whitelist of commands which accept `--color` rather than a blacklist of those who don't, because you can add arbitrary subcommands to cargo, and they might not like `--color`, just like `cargo fmt`.

alygin: You're right, I'll add the list.

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -104,4 +117,8 @@ class Cargo(
             throw ExecutionException(e)
         }
     }
+
+    private companion object {
```
matklad: TIL that you can mark companion object with private! Thanks!

matklad:
Great!

PR #803, opened 2016-12-02
(HINT): Functions/methods arguments hint

### LINKED ISSUE #437: Parameter info in function and method call arguments
Extension point `codeInsight.parameterInfo` via ParameterInfoHandler interface allows function parameters to be shown when the users presses `Ctrl+P` where relevant. For fn calls, this is trivial so long as the fn is resolved properly, but robust type information won't be available without type system support. Method calls require type system support.

ParameterInfoHandler interface: https://upsource.jetbrains.com/idea-ce/file/idea-ce-1731d054af4ca27aa827c03929e27eeb0e6a8366/platform/lang-api/src/com/intellij/lang/parameterInfo/ParameterInfoHandler.java

alygin:
This one adds arguments info hints to function and method calls (Cmd+P). It works where the current resolvers work:

screen shot 2016-12-02 at 08 25 39

A possible further improvement would be to extract descriptions of the aliased types in generic functions and show them instead of aliases. It might be the common mechanism for both these hints and the completion engine (and something else?).

Fixes #437

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
@@ -1,4 +1,4 @@
-package org.rust.ide.info
+package org.rust.ide.hints
 
 import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.lang.parameterInfo.*
@@ -12,6 +12,8 @@ import org.rust.lang.core.psi.util.parentOfType
  * Provides funcions/methods afruments hint.
  */
 class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsDescription> {
+
+    var hintText: String = ""
 
     override fun couldShowInLookup() = true
 
@@ -43,7 +45,7 @@ class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsD
 
     override fun showParameterInfo(element: PsiElement, context: CreateParameterInfoContext) {
         if (element !is RustArgListElement) return
-        val argsDescr = RustArgumentsDescription.tryGetDescription(element) ?: return
+        val argsDescr = RustArgumentsDescription.findDescription(element) ?: return
         context.itemsToShow = arrayOf(argsDescr)
         context.showHint(element, element.textRange.startOffset, this)
     }
@@ -70,8 +72,9 @@ class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsD
             return
         }
         val range = p.getArgumentRange(context.currentParameterIndex)
+        hintText = p.presentText
         context.setupUIComponentPresentation(
-            p.parametersPresentText,
+            hintText,
             range.startOffset,
             range.endOffset,
             !context.isUIComponentEnabled,
@@ -85,7 +88,7 @@ class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsD
      */
     private fun findArgumentIndex(place: PsiElement): Int {
         val callArgs = place.parentOfType<RustArgListElement>() ?: return INVALID_INDEX
-        val descr = RustArgumentsDescription.tryGetDescription(callArgs) ?: return INVALID_INDEX
+        val descr = RustArgumentsDescription.findDescription(callArgs) ?: return INVALID_INDEX
         var index = -1
         if (descr.arguments.isNotEmpty()) {
             index += generateSequence(callArgs.firstChild, { c -> c.nextSibling})
@@ -104,20 +107,26 @@ class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsD
     }
 }
 
+/**
+ * Holds information about arguments from func/method declaration
+ */
 class RustArgumentsDescription(
     val arguments: Array<RustArgumentDescription>
 ) {
     fun getArgumentRange(index: Int): TextRange {
-        if (index < 0) return TextRange.EMPTY_RANGE
+        if (index < 0 || index >= arguments.size) return TextRange.EMPTY_RANGE
         val start = (0..index - 1).sumBy { arguments[it].textLen + 2 }
         val range = TextRange(start, start + arguments[index].textLen)
         return range
     }
 
-    val parametersPresentText = if (arguments.isEmpty()) "<no arguments>" else arguments.joinToString(", ")
+    val presentText = if (arguments.isEmpty()) "<no arguments>" else arguments.joinToString(", ")
 
     companion object {
-        fun tryGetDescription(args: RustArgListElement): RustArgumentsDescription? {
+        /**
+         * Finds declaration of the func/method and creates description of its arguments
+         */
+        fun findDescription(args: RustArgListElement): RustArgumentsDescription? {
             val call = args.parent
             val decl = when (call) {
                 is RustCallExprElement -> call.declaration
--- a/src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
@@ -4,12 +4,11 @@ import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.lang.parameterInfo.*
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
-import com.intellij.psi.util.PsiTreeUtil
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.util.parentOfType
 
 /**
- * Provides funcions/methods afruments hint.
+ * Provides functions/methods arguments hint.
  */
 class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsDescription> {
 
@@ -36,7 +35,7 @@ class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsD
     }
 
     fun findElementForParameterInfo(contextElement: PsiElement) =
-        PsiTreeUtil.getParentOfType(contextElement, RustArgListElement::class.java)
+        contextElement.parentOfType<RustArgListElement>()
 
     override fun findElementForUpdatingParameterInfo(context: UpdateParameterInfoContext) =
         context.file.findElementAt(context.editor.caretModel.offset)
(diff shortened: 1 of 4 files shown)


alygin:
I wonder if there's a new public API to implement those parameter name hints from the latest IDEA release?

screen shot 2016-12-02 at 09 02 34

They might be of help in Rust, especially on the early stages of learning.

matklad:
>I wonder if there's a new public API to implement those parameter name hints from the latest IDEA release?

There's com.intellij.codeInsight.hints.InlayParameterHintsProvider for this, but it is experimental and not available in older versions which we do support.

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+import com.intellij.codeInsight.lookup.LookupElement
+import com.intellij.lang.parameterInfo.*
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.parentOfType
+
+/**
+ * Provides funcions/methods afruments hint.
```
matklad: s/funcions/functions

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+    override fun getParametersForDocumentation(p: RustArgumentsDescription, context: ParameterInfoContext?) =
+        p.arguments
+
+    override fun findElementForParameterInfo(context: CreateParameterInfoContext): PsiElement? {
+        val contextElement = context.file.findElementAt(context.editor.caretModel.offset) ?: return null
+        return findElementForParameterInfo(contextElement)
+    }
+
+    fun findElementForParameterInfo(contextElement: PsiElement) =
+        PsiTreeUtil.getParentOfType(contextElement, RustArgListElement::class.java)
```
matklad: `contextElement.parentOfType<RustArgListElement>`

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+     * Finds index of the argument in the given place
+     */
+    private fun findArgumentIndex(place: PsiElement): Int {
+        val callArgs = place.parentOfType<RustArgListElement>() ?: return INVALID_INDEX
+        val descr = RustArgumentsDescription.findDescription(callArgs) ?: return INVALID_INDEX
+        var index = -1
+        if (descr.arguments.isNotEmpty()) {
+            index += generateSequence(callArgs.firstChild, { c -> c.nextSibling})
+                .filter { it.text == "," }
+                .takeWhile { it.textRange.startOffset < place.textRange.startOffset }
```
matklad: You don't need `takeWhile` here, count accepts a predicate.

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Holds information about arguments from func/method declaration
+ */
+class RustArgumentsDescription(
+    val arguments: Array<RustArgumentDescription>
+) {
+    fun getArgumentRange(index: Int): TextRange {
+        if (index < 0 || index >= arguments.size) return TextRange.EMPTY_RANGE
+        val start = (0..index - 1).sumBy { arguments[it].textLen + 2 }
```
matklad: `arguments.take(index).sumBy { it.textLen + 2 }`

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.parentOfType
+
+/**
+ * Provides funcions/methods afruments hint.
+ */
+class RustParameterInfoHandler : ParameterInfoHandler<PsiElement, RustArgumentsDescription> {
```
matklad: Can we change `PsiElement` to `RustArgListElement`?

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+    val presentText = if (arguments.isEmpty()) "<no arguments>" else arguments.joinToString(", ")
+
+    companion object {
+        /**
+         * Finds declaration of the func/method and creates description of its arguments
+         */
+        fun findDescription(args: RustArgListElement): RustArgumentsDescription? {
+            val call = args.parent
+            val decl = when (call) {
+                is RustCallExprElement -> call.declaration
```
matklad: If you write `call.declaration?.paramenters` then you wouldn't need the second `when`.

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+            return RustArgumentsDescription(params ?: emptyArray())
+        }
+    }
+}
+
+class RustArgumentDescription(
+    val name: String?,
+    val type: String?
+) {
+    val textLen = (name?.length ?: 1) + 2 + (type?.length ?: 1)
```
matklad: `val textLen = toString().length`. Or perhaps we don't need at all?

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+    override fun toString() = (name ?: "?") + ": " + (type ?: "?")
```
matklad: I'd use a template here: `"${name ?: "?"}: ${type ?: "?"}"`

src/test/kotlin/org/rust/ide/hints/RustParameterInfoHandlerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    private fun checkByText(code: String, hint: String, index: Int) {
+        myFixture.configureByText("main.rs", code)
+        val handler = RustParameterInfoHandler()
+        val createContext = MockCreateParameterInfoContext(myFixture.editor, myFixture.file)
+
+        // Check hint
+        val elt = handler.findElementForParameterInfo(createContext)
+        if (hint.isNotEmpty()) {
+            elt ?: throw AssertionFailedError("Element not found")
```
matklad: `checkNotNull(elt) { "Element not found" }` would be a bit more direct.

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    override fun updateParameterInfo(place: PsiElement, context: UpdateParameterInfoContext) {
+        val argIndex = findArgumentIndex(place)
+        if (argIndex == INVALID_INDEX) {
+            context.removeHint()
+            return
+        }
+        context.setCurrentParameter(argIndex)
+        if (context.parameterOwner == null) {
```
matklad: Consider using `when` here.

src/test/kotlin/org/rust/ide/hints/RustParameterInfoHandlerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+            handler.showParameterInfo(elt, createContext)
+            val items = createContext.itemsToShow ?: throw AssertionFailedError("Paremeters are not shown")
```
matklad: s/Paremetets/Parameters

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
alygin: There's not much of documentation on how this handler exactly works, but it doesn't show anything with `RustArgListElement`.

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
alygin: A bit of over overengineering here. I've got rid of this class entirely.

src/test/kotlin/org/rust/ide/hints/RustParameterInfoHandlerTest.kt
alygin: I need to switch the spellchecker back on )

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
matklad: This type parameter just ties together signatures of `find*` and `update/show*` methods. If you use `RustArgListElement`, you would not need `if (element !is RustArgListElement) return` check in `showParameterInfo`.

src/test/kotlin/org/rust/ide/hints/RustParameterInfoHandlerTest.kt
matklad: The idiom is to bind the result of `checkNotNull`:

```
val nonNullElt = checkNotNull(elt)
```

though, it'd be inconvenient here because of the shadowing :(

src/main/kotlin/org/rust/ide/hints/RustParameterInfoHandler.kt
alygin: When I change `PsiElement` to `RustArgListElement` it just stops displaying hints and doesn't call `updateParameterInfo()`. It looks like there's more behind the scenes than just tying methods together. I checked a couple of implementations in other plugins that I could find quickly ([Dart](https://github.com/JetBrains/intellij-plugins/blob/d0fcca865b56466415f13763cb3c517830c8a708/Dart/src/com/jetbrains/lang/dart/ide/info/DartParameterInfoHandler.java) and [ColdFusion](https://github.com/JetBrains/intellij-plugins/blob/d0fcca865b56466415f13763cb3c517830c8a708/CFML/src/com/intellij/coldFusion/UI/editorActions/CfmlParameterInfoHandler.java)), they use `PsiElement` as well.

But if you're sure it must work with `RustArgListElement`, I'll have a look under the hood a bit later.

matklad: >But if you're sure it must work with RustArgListElement, I'll have a look under the hood a bit later.

No, I am not sure, let's merge this as is. Although com.intellij.codeInsight.hint.api.impls.MethodParameterInfoHandler uses more specific type there. But it is `ParameterInfoHandlerWithTabActionSupport` (so that you can use <kbd>Tab</kbd> to navigate through parameters).

matklad:
Thanks!

PR #805, opened 2016-12-02
(MISC): Error links improvements

### LINKED ISSUE #804: Error code links improvement
The error code links [are back](https://github.com/intellij-rust/intellij-rust/pull/788), which is cool. But I think they could be improved a bit:

1. The compiler emits codes not only for errors but also for warnings (for instance, the [E0122](https://doc.rust-lang.org/error-index.html#E0122)). Such warning codes are ignored now though could be "linkified" too.
2. The code links now wrap the whole phrase `error[E0XXX]`. I think it's only the code with the braces that should become a link. When using the colored output it's easier to quickly scan the very left side of the output winodow and find the boundaries of the error/warning messages. Some errors and warnings have no codes and their markers now look different.

Here's an example of the described problems (three issues with three different styles):

screen shot 2016-12-02 at 11 10 26

alygin:
Fixes #804

Those messages now look like this:

screen shot 2016-12-02 at 19 00 03


matklad:
Awesome!

PR #811, opened 2016-12-03
(MISC): Gutter icon position fix

alygin:
Fixes positioning of the "go to declaration" gutter icon when comments or attributes are provided for the function/method.

Before the fix, it was placed at the beginning of the comment/attribute.


matklad:
Thanks!

PR #812, opened 2016-12-03
(MISC): Coloring ? operator

### LINKED ISSUE #785: Syntax coloring for ? symbol.
When use `?` instead of `try!()` macro there are no evidence of it in the syntax highlighting. That make very hard to identify these exit points.

If you have no time to do it please point me where I should looking for to implement it.

alygin:
Makes the `?` operator colored as a keyword ([screenshots](https://github.com/intellij-rust/intellij-rust/issues/785#issuecomment-264547200)). A setting to customise the color is also provided.

We can decide what color to use before merging the PR.

Fixes #785. Also fixes the `self` parameter display in the color settings window (it's displayed as a simple parameter now).


matklad:
Yep, I think coloring it as a keyword makes a lot of sense!

PR #819, opened 2016-12-07
(Misc) Local variable names suggestions

LiamClark:
This pr expands the functionality introduced in #796

So far it follows these rules:

If the type is resolved suggest the first letter of the type name.
If its an argument to a function call, suggest the name of the argument in the function definition.
If its a function call suggest the name of the function and the enclosing namespace (if any).
If a name is already bound in the local scope do not suggest it.

This is mainly based on what the go plugin does.

Intellij in java also does the following if you are introducing a variable for a string literal, it uses the contents of the literal for the name, would this special case also make sense for rust?

We might want to use the is array setting on the Intellij NameUtil, however i'm not completely certain what it does and how much value it might add, any thoughts on this?

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
@@ -1,13 +1,10 @@
 package org.rust.lang.refactoring
 
-import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.codeStyle.NameUtil
 import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.inspections.toSnakeCase
-import org.rust.ide.navigation.goto.RustSymbolNavigationContributor
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.util.childOfType
 import org.rust.lang.core.psi.util.parentOfType
 import org.rust.lang.core.types.RustUnknownType
 import org.rust.lang.core.types.util.resolvedType
@@ -19,6 +16,7 @@ import java.util.*
  * If the type is resolved suggest the first letter of the type name.
  * If its an argument to a function call, suggest the name of the argument in the function definition.
  * If its a function call suggest the name of the function and the enclosing name space (if any).
+ * If the expression is in a struct literal (Foo {x: 5, y: 6}) suggest the tag for the expression as a name
  * If a name is already bound in the local scope do not suggest it.
  */
 fun PsiElement.suggestNames(): LinkedHashSet<String> {
@@ -28,6 +26,7 @@ fun PsiElement.suggestNames(): LinkedHashSet<String> {
     val foundNames = when {
         this.isArgument() -> rustNameUtil(nameForArgument())
         this is RustCallExprElement -> nameForCall(this).flatMap(::rustNameUtil)
+        isStructField() -> rustNameUtil((this.parent as? RustStructExprFieldElement)?.identifier?.text ?: "")
         else -> emptyList()
     }
 
@@ -38,6 +37,7 @@ fun PsiElement.suggestNames(): LinkedHashSet<String> {
 
     return names
 }
+
 
 private fun nameForType(expr: PsiElement): String? {
     val type = (expr as? RustExprElement)?.resolvedType
@@ -83,5 +83,6 @@ fun findNamesInLocalScope(expr: PsiElement): List<String> {
 }
 
 private fun PsiElement.isArgument() = this.parent is RustArgListElement
+private fun PsiElement.isStructField() = this.parent is RustStructExprFieldElement
 
 private fun rustNameUtil(name: String) = NameUtil.getSuggestionsByName(name, "", "", false, false, false).map { it.toSnakeCase(false) }
--- a/src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
@@ -1,37 +1,53 @@
-package org.rust.lang.refactoring
+    package org.rust.lang.refactoring
 
 import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
 import com.intellij.psi.codeStyle.NameUtil
 import org.rust.ide.navigation.goto.RustSymbolNavigationContributor
-import org.rust.lang.core.psi.RustArgListElement
-import org.rust.lang.core.psi.RustCallExprElement
-import org.rust.lang.core.psi.RustExprElement
-import org.rust.lang.core.psi.RustFnItemElement
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.util.parentOfType
 import org.rust.lang.core.types.util.resolvedType
 import java.util.*
 
-fun suggestedNames(project: Project, expr: RustExprElement): LinkedHashSet<String> {
+fun suggestedNames(project: Project, expr: PsiElement): LinkedHashSet<String> {
     val names = LinkedHashSet<String>()
 
-    val typeString = expr.resolvedType.toString()
-    //doesn't really seem to do that much.
-    if (typeString != "unkown") {
-        names.addAll(RustNameUtil(typeString.take(1)))
-    }
+
+    nameForType(expr)?.let { names.addAll(RustNameUtil(it)) }
 
     if (expr.isArgument()) {
         val name = nameForArgument(project, expr)
         val suggestionsByName = RustNameUtil(name)
         names.addAll(suggestionsByName)
     } else if (expr is RustCallExprElement) {
-        names.addAll(RustNameUtil(expr.expr.text))
+        names.addAll(nameForCall(expr).flatMap(::RustNameUtil))
     }
 
     return names
 }
 
-fun nameForArgument(project: Project, expr: RustExprElement): String {
+fun nameForType(expr: PsiElement): String? {
+    val typeString = (expr as? RustExprElement)?.resolvedType.toString()
+    //doesn't really seem to do that much.
+    if (typeString != "unkown" && typeString != "null") {
+        return typeString.take(1)
+    }
+
+    return null
+}
+
+fun nameForCall(expr: RustCallExprElement): List<String> {
+    val pathElement = expr.expr
+    if (pathElement is RustPathExprElement) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * This suggests names for an expression about to be bound to a local variable
+ *
+ * If the type is resolved suggest the first letter of the type name.
+ * If its an argument to a function call, suggest the name of the argument in the function definition.
+ * If its a function call suggest the name of the function and the enclosing name space (if any).
+ * If a name is already bound in the local scope do not suggest it.
+ */
+fun suggestedNames(project: Project, expr: PsiElement): LinkedHashSet<String> {
```
mkaput: Maybe:
```kotlin
fun PsiElement.suggestNames()
```

?

You should be able to get project using `PsiElement#getProject()` method.

LiamClark: Indeed looks a lot cleaner, thanks! didn't know that one existed

LiamClark:
Also really hoping for more suggestions of information to use to find even better names

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.ide.navigation.goto.RustSymbolNavigationContributor
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.childOfType
+import org.rust.lang.core.psi.util.parentOfType
+import org.rust.lang.core.types.util.resolvedType
+import java.util.*
+
+/**
+ * This suggests names for an expression about to be bound to a local variable
```
matklad: If this is applicable only to the expressions than it makes sense to use `RustExprElement` instead of plain `PsiElement`.

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -0,0 +1,101 @@
...
+fun findNamesInLocalScope(expr: PsiElement): List<String> {
+    val blockScope = expr.parentOfType<RustBlockElement>(strict = false)
+    val letDecls = PsiTreeUtil.findChildrenOfType(blockScope, RustLetDeclElement::class.java)
+
+    return letDecls.map { it.pat?.text }.filterNotNull()
+}
+
+private fun PsiElement.isArgument() = this.parent is RustArgListElement
+
+private fun RustNameUtil(name: String) = NameUtil.getSuggestionsByName(name, "", "", false, false, false).map(String::toSnakeCase)
```
matklad: This should start with a lowercase latter.

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -0,0 +1,101 @@
...
+}
+
+private fun findFnImpl(project: Project, callExpr: RustCallExprElement): RustFnItemElement? {
+    val navigator = RustSymbolNavigationContributor()
+    val items = navigator.getItemsByName(callExpr.firstChild.text, null, project, false)
+
+    return items.firstOrNull() as? RustFnItemElement
+}
+
+fun String.toSnakeCase(): String {
```
matklad: We already have one `toSnakeCase` function in `RustNamingInspection` already :)

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -0,0 +1,101 @@
...
+    val usedNames = findNamesInLocalScope(this)
+    names.removeAll(usedNames)
+
+    return names
+}
+
+private fun nameForType(expr: PsiElement): String? {
+    val typeString = (expr as? RustExprElement)?.resolvedType.toString()
+    //doesn't really seem to do that much.
+    if (typeString != "unknown" && typeString != "null") {
```
matklad: Let's be more specific here and return `null` if `type is RustUnknownType || type == null`

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -0,0 +1,101 @@
...
+fun PsiElement.nameForArgument(): String {
+    val call = this.parentOfType<RustCallExprElement>(strict = false) ?: return ""
+
+    val parameterIndex = call.argList.children.indexOf(this)
+    val fn = findFnImpl(project, call)
+
+    return fn?.parameters?.parameterList?.get(parameterIndex)?.pat?.text ?: ""
+}
+
+private fun findFnImpl(project: Project, callExpr: RustCallExprElement): RustFnItemElement? {
```
matklad: I think you want to use `callExpr.reference.resolve()` here.

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
LiamClark: I'm not sure actually, I think there could be other cases, this is mainly done like this right now, due to the call site in the introduce local variable handler only having type information for psi element.

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
mkaput: And I believe there could be one in platform utils

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
LiamClark: Yes this seemed very hacky indeed, however callExpr.reference returns null in my argument test case.

What exactly does do?

LiamClark: Also what's the preferred way to do this for a methods?

matklad:
PSA: I'm refactoring `RustElementFactory` a bit, so expect slight merge conflicts! :)

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
matklad: >due to the call site in the introduce local variable handler only having type information for psi element.

If we expect to see only expressions there, it makes sense to assert that fact:

```
if (psiElement !is RustExprElement) error("Expected to see an expression during introduca variable refactoring: `${psiElement.text}"`)
```

matklad:
>Also really hoping for more suggestions of information to use to find even better names

I think this is a good start, and we should merge this! One possible improvement is to try to use the name of the type for factory-like functions (that is, suggest `string` instead of `new` in on of the tests).

LiamClark:
It actually already does this, its just the second option in the dialog

himikof:
Considering that there is a strong convention in Rust for calling the constructor associated function `T::new`, and `new` is a totally useless name for the call result (the thing being constructed), I would suggest special casing the function name `new` so it ranks below other names (or even not suggested at all). In my opinion, the most relevant names for the test case in question are `buf` (from `std::io::Read::read_to_string` argument name) and `string`. The suggested special casing could also apply to `Default::default`, as the situation is similar there.

LiamClark:
@himikof 

I introduced the special case for new, however with default i'm not sure how much good it would do.

If the default was used as an argument that information will take priority.
In the case where it's just a single default statement:

`Default::defa<caret>ult;`
We don't have a type or anything else to suggest, but the name default.
Do you know any case where it would matter?

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
```diff
@@ -28,6 +26,7 @@ fun PsiElement.suggestNames(): LinkedHashSet<String> {
     val foundNames = when {
         this.isArgument() -> rustNameUtil(nameForArgument())
         this is RustCallExprElement -> nameForCall(this).flatMap(::rustNameUtil)
+        isStructField() -> rustNameUtil((this.parent as? RustStructExprFieldElement)?.identifier?.text ?: "")
```
matklad: I would have just saved `parent` in the local variable and then inlined the functions:

```
val parent = parent //
val foundNames = when {
    if parent is RustStructFieldElement -> .... 
}
```

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
matklad: @LiamClark what do yo think about this? Looks like a quite lot of places inside `RustIntroduceVariableRefactoring` would need to change from `PsiElement` to `RustExprElement` (like `findOccurrences` and `OccurrencesChooser.simpleChooser<RustExprElement>`), but this is actually a good idea to do anyway.

matklad: `org.rust.lang.core.psi.RustPsiFactory#createLetDeclaration` can also use `RustExprElement`!

himikof:
@LiamClark No, I didn't mean directly calling `Default::default` (it is not actually possible, because `Default` is a trait and this method is not object-safe). I meant for some type `T : Default` calling `T::default()` (for example, `Vec<usize>::default()`). In this case, variable name suggestion `default` seems totally useless. I understand that implementing it probably requires working trait method resolution (on the other hand, `new` is just a naming convention, so string comparison should be fine there).

src/main/kotlin/org/rust/lang/refactoring/RustNameSuggestions.kt
alygin: @mkaput, the platform provides several methods in the [StringUtil](https://github.com/JetBrains/intellij-community/blob/883883fd0a5f27ba1f0eee23ee46898fe4072a0d/platform/util/src/com/intellij/openapi/util/text/StringUtil.java) class, but there's nothing that could replace this implementation.

There're several subtleties in how `rustc` transforms names from one notation to another when it suggests name substitutions. This function takes them into account. Since other classes also need it, maybe, it's the right time to move it to some utility class, or to `RustNamedElement`?

matklad:
Looks great, thanks!

Issue #821, opened 2016-12-07
Function links in stack backtraces


alygin:
For Java, Kotlin and perhaps other JVM languages, IDEA provides a nice stacktrace presentation. It adds links to exact positions in the source code for every line, it separates our methods calls from calls to external libraries, it supports folding etc.

I've prototyped a similar feature for Rust. So far, it makes stack backtraces look like this:

<img width="641" alt="screen shot 2016-12-08 at 00 26 22" src="https://cloud.githubusercontent.com/assets/2101250/20987268/0587a2e4-bcdd-11e6-9ca6-5731dae60e69.png">

Here, the usual hyperlinks lead to the source code of the project we are working on. The grayed hyperlinks lead to external crates. And those not linked are from a special skip list. Every link leads to the declaration of the corresponding function. Trailing hashcodes are dimmed a bit to reduce noise.

But before moving further with this feature, I'd like to discuss it here. What do you think, does it worth implementing at all? Any ideas on how to improve it?

There's also a problem I wasn't able to solve quickly &mdash; it seems that `RustResolveEngine.resolve(path, pivot)` doesn't work for struct methods, it only works for pure and associated functions. Am I right?

alygin:
And to be objective, here are my arguments against this feature:

1. If the described problem with resolving methods is really unsolvable now, we'll have many unlinked lines which will be rather confusing.
2. The mechanism is not sophisticated enough to provide links to the exact call positions, it only leads to method declarations. The developer still needs to search for the exact calls.
3. On Linux, stack backtraces already come with the exact source code links (we now parse them and wrap into hyperlinks). Maybe we just need to wait until such feature comes to other platforms.

The feature might possibly be a bit different &mdash; just highlight parts of stack backtraces to make its analysis easier, but don't add hyperlinks. Or to add hyperlinks only on non-linux platforms.

matklad:
Sorry, looks like I've missed notification about this issue :( 

We surely do need to hyperlink qualified names in stacktraces and in general we should have an API like `resolveQualifiedNameString(path: String) -> RustPathNamedElement`. 


> it only works for pure and associated functions.

Yes, and this is a bug I think.

alygin:
OK, then I'll try to tackle the bug and move the function that searches a `RustPathNamedElement` into a utility class (into `RustResolveUtil`?).

matklad:
Into `RustResolveEngine`. `RustResovleUtil` should ideally be dead, but I can;t

[...]

PR #827, opened 2016-12-11
(MISC): Add source code links to stack backtraces

alygin:
Work is still in progress on this feature (#821), but it's already pretty functional.

I'm creating the PR to have an early feedback on the general approach and on some parts of the code related to elements resolving and working with cargo project metadata. Those of particular interest are marked with TODO's.

@matklad, could you, please, have a look at it?

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -26,7 +26,6 @@ class CargoProjectDescription private constructor(
         val dependencies: List<Package>,
         val isWorkspaceMember: Boolean
     ) {
-        val isModule: Boolean get() = source == null
         val libTarget: Target? get() = targets.find { it.isLib }
         val contentRoot: VirtualFile? get() = VirtualFileManager.getInstance().findFileByUrl(contentRootUrl)
 
@@ -120,6 +119,8 @@ class CargoProjectDescription private constructor(
     }
 
     val hasStandardLibrary: Boolean get() = findExternCrateRootByName(AutoInjectedCrates.std) != null
+
+    fun findPackage(name: String): Package? = packages.find { it.name == name }
 
     companion object {
         fun deserialize(data: CleanCargoMetadata): CargoProjectDescription? {
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -27,8 +27,8 @@ class CargoProjectDescription private constructor(
     ) {
         val isModule: Boolean get() = source == null
         val libTarget: Target? get() = targets.find { it.isLib }
-
         val contentRoot: VirtualFile? get() = VirtualFileManager.getInstance().findFileByUrl(contentRootUrl)
+        val isExternal: Boolean get() = targets.asSequence().filter { it.crateRoot != null }.any { it.isLib }
 
         override fun toString(): String = "Package(contentRootUrl='$contentRootUrl', name='$name')"
     }
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -23,12 +23,12 @@ class CargoProjectDescription private constructor(
         val version: String,
         val targets: Collection<Target>,
         val source: String?,
-        val dependencies: List<Package>
+        val dependencies: List<Package>,
+        val isExternal: Boolean
     ) {
         val isModule: Boolean get() = source == null
         val libTarget: Target? get() = targets.find { it.isLib }
         val contentRoot: VirtualFile? get() = VirtualFileManager.getInstance().findFileByUrl(contentRootUrl)
-        val isExternal: Boolean get() = targets.asSequence().filter { it.crateRoot != null }.any { it.isLib }
 
         override fun toString(): String = "Package(contentRootUrl='$contentRootUrl', name='$name')"
     }
@@ -112,7 +112,8 @@ class CargoProjectDescription private constructor(
                 version = "",
                 targets = listOf(Target(crateRoot.url, name = crateName, kind = TargetKind.LIB)),
                 source = null,
-                dependencies = emptyList()
+                dependencies = emptyList(),
+                isExternal = true
             )
         }
         return CargoProjectDescription(packages + stdlibPackages)
@@ -136,7 +137,8 @@ class CargoProjectDescription private constructor(
                     pkg.version,
                     pkg.targets.map { Target(it.url, it.name, it.kind) },
                     pkg.source,
-                    deps
+                    deps,
+                    pkg.isExternal
                 ) to deps
             }.unzip()
 
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -24,7 +24,7 @@ class CargoProjectDescription private constructor(
         val targets: Collection<Target>,
         val source: String?,
         val dependencies: List<Package>,
-        val isExternal: Boolean
+        val isWorkspaceMember: Boolean
     ) {
         val isModule: Boolean get() = source == null
         val libTarget: Target? get() = targets.find { it.isLib }
@@ -113,7 +113,7 @@ class CargoProjectDescription private constructor(
                 targets = listOf(Target(crateRoot.url, name = crateName, kind = TargetKind.LIB)),
                 source = null,
                 dependencies = emptyList(),
-                isExternal = true
+                isWorkspaceMember = false
             )
         }
         return CargoProjectDescription(packages + stdlibPackages)
@@ -138,7 +138,7 @@ class CargoProjectDescription private constructor(
                     pkg.targets.map { Target(it.url, it.name, it.kind) },
                     pkg.source,
                     deps,
-                    pkg.isExternal
+                    pkg.isWorkspaceMember
                 ) to deps
             }.unzip()
 
--- a/src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
@@ -132,6 +132,10 @@ private class RustBacktraceItemFilter(
             .getAttributes(TextAttributesKey.createTextAttributesKey("org.rust.DIMMED_TEXT"))!!
         val GRAYED_LINK_COLOR = Color(135, 135, 135)
         val GRAYED_LINK = TextAttributes(GRAYED_LINK_COLOR, null, GRAYED_LINK_COLOR, EffectType.LINE_UNDERSCORE, Font.PLAIN)
-        val SKIP_PREFIXES = arrayOf("std::", "core::")
+        val SKIP_PREFIXES = arrayOf(
+            "std::rt::lang_start",
+            "std::panicking",
+            "std::sys::backtrace",
+            "core::panicking")
     }
 }
--- a/src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
@@ -5,6 +5,8 @@ import com.intellij.execution.filters.OpenFileHyperlinkInfo
 import com.intellij.execution.filters.RegexpFilter
 import com.intellij.openapi.editor.colors.EditorColorsManager
 import com.intellij.openapi.editor.colors.TextAttributesKey
+import com.intellij.openapi.editor.markup.EffectType
+import com.intellij.openapi.editor.markup.TextAttributes
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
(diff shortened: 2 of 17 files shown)


alygin:
Looks like IDEA 15 had no `grayedHyperlynk` support and I have to paint them manually.

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,137 @@
...
-) : RegexpFileLinkFilter(
-    project,
-    cargoProjectDir,
-    "^\\s+at ${RegexpFilter.FILE_PATH_MACROS}:${RegexpFilter.LINE_MACROS}$") {
+) : Filter {
+    private val sourceLinkFilter = RegexpFileLinkFilter(project, cargoProjectDir, "^\\s+at ${RegexpFilter.FILE_PATH_MACROS}:${RegexpFilter.LINE_MACROS}$")
+    private val backtraceItemFilter = RustBacktraceItemFilter(project)
+
+    override fun applyFilter(line: String, entireLength: Int): Filter.Result? {
+        backtraceItemFilter.applyFilter(line, entireLength)?.let { return it }
```
matklad: You can chain stuff like this with `?:` operator more concisely:

```
return backtraceItemFilter.applyFilter()
    ?: sourceLinkFilter.applyFilter()
```

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,137 @@
...
+private class RustBacktraceItemFilter(
+    val project: Project
+) : Filter {
+    private val pattern = Pattern.compile("^(\\s*\\d+:\\s+0x[a-f0-9]+ - )(.+)(::h[0-9a-f]+)$")!!
+    private val docManager = PsiDocumentManager.getInstance(project)
+
+    override fun applyFilter(line: String, entireLength: Int): Filter.Result? {
+        val matcher = pattern.matcher(line)
+        if (!matcher.find()) return null
+        val funcName = matcher.group(2)
```
matklad: Small stylistic not, but what do you think about binding all groups in the same place, like

```
val funcPath = matcher.group(1)
val funcName = matcher.group(2)
val funcHash = matcher.group(3)
```

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,137 @@
...
+import org.rust.lang.core.symbols.RustPathSegment
+import java.util.*
+import java.util.regex.Pattern
 
 /**
- * Detects messages about panics and adds source code links to them.
+ * Adds features to stack backtraces:
+ * - Wrap function calls into hyperlinks to source code.
+ * - Turn source code links into hyperlinks.
+ * - Dims funcion hashcodes to remove noise.
```
matklad: s/funcion/function.

Has you re-enabled the spellchecker? :)

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,137 @@
...
+//            "std::sys::backtrace::",
+//            "std::panicking::",
+            "std::rt::lang_start"
+        )
+    }
+}
+
+
+private object ElementResolver {
+    private val vfm = VirtualFileManager.getInstance()
```
matklad: This filed of an object (~ static field) seems a bit dangerous. It is lazily evaluated the first time this object is accessed, while `VirtualFileManager` is available only after the application has been initialized. Let's just query it inside the method.

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,137 @@
...
+    fun resolve(pathStr: String, project: Project) : Result? {
+        val segments = pathStr.segments
+        if (segments.isEmpty()) return null
+        val pkg = project.getPackage(segments[0].name) ?: return null
+        val path = RustPath(RustPathHead.Absolute, segments.drop(1))
+        val el = pkg.targets
+            .mapNotNull { vfm.findFileByUrl(it.crateRootUrl) }
+            .mapNotNull { project.getPsiFor(it) as? RustCompositeElement }
+            .flatMap { RustResolveEngine.resolve(path, it) }
+            .mapNotNull { it as? RustNamedElement }
```
matklad: This is `filterIsInstance`.

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,137 @@
...
+
+        return Result(el, pkg)
+    }
+
+    private fun Project.getPackage(name: String): CargoProjectDescription.Package? =
+        modulesWithCargoProject
+            .mapNotNull { it.cargoProject }
+            .flatMap { it.packages }
+            .filter { it.isModule && it.name == name }
+            .firstOrNull()
```
matklad: `.filter {} .firstOrNull()` is `find`

matklad:
Overall looks good! We definitely need some tests here though!

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
```diff
@@ -27,8 +27,8 @@ class CargoProjectDescription private constructor(
     ) {
         val isModule: Boolean get() = source == null
         val libTarget: Target? get() = targets.find { it.isLib }
-
         val contentRoot: VirtualFile? get() = VirtualFileManager.getInstance().findFileByUrl(contentRootUrl)
+        val isExternal: Boolean get() = targets.asSequence().filter { it.crateRoot != null }.any { it.isLib }
```
alygin: The idea here is to produce regular hyperlinks for the source code we control (it's in our workspace) from the source code we don't (external dependencies). It doesn't depend on the particular target we're currently running. We produce regular hyperlinks for our own code and grayed ones for everything else. This is how this feature works in Java.

`isExternal` seems to work pretty reliably now for the described task, while `isModule` returns `true` for both our own crates and external dependencies.

But maybe things are more complicated than I think and I don't understand your point? We could look at a particular example that you think won't work.

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
alygin: Oh, sure, thanks!

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
alygin: Yeah, I have. But only now :)

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
alygin: Sure

alygin:
Sure. Some of my tests are broken now, so I decided not to push them so far.

alygin:
And what about moving things to places they fit better? Would be `RustResolveEngine` a good place for `resolve(String, Project)`?

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
matklad: >The idea here is to separate the source code we control (it's in our workspace) from the source code we don't (external dependencies).

Yes, this is exactly what I want. I think that this implementation of `isExternal` won't work always, and it is quite possible that `isModule` is broken as well. 

There's no need to make it super correct right now, it can always be fixed later. 

The problem with `external == any target is a library` happes when, for example, you are implementing the library. 

If you have

```
src/
  main.rs
  lib.rs
```

and `main.rs` panics

matklad:
>Would be RustResolveEngine a good place for resolve(String, Project)?

Absolutely!

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
alygin: Thanks for pointing that out! I'll have a look at it.

alygin:
Could you, please, help me a little with converting my code, so it could work after the refactoring? I use the following lines to resolve absolute paths:

```rust
val rustPath = RustPath(RustPathHead.Absolute, segments.drop(1))
// ...
RustResolveEngine.resolve(rustPath, pivot)
```

Here, `segments` represents an absolute path, and `pivot` is the crate root module.

While merging it with the latest changes, I tried transforming the `rustPath` into:

```rust
val rustPath = RustPath.CrateRelative(segments.drop(1))
```

and

```rust
val rustPath = RustPath.Named(segments[0], segments.drop(1))
```

The one that uses `CrateRelative` works for some paths, but hangs on [these lines](https://github.com/intellij-rust/intellij-rust/blob/ff84d8715b62c7061f1eb0c26f2d0fe939aaf759/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L38-L40) for others (for instance, it hangs on resolving `std::sys::backtrace::tracing::imp::write`).

The one with `Named` just never resolves things because `resolve(path, pivot)` always returns an empty list for the described arguments.

What would be the right way to resolve absolute paths now?

matklad:
Hm, I believe the refactoring should not have changed the resolve algorithm. 

I've just merged `RustPath` and `RustPathHead` classes (https://github.com/intellij-rust/intellij-rust/commit/17519142e544b93049df7974098e55bb4cff7aff). 

Are you sure that this worked differently before? 

>for instance, it hangs on resolving 

Yeah, in general resolve functions can end up in the infinite loop, if the original source code contains reference cycle. So we explicitly guard against such cases:

https://github.com/intellij-rust/intellij-rust/blob/ff84d8715b62c7061f1eb0c26f2d0fe939aaf759/src/main/kotlin/org/rust/lang/core/resolve/ref/RustReferenceBase.kt#L25

Also see `RecursionGuard` class in the platform.

Although `std::sys::backtrace::tracing::imp::write` should resolve just fine.

TLDR, I don't really understand what's going on here, will take a closer look tomorrow!

alygin:
Yes, it resolved everything inside `std::`, `core::` and our own packages. There were (and still are) some problems with resolving things in external dependencies, but they might be caused by the way I attached the sources of those external crates (by means of path rewriting), I haven't dig deep here yet.

Now, things in `core::` and in our own packages are resolved correctly (when I use `RustPath.CrateRelative`), while all attempts to resolve something within `std::` just hang. I'll also try to debug such calls when time allows.

src/main/kotlin/org/rust/cargo/runconfig/RustBacktraceFilter.kt
```diff
@@ -1,17 +1,185 @@
...
+        val segments = pathStr.segments
+        if (segments.isEmpty()) return null
+        val pkg = project.getPackage(segments[0].name) ?: return null
+        val vfm = VirtualFileManager.getInstance()
+        val path = RustPath(RustPathHead.Absolute, segments.drop(1))
+        val el = pkg.targets.asSequence()
+            .mapNotNull { vfm.findFileByUrl(it.crateRootUrl) }
+            .mapNotNull { project.getPsiFor(it) as? RustCompositeElement }
+            .flatMap { RustResolveEngine.resolve(path, it).asSequence() }
+            .filterIsInstance(RustNamedElement::class.java)
```
matklad: You can call it as `filterIsInstance<RustNamedElement>()`

matklad:
So, changing paths is irrelevant here, here's the commit that must have introduced problems: https://github.com/intellij-rust/intellij-rust/commit/ff84d8715b62c7061f1eb0c26f2d0fe939aaf759

I've forgot that `RustFile` is `RustCompositeElement` =/ Not sure what is the best course of action here...

matklad:
Looks like making `containingMod` not null is in general a bad idea: there's no containing module for the topmost one...

alygin:
Several changes:
- `resolve()` method moved to `RustResolveEngine`
- `Package.isExternal` is now set based on the cargo's `workspace_members` parameter
- tests added

alygin:
Changed `Package.isExternal` to `Package.isWorkspaceMember` to follow cargo's metadata naming.

src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt
```diff
@@ -159,6 +182,17 @@ object RustResolveEngine {
             else
                 current.parentOfType<RustResolveScope>()
         }
+
+    private fun Project.getPackage(name: String): CargoProjectDescription.Package? =
```
matklad: This should not be a `Project` extension. In intellij, the global context for a "project" is actually a `Module`. You can have two completely different Cargo projects added as different intellij modules in the same IDEA project.

More specifically, the IDEA itslf allows you to create several `Modules` within a single project, while IDEs like PyCharm generally create only a single `Module` for a `Project`.

So, let's make this a method of `CargoProjectDescription` (which is properly attached to a module) and make it a caller's responsibility to retrieve the necessary module. The module is available in `CargoCommandConfiguration`: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/runconfig/CargoCommandConfiguration.kt#L33, you'll need to thread it to the filter.

alygin:
There's only [one usage](https://github.com/intellij-rust/intellij-rust/blob/2b92a00f5b52f84190694f0e200be41da4a76b8e/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceImpl.kt#L140) of `Package.isModule` left in the project. Would it be OK to substitute it to the new `Package.isWorkspaceMember` property and get rid of `isModule` completely?

matklad:
>Would it be OK to substitute it to the new Package.isWorkspaceMember property and get rid of isModule completely?

Totally! `isModule` is there only because cargo had not had workspaces when this feature was implemented in the plugin. `isWorspaceMember` is a more correct solution!

matklad:
Great! What do you think about enable backtracks by default?

alygin:
I'm totally positive about switching them on by default. I always have to spend that extra mouse click per every run configuration )

PR #828, opened 2016-12-11
(MISC): Fix color argument position

alygin:
Fixes the `--color=always` position [problem](https://github.com/intellij-rust/intellij-rust/issues/646#issuecomment-266288142).


matklad:
Thanks!

PR #832, opened 2016-12-12
(TEMPL): parentheses and lambda

ujpv:


src/test/kotlin/org/rust/ide/template/postfix/LambdaPostfixTemplateTest.kt
```diff
@@ -0,0 +1,28 @@
+package org.rust.ide.template.postfix
+
+/**
+ * Created by chainic-vina on 12.12.16.
```
matklad: :) We try to avoid authorship comments :)

ujpv: I added it unintentionally. :)

matklad:
Thanks!

Issue #769, opened 2016-11-08
Intentions to un-elide lifetimes


matklad:
Go from

```rust
fn f(&self) -> &str 
```

to

```rust
fn f<'a>(&'a self) -> &'a str
```

matklad:
There's preliminary support for this feature (that is, it does not always do the right thing), would be nice to complete it!

matklad:
In particular, current implementation does not handle generics, it can't turn this

```
struct S<'a> { x: &'a i32 }

fn make_s(x: &i32) -> S { unimplemented!() }
```

into this

```
struct S<'a> { x: &'a i32 }

fn make_s<'a>(x: &'a i32) -> S<'a> { unimplemented!() }
```

PR #833, opened 2016-12-14
(INT): Un-elide lifetimes

ujpv:
Issue #769


matklad:
Seems like a good start! There are a lot of more tough cases though...

Lifetimes may be elided in generics:

```
struct S<'a> { x: &'a i32 }

fn make_s(x: &i32) -> S { unimplemented!() }
fn make_s_unelided<'a>(x: &i32) -> S { unimplemented!() }
```

Also, `self` parameters receive special treatment.

PR #837, opened 2016-12-15
Disable inline macro use by default

### LINKED ISSUE #834: Formatter inlines [#macro_use] on crates and functions
Before:
```rust
#[macro_use(foo, bar)]
extern crate baz;

#[macro_use]
mod baa;
```

After formatter:
```rust
#[macro_use(foo, bar)] extern crate baz;

#[macro_use] mod baa;
```

While I couldn't find anything specific about this case in the [Rust style guide](https://aturon.github.io/style/whitespace.html), I does seem contrary to how examples are used in [the book](https://doc.rust-lang.org/beta/book/macros.html#scoping-and-macro-importexport).

johnthagen:
Closes #834.

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
@@ -190,13 +190,6 @@ fun Block.computeSpacing(child1: Block?, child2: Block, ctx: RustFmtContext): Sp
     return ctx.spacingBuilder.getSpacing(this, child1, child2)
 }
 
-/**
- * Attribute name -> parent PSI classes in which this attribute should be placed in-line.
- */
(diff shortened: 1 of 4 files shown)


mkaput:
Maybe delete whole feature if it's against RFCs, what do you think?

matklad:
Yeah, let's just remove it altogether!

johnthagen:
@mkaput @matklad I'm not super familiar with the code base, could you give me a short run down on what needs to be removed?

matklad:
@johnthagen you need just to remove all usages of this constant: https://github.com/intellij-rust/intellij-rust/search?utf8=%E2%9C%93&q=INLINE_MACRO_USE_ATTR

matklad:
Thanks!

PR #839, opened 2016-12-15
(TEMPL): Let postfix template

ujpv:


mkaput:
I think it would be nice to add `var` alias for this.

matklad:
>I think it would be nice to add var alias for this.

Then `var` for `let mut`, `val`/ `let` for `let`.

matklad:
Let's merge this with just let and add aliases later, I need to introduce a variable right now!

PR #842, opened 2016-12-18
Fix logging macros parsing

HeyLey:
Fix parsing problem with trace macro

```
trace!(target: "smbc", "open_with {:?}", options);
```


src/main/grammars/RustParser.bnf
```diff
@@ -1102,6 +1098,9 @@ item_macro_arg ::= '(' token_trees ')' ';'
 try_macro_args ::= <<macro_args_delim any_expr>>
 vec_macro_args ::= '[' expr_list ']' | '[' any_expr';' any_expr ']' | '[' ']'  { pin = 1 }
 
+private zz_log_macro_args ::= ('target' ':' expr ','?)? (<<comma_separated_list format_macro_arg>>)?
```
matklad: This will work only for this particular logging macro. I think we should fallback to parsing token trees if we fail to parse a `format_macro_arg`. That is, even the crazy stuff like `trace!({} <=> +++)` should not result in a parsing error, while usual invocations like `trace!("Bla bla {bar}", bar = some_expression)` should be parsed as format-like macros.

HeyLey: From the definition of trace https://doc.rust-lang.org/log/log/macro.log.html this is normal usage.
But I can add `token trees` fallback as well.

matklad: Indeed! Thanks for the clarification!

matklad:
Thanks!

PR #843, opened 2016-12-18
(COMPL): Add parentheses when autocomplete functions/methods

alygin:
Adds parentheses to functions/methods on autocompletion.

If the function has arguments, the cursor is positioned between parentheses, otherwise it moves after the closing parenthesis.

Parentheses are not added when the completion is performed within a `use` block.


matklad:
:+1:

PR #844, opened 2016-12-20
(MISC): Coloring type aliases

### LINKED ISSUE #830: Add coloring option for type aliases
There doesn't seem to be a way to color [type aliases](https://doc.rust-lang.org/book/type-aliases.html), for example:

```rust
type Name = String;

let x: Name = "Hello".to_string();
```

I would think a good default color would be the same as `struct` or `enum` since it's just an alias to a type.

alygin:
Adds a setting for customizing type aliases color. The default value is the platform's class name color.

Fixes #830


matklad:
Thanks!

PR #846, opened 2016-12-21
(MISC): Color customization terms cleanup

alygin:
Brings element names that are used in the color customization panel into line with the Rust lingo (used in the official documentation). Also provides an example of an associated function call in the demo source code.

I would also rename some of the `RustColor` variant names accordingly, but that will reset colors for those users who have customized them and there're no simple way (at least, I didn't find it) to just copy values from the old settings. So, I'm not sure if it's ok to do so.

### DIFF
--- a/src/main/kotlin/org/rust/ide/colors/RustColors.kt
+++ b/src/main/kotlin/org/rust/ide/colors/RustColors.kt
@@ -9,10 +9,10 @@ import com.intellij.openapi.editor.DefaultLanguageHighlighterColors as Default
  */
 enum class RustColor(humanName: String, val default: TextAttributesKey) {
     IDENTIFIER            ("Identifier",                  Default.IDENTIFIER),
-    FUNCTION_DECLARATION  ("Function declaration",        Default.FUNCTION_DECLARATION),
-    INSTANCE_METHOD       ("Instance method declaration", Default.INSTANCE_METHOD),
-    STATIC_METHOD         ("Static method declaration",   Default.STATIC_METHOD),
-    PARAMETER             ("Function parameter",          Default.PARAMETER),
+    FUNCTION              ("Function",                    Default.FUNCTION_DECLARATION),
+    METHOD                ("Method",                      Default.INSTANCE_METHOD),
+    ASSOC_FUNCTION        ("Associated function",         Default.STATIC_METHOD),
+    PARAMETER             ("Parameter",                   Default.PARAMETER),
     SELF_PARAMETER        ("Self parameter",              Default.KEYWORD),
     Q_OPERATOR            ("? operator",                  Default.KEYWORD),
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/colors/RustColors.kt
```diff
@@ -9,11 +9,11 @@ import com.intellij.openapi.editor.DefaultLanguageHighlighterColors as Default
...
 enum class RustColor(humanName: String, val default: TextAttributesKey) {
     IDENTIFIER            ("Identifier",                  Default.IDENTIFIER),
-    FUNCTION_DECLARATION  ("Function declaration",        Default.FUNCTION_DECLARATION),
-    INSTANCE_METHOD       ("Instance method declaration", Default.INSTANCE_METHOD),
-    STATIC_METHOD         ("Static method declaration",   Default.STATIC_METHOD),
-    PARAMETER             ("Function parameter",          Default.PARAMETER),
-    SELF_PARAMETER        ("Self parameter",              Default.KEYWORD),
+    FUNCTION_DECLARATION  ("Function",                    Default.FUNCTION_DECLARATION),
+    INSTANCE_METHOD       ("Method",                      Default.INSTANCE_METHOD),
+    STATIC_METHOD         ("Associated function",         Default.STATIC_METHOD),
```
matklad: A nice change! I think we should use "associated" instead of "static" everywhere in the code.

src/main/kotlin/org/rust/ide/colors/RustColors.kt
```diff
@@ -9,11 +9,11 @@ import com.intellij.openapi.editor.DefaultLanguageHighlighterColors as Default
...
+    PARAMETER             ("Function argument",           Default.PARAMETER),
```
matklad: Let's change "argument" to 'parameter". This is not really cast in stone, but usually when people refer to "parameter", they mean the formal parameters of function *declaration* `fn foo(i_am_a_format_parameter: i32)`. And "argument" is usually the actual value that is passed to the function *call* `foo(1 + 2 + 3)`.

src/main/kotlin/org/rust/ide/colors/RustColors.kt
matklad: >I think we should use "associated" instead of "static" everywhere in the code.

It's just a general comment and a note to myself, there's no need to fix everything in this PR

src/main/kotlin/org/rust/ide/colors/RustColors.kt
alygin: Hm, interesting, I never thought of arguments/parameters in this way.

It seems that the book (and the error index as well) is quite definite about using "argument" when talking [about functions](http://rust-lang.github.io/book/ch03-03-how-functions-work.html) and "parameter" when talking [about generic parameters](https://doc.rust-lang.org/book/generics.html) with not a single exception, while `rustc` emits messages that contain both terms and don't follow any rules about them.

I'm not an adherent of any of these terms, but the "parameter" seems to contradict even the very next "Self argument" (though named SELF_PARAMETER internally) element. Should it be "Self parameter"?

matklad: >Hm, interesting, I never thought of arguments/parameters in this way.

Yep, this terminological distinction really matters only when you "implement" a language, not when you use it. But it is a pretty well established terminology: 

https://en.wikipedia.org/wiki/Parameter_(computer_programming)#Parameters_and_arguments

It's definitely helps to use the precise term in the code (there are `RustParametersElement` and `RustArgListElement`). 


It's not that important for user-facing names, but I'd still prefer parameter, because Kotlin and Java use this name in color settings as well.

alygin: Thanks for the link, that's interesting. Given the fact that the "parameter" term is everywhere in the IDE (like the "Parameter Info" command), I agree that it makes sense to stick to it despite the fact that it's used in the Rust documentation differently.

matklad:
Thanks!

PR #850, opened 2016-12-23
(RUN): New backtrace write function location in Rust 1.14

alygin:
In Rust 1.14, the function that prints stack backtraces has been moved to another module. The new location is added to the `RustBacktraceFilter`'s skip list.


matklad:
Thanks!

PR #852, opened 2016-12-23
(COMPL): Enum variants autocompletion improvement

alygin:
Adds parentheses / curly braces to enum variants with tuple / block fields on autocompletion.


src/main/kotlin/org/rust/lang/core/completion/RustCompletionEngine.kt
```diff
@@ -119,3 +127,9 @@ private fun RustPath.dropLastSegment(): RustPath {
         is RustPath.Named -> RustPath.Named(head, segments)
     }
 }
+
+private fun InsertionContext.isInUseBlock(): Boolean {
+    val element = file.findElementAt(startOffset - 1)!!
+    return (element.parentOfType<RustUseGlobListElement>() != null)
+            || (element.parentOfType<RustUseItemElement>() != null)
```
matklad: Hh, the second check seems strictly more general than the first one.

alygin: You're right, it should be simplified.

matklad:
Nice!

PR #853, opened 2016-12-24
(PROJ): Introduce PackageOrigin

alygin:
`Package.isWorkspaceMember` is substituted for `Package.origin`, which allows to make distinction between stdandard libraries, workspace packages, direct dependencies and transitive dependencies.

I've also removed `Package.dependencies` since it has always been empty and none used it.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -3,7 +3,6 @@ package org.rust.cargo.project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
 import org.rust.cargo.toolchain.impl.CleanCargoMetadata
-import org.rust.cargo.util.AutoInjectedCrates
 import java.util.*
 import java.util.concurrent.atomic.AtomicReference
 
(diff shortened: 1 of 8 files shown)


matklad:
>since it has always been empty and none used it.

That does seem strange! Though it's OK to remove it for now, some day we'll want to look into dependencies.

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
```diff
@@ -102,7 +100,7 @@ class CargoProjectDescription private constructor(
 
     fun isCrateRoot(file: VirtualFile): Boolean = findTargetForCrateRootFile(file) != null
 
-    fun withAdditionalPackages(additionalPackages: Collection<Pair<String, VirtualFile>>): CargoProjectDescription {
+    fun withAdditionalPackages(additionalPackages: Collection<Pair<String, VirtualFile>>, origin: PackageOrigin): CargoProjectDescription {
```
matklad: This function has only usage, inside a function that has only one usage... Perhaps we can rename this to `withStdlib`, remove `origin` argument and maybe even kill `extendProjectDescriptionWithStandardLibraryCrates` function?

alygin: Totally agree! I thought there was some bigger idea behind this function, but if it's that simple, it's better to do as you said.

alygin: Though I'm not sure on killing `extendProjectDescriptionWithStandardLibraryCrates`. It uses `project` which is not available inside `CargoProjectDescription`, so there still has to be something that will call `withStdLib()` from outside and provide it with a `Library`.

matklad: Perhaps the work done by the `ePDWSLC` can be moved to `Module.cargoProject` extension?

alygin: There're some other dependencies within `ePDWSLC` that must be moved (or removed). I'll get back to it shortly and try to figure out how to do it properly.

alygin:
@matklad, what about this last refactoring?

matklad:
LGTM!

PR #854, opened 2016-12-25
(COMPL): Keywords completion contributor

alygin:
Adds a framework for implementing keywords completion. Several examples are provided to show how it works: `extern crate`, `use`, `let` and `return` completions are implemented for the most straightforward cases. So, it's more a proof of concept and should be extended and improved in the future.

One of the architectural changes is introduction of the `RustPsiPattern` class that can be used as a source of predefined high-level `PsiElementPattern`s for completions, intentions, inspections and other tools that rely on code patterns analysis.


src/main/kotlin/org/rust/lang/core/RustPsiPattern.kt
```diff
@@ -0,0 +1,35 @@
...
+    class StatementBeginning : PatternCondition<PsiElement>("on statement beginning") {
+        override fun accepts(t: PsiElement, context: ProcessingContext?): Boolean {
+            val prev = t.prevVisibleOrNewLine ?: return true
+            if (prev is PsiWhiteSpace) return true
+            return prev.node.elementType in STATEMENT_BOUNDARIES
+        }
+    }
+
+    private companion object {
+        val STATEMENT_BOUNDARIES = listOf(RustTokenElementTypes.SEMICOLON, RustTokenElementTypes.LBRACE, RustTokenElementTypes.RBRACE)
```
matklad: It's better to use `TokenSet` here.

src/main/kotlin/org/rust/lang/core/RustPsiPattern.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.intellij.psi.PsiErrorElement
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.util.ProcessingContext
+import org.rust.lang.core.psi.RustTokenElementTypes
+
+/**
+ * Rust PSI tree patterns.
+ */
+class RustPsiPattern {
```
matklad: There's a bunch of patterns in `RustAttributeCompletionProvoder`, and they have some nice tests at `RustAttributeCompletionPatternTest`.

matklad: Although majority of test are integrated, it makes sense to add "unit tests" for patterns specifically, because they are very sensitive to PSI strucutre, that changes oftentimes.

alygin: Great, I didn't notice that. Though I think it would be better to move those patterns into a more general class than `AttributeCompletionProvider` if we want to reuse them. What do you think about `RustPsiPattern`? Would it be a suitable place for them?

I don't yet know how to properly unit test them in isolation, but I'll figure something out.

matklad: `RustAttributeCompletionPatternTest` tests patterns in isolation. 

>What do you think about RustPsiPattern? Would it be a suitable place for them?

Sounds reasonable!

alygin:
1. PSI patterns from `AttributeCompletionProvider` moved to `RustPsiPattern`
2. Patterns unit tests moved to `RustPsiPatternTest` and expanded
3. Completion tests added for `AttributeCompletionProvider`
4. Keywords `enum`, `struct`, `trait` and `type` added to the completion list

The current implementation of the patterns moved from `AttributeCompletionProvider` looks very specific to attributes matching. Some refactoring will surely be needed if they are expected to become general patterns. I think I might do that a bit later, along with keywords completion fine tuning.

matklad:
LGTM! There are a couple of minor nits, like a missing package directive, I'll fix them myself to save on a round trip time.

PR #860, opened 2016-12-27
(COMPL): Fix hyphenated crate names completion

alygin:
Hyphens in crate names are replaced with underscores on autocompletion.


matklad:
Thanks!

PR #869, opened 2016-12-30
(COMPL): Keywords completion improvement

alygin:
1. New keywords added: `crate`, `extern`, `fn`, `impl`, `mod`, `mut`, `pub`, `unsafe`, `use`.
2. Each part of a keywords phrase is now completed separately. For instance, the phrase `pub unsafe fn` completes word by word, each completion variant depends on the preceding keyword.
3. Many keywords (`enum`, `struct`, `use` etc) are now also completed within functions/methods.


matklad:
Nice!

PR #870, opened 2016-12-30
(COMPL): Fix icon for stdlib module in autocompletion

alygin:
Stdlib modules now use the `RUST_FILE` icon on autocompletion. This fix rewrites it to `MODULE`, the same icon as other modules use.


matklad:
Yep, that makes sense! 

An alternative is to change the icon of the file itself. The same way in Java class icon is usually shown for a file. Perhaps we can even use a different icon for `mod.rs` / `lib.rs` / `main.rs` files, so that it's easy to notice which files are crate roots.

alygin:
I agree, but I think this will make some side effects somewhere else. For instance, in the Project Structure view. We need either provide the correct module name for the current file there, or hide it completely. Maybe, hiding is the right choice, other plugins don't show the file at the top level in the Project Structure. I decided not to change other things so far, so made a local fix for autocompletion.

PR #872, opened 2016-12-31
Fix prelude resolve in child modules

himikof:


matklad:
Thanks a lot! I hope that after the last refactoring resolve code is a bit easier to follow!

PR #873, opened 2016-12-31
(COMPL): Attributes completion improvement

alygin:
A couple of improvements of attributes completion:
1. Add parenthesis to attributes that require additional arguments.
2. Add icon to attributes (the same as used for annotations in Java/Kotlin) in completion list.


matklad:
Thanks!

PR #878, opened 2017-01-01
(COMPL): Autocomplete break and continue keywords

alygin:
Autocompletion for the `break` and `continue` keywords inside loops.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/RustPsiPattern.kt
+++ b/src/main/kotlin/org/rust/lang/core/RustPsiPattern.kt
@@ -1,6 +1,7 @@
 package org.rust.lang.core
 
 import com.intellij.patterns.*
+import com.intellij.patterns.StandardPatterns.or
 import com.intellij.patterns.PlatformPatterns.psiElement
 import com.intellij.psi.*
 import com.intellij.psi.tree.TokenSet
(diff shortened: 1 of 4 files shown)


alygin:
Tests a broken because of the backtraces setting rollback.

src/main/kotlin/org/rust/lang/core/RustPsiPattern.kt
```diff
@@ -80,6 +81,12 @@ object RustPsiPattern {
     val onTestFn: PsiElementPattern.Capture<PsiElement> = onItem(psiElement<RustFunctionElement>()
         .withChild(psiElement<RustOuterAttrElement>().withText("#[test]")))
 
+    val inAnyLoop: PsiElementPattern.Capture<PsiElement> =
+        psiElement().inside(true, psiElement<RustBlockElement>().withParent(or(
```
matklad: One possible improvement in performance and correctness is to try to stop earlier when searching for parent. That is, `break` should not be suggested inside a lambda inside a loop.

matklad:
LGTM, I've restarted the travis build

src/main/kotlin/org/rust/lang/core/RustPsiPattern.kt
alygin: Sure, you're right. I'll add a proper `stopAt`.

matklad:
Thanks

PR #884, opened 2017-01-02
(T): Move tests from fixture-files to test classes

alygin:
Test source code for spellchecking, goto and some intentions moved from fixture-files to test classes.

There's a formatting issue in `AddDeriveIntentionTest`, but according to the `FIXME`-comment, it was there before.


matklad:
Thanks a lot!

PR #887, opened 2017-01-03
(DOC): Fix typos

alygin:
Some small typos are fixed.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -160,15 +160,15 @@ currently opened in the editor and its dependencies, ignoring most of the
 crates. The major chunk of resolve is handled by the `innerDeclarations` and
 `outerDeclarations` functions. 
 
-`innerdeclarations` takes a reference and walks the AST tree upwords, collecting
-all the let declarations, function parameters, item definitions and other
+`innerDeclarations` takes a reference and walks the AST tree upwards, collecting
+all the `let` declarations, function parameters, item definitions and other
 "local" declarations. This is the function which does the work of resolving a
 local variable or a first component `foo` of the `foo::bar::baz` path.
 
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -189,8 +189,8 @@ grandparent module. This restriction allows fast recalculation of the index:
 only the part corresponding to the changed files needs to be flushed.
 
 These indexes power go to class and go to symbol functionality. They are also
-used during resolve to find parent module for a file and to get the list of
-impls for a type.
+used during resolve to find the parent module for a file and to get the list of
```
mkaput: this `the` is incorrect imho, but I'm not professional english-speaker 😉

alygin: Hm, neither am I, but I'm pretty sure it must be there. It's a countable, non-abstract noun, so it can't just sit without an article. And there's only one parent module for a file, so it's `the`. I'll go find some examples to prove or rebut it :)

alygin: [Here](http://www.ibm.com/support/knowledgecenter/SSNW2F_5.2.1/com.ibm.p8.security.doc/p8psa074.htm)'s IBM talking about _**the** parent of a subfolder_.
[Here](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html)'s Oracle talking about _**the** parent of a `ClassLoader` instance_.
[Here](https://msdn.microsoft.com/en-us/windows/hardware/drivers/install/determining-the-parent-of-a-device)'s Microsoft talking about _**the** parent of a device_.

mkaput: ok

matklad:
Thanks!

PR #888, opened 2017-01-03
(COMPL): Complete self and super after use

alygin:
Autocomplete `self::` and `super::` after `use`.


matklad:
Hm, use declarations are indeed special!

If you want to add `self`/`super` completion in function:

```
fn foo() {
    self::foo
   //^ this self should be completed. 
}
```

you need to do it the same way `Self` is added: https://github.com/intellij-rust/intellij-rust/blob/42a36443ddca9d9041e2ada41cc30f01c309d5b5/src/main/kotlin/org/rust/lang/core/resolve/RustResolveEngine.kt#L284

matklad:
Thanks!

Issue #864, opened 2016-12-29
Use annotators instead of duplicating grammar rules


matklad:
In out `RustParser.bnf` file, there's a pattern repeated thrice, for `function`, `constant` and `type_alias`. 

A `fake` rule is defined to describe a single PSI element, and then a number of implementations (`elementType = bla-bla-bla`) of this rules are provide. The goal of this implementations is to enforce small syntactic differences. For example, top level functions always have body, but functions in interfaces can just end in `;`. 

A better way to handle this is to have exactly one non fake rule in `.bnf` and write a custom annotator which would say something like "top level functions must have a body, change `;` to `{}`?". See, how this is done in Kotlin: if you write `override fun foo()` at the top level, the PSI will containg a `FUN`, which would be annotated with two errors (an extra override and a missing body). 

So this issue is exactly about removing special casing in the parser and writing an annotator. The best place to start is `constant`, then `type_alias`. `function` might prove to be tricky, because different functions have different grammar for parameters ([I hate this](https://github.com/rust-lang/rfcs/pull/1685)). It might be useful to look at the Kotlin's annotator!

alygin:
I'll take constants for starters. It looks like many of [these errors](https://github.com/intellij-rust/intellij-rust/issues/886) should also be handled by annotators, so it'll be a good practice.

alygin:
BTW, what is the `default` contextual keyword referenced from many grammar rules but absent in the Rust docs? Is it extinct or did I neglect something important in the docs?

matklad:
It's a part of specialization: https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md

The specialization hasn't hit stable yet, but we have to support it on the syntactic level because the standard library uses nightly features.

alygin:
Oh, good to know. But why doesn't the grammar kit generate a field in `RustConstantElement` for it?

matklad:
It's because `default` is generated by an external rule, and GrammarKit can't undestand external rules. The external rule itself is needed because the keyword is contextual, so we need to remap identifier to the keyword. 

You need to define `fun RustConstantElement.getDefault(): PsiElement? = findChildByType(RustTokenElementTypes.DEFAULT)` manually.

alygin:
@matklad, could you, please, [have a look](https://github.com/alygin/intellij-rust/commit/a03d1d17f693c46056ebda3d4f76596cd4f444ef), if I got the idea correctly?

[...]

PR #892, opened 2017-01-04
(GRAM): Switch from multi-rule mode to annotations for constants

alygin:
The first part of changes to make #864 happen. The `constant` grammar rule has no variants any more and all the specifics of its usage in different contexts are handled by  `RustInvalidSyntaxAnnotator`.

There's a slight problem that I described in [this comment](https://github.com/intellij-rust/intellij-rust/issues/864#issuecomment-270430613). I've TODOed tests ([one](https://github.com/intellij-rust/intellij-rust/compare/master...alygin:gram-to-annot?expand=1#diff-14ef865dc1201f0af9825a321e09bc3dR52), [two](https://github.com/intellij-rust/intellij-rust/compare/master...alygin:gram-to-annot?expand=1#diff-14ef865dc1201f0af9825a321e09bc3dR64)) that break if you uncomment the provided lines and add them to the test because `GeneratedParserUtilBase` adds its own errors.


matklad:
Hm, the example from the comment seems to work as expected here:

constatns


There are two errors generated by the annotator, with nice messages.

Have you run `Generate Parser` task after modifying `Rust.bnf`?

matklad:
Tough TODO tests are failing... Investigating.

matklad:
Aha, I've found the culprit! It's

```
private trait_member_recover ::= !('#' | vis | EXTERN | CONST | UNSAFE | FN | TYPE_KW | '}')
```

This is the rule defining which tokens should be skipped after parsing a declaration inside a trait. There's no `default` in this list, so in

```
trait T {
  fn foo();
  default const S: i32 = 0;
}
```

parser will parse `foo` and then skip over `default`, marking it as an error. To fix it, you should change the recover rule to

```
private trait_member_recover ::= !('#' | vis | EXTERN | CONST | UNSAFE | FN | TYPE_KW | '}' | "default" )
```

It's important to undestand that this recover rule is always called, even if the previous element was parsed correctly. So a wrong `recoverWhile` can introduce spurious syntax errors into the valid file.

alygin:
Thanks! Then maybe `STATIC` should also be added to `trait_member_recover` and `impl_member_recover`?

src/main/kotlin/org/rust/ide/annotator/RustInvalidSyntaxAnnotator.kt
```diff
@@ -19,6 +24,45 @@ class RustInvalidSyntaxAnnotator : Annotator {
...
+            if (el != null) null else holder.createErrorAnnotation(highlightElements.combinedRange() ?: TextRange.EMPTY_RANGE, message)
+
+        private fun deny(el: PsiElement?, message: String, vararg highlightElements: PsiElement?): Annotation? =
+            if (el == null) null else holder.createErrorAnnotation(highlightElements.combinedRange() ?: el.textRange, message)
+
+        private fun Array<out PsiElement?>.combinedRange(): TextRange? {
+            var range: TextRange? = null
+            filterNotNull()
+                .map { it.textRange }
+                .forEach { range = range?.union(it) ?: it }
```
matklad: `reduce(TextRange::union)`

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustConstantImplMixin.kt
```diff
@@ -34,3 +34,23 @@ val RustConstantElement.kind: RustConstantKind get() = when {
...
+        is RustItemsOwner -> RustConstantRole.FREE
+        is RustTraitItemElement -> RustConstantRole.TRAIT_CONSTANT
+        is RustImplItemElement -> RustConstantRole.IMPL_CONSTANT
+        is RustForeignModItemElement -> RustConstantRole.FOREIGN
+        else -> error("Unexpected constant parent: $parent")
+    }
+}
+
+val RustConstantElement.default: PsiElement?
+    get() = node.findChildByType(RustTokenElementTypes.DEFAULT) as? PsiElement
```
matklad: `?.psi` looks nicer that a cast.

matklad:
>Thanks! Then maybe STATIC should also be added to trait_member_recover and impl_member_recover?

Yep! If this is not covered by the annotator tests, you can add it to `org.rust.lang.core.parser.RustCompleteParsingTestCase`.

matklad:
>If this is not covered by the annotator tests

Though annotator test is preferable. It tests the user visible functionality, while parser tests test internal representation, so they are way more fragile and don't actually test the stuff we are really interesting in!

src/main/kotlin/org/rust/ide/annotator/RustInvalidSyntaxAnnotator.kt
```diff
@@ -19,6 +24,44 @@ class RustInvalidSyntaxAnnotator : Annotator {
...
+        }
+
+        private fun require(el: PsiElement?, message: String, vararg highlightElements: PsiElement?): Annotation? =
+            if (el != null) null else holder.createErrorAnnotation(highlightElements.combinedRange ?: TextRange.EMPTY_RANGE, message)
+
+        private fun deny(el: PsiElement?, message: String, vararg highlightElements: PsiElement?): Annotation? =
+            if (el == null) null else holder.createErrorAnnotation(highlightElements.combinedRange ?: el.textRange, message)
+
+        private val Array<out PsiElement?>.combinedRange: TextRange?
+            get() = if (isEmpty())
```
alygin: Unfortunately, `fold` with an empty initial range won't work as needed here because it will always produce ranges that start from (0, 0).

matklad:
@alygin anything else you want to do in this PR?

alygin:
I could add parsing tests for constants if you think they are necessary, though it'll take some time. Otherwise no. I think I'll add quick fixes in a separate PR.

matklad:
Let's merge this then! Thanks a lot!

PR #893, opened 2017-01-04
(GRAM): Switch from multi-rule mode to annotator for type aliases

alygin:
Here's the second part of #864. Type aliases grammar rule is unified and all the checks are moved to `RustInvalidSyntaxAnnotator`. I've also extracted detection of two Rust errors related to the issue (E0202, E0449) and provided them with the same messages as rustc produces.

Unfortunately, there's one broken parsing test I can't fix: `RustPartialParsingTestCase > testTraitBody`. It looks like a recovery problem we faced earlier, but I have no clue how to solve it. The parser just interprets some gibberish as a beginning of a type alias statement.

@matklad, could you, please, take a look at it?

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -269,7 +269,7 @@ private attrs_and_vis ::= outer_attr* vis?
 private attrs_default_vis ::= outer_attr* DEFAULT? vis?
 
 private item ::= constant
-               | type_item
+               | type_alias
                | fn_item
                | trait_item
                | impl_item
@@ -690,7 +690,7 @@ trait_item ::= attrs_and_vis UNSAFE? TRAIT IDENTIFIER generic_params? type_param
 private trait_body ::= '{' trait_member* '}' { pin = 1 }
 
 // TODO: use upper rules and inheritance here
-private trait_member ::= !'}' ( constant | trait_type_member | trait_method_member ) {
+private trait_member ::= !'}' ( constant | type_alias | trait_method_member ) {
   pin = 1
   recoverWhile = trait_member_recover
 }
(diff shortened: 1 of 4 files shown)


matklad:
Do you see  ` PsiErrorElement:type expected, got 'intellijidearulezz'` instead of `PsiErrorElement:'intellijidearulezz' unexpected`? This is perfectly fine: we don't pay any attention at the moment to the error messages generated by grammar kit. Just change the `.txt` file with results. 

The purpose of partial tests is not to check for error messages, but to make sure that the code after the error is properly parsed. Try removing `recoverWhile` attribute form `trait_member` rule and see how this affects this test case.

matklad:
BTW, this is exactly "parser tests are fragile" instance I was complaining about :)

matklad:
I'll fix parser tests myself and merge this now: I want to change PSI a bit and don't want to cause merge conflicts within the `.txt` files of parser tests

alygin:
Yes, I saw the message. The problem is that I didn't understand the reason this message appeared. There's nothing significant had been changed in grammar that should have changed the message from 'unexpected `iir`' to 'type expected, got `iir`'. Why in the world is it expecting a type in this code snippet:

```rust
trait Foo {
    intellijidearulezz
}
```

?

Thanks for fixing and merging!

matklad:
Frankly speaking I don't know :) `type` here means the `type` keyword (the beginning of type alias. And it is reasonable to see it here, because we are parsing a list of `trait_member`s at this point
 
```
private trait_body ::= '{' trait_member* '}' { pin = 1 }

private trait_member ::= !'}' ( constant | trait_type_member | trait_method_member ) {
  pin = 1
  recoverWhile = trait_member_recover
}
```

and `type` token belongs to `FIRST(trait_member)`. I think that before Grammar Kit could have got confused before because of the `elementType = ` attributes? `consumeTokenMethod(".*") = "consumeTokenFast"` is another thing that affects which error messages are generated.

Issue #886, opened 2017-01-03
Detect Rust errors


alygin:
Here's the list of the Rust compiler errors that I think are relatively easy to detect and highlight given the current state of the parser and resolve engine. I bet I neglected some other errors that could be in this list because of underestimation of the resolve engine or overestimation of the error detection complexity or just because of a simple carelessness. So, any corrections and additions are very welcome.

Some of these errors could only be detected in a limited number of cases so far, but it's still enough for them to be helpful.

I think this list could be treated as a meta issue from which anyone could grab tasks to accomplish. I will surely do that.

I've also made so bold as to extract a short list of those errors that seem to be more common and their detection will provide the most benefit. But, again, I might be wrong here, so feel free to criticize this list and suggest improvements.

Errors that could have a quick-fix are marked as (fixable).

More common:
- [ ] [E0050](https://doc.rust-lang.org/error-index.html#E0050): Wrong number of parameters in method implementation.
- [ ] [E0061](https://doc.rust-lang.org/error-index.html#E0061): Wrong number of arguments / enum variant tuple fields.
- [ ] [E0106](https://doc.rust-lang.org/error-index.html#E0106): Lifetime is missing from a type (fixable).
- [ ] [E0107](https://doc.rust-lang.org/error-index.html#E0107): Incorrect number of lifetime parameters provided for a type 
- [ ] [E0185](https://doc.rust-lang.org/error-index.html#E0185) / [E0186](https://doc.rust-lang.org/error-index.html#E0186): Trait associated function is a method in implementation / vice versa.
- [ ] [E0201](https://doc.rust-lang.org/error-index.html#E0201): Function name duplication in an `impl` block.
- [ ] [E0232](https://doc.rust-lang.org/error-index.html#E0232): The attribute must have a value.
- [ ] [E0261](https://doc.rust-lang.org/error-index.html#E0261): Lifetime must be declared before being used (fixable).
- [ ] [E0262](https://doc.rust-lang.org/error-index.html#E0262): Declaring certain lifetime names in parameters is disallowed.
- [ ] [E0306](https://doc.rust-lang.org/error-index.html#E0306): The number of elements in the array must be an unsigned integer.
- [ ] [E0308](https://doc.rust-lang.org/error-index.html#E0308)(1): Mismatched types on function call. Can be implemented for at least simple cases where the arguments types are known.
[...]

PR #897, opened 2017-01-05
(REF): Annotators refactoring

alygin:
I've preformed a minor annotators refactoring to prepare for working on #886:
- Fixes moved to the `annotator.fixes` package, the same way it's done for inspections
- Some tests moved from fixture-files
- A couple of fixture-test files removed because they had been moved to `RustExpressionAnnotatorTest` long before.

There're now two annotators &mdash; `RustInvalidSyntaxAnnotator` and `RustItemsAnnotator` &mdash; that do pretty the same things: highlight both syntax and compile errors. I think they could be combined into a single `RustErrorsAnnotator` that would be responsible for all the errors hunting. The difference between syntax and all other errors is pretty subtle (for instance, `pub` within a trait block is a syntax error, while `pub` within a trait impl block is an error with its own code) and they all use the same analysis patterns, so I personally see no reasons to separate them.

What do you think?


matklad:
Looks reasonable, thanks!

>There're now two annotators — RustInvalidSyntaxAnnotator and RustItemsAnnotator — that do pretty the same things: highlight both syntax and compile errors. I think they could be combined into a single RustErrorsAnnotator that would be responsible for all the errors hunting. 

Yes, I think it makes sense to combine the annotators.

Issue #866, opened 2016-12-30
Support code completion for default functions in traits


kirillkh:
```
trait X {
    fn a(&self) {
        self.b();  // <- code completion doesn't work after "self."
    }

    fn b(&self) {}
}
```

himikof:
Duplicated by #875, should be fixed in f0a252d.

matklad:
Yes, thanks @himikof !

kirillkh:
Still not fixed for me in the latest plugin release (0.1.0.1385). Navigation between functions works now, but code completion doesn't. E.g.: [line 56 here](https://github.com/kirillkh/rs_teardown_tree/blob/41afe177c86466be2977211b4fc7e65e39457394/src/applied/plain_tree.rs).

PR #900, opened 2017-01-06
(ANNOT): Trait impl errors annotation

alygin:
A few trait implementation related errors detection from #866:
- E0046 and E0407 had already been implemented, I just changed the messages slightly to match what rustc emits
- E0050 - Method `foo` has N parameters but the declaration in trait `Foo` has M
- E0185 / E0186 - Method `foo` has a `self` declaration in the impl / trait, but not in the trait / impl


src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
```diff
@@ -148,21 +153,56 @@ class RustErrorAnnotator : Annotator {
...
             .reduce(TextRange::union)
 
-    private fun isInTraitImpl(o: RustVisElement): Boolean {
-        val impl = o.parent?.parent
-        return impl is RustImplItemElement && impl.traitRef != null
+    private val Collection<String?>.namesList: String
+        get() = mapNotNull{ "`$it`" }.joinToString(", ")
+
+    private fun RustSelfArgumentElement.name(): String {
+        val sb = StringBuilder(9)
```
matklad: There's a nice `buildString` function in Kotlin.

src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
```diff
@@ -106,18 +103,26 @@ class RustErrorAnnotator : Annotator {
...
 
         val notMembers = implemented.filterKeys { it !in canImplement }
         for (method in notMembers.values) {
             holder.createErrorAnnotation(method.identifier,
-                "Method is not a member of trait `${trait.name}`")
+                "Method `${method.name}` is not a member of trait `$traitName` [E0407]")
         }
+
+        implemented
+            .map { it.value to canImplement[it.key] }
```
matklad: This can be rewritten using `mapNotNull` to avoid the `!!`, not sure it's worth doing though.

alygin: It's only the `second` that is checked for `null`, so it can't be transformed to `mapNotNull`. Or did you mean something else?

src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
alygin: I'll refactor it.

src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
matklad: Something like

```
.mapNotNull { (name, fn) ->
  canImplement[fn]?.let { fn to it }  
}
```

alygin: Hm, maybe I'm doing something wrong, but such syntax just doesn't work.

matklad: Yeah, you can't destruct in lambdas at the moment (this will be fixed in Kotlin 1.1 I believe), so it has to be more ugly:

```
.mapNotNull {
  val (name, fn) = it
  canImplement[fn]?.let { fn to it }  
}
```

alygin: Kotlin can't infer types in such expression and asks to add them somewhere explicitly (even with the `forEach` block). So, it'll be even uglier.

matklad: Let's leave it as is then.

matklad:
Thanks!

PR #901, opened 2017-01-07
(ANNOT): Duplicate name annotations

alygin:
This is a part of #886 that detects errors E0201 (name duplication in `impl` blocks) and E0428 (name duplication in modules, enums and code blocks).

Unfortunately, there's a problem that prevents duplication detection on the file level: all annotations, added inside `visitFile()` are displayed on the screen for a fraction of a second and then just disappear. They pass unit tests, but just are not displayed correctly. It looks for me as a platform issue, though I might just not know something. So, any help here will be appreciated.

I decided not to include the `visitFile()` part in this PR so far, though the code is absolutely the same as for `visitModItem()`.


matklad:
>It looks for me as a platform issue, though I might just not know something. So, any help here will be appreciated.

Looks strange indeed. I doubt this is really a platform issue though: similar features in other languages work just fine.

And (that's a bit embarrassing for me :) ) turns out we already have some duplicates implemented via inspection: `org.rust.ide.inspections.duplicates` :) We need to do something with it...

matklad:
Ok, so here are three options we have I think for duplicates:

1) implement them as an inspection (Go plugin does this)
2) implement them as a part of annotator (Kotlin does this)
3) implement `highlightVisitor` as Java does. 

I think annotator is the right approach here, but we need to figure out how to fix it for the file case.

alygin:
Regarding the existing duplication inspections: there're four of them now (for struct fields, for trait types, for trait functions and for trait constants). I can move them to annotators as a part of this PR, they will only gain from this because they now don't catch duplications between different types (for instance, a constant `A` and a function `A()` in the same trait).

matklad:
And there's one more tricky bit, code like

```
mod foo;

fn foo() {}
```

Should not trigger the error, it's a valid Rust because modules and functions live in different namespaces.

alygin:
Yeah, I noticed that, modules and types live in one namespace, funcs/consts/structs/traits/enums &mdash; in another. I need to take that into account in the annotator.

matklad:
There's org.rust.lang.core.resolve.ScopeEntryKt#getNamespaces`

alygin:
Oh, so things are a bit more complicated. All right, I'll implement differentiation in the duplication analysis.

alygin:
> similar features in other languages work just fine

Yes, the features work fine, but I'm not sure they are implemented in `visitFile()`. I went through several language plugins but none of them actually used this method for annotating.

I also tried `visitItemsOwner`, but it doesn't work for files.

alygin:
@matklad, I think `RustTypeAliasElement` must fall into the `TYPES` namespace group [here](https://github.com/intellij-rust/intellij-rust/blob/2662757ba542b2d11a09d6bff057ade4ff526f75/src/main/kotlin/org/rust/lang/core/resolve/ScopeEntry.kt#L86). For instance, this code is valid in Rust:

```rust
type Foo = bool;
fn Foo() {}            // no conflict with functions

trait T {
    type Bar;
    const Bar: u8;}    // no conflict with constants
}
```

Am I correct or there's something else I don't understand here?

matklad:
> I think RustTypeAliasElement must fall into the TYPES namespace group here

Yes, that is correct.

alygin:
And the same is `RustTraitItemElement`

alygin:
Actually, enum variants are `BOTH`:

```rust
mod stuff {
    pub enum E { SOUTH }
}

use stuff::E::SOUTH;

const SOUTH: u32 = 10;   // error: a value named `SOUTH` has already been imported in this module
type SOUTH = u32;        // error: a type named `SOUTH` has already been imported in this module
```

matklad:
>Actually, enum variants are BOTH:

Hm, this sounds like a bug in Rustc to me!

matklad:
Or may be this is for forward compatibility reasons I can imagine extension of the language where enum variants are proper types. I think it makes sense to ask in the forum about this and treat enum variants as *values only* for the time being in the plugin.

alygin:
Just found [this issue](https://github.com/rust-lang/rfcs/pull/1450), reading.

alygin:
Yep, I left them as is at the moment. Will try to find more information on the issue. At the worst case it will result in false negatives which is not critical.

I've removed the mentioned inspections and converted them into annotations, so the error E0124 (name duplication in struct) is also covered now. When looking for duplications, all names are grouped into namespaces at first (with elements from `BOTH` going into both groups) and then the analysis is performed separately in each group.

matklad:
Thanks!

I've tried to figure out what's the problem with file, but I have no explanation for this at the moment :(

alygin:
Neither have I. I've tried lots of different approaches to this issue, but it just doesn't work on `visitFile()`, though the annotations are shown when executing 'Analyse Code...' The only annotation that works properly inside `visitFile()` is the one applied to the whole file.

BTW, there's the `RustFile` class that extends `PsiElement`, but `visitFile()` accepts `PsiElement`, not `RustFile`. Maybe something should be done to make it accept `RustFile`? I don't understand how `RustElementVisitor` is generated. On the one hand, it should be generated automatically, on the other had, I see comments in it that don't look like they were autogenerated but can't find the source they came from.

PR #902, opened 2017-01-07
(TYPE): i128/u128 types support

### LINKED ISSUE #895: Support u128 and i128
This would require adding new primitives to type system afair, and add them to all suffix lists in our code base.

alygin:
As far as I understand, it should be enough to fix #895.


matklad:
Thanks @alygin and @mkaput for review!

PR #904, opened 2017-01-07
(COMPL): Extern crates completion fix

### LINKED ISSUE #903: extern crate completion uses wrong name
Cargo package and the library inside the package can have different names. For example, `rust_crypto` package defines the `crypto` crate, so this completion is wrong: 

crypto

By the way, resolve works correctly (resolves `extern crate crypto` and not `extern crate rust_crypto`).

cc @alygin

alygin:
Use target with `kind = LIB` in autocompletion instead of the package name.
Fixes #903

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -27,9 +27,6 @@ class CargoProjectDescription private constructor(
         val source: String?,
         val origin: PackageOrigin
     ) {
-        // crate name must be a valid Rust identifier, so normalize it by mapping `-` to `_`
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
```diff
@@ -29,7 +29,6 @@ class CargoProjectDescription private constructor(
     ) {
         // crate name must be a valid Rust identifier, so normalize it by mapping `-` to `_`
         // https://github.com/rust-lang/cargo/blob/ece4e963a3054cdd078a46449ef0270b88f74d45/src/cargo/core/manifest.rs#L299
-        val normName = name.replace('-', '_')
```
matklad: Can you move this comment to another instance of this replace down in the file?

alygin: Oh, sure, didn't notice it.

matklad:
Thanks!

PR #905, opened 2017-01-08
(ANNOT): Detect type param names duplication

alygin:
Annotation to highlight the E0403 error &mdash; duplication of type parameter names.
Part of #886


matklad:
Thanks!

BTW, I am not 100% sure, but you don't have to repeate `descr` field several times in tests. That is, you can have on `<error descr="bla-bla-bla">...</error>` to check the message and all the rest can be just `<error>...</error>`

alygin:
Yes, the simple `<error>...</error>` works for checking highlighting, thanks.

BTW, I wanted to implement the same duplication check for lifetime parameters, but it occurred that they are not `RustNamedElement`s. Was it done intentionally?

matklad:
>Was it done intentionally?

Nope, answered on the related PR!

PR #906, opened 2017-01-08
(ANNOT): Detect lifetime names duplication

alygin:
It's E0263 (lifetime name duplication) from #886.

Values `uniqueName` and `uniqueNameElement` are extracted to be used on non-`RustNamedElement`s for other similar errors.


matklad:
I think we need just to make lifetemes into named elements (and include `'` in the name). They are not named at the moment just because nobody has implemented this (we don't do any resolve for lifetimes, although we could, and arguably, we can even make it false positive-free).

alygin:
Perfect! Then I'll make them named and will use the same duplication detector I used for other elements. Till then, I close the PR.

Another question here is what namespace should liftimes and type parameters belong to? `TYPES`?

matklad:
>Another question here is what namespace should liftimes and type parameters belong to? TYPES?

I believe they should live in a separate namespace `LIFETIMES`.

alygin:
OK

PR #910, opened 2017-01-09
(ANNOT): Detect lifetime names duplication

alygin:
It's E0263 (lifetime name duplication) from #886.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
@@ -129,7 +129,10 @@ class RustErrorAnnotator : Annotator {
 
     private fun checkGenericParams(holder: AnnotationHolder, params: RustGenericParamsElement) =
         findDuplicates(holder, params, { ns, name ->
-            "The name `$name` is already used for a type parameter in this type parameter list [E0403]"
+            if (ns == Namespace.Lifetimes)
+                "Lifetime name `$name` declared twice in the same scope [E0263]"
+            else
+                "The name `$name` is already used for a type parameter in this type parameter list [E0403]"
         })
 
     private fun checkImpl(holder: AnnotationHolder, impl: RustImplItemElement) {
--- a/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustLifetimeParamElementImplMixin.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustLifetimeParamElementImplMixin.kt
@@ -1,0 +1,11 @@
+package org.rust.lang.core.psi.impl.mixin
+
+import com.intellij.lang.ASTNode
+import org.rust.lang.core.psi.RustLifetimeParamElement
+import org.rust.lang.core.psi.impl.RustNamedElementImpl
+
+abstract class RustLifetimeParamElementImplMixin(node: ASTNode) : RustNamedElementImpl(node), RustLifetimeParamElement {
+
+    override fun getNameIdentifier() = lifetime
+
+}
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -336,7 +336,11 @@ generic_params ::= '<' lifetimes_params type_params '>' {
 private lifetimes_params ::= <<list_element lifetime_param>>*
 private type_params      ::= <<list_element type_param>>*
 
-lifetime_param ::= LIFETIME lifetime_param_bounds?
+lifetime_param ::= LIFETIME lifetime_param_bounds? {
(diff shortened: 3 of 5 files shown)


src/main/kotlin/org/rust/ide/annotator/RustErrorAnnotator.kt
```diff
@@ -128,8 +128,11 @@ class RustErrorAnnotator : Annotator {
         })
 
     private fun checkGenericParams(holder: AnnotationHolder, params: RustGenericParamsElement) =
-        findDuplicates(holder, params, { ns, name ->
-            "The name `$name` is already used for a type parameter in this type parameter list [E0403]"
+        findDuplicates(holder, params, name@ { ns, name ->
+            return@name if (ns == Namespace.Lifetimes)
+                "Lifetime name `$name` declared twice in the same scope [E0263]"
+            else
+                "The name `$name` is already used for a type parameter in this type parameter list [E0403]"
```
mkaput: this return is not neccessary, `if/else` is expression

alygin: Oh, sure, you're right

src/main/kotlin/org/rust/lang/core/psi/impl/RustNamedElementImpl.kt
```diff
@@ -6,12 +6,14 @@ import com.intellij.psi.PsiNameIdentifierOwner
...
 
-abstract class RustNamedElementImpl(node: ASTNode) : RustCompositeElementImpl(node),
-                                                     RustNamedElement,
-                                                     PsiNameIdentifierOwner {
+abstract class RustNamedElementImpl(
+    node: ASTNode,
+    val idTokenType: RustTokenType = RustTokenElementTypes.IDENTIFIER
+) : RustCompositeElementImpl(node), RustNamedElement, PsiNameIdentifierOwner {
 
-    override fun getNameIdentifier(): PsiElement? = findChildByType(RustTokenElementTypes.IDENTIFIER)
```
matklad: This non final implementation is here exactly to avoid declaring a `mixin` class. But, because you name element is identifier, you have to declare mixin, so just override this method there instead of modifying the base class. You'll need something like `nameIdentifier = lifetime` I think.

alygin: Yeah, I see, that makes sense.

matklad: As an aside, ideally such "one-mehod" overrides should work via "method injection": https://github.com/JetBrains/Grammar-Kit/blob/master/HOWTO.md#34-implement-interface-via-method-injection, but I have not yet figured out how to make this work with our build system.  "no need to specify arguments or return type" means that grammar kit uses IDEA API to fetch the signatures, which does not exactly work in standalone mode...

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustLifetimeParamElementImplMixin.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.intellij.lang.ASTNode
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RustLifetimeParamElement
+import org.rust.lang.core.psi.RustTokenElementTypes
+import org.rust.lang.core.psi.impl.RustNamedElementImpl
+
+abstract class RustLifetimeParamElementImplMixin(node: ASTNode) : RustNamedElementImpl(node), RustLifetimeParamElement {
+
+    override fun getNameIdentifier(): PsiElement? = findChildByType(RustTokenElementTypes.LIFETIME)
```
matklad: Can you use just `getLifetime()` function from `RustLifetimeParamElemnt`?

alygin: Sure

matklad:
Thanks!

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RustLifetimeParamElementImplMixin.kt
matklad: To expand on this a bit, here's an interesting entanglement of the base and derived class happens!

The leaf class of the hierarchy is `RustLifetimeParamElementImpl`, and we want to define `getNameIdentifier` there. We don't have direct control over the leaf, so we define a base class mixin and implement the method there. 

However, we employ `getLifetime` which is itself defined in the leaf class! In the ancestor, we are able to use the method defined in the descendant because both ancestor and the descendant share the same interface `RustLifetimeParamElement`!

PR #917, opened 2017-01-11
(INT): add a 'Simplify boolean expression' intention

slavam2605:
This is my first contribution so I'm not quite sure that all is ok.
There is a problem: name of intention 'Simplify boolean expression' clashes with Java intention and when in editor I try to change settings of rust's intention IDEA shows me 'Java/Boolean/Simplify boolean expression' intention. Maybe we shoud rename new intention a bit.
Initially I wanted to do integer expression simplification, but I realized, that it is quite hard to infer type of integer literal. So if you can give a piece of advice about how it could be implemented I could try to do it. It would be very useful intention.


slavam2605:
Also it would be great anyway to move RustUnaryExprElement.operatorType() and RustUnaryExprElement.operator() extensions to somewhere public (they are copies of the same extensions for RustBinaryExprElement, it's very complicated to work with expressions without these extensions. Also some kind of extensions are needed for RustLitExprElement to pattern match over different kinds of literals, like `true`/`false`, string literals and so on).

slavam2605:
And testCargoFmt test FAILED -- but I don't think that it's connected with my PR. I had this problem locally too. I've installed rustfmt to make this test green.

matklad:
Great, will take a look shortly! 

>There is a problem: name of intention 'Simplify boolean expression' clashes with Java intention and when in editor I try to change settings of rust's intention IDEA shows me 'Java/Boolean/Simplify boolean expression' intention. Maybe we shoud rename new intention a bit.

I think `familyName` should be changed to avoid this problem. It is probably something that can be tweaked via `plugin.xml` as well. 

>Initially I wanted to do integer expression simplification, but I realized, that it is quite hard to infer type of integer literal. So if you can give a piece of advice about how it could be implemented I could try to do it. It would be very useful intention.

I think that we should not add this to the initial implementation: it's easier to do the ground work in one limited in scope PR and then expand on it further. And yes, I think there is indeed a problem with inferring types of integer literals if they don't have an explicit type/suffix. I think this isn't something that can be easily done at this stage, because it requires more or less complete and precise type inference, and we are far from there yet. 

>Also it would be great anyway to move RustUnaryExprElement.operatorType() and RustUnaryExprElement.operator() extensions to somewhere public (they are copies of the same extensions for RustBinaryExprElement, it's very complicated to work with expressions without these extensions. Also some kind of extensions are needed for RustLitExprElement to pattern match over different kinds of literals, like true/false, string literals and so on).

Yes, there's little functionality implemented on this level yet, so the API is very rough. If you find that you need some kind of extension function, or if you think you can somehow simplify and refactor existing code, please go ahead! Also, take a look at the DeMorgan laws intention: I think it may also use some more convenient extensions!

matklad:
>And testCargoFmt test FAILED -- but I don't think that it's connected with my PR. I had this problem locally too. I've installed rustfmt to make this test green.

Ah, some tests are flaky because of the `ThreadLeaked` problem :( I know exactly what is the cause of this exceptions, but I haven't yet figured out how to fix it most properly, so just ignore this failures for the time being and sorry for inconvenience!

src/main/kotlin/org/rust/ide/intentions/SimplifyBooleanExpressionIntention.kt
```diff
@@ -0,0 +1,122 @@
...
+                null
+            }
+            else -> null
+        }
+    }
+
+    val RustUnaryExprElement.operator: PsiElement
+        get() = requireNotNull(node.findChildByType(UNARY_OPS)) { "guaranteed to be not-null by parser" }.psi
+
+    val RustUnaryExprElement.operatorType: IElementType
```
matklad: What do you think about introducing `enum class UnaryOperator`? It'd give us a bit more typesafety in comparison with the `IElementType`.

src/main/kotlin/org/rust/ide/intentions/SimplifyBooleanExpressionIntention.kt
```diff
@@ -0,0 +1,122 @@
...
+import org.rust.lang.core.psi.RustTokenElementTypes.*
+import org.rust.lang.core.psi.util.elementType
+import org.rust.lang.core.psi.util.parentOfType
+
+class SimplifyBooleanExpressionIntention : RustElementBaseIntentionAction<SimplifyBooleanExpressionIntention.Context>() {
+    override fun getText() = "Simplify boolean expression"
+    override fun getFamilyName() = text
+
+    data class Context(
+        val expr: RustExprElement
```
matklad: Are you planing to expand the `Context` somehow? If not, you can use `RustExprElement` as a context directly.

src/main/kotlin/org/rust/ide/intentions/SimplifyBooleanExpressionIntention.kt
```diff
@@ -0,0 +1,122 @@
...
+        var expr = element.parentOfType<RustExprElement>() ?: return null
+        var lastSimplifiable: RustExprElement? = null
+        if (isSimplifiableExpression(expr))
+            lastSimplifiable = expr
+
+        while (true) {
+            expr = expr.parentOfType<RustExprElement>() ?: break
+            if (isSimplifiableExpression(expr))
+                lastSimplifiable = expr
+        }
```
matklad: I think this can be rewritten in a more functional style using Kotlin seqnences. Take a look at the `ancestors` method and it's implementation. 

I think it should be possible to do something like `expr.ancestors.takeWhile { it is RustExprElement }.find(isSimplifiableExpression)`.

matklad:
Ah, I also will do a **massive** rename shortly, so unfortunately expect merge conflicts :(

src/main/kotlin/org/rust/ide/intentions/SimplifyBooleanExpressionIntention.kt
slavam2605: I thought that I would add other types support and I would need some additional meta-information. But for now `expr` is enough, so I will eliminate useless data class.

src/main/kotlin/org/rust/ide/intentions/SimplifyBooleanExpressionIntention.kt
slavam2605: My first idea was the same. But then I saw the same thing for RustBinaryExprElement and made it the same to ask later what is better. I will change it now.

slavam2605:
I'm really love new names of classes! It looks consistently and neat now.
(Maybe you forgot to rename `UnescapeRustTest` to `UnescapeRsTest`, don't know.)
I rewrote `findApplicableContext` in more functional way, eliminate `Context` class and added `UnaryOparator` enum with a small documentation and an extension for `RsUnaryExpr`.
I've added one test to be sure that my new implementation of `findApplicableContext` is really correct.
Also about `getFamilyName()` -- I think that it shoud be named "Simplify boolean expression" and none else. But somehow IDEA doesn't take into account category of intention when references a menu option. So I made a dirty trick -- be aware, maybe you can make it in a better way -- I replaced second whitespace in "Simplify boolean expression" with nbsp; (non-breakable space). Looks the same, but in fact is another string. Now it works, I've tested.

matklad:
Thanks a lot!

>Maybe you forgot to rename UnescapeRustTest to UnescapeRsTest, don't know.

Yeah, there's a ton of classes to rename :) 

> Looks the same, but in fact is another string.

Hm, I am almost sure there should be another way, but if this works, then it is fine!

Issue #912, opened 2017-01-11
Highlight matching angle brackerts


matklad:
In something like `Result<Foo, Bar>`, `<` should be highlighted when the cursor is over `>`.

This should not be hard (I hope :) ), but I don't know off the top of my head which extension point do we need exactly. A good idea would be to find out how Java does this.

mkaput:
I worked on that some time ago, they do some mangling in brace matcher: https://upsource.jetbrains.com/idea-ce/file/idea-ce-1ee8a1fbfebb243266c84879345f8030dd3c2bbe/java/java-impl/src/com/intellij/codeInsight/highlighting/JavaBraceMatcher.java?nav=0:0:focused

Test: https://upsource.jetbrains.com/idea-ce/file/idea-ce-1ee8a1fbfebb243266c84879345f8030dd3c2bbe/java/java-tests/testSrc/com/intellij/codeInsight/highlighting/JavaBraceMatcherTest.java?nav=0:0:focused

matklad:
Fixed and in alpha

PR #923, opened 2017-01-12
(HIGHLIGHTER) Highlight matching angle brackets #912

oistein:


matklad:
Thanks! As written, this works a little bit too good: it will match `<` and `>` in the following code: `let _ = 1 < 2 && 4 > 3;`

I think this can be handled if you add another  `RsPairedBraceMatcher` inhereting from `PairedBraceMatcherAdapter`. Take a look at `JavaPairedBraceMatcher` which uses a more simple `JavaBraceMatcher`.

oistein:
Yes . Will fix tonight.

matklad:
Hm, wait a second, looks like this is more complicated then I thought....

So, the problem here is that brace paring is done on the lexical level in the API. That is, you have access to the highlighting iterator, but not to the PSI, and this poses a problem.

In Java, the set of tokens that an appear between paired `< >` is very limited (they are **all** listed in the `Holder` struct), and is distinct from the tokens that may appear in the expression context. 

In Rust, the type grammar is more complex, so a lot of stuff can appear between `< >`. Crucially, whole expressions may appear there (array type is `[type; expr]`)! So looks like we just can't use this API here :( In particular, I believe this implementation would not match `Vec<[u8; 4]>`.

I wonder if this is the reason why Kotlin doesn't match `< >` either...

Though I think if we account only for constants, this will give us a nice and useful approximation! Can you extend this a bit (implementation and tests) such that the following cases work:

* paths `<foo::bar>`, `<self::foo>`, `<super::foo>`
* arrays with constant lengths: `<[u8; 92]>`
* tuples `<((), (i32))>`
* references `<&foo>`, `<&mut foo>`
* Fn traits `<Fn(u32) -> F64>`

And I think we should add some more flavorful not matched tests as well, including false positives like `(1 < 2) > 3`.


Also, a short comment explaining the issue (we want PSI tree, but we got only highlighting iterator, so let's use heuristics) would be useful!

matklad:
Also I imagine we can write a more "property based" flavored test here. What do you think about the following setup:

* We basically have a single tests, which receives as an input a big string literal with Rust code, which contains all the interesting positive cases. 

* We parse this file to PSI, iterate over all elements that should have matched `<` (`RsTypeParameterList` and `RsTypeArgumentList` I believe) and make sure that they are matched

* We do the same for negative cases, but this time we iterate over characters and test that every `<` does not match. 


Sorry that this turned out into a rather substantiation project :)

src/main/kotlin/org/rust/ide/typing/RsPairedBraceMatcher.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.intellij.openapi.fileTypes.FileType
+import com.intellij.psi.tree.TokenSet
+import org.rust.lang.RsFileType
+import org.rust.lang.RsLanguage
+import org.rust.lang.core.psi.RsTokenElementTypes
+import org.rust.lang.core.psi.RsTokenElementTypes.*
+
+class RsPairedBraceMatcher : PairedBraceMatcherAdapter(RsBraceMatcher(), RsLanguage) {
+
+    private object Holder {
```
matklad: You can use `companion object` here: 

```
companion object {
   val TYPE_TOKENS = ....
}
```

src/main/kotlin/org/rust/ide/typing/RsPairedBraceMatcher.kt
```diff
@@ -0,0 +1,80 @@
...
+        val TYPE_TOKENS = TokenSet.orSet(COMMENTS_OR_WHITE_SPACE_TOKEN_SET,
+            TokenSet.create(
+                RsTokenElementTypes.IDENTIFIER,
```
matklad: There will be a lot of tokens here, so it make sense to import `RsTokenElemntTypes.*`

matklad:
@mkaput you've worked this a bit, does https://github.com/intellij-rust/intellij-rust/pull/923#issuecomment-272612504 sounds reasonable?

mkaput:
If I recall correctly there was some, maybe hacky (getter chain), way to obtain PSI from HighlighterIterator, but I worked on it a while ago and I don't exactly remember how things worked. 

As for now such heuristics are OK. We can take deeper look later on this problem. 

This feature may not have been implemented **yet** for Kotlin because it's rather low priority thing. 

I have also wild idea to make an adapter from highlighter iterator to parser, something like custom PsiBuilder, so we could use our parser to perform some creative queries on HI. What do you think?

matklad:
>I have also wild idea to make an adapter from highlighter iterator to parser, something like custom PsiBuilder, so we could use our parser to perform some creative queries on HI. What do you think?

I am not sure that this is a good idea... Using tokens directly is faster than waiting for the PSI to be parsed.

mkaput:
Yeah, that's why I called this idea it *wild* :p

oistein:
Well that was a handful :) 

> We basically have a single tests, which receives as an input a big string literal with Rust code, which contains all the interesting positive cases.

Can you point to any similar testcases and/or APIs for this? 

> Also, a short comment explaining the issue (we want PSI tree, but we got only highlighting iterator, so let's use heuristics) would be useful!

In test or *Highlighter?

matklad:
>Can you point to any similar testcases and/or APIs for this?

Something similar happens in Stub access test, but it is vastly more complicated there.

I imagine this test like this:

```Kotlin
    fun test() = checkPairs("""
        struct S<T, Foo>;
        // and all other interesting cases
    """)
    
    fun checkPairs(@Language("Rust") code: String) {
        myFixture.configureByText("main.rs", code)
        // need also to check `TypeArgumentList` and maybe some others?
        val params =myFixture.file.descendantsOfType<RsTypeParameterList>() 
        for (param in params) {
            val leftAngleBrackert = param.lt // .textOffset
            val rightAngleBracket = param.gt
            // check that brackets match according to matcher
        }
    }
```

>In test or *Highlighter?

In the highlighter.

matklad:
@oistein I'll try to bring this other the finish line myself!

PR #925, opened 2017-01-13
COMP: Add type of local bindings

slavam2605:
In completion popup added a type for bindings for which `RustTypificationEngine` is able to infer a type.
Example: 
```
let a = 12u32;
let b = |x| x;
<caret>
```
Here completion will show [a ........ u32] and [b ......... ] with no type for b.


matklad:
Thanks!

PR #926, opened 2017-01-13
GRAM, TY: Correct parsing and type inference of 1-tuple

slavam2605:
Following the rust documentation, `(i32)`, `((i32))` and so on are just the `i32` type and `(i32, )` denotes a 1-tuple of `i32`.
With old grammar `(i32)` parsed as 1-tuple of `i32`. This PR fixes parsing and type inference about 1-tuples.


matklad:
Hm, does it passes the `way_too_many_parens` tests? If it does, then perhaps we can simplify `TupleOrParenExpr` in the same fashion? It currently uses an external rule to avoid the exponential time parsing, but it'd be great to rewrite it with simpler rules.

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustTypificationEngine.kt
```diff
@@ -217,7 +217,7 @@ private class RustTypeTypificationVisitor(val pivot: RsType) : RustComputingVisi
             RustUnitType
     }
 
-    override fun visitBaseType(o: RsBaseType) = set {
+    override fun visitBaseType(o: RsBaseType): Unit = o.baseType?.let { visitBaseType(it) } ?: set {
```
matklad: This `ComputingVisitor`s are tricky. By convention, every method should should start with `= set`, to make sure that the result is set exactly once. Can you move this inside the `set`?

slavam2605: I made it inside at first, but than decided to extract for cleaner code.
I can return it back.

slavam2605:
However, tests are not passing now and I thought that it is not my problem. I would prefer to wait until the master branch will pass all tests and then I will test mine.

src/main/kotlin/org/rust/lang/core/types/visitors/impl/RustTypificationEngine.kt
matklad: Yeah, I think calling `RustTypificationEngine.typifyType(baseType)` would be a little bit more clear and consistent with other branches in this visitor.

matklad:
>However, tests are not passing now and I thought that it is not my problem. I would prefer to wait until the master branch will pass all tests and then I will test mine.

Should be OK now.

slavam2605:
@matklad 
I've done with fixing grammar in a such way that it quickly parses all cases like (((i32, ), ), ), (((i32))), (((i32, i32), i32), i32) and so on (with a big number (59) of pairs of brackets, of course)
I'm now trying to cope with hell of merging and I think I will recreate the PR, but I have a question:
I want to extend test "way_too_many_parens". How I should generate .txt file with PSI? I can't find a way to do it automatically.

matklad:
>I've done with fixing grammar in a such way that it quickly parses all cases

Splendid! 

>How I should generate .txt file with PSI? I can't find a way to do it automatically.

You need to delete the existing `.txt` file and run the test. The test would fail, but it would write the updated `.txt` file and the next time it is run, it should pass.

slavam2605:
Ok, my tests failed and I realize what is totally wrong.
To distinguish ( type ) from ( type, ) we need to parse all type and then check if there is a comma.
It can be solved by more straightforward parser which can return different nodes of AST or PSI from the same parsing branch, something like this:
```
parseType() {
    c = peekChar();
    if c is LBRACKET {
        list = List::new();
        loop {
            type = parseType();
            list.push(type);
            c2 = takeChar();
            if c2 is COMMA {
                if peekChar() is RBRACKET: return Tuple(list)
            } else {
                if peekChar() is RBRACKET: 
                    if list.len() == 1: return list[0] else: return Tuple(list)
            }
        }
    } else {
        // parse rest branches...
    }
}
```
But I can't do this with GrammarKit because I hardly understand how all this parsing works, especially stubs generation. If it is possible to write such parser with GrammarKit or to write an external parser for this rule then it will work (and in linear time).
Now the only thing I can is to introduce parser changes that correctly handle 1-tuples and support parens only around a BaseType (like (((i32))) ). But this breaks way_too_many_parens test since there is one case (((((((((    (i32, )   )))))))))) and new parser does not support this. But it is wrong anyway: (((... (1, ) ...))) has a type (i32, ), but the current version of parser thinks that `k` has type (((... i32, ), ), ...), ).

slavam2605:
I found `TupleOrParenExpr` and I think that it is really the only way to do the same thing for types.

matklad:
>I found TupleOrParenExpr and I think that it is really the only way to do the same thing for types.I found 

Yes, external rules is one way forwad. Another option is to try to use `upper`  rules, they allow to make "last-minute" decisions about the node type. See how `ExprStmtOrLastExpr` is done, for example. 

I've tried to apply `upper` to tuples before, but I wasn't successful: they interacted badly with expression parsing. 

The third option would be not to distinguish tuples and parenthesized expressions on the PSI level and instead some method like `isParen: Boolean get() = exprs.size() == 1 && hasTrailingComma` . At the moment I think it's better to handle the distinction in the parser, but I am not 100% sure.

slavam2605:
I'm closing the PR and will try to do it a bit later, when I will get some free time (if it won't be already done).

PR #928, opened 2017-01-14
(GRAM): Unify functions grammar

alygin:
I'd like to commit this part of the functions grammar unification (#864) to make it survive the next wave of the renaming armageddon :)

Free functions, impl and trait functions/methods are unified along with their parameters.

Foreign functions will be the next step.

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -77,7 +77,7 @@
     ]
 
     elementType("(Fn|Anon|Lambda|Path)Parameter") = ValueParameter
-    elementType("(Fn|Lambda|Path|FnType|ForeignFn|Method|MethodDecl)Parameters") = ValueParameterList
+    elementType("(Fn|Lambda|Path|FnType|ForeignFn)Parameters") = ValueParameterList
 
     extends(".+Expr") = Expr
     elementType(".+BinExpr") = BinaryExpr
@@ -238,7 +238,7 @@ fake ValueParameter ::= Pat? Type? {
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-FnParameter ::= Pat TypeAscription
+FnParameter ::= [ Pat ':' ] Type
 LambdaParameter ::= Pat TypeAscription?
 AnonParameter ::= [ RestrictedPat ':' ] Type
 PathParameter ::= Type !'='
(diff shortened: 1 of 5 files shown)


src/main/grammars/RustParser.bnf
```diff
@@ -256,19 +256,15 @@ fake ValueParameterList ::= SelfParameter? ValueParameter* '...'? {
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-FnParameters          ::= '(' !',' [ <<comma_separated_list FnParameter>> ] ')' { pin = 1 }
+FnParameters          ::= '(' !',' ( SelfParameter [ ',' [ <<comma_separated_list FnParameter>> ] ]
+                                   | [ <<comma_separated_list FnParameter>> ] ) ')' { pin = 1 }
```
matklad: I think this two cases can be coalesced: ` [ SelfParameter (',' | &')') ] [ <<comma ... >>] ) `.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -90,6 +93,49 @@ class RsErrorAnnotator : Annotator {
         }
     }
 
+    private fun checkValueParameter(holder: AnnotationHolder, param: RsValueParameter) {
+        val fn = param.parent.parent as? RsFunction ?: return
+        when (fn.role) {
+            RsFunctionRole.FREE -> {
+                require(param.pat, holder, "${fn.title} cannot have anonymous parameters", param)
```
matklad: This and the last branch are identical. So perhaps

```
when (fn.role) {
  RsFunctionRole.Free, RsFunctionRole.IMPL_METHOD -> 
}
```

matklad:
>to make it survive the next wave of the renaming armageddon :)

Hey, I think there's nothing else left to rename :)  Next I am planning to tweak types grammar such that it is easier  to implement stubs.

src/main/grammars/RustParser.bnf
alygin: Yep, that works!

alygin:
I noticed a problem here that I don't know how to deal with. Let's assume we have a simple function definition with an error that the annotator catches properly:

```rust
    default fn def() {}
//  ^^^^^^^ Function `def` cannot have the `default` qualifier
```

So far so good. Then we add another error &mdash; an anonymous parameter &mdash; which is also caught by the annotator:

```rust
    default fn def(u128) {}
               //  ^^^^ Function `def` cannot have anonymous parameters
```

The problem is that the first error is not caught anymore. It happens because in the second case the `default` word isn't treated as a keyword anymore &mdash; it's not highlighted and `RsFunction.default` returns `null`.

@matklad, do you know why it happens that way and how it could be fixed?

matklad:
>do you know why it happens that way and how it could be fixed?

https://github.com/intellij-rust/intellij-rust/commit/32b916f637fb40114423bb1422c57a5c0da1a00d

matklad:
Thanks!

alygin:
@matklad, thanks for helping with the multi-errors problem.
I noticed that it broke the `RsErrorAnnotatorTest.testPaths()` test. It now highlights all the path instead of the single `self`- or `super`-part.

I could change the test to follow this behaviour, but I'm not sure if it should be fixed this way. Actually, I don't understand why it had worked before, because the `checkPath()` function calls `holder.createErrorAnnotation(path, ...)` that should have been highlighted the whole path all the time. So, I'm a bit perplexed by this test.

matklad:
Yeah, I've fixed it in https://github.com/intellij-rust/intellij-rust/commit/2f39046ab30301f8b3bd03dfba3f2ac99af31179.

>So, I'm a bit perplexed by this test.

Paths are nested. Although `foo::bar::baz` looks linear in the source, in PSI tree there are three nested nodes. So previously annotator would annotate the inner most path, and skip its parents if there are errors. Now we force it to skip the parents.

PR #929, opened 2017-01-15
(ANNOT): Annotate duplications at file level

alygin:
@matklad, your [32b916f](https://github.com/intellij-rust/intellij-rust/commit/32b916f637fb40114423bb1422c57a5c0da1a00d) commit has also fixed [the problem](https://github.com/intellij-rust/intellij-rust/pull/901#issuecomment-271079881) with duplication higlights at the file level. So, it's now much more useful.


matklad:
Hm, are you sure? I still don't get file-level highlighting consistently: 

d

alygin:
Hm, it works correctly and very reliably for me:

screen shot 2017-01-15 at 13 44 38

Could it be that something didn't recompile when you started it?

matklad:
I've rechecked and it's definitely  still flickers for me, especially when I add/remove the trailing whitespace in the file.

alygin:
I think you're right, it's not fixed. I don't know what it depends on, but the highlighting works fine in one files and doesn't work in others. I've noticed that the more code I have in a file, the better it works. Sounds weird, but that's how it is :(

I can't explain that, so I'm closing the PR and will carry the investigation further.

PR #930, opened 2017-01-15
(GRAM): Merge function grammar rules

alygin:
The final part of the functions grammar unification (by #864) &mdash; foreign functions have merged into  the `Function` rule.


src/main/grammars/RustParser.bnf
```diff
@@ -256,13 +256,14 @@ fake ValueParameterList ::= SelfParameter? ValueParameter* '...'? {
...
-FnParameters          ::= '(' !',' [ SelfParameter (',' | &')') ] [ <<comma_separated_list FnParameter>> ] ')' { pin = 1 }
+FnParameters          ::= '(' !',' [ SelfParameter (',' | &')') ] [ <<fn_params_list FnParameter>> ] ')' { pin = 1 }
 LambdaParameters      ::= '|' !',' [ <<comma_separated_list LambdaParameter>> ] '|'
 
-ForeignFnParameters   ::= <<variadic_params_impl FnParameter>>
 FnTypeParameters      ::= <<variadic_params_impl AnonParameter>>
 PathParameters        ::= '(' [ <<comma_separated_list PathParameter>> ] ')' { pin = 1 }
 
+
+private meta fn_params_list ::= <<param>> (',' <<param>>)* [ ',' '...'? ] ','?
```
matklad: This `meta` rule is used only once, so you don't need it.

src/main/grammars/RustParser.bnf
```diff
@@ -308,7 +309,7 @@ ExternAbi ::= extern STRING_LITERAL?
 
 RetType ::= '->' Type { pin = 1 }
 
-fake Function ::=
+Function ::=
   OuterAttr* default_? Vis? FnAttrs FnHeader FnParameters FnTail (';' | InnerAttrsAndBlock)
```
matklad: `FnAttrs`, `FnHeader` and `FnTail` are used only once, so we can inline them and split the definition of `Function` over several lines.

They don't add clarity, and were used just to remove duplication between various function flavors.

alygin: OK. Should it be `FnItem` instead of `Function`?

matklad: I like the `Function` more: `Item` suffix does not much information. The whole `Item` concept is too fuzzy because of extern blocks, use declarations and anonymous impls.

matklad:
Nice!

PR #932, opened 2017-01-15
(GRAM): struct / enum variant fields decl grammar unification

alygin:
One more grammar rules unification from #864: this one is for struct / enum variant field declarations, including tuple-like.

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -525,13 +525,14 @@ EnumVariant ::= OuterAttr* identifier VariantArgs? {
 
 private VariantArgs ::= BlockFields | TupleFields | VariantDiscriminant
 
-BlockFields ::= '{' <<field_with_recover FieldDecl>>* '}' {
+BlockFields ::= '{' FieldDecl_with_recover* '}' {
   pin = 1
   extends = "org.rust.lang.core.psi.impl.RsStubbedElementImpl.WithParent<?>"
   stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
-TupleFields ::= '(' <<comma_separated_list TupleFieldDecl >> ')' { pin = 1 }
+
+TupleFields ::= '(' <<comma_separated_list TupleFieldDecl>> ')' { pin = 1 }
 
 VariantDiscriminant ::= '=' AnyExpr { pin = 1 }
 
@@ -555,12 +556,11 @@ TupleFieldDecl ::= AttrsAndVis Type {
   hooks = [ leftBinder = "DOC_COMMENT_BINDER" ]
 }
 
-private meta field_with_recover ::= !'}' <<param>> (',' | &'}') {
+private FieldDecl_with_recover ::= !'}' FieldDecl (',' | &'}') {
   pin = 1
   recoverWhile = Field_recover
 }
 private Field_recover ::= !('}' | '#' | pub | identifier)
-
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 // Trait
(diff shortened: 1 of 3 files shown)


src/main/grammars/RustParser.bnf
```diff
@@ -532,9 +532,6 @@ BlockFields ::= <<mk_block_fileds FieldDecl>> {
 }
 TupleFields ::= <<mk_tuple_fields TupleFieldDecl>>
 
-VariantBlockFields ::= <<mk_block_fileds VariantFieldDecl>> { elementType = BlockFields }
-VariantTupleFields ::= <<mk_tuple_fields VariantTupleFieldDecl>> { elementType = TupleFields }
```
matklad: I think these `mk_` macros are no longer needed?

alygin: Yeah, you're right.

src/main/grammars/RustParser.bnf
```diff
@@ -523,17 +523,15 @@ EnumVariant ::= OuterAttr* identifier VariantArgs? {
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-private VariantArgs ::= VariantBlockFields | VariantTupleFields | VariantDiscriminant
+private VariantArgs ::= BlockFields | TupleFields | VariantDiscriminant
 
-BlockFields ::= <<mk_block_fileds FieldDecl>> {
+BlockFields ::= '{' <<field_with_recover FieldDecl>>* '}' {
```
matklad: And `filed_with_recover` also no longer needed :)  :+1:

src/main/grammars/RustParser.bnf
```diff
@@ -523,17 +523,15 @@ EnumVariant ::= OuterAttr* identifier VariantArgs? {
...
+  pin = 1
   extends = "org.rust.lang.core.psi.impl.RsStubbedElementImpl.WithParent<?>"
   stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
-TupleFields ::= <<mk_tuple_fields TupleFieldDecl>>
-
-VariantBlockFields ::= <<mk_block_fileds VariantFieldDecl>> { elementType = BlockFields }
-VariantTupleFields ::= <<mk_tuple_fields VariantTupleFieldDecl>> { elementType = TupleFields }
+TupleFields ::= '(' <<comma_separated_list TupleFieldDecl >> ')' { pin = 1 }
```
matklad: Nit: extra space before `>>`

src/main/grammars/RustParser.bnf
matklad: To be more precise, we need `FieldDecl_with_recover`, but it don't have to be a macro.

alygin: But will it be possible to preserve the recovery policy that is defined for `filed_with_recover` if we collapse this rule?

A dozen of tests break when I use

```rust
BlockFields ::= '{' ( !'}' FieldDecl (',' | &'}') )* '}'
```

and hundreds break when I try to make it recoverable :)

matklad: This should work

```
BlockFields ::= '{' FieldDecl_with_recover* '}'

 private FieldDecl_with_recover ::= !'}' FieldDecl (',' | &'}') {
    pin = 1
    recoverWhile = Field_recover
  }
  private Field_recover ::= !('}' | '#' | pub | identifier)
  ```

We can't remove this rule entirely, but we can get rid of meta `<<` :)

matklad:
Great! So, #864 can be closed? Or is there something we've missed?

alygin:
I think it's completed. At least, I can't find anything that could be merged the same way.

PR #936, opened 2017-01-15
RES: fix module resolve for case-insensitive file systems

matklad:
take 3

### DIFF
--- a/src/test/kotlin/org/rust/cargo/toolchain/CargoProjectResolveTest.kt
+++ b/src/test/kotlin/org/rust/cargo/toolchain/CargoProjectResolveTest.kt
@@ -1,6 +1,7 @@
 package org.rust.cargo.toolchain
 
 import com.intellij.openapi.command.WriteCommandAction
+import com.intellij.openapi.util.SystemInfo
 import com.intellij.openapi.vfs.LocalFileSystem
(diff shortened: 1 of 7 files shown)


src/test/kotlin/org/rust/cargo/toolchain/CargoProjectResolveTest.kt
```diff
@@ -19,6 +20,10 @@ class CargoProjectResolveTest : RustWithToolchainTestBase() {
     fun testResolveLocalPackageMod() = resolveRefInFile("local_package_mod", "src/bar.rs")
     fun testModuleRelations() = resolveRefInFile("mods", "src/foo.rs")
     fun testKebabCase() = resolveRefInFile("kebab-case", "src/main.rs")
+    fun testCaseInsensitiveMods() {
+        if (!SystemInfo.isWindows) return
+        resolveRefInFile("case_insensitive_mods", "src/BAR.rs")
```
mkaput: Shouldn't this be `bar` lowercase? (just guessing)

matklad: No, it should be the same as the name of the file.

Issue #931, opened 2017-01-15
Gutter icon to open docs.rs documentation of the external crate


matklad:
It'd be great to place an icon to the left of

```
extern crate num_bigint;
```

which would open the browser at https://docs.rs/num-bigint/0.1.35/num_bigint/.

mkaput:
Shouldn't this be rather integrated into <kbd>Ctrl+Q</kbd>?

matklad:
Yeah, we can add a like to `doc.rs` when we render documentation as well, but I think gutter icon is a bit more important because it is more easy to discover.

alygin:
<img width="249" alt="screen shot 2017-01-17 at 01 20 59" src="https://cloud.githubusercontent.com/assets/2101250/22001412/1dea15b8-dc56-11e6-989d-c92842bd7018.png">

or

<img width="258" alt="screen shot 2017-01-17 at 01 22 00" src="https://cloud.githubusercontent.com/assets/2101250/22001415/2259a79e-dc56-11e6-92f5-2596240008de.png">

?

matklad:
I like `?` :)

Issue #914, opened 2017-01-11
Recursive Call Marker 4 Times


Kausta:
When editing a line with recursive call it adds the marker up to 4 times and expands the left column of the editor where the marker is. When the line is deleted it leaves the left column of the editor expanded. 

Problem occurs with Intellij IDEA 2016.2.4.

matklad:
This regressed after https://github.com/intellij-rust/intellij-rust/commit/5782a9adb69c742d1914edf56b6bfe3f6ffa6d5a apparently. I remember checking this with 2016.1, but may be only 2016.2 affected? 

I'll look into this! The workaround is to upgrade to 2016.3 :)

matklad:
>but may be only 2016.2 is affected?

Nopt, 2016.1 is affected as well. Looks I've failed to check that properly.

PR #940, opened 2017-01-16
(ANNOT): Crate docs gutter icon

alygin:
Adds gutter icons that open external crate documentation as described in #931.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -142,6 +142,16 @@ class CargoProjectDescription private constructor(
 
     fun findPackage(name: String): Package? = packages.find { it.name == name }
 
+    fun findLibTargetByNormName(normName: String): Pair<Package, Target>? =
+        packages.asSequence()
+            .mapNotNull { pkg ->
+                pkg.targets.asSequence()
+                    .filter { it.normName == normName }
+                    .map { Pair(pkg, it) }
+                    .firstOrNull()
+            }
+            .firstOrNull()
+
     companion object {
         fun deserialize(data: CleanCargoMetadata): CargoProjectDescription {
             // Packages form mostly a DAG. "Why mostly?", you say.
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -72,34 +72,12 @@ class CargoProjectDescription private constructor(
         LIB, BIN, TEST, EXAMPLE, BENCH, UNKNOWN
     }
 
-    data class ExternCrate(
-        /**
-         * Name of a crate as appears in `extern crate foo;`
-         */
-        val name: String,
-
-        /**
-         * Root module file (typically `src/lib.rs`)
-         */
-        val virtualFile: VirtualFile
-    ) {
-        init {
-            check('-' !in name)
-        }
-    }
-
     private val targetByCrateRootUrl = packages.flatMap { it.targets }.associateBy { it.crateRootUrl }
 
-    private val externCrates: Collection<ExternCrate> get() = packages.mapNotNull { pkg ->
-        val target = pkg.libTarget ?: return@mapNotNull null
-        target.crateRoot?.let { ExternCrate(target.normName, it) }
-    }
-
-    /**
-     * Searches for the `VirtualFile` of the root mod of the crate
-     */
-    fun findExternCrateRootByName(crateName: String): VirtualFile? =
-        externCrates.orEmpty().find { it.name == crateName }?.virtualFile
+    fun findCrateByName(normName: String): Target? =
+        packages
+            .mapNotNull { it.libTarget }
+            .find { it.normName == normName }
 
     /**
      * If the [file] is a crate root, returns the corresponding [Target]
@@ -108,6 +86,8 @@ class CargoProjectDescription private constructor(
         val canonicalFile = file.canonicalFile ?: return null
         return targetByCrateRootUrl[canonicalFile.url]
     }
+
+    fun findPackage(name: String): Package? = packages.find { it.name == name }
 
     fun isCrateRoot(file: VirtualFile): Boolean = findTargetForCrateRootFile(file) != null
 
@@ -139,15 +119,6 @@ class CargoProjectDescription private constructor(
     }
 
     val hasStandardLibrary: Boolean get() = packages.any { it.origin == PackageOrigin.STDLIB }
-
-    fun findPackage(name: String): Package? = packages.find { it.name == name }
-
-    fun findLibTargetByNormName(normName: String): Pair<Package, Target>? =
-        packages.asSequence()
-            .mapNotNull { pkg ->
-                pkg.libTarget?.let { if (it.normName == normName) Pair(pkg, it) else null }
-            }
-            .firstOrNull()
 
     companion object {
         fun deserialize(data: CleanCargoMetadata): CargoProjectDescription {
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
@@ -145,10 +145,7 @@ class CargoProjectDescription private constructor(
     fun findLibTargetByNormName(normName: String): Pair<Package, Target>? =
         packages.asSequence()
             .mapNotNull { pkg ->
-                pkg.targets.asSequence()
-                    .filter { it.normName == normName }
-                    .map { Pair(pkg, it) }
-                    .firstOrNull()
+                pkg.libTarget?.let { if (it.normName == normName) Pair(pkg, it) else null }
             }
             .firstOrNull()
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+        elements.asSequence()
+            .filterIsInstance<RsExternCrateItem>()
+            .forEach { crateItem ->
+                val cargoProject = crateItem.module?.cargoProject ?: return@forEach
+                val (pkg, crate) = cargoProject.findLibTargetByNormName(crateItem.identifier.text)?: return@forEach
+                result.add(LineMarkerInfo(
+                    crateItem.crate,
+                    crateItem.crate.textRange,
+                    RsIcons.DOCS_MARK,
+                    Pass.LINE_MARKERS,
```
matklad: Can you please check that this doesn't cause duplicate markers under 2016.1 (https://github.com/intellij-rust/intellij-rust/commit/55d34bfe511ddfed9ade998db6d998539fea354f)?

src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.rust.lang.core.psi.util.module
+
+/**
+ * Provides an external crate imports with gutter icons that open documentation on docs.rs.
+ */
+class RsCrateDocLineMarkerProvider : LineMarkerProvider {
+
+    override fun getLineMarkerInfo(element: PsiElement): LineMarkerInfo<PsiElement>? = null
+
+    override fun collectSlowLineMarkers(elements: MutableList<PsiElement>, result: MutableCollection<LineMarkerInfo<PsiElement>>) {
```
matklad: Can we use `List<PsiElement>` here?

src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+        elements.asSequence()
+            .filterIsInstance<RsExternCrateItem>()
+            .forEach { crateItem ->
+                val cargoProject = crateItem.module?.cargoProject ?: return@forEach
```
matklad: I think a `for` loop and `continue` would be a little clearer here :)

src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+                val (pkg, crate) = cargoProject.findLibTargetByNormName(crateItem.identifier.text)?: return@forEach
+                result.add(LineMarkerInfo(
```
matklad: Alternatively, I believe you can go fully functional with
```
mySeq.mapNotNullTo(result) {
   ...
}
```

alygin:
Timeout while running `RunConfigurationTestCase` on one of the Travis jobs. Looks unrelated to my changes.

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
```diff
@@ -142,6 +142,16 @@ class CargoProjectDescription private constructor(
 
     fun findPackage(name: String): Package? = packages.find { it.name == name }
 
+    fun findLibTargetByNormName(normName: String): Pair<Package, Target>? =
+        packages.asSequence()
+            .mapNotNull { pkg ->
+                pkg.targets.asSequence()
```
matklad: Should we filter only `lib` targets? And I bet this function is already written somewhere: goto definition works for extern crates, so we must be able to find target by name :)

src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
alygin: Do I understand correctly that the problem with the recursive icon provider was that it added an icon for each call if there were more than one of them? Like in `foo(foo())`.

The doc marker adds two icons if there's two crates on the same line, like 

```rust
extern crate foo; extern crate bar;
```

But each icons has its own link to the corresponding documentation, which I think is expected.

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
alygin: Oh, sure, it lost when I refactored this code. Thanks for pointing this out!

src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
matklad: No, there was issue when the same marker repeats four times: https://github.com/intellij-rust/intellij-rust/issues/914

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
```diff
@@ -142,6 +142,13 @@ class CargoProjectDescription private constructor(
 
     fun findPackage(name: String): Package? = packages.find { it.name == name }
 
+    fun findLibTargetByNormName(normName: String): Pair<Package, Target>? =
+        packages.asSequence()
+            .mapNotNull { pkg ->
+                pkg.libTarget?.let { if (it.normName == normName) Pair(pkg, it) else null }
+            }
+            .firstOrNull()
```
matklad: I think there are two simplifications possible here. We don't need to return a `Pair<Package, Traget>`, because `Target` has a backling back to package: https://github.com/intellij-rust/intellij-rust/blob/bf18a3c19a8db21a8c925a6996e9255c45ccfe02/src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt#L65

Also, this function uses essentially the same algorithm as `findExternCrateRootByName` (list packages, find library targets, filter by normName), so perhaps they can be unified somehow? 

There's `ExternCrate` class nearby, which is similar to `Target`, I think we should use it (and add fields like package name and target). Alternatively, we can jettison `ExternCrate` and just use `Target` instead, with the contract that the kind of target must be `lib`.

src/main/kotlin/org/rust/ide/annotator/RsCrateDocLineMarkerProvider.kt
alygin: Yes, you're right, it's broken on 2016.1. I'll use the fix from the recursive call icons provider.

src/main/kotlin/org/rust/cargo/project/CargoProjectDescription.kt
matklad: Also, you can use `to` function to construct a pair: `pkg to it` :)

alygin: > we can jettison ExternCrate and just use Target instead

I think that would be the right way to go, `ExternCrate` looks redundant.

matklad: And there's one more subtlety I think. Not every crate has documentation on crates.io. For example, local path dependencies don't. I think you can look at `CargoMetadata.Package.source` to learn this.

alygin: > you can look at CargoMetadata.Package.source to learn this

How exactly do you think we can use this property? It can be both empty and filled for documented crates.

matklad: >How exactly do you think we can use this property? It can be both empty and filled for documented crates.

Really? I think it should be `"source": "registry+https://github.com/rust-lang/crates.io-index"` for all creates from crates.io?

alygin: It happened that one of the projects I use to check this feature is [ripgrep](https://github.com/BurntSushi/ripgrep). Besides the executable, it includes five nested libraries that are used as local dependencies (referenced by relative paths, so have `source = null`), but all of them are published on crates.io and have documentation on docs.rs.

Though, we can ignore such cases since we're the authors of the crates and so don't need external documentation for them.

matklad:
LGTM, but tests failures seem legitimate :)

alygin:
Yep, I need to provide valid file paths for the test external crates, but haven't found how to do it correctly yet.

matklad:
>Yep, I need to provide valid file paths for the test external crates, but haven't found how to do it correctly yet.

I think you need to inherit from `RustProjectDescriptorBase` and override `testCargoProject`. I think this should be enough, because you don't need to run a real Cargo.

alygin:
That's what I do in `WithStdlibAndDependencyRustProjectDescriptor`, if I understand you correctly. I need to emulate external dependencies for these tests and they have to point to a correct source files if I want `Target.crateRoot` to return a non-null value in order to bypass [this check](https://github.com/intellij-rust/intellij-rust/pull/940/files#diff-00f69c028b4731d34cd975d459e5757eR25). Unfortunately, `"$contentRoot/main.rs"` (as is used in `RustProjectDescriptorBase.testCargoPackage()`) doesn't work.

matklad:
> I want `Target.crateRoot` to return a non-null value

Are we touching `crateRoot` here? We don't really need it for this particular use case. 

I think you are interested in this source: https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/lang/RsTestBase.kt#L233

alygin:
> Are we touching crateRoot here?

Nope, we are not. That was a temporary distraction )

matklad:
Super!

PR #942, opened 2017-01-17
Sidetable duplicates

matklad:
Seems to work fine on the file level now :tada: 

@alygin please review


matklad:
>Here's a false positive:

Good catch! Should be OK now.

matklad:
>it doesn't emit an error for the first occurrence, only for the subsequent ones.

There's a history to this:  when I first implemented the [first duplicates inspection](https://github.com/intellij-rust/intellij-rust/commit/b0bc03d126f1c165868bef4ef40ccb9cbc9ef3b4), I highlighted only the second and subsequent occurrences because it was the simplest liner time algorithm which could fit in a single loop. Highlighting all occurrences is more consistent with the rest of the platform. Also, I wouldn't be surprised if  `rustc` doesn't highlight the first occurrence for the same reason :) 

Thanks for the review!

alygin:
It looks like rustc [does it deliberately](https://github.com/rust-lang/rust/blob/7d82d95af9c54c3947f6c6e21a6d632d9ee468de/src/librustc_resolve/lib.rs#L3124). But rustc has auxiliary spans that point out to the first occurrence, so I think it makes sense for us to highlight all of them.

PR #947, opened 2017-01-18
(COMPL): Derive completion improvement

alygin:
Minor `derive` attribute completion improvements:

- Complete `derive` with parentheses on `struct`s
- Complete [RustcEncodable](https://doc.rust-lang.org/rustc-serialize/rustc_serialize/trait.Encodable.html) and [RustcDecodable](https://doc.rust-lang.org/rustc-serialize/rustc_serialize/trait.Decodable.html) derivables
- Debug code removed
- `DeriveCompletionProvider` and its test class renamed to get in line with other classes and to make navigation easier

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProvider.kt
@@ -5,17 +5,16 @@ import com.intellij.codeInsight.completion.CompletionProvider
 import com.intellij.codeInsight.completion.CompletionResultSet
 import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.patterns.ElementPattern
-import com.intellij.patterns.ElementPatternCondition
 import com.intellij.patterns.PatternCondition
 import com.intellij.patterns.PlatformPatterns.psiElement
 import com.intellij.psi.PsiElement
 import com.intellij.util.ProcessingContext
 import org.rust.lang.RsLanguage
-import org.rust.lang.core.psi.RsOuterAttr
 import org.rust.lang.core.psi.RsElementTypes.*
+import org.rust.lang.core.psi.RsOuterAttr
 import org.rust.lang.core.psi.util.parentOfType
 
-object DeriveCompletionProvider : CompletionProvider<CompletionParameters>() {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProvider.kt
```diff
@@ -5,25 +5,25 @@ import com.intellij.codeInsight.completion.CompletionProvider
...
 
     private val DERIVABLE_TRAITS = listOf(
         "Eq", "PartialEq",
         "Ord", "PartialOrd",
         "Copy", "Clone",
         "Hash",
         "Default",
-        "Debug"
+        "Debug",
+        "RustcEncodable", "RustcDecodable"
```
mkaput: Maybe we should check whether there is `rustc_serializable` in dependencies?

matklad: I think this is not necessary: you can make some public struct `RustcEncodable`, and actually serialize it in the dependent crate.

alygin: @matklad, but you still have to attach `rustc_serializable` to be able to use `RustcEncodable`.

In this "don't show completions we can't use right away" / "let the user complete what he wants and resolve dependency problems later" dilemma, I have a slight inclination to the latter. I think we shouldn't prevent completion just because the user hasn't added the dependency yet. If he or she has already decided to make a struct De/Encodable, they eventually add the dependency. But that only relates to the standard derivables the compiler knows of. This approach can't be spread to third party derivables like `(De)Serialize` from serde.

alygin: Another option is to make the error annotator check if `rustc_serializable` is added when one of these derivables is used, since the compiler has a special E0433 error for that.

matklad: > I have a slight inclination to the latter. I think we shouldn't prevent completion just because the user hasn't added the dependency yet.

Yes, this is the right approach I think. We need to explore auto importing stuff as well.

matklad:
Thanks!

PR #951, opened 2017-01-19
(INSP): Fix self convention inspection for `into_`

### LINKED ISSUE #950: `mut self` trips lint for `into_*` methods
Using `mut self` (`self` by-value with a mutable binding) in a method that starts with `into` trips the following lint: 

> methods called `into_*` usually take self by value; consider using a less ambiguous name)
> Under construction

alygin:
Makes the Self Convention Inspection handle `mut self` correctly for the `into_*` case, the same way as Clippy works.
Fixes #950.


matklad:
Thanks!

PR #954, opened 2017-01-20
(ANNOT): Detect function param names duplication

alygin:
E0415 detection (from #886).

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -16,6 +16,7 @@ import org.rust.ide.annotator.fixes.ImplementMethodsFix
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.RsFile
 import org.rust.lang.core.psi.impl.mixin.*
+import org.rust.lang.core.psi.util.descendantsOfType
 import org.rust.lang.core.psi.util.module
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -393,6 +399,16 @@ private fun AnnotationSession.duplicatesByNamespace(owner: PsiElement): Map<Name
     return duplicates
 }
 
+private fun PsiElement.allChildren(recursively: Boolean): Sequence<PsiElement> {
```
matklad: `org.rust.lang.core.psi.util.RustPsiNavigationExtensionsKt#descendantsOfType` can help here I think :)

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -49,6 +51,10 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         element.accept(visitor)
     }
 
+    private fun checkPatBinding(holder: AnnotationHolder, binding: RsPatBinding) {
+        binding.parentOfType<RsValueParameterList>()?.let { checkDuplicates(holder, binding, it, true) }
```
matklad: Let's add parameter name here: `recursively = true` for readability.

matklad:
Thanks!

PR #967, opened 2017-01-28
(RES): Resolve conflicts between inherent / trait functions

alygin:
While working on [E0061](https://doc.rust-lang.org/error-index.html#E0061) (wrong arguments number) detection I stumbled upon a problem with methods and associated functions resolve when there's a conflict between the inherent implementation and a trait implementation.

For instance:

```rust
struct Foo;
impl Foo {
    fn bar(&self) {}
}

trait Bar {
    fn bar(&self);
}
impl Bar for Foo {
    fn bar(&self) {}
}

fn main() {
    let foo = Foo;
    foo.bar();  // <-- resolves to Bar::bar() now, but must be resolved to Foo::bar()
}
```

I've fixed the problem for methods, but left associated functions intact so far (though, I've provided an ignored failing test for them).

If this fix looks fine, I'll get back to associated functions. Do I understand correctly that I have to extract something like `RsFnCallReferenceImpl` from `RsPathReferenceImpl` first?

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
@@ -36,7 +36,7 @@ object ResolveEngine {
      * Resolves abstract qualified-path [path] in such a way, like it was a qualified-reference
      * used at [pivot]
      */
-    fun resolve(path: RustPath, pivot: RsCompositeElement, namespace: Namespace? = null): List<RsCompositeElement> {
+    fun resolve(path: RustPath, pivot: RsCompositeElement, namespace: Namespace? = null, single: Boolean = true): List<RsCompositeElement> {
         val start: Sequence<ScopeEntry> = when (path) {
             is RustPath.CrateRelative -> sequenceOfNotNull(
                 pivot.crateRoot?.let { ScopeEntry.of(it) }
@@ -72,7 +72,7 @@ object ResolveEngine {
             current = scope.filter { it.name == name }
         }
 
-        val filteredByNs = if (namespace == null) current else current.filterByNamespace(namespace).take(1)
+        val filteredByNs = if (namespace == null) current else current.filterByNamespace(namespace).take(if (single) 1 else Int.MAX_VALUE)
         return filteredByNs
             .mapNotNull { it.element }
             .toList()
@@ -128,26 +128,22 @@ object ResolveEngine {
         return emptyList()
     }
 
+    fun resolveCallExpr(path: RustPath, element: RsPath, namespace: Namespace?): List<RsCompositeElement> {
+        val fn = resolve(path, element, namespace, false).asSequence()
+            .filterIsInstance<RsNamedElement>()
+            .chooseMajor()
+        return if (fn == null) emptyList() else listOf(fn)
+    }
+
     /**
      * Resolves method-call expressions
      */
     fun resolveMethodCallExpr(call: RsMethodCallExpr): RsNamedElement? {
         val receiverType = call.expr.type
         val name = call.identifier.text
-        var method: RsFunction? = null
-
-        receiverType.getMethodsIn(call.project)
+        return receiverType.getMethodsIn(call.project)
             .filter { it.name == name }
-            .forEach {
-                (it.parent as? RsImplItem)?.let {impl ->
-                    if (impl.traitRef == null) return it
-                }
-                if (method == null) {
-                    method = it
-                }
-            }
-
-        return method
+            .chooseMajor()
     }
 
     fun resolveUseGlob(ref: RsUseGlob): List<RsCompositeElement> = recursionGuard(ref, Computable {
@@ -220,6 +216,19 @@ object ResolveEngine {
         get() = splitToSequence("::")
             .map { RustPathSegment(it, emptyList()) }
             .toList()
+
+    /**
+     * Chooses the major element from the given sequence of candidates. For instance,
+     * a function/method from inherent implementation takes precedence over trait implementations.
+     */
+    private fun Sequence<RsNamedElement>.chooseMajor(): RsNamedElement? =
+        partition { it.isInherent }.let {
+            when {
+                it.first.isNotEmpty() -> it.first[0]
+                it.second.isNotEmpty() -> it.second[0]
+                else -> null
+            }
+        }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
```diff
@@ -485,8 +485,8 @@ class RsResolveTest : RsResolveTestBase() {
     """)
 
     fun testForeignStatic() = checkByCode("""
-        extern "C" { static FOO: i32; }
-                            //X
+        extern "C" { static mut FOO: i32; }
+                                //X
```
matklad: Is this accidental or on purpose?

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
```diff
@@ -591,5 +591,73 @@ class RsResolveTest : RsResolveTestBase() {
                 //^ unresolved
         }
     """)
-}
 
+    fun testMethod() = checkByCode("""
```
matklad: These tests belong to `TypeAwareResolveTest`. The split is done only to make the test file size manageable.

matklad: And I believe there should be tests already for inherent and non-inherent methods, so only the one with conflict should be added.

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
```diff
@@ -134,9 +134,20 @@ object ResolveEngine {
...
         val receiverType = call.expr.type
         val name = call.identifier.text
+        var method: RsFunction? = null
 
-        return receiverType.getMethodsIn(call.project)
-            .find { it.name == name }
+        receiverType.getMethodsIn(call.project)
+            .filter { it.name == name }
+            .forEach {
+                (it.parent as? RsImplItem)?.let {impl ->
```
matklad: Could you add a comment here explaining that if there's a conflict, we want to prefer the inherent method?

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
```diff
@@ -134,9 +134,20 @@ object ResolveEngine {
     fun resolveMethodCallExpr(call: RsMethodCallExpr): RsNamedElement? {
         val receiverType = call.expr.type
         val name = call.identifier.text
+        var method: RsFunction? = null
 
-        return receiverType.getMethodsIn(call.project)
-            .find { it.name == name }
+        receiverType.getMethodsIn(call.project)
```
matklad: This `forEach` with early return and `var method` are not elegant, but I think it's Ok for the time being. I'd like to rethink API around trait-based resolve some day. https://gitter.im/intellij-rust/intellij-rust?at=586d8908058ca96737cc43ea has  interesting ideas.

But perhaps for the problem at hand we can do something like `methods.filter {it.name == name}.partition { it.isInherent }`?

matklad:
>Do I understand correctly that I have to extract something like RsFnCallReferenceImpl from RsPathReferenceImpl first?

Why do you need it?  Aren't inherent function always preferred? (like in `let fn = Foo::bar`)

I think it's better to look at the parent of element if you need to figure the context you are in. I am thinking about *reducing* the number of reference classes, and using `ReferenceExpression` for most things, like Kotlin does.

matklad:
`org.rust.lang.core.resolve.RsTypeAwareResolveTest > testMethodCallOnTraitObject`

fails on travis on 2016.1 because `parent` will cause switch to `AST`. I think it's best just to disable this test on 2016.1: https://github.com/intellij-rust/intellij-rust/commit/4c59102038de2252b150c8c234d39a083632d976#diff-e8a0507a75ea09da38f71fb46c7f066eR189

On 2016.2 and later, stub nodes remember if the parent was stub as well, so `getParent` will work with stubs if possible.

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
alygin: Hm, that's weird. I made it on purpose because we have an annotation "Non mutable static constants are not allowed in extern blocks". So I just made it mutable to remove this error message from the test. But now I see that Rust allows such constants, so it's just a wrong annotation that must be removed, I'll fix it in a separate PR.

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
alygin: Actually, I didn't find any, so decided to add a couple.

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
matklad: > But now I see that Rust allows such constants

I think there was a change to allow them recently.

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
matklad: `testSelfMethodCallExpr` is what I was thinking about.

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
alygin: OK, I'll try that

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
alygin: Sure

alygin:
> Why do you need it?

I just see that `RsMethodCallReferenceImpl` is separated from `RsPathReferenceImpl` and has its own resolve logic (like `ResolveEngine.resolveMethodCallExpr()`), so I thought functions should also  be extracted if I need to resolve them differently. But you're right, I think _"it's better to look at the parent of element"_ will be enough.

> Aren't inherent function always preferred?

They must be preferred but they are not, that's what I'd like to fix.

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
alygin: I also think so, though I didn't see this change announced.

matklad:
>I just see that RsMethodCallReferenceImpl is separated from RsPathReferenceImpl

This is done to separate method calls from fields: in Rust, you can have a field and a method with the same name, so we parse `foo.bar()` not as a call on `foo.bar`, but as a special method call node.

alygin:
> fails on travis on 2016.1 because parent will cause switch to AST. I think it's best just to disable this test on 2016.1

Are all those `java.lang.AssertionError: Access to tree elements not allowed in tests.` in Travis also caused by 2016.1 and should be eliminated the same way?

matklad:
>Are all those java.lang.AssertionError: Access to tree elements not allowed in tests. in Travis also caused by 2016.1 and should be eliminated the same way?

Yes, there's `getParent` in the stack trace. I am thinking of dropping support for 2016.1 altogether: need to recheck this, but 2016.1 is used by a small fraction of people.

matklad:
Thanks!

PR #969, opened 2017-01-28
(ANNOT): Allow foreign mutable static consts

alygin:
Removes the annotation on mutable static constants in `extern` blocks as they are allowed by Rust (discussed [here](https://github.com/intellij-rust/intellij-rust/pull/967#pullrequestreview-18971241)).


matklad:
:+1:

Issue #962, opened 2017-01-24
Intention: if let to match


matklad:
We have match to if-let intention: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt

It would be nice to add the opposite transformation!

Shiroy:
I'm working on this issue.

PR #973, opened 2017-01-29
INT: Add IfLetToMatchIntention

Shiroy:
PR for issue #962 

This intention convert
```rust
if let A(value) = x {
    //Stuff for A
} else if let B(value) = x {
    //Stuff for B
} else {
    //Stuff for something else
}
```
to
```rust
match x {
    A(value) => {
        //Stuff for A
    },
    B(value) => {
        //Stuff for B
    },
    _ => {
        //Stuff for something else
    }
}
```

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/IfLetToMatchIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/IfLetToMatchIntention.kt
@@ -1,0 +1,119 @@
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.parentOfType
+
+class IfLetToMatchIntention : RsElementBaseIntentionAction<IfLetToMatchIntention.Context>() {
+    override fun getText(): String = "Convert if let statement to match"
+    override fun getFamilyName(): String = text
+
+    data class Context(
+        val ifStmt: RsIfExpr,
+        val target: RsExpr,
+        val matchArms: MutableList<MatchArm>,
+        var elseBody: RsBlock?
+    )
+
+    data class MatchArm(
+        val matchArm: RsPat,
+        val body: RsBlock
+    )
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        //1) Check that we have an if statement
+        var ifStatement = element.parentOfType<RsIfExpr>() ?: return null
+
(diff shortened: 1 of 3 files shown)


Shiroy:
Is there a way I can write unit tests for that intention

matklad:
>Is there a way I can write unit tests for that intention?

Yes, they are in /home/matklad/projects/intellij-rust/src/test/kotlin/org/rust/ide/intentions

`doAvailableTest` takes two string literals with Rust code and a caret marker `/*caret*/` and checks that after applying the intention to the first string, we get the second one.

`doUnavailableTest` checks that the intention is not available. 


I haven't reviewed the code in detail, because it is WIP, but it looks good on the first blush! Could you run a **reformat file** action though? We place a space between `if` and `(` in condition :)

src/main/kotlin/org/rust/ide/intentions/IfLetToMatchIntention.kt
```diff
@@ -0,0 +1,118 @@
...
+
+        //Here we have extracted the upper most if statement node
+
+        return extractIfLetStatementIfAny(ifStatement)
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        val (ifStmt, target, matchArms, elseBody) = ctx
+
+        var generatedCode = "match ${target.text} {"
```
matklad: You can use `buildString` for more efficient and idiomatic code:

```
        buildString { 
            append("foo")
            append("bar")
        }
```

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/build-string.html

Shiroy:
I'm done :)

matklad:
Super!

matklad:
(this will be in nightly in 24 hours, and in alpha the next monday).

alygin:
I believe, there must be an intention description in `src/main/resources/intentionDescriptions`. Without it, I got the following error:

```
Assertion failed: Intention Description Dir URL is null: Convert if let statement to match; IfLetToMatchIntention, PluginClassLoader[org.rust.lang, 0.1.0.SNAPSHOT]
java.lang.Throwable
	at com.intellij.openapi.diagnostic.Logger.assertTrue(Logger.java:168)
	at com.intellij.codeInsight.intention.impl.config.IntentionActionMetaData.getDirURL(IntentionActionMetaData.java:109)
	at com.intellij.codeInsight.intention.impl.config.BeforeAfterActionMetaData.getDescription(BeforeAfterActionMetaData.java:155)
```

matklad:
Yep, I need to add a test for this.

matklad:
https://github.com/intellij-rust/intellij-rust/commit/f3ca2f638e501558b9e35d5cbdafbaecae32c733

PR #975, opened 2017-01-30
Mod completion

### LINKED ISSUE #883: Add completion for mod declarations
`mod fo<caret>` should suggest all the files in the current directory.

See `ModDeclReference` and it's `getVariants` method, and `RustCompletionEngine` as well.

Shiroy:
This PR add autocomplete for module declaration statements.

The autocomplete is created by taking all the `.rs` filenames without the extension + all the directory name that contains a `mod.rs` file inside.

Fix #883

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
@@ -2,8 +2,10 @@ package org.rust.lang.core.completion
 
 import com.intellij.codeInsight.completion.InsertionContext
 import com.intellij.codeInsight.lookup.LookupElement
+import com.intellij.codeInsight.lookup.LookupElement.EMPTY_ARRAY
 import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.openapi.editor.EditorModificationUtil
+import com.intellij.psi.PsiFile
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.cargo.project.workspace.cargoWorkspace
 import org.rust.ide.icons.RsIcons
@@ -72,6 +74,24 @@ object CompletionEngine {
             ?.mapNotNull { it.libTarget }
             ?.map { LookupElementBuilder.create(extCrate, it.normName).withIcon(extCrate.getIcon(0)) }
             ?.toTypedArray() ?: emptyArray()
+
+    fun completeMod(mod: RsModDeclItem): Array<out LookupElement> {
+        val directory = mod.containingMod.ownedDirectory ?: return EMPTY_ARRAY
+
+        val currentModuleName = mod.parentOfType<PsiFile>()?.name?.substringBeforeLast('.')
+
+        val modFromRsFile = directory.files
+            .filter { it.name.endsWith(".rs") }
+            .map { it.name.substringBeforeLast('.') }
+            .filter { it != "mod" && it != currentModuleName }
+            .map { LookupElementBuilder.create(it).withIcon(RsIcons.MODULE) }
+
+        val modFromDirectoryModRsFile = directory.subdirectories
+            .filter { it.findFile("mod.rs") != null }
+            .map { LookupElementBuilder.create(it.name).withIcon(RsIcons.MODULE) }
+
+        return (modFromRsFile + modFromDirectoryModRsFile).toTypedArray()
+    }
 }
 
 private fun RsCompositeElement?.completionsFromResolveScope(): Array<LookupElement> =
--- a/src/test/kotlin/org/rust/lang/core/completion/RsCompletionTestBase.kt
+++ b/src/test/kotlin/org/rust/lang/core/completion/RsCompletionTestBase.kt
@@ -11,28 +11,7 @@ abstract class RsCompletionTestBase : RsTestBase() {
 
     protected fun checkSingleCompletion(target: String, @Language("Rust") code: String) {
         InlineFile(code).withCaret()
-        val variants = myFixture.completeBasic()
-
-        fun LookupElement.debug(): String = "$lookupString ($psiElement)"
-        check(variants == null) {
-            "Expected a single completion, but got ${variants.size}:\n" +
-                variants.map { it.debug() }.joinToString("\n")
-        }
-        val normName = target
-            .substringBeforeLast("()")
-            .substringBeforeLast(" {}")
-            .substringAfterLast("::")
-            .substringAfterLast(".")
-        val shift = when {
-            target.endsWith("()") || target.endsWith("::") -> 3
-            target.endsWith(" {}") -> 4
-            else -> 2
-        }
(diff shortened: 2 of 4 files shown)


src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
```diff
@@ -72,6 +73,22 @@ object CompletionEngine {
             ?.mapNotNull { it.libTarget }
             ?.map { LookupElementBuilder.create(extCrate, it.normName).withIcon(extCrate.getIcon(0)) }
             ?.toTypedArray() ?: emptyArray()
+
+    fun completeMod(mod: RsModDeclItem): Array<out LookupElement> {
+        val directory = mod.containingMod.ownedDirectory ?: return EMPTY_ARRAY
+
+        val modFromRsFile = directory.files
+            .filter { file -> file.name.endsWith(".rs") }
```
matklad: It's more idiomatic to use an implicit `it` argument for such short lambdas: `.filter { it.name.endsWith(".rs") }`

src/test/kotlin/org/rust/lang/core/completion/RsModCompletionTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    fn main() {
+
+    }
+
+    //- my_mod.rs
+    pub fn test() {
+
+    }
+    """)
```
matklad: Could you reformat this a bit? Indent the code between `//-` markers:

```
fun testFoo()  = check("""
//- main.rs
    bla-bla-bla
//- foo.rs
    bla-bla-bla
""")
```

Also, let's format empty functions as `fn main() {}` to save the line count.

src/test/kotlin/org/rust/lang/core/completion/RsCompletionTestBase.kt
```diff
@@ -35,6 +35,38 @@ abstract class RsCompletionTestBase : RsTestBase() {
...
 
+    protected fun checkSingleCompletionWithMultipleFiles(target: String, @Language("Rust") code: String) {
+        val files = ProjectFile.parseFileCollection(code)
+        for ((path, text) in files) {
+            myFixture.tempDirFixture.createFile(path, replaceCaretMarker(text))
+        }
+
+        openFileInEditor(files[0].path)
+
+        val variants = myFixture.completeBasic()
```
alygin: From here and down to the end of the method everything looks like a copy of `checkSingleCompletion()`. It seems that this block is a good candidate to be extracted into a separate method to remove duplication.

src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
```diff
@@ -72,6 +73,22 @@ object CompletionEngine {
...
             ?.map { LookupElementBuilder.create(extCrate, it.normName).withIcon(extCrate.getIcon(0)) }
             ?.toTypedArray() ?: emptyArray()
+
+    fun completeMod(mod: RsModDeclItem): Array<out LookupElement> {
+        val directory = mod.containingMod.ownedDirectory ?: return EMPTY_ARRAY
+
+        val modFromRsFile = directory.files
+            .filter { it.name.endsWith(".rs") }
+            .map { it.name.substringBeforeLast('.') }
+            .filter { it != "mod" }
```
alygin: I think we could also filter out the name of the module we're in. Something like `it != "mod" && it != mod.parentOfType<PsiFile>()?.name?.substringBeforeLast('.')`.

Shiroy: For instance, not proposing `mod main;` in `main.rs`

alygin: Yep

alygin: Just not exactly as I wrote, of course. It's better to declare a value with our module name before scanning files for not to do it on every iteration.

Shiroy: How do you how the source code is mapped to the PSI tree ? Is it written in the BNF grammar ?

alygin: Yes, Grammar Kit generate a parser based on the rules defined in the `.bnf` file. The parser, in its turn, produces PSI trees from the source code. [Here's the overview](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md) of how it works and links to the Grammar Kit documentation.

src/test/kotlin/org/rust/lang/core/completion/RsModCompletionTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    //- main.rs
+        mod my_m/*caret*/;
+
+        fn main() {}
+
+    //- my_mod/mod.rs
+        pub fn test() {}
+    """)
+
+    fun testModCompletionSameDirecotryNoModRS() = checkNoCompletionWithMultipleFiles("""
```
alygin: `Direcotry` &mdash; a typo, here and below.

Shiroy: Fix. Thanks for pointing this out.

alygin:
And one more simple improvement: I think we should check if the `mod` is at the top level of the file:

```rust
val file = mod.parent as? PsiFile ?: return EMPTY_ARRAY
val directory = ...
val thisModuleName = file.name.substringBeforeLast('.')
```

This will prevent completion within nested blocks (where such completion makes no sense). It could be covered by a simple no-completion test like `fn foo() { mod my_m/*caret*/; }`.

Shiroy:
I'm not sure. We can also have code like
```rust
mod english {
    mod hello;
    mod farewell;
}
```

In that case, which completion should I propose ?

alygin:
Hm, I believe such code won't work in Rust. You can only have nested modules in such case:

```rust
mod english {
    mod hello { ... }
    mod farewell { ... }
}
```

And here we should suppress completion for `hello` and `farewell` because their code can't be placed into separate files. That was my idea.

Or am I wrong?

Shiroy:
I'll write short Rust to program to test that.

alygin:
Actually, you can have a not in-line module within a block if you use `#[path=...]`. For instance:

```rust
mod english {
    #[path="hello.rs"]
    mod hello;
}
```

But the completion doesn't cover such case because you need to scan files inside the `english` directory to complete it properly. Moreover, the completion then should add the `path` attribute. So, I think we should omit such exotic cases at the moment.

alygin:
Just a bit more details: https://github.com/rust-lang/rust/pull/31534

alygin:
Oh, you're actually right. Your case is this _"in an inline module whose containing file is not a directory owner"_ and it's covered by the completion now. Sorry for wasting your time.

matklad:
Thanks @Shiroy for implementing this and @alygin for a thorough review!

PR #976, opened 2017-01-30
(ANNOT): Wrong arguments number annotation

alygin:
Wrong arguments number detection ([E0061](https://doc.rust-lang.org/error-index.html#E0061) from #886).

I had to skip cases when a function call resolves to a trait implementation to eliminate some tricky false positives (see tests).


matklad:
Those false positives are indeed tricky... Let's hope that we'll get a nice description of trait matching algorithm from http://smallcultfollowing.com/babysteps/blog/2017/01/26/lowering-rust-traits-to-logic/ work. 

Thanks!

PR #978, opened 2017-01-31
(ANNOT): Detect E0061 for tuple-like structs

alygin:
Extends E0061 detection (from #886) to tuple-like structs and enum variants.


matklad:
Thanks!

Issue #896, opened 2017-01-05
Add test for running annotators with rustc source 


mkaput:
Along with #864 and #886 it may be worth to run annotations throughout rustc source (just like we currently do with parser) and ensure we do not output any false positives for error annotations. This will be also perfect performance test.

alygin:
> just like we currently do with parser

Do you mean `RustParserPerformanceTest` or are there other similar tests?

matklad:
There are `RustHighlightingPerformanceTest` and `RustParserPerformanceTest`. This separation is more or less arbitrary at the moment.

alygin:
Just for information: since `RustParserPerformanceTest` uses `checkHighlighting()` &mdash; which among other things also runs annotators &mdash; it already catches false positive annotations in stdlib sources and reports performance metrics.

dima74:
`RsRealProjectAnalysisTest` was added in #3419, so I guess the issue can be closed

PR #979, opened 2017-01-31
(ANNOT): Missing lifetime annotation

alygin:
Detects missing lifetimes ([E0106](https://doc.rust-lang.org/error-index.html#E0106) from #886) in structs, enums and type aliases.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -16,10 +16,7 @@ import org.rust.ide.annotator.fixes.ImplementMethodsFix
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.impl.RsFile
 import org.rust.lang.core.psi.impl.mixin.*
-import org.rust.lang.core.psi.util.descendantsOfType
-import org.rust.lang.core.psi.util.module
-import org.rust.lang.core.psi.util.parentOfType
-import org.rust.lang.core.psi.util.trait
+import org.rust.lang.core.psi.util.*
 import org.rust.lang.core.resolve.Namespace
 import org.rust.lang.core.resolve.namespaces
 
@@ -40,6 +37,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
             override fun visitPatBinding(o: RsPatBinding) = checkPatBinding(holder, o)
             override fun visitPath(o: RsPath) = checkPath(holder, o)
             override fun visitFieldDecl(o: RsFieldDecl) = checkDuplicates(holder, o)
+            override fun visitRefLikeType(o: RsRefLikeType) = checkRefLikeType(holder, o)
             override fun visitTraitItem(o: RsTraitItem) = checkDuplicates(holder, o)
             override fun visitTypeAlias(o: RsTypeAlias) = checkTypeAlias(holder, o)
             override fun visitTypeParameter(o: RsTypeParameter) = checkDuplicates(holder, o)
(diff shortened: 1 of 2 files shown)


alygin:
Added detection of missing lifetimes in return types.

alygin:
Added detection of missing lifetimes in return types of Fn-parameters. Like in the following snippet:

```rust
fn foo(f: &Fn (&u32, &u8) -> &u32) {}
                          // ^ Missing lifetime specifier
```

I hope this is the last touch of this PR, I know no other cases of E0106.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -331,6 +345,20 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     private fun pluralise(count: Int, singular: String, plural: String): String =
         if (count == 1) singular else plural
 
+    private fun canElideReturnLifetime(fn: PsiElement): Boolean {
```
matklad: Hm, not sure that we already can handle this properly: the lifetime might be hidden inside a generic parameter:

```
struct S<'a> {
    foo: &'a u32,
}

fn main() {
    let x = 92;
    let s = S { foo: &x };
    let y = get_ref(s);
    println!("y = {:?}", y);
}

fn get_ref(s: S) -> &u32 {
    s.foo
}
```

alygin: What it we just resolve a non-ref parameter's type and if it has lifetimes we'll count them in `refsCount`? And we also need to provide special treatment to primitive types as they don't resolve at the moment.

matklad: I think this can work, provided that we gracefully handle the case where we fail to resolve. Would be interesting to find some documentation/rustc source code which precisely specifies how lifetimes can be elided. 

We can also scale this PR to fields, which are handled correctly, and solve lifetime elision separately.

alygin: > Would be interesting to find some documentation

[E0106](https://doc.rust-lang.org/error-index.html#E0106) description has formal rules of lifetime elision.

> We can also scale this PR to fields

I totally agree. I'll remove the return types analysis.

alygin: And [here's the rustc code](https://github.com/rust-lang/rust/blob/0f49616a53465d3ee7a7069cd13c7f86a483b655/src/librustc/middle/resolve_lifetime.rs#L1275).

alygin: I also wonder what would be a reliable way to catch such false positives? I always run annotations against several relatively large projects like cargo, hyper etc. But that's obviously not enough. @mkaput suggested #896, but a) I didn't figure out how to implement it, and b) I'm not sure if it'll be enough because rust code allows very convoluted code structures that people don't use in real code.

Maybe we could check annotations against some tests that rust uses? Some (chosen set of) files from [rust/test](https://github.com/rust-lang/rust/tree/b1363a73ede57ae595f3a1be2bb75d308ba4f7f6/src/test)?

alygin:
Rolled back to structs, enums and type aliases.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
matklad: Interesting question!

I like the idea of running annotator on the large code base, like cargo + it's dependencies. I wouldn't worry too much about potentially missing cases, because we are primarily interested in what people actually write. There will be some amount of false positives, bug reports and fixing, but this perfectly OK. And sometimes false positives may actually be on the Rust side, as in the case with naming conventions for static variables. 

I am more worried about the effect on the testing times. Currently, we have similar parsing tests, and they take too much time to my taste. They are valuable of course, but still... I need to figure out how to split our test suite into fast and slow parts. Then, we can run inspections on Cargo as a part of slow tests, and then add any false positives as fast tests.

matklad:
Thanks!

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
alygin: > split our test suite into fast and slow parts

That would be great!

PR #982, opened 2017-02-01
INT: Purity check of expressions, improved simplification of boolean expressions

slavam2605:
First sketch of advanced analisys of expressions and its usage in simplifications.
`expr.isPure()` tries to infer if an expression is functionally pure and throws no errors.
Now expressions like `a || true` are simplified to `true` (since paths like `a` are always pure, so expression `a || true` has no side effects, no panics and `||` is not overloadable).
Expressions like `f() || true` are not simplified because function call can have side effects.
There is a lot of work to do, but I want to know, is a design like this ok or it should be done in the completely another way.

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
+++ b/src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
@@ -1,0 +1,102 @@
+package org.rust.ide.utils
+
+import org.rust.lang.core.psi.*
+
+/**
+ * Returns `true` if all elements are `true`, `false` if there exists
+ * `false` element and `null` otherwise.
+ */
+private fun <T> List<T>.allMaybe(predicate: (T) -> Boolean?): Boolean? {
+    val values = map(predicate)
+    val nullsTrue = values.all {
+        it ?: true
+    }
+    val nullsFalse = values.all {
+        it ?: false
+    }
+    return if (nullsTrue == nullsFalse) nullsTrue else null
+}
+
+/**
+ * Check if an expression is functionally pure
+ * (has no side-effects and throws no errors).
+ *
+ * @return `true` if the expression is pure, `false` if
+ *        > it is not pure (has side-effects / throws errors)
+ *        > or `null` if it is unknown.
+ */
+fun RsExpr.isPure(): Boolean? {
+    return when (this) {
+        is RsArrayExpr -> when (semicolon) {
+            null -> exprList.allMaybe { it.isPure() }
+            else -> exprList[0].isPure() // Array literal of form [expr; size],
+        // size is a compile-time constant, so it is always pure
+        }
+        is RsStructExpr -> when (structExprBody.dotdot) {
+            null -> structExprBody.structExprFieldList
+                .map { it.expr }
+                .allMaybe { it?.isPure() } // TODO: Why `it` can be null?
+            else -> null // TODO: handle update case (`Point{ y: 0, z: 10, .. base}`)
+        }
+        is RsTupleExpr -> exprList.allMaybe { it.isPure() }
+        is RsFieldExpr -> expr.isPure()
+        is RsParenExpr -> expr.isPure()
+        is RsBreakExpr -> false // Changes execution flow
+        is RsContExpr -> false  // -//-
+        is RsRetExpr -> false   // -//-
+        is RsTryExpr -> false   // -//-
+        is RsPathExpr -> true   // Paths are always pure
+        is RsQualPathExpr -> true
+        is RsLitExpr -> true
+        is RsUnitExpr -> true
+
+        // TODO: more complex analysis of blocks of code and search of implemented traits
+        is RsBinaryExpr -> null // Have to search if operation is overloaded
+        is RsBlockExpr -> null  // Have to analyze lines, very hard case
+        is RsCastExpr -> null;  // `expr.isPure()` maybe not true, think about side-effects, may panic while cast
+        is RsCallExpr -> null   // All arguments and function itself must be pure, very hard case
+        is RsForExpr -> null    // Always return (), if pure then can be replaced with it
+        is RsIfExpr -> null
+        is RsIndexExpr -> null  // Index trait can be overloaded, can panic if out of bounds
+        is RsLambdaExpr -> null
+        is RsLoopExpr -> null
+        is RsMacroExpr -> null
+        is RsMatchExpr -> null
+        is RsMethodCallExpr -> null
+        is RsRangeExpr -> null
+        is RsUnaryExpr -> null  // May be overloaded
+        is RsWhileExpr -> null
+        else -> null
+    }
+}
+
+/**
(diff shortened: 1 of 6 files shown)


matklad:
Yeah, I am a bit worried about this purity concept, because it is not something we can check reliably.

And, even you have some impure function, I think it makes sense to flag code like `foo() || true` anyway.

For example, Kotlin does suggest to simplify this:

```
fun foo(): Boolean {
    println("Hello")
    return false
}

fun main(args: Array<String>) {
    val x = foo() || true
}
```

Also, it's impossible to overload boolean operators in Rust, only bitwise ones (is this correct?), so I think we can just use partial constant evaluation here.

slavam2605:
Yes, indeed impurity is something that is hard to determine. But there are some cases where such transformation is 100% safe, like `a || true` or `12 > 13 && 12 > 13`. I think it would be great to have at least some private mechanism to determine if an expession is completely pure and then make some weak warning that the expression can be simplified.
What do you think about it?

matklad:
>But there are some cases where such transformation is 100% safe, like a || true or 12 > 13 && 12 > 13.

Hm, why do you need purity for this? It looks like plain constant folding.

slavam2605:
@matklad 
If some expression like `foo()` has side effects then it is not equivalent to transform `foo() || true` into `true`.
And if some expression is not pure because it can return different values then `randomBool() || randomBool()` is not equivalent to `randomBool()`.
But for pure expressions such transformations are equivalent (`a || a` → `a`, `a && a ` → `a`, `a || true` → `true` and so on).

mkaput:
>If some expression like foo() has side effects then it is not equivalent to transform foo() || true into true.

I think doing purity analysis on entire functions and storing results in stubs would be very helpful, not only here.

matklad:
>If some expression like foo() has side effects then it is not equivalent to transform foo() || true into true.

It's not equivalent, but I think we should suggest this transformation anyway, like Kotlin does.

slavam2605:
I agree that we can make an action for this.
I will rewrite intention according to this.
But I think it would be great if we will annotate pure simplifiable expressions with weak_warning.

matklad:
>But I think it would be great if we will additionally annotate pure simplifiable expressions with weak_warning.

We should annotate impure ones as well :)  I doubt that anyone really intended to write `foo() || true`. `{foo(); true}` would be a much clearer version.

slavam2605:
Updated, first stub for purity and simplification of `a || true` and so on.
I also wanted to simplify cases like `a || a` and `a && a`, but psi.isEquivalentTo() makes something different than I've expected. Is there a simple way to determine that two expressions are equivalent at least at AST level?

slavam2605:
Update: added an inspection for simplifying pure boolean expressions + some refactorings.
@matklad Is it ok now or I should fix something?

matklad:
Hm, there's RsSimplifyBooleanExpressionInspection.kt and SimplifyBooleanExpressionInspection.kt in ths PR. I think one is extra? :)

matklad:
I am still not convinced that we need purity analysis at this stage. My arguments are:

* More or less precise purity analysis is hard to do, there's no `pure` function annotation in Rust (there used to be one, before the borrow checker was implemented though :) )

* I think expressions should be simplified regardless of purity: if you use expression solely for it's side effects, then using `{}` is much better solution.

slavam2605:
There are Inspection and Intention, because an intention simplifies all expressions, but an Inspection annotates and fix only pure boolean expressions.
Alt-Enter action simplifies all expressions regardless of purity.
Only 100% safe to simplify expressions are annotated with warning and have a quick-fix to simplify them.

src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
```diff
@@ -0,0 +1,212 @@
...
+                                "true" -> return leftExpr to true
+                                "false" -> return createPsiElement(project, "false") to true
+                            }
+                        OROR ->
+                            when (it.text) {
+                                "true" -> return createPsiElement(project, "true") to true
+                                "false" -> return leftExpr to true
+                            }
+                    }
+                    {}
```
matklad: I think it would be slightly better to write `Unit`, which is a synonym of Rust's `()`. `{}` is a trivial function.

matklad: Can we somehow remove duplication between these two branches?

src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
```diff
@@ -0,0 +1,212 @@
...
+            else -> null // TODO: handle update case (`Point{ y: 0, z: 10, .. base}`)
+        }
+        is RsBinaryExpr -> when (operatorType) {
+            ANDAND, OROR -> listOf(left, right!!).allMaybe(RsExpr::isPure)
+            else -> null // Have to search if operation is overloaded
+        }
+        is RsTupleExpr -> exprList.allMaybe(RsExpr::isPure)
+        is RsFieldExpr -> expr.isPure()
+        is RsParenExpr -> expr.isPure()
+        is RsBreakExpr -> false // Changes execution flow
```
matklad: You can write condense this a little, like this: `is RsTupleExpr, is RsFieldExpr` ...

matklad:
Looks nice, though I am slightly worried about amount of code here...

matklad:
Could you try running this inspection manually on some large Rust code bases to test that everything works OK? A good test would be to open `~/.cargo/` folder and to run this inspection on all downloaded crates. 

I am also worried about the performance implications of this check: could it be the case that we traverse some deep expressions over and over?

slavam2605:
How can I run an inspection on some file?
Open it in IDEA or somehow else?

matklad:
>How can I run an inspection on some file?
Open it in IDEA or somehow else?

Yeah, I mean just a manual testing: open the directory with rust code in IDEA and then use **run inspection by name** on the directory.

slavam2605:
I've opened $HOME$/.cargo and run an inspection over all files -- "No suspicious code found. 1 161 files processed in 'Project 'src'', completed in several (~3-5) seconds

slavam2605:
Update: refactored.

src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
slavam2605: Yes, good idea.

src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
slavam2605: After refactoring Unit even not needed. Code have become much nicer)

slavam2605:
@matklad Is it ok for now or something else is needed?

matklad:
Yeah, looks great now, thanks! Sorry for the slow replies and thanks for the heads up: loads of stuff happened recently :)

PR #983, opened 2017-02-01
(ANNOT): Detect undeclared lifetimes

alygin:
Detects usage of undefined lifetimes ([E0261](https://doc.rust-lang.org/error-index.html#E0261) from #886). Covers cases for structs / enums / type aliases.

It could be easily extended to type arguments, but I faced a problem I don't know how to solve. In the following code:

```rust
trait Foo<'a> {
    type It = &'a u8;
}
struct Bar;
impl<'a> Foo<'a> for Bar {}
         // ^^^^ the problem is here
```

the marked part (`<'a>`) is represented by a `RsTypeArgumentList`, but in the PSI tree it's just a single leaf node that has no children. Both its `assocTypeBindingList` and `typeReferenceList` properties are empty. The PSI tree viewer shows me three elements under the `TYPE_ARGUMENT_LIST` (`<`, `lifetime`, and `>`), but when I analyse this arguments list in code, it returns no children.

Is there something that should be done in the grammar to make it work? Or am I missing something?

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -306,8 +306,34 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     }
 
     private fun checkRefLikeType(holder: AnnotationHolder, type: RsRefLikeType) {
-        if (type.needsLifetime()) {
-            require(type.lifetime, holder, "Missing lifetime specifier [E0106]", type.and ?: type)
+        val owner = type.typeOwner
+        val needLifetime = owner is RsFieldDecl || owner is RsTupleFieldDecl || owner is RsTypeAlias
+        if (!needLifetime) return
+
+        val lifetime = type.lifetime
+        if (lifetime == null) {
+            holder.createErrorAnnotation(type.and?.textRange ?: type.textRange, "Missing lifetime specifier [E0106]")
+        } else {
+            checkLifetimeDeclaration(holder, lifetime, owner)
+        }
+    }
+
+    private fun checkLifetimeDeclaration(holder: AnnotationHolder, lifetime: PsiElement, owner: PsiElement) {
+        val name = lifetime.text
+        if (name == "'static") return
+        val declaredLifetimes = owner.branch
+            .mapNotNull {
+                when (it) {
+                    is RsStructItem -> it.typeParameterList?.lifetimeParameterList ?: emptyList()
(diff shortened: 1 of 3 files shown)


matklad:
Hm, looks like you've implemented reference resolution and go to definition for lifetimes :) ! What do you think about moving this logic to new `RsLifetimeReference` class, so that `Ctrl+B` works for lifetimes, and implementing  this inspection as `lifetime.refernece().resolve() == null`?

alygin:
Oh, that's cool :) Sure, I'll do that. I'll close this PR and break it into two parts to keep them focused.

But what about the problem with `RsTypeArgumentList`? I don't get why it behaves this way.

alygin:
@matklad, do I understand correctly that I also need to implement `RsLifetimeDecl` that will be `RsReferenceElement` (and probably `RsNamedElement`)?

matklad:
Oups, sorry, missed your comment, there's an avalanche of discussion around me recently :)

Yeah, you need to modify the PSI structure for the resolve, and it should solve your question about `RsTypeArgumentList` as well. Currently, `lifetime` is just a leaf token, and GK won't generate accessors for leaf tokens with "many" cardinality (i.e, there may be `getLifetime`, but no `getLifetimes`).

So, you need to introduce two new composite nodes: `LifetimeDefinition` which is a `PsiNamedElement` and occurs in `TypeParameterList` and `Label` elements, and `LifetimeReference`, which is present in `TypeArgumentList`, and the break/continue, goto (haha :) ). 

Syntactically, both definition and reference are the same, they are just `lifetime`, semantically, they are different. `Ctrl+B` on reference will bring you to the corresponding def, `Ctrl+B` on the def should show usages.


When implementing this, please don't add lifetimes directly to `lexicalDeclarations`. It is a mistake that `lexicalDeclarations` now handles both types and values namespace, and the filtering is done on the top level. The proper thing to do is to push the namespace declaration down to `lexicalValuesDeclarations`, `lexicalTypesDeclaratinos`, because it'll help to avoid loading stubs when resolving type parameters. Currently this causes this problem: https://github.com/intellij-rust/intellij-rust/blob/0c58f9f914bad73f5a909ac1916aefbd30996748/src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt#L316-L328

alygin:
> Oups, sorry, missed your comment, there's an avalanche of discussion around me recently :)

No worries at all, I've noticed that these days have been hectic. Popularity doesn't come cheap :)

Thanks for the explanation! It looks like this annotation has brought me to yet another area I haven't explored before. So, I need some time to digest it all.

PR #990, opened 2017-02-07
INSP: self in static context inspection [E424]

sirgl:
The self keyword was used in a static method inspection.


src/main/kotlin/org/rust/ide/inspections/RsSelfInStaticContextInspection.kt
```diff
@@ -0,0 +1,32 @@
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.impl.mixin.isAssocFn
+import org.rust.lang.core.psi.util.parentOfType
+
+class RsSelfInStaticContextInspection : RsLocalInspectionTool() {
```
alygin: All the Rust error detection is now performed in the [RsErrorAnnotator](https://github.com/intellij-rust/intellij-rust/blob/28b61a3e809a1034f6cfb5727a023d0f4a00e7c7/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt) class. I think it would be more convenient to move it there.

src/main/kotlin/org/rust/ide/inspections/RsSelfInStaticContextInspection.kt
```diff
@@ -0,0 +1,32 @@
...
+        if (path.self == null) {
+            return
+        }
+
+        val function = path.parentOfType<RsFunction>() ?: return
+
+        if (function.isAssocFn) {
+            holder.registerProblem(
+                path,
+                "The self keyword was used in a static method [E424]",
```
alygin: It's [E0424]. And I think the real Rust message differs from this one because you actually can use the `self` keyword in a static context, just not everywhere (see the comment about a false positive).

src/main/kotlin/org/rust/ide/inspections/RsSelfInStaticContextInspection.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitPath(path: RsPath) = inspectPath(holder, path)
+        }
+
+    private fun inspectPath(holder: ProblemsHolder, path: RsPath) {
+        if (path.self == null) {
+            return
+        }
```
alygin: It could be simplified: `path.self ?: return`

matklad:
Nice!

I think this has one false positive at the moment: what if `self` is used to refer to a module, like this:

```
fn foo() {
}

fn bar() {
    self::foo()
}
```

sirgl:
Fix it and add quick fix to add self

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -60,6 +63,18 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         if ((child == null || child.asRustPath != null) && path.asRustPath == null) {
             holder.createErrorAnnotation(path, "Invalid path: self and super are allowed only at the beginning")
         }
+
+        path.self ?: return
+        val function = path.parentOfType<RsFunction>() ?: return
+        function.isInherentImpl
```
alygin: Looks like an unnecessary line

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -60,6 +63,18 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
+        val rustPath = path.asRustPath
+        if (rustPath != null && rustPath !is RustPath.ModRelative && function.isAssocFn) {
+            val error = "The self keyword was used in a static method [E0424]"
+            val annotation = holder.createErrorAnnotation(path, error)
+            if (function.role == RsFunctionRole.IMPL_METHOD) {
```
alygin: I think this `if` is unnecessary here. The fix could also work for traits:

```rust
trait Foo {
    fn foo(&self);
    fn bar() { self.foo(); }
}
```

src/main/resources/inspectionDescriptions/RsSelfInStaticContext.html
```diff
@@ -0,0 +1 @@
+Checks if self was used in static function.
```
alygin: This file can be removed now

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -10,6 +10,10 @@ import org.rust.lang.core.psi.impl.mixin.valueParameters
 import org.rust.lang.core.psi.util.childOfType
 
 class RsPsiFactory(private val project: Project) {
+    fun createSelf(): RsSelfParameter =
+        createFromText<RsFunction>("fn main(self){}")?.selfParameter
+            ?: error("Failed to create self element")
```
alygin: I think ideally we should let the user chose between `self`, `&self` and `&mut self`, but as a basic implementation I would expect `&self` to be added, it seems much more common. What do you think?

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -784,4 +786,36 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase() {
             <error>pub</error> const C: u32 = 10;
         }
     """)
+
+    @Test
```
alygin: You don't need `@Test` attributes here. It's enough that the method names start with `test`.

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
sirgl: Ok, I'll try to do it like in Kotlin when introduce val/var

matklad: Yep, `&self` seems like a reasonable default.

sirgl:
According to chat decided to left only &self

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -60,6 +63,17 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
             holder.createErrorAnnotation(path, "Invalid path: self and super are allowed only at the beginning")
         }
+
+        path.self ?: return
+        val function = path.parentOfType<RsFunction>() ?: return
+        val rustPath = path.asRustPath
+        if (rustPath != null && rustPath !is RustPath.ModRelative && function.isAssocFn) {
+            val error = "The self keyword was used in a static method [E0424]"
+            val annotation = holder.createErrorAnnotation(path, error)
+            if (function.role != RsFunctionRole.FREE) {
```
alygin: I would guess this is always `true` because the condition `... && function.isAssocFn` filters free functions out. Does it?

sirgl: nope, it passes free functions in this case

alygin: @matklad, it it intentional? I would expect `fn.isAssocFn` to return `false` if it's not an impl or a trait function.

matklad: No, I believe this is a bug. `isAssoc` should be implemented in terms of `role` (or perhaps we can get rid of it altogether),  it should have a longer name, and we should drop `isStatic` field from the stub.

sirgl: it was very helpful that there is a function that simply watches for self keyword, btw

matklad: I'll fix this issue myself then, it's likely that we misuse `isAssocFn` elsewhere. Thanks @alygin and @sirgl for spotting this issue :+1: .

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -1,5 +1,7 @@
 package org.rust.ide.annotator
 
+import org.junit.Test
```
matklad: Unused imoprt

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -784,4 +786,35 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase() {
...
+        fn foo() {
+        }
+
+        fn bar() {
+            self::foo()
+        }
+    """)
 }
+
+data class A (val s: String)
```
matklad: This probably should not be here.

src/test/kotlin/org/rust/ide/annotator/fixes/AddSelfTest.kt
```diff
@@ -0,0 +1,44 @@
+package org.rust.ide.annotator.fixes
+
+import org.junit.Test
+import org.rust.ide.annotator.RsAnnotatorTestBase
+
+class AddSelfTest : RsAnnotatorTestBase() {
+    @Test
+    fun testNoParameters() = checkQuickFix("Add self to function", """
```
matklad: No need to annotate tests with `@Test`. Also, we are switching to the more readable test names format, and I suggest we use it here:

```Rust
fun `test no parameters`() = check(..)
```

src/test/kotlin/org/rust/ide/annotator/fixes/AddSelfTest.kt
```diff
@@ -0,0 +1,44 @@
...
+""", """
+struct S;
+
+impl S {
+    fn foo(&self, a: i32) {
+        self/*caret*/;
+    }
+}
+""")
+}
```
matklad: Could you indent the literals here, using the pattern like in https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/actions/RsJoinRawLinesHandlerTest.kt#L4 ?

matklad:
Thanks @sirgl !

Issue #988, opened 2017-02-05
Broken terminal output on Linux


jonas-schievink:
This is the output I get (happens with every project):

![1486338147](https://cloud.githubusercontent.com/assets/1786438/22631004/6425df44-ec05-11e6-94cd-b40c7ef6d4fb.png)

Not sure where the extra characters come from, might be an issue with my system, but I thought I'd file an issue in case someone can reproduce.

According to a [quick search](https://unicode-table.com/en/000F/), these are "Shift In" characters (U+000F), which don't seem to belong there.

matklad:
We supply the `TERM` environmental variable to Cargo by default to force it to produce colored output: https://github.com/intellij-rust/intellij-rust/blob/281baf3bec9459d821007f1ec8a1fd7b4fa72cc8/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt#L88

Not sure what's going on here, but looks like somehow Cargo uses the wrong escape codes here. 

You can explicitly set `TERM` environment variable and `--color` option in the Run Configuration to workaround this issue. 


Just to make sure, do you get the same extra characters if you run a simple "hello world" project, without the rocket?

la10736:
I didn't report this issue too but I experimented the same behavior on my ubuntu 16:10 at home... on 16:04 at office work well.

jonas-schievink:
> Just to make sure, do you get the same extra characters if you run a simple "hello world" project, without the rocket?

Yep, same thing.

@matklad The code you linked seems to suggest that it will overwrite the `TERM` variable with `linux` (and a test confirms that), so I can't change it in the Run Configuration. I don't know much about terminfo, but "linux" seems like a weird default (that... looks like the in-kernel console).

alygin:
@la10736, could you, please, open a terminal and run the following commands on both your computers (at home and at office) within a rust project directory (any project will do, though ideally it should be the same project):

```
script -q cargo.log
TERM=linux cargo build --color=always
<ctrl+d>
```

and upload here the resulting files `cargo.log`?

That should help to find out whether it's `cargo`/`rustc` emit different output or the IDE handles the same output differently.

la10736:
@alygin I tried at office and on ah fresh hello word project `cargo.log` is empty... I' ll try at home this evening.

alygin:
[...]

PR #991, opened 2017-02-07
Fix TERM environment variable setting

jonas-schievink:
Fixes the issues reported in #988 (at least on Arch Linux).

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
@@ -82,7 +82,7 @@ class Cargo(
             && command in COLOR_ACCEPTING_COMMANDS
             && additionalArguments.none { it.startsWith("--color") }) {
             cmdLine
-                .withEnvironment("TERM", "linux")
+                .withEnvironment("TERM", "xterm+256color")
                 .withRedirectErrorStream(true)
                 .withParameters("--color=always") // Must come first in order not to corrupt the running program arguments
         }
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -73,25 +73,24 @@ class Cargo(
...
-            .withEnvironment(CargoConstants.RUSTC_ENV_VAR, pathToRustExecutable)
-            .withEnvironment(environmentVariables)
 
         // Make output colored
         if ((SystemInfo.isLinux || SystemInfo.isMac)
             && command in COLOR_ACCEPTING_COMMANDS
             && additionalArguments.none { it.startsWith("--color") }) {
             cmdLine
-                .withEnvironment("TERM", "linux")
+                .withEnvironment("TERM", "xterm+256color")
```
alygin: Unfortunately, switching to `xterm+256color` breaks coloring on macs. It looks like we need to provide different terminal types for different platforms.

jonas-schievink: *sigh* I knew it wasn't going to be this easy :)

jonas-schievink:
@alygin Can you (or someone else) take a look at the latest commit I just pushed? I don't have a Mac to test this on. We might want to set `TERM` to a value that makes more sense on Macs.

alygin:
It works on my mac

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
alygin: `linux` uses the same escape codes as `ansi`, while `xterm-256color` differs significantly from them.

BTW, I think we could also try `ansi` as a common ground. It works on mac.

matklad:
Thanks a lot @jonas-schievink and @alygin !

PR #993, opened 2017-02-08
(ANNOT): Ignore raw pointers in E0106 annotation

### LINKED ISSUE #992: missing lifetime on raw pointers
sometimes the IDE underlines a raw pointer member of field with red and says missing lifetime parameter.

alygin:
Fixes #992

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -326,6 +326,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     }
 
     private fun RsRefLikeType.needsLifetime(): Boolean {
+        if (mul != null) return false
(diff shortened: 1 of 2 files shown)


matklad:
Thanks!

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -326,6 +326,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     }
 
     private fun RsRefLikeType.needsLifetime(): Boolean {
+        if (mul != null) return false
```
matklad: Based on discussion in https://github.com/intellij-rust/intellij-rust/commit/02c242ce0653609eb0c271993dbd5f69e25f1f46#commitcomment-20830146, I think that perhaps instead of `mul != null` check, you should check if perhaps the owning `RsTypeReference` type is `RustReferenceType`?

alygin: OK, I'll try this approach

Issue #944, opened 2017-01-17
Replace `.unwrap()` with `?` intention


matklad:
Nothing fancy, no need to change type of anything, just a simple text substitution.

mkaput:
>  just a simple text substitution

Is it? You need to ensure that the block returns type of unwrapped expression imho

mkaput:
Or maybe I'm to strict 😉

matklad:
>Is it? You need to ensure that the block returns type of unwrapped expression imho

We'll do that in a couple of years ;)

PR #995, opened 2017-02-10
INT: Add UnwrapToTryIntention

slavam2605:
PR for issue [#944](https://github.com/intellij-rust/intellij-rust/issues/944)


src/main/resources/intentionDescriptions/UnwrapToTryIntention/before.rs.template
```diff
@@ -0,0 +1 @@
+let a = <spot>Some(12).unwrap()</spot>
```
matklad: `?` is not available for `Option` yet, let's change `Some` to `Ok`?

src/main/kotlin/org/rust/ide/intentions/UnwrapToTryIntention.kt
```diff
@@ -0,0 +1,31 @@
...
+            methodCall.typeArgumentList == null &&
+            methodCall.valueArgumentList.exprList.isEmpty()) {
+            return methodCall
+        }
+        return null
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: RsMethodCallExpr) {
+        val newPsiElement = RsPsiFactory(project).createExpression("a?")
+        (newPsiElement as? RsTryExpr)?.expr?.replace(ctx.expr) ?: return
```
matklad: In this cases, I think it's better to through error (which contains the troublesome expression text) than to silently do nothing. 

If the user sees an exception and a stack trace, she is more likely to send a bug report, and we want bug reports to fix the bugs. 

So, lets add `createTryExpression(inner: RsExpr): RsTryExpr` to the `RsPsiFactory`.

slavam2605: I should throw an exception instead of `?: return`? I just can't understand why you're talking about `createTryExpression`. Or I shoud add some error throwing in the rust code?

matklad: Sorry, I've conflated two thoughts here.

* Yeah, `?: error("Faild to foo expression ${bar.text}")` is better than `?: return`

* Instead of casting an expression to try expression, let's add a function to `RsPsiFactory` which construct `RsTryExpression`.

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -18,6 +18,15 @@ class RsPsiFactory(private val project: Project) {
         createFromText("fn main() { $text; }")
             ?: error("Failed to create expression from text: `$text`")
 
+    fun createTryExpression(expr: RsExpr): RsTryExpr {
+        val newElement = createExpression("a?")
+        if (newElement is RsTryExpr) {
+            newElement.expr.replace(expr)
+            return newElement
+        }
+        error("Failed to create try expression ${expr.text}?")
```
matklad: Could you add backtics around `${expr.text}`, as is done in other cases in this file? 

This is useful, because it will nicely delimit trailing/leading whitespace, if there's any (Rust's `Debug` for strings which encloses the string in `"` and escapes `\n` is a better solution for this problem, but there isn't `Debug` in Java).

slavam2605: Done.

matklad:
Splendid! Looks like both Travis and AppVeyor are having a bad day today (or is it GitHub), so I'll just merge this, hope you've run the tests locally ;)

alygin:
> so I'll just merge this, hope you've run the tests locally

@matklad, `bors` is looking indignantly at you :grinning:

PR #996, opened 2017-02-11
GRAM + RES: Resolve lifetimes and labels

alygin:
Resolve lifetimes and loop labels. A bit of context [here](https://github.com/intellij-rust/intellij-rust/pull/983#issuecomment-276801669).

I don't like this `LifetimeReference` name much because it's long and leads to things like `RsLifetimeReferenceReferenceImpl`. Maybe, it's better to rename `lifetime` to `lifetime_identifier` (it's also long, but it's not used that much) and use the `Lifetime` name for lifetime references? Or maybe, my implementation can be changed somehow to evade those awkward names? Or am I exaggerating and it looks ok?

And I think we should separate lifetimes and loop labels completely? They only share the naming convention, otherwise they are absolutely different elements. @matklad, what do you think?


alygin:
I've just encountered an exception that relates to these changes. It says `Stub and PSI element type mismatch in osstringext.rs: stub RsPlaceholderStub, AST LIFETIME_DECL; Element(LIFETIME_DECL)`.

I don't fully understand how stubs work and can't find the cause of the problem quickly, so I'll just mark the PR as WIP until I figure out what that means and how to fix it.

alygin:
Oh, got it, `stubVersion`.

alygin:
I've separated lifetimes and loop labels. There're now `LifetimeReference` &#8594; `LifetimeDecl` and `Label` &#8594; `LabelDecl` without resolve conflicts.

matklad:
@alygin as the author of PRs, what form of review is more convenient to you:

* batching all comments into a single GitHub review

* a number of "single comment"s

? 

I've just noticed that I've forgotten to click "finish the review" button, although I have made this comments earlier =/

alygin:
That's really no big deal, I can easily imagine situations when one works much better than another and vice versa. Just choose the way that feels better for the particular PR at the particular moment. I'm totally OK with both single phrases thrown at me randomly and comments batched into a reviews the size of short novel )

alygin:
@matklad, what do you think of changing the `lifetime` token name to something unrelated to lifetimes exclusively? I'm a bit irritated by two problems that this token type brings: we have a `LabelDecl.lifetime` property which is just an identifier, not a lifetime, and we have to use `LifetimeReference` (let alone the ugly `LifetimeReferenceReference`) instead of shorter `Lifetime` (as in `Label`) just to evade name collision. 

There's the `identifier` token type and the `lifetime` is also a sort of identifier that can be used for lifetimes and labels (for now), so maybe it should be named appropriately? Unfortunately, I can't come up with an obvious and good name for it, but maybe something like `l_identifier` (`l` from **L**ifetime and **L**abel) or `q_identifier` (`q` from **Q**uoted identifier) will do?

matklad:
>what do you think of changing the lifetime token name to something unrelated to lifetimes exclusively?

That's reasonable! `l` and `q` prefixes are ok, but I would prefer something more self-describing, `quote_identifier` perhaps?

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
```diff
@@ -212,6 +209,34 @@ object ResolveEngine {
         return module.project.getPsiFor(module.cargoWorkspace?.findCrateByName(name)?.crateRoot)?.rustMod
     }
 
+    fun resolveLabel(label: RsLabel): RsLabelDecl? =
+        label.branch
+            .mapNotNull {
+                when (it) {
+                    is RsLoopExpr -> it.labelDecl
+                    is RsForExpr -> it.labelDecl
+                    is RsWhileExpr -> it.labelDecl
```
alygin: Do you mean an interface or a base class?

matklad: An interface. If the name of interface methods match accessors generated by grammar kit, everything will work automatically.

alygin:
To summarize:

1. `lifetime` token type renamed to `quote_identifier`
2. Separate lifetimes and loop labels in grammar
3. Resolve `Lifetime` &#8594; `LifetimeDecl` and `Label` &#8594; `LabelDecl`
4. Tweak the renaming machinery a bit to support lifetimes and labels

There's a small problem with renaming: `RsNamesValidator` knows nothing about the type of the element being renamed, so the only way to make it work for both `identifier` and `quote_identifier` is to always allow both types. I think we need `RenamePsiElementProcessor` to make renaming more robust, I'd like to implement it as a separate task.

alygin:
BTW, how does `RustParser.bnf` know about the `identifier` and other token types? I've added the `quote_identifier` token type to the lexer, everything compiles and works, but IDEA highlights all the occurrences of `quote_identifier` as a "reference to a missing rule". How do I make it aware about this new type?

matklad:
>BTW, how does RustParser.bnf know about the identifier and other token types? I've added the quote_identifier token type to the lexer, everything compiles and works, but IDEA highlights all the occurrences of quote_identifier as a "reference to a missing rule". How do I make it aware about this new type?

All token types are generated by the grammar kit. This somewhat confusingly means that the generated lexer code depends on the generated parser code. 

GK infers the set of tokens from the actual rules (if you use `foo`, it will generate the `foo` token) and additional tokens can be listed in `tokens` attribute: https://github.com/intellij-rust/intellij-rust/blob/83efd81f645819cada6a85b92a78e106f6de874c/src/main/grammars/RustParser.bnf#L23.

I think you need to list tokens with `_` in them explicitelly: https://github.com/intellij-rust/intellij-rust/blob/83efd81f645819cada6a85b92a78e106f6de874c/src/main/grammars/RustParser.bnf#L71

alygin:
Hm, but there's no `identifier` token, there's no `loop` token and lots of other tokens are not presented in the `tokens` collection, yet they are still resolved as rules. But my new `quote_identifier` rule is not resolved which makes me think that I missed something.

matklad:
>Hm, but there's no identifier token, there's no loop token and lots of other tokens are not presented in the tokens collection, yet they are still resolved as rules. But my new quote_identifier rule is not resolved which makes me think I missed something.

`loop` does not have `_` in it, while `quote_identifier` does and thus needs to be listed explicitelly. I may be mistaken here, but this is how I remember GK works.

alygin:
I found out that it also works without adding to `tokens` if I use uppercase (`QUOTE_IDENTIFIER`).

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
```diff
@@ -212,6 +209,26 @@ object ResolveEngine {
...
+            .mapNotNull { it as? RsLabeledExpression }
+            .map { it.labelDecl }
+            .find { it.name ==  label.quoteIdentifier.text}
+
+    fun resolveLifetime(lifetimeRef: RsLifetime): RsLifetimeDecl? {
+        if (lifetimeRef.isPredefined) return null
+        return lifetimeRef.ancestors
+            .mapNotNull {
+                when (it) {
+                    is RsTypeAlias -> if (it.parent is RsImplItem || it.parent is RsTraitItem) null else it.typeParameterList
```
matklad: Why do we need `if (it.parent is RsImplItem || it.parent is RsTraitItem)` condition? 

We can just resolve this, and annotate lifetimes in associated types as forbidden (or just wait until they implement that HKT RFC :) )

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -4,15 +4,23 @@ import com.intellij.openapi.project.Project
...
+import org.rust.lang.core.psi.RsElementTypes.QUOTE_IDENTIFIER
 import org.rust.lang.core.psi.impl.mixin.selfParameter
 import org.rust.lang.core.psi.impl.mixin.valueParameters
 import org.rust.lang.core.psi.util.childOfType
 
 class RsPsiFactory(private val project: Project) {
-    fun createIdentifier(text: String): PsiElement =
-        createFromText<RsModDeclItem>("mod $text;")?.identifier
-            ?: error("Failed to create identifier: `$text`")
+    fun createIdentifier(text: String, tokenType: IElementType): PsiElement =
```
matklad: I think it's better to crate two functions: `createIdentifier` and `createQuoteIdentifier`. We will have to overload `setName` in label and lifetime reference, but it is probably a good thing to do anyway.

src/main/kotlin/org/rust/lang/refactoring/RsNamesValidator.kt
```diff
@@ -18,10 +19,10 @@ class RsNamesValidator : NamesValidator {
         isIdentifier(name, project, true)
 
     fun isIdentifier(name: String, @Suppress("UNUSED_PARAMETER") project: Project?, withPrimitives: Boolean): Boolean =
-        if (withPrimitives) {
-            getLexerType(name) == IDENTIFIER && name !in PrimitiveTypes
-        } else {
-            getLexerType(name) == IDENTIFIER
+        when (getLexerType(name)) {
+            IDENTIFIER -> !withPrimitives || name !in PrimitiveTypes
+            QUOTE_IDENTIFIER -> name !in PredefinedLifetimes
```
matklad: Perhaps we should omit this check? Renaming lifetime to `'static` seems a useful feature. We can annotate `'static` lifetime declarations separately. 

A similar thing happens if you rename `z` to `x` in this Kotlin code `fun foo(x: Int, z: Int) {}`: idea does the rename, but warns you about the conflict.

src/main/kotlin/org/rust/lang/core/resolve/ResolveEngine.kt
alygin: > and annotate lifetimes in associated types as forbidden

We already do that.

I just thought we could annotate an undeclared lifetime earlier, without waiting for the user to remove an invalid lifetime declaration from the type alias in cases like this:

```rust
trait Foo {
    type Bar<'a> = &'a str;
}
```

But ok, it's a very minor improvement and maybe doesn't worth carrying this logic.

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
alygin: I just tried to unify identified and quote_identified elements by providing a token type to the `RsNamedElementImpl` constructor. But that didn't work well when it came to renaming support, so I agree, maybe it's better to separate them at all levels.

src/main/kotlin/org/rust/lang/refactoring/RsNamesValidator.kt
alygin: Yeah, renaming to `'static` should be allowed. And its declaration could be removed after renaming to `'static` with help of `RenamePsiElementProcessor`, that's a good idea.

matklad: >And its declaration could be removed after renaming to 'static with help of RenamePsiElementProcessor

That's a bit to magical I think. Let's stick with just renaming to `'static` for now, and then add conflicts dialog like the one in Kotlin?

alygin: Though, labels cannot be named `'static` )

alygin: > Let's stick with just renaming

Sure, I didn't mean implementing it in this PR.

alygin:
@matklad, thaks for the example on [rust-lang/rust](https://gitter.im/rust-lang/rust). Who else would remember `for<'a>`? :)
I've added this case.

matklad:
Splendid!

Issue #859, opened 2016-12-27
Implement a list selector


matklad:
For `fn foo(a: i32, b: String)` it would be useful if semantic selection (`Ctrl+W`) could select all the arguments, but without parenthesis (how it is done in Kotlin):

```
fn foo(<selection>a: i32, b: String</selection>)
```

This can be done via `ExtendWordSelectionHandler`.

Here's the relevant code and tests in Kotlin:

https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinListSelectioner.kt

https://github.com/JetBrains/kotlin/blob/master/idea/tests/org/jetbrains/kotlin/idea/WordSelectionTest.java

zjhmale:
@matklad You mean `Extend Selection` and `Shrink Selection` in keymap right?

matklad:
Yes

mrhota:
@matklad looks like that PR implements this, so you can close this issue

matklad:
Thanks @mrhota !

PR #998, opened 2017-02-12
SEL: Select args and params lists without parentheses

alygin:
Implements #859


src/main/kotlin/org/rust/ide/wordSelection/RsListSelectionHandler.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.rust.lang.core.psi.*
+
+class RsListSelectionHandler : ExtendWordSelectionHandlerBase() {
+    override fun canSelect(e: PsiElement): Boolean =
+        e is RsTypeArgumentList || e is RsValueArgumentList || e is RsTypeParameterList || e is RsValueParameterList
+
+    override fun select(e: PsiElement, editorText: CharSequence, cursorOffset: Int, editor: Editor): List<TextRange>? {
+        val node = e.node!!
+        val startNode = node.findChildByType(TokenSet.create(RsElementTypes.LPAREN,RsElementTypes.LT)) ?: return null
+        val endNode = node.findChildByType(TokenSet.create(RsElementTypes.RPAREN,RsElementTypes.GT)) ?: return null
```
matklad: Space is missing after `,` :) Also, let's use unqualified form for tokens: `tokenSetOf(LPAREN, LT)`.  (we do have `tokenSetOf somewhere`).

matklad:
:+1:

Issue #489, opened 2016-06-24
Implement "chain comparison" annotator.


matklad:
Problem: some users (me included) often times forget to place `::` before generics in expression context. That is, they write `Vec::new<i32>()` instead of `Vec::new::<i32>()`

Solution: this problem is neatly solved in Rust. Chained comparisons like `1 < 2 < 3` are forbidden on the syntactic level (in contrast to other languages, where they are ruled out during type checking), so the parser can provide a meaningful "did you mean `::<>`" error. For the plugin, I thing the best way to handle this is to add the case to the [InvalidSyntaxAnnotator](https://github.com/intellij-rust/intellij-rust/blob/cdf2cb4830c0cbb4df2ac4c406f267c9bd23c86b/src/main/kotlin/org/rust/ide/annotator/RustInvalidSyntaxAnnotator.kt).

The plan:
- Make sure that chained comparisons are indeed always forbidden in Rust and collect a bunch of test cases. 
- Add a new test and new pattern to the annotator. 
- (optional, separate PR) add a quick fix to turn `foo<T>()` into `foo::<T>()`.

mkaput:
Some real-life test case (should be annotated as error, but is not):
```rust
fn main() {
    let lit = b
    <caret>"Hello, World";
}
```

Generated PSI (as of 2016-08-07):
```
FILE(0,57)
  RustFnItemElementImpl(FN_ITEM)(0,56)
    PsiElement(fn)('fn')(0,2)
    PsiWhiteSpace(' ')(2,3)
    PsiElement(<IDENTIFIER>)('main')(3,7)
    RustParametersElementImpl(PARAMETERS)(7,9)
      PsiElement(()('(')(7,8)
      PsiElement())(')')(8,9)
    PsiWhiteSpace(' ')(9,10)
    RustBlockElementImpl(BLOCK)(10,56)
      PsiElement({)('{')(10,11)
      PsiWhiteSpace('\n    ')(11,16)
      RustLetDeclElementImpl(LET_DECL)(16,54)
        PsiElement(let)('let')(16,19)
        PsiWhiteSpace(' ')(19,20)
        RustPatIdentElementImpl(PAT_IDENT)(20,23)
          RustPatBindingElementImpl(PAT_BINDING)(20,23)
            PsiElement(<IDENTIFIER>)('lit')(20,23)
        PsiWhiteSpace(' ')(23,24)
        PsiElement(=)('=')(24,25)
[...]

PR #999, opened 2017-02-12
INT: Add namespace intention #489 for trivial cases

la10736:
I implemented the intention just in the trivial but frequent case where you have just one generic parameter. I wrote a test case for the general case but comment it out. I don't know how can I try to solve it... any hints will be very appreciated. Maybe it is better spread general case in another PR.


matklad:
Nice! I think that instead of the intention, this should be a quick fix associated with the error annotation.

You can add a quick fix like this:

```Kotlin
    private fun checkBinary(holder: AnnotationHolder, o: RsBinaryExpr) {
        if (o.isComparisonBinaryExpr() && (o.left.isComparisonBinaryExpr() || o.right.isComparisonBinaryExpr())) {
            holder.createErrorAnnotation(o, "Chained comparison operator require parentheses")
                .registerFix(AddTurbofishOperatorFix())
        }
    }
```

matklad:
Oups, didn't mean to close this :sweat_smile:

src/main/kotlin/org/rust/ide/intentions/AddNamespaceIntention.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.util.parentOfType
+import org.rust.lang.core.psi.RsElementTypes.*
+
+class AddNamespaceIntention : RsElementBaseIntentionAction<AddNamespaceIntention.Context>() {
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        val (matchExpr, caller, gen, more) = ctx
+        val callWithNamespace =
+            RsPsiFactory(project).createExpression("""${caller.text}::<${gen.text}>${more.text}""") as RsCallExpr
```
matklad: Single quotes should be enough here. Also you don't need this cast

src/main/kotlin/org/rust/ide/intentions/AddNamespaceIntention.kt
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    override fun getText() = "Add missed namespace"
+    override fun getFamilyName() = text
+
+    private fun resolveMatchExpression(element: PsiElement): RsBinaryExpr? {
+        val base = element.parentOfType<RsBinaryExpr>() ?: return null
+        return if (base.left is RsBinaryExpr) {
+                base
+            } else {
```
matklad: Indentation is a bit messed up. Reformat file (`Ctrl+Alt+l`) should cleanup this.

matklad:
> I don't know how can I try to solve it... any hints will be very appreciated.

That's an interesting problem indeed!  When there are several type arguments, you don't get any reasonable PSI tree because the parser expects expression after the `<` and `T, U` does not look like an expression at all:

train_wreck

The good news is that we still trigger "chain comparison" annotator in this cases sometimes. For example, for 

```
.collect<Result<Vec<_>, _>>();
```

We will parse `< Result <` part. I think we can handle this, if we base our modifications not on the `PSI`, but on the text instead. That is, 

```
editor.document.insertString(lt.textOffset - 1, "::")
```

should work in more cases. 


This still won't handle the case like `.collect<T, U>();`, but the problem here is that even the original error annotation does not work here, because we fail to reconstruct any reasonable PSI. I suggest we don't handle this case in this PR.

matklad:
>but the problem here is that even the original error annotation does not work here, because we fail to reconstruct any reasonable PSI. 

How should we fix it? I don't know a good solution unfortunately. The nice trick would be to check all `a < b` expressions, where the node immediately after `b` is an error element. Then, we can try to copy the file, insert `::` before `<` and see if that fixes syntax errors. If it does, we can suggest a fix.

la10736:
@matklad 
> You can add a quick fix like this:

I didn't know as do it... Is it possible also filter the cases where register the quick fix? I'm not sure if it is a good idea try to fix all chained comparison.

matklad:
>I didn't know as do it... Is it possible also filter the cases where register the quick fix?

Sure! The code should like this:

```
val annotation = holder.createAnnotation()

if (looksLikeTheUserForgotTheTurbofish) {
    annotation.registerFix(RsAddTurbofishOperatorFix())
}
```

la10736:
I don't think we should relay on the *lucky* case where chained comparison is triggered. Is better try to write something that is quite closer this issue like a  `,` in a `PsiErrorElement` and find a `RsBinaryExpression` thats contain it in the right branch. I think that can be more robust.

I'll try to handle this special cases in an other PR. I'll take in account your considerations and fix this simple case by use quickfix instead annotation (maybe is better a new PR too to take commits clean).

matklad:
>I don't think we should relay on the lucky case where chained comparison is triggered.

My though was that we can fix the "chained comparison" to trigger in more cases. If we can fix something, we must be aware that the problem exists anyway :) 

>maybe is better a new PR too to take commits clean

That sounds reasonable! Or you can just force-push to this branch.

la10736:
I'm working (slowly) on handling some more cases. I'll move these changes on another PR.

Issue #1000, opened 2017-02-13
Simplify boolean expression inspection should not modify PSI


matklad:
See https://github.com/intellij-rust/intellij-rust/commit/e0962ec008e9ed9f8031ecd3d12ca57be1b20fbb

The problem is that inspection tries to change PSI. I've disabled the inspection by default for now. 

This can be hacked around by coping the PSI node, like we do in the intention, but I wonder if this can be done more efficiently, without coping? What if we parametrize the simplification function with the return value, such that it returns `Unit` in query mode and `RsExpr` in execute mode? 

cc @slavam2605

matklad:
:100:0

slavam2605:
@matklad Should it be closed?

matklad:
Indeed!

PR #1001, opened 2017-02-13
INSP: Fix of simplify boolean expression inspection not to change PSI tree

slavam2605:
Fix of [#1000](https://github.com/intellij-rust/intellij-rust/issues/1000).
Happy anniversary 1000-th issue :)

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsSimplifyBooleanExpressionInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSimplifyBooleanExpressionInspection.kt
@@ -20,7 +20,7 @@ class RsSimplifyBooleanExpressionInspection : RsLocalInspectionTool() {
         override fun visitExpr(o: RsExpr) {
             if (o.isPure() != true)
                 return
-            val (expr, result) = o.simplifyBooleanExpression()
+            val (ignored, result) = o.simplifyBooleanExpression(true)
             if (!result)
                 return
             holder.registerProblem(
--- a/src/main/kotlin/org/rust/ide/inspections/RsSimplifyBooleanExpressionInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSimplifyBooleanExpressionInspection.kt
@@ -4,6 +4,7 @@ import com.intellij.codeInspection.LocalQuickFix
 import com.intellij.codeInspection.ProblemDescriptor
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
+import org.rust.ide.utils.canBeSimplified
 import org.rust.ide.utils.isPure
 import org.rust.ide.utils.simplifyBooleanExpression
 import org.rust.lang.core.psi.RsExpr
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/inspections/RsSimplifyBooleanExpressionInspection.kt
```diff
@@ -20,7 +20,7 @@ class RsSimplifyBooleanExpressionInspection : RsLocalInspectionTool() {
         override fun visitExpr(o: RsExpr) {
             if (o.isPure() != true)
                 return
-            val (expr, result) = o.simplifyBooleanExpression()
+            val (ignored, result) = o.simplifyBooleanExpression(true)
```
slavam2605: It's a pity that Kotlin have no `_` in destructuring patterns for now :(

slavam2605: At least my compiler haven't. Kotling plugin suggest to rename to `_`, but this does not compile for me.

matklad: Yep, `_` will be available with Kotlin 1.1.

src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
```diff
@@ -103,37 +103,46 @@ val RsUnaryExpr.operatorType: UnaryOperator?
...
  * Simplifies a boolean expression if can.
  *
+ * @param peek if true then does not perform any changes on PSI, 
+ *             `expr` is not defined and `result` indicates if this expression
+ *             can be simplified
  * @return `(expr, result)` where `expr` is a resulting expression,
  *         `result` is true if an expression was actually simplified.
  */
-fun RsExpr.simplifyBooleanExpression(): Pair<RsExpr, Boolean> {
+fun RsExpr.simplifyBooleanExpression(peek: Boolean = false): Pair<RsExpr, Boolean> {
```
matklad: Cosmetics: function with a boolean argument whose value is a literal is usually a code smell. It's better to split this function in two with different names to enhance readability. 

In this case, I think we can do roughly like this:


```Kotlin
fn RsExpr.canBeSimplified(): Boolean {
    return simplifyBooleanExpression(dryRun=true).second
}

fn RsExpr.simplifyBooleanExpression(): RsExpr {
    return simplifyBooleanExpression(dryRun=false).first
}

// note that we hide ugly boolean argument API inside a private function
private fn RsExpr.simplifyBooleanExpression(dryRun: Boolean): Pair<RsExpr, Bool> {
    ....
}


```

slavam2605: Fixed.
`simplifyBooleanExpression` still returns a pair to prevent twice traversal of PSI tree.

matklad:
Thanks!

src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
matklad: >simplifyBooleanExpression still returns a pair to prevent twice traversal of PSI tree.

This should not matter much: the second traversal is triggered only when you actually invoke the fix, and it's OK to spend more time than, because fixes are applied much rarely then the inspections are run.

PR #1012, opened 2017-02-15
ANNOT: Undeclared lifetimes and labels annotation E0261/E0426

alygin:
Annotate undeclared lifetimes and labels (E0261 and E0426 from #886)


matklad:
Super!

Issue #997, opened 2017-02-12
Support inferring types for operations


sirgl:
Current implementation of type inferrence support inferrence only for boolean operations.
For example:
```
let a = 2;
let b = 3;
let c = a + b; // Plugin doesn't know what is the type of c
```
In Rust operations supported by implementations of appropirate traits : `Add`, `Mul`...
That's mean, that `RustTypificationEngine` will require to have access to traits, implemented for current type.
But method `getTraitsImplementedIn()` require project, that is not accessible in `RustTypificationEngine`.
Is it ok to bring to `typifyExpr()` project?
Or it is required to precompute it somehow?

As I can see, type inferrence is the most important feature now.

matklad:
>that is not accessible in RustTypificationEngine

Any `PsiElement` has a `getProject` method.

sirgl:
Sorry for asking stupid questions again, try to implement it :)

matklad:
Was fixed in https://github.com/intellij-rust/intellij-rust/pull/1267, thanks @sirgl  and @Undin !

PR #1016, opened 2017-02-16
Inference for simple operations

sirgl:
Add inference for Add, Sub, Rem, Mul and Div traits and fallbacks for primitive types. Planning to add unary operations also. Am I going right direction? #997


matklad:
Yes, this looks reasonable. 

In general, we need a better infrastructure for finding known items, like operator traits, but matching solely by name is Ok for now.

sirgl:
This is not very precise implementation as it not using subtyping for automatic type conversion. As I could understand, we have pretty limited type inference for now.
Mostly, notes for myself:
Firstly, we need some function that will check if one type is convertible to another. I started to do it according to [coercions](https://doc.rust-lang.org/nomicon/coercions.html), and [deref coencions](https://doc.rust-lang.org/book/deref-coercions.html), but some rules requires raw pointers to implement it that we don't have for now.
Secondly, we need to add recognition of lvalue, rvalue context required by `Deref`, `DerefMut`, `Index` and `IndexMut`
Also, add slice [T] as type

src/test/kotlin/org/rust/lang/core/type/RsExpressionTypeInferenceTest.kt
```diff
@@ -354,5 +357,75 @@ class RsExpressionTypeInferenceTest : RsTypificationTestBase() {
...
+            "Mul" to "*",
+            "Shl" to "<<",
+            "Shr" to ">>",
+            "BitOr" to "|",
+            "BitXor" to "^",
+            "BitAnd" to "&"
+        )
+        traitNames.forEach { entry ->
+            val traitName = entry.key
+            val operation = entry.value
```
matklad: This can be simplified to `for ((traitName, operation) in traitNames) { ... }`

src/test/kotlin/org/rust/lang/core/type/RsExpressionTypeInferenceTest.kt
```diff
@@ -1,6 +1,9 @@
 package org.rust.lang.core.type
 
 import org.junit.ComparisonFailure
+import org.junit.Test
+import org.rust.lang.core.types.types.RustFloatType
+import org.rust.lang.core.types.types.RustIntegerType
 
 class RsExpressionTypeInferenceTest : RsTypificationTestBase() {
```
matklad: Let's also add an integration tests, that will check that this works with real `::std::ops`. 

I think a test in `org.rust.lang.core.resolve.RsStdlibResolveTest` like this one

```Kotlin
    fun `test resolve operators`() = stubOnlyResolve("""
    //- main.rs
        fn main() {
            let x = String::new();
            let y = x + "World!";
            y.as_bytes()
        }       //^ ...libcollections/string.rs
    """)
```

should work. 

I think it fails currently for two reasons:

* it switches from stubs to AST, which is probably not that important
* the traits in `::std::ops` are generic, like `impl Add<&str> for String`, and this breaks  `.filter { it.traitRef?.text == traitName }`

sirgl:
To finish this work I need support for generics in stubs indexing and subtyping support from another my request, that makes possible using subtype instead of original in ops. But I think that PR will become huge...

sirgl:
But these remarks I'll fix

matklad:
>To finish this work I need support for generics in stubs indexing and subtyping support from another my request,

I think the test I've mentioned will work if instead of `traitReference.text` you will use `traitRefernece.path.referenceName`?

sirgl:
Ah, ok. But I meant that not only exact class match can be appliable, but also traits that can be casted by deref and [T,N] -> [T] and so on.

matklad:
>Ah, ok. But I meant that not only exact class match can be appliable, but also traits that can be casted by deref and [T,N] -> [T] and so on.

Yep, sure, there's a lot ground yet to cover! But it's always nice to be able to handle some common cases

Shiroy:
@sirgl don't you think it would be more reliable to think about a way to implement the Hindley-Milner type inference algorithm in the plugin ? I think it would give better results and could also allow support of generic.

I've started some work in that way a couple of weeks ago but I first have to understand how the HM algorithm work and which constraint, specific to IDEs, may influence the results. I'd be very pleased to work with you on this problem.

sirgl:
@Shiroy To tell the truth I didn't know about this algorithm before =) I'm a beginner in compiler and IDE development, so I do it just for reasons of common sense. But sure, I will look at this algo and try to help you with its implementation if it will be considered resonable.

matklad:
@Shiroy yeah, a full H-M is what we want ideally here. However, that's not that simple. Rust uses a specific flavor of H-M, mainly because of the complicated method lookup (you need to know the type to do trait matching, so method selection should be run after unification, but you need the return and argument types of methods for unification, so this becomes somewhat entangled), and it's note really specified anywhere except the code of the compiler, so matching compiler's behavior exactly will be hard. 

I don't thing there will be much IDE specificity in the unification itself: it's local, because you need to look only at a single function's body, so you can do it eagerly, without typical IDE laziness. 


But this stuff is complicated. I think at this stage a good strategy is to use simplistic type inference to get the major benefits for completion and resolve. When it becomes clear if we should or should not use the RLS, we can either strip our own inference engine, or make it to replicate what rustc does precisely.

Shiroy:
Ok I see your point. It is the same kind of algorithms that are used in the Kotlin plugin/compiler ? I'd be very interested to have a look

matklad:
> It is the same kind of algorithms that are used in the Kotlin plugin/compiler ?

I am not really familiar with Kotlin implementation, but yes, there's some unification style inference it Kotlin. It is both simpler and more complex though: simpler, because there are no impls, and all extensions should be imported. More complex because of function overloading.

src/main/kotlin/org/rust/lang/core/types/OperationsSupport.kt
```diff
@@ -0,0 +1,53 @@
+package org.rust.lang.core.types
+
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.resolve.indexes.RsImplIndex
+import org.rust.lang.core.types.types.*
+
+
+fun RsExpr.findImpl(traitName: String): RsImplItem? {
```
farodin91: It would be better to use the language attribute, see here #1177

farodin91: https://doc.rust-lang.org/nightly/src/core/ops.rs.html#243-254

PR #1017, opened 2017-02-16
ANNOT: Annotate type placeholders in item signatures

alygin:
Detects usage of type placeholders in function signatures and constants ([E0121](https://doc.rust-lang.org/error-index.html#E0121) from #886).


matklad:
Thanks!

PR #1019, opened 2017-02-17
DOC: Expand QuickNavagateInfo to all named elements

### LINKED ISSUE #1018: Suggestion: Display const value when inspecting it with ctrl+mouse pointer
Hey guys!  
  
I've got a suggestion for a feature that I often use in php. I only use PhpStorm but I guess that's something available in other Jetbrains IDE.  
As you may have guessed from the title the idea would be to display a const value when hovering it while ctrl is pressed. This looks like follow in php:  
   
image  
*I couldn't capture my mouse pointer but it was just on `TEST`* 
  
And here is the current behavior in Rust:  
image  
*Here my cursor was just on `LINE_ID`*  
  
I would have gladly added it myself but my Java is kinda rusty (see what I did there? :smile: ) and I've got no experience in Kotlin.  
Toughts?

alygin:
What the title says + quick nav info tests inlining

Fixes #1018


alygin:
Fixed `self` parameter handling. The overall implementation is not very elegant, I'm pressed for time now, but could get back to it a bit later.

matklad:
>The overall implementation is not very elegant, I'm pressed for time now, but could get back to it a bit later.

Yeah, I think at least some renames and unification between `PatBinding` and all other named elements would be useful. 

Do you want this to be merge and send refactoring in the follow up pull request?

alygin:
> Do you want this to be merge and send refactoring in the follow up pull request?

Yes, the changes look safe (at least to me) and bear tangible fruit, so I'd prefer to have them merged and get back to refactoring in a separate PR when time allows.

matklad:
Merged!

alygin:
Thanks!

PR #1020, opened 2017-02-17
DOC: Better quick navigation info

alygin:
Refactoring and various improvements of quick navigation info.

### DIFF
--- a/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
@@ -1,13 +1,12 @@
 package org.rust.ide.docs
 
 import com.intellij.lang.documentation.AbstractDocumentationProvider
-import com.intellij.openapi.util.text.StringUtil
-import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
-import com.intellij.psi.PsiWhiteSpace
-import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.impl.RsFile
-import org.rust.lang.core.psi.util.ancestors
+import org.rust.ide.utils.presentationInfo
+import org.rust.lang.core.psi.RsDocAndAttributeOwner
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsMod
+import org.rust.lang.core.psi.RsNamedElement
 import org.rust.lang.doc.documentationAsHtml
 
 class RsDocumentationProvider : AbstractDocumentationProvider() {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -17,89 +17,159 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
...
+            is RsStructItem -> element.signatureText
+            is RsFieldDecl -> element.signatureText
+            is RsEnumItem -> element.signatureText
+            is RsEnumVariant -> element.signatureText
+            is RsTraitItem -> element.signatureText
+            is RsTypeAlias -> element.signatureText
+            is RsConstant -> element.signatureText
+            is RsSelfParameter -> element.signatureText
+            is RsTypeParameter -> element.signatureText
+            is RsLifetimeDecl -> element.signatureText
```
matklad: The ton of short `signatureText` functions seems a bit repetitive. Perhaps it's better to inline them into this `when`? 

Also, some of `signatureText` are `fun`s and some are `val`s

alygin: > Perhaps it's better to inline them into this `when`

Absolutely

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -17,89 +17,101 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
...
     }
 
-    private fun RsNamedElement.getQuickNavigateInfo(): String =
-        "${formatSignature()}$locationString"
+    private val RsFunction.signatureText: String?
+        get() = createSignatureText(this, identifier, listOf(whereClause, retType, valueParameterList))
 
+    private val RsPatBinding.owner: PsiElement?
+        get() = ancestors
+            .drop(1).dropWhile { it is RsPat || it is RsMatchPat || it is RsPatField }
```
alygin: @matklad, why `RsMatchPat` and `RsPatField` aren't `RsPat`? Is there some idea behind it? Or they were simply neglected?

matklad: Yes, that's intentional. Although the name may be misleading. `RsMatchPat` is not a pattern because you can't stick it anywhere they pattern can go:

```
let (None | Some(x)) = None; // syntax error
let (None if cond) = None; // ditto
```

The same goes for `RsPatField`.

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -17,89 +17,101 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
...
+            .filterIsInstance<RsCompositeElement>()
+            .firstOrNull()
```
matklad: new line is missing :)

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
matklad: Hm, this reminds me of [this code](https://github.com/intellij-rust/intellij-rust/blob/a4e6f0b4ddeee50217500832ca416d37f7181fa1/src/main/kotlin/org/rust/lang/core/types/PatternMatching.kt#L13).

I think this should be extracted to

```
val RsPatBinding.topLevelPattern: RsPat // no ?
```

Then an owner is just `topLevelPattern.parent` I think.

Note that `topLevelPattern` should walk through `PatField`, but stop at `MatchPat`.

matklad:
We also need to add a test for https://gitter.im/intellij-rust/intellij-rust?at=58adef1c00c00c3d4f743494

matklad:
Have written a thing for discovering this sort of bugs earlier: https://github.com/intellij-rust/intellij-rust/commit/39706a8b062521e77bbc5bcf7952f82d43dcc66e

alygin:
Thanks! I'll then make final touches here without waiting for #1023 to finish.

alygin:
> Have written a thing for discovering this sort of bugs earlier

I don't quite understand how it catches bugs like the one mentioned. It seems that it checks parsing. Now, it catches a syntax error in the following line that is correct from the `rustc`'s point of view:

https://github.com/rust-lang/rust/blob/5509ae399e4ed6f041b95bf948751f35b7d65517/src/test/run-pass/union/union-backcomp.rs#L17

But it catches nothing when I put `check(false)` [here](https://github.com/intellij-rust/intellij-rust/blob/39706a8b062521e77bbc5bcf7952f82d43dcc66e/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt#L83).

alygin:
So, I've extracted `createSignatureText()` and generalized it to represent a named element in various contexts. It's now `RsNamedElement.presentationInfo` of type `PresentationInfo` that allows to get various presentation of the element.

There's the `ItemPresentation` interface that seems similar, but I don't know if it's possible and reasonable to merge them somehow.

This class can be used in numerous other places, I'll do the refactoring in a separate PR (`typeNameText` and `projectStructureItemText` are exactly for that).

@slavam2605, you can now rebase and use `member.presentationInfo.projectStructureItemText` [here](https://github.com/intellij-rust/intellij-rust/pull/1023/files#diff-d3d55455ab063cc1a2876ba565322b64R21).

alygin:
In one of the previous commits I also fixed those range check failures. So, it fixes #1032.

slavam2605:
Looks great)
I will rebase when I will complete tests

matklad:
Hm, 

```
    fun `test anything`() = forAllPsiElements { element ->
        if (element is RsNamedElement) {
            try {
                RsDocumentationProvider().getQuickNavigateInfo(element, element)
            } catch(e: Exception) {
                println(element)
                throw e
            }
        }
    }
```

throws

alygin:
> Hm, ... throws

This test found one real problem, I've fixed it. But it keeps falling with `FileNotFoundException` in [this line](https://github.com/intellij-rust/intellij-rust/commit/39706a8b062521e77bbc5bcf7952f82d43dcc66e#diff-b4dea1972a2f0a54cb7a7ba528bf466aR122). Why would that be? My rust sources are up to date for all channels.

matklad:
>Why would that be? My rust sources are up to date for all channels.

VFS is not always refreshed during tests, and can be dissinchronized with the real world. To clean it up, you can use something like

```
rm -rf ~/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/2016.3.2/78432319336fa253b797390a1ad66e0c5b69fee1/ideaIC-2016.3.2/system/{caches,index,LocalHistory}
```

That is, when you run tests via gradle, sandbox directory is `intellij-rust/build/sandbox`. When  you run them via IDEA runner, it uses IDEA from gradle as a sandbox.

alygin:
Thanks! The test is green now.

matklad:
Great!

matklad:
Will publish a new alpha now!

Issue #1010, opened 2017-02-15
Detect missing associated types in trait impls


alygin:
There's the "Missing trait item" annotation that detects which functions are missing in a trait implementation. It would be nice to also detect missing associated types:

```rust
struct Foo;
trait Bar {
    type SomeType;
}
impl Bar for Foo {  // <-- should complain about missing `SomeType`
}
```

Even better would be to add the missing associated types in the fix, provided by the existing annotation.

alygin:
closed in https://github.com/intellij-rust/intellij-rust/commit/75a77e2b5027be8f7cb54aa4c90379c1943cc35c

Issue #1009, opened 2017-02-15
Missing trait items false positive


alygin:
[Specialization](https://github.com/rust-lang/rfcs/pull/1210/) makes the current implementation of "Missing trait items" annotation prone to false positives:

```rust
#[derive(Clone)]
struct Foo<T> { t: T }

trait Bar<T> {
    fn bar(t: T);
}

impl<T> Bar<T> for Foo<T> {
    default fn bar(t: T) {}
}

impl<T: Clone> Bar<T> for Foo<T> {}
//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  Not all trait items implemented, missing `bar`
```

Just noticed it, didn't think of what would be the better way to deal with it yet.

alygin:
I think the current state of the resolve engine doesn't allow us to implement an absolutely correct specialization analysis and tell for sure which trait items are implemented for a struct and which are missing. Please, correct me if I'm wrong.

So, I would go with a simplified solution that is prone to false negatives, but removes the mentioned false positives. Particularly, I have two solutions in mind:

1. Don't annotate `impl` blocks that have type parameters polybounds (like `impl<T: Clone> ...` or `impl<T> ... where T: Send` etc) since they are potential specializations. This solution is simple and lightweight but switches off the annotation for many implementations that are not real specializations.

2. Extend the solution 1 by finding all the implementations of the given trait for the given struct and check if there're `default` implementations of the missing items from the `impl` block in question. Remove what we found from the list of missing items without additional checks. But I'm actually not sure that such solution won't bring some new kind of false positives, and it's pretty heavy (though we can do the analysis only when there're really missing items and the `impl` block is a specialization candidate as described above).

@matklad, any thoughts on this?

matklad:
2 seems to be a reasonable approach. 

>we can do the analysis only when there're really missing items

I think this will cover absolute majority of cases, so I wouldn't be afraid about performance. And we need to make impl search fast anyway :)

matklad:
But another option is just to wait until someone reports a false positive bug report: I think that perhaps specialization is not that abundant in practice? Though I don't have any real numbers.

alygin:
[...]

PR #1023, opened 2017-02-17
INT?: Implement trait methods and type aliases

slavam2605:
This PR extends `Implement members` functionality.
1. Added automatic implementation of type aliases (with defaulting type `()`)
2. `Implement members` quick fix now shows a chooser dialog to choose members to implement
3. `Ctrl+I` (implement members code insight feature) is now available

Example of chooser dialog:
image

Example (when all members were chosen):
```rust
#![feature(associated_type_defaults)]

trait Foo {
    type Type1 = i32;
    type Type2;
    type Type3;
    fn foo();
    fn bar();
    fn baz() {}
}

struct Bar;

impl Foo for Bar {
    type Type2 = ();
    type Type3 = ();

    fn foo() {
        unimplemented!()
    }

    fn bar() {
        unimplemented!()
    }
}
```

**Note**: I commented out one test, because now quick fix does not automatically generate all members, only after dialog choosing. I don't know how to write a new version of test.


alygin:
Cool! I guess it implements the second part of #1010.

slavam2605:
Actually I've updated an error annotator too.

src/main/kotlin/org/rust/utils/formatterUtils.kt
```diff
@@ -0,0 +1,63 @@
+package org.rust.utils
+
+import com.intellij.psi.PsiComment
+import com.intellij.psi.PsiWhiteSpace
+import org.rust.lang.core.psi.*
+
+fun RsNamedElement.getFormattedParts(): Pair<String, String> {
```
alygin: This part will cause conflict with #1020. There's now better implementation of signatures of various elements. And we also do similar work in [tree elements](https://github.com/intellij-rust/intellij-rust/tree/9f545171d9c40b5e223e3c85a4d6bcdcb2aac3d0/src/main/kotlin/org/rust/ide/structure) to show them in the project structure. So I'm totally for moving this functionality into a separate class and reuse it where needed.

So, I see three steps to achieve this:
1. Merge this PR.
2. I'll rebase my #1010 branch and replace this implementation of `getFormattedParts()` with the improved version.
3. I'll do the refactoring in tree structure to reuse this class and to move some ideas from there to `getFormattedParts()`. For instance, we effectively remove internal comments in tree structure, but leave them here so far, it should be fixed.

@matklad

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,112 @@
+package org.rust.ide.core.overrideImplement
```
alygin: The name of a `.kt` file should start with an uppercase letter

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,112 @@
...
+                text = "${member.type.name}"
+            }
+        }
+    }
+
+    override fun renderTreeNode(component: SimpleColoredComponent?, tree: JTree?) {
+        when (member) {
+            is FunctionOrType.Function -> {
+                component?.append("fn ", SimpleTextAttributes.REGULAR_BOLD_ATTRIBUTES)
+                component?.icon = RsIcons.FUNCTION
```
alygin: `member.function.getIcon(0)` will provide a tailored icon instead of the general one. The same with types.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
slavam2605: I was comparing with Kotlin plugin where all kotlin files (not classes) start with a lowercase letter.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,112 @@
...
+    val toImplementTypes = trait.typeAliasList.filter { it.name in notImplementedTypes }
+
+    if (toImplement.isEmpty() && toImplementTypes.isEmpty())
+        return
+    val base = MemberChooserObjectBase(traitName, RsIcons.TRAIT)
+    val members = toImplement.map { RsFunctionChooserMember(base, FunctionOrType.Function(it)) } +
+            toImplementTypes.map { RsFunctionChooserMember(base, FunctionOrType.Type(it)) }
+
+    val chooser = MemberChooser(members.toTypedArray(), true, true, project)
+    chooser.title = "Implement members"
```
alygin: It should be in Title Case: Implement **M**embers

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
alygin: Hm, I never noticed that. So, maybe I'm wrong.

slavam2605: [one](https://github.com/JetBrains/kotlin/tree/master/idea/src/org/jetbrains/kotlin/idea/util), [two](https://github.com/JetBrains/kotlin/tree/master/idea/idea-core/src/org/jetbrains/kotlin/idea/core), [three](https://github.com/JetBrains/kotlin/tree/master/idea/ide-common/src/org/jetbrains/kotlin/util)
Looks like a good practice to distinguish Kotlin files from Kotlin classes.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
alygin: Let's also set `chooser.copyJavaDocVisible = false` to remove that checkbox at the bottom.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    override fun renderTreeNode(component: SimpleColoredComponent?, tree: JTree?) {
+        when (member) {
+            is FunctionOrType.Function -> {
+                component?.append("fn ", SimpleTextAttributes.REGULAR_BOLD_ATTRIBUTES)
+                component?.icon = RsIcons.FUNCTION
+            }
+            is FunctionOrType.Type -> {
+                component?.append("type ", SimpleTextAttributes.REGULAR_BOLD_ATTRIBUTES)
```
alygin: I see that Kotlin doesn't add `fun` to functions in the same window (neither do Java or CLion, but they don't have such keywords). Maybe we don't need `fn` and `type` labels here because can simply understand the type of the item by the icon and signature. What do you think?

slavam2605: I don't know, I thought that it was a good idea to mark them but now I don't know what is better.

alygin: My arguments are:

- This window shows a part of the project structure tree (ctrl + 7) that relates to a particular trait. So, it needs to follow the same representation rules as the project structure. And neither we nor other plugins I saw show such keywords in the project structure because icons and signatures provide enough information to easily understand what the element is.
- Those keywords are the only bold elements in the tree, so they detract attention a bit from the names of the elements, but are really not important parts of what we want to show here (signatures of items to be implemented). And we also don't bold other keywords in the signatures.
- They introduce indentation for item names, so you need to adjust to it when you're trying to find an element you want to select with eyes. This argument plays together with my next comment that is going to appear shortly :)

alygin:
There're two approaches to what to show in this window: either the elements that should be implemented to make the code compile (Kotlin, CLion), or all the elements that could be implemented (Java) and automatically select those not providing a default implementation:

screen shot 2017-02-18 at 12 50 29

The idea in the second approach is that you see all the functionality that the trait provides and can make an informative choice if you want to reimplement something. Icons also show you what elements are already implemented and what are not. And automatic selections make it easy to fall into option 1 with one click.

I think the second approach could be an idea for further improvements of the window (not in this PR). Or, if it's easy to add, implemented from the very beginning.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
alygin: OK, good to know.

slavam2605:
Yes, it is a good idea, but in Java there are imterfaces and base classes and so IDEA can distinguish "implement" and "override". But in Rust we have only traits and my idea was to show only mandatory members in "implement" action and all members in "override".
I don't know what is better.

alygin:
You're right that Java has both "implement" and "override" actions, but the Java's advantage over the same actions in, for example Kotlin, is that it combines them in the interface implementation window, narrowing the list of function to override to those from the chosen interface. You can quickly grasp the interface functionality during implementation and override optional items if you need, without necessity to open another window.

And you're absolutely right that in Java the "override" command is very important and has a great value per se because the language provides class inheritance and multiple interfaces implementations in the same scope (in the class body). In Rust, a trait implementation always narrows you to this particular trait and nothing else. Even if the trait uses inheritance, you need to implement/override the items of the parent trait in a different scope (different `impl` block).

So, "override" would be a cool separate feature in Rust too, but it's always limited to a particular trait and so it would be natural to also incorporate it into "implement", to provide a better user experience.

slavam2605:
Ok, I will try fix all remarks.

slavam2605:
@alygin @matklad 
I fixed some problems, but I haven't change a behavior of `Implement Members` for now (about selecting all mandatory elements etc.). If this PR will be ok I will add this in the next PR with something like `Override Members` and will think about a lot of things that I haven't implement here.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    val canImplementTypes = trait.typeAliasList.associateBy { it.name }
+    val mustImplementTypes = canImplementTypes.filterValues { it.typeReference == null }
+    val implementedTypes = impl.typeAliasList.associateBy { it.name }
+
+    val notImplemented = mustImplement.keys - implemented.keys
+    val notImplementedTypes = mustImplementTypes.keys - implementedTypes.keys
+
+    val toImplement = trait.functionList.filter { it.name in notImplemented }
+    val toImplementTypes = trait.typeAliasList.filter { it.name in notImplementedTypes }
```
alygin: The block above is a duplicate of what we already have in `RsErrorAnnotator`. Maybe we could move it somewhere (`RsImplItemImplMixin`?) and reuse? It could return `toImplement` (must to have yet not implemented items) and `canOverride` (optional and not yet implemented), so it could be used in error annotations, "implement" and "override" commands.

The trait items analysis we do now is straightforward and doesn't care of specialization (see #1009), so we'll soon have to make it smarter. It's better to do it in one place.

alygin:
Apart from the mentioned duplication, it looks good to me.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
slavam2605: Should we mix types and functions together or do separate methods like `toImplementFunctions`, `toImplementTypes`?

slavam2605:
I don't understand, how I can add an arbitrary method in a mixin, because only implementation extends from it, so I've extracted methods to util file.
But I can't use these methods in an error annotator, because it actually requires almost all values from these methods so it looks reasonable to leave it as it is.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
alygin: I would combine them in a single function. The critical point here is the annotator speed, but adding `canOverride` to it doesn't look as a big overhead (especially if make it a `Sequence` instead of `List`), at least for the time being. We can always separate them later.

alygin:
I mean not the `RsImplItemImplMixin` class itself, but the file `RsImplItemImplMixin.kt`, like we do with other items to combine reusable code related to a particular element type: [RsFunctionImplMixin](https://github.com/intellij-rust/intellij-rust/blob/12b4b8fdcc0ffbcfc2881b77a5d5048378847208/src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RsFunctionImplMixin.kt), for example.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -249,17 +249,21 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         val mustImplement = canImplement.filterValues { it.isAbstract }
         val implemented = impl.functionList.associateBy { it.name }
 
+        val canImplementTypes = trait.typeAliasList.associateBy { it.name }
+        val mustImplementTypes = canImplementTypes.filterValues { it.typeReference == null }
```
alygin: @matklad, what do you think of introducing an interface (RsAbstractable) with the `isAbstract` value? We then could get rid of such repentances. Now, `RsTypeAlias` and `RsFunction` (for which we already have `isAbstract`) are good candidates to share this interface.

slavam2605:
I've added some functionality I was going to, but had no time earlier.
Now `implement members` support also associated constants.
Thanks to alygin's good idea, members list now shows all members and select all mandatory ones.

Here is an example of new window:
image

And the code it was showed for:
```rust
#![feature(associated_type_defaults)]
#![feature(associated_consts)]

trait Foo {
    const A: i32;
    const B: i64;
    const C: f64 = 2.0;
    fn foo(i32) -> i32;
    fn bar(i64) -> i64;
    fn baz(i64) {}
    type Type1;
    type Type2;
    type Type3 = &'static str;
}

struct Bar;

impl Foo for Bar {
    const A: i32 = unimplemented!();
    const B: i64 = unimplemented!();
    type Type1 = ();
    type Type2 = ();

    fn foo(_: i32) -> i32 {
        unimplemented!()
    }

    fn bar(_: i64) -> i64 {
        unimplemented!()
    }
}
```

I don't know what is the default value of a constant should be, so it is `unimplemented!()` for now to be compilable.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
slavam2605: Also `RsConstant`, I think.

alygin:
Looks cool!

So, you moved the analysis to the mixin but decided not to use it in the error annotator. You now have all the functions you need to annotate `impl`s with missing items, so why not reuse it in the annotator?

And you decided to break `toImplement*` into three functions (as well as `canOverride*`). The problem here is that they all call `traitRef?.trait` that hides `resolve()` inside, and that's the heaviest operation here. Moreover, these three function will always be used together (error annotator needs all three, "implement" calls all three in row (and then the other three), "override" will always call all three `canOverride`. So, maybe combine them into two functions `toImplementMembers` and `canOverrideMemebers`, each of which returns a list of `RsNamedElement`s?

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.rust.lang.core.psi.impl.mixin.*
+import org.rust.lang.core.psi.util.trait
+import org.rust.utils.getFormattedParts
+import javax.swing.JTree
+
+sealed class TraitMember {
+    class Function(val function: RsFunction) : TraitMember()
+    class Type(val type: RsTypeAlias) : TraitMember()
+    class Constant(val constant: RsConstant) : TraitMember()
+}
```
alygin: I just thought that if you use `RsNamedElement`, you can actually get rid of `TraitMember` and simplify things in this file. Functions, type aliases and constants are all `RsNamedElement`s, and hence inherit `name` and `getIcon()`, so there's no need in those `when`s in `renderTreeNode()` and `getText()`. The only place you need to know the exact type of the element, is when you define the `text`, but you can simply use `is RsFunction` etc there. And after my refactoring of #1020 we'll also remove this last `when`.

slavam2605:
Indeed, I will try to merge six functions into two and maybe optimize it somehow.
And maybe now I will be able to reuse these functions in the error annotator.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
slavam2605: I decided to use sealed class to make the code cleaner and make all `when`s exhaustive, but now it is not so important.

slavam2605:
I've stuck combining functions `toImplement*` because I'm really need a `RsAbstractable` interface to write a good code. It would be extremely great to introduce such interface, but I don't know how I can do this, because such interfaces are generating automatically.

alygin:
>  because such interfaces are generating automatically

They actually don't. Have a look at [this commit](https://github.com/intellij-rust/intellij-rust/commit/27607a17ae2a05ed6aa6b7195fe1a0ba63491dc1). Among other things it introduces the `RsLabeledExpression` interface for `RsForExpr`, `RsLoopExpr` and `RsWhileExpr` which is pretty the same as what you need for `RsAbstractable`. The steps are:
1. Declare the interface yourself in `org.rust.lang.core.psi`.
2. Add it to `implements` collection of those elements you want to implement it (`Const`, `TraitItem` and `Function`)
3. Implement `isAbstract` in the mixin classes of these elements.

@matklad, any arguments against introducing such an interface?

slavam2605:
Yes, now I understand. I will update the PR recently.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -12,51 +12,38 @@ import org.rust.lang.core.psi.util.trait
...
-                text = "${member.type.name}"
+            is RsTypeAlias -> {
+                text = "${member.name}"
             }
-            is TraitMember.Constant -> {
-                text = "${member.constant.name}: ${member.constant.typeReference?.text}"
+            is RsConstant -> {
+                text = "${member.name}: ${member.typeReference?.text}"
             }
+            else -> error("Unknown trait member: " + member)
```
alygin: Could be a bit shorter and a bit clearer:

```rust
text = when (member) {
    is RsFunction -> {
        val (before, after) = member.getFormattedParts()
        "${member.name}$after"
    }
    is RsTypeAlias -> "${member.name}"
    is RsConstant -> "${member.name}: ${member.typeReference?.text}"
    else -> error("Unknown trait member: $member")
}
```

slavam2605:
Oops, something is broken. 
```
e: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class org.rust.lang.core.psi.RsFunction, unresolved supertypes: RsAbstractable
    class org.rust.lang.core.psi.RsConstant, unresolved supertypes: RsAbstractable
    class org.rust.lang.core.psi.RsTypeAlias, unresolved supertypes: RsAbstractable
```
I don't undertand how to fix it honestly.

slavam2605:
Seems like I commit not all files. I hope this commit will fix the problem.

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RsImplItemImplMixin.kt
```diff
@@ -31,3 +30,19 @@ abstract class RsImplItemImplMixin : RsStubbedElementImpl<RsImplItemStub>, RsImp
         return PresentationData(typeReference?.text ?: "Impl", null, RsIcons.IMPL, null)
     }
 }
+
+/**
+ * @return pair of two lists: (mandatory trait members, optional trait members)
+ */
+fun RsImplItem.toImplementOverride(): Pair<List<RsNamedElement>, List<RsNamedElement>> {
+    val trait = traitRef?.trait ?: error("No trait ref")
```
alygin: Here should be `?: return null` instead of throwing error. Resolve engine is not perfect now, so we don't guarantee resolving all the possible trait references (let alone the cases when the a trait name is typed incorrectly). In such cases it's better to tell the caller that we can't collect items and just don't show the window instead of producing an IDE error.

src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RsImplItemImplMixin.kt
```diff
@@ -31,3 +30,19 @@ abstract class RsImplItemImplMixin : RsStubbedElementImpl<RsImplItemStub>, RsImp
...
+    val traitMembers = trait.functionList as List<RsAbstractable> + trait.typeAliasList + trait.constantList
```
alygin: `traitMembers = trait.children.filterIsInstance<RsAbstractable>()` will cover all possible elements without naming them explicitly and without creating several temporary lists. The same with `members`.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -244,22 +244,22 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     private fun checkImpl(holder: AnnotationHolder, impl: RsImplItem) {
         val trait = impl.traitRef?.trait ?: return
         val traitName = trait.name ?: return
-
-        val canImplement = trait.functionList.associateBy { it.name }
-        val mustImplement = canImplement.filterValues { it.isAbstract }
+
```
alygin: We now resolve the trait twice: here and in the `toImplementOverride`. We could just return `null` from `toImplementOverride` if it's not resolved and return from here (see a comment for `toImplementOverride`). Here, we could only check that the trait reference exists and skip the further analysis if it doesn't: `if (impl.traitRef == null) return`

slavam2605: Sorry, I don't understand clearly. In the beginning of `checkImpl` method if `traitRef` is null then we immediately return from function. You suggest to reuse already resolved value of `trait` somehow?

alygin: I mean that we don't need to call `traitRef?.trait` (that resolves trait) anymore since we don't use the trait itself in this method. `toImplementOverride()` will collect all the information we need from the trait and either will return it or will return `null` if it can't resolve the trait. In the latter case we just can return from `checkImpl()` as we do now in `impl.traitRef?.trait ?: return`.

slavam2605: But we use it for `traitName` in this method.

alygin: I think the name can be obtained from `traitRef.text`. Or we could only resolve the trait if we really found an error and need to emit the annotation. But not always.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
matklad: I think there isn't any convention yet? Kotlin itself is not consistent (example [three](https://github.com/JetBrains/kotlin/tree/master/idea/ide-common/src/org/jetbrains/kotlin/util)).

src/main/kotlin/org/rust/ide/core/overrideImplement/ImplementMembersHandler.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.impl.RsFile
+import org.rust.lang.core.psi.util.parentOfType
+
+class ImplementMembersHandler: LanguageCodeInsightActionHandler {
+    override fun isValidFor(editor: Editor, file: PsiFile): Boolean {
+        if (file !is RsFile)
+            return false
```
matklad: Let's either format this as a single line `if (file !is RsFile) return` or add braces to avoid goto fail bug?

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.rust.lang.core.psi.impl.mixin.*
+import org.rust.lang.core.psi.util.trait
+import org.rust.utils.getFormattedParts
+import javax.swing.JTree
+
+private class RsTraitMemberChooserMember(val base: MemberChooserObjectBase, val member: RsNamedElement) : ClassMember {
+    private val text: String;
+
+    init {
+        when (member) {
```
matklad: This can be `val text = when(member) { ... }`

src/main/kotlin/org/rust/ide/core/overrideImplement/ImplementMembersHandler.kt
```diff
@@ -0,0 +1,27 @@
...
+        val elementAtCaret = file.findElementAt(editor.caretModel.offset)
+        val classOrObject = elementAtCaret?.parentOfType<RsImplItem>(strict = false)
+        return classOrObject != null 
+    }
+
+    override fun startInWriteAction() = false
+
+    override fun invoke(project: Project, editor: Editor, file: PsiFile) {
+        val elementAtCaret = file.findElementAt(editor.caretModel.offset)
+        val implItem = elementAtCaret?.parentOfType<RsImplItem>(strict = false) ?: error("No impl trait item")
```
matklad: I think we should not `error` here, and just do nothing? 

`error` is useful to check invariants about PSI you control. For example, if you use `psiFactory` to generate a function, you can be sure that it has a body, and you want to assert this in code. 

In contrast, when the PSI is supplied by the user, it can be broken in different ways, and one should handle `null` gracefully, either by doing nothing, or by showing `Action is not applicable` error popup.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,87 @@
...
+    val allMembers = toOverride.map { RsTraitMemberChooserMember(base, it) }
+    
+    if (allMembers.isEmpty())
+        return
+
+    val chooser = MemberChooser(allMembers.toTypedArray(), true, true, project)
+    chooser.title = "Implement Members"
+    chooser.selectElements(mandatoryMembers.toTypedArray())
+    chooser.setCopyJavadocVisible(false)
+    chooser.show()
```
matklad: `apply` would be a more kotlinish:

```Kotlin
val chooser = MemberChooser(...).apply {
    title = "foo bar"
    show()
}
```

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    val selected = chooser.selectedElements ?: listOf()
+    if (selected.isEmpty())
+        return
+    val templateImpl = RsPsiFactory(project).createTraitImplItem(
+        selected.mapNotNull { it.member as? RsFunction },
+        selected.mapNotNull { it.member as? RsTypeAlias },
+        selected.mapNotNull { it.member as? RsConstant }
```
matklad: Another way to express this is `members.filterIsInstance<RsFunction>`.

src/main/kotlin/org/rust/utils/formatterUtils.kt
matklad: That's a nice plan! Actually, I think tree elements can use some refactoring: I don't think we need separate tree classes for different elements.

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
mkaput: Personally I prefer `camelCase` for files without *main* class, but in intellij-rust codebase we use mostly `UpperCase` similar to class files.

src/main/kotlin/org/rust/lang/core/psi/RsAbstractable.kt
```diff
@@ -0,0 +1,5 @@
+package org.rust.lang.core.psi
+
+interface RsAbstractable : RsNamedElement {
```
matklad: :+1:

matklad:
This looks awesome! The only thing missing are tests!

The main thing we want to test is [generateTraitMembers](https://github.com/intellij-rust/intellij-rust/pull/1023/files#diff-d3d55455ab063cc1a2876ba565322b64R48) function. 

It has a following pattern: 

* collect some data A from the PSI
* show the data in the UI, to get some response B
* use B to apply changes to the PSI. 

I suggest we split this function into these three parts, to be able to test parts one and three. 

I imagine the test can look like this:

```Kotlin
class OverrideTest {
    fun `test simple`() = doTest("""
        trait T {
            fn foo();
            fn bar(&self) {}
            type T;
        }

        struct S;
        impl T for S {
            /*caret*/
        }
    """, """
        // `*` means the user selected this item, `s` is the initial state of the selection    
        * s foo()
        *   bar(&self)
        * s T
    """, """
        trait T {
            fn foo();
            fn bar(&self) {}
            type T;
        }
        
        impl T for () {
            type T = ();
        
            fn foo() {
                unimplemented!()
            }
        
            fn bar(&self) {
                unimplemented!()
            }
        }
    """)

    private fun doTest(
        @Language("Rust") code: String,
        abstractMembers: String,
        @Language("Rust") expected: String) {
    }
}
```

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -73,11 +73,14 @@ class RsPsiFactory(private val project: Project) {
...
-            .joinToString("\n\n")
-        val text = "impl T for S { $methods }"
+    fun createTraitImplItem(traitMethods: List<RsFunction>, traitTypeAliases: List<RsTypeAlias>, traitConstants: List<RsConstant>): RsImplItem {
+        val members = (
+                    traitConstants.map { " const ${it.identifier.text}: ${it.typeReference?.text} = unimplemented!();" } +
+                    traitTypeAliases.map { " type ${it.name} = ();" } +
+                    traitMethods.map { " \n${it.signatureText} {\nunimplemented!()\n}" }
+                ).joinToString("\n")
+        val text = "impl T for S { \n$members\n }"
+        println(createFromText<RsImplItem>(text)?.text)
```
matklad: :)

slavam2605: Oops)

slavam2605:
How can we use a `abstractMembers` argument? It shoud be parsed somehow? Should we compare it with actual UI created? Something like take a `chooser` and check that selected items are correct?

matklad:
>How can we use a abstractMembers argument? It shoud be parsed somehow? Should we compare it with actual UI created? Something like take a chooser and check that selected items are correct?

Yep, I was thinking that chooser is parametrized by a list of `Entry(val item: RsAbstractable, val shouldBeSelectedByDefault: true)`, and that you can render a list of entries to string and compare that with `abstractMembers` argument. That is, instead of parsing `abstractMemmbers`, go in the opposite direction and render the actual abstract members to string.

slavam2605:
I don't understand the meaning of `RsHighlightingAnnotatorText.testContextualKeywords`. Could someone explain what does it test? And what meaning `defult fn` has in Rust?

matklad:
In rust, some keywords (`default` and `union`). That means that they can be used both for names of local variables, and to designate special syntax: `union S {}` is an untagged union, and `default fn foo()` is a function subject to specializaton.

So, during the lexer pass, `default` and `union` are treated as usual identifiers. Then, during the parsing, we see if `default` or `union` are in the keyword position, and *remap* their tokens to keyword. 

One problem with them is that we need to highlight them as keywords, but only if they are actually keywords. This can't be done in the lexer-based highlighting pass, so this is handled by the annotator. This test checks that they are actually highlighted as keyword.

mkaput:
Rust has a thing called contextual keywords: they are not part of lexical definition of the language, and therefore are valid identifiers normally, but in some contexts they are treated as keywords, e.g. `default`, which is only valid before methods definition in trait. It is used in specialization: https://github.com/rust-lang/rfcs/blob/a4a22d7c5dd71724bb2cd0fe2db5026338d0b270/text/1210-impl-specialization.md

slavam2605:
Thanks a lot :)

src/test/kotlin/org/rust/ide/core/overrideImplement/OverrideTraitMembersTest.kt
```diff
@@ -0,0 +1,82 @@
...
+            if (member in selectedSet)
+                builder.append("*  ")
+            else
+                builder.append("   ")
+            builder.append(member.formattedText()).append("\n")
+        }
+        builder.deleteCharAt(builder.lastIndex)
+        return builder.toString()
+    }
+}
```
matklad: Yes, this looks like a nice way to test this feature! 

Can we add a couple more test cases, including implementing type aliases, and not implementing already present members?

src/test/kotlin/org/rust/ide/core/overrideImplement/OverrideTraitMembersTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    """, """
+        
+    """)
+
+    private fun doTest(@Language("Rust") code: String,
+                       chooser: String,
+                       @Language("Rust") expected: String) {
+        checkByText(code, expected) {
+            val impl = myFixture.elementAtCaret.parentOfType<RsImplItem>()
+                    ?: fail("Caret is not in an impl block")
```
matklad: Nice! It didn't occur to me that one can use JUnit's `fail` instead of Kotlin's `error` in tests!

src/test/kotlin/org/rust/ide/core/overrideImplement/OverrideTraitMembersTest.kt
slavam2605: Of course, I'm sorry, I was rather busy. Now I'm rebasing and apply alygin's changes to this code.

src/test/kotlin/org/rust/ide/core/overrideImplement/OverrideTraitMembersTest.kt
slavam2605: This approach has a problem: JUnit's `fail` returns `Unit` instead of `Nothing`, so I had to write a helper method for that.

src/test/kotlin/org/rust/ide/core/overrideImplement/OverrideTraitMembersTest.kt
matklad: No worries! Thanks a lot for this work!

src/test/kotlin/org/rust/ide/core/overrideImplement/OverrideTraitMembersTest.kt
matklad: >This approach has a problem: JUnit's fail returns Unit instead of Nothing, so I had to write a helper method for that.

Hm, I've missed your comment and written a similar method myself :) 

https://github.com/intellij-rust/intellij-rust/blob/8a3be2879315d29df6a46a4e99929cbd40fc1bbc/src/test/kotlin/org/rust/lang/RsTestBase.kt#L278

slavam2605:
Hey, I've eventally solve all my problems and made a rebasing and some tests :)
Seems working now, I hope.
But there is a last problem: to generate function definition I use @alygin's `presentationInfo?.projectStructureItemText`, but in the case of anonymous function's argument `foo(i32)` it generates text `foo(i32)`, but I want to generate function with signature `foo(_: i32)`, because the first one is invalid.
So maybe there are already some tools to do it?

matklad:
Cool!

`RsFunction.signatureText` handles adding `_`.

slavam2605:
`signatureText` inserts two pairs of extra spaces around the `()`. I will remove them and commit again.

slavam2605:
Or maybe I can't do it.
`signatureText` returns `"fn $name $generics (${allArguments.joinToString(",")}) $ret $where"`
For empty `$generics` and `$ret` it generates extra spaces.
But I can't just remove them, because instead it will generate something like `fn foo() -> i32where ...`
So I don't know how to solve it.

alygin:
I think fixing `signatureText` should be quite simple. Just make `generics` (and `where` as well) include the delimiting space when it's present and be empty string when it's not. The template will be `fn $name$generics(${allArguments.joinToString(",")}) $ret$where`. That should remove extra spaces.

alygin:
I even think `generics` never needs spaces. So, it should be something like:

```kotlin
private val RsFunction.signatureText: String? get() {
    // ...
    val ret = (retType?.text ?: "").withLeftSpace
    val where = (whereClause?.text ?: "").withLeftSpace
    return "fn $name$generics(${allArguments.joinToString(",")})$ret$where"
}

private val String.withLeftSpace: String get() = if (isEmpty()) this else " " + this
```

slavam2605:
Thanks) I've fixed it and make last test more complex.

slavam2605:
I hope it's done now.

matklad:
:rocket:

matklad:
Thanks a lot!

src/main/kotlin/org/rust/ide/core/overrideImplement/generateUtil.kt
```diff
@@ -87,13 +81,11 @@ fun insertNewTraitMembers(selected: Collection<RsTraitMemberChooserMember>, impl
...
     )
     val lastMethodOrBrace = impl.functionList.lastOrNull() ?: impl.lbrace ?: return
-    runWriteAction {
-        impl.addRangeAfter(
-                templateImpl.lbrace?.nextSibling,
-                templateImpl.rbrace?.prevSibling,
-                lastMethodOrBrace
-        )
-    }
+    impl.addRangeAfter(
```
alygin: I've got an exception while performing automatic trait implementation. It looks like its cause could be this `runWriteAction` removal? @slavam2605, @matklad, how do you think, should we put it back?

```
AWT events are not allowed inside write action: java.awt.event.InvocationEvent[INVOCATION_DEFAULT,runnable=com.intellij.openapi.util.ActionCallback$$Lambda$257/2117917341@55337b94,notifier=null,catchExceptions=false,when=1490370577201] on sun.lwawt.macosx.LWCToolkit@1ae77ddd
java.lang.Throwable
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:132)
	at com.intellij.openapi.application.impl.NoSwingUnderWriteAction.lambda$watchForEvents$0(NoSwingUnderWriteAction.java:36)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:376)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:184)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:229)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:227)
   ...
```

slavam2605: I think I removed it because tests are running in write action (I set them up this way), so I forgot that in real work it can crash.

matklad: Should be fixed in: https://github.com/intellij-rust/intellij-rust/commit/c8811197553f69257fb53721dac52071fa128fe2

PR #1029, opened 2017-02-19
[EARLY WIP]: Test runner

mkaput:
This PR is going to implement test runner for cargo tests. And yes, it's going to be huge.

This is early work in progress. I'm posting this because I need to discuss some design details.


src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
```diff
@@ -0,0 +1,203 @@
...
+import jetbrains.buildServer.messages.serviceMessages.ServiceMessageVisitor
+
+/**
+ * Translate Cargo Test output to TeamCity Service Messages which SM Test Runner understands.
+ *
+ * Unfortunately, as for now Cargo Test only supports `human` message format, which is pretty
+ * obscure to handle programmatically. First of all, it is split into some separate sections,
+ * which are not easily distinguishable, that's why we do state machines here. Next problem is,
+ * that Cargo collects failed tests stdouts and print them yet at the end of test session.
+ * Lastly, we do not have exact information when given test started.
```
mkaput: I have no idea whether the thing I am doing here makes sense.

Cargo test output is terrible for machine parsing, especially test failures. This is fact. Currently I'm trying to collect failed tests, then collect their stdouts and print everything at the end of the test session (just like TeamCity plugin does). This has the advantage that we bind tests with their outputs, but we do not have information about them until end of run, also it complicates thing drastically.

Otherwise we can put this PR on hold, and do some cargo refactorings to support SM or other more machine-friendly format. I can try to do it, but I'm short on spare time and haven't made any contribution to Rust core projects, so this would take a long time to complete.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
```diff
@@ -0,0 +1,45 @@
...
+    override fun getLocation(
+        protocol: String,
+        path: String,
+        project: Project,
+        scope: GlobalSearchScope
+    ): List<Location<PsiElement>> {
+        if (protocol == TEST_PROTOCOL) {
+            val resolveResult = ResolveEngine.resolve(path, project) ?: return emptyList()
+            return listOf(PsiLocation.fromPsiElement<PsiElement>(project, resolveResult.element))
+        }
```
mkaput: I think using resolve engine isn't the way to go here, but this should use `GlobalSearchScope`, especially that we can set up custom one in `CargoTestConsoleProperties#initScope`. But does anyone know how these scopes work? I struggle to find any sensible resource about them.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestConfiguration.kt
```diff
@@ -0,0 +1,80 @@
...
+class CargoTestConfiguration(
+    project: Project,
+    name: String,
+    factory: ConfigurationFactory
+) : CargoConfigurationBase(project, name, factory),
+    RunConfigurationWithSuppressedDefaultDebugAction {
+
+    // FIXME This is stub implementation and almost 1:1 copy-paste of CargoCommandConfiguration!
+    //       We should store direct test path and provide dedicated UI, so user would be able
+    //       to set up run config manually in friendly manner.
```
mkaput: Pinging myself to not forget to implement this 😉

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
```diff
@@ -0,0 +1,203 @@
...
+
+    private fun message(sm: String) {
+        val message = requireNotNull(ServiceMessage.parse(sm)) { "message processor should build valid SMs" }
+        message.visit(requireNotNull(visitor) { "visitor hasn't be initialized" })
+    }
+
+    private fun message(smb: ServiceMessageBuilder) = message(smb.toString())
+
+    // Kotlin chokes on inner sealed classes, that's why we use enums here
+    private enum class State {
```
matklad: Ah, reminds me of good ol' days of jps: https://github.com/intellij-rust/intellij-rust/blob/239096bfc0d486acac17e1669ed9b34e4b665241/jps/src/main/kotlin/org/rust/jps/RustBuilder.kt#L123

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
matklad: Basically there are two ways to search for a thing in Rust: given a crate root and a path, resolve the path inside the crate. This is what Resolve Engine does.

Another option is to search for the last identifier in the index (we can't store the whole path in index because it depends on more than one file. Compare this with Kotlin, where you always have a handy `package` clause at the beginning of the file).

`SearchScopes` are used in the second case in the query to the index to do `VirtualFile` based filtering. 

I think either approach would work in Rust, but we need to know which Cargo package/target are we talking about. This info should be available to `CargoTestConsoleProperties` because it knows the cargo command line it will execute.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
mkaput: > I think this can be worked around by test -- --nocapture --test-threads 1.

Nice! This is probably the way to go (as for now).

> Without --test-threads the output can be interleaved in arbitrary ways I think.

Yep, that's probably why rust test is collecting stdout and printing them at the end

> Adding support for custom reporters to Rust tools would be great, but this is more complicated than it seems. The first thing to understand is that you are not interacting with cargo here, but with a test harness binary. I think this is it https://github.com/rust-lang/rust/blob/master/src/libtest/lib.rs, but I haven't worked with it.

I don't see any problems with this. We are already passing some arguments to test binary (note `test -- ...`) and it doesn't look very complicated. So I think extending it shouldn't be a huge problem, especially as not only we would benefit from it.

> There is a consensus that it should be possible to plug a custom test runner instead, which would solve a whole host of problems, including a custom error reporting, but looks like nobody is actively working on it (because this is not a small task!).

Indeed, this is not small task

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
mkaput: Ok. I think I can add module name to url

mkaput:
> I am not sure that we need a separate CargoTestConfiguration and associated classes though?

My idea is/was to provide dedicated UI (like JUnit or Python tests) and hide passing custom arguments to cargo from user.

> Can we just automatically create the test console view if we see test command in the command line? Like gradle does, when you execute :test gradle task

Yes we can. Hmm I can try to do this (run test console if, and only if we issue `cargo test <foo> -- --nocapture --test-threads 1`). What do you think?

matklad:
>My idea is/was to provide dedicated UI (like JUnit or Python tests) and hide passing custom arguments to cargo from user.

Why do you want a dedicated UI? In general, I'd love to make UI exactly as invoking `cargo` through command line. Action to run cargo commands is a step in the this direction. I think this is better then our own custom GUI because Cargo is ubiquitous in Rust ecosystem, while in Python and Java there's more different tools your can use. 

An important concrete benefit of a single UI would be that if the user crafts a test command by hand, we will still create a test console. 


>run test console if, and only if we issue `cargo test <foo> -- --nocapture --test-threads 1`

I think we should go a bit further and just add `--nocapture` and `--test-threads 1` to any test command.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
matklad: >So I think extending it shouldn't be a huge problem, especially as not only we would benefit from it.

The problem would be to agree on the format I think, it's probably worth an RFC. A nice prior art is https://github.com/dart-lang/test/blob/master/doc/json_reporter.md

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
matklad: >Ok. I think I can add module name to url

Ideally, we want to know an IDEA module to get hold of the relevant `Workspace`, and we want Cargo's (package spec, target) pair to identify a crate within workspace.

matklad:
Together with `--nocapture`, I think it may be worthwhile to add `--no-fail-fast`. 

Changing defaults makes sense because they are oriented for the command line usage, where it is difficult to filter by test.

mkaput:
Unfortunately, I am currently completely busy because of both my studies and work, and I have no time to work on this anymore. I'm sorry. Hoping someone would like to finish this idea :)

PR #1033, opened 2017-02-23
REF: Use PresentationInfo in project structure and navigation elements

alygin:
Leverage `RsNamedElement.presentationInfo` to show elements in the project structure tree and in various navigation elements (like `Show usages of struct 'Foo'` instead of `Show usages of Rs Struct Item Impl 'Foo'`).

The fuzzy-test is ok :)


alygin:
Also inlined project structure view tests

matklad:
Thanks!

PR #1034, opened 2017-02-24
REN: Don’t fail on wrong id type while renaming

alygin:
There's a problem with renaming now: `RsNameValidator` hasn't a clue about the type of the element being renamed. So, it has to always allow both regular `identifier`s and `quote_identifier`s because we want to be able to rename everything. This leads to throwing an error when you provide a name of a wrong type (for instance, `foo` for a lifetime, or `'foo` for a function).

This change introduces an implementation of `RenamePsiElementProcessor` that substitutes wrong names to valid ones (by adding or removing a quote sign when needed) thus preventing mentioned errors.

The problem could have been solved in the `RsReferenceBase.doRename()`, but `RenamePsiElementProcessor` provides additional functionality that will allow us to do other refactoring improvements (like handling name conflicts), so I think it's good to have one.

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/RsRenameProcessor.kt
@@ -1,0 +1,26 @@
+package org.rust.lang.refactoring
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.search.SearchScope
+import com.intellij.refactoring.rename.RenamePsiElementProcessor
+import org.rust.lang.core.psi.RsLabel
+import org.rust.lang.core.psi.RsLabelDecl
+import org.rust.lang.core.psi.RsLifetime
+import org.rust.lang.core.psi.RsLifetimeDecl
+
+class RsRenameProcessor : RenamePsiElementProcessor() {
+
+    override fun canProcessElement(element: PsiElement): Boolean = true
+
+    override fun prepareRenaming(element: PsiElement, newName: String?, allRenames: MutableMap<PsiElement, String>, scope: SearchScope) {
+        if (newName == null) return
+        if (element is RsLifetime || element is RsLifetimeDecl || element is RsLabel || element is RsLabelDecl) {
+            allRenames.put(element, newName.ensureQuote())
+        } else {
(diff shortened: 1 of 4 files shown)


alygin:
BTW, `RsNameValidator` rejects names of primitive types, but Rust treat them as valid names. Here's a valid Rust code:

```rust
struct u8 {
    f64: f64,
    str: &'static str
}

impl u8 {
    fn bool() {}
}
```

@mkaput (git says you're the author of the validation code :), what was the reason of such limitations? Maybe we should remove this check?

src/main/kotlin/org/rust/lang/refactoring/RsRenameProcessor.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.rust.lang.core.psi.RsLabel
+import org.rust.lang.core.psi.RsLabelDecl
+import org.rust.lang.core.psi.RsLifetime
+import org.rust.lang.core.psi.RsLifetimeDecl
+
+class RsRenameProcessor : RenamePsiElementProcessor() {
+
+    override fun canProcessElement(element: PsiElement): Boolean = true
+
+    override fun prepareRenaming(element: PsiElement, newName: String?, allRenames: MutableMap<PsiElement, String>, scope: SearchScope) {
```
matklad: I think `newName` should be just `String`, without `?`

alygin: I thought that too, but I see that Kotlin [allows nulls and has checks for that](https://github.com/JetBrains/kotlin/blob/3471b3311f8358a14ab21d957c6137a84615484a/idea/src/org/jetbrains/kotlin/idea/refactoring/rename/RenameKotlinPsiProcessor.kt#L61), so maybe there're some special cases? Though it seems unlikely, I'll remove it.

mkaput:
Well I thought they are invalid :p

alygin:
Then I propose to remove the check and let the user do weird things if he or she wants. Any objections?

matklad:
>Then I propose to remove the check and let the user do weird things if he or she wants. Any objections?

Sounds good to me!

matklad:
Thanks!

PR #1035, opened 2017-02-24
TY: Type inference for array sizes and enum variant discriminants

alygin:
Infer integer literal kinds for array sizes and enum variant discriminants. It will help with a couple of error annotations.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/util/RustExprExtensions.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/util/RustExprExtensions.kt
@@ -5,7 +5,7 @@ import org.rust.lang.core.psi.*
 /**
  *  `RsExpr` related extensions
  */
-// TODO: rename to make it clear that these are filed of the type and not of the expression.
+// TODO: rename to make it clear that these are fields of the type and not of the expression.
 val RsStructExpr.fields: List<RsNamedElement>
     get() = (path.reference.resolve() as? RsFieldsOwner)?.namedFields.orEmpty()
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/psi/impl/mixin/RsArrayExprMixin.kt
```diff
@@ -0,0 +1,11 @@
+package org.rust.lang.core.psi.impl.mixin
+
+import com.intellij.lang.ASTNode
+import org.rust.lang.core.psi.RsArrayExpr
+import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.psi.impl.RsCompositeElementImpl
+
+abstract class RsArrayExprMixin(node: ASTNode) : RsCompositeElementImpl(node), RsArrayExpr
```
matklad: Why do we need mixin? Can we just define an extension method? 

We don't have a single place to define mixins for elements, but you could put this in org/rust/lang/core/psi/util/RustExprExtensions.kt.

I think it's wrong that we place extension methods in `impl/mixin` files. It's confusing to find an interface in an impl  package, and not every PSI element has a mixin. I'll try to organize this better.

alygin: Given how the current code is organized now, I thought this was the right place for such extension method. `RustExprExtensions` doesn't look like a popular place among extensions at the time being :)

OK, I'll move the method there if it's the right place in the long term.

matklad:
>Given how the code is organized now, I thought this was the right place for such extension method

Given the current hectic organization the place for extension itself is OK. But we definitely do not need mixin, because we do not override anything. Mixins are useful only for providing implementations of virtual functions :)

matklad:
Thanks! I'll try to reorganize extension methods then!

PR #1041, opened 2017-02-27
more icons

### LINKED ISSUE #1038: Use different icon for `mod.rs` files
In structure view, it's difficult to spot `mod.rs` file (unlike, say `__init__.py`). We could use a different icon for it I think...

mkaput:
New icons for `build.rs`, `mod.rs` & `main.rs/lib.rs`. I had no idea what symbol to choose for two former ones, so I opted for colouristic variants.

illustrator_2017-02-27_21-24-57
java_2017-02-27_21-24-23
java_2017-02-27_21-24-35

I have removed svgs in favour of Illustrator files, because this format is easier to modify

fix #1038


matklad:
Super!

matklad:
We might want to think about replacing colors with different icons (like in the build.rs case), to help colorblind people. And I'll take a look if we can easily move `mod.rs` to the top.

matklad:
https://github.com/intellij-rust/intellij-rust/commit/29d42e6bdfe3f19bc0d697f6b88d8d9558ebcd58

matklad:
fire

matklad:
And we might want to look at something like `NestingTreeStructureProvider`, to really display the module structure instead of file structure. 

Though I'd prefer to stick with the current solution: it's less magic there.

johnthagen:
Just want to say this is one of those features I didn't know I needed until it was added. Great job! 👍

PR #1042, opened 2017-02-27
RES: Resolve lifetime declarations in polybounds

### LINKED ISSUE #1040: Highlight error
Consider this code
```
fn main() {
    fn call_with_ref<F>(some_closure: F) -> i32
        where F: for<'a> Fn(&'a i32) -> i32 {
        1
    }
}
```
You will notice that the `'a` is highlighted in red color.

Screenshot below:
http://joxi.ru/l2Z16e7cwaDBKm

alygin:
Fixes #1040


matklad:
Thanks!

It's interesting that I don't actually know what this polybound stuff means in Rust. I know that you can write `T: for<'a> X` and `T: (for<'a> X)`, and they parse differently by rustc. You can even write `T: for<'a> for<'b> X` and the two for's are different.

PR #1047, opened 2017-03-01
ANNOT: Add unsafe error annotator and add quick fix

farodin91:


farodin91:
@alygin r+

alygin:
Sure! I'll have a look as soon as time allows. Hopefully, tomorrow.

src/main/kotlin/org/rust/ide/annotator/fixes/AddUnsafeFix.kt
```diff
@@ -0,0 +1,27 @@
...
+    override fun invoke(project: Project, file: PsiFile, editor: Editor?, startElement: PsiElement, endElement: PsiElement) {
+        val unsafe = RsPsiFactory(project).createUnsafe()
+        when(expr) {
+            is RsBlockExpr -> {
+                expr.addBefore(unsafe, expr.block)
+            }
+            is RsFunction -> {
+                expr.addBefore(unsafe, expr.fn)
+            }
+        }
```
alygin: Could be simplified:

```rust
when (expr) {
    is RsBlockExpr -> expr.addBefore(unsafe, expr.block)
    is RsFunction -> expr.addBefore(unsafe, expr.fn)
}
```

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -51,11 +49,41 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
             override fun visitValueParameter(o: RsValueParameter) = checkValueParameter(holder, o)
             override fun visitVis(o: RsVis) = checkVis(holder, o)
             override fun visitBinaryExpr(o: RsBinaryExpr) = checkBinary(holder, o)
+            override fun visitCallExpr(o: RsCallExpr) = checkCall(holder, o)
```
alygin: Let's also stick with naming handlers as the corresponding elements are named: `checkCallExpr()` instead of `checkCall()` in this case. It makes searching for visitors and handlers of a particular elements easier.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -51,11 +49,48 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
+                    is RsBlockExpr -> it.unsafe != null
+                    is RsFunction -> true
+                    else -> false
+                }
+            }
+            .firstOrNull() ?: return
+
+        if (parent is RsBlock || (parent is RsFunction && parent.unsafe == null)) {
+            val annotation = holder.createErrorAnnotation(o, "Call to unsafe function requires unsafe function or block [E0133]")
+            annotation.registerFix(AddWrapUnsafeFix(o))
```
alygin: There're cases when we can't apply such fix because it would break code:

```rust
unsafe fn pi() -> f64 { 3.14 }
fn main() {
    let s = pi() * 10;
}
```

I think we could only add this fix when the call expression's parent is `RsBlock`.

src/main/kotlin/org/rust/ide/annotator/fixes/AddWrapUnsafeFix.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsBlock
+import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.psi.RsExprStmt
+import org.rust.lang.core.psi.RsPsiFactory
+
+class AddWrapUnsafeFix(val expr: RsExpr) : LocalQuickFixAndIntentionActionOnPsiElement(expr) {
+    override fun getFamilyName() = text
+    override fun getText() = "Add unsafe block around function call"
```
alygin: I'm unaware of strict guidelines, but names of such fixes usually start with "Surround". I think this one could be named "Surround with unsafe block", so its meaning would be easy to grasp quickly.

src/test/kotlin/org/rust/ide/annotator/fixes/AddUnsafeTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+""")
+
+    fun `test wrap function with a unsafe block`() = checkQuickFix("Add unsafe block around function call", """
+        unsafe fn foo() { return; }
+
+        fn main() {
+            foo()/*caret*/;
+        }
+""", """
```
alygin: quotes should be indented too :)

src/test/kotlin/org/rust/ide/annotator/RsExpressionAnnotatorTest.kt
```diff
@@ -154,4 +154,64 @@ class RsExpressionAnnotatorTest : RsAnnotatorTestBase() {
...
+        unsafe fn foo() {}
+
+        fn main() {
+            unsafe {
+                fn bar() {
+                    <error>foo()</error>;
+                }
+            }
+        }
+    """)
```
alygin: I think we need at least one more test to check that it's ok to make unsafe calls within unsafe functions:

```rust
unsafe fn foo() {}
unsafe fn bar() { foo(); }
```

src/test/kotlin/org/rust/ide/annotator/fixes/AddUnsafeTest.kt
```diff
@@ -0,0 +1,89 @@
...
+""")
+
+    fun `test wrap function with a unsafe block 2`() = checkQuickFix("Add unsafe block around function call", """
+        unsafe fn foo() { return; }
+
+        fn main() {
+            foo()/*caret*/
+        }
+""", """
+        unsafe fn foo() { return; }
```
alygin: Just a small remark: there's no need in these `return`s in empty blocks.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
farodin91: I found a correct solution.
```rust
unsafe fn pi() -> f64 { 3.14 }
fn main() {
    let s = unsafe { pi() } * 10.0;
}
```

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -40,6 +40,9 @@ class RsPsiFactory(private val project: Project) {
     fun createBlockExpr(body: String): RsBlockExpr =
         createExpression("{ $body }") as RsBlockExpr
 
+    fun createUnsafeBlockExpr(body: String): RsBlockExpr =
+            createExpression("unsafe { $body }") as RsBlockExpr
```
alygin: broken indentation

src/main/kotlin/org/rust/ide/annotator/fixes/SurroundWithUnsafeFix.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.rust.lang.core.psi.RsExprStmt
+import org.rust.lang.core.psi.RsPsiFactory
+
+class SurroundWithUnsafeFix(val expr: RsExpr) : LocalQuickFixAndIntentionActionOnPsiElement(expr) {
+    override fun getFamilyName() = text
+    override fun getText() = "Surround with unsafe block"
+
+    override fun invoke(project: Project, file: PsiFile, editor: Editor?, startElement: PsiElement, endElement: PsiElement) {
+        val parent = expr.parent
+        when(parent) {
```
alygin: missing space after `when`

farodin91:
@alygin I fixed the formatting nits.

matklad:
Thanks @farodin91 @alygin !

PR #1057, opened 2017-03-07
GRAM: Fix cast expressions parsing

### LINKED ISSUE #1056: Const binding incorrectly highlighted when after "as" cast
The following code snippet is not highlighted correctly in intellij-rust 0.1.0.1698

```rust
const CONSTANT: f64 = 10.0;

struct MyStructure;

pub fn fun() {
    let x = 1 as f64 + CONSTANT;
    //                 ^------- incorrect highlighting
}
```

screen shot 2017-03-07 at 8 37 26 am

alygin:
Only allow scalar types in cast expressions.

Though, I'm not sure this is the optimal way to extract scalar types from `TypeReference`. I tried to make `ScalarTypeReference` a part of `TypeReference` but it caused warnings about left-recursion, so I left it this way for the time being.

Fixes #1056


src/main/grammars/RustParser.bnf
```diff
@@ -657,6 +657,10 @@ private ImplMember_recover ::= !('#' | Vis | extern | const | static | unsafe |
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 
 private TypeAscription ::= ':' TypeReference
+private ScalarTypeReference ::= ArrayType
+                              | RefLikeType
+                              | FnPointerType
+                              | BaseType
```
matklad: Yes, this is the best way to do this!

matklad:
:+1:

PR #1058, opened 2017-03-07
Various small fixes

kumbayo:
This are various small problems that were found while working on other code parts.

If you have an idea how to properly test for any of the changed behaviour I would be happy to add new tests.


matklad:
Oh my god, that's some embarrassing issues! Thank you very much!

I am not sure if it is possible to test this nicely: generally, tests work with AST and not stubs. We can write a test that executes some functions on AST based element and then on the stub based one, but I am afraid that that would add even more boilerplate to already boilerplatish stubs...

Issue #1059, opened 2017-03-07
NoClassDefFoundError when trying to use the Debugger Plugin


ashleysommer:
I downloaded the Debugger Plugin v 0.0.1.55 from TeamCity and installed it on CLion 2016.3.3.
When attempting to add a "Cargo Debug" run configuration, the plugin crashes with this stacktrace:
```
org/rust/cargo/runconfig/RustRunConfigurationModule
java.lang.NoClassDefFoundError: org/rust/cargo/runconfig/RustRunConfigurationModule
	at org.rust.debugger.runconfig.CargoDebugConfiguration.<init>(CargoDebugConfiguration.kt:38)
	at org.rust.debugger.runconfig.CargoDebugConfigurationType$1.createTemplateConfiguration(CargoDebugConfiguration.kt:32)
	at com.intellij.execution.configurations.ConfigurationFactory.createTemplateConfiguration(ConfigurationFactory.java:76)
	at com.intellij.execution.impl.RunManagerImpl.getConfigurationTemplate(RunManagerImpl.java:323)
	at com.intellij.execution.impl.RunManagerImpl.doCreateConfiguration(RunManagerImpl.java:142)
	at com.intellij.execution.impl.RunManagerImpl.createConfiguration(RunManagerImpl.java:137)
	at com.intellij.execution.impl.RunConfigurable.createNewConfiguration(RunConfigurable.java:1104)
	at com.intellij.execution.impl.RunConfigurable$MyToolbarAddAction.a(RunConfigurable.java:1138)
	at com.intellij.execution.impl.NewRunConfigurationPopup$1.onChosen(NewRunConfigurationPopup.java:82)
	at com.intellij.execution.impl.NewRunConfigurationPopup$1.onChosen(NewRunConfigurationPopup.java:48)
	at com.intellij.ui.popup.list.ListPopupImpl.a(ListPopupImpl.java:399)
	at com.intellij.ui.popup.list.ListPopupImpl.handleSelect(ListPopupImpl.java:351)
	at com.intellij.ui.popup.list.ListPopupImpl$MyMouseListener.mouseReleased(ListPopupImpl.java:504)

```
I know on the 'Implement Debugger' issue page here: https://github.com/intellij-rust/intellij-rust/issues/535 it says "It may work with CLion `2016.3.2` and intellij-rust `0.1.0.1611`". I am using CLion `2016.3.3` with intellij-rust `0.1.0.1698`. Might that be the cause of the problem?

[...]

PR #1060, opened 2017-03-08
Update Debugger plugin to work with intellij-rust plugin 0.1.0.1698

ashleysommer:
Update Debugger plugin to work with intellij-rust plugin 0.1.0.1698
Also updated CLion dependency from 16.3.2 to 16.3.3.
Tested and now works fine on CLion 16.3.3 with intellij-rust-plugin 0.1.0.1698

Fixes https://github.com/intellij-rust/intellij-rust/issues/1059


matklad:
Thanks!

PR #1065, opened 2017-03-11
RUN: Run config UI fixes

alygin:
Various small fixes of the run configuration edit window:

1. Add title to the "Additional arguments" window to fix `Assertion failed: Dialog title shouldn't be empty or null: []` exception that sometimes occurs.
2. Make fields wide again!
3. Add quick focus hotkeys to the fields with no such keys.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/ui/CargoRunConfigurationEditorForm.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/ui/CargoRunConfigurationEditorForm.kt
@@ -11,7 +11,9 @@ import com.intellij.ui.RawCommandLineEditor
 import com.intellij.util.execution.ParametersListUtil
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
 import org.rust.cargo.toolchain.CargoCommandLine
+import java.awt.Dimension
 import javax.swing.JComponent
+import javax.swing.JPanel
 import javax.swing.JTextField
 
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/backcompat/ui/layout/MigLayoutBuilder.kt
```diff
@@ -281,9 +281,9 @@ private fun createComponentConstraints(constraints: Array<out CCFlags>? = null,
         cc().split = split
     }
 
-    if (growPolicy == GrowPolicy.SHORT_TEXT) {
-//    cc().minWidth("210")
-        cc().maxWidth("210")
+    when (growPolicy) {
+        GrowPolicy.SHORT_TEXT -> cc().maxWidth("210")
+        GrowPolicy.GROW_X -> cc().growX()
```
matklad: This file is copy-pasted from the IDEA, and we'll get rid of it once we drop support for 2016.2, so it's better to handle this elsewhere.

matklad: I've done something similar here: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/runconfig/ui/RunCargoCommandDialog.kt#L27

https://github.com/JetBrains/intellij-community/commit/8a6de5742bcb821f4a56fdd9f67ed495a42212f2

matklad:
Thanks!

PR #1066, opened 2017-03-13
RUN: Support short backtraces

alygin:
Rust now supports [short backtraces](https://github.com/rust-lang/rust/pull/38165) in nightly.

This PR:
1. Changes the "Print backtrace" flag to backtrace mode combo box (No, Short, Full). "Short" will work the same way as "Full" in beta and stable until the feature is promoted. Full is default.
2. Adds support of short backtraces to the filter that "linkifies" function names.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/BacktraceMode.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/BacktraceMode.kt
@@ -1,0 +1,14 @@
+package org.rust.cargo.toolchain
+
+enum class BacktraceMode(val index: Int, val title: String) {
+    NO(0, "No"),
+    SHORT(1, "Short"),
+    FULL(2, "Full");
+
+    override fun toString(): String = title
+
+    companion object {
+        val DEFAULT = FULL
+        fun fromIndex(index: Int) = BacktraceMode.values().find { it.index == index } ?: DEFAULT
+    }
+}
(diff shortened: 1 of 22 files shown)


src/main/kotlin/org/rust/cargo/toolchain/BacktraceMode.kt
```diff
@@ -0,0 +1,9 @@
+package org.rust.cargo.toolchain
+
+object BacktraceMode {
```
matklad: This looks like a `enum class` to me.

matklad:
Another option is to always use the full format and hide irrelevant details on the IDE side, like in Java backtracks. 

But giving an explicit control is certainly a better solution in a short term!

alygin:
> Another option is to always use the full format and hide irrelevant details on the IDE side, like in Java backtracks.

Actually, I implemented `RustConsoleFolding` long time ago just to explore this feature, but never published it :) I thought anyone who doesn't like long stack traces can configure folding in the Console Preferences.

I think I could revive that folding implementation and add an option to switch if off in case someone needs literally full output. Java has no such option (at least I'm unaware of one) but it provides the exact set of folding patterns that anyone can edit and it's not a part of public API (or is it?) so we can't use exactly the same approach.

alygin:
Here's how it looks like:

screen shot 2017-03-15 at 13 09 27

alygin:
> Another option is to always use the full format and hide irrelevant details on the IDE side, like in Java backtracks.

@matklad, did you mean what the screenshot shows? Or something else?

matklad:
>@matklad, did you mean what the screenshot shows? Or something else?

The screenshot. 

But now I think that maybe short backtracks is better solution than custom folding on the IDE side?

alygin:
> maybe short backtracks is better solution than custom folding on the IDE side?

In my opinion, after short backtraces make it to stable (hopefully, it won't take long), we can set them as the default option, still allow the user to switch to the Full mode but fold nothing ourselves. Those who really want folding in the Full mode can always configure it via Console Preferences, even without support on the plugin side.

PR #1067, opened 2017-03-14
error checking plugin

matomer:
only for code review for now


matklad:
Just a cursory review for now:

* I think code can get some enterprisy cleanups: let's not use abbreviated names (`ExportInfo`, `EI`) and remove commented code and extra debug output. 

* Do we really need that `DEBUG` logger? There's quite a bit of UI code there... Perhaps we can use the usual logging framework instead, like [here](https://github.com/intellij-rust/intellij-rust/blob/58a3d03838f3b81116a51b0c3c6ad489126a91cd/src/main/kotlin/org/rust/cargo/project/workspace/impl/CargoProjectWorkspaceServiceImpl.kt#L31)?

* Why do we want to "use this server to cache the cargo check result instead of running it for every file"? I think `cargo check` should be instantaneous, if there are no changes?

matomer:
1) EI will be removed (it was only for logging)
2) Same :)
3) IJ calls the annotate very often - even when there were no file changes (e.g. switching tabs). The server is there because I don't want to run a cargo check command each time a file needs annotation. It caches the last cargo check run and only updates it when it sees that a project file has been modified. This is even more important when you have multiple files in the project, since each one runs a `cargo check` that is a global command so they cause each other to block and slow down.

matomer:
Cleaned up the logging and changed jackson to gson. Please review again and tell me if there's more that needs to be done. Currently the only thing I have left is pulling the cargo exe path from IJ-rust (which I'll ask you about on gitter).

matklad:
>Currently the only thing I have left is pulling the cargo exe path from IJ-rust (which I'll ask you about on gitter).

See the usages of `org.rust.cargo.toolchain.Cargo` class in the plugin. To get an instance of `Cargo`, you need a `Toolchain` and a project directory. You can get toolchain using `Project.toolchain` extension, and a directory using `Module.cargoProjectRoot`.

src/main/kotlin/org/rust/ide/annotator/CargoCheckExtAnnotator.kt
```diff
@@ -0,0 +1,630 @@
...
+object IJDebug {
+    val defaultTitle = "CARGO-CHECK"
+
+    fun log(msg : String, title : String = defaultTitle) {
+        IJLogger.getInstance(title).debug(msg)
+    }
+
+}
+
+object CargoJson {
```
matklad: This object is **huge**. I think it is so because it tries to do two rather orthogonal things:

* parse json text into a typed Kotlin data structure,
* use that data structure to annotate the file.

Let's try to split into two parts. 

The first part, `CargoMessages`, should specify `data class`es and provide an ability to convert text output of cargo into Kotlin objects. It can live in a separate file and should be independent of the annotator (because it could be reused elsewhere). 

All the annotation logic should live in the annotator. 

So, the end result could look like this:

```Kotlin
for (line in cargoOutput) {
  val message = CargoMessage.parse(line)
  if (message.level == CargoMessageLevel.ERROR) {
    annotate(holder, message)
  }
}
```

src/main/kotlin/org/rust/ide/annotator/CargoCheckExtAnnotator.kt
```diff
@@ -0,0 +1,630 @@
...
+        //Used for both error and warning
+        abstract class ErrorBaseMessage(
+            val children: List<Message>,
+            val code: Code?,
+            override val message: String,
+            override val spans: List<Span>,
+            val level : String //Only "warning" and "error" allowed
+        ) : Message() {
+            init {
+                children.forEach { assert(it is NoteMessage || it is HelpMessage) }
```
matklad: It's preferable to use `check` instead of `assert`, because assertions are usually disabled.

src/main/kotlin/org/rust/ide/annotator/CargoCheckExtAnnotator.kt
```diff
@@ -0,0 +1,630 @@
...
+
+            //Check if this we don't need to do annotation for this error
+            //and can skip it.
+            private fun canSkipAnnotation() : Boolean {
+                if (code != null || children.isNotEmpty() || spans.isNotEmpty())
+                    return false
+
+                //Check the message itself - some messages don't need to be displayed as errors
+                val regexes = arrayOf(
+                    Regex("aborting due to \\d+ previous errors"),
```
matklad: There's a handy

```
"aborting due to \\d+ previous errors".toRegex()
```

extension :)

src/main/kotlin/org/rust/ide/annotator/CargoCheckExtAnnotator.kt
```diff
@@ -0,0 +1,630 @@
...
+        //file - The file to annotate
+        fun doAnnotate(holder: AnnotationHolder)  {
+            message.doAnnotate(holder, target)
+        }
+    }
+
+
+}
+
+data class TEAAnnotationInfo(val file: PsiFile, val editor: Editor)
```
matklad: It's better to avoid acronyms in names, unless the acronym is very popular, like HTML.

src/main/kotlin/org/rust/ide/annotator/CargoCheckExtAnnotator.kt
```diff
@@ -0,0 +1,630 @@
...
+            IJDebug.log("executing command `$cmd`")
+
+            val err : String
+            val out : String
+            val retVal : Int
+            var process : Process? = null
+
+            try {
+
+                process = ProcessBuilder(args).start()
```
matklad: Can we use `CapturingProcessHandler` instead? It already can split output by lines. 

Or perhaps we can just synchronously wait for all the output to be available?

src/main/kotlin/org/rust/ide/annotator/CargoCheckExtAnnotator.kt
```diff
@@ -0,0 +1,630 @@
...
+            //We check if there were changes to any project files and
+
+            ApplicationManager.getApplication().messageBus.connect().subscribe(VirtualFileManager.VFS_CHANGES,
+                object : BulkFileListener {
+                    override fun before(events: MutableList<out VFileEvent>) {
+                    }
+
+                    override fun after(events: MutableList<out VFileEvent>) {
+
+                        //fun printEv(l : MutableList<VFileEvent>) : String =events.map { it.file }.joinToString(",")
```
matklad: Some more debug comments should be removed.

matklad:
Looks better now!

Sorry for the amount of style and organization related comments, but it's much easier to review the code which is structured neatly :)

PR #1075, opened 2017-03-19
Fix a few small type related bugs

kumbayo:


kumbayo:
This is for correctness and will not improve the auto completion significantly as we currently have limited auto completion for primitive types.

matklad:
Thanks!

PR #1088, opened 2017-03-25
RES: Resolve to correct crate version

### LINKED ISSUE #1015: When resolving external crates, use dependency information
Rust has an awesome library/module split at the language level, which allows for linking in two versions of the same library. 

That is, `extern crate foo;` can resolve to different crates depending on which crate this declaration is in. 

Precise dependency DAG is available in `cargo metadata`, but we just lump all external crates to a global namespace: https://github.com/intellij-rust/intellij-rust/blob/723969c995e7991e1bc002454656f1cdb27edd2f/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt#L84

This causes a trouble if, for example, there are two hypers in the dependency graph, and we resolve to the wrong one and produce false positive error annotations because some functions have changed arity. 

@alygin you might be interested in this bug :)

alygin:
When multiple versions of a particular crate are used in the same project (transitively, through dependencies), resolve references to the correct version of the crate depending on what package the reference belongs to.

Fixes #1015

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -24,6 +24,7 @@ class CargoWorkspace private constructor(
         val source: String?,
         val origin: PackageOrigin
     ) {
+        val normName = name.replace('-', '_')
         val dependencies: MutableList<Package> = ArrayList()
         val libTarget: Target? get() = targets.find { it.isLib }
         val contentRoot: VirtualFile? get() = VirtualFileManager.getInstance().findFileByUrl(contentRootUrl)
@@ -34,6 +35,9 @@ class CargoWorkspace private constructor(
             targets.forEach { it.initPackage(this) }
             return this
         }
+
+        fun findCrateByName(normName: String): Target? =
+            if (this.normName == normName) libTarget else dependencies.findLibrary(normName)
     }
 
     class Target(
@@ -77,10 +81,7 @@ class CargoWorkspace private constructor(
 
     private val targetByCrateRootUrl = packages.flatMap { it.targets }.associateBy { it.crateRootUrl }
 
-    fun findCrateByName(normName: String, inPackage: Package? = null): Target? =
-        inPackage?.dependencies?.findLibrary(normName)
-            ?.let { return it }
-            ?: packages.findLibrary(normName)
+    fun findCrateByNameApproximately(normName: String): Target? = packages.findLibrary(normName)
 
     /**
      * If the [file] is a crate root, returns the corresponding [Target]
@@ -96,7 +97,7 @@ class CargoWorkspace private constructor(
 
     fun withStdlib(libs: List<StandardLibraryRoots.StdCrate>): CargoWorkspace {
         val stdlib = libs.map { crate ->
-            Package(
+            val pkg = Package(
                 contentRootUrl = crate.packageRootUrl,
                 name = crate.name,
                 version = "",
@@ -104,16 +105,26 @@ class CargoWorkspace private constructor(
                 source = null,
                 origin = PackageOrigin.STDLIB
             ).initTargets()
+            (crate.name to pkg)
+        }.toMap()
+
+        // Bind dependencies and collect roots
+        val roots = ArrayList<Package>()
+        libs.forEach { lib ->
+            val slib = stdlib[lib.name] ?: error("Std lib ${lib.name} not found")
+            val depPackages = lib.dependencies.mapNotNull { stdlib[it] }
+            slib.dependencies.addAll(depPackages)
+            if (lib.isRoot) {
+                roots.add(slib)
+            }
         }
-        packages.forEach { it.dependencies.addAll(stdlib) }
-        return CargoWorkspace(packages + stdlib)
+
+        roots.forEach { it.dependencies.addAll(roots) }
+        packages.forEach { it.dependencies.addAll(roots) }
+        return CargoWorkspace(packages + roots)
     }
 
     val hasStandardLibrary: Boolean get() = packages.any { it.origin == PackageOrigin.STDLIB }
-
-    fun Collection<CargoWorkspace.Package>.findLibrary(normName: String): CargoWorkspace.Target? =
-        mapNotNull { it.libTarget }
-            .find { it.normName == normName }
 
     companion object {
         fun deserialize(data: CleanCargoMetadata): CargoWorkspace {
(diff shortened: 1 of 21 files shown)


alygin:
If falls back to the current crate lookup mechanism (search by name among all the packages) when cannot find it in the dependencies list. So, I hope there won't be regression.

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -78,10 +77,10 @@ class CargoWorkspace private constructor(
 
     private val targetByCrateRootUrl = packages.flatMap { it.targets }.associateBy { it.crateRootUrl }
 
-    fun findCrateByName(normName: String): Target? =
-        packages
-            .mapNotNull { it.libTarget }
-            .find { it.normName == normName }
+    fun findCrateByName(normName: String, inPackage: Package? = null): Target? =
```
matklad: Can we make this last argument non-optional (or perhaps make this function a method of `Package`)? Current signature is (and always was) a bit of a lie, because if you don't know the package, you might get wrong results.

If we indeed sometime need to find crate by name "approximately", it's better to create a separate function with `approximately` or something in the name :)

matklad: Hm, and it may be the case that we need `Target` and not the `Package` here. `dev-dependencies` affect only tests and build script, but not the main target.

alygin: I totally agree on moving the method to `Package` and only leaving the "approximate" part in the workspace. But I have two concerns about the `Target`:

First, I don't see the benefits of having the method there. I understand that Rust itself tracks dependencies at the target level, but why should we do the same in IDE? If I'm not mistaken, one cannot have two different versions of the same crate in different targets, so it's safe to search in a common list of packages and separating them will only bring unnecessary complexity and potential problems. Or am I wrong? Is there an example where such separation is needed for correct resolving?

Second, at first glance, `cargo metadata` doesn't provide enough information to properly separate package dependencies among different targets. Resolve nodes have no information about targets, and values in the `packages.dependencies.target` field look sort of useless to me. But honestly, I didn't dig deep here yet.

alygin:
While testing the feature, I also found that we don't resolve some crates. The cases I studied were all about having optional dependencies that are only included for particular features. At first, I thought the reason was that we didn't add `--all-features` when run `cargo metadata`, but I see this flag in code, so it shouldn't be the case.

I'll try to figure out what's the cause of the problem, but maybe you already know the answer?

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
matklad: Hm, that's indeed the case that Cargo itself tracks deps on the package granularity, so it's totally OK to do the same.

There is a bug filed to support "binary only" dependencies in Cargo though, but it probably won't be fixed in the nearish future.

matklad:
>I'll try to figure out what's the cause of the problem, but maybe you already know the answer?

matklad:
No, I don't. Maybe features can be mutually exclusive?

alygin:
> Maybe features can be mutually exclusive?

No, it's not. One of such cases is the [unicase](https://github.com/seanmonstar/unicase) package (you need version 1.4.0 that comes with hyper) that optionally depends on `heapsize` that we don't resolve for some reason.

Another open case is the standard library. We don't have information about dependencies of its packages. Any ideas on how that could be implemented? Run additional `cargo metadata` inside every package from stdlib?

matklad:
>Another open case is the standard library. We don't have information about dependencies of its packages. Any ideas on how that could be implemented? Run additional cargo metadata inside every package from stdlib?

Ideally, Cargo should just put everything into metadata, but the relevant RFC is far from being implemented: https://github.com/rust-lang/rfcs/pull/1133

So I think we can just hard-code that? Presumably, there are quite a few stdlib crates, and their dependencies should not change much. I think it should be possible to run cargo separately on the workspace with standard library, but I expect that this is more work.

alygin:
There will be about 15 additional dependencies. I think in practise it's better to hardcode them and wait for the RFC implementation. If it'll take really long and we encounter some problems with resolving, we can get back to the idea of extracting metadata separately from each stdlib package.

alygin:
The problem with unresolved crates is that cargo doesn't propagate `--all-features` to dependencies. That is, if you add the dependency `foo` that optionally depends on `bar` which is bound to some feature, and you don't specify this feature when adding `foo`, you won't be able to jump to `bar` while exploring `foo` because it will be absent in the metadata.

Such behaviour looks reasonable for cargo, but leaves room for unresolvable crate references. The good part is that such crates can only be outside of our workspace. I don't think we can fix it at the moment.

alygin:
I did the refactoring we discussed above and switched to `Package.findCrateByName()` almost everywhere. I'd like to leave the "approximate" version of the method only for special cases and never fall back to it from `Package.findCrateByName()`. Unfortunately, there's a problem that prevents it now: [in this line](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/ext/RsCompositeElement.kt#L40), `virtualFile` is often `null` and it breaks autocompletion in some cases.

I can't find the reason of such nullability. The crate name completion test is broken now because of it, while very similar test for crate docs line marker works fine. Unfortunately, the problem is not only in tests. In real projects, `RsExternCrateItem`s contain valid `crateRoot.containingFile`s, but those containing files has no virtual files associated with them.

At the same time, the doc line markers are shown for the very same `RsExternCrateItem`s which means the virtual files are good when the marker collector works. It seems that the problem only occurs on autocompletion and somehow relates to virtual files lifecycle, but I don't know the platform internals good enough to fix it.

Any ideas on why that could happen and how to make it work? I pushed the changes so you could have a look at what was done.

matklad:
>I can't find the reason of such nullability.

This is explained in this commit message:
https://github.com/intellij-rust/intellij-rust/commit/f9a6e4aba9e4d7110b0ac08d026e937efe87bbbe. A ghastly issue indeed!

alygin:
Thanks! It works.

alygin:
To sum it up:

1. `Package.dependencies` contains references to particular versions of its dependencies.
2. Usual crate resolving now goes through `Package.findCrateByName()`, while `CargoWorkspace.findCrateByNameApproximately()` searches for crates globally by their names and is only used where we have no other options.
3. Stdlib contains not only root crates, but also most of their dependencies. So, I renamed `StandardLibraryRoots` to `StandardLibrary` to better fit its role.
4. Every package has the stdlib roots among its dependencies.

The known issues are: a) feature-specific transient dependencies for which `cargo metadata` doesn't provide information, b) some stdlib transitive dependencies specific for stdlib testing.

matklad:
Awesome!

PR #1099, opened 2017-03-27
ANN: Improvement on add turbofish quick fix.Support nested generics.

la10736:
Now addressing:

* nesting `parse<Option<Option<i32>>("41")`
* boundary `3 < parse<i32>("41") > 10`

But not yet:

* `Something<Gen>::func()`
* `func<Some<_>>()`
* `func<A, B>()`

*Note*: `create()` and `createFromText()` come from `RsPsiFactory` because I didn't find any other usable and public implementation. Thoughts are welcome.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFix.kt
@@ -94,7 +94,6 @@ class AddTurbofishFix : RsElementBaseIntentionAction<AddTurbofishFix.Context>() 
 
     private fun isCallExpression(expr: RsExpr) = expr is RsParenExpr || expr.firstChild is RsParenExpr
 
-    /* Maybe there' s a better way than cut and paste these methods from RsFactory */
     private inline fun <reified T : RsCompositeElement> create(project: Project, text: String): T? =
         createFromText(project, "fn main() { $text; }")
 
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFix.kt
@@ -3,27 +3,32 @@ package org.rust.ide.annotator.fixes
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFileFactory
 import org.rust.ide.intentions.RsElementBaseIntentionAction
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.RsElementTypes.*
 import org.rust.lang.core.psi.ext.operatorType
 import org.rust.lang.core.psi.ext.parentOfType
+import org.rust.lang.core.psi.RsElementTypes.*
+import org.rust.lang.RsFileType
+import org.rust.lang.core.psi.ext.RsCompositeElement
+import org.rust.lang.core.psi.ext.childOfType
+
 
 class AddTurbofishFix : RsElementBaseIntentionAction<AddTurbofishFix.Context>() {
     private val TURBOFISH = "::"
 
     override fun invoke(project: Project, editor: Editor, ctx: Context) {
-        val (matchExpr, caller, gen, more) = ctx
-        val callWithNamespace =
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFix.kt
```diff
@@ -3,55 +3,103 @@ package org.rust.ide.annotator.fixes
...
-        if (!isCallExpression(more)) {
-            return null
-        };
-        return Context(matchExpr, caller, gen, more)
     }
 
     private fun isCallExpression(expr: RsExpr) = expr is RsParenExpr || expr.firstChild is RsParenExpr
+
+    /* Maybe there' s a better way than cut and paste these methods from RsFactory */
+    private inline fun <reified T : RsCompositeElement> create(project: Project, text: String): T? =
```
la10736: Is there a public implementation for this?

matklad: No, currently `RsFactory` creates only valid elements. You can add a `createDummyFile(text: String)` method to it, but this particular fallible create belongs to this fix.

src/test/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFixTest.kt
```diff
@@ -48,14 +47,54 @@ class AddTurbofishFixTest : RsAnnotatorTestBase() {
...
+        }
+    }
+
+    fun `test should guess the correct boundary`() {
+        val cases = listOf(
+            Pair("", " >> 3"), Pair("", " > 3 > 5"), Pair("", " == call_something()"),
+            Pair("3 < ", ""), Pair("3 == ", ""), Pair("bye() == 3 < ", " > more() >> 2")
+        )
+
+        for ((before, after) in cases) {
```
matklad: I think this loop is a bit too unreadable. What do you think about creating a function that checks two expressions, and then just write a large number of cases explicitelly?

```Rust
fun doCheck(expr1: String, expr2: String) { ... }

fun `test should guess correct boundary` {
  doCheck("something<Option<i32>>(/*caret*/"42")>> 3", "something::<Option<i32>>(/*caret*/"42")>> 3")
  ...
}
```

It's a bit more repetition that way, but it is super obvious what exactly is expression before and after.

matklad:
Looks good to me!

I am a bit worried about all this complicated logic here, but I don't have better suggestions. I think the precise logic does not matter as long as we have a good number of test cases that pass :) 

Another possible alternative is text based. Find all `<` tokens to the left of the cursor, find the matching `>`, and then try to create the expression with text from the beginig of the original expression to the last `>` plus `()`. 

That is, if we look at the PSI for

```
xs.iter()
        .map(|i| i * 2)
        .collect<Result<Vec<_>, Option<_>>>();
```

we are correctly see the beginning of the expression, but we fail to parse anything past `Vec<_`. But we can find the expression's end by matching `<` with the corresponding `>`. Then we can try add `:: ... ()` and see if that parses. 

I have no idea if this would actually work, so let's the tests and `create` thing in this approach and merge it first.

src/test/kotlin/org/rust/ide/annotator/fixes/AddTurbofishFixTest.kt
la10736: Ok sounds good. Just a note: by use `before` and `after` the test intent become clear but find then in string make things too complicated...

I'll change it

la10736:
My idea was to not use text based search but psi list elements: text based make hard to find the boundary and stop clause. But now I see your point and I'll give it a try.

If you are OK I would like to take these cases out this PR and create a new one.

matklad:
>but psi list elements: text based make hard to find the boundary and stop clause. But now I see your point and I'll give it a try.

We need PSI to find the left boundary, and we need text to find the right boundary :) 

>If you are OK I would like to take these cases out this PR and create a new one.

OK!

la10736:
Is it ok now or I have missed something?

matklad:
@la10736 I have not yet had a chance to review this properly, will try do take a look tomorrow!

la10736:
No problem.... I'm fine. I'd just asked in the case I'd forgot something... :smile:

matklad:
LGTM! Thanks!

PR #1105, opened 2017-04-04
ANN: More lifetime error annotations

alygin:
E0107 (wrong number of lifetime parameters) and more cases of E0106 (missing lifetime) annotations. From #886


src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeParametersOwner.kt
```diff
@@ -0,0 +1,7 @@
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsTypeParameterList
+
+interface RsTypeParametersOwner : RsCompositeElement {
+    val typeParameterList: RsTypeParameterList?
```
matklad: I think this is the same as `RsGenericDeclaration`?

alygin: Oh, it is indeed.

matklad:
Great!

Issue #1097, opened 2017-03-27
Fold one line methods


matklad:
It would be nice if one liners like

```Rust
pub fn root(&self) -> Node {
    Node { id: self.root, file: self }
}
```

Were folded by default, as in Java:

![one-line-method](https://cloud.githubusercontent.com/assets/1711539/24377779/31170b9e-1349-11e7-8de5-91556a8e4b06.png)

alygin:
@matklad, closed?

matklad:
Yep!

miedzinski:
Is it possible to disable this?

matklad:
Yep!

![folding](https://cloud.githubusercontent.com/assets/1711539/24863934/5dea4d30-1e0b-11e7-8612-bb91b84c98d5.png)

The option at the very bottom.

miedzinski:
@matklad Thanks!

PR #1106, opened 2017-04-05
FOLD: Fold one liner functions

alygin:
Implements #1097 

I'd like to only fold one-liners by default when the flag Preferences -> Editor -> General -> Code Folding -> One-line methods is on, but don't know how to get its value. There's the `CodeFoldingSettings` class, but it doesn't provide access to the "Collapse by default" preferences block. Any hints on how that could be achieved?

### DIFF
--- a/src/main/kotlin/org/rust/ide/folding/RsCodeFoldingOptionsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/folding/RsCodeFoldingOptionsProvider.kt
@@ -1,0 +1,17 @@
+package org.rust.ide.folding
+
+import com.intellij.application.options.editor.CodeFoldingOptionsProvider
+import com.intellij.openapi.options.BeanConfigurable
+
+class RsCodeFoldingOptionsProvider :
+    BeanConfigurable<RsCodeFoldingSettings>(RsCodeFoldingSettings.instance),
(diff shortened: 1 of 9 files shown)


matklad:
I think this setting comes from `JavaCodeFoldingSettings`. Not sure if it is present in minor IDEs

alygin:
You're right, it's absent from CLion. This "General" misled me.

Is it possible to provide our own flag in this block? Or should it go to somewhere in Code Style -> Rust? I'm afraid not everyone likes the feature and it's better to make it customizable.

matklad:
There's `CodeFoldingOptionsProvider`. I think it's OK to add "fold Rust one-line methods". 

The ideal solution would probably reuse Java's options in IDEA and added our own thing in CLion, but that's probably impossible to implement without reflection/messing with IDE-specific configuration (there's `idea-only.xml`).

alygin:
> I think it's OK to add "fold Rust one-line methods".

Added

matklad:
Thanks!

matklad:
Tweaked logic a bit in https://github.com/intellij-rust/intellij-rust/commit/80f1dbc755afcd2bc52daf89ce45f76f4da00c3b so that one-line automatically unfolds once you've typed `\n`.

PR #1111, opened 2017-04-05
RES: Fix lifetime resolve in where clauses

alygin:
Fix lifetime resolve when it's declared in a `where for<...>` clause.


matklad:
Thanks!

It would be interesting to dig what these `for` clauses in `where` actually mean in the Rust language and how they interact with for in types. Perhaps it's just some think that should have been obsoleted by `for-in` types?

PR #1118, opened 2017-04-08
ANN: Wrong type arguments number annotations

alygin:
Annotates [E0243](https://doc.rust-lang.org/error-index.html#E0243)/[E0244](https://doc.rust-lang.org/error-index.html#E0244) errors &ndash; not enough / too many type arguments.
From #886


matklad:
:+1:

PR #1121, opened 2017-04-09
COMP: Complete trait methods implemented for struct or enum (Fixes #1120)

### LINKED ISSUE #1120: Complete trait methods implemented for struct
Minimal setup
```rust
trait SomeTrait {
    fn some_fn(&self);
}
struct SomeStruct {
    value: i32
}
impl SomeTrait for SomeStruct {
    fn some_fn(&self) {
        println!("Hello, world!");
    }
}
fn main() {
    let v = SomeStruct {value: 5i32};
    v.s/*caret*/
}
```
I expect to see "some_fn" completion option, but "No suggestions".

Look at [this line](https://github.com/intellij-rust/intellij-rust/blob/47ffbfa3afc2d2ce915e8370c786a33de3bdc8a0/src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt#L67). If I understand correctly, this line is needed to filter trait implementations for different type arguments (we may compare only [`RustStructType.typeArguments`](https://github.com/intellij-rust/intellij-rust/blob/340102c272a24b6da776469bce297e1983234cd3/src/main/kotlin/org/rust/lang/core/types/types/RustStructType.kt#L9) at that line). But actually we compare structs by type arguments [and psi element (`RustStructType.struct`)](https://github.com/intellij-rust/intellij-rust/blob/340102c272a24b6da776469bce297e1983234cd3/src/main/kotlin/org/rust/lang/core/types/types/RustStructType.kt#L8), which are not equals (by referense) while completion.

clip2net_170409024423

I think we just should compare RustStructType's by `item` property, which are canonicalized and really equal. Or compare only type arguments at that place.

vlad20012:
See #1120 for more info. I removed `RustStructType.struct` field and made equality by `item` (same with `RustEnumType`). I'm not sure about the implementation, but this PR contains useful tests that demonstrate the problem (and its fix).

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/types/RustEnumType.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/types/RustEnumType.kt
@@ -4,8 +4,8 @@ import com.intellij.codeInsight.completion.CompletionUtil
 import org.rust.lang.core.psi.RsEnumItem
 import org.rust.lang.core.types.RustType
 
-data class RustEnumType(
-    val enum: RsEnumItem,
+class RustEnumType(
+    enum: RsEnumItem,
     override val typeArguments: List<RustType> = emptyList()
 ) : RustStructOrEnumTypeBase {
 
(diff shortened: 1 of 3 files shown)


matklad:
Hm, that's a one tricky bug, thanks! 

Implementation-wise, I'd rather lean on Kotlin data classes for implementing `.equals` and `hashCode`. Perhaps we can change constructors to call `getOriginalOrSelf` before storing the field?

Issue #1114, opened 2017-04-06
Cannot borrow immutable local variable `None` as mutable


eugene-bulkin:
I have a function that takes a `&mut Option<T>`, and in cases where it doesn't need anything, I can pass a `&mut None` in. However, this now is noted as an error in IntelliJ, when it wasn't before (and it isn't an actual Rust error).

farodin91:
I think the checks thinks None is an variable. And this variable isn't mutable. 

I will look into it deeper. The checks need to be refactor decided between mutable and assignable.

farodin91:
Assignable = reassignable

matklad:
closed by https://github.com/intellij-rust/intellij-rust/pull/1122

PR #1122, opened 2017-04-10
ANN: Fix mutability annotation on enum variants

alygin:
Fixes the #1114 false positive. I'm not sure this should be the final solution, just want to provide some fix before the alpha release.

cc @farodin91


matklad:
LGTM! Perhaps the better overall strategy would be to flip similar logic and default to "no error" case.

alygin:
> Perhaps the better overall strategy would be to flip similar logic and default to "no error" case.

Totally agree. Unfortunately, just changing the default to "no error" in the current code breaks several tests, thus the "logic flipping" will require some additional effort.

PR #1123, opened 2017-04-10
tbn -> bfn

Keats:
tbn doesn't really mean anything, I didn't think enough when I added that


matklad:
Yep, `bfn` sounds better!

PR #1124, opened 2017-04-10
TY: infer struct or enum expr type parameters

vlad20012:
provides generic types inference for struct/enum expressions
```rust
struct SomeStruct<T1, T2> { a: T1, b: T2 }
fn main() {
    let x = SomeStruct { a: 5u16, b: 5u8 };
    (x.a, x.b)
  //^ (u16, u8)
}
```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/RustTypificationEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/RustTypificationEngine.kt
@@ -1,5 +1,6 @@
 package org.rust.lang.core.types
 
+import org.rust.ide.utils.isNullOrEmpty
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.*
 import org.rust.lang.core.psi.ext.*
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/types/RustTypificationEngine.kt
```diff
@@ -237,3 +240,70 @@ private fun RsImplItem.remapTypeParameters(
                 null
             }
         }.toMap()
+
+private fun inferStructTypeParameters(o: RsStructExpr, item: RsStructItem): RustType {
+    val baseType = item.type
+    if ((baseType as? RustStructOrEnumTypeBase)?.typeArguments.isNullOrEmpty())
+        return baseType
```
matklad: Let's add braces to this and similar ifs, to be consistent with other code? 

```Kotlin
if ((baseType as? RustStructOrEnumTypeBase)?.typeArguments.isNullOrEmpty()) {
    return baseType
}
```

If you want to avoid braces, and both the condition and the expression are short, you can put the return on the same line:

```Kotlin
if (condition) return foo
```

matklad:
bors: r+

Thanks!

PR #1132, opened 2017-04-10
TY: basic infer non-suffixed numeric types

vlad20012:
Now correctly infer types in cases like this
```rust
struct SomeStruct<T> { a: T, b: T }
fn main() {
    let x = SomeStruct { a: 0, b: 5u16 };
    x.a
    //^ u16
}
```


matklad:
bors r+

:+1:

PR #1135, opened 2017-04-11
Ignore 'isKindWeak' in type comparision

vlad20012:
Related to #1132


matklad:
bors r+

Would be nice to try to pin this down with some test though!

Issue #1022, opened 2017-02-17
Automatically insert `>` after `::<`


matklad:

la10736:
If there is no too much pressure about this feature I can take care of it.

matklad:
>If there is no too much pressure about this feature I can take care of it.

Sure! Although there's a change that it would be more complicated than it seams. Somewhat similar https://github.com/intellij-rust/intellij-rust/pull/923 turned out to be pretty complicated.

matklad:
closed by https://github.com/intellij-rust/intellij-rust/pull/1142

PR #1142, opened 2017-04-17
TYPE: automatically insert '>' after '::<'

Undin:
https://github.com/intellij-rust/intellij-rust/issues/1022
This case is similar with inserting `>` after `.<` in Java so my implementation is based on
https://github.com/JetBrains/intellij-community/blob/f8093cba578089b13314197faee7b3436ab831f5/java/java-impl/src/com/intellij/codeInsight/editorActions/JavaTypedHandler.java#L290

### DIFF
--- a/src/main/kotlin/org/rust/ide/typing/RsTypingHandler.kt
+++ b/src/main/kotlin/org/rust/ide/typing/RsTypingHandler.kt
@@ -1,0 +1,53 @@
+package org.rust.ide.typing
+
+import com.intellij.codeInsight.CodeInsightSettings
+import com.intellij.codeInsight.editorActions.TypedHandlerDelegate
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.ex.EditorEx
+import com.intellij.openapi.fileTypes.FileType
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsElementTypes
+import org.rust.lang.core.psi.RsFile
+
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/typing/RsTypingHandler.kt
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        return Result.CONTINUE
+    }
+
+    override fun charTyped(c: Char, project: Project, editor: Editor, file: PsiFile): Result {
+        if (file !is RsFile) {
+            return Result.CONTINUE
+        }
+        if (rsLTTyped && CodeInsightSettings.getInstance().AUTOINSERT_PAIR_BRACKET) {
```
alygin: Isn't this `AUTOINSERT_PAIR_BRACKET` part of the condition redundant here? We cheack `rsLTTyped` that can only be `true` if `AUTOINSERT_PAIR_BRACKET` is `true`.

Undin: Yes, you are right

Undin:
@alygin I've fixed your comment. Please, check again when you have time

alygin:
Looks good to me
r+

matklad:
@alygin 

You need to say bors before erh plus :)

alygin:
Yeah, I know, but I wasn't going to merge it. I thought you would also have a look at the changes before merging, since I had never worked with this part of code and can miss something. Sorry for the misleading comment :)

matklad:
bors r=alygin

Thanks @Undin !

PR #1143, opened 2017-04-18
INSP: Inspection for mutable

farodin91:


alygin:
That would be great to have such inspection, but I'm afraid it's impossible to have a reliable implementation of it at the time. Partly because of to the imperfection of the current resolve mechanism (it can just miss some usages), partly because of the task is not so simple as it might seem even with the absolutely accurate resolving. For instance, here's the trivial case that emits a false positive warning:

```rust
struct Foo { foo: bool }
fn foo(b: &mut bool) {}
fn bar() {
    let mut f = Foo { foo: true };
     // ^^^^^ Highlighted incorrectly
    foo(&mut f.foo);
}
```

I believe running this inspection on any even mid-size real project will reveal great deal of various false positive cases.

alygin:
Here's even simpler false positive:

```rust
fn foo() {
    let mut v = Vec::new();
    v.push(42);
}
```

farodin91:
I'll start with disabling this by default and mark it as experimental. But first I'll extend my implementation to reduce some false positives and support function calls.

farodin91:
@alygin Could you review my changes? What do you think of marking it as experimental?

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,62 @@
...
+    override fun getDisplayName() = "No mutable required"
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitPatBinding(o: RsPatBinding) {
+                if (!o.isMut) return
+                val block = o.parentOfType<RsBlock>() ?: o.parentOfType<RsFunction>() ?: return
+                val occurrences = ReferencesSearch.search(o, LocalSearchScope(block))
+                    .map { it.element.parent }
+                var isMutableNeeded = false
+                if (!occurrences.isEmpty()) {
```
alygin: This is a redundant check, the following loop will handle the case.

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,62 @@
...
+                    loop@ for (occurrence in occurrences) {
+                        val parent = occurrence.parent
```
alygin: This, actually, only handles very simple cases. Just put the binding inside a tuple and the whole analysis breaks:

```rust
fn foo(s: &mut (u8, u8)) {}

fn main() {
    let mut a = 1u8;
    let mut b = 2u8;
    bar(&mut (a, b));
}
```

And Rust allows much more convoluted code structures.

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,62 @@
...
+                        when (parent) {
+                            is RsUnaryExpr -> {
+                                if (parent.mut != null) {
```
alygin: Unnecessary nested blocks here and below: `is RsUnaryExpr -> if (...) {` will work. And overall the code is hard to analyse because of many nested levels. It would be useful to break it into small and focused functions.

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.rust.lang.core.resolve.ref.RsMethodCallReferenceImpl
+
+class RsVariableMutableInspection : RsLocalInspectionTool() {
+    override fun getDisplayName() = "No mutable required"
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitPatBinding(o: RsPatBinding) {
+                if (!o.isMut) return
+                val block = o.parentOfType<RsBlock>() ?: o.parentOfType<RsFunction>() ?: return
+                val occurrences = ReferencesSearch.search(o, LocalSearchScope(block))
```
alygin: `occurrences` declaration is redundant here. This could be:

```rust
ReferenceSearch.search(...)
    .asSequence()
    .map { ... }
    .filter { ... }
    ...
```

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+                return false
+            }
+            is RsMethodCallExpr -> {
+                val ref = parent.reference.resolve() as? RsFunction ?: return false
+                val self = ref.selfParameter ?: return false
+                if (self.isMut) {
+                    return false
+                }
+            }
+            is RsTupleExpr,
```
alygin: I don't remember if this case was in your previous version, but actually, just checking for `RsTupleExpr` at a particular level doesn't change much. You need to traverse ancestors of your occurrence up while they are tuples and check mutability at every level. As far as I understand, the following simple code will emit false positive:

```rust
fn foo(a: &mut (u8)) {}
fn main() {
    let mut a = 10;
    foo(&mut (a));
}
```

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
alygin: `occurrences` declaration is redundant here. And the double negative logic (`filterNot { somethingNot() }`) could also be flipped to make the code shorter, simpler to read and more idiomatic):

```rust
if (ReferenceSearch.search(...)
        .asSequence()
        .map { ... }
        .any { checkOccurrenceNeedsMutable(it) }) return
```

Here, `checkOccurrenceNeedsMutable()` should default to `true`. It'll remove false positives in cases you don't handle yet.

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+                val expr = parent.parent as? RsUnaryExpr ?: return false
+                if (handleUnaryExpr(expr)) {
+                    return false
+                }
+            }
+        }
+        return true
+    }
+
+    fun handleUnaryExpr(expr: RsUnaryExpr) = expr.mut != null
```
alygin: Could also be more idiomatic:

```rust
private val RsUnaryExpr.isMutable: Boolean get() = expr.mut != null
```

alygin:
Here's one more simple case of false positive you might want to address:

```rust
fn main() {
    let mut a = 10;
    let mut foo = || a = 20;
     // ^^^^^^^ Should not be highlighted
    foo();
}
```

alygin:
Great! That looks much more immune to false positives.

I guess there're only a couple of common cases left to cover so we could say that false positives are rare enough for the inspection to be of use at least in simple code:

1. `ref mut`s in match patterns:

```rust
fn main() {
    let mut a = Some(20);
    if let Some(ref mut b) = a { *b = 10; }
             // ^^^^^^^^^ Should not be highlighted

    match a {
        Some(ref mut b) => *b = 10,
          // ^^^^^^^^^ Should not be highlighted
        None => {},
    }
}
```

2. Macros:

```rust
fn foo(a: &mut u32) { }
macro_rules! call_foo { ( $name:ident ) => { foo(&mut $name); }; }
fn main() {
    let mut w = 10;
    call_foo!(w);
}
```

Macros are dark territory at the time, you won't even get occurrences within macro calls. I think the only safe thing to do here is don't emit warnings if the block under analysis contains any macro call. There're only several macros from the stdlib that the plugin recognizes, they can be added as exceptions (maybe later).

mkaput:
I think you can lookup macro for identifiers matching tested variable as macro hygiene prevents from silent access to outer scope from macros

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,49 @@
...
+    override fun getDisplayName() = "No mutable required"
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitPatBinding(o: RsPatBinding) {
+                if (!o.isMut) return
+                val block = o.parentOfType<RsBlock>() ?: o.parentOfType<RsFunction>() ?: return
+                if (ReferencesSearch.search(o, LocalSearchScope(block))
+                    .asSequence()
+                    .any { checkOccurrenceNeedMutable(it.element.parent) }) return
+                holder.registerProblem(o, "No mutable required for ${o.identifier.text}", *arrayOf(RemoveMutableFix(o)))
```
alygin: Let's also change the warning text to better match what Rust uses: "Variable `foo` does not need to be mutable".

farodin91:
Any tip to find a macro invocation in block?

alygin:
The most straightforward way is to add the following line to the visitor:

```rust
if (block.childOfType<RsMacroExpr>() != null) return
```

But that's too wide-ranging. A more accurate version that does what @mkaput said will include recursive children traversal along with checking their arguments to find if they contain an identifier `o.identifier.text` and analysing their positions within the block to handle rebindings. But I think we can cope with these cases later.

farodin91:
@alygin I check a bit more detail than only check for existent.

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.rust.lang.core.resolve.ref.RsMethodCallReferenceImpl
+
+class RsVariableMutableInspection : RsLocalInspectionTool() {
+    override fun getDisplayName() = "No mutable required"
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitPatBinding(o: RsPatBinding) {
+                if (!o.isMut) return
+                val block = o.parentOfType<RsBlock>() ?: o.parentOfType<RsFunction>() ?: return
+                val macro = block.descendantsOfType<RsMacroExpr>()
```
alygin: There could be a small optimization: you don't need to check macros if you find a mutable usage. So, this call to `descendantsOfType` (which builds the whole collection of macro calls) should go after the occurrences check to skip it sometimes.

It also can be used the same way as `checkOccurence()` instead of introducing a local variable:

```kotlin
if (block.descendantsOfType<RsMacroExpr>().any { ... }) return
```

src/main/resources/META-INF/plugin.xml
```diff
@@ -340,6 +340,11 @@
                          enabledByDefault="false" level="WARNING"
                          implementationClass="org.rust.ide.inspections.RsSimplifyBooleanExpressionInspection"/>
 
+        <localInspection language="Rust" groupName="Rust"
+                         displayName="[Experimental] No mutable required"
```
alygin: Let's don't put "[Experimental]" here. I think the caveat in the description is enough.

src/main/kotlin/org/rust/ide/inspections/RsVariableMutableInspection.kt
```diff
@@ -0,0 +1,53 @@
...
+                if (macro.any { checkExprPosition(o, it) }) return
+                if (ReferencesSearch.search(o, LocalSearchScope(block))
+                    .asSequence()
+                    .any { checkOccurrenceNeedMutable(it.element.parent) }) return
+                holder.registerProblem(o, "Variable ${o.identifier.text} does not need to be mutable", *arrayOf(RemoveMutableFix(o)))
```
alygin: There should be backquotes around the variable id in the message. Markdown formatting in my comment just hid them :)

alygin:
Could you, please, perform the "Optimize imports" command on both the inspection and fix files to remove unused imports? And then, I think, we could merge.

farodin91:
Done

alygin:
Thanks!

bors r+

Issue #1137, opened 2017-04-13
"Find Usages" doesn't work in nested packages


alygin:
While working on one of the error annotations I stumbled upon a problem with the "Find Usages" command. It finds nothing when performed in a nested package. Here's a simple project that reproduces this problem: [alygin/usage](https://github.com/alygin/usage). [This simple case](https://github.com/alygin/usage/blob/243427bb263ae56b6c78dca08024c22b3832eb43/nested/src/lib.rs) doesn't work in the nested package.

I noticed that it stops working as soon as you add the dependency block to the root `Cargo.toml`. That is, the command works as expected when [this block](https://github.com/alygin/usage/blob/243427bb263ae56b6c78dca08024c22b3832eb43/Cargo.toml#L8) is removed.

I tried to find the way to fix it, but there's not much code in `RsFindUsagesProvider` and it seems that the platform does all the magic itself. Any hints on what could be the reason of such behaviour and how to fix it?

matklad:
I have a vague idea that we add path dependencies as library roots (which we probably shouldn't do, and we definitely shouldn't do for workspace members), so these usages are filtered out by the default search scope ("project", and not "project and libraries").

alygin:
You're right, the problem can be fixed by setting `origin = WORKSPACE` on the `nested` package. But we don't do that now because cargo doesn't add this package to the `workspace_members` list since there's no `[workspace]` block in the root `Cargo.toml` that contains `nested` as a member.

So, on the one hand, it's totally correct on the cargo side. On the other hand, many projects with nested packages don't use explicit workspaces, but it would be nice to handle them properly.

We could require explicit workspaces for such project layouts to work properly (in accordance with cargo) or treat all the packages that are located within the project directory as workspace members. What do you think?

matklad:
>We could require explicit workspaces for such project layouts to work properly (in accordance with cargo) or treat all the packages that are located within the project directory as workspace members. What do you think?

I think we shouldn't treat a package as a library if it is an explicit member of the workspace regardless of the location, or if it is a path dependency and resides under module's content root .

alygin:
> we shouldn't treat a package as a library if it is an explicit member of the workspace regardless of the location

It already works

> or if it is a path dependency and resides under module's content root

That's a thing to implement. I'll to that.

PR #1145, opened 2017-04-19
"Unknown crate" inspection, "Unresolved reference" updated

Litarvan:
There is now a "Unknown crate" inspection, that detects unknown crate of "extern crate" expressions.
The "Unresolved reference" is now enabled by default, but as warning. It can also detects bad method references now, and more expressions.


alygin:
"Unknown crate" would ideally be an error annotation, not an inspection, since it's an official rustc error [E0463](https://doc.rust-lang.org/error-index.html#E0463). But the problem here is that it's susceptible to false positives due to inaccuracy of crates resolve at the moment. You can find more information about problem cases in the discussion in this [pull-request](https://github.com/intellij-rust/intellij-rust/pull/1088).

Good news is that false positives should only occur in dependencies, not in our own project. Unfortunately, we sometimes can't tell for sure which package is a dependency and which is a member of our project ([here're some details](https://github.com/intellij-rust/intellij-rust/issues/1137)).

So, my opinion is that a reasonable approach would be to wait until we close #1137 (it shouldn't take long, we just need to agree on the way to cope with it) and then move this inspection to `RsErrorAnnotator` but limit its scope by packages with `origin = WORKSPACE`. What do you think?

alygin:
> So, my opinion is that a reasonable approach would be to wait until we close #1137 (it shouldn't take long, we just need to agree on the way to cope with it) and then move this inspection to RsErrorAnnotator but limit its scope by packages with origin = WORKSPACE.

Oh, actually, we don't need to wait for closing #1137 if we limit the scope of the annotation by workspace members. The worst scenario will be potential false negatives in projects with nested packages, which is not so bad.

alygin:
And it would also be great to add some tests. We have a [special test project descriptor](https://github.com/intellij-rust/intellij-rust/blob/f1f86f04231f6deeb7006fab1352de6e389b958b/src/test/kotlin/org/rust/lang/RsTestBase.kt#L264) with external dependencies. You can find an example of its usage in `RsExternCrateCompletionTest`.

Litarvan:
Ok i'll try to edit what you said :)

Litarvan:
Ok I did it but, what do you mean by "setting origine = WORKSPACE" ?

Litarvan:
Ok so I don't know why but RANDOMLY the test fails. Also i didn't managed to put the alias crate test in the same class of my tests, because in fact, only the first test executed works, every other will drop an unknown crate error. (The random fail is probably linked to that).

Litarvan:
I deleted the "valid" crate test, because some other test do nearly the same, so the bug that i mentionned can't appear now.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -457,6 +458,11 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         holder.createErrorAnnotation(ret, "`return;` in a function whose return type is not `()` [E0069]")
     }
 
+    private fun checkExternCrate(holder: AnnotationHolder, el: RsExternCrateItem) {
+        if (el.reference.resolve() != null) return
```
alygin: >  what do you mean by "setting origin = WORKSPACE" ?

I mean adding an additional check here:

```rust
if (el.containingCargoPackage?.origin != PackageOrigin.WORKSPACE) return
```

It will limit the scope of the check by our own project and prevent it from running within dependencies because we can't fully rely on correct resolving there as I mentioned earlier.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -457,6 +458,11 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
+        holder.createErrorAnnotation(el.textRange, "Unknown crate '" + el.identifier.text + "' [E0463]")
```
alygin: Let's change this message to match what rustc emits: 

```
Can't find crate for `${el.identifier.text}` [E0463]
```

as we do in all such annotations.

alygin:
> I don't know why but RANDOMLY the test fails

I'll have a look at the tests a bit later when time allows.

Litarvan:
Done !

alygin:
There's one more problem we need to solve to make this annotation safe to use. Rust's stdlib has several root libraries that we now plug as dependencies [here](https://github.com/intellij-rust/intellij-rust/blob/34f01038f712525309b220c66785b0999df1e2aa/src/main/kotlin/org/rust/cargo/util/RustCrateUtil.kt#L15).

For instance, you can have the following valid code:

```rust
#![feature(rand)]
extern crate rand;
```

that will produce an error annotation because `rand` is declared as an stdlib dependency, not a root library.

To fix it, all the libraries in `AutoInjectCrates` should have `isRoot = true` except for `build_helper` and four `rustc_*` libraries.

alygin:
> To fix it, all the libraries in AutoInjectCrates should have isRoot = true

And that will lead us to yet another problem with multi-resolve :)  If someone explicitly adds `rand` as a dependency to the `Cargo.toml`, we'll have two `rand`s and `resolve()` will return `null`. So, we need to change `.resolve() != null` to `.multiResolve().isNotEmpty()`.

Litarvan:
The tests fail because of the isRoot = true in the AutoInjectedCrates :/
I don't really know what to do

alygin:
Hm, it looks like marking stdlib's `rand` as a root brings unnecessary ambiguity to resolving because it exists as both an external crate and stdlib part. The tests could be fixed by choosing another crate (though I'm not sure at the moment), but that would be just masking the problem. I'll have a look at this case and try to fix the problem in a separated PR. Then we'll be able to merge this one.

alygin:
@Litarvan, when #1223 is merged, you can rollback the `isRoot=true` changes and rebase. Tests should be ok then.

Litarvan:
Done :)

alygin:
Thanks! It turned out to be harder than expected, but revealed and helped to solve several issues with resolving!

bors r+

PR #1156, opened 2017-04-22
Filter methods by trait bounds

vlad20012:


src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
```diff
@@ -47,8 +43,12 @@ class RsImplIndex : AbstractStubIndex<RustTypeFingerprint, RsImplItem>() {
...
             ).asSequence()
             return when {
-                else -> elements.filter {
-                    canUnify(target, it.typeReference?.type ?: RustUnknownType)
+                else -> elements.filter { impl ->
+                    val ty = impl.typeReference?.type
+                    // Addition class check is a temporal solution to filter impls for type parameter
+                    // with the same name
+                    // struct S; impl<S: Tr1> Tr2 for S {}
+                    ty != null && ty.javaClass == target.javaClass && ty.isAssignableFrom(target, project)
```
matklad: That's unlikely to bite us in practice, so I suggest to remove this special case and instead write a test that verifies current  wrong behavior with a `FIXME` mark. 

Ideally, purely generic impls of this form should be handled separately. We should collect them in a separate index (maybe keyed by the trait) and then check if they are implemented for particular type.

src/main/grammars/RustParser.bnf
```diff
@@ -311,12 +311,29 @@ TypeParameter ::= OuterAttr* identifier TypeParamBounds? [ '=' TypeReference ] {
...
+  extends = "org.rust.lang.core.psi.ext.RsStubbedElementImpl<?>"
+  stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
+  elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
+}
+
+WherePred ::= Lifetime LifetimeParamBounds | ForLifetimes? TypeReference TypeParamBounds {
+  extends = "org.rust.lang.core.psi.ext.RsStubbedElementImpl<?>"
+  stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
+  elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
+}
```
matklad: :+1:

src/main/kotlin/org/rust/lang/core/types/RustType.kt
```diff
@@ -19,6 +19,13 @@ interface RustType {
         RsImplIndex.findMethodsFor(this, project)
 
     /**
+     * Checks if `other` type may be represented as this type.
+     *
+     * Note that `t1.isAssignableFrom(t2)` is not the same as `t2.isAssignableFrom(t1)`.
+     */
+    fun isAssignableFrom(other: RustType, project: Project): Boolean
```
matklad: This is not the best name. In Java, `isAssignableFrom` is about subtyping, but there's no subtypnig in Rust (except for the lifetimes). I suggest `canUnifyWith` instead: a type parameter `T` can unify with some concrete type `S` iff `S` satisfies all the bounds imposed on `T`.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
vlad20012: This would not be a problem, but the code given in the comment leads to a StackOverflowError.

matklad:
bors r+

Thanks!

PR #1157, opened 2017-04-22
CARGO: Treat nested packages as workspace members

### LINKED ISSUE #1137: "Find Usages" doesn't work in nested packages

alygin:
Makes all the packages that reside within our project treated as workspace members even if they're not listed by cargo in the `workspace_members` block.

Since cargo doesn't explicitly provide package paths, I extract them from the package `id`s and check if manifests of other packages sit somewhere within them.

Fixes #1137

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -135,12 +135,17 @@ class CargoWorkspace private constructor(
             // handle cycles here.
 
             // Figure out packages origins:
-            // - if a package is a workspace member, it's WORKSPACE
+            // - if a package is a workspace member, or if it resides inside a workspace member directory, it's WORKSPACE
             // - if a package is a direct dependency of a workspace member, it's DEPENDENCY
             // - otherwise, it's TRANSITIVE_DEPENDENCY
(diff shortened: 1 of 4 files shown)


matklad:
>Since cargo doesn't explicitly provide package paths, I extract them from the package ids and check if manifests of other packages sit somewhere within them.

I think `manifest_path` is what you are looking for. PackageIds are intended to be opaque identifiers.

matklad:
bors r+

Thanks!

PR #1158, opened 2017-04-22
Angle brace typed and backspace handlers

Undin:
Support of angle brace completion and removing paired brace.

Completion/removing paired brace cases:
* after `::`
* after any identifier if we are sure that it is name of generic entity: function, struct, enum, trait or type alias. 
* after identifier if it looks like type name (implementation similar with https://github.com/JetBrains/intellij-community/blob/cb371763404452f009f8ff7edba84e4881959882/java/java-impl/src/com/intellij/codeInsight/editorActions/JavaTypedHandler.java#L299)

Angle brace balance check is similar with https://github.com/JetBrains/intellij-community/blob/cb371763404452f009f8ff7edba84e4881959882/java/java-impl/src/com/intellij/codeInsight/editorActions/JavaTypedHandler.java#L246


src/main/kotlin/org/rust/ide/typing/RsAngleBraceHandlers.kt
```diff
@@ -0,0 +1,136 @@
...
+import com.intellij.openapi.editor.highlighter.HighlighterIterator
+import com.intellij.openapi.fileTypes.FileType
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsElementTypes
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.tokenSetOf
+
+private val GENERIC_NAMED_ENTITY_KEYWORDS = tokenSetOf(RsElementTypes.FN, RsElementTypes.STRUCT,
+    RsElementTypes.ENUM, RsElementTypes.TRAIT, RsElementTypes.TYPE_KW)
```
matklad: For `ElementType`s, we usually prefer unqualified form: `tokenSetOf(FN, STRUCT, ...)`.

src/main/kotlin/org/rust/ide/typing/RsAngleBraceHandlers.kt
```diff
@@ -0,0 +1,136 @@
...
+    RsElementTypes.SEMICOLON)
+
+class RsAngleBraceTypedHandler : TypedHandlerDelegate() {
+
+    private var rsLTTyped = false
+
+    override fun beforeCharTyped(c: Char, project: Project, editor: Editor, file: PsiFile, fileType: FileType): Result {
+        if (file !is RsFile) {
+            return Result.CONTINUE
+        }
```
matklad: It's Ok to write simple  "bail out" ifs on one line: `if (file !is RsFile) return Result.CONTINUE`.

matklad:
bors r+

Sweet!

matklad:
@Undin  there's one more possible improvement: when typing `(` followed by `)`, the result is `()/*caret*/`. With angle braces, we currently get `<>>`. I think it should be possible to use the same logic as with `()`. I have not looked into this yet, so I am not sure which extension point is responsible for this. Typing handler perhaps?

PR #1159, opened 2017-04-22
add .DS_Store to .gitignore

Undin:


mkaput:
I think this rule should be in your global gitignore file :)

Undin:
Oh, it's cool! I didn't know about global gitignore. Thanks :)

But I'm still thinking that it's the ideal solution for one developer. But it's simpler to add one line in .gitignore than each developer with mac adds this line into its own global gitignore.
Maybe it's not the best example but IDEA's .gitignore contains `.DS_Store` line too https://github.com/JetBrains/intellij-community/blob/10df87d7a9840e5901d4901ac4fff7ba035501c2/.gitignore#L8

matklad:
Yeah, in theory this is something you should put in your global .gitignore (the total number of repositories is much greater then the total number of developers, hence it's more efficient to store this once per developer), but in practice this is a nice little improvement for mac users, so thanks a lot and

bors r+

:)

PR #1160, opened 2017-04-23
TY: infer return type of generic function

vlad20012:
```rust
fn f<T>(t: T) -> T {
    return t;
}
fn main() {
    let a = f(0i32);
    a
  //^ i32
}
```
It also works for static methods. Yes, Box::new() now return a correct type.


src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -151,6 +151,45 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
...
+            a
+          //^ (u8, u16)
+        }
+    """)
+
+    fun testStaticMethod() = testExpr("""
+        struct S<T> { value: T }
+        impl<T> S<T> {
+            fn new(t: T) -> S<T> {
+                return S{value: t};
```
matklad: Missing space after `S` :)

src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -151,6 +151,45 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
         }
     """)
 
+    fun testGenericFunction() = testExpr("""
+        fn f<T>(t: T) -> T {
+            return t;
+        }
```
matklad: I usually prefer to keep test fixtures maximally compact, so this I would write as `fn f<T>(t: T) -> T { t }`.

Not 100% sure that this is a good idea, but I like the increased density in this case.

matklad:
Thanks!

bors r+

PR #1169, opened 2017-04-26
ANN: Fix a false postive case of mutable annotator

farodin91:


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -60,12 +60,23 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
-        val fn = o.reference.resolve() as? RsFunction ?: return
+    private fun checkMethodForNeededMutable(o: RsMethodCallExpr, fn: RsFunction): Boolean {
         if (!o.expr.isMutable() &&
             fn.selfParameter != null &&
             fn.selfParameter?.isMut ?: false &&
             fn.selfParameter?.isRef ?: false) {
+            val typeRef = o.parentOfType<RsImplItem>()?.typeReference as? RsRefLikeType ?: return true
+            if (typeRef.isMut) return false
+
+            return true
```
mkaput:
```kotlin
return !typeRef.isMut
```

farodin91: Resolved

PR #1177, opened 2017-04-27
RES: Support autoderef using language attribute

farodin91:
I uses an idea by @kumbayo. Is it okay?

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
@@ -41,16 +41,18 @@ class RsImplIndex : AbstractStubIndex<RustTypeFingerprint, RsImplItem>() {
                 .filter { !it.isAssocFn }
 
         fun findMethodsAndAssociatedFunctionsFor(target: RustType, project: Project): Collection<RsFunction> =
-            findImplsFor(target, project)
+            findImplsFor(target, project, true)
                 .flatMap { it.allMethodsAndAssocFunctions }
 
         fun findImplsFor(
             target: RustType,
             project: Project,
+            include_deref: Boolean = false,
             deref_hierarchic: Collection<RustType> = emptyList()
         ): Collection<RsImplItem> =
             project.implsCache
                 .getOrPut(target) { doFindImplsFor(target, project, deref_hierarchic) }
+                .filter { include_deref || it.typeReference?.type?.canUnifyWith(target, project) ?: false }
 
         private fun doFindImplsFor(
             target: RustType,
@@ -75,7 +77,7 @@ class RsImplIndex : AbstractStubIndex<RustTypeFingerprint, RsImplItem>() {
             }
             val derefTargetType = collection.find(RsImplItem::isDerefTrait)?.targetType
             val derefImpls = if (derefTargetType != null && !deref_hierarchic.any { it == derefTargetType })
-                findImplsFor(derefTargetType, project, deref_hierarchic + listOf(derefTargetType))
+                findImplsFor(derefTargetType, project, true, deref_hierarchic + listOf(derefTargetType))
             else
                 emptyList()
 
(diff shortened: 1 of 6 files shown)


matklad:
Yep, the implementation looks right! 

I think we would need to be more precise one day about autoderef behavior (current `stripAnyRefs` thing is an adhoc hack), but this can be handled later, as long as we have necessary tests in place!

But one thing we need to check is that we don't enter an infinite loop if A derefs to B and B derefs to A. Could you add a test for this as well?

farodin91:
Will also look for lang attributes, for better correlations. I will also support `*`.

@matklad Any idea to fix the bug access tree?

matklad:
This line causes the exception: `it.traitRef?.resolveToTrait?.identifier?.text`.

When you access the text of any node, a compact [Stub based](http://www.jetbrains.org/intellij/sdk/docs/basics/indexing_and_psi_stubs/stub_indexes.html) is switched to a more more slow PSI-based one. This test checks that resolve operates purely on stubs. 

You can see what info is available via stubs in `StubImplementations.kt`. For example, the trait's name property is stubified.

matklad:
You can see a recent PR (https://github.com/intellij-rust/intellij-rust/pull/1156) for an example of how to add more information to stubs, should you need it.

vlad20012:
I found a better solution that use "lang" attributes in this [branch](https://github.com/kumbayo/intellij-rust/commits/autocompletion_primitive_and_deref) and [commit](https://github.com/kumbayo/intellij-rust/commit/47f462777205b3f01327289ef4e6092fd00d7c8f)

farodin91:
@vlad20012 I know started to implement this and i will move lang attributes to stubs.

farodin91:
@matklad It should solve all current issues? 

My intention was to support `Vec::new().iter().next()` and so on? For this i need to add a default marco. This should be in a different PR.

farodin91:
@matklad Any tip to clean my local history?

farodin91:
I want also to extend the popup to show this is a autoderef function. Is this wanted and how could i implement it?

farodin91:
@matklad What do you think here?
> I think we would need to be more precise one day about autoderef behavior (current stripAnyRefs thing is an adhoc hack), but this can be handled later, as long as we have necessary tests in place!

vlad20012:
I think it's desirable to add several tests independent of the std to `RsTypeAwareResolveTest`

vlad20012:
And tests for multiple deref (A to B & B to C)
And test for recursive deref (A to B & B to A)

farodin91:
I will also add support for `deref_mut`.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
```diff
@@ -56,6 +72,14 @@ class RsImplIndex : AbstractStubIndex<RustTypeFingerprint, RsImplItem>() {
                 // struct S; impl<S: Tr1> Tr2 for S {}
                 ty != null && ty.javaClass == target.javaClass && ty.canUnifyWith(target, project)
             }
+            val traitDeref = collection.find { it.traitRef?.resolveToTrait?.langAttribute == "deref" }
```
matklad: Yeah, the `langItems` is the right approach here! 

Let's perhaps create a `src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt` file and move the lang-items related logic there? We have a ton of similar use case already, and it's high time we start consolidate it. 

For know, I think the file needs just one `fun isDerefTrait(trait: RsTraitItem): Boolean` one-liner.

matklad:
>I will also add support for deref_mut.

We can make it in a separate PR I think

farodin91:
> We can make it in a separate PR I think
That should be a easy one.

vlad20012:
I found a little problem. If you resolve all impls for deref `Target` type as impls for this type, it leads to false hits to trait bound. E.g. this test fails:
```rust
struct A;
struct B;
trait Tr {}
impl Tr for B {}
#[lang = "deref"]
trait Deref { type Target; }
impl Deref for A { type Target = B; }

struct S<T>(T);
impl<T: Tr> S<T> { fn bar(&self) {} }

fn foo(a: S<A>) {
    a.bar();
    //^ unresolved
}
```
I think it's reasonable to apply deref logic not for impls resolution, but for method resolution. I.e. move your code to `findMethodsAndAssociatedFunctionsFor`.

farodin91:
@matklad Resolved
@vlad20012 Done a bit different to use caching.

matklad:
Let's merge this!

bors r+

I really don't like how dereferencing ends up handled directly while queering the index, will try to move it elsewhere later today, together with `stripAnyRef` and fingerprinting logic. 

Thanks!

PR #1186, opened 2017-04-29
local variable documentation

Undin:
Add quick documentation for local variables.


Undin:
At this moment I use `PresentationInfo.type` as a prefix of documentation string.
It's quite good for function params and `let` bindings, and not so good for other cases. So if someone suggests something better I'm happy to use this suggestion.

matklad:
bors r+

Thanks!

PR #1187, opened 2017-04-30
MACRO: Working on matcher

farodin91:
TODO:
- Parsing of complex types, should work in the most cases (Any idea for false positives)
- Add more tests (partial done by using diesel macros)

@matklad Any feedback?


src/main/grammars/RustParser.bnf
```diff
@@ -983,6 +983,25 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+// block: a brace-delimited sequence of statements and optionally an expression. Example: { log(error, "hi"); return 12; }.
+// item: an item. Examples: fn foo() { }; struct Bar;.
+// meta: a "meta item", as found in attributes. Example: cfg(target_os = "windows").
+// tt: a single token tree.
+
+MacroDefinitionPatternList ::= <<macro_args_delim (MacroDefinitionPattern ';'?)* >>
+MacroMatcher ::= MacroMatching ((','| ';') MacroMatching)*
+MacroMatching ::= (MacroSimpleMatching | MacroComplexMatching)
+MacroComplexMatching ::= '$' '(' MacroSimpleMatching ')' (',' | ';')? ('*' | '+')
+MacroSimpleMatching ::= '$' identifier ':' identifier
```
farodin91: I want to write `MacroSimpleMatching ::= '$' identifier ':' MacroPatternType`, but than i get an stub error.
I don't have an idea to fix this here?

src/main/grammars/RustParser.bnf
```diff
@@ -983,6 +983,45 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+                 | (MacroSimpleMatching | MacroComplexMatching)+
+                 | << placeholderMacroToken >>+
+
+private MacroMatchingTrees ::=  MacroMatchingTree*
+private PossibleTokenAfterComplex ::= (',' | ';' | '::' | '.')
+MacroComplexMatching ::= '$' '(' MacroMatchingTrees ')' PossibleTokenAfterComplex? ('*' | '+')
+MacroSimpleMatching ::= '$' MacroBinding ':' MacroPatternType
+MacroBinding ::= identifier
+private MacroPatternType ::= ( "ident" | "path" | "expr" | "ty" | "pat" | "stmt" | "block" | "item" | "meta" | "tt" )
+MacroDefinitionPattern ::= '(' MacroMatcher ')' '=>' (MacroRightTokenTree | TokenTree)
```
matklad: Why do we have this `| TokenTree` branch? I think `MacroRightTokenTree` should be enough and should be more efficient.

src/main/grammars/RustParser.bnf
```diff
@@ -983,6 +983,45 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+
+MacroRightComplexMatching ::= '$' '(' MacroTokenTrees ')' PossibleTokenAfterComplex? ('*' | '+')
```
matklad: Instead of `Right`, lets use `MacroPattern` and `MacroBody`? Or perhaps there are even better names?

src/main/grammars/RustParser.bnf
```diff
@@ -983,6 +983,45 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+                 | '[' MacroMatchingTrees ']'
+                 | (MacroSimpleMatching | MacroComplexMatching)+
+                 | << placeholderMacroToken >>+
+
+private MacroMatchingTrees ::=  MacroMatchingTree*
+private PossibleTokenAfterComplex ::= (',' | ';' | '::' | '.')
+MacroComplexMatching ::= '$' '(' MacroMatchingTrees ')' PossibleTokenAfterComplex? ('*' | '+')
+MacroSimpleMatching ::= '$' MacroBinding ':' MacroPatternType
+MacroBinding ::= identifier
+private MacroPatternType ::= ( "ident" | "path" | "expr" | "ty" | "pat" | "stmt" | "block" | "item" | "meta" | "tt" )
```
matklad: I would rather use `identiier` here and then write an annotator/completion to check possible variants.

src/main/grammars/RustParser.bnf
```diff
@@ -983,6 +983,45 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+
+MacroDefinitionPatternList ::= <<macro_args_delim (MacroDefinitionPattern ';'?)* >>
+MacroMatcher ::= MacroMatchingTree*
+MacroMatchingTree ::=  '(' MacroMatchingTrees ')'
+                 | '{' MacroMatchingTrees '}'
+                 | '[' MacroMatchingTrees ']'
+                 | (MacroSimpleMatching | MacroComplexMatching)+
+                 | << placeholderMacroToken >>+
+
+private MacroMatchingTrees ::=  MacroMatchingTree*
```
matklad: I'd rather use `MacroMathingTree*` directly instead of a private rule: it's the same number of characters, and the meaning is more obvious! :)

src/main/grammars/RustParser.bnf
```diff
@@ -983,6 +983,45 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+private PossibleTokenAfterComplex ::= (',' | ';' | '::' | '.')
+MacroComplexMatching ::= '$' '(' MacroMatchingTrees ')' PossibleTokenAfterComplex? ('*' | '+')
+MacroSimpleMatching ::= '$' MacroBinding ':' MacroPatternType
+MacroBinding ::= identifier
+private MacroPatternType ::= ( "ident" | "path" | "expr" | "ty" | "pat" | "stmt" | "block" | "item" | "meta" | "tt" )
+MacroDefinitionPattern ::= '(' MacroMatcher ')' '=>' (MacroRightTokenTree | TokenTree)
+
+MacroRightComplexMatching ::= '$' '(' MacroTokenTrees ')' PossibleTokenAfterComplex? ('*' | '+')
+MacroRightSimpleMatching ::= '$' (MacroBinding | "crate")
+private MacroTokenTrees ::= MacroTokenTree*
```
matklad: Same here, I think we can get rid of this rule.

matklad:
Looks great! Parsing `macro_rules` is definitely the first thing to do!

We need to make sure that we parse all reasonable macro_rules though! Parsing tests in `Test slow` should handle that, but it'd be great to create a special file with macro_rules for `CompleteParsingTestCase`.

matklad:
>Parsing of complex types, should work in the most cases (Any idea for false positives)

Just throw some existing macro heavy code at it :) Serde/Diesel/pest/nom seem like good examples.

matklad:
>Better highlighting

This definitely should be handled in a separate PR, after the basic grammar is working.

src/main/grammars/RustParser.bnf
```diff
@@ -984,6 +984,37 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
...
+               | '{' MacroMatchingTree* '}'
+               | '[' MacroMatchingTree* ']'
+MacroMatchingTree ::=  '(' MacroMatchingTree* ')'
+                     | '{' MacroMatchingTree* '}'
+                     | '[' MacroMatchingTree* ']'
+                     | (MacroPatternSimpleMatching | MacroPatternComplexMatching)+
+                     | << placeholderMacroToken >>+
+
+MacroPatternComplexMatching ::= '$' '(' MacroMatchingTree* ')' <<tokenAfterComplexPattern>>* ('*' | '+')
+MacroPatternSimpleMatching ::= '$' MacroBinding ':' identifier
```
farodin91: I need to support also type.

src/test/resources/org/rust/lang/core/parser/fixtures/complete/diesel_macros.rs
```diff
@@ -0,0 +1,712 @@
...
+            my_type -> MyCustomType,
+        }
+    }
+
+    #[test]
+    fn table_with_custom_schema() {
+        let expected_sql = "SELECT `foo`.`bars`.`baz` FROM `foo`.`bars`";
+        assert_eq!(expected_sql, debug_sql!(bars::table.select(bars::baz)));
+    }
+}
```
matklad: Let's remove comments, tests and everything else which is not a `macro_rules` definition. 

These parser tests are a bit annoying in that if you tweak parser or PSI structure, they tend to get invalidated as collateral damage, even if the changes were done to an unrelated part of the grammar. So it makes sense to keep them short and dense!

src/main/grammars/RustParser.bnf
```diff
@@ -984,6 +984,37 @@ MacroDefinition ::= MacroRulesInvocation identifier ItemMacroArg {
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
+MacroDefinitionPatternList ::= <<macro_definition_delim (MacroDefinitionPattern ';'?)* >>
+
+private meta macro_definition_delim ::= '(' <<param>> ')' ';'
+                                      | '{' <<param>> '}'
+                                      | '[' <<param>> ']' ';' { pin = 1 }
+MacroDefinitionPattern ::= MacroPattern '=>' MacroBody
+MacroBinding ::= (identifier | "type")
```
farodin91: @matklad Any tip to all nearly keywords here?

matklad: Hm, so inside a macro you are allowed to use keyword as an identifier? Weird. I think that perhaps just listing all the keywords here is the best solution.

farodin91: Macro pattern is only crate not allowed.

matklad:
Looking good! The next step would be to implement `RsReference` for `$foo` references in macros.

PR #1188, opened 2017-04-30
TY: initial support of array expression type inference

Undin:
Support inference of:
* sequence expressions like `[1, 2, 3]`. Array element type is the same with type of first element of the sequence. Should we infer type of all elements?
* expressions with initialization value like `[0; 8]`. Size can be extracted only from integer literal.

TODO:
* support compile time expressions as array size

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
@@ -1,16 +1,28 @@
 package org.rust.lang.core.psi.ext
 
 import org.rust.lang.core.psi.RsArrayType
+import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.psi.RsLitExpr
+import org.rust.lang.core.types.types.RustIntegerType
 
 val RsArrayType.arraySize: Int? get() {
     val stub = stub
     if (stub != null) {
         return if (stub.arraySize == -1) null else stub.arraySize
     }
+    return calculateArraySize(expr)
+}
 
+// TODO: support constants and compile time expressions
+fun calculateArraySize(expr: RsExpr?): Int? {
+    val sizeLiteral = (expr as? RsLitExpr)
+        ?.integerLiteral
+        ?.text
+        ?.removeSuffix(RustIntegerType.Kind.usize.name)
+    // BACKCOMPAT: Kotlin API 1.0
     return try {
-        expr?.text?.toInt() // TODO need more precise handling
+        sizeLiteral?.toInt()
     } catch (e: NumberFormatException) {
-        return null
+        null
     }
 }
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
@@ -127,10 +127,25 @@ fun inferExpressionType(expr: RsExpr): RustType {
                 RustUnknownType
         }
 
+        is RsArrayExpr -> inferArrayType(expr)
+
         else -> RustUnknownType
     }
 }
 
+private fun inferArrayType(expr: RsArrayExpr): RustType {
+    val (elementType, size) = if (expr.semicolon != null) {
+        val elementType = expr.initializer?.type ?: return RustSliceType(RustUnknownType)
+        val size = calculateArraySize(expr.sizeExpr) ?: return RustSliceType(elementType)
+        elementType to size
+    } else {
+        val elements = expr.arrayElements
+        if (elements.isNullOrEmpty()) return RustSliceType(RustUnknownType)
+        // '!!' is safe here because we've just checked that elements isn't null
+        elements!![0].type to elements.size
+    }
+    return RustArrayType(elementType, size)
+}
 
 private val RsBlock.type: RustType get() = expr?.type ?: RustUnitType
 
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayExpr.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayExpr.kt
@@ -1,0 +1,22 @@
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsArrayExpr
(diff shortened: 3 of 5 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -127,10 +127,38 @@ fun inferExpressionType(expr: RsExpr): RustType {
...
+        Pair(elementExpr.type, size)
+    } else {
+        if (expr.exprList.isEmpty()) return RustUnknownType
+        Pair(expr.exprList[0].type, expr.exprList.size)
+    }
+    return RustArrayType(elementType, size)
+}
+
+// TODO: support constants and compile time expressions
+private fun calculateArraySize(expr: RsExpr): Int? {
```
matklad: This logic should be used for `org.rust.lang.core.psi.ext.RsArrayTypeKt#getArraySize` as well. Looks like making this function public would do the trick?

Undin: Yes, it's a good idea. I've made it public, moved into `org.rust.lang.core.psi.ext.RsArrayType.Kt`(I think it's better place for this function if we want to use it for size calculation of `RsArrayType`) and use from both places.

src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
```diff
@@ -1,16 +1,28 @@
...
+    return calculateArraySize(expr)
+}
 
+// TODO: support constants and compile time expressions
+fun calculateArraySize(expr: RsExpr?): Int? {
+    val sizeLiteral = (expr as? RsLitExpr)
+        ?.integerLiteral
+        ?.text
+        ?.removeSuffix(RustIntegerType.Kind.usize.name)
+    // TODO: use String.toIntOrNull when we'll use kotlin api >= 1.1
```
matklad: We usually mark such things with 

```
BACKCOMPAT: Kotlin API 1.0 / 2016.1 / Rust 1.12
```

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -127,10 +127,24 @@ fun inferExpressionType(expr: RsExpr): RustType {
...
 
+        is RsArrayExpr -> inferArrayType(expr)
+
         else -> RustUnknownType
     }
 }
 
+private fun inferArrayType(expr: RsArrayExpr): RustType {
+    val (elementType, size) = if (expr.semicolon != null) {
+        if (expr.exprList.size != 2) return RustUnknownType
```
matklad: It's better to return `RsArrayType(RustUnknownType)` rather than `RustUnknownType`. 

In general in the IDE the goal is to infer as may types as possible, and not to find a type error as early as possible.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -127,10 +127,24 @@ fun inferExpressionType(expr: RsExpr): RustType {
...
+        val (elementExpr, sizeExpr) = expr.exprList
+        val size = calculateArraySize(sizeExpr) ?: return RustUnknownType
```
matklad: Same here.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -127,10 +127,24 @@ fun inferExpressionType(expr: RsExpr): RustType {
...
+        Pair(elementExpr.type, size)
```
matklad: It's usually better to use the `to` infix function to construct pairs:

```Rust
elementExpr.type to size
```

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
Undin: But `RsArrayType` requires not only element type but size too. What is it supposed to be array size when we can't calculate it?

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
Undin: Generally, I'm not sure that this check is necessary here because this is contrary to grammar https://github.com/intellij-rust/intellij-rust/blob/c495d7caf2e7272e80cf53fc3dcf5c1e49f144dc/src/main/grammars/RustParser.bnf#L922.
Maybe drop it at all?

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
matklad: We use `RustSliceType(componentType)` in a similar situation for expressions. It does not feel exactly right, but it's better than `UnknownType`.

matklad: Perhaps we should just merge slice and array types?

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
matklad: >Maybe drop it at all?

Yeah, it should be safe to just `check(expr.exprList.size == 2)`. Better still would be to add general accessors to `RsArrayExpr` like `RsArrayExpr.initializer: RsExpr?`, `RsArrayExpr.size: RsExpr?` and ``RsArrayExpr.arrayElements: List<RsExpr>?`.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
Undin: At this moment I can't say what it will affect and that will be better because I'm still not very familiar with the internal type system.
Let's use `RustSliceType` in such ambiguous cases for now and if you consider that `RsArrayType` and `RustSliceType` must be represented with one class we will merge them in other PR

Undin:
Fixed all comments (without merging of `RustArrayType` and `RustSliceType`)

matklad:
:+1:

bors r+

PR #1200, opened 2017-05-03
Improve quick documentation of struct fields

### LINKED ISSUE #1089: Display type of struct field names when viewed in quick-doc
It would be nice if in the following example:

```rust
struct S {
    /// A number.
    x: i32,
}

fn main() {
    let s = S{x: 0};
    let y = s.x;
}
```

Quick doc on the `.x` of `s` would show you that `x` is of type `i32` in addition to its doc string. This becomes even more useful when working with libraries that have `struct`s inside `struct`s.

Undin:
Fixes #1089

### DIFF
--- a/src/test/kotlin/org/rust/ide/docs/RsQuickDocumentationTest.kt
+++ b/src/test/kotlin/org/rust/ide/docs/RsQuickDocumentationTest.kt
@@ -23,6 +23,7 @@ class RsQuickDocumentationTest : RsDocumentationProviderTest() {
     fun testGenericStructVariable() = checkDoc()
     fun testTupleDestructuring() = checkDoc()
     fun testConditionalBinding() = checkDoc()
+    fun testStructField() = checkDoc()
 
     private fun checkDoc() = compareByHtml { element, originalElement ->
         RsDocumentationProvider().generateDoc(element, originalElement)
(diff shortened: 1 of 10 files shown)


src/test/kotlin/org/rust/ide/docs/RsQuickDocumentationTest.kt
```diff
@@ -23,6 +23,7 @@ class RsQuickDocumentationTest : RsDocumentationProviderTest() {
     fun testGenericStructVariable() = checkDoc()
     fun testTupleDestructuring() = checkDoc()
     fun testConditionalBinding() = checkDoc()
+    fun testStructField() = checkDoc()
```
matklad: Perhaps we should change this tests to use inline string literal fixtures instead of file fixtures? Would this be more convenient? One can mark string literals with `@Language("Rust")` or `@Langauge("html")` to get highlighting inside of the literals.

matklad:
bors r+

Thanks!

I should be using quick doc/quick definition more myself :( Can't break the habit of doing "go to definition, go back".

src/test/kotlin/org/rust/ide/docs/RsQuickDocumentationTest.kt
Undin: Yes, I like the idea with string literals. For me, it's much more convenient to look at the content of tests in the same file.

matklad: Would send a PR ;) ?

Undin: No problem :)

matklad: Thanks!

PR #1204, opened 2017-05-04
TYPE: slightly refactor search of deref'ed types

matklad:
Separate "look in the index and cache" and "run dfs on derefs" phases


matklad:
cc @farodin91

PR #1206, opened 2017-05-05
T: inline documentation tests

Undin:
* add new base `findElement*InEditor` method to `RsTestBase` to get caret offset of marker too
* create base method for inline doc tests in `RsDocumentationProviderTest`
* inline source of documentation tests into `RsQuickDocumentationTest`
* use `//^` marker instead of `<caret>` to prevent errors in syntax highlighting


matklad:
bors r+

Much better now, thanks a lot!

PR #1213, opened 2017-05-05
COMP: Show the trait of an method

### LINKED ISSUE #1190: Show trait of a method in autocomplete popup
It would a cool function to see some background infos.

farodin91:
Fixes #1190

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
@@ -17,6 +17,9 @@ object CompletionEngine {
     const val KEYWORD_PRIORITY = 10.0
 }
 
+private val RsFunction.extraTailText: String
+    get() = parentOfType<RsImplItem>()?.traitRef?.text?.let { " of $it" } ?: ""
+
 fun RsCompositeElement.createLookupElement(scopeName: String): LookupElement {
     val base = LookupElementBuilder.create(this, scopeName)
         .withIcon(if (this is RsFile) RsIcons.MODULE else getIcon(0))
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
```diff
@@ -17,6 +17,11 @@ object CompletionEngine {
     const val KEYWORD_PRIORITY = 10.0
 }
 
+fun RsFunction.generateAppendTailText(): String {
```
matklad: This can be `private`.

src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
```diff
@@ -35,6 +40,7 @@ fun RsCompositeElement.createLookupElement(scopeName: String): LookupElement {
         is RsFunction -> base
             .withTypeText(retType?.typeReference?.text ?: "()")
             .withTailText(valueParameterList?.text?.replace("\\s+".toRegex(), " ") ?: "()")
+            .appendTailText(generateAppendTailText(), false)
```
alygin: Don't you think it'll look better with `grayed = true`? Won't distract from the function signature so much.

src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
```diff
@@ -17,6 +17,11 @@ object CompletionEngine {
     const val KEYWORD_PRIORITY = 10.0
 }
 
+private fun RsFunction.generateAppendTailText(): String {
+    val trait = parentOfType<RsImplItem>()?.traitRef?.text ?: return ""
+    return " by $trait"
+}
```
alygin: It looks more like a property. Could also be more compact:

```rust
private val RsFunction.extraTailText: String
    get() = parentOfType<RsImplItem>()?.traitRef?.text?.let { " of $it" } ?: ""
```

src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
farodin91: Look better

matklad:
Thanks!

PR #1215, opened 2017-05-05
COMP: Should not complete for function with test attribute

### LINKED ISSUE #1189: Autocomplete currently suggests test functions
Using Hyper for example the functions defined with #[test] are suggested in the autocomplete. If its possible to exclude any #[test] method from autocompleting that would be superb and would let us more easily see the methods we are interested in.

Btw, keep up the good work - you're doing brilliantly!

farodin91:
Fixes #1189

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
@@ -2,7 +2,10 @@ package org.rust.lang.core.resolve
 
 import com.intellij.codeInsight.lookup.LookupElement
 import org.rust.lang.core.completion.createLookupElement
+import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.ext.RsCompositeElement
+import org.rust.lang.core.psi.ext.allAttributes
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -52,6 +55,8 @@ fun collectResolveVariants(referenceName: String, f: (RsResolveProcessor) -> Uni
 fun collectCompletionVariants(f: (RsResolveProcessor) -> Unit): Array<LookupElement> {
     val result = mutableListOf<LookupElement>()
     f { e ->
+        if((e.element as? RsFunction?)?.isTest ?: false)
```
matklad: A space is missing after the `if` ;) 

Also we usually format short ifs on a single line: `if (cond) return false`, or add `{}`.

farodin91: Is this the correct position?

matklad: yep!

matklad:
Thanks!

bors r+

PR #1216, opened 2017-05-06
RUN: Select channel in run configuration

alygin:
If rustup is available, allows to select a channel in run configuration:

screen shot 2017-05-06 at 14 38 26

If the channel is not `[Default]`, `rustup run {channel}` is used to execute cargo.

At the moment, the set of channels is predefined and their presence is not checked. In the future, the setting can be changed to "Toolchain" and use `rustup show` to fill the list automatically. But I guess such improvement should somehow merge into the current `RustToolchain` infrastructure, and thus be carefully thought over beforehand.

What do you think? Is it ok to have the feature with this caveat?


matklad:
Yeah, I think it's ok to hardcode a list of channels for now. I also though that maybe the `channel` should be a part of global configuration (i.e., channel per project). In theory, this seems better, but in practice you want nightly for clippy, so putting to runconfig is a way to go.

alygin:
> In theory, this seems better, but in practice you want nightly for clippy, so putting to runconfig is a way to go.

Yeah, and working on the same crate using different channels is also uncommon:

```rust
#[cfg(feature = "nightly")]
fn foo() { println!("nightly foo"); }

#[cfg(feature = "stable")]
fn foo() { println!("stable foo");  }
```

So I think we'll have to support quick and easy switching between channels anyway.

PR #1218, opened 2017-05-06
TY: initial support of index expression type inference

Undin:
* more flexible `canUnifyWith` for RustNumericType: numeric type can be unified with type with different kind if it's weak
* initial support for index expression type inference using `Index` trait

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt
@@ -9,9 +9,9 @@ import org.rust.lang.core.resolve.indexes.RsImplIndex
 import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.ty.findImplsAndTraits
 import org.rust.lang.core.types.infer.remapTypeParameters
+import org.rust.lang.core.types.ty.TyReference
 import org.rust.lang.core.types.type
 import org.rust.lang.core.types.ty.TyUnknown
-
 
 fun findDerefTarget(project: Project, ty: Ty): Ty? {
     val impls = RsImplIndex.findImpls(project, ty)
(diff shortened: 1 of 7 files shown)


farodin91:
It would better to access the language attribute.

Undin:
@farodin91 What do you mean?
I just use language attribute `index` to find correct trait

farodin91:
Look at wrong place

src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt
```diff
@@ -35,13 +36,39 @@ fun findIteratorItemType(project: Project, ty: RustType): RustType {
...
+            val indexType = impl.functionList
+                .find { it.name == "index" }
+                ?.valueParameterList
+                ?.valueParameterList
+                // 'index' function have only one value parameter
+                // fn index(&self, index: Idx) -> &Self::Output;
+                ?.getOrNull(0)
+                ?.typeReference
+                ?.type ?: continue
+            if (indexType.canUnifyWith(indexExprType, project)) {
```
matklad: Let's make the check more optimistic: if there's only impl `Index` for the type (common case), we should just use it, without the check that `indexExprType` match.

src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt
```diff
@@ -35,13 +36,39 @@ fun findIteratorItemType(project: Project, ty: RustType): RustType {
     return rawType.substitute(typeParameterMap)
 }
 
+fun findIndexOutputType(project: Project, containerType: RustType, indexExprType: RustType): RustType {
+    val impls = RsImplIndex.findImpls(project, containerType)
+    for (impl in impls) {
+        val trait = impl.traitRef?.resolveToTrait ?: continue
+        if (trait.isIndex) {
+            // TODO: get index type from impl declaration
+            val indexType = impl.functionList
```
matklad: Ideally, we should be able to get type from `Type Output = ` if we fix the logic over there: https://github.com/intellij-rust/intellij-rust/blob/8ae5457c32c48e10d8b99ee17aeb112ac204abf6/src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt#L17

Undin: But the declaration of output type doesn't contain type of possible index expression. Or do you suggest it instead of using `lookupAssociatedType` function?

matklad: Sorry, I for some reason thought that this is not for the `Index` type, but for the output type. 

Yeah, for the index type I think it's easiest just to look at the function signature, but, as per comment below, we should only use it if we find more than one `Index` for one type and need to select between them.

Undin:
Fix all comments

matklad:
Ah, one more thing: could you add an "integration style" test to `RsTypeAwareResolveTest` or, even better to `RsStdlibResolveTest` with a real `Vec<T>`? 

Type inference tests are pretty low level and I feel more confident when I know that `go to definition` actually work!

Undin:
No problem :)
Done

matklad:
Splendid!

bors r+

PR #1220, opened 2017-05-07
RES: Support auto deref_mut

farodin91:


src/test/kotlin/org/rust/lang/core/resolve/RsMultiResolveTest.kt
```diff
@@ -21,6 +21,29 @@ class RsMultiResolveTest : RsResolveTestBase() {
         }
     """)
 
+    //FIXME: should resolve to a single mut method!
+    fun `test auto deref and deref mut`() = doTest("""
+        #[lang = "deref"]
+        trait Deref { type Target; }
+        #[lang = "deref_mut"]
+        trait DerefMut { type Target; }
```
matklad: Hm, there's no  `Target` type in `DerefMut` in Rust. Instead, it extends `Deref` (`DerefMut: Deref`) and uses its `Target`.

matklad:
Hm, I've took at closer look at this and I don't understand why we need to handle `DerefMut` at all at this stage.

`DerefMut` is exactly like `Deref`, except that you can get `&mut Target` instead of `& Target`. And for resolve purposes `&mut T` and `&T` are equivalent. That is, I don't believe that by recognizing `DerefMut` we will be able to resolve more things. 

We could use `DerefMut` to implement "type cannot be indexed mutably" annotation, but we need to implement near perfect name resolution before.

farodin91:
A possible case could be in future:
```rust
impl &mut Test {
 fn foo(self) {
 }
}
```

matklad:
>A possible case could be in future:

Hm, but there's no `DerefMut` trait in there? 

I also worry about implementing something for the future, because it often turns out that the initial implementation does not fit the future use case perfectly. My rule of thumb is that additions to code should be justified by "integration" resolve or completion tests. Looks like all tests added in this PR will pass on `master` branch just fine because we already have Deref.

farodin91:
I'll close this PR. I only want to show up a possible case.

PR #1222, opened 2017-05-09
CARGO: Support proc-macro crate kind

alygin:
Cargo reports library crates with `proc-macro = true` as `kind = proc-macro`, not `kind = lib` (for example, `serde_derive`). As a result, such crates are not resolved in `extern crate` clauses. This fix makes proc-macro crates treated as lib crates and correctly resolved.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
@@ -151,6 +151,7 @@ object CargoMetadata {
             "example" -> CargoWorkspace.TargetKind.EXAMPLE
             "test" -> CargoWorkspace.TargetKind.TEST
             "bench" -> CargoWorkspace.TargetKind.BENCH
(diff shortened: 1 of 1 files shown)


mkaput:
👍

matklad:
The situation with cargo `kind` is a bit murky at the moment. 

There are two dimension of variability here:

* `rustc` `--crate-type` argument (which can be a vector of strings). It is documented in https://doc.rust-lang.org/reference/linkage.html.

* Cargo `--lib`, `--bin NAME`, `--example NAME` flags.

These two dimensions used to be collinear (`--bin` implies `--crate-type="bin"`), but they are not any more. There are `bin` and `lib` examples, for example. 

This was cleaned up on the Cargo's side in https://github.com/rust-lang/cargo/pull/3862/.

`crate_type` field is `--crate-type` from compiler. The old `kind` is a roundabout way to express `--bin` / `--lib` thing. I think that out `TargetKind` should mirror the flag directly, so we don't need `PROC_MACRO` variant at all and can use `LIB` instead. 

When we'll need to decide if something is a `proc_macro`, we'll look at `crate_type` then.

PR #1223, opened 2017-05-09
CARGO: Support feature gated stdlib dependencies

alygin:
Rust stdlib includes several crates that can be used as dependencies without mentioning in Cargo.toml yet are not a part of the prelude. For instance, we can use `extern crate rand;` in nightly when `#![feature(rand)]` is provided. So, such crates should be added as stdlib roots to be resolved (they now have `isRoot = false`).

The problem is that there are external crates with the same names (`rand` and `term` are main troublemakers). Adding such crates as dependencies to Cargo.toml brings name conflicts and breaks resolving when [we use the mentioned approach](https://github.com/intellij-rust/intellij-rust/pull/1145#issuecomment-299711548).

This fix introduces stdlib crate types: root, featured, dependency. Unlike roots, featured crates are only added to packages as a part of stdlib when there're no explicit dependencies with the same names.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -3,6 +3,7 @@ package org.rust.cargo.project.workspace
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
 import org.rust.cargo.toolchain.impl.CleanCargoMetadata
+import org.rust.cargo.util.StdLibType
 import java.util.*
 import java.util.concurrent.atomic.AtomicReference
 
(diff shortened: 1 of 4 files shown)


himikof:
Is "featured" an appropriate term to use? I believe it is usually called "feature gated" in the Rust world. And their semantics of being used only when no explicit dependency exists are better described as "fallback" crates, I think.

alygin:
Yeah, I agree that "featured" is a bad choice. I like "feature gated" more than "fallback" because it better suits its original meaning, while "fallback" relates only to this particular task.

If there will be no more suggestions, I'll rename it to `FEATURE_GATED`.

matklad:
`FEATURE_GATED` sounds right!

alygin:
Feature gated it is.

matklad:
bors r+

Thanks @alygin & @himikof !

PR #1224, opened 2017-05-09
MACRO: Simple pattern resolve

farodin91:


farodin91:
@matklad Any tip for scoping?

src/main/grammars/RustParser.bnf
```diff
@@ -1001,8 +1001,17 @@ MacroMatchingTree ::=  '(' MacroMatchingTree* ')'
                      | (MacroPatternSimpleMatching | MacroPatternComplexMatching)+
                      | << placeholderMacroToken >>+
 
-MacroPatternComplexMatching ::= '$' '(' MacroMatchingTree* ')' <<tokenAfterComplexPattern>>* ('*' | '+')
-MacroPatternSimpleMatching ::= '$' MacroBinding ':' identifier
+MacroPatternComplexMatchingBody ::= '(' MacroMatchingTree* ')' {
+  pin = 1
+}
+MacroPatternComplexMatching ::= '$' MacroPatternComplexMatchingBody <<tokenAfterComplexPattern>>* ('*' | '+')  {
+  implements = [ "org.rust.lang.core.psi.ext.RsNamedElement" ]
```
matklad: Only simple atomic bindings like `$foo:ty` should be `RsNamedElements` and `RsReferences`, you don't need to mach the whole complex matching structure during resolve.

src/test/kotlin/org/rust/lang/core/resolve/RsMarcoResolveTest.kt
```diff
@@ -0,0 +1,56 @@
+package org.rust.lang.core.resolve
+
+class RsMacroResolveTest : RsResolveTestBase() {
+    fun `test resolve simple matching with multiple pattern definition`() = checkByCode("""
+        macro_rules! test {
+            (${'$'}test:expr) => (
```
matklad: Gosh, this `${'$'}` thing looks ugly :(

I wonder if this would look a tiny bit better if we define

```
private val `$` = '$'
```
at the file level and use

```
$`$`
```

instead?

matklad:
>@matklad Any tip for scoping?

macro definitions don't have nesting and shadowing, so we don't really need to handle scoping here I guess. `parentOfType` followed by `descendantsOfType` looks like a solid implementation to me!

farodin91:
Your are able to nest complex into complex struct.

matklad:
>Your are able to nest complex into complex struct.

But `$names` must be unique, so we don't need to consider the nesting while resolving.

farodin91:
```rust
macro_rules! test {
          ($($test:expr)+) => ($test)
                                          //^ unresolved
}
```

matklad:
>This should be case of unresolve due not nested.

I think it should be resolved, and then we should provide additional error annotation on top that "$test pattern is on the wrong level of nesting". In general, in IDE you should resolve as much things as possible, and handle all "mis-resolves" by annotators. Such "optimistic" resolution is very helpful, because it allows to provide fixes and assistances even if the code in question is broken.

matklad:
bors r+

Thanks!

One more thing to add would be highlighting of simple patterns. This should be handled `org.rust.ide.annotator.RsHighlightingAnnotator`. 

`colorFor` function should color the named element. The references should then be colored automatically by `highlightReference` function.

PR #1235, opened 2017-05-12
Advanced resolve

matklad:
This is a rather major improvement of type inference (at least, ground work for this improvement). 

This allows us to resolve this case:

```Rust
        trait I<T> {
            fn foo(&self) -> T;
        }

        struct S;
        impl S {
            fn bar(&self) {}
        }     //X

        fn baz<T: I<S>>(t: T) {
            t.foo().bar()
        }         //^
```

Previously, we were able to resolve `.foo()`, but not `.bar()`. The problem was that while we know the function `foo`, we lost the knowlage of its type parameter `T`.

The crux of the fix is to extend `ScopeEntry` with `typeArguments` field, so that once we resolve `foo`, we know that `T` is in fact maps to `S` for this particular reference.

This type arguments must be threaded through all of the resolve infrastructure, so `BoundElement` class is introduced to hold a pair of PsiElement and TypeArguments, and `advanceResolve` function is added to `RsReference`, which returns bound PsiElement .

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
@@ -10,22 +10,25 @@ import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.RustPsiImplUtil
 import org.rust.lang.core.stubs.RsTraitItemStub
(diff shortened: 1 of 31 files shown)


matklad:
This actually is pretty similar to how Java handles generics: `TypeArguments` are a watered-down version of `PsiSubstitutor`, and `ScopeEntry/BoundElement` correspond to `JaveResolveResult`.

Issue #601, opened 2016-08-16
Type-inference for closure's params


alexeykudinkin:
This meta-issue, was created not only for the purpose of tracking progress on the matter, but also to accommodate all discussions about implementation details.

```
trait T {
    fn bar();
}

fn foo<F>(f: F) where F: Fn(T) -> Unit {}

fn main() {
    foo(|t| { t.bar() /* T::bar() */; () })
}
```

alexeykudinkin:
Primarily blocked by #602.

PR #1236, opened 2017-05-13
TY: Basic support for closure type inference

farodin91:
Initial support for #601


matklad:
I definitely like the tests here! 

The implementation should be a bit different and leverage existing `TyTypeParameter` logic, but this can be handled separately (and there's some work needed with the grammar as well).

bors r+

farodin91:
I'll try to support iterator with an up coming pull request.

PR #1241, opened 2017-05-15
CARGO: Add "Run Clippy" action

### LINKED ISSUE #938: Add action to run Clippy
We already can run rustfmt: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt

Running clippy should be done in a similar fashion, but it should automatically use the nightly toolchain if Rustup is available.

alygin:
Adds the "Run Clippy" action:

screen shot 2017-05-15 at 23 47 30

Uses nightly if rustup is available, otherwise uses the default toolchain.

No more excuses not to use clippy ;)

Closes #938


matklad:
bors r+

Thanks!

matklad:
I guess I'll need to clean up all my old Rust projects now =/

PR #1242, opened 2017-05-17
TY: Infer for associated types

farodin91:
Didn't work on std in this version.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -186,7 +186,7 @@ fun processPathResolveVariants(path: RsPath, isCompletion: Boolean, processor: R
         if (base is RsTraitItem && qualifier.cself != null) {
             if (processAll(base.typeAliasList, processor)) return true
         }
-        if(base is RsStructItem && qualifier.cself != null) {
+        if (base is RsStructItem && qualifier.cself != null) {
             val traitItem = path.parentOfType<RsImplItem>()
             if (traitItem != null && processAll(traitItem.typeAliasList, processor)) return true
         }
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -183,10 +183,10 @@ fun processPathResolveVariants(path: RsPath, isCompletion: Boolean, processor: R
             val s = base.`super`
             if (s != null && processor("super", s)) return true
         }
-        if (base is RsTraitItem && qualifier.cself != null) {
+        if (base is RsTraitItem && qualifier.referenceName == "Self") {
             if (processAll(base.typeAliasList, processor)) return true
         }
-        if (base is RsStructItem && qualifier.cself != null) {
+        if (base is RsStructItem && qualifier.referenceName == "Self") {
             val traitItem = path.parentOfType<RsImplItem>()
             if (traitItem != null && processAll(traitItem.typeAliasList, processor)) return true
         }
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -186,6 +186,10 @@ fun processPathResolveVariants(path: RsPath, isCompletion: Boolean, processor: R
         if (base is RsTraitItem && qualifier.cself != null) {
             if (processAll(base.typeAliasList, processor)) return true
         }
+        if(base is RsStructItem && qualifier.cself != null) {
+            val traitItem = path.parentOfType<RsImplItem>()
+            if (traitItem != null && processAll(traitItem.typeAliasList, processor)) return true
+        }
         if (processItemOrEnumVariantDeclarations(base, ns, processor, isSuperChain(qualifier))) return true
         if (base is RsTypeBearingItemElement && parent !is RsUseItem) {
             if (processAssociatedFunctionsAndMethodsDeclarations(base.project, base.type, processor)) return true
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -185,6 +185,10 @@ fun processPathResolveVariants(path: RsPath, isCompletion: Boolean, processor: R
         if (base is RsTraitItem && qualifier.cself != null) {
             if (processAll(base.typeAliasList, processor)) return true
         }
+        if(base is RsStructItem && qualifier.cself != null) {
```
matklad: Formatting is off :)

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -136,6 +144,19 @@ fun inferExpressionType(expr: RsExpr): Ty {
     }
 }
 
+private fun inferTypeForExprAndDefaultMethodImpl(expr: RsMethodCallExpr, method: RsFunction, type: RsTypeReference?): Ty {
+    val traitItem = method.parentOfType<RsTraitItem>() ?: return TyUnknown
+    val caller = expr.expr as? RsPathExpr ?: return TyUnknown
+    val typeStruct = caller.type as? TyStruct ?: return TyUnknown
+    val project = PsiUtilCore.getProjectInReadAction(typeStruct.item)
```
matklad: `expr.project` should work here.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -136,6 +144,19 @@ fun inferExpressionType(expr: RsExpr): Ty {
...
+    val impls = RsImplIndex.findImpls(project, typeStruct)
```
matklad: It's better not to call into `RsImplIndex` directly, but to use `findImplsAndTraits` function, because it handles stuff like bounds on type parameters.

src/test/kotlin/org/rust/lang/core/type/RsExpressionTypeInferenceTest.kt
```diff
@@ -476,4 +476,58 @@ class RsExpressionTypeInferenceTest : RsTypificationTestBase() {
...
+
+    fun `test associated types for impl`() = testExpr("""
+        trait A {
+            type Item;
+            fn foo(self) -> Self::Item;
+        }
+        struct S;
+        impl A for S {
+            type Item = S;
+            fn foo(self) -> Self::Item {S}
```
matklad: Let's add spaces here: `-> Self::Item { S }`. Sorry that rust autoformatting does not work with langauge injection :) !

src/test/kotlin/org/rust/lang/core/type/RsTypeResolvingTest.kt
```diff
@@ -156,6 +156,18 @@ class RsTypeResolvingTest : RsTypificationTestBase() {
...
 
+    fun `test associated types for impl`() = testType("""
+        trait A {
+            type Item;
+            fn foo(self) -> Self::Item;
+        }
+        struct S;
+        impl A for S {
+            type Item = S;
+            fn foo(self) -> Self::Item {S}
```
matklad: and some spaces here.

src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
```diff
@@ -1024,4 +1024,26 @@ class RsTypeAwareResolveTest : RsResolveTestBase() {
             t.bar().foo(|e| { e.bar(); })
         }                      //^
     """)
+
+    fun `test associated type resolve for closure`() = checkByCode("""
+        trait A {
+            type Item;
+            fn filter<P>(self, predicate: P) -> Filter<Self, P> where Self: Sized, P: FnMut(&Self::Item) {
+                unimplemented!()
```
alygin: There's no need in these `unimplemented!()` macro invocations in test code snippets. `fn filter<P>(...) -> ... {}` will effectively do the same work.

farodin91:
Resolved!

farodin91:
@matklad Any tip to fix the error getting during resolving lamba expr for std?

matklad:
Looks like the error happens when accessing `cself` field of `RsPath`, which is not stubbed. However, `referenceName` is present in stubs, so you should be able to use something like `path.referenceName == "Self"` instead of accessing `cself` directly.

farodin91:
@matklad Any idea to fix associcated types for default impl functions ?
I don't like my current idea (the first two commits).

farodin91:
@matklad I build a working version could you look into?

matklad:
@farodin91 sorry for not responding in a timely manner: I am a bit overloaded with other stuff at the moment, but I'll take a look as soon as I can!

matklad:
Looks like test failures are legitimate?

farodin91:
I tested it and it work for me.

matklad:
Hm, that's weird, CI fails pretty consistently: https://travis-ci.org/intellij-rust/intellij-rust/jobs/244310883#L353

farodin91:
Different compuiter it didn't work

farodin91:
It didn't work to resolve into_iter on an Vec<String>

farodin91:
@matklad The test was incorrect.

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -16,28 +17,38 @@ data class TyTypeParameter private constructor(
...
+    fun getTraitRefs(): Sequence<RsTraitRef> = parameter.traitRefs
+
     override fun canUnifyWith(other: Ty, project: Project, mapping: TypeMapping?): Boolean {
         mapping?.merge(mutableMapOf(this to other))
         return true
     }
 
-    override fun substitute(map: TypeArguments): Ty = map[this] ?: this
+    override fun substitute(map: TypeArguments): Ty {
+        return map.filterKeys { it.toString() == this.toString() }.values.firstOrNull() ?: return this
```
matklad: This reads really weird... Why `map[this]` does not work here? If it doesn't that means that `map` is constructed incorrectly, not that we should resort to approximate by name only comparison.

farodin91: Resolved. Work but I don't have a idea to fix this easly.

matklad:
Thanks a lot @farodin91 !

This'll land as a part of https://github.com/intellij-rust/intellij-rust/pull/1375!

Issue #1195, opened 2017-05-02
Autocomplete for types with type constraints


zummenix:
Probably it hasn't been implemented yet but I've decided to open the issue anyway.

Example:
```rust
fn main() {
    println!("{:?}", foo()./* autocomplete here */);
}

fn foo() -> Result<i32, i32> {
    Ok(42)
}
```
Autocomplete works for methods on `impl<T, E> Result<T, E>` but doesn't work on `impl<T, E> Result<T, E> where E: Debug` and bunch of others with type constraints, i.e. doesn't autocomplete `unwrap`, `expect`, etc.

vlad20012:
Automatically implemented currently not supported. I'm working on it.
- [ ] Sized
- [ ] Sync
- [ ] Send
- [ ] #[derive()]

torkleyy:
@vlad20012 Will custom `#[derive]`d traits also work?

vlad20012:
@torkleyy I'm not sure yet, but I think so.

matklad:
Hm, I wonder if we should be less conservative here and just resolve all methods regardless of trait bouds.

We'll use trait bounds to disambiguate possible conflicts and (somewhere in the future), to annotate errors like "you are calling `.contains_key` on `BTreeMap`, but `Key` is not `Hash`. Add `#[derive(Hash)]`?"

farodin91:
@matklad I would prefer your idea.
For example
```rust
let mut map = HashMap::new();
                                         \\^unresolved
```
At this intellij can't know type constraints.

PR #1245, opened 2017-05-19
Relaxed trait matching

matklad:
Closes https://github.com/intellij-rust/intellij-rust/issues/1195

@vlad20012 this unfortunately reverts some of the recent trait matching work (all the test remain though!), in order to make two tests in `StdilbResolveTestCase` work :( 

The general problem here is that we want to resolve a superset of methods, and add filtering at the latest possible stage. This approach useful for several reasons: 

* our type inference is not perfect at the moment, and false positives in name resolution are generally better than false negatives. 

* even with perfect name resolution, it's better to resolve some method and show "T does not implement Foo" error message rather then generic "unresolved reference".

* code might be incomplete, and so we might not be able to infer necessary trait bounds.


vlad20012:
Ok, think it's reasonable for now. In future precise trait bound matching should be used at least for completion, I guess.

matklad:
>In future, precise trait bound matching should be used at least for completion, I guess.

Yep, we absolutely should have both precise and approximate matching! The problem is how to integrate it into the current resolve in such a way that everything works reasonably well. I've been trying to do this for the past few days on an off, and this implementation is the best I've managed to pull off :(

PR #1253, opened 2017-05-20
INSP: Convert some annotations into inspections

### LINKED ISSUE #1243: error_chain's Result<T> support
When using the error_chain crate it's common to define the errors in a "errors.rs" mod and import them where needed with `use errors::*`.

This overrides the default `Result<T,E>` enum with a `Result<T>`, and intellij-rust reports all the functions using it as errors (Wrong number of arguments: expected 2, found 1 [E0243]).

I know it's hard to parse macros (see #955), but I was wondering if there is a way to use annotations to declare the definition of enums, structs or functions (like `// @defines Result<T>` or similar) to use in the mod defining them, so that the parser recognizes them.

P.S.: sorry for the empty bug report earlier, I don't know what I pressed to post it while writing... :)

alygin:
Transforms E0243/E0244 (wrong type parameters number) and E0106/E0107 (wrong lifetime parameters number) annotations into two inspections.

Closes #1243


matklad:
bors r+

Thanks!

PR #1257, opened 2017-05-21
Improve primitive type autocompletion

kumbayo:
CC #371

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -24,7 +24,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
         // Bump this number if Stub structure changes
-        override fun getStubVersion(): Int = 70
+        override fun getStubVersion(): Int = 71
 
         override fun getBuilder(): StubBuilder = object : DefaultStubBuilder() {
             override fun createStubForFile(file: PsiFile): StubElement<*> = RsFileStub(file as RsFile)
@@ -715,6 +715,7 @@ class RsTypeReferenceStub(
     parent: StubElement<*>?, elementType: IStubElementType<*, *>,
     val isMut: Boolean,
     val isRef: Boolean,
+    val isPointer: Boolean,
     val arraySize: Int
 ) : StubBase<RsTypeReference>(parent, elementType) {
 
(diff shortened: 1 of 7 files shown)


kumbayo:
This does not solve all primitive type related problems but is a start.

The current state of primitive type autocompletion:
* No autocompletion for methods on u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, isize
  as they are implemented by macros.
* No support for UFCS on primitive types (i have some WIP code for this)
* *const and *mut tests fail because TyPointer.canUnifyWith does not check mutability yet
  and we get 2 methods instead of one (i have some WIP code for this)
* No autocompletion on arrays as they do not coerce to slices yet

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
```diff
@@ -82,7 +82,7 @@ fun findImplsAndTraits(project: Project, ty: Ty): Pair<Collection<BoundElement<R
 
     //  XXX: TyStr is TyPrimitive, but we want to handle it separately
         is TyStr -> RsImplIndex.findImpls(project, ty).map { impl -> BoundElement(impl) } to noTraits
-        is TyPrimitive, is TyUnit, is TyUnknown -> noImpls to noTraits
+        is TyUnit, is TyUnknown -> noImpls to noTraits
```
kumbayo: I am unsure about this part.
My gut feeling says the following:
We should use the TyStr path for all primitive types.
We should use the else path for TyPointer/TyReference since they need to resolve type parameters.

I also think for TyPointers the type parameter machinery from TyStructOrEnum needs to be adapted/generalized.
Because currently the following code does not resolve types
```rust
let my_ptr: *const char = 42 as *const _;
let new_ptr = my_ptr.offset(42); // evaluates to *const T instead of *const char
```

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -24,7 +24,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
         // Bump this number if Stub structure changes
-        override fun getStubVersion(): Int = 69
+        override fun getStubVersion(): Int = 71
```
farodin91: Why you increase the stub version up 71 not to 70?

kumbayo:
This view shows 2 commits squashed together.Both commits increment the number by 1 so they are testable independently.The first commit changes what inpls are indexed. The second commit changes what is stored in the index.

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
matklad: Yeah, an arbitrary increment is fine here, we probably won't run out of Ints in the nearest future :)

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
matklad: >We should use the TyStr path for all primitive types.

That is correct. 

>I also think for TyPointers the type parameter machinery from TyStructOrEnum needs to be adapted/generalized.

Yep!

matklad:
Looks great!  Do you want to work on those `except<IllegalStateException>` test in this PR or are they left for future work? Both options are fine!

kumbayo:
I think those improvements will take me some time, so it might be a good idea to merge this part now.

matklad:
bors r+

Thanks!

PR #1260, opened 2017-05-22
ANN: Fix E0384 false positive

alygin:
Fixes an E0384 false positive when a mutable static constant is reassigned. Also a minor refactoring.

Though technically, it's not a E0384, I think we can still handle it here because Rust has no designated code for this case.

cc @farodin91


farodin91:
bors r+

Looks great!

alygin:
bors is very strict )

bors r+

notriddle:
BTW, @alygin, you can use r=SOME_USER to approve a commit "on behalf of" someone else. It's a decent way to do "reviewer internships."

PR #1263, opened 2017-05-24
TY: improve type inference of function call expr

Undin:
Improve type inference of function call expressions. Now we can infer type parameter not only directly:
```rust
fn foo<T>(arg: T) -> T { ... }

let x = foo(123) // x: i32
```
but also from more complex generic arguments:
```rust
fn foo<T>(arg: (T, T)) -> [T; 2] { ... }

let x = foo((1, 2)); // x: [i32; 2]
``` 
Slightly modified `canUnifyWith` methods are used to achieve it.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
@@ -215,17 +215,8 @@ private fun mapTypeParameters(
     return argsMapping
 }
 
-private fun addTypeMapping(
-    argsMapping: MutableMap<TyTypeParameter, Ty>,
-    fieldType: Ty?,
-    expr: RsExpr
-) {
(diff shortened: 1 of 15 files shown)


matklad:
Nice! One step close to proper unification algorithm!

bors r+

matklad:
>Nice! One step close to proper unification algorithm!

The next step should be pretty biggish I think. Currently, we infer types by looking at a single expression only, and this is wrong, because types of the following expressions may influence types of the preceding ones.  (`let mut x = Vec::new(); x.push(92i32)`).

What we need to do is to infer types of all expressions in a function simultaneously. That is, for each function `fn` we should have something like `fn.exprTypes: Map<RsExpr, Ty>`, and we should populate this map in one go by traversing the tree of expressions inside the function body. 

cc @vlad20012

PR #1267, opened 2017-05-25
TY: initial support of arithmetic binary expression type inference

Undin:
Infer type of arithmetic binary expression.

Now inference doesn't work in following cases:
* for primitive types because their implementations are created with macros
* for generic types when `Output = Self`:
```rust
impl<T> Add for Foo<T> {
    type Output = Self;
    fn add(self, rhs: Self) -> Self { ... }
}
```
* if `Output` is generic and its type parameters should be inferred from `RHS` type:
```rust
impl<T1, T2> Add<Bar<T2>> for Foo<T1> {
    type Output = FooBar<T1, T2>;
    fn add(self, rhs: Bar<T2>) -> FooBar<T1, T2> { ... }
}
```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsExpr.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsExpr.kt
@@ -1,7 +1,6 @@
 package org.rust.lang.core.psi.ext
 
 import com.intellij.psi.PsiElement
-import com.intellij.psi.tree.IElementType
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.*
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsExpr.kt
```diff
@@ -29,13 +29,67 @@ val RsUnaryExpr.operatorType: UnaryOperator get() = when {
...
+    operator fun component1(): String = traitName
+    operator fun component2(): String = itemName
+    operator fun component3(): String = sign
+}
+
 val RsBinaryExpr.operator: PsiElement
     get() = requireNotNull(node.findChildByType(BINARY_OPS)) { "guaranteed to be not-null by parser" }.psi
 
 // TODO: probably want to use a special `enum` here instead of `IElementType`.
 val RsBinaryExpr.operatorType: IElementType
```
matklad: Let's get rid of `operatorType` and use `arithmeticOp` instead?

Undin: Unfortunately, `operatorType` is used to support not only arithmetic binary expressions, so it's not very convenient to replace `operatorType` with some enum. But I can create sealed class and split it into some subtypes grouped by type of operation: arithmetic, logic, comparison, etc. What do you think about it?

matklad: Hm, Indeed. Yeah, let's create a sealed class then: having two similar ways to do the same thing is a recipe for problems.

Undin:
Done

matklad:
Thanks!

bors r+

PR #1271, opened 2017-05-26
TY: tuple field type inference

### LINKED ISSUE #1269: No type is shown for tuple elements accessed by index
image

Undin:
Fixes #1269

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
@@ -52,7 +52,11 @@ fun inferExpressionType(expr: RsExpr): Ty {
 
         is RsFieldExpr -> {
             val boundField = expr.reference.advancedResolve()
-                ?: return TyUnknown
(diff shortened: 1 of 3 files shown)


vasily-kirichenko:
Wow. That was fast :)

vasily-kirichenko:
I have a long list of other snippets where types are not inferred. Should I create issues or you guys are already aware of all the restrictions?

Undin:
I think @matklad can answer this question

Undin:
But I personally think that it's a good idea to have list with all (maybe not all, but the most part) of language structs where type inference doesn't work

matklad:
Looks good @Undin , but perhaps we can handle tuple structs here as well?

matklad:
Nevermind, they are already handled here :)

matklad:
@vasily-kirichenko we are aware that a lot of stuff just does not work, but having small isolated snippets would be super useful! I've created a meta-issue for this: https://github.com/intellij-rust/intellij-rust/issues/1272

PR #1279, opened 2017-05-27
Typification improvements

kumbayo:
The 2 main improvements are:
* Ranges ".." and "..." are typified to ops::Rangexxx
* The macros vec![] and format!("") are typified to vec::Vec<T> and String

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
@@ -117,12 +117,9 @@ fun inferExpressionType(expr: RsExpr): Ty {
 
         is RsTryExpr -> {
             val base = expr.expr.type
-            // This is super hackish. Need to figure out how to
-            // identify known ty (See also the CString inspection).
-            // Java uses fully qualified names for this, perhaps we
-            // can do this as well? Will be harder to test though :(
-            if (base is TyEnum && base.item.name == "Result")
-                base.typeArguments.firstOrNull() ?: TyUnknown
+
+            if (isStdResult(base))
+                (base as TyEnum).typeArguments.firstOrNull() ?: TyUnknown
             else
                 TyUnknown
         }
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
@@ -116,6 +116,7 @@ fun inferExpressionType(expr: RsExpr): Ty {
         }
 
         is RsTryExpr -> {
+            // See RsMacroExpr where we handle the try! macro in a similar way
             val base = expr.expr.type
 
             if (isStdResult(base))
@@ -166,6 +167,52 @@ fun inferExpressionType(expr: RsExpr): Ty {
             val containerType = expr.containerExpr?.type ?: return TyUnknown
             val indexType = expr.indexExpr?.type ?: return TyUnknown
             findIndexOutputType(expr.project, containerType, indexType)
+        }
+
+        is RsMacroExpr -> {
+            if (expr.vecMacro != null) {
+                val elements = expr.vecMacro!!.vecMacroArgs?.exprList ?: emptyList()
+                var elementType: Ty = TyUnknown;
+                for (e in elements) {
+                    elementType = getMoreCompleteType(e.type, elementType);
+                }
+
+                findStdVec(elementType, expr)
+            } else if (expr.logMacro != null) {
+                TyUnit
+            } else if (expr.macro != null) {
+                val macro = expr.macro ?: return TyUnknown;
+                when (macro) {
+                    is RsTryMacro -> {
+                        // See RsTryExpr where we handle the ? expression in a similar way
+                        val base = macro.tryMacroArgs?.expr?.type ?: return TyUnknown;
+
+                        if (isStdResult(base))
+                            (base as TyEnum).typeArguments.firstOrNull() ?: TyUnknown
+                        else
+                            TyUnknown
+                    }
+
+                    is RsFormatLikeMacro -> {
+                        val name = macro.macroInvocation.referenceName;
+
+                        if (name == "format")
+                            findStdString(expr)
+                        else if (name == "format_args")
+                            findStdArguments(expr)
+                        else if (name == "write" || name == "writeln")
+                            TyUnknown //Returns different types depending on first argument
+                        else if (name == "panic" || name == "print" || name == "println")
+                            TyUnit
+                        else
+                            TyUnknown
+                    }
+                    is RsAssertMacro,
+                    is RsAssertEqMacro -> TyUnit
+
+                    else -> TyUnknown
+                }
+            } else return TyUnknown;
         }
 
         else -> TyUnknown
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
@@ -238,8 +238,13 @@ private fun inferArrayType(expr: RsArrayExpr): Ty {
     } else {
         val elements = expr.arrayElements
         if (elements.isNullOrEmpty()) return TySlice(TyUnknown)
+
+        var elementType: Ty = TyUnknown;
         // '!!' is safe here because we've just checked that elements isn't null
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -142,6 +185,27 @@ fun inferExpressionType(expr: RsExpr): Ty {
...
+    val module = elementForModule.module ?: return TyUnknown;
+    val crateRoot = elementForModule.crateRoot as? RsFile ?: return TyUnknown
+    val prefix = if (crateRoot.attributes == RsFile.Attributes.NONE) prefixStd else prefixNoStd;
+    val fullName = prefix + "::" + name;
+
+    val (element, _) = org.rust.lang.core.resolve.resolveStringPath(fullName, module) ?: return TyUnknown;
+    return (element as? RsTypeBearingItemElement)?.type ?: TyUnknown;
+}
+
+private fun getMoreCompleteType(t1: Ty, t2: Ty): Ty {
```
kumbayo: This is kind of limited and hacky but provides a nice type when only the first or second argument of a range have a concrete type.
e.g.
```rust
let len: usize = xxx;
let range = 0..len;
   //^ Range<usize>
```

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -114,7 +114,7 @@ fun inferExpressionType(expr: RsExpr): Ty {
             RsElementTypes.GT,
             RsElementTypes.GTEQ,
             RsElementTypes.LTEQ -> TyBool
-
+            RsElementTypes.EQ -> TyUnit
```
kumbayo: This will (trivially) conflict with #1267

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -267,8 +267,13 @@ private fun inferArrayType(expr: RsArrayExpr): Ty {
     } else {
         val elements = expr.arrayElements
         if (elements.isNullOrEmpty()) return TySlice(TyUnknown)
+
+        var elementType: Ty = TyUnknown;
         // '!!' is safe here because we've just checked that elements isn't null
-        elements!![0].type to elements.size
+        for (e in elements!!) {
```
kumbayo: Do you think it is OK if we calculate the type for all array/vec elements?
Or should be an upper bound (e.g. only look at the first 20 array elements, quit after the first full type?)

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -132,16 +133,132 @@ fun inferExpressionType(expr: RsExpr): Ty {
...
+                        return TyUnknown
+                    } else {
+                        "RangeToInclusive" to e.type
+                    }
+                }
+                dot3 != null && el.size == 2 -> {
+                    "RangeInclusive" to getMoreCompleteType(el[0].type, el[1].type)
+                }
+
+                else -> return TyUnknown //Should never happen
```
matklad: I usually prefer to use `error` for things that should never happen, for once they do happen, it's much easier to discover and fix :)

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -132,16 +133,132 @@ fun inferExpressionType(expr: RsExpr): Ty {
...
 
+        is RsMacroExpr -> {
+            if (expr.vecMacro != null) {
+                val elements = expr.vecMacro!!.vecMacroArgs?.exprList ?: emptyList()
+                var elementType: Ty = TyUnknown;
+                for (e in elements) {
+                    elementType = getMoreCompleteType(e.type, elementType);
+                }
+
+                val ty = tyFromAbsolutePath("std", "collections", "vec::Vec", expr);
```
matklad: We have `langItems` file for various "well-known" types (not only literally lang-items). Let's move these there as well? 

I think that `tyFromAbsolutePath` could be a `private` implementation detail, and the API would be something like `findStdVec(m: Module)`.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -132,16 +133,132 @@ fun inferExpressionType(expr: RsExpr): Ty {
...
+            } else if (expr.macro != null) {
+                val macro = expr.macro ?: return TyUnknown;
+                when (macro) {
+                    is RsTryMacro -> {
+                        val base = macro.tryMacroArgs?.expr?.type ?: return TyUnknown;
+                        // This is super hackish. Need to figure out how to
+                        // identify known ty (See also the CString inspection).
+                        // Java uses fully qualified names for this, perhaps we
+                        // can do this as well? Will be harder to test though :(
+                        // See RsTryExpr where we handle the ? expression
```
matklad: Yeah, we need to follow similar approach (create a sort of registry of well-known types, and query it using crate-qualified paths), the only thing different is that we don't have truly fully qualified names, so we'll need to actually use resolver to do this, and that's why we probably need to cache it on psi modification count. This should go to `LangItems.kt`. We don't need to implement registry with caching yet, but it's nice to keep all similar stuff together.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
matklad: No, this is precisely how things should work :)

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
matklad: An interesting question. I think it's OK to inspect all the elements of the array. Of course there are some awful code-generated cases with huge arrays which can hit us hard, but we probably will calculate the types of elements anyway for highlighting.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
kumbayo: I could not find a way to get the top level RsFile from a Module so i am still passing a RsCompositeElement for getting the Module and also the RsFile.
Do you have an idea on how to do this?

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
kumbayo: I changed this part to also cover other assignments like +=, -=, ...

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
kumbayo: In this case i just copied the code+comment that was already available in the handling of the ? expression and reused it for try!.

I am not sure how to properly do it. Do you envision something like LangItems.isRange(type: Ty): Boolean or something like LangItems.isType(type: Ty, typeName: String): Boolean

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
kumbayo: Done

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
matklad: Yeah, the `?` was implemented before the `LangItems`. 

>I am not sure how to properly do it. Do you envision something like LangItems.isRange(type: Ty): Boolean or something like LangItems.isStdType(type: Ty, typeName: String): Boolean

Something like `fun isStdResult(ty: Ty)`. I don't think we need a super generic `typeName: String` variant, because there are only so many interesting types in Rust. The precise interface probably does not matter much at this stage: I think we can provide a nice API later, once we've collected all the use-cases for it in `LangItems.kt`.

kumbayo: Done

matklad:
bors r+

Thanks!

Undin:
Looks like bors is hanging

notriddle:
@matklad You should be able to r+ it again.

matklad:
bore r+

Undin:
I guess bore -> bors :)

PR #1282, opened 2017-05-28
TYPE: fold doc comments

vasily-kirichenko:

### DIFF
--- a/src/main/kotlin/org/rust/ide/folding/RsFoldingBuilder.kt
+++ b/src/main/kotlin/org/rust/ide/folding/RsFoldingBuilder.kt
@@ -18,6 +18,7 @@ import org.rust.lang.RsLanguage
 import org.rust.lang.core.leftLeaves
 import org.rust.lang.core.leftSiblings
 import org.rust.lang.core.parser.RustParserDefinition.Companion.BLOCK_COMMENT
+import org.rust.lang.core.parser.RustParserDefinition.Companion.OUTER_EOL_DOC_COMMENT
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.LBRACE
 import org.rust.lang.core.psi.RsElementTypes.RBRACE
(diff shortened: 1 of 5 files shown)


matklad:
Great! Let's also handle INNER_EOL_DOC_COMMENT and add test as well ;)

There's org.rust.ide.folding.RsFoldingTest for it, and it had a `testBlockComment` method already. So let's rename the method into `testComments`, rename `test/resources/org/rust/ide/folding/fixtures/block_comment.rs` file into `test/resources/org/rust/ide/folding/fixtures/commenst.rs` and make it's contents more flavorful, something like

```
/*
hello
*/

///
/// Otter doc
///
fn foo() {}

mod m {
    //!
    //! Inner doc
    //! 
}
```,

with `<fold text='/* ... */'></fold>` tags enclosing the folds.

**Comment by matklad** [comment]
Oh, while we are at it, could you rename `RsFoldingTest` into `RsFoldingBuilderTest` as well, so that **Goto test** (`ctrl+shift+t`) action works in the IDE?

File: src/main/kotlin/org/rust/ide/folding/RsFoldingBuilder.kt [hunk 1]
```diff
@@ -129,7 +131,8 @@ class RsFoldingBuilder : FoldingBuilderEx(), DumbAware {
     }
 
     override fun isCollapsedByDefault(node: ASTNode): Boolean =
-        RsCodeFoldingSettings.instance.collapsibleOneLineMethods && node.elementType in COLLAPSED_BY_DEFAULT
+        (RsCodeFoldingSettings.instance.collapsibleOneLineMethods && node.elementType in COLLAPSED_BY_DEFAULT)
+            || (CodeFoldingSettings.getInstance().COLLAPSE_DOC_COMMENTS && node.elementType == OUTER_EOL_DOC_COMMENT)
```
matklad: :+1:, didn't think of that!

vasily-kirichenko:
Done :) I'm not sure the test content is what it should be. I just copy pasted it from test failure diff.

matklad:
Thanks!  (And especially for the GIF, the make changelogs so much lively!)

bors r+

>I'm not sure the test content is what it should be. I just copy pasted it from test failure diff.

Yeah, this is precisely how these sort of test work: you make sure that code works correctly by manually checking that it does the right thing when opening a rust file, and then you use the tests just to "freeze" this behavior.

vasily-kirichenko:
Fixed the autoformatted test and renamed `RsFoldingTest`.

matklad:
:+1: 

bors r+

PR #1283, opened 2017-05-28
Intelligently complete return with `;` or ` `

matklad:
r? @alygin

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionProvider.kt
@@ -8,6 +8,8 @@ import com.intellij.util.ProcessingContext
 import org.rust.lang.core.completion.CompletionEngine.KEYWORD_PRIORITY
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.ext.parentOfType
+import org.rust.lang.core.psi.ext.returnType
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionProvider.kt
```diff
@@ -36,3 +27,20 @@ class AddSuffixInsertionHandler(val suffix: String) : InsertHandler<LookupElemen
...
+private val ALWAYS_NEEDS_SPACE = setOf("crate", "const", "enum", "extern", "fn", "impl", "let", "mod", "mut", "pub",
+    "static", "struct", "trait", "type", "unsafe", "use")
+
+
+private fun addInsertionHandler(keyword: String, builder: LookupElementBuilder, parameters: CompletionParameters): LookupElementBuilder {
+    val suffix = when (keyword) {
+        in ALWAYS_NEEDS_SPACE -> " "
+        "return" -> {
+            val fn = parameters.position.parentOfType<RsFunction>() ?: return builder
+            if (fn.retType != null) " " else ";"
```
alygin: Inside `fn foo() -> ()` it could also be completed with `;`

Diverging functions probably also need a special case (don't suggest `return` at all), but that may be a subject for a separate PR.

matklad: Good idea!

matklad:
bors r=alygin

PR #1288, opened 2017-05-28
INT: do not show "remove unnecessary parens" for struct literals

### LINKED ISSUE #1285: Wrong suggestion to remove parens around struct construction literal at `match ..` site

vasily-kirichenko:
close #1285

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
@@ -8,13 +8,7 @@ import com.intellij.util.SmartList
 import org.rust.ide.annotator.fixes.AddStructFieldsFix
 import org.rust.ide.intentions.RemoveParenthesesFromExprIntention
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsFieldsOwner
-import org.rust.lang.core.psi.ext.RsReferenceElement
-import org.rust.lang.core.psi.ext.namedFields
-import org.rust.lang.core.psi.ext.queryAttributes
-import org.rust.lang.core.psi.ext.RsStructKind
-import org.rust.lang.core.psi.ext.kind
-import org.rust.lang.core.psi.impl.RsStructLiteralImpl
+import org.rust.lang.core.psi.ext.*
 import java.util.*
 
 class RsExpressionAnnotator : Annotator {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
```diff
@@ -81,8 +82,12 @@ private class RedundantParenthesisVisitor(private val holder: AnnotationHolder)
 
     private fun RsExpr?.warnIfParens(message: String) {
         if (this is RsParenExpr) {
-            holder.createWeakWarningAnnotation(this, message)
-                .registerFix(RemoveParenthesesFromExprIntention())
+            when (this.children.singleOrNull()) {
```
matklad: I think this overshots a bit. Sometimes is totally legit to remove parenthesis around struct literals:

```
let s = (S { foo: 92 });
```

We need to check that the parent of the exression is a conditon/match target.

src/main/kotlin/org/rust/ide/intentions/RemoveParenthesesFromExprIntention.kt
```diff
@@ -5,13 +5,20 @@ import com.intellij.openapi.project.Project
...
     override fun getText(): String = "Remove parentheses from expression"
     override fun getFamilyName(): String = text
 
-    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsParenExpr? =
-        element.parentOfType<RsParenExpr>()
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsParenExpr? {
+        val parenExpr = element.parentOfType<RsParenExpr>()
+
+        return when (parenExpr?.children?.singleOrNull()) {
+            null, is RsStructLiteralImpl -> null
```
matklad: Similarly here, we want' to disable intention only in conditionals.

src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
```diff
@@ -79,10 +74,17 @@ private class RedundantParenthesisVisitor(private val holder: AnnotationHolder)
...
+        o.expr.warnIfParens("Redundant parentheses in let declaration")
+    }
+
     private fun RsExpr?.warnIfParens(message: String) {
         if (this is RsParenExpr) {
-            holder.createWeakWarningAnnotation(this, message)
-                .registerFix(RemoveParenthesesFromExprIntention())
+            if ((this.parentOfType<RsCondition>() != null || this.parentOfType<RsMatchExpr>() != null)
+                && this.children.singleOrNull() is RsStructLiteral) {
+            } else
```
matklad: That's one puzzling `if`, with an empty `then` branch, and multiline `else` branch without braces :) 

I wonder if we can simplify this a bit, and remove duplication with the intention as well? Something like this should work, perhaps? 

```Kotlin
class RemoveParenthesesFromExprIntention {
  fun isAvailable(expr: RsParanExpr): Boolean { /* check that we can safely remove parenthesis */  }
  override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsParenExpr? {
    val parenExpr = element.parentOfType<RsParenExpr>()
    return if (isAvailable(parenExpr))  parenExpr else null
  }
}

// In the annotator
 private fun RsExpr?.warnIfParens(message: String) {
    if (this !is RsParenExpr) return
    val fix = RemoveParenthesesFromExprIntention()
    if (fix.isAvailable(this)) holder.createWeakWarningAnnotation(...)
}

```

vasily-kirichenko:
Thanks for the review! All done.

matklad:
LGTM!

bors r+

PR #1289, opened 2017-05-28
INSP: Borrow checker and Reassign Immutable inpsections

### LINKED ISSUE #1266: "Assign immutable" error annotation should not warn by default
[Assign immutable](https://github.com/intellij-rust/intellij-rust/blob/f04bee83267c275f146ada5992f3c62ebec226ff/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt#L489) error annotation is known to produce false positives: https://gitter.im/intellij-rust/intellij-rust?at=5927204f2b926f8a67919709

The fundamental problem with this annotation is that it assumes "immutable" by default, so if we are unable to infer some types, it causes an FP. We should switch the logic to not warn by default.

alygin:
This one makes the E0384 (re-assing immutable variable) and "cannot borrow immutable as mutable" detection safer:

1. The related error annotations are extracted into two separate inspections &mdash; "Re-assign immutable variable" (an [E0384](https://doc.rust-lang.org/error-index.html#E0384) analogue) and "Borrow checker" &mdash; to make them optional.
2. The part of the mutability analysis that led to various false positives (including the one from #1266) is commented out. Effectively, it removes those false positives, though brings some false negatives.

In this fix, I wasn't aiming to change the general approach to the existing mutability analysis. The majority of the problems with it (actually, all the problems that I'm aware of) stemmed from a naive approach in one particular place related to structs and their fields/methods, that I removed (so it defaults to "mutable" now). We can come back to it later and try to implement a deeper analysis that doesn't bring false positives. Until then, I've marked broken tests as `@Ignored`.

Closes #1266


src/test/kotlin/org/rust/ide/inspections/RsBorrowCheckerInspectionTest.kt
```diff
@@ -0,0 +1,292 @@
...
+    fun `test simple enum variant is treated as mutable`() = checkByText("""
+        enum Foo { FOO }
+        fn foo (f: &mut Foo) {}
+        fn bar () {
+            foo(&mut Foo::FOO);     // Must not be highlighted
+        }
+    """)
+
+    @Ignore
+    fun `test fix let at method call (self)`() = checkFixByText("Make `test` mutable", """
```
matklad: Let's not `@Ignore` the tests, but change the expected result with a comment that it's a false negative. This way, we'll notice when it is fixed, and the Ignored tests could be just forgotten.

matklad: The most interesting fact is that all `@Ignore` tests actually do work in this PR :)

matklad:
Thanks @alygin ! I am merging this in https://github.com/intellij-rust/intellij-rust/pull/1379!

PR #1291, opened 2017-05-28
PSI: error_chain macro expansion (WIP)

alygin:
I've drafted an implementation of the `error_chain` macro expansion, but need a review before moving further.

The simple case of the expansion (with predefined names) works fine, but needs several tweaks. The main problems I see now are:
- The expansion is not stubbed, so resolving is not as fast as it could be. It also breaks unit tests because of the fallbacks to the PSI tree. I don't know what would be the right approach to stubbing these implied items. Should the macro invocation be stubbed with it's `expansion` property?
- When I change something inside the `error_chain` macro invocation, it doesn't affect resolving immediately. For instance, if I have an `error_chain! {}` invocation and then use `Result<T>`, this `Result<T>` resolves correctly. Then, if I make some changes: `error_chain! { types {...} }`, the `Result<T>` is still resolved, though the macro is not expanded anymore. Reopening the file solves the problem. Maybe it relates to the previous problem with stubs. Any ideas on that?

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -961,7 +961,9 @@ LitExpr ::= STRING_LITERAL | BYTE_STRING_LITERAL
 // Macros
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 
-fake Macro ::= MacroInvocation MacroArg?
+fake Macro ::= MacroInvocation MacroArg? {
+  mixin = "org.rust.lang.core.psi.ext.RsMacroImplMixin"
+}
 
 // Various kinds of macros
 ExprLikeMacro ::= MacroInvocation MacroArg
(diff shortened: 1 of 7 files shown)


matklad:
>The expansion is not stubbed, so resolving is not as fast as it could be. 

I think fully stubbing/caching macros would be very difficult, because macro definition and macro invocation may be in different files, so the usual indexing infrastructure won't help much. 

But I don't think we need to do this: I expect that macro invocations to be relatively rare. So it's totally OK if new tests use AST. I think that some old tests may fail though because of macro invocations in the standard library. To deal with this, I think we can store the name of the macro incokaction in the index, and skip expansion step altogether, if it is not the error chain. 

>When I change something inside the error_chain macro invocation, it doesn't affect resolving immediately.

I think this happens because we create expansion in the constructor: https://github.com/intellij-rust/intellij-rust/pull/1291/files#diff-2371a04722cf0161dc1e4979a9851485R14

PSI is mutable, so node's children may change even if the node stays the same, which invalidates the expansion. The simplest thing to do would be to compute expansion on the fly, and cache it on `PsiModificationTracker.MODIFICATION_COUNT`, how it's done in `org/rust/lang/core/types/Extensions.kt` and in `org.rust.lang.core.resolve.indexes.RsImplIndex`.

matklad:
Closing due to inactivity. We've now made a first step towards macro expansion though: https://github.com/intellij-rust/intellij-rust/pull/1974

PR #1300, opened 2017-05-30
Less obtrusive missing struct fields error highlighting

### LINKED ISSUE #1250: Too angry when missing a field in a struct
Until the very last field is added when instantiating a struct, there is no coloring / syntax highlighting of the fields you have filled in. Everything is an angry red squiggle telling you you've not completed all of it.

It would be great if it could give better support for not totally filled in structs as this is quite a common occurrence when coding. Maybe the red warning could be just be on the struct name rather than the full contents of the struct? (Apologies if what I'm asking for is very difficult).

(Also if you're in the area, if you autocomplete setting a struct member when instantiating  a struct it would be great if it also added the ': ' by default - that would be super slick)

vasily-kirichenko:
Close #1250


matklad:
Thanks!

bors r+

Issue #640, opened 2016-08-30
Show type of variable when Ctrl-Hover


johnthagen:
When Ctrl-Hovering, it would be neat if you could see the type of the variable if that is available to `intellij-rust`.

![rust-ctrl-hover](https://cloud.githubusercontent.com/assets/10340167/18107583/7c26a3f0-6ed6-11e6-9edb-8d98dedb42e2.png)

PyCharm does something like this for Python code:

![pycharm-type-hover](https://cloud.githubusercontent.com/assets/10340167/18107667/e4999d3e-6ed6-11e6-87a1-8ff6b18c3577.png)

kindlychung:
Scala plugin does this with shift+ctrl+p, which I like a lot (better than ctrl hovering).

matklad:
@kindlychung we already have `Shift+Ctrl+p` functionality, though type inference itself is very poor at the moment, so it's not very usable.

johnthagen:
@kindlychung  Thanks for the tip, hadn't used `Shift+Ctrl+P` before.

Currently the types are shown for functions with `Ctrl+Hover`, so it seems like it would be a natural addition for variables as well.  If they type inference is poor for a while, seems fine to me as that's a separate effort.

the8472:
In addition to variables it would be even better if the information were available in chained calls. For a beginner it is quite helpful to know what the type of some intermediate result is so one can call the right transformations on it.

![eclipse-tooltip](https://cloud.githubusercontent.com/assets/1065730/19747257/e9f20970-9bda-11e6-875b-00f154d56094.png)

matklad:
String displayed by hover is controlled by `org.rust.ide.docs.RsDocumentationProvider#getQuickNavigateInfo`.

PedroD:
https://stackoverflow.com/questions/54851861/how-do-i-activate-type-annotations-hints-in-kotlin-like-depicted/54852497#54852497

PR #1302, opened 2017-05-30
Show the type of variable on hover

### LINKED ISSUE #1217: Show the type of variable on hover
vscode has this feature and it's very useful for determining the type of a variable.

Undin:
Сloses #640
Closes #1217

It temporary breaks potentially useful quick doc in some cases where we can't infer variable type, for example, tuple destructuring in function declaration, but it should be fixed from type inference side.


vasily-kirichenko:
This is great. But the mouse cursor is blinking as it's moved:

Undin:
Yeah, I can reproduce it too, but it happens not only with Rust code, so I think it's global problem.

matklad:
bors r+

:+1:

Undin:
Hm, #640 should be closed too.
@matklad do you know how to say bors to close several issues after PR merging?

jrmuizel:
This doesn't seem to work for me in IntelliJ 2017.1.4 on OS X. I don't get any popup when I hover over struct1 using the example above.

PR #1305, opened 2017-05-31
INSP: Extend wrong type parameters inspection and add a simple fix.

farodin91:
#886

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsWrongTypeParametersNumberInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsWrongTypeParametersNumberInspection.kt
@@ -6,12 +6,13 @@
 package org.rust.ide.inspections
 
 import com.intellij.codeInspection.ProblemsHolder
-import org.rust.lang.core.psi.RsBaseType
-import org.rust.lang.core.psi.RsVisitor
+import org.rust.ide.inspections.fixes.RemoveTypeParameter
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsCompositeElement
 import org.rust.lang.core.psi.ext.RsGenericDeclaration
 
 /**
- * Inspection that detects E0243/E0244 errors.
+ * Inspection that detects E0243/E0244/E0087/E0089/E0035/E0036 errors.
  */
 class RsWrongTypeParametersNumberInspection : RsLocalInspectionTool() {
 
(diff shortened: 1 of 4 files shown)


farodin91:
#886

src/main/kotlin/org/rust/ide/inspections/RsWrongTypeParametersNumberInspection.kt
```diff
@@ -16,18 +17,74 @@ class RsWrongTypeParametersNumberInspection : RsLocalInspectionTool() {
                 // Don't apply generic declaration checks to Fn-traits and `Self`
                 if (type.path?.valueParameterList != null) return
                 if (type.path?.cself != null) return
+                checkMethod(holder, type)
+            }
+
+            override fun visitCallExpr(o: RsCallExpr) {
+                checkMethod(holder, o)
+            }
```
alygin: Could be a one-liner:
`override fun visitCallExpr(o: RsCallExpr) = checkMethod(holder, o)`

src/main/kotlin/org/rust/ide/inspections/RsWrongTypeParametersNumberInspection.kt
```diff
@@ -16,18 +17,74 @@ class RsWrongTypeParametersNumberInspection : RsLocalInspectionTool() {
...
+            is RsCallExpr -> {
+                val pathExpr = o.expr as RsPathExpr?
+                pathExpr?.path?.typeArgumentList?.typeReferenceList?.size ?: 0
+            }
+            is RsBaseType -> {
+                o.path?.typeArgumentList?.typeReferenceList?.size ?: 0
+            }
+            else -> {
+                0
+            }
```
alygin: The same thing here and below, could be much shorter:
```rust
val actualArgs = when (o) {
    is RsMethodCallExpr -> o.typeArgumentList?.typeReferenceList?.size ?: 0
    is RsCallExpr -> (o.expr as? RsPathExpr)?.path?.typeArgumentList?.typeReferenceList?.size ?: 0
    is RsBaseType -> o.path?.typeArgumentList?.typeReferenceList?.size ?: 0
    else -> 0
}
```

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveTypeParameter.kt
```diff
@@ -0,0 +1,37 @@
...
+            }
+        } ?: return
+        val lifetime = typeArgumentList.lifetimeList.size
+        if (lifetime == 0) {
+            typeArgumentList.delete()
+        }
+    }
+
+    override fun getName() = "Remove all type parameter"
+    override fun getFamilyName() = name
```
alygin: We usually put properties at the beginning of the class, before methods, so it would be easier to find them.
Also a typo: parameter**s**

farodin91:
@alygin Resolved!

src/main/kotlin/org/rust/ide/inspections/RsWrongTypeParametersNumberInspection.kt
```diff
@@ -16,18 +17,51 @@ class RsWrongTypeParametersNumberInspection : RsLocalInspectionTool() {
...
             }
+
+            override fun visitCallExpr(o: RsCallExpr) = checkMethod(holder, o)
+            override fun visitMethodCallExpr(o: RsMethodCallExpr) = checkMethod(holder, o)
+        }
+
+    private fun checkMethod(holder: ProblemsHolder, o: RsCompositeElement) {
+        val actualArgs = when (o) {
+            is RsMethodCallExpr -> o.typeArgumentList?.typeReferenceList?.size ?: 0
+            is RsCallExpr -> (o.expr as RsPathExpr?)?.path?.typeArgumentList?.typeReferenceList?.size ?: 0
```
alygin: Did you mean `o.expr as? RsPathExpr`? Otherwise this can produce ClassCastExceptions.

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveTypeParameter.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.openapi.project.Project
+import org.rust.lang.core.psi.RsBaseType
+import org.rust.lang.core.psi.RsCallExpr
+import org.rust.lang.core.psi.RsMethodCallExpr
+import org.rust.lang.core.psi.RsPathExpr
+import org.rust.lang.core.psi.ext.RsCompositeElement
+
+class RemoveTypeParameter(val element: RsCompositeElement) : LocalQuickFix {
```
alygin: All fix classes in the plugin have `Fix` suffix. Let's stick to this naming convention.

alygin: And again, given what it does, it'll be "Parameter**s**", the plural form.

farodin91:
@matklad Any suggestions?

matklad:
LGTM, thanks @farodin91  and @alygin !

bors r+

farodin91:
@matklad I need to add a MIT header.

farodin91:
@matklad  resolved

matklad:
@farodin91 hm, looks like there are still some false positives left:

```Rust
fn foo<U, V>(u: U, v: V) {}

fn main() {
    foo(1, 2); // this is annotated as an error, but it shouldn't be
```

I think we need to add annotation only if there's at least one parameter. Will you send a PR with a fix?

farodin91:
I could send an PR. I will start this evening.

PR #1306, opened 2017-05-31
ANNOT: Detect wrong safe or unsafe impls E0199/E0200

farodin91:
#886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -309,8 +309,14 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     }
 
     private fun checkImpl(holder: AnnotationHolder, impl: RsImplItem) {
-        val trait = impl.traitRef?.resolveToTrait ?: return
+        val traitRef = impl.traitRef ?: return
+        val trait = traitRef.resolveToTrait ?: return
         val traitName = trait.name ?: return
+        if (impl.unsafe != null && trait.unsafe == null) {
+            holder.createErrorAnnotation(traitRef, "Implementing the trait `$traitName` is not unsafe [E0199]")
+        } else if (impl.unsafe == null && trait.unsafe != null) {
+            holder.createErrorAnnotation(traitRef, "The trait `$traitName` requires an `unsafe impl` declaration [E0200]")
+        }
         // Macros can add methods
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -311,6 +311,11 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     private fun checkImpl(holder: AnnotationHolder, impl: RsImplItem) {
         val trait = impl.traitRef?.resolveToTrait ?: return
         val traitName = trait.name ?: return
+        if (impl.unsafe != null && trait.unsafe == null) {
+            holder.createErrorAnnotation(impl.traitRef!!, "Safe traits should not have unsafe implementations [E0199]")
```
alygin: We usually try to stick to the exact (or at least as close as we can) error messages that the compiler emits. In these cases they are "Implementing the trait \`Foo\` is not unsafe [E0199]" and "The trait \`Foo\` requires an \`unsafe impl\` declaration [E0200]". Could you, please, change them accordingly?

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -860,4 +860,24 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase() {
             self::foo()
         }
     """)
+
+    fun `testE0199 Only safe impls for safe traits`() = checkErrors("""
```
alygin: Annotation tests in this file are ordered respectively to their error codes. Would you, please, move them to the right positions? It makes maintenance a bit easier.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -311,6 +311,11 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
+        } else if (impl.unsafe == null && trait.unsafe != null) {
+            holder.createErrorAnnotation(impl.traitRef!!, "Unsafe traits must have unsafe implementations [E0200]")
```
alygin: Let's declare `traitRef = impl.traitRef ?: return` at the very method beginning and get rid of these `!!`s.

farodin91:
@alygin Resolved!

alygin:
Thanks!

bors r+

PR #1307, opened 2017-06-01
(TYPE): join struct lines removing last comma

### LINKED ISSUE #1091: Join lines should remove `,` in struct literal
```
S {
  foo: 1,
}
```

`Ctrl+Shift+J` ->

```
S { foo: 1 }
```

greg-kargin:
Close #1091 

struct-join


src/main/kotlin/org/rust/ide/actions/RsJoinRawLinesHandler.kt
```diff
@@ -19,9 +18,23 @@ class RsJoinRawLinesHandler : JoinRawLinesHandlerDelegate {
...
+
+        val joinStruct = tryJoinStruct(file, start)
+        if (joinStruct != CANNOT_JOIN)
+            return joinStruct
+
         return tryJoinSingleExpressionBlock(file, start)
     }
 
+    fun tryJoinStruct(file: RsFile, start: Int): Int {
+        val body = file.findElementAt(start)!!.parent
```
matklad: Hm, I think we need a somewhat more precise check here. Perhaps using `.parentOfType<RsStructLiteral>` and 
checking that the opening brace and the caret are on the same lines. otherwise this example won't work

```
let _ = S {
   foo: 92 <caret>,
}
```

And that gets me thinking about what actual behavior do we want here? Joining the *whole* structure to the single seems a bit out of scope for joining lines (that means that my original test case is wrong). The better behavior would be to just join two lines, as usual, but to make sure that if `}` is on the same line as the last field, there's no trailing comma. It would be nice to check if it works well with selection as well!

So, I am thinking about the following test cases:

```rust
S { |
    foo: 92,
}

->

S { foo: 92, |
}

->

S { foo: 92 }

---------------------------------

S {
  foo: 92,
  bar: |92,
}

->

S {
  foo: 92,
  bar: 92 }

---------------------------------

S <selection>{
  foo: 92,
  bar: 92,
}</selection>

->

S { foo: 92, bar: 92 }
```

greg-kargin:
join-lines

This commit responds the tests you mentioned, except it takes an additional click when joining lines in the middle of struct:
```
let _ = S {
    foo: 42,
    |bar: 62,
};

- ctrl-shift-j ->

let _ = S {
    foo: 42,
    bar: 62|
};

- ctrl-shift-j ->

let _ = S {
    foo: 42,
    bar: 62 };
```
I'm not sure how to deal with it.

src/test/kotlin/org/rust/ide/actions/RsJoinRawLinesHandlerTest.kt
```diff
@@ -4,12 +4,90 @@ class RsJoinRawLinesHandlerTest : RsJoinLinesHandlerTestBase() {
...
+                bar: /*caret*/42,
+             };
+        }
+    ""","""
+        struct S { foo: i32, bar: i32 }
+        fn main() {
+            let _ = S {
+                foo: 42,
+                bar: 42/*caret*/
+            };
```
matklad: This seems a bit strange: we've removed the comma, but we've failed to actually join the lines...

src/main/kotlin/org/rust/ide/actions/RsJoinRawLinesHandler.kt
```diff
@@ -19,9 +19,36 @@ class RsJoinRawLinesHandler : JoinRawLinesHandlerDelegate {
...
+        while (elem != null && elem.elementType != STRUCT_LITERAL_FIELD) elem = elem.prevSibling
+        return elem
+    }
+
+    fun tryJoinStructLastLine(file: RsFile, start: Int): Int {
+        val struct = file.findElementAt(start)?.parentOfType<RsStructLiteral>() ?: return CANNOT_JOIN
+        val lastField = struct.structLiteralBody.structLiteralFieldList.last() ?: return CANNOT_JOIN
+        val currentField = getCurrentField(file, start) ?: return CANNOT_JOIN
+
+        if (currentField != lastField) return CANNOT_JOIN
```
matklad: Hm, the logic here seems a bit complicated. Perhaps it would be easier to add this to `RsJoinLinesHandler`, which is executed *after* the lines are joined? That way, the condition would be "at the join point, comma is followed by `}`".

matklad:
Looks perfect now, thanks a lot! And sorry for a slow response: I've been terribly busy with other stuff recently :( 

bors r+

greg-kargin:
@matklad thanks! my first merged pr ever:)

PR #1310, opened 2017-06-01
Cargo check annotator

vasily-kirichenko:
Please, help.

I added a check command to cargo module, so I can call it for current project and got errors and warnings in json format like this:

```json
{
  "message": {
    "children": [
      {
        "children": [],
        "code": null,
        "level": "note",
        "message": "#[warn(unused_parens)] on by default",
        "rendered": null,
        "spans": []
      }
    ],
    "code": null,
    "level": "warning",
    "message": "unnecessary parentheses around `if` condition",
    "rendered": null,
    "spans": [
      {
        "byte_end": 73,
        "byte_start": 67,
        "column_end": 14,
        "column_start": 8,
        "expansion": null,
        "file_name": "src\\main.rs",
        "is_primary": true,
        "label": null,
        "line_end": 9,
        "line_start": 9,
        "suggested_replacement": null,
        "text": [
          {
            "highlight_end": 14,
            "highlight_start": 8,
            "text": "    if (true) {"
          }
        ]
      }
    ]
  },
  "package_id": "test1 0.1.0 (path+file:///C:/Users/vaski/IdeaProjects/test1)",
  "reason": "compiler-message",
  "target": {
    "crate_types": [
      "bin"
    ],
    "kind": [
      "bin"
    ],
    "name": "test1",
    "src_path": "C:\\Users\\vaski\\IdeaProjects\\test1\\src\\main.rs"
  }
}
```
I added a (temporary) action "Cargo check", which executes the above function, just to check it works (it does).

I added an annotator that marks everything with "foo" error.

Now I need to wire these together: on any source file change (and Cargo.toml change?) launch the command and push results to the annotator, which creates annotations from given json and notify IDEA about them, then IDEA got the annotation by calling our annotator back. 

Is this how things work in IDEA? An example of an annotator that works on an external data instead of AST would be great.


alygin:
There already [was an attempt](https://github.com/intellij-rust/intellij-rust/pull/1067) to use `cargo check` as an external annotator. You can find some useful information in the discussion in gitter (starting Mar 14, 2017). But the author decided to make it a separate plugin, so it's not a part of IntellJ-Rust. I'm not sure if this plugin is already available though.

vasily-kirichenko:
@alygin thanks, that PR closes all questions. Cool :)

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,170 @@
...
+            parser.parse(it)
+        }.mapNotNull {
+            if (it.isJsonObject)
+                CargoTopMessage.fromJson(it.asJsonObject)
+            else
+                null
+        }
+    }
+
+    override fun getModificationCount() = project?.baseDir?.modificationStamp ?: 0
```
vasily-kirichenko: I think it's a wrong way to determine that anything in the project has been changed. What's the right way?

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,170 @@
...
+}
+
+class RsCargoCheckAnnotator : ExternalAnnotator<CargoCheckAnnotationInfo, CargoCheckAnnotationResult>() {
+
+    private fun getCachedResult(file: PsiFile) =
+        CachedValuesManager.getManager(file.project).createCachedValue {
+            CachedValueProvider.Result.create(
+                checkProject(file),
+                PsiModificationTracker.MODIFICATION_COUNT)
+
```
vasily-kirichenko: 1. Do I understand right that `getManager(file.project)` returns a cache stored in the `Project`, but it does not use it to invalidate itself?

2. `checkProject(file)` returns `CargoCheckAnnotationResult` which implements `ModificationTracker` and it's used to determine if a cache entry should be invalidated?

vasily-kirichenko:
It seems to work OK:

image

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+
+
+        val extraMessage = run {
+            //val code = if (message.code?.code.isNullOrBlank()) "" else "${message.code?.code}"
+            val label =
+                if (span.is_primary && span.label != null)
+                    "${escapeHtml(span.label)}<br/>"
+                else ""
```
alygin: This `if` looks exactly as the previous `when`. Maybe they could be combined?

```
   val (shortMessage, extraMessage) = if (span.is_primary && span.label != null)
            Pair(message.message, "${escapeHtml(span.label)}<br/>")
        else
            Pair(span.label, "")
```

alygin:
We should also skip the "aborting due to previous error" message. It's a meta message that is always present when there're errors in the code. It doesn't actually relate to our project.

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,203 @@
...
+        // The compiler message lines and columns are 1 based while intellij idea are 0 based
+        val textRange =
+            TextRange(toOffset(span.line_start - 1, span.column_start - 1),
+                toOffset(span.line_end - 1, span.column_end - 1))
+
+        val shortMessage =
+            when (span.is_primary) {
+                true -> message.message
+                else -> span.label ?: message.message
+            }
```
alygin: It would be nice to add an error code (if it's available) to this message, as we do in error annotators: "can't find crate for `foo` [E0463]"

src/main/kotlin/org/rust/cargo/toolchain/CargoMessage.kt
```diff
@@ -0,0 +1,77 @@
...
+)
+
+data class Code(
+    val code: String,
+    val explanation: String?)
+{
+    fun formatAsLink() : String {
+        val rustErrorIndexPrefix = "https://doc.rust-lang.org/error-index.html"
+
+        return "<a href=\"$rustErrorIndexPrefix#$code\">$code</a>"
```
alygin: Actually, this `formatAsLink()` is unused and can be removed. But if you decide to add links to error codes (which would be very useful), it'll be nice to move the url or the whole error link format string to `RsConstants` and use it here and in `RsExplainFilter` that also contains this constant.

alygin:
In my opinion, this annotator should be optional, otherwise it'll drain laptop batteries in no time. What do you think?

vasily-kirichenko:
> In my opinion, this annotator should be optional, otherwise it'll drain laptop batteries in no time. What do you think?

I agree. Any points where to add a setting?

alygin:
`RustProjectSettingsPanel` looks suitable for this setting.

I also wander if we should disable our own error annotations and some of inspections when this one is enabled? On the one hand, we don't want to have duplicate messages. On the other hand, `cargo check` doesn't always analyse all the file. We could have three options for the setting: Use cargo check, use our own annotator, use both. But I'm not sure if it's right and what should be the default option. Just thinking out loud.

alygin:
Empty blocks sometimes sneak into tooltips:

screen shot 2017-06-04 at 12 34 09

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,203 @@
...
+            val severity = when (message.level) {
+                "error" -> HighlightSeverity.ERROR
+                "warning" -> HighlightSeverity.WEAK_WARNING
+                else -> throw AssertionError()
+            }
+
+            val codeStr = message.code?.code
+
+            // If spans are empty we add a "global" error
+            if (message.spans.isEmpty()) {
```
alygin: Though I see that the current implementation works fine with all the corner cases that I was able to remember. It seems that the compiler output is more friendly now, so I think this additional `span.extension` walking can be skipped, at least at this moment.

vasily-kirichenko:
I have a problem: if there is a build error I cannot find what the error is, what I can see is only this:

image

vasily-kirichenko:
All done.

vasily-kirichenko:
BTW, the way settings are handled is very fragile. I had to add same field in several places, copy values from one structure to another in several places, add it to the condition that checks if settings's been modified...

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,179 @@
...
+        private val messageRegex = """\s*\{\s*"message".*""".toRegex()
+    }
+
+    val messages =
+        commandOutput
+            .filter { messageRegex.matches(it) }
+            .map { parser.parse(it) }
+            .filter { it.isJsonObject }
+            .mapNotNull { CargoTopMessage.fromJson(it.asJsonObject) }
+            .filter { it.message.message != "aborting due to previous error" }
```
alygin: Here should be either a regexp or `.startsWith("aborting due to ")`, because the message can also be "aborting due to N previous errors".

vasily-kirichenko: Done.

alygin:
Cool!

I think we now need to bring some order to the displayed messages. There're three problems I see now:

1. Too many redundant message blocks. In IDEA, a horizontal line (`<hr>`) in tooltips is usually used to separate _different_ problems that relate to the same span or to the same line. We now show several blocks that are actually describe a single error, they just duplicate each other:

screen shot 2017-06-04 at 18 10 56

I think it should look like this:

screen shot 2017-06-04 at 18 10 56 - 2

I. e. we need to a) skip span labels if the main error message starts with it (there're [more deduplication rules](https://github.com/AtomBuild/atom-build-cargo/blob/736f12bb2fb0829fa6ab51f4651404fa20fb920e/lib/errors.js#L17), but I think this simple one is enough at the time), and b) add the error code to the main message, not as a separate message.

2. Don't add non-primary spans. Every non-primary span only has value in the context of a particular error message. We don't connect secondary spans to their messages (there're no tools for that), so I think we just need to skip them and not create confusing annotations like this that pop out from nowhere:

screen shot 2017-06-04 at 18 24 45

3. Messages and labels are sometimes connected wrongly. In this example, the span label is attached to the help message:

screen shot 2017-06-04 at 18 28 25

It would be nice to have it like this:

screen shot 2017-06-04 at 18 28 25 - 2

If we sum it up, the rules will be:
- Add a single annotation per error message with the primary span.
- Add error code (if available) to the main message text.
- Add a primary span label in a separate line if the main message doesn't start with it.
- Add note- and help-messages in a separate lines, replacing line breaks with `<br>`s.
- When creating a new annotation, the main message should go to the `message` parameter, while main message with the code, the primary span label and additional help messages should go to `htmlTooltip`.
- Ignore secondary spans because they have no connection with the main message and might be confusing.

In the future, we can try to bring secondary span labels back and provide them with the links to source code (like in `RegexpFileLinkFilter`, if IDEA allows that). So, they will be connected to their master error messages and don't confuse the user.

How do you think, do these rules make sense?

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
alygin: Oh, and [here's one more meta error](https://github.com/AtomBuild/atom-build-cargo/blob/736f12bb2fb0829fa6ab51f4651404fa20fb920e/lib/errors.js#L12) that should be handled the same way.

vasily-kirichenko:
image

image

src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
```diff
@@ -98,6 +104,7 @@ class RustProjectSettingsPanel(private val cargoProjectDir: String = ".") : Disp
         if (PlatformUtils.isIntelliJ()) {
             row("Use cargo check:") { useCargoCheckForBuildCheckbox() }
         }
+        row(label = Label("Use cargo check to analyze code:")) { useCargoCheckAnnotatorCheckbox() }
```
alygin: Having two settings with nearly identical names is a bit confusing. Let's rename "Use cargo check:" to "Use cargo check when build project:"
@matklad?

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,179 @@
...
+                toOffset(span.line_start - 1, span.column_start - 1),
+                toOffset(span.line_end - 1, span.column_end - 1))
+
+        val tooltip = run {
+            val lines = ArrayList<String>()
+            val code = message.code.formatAsLink()
+            lines.add(escapeHtml(message.message) + if (code == null) "" else " $code")
+
+            if (span.label != null && !message.message.startsWith(span.label))
+                lines.add(escapeHtml(span.label))
```
alygin: We're trying to format such `if`s either as one-liners, or add braces around the body. It helps to reduce the possibility of accidental bugs when adding new code.

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,179 @@
...
+
+            message.children
+                .filter { !it.message.isBlank() }
+                .forEach {
+                    val prefix = when (it.level) {
+                        "note" -> "Note: "
+                        "help" -> "Help: "
+                        else -> ""
+                    }
```
alygin: Could be shorter yet more general: `val prefix = "${it.level.capitalize()}: "`

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,179 @@
...
+                        .filter { it.is_primary }
+                        .filter { isValidSpan(it) }
+                        .filter { filePath.endsWith(PathUtil.getCanonicalPath(it.file_name)) }
+                        .firstOrNull()
+
+                if (primarySpan != null) {
+                    val annotation = createAnnotation(primarySpan, message, severity, doc, holder)
+                    annotation.problemGroup = problemGroup
+                    annotation.setNeedsUpdateOnTyping(true)
+                }
```
alygin: After the refactoring, maybe `primarySpan == null` is the right condition for global annotations, not `message.spans.isEmpty()`? Otherwise we might lose some messages and don't notice it.

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,179 @@
...
+                        "help" -> "Help: "
+                        else -> ""
+                    }
+                    lines.add(prefix + escapeHtml(it.message))
+                }
+
+            lines.joinToString("<br>").replace("\n", "<br>")
+        }
+
+        return holder.createAnnotation(severity, textRange, message.message, "<html>$tooltip</html>")
```
alygin: Html escaping is broken somewhere:

screen shot 2017-06-05 at 01 32 06

IDEA expects the tooltip to be an html-string, there's no need to put it in `<html>...</html>`. But I don't think this causes the problem.

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private val ERROR_INDEX_URL = "https://doc.rust-lang.org/error-index.html"
```
alygin: Let's move this constant to `RsConstants` and reuse it here and in `RsExplainFilter`.

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
vasily-kirichenko: I cannot reproduce this tooltip:

image

Could you provide full repro snippet?

alygin: It looks like we use different compiler versions, but actually, you've reproduced the problem. Have a closer look at this `std::ops::Add&str>`, the opening `<` is missing, it should be `std::ops::Add<&str>`.

vasily-kirichenko:
All done.

image

matklad:
Wow, @vasily-kirichenko , this looks super awesome!

Would you like to add a test here as well? Tests that interract with Cargo can inherit `RustWithToolchainTestBase`.

vasily-kirichenko:
@matklad sure, will add tests tonight or tomorrow.

vasily-kirichenko:
I added a couple of simple tests that call cargo check and verify that it returns zero and non zero code in case of success and error. 

However, I've no idea how to check the annotator because it requires to inherit `RsAnnotatorTestBase`, which cannot run external annotators AFAIK.

vasily-kirichenko:
The test fails with this:

```
org.rust.cargo.CargoCheckTest > testZeroErrorCodeIfProjectHasNoErrors FAILED
    junit.framework.AssertionFailedError: Expected zero error code, but got 101. cmd = /home/travis/.cargo/bin/cargo check --color=always --message-format=json, stdout = error: no such subcommand: `check`
    	Did you mean `clean`?
```

It passes on my machine on windows though. Any ideas?

matklad:
We use Rust 1.13 on the CI, and 1.13 does not have `cargo check`. Should be OK to just bump this version to the first Rust which has `check` here: https://github.com/intellij-rust/intellij-rust/blob/master/.travis.yml#L12

vasily-kirichenko:
OMG the tests passed :)

matklad:
bors r+

Thank you very much, @vasily-kirichenko!

alygin:
The problem with html-formatting is still here. It looks like the platform bug to me, because the html looks correct. @matklad, maybe you know something about it?

PR #1314, opened 2017-06-02
RES: Move macro resolve to get completion

farodin91:
ToDo
- [x] resolve in large libs like std
- [x] support special macros?
- [x] support reexport
- [x] completion legacy provider never call RsMacroItemRef


farodin91:
@matklad I get resolve for local macro_rules done but not for imports

farodin91:
@matklad Could you review the current state?
ToDo
- resolve in large libs like std
- support special macros?
- completion legacy provider never call RsMacroItemRef
- support reexport

Some tips?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -293,6 +294,69 @@ fun processMacroSimpleResolveVariants(element: RsMacroBodySimpleMatching, proces
...
+    }
+    if (add_std_crate) {
+        val stdCrate = scope.containingCargoPackage?.findCrateByName("std")?.crateRoot
+        val prelude = scope.project.getPsiFor(stdCrate)?.rustMod
+        if (prelude is RsMod && processItemMacroDeclarations(prelude, processor, true)) return true
+    }
+
+    for (modDecl in scope.modDeclItemList.filter { it.hasMacroUse }) {
+        val mod = modDecl.reference.resolve() ?: continue
+        if (mod is RsMod && processItemMacroDeclarations(mod, processor, true)) return true
```
matklad: So in the worst case, we'll walk the whole crate here, but we hope only some modules have `macro_use`?

I think it is very sensible approach, though I think that perhaps we need to cache this information? Something like 

```
fun exportedMacros(crateRoot: RsFille): Collection<RsMacroDefinition> {
     CachedValuesManager.getCachedValue ... 
}
```

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -293,6 +294,69 @@ fun processMacroSimpleResolveVariants(element: RsMacroBodySimpleMatching, proces
     return simple.any { processor(it) }
 }
 
+fun processMacroReferenceVariants(referenceElement: RsReferenceElement, isCompletion: Boolean, processor: RsResolveProcessor): Boolean {
+    val prevScope = mutableSetOf<String>()
+    walkUp(referenceElement, { false }) { cameFrom, scope ->
+        val currScope = mutableListOf<String>()
+        val shadowingProcessor = { e: ScopeEntry ->
```
matklad: Hm, do macros in Rust support real scope based shadowing?

farodin91: Rust don't allow shadowing

farodin91: https://doc.rust-lang.org/book/first-edition/macros.html

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
farodin91: Do you like my implementation of Caching

farodin91:
@matklad How to fix completion using the legacy provider?

farodin91:
@matklad Now this support the same as the old marco resolve and added stubbed attr?
But i'm not shure how to add completion easly?

farodin91:
@matklad Tried to implement a PsiReferenceContributor, but it didn't get called.
kwww.jetbrains.org/intellij/sdk/docs/tutorials/custom_language_support/reference_contributor.html

farodin91:
@matklad Could you review?

matklad:
@farodin91 hopefully I'll get to this this weekend!

src/main/kotlin/org/rust/lang/core/resolve/ref/RsMarcoItemReferenceContributor.kt
```diff
@@ -0,0 +1,24 @@
...
+package org.rust.lang.core.resolve.ref
+
+import com.intellij.patterns.PlatformPatterns
+import com.intellij.psi.*
+import com.intellij.util.ProcessingContext
+import org.rust.lang.core.psi.RsElementTypes.IDENTIFIER
+
+class RsMarcoItemReferenceContributor : PsiReferenceContributor() {
+    override fun registerReferenceProviders(registrar: PsiReferenceRegistrar) {
+        val pattern = PlatformPatterns.psiElement(IDENTIFIER)//never called
```
matklad: Reference providers are only called for elements implementing `ContributedReferenceHost` and I don't think we need to use it here. 

Contributed references are useful if you want to extend go to definition for some language *from outside*. For example, if you are creating support for a particular framework, which renders template like `ctx.render("template_file_name.ctx")` , you can *inject* reference to a file into the string literal, provided that string literal in the language is a `ContributedReferenceHost`. 

Here, we have control over PSI, so we can return the reference from `getReference` directly.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
matklad: Yep, it looks all right!

I am a bit worried though that there's a bit of duplication between this function and `exportedCrateMacros`.  The loop
`for (mod in crateRoot.modItemList.filter { it.hasMacroUse })` is essentially repeated in both functions.

src/main/kotlin/org/rust/lang/core/resolve/ref/RsMarcoItemReferenceContributor.kt
farodin91: I tried this to get completion working for marcos

matklad: I think completion should be handled by `CompletionProvider`. Doing completion via the usual reference machinery won't work, because we don't know if we are typing a macro until we see `!`.

farodin91: It is possible to create a reference to use variants.

matklad: I don't think that creating a reference will actually work here, even if we implement `ContributedReferenceHost`. Consider this (incomplete) code:

```
macro_rules foobar! { ... }
fn fooquux() { ... }

fn main() {
    let _ = foo
}
```

Here, we can't really say if `foo` is a macro refernce, or a "usual" function reference. However, completion contributor should still be able to provide completions.

farodin91: I want to use my current process alg. To write an completion contributor. Else I need to write some parts more than once.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
farodin91: I see

farodin91:
@matklad All issues are resolved.

matklad:
It seems to work now! Looks like I wouldn't have to type `unimplemented!()` fully again, once this is merge. 

However, I get a `ClassCastException` in breadcrumbs: looks like `MaroDefinitions` stopped to be a `NamedElement` in this PR. I'll push a test for this soon!

matklad:
Here's the test: https://github.com/intellij-rust/intellij-rust/pull/1461

matklad:
@farodin91 I've looked through this carefully, and I think I'll push this over the finish line myself: our grammar and PSI structure for macros is quite incomprehencible at the moment, so I hope to use this PR as an excuse to try to clean it up!

matklad:
Ok, this lands as part of https://github.com/intellij-rust/intellij-rust/pull/1467, thanks a lot, @farodin91 !

PR #1316, opened 2017-06-03
TY: improve type inference of method call expr

Undin:
Support type inference of generic method call exprs.


matklad:
Nice!

bors r+

PR #1319, opened 2017-06-07
T: vec slice resolve test

Undin:
Add integration test to check interaction between range expression and index expression type inference


matklad:
bors r+

Test pull requests are the best, thanks a lot!

PR #1326, opened 2017-06-09
TYPE: Add inlay parameter hints

### LINKED ISSUE #1141: Show function parameters automatically
It would be really helpful to see that you have to pass `(foo: Foo, bar: Bar)` while calling a function.

farodin91:
Fixes #1141

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
@@ -1,0 +1,101 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.hints
+
+import com.intellij.codeInsight.hints.HintInfo
+import com.intellij.codeInsight.hints.InlayInfo
+import com.intellij.codeInsight.hints.InlayParameterHintsProvider
+import com.intellij.codeInsight.hints.Option
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.selfParameter
+import org.rust.lang.core.psi.ext.valueParameters
+import org.rust.lang.core.types.infer.inferDeclarationType
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.type
+
+private val RsCallExpr.declaration: RsFunction?
+    get() = (expr as? RsPathExpr)?.path?.reference?.resolve() as? RsFunction
+
+private val RsMethodCallExpr.declaration: RsFunction?
+    get() = reference.resolve() as? RsFunction
+
+enum class HintType(desc: String, enabled: Boolean) {
+    LET_BINDING_HINT("Show local variable type hints", true) {
+        override fun provideHints(elem: PsiElement): List<InlayInfo> {
+            val element = elem as? RsLetDecl ?: return emptyList()
+            if (element.typeReference != null) return emptyList()
+            val ident = element.pat as? RsPatIdent ?: return emptyList()
+            val patBinding = ident.patBinding
+            val type = patBinding.type
+            if (type is TyUnknown) return emptyList()
+            return listOf(InlayInfo(": " + type.toString(), patBinding.textRange.endOffset))
+        }
+
+        override fun isApplicable(elem: PsiElement): Boolean
+            = elem is RsLetDecl
+    },
+    PARAMETER_HINT("Show argument name hints", true) {
+        override fun provideHints(elem: PsiElement): List<InlayInfo> {
+            val (declaration, valueArgumentList) = when (elem) {
+                is RsCallExpr -> (elem.declaration to elem.valueArgumentList)
+                is RsMethodCallExpr -> (elem.declaration to elem.valueArgumentList)
+                else -> null
+            } ?: return emptyList()
+            val selfParameter = if (declaration?.selfParameter != null && elem is RsCallExpr) {
+                1
+            } else {
+                0
+            }
+            return declaration?.valueParameters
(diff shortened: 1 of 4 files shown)


farodin91:
cc @vasily-kirichenko

vasily-kirichenko:
wow. let me try!

vasily-kirichenko:
Pure beauty :)

vasily-kirichenko:
@farodin91 I found quite a lot of bugs :)

image

farodin91:
I don't show type if i can't infer a type != TyUnknown.

farodin91:
Adding tests and Options

vasily-kirichenko:
But isn't `1 + 1` has `i32` type?..

Undin:
Not now:)
At this moment plugin can't infer result of arithmetic operations for primitive types

vasily-kirichenko:
image

Cool :) But generic parameters are not inferred.

farodin91:
If you set the parameter it should infer parameters. For example `let x:HashMap<String, String>` after this is should infer.

farodin91:
@alygin Could you review?

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -0,0 +1,89 @@
...
+                ?: emptyList()
+        }
+
+        override fun isApplicable(elem: PsiElement): Boolean
+            = elem is RsCallExpr || elem is RsMethodCallExpr
+
+        fun transform(index: Int, valueParameter: RsValueParameter, valueArgumentList: RsValueArgumentList): InlayInfo? {
+            if (index >= valueArgumentList.exprList.size)
+                return null
+            return InlayInfo(valueParameter.pat?.text + ":", valueArgumentList.exprList[index].textRange.startOffset)
```
alygin: You also need to return here if `valueParameter.pat?` is `null` here to handle anonymous arguments properly. Here's a snippet to test this case:

```rust
trait Foo {
    fn foo(&self, u32);
    fn bar(&self) {
        self.foo(10);    // <-- shouldn't add `null: `
    }
}
```

kumbayo:
When calling a method using UFCS syntax, the labels are shifted by 1 for the extra self argument.
```rust
struct S;

impl S {
    fn foo_self(self, param1: u32) {}
    fn foo_self_ref(&self, param1: u32) {}
    fn foo_self_mut_ref(&mut self, param1: u32) {}
}

fn test() {
    let mut b = S;

    b.foo_self_ref(1);
    b.foo_self_mut_ref(1);
    b.foo_self(1);

    S::foo_self_ref(&b, 1); //bad
    S::foo_self_mut_ref(&mut b, 1); //bad
    S::foo_self(b, 1); //bad
}
```

inlay_parameter_hints_ufcs

farodin91:
@alygin For IDEA you can de/select a every single feature. Any idea to integrate it into rust settings panel and in the configure area? At the same time.

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -0,0 +1,100 @@
...
+
+enum class HintType(desc: String, enabled: Boolean) {
+    LET_BINDING_HINT("Show local variable type hints", true) {
+        override fun provideHints(elem: PsiElement): List<InlayInfo> {
+            val element = elem as? RsLetDecl ?: return emptyList()
+            if (element.typeReference != null) return emptyList()
+            val ident = element.pat as? RsPatIdent ?: return emptyList()
+            val patBinding = ident.patBinding
+            val type = inferDeclarationType(patBinding)
+            if (type is TyUnknown || type is TyUnit) return emptyList()
```
kumbayo: Is there a reason / use case why you do not want to show the unit type on local variables?
Noise during typing new code?

I think it might be useful to show it for situations like.
```rust
let z = accidentally_returns_nothing();

fn accidentally_returns_nothing() -> () {
    ;
}
```

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -0,0 +1,100 @@
+package org.rust.ide.hints
+
+import com.intellij.codeInsight.hints.HintInfo
+import com.intellij.codeInsight.hints.InlayInfo
+import com.intellij.codeInsight.hints.InlayParameterHintsProvider
+import com.intellij.codeInsight.hints.Option
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsCompositeElement
+import org.rust.lang.core.psi.ext.valueParameters
```
kumbayo: IntelliJ claims this import is unused.

matklad:
Hm, compilation fails on 2016.3... I need to check if we can make extensions in `plugin.xml` dependent on the version of IDE...

farodin91:
But my implementation of settings shouldn't work in any case for 2016.3.

kumbayo:
It seems the `S::foo_self_ref(&b, 1);` case has regressed.
In an intermediate version it already worked.
But now it is not working anymore and the test for it is gone again.

farodin91:
@kumbayo Thank i pushed something wrong.

farodin91:
@matklad Any progress to fix compilation fails on 2016.3?

matklad:
@farodin91 looks like the easiest solution is just to wait while we drop support for 2016.3. This'll happen once 2017.2 is out, which is pretty soon.

vlad20012:
2017.2 just released

farodin91:
I will rebase after #1487 is merged.

farodin91:
Resolved

farodin91:
@matklad Do you like to merge it?

matklad:
@farodin91 sure!

bors r+

PR #1328, opened 2017-06-10
ANNOT: Handle wrong number of arguments passed to variadic functions

kumbayo:
External C functions are allowed to be variadic. Variadic functions
take a minimum amount of arguments and passing more arguments is valid/expected.

Our current implementation always expects the exact number of arguments
to be passed and displays an error otherwise.

Fix this behavior and provide an show error E0060
when not passing enough arguments to a variadic function.

RsMethodCallExpr.expectedParamsCount() was only adjusted for consistency
as variadic calls are only allowed for extern C functions.


matklad:
bors r+

Thanks!

PR #1329, opened 2017-06-10
Check test

matklad:
Make it easier to test `CargoCheckAnnotator`. 

cc @vasily-kirichenko

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
@@ -16,6 +16,7 @@ import com.intellij.psi.PsiManager
 import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
 import com.intellij.psi.util.PsiModificationTracker
+import com.intellij.util.PathUtil
 import org.apache.commons.lang.StringEscapeUtils.escapeHtml
 import org.rust.cargo.project.settings.rustSettings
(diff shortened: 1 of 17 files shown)


src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -82,9 +82,12 @@ class RsCargoCheckAnnotator : ExternalAnnotator<CargoCheckAnnotationInfo, CargoC
             val primarySpan = message.spans
                 .filter { it.is_primary }
                 .filter { it.isValid() }
-                .filter { filePath.endsWith(PathUtil.getCanonicalPath(it.file_name)) }
```
matklad: @vasily-kirichenko is there any particular reason why we need `getCanonicalPath` here? It is a somewhat expensive operation, because it resolves symlinks, and thus required a syscall per span.

vasily-kirichenko: on windows filePath and it.file_name have different kind of slashes, that's why.

matklad: Thanks! Let's check that appveyor fails then, to make sure this is tested :) !

matklad: Perfect timing: https://ci.appveyor.com/project/matklad/intellij-rust/build/1.0.2172 :P

PR #1332, opened 2017-06-11
INT: add specify type explicitly intention

sirgl:
peek 2017-06-11 18-31

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/SpecifyTypeExplicitlyIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/SpecifyTypeExplicitlyIntention.kt
@@ -23,7 +23,7 @@ class SpecifyTypeExplicitlyIntention : RsElementBaseIntentionAction<SpecifyTypeE
             return null
         }
         val ident = letDecl.pat as? RsPatIdent ?: return null
-        val type: Ty = inferDeclarationType(ident.patBinding)
+        val type = inferDeclarationType(ident.patBinding)
         if (type is TyUnknown) {
             return null
         }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/intentions/SpecifyTypeExplicitlyIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    override fun getText() = "Specify type explicitly"
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val letDecl = element.parentOfType<RsLetDecl>() ?: return null
+        if(letDecl.typeReference != null) {
+            return null
+        }
+        val ident = letDecl.pat as? RsPatIdent ?: return null
+        val type: Ty = inferDeclarationType(ident.patBinding)
```
mkaput: heh, the explicit type declaration here is unnecessary 😉

src/main/kotlin/org/rust/ide/intentions/SpecifyTypeExplicitlyIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return Context(type, letDecl)
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        val factory = RsPsiFactory(project)
+        val createdType = factory.createType(ctx.type.toString())
+        val letDecl = ctx.letDecl
+        val colon = letDecl.addAfter(factory.createColon(), letDecl.pat)
+        colon.parent.addAfter(createdType, colon)
```
mkaput: why not just `letDecl` instead of `colon.parent`?

sirgl:
I think shorter means better :)

mkaput:
I think this is small enough that I can shamelessly type it for the first time :D

bors r+

Thanks!

PR #1334, opened 2017-06-12
TYPE: Join on trailing commas

### LINKED ISSUE #1091: Join lines should remove `,` in struct literal

d9n:
Any join on a line that ends in a comma where the next line begins in an
rbrace will result in the comma being removed, e.g.

```rust
...
    val: i32,
}
```

joins into

```rust
    val: i32 }
```

This allows the user to easily compress a multi-line struct into a
single-line one, e.g.

```rust
struct s {
    val1: i32,
    val2: i32,
    val3: i32,
}
```

into

```rust
struct s { val1: i32, val2: i32, val3: i32 }
```

Fixes #1091

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RsJoinRawLinesHandler.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsJoinRawLinesHandler.kt
@@ -4,10 +4,10 @@ import com.intellij.codeInsight.editorActions.JoinLinesHandlerDelegate.CANNOT_JO
 import com.intellij.codeInsight.editorActions.JoinRawLinesHandlerDelegate
 import com.intellij.openapi.editor.Document
 import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.util.PsiTreeUtil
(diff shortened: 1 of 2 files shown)


Undin:
Duplicates #1307

matklad:
@d9n sorry for missing you PR at first! (I am very busy with other stuff at the moment unfortunately)

Actually, I think that your approach of not restricting this to struct literals is superior to what was originally suggested! That is, I think that joining lines in struct definitions:

```
struct S {
  f: i32,
  x: String,/*caret*/
}
```

should probably remove trailing `,` as well.

Would you like to open a PR which remove the restriction and adds more tests (struct definitions, patterns, maybe something else?)

As for the question about the process, I probably should have noticed this beforehand, because I new that #1307 was in flight :( Also there's a link to #1307 from the issue page! I am not super in favor of requiring to claiming an issue via comment, because it places a bit more burden on contributor (if one claims an issue, one may feel that it is necessary to actually fix it, which may cause some stress if there are unforeseen difficulties).

PR #1338, opened 2017-06-13
Debug mysterious bug, probably in Cargo

matklad:


src/main/kotlin/org/rust/cargo/toolchain/CargoMessage.kt
```diff
@@ -12,6 +12,9 @@ data class CargoTopMessage(
 ) {
     companion object {
         fun fromJson(json: JsonObject): CargoTopMessage? {
+            if (json.getAsJsonPrimitive("reason") == null) {
+                error("Maformed JSON from Cargo: $json")
```
mkaput: Malformed

matklad: Yeah, I don't expect this to land, I just want to debug what's goining on, because the failure fails to reproduce locally :( 

Hope the test will fail :)

notriddle:
BTW, @matklad, for this kind of test you probably want to use "bors try". That way you don't accidentally merge if the tests "pass."

PR #1351, opened 2017-06-16
RES: resolve traits in derive attribute

Undin:


src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
```diff
@@ -351,4 +351,27 @@ class RsStdlibResolveTest : RsResolveTestBase() {
...
+            "Hash" to "hash/mod.rs",
+            "Ord" to "cmp.rs",
+            "PartialEq" to "cmp.rs",
+            "PartialOrd" to "cmp.rs"
+        )
+        for ((trait, path) in traitToPath) {
+            // use different file names to avoid exception because of
+            // creation of already existing file
+            stubOnlyResolve("""
+            //- ${trait.toLowerCase()}.rs
```
Undin: Is there a good way to remove the temporary file after `stubOnlyResolve` method call?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -293,6 +296,26 @@ fun processMacroSimpleResolveVariants(element: RsMacroBodySimpleMatching, proces
...
+    val traitName = element.identifier.text
+    val stdTrait = STD_DERIVE_TRAITS[traitName]
+    val project = element.project
+    val trait = StubIndex.getElements(
+        RsNamedElementIndex.KEY,
+        traitName,
+        project,
+        GlobalSearchScope.allScope(project),
+        RsNamedElement::class.java
+    ).find { e ->
```
matklad: This is going to hit an index on every resolve, but I think we can easily add some caching here. What do you thing about adding something like `fun derivableTraits(module: Module): Collection<TraitItem>`, which would cache the results based on PsiModifictionCount?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -293,6 +296,26 @@ fun processMacroSimpleResolveVariants(element: RsMacroBodySimpleMatching, proces
     return simple.any { processor(it) }
 }
 
+fun processMetaItemResolveVariants(element: RsMetaItem, processor: RsResolveProcessor): Boolean {
+    if (element.parentOfType<RsMetaItem>()?.identifier?.text != "derive") return false
+    val traitName = element.identifier.text
+    val stdTrait = STD_DERIVE_TRAITS[traitName]
```
matklad: Looks like it can through an exception if `traitName` is not in `STD_DERIVE_TRAITS`.

src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
matklad: I don't think so, `stubOnlyResolve` is intended to be used as a single expression inside the test method body. I think the way to go is to deconstruct `stubOnlyResolve` into helper functions and use them here.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
Undin: No, if `traitName` is not in `STD_DERIVE_TRAITS` it means that it's some other name, for example, `Serialize` from some library. In this case, I consider that it's trait name and I try to find it but if there are at least two different traits with the same name I can find the wrong declaration. If you consider that it isn't ok, I'll resolve only derivable traits from stdlib.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
Undin: Ok, I think it's a good idea.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
Undin: Or I can pass all not std traits with `traitName` as resolving result and if there are more than one variant user will be able to choose appropriate declaration and in future, we'll probably implement more precise resolving. I think it's better than without resolving at all.

Undin:
I've fixed all comments.
@matklad can you check it again?

matklad:
Looks great now, and thanks for the poke!

bors r+

PR #1353, opened 2017-06-17
ANNOT: Do not require/allow unsafe on negative impls of traits [E0198]

kumbayo:
Negative impls can sometimes be necessary when you want to remove certain
automatically derived marker traits like Send and Sync.

This negative impls does not require (or allow) the unsafe keyword even
when the trait itself is marked as unsafe.

This fixes a wrong error being displayed in libcore/marker.rs for
impl<T: ?Sized> !Send for *const T { }

CC #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -312,9 +312,11 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         val traitRef = impl.traitRef ?: return
         val trait = traitRef.resolveToTrait ?: return
         val traitName = trait.name ?: return
-        if (impl.unsafe != null && trait.unsafe == null) {
+        if (impl.unsafe != null && impl.excl != null) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -312,9 +312,11 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         val traitRef = impl.traitRef ?: return
         val trait = traitRef.resolveToTrait ?: return
         val traitName = trait.name ?: return
-        if (impl.unsafe != null && trait.unsafe == null) {
+        if (impl.unsafe != null && impl.excl != null) {
+            holder.createErrorAnnotation(traitRef, "Negative implementations are not unsafe [E0198]")
```
kumbayo: The already existing errors are displayed on the trait name. So i mimicked that behavior.
rustc seems to display the error on the whole line instead for E0198, E0199 and E0200.

But i guess thats not a huge deal.

matklad:
bors r+

Thank you!

PR #1356, opened 2017-06-18
RES: Handle pat binding names overlapping with module names

kumbayo:
This allows to properly resolve modules when a pat binding with the same name exists.
It covers the following situations creating pat bindings:
- if let
- while let
- for
- match arms
- lambdas


matklad:
bors r+

Thanks!

PR #1367, opened 2017-06-22
CARGO: RUN: added 'Cargo' submenu to project view context menu

jchlapinski:
Hi, I often find myself wanting to execute some auxiliary cargo command and I find it slightly inconvenient to be mixed with my regular run/test configurations.

My proposition is to add a sub menu to project context menu to execute certain cargo commands (namely `cargo check | update | clean` and possibly others, which are not strictly related to run/test/bench the code.

Perhaps it would be better to even let user add his/her own commands into the menu as well.

Regards,
Jakub

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RsCargoAction.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsCargoAction.kt
@@ -1,0 +1,67 @@
+package org.rust.ide.actions
+
+import com.intellij.execution.ExecutorRegistry
+import com.intellij.execution.ProgramRunnerUtil
+import com.intellij.execution.RunManagerEx
+import com.intellij.execution.RunnerAndConfigurationSettings
+import com.intellij.execution.executors.DefaultRunExecutor
+import com.intellij.notification.NotificationType
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.LangDataKeys
+import com.intellij.openapi.module.Module
+import com.intellij.project.isEqualToProjectFileStorePath
+import org.rust.cargo.icons.CargoIcons
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.runconfig.createCargoCommandRunConfiguration
+import org.rust.cargo.toolchain.CargoCommandLine
+import org.rust.cargo.util.modulesWithCargoProject
+import org.rust.ide.icons.RsIcons
+import org.rust.ide.notifications.showBalloon
+import org.rust.ide.utils.isNullOrEmpty
+
+class RsCargoAction(command: String) : AnAction() {
+    val command: String
+
+    init {
+        templatePresentation.text = "Run 'cargo $command'"
+        this.command = command
+    }
+
+    override fun update(e: AnActionEvent) {
+        if (e.project?.toolchain == null || e.project?.modulesWithCargoProject.isNullOrEmpty()) {
+            e.presentation.isEnabled = false
+        }
+    }
+
+    override fun actionPerformed(e: AnActionEvent) {
+        val module = getModule(e) ?: return
+        val commandLine = CargoCommandLine(command)
+
+        runCommand(module, commandLine)
+    }
+
+    private fun getModule(e: AnActionEvent): Module? {
+        val cargoModules = e.project?.modulesWithCargoProject.orEmpty()
+        val current = e.getData(LangDataKeys.MODULE)
+
+        return if (current in cargoModules)
+            current
+        else
+            cargoModules.firstOrNull()
+    }
+
+    private fun runCommand(module: Module, cargoCommandLine: CargoCommandLine) {
+        val runConfiguration = createRunConfiguration(module, cargoCommandLine)
+        val executor = ExecutorRegistry.getInstance().getExecutorById(DefaultRunExecutor.EXECUTOR_ID)
+        ProgramRunnerUtil.executeConfiguration(module.project, runConfiguration, executor)
+    }
+
+    private fun createRunConfiguration(module: Module, cargoCommandLine: CargoCommandLine): RunnerAndConfigurationSettings {
+        val runManager = RunManagerEx.getInstanceEx(module.project)
+
+        val configuration = runManager.createCargoCommandRunConfiguration(cargoCommandLine)
+        configuration.isTemporary = true
+        return configuration
+    }
+}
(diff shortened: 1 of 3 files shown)


matklad:
@jchlapinski are you aware of **Run Cargo Command** action? If you are, why is it inconvenient for your use case?

matklad:
I am somewhat hesitant to add actions directly to the project context menu: not every project uses Rust, and the menu is already very crowded. I personally am happy with **Run Cargo Command** approach (though it could use some command completion!), but the discoverability of this action is very bad. And it certainly would be nice to have some buttons for most common actions preconfigured. I wonder if we should create a Cargo Toolbar? http://www.jetbrains.org/intellij/sdk/docs/user_interface_components/tool_windows.html

jchlapinski:
@matklad 1) it is not accessible easily from anywhere (?) aside from "Help > Find Action..." or adding it to run configurations. 2) When used, it automatically registers itself in run configurations. Somehow I feel that `cargo update` is not something I want to have in run configurations, but that is maybe just my opinion.

matklad:
Yeah, both 1) and 2) are very valid points! For one, I use `ctrl+shif+a` shortcut, but it is definitely not the most convenient thing. 

Would you like to try to implement a Cargo toolbar, somewhat akin the the Gradle one:
gradle-toolbar

jchlapinski:
I was rather going for something similar to this maven submenu
maven-submenu

Of course this submenu is only visible if the current module uses maven toolchain

I do not think full-blown toolbar for cargo is all that useful because: 1) cargo does not really have any custom (project dependent) tasks like those defined in gradle, maven, gulp or other task runners. 2) Having a few common commands at hand is all I really see useful personally.

jchlapinski:
However if there will be good idea what such toolbar would do, I am happy to help implementing it, as much as I can.

matklad:
>However if there will be good idea what such toolbar would do

I think that the main benefit of the toolbar is that it has an always visible icon, which is easier to discover than an entry in the context menu. We can also put some auxilary information there, like the location of `Cargo.toml`, version of toolchain and an error message if current project failed to update successfully. 

That said, toolbar will need a much larger amount of work, so perhaps context menu is a better start? @mkaput, @alygin  what do you think?

jchlapinski:
@matklad well, actually maven has both the context menu and the toolbar. Perhaps this is the way to go for cargo as well. Now that I think of it, there would be nice to show dependencies from Cargo.toml and Cargo.lock in form of a tree in such toolbar... I often find myself seeing the same dependency but with different versions compiled multiple times in my rust projects. It would be nice to see where those dependencies come from. I think I will give it a try this weekend.

mkaput:
Well, I never knew about the context menu for Maven 😄 

Personally I would expect toolbar existing, just because, well, I'm used to Gradle one as this is pretty standard IntelliJ UX. Because of workspaces, there is sense to present list of Cargo tasks in tree form.

matklad:
Sounds great @jchlapinski! 

So I propose that we fix a couple of nits in this PR, merge it just to have something working, and then replace it with a toolbar once it's ready! I am not a big fan of having both the toolbar and the context menu, because it's better not to create duplication in the already quite dense UI.

src/main/kotlin/org/rust/ide/actions/RsCargoAction.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.rust.cargo.icons.CargoIcons
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.runconfig.createCargoCommandRunConfiguration
+import org.rust.cargo.toolchain.CargoCommandLine
+import org.rust.cargo.util.modulesWithCargoProject
+import org.rust.ide.icons.RsIcons
+import org.rust.ide.notifications.showBalloon
+import org.rust.ide.utils.isNullOrEmpty
+
+class RsCargoAction(command: String) : AnAction() {
```
matklad: This could be `private class RsCargoAction(private val command: String)`

src/main/kotlin/org/rust/ide/actions/RsCargoAction.kt
```diff
@@ -0,0 +1,67 @@
...
+        val executor = ExecutorRegistry.getInstance().getExecutorById(DefaultRunExecutor.EXECUTOR_ID)
+        ProgramRunnerUtil.executeConfiguration(module.project, runConfiguration, executor)
+    }
+
+    private fun createRunConfiguration(module: Module, cargoCommandLine: CargoCommandLine): RunnerAndConfigurationSettings {
+        val runManager = RunManagerEx.getInstanceEx(module.project)
+
+        val configuration = runManager.createCargoCommandRunConfiguration(cargoCommandLine)
+        configuration.isTemporary = true
+        return configuration
```
matklad: A bit more idiomatic Kotlin would be 

```Kotlin
return runManager.createCargoCommandRunConfiguration(cargoCommandLine).apply { isTemporary = true }
```

src/main/kotlin/org/rust/ide/actions/RsCargoActionGroup.kt
```diff
@@ -0,0 +1,32 @@
...
+class RsCargoActionGroup : ActionGroup() {
+    val actions : Array<AnAction>
+
+    init {
+        templatePresentation.text = "_Cargo"
+        templatePresentation.description = "Run cargo command"
+        templatePresentation.icon = CargoIcons.ICON
+        templatePresentation.isEnabledAndVisible = true
+        isPopup = true
+        actions = arrayOf(RsCargoAction("clean"), RsCargoAction("check"), RsCargoAction("update"))
```
matklad: Hm, I think that `cargo check` and `cargo clean` would fail if executed in root virual package of the multi-package workspace. Let's add `--all` flag by default?

src/main/kotlin/org/rust/ide/actions/RsCargoActionGroup.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.openapi.actionSystem.ActionGroup
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import org.rust.cargo.icons.CargoIcons
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.util.modulesWithCargoProject
+import org.rust.ide.utils.isNullOrEmpty
+
+class RsCargoActionGroup : ActionGroup() {
+    val actions : Array<AnAction>
```
matklad: Could be private

matklad:
Closing in favor of https://github.com/intellij-rust/intellij-rust/pull/1670!

We now have a toolbar, yay!

PR #1375, opened 2017-06-26
Type infer item 2

matklad:


matklad:
@farodin91 this is basically slightly refactored #1242. I've tried to thread all the necessary information about type parameters via `Ty`s and `BoundElement`s, instead of relying on paths and trait refs.

matklad:
bors r+

closes #1242

PR #1377, opened 2017-06-27
(c)

matklad:


mkaput:
I think we could configure intellij to automatically manage license comments

matklad:
>I think we could configure intellij to automatically manage license comments

Yep, I've used the copyright plugin to actually update the copyrights. However, it is important to check them during CI, and looks like nothing can beat a little bash script here :) There are a couple of Gradle plugins for licenses there, but they all seemed kind of clunky to me :(

mkaput:
I have nothing against bash, just it would be nice if intellij would automatically insert license comment in new files :)

matklad:
Hm, yeah, I think I can try to figure out how to share copyright's plugin config...

mkaput:
It's stored in `.idea/copyright` if I recall correctly

matklad:
Yep, you are right! https://github.com/intellij-rust/intellij-rust/commit/4f40440a0033eaae0ed77418e382001011e5fa07

PR #1378, opened 2017-06-27
TYPE: Rainbow highlighter for variable

farodin91:
Has anyone a idea for a better icon?
screenshot from 2017-06-27 10-10-16

### DIFF
--- a/src/main/kotlin/org/rust/ide/highlight/RsRainbowVisitor.kt
+++ b/src/main/kotlin/org/rust/ide/highlight/RsRainbowVisitor.kt
@@ -1,0 +1,47 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.highlight
+
+import com.intellij.codeInsight.daemon.RainbowVisitor
+import com.intellij.codeInsight.daemon.impl.HighlightVisitor
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.descendantsOfType
+import org.rust.lang.core.psi.ext.parentOfType
+
+class RsRainbowVisitor : RainbowVisitor() {
+    override fun suitableForFile(file: PsiFile): Boolean = file is RsFile
+
+    override fun clone(): HighlightVisitor = RsRainbowVisitor()
+
+    override fun visit(element: PsiElement) {
+        val visitor = object : RsVisitor() {
+            override fun visitFunction(fn: RsFunction) = processFunctionVisitor(fn)
+        }
+        element.accept(visitor)
+    }
+
+    private fun processFunctionVisitor(fn: RsFunction) {
+        val patBindings = fn.descendantsOfType<RsPatBinding>()
+        val bindings = patBindings
+            .map { b -> b to patBindings.filter { it.name == b.name }.indexOf(b) }
+        val paths = fn.descendantsOfType<RsPath>()
+            .filter { it.parent !is RsCallExpr }
+        bindings.forEach { (first, second) ->
+            addInfo(first, first.referenceNameElement, second)
+        }
+        for ((first, second) in paths.map { it to it.reference.resolve() as? RsPatBinding }) {
+            val binding = second ?: continue
+            val num = bindings.find { it.first == binding }?.second ?: continue
+            addInfo(binding, first, num)
+        }
+    }
+
+    private fun addInfo(ref: RsPatBinding, rainbowElement: PsiElement, num: Int) {
+        addInfo(getInfo(ref, rainbowElement, "${ref.name}_$num", null))
+    }
+}
--- a/src/test/kotlin/org/rust/ide/highlight/RsRainbowVisitorTest.kt
+++ b/src/test/kotlin/org/rust/ide/highlight/RsRainbowVisitorTest.kt
@@ -1,0 +1,52 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.highlight
+
+import org.rust.lang.RsTestBase
+
+class RsRainbowVisitorTest : RsTestBase() {
+
+    fun checkRainbow(code: String, isRainbowOn: Boolean = true, withColor: Boolean = false) {
+        myFixture.testRainbow(
+            "main.rs",
+            code,
+            isRainbowOn, withColor)
+    }
+
+    fun testPathBinding() = checkRainbow("""
+        fn main() {
+            let mut <rainbow>test</rainbow> = "";
+            <rainbow>test</rainbow> = "";
+        }
+    """)
+
+    fun testFunctionBinding() = checkRainbow("""
+        fn foo() {}
+        fn main() {
+            foo();
+        }
+    """, true)
+
+    fun testDifferentColor() = checkRainbow("""
+        fn main() {
+            let mut <rainbow color='ff000002'>test</rainbow> = "";
+            <rainbow color='ff000002'>test</rainbow> = "";
+            let mut <rainbow color='ff000004'>test</rainbow> = "";
+            <rainbow color='ff000004'>test</rainbow> = "";
+        }
+    """, withColor = true)
+
+    fun testComplexDifferentColor() = checkRainbow("""
+        fn foo(<rainbow color='ff000002'>test</rainbow>: i32) {
+            let <rainbow color='ff000002'>x</rainbow> = <rainbow color='ff000002'>test</rainbow> + <rainbow color='ff000002'>test</rainbow>;
+            let <rainbow color='ff000002'>y</rainbow> = {
+               let <rainbow color='ff000004'>test</rainbow> = <rainbow color='ff000002'>x</rainbow>;
+            };
+            <rainbow color='ff000002'>test</rainbow>
+        }
+    """, withColor = true)
+
+}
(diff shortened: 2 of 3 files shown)


src/main/kotlin/org/rust/ide/icons/RsIcons.kt
```diff
@@ -68,6 +68,7 @@ object RsIcons {
     val IMPLEMENTING_METHOD = AllIcons.Gutter.ImplementingMethod!!
     val OVERRIDING_METHOD = AllIcons.Gutter.OverridingMethod!!
     val RECURSIVE_CALL = AllIcons.Gutter.RecursiveMethod!!
+    val VARIABLEN_SHADOWING = AllIcons.Gutter.Colors!!
```
mkaput: Variablen

mkaput:
Is there any language plugin that does such checks? I'm looking whether there is any precedence for icons for such gutter

matklad:
I feel that the underlying problem here should probably be solved with semantic highlighting: https://blog.jetbrains.com/pycharm/2017/01/make-sense-of-your-variables-at-a-glance-with-semantic-highlighting/

farodin91:
@matklad This idea looks good.

farodin91:
@matklad Do you have any example for a semantic highlight?

matklad:
@farodin91 I think `JavaRainbowVisitor` and `JavaRainbowVisitor` classes are worth looking into. 

By the way, I didn't know about this example beforehand, it's just possible to find it in the idea repository, following the trace starting from the linked blog post.

It says 

>This feature is disabled by default but can be found in Settings → Editor → Colors & Fonts → Language Defaults → Semantic highlighting.

And indeed there's a "Semantic highlighting" string in the IDEA UI. So it's possible to search for it: 

https://github.com/JetBrains/intellij-community/search?utf8=%E2%9C%93&q=Semantic+Highlighting&type=

Luckily, there aren't that many results, and there's only one entry in `.properties` file, which holds the UI string:

> color.settings.rainbow.demo.header.1=Semantic highlighting

So, it tells us that code word for this feature inside a platform is `ranbow`, and it also tells us the property key. After this step, we can either search for classes with `Rainbow` in class name, or we can search for `color.settings.rainbow.demo.header.1`, both of which bring us `com.intellij.codeHighlighting.RainbowHighlighter` which I think is an entry point to this whole subsystem. Hope this are some useful hints! :)

farodin91:
@matklad thank you

farodin91:
screenshot from 2017-06-29 08-38-29
@matklad Which variant do you like more?

matklad:
@farodin91 rainbow highlighting looks better :)

farodin91:
I will remove the gutter.

farodin91:
@matklad Any idea how to rewrite an test?

matklad:
Doesn't `RsHighlightingAnnotatorTest` work for this case?

matklad:
Hm, there's `com.intellij.testFramework.fixtures.CodeInsightTestFixture#testRainbow`.

farodin91:
I add a test

src/test/kotlin/org/rust/ide/highlight/RsRainbowVisitorTest.kt
```diff
@@ -0,0 +1,39 @@
...
+            let mut test = "";
+            test = "";
+            //X
+        }
+    """)
+
+    fun testFunctionBinding() = checkRainbow("""
+        fn foo() {}
+        fn main() {
+            foo();
```
matklad: Hm, shouldn't we have some `<rainbow>` tags in here? Like this: https://github.com/JetBrains/intellij-community/blob/fb1792bc5d19aa616429a26ac447b7f3cee8f355/java/java-tests/testSrc/com/intellij/java/codeInsight/daemon/LightRainbowHighlightingTest.java#L41

Does this test actually fail if there's a bug in rainbow highlighter? 

I thing we should also add more flawourful tests here, to make sure that we actually handle shadowing right.

Something like

```
fn foo(param: i32) {
   let x = param + param; // the first three params should be colored using the same color;
let y = {
   let param = x + y; // now *first* param should use  use a different color;
};
  param // and this shoulduse the original color. 
}

```

src/test/kotlin/org/rust/ide/highlight/RsRainbowVisitorTest.kt
```diff
@@ -0,0 +1,39 @@
...
+package org.rust.ide.highlight
+
+import org.intellij.lang.annotations.Language
+import org.rust.lang.RsTestBase
+import org.rust.lang.core.psi.RsPatBinding
+import org.rust.lang.core.psi.RsPath
+
+class RsRainbowVisitorTest : RsTestBase() {
+    fun checkRainbow(@Language("Rust") text: String) {
+        InlineFile(text)
```
matklad: Looks like `testRainbow` does `configureByFile` internally, so `InlineFile` does not have any effect.

farodin91:
@matklad Resolved!

src/test/kotlin/org/rust/ide/highlight/RsRainbowVisitorTest.kt
```diff
@@ -0,0 +1,52 @@
...
+            let mut <rainbow color='ff000004'>test</rainbow> = "";
+            <rainbow color='ff000004'>test</rainbow> = "";
+        }
+    """, withColor = true)
+
+    fun testComplexDifferentColor() = checkRainbow("""
+        fn foo(<rainbow color='ff000003'>param</rainbow>: i32) {
+            let <rainbow color='ff000003'>x</rainbow> = <rainbow color='ff000003'>param</rainbow> + <rainbow color='ff000003'>param</rainbow>;
+            let <rainbow color='ff000001'>y</rainbow> = {
+               let <rainbow color='ff000003'>param</rainbow> = <rainbow color='ff000003'>x</rainbow>;
```
matklad: Hm, ` <rainbow color='ff000003'>param</rainbow> ` does not look right... It's not the same param as in the `fn foo(param: i32)`, so it should probably use a different color.

matklad: But it works for me in the editor... I wonder why is that...

src/main/kotlin/org/rust/ide/highlight/RsRainbowVisitor.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    private fun processPatBindingVisitor(ref: RsPatBinding) {
+        addInfo(ref, ref.referenceNameElement, "${ref.referenceName}_${ref.textOffset}")
+    }
+
+    private fun processPathVisitor(element: RsPath) {
+        val ref = element.reference.resolve() ?: return
+        when (ref) {
+            is RsPatBinding ->
+                addInfo(ref, element, "${ref.referenceName}_${ref.textOffset}", RsColor.IDENTIFIER.textAttributesKey)
```
matklad: Why do we need to include a `textOffest` here? It causes all of the colors change after the char is typed, because `textOffset`s of all the following identifiers are shifted.

Local typing should not change colors of other variables.

farodin91: I add the offset to get a different color

matklad: Hm, that's interesting! looks like the upstream API is indeed unprepared for handling shadowing... However, prepending an offset is not an option, exactly because it changes all colors on typing. I wonder if we can come up with something more stabe than offset to disambiguate between shadowed names.

farodin91:
@matklad Both are resolved. I use the hashcode as the name.

farodin91:
I like the a idea of hashcode but the test is inconsistent.

farodin91:
My updated version is the name and the index of patbinding with the same name.

matklad:
Looks great now! The only thing I am worried about is that the current imple `PatBinding` inside a function. mentation is quadratic: we walk the whole function via `descendentsOfType` for each call to `addInfo`. Let's refactor this slightly, and use `visitFunction`, in the visitor, instead of `visitPath`/`visitPathBinding`.

Inside this `visitFunction`, we can collect all the shadowing info once (I think we need something like `Map<RsPatBinding, String>`), and then use it to highlight each binding.

farodin91:
@matklad Done. Code is a bit more complex now.

farodin91:
path doesn't work correct

farodin91:
@matklad Now path should work.

src/main/kotlin/org/rust/ide/highlight/RsRainbowVisitor.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.descendantsOfType
+import org.rust.lang.core.psi.ext.parentOfType
+
+class RsRainbowVisitor : RainbowVisitor() {
+    override fun suitableForFile(file: PsiFile): Boolean = file is RsFile
+
+    override fun clone(): HighlightVisitor = RsRainbowVisitor()
+
+    override fun visit(element: PsiElement) {
```
matklad: I think this can be `override fun visitFunction(fn: RsFunction)`?

farodin91: Do you think of this variant?

matklad: Ah, this is another kind of visitor altogether... Sorry for causing confusion then! Let's merge it, and I'll try to do some cleanups myself!

PR #1381, opened 2017-06-27
RES: resolve methods of derived traits

Undin:


src/main/grammars/RustParser.bnf
```diff
@@ -113,17 +113,32 @@ File ::= [ SHEBANG_LINE ] InnerAttr* Items
 // Attributes
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 
-fake Attr ::= MetaItem
+fake Attr ::= MetaItem {
+  extends = "org.rust.lang.core.psi.ext.RsStubbedElementImpl<?>"
+  stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
+}
```
matklad: Yep, stubbing all attributes seems like the right approach to me! And looks like we can now remove `val langAttribute: String?` from `RsTraitItemStub`, and `val pathAttribute: String?` from `RsModDeclItemStub`.

Undin: You want to save all `RsMetaItem` elements (identifier and literal value) into stub and use them, don't you?

matklad: Hm, yeah, perhaps we don't want to create stub nodes for all different Literal elements. I think we can store value as a `String?` for now!

Undin: Done

matklad:
bors r+

Thanks!

PR #1387, opened 2017-06-28
Trait impl unification

matklad:


matklad:
bros r+

mkaput:
Wip commit has slipped in

matklad:
hm, that's why I subconsciously typed bros =P

Undin:
It looks like bors hangs again

PR #1399, opened 2017-06-30
Executable and library rust project types

### LINKED ISSUE #1277: Provide option to create a library project
Currently there's no way to create a library project, not an executable one:

image

Undin:
Closes #1277 

Add checkbox into new project wizard to choose project/module type: executable or library.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
+++ b/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
@@ -36,11 +36,11 @@ class RustProjectSettingsPanel(private val cargoProjectDir: Path = Paths.get("."
     ) {
         fun applyTo(settings: RustProjectSettingsService) {
             settings.data = RustProjectSettingsService.Data(
-                toolchain,
-                autoUpdateEnabled,
-                explicitPathToStdlib,
-                useCargoCheckForBuild,
-                useCargoCheckAnnotator
+                toolchain = toolchain,
+                autoUpdateEnabled = autoUpdateEnabled,
+                explicitPathToStdlib = explicitPathToStdlib,
+                useCargoCheckForBuild = useCargoCheckForBuild,
+                useCargoCheckAnnotator = useCargoCheckAnnotator
             )
         }
     }
@@ -78,11 +78,11 @@ class RustProjectSettingsPanel(private val cargoProjectDir: Path = Paths.get("."
 
     var data: Data
         get() = Data(
-            RustToolchain(pathToToolchainField.text),
-            autoUpdateEnabled.isSelected,
-            (if (downloadStdlibLink.isVisible) null else pathToStdlibField.text.blankToNull()),
-            useCargoCheckForBuildCheckbox.isSelected,
-            useCargoCheckAnnotatorCheckbox.isSelected
+            toolchain = RustToolchain(pathToToolchainField.text),
+            autoUpdateEnabled = autoUpdateEnabled.isSelected,
+            explicitPathToStdlib = (if (downloadStdlibLink.isVisible) null else pathToStdlibField.text.blankToNull()),
+            useCargoCheckForBuild = useCargoCheckForBuildCheckbox.isSelected,
+            useCargoCheckAnnotator = useCargoCheckAnnotatorCheckbox.isSelected
         )
         set(value) {
             // https://youtrack.jetbrains.com/issue/KT-16367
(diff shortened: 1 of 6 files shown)


src/main/resources/META-INF/idea-only.xml
```diff
@@ -1,6 +1,5 @@
 <idea-plugin version="2">
     <extensions defaultExtensionNs="com.intellij">
-        <moduleBuilder builderClass="org.rust.ide.idea.RsModuleBuilder"/>
```
Undin: Do we really need this? I dropped it because `RsModuleBuilder` has not empty constructor now and IDEA fails to create it with reflection but I don't know the purpose of this extension (kotlin plugin doesn't have it too).
For me, the plugin works well without it:)

src/main/kotlin/org/rust/ide/idea/RsModuleBuilder.kt
```diff
@@ -9,29 +9,34 @@ import com.intellij.execution.ExecutionException
...
+    // make builder created by RsModuleType not available
+    // to prevent additional template in new project wizard
+    override fun isAvailable(): Boolean = builderType != ROOT
 
-    override fun getCustomOptionsStep(context: WizardContext, parentDisposable: Disposable): ModuleWizardStep =
-        CargoConfigurationWizardStep(context).apply {
-            Disposer.register(parentDisposable, Disposable { this.disposeUIResources() })
-        }
+    override fun createWizardSteps(wizardContext: WizardContext, modulesProvider: ModulesProvider): Array<ModuleWizardStep> =
+        arrayOf(CargoConfigurationWizardStep(wizardContext))
```
Undin: I'm not sure how to register `Disposable` here because IDEA doesn't pass `parentDisposable` into `createWizardSteps` method.

src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
```diff
@@ -94,11 +94,11 @@ class RustProjectSettingsPanel(private val cargoProjectDir: String = ".") : Disp
...
-            RustToolchain.suggest(),
-            true,
-            null,
-            false,
-            true
+            toolchain = RustToolchain.suggest(),
+            autoUpdateEnabled = true,
+            explicitPathToStdlib = null,
+            useCargoCheckForBuild = false,
+            useCargoCheckAnnotator = true
```
matklad: By the way, it is absolutely horrendous that there is this super boring similar code in like six places! I definitely over engineered it. If you have any ideas on how to remove the duplication, please go ahead!

src/main/kotlin/org/rust/ide/idea/RsModuleBuilder.kt
matklad: Frankly, I am not even sure we need to call `disposeUIresources` at all.

matklad:
The implementation looks great, however, I have a few questions about the general UI structure :) 

I am not really a fan of multistep wizards, and I in general I think the better approach is "create something runnable in a single click, and make it easy to change it later". So I would prefer if instead of this separate template step, we just had a `--bin` checkbox on the single config step we have now.

It might be rather challenging to add this checkbox, because we try to share the same form in several places. However, perhaps this is just wrong? I think the following options are just useless on this dialog, because they don't affect creation of the project in any way

new-project

So what do you thing about removing a reusable `RustProjectSettingsPanel` altogether, and building a custom UI for project creation step, reusing only some fields?

matklad:
>make it easy to change it later"

I think we have quite a few opportunities here to add some inspections/intentions to `Cargo.toml` file, like "add binary", "add library", "warning, file `foo.rs` does not belong to any crate, do you want to create a foo binary"

Undin:
Do you really want to remove "checkbox" settings from new project wizard and keep them only in project settings panel? Yes, these settings don't affect the creation of the project but they are project settings too and I think it can be annoying to change some setting(s) for every new project if you don't like default value.

Undin:
I dropped template wizard step and added checkbox for executable project creation.
I didn't remove other checkboxes because I'm not sure it's good decision but if you consider so, I'll move them from `RustProjectSettingsPanel` into `RustProjectConfigurable`

matklad:
>Do you really want to remove "checkbox" settings from new project wizard and keep them only in project settings panel?

Yeah I still think we don't need those checkboxes.

> it can be annoying to change some setting(s) for every new project if you don't like default value.

There's a little known feature that you can configure default settings for the project :) `Ctrl+Shift+A > Default Settings`. For non default settings, I think it doesn't make much difference if you configure them on creating a project, or afterwards via settings. 


That said, 

bors r+

because changing UI here is a separate task.

matklad:
Ah, one more thing though: let's keep `--bin` as the default? I think it makes more sense, and actually we want to switch to `--bin` by default in Cargo as well!

matklad:
(test failure is a spurious network error)

Undin:
> Yeah I still think we don't need those checkboxes.
Ok, I can remove them from new project wizard in next PR:)

Undin:
> Ah, one more thing though: let's keep --bin as the default? I think it makes more sense, and actually we want to switch to --bin by default in Cargo as well!

No problem:)
Done!

matklad:
Thanks!

PR #1411, opened 2017-07-03
Infer `Self` type for implicit trait implementation

Undin:


src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -56,8 +56,20 @@ fun inferExpressionType(expr: RsExpr): Ty {
...
-                .substitute(boundMethod.typeArguments)
+            val typeRef = method.retType?.typeReference
+            // if calling method belongs to some trait `Tr`
+            // and struct or enum doesn't have explicit implementation of this trait,
+            // for example, this trait is implemented with `derive` attribute
+            // and return type is `Self`
+            // then `typeRef.type` returns `Tr`
+            // although the real type of expression is type of struct or enum
+            // that implements trait `Tr`
+            val returnType = if (typeRef.isCselfRef) {
```
matklad: Hm, I think this won't work for a more complicated case like

```
trait BiClone {
    fn bi_clone(&self) -> (Self, Self)
}
```

What we really need is to add `Self -> StructType` mapping to the `boundMethod.typeArguments`, so that the original code of just `.substitute(mapping)` would do conversion of `Self` to an actual type.

Undin: Yes, it's more general solution)
I'll try to implement it.

Undin: Done

matklad:
bors r+

Thanks!

PR #1414, opened 2017-07-03
Expressions refactoring

Undin:
`inferExpressionType` is getting too big: at this moment it takes more than 200 lines.
I've split it into several functions for each expression type except one-liner expressions.


matklad:
Thank you!

bors r+

cc @vlad20012

PR #1418, opened 2017-07-04
Remove checkboxes from new project wizard

Undin:
Move all checkboxes which aren't retaled to project creation from `RustProjectSettingsPanel` into `RustProjectConfigurable` component.


matklad:
bors r+

Thanks!

PR #1421, opened 2017-07-04
TY: index operation must return 'Output' type instead of '&Output'

### LINKED ISSUE #1420: Result of indexing a vec given wrong type
Given the following code:
```rust
let x: Vec<u32> = vec![1, 2];
let a = x[0];
println!("food  {}", a);
```
the type of a should be u32. Intellij gives &u32.

Undin:
Closes #1420 

`container[index]` is syntactic sugar for `*container.index(index)` (see [index doc](https://doc.rust-lang.org/std/ops/trait.Index.html)) so type of `container[index]` must be `Output` instead of `&Output`.


matklad:
bors r+

:+1:

PR #1427, opened 2017-07-05
FMT: fix log-like macro formatting

matklad:


farodin91:
Working this for assert_eq?

PR #1428, opened 2017-07-05
TY: supported explicit type parameters it struct literals and function calls

vlad20012:
```rust
struct S<T1, T2> {a: T1, b: T2}
fn main() {
    let x = S::<u8, _>{a: 1, b: 2};
    (x.a, x.b)
  //^ (u8, i32)
}
```

```rust
struct S<T1, T2> (T1, T2);
fn main() {
    let x = S::<u8, _>(1, 2);
    (x.0, x.1)
  //^ (u8, i32)
}
```

```rust
fn f<T>(t: T) -> T { t }

fn main() {
    let a = f::<u8>(1);
    a
  //^ u8
}
```
```rust
fn f<T>(t: T) -> T { t }

fn main() {
    let f = f::<u8>;
    let r = f(1);
    r
  //^ u8
}
```


src/main/kotlin/org/rust/lang/core/psi/ext/RsStructLiteralField.kt
```diff
@@ -7,13 +7,15 @@ package org.rust.lang.core.psi.ext
...
 import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsFieldDecl
 import org.rust.lang.core.psi.RsStructLiteral
 import org.rust.lang.core.psi.RsStructLiteralField
 import org.rust.lang.core.resolve.ref.RsReference
 import org.rust.lang.core.resolve.ref.RsStructLiteralFieldReferenceImpl
 
 
 val RsStructLiteralField.parentStructLiteral: RsStructLiteral get() = parentOfType<RsStructLiteral>()!!
+val RsStructLiteralField.declaration: RsFieldDecl? get() = reference.resolve() as? RsFieldDecl
```
matklad: Let's call it `resolveToDeclaration` (compare with `RsTraitItem#resolveToTrait`)?

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -40,28 +40,38 @@ fun inferExpressionType(expr: RsExpr) = when (expr) {
...
 }
 
 private fun inferCallExprType(expr: RsCallExpr): Ty {
     val fn = expr.expr
     if (fn is RsPathExpr) {
-        val variant = fn.path.reference.resolve()
-        when (variant) {
-            is RsEnumVariant -> return inferTupleEnumTypeParameters(expr, variant)
-            is RsStructItem -> return inferTupleStructTypeParameters(expr, variant)
+        val boundElement = fn.path.reference.advancedResolve()
```
matklad: Hm, I was going to say that we should use `val (elemet, subst) = fn.path.refenrece.advancedResolve()` but looks like it's not going to work because of nulls. 

Still, let's make it a convention to destruct `BoundElement`s, and to use `subst` for the substitution. (And yeah, should probably rename `TypeArguments` to `Substitution`).

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
```diff
@@ -290,73 +300,38 @@ private fun getMoreCompleteType(t1: Ty, t2: Ty): Ty {
 }
 
 private val RsBlock.type: Ty get() = expr?.type ?: TyUnit
+private val RsStructLiteralField.type: Ty? get() = declaration?.typeReference?.type
```
matklad: `Ty?` seems smelly. We have a null-object, `TyUnknown`, and should probably use it here.

src/main/kotlin/org/rust/lang/core/types/infer/Expressions.kt
vlad20012: So nothing to fix here? (because of nulls)

matklad: Yep!

PR #1431, opened 2017-07-05
FMT: fix log-like macro formatting for assert_eq

farodin91:


mkaput:
I don't remember exactly, but wasn't there also accompanying `assert_neq!`?

matklad:
@mkaput there's also `debug_assert_eq/ne`  :) But I believe these all should be handled by this code!

Thanks @farodin91 !

bors r+

mkaput:
> But I believe these all should be handled by this code!

That's what I wanted to tell :)

PR #1454, opened 2017-07-08
QDOC: quick documentation for type parameters

Undin:
**NOTE**:
This implementation uses original text of type parameter bound to create quick doc. So if type parameter bound contains extra whitespace symbols:
```rust
fn foo<T: Into< String >> { ... }
```
quick doc will contain them too:
```
type parameter T: Into< String >
```
But quick doc for some other language items, for example, struct fields, has the same problem, so I think it should be fixed at the same time in other PR.


matklad:
bors r+

Thanks! I expected something like this after the `QDOC` PR :)

PR #1462, opened 2017-07-11
PERF: add a poor man's profiler

matklad:


src/main/kotlin/org/rust/utils/RsStopWatch.kt
```diff
@@ -0,0 +1,40 @@
...
+            override fun run() {
+                println("${this@RsStopWatch.name}: ${totalNs.get() / 1_000_000} ms")
+            }
+        })
+    }
+
+    fun <T> measure(block: () -> T): T {
+        var result: T? = null
+        totalNs.addAndGet(measureNanoTime { result = block() })
+        return result!!
```
mkaput: I think it should be good to note that returning `null` from block will result in NPE

matklad: Excellent catch!

matklad:
bors r=vlad20012

Issue #1459, opened 2017-07-11
Implement function move up or down with step over function


farodin91:
Would like to implement it?
@matklad Any suggestions to start?

matklad:
Is this about moving functions with `Ctrl+Shift+Up/Down`? I would **love** to have that! Not sure what extension points are relevant here, but there's `MethodUpDownUtil` in IntelliJ IDEA, which sounds like it might be an entry point to this stuff.

farodin91:
Yes is about moving functions with Ctrl+Shift+Up/Down.

farodin91:
MethodDownHandler in java could the function

PR #1463, opened 2017-07-11
TYPE: Add a StatementUpDownMover for function

farodin91:
Todo

- [x] function
- [x] struct
- [x] tests 
- [x] edge case
- [x] move documentation
- [x] more generic test much duplicated code
- [ ] more generic for example mod, impl, block, trait, if, loops, match pattern, macro pattern, macros, ...


farodin91:
Fixes #1459

src/test/kotlin/org/rust/ide/actions/RsFunctionMoverTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        """, """
+fn test() {
+
+}
+fn foo() {}
+        """, false)
+
+    fun moveTest(
+        @Language("Rust") before: String,
+        @Language("Rust") after: String, down: Boolean) {
```
matklad: Boolean parameters like `down: Boolean` often are code smell. Let's make two functions instead, `testMoveUp` and `testMoveDown`, to make it easier to find out what is supposed to happen at a glance!

matklad:
Looks nice! It looks like the logic here could grow very complex, so it would be helpful to have more tests here!

farodin91:
How to run format or ignore ident after an action? I didn't like the current test compare.

farodin91:
@matklad I will start refactor the code. Now I have unit tests.

farodin91:
Some refactor done.

farodin91:
@matklad Do you like to review the current state.

src/main/kotlin/org/rust/ide/actions/RsAbstractMover.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.ide.utils.findElementAtIgnoreWhitespaceAfter
+import org.rust.ide.utils.findElementAtIgnoreWhitespaceBefore
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsMod
+import org.rust.lang.core.psi.ext.parentOfType
+
+val PsiElement.line: Int? get() = containingFile.viewProvider.document?.getLineNumber(textRange.startOffset)
```
matklad: This probably wants to be private?

src/main/kotlin/org/rust/ide/actions/RsAbstractMover.kt
```diff
@@ -0,0 +1,90 @@
...
+
+abstract class RsAbstractMover : StatementUpDownMover() {
+
+    abstract val listOfContainers: List<Class<out PsiElement>>
```
matklad: Let's name it just `containers`. Usually it is not necessary to repeat the name of the type in the name of variable

src/main/kotlin/org/rust/ide/actions/RsAbstractMover.kt
```diff
@@ -0,0 +1,90 @@
...
+            info.toMove.endLine + 1
+        }
+        if (line < 0 || line >= editor.document.lineCount) {
+            return info.prohibitMove()
+        }
+        val offset = editor.document.getLineStartOffset(line)
+        var element = if (!down) {
+            file.findElementAtIgnoreWhitespaceBefore(offset) ?: return false
+        } else {
+            file.findElementAtIgnoreWhitespaceAfter(offset) ?: return false
```
matklad: `?: return false` can be pulled out of the `if` expression.

src/main/kotlin/org/rust/ide/actions/RsAbstractMover.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+        while (element != null && element !is RsFile) {
+            if (listOfContainers.any { it.isAssignableFrom(element.javaClass) }) {
```
matklad: This reflection does not look exactly pretty...

What do you think about using `IElementTypes` instead of classes (like `STRUCT_ITEM`). Alternatively, subclasses can provide a method like `isContainer(psi: PsiElement)`, and use a `when` inside.

farodin91:
Other `mover` could be also handled in a different PR.

src/main/kotlin/org/rust/ide/actions/RsAbstractMover.kt
```diff
@@ -0,0 +1,90 @@
...
+                        return info.prohibitMove()
+                    }
+                    parentOfFn = parentOfFn.parent
+                }
+            }
+
+            if (jumpOver.any { it.isAssignableFrom(element.javaClass) }) {
+                break
+            }
+            val context = element.context ?: break
```
matklad: Up/Down always operates with real files, so there's no need to use `context` here, `parent` is sufficient.

src/test/kotlin/org/rust/ide/actions/RsStructItemMoverTest.kt
```diff
@@ -0,0 +1,658 @@
...
+        """, """
+            // - main.rs
+            fn s() {
+                struct /*caret*/ A{
+                    test: u32
+                }
+            }
+        """)
+
+    fun moveUpTest(@Language("Rust") before: String, @Language("Rust") after: String) {
```
matklad: These functions could be `private`.

farodin91:
Resolved

farodin91:
@matklad Resolved merge conflict. Any thing to do?

farodin91:
I found a bug.

src/test/kotlin/org/rust/ide/actions/RsFunctionMoverTest.kt
```diff
@@ -0,0 +1,422 @@
...
+            // - main.rs
+            fn foo() {
+
+            }
+            fn /*caret*/test() {
+
+            }
+        """)
+
+    // This test fails in run intellij
```
farodin91: @matklad Any suggestions why this happened?

matklad: No, of the top of my head I don't know what happens here. So looks like the best way to debug this is to launch IDEA under debugger and just step through. 

However, the behavior in IntelliJ seems more reasonable: if there's a single blank line between two functions, we should swap them. Kotlin preserves all the whitespace between functions as well, so perhaps we should do the same?

src/main/kotlin/org/rust/ide/actions/RsAbstractMover.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.intellij.psi.PsiFile
+import com.intellij.psi.tree.IElementType
+import org.rust.ide.utils.findElementAtIgnoreWhitespaceAfter
+import org.rust.ide.utils.findElementAtIgnoreWhitespaceBefore
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.elementType
+
+val PsiElement.line: Int? get() = containingFile.viewProvider.document?.getLineNumber(textRange.startOffset)
+
+abstract class RsAbstractMover : StatementUpDownMover() {
```
matklad: Let's name it `RsAbstractUpDownMover`, to make it easier to discover via <kbd>Ctrl+N</kbd>? The descendants should probably have `UpDown` in their names as well.

src/test/kotlin/org/rust/ide/actions/RsFunctionMoverTest.kt
farodin91: I known, but test passes.

matklad: Yeah, that's a complete mystery to me as well!

Unfortunately, I have no better hints then step through in test and in live IDEA and to try to find out, what's different.

farodin91: I find the problem space are different handle to new lines.

matklad:
For this code

```
struct Foo {
    foo: u32,
}

impl Foo {}
```

moving the `impl` up won't jump over `struct`, but it should I think.

farodin91:
I only support currently structs and function.

matklad:
Ok, it can be handled in a separate PR!

Thanks!

bors r+

farodin91:
bors failed?

PR #1469, opened 2017-07-12
TY: first step to the full type inference

vlad20012:
This commit does not really improve a type inference. You can treat it as a large refactoring, or a first (baby) step to the full type inference.

In short, now we infer type of each expression in a function at once. Each function now has the property `inferenceContext`, which contains information about all types inside a function.


src/main/kotlin/org/rust/lang/core/completion/CompletionEngine.kt
```diff
@@ -84,7 +84,7 @@ fun RsCompositeElement.createLookupElement(scopeName: String): LookupElement {
             }
 
         is RsPatBinding -> base
-            .withTypeText(inferDeclarationType(this).let {
+            .withTypeText(this.type.let {
```
matklad: Probably don't need explicit `this` anymore. Also usually any code with `let` can be made a tiny bit better by avoiding `let` :)

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+import org.rust.lang.core.types.type
+import org.rust.utils.toIntOrNull
+
+fun inferTypesIn(fn: RsFunction): RsInferenceContext {
+    val ctx = RsInferenceContext()
+    TypeInferenceRecursionGuard.guarded { ctx.infer(fn) }
+    return ctx
+}
+
+class RsInferenceContext {
```
matklad: Not sure that we need the `Rs` prefix here: it is usually used to avoid clashes with classes in the platform. Or do we already have some `InferenceContext` in classpath?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+    val bindings: MutableMap<RsPatBinding, Ty> = mutableMapOf()
+    val exprTypes: MutableMap<RsExpr, Ty> = mutableMapOf()
```
matklad: Could be private?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+    private val RsBlock.ty: Ty get() = expr?.ty ?: TyUnit
+    private val RsStructLiteralField.type: Ty get() = resolveToDeclaration?.typeReference?.type ?: TyUnknown
+
+    fun walkBlock(block: RsBlock) {
+        for (stmt in block.stmtList) walk(stmt)
+        walkNullable(block.expr)
+    }
+
+    private fun walkNullable(psi: PsiElement?) {
+        if(psi != null) walk(psi)
```
matklad: <kbd>Ctrl+Shift+L</kbd>

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    private fun walkChildren(psi: PsiElement) {
+        for (psi1 in psi.children) walk(psi1)
```
matklad: I'm usually not a fan of `forEach`, but in this case, `psi.children.forEach(walk)` would probably be rather neat.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+                walkNullable(psi.expr)
+                ctx.collectBindings(psi.pat, psi.typeReference?.type ?: psi.expr?.ty ?: TyUnknown)
+            }
+            is RsCondition -> {
+                walkNullable(psi.expr)
+                ctx.collectBindings(psi.pat, psi.expr.ty)
+            }
+            is RsMatchArm -> {
+                for (pat in psi.patList)
+                    ctx.collectBindings(pat, psi.parentOfType<RsMatchExpr>()?.expr?.ty ?: TyUnknown)
```
matklad: Can we have braces here and in similar cases elsewhere ? :) 

https://www.imperialviolet.org/2014/02/22/applebug.html

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+            }
+            is RsForExpr -> {
+                walkNullable(psi.expr)
+                ctx.collectBindings(psi.pat, findIteratorItemType(psi.project, psi.expr?.ty ?: TyUnknown))
+                psi.block?.let { walkBlock(it) }
+                ctx.writeTy(psi, TyUnit)
+            }
+            is RsLambdaExpr -> {
+                val params = psi.valueParameterList.valueParameterList
+                params.forEach { p -> p.typeReference?.type?.let { ctx.collectBindings(p.pat, it) } }
```
matklad: If you need to introduce a dummy `p` name just to avoid nested lambdas, it might be better to write an explicit `for`-loop :)

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+}
+
+fun collectBindingsNullable(pattern: RsPat?, baseType: Ty): Map<RsPatBinding, Ty> {
+    return if (pattern != null) collectBindings(pattern, baseType) else emptyMap()
+}
+
+private object TypeInferenceRecursionGuard {
+    private val local: ThreadLocal<Boolean> = ThreadLocal.withInitial { false }
+
+    inline fun guarded(action: () -> Unit) {
```
matklad: Let's write a short docstring here explaining that this is for assertions only. Also, looks like we don't need `TypeInferenceRecursionGuard` wrapper object?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+        fieldType?.canUnifyWith(expr.ty, expr.project, argsMapping)
+
+    private fun inferArithmeticBinaryExprType(expr: RsBinaryExpr, op: ArithmeticOp): Ty {
+        val lhsType = expr.left.ty
+        val rhsType = expr.right?.ty ?: TyUnknown
+        return findArithmeticBinaryExprOutputType(expr.project, lhsType, rhsType, op)
+    }
+}
+
+fun collectBindingsNullable(pattern: RsPat?, baseType: Ty): Map<RsPatBinding, Ty> {
```
matklad: This could be private, and it has only one usage, so maybe inline it?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -0,0 +1,434 @@
...
+import org.rust.lang.core.resolve.ref.resolveFieldExprReferenceWithReceiverType
+import org.rust.lang.core.resolve.ref.resolveMethodCallReferenceWithReceiverType
+import org.rust.lang.core.types.ty.*
+import org.rust.lang.core.types.type
+import org.rust.utils.toIntOrNull
+
+fun inferTypesIn(fn: RsFunction): RsInferenceContext {
+    val ctx = RsInferenceContext()
+    TypeInferenceRecursionGuard.guarded { ctx.infer(fn) }
+    return ctx
```
matklad: Could use Kotlin's `apply` here.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
vlad20012: org.eclipse.jdt.internal.compiler.lookup.InferenceContext

matklad: Let's leave Rs then!

PR #1471, opened 2017-07-12
TY: infer type parameters from trait bounds

vlad20012:
```rust
struct S<X>(X);
trait Tr<Y> { fn foo(&self) -> Y; }
impl<Z> Tr<Z> for S<Z> { fn foo(&self) -> Z { unimplemented!() } }
fn bar<A, B: Tr<A>>(b: B) -> A { b.foo() }
fn main() {
    let a = bar(S(1));
    a
} //^ i32
```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
@@ -34,7 +34,16 @@ class TyTypeParameter private constructor(
         bounds.flatMap { it.flattenHierarchy }
 
     override fun canUnifyWith(other: Ty, project: Project, mapping: TypeMapping?): Boolean {
-        mapping?.merge(mutableMapOf(this to other))
+        if (mapping != null) {
+            val traits = findImplsAndTraits(project, other)
+            for (bound in bounds) {
+                val trait = traits.find { it.element.implementedTrait?.element == bound.element }
+                if (trait != null) {
+                    mapping.merge(bound.subst.reverse().substituteInValues(trait.subst))
+                }
+            }
+            mapping.merge(mapOf(this to other))
+        }
         return true
     }
 
--- a/src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
@@ -11,6 +11,8 @@ import org.rust.lang.core.psi.RsTypeParameter
 import org.rust.lang.core.psi.ext.bounds
 import org.rust.lang.core.psi.ext.flattenHierarchy
 import org.rust.lang.core.psi.ext.resolveToBoundTrait
+import org.rust.lang.core.resolve.asFunctionType
+import org.rust.lang.core.resolve.isAnyFnTrait
 import org.rust.lang.core.types.BoundElement
 
 class TyTypeParameter private constructor(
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -541,4 +541,14 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
...
     """)
+
+    fun `test infer generic argument from trait bound`() = testExpr("""
+        struct S<X>(X);
+        trait Tr<Y> { fn foo(&self) -> Y; }
+        impl<Z> Tr<Z> for S<Z> { fn foo(&self) -> Z { unimplemented!() } }
+        fn bar<A, B: Tr<A>>(b: B) -> A { b.foo() }
+        fn main() {
+            bar(S(1))
+        }    //^ i32
```
matklad: Let's change this a little bit such that it is not ambiguous which expression gets selected by `^` :

```
let x = bar(S(1)); x
                 //^ i32
```

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -34,7 +34,15 @@ class TyTypeParameter private constructor(
         bounds.flatMap { it.flattenHierarchy }
 
     override fun canUnifyWith(other: Ty, project: Project, mapping: TypeMapping?): Boolean {
-        mapping?.merge(mutableMapOf(this to other))
+        if (mapping != null) {
+            val traits = findImplsAndTraits(project, other)
+            for (bound in bounds) {
+                val trait = traits.find { it.element.implementedTrait?.element == bound.element }
+                if(trait != null)
+                    mapping.merge(bound.subst.reverse().substituteInValues(trait.subst))
```
matklad: spaces & braces ;)

matklad:
Looks good! 

By the way, is there any change in performance after that big refactoring?

vlad20012:
> By the way, is there any change in performance after that big refactoring?

No changes

vlad20012:
And more:
```rust
struct X;
fn apply<T1, T2, F: Fn(T1) -> T2>(t: T1, f: F) -> T2 {
    return f(t);
}

fn main() {
    let a = apply(X, |x| x);
    a
} //^ X
```

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -34,7 +36,27 @@ class TyTypeParameter private constructor(
         bounds.flatMap { it.flattenHierarchy }
 
     override fun canUnifyWith(other: Ty, project: Project, mapping: TypeMapping?): Boolean {
-        mapping?.merge(mutableMapOf(this to other))
+        if (mapping != null) {
```
matklad: Usually, it's better to avoid deep nesting and deal with short easy cases upfront:

```
if (mapping == null) return true
// Kotlin undestands that `mapping` is not null here
```

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -34,7 +36,27 @@ class TyTypeParameter private constructor(
...
+            if(other is TyFunction) {
```
matklad: <kbd>Ctrl+Alt+L</kbd>

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -34,7 +36,27 @@ class TyTypeParameter private constructor(
         bounds.flatMap { it.flattenHierarchy }
 
     override fun canUnifyWith(other: Ty, project: Project, mapping: TypeMapping?): Boolean {
```
matklad: I am not entirely sure, but it may be the case that we want two different `canUnifyWith`: one which is run during type infernece and which can mutate the type mapping, and the other which is run during `impl` search, which does not actually update anything.

vlad20012: In the near future all this will be replaced with normal type inference. I think this few lines will not worsen this set of hacks =)

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
vlad20012: Pfffkhhhbaaa, how can I be so inconsiderate...

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
vlad20012: You can rename this method (mentally, of course) to `tryUnifyWith` and `mapping` to something like `unificationTable`. In fact, if we checked a trait bounds, there would be a very similar code insted of `return true`

PR #1478, opened 2017-07-13
COMP: add match expression template

Avi-D:


matklad:
Yeah, we definitely could help writing `match`s and `if`s better! However I am not sure that live template is the best way forward here.

@Avi-D, are you aware of the postfix completion feature? That is, after typing `foo.match` in the expression context, this gets expanded to `match foo { }`, which is quite useful in practice. 

We also need a `prefix` template of sorts, but this can be more naturally handled by keyword completion over there: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt#L49-L75

That is, we can define completion providers for `if` and `match` keywords, which would add curly braces and place the cursor on the condition.

Avi-D:
I was not aware of postfix completion. It is very useful. I am new to intellij. I was also wondering about completions for `panic!()` and `#[derive(Debug)]`?

matklad:
`panic` will hopefully be in the next release, and for `derive` stuff, you can press <kbd>alt+enter</kbd> while the cursor is inside struct definition to invoke "add derive clause" intention, and inside `derive` usual completion works.

PR #1480, opened 2017-07-14
COMP: improve derive completion

### LINKED ISSUE #1465: Improve Derive attribute compltion
We have special cased completion for `derive` attribute

```
#[derive(|)]
struct S {
}
```

it suggests stuff like `Clone`, `Eq`, etc. I think it can be improved a little:

1) add suggestion to add `Clone, Copy`: this is a very often combo

2) the same for `PartialEq, Eq` combo.

3) remove suggestions for `RustcSerialize`, as it is superseded by serde.

The code is here:

https://github.com/intellij-rust/intellij-rust/blob/3c948dd52b61a6e537b0465a29e64326e48ec397/src/main/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProvider.kt

The tests are here:

https://github.com/intellij-rust/intellij-rust/blob/3c948dd52b61a6e537b0465a29e64326e48ec397/src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt

Undin:
Closes #1465 

* Drop suggestions of `RustcEncodable` and `RustcDecodable` traits
* Add completion variants with dependencies

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProvider.kt
@@ -8,6 +8,7 @@ package org.rust.lang.core.completion
 import com.intellij.codeInsight.completion.CompletionParameters
 import com.intellij.codeInsight.completion.CompletionProvider
 import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.completion.PrioritizedLookupElement
 import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.patterns.ElementPattern
 import com.intellij.patterns.PatternCondition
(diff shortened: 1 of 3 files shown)


Undin:
Two questions:
* How can I sort completion results? I think that `Copy, Clone` completion must be in front of single `Copy` suggestion. But maybe I'm wrong or it's not important.
* What do you think about checking which derivable traits are already implemented manually and skip then in completion suggestions?

Undin:
I've found answer at the question about suggestions sorting.

matklad:
>How can I sort completion results? I think that Copy, Clone completion must be in front of single Copy suggestion. But maybe I'm wrong or it's not important.

Yep, completion priority is the way to go for simple cases! 

> What do you think about checking which derivable traits are already implemented manually and skip them in completion suggestions?

Seems most reasonable to add!

matklad:
bors r+

Thanks!

Issue #1430, opened 2017-07-05
Reference in 'for in' loop erases types


jrmuizel:
In the following code e is resolved as (&K, &V) instead of (&u32, &u32)

``` rust
        let h: HashMap<u32, u32> = HashMap::new();
        for e in &h {
            e
        }
```

However, if we drop the reference on 'h', e is properly resolved to (u32, u32)
```rust
        let h: HashMap<u32, u32> = HashMap::new();
        for e in h {
            e
        }
```

farodin91:
I looked into it this problem, but looks like a huge problem. Erases of references and type parameter.

PR #1482, opened 2017-07-15
TY: `Self` & recursive receiver substitution (simple iterator chains works now!)

vlad20012:
1. `Self` type in `impl` block is now resolves to corresponding impl, not to particular type. Then `Self` type substitutes with with the receiver type. It allows to substitute complex types.
```rust
trait Tr<A> { fn wrap(self) -> S<Self> where Self: Sized { unimplemented!() } }
struct X;
struct S<C>(C);
impl<D> Tr<D> for S<D> {}
fn main() {
    let a = S(X).wrap().wrap().wrap();
    a
} //^ S<S<S<S<X>>>>
```
2. Now we substitute `impl` type parameters with receiver recursively (`RsImplItem.remapTypeParameters`). 
```rust
trait Tr<A> {
    fn wrap(self) -> S<Self> where Self: Sized { unimplemented!() }
    fn fold(self) -> A where Self: Sized { unimplemented!() }
}

struct X;
struct S1<B>(B);
struct S<C>(C);

impl<D> Tr<D> for S1<D> {}
impl<Src, Dst> Tr<Dst> for S<Src> where Src: Tr<Dst> {}

fn main() {
    let a = S1(X).wrap().wrap().wrap().fold();
    a
} //^ X
```
The magic is in line
```rust
impl<Src, Dst> Tr<Dst> for S<Src> where Src: Tr<Dst> {}
```
Many traits are implemented in a similar way, e.g. std iterators. And iterator map chains are working now!
```rust
struct S<T>(T);
fn main() {
    let test: Vec<String> = Vec::new();
    let a = test.into_iter()
    .map(|x| x.to_string())
    .map(|x| S(x))
    .map(|x| x.0)
    .next().unwrap();
    a
  //^ String
}
```
Also, the complex @matklad's test now passes (`test layered visitor example`).

3. This improvements resulted in a 40-fold performance regression... But it was enough to cache substituted (remapped) impls instead of raw. I.e. move caching from `RsImplIndex.findImpls` to `findImplsAndTraits`. Now it's even 15% faster than baseline.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/LangItems.kt
@@ -7,12 +7,9 @@ package org.rust.lang.core.resolve
 
 import com.intellij.openapi.project.Project
 import org.rust.lang.core.psi.RsFile
-import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.ext.*
-import org.rust.lang.core.resolve.indexes.RsImplIndex
 import org.rust.lang.core.types.BoundElement
-import org.rust.lang.core.types.infer.remapTypeParameters
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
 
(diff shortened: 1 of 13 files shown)


farodin91:
Do you fix with the pr also this issue #1430 ?

vlad20012:
Hmm, I just checked, and YES, it works!

vlad20012:
Included into #1516

PR #1483, opened 2017-07-15
Fixed associated type name

vlad20012:
Use of string interpolation is fraught with risks...


matklad:
>Use of string interpolation is fraught with risks...

Another common gotcha is that `"$foo"` works even if `foo` is `T?`.

:+1:

PR #1486, opened 2017-07-18
COMP: 'Where' keyword completion

Undin:


src/main/grammars/RustParser.bnf
```diff
@@ -562,6 +562,7 @@ private UnitStructTail ::= WhereClause? ';'
 private BlockStructTail ::= WhereClause? BlockFields
 
 EnumItem ::= AttrsAndVis enum identifier TypeParameterList? WhereClause? EnumBody {
+  pin = "enum"
```
matklad: 👍

src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -96,24 +95,48 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
...
+            .andNot(psiElement().withParent(RsTraitOrImpl::class.java))
+
+        val trait = psiElement<RsTraitItem>()
+            .withLastChildSkipping(RsPsiPattern.error, or(psiElement(IDENTIFIER), typeParameters))
+
+        val impl = psiElement<RsImplItem>()
+            .withLastChildSkipping(RsPsiPattern.error, psiElement<RsTypeReference>())
+
+        return psiElement()
+            .withPrevSiblingSkipping(RsPsiPattern.whitespace, or(function, struct, enum, typeAlias, trait, impl))
```
matklad: Ideally, we should use parser to provide keyword completion, like Kotlin does: https://github.com/JetBrains/kotlin/blob/master/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KeywordCompletion.kt#L199.

But pattern based approach should work quite well in practice!

matklad:
bors r+

Thanks!

matklad:
Looks like this needs some rebasing now because of https://github.com/intellij-rust/intellij-rust/commit/85cc8ce21185e61c62d4f398606ccb8fee9ecb63

Undin:
Done

PR #1490, opened 2017-07-20
TYPE: Move left/right action

Undin:


src/main/kotlin/org/rust/ide/actions/RsMoveLeftRightHandler.kt
```diff
@@ -0,0 +1,44 @@
...
+            is RsTypeParamBounds -> element.polyboundList
+            is RsTypeParameterList -> {
+                // We can't mix lifetime parameters with type parameters
+                // because lifetime parameters must be declared prior to type parameters
+                // and after moving type parameter before lifetime parameter
+                // code becomes invalid that prevents reverse action.
+                // So if there are more than one type parameters only type parameters will be returned
+                // because it's more common action. Otherwise lifetime parameters will be returned.
+                val typeParameters = element.typeParameterList
+                if (typeParameters.size > 1) typeParameters else element.lifetimeParameterList
```
Undin: @matklad How do you think if it's worth to make special psi nodes for `typeParameterList` and `lifetimeParameterList` to allow moving lifetimes and type params in the case when there are more than one lifetimes and more than one type params?

matklad: I think if we want to get fancy here, we should move in the opposite direction and allow to mix lifetime and type parameters in the parser, and use `RsSyntaxErrorsAnnotator` (split off of the Errors annotator yesterday) to add error messages.

matklad:
bors r+

Thanks!

src/main/kotlin/org/rust/ide/actions/RsMoveLeftRightHandler.kt
Undin: Good idea)
I can implement it

PR #1496, opened 2017-07-22
Delete branches after merging them with bors

notriddle:
By turning on this switch, bors will delete your branch after it merges it, as long as that branch is opened on the intellij-rust repository (it can't delete branches off of forks).


Undin:
Landed in #1517

matklad:
Thanks @notriddle and @khodzha, and sorry for taking so long to respond to the PR!

PR #1502, opened 2017-07-23
T: Some tests about function's gutter icon

la10736:


matklad:
Thanks a lot for the tests, @la10736 !

Sorry for taking so long to respond to the recent pull request and issues: there were some circumstances preventing me from contributing, but they are over for good :)

PR #1505, opened 2017-07-23
TYPE: Highlight exit points of the function

### LINKED ISSUE #778: Highlight exit points of the function
It is much easier to analyze the control flow of the function, if the highlighted (background / Gutter) all exit points. `return`, `{ re_val }`, `try!`, `panic!`, `?` and so on.

A la [Navigate To Function Exits](https://www.jetbrains.com/help/resharper/2016.2/Navigation_and_Search__Navigate_from_Here__Function_Exits.html)  from ReSharper.

farodin91:
Fixes #778

### DIFF
--- a/src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
+++ b/src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
@@ -13,6 +13,7 @@ import com.intellij.psi.PsiFile
 import com.intellij.util.Consumer
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.*
+import org.rust.lang.core.psi.ext.ancestors
 import org.rust.lang.core.psi.ext.elementType
 
 class RsHighlightExitPointsHandlerFactory : HighlightUsagesHandlerFactoryBase() {
@@ -21,31 +22,10 @@ class RsHighlightExitPointsHandlerFactory : HighlightUsagesHandlerFactoryBase() 
             return null
         }
         val elementType = target.elementType
-        if (elementType == RETURN || elementType == TRY_EXPR || elementType == Q) {
-            return RsHighlightExitPointsHandler(editor, file, target)
-        }
-        if (elementType == IDENTIFIER) {
-            val parent = target.parent
-            if (parent is RsMacroCall && (parent.referenceName == "panic" || parent.referenceName == "try")) {
-                return RsHighlightExitPointsHandler(editor, file, target)
-            }
-        }
-        if (testForExprBeforeBlock(target)) {
+        if (elementType == RETURN || elementType == Q) {
             return RsHighlightExitPointsHandler(editor, file, target)
         }
         return null
-    }
-
-    private fun testForExprBeforeBlock(target: PsiElement): Boolean {
-        var context: PsiElement? = target
-        while (context != null) {
-            val parent = context.parent
-            if (context is RsExpr && parent is RsBlock) {
-                return true
-            }
-            context = parent
-        }
-        return false
     }
 
     private class RsHighlightExitPointsHandler(editor: Editor, file: PsiFile, var target: PsiElement) : HighlightUsagesHandlerBase<PsiElement>(editor, file) {
@@ -55,15 +35,25 @@ class RsHighlightExitPointsHandlerFactory : HighlightUsagesHandlerFactoryBase() 
             selectionConsumer?.consume(targets)
         }
 
-        private fun getFunctionOrLambda(target: PsiElement): PsiElement? {
-            var context: PsiElement? = target
-            while (context != null) {
-                if (context is RsFunction || context is RsLambdaExpr) {
-                    return context
+        private fun getFunctionOrLambda(target: PsiElement): PsiElement? =
+            target.ancestors.first { it is RsFunction || it is RsLambdaExpr }
+
+        private fun onlyExpressionAboveUntilFn(expr: RsExpr): Boolean {
+            expr.ancestors.forEach {
+                val parent = it.parent
+                when (it) {
+                    is RsExpr -> when (parent) {
+                        is RsMatchExpr,
+                        is RsPatRange,
+                        is RsMatchArmGuard -> return false
+                    }
+                    is RsFunction,
+                    is RsLambdaExpr -> return true
+                    is RsCondition,
+                    is RsStmt -> return false
                 }
-                context = context.parent
             }
-            return null
+            return false
         }
 
         override fun computeUsages(targets: MutableList<PsiElement>?) {
@@ -76,7 +66,11 @@ class RsHighlightExitPointsHandlerFactory : HighlightUsagesHandlerFactoryBase() 
                 override fun visitFunction(o: RsFunction) {}
                 override fun visitLambdaExpr(o: RsLambdaExpr) {}
                 override fun visitRetExpr(o: RsRetExpr) = addOccurrence(o)
-                override fun visitTryExpr(o: RsTryExpr) = addOccurrence(o)
+
+                override fun visitTryExpr(o: RsTryExpr) {
+                    (o.expr as? RsMethodCallExpr)?.acceptChildren(this)
+                    addOccurrence(o.q)
+                }
 
                 override fun visitTryMacroArgument(o: RsTryMacroArgument) {
                     val macroCall = o.parent as? RsMacroCall ?: return
@@ -91,10 +85,15 @@ class RsHighlightExitPointsHandlerFactory : HighlightUsagesHandlerFactoryBase() 
                 }
 
                 override fun visitExpr(o: RsExpr) {
-                    if (o.parent is RsBlock) {
-                        addOccurrence(o)
-                    } else {
-                        super.visitExpr(o)
+                    when (o) {
+                        is RsIfExpr,
+                        is RsBlockExpr,
+                        is RsMatchExpr -> o.acceptChildren(this)
+                        else -> if (onlyExpressionAboveUntilFn(o)) {
+                            addOccurrence(o)
+                        } else {
+                            super.visitExpr(o)
+                        }
                     }
                 }
             })
--- a/src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
+++ b/src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
@@ -14,7 +14,6 @@ import com.intellij.util.Consumer
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.*
 import org.rust.lang.core.psi.ext.elementType
-import org.rust.lang.core.psi.ext.parentOfType
 
 class RsHighlightExitPointsHandlerFactory : HighlightUsagesHandlerFactoryBase() {
     override fun createHighlightUsagesHandler(editor: Editor, file: PsiFile, target: PsiElement): HighlightUsagesHandlerBase<*>? {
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -55,6 +55,7 @@
 
         <lang.syntaxHighlighterFactory language="Rust"
                                        implementationClass="org.rust.ide.highlight.RsHighlighterFactory"/>
+        <highlightUsagesHandlerFactory implementation="org.rust.ide.highlight.RsHighlightExitPointsHandlerFactory"/>
```
alygin: Shouldn't we add the `language="Rust"` attribute here?

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -0,0 +1,97 @@
...
+                return RsHighlightExitPointsHandler(editor, file, target)
+            }
+        }
+        if (testForExprBeforeBlock(target)) {
+            return RsHighlightExitPointsHandler(editor, file, target)
+        }
+        return null
+    }
+     private fun testForExprBeforeBlock(target: PsiElement) : Boolean {
+         var context = target
```
alygin: You need to specify the `context`'s type (`PsiElement?`) explicitly here. Otherwise, it's inferred as non-nullable `PsiElement` and NPE is thrown when `context = context.parent` hits `null`.

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -0,0 +1,97 @@
...
+     }
+
+    private class RsHighlightExitPointsHandler(editor: Editor, file: PsiFile, var target: PsiElement) : HighlightUsagesHandlerBase<PsiElement>(editor, file) {
+        override fun getTargets() = listOf(target)
+
+        override fun selectTargets(targets: MutableList<PsiElement>?, selectionConsumer: Consumer<MutableList<PsiElement>>?) {
+            selectionConsumer?.consume(targets)
+        }
+
+        override fun computeUsages(targets: MutableList<PsiElement>?) {
```
alygin: Some cases are handled incorrectly in this method. Here's a couple of problems:

```rust
fn foo() {
    let one = || { return 1 };
                  // ^ Caret.
    return  // <-- this one shouldn't be highlighted
}
```

The same problem with inner functions:

```rust
fn foo() {
    fn bar() {
        return  // <-- shouldn't be highlighted
    }
    return
     // ^ Caret
}
```

farodin91:
Resolved

src/main/resources/META-INF/plugin.xml
farodin91: Impossible

la10736:
Just to be sure that I've understand correctly your approach... Case like the following are not covered:

```
fn main() {
    let data = vec!["1", "pippo"];
    let results = data.iter().map(|s| Ok(s.parse::<i32>()?)).collect::<Vec<Result<_, std::num::ParseIntError>>>();

    println!("Results = {:?}", results);
}
```

and 

```
fn() {
    assert!(false);
    assert!(true);
}
```

Is there any way to find exit point without try to guess them by check every statement that can produce an exit point?

farodin91:
@la10736 For the first case, I have no idea what to highlight?

farodin91:
An another way, to highlight exit point would be a control flow analysis, but this would be a much complexer task.

la10736:
I think that what you've done till now is amazing. A control flow analysis can be the final solution that can check also when a called function can call a `panic!()`.  But now is this approach is ok.

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -0,0 +1,105 @@
...
+                override fun visitFormatMacroArgument(o: RsFormatMacroArgument) {
+                    val macroCall = o.parent as? RsMacroCall ?: return
+                    if (macroCall.referenceName == "panic") {
+                        addOccurrence(macroCall)
+                    }
+                }
+
+                override fun visitExpr(o: RsExpr) {
+                    if (o.parent is RsBlock) {
+                        addOccurrence(o)
```
alygin: If I understand correctly, you use this block to highlight the last expression in the function block. Unfortunately, it brings false highlights:

```rust
fn foo() {
    let a = { 10 };  // <-- shouldn't be highlighted
    loop {
        break  // <-- shouldn't be highlighted
    }
    return
    // ^ Caret
}
```

In general, highlighting implicit return looks nice in many functions. But if the expression is large, you might not want to have it highlighted. For instance, this is a quite common pattern with a long `match`-expression:

```rust
fn foo(a: u32) -> u32 {
    match a {
        0: 0,
        1..7: 1,
        8...21: 2,
        // many-many other arms
    }
} 
```

Maybe we can find a more intelligent way to handle such expressions? Maybe we shouldn't highlight them at all? Or only highlight "header"s of long expressions (words `match`, `if`, `loop`, function name, etc)?

Any ideas or examples from other languages that allow omit `return`s?

alygin: I had a look at how IntelliJ handles such cases for Groovy, which allows to omit `return`s the same way as Rust does.

It only highlights exit points when you put the cursor to a `return` word and doesn't highlight them when the cursor is inside an expression, even if this expression's result returns from the function:

screen shot 2017-07-24 at 14 25 54

screen shot 2017-07-24 at 14 26 07

screen shot 2017-07-24 at 14 26 23

It looks as a reasonable approach to me.

farodin91: I didn't like only to highlight with cursor on `return`. In rust is less used. I will try to import my current version.

alygin: The problem I want to address stems from the fact that in real projects, expressions are not so short and simple as you use in tests and in examples. Try to open a couple of real projects and move the cursor through the code. You'll see that very often the exit point highlighting spans huge parts of functions. One-liners and "one-expression"ers are always 100% highlighted. Here's an example from cargo:

screen shot 2017-07-24 at 18 14 27

It's just too much noise for the functionality you need only occasionally.

Moreover, this highlighting now subdues another very useful feature &mdash; symbol occurrence highlighting. Let's assume I want to find all the occurrences of the `self` argument in a function, so I put the cursor to the nearest `self` (another example from cargo):

screen shot 2017-07-24 at 18 16 41

Here, a) most of the function is green because of too eager exit points highlighting, so I don't see `self`s, and b) the first `self` occurrence is not highlighted at all.

As the result, the occurrences highlighting feature effectively becomes non-working. It doesn't look like a reasonable trade-off for having exit points highlighting. That's why I think we need to be careful here and try to find a better solution.

The Groovy-solution I described above is devoid of these problems because it doesn't highlight exit points when the cursor is placed inside an expression. Only when it's placed on keywords (`return`, `throw`). In Rust, things are more complicated because, for example, `panic!()` is both an exit point highlighting trigger and a valid part of an expression. What should we do when someone places the cursor on `panic!()`? Highlight exit points? But maybe I meant to highlight all the `painc!()`s in the file and don't want to see `return`s, `?`s and other exit-stuff. Let alone diverging functions (`fn foo() -> !`) that are also exit-points, but not covered in the current implementation.

So, the feature is much more complicated than it might seem at the first glance. My opinion is that it's better to start with something simple and safe (the Groovy-style).

farodin91: I'll try something

farodin91:
@alygin What do you think?

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -0,0 +1,115 @@
...
+                    }
+                }
+                when (context) {
+                    is RsFunction,
+                    is RsLambdaExpr -> return true
+                    is RsStmt -> return false
+                }
+                context = parent
+            }
+            return false
```
alygin: Same here: could be much shorter:

```kotlin
expr.ancestors.forEach {
    val parent = it.parent
    when (it) {
        is RsCondition -> return false
        is RsExpr -> when (parent) {
            is RsMatchExpr,
            is RsPatRange,
            is RsMatchArmGuard -> return false
        }
        is RsFunction,
        is RsLambdaExpr -> return true
        is RsStmt -> return false
    }
}
return false
```

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -0,0 +1,115 @@
...
+                override fun visitRetExpr(o: RsRetExpr) = addOccurrence(o)
+                override fun visitTryExpr(o: RsTryExpr) {
+                    val expr = o.expr
+                    when (expr) {
+                        is RsMethodCallExpr -> {
+                            addOccurrence(o.q)
+                            expr.acceptChildren(this)
+                        }
+                        else -> addOccurrence(o.q)
+                    }
```
alygin: Also could be shorter:

```kotlin
(o.expr as? RsMethodCallExpr)?.acceptChildren(this) 
addOccurrence(o.q)
```

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -0,0 +1,104 @@
...
+                    element?.acceptChildren(this)
+                }
+
+                override fun visitFunction(o: RsFunction) {}
+                override fun visitLambdaExpr(o: RsLambdaExpr) {}
+                override fun visitRetExpr(o: RsRetExpr) = addOccurrence(o)
+
+                override fun visitTryExpr(o: RsTryExpr) {
+                    (o.expr as? RsMethodCallExpr)?.acceptChildren(this)
+                    addOccurrence(o.q)
```
alygin: Maybe `addOccurrence(o)` would be better here. It's ok, because we don't trigger highlighting when we step onto this expression, just collect spans, so it won't overlap with the "find symbol occurrences" feature.

I think the ideal solution would be to highlight only the last function/method call in the path (to which the `?` really belongs), but the platform API doesn't allow arbitrary spans in this highlight handler (and highlighting two elements for one logical span would be incorrect), so I personally don't have strong preference here.

The current implementation also looks good to me, but feel free to return the previous logic in this place.

alygin:
Ideas for possible further improvements:

1. Treat calls to diverging functions as exit points.
2. Figure out a better way to highlight `?`s with the last function call in the expression.

@bors r+

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
farodin91: I try to highlight only the last part and it becomes too fast to large.

farodin91:
@alygin Any suggestions?

la10736:
I've got it this Sunday by `gradle test` on my laptop but sincerely I cannot understand why. If I remember well it happen since 1 over 2 times.

mkaput:
bors retry

mkaput:
bors r=@alygin

alygin:
This error pops up every now and then on my laptop. I guess it somehow depends on the `cargo check` response time, but I'm not sure. Let's try again.

bors r+

farodin91:
I will try to fix this issue in another PR.

la10736:
I guess that the problem is more frequent when the host is faster: on my slower work laptop it never happen but at home. Appveyor is quite fast and more frequent.

@farodin91 I tried yesterday but I wasn't able to reproduce it :cry:

farodin91:
@la10736 I tried yesterday but I wasn't also able to reproduce it :disappointed_relieved:

matklad:
Looks great to me @farodin91! The flaky is removed now: I am not sure how it is best to to test what I want to test there.

bors r+

Sorry for the long response time :(

PR #1515, opened 2017-07-26
FMT: Should not break the code by leave `use some::self`

### LINKED ISSUE #1512: Bug - removing {} from {self} leads to compilation error
Please don't remove {} around "self" word.
Also I would be very pleasure to see some setting "Never remove non-space symbols from my code, commas, brackets and so on".

la10736:
Fix #1512 and don't leave broken code.


src/test/kotlin/org/rust/ide/formatter/RsFormatterImportBracesTest.kt
```diff
@@ -12,6 +12,9 @@ class RsFormatterImportBracesTest : RsFormatterTestBase() {
     fun testWontRemoveBracesIfMultiImport() = doTextTest("use getopts::{optopt, optarg};",
         "use getopts::{optopt, optarg};")
 
+    fun `test wan't remove braces for single self`() = doTextTest("use getopts::{self};",
```
mkaput: won't

la10736: THX

PR #1521, opened 2017-07-27
Use real std::Result type in try expr and macro

vlad20012:
Previously we just compared type name with 'Result'


src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
```diff
@@ -422,4 +422,47 @@ class RsStdlibResolveTest : RsResolveTestBase() {
...
+            pub fn foo() -> IoResult<S> { unimplemented!() }
+
+        }
+
+        fn main() {
+            let s = io::foo()?;
+            s.field;
+              //^
+        }
+    """)
```
matklad: I wounder if we should do something similar to "ClosureResolveTest" here, so that you can test with without stdlib. 

Stdlib aware tests are slower, and much more difficult to debug if something goes wrong.

vlad20012: ClosureResolveTest uses lang attributes. Withoud stdlib we should place Result type into core::result module... Hmm, I did not not try it. It will probably work

matklad: I doubt it'll work out of the box, because one needs to bind a particular `core::result` module with Stdlib from the workspace, and IIRC workspace in these tests doesn't have stdlib at all, but we can easily fake it.

PR #1528, opened 2017-07-28
FMT: better code samples

### LINKED ISSUE #1129: Create better sample code for formatter
Our samples for formatter looks rather scary: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/resources/org/rust/ide/formatter/settings/wrap_sample.rs

It'd be great to some reduce the size of the file, and maybe make an example that makes a little more sense.

matklad:
closes #1129


matklad:
bors r+
 
@mkaput do you remember why we kept this samples in resources originally? Looks like just storing them inline is simpler.

mkaput:
I just personally prefer to store such stuff in resource files 😉

PR #1531, opened 2017-07-28
Allow mixing lifetime and type parameters

Undin:


src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -168,6 +169,17 @@ private fun checkValueParameter(holder: AnnotationHolder, param: RsValueParamete
     }
 }
 
+private fun checkTypeParameterList(holder: AnnotationHolder, element: RsTypeParameterList) {
+    val lifetimeParams = element.lifetimeParameterList
+    if (lifetimeParams.isEmpty()) return
+    val startOfTypeParams = element.typeParameterList.minBy { it.textOffset }?.textOffset ?: return
```
matklad: Will `.firstOrNull()?.textOffset` work here?

Undin: Can we be sure that list of children in psi element has same order that in code?
If yes then `.firstOrNull()?.textOffset` is the best variant.

matklad: Yep! `getFooList` works by traversing the Ast using `.getFirstChild()`, `.getNextSibling` methods and filtering by `IElementType`, so ordering is preserved. If the ordering was indeterminate, then `Collection` would be used for return type, and not `List`.

Undin: Cool!

Undin: Fixed

matklad:
bors r+

Thanks!

PR #1533, opened 2017-07-28
COMP: Improve derive completion 2

Undin:
* Fix relation between `Copy` and `Clone` traits
* Find manual implementations of derivable traits and skip them in completion suggestions

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
@@ -5,19 +5,33 @@
 
 package org.rust.lang.core.completion
 
+import com.intellij.testFramework.LightProjectDescriptor
 import org.rust.lang.core.resolve.StdDerivableTrait
 import org.rust.lang.core.resolve.withDependencies
 
 class RsDeriveCompletionProviderTest : RsCompletionTestBase() {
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
```diff
@@ -39,6 +43,15 @@ class RsDeriveCompletionProviderTest : RsCompletionTestBase() {
...
     """)
 
+    fun `test complete with manually implemented dependencies`() = checkSingleCompletion("Copy", """
+        #[derive(Cop/*caret*/)]
+        enum Foo { Something }
+
+        impl Clone for Foo {
+            fn clone(&self) -> Foo { Foo::Something }
+        }
+    """)
```
matklad: Let's use `doSingleCompletion` for new completion tests. It allows to verify caret placement as well, and is much less magical then `checkSingleCompletion`: https://github.com/intellij-rust/intellij-rust/blob/a900832640fefaacf1c698b5bac69479b0791567/src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt#L17-L31

Undin: No problem:)
Done

Issue #825, opened 2016-12-09
Line comment starting at block level


alexander-irbis:
With `Comment with Line Comment` action (`Ctrl+Slash`) it makes such commented lines with a comment starting at the begin of a line:
```rust
fn example(u: u8) -> char {
    match u {
//        0 => 'a',
//        1 => 'b',
        _ => '_',
    }
}
```
And then, after copy-paste or any other action followed by code formatting, it transforms into
```rust
fn example(u: u8) -> char {
    match u {
        //        0 => 'a',
        //        1 => 'b',
        _ => '_',
    }
}
```
So, the consistent behaviour for `Comment with Line Comment` action would be comments, starting at block level
```rust
fn example(u: u8) -> char {
    match u {
        // 0 => 'a',
        // 1 => 'b',
        _ => '_',
    }
}
```

x-ji:
Is this solved with #1972 ? Currently the comments still seem to be inserted at the first column, instead of spaces consistent with the original indentation.

PR #1534, opened 2017-07-29
FMT: Line comment starting at block level #825

andreycizov:
Fix for #825.


matklad:
Hm, looks like the fix is going to be a bit more involved. The problem is that for "comment some code out" we want to keep `//` on the beginning of lines, but for "comment some bit of code" we want to indent `//`. I believe this can be handled by com.intellij.formatting.Spacing#createKeepingFirstColumnSpacing function. The plugin's side of spacing handing is here: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt

Issue #1503, opened 2017-07-23
Gutter test icons for modules that contain tests


la10736:
It would be nice if we can provide a gutter icon to start test for modules too. On thinking about that I found that is little bit hard to find a good definition of what modules should have this icon and what not.

If we just say that every module that contains at least one test should have the test icon we will fall often in the cases where all modules have test icon (at least in my projects).

A pragmatic definition for _test module_ could be:
1. A module `test` or `tests` annotated by `#[cfg(test)]` that contain at lest a test function
2. A module contained in a _test module_ that have at least a test function
3. Have a test function means that the test function is in the module or in any of its submodules

Any thought?

If you are ok I can implement it.

alygin:
It would be great to have such an icon. But I think in practice, only the point 1 seems reasonable to implement. In other cases, you usually don't have a line to put the icon on. I mean, modules are usually files and I can't recall seeing a nested module that is not fully devoted to tests (and thus having the `#[cfg(test)]` attribute) but containing individual test functions.

In other languages, when the whole file is a test module, we usually have a file context menu item to run tests, not a gutter icon. But that would be a separate feature.

la10736:
@alygin 1, 2, 3 describe a recursive algorithm and not some possibility. Even when the module are file you have a line like
```rust
mod foo;
```
that include it.

Maybe is better to write an example:

`lib.rs`

```rust
mod out_no_test;

#[cfg(test)]
mod tests {
    #[test]
    fn foo_fn() {}
    mod inner_test {
         #[test]
         fn inner_test_fn()  {}
    }
    mod inner_no_test {
    }
    mod inner_file_test;
    mod inner_file_no_test;
}

mod no_test {
    fn some() {}
}
```
`inner_file_test.rs`
```rust
#[test]
fn bar() {}
```

Finally `inner_file_no_test.rs` doesn't contains any test function;

In this case we following gutter test icon on the mod lines:

* `tests`
* `inner_test`
* `inner_file_test`

Of course there are a lot of other corner cases that this example doesn't expose but I hope it is clear enough.

I'm using a lot of nesting in my tests because I find namespace a good way to group test topics and I can specialize the fixtures

alygin:
> Even when the module are file you have a line like `mod foo;`

Ah, that's what you mean. Yeah, I think it makes sense.

farodin91:
I would start with marking inner mods with a #[cfg(test)], because for me the use case is to run test for i'm working on.

[...]

PR #1535, opened 2017-07-29
CARGO: Support basic run tests in mod

farodin91:
Partial fix of #1503

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducer.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducer.kt
@@ -6,6 +6,7 @@
 package org.rust.cargo.runconfig.test
 
 import com.intellij.execution.Location
+import com.intellij.execution.PsiLocation
 import com.intellij.execution.actions.ConfigurationContext
 import com.intellij.execution.actions.RunConfigurationProducer
 import com.intellij.openapi.util.Ref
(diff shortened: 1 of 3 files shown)


farodin91:
@matklad Do you like to review my PR?

matklad:
oups, sorry, I've reviewed it yesterday and forget to submit the review=/

farodin91:
@matklad Resolved!

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducer.kt
```diff
@@ -70,26 +70,42 @@ class CargoTestRunConfigurationProducer : RunConfigurationProducer<CargoCommandC
...
 
         val testPath = mod.crateRelativePath.configPath() ?: ""
         val target = mod.containingCargoTarget ?: return null
-        if (!mod.functionList.any { it.isTest }) return null
 
         return TestConfig(mod, testName, testPath, target)
     }
 }
 
+fun PsiElement.testName(): String? {
```
matklad: I would rather try to avoid such all encompassing extensions methods: now every PSI element anywhere in the project has `testName` method. 

What do you think about the design similar to that of `ExecutableRunLineMarkerContributor`? https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/runconfig/command/CargoExecutableRunLineMarkerContributor.kt#L21

There, line marker calls into `CargoExecutableRunConfigurationProducer` which acts like a single source of truth for test discovery. Here, we won't be able to use a simple `Boolean` is test function, because we need a more rich return type with two variants "Nope, that's not a test!" and "Yep, it's a test, and it has a name "foo_test" ."

farodin91:
@matklad What do you think?

matklad:
Looks nice! I am somewhat not sure that abstracting on the `Location` level here is right, because we probably don't want to crawl the PSI tree upwards in line markers. But it is just a gut feeling, I don't see immediate problems with this approach. 

bors r+

Issue #1524, opened 2017-07-28
More intelligent hints


alygin:
Having [inlay parameter hints](https://github.com/intellij-rust/intellij-rust/pull/1326) is cool, but sometimes it can bee quite annoying. In mature languages, this feature uses additional rules to hide hints when they are redundant. I don't know all the rules, but here're a few that are obvious:

1. Local variable that is passed as an argument, has the same name as the parameter:

<img width="245" alt="screen shot 2017-07-28 at 09 47 11" src="https://user-images.githubusercontent.com/2101250/28705741-d948f7ec-7379-11e7-8373-04882011b0de.png">

2. There's a single parameter (not counting `self` in methods) and its name matches the function's name:

<img width="145" alt="screen shot 2017-07-28 at 09 52 20" src="https://user-images.githubusercontent.com/2101250/28705894-80b2f9f6-737a-11e7-904f-de246fb7dc7e.png">

3. Functions that start with `set_` and have single parameter (not counting `self` in methods):

<img width="172" alt="screen shot 2017-07-28 at 09 55 25" src="https://user-images.githubusercontent.com/2101250/28706001-ead537a4-737a-11e7-967e-a2e9c45008b3.png">

In all these cases hints could be safely omitted.

farodin91:
I like the idea of intelligent hints, but I would prefer to be able to switch between all and intelligent hints.

alygin:
Here's the exact rules that Java uses (they can be found in "Show parameter name hints -> Configure... -> Java") setting:

- White list of name patterns (a couple dozens by default, customizable)
- Do not show if method name contains parameter name
- Show for non-literals in case of multiple params with the same type (off by default)
- Do not show for builder-like methods

Issue #1522, opened 2017-07-28
Add type hint for lambda parameter


farodin91:
This should be configurable.

PR #1536, opened 2017-07-29
TYPE: Extend hints

farodin91:
Fixes: #1524 & #1522

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/configurable/RustProjectConfigurable.kt
+++ b/src/main/kotlin/org/rust/cargo/project/configurable/RustProjectConfigurable.kt
@@ -38,6 +38,8 @@ class RustProjectConfigurable(
     private val useCargoCheckAnnotatorCheckbox = JBCheckBox()
     private val letBindingHintCheckbox = JBCheckBox()
     private val parameterHintCheckbox = JBCheckBox()
+    private val smartHintCheckbox = JBCheckBox()
+    private val lambdaParameterHintCheckbox = JBCheckBox()
     private val cargoTomlLocation = Label("N/A")
 
     private var autoUpdateEnabled: Boolean
@@ -70,6 +72,18 @@ class RustProjectConfigurable(
             parameterHintCheckbox.isSelected = value
         }
 
+    private var lambdaParameterHint: Boolean
+        get() = lambdaParameterHintCheckbox.isSelected
+        set(value) {
+            lambdaParameterHintCheckbox.isSelected = value
+        }
+
+    private var smartHint: Boolean
+        get() = smartHintCheckbox.isSelected
+        set(value) {
+            smartHintCheckbox.isSelected = value
+        }
+
     override fun createComponent(): JComponent = panel {
         rustProjectSettings.attachTo(this)
         row(label = Label("Watch Cargo.toml:")) { autoUpdateEnabledCheckbox() }
@@ -77,8 +91,10 @@ class RustProjectConfigurable(
             row("Use cargo check when build project:") { useCargoCheckForBuildCheckbox() }
         }
         row(label = Label("Use cargo check to analyze code:")) { useCargoCheckAnnotatorCheckbox() }
-        row(label = Label("Show local variable type hints:")) { letBindingHintCheckbox() }
-        row(label = Label("Show argument name hints:")) { parameterHintCheckbox() }
+        row(label = Label("${HintType.LET_BINDING_HINT.option.name}:")) { letBindingHintCheckbox() }
+        row(label = Label("${HintType.PARAMETER_HINT.option.name}:")) { parameterHintCheckbox() }
+        row(label = Label("${HintType.LAMBDA_PARAMETER_HINT.option.name}:")) { lambdaParameterHintCheckbox() }
+        row(label = Label("${HintType.SMART_HINTING.name}:")) { smartHintCheckbox() }
         row("Cargo.toml") { cargoTomlLocation() }
     }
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
```diff
@@ -46,20 +47,95 @@ class RsInlayParameterHintsProviderTest : RsTestBase() {
...
+            let s = S;
+            S::foo(s, 10);
+        }    //^
+    """)
+
+    fun `test lamdba type hint`() = checkByText<RsLambdaExpr>("""
+        struct S;
+        fn with_s<F: Fn(S)>(f: F) {}
+        fn main() {
+            with_s(|s/*caret*/| s.bar())
```
farodin91: @alygin Work in really world.

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -50,6 +52,14 @@ enum class HintType(desc: String, enabled: Boolean) {
             } else {
                 0
             }
+            if (smart && declaration?.valueParameters?.size ?: 0 == 1) {
```
alygin: We could add

```rust
declaration ?: return emptyList()
```

above and get rid of most of those `?`s.

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -50,6 +52,14 @@ enum class HintType(desc: String, enabled: Boolean) {
...
+                if (declaration?.name == declaration?.valueParameters?.getOrNull(0)?.pat?.text) {
```
alygin: I found the exact rules that Java uses. There's `methodName.contains(paramName)`. I think it makes sense and covers a bit more cases. We could do the same.

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -59,14 +69,35 @@ enum class HintType(desc: String, enabled: Boolean) {
             = elem is RsCallExpr || elem is RsMethodCallExpr
 
         fun transform(index: Int, valueParameter: RsValueParameter, valueArgumentList: RsValueArgumentList): InlayInfo? {
-            if (index >= valueArgumentList.exprList.size)
+            if (index >= valueArgumentList.exprList.size) {
                 return null
+            }
             val pat = valueParameter.pat?.text ?: return null
+            if (smart) {
+                if (pat == valueArgumentList.exprList[index].text) {
```
alygin: These `if`s could be combined.

src/main/kotlin/org/rust/cargo/project/configurable/RustProjectConfigurable.kt
```diff
@@ -97,6 +113,8 @@ class RustProjectConfigurable(
         useCargoCheckAnnotator = settings.useCargoCheckAnnotator
         letBindingHint = HintType.LET_BINDING_HINT.enabled
         parameterHint = HintType.PARAMETER_HINT.enabled
+        lambdaParameterHint = HintType.LAMBDA_PARAMETER_HINT.enabled
+        smartHint = HintType.SMART_HINTING.get()
```
alygin: At the moment, we have several checkboxes in the Rust settings panel, and they use three different approaches to getting/setting their values. Is this variety necessary?

farodin91: I want to support settings in Clion for inlay parameter hints. But also support settings for inlay parameter like java or kotlin. You need to support getSupportedOptions.

src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
farodin91: @alygin Any suggestions to fix this problem?

farodin91:
@alygin it will fail

farodin91:
@alygin The lambda test fails

farodin91:
#1548

src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
farodin91: @matklad Any suggestions to fix this problem?

alygin:
Oh, sorry, I was in a hurry and didn't notice your question. I think the problem is that the test is not stdlib-aware, yet tries to use `Fn`. Just add the following line at the beginning of the class:

```kotlin
override fun getProjectDescriptor() = WithStdlibRustProjectDescriptor
```

That should fix it.

alygin:
Just a notice, not critical: Two versions of `checkNoHint()` look very similar. Maybe they could be combined into one with default parameter values?

src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
matklad: We need to provide an `Fn` lang item for this to work. Take a look how `org.rust.lang.core.resolve.RsClosuresResolveTest` manages to do it without depending on the acutal standard library (it's better to avoid stdlib aware tests, because they are slower).

Bascically, we need to add `#[lang="fn"] trant Fn<Args> { type Output; }` code to this one test case.

farodin91:
@matklad What do you think?

matklad:
bors r+

Thanks @farodin91 !

I am a bit not sure about the amount of the checkboxes we have in UI here, perhaps we should group then in a "Hints" pane or something, but this can be worked on later.

PR #1498, opened 2017-07-22
ANN: don't try to annotate non project files

Undin:
Closes #1497 

In `RsCargoCheckAnnotator` we check that message belongs to file by [equality of path suffix](https://github.com/intellij-rust/intellij-rust/blob/78754355e3cc823d75292f596d5c4ca1c17b26ad/src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt#L147-L148) with path from the message. But if the current file is not project file (from stdlib or dependency) and it has path with the same suffix, we calculate offsets for incorrect file and sometimes it can lead to invalid range.

PR #1540, opened 2017-07-29
Literals

matklad:


src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -147,18 +147,23 @@ private fun filterMessage(file: PsiFile, document: Document, message: RustcMessa
...
     }
 
     // The compiler message lines and columns are 1 based while intellij idea are 0 based
-    var textRange = TextRange(
-        toOffset(span.line_start - 1, span.column_start - 1) ?: return null,
-        toOffset(span.line_end - 1, span.column_end - 1) ?: return null
-    )
+    val startOffset = toOffset(span.line_start, span.column_start)
+    val endOffset = toOffset(span.line_end, span.column_end)
+    var textRange = if (startOffset != null && endOffset != null && startOffset < endOffset) {
```
Undin: But it doesn't solve the main problem, it only allows us to avoid invalid range.
The main problem is that we try to apply the result of `cargo check` to incorrect file with the same suffix of path.
I think we should differ project files from libraries files with the same path suffix and don't annotate incorrect files. In #1498 I suggest annotating only project files as a possible solution.

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -147,18 +147,23 @@ private fun filterMessage(file: PsiFile, document: Document, message: RustcMessa
     val spanFilePath = PathUtil.toSystemIndependentName(span.file_name)
     if (!file.virtualFile.path.endsWith(spanFilePath)) return null
 
+    @Suppress("NAME_SHADOWING")
     fun toOffset(line: Int, column: Int): Int? {
+        val line = line - 1
+        val column = column - 1
```
mkaput: Am I the only one who wishes every language had Haskell's treating `'` as valid identifier character, so it is possible to name variables `line'`, `column'` ?

matklad: Well, you could do it in Kotlin :) `` `line'` `` 

And actually for Haskell I would prefer if `ghc` didn't treat shadowing as a warning: this would greatly reduce the need for primes and the amount of bugs in my code :)

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
matklad: Oh my bad, you are absolutely right 😶 I think that we still better leave this check in, because we don't control the source of messages, but of course I should have merged #1498!

Undin: Yes, of course, these changes also allow us to avoid unexpected problems with invalid range in all possible cases:)

src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
mkaput: Well, my opinion seems to be totally opposite :)

I am currently working in Elixir and I don't remember any situation where I used shadowing except in `with` clause where pipe operator doesn't work:

```elixir
with \
  {:ok, foo} <- some_function(...),
  {:ok, foo} <- some_other_function(foo, ...)
do
    ...
end
```

What I don't like about such shadowing is, it is in fact *pure mutability* and can introduce nasty bugs when you forget that there is a shadow somewhere in a function.

matklad: >I am currently working in Elixir

\*green with envy\*

>can introduce nasty bugs when you forget that there is a shadow somewhere in a function.

In my own code, I think I find the opposite is true more often: without shadowing, I often have two variables of the same type in scope, like `foo` and `foo_some_suffix`, and then I use the original `foo` while I should have used `foo_some_suffix`, and even Haskell's type system does not help me. With shadowing, there's only one variable in scope, so you can't use the wrong one =P To me, shadowing is more about Bloch's "minimize visibility", rather then "mutability". 

Would be great to be able to measure and quantify the effects of shadowing though!

mkaput: > *green with envy*

It's funny (Ruby-syntax) but pretty nice.

But I miss real type system. For example dialyzer can't handle some parts of my work project (where I use lots of nested `:ok` tuples, lists and named types) and just doesn't type check them :(

Though I wish I could write NIFs in Rust! (Rustler looks amazing)

> I often have two variables of the same type in scope, like foo and foo_some_suffix, and then I use the original foo while I should have used foo_some_suffix

Well, it happens my *mistake processing unit* in my brain works opposite :)

PR #1544, opened 2017-07-31
Refactored Ty.canUnifyWith

vlad20012:
Refactored extremely hacky `Ty.canUnifyWith` method.

Now it's possible to implement precise/fuzzy impl lookup originally discussed at #1245 (also you can look at `RsPreciseTraitMatchingTest` TODO tests).

To do it we should somehow filter fuzzy matched impls after `findImplsAndTraits` call (currently there are no any flags and return type `Collection<BoundElement<RsTraitOrImpl>>` already seems quite complicated)

And I guess after it we'll be able to easily implement `impl<T> for T` resolution.


src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -33,14 +33,16 @@ class TyTypeParameter private constructor(
...
                     }
+                } else {
+                    result = result.merge(UnifyResult.fuzzy)
                 }
             }
         }
 
-        mapping.merge(mapOf(this to other))
-        return true
+        result = result.merge(mapOf(this to other))
```
matklad: `return result.merge...`

PR #1558, opened 2017-08-01
[WIP]: CARGO: fix proxy

### LINKED ISSUE #1021: Working behind proxy
Hello,
It seems that working behind proxy is not supported, am I wrong?
When I add dependency, they don't get downloaded, moreover the auto-completion is not working with the new dependencies, Although proxy setting is set in intellij idea.
I am on WINDOWS 8.1
Thanks in advance.

anton-okolelov:
fixes #1021

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -443,9 +443,6 @@ private fun RsMethodCallExpr.expectedParamsCount(): Pair<Int, Boolean>? {
         .takeIf { fn.owner.isInherentImpl }
 }
 
-private val RsCallExpr.declaration: RsFunction?
-    get() = (expr as? RsPathExpr)?.path?.reference?.resolve() as? RsFunction
-
 private fun isValidSelfSuperPrefix(path: RsPath): Boolean {
     if (path.self == null && path.`super` == null) return true
     if (path.path == null && path.coloncolon != null) return false
--- a/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
@@ -10,18 +10,11 @@ import com.intellij.codeInsight.hints.InlayInfo
 import com.intellij.codeInsight.hints.InlayParameterHintsProvider
 import com.intellij.codeInsight.hints.Option
 import com.intellij.psi.PsiElement
+import org.rust.ide.utils.CallInfo
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.selfParameter
-import org.rust.lang.core.psi.ext.valueParameters
-import org.rust.lang.core.types.infer.inferDeclarationType
 import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
-
-private val RsCallExpr.declaration: RsFunction?
-    get() = (expr as? RsPathExpr)?.path?.reference?.resolve() as? RsFunction
-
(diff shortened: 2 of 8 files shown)


src/main/kotlin/org/rust/cargo/toolchain/ProxyHelper.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.toolchain
+
+import com.intellij.util.net.HttpConfigurable
+import java.net.URI
+
+class ProxyHelper(var httpSettings: HttpConfigurable) {
```
matklad: Field could be private. Also, why do we need it as a constructor parameter? Perhaps it would be simpler to create a zero-argument constructor and call `HttpConfigurable.getInstance()` inside it?

src/main/kotlin/org/rust/cargo/toolchain/ProxyHelper.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    fun isHttpProxyEnabled(): Boolean {
```
matklad: Let's make it a property: `val isHttpProxyEnabled: Boolean get() = ... `

src/main/kotlin/org/rust/cargo/toolchain/ProxyHelper.kt
```diff
@@ -0,0 +1,27 @@
...
+        return httpSettings.USE_HTTP_PROXY && httpSettings.PROXY_HOST.isNotEmpty()
+    }
+
+    fun getProxyUri(): URI {
```
matklad: Could be a property as well

src/main/kotlin/org/rust/cargo/toolchain/ProxyHelper.kt
```diff
@@ -0,0 +1,27 @@
...
+        var userInfo: String? = null
+        if (httpSettings.PROXY_AUTHENTICATION
+            && httpSettings.proxyLogin!!.isNotEmpty() && httpSettings.plainProxyPassword != null) {
```
matklad: Hm, why are these `!!` safe? Is it because the user of this class required to call `isHttpProxyEnabled` before?

Could we fold both the checking and modification logic into this class?

Something like this:

```Rust
fun withProxyIfNeeded(cmd: GeneralCommandLine): GeneralCommandLine {
    if (!proxyEnabled) return cmd
    ....
}
```

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -109,6 +110,12 @@ class Cargo(
             .withParameters(commandLine.command)
             .withEnvironment(CargoConstants.RUSTC_ENV_VAR, rustExecutable.toString())
 
+        val proxyHelper = ProxyHelper(HttpConfigurable.getInstance())
+
+        if (proxyHelper.isHttpProxyEnabled()) {
+            cmdLine.withEnvironment("http_proxy", proxyHelper.getProxyUri().toString())
+        }
```
matklad: Let's add a unit-test style test for this? It could work along this lines:

1. Set proxy settings
2. Create a `Cargo` and `CargoCommandLine` instances
3. Call `toGeneralCommandLine`
4. Observe that proxy env var is present.

src/main/kotlin/org/rust/cargo/toolchain/ProxyHelper.kt
anton-okolelov: Ok, I'll fix that

anton-okolelov:
I'm sorry. I accidentally removed my commit and this PR has been closed automatically. Have no idea how to reopen it. I'm gonna create another one

PR #1564, opened 2017-08-02
T: Added option mysql_async test to RsHighlightingPerformanceTest

vlad20012:
There were BIG problems (OOME #1563) with type inference in mysql_async project, so i thought it's reasonable to embed it into our performance tests.

// Also fixed project dir refresh in test's VFS


matklad:
bors r+

Thanks!

PR #1567, opened 2017-08-02
Fix quick definition

### LINKED ISSUE #1555: Quick Definition (Ctrl+Shift+I) should show the actual definition
Looks like quick definition shows only the first line of the definition, which is quite odd!

image

Here, we should see the full `struct S` text.

Undin:
Closes #1555

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsNamedElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsNamedElement.kt
@@ -20,10 +20,10 @@ import org.rust.lang.core.stubs.RsNamedStub
 
 interface RsNamedElement : RsCompositeElement, PsiNamedElement, NavigatablePsiElement
 
+interface RsNameIdentifierOwner : RsNamedElement, PsiNameIdentifierOwner
 
 abstract class RsNamedElementImpl(node: ASTNode) : RsCompositeElementImpl(node),
-                                                   RsNamedElement,
-                                                   PsiNameIdentifierOwner {
+                                                   RsNameIdentifierOwner {
 
     override fun getNameIdentifier(): PsiElement? = findChildByType(IDENTIFIER)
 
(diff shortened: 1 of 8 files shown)


Undin:
Some notes/questions:
* I'm not sure that `ImplementationTextSelectioner` is the best place to fix the problem. For example, kotlin doesn't have `ImplementationTextSelectioner` at all. The main difference is that psi elements in kotlin plugin return themselves as navigation element [here](https://github.com/JetBrains/intellij-community/blob/d31f26bc3445b1276491ba2a0c163fbb2f87b94b/platform/lang-impl/src/com/intellij/codeInsight/hint/ImplementationViewComponent.java#L379) while we return identifier child element.
* Is there good way to test this? I think testing of `RsImplementationTextSelectioner` itself is not very useful because quick definition action invokes a lot of code besides `RsImplementationTextSelectioner`. So it would be great to get text from popup window but quick definition is not editor action and 
`myFixture.performEditorAction(IdeActions.ACTION_QUICK_IMPLEMENTATIONS)` fails because it can't show popup.

matklad:
>The main difference is that psi elements in kotlin plugin return themselves as navigation element here while we return identifier child element.

An excellent catch! Of course this is a bug: `navigationElement` is for the "I have a compiled class-file PSI element, but I want goto declaration to focus on the corresponding source element" use case. That is, it can substitute a completely unrelated PSI element. For the use case "I want goto definition to place cursor on the identifier", there's a `textOffset` method.

Looks like some two years ago someone failed to read the javadoc: https://github.com/intellij-rust/intellij-rust/commit/211da249631d3932a030ccc9bb05eef5d4987d50#diff-082a2875e15be79cb8aecf7d72c59e01R22. I think this is the oldest bug in the plugin so far, so congratulations, Indiana Jones achievement unlocked!

Undin:
Ready to review:)

src/main/kotlin/org/rust/lang/core/psi/ext/RsNamedElement.kt
```diff
@@ -20,10 +20,10 @@ import org.rust.lang.core.stubs.RsNamedStub
 
 interface RsNamedElement : RsCompositeElement, PsiNamedElement, NavigatablePsiElement
 
+interface RsNameIdentifierOwner : RsNamedElement, PsiNameIdentifierOwner
```
matklad: :+1: :+1: :+1:

matklad:
bors r+

This is super great!

PR #1576, opened 2017-08-05
TYPE: Classify usage type

farodin91:
@matklad What do you think?


matklad:
Excellent, @farodin91 !

bors r+

farodin91:
@matklad Would be better to replace naming by https://github.com/JetBrains/intellij-community/blob/306d705e1829bd3c74afc2489bfb7ed59d686b84/platform/usageView/src/com/intellij/usages/impl/rules/UsageType.java#L25 ?

matklad:
It would be better to reusing existing usage types, if they fit. But for something Rust-specific like `impl` we should invent our own usage type.

PR #1601, opened 2017-08-09
QDOC: links

Undin:
Support links in quick documentation.

### DIFF
--- a/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
@@ -5,6 +5,7 @@
 
 package org.rust.ide.docs
 
+import com.intellij.codeInsight.documentation.DocumentationManagerUtil
 import com.intellij.lang.documentation.AbstractDocumentationProvider
 import com.intellij.psi.PsiElement
 import org.rust.ide.presentation.escaped
@@ -45,10 +46,15 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
 
     private fun generateDoc(element: RsTypeParameter): String? {
         val name = element.name ?: return null
-        val typeBounds = element.bounds
-        val bounds = if (typeBounds.isEmpty()) "" else typeBounds.joinToString(" + ", ": ") { it.text.escaped }
-        val defaultValue = element.typeReference?.let { " = ${it.text}" } ?: ""
-        return "type parameter <b>$name</b>$bounds$defaultValue"
+        return buildString {
+            append("type parameter ")
+            b { it += name }
+            val typeBounds = element.bounds
+            if (typeBounds.isNotEmpty()) {
+                typeBounds.joinTo(this, " + ", ": ") { generateDocumentation(it) }
+            }
+            element.typeReference?.generateDocumentation(this, " = ")
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/docs/DocumentationUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+
+package org.rust.ide.docs
+
+import com.intellij.codeInsight.documentation.DocumentationManagerUtil
+import com.intellij.psi.PsiElement
+import org.rust.ide.utils.escaped
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+
+internal fun PsiElement.generateDocumentation(buffer: StringBuilder, prefix: String = "", suffix: String = "") {
```
mkaput: `internal` doesn't make sense in our codebase, as it means *private outside module* not *private outside package* as Java's `package-local` does

Let's just make it public, eventually leave a comment that this function is package internal

src/main/kotlin/org/rust/ide/docs/DocumentationUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+            buffer += "&gt;::"
+        }
+        else -> buffer += text.escaped
+    }
+    buffer += suffix
+}
+
+private fun generatePathDocumentation(element: RsPath, buffer: StringBuilder) {
+    val path = element.path
+    path?.let {
```
mkaput: nitpicking: `if(path != null)` maybe? seems slightly more readable to me

src/main/kotlin/org/rust/ide/docs/DocumentationUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+
+private fun RsPath.isLinkNeeded(): Boolean {
+    val element = reference.resolve()
+    // If it'll find out that links for type parameters are useful
+    // just check element for null
+    return !(element == null || element is RsTypeParameter)
+}
+
+private fun RsPath.link(): String {
+    val path = path
```
mkaput: ❤️

src/main/kotlin/org/rust/ide/docs/DocumentationUtils.kt
Undin: Hm, indeed.
Maybe drop `DocumentataionUtils` at all and move all functions into `RsDocumentationProvider` as private functions? It'll make `RsDocumentationProvider.kt` bigger but allow not to provide internal methods into global namespace.

Undin:
Comments are fixed

src/main/kotlin/org/rust/ide/docs/DocumentationUtils.kt
mkaput: seems fine to me

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -45,10 +47,22 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
 
     private fun generateDoc(element: RsTypeParameter): String? {
         val name = element.name ?: return null
+        val buffer = StringBuilder("type parameter ")
```
mkaput: I wish there were extension `+=` for `StringBuilder` in Kotlin like in Scala

matklad:
Hm, overall, I am not quite happy with the logic in `PresentaionUtil.kt` and `RsDocumentationProvider`: there are just too many casts, and everything looks extremely brittle. I don't have a good plan of how to fix this, except for writing more tests though :)

matklad:
Oups, sorry, I've pushed some refactorings in the documentation area and I've totally forgot that it'll conflict with this PR :(

Undin:
No problem:)
I'll just rebase

Undin:
@matklad Do you have cases what you want to check in tests? If yes I'll be glad to add them.
At this moment there are more than 100 tests for `RsDocumentationProvider` and in my opinion, they test all common cases.

matklad:
@Undin no, I think the current tests cover most of the reasonable use cases. The problem is, it's difficult to make sure that weird use-cases also work. Recently, after the `navigationElement` fix, similar code in Presentation utils start throwing assertion errors when opening a real project, despite the green tests: https://github.com/intellij-rust/intellij-rust/commit/9a93b97e9c8ac7ff4409c31495786c438739a854

I don't really know how we can improve situation here though, other then carefully scrutinize our PSI hierrarchy and try to extract some more focus intefaces to cut down on `when (randomPsiElement)` switches.

Issue #1548, opened 2017-07-31
Hide type hint if it is too long


vlad20012:
![image](https://user-images.githubusercontent.com/3221931/28794400-e2687da8-763e-11e7-8b70-66b3d44455b3.png)
I don't completely know what is "too long". May be optional max length?

farodin91:
@matklad Any suggestions?

mkaput:
I would place the limit of 10-15 chars and

1. If this is generic, try to replace most nested and then right-most generic arguments with `...

```rust
Filter<Map<...>, ...>
```

2. If it's not possible (just Java-esque name) , maybe just cut down whole string 

```
FilterSomethi...
```

3. Check out what Java implementation does ;)

farodin91:
I would replace it underscore not ..., because underscore is a bit more rusty.

polachok:
![screenshot from 2017-08-01 16-16-39](https://user-images.githubusercontent.com/94035/28827091-1d753026-76d5-11e7-971f-b5d256566687.png)

matklad:
Working with visitors...

![image](https://user-images.githubusercontent.com/1711539/28827287-c0180362-76d5-11e7-883b-59d0cbd23a08.png)

matklad:
Just to make sure folks are were, you can disable hints for local type parameters in settings: 
![image](https://user-images.githubusercontent.com/1711539/28827453-49bb5628-76d6-11e7-8ec8-efc3c409d9c2.png)

farodin91:
You can disable it also in parameter hints settings.

alygin:
> Working with futures:

> Working with visitors...

Zero-cost abstractions they said...

vlad20012:
> Zero-cost abstractions they said...
Awaiting [RFC 1522](https://github.com/rust-lang/rfcs/blob/master/text/1522-conservative-impl-trait.md)

matklad:
So I think that limiting this by the line length would be cool, but a bit difficult to implement, because you'll need to decide how to budget line length limit among children. I think a nice first solution would be to use the depth of a type to decide if we should print `_` or a full representation. 

That is, I imagine something like this:

```
fun compactPresentation(ty: Ty, maxDepth = 4) -> String
  if (maxDepth == 0) return "_"
  // switch on cases and recurse with maxDepth - 1
```

PR #1608, opened 2017-08-10
Shorten long types in hints

matklad:
closes https://github.com/intellij-rust/intellij-rust/issues/1548


farodin91:
@matklad You should implement this for lambda to.

matklad:
Do lambdas with huge parameter types occur in practice?

farodin91:
I have no idea, but it is cool to have the same behavior for both.

PR #1611, opened 2017-08-10
TypeInference refactoring

vlad20012:
Now traversing PSI manually, looking for each element type and its children we should process.

(Last preparations before implementing honest type inference)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -18,11 +18,11 @@ import org.rust.lang.core.types.type
 import org.rust.utils.forEachChild
 
 fun inferTypesIn(fn: RsFunction): RsInferenceContext =
-    RsInferenceContext().apply { recursionGuard { infer(fn) } }
+    RsInferenceContext().apply { preventRecursion { infer(fn) } }
 
 class RsInferenceContext {
-    private val bindings: MutableMap<RsPatBinding, Ty> = mutableMapOf()
-    private val exprTypes: MutableMap<RsExpr, Ty> = mutableMapOf()
+    private val bindings: MutableMap<RsPatBinding, Ty> = HashMap()
+    private val exprTypes: MutableMap<RsExpr, Ty> = HashMap()
 
     fun infer(fn: RsFunction) {
         extractParameterBindings(fn)
@@ -30,7 +30,7 @@ class RsInferenceContext {
         val block = fn.block
         if (block != null) {
             val items = StdKnownItems.relativeTo(fn)
-            RsFnInferenceContext(this, ImplLookup(fn.project, items), items).walkBlock(block)
+            RsFnInferenceContext(this, ImplLookup(fn.project, items), items).inferBlockType(block)
         }
     }
 
@@ -42,6 +42,10 @@ class RsInferenceContext {
 
     fun getExprType(expr: RsExpr): Ty {
         return exprTypes[expr] ?: TyUnknown
+    }
+
+    fun isTypeInferred(expr: RsExpr): Boolean {
+        return exprTypes.containsKey(expr)
     }
 
     fun getBindingType(binding: RsPatBinding): Ty {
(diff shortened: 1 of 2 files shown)


matklad:
Should we use HashMap/THashMap instead of linked one here?

https://github.com/vlad20012/intellij-rust/blob/7d917e3972aac420560f962c9dcdade642358155/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt#L24-L25

matklad:
Let's rename this recursion gurad to `[forbid|checkNo][Recursion|Reentrancy]`: https://github.com/vlad20012/intellij-rust/blob/7d917e3972aac420560f962c9dcdade642358155/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt#L21

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -44,6 +44,10 @@ class RsInferenceContext {
         return exprTypes[expr] ?: TyUnknown
     }
 
+    fun getExprTypeOrNull(expr: RsExpr): Ty? {
```
matklad: The only call-site does `== null`, so perhaps `fn alreadyProcessed(expt: RsExpt): Boolan`?

matklad:
It's a mild rule of thumb, but I usually find it better to push null-checks to the caller: https://github.com/vlad20012/intellij-rust/blob/7d917e3972aac420560f962c9dcdade642358155/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt#L59-L61. I can't really say why though, so I might be wrong about this one!

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -66,98 +70,66 @@ private class RsFnInferenceContext(
...
-        is RsCallExpr -> inferCallExprType(expr)
-        is RsDotExpr -> {
-            val methodCall = expr.methodCall
-            val fieldLookup = expr.fieldLookup
-            when {
-                methodCall != null -> inferMethodCallExprType(methodCall)
-                fieldLookup != null ->inferFieldExprType(fieldLookup)
-                else -> TyUnknown
-            }
+    private fun inferExprType(expr: RsExpr): Ty {
```
matklad: This could be a local extension property, so that `expr?.inferType :? TyUnknown` works.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -66,98 +70,66 @@ private class RsFnInferenceContext(
...
-                walkNullable(psi.expr)
-                ctx.writeTy(psi, if (ty is TyFunction) TyFunction(ty.paramTypes, psi.expr?.ty ?: ty.retType) else TyUnknown)
-            }
-            is RsExpr -> {
-                walkChildren(psi)
-                ctx.writeTy(psi, inferExpressionType(psi))
-            }
-            else -> {
-                walkChildren(psi)
+            is RsExprStmt -> {
```
matklad: No need for `{}` here.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -66,98 +70,66 @@ private class RsFnInferenceContext(
...
-        is RsDotExpr -> {
-            val methodCall = expr.methodCall
-            val fieldLookup = expr.fieldLookup
-            when {
-                methodCall != null -> inferMethodCallExprType(methodCall)
-                fieldLookup != null ->inferFieldExprType(fieldLookup)
-                else -> TyUnknown
-            }
+    private fun inferExprType(expr: RsExpr): Ty {
+        if (ctx.getExprTypeOrNull(expr) != null) error("Trying to infer expression type twice")
```
matklad: 👍

matklad:
To much stuff happens in this three lines: https://github.com/vlad20012/intellij-rust/blob/7d917e3972aac420560f962c9dcdade642358155/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt#L138-L140

I'd rather unpack them into something less dense :)

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -169,48 +141,66 @@ private class RsFnInferenceContext(
...
             else -> TyUnknown
-        }
+        }.substitute(subst).substitute(inferStructTypeArguments(expr))
+    }
+
+    private fun inferRsTupleExprType(expr: RsTupleExpr): Ty {
+        return TyTuple(expr.exprList.map { inferExprType(it) })
+    }
+
+    fun inferCastExprType(expr: RsCastExpr): Ty {
```
matklad: Looks like private is missing here, so even GH highlighting is confused :) And some other functions probably miss private as well.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -169,48 +141,66 @@ private class RsFnInferenceContext(
...
 
-        val methodType = method.type as? TyFunction ?: return returnType
+        val methodType = method.type as? TyFunction ?: return inferUnresolvedFnCallExprType(argExprs)
         // drop first element of paramTypes because it's `self` param
         // and it doesn't have value in `methodCall.valueArgumentList.exprList`
-        return returnType.substitute(mapTypeParameters(methodType.paramTypes.drop(1), methodCall.valueArgumentList.exprList))
+        return returnType.substitute(mapTypeParameters(methodType.paramTypes.drop(1), argExprs))
+    }
+
+    private fun inferUnresolvedFnCallExprType(argExprs: List<RsExpr>): Ty {
```
matklad: Hm, are we walking arguments if function call is resolved? Looks like there's a subtle invariant here, that on every exit arguments must be walked. Let's just walk arguments unconditionally on entering the function, and `return TyUnknown` elsewhere. 

The particular smell here is that this is a function, intended primarily for side effects, but whose return value is not Unit.

matklad: Yeah, we might want to introduce `RsCall`, such that `RsCallExpr: RsCall` and `RsMethodCall: RsCall`.

matklad:
Forgot an r+ 😅 

bors r+

PR #1614, opened 2017-08-11
Wrap LibraryRootsProvider in read action

matklad:


matklad:
Quite surprisingly, there is a case where roots provide is run outside the read action.

bors r+

PR #1615, opened 2017-08-11
QDOC: support path value parameters

Undin:
Support path value parameters in quick documentation.
It allows us to correctly show `FnOnce`, `FnMut` and `Fn` trait bounds.


matklad:
bors r+

Thanks!

PR #1620, opened 2017-08-12
Short types with ignore ref as level

### LINKED ISSUE #1613: Short hint type
Ignore ref as level. 
For example:
```rust
let option/*: &Option<&_> */ = &Some(&String::new());
let error/*: Option<Result<u64, E>> */ = Some(Ok(0u64));
```

farodin91:
Fixes #1613

### DIFF
--- a/src/main/kotlin/org/rust/ide/presentation/Ty.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/Ty.kt
@@ -42,7 +42,7 @@ private fun render(ty: Ty, level: Int): String {
 
         is TyTuple -> ty.types.map(r).joinToString(", ", "(", ")")
         is TyArray -> "[${r(ty.base)}; ${ty.size}]"
-        is TyReference -> "${if (ty.mutable) "&mut " else "&"}${r(ty.referenced)}"
+        is TyReference -> "${if (ty.mutable) "&mut " else "&"}${render(ty.referenced, level)}"
         is TyPointer -> "*${if (ty.mutable) "mut" else "const"} ${r(ty.referenced)}"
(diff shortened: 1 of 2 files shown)


farodin91:
@matklad r?

src/test/kotlin/org/rust/lang/core/type/RsShortTypeInferenceTest.kt
```diff
@@ -0,0 +1,39 @@
...
+            let s: S<(), ()> = unimplemented!();
+            let foo = s
+                .wrap(|x: i32| x + 1)
+                .wrap(|x: i32| x + 1)
+                .wrap(|x: i32| x + 1)
+                .wrap(|x: i32| x + 1);
+            foo;
+            //^ S<fn() -> <unknown>, S<fn() -> <unknown>, S<_, _>>>
+        }
+    """, short = true)
```
matklad: Let's use hint tests to check this behavior? Typification tests are about what types are actually inferred, and hint tests are more presentation oriented.

farodin91: I want to split this from the logic of hinting

matklad: Seems reasonable! Then let's add a test to `org.rust.lang.core.ext.RsPsiStructureTest`, which allows you to assert about the outcome of arbitrary function of PSI element. This way, we won't tie presentation and type inference tests.

farodin91: Done

PR #1621, opened 2017-08-12
TY: infer types with expectation

vlad20012:
The `Expectation` comes!
This is a next step to the real type inference. 
One advantage, tangible right now - lamda argument types now can be inferred from any contect, not only from function arguments.


vlad20012:
r? @matklad 
And can it be better to move these tests to another class?

PR #1625, opened 2017-08-13
GRD: gradle 4.1

Undin:


matklad:
Yeah, this is now in the master, and the instructions about upgrading gradle are added to the new MAINTAINING.md: it's a bit tricky. Sorry for not commenting here before: I've locked myself out of GitHub's UI today via two-factor authentication =P

PR #1628, opened 2017-08-14
TY: fixed inference consts and statics inside functions

vlad20012:
Fixed:
```rust
fn main() {
    const X: i32 = 1;
}                //^ i32
```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/Extensions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/Extensions.kt
@@ -6,6 +6,7 @@
 package org.rust.lang.core.types
 
 import com.intellij.openapi.util.Computable
+import com.intellij.psi.PsiElement
 import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
 import com.intellij.psi.util.PsiModificationTracker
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/types/Extensions.kt
```diff
@@ -39,11 +40,14 @@ val RsFunction.inferenceContext: RsInferenceContext
         CachedValueProvider.Result.create(inferTypesIn(this), PsiModificationTracker.MODIFICATION_COUNT)
     })
 
+val PsiElement.inferenceContext: RsInferenceContext?
```
matklad: Let's make this private and move after the public API of the file?

src/test/kotlin/org/rust/lang/core/type/RsExpressionTypeInferenceTest.kt
```diff
@@ -635,4 +635,16 @@ class RsExpressionTypeInferenceTest : RsTypificationTestBase() {
...
+        fn main() {
+            const X: i32 = 1;
+        }                //^ i32
+    """)
+
+    fun `test infer static inside a function`() = testExpr("""
+        fn main() {
+            const X: i32 = 1;
+        }                //^ i32
+    """)
```
matklad: These tests are letterwise identical =)

vlad20012: Different keywords! =D
Ok...

vlad20012: Wow... really... hmmm...

vlad20012:
Done

Issue #1599, opened 2017-08-09
Wrong type from match result with return statement


jrmuizel:
Intellij thinks the type 'full' is () when it should be u32

```rust
enum Buck {
    Full(u32),
    Empty,
}

fn peek() -> Buck {
    Buck::Empty
}

fn stuff() -> u32 {
    let full = match peek() {
        Buck::Empty => {
            return 0;
        }
        Buck::Full(bucket) => bucket,
    };
    full
}
```

vlad20012:
Closed with #2010

PR #1629, opened 2017-08-14
TY: implemented `!` type

vlad20012:
`!` type (TyNever) is result of `return`, `break` & `continue` exprs, and also `panic!()`, `unimplemented!()` and so on. Rust handles `!` in a special way, e.g. allow it to become any type, or ignore match arm that return it.

It isn't fixing #1599 completely, but if remove semicolon after `return 0` in that example,  it will work.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/ty/TyPrimitive.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/TyPrimitive.kt
@@ -56,6 +56,11 @@ object TyUnit : TyPrimitive {
     override fun toString(): String = tyToString(this)
 }
 
+/** The `!` type. E.g. `unimplemented!()` */
+object TyNever : TyPrimitive {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/types/ty/TyPrimitive.kt
```diff
@@ -56,6 +56,11 @@ object TyUnit : TyPrimitive {
     override fun toString(): String = tyToString(this)
 }
 
+/** The `!` type. E.g. `unimplemented!()` */
+object TyNever : TyPrimitive {
+    override fun toString(): String = tyToString(this)
```
matklad: `tyToString` should be updated as well. I wonder whether we should make `TyPrimitive` a sealed class... IIRC, it is not possible for some stupid reason, but I may be recalling wrgon!

PR #1634, opened 2017-08-16
Cargo command editor

matklad:
Add UI component for adding Cargo commands with completion. Currently it is used only in **Run Cargo Command** action, but the plan is to plug it into run configurations as well.

https://zippy.gfycat.com/FixedGiantAiredaleterrier.webm

r? @vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/cargo/util/CargoCommandCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/cargo/util/CargoCommandCompletionProvider.kt
@@ -1,0 +1,52 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.util
+
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.lookup.LookupElement
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.util.TextFieldCompletionProvider
+import com.intellij.util.execution.ParametersListUtil
+import org.rust.cargo.project.workspace.CargoWorkspace
+import org.rust.utils.buildList
+
+class CargoCommandCompletionProvider(
+    private val workspace: CargoWorkspace?
+) : TextFieldCompletionProvider() {
+    override fun addCompletionVariants(text: String, offset: Int, prefix: String, result: CompletionResultSet) {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/cargo/util/CargoCommandCompletionProvider.kt
```diff
@@ -0,0 +1,257 @@
...
+
+    override fun acceptChar(c: Char): CharFilter.Result? =
+        if (c == '-') CharFilter.Result.ADD_TO_PREFIX else null
+
+    override fun addCompletionVariants(text: String, offset: Int, prefix: String, result: CompletionResultSet) {
+        val (ctx, _) = splitContextPrefix(text)
+        result.addAllElements(complete(ctx))
+    }
+
+    // public for testing
```
mkaput: https://upsource.jetbrains.com/idea-ce/file/idea-ce-200f59d7cf3b0f66feb3a4abebbb90864dc5edc7/platform/annotations/common/src/org/jetbrains/annotations/TestOnly.java

matklad: Hm, this is called not only in tests, but in implementation as well. So what we really want here is https://google.github.io/guava/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html, but Guava annotations are not generally used in IntelliJ.

vlad20012: You can make it private and add separate TestOnly accessors

mkaput: > Hm, this is called not only in tests, but in the implementation as well. So what we really want here is https://google.github.io/guava/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html, but Guava annotations are not generally used in IntelliJ.

Didn't read carefully what `@TestOnly` means, heh

> You can make it private and add separate TestOnly accessors

Nah, too much hassle 😉

matklad: Hm, generally this sounds like an excellent approach, but I am afraid I am too lazy in the end :)

matklad:
bors r=vlad20012

Issue #1632, opened 2017-08-15
Overhaul Run Configurations for Cargo


matklad:
Our current run configurations basically work, but looks like there's a big room for improvement. 

To me as a user of IntelliJ Rust the biggest problem UI-wise is a mismatch between IDEA's GUI and Cargo's usual command line. In particular, I always mess up `Command` and `Additional arguments` fields and write args in the first field.

So the biggest change is to merge `Command` and `Additional arguments` fields into a single field with autocomplete for command names themselves common static options like `--release` and common dynamic options like `--bin name` or `--package name`.  The same should work for **Run Cargo Command** action.

At the same time there are a number of bugs which should be fixed:

* Control for propagating parent environmental variables does not work. 
* There's no control for working directory.

PR #1637, opened 2017-08-17
Reorganize CargoCommand Run configuration

matklad:
This simplifies `CargCommadConfiguration` a bit by switching from a reflection-based XMLSerializer to hand written serialization code, and by keeping stored data closed to the XML/GUI. As a bonus, "Include parent environment variables" now works properly.

This isn't ready for merging yet and will break tests under 2017.1. I'll come back to this after RustConf.


matklad:
Part of https://github.com/intellij-rust/intellij-rust/issues/1632

notriddle:
@matklad You're going to need to run it again. [We went down](https://bors.tech/pants-on-head-stupid/downtime/retrospective/2017/08/24/we-were-down/).

PR #1642, opened 2017-08-21
INT: `Fill match arms` intention

### LINKED ISSUE #1551: "Fill match arms" intention.
For an empty `match`, add an intention to prefil match arms with enum variants, like

```Rust
fn foo(x: Option<i32>) {
    match x {
        /*caret*/
    }
}
```

->

```Rust
fn foo(x: Option<i32>) {
    match x {
        Some(_) => { /*caret*/ }
        None => {}
    }
}
```

Undin:
Closes #1551


matklad:
Splendid!

bors r+

PR #1648, opened 2017-08-23
COMP: Fix function self ref and double colon

### LINKED ISSUE #1639: self:: autocompletion in function signatures
[Imgur](http://i.imgur.com/w0AJFIE.png)

Notice the colons.

This happens even outside impl blocks. I don't think any other paths trigger it; only `self::`.

edit:
PyCharm Pro 2017.2.1
intellij-rust 0.1.0.2065
rustc 1.19.0 (0ade33941 2017-07-17)

farodin91:
Fixes #1639 
@matklad @Undin r?

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -227,9 +227,7 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
         if (Namespace.Types in ns) {
             if (processor("self", containingMod)) return true
             val superMod = containingMod.`super`
-            if (superMod != null) {
-                if (processor("super", superMod)) return true
-            }
+            if (superMod != null && processor("super", superMod)) return true
         }
     }
 
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -11,6 +11,7 @@ import com.intellij.codeInsight.completion.PrioritizedLookupElement
 import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.openapi.editor.EditorModificationUtil
+import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.icons.RsIcons
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
@@ -27,8 +28,10 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
         is RsMod -> if (scopeName == "self" || scopeName == "super") {
             base.withTailText("::")
                 .withInsertHandler({ ctx, _ ->
-                    val offset = ctx.editor.caretModel.offset
-                    if (ctx.file.findElementAt(offset)?.parentOfType<RsUseGlobList>() == null) {
+                    val offset = ctx.tailOffset - 1
+                    val inUse = PsiTreeUtil.findElementOfClassAtOffset(ctx.file, offset, RsUseGlobList::class.java, false) != null
+                    val inSelfParam = PsiTreeUtil.findElementOfClassAtOffset(ctx.file, offset, RsSelfParameter::class.java, false) != null
+                    if (!(inUse || inSelfParam)) {
                         ctx.addSuffix("::")
                     }
                 })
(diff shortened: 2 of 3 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -238,6 +242,15 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
     return processNestedScopesUpwards(path, processor, ns)
 }
 
+private inline fun <reified T : PsiElement> PsiElement.contextOfType(strict: Boolean = true): T? {
```
Undin: Looks like we already have this method [here](https://github.com/intellij-rust/intellij-rust/blob/1a1ea675f6c72db2358845518c85986b879aa394/src/main/kotlin/org/rust/lang/core/psi/ext/PsiElement.kt#L35-L36)

src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -8,6 +8,7 @@ package org.rust.lang.core.resolve
 import com.intellij.codeInsight.lookup.LookupElement
 import org.rust.lang.core.completion.createLookupElement
 import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsValueParameter
```
Undin: Unused import

farodin91:
Resolved

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -219,11 +219,15 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
     val crateRoot = path.crateRoot
     if (!path.hasColonColon) {
         if (Namespace.Types in ns) {
-            if (processor("self", containingMod)) return true
-            val superMod = containingMod.`super`
-            if (superMod != null) {
-                if (processor("super", superMod)) return true
+            val context = parent?.contextOfType<RsValueParameter>()
```
matklad: Hm, I think `contextOfType` will overshoot here. What if we are completing something like

```
struct Foo;

fn foo(foo: se/*caret*/lf::Foo) {
}
```

farodin91: What should i do here?

matklad: Hm, it seems to me that perhaps we should fix this at the level of insertion handler?

Reference completion for `self` parameter does not make much sense anyway, because it's more like a named element, and not a reference.

farodin91: I don't have time to look in depth until end of next week.

farodin91: I added both self and self:: to resolve this issue. I have no idea to fix it at the insertion handler.

farodin91:
@matklad any suggestions?

matklad:
Yeah, I still think that it's valid to suggest modules here, because the following is a valid Rust code:

```Rust
fn main() {
    println!("Hello, World");
}

struct S(u32);

impl S {
    fn foo(self::S(x): S) {

    }
}
```

So I still think that the insertion handle should be fixed here: https://github.com/intellij-rust/intellij-rust/blob/cbba4d044c37fe466ed409a871faf33363491e55/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt#L27


Also care must be taken to disable `::` only on top level in self parameter, so that it is autoinserted in, for example, this case:

```
fn foo(xs: Vec<self::S>) { }
```

farodin91:
Resolved but a bit different

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -220,11 +220,13 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
...
+            if (context is RsValueParameter) {
+                if (processor("self", context)) return true
+            } else if (processor("self", containingMod)) return true
+
             val superMod = containingMod.`super`
-            if (superMod != null) {
-                if (processor("super", superMod)) return true
-            }
+            if (superMod != null && processor("super", superMod)) return true
```
matklad: It still seems to me that it is wrong to modify name resolution hear, because semantically resolving a self parameter to itself does not make sense.

I think that modifying insertion handler to handle self parameters simlarly to use globs should be enough:


```Rust
private fun doubleColonLookup(base: LookupElementBuilder): LookupElementBuilder =
    base.withTailText("::")
        .withInsertHandler({ ctx, _ ->
            val offset = ctx.tailOffset - 1
            val inUse = PsiTreeUtil.findElementOfClassAtOffset(ctx.file, offset, RsUseGlobList::class.java, false) != null
            val inSelfParam = PsiTreeUtil.findElementOfClassAtOffset(ctx.file, offset, RsSelfParameter::class.java, false) != null
            if (!(inUse || inSelfParam)) {
                ctx.addSuffix("::")
            }
        })
```

farodin91: For me, it was unclear, how lookup builder work.

farodin91:
Resolved

matklad:
bors r+

Thanks!

PR #1654, opened 2017-08-25
STUB: less AST access

matklad:


matklad:
Ok, this'll be folded as a part of another PR anyways!

Issue #1622, opened 2017-08-13
Can not resolve items from the libc crate


kumbayo:
Items from the libc crate are not resolved properly when using the crate from crates.io.

This is likely related to the libc crate being available twice:
* From crates.io via Cargo.toml
* From rust itself, which can be activated by `#![feature(libc)]` rust-lang/rust#27783

```rust
extern crate libc;

use libc::c_char;

fn main() {
    let c: c_char = 'a' as c_char;
    println!("Hello, world!");
}
```
Cargo.toml:
```toml
[dependencies]
libc="*"
```

![intellij_libc](https://user-images.githubusercontent.com/1041267/29250047-c37b5d8c-803b-11e7-86c1-fe7d29db8758.png)

matklad:
Thanks for the report!

mkaput:
not closed

matklad:
Ah, I always forget that closes xxx should go into the PR description =/

thomcc:
This still happens... (or came back if it was fixed)

PR #1662, opened 2017-08-28
RES: explicit deps hide compiler bulid-ins

matklad:


matklad:
bors r+

closes #1622

PR #1664, opened 2017-08-28
Coerce arrays to slices

kumbayo:
This enables auto completion on arrays.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -171,15 +171,14 @@ class ImplLookup(private val project: Project, private val items: StdKnownItems)
     fun derefTransitively(baseTy: Ty): Set<Ty> {
         val result = mutableSetOf<Ty>()
 
-        var ty = baseTy
-        while (true) {
+        var ty: Ty? = baseTy
+        while (ty != null) {
             if (ty in result) break
             result += ty
-            ty = if (ty is TyReference) {
-                ty.referenced
-            } else {
-                findDerefTarget(ty)
-                    ?: break
+            ty = when (ty) {
+                is TyReference -> ty.referenced
+                is TyArray -> TySlice(ty.base)
+                else -> findDerefTarget(ty)
             }
         }
 
--- a/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
@@ -5,6 +5,8 @@
 
 package org.rust.lang.core.resolve
 
+import junit.framework.AssertionFailedError
+
 class RsTypeAwareResolveTest : RsResolveTestBase() {
     fun testSelfMethodCallExpr() = checkByCode("""
         struct S;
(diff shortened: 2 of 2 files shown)


src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
```diff
@@ -254,6 +254,30 @@ class RsStdlibResolveTest : RsResolveTestBase() {
     """)
     }
 
+    fun `test inherent impl array coercing to slice`() = stubOnlyResolve("""
+    //- main.rs
+        fn main() { [1u8, 2u8, 3u8].len(); }
```
vlad20012: All the tests are depends on a single std method len(), so it is reasonable to move they into `RsGenericExpressionTypeInferenceTest` and implement slice.len() manually

src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
```diff
@@ -254,6 +254,30 @@ class RsStdlibResolveTest : RsResolveTestBase() {
...
+                                  //^ ...slice.rs
+    """)
+
+    fun `test inherent impl slice 1`() = stubOnlyResolve("""
```
vlad20012: This test passes right now without the current fix, so may be removed

src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
```diff
@@ -254,6 +254,30 @@ class RsStdlibResolveTest : RsResolveTestBase() {
...
+            my_slice.len();
+        }          //^ ...slice.rs
+    """)
+
+    fun `test inherent impl slice 2`() = expect<IllegalStateException> {
+        stubOnlyResolve("""
+    //- main.rs
+        fn main() {
+            let my_slice: &[u8] = &[1u8, 2u8, 3u8];
+            <[u8]>::len(my_slice);
```
vlad20012: Is it really passes now? I thought we don't support UFCS

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -177,6 +177,8 @@ class ImplLookup(private val project: Project, private val items: StdKnownItems)
             result += ty
             ty = if (ty is TyReference) {
                 ty.referenced
+            } else if (ty is TyArray) {
```
Undin: Maybe convert these `if`s to one `when` construction?

src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
kumbayo: Is placing it in `RsTypeAwareResolveTest` also OK?
There is already a similar `test slice resolve` there.

vlad20012: All current tests passes without your fix, so they are not enough "similiar" :) So place it next to the exists one.

vlad20012: Hmm, really there are some mirroring between our isolated tests and std-base. But non-std-based unit-like tests should be preferred because they are faster and more clear & visible. but `RsGenericExpressionTypeInferenceTest` is really wrong here, we are speaking about method resolution but not type inference. So I think we should use `RsTypeAwareResolveTest`, there are already some `slice resolve` tests

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
kumbayo: Done

matklad:
ping @vlad20012

vlad20012:
bors r+

Thanks!

matklad:
Test failure looks mysterious, but is probably unrelated: https://travis-ci.org/intellij-rust/intellij-rust/jobs/270737008

bors r=vlad20012

PR #1679, opened 2017-09-01
RES: support associated constants

matklad:
r? @vlad20012


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -216,7 +216,11 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             // then resolving associated type to <Self as Tr>::Item, and then substituting `{Self => T}` into it
 
             for ((element, subst) in lookup.findImplsAndTraits(TyTypeParameter.named(base))) {
```
matklad: Does it make sense to fold this for loop with the previous if, which handles functions?

matklad:
wait, where's my test? o_O

matklad:
Here we go!

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
matklad: Probably yes, we need the same "inherent wins" logic for constants.

PR #1683, opened 2017-09-01
RES: uniform handling of associated items

matklad:
r? @vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -203,22 +203,35 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             val s = base.`super`
             if (s != null && processor("super", s)) return true
         }
-        if (base is RsTraitOrImpl && qualifier.referenceName == "Self") {
-            if (processAll(base.associatedTypesTransitively, processor)) return true
-        }
         if (processItemOrEnumVariantDeclarations(base, ns, processor, isSuperChain(qualifier))) return true
         if (base is RsTypeDeclarationElement && parent !is RsUseItem) {
-            if (processAssociatedFunctionsAndMethodsDeclarations(lookup, base.declaredType, processor)) return true
-        }
-        if (base is RsTypeParameter) {
+            val impls = lookup.findImplsAndTraits(base.declaredType)
+            if (Namespace.Values in ns) {
+                val assocFunctions = impls.flatMap { it.functionsWithInherited }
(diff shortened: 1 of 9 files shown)


vlad20012:
docs again?

matklad:
Yeah, seems like this is a genuinely non-deterministic test :( I'd like to see if it fails deterministically on travis.

matklad:
Hm, nope, that's an actual resolve failure, which was not caught by our resolve test:

```
    fun `test foo`() = checkByCode("""
        pub trait Iter {
            type Item;
                //X

            fn scan<St, B, F>(self, initial_state: St, f: F) -> Scan<Self, St, F>
                where Self: Sized, F: FnMut(&mut St, Self::Item) -> Option<B> { unimplemented!() }
        }                                                 //^
    """)
```

notriddle:
Protip: if you're trying to find out if a test is deterministic or not, you probably want to use [bors try](https://bors.tech/documentation/reference/). That way, you don't accidentally merge flaky code.

matklad:
@notriddle actually I was hoping that the build passes by accident, because obviously my 100% correct changes could not have caused this unrelated build failure. I am so glad that bors forbids me to make such silly and unprofessional mistakes, which are nether the less very tempting after finishing a chunk of code :)

matklad:
Spotted one more bit of logical duplication: `processFnsWithInherentPriority` and `associatedTypesTransitively` are doing a similar logic. I think what we probably need here is 

`RsTraitOrImpl.logicalMembers`, which for traits would be just direct memebers, and for impls direct members + members from the trait, which are not implemented in the impl directly. "transitiveness" (traits from bounds) should probably be handled separately, via `findImplsAndTraits`.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -435,27 +419,56 @@ private fun processFieldDeclarations(struct: RsFieldsOwner, subst: Substitution,
...
+    }
+    return lookup.derefTransitively(receiver).any { processAssociatedItems(lookup, it, VALUES, methodProcessor) }
+}
+
+private fun processAssociatedItems(lookup: ImplLookup, type: Ty, ns: Set<Namespace>, processor: RsResolveProcessor): Boolean {
+    val visitedInherent = mutableSetOf<String>()
+    fun processTraitOrImpl(traitOrImpl: BoundElement<RsTraitOrImpl>, inherent: Boolean): Boolean {
+        fun inherentProcessor(entry: BoundElement<RsNamedElement>): Boolean {
+            val name = entry.element.name ?: return false
+            if (inherent) visitedInherent.add(name)
```
vlad20012: Why name but not the element? The same name may be used in different scopes

PR #1686, opened 2017-09-01
Fix exceptions happening after running the "Change try! to ?" inspection

### LINKED ISSUE #1685: Running "Change try! to ?" causes an exception
When running "Change try! to ?" on the following code an exception happens.

```rust
fn foo() -> Result<(), ()> {
    try!(Err(()));
    Ok(())
}
```

```returns_ok()?
 parent=let f = returns_ok()?;
java.lang.Throwable
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:132)
	at com.intellij.psi.impl.PsiElementBase.notNullChild(PsiElementBase.java:285)
	at org.rust.lang.core.psi.impl.RsMacroExprImpl.getMacroCall(RsMacroExprImpl.java:31)
	at org.rust.lang.core.types.infer.RsFnInferenceContext.inferMacroExprType(TypeInference.kt:432)
	at org.rust.lang.core.types.infer.RsFnInferenceContext.inferType(TypeInference.kt:124)
	at org.rust.lang.core.types.infer.RsFnInferenceContext.processStatement(TypeInference.kt:92)
	at org.rust.lang.core.types.infer.RsFnInferenceContext.inferType(TypeInference.kt:82)
	at org.rust.lang.core.types.infer.RsFnInferenceContext.inferType$default(TypeInference.kt:80)
	at org.rust.lang.core.types.infer.RsFnInferenceContext.inferBlockType(TypeInference.kt:78)
	at org.rust.lang.core.types.infer.RsInferenceContext.infer(TypeInference.kt:35)
	at org.rust.lang.core.types.infer.TypeInferenceKt.inferTypesIn(TypeInference.kt:23)
```

In the Psi the above code is represented the following way:
RsMacroExpr
-RsMacroCall
--identifier
--!
--RsTryMacroArgument

RsTryMacroInspection registers the fix on the RsMacroCall.
When the fix is applied, it replaces the RsMacroCall node with a RsTryExpr, leaving back a corrupt RsMacroExpr missing a required RsMacroCall child.
Later some other code trips over this corrupt psi node.

In the tests we do not notice this problem as we probably do not look at the psi after the fix.

kumbayo:
Fixes #1685

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsTryMacroInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsTryMacroInspection.kt
@@ -9,10 +9,7 @@ import com.intellij.codeInspection.LocalQuickFix
 import com.intellij.codeInspection.ProblemDescriptor
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
-import org.rust.lang.core.psi.RsMacroCall
-import org.rust.lang.core.psi.RsPsiFactory
-import org.rust.lang.core.psi.RsTryExpr
-import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.macroName
 
 /**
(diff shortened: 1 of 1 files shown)


kumbayo:
I am not sure on how to test that we don't corrupt the PSI anymore.

matklad:
Thanks! I'll ask colleagues about asserting PSI correction in tests (though, I am not sure there's a super great answer here).

bors r+

PR #1687, opened 2017-09-01
Fix the return type of panic!() and unreachable!()

kumbayo:
panic!(), unimplemented!() and unreachable!() unconditionally panic
and thus have a return type of "!".

All 3 support a format!() syntax so support that as well.


matklad:
bors r+

Thanks!

kumbayo:
@matklad It seems bors is stuck :-(

Is this related to bors-ng/bors-ng#206?
(I can't access the link to intellij-rust History posted there)

matklad:
Hm, it's not even in the logs now...


bors r+

Issue #1667, opened 2017-08-29
False positive: Cannot borrow immutable local variable as mutable


knsd:
```rust
let mut xs: Vec<Vec<usize>> = vec![vec![1, 2], vec![3, 4]];

for elt in &mut xs {
    elt.push(0)
//  ~~~ Error here: Cannot borrow immutable local variable `elt` as mutable
}
```

alygin:
Looks like the same problem as in #1559

farodin91:
@vlad20012 The main problem is type inference fails.
```rust
fn main() {
  let mut b: Vec<Vec<usize>> = vec![vec![1, 2]];
  for a in &mut b {
      a
    //^ &mut Vec<usize>
  }
}```

vlad20012:
Okaaay, everything closes on the type inference.

farodin91:
I will open a PR with a bit clean up code

PR #1691, opened 2017-09-04
INSP: Refactor borrow checker

farodin91:


vlad20012:
It is not just refactoring. It make the inspection less strict to solve #1667, isn't it? But I think we should just make type inference better... (don't sure it is the right vision, may be we should fix it as soon as possible and then deal with type inference)

farodin91:
I try to fix it directly, but i want to support it for all reference mutable iterators

vlad20012:
Real problem is not in type inference directly, it is about exact/fuzzy impl matching (look at `RsPreciseTraitMatchingTest`). If you want to fix it, I have some suggestions. (And as I see, @kumbayo is looking in the same direction with #1697))

farodin91:
What do suggestions do you think of? I want in this PR only fix issue directly correlate to the borrow checker.

src/test/kotlin/org/rust/ide/inspections/RsBorrowCheckerInspectionTest.kt
```diff
@@ -125,6 +125,16 @@ class RsBorrowCheckerInspectionTest : RsInspectionsTestBase(RsBorrowCheckerInspe
         }
     """)
 
+    fun `test no highlight for mutable for loops`() = checkByText("""
+        fn test() {
+            let mut xs: Vec<Vec<usize>> = vec![vec![1, 2], vec![3, 4]];
+
+            for test in &mut xs {
+                <error>test</error>.push(0)
```
matklad: Hm, I think there shouldn't be an error here? At lest the following compiles fine:

```Rust
fn main() {
    let mut xs: Vec<Vec<usize>> = vec![vec![1, 2], vec![3, 4]];
    for test in &mut xs {
        test.push(0)
    }
}
```

matklad: If this is a test for the current behavior which should be fixed later, could you add a `TODO: this should not show error here` ?

farodin91: I add a note

vlad20012:
Hmm, I think this test confused me. So it is really just a refactoring. So thanks!
bors r+

PR #1697, opened 2017-09-05
Primitive type resolving improvements

### LINKED ISSUE #1667: False positive: Cannot borrow immutable local variable as mutable

kumbayo:
- Distinguish between *const and *mut pointers to help resolving to only 1 candidate
- Distinguish between & and &mut references to help resolving to only 1 candidate
- Resolve inherent implementations on primitives for "primitive::method()" syntax
CC #371
Fixes #1667

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -198,6 +198,11 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
     }
 
     if (qualifier != null) {
+        val primitiveType = TyPrimitive.fromPath(qualifier);
+        if (primitiveType != null) {
+            if (processAssociatedItems(lookup, primitiveType, ns, processor)) return true
+        }
+
         val base = qualifier.reference.resolve() ?: return false
         if (base is RsMod) {
             val s = base.`super`
--- a/src/main/kotlin/org/rust/lang/core/types/ty/TyPointer.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/TyPointer.kt
@@ -10,7 +10,10 @@ import org.rust.lang.core.resolve.ImplLookup
 data class TyPointer(val referenced: Ty, val mutability: Mutability) : Ty {
 
     override fun unifyWith(other: Ty, lookup: ImplLookup): UnifyResult =
-        if (other is TyPointer) referenced.unifyWith(other.referenced, lookup) else UnifyResult.fail
+        if (other is TyPointer && mutability == other.mutability)
+            referenced.unifyWith(other.referenced, lookup)
+        else
+            UnifyResult.fail
 
     override fun substitute(subst: Substitution): Ty =
         TyPointer(referenced.substitute(subst), mutability)
--- a/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
@@ -558,4 +558,34 @@ class RsTypeAwareResolveTest : RsResolveTestBase() {
                //^
         }
     """)
+
+    fun `test impl trait for mutable reference`() = expect<AssertionFailedError> {
+        checkByCode("""
+        struct Foo;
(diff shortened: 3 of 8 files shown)


src/main/kotlin/org/rust/lang/core/types/ty/TyPointer.kt
```diff
@@ -10,7 +10,10 @@ import org.rust.lang.core.resolve.ImplLookup
 data class TyPointer(val referenced: Ty, val mutability: Mutability) : Ty {
 
     override fun unifyWith(other: Ty, lookup: ImplLookup): UnifyResult =
-        if (other is TyPointer) referenced.unifyWith(other.referenced, lookup) else UnifyResult.fail
+        if (other is TyPointer && mutability == other.mutability)
```
kumbayo: This fails unification when the mutability does not match.
I am not fully sure about this.
I tried resolving fuzzy in this case but I still get no good resolving because of "Multiple implementations"

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -198,6 +198,11 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
     }
 
     if (qualifier != null) {
+        val primitiveType = TyPrimitive.fromPath(qualifier);
+        if (primitiveType != null) {
+            if (processAssociatedFunctionsAndMethodsDeclarations(lookup, primitiveType, processor)) return true
```
kumbayo: Not sure if processAssociatedFunctionsAndMethodsDeclarations is the right thing to call here but it seems to work nicely.

matklad: Turns out that processing associated items is a mess right now, see #1683. I hope I'll fix it today.

src/main/kotlin/org/rust/lang/core/types/ty/TyPointer.kt
kumbayo: I am currently not sure how to use the exact/fuzzy match with resolving but I will look into what i can figure out.

vlad20012: Hmm, I was mistaken here. Really this code does not compile:
```rust
struct S;
impl<'a> Tr for &'a S {}
trait Tr { fn foo(&self) {} }

fn main() {
    let a = S;
    let b = &mut a;
    b.foo();
}
```
But it will compile if we replace `&mut a` to just `&a`

vlad20012: And try the same with TyReference. It may fix #1667

kumbayo: I am not sure I understand what this means. Is the TyPointer unification change OK as it is now?

Yes adapting TyReference the same ways as TyPointer fixes the linked issue :-)
In the past I have tried the same code in TyReference but could not find a case where it mattered so i did not include it in my patch.

I will rebase for the associated type bugfix and push tomorrow when I have more time.

src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
```diff
@@ -558,4 +558,30 @@ class RsTypeAwareResolveTest : RsResolveTestBase() {
                //^
         }
     """)
+
+    fun `test impl trait for mutable reference`() = checkByCode("""
```
vlad20012: This tests really should be moved to  `RsPreciseTraitMatchingTest`

vlad20012:
bors r+
Thanks! And sorry for the long response, I just missed the notification. For the future, you can ping me or matklad by mentioning in comments here (or also on gitter), if we don't answer for a long time.

src/main/kotlin/org/rust/lang/core/types/ty/TyPointer.kt
vlad20012: You understood correctly, it is OK =)

PR #1699, opened 2017-09-06
PSI: Cache file items

matklad:
Not sure if the implementation can be less copy-pasty...

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
@@ -13,15 +13,16 @@ import com.intellij.psi.FileViewProvider
 import com.intellij.psi.PsiDirectory
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
-import com.intellij.psi.tree.IElementType
 import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
 import com.intellij.psi.util.PsiModificationTracker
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -70,50 +71,134 @@ class RsFile(
...
+
+    @Volatile
+    private var _items_cache: ItemsCache? = null
+
+    override fun subtreeChanged() {
+        super.subtreeChanged()
+        _items_cache = null
     }
+
+    private val items_cache: ItemsCache
```
vlad20012: `_`

matklad: 😑

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -70,50 +71,134 @@ class RsFile(
...
+
+    @Volatile
+    private var _itemsCache: ItemsCache? = null
+
+    override fun subtreeChanged() {
+        super.subtreeChanged()
+        _itemsCache = null
     }
+
+    private val itemsCache: ItemsCache
```
vlad20012: May be it's better to use `by lazy(PUBLICATION)` instead?

matklad: It won't work unfortunately because we need to clear cache manually in `subtreeChanged`.

PR #1705, opened 2017-09-07
COMP: Refactor macro comp and walkup

### LINKED ISSUE #1468: Reduce code duplication for completing/resolving macros
There seems to be a bit of code duplication between `RsMacroDefinitionCompletionProvider` and `NameResolution.kt`: functions like `processMacroReferenceVariants` and `walkUp` seems to be duplicated. Would be nice to somehow get rid of this duplication!

farodin91:
Fixes #1468

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -322,26 +322,7 @@ fun processMetaItemResolveVariants(element: RsMetaItem, processor: RsResolveProc
 }
 
 fun processMacroCallVariants(invocation: PsiElement, processor: RsResolveProcessor): Boolean =
-    macroWalkUp(invocation) { scope ->
-        processMacroDeclarations(scope, processor)
-    }
-
-private fun macroWalkUp(
-    start: PsiElement,
-    processor: (scope: RsCompositeElement) -> Boolean
-): Boolean {
-    var scope = start.context as? RsCompositeElement
-    while (scope != null) {
-        if (processor(scope)) return true
-        val cameFrom = scope
-        scope = scope.context as? RsCompositeElement
-        if (scope == null && cameFrom is RsFile) {
-            scope = cameFrom.`super`
-        }
-    }
-
-    return false
-}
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -779,21 +767,20 @@ private fun processNestedScopesUpwards(scopeStart: RsCompositeElement, processor
...
 }
 
-
 // There's already similar functions in TreeUtils, should use it
 private fun walkUp(
-    start: RsCompositeElement,
+    start: PsiElement,
     stopAfter: (RsCompositeElement) -> Boolean,
-    processor: (cameFrom: RsCompositeElement, scope: RsCompositeElement) -> Boolean
+    context: (cameFrom: PsiElement, scope: RsCompositeElement) -> RsCompositeElement?,
```
matklad: Hm, maybe we can add a `stopAtFileBoundary: Boolean = false` flag instead of making this fully configurable via a closure?

farodin91:
I changed this

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -779,23 +760,26 @@ private fun processNestedScopesUpwards(scopeStart: RsCompositeElement, processor
...
-    var cameFrom: RsCompositeElement = start
+    var cameFrom = start
     var scope = start.context as RsCompositeElement?
     while (scope != null) {
         if (processor(cameFrom, scope)) return true
         if (stopAfter(scope)) break
         cameFrom = scope
-        scope = scope.context as RsCompositeElement?
+        scope = scope.context as? RsCompositeElement
+        if (stopAtFileBoundary) {
```
matklad: Looks like the meaning of this variable is inverted? That is, it says "stop at file boundary", but, when `true`, it actually causes the search to continue beyond the file boundary.

farodin91: Ups

matklad:
bors r+

Thanks!

Issue #1559, opened 2017-08-02
mut check fail


biolee:
https://github.com/pingcap/tikv/blob/master/src/coprocessor/endpoint.rs#L1017

func is mut, but it do not get underscore and get an err for call update who need &mut self

![screenshot](https://user-images.githubusercontent.com/7426917/28854905-376dc392-76ff-11e7-9e16-7119c9c6e720.png)

kumbayo:
This error is not visible anymore since 61eeac6d6b99ae0bb37b43530f49afac960eae64.
In this commit the inferred type of `func` changed from
`&Box<AggrFunc>` to `<B as Iterator>::Item` which does not allow to resolve `.update(...)` and thus hides the error.

Minimized example:
```rust
struct Foo;
fn main() {
    let a: &[Foo] = &[];
    let b: &Vec<Foo> = &vec![];
    for (it_a, it_b) in a.iter().zip(b) {
        it_b;
        //^Should be &Foo but is <B as Iterator>::Item
    }
}
```

vlad20012:
I think I'll fix it on the next iteration of type inference system.

PR #1711, opened 2017-09-09
TY: advanced type inference & unification

vlad20012:
Really it's just a part of The Ultimate Type Inference™. I just found something like an "incremental point", so await a new PRs soon.
Good news is that a lot of really advanced type inference are working now:

```rust
fn main() {
    let a = 0;
          //^ u8
    let mut vec = Vec::new();
    vec.push(a);
    vec.push(1u8);
}
```
```rust
fn main() {
    let vec = vec![1, 2, 3];
    let b: Vec<_> = vec.into_iter().collect();
    b
  //^ Vec<i32>
}
```

Bad news is that the code is now like an awful Frankenstein molded out of our old type inference (with substitutions, Ty.unifyWith and so on) and the new one. I think we can live with it for a bit :)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Fulfillment.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Fulfillment.kt
@@ -1,0 +1,184 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types.infer
+
+import org.rust.lang.core.psi.RsTraitItem
+import org.rust.lang.core.psi.RsTypeAlias
+import org.rust.lang.core.resolve.ImplLookup
+import org.rust.lang.core.types.TraitRef
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyInfer
+import org.rust.lang.core.types.ty.TyTypeParameter
+
+sealed class Predicate: TypeFoldable<Predicate> {
+    /** where T : Bar<A,B,C> */
+    data class Trait(val trait: TraitRef) : Predicate() {
+        override fun superFoldWith(folder: TypeFolder): Trait =
+            Trait(trait.foldWith(folder))
+    }
+
+    /** where <T as TraitRef>::Name == X */
+    data class Projection(
+        val selfTy: Ty,
+        val trait: RsTraitItem,
+        val target: RsTypeAlias,
+        val ty: Ty
+    ): Predicate() {
+        override fun superFoldWith(folder: TypeFolder): Projection =
+            Projection(selfTy.foldWith(folder), trait, target, ty.foldWith(folder))
+
+        override fun toString(): String =
+            "<$selfTy as ${trait.name}>::${target.name} == $ty"
+    }
+
+    /** where `T1 == T2` */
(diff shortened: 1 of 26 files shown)


src/main/kotlin/org/rust/lang/core/types/ty/Fold.kt
```diff
@@ -0,0 +1,35 @@
...
+            if (ty is TyTypeParameter) folder(ty) else ty.superFoldWith(this)
+    }
+}
+
+fun substFolder(subst: Substitution): TypeFolder {
+    return object : TypeFolder {
+        override fun invoke(ty: Ty): Ty =
+            subst[ty] ?: ty.superFoldWith(this)
+    }
+}
```
matklad: This file badly needs some comments with high-level overview :)

matklad: Also, let's perhaps leave only types themselves in the `ty` package?

src/main/kotlin/org/rust/lang/core/types/infer/Fulfillment.kt
```diff
@@ -0,0 +1,175 @@
...
+data class Obligation(val recursionDepth: Int, var predicate: Predicate): TypeFoldable<Obligation> {
+    constructor(predicate: Predicate) : this(0, predicate)
+
+    override fun superFoldWith(folder: TypeFolder): Obligation =
+        copy(predicate = predicate.foldWith(folder))
+}
+
+data class PendingPredicateObligation(val obligation: Obligation, val stalledOn: MutableList<Ty>)
+
+class ObligationForest {
```
matklad: Yeah, I kinda expected this one...

matklad:
Looks just awesome @vlad20012 ! Now I understand what you've been doing all this time =)

The two big questions here are performance and documentation. For docs, I think we want short docstrings at the core types, and a couple of paragraphs in ARCHITECTURE.md

matklad:
Perf seems OK! Waiting for the docs to do a thorough review ;)

I am so pumped about this! This an awesome improvement!

matklad:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

sirgl:
Is it enough to close type system(to react with error on unknown type)? If there is no macroses of course.

vlad20012:
@sirgl No, type inference system still not complete. We failing on associated types [sometimes](https://github.com/intellij-rust/intellij-rust/issues/1559#issuecomment-328363829) and even don't support some syntax (`I: Iterator<Item=T>`). But some minimal type checking is possible now.
What about unknown types, it is mostly the problem of macros or the name resolution at all. Currently it's hard to get unknown type if all names successfully resolved.
If you worry about auto import, it is stalled not on type inference, but on macros.

vlad20012:
Mmm, there was some problems with CI. Anyway, this PR just merged!

supercurio:
Maybe what introduced #1753; but great direction for this plugin nonetheless!

vlad20012:
@supercurio it is in release already. #1753 is just a bug! I working on it!

PR #1713, opened 2017-09-09
Fix URL of the Changelog

kumbayo:
Commit eecdf544ae002 updated the date but forgot to increment the number.


matklad:
bors r+

Thanks!

PR #1714, opened 2017-09-10
ANN: Fix wrong mutability annotation on empty structs without braces

kumbayo:
A mutable reference to a empty struct without braces would produce an error
"Cannot borrow immutable local variable `EmptyStruct` as mutable"

struct EmptyStruct;
let error = &mut EmptyStruct;
let ok = &mut EmptyStruct { };

The solution for this is very similar to 6842c263b91feeb99c78fb7319d8757203d47ebf for #1114

An empty struct expression with and without braces have a different
PSI representation.
EmptyStruct is a RsPathExpr which resolves to a RsStructItem
EmptyStruct {} is a RsStructLiteral which is handled by the else in RsExpr.isMutable


vlad20012:
Yep, looks ok. Thanks!
bors r+

PR #1724, opened 2017-09-13
TYPE: Extend Statement Mover and refactor some tests

### LINKED ISSUE #1575: Refactor AbstractUpDownMover and support more types
Todo
- [ ] impl
- [ ] mod
- [ ] trait
- [ ] match pattern
- [ ] macro_definition
- [ ] macro_call
- [ ] more code reuse in test
- [ ] clean up code in `checkAvailable`

farodin91:
Fixes #1575 
@matklad  r?


matklad:
`RsAbstractUpDownMover` has a single descendant `RsBasicUpDownMover`. Perhaps we could collapse this two classes to avoid unnecessary inheritance?

src/main/kotlin/org/rust/ide/actions/mover/RsAbstractUpDownMover.kt
```diff
@@ -76,19 +63,62 @@ abstract class RsAbstractUpDownMover : StatementUpDownMover() {
             }
             element = parent
         }
+        return element
+    }
 
-        if (element != null) {
-            info.toMove2 = LineRange(element)
+    private fun checkContainerBoundaries(element: PsiElement?, collectedElement: PsiElement): Boolean {
```
matklad: Looks like this could be `element: PsiElement?`

In general, functions which take a nullable parameter and do nothing if the parameter is null are code smell: it's usually better to push null-checks to the caller (and to avoid null checks at all if possible, by threading non-null types).

src/main/kotlin/org/rust/ide/actions/mover/RsAbstractUpDownMover.kt
```diff
@@ -76,19 +63,62 @@ abstract class RsAbstractUpDownMover : StatementUpDownMover() {
...
+        if (containers.any { element?.elementType == it }) {
+            var parentOfFn = collectedElement.parent
+            while (parentOfFn !is RsFile) {
```
matklad: This could use `.ancestors` function perhaps?

src/main/resources/META-INF/plugin.xml
```diff
@@ -577,8 +577,13 @@
...
         <moveLeftRightHandler language="Rust" implementationClass="org.rust.ide.actions.RsMoveLeftRightHandler"/>
-        <statementUpDownMover implementation="org.rust.ide.actions.RsFunctionUpDownMover"/>
-        <statementUpDownMover implementation="org.rust.ide.actions.RsStructItemUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsFunctionUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsStructItemUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsImplItemUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsTraitItemUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsModItemUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsMacroCallUpDownMover"/>
+        <statementUpDownMover implementation="org.rust.ide.actions.mover.RsMacroDefinitionUpDownMover"/>
```
matklad: It seems to me that all these movers work similarly: they move items inside a module. So perhaps we could create only a single mover instead, which will handle all items?

farodin91: Yes. I need to find a nice solution to support easy extension for match pattern usw...

farodin91:
@matklad I merged all class

farodin91:
@matklad Anything to do?

matklad:
@farodin91, no, this looks great to me now, thanks!

@bots r+

PR #1725, opened 2017-09-13
Fix unsound `findWithCache`

matklad:
`getCachedValue` relies on identity of passed-in lambdas, and Kotlin's
inline does not help us to maintain these identities. So let's create
keys ourselves instead.

r? @vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/ProjectCache.kt
+++ b/src/main/kotlin/org/rust/lang/utils/ProjectCache.kt
@@ -1,0 +1,45 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.utils
+
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Key
+import com.intellij.psi.util.CachedValue
+import com.intellij.psi.util.CachedValueProvider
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/utils/ProjectCache.kt
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    fun getOrPut(project: Project, key: T, defaultValue: () -> R): R {
+        val cache = CachedValuesManager.getManager(project)
+            .getCachedValue(project, cacheKey, provider, false)
+        return cache.getOrPut(key) { defaultValue() }
+    }
+
+    private val cacheKey: Key<CachedValue<ConcurrentMap<T, R>>> = Key.create(cacheName)
```
vlad20012: Should we put fields before methods?

matklad: Yeah, style guide says:

> Do not sort the method declarations alphabetically or by visibility

https://github.com/JetBrains/kotlin-web-site/blob/38a85989b11977559fad17b60001cd96736d9470/pages/docs/reference/coding-conventions.md#class-layout

will rebase, 👍

matklad: Addressed!

PR #1727, opened 2017-09-13
Experimantal basic type checking

vlad20012:
It is very experimental and very basic (=D). It has a lot of false negatives, but should not contain false positives. Anyway, false positive is a bug. Report it.

image


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -281,19 +296,46 @@ private class RsFnInferenceContext(
...
+                TyTraitObject::class.java
+            )
+
+            if (!expected.containsTyOfClass(ignoredTys) && !inferred.containsTyOfClass(ignoredTys)) {
+                // another awful hack: check that inner expressions did not annotated as an error
+                // to disallow annotation intersections. This should be done in a different way
+                fun PsiElement.isChildOf(psi: PsiElement) = this.ancestors.contains(psi)
+                if (ctx.diagnostics.all { !it.element.isChildOf(expr) }) {
+                    ctx.reportTypeMismatch(expr, expected, inferred)
+                }
```
matklad: This is probably OK for now, but it's definitely superlinear in complexity. 

I think a better architecture here would be to collect **all** diagnostics at first and the filter out some of them at the very end.

PR #1733, opened 2017-09-13
Fixed kotlin compiler warnings in type inference

vlad20012:


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -933,11 +935,11 @@ private fun Sequence<Ty>.infiniteWithTyUnknown(): Sequence<Ty> =
     this + generateSequence { TyUnknown }
 
 private fun unwrapParenExprs(expr: RsExpr): RsExpr {
```
matklad:
```
tailrec private fun unwrapParenExprs(expr: RsExpr): RsExpr =
    if (expr is RsParenExpr) unwrapParenExprs(expr.expr) else expr
```

;)

PR #1736, opened 2017-09-16
Fixed "unresolved reference" inspection

vlad20012:
Was broken with bf9dad5300081f4d3b979cb2570594ef87b4e41b


vlad20012:
it would be nice if someone wrote tests for it =)

Issue #1696, opened 2017-09-05
Type information on declaration of var


Zteve:
Getting type information on the use of a variable is fine and dandy (although too generic sometimes) but often the type is implied in the declaration, and so it is not clear what this is. This most often happens when `|x| expression` closures are used. It is often hard to deduce what type `x` is, even though the IDE knows it (it can tell you if you refer to `x` in the body of the closure).

Just saying that it is defined/declared/bound in a pattern, or in a function signature, or in a block somewhere is not very helpful, since I can usually see that.  Give me the type information here as well.

PR #1738, opened 2017-09-17
Type information on var binding #1696

vlad20012:
Is it legal if method named `getExpressionsAt` now return not only expressions?)


src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
```diff
@@ -8,23 +8,24 @@ package org.rust.ide.hints
...
+        pivot.ancestors
             .takeWhile { it !is RsItemElement }
-            .filter(RsExpr::class.java)
+            .filter { it is RsExpr || it is RsPatBinding }
             .toList()
 
-    override fun getInformationHint(element: RsExpr): String =
-        StringUtil.escapeXml(element.type.toString())
+    override fun getInformationHint(element: PsiElement): String =
+        StringUtil.escapeXml(((element as? RsExpr)?.type ?: (element as RsPatBinding).type).toString())
```
matklad: Let's extract `val type = when (element)` bit to keep `toString` logic separate?

PR #1741, opened 2017-09-18
TY: Correct infer types of byte literals `b'a'` and `b"abc"`

### LINKED ISSUE #1739: Error for byte literal
```rust
fn main() {
	let _: u8 = b'a';
}
```

This code isn't an error when using the command line compiler, but does cause an error inside the editor:
```
Mismatched types E0308
expected u8, found char
```

There is no error for byte string literals: `let _: &[u8] = b"abc";` works.

vlad20012:
Fixes #1739

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsLiteralKind.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsLiteralKind.kt
@@ -57,7 +57,7 @@ sealed class RsLiteralKind(val node: ASTNode) {
         override val offsets: LiteralOffsets by lazy { offsetsForNumber(node) }
     }
 
-    class String(node: ASTNode) : RsLiteralKind(node), RsLiteralWithSuffix, RsTextLiteral {
+    class String(node: ASTNode, val isByte: kotlin.Boolean) : RsLiteralKind(node), RsLiteralWithSuffix, RsTextLiteral {
         override val offsets: LiteralOffsets by lazy { offsetsForText(node) }
 
         override val validSuffixes: List<kotlin.String> get() = emptyList()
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsLiteralKind.kt
```diff
@@ -75,6 +75,24 @@ sealed class RsLiteralKind(val node: ASTNode) {
         }
     }
 
+    class ByteString(node: ASTNode) : RsLiteralKind(node), RsLiteralWithSuffix, RsTextLiteral {
```
vlad20012: This is a pure copy-paste of `String` class. It seems, we should extract superclass, but with sealed class it is... strange, I guess

matklad: Can we have a single `String` class with `isByte: bool` field?

vlad20012: Hmm, sounds reasonable

vlad20012:
Done

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -343,10 +343,17 @@ private class RsFnInferenceContext(
     }
 
     fun inferLitExprType(expr: RsLitExpr, expected: Ty?): Ty {
```
matklad: Why do we need these two similar functions? :)

vlad20012: Oh, it's another story =D

First works in function inference context, second - in out-of-function context (constanst/statics)

matklad: Could we unify them?

vlad20012: Really we currently very badly support out-of-fn expressions, and it requires detailed consideration. I'll open an issue

matklad: OK :) In general, I believe such low-hanging refactoring fruit should be picked up as soon as possible, otherwise it tends to turn into wine...

PR #1747, opened 2017-09-19
Consider type parameters in paths when resolving associated items

kumbayo:
This helps in the case of
```
let v = Vec::<u16>::new();
```
which is now correctly typified as Vec<u16>

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -20,6 +20,7 @@ import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ref.RsReference
 import org.rust.lang.core.stubs.index.RsNamedElementIndex
 import org.rust.lang.core.types.BoundElement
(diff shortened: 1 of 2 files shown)


kumbayo:
@vlad20012 Do you think this is an OK way to support type parameters in paths until the ultimate type inference?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -203,14 +204,14 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             if (processAssociatedItems(lookup, primitiveType, ns, processor)) return true
         }
 
-        val base = qualifier.reference.resolve() ?: return false
-        if (base is RsMod) {
-            val s = base.`super`
+        val base = qualifier.reference.advancedResolve() ?: return false
```
vlad20012: Here we can use destructuring declaration `val (base, subst) = ...`

vlad20012:
Yep, seems OK. Thanks!
bors r+

PR #1749, opened 2017-09-20
RES: Support impl for type parameter `impl<T> Trait for T`

### LINKED ISSUE #1735: Support impl for type parameter `impl<T> Trait for T`
Currently we don't look for impls like `impl<T> Trait for T`, so this test fails
```rust
trait Trait {
    fn foo(&self) {}
}    //X
impl<T> Trait for T {}
struct X;
fn main() {
    X.foo();
}   //^
```
It is pretty easy to implement. [Here](https://github.com/intellij-rust/intellij-rust/blob/9603f049d96b1ec515830b1db03737db14e3af1a/src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt#L48) we indexing the `impl` by the type. We should check if type is a type parameter of this impl (go to parent psi, find type parameters declaration and check the type name is the same), then if it is a type parameter, save it to index with a specific key that is invalid identifier (like #). Then while finding impls retreive index twice: with actual type name and with our specific key.

---
You can find more info about how to start contributing to Intellij-Rust in [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md), and some info about our architecure in [ARCHITECTURE.md](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md). If you have any questions about this issue, you can ask them right here or in our [gitter](https://gitter.im/intellij-rust/intellij-rust). If you decided to grab this issue, please leave a comment here and put yourself in the [worklist](https://github.com/intellij-rust/intellij-rust/wiki/Worklist).

Undin:
Closes #1735

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
@@ -11,7 +11,12 @@ import com.intellij.psi.stubs.AbstractStubIndex
 import com.intellij.psi.stubs.IndexSink
 import com.intellij.psi.stubs.StubIndexKey
 import com.intellij.util.io.KeyDescriptor
+import org.rust.lang.core.psi.RsBaseType
 import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.ext.bounds
+import org.rust.lang.core.psi.ext.name
+import org.rust.lang.core.psi.ext.typeElement
+import org.rust.lang.core.psi.ext.typeParameters
 import org.rust.lang.core.stubs.RsFileStub
 import org.rust.lang.core.stubs.RsImplItemStub
 import org.rust.lang.core.types.TyFingerprint
@@ -33,7 +38,7 @@ class RsImplIndex : AbstractStubIndex<TyFingerprint, RsImplItem>() {
             val fingerprint = TyFingerprint.create(target)
                 ?: return emptyList()
 
-            return getElements(KEY, fingerprint, project, GlobalSearchScope.allScope(project))
+            val impls = getElements(KEY, fingerprint, project, GlobalSearchScope.allScope(project))
                 .filter { impl ->
                     val ty = impl.typeReference?.type
                     // Addition class check is a temporal solution to filter impls for type parameter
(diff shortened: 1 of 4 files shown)


Undin:
This implementation checks if type parameter is unbounded because otherwise it leads to stack overflow. Solution of this problem may require a lot of changes in code and I think it should be fixed while supporting trait bounds in separate issue.

src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
```diff
@@ -33,19 +38,36 @@ class RsImplIndex : AbstractStubIndex<TyFingerprint, RsImplItem>() {
...
+            val impl = stub.psi
+            val typeRef = impl.typeReference ?: return
+            val type = typeRef.typeElement
+
+            val key = if (type is RsBaseType) {
+                val typeParam = impl.typeParameters.find { it.name == type.name }
+                // At this moment we support only free trait impls without bounds
+                // to avoid stack overflow while resolving
+                if (typeParam != null && typeParam.bounds.isEmpty()) {
+                    TyFingerprint.TYPE_PARAMETER_FINGERPRINT
```
vlad20012: If type params is not empty we should just ignore the impl (exclude from index)
Also, what about whew clauses?

vlad20012:
Ok, I understood why SO here. So... this will be fixed when I finish replacing `Ty.unifyWith` with `FulfillmentContext`

src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
Undin: `RsTypeParameter.bounds` is extension property (see [here](https://github.com/intellij-rust/intellij-rust/blob/4c7c1e331e49b84aa262fcc7471cef72605a3920/src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeParameter.kt#L17-L25])) which collect all bounds including where clause

vlad20012: Oh, OK

vlad20012:
bors r+
Many thanks!

vlad20012:
Oh, we forgot to bump stub/index version...

Undin:
@vlad20012 Let's fix it!

Undin:
Not very successful PR :(

PR #1764, opened 2017-09-22
TY: Coerce container type while index expr type inference

### LINKED ISSUE #1748: type inference doesn't work through array indexes on references to Vecs.
'm' has an unknown type in the following example:
```rust
fn foo() -> u32 {
    let r: Vec<u32> = Vec::new();
    let p = &r;
    let m = p[0];
    m
}
```

Undin:
Fixes #1748

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -163,7 +163,7 @@ class ImplLookup(private val project: Project, private val items: StdKnownItems)
         return subst + associated
     }
 
-    fun derefSequence(baseTy: Ty): Sequence<Ty> {
+    fun coercionSequence(baseTy: Ty): Sequence<Ty> {
         val result = mutableSetOf<Ty>()
         return generateSequence(baseTy) {
             if (result.add(it)) {
@@ -200,7 +200,7 @@ class ImplLookup(private val project: Project, private val items: StdKnownItems)
         return rawType.substitute(impl.subst)
     }
 
-    fun findIndexOutputType(containerType: Ty, indexType: Ty): Ty {
+    fun findIndexOutputType(containerType: Ty, indexType: Ty): Ty? {
         val impls = findImplsAndTraits(containerType)
             .filter { it.element.implementedTrait?.element?.isIndex ?: false }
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/lang/core/type/RsIndexTypeInferenceTest.kt
```diff
@@ -129,4 +129,61 @@ class RsIndexTypeInferenceTest : RsTypificationTestBase() {
...
+        pub trait Deref {
+            type Target: ?Sized;
+            fn deref(&self) -> &Self::Target;
+        }
+
+        #[lang = "index"]
+        pub trait Index<Idx: ?Sized> {
+            type Output: ?Sized;
+            fn index(&self, index: Idx) -> &Self::Output;
+        }
```
matklad: Let's do the same trick here as for `org.rust.lang.core.resolve.RsClosuresResolveTest` to avoid repeating index impl every time?

src/test/kotlin/org/rust/lang/core/type/RsIndexTypeInferenceTest.kt
```diff
@@ -129,4 +129,61 @@ class RsIndexTypeInferenceTest : RsTypificationTestBase() {
...
+
+        fn foo(v: Vec<u32>) -> u32 {
+            let p = &v;
+            let m = p[0];
+            m
+          //^ u32
+        }
+    """)
+
+    fun testTransitiveCoercion() = testExpr("""
```
matklad: Let's use `test transitive coercion`'like names for new tests?

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -163,7 +163,7 @@ class ImplLookup(private val project: Project, private val items: StdKnownItems)
         return subst + associated
     }
 
-    fun derefSequence(baseTy: Ty): Sequence<Ty> {
+    fun coercionSequence(baseTy: Ty): Sequence<Ty> {
```
matklad: Is this intended to include real coercions in the future? For the current implementation, `deref` makes more sense to me as a name...

Undin: This method already does coercion because it converts `TyArray` to `TySlice`.
But if you consider the previous name is better I'll revert renaming - I don't insist:)

matklad: Let's make it `coercion` then, @vlad20012 could change it later if he disagrees :)

Undin:
Fixed comments

matklad:
bors r+

Thanks!

PR #1767, opened 2017-09-22
TY: supported UFCS & multiple impls of the same trait

### LINKED ISSUE #1753: Mismatched types error E0308 when there is no error
With the new plugin 52, Version: 0.1.0.2069 and using the [jni-rs crate](https://crates.io/crates/jni), this error is reported at many places in a Rust / JNI codebase.

```rust
JObject::from(env.new_string("hello")?);
```

> mismatched types E0308
> expected JThrowable, found JString

```rust
JString::from(elem_obj)
```

> mismatched types E0308
> expected *mut _jobject, found JObject

JObject and JString can be converted to each other using ::from.
source: https://github.com/prevoty/jni-rs/blob/master/src/wrapper/objects/jstring.rs#L25

In the meantime, would there be an way to disable this check? It persists after disabling all inspections in IDEA Settings.
Thanks!

vlad20012:
Partially f'ixes #1753 
It doesn't affect path resolution for now, just type inference

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsReference.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsReference.kt
@@ -17,6 +17,8 @@ interface RsReference : PsiPolyVariantReference {
 
     fun advancedResolve(): BoundElement<RsCompositeElement>? = resolve()?.let { BoundElement(it) }
 
+    fun advancedCachedMultiResolve(): List<BoundElement<RsCompositeElement>> = multiResolve().map { BoundElement(it) }
+
     fun multiResolve(): List<RsCompositeElement>
 }
 
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -21,6 +21,7 @@ import org.rust.lang.core.types.ty.Mutability.IMMUTABLE
 import org.rust.lang.core.types.ty.Mutability.MUTABLE
 import org.rust.lang.core.types.type
 import org.rust.utils.forEachChild
+import org.rust.utils.zipValues
 
 fun inferTypesIn(fn: RsFunction): RsInferenceResult =
     RsInferenceContext().run { preventRecursion { infer(fn) } }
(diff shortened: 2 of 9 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -374,7 +375,44 @@ private class RsFnInferenceContext(
...
+        if (elements.size <= 1) return null
+
+        val traits = elements.asSequence()
+            .mapNotNull { it.element as? RsFunction }
+            .mapNotNull { it.owner as? RsFunctionOwner.Impl }
+            .mapNotNull { it.impl.traitRef?.resolveToTrait }
+            .toList()
+
+        if (traits.size == elements.size && traits.toSet().size == 1) {
+            val fnName = (elements.first().element).name
```
matklad: Useless parents?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -374,7 +375,44 @@ private class RsFnInferenceContext(
...
+    }
+
+    /** This works for `String::from` where impls of `From` trait found for `String` */
+    private fun resolveAmbiguity(elements: List<BoundElement<RsNamedElement>>): BoundElement<RsFunction>? {
+        if (elements.size <= 1) return null
+
+        val traits = elements.asSequence()
+            .mapNotNull { it.element as? RsFunction }
+            .mapNotNull { it.owner as? RsFunctionOwner.Impl }
+            .mapNotNull { it.impl.traitRef?.resolveToTrait }
```
matklad: Perhaps it's better to fuse these mapNotNull calls together?

src/main/kotlin/org/rust/lang/core/resolve/ref/RsReference.kt
```diff
@@ -17,6 +17,8 @@ interface RsReference : PsiPolyVariantReference {
 
     fun advancedResolve(): BoundElement<RsCompositeElement>? = resolve()?.let { BoundElement(it) }
 
+    fun advancedCachedMultiResolve(): List<BoundElement<RsCompositeElement>> = multiResolve().map { BoundElement(it) }
```
matklad: API for references is getting uglier and uglier :(

Not sure if we can significantly improve it ...

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
vlad20012: Heh, looks awful
```kotlin
val traits = elements.map {
            ((it.element as? RsFunction)?.owner as? RsFunctionOwner.Impl)?.impl?.traitRef?.resolveToTrait
        }
```

src/main/kotlin/org/rust/lang/core/resolve/ref/RsReference.kt
vlad20012: I'm going to move `RsPathReference.resolveInner` to a separate function (like `RsMethodReference`) and then dismantle `advancedResolve()` at all.

farodin91:
@vlad20012 You say this or should fix method in my example. Wouldn’t be cool to have a test

PR #1773, opened 2017-09-25
RES: resolve UFCS & assoc fn calls to a particular impls

vlad20012:
The PR make this `default` call to be resolved to the particular impl of the `Default` trait for `String`.
```rust
let a: String = Default::default();
```
And even more complicated stuff:
```rust
let a = From::from;
let b = a("");
let c: String = b;
```
Yes, this compiles by rustc and `from` now resolved to the particular impl `From` for `String` =)
Same with UFCS.

And the same approach should also be applied for regular method calls.

Really the current implementation does not suit me very much. It looks a bit ugly. Discussion is required =D

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -55,7 +55,7 @@ class RsPsiFactory(private val project: Project) {
         createExpressionOfType("unsafe { $body }")
 
     fun tryCreatePath(text: String): RsPath? {
-        val path = createFromText<RsPathExpr>("fn main() { $text;}")?.path ?: return null
+        val path = createFromText<RsPathExpr>("const a = $text;")?.path ?: return null
         if (path.text != text) return null
         return path
     }
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
@@ -9,6 +9,7 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.ResolveResult
 import com.intellij.psi.impl.source.resolve.ResolveCache
 import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.RsPathExpr
 import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.ext.RsCompositeElement
 import org.rust.lang.core.psi.ext.RsGenericDeclaration
@@ -19,6 +20,7 @@ import org.rust.lang.core.resolve.collectCompletionVariants
 import org.rust.lang.core.resolve.collectPathResolveVariants
 import org.rust.lang.core.resolve.processPathResolveVariants
 import org.rust.lang.core.types.BoundElement
+import org.rust.lang.core.types.inference
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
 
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -54,7 +54,7 @@ class RsPsiFactory(private val project: Project) {
         createExpressionOfType("unsafe { $body }")
 
     fun tryCreatePath(text: String): RsPath? {
-        val path = createFromText<RsPathExpr>("fn main() { $text;}")?.path ?: return null
+        val path = createFromText<RsPathExpr>("const a = $text;")?.path ?: return null
```
vlad20012: Since path expr resolution now delegated to type inference, `fn main() { $text;}` will perform type inference first, sometimes during type inference of another fn. Not only is it forbidden, it's also just useless work. `const a = $text;` is a very temporary solution because const exprs really should call type inference too. But there is no other syntax for **value** path including paths to associated items. Good news is that path to associated item (like `<A as B>::C`) used is tests only (khm, really in just a single test), so I think here we can use just `use $text;` and leave plain old `fn main() { $text;}` for test cases.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -33,6 +35,7 @@ fun inferTypesIn(fn: RsFunction): RsInferenceResult =
 class RsInferenceResult(
     private val bindings: Map<RsPatBinding, Ty>,
     private val exprTypes: Map<RsExpr, Ty>,
+    private val resolvedPaths: Map<RsPathExpr, List<RsCompositeElement>>,
```
vlad20012: Here we use `List<RsCompositeElement>` instead of list of bould elements because bound elements used only for type inference and also because is't hard get substitution for UFCS-like stuff. But it's not clean anyway...

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -51,6 +57,8 @@ class RsInferenceResult(
 class RsInferenceContext {
     private val bindings: MutableMap<RsPatBinding, Ty> = HashMap()
     private val exprTypes: MutableMap<RsExpr, Ty> = HashMap()
+    private val resolvedPaths: MutableMap<RsPathExpr, List<RsCompositeElement>> = HashMap()
+    private val pathRefinements: MutableList<Pair<RsPathExpr, TraitRef>> = mutableListOf()
```
vlad20012: I don't sure in `refinement` name. It's something like additional info for resolved path that should be used after type inference to provide more precise path (not to trait but to impl)

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -27,60 +29,62 @@ class RsPathReferenceImpl(
 
     override val RsPath.referenceAnchor: PsiElement get() = referenceNameElement
 
-    override fun resolveInner(): List<BoundElement<RsCompositeElement>> {
-        val lookup = ImplLookup.relativeTo(element)
-        val result = collectResolveVariants(element.referenceName) {
-            processPathResolveVariants(lookup, element, false, it)
-        }
+    override fun resolveInner(): List<BoundElement<RsCompositeElement>> =
+        (element.parent as? RsPathExpr)?.let { it.inference?.getResolvedPath(it) } ?: resolvePath(element)
```
vlad20012: We can't (and don't need to) delegate all path resolution to type inference, but it seems not clear whether path depends on type inference or not. May be create a special reference class for path of `RsPathExpr`?..

matklad:
I don't have enough time today to review this thoroughly, so I'd just say r=me with two notices:

1) The tests should pass :) 
2) Could you investigate performance/memory usage after this PR a bit thoroughly? There were some rather troubling reports recently :(

vlad20012:
Ready for review

PR #1782, opened 2017-09-25
ANN: Coerce *mut pointers to *const pointers to fix a wrong E0308

kumbayo:
In places that expect a *const T it is possible to pass a *mut T due to coercion.
See https://doc.rust-lang.org/nomicon/coercions.html

This fixes a wrong E0308 error when using a *mut T at a points where a *const T is expected.
(function argument, let statement, ...)

CC: #1753

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -340,6 +340,10 @@ private class RsFnInferenceContext(
                 expected is TyReference && expected.referenced is TySlice -> {
                 ctx.combineTypes(inferred.referenced.base, expected.referenced.elementType)
             }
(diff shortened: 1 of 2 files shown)


vlad20012:
bors r+

Excellent! This is done exactly as it should.
Thanks!

If you interesting on something like this, you can take a look rustc's implementation ([click](https://github.com/rust-lang/rust/blob/91dbf52af3de0436bcc032229540db1fe14b6df8/src/librustc_typeck/check/coercion.rs#L207)). That `coerce_unsafe_ptr` function is a something like you just wrote, but also coerce & to * ptrs.

kumbayo:
I will have a look at it.
Currently it does not matter yet because of
https://github.com/intellij-rust/intellij-rust/blob/2ba6e22f36fe58362a58f286baaca0d753fde595/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt#L316-L322

PR #1783, opened 2017-09-25
Show macro definitions and extern items in the Structure viewer 

kumbayo:


kumbayo:
Extern and non extern items use the same icons and are currently not distinguishable in the Structure viewer. Not sure if this is OK.

vlad20012:
@kumbayo it's anyway better than don't display them =)
Thanks!

bors r+

Issue #1732, opened 2017-09-13
Unify error annotation messages style


vlad20012:
Compare this two:
![image](https://user-images.githubusercontent.com/3221931/30381680-c4666d40-98a5-11e7-984e-1ae94bca7e6d.png)

![image](https://user-images.githubusercontent.com/3221931/30381613-904ace20-98a5-11e7-8484-5bf8d2183bf2.png)

First is from cargo check, second - our own inspection. 

Now we have a [good utils](https://github.com/intellij-rust/intellij-rust/blob/b42286d6b582bffa3b103cf05a9d0ad35d70b101/src/main/kotlin/org/rust/lang/core/types/RsDiagnostic.kt) for error messages formatting, and I think we shouldn't write [E0000] string manually anywhere.

We have a lot of hand-written messages [here](https://github.com/intellij-rust/intellij-rust/blob/b42286d6b582bffa3b103cf05a9d0ad35d70b101/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt), and they should be replaced with the new API.

shssoichiro:
I'll give this one a try.

farodin91:
@shssoichiro I could help if you need some?

shssoichiro:
@farodin91 Thanks. I might need help just figuring out how to use the RsDiagnostic class, I couldn't find any locations where it's being used right now based on IntelliJ's go-to usage, although I haven't had a lot of time to look quite yet.

farodin91:
1. I would move `RsDiagnostic` into `org.rust.core.utils` to split from type inference. 
2. Make `expectedFound` part `TypeError` until we need in a different location.
3. Switch the link to error description by using the channel.
4. Now start converting errors like https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt#L87-L89 into more clean error like `TypeError`
5. ... converting more

shssoichiro:
Thanks! I think I've got the idea now. I was having trouble grokking Kotlin's syntax at first, but now that I've figured out what the code is doing I should have a PR up in a day or two.

Jezza:
This issue seems OK to close, as it was fixed with the aforementioned PR.

PR #1789, opened 2017-09-28
Unify error annotation messages style

shssoichiro:
Addresses #1732. I'm not sure if the approach I took is the best one. I'm open to feedback to improve this solution.


vlad20012:
@shssoichiro btw we need a CLA signed first

The CLA process is described in [contributing.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#contributing). Copy-pasting here:

We require a contributor license agreement for all contributions. You can sign it electronically at https://www.jetbrains.com/agreements/cla/ via DocuSign. The procedure consists of a couple of emails and clicks: no need to scan or print or snail-mail anything :) CONTRIBUTORS.txt file stores GitHub usernames of people who already signed a CLA.

I will now review this PR. Could you also add your github nick to https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTORS.txt (the list is sorted alphabetically)

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -83,13 +86,13 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
                 is RsTraitItem -> return
             }
         }
-        val (elem, desc) = when (field) {
-            is RsFieldLookup -> field to "Attempted to access a private field on a struct. [E0616]"
-            is RsMethodCall -> field.identifier to "A private item was used outside of its scope. [E0624]"
-            is RsPath -> field to "A private item was used outside its scope. [E0603]"
+        val error = when (field) {
```
vlad20012: Maybe also rename `field` to simple `element`. It's not the only field now, but name remains old

shssoichiro:
I've signed the CLA and updated the CONTRIBUTORS.txt.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -0,0 +1,634 @@
...
+        }
+
+        private fun expectedFound(expectedTy: Ty, actualTy: Ty): String {
+            val expectedTyS = escapeString(expectedTy.toString())
+            val actualTyS = escapeString(actualTy.toString())
+            return "expected `$expectedTyS`, found `$actualTyS`"
+        }
+    }
+
+    class FieldAccessError(
```
vlad20012: There is a lot of copy-paste. FieldAccessError/MethodAccessError/PathAccessError seems a very similiar. I think we can generalize it as just `AccessError` and then save `RsErrorCode` as a field (and second constructor argument) and then switch aroung it.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -0,0 +1,634 @@
...
+    }
+}
+
+private class AnnotationHeader(
+    val plainText: String,
+    val html: String
+)
+
+private enum class RsErrorCode {
+    E0046,
```
vlad20012: I need to check kotlin style guide, but I think it's ok to write a multiple declarations on a single line

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -0,0 +1,634 @@
...
+        element.textRange,
+        header,
+        detailsHtml
+    )
+}
+
+private fun AnnotationHolder.createErrorAnnotation(
+    textRange: TextRange,
+    header: AnnotationHeader,
+    detailsHtml: String
```
vlad20012: detailsHtml: String = ""

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -123,7 +126,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
 
     private fun checkUnsafeCall(holder: AnnotationHolder, o: RsExpr) {
         if (!o.isInUnsafeBlockOrFn(/* skip the expression itself*/ 1)) {
-            val annotation = holder.createErrorAnnotation(o, "Call to unsafe function requires unsafe function or block [E0133]")
+            val annotation = RsDiagnostic.UnsafeError(o, "Call to unsafe function requires unsafe function or block").addToHolder(holder)
             annotation.registerFix(SurroundWithUnsafeFix(o))
```
vlad20012: On the idea, all the fixes should be added inside a `addToHolder` method, and so it shouldn't return an `Annotation`

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -41,7 +42,7 @@ private fun checkFunction(holder: AnnotationHolder, fn: RsFunction) {
         is RsFunctionOwner.Trait -> {
             deny(fn.default, holder, "${fn.title} cannot have the `default` qualifier")
             deny(fn.vis, holder, "${fn.title} cannot have the `pub` qualifier")
-            deny(fn.const, holder, "Trait functions cannot be declared const [E0379]")
+            if (fn.const != null) RsDiagnostic.ConstTraitFnError(fn.const!!).addToHolder(holder)
```
vlad20012: This can be replaced with `fn.const?.let { RsDiagnostic.ConstTraitFnError(it).addToHolder(holder) }` or just unwrapped to a local variable. Using of `!!` is a very bad practice (but it used widely in this file, I know)

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -0,0 +1,634 @@
...
+
+    class UnsafeError(
+        element: PsiElement,
+        private val message: String
+    ) : RsDiagnostic(element) {
+        override fun addToHolder(holder: AnnotationHolder): Annotation {
+            return holder.createErrorAnnotation(
+                element,
+                header(RsErrorCode.E0133, message),
+                ""
```
vlad20012: Let's make last parameter `= ""` by default ([kotlin default arguments](https://kotlinlang.org/docs/reference/functions.html#default-arguments))

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -276,18 +278,16 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         val selfArg = fn.selfParameter
 
         if (selfArg != null && superFn.selfParameter == null) {
-            holder.createErrorAnnotation(selfArg,
-                "Method `${fn.name}` has a `${selfArg.canonicalDecl}` declaration in the impl, but not in the trait [E0185]")
+            RsDiagnostic.DeclMissingFromTraitError(selfArg, fn, selfArg.canonicalDecl).addToHolder(holder)
         } else if (selfArg == null && superFn.selfParameter != null) {
-            holder.createErrorAnnotation(params,
-                "Method `${fn.name}` has a `${superFn.selfParameter?.canonicalDecl}` declaration in the trait, but not in the impl [E0186]")
+            RsDiagnostic.DeclMissingFromImplError(params, fn, "${superFn.selfParameter?.canonicalDecl}").addToHolder(holder)
```
vlad20012: It's preffered to pass a concrete objects to `RsDiagnostic`, not their string representation

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -0,0 +1,634 @@
...
+
+    class TypeError(
+        element: PsiElement,
+        private val expectedTy: Ty,
+        private val actualTy: Ty
+    ) : RsDiagnostic(element) {
+        override fun addToHolder(holder: AnnotationHolder): Annotation {
+            return holder.createErrorAnnotation(
+                element,
+                header(RsErrorCode.E0308, "mismatched types"),
```
vlad20012: I also suggest to add an import `RsErrorCode.*` and so use simply `E0308` and so on without this explicit `RsErrorCode.` everywhere

vlad20012:
Good job! I think we can r+ this after these minor fixes =)

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
vlad20012: Hmm, no, I decided to leave it as is =D. In future we will insert more codes into middle of the list, so any complex formatting will be hard to maintain...

shssoichiro:
Changes pushed.

vlad20012:
Oh, I just noticed `RsTextRangeDiagnostic`/`RsRequireResolveDiagnostic`.

`RsRequireResolveDiagnostic` seems really useless. You can just use something like `requireResolve` function that yoy removed from `RsErrorAnnotator`. I.E. if some diagnostics created, it should definitely be an annotation, not contextual "may be or may not".

`RsTextRangeDiagnostic` seems really not ok, but I can not quickly think of anything better.

vlad20012:
@shssoichiro Hi, do you need any help with moving this PR forward? If you don't have any spare time, I can help to push this over the finish line :)

shssoichiro:
Sorry, I've been busy for the past couple of days. I will get some changes pushed to this today. I'd like to get to a happy solution with no duplication of abstract classes, so hopefully I'll find some way to reach that now that I have the time to sit and think about it.

shssoichiro:
I've removed `RsRequireResolveDiagnostic`, but can't figure out a good way to remove `RsTextRangeDiagnostic`, since `TextRange` does not extend `PsiElement`.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -0,0 +1,567 @@
...
+): Annotation {
+    return createAnnotation(HighlightSeverity.ERROR,
+        textRange,
+        header.plainText,
+        "<html>${header.html}<br>$detailsHtml</html>"
+    )
+}
+
+private fun AnnotationHolder.requireResolve(el: RsReferenceElement, header: AnnotationHeader, message: String = "") {
+    if (el.reference.resolve() != null || el.reference.multiResolve().size > 1) return
```
vlad20012: Hmm, there is still some inconsistency. We creating `RsDiagnostics` first, then converting it to `Annotation`. But here we may or **may not** create an annotation due to this condition. Can you please try to move it on the upper level, before creating `RsDiagnostics`?

shssoichiro: Sorry, I must have missed the notification for this. I will make this change shortly.

vlad20012: Oh, and now it need to be rebased... There was an API changes

shssoichiro:
Should be good to go now

vlad20012:
Thanks!
bors r+

shssoichiro:
Conflicts fixed

shssoichiro:
Sorry, my fault. New commit pushed that compiles correctly.

PR #1791, opened 2017-09-28
Project model 3.0

matklad:
This is the third iteration of our project model.

The main technical change is that we no longer bind our project model to IDEA `Module`. This means that now it is possible to have several Cargo projects open simultaneously in all IDEs. It's also possible to provide a custom path to Cargo.toml file.

Another nice usability improvement is that now a list of projects is present in the Cargo Toolbar, together with their update status (is the info fresh, or were there some error during the previous update):

image

Those `+` `-` buttons on the toolbar are the main way to control cargo projects, if autoimport fails for some reason (for example, if Cargo.toml is not in the root of the project). 

r? @vlad20012 and anyone else! :)


src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    interface CargoProjectsListener {
+        fun cargoProjectsUpdated(projects: Collection<CargoProject>)
+    }
+}
+
+val Project.cargoProjects get() = service<CargoProjectsService>()
+
+interface CargoProject {
```
farodin91: Extra file?

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -0,0 +1,378 @@
...
+        // So the actual "Let's guess a project model if it is not imported
+        // explicitly" happens in [org.rust.ide.notifications.MissingToolchainNotificationProvider]
+    }
+
+    override fun toString(): String {
+        return "CargoProjectsService(projects = $allProjects)"
+    }
+}
+
+data class CargoProjectImpl(
```
farodin91: Extra file?

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -0,0 +1,378 @@
...
+import org.rust.utils.runAsyncTask
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicReference
+
+private val LOG = Logger.getInstance(CargoProjectsServiceImpl::class.java)
+
+@State(name = "CargoProjects")
+class CargoProjectsServiceImpl(
```
farodin91: I think testing of this service would be cool

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -0,0 +1,378 @@
...
+        // Do nothing: in theory, we might try to do [discoverAndRefresh]
+        // here, but the `RustToolchain` is most likely not ready.
+        //
+        // So the actual "Let's guess a project model if it is not imported
+        // explicitly" happens in [org.rust.ide.notifications.MissingToolchainNotificationProvider]
+    }
+
+    override fun toString(): String {
+        return "CargoProjectsService(projects = $allProjects)"
+    }
```
mkaput: =

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -0,0 +1,378 @@
...
+
+    private fun withWorkspace(result: TaskResult<CargoWorkspace>): CargoProjectImpl = when (result) {
+        is TaskResult.Ok -> copy(rawWorkspace = result.value, workspaceStatus = UpdateStatus.UpToDate)
+        is TaskResult.Err -> copy(workspaceStatus = UpdateStatus.UpdateFailed(result.reason))
+    }
+
+    override fun toString(): String {
+        return "CargoProject(manifest = $manifest)"
+    }
```
mkaput: =

src/main/kotlin/org/rust/lang/core/psi/ext/RsCompositeElement.kt
```diff
@@ -25,40 +24,36 @@ interface RsCompositeElement : PsiElement {
...
-        val root = superMods.lastOrNull()
-        if (root != null && root.isCrateRoot)
-            root
-        else
-            null
-    } else {
-        (context as? RsCompositeElement)?.crateRoot
+val RsCompositeElement.crateRoot: RsMod?
+    get() {
+        return if (this is RsFile) {
```
mkaput: =

mkaput: this extension feels like it should be just in RsCompositeElement as virtual method

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -199,7 +199,7 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
     }
 
     if (qualifier != null) {
-        val primitiveType = TyPrimitive.fromPath(qualifier);
+        val primitiveType = TyPrimitive.fromPath(qualifier)
```
mkaput: kill 'em all 👍

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
matklad: No, I intentionally have separated interface and `Impl` classes. Although in most cases it is stupid to have an interface with a single implementation only, I think it makes sense in this particular use-case, because the interface is rather important and should be used throughout the project, and the impl is quite involved.

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
matklad: An excellent suggestion @farodin91 !

I've added some additional test for `AsyncValue`, and looks like `Promise` has a bug, and we should switch to `CompletableFuture`!

src/main/kotlin/org/rust/lang/core/psi/ext/RsCompositeElement.kt
matklad: :+1: 💯

matklad:
image

PR #1792, opened 2017-09-28
Coerce references to pointers

kumbayo:
In places that expect a pointer it is possible to pass a reference due to coercion.
Coercion can also remove mutability.
See https://doc.rust-lang.org/nomicon/coercions.html

&T     -> *const T
&mut T -> *mut T
&mut T -> *const T

### DIFF
--- a/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
@@ -5,6 +5,7 @@
 
 package org.rust.ide.annotator
 
+import org.junit.ComparisonFailure
 import org.rust.fileTreeFromText
 
 class RsErrorAnnotatorTest : RsAnnotatorTestBase() {
@@ -948,7 +949,7 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase() {
         }
     """)
 
-    fun `test type mismatch E0308 ptr mutability`() = checkErrors("""
+    fun `test type mismatch E0308 coerce ptr mutability`() = checkErrors("""
         fn fn_const(p: *const u8) { }
         fn fn_mut(p: *mut u8) { }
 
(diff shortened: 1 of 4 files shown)


kumbayo:
This is a WIP not meant for immediate inclusion.
It removes a test from RsCargoCheckAnnotatorTest.kt that would fail because of new improved type checking.
I am not sure how to properly fix this.

This commit would also cause new false positives because of errors around TyReference.
See TypeInference.coerceResolved()

vlad20012:
I suggest to focus not on type checking but on type inference. The main meaning of `coerce` method is unification of two types passed to it. It's simple to write a type inference test that fails now, but will pass after your improvements:
```rust
fn foo(_: *const u8) {}
fn main() {
    let a = 0;
          //^ u8
    foo(&a)
}
```
You do not really need to write annotator tests, type inference tests is enough. If we move this way, we don't need to exclude `TyReference` from `ignoredTys` =)

kumbayo:
Oh at the beginning i did not notice that the result of tryCoerce() already influences type inference. :-)
I agree with not changing the TyReference part now.
But i still think there is some usefulness to the annotator test as it shows when coercion is possible and when it is not possible due to wrong mutability.

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -966,6 +967,22 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase() {
         }
     """)
 
+    fun `test type mismatch E0308 coerce reference to ptr`() = expect<ComparisonFailure> {
```
vlad20012: Hmm, if the test currently not passes let's add some TODO comment describes it

vlad20012:
> But i still think there is some usefulness to the annotator test as it shows when coercion is possible and when it is not possible due to wrong mutability.

Hmm, yes, I agree. We have a very few typechech tests for now =(

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
kumbayo: Done

vlad20012:
bors r+

Thanks!

PR #1796, opened 2017-09-29
TY: Improve array expr type inference

### LINKED ISSUE #1790: TY: Introduce array of unknown size
This code is ok, but syntax checker marks array as wrong

```rust
const COUNT: usize = 2;

struct Foo {
    v: [usize; COUNT]
}

fn bar(f: Foo) {}

fn main() {
    bar(Foo{
        v:[10, 20] // E308, expected [usize], found [usize; 2]
    });
}
```

[Mentoring instructions can be found in this comment](https://github.com/intellij-rust/intellij-rust/issues/1790#issuecomment-333066164)

Undin:
* Introduce arrays with unknown size
* ~~Implement rough array size expression calculation~~
* Use long for array size value

Fixes #1790

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
@@ -10,17 +10,19 @@ import org.rust.lang.core.psi.RsExpr
 import org.rust.lang.core.psi.RsLitExpr
 import org.rust.lang.core.types.ty.TyInteger
 
-val RsArrayType.arraySize: Int? get() {
+val RsArrayType.isSlice: Boolean get() = stub?.isSlice ?: (expr == null)
+
+val RsArrayType.arraySize: Long? get() {
     val stub = stub
     if (stub != null) {
-        return if (stub.arraySize == -1) null else stub.arraySize
+        return if (stub.arraySize == -1L) null else stub.arraySize
     }
     return calculateArraySize(expr)
 }
 
 // TODO: support constants and compile time expressions
-fun calculateArraySize(expr: RsExpr?): Int? = (expr as? RsLitExpr)
+fun calculateArraySize(expr: RsExpr?): Long? = (expr as? RsLitExpr)
     ?.integerLiteral
     ?.text
     ?.removeSuffix(TyInteger.Kind.usize.name)
-    ?.toIntOrNull()
+    ?.toLongOrNull()
--- a/src/main/kotlin/org/rust/lang/core/types/ty/TyArray.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/TyArray.kt
@@ -10,7 +10,7 @@ import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.infer.TypeFolder
 import org.rust.lang.core.types.infer.TypeVisitor
 
-class TyArray(val base: Ty, val size: Int?) : Ty {
+class TyArray(val base: Ty, val size: Long?) : Ty {
     override fun unifyWith(other: Ty, lookup: ImplLookup): UnifyResult =
         if (other is TyArray && (size == null || other.size == null || size == other.size))
             base.unifyWith(other.base, lookup) else UnifyResult.fail
--- a/src/main/kotlin/org/rust/lang/core/types/ty/TyArray.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/TyArray.kt
@@ -10,9 +10,10 @@ import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.infer.TypeFolder
 import org.rust.lang.core.types.infer.TypeVisitor
 
-class TyArray(val base: Ty, val size: Int) : Ty {
+class TyArray(val base: Ty, val size: Int?) : Ty {
     override fun unifyWith(other: Ty, lookup: ImplLookup): UnifyResult =
-        if (other is TyArray && size == other.size) base.unifyWith(other.base, lookup) else UnifyResult.fail
+        if (other is TyArray && (size == null || other.size == null || size == other.size))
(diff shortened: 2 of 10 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
```diff
@@ -18,9 +17,42 @@ val RsArrayType.arraySize: Int? get() {
...
+
+private fun tryEvaluateExpr(expr: RsExpr?): Int? {
+    return when (expr) {
+        is RsLitExpr -> expr.integerLiteral
+            ?.text
+            ?.removeSuffix(TyInteger.Kind.usize.name)
+            ?.toIntOrNull()
+            ?.notNegativeOrNull()
+        is RsPathExpr -> {
+            val const = expr.path.reference.resolve() as? RsConstant ?: return null
```
Undin: Call of `resolve` method leads to `IndexNotReadyException` because we save array size in stub and try to calculate it while indexing.
I suggest to drop array size filed from `RsArrayTypeStub` because it's more semantic then syntax part of language. And it'll allow to calculate array size expressions, of course:)

src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
```diff
@@ -18,9 +17,42 @@ val RsArrayType.arraySize: Int? get() {
...
+            val integerType = const.typeReference?.type as? TyInteger ?: return null
```
Undin: Maybe use `TyPrimitive.fromPath` directly instead of full type inference?

src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
vlad20012: Hmm, here we need to be very careful. We have an invariant that we looking only at a single file while indexing, so we can't resolve constants (because they can be defined in another files) during it. We need to looking deeply where the indexing need an array size... and is it really need it.

Also we don't want to falling back to AST while resolve (e.g. we resolved a function that return an array with constexpr size, and falled back to AST). Hmm...

And also maybe we want to save the size to stub only when it is trivial (the most case)

src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
vlad20012: Umm... maybe =D

src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
vlad20012: Also note that `RsConstant` element describes both `const` and `static`!

Undin: > Hmm, here we need to be very careful. We have an invariant that we looking only at a single file while indexing, so we can't resolve constants during it (because they can be defined in another files).

Yes, I know about this restriction, that's why I've proposed to drop indexing of array size - it becomes useless because of necessity of complex computation.

> Also we don't want to falling back to AST during resolve (e.g. we resolved a function that return an array with constexpr size, and falled back to AST). Hmm...
>
>And also maybe we want to save the size to stub only when it is trivial (the most case)

I think it won't help - we still need to calculate value of integer literal expressions and while it we fall back to AST because we don't create stub for them.
To solve this problem I think we can create stub for literal expressions and save integer value in it. It'll help us not to fall back to AST during type inference but at same time index will contain **all** integers values not only part of array size expr.

Undin: Although we can save values only for constants and array index expr

vlad20012:
Umm, there is a some unresolved (or at least hard to solve) questions with constexpr evaluation, so maybe we can land a trivial part with unknown size first?

Undin:
Ok, I'll make corresponding changes a bit later:)

Undin:
Done

Undin:
Done

src/main/kotlin/org/rust/lang/core/psi/ext/RsArrayType.kt
```diff
@@ -10,17 +10,19 @@ import org.rust.lang.core.psi.RsExpr
...
 import org.rust.lang.core.types.ty.TyInteger
 
-val RsArrayType.arraySize: Int? get() {
+val RsArrayType.isSlice: Boolean get() = stub?.isSlice ?: (expr == null)
+
+val RsArrayType.arraySize: Long? get() {
     val stub = stub
     if (stub != null) {
-        return if (stub.arraySize == -1) null else stub.arraySize
+        return if (stub.arraySize == -1L) null else stub.arraySize
```
vlad20012: Hmm, interestring. We can also use `-1` in TyArray too to avoid boxing...

vlad20012:
bors r+

Thanks!

src/main/kotlin/org/rust/lang/core/types/ty/TyArray.kt
```diff
@@ -10,9 +10,10 @@ import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.infer.TypeFolder
 import org.rust.lang.core.types.infer.TypeVisitor
 
-class TyArray(val base: Ty, val size: Int) : Ty {
+class TyArray(val base: Ty, val size: Long?) : Ty {
     override fun unifyWith(other: Ty, lookup: ImplLookup): UnifyResult =
-        if (other is TyArray && size == other.size) base.unifyWith(other.base, lookup) else UnifyResult.fail
+        if (other is TyArray && (size == null || other.size == null || size == other.size))
+            base.unifyWith(other.base, lookup) else UnifyResult.fail
```
vlad20012: Usually when `if` is multiline we wrap it with `{ }`, but I want to merge it right now and `unifyWith` will be dropped soon :)

PR #1804, opened 2017-10-01
Support explicit self type in method parameter 

kumbayo:
The self parameter of a method can be written in a short hand and in an explicit variant.
The explicit variant also supports a special case for Box<Self>.

```
self      -> self: Self
&self     -> self: &Self
&mut self -> self: &mut Self
-         -> self: Box<Self>
```

Fix type the inference for this explicit type syntax on the self parameter.

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
@@ -241,6 +241,96 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
         }
     """)
 
+    fun `test self 1`() = testExpr("""
+        struct Foo;
+        impl Foo {
+            fn foo(self) {
+                self;
+              //^ Foo
+            }
+        }
+    """)
+
+    fun `test self 2`() = testExpr("""
+        struct Foo;
+        impl Foo {
+            fn foo(self: Self) {
+                self;
+              //^ Foo
+            }
+        }
+    """)
+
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -241,6 +241,96 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
         }
     """)
 
+    fun `test self 1`() = testExpr("""
```
vlad20012: We still did not have plain `self` param tests? I'm failed to find them, but it looks really weird...

src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -241,6 +241,96 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
...
+            fn foo(self) {
+                self;
+              //^ Foo
+            }
+        }
+    """)
+
+    fun `test self 2`() = testExpr("""
+        struct Foo(u8);
+        fn main() { }
```
vlad20012: Why we need `fn main() { }` and that `u8` field?

kumbayo: Not needed, just copy paste leftovers.
Should be fixed now.

vlad20012:
bors r+

Thanks!

PR #1805, opened 2017-10-02
Let the user know to install cargo fmt

### LINKED ISSUE #1131: "Reformat file with rustfmt" gives confusing error if rustfmt is not installed
I hadn't realized that `rustfmt` is a separate package that isn't included by default with Cargo, and I hadn't installed it.

Without `rustfmt` installed, IntelliJ-Rust gives this error message:

 > C:\Users\<me>\.cargo\bin\cargo.exe fmt -- --write-mode=overwrite --skip-children <my-project>/<my-file>.rs stdout : stderr : error: no such subcommand: `fmt` Did you mean `doc`?

No, I did not mean `doc`, silly tool!

Obviously this is user error, but I believe IntelliJ-Rust could probably handle this fairly easily, i.e. by checking for the above error and suggesting to the user to manually run `cargo install rustfmt`, or by simply automatically checking if `rustfmt` is installed and doing the install automatically if not.

maraisr:
Never written a single line of Kotlin before - so please be nice! :heart: 

fixes #1131

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt
@@ -49,14 +49,10 @@ class RsFmtFileAction : DumbAwareAction() {
             cargo.reformatFile(module, file)
         } catch (e: ExecutionException) {
             // #1131 - Check if we get a `no such subcommand: fmt` and let the user know to install fmt
-            val messageCheck = Regex("no such subcommand: fmt");
-
-            messageCheck.matchEntrie(e.message)?.let { match ->
-                if ( match != null ) {
-                    project.showBalloon("Install Cargo fmt: https://github.com/rust-lang-nursery/rustfmt", NotificationType.ERROR)
-                } else {
-                    project.showBalloon(e.message ?: "", NotificationType.ERROR)
-                }
+            if ("no such subcommand: fmt" in e.message) {
+                project.showBalloon("Install rustfmt: https://github.com/rust-lang-nursery/rustfmt", NotificationType.ERROR)
+            } else {
+                project.showBalloon(e.message ?: "", NotificationType.ERROR)
             }
         }
     }
--- a/src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt
@@ -48,11 +48,13 @@ class RsFmtFileAction : DumbAwareAction() {
         try {
             cargo.reformatFile(module, file)
         } catch (e: ExecutionException) {
+            val message = e.message ?: ""
+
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt
```diff
@@ -48,7 +48,16 @@ class RsFmtFileAction : DumbAwareAction() {
         try {
             cargo.reformatFile(module, file)
         } catch (e: ExecutionException) {
-            project.showBalloon(e.message ?: "", NotificationType.ERROR)
+            // #1131 - Check if we get a `no such subcommand: fmt` and let the user know to install fmt
+            val messageCheck = Regex("no such subcommand: fmt");
```
matklad: In Kotlin, there's a handy `toRegex` extension, so it's more idiomatic to write `"no such subcommand: fmt".toRegex()`.

However in this case the regex is trivial, so we can just search for a substring, using the `in` operator:

```
if ("no such subcommand: fmt" in e.message) {
    ...
}
```

matklad:
Great, thanks @maraisr !

Please note that we require CLA's for pull request, as per our [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#contributing):

>We require a contributor license agreement for all contributions. You can sign it electronically at https://www.jetbrains.com/agreements/cla/ via DocuSign. The procedure consists of a couple of emails and clicks: no need to scan or print or snail-mail anything :) CONTRIBUTORS.txt file stores GitHub usernames of people who already signed a CLA.

Feel free to add yourself to CONTRIBUTORS.txt after signing :)

src/main/kotlin/org/rust/ide/actions/RsFmtFileAction.kt
```diff
@@ -48,7 +48,12 @@ class RsFmtFileAction : DumbAwareAction() {
         try {
             cargo.reformatFile(module, file)
         } catch (e: ExecutionException) {
-            project.showBalloon(e.message ?: "", NotificationType.ERROR)
+            // #1131 - Check if we get a `no such subcommand: fmt` and let the user know to install fmt
+            if ("no such subcommand: fmt" in e.message) {
```
matklad: Hm, looks like `e.message` is nullable on at least one version of IJ. I think we can do

```
val message = e.message ?: ""
```

and then use it in this condition and also in the else branch.

maraisr: I'll after this PR definitely go learn Kotlin! Silly mistakes like this shouldn't be happening.

maraisr:
@matklad  thank you again so much for this!

matklad:
This is already released: https://intellij-rust.github.io/2017/10/02/changelog-53.html :)

PR #1810, opened 2017-10-02
[#1797] Unsafe block checked

### LINKED ISSUE #1797: "Remove braces from single expression" false positive for "unsafe" block
```rust
let slice = &[1];
let start = slice.as_ptr();
let end = unsafe { start.offset(1) };
```
Put your cursor on the `unsafe` keyword -- there is a suggestion to "Remove braces from single expression". But I apply the quickfix, I will get a compile error `error[E0133]: call to unsafe function requires unsafe function or block`.

maraisr:
Not too sure what the issue #1797 was talking about - but thinking it's something along these lines?

fixes #1797

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/UnwrapSingleExprIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/UnwrapSingleExprIntention.kt
@@ -19,6 +19,7 @@ class UnwrapSingleExprIntention : RsElementBaseIntentionAction<RsBlockExpr>() {
 
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsBlockExpr? {
         val blockExpr = element.parentOfType<RsBlockExpr>() ?: return null
+        if (blockExpr.unsafe != null) return null
         val block = blockExpr.block
 
         return if (block.expr != null && block.lbrace.getNextNonCommentSibling() == block.expr)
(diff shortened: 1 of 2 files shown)


vlad20012:
Yep, this line should fix the issue. But it would be nice to provide some tests of this case :)

The link to the tests provided in the ofirinal issue - ([click](https://github.com/intellij-rust/intellij-rust/blob/fcb26cbdbd7ae7a654597413b52862f3e115f58b/src/test/kotlin/org/rust/ide/intentions/UnwrapSingleExprIntentionTest.kt#L85)). Try to write some `doUnavailableTest` test that fails without your fix, but pass with it.

vlad20012:
This test does not actually a test for this case. I.e. it passes without your fix :)

Also we have some commit messages conventions https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#commit-messages
Here it's `ANN:` and `T:`, but you can also commit push tests in the same commit as the fix, so the only `ANN:` remains.

vlad20012:
@maraisr Yep, not `ANN:` but `INT:` here =)
Github don't send notifications on commit amending, so I did not notice that you fixed it. 

Everything is OK now, thanks!

bors r+

maraisr:
My pleasure @vlad20012  - onto the next one!

PR #1811, opened 2017-10-03
Improve AddStructFieldsFix to support fill struct & fill struct recurively.

### LINKED ISSUE #1794: Fill struct & fill struct recursively intention
Shortly, implement this for Rust: https://twitter.com/dlsniper/status/913458055039373312

pocket7878:
Fixes #1794 .

I'll try to support by checking type name.
But is it correct way to check type of struct field?

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
@@ -59,9 +59,11 @@ class RsExpressionAnnotator : Annotator {
         } else {
             if (missingFields.isNotEmpty()) {
                 val structNameRange = expr.parent.childOfType<RsPath>()?.textRange
-                if (structNameRange != null)
-                    holder.createErrorAnnotation(structNameRange, "Some fields are missing")
-                        .registerFix(AddStructFieldsFix(decl.namedFields, missingFields, expr), expr.parent.textRange)
+                if (structNameRange != null) {
+                    val annotation = holder.createErrorAnnotation(structNameRange, "Some fields are missing")
+                    annotation.registerFix(AddStructFieldsFix(decl.namedFields, missingFields, expr), expr.parent.textRange)
+                    annotation.registerFix(AddStructFieldsFix(decl.namedFields, missingFields, expr, recursive = true), expr.parent.textRange)
+                }
             }
         }
 
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
@@ -10,11 +10,18 @@ import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiElementFactory
 import com.intellij.psi.PsiFile
 import org.rust.ide.formatter.RsTrailingCommaFormatProcessor
 import org.rust.ide.formatter.impl.CommaList
+import org.rust.ide.navigation.goto.RsClassNavigationContributor
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.elementType
+import org.rust.lang.core.psi.ext.namedFields
+import org.rust.lang.core.resolve.StdKnownItems
+import org.rust.lang.core.types.ty.*
+import org.rust.lang.core.types.type
+import kotlin.reflect.jvm.internal.impl.resolve.scopes.TypeIntersectionScope
 
 /**
  * Adds the given fields to the stricture defined by `expr`
@@ -22,9 +29,16 @@ import org.rust.lang.core.psi.ext.elementType
 class AddStructFieldsFix(
     private val declaredFields: List<RsFieldDecl>,
     private val fieldsToAdd: List<RsFieldDecl>,
-    structBody: RsStructLiteralBody
+    structBody: RsStructLiteralBody,
+    private val recursive: Boolean = false
 ) : LocalQuickFixAndIntentionActionOnPsiElement(structBody) {
-    override fun getText(): String = "Add missing fields"
+    override fun getText(): String {
+        return if (recursive) {
+            "Recursively add missing fields"
+        } else {
+            "Add missing fields"
+        }
+    }
 
     override fun getFamilyName(): String = text
 
@@ -37,12 +51,26 @@ class AddStructFieldsFix(
     ) {
         val psiFactory = RsPsiFactory(project)
         var structLiteral = startElement as RsStructLiteralBody
+        val (firstAdded, _) = fillStruct(psiFactory, structLiteral, declaredFields, fieldsToAdd, true, recursive)
 
+        if (editor != null && firstAdded != null) {
+            editor.caretModel.moveToOffset(firstAdded.expr!!.textOffset)
+        }
+    }
+
+    private fun fillStruct(
+        psiFactory: RsPsiFactory,
+        structLiteralBody: RsStructLiteralBody,
+        declaredFields: List<RsFieldDecl>,
+        fieldsToAdd: List<RsFieldDecl>,
+        postProcess: Boolean = true,
+        recursive: Boolean): Pair<RsStructLiteralField?, RsStructLiteralBody> {
+        var structLiteral = structLiteralBody
         val forceMultiline = structLiteral.structLiteralFieldList.isEmpty() && fieldsToAdd.size > 2
 
         var firstAdded: RsStructLiteralField? = null
         for (fieldDecl in fieldsToAdd) {
-            val field = psiFactory.createStructLiteralField(fieldDecl.name!!)
+            val field = specializedCreateStructLiteralField(psiFactory, fieldDecl)!!
             val addBefore = findPlaceToAdd(field, structLiteral.structLiteralFieldList, declaredFields)
             ensureTrailingComma(structLiteral.structLiteralFieldList)
 
@@ -58,12 +86,13 @@ class AddStructFieldsFix(
             structLiteral.addAfter(psiFactory.createNewline(), structLiteral.lbrace)
         }
 
-        structLiteral = CodeInsightUtilBase.forcePsiPostprocessAndRestoreElement(structLiteral)
-        RsTrailingCommaFormatProcessor.fixSingleLineBracedBlock(structLiteral, CommaList.forElement(structLiteral.elementType)!!)
+        if (postProcess) structLiteral = CodeInsightUtilBase.forcePsiPostprocessAndRestoreElement(structLiteral)
 
-        if (editor != null && firstAdded != null) {
-            editor.caretModel.moveToOffset(firstAdded.expr!!.textOffset)
-        }
+        RsTrailingCommaFormatProcessor.fixSingleLineBracedBlock(
+            structLiteral,
+            CommaList.forElement(structLiteral.elementType)!!)
+
+        return Pair(firstAdded, structLiteral)
     }
 
     private fun findPlaceToAdd(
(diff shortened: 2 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +106,19 @@ class AddStructFieldsFix(
 
         return null
     }
+
+    private fun specializedCreateStructLiteralField(factory: RsPsiFactory, fieldDecl: RsFieldDecl): RsStructLiteralField? {
+        val fieldTypeRef = fieldDecl.typeReference ?: return null
```
matklad: The types are represented by `Ty` classes in our plugin. To get a `Ty` from a `typeReference`, you can use `.type` extension method. 

So this could look like

```
private val Ty.defaultValue: String = ...
```

and then 

```
fieldDecl.typeReference?.type?.defaultValue
```

farodin91:
An idea would look for default impl and take these values.

pocket7878:
Maybe we can lookup Default trait implementation. 
But I think it will be hard to calculate default()'s return variable. 
Because there is no restriction to default()'s implementation, so implementation can be complex code.
Is there any way execute code on-the-fly(with low overhead)?

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +109,47 @@ class AddStructFieldsFix(
...
+                }
+                is TyArray -> {
+                    val size = this.size ?: 0
+                    val elementDefault = this.base.defaultValue
+                    val accum = mutableListOf<String>()
+                    //Kotlin array generator's size is only allowed Int, But TyArray's size is Long.
+                    for (i in 1..size) {
+                        accum += elementDefault
+                    }
+                    return "[" + accum.joinToString(" , ") + "]"
```
matklad: This could be simplified as `accum.joinString(prefix="[", postfix="]")`

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +109,47 @@ class AddStructFieldsFix(
...
+
+    private val Ty.defaultValue: String
+        get() {
+            return when (this) {
+                is TyBool -> "false"
+                is TyInteger -> "0"
+                is TyFloat -> "0.0"
+                is TyStr -> "\"\""
+                is TyStruct -> {
+                    when (this.item.name) {
```
matklad: You don't have to write explicit `this` here.

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +109,47 @@ class AddStructFieldsFix(
...
+                        //FIXME: should be check by absolute name e.g "std::option::Option"
+                        "Option" -> "None"
+                        else -> "()"
+                    }
+                }
+                is TyArray -> {
+                    val size = this.size ?: 0
+                    val elementDefault = this.base.defaultValue
+                    val accum = mutableListOf<String>()
+                    //Kotlin array generator's size is only allowed Int, But TyArray's size is Long.
```
matklad: Hm, what if the type is `[i8; 256]`? I don't think we should insert 256 values. Let's use `[]`? Though it has a wrong type, I think it's a reasonable default nonetheless.

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +109,47 @@ class AddStructFieldsFix(
...
+                    for (i in 1..size) {
+                        accum += elementDefault
+                    }
+                    return "[" + accum.joinToString(" , ") + "]"
+                }
+                is TyTuple -> {
+                    return "(" + this.types.map { it.defaultValue }.joinToString(" , ") + ")"
```
matklad: Can be simplified using prefix, postfix and transform args of `joinToString`.

matklad:
>Maybe we can lookup Default trait implementation.
> But I think it will be hard to calculate default()'s return variable.

Yeah, I agree with this reasoning. We could use `Default::default()`, but I think it's better to stick to a single simple literal if possible, because the user will most likely change it anyway, so it must be easy to delete even if one does not know about expand selection.


Looks great so far @pocket7878 !

pocket7878:
I'd just updated my code to support recursive fill struct.
But code looks little complicated. I'll try to refactoring.

src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
```diff
@@ -59,9 +59,12 @@ class RsExpressionAnnotator : Annotator {
         } else {
             if (missingFields.isNotEmpty()) {
                 val structNameRange = expr.parent.childOfType<RsPath>()?.textRange
-                if (structNameRange != null)
+                if (structNameRange != null) {
                     holder.createErrorAnnotation(structNameRange, "Some fields are missing")
                         .registerFix(AddStructFieldsFix(decl.namedFields, missingFields, expr), expr.parent.textRange)
+                    holder.createErrorAnnotation(structNameRange, "Some fields are missing")
+                        .registerFix(AddStructFieldsFix(decl.namedFields, missingFields, expr, true), expr.parent.textRange)
```
matklad: For boolean arguments, it's better to specify parameter name at the call site for better readability.

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -60,12 +69,50 @@ class AddStructFieldsFix(
...
+            if (firstAdded == null) {
+                firstAdded = added
+            }
+        }
+
+        if (forceMultiline) {
+            structLiteral.addAfter(psiFactory.createNewline(), structLiteral.lbrace)
+        }
+
+        if (postProcess)
```
matklad: Missing braces?

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -59,8 +59,17 @@ class RsPsiFactory(private val project: Project) {
...
-        createExpressionOfType<RsStructLiteral>("S { $name: () }").structLiteralBody.structLiteralFieldList[0]
+    fun createStructLiteral(name: String): RsStructLiteral =
+        createExpressionOfType<RsStructLiteral>("$name { }")
+
+    fun createStructLiteralField(name: String, value: RsExpr? = null): RsStructLiteralField {
+        val structLiteralField = createExpressionOfType<RsStructLiteral>("S { $name: () }")
+            .structLiteralBody
+            .structLiteralFieldList[0]
+        if (value != null)
+            structLiteralField.expr?.replace(value)
```
matklad: For short braceless ifs we usually place condition and action on the same line, to avoid gotofail bug:

```
if (value != null) structLIteralField.expr?.replace(value)
```

src/test/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFixTest.kt
```diff
@@ -136,10 +136,10 @@ class AddStructFieldsFixTest : RsAnnotatorTestBase() {
...
         fn main() {
             Mesh{
-                name: (),
+                name: String::from(""),
                 vertices: (),
                 faces: (),
-                material: (),
+                material: None,
             }
         }
```
matklad: Let's add a recursive test as well?

pocket7878:
Hmm.. can't detect struct type in test case. Got a <unknown> as a type name..

pocket7878:
Added TyUnknown fallback and Vec support!

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +158,68 @@ class AddStructFieldsFix(
...
+            is TyInteger -> factory.createExpression("0")
+            is TyFloat -> factory.createExpression("0.0")
+            is TyStr -> factory.createExpression("\"\"")
+            is TyReference -> {
+                defaultValueExprFor(factory, fieldDecl, ty.referenced)
+            }
+            is TyStruct -> {
+                val structName = ty.item.name ?: return factory.createExpression("()")
+                when (structName) {
+                    "String" -> factory.createExpression("String::from(\"\")")
```
vlad20012: It's better to use `StdKnownItems` here to match type exaclty

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +158,68 @@ class AddStructFieldsFix(
...
+                }
+            }
+            is TyArray -> {
+                factory.createExpression("[]")
+            }
+            is TyTuple -> {
+                factory.createExpression(ty.types.map { defaultValueExprFor(factory, fieldDecl, it).text }.joinToString(prefix = "(", separator = ",", postfix = ")"))
+            }
+            is TyUnknown -> {
+                val typeName = fieldDecl.typeReference?.text ?: return factory.createExpression("()")
```
vlad20012: Hmm, this path seems not much useful? I'll venture to assume that this path was added just because current tests are not std-aware. To make them so, just add `override fun getProjectDescriptor() = WithStdlibRustProjectDescriptor` line to `AddStructFieldsFixTest ` class.

vlad20012:
> Maybe we can lookup Default trait implementation.

I pondered this question a bit, and came to the conclusion that this is not the best idea. But I wrote some code before... so here is it =D
```rust
val items = StdKnownItems.relativeTo(structLiteral)
val lookup = ImplLookup(structLiteral.project, items)

val defaultImplBody = items.findCoreItem("default::Default")
    ?.let { it as? RsTraitItem }
    ?.let { lookup.select(TraitRef(ty, BoundElement(it))) }
    ?.let { (it as? SelectionResult.Ok)?.impl?.members?.functionList?.singleOrNull()?.block?.expr }
```

pocket7878:
@vlad20012 Thank you for your review! Fixed!

vlad20012:
As I see, this is almost complete. Good work, @pocket7878! Thanks!
I just noticed that some of the paths have not been tested. It would be nice to add fileds of all supported types to testing structs. I.e. bool, char, str and so on.
And if all work is completed, could you please squash all your commits in this PR and prefix the message of resulting commit with "INT: ".

src/main/kotlin/org/rust/ide/annotator/RsExpressionAnnotator.kt
```diff
@@ -59,9 +59,12 @@ class RsExpressionAnnotator : Annotator {
         } else {
             if (missingFields.isNotEmpty()) {
                 val structNameRange = expr.parent.childOfType<RsPath>()?.textRange
-                if (structNameRange != null)
+                if (structNameRange != null) {
                     holder.createErrorAnnotation(structNameRange, "Some fields are missing")
                         .registerFix(AddStructFieldsFix(decl.namedFields, missingFields, expr), expr.parent.textRange)
+                    holder.createErrorAnnotation(structNameRange, "Some fields are missing")
```
vlad20012: Just noticed this one. You don't need to created multiple annotations to add multiple fixed. Just extract `Annotation` to a variable and then call `registerFix` on it as much as you need.

```kotlin
val ann = holder.createErrorAnnotation(structNameRange, "Some fields are missing")
ann.registerFix()
ann.registerFix()
```

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +134,58 @@ class AddStructFieldsFix(
...
+        return when (ty) {
+            is TyBool -> factory.createExpression("false")
+            is TyInteger -> factory.createExpression("0")
+            is TyFloat -> factory.createExpression("0.0")
+            is TyChar -> factory.createExpression("''")
+            is TyStr -> factory.createExpression("\"\"")
+            is TyReference -> {
+                defaultValueExprFor(factory, fieldDecl, ty.referenced)
+            }
+            is TyEnum -> {
```
vlad20012: I don't think we should separate handling of struct and enums. We can manage with single branch `is TyStructOrEnumBase`.

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +134,58 @@ class AddStructFieldsFix(
...
+                when(ty.item.declaredType) {
+                    stdknownItems.findCoreTy("option::Option") -> factory.createExpression("None")
```
vlad20012: This looks like useless work. We can switch `ty.item` and match it with `stdknownItems.findCoreItem()`

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -105,4 +134,58 @@ class AddStructFieldsFix(
...
+                            factory.createExpression("()")
+                        }
+                    }
+                }
+            }
+            is TyArray -> {
+                factory.createExpression("[]")
+            }
+            is TyTuple -> {
+                factory.createExpression(ty.types.map { defaultValueExprFor(factory, fieldDecl, it).text }.joinToString(prefix = "(", separator = ",", postfix = ")"))
```
vlad20012: Should we use ", " separator intead?

pocket7878:
@vlad20012 Thank you for your mentoring! I just squashed my commits!.

vlad20012:
Excellent! Thanks!
bors r+

vlad20012:
How `AddStructFieldsFix` can be improved further:
1. Better support &ref types; i.e. in struct like `struct S { f: &i32 }` insert `&` op before literal: `S { f: &0 }`
2. Support more container types: `Box`, `Rc`, `Arc`, `Weak`, `Cell`, `RefCell`, `UnsafeCell`, `Mutex` (etc?)

PR #1828, opened 2017-10-06
CARGO: Fix channel selection command lines

### LINKED ISSUE #1827: Commands don't run `error: toolchain 'Nightly' is not installed`
This bug is due to #1826. Lack of testing as warned for in the PR ...

sanmai-NL:
Rustup channel names are case sensitive. Remove distinction between actual
channel names and their UI caption (with initial capitals).

Resolves #1827


matklad:
@sanmai-NL added a test in https://github.com/intellij-rust/intellij-rust/pull/1838, thanks again!

PR #1829, opened 2017-10-07
TY: evaluate const expr array size

Undin:


Undin:
Do we have good way to check performance of stub creation?

vlad20012:
Is lit expr enough? What about bin expr?

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -978,6 +978,38 @@ class RsMetaItemStub(
     }
 }
 
+class RsLitExprStub(
+    parent: StubElement<*>?, elementType: IStubElementType<*, *>,
+    val integerLiteralValue: String?
+) : StubBase<RsLitExpr>(parent, elementType) {
+    object Type : RsStubElementType<RsLitExprStub, RsLitExpr>("LIT_EXPR") {
+
+        override fun shouldCreateStub(node: ASTNode): Boolean {
```
vlad20012: Why `createStubIfParentIsStub(node)` isn't work here?

Undin:
Our grammar is supposed not to generate a direct object of `RsExprImpl` type, isn't it?

My problem: we should make `RsExpr` to extend `StubBasedPsiElement` (and correspondingly
`RsExprImpl` should implement `StubBasedPsiElement` somehow) because otherwise all its child classes can't implement `StubBasedPsiElement`. 

On the one hand, `RsExprImpl` has `getElementType` method and it should return `IStubElementType` (or it fails at runtime while call) and if there are real direct objects of `RsExprImpl` we must generate implementation of `IStubElementType` for `RsExpr` (test fails because there isn't such implementation). On the other hand, if we add corresponding code into `org.rust.lang.core.stubs.StubImplementationsKt.factory`, stubs for expr PSI elements will be created incorrectly.
 
So direct objects of `RsExprImpl` should not exist and there is some mistake in grammar (test fails at call of `RsExprImpl.getElementType`) or I have no idea how to create stubs for expressions.

Undin:
@vlad20012 @matklad any tips?

matklad:
Yeah, I think that ideally `RsExprImpl` should be an `abstract` class. However it's also true that inheritance, Grammar Kit, and stubs together don't mesh up really well.

The trick we used for TypeReferences (when they used `extends` in the grammar) was to create a single stub type for all various classes.

Undin:
My current implementation uses single stub type for all expressions, but current grammar allows to generate direct objects of `RsExprImpl` somehow and current tests fail because of it. It is bug, isn't it?

Undin:
Hm, no, I use different stub class, sorry
But it seems it's not a problem

Undin:
Hmm, no, I use different stub classes, sorry
But it seems that it isn't problem in this case

matklad:
>but current grammar generates direct objects of RsExprImpl somehow

Yeah, I *think* this should not happen (this class should be abstract), but I am not sure what's causes this.

Undin:
Ok, thanks:)
I'll try to find out how it happens

Undin:
Hm, looks like that
```rust
#[attr]
{}
```
in 
```rust
fn main() {
    #[attr]
    {}
}
```
generates direct object of `RsExprImpl`

Undin:
@matklad thanks for #2039! `RsCompilerSourcesPerformance` test works!

Undin:
Finally, it seems it works! 🎉

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -534,27 +535,27 @@ private class RsFnInferenceContext(
 
     private fun tryCoerce(inferred: Ty, expected: Ty): Boolean {
         return when {
-        // Coerce array to slice
+            // Coerce array to slice
```
vlad20012: Hmm, intellij auto-format thinks differently about this

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1255,7 +1260,7 @@ private class RsFnInferenceContext(
                 ?: expr.initializer?.inferType()
                 ?: return TySlice(TyUnknown)
             expr.sizeExpr?.inferType(TyInteger(TyInteger.Kind.usize))
-            val size = calculateArraySize(expr.sizeExpr)
+            val size = calculateArraySize(expr.sizeExpr) { resolvePathExpr(it).singleOrNull()?.element }
```
vlad20012: Here we don't need any additional `resolvePathExpr`. We can just lookup already resolved path:
```rust
calculateArraySize(expr.sizeExpr) { ctx.resolvedPaths[it] }
```

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
Undin: It's weird but I've reverted this change

vlad20012:
I was very worried about the assumption that extra stubs (e.g. full function bodies) may be created as a side-effect of thid PR. So I write some tests that check stub existence. [Try it](https://github.com/vlad20012/intellij-rust/commit/9bf0119e29ce072c0a6ed271d2ffb0f4fa927dcf) (don't merge, just copy-paste the code). I've tried it and it seems like even expressions inside statements are stubbed =(
Or maybe my tests are broken.
c.c. @matklad

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1038,14 +1028,96 @@ class RsMetaItemStub(
...
+}
+
+class RsLitExprStub(
+    parent: StubElement<*>?, elementType: IStubElementType<*, *>,
+    val type: RsStubLiteralType?,
+    val integerLiteralValue: String?
+) : RsPlaceholderStub(parent, elementType) {
+    object Type : RsStubElementType<RsLitExprStub, RsLitExpr>("LIT_EXPR") {
+
+        override fun serialize(stub: RsLitExprStub, dataStream: StubOutputStream) {
```
vlad20012: :thinking: 
```kotlin
 override fun shouldCreateStub(node: ASTNode): Boolean = createStubIfParentIsStub(node)
```

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1038,14 +1028,96 @@ class RsMetaItemStub(
...
+    }
+}
+
+class RsLitExprStub(
+    parent: StubElement<*>?, elementType: IStubElementType<*, *>,
+    val type: RsStubLiteralType?,
+    val integerLiteralValue: String?
+) : RsPlaceholderStub(parent, elementType) {
+    object Type : RsStubElementType<RsLitExprStub, RsLitExpr>("LIT_EXPR") {
+
```
vlad20012: :thinking: 
```kotlin
 override fun shouldCreateStub(node: ASTNode): Boolean = createStubIfParentIsStub(node)
```

vlad20012: :thinking: 
```kotlin
override fun shouldCreateStub(node: ASTNode): Boolean = createStubIfParentIsStub(node)
```

vlad20012: Yep, it fixes most cases except function tail expr

vlad20012: :thinking: :thinking: 
```kotlin
override fun shouldCreateStub(node: ASTNode): Boolean = createStubIfParentIsStub(node) &&
            node.psi.parent.parent !is RsFunction
```

Undin: 🤦
I was completely sure that `RsLitExprStub.Type` is inheritor of `RsPlaceholderStub.Type` and it's already used `createStubIfParentIsStub`
Sorry, my fault

vlad20012:
```rust
        fn foo() -> i32 { 2 + 2 }
                          //^
```
This test is still failing

vlad20012:
Anyway, I think I should finally merge it. 4 month, wow :)
I'll add stub tests and fix tail exprs right after the merge.
bors r+

PR #1833, opened 2017-10-08
T: stub only type inference

Undin:


src/test/kotlin/org/rust/lang/core/type/RsTypificationTestBase.kt
```diff
@@ -5,14 +5,31 @@
...
         InlineFile(code)
+        check(description)
+    }
+
+    protected fun stubOnlyTypeInfer(@Language("Rust") code: String, description: String = "") {
+        val testProject = fileTreeFromText(code)
+            .createAndOpenFileWithCaretMarker()
+
+        checkAstNotLoaded(VirtualFileFilter { file ->
+            !file.path.endsWith(testProject.fileWithCaret)
```
matklad: Have you verified that this assert really fires if type inferences touches AST?

Undin: At least, my current implementation of const expr evaluation fails:)

matklad: :)

matklad:
bors r+

An excellent change @Undin !

PR #1840, opened 2017-10-09
INT: Add arms before right brace

### LINKED ISSUE #1831: Intention "Fill match arms" deletes comments inside `match`-expression
# Expected Behaviour
The intention "Fill match arms" should keep the comments inside the `match`-expression, either below or above the generated content. Running it on the following code snippet...
```rust
enum LineStyle {
    Solid,
    Dashed,
    Dotted,
    // ...
}
let ls = LineStyle::Dashed;
match ls {
    // check for line style...
    // LineStyle::Dashed => println!("dashed"),
}
```
...should produce something like the following:
```rust
// ...
match ls {
    // check for line style...
    // LineStyle::Dashed => println!("dashed"),
    LineStyle::Solid => {},
    LineStyle::Dashed => {},
    LineStyle::Dotted => {},
}
```
# Actual Behaviour
Using the intention results in the comments being deleted:
```rust
// ...
match ls {
    LineStyle::Solid => {},
    LineStyle::Dashed => {},
    LineStyle::Dotted => {},
}
```

# Steps to Reproduce the Problem
1. Create a new Rust binary project using IntelliJ IDEA.
2. Paste the snippet from the first section of this issue into the generated `main`-function.
3. Place the caret inside of the match expression.
4. Press `Alt`+`Return` (Show Intention Actions).
5. Select "Fill match arms" and press `Return`.

# Specifications
- IntelliJ IDEA: Ultimate 2017.2 (Student License). Build #IU-172.4343.14 built on September 26, 2017
- Rust plugin: 0.1.0.2070 from October 2, 2017
- OS: ArchLinux x64, Kernel: 4.13.4-1-ARCH
- JRE: 1.8.0_152-release-915-b12 amd64
- JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o

farodin91:
Fixes #1831 
@vlad20012 r?

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/FillMatchArmsIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/FillMatchArmsIntention.kt
@@ -35,7 +35,16 @@ class FillMatchArmsIntention : RsElementBaseIntentionAction<FillMatchArmsIntenti
     override fun invoke(project: Project, editor: Editor, ctx: Context) {
         val (expr, name, variants) = ctx
         var body = RsPsiFactory(project).createMatchBody(name, variants)
-        body = (expr.matchBody?.replace(body) ?: expr.addAfter(body, expr.expr)) as RsMatchBody
+        val matchBody = expr.matchBody
+        if (matchBody != null) {
+            val rbrace = matchBody.rbrace
+            for (arm in body.matchArmList){
+                matchBody.addBefore(arm, rbrace)
+            }
+            body = matchBody
+        } else {
+            body = expr.addAfter(body, expr.expr) as RsMatchBody
+        }
 
         val lbraceOffset = (body.matchArmList.firstOrNull()?.expr as? RsBlockExpr)
             ?.block?.lbrace?.textOffset ?: return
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/intentions/FillMatchArmsIntention.kt
```diff
@@ -35,7 +35,16 @@ class FillMatchArmsIntention : RsElementBaseIntentionAction<FillMatchArmsIntenti
     override fun invoke(project: Project, editor: Editor, ctx: Context) {
         val (expr, name, variants) = ctx
         var body = RsPsiFactory(project).createMatchBody(name, variants)
-        body = (expr.matchBody?.replace(body) ?: expr.addAfter(body, expr.expr)) as RsMatchBody
+        val matchBody = expr.matchBody
+        if (matchBody != null) {
+            val rbrace = matchBody.rbrace
+            for (arm in body.matchArmList){
+                matchBody.addBefore(arm, rbrace)
```
vlad20012: Looks like like we could use `matchBody.addRangeBefore(body.firstChild, body.lastChild, rbrace)` here (but then needs to check the corner case where enum body is empty)

vlad20012:
bors r+
Thanks!

PR #1846, opened 2017-10-10
TYPE: Extend extract to support return parameter

### LINKED ISSUE #1774: Extracting a function should create the necessary parameters.
It would be nice if when extracting a function from code all the variables used by that range of code became parameters to the new function.

For example:

```rust
fn foo(x: u32) -> u32 {
   x *= 2;
   x += 1;
   x
}
```
became
```rust

fn foo(x: u32) -> u32 {
    x *= 2;
    new_fn(x)
}

fn new_fn(x: u32) -> u32 {
   x += 1
   x
}
```

farodin91:
Partial fixes #1774 
@vlad20012 r?

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
@@ -7,15 +7,22 @@ package org.rust.lang.refactoring.extractFunction
 
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.search.LocalSearchScope
+import com.intellij.psi.search.searches.ReferencesSearch
 import org.rust.ide.utils.findStatementsInRange
 import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsPatBinding
+import org.rust.lang.core.psi.ext.descendantsOfType
 import org.rust.lang.core.psi.ext.isAssocFn
 import org.rust.lang.core.psi.ext.owner
 import org.rust.lang.core.psi.ext.parentOfType
+import org.rust.lang.core.types.type
 
 class RsExtractFunctionConfig private constructor(
     val containingFunction: RsFunction,
     val elements: List<PsiElement>,
+    val returnType: String? = null,
+    val returnBindingName: String? = null,
     var name: String = "",
     var visibilityLevelPublic: Boolean = false,
     val needsSelf: Boolean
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -31,11 +38,24 @@ class RsExtractFunctionConfig private constructor(
             // check element should be a part of one block
             val fn = first.parentOfType<RsFunction>() ?: return null
             if (fn != last.parentOfType<RsFunction>()) return null
+
+            val letBindings = fn.descendantsOfType<RsPatBinding>().filter { it.textOffset <= end }
+
+            //val beforeBindings = letBindings.filter { it.textOffset < start }
```
matklad: This probably should be deleted?

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionDialog.kt
```diff
@@ -6,47 +6,76 @@
...
     }
     visibilityBox.selectedItem = "Private"
 
-    with(contentPanel) {
-        add(visibilityLabel)
-        add(functionNameLabel)
-        add(visibilityBox)
-        add(functionNameField)
+    val signatureComponent = RsSignatureComponent("fn test() {}", project)
+    signatureComponent.minimumSize = Dimension(300,30)
```
matklad: Formatting is off: there should be a space after `,`.

src/test/kotlin/org/rust/lang/refactoring/RsExtractFunctionTest.kt
```diff
@@ -33,6 +33,48 @@ class RsExtractFunctionTest : RsTestBase() {
...
+    fun `test extract take last possible return binding`() = doTest("""
+            fn main() {
+                <selection>let test2 = 10i32;
+                let test = 10i32;</selection>
+                println!("{}", test);
+                println!("{}", test2);
+            }
+        """, """
+            fn main() {
+                let test = test();
```
matklad: Yeah, ideally, we should be able to do something like `let (test, test2) = test()`...

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
farodin91: This is prep for function parameters.

farodin91:
@matklad all resolved

matklad:
bors r+

Thanks @farodin91 !

PR #1849, opened 2017-10-11
INSP: Don't warn about conventions when implementing traits

### LINKED ISSUE #1715: Don't warn about conventions when implementing traits
If `SomeTrait` is in a crate I don't control, I can't fix its bad conventions, so don't warn me about them when I implement it.

```rust
extern crate not_my_crate;

use not_my_crate::SomeTrait;

struct MyStruct;

impl SomeTrait for MyStruct {
  // This will warn as follows
  // methods called `as_*` usually take self by reference or self by mutable reference; consider choosing a less ambiguous name
  // This isn't my choice; don't warn me about it
  fn as_i64(self) -> i64 {
    unimplemented!();
  }
}
```

farodin91:
Fixes #1715 
@vlad20012 r?

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
@@ -9,8 +9,10 @@ import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.RsVisitor
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.ty.TyStructOrEnumBase
 import org.rust.lang.core.types.type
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
```diff
@@ -19,12 +19,13 @@ class RsSelfConventionInspection : RsLocalInspectionTool() {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
         object : RsVisitor() {
             override fun visitFunction(m: RsFunction) {
-                // Should this handle traits as well perhaps?
-                if (m.owner !is RsFunctionOwner.Impl) return
+                val owner = m.owner
+                if (!owner.isImplOrTrait) return
+                if (owner.isTraitImpl) return
```
matklad: I think it would be more readable if all owner checks were consolidated in a single if, like `if (!(owner.isInherentImpl || owner.isTrait)) return`

farodin91:
Resolved

src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
```diff
@@ -19,12 +19,12 @@ class RsSelfConventionInspection : RsLocalInspectionTool() {
...
             override fun visitFunction(m: RsFunction) {
-                // Should this handle traits as well perhaps?
-                if (m.owner !is RsFunctionOwner.Impl) return
+                val owner = m.owner
+                if (!(owner.isInherentImpl || owner.isTrait)) return
 
                 val convention = SELF_CONVENTIONS.find { m.identifier.text.startsWith(it.prefix) } ?: return
                 if (m.selfType in convention.selfTypes) return
-                if (m.selfType == SelfType.SELF && m.isOwnerCopyable()) return
+                if (owner.isInherentImpl && m.selfType == SelfType.SELF && m.isOwnerCopyable()) return
```
matklad: Why do we need `owner.isInherentImpl` here?

farodin91: isOwnerCopy works only on Impl

farodin91:
Resolved!

src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
```diff
@@ -30,9 +32,19 @@ class RsSelfConventionInspection : RsLocalInspectionTool() {
         }
 
     private fun RsFunction.isOwnerCopyable(): Boolean {
-        val implBlock = parentOfType<RsImplItem>() ?: return false
-        val owner = implBlock.typeReference?.type as? TyStructOrEnumBase ?: return false
-        return owner.item.queryAttributes.hasAttributeWithArg("derive", "Copy")
+        val traitOrImpl = parentOfType<RsTraitOrImpl>() ?: return false
```
matklad: Let's use `RsFunction#owner` here: `parentOfType` might be wrong in cases like this

```
impl X {
  fn foo() { fn bar() { /* no owner here, but there's a parent of type*/ } }
}
```

I *think* that in this particular case it's OK, because we check that the owner is impl or trait, but this relise on some interprocedural reasoning, which makes code harder to read.

Could we switch on `m.owner` directly and fold `copyble` and `owner.isInherentImpl || owner.isTrait` checks into one?

farodin91: owner doesn't use `parentOfType`

farodin91: Look at the wrong place

matklad: >owner doesn't use parentOfType

Yes, and it's the right thing to do because `parentOfType` might give wrong results. So it's preferable to use `owner` if possible, and I think we can use `owner` here?

farodin91: I try to improve the solution

matklad:
bors r+

Thanks!

PR #1855, opened 2017-10-12
HINT: Remove redundant local variable type hint

### LINKED ISSUE #1850: Local variable type hint is sometimes redundant
Local variable type hints are generally useful, but sometimes they are reundant, for example:

```
struct Point {
  x: i32,
  y: i32
};

fn foo() {
  let x = Point{ x: 42, y: 42 };
}
```

Here, showing a local variable hint for `x` is redundant, because the type name is right there as the first token of the initializer expression. I think in a case like this, the editor could be smarter and not show the hint.

farodin91:
Fixes #1850 
@matklad r?

### DIFF
--- a/src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
+++ b/src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
@@ -11,6 +11,7 @@ import org.intellij.lang.annotations.Language
 import org.rust.fileTreeFromText
 import org.rust.lang.RsTestBase
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.descendantsOfType
 
 
 class RsInlayParameterHintsProviderTest : RsTestBase() {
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
```diff
@@ -60,7 +60,14 @@ class RsInlayParameterHintsProviderTest : RsTestBase() {
         fn main() {
             let s/*caret*/ = S;
         }  //^
-    """, ": S", 0)
+    """, ": S", 0, smart = false)
+
+    fun `test let should not show a hint directly after initial struct smart hint`() = checkNoHint<RsLetDecl>("""
```
matklad: Let's add a couple of more tests here, where the hints should not be shown?

I am thinking about

```Rust
struct TupleStruct(f32)
struct BracedStruct { f: f32 }
enum E {
   C, B { f: f32 }, T(f32)
}

fn main() {
    let no_hint = TupleStruct(1.0);
    let no_hint = BracedStruct { f: 1.0 };
    let no_hint = E::C;
    let no_hint = E::B { f: 1.0 };
    let no_hint = E::T(1.0);
}
```

farodin91:
Resolved

matklad:
bors r+

Thanks!

PR #1862, opened 2017-10-15
Resolve #1587: It is not possible to debug applications with command line parameters

### LINKED ISSUE #1587: It is not possible to debug applications with command line parameters
I create a Run configuration using parameters. When I Run it, it correctly executes a command like:
```
cargo run --color=always --bin myProg -- client peperulo
```
However, when I launch it with Debug, it executes other cargo line like:
```
cargo build --color=always --bin myProg -- client peperulo
```
and Cargo reports an error: 
```
error: Invalid arguments.

Usage:
    cargo build [options]
```

It works ok if I remove my command lines after --

bgard6977:
I was unable to debug a program in CLion because it was not possible in the Run Configuration dialog box to set parameters to pass to the application being debugged.

I added a new "Run Arguments" field to the dialog box, plumbed it through to both run and debug mode, and tested as much as I could with IntelliJ (I'm not sure how to debug it in CLion, but it appears to launch with the correct parameters...)

screenshot at 2017-10-15 17 23 01

### DIFF
--- a/src/test/kotlin/org/rust/cargo/toolchain/CargoTest.kt
+++ b/src/test/kotlin/org/rust/cargo/toolchain/CargoTest.kt
@@ -8,10 +8,24 @@ package org.rust.cargo.toolchain
 import com.intellij.execution.configurations.GeneralCommandLine
 import com.intellij.openapi.util.SystemInfo
 import com.intellij.util.net.HttpConfigurable
+import org.junit.Test
 import org.rust.lang.RsTestBase
+import org.rust.openapiext.pathAsPath
 import java.nio.file.Paths
 
 class CargoTest : RsTestBase() {
+
+    fun `test run arguments preserved`() = checkCommandLine(
+        cargo
+            .toColoredCommandLine(CargoCommandLine("run", listOf("--bin", "parity", "--", "--prune", "archive")))
+            .withCharset(Charsets.UTF_8),
+        "cmd: /usr/bin/cargo run --color=always --bin parity -- --prune archive\n" +
+            "env: RUSTC=/usr/bin/rustc, RUST_BACKTRACE=full, TERM=ansi"
+        , """
+            cmd: C:/usr/bin/cargo.exe run --bin parity -- --prune archive
+            env: RUSTC=C:/usr/bin/rustc.exe, RUST_BACKTRACE=full, TERM=ansi
+    """)
+
     fun `test basic command`() = checkCommandLine(
         cargo.toGeneralCommandLine(CargoCommandLine("test", listOf("--all"))), """
         cmd: /usr/bin/cargo test --all -- --nocapture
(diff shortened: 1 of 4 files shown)


matklad:
The implementation looks correct to me!

 However, am not really keen on adding more fields to the run configuration form: I think there are benefits in keeping it as close to the raw Cargo command line as possible. 

What do you think about parsing the `command` field instead? I think it could work by calling `ParametersListUtil.parse(runArguments)` and then taking what's after the `--` arg. 

Please note that we need an electronic CLA for the pull requests: https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#contributing

bgard6977:
@matklad I updated it to fulfill the same functional requirements, but using a parse/join approach, and verified it works in CLion with a real app. I also submitted the CLA, and rebased onto master. Please let me know if there's any other changes I should make to get this merged.

Thanks,

Brent

src/main/kotlin/org/rust/cargo/runconfig/CargoRunState.kt
```diff
@@ -42,6 +42,8 @@ class CargoRunState(
             // Even though default system encoding is usually not UTF-8 on windows,
             // most Rust programs are UTF-8 only.
             .withCharset(Charsets.UTF_8)
+            .withParameters("--") // Tell cargo remaining args are for the debug target
+            .withParameters(commandLine.getRunArguments())
```
matklad: I think these two lines might not be needed here? All arguments should be added as part of `additionalArguments` inside `toColoredCommandLine`.

bgard6977: - Removed those lines
- Wrote a unit test
- Verified in CLion (debug and run)
- Rebased
- Pushed

src/test/kotlin/org/rust/cargo/toolchain/CargoTest.kt
```diff
@@ -8,10 +8,22 @@ package org.rust.cargo.toolchain
...
 class CargoTest : RsTestBase() {
+
+    fun `test run arguments preserved`() = checkCommandLine(
+        cargo
+            .toColoredCommandLine(CargoCommandLine("run", listOf("--bin", "parity", "--", "--prune", "archive")))
+            .withCharset(Charsets.UTF_8),
+        "cmd: /usr/bin/cargo run --color=always --bin parity -- --prune archive\n" +
+            "env: RUSTC=/usr/bin/rustc, RUST_BACKTRACE=full, TERM=ansi",
+        ""
+    )
```
matklad: Excellent, thanks!

I am afraid this'll fail on windows though (that's what the second argument means). See the tests below to find out how this second string should look like.

Oh, and while we are at it, perhaps it would be prudent to write a couple of unit-tests for `getRunArguments` and `getBuildArguments` as well? Nothing fancy, just constructing a `CargoCommandLine` with `listOf("--bin", "parity", "--", "--prune", "archive")` and checking result exactly.

matklad:
bors try

Let's check the window build!

matklad:
bors: try

:)

matklad:
bors: try

matklad:
Ok, let's try it in this way:

bors: r+

bgard6977:
bors: r+

bgard6977:
@matklad Updated with the requested changes. I guessed at the windows syntax, as I don't have an environment readily available. Hopefully I guess right. Could you kick off bors when you have a moment? Thanks!

matklad:
bors r+

Thanks @bgard6977 !

PR #1863, opened 2017-10-16
Add a PR template which mentions CLA

matklad:
We really need to make a real bot some day, but a PR template should be a good stop-gap solution.

r? @vlad20012


vlad20012:
Hmm, nice solution :)

bors r+

PR #1866, opened 2017-10-16
Run

matklad:
Cleanup our runconfigs a bit and make sure that `Build Project` (<kbd>Ctrl+F9</kbd>) works with several projects. 

r? @vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/CargoCommandLine.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/CargoCommandLine.kt
@@ -6,6 +6,7 @@
 package org.rust.cargo.toolchain
 
 import com.intellij.execution.configuration.EnvironmentVariablesData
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/cargo/toolchain/CargoCommandLine.kt
```diff
@@ -16,4 +17,21 @@ data class CargoCommandLine(
     val workingDirectory: Path? = null,
     val environmentVariables: EnvironmentVariablesData = EnvironmentVariablesData.DEFAULT,
     val nocapture: Boolean = true
-)
+) {
+    companion object {
+        fun forProject(
```
matklad: Found a better solution for this :) 

https://github.com/intellij-rust/intellij-rust/commit/559218c81dcce533b5ebc09d4615663f019ad93c#diff-96ed2d1e44784b834de08e4c720d9f9bL21

PR #1869, opened 2017-10-16
Cargo comp

matklad:
Add `--manifest-path` to completion.


vlad20012:
Hah, `test complete run args` test failed :) Now there should be `--manifest-path` in completion.

matklad:
bors r=vlad20012

PR #1872, opened 2017-10-16
Fix project settings serialization

### LINKED ISSUE #1870: Toolchain settings forgotten on project load
After the latest update (0.2.0.2073), no matter what I do I get the following error when opening a project:

```
Cargo project update failed:
	Can't update Cargo project, no Rust toolchain
```

Clicking "Setup toolchain", then "Ok" fixes the problem, but on the next reload the plugin loses the toolchain information again.

matklad:
closes #1870

Ideas serialization can't deserialize `java.nio.Path`.

r? @vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/utils.kt
+++ b/src/main/kotlin/org/rust/openapiext/utils.kt
@@ -24,6 +24,7 @@ import com.intellij.psi.stubs.StubIndex
 import com.intellij.psi.stubs.StubIndexKey
 import com.intellij.util.io.systemIndependentPath
 import org.jdom.Element
+import org.jdom.input.SAXBuilder
 import java.nio.file.Path
 import java.nio.file.Paths
 
@@ -42,6 +43,7 @@ fun checkWriteAccessAllowed() {
     }
 }
 
+@Suppress("unused")
 fun checkReadAccessAllowed() {
     check(ApplicationManager.getApplication().isReadAccessAllowed) {
         "Needs read action"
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
```diff
@@ -90,13 +90,16 @@ interface CargoProject {
 }
 
 fun guessAndSetupRustProject(project: Project): Boolean {
+    println("project.rustSettings.toolchain = ${project.rustSettings.toolchain}")
```
vlad20012: Hmmmmmmm ?

matklad: 😊 fixed

matklad:
bors try

matklad:
bors try

src/main/kotlin/org/rust/openapiext/utils.kt
```diff
@@ -40,6 +43,7 @@ fun checkWriteAccessAllowed() {
     }
 }
 
+@Suppress("unused")
```
mkaput: Why?

matklad:
Merge manually so as to make a new alpha release :)

src/main/kotlin/org/rust/openapiext/utils.kt
matklad: This function is indeed unused, so we probably can just remove it...

But perhaps it should be used somewhere? :) I think all our services are now thread safe, so we don't actually need to enforce  usages of read actions, but it be nice to audit them never the less :)

mkaput: I think it is nice to leave comments every time such annotation is used :)

I thought it might be used via reflection or sth

PR #1881, opened 2017-10-19
[COMP] added semicolon completion when completing a function in use statement

### LINKED ISSUE #1856: Completion of function in "use" declaration could insert semicolon
When performing completion of a function in a "use" declaration, we know we'll want a semicolon afterwards, so the completion might as well insert it.

mhaessig:
Added the completion of semicolons after a function in a use statement. 

Fixes #1856

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -12,6 +12,7 @@ import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.openapi.editor.EditorModificationUtil
 import com.intellij.psi.util.PsiTreeUtil
+import org.jetbrains.debugger.createVariableContext
 import org.rust.ide.icons.RsIcons
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
@@ -30,9 +31,8 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
             base.withTailText("::")
                 .withInsertHandler({ ctx, _ ->
                     val offset = ctx.tailOffset - 1
-                    val inUse = PsiTreeUtil.findElementOfClassAtOffset(ctx.file, offset, RsUseGlobList::class.java, false) != null
                     val inSelfParam = PsiTreeUtil.findElementOfClassAtOffset(ctx.file, offset, RsSelfParameter::class.java, false) != null
-                    if (!(inUse || inSelfParam)) {
+                    if (!(ctx.isInGlob || inSelfParam)) {
                         ctx.addSuffix("::")
                     }
                 })
@@ -48,7 +48,15 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
             .withTailText(element.valueParameterList?.text?.replace("\\s+".toRegex(), " ") ?: "()")
             .appendTailText(element.extraTailText, true)
             .withInsertHandler handler@ { context: InsertionContext, _: LookupElement ->
-                if (context.isInUseBlock) return@handler
+                val curUseItem = context.getUseItem
+                if (curUseItem != null) {
+                    val hasSemicolon = curUseItem.lastChild!!.elementType == RsElementTypes.SEMICOLON
+                    if (!(hasSemicolon || context.isInGlob)) {
+                        context.addSuffix(";")
+                    }
+                    return@handler
+                }
+
                 if (!context.alreadyHasCallParens) {
                     context.document.insertString(context.selectionEndOffset, "()")
                 }
@@ -76,7 +84,7 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
                 else -> ""
             })
             .withInsertHandler handler@ { context, _ ->
-                if (context.isInUseBlock) return@handler
+                if (context.getUseItem != null) return@handler
                 val (text, shift) = when {
                     element.tupleFields != null -> Pair("()", 1)
                     element.blockFields != null -> Pair(" {}", 2)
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt
```diff
@@ -150,6 +150,14 @@ class RsCompletionTest : RsCompletionTestBase() {
         use self::foo::quux/*caret*/;
     """)
```
matklad: Let's add another test for glob imports:

    use self::{f}


Here, the semicolon should not be added, but nothing should explode as well

mhaessig: So the test would be 
```kotlin
fun `test use glob no semicolon`() = doSingleCompletion("""
        mod foo { pub fn quux() {} }
        use self::{f/*caret*/}
    """, """
        mod foo { pub fn quux() {} }
        use self::{f/*caret*/}
    """)
```
?

matklad: Rater

```

    fun `test some illuminating name`() = doSingleCompletion("""
        use self::{f/*caret*/}
        fn foo() {}
    """, """
        use self::{foo/*caret*/}
        fn foo() {}
    """)
```

mhaessig: I couldn't use your suggestion because it is not a single completion. I just put the function in a module.

matklad:
LGTM, but looks like CI is failing legitimately? 

I think we can extend this logic to structs, traits, const, statics (but not enums, bc you can import individual variants). Wanna send a follow up PR once this one is merged? :)

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -48,7 +47,16 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
             .withTailText(element.valueParameterList?.text?.replace("\\s+".toRegex(), " ") ?: "()")
             .appendTailText(element.extraTailText, true)
             .withInsertHandler handler@ { context: InsertionContext, _: LookupElement ->
-                if (context.isInUseBlock) return@handler
+                if (context.isInUseBlock) {
+                    val offset = context.tailOffset - 1
+                    val useItem = PsiTreeUtil.findElementOfClassAtOffset(context.file, offset, RsUseItem::class.java, false)
+                    val hasSemicolon = useItem!!.lastChild!!.elementType != RsElementTypes.SEMICOLON
```
matklad: Hm, now that I think of it, could we get rid of those `!!` in `useItem!!`? What if `isInUseBlock` returned not a `Bool`, but a `RsUseItem?`? So that we can compare it with `null`.

mhaessig: I refactored it to your suggestion.

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -118,8 +124,11 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
 fun LookupElementBuilder.withPriority(priority: Double): LookupElement =
     PrioritizedLookupElement.withPriority(this, priority)
 
-private val InsertionContext.isInUseBlock: Boolean
-    get() = file.findElementAt(startOffset - 1)!!.parentOfType<RsUseItem>() != null
+private val InsertionContext.getUseItem: RsUseItem
+    get() = PsiTreeUtil.findElementOfClassAtOffset(file, tailOffset - 1, RsUseItem::class.java, false)!!
```
matklad: Hm, this might thrown a null pointer exception at runtime. Kotlin's `!!` are very similar to Rust's `.unwraps()`: can help to get the job done, but might cause your program to die in flames in production. 

The safer way would be:

```kotlin
                val useItem = context.getUseItem()
                if (useItem != null) {
                    // Kotlin nows that useItem is not null here, so it does not require `!!` after `useItem`.
                    val hasSemicolon = useItem.lastChild!!.elementType == RsElementTypes.SEMICOLON
                    if (!(hasSemicolon || context.isInGlob)) {
                        context.addSuffix(";")
                    }
                    return@handler
}

...

private fun InsertionContext.getUseItem(): RsUseItem
    = PsiTreeUtil.findElementOfClassAtOffset(file, tailOffset - 1, RsUseItem::class.java, false)
```

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -48,7 +47,14 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
             .withTailText(element.valueParameterList?.text?.replace("\\s+".toRegex(), " ") ?: "()")
             .appendTailText(element.extraTailText, true)
             .withInsertHandler handler@ { context: InsertionContext, _: LookupElement ->
-                if (context.isInUseBlock) return@handler
+                if (context.getUseItem != null) {
+                    val hasSemicolon = context.getUseItem?.lastChild!!.elementType == RsElementTypes.SEMICOLON
```
matklad: This question mark (`context.getUseItem?`) may be avoided if you save `context.getUseItem` to local variable. It's also good for performance, because you won't be traversing PSI two times.

Here are some docs about this "control flow sensitive typing" feature of Kotlin langauge:

https://kotlinlang.org/docs/reference/null-safety.html#checking-for-null-in-conditions

mhaessig: I incorporated your suggestion in the last commit.

Once again RTFM holds true.

matklad:
bors r+

Thanks!

mhaessig:
I'll gladly send a follow up PR. Should I file a new issue?

matklad:
@mhaessig yeah, that'd be useful!

While we are at it, perhaps `PsiTreeUtil.findElementOfClassAtOffset(file, tailOffset - 1, RsUseItem::class.java, false)` can be extracted into a helper function (see how `parentOfType` is impelented via inlined lambdas & reified type parameters).

PR #1883, opened 2017-10-19
Add `isCopy(Ty)` method

matklad:
r? @vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
@@ -86,7 +86,7 @@ private val RsFunction.selfType: SelfType
 data class SelfConvention(val prefix: String, val selfTypes: Collection<SelfType>)
 
 private fun ProblemsHolder.registerProblem(element: PsiElement, convention: SelfConvention) {
-    val selfTypes = convention.selfTypes.map { it.description }.joinToString(" or ")
+    val selfTypes = convention.selfTypes.joinToString(" or ") { it.description }
 
     val description = "methods called `${convention.prefix}*` usually take $selfTypes; " +
         "consider choosing a less ambiguous name"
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
```diff
@@ -13,49 +13,53 @@ import org.rust.lang.core.psi.RsTraitItem
...
     NO_SELF("no self"),
-    SELF("self by value"),
-    REF_SELF("self by reference"),
-    REF_MUT_SELF("self by mutable reference");
+    BY_VAL("self by value"),
+    BY_REF("self by reference"),
+    BY_MUT_REF("self by mutable reference");
 }
 
 private val RsTraitOrImpl.traits: Sequence<RsTraitItem>
```
farodin91: You can remove this function and make supertraits private again.

matklad: Excellent, thanks @farodin91 !

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -55,6 +56,9 @@ class ImplLookup(private val project: Project, private val items: StdKnownItems)
     val fnOutputParam by lazy(NONE) {
         RsLangItemIndex.findLangItem(project, "fn_once")?.let { findFreshAssociatedType(it, "Output") }
     }
+    private val copyTrait: RsTraitItem? by lazy(NONE) {
```
farodin91: Is it possbile to make this more generic?

matklad: Hm, could you elaborate a bit?

farodin91: I would change it to a function, than you say copy or debug, for example: driveTrait("Debug")...

matklad: Yeah, that makes sense, and we should definitely do it once we handle more then one derived trait specially. But for a single use case, I usually prefer not to create an abstraction upfront.

farodin91: Okay

PR #1894, opened 2017-10-22
RES: resolve binary operators

### LINKED ISSUE #1892: Treat overloaded operators as methods (QuickRef, Ctrl-Hover)
`Ctrl+Q`, hovering with `Ctrl` and `Ctrl+B` (go to definition/uses) should work on overloaded operators as well. (CLion does this quite good)
(I'm not sure what `Refactor/Extract type parameter` does, if anything)
And finally show the operators themselves (not only the corresponding method) as suggestion:
image
(CLion, again)

related: https://github.com/intellij-rust/intellij-rust/issues/1890

Undin:
Closes #1892


mkaput:
formatter and parser tests failed

src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt
```diff
@@ -138,6 +138,7 @@ fun createSpacingBuilder(commonSettings: CommonCodeStyleSettings, rustSettings:
 
         //== rules with very large area of application
         .around(NO_SPACE_AROUND_OPS).spaces(0)
+        .around(BINARY_OP).spaces(1)
```
Undin: I'm not very sure that it's the best fix but tests work:)

mkaput: I think this rule might duplicate the rule below

Undin: Yeah, but we can't just exclude `RS_BINARY_OPS` tokens from `SPACE_AROUND_OPS` because, for example, `=` and `|` tokens can be part not only binary expressions.
We can exclude all `RS_BINARY_OPS` tokens except `=` and `|` but I'm afraid of that there are other cases which are not in our tests.

Undin: If you know a good way to get rid of this duplication, I will be glad to use it

matklad:
Excellent, thanks @Undin !

bors r+

Issue #1888, opened 2017-10-19
Add semicolon on completion of use item


mhaessig:
When completing a use item with a struct, traits const or static add the semicolon at the end if needed (like #1881).

mhaessig:
@matklad Could we also insert a semicolon after a glob list?

matklad:
Yeah, I think we can make typing `{` after `use foo::` to auto-insert `};`, but this is handled by completely separate machinery.

mhaessig:
So a separate issue?

matklad:
Yep!

Kobzol:
This has been already fixed by https://github.com/intellij-rust/intellij-rust/pull/1901 and https://github.com/intellij-rust/intellij-rust/pull/2142.

PR #1901, opened 2017-10-23
[COMP] Semicolon completion in use statements

mhaessig:
PR for issue #1888. Currently work in progress.

Tasks:
- [X] Refactor `getUseItem` to helper function `getItemOfType`
- [x] Semicolon completion for structs
- [x] Semicolon completion for traits
- [x] Semicolon completion for const
- [x] Semicolon completion for statics

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -48,7 +48,7 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
             .withTailText(element.valueParameterList?.text?.replace("\\s+".toRegex(), " ") ?: "()")
             .appendTailText(element.extraTailText, true)
             .withInsertHandler handler@ { context: InsertionContext, _: LookupElement ->
-                val curUseItem = context.getUseItem
+                val curUseItem = context.getItemOfType<RsUseItem>()
                 if (curUseItem != null) {
                     val hasSemicolon = curUseItem.lastChild!!.elementType == RsElementTypes.SEMICOLON
                     if (!(hasSemicolon || context.isInGlob)) {
@@ -84,7 +84,7 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
                 else -> ""
             })
             .withInsertHandler handler@ { context, _ ->
-                if (context.getUseItem != null) return@handler
+                if (context.getItemOfType<RsUseItem>() != null) return@handler
                 val (text, shift) = when {
                     element.tupleFields != null -> Pair("()", 1)
                     element.blockFields != null -> Pair(" {}", 2)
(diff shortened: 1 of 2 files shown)


mhaessig:
The failed check `./gradlew assemble testClasses -Pkotlin.incremental=false --no-daemon` works on my machine.

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -126,8 +130,19 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
 fun LookupElementBuilder.withPriority(priority: Double): LookupElement =
     PrioritizedLookupElement.withPriority(this, priority)
 
-private val InsertionContext.getUseItem: RsUseItem?
-    get() = PsiTreeUtil.findElementOfClassAtOffset(file, tailOffset - 1, RsUseItem::class.java, false)
+fun appendSemicolon(context: InsertionContext) {
```
farodin91: This should be private

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -126,8 +130,19 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
...
+    if (curUseItem != null) {
+        val hasSemicolon = curUseItem.lastChild!!.elementType == RsElementTypes.SEMICOLON
+        if (!(hasSemicolon || context.isInGlob)) {
+            context.addSuffix(";")
+        }
+    }
+}
+
+
+inline fun <reified T : RsItemElement> InsertionContext.getItemOfType(strict: Boolean = false): T? =
```
farodin91: same here

mhaessig:
You're absolutely . I fixed it in the newest iteration of my last commit.

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -126,8 +130,19 @@ fun createLookupElement(element: RsCompositeElement, scopeName: String): LookupE
...
-    get() = PsiTreeUtil.findElementOfClassAtOffset(file, tailOffset - 1, RsUseItem::class.java, false)
+private fun appendSemicolon(context: InsertionContext) {
+    val curUseItem = context.getItemOfType<RsUseItem>()
+    if (curUseItem != null) {
+        val hasSemicolon = curUseItem.lastChild!!.elementType == RsElementTypes.SEMICOLON
+        if (!(hasSemicolon || context.isInGlob)) {
+            context.addSuffix(";")
+        }
+    }
+}
```
matklad: There's a bit of code duplication between this method and `is RsFunction` branch. Could we somehow find out, how to get rid of it? 

One possible solution I have in mind is to attach a single insertion handler to all our elements, and *inside* that handler determine the context by `tailOffset` and the element to be inserted via a `when` switch. That is, leave only a single `.withInsertionHandler` in this file, and move all inertion handling logic to a separate file. 

What do you think about this?

mhaessig: The reason I left the duplication for functions is because they need the return of `@handler` in order to work.
But I think I see the idea behind your proposed refactoring and I like it. Although I am not quite sure how we can decouple `.withInsertionHandler` from the types while still having the different type- and tail-texts.

matklad: >Although I am not quite sure how we can decouple .withInsertionHandler from the types while still having the different type- and tail-texts.

Hm, I think we need some function which switch over element type and returns an `LookupElementBuilder` with type text, tail text, but without insertion handler. 

Then, we add an insertion handler to that Builder

matklad: Ok, I'll r+ this now, but it'd be great to remove duplication in a separate PR!

PR #1932, opened 2017-10-26
Fix compiler warnings

matklad:


matklad:
@farodin91 there were some shadowing warnings in extract function refactoring, an I've tried to fix them. In general, we try to maintain zero compiler warning policy, but unfortunately we can't enforce it via CI until Kotlin 1.2.

PR #1935, opened 2017-10-26
TYPE: Extract function replace unknown types

### LINKED ISSUE #1933: Extract function refactoring creates syntatically invalid code
Before

```
fn parse_test(call: Call) -> JsResult<JsValue> {
    <selection>let scope = call.scope;
    let test = call.arguments.require(scope, 0)?.check::<JsInteger>()?.value() as usize;
    let callback = call.arguments.require(scope, 1)?.check::<JsFunction>()?;
    let file = FILE.lock().unwrap();
    let file = get_file_or_return_null!(file).clone();</selection>

    struct RenderTask(Arc<File>, usize);
    impl Task for RenderTask {
        type Output = String;
        type Error = ();
        type JsEvent = JsString;

        fn perform(&self) -> Result<String, ()> {
            let mut renderer = renderer();
            let tree = renderer.render_one(&self.0, self.1);
            Ok(tree)
        }

        fn complete<'a, T: Scope<'a>>(self, scope: &'a mut T, result: Result<String, ()>) -> JsResult<JsString> {
            Ok(JsString::new(scope, &result.unwrap()).unwrap())
        }
    }

    RenderTask(file, test).schedule(callback);
    Ok(JsNull::new().upcast())
}
```

After ->

```
fn parse_test(call: Call) -> JsResult<JsValue> {
    let (test, callback, file) = foo();

    struct RenderTask(Arc<File>, usize);
    impl Task for RenderTask {
        type Output = String;
        type Error = ();
        type JsEvent = JsString;

        fn perform(&self) -> Result<String, ()> {
            let mut renderer = renderer();
            let tree = renderer.render_one(&self.0, self.1);
            Ok(tree)
        }

        fn complete<'a, T: Scope<'a>>(self, scope: &'a mut T, result: Result<String, ()>) -> JsResult<JsString> {
            Ok(JsString::new(scope, &result.unwrap()).unwrap())
        }
    }

    RenderTask(file, test).schedule(callback);
    Ok(JsNull::new().upcast())
}

fn foo() ->
```

cc @farodin91

farodin91:
Fixes #1933 
@matklad r?

### DIFF
--- a/src/main/kotlin/org/rust/ide/presentation/Ty.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/Ty.kt
@@ -9,12 +9,21 @@ import org.rust.lang.core.types.ty.*
 
 
 val Ty.shortPresentableText: String get() = render(this, level = 3)
+val Ty.insertionSafeText: String
+    get() = render(this, level = Int.MAX_VALUE, unknown = "_", anonymous = "_", integer = "_", float = "_")
 
 fun tyToString(ty: Ty) = render(ty, Int.MAX_VALUE)
 
-private fun render(ty: Ty, level: Int): String {
+private fun render(
+    ty: Ty,
+    level: Int,
+    unknown: String = "<unknown>",
+    anonymous: String = "<anonymous>",
+    integer: String = "{integer}",
+    float: String = "{float}"
+): String {
     check(level >= 0)
-    if (ty is TyUnknown) return "<unknown>"
+    if (ty is TyUnknown) return unknown
     if (ty is TyPrimitive) {
         return when (ty) {
             is TyBool -> "bool"
@@ -30,8 +39,7 @@ private fun render(ty: Ty, level: Int): String {
 
     if (level == 0) return "_"
 
-    val r = { subTy: Ty -> render(subTy, level - 1) }
-    val anonymous = "<anonymous>"
+    val r = { subTy: Ty -> render(subTy, level - 1, unknown, anonymous, integer, float) }
 
     return when (ty) {
         is TyFunction -> {
@@ -42,8 +50,10 @@ private fun render(ty: Ty, level: Int): String {
         is TySlice -> "[${r(ty.elementType)}]"
 
         is TyTuple -> ty.types.map(r).joinToString(", ", "(", ")")
-        is TyArray -> "[${r(ty.base)}; ${ty.size ?: "<unknown>"}]"
-        is TyReference -> "${if (ty.mutability.isMut) "&mut " else "&"}${render(ty.referenced, level)}"
+        is TyArray -> "[${r(ty.base)}; ${ty.size ?: unknown}]"
+        is TyReference -> "${if (ty.mutability.isMut) "&mut " else "&"}${
+        render(ty.referenced, level, unknown, anonymous, integer, float)
+        }"
         is TyPointer -> "*${if (ty.mutability.isMut) "mut" else "const"} ${r(ty.referenced)}"
         is TyTraitObject -> ty.trait.name ?: anonymous
         is TyTypeParameter -> ty.name ?: anonymous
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -50,9 +50,26 @@ class RsExtractFunctionConfig private constructor(
...
     val signature: String
         get() {
-            var signature = "fn $name(${if (needsSelf) "self" else ""})"
-            if (returnValue != null) {
-                signature += " -> ${returnValue.type}"
+            var signature = "fn $name"
+            val arguments = "(${if (needsSelf) "self" else ""})"
+            signature += if (returnValue != null) {
+                var type = returnValue.type.toString()
+                val unknowns = Regex("<unknown>").findAll(type)
```
matklad: Ouch, that seems gross... Could we check this in a more structured way, rather then relying on a `toString` representation?

farodin91: I tried to find a fix solve for it. Besides, i also don't like the complete signature function.

matklad: There's `org.rust.ide.presentation.TyKt#getShortPresentableText`, and I think we can make `Ty.insertionSafeText` using the same machinery, which would replace `<unknowns>` with something reasonable.

matklad: But in general yeah, I think `ExtractFunction` can use some refactoring and restructuring: this is really complex stuff, with a ton of edge cases, so code must me architectured with care to avoid drowning in the error handing and such.  

We might add some infrastructure for showing error popup with "Sorry, can't extract this function, because it contains this stuff, which we can't handle yet"

farodin91: We can start with making _ and replace this in the future

matklad: Sounds reasonable! But we still need to extract this into a separate method, to make it *easy* to replace in future.

farodin91:
@matklad done

src/main/kotlin/org/rust/ide/presentation/Ty.kt
```diff
@@ -9,12 +9,21 @@ import org.rust.lang.core.types.ty.*
 
 
 val Ty.shortPresentableText: String get() = render(this, level = 3)
+val Ty.insertionSafeText: String
+    get() = render(this, level = 3, unknown = "_", anonymous = "_", integer = "_", float = "_")
```
matklad: Hm, we probably should not use `level` when rendering types for insertion in code?

farodin91: Okay

farodin91:
@matklad Resolved

src/test/kotlin/org/rust/lang/refactoring/RsExtractFunctionTest.kt
```diff
@@ -33,6 +33,71 @@ class RsExtractFunctionTest : RsTestBase() {
...
+            let scope = call.scope;
+            let test = call.arguments.require(scope, 0)?.check::<JsInteger>()?.value() as usize;
+            let callback = call.arguments.require(scope, 1)?.check::<JsFunction>()?;
+            let file = FILE.lock().unwrap();
+            let file = get_file_or_return_null!(file).clone();
+            (test, callback, file)
+        }
+        """,
+        false,
+        "foo")
```
matklad: This test looks way to big for the particular issue at hand, but let's leave it as is as a stress test.

matklad:
bors r+

Thanks!

PR #1938, opened 2017-10-27
COMP: Implement unsafe methods.

### LINKED ISSUE #1895: `unsafe` is not added for "Implement methods"
I have code like this:

```Rust
trait Trait {
    unsafe fn f();
}

impl Trait for () {
}
```

I click Alt+Enter on `impl Trait for ()` and select "Implement methods". IDE then adds stub `fn f() ` instead of `unsafe fn f()`.

ttaomae:
Fixes #1895.

Properly include `unsafe` keyword when implementing trait methods.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -249,6 +249,7 @@ class RsPsiFactory(private val project: Project) {
 
 private val RsFunction.signatureText: String?
     get() {
+        val unsafe = if (isUnsafe) "unsafe " else ""
         // We can't simply take a substring of original method declaration
         // because of anonymous parameters.
         val name = name ?: return null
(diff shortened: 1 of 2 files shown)


matklad:
bors r+

Thanks, @ttaomae!

ttaomae:
It took me a little while to figure out what the `*` and `s` meant in the `chooser` argument of the `doTest` method, but from my understanding `*` indicates that the method or field is selected by default and `s` indicates that we want to implement the specified method/field. Please correct me if I'm wrong.

Assuming this is correct, I noticed that there are no tests where a method/field is selected by default, but we choose not to implement it. Is this something worth testing (in a separate pull request)?

matklad:
>It took me a little while to figure out what the * and s meant in the chooser argument of the doTest method, but from my understanding * indicates that the method or field is selected by default and s indicates that we want to implement the specified method/field. Please correct me if I'm wrong.

Yeah, we have a mini-DSL to specify the result of the interaction of the user with the chooser dialog. This is indeed not super obvious, so it would be cool if you could invent a more straightforward way to test this, or just write a short comment explaining the DSL :) 

> Assuming this is correct, I noticed that there are no tests where a method/field is selected by default, but we choose not to implement it. Is this something worth testing (in a separate pull request)?

Sure, we :heart:**love**:heart: tests!

ttaomae:
> Sure, we ❤️love❤️ tests!
Great! I'll work on this and hopefully submit a pull request in a few days.

> so it would be cool if you could invent a more straightforward way to test this, or just write a short comment explaining the DSL :)
I'll give this some thought and incorporate it into the new pull request if I come up with anything. If not I'll definitely be sure to add an explanation.

PR #1957, opened 2017-11-01
INSP: Enable automatic fixing of macro and lifetime parameter naming conventions

kumbayo:
Renaming macros and lifetime parameters works quite well.
So enable the quick fix which offers to automatically rename the item.


kumbayo:
Automatic renaming of associated types is still left disabled as the "type assoc_type = xxx" in the impl block is no renamed.

matklad:
bors r+

Thanks @kumbayo !

PR #1958, opened 2017-11-01
COMP: Auto complete #[macro_use] on "mod mod_name;"

kumbayo:


src/test/kotlin/org/rust/lang/core/completion/RsAttributeCompletionTest.kt
```diff
@@ -63,6 +63,11 @@ class RsAttributeCompletionTest : RsCompletionTestBase() {
         mod foo {}
     """)
 
+    fun testMacroUseOnMod2() = checkSingleCompletion("macro_use", """
```
matklad: Let's use `doSingleCompletion` function for the new tests!

There are two problems with `checkSingleCompletion`: first, it uses a very fiddly code to find out what exactly was completed. It also does not check the caret position after completion, which is important. 

I've added a deprecation for `checkSingleCompletion` in https://github.com/intellij-rust/intellij-rust/pull/1959, and I plan to lazily phase it out of existence :)

kumbayo: Done

matklad:
bors r+

Thanks!

PR #1972, opened 2017-11-02
Keep comments on the first line when reformatting

matklad:
closes https://github.com/intellij-rust/intellij-rust/issues/825
closes https://github.com/intellij-rust/intellij-rust/pull/1534


matklad:
bors r+

Took me a little longer than expected 😆

Issue #1028, opened 2017-02-19
TOML plugin should be compatible with canonical TOML grammar


himikof:
Now that there is a ABNF TOML [grammar](https://github.com/toml-lang/toml/blob/master/toml.abnf) upstream, TOML plugin grammar should be updated.

Related to #1027.

matklad:
Working on TOML plugin right now. And of course grammar differs from speck!

matklad:
closed by #1978

PR #1978, opened 2017-11-03
Toml

matklad:


matklad:
closes https://github.com/intellij-rust/intellij-rust/issues/1028

Not that I've rigorously tested it, but it should most stuff correctly, and we should fix remaining bugs one-by-one: official grammar contains bugs anyway https://github.com/toml-lang/toml/pull/484

Issue #1777, opened 2017-09-25
Add keyword support for cargo.toml


farodin91:
http://doc.crates.io/manifest.html
Todo 
* auto complete
* error annotate (unknown keyword, unknown crate or wrong type)
* typing

PR #1982, opened 2017-11-04
Complete keys from Cargo.toml

matklad:


matklad:
bors try

matklad:
bors try

matklad:
bors try

matklad:
Note: CLion stuff is landing separately in https://github.com/intellij-rust/intellij-rust/pull/1984.

farodin91:
Fixes #1777

src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
```diff
@@ -0,0 +1,192 @@
...
+            get() = PlatformPatterns.psiElement()
+                .inVirtualFile(VirtualFilePattern().withName("Cargo.toml"))
+                .withParent(TomlKey::class.java)
+    }
+}
+
+
+
+@Language("TOML")
+private val EXAMPLE_CARGO_TOML = """
```
vlad20012: Hmm, this looks like something temporary? Really we want to implement it like "cargo run" completion works? (it also make possible to complete values like paths or crates).

src/main/kotlin/org/rust/toml/TomlSchema.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.toml.lang.psi.TomlArrayTable
+import org.toml.lang.psi.TomlFileType
+import org.toml.lang.psi.TomlKeyValueOwner
+import org.toml.lang.psi.TomlTable
+
+class TomlSchema(project: Project, example: String) {
+    private val tables: List<TomlTableSchema>
+
+    init {
+        val toml = PsiFileFactory.getInstance(project)
```
vlad20012: I prefer to move a complex logic from constructors to factory methods :)

src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
matklad: I think that for basic "what keys are available in this table" this example approach is the long-term preferable. The toml is from http://doc.crates.io/manifest.html, so it should be easy to update it when new features are implemented in Cargo. I'll ad the manifest.html URL as a comment here. 

I wish we had a similar example with all Cargo command line options, so that the basic ones (without arguments) could be generated and not hard-coded. 

But yeah, we'll need to implement smarter completion here in the future, like suggesting a name of the package in `[dependenices]` section from the crates on crates.io.

src/main/kotlin/org/rust/toml/TomlSchema.kt
matklad: Will do!

src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
vlad20012: So may be it's better to move it to a separate file instead of inline string?

matklad: And what are the downsides to using an inline string? :) 

The upside is that we don't need to do IO to get the text from resources.

vlad20012: The IO is just hidden with java class loading =) And anyway it performs once. So I don't think it is something major.

The downsides is that, hmm, we include a large fragment of non-kotlin code to kotiln file. This looks ugly for me, but I can't say more precisely :)

src/main/kotlin/org/rust/toml/TomlSchema.kt
matklad: Fixed

src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
matklad: I don't feel strongly either way, and I am to lazy to extract this into a separate file :P

vlad20012: Okay =)

matklad:
bors r=vlad20012

PR #1996, opened 2017-11-06
Improve handling of explicit #[path = "foo/mod.rs"] attribute

kumbayo:
When the explicit path ends with "/mod.rs" then resolving inside this module would be broken.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/index/RsModulesIndex.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/index/RsModulesIndex.kt
@@ -12,6 +12,7 @@ import com.intellij.psi.stubs.StringStubIndexExtension
 import com.intellij.psi.stubs.StubIndex
 import com.intellij.psi.stubs.StubIndexKey
 import com.intellij.util.PathUtil
+import org.rust.ide.RsConstants
 import org.rust.lang.core.psi.ext.RsMod
 import org.rust.lang.core.psi.RsModDeclItem
 import org.rust.lang.core.psi.RsFile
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/stubs/index/RsModulesIndex.kt
```diff
@@ -63,7 +64,11 @@ class RsModulesIndex : StringStubIndexExtension<RsModDeclItem>() {
         private fun key(mod: RsModDeclItem): String? {
             val pathAttribute = mod.pathAttribute
             return if (pathAttribute != null) {
-                FileUtil.getNameWithoutExtension(PathUtil.getFileName(pathAttribute))
+                val fileName = PathUtil.getFileName(pathAttribute)
+                if (fileName == RsConstants.MOD_RS_FILE)
```
kumbayo: This works but I wonder if is there a better way to extract the module name from a string like "foo1/foo2/mod.rs"

matklad: No, I think this is fine!

src/main/kotlin/org/rust/lang/core/stubs/index/RsModulesIndex.kt
```diff
@@ -63,7 +64,11 @@ class RsModulesIndex : StringStubIndexExtension<RsModDeclItem>() {
...
+                    PathUtil.getParentPath(pathAttribute).split('/').last()
```
matklad: I think `pathAttribute` can use `\` here, because it's a string from a source code, which in theory might be windows only. Let's use `getFileName` here as well? 

Or maybe convert the whole thing to `nio.Path`?

kumbayo: I used `getFileName()` here.
I also adapted the `NameResolution.kt` to support Windows style paths (even on Linux systems).
Hope thats OK.

matklad:
Nice!

bors r+

Btw, @kumbayo, could you take a look at https://github.com/intellij-rust/intellij-rust/pull/2027? I think I have an idea to make our test suite even better, but I am not sure that it is such a great idea after all, so I am soliciting feedback :)

matklad:
bors delegate

matklad:
bors delegate+

PR #1998, opened 2017-11-07
WIP: Improve "implement members fix" tests.

ttaomae:
Follow up to https://github.com/intellij-rust/intellij-rust/pull/1938#issuecomment-339894860.

### DIFF
--- a/src/test/kotlin/org/rust/ide/core/overrideImplement/ImplementMembersFixTest.kt
+++ b/src/test/kotlin/org/rust/ide/core/overrideImplement/ImplementMembersFixTest.kt
@@ -5,274 +5,251 @@
 
 package org.rust.ide.core.overrideImplement
 
+import com.intellij.psi.PsiDocumentManager
 import junit.framework.TestCase
 import org.intellij.lang.annotations.Language
 import org.rust.lang.RsTestBase
 import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.ext.childOfType
-import java.util.*
 
 class ImplementMembersFixTest : RsTestBase() {
     override fun isWriteActionRequired() = true
 
-    fun test1() = doTest("""
+    fun `test implement methods`() = doTest("""
         trait T {
-            fn foo();
-            fn bar() {}
+            fn f1();
+            fn f2();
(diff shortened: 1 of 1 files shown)


ttaomae:
Some notes on the current version.

Rather than using the `*s` notation, I've created a data class to represent the same information in a hopefully more meaningful way.

I've reorganized the tests and gave them more informative names. I'm not sure if I've covered all the same test-cases, so please let me know if there is anything that should be added.

`testImplementMoreMethods()` currently fails because the actual result does not contain spaces between the method parameters. e.g.
```rust
// expected
fn f(a: i32, b: i32)
```
versus
```rust
// actual
fn f(a: i32,b: i32)
```
I'm not sure if this is the intended behavior since actually using the fix in the IDE properly inserts spaces between parameters.

matklad:
Oh, this is looking great (travis is failing though :( )! Unfortunately, I won't be able to review this until the evening :( 


One comment I had is that we try to use ``test `space separated words`()``  naming convention for the new tests. You can quickly convert from camel case to words using the following incantation:

1. select `fun test` substring
2. press `Alt+J` repeatedly to select all tests with multiple carets
3. Move caret right and press extend selection to select only the names of the tests
4. `Ctrl+X` to cut out all the names
5. Backtick to insert a pair of backtics
6. `Ctrl+V` to paste back test names
7. (Via string manipulation plugin) `Alt+Shift+M > to words` :)

matklad:
>I'm not sure if this is the intended behavior since actually using the fix in the IDE properly inserts spaces between parameters.

You need to add 

```          PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(myFixture.editor.document)
```

incantation to the test to make postponed formatting operations

ttaomae:
> One comment I had is that we try to use test ``test `space separated words`()``

Changed method names.

> You need to add [...]

It looks like something was wrong with your formatting, but I was able to get the line by viewing the page source. Added postponed formatting operations.

matklad:
>It looks like something was wrong with your formatting, but I was able to get the line by viewing the page source.

:fearful: 

bors r+

Thanks @ttaomae !

PR #2002, opened 2017-11-08
Delegate method/field resolution to type inference engine

vlad20012:
And some references refactoring. Now the one reference that can `advancedResolve` is path reference

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsPathReferenceElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsPathReferenceElement.kt
@@ -1,0 +1,12 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.resolve.ref.RsPathReference
+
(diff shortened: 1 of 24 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsPathReferenceElement.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.resolve.ref.RsPathReference
+
+interface RsPathReferenceElement : RsReferenceElement{
```
vlad20012: And do we really need this class?

vlad20012:
Then #1773 !

src/main/kotlin/org/rust/lang/core/psi/ext/RsPathReferenceElement.kt
matklad: I think we need it at the moment, though I still think that perhaps we should dramatically reduce the number of reference classes and separate "references for integration with the platform"  from "Rust-specific name resolution".

PR #2010, opened 2017-11-09
TY: block with bot expressions has bot type

### LINKED ISSUE #2006: Plugin incorrectly thinks that function returns ()
```Rust
fn a() -> i32 {
    if true {
        return 0;
    } else {
        return 6;
    };
}
```

While this code is bad style it is still valid, but plugin highlights this code as invalid: Function returns () instead of i32.

matklad:
closes #2006 

r? @vlad20012


vlad20012:
Hmm, does it also fixes #1599?

matklad:
Yep, added a test

bors r=vlad20012

matklad:
bors r+

?

PR #2027, opened 2017-11-13
T: add Testcrumbs to make testsuite more resilient to refactorings

matklad:
Sooo, I've got this idea a while ago (actually, I've stolen it from Dart), that it would be cool to connect certain tests and `if`s in code, to get precise coverage information. Today I've figured out that it can be 90% done without some custom code-generation tricks. 

What do you think?

cc @vlad20012 @Undin @mkaput

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/Testmark.kt
+++ b/src/main/kotlin/org/rust/openapiext/Testmark.kt
@@ -1,0 +1,74 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.openapiext
+
+import junit.framework.TestCase.fail
+import org.jetbrains.annotations.TestOnly
+
+
+/**
+ * Testmarks allow easy navigation between code and tests.
+ * That is, to find a relevant test for some piece of logic,
+ * and to find the precise condition which is supposed to be
+ * tested by a particular test. In a sense, testmarks are
+ * a hand-made artisanal code-coverage tool.
+ *
+ * To use a testmark, first define a static [Testmark] value
+ *
+ * ```
+ * val emptyFrobnicator = Testmark("emptyFrobnicator")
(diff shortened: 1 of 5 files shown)


kumbayo:
Looks like a nice idea for "negative" or "failing" tests.
I tried adding it to a test and wonder if it would be useful to be able to differentiate between hit once or multiple times or if that is overkill.

src/main/kotlin/org/rust/openapiext/Testmark.kt
```diff
@@ -0,0 +1,80 @@
...
+ * fun `test don't blow up on empty frobnicator`() = emptyFrobnicator.checkHit {
+ *    arrange()
+ *    act()
+ *    assert()
+ * }
+ * ```
+ *
+ * You can `Cltr+B` on `emptyFrobnicator` to navigate
+ * between code and tests. If after refactoring the
+ * test fails to hit the breadmark, you'll be notified
```
kumbayo: This still uses the name breadmark :-)

mkaput:
Nice

src/main/kotlin/org/rust/openapiext/Testmark.kt
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Note that if testmark is used in "production", but
+ * not in "test" code, it's useless, but we can't
+ * provide a runtime assertion for this case.
+ */
+class Testmark(val name: String) {
+    @Volatile
+    private var state = 0
+
+    fun hit() = hitOrSkip(this)
```
vlad20012: I think it will be faster if we just leave out `if (state == 1) state = 2` instead of this magic =)
Volatile read on x86 is just a regular read, and the condition satisfies in tests only.

matklad: Fixed!

PR #2030, opened 2017-11-14
Fix handling of #[macro_export] and #[macro_use] attributes

kumbayo:
Properly interpret the #[macro_use] and #[macro_export] attributes.


kumbayo:
@farodin91 You have implemented the current macro resolving. Maybe you can have a quick look if this change looks correct.

matklad:
@kumbayo oh, this is no longer WIP, I've missed that 😅 I'll try to review it today after the release!

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -384,9 +386,15 @@ val RsFile.exportedCrateMacros: List<RsMacroDefinition>
         CachedValueProvider.Result.create(macros, PsiModificationTracker.MODIFICATION_COUNT)
     }
 
+fun Testmark.hitOnFalse(b: Boolean): Boolean {
+    if (!b)
```
matklad: Let's join lines here. The style we use is either add braces, or put action at the same line as a condition.

kumbayo: Done, I also rebased and switched to the new ways of checking Testmark in a tests.

matklad:
bors r+

Thanks!

matklad:
bors r+

Not sure what's up with CI

Issue #1751, opened 2017-09-20
Support Sized, Send & Sync traits auto-derive


vlad20012:
When we [looking up for implemented traits](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt#L65) we should provide auto-derived traits.

- [ ] Sized
- [ ] Send
- [ ] Sync

`Sized` is the simplest one. The type is sized if it is not a DST. The type is DST if it is a trait object, slice or a struct/enum/tuple which **last** field is DST. So what we need is just a `tailrec` recursive function that checks that last field of is (not) trait object/slice.

Send/Sync looks like this:
1. All primitive types is Send and Sync
2. If the type directly implements Send/Sync, it is definitely Send/Sync (works now)
3. It the type directly implements !Send/!Sync, it is definitely **not** Send/Sync. We should create another index for `impl !Trait for S`, I guess
4. Otherwise the type is Send/Sync only if all it's fields is Send/Sync. We can use [TypeVisitor](https://github.com/intellij-rust/intellij-rust/blob/9e50411e338b4728bfc607bc19fd06a1db3977ba/src/main/kotlin/org/rust/lang/core/types/infer/Fold.kt#L14) (e.g. [`containsTyOfClass`](https://github.com/intellij-rust/intellij-rust/blob/9e50411e338b4728bfc607bc19fd06a1db3977ba/src/main/kotlin/org/rust/lang/core/types/infer/Fold.kt#L98)) to check it.

---
You can find more info about how to start contributing to Intellij-Rust in [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md), and some info about our architecure in [ARCHITECTURE.md](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md). If you have any questions about this issue, you can ask them right here or in our [gitter](https://gitter.im/intellij-rust/intellij-rust). If you decided to grab this issue, please leave a comment here and put yourself in the [worklist](https://github.com/intellij-rust/intellij-rust/wiki/Worklist).

Undin:
I'll implement it

PR #2038, opened 2017-11-15
TY: derive Sized trait impl

Undin:
Part of #1751

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeParameter.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeParameter.kt
@@ -14,6 +14,11 @@ import org.rust.lang.core.stubs.RsTypeParameterStub
 import org.rust.lang.core.types.RsPsiTypeImplUtil
 import org.rust.lang.core.types.ty.Ty
 
+/**
+ * Returns all bounds for type parameter.
+ *
+ * Don't use it for stub creation because it will cause [com.intellij.openapi.project.IndexNotReadyException]!
+ */
 val RsTypeParameter.bounds: List<RsPolybound> get() {
     val owner = parent?.parent as? RsGenericDeclaration
     val whereBounds =
@@ -22,6 +27,22 @@ val RsTypeParameter.bounds: List<RsPolybound> get() {
             .flatMap { it.typeParamBounds?.polyboundList.orEmpty() }
 
     return typeParamBounds?.polyboundList.orEmpty() + whereBounds
+}
+
+val RsTypeParameter.isSized: Boolean get() {
+    val stub = stub
+    if (stub != null) return stub.isSized
+
+    // We can't use `resolve` here to find `?Sized` bound because it causes `IndexNotReadyException` while indexing.
+    // Instead of it we just check `?` before trait name in bound
+    // because at this moment only `Sized` trait can have `?` modifier
+    val owner = parent?.parent as? RsGenericDeclaration
+    val whereBounds =
+        owner?.whereClause?.wherePredList.orEmpty()
+            .filter { (it.typeReference?.typeElement as? RsBaseType)?.name == name }
+            .flatMap { it.typeParamBounds?.polyboundList.orEmpty() }
+    val bounds = typeParamBounds?.polyboundList.orEmpty() + whereBounds
+    return bounds.none { it.q != null }
 }
 
 abstract class RsTypeParameterImplMixin : RsStubbedNamedElementImpl<RsTypeParameterStub>, RsTypeParameter {
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -21,8 +21,7 @@ import org.rust.lang.core.types.ty.Mutability.IMMUTABLE
 import org.rust.lang.core.types.ty.Mutability.MUTABLE
 import org.rust.lang.core.types.type
 import org.rust.openapiext.ProjectCache
-import org.rust.stdext.buildList
-import org.rust.stdext.concatLists
+import org.rust.stdext.buildSet
 import kotlin.LazyThreadSafetyMode.NONE
 
 enum class StdDerivableTrait(val modName: String, val dependencies: Array<StdDerivableTrait> = emptyArray()) {
@@ -91,24 +90,27 @@ class ImplLookup(
             return _ctx!!
         }
 
-    fun findImplsAndTraits(ty: Ty): Collection<RsTraitOrImpl> {
+    fun findImplsAndTraits(ty: Ty): Set<RsTraitOrImpl> {
         return findImplsAndTraitsCache.getOrPut(project, freshen(ty)) { rawFindImplsAndTraits(ty) }
     }
 
-    private fun rawFindImplsAndTraits(ty: Ty): Collection<RsTraitOrImpl> {
-        return when (ty) {
-            is TyTypeParameter -> ty.getTraitBoundsTransitively().map { it.element }
-            is TyTraitObject -> BoundElement(ty.trait).flattenHierarchy.map { it.element }
+    private fun rawFindImplsAndTraits(ty: Ty): Set<RsTraitOrImpl> {
+        val implsAndTraits = mutableSetOf<RsTraitOrImpl>()
+        when (ty) {
+            is TyTypeParameter -> ty.getTraitBoundsTransitively().mapTo(implsAndTraits) { it.element }
+            is TyTraitObject -> BoundElement(ty.trait).flattenHierarchy.mapTo(implsAndTraits) { it.element }
             is TyFunction -> {
-                findSimpleImpls(ty) + fnTraits
+                implsAndTraits += findSimpleImpls(ty)
+                implsAndTraits += fnTraits
             }
-            is TyUnknown -> emptyList()
-            else -> concatLists(
-                findDerivedTraits(ty),
-                getHardcodedImpls(ty).map { it.element },
-                findSimpleImpls(ty)
-            )
+            is TyUnknown -> Unit
+            else -> {
+                implsAndTraits += findDerivedTraits(ty)
+                implsAndTraits += findSimpleImpls(ty)
+                getHardcodedImpls(ty).mapTo(implsAndTraits) { it.element }
+            }
         }
+        return implsAndTraits
     }
 
     private fun findDerivedTraits(ty: Ty): Collection<RsTraitItem> {
@@ -219,7 +221,7 @@ class ImplLookup(
         }
     }
 
-    private fun assembleCandidates(ref: TraitRef): List<SelectionCandidate> = buildList {
+    private fun assembleCandidates(ref: TraitRef): Set<SelectionCandidate> = buildSet {
         addAll(assembleImplCandidates(ref))
         addAll(assembleDerivedCandidates(ref))
         if (ref.selfTy is TyFunction && ref.trait.element in fnTraits) add(SelectionCandidate.Closure)
@@ -448,7 +450,7 @@ class ImplLookup(
             ImplLookup(psi.project, StdKnownItems.relativeTo(psi))
 
         private val findImplsAndTraitsCache =
-            ProjectCache<Ty, Collection<RsTraitOrImpl>>("findImplsAndTraitsCache")
+            ProjectCache<Ty, Set<RsTraitOrImpl>>("findImplsAndTraitsCache")
 
         private val traitSelectionCache =
             ProjectCache<TraitRef, SelectionResult<SelectionCandidate>>("traitSelectionCache")
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -67,6 +67,9 @@ class ImplLookup(
     private val copyTrait: RsTraitItem? by lazy(NONE) {
         RsNamedElementIndex.findDerivableTraits(project, "Copy").firstOrNull()
     }
+    private val sizedTrait: RsTraitItem? by lazy(NONE) {
+        RsLangItemIndex.findLangItem(project, "sized")
+    }
     private val derefTraitAndTarget: Pair<RsTraitItem, RsTypeAlias>? = run {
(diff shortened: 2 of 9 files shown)


src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
```diff
@@ -85,3 +88,30 @@ fun mergeFlags(element: BoundElement<*>): TypeFlags =
 
 fun mergeFlags(tys: List<Ty>): TypeFlags =
     tys.fold(0) { a, b -> a or b.flags }
+
+tailrec fun Ty.isSized(): Boolean {
```
Undin: I don't really like that this function is public but I haven't come up with how to make it private because I use it in two places

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -58,15 +61,22 @@ class TyTypeParameter private constructor(
...
+        private val self = TyTypeParameter(Self, false, { emptyList() })
 
         fun self(): TyTypeParameter = self
 
-        fun self(trait: RsTraitOrImpl): TyTypeParameter =
-            TyTypeParameter(Self, { listOfNotNull(trait.implementedTrait) })
+        fun self(item: RsTraitOrImpl): TyTypeParameter {
+            val isSized = when (item) {
+                // TODO: support trait bounds
+                is RsTraitItem -> false
```
Undin: I don't know if we really want to check trait bounds here because we get all bounds in `ImplLookup#rawFindImplsAndTraits` and it brings `Sized` trait in result set so related test passes because of this hack.

src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeParameter.kt
```diff
@@ -24,6 +24,13 @@ val RsTypeParameter.bounds: List<RsPolybound> get() {
     return typeParamBounds?.polyboundList.orEmpty() + whereBounds
 }
 
+val RsTypeParameter.isSized: Boolean get() {
+    return stub?.isSized ?: bounds.none {
+        if (it.q == null) return@none false
+        it.bound.traitRef?.resolveToTrait?.isSizedTrait ?: false
```
matklad: For dealing with nullable booleans, ` == true` is more idiomatic: https://github.com/JetBrains/kotlin-web-site/blob/465783bf395acc5d34f089291eca594ef3d7d4ce/pages/docs/reference/coding-conventions.md#using-nullable-boolean-values-in-conditions

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -91,24 +93,28 @@ class ImplLookup(
...
             }
-            is TyUnknown -> emptyList()
-            else -> concatLists(
-                findDerivedTraits(ty),
-                getHardcodedImpls(ty).map { it.element },
-                findSimpleImpls(ty)
-            )
         }
+        implsAndTraits += findImplicitImpls(ty)
+        return implsAndTraits
```
matklad: 👍

matklad: In general, I think we should switch our style from recklessly concatenating read only collections to  the pattern of creating a mutable collection, filling it with values and then freezing it. There are helper functions like `buildList` which could help here, though in this particular case just creating a mutable set seems clearer.

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
matklad: That's fine I think.

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
```diff
@@ -85,3 +88,30 @@ fun mergeFlags(element: BoundElement<*>): TypeFlags =
...
+                namedFields.isNotEmpty() -> namedFields.last().typeReference
+                tupleFields.isNotEmpty() -> tupleFields.last().typeReference
+                else -> null
+            }
+            val type = typeRef?.type?.substitute(typeParameterValues) ?: return true
+            type.isSized()
+        }
+
+        is TyTuple -> types.last().isSized()
+        else -> false
```
matklad: Is `TyNever` unhanded? 

I think we should return `true` by default here, because you have to opt-out of sizedness usually.

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
```diff
@@ -80,10 +90,17 @@ class TyTypeParameter private constructor(
         fun associated(type: Ty, trait: RsTraitItem, target: RsTypeAlias): TyTypeParameter {
             return TyTypeParameter(
                 AssociatedType(type, trait, target),
-                { emptyList() } )
+                // TODO: support associated type bounds
+                true,
```
matklad: Bare `true` looks not very readable on GitHub, let's add `isSized = true` for readability?

matklad:
LGTM!

r? @vlad20012

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
Undin: > Is TyNever unhanded?

`TyNever` is primitive type so it is sized too)

Undin:
Fixed comments

src/test/kotlin/org/rust/lang/core/type/RsImplicitTraitsTest.kt
```diff
@@ -0,0 +1,165 @@
...
+                   //^ Sized
+    """)
+
+    fun `test type parameter is Sized by default`() = doTest("""
+        fn foo<T>() -> T { unimplemented!() }
+                     //^ Sized
+    """)
+
+    fun `test type parameter with ?Sized bound is not Sized`() = doTest("""
+        fn foo<T: ?Sized>() -> Box<T> { unimplemented!() }
```
vlad20012: Let's add a test where `T: Sized` explicitly: `fn foo<T: Sized>`. It allows to make sure that we handle `?` correctly

src/main/kotlin/org/rust/lang/core/types/ty/TyTypeParameter.kt
vlad20012: All this code is a set of hacks and will be dismantled soon, so everything is OK =)

src/test/kotlin/org/rust/lang/core/type/RsImplicitTraitsTest.kt
Undin: Done

vlad20012:
Thanks a lot!
bors r+

vlad20012:
Hmm, I just found what's done wrong. When I wrote the issue, I mentioned `findImplsAndTraits`, and that was true at that time. But now I even don't sure that we need to change anything in `findImplsAndTraits`. The main place we now need Sized/Send/Sync auto-derive is `assembleCandidates`. And `isSized`/`isCopy` should be implemented like `select(TraitRef(ty, sizedTrait .withSubst()))`

Undin:
Ok, I'll fix it a little bit later)

Undin:
@vlad20012 fixed

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -91,24 +93,27 @@ class ImplLookup(
...
-    fun findImplsAndTraits(ty: Ty): Collection<RsTraitOrImpl> {
+    fun findImplsAndTraits(ty: Ty): Set<RsTraitOrImpl> {
         return findImplsAndTraitsCache.getOrPut(project, freshen(ty)) { rawFindImplsAndTraits(ty) }
     }
 
-    private fun rawFindImplsAndTraits(ty: Ty): Collection<RsTraitOrImpl> {
-        return when (ty) {
-            is TyTypeParameter -> ty.getTraitBoundsTransitively().map { it.element }
-            is TyTraitObject -> BoundElement(ty.trait).flattenHierarchy.map { it.element }
+    private fun rawFindImplsAndTraits(ty: Ty): Set<RsTraitOrImpl> {
```
vlad20012: As I see, nothing changed here except we now use `Set` instead of `List`. What are the reasons for this? I'm worried about this also because the returned collection always stores to the cache, so it is long-leaving object, and HashSet consumes much more memoty than plain ArrayList.

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -219,9 +224,10 @@ class ImplLookup(
         }
     }
 
-    private fun assembleCandidates(ref: TraitRef): List<SelectionCandidate> = buildList {
+    private fun assembleCandidates(ref: TraitRef): Set<SelectionCandidate> = buildSet {
```
vlad20012: Same question about `Set`, but here, I guess, the point is that we can somehow run into multiple implementations of `Sized` trait (but I don't understand how). Anyway, think it's ok for now because I going to optimize this method (i.e. why we need to call the costly `assembleImplCandidates` for type parameters?), and we can handle `Sized` trait as a special case in the same manner (i.e. why we need to call the costly `assembleImplCandidates` if we selecting `Sized` trait that can't be implemented manually?).

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
matklad: Semantically, we want a `Set` of impls here. Is there a compact array-based readonly set implementation somewhere?

vlad20012: Furthermore, `mutableSetOf` used there, so actually it's the `LinkedHashSet` =)

@matklad We don't have something like `TObjectHashSet` (but have in `Koloboke` collections! which we don't use...). Anyway I think that semantic meaning is secondary in as hot place as this.

matklad: Hm, but can we have both proper semantic & good performance? 

```
class ArraySet<T>(private val data: Array<T>) : Set<T> {
    override val size: Int get() = data.size
    override fun contains(element: T): Boolean = data.contains(element)
    override fun containsAll(elements: Collection<T>): Boolean = elements.all { data.contains(it) }
    override fun isEmpty(): Boolean = data.isEmpty()
    override fun iterator(): Iterator<T> = data.iterator()

    companion object {
        inline fun <reified T> fromSet(s: Set<T>): ArraySet<T> = ArraySet(s.toTypedArray())
    }
}
```

> so actually it's the LinkedHashSet =)

Being deterministic is a nice property, for tests especially. So, would constructing a LinkedHashSet and then converting it to an ArraySet be OK performance-wise?

vlad20012: @matklad As I know, java's `Set` guarantees O(1) for `contains`... And it looks rather like a hack than more beautiful solution =)

Undin:
Oh, I've found out that these changes cause my favorite `IndexNotReadyException`

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
matklad: >As I know, java's Set must provide O(1) for contains.

Even `TreeSet`? :trollface:

vlad20012: Hmm, but not O(n) anyway! =D

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
Undin: Very stupid case is
```rust
fn foo<T: Trait>(t: T) where T: Trait { ... }
```
It looks like valid rust function for any `Trait`.

My case is
```rust
fn foo<T: Sized>(t: T) { ... }
```
In this case `Sized` trait comes from explicit type bound and from `assembleImplicitCandidates`. Of course, I can check this case manually but I think `Set`
 can solve both problems more elegantly.

Undin: Hm, if we want to have special case for `Sized` trait in `assembleCandidates` I can implement it in this PR - we already have `Ty.isSized` function and we use it in `assembleCandidates` directly for `Sized` trait case. What do you think about it?

vlad20012: Ok

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
vlad20012: About @Undin's description for `Set` in the comment below:
Hmm, it's reasonable. It would be cool to add some tests for method lookup with `fn foo<T: Trait>(t: T) where T: Trait { ... }`. Without a `Set` we will run into ambiguity?

Undin:
* fixed `IndexNotReadyException`
* created special case in `assembleCandidates` for `Sized` trait
* changed run configuration to show `idea.log` in IDE

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -260,6 +272,21 @@ class ImplLookup(
             .map { SelectionCandidate.DerivedTrait(it) }
     }
 
+    private fun sizedTraitCandidates(ty: Ty, sizedTrait: RsTraitItem): Set<SelectionCandidate> {
+        var isSized = ty.isSized()
+        if (!isSized && ty is TyTypeParameter) {
+            isSized = ty.getTraitBoundsTransitively().any { it.element == sizedTrait }
```
vlad20012: Hmm, why we need this branch?

Undin: In this case
```rust
trait Foo : Sized {
    fn foo(self: Self);
}
```
`Self` type should be `Sized` and we need to check super traits.
But constructing of `TyTypeParameter.Self` is better place to do it so I've moved corresponding code there.

vlad20012: Oh, Ok

vlad20012:
Thanks again :)

PR #2039, opened 2017-11-16
GRAM: pull attributes inside expressions

matklad:
Fix for https://github.com/intellij-rust/intellij-rust/pull/1829#issuecomment-338390755

So there were two problem here:

1) we attached outer attributes to statements, while the (nightly) rust language allows attaching attributes to expressions. So we failed to parse 100% legit Rust code like `#[foo]-#[bar]-#[baz]92` :) 

2) The rule for the tail expression of the block was horribly broken and produced bogus AST in presence of attributes. 

The rule looked like this:

Place an `Expr` marker, parse outer attributes, parse an expression, if there's a semicolon, switch to `ExprStmt`, else use some concrete `Expr`.

The problem here is that if there are no attributes, the tree looks like this:

```
Expr
  BlockExpr
```

That is, `BlockExpr` is the sole child of `Expr`, and the `Expr` is flattened away, because BlockExpr extends expr. 

**But**, if ther's an attribute, the tree looks like this:

```
Expr
Attr BlockExpr
```

And we can't flatten out `Expr` in a similar way.

The fix to both issues is to move attributes *inside* expressions.

r? @Undin


Undin:
Looks good:)
bors r+

Issue #2042, opened 2017-11-18
[Minor] Refactoring suggestion from assert! to assert_eq removes explicit error


FatalCatharsis:
The refactor suggestion to change a general assert statement from assert! to assert_eq! removes the specific error message.

Before:
![image](https://user-images.githubusercontent.com/2822917/32976382-655870d8-cbda-11e7-87e4-cf7b17fb92c0.png)

After applying refactor:
![image](https://user-images.githubusercontent.com/2822917/32976388-7732a90e-cbda-11e7-8c8a-200350a9e12d.png)

PR #2044, opened 2017-11-18
INSP Apply assert arg on refactor of AssertEqualInspection

FatalCatharsis:
Addresses issue #2042

Changes the AssertEqualInspection refactor suggestion to append the rest of the argument list.

Originally, applying the refactor would pull apart the binary expression and replace it with two args, but any subsequent args on the assert were deleted.

(this is my first PR to this project so I wanted to start small)

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsAssertEqualInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsAssertEqualInspection.kt
@@ -75,19 +75,17 @@ class RsAssertEqualInspection : RsLocalInspectionTool() {
             val factory = RsPsiFactory(project)
             val newAssert = factory.createExpression("$assertName!()") as RsCallExpr
 
-            val argPair = comparedAssertArgs(assertArgument) ?: return null
+            val (first, second) = comparedAssertArgs(assertArgument) ?: return null
 
             val args: MutableList<PsiElement> = mutableListOf(
-                argPair.first,
+                first,
                 factory.createComma(),
-                argPair.second
+                second
             )
 
-            if (assertArgument.formatMacroArgList.isNotEmpty()) {
-                for (arg: PsiElement in assertArgument.formatMacroArgList) {
-                    args.add(factory.createComma())
-                    args.add(arg)
-                }
+            for (arg: PsiElement in assertArgument.formatMacroArgList) {
+                args.add(factory.createComma())
+                args.add(arg)
             }
 
             addArguments(args, newAssert.valueArgumentList.lparen, newAssert.valueArgumentList)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/inspections/RsAssertEqualInspection.kt
```diff
@@ -71,12 +71,25 @@ class RsAssertEqualInspection : RsLocalInspectionTool() {
...
+
+            val argPair = comparedAssertArgs(assertArgument) ?: return null
+
+            val args: MutableList<PsiElement> = mutableListOf(
+                argPair.first,
+                factory.createComma(),
+                argPair.second
+            )
+
+            if (assertArgument.formatMacroArgList.isNotEmpty()) {
```
matklad: I think we could remove this `if`, because the loop would be empty anyway.

src/main/kotlin/org/rust/ide/inspections/RsAssertEqualInspection.kt
```diff
@@ -71,12 +71,25 @@ class RsAssertEqualInspection : RsLocalInspectionTool() {
...
-        private fun buildAssert(project: Project, str: String, argPair: Pair<PsiElement, PsiElement>): PsiElement {
+        private fun buildAssert(project: Project, assertArgument: RsAssertMacroArgument): PsiElement? {
             val factory = RsPsiFactory(project)
-            val newAssert = factory.createExpression(str) as RsCallExpr
-            val args: MutableList<PsiElement> = mutableListOf(argPair.first)
-            args.add(factory.createComma())
-            args.add(argPair.second)
+            val newAssert = factory.createExpression("$assertName!()") as RsCallExpr
+
+            val argPair = comparedAssertArgs(assertArgument) ?: return null
```
matklad: It would be slightly more idiomatic to write `let (left, right) = comparedAssert`

FatalCatharsis:
Not sure if I did the CLA right. I went through the docusign portal and filled it out earlier. Do I need to paste it anywhere here or email it to someone?

matklad:
>Do I need to paste it anywhere here or email it to someone?

No, just add yourself to CONTRIBUTORS.txt in this PR :)

FatalCatharsis:
Hey, I'm

src/main/kotlin/org/rust/ide/inspections/RsAssertEqualInspection.kt
FatalCatharsis: Agreed, that's much nicer. First time I've used Kotlin, so I'm still picking things up. 
1f7dd3a

src/main/kotlin/org/rust/ide/inspections/RsAssertEqualInspection.kt
FatalCatharsis: Yup, redundant check. 
1f7dd3a

FatalCatharsis:
Added my github username to CONTRIBUTORS.txt 8f7667a and added some tests 829c226. Had trouble understanding the syntax for a bit, but I got the gist. Question, do you know why the tests required that <caret> block at the end? My test kept failing without it.

matklad:
bors r+

Thanks!

> Question, do you know why the tests required that <caret> block at the end? My test kept failing without it.

The `<caret>` specifies the position of the editor caret in the text document. It is required because what the tests do is that they invoke the light-bulb <kdb>Alt+Enter</kbd> keybinding, which looks for suggestions for the current context around the caret

Issue #1692, opened 2017-09-04
Type inference of `impl Trait` RFC 1522


vlad20012:
[RFC 1522](https://github.com/rust-lang/rfcs/pull/1522)
```rust
trait Trait {
    fn bar() {}
}     //X
fn foo() -> impl Trait { unimplemented!() }
fn main() {
    foo().bar();
}        //^
```
I know that `impl Trait` is an unstable feature, but is used increasingly in rustc itself, and I tired of stumbling over type inference failures caused by this =D

It is pretty easy to implement - just create new type `TyAnon` in package `org.rust.lang.core.types.ty` with single field of type `List<BoundElement<RsTraitItem>>`, instantiate it in `Declarations.inferTypeReferenceType` for element of type `RsImplTraitType` and then add a special case for it in the `ImplLookup.findImplsAndTraits` just like `TyTypeParameter`.

---
You can find more info about how to start contributing to Intellij-Rust in [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md), and some info about our architecure in [ARCHITECTURE.md](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md). If you have any questions about this issue, you can ask them right here or in our [gitter](https://gitter.im/intellij-rust/intellij-rust). If you decided to grab this issue, please leave a comment here and put yourself in the [worklist](https://github.com/intellij-rust/intellij-rust/wiki/Worklist).

oleg-semenov:
Hi @vlad20012, i decided to try to do this.

PR #2048, opened 2017-11-19
(TY): add inference of `impl Trait`

oleg-semenov:
This fulfills the issue:
https://github.com/intellij-rust/intellij-rust/issues/1692

### DIFF
--- a/src/main/kotlin/org/rust/ide/presentation/Ty.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/Ty.kt
@@ -67,6 +67,7 @@ private fun render(
             is TyInfer.FloatVar -> float
         }
         is FreshTyInfer -> "<fresh>" // really should never be displayed; debug only
+        is TyAnon -> ty.traits.map { it.element.name ?: anonymous }.joinToString("+", "impl ")
         else -> error("unreachable")
     }
 }
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -47,6 +47,7 @@ mrhota
 netvl
 ninjabear
 oistein
+oleg-semenov
 osialr
 pasa
 pavel-v-chernykh
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -106,6 +106,7 @@ class ImplLookup(
                 implsAndTraits += findSimpleImpls(ty)
                 implsAndTraits += fnTraits
             }
+            is TyAnon -> ty.traits.mapTo(implsAndTraits) { it.element }
             is TyUnknown -> Unit
             else -> {
                 implsAndTraits += findDerivedTraits(ty)
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt
@@ -59,6 +59,9 @@ fun inferTypeReferenceType(ref: RsTypeReference): Ty {
             TyFunction(paramTypes, type.retType?.let { it.typeReference?.type ?: TyUnknown } ?: TyUnit)
         }
 
+        is RsImplTraitType -> {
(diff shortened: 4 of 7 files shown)


src/main/kotlin/org/rust/lang/core/types/TyFingerprint.kt
```diff
@@ -42,6 +42,7 @@ data class TyFingerprint constructor(
                 }
                 is RsArrayType -> TyFingerprint("[T]")
                 is RsFnPointerType -> TyFingerprint("fn()")
+                is RsImplTraitType -> TyFingerprint("Anon")
```
vlad20012: In the case of `impl Trait` we don't need to change `TyFingerprint`.
`TyFingerprint` is used as a key for indexing (caching) rust impls (`impl X { }`), and it's impossible to write something like `impl impl Trait {}`, so this line is useless :)

src/main/kotlin/org/rust/ide/presentation/Ty.kt
```diff
@@ -67,6 +67,7 @@ private fun render(
             is TyInfer.FloatVar -> float
         }
         is FreshTyInfer -> "<fresh>" // really should never be displayed; debug only
+        is TyAnon -> ty.traits.map { it.element.name ?: anonymous }.joinToString()
```
vlad20012: Hm, it's interesting question - how we should display `TyAnon`. I think it should be displayed as close to it's syntax as possibe, i.e. `impl A+B+C`. So I think we should write here something like `ty.traits.map { it.element.name ?: anonymous }.joinToString("+", "impl ")`

vlad20012: And this code isn't tested. Let's add one simple test to `RsTypeResolvingTest`! :)

src/test/kotlin/org/rust/lang/core/type/RsExpressionTypeInferenceTest.kt
```diff
@@ -810,4 +810,16 @@ class RsExpressionTypeInferenceTest : RsTypificationTestBase() {
             a
         } //^ <unknown>
     """)
+
+    fun `test impl Trait`() = testExpr("""
```
vlad20012: Hmm, this test is better to move to `RsTypeAwareResolveTest`, because we really want to check that `bar` is resolved to the method we want. I also suggest to add there similar test, but with multiple traits `impl Trait1+Trait2`

vlad20012:
Oh, I waited a long time for someone to do it =D Thanks a lot!

As I see, it's your first contribution here. We need an electronic CLA for contributions (it's very simple, just few clicks): https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#cla

After you sign the CLA, please add your name to
https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTORS.txt

src/main/kotlin/org/rust/ide/presentation/Ty.kt
oleg-semenov: Sounds good. Done.

src/main/kotlin/org/rust/lang/core/types/TyFingerprint.kt
oleg-semenov: Done.

src/test/kotlin/org/rust/lang/core/type/RsExpressionTypeInferenceTest.kt
oleg-semenov: Hm maybe done. For the Trait1+Trait2 case it looks like i can't have multiple `//^`in one test. So I made 2 tests testing the same source code but with different markers.

oleg-semenov:
@vlad20012, thank you for your review, I updated the commit, and will do the electronic CLA for contributions + CONTRIBUTORS.txt update soon.

vlad20012:
Ok, now everything looks fine! I'm awaiting for CLA to merge it =)

oleg-semenov:
CLA should be done

vlad20012:
bors r+
Thanks again!

oleg-semenov:
Thank you, I’ll try pick something to work on soon.

PR #2049, opened 2017-11-20
ANN: function args and return type must be Sized

Undin:
* check that types of function args and return type have `Sized` trait bound
* add quick fixes: convert to reference and convert to `Box`


Undin:
I've found out that `ImplLookup.isSized()` implemented with `assembleCandidates` method is not convenient to use because we need `Sized` trait in our tests.
For these annotations, I use `Ty.isSized()` directly. Maybe drop `ImplLookup.isSized()` to have only one API for this goal?

vlad20012:
@Undin as I remember, `ImplLookup.isSized()` and `Ty.isSized()` are not fully equivalent, but now it's so. You changed something before I merged the PR and I missed it =)
Then yes, now `ImplLookup.isSized()` is useless, I guess

vlad20012:
Thanks!
bors r+

Undin:
@vlad20012 yes, at the beginning `Ty.isSized()` didn't check supertraits of trait itself but when we decided to create special case for `Sized` trait in `ImplLookup.assembleCandidates()` I had to do it.

PR #2054, opened 2017-11-21
Fixed "unification error"

vlad20012:


matklad:
bors r+

I can't reproduce the error, but given that it was pretty hard to reproduce initially, I am still no 100% sure that this fixes this error, so let's hope for the best :(

To bad Kotlin is not Rust and can't statically prevent such problems :(

PR #2058, opened 2017-11-22
Method ambiguity resolution

vlad20012:
All `RsPreciseTraitMatchingTest` tests now passes
1. We now filter methods of traits that are not imported
2. We now check trait bounds of impls

We do not do anything above if we don't have method name ambiguity, i.e. name resolution succeeded with exactly single entity

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -6,12 +6,15 @@
 package org.rust.lang.core.resolve
 
 import com.intellij.openapi.project.Project
+import org.rust.lang.core.macros.setContext
 import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.RsPsiFactory
 import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.RsTypeAlias
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.indexes.RsImplIndex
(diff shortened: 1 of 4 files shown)


vlad20012:
About the last commit: The idea is that if we resolved exactly 1 method, we don't perform
any additional filtering and just use this 1 method.

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -472,6 +475,12 @@ class ImplLookup(
             return TyFunction(argumentTypes, outputType)
         }
 
+    fun isTraitVisibleFrom(trait: RsTraitItem, scope: RsElement): Boolean {
+        val name = trait.name ?: return true
+        val path = RsPsiFactory(project).tryCreatePath(name)?.apply { setContext(scope) } ?: return true
+        return resolvePath(path, this).any { it.element == trait }
```
matklad: I wonder if these uncaheble resolves have any impact on performance.

vlad20012: This code anyway executes very rarely - only if we run into method name ambiguity (resolved 2 methods with the same name). I've checked it with our `the best` :) performance tests and did not noticed any regression

matklad: Yeah, a fair point about this being rare!

I foresee though that one day we'll start to use this for "import trait" quick fix, and it'll become a bottleneck :) But we definitely should not try to optimize it right now!

PR #1997, opened 2017-11-07
TY: Finally dismantled `Ty.unifyWith` & lot of other type inference changes

vlad20012:

PR #2061, opened 2017-11-23
RES: Enable `impl<T: Bound> for T`

vlad20012:
Some infrastructure for this was prepared in #1997 and #2058. Additionally now we resolve ambiguous trait impls during trait selection

@Undin, it's done =)

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareGenericResolveTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareGenericResolveTest.kt
@@ -264,14 +264,16 @@ class RsTypeAwareGenericResolveTest : RsResolveTestBase() {
         }
     """)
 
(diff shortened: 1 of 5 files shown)


vlad20012:
Hmm, I think I should also remove these "unresolved" tests...

src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareGenericResolveTest.kt
```diff
@@ -264,14 +264,16 @@ class RsTypeAwareGenericResolveTest : RsResolveTestBase() {
         }
     """)
 
+    // TODO should be unresolved
     fun `test impl not resolved by accident if struct name is the same as generic type`() = checkByCode("""
         struct S;
         trait Tr1{}
         trait Tr2{ fn some_fn(&self) {} }
```
matklad: missing space before `{`

PR #2066, opened 2017-11-23
Resolve method ambiguity based on deref level / if there are multiple impls of the same trait to the same type

vlad20012:


matklad:
This could probably use a couple of `Testmark`s? :)

vlad20012:
+ one more commit
+ use Teskmark

src/main/kotlin/org/rust/stdext/Collections.kt
```diff
@@ -36,3 +36,13 @@ fun makeBitMask(bitToSet: Int): Int = 1 shl bitToSet
 
 fun <K, V1, V2> zipValues(map1: Map<K, V1>, map2: Map<K, V2>): List<Pair<V1, V2>> =
     map1.mapNotNull { (k, v1) -> map2[k]?.let { v2 -> Pair(v1, v2) } }
+
+inline fun <T> List<T>.singleOrFilter(predicate: (T) -> Boolean): List<T> = when {
+    size < 2 -> this
+    else -> filter(predicate)
+}
+
+inline fun <T> List<T>.singleOrLet(function: (List<T>) -> List<T>): List<T> = when {
```
vlad20012: Can't think of a better name :(

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTestBase.kt
```diff
@@ -42,6 +43,9 @@ abstract class RsResolveTestBase : RsTestBase() {
         check(resolved == target)
     }
 
+    protected fun checkByCode(@Language("Rust") code: String, mark: Testmark) =
```
vlad20012: I think something like this make usage of `Testmark`s much more clear

vlad20012:
+ more docs

vlad20012:
bors r=matklad

vlad20012:
There is a real problem with that test on windows. It fails locally somtimes...

PR #2071, opened 2017-11-26
PRJ: support rust project creation in small IDEs

### LINKED ISSUE #557: Support creating cargo projects in small IDEs
That is, implement #401 for PyCharm, CLion, etc. 

See `com.intellij.platform.DirectoryProjectGenerator`

Undin:
Partially closes #557 

This implementation should work with:
* **PyCharm Community** (checked with 172.4343.24 and 173.3727.88)
* **PyCharm Professional** (checked with 172.4343.24 and 173.3727.88)
* **PyCharm Edu** (checked with 172.4539)
* **WebStorm** (checked with 172.4343.25 and 173.3727.69)
* **RubyMine** (checked with 172.4155.44 and 173.3727.77)
* **PhpStorm** (checked with 172.4155.41 and 173.3727.84)

Unfortunately, project creation in **CLion** still **doesn't work**

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
+++ b/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
@@ -21,7 +21,10 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import javax.swing.JLabel
 
-class RustProjectSettingsPanel(private val cargoProjectDir: Path = Paths.get(".")) : Disposable {
+class RustProjectSettingsPanel(
+    private val cargoProjectDir: Path = Paths.get("."),
+    private val updateListener: (() -> Unit)? = null
+) : Disposable {
     data class Data(
         val toolchain: RustToolchain?,
         val explicitPathToStdlib: String?
@@ -114,6 +117,7 @@ class RustProjectSettingsPanel(private val cargoProjectDir: Path = Paths.get("."
                     toolchainVersion.text = rustcVersion.parsedVersion
                     toolchainVersion.foreground = JBColor.foreground()
                 }
+                updateListener?.invoke()
             }
         )
     }
--- a/src/main/kotlin/org/rust/ide/newProject/RsDirectoryProjectGenerator.kt
+++ b/src/main/kotlin/org/rust/ide/newProject/RsDirectoryProjectGenerator.kt
@@ -5,15 +5,23 @@
 
 package org.rust.ide.newProject
 
+import com.intellij.ide.util.projectWizard.AbstractNewProjectStep
+import com.intellij.ide.util.projectWizard.CustomStepProjectGenerator
+import com.intellij.ide.util.projectWizard.ProjectSettingsStepBase
 import com.intellij.openapi.module.Module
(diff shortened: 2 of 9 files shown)


src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
```diff
@@ -21,7 +21,10 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import javax.swing.JLabel
 
-class RustProjectSettingsPanel(private val cargoProjectDir: Path = Paths.get(".")) : Disposable {
+class RustProjectSettingsPanel(
+    private val cargoProjectDir: Path = Paths.get("."),
+    private val updateListener: (() -> Unit)? = null
+) : Disposable {
```
Undin: Do we really need to implement `Disposable` here?
We do nothing in `dispose` method

matklad: We don't own any resources that need cleanup directly, but `UiDebouncer` does: it needs a parentDisposable in constructor.

src/main/kotlin/org/rust/ide/newProject/RsDirectoryProjectGenerator.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.openapi.wm.impl.welcomeScreen.AbstractActionWithPanel
+import com.intellij.platform.DirectoryProjectGenerator
+import com.intellij.platform.DirectoryProjectGeneratorBase
+import com.intellij.platform.ProjectGeneratorPeer
+import org.rust.ide.icons.RsIcons
+import javax.swing.Icon
+
+// We implement `CustomStepProjectGenerator` as well to correctly show settings UI
+// because otherwise PyCharm doesn't add peer's component into project settings panel
```
matklad: Ah, so this is the magic ingredient to make it work in PyCharm! Ive pulled a lot of hair out over this one! Kudos!

src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
matklad: Hm, I think that perhaps instead `RustProjectSettingsPanel` can accept parent disposable as an argument instead?

matklad:
LGTM! r=me when we figure out what's the proper way to handle Disposables here.

src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
Undin: Oh, I've understood why we need `Disposable` here.

> Hm, I think that perhaps instead `RustProjectSettingsPanel` can accept parent disposable as an argument instead?

I think it would be not very convenient because we still use `RustProjectSettingsPanel` in `RustProjectConfigurable` directly and we don't have parent disposable.
Maybe don't change it at all?

PR #2072, opened 2017-11-26
DOC: the oldest current supported IDE version is 2017.2

Undin:


matklad:
Though, perhaps we can strip this info altogether? 

It was relevant when the plugin only appeared, and there existed IDEA version for which not a single version of plugin worked. Now, you can install plugin anywhere, but you won't received updates for older IDEs.

Undin:
Hm, but still there are old versions of IDEA where plugin doesn't work at all.
But it is really not very useful info because plugin repository contains this info (and info about each plugin version) and it is more relevant in case of plugin installation.

matklad:
>Hm, but still there are old versions of IDEA where plugin doesn't work at all.

That's like 14.something versions, I doubt anyone uses them :)

PR #2085, opened 2017-11-30
MACRO: expand `&item*` pattern

matklad:
if_std in futures-rs and failure now works

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -1,0 +1,77 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.macros
+
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.CachedValueProvider
+import com.intellij.psi.util.CachedValuesManager
+import com.intellij.psi.util.PsiModificationTracker
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.childrenOfType
+
+class MacroExpander(project: Project) {
+    val psiFactory = RsPsiFactory(project)
+
+    fun expandMacro(def: RsMacroDefinition, call: RsMacroCall): List<ExpansionResult>? {
+        val case = def.macroDefinitionBodyStubbed
+            ?.macroDefinitionCaseList
+            ?.singleOrNull()
+            ?: return null
+
+        val patGroup = case.macroPattern.macroBindingGroupList.singleOrNull()
+            ?: return null
+        val patBinding = patGroup.macroBindingList.singleOrNull()
+            ?: return null
+        val name = patBinding.colon.prevSibling?.text
+        val type = patBinding.colon.nextSibling?.text
+        if (!(name != null && type != null && type == "item")) return null
+
+        val expGroup = case.macroExpansion.macroExpansionReferenceGroupList.singleOrNull()
+            ?: return null
+        val expansionText = expGroup.textBetweenParens(expGroup.lparen, expGroup.rparen)?.toString() ?: return null
+
+        val items = parseBodyAsItemList(call) ?: return null
+        val newText = items.joinToString("\n\n") { item ->
+            expansionText.replace("$" + name, item.text)
+        }
+
+        val expandedFile = psiFactory.createFile(newText)
+
+        return expandedFile.childrenOfType()
+    }
+
+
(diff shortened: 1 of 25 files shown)


matklad:
r? @vlad20012 

Have not measure the performance impact, will do now. 

I think the implementation is more or less solid, but I absolutely hate this `ItemCache` stuff now :(

matklad:
Performance regressed for ~ 30% 😞 

<details>

```
BEFORE:

highlighting mysql_async:
collecting  : 4 ms
resolve     : 2817 ms
highlighting: 184 ms
3005 ms total.

highlighting Cargo:
collecting  : 4 ms
resolve     : 850 ms
highlighting: 91 ms
945 ms total.

AFTER:

highlighting mysql_async:
collecting  : 3 ms
resolve     : 4139 ms
highlighting: 198 ms
4340 ms total.

highlighting Cargo:
collecting  : 4 ms
resolve     : 1114 ms
highlighting: 98 ms
1216 ms total.
```

</details>

src/main/kotlin/org/rust/lang/core/macros/MacroExpansion.kt
```diff
@@ -5,16 +5,103 @@
...
+
+inline fun <reified I : RsElement> createInMemoryPsi(project: Project, code: String): I? {
+    return createMacroExpansionFile(project, code)
+        .descendantOfTypeStrict()
+}
+
+private val RsMacroDefinition.body: RsMacroDefinitionBody?
+    get() {
+        val stub = stub ?: return macroDefinitionBody
+        val text = stub.macroBody ?: return null
```
vlad20012: Hmm, can you please explain why we storing a raw text to stub instead of stubbing all nodes? Is it something fundamental or it's just easier to implement?

vlad20012: And we even don't cache expanded body anywhere and should parse it again for every usage :(

src/test/kotlin/org/rust/lang/core/resolve/RsMacroExpansionResolveTest.kt
```diff
@@ -7,6 +7,9 @@ package org.rust.lang.core.resolve
 
 
 class RsMacroExpansionResolveTest: RsResolveTestBase() {
+    @Suppress("PrivatePropertyName")
+    private val `$` = '$'
```
vlad20012: `private const val`

src/main/kotlin/org/rust/lang/core/macros/MacroExpansion.kt
matklad: It's much easier to implement (it's hard to stub leaf elements, and we need them here) and, I think, might actually be more efficient, because we wan't be storing the actual tree structure.

vlad20012: Let's try to insert a cache here. I wonder how this will impact performance.

matklad:
Added a cache (which helped a bit!) and make sure we use a single instance of RsPsiFactory (not that it should impact performance, but it should nudge us into grouping macro-related stuff together). 

So, this is ready for review I guess :)

src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        return expandedFile.childrenOfType()
+    }
+
+
+    private val RsMacroDefinition.macroDefinitionBodyStubbed: RsMacroDefinitionBody?
+        get() {
+            val stub = stub ?: return macroDefinitionBody
+            val text = stub.macroBody ?: return null
+            return CachedValuesManager.getCachedValue(this) {
```
Undin: Hm, we have a lot of similar code with caches.
Maybe create extension function for that?
Something like
```rust
fun <T> RsElement.getCachedValue(provider: () -> T): T {
    return CachedValuesManager.getCachedValue(this) {
        CachedValueProvider.Result.create(
            provider(),
            PsiModificationTracker.MODIFICATION_COUNT
        )
    }
}
```

vlad20012: It use java class of second parameter of `getCachedValue` as an identity (key) of our cache, so we must write it manually each time we want to create a new cache ((

Undin: Oh, I've understood
Thanks)

Undin:
If open `future-rs` project and go to `src/stream/collect.rs` all items from other files can't be resolved. But if you comment usage of `if_std` macros in `src/stream/mod.rs` resolve works

matklad:
Hm, yeah, that one is going to be tough =/ The problem is that we can't determine parent module in this case, because `mod collect;` mod decl is inside `if_std!`, so it is not present in indexes ...

PR #2092, opened 2017-12-01
MACRO: fix stub based code path for macro definition

matklad:
This is embarrassing 😓 

@Undin this might fix the problem with futures.


Undin:
Unfortunately, it still doesn't work for me :(
But looks like 
```rust
if_std! {
   struct Foo;
}

fn foo() -> Foo { unimplemented!() }
           //^
```
resolves now but without correct navigation (but it expected)

Issue #2081, opened 2017-11-27
Wrong internal text instead of item name for E0603 (accessing private items)


kumbayo:
Example code:
```rust
mod foo {
    const BAR: u32 = 0x_a_bad_1dea_u32;
}
use foo::BAR;
```

Current error message:
![unification_after](https://user-images.githubusercontent.com/1041267/33285797-a2604c18-d3b3-11e7-9c48-43c8e9cda70e.png)

Before #1789 (where we did not print the item name)
![unification_before](https://user-images.githubusercontent.com/1041267/33285788-9e8ad220-d3b3-11e7-9406-94400dcb196f.png)

Rustc error message:
```
error[E0603]: constant `BAR` is private
   --> src/main.rs:232:5
    |
232 | use foo::BAR;
    |
```

Undin:
Fixed in #2097

PR #2097, opened 2017-12-03
Module visibility

Undin:
* Introduce `RsVisible` interface and implement it for `RsMod`
* Annotate if you try to use private module
* Minor improvement in 'outside of scope' error description (including fix of #2081)

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -89,11 +89,25 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
             }
         }
 
-        val error = when (ref) {
-            is RsFieldLookup -> RsDiagnostic.AccessError(ref, RsErrorCode.E0616)
-            is RsMethodCall -> RsDiagnostic.AccessError(ref.identifier, RsErrorCode.E0624)
-            is RsPath -> RsDiagnostic.AccessError(ref, RsErrorCode.E0603)
-            else -> return
+        val error = when {
+            element is RsFieldDecl -> {
+                val structName = element.ancestorStrict<RsStructItem>()?.crateRelativePath?.removePrefix("::") ?: ""
+                RsDiagnostic.StructFieldAccessError(ref, ref.referenceName, structName)
+            }
+            ref is RsMethodCall -> RsDiagnostic.AccessError(ref.identifier, RsErrorCode.E0624, "Method")
+            else -> {
+                val itemType = when (element) {
+                    is RsMod -> "Module"
+                    is RsConstant -> "Constant"
+                    is RsFunction -> "Function"
+                    is RsStructItem -> "Struct"
+                    is RsEnumItem -> "Enum"
+                    is RsTraitItem -> "Trait"
+                    is RsTypeAlias -> "Type alias"
+                    else -> "Item"
+                }
+                RsDiagnostic.AccessError(ref, RsErrorCode.E0603, itemType)
+            }
         }
         error.addToHolder(holder)
     }
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -17,7 +17,6 @@ import org.rust.ide.annotator.fixes.AddModuleFileFix
 import org.rust.ide.annotator.fixes.AddTurbofishFix
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
-import org.rust.lang.core.psi.impl.RsMembersImpl
 import org.rust.lang.core.resolve.Namespace
 import org.rust.lang.core.resolve.namespaces
 import org.rust.lang.core.types.inference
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -77,6 +77,12 @@ class RsFile(
             return cargoWorkspace?.isCrateRoot(file) ?: false
         }
 
+    override val isPublic: Boolean get() {
+        if (isCrateRoot) return true
+        val parentMod = `super` ?: return false
+        return parentMod.itemsAndMacros.any { it is RsModDeclItem && it.name == modName && it.isPublic }
```
matklad: This does the same work twice, because `super` also has to find a `ModDecl`. Let's add something like `RsFile.declaration: RsModDecl?` and use it to implement both `super` and `isPublic`?

`org.rust.ide.navigation.goto.RsGotoSuperHandler` could also use this functionality.

src/main/kotlin/org/rust/lang/core/psi/ext/RsMod.kt
```diff
@@ -33,6 +33,8 @@ interface RsMod : RsQualifiedNamedElement, RsItemsOwner {
 
     val isCrateRoot: Boolean
 
+    val isPublic: Boolean
```
matklad: Hm, perhaps we should split `VisibilityOwner` into `Visible` and `PubOwner: Visible`?

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
```diff
@@ -49,7 +49,11 @@ fun gotoSuperTarget(source: PsiElement): NavigatablePsiElement? {
...
 
     val mod = modOrMethod as RsMod
-    val superMod = mod.`super` ?: return null
-    return superMod.itemsAndMacros.filterIsInstance<RsModDeclItem>().find { it.reference.resolve() == mod }
-        ?: superMod
+    return when (mod) {
+        is RsFile -> mod.declaration
+        else -> {
+            val superMod = mod.`super` ?: return null
+            superMod.itemsAndMacros.filterIsInstance<RsModDeclItem>().find { it.reference.resolve() == mod } ?: superMod
```
Undin: @matklad we can find `RsModDeclItem` only in `RsFile`, can't we?
If it's true, then it should be simplified to
```kotlin
return when (mod) {
    is RsFile -> mod.declaration
    else -> mod.`super`
}
```

matklad: >we can find RsModDeclItem only in RsFile, can't we?

Nope :) 

```
fn foo() {
    #[path="bar.rs"]
    mod bar;
}
```

matklad: My gut feeling is that it can be completely removed, because we actually find `modDecl` when we are evaluating `` `super` ``.

Undin: Do you mean just return `mod.super` here?
Then we will navigate incorrectly in next case
```rust
//- foo/qqq.rs
/*caret*/

//- main.rs
mod foo { // <- we will go here
    #[path="qqq.rs"]
    mod bar; // <- but should go here
}
```

But looks like that
```rust
return when (mod) {
    is RsFile -> mod.declaration
    else -> mod.`super`
}
```
works as expected.
I've added some tests with `path` attribute

matklad: Yeah, the current implementation looks great: we now longer do the work of finding the declaration twice, and we don't encode the same logic in several functions anymore!

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -69,26 +68,46 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
+            ref is RsMethodCall -> RsDiagnostic.AccessError(ref.identifier, RsErrorCode.E0624, "Method")
+            else -> {
+                val itemType = when (element) {
+                    is RsMod -> "Module"
+                    is RsConstant -> "Constant"
+                    is RsFunction -> "Function"
+                    is RsStructItem -> "Struct"
+                    is RsEnumItem -> "Enum"
+                    is RsTraitItem -> "Trait"
+                    is RsTypeAlias -> "Type alias"
```
matklad: On day, we should consolidate all rendering logic in once place....

src/test/kotlin/org/rust/ide/annotator/RsAnnotatorTestBase.kt
```diff
@@ -36,8 +36,8 @@ abstract class RsAnnotatorTestBase : RsTestBase() {
         @Language("Rust") after: String
     ) = checkByText(before, after) { applyQuickFix(fixName) }
 
-    protected fun checkDontTouchAstInOtherFiles(fileTree: FileTree, checkInfo: Boolean = false, filePath: String? = null) {
-        fileTree.create()
+    protected fun checkDontTouchAstInOtherFiles(@Language("Rust") text: String, checkInfo: Boolean = false, filePath: String? = null) {
```
matklad: :+1:

src/test/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandlerTest.kt
```diff
@@ -72,17 +73,45 @@ class RsGotoSuperHandlerTest : RsTestBase() {
...
+        /*caret*/    // only comment
+
+        //- main.rs
+        mod foo {
+            #[path="qqq.rs"]
+            mod bar;
+        }
+    """, expected = """
+        #[path="qqq.rs"]
+            mod bar;
```
matklad: Indentation is off here.

Undin: Yeah, but target text has this weird indent too. I don't really know why

matklad: Ah,  because we are looking at `element.text`, which includes indentation for all but the first line!

Yeah, let's leave this as is, though it would be cool to invent some better way to mark target elements for tests.

Undin: It would be cool to compare file tree before and after as we do in single file case

matklad: Hm, yeah, I think we can use file tree to compare caret positions!

Issue #1939, opened 2017-10-27
Collapse use statements


itmuckel:
Some of my rust files have a lot of use statements.

As a programmer
I'd like to collapse the use statements
to reduce noise, when working on my code.

Please add a possibility to collapse use statements like you can collpase functions.

PS: Maybe a different issue, but collapsing crates would be nice, too.

matklad:
If anyone wants to look into this, the code is in https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/folding/RsFoldingBuilder.kt.

framlog:
I'm going to do this.

Undin:
Implemented in #2101

PR #2101, opened 2017-12-04
Code foldings

Filippok:
I have implemented some code foldings. Now it can fold repeating items like uses, mods and, crates.

Fulfilled #1939 request.


matklad:
Excellent, thanks @Filippok !

bors r+

PR #2103, opened 2017-12-05
Sibling mod visibility

Undin:
Correctly check access to sibling modules

### DIFF
--- a/src/test/kotlin/org/rust/ide/annotator/RsAnnotatorTestBase.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsAnnotatorTestBase.kt
@@ -38,16 +38,11 @@ abstract class RsAnnotatorTestBase : RsTestBase() {
 
     protected fun checkDontTouchAstInOtherFiles(@Language("Rust") text: String, checkInfo: Boolean = false, filePath: String? = null) {
         fileTreeFromText(text).create()
-        myFixture.configureFromTempProjectFile("main.rs")
-
+        val testFilePath = filePath ?: "main.rs"
         (myFixture as CodeInsightTestFixtureImpl) // meh
-            .setVirtualFileFilter { !it.path.endsWith("main.rs") }
-        if (filePath == null) {
-            myFixture.testHighlighting(false, checkInfo, false)
-        } else {
-            myFixture.testHighlighting(false, checkInfo, false, filePath)
-        }
+                    .setVirtualFileFilter { !it.path.endsWith(testFilePath) }
+
+        myFixture.configureFromTempProjectFile(testFilePath)
+        myFixture.testHighlighting(false, checkInfo, false)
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/ide/annotator/RsAnnotatorTestBase.kt
```diff
@@ -38,16 +38,10 @@ abstract class RsAnnotatorTestBase : RsTestBase() {
 
     protected fun checkDontTouchAstInOtherFiles(@Language("Rust") text: String, checkInfo: Boolean = false, filePath: String? = null) {
         fileTreeFromText(text).create()
-        myFixture.configureFromTempProjectFile("main.rs")
-
-        (myFixture as CodeInsightTestFixtureImpl) // meh
-            .setVirtualFileFilter { !it.path.endsWith("main.rs") }
```
matklad: Hm, is this hack no longer necessary in current versions of IDE?

I remember that it was necessary because `checkAstNotLoaded` would not work as expected, because `testHighlighting` would override the `setAssertOnFileLoadingFilter`.


Could you check that tests indeed fail if you touch AST (i.e, try calling `mod.superMods.forEach { it.text }`)

Undin: Hm, at this moment it looks like a magic for me:)
`mod.superMods.forEach { it.text }` doesn't trigger file filter at all and it doesn't matter how we set this filter: trough `myFixture.setVirtualFileFilter` or through `checkAstNotLoaded`.
But `testHighlighting` really set own file filter.
I will investigate it later.

matklad: Yeah, I remember spending couple of hours on this line **two times** :) It's definitely fishy.

Undin: I've understood how it works.

Some code invoked by `testHighlighting`
```java
final VirtualFileFilter fileTreeAccessFilter = myVirtualFileFilter;
Disposable disposable = Disposer.newDisposable();
if (fileTreeAccessFilter != null) {
  PsiManagerEx.getInstanceEx(project).setAssertOnFileLoadingFilter(fileTreeAccessFilter, disposable);
}
try {
  doHighlighting();
} finally {
  Disposer.dispose(disposable);
}
```

So we really should set file filter through `myFixture.setVirtualFileFilter`.

But 
```kotlin
oMod.superMods.forEach { it.text }
```
in `RsErrorAnnotator.checkReferenceIsPublic` still doesn't trigger file filter.
Although, for example,
```kotlin
oMod.superMods.forEach { (it as? RsFile)?.declaration?.text }
```
 does it.

PR #2106, opened 2017-12-07
Up down

matklad:
This adds <kbd>Ctrl+Shift+Up/Down</kbd> support for all comma lists, like function parameters & arguments, struct fields, etc. 

One complication with this is that you need to automatically insert a comma sometimes. Having written the code to do it, I though that it would be cool to add it to our formatter as well.

r? @vlad20012 or @Undin


Undin:
Looks like that in some cases we have two different ways to do this action: <kbd>Ctrl+Shift+Up/Down</kbd> and <kbd>Ctrl+Shift+Left/Right</kbd>
Maybe it's simpler to use `RsMoveLeftRightHandler` for this case because it's much simpler to implement (just return items of comma-separated list)? Yes, it doesn't insert comma at the end of list, but do we really want to insert it in the end of list?

matklad:
Yeah, that's true that we have two implementations for this action. But I am pretty sure that the users expect <kbd>ctrl+alt+Up/Down</kbd> to work, especially with Rust convention of putting parameters on separate lines, so we need to support this. Perhaps we could somehow reuse the implementation? I am not sure it is possible tough

Issue #1989, opened 2017-11-04
Implement members fix lists members that are already implemented


ttaomae:
Compare the behavior for Rust:
![rust-implement-members](https://user-images.githubusercontent.com/1213865/32408717-8f4b55b6-c141-11e7-9435-547cbc1a8fa6.png)

versus Kotlin:
![kotlin-implement-members](https://user-images.githubusercontent.com/1213865/32408745-4ed6b132-c142-11e7-82ec-9e4808768c43.png)

ttaomae:
It looks like this has been resolved. Closing.

PR #2108, opened 2017-12-07
#1989: Implement trait members - remove already implement members

GuillaumeEveillard:
Hello,

I propose a fix for "Implement members fix lists members that are already implemented" #1989

I'm not sure the change I made in TraitImplementationInfo is the best solution. If someone has a better idea, I will be happy to update my fix.

Guillaume


matklad:
Looks great @GuillaumeEveillard!

bors r+

PR #2111, opened 2017-12-10
PRJ: create temporary action for new rust project creation in CLion

Undin:
Temporary workaround for new project creation in CLion.
Should be removed when CLion will support not CMake projects.

* Introduce 'New Rust Project' action
* Add this action to welcome screen in CLion because otherwise it can be found only from search


matklad:
A great idea!

bors r+

PR #2113, opened 2017-12-10
Improve handling of attributes on literals and type/lifetime parameters

### LINKED ISSUE #1893: Parser does not allow cfg attribs in const slices
## Example

```rust
const EXTENSIONS: &'static [&'static str] = &[
    #[cfg(debug_assertions)]
    "VK_EXT_debug_report",
];
```

kumbayo:
Do not cause exceptions when attributes are present on literals.
This fixes #1893

Do not show E0199 when unsafely implementing a safe trait but a #[may_dangle] attribute is present.
This avoids showing errors for valid code in the rust standard library.


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -241,15 +241,25 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         val traitRef = impl.traitRef ?: return
         val trait = traitRef.resolveToTrait ?: return
         val traitName = trait.name ?: return
+
+        fun mayDangleOnTypeOrLifetimeParameters(impl: RsImplItem): Boolean {
+            return impl.typeParameters.any() { it.queryAttributes.hasAtomAttribute("may_dangle") } ||
```
kumbayo: I am not sure if this added queryAttributes could have a performance impact.

kumbayo:
We do not yet support auto formatting of attributes on literals, so I am not sure if we should consider #1893 fully fixed or not.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
matklad: Should be fine I think.

matklad:
bors r+

>We do not yet support auto formatting of attributes on literals, so I am not sure if we should consider #1893 fully fixed or not.

Let's consider it fixed: we'll tighten-up our formatter once rustfmt 1.0 is released.

Issue #1624, opened 2017-08-13
Wrong E0428 is shown when a macros has the same name as a type/module 


kumbayo:
When a macro and a module or a macro and a struct have the same name, a wrong error is shown
`A type named xxx has already been defined in this module [E0428]`

```rust
fn main() {
    println!("Hello, world!");
}

macro_rules! example {
    () => ()
}

//struct example { } // <-Error here when uncommented

mod example { } // <-Error here
```
![intellij_e0428](https://user-images.githubusercontent.com/1041267/29252880-b41d41f0-806f-11e7-9b7f-5bc174890638.png)

kumbayo:
I am working on this one.

PR #2114, opened 2017-12-10
Fix namespacing of macro definitions

kumbayo:
In Rust macros are in a different namespace than names and types.

This change fixes a wrong E0428 error when a macro has the same name
as a module, struct or function.


matklad:
bors r+

Thanks!

kumbayo:
This fixes #1624

PR #2117, opened 2017-12-11
INT: initial support of auto import

Undin:
Note: 
* It doesn't support reexports yet
* It is intention temporarily to avoid two problems:
    * unresolved references because of macros
    * lack of reexport support

  So it should allow user to use auto import in some simple cases and don't annoy him a lot


src/test/kotlin/org/rust/ide/intentions/ImportNameIntentionStdTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun `test import struct from std crate`() = doAvailableTest("""
+        fn foo<T: Read/*caret*/>(t: T) {}
+    """, """
+        use std::io::Read;
+
+        fn foo<T: Read/*caret*/>(t: T) {}
+    """, ImportNameIntention.Testmarks.autoInjectedCrate)
+
+    // TODO: change this test after reexports support
+    fun `test import struct from not std crate`() = doAvailableTest("""
```
Undin: Is there a good way to create tests with some extern crates (not from stdlib)?

matklad: This could be handled via custom project descriptor, see `WithStdlibAndDependencyRustProjectDescriptor` for example.

Undin: Oh, thanks)
Then I'll change these tests to avoid unnecessary changes in future when we will support reexports in auto import

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,230 @@
...
+        if (TyPrimitive.fromPath(path) != null || path.reference.resolve() != null) return null
+
+        val parent = path.path
+        val parentRes = parent?.reference?.resolve()
+        if (parent == null || parentRes is RsMod || parentRes is RsEnumItem) {
+            val basePath = getBasePath(path)
+            val pathMod = path.containingMod
+            // TODO: support reexports
+            val candidates = RsNamedElementIndex.findElementsByName(project, basePath.referenceName)
+                .filterIsInstance(RsQualifiedNamedElement::class.java)
```
matklad: `.filterIsInstance<RsQualifiedNameElement>()`

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,230 @@
...
+
+        return null
+    }
+
+    private fun RsVisible.isVisibleIn(mod: RsMod): Boolean {
+        val parentMod = (if (this is RsMod) `super` else containingMod) ?: return false
+        val ourSuperMods = parentMod.superMods
+        val superMods = mod.superMods
+
+        val superModsSet = HashSet(superMods)
```
matklad: Microoptimzation, but we can calculate this once for all elements.

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,232 @@
...
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val path = element.ancestorStrict<RsPath>() ?: return null
+        if (TyPrimitive.fromPath(path) != null || path.reference.resolve() != null) return null
+
+        val parent = path.path
+        val parentRes = parent?.reference?.resolve()
+        if (parent == null || parentRes is RsMod || parentRes is RsEnumItem) {
+            val basePath = getBasePath(path)
```
matklad: I am a bit confused over the logic here. We can autoimport only the first component of the path. However, we can invoke intention on any part of the path

```
foo::bar::/*caret*/baz
^^^ unresolved
```

This sounds just right! What I can't understand is why do we try to resolve the first qulifier here? 


```
foo::bar::baz::quux::fr/*caret*/ob;
^               ^- why do we try to resolve this component?
| - this is what we want to auto import
```

Shouldn't the logic be like this instead:

```
base = getBasePath(path)
if (base.reference.resolve() != null) return
try to import base
```

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,232 @@
...
+            val pathMod = path.containingMod
+            val pathSuperMods = HashSet(pathMod.superMods)
+            // TODO: support reexports
+            val candidates = RsNamedElementIndex.findElementsByName(project, basePath.referenceName)
```
matklad: In the future, we can do something smart with SearchScopes here to quickly filter out types from transitive dependencies, etc.

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,232 @@
...
+    // Semantic signature of method is `RsVisible.isVisibleIn(mod: RsMod)`
+    // but in our case `mod` is always same and `mod` needs only to get set of its super mods
+    // so we pass `superMods` instead of `mod` for optimization
+    private fun RsVisible.isVisibleIn(superMods: Set<RsMod>): Boolean {
+        val parentMod = (if (this is RsMod) `super` else containingMod) ?: return false
+        val ourSuperMods = parentMod.superMods
+
+        // try to find latest common ancestor module of `parentMod` and `mod` in module tree
+        // we need to do it because we can use direct child items of any super mod with any visibility
+        val lca = ourSuperMods.find { it in superMods }
```
matklad: Interesting! I suppose that we might need an `lca` when we are figuring out the optimal import path. So perhaps this function should return not only a boolean result, but also a path (perhaps as a string) which should be used in the corresponding `use` item?

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,232 @@
...
+                    panel.add(psiRenderer.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus))
+                    panel
+                }
+            }
+        }
+        NavigationUtil.hidePopupIfDumbModeStarts(popup, project)
+        popup.showInBestPositionFor(editor)
+    }
+
+    private fun importItem(project: Project, item: RsQualifiedNamedElement, originalPath: RsPath) = WriteCommandAction.runWriteCommandAction(project) {
```
matklad: `project.runWriteCommandAction {}`

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,232 @@
...
+            override fun isSpeedSearchEnabled(): Boolean = true
+            override fun hasSubstep(selectedValue: RsQualifiedNamedElement?): Boolean = false
+
+            override fun onChosen(selectedValue: RsQualifiedNamedElement?, finalChoice: Boolean): PopupStep<*>? {
+                if (selectedValue == null) return PopupStep.FINAL_CHOICE
+                return doFinalStep { importItem(project, selectedValue, originalPath) }
+            }
+
+            override fun getTextFor(value: RsQualifiedNamedElement): String =
+                ObjectUtils.assertNotNull(value.qualifiedName)
```
matklad: Just `value.qualifiedName!!`?

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
```diff
@@ -0,0 +1,232 @@
...
+                    baseRenderer.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus)
+                    panel.add(baseRenderer.nextStepLabel, BorderLayout.EAST)
+                    panel.add(psiRenderer.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus))
+                    panel
+                }
+            }
+        }
+        NavigationUtil.hidePopupIfDumbModeStarts(popup, project)
+        popup.showInBestPositionFor(editor)
+    }
```
matklad: I expect auto import machinery to grow significantly over time, so we might want to create an `import` package and move all UI to a separate file.

matklad: See also src/main/kotlin/org/rust/lang/refactoring/introduceVariable/ui.kt for an example of mock-based testing infrastructure for similar UI-dialogs.

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
matklad: Ah, so we need to handle extern crates as well... Perhaps `RsVisible.canBeImported() -> ImportDescription?` is needed here, where Import Description is a sealed class?

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
Undin: Yeah, `ObjectUtils.assertNotNull` is used in similar java import action and remained here by accident

src/main/kotlin/org/rust/ide/intentions/ImportNameIntention.kt
Undin: Yeah, it makes sense!

vlad20012:
bors r+
🎉🎉🎉

vlad20012:
bors try

Undin:
bors r=vlad20012

Undin:
bors r=vlad20012

PR #2118, opened 2017-12-11
Fix some compiler warnings

Undin:


matklad:
CI is failing, so we might want to silcence some of them with `//BACKCOMPAT:` + `@Supress` combo

Undin:
I've forgotten we still support IDEA 2017.2 :)

PR #2119, opened 2017-12-11
Do not show errors on &mut ref to a function/method

### LINKED ISSUE #1858: False positive with global functions: Cannot borrow immutable local variable as mutable
Hi,

The [following valid code](https://play.rust-lang.org/?gist=e9210d63cca54810f65785aed56cc58b&version=stable) triggers this false positive:

```
fn global() {}

fn main() {
	let local = &mut global;
}
```

screenshot_20171013_160308

kumbayo:
Getting a &mut reference to a function/method works in rustc so we must show an error for it.

Fixes #1858


matklad:
bors r+

Thaks!

Issue #2063, opened 2017-11-23
Closure type inference seems incorrect


slavam2605:
Here is a piece of code:
```rust
let f = |x: i32| {
    if x > 0 {
        return x;
    }
    panic!("Negative!");
};
```
Plugin infers a type `f: fn() -> !`, since I think it is `fn() -> i32`.

RikardLegge:
I have done som investigation on the latest master whilst trying to get familiar with the codebase.
My initial hypotesis was that the type inference prioritised the diverging function over the i32 which is true but it's not the only factor at play. 

The following two examples, even though they might not be idiomatic rust, they compile and run. `rustc` infers the type `(i32) -> i32` to the closure whilst the intellij plugin infers `(i32) -> !`.
```rust
let f = |x: i32| {
    if x > 0 {
        return x;
    } else {
        return x;
    };
};
```
or even
```rust
let f = |x: i32| {
    return x;
};
```
It seems that when using the return keyword, closures are unable to automatically resolve their type. Does this sound familiar? 
Not using a return statement in the second examples and using an implicit return produces the desired result `(i32) -> i32`.

Issue #2029, opened 2017-11-14
Highlight bindings implementing Copy


gilescope:
Would it be possible to colour or highlight differently literals and things that implement the Copy trait versus non-copyable things? It would be great to try this out as a proof of concept to see if it helped people. Is it possible / easy to try it?

matklad:
Yeah, I think it should be not very hard to do: there'll be false positives, but they don't matter for highlighting a lot.

PR #2121, opened 2017-12-12
highlight structs implementing Copy

framlog:
implemented #2029 

Edit:
<del>fixed up #2063 perhaps.</del> (it's moved to #2150 )


Undin:
@framlog Could you revert files removing from `.idea` folder?

kumbayo:
From some very quick tests it seems to not recognize structs containing immutable references as Copy.
```
#[derive(Copy, Clone)]
struct A2<'a> {
    a: &'a u32
}
```
Or structs containing tuples
```
#[derive(Copy, Clone)]
struct A3 {
    t: (u32, u16),
}
```
I guess this is because our `TyReference` and `TyTuple` are not derived from `TyPrimitive`

I think it would be nice to have some test coverage for this especially around primitives.
For example `&u32` implements Copy while `&mut u32` does not.

framlog:
@kumbayo fixed up, thx. But I don't know how to test them properly. I don't find any useful document about this.

kumbayo:
@framlog I do not know the best way to test this on the annotation level. :-/ Maybe others can chime in for that.

If you want to test the "Copy" trait on the type system level you can have a look at RsImplicitTraitsTest.
The test currently only supports the "Sized" trait but I have a branch that should allow testing for "Copy".
https://github.com/kumbayo/intellij-rust/commits/preparation-copy-tests

I guess `Ty.isSized()` and `ImplLookup.assembleCandidates` might be good place to start investigation for hooking up the Copy information to the type system (if you plan to do that).

src/main/kotlin/org/rust/lang/core/types/ty/TyTuple.kt
```diff
@@ -18,5 +18,7 @@ data class TyTuple(val types: List<Ty>) : Ty(mergeFlags(types)) {
         types.any(visitor)
 
     override fun toString(): String = tyToString(this)
+
+    override val isCopyable = types.fold(true) { acc, ty -> acc && ty.isCopyable }
```
kumbayo: Would this also work as `types.all { it.isCopyable }` ?
(Same at other places)

src/main/kotlin/org/rust/ide/colors/RsColors.kt
```diff
@@ -32,9 +32,11 @@ enum class RsColor(humanName: String, val default: TextAttributesKey) {
 
     CRATE("Crate", Default.IDENTIFIER),
     STRUCT("Struct", Default.CLASS_NAME),
+    STRUCT_COPYABLE("Copyable struct", Default.MARKUP_TAG),
```
kumbayo: With new highlighting elements it would be a good idea to also update the color highlighting demo.
See RsColorSettingsPage.kt / highlighterDemoText.rs

src/main/kotlin/org/rust/lang/core/types/ty/TyTuple.kt
framlog: yep, `types.all` is better.

src/main/kotlin/org/rust/ide/colors/RsColors.kt
framlog: Thanks for reminding me

framlog:
@kumbayo Thanks for suggestions. More tests will be added when I'm available for these

src/test/kotlin/org/rust/lang/core/type/RsImplicitTraitsTest.kt
```diff
@@ -135,6 +136,67 @@ class RsImplicitTraitsTest : RsTypificationTestBase() {
         }
     """)
 
+    fun `test primitive types are Copy`() = checkPrimitiveTypes("Copy")
+
+    fun `test slice impl Copy inference 1`() = doTest("""
+        struct S;
+        fn foo() -> &[S] { unimplemented!() }
+                   //^ !Copy
```
kumbayo: Are you sure this should not be Copy? My tests seem to indicate otherwise.

kumbayo:
I wonder why is there a check for `Deref`?
As far as i can see implementing `Deref` has no influence on the `Copy` trait...
If there is a reason to not highlight it as such, it should probably be mentioned in the commit message or the code.

According to the documentation arrays are also Copy if the type they contain is Copy.

Can you replace the `#[derive(Copy)]` by `#[derive(Copy, Clone)]` to avoid future errors in the tests when we notice that deriving only `Copy` is not allowed.

src/test/kotlin/org/rust/lang/core/type/RsFunctorTypeInferenceTest.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.rust.lang.core.types.type
+
+
+class RsFunctorTypeInferenceTest: RsTypificationTestBase() {
+    fun `test if with all branches diverging`() = testFunctor("""
+        fn main() {
+            let x = if true { return; } else { return; };
+            x;
+        }
+      //^ ()
```
kumbayo: Previously this tests all checked the type of x.
Is this change really indented to check the type of the fn?

kumbayo:
The commit `refined type inference for blocks` introduces changes to both type inference and also to handling of the Copy trait. Can you separate those changes in different commits?

I also do not know how strongly others feel about this but i think it would be nice if the block type inference changes were in a separate pull request.
(The 2 topics are quite different from each other and i think it makes reviewing and accepting the changes individually easier).
Thanks

src/test/kotlin/org/rust/lang/core/type/RsFunctorTypeInferenceTest.kt
```diff
@@ -0,0 +1,73 @@
...
+    """)
+
+    fun `test one branch diverges`() = testFunctor("""
+        fn main() {
+            let x = if true { return; } else { 92 };
```
kumbayo: Your change has changed the type of `x` from `i32` to `()` which does not match what rustc does.

framlog:
Thanks for reviewing. The ‘deref’ should be replaced by ‘drop’. It’s a mistake I think. sorry for that. I’ll check other suggestions when I back to my laptop.

src/main/kotlin/org/rust/lang/core/types/ty/TySlice.kt
```diff
@@ -18,4 +18,6 @@ data class TySlice(val elementType: Ty) : Ty(elementType.flags) {
         elementType.visitWith(visitor)
 
     override fun toString(): String = tyToString(this)
+
+    override val isCopyable = elementType.isCopyable
```
kumbayo: In most cases it wont matter (as slices are typically used behind references) but as far as i can see slices are currently not `Copy` in rust.
https://github.com/rust-lang/rust/issues/30483
```rust
trait T1 {
    type Assoc1: Copy + ?Sized;
}

struct S1 {}

impl T1 for S1 {
    type Assoc1 = [u8];
}
```
During compiling rustc  says `the trait std::marker::Copy is not implemented for [u8]`

kumbayo:
This seems to include 3 of my commits (with you as the comitter, probably due to `git rebase -i`) which are already merged in master.
Can you please rebase this branch on top of origin/master
or drop my patches from your branch (using for example `git rebase -i`)?

framlog:
@kumbayo fixed

src/main/kotlin/org/rust/lang/core/types/ty/TyStructOrEnum.kt
```diff
@@ -22,10 +24,23 @@ abstract class TyStructOrEnumBase(flags: TypeFlags) : Ty(flags) {
...
+    } ?: true
+}
+
 class TyStruct private constructor(
     private val boundElement: BoundElement<RsStructItem>
 ) : TyStructOrEnumBase(mergeFlags(boundElement)) {
 
+    override val isCopyable: Boolean get() {
+        val lookup = ImplLookup.relativeTo(boundElement.element)
+        return lookup.isCopy(this) && !lookup.isDrop(this) && boundElement.element.canBeCopyable()
```
Undin: I think we should check only `Copy` trait implementation for this feature even rustc will emit an error while compiling.
These checks can be part of inspection to highlight incompatibilities in `Copy` implementation.

src/main/kotlin/org/rust/lang/core/psi/ext/RsEnumItem.kt
```diff
@@ -12,6 +12,8 @@ import org.rust.ide.icons.RsIcons
 import org.rust.lang.core.macros.ExpansionResult
 import org.rust.lang.core.psi.RsEnumItem
 import org.rust.lang.core.psi.RsPsiImplUtil
+import org.rust.lang.core.psi.RsStructItem
+import org.rust.lang.core.resolve.ImplLookup
```
Undin: Unused imports

src/main/kotlin/org/rust/lang/core/psi/ext/RsStructItem.kt
```diff
@@ -13,9 +13,12 @@ import org.rust.lang.core.macros.ExpansionResult
 import org.rust.lang.core.psi.RsElementTypes.UNION
 import org.rust.lang.core.psi.RsStructItem
 import org.rust.lang.core.psi.RsPsiImplUtil
+import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.stubs.RsStructItemStub
 import org.rust.lang.core.types.RsPsiTypeImplUtil
 import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.type
```
Undin: Unused imports

Undin:
And it looks like highlighting of struct and enum declarations is not very useful itself because in the most cases `Copy` trait is implemented by `#[derive]` attribute and user has already known about it.
More useful feature (and I think idea of the corresponding issue) to highlight variables which use copying instead of moving. @gilescope could you comment it?

framlog:
I think both are useful since the structs or enums aren't always `Copy` even though you add `#[derive]` attribute.

Issue #1899, opened 2017-10-23
Structure view does not show template constraints


jbayardo:
As of 0.2.0.2074, impls of templated types with different trait constraints on the template will display the same, though being different. For example

```
struct A<K> { ... }

impl<K: Ord> A<K> { ... }
impl<K: Display> A<K> { ... }
```

Will show in the Structure view as `A<K>` (although in separate items). I think they should be displayed separately, perhaps something like

```
A<K : Ord> (dropdown)
A<K : Display> (dropdown)
```

matklad:
The relevant function is `org.rust.ide.presentation.UtilsKt#getPresentationForStructure`.

PR #2123, opened 2017-12-12
StructureView display of template impls

Filippok:
Fulfilled #1899 request.

### DIFF
--- a/src/main/kotlin/org/rust/ide/presentation/Utils.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/Utils.kt
@@ -12,7 +12,6 @@ import org.rust.ide.icons.addVisibilityIcon
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.type
-import kotlin.collections.ArrayList
 
 
 fun getPresentation(psi: RsElement): ItemPresentation {
@@ -62,26 +61,19 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
 }
 
 private fun buildTemplateImplTypeName(psi: RsImplItem): String? {
-    val paramsList = ArrayList<String>()
+    val paramsList = mutableListOf<String>()
     val baseName = psi.typeReference?.baseType?.name ?: return null
-    for(typeParameter in psi.typeParameters) {
+    for (typeParameter in psi.typeParameters) {
         val name = typeParameter.name ?: continue
         val typeBounds = typeParameter.typeParamBounds?.childrenOfType<RsPolybound>()
-        if(typeBounds != null) {
-            for(bound in typeBounds) {
-                val value = bound.bound.traitRef?.path?.referenceName
-                if(value != null) {
-                    val builder = StringBuilder()
-                    builder.append(name)
-                    builder.append(": ")
-                    builder.append(value)
-                    paramsList.add(builder.toString())
-                    break
-                }
-            }
+        val boundsList = mutableListOf<String>()
+        if (typeBounds != null) {
+            typeBounds.mapNotNullTo(boundsList) { it.bound.traitRef?.path?.referenceName }
         }
+        val boundsString = boundsList.joinToString(separator = " + ")
+        paramsList.add("$name: $boundsString")
     }
-    if(paramsList.isEmpty()) return null
+    if (paramsList.isEmpty()) return null
     val paramsString = paramsList.joinToString(prefix = "<", postfix = ">", separator = ", ")
     return "$baseName$paramsString"
 }
--- a/src/main/kotlin/org/rust/ide/presentation/Utils.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/Utils.kt
@@ -61,31 +61,16 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
 }
 
 private fun buildTemplateImplTypeName(psi: RsImplItem): String? {
-
-    val populatePolyboundsMap = fun(type: String, typeBounds: List<RsPolybound>?, map: LinkedHashMap<String, MutableList<String>>): Unit {
-        val boundsList = mutableListOf<String>()
-        if (typeBounds != null) {
-            typeBounds.mapNotNullTo(boundsList) { it.bound.traitRef?.path?.referenceName }
-        }
-        if (map.containsKey(type))
-            map[type]!!.addAll(boundsList)
-        else
-            map.put(type, boundsList)
-    }
-
     val boundsMap = LinkedHashMap<String, MutableList<String>>()
     val baseName = psi.typeReference?.baseType?.name ?: return null
     for (typeParameter in psi.typeParameters) {
         val name = typeParameter.name ?: continue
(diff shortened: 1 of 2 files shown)


kumbayo:
I would suggest having a commit message with information on what this commit does.

Can you please also use the auto format action for the code?
It is accessible via Ctrl-Shift-A "Reformat code Ctrl+Alt+L"

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -60,13 +61,40 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
...
+    val paramsList = ArrayList<String>()
+    val baseName = psi.typeReference?.baseType?.name ?: return null
+    for(typeParameter in psi.typeParameters) {
+        val name = typeParameter.name ?: continue
+        val typeBounds = typeParameter.typeParamBounds?.childrenOfType<RsPolybound>()
+        if(typeBounds != null) {
+            for(bound in typeBounds) {
+                val value = bound.bound.traitRef?.path?.referenceName
+                if(value != null) {
+                    val builder = StringBuilder()
```
kumbayo: This can probably be simplified to `paramsList.add("$name: $value")`

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -12,6 +12,7 @@ import org.rust.ide.icons.addVisibilityIcon
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.type
+import kotlin.collections.ArrayList
```
kumbayo: For me this shows a "unresolved reference" error in IntelliJ but compiles fine in Gradle.
Maybe you can use `mutableListOf<String>()` below which should work the same way.

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -60,13 +61,40 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
...
+        if(typeBounds != null) {
+            for(bound in typeBounds) {
+                val value = bound.bound.traitRef?.path?.referenceName
+                if(value != null) {
+                    val builder = StringBuilder()
+                    builder.append(name)
+                    builder.append(": ")
+                    builder.append(value)
+                    paramsList.add(builder.toString())
+                    break
```
kumbayo: Is it intentional that the code here does not handle multiple type bounds?
This code
```rust
impl<K: Display + Ord> A<K> {
    pub fn ccc() {}
}
```
shows up the same way as
```rust
impl<K: Display> A<K> {
    pub fn aaa() {}
}
```

Filippok:
Thanks for review, @kumbayo. I'll fix issues and reopen request.

kumbayo:
I wonder what should be displayed when there are no type bounds like
```rust
impl<K> A<K> {
    pub fn ddd() {}
}
```
Now it displays `A<K: >`.  I have not thought about it deeply but I would have expect `A<K>` instead.

No matter which way we choose, there should probably be a test for it to notice if it changes unexpectedly ;-)

kumbayo:
Sorry for not catching this earlier but this does not handle the case where the type constraint is in a `where` clause (which should behave exactly the same functionality wise).

These all look the same just like before the change.
```rust
impl<K> A<K> where K: Display {
    pub fn aaa(&self) {}
}
impl<K> A<K> where K: Ord {
    pub fn bbb(&self) {}
}
impl<K> A<K> where K: Display + Ord {
    pub fn ccc(&self) {}
}

//looks strange but valid
impl<K> A<K> where {
    pub fn ddd(&self) {}
}
//looks strange but valid
impl<K> A<K> where K: {
    pub fn eee(&self) {}
}
```

Filippok:
Oh, my bad. Completely forgot about where clauses.

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -61,19 +61,39 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
 }
 
 private fun buildTemplateImplTypeName(psi: RsImplItem): String? {
-    val paramsList = mutableListOf<String>()
-    val baseName = psi.typeReference?.baseType?.name ?: return null
-    for (typeParameter in psi.typeParameters) {
-        val name = typeParameter.name ?: continue
-        val typeBounds = typeParameter.typeParamBounds?.childrenOfType<RsPolybound>()
+
+    val populatePolyboundsMap = fun(type: String, typeBounds: List<RsPolybound>?, map: LinkedHashMap<String, MutableList<String>>): Unit {
```
kumbayo: I wonder why `populatePolyboundsMap` needs to be a val of a fun instead of a normal fun?
Also `RsTypeParameter.bounds` seems to already implement all logic to extract bounds from both places. Can we reuse that instead of recoding it or is there a reason why it can't be used?

This would allow to simplify the code to something like:
```rust
    val boundsMap = LinkedHashMap<String, MutableList<String>>()
    val baseName = psi.typeReference?.baseType?.name ?: return null
    for (typeParameter in psi.typeParameters) {
        val name = typeParameter.name ?: continue

        val boundsList = mutableListOf<String>()
        typeParameter.bounds.mapNotNullTo(boundsList) { it.bound.traitRef?.path?.referenceName }
        if (boundsMap.containsKey(name))
            boundsMap[name]!!.addAll(boundsList)
        else
            boundsMap.put(name, boundsList)
    }
```
And in my (very quick) tests it yielded the same results

Filippok: ```populateBoundsMap``` is used only in ```buildTemplateImplTypeName``` function and nowhere else, so I don't see any point in introducing it in higher level scope.

Thanks for pointing to ```RsTypeParameter.bounds``` - I will simplify code now. Also, ```populateBoundsMap``` becomes obsolete and will be removed.

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -60,13 +60,42 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
...
+
 private fun presentableName(psi: RsElement): String? = when (psi) {
     is RsNamedElement -> psi.name
     is RsImplItem -> {
         val typeName = psi.typeReference?.type?.toString()
         val traitName = psi.traitRef?.path?.referenceName
+        val templateImplName = buildTemplateImplTypeName(psi)
         when {
             typeName == null -> null
+            templateImplName != null -> templateImplName
```
Undin: It doesn't handle trait implementations with type params for some type.
For instance, in this case
```rust
trait Foo {}
struct S<T>(T);
impl<T> Foo for S<T> {}
```
it shows `S<T>` instead of `Foo for S<T>`
And maybe move all code associated with `RsImplItem` into separate function?

Undin:
@Filippok I've fixed comments and merge this changes in #2428 
Thanks for the contribution!

Filippok:
Oh, thanks, @Undin. Actually, I was too busy for the last couple of months, so eventually, I forgot about this PR :-/

PR #2125, opened 2017-12-13
GRD: update gradle and kotlin

Undin:
* gradle 4.4
* kotlin 1.2.0


build.gradle.kts
```diff
@@ -31,7 +31,7 @@ val channel = prop("publishChannel")
 
 plugins {
     idea
-    kotlin("jvm") version "1.1.4"
+    kotlin("jvm") version "1.2.0"
```
matklad: We'll need to switch apiVersion and langaugeVersion to 1.1 once we drop 2017.2

Perhaps we can even switch langaugeVersion now?

Undin: Looks like IDEA 2017.2 has kotlin 1.1 libraries so we can't switch to 1.2 apiVersion

Some warnings from logs if you try to do it:

<details><summary>log</summary>
<p>

```
w: Runtime JAR files in the classpath have the version 1.1, which is older than the API version 1.2. Consider using the runtime of version 1.2, or pass '-api-version 1.1' explicitly to restrict the available APIs to the runtime of version 1.1. You can also pass '-language-version 1.1' instead, which will restrict not only the APIs to the specified version, but also the language features. Alternatively, you can use '-Xskip-runtime-version-check' to suppress this warning
w: /Users/warrior/Programming/intellij-rust/deps/ideaIC-2017.2.5/lib/kotlin-reflect.jar: Runtime JAR file has version 1.1 which is older than required for API version 1.2
w: /Users/warrior/Programming/intellij-rust/deps/ideaIC-2017.2.5/lib/kotlin-runtime.jar: Runtime JAR file has version 1.1 which is older than required for API version 1.2
w: /Users/warrior/Programming/intellij-rust/src/main/kotlin/org/rust/toml/CargoCrateDocLineMarkerProvider.kt: (49, 13): Name shadowed: version
w: Runtime JAR files in the classpath have the version 1.1, which is older than the API version 1.2. Consider using the runtime of version 1.2, or pass '-api-version 1.1' explicitly to restrict the available APIs to the runtime of version 1.1. You can also pass '-language-version 1.1' instead, which will restrict not only the APIs to the specified version, but also the language features. Alternatively, you can use '-Xskip-runtime-version-check' to suppress this warning
w: /Users/warrior/Programming/intellij-rust/deps/ideaIC-2017.2.5/lib/kotlin-reflect.jar: Runtime JAR file has version 1.1 which is older than required for API version 1.2
w: /Users/warrior/Programming/intellij-rust/deps/ideaIC-2017.2.5/lib/kotlin-runtime.jar: Runtime JAR file has version 1.1 which is older than required for API version 1.2
```

</p>
</details>

Issue #1730, opened 2017-09-13
Quick fixes on type mismatch


vlad20012:
A very basic type checking just merged #1727
Now it's possible to attach quick fixes to a `type mismatch` annotations (on [this line](https://github.com/intellij-rust/intellij-rust/blob/b42286d6b582bffa3b103cf05a9d0ad35d70b101/src/main/kotlin/org/rust/lang/core/types/RsDiagnostic.kt#L25)). I can imagine the following quick fixes (we assume that `A` is actual type, `B` is expected and so `a` is a var of type `A`. e.g. `let a: A = ...; let _: B = a;`):
* If both types are numerics, we can suggest add a cast `a as B`.
* We can have a set of special cases like `str.to_string()`/`String.as_str()`
* We can try to find `impl From<A> for B`, and if it is exists, suggest `B::from(a)`
* `From` is not the one type conversion trait. We can also deal with `ToOwned`/`AsRef`/`AsMut` or even `TryFrom`/`FromStr`
* We can try to deref `*a` to match required type
* We can also get more references `&a`
* And any combinations of these, like `&*a` or even `&S::from(*a)`

You can find examples of quick fixes in the [`org.rust.ide.annotator.fixes`](https://github.com/intellij-rust/intellij-rust/tree/master/src/main/kotlin/org/rust/ide/annotator/fixes) package.

oleg-semenov:
@vlad20012 I think i can try this. For the Trait bullet points, I think I can use org.rust.lang.core.resolve.ImplLookup and also for check the deref. Does that sound about right? 

For the Refactorings  those are not implemented somewhere, or are they?

Is it ok to have a couple of pull requests for this issue?

vlad20012:
1. Yep, `ImplLookup` is what you need. `ImplLookup.select()` allows you to check that some trait is implemented for some type, e.g.
```kotlin
val fromTrait = items.findCoreItem("convert::From") as? RsTraitItem ?: return
select(TraitRef(expectedTy, fromTrait.withSubst(actualTy)))RsTraitItem)?.withSubst(actualTy)))
```
2. No :)
3. Yes, definitely yes. I even suggest to implement each item as a separate PR.

oleg-semenov:
Thanks @vlad20012, no prob, hope you feel better now. Cool, I'll add the simple cases first.

oleg-semenov:
[...]

PR #2130, opened 2017-12-17
ANN: add quick fix on numeric type mismatch

oleg-semenov:
This commit adds quick-fix for the numeric types mismatch,
which fulfills the first bullet-point of the issue:

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -19,9 +19,12 @@ import org.rust.ide.annotator.fixes.*
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyInfer
+import org.rust.lang.core.types.ty.TyNumeric
 import org.rust.lang.refactoring.implementMembers.ImplementMembersFix
 import org.rust.lang.utils.RsErrorCode.*
 import org.rust.lang.utils.Severity.*
+import org.rust.stdext.buildList
 
 sealed class RsDiagnostic(
     val element: PsiElement,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -79,7 +88,7 @@ sealed class RsDiagnostic(
     ) : RsDiagnostic(element) {
         override fun prepare(): PreparedAnnotation {
             val block = element.ancestorStrict<RsBlock>()?.parent
-            val fixes = mutableListOf<LocalQuickFix>(SurroundWithUnsafeFix(element as RsExpr))
+                val fixes = mutableListOf<LocalQuickFix>(SurroundWithUnsafeFix(element as RsExpr))
```
vlad20012: Hmm?

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -35,12 +37,19 @@ sealed class RsDiagnostic(
...
-        )
+        override fun prepare(): PreparedAnnotation {
+            val fixes = mutableListOf<LocalQuickFix>()
+            if (expectedTy is TyNumeric && isActualTyNumeric()) fixes.add(AddAsTyFix(element, expectedTy))
+            return PreparedAnnotation(
+                ERROR,
+                E0308,
+                "mismatched types",
+                expectedFound(expectedTy, actualTy),
+                fixes = fixes
```
vlad20012: Try to use our `buildList` function here, it may look much cleaner.

vlad20012:
Everithing (except some formatting/style :)) looks fine! Tests in the right place and written as expected. Thanks!

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
oleg-semenov: done

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
oleg-semenov: done

vlad20012:
bors r+
Thanks!

PR #2131, opened 2017-12-18
ACT: implement 'goto type declaration' action

### LINKED ISSUE #2102: Implement *Go to type declaration* (Ctrl-Shift-B)
I'm used to it from the Scala plugin and I think it is quite useful. 

My guess is it shouldn't be too difficult to implement, as both of the sub-actions (find the type of symbol, go to declaration) are already implemented.

Undin:
Closes #2102

### DIFF
--- a/src/main/kotlin/org/rust/ide/navigation/goto/RsTypeDeclarationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/navigation/goto/RsTypeDeclarationProvider.kt
@@ -1,0 +1,52 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.navigation.goto
+
+import com.intellij.codeInsight.navigation.actions.TypeDeclarationProvider
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsConstant
+import org.rust.lang.core.psi.RsFieldDecl
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsPatBinding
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/navigation/goto/RsTypeDeclarationProvider.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.rust.lang.core.types.type
+
+class RsTypeDeclarationProvider : TypeDeclarationProvider {
+
+    override fun getSymbolTypeDeclarations(element: PsiElement): Array<PsiElement>? {
+        val type = when (element) {
+            is RsFunction -> element.retType?.typeReference?.type
+            is RsFieldDecl -> element.typeReference?.type
+            is RsConstant -> element.typeReference?.type
+            is RsPatBinding -> element.type
```
matklad: Hm, shouldn't this handle arbitrary RsExpressions?

Undin: Looks like no
If I understand correctly corresponding action looks for only references, named elements and lookup elements(https://github.com/JetBrains/intellij-community/blob/master/platform/lang-impl/src/com/intellij/codeInsight/navigation/actions/GotoTypeDeclarationAction.java#L112-L116)
I really don't know which elements are lookup elements, but it seems that arbitrary `RsExpr` don't belong to these groups

Undin:
@matklad Do you have any other comments?

Undin:
bors=matklad

Undin:
bors r=matklad

PR #2142, opened 2017-12-22
[COMP] Refactored LookupElements.kt

mhaessig:
Performed proposed refactoring from PR #1901 and thus finally resolving issue #1888.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -22,43 +22,91 @@ const val KEYWORD_PRIORITY = 10.0
 private const val MACRO_PRIORITY = -0.1
 
 fun createLookupElement(element: RsElement, scopeName: String): LookupElement {
-    val base = LookupElementBuilder.create(element, scopeName)
-        .withIcon(if (element is RsFile) RsIcons.MODULE else element.getIcon(0))
+    val base = element.getLookupElementBuilder(scopeName)
+        .withInsertHandler { context: InsertionContext, _ -> getInsertHandler(element, scopeName, context) }
 
-    return when (element) {
+    if (element is RsMacroDefinition) return base.withPriority(MACRO_PRIORITY)
+
+    return base
+}
+
+
+fun LookupElementBuilder.withPriority(priority: Double): LookupElement =
+    PrioritizedLookupElement.withPriority(this, priority)
+
+private fun RsElement.getLookupElementBuilder(scopeName: String): LookupElementBuilder {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -22,113 +22,131 @@ const val KEYWORD_PRIORITY = 10.0
...
         } else {
             base
         }
 
         is RsConstant -> base
-            .withTypeText(element.typeReference?.text)
-            .withInsertHandler({ context: InsertionContext, _ -> appendSemicolon(context) })
-        is RsFieldDecl -> base.withTypeText(element.typeReference?.text)
-        is RsTraitItem -> base.withInsertHandler({ context: InsertionContext, _ -> appendSemicolon(context) })
+            .withTypeText(this.typeReference?.text)
```
Undin: Explicit `this` here and below can be dropped

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -22,113 +22,131 @@ const val KEYWORD_PRIORITY = 10.0
 private const val MACRO_PRIORITY = -0.1
 
 fun createLookupElement(element: RsElement, scopeName: String): LookupElement {
-    val base = LookupElementBuilder.create(element, scopeName)
-        .withIcon(if (element is RsFile) RsIcons.MODULE else element.getIcon(0))
+    val base = element.getLookupElementBuilder(scopeName)
+        .withInsertHandler({ context: InsertionContext, _ ->
+            run {
```
Undin: It seems that if you drop explicit `()` at `withInsertHandler` call and `run` function, code would be a little bit clearer

Undin: And maybe extract this lambda as separate function?

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -22,113 +22,131 @@ const val KEYWORD_PRIORITY = 10.0
...
+                        context.document.insertString(context.selectionEndOffset, "!$parens")
+                        EditorModificationUtil.moveCaretRelatively(context.editor, 2)
+                    }
+
+                }
+            }
+        })
+
+    if (element is RsMacroDefinition) {
+        return base.withPriority(MACRO_PRIORITY)
```
Undin: We usually use `if (condition) return value` pattern in such cases

mhaessig:
Incorporated changes suggested in the review.

Undin:
bors r+
Thanks!

PR #2143, opened 2017-12-22
Handle type aliases used in struct literals

kumbayo:
We resolve all type aliases in the struct literal and use the actual struct type as result.


Undin:
bors r+
Thanks!

PR #2150, opened 2017-12-24
refine type inference for blocks

framlog:
fixup #2063 

More specifically, it forces the inference mechanism to take `return` clauses into consideration.


kumbayo:
This seems to include 3 of my commits (with you as the comitter, probably due to `git rebase -i`) which are already merged in master.
Can you please rebase this branch on top of origin/master
or drop my patches from your branch (using for example `git rebase -i`)?

src/main/resources/org/rust/ide/colors/highlighterDemoText.rs
```diff
@@ -21,6 +21,11 @@ struct <STRUCT>Object</STRUCT><<TYPE_PARAMETER>T</TYPE_PARAMETER>> {
     <FIELD>fields</FIELD>: <STRUCT>HashMap</STRUCT><<TYPE_PARAMETER>T</TYPE_PARAMETER>, <PRIMITIVE_TYPE>u64</PRIMITIVE_TYPE>>
 }
 
+<ATTRIBUTE>#[derive(Copy)]</ATTRIBUTE>
```
kumbayo: This change probably belongs to the other merge request

src/test/kotlin/org/rust/lang/core/type/RsDivergingTypeInferenceTest.kt
```diff
@@ -1,45 +0,0 @@
...
- * found in the LICENSE file.
- */
-
-package org.rust.lang.core.type
-
-
-class RsDivergingTypeInferenceTest : RsTypificationTestBase() {
-    fun `test if with all branches diverging`() = testExpr("""
-        fn main() {
-            let x = if true { return; } else { return; };
```
kumbayo: This now infers to `unknown` instead of `!`.
Is this intentional?

Also you seem to have changed the test in a way that this change in behavior is not visible...
Is it intentional that the tests now test something different (the whole block instead of only `x`)?

framlog: Yes. I think rustc can't infer the type of `if true { return; } else { return; }`

kumbayo: I am pretty sure rust can infer the type as `!` as the following code compiles
```rust
fn main() {
    let x = if true { return; } else { return; };
    let y: String = x; //only possible because x is !
}
```

framlog: Well, I still don't think rustc could infer a proper type for `if true { return; } else { return; }` by itself after some tests.

I tried run `rustc test.rs -Z save-analysis` and read the dumped json then, which gave me that the type of `x` is `std::string::String`, which's what the `y`'s type ascription says. And after I added `feature(never_type)` and set the type ascription of `x` to `!`, the type turned into `!` in the dumped json file.

And I got an empty funtion when I tried to dump mir of those code.

framlog: I start to think `x` should be infer to `!` after reading rfc 1216 and its related comments. I'll fix this

framlog: fixed

vlad20012:
I'm very sorry, but I've solved it in another way in #2201. Your solution is also working, but when we talking about Rust type inference system, it is very important to keep it in accordance with rustc implementation, because it is enough hacky and we want exactly the same behavior in our IDE as in the compiler.

Also, the issue #2063 was assigned to me, so it was a kind of mistake to start working on that without asking me or any discussion there.

Anyway, thanks for your great work! And sorry again.

PR #2167, opened 2017-12-30
Better handle multiple attributes with the same name

kumbayo:
Also add some comments to make the code easier to understand.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -99,17 +99,16 @@ class QueryAttributes(
             .singleOrNull()
 
     val langAttribute: String?
-        get() = getStringAttribute("lang")
+        get() = getStringAttributes("lang").firstOrNull()
 
     val deriveAttributes: Sequence<RsMetaItem>
         get() = attrsByName("derive")
 
-    private fun getStringAttribute(attributeName: String): String? = attrByName(attributeName)?.value
+    // `#[attributeName = "Xxx"]`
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -94,17 +99,21 @@ class QueryAttributes(
...
+    private fun getStringAttributes(attributeName: String): Sequence<String?> = attrsByName(attributeName).map { it.value }
 
     val metaItems: Sequence<RsMetaItem>
         get() = attributes.mapNotNull { it.metaItem }
 
-    private fun attrByName(name: String) = metaItems.find { it.referenceName == name }
+    /**
+     * Get a sequence of all attributes named [name]
+     */
+    private fun attrsByName(name: String) = metaItems.filter { it.referenceName == name }
```
Undin: Let's add return type for this function to have explicit returning type for all functions

src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt
```diff
@@ -539,7 +539,9 @@ class RsCompletionTest : RsCompletionTestBase() {
     """)
 
     fun `test hidden macro is hidden in other module`() = checkNoCompletion("""
+        #[doc="No problems with"]
         #[doc(hidden)]
+        #[doc="explicit docs"]
```
Undin: Let's create separate test with several `#[doc]` attributes

kumbayo: Done, i just created a separate test for this instead of modifying the existing test.

Undin:
bors r+
Thanks!

PR #2169, opened 2017-12-31
INT: Improve auto import

### LINKED ISSUE #2159: Automatic import importing wrong things
_From @Jiehong on December 27, 2017 21:51_

Hi,

With the new semi-auto import feature being there (amazing! thanks!), I tried a quick piece of code, and here are my results:

``` rust
fn sqr(number: i64) -> i64 {
    thread::sleep(Duration::from_secs(2));
    return number * number;
}
```
In this code, you'll be able to try to import `thread` from line `thread::sleep`. If you do so, the following line will be added:
``` rust
use std::thread::thread;
```

While it should be:
``` rust
use std::thread;
```

Once you've fixed this import, it seems that you can try to import `Duration`, but using `Alt+Enter` only offers you to add parenthesis around the expression.

It should offer to import the following line:

``` rust
use std::time::Duration;
```

IDE version: Intellij IDE Community 2017.3.2

_Copied from original issue: intellij-rust/intellij-rust.github.io#57_

Undin:
Fixes #2159

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
@@ -13,10 +13,7 @@ import org.rust.cargo.project.workspace.CargoWorkspace
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.cargo.util.AutoInjectedCrates
 import org.rust.ide.intentions.RsElementBaseIntentionAction
-import org.rust.lang.core.psi.RsExternCrateItem
-import org.rust.lang.core.psi.RsPath
-import org.rust.lang.core.psi.RsPsiFactory
-import org.rust.lang.core.psi.RsUseItem
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.stubs.index.RsNamedElementIndex
 import org.rust.lang.core.types.ty.TyPrimitive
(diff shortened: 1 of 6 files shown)


Undin:
@matklad r?

src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
```diff
@@ -38,8 +36,16 @@ class ImportNameIntention : RsElementBaseIntentionAction<ImportNameIntention.Con
...
         val candidates = RsNamedElementIndex.findElementsByName(project, basePath.referenceName)
             .filterIsInstance<RsQualifiedNamedElement>()
+            .filter { basePath != path || !(it is RsMod || it is RsModDeclItem || it.parent is RsMembers) }
             .mapNotNull {
                 val info = (it as? RsVisible)?.canBeImported(pathSuperMods) ?: return@mapNotNull null
+                // check that result after import can be resolved
+                // if no, don't add it in candidate list
+                if (basePath != path) {
+                    val pkg = info.target.pkg
+                    val fullPath = "${pkg.name}::${info.crateRelativePath.removeSuffix(basePath.text)}${path.text}"
```
matklad: Yeah, so here's where we start to use `QualifiedNameOwner` in a precise way :) It is ok, but we need to make sure that when we add `trait::function` qualified names (which seems reasonable), we won't try to use them for importing.

src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
```diff
@@ -38,8 +36,16 @@ class ImportNameIntention : RsElementBaseIntentionAction<ImportNameIntention.Con
...
+                    resolveStringPath(fullPath, pkg.workspace, project) ?: return@mapNotNull null
```
matklad: Hm, an excellent idea! I think we need to go a bit further though and to try to actually insert a fake use declaration and resolve it. This is what `resolveStringPath` does under the hood actually, but it always inserts the path into the root module, and we preferably should use the current module instead.

matklad:
LGTM!

Undin:
bors r=matklad

PR #2166, opened 2017-12-29
WIP: Parse and evaluate #[cfg(...)] attributes during resolving

kumbayo:
This should not be merged yet as it is still work in progress.
I plan to submit this patches as a series of smaller PRs.

CC #1191
CC #461

PR #2172, opened 2018-01-02
Stub and PSI improvements related to inner and outer attributes

kumbayo:
Avoid loading the PSI when reading inner/outer attributes.
Correctly allow and attach comments and inner/outer attributes on various items.

This is in preparation for #2166

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
@@ -21,7 +21,7 @@ abstract class RsForeignModItemImplMixin : RsStubbedElementImpl<RsPlaceholderStu
     constructor(stub: RsPlaceholderStub, elementType: IStubElementType<*, *>) : super(stub, elementType)
 
     override val outerAttrList: List<RsOuterAttr>
-        get() = childrenOfType()
+        get() = PsiTreeUtil.getStubChildrenOfTypeAsList(this, RsOuterAttr::class.java)
 
     override val isPublic: Boolean get() = false // visibility does not affect foreign mods
 
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
@@ -22,10 +22,10 @@ abstract class RsForeignModItemImplMixin : RsStubbedElementImpl<RsPlaceholderStu
     constructor(stub: RsPlaceholderStub, elementType: IStubElementType<*, *>) : super(stub, elementType)
 
     override val innerAttrList: List<RsInnerAttr>
-        get() = PsiTreeUtil.getStubChildrenOfTypeAsList(this, RsInnerAttr::class.java)
+        get() = stubChildrenOfType()
 
     override val outerAttrList: List<RsOuterAttr>
-        get() = PsiTreeUtil.getStubChildrenOfTypeAsList(this, RsOuterAttr::class.java)
+        get() = stubChildrenOfType()
 
     override val isPublic: Boolean get() = false // visibility does not affect foreign mods
 
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
@@ -10,6 +10,7 @@ import com.intellij.psi.stubs.IStubElementType
 import com.intellij.psi.util.PsiTreeUtil
 import org.rust.lang.core.macros.ExpansionResult
 import org.rust.lang.core.psi.RsForeignModItem
+import org.rust.lang.core.psi.RsInnerAttr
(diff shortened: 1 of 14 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
```diff
@@ -21,7 +21,7 @@ abstract class RsForeignModItemImplMixin : RsStubbedElementImpl<RsPlaceholderStu
     constructor(stub: RsPlaceholderStub, elementType: IStubElementType<*, *>) : super(stub, elementType)
 
     override val outerAttrList: List<RsOuterAttr>
-        get() = childrenOfType()
+        get() = PsiTreeUtil.getStubChildrenOfTypeAsList(this, RsOuterAttr::class.java)
```
Undin: Maybe extract it as separate function like `childrenOfType`

src/test/kotlin/org/rust/lang/core/stubs/RsStubAccessTest.kt
```diff
@@ -45,6 +48,15 @@ class RsStubAccessTest : RsTestBase() {
         processStubsWithoutAstAccess<RsElement> { it.reference }
     }
 
+    fun `test inner outer attributes does not need ast`() {
+        processStubsWithoutAstAccess<RsElement> { element ->
+            if (element is RsInnerAttributeOwner && element !is RsFile)
+                element.innerAttrList
```
Undin: Let's use braces for blocks even with one line action except blocks with single `return` expr and `if (condition) expr1 else expr2` expressions

src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
kumbayo: Can we just change `PsiElement.childrenOfType` to call `getStubChildrenOfTypeAsList` instead of `getChildrenOfTypeAsList` ? Or is there any drawback to such a change?

src/test/kotlin/org/rust/lang/core/stubs/RsStubAccessTest.kt
kumbayo: Done, i also removed 2 unnecessary implementations of `val outerAttrList: List<RsOuterAttr>`

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -121,7 +122,7 @@ abstract class RsFunctionImplMixin : RsStubbedNamedElementImpl<RsFunctionStub>,
     override val crateRelativePath: String? get() = RsPsiImplUtil.crateRelativePath(this)
 
     final override val innerAttrList: List<RsInnerAttr>
-        get() = block?.innerAttrList.orEmpty()
+        get() = PsiTreeUtil.getStubChildrenOfTypeAsList(this, RsInnerAttr::class.java)
```
kumbayo: This change is actually wrong.
The RsInnerAttr are children of the RsBlock.
But the new code searches them inside the RsFunction

src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
Undin: No, we can't replace `childrenOfType` by `getStubChildrenOfTypeAsList` because if element is `StubBasedPsiElement`, `getStubChildrenOfTypeAsList` returns only inheritors of `StubBasedPsiElement` while `childrenOfType` returns all children.
For example, if `element` is `RsArrayType`, result of `PsiTreeUtil.getStubChildrenOfTypeAsList(element, RsElement::class.java)` doesn't contain expr child because `RsExpr` is not `StubBasedPsiElement`

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
kumbayo: Fixed, i also added tests for inner/outer attribute owners and some additional fixes for problems i found.

src/main/kotlin/org/rust/lang/core/psi/ext/RsForeignModItem.kt
kumbayo: Ah OK. I have created a wrapper `PsiElement.stubChildrenOfType()`

src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
```diff
@@ -24,6 +25,10 @@ val RsItemsOwner.itemsAndMacros: Sequence<RsElement>
                     val stub = stub
                     if (stub != null) return@run stub.childrenStubs
                 }
+                is RsBlock -> {
+                    val stub = stub
+                    if(stub != null) return@run stub.childrenStubs
```
Undin: Formatting

Undin:
Thanks!
bors r+

PR #2179, opened 2018-01-08
ART: svg icons

Nilera:
Redrew icons into SVG format, and move `rust-file.ai` into *art/icons* directory to reduce the size of output jar file.


Undin:
bors r+
Thanks!

Issue #2185, opened 2018-01-10
Correctly extract associated function in trait item


Undin:
We can extract associated function in `impl Ty { ... }` block but not for trait item.

```rust
trait Foo {
    fn foo(&self) {
        let b = 1;
        <selection>println!("{}", b);</selection>
    }
}
```

**Expected**:
```rust
trait Foo {
    fn foo(&self) {
        let b = 1;
        Self::bar(b);
    }

    fn bar(b: i32) {
        println!("{}", b);
    }
}
```

**Actual**:
```rust
trait Foo {
    fn foo(&self) {
        let b = 1;
        bar(b);  // error[E0425]: cannot find function `bar` in this scope
    }

    fn bar(b: i32) {
        println!("{}", b);
    }
}
```

Also, looks like we can use `Self` type for all associated function extraction not only in trait item to unify implementation of these cases

t-kameyama:
I will work on this issue.

Undin:
Fixed in #2194

PR #2194, opened 2018-01-13
TYPE: Correctly extract associated function in trait item #2185

t-kameyama:
Fix for #2185


src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionHandlerAction.kt
```diff
@@ -61,7 +62,12 @@ class RsExtractFunctionHandlerAction(
         stmt += if (firstParameter != null && firstParameter.name.endsWith("self") && firstParameter.type == null) {
             "self.${config.name}(${config.argumentsText})"
         } else {
-            val type = (config.containingFunction.owner as? RsFunctionOwner.Impl)?.impl?.typeReference?.text
+            val owner = config.containingFunction.owner
+            val type = when (owner) {
+                is RsFunctionOwner.Impl -> owner.impl.typeReference?.text
+                is RsFunctionOwner.Trait -> TyTypeParameter.Self.name
```
Undin: I think it's better to use own "Self" literal instead of `TyTypeParameter.Self.name` to avoid unexpected changes while refactoring of `TyTypeParameter`

Undin:
Thanks!
bors r+

PR #2195, opened 2018-01-14
Improve typification and resolving in match guard clauses and struct update syntax

kumbayo:
Currently we do not correctly typify and resolve paths (function calls or variables)
inside match guard clauses and in the struct literals using struct update syntax.


Undin:
Looks good!
Let's add resolve test for "struct literal update syntax".
I mean resolve something like `S::new()` in  
```rust
struct S {
    f1: u32,
    f2: u8,
}

impl S {
    fn new() -> Self {
        S { f1: 0, f2: 0 }
    }
}

fn main() {
    let a = S { f1: 1, ..S::new() };
}
```

Undin:
bors r+
Thanks!

PR #2196, opened 2018-01-14
ANN: add quick fix for type mismatch when impl ToOwned for A exists and B is Owned type of A

oleg-semenov:
impl `ToOwned` for A exists and B is `Owned` type of A
where A is actual type and B is expected type,
which fulfills part of the forth bullet-point of the issue:

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -21,6 +21,7 @@ import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.resolve.StdKnownItems
 import org.rust.lang.core.resolve.withSubst
+import org.rust.lang.core.types.BoundElement
 import org.rust.lang.core.types.TraitRef
 import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.ty.TyInfer
@@ -51,8 +52,15 @@ sealed class RsDiagnostic(
                 fixes = buildList {
                     if (expectedTy is TyNumeric && isActualTyNumeric()) {
                         add(AddAsTyFix(element, expectedTy))
-                    } else if (isFromActualImplForExpected(element)) {
-                        add(ConvertToTyUsingFromTraitFix(element, expectedTy))
+                    } else  if (element is RsElement){
+                        val items = StdKnownItems.relativeTo(element)
+                        val lookup = ImplLookup(element.project, items)
+                        if (isFromActualImplForExpected(items, lookup)) {
+                            add(ConvertToTyUsingFromTraitFix(element, expectedTy))
+                        }
+                        if (isToOwnedImplWithExcpectedForActual(items, lookup)) {
+                            add(ConvertToOwnedTyFix(element))
+                        }
                     }
                 }
             )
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -51,22 +52,34 @@ sealed class RsDiagnostic(
...
-            val items = StdKnownItems.relativeTo(element as? RsElement ?: return false)
-            val lookup = ImplLookup(element.project, items)
+        private fun isFromActualImplForExpected(items: StdKnownItems, lookup: ImplLookup): Boolean {
             val fromTrait = items.findFromTrait() ?: return false
             return lookup.select(TraitRef(expectedTy, fromTrait.withSubst(actualTy))).ok() != null
         }
 
+        private fun isToOwnedImplWithExcpectedForActual(items: StdKnownItems, lookup: ImplLookup): Boolean {
+            val toOwnedTrait = items.findToOwnedTrait() ?: return false
+            val declaredTy = lookup.select(TraitRef(actualTy, BoundElement(toOwnedTrait))).ok()?.impl?.members
```
vlad20012: Let's use `selectProjection` here. It is more generic API for associated type selection. I can't quickly imagine a case where `selectProjection` works better right now, but it is more clear and in future it should work, for example, with associated type bindings:
```rust
fn foo<T: ToOwned<Owned=X>>(t: T) { bar(t); }
```

src/test/kotlin/org/rust/ide/annotator/fixes/ConvertToOwnedTyFixTest.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.rust.ide.inspections.RsExperimentalChecksInspection
+import org.rust.ide.inspections.RsInspectionsTestBase
+
+class ConvertToOwnedTyFixTest : RsInspectionsTestBase(RsExperimentalChecksInspection()) {
+    override fun getProjectDescriptor() = WithStdlibRustProjectDescriptor
+
+    fun `test B is Owned type of A`() = checkFixByText("Convert to owned type using `ToOwned` trait","""
+        use std::borrow::Borrow;
+
+        struct A{}
```
vlad20012: Hmm, should be a space before `{`?

src/test/kotlin/org/rust/ide/annotator/fixes/ConvertToOwnedTyFixTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        struct B{}
+
+        impl Borrow<A> for B { fn borrow(&self) -> &A { &A {} } }
+
+        impl ToOwned for A {
+            type Owned = B;
+            fn to_owned(&self) -> B { B {} }
```
vlad20012: AFAIK if you want to instantiate a struct without fields, you should just write `B` without braces.

oleg-semenov: only if `B` is also declared as `struct B;`, if it's declared as `struct B {}` it has to be instantiated with `B {}` too, don't really know why

src/test/kotlin/org/rust/ide/annotator/fixes/ConvertToOwnedTyFixTest.kt
oleg-semenov: done

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
oleg-semenov: maybe done, not sure i used the selectProjection the best/right way

vlad20012: Hmm, it can look a bit cleaner if we extract `toOwnedTrait.associatedTypesTransitively.find { it.name == "Owned" }` to a separate function (or make `findAssociatedType` in `ImplLookup` public and just use it), but it is something minor.

src/test/kotlin/org/rust/ide/annotator/fixes/ConvertToOwnedTyFixTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        struct A;
+        struct B;
+        struct C;
+
+        impl Borrow<A> for C { fn borrow(&self) -> &A { &A; } }
+
+        impl ToOwned for A {
+            type Owned = C;
+            fn to_owned(&self) -> C { C; }
```
vlad20012: Extra `;` :)

vlad20012:
:+1: Thanks!
bors r+

Issue #2187, opened 2018-01-10
Support generic parameters in "extract function" refactoring


Undin:
Current "extract function" implementation doesn't add type parameters and their bounds for generic arguments into function signature.
It would be great to support it.

```rust
fn foo<T: Default>() -> T {
    <selection>T::default()</selection>
}
```

**Expected:**
```rust
fn foo<T: Default>() -> T {
    bar()
}

fn bar<T: Default>() -> T {
    T::default()
}
```

**Actual:**
```rust
fn foo<T: Default>() -> T {
    bar()
}

fn bar() -> T {
    T::default()
}
```

t-kameyama:
I will work on this issue.

Undin:
Implemented in #2197

PR #2197, opened 2018-01-15
TYPE: Support generic parameters in "extract function" refactoring

t-kameyama:
Fix for #2187


Undin:
@t-kameyama your tests failed because of compilation of `rustfmt` fails on the latest nightly rust
I've fixed it in master so you should rebase your code on master branch to avoid this problem

t-kameyama:
Rebased

t-kameyama:
Rebased.

Undin:
I haven't reviewed code yet but I've checked how it works and current implementation doesn't handle some cases:

* trait bounds in `where` clause. Extracted function doesn't contain these type parameters
```rust
fn foo<T>(t: T) where T: Debug {
    <selection>println!("{:?}", t);</selection>
}
```
I think [bounds](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeParameter.kt#L22) function can help you

* associated function extraction in impl for generic types. It generates invalid code
```rust
impl<T> S<T> {
    fn foo() {
        <selection>println!("Hello!");</selection>
    }
}
```

* if generic bound depends on another generic parameter(s)
```rust
trait Foo<T> {}
trait Bar<T> {}
fn foo<T, F: Foo<T>, B: Bar<F>>(b: B) {
    <selection>let a = b;</selection>
}
```

t-kameyama:
Fixed

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -65,25 +66,65 @@ class RsExtractFunctionConfig private constructor(
...
+        get() {
+            val whereTypes = containingFunction.whereClause?.wherePredList?.mapNotNull {
+                it.typeReference?.type
+            } ?: return ""
+            return typeParameters
+                .mapNotNull {
+                    if (!whereTypes.contains(it.declaredType) || it.bounds.isEmpty())
+                        null
+                    else
+                        "${it.text}:${it.bounds.joinToString("+") { it.text }}"
```
Undin: Hm, it doesn't support the following case
```rust
fn foo<T: Debug>(t: T) where T: Default {
    <selection>println!("{}", t);</selection>
}
```
Probably it is because `it.bounds` returns all bounds not only from where clause

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -138,3 +179,17 @@ class RsExtractFunctionConfig private constructor(
...
+else
+    listOf(this) + typeParameterValues.values.flatMap { it.types() }
+
+private fun Ty.dependTypes(bounds: Map<Ty, List<Ty>>, last: List<Ty> = emptyList()): List<Ty> {
+    val types = bounds[this]
+    return if (types == null || types.isEmpty())
+        emptyList()
+    else
+        types + types.flatMap {
+            if (last.contains(it)) emptyList() else it.dependTypes(bounds, last + types)
```
Undin: Let's avoid list concatenation in recursive function - it creates a lot of temporary objects and copy both lists for every `+`
If you want/need to collect something in recursive function you can use the following pattern:
```kotlin
private fun collectSomething(arg): SomeCollectionInterface {
    val collection = SomeMutableCollection()

    fun collect(args) {
        // do something and put result into collection
        // call recursively if needed
    }

    collect(arg)
    return collection
}
```

t-kameyama:
Fixed

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -65,26 +66,68 @@ class RsExtractFunctionConfig private constructor(
...
+            return filtered.joinToString(separator = ",", prefix = " where ") { it.text }
+        }
+
+    private fun typeParameters(): List<RsTypeParameter> {
+        val bounds = typeParameterBounds()
+        val paramAndReturnTypes = (parameters.mapNotNull { it.type } + listOf(returnValue).mapNotNull { it?.type })
+        return containingFunction.typeParameters.filter {
+            val typeParameter = it.declaredType
+            paramAndReturnTypes.any {
+                it.types().contains(typeParameter) || it.dependTypes(bounds).contains(typeParameter)
```
Undin: Hm, it seems we can do it more efficiently - collect all types and dependent types for each element from `paramAndReturnTypes` to single set once and just check if this set contains `it.declaredType`

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -65,26 +66,68 @@ class RsExtractFunctionConfig private constructor(
...
+            val typeParams = typeParameters().map { it.declaredType }
+            if (typeParams.isEmpty()) return ""
+            val filtered = wherePredList.filter { typeParams.contains(it.typeReference?.type) }
+            if (filtered.isEmpty()) return ""
+            return filtered.joinToString(separator = ",", prefix = " where ") { it.text }
+        }
+
+    private fun typeParameters(): List<RsTypeParameter> {
+        val bounds = typeParameterBounds()
+        val paramAndReturnTypes = (parameters.mapNotNull { it.type } + listOf(returnValue).mapNotNull { it?.type })
```
Undin: `listOf(returnValue).mapNotNull { it?.type }` can be simplified to `listOfNotNull(returnValue?.type)`

t-kameyama:
@Undin Thanks for your review. I'd like to fix for your comments. Can I fix it?

Undin:
Oh, of course!)

Issue #2174, opened 2018-01-05
Auto import


Undin:
Meta issue to support auto import:
- [x] Naive support (without reexports) as intention
- [x] Support reexports
- [x] Convert intention to inspection with quick fix
- [x] Unimported methods from impls
- [x] Suggest import while typing ~~#789~~ ~~#2721~~
- [x] Import trait while trait method completion

Related issues:
* ~~#789~~
* ~~#1366~~
* ~~#1488~~
* ~~#1737~~
* ~~#2034~~
* ~~#2157~~
* ~~#2159~~
* ~~#2299~~
* ~~#2347~~ 
* ~~#2367~~
* ~~#2388~~
* ~~#2394~~
* ~~#2402~~
* #2438
* ~~#2444~~
* ~~#2721~~
* ~~#2822~~
* ~~#2889~~
* ~~#3269~~
* ~~#3275~~

Undin:
Looks like we've implemented all subissues

Issue #1737, opened 2017-09-16
Experimental auto-import intention


vlad20012:
The "real" auto-import should be a specific quick fix attached to an "unresolved reference" error annotation. The problem with this approach is that "unresolved reference" inspections produces a lot of false positives because of unsupported macros, so macros blocks this feature.

I have another idea. We can temporary implement auto-import as a regular intention. You don't see any error annotations, but just alt+enter to a name that should be imported and apply the intention. Then, when macros will be implemented, this intention will be transformed to a quick fix. So, macros now is not blocks this task and they may be implemented in parallel.

I just made a sample intention for qick start ([click](https://github.com/vlad20012/intellij-rust/commit/6848ffa9b71adf42766613f775151d69db3dafe0)). You can start from this commit. There are a large comment about how to continue.

A large discussion about auto-import also located here #1488.

---
You can find more info about how to start contributing to Intellij-Rust in [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md), and some info about our architecure in [ARCHITECTURE.md](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md). If you have any questions about this issue, you can ask them right here or in our [gitter](https://gitter.im/intellij-rust/intellij-rust). If you decided to grab this issue, please leave a comment here and put yourself in the [worklist](https://github.com/intellij-rust/intellij-rust/wiki/Worklist).

Undin:
I think it can be closed - main functionality is implemented

PR #2199, opened 2018-01-15
Initial reexports support for auto import

### LINKED ISSUE #2157: Auto import doesn't work
I just updated this plugin to version `0.2.0.2083` and saw in the changelog

> Initial support of auto import :tada: Now it is implemented as intention so you can try it via Alt+Enter at unresolved reference

and that sounded super awesome! But after updating and restarting IntelliJ IDEA (see below for meta stuff), I could not see any intention for the following code (simple, single-crate project) even though the "Import" intention is enabled:

```rust
fn simple() {
    let mut map = HashMap::new();
}
```

My Rust version is

```
rustc 1.24.0-nightly (7eb64b86c 2017-12-20)
```

and IDEA -> Help -> About gives me

```
IntelliJ IDEA 2017.3.1 (Community Edition)
Build #IC-173.3942.27, built on December 11, 2017
JRE: 1.8.0_152-release-1024-b8 amd64
Linux 4.4.0-104-generic
```

Undin:
Part of #1737 and #2174
Fixes #2157

TODO:
* support `pub use` with wildcards
* filter import candidates for the same item (for example, `std::sync::Arc` and `alloc::arc::Arc`)

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
@@ -48,7 +48,7 @@ class ImportNameIntention : RsElementBaseIntentionAction<ImportNameIntention.Con
 
         val candidates = (explicitItems + reexportedItems)
             .filter { basePath != path || !(it.item is RsMod || it.item is RsModDeclItem || it.item.parent is RsMembers) }
-            .mapNotNull { importItem -> importItem.canBeImported(pathSuperMods)?.let { ImportCandidate(importItem.item, it) } }
+            .mapNotNull { importItem -> importItem.canBeImported(pathSuperMods)?.let { ImportCandidate(importItem, it) } }
             // check that result after import can be resolved and resolved element is suitable
             // if no, don't add it in candidate list
             .filter { path.canBeResolvedToSuitableItem(project, pathMod, it.info) }
@@ -178,7 +178,7 @@ sealed class ImportItem(val item: RsQualifiedNamedElement) {
     abstract val containingCargoTarget: CargoWorkspace.Target?
 
     class ReexportedItem(
-        private val useSpeck: RsUseSpeck,
+        val useSpeck: RsUseSpeck,
         item: RsQualifiedNamedElement
     ) : ImportItem(item) {
         override val parentMod: RsMod? get() = useSpeck.containingMod
@@ -211,7 +211,7 @@ sealed class ImportInfo {
     }
 }
 
-data class ImportCandidate(val item: RsQualifiedNamedElement, val info: ImportInfo)
+data class ImportCandidate(val importItem: ImportItem, val info: ImportInfo)
 
 private fun RsItemsOwner.firstItem(): RsElement = itemsAndMacros.first()
 
--- a/src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
@@ -16,6 +16,7 @@ import org.rust.ide.intentions.RsElementBaseIntentionAction
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.stubs.index.RsNamedElementIndex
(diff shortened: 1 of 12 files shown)


Undin:
@matklad r?

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
```diff
@@ -140,16 +140,6 @@ fun processItemDeclarations(
     return false
 }
 
-private val RsUseSpeck.nameInScope: String? get() {
-    alias?.name?.let { return it }
-    val baseName = path?.referenceName ?: return null
-    if (baseName == "self") {
-        NameResolutionTestmarks.selfInGroupName.hit()
```
Undin: Unfortunately, I don't know good way to make this code public, not to remove this testmark and not to add some artificial code at the same time
Any idea?

src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
```diff
@@ -32,31 +33,48 @@ class ImportNameIntention : RsElementBaseIntentionAction<ImportNameIntention.Con
...
+
+        val parentMod = parentMod ?: return null
         val ourSuperMods = parentMod.superMods
 
         // try to find latest common ancestor module of `parentMod` and `mod` in module tree
         // we need to do it because we can use direct child items of any super mod with any visibility
         val lca = ourSuperMods.find { it in superMods }
-        val crateRelativePath = (this as? RsQualifiedNamedElement)?.crateRelativePath?.removePrefix("::") ?: return null
+        val crateRelativePath = crateRelativePath ?: return null
+        val isPublic = when (this) {
```
matklad: Can be a virtual function of `ImportItem`.

src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
```diff
@@ -32,31 +33,48 @@ class ImportNameIntention : RsElementBaseIntentionAction<ImportNameIntention.Con
...
+            .map { ImportItem.ExplicitItem(it) }
+
+        val reexportedItems = RsReexportIndex.findReexportsByName(project, basePath.referenceName)
+            .asSequence()
+            .mapNotNull {
+                val item = it.path?.reference?.resolve() as? RsQualifiedNamedElement ?: return@mapNotNull null
+                ImportItem.ReexportedItem(it, item)
+            }
+
+        val candidates = (explicitItems + reexportedItems)
```
matklad: I am a bit worried that a single item can be both inside explicit and reexported items (if it is reexported under it's original name). I don't know a way to avoid it though :(

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
matklad: I think this is fine :)

src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
Undin: It's next step of auto import improvement
I think we can use some heuristic rules, for example, if there are items from `std` and from other crates, exclude items from other crates. Or drop `mod::sub_mod::item` if there is reexport of this item in `mod`.

Undin: At least `PyCharm` uses the last rule

src/main/kotlin/org/rust/ide/intentions/import/ImportNameIntention.kt
Undin: Hm, the reason why I didn't it before is check for `RsVisibile` at the beginning of method and next smart cast in `when` expression
So I can move it into separate method of `ImportItem` but in my opinion, it will be not very gracefully
Anyway, it should be refactored to support enum variants which are not `RsVisible` at all and can't be import now at all.

matklad: Sounds reasonable!

matklad:
LGTM

Undin:
bors r=matklad

PR #2201, opened 2018-01-16
TY: unify return & tail exprs with current function return type

### LINKED ISSUE #2063: Closure type inference seems incorrect

vlad20012:
Fixes #2063


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -32,8 +32,13 @@ import org.rust.stdext.singleOrFilter
...
 
-fun inferTypesIn(fn: RsFunction): RsInferenceResult =
-    RsInferenceContext().run { preventRecursion { infer(fn) } }
+fun inferTypesIn(fn: RsFunction): RsInferenceResult {
+    val items = StdKnownItems.relativeTo(fn)
+    val lookup = ImplLookup(fn.project, items)
+    val ctx = RsInferenceContext(lookup, items)
+    lookup.ctx = ctx
+    return preventRecursion { ctx.infer(fn) }
+}
```
Undin: Looks like that you can simplify it to
```kotlin
fun inferTypesIn(fn: RsFunction): RsInferenceResult {
    val lookup = ImplLookup.relativeTo(fn)
    return preventRecursion { lookup.ctx.infer(fn) }
}
```
and make `Lookup.ctx` property readonly again

vlad20012: Done

PR #2202, opened 2018-01-16
DBG: allow to debug example binaries

### LINKED ISSUE #2200: Can't find a binary to debug
screen shot 2018-01-16 at 20 08 41
repo: https://github.com/exonum/exonum-configuration

Undin:
Don't filter example targets to allow launching them in debug mode
Fixes #2200

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -129,6 +129,9 @@ other files do.
 Consider prefixing commit with a `TAG:` which describes the area of the
 change. Common tags are:
 
+<details>
+<summary>Tags list</summary>
+
   * GRAM for changes to `.bnf` files
   * PSI for other PSI related changes
   * RES for name resolution
@@ -148,12 +151,14 @@ change. Common tags are:
   * PERF for performance optimizations
   * PRJ for project creation changes 
   * ACT for actions
+  * DBG for debugger
 
 
   * CARGO for cargo-related changes
   * GRD for build changes
   * T for tests
   * DOC for documentation
+</details>
 
 Try to keep the summary line of a commit message under 50 characters.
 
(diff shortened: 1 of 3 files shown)


Undin:
@matklad r?

matklad:
The tricky bit here is that examples can be either binary or library crates. That is, it is not sufficient to look at the kind field to determine if the thing could be debugged. 

The `cargo metadata` description of this stuff is weird for historical reasons. Here's how it looks:
https://github.com/rust-lang/cargo/blob/c5e21db8d450a46fb092cce0e634b71c0d75d5fb/src/cargo/core/manifest.rs#L231

The `kind` field tells us about the command-line flag used to refer to this target. The `crate_type` field tells us if this is actually a binary. I think for example `kind`, we also need to check that `crate_type` is `bin`.

matklad:
Here's the doc with an example which specifies non-binary crate type for an example: https://doc.rust-lang.org/cargo/reference/manifest.html#examples

Undin:
Oh, it's very useful info
Thanks!

Undin:
@matklad Done

CONTRIBUTING.md
```diff
@@ -148,6 +148,7 @@ change. Common tags are:
   * PERF for performance optimizations
   * PRJ for project creation changes 
   * ACT for actions
+  * DBG for debugger
```
matklad: Btw, this section now uses **a lot** of space, so maybe hind it behind `\<details\>\</details\>`?

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
```diff
@@ -146,6 +146,8 @@ private fun buildProjectAndGetBinaryArtifactPath(project: Project, command: Carg
                             .mapNotNull { CargoMetadata.Artifact.fromJson(it) }
                             .filter { (target, profile) ->
                                 target.cleanKind == CargoWorkspace.TargetKind.BIN
+                                    // TODO: support cases when `crate_types` contains not only binary
+                                    || target.cleanKind == CargoWorkspace.TargetKind.EXAMPLE && target.crate_types.singleOrNull() == "bin"
```
matklad: `target.crate_types.singleOrNull() == "bin"` does not look too great :) 

Let's add an enum for this, sort-of like `cleanKind` works, and also add a comment with a link to the relevant comment in Cargo, which explains what's going on. 

Other than this stylistic drawback, r+!

Undin:
bors r=matklad

Issue #2205, opened 2018-01-17
Intention/inspection to sort method in impl acroding to the order in the trait


matklad:
In Rust, traits are usually small, so it makes sense to sort methods in the impl in the same oreder as in the trait. Would love to have this! :-)

t-kameyama:
I will work on this.

Undin:
Implemented in #2208

PR #2208, opened 2018-01-18
INT/INSP: sort method in impl according to the order in the trait

t-kameyama:
Fix for #2205


src/main/kotlin/org/rust/ide/intentions/SortImplTraitMethodsIntention.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.resolveToTrait
+
+class SortImplTraitMethodsIntention : RsElementBaseIntentionAction<RsImplItem>() {
+
+    override fun getText() = "Sort methods in the same order as in the trait"
```
mkaput: A bit too long for me, maybe `Apply same method order` ?

src/main/kotlin/org/rust/ide/intentions/SortImplTraitMethodsIntention.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    companion object {
+        fun isApplicableTo(implItem: RsImplItem): Boolean {
+            val implMethods = implItem.members?.functionList ?: return false
+            val traitMethods = implItem.traitRef?.resolveToTrait?.members?.functionList ?: return false
+
+            if (implMethods.size != traitMethods.size) return false
+            if (traitMethods.zip(implMethods).all { it.first.name == it.second.name }) return false
+
+            return true
```
mkaput: Microoptimisation

```kotlin
return traitMethods.zip(implMethods).any { it.first.name != it.second.name }
```

t-kameyama:
Fixed.

src/main/kotlin/org/rust/ide/intentions/SortImplTraitMethodsIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+        val implItem = element.ancestorStrict<RsImplItem>() ?: return null
+        if (!isApplicableTo(implItem)) return null
+        return implItem
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: RsImplItem) {
+        val implMethods = ctx.members?.functionList ?: return
+        val traitMethods = ctx.traitRef?.resolveToTrait?.members?.functionList ?: return
+
+        val implMethodMap = implMethods.map { it.name to it }.toMap()
```
Undin: Can be simplified to `implMethods.associate { it.name to it }`

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMethodsInspection.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.IntentionWrapper
+import com.intellij.codeInspection.ProblemsHolder
+import org.rust.ide.intentions.SortImplTraitMethodsIntention
+import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.RsVisitor
+
+class RsSortImplTraitMethodsInspection : RsLocalInspectionTool() {
```
Undin: I think different methods order is not potential error so I prefer to drop inspection and leave the only intention
@matklad what do you think?

src/main/kotlin/org/rust/ide/intentions/SortImplTraitMethodsIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+    override fun getFamilyName(): String = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsImplItem? {
+        val implItem = element.ancestorStrict<RsImplItem>() ?: return null
+        if (!isApplicableTo(implItem)) return null
+        return implItem
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: RsImplItem) {
+        val implMethods = ctx.members?.functionList ?: return
```
Undin: Maybe sort not only functions but all members: functions, associated types and associated consts?

src/main/kotlin/org/rust/ide/intentions/SortImplTraitMethodsIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+        val implMethodMap = implMethods.map { it.name to it }.toMap()
+        val sortedImplMethods = traitMethods.mapNotNull { implMethodMap[it.name]?.copy() }
+
+        traitMethods.zip(implMethods).forEachIndexed { index, (traitMethod, implMethod) ->
+            if (traitMethod.name != implMethod.name) implMethod.replace(sortedImplMethods[index])
+        }
+    }
+
+    companion object {
+        fun isApplicableTo(implItem: RsImplItem): Boolean {
```
Undin: If we drop inspection, let's return something like 
```kotlin
data class Context(
    val traitItems: List<RsItemElement>,
    val implItems: List<RsItemElement>
)
```
here and use it as applicable context to avoid items search twice: in `isApplicableTo` and in `invoke`

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMethodsInspection.kt
matklad: I suggest that we make it a *weak warning* at least at first. If the large number of people complain, we can turn this into inspection, **but** by this time we'll educate people that this stuff is available. 

However, there's a high chance that most of the folks would be okay with the weak warning: you can always disable it in the settings. 

As a comparison, Kotlin's "replace a nice for-loop with an unreadable sidefectful sequence of high-order ops which allocate on every intermediate step" definitely has false positives, but it's a weak warning non the less because it advertises the feature.

Undin: > we can turn this into inspection

You mean intention, don't you?

Hm, then let's drop intention and make this inspection enabled by default
I really think that we need only one of them because user always can suppress highlighting of inspection and use only quick fix and if we have both intention and inspection we get following situation
screen shot 2018-01-25 at 02 25 13

matklad: Yep, I meant "let's have a thing with a squiggly line and, if user complain, downgrade it to lightbulb only thing".

t-kameyama:
Fixed

src/main/kotlin/org/rust/ide/intentions/SortImplTraitMembersIntention.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    companion object {
+        fun isApplicableTo(implItem: RsImplItem): Context? {
+            val traitItems = targetItems(implItem.traitRef?.resolveToTrait).takeIf { it.isNotEmpty() } ?: return null
+            val implItems =  targetItems(implItem).takeIf { it.isNotEmpty() } ?: return null
+
+            if (traitItems.size != implItems.size) return null
+            if (traitItems.zip(implItems).all { it.first.name == it.second.name }) return null
```
Undin: Hm, it seems it's not enough to check only items name because items with different types of can have same names.
Pay attention that you use name as unique identifier of item in `invoke` method too

t-kameyama:
Fixed

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
```diff
@@ -0,0 +1,58 @@
...
+                SortImplTraitMembersFix(traitItems, implItems)
+            )
+        }
+    }
+
+    private fun targetItems(item: RsTraitOrImpl?): List<RsItemElement> {
+        val children = item?.members?.children
+        return children?.mapNotNull {
+            if (it is RsFunction || it is RsConstant || it is RsTypeAlias) it as? RsItemElement else null
+        }?.takeIf { it.size == children.size } ?: emptyList()
```
Undin: Why we need `takeIf { it.size == children.size }` here?
We return empty list if trait or impl has children that are not `RsFunction`, `RsConstant` or `RsTypeAlias` - RsInnerAttr for example
Looks like that everything will be ok if we just drop `takeIf`

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
```diff
@@ -0,0 +1,58 @@
...
+    private class SortImplTraitMembersFix(
+        private val traitItems: List<RsItemElement>,
+        private val implItems: List<RsItemElement>
+    ) : LocalQuickFix {
+        override fun getName() = "Apply same member order"
+
+        override fun getFamilyName() = name
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            val implItemMap = implItems.associate { it.name + it.elementType to it }
```
Undin: Maybe use pair of name and type element as map key instead of concatenation string representations?

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
```diff
@@ -0,0 +1,58 @@
...
+    private fun targetItems(item: RsTraitOrImpl?): List<RsItemElement> {
+        val children = item?.members?.children
+        return children?.mapNotNull {
+            if (it is RsFunction || it is RsConstant || it is RsTypeAlias) it as? RsItemElement else null
+        }?.takeIf { it.size == children.size } ?: emptyList()
+    }
+
+    private class SortImplTraitMembersFix(
+        private val traitItems: List<RsItemElement>,
+        private val implItems: List<RsItemElement>
```
Undin: Let's avoid holding psi elements in quick fix object - it can lead to memory leak.
Usually, if you want to use some psi element in quick fix, you have several options:
* use `com.intellij.psi.SmartPsiElementPointer`
* get psi element from `ProblemDescriptor` in `applyFix` method. You get same psi element that you use in `registerProblem`
* use `com.intellij.codeInspection.LocalQuickFixOnPsiElement` as parent class of your quick fix

2nd and 3rd options use `com.intellij.psi.SmartPsiElementPointer` under the hood so basically there is only one way to do right.

I think here we can use `LocalQuickFixOnPsiElement` to pass impl and trait objects, create separate method to get items from them and check all conditions (like `findApplicableContext` in `RsElementBaseIntentionAction`) and use this method in `visitImplItem` and in `applyFix`.

t-kameyama:
Fixed

Undin:
Looks great!
Thanks!
bors r+

kumbayo:
I have seen some exceptions which seem to be related to this change.
```
java.lang.Throwable: Assertion failed: Both elements must be from the same file
	at com.intellij.openapi.diagnostic.Logger.assertTrue(Logger.java:172)
	at com.intellij.codeInspection.LocalQuickFixOnPsiElement.<init>(LocalQuickFixOnPsiElement.java:45)
	at org.rust.ide.inspections.RsSortImplTraitMembersInspection$SortImplTraitMembersFix.<init>(RsSortImplTraitMembersInspection.kt:41)
	at org.rust.ide.inspections.RsSortImplTraitMembersInspection$buildVisitor$1.visitImplItem(RsSortImplTraitMembersInspection.kt:27)
	at org.rust.lang.core.psi.impl.RsImplItemImpl.accept(RsImplItemImpl.java:27)
	at org.rust.lang.core.psi.impl.RsImplItemImpl.accept(RsImplItemImpl.java:31)
	at com.intellij.codeInspection.InspectionEngine.acceptElements(InspectionEngine.java:82)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.lambda$visitRestElementsAndCleanup$4(LocalInspectionsPass.java:309)
	at com.intellij.concurrency.ApplierCompleter.execAndForkSubTasks(ApplierCompleter.java:133)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:1127)
	at com.intellij.concurrency.ApplierCompleter.lambda$wrapInReadActionAndIndicator$1(ApplierCompleter.java:105)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:543)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:488)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:94)
	at com.intellij.concurrency.ApplierCompleter.wrapInReadActionAndIndicator(ApplierCompleter.java:116)
	at com.intellij.concurrency.ApplierCompleter.lambda$compute$0(ApplierCompleter.java:96)
	at com.intellij.openapi.application.impl.ReadMostlyRWLock.executeByImpatientReader(ReadMostlyRWLock.java:143)
	at com.intellij.openapi.application.impl.ApplicationImpl.executeByImpatientReader(ApplicationImpl.java:229)
	at com.intellij.concurrency.ApplierCompleter.compute(ApplierCompleter.java:96)
	at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

The code currently uses `startElement` and `endElement` to hold a `RsImplItem` and a `RsTraitItem`.
But those can be in 2 different files which the code does not expect.

Undin:
Fixed in #2252

PR #2209, opened 2018-01-18
Fixed indexing of `impl<T> for &T`

vlad20012:


src/main/kotlin/org/rust/lang/core/types/TyFingerprint.kt
```diff
@@ -25,20 +25,27 @@ data class TyFingerprint constructor(
...
+                            TyFingerprint(name)
+                        }
+                    }
                 }
                 is RsRefLikeType -> {
                     if (type.isPointer)
                         TyFingerprint("*T")
                     else
-                        create(type.typeReference)
+                        create(type.typeReference, typeParameters)
```
Undin: Hm, it's interesting that we use different logic for references and pointers

PR #2163, opened 2017-12-28
WIP: Improve navigation related to trait implementations

kumbayo:
This should not be merged yet as it is still work in progress.
Fixes #2122

PR #2217, opened 2018-01-21
Improve navigation around const and type in traits

kumbayo:
Allow to navigate from "const" and "type" in trait impl blocks to the according place in the trait the same way we already do for "fn".

Also slightly improve documentation and the breadcrumb bar entries related to "const" and "type".

This is a preparation for #2163

### DIFF
--- a/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
@@ -73,30 +73,13 @@ private fun RsDocAndAttributeOwner.header(usePreTag: Boolean): String {
     val rawLines = when (this) {
         is RsFieldDecl -> listOfNotNull((parent?.parent as? RsDocAndAttributeOwner)?.presentableQualifiedName)
         is RsStructOrEnumItemElement, is RsTraitItem -> listOfNotNull(presentableQualifiedModName)
-        is RsFunction -> {
+        is RsAbstractable -> {
             val owner = owner
             when (owner) {
-                is RsFunctionOwner.Foreign,
-                is RsFunctionOwner.Free -> listOfNotNull(presentableQualifiedModName)
-                is RsFunctionOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
-                is RsFunctionOwner.Trait -> owner.trait.declarationText
-            }
-        }
-        is RsConstant -> {
-            val owner = owner
-            when (owner) {
-                is RsConstantOwner.Foreign,
-                is RsConstantOwner.Free -> listOfNotNull(presentableQualifiedModName)
-                is RsConstantOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
-                is RsConstantOwner.Trait -> owner.trait.declarationText
-            }
-        }
-        is RsTypeAlias -> {
-            val owner = owner
-            when (owner) {
-                is RsTypeAliasOwner.Free -> listOfNotNull(presentableQualifiedModName)
-                is RsTypeAliasOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
-                is RsTypeAliasOwner.Trait -> owner.trait.declarationText
+                is RsAbstractableOwner.Foreign,
+                is RsAbstractableOwner.Free -> listOfNotNull(presentableQualifiedModName)
+                is RsAbstractableOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
+                is RsAbstractableOwner.Trait -> owner.trait.declarationText
             }
         }
         else -> listOfNotNull(presentableQualifiedName)
--- a/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
@@ -72,22 +72,22 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
 private fun RsDocAndAttributeOwner.header(usePreTag: Boolean): String {
     val rawLines = when (this) {
         is RsFieldDecl -> listOfNotNull((parent?.parent as? RsDocAndAttributeOwner)?.presentableQualifiedName)
+        is RsStructOrEnumItemElement, is RsTraitItem -> listOfNotNull(presentableQualifiedModName)
         is RsFunction -> {
             val owner = owner
             when (owner) {
-                is RsFunctionOwner.Foreign, is RsFunctionOwner.Free -> listOfNotNull(presentableQualifiedModName)
-                is RsFunctionOwner.Impl ->
-                    listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
+                is RsFunctionOwner.Foreign,
+                is RsFunctionOwner.Free -> listOfNotNull(presentableQualifiedModName)
+                is RsFunctionOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
                 is RsFunctionOwner.Trait -> owner.trait.declarationText
             }
         }
-        is RsStructOrEnumItemElement, is RsTraitItem -> listOfNotNull(presentableQualifiedModName)
         is RsTypeAlias -> {
             val owner = owner
             when (owner) {
+                is RsTypeAliasOwner.Free -> listOfNotNull(presentableQualifiedModName)
                 is RsTypeAliasOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
                 is RsTypeAliasOwner.Trait -> owner.trait.declarationText
-                is RsTypeAliasOwner.Free -> listOfNotNull(presentableQualifiedModName)
             }
         }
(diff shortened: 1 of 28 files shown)


src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -118,8 +127,15 @@ private fun RsDocAndAttributeOwner.signature(usePreTag: Boolean): String {
             retType?.generateDocumentation(buffer)
             listOf(buffer.toString()) + whereClause?.documentationText.orEmpty()
         }
-        // All these types extend RsTypeBearingItemElement and RsGenericDeclaration interfaces
```
Undin: Let's replace mention about `RsTypeBearingItemElement` with `RsItemElement` - currently we use cast to `RsItemElement`.

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -72,22 +72,31 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
...
+                is RsFunctionOwner.Free -> listOfNotNull(presentableQualifiedModName)
+                is RsFunctionOwner.Impl -> listOfNotNull(presentableQualifiedModName) + owner.impl.declarationText
                 is RsFunctionOwner.Trait -> owner.trait.declarationText
             }
         }
-        is RsStructOrEnumItemElement, is RsTraitItem -> listOfNotNull(presentableQualifiedModName)
+        is RsConstant -> {
+            val owner = owner
+            when (owner) {
+                is RsConstantOwner.Foreign,
```
Undin: Hm, maybe we want to unity `Rs*Owner` classes to unify work with them? I'm not sure that it's very good decision because type alias owner can't be foreign

src/test/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandlerTest.kt
```diff
@@ -26,22 +26,54 @@ class RsGotoSuperHandlerTest : RsTestBase() {
         }
     """)
 
-    fun `test method declataion from impl`() = checkNavigation("""
+    fun `test method declaration from impl`() = checkNavigation("""
```
Undin: 👍

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
kumbayo: I initially tried to do that while creating this series but abandoned the idea because i hit a lot of problems and code churn at that time. This time i tried again and it went much more smoothly.
And apart of the size of the code change i think its a good idea.

src/main/kotlin/org/rust/lang/core/psi/ext/RsAbstractable.kt
```diff
@@ -6,7 +6,24 @@
...
+
+// Resolve a const, fn or type in a impl block to the corresponding item in the trait block
+val RsAbstractable.superItem: RsAbstractable?
+    get() {
+        val rustImplItem = ancestorStrict<RsImplItem>() ?: return null
+        val superTrait = rustImplItem.traitRef?.resolveToTrait ?: return null
+        return when (this) {
+            is RsConstant -> superTrait.members?.constantList.orEmpty().find { it.name == this.name }
+            is RsFunction -> superTrait.members?.functionList.orEmpty().find { it.name == this.name }
+            is RsTypeAlias -> superTrait.members?.typeAliasList.orEmpty().find { it.name == this.name }
```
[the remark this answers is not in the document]
kumbayo: Done

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
kumbayo: Done

src/main/kotlin/org/rust/lang/core/psi/ext/RsAbstractable.kt
```diff
@@ -6,7 +6,54 @@
...
+    class Impl(val impl: RsImplItem, val isInherent: Boolean) : RsAbstractableOwner()
+
+    val isInherentImpl: Boolean get() = this is Impl && isInherent
+    val isTraitImpl: Boolean get() = this is Impl && !isInherent
+    val isImplOrTrait: Boolean get() = this is Impl || this is Trait
+}
+
+val RsAbstractable.owner: RsAbstractableOwner
+    get() {
+        val stubOnlyParent = when (this) {
```
kumbayo: I could not find a nice way to do this for all 3 different types. Hope this is OK.

src/main/kotlin/org/rust/lang/core/psi/ext/RsAbstractable.kt
```diff
@@ -6,7 +6,54 @@
...
+            is RsForeignModItem -> RsAbstractableOwner.Foreign
+            is RsMembers -> {
+                val traitOrImpl = parent.parent
+                when (traitOrImpl) {
+                    is RsImplItem -> RsAbstractableOwner.Impl(traitOrImpl, isInherent = traitOrImpl.traitRef == null)
+                    is RsTraitItem -> RsAbstractableOwner.Trait(traitOrImpl)
+                    else -> error("unreachable")
+                }
+            }
+            else -> RsAbstractableOwner.Free
```
kumbayo: `is RsItemsOwner -> RsConstantOwner.Free` from `RsConstant` is now handled by the `else` just like already done in `RsFunction`

src/main/kotlin/org/rust/lang/core/types/RsPsiTypeImplUtil.kt
```diff
@@ -20,9 +20,10 @@ object RsPsiTypeImplUtil {
         val typeReference = psi.typeReference
         if (typeReference != null) return typeReference.type
         return when (psi.owner) {
-            is RsTypeAliasOwner.Free -> TyUnknown
-            is RsTypeAliasOwner.Trait -> TyTypeParameter.associated(psi)
-            is RsTypeAliasOwner.Impl -> TyUnknown
+            is RsAbstractableOwner.Free -> TyUnknown
+            is RsAbstractableOwner.Trait -> TyTypeParameter.associated(psi)
+            is RsAbstractableOwner.Impl -> TyUnknown
+            is RsAbstractableOwner.Foreign -> TyUnknown
```
kumbayo: This `is RsAbstractableOwner.Foreign` is new because of the unification.

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
```diff
@@ -32,21 +31,22 @@ class RsGotoSuperHandler : LanguageCodeInsightActionHandler {
...
+        RsAbstractable::class.java,
         RsMod::class.java
     ) ?: return null
 
-    if (modOrMethod is RsFunction) {
-        return if (modOrMethod.owner.isTraitImpl) {
-            modOrMethod.superMethod
+    if (modOrAbstractable is RsAbstractable) {
+        val superItem = modOrAbstractable.superItem
+        return if (superItem != null) {
```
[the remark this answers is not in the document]
kumbayo: Please let me think about this part a bit more before merging it.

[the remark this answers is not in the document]
kumbayo: Not sure if i understood your request correctly. The modified code should now behave just as before the patch but also support const and type alias.
This was made possible by the `RsAbstractableOwner` unification :-)

kumbayo:
@Undin Your comments should be addressed now.

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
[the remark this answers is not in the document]
Undin: Yeah, now behavior is the same as before. But initial implementation could return super mod for any abstractable item from trait impl if `superItem` returned null. But `superItem` can return null for item even if item from trait impl, for example, if we can't resolve trait ref. And prev implementation tried to find super mod in this case that is not correct behavior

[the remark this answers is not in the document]
Undin: > This was made possible by the RsAbstractableOwner unification :-)

👍 
I thought about it too :)

src/main/kotlin/org/rust/lang/core/types/RsPsiTypeImplUtil.kt
Undin: I think it's acceptable price for this unification

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -181,6 +180,7 @@ private val RsItemElement.declarationModifiers: List<String> get() {
         }
         is RsStructItem -> modifiers += "struct"
         is RsEnumItem -> modifiers += "enum"
+        is RsConstant -> modifiers += "const"
```
Undin: Oops, we forgot about static here)
I'll fix it

kumbayo: Thanks for the fix!

PR #2227, opened 2018-01-23
TY: attach type inference context to constants and array size exprs

vlad20012:
Provides correct type inference in constants


src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeInferenceRoot.kt
```diff
@@ -0,0 +1,8 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+interface RsTypeInferenceRoot : RsElement
```
vlad20012: maybe someone can suggest a better name for it?

src/main/grammars/RustParser.bnf
```diff
@@ -707,7 +712,7 @@ TypeReference ::= TypeReferenceInner {
 ScalarTypeReference ::= ScalarTypeReferenceInner { elementType = TypeReference }
 
 ArrayType ::= '[' TypeReference [';' AnyExpr] ']' {
-  implements = "org.rust.lang.core.psi.ext.RsTypeElement"
+  implements = ["org.rust.lang.core.psi.ext.RsTypeElement" "org.rust.lang.core.psi.ext.RsTypeInferenceRoot"]
```
matklad: I think other cases in `.bnf` place a space between `[` and `"`?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -126,20 +129,35 @@ class RsInferenceContext(
         }
     }
 
-    fun infer(fn: RsFunction): RsInferenceResult {
-        extractParameterBindings(fn)
+    fun infer(fn: RsTypeInferenceRoot): RsInferenceResult {
```
matklad: `fn` does not look like the best name anymore :)

src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeInferenceRoot.kt
matklad: Regardless of the name, it probably needs a doc string and a mention in the Architecture.md.

However, I find `TypeInferenceRoot` name to be pretty great actually!

src/main/kotlin/org/rust/lang/core/types/Extensions.kt
```diff
@@ -32,19 +31,19 @@ val RsTypeElement.lifetimeElidable: Boolean get() {
     return typeOwner !is RsFieldDecl && typeOwner !is RsTupleFieldDecl && typeOwner !is RsTypeAlias
 }
 
-val RsFunction.inference: RsInferenceResult
+val RsTypeInferenceRoot.inference: RsInferenceResult
     get() = CachedValuesManager.getCachedValue(this, {
         CachedValueProvider.Result.create(inferTypesIn(this), PsiModificationTracker.MODIFICATION_COUNT)
     })
 
 val PsiElement.inference: RsInferenceResult?
```
matklad: Could `PsiElement.inference` and `RsTypeInferenceRoot.inference` have different names? They do slightly different things, and the case where `x`'s static type is `RsCompositteElement`, but its runtime type is `RsFunction` is murky.

matklad: And can we make `PsiElement.inference` private? It looks overly general I think :)

matklad:
>maybe someone can suggest a better name for it?

What about `InferenceContext` / `InferenceContextOwner`?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
Undin: Same issue about name of `RsFnInferenceContext`

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -126,20 +129,35 @@ class RsInferenceContext(
...
+            }
+            is RsArrayType -> {
+                val fctx = RsFnInferenceContext(this, TyInteger(TyInteger.Kind.usize))
+                fn.expr?.let { fctx.inferLambdaBody(it) }
+            }
+            is RsVariantDiscriminant -> {
+                val fctx = RsFnInferenceContext(this, TyInteger(TyInteger.Kind.isize))
+                fn.expr?.let { fctx.inferLambdaBody(it) }
+            }
+            else -> error("Type inference is not implemented for PSI element of type `${fn.javaClass}` that " +
```
Undin: It seems that `ArrayExpr` should be `RsTypeInferenceRoot` as well

src/main/kotlin/org/rust/lang/core/types/Extensions.kt
vlad20012: :thinking: Hmm, I think we can just unify behavior of these props. Just use `ancestorOrSelf` here. 
Anyway, the name `inference` looks really ugly.
I can't make it private because it is used in method/field/path reference impls

vlad20012:
I've used `RsInferenceContextOwner`. We already have a lot of `*Owner` interfaces, so it seems uniform.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -126,20 +129,30 @@ class RsInferenceContext(
...
+        } else {
+            val (retTy, expr) = when (element) {
+                is RsConstant -> element.typeReference?.type to element.expr
+                is RsArrayType -> TyInteger(TyInteger.Kind.usize) to element.expr
+                is RsVariantDiscriminant -> TyInteger(TyInteger.Kind.isize) to element.expr
+                else -> error("Type inference is not implemented for PSI element of type " +
+                        "`${element.javaClass}` that implement `RsInferenceContextOwner`")
+            }
+            if (expr != null) {
+                RsFnInferenceContext(this, retTy ?: TyUnknown).inferLambdaBody(expr)
```
matklad: Should we rename `RsFnInferenceContex` to ` RsInferenceContex`?

And we probably can `drop` Rs prefix here, because this is IJ-Rust specific thing and unlikely to clash with something else.

vlad20012: The name inspired by rustc (`FnCtxt`)

vlad20012: The name inspired by rustc (FnCtxt). We have 2 types: `RsInferenceContext` and `RsFnInferenceContext`, where `RsFnInferenceContext` creates for something that have a return type. For example, for each lambda expr. `RsInferenceContext` is single for each function/constant. So I can't think of a better name for it. The prefix `Rs` exists because there is `InferenceContext` class in JDT and javac.

PR #2230, opened 2018-01-27
TY: unify match pattern with matching variable

vlad20012:
```rust
fn main() {
    match 0u8 {
        0 => {},
      //^ u8
        _ => {},
    };
}
```
Previously we ignored patterns in type inference at all, i.e. that `0` in the pattern was `<unknown>`.
It also simplifies `RsPathReferenceImpl` because path resolution in patterns is now provided by type inference engine.


vlad20012:
bors try

Undin:
bors r=vlad20012

PR #2235, opened 2018-01-28
RES: correctly resolve modules with 'path' attribute

### LINKED ISSUE #2145: Improve support of path attribute in module system
Rust module system allows mapping module system to a filesystem.
Without `path` attribute it is quite straightforward - module name directly matches on directory or file with the same name.
But `path` attribute allows mapping module to directory or file with different name.

We already support simple cases like this
```rust
// main.rs
#[path="bar/baz.rs"]
mod foo;
```
But we don't support more complex cases when path attribute is added to root module of some hierarchy
```rust
//main.rs
#[path="bar"]
mod foo {
    mod baz;
}
```
In this case, code of `foo::baz` module should be in `bar/baz.rs` file.
Or even more complex case
```rust
// main.rs
#[path="bar"]
mod foo {
    #[path="qwe.rs"]
    mod baz;
}
```
where code of `foo::baz` module should be in `bar/qwe.rs` file.
It would be great to support such cases too.

Undin:
Closes #2145

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsModItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsModItem.kt
@@ -6,13 +6,14 @@
 package org.rust.lang.core.psi.ext
 
 import com.intellij.lang.ASTNode
+import com.intellij.openapi.util.io.FileUtil
 import com.intellij.psi.PsiDirectory
 import com.intellij.psi.stubs.IStubElementType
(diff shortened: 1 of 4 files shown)


Undin:
@matklad or @vlad20012 r?

Undin:
bors r=matklad

Undin:
bors r=matklad

Undin:
Okay(

Undin:
bors r=matklad

PR #2239, opened 2018-01-29
HINT: Remove unnecessary inlay hint in lambda expr

### LINKED ISSUE #2236: In certain closures, parameter type is displayed as if it's inferred
image

Notice on the third line `|e: CreateEmbed: CreateEmbed|`. It's clear that `e` is `CreateEmbed`.

farodin91:
Fixes  #2236!

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
@@ -98,9 +98,9 @@ enum class HintType(desc: String, enabled: Boolean) {
             val element = elem as? RsLambdaExpr ?: return emptyList()
             val type = element.type as? TyFunction ?: return emptyList()
             return element.valueParameterList.valueParameterList
-                .mapIndexed { index, parameter -> type.paramTypes.getOrNull(index) to parameter.textRange.endOffset }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -98,6 +98,7 @@ enum class HintType(desc: String, enabled: Boolean) {
             val element = elem as? RsLambdaExpr ?: return emptyList()
             val type = element.type as? TyFunction ?: return emptyList()
             return element.valueParameterList.valueParameterList
+                .filter { it.typeReference == null }
```
Undin: These changes don't take into account cases when lambda has several arguments and only some of them have explicit type parameters
For example:
```rust
struct S;
fn foo<T: Fn(S, S, (S, S)) -> ()>(action: T) {}
fn main() {
    foo(|x, y: S, z| {});
}
```
Here type hint for `z` is wrong
screen shot 2018-01-30 at 11 32 21

Undin:
Looks like you missed `//^` in `test lambda type should show after an defined type correct`

farodin91:
I implemented this with an working environment.

farodin91:
@Undin Do you like to review?

PR #2241, opened 2018-02-01
Constexpr eval fixes

vlad20012:


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1104,7 +1126,7 @@ private fun RsStubLiteralType?.serialize(dataStream: StubOutputStream) {
         dataStream.writeInt(-1)
```
Undin: `writeInt` -> `writeByte`

vlad20012: Hmm =)

vlad20012: Fixed

vlad20012:
bors try

Issue #2258, opened 2018-02-04
'Not all trait items implemented' checks items only by their names


Undin:
```rust
trait A {
    type C;
    const C: i32;
}

struct S;

impl A for S {
    type C = ();
}
```

**Expected**: there is error highlighting with `Not all trait items implemented` error
**Actual**: there isn't any highlighting

Related code can be found [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt#L266-L272)

Undin:
Fixed in #2262

PR #2262, opened 2018-02-05
PSI: 'Not all trait items implemented' checks items only by their names

t-kameyama:
Fix for #2258


Undin:
Let's add test that corresponding annotation really works in these cases. Related tests can be found [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt#L59-L90)

t-kameyama:
Added test

Undin:
bors r+
Thanks!

PR #2266, opened 2018-02-05
TY: Introduced TyProjection (replacement for associated type)

vlad20012:


src/main/kotlin/org/rust/lang/core/types/ty/TyProjection.kt
```diff
@@ -0,0 +1,55 @@
...
+ * ```
+ * trait Trait { type Item; }
+ * fn foo<T: Trait>() -> T::Item { unimplemented!() }
+ * fn bar<T: Trait>() -> <T as Trait>::Item { unimplemented!() }
+ * ```
+ *
+ * Fields meaning:
+ * ```
+ * <T as Trait>::Item
+ * |     |       |
```
Undin: I think we need extra space before first `|`

src/main/kotlin/org/rust/lang/core/types/BoundElement.kt
```diff
@@ -7,36 +7,38 @@ package org.rust.lang.core.types
...
  * `fn make_t<T>() -> T { ... }`, together with actual
  * type arguments, like `T := i32`.
  */
 data class BoundElement<out E : RsElement>(
     val element: E,
-    val subst: Substitution = emptySubstitution
+    /** Generic type bindings, e.g. `type T = Box<u8>;` */
+    val subst: Substitution = emptySubstitution,
+    /** Associated type bindings, e.g. `type T = Iterator<Item=u8>;` */
+    val assoc: Map<RsTypeAlias, Ty> = emptyMap()
```
Undin: Maybe slightly change docs for these fields? It's a little bit confusing and I spent some time to understand it.
Maybe something like this:
"Generic type bindings, e.g. if we have concrete type `Box<u8>` with a generic declaration as `Box<T>`, `subst` maps `T` to `u8`"? 
And something similar for `assoc` map as well

src/main/kotlin/org/rust/lang/core/types/BoundElement.kt
```diff
@@ -7,36 +7,38 @@ package org.rust.lang.core.types
...
 import org.rust.lang.core.types.ty.emptySubstitution
 import org.rust.lang.core.types.ty.foldValues
-import org.rust.lang.core.types.ty.substituteInValues
 
-/* Represents a potentially generic Psi Element, like
+/**
+ * Represents a potentially generic Psi Element, like
  * `fn make_t<T>() -> T { ... }`, together with actual
  * type arguments, like `T := i32`.
  */
```
Undin: Let's add some info about associated types too

vlad20012:
Done :)

PR #2268, opened 2018-02-06
INSP: highlight only part of impl 

Undin:
Improve `Different impl member order from the trait` inspection highlighting to underline only `impl Trait for Type` part instead of the whole impl element


Undin:
@matklad r?

PR #2270, opened 2018-02-06
 ANN: Move mutable annotation into a own annotator 

### LINKED ISSUE #1950: Add underline to &mut self?
Add underline under &mut self to be consistent with display of other mut variables?

farodin91:
Fixes #1950

@Undin r? Any idea for tests?

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
@@ -1,0 +1,82 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator
+
+import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.lang.annotation.Annotator
+import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.psi.PsiElement
+import org.rust.ide.colors.RsColor
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.mutability
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.type
+import org.rust.openapiext.isUnitTestMode
+
+class RsHighlightingMutableAnnotator : Annotator {
+    companion object {
+        private val MUTABLE_HIGHTLIGHTING = HighlightSeverity(
+            "MUTABLE_HIGHTLIGHTING",
+            HighlightSeverity.INFORMATION.myVal + 3
+        )
+    }
+
+
+    override fun annotate(element: PsiElement, holder: AnnotationHolder) {
+        val ref = when (element) {
+            is RsPath -> element.reference.resolve() ?: return
+            is RsSelfParameter -> element
+            is RsPatBinding -> element
+            else -> return
+        }
+        distinctAnnotation(element, ref, holder)
+    }
+
+    private fun annotationFor(ref: RsElement): RsColor? = when (ref) {
+        is RsSelfParameter -> RsColor.MUT_PARAMETER
+        is RsPatBinding -> if (ref.ancestorStrict<RsValueParameter>() != null) {
+            RsColor.MUT_PARAMETER
+        } else {
+            RsColor.MUT_BINDING
+        }
+        else -> null
+    }
+
+    private fun distinctAnnotation(element: PsiElement, ref: RsElement, holder: AnnotationHolder) {
+        val color = annotationFor(ref) ?: return
+        if (ref.isMut) {
+            val element = partToHighlight(element)
+            addHighlightingAnnotation(holder, element, color)
+        }
+    }
+
+    private fun partToHighlight(element: PsiElement): PsiElement = when (element) {
+        is RsSelfParameter -> element.self
+        is RsPatBinding -> element.identifier
+        else -> element
+    }
+
+    private fun addHighlightingAnnotation(holder: AnnotationHolder, target: PsiElement, key: RsColor) {
+        // These following lines allow to test the mutable highlight for it own using INFORMATION and the description.
+        val annotationSeverity = if (isUnitTestMode) HighlightSeverity.INFORMATION else MUTABLE_HIGHTLIGHTING
+        val annotationText = if (isUnitTestMode) key.attributesDescriptor.displayName else null
+        holder.createAnnotation(
+            annotationSeverity,
+            target.textRange,
+            annotationText
+        ).textAttributes = key.textAttributesKey
+    }
+
+}
+
+val RsElement.isMut: Boolean
+    get() = when (this) {
+        is RsPatBinding -> mutability.isMut || type.let { it is TyReference && it.mutability.isMut }
+        is RsSelfParameter -> mutability.isMut
+        else -> false
+    }
--- a/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
@@ -99,14 +99,7 @@ private fun colorFor(element: RsElement): RsColor? = when (element) {
     is RsModDeclItem -> RsColor.MODULE
     is RsModItem -> RsColor.MODULE
     is RsPatBinding -> {
-        val isParameter = element.ancestorStrict<RsValueParameter>() != null
-        val isMut = element.mutability.isMut || element.type.let { it is TyReference && it.mutability.isMut }
-        when {
-            isMut && isParameter -> RsColor.MUT_PARAMETER
-            isMut -> RsColor.MUT_BINDING
-            isParameter -> RsColor.PARAMETER
-            else -> null
-        }
+        if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else null
     }
     is RsStructItem -> RsColor.STRUCT
     is RsTraitItem -> RsColor.TRAIT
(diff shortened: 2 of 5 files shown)


mfarrugi:
The self parameter can already be configured to be underlined. 
This PR highlights `&mut self ` as a mutable parameter, if that's the intent the title should be changed.

farodin91:
I have no idea why this currently is working correctly with my changes. Magic detects my name or description properly.

Undin:
@farodin91 It's not really true. It works correctly only with default color settings.

Final text attributes of element are composition of text attributes from `RsHighlighter` and from different annotators (in our case, it is `RsHighlightingAnnotator` generally) and text attributes from annotation with higher priority replace text attributes from annotations with lower priority.

So, currently, color of `self` is `RsColor.KEYWORD` (provided by `RsHighlighter`) + `RsColor.SELF_PARAMETER` (provided by `RsHighlightingAnnotator`) and color of `mut self` is `RsColor.KEYWORD` + `RsColor.MUT_PARAMETER`. By default, `RsColor.SELF_PARAMETER` has same text attributes as `RsColor.KEYWORD`, what is why your implementation works. But if you change some attributes for `RsColor.SELF_PARAMETER`, it won't apply for `mut self` element.
screen shot 2018-02-26 at 11 57 17

Our final goal is to achieve that the color for `mut self` will be `RsColor.KEYWORD` (provided by `RsHighlighter`) + `RsColor.SELF_PARAMETER` (provided by `RsHighlightingAnnotator`) + `RsColor.MUT_PARAMETER` (provided by some annotator which I suggest to implement)

farodin91:
@Undin Correct?

src/main/kotlin/org/rust/ide/annotator/RsMutableSelfParameterAnnotator.kt
```diff
@@ -0,0 +1,58 @@
...
+        val testMode = ApplicationManager.getApplication().isUnitTestMode
+        val actualSeverity = if (testMode && severity.myVal < HighlightSeverity.INFORMATION.myVal) {
+            HighlightSeverity.INFORMATION
+        } else {
+            severity
+        }
+        holder.createAnnotation(
+            actualSeverity,
+            target.textRange,
+            key.attributesDescriptor.displayName
```
Undin: We need non null message only for tests

src/main/kotlin/org/rust/ide/annotator/RsMutableSelfParameterAnnotator.kt
```diff
@@ -0,0 +1,58 @@
...
+                addHighlightingAnnotation(holder, element.self, RsColor.MUT_PARAMETER, LOW_PRIORITY_HIGHLIGHTING)
+            }
+        }
+    }
+
+    private fun addHighlightingAnnotation(holder: AnnotationHolder,
+                                          target: PsiElement,
+                                          key: RsColor,
+                                          severity: HighlightSeverity = HighlightSeverity.INFORMATION) {
+        val testMode = ApplicationManager.getApplication().isUnitTestMode
```
Undin: You can use `org.rust.openapiext.isUnitTestMode`

src/main/kotlin/org/rust/ide/annotator/RsMutableSelfParameterAnnotator.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.psi.PsiElement
+import org.rust.ide.colors.RsColor
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.RsSelfParameter
+import org.rust.lang.core.psi.ext.mutability
+import org.rust.lang.core.types.ty.Mutability
+
+class RsMutableSelfParameterAnnotator : Annotator {
+    private val LOW_PRIORITY_HIGHLIGHTING = HighlightSeverity(
```
Undin: I think in our case annotation priority of this annotator should be greater than `RsHighlightingAnnotator`, i.e. greater than `HighlightSeverity.INFORMATION` but less than `HighlightSeverity.WEAK_WARNING`

src/test/kotlin/org/rust/ide/annotator/RsMutableAnnotatorTest.kt
```diff
@@ -0,0 +1,18 @@
...
+class RsMutableAnnotatorTest : RsAnnotatorTestBase() {
+
+    fun `test mut self highlight`() = checkInfo("""
+        struct <info>Foo</info> {}
+        impl <info>Foo</info> {
+            fn <info>bar</info>(&mut <info><info>self</info></info>) {
+                <info>self</info>.<info>bar</info>();
+            }
+        }
+    """)
```
Undin: Let's check `descr` property of annotation tags too

src/main/kotlin/org/rust/ide/annotator/RsMutableSelfParameterAnnotator.kt
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+    }
+
+    private fun addHighlightingAnnotation(holder: AnnotationHolder,
+                                          target: PsiElement,
+                                          key: RsColor,
+                                          severity: HighlightSeverity = HighlightSeverity.INFORMATION) {
+        val testMode = ApplicationManager.getApplication().isUnitTestMode
+        val actualSeverity = if (testMode && severity.myVal < HighlightSeverity.INFORMATION.myVal) {
```
Undin: Let's add a comment why we change annotation priority here because it isn't obvious

farodin91:
@Undin r?

src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.rust.lang.core.psi.ext.RsNameIdentifierOwner
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.mutability
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.type
+import org.rust.openapiext.isUnitTestMode
+
+class RsHighlightingMutableAnnotator : Annotator {
+    private val HIGH_PRIORITY_HIGHLIGHTING = HighlightSeverity(
+        "HIGH_PRIORITY_HIGHLIGHTING",
```
Undin: Let's rename it to show that it's about mutable variables.
And move it into `companion object`

src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
```diff
@@ -0,0 +1,83 @@
...
+    private fun distinctAnnotation(element: PsiElement, ref: RsElement, holder: AnnotationHolder) {
+        val (ref, color) = annotationFor(ref) ?: return
+        if (ref.isMut) {
+            val element = partToHighlight(element)
+            addHighlightingAnnotation(holder, element, color)
+        }
+    }
+
+    private fun partToHighlight(element: PsiElement): PsiElement {
+        val element = when (element) {
```
Undin: Maybe return result of `when` directly and avoid warning about `Name shadowing` here?

src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
```diff
@@ -0,0 +1,83 @@
...
+        val ref = when (element) {
+            is RsPath -> element.reference.resolve() ?: return
+            is RsSelfParameter -> element
+            is RsPatBinding -> element
+            else -> return
+        }
+        distinctAnnotation(element, ref, holder)
+    }
+
+    private fun annotationFor(ref: RsElement): Pair<RsNameIdentifierOwner, RsColor>? = when (ref) {
```
Undin: Looks like that you always return `ref` in result pair. Maybe return only color?

src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
```diff
@@ -0,0 +1,83 @@
...
+            is RsSelfParameter -> element.self
+            is RsPatBinding -> element.identifier
+            else -> element
+        }
+        return element
+    }
+
+    private fun addHighlightingAnnotation(holder: AnnotationHolder, target: PsiElement, key: RsColor) {
+        // These following lines allow to test the mutable highlight for it own using WEAK_WARNING and the description.
+        val annotationSeverity = if (isUnitTestMode) HighlightSeverity.WEAK_WARNING else HIGH_PRIORITY_HIGHLIGHTING
```
Undin: Let's use `HighlightSeverity.INFORMATION` for tests because it's more suitable for this annotation and allows fixing tests in `RsHighlightingAnnotatorTest` without unnecessary changes.
Don't forget to correct comment too :wink:

src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
```diff
@@ -0,0 +1,83 @@
...
+            is RsPatBinding -> element
+            else -> return
+        }
+        distinctAnnotation(element, ref, holder)
+    }
+
+    private fun annotationFor(ref: RsElement): Pair<RsNameIdentifierOwner, RsColor>? = when (ref) {
+        is RsSelfParameter -> ref to RsColor.MUT_PARAMETER
+        is RsPatBinding -> if (ref.ancestorStrict<RsValueParameter>() != null) {
+            ref to RsColor.MUT_PARAMETER
```
Undin: Let's add test for this case too

farodin91:
@Undin all things changed.

Undin:
Thanks a lot!
bors r+

PR #2271, opened 2018-02-06
DOC: Fix names of recommended plugins in contributor docs

devurandom:
Change the names to reflect what the actual plugins are called, to simplify search for future contributors.

CLA will come later, since the sign-up form is currently broken.


Undin:
bors r+
Thanks!

PR #2272, opened 2018-02-06
Fix self-assignment of state in RustProjectSettingsServiceImpl

devurandom:
IntelliJ IDEA complains about this, so I removed it.

CLA will come later, since the sign-up form is currently broken.

Since the fix is so small and should not actually change anything, I am unsure which prefix to pick from https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#commit-messages

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/settings/impl/RustProjectSettingsServiceImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/settings/impl/RustProjectSettingsServiceImpl.kt
@@ -42,7 +42,7 @@ class RustProjectSettingsServiceImpl(
 
     override var data: RustProjectSettingsService.Data
         get() {
-            state = state
+            val state = state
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/project/settings/impl/RustProjectSettingsServiceImpl.kt
```diff
@@ -42,7 +42,6 @@ class RustProjectSettingsServiceImpl(
 
     override var data: RustProjectSettingsService.Data
         get() {
-            state = state
```
matklad: Ouch, what a blunder 😅 

I intended to write this as `val state = state`: this class should be thread safe, so it is nice to read the `state` field only once to gurantee that the return value is consistent, and not a mixture from several different states.

vlad20012:
bors r+
Thanks!

Issue #2267, opened 2018-02-05
Cargo project update failed: no explicit stdlib or rustup found


devurandom:
When working on the Rust source code itself, IntelliJ Rust complains "Cargo project update failed: no explicit stdlib or rustup found". I am not sure what this means, exactly.

Further, IntelliJ Rust says at the top of any file that it "Can not attach stdlib sources automatically without rustup", but when I try to "Attach manually", it tells me that "Invalid Rust standard library source path: <.../rust/src/libstd>", where rust/src/libstd is the path to the source code of libstd. What exactly is IntelliJ Rust search for, when it asks for the location of that library source code?

It appears that these error message could be enriched to make it easier for the user to solve the issue. I would be glad to contribute such a patch, but am unable to do so without additional help and information.

devurandom:
Is it possible, that what IntelliJ Rust actually wants, is the path to the Rust source code containing a "libstd" directory?

In which case it might get a little bit more automatism by looking at src/Cargo.toml and seeing that this is a copy of the Rust source code, and ask me whether I want to use that one.

devurandom:
I'd like to assign this to myself, if possible, and see whether I can come up with some wording that's more detailed and easier to understand.

devurandom:
Please note that #2275 does NOT yet detect that the currently open project is a copy of the Rust source code and hence it does not yet offer the user to just use this source.

QasimK:
I have also tried to solve this by providing the path "/usr/lib/rustlib/src/rust/src/libstd" but it is not being accepted :(

dkandalov:
The problem still exists in the latest IJ (Build #IU-182.3911.13), Rust plugin Version: 0.2.0.2101-182.

In my case the problem seemed to be with rust installed via brew. After `brew uninstall rust` I followed installation instructions here https://www.rust-lang.org/en-US/install.html, i.e. `curl https://sh.rustup.rs -sSf | sh`. I had to point IJ to `~/.cargo/bin/rustc` in a UI dialog but the error message was gone.

TriplEight:
Faced the same problem, that's because I was following setup instructions from TheBook.

To prove it won't work type $ which rustup and it won't answer.

So to avoid that in IDEA you should uninstall Rust first:

run /usr/local/lib/rustlib/uninstall.sh as root to uninstall Rust

[...]

PR #2275, opened 2018-02-07
Improve error messages, distinguish "standard library" and "Rust source"

devurandom:
The original intention was to improve error message, which was confusing
 some users.  Since the source code itself was also referring to the
 Rust source as "stdlib" or "standard library", I fixed that as well.

Fixes: intellij-rust/intellij-rust#2267

Signed-off-by: Dennis Schridde <devurandom@gmx.net>

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/configurable/RsProjectConfigurable.kt
+++ b/src/main/kotlin/org/rust/cargo/project/configurable/RsProjectConfigurable.kt
@@ -62,7 +62,7 @@ class RsProjectConfigurable(
 
         rustProjectSettings.data = RustProjectSettingsPanel.Data(
             toolchain = toolchain,
-            explicitPathToStdlib = settings.explicitPathToStdlib
+            explicitPathToRustSource = settings.explicitPathToRustSource
         )
         expandMacros = settings.expandMacros
 
@@ -82,7 +82,7 @@ class RsProjectConfigurable(
         val currentData = settings.data
         settings.data = currentData.copy(
             toolchain = rustProjectSettings.data.toolchain,
-            explicitPathToStdlib = rustProjectSettings.data.explicitPathToStdlib,
+            explicitPathToRustSource = rustProjectSettings.data.explicitPathToRustSource,
             expandMacros = expandMacros
         )
     }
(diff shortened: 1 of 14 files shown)


devurandom:
P.S. This does NOT yet detect that the currently open project is a copy of the Rust source code and hence it does not yet offer the user to just use this source.

vlad20012:
Hmm, but it's not a rust sources, it's a rust **standard library** sources :thinking:

vlad20012:
@matklad r?

devurandom:
Ok, I can try to improve that this weekend.

My situation was that I had the Rust source checked out, pointed IntelliJ Rust at the sources in rust/src/libstd and it kept complaining. Through asking around on Gitter I figured out that it did not want the standard library (shared objects or Rust libraries) and not the sources of libstd, but the sources of the whole package. So I set out to improve the error message to prevent such confusion for future users.

QasimK:
@devurandom Which path worked for you? I am trying `/usr/lib/rustlib/src/rust/src/libstd` to no avail

devurandom:
I just rebased onto current master.

Repeated research shows that intellij-rust is not happy with being given only the libstd source code, hence I still believe that "Rust source" is more appropriate and will help users better understand what is necessary.

devurandom:
I understand your concern that what IntelliJ Rust needs is not the full Rust (compiler) source, but only the source to its libraries (i.e. those listed with `type` `StdLibType.ROOT` and `StdLibType.FEATURE_GATED` in `src/main/kotlin/org/rust/cargo/util/RustCrateUtil.kt:AutoInjectedCrates`).

I think the main points of confusion for users are:
* People might understand that the path to the binary *library* is being requested
* They might understand that standard library (singular) refers to *libstd*

But:
* IntelliJ Rust wants the *source*, not the binary library
* It wants the path to the parent directory of the source of *all standard Rust libraries* (plural)

I will try to work out a patch that is less invasive and still ensures that "standard library" / "stdlib" in the Ui and in the source do not refer to both the source code of the standard libraries (pl) and the binary standard libraries (pl), which would cause confusion.

vlad20012:
I think I should close it for now.
I recognize the problem, but I think it should be solved in such order:
1. Come up with a new name. Yes, "standard library" is too ambiguous, but I don't sure that "Rust source" is much better (because we need the only stdlib sources, not entire rust sources). Maybe "Rust standard library source"?
2. Change the name in the UI only. 
3. (optional?) Refactor the code. We used to use name "stdlib" because our plugin know nothing about `.rlib` files and "rust-src" is a real "standard library" for us. So I'm not sure that we should even change anything at all.

PR #2278, opened 2018-02-08
RUN: Fix console links

Undin:
Current beta/nightly cargo returns paths relative to workspace directory but stable returns paths relative to project dir.
This fix takes into account these changes to add links in console properly

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
@@ -40,10 +40,7 @@ import org.rust.cargo.project.workspace.StandardLibrary
 import org.rust.cargo.toolchain.RustToolchain
 import org.rust.cargo.toolchain.Rustup
 import org.rust.ide.notifications.showBalloon
-import org.rust.openapiext.TaskResult
-import org.rust.openapiext.modules
-import org.rust.openapiext.pathAsPath
-import org.rust.openapiext.runAsyncTask
+import org.rust.openapiext.*
 import org.rust.stdext.AsyncValue
 import org.rust.stdext.joinAll
 import java.nio.file.Path
@@ -262,6 +259,8 @@ data class CargoProjectImpl(
             rootDirCache.set(file)
             return file
         }
+
+    override val workspaceRootDir: VirtualFile? by CachedVirtualFile(workspace?.workspaceRootPath?.toUri()?.toString())
 
     @TestOnly
     fun setRootDir(dir: VirtualFile) = rootDirCache.set(dir)
--- a/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
@@ -42,7 +42,13 @@ object CargoMetadata {
         /**
          * Ids of packages that are members of the cargo workspace
          */
-        val workspace_members: List<String>?
+        val workspace_members: List<String>?,
+
+        /**
+         * Path to workspace root folder. Can be null for old cargo version
+         */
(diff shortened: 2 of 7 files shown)


Undin:
matklad r?

Undin:
@matklad r?

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -314,6 +313,15 @@ data class CargoProjectImpl(
...
 }
 
+private inline fun AtomicReference<VirtualFile>.getCachedFile(ioFileGenerator: () -> File?): VirtualFile? {
+    val cached = get()
+    if (cached != null && cached.isValid) return cached
+    val ioFile = ioFileGenerator()
+    val file = if (ioFile != null) LocalFileSystem.getInstance().findFileByIoFile(ioFile) else null
+    set(file)
+    return file
+}
```
matklad: FWIW, I am totally unsure if this hand-rolled caching is valid :-) There are some `VirtualFilePointerSomething` classes in IntelliJ which do something similar. I've investigated them, but the results were inconclusive.

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -314,6 +313,15 @@ data class CargoProjectImpl(
         "CargoProject(manifest = $manifest)"
 }
 
+private inline fun AtomicReference<VirtualFile>.getCachedFile(ioFileGenerator: () -> File?): VirtualFile? {
```
matklad: Hm, why we parametrize it over a function? Can we give it a plain `path: Path` argument?

BTW, I think one can do someting like `val fooFile: VirtualFile by CachedVirtualFile(fooPath)`.

Undin: Oh, we've already had the separate class with this code!
But maybe we should use `VirtualFilePointer` instead of `CachedVirtualFile ` indeed? Looks like it does exactly what we need (although there isn't documentation about it)

matklad: >But maybe we should use VirtualFilePointer instead of CachedVirtualFile indeed? 

Quite probably :-)

src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -42,7 +42,12 @@ object CargoMetadata {
         /**
          * Ids of packages that are members of the cargo workspace
          */
-        val workspace_members: List<String>?
+        val workspace_members: List<String>?,
+
+        /**
+         * Path to workspace root folder. Can be null for old cargo version
+         */
+        val workspace_root: String?
```
matklad: Let's add `// BACKCOMPAT: Rust 1.23.0`, so that we remove `?` once we drop support for 1.23?

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
Undin: I've tried to use `VirtualFilePointer` in `CachedVirtualFile` to have convenient syntax with delegating but it requires `Disposable` parent and I've not found good way to pass it - all my attempts looked awful
Let's not change `CachedVirtualFile` now and I'll try to investigate it later

matklad: >Let's not change CachedVirtualFile now and I'll try to investigate it later

That was my conclusion when I've poked this thing the last time 🤣

PR #2284, opened 2018-02-11
Parse crate visibility

mrhota:
This only lets the plugin correctly parse `crate` as a visibility modifier. The plugin will still mark such items as private if they're used in other modules.


Undin:
@mhrota you can find related code [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt#L73-L112)

PR #2286, opened 2018-02-11
TYPE: clean up "extract function" refactoring

t-kameyama:
https://github.com/intellij-rust/intellij-rust/pull/2197#pullrequestreview-95631728

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
@@ -121,10 +121,11 @@ class RsExtractFunctionConfig private constructor(
     private fun typeParameterBounds(): Map<Ty, Set<Ty>> {
         return containingFunction.typeParameters.associate {
             val type = it.declaredType
-            val bounds = it.bounds.flatMap {
+            val bounds = mutableSetOf<Ty>()
+            it.bounds.flatMapTo(bounds) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -103,29 +103,28 @@ class RsExtractFunctionConfig private constructor(
...
 
-    private fun typeParameterBounds(): Map<Ty, List<Ty>> {
+    private fun typeParameterBounds(): Map<Ty, Set<Ty>> {
         return containingFunction.typeParameters.associate {
             val type = it.declaredType
             val bounds = it.bounds.flatMap {
                 it.bound.traitRef?.path?.typeArgumentList?.typeReferenceList?.flatMap { it.type.types() } ?: emptyList()
             }
-            type to bounds
+            type to bounds.toSet()
```
Undin: You can use `flatMapTo(HashSet<Ty>())` instead of `flatMap` above to put results into set immediately and avoid this transformation here

t-kameyama: use flatMapTo?

Undin: Oops, my mistake
yep, I've meant `flatMapTo`

Undin:
bors r+
Thanks!

PR #2287, opened 2018-02-12
PERF: Fulfillment (type inference) optimization

vlad20012:
This PR brings a LARGE optimization (more than x2 on mysq_async test) and opens the possibility to implement associated type bindings (`Iterator<Item=T>`)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/ProjectionCache.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/ProjectionCache.kt
@@ -1,0 +1,62 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types.infer
+
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyProjection
+
+/**
+ * The projection cache. Unlike the standard caches, this can
+ * include infcx-dependent type variables - therefore, we have to roll
+ * the cache back each time we roll a snapshot back, to avoid assumptions
+ * on yet-unresolved inference variables.
+ *
+ * Because of that, projection cache entries can be "stranded" and left
+ * inaccessible when type variables inside the key are resolved. We make no
+ * attempt to recover or remove "stranded" entries, but rather let them be
+ * (for the lifetime of the [RsInferenceContext]).
+ */
+class ProjectionCache {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/SnapshotMap.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val inner: MutableMap<K, V> = HashMap()
+    private val undoLog: UndoLog = UndoLog()
+
+    fun put(key: K, value: V): V? {
+        val oldValue = inner.put(key, value)
+        undoLog.logChange(if (oldValue == null) Insert(inner, key) else Overwrite(inner, key, oldValue))
+        return oldValue
+    }
+
+    fun get(key: K) = inner[key]
```
Undin: Let's add returning type of `get`-  I like rust rule to have explicit function signature

src/main/kotlin/org/rust/lang/core/types/infer/ProjectionCache.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    fun ambiguous(key: TyProjection) {
+        map.put(key, ProjectionCacheEntry.Ambiguous)
+            ?: error("never started projecting for `$key`")
+    }
+
+    fun error(key: TyProjection) {
+        map.put(key, ProjectionCacheEntry.Error)
+            ?: error("never started projecting for `$key`")
```
Undin: Maybe create something like
```kotlin
private fun put(key: TyProjection, value: ProjectionCacheEntry) {
    map.put(key, value) ?: error("never started projecting for `$key`")
}
```
and use it in `putTy`, `ambiguous` and `error` to avoid code duplication?

src/main/kotlin/org/rust/lang/core/types/infer/Snapshot.kt
```diff
@@ -0,0 +1,77 @@
...
+            undoLog.clear()
+        } else {
+            undoLog[position] = CommittedSnapshot
+        }
+    }
+
+    override fun rollback(){
+        assertOpenSnapshot()
+        val toRollback = undoLog.subList(position + 1, undoLog.size)
+        toRollback.reversed().forEach(Undoable::undo)
```
Undin: Looks like that `asReversed` is more efficient than `reversed` here

vlad20012:
Done

Undin:
Looks good)
bors r+

PR #2289, opened 2018-02-13
INT: move type parameter bouds from the where clause to parameter list

### LINKED ISSUE #2279: Intention to move type parameter bouds from the where clause to parameter list
Intention to convert this

```Rust
fn foo<T>() 
where
   T: Clone + Copy,
   T: Debug,
{}
```

into this:

```Rust
fn foo<T: Clone + Copy + Debug>() {}
```

In a sense, this is an opposite of this intention: https://github.com/intellij-rust/intellij-rust/blob/de894ebafb978825fcab95e455c0a89785a25c73/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToWhereClauseIntention.kt

It was introduced in this https://github.com/intellij-rust/intellij-rust/pull/762, take a look at it to learn about where to put tests/docs, etc.

t-kameyama:
Fixes #2279

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
@@ -23,10 +23,11 @@ class MoveTypeConstraintToParameterListIntention : RsElementBaseIntentionAction<
 
         val typeParameterList = whereClause.ancestorStrict<RsGenericDeclaration>()?.typeParameterList ?: return null
         val lifetimes = typeParameterList.lifetimeParameterList
-        val types = typeParameterList.typeParameterList.mapNotNull { it.name }
-        if (wherePredList.any { it.lifetime?.reference?.resolve() !in lifetimes && it.typeReference?.text !in types })
-            return null
-
+        val types = typeParameterList.typeParameterList
+        if (wherePredList.any {
+                it.lifetime?.reference?.resolve() !in lifetimes &&
+                    (it.typeReference?.typeElement as? RsBaseType)?.path?.reference?.resolve() !in types
+            }) return null
         return whereClause
     }
 
--- a/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
@@ -22,9 +22,9 @@ class MoveTypeConstraintToParameterListIntention : RsElementBaseIntentionAction<
         if (wherePredList.isEmpty()) return null
 
         val typeParameterList = whereClause.ancestorStrict<RsGenericDeclaration>()?.typeParameterList ?: return null
-        val lifetimes = typeParameterList.lifetimeParameterList.mapNotNull { it.quoteIdentifier.text }
+        val lifetimes = typeParameterList.lifetimeParameterList
         val types = typeParameterList.typeParameterList.mapNotNull { it.name }
-        if (wherePredList.any { it.lifetime?.quoteIdentifier?.text !in lifetimes && it.typeReference?.text !in types })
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsLifetimeParameter.kt
```diff
@@ -0,0 +1,18 @@
...
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsLifetime
+import org.rust.lang.core.psi.RsLifetimeParameter
+
+val RsLifetimeParameter.bounds: List<RsLifetime> get() {
+    val owner = parent?.parent as? RsGenericDeclaration
+    val whereBounds =
+        owner?.whereClause?.wherePredList.orEmpty()
+            .filter { it.lifetime?.quoteIdentifier?.text == this.quoteIdentifier.text }
```
Undin: Hm, I think we should use `resolve` here
Something like this: `.filter { it.lifetime?.reference?.resolve() == this }`

src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsWhereClause? {
+        val whereClause = element.ancestorStrict<RsWhereClause>() ?: return null
+        val wherePredList = whereClause.wherePredList
+        if (wherePredList.isEmpty()) return null
+
+        val typeParameterList = whereClause.ancestorStrict<RsGenericDeclaration>()?.typeParameterList ?: return null
+        val lifetimes = typeParameterList.lifetimeParameterList.mapNotNull { it.quoteIdentifier.text }
+        val types = typeParameterList.typeParameterList.mapNotNull { it.name }
+        if (wherePredList.any { it.lifetime?.quoteIdentifier?.text !in lifetimes && it.typeReference?.text !in types })
```
Undin: I think it's better to check equality using `resolve` approach - get reference from some `RsReferenceElement`, call `resolve` and use result of this call for equality check

src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private fun typeParameterText(param: RsNamedElement, bounds: List<RsElement>): String? {
+        val name = param.name ?: return null
+        val bs = bounds.distinctBy { it.text }
+        return buildString {
+            append(name)
+            if (bs.isNotEmpty()) {
+                append(":")
+                append(bs.joinToString("+") { it.text })
```
Undin: You can pass `:` as `prefix` parameter into `joinToString` method

src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsWhereClause? {
+        val whereClause = element.ancestorStrict<RsWhereClause>() ?: return null
+        val wherePredList = whereClause.wherePredList
+        if (wherePredList.isEmpty()) return null
+
+        val typeParameterList = whereClause.ancestorStrict<RsGenericDeclaration>()?.typeParameterList ?: return null
+        val lifetimes = typeParameterList.lifetimeParameterList
+        val types = typeParameterList.typeParameterList.mapNotNull { it.name }
+        if (wherePredList.any { it.lifetime?.reference?.resolve() !in lifetimes && it.typeReference?.text !in types })
```
Undin: Let's use `resolve` and for type parameters

t-kameyama: Fixed

Undin:
bors r+
Thanks!

PR #2291, opened 2018-02-13
Improved TypeReference grammar; supported `dyn Trait`

### LINKED ISSUE #2225: Lifetime bounds are wrongly marked as erroneous in initial positions for trait objects
All of the following is valid in Rust, but presently the variants with lifetime in front are highlighted like invalid, red-underlining the lifetime in question.

```rust
type WorksOk<'a> = Box<AsRef<str> + 'a>;
type ShowsAsError<'a> = Box<'a + AsRef<str>>;

type WorksOkAgain<'a> = Box<Fn() -> () + 'a>;
type ShowsAsErrorAgain<'a> = Box<'a + Fn() -> ()>;
```

It appears that there is conflict with normal parameter parsing, since errors are like

`Wrong number of lifetime parameters: expected 0, found 1 [E0107]`
`Wrong number of type parameters: expected 1, found 0 [E0243]`

vlad20012:
Also fixes #2225


src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -243,6 +243,9 @@ object RustParserUtil : GeneratedParserUtilBase() {
     @JvmStatic
     fun autoKeyword(b: PsiBuilder, level: Int): Boolean = contextualKeyword(b, "auto", AUTO)
 
+    @JvmStatic
+    fun dynKeyword(b: PsiBuilder, level: Int): Boolean = contextualKeyword(b, "dyn", AUTO)
```
Undin: `AUTO` -> `DYN` ?

vlad20012: =/

matklad:
Seems legit! 

I think we could use the same PSI node for both `dyn` and `impl`, sort of like we use the same for `struct` and `enum`, but I don't know whether it is a good or a bad idea =)

vlad20012:
bors try

vlad20012:
Umm :thinking: fail just reproduced.

vlad20012:
> I think we could use the same PSI node for both dyn and impl

We can merge ImplicitDynTraitType/ExplicitDynTraitType/ImplTraitType to a single PSI node, but how to name it :thinking: ? TraitType?..

src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitType.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsTraitType
+
+val RsTraitType.isImpl: Boolean get() = (stub?.isImpl) ?: impl != null
```
Undin: Oh, it's my favorite part of kotlin operator precedence 😃 
I think you meant `stub?.isImpl ?: (impl != null)` because `(stub?.isImpl) ?: impl != null` is same as `((stub?.isImpl) ?: impl) != null`

src/main/kotlin/org/rust/lang/core/types/ty/TyAnon.kt
```diff
@@ -5,15 +5,15 @@
...
 import org.rust.lang.core.psi.RsTraitItem
+import org.rust.lang.core.psi.RsTraitType
 import org.rust.lang.core.psi.ext.flattenHierarchy
 import org.rust.lang.core.types.BoundElement
 
 /**
  * Represents "impl Trait".
  */
-data class TyAnon(val definition: RsImplTraitType, val traits: List<BoundElement<RsTraitItem>>) : Ty() {
+data class TyAnon(val definition: RsTraitType, val traits: List<BoundElement<RsTraitItem>>) : Ty() {
```
Undin: Maybe add check that `definition.isImpl` to prevent possible future mistakes?

vlad20012:
Done

Undin:
LGTM
bors r+

PR #2293, opened 2018-02-14
FMT: add semicolon after break and continue

### LINKED ISSUE #2249: Formatter should add `;` after `break and continute
```rust
loop {
    if cond {
        break
    }
}
```

->

```rust
loop {
    if cond {
        break;
    }
}
```

We already have this logic for return: https://github.com/intellij-rust/intellij-rust/blob/70c9135dbcdc2347f8b85a8d4a542320b4503c97/src/main/kotlin/org/rust/ide/formatter/processors/RsReturnStatementFormatProcessor.kt.

It should be generalized to `break` and  `continue` as well.

t-kameyama:
Fixes #2249


src/main/kotlin/org/rust/ide/formatter/processors/RsStatementSemicolonFormatProcessor.kt
```diff
@@ -10,37 +10,44 @@ import com.intellij.openapi.util.TextRange
...
 import org.rust.lang.core.psi.ext.getNextNonCommentSibling
-import java.util.*
 
-class RsReturnStatementFormatProcessor : PreFormatProcessor {
+class RsStatementSemicolonFormatProcessor : PreFormatProcessor {
     override fun process(element: ASTNode, range: TextRange): TextRange {
         if (!shouldRunPunctuationProcessor(element)) return range
 
-        val returnElements = ArrayList<RsRetExpr>()
+        val returns = arrayListOf<RsRetExpr>()
```
vlad20012: Hmm, I think we don't need 3 different lists here, we can just leave a single list of `RsExpr` type

t-kameyama:
Fixed

vlad20012:
Thanks!
bors r+

PR #2294, opened 2018-02-14
TY: Fixes #2144 var with later initialization

### LINKED ISSUE #2144: Local variable initialized after declaration does not get a type annotation
In the following code:

```
fn main() {
    let answer;
    answer = 42;
}
```

the variable `answer` does not get a type annotation, the way it would if it were initialized on the same line as its declaration.

vlad20012:
Fixes #2144

```rust
fn main() {
    let a;
    a = 0;
    a;
} //^ i32
```


Undin:
Looks good!
Let's add tests with type hints because original issue was about it

vlad20012:
image
Type hint definitely works. I think that add more tests is excessive here.

PR #2295, opened 2018-02-15
Don't drop unsafe keyword on join lines

pcholakov:
Contributor agreement signed as requested :-)


vlad20012:
Looks great. Thanks a lot!
And I apologize for the long response.

bors r+

PR #2303, opened 2018-02-17
FMT: Improved braces deletion for use-groups with aliases

AtomicInteger:
FMT: Improved braces deletion for use-groups with aliases(see tests).


vlad20012:
Thanks!
bors r+

PR #2305, opened 2018-02-18
ANN: add quick fix on String type mismatch

oleg-semenov:
This commit adds quick-fix for the types mismatch when String is
expected type and impl `ToString` for actual type exists.
This fulfills part of the 2nd bullet-point of the issue:

Also unifying naming of couple of Fixes.


vlad20012:
image

Hmm, may be we should unify messages? "Convert ty **type** String using `to_string()` **method**"

Undin:
Maybe it's better to drop **type** from all such quick fixes?

oleg-semenov:
Updated it to be "Convert to String using `ToString` trait" to be more like "Convert to $ty using `From` trait". Probably, not the best naming possible, so can change to something different.

vlad20012:
Thanks!
I really like this kind of stuff. Quick fixes, intentions and so on are the things that really about *IDE* experience. Thank you for doing this while I'm _just_ working on a language support =)

bors r+

PR #2306, opened 2018-02-19
TY: Support associated type bindings

vlad20012:
We now support this stuff:
```rust
        trait Tr { type Item; }

        fn foo<B: Tr<Item=u8>>(_: B) {
            let a: B::Item = 0;
            a;
        } //^ u8
```

And it's even don't affect performance because of #2287

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsAssocTypeBinding.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsAssocTypeBinding.kt
@@ -1,0 +1,34 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import com.intellij.lang.ASTNode
+import com.intellij.psi.PsiElement
+import com.intellij.psi.stubs.IStubElementType
+import org.rust.lang.core.psi.RsAssocTypeBinding
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.resolve.ref.RsAssocTypeBindingReferenceImpl
+import org.rust.lang.core.resolve.ref.RsReference
+import org.rust.lang.core.stubs.RsAssocTypeBindingStub
+
+// Current grammar allows to write assoc type bindings in method calls, e.g.
+// `a.foo::<Item = i32>()`, so it's nullable
+val RsAssocTypeBinding.parentPath: RsPath?
+    get() = ancestorStrict()
+
+abstract class RsAssocTypeBindingMixin : RsStubbedNamedElementImpl<RsAssocTypeBindingStub>,
+                                         RsAssocTypeBinding {
+
+    constructor(node: ASTNode) : super(node)
+
+    constructor(stub: RsAssocTypeBindingStub, nodeType: IStubElementType<*, *>) : super(stub, nodeType)
+
+    override fun getReference(): RsReference = RsAssocTypeBindingReferenceImpl(this)
+
+    override val referenceNameElement: PsiElement get() = identifier
+
+    override val referenceName: String get() = stub?.name ?: referenceNameElement.text
+}
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsAssocTypeBindingReferenceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsAssocTypeBindingReferenceImpl.kt
@@ -1,0 +1,26 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.resolve.ref
+
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsAssocTypeBinding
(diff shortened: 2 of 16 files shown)


src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -98,8 +98,17 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
...
+                    val name = it.name
+                    val type = it.typeReference.type
+                    element.associatedTypesTransitively.find { it.name == name }?.let {
+                        it to type
+                    }
+                }?.toMap() ?: emptyMap()
+
                 val outputParam = lookup.fnOnceOutput
-                return@run if (outputArg != null && outputParam != null) {
+                return@run assocBindings + if (outputArg != null && outputParam != null) {
```
Undin: Let's create one mutable map and collect all results into it directly

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1167,6 +1168,35 @@ private fun RsStubLiteralType?.serialize(dataStream: StubOutputStream) {
...
+            }
+
+        override fun createPsi(stub: RsAssocTypeBindingStub): RsAssocTypeBinding =
+            RsAssocTypeBindingImpl(stub, this)
+
+        override fun createStub(psi: RsAssocTypeBinding, parentStub: StubElement<*>?) =
+            RsAssocTypeBindingStub(parentStub, this, psi.identifier.text)
+
+
+        override fun indexStub(stub: RsAssocTypeBindingStub, sink: IndexSink) {}
```
Undin: You can drop this method because `RsStubElementType` already has default empty implementation

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -33,7 +33,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
         // Bump this number if Stub structure changes
-        override fun getStubVersion(): Int = 123
+        override fun getStubVersion(): Int = 124
```
Undin: Let's increase it - current stub version is already `124`

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -98,8 +98,17 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
 
         val assocTypes = run {
             if (element is RsTraitItem) {
+                // Iterator<Item=T>
+                val assocBindings =  path.typeArgumentList?.assocTypeBindingList?.mapNotNull {
+                    val name = it.name
+                    val type = it.typeReference.type
+                    element.associatedTypesTransitively.find { it.name == name }?.let {
```
Undin: Is it ok that we check name equality for search instead of using `resolve` or something else?

src/main/kotlin/org/rust/lang/core/psi/ext/RsAssocTypeBinding.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.stubs.IStubElementType
+import org.rust.lang.core.psi.RsAssocTypeBinding
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.resolve.ref.RsAssocTypeBindingReferenceImpl
+import org.rust.lang.core.resolve.ref.RsReference
+import org.rust.lang.core.stubs.RsAssocTypeBindingStub
+
+val RsAssocTypeBinding.parentPath: RsPath
+    get() = ancestorStrict()!! // non-null guaranteed by the grammar
```
Undin: Unfortunately, it's not true
Current grammar allows writing following things
```rust
fn main() {
    let a = S(123);
    let b = a.foo::<Ty = i32>();
}
```
It throws NPE becase `a.foo::<Ty = i32>()` is `RsMethodCall`

src/main/kotlin/org/rust/lang/core/resolve/ref/RsAssocTypeBindingReferenceImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+        get() = referenceNameElement
+
+    override fun resolveInner(): List<RsElement> =
+        collectResolveVariants(element.referenceName) { processAssocTypeVariants(element, it) }
+
+    override fun getVariants(): Array<out Any> =
+        collectCompletionVariants { processAssocTypeVariants(element, it) }
+}
+
+fun resolveAssocTypeBinding(trait: RsTraitItem, binding: RsAssocTypeBinding): RsTypeAlias? =
```
Undin: Let's move it into `RsPathReferenceImpl.kt` and make it private

vlad20012:
Done

PR #2309, opened 2018-02-20
ANN: add quick fix on str/String type mismatch

oleg-semenov:
This commit adds quick-fix for the types mismatch when `&str`
or `&mut str` is expected type and `String` is actual type.
This fulfills part of the 2nd bullet-point of the issue:

Also fixing the name of ConvertToStringFix class.


vlad20012:
:+1: Thanks!

bors r+

Issue #2184, opened 2018-01-10
Handle function arguments in extract function


Undin:
The current implementation doesn't take into account reference to function arguments.
And it would be great to support it.

```rust
fn foo(a: i32) {
    <selection>println!("{}", a);</selection>
}
```

**Expected**:
```rust
fn foo(a: i32) {
    bar(a);
}

fn bar(a: i32) {
    println!("{}", a);
}
```

**Actual**:
```rust
fn foo(a: i32) {
    bar();
}

fn bar() {
    println!("{}", a); // error[E0425]: cannot find value `a` in this scope
}
```

Undin:
Looks like it works now

PR #2314, opened 2018-02-21
Support mutablity of a variable during extraction

### LINKED ISSUE #2189: Variable mutability in extracted function
Now we don't care about variable mutability in "extract function" refactoring and we can generate incorrect code in some cases

```rust
fn foo() {
    let mut vec = vec![1, 2, 3];
    <selection>vec.push(1);</selection>
}
```

**Expected:**
```rust
fn foo() {
    let mut vec = vec![1, 2, 3];
    bar(vec);
}

fn bar(mut vec: Vec<i32>) {
    vec.push(1);
}
```

or even

```rust
fn foo() {
    let mut vec = vec![1, 2, 3];
    bar(&mut vec);
}

fn bar(vec: &mut Vec<i32>) {
    vec.push(1);
}
```

**Actual:**
```rust
fn foo() {
    let mut vec = vec![1, 2, 3];
    bar(vec);
}

fn bar(vec: Vec<i32>) {
    vec.push(1); // error[E0596]: cannot borrow immutable argument `vec` as mutable
}
```

farodin91:
Fixes #2189
Added a test for #2184 works without any changes.

@Undin r?


Undin:
LGTM
Thanks!
bors r+

PR #2315, opened 2018-02-21
Introduce & optimize `TyAdt` as a replacement for `TyStruct` & `TyEnum`

vlad20012:


src/main/kotlin/org/rust/lang/core/types/ty/TyAdt.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Represents struct/enum/union.
+ * "ADT" may be read as "Algebraic Data Type".
+ * The name is inspired by rustc
+ */
+@Suppress("DataClassPrivateConstructor")
+data class TyAdt private constructor(
+    val item: RsStructOrEnumItemElement,
+    val typeArguments: List<Ty>
```
vlad20012: I think we can even use a raw Array here

Undin: I think it won't bring any benefits

PR #2318, opened 2018-02-22
TYPE: Properly pass variables into extracted function

### LINKED ISSUE #2188: Properly pass variables into extracted function
We always pass variables by value and it can lead to incorrect code.
We should correctly determine how to pass variable: by value or by reference.

```rust
fn foo() {
    let vec = vec![1, 2, 3];
    <selection>println!("{}", vec.len());</selection>
    println!("{}", vec[0]);
}
```

**Expected:**
```rust
fn foo() {
    let vec = vec![1, 2, 3];
    bar(&vec);
    println!("{}", vec[0]);
}

fn bar(vec: &Vec<i32>) {
    println!("{}", vec.len());
}
```

**Actual:**
```rust
fn foo() {
    let vec = vec![1, 2, 3];
    bar(vec);
    println!("{}", vec[0]); // error[E0382]: use of moved value: `vec`
}

fn bar(vec: Vec<i32>) {
    println!("{}", vec.len()); 
}
```

t-kameyama:
Fixes #2188

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
@@ -11,10 +11,7 @@ import com.intellij.psi.search.LocalSearchScope
 import com.intellij.psi.search.searches.ReferencesSearch
 import org.rust.ide.presentation.insertionSafeText
 import org.rust.ide.utils.findStatementsInRange
-import org.rust.lang.core.psi.RsExpr
-import org.rust.lang.core.psi.RsFunction
-import org.rust.lang.core.psi.RsPatBinding
-import org.rust.lang.core.psi.RsTypeParameter
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.ty.*
@@ -42,21 +39,23 @@ class ReturnValue(val expression: String?, val type: Ty) {
 class Parameter(
     val name: String,
     val type: Ty? = null,
-    val reference: Reference = Reference.NONE
+    val reference: Reference = Reference.NONE,
+    isMutableValue: Boolean = false
 ) {
 
     enum class Reference(val text: String) {
-        MUTABLE("&mut "), BORROWING("& "), NONE("")
+        MUTABLE("&mut "), IMMUTABLE("& "), NONE("")
     }
 
     companion object {
-        fun direct(value: RsPatBinding, requiredBorrowing: Boolean): Parameter {
+        fun direct(value: RsPatBinding, requiredBorrowing: Boolean, requiredMutableValue: Boolean): Parameter {
             val reference = when {
+                requiredMutableValue -> Reference.NONE
                 value.mutability.isMut -> Reference.MUTABLE
-                requiredBorrowing -> Reference.BORROWING
+                requiredBorrowing -> Reference.IMMUTABLE
                 else -> Reference.NONE
             }
-            return Parameter(value.referenceName, value.type, reference)
+            return Parameter(value.referenceName, value.type, reference, requiredMutableValue)
         }
 
         fun self(value: String): Parameter {
@@ -64,8 +63,10 @@ class Parameter(
         }
     }
 
+    private val mut = if (isMutableValue) "mut " else ""
+
     val parameterText: String
-        get() = if (type != null) "$name: ${reference.text}${type.insertionSafeText}" else name
+        get() = if (type != null) "$mut$name: ${reference.text}${type.insertionSafeText}" else name
 
     val argumentText: String
         get() = "${reference.text}$name"
@@ -162,11 +163,21 @@ class RsExtractFunctionConfig private constructor(
                     if (it.textOffset > start) return@mapNotNull null
                     val result = ReferencesSearch.search(it, LocalSearchScope(fn))
                     if (result.none { it.element.textOffset in start..end }) return@mapNotNull null
+
                     val requiredBorrowing = result.any { it.element.textOffset > end }
                         && it.type !== TyBool && it.type !== TyChar && it.type !is TyInteger && it.type !is TyFloat
                         && it.type !is TyReference
                         && !implLookup.isCopy(it.type)
-                    Parameter.direct(it, requiredBorrowing)
+
+                    val requiredMutableValue = it.mutability.isMut && result.any {
+                        val element = it.element
+                        if (element.textOffset !in start..end) return@any false
+                        if (element.ancestorStrict<RsValueArgumentList>() == null) return@any false
+                        val unaryExpr = element.ancestorStrict<RsUnaryExpr>()
+                        unaryExpr == null || unaryExpr.operatorType == UnaryOperator.REF_MUT
+                    }
+
+                    Parameter.direct(it, requiredBorrowing, requiredMutableValue)
                 }
                 .toMutableList()
 
--- a/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
@@ -17,10 +17,7 @@ import org.rust.lang.core.psi.RsPatBinding
 import org.rust.lang.core.psi.RsTypeParameter
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ImplLookup
-import org.rust.lang.core.types.ty.Mutability
-import org.rust.lang.core.types.ty.Ty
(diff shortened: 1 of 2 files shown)


Undin:
Rebase onto master, please)

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -150,12 +155,14 @@ class RsExtractFunctionConfig private constructor(
...
             val parameters = letBindings
-                .filter { it.textOffset <= start }
-                .filter {
-                    ReferencesSearch.search(it, LocalSearchScope(fn))
-                        .any { ref -> ref.element.textOffset in start..end }
+                .mapNotNull {
+                    if (it.textOffset > start) return@mapNotNull null
+                    val result = ReferencesSearch.search(it, LocalSearchScope(fn))
+                    if (result.none { it.element.textOffset in start..end }) return@mapNotNull null
+                    val requiredBorrowing = result.any { it.element.textOffset > end }
```
Undin: Let's use a little bit different strategy here - pass a variable with minimal required "attribute". I mean if it is enough to pass only reference, let's pass reference even we can move variable and don't break code.
So, for 
```rust
fn foo() {
    let mut v = Vec::new();
    <selection>v.push(123);</selection>
}
```
result should be
```rust
fn foo() {
    let mut v = Vec::new();
    bar(&mut v);
}

fn bar(v: &mut Vec<i32>) {
    v.push(123);
}
```

Undin: Also current implementation can pass references by reference
```rust
fn foo() {
    let s = "str";
    <selection>println!("{}", s);</selection>
    println!("{}", s);
}
```
=>
```rust
fn foo() {
    let s = "str";
    bar(&s);
    println!("{}", s);
}

fn bar(s: &&str) {
    println!("{}", s);
}
```

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -150,12 +155,14 @@ class RsExtractFunctionConfig private constructor(
...
+                        && it.type !is TyPrimitive && !ImplLookup.relativeTo(it).isCopy(it.type)
```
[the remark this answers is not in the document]
Undin: Hm, looks like that not all primitive types implement `Copy` trait (for example `str`).
And add a comment why we have this check here to drop it when `ImplLookup. isCopy` will support primitive types

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -40,28 +42,31 @@ class ReturnValue(val expression: String?, val type: Ty) {
     }
 }
 
-class Parameter(val name: String, val type: Ty? = null, val mutability: Mutability = Mutability.IMMUTABLE) {
+class Parameter(val name: String, val type: Ty? = null, val mutability: Mutability = Mutability.IMMUTABLE,
```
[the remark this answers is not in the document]
t-kameyama: It seems `ImplLookup.isCopy` does not support primitive types.

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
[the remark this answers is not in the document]
t-kameyama: It seems ImplLookup.isCopy does not support primitive types.

src/test/kotlin/org/rust/lang/refactoring/RsExtractFunctionTest.kt
```diff
@@ -726,6 +726,76 @@ class RsExtractFunctionTest : RsTestBase() {
         false,
         "bar")
 
+        fun `test extract a function with reference parameter`() = doTest("""
```
Undin: Let's split this test into several small tests to check one case per tests
I think it's more convenient to check what cases are tested

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -40,28 +39,36 @@ class ReturnValue(val expression: String?, val type: Ty) {
...
 
-class Parameter(val name: String, val type: Ty? = null, val mutability: Mutability = Mutability.IMMUTABLE) {
+class Parameter(
+    val name: String,
+    val type: Ty? = null,
+    val reference: Reference = Reference.NONE
+) {
+
+    enum class Reference(val text: String) {
+        MUTABLE("&mut "), BORROWING("& "), NONE("")
```
Undin: What about passing a variable by mutable value?
```rust
fn foo(mut v: Vec<i32>) {}

fn main() {
    let mut v = Vec::new();
    <selection>v.push(1);
    foo(v);</selection>
}
```
should be
```rust
fn foo(mut v: Vec<i32>) {}

fn main() {
    let mut v = Vec::new();
    bar(v);
}

fn bar(mut v: Vec<i32>) {
    v.push(1);
    foo(v);
}

```
but current version generates the following
```rust
fn foo(mut v: Vec<i32>) {}

fn main() {
    let mut v = Vec::new();
    bar(&mut v);
}

fn bar(v: &mut Vec<i32>) {
    v.push(1);
    foo(v);
}

```

t-kameyama:
Fixed

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -149,13 +157,28 @@ class RsExtractFunctionConfig private constructor(
...
-                .filter { it.textOffset <= start }
-                .filter {
-                    ReferencesSearch.search(it, LocalSearchScope(fn))
-                        .any { ref -> ref.element.textOffset in start..end }
+                .mapNotNull {
+                    if (it.textOffset > start) return@mapNotNull null
+                    val result = ReferencesSearch.search(it, LocalSearchScope(fn))
+                    if (result.none { it.element.textOffset in start..end }) return@mapNotNull null
+
+                    val requiredBorrowing = result.any { it.element.textOffset > end }
```
Undin: `result.any { it.element.textOffset > end }` condition is not enough here
In the following case we still can pass variable by reference
```rust
fn foo() {
    let a = Vec::new();
    <selection>f(&a);</selection>
}

fn f(a: &Vec<i32>) {}
```
but pass it by value
```rust
fn foo() {
    let a = Vec::new();
    bar(a);
}

fn bar(a: Vec<i32>) {
    f(&a);
}

fn f(a: &Vec<i32>) {}
```

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -149,13 +157,28 @@ class RsExtractFunctionConfig private constructor(
...
+                        && it.type !== TyBool && it.type !== TyChar && it.type !is TyInteger && it.type !is TyFloat
```
Undin: Let's add a comment here to explain why we manually check primitive types

Undin: And looks like there isn't difference between `!==` and `!=` here

src/main/kotlin/org/rust/lang/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -40,28 +36,40 @@ class ReturnValue(val expression: String?, val type: Ty) {
...
+    enum class Reference(val text: String) {
+        MUTABLE("&mut "), IMMUTABLE("& "), NONE("")
+    }
+
     companion object {
-        fun direct(value: RsPatBinding): Parameter {
-            return Parameter(value.referenceName, value.type, value.mutability)
+        fun direct(value: RsPatBinding, requiredBorrowing: Boolean, requiredMutableValue: Boolean): Parameter {
+            val reference = when {
+                requiredMutableValue -> Reference.NONE
```
Undin:
```rust
fn foo(mut a: Vec<i32>) {
    <selection>f(&mut a);</selection>
    f(&mut a);
}

fn f(a: &mut Vec<i32>) {}
```
=>
```rust
fn foo(mut a: Vec<i32>) {
    bar(a);
    f(&mut a); // error[E0382]: use of moved value: `a`
}

fn bar(mut a: Vec<i32>) {
    f(&mut a);
}

fn f(a: &mut Vec<i32>) {}
```

t-kameyama:
Fixed

Undin:
Sorry for so slow response
Thanks a lot!
bors r+

PR #2321, opened 2018-02-22
PERF: better cache macros during resolve

vlad20012:
It increases performance up to 20% on `mysq_async` test.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -8,6 +8,7 @@
 package org.rust.lang.core.resolve
 
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Key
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -400,18 +402,16 @@ fun processAssocTypeVariants(trait: RsTraitItem, processor: RsResolveProcessor):
     return false
 }
 
-val RsFile.exportedCrateMacros: List<RsMacroDefinition>
-    get() = CachedValuesManager.getCachedValue(this) {
-        val macros = exportedCrateMacros(this, true)
+private val EXPORTED_MACROS_KEY: Key<CachedValue<List<RsMacroDefinition>>> = Key.create("123")
+private val ALL_MACROS_KEY: Key<CachedValue<List<RsMacroDefinition>>> = Key.create("321")
```
Undin: Don't forget change string names of these keys

PR #2324, opened 2018-02-24
ANN: Fixed #2082 correctly display generic type in tooltips

### LINKED ISSUE #2082: Wrong HTML escaping in E0308 tooltips causes problems around < and >
Example code:
```rust
struct Example<T>(T);

fn testit() {
    let wrongAssignment: Example<u32> = Example(Example(1u8));
}
```

html_escaping_e0308
It wants to output `expected 'Example<u32>', found 'Example<Example<u8>>'`
but `LocalInspectionPass.createHighlightInfo()` calls `@NonNls String message = ProblemDescriptorUtil.renderDescriptionMessage(descriptor, element);`
which replaces '&lt;' by '<' but leaves `<html>` tags around the message.

This error was inspired by https://github.com/intellij-rust/intellij-rust/issues/1753#issuecomment-331313940 where the error message was probably OK.

Rustc error message:
```
error[E0308]: mismatched types
   --> src/main.rs:240:49
    |
240 |     let wrongAssignment: Example<u32> = Example(Example(1u8));
    |                                                 ^^^^^^^^^^^^ expected u32, found struct `Example`
    |
    = note: expected type `u32`
               found type `Example<u8>`
```

vlad20012:
Fixes #2082

Temorary replace `<` & `>` with `❮` & `❯`. But I don't sure it's the best possible symbols.

image

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -99,10 +99,18 @@ sealed class RsDiagnostic(
         }
 
         private fun expectedFound(expectedTy: Ty, actualTy: Ty): String {
-            val expectedTyS = escapeString(expectedTy.toString())
-            val actualTyS = escapeString(actualTy.toString())
+            val expectedTyS = escapeTy(expectedTy.toString())
+            val actualTyS = escapeTy(actualTy.toString())
             return "expected `$expectedTyS`, found `$actualTyS`"
         }
+
+        // BACKCOMPAT: ???
+        // Fix for IntelliJ platform bug: https://youtrack.jetbrains.com/issue/IDEA-186991
+        // replace it with `escapeString()` after the end of support IDEs with the bug
+        private fun escapeTy(str: String): String = str
+            .replace("<", "&#60;")
+            .replace(">", "&#62;")
+            .replace("&", "&amp;")
     }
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -99,10 +99,17 @@ sealed class RsDiagnostic(
...
+            val actualTyS = escapeTy(actualTy.toString())
             return "expected `$expectedTyS`, found `$actualTyS`"
         }
+
+        // BACKCOMPAT: ???
+        // Fix for IntelliJ platform bug: https://youtrack.jetbrains.com/issue/IDEA-186991
+        // replace it with `escapeString()` after the end of support IDEs with the bug
+        private fun escapeTy(str: String): String = str
+            .replace("<", "&#10094;")
+            .replace(">", "&#10095;")
```
Undin: You can use code of common angle brackets (60 and 62) here - it works fine with such escaping

Undin: And are you sure that it is enough for escaping? what about `&`?
maybe create a copy of `XmlStringUtil.escapeString` with correct escaping of `<` and `>`?

vlad20012:
Ok, replacing with `&#60;` & `&#62;` codes works fine

PR #2325, opened 2018-02-24
COMP: Filter method completion variants by trait bounds of impls

vlad20012:
After supporting of `impl<T> for T` we've got a tons of irrelevant method completions for ANY TYPE. This PR tries to fix it via trait bounds checking.


Undin:
LGTM
bors r+

PR #2329, opened 2018-02-25
TY: Correctly typify length of byte string literals containing escape characters

kumbayo:
Byte string literals like b"Example" are &[u8; ?] arrays.
Currently we infer the array length from the raw text length.
But if the string contains escape characters this will give us
the wrong array length.

Instead first unescape the string and then check its length.

This also requires bumping he stub version
as the literal length is stored in stubs.


Undin:
bors r+
Thanks!

PR #2334, opened 2018-02-26
Add a workaround for loading of the intellij-toml plugin

kumbayo:
The current way (since #2213) of loading the intellij-toml plugin
from build.gradle.kts requires a fix for
JetBrains/gradle-intellij-plugin#238.
But the current latest stable version 0.2.18 of gradle-intellij-plugin
does not yet contain this fix.
This causes toml support to not work during development.

Go back to the old way of loading the intellij-toml plugin
until a bugfix version of intellij-toml containing the fix is released.

See JetBrains/gradle-intellij-plugin#258 for more details.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -104,7 +104,7 @@ project(":") {
     intellij {
         pluginName = "intellij-rust"
 //        alternativeIdePath = "deps/clion-$clionVersion"
-        setPlugins(project(":intellij-toml"))
+        setPlugins("org.toml.lang:0.2.0.12")
     }
 
     repositories {


kumbayo:
@Undin would this workaround make sense?
This way to specify/load the plugin seems to worked before so i am not quite sure how correct it is.
(and what are its tradeoffs)

Or should we wait until somebody can create a bugfix version of JetBrains/gradle-intellij-plugin?

Undin:
Let's merge it because it's really inconvenient without toml plugin and will revert when `gradle-intellij-plugin` will be able to load local plugin dependencies too

The original issue was the following: if you try to support new IDE version (in #2213 it was 2018.1) all your plugin dependencies should be compatible with this version. But then toml 0.2.0.9 didn't support 2018.1. The first decision is to build new toml plugin version, upload it and update the corresponding dependency for rust plugin. But it doesn't work because of the following reasons: 
* toml plugin and rust plugin are one gradle project. 
* toml and rust plugin have the same `ideaVersion` property. 
* `gradle-intellij-plugin` downloads and checks all plugin dependencies while gradle configuration phase. Note gradle runs configuration phase for all project even you try to run task for some subproject

So, when you try to build toml plugin with new IDE version , you need to change `ideaVersion`. It forces you to upgrade `until-build` in `plugin.xml` for rust plugin too because otherwise `gradle-intellij-plugin` says that rust plugin is not compatible with new IDE dependency. But if you do it, you will get another error - toml dependency of rust plugin is not compatible with new IDE version 😆 

Yes, we can separate version of IDE dependency for toml and rust plugins but looks like it's not good idea. So to have dependency on local toml plugin in rust plugin is the best solution (in my opinion, of course) except issue about loading local plugin dependency

Undin:
bors r+
Thanks!

kumbayo:
Thanks for the detailed explanation!

PR #2340, opened 2018-03-03
Improvements of trait types support

Undin:
* implement move left/right action for bounds of trait types (trait objects, `impl`/`dyn Trait`)
* support `goto type declaration` for `impl Trait`


Undin:
@vlad20012 I think these changes can complicate fixing bug with endless indexing so merge it when you'll manage with that bug

PR #2341, opened 2018-03-04
ANN: add type mismatch quick fix using Borrow/BorrowMut and AsRef/Mut

oleg-semenov:
This commit adds series of quick-fix for the types mismatch when
impl `Borrow<B>`/`BorrowMut<B>/AsRef<B>/AsMut<B>` for A exists
where A is actual type and B is expected type,
which fulfills part of the forth bullet-point of the issue:

Also refactoring a couple of existing quick-fixes of the similar form
to reduce code duplication a little bit, and to make keeping the
messaging/naming unified easier.

Also fixing ConvertToOwnedTyFixTest.


src/main/kotlin/org/rust/ide/annotator/fixes/ConvertToRefTyFix.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.types.ty.Ty
+
+/**
+ * For the given `expr` converts it to the reference type with `as_ref()` method.
+ */
+class ConvertToRefTyFix(expr: PsiElement, ty: Ty) : ConvertToTyUsingTraitFix(expr, ty, "AsRef", "as_ref")
```
vlad20012: I think it's ok to move all these single-line classes to `ConvertToTyUsingTraitFix.kt` file

oleg-semenov: Done

vlad20012:
Thanks! :+1::+1::+1::fire::fire::fire:
bors r+

PR #2350, opened 2018-03-05
Support new inclusive range syntax a..=b

### LINKED ISSUE #2335: Wrong Formatting on inclusive_range_syntax
```
#![feature(inclusive_range_syntax)]
...
for i in 1..=10 {
```
When pressing ctrl+alt+L, it adds a space between `..` and `=`.

```
for i in 1.. = 10 {
```

kumbayo:
Improve formatting of inclusive range syntax.
Support exclusive range syntax a..b in range patterns (if let, match arms).

Fixes #2335


src/test/kotlin/org/rust/lang/core/type/RsStdlibExpressionTypeInferenceTest.kt
```diff
@@ -74,6 +74,15 @@ class RsStdlibExpressionTypeInferenceTest : RsTypificationTestBase() {
         }
     """)
 
+    fun `test RangeToInclusive new syntax`() = stubOnlyTypeInfer("""
+    //- main.rs
+        fn main() {
+            let x = ..=42u16;
+            x
```
Undin: Let's add `;` to have correct code in tests
It doesn't affect type inference here but I think it's a good habit to write correct code in such tests to avoid weird results in some cases

kumbayo: Thanks fixed, i also adjusted the code to make sure we do not reintroduce type errors in the future.

kumbayo:
bors r=Undin

PR #2352, opened 2018-03-06
INSP: Auto import improvements

### LINKED ISSUE #2347: Autoimport add `extern crate %crate_name%;` that already imported in main.rs/lib.rs
Autoimport add `extern crate %crate_name%;` that already imported in main.rs/lib.rs:
image

Undin:
* don't import path in use item
* don't highlight references with multiple resolutions
* correctly import items from extern crate

Fixes #2347

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/import/AutoImportFix.kt
@@ -117,6 +117,13 @@ class AutoImportFix(path: RsPath) : LocalQuickFixOnPsiElement(path), HighPriorit
             if (TyPrimitive.fromPath(path) != null) return null
             val basePath = getBasePath(path)
             if (basePath.reference.resolve() != null) return null
+
+            // Don't try to import path in use item
+            if (path.ancestorStrict<RsUseSpeck>() != null) {
+                Testmarks.pathInUseItem.hit()
+                return Context(basePath, emptyList())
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/inspections/fixes/import/AutoImportFix.kt
```diff
@@ -116,7 +136,14 @@ class AutoImportFix(path: RsPath) : LocalQuickFixOnPsiElement(path), HighPriorit
         fun findApplicableContext(project: Project, path: RsPath): Context? {
             if (TyPrimitive.fromPath(path) != null) return null
             val basePath = getBasePath(path)
-            if (basePath.reference.resolve() != null) return null
+            if (basePath.reference.multiResolve().isNotEmpty()) return null
+
+            // Don't try to import path in use item
+            if (path.ancestorStrict<RsUseSpeck>() != null) {
```
matklad: This check seems cheaper than multiresolve, so let's do it first?

Undin: Order is important here because we use `findApplicableContext` in `RsUnresolvedReferenceInspection`
If result of `findApplicableContext` is null => this path shouldn't be highlighted as unresolved
If result is not null then the path should be highlighted even it's in `use` item but without quick fix

vlad20012:
:+1: 
bors r+

PR #2360, opened 2018-03-09
GRD: Local toml plugin dependency

Undin:
* update kotlin, gradle & intellij-gradle-plugin
* revert changes of #2334 because intellij-gradle plugin correctly loads local plugin dependencies since 0.2.19
* fix intellij-toml version suffix


kumbayo:
Works nicely. Thanks!
bors r+

PR #2366, opened 2018-03-12
INSP: filter import candidates by namespace

Undin:


src/main/kotlin/org/rust/lang/core/resolve/Namespace.kt
```diff
@@ -11,10 +11,16 @@ import org.rust.lang.core.psi.ext.RsNamedElement
 import java.util.*
 
 enum class Namespace(val itemName: String) {
-    Values("value"), Types("type"), Lifetimes("lifetime"), Macros("macro")
+    Values("value"),
+    Types("type"),
+    Lifetimes("lifetime"),
+    Macros("macro"),
+    Traits("trait")
```
matklad: Not sure that this is a good idea :) All previous namespaces are disjoints, you can have a value, a type, a lifetime and a macro in the same scope which share a name but do not conflict. It's not true for traits, because traits are in types namespace:

https://play.rust-lang.org/?gist=f8686cc9716100527bf19468e8fbbdda&version=stable

More concretely, `Namespace`'s main purpose is to direct name resolution, and, if we don't single out traits specifically during the resolution, we probably should not have this namespace variant. I suggest creating some additional mechanism for filtering items based on their types.

Undin: What do you think about separate `Namespace` enum only for auto import with related changes? It will lead to some code duplication but I like current approach and it seems if I use current implementation of namespace and some other mechanism for traits it will be rather ugly

matklad: Hm, I am not sure if a separate enum is required... Could one use a `(PsiElement) -> Boolean` for `expectedNamespaces`? Ie, use a lambda function as a condition, and filter by namespace (or, for trait references, by the class of `PsiElement`) inside?

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -423,6 +426,40 @@ sealed class ImportInfo {
...
+            is RsTraitItem,
+            is RsTypeAlias -> true
+            else -> false
+        }
+    }
+    is RsPathExpr -> { e ->
+        when (e) {
+            is RsConstant,
+            is RsFunction -> true
+            is RsFieldsOwner -> e.blockFields == null
```
vlad20012: There is a case when we typing a struct literal and want to import struct before typing `{`, i.e.
```rust
let a = Foo/*caret*/
```
This `Foo` will be parsed as `RsPathExpr`, but we may want to import a struct.
```rust
let a = Foo { ... };
```
In short, I think this `e.blockFields == null` is not necessary

Undin: > In short, I think this e.blockFields == null is not necessary

I think we should understand such things to more precise suggestions in quick fix but your case with typing is reasonable so let's drop check of fields type for now

> P.S. + RsTypeAlias
Do you mean to add `RsTypeAlias` as yet another element in `when`? But I can't come up with case when you can use type alias in path expr. Maybe you mean such cases for `RsStructLiteral` branch?

src/main/kotlin/org/rust/lang/core/psi/ext/RsTypeAlias.kt
```diff
@@ -20,6 +20,16 @@ import javax.swing.Icon
 val RsTypeAlias.default: PsiElement?
     get() = node.findChildByType(DEFAULT)?.psi
 
+fun RsTypeAlias.baseType(): RsElement? {
+    var base: RsElement? = this
+    val visited = mutableSetOf<RsElement>(this)
```
kumbayo: This newly extracted method seems to prevent endless loops when a type aliases form a cycle.
Can you also update the very similar code in `TypeInference.kt` `inferStructLiteralType()` to do the same?
If this loop prevention is intentional, I guess it would also make sense to add tests for it and to mention it in the commit message?

Undin:
@kumbayo I created separate PR about cyclic type aliases

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
vlad20012: > RsTypeAlias

I meant the same case with incomplete struct literal, where we refer not a `RsFieldsOwner` directly, but a `RsTypeAlias`

Undin: Oh, okay:)
I've fixed it too

Issue #1806, opened 2017-10-02
No type inference for elements of arrays


doxxx:
In the following code:

```rust
let a: [f64; 3] = [1.0, 2.0, 3.0];
let b = a[0];
let c = b.powi(2);
```

The type of `b` is not inferred -- it's shown as `<unknown>` -- and `powi` is not resolvable and thus not highlighted as a method of `f64`.

Plugin Version: 0.1.0.2069

vlad20012:
Ok, this should be fixed by these 3 PRs

Undin:
Fixed

PR #2368, opened 2018-03-14
TY: take into account type arguments when projecting associated types

vlad20012:
Part of #1806

### DIFF
--- a/src/main/kotlin/org/rust/ide/presentation/Ty.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/Ty.kt
@@ -61,7 +61,9 @@ private fun render(
         }"
         is TyPointer -> "*${if (ty.mutability.isMut) "mut" else "const"} ${r(ty.referenced)}"
         is TyTypeParameter -> ty.name ?: anonymous
-        is TyProjection -> "<${ty.type} as ${ty.trait.name}>::${ty.target.name}"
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/presentation/Ty.kt
```diff
@@ -61,7 +61,9 @@ private fun render(
         }"
         is TyPointer -> "*${if (ty.mutability.isMut) "mut" else "const"} ${r(ty.referenced)}"
         is TyTypeParameter -> ty.name ?: anonymous
-        is TyProjection -> "<${ty.type} as ${ty.trait.name}>::${ty.target.name}"
+        is TyProjection -> "<${ty.type} as ${ty.trait.element.name ?: return anonymous}${
+        if (includeTypeArguments) formatTraitTypeArguments(ty.trait, r, includeAssoc = false) else ""
+        }>::${ty.target.name}"
```
Undin: Maybe split this line into some separate statements?
It's rather hard to understand what happens - too many operations per line
Also a little bit weird formatting

vlad20012: There are other branches that look as awful. This function needs to be cleaned up. But we can do it in another PR.

Undin: Okay:)

PR #2374, opened 2018-03-15
TY&RES: find impls for unconstrained integers/floats

vlad20012:
Now we corrrectly resolve method calls on unconstrained integers/floats, e.g. `0.foo()`.

Also a part of #1806

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/TyFingerprint.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/TyFingerprint.kt
@@ -23,34 +23,39 @@ data class TyFingerprint constructor(
     companion object {
 
         val TYPE_PARAMETER_FINGERPRINT = TyFingerprint("#T")
+        private val ANY_INTEGER_FINGERPRINT = TyFingerprint("{integer}")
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/types/TyFingerprint.kt
```diff
@@ -23,34 +23,38 @@ data class TyFingerprint constructor(
...
         val TYPE_PARAMETER_FINGERPRINT = TyFingerprint("#T")
+        private val ANY_INTEGER_FINGERPRINT = TyFingerprint("{integer}")
+        private val ANY_FLOAT_FINGERPRINT = TyFingerprint("{float}")
 
         // Keep in sync with Declarations-inferTypeReferenceType
-        fun create(ref: RsTypeReference, typeParameters: List<String>): TyFingerprint? {
+        fun create(ref: RsTypeReference, typeParameters: List<String>): List<TyFingerprint> {
             val type = ref.typeElement
-            return when (type) {
+            return listOf(when (type) {
```
Undin: Let's split it into two statements:
```kotlin
val fingerprint = when (type) { ... }
return listOf(fingerprint)
```

vlad20012:
Done

PR #2375, opened 2018-03-16
GRAM: Fixed #2371 generate correct PSI and emit parse error on `() + A`

### LINKED ISSUE #2371: Keep getting () error
I keep getting [this](https://gist.github.com/Juici/1ed61c63e2f451a3f8d3afbb66dbfaf6) error whenever I try to do code completion.

```
CLion 2018.1 EAP
Build #CL-181.4203.106, built on March 14, 2018
CLion EAP User
Expiration date: April 13, 2018
JRE: 1.8.0_152-release-1136-b20 amd64
Linux 4.14.25-1-MANJARO
```
```
rustc 1.26.0-nightly (2789b067d 2018-03-06)
cargo 0.26.0-nightly (1d6dfea44 2018-01-26)
rustup 1.11.0 (e751ff9f8 2018-02-13)
```
\
EDIT: Might be related to this? https://github.com/rust-lang-nursery/rls/issues/756
\
EDIT2: Running `rls` command from terminal results
```
$ rls
error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not have the binary `rls`
```

vlad20012:


Undin:
I've found out where our grammar failed
```rust
pub trait CbFunc: Send {
    /// Calls the function.
    fn call_box(self: Box<Self>, &mut Cursive);
}

impl<F: FnOnce(&mut Cursive) -> () + Send> CbFunc for F {
    fn call_box(self: Box<Self>, siv: &mut Cursive) {
        (*self)(siv)
    }
}
```

Current master grammar parses `FnOnce(&mut Cursive) -> () + Send` as `FnOnce` with `() + Send` ret type

src/test/resources/org/rust/lang/core/parser/fixtures/partial/types.rs
```diff
@@ -1 +1,3 @@
 fn foo<'a, 'b>() where for<'c> 'a : 'b {}
+
+type T = () + A;
```
Undin: Let's add something like `type T = FnOnce(&mut i32) -> () + Send` into complete parser tests to ensure that it's parsed correctly

src/main/grammars/RustParser.bnf
```diff
@@ -703,8 +703,9 @@ private TypeReferenceInner ::= ArrayType
                                // External rule to find the difference between `(T)` and `(T,)`
                                | <<tupleOrParenType TypeReferenceInner TupleType_upper>>
                                | ExplicitTraitType
+                               | TrivialBaseType
                                // External rule to find the difference between `T` and `T+T`
-                               | <<baseOrTraitType BaseTypeInner ImplicitTraitType TraitType_upper>>
+                               | <<baseOrTraitType PathGenericArgsWithoutColons ImplicitTraitType TraitType_upper>>
```
Undin: It would be great to parse `() + Send` too and add error annotation via some inspection
But I think it can be done in other PR - looks like that it will require a lot of changes

Undin: Hm, we steel parse `FnOnce() -> i32 + Send` as `FnOnce` with `i32 + Send` ret type

Undin:
Grammar looks good
But it seems that we should bump stub version because of these changes

PR #2379, opened 2018-03-18
COMP: implement crates.io deps completion

### LINKED ISSUE #2190: autocompletion in [dependencies] in Cargo.toml
Hi.
Currently, if one wants to specify dependency in Cargo.toml
he ought to go to crates.io, find crate, look at version information 
and only after that he can specify dependency by specifying version.

Could you please add autocompletion of crate's version?
For instance, I type toml = "
then I press Ctrl-Space and it shows 0.4, 0.4.5 and few minor versions too.

vitvakatu:
closes #2190 

In fact, it's my first Kotlin code, so be careful during review :wink:


src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
```diff
@@ -58,6 +68,34 @@ class CargoTomlKeysCompletionProvider : CompletionProvider<CompletionParameters>
         })
     }
 
+    private fun dependencies(parent: TomlKeyValue): Collection<String> {
+        val response = try {
+            val name = parent.text.replace("IntellijIdeaRulezzz", "")
```
vlad20012: Let's use `CompletionUtil.getOriginalElement(parent)?.text ?: ""` instead.

src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
```diff
@@ -58,6 +68,34 @@ class CargoTomlKeysCompletionProvider : CompletionProvider<CompletionParameters>
...
+                .use { it.readText() }
+        } catch (e: IOException) {
+            parent.project.showBalloon("Could not reach crates.io", NotificationType.WARNING)
+            return emptyList()
+        }
+        return Gson().fromJson(response, Crates::class.java).crates.map { "${it.name} = \"${it.maxVersion}\"" }
+    }
+
+    private fun isDependenciesTable(tableName: String): Boolean {
+        /* Matches all relevant names of table, like [dependencies], [target.'cfg(unix)'.dev-dependencies],
```
vlad20012: We prefer not to use multi-line comments

src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt
vitvakatu: Thanks, didn't find the way to do it in an accurate way

vlad20012:
Thanks!
bors r+

PR #2382, opened 2018-03-19
T: check channel of rustc while resolve tests

Undin:


src/main/kotlin/org/rust/cargo/toolchain/RustToolchain.kt
```diff
@@ -27,9 +28,12 @@ data class RustToolchain(val location: Path) {
 
     fun queryVersions(): VersionInfo {
         checkIsBackgroundThread()
-        return VersionInfo(scrapeRustcVersion(pathToExecutable(RUSTC)))
+        return queryVersionsSync()
     }
 
+    @VisibleForTesting
```
matklad: We usually use `@TestOnly` I think. Have no idea what's the actual difference between the two =)

Undin: `@TestOnly` - you can use method/constructor only in tests sources
`@VisibleForTesting` -  you can use method/constructor as a protected method and in tests
So `@VisibleForTesting` is more suitable for us
Unfortunately, the corresponding inspection works only with java code so now there isn't a real difference between them in kotlin project

Undin:
bors r=matklad

PR #2396, opened 2018-03-25
RES & TY: Prevent infinite loops with cyclic type aliases

Undin:
Collects already visited type alias elements to prevent infinite loops while resolving fields in struct literals and struct literal type inference


kumbayo:
@Undin Looks fine to me. Thanks for fixing the bug i introduced!
One thing i noticed is that it seems to break highlighting of tests in `RsResolveTest.kt` and `RsExpressionTypeInferenceTest.kt`.
Not sure if we should do something or just accept it until the next plugin release ;-)

Undin:
> One thing i noticed is that it seems to break highlighting of tests in RsResolveTest.kt and RsExpressionTypeInferenceTest.kt.
Not sure if we should do something or just accept it until the next plugin release ;-)

I can't come up with any solution to this problem, but I don't think it's a real issue - just a minor inconvenience. And it will be fixed by new plugin version
So let's merge it!

bors r=kumbayo

PR #2398, opened 2018-03-25
INT: Fix make mutable intention

rrevenantt:
When variable is assigned with "ref" modifier like `let ref a = ... `, and you apply make mutable quick fix, it will result in wrong `let mut a = ... ` instead of `let ref mut a = ... `.

This PR fixes this issue.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
@@ -43,6 +43,7 @@ fun updateMutable(project: Project, binding: RsNamedElement, mutable: Boolean = 
         is RsPatBinding -> {
             val tuple = binding.ancestorStrict<RsPatTup>()
             val parameter = binding.ancestorStrict<RsValueParameter>()
+
             if (tuple != null && parameter != null) {
                 return
             }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
```diff
@@ -54,7 +55,8 @@ fun updateMutable(project: Project, binding: RsNamedElement, mutable: Boolean =
                 parameter.replace(newParameterExpr)
                 return
             }
-            val newPatBinding = RsPsiFactory(project).createPatBinding(binding.identifier.text, mutable)
+            val isRefBinding = (binding.bindingMode?.ref) != null
```
Undin: You can use `RsPatBinding.isRef` extension property here

src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
```diff
@@ -54,7 +55,8 @@ fun updateMutable(project: Project, binding: RsNamedElement, mutable: Boolean =
...
+            val newPatBinding = RsPsiFactory(project).createPatBinding(binding.identifier.text, mutable, isRefBinding)
```
Undin: Let's use named parameters for boolean args here to be sure that we don't mix up them

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -271,8 +271,8 @@ class RsPsiFactory(private val project: Project) {
             ?: error("Failed to create parameter element")
     }
 
-    fun createPatBinding(name: String, mutable: Boolean = false): RsPatBinding =
-        (createStatement("let ${if (mutable) "mut " else ""}$name = 10;") as RsLetDecl).pat
+    fun createPatBinding(name: String, mutable: Boolean = false, ref: Boolean = false): RsPatBinding =
+        (createStatement("let ${if (ref) "ref " else ""}${if (mutable) "mut " else ""}$name = 10;") as RsLetDecl).pat
```
Undin: Maybe extract `if` expressions into separate variables to make the code more readable?

Undin:
bors r+
Thanks!

PR #2403, opened 2018-03-25
ANN: add type mismatch quick fix using TryFrom

oleg-semenov:
`impl TryFrom<A> for B` exists where A is actual type and B is expected
type, which fulfills part of the forth bullet-point of the issue:

The Fix skips suggesting the "TryFrom" when it's implemented through
`impl<T, U> TryFrom<U> for T where T: From<U> { type Error = Infallible; ...}`, 
since the regular "From" should is already suggested in those cases.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/ConvertToTyUsingTryFromTraitFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/ConvertToTyUsingTryFromTraitFix.kt
@@ -1,0 +1,103 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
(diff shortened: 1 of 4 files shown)


vlad20012:
@oleg-semenov, Hmm, please, fix the commit message :)

oleg-semenov:
Updated the message.

vlad20012:
> Updated the message.

Umm, but nothing changed. I'm about `TrtFrom` :)

oleg-semenov:
Ah right. (didn't push actual commit). I updated it now

vlad20012:
Hmm, `convert::Infallible` is now removed from stdlib and replaced with `!` https://github.com/rust-lang/rust/pull/49038
Ok, I'll merge it and then I'll try to fix it somehow. Thanks!
bors r+

oleg-semenov:
oh interesting, "good" timing, i'll try to keep my rust-nightly more up-to-date next time, thanks

PR #2407, opened 2018-03-26
CARGO: Favor selected project when running cargo command

rrevenantt:
When there are multiple cargo projects, use project that was selected in tool window to run cargo command. Currently first project is always used.

Also I've added project name to generated run configuration to have separate run configurations for each project, otherwise it would be hard to keep in mind what project was run last time.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
@@ -24,7 +24,7 @@ fun CargoCommandLine.mergeWithDefault(default: CargoCommandConfiguration): Cargo
         this
 
 fun RunManager.createCargoCommandRunConfiguration(cargoCommandLine: CargoCommandLine): RunnerAndConfigurationSettings {
-    val runnerAndConfigurationSettings = createRunConfiguration(cargoCommandLine.command, CargoCommandConfigurationType().factory)
+    val runnerAndConfigurationSettings = createRunConfiguration(cargoCommandLine.name, CargoCommandConfigurationType().factory)
     val configuration = runnerAndConfigurationSettings.configuration as CargoCommandConfiguration
     configuration.setFromCmd(cargoCommandLine)
     return runnerAndConfigurationSettings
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/cargo/runconfig/command/RunCargoCommandActionBase.kt
```diff
@@ -35,10 +37,19 @@ abstract class RunCargoCommandActionBase(icon: Icon) : AnAction(icon) {
             ?.let { cargoProjects.findProjectForFile(it) }
             ?.let { return it }
 
-        return cargoProjects.allProjects.firstOrNull()
+        val cargoPanel = ToolWindowManager.getInstance(e.project!!)
```
Undin: IntelliJ platform has `DataContext` interface to pass some data into actions.
Usually you just pass some key into `AnActionEvent.getData()` and get the corresponding data.
To provide data you need to implement `DataProvider` interface.

In our case, `CargoToolWindowPanel` already implements `DataProvider` as an inheritor of `SimpleToolWindowPanel`. Moreover, it provides a selected project for `DetachCargoProjectAction` via `DetachCargoProjectAction.CARGO_PROJECT_TO_DETACH` key.
Let's generalize this api to use it in this case too.

I think we should to the following:
* move out `DetachCargoProjectAction.CARGO_PROJECT_TO_DETACH` key from `DetachCargoProjectAction`, for example, into `CargoToolWindow.kt` file
* rename it to `SELECTED_CARGO_PROJECT` and use `"Cargo.selectedProject"` as key name (or use other informative names)
* use it in `RunCargoCommandActionBase` to get selected cargo project

src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
```diff
@@ -24,7 +24,7 @@ fun CargoCommandLine.mergeWithDefault(default: CargoCommandConfiguration): Cargo
         this
 
 fun RunManager.createCargoCommandRunConfiguration(cargoCommandLine: CargoCommandLine): RunnerAndConfigurationSettings {
-    val runnerAndConfigurationSettings = createRunConfiguration(cargoCommandLine.command, CargoCommandConfigurationType().factory)
+    val runnerAndConfigurationSettings = createRunConfiguration(cargoCommandLine.name, CargoCommandConfigurationType().factory)
```
Undin: Maybe add `name` parameter into `createCargoCommandRunConfiguration` to avoid mutable state in `CargoCommandLine`?
We can make it nullable with `null` as default value and pass name only when it should differ from command name. In other cases use `cargoCommandLine.command`

Undin:
Thanks!
bors r+

PR #2418, opened 2018-03-29
Completion for dev-dependencies in TOML

matklad:
Have no idea if it actually works :( WiFi here is garbage :(


vlad20012:
Thanks!
bors r+

PR #2427, opened 2018-04-02
ANN: add type mismatch quick fix using FromStr

oleg-semenov:
`impl FromStr for A` exists where A is expected type and
&str/&mut str/&String/&mut String is actual type, which fulfills part of
the forth bullet-point of the issue:

Also reorganizing conditions for the fixes little bit.


vlad20012:
Hmm, Is it true that if we found type mismatch, we can always replace `Aa::from_str("")` to `"".parse()`? I think yes, and think we should use `.parse()` instead

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -124,6 +130,14 @@ sealed class RsDiagnostic(
             return result.ok()?.value
         }
 
+        private fun ifActualIsStrGetErrTyOfFromStrImplForTy(stringTy: Ty, ty: Ty, items: StdKnownItems, lookup: ImplLookup): Ty? {
+            if (actualTy !is TyReference || actualTy.referenced != TyStr && actualTy.referenced != stringTy) return null
+            val fromStr = items.findFromStrTrait() ?: return null
+            val result = lookup.selectProjectionStrict(TraitRef(ty, BoundElement(fromStr)),
+                fromStr.associatedTypesTransitively.find { it.name == "Err"} ?: return null)
```
vlad20012: You can now use `fromStr.findAssociatedType("Err")` instead

oleg-semenov: done

oleg-semenov:
Cool. Switched to `.parse()`.

vlad20012:
Thank!
bors r+

PR #2431, opened 2018-04-02
TY: Fixed type inference for `Self {}` struct literal

### LINKED ISSUE #2361: Unresolved reference `None`
In this code:

```rust
struct Node<'a> {
    depth: u32,
    zero: Option<&'a Node<'a>>,
    one: Option<&'a Node<'a>>,
    value: Option<u32>,
}

impl<'a> Node<'a> {
    pub fn new() -> Self {
        Self {
            depth: 0,
            zero: None,
            one: None,
            value: None,
        }
    }
}

fn main() {
    let root = Node::new();
    println!("{}", root.value.is_some());
}
```

None in `new` function are marked as Unresolved, but `cargo check` show just warnings for never used fields.

unresolver reference

vlad20012:
Fixes #2361

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
@@ -9,13 +9,11 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.ResolveResult
 import com.intellij.psi.impl.source.resolve.ResolveCache
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsElement
-import org.rust.lang.core.psi.ext.RsGenericDeclaration
-import org.rust.lang.core.psi.ext.RsNamedElement
-import org.rust.lang.core.psi.ext.typeParameters
+import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.*
 import org.rust.lang.core.types.BoundElement
 import org.rust.lang.core.types.infer.foldTyInferWith
+import org.rust.lang.core.types.infer.substitute
 import org.rust.lang.core.types.inference
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -128,3 +126,40 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
 private fun resolveAssocTypeBinding(trait: RsTraitItem, binding: RsAssocTypeBinding): RsTypeAlias? =
     collectResolveVariants(binding.referenceName) { processAssocTypeVariants(trait, it) }
         .singleOrNull() as? RsTypeAlias?
+
+/** Resolves a reference through type aliases */
+fun RsPathReference.throughResolve(): RsElement? =
+    advancedThroughResolve()?.element
+
+/** Resolves a reference through type aliases */
+fun RsPathReference.advancedThroughResolve(): BoundElement<RsElement>? {
```
Undin: Maybe `deepResolve` is better?

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -128,3 +126,40 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
...
+
+    // Resolve potential type aliases
+    return if (boundElement != null && boundElement.element is RsTypeAlias) {
+        resolveThroughTupeAliases(boundElement)
+    } else {
+        boundElement
+    }
+}
+
+private fun resolveThroughTupeAliases(boundElement: BoundElement<RsElement>): BoundElement<RsElement>? {
```
Undin: `Tupe` -> `Type`

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -128,3 +126,40 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
...
+    var base: BoundElement<RsElement> = boundElement
+    @Suppress("NAME_SHADOWING")
```
vlad20012: ^

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
vlad20012: I hesitated a long time with the choice, but ok, will be `deepResolve` :)

vlad20012:
bors r=undin

PR #2441, opened 2018-04-05
ANN: Build/check all targets for both annotations and project build.

### LINKED ISSUE #1433: Errors in tests are not shown
The new `cargo check` feature does not show errors in tests (under the attributes `#[cfg(test)]`/`#[test]`).

### LINKED ISSUE #2178: Building/error checking doesn't compile/check examples and such
When developing, I want to be immediately sure that "examples" continue building. But currently the plugin simply ignores any and all code in "examples" subdirectory. No errors are highlighted there and `Ctrl-F9` doesn't build it.

I assume the same goes for "tests" and "benches", because having looked at the plugin source it seems to me that mere `cargo check --all` is called for building, whereas to also check examples and such it also has to have `--all-targets` switch (or maybe its more supported equivalent `--tests --examples --benches --bins --lib`).

I'm not sure why the build command is simply not fully customizable (provided it outputs in "--message-format=json"), so that the user could run whatever they want, be it `cargo check --all-targets`, `cargo check --all-features`, `cargo build` or `cargo test --no-run` or whatever their fancy is.

But at least provide the option to check everything in a crate, so as to have errors properly highlighted in the editor.

orium:
This uses `--all-targets` for error/warning annotations and to build/check the project.  

Fixes #1433 
Fixes #2178

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
@@ -5,17 +5,38 @@
 
 package org.rust.cargo.runconfig
 
+import com.intellij.execution.configurations.GeneralCommandLine
+import com.intellij.execution.ExecutionException
 import com.intellij.execution.ExecutorRegistry
+import com.intellij.execution.process.CapturingProcessHandler
 import com.intellij.execution.ProgramRunnerUtil
 import com.intellij.execution.RunManager
 import com.intellij.execution.RunnerAndConfigurationSettings
 import com.intellij.execution.executors.DefaultRunExecutor
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import org.rust.cargo.project.model.cargoProjects
 import org.rust.cargo.project.settings.rustSettings
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
 import org.rust.cargo.runconfig.command.CargoCommandConfigurationType
 import org.rust.cargo.toolchain.CargoCommandLine
+
+private val LOG_GENERAL_COMMAND_LINE = Logger.getInstance(GeneralCommandLine::class.java)
+
+fun GeneralCommandLine.runExecutable(): List<String>? {
+    val procOut = try {
+        val timeoutMs = 1000
+        CapturingProcessHandler(this).runProcess(timeoutMs)
+    } catch (e: ExecutionException) {
+        LOG_GENERAL_COMMAND_LINE.warn("Failed to run executable!", e)
+        return null
+    }
+
+    if (procOut.exitCode != 0 || procOut.isCancelled || procOut.isTimeout)
+        return null
+
+    return procOut.stdoutLines
+}
 
 fun CargoCommandLine.mergeWithDefault(default: CargoCommandConfiguration): CargoCommandLine =
     if (environmentVariables.envs.isEmpty())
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -45,6 +45,16 @@ class Cargo(
     private val cargoExecutable: Path,
     private val rustExecutable: Path
 ) {
+    val supportsBuildCheckAllTargets: Boolean = checkSupportForBuildCheckAllTargets()
```
Undin: Let's use `checkSupportForBuildCheckAllTargets` directly where we need this info to avoid unnecessary invocation of `checkSupportForBuildCheckAllTargets` for every `Cargo` object creation

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -45,6 +45,16 @@ class Cargo(
...
+
+    private fun checkSupportForBuildCheckAllTargets(): Boolean {
+        val lines = GeneralCommandLine(cargoExecutable)
```
Undin: I think we can use `CargoCommandLine` here
It will require some `Disposable` to execute but we have it in all places where we need info about `--all-targets` flag

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -97,9 +107,16 @@ class Cargo(
     }
 
     @Throws(ExecutionException::class)
-    fun checkProject(owner: Disposable, projectDirectory: Path): ProcessOutput =
-        CargoCommandLine("check", projectDirectory, listOf("--message-format=json", "--all"))
+    fun checkProject(owner: Disposable, projectDirectory: Path): ProcessOutput {
+        var arguments = listOf("--message-format=json", "--all")
```
Undin: Let's use `mutableListOf` instead of `listOf `. It allows us to make `arguments` final

src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
```diff
@@ -36,7 +36,16 @@ fun Project.buildProject() {
...
     for (cargoProject in cargoProjects.allProjects) {
-        val cmd = CargoCommandLine.forProject(cargoProject, command, listOf("--all"))
+        val allTargets = rustSettings.toolchain
+            ?.cargoOrWrapper(cargoProject.manifest.parent)
+            ?.supportsBuildCheckAllTargets ?: false
+        var arguments = listOf("--all")
+
+        if (allTargets) {
+            arguments += "--all-targets"
+        }
```
Undin: It seems that we can create argument list one time outside of the loop

orium: Can we safely assume that the version of cargo given by `toolchain.rawCargo()` is the same as `cargoOrWrapper(cargoProject.manifest.parent)` for all `cargoProject`s?

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
orium: Not only that, we would also have to pass the project directory around to create `CargoCommandLine`.  Is it worth it?

src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
Undin: Why not? We use same cargo binary for all projects
Or do you know a case when we can't assume it?

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
Undin: Hm, maybe not
Then let's move `GeneralCommandLine.runExecutable` into `Utils` and make it public

src/main/kotlin/org/rust/cargo/runconfig/Utils.kt
orium: > Or do you know a case when we can't assume it?

I do not.  I did it this way just in case xargo could wrap around different cargo versions, but it doesn't seem to do anything like it (from a quick google search).

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
orium: The `Utils` in `org.rust.cargo.runconfig`?

Undin: in existing `org.rust.cargo.runconfig.Utils`

orium:
Done

Undin:
Thanks!
bors r+

PR #2445, opened 2018-04-07
INSP: extern crate alias

### LINKED ISSUE #2444: Auto import doesn't take into account extern crate alias
Crate `X`:
```rust
pub mod foo {
    pub struct Foo;
}
```
Out project:
```rust
extern crate X as Y;

fn main() {
    let foo = Foo/*caret*/;
}
```
Try to import `Foo`

**Expected**:
```rust
extern crate X as Y;

use Y::foo::Foo;

fn main() {
    let foo = Foo;
}
```

**Actual**:
```rust
extern crate X as Y;

use X::foo::Foo;

fn main() {
    let foo = Foo;
}
```

Undin:
Take into account extern crate alias while auto import

Fixes #2444


vlad20012:
:+1: 
bors r+

PR #2446, opened 2018-04-08
ACT: "Implement members" should honor thе оrder of trait members in the definition

### LINKED ISSUE #2257: "Implement Member" doesn't take into account trait member items order
`Implement Member` quick fix creates default member implementation in the following order: assoc types, assoc consts and methods, - even original order differs.
In combination with RsSortImplTraitMembersInspection it leads to situation when our feature generates code with warning.

Need to generate member items in the same order as in trait.

Related code can be found [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt#L45-L60) and [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt#L141-L155).

JakubAdamWieczorek:
Fixes #2257.

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
@@ -9,13 +9,15 @@ import com.intellij.codeInsight.hint.HintManager
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.tree.IElementType
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.leftSiblings
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsNamedElement
-import org.rust.lang.core.psi.ext.TraitImplementationInfo
-import org.rust.lang.core.psi.ext.resolveToBoundTrait
-import org.rust.lang.core.psi.ext.resolveToTrait
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.rightSiblings
 import org.rust.lang.core.types.BoundElement
-import org.rust.lang.core.types.ty.Substitution
 import org.rust.openapiext.checkReadAccessAllowed
 import org.rust.openapiext.checkWriteAccessAllowed
 
@@ -32,7 +34,7 @@ fun generateTraitMembers(impl: RsImplItem, editor: Editor?) {
     if (chosen.isEmpty()) return
     runWriteAction {
         // Non-null was checked by `findMembersToImplement`.
-        insertNewTraitMembers(chosen, impl.members!!, trait.subst)
+        insertNewTraitMembers(chosen, impl.members!!, trait)
     }
 }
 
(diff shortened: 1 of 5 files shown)


Undin:
It seems that `createMembers` method should just create default implementation of members but the logic of their location in `impl` item should be in `insertNewTraitMembers`
It allows us to support right order not only in empty `impl` item but add new members in necessary order as well
```rust
trait Foo {
    type T;
    const C: i32;
}

struct S;

impl Foo for S {
    type T = ();
}
```
we insert default implementation of const at the beginning of impl

JakubAdamWieczorek:
@Undin, you are absolutely right. I will revise this change, though only at the end of the week. :)

JakubAdamWieczorek:
@Undin, what do you think would be the best behaviour for the action when there already exist some member implementations that are not in the right order? There are a few options to choose from.

- Ignore the desired order and add new members at the end.
- Try to insert new members in the right positions (though obviously the result might not be of value if the existing order is completely arbitrary).
- Allow the user to decide, should the action perhaps sort the existing members as well, so that the resulting trait implementation satisfies the constraint.

Undin:
I think we should try to use the second option and fallback to the first one if the current order doesn't allow us to insert new members in right positions

JakubAdamWieczorek:
@Undin, I revised the PR. :)

JakubAdamWieczorek:
@Undin, какие-нибудь мысли у вас есть? 🙂

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
```diff
@@ -45,20 +47,63 @@ private fun findMembersToImplement(impl: RsImplItem): Pair<TraitImplementationIn
...
+            existingMembers.childrenOfType<RsAbstractable>().findLast {
+                existingMember -> traitMembers.indexOfFirst { it.name == existingMember.name } < posInTrait
+            }
+        } ?: existingMembers.lbrace ?: return
+
+        // If the newly added item is a function, we add an extra line between it and each of its siblings.
+        val addedMember = existingMembers.addAfter(newMember, anchor)
+
+        val previousAbstractable = addedMember.leftSiblings.find { it is RsAbstractable }
+        if (previousAbstractable != null) {
```
Undin: check for null is redundant here

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
```diff
@@ -45,20 +47,63 @@ private fun findMembersToImplement(impl: RsImplItem): Pair<TraitImplementationIn
...
+
+        // If [1] does not hold, the first new member we will append at the end of the implementation.
+        // All the other ones will consequently be inserted at the right position in relation to that very first one.
+        val anchor = if (!areExistingMembersInTheRightOrder && index == 0) {
+            existingMembers.childrenOfType<RsAbstractable>().lastOrNull()
+        } else {
+            existingMembers.childrenOfType<RsAbstractable>().findLast {
+                existingMember -> traitMembers.indexOfFirst { it.name == existingMember.name } < posInTrait
+            }
+        } ?: existingMembers.lbrace ?: return
```
Undin: Hm, call `existingMembers.childrenOfType<RsAbstractable>()` for each iteration looks not very efficient
Looks like we can calculate necessary info only one time before the loop

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
JakubAdamWieczorek: `previousAbstractable` will be null when inserting a stub implementation of a member at the very first position in the `impl`. In fact, when I remove this check, I get 10 test failures.

JakubAdamWieczorek:
@Undin, thanks for taking a look. I addressed all your comments!

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
```diff
@@ -45,20 +47,63 @@ private fun findMembersToImplement(impl: RsImplItem): Pair<TraitImplementationIn
...
-    )
+    val templateImpl = RsPsiFactory(existingMembers.project).createMembers(selected, trait.subst)
+
+    val traitMembers = trait.element.members!!.childrenOfType<RsAbstractable>()
+    val newMembers = templateImpl.childrenOfType<RsAbstractable>()
+
+    // [1] First, check if the order of the existingMembers already implemented
+    // matches the order of existingMembers in the trait declaration.
+    val existingMembersOrder = existingMembers.childrenOfType<RsAbstractable>().map {
+        existingMember -> traitMembers.indexOfFirst { it.name == existingMember.name }
```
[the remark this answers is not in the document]
JakubAdamWieczorek: A good catch, thanks!

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
JakubAdamWieczorek: Well, I'd previously thought about it as well and decided on an extra list to bookkeep these positions but later figured it'd add complexity for close to no benefit in return as we still have to do O(n) lookups, just that instead of traversing the element tree we traverse the list in question. But now I did this change anyway and I suppose that with the other fix that addresses the name collission issue this change becomes much more desirable as it helps reduce code redundance.

Undin:
bors r+
Thanks!

Issue #2393, opened 2018-03-23
E0384 false positive when variables of type &str are declared first and initialized later.


Gowee:
When declaring an immutable variable of type `&str` without initializing at the same time and initialize it later instead, it prompts with  *Re-assignment of immutable variable [E0384]*.

*e.g.*
```
fn main() {
    let _blah: &str;

    _blah = "blah"; // E0384

    println!("{}", _blah);
}
```

plugin version: 0.2.0.2090-173
IDE: PyCharm Pro 2017.3.3
rustup toolchain: rustc 1.26.0-nightly (75af15ee6 2018-03-20)

Undin:
Fixed in #2447

PR #2447, opened 2018-04-08
INSP: Fix false warning of reassignment to immutable type for referen…

sgift:
First check if there was already an assignment and then check for reference types.

Fix for #2393


Undin:
Looks good!
But I've found that `RsExpr.isMutable` is rather hacky and should be rewritten sometime

Undin:
And as a new intellij-rust contributor please add your github nickname into CONTRIBUTORS.txt :)

Undin:
Welcome to intellij-rust contributors!

PR #2468, opened 2018-04-14
TY: Fixed `write!` & `writeln!` macro return type

### LINKED ISSUE #2229: Wrong return type inferred for write!() and writeln!()
Currently both `write!()` and `writeln!()` macros are inferred to return `()`.
In most cases the macros should actually return a different type.

The implementation of the `write!()` and `writeln!()` macros calls write_fmt() on their first argument and return the result of that call.

When called on a expression implementing the `io::Write` trait it should return `io::Result<()>`
When called on a expression implementing the `fmt::Write` trait it should return `fmt::Result<()>`

This was reported at https://users.rust-lang.org/t/what-is-the-expect-function-for-writeln-macro/15200/

### LINKED ISSUE #2452: Incorrect "mismatched types [E0308]" on `write!`
```rust
struct Point {
    x: i32,
    y: i32,
}

impl fmt::Display for Point {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "({}, {})", self.x, self.y) //Intellij marks this line 
    }
}
```
Intellij marks this code in red and displays "mismatched types [E0308] expected Result<(),Error>, found()". But the code is the code is compiled without problem.
The parser seems to incorrectly infer the return value.

### LINKED ISSUE #2461: Wrong mismatched type error for write! in match
Put the following code in a rust file: https://play.rust-lang.org/?gist=6e240f1d2307a736597f7862ed1b49c3&version=stable

This compiles and runs fine, but the plugin shows an error:

image

    mismatched types [E0308]
    expected Result<(), Error>, found `()`

vlad20012:
Really this commit as a first experiment with
expression-context macro expansion and type inference inside macros

Fixes #2229
Fixes #2452 
Fixes #2461

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -856,7 +856,8 @@ Expr ::= RetExpr
        | UnaryExpr
        | TryExpr
        | AtomExpr {
-  extends = "org.rust.lang.core.psi.ext.RsStubbedElementImpl<?>"
+  implements = "org.rust.lang.core.macros.ExpansionResult"
+  mixin = "org.rust.lang.core.psi.ext.RsExprMixin"
   stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
 }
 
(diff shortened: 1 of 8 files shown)


Undin:
Hm
bors r=vlad20012

PR #2481, opened 2018-04-18
FMT: Do not put a space between '?' and 'Sized'

### LINKED ISSUE #2377: Formatting adds space between ? and Sized
Formatting the following:

```rust
    pub fn add<T: Serialize + ?Sized>(&mut self, key: &str, val: &T) {
        self.data.insert(key.to_owned(), to_value(val).unwrap());
    }
```

gives:

```rust
    pub fn add<T: Serialize + ? Sized>(&mut self, key: &str, val: &T) {
        self.data.insert(key.to_owned(), to_value(val).unwrap());
    }
```

kumbayo:
We currently put a space between the '?' and 'Sized' so it looks like "T: ? Sized".

rustfmt does not add a space, so lets match its behavior.

Fixes #2377


vlad20012:
Thanks!
bors r+

PR #2484, opened 2018-04-19
INSP: Fix "Apply same member order" for incomplete `impls`

### LINKED ISSUE #2476: "Apply same member order" generates wrong code
**Steps to reproduce:**
Apply `Apply same member order` quick fix for the following code
```rust
trait Foo {
    type T;
    const C: i32;
    fn foo();
}

struct S;

impl Foo for S/*caret*/ {
    fn foo() { unimplemented!() }
    const C: i32 = unimplemented!();
}
```

**Expected:**
```rust
impl Foo for S {
    const C: i32 = unimplemented!();
    fn foo() { unimplemented!() }
}
```

**Actual:**
```rust
impl Foo for S {
    const C: i32 = unimplemented!();
    const C: i32 = unimplemented!();
}
```

See code of the corresponding inspection and quick fix [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt)

JakubAdamWieczorek:
`SortImplTraitMembersFix` would not work properly if the `impl`
was missing implementations of one or more members of the trait.

Fixes #2476.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
@@ -15,6 +15,7 @@ import org.rust.lang.core.psi.ext.RsItemElement
 import org.rust.lang.core.psi.ext.RsTraitOrImpl
 import org.rust.lang.core.psi.ext.elementType
 import org.rust.lang.core.psi.ext.resolveToTrait
+import org.rust.openapiext.Testmark
 
 class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
@@ -33,9 +34,13 @@ class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
 
     companion object {
         private fun sortedImplItems(implItems: List<RsItemElement>, traitItems: List<RsItemElement>): List<RsItemElement>? {
-            val implItemMap = implItems.associate { it.key() to it }
-            val sortedImplItems = traitItems.mapNotNull { implItemMap[it.key()] }
-            if (sortedImplItems.size != implItems.size || sortedImplItems == implItems) return null
+            val traitItemMap = traitItems.withIndex().associate { it.value.key() to it.index }
+            if (implItems.any { it.key() !in traitItemMap }) {
+                Testmarks.implMemberNotInTrait.hit()
+                return null
+            }
+            val sortedImplItems = implItems.sortedBy { traitItemMap[it.key()] }
+            if (sortedImplItems == implItems) return null
             return sortedImplItems
         }
     }
@@ -48,11 +53,18 @@ class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
             val trait = impl.traitRef?.resolveToTrait ?: return
             val implItems = impl.items()
             val traitItems = trait.items()
-            val sortedImplItems = sortedImplItems(implItems, traitItems)?.map { it.copy() } ?: return
-            traitItems.zip(implItems).forEachIndexed { index, (traitItem, implItem) ->
-                if (traitItem.key() != implItem.key()) implItem.replace(sortedImplItems[index])
+
+            // As we're applying the fix, this will be non-null.
+            val sortedImplItems = sortedImplItems(implItems, traitItems) ?: return
+            implItems.zip(sortedImplItems).forEachIndexed { index, (implItem, expectedImplItem) ->
+                if (implItem.key() != expectedImplItem.key())
+                    implItem.replace(sortedImplItems[index].copy())
             }
         }
+    }
+
+    object Testmarks {
+        val implMemberNotInTrait = Testmark("implMemberNotInTrait")
     }
 }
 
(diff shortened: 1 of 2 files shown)


JakubAdamWieczorek:
This isn't quite correct, I'll revise.

JakubAdamWieczorek:
Okay, I've fixed it and added one more test. This should be good to go. :)

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
```diff
@@ -33,9 +33,11 @@ class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
 
     companion object {
         private fun sortedImplItems(implItems: List<RsItemElement>, traitItems: List<RsItemElement>): List<RsItemElement>? {
-            val implItemMap = implItems.associate { it.key() to it }
-            val sortedImplItems = traitItems.mapNotNull { implItemMap[it.key()] }
-            if (sortedImplItems.size != implItems.size || sortedImplItems == implItems) return null
+            val traitItemMap = traitItems.withIndex().associate { it.value.key() to it.index }
+            val implItemsInTrait = implItems.filter { traitItemMap.contains(it.key()) }
+            if (implItemsInTrait.size < implItems.size) return null
```
Undin: Do I understand correctly that here you check that all items from `implItems` are in the corresponding trait item?
If it's true, looks like that you let's just check this case more explicitly (for instance, somehow like this`if (implItems.any { it.key() !in traitItemMap }) return null`)

Undin: Oh, let's create `Testmark` for this case
The goal of `Testmark` is to link rare cases with their tests and ensure that you don't drop support of such cases during refactorings
You can see example of testmark using [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt#L78)
Don't forget to use it in test - just pass testmark object in `check*` function call in the corresponding test

src/test/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspectionTest.kt
```diff
@@ -234,4 +245,32 @@ class RsSortImplTraitMembersInspectionTest : RsInspectionsTestBase(RsSortImplTra
...
+    fun `test does the right thing when the impl is missing some of the members`() = checkFixByText("Apply same member order", """
+        trait Foo {
+            type T;
+            const C: i32;
+            fn foo();
+        }
+
+        struct S;
+
+        <error descr="Not all trait items implemented, missing: `T` [E0046]">impl Foo for S/*caret*/</error> {
```
Undin: Let's check that we show description of `RsSortImplTraitMembersInspection` too. You need to add `<weak_warning>` tag (see other tests for this inspection) and enable weak warning highlighting (pass the corresponding param to `checkFixByText ` function)

Undin: Oh, I've just noticed that we don't enable weak warning checking in other tests so really we don't check inspection message at all.
Could you please fix it too?

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
```diff
@@ -33,9 +33,11 @@ class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
 
     companion object {
         private fun sortedImplItems(implItems: List<RsItemElement>, traitItems: List<RsItemElement>): List<RsItemElement>? {
-            val implItemMap = implItems.associate { it.key() to it }
-            val sortedImplItems = traitItems.mapNotNull { implItemMap[it.key()] }
-            if (sortedImplItems.size != implItems.size || sortedImplItems == implItems) return null
+            val traitItemMap = traitItems.withIndex().associate { it.value.key() to it.index }
+            val implItemsInTrait = implItems.filter { traitItemMap.contains(it.key()) }
```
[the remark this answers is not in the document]
JakubAdamWieczorek: Fixed.

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
```diff
@@ -48,9 +50,12 @@ class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
             val trait = impl.traitRef?.resolveToTrait ?: return
             val implItems = impl.items()
             val traitItems = trait.items()
-            val sortedImplItems = sortedImplItems(implItems, traitItems)?.map { it.copy() } ?: return
-            traitItems.zip(implItems).forEachIndexed { index, (traitItem, implItem) ->
-                if (traitItem.key() != implItem.key()) implItem.replace(sortedImplItems[index])
+
+            // As we're applying the fix, this will be non-null.
+            val sortedImplItems = sortedImplItems(implItems, traitItems)!!
```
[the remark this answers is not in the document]
JakubAdamWieczorek: Interesting! Fixed.

src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
JakubAdamWieczorek: Done! 🙂

src/test/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspectionTest.kt
JakubAdamWieczorek: Sure! Done. 🙂

Undin:
Great!
Thanks a lot!
bors r+

PR #2489, opened 2018-04-23
UI & CARGO: Improve workspace tree

### LINKED ISSUE #2151: Import all cargo workspace member crates when importing a project
When "Import Project" dialog invoked, intellij-rust creates a project with only with root cargo crate, but not with imported crates.

For example, with [futures-rs](https://github.com/alexcrichton/futures-rs/):

0. Clone https://github.com/alexcrichton/futures-rs/
1. Lauch "Import Project" dialog.
2. Select root folder of cloned futures-rs
3. Click next, next
4. Intellij-rust sees only root project:

Although `Cargo.toml` of futures-rs [references futures-cpupool](https://github.com/alexcrichton/futures-rs/blob/bf5dc304ccc37e05e3dbe57c17c70ef1ca175136/Cargo.toml#L28-L29):

```
[workspace]
members = ["futures-cpupool"]
```

I would be convenient if `future-cpupool` also added to the project. It's important for project with 10 crates or so, because otherwise these crates need to be added to project manually.

Versions:
Intellij IDEA CE: 2017.3.1
intellij-rust: 0.2.0.2082
cargo: 0.25.0-nightly (930f9d949 2017-12-05)

Undin:
* Show workspace members and targets in cargo toolwindow
* Allow running targets from cargo toolwindow by double click

TODO:
* Come up with icons for target nodes

Fixes #2151

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
@@ -28,13 +28,13 @@ import com.intellij.xdebugger.impl.XDebugProcessConfiguratorStarter
 import com.intellij.xdebugger.impl.ui.XDebugSessionData
 import org.jetbrains.concurrency.AsyncPromise
 import org.jetbrains.concurrency.Promise
+import org.rust.cargo.project.workspace.CargoWorkspace.CrateType
 import org.rust.cargo.project.workspace.CargoWorkspace.TargetKind
 import org.rust.cargo.runconfig.CargoRunState
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
 import org.rust.cargo.toolchain.Cargo
 import org.rust.cargo.toolchain.CargoCommandLine
 import org.rust.cargo.toolchain.impl.CargoMetadata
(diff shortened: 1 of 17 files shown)


matklad:
LGTM, but the absence of tests bothers me...

Do other intellij projects test similar stuff somehow?

Undin:
@matklad unfortunately, I don't know any API to create tool window tests in a headless environment
So I test tree structure directly

matklad:
>I don't know any API to create tool window tests in a headless environment

I don't know either! Might be a good idea to ask UI folks about how they approach testing! I expect that "manually" would be the answer :-) 

Historically, we basically ignored UI testing, because there wasn't a lot of UI to begin with. The Cargo toolbar I think is our the first major UI component, and we probably should do *something explicit* about testing it, even if that is "nah, it's to complicated to test automatically, and to burdensome to test manually, so we'll just monitor twitter for angry complaints" :) 

The simplest testing option, in theory, is just to create a short text document, "UI test plan", which describes the most complex and important UI interactions, and to click it through manually before important releases. I have never used this approach in practice, so it might not work at all.

Anyway, it's just some ramblings to thing about, the changes themselves look good to me!

PR #2494, opened 2018-04-25
DBG: load default pretty printers for rust

### LINKED ISSUE #2140: Make debugger to show string values
It's very important to see what strings values are. But today the only way to see them is just log it somewhere. It's not really convinient and sometimes require large code rewrite to satisfy all rust rules to just borrow some value and print it, especially for nested values.

In a nutshell: currently debugger shows strings as `Vec<u8>`, but we need more human-readable representation.

Undin:
Loads pretty printers into IDE debugger to show values of rust variables in more convenient for user way. `rust-gdb` and `rust-lldb` use the same pretty-printers.

Some samples:
**lldb (without pretty-printers)**
screen shot 2018-04-25 at 13 48 06

**gdb (with rust pretty-printers)**
deepinscreenshot_select-area_20180425181712
**lldb (with rust pretty-printers)**
screen shot 2018-04-25 at 13 45 34

It's still far from perfect when lldb is used because doesn't allow to see separate elements of `Vec` but it's better than now

Fixes #2140

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
@@ -1,0 +1,69 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger.runconfig
+
+import com.intellij.execution.filters.TextConsoleBuilder
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.util.io.FileUtil
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.xdebugger.XDebugSession
+import com.jetbrains.cidr.execution.RunParameters
+import com.jetbrains.cidr.execution.debugger.CidrLocalDebugProcess
+import com.jetbrains.cidr.execution.debugger.backend.DebuggerCommandException
+import com.jetbrains.cidr.execution.debugger.backend.gdb.GDBDriver
+import com.jetbrains.cidr.execution.debugger.backend.lldb.LLDBDriver
+
+class RsLocalDebugProcess(
+    parameters: RunParameters,
+    debugSession: XDebugSession,
+    consoleBuilder: TextConsoleBuilder
+) : CidrLocalDebugProcess(parameters, debugSession, consoleBuilder) {
+
+    fun loadPrettyPrinters(sysroot: String) {
+        val threadId = currentThreadId
+        val frameIndex = currentFrameIndex
+        postCommand { driver ->
+            when (driver) {
+                is LLDBDriver -> driver.loadPrettyPrinters(threadId, frameIndex, sysroot)
+                is GDBDriver -> driver.loadPrettyPrinters(threadId, frameIndex, sysroot)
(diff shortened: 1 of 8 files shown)


Undin:
@matklad @vlad20012 Could you check that it really works?

src/main/kotlin/org/rust/cargo/runconfig/CargoRunState.kt
```diff
@@ -49,6 +47,13 @@ class CargoRunState(
         }
     }
 
+    fun cargo(): Cargo = toolchain.cargoOrWrapper(cargoProject?.manifest?.parent)
+
+    fun sysroot(): String? {
```
matklad: Inconsistent naming: sysroot vs getSysroot. 

I would called this `computeSysroot`, because this does IO, and might fail with ExecutionException.

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    private fun LLDBDriver.loadPrettyPrinters(threadId: Long, frameIndex: Int, sysroot: String) {
+        val path = "$sysroot/lib/rustlib/etc/lldb_rust_formatters.py".systemDependentAndEscaped()
+        try {
+            executeConsoleCommand(threadId, frameIndex, """command script import "$path"""")
+            executeConsoleCommand(threadId, frameIndex, """type summary add --no-value --python-function lldb_rust_formatters.print_val -x ".*" --category Rust""")
+            executeConsoleCommand(threadId, frameIndex, """type category enable Rust""")
+        } catch (e: DebuggerCommandException) {
```
matklad: LGTM for now, but make sure that CLion folks are aware that we load pretty printers this way. They might add a proper API for this in the future hopefully!

matklad:
r=me with an opt-out checkbox and naming fixed!

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
Undin: Currently, there isn't special API for pretty printers loading and gdb pretty-printers are loaded in the same way

Undin:
bors r=matklad

ice1000:
Is it even possible to view an `enum` value? I'm debugging on Windows with msys' gdb toolchain, with the nightly plugin version I've just downloaded from the plugins.jetbrains.com page. Seems that I cannot see pretty-printed `BTreeMap` or an instance (`Value`) of an `enum`:

image

kevleyski:
I still only see vec bytes when debugging a String - am I doing something wrong?

Undin:
@kevleyski could you file an issue, please? `Help | Create New Issue` action may help to collect all basic info.
Also, it would be great to get detailed debugger logs (see https://intellij-support.jetbrains.com/hc/en-us/articles/360001436079-Collecting-additional-logs-in-CLion- instruction, the necessary setting is `#com.jetbrains.cidr.execution.debugger`).
Also, if you use CLion, it would be great to get info about your debugger. See `Preferences | Build, Execution, Deployment | Toolchains` settings

PR #2497, opened 2018-04-26
LEX: Simple support raw identifiers

### LINKED ISSUE #2456: Support raw identifiers (RFC 2151)
Allows identifiers to start with `r#`. 
* [x] Lexer
* [ ] Parser and PSI

See [RFC 2151](https://rust-lang.github.io/rfcs/2151-raw-identifiers.html)

Dimonchik0036:
Fixes #2456


vlad20012:
Hi!
I found that you went a little too far with the tests =)
1. It's better to use "resolve tests" instead of completion tests in this case. I.e. `RsResolveTest`
2. One test is enough to check identifier parsing,

Dimonchik0036:
Fixed.

vlad20012:
Thanks!
bors r+

Undin:
Hm, but the current implementation for `r#x` creates `r#x` identifier although it should create `x` 
Because of it resolve in real case doesn't work
```rust
fn main() {
    let x = 123;
      //X should resolve here
    println!("{}", r#x);
                  //^
}
```

vlad20012:
@Undin Yep, it's true, but the original issue was only about lexer-level support, i.e. now we correctly parse it at least

Undin:
Okay:)

PR #2507, opened 2018-04-30
TY: Do not cause IllegalStateException when typifying vec![0; 1]

### LINKED ISSUE #2500: vec![x;y] stops analysis
Code isn't being analysed when I write vec![x;y].

kumbayo:
When typifying `vec![0; 1]` we would trigger `error("Trying to infer expression type twice")`
in `RsExpr.inferType(expected: Ty?)`

This is because `inferChildExprsRecursively(expr.macroCall)`
already typified all children inside the vec[] and
`vecArg.exprList.firstOrNull()?.inferType()`
would typify the first child a second time.

Adapt a test from #2469 that was not actually testing vec![] repeat mode.
Fixes #2500


kumbayo:
@vlad20012 Can you have a look at this?

vlad20012:
Thanks!
(really nice to have a bug both in impl and in tests =/)
bors r+

PR #2511, opened 2018-05-02
GRAM: Raw identifiers

Dimonchik0036:
Adds a new rule for the identifier.


Undin:
Hm, it works in general, but I worry that `r#` is not part of identifier
You can see it, for example, while `goto definition` action (via `ctrl/cmd` + click) - `r#` is not highlighted with other part of identifier. Kotlin plugin with same feature (``` `identifier` ```) considers that all ``` `identifier` ``` text is identifier and highlights it in the same case.

Maybe we want to have the same behavior?
If yes, looks like it requires changes in the following places:
* `RsNamedElement#getName`
* `RsNamedElement#setName`
* `RsReferenceElement#referenceName`
* maybe somewhere else

@matklad @vlad20012 what do you think?

Dimonchik0036:
@Undin You're right, this behavior will be more logical. I'll change the solution.

PR #2523, opened 2018-05-07
RUN: Run single test in debug properly

### LINKED ISSUE #1843: Debugging just one test actually runs all tests
The UI implies that you can debug just a single test, but will actually run all tests. This is annoying if you want to use breakpoints, or if you have hundreds of tests.

Example code, with "run" buttons next to each test case, as well as the whole module:
image

Configuration for a generated run/debug profile from clicking on one of these. Note the "Command:" `test --lib tests::second`:
image

However, when the debugging session is started, it runs just the raw test binary (not the cargo command, as promised), without even `--nocapture`, and certainly without `second`, to limit it just to that test:
```
...testcase../target/debug/debug_one_test-90d09fece0178266

running 2 tests
test tests::first ... ok
test tests::second ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

..and the breakpoint will be hit for every test.

This is contrary to the `cargo` behaviour:
```
% cargo test second
...
running 1 test
test tests::second ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out
```

The code, in non-screenshot form:
```rust
fn foo(msg: &str) { println!("Hi, I'm running {}!", msg); }
#[cfg(test)] mod tests {
    #[test] fn first() { ::foo("first"); }
    #[test] fn second() { ::foo("second"); }
}
```

### LINKED ISSUE #2373: Start single test in debugger runs all test in module
I have tried to find if this issue already was reported but could not find it. If it was already a known issue, just let me know.

When I run a single test in CLion with the regular executor it will run a single test and all other tests in the module are marked as filtered out:
`test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 9 filtered out`

But if I run this same test in debugmode, it will run all tests in the module:
`test result: FAILED. 8 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out`.
I expected it to behave the same as the regular run.

Side-note: It even appears to run everything in parallel.

Undin:
Fixes #1843
Fixes #2373

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
@@ -38,6 +38,7 @@ import org.rust.cargo.runconfig.CargoRunState
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
 import org.rust.cargo.toolchain.CargoCommandLine
 import org.rust.cargo.toolchain.impl.CargoMetadata
+import org.rust.cargo.util.CargoArgsParser
 import org.rust.debugger.settings.RsDebuggerSettings
 import org.rust.openapiext.GeneralCommandLine
 import org.rust.openapiext.withWorkDirectory
@@ -64,20 +65,18 @@ class RsDebugRunner : AsyncProgramRunner<RunnerSettings>() {
     override fun execute(environment: ExecutionEnvironment, state: RunProfileState): Promise<RunContentDescriptor?> {
         state as CargoRunState
         val cmd = state.commandLine
-
-        val (buildArgs, execArgs) = cmd.splitOnDoubleDash()
+        val (commandArguments, executableArguments) = CargoArgsParser.parseArgs(cmd.command, cmd.additionalArguments)
 
         val buildCommand = if (cmd.command == "run") {
-            cmd.copy(command = "build", additionalArguments = buildArgs)
+            cmd.copy(command = "build", additionalArguments = commandArguments)
         } else {
-            check(cmd.command == "test")
             cmd.prependArgument("--no-run")
         }
 
         val runCommand = { executablePath: Path ->
             GeneralCommandLine(executablePath)
                 .withWorkDirectory(cmd.workingDirectory)
-                .withParameters(execArgs)
+                .withParameters(executableArguments)
                 .withEnvironment(cmd.environmentVariables.envs)
         }
 
--- a/src/main/kotlin/org/rust/cargo/toolchain/CargoCommandLine.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/CargoCommandLine.kt
@@ -52,11 +52,8 @@ data class CargoCommandLine(
      * Splits [additionalArguments] into parts before and after `--`.
      * For `cargo run --release -- foo bar`, returns (["--release"], ["foo", "bar"])
      */
-    fun splitOnDoubleDash(): Pair<List<String>, List<String>> {
-        val idx = additionalArguments.indexOf("--")
-        if (idx == -1) return additionalArguments to emptyList()
-        return additionalArguments.take(idx) to additionalArguments.drop(idx + 1)
-    }
+    fun splitOnDoubleDash(): Pair<List<String>, List<String>> =
+        org.rust.cargo.util.splitOnDoubleDash(additionalArguments)
 
     companion object {
         fun forTargets(
--- a/src/main/kotlin/org/rust/cargo/util/CargoArgsParser.kt
+++ b/src/main/kotlin/org/rust/cargo/util/CargoArgsParser.kt
@@ -1,0 +1,131 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.util
+
+class CargoArgsParser private constructor(
+    private val cargoArgs: List<String>,
+    private val optionsToArgsCountRange: Map<String, OptionArgsCountRange>
+) {
+    private fun splitArgs(): SplitCargoArgs {
+        val commandOptions = mutableListOf<String>()
+        val positionalArguments = mutableListOf<String>()
+
+        var i = 0
+        optionLoop@ while (i < cargoArgs.size) {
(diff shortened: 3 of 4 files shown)


src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducer.kt
```diff
@@ -96,10 +96,15 @@ class TestConfig(
...
     target: CargoWorkspace.Target,
-    private val exact: Boolean
+    exact: Boolean
 ) {
-    val cargoCommandLine: CargoCommandLine = CargoCommandLine.forTarget(target, "test", listOf(testPath))
-        .let { if (exact) it.withDoubleDashFlag("--exact") else it }
+    val cargoCommandLine: CargoCommandLine = if (exact) {
+        CargoCommandLine.forTarget(target, "test")
+            .withDoubleDashFlag("--exact")
+            .withDoubleDashFlag(testPath)
```
Undin: I'm not sure if it's supposed to work but it works in practice

matklad: Hm, I think the bug is fixed in the wrong place in this instance. When running `cargo test foo -- bar`, Cargo runs the test binary as `./target/debug/test-hash foo bar`. That is, the last positional argument and all args after `--`  are passed to the binary. 

I think the logic here needs to be adjusted: https://github.com/intellij-rust/intellij-rust/blob/2bbb484903b726e74e30e5a931bd3f2ba596f995/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt#L63. It should not be `splitOnDoubleDash`, but `getTestBinaryArguments`, and should basically replicate Cargo's argument parsing.

elichai:
Any news about it? 
I can run a single test but I can't seem to debug a single test, tried tweaking the configuration however I could without luck.

mchernyavsky:
@elichai Yes, I have updated this PR a few days ago, so everything should work after this PR would be reviewed and merged.

Undin:
@mchernyavsky I'll review it asap

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
```diff
@@ -0,0 +1,123 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.util
+
+class CargoRunArgsSeparator(cargoArgs: List<String>) : AbstractCargoArgsSeparator(cargoArgs, RUN_OPTIONS) {
+    override val commandOptions: List<String> by lazy {
```
Undin: Looks like `lazy` is redundant here

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
```diff
@@ -0,0 +1,123 @@
...
+        separatedArgs.first
+    }
+
+    override val executableArguments: List<String> by lazy {
+        val positionalArguments = separatedArgs.second
```
Undin: Hm, is it right name? Looks like it should be `executableArguments`

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
```diff
@@ -0,0 +1,123 @@
...
+        if (positionalArguments.isNotEmpty() && positionalArguments.first() == "--") {
+            positionalArguments.drop(1)
+        } else {
+            positionalArguments
+        }
```
Undin: Why can we drop `--` while calculation of `separatedArgs`?

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+class CargoTestArgsSeparator(cargoArgs: List<String>) : AbstractCargoArgsSeparator(cargoArgs, TEST_OPTIONS) {
+    override val commandOptions: List<String> by lazy {
+        val (commandOptions, positionalArguments) = separatedArgs
+        val (pre, _) = splitOnDoubleDash(positionalArguments)
+        commandOptions + pre.dropLast(1)
```
Undin: Сan we add positional arg to command and executable options while `separatedArgs` calculation to avoid this postprocessing?

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
mchernyavsky: This is the correct name, but I chose the wrong name for `executableArguments` in the `separatedArgs` property.

Now I use the following names:
* `commandOptions` -- options along with their arguments
* `positionalArguments` -- all positional arguments
* `commandArguments` -- `commandOptions` + part of `positionalArguments`
* `executableArguments` -- rest of `positionalArguments`

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
mchernyavsky: Cargo does it too.

```rust
fn main() {
    use std::env;
    let args: Vec<String> = env::args().collect();
    println!("{:?}", args);
}
```

```
cargo run a -- -- b
["target/debug/rust-sandbox", "a", "--", "--", "b"]
```

```
cargo run -- -- b
["target/debug/rust-sandbox", "--", "b"]
```

src/main/kotlin/org/rust/cargo/util/CargoArgsSeparator.kt
mchernyavsky: Oh, actually, we must do it. `cargo test` can't accept more than one positional argument. So there is an error in this implementation: If in the original command we passed exactly two positional arguments, Cargo will not report an error.
I have added a comment about this.

mchernyavsky:
bors retry

Undin:
@mchernyavsky Looks like it won't help
There is an exception while test launch. See logs on appveyor for more details

mchernyavsky:
bors try

Issue #2472, opened 2018-04-14
Mismatched type [E0308] offers Intention for fix but does not actual apply fix


schoenja:
I am getting an `mismatched types [E0308] expected '&mut String', found '&String'` error with this code on line 4 (code is from [this tutorial](https://doc.rust-lang.org/book/second-edition/ch04-02-references-and-borrowing.html#mutable-references))
```rust
fn main() {
    let mut s = String::from("hello");

    change(&s);
}

fn change(some_string: &mut String) {
    some_string.push_str(", world");
}
```
The intention offers to `Convert to &mut String using 'BorrowMut' trait`.

**Expected behaviour**:
The line is changed to `change(&mut s);`.

**Actual behaviour**:
The line is not changed.

Additional information:
IntelliJ IDEA 2018.1.1 (Ultimate Edition)
Build #IU-181.4445.78, built on April 9, 2018
JRE: 1.8.0_162-b12 amd64
JVM: OpenJDK 64-Bit Server VM by Oracle Corporation
Linux 4.15.15-1-ARCH
Rust plugin: Version: 0.2.0.2092-181

vlad20012:
c.c. #1730
c.c. @oleg-semenov

oleg-semenov:
hm it works little bit differently for me, it replaces `&s` in the line 4 with `(&s).borrow_mut()` which is still incorrect; i didn't try to match any versions though; I can make a bug-fix that adds a check that the expression is mutable before suggesting applying `BorrowMut` trait. So `BorrowMut` won't be suggested in this case.

As for the desired `&s` => `&mut s` fix we just currently don't have a quick-fix/intention for that. So far all quick-fixes/intentions for types mismatches don't try to replace or delete anything from the affected source-code; they add stuff before or around it. I don't know if we want to keep this "policy" (or it's just a coincidence). But i think in this case we can add a simple quick-fix/intention that in cases like this can try to check the mutability of `s` and suggest replacing `&` with `&mut`. So, more precisely it would be like if expected type it &mut T actual type is `& T` and the expression is mutable suggest replacing `&` with `&mut`.

@vlad20012 what do you think?

PR #2526, opened 2018-05-12
ANN: Only allow BororwMut and AsMut quick-fixes on mutable expressions

oleg-semenov:
This commit makes the quick-fixes that use BororwMut and AsMut traits
to be applicable only for truly mutable expressions. This fixes part of
the: https://github.com/intellij-rust/intellij-rust/issues/2472


vlad20012:
bors r+
Thanks!

PR #2529, opened 2018-05-13
MACRO: Surround expressions with parenthesis during macro expansion

kumbayo:
We currently expand macros textually instead of based on AST nodes
as rustc does.
This can cause problems where the macro expansion interacts badly
with operator precedence rules and produces wrong results.

For example this code (from the documentation of write!())
```rust
let mut s = String::new();
let a = write!(&mut s, "text");
```

would be wrongly expanded to
```rust
let a = &mut (s.write_fmt(...))
```
and typified as `&mut Result<(), Error>`

instead of
```rust
let a = (&mut s).write_fmt(...)
```
and being typified as `Result<(), Error>`.

Work around this problem by surrounding all expressions with parenthesis
during expansion.
CC #2229

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -239,7 +239,13 @@ private class MacroPattern private constructor(
                         MacroExpansionMarks.failMatchPatternByBindingType.hit()
                         return null
                     }
-                    map[name] = macroCallBody.originalText.substring(lastOffset, macroCallBody.currentOffset)
+                    val text = macroCallBody.originalText.substring(lastOffset, macroCallBody.currentOffset)
+
+                    // Wrap expressions in () to avoid problems related to operator precedence during expansion
+                    if (type == "expr")
+                        map[name] = "($text)"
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -239,7 +239,13 @@ private class MacroPattern private constructor(
                         MacroExpansionMarks.failMatchPatternByBindingType.hit()
                         return null
                     }
-                    map[name] = macroCallBody.originalText.substring(lastOffset, macroCallBody.currentOffset)
+                    val text = macroCallBody.originalText.substring(lastOffset, macroCallBody.currentOffset)
+
+                    // Wrap expressions in () to avoid problems related to operator precedence during expansion
```
kumbayo: I chose to immediately add the () here instead of storing type and the unaltered text and adding them during the actual expansion. This way it requires a lot less changes to the code.

kumbayo:
@vlad20012 Could you have a look at this change? Thanks

vlad20012:
:+1: Thanks!
bors r+

Issue #2544, opened 2018-05-21
Improve "Copy Reference" action for Rust


Undin:
Currently, `Copy Reference`  returns `$absolute-file-name:$line-number` for rust items
It's default platform implemenetation and it's not quite useful. For example, for `getQualifiedName` method in java `QualifiedNameProvider` interface it returns `com.intellij.ide.actions.QualifiedNameProvider#getQualifiedName`.

It would be great to implement similar bahavior for rust as well.
Looks like, you just need to implement [QualifiedNameProvider](/https://github.com/JetBrains/intellij-community/tree/master/platform/lang-impl/src/com/intellij/ide/actions/QualifiedNameProvider.java) for rust language. Note, we already have [RsQualifiedNamedElement](/https://github.com/intellij-rust/intellij-rust/tree/master/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt) class.

PR #2545, opened 2018-05-21
ACT: Add impl for QualifiedNameProvider

dimonchik0036:
Impl #2544


Dimonchik0036:
Now if we have this train
```
trait Show {
    fn show(&self) -> String;
}
```
it  look like as `package::show`

Dimonchik0036:
```rust
mod world {
    trait House {
        fn say_hello(&self) -> String;
    }
}
```
Look like as `package::world::House#say_hello`

Dimonchik0036:
@Undin what do you think?

mehcode:
Why `#` ? Rust uses `::` everywhere normally.

Dimonchik0036:
@mehcode `#` is used to show that the function belongs to the trait. Perhaps it really is not necessary.

mehcode:
Universal function call syntax in rust still uses `::` for functions in traits. 

E.g.. 

```rust
package::world::House::say_hello(foo)
```

Given the issue (copying a qualified reference for intended usage in code) I would assume the intention is to make UFCS easier to write if necessary.

Dimonchik0036:
@mehcode You are right, I will correct it.

Kobzol:
@dimonchik0036 Are you planning to finish this PR? :) Otherwise I can try to finish it.

dimonchik0036:
@Kobzol Too much time has passed that I even forgot. 😄 
You can finish this PR.

PR #2555, opened 2018-05-28
QDOC: Provide links to external documentation

### LINKED ISSUE #1875: Quick documentaion should contain link to the on-line docs
PyCharm has a useful feature where it adds an "View external documentaion" link to the quick-doc popup (<kbd>Ctrl+Q</kbd>):

image

I think we should do the same for Rust, using `docs.rs` and `https://doc.rust-lang.org/std/`. 

The entry point for the relevant feature is `org.rust.ide.docs.RsDocumentationProvider`.

Undin:
It's initial implementation - it provides links only for stdlib items and it doesn't support reexports (as well as all our quick documentation does).
It means:
* we generate link according to a fully qualified name of resolved item and don't take into account its reexport path. For example, we generate https://doc.rust-lang.org/alloc/string/struct.String.html instead of https://doc.rust-lang.org/std/string/struct.String.html. It's not a big problem, because doc.rust-lang.org provides very similar pages for both links.
* we don't provide links to element from private module but available only via reexport. For example, `Iterator` trait.

Also, it's the first part of big plan to generate fully qualified name with similar schema as documentation link.

Part of #1875

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
@@ -5,6 +5,11 @@
 
 package org.rust.lang.core.psi.ext
 
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsQualifiedName.ChildItemType.*
+import org.rust.lang.core.psi.ext.RsQualifiedName.ParentItemType.*
+import org.rust.lang.core.types.ty.TyPrimitive
+
 interface RsQualifiedNamedElement : RsNamedElement {
     val crateRelativePath: String?
 }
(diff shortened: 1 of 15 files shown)


mehcode:
Why is the focus on external documentation when easily 90% of rust users have documentation downloaded locally?

src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
```diff
@@ -14,3 +19,162 @@ val RsQualifiedNamedElement.qualifiedName: String? get() {
...
+                    when (owner) {
+                        is RsAbstractableOwner.Trait -> owner.trait
+                        is RsAbstractableOwner.Impl -> {
+                            return owner.impl.typeReference?.toParentItem()
+                        }
+                        else -> return null
+                    }
+                }
+                is RsEnumVariant -> element.parentEnum
+                is RsFieldDecl -> element.parentStruct
```
vlad20012: Should we take into account that enums can contain fields too?

Undin:
@mehcode 
Yeah, it makes sense and we will implement such behavior eventually
But local link creation requires some infrastructural changes and we will make them in separate PR

src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
Undin: It's an interesting question
On the one hand, named field in enum variant can have qualified name
On the other hand, rustdoc doesn't add info about it in a documentation
So, let's do same things as rustdoc and don't generate qualified name for fields in enum variant now and will implement it later if it is necessary

Undin:
bors r=vlad20012

PR #2563, opened 2018-05-31
MACRO: Fix empty group 

### LINKED ISSUE #2558: macro expansion can lockup on a macro containing $()* under some circumstances
Here is a code sample that causes CLion to lock up when "Expand macros" is enabled.  The best I was able to reduce it down is to two modules:

**`lib.rs`**
```rust
#[macro_use] mod index_vec;

newtype_index!{
    #[derive(Serialize, Deserialize)]
    Layer
}
```

**`index_vec.rs`**
```rust
use std::hash::Hash;

#[macro_export]
macro_rules! newtype_index {
    ($()* $type:ident) => (
        pub struct $type(usize);

    );
}
```

---

See https://github.com/ExpHP/IntelliJ-Rust-macro-freeze for a clonable repo.

Thread dumps that appear during the lock up:

* https://raw.githubusercontent.com/ExpHP/IntelliJ-Rust-macro-freeze/master/threadDump-20180529-151600.txt
* https://raw.githubusercontent.com/ExpHP/IntelliJ-Rust-macro-freeze/master/threadDump-20180529-151611.txt

Dimonchik0036:
Fix #2558


Undin:
@Dimonchik0036 #2558 is not about problem in grammar - current grammar can parse such constructions correctly
It's a bug (infinite loop more precisely) in macro expander and let @vlad20012 fix it himself

But anyway, thanks for the attempt!

PR #2564, opened 2018-06-01
TYPE: Narrow the search scope of lambda expression parameter bindings

### LINKED ISSUE #2557: Setup search scope for lambda arguments correctly
Looks like the search scope for lambda arguments is bigger than it should be, so renaming it suggest wrong text occurrences:

image

JakubAdamWieczorek:
This makes sure that renaming a lambda parameter, with text occurrences,
does not touch comments outside the lambda expression.

Fixes #2557.


Undin:
bors r+
Thanks!

PR #2569, opened 2018-06-03
rustfmt: use write-mode and skip-children flags only when supported

### LINKED ISSUE #2579: "Reformat file with rustfmt" does not work due to unrecognized option 'write-mode'
When invoking the "Reformat file with rustfmt" shortcut, cargo fails to execute:

Cargo execution failed (exit code 1). C:/Users/Lehona/.cargo/bin/cargo.exe fmt --all -- --write-mode=overwrite --skip-children E:/documents/ChainOfResponsibility/src/main.rs stdout : Unrecognized option: 'write-mode' stderr :

I guess this option has been deprecated or changed names, although I don't know what to change it to for now.

king6cong:
See also: https://github.com/rust-lang-nursery/rustfmt/commit/5d9f5aa05a668e7dfef87b46d89dad2884df9d41

Fixes #2579

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -39,6 +39,7 @@ johnthagen
 jonas-schievink
 Keats
 kumbayo
+king6cong
 la10736
 LiamClark
 Litarvan
--- a/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
@@ -42,9 +42,12 @@ import java.nio.file.Path
  * It is impossible to guarantee that paths to the project or executables are valid,
  * because the user can always just `rm ~/.cargo/bin -rf`.
  */
+
+
 class Cargo(
     private val cargoExecutable: Path,
-    private val rustExecutable: Path
+    private val rustExecutable: Path,
+    private val rustfmtExecutable: Path
 ) {
     fun checkSupportForBuildCheckAllTargets(): Boolean {
         val lines = GeneralCommandLine(cargoExecutable)
@@ -53,6 +56,18 @@ class Cargo(
             ?: return false
 
         return lines.any { it.contains(" --all-targets ") }
+    }
+
+    data class RustfmtFlags(val emit: Boolean, val skip_children: Boolean)
+
+    private fun checkSupportForRustfmtFlags(): RustfmtFlags {
+        val lines = GeneralCommandLine(rustfmtExecutable)
+            .withParameters("-h")
+            .runExecutable()
+            ?: return RustfmtFlags(false, false)
+
+        return RustfmtFlags(lines.any { it.contains(" --emit ") },
+            lines.any { it.contains(" --skip-children ") })
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -98,7 +98,7 @@ class Cargo(
     fun reformatFile(owner: Disposable, file: VirtualFile, listener: ProcessListener? = null): ProcessOutput {
         val cmd = CargoCommandLine(
             "fmt", file.parent.pathAsPath,
-            listOf("--all", "--", "--write-mode=overwrite", "--skip-children", file.path)
+            listOf("--all", "--", "--skip-children", file.path)
```
Undin: We can't just drop this parameter because `write-mode` flag is dropped only in nightly builds
Let's check that current rustfmt don't support `write-mode` and drop this flag only in this case
You can do it in a similar way as we do for `all-targets` flag of `cargo check` command (see [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt#L112-L113))
Also, looks like that it makes sense to use `--emit=files` flag to commit behaviour of `rustfmt`

CONTRIBUTORS.txt
```diff
@@ -2,6 +2,7 @@
 # If you are wondering, how to sort lines in IDEA, then the answer is
 # https://plugins.jetbrains.com/plugin/2162
 
+king6cong
```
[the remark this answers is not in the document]
king6cong: updated

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
king6cong: updated. 
`rustfmt` is used instead of `cargo fmt` because https://github.com/rust-lang-nursery/rustfmt/pull/2768 , and I don't know how to get the project root as workingDirectory easily. 
support of `write-mode` and `skip-children` is checkd

king6cong:
@Undin ping

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -96,9 +111,21 @@ class Cargo(
 
     @Throws(ExecutionException::class)
     fun reformatFile(owner: Disposable, file: VirtualFile, listener: ProcessListener? = null): ProcessOutput {
+
+        val arguments = mutableListOf("--all", "--", file.path)
+
+        var (emit, skip_children) = checkSupportForRustfmtFlags()
+        if (!emit) {
+            arguments += "--write-mode=overwrite"
+        }
```
Undin: I'd prefer to pass `--emit=files` explicitly to commit our intention
It's true that it's current default value but default value can be changed in the future and it will change our behavior too (probably in an unexpected way)

Undin:
Ok, let's merge it
I'll fix comments myself
bors r+
Thanks!

Undin:
Hm...
bors r+

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -96,9 +111,21 @@ class Cargo(
 
     @Throws(ExecutionException::class)
     fun reformatFile(owner: Disposable, file: VirtualFile, listener: ProcessListener? = null): ProcessOutput {
+
+        val arguments = mutableListOf("--all", "--", file.path)
+
+        var (emit, skip_children) = checkSupportForRustfmtFlags()
```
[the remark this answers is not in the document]
king6cong: About working directory - before https://github.com/rust-lang-nursery/rustfmt/pull/2768 lands, the working directory has to be the directory that Cargo.toml residents. I have tried `file.parent.pathAsPath` and it's not enough.

```
congs-iMac:lib cong$ pwd
/Users/cong/repos/nebulet/lib
congs-iMac:lib cong$ cargo fmt -- --help
Failed to find targets
usage: cargo fmt [options]
...
```

king6cong:
Thanks, was on a vacation and sorry about not responding timely 😄

PR #2585, opened 2018-06-15
ModName should take into account #[path] attribute of mod declaration

### LINKED ISSUE #2584: In some cases module names deterined by file name, not by mod directive
I have next code:

src/lib.rs:
```
#[cfg(test)]
#[path="foo.rs"]
mod test;
```
src/foo.rs:

```
#[test]
fn test_foo(){
    println!("Foo");
}

```

When I clicked to ">" icon on the line with `fn test_foo()`, command `cargo test --color=always --package proba8 --lib foo::test_foo -- --nocapture --exact` invoked, but no tests are executed (with `ok` result). This can be confusing

Correct argument may be `--lib test::test_foo `

Undin:
In the following case
```rust
//- main.rs
#[path="bar.rs"]
mod foo;

//- bar.rs
fn foo() {}
```
fully qualified name of `foo` function is `foo::foo`.
But currently, we don't take into account module declaration and its attributes and return file system name as mod name for file. Because of it we return incorrect `bar::foo` in case above.

This change fixes it.

Fixes #2584

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/RsDowngradeModuleToFile.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/RsDowngradeModuleToFile.kt
@@ -5,55 +5,70 @@
 
 package org.rust.lang.refactoring
 
+import com.intellij.lang.Language
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.command.WriteCommandAction.runWriteCommandAction
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiDirectory
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
 import com.intellij.psi.PsiFileSystemItem
+import com.intellij.psi.impl.file.PsiFileImplUtil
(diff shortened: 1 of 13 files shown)


src/main/kotlin/org/rust/lang/refactoring/RsDowngradeModuleToFile.kt
```diff
@@ -50,10 +51,10 @@ private fun contractModule(fileOrDirectory: com.intellij.psi.PsiFileSystemItem)
 
 private val com.intellij.psi.PsiElement.isDirectoryMod: Boolean get() {
     return when (this) {
-        is org.rust.lang.core.psi.RsFile -> name == org.rust.lang.core.psi.ext.RsMod.Companion.MOD_RS && containingDirectory?.children?.size == 1
+        is org.rust.lang.core.psi.RsFile -> name == RsConstants.MOD_RS_FILE && containingDirectory?.children?.size == 1
```
vlad20012: Hmm, why we don't use imports here? If we anyway editing this line, maybe import it?

Undin: done

PR #2593, opened 2018-06-19
TY&RES: fixed method pick logic

### LINKED ISSUE #2440: Incorrect "mismatched types [E0308]" on `clone`
In a method which adds two sets and returns the result as a new set:
```rust
    fn add_set(&self, other: &USet) -> USet {
        if self.is_empty() {
            other.clone()
        } else if other.is_empty() {
            self.clone()
        } else {
             // ...           
             USet { /* ... */ }
        }
    }
```

Intellij marks this code in red and displays "mismatched types [E0308] expected `USet`, found `&USet`". The code is valid. The problem lies in determining what is the result of the `clone()` operation. If the `if/else if/else` sequence is commented out, Intellij stops treating it as an error. It stops as well if I attempt to dereference the result of `clone`, ie. switching to:
```
...
*other.clone()
...
*self.clone()
...
```
stops Intellij from marking the code as erroneous. But this in fact is now an error, and `rustc` will not compile it.

Intellij IDEA 2018.1 CE, 181.4203.550
Rust plugin 0.2.0.2092.181

### LINKED ISSUE #2460: (False) type mismatch errors for &f64
Since:

* IntelliJ Rust Plugin 0.2.0.2092.181
* IntelliJ IDEA 2018.1.1 (Ultimate Edition)
* Build #IU-181.4445.78, built on April 9, 2018
* JRE: 1.8.0_152-release-1136-b27 x86_64
* JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
* macOS 10.13.4

I get lots of (false) type mismatch errors, rendering IDE unusable.

One of the simpler ones is:

```
/// simple-minded comparison of f64s; may panic if NaN etc.
/// 
fn f64compare(x: &f64, y: &f64) -> ::std::cmp::Ordering
{
    x.partial_cmp(y).unwrap()
}
```

Flags the `y` parameter to `.partial_cmp` as an error with the message:

```
mismatched types [E0308] expected `&&f64`, found `&f64`
```

Though it compiles and runs perfectly.

I notice lots of issues (#2449 #2452 #2451 #2448) around this area of type mismatches recently — could they be related?

### LINKED ISSUE #2482: Mismatched types
Clion 2018.1
OS X 10.13.4

```rust
    let string: String = "string".to_owned();  // Mismatched types [#0308]
    println!("{}", string);
```

Actually, `to_owned()` can create an owned `String` from a string slice.

### LINKED ISSUE #2514: TY does not work after clone()
When we are doing `.clone()` after data container unwrapping, type inference does not work.
Tested on: `Vec<>`, `Box<>`, `Rc<>`, `RefCell<>` and `Arc<>`.
Example:
```rust
struct Foo {
    bar: Box<i32>
}

impl Foo {
    fn test(&self) {
        let b = self.bar.as_ref().clone();
        b // variable b: <unknown>
    }
}
```
System info:
OS: Windows 7
intellij-rust version: 0.2.0.2093-181
IDE: IntelliJ IDEA Community Edition 2018.1.3 EAP(Build 181.4892.5)
rustup: 1.11.0 (e751ff9f8 2018-02-13)
rustc: 1.27.0-nightly (ac3c2288f 2018-04-18)
Toolchain: nightly-x86_64-pc-windows-msvc (default)

vlad20012:
Fixes #2482 
Fixes #2460 
Fixes #2440 
Fixes #2514

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -31,6 +31,7 @@ import org.rust.lang.utils.RsDiagnostic
 import org.rust.openapiext.Testmark
 import org.rust.openapiext.forEachChild
 import org.rust.openapiext.recursionGuard
+import org.rust.stdext.notEmptyOrLet
 import org.rust.stdext.singleOrFilter
 import org.rust.stdext.singleOrLet
 import org.rust.stdext.zipValues
@@ -958,7 +959,7 @@ private class RsFnInferenceContext(
         val argExprs = methodCall.valueArgumentList.exprList
         val callee = run {
             val variants = resolveMethodCallReferenceWithReceiverType(lookup, receiver, methodCall)
-            val callee = pickSingleMethod(variants, methodCall)
+            val callee = pickSingleMethod(receiver, variants, methodCall)
             // If we failed to resolve ambiguity just write the all possible methods
             val variantsForDisplay = (callee?.let(::listOf) ?: variants).map { it.element }
             ctx.writeResolvedMethod(methodCall, variantsForDisplay)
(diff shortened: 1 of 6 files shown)


vlad20012:
bors try

Undin:
LGTM
bors r+

PR #2594, opened 2018-06-19
RES: New macro resolve

Undin:
* Rename `RsMacroDefinition*` to `RsMacro*`
* Implement new macro resulition logic
* Refactor `RustProjectDescriptor`s

PR #2607, opened 2018-06-24
RES: New macro resolve 2

Undin:
It's the second attempt to implement macro resolution according to the latest changes in macro reexport syntax.

Improvements in comparison with the [previous version](https://github.com/intellij-rust/intellij-rust/pull/2594):
* Don't try to resolve base path of use item if its name is not extern crate name. As a result new macro resolution with this check improves whole performance of our highlighting tests by 10%-20%
* Fix `NoClassDefFoundError` on CLion described [here](https://github.com/intellij-rust/intellij-rust/issues/2597#issuecomment-399089894).


Undin:
bors r=vlad20012

PR #2608, opened 2018-06-24
GRAM: fall back to common macro parsing if log parser fails

Undin:
Currently, we have special rules in grammar for several types of macros including logs macros.
Rules for log macros are based on macros from [log](https://docs.rs/log) crate.
But macros from [slog](https://docs.rs/slog) crate have same names but different grammar.
Unfortunately, current grammar doesn't fall back to common rule for macros if log macro parser fails and it leads to syntax error for call of macros from slog crate.

As I understand such special rules were introduced to show syntax error for well-known macros and maybe do other useful code insight things. But in this case, I think we should keep special rule for log macros but use common grammar when parser can't apply the special rule to parse macro call. To show syntax errors we can implement some annotator.


Undin:
bors r=vlad20012

Issue #2327, opened 2018-02-24
Could't resolve URL error in quick doc


JonathanxD:
Plugin does not resolve relative documentation links, example, in `println!`documentation there is a relative link to `format!` (`../std/macro.format.html`):

![println!](https://cdn.pbrd.co/images/H98x5Mu.png)

But when I click in the `format!`:

![format!](https://cdn.pbrd.co/images/H98xEOC.png)

Plugin Configuration:

![rust](https://cdn.pbrd.co/images/H98yan3.png)

The path to documentation (in linux at least) is: `~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/`. (Relative to toolchain location is: `../share/doc/rust/html/`).

**Note: clicking on *project settings* link does nothing, I don't know if this is a bug too.***

Issue #667, opened 2016-09-14
Follow relative Markdown reference links in stdlib documentation


johnthagen:
Currently the Rust stdlib documentation uses a lot of relative references in the Markdown comments.  The fact that `intellij-rust` recognizes that they are links and tries to resolve them is very promising, but currently they do not resolve correctly.

If we consider [`Box`](https://doc.rust-lang.org/std/boxed/struct.Box.html), note the relative path `../../std/boxed/index.html`. 

``` rust
/// A pointer type for heap allocation.
///
/// See the [module-level documentation](../../std/boxed/index.html) for more.
#[lang = "owned_box"]
#[stable(feature = "rust1", since = "1.0.0")]
pub struct Box<T: ?Sized>(Unique<T>);
```

![box-doc](https://cloud.githubusercontent.com/assets/10340167/18495745/5951a8de-79ee-11e6-800f-b3dfffde5735.png)

Clinking the link results in:

![box-link-error](https://cloud.githubusercontent.com/assets/10340167/18495748/6186093c-79ee-11e6-8633-6d62a891cdd7.png)

 Could this somehow be translated into:
- `https://doc.rust-lang.org/std/boxed/index.html`

I think this could be a powerful tool, especially for those new to Rust and browsing documentation as they develop.

mkaput:
Unfortunately, we don't do any detection, we only highlight them :( The problem is, rustdoc links haven't been standardised yet afaik.

Kobzol:
@Undin it seems to me that stdlib links (both relative and absolute) work in quick documentation. Is this issue resolved or is the plan to also support them in normal documentation comments?

Undin:
@Kobzol we still can't resolve some links. For example, you can check `Eq` link in `HashMap` documentation
In general, the plugin can't resolve relative links when both documentation owner and link target are re-exported and link uses path between reexported names
But it's a more narrow thing so I think we cal close this issue and open another one with more specific problem

Undin:
In general, link resolution works. Known bug is filed in #5374

PR #2618, opened 2018-06-28
QDOC: initial support of documentation links

Undin:
Rust doc comments may contain relative links like `struct.String.html`, `#method.clear`, `../../std/macro.vec.html`, etc.
Currently, we don't process them and final documentation text in quick doc contains relative links
which can't be opened.
These changes do the following things:
* resolve all relative links with base URI, where base URI is constructed from current item path
* use `psi_element://` scheme to make quick doc processes such links
* try to parse link content in `RsDocumentationProvider#getDocumentationElementForLink` into `RsQualifiedName` and find `RsElement` with the same qualified name object.

Note: a lot of links in stdlib provide paths based on reexports. This implementation does **NOT** support such path, i.e. `RsDocumentationProvider#getDocumentationElementForLink` fails to find element with required qualified name.

Note 2: there are relative links in documentation that refer to not language items, for example, to some documentation page. Now we don't distinguish them and process as language items. For users it doesn't break anything because such links didn't work before.

Initial part of #667 and #2327

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
@@ -32,8 +32,7 @@ data class RsQualifiedName private constructor(
     fun toUrlPath(): String {
         val segments = mutableListOf(crateName)
         segments += modSegments
-        val (pageName, anchor) =
-        if (parentItem.type == MOD) {
+        val (pageName, anchor) = if (parentItem.type == MOD || parentItem.type == CRATE) {
             "index.html" to ""
         } else {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
```diff
@@ -40,11 +40,80 @@ class RsQualifiedName private constructor(
         return segments.joinToString(separator = "/", postfix = anchor)
     }
 
+    override fun equals(other: Any?): Boolean {
```
vlad20012: I suspect that we can't just use data class here? It would be nice to leave a comment, why so.

Undin: fixed

PR #2621, opened 2018-06-29
PERF: introduce fn-local modification trackers

vlad20012:
Now we will invalidate type inference cache only if a current function or program structure was changed. I.e. if we type in some function, inferred types of other functions will not be flushed.

Should fix most of "high CPU usage" issues

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiManager.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiManager.kt
@@ -1,0 +1,88 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi
+
+import com.intellij.openapi.components.ProjectComponent
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.ModificationTracker
+import com.intellij.openapi.util.SimpleModificationTracker
+import com.intellij.psi.*
+import com.intellij.psi.impl.PsiTreeChangeEventImpl
+import com.intellij.psi.util.PsiTreeUtil.getContextOfType
+import org.rust.lang.RsFileType
+import org.rust.lang.core.psi.ext.RsItemElement
+import org.rust.lang.core.psi.ext.RsModificationTrackerOwner
+
+interface RsPsiManager {
+    /**
+     * A project-global modification tracker that increments on each PSI change that can affect
+     * name resolution or type inference. It will be incremented with a change of most types of
+     * PSI element excluding function bodies (expressions and statements)
+     */
+    val rustStructureModificationTracker: ModificationTracker
+}
+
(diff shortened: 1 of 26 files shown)


vlad20012:
Hmm, carg-fmt failed on nightly

src/main/kotlin/org/rust/lang/core/psi/RsPsiManager.kt
```diff
@@ -0,0 +1,88 @@
...
+            updateModificationCount(child ?: event.parent)
+        }
+    }
+
+    private fun updateModificationCount(psi: PsiElement) {
+        // We find a nearest parent item or macro call (because macro call can produce items)
+        // If found item implements RsModificationTrackerOwner, we increment it's own
+        // modification counter. Otherwise we increment global modification counter.
+        //
+        // So, if we changing something inside a functions and it is not an item,
```
Undin: Let's better write something like this:
"So, if something is changed inside a function except an item, we will only increment the function local modification counter."

src/main/kotlin/org/rust/lang/core/psi/RsPsiManager.kt
```diff
@@ -0,0 +1,88 @@
...
+        // we incrementing the local modification counter of the function only.
+        //
+        // It may not be intuitive that if we changing an item inside a function,
```
Undin: changing -> change

src/main/kotlin/org/rust/lang/core/psi/ext/PsiElement.kt
```diff
@@ -96,3 +96,6 @@ fun PsiElement?.getPrevNonCommentSibling(): PsiElement? =
  */
 fun PsiElement?.getNextNonCommentSibling(): PsiElement? =
     PsiTreeUtil.skipSiblingsForward(this, PsiWhiteSpace::class.java, PsiComment::class.java)
+
+fun RsElement.isParentOf(child: PsiElement): Boolean =
```
matklad: `isAncestorOf`? Or change `ancestors` to `parents` :)

src/test/kotlin/org/rustSlowTests/RsHighlightingPerformanceTest.kt
```diff
@@ -67,6 +69,20 @@ class RsHighlightingPerformanceTest : RustWithToolchainTestBase() {
...
+
+        val stmt = myFixture.file.descendantsOfType<RsFunction>().asSequence()
+            .mapNotNull { it.block?.stmtList?.lastOrNull() }.first()
+        myFixture.editor.caretModel.moveToOffset(stmt.textOffset)
+        myFixture.type("2+2;")
+        PsiDocumentManager.getInstance(project).commitAllDocuments() // process PSI modification events
+
+        timings.measure("resolve_after_typing") {
+            refs.forEach { it.reference.resolve() }
+        }
```
matklad: :+1:

PR #2628, opened 2018-07-02
TY: supported match ergonomics

### LINKED ISSUE #2457: Support match ergonomics
See [RFC 2005](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md)

### LINKED ISSUE #2478: Reference when destructuring from a struct
Given the following code:
```rust
struct Struct {
    foo: String,
}
let mut s = Struct {
    foo: String::from("bar"),
};
let Struct { ref foo } = &mut s;
```
Intellij-rust seems to think the destructured `foo` is of type `String` when in actuality it's a `&String`.

vlad20012:
Supported match ergonomics ([RFC 2005](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md))
```rust
fn main() {
    let (a, b) = &(1, 2);
    (a, b);
} //^ (&i32, &i32)
```

Fixes #2457 
Fixes #2478


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1482,74 +1482,74 @@ private class RsFnInferenceContext(
...
                 ctx.writeBindingTy(patBinding, bindingType)
                 pat?.extractBindings(type)
             }
             is RsPatTup -> {
-                val types = (type as? TyTuple)?.types.orEmpty()
+                val (derefTy, mut) = type.stripReferences()
+                val types = (derefTy as? TyTuple)?.types.orEmpty()
                 for ((idx, p) in patList.withIndex()) {
-                    p.extractBindings(types.getOrElse(idx, { TyUnknown }))
+                    p.extractBindings(types.getOrElse(idx) { TyUnknown }.toRef(mut))
```
Undin: Looks like that we need pass `mut != null` as the second argument
It should fix the following case:
```rust
fn main() {
    let p = (123, 123);
    let (x, ref y) = &p;
}
```
I this case `x` and `y` have `&i32` type but we infer that type of `y` is `&&i32`
screen shot 2018-07-09 at 15 01 38

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1482,74 +1482,74 @@ private class RsFnInferenceContext(
         }
     }
 
-    private fun RsPat.extractBindings(type: Ty) {
+    private fun RsPat.extractBindings(type: Ty, ignoreRef: Boolean = false) {
```
Undin: Maybe better to convert `ignoreRef` to `addRef` (or some better name with similar meaning)?
Looks like it would be closer to rustc desugaring described in RFC [examples](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md#examples)

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1648,6 +1648,19 @@ data class TyWithObligations<out T>(
...
+    var isMut = true
+    var ty = this
+    while (ty is TyReference) {
+        isMut = isMut && ty.mutability.isMut
+        ty = ty.referenced
+    }
+    return ty to (if (this is TyReference) Mutability.valueOf(isMut) else null)
+}
+
+private fun Ty.toRef(mut: Mutability?): Ty =
```
Undin: I don't like method name because its calls make me think that it always converts type to reference type
Maybe rename it to something like `toRefIfNeeded` (really, I don't like this name - it's too hard to read) and move it into `RsPat.extractBindings` because this method is too specific and should be used only in `extractBindings`

src/test/kotlin/org/rust/lang/core/type/RsPatternMatchingTest.kt
```diff
@@ -181,4 +217,100 @@ class RsPatternMatchingTest : RsTypificationTestBase() {
...
+            let [a, b] = &[1, 2];
+            a;
+        } //^ &i32
+    """)
+
+    fun `test match ergonomics slice 2`() = testExpr("""
+        fn main() {
+            let [ref a, b] = &[1, 2];
+            a;
+        } //^ &i32
```
Undin: Let's add similar tests for all pat binding types to make sure that we correctly support `ref` pattern

PR #2632, opened 2018-07-02
PERF: use rust structure mod tracker in `ProjectCache`

vlad20012:
ProjectCache is intensive used in type inference, so this change should speedup type inference in some function after its body change (e.g. completion)

As I see, all current usages of ProjectCache can safely depend on rust structure mod tracker instead of global psi mod tracker


Undin:
Performance improvement by this small change looks awesome!

PR #2643, opened 2018-07-05
Expand macros in trait and impl bodies

vlad20012:
Now we can expand macros in this context:
```rust
impl Foo {
    foo!();
    bar!();
}
```
and can resolve method generated by these macros.
Also fixed some inspections to work properly with members defined by a macro.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
@@ -10,7 +10,6 @@ import org.rust.lang.core.psi.RsBlock
 import org.rust.lang.core.psi.RsFile
 import org.rust.lang.core.psi.RsMacroCall
 import org.rust.lang.core.psi.RsModItem
-import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
 
 interface RsItemsOwner : RsElement
 
@@ -42,27 +41,11 @@ val RsItemsOwner.itemsAndMacros: Sequence<RsElement>
     }
 
 fun RsItemsOwner.processExpandedItems(processor: (RsItemElement) -> Boolean): Boolean {
-    for (psi in itemsAndMacros) {
-        when (psi) {
-            is RsMacroCall ->
-                if (processMacroCall(psi, processor, 0)) return true
-
-            is RsItemElement ->
-                if (processor(psi)) return true
-        }
-    }
-
-    return false
+    return itemsAndMacros.any { it.processItem(processor) }
 }
 
-fun processMacroCall(call: RsMacroCall, processor: (RsItemElement) -> Boolean, recursionDepth: Int): Boolean {
-    if (recursionDepth > DEFAULT_RECURSION_LIMIT) return true
-    for (expanded in call.expansion.orEmpty()) {
-        when (expanded) {
-            is RsItemElement -> if (processor(expanded)) return true
-            is RsMacroCall -> processMacroCall(expanded, processor, recursionDepth + 1)
-        }
-    }
-
-    return false
+private fun RsElement.processItem(processor: (RsItemElement) -> Boolean) = when (this) {
+    is RsMacroCall -> processExpansionRecursively { it is RsItemElement && processor(it) }
+    is RsItemElement -> processor(this)
+    else -> false
 }
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsMacroCall.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsMacroCall.kt
@@ -13,6 +13,7 @@ import org.rust.lang.core.macros.ExpansionResult
 import org.rust.lang.core.macros.expandMacro
 import org.rust.lang.core.psi.RsElementTypes.IDENTIFIER
 import org.rust.lang.core.psi.RsMacroCall
+import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
 import org.rust.lang.core.resolve.ref.RsMacroCallReferenceImpl
 import org.rust.lang.core.resolve.ref.RsReference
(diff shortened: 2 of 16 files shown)


src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
```diff
@@ -80,28 +79,26 @@ private fun insertNewTraitMembers(selected: Collection<RsAbstractable>, existing
         }
 
         val anchor = indexedExistingMembers
-            .map { IndexedValue(it.index, it.value.first) }
-            .lastOrNull()
+            .map {
+                val member = it.value.first
+                IndexedValue(it.index, member.expandedFromRecursively ?: member)
+            }.lastOrNull()
             ?: IndexedValue(-1, existingMembers.lbrace) ?: return
```
fedochet: This is strange `return`; do we really need it? Constructor cannot return nullable value (I guess)

src/main/kotlin/org/rust/lang/core/psi/ext/RsMembers.kt
```diff
@@ -0,0 +1,41 @@
...
+
+/** Returns all members including those produced by macros */
+val RsMembers.expandedMembers: List<RsAbstractable>
+    get() {
+        val members = mutableListOf<RsAbstractable>()
+
+        for (member in childrenSequence) {
+            when (member) {
+                is RsAbstractable -> members.add(member)
+                is RsMacroCall -> member.processExpansionRecursively {
```
fedochet: Maybe we could add addditional function to this place, so this `false` in the air will not look so strange?
Like this: 
```kotlin
private fun RsMacroCall.collectAbstractableMembersRecursively(): List<RsAbstractable> {
    val members = mutableListOf<RsAbstractable>()

    this.processExpansionRecursively { 
        if (it is RsAbstractable) {
            members.add(it)
        }

        false
    }

    return members
}
```
and then use it like 
```kotlin
... -> members.addAll(member.collectAbstractableMembersRecursively())
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsMacroCall.kt
```diff
@@ -53,6 +54,21 @@ val RsMacroCall.expansion: List<ExpansionResult>?
         expandMacro(this)
     }
 
+fun RsMacroCall.processExpansionRecursively(processor: (ExpansionResult) -> Boolean): Boolean =
+    processExpansionRecursively(processor, 0)
+
+private fun RsMacroCall.processExpansionRecursively(processor: (ExpansionResult) -> Boolean, recursionDepth: Int): Boolean {
```
fedochet: Same here: we can improve this function by using `any` method of the list, like this:
```kotlin
private fun RsMacroCall.processExpansionRecursively(processor: (ExpansionResult) -> Boolean, recursionDepth: Int): Boolean {
    if (recursionDepth > DEFAULT_RECURSION_LIMIT) return true

    return expansion.orEmpty().any { it.processRecursively(processor, recursionDepth) }
}

private fun ExpansionResult.processRecursively(processor: (ExpansionResult) -> Boolean, recursionDepth: Int) = when (this) {
    is RsMacroCall -> this.processExpansionRecursively(processor, recursionDepth + 1)
    else -> processor(this)
}
```

This way there would be only one return point and it is easier to read (imo)

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
```diff
@@ -80,28 +79,26 @@ private fun insertNewTraitMembers(selected: Collection<RsAbstractable>, existing
...
         existingMembersWithPosInTrait.add(anchor.index + 1, Pair(addedMember, posInTrait))
 
         // If the newly added item is a function, we add an extra line between it and each of its siblings.
-        val previousAbstractable = addedMember.leftSiblings.find { it is RsAbstractable }
-        if (previousAbstractable != null) {
-            if (previousAbstractable is RsFunction || addedMember is RsFunction) {
-                val whitespaces = createExtraWhitespacesAroundFunction(previousAbstractable, addedMember)
-                existingMembers.addBefore(whitespaces, addedMember)
-            }
+        val prev = addedMember.leftSiblings.find { it is RsAbstractable || it is RsMacroCall }
```
fedochet: Those two pieces are almost identical, maybe we should extract a function here

src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
```diff
@@ -45,7 +44,7 @@ fun RsItemsOwner.processExpandedItems(processor: (RsItemElement) -> Boolean): Bo
     for (psi in itemsAndMacros) {
```
fedochet: Forgot to add that here we can also use `any` method:
```kotlin
fun RsItemsOwner.processExpandedItems(processor: (RsItemElement) -> Boolean): Boolean {
    return itemsAndMacros.any { it.processItem(processor) }
}

private fun RsElement.processItem(processor: (RsItemElement) -> Boolean) = when (this) {
    is RsMacroCall -> this.processExpansionRecursively { it is RsItemElement && processor(it) }

    is RsItemElement -> processor(this)

    else -> false
}

```

src/main/kotlin/org/rust/lang/refactoring/implementMembers/impl.kt
vlad20012: There are a lot of differences: leftSiblings/rightSiblings prev/nex/addedMember order in `createExtraWhitespacesAroundFunction` call parameters, addBefore/addAfter. So I don't know how to generalize it.

fedochet: You're right, missed that

vlad20012:
bors r=fedochet

PR #2646, opened 2018-07-06
TY: Fix type inference in FieldExpr

ortem:


src/main/kotlin/org/rust/lang/core/resolve/ref/RsMethodCallReferenceImpl.kt
```diff
@@ -78,13 +78,27 @@ fun resolveFieldLookupReferenceWithReceiverType(
...
-data class MethodCallee(
+interface DotExprResolveVariant : ScopeEntry {
+    val selfTy: Ty;
+    val derefCount: Int
+}
+
+data class FieldResolveVariant(
+    override val name: String,
+    override val element: RsElement,
+    /** The receiver type after possible derefs performed */
```
vlad20012: Let's move these docs to the parent class

src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -1431,4 +1431,34 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
             a;
         } //^ X
     """)
+
+    fun `test field of parent with deref`() = testExpr("""
```
vlad20012: Let's rename it to "test field type substitution after deref".
Also, this test can be written much more compact. We can write structs in a single line `struct S1<A> { field1: A }`, remove field2 from S2, remove `deref` method from Deref trait and impl.

PR #2647, opened 2018-07-06
Improve dependencies completion in Cargo.toml

### LINKED ISSUE #2856: Cargo.toml crate completion doesn't work for target-specific dependencies
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2104-181
* **Rust toolchain version:** rustc 1.28.0 (9634041f0 2018-07-30)
* **IDE name and version:** IntelliJ IDEA 2018.1.5 (Community Edition)

## Problem description

Completion of crates.io crates doesn't seem to work inside target-specific dependency keys like `[target.'cfg(windows)'.dependencies]`

## Steps to reproduce

Try typing crate names in this Cargo.toml:

```toml
[package]
name = "anything"
version = "0.1.0"
authors = ["Jonas Schievink <jonasschievink@gmail.com>"]

[dependencies]

[dev-dependencies]

[target.'cfg(windows)'.dependencies]

[target.'cfg(unix)'.dependencies]
```

The first 2 dependency keys (`dependencies` and `dev-dependencies`) work fine, the last 2 don't.

Presumably, this is caused by this line, which only looks at the first key (here, `target`) to determine whether completion should be done:

https://github.com/intellij-rust/intellij-rust/blob/6a24958d7dcd86b28edbfd0ad29da2f169fe564b/src/main/kotlin/org/rust/toml/RsTomlCompletionProvider.kt#L56

Instead of looking at the first element, it should probably check the *last* one.

vlad20012:
With this PR we should be able to complete versions of dependencies in cargo toml when caret is placed to toml value position, i.e.
```
[dependencies]
rand <caret>
rand = <caret>
rand = "<caret>"

[dependencies.<caret>]
<caret>
```

Fixes #2856

### DIFF
--- a/src/main/kotlin/org/rust/toml/CargoTomlPsiPattern.kt
+++ b/src/main/kotlin/org/rust/toml/CargoTomlPsiPattern.kt
@@ -1,0 +1,149 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml
+
+import com.intellij.patterns.PsiElementPattern
+import com.intellij.patterns.StandardPatterns
+import com.intellij.patterns.VirtualFilePattern
+import com.intellij.psi.PsiElement
+import org.rust.cargo.toolchain.RustToolchain
+import org.rust.lang.core.psiElement
+import org.rust.lang.core.with
+import org.toml.lang.psi.TomlKey
+import org.toml.lang.psi.TomlKeyValue
+import org.toml.lang.psi.TomlTable
+import org.toml.lang.psi.TomlTableHeader
+
+object CargoTomlPsiPattern {
+    private const val TOML_KEY_CONTEXT_NAME = "key"
+    private const val TOML_KEY_VALUE_CONTEXT_NAME = "keyValue"
+
+    private inline fun <reified I : PsiElement> cargoTomlPsiElement(): PsiElementPattern.Capture<I> {
+        return psiElement<I>().inVirtualFile(
+            StandardPatterns.or(
+                VirtualFilePattern().withName(RustToolchain.CARGO_TOML),
+                VirtualFilePattern().withName(RustToolchain.XARGO_TOML)
+            )
+        )
+    }
+
+    private inline fun <reified I : PsiElement> cargoTomlPsiElement(contextName: String): PsiElementPattern.Capture<I> {
+        return cargoTomlPsiElement<I>().with("putIntoContext") { e, context ->
+            context?.put(contextName, e)
+            true
+        }
+    }
+
+    /** Any element inside any TomlKey in Cargo.toml */
+    val inKey: PsiElementPattern.Capture<PsiElement> =
+        cargoTomlPsiElement<PsiElement>()
+            .withParent(TomlKey::class.java)
+
+    private val onDependencyTableHeader: PsiElementPattern.Capture<TomlTableHeader> =
+        cargoTomlPsiElement<TomlTableHeader>()
+            .with("dependenciesCondition") { header ->
+                header.isDependencyListHeader
+            }
+
+    private val onDependencyTable: PsiElementPattern.Capture<TomlTable> =
+        cargoTomlPsiElement<TomlTable>()
+            .withChild(onDependencyTableHeader)
+
+    /**
+     * ```
+     * [dependencies]
+     * regex = "1"
+     *   ^
+     * ```
+     */
+    val onDependencyKey: PsiElementPattern.Capture<TomlKey> =
+        cargoTomlPsiElement<TomlKey>()
+            .withSuperParent(
+                2,
+                onDependencyTable
+            )
+
+    private val onSpecificDependencyTable: PsiElementPattern.Capture<TomlTable> =
+        cargoTomlPsiElement<TomlTable>()
+            .withChild(
+                psiElement<TomlTableHeader>()
+                    .with("specificDependencyCondition") { header, context ->
+                        val names = header.names
+                        names.getOrNull(names.size - 2)?.isDependencyKey == true
+                    }
+            )
+
+    /**
+     * ```
+     * [dependencies.regex]
+     *                 ^
+     * ```
+     */
+    val onSpecificDependencyHeaderKey: PsiElementPattern.Capture<TomlKey> =
+        cargoTomlPsiElement<TomlKey>(TOML_KEY_CONTEXT_NAME)
+            .withParent(
+                psiElement<TomlTableHeader>()
+                    .with("specificDependencyCondition") { header, context ->
+                        val key = context?.get(TOML_KEY_CONTEXT_NAME) ?: return@with false
+                        val names = header.names
+                        names.getOrNull(names.size - 2)?.isDependencyKey == true && names.lastOrNull() == key
+                    }
+            )
+
+    /** Any element inside [onSpecificDependencyHeaderKey] */
+    val inSpecificDependencyHeaderKey: PsiElementPattern.Capture<PsiElement> =
+        cargoTomlPsiElement<PsiElement>()
+            .withParent(onSpecificDependencyHeaderKey)
+
+    /**
+     * Any element inside [onDependencyKeyValue].
+     * This pattern doesn't permit nested keyValues, e.g. false for
+     * ```
+     * [dependencies]
+     * regex = { version = "1" }
+     *                   ^
+     * ```
+     */
+    val inDependencyKeyValue: PsiElementPattern.Capture<PsiElement> =
+        cargoTomlPsiElement<PsiElement>()
+            .inside(psiElement<TomlKeyValue>(TOML_KEY_VALUE_CONTEXT_NAME))
+            .with("dependencyKeyValueCondition") { _, context ->
+                val keyValue = context?.get(TOML_KEY_VALUE_CONTEXT_NAME) ?: return@with false
+                onDependencyKeyValue.accepts(keyValue)
+            }
+
+    /**
+     * ```
+     * [dependencies]
+     * regex = "1"
+     *       ^
+     * ```
+     */
+    private val onDependencyKeyValue: PsiElementPattern.Capture<TomlKeyValue> =
+        cargoTomlPsiElement<TomlKeyValue>()
+            .withParent(onDependencyTable)
+
+    /** Any element inside [onSpecificDependencyKeyValue] */
+    val inSpecificDependencyKeyValue: PsiElementPattern.Capture<PsiElement> =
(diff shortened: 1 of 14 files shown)


vlad20012:
Ready for review

src/main/kotlin/org/rust/toml/CratesIoApi.kt
```diff
@@ -0,0 +1,88 @@
...
+    val name: String,
+    @SerializedName("max_version")
+    val maxVersion: String
+)
+
+val CrateDescription.dependencyLine: String
+    get() = "$name = \"$maxVersion\""
+
+fun searchCrate(key: TomlKey): Collection<CrateDescription> {
+    MOCK?.let { return it }
```
Undin: I like the way how we implement other mocks, for example, for extract function refactoring
More precisely, I like the error if `MOCK` variable in tests is null and error message.
Such error prevents you from accidental errors in tests and message says exactly what you should do
It would be great use such pattern here as well

src/main/kotlin/org/rust/toml/CratesIoApi.kt
```diff
@@ -0,0 +1,88 @@
...
+
+private inline fun <reified T> requestCratesIo(context: PsiElement, path: String): T? {
+    return requestCratesIo(context, path, T::class.java)
+}
+
+private fun <T> requestCratesIo(context: PsiElement, path: String, cls: Class<T>): T? {
+    return try {
+        runWithCheckCanceled {
+            val response = HttpRequests.request("https://crates.io/api/v1/$path")
+                .readString(ProgressManager.getInstance().progressIndicator)
```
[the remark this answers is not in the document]
Undin: Okay, `HttpURLConnection` can reuse connections, but I still don't like it because of other comments))

src/main/kotlin/org/rust/toml/CargoTomlDependenciesCompletionProvider.kt
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    protected abstract fun completeDependencyKey(key: TomlKey, result: CompletionResultSet)
+
+    protected abstract fun completeDependencyValue(keyValue: TomlKeyValue, result: CompletionResultSet)
+}
+
+/** @see CargoTomlPsiPattern.inDependencyKeyValue */
+class CargoTomlDependenciesCompletionProvider : AbstractTomlKeyValueCompletionProvider() {
```
Undin: Let's use the same naming style for these providers and replace `Dependencies` to `Dependency`

src/main/kotlin/org/rust/toml/CargoTomlDependenciesCompletionProvider.kt
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        if (!hasQuotes) {
+            context.document.insertString(startOffset, "\"")
+            context.document.insertString(context.selectionEndOffset, "\"")
+        }
+    }
+}
+
+private inline fun <reified T : PsiElement> InsertionContext.getItemOfType(strict: Boolean = false): T? =
```
Undin: Let's drop this function and use `org.rust.lang.core.completion.LookupElementsKt#getElementOfType` instead

src/main/kotlin/org/rust/toml/CargoTomlDependenciesCompletionProvider.kt
```diff
@@ -0,0 +1,136 @@
...
+            ?: error("PsiElementPattern must not allow KeyValues in tables without header")
+    }
+}
+
+/** Inserts `=` between key and value if missed and wraps inserted string with quotes if needed */
+private class StringValueInsertionHandler(val keyValue: TomlKeyValue) : InsertHandler<LookupElement> {
+    override fun handleInsert(context: InsertionContext, item: LookupElement) {
+        var startOffset = context.startOffset
+        val value = context.getItemOfType<TomlValue>()
+        val hasEq = keyValue.children.find { it.elementType == TomlElementTypes.EQ } != null
```
Undin: `find { someCondition } != null` -> `any { someCondition }`?

src/main/kotlin/org/rust/toml/CargoTomlPsiPattern.kt
```diff
@@ -0,0 +1,158 @@
...
+                        val key = context?.get(TOML_KEY_CONTEXT_NAME) ?: return@with false
+                        val names = header.names
+                        names.getOrNull(names.size - 2)?.isDependencyKey == true && names.lastOrNull() == key
+                    }
+            )
+
+    /** Any element inside [onSpecificDependencyHeaderKey] */
+    val inSpecificDependencyHeaderKey: PsiElementPattern.Capture<PsiElement> =
+        cargoTomlPsiElement<PsiElement>()
+            .withSuperParent(1, onSpecificDependencyHeaderKey)
```
Undin: `withSuperParent(1, onSpecificDependencyHeaderKey)` -> `withParent(onSpecificDependencyHeaderKey)`

src/main/kotlin/org/rust/toml/CargoTomlPsiPattern.kt
```diff
@@ -0,0 +1,158 @@
...
+     * regex = "1"
+     *       ^
+     * ```
+     */
+    private val onDependencyKeyValue: PsiElementPattern.Capture<TomlKeyValue> =
+        cargoTomlPsiElement<TomlKeyValue>()
+            .withSuperParent(
+                1,
+                onDependencyTable
+            )
```
Undin: `withParent(onDependencyTable)`

src/main/kotlin/org/rust/toml/CargoTomlPsiPattern.kt
```diff
@@ -0,0 +1,158 @@
...
+     * version = "1"
+     *         ^
+     * ```
+     */
+    private val onSpecificDependencyKeyValue: PsiElementPattern.Capture<TomlKeyValue> =
+        cargoTomlPsiElement<TomlKeyValue>()
+            .withSuperParent(
+                1,
+                onSpecificDependencyTable
+            )
```
Undin: `withParent(onSpecificDependencyTable)`

src/test/kotlin/org/rust/toml/CargoTomlPsiPatternTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    fun `test inSpecificDependencyKeyValue 3`() = testPatternNegative(inSpecificDependencyKeyValue, """
+        [dependencies]
+        version = ""
+        #^
+    """)
+
+    private inline fun <reified T : PsiElement> testPattern(pattern: ElementPattern<T>, @Language("Toml") code: String) {
+        InlineFile(code, "Cargo.toml")
+        val element = findElementInEditor<T>()
+        LightPlatformCodeInsightFixtureTestCase.assertTrue(pattern.accepts(element))
```
Undin: Looks like you can omit `LightPlatformCodeInsightFixtureTestCase` here

src/test/kotlin/org/rust/toml/CargoTomlPsiPatternTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private fun <T> testPatternNegative(pattern: ElementPattern<T>, @Language("Toml") code: String) {
+        InlineFile(code, "Cargo.toml")
+        val element = findElementInEditor<PsiElement>()
+        LightPlatformCodeInsightFixtureTestCase.assertFalse(pattern.accepts(element))
```
Undin: The same here

src/test/kotlin/org/rust/toml/CargoTomlDependenciesCompletionTest.kt
```diff
@@ -0,0 +1,164 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml
+
+import org.intellij.lang.annotations.Language
+
+class CargoTomlDependenciesCompletionTest : CargoTomlCompletionTestBase() {
```
Undin: Now dependency completion works with complex keys like `target.'cfg(windows)'.dependencies`  and `target.'cfg(windows)'.dependencies.dep`.
But we still don't have the corresponding tests. Let's add them.

src/main/kotlin/org/rust/toml/CargoTomlDependenciesCompletionProvider.kt
```diff
@@ -0,0 +1,136 @@
...
+import com.intellij.psi.tree.IElementType
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.util.ProcessingContext
+import org.rust.lang.core.psi.ext.ancestorOrSelf
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.elementType
+import org.rust.lang.core.psi.ext.isAncestorOf
+import org.toml.lang.psi.*
+
+abstract class AbstractTomlKeyValueCompletionProvider : CompletionProvider<CompletionParameters>() {
```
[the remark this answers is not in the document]
vlad20012: Partially resolved: I don't think that the name should contains "Dependency" because this class can be used for any TomlKeyValue completion. I removed any mentions of "Dependency" word from this class

src/test/kotlin/org/rust/toml/CargoTomlDependenciesCompletionTest.kt
vlad20012: These cases are covered by `CargoTomlPsiPatternTest`. Should be double-check it?

Undin: But it's not really true.
Yeah, we indeed have tests for patterns. But are we sure that we correctly use these patterns in other code to make completion works in a necessary way?
In my opinion, pattern tests are convenient way for developers to find out a problem when tests fail. User know nothing about patterns but he wants that completion works. So, first of all, we should test the user use case and other tests are optional/auxiliary

vlad20012: Ok, I've added this test:
```kotlin
fun `test empty key with complex dependency head`() = doTest("""
        [target.'cfg(windows)'.dev-dependencies]
        <caret>
    """, """
        [target.'cfg(windows)'.dev-dependencies]
        dep = "1.0"
    """, "dep" to "1.0")
```

Undin: 👍

PR #2651, opened 2018-07-07
WIP display items expanded from macros in structure view

vlad20012:
image


kumbayo:
Which kind of items do you want to display in the structure view for expanded macros?
Only stuff that we display in modules or also others things like `use` and `extern` to "debug" the macro expansion.

In libc-0.2.40/src/lib.rs this causes `null` to be displayed in expanded `cfg_if!` macros.
image

It seems `psi.expansion.orEmpty()` does not filter by type and `Utils.presentableName()` does not handle certain types like `RsUseItemImpl`, `RsForeignModImpl`.

vlad20012:
@kumbayo
1. I want expanded items to be displayed flatten, i.e. there is a single group `cfg_if` and all resulting items inside it without intermediate macros like `__cfg_if_items`
2. I didn't think about `use` displaying... I think we should not display it. Instead, we can invent an action to display expanded text to debug macros

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -118,3 +125,6 @@ fun breadcrumbName(e: RsElement): String? {
         else -> null
     }
 }
+
+private val GENERATED_ITEM_ATTR: TextAttributesKey =
```
Undin: Let's move it into `RsColors`

PR #2652, opened 2018-07-07
PERF: macro resolve cache invalidation

Undin:
Use `rustStructureModificationTracker` for macro resolve cache invalidaiton.

Also measure average time of references resolution after typing over all functions in file.
It should show more representative result because resolution in single function can depend on not all
cases and such measurement can mislead us


vlad20012:
:+1: Cool, now we have file type inference latency benchmark.
bors r+

PR #2653, opened 2018-07-07
RUN: add UI option to disable using '--all-target' flag while cargo build/check

### LINKED ISSUE #2604: --all-targets not always desirable
When building a project for a specific no-std target (in my case for `thumbv6m-none-eabi`), the argument `--all-targets` casues `cargo check` to fail because there is no crate `test`. The same thing happens when I build from the command line.

Please make it configurable or even just a tickbox or something! I've had to go back to an older version of the plugin before the `--all-targets` change was introduced and am stuck there.

An example log is shown:
```
C:/Users/David/.cargo/bin/cargo.exe build --all --all-targets
   Compiling coffee-compass v0.1.0 (file:///C:/Users/David/Documents/code/coffee-compass)
error[E0463]: can't find crate for `test`
 --> src\main.rs:1:1
  |
1 | #![no_std]
  | ^ can't find crate

error: aborting due to previous error
```

Undin:
Fixes #2604

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/configurable/RustProjectConfigurable.kt
+++ b/src/main/kotlin/org/rust/cargo/project/configurable/RustProjectConfigurable.kt
@@ -32,26 +32,29 @@ class RustProjectConfigurable(
         project.cargoProjects.allProjects.firstOrNull()?.rootDir?.pathAsPath ?: Paths.get(".")
     )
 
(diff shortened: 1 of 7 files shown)


axos88:
Where can I find this option?

axos88:
image

axos88:
File | Settings for Windows and Linux 
CLion | Preferences for macOS

PR #2655, opened 2018-07-08
 TY: Basic implementation of lifetimes

mchernyavsky:
Providing information about lifetime parameters; substituting and rendering for lifetimes.

@vlad20012

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
@@ -11,18 +11,21 @@ import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.*
 import org.rust.lang.core.types.BoundElement
+import org.rust.lang.core.types.Substitution
 import org.rust.lang.core.types.infer.foldTyInferWith
+import org.rust.lang.core.types.infer.resolve
 import org.rust.lang.core.types.infer.substitute
 import org.rust.lang.core.types.inference
+import org.rust.lang.core.types.regions.ReEarlyBound
+import org.rust.lang.core.types.regions.Region
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
 import org.rust.stdext.buildMap
 
-
 class RsPathReferenceImpl(
     element: RsPath
 ) : RsReferenceBase<RsPath>(element),
-    RsPathReference{
+    RsPathReference {
 
     override val RsPath.referenceAnchor: PsiElement get() = referenceNameElement
 
@@ -81,6 +84,8 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
         }
     }
 
+    val lifetimeArguments: List<Region>? = path.typeArgumentList?.lifetimeList?.map { it.resolve() }
+
     val outputArg = path.retType?.typeReference?.type
 
     return result.map { boundElement ->
@@ -111,12 +116,12 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
             }
         }
 
-        val parameters = element.typeParameters.map { TyTypeParameter.named(it) }
-
-        BoundElement(element,
-            subst + parameters.zip(typeArguments ?: parameters).toMap(),
-            assocTypes
-        )
+        val typeParameters = element.typeParameters.map { TyTypeParameter.named(it) }
+        val lifetimeParameters = element.lifetimeParameters.map { ReEarlyBound(it) }
+        val typeSubst = typeParameters.zip(typeArguments ?: typeParameters).toMap()
+        val lifetimeSubst = lifetimeParameters.zip(lifetimeArguments ?: lifetimeParameters).toMap()
+        val newSubst = Substitution(typeSubst, lifetimeSubst)
+        BoundElement(element, subst + newSubst, assocTypes)
     }
 }
 
--- a/src/main/kotlin/org/rust/lang/core/types/Kind.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/Kind.kt
@@ -1,0 +1,23 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types
+
+typealias TypeFlags = Int
+const val HAS_TY_INFER_MASK: Int = 1
+const val HAS_TY_TYPE_PARAMETER_MASK: Int = 2
+const val HAS_TY_PROJECTION_MASK: Int = 4
+const val HAS_RE_EARLY_BOUND_MASK: Int = 8
+
+/**
+ * An entity in the Rust typesystem, which can be one of
+ * several kinds (only types and lifetimes for now).
+ */
+abstract class Kind(val flags: TypeFlags = 0)
+
+fun mergeFlags(kinds: Collection<Kind>): TypeFlags = kinds.fold(0) { a, b -> a or b.flags }
+
+fun mergeFlags(element: BoundElement<*>): TypeFlags =
+    mergeFlags(element.subst.kinds) or mergeFlags(element.assoc.values)
--- a/src/main/kotlin/org/rust/lang/core/types/Substitution.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/Substitution.kt
@@ -1,0 +1,66 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types
+
+import org.rust.lang.core.psi.RsLifetimeParameter
+import org.rust.lang.core.psi.RsTypeParameter
+import org.rust.lang.core.types.infer.TypeFolder
+import org.rust.lang.core.types.infer.substitute
+import org.rust.lang.core.types.regions.ReEarlyBound
+import org.rust.lang.core.types.regions.Region
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyTypeParameter
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.stdext.zipValues
+
+open class Substitution(
+    val typeSubst: Map<TyTypeParameter, Ty> = emptyMap(),
+    val lifetimeSubst: Map<ReEarlyBound, Region> = emptyMap()
+) {
+    val types: Collection<Ty> get() = typeSubst.values
+    val lifetimes: Collection<Region> get() = lifetimeSubst.values
+    val kinds: Collection<Kind> get() = types + lifetimes
+
+    operator fun plus(other: Substitution): Substitution =
+        Substitution(mergeMaps(typeSubst, other.typeSubst), mergeMaps(lifetimeSubst, other.lifetimeSubst))
+
+    operator fun get(key: TyTypeParameter) = typeSubst[key]
+    operator fun get(key: ReEarlyBound) = lifetimeSubst[key]
+    operator fun get(psi: RsTypeParameter) = typeSubst[TyTypeParameter.named((psi))]
+    operator fun get(psi: RsLifetimeParameter) = lifetimeSubst[ReEarlyBound((psi))]
+
+    fun typeByName(name: String): Ty =
+        typeSubst.entries.find { it.key.toString() == name }?.value ?: TyUnknown
+
+    fun typeParameterByName(name: String): TyTypeParameter? =
+        typeSubst.keys.find { it.toString() == name }
+
+    fun substituteInValues(map: Substitution): Substitution =
+        Substitution(
+            typeSubst.mapValues { (_, value) -> value.substitute(map) },
+            lifetimeSubst.mapValues { (_, value) -> value.substitute(map) }
+        )
+
+    fun foldValues(folder: TypeFolder): Substitution =
+        Substitution(
+            typeSubst.mapValues { (_, value) -> value.foldWith(folder) },
+            lifetimeSubst.mapValues { (_, value) -> value.foldWith(folder) }
+        )
+
+    fun zipTypeValues(other: Substitution): List<Pair<Ty, Ty>> = zipValues(typeSubst, other.typeSubst)
+
+    fun mapTypeValues(transform: (Map.Entry<TyTypeParameter, Ty>) -> Ty): Substitution =
+        Substitution(typeSubst.mapValues(transform), lifetimeSubst)
+}
+
+private object EmptySubstitution : Substitution()
+
+val emptySubstitution: Substitution = EmptySubstitution
+
+fun Map<TyTypeParameter, Ty>.toTypeSubst() = Substitution(typeSubst = this)
+
+private fun <K, V> mergeMaps(map1: Map<K, V>, map2: Map<K, V>): Map<K, V> =
+    if (map1.isEmpty() && map2.isEmpty()) emptyMap() else HashMap(map1).apply { putAll(map2) }
--- a/src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt
@@ -8,6 +8,10 @@ package org.rust.lang.core.types.infer
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ref.advancedDeepResolve
+import org.rust.lang.core.types.regions.ReEarlyBound
+import org.rust.lang.core.types.regions.ReStatic
+import org.rust.lang.core.types.regions.ReUnknown
+import org.rust.lang.core.types.regions.Region
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
 
@@ -46,7 +50,7 @@ fun inferTypeReferenceType(ref: RsTypeReference): Ty {
         is RsRefLikeType -> {
             val base = type.typeReference
             when {
-                type.isRef -> TyReference(inferTypeReferenceType(base), type.mutability)
+                type.isRef -> TyReference(inferTypeReferenceType(base), type.mutability, type.lifetime.resolve())
                 type.isPointer -> TyPointer(inferTypeReferenceType(base), type.mutability)
                 else -> TyUnknown
             }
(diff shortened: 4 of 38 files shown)


src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -81,6 +85,14 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
         }
     }
 
+    val lifetimeArguments: List<Re>? =
```
fedochet: You can remove `?` from type if you're returning `orEmpty()`, I guess

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
```diff
@@ -79,21 +69,17 @@ val Ty.hasTyTypeParameters
...
-
-fun Substitution.get(psi: RsTypeParameter): Ty? {
-    return get(TyTypeParameter.named((psi)))
-}
+val Ty.needToSubstitute
+    get(): Boolean = hasTyTypeParameters || hasReLifetimeParameters
 
-fun mergeFlags(element: BoundElement<*>): TypeFlags =
-    element.subst.values.fold(0) { a, b -> a or b.flags } or element.assoc.values.fold(0) { a, b -> a or b.flags }
+fun mergeFlags(element: BoundElement<*>): KindFlags =
```
fedochet: Probably this will be better like this:
```
fun mergeFlags(element: BoundElement<*>): KindFlags =
    element.subst.kinds.mergeFlags() or element.assoc.values.mergeFlags()

fun mergeFlags(kinds: List<Kind>): KindFlags = kinds.mergeFlags()

private fun Collection<Kind>.mergeFlags() = this.fold(0) { a, b -> a or b.flags }
```

This may cause some compilaton troubles, but you get the idea

src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt
```diff
@@ -80,3 +82,17 @@ fun inferTypeReferenceType(ref: RsTypeReference): Ty {
         else -> TyUnknown
     }
 }
+
+internal fun resolveLifetime(element: RsElement): Re {
```
fedochet: This will probably be more readable if rewritten with `when`, like this:
```kotlin
    val lifetime = when (element) {
        is RsRefLikeType -> element.lifetime
        is RsSelfParameter -> element.lifetime
        else -> null
    } ?: return ReUnknown

    val lifetimeName = lifetime.referenceName

    when (lifetimeName) {
        "'_" -> return ReUnderscore
        "'static" -> return ReStatic
    }

    val resolved = lifetime.reference.resolve()
    return when (resolved) {
        is RsLifetimeParameter -> ReEarlyBound.named(resolved)
        else -> ReUnknown
    }
```

Undin: Let's use `if` expr for binary condition instead of `when`
```kotlin
val resolved = lifetime.reference.resolve()
return if (resolved is RsLifetimeParameter) ReEarlyBound.named(resolved) else ReUnknown
```

fedochet: @Undin I dislike this syntax very much :D
IMO, it is badly readable
Maybe spliting branches on separate lines will be better?

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
mchernyavsky: For consistency with `typeArguments` I have removed `orEmpty()`.

mchernyavsky:
Fixed

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -81,6 +85,13 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
...
+        path.typeArgumentList
+            ?.lifetimeList
+            ?.map { it.reference }
+            ?.map { it.resolve() as? RsLifetimeParameter }
+            ?.map { it?.let { ReEarlyBound.named(it) } ?: ReUnknown }
```
Undin: Let's collapse all `map` calls into a single `map` call to avoid unnecessary creation of intermediate list objects

src/main/kotlin/org/rust/lang/core/types/infer/Declarations.kt
Undin: @fedochet I agree that one-line `if` expr is not readable if child expressions are long and in such cases I'd prefer to split it into different lines
But in this case it looks like good enough

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
mchernyavsky: Done it

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -81,6 +85,14 @@ fun resolvePath(path: RsPath, lookup: ImplLookup = ImplLookup.relativeTo(path)):
         }
     }
 
+    val lifetimeArguments: List<Re>? =
+        path.typeArgumentList
+            ?.lifetimeList
+            ?.map { lifetime ->
+                val resolved = lifetime.reference.resolve() as? RsLifetimeParameter
+                resolved?.let { ReEarlyBound.named(it) } ?: ReUnknown
```
Undin: You can write something like this
```kotlin
val resolved = lifetime.reference.resolve() as? RsLifetimeParameter ?: return@map ReUnknown
ReEarlyBound.named(it)
```
to get rid of `let` here - it ususally makes code readability worse

mchernyavsky: Fixed

src/main/kotlin/org/rust/lang/core/types/infer/Fold.kt
```diff
@@ -89,14 +99,27 @@ fun <T> TypeFoldable<T>.foldTyProjectionWith(folder: (TyProjection) -> Ty): T =
  */
 fun <T> TypeFoldable<T>.substitute(subst: Substitution): T =
     foldWith(object : TypeFolder {
-        override fun invoke(ty: Ty): Ty =
-            subst[ty] ?: if (ty.hasTyTypeParameters) ty.superFoldWith(this) else ty
+        override fun foldTy(type: Ty): Ty {
+            val typeWithLifetime = (type as? TyReference)?.copy(lifetime = foldRe(type.lifetime)) ?: type
+            val value = (typeWithLifetime as? TyTypeParameter)?.let { subst[it] }
+            if (value != null) return value
```
vlad20012: Can be simplified to `(typeWithLifetime as? TyTypeParameter)?.let { return subst[it] }`

src/main/kotlin/org/rust/lang/core/types/infer/Fold.kt
```diff
@@ -89,14 +99,27 @@ fun <T> TypeFoldable<T>.foldTyProjectionWith(folder: (TyProjection) -> Ty): T =
...
+            return if (typeWithLifetime.needToSubstitute) {
+                typeWithLifetime.superFoldWith(this)
+            } else {
+                typeWithLifetime
+            }
+        }
+
+        override fun foldRe(region: Re): Re {
+            val value = (region as? ReEarlyBound)?.let { subst[it] }
+            return value ?: region
```
vlad20012: Can be simplified to `return (region as? ReEarlyBound)?.let { subst[it] } ?: region` and then converted to "expression body" function

src/main/kotlin/org/rust/lang/core/types/infer/Fold.kt
```diff
@@ -5,10 +5,20 @@
...
 
+import org.rust.lang.core.types.Substitution
+import org.rust.lang.core.types.re.Re
+import org.rust.lang.core.types.re.ReEarlyBound
 import org.rust.lang.core.types.ty.*
 
-typealias TypeFolder = (Ty) -> Ty
-typealias TypeVisitor = (Ty) -> Boolean
+interface TypeFolder {
+    fun foldTy(type: Ty): Ty = type
```
vlad20012: Why do you use `type` name for Tys? Not only here, in multiple places. `ty` is conventional abbreviation in this project

src/main/kotlin/org/rust/lang/core/types/Substitution.kt
```diff
@@ -0,0 +1,66 @@
...
+    val typeSubst: Map<TyTypeParameter, Ty> = emptyMap(),
+    val lifetimeSubst: Map<ReEarlyBound, Re> = emptyMap()
+) {
+    val types: Collection<Ty> get() = typeSubst.values
+    val lifetimes: Collection<Re> get() = lifetimeSubst.values
+    val kinds: Collection<Kind> get() = types + lifetimes
+
+    operator fun plus(other: Substitution): Substitution =
+        Substitution(
+            typeSubst + other.typeSubst,
```
vlad20012: Since now we have a single place of maps merging, use `HashMap`s here instead of `LinkedHashMap`, produced by kotlin's `+`. It may looks like `HashMap(typeSubst).apply { putAll(other.typeSubst) }`. But really we can write our own map merging function with some optimizations, i.e. if both maps are empty, return `emptyMap()`. I didn't do it previously because of multiple points of merging, but now it's the single place. Mmm, anyway, we can do in a separate PR....

src/main/kotlin/org/rust/lang/core/types/re/Re.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import org.rust.lang.core.types.Kind
+import org.rust.lang.core.types.KindFlags
+import org.rust.lang.core.types.infer.TypeFoldable
+import org.rust.lang.core.types.infer.TypeFolder
+import org.rust.lang.core.types.infer.TypeVisitor
+
+/**
+ * The name `Re` is short for `Region`, inspired by the Rust
+ * compiler.
```
vlad20012: It's a lie - in rustc it's named "Region"

src/main/kotlin/org/rust/lang/core/types/re/ReUnderscore.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types.re
+
+object ReUnderscore : Re() {
```
vlad20012: Do we have this type of region in rustc? If not, I this it's processed on another level (like lowering) and we dont need it on type level too

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
```diff
@@ -79,21 +69,11 @@ val Ty.hasTyTypeParameters
...
-
-fun mergeFlags(element: BoundElement<*>): TypeFlags =
-    element.subst.values.fold(0) { a, b -> a or b.flags } or element.assoc.values.fold(0) { a, b -> a or b.flags }
+val Ty.hasReLifetimeParameters
+    get(): Boolean = BitUtil.isSet(flags, HAS_RE_LIFETIME_PARAMETER_MASK)
 
-fun mergeFlags(tys: List<Ty>): TypeFlags =
-    tys.fold(0) { a, b -> a or b.flags }
+val Ty.needToSubstitute
+    get(): Boolean = hasTyTypeParameters || hasReLifetimeParameters
```
vlad20012: Let's use some bitmagic here: `BitUtil.isSet(flags, HAS_TY_TYPE_PARAMETER_MASK or HAS_RE_LIFETIME_PARAMETER_MASK)`

src/main/kotlin/org/rust/lang/core/types/ty/TyReference.kt
```diff
@@ -7,12 +7,26 @@ package org.rust.lang.core.types.ty
 
 import org.rust.lang.core.types.infer.TypeFolder
 import org.rust.lang.core.types.infer.TypeVisitor
+import org.rust.lang.core.types.re.Re
 
-data class TyReference(val referenced: Ty, val mutability: Mutability) : Ty(referenced.flags) {
+data class TyReference(
+    val referenced: Ty,
+    val mutability: Mutability,
+    val lifetime: Re
```
vlad20012: Maybe it's better to use default parameter value?
`val lifetime: Re = ReUnknown`
And so remove a lot of changes from this PR related to this signature change

src/main/kotlin/org/rust/lang/core/types/ty/TyReference.kt
```diff
@@ -7,12 +7,26 @@ package org.rust.lang.core.types.ty
...
 
-    override fun superFoldWith(folder: TypeFolder): Ty =
-        TyReference(referenced.foldWith(folder), mutability)
+    override fun superFoldWith(folder: TypeFolder): Ty = TyReference(referenced.foldWith(folder), mutability, lifetime)
 
-    override fun superVisitWith(visitor: TypeVisitor): Boolean =
-        referenced.visitWith(visitor)
+    override fun superVisitWith(visitor: TypeVisitor): Boolean = referenced.visitWith(visitor)
+
+    override fun equals(other: Any?): Boolean = when {
```
vlad20012: Could you please add some comments here about why do you exclude lifetime from comparision

src/main/kotlin/org/rust/lang/core/types/ty/TyReference.kt
vlad20012: And let's rename it to `region`

src/main/kotlin/org/rust/lang/core/types/GenericParameter.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types
+
+interface GenericParameter {
```
vlad20012: Is this abstraction now useless?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1020,15 +1022,17 @@ private class RsFnInferenceContext(
...
             }
         }
 
         unifySubst(fnSubst, typeParameters)
 
         val methodType = (callee.element.typeOfValue)
             .substitute(typeParameters)
-            .foldWith(this::normalizeAssociatedTypesIn) as TyFunction
+            .foldWith(object: TypeFolder {
+                override fun foldTy(type: Ty): Ty = normalizeAssociatedTypesIn(type)
```
vlad20012: This code occurs in multiple places. Maybe it's better to extract `AssociatedTypeNormalizer` to a class level?

src/main/kotlin/org/rust/lang/core/types/ty/TyAdt.kt
```diff
@@ -20,26 +26,39 @@ import org.rust.lang.core.types.type
...
     // This method is rarely called (in comparison with folding),
     // so we can implement it in a such inefficient way
     override val typeParameterValues: Substitution
-        get() = item.typeParameters.withIndex().associate { (i, param) ->
-            TyTypeParameter.named(param) to typeArguments.getOrElse(i) { TyUnknown }
+        get() {
+            val typeSubst = item
+                .typeParameters
+                .withIndex()
+                .associate { (i, param) ->
```
vlad20012: I believe it's ok to write it in a single line:
` val typeSubst = item.typeParameters.withIndex().associate { (i, param) ->`

src/main/kotlin/org/rust/lang/core/types/ty/TyTuple.kt
```diff
@@ -7,13 +7,14 @@ package org.rust.lang.core.types.ty
...
+import org.rust.lang.core.types.mergeFlags
 
 data class TyTuple(val types: List<Ty>) : Ty(mergeFlags(types)) {
 
     override fun superFoldWith(folder: TypeFolder): Ty =
         TyTuple(types.map { it.foldWith(folder) })
 
     override fun superVisitWith(visitor: TypeVisitor): Boolean =
-        types.any(visitor)
+        types.any { visitor.visitTy(it) }
```
vlad20012: -> `types.any { it.visitWith(visitor) }`

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
fedochet: I am a little late here, but maybe we should rewrite it like this?
```kotlin
       map { lifetime ->
            val resolved = lifetime.reference.resolve()
            if (resolved is RsLifetimeParameter)
                ReEarlyBound.named(resolved)
            else
                ReUnknown
        }
```

It is much more clean about what kind of decision we are doing in this closure, IMO

mchernyavsky: I prefer to use [Google's convention](https://android.github.io/kotlin-guides/style.html#expressions) about `if`/`else` expressions:
> An `if`/`else` conditional that is used as an expression may omit braces only if the entire expression fits on one line.

src/main/kotlin/org/rust/lang/core/types/Kind.kt
```diff
@@ -0,0 +1,23 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types
+
+typealias KindFlags = Int
```
vlad20012: And I really dislike this name. Let's rename it back to `TypeFlags` anyway

src/main/kotlin/org/rust/lang/core/types/infer/Fold.kt
mchernyavsky: No, because `subst[it]` has type `Ty?`.

vlad20012: :thinking: 
`(typeWithLifetime as? TyTypeParameter)?.let { subst[it] }?.let { return it }`
?

src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
mchernyavsky: > java.lang.AssertionError: Mask must have only one bit set, but got: 1010

vlad20012:
Please rebase to the actual master

src/main/kotlin/org/rust/lang/core/types/ty/TyReference.kt
mchernyavsky: I have added a comment in the code

phase:
:tada: :tada: :tada: :tada:

Undin:
@mchernyavsky Looks like after this changes we have an endless loop somewhere in resolve/type inference because `org.rustSlowTests.RsHighlightingPerformanceTest#test highlighting mysql_async` can't finish (at least on my machine)

mchernyavsky:
@Undin I have fixed it in #2677

PR #2657, opened 2018-07-09
Add mutability categories

ortem:
Add mutability categories support for checking assignments to immutable memory (see [mem_categorization.rs](https://github.com/rust-lang/rust/blob/master/src/librustc/middle/mem_categorization.rs)).

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
@@ -10,7 +10,10 @@ import com.intellij.lang.annotation.Annotator
 import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.psi.PsiElement
 import org.rust.ide.colors.RsColor
-import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsPatBinding
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.RsSelfParameter
+import org.rust.lang.core.psi.RsValueParameter
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.ancestorStrict
 import org.rust.lang.core.psi.ext.mutability
--- a/src/main/kotlin/org/rust/lang/core/types/Extensions.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/Extensions.kt
@@ -13,11 +13,11 @@ import com.intellij.psi.util.CachedValueProvider.Result
 import com.intellij.psi.util.CachedValuesManager
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
-import org.rust.lang.core.types.infer.*
-import org.rust.lang.core.types.ty.Ty
-import org.rust.lang.core.types.ty.TyReference
-import org.rust.lang.core.types.ty.TyTypeParameter
-import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.infer.RsInferenceResult
+import org.rust.lang.core.types.infer.inferTypeReferenceType
+import org.rust.lang.core.types.infer.inferTypesIn
+import org.rust.lang.core.types.infer.mutabilityCategory
+import org.rust.lang.core.types.ty.*
 import org.rust.openapiext.recursionGuard
 
 
@@ -25,10 +25,11 @@ val RsTypeReference.type: Ty
     get() = recursionGuard(this, Computable { inferTypeReferenceType(this) })
         ?: TyUnknown
 
-val RsTypeElement.lifetimeElidable: Boolean get() {
-    val typeOwner = owner.parent
-    return typeOwner !is RsFieldDecl && typeOwner !is RsTupleFieldDecl && typeOwner !is RsTypeAlias
-}
+val RsTypeElement.lifetimeElidable: Boolean
+    get() {
+        val typeOwner = owner.parent
+        return typeOwner !is RsFieldDecl && typeOwner !is RsTupleFieldDecl && typeOwner !is RsTypeAlias
+    }
 
 private val TYPE_INFERENCE_KEY: Key<CachedValue<RsInferenceResult>> = Key.create("TYPE_INFERENCE_KEY")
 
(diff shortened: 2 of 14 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    val isMutable: Boolean get() {
+        return when (this) {
+            MutabilityCategory.Immutable -> false
+            MutabilityCategory.Declared, MutabilityCategory.Inherited -> true
+        }
+    }
+}
+
+val RsExpr.mutabilityCategory: MutabilityCategory? get() {
```
fedochet: I think that this piece of code may be rewritten like this: 
```kotlin
val RsExpr.mutabilityCategory: MutabilityCategory? get() {
    return when (this) {
        is RsUnaryExpr -> this.unaryExprMutability

        is RsDotExpr -> this.dotExpressionMutability

        is RsIndexExpr -> this.containerExpr?.mutabilityCategory?.inherit()

        is RsPathExpr -> this.pathMutability

        is RsParenExpr -> this.expr.mutabilityCategory

        else -> null
    }
}

private val RsUnaryExpr.unaryExprMutability: MutabilityCategory?
    get() {
        val expr = expr ?: return null
        if (mul == null) return null
        val pointer = expr.type.pointerKind ?: return null

        return MutabilityCategory.valueOf(pointer.mutability)
    }

private val Ty.pointerKind: PointerKind?
    get() = when (this) {
        is TyPointer -> UnsafePointer(mutability)
        is TyReference -> BorrowedPointer(mutability)
        else -> null
    }


private val RsDotExpr.dotExpressionMutability: MutabilityCategory?
    get() {
        val type = expr.type
        val isDeref = expr.inference?.adjustments?.get(this)?.any { it.kind == Adjust.DEREF } ?: false

        return if (isDeref && type is TyReference) {
            MutabilityCategory.valueOf(type.mutability)
        } else {
            expr.mutabilityCategory?.inherit()
        }
    }

private val RsPatBinding.patBindingMutability: MutabilityCategory
    get() {
        return if (kind is BindByValue && mutability.isMut) {
            MutabilityCategory.Declared
        } else {
            MutabilityCategory.Immutable
        }
    }

private val RsPathExpr.pathMutability: MutabilityCategory?
    get() {
        val declaration = path.reference.resolve() ?: return null

        // this brings false-negative, because such variable should has Immutable category:
        // let x; x = 1;
        // x = 2; <-- `x` is immutable, but it determined as mutable
        //
        // so it would be replaced with some kind of data-flow analysis
        if (declaration.hasNoAssignment()) return MutabilityCategory.Declared

        return when (declaration) {
            is RsConstant, is RsEnumVariant, is RsStructItem, is RsFunction -> MutabilityCategory.Declared

            is RsPatBinding -> declaration.patBindingMutability

            is RsSelfParameter -> MutabilityCategory.valueOf(declaration.mutability)

            else -> null
        }
    }

private fun RsElement.hasNoAssignment(): Boolean {
    val letExpr = ancestorStrict<RsLetDecl>()
    return letExpr != null && letExpr.eq == null
}
```

It is the same thing; basically I just eliminated different types of returns and moved logic to separate properties
It seems more readable that way IMO

vlad20012: :+1: but I recommend to use a regular functions without receiver instead of properties and leave as a property just `val RsExpr.mutabilityCategory`

fedochet: @vlad20012 if you say so :sweat_smile: 
I am still not sure when to use propertie or functions when it comes to extensions in kotlin

ortem: @fedochet, @vlad20012  Great remark, thank you! Also what do you think about using visitor pattern instead of just functions/properties?

src/main/kotlin/org/rust/lang/core/types/Extensions.kt
```diff
@@ -73,31 +73,8 @@ val RsTraitOrImpl.selfType: Ty get() {
...
-
-            val type = this.type
-            if (type is TyReference) return type.mutability.isMut
-            if (type is TyUnknown) return DEFAULT_MUTABILITY
+val RsExpr.isDereference: Boolean get() = this is RsUnaryExpr && this.mul != null
 
-            if (declaration is RsEnumVariant) return true
-            if (declaration is RsStructItem) return true
-            if (declaration is RsFunction) return true
+private val DEFAULT_MUTABILITY = true
```
mchernyavsky: Kotlin has `const` keyword for compile-time constants:
```kotlin
private const val DEFAULT_MUTABILITY = true
```

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
fedochet: @ortem it depends from how many possible operations on this hierarchy will you actually need
I think that if you will need another one, you may refactor current approach and use visitors

fedochet: You mean this? :grinning: 
```kotlin
    private fun RsExpr.inferType(expected: Ty? = null): Ty {
        if (ctx.isTypeInferred(this)) error("Trying to infer expression type twice")

        val ty = when (this) {
            is RsPathExpr -> inferPathExprType(this)
            is RsStructLiteral -> inferStructLiteralType(this, expected)
            is RsTupleExpr -> inferRsTupleExprType(this, expected)
            is RsParenExpr -> this.expr.inferType(expected)
            is RsUnitExpr -> TyUnit
            is RsCastExpr -> inferCastExprType(this)
            is RsCallExpr -> inferCallExprType(this, expected)
            is RsDotExpr -> inferDotExprType(this, expected)
            is RsLitExpr -> inferLitExprType(this, expected)
            is RsBlockExpr -> this.block.inferType(expected)
            is RsIfExpr -> inferIfExprType(this, expected)
            is RsLoopExpr -> inferLoopExprType(this)
            is RsWhileExpr -> inferWhileExprType(this)
            is RsForExpr -> inferForExprType(this)
            is RsMatchExpr -> inferMatchExprType(this, expected)
            is RsUnaryExpr -> inferUnaryExprType(this, expected)
            is RsBinaryExpr -> inferBinaryExprType(this)
            is RsTryExpr -> inferTryExprType(this)
            is RsArrayExpr -> inferArrayType(this, expected)
            is RsRangeExpr -> inferRangeType(this)
            is RsIndexExpr -> inferIndexExprType(this)
            is RsMacroExpr -> inferMacroExprType(this)
            is RsLambdaExpr -> inferLambdaExprType(this, expected)
            is RsRetExpr -> inferRetExprType(this)
            is RsBreakExpr -> inferBreakExprType(this)
            is RsContExpr -> TyNever
            else -> TyUnknown
        }

        ctx.writeExprTy(this, ty)
        return ty
    }

```

vlad20012: Yep

src/main/kotlin/org/rust/ide/annotator/RsHighlightingMutableAnnotator.kt
```diff
@@ -81,7 +80,7 @@ class RsHighlightingMutableAnnotator : Annotator {
 
 val RsElement.isMut: Boolean
     get() = when (this) {
-        is RsPatBinding -> mutability.isMut || kind.isRefMut || type.let { it is TyReference && it.mutability.isMut }
+        is RsPatBinding -> mutability.isMut || type.let { it is TyReference && it.mutability.isMut }
```
fedochet: Maybe `type is TyReference && type.nmutability.isMut` is better? Or `type` is `var`?

ortem: @fedochet `type` is a property with custom getter, so it's impossible.

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -0,0 +1,158 @@
...
+        } else {
+            processExprUnadjusted(expr)
+        }
+    }
+
+    val adjustments = expr.inference?.adjustments?.get(expr) ?: emptyList()
+    return helper(expr, adjustments)
+}
+
+fun processExprAdjustedWith(expr: RsExpr, previous: () -> Cmt, adjustment: Adjustment): Cmt {
```
vlad20012: Can be private?

PR #2664, opened 2018-07-12
MACRO: fixed expansions of a macro with empty group (cfg_if! case)

### LINKED ISSUE #2601: cfg_if macro expansion
Macros like `cfg_if` don't seem to be expanding, even with macro expansion option enabled.

It's most noticeable in `std::os` and crates like `libc` or other FFI crates that make heavy use of `cfg_if` and similar macros.

vlad20012:
FIxes #2601

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -138,18 +138,19 @@ class MacroExpander(val project: Project) {
     ): Pair<RsMacroCase, MacroSubstitution>? {
         val macroCallBody = createPsiBuilder(call, call.macroBody ?: return null)
         var start = macroCallBody.mark()
-        return def.macroBodyStubbed?.macroCaseList?.asSequence()
-            ?.mapNotNull { case ->
-                val subst = case.pattern.match(macroCallBody)
-                if (subst != null) {
-                    case to subst
-                } else {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -164,10 +168,17 @@ class MacroExpander(val project: Project) {
                     child.macroExpansionContents?.let { contents ->
```
fedochet: I guess this needs some refactorings :sweat_smile:

src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -380,6 +395,45 @@ private fun PsiBuilder.isSameToken(psi: PsiElement): Boolean {
...
+        CachedValueProvider.Result.create(
+            collectAvailableVars(this),
+            PsiModificationTracker.MODIFICATION_COUNT // TODO use modtracker attached to the macro
+        )
+    }
+
+private fun collectAvailableVars(groupDefinition: RsMacroBindingGroup): Set<String> {
+    val vars = groupDefinition.descendantsOfType<RsMacroBinding>().toMutableList()
+
+    fun go(psi: PsiElement): Unit = when (psi) {
```
ortem: Move `when (psi) {` to a new line

PR #2673, opened 2018-07-16
Rust features

Undin:
* Introduce rust compiler features. We need some info about compiler features to highlight code with experimental features. Feature info can be found [here](https://github.com/rust-lang/rust/commits/master/src/libsyntax/feature_gate.rs)
* New index for `#![feature(...)]` attributes. Rustc allow declaring feature attributes in any module and this index helps us to quickly search feature attributes
* Add error highlighting for `crate_visibility_modifier` feature as "proof of concept"
* Create `MockRustcVersion` annotation to set rustc version for test

TODO: add some script to automatically update feature list from rustc source


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -696,6 +693,19 @@ sealed class RsDiagnostic(
...
+
+    class ExperimentalFeature(
+        element: PsiElement,
+        private val presentableFeatureName: String,
+        private val fix: AddFeatureAttributeFix? = null
+    ) : RsDiagnostic(element) {
+        override fun prepare(): PreparedAnnotation = PreparedAnnotation(
+            ERROR,
+            E0658,
+            header = escapeString("$presentableFeatureName is experimental"),
```
vlad20012: We don't really need `escapeString` here?

src/test/kotlin/org/rust/ide/annotator/fixes/ConvertToSizedTypeFixTest.kt
```diff
@@ -5,33 +5,33 @@
...
     """, """
         trait Foo {}
         fn foo(foo: Box<Foo>) {}
     """)
 
-    fun `test convert function return type to Box`() = checkQuickFix("Convert to Box", """
+    fun `test convert function return type to Box`() = checkFixByText("Convert to Box", """
         trait Foo {}
-        fn foo() -> Foo/*caret*/ { unimplemented!() }
+        fn foo() -> <error>Foo/*caret*/</error> { unimplemented!() }
```
vlad20012: It looks a bit weird that now we have to specify both error and caret...

src/test/kotlin/org/rust/lang/MockRustcVersion.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang
+
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class MockRustcVersion(val rustcVersion: String)
```
vlad20012: It would be nice to add some docs that refer to CargoProject.rustcInfo.

src/test/kotlin/org/rust/ide/annotator/RsAnnotationTestBase.kt
```diff
@@ -0,0 +1,186 @@
...
+        fixName: String,
+        @Language("Rust") before: String,
+        @Language("Rust") after: String,
+        checkWarn: Boolean = true,
+        checkInfo: Boolean = false,
+        checkWeakWarn: Boolean = false,
+        testmark: Testmark? = null
+    ) = checkFix(fixName, before, after,
+        configure = this::configureByFileTree,
+        checkBefore = { myFixture.checkHighlighting(checkWarn, checkInfo, checkWeakWarn) },
```
vlad20012: Why are we running `checkHighlighting` for quickFix tests? Is it true that here we running all annotators here?

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
Undin: `presentableFeatureName ` can potentially contain symbols which we need to escape

src/test/kotlin/org/rust/ide/annotator/RsAnnotationTestBase.kt
Undin: > Why are we running checkHighlighting for quickFix tests?

It allows us to check inspection message and result of quick fix in one test case. For example, we use it in auto import tests.

> Is it true that here we running all annotators here?

Yeah. But if you don't need highlighting, we have `checkFixByTextWithoutHighlighting` method

src/test/kotlin/org/rust/ide/annotator/fixes/ConvertToSizedTypeFixTest.kt
Undin: See [this](https://github.com/intellij-rust/intellij-rust/pull/2673#discussion_r204208425) comment

src/test/kotlin/org/rust/ide/annotator/RsAnnotationTestBase.kt
vlad20012: Okaaaay

PR #2674, opened 2018-07-16
Fixed resolve cache invalidation

vlad20012:


src/main/kotlin/org/rust/lang/core/resolve/ref/RsResolveCache.kt
```diff
@@ -49,26 +55,53 @@ class RsResolveCache(messageBus: MessageBus) {
      * Expected resolve results: [PsiElement], [ResolveResult] or a [List]/[Array] of [ResolveResult]
      */
     @Suppress("UNCHECKED_CAST")
-    fun <K: Any, V> resolveWithCaching(key: K, resolver: (K) -> V): V? {
+    fun <K : PsiElement, V> resolveWithCaching(key: K, resolver: (K) -> V): V? {
         ProgressManager.checkCanceled()
-        return cache[key] as V? ?: run {
+        val map = getCacheFor(key)
+        return map[key] as V? ?: run {
```
fedochet: Isn't it's better to use `as? V` instead of `as V?`? If `map` contains object of type  other than `V`, this will cause `ClassCastException`

vlad20012: ClassCastException will be anyway because of type erasure

src/main/kotlin/org/rust/lang/core/psi/ext/RsModificationTrackerOwner.kt
```diff
@@ -28,3 +31,13 @@ interface RsModificationTrackerOwner : RsItemElement {
      */
     fun incModificationCount(element: PsiElement): Boolean
 }
+
+fun findModificationTrackerOwnerFor(element: PsiElement): RsModificationTrackerOwner? {
```
Undin: Maybe convert to extension function and rename to `findModificationTrackerOwner`?

src/main/kotlin/org/rust/lang/core/resolve/ref/RsResolveCache.kt
```diff
@@ -49,26 +55,53 @@ class RsResolveCache(messageBus: MessageBus) {
...
             result
         }
     }
 
+    private fun getCacheFor(element: PsiElement): ConcurrentMap<PsiElement, Any?> {
+        // 1. The global resolve cache invalidates only on rust structure changes.
+        // 2. If some reference element is located inside RsModificationTrackerOwner,
+        //    the global cache will not be invalidated on its change
+        // 3. PSI uses default identity-based equals/hashCode
+        // 4. Intellij do incremental updates of a mutable PSI tree.
```
Undin: do -> does

Undin:
LGTM except minor comments

vlad20012:
bors r=undin

src/main/kotlin/org/rust/lang/core/resolve/ref/RsResolveCache.kt
```diff
@@ -49,26 +55,53 @@ class RsResolveCache(messageBus: MessageBus) {
...
+        //    the global cache will not be invalidated on its change
+        // 3. PSI uses default identity-based equals/hashCode
+        // 4. Intellij does incremental updates of a mutable PSI tree.
+        //
+        // It means that some reference element may be changed and then should be
+        // re-resolved to another target. But if we use the global cache, we will
+        // retrieve its previous resolve result from the cache. So if the reference
+        // element is located inside RsModificationTrackerOwner, we should use a
+        // separate cache for it.
+        val owner = element.findModificationTrackerOwner()
```
fedochet: It may be better to rewrite it like this:

```kotlin

val owner = element.findModificationTrackerOwner() ?: return cache

return CachedValuesManager.getCachedValue(owner, LOCAL_CACHE_KEY) {
    Result.create(
        createWeakMap(),
        owner.project.rustStructureModificationTracker,
        owner.modificationTracker
    )
}
```

vlad20012: :+1: But too late :)

PR #2680, opened 2018-07-17
INSP: Needless lifetimes inspection

mchernyavsky:
Checks for lifetime annotations which can be removed by relying on lifetime elision.

Corresponds to [needless_lifetimes](https://github.com/rust-lang-nursery/rust-clippy/wiki#needless_lifetimes) lint from Rust Clippy (see [lifetimes.rs](https://github.com/rust-lang-nursery/rust-clippy/blob/master/clippy_lints/src/lifetimes.rs)).

Also implements quickfix for removing needless lifetimes.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
@@ -1,0 +1,228 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.util.TextRange
+import org.rust.ide.inspections.ReferenceLifetime.*
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.openapiext.forEachChild
+import org.rust.stdext.chain
+
+/**
+ * Checks for lifetime annotations which can be removed by relying on lifetime elision.
+ * Corresponds to needless_lifetimes lint from Rust Clippy.
+ */
+class RsNeedlessLifetimesInspection : RsLocalInspectionTool() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): RsVisitor = object : RsVisitor() {
+        override fun visitFunction(fn: RsFunction) {
+            if (couldUseElision(fn)) registerProblem(holder, fn)
+        }
+    }
+}
+
+/**
+ * There are two scenarios where elision works:
+ * - no output references, all input references have different LT
+ * - output references, exactly one input reference with same LT
+ * All lifetimes must be unnamed, 'static or defined without bounds on the level of the current item.
+ */
+private fun couldUseElision(fn: RsFunction): Boolean {
+    if (hasWhereLifetimes(fn.whereClause)) return false
+
+    val typeParametersBounds = fn.typeParameters.flatMap { it.bounds }.map { it.bound }
+    if (typeParametersBounds.any { hasNamedReferenceLifetime(it) }) return false
+
+    val (inputLifetimes, outputLifetimes) = collectLifetimesFromFnSignature(fn)
+
+    // no input lifetimes? easy case!
+    if (inputLifetimes.isEmpty()) return false
+
+    if (checkLifetimesUsedInBody(fn.block)) return false
+
+    // check for lifetimes from higher scopes
+    val allowedLifetimes = allowedLifetimesFrom(fn.lifetimeParameters)
+    if (!inputLifetimes.chain(outputLifetimes).all { allowedLifetimes.contains(it) }) return false
+
+    val areInputsDistinct = inputLifetimes.size == inputLifetimes.distinct().size
+
+    // no output lifetimes, check distinctness of input lifetimes
+    if (outputLifetimes.isEmpty()) {
+        // only unnamed and static, ok
+        if (inputLifetimes.none { it is Named }) return false
+        // we have no output reference, so we only need all distinct lifetimes
+        return areInputsDistinct
+    }
+
+    return when {
+        outputLifetimes.distinct().size > 1 -> false
+        inputLifetimes.size == 1 || fn.selfParameter?.isRef == true && areInputsDistinct -> {
+            val input = inputLifetimes.first()
+            val output = outputLifetimes.first()
+            when {
+                input is Named && output is Named && input.name == output.name -> true
+                input is Named && output === Unnamed -> true
+                input === Unnamed && output === Unnamed && inputLifetimes.any { it is Named } -> true
+                else -> false  // different named lifetimes or something static going on
+            }
+        }
+        else -> false
+    }
+}
+
+private class LifetimesCollector : RsVisitor() {
+    val lifetimes = mutableListOf<ReferenceLifetime>()
+
+    override fun visitSelfParameter(selfParameter: RsSelfParameter) {
+        selfParameter.lifetime?.let { visitLifetime(it) }
+        if (selfParameter.isRef && selfParameter.lifetime.isElided) record(null)
+        selfParameter.typeReference?.let { visitTypeReference(it) }
+    }
+
+    override fun visitRefLikeType(refLike: RsRefLikeType) {
+        if (refLike.isRef && refLike.lifetime.isElided) record(null)
+        super.visitRefLikeType(refLike)
+    }
+
+    override fun visitPath(path: RsPath) {
+        collectAnonymousLifetimes(path)
+        super.visitPath(path)
+    }
+
+    override fun visitLifetime(lifetime: RsLifetime) = record(lifetime)
+
+    override fun visitElement(element: RsElement) {
+        if (element is RsItemElement /* ignore nested items */) return
+        element.forEachChild { it.accept(this) }
+    }
+
+    private fun record(lifetime: RsLifetime?) {
+        lifetimes.add(lifetime.name.toReferenceLifetime())
+    }
+
+    private fun collectAnonymousLifetimes(path: RsPath) {
+        if (path.typeArgumentList != null) return
+        val resolved = path.reference.resolve()
+        when (resolved) {
+            is RsStructItem, is RsTraitItem, is RsTypeAlias -> {
+                val declaration = resolved as RsGenericDeclaration
+                val genericsCount = declaration.lifetimeParameters.size + declaration.typeParameters.size
+                repeat(genericsCount) { record(null) }
+            }
+        }
+    }
+}
+
+private class BodyLifetimeChecker : RsVisitor() {
+    var lifetimesUsedInBody: Boolean = false
+        private set
+
+    override fun visitLifetime(lifetime: RsLifetime) {
+        if (lifetime.name is LifetimeName.Parameter) {
+            lifetimesUsedInBody = true
+        }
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+                    val collector = LifetimesCollector()
+                    bound.accept(collector)
+                    if (collector.lifetimes.any { it is ReferenceLifetime.Named }) return
+
+                    val lifetimes = bound.traitRef?.path?.typeArgumentList?.lifetimeList.orEmpty()
+                    if (lifetimes.any { !it.isStatic && !it.isElided }) return
+                    lifetimes
+                }
+            if (couldUseElision(fn, boundsLifetimes)) {
+                holder.registerProblem(
```
ortem: I'd extract it to `private fun registerProblem(holder: ProblemsHolder, fn: RsFunction)`

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+class RsNeedlessLifetimesInspection : RsLocalInspectionTool() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): RsVisitor = object : RsVisitor() {
+        override fun visitFunction(fn: RsFunction) {
+            if (hasWhereLifetimes(fn.whereClause)) return
+
+            val boundsLifetimes = fn.typeParameters
+                .flatMap { it.bounds }
+                .map { it.bound }
+                .flatMap { bound ->
+                    val collector = LifetimesCollector()
```
ortem: I'd extract something like that
```
private fun hasNamedReferenceLifetimes(bound: RsBound): Boolean {
    val collector = LifetimesCollector()
    bound.accept(collector)
    return collector.lifetimes.any { it is ReferenceLifetime.Named }
}
```

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+    val inputLifetimes = inputCollector.lifetimes + lifetimesFromBounds(boundsLifetimes)
+    val outputLifetimes = outputCollector.lifetimes
+
+    if (BodyLifetimeChecker.checkLifetimesUsedInBody(fn.block)) return false
+
+    // check for lifetimes from higher scopes
+    if (inputLifetimes.chain(outputLifetimes).any { !allowedLifetimes.contains(it) }) return false
+
+    // no input lifetimes? easy case!
+    if (inputLifetimes.isEmpty()) return false
```
ortem: I'd move empty checking just below `inputLifetimes` and `outputLifetimes` declarations.

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+        .toCollection(allowedLifetimes)
+    allowedLifetimes.add(ReferenceLifetime.Unnamed)
+    allowedLifetimes.add(ReferenceLifetime.Static)
+    return allowedLifetimes
+}
+
+private fun lifetimesFromBounds(bounds: List<RsLifetime>): List<ReferenceLifetime> =
+    bounds.map { it.name }.filterIsInstance<ReferenceLifetime.Named>()
+
+private fun countUniqueLifetimes(lifetimes: List<ReferenceLifetime>): Int = HashSet(lifetimes).size
```
fedochet: You can use `distinct().size` here, I guess

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+//        }
+//    }
+
+//    override fun visitPath(o: RsPath) {
+//        collectAnonymousLifetimes(o)
+//        // TODO: handle existential types
+//    }
+
+    override fun visitElement(o: RsElement) {
+        if (!abort && o !is RsItemElement /* ignore nested items */) {
```
fedochet: I think it's better to invert this condition like this: 
```kotlin
if (abort || o is RsItemElement) return
o.childrenOfType<RsElement>().forEach { it.accept(this) }
```

It's always easier to read conditions without negations

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+private fun lifetimesFromBounds(bounds: List<RsLifetime>): List<ReferenceLifetime> =
+    bounds.map { it.name }.filterIsInstance<ReferenceLifetime.Named>()
+
+private fun countUniqueLifetimes(lifetimes: List<ReferenceLifetime>): Int = HashSet(lifetimes).size
+
+/**
+ * Are any lifetimes mentioned in the `where` clause? If yes, we don't try to reason about elision.
+ */
+private fun hasWhereLifetimes(whereClause: RsWhereClause?): Boolean {
+    if (whereClause == null) return false
```
fedochet: You can write it like `whereClause ?: return false`, like you did in `checkLifetimesUsedInBody`

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
ortem: @fedochet `distinct()` preserve original ordering and (I guess) has worse performance than `HashSet`, isn't it?

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,240 @@
...
+
+        // if the bounds define new lifetimes, they are fine to occur
+        val boundLifetimeParams = predicate.forLifetimes?.lifetimeParameterList.orEmpty()
+        val allowedLifetimes = allowedLifetimesFrom(boundLifetimeParams)
+        // now walk the bounds
+        val collector = LifetimesCollector()
+        predicate.typeParamBounds?.polyboundList?.map { it.bound }?.forEach { it.accept(collector) }
+        if (collector.abort) return false
+        // and check that all lifetimes are allowed
+        if (collector.lifetimes.any { !allowedLifetimes.contains(it) }) return false
```
fedochet: I guess you can use `none` instead of `any` and remove this negation. I didn't even noticed it, it is really subtle :sweat_smile:

fedochet: Or you can use 
```kotlin
if (!allowedLifetimes.containsAll(collector.lifetimes)) return false
```
It is more concise

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
fedochet: @ortem I don't know really, but I think that they didn't put O(n^2) perfomance in here :sweat_smile:

fedochet: In source code they just use LinkedHashSet instead, so I don't think there would be some noticeable penalty, while (IMO) code is clearer with `.distinct()`

sirgl:
Maybe there should be a fix to remove lifetime? It should be pretty straightforward to implement, but very convenient in use.

mchernyavsky:
Fixed

src/main/kotlin/org/rust/ide/inspections/fixes/ElideLifetimesFix.kt
```diff
@@ -0,0 +1,85 @@
...
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val fn = descriptor.psiElement as? RsFunction ?: return
+        LifetimeRemover().visitFunction(fn)
+    }
+}
+
+private class LifetimeRemover : RsVisitor() {
+    private val boundsLifetimes = mutableListOf<RsLifetimeParameter>()
+
+    override fun visitFunction(o: RsFunction) {
```
ortem: IMO it's better to rename all `o` inside `visit` methods to more understandable names.

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,246 @@
...
+private fun checkLifetimesUsedInBody(body: RsBlock?): Boolean {
+    body ?: return false
+    val checker = BodyLifetimeChecker()
+    body.accept(checker)
+    return checker.lifetimesUsedInBody
+}
+
+private fun LifetimeName.toReferenceLifetime(): ReferenceLifetime =
+    when (this) {
+        is LifetimeName.Parameter -> ReferenceLifetime.Named(name)
```
ortem: Maybe import `org.rust.lang.core.types.LifetimeName.*` and use just `Parameter`, `Static` and so on?

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,260 @@
...
+        // now walk the bounds
+        predicate.typeParamBounds?.polyboundList?.map { it.bound }?.forEach { it.accept(collector) }
+        // and check that all lifetimes are allowed
+        if (!allowedLifetimes.containsAll(collector.lifetimes)) return false
+    }
+    return false
+}
+
+private fun checkLifetimesUsedInBody(body: RsBlock?): Boolean {
+    body ?: return false
```
ortem: `body ?: return false` is expression of type `RsBlock`. IMO it's better to use statement `if (body == null) return false`.

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,260 @@
...
+
+    private fun collectAnonymousLifetimes(path: RsPath) {
+        if (path.typeArgumentList != null) return
+        val resolved = path.reference.resolve()
+        when (resolved) {
+            is RsStructItem,
+            is RsTraitItem,
+            is RsTypeAlias -> {
+                val genericDecl = resolved as RsGenericDeclaration
+                val numGenerics = genericDecl.lifetimeParameters.size + genericDecl.typeParameters.size
```
ortem: Maybe `genericsCount` is a little clearer than `numGenerics`

mchernyavsky:
Fixed

src/main/kotlin/org/rust/ide/inspections/RsNeedlessLifetimesInspection.kt
```diff
@@ -0,0 +1,232 @@
...
+    override fun visitPath(path: RsPath) {
+        collectAnonymousLifetimes(path)
+        super.visitPath(path)
+    }
+
+    override fun visitLifetime(lifetime: RsLifetime) = record(lifetime)
+
+    override fun visitElement(element: RsElement) {
+        if (element is RsItemElement /* ignore nested items */) return
+        element.childrenOfType<RsElement>().forEach { it.accept(this) }
```
vlad20012: We can avoid list allocation here by using `forEachChild` extension method

PR #2683, opened 2018-07-19
RES: make "go to declaration" works with items defined with a macro

vlad20012:
Actually resolve items defined with a macro to the corresponding
macro call

### DIFF
--- a/src/main/kotlin/org/rust/ide/navigation/goto/RsTargetElementEvaluator.kt
+++ b/src/main/kotlin/org/rust/ide/navigation/goto/RsTargetElementEvaluator.kt
@@ -1,0 +1,33 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.navigation.goto
+
+import com.intellij.codeInsight.TargetElementEvaluatorEx2
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiReference
+import org.rust.lang.core.macros.findMacroCallExpandedFrom
+
+class RsTargetElementEvaluator : TargetElementEvaluatorEx2() {
+    /**
+     * Allows to intercept platform calls to [PsiReference.resolve]
+     *
+     * Note that if this method returns null, it means
+     * "use default logic", i.e. call `ref.resolve()`
+     */
+    override fun getElementByReference(ref: PsiReference, flags: Int): PsiElement? = null
+
+    /**
+     * Allows to refine GotoDeclaration target
+     *
+     * Note that if this method returns null, it means
+     * "use default logic", i.e. just use `navElement`
+     *
+     * @param element the resolved element (basically via `element.reference.resolve()`)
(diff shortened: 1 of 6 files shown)


Undin:
Hm, looks like that we should use `TargetElementEvaluatorEx2` and implement the corresponding logic in `getGotoDeclarationTarget` method

src/test/kotlin/org/rust/ide/navigation/goto/RsGotoDeclarationTest.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package org.rust.ide.navigation.goto
+
+import com.intellij.openapi.actionSystem.IdeActions
+import org.intellij.lang.annotations.Language
+import org.rust.lang.RsTestBase
+
+class RsGotoDeclarationTest : RsTestBase() {
+    private fun doTest(@Language("Rust") before: String, @Language("Rust") after: String) = checkByText(before, after) {
```
Undin: I don't insist but maybe move private methods below all test methods to have the same style in all our test classes?
It seems we use such style almost in all test classes

vlad20012:
Resolved

src/main/kotlin/org/rust/ide/navigation/goto/RsTargetElementEvaluator.kt
```diff
@@ -0,0 +1,25 @@
...
+class RsTargetElementEvaluator : TargetElementEvaluatorEx2() {
+    // `true` is a default value if TargetElementEvaluator is not defined
+    override fun includeSelfInGotoImplementation(element: PsiElement): Boolean = true
+
+    // Note that if this method returns null, it means "use default logic",
+    // i.e. call `ref.resolve()`
+    override fun getElementByReference(ref: PsiReference, flags: Int): PsiElement? = null
+
+    override fun getGotoDeclarationTarget(element: PsiElement, navElement: PsiElement?): PsiElement? =
+        (element as? ExpansionResult)?.expandedFromRecursively?.referenceNameElement ?: element
```
Undin: I think it's better to return null if element is not generated by macro

src/main/kotlin/org/rust/ide/navigation/goto/RsTargetElementEvaluator.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package org.rust.ide.navigation.goto
+
+import com.intellij.codeInsight.TargetElementEvaluatorEx2
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiReference
+import org.rust.lang.core.macros.ExpansionResult
+import org.rust.lang.core.macros.expandedFromRecursively
+
+/** This handler allows to intercept platform calls to [PsiReference.resolve] */
```
Undin: Really this class doesn't intercept `PsiReference.resolve`, it only returns custom element for "goto declaration" action sometimes. Let's change this doc according to real purpose of the class

PR #2686, opened 2018-07-19
INSP: Assign to immutable inspection

ortem:
Add `RsAssignToImmutableInspection` corresponds to [E0594](https://doc.rust-lang.org/error-index.html#E0594) Rust error.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
@@ -1,0 +1,48 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemsHolder
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.isAssignBinaryExpr
+import org.rust.lang.core.psi.ext.unwrapParenExprs
+import org.rust.lang.core.types.isDereference
+import org.rust.lang.core.types.isMutable
+import org.rust.lang.core.types.ty.TyPointer
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.type
+
+class RsAssignToImmutableInspection : RsLocalInspectionTool() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitBinaryExpr(expr: RsBinaryExpr) {
+                if (expr.isAssignBinaryExpr) checkAssignment(expr, holder)
+            }
+        }
+
+    private fun checkAssignment(expr: RsBinaryExpr, holder: ProblemsHolder) {
+        val left = unwrapParenExprs(expr.left)
+        if (left.isMutable) return
+
+        when (left) {
+            is RsDotExpr -> registerProblem(holder, expr, "field of immutable binding")
+            is RsIndexExpr -> registerProblem(holder, expr, "indexed content of immutable binding")
+            is RsUnaryExpr -> if (left.isDereference) registerDereferenceProblem(left, holder, expr)
+        }
+    }
+
+    private fun registerDereferenceProblem(left: RsUnaryExpr, holder: ProblemsHolder, expr: RsBinaryExpr) {
+        when (left.expr?.type) {
+            is TyReference -> registerProblem(holder, expr, "immutable borrowed content")
+            is TyPointer -> registerProblem(holder, expr, "immutable dereference of raw pointer")
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
```diff
@@ -0,0 +1,51 @@
...
+                }
+        }
+    }
+
+    private fun registerProblem(holder: ProblemsHolder, expr: RsExpr, message: String) {
+        holder.registerProblem(expr, "Cannot assign to $message [E0594]")
+    }
+}
+
+private fun RsExpr?.isAssignBinaryExpr(): Boolean {
```
fedochet: Why there is a `RsExpr?`, not just `RsExpr`?

src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
```diff
@@ -0,0 +1,51 @@
...
+    val op = this as? RsBinaryExpr ?: return false
```
fedochet: Maybe `if (this !is RsBinaryExpr) return false`? Don't know what is more readable though

fedochet: Maybe it is better to return `this is BinaryExpr && this.operatorType is AssignmentOp`?

src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    private fun checkAssignment(expr: RsBinaryExpr, holder: ProblemsHolder) {
+        val left = unwrapParenExprs(expr.left)
+        if (left.isMutable) return
+
+        when (left) {
+            is RsDotExpr -> registerProblem(holder, expr, "field of immutable binding")
+            is RsIndexExpr -> registerProblem(holder, expr, "indexed content of immutable binding")
+            is RsUnaryExpr ->
+                if (left.isDereference) when (left.expr?.type) {
```
fedochet: This looks odd; consider wrapping this `when` in braces

src/main/kotlin/org/rust/lang/core/types/Extensions.kt
```diff
@@ -80,5 +80,16 @@ fun Ty.builtinDeref(explicit: Boolean = true): Pair<Ty, Mutability>? =
 
 val RsExpr.isDereference: Boolean get() = this is RsUnaryExpr && this.mul != null
 
-
 val RsExpr.isMutable: Boolean get() = mutabilityCategory?.isMutable ?: Mutability.DEFAULT_MUTABILITY.isMut
+
+tailrec fun unwrapParenExprs(expr: RsExpr): RsExpr =
+    when (expr) {
```
fedochet: I don't like `if expr`s myself, but there are only two options
We should decide something for such cases

src/main/kotlin/org/rust/lang/core/types/Extensions.kt
```diff
@@ -80,5 +80,16 @@ fun Ty.builtinDeref(explicit: Boolean = true): Pair<Ty, Mutability>? =
...
+        is RsParenExpr -> unwrapParenExprs(expr.expr)
+        else -> expr
+    }
+
+val RsExpr.isAssignBinaryExpr: Boolean
+    get() {
+        val op = this as? RsBinaryExpr ?: return false
+        return op.operatorType is AssignmentOp
```
fedochet: How about `return op is RsBinaryExpr && op.operatorType is AssignmentOp`?

src/main/kotlin/org/rust/lang/core/types/Extensions.kt
```diff
@@ -78,7 +78,12 @@ fun Ty.builtinDeref(explicit: Boolean = true): Pair<Ty, Mutability>? =
...
-
+val RsUnaryExpr.isDereference: Boolean get() = this.mul != null
 
 val RsExpr.isMutable: Boolean get() = mutabilityCategory?.isMutable ?: Mutability.DEFAULT_MUTABILITY.isMut
+
+tailrec fun unwrapParenExprs(expr: RsExpr): RsExpr =
+    if (expr is RsParenExpr) unwrapParenExprs(expr.expr) else expr
+
+val RsExpr.isAssignBinaryExpr: Boolean
+    get() = this is RsBinaryExpr && this.operatorType is AssignmentOp
```
vlad20012: Let's move these functions to `RsExpr.kt`.
`org.rust.lang.core.types.Extensions.kt` is for extending PSI with types. All simple PSI extensions should be located inside `org.rust.lang.core.psi.ext` package

Undin:
Interesting
https://doc.rust-lang.org/error-index.html doesn't contain E0594 error although compiler produces it

PR #2687, opened 2018-07-19
Control flow graph implementation

ortem:
Rust control flow graph implementation. Will be used for future data flow analysis and borrow checker.


sirgl:
It would be nice to have tests, that covers Rust control flow construction. You can see example for java [here](https://github.com/JetBrains/intellij-community/blob/d1a470cbe2e416099f776d1f4c6fb61f54582ea8/java/java-tests/testSrc/com/intellij/java/codeInsight/psi/ControlFlowTest.java) and test data [here](https://github.com/JetBrains/intellij-community/tree/master/java/java-tests/testData/psi/controlFlow)
Probably you have to linearise cfg to do it, but I think correctness definitely worth it.
Also, as I understand, you want to attach some attributes to cfg in future, so debugging borrow checker and analysis will be simpler with good testing infrastructure.

ortem:
@sirgl Good idea! What do you think about tests based on depth-first traversal traces?

src/main/kotlin/org/rust/lang/core/Graph.kt
```diff
@@ -0,0 +1,117 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core
+
+import org.rust.stdext.nextOrNull
+import java.util.*
+
+class Graph<N, E>(val nodes: MutableList<Node<N>>, val edges: MutableList<Edge<E>>) {
+    constructor() : this(mutableListOf(), mutableListOf())
```
mchernyavsky: It's better to use default arguments.

src/main/kotlin/org/rust/lang/core/CFGBuilder.kt
```diff
@@ -0,0 +1,366 @@
...
+    fun process(element: RsElement?, pred: CFGIndex): CFGIndex {
+        if (element == null) return pred
+
+        val oldPredsSize = preds.size
+        preds.push(pred)
+        element.accept(this)
+        preds.pop()
+        assert(preds.size == oldPredsSize)
+
+        return result!!
```
mchernyavsky: Maybe It's better to use `checkNotNull`:
```kotlin
return checkNotNull(result) { "Optional (lazy) error message" }
```
This method throws `IllegalStateException` instead of `NullPointerException`.

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    companion object {
+        fun process(fn: RsFunction, sink: (ExitPoint) -> Unit) = fn.block?.acceptChildren(ExitPointVisitor(sink))
+        fun process(fn: RsLambdaExpr, sink: (ExitPoint) -> Unit) = fn.expr?.acceptChildren(ExitPointVisitor(sink))
+    }
+}
+
+private class ExitPointVisitor(
+    private val sink: (ExitPoint) -> Unit
+) : RsVisitor() {
```
mchernyavsky: > IMO it's better to rename all `o` inside `visit` methods to more understandable names.

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }
+    }
+
+    override fun visitExprStmt(exprStmt: RsExprStmt) {
+        exprStmt.acceptChildren(this)
+        val block = exprStmt.parent as? RsBlock ?: return
+        if (!(block.expr == null && block.stmtList.lastOrNull() == exprStmt)) return
+        val parent = block.parent
+        if ((parent is RsFunction || parent is RsExpr && parent.isInTailPosition) && exprStmt.expr.type != TyNever) {
```
fedochet: This condition seems heavy; maybe we should give it some name?

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,188 @@
...
+            else -> {
+                if (expr.isInTailPosition) sink(ExitPoint.TailExpr(expr)) else expr.acceptChildren(this)
+            }
+        }
+    }
+
+    override fun visitExprStmt(exprStmt: RsExprStmt) {
+        exprStmt.acceptChildren(this)
+        val block = exprStmt.parent as? RsBlock ?: return
+        if (!(block.expr == null && block.stmtList.lastOrNull() == exprStmt)) return
```
fedochet: Probably this will be better looking like this:
```kotlin
if (block.expr != null) return        
if (block.stmtList.lastOrNull() != exprStmt) return
```

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,188 @@
...
+
+sealed class ExitPoint {
+    class Return(val e: RsRetExpr) : ExitPoint()
+    class TryExpr(val e: RsExpr) : ExitPoint() // `?` or `try!`
+    class DivergingExpr(val e: RsExpr) : ExitPoint()
+    class TailExpr(val e: RsExpr) : ExitPoint()
+    class TailStatement(val stmt: RsExprStmt) : ExitPoint()
+
+    companion object {
+        fun process(fn: RsFunction, sink: (ExitPoint) -> Unit) = fn.block?.acceptChildren(ExitPointVisitor(sink))
```
fedochet: I think you should use ordinary body functions instead of this. You don't really need `process` to return `Unit?`

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    fun depthFirstTraversalTrace(): String =
+        graph.depthFirstTraversal(this.entry).map { it.data.text }.joinToString("\n")
+
+
+    /**
+     * Creates graph description written in the DOT language.
+     * Usage: copy the output into `cfg.dot` file and run `dot -Tpng cfg.dot -o cfg.png`
+     */
+    fun createDotDescription(): String {
```
fedochet: Maybe use `buildString`: 

```kotlin
    /**
     * Creates graph description written in the DOT language.
     * Usage: copy the output into `cfg.dot` file and run `dot -Tpng cfg.dot -o cfg.png`
     */
    fun createDotDescription(): String =
        buildString {
            append("digraph {\n")

            graph.forEachEdge { edge ->
                val source = edge.source
                val target = edge.target
                val sourceNode = source.data
                val targetNode = target.data

                append("    \"${source.index}: ${sourceNode.text}\" -> \"${target.index}: ${targetNode.text}\";\n")
            }

            append("}\n")
        }
```

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,199 @@
...
+            is RsMatchExpr -> expr.acceptChildren(this)
+            else -> {
+                if (expr.isInTailPosition) sink(ExitPoint.TailExpr(expr)) else expr.acceptChildren(this)
+            }
+        }
+    }
+
+    override fun visitExprStmt(exprStmt: RsExprStmt) {
+        exprStmt.acceptChildren(this)
+        val block = exprStmt.parent as? RsBlock ?: return
```
fedochet: Maybe you can put those computations in some separate method with descriptive name

src/main/kotlin/org/rust/lang/core/CFGBuilder.kt
```diff
@@ -0,0 +1,345 @@
...
+
+    override fun visitLabelDecl(labelDecl: RsLabelDecl) = finishWith(pred)
+
+    override fun visitExprStmt(exprStmt: RsExprStmt) {
+        val exit = process(exprStmt.expr, pred)
+        finishWith { addAstNode(exprStmt, listOf(exit)) }
+    }
+
+    override fun visitPatBinding(patBinding: RsPatBinding) =
+        finishWith { addAstNode(patBinding, listOf(pred)) }
```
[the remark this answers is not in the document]
ortem: Nice! What do you think of
```
private fun finishWithAstNode(element: RsElement, pred: CFGNode) =
    finishWith { addAstNode(element, listOf(pred)) }
```
```
override fun visitPatBinding(patBinding: RsPatBinding) =
    finishWithAstNode(patBinding, pred)
```
?

src/test/kotlin/org/rust/lang/core/RsControlFlowGraphTest.kt
```diff
@@ -0,0 +1,260 @@
...
+
+package org.rust.lang.core
+
+import junit.framework.ComparisonFailure
+import org.intellij.lang.annotations.Language
+import org.rust.lang.RsTestBase
+import org.rust.lang.core.psi.RsBlock
+import org.rust.lang.core.psi.ext.descendantsOfType
+
+class RsControlFlowGraphTest : RsTestBase() {
```
vlad20012: There are very few tests. Especially, there are no tests for loop expression and loop expression with breaks. I.e. loop expr without breaks is really an infinite loop and should be the last node in CFG.

Try to run these tests with coverage and ensure all methods in CFGBuilder are executed

src/main/kotlin/org/rust/lang/core/CFGBuilder.kt
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    override fun visitForExpr(forExpr: RsForExpr) {
+        val loopBack = addDummyNode(listOf(pred))
+        val exprExit = addAstNode(forExpr, emptyList())
+        val loopScope = LoopScope(forExpr, loopBack, exprExit)
+
+        withLoopScope(loopScope) {
+            val bodyExit = process(forExpr.block, loopBack)
+            addContainedEdge(bodyExit, loopBack)
```
vlad20012: Looks like we processing for expression just like it is loop expression, that is incorrect.

src/main/kotlin/org/rust/lang/core/CFGBuilder.kt
```diff
@@ -0,0 +1,348 @@
...
+
+        finishWith(exprExit)
+    }
+
+    override fun visitParenExpr(parenExpr: RsParenExpr) = parenExpr.expr.accept(this)
+
+    override fun visitTryExpr(tryExpr: RsTryExpr) {
+        val exprExit = addAstNode(tryExpr, emptyList())
+        val checkExpr = addDummyNode(listOf(pred))
+        val expr = addAstNode(tryExpr.expr, listOf(checkExpr))
```
vlad20012: Looks like we have to process that expr too? Also, add some tests for try expr

src/main/kotlin/org/rust/lang/core/CFGBuilder.kt
```diff
@@ -0,0 +1,348 @@
...
+    fun process(element: RsElement?, pred: CFGNode): CFGNode {
+        if (element == null) return pred
+
+        val oldPredsSize = preds.size
+        preds.push(pred)
+        element.accept(this)
+        preds.pop()
+        assert(preds.size == oldPredsSize)
+
+        return checkNotNull(result) { "Processing ended inconclusively" }
```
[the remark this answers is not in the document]
ortem: Now `result` sets to `null` before `element.accept(this)`.

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
```diff
@@ -0,0 +1,199 @@
...
+            is RsExprStmt -> expr.cfgText() + ";"
+            else -> this.text
+        }
+}
+
+class CFGEdgeData(val exitingScopes: List<RsElement>)
+
+typealias CFGGraph = Graph<CFGNodeData, CFGEdgeData>
+typealias CFGNode = Node<CFGNodeData, CFGEdgeData>
+typealias CFGEdge = Edge<CFGNodeData, CFGEdgeData>
```
vlad20012: CFGGraph & CFGEdge are not used anywhere

src/main/kotlin/org/rust/lang/core/CFGBuilder.kt
```diff
@@ -0,0 +1,363 @@
...
+
+    override fun visitLabelDecl(labelDecl: RsLabelDecl) = finishWith(pred)
+
+    override fun visitExprStmt(exprStmt: RsExprStmt) {
+        val exit = process(exprStmt.expr, pred)
+        finishWithAstNode(exprStmt, exit)
+    }
+
+    override fun visitPatBinding(patBinding: RsPatBinding) =
+        finishWithAstNode(patBinding, pred)
```
vlad20012: Looks like this method is never executed. We process RsPatBinding in `visitPatIdent` method (b/c RsPatBinding is always a child of RsPatIdent)

vlad20012: Not really. It is used in the case of `RsPatField`. But it's enough weird. Maybe it's better to override `visitPatField`? and also please add a test for the "PatField" case.

src/main/kotlin/org/rust/lang/core/ControlFlowGraph.kt
ortem: Got rid of `CFGGraph`. `CFGEdge` will be used in data-flow analysis.

mchernyavsky:
bors retry

vlad20012:
bors retry

Undin:
bors r=vlad20012

Undin:
bors r=vlad20012

vlad20012:
bors retry

PR #2684, opened 2018-07-19
WIP PERF: better cache macro expansions

vlad20012:

PR #2692, opened 2018-07-20
MACRO&PERF: Store macro expansions as stubs

vlad20012:
I want to store results of macro expansion in memory as stubs instead of AST. This will **significantly** reduce memory consumption.


vlad20012:
Merge it only after #2683 and #2684

vlad20012:
Will be implemented in another way

vlad20012:
Will be implemented in another way

PR #2698, opened 2018-07-22
One branch

Undin:
Some time ago we created a separate branch for 181 IDEA version to be able to build the plugin with 173 and 181 platform versions because of different API in platforms.
Currently, we have master and 182 branches for the same purpose.

These changes move all source code into one branch but introduce separate folders for each supported platform version (now 181 and 182) where we should keep platform dependent implementations.

To change current platform version, you need to change `platformVersion` property in `gradle.properties` and re-import gradle project

TODO:
* fix `commitNightly` in `build.gradle.kts`. I'll do it in other PR


vlad20012:
:tada: :tada: :tada: :tada: :tada: 
bors r+

PR #2699, opened 2018-07-23
ANN: Introduce method separators

### LINKED ISSUE #2602: Implement method separators
I agree with the first paragraph of https://dzone.com/articles/intellij-idea-small-tips-big , method separators improve readability. I'm used to them from the Scala plugin.

t-kameyama:
Fixes #2602

### DIFF
--- a/src/main/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProvider.kt
+++ b/src/main/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProvider.kt
@@ -1,0 +1,51 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.lineMarkers
+
+import com.intellij.codeHighlighting.Pass
+import com.intellij.codeInsight.daemon.DaemonCodeAnalyzerSettings
+import com.intellij.codeInsight.daemon.LineMarkerInfo
+import com.intellij.codeInsight.daemon.LineMarkerProvider
+import com.intellij.openapi.editor.colors.CodeInsightColors
+import com.intellij.openapi.editor.colors.EditorColorsManager
+import com.intellij.openapi.editor.markup.GutterIconRenderer
+import com.intellij.openapi.editor.markup.SeparatorPlacement
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.getPrevNonCommentSibling
+
+class RsMethodLineSeparatorProvider : LineMarkerProvider {
+
+    override fun getLineMarkerInfo(element: PsiElement): LineMarkerInfo<PsiElement>? {
+        if (DaemonCodeAnalyzerSettings.getInstance().SHOW_METHOD_SEPARATORS) {
+            if (element.canHaveSeparator()) {
+                val prevSibling = element.getPrevNonCommentSibling()
+                if (prevSibling.canHaveSeparator() && (element.wantsSeparator() || prevSibling.wantsSeparator())) {
+                    return createLineSeparatorByElement(element)
+                }
+            }
+        }
+        return null
+    }
+
+    private fun PsiElement?.canHaveSeparator() = this is RsFunction
+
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.getPrevNonCommentSibling
+
+class RsMethodLineSeparatorProvider : LineMarkerProvider {
+
+    override fun getLineMarkerInfo(element: PsiElement): LineMarkerInfo<PsiElement>? {
+        if (DaemonCodeAnalyzerSettings.getInstance().SHOW_METHOD_SEPARATORS) {
```
Undin: Let's save `DaemonCodeAnalyzerSettings` and `EditorColorsManager` instances in constructor as `JavaLineMarkerProvider` does

src/main/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    private fun createLineSeparatorByElement(element: PsiElement): LineMarkerInfo<PsiElement> {
+        val anchor = PsiTreeUtil.getDeepestFirst(element)
+        return LineMarkerInfo(anchor, anchor.textRange, null, Pass.LINE_MARKERS, null, null, GutterIconRenderer.Alignment.RIGHT).apply {
+            separatorColor = EditorColorsManager.getInstance().globalScheme.getColor(CodeInsightColors.METHOD_SEPARATORS_COLOR)
+            separatorPlacement = SeparatorPlacement.TOP
+        }
+    }
+
+    override fun collectSlowLineMarkers(p0: MutableList<PsiElement>, p1: MutableCollection<LineMarkerInfo<PsiElement>>) {
```
Undin:
```kotlin
fun collectSlowLineMarkers(p0: MutableList<PsiElement>, p1: MutableCollection<LineMarkerInfo<PsiElement>>)
``` 
-> 
```kotlin
fun collectSlowLineMarkers(
    elements: List<PsiElement>, 
    result: MutableCollection<LineMarkerInfo<PsiElement>>
)
```

src/test/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProviderTest.kt
```diff
@@ -0,0 +1,52 @@
...
+                const C1: i32 = 1;
+                fn f1() {}
+                fn f2() {}
+                fn f3() { // - null
+                }
+                fn f4() { // - null
+                }
+                fn f5() {} // - null
+            }
+        """)
```
Undin: Let's add tests for functions outside of impl block: top level function, trait methods, functions in extern block

src/main/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+                    return createLineSeparatorByElement(element)
+                }
+            }
+        }
+        return null
+    }
+
+    private fun PsiElement?.canHaveSeparator() = this is RsFunction
+
+    private fun PsiElement?.wantsSeparator() = if (this == null) false else StringUtil.getLineBreakCount(text) > 0
```
Undin: Are we sure that we don't want line separator for one line functions?
For example, kotlin doesn't show separators only for one-line single-expression functions like this
```kotlin
fun foo(x): Int = x * x
```
 But rust doesn't have such syntax at all, so maybe drop this condition?

t-kameyama: I think this condition is necessary for this case:
```rust
trait Foo {
    fn f1();
    fn f2();
    fn f3();
}
```

t-kameyama:
Fixed

Undin:
bors r+
Thanks!

PR #2700, opened 2018-07-23
Fix memory categorization of rvalue expressions

### LINKED ISSUE #2696: "Cannot borrow immutable as mutable" when the code is correct
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2372-181-nightly
* **IDE name and version:** IntelliJ IDEA 2018.1.6 (Ultimate Edition), Build #IU-181.5540.7
* **Operating system:** Linux

## Problem description
The IDE reports a "Cannot borrow immutable as mutable" error (indicated by a red underline) when there are no real errors in the code.

## Steps to reproduce

Create a new binary project with the following code:
```rust
fn main() {
  let x = vec![1];
  if x.iter().any(|c| *c == 2) {

  }
}
```
The IDE highlights `x.iter()`. The popup message says:

> Cannot borrow immutable local variable `x.iter()` as mutable less... (Ctrl+F1) 
Implements simple cases of borrow checking.

ortem:
Fixes #2696

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
@@ -61,6 +61,9 @@ private fun processUnaryExpr(unaryExpr: RsUnaryExpr): Cmt {
 }
 
 private fun processDotExpr(dotExpr: RsDotExpr): Cmt {
+    if (dotExpr.methodCall != null) {
+        return processRvalue(dotExpr)
+    }
     val type = dotExpr.type
     val base = dotExpr.expr
     val baseCmt = processExpr(base)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -123,14 +126,22 @@ private fun processDeref(baseCmt: Cmt): Cmt {
     }
 }
 
+private fun processRvalue(expr: RsExpr): Cmt =
+    Cmt(expr.type, MutabilityCategory.Declared)
+
+private fun isRvalue(expr: RsExpr): Boolean =
```
Undin: Maybe create some marker interface for this expression types to use `expr is RsSomeInterfaceName` instead of this function?

ortem: Actually this code is redundant.

PR #2705, opened 2018-07-24
Implement categorization of expressions

ortem:
Add `category` to `Cmt` (category, mutability, and type); implement memory categorization cache.
Some parts of memory categorization are unused now, but they will be necessary for borrow checking.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
@@ -6,18 +6,117 @@
 package org.rust.lang.core.types.infer
 
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.containerExpr
+import org.rust.lang.core.psi.ext.descendantsOfType
 import org.rust.lang.core.psi.ext.mutability
+import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.builtinDeref
-import org.rust.lang.core.types.infer.Adjustment.Deref
-import org.rust.lang.core.types.inference
+import org.rust.lang.core.types.infer.Aliasability.FreelyAliasable
+import org.rust.lang.core.types.infer.Aliasability.NonAliasable
+import org.rust.lang.core.types.infer.AliasableReason.*
+import org.rust.lang.core.types.infer.BorrowKind.ImmutableBorrow
+import org.rust.lang.core.types.infer.BorrowKind.MutableBorrow
+import org.rust.lang.core.types.infer.Categorization.*
+import org.rust.lang.core.types.infer.ImmutabilityBlame.*
+import org.rust.lang.core.types.infer.InteriorKind.*
+import org.rust.lang.core.types.infer.MutabilityCategory.Declared
+import org.rust.lang.core.types.infer.PointerKind.BorrowedPointer
+import org.rust.lang.core.types.infer.PointerKind.UnsafePointer
 import org.rust.lang.core.types.isDereference
+import org.rust.lang.core.types.regions.ReStatic
+import org.rust.lang.core.types.regions.Region
 import org.rust.lang.core.types.ty.*
-import org.rust.lang.core.types.type
 import org.rust.stdext.nextOrNull
 
+/** [Categorization] is a subset of the full expression forms */
+sealed class Categorization {
+    /** Temporary value */
+    data class Rvalue(val region: Region) : Categorization()
 
+    /** Static value */
+    object StaticItem : Categorization()
+
+    /** Local variable */
+    data class Local(val element: RsElement) : Categorization()
+
+    /** Dereference of a pointer */
+    data class Deref(val cmt: Cmt, val pointerKind: PointerKind) : Categorization()
+
+    /** Something reachable from the base without a pointer dereference (e.g. field) */
+    data class Interior(val cmt: Cmt, val interiorKind: InteriorKind) : Categorization()
+
+    /** Selects a particular enum variant (if enum has more than one variant */
+    data class Downcast(val cmt: Cmt, val element: RsElement) : Categorization()
+}
+
+sealed class BorrowKind {
+    object ImmutableBorrow : BorrowKind()
+    object MutableBorrow : BorrowKind()
+
+    companion object {
+        fun from(mutability: Mutability): BorrowKind =
+            when (mutability) {
+                Mutability.IMMUTABLE -> ImmutableBorrow
+                Mutability.MUTABLE -> MutableBorrow
+            }
+
+        fun isCompatible(firstKind: BorrowKind, secondKind: BorrowKind): Boolean =
+            firstKind == ImmutableBorrow && secondKind == ImmutableBorrow
+    }
+}
+
+sealed class PointerKind {
+    data class BorrowedPointer(val borrowKind: BorrowKind, val region: Region) : PointerKind()
+    data class UnsafePointer(val mutability: Mutability) : PointerKind()
+}
+
+/** "interior" means "something reachable from the base without a pointer dereference" */
+sealed class InteriorKind {
+    /** e.g. `s.field` */
+    class InteriorField(val fieldName: String?) : InteriorKind()
+
+    /** e.g. `arr[0]` */
+    object InteriorIndex : InteriorKind()
+
+    /** e.g. `fn foo([_, a, _, _]: [A; 4]) { ... }` */
+    object InteriorPattern : InteriorKind()
+}
+
+/** Reason why something is immutable */
+sealed class ImmutabilityBlame {
+    /** Immutable as immutable variable */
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,57 +5,167 @@
...
+/** [Categorization] is a subset of the full expression forms */
+sealed class Categorization {
+    /** Temporary value */
+    data class Rvalue(val region: Region) : Categorization()
+
+    /** Static value */
+    object StaticItem : Categorization()
+
+    /** Variable captured by closure */
+    object Upvar : Categorization()
```
mchernyavsky: Object?

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,57 +5,167 @@
...
+
+    /** Local variable */
+    data class Local(val element: RsElement) : Categorization()
+
+    /** Dereference of a pointer */
+    data class Deref(val cmt: Cmt, val pointerKind: PointerKind?) : Categorization()
```
mchernyavsky: Why is `PointerKind` nullable?

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,57 +5,167 @@
...
+
+    /** Something reachable from the base without a pointer dereference (e.g. field) */
+    data class Interior(val cmt: Cmt, val interiorKind: InteriorKind) : Categorization()
+
+    /** Selects a particular enum variant (if enum has more than one variant */
+    data class Downcast(val cmt: Cmt, val element: RsElement) : Categorization()
+}
+
+sealed class BorrowKind {
```
mchernyavsky: Maybe `enum` is better suited in this case?

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,57 +5,167 @@
...
-            MutabilityCategory.Declared, MutabilityCategory.Inherited -> true
+            Immutable -> false
+            Declared, Inherited -> true
         }
 }
 
 /**
- * Category, MutabilityCategory, and Type
- * Currently supports only MutabilityCategory and Type
+ * [Cmt]: Category, MutabilityCategory, and Type
```
mchernyavsky: 🤔

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
ortem: In rustc, `Upvar` contains `ClosureKind` (`Fn`, `FnMut`, and `FnOnce`, which are not currently supported in our type inference as far as I understand) and `UpvarId` (it would be the pair of the original `RsElement` and closure, but I'm not sure it would be required). So for now I keep this class empty; do you prefer `class Upvar() : Categorization()` instead of `object`?

mchernyavsky: Yes, if we need to distinguish between `Upvar` instances (e.g., to store them in a `Set`).

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
ortem: Maybe, but I use `sealed class` to ensure uniformity (see `PointerKind`)

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
ortem: OK, it has a point.

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
ortem: Are you concerned about the name `Cmt`? Do you have better name for it?

mchernyavsky: `ExprInfo`? 🤔

```Kotlin
val RsExpr.info: ExprInfo? get() = processExpr(this)
```
looks better than
```Kotlin
val RsExpr.cmt: Cmt? get() = processExpr(this)
```
IMO

ortem: `ExprInfo` is a quite ambiguous name. Also `cmt` (as parameter/variable) is used widely in rustc (as well as `ty`).

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -6,56 +6,165 @@
...
+    data class BorrowedPointer(val borrowKind: BorrowKind, val region: Region) : PointerKind()
+    data class UnsafePointer(val mutability: Mutability) : PointerKind()
+}
+
+sealed class InteriorKind {
+    object InteriorField : InteriorKind()
+    object InteriorElement : InteriorKind()
+}
+
+sealed class ImmutabilityBlame {
```
vlad20012: Some documentation would be welcome :)

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,140 +5,368 @@
...
+}
+
+sealed class PointerKind {
+    data class BorrowedPointer(val borrowKind: BorrowKind, val region: Region) : PointerKind()
+    data class UnsafePointer(val mutability: Mutability) : PointerKind()
+}
+
+/** "interior" means "something reachable from the base without a pointer dereference" */
+sealed class InteriorKind {
+    class InteriorField(val fieldName: String?) : InteriorKind()    // e.g. `s.field`
```
vlad20012: Lest use javadoc instead of raw comments

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,140 +5,368 @@
...
+    object InteriorIndex : InteriorKind()                           // e.g. `arr[0]`
+    object InteriorPattern : InteriorKind()                         // e.g. `fn foo([_, a, _, _]: [A; 4]) { ... }`
+}
+
+sealed class ImmutabilityBlame {
```
vlad20012: Pls add some comment about what is the ImmutabilityBlame at all

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,140 +5,368 @@
...
+    class ImmutableLocal(val element: RsElement) : ImmutabilityBlame()  // immutable as immutable variable
+    class LocalDeref(val element: RsElement) : ImmutabilityBlame()      // immutable as dereference of immutable variable
+    object AdtFieldDeref : ImmutabilityBlame()                          // immutable as interior of immutable
+}
+
+sealed class Aliasability {
```
vlad20012: Docs

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,140 +5,368 @@
...
+            category is Interior -> category.cmt.aliasability
+            category is Downcast -> category.cmt.aliasability
+            else -> NonAliasable
+        }
 }
 
-private fun processDotExpr(dotExpr: RsDotExpr): Cmt {
-    if (dotExpr.methodCall != null) {
-        return processRvalue(dotExpr)
+class MemoryCategorizationContext(val infcx: RsInferenceContext) {
```
vlad20012: Let's use ImplLookup here

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -5,140 +5,368 @@
...
+        return Cmt(indexExpr, Interior(baseCmt, InteriorIndex), baseCmt.mutabilityCategory.inherit(), type)
+    }
 
-private fun processExpr(expr: RsExpr): Cmt {
-    val adjustments = expr.inference?.adjustments?.get(expr) ?: emptyList()
-    return processExprAdjustedWith(expr, adjustments.asReversed().iterator())
-}
+    private fun processPathExpr(pathExpr: RsPathExpr): Cmt {
+        val type = pathExpr.type
+        val declaration = pathExpr.path.reference.resolve() ?: return Cmt(pathExpr, ty = type)
```
vlad20012: Let's use `pathExpr.path.reference.deepResolve()` to make it work with type aliases

PR #2708, opened 2018-07-24
Add deref adjustments inside inferIndexExprType

### LINKED ISSUE #2711: Warning for valid code: "Cannot borrow immutable local variable as mutable"
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2101-181
* **Rust toolchain version:** rustc 1.29.0-nightly (6a1c0637c 2018-07-23)
* **IDE name and version:** Intellij IDEA Ultimate 2018.6.1
* **Operating system:** Ubuntu 16.04

## Problem description
In the latest version of the plugin, I'm getting warnings about mutably borrowing an immutable local variable. The local variable is immutable, but the value is a mutable reference to a Vec, and mutably borrowing an element from such a mutable Vec reference is allowed.

## Steps to reproduce
```rust
fn example() {
    let mut a = vec![1, 2, 3];
    let b = &mut a;
    let c = &mut b[0];
}
```

Will show

mutability

This code will compile just fine however.

ortem:
Fixes #2711

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -233,8 +233,10 @@ class RsInferenceContext(
         }
     }
 
-    fun addAdjustment(expression: RsExpr, adjustment: Adjustment) {
-        adjustments.getOrPut(expression) { mutableListOf() }.add(adjustment)
+    fun addAdjustment(expression: RsExpr, adjustment: Adjustment, count: Int = 1) {
+        repeat(count) {
+            adjustments.getOrPut(expression) { mutableListOf() }.add(adjustment)
+        }
     }
 
     fun reportTypeMismatch(expr: RsExpr, expected: Ty, actual: Ty) {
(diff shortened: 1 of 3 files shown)


Undin:
@ortem Let's add test(s) related to #2711 to ensure that we really fix the bug

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1363,13 +1363,27 @@ class RsFnInferenceContext(
...
-            ?: TyUnknown
+
+        var derefCount = -1 // starts with -1 because the fist element of the coercion sequence is the type itself
+        var prevType: Ty? = null
+        for (type in lookup.coercionSequence(containerType)) {
+            val outputType = lookup.findIndexOutputType(type, indexType)
+            if (!isArrayToSlice(prevType, type)) derefCount++
+
+            if (outputType != null) {
+                expr.containerExpr?.let { ctx.addAdjustment(it, Adjustment.Deref(containerType), derefCount) }
```
vlad20012: I don't sure, but do we N^2 adjustments instead of N? I.e. this `addAdjustment` call is inside a loop, and athere is a loop inside `addAdjustment`

ortem:
bors r=vlad20012

Undin:
bors r=vlad20012

PR #2718, opened 2018-07-29
RES: dependency key reference in Cargo.toml

### LINKED ISSUE #1779: Go to lib.rs from crago.toml deps
>  @matklad: The minimal feature for "smart" Cargo.toml would be a reference from foo string in [dependencies] foo = "1.0.0" to the lib.rs of the foo crate.

Undin:
Closes #1779


src/main/kotlin/org/rust/toml/CargoTomlKeyReferenceContributor.kt
```diff
@@ -0,0 +1,77 @@
...
+                        e.names.lastOrNull()?.isDependencyKey == true
+                    }
+                )
+            )
+
+        private val specificDependencyKeyPattern: PsiElementPattern.Capture<TomlKey> = psiElement<TomlKey>(TOML_KEY_CONTEXT_NAME)
+            .withParent(psiElement<TomlTableHeader>()
+                .with("specificDependencyCondition") { e, context ->
+                    val key = context?.get(TOML_KEY_CONTEXT_NAME) ?: return@with false
+                    if (e !is TomlTableHeader) return@with false
```
fedochet: This check better be replaced with `if (e == null) return@with false`
And maybe the whole lambda body should be moved to the separate function to avoid so many local returns?

src/main/kotlin/org/rust/toml/CargoTomlKeyReferenceContributor.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    companion object {
+
+        private const val TOML_KEY_CONTEXT_NAME = "key"
+
+        private val dependencyKeyPattern: PsiElementPattern.Capture<TomlKey> = psiElement<TomlKey>()
+            .withSuperParent(2, psiElement<TomlTable>()
+                .withChild(psiElement<TomlTableHeader>()
+                    .with("dependenciesCondition") { e ->
+                        if (e !is TomlTableHeader) return@with false
```
fedochet: Same here, I guess - can be just checked for null

src/main/kotlin/org/rust/toml/CargoTomlKeyReferenceContributor.kt
Undin: Hm, I was sure that I need a cast here but I don't know why.
Really, we don't such check at all because `e` is always not null

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTestBase.kt
```diff
@@ -41,27 +45,33 @@ abstract class RsResolveTestBase : RsTestBase() {
...
     protected fun stubOnlyResolve(@Language("Rust") code: String) {
-        val testProject = fileTreeFromText(code)
-            .createAndOpenFileWithCaretMarker()
+        stubOnlyResolve<RsWeakReferenceElement>(fileTreeFromText(code))
+    }
+
+    protected fun stubOnlyResolve(@Language("Rust") code: String, mark: Testmark) =
+        mark.checkHit { stubOnlyResolve(code) }
+
+    protected inline fun <reified T: PsiElement> stubOnlyResolve(fileTree: FileTree) {
```
vlad20012: I think it's a bad practice to create large inline functions. It the case of reified types it's simple to split it into 2 methods like here https://github.com/intellij-rust/intellij-rust/pull/2647/files#diff-c2a83b6bef270a72d06a7ba0972ecf61R56

PR #2744, opened 2018-08-04
RES: use correct prelude module

Undin:
Currently, we add prelude from `std` crate to all modules.
But if crate has `#![no_std]` or `#![no_core]` attributes we shouldn't do it.
In the case of `#![no_std]` compiler implicitly adds prelude from `core` crate and
in the case of `#![no_core]` compiler does nothing.

Also, there is unstable `#[prelude_import]` attribute that used in `std` and `core`
to explicitly add some custom prelude (actually, it is used to add prelude of `std` and `core`
to the corresponding crate because they use `#![no_std]` and `#![no_core]` respectively)

These changes implement similar logic as compiler uses. Note, we don't search
`#[prelude_import]` and just add `std::prelude` to `std` and `core::prelude` to `core` crates.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/util/RustCrateUtil.kt
+++ b/src/main/kotlin/org/rust/cargo/util/RustCrateUtil.kt
@@ -35,14 +35,14 @@ data class StdLibInfo (
 )
 
 object AutoInjectedCrates {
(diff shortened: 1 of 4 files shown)


Undin:
bors r=vlad20012

PR #2764, opened 2018-08-11
CARGO: improve rustfmt formating

### LINKED ISSUE #2663: No documentation for "Use cargo check to analyze code" and other Languages & Frameworks options
Using the latest versions of intellij-rust (0.2.0.2100-182 ) and CLion (2018.5).

Not sure if this is captured anywhere, but there doesn't appear to be any tooltips or documentation for the Languages & Frameworks options. Eg to figure out what "cargo check to analyze code" does, I had to dig through bug reports. And I'm still only vaguely aware of what this means.

It'd be nice if somebody spent a small amount of time to provide a paragraph-sized tooltip of what each option means. I don't know if this is the Intellj way, I've only started using Intellj because of Rust. But in my opinion this would be much less cognitive overhead than having to context-switch and search the web to figure out what every option does.

Thanks.

### LINKED ISSUE #2761: Cargo format returns with an error: Unstable option (`--skip-children`) used without `--unstable-features`
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2103-182
* **Rust toolchain version:** nightly-x86_64-pc-windows-gnu rustc 1.30.0-nightly (73c78734b 2018-08-05)
* **IDE name and version:** CLion 2018.2

## Problem description

Cargo format doesn't run. I get the following error:

```
Cargo execution failed (exit code 1).
    C:/Users/USER/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/bin/cargo.exe fmt --all -- --emit=files --skip-children C:/Path/to/mod.rs
    stdout : Unstable option (`--skip-children`) used without `--unstable-features`
    stderr : 
```

## Steps to reproduce

Setup CLion as documented and run rust format. I'm running on the current latest gnu stable nightly toolchain and MinGW.

Undin:
* Add panel for rustfmt with options to use `--skip-children` flag
* Use `rustfmt` directly instead of `cargo fmt` to reformat only one file
* Allow a user to install `rustfmt` from IDE

Fixes #2761

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/Rustup.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Rustup.kt
@@ -30,16 +30,16 @@ class Rustup(
         val downloadProcessOutput = GeneralCommandLine(rustup)
             .withWorkDirectory(projectDirectory)
             .withParameters("component", "add", "rust-src")
-            .exec()
+            .execute(null)
 
-        return if (downloadProcessOutput.exitCode != 0) {
-            val message = "rustup failed: `${downloadProcessOutput.stderr}`"
-            LOG.warn(message)
-            DownloadResult.Err(message)
-        } else {
+        return if (downloadProcessOutput?.isSuccess == true) {
             val sources = getStdlibFromSysroot() ?: return DownloadResult.Err("Failed to find stdlib in sysroot")
             fullyRefreshDirectory(sources)
             DownloadResult.Ok(sources)
+        } else {
+            val message = "rustup failed: `${downloadProcessOutput?.stderr ?: ""}`"
+            LOG.warn(message)
+            DownloadResult.Err(message)
         }
     }
 
--- a/src/main/kotlin/org/rust/cargo/toolchain/Rustup.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Rustup.kt
@@ -5,6 +5,8 @@
 
 package org.rust.cargo.toolchain
 
+import com.intellij.execution.ExecutionException
+import com.intellij.openapi.Disposable
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.util.io.FileUtil
 import com.intellij.openapi.vfs.LocalFileSystem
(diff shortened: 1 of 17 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Rustup.kt
```diff
@@ -21,25 +23,40 @@ class Rustup(
     private val rustup: Path,
     private val projectDirectory: Path
 ) {
-    sealed class DownloadResult {
-        class Ok(val library: VirtualFile) : DownloadResult()
-        class Err(val error: String) : DownloadResult()
+    @Suppress("unused")
+    sealed class DownloadResult<out T> {
+        class Ok<T>(val value: T) : DownloadResult<T>()
+        class Err<T>(val error: String) : DownloadResult<T>()
```
vlad20012: I released that we can define Err class as
`class Err(val error: String) : DownloadResult<Nothing>`

src/main/kotlin/org/rust/ide/actions/RustfmtFileAction.kt
```diff
@@ -26,15 +37,26 @@ class RustfmtFileAction : DumbAwareAction() {
...
 
-        val cargo = toolchain.rawCargo()
+        FileDocumentManager.getInstance().saveAllDocuments()
+
+        val rustfmt = toolchain.rustfmt()
         try {
-            cargo.reformatFile(project, file)
+            ProgressManager.getInstance().runProcessWithProgressSynchronously<ProcessOutput, ExecutionException>({
+                rustfmt.reformatFile(project, file)
+            }, "Reformat File with Rustfmt", true, project)
```
vlad20012: The title describes a process that is currently running, so it should be in continuous form and with ellipsis in the end: "Reformatting File with Rustfmt..."
image

src/main/kotlin/org/rust/cargo/toolchain/Rustup.kt
Undin: Indeed:)
And probably, we want to implement a `Result` class for common purposes because we already have at least another two implementations: `TaskResult` and `ResolveResult`

vlad20012: Maybe we want to use the library https://github.com/michaelbull/kotlin-result

src/test/kotlin/org/rust/cargo/commands/RustfmtTest.kt
```diff
@@ -5,13 +5,16 @@
...
+import com.intellij.testFramework.TestActionEvent
 import org.rust.cargo.RustWithToolchainTestBase
-import org.rust.cargo.project.settings.toolchain
 import org.rust.fileTree
+import org.rust.ide.actions.RustfmtFileAction
 
-class CargoFmtTest : RustWithToolchainTestBase() {
+class RustfmtTest : RustWithToolchainTestBase() {
 
     fun `test cargo fmt`() {
```
vlad20012: "test rustfmt"

PR #2770, opened 2018-08-13
TY: Implement scopes and free regions

mchernyavsky:
Will be used in the future for region inference and borrow checker.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/outlives/TransitiveRelation.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/outlives/TransitiveRelation.kt
@@ -1,0 +1,217 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.types.infer.outlives
+
+import org.rust.stdext.mapToMutableList
+import org.rust.stdext.removeLast
+import java.util.*
+import java.util.stream.Collectors
+
+class TransitiveRelation<T> {
+    /** List of elements. This is used to map from a T to an index. */
+    private val elements: MutableList<T> = mutableListOf()
+
+    /** Maps each element to an index. */
+    private val map: MutableMap<T, Int> = hashMapOf()
+
+    /** List of base edges in the graph. Require to compute transitive closure. */
+    private val edges: MutableList<Edge> = mutableListOf()
+
+    /** Cached transitive closure derived from the edges. */
+    private var closure: BitMatrix? = null
+
+    fun isEmpty(): Boolean = edges.isEmpty()
+
+    /** Indicate that `[element1] < [element2]` (where `<` is this relation). */
+    fun add(element1: T, element2: T) {
+        val source = addIndex(element1)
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/outlives/TransitiveRelation.kt
```diff
@@ -0,0 +1,196 @@
...
+        return bits.get(getIndex(row, column))
+    }
+
+    /** @return those indices that are true in [row1] and [row2]. */
+    fun intersection(row1: Int, row2: Int): MutableList<Int> {
+        require(row1 < rows && row2 < rows)
+        val bits1 = bits.get(getIndex(row1), getIndex(row1 + 1))
+        val bits2 = bits.get(getIndex(row2), getIndex(row2 + 1))
+        bits1.and(bits2)
+        return bits1.stream().toList().toMutableList()
```
Undin: Maybe we implement own conversion to list for `BitSet`? Because current code copies bit set two times

src/main/kotlin/org/rust/lang/core/types/infer/outlives/TransitiveRelation.kt
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    private fun getIndex(row: Int, column: Int = 0): Int = row * columns + column
+}
+
+private data class Edge(val source: Int, val target: Int)
+
+private fun <T> MutableList<T>.removeLast(): T = removeAt(size - 1)
```
Undin: Probably we want to have this method in `org.rust.stdext.Collections.kt`

src/main/kotlin/org/rust/lang/core/types/infer/outlives/TransitiveRelation.kt
fedochet: We can use `toSequence` instead of `toList` to avoid that, I guess

src/main/kotlin/org/rust/lang/core/types/infer/outlives/TransitiveRelation.kt
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return matrix
+    }
+}
+
+/**
+ * Pare down is used as a step in the LUB computation. It edits the candidates array in place by removing any element
+ * j for which there exists an earlier element `i < j` such that `i -> j`.
+ */
+private fun pareDown(candidates: MutableList<Int>, closure: BitMatrix) {
```
ortem: Probably replace with something like that:
```
private fun pareDown(candidates: MutableList<Int>, closure: BitMatrix) {
    for (leftIndex in 0..candidates.size) {
        var dead = 0
        for (rightIndex in leftIndex+1..candidates.size) {
            val left = candidates[leftIndex]
            val right = candidates[rightIndex]
            if (closure[left, right]) {
                dead++
            } else {
                candidates[rightIndex - dead] = right
            }
        }
        candidates.subList(candidates.size - dead, candidates.size).clear()
    }
}
```

PR #2774, opened 2018-08-14
INSP: Unresolved reference inspection and auto import quick fix for trait method references

### LINKED ISSUE #2402: Highlight unimported methods as unresolved
Now our `Unresolved reference` inspection uses `resolve` method to determine if a path can be resolved or not.
But `resolve` works for unimported methods too.
In the following case
```rust
mod foo {
    pub trait Foo {
        fn foo(&self);
    }

    impl Foo for i32 {
        fn foo(&self) {
            println!("{}", self);
        }
    }
}

fn main() {
    1i32.foo()
}
```
we should highlight call of `foo` method as unresolved and add quick fix to import the corresponding trait

Undin:
Closes #2402

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -18,6 +18,7 @@ import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.cargo.util.AutoInjectedCrates
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.resolve.ref.deepResolve
 import org.rust.lang.core.stubs.index.RsNamedElementIndex
 import org.rust.lang.core.stubs.index.RsReexportIndex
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -173,6 +187,44 @@ class AutoImportFix(path: RsPath) : LocalQuickFixOnPsiElement(path), HighPriorit
             return Context(basePath, candidates)
         }
 
+        fun findApplicableContext(project: Project, methodCall: RsMethodCall): Context<Unit>? {
+            val results = methodCall.reference.multiResolve()
+            if (results.isEmpty()) return Context(Unit, emptyList())
+
+            val lookup = ImplLookup.relativeTo(methodCall)
+            val mod = methodCall.containingMod
```
vlad20012: Why do we skip scopes to the mod declaration? I can imagine an example where it is wrong:
```rust
mod foo {
    fn foo(a: A) {
        use bar::Baz;
        a.baz(); // baz defined on Baz trait that is implemented for `A` type
    }
}
```

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -354,8 +406,8 @@ class AutoImportFix(path: RsPath) : LocalQuickFixOnPsiElement(path), HighPriorit
         }
     }
 
-    data class Context(
-        val basePath: RsPath,
+    data class Context<Data>(
```
vlad20012: I think it's better to use single letter generic name according to java conventions

PR #2777, opened 2018-08-16
Create RsShowControlFlowGraphAction

ortem:
The action builds the control-flow graph of a function under caret and shows the graph. To use this action you have to install `graphviz` package.
screen shot 2018-08-16 at 12 35 32
screen shot 2018-08-16 at 12 36 32

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
@@ -1,0 +1,87 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.actions
+
+import com.intellij.execution.ExecutionException
+import com.intellij.execution.configurations.GeneralCommandLine
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.CommonDataKeys
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.application.PathManager
+import com.intellij.openapi.progress.ProgressManager
+import com.intellij.openapi.ui.Messages
+import com.intellij.util.io.createDirectories
+import com.intellij.util.io.delete
+import org.rust.lang.core.cfg.ControlFlowGraph
+import org.rust.lang.core.psi.RsBlock
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.ext.ancestorOrSelf
+import org.rust.openapiext.execute
+import org.rust.openapiext.isSuccess
+import org.rust.openapiext.withWorkDirectory
+import java.awt.Desktop
+import java.io.File
+import java.nio.file.Paths
+
+class RsShowControlFlowGraphAction : AnAction() {
+    private val tempDir = Paths.get(PathManager.getTempPath()).resolve(CONTROL_FLOW_GRAPH)
+
+    init {
+        tempDir.delete()
+    }
+
+    override fun update(event: AnActionEvent) {
+        event.presentation.isEnabledAndVisible = getBody(event.dataContext) != null
+    }
+
+    override fun actionPerformed(event: AnActionEvent) {
+        val body = getBody(event.dataContext) ?: return
+
+        val cfg = ControlFlowGraph.buildFor(body)
+        val dotDescription = cfg.createDotDescription()
+
+        val dotFileName = "cfg${cfg.hashCode()}.dot"
+        val pngFileName = "cfg${cfg.hashCode()}.png"
+
+        val dotFile = ProgressManager.getInstance().runProcessWithProgressSynchronously<File, ExecutionException>({
+            createTempFileWithContent(dotFileName, dotDescription)
+        }, "Creating dot file...", true, event.project)
+
+        val pngFile = tempDir.resolve(pngFileName).toFile()
+
+        val result = GeneralCommandLine("dot", "-Tpng", dotFileName, "-o", pngFileName)
+            .withWorkDirectory(tempDir)
+            .execute()
+
+        if (result != null && result.isSuccess && pngFile.exists()) {
+            Desktop.getDesktop().open(pngFile)
+        } else {
+            Messages.showInfoMessage("Please ensure that you have Graphviz installed", "Cannot show graph")
+        }
+
+        dotFile.deleteOnExit()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+        val dotFileName = "cfg${cfg.hashCode()}.dot"
+        val pngFileName = "cfg${cfg.hashCode()}.png"
+
+        val dotFile = createTempFileWithContent(dotFileName, dotDescription)
+        val pngFile = tempDir.resolve(pngFileName).toFile()
+
+        try {
+            val dotFilePath = tempDir.resolve(dotFileName)
+            val pngFilePath = tempDir.resolve(pngFileName)
+            Runtime.getRuntime().exec("dot -Tpng $dotFilePath -o $pngFilePath").waitFor()
```
Undin: The platform has `GeneralCommandLine` class to invoke command line actions
Also we have some extensions for it (I've just moved them into more proper package, so see [this](https://github.com/intellij-rust/intellij-rust/pull/2785) PR)

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    init {
+        tempDir.delete()
+    }
+
+    override fun actionPerformed(event: AnActionEvent) {
+        val body = getBody(event.dataContext) ?: return
+
+        val cfg = ControlFlowGraph.buildFor(body)
+        val dotDescription = cfg.createDotDescription()
```
Undin: Looks like there is a bug in `org.rust.lang.core.cfg.ControlFlowGraph#createDotDescription` because it creates invalid dot file (at least for `graphviz version 2.40.1 (20161225.0304)` from homebrew on my mac) for the following code
```rust
struct EE {
    pub field1: u32,
    pub field2: u8,
}

type XX = EE;

impl EE {
    pub fn new() -> XX {
        XX { field1: 2, field2: 3 }
    }
}

fn main() {
    let xx = XX::new(); // new() is not recognized by intellisense
    println!("{:?}", xx.field2); //xx type in not recognized by intellisense
/*caret*/
    let yy = EE::new();// new() is not recognized by intellisense
    println!("{:?}", yy.field2); //yy type in not recognized by intellisense
}
```

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.rust.lang.core.psi.RsFunction
+import java.awt.Desktop
+import java.io.File
+import java.io.IOException
+import java.nio.file.Paths
+
+
+class RsShowControlFlowGraphAction : AnAction() {
+    private val CONTROL_FLOW_GRAPH = "control_flow_graph"
+    private val tempDir = Paths.get(PathManager.getTempPath()).resolve(CONTROL_FLOW_GRAPH)
```
Undin: Are we sure that we want to create such files in tmp directory?
It's rather hard and inconvenient to find them there that can be important for such internal actions (for example, for bug fixing or something else)
Maybe it's better to create them in a folder in project base dir?
@vlad20012 what do you think?

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    init {
+        tempDir.delete()
```
Undin: Why do we need to drop this folder at all?

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+        } catch (e: IOException) {
+            Messages.showInfoMessage(e.message, "Cannot show graph")
+        }
+
+        if (pngFile.exists()) {
+            Desktop.getDesktop().open(pngFile)
+        }
+
+        dotFile.deleteOnExit()
+        pngFile.deleteOnExit()
```
Undin: Related to comment about tmp directory - I'm not sure that it's a good idea to delete these files because they can be useful outside of IDE in some cases

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
ortem: Thanks, I forgot to shield quotes.

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        try {
+            val dotFilePath = tempDir.resolve(dotFileName)
+            val pngFilePath = tempDir.resolve(pngFileName)
+            Runtime.getRuntime().exec("dot -Tpng $dotFilePath -o $pngFilePath").waitFor()
+        } catch (e: IOException) {
+            Messages.showInfoMessage(e.message, "Cannot show graph")
+        }
+
+        if (pngFile.exists()) {
```
mchernyavsky: Maybe we should move this block to the try-catch above?

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.rust.lang.core.cfg.ControlFlowGraph
+import org.rust.lang.core.psi.RsBlock
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.ext.ancestorOrSelf
+import java.awt.Desktop
+import java.io.File
+import java.io.IOException
+import java.nio.file.Paths
+
+class RsShowControlFlowGraphAction : AnAction() {
```
mchernyavsky: I think this action should be available only for Rust files.
You can override `update` method for this purpose:
```kotlin
override fun update(event: AnActionEvent) {
    val file = event.getData(CommonDataKeys.PSI_FILE)
    event.presentation.isEnabled = file is RsFile
}
```

src/main/kotlin/org/rust/ide/actions/RsShowControlFlowGraphAction.kt
ortem: I think it's not so useful because you can just perform the action again.

PR #2782, opened 2018-08-20
Test that Sized win ?Sized if they both are listed

dem1tris:
Hello! Just added a couple of tests for case when both of `Sized` (itself or as a supertrait) and `?Sized` are bounds for type parameter. They can't be passed for now, but should be.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -20,6 +20,7 @@ chriskrycho
 contradictioned
 d9n
 daschl
+dem1tris
 devurandom
 Dimonchik0036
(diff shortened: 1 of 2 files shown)


ortem:
@dem1tris Thanks for your test cases!

Pull request with failing tests cannot be merged by our CI. You should create an issue with your test cases instead of the pull request.

dem1tris:
@ortem, done: https://github.com/intellij-rust/intellij-rust/issues/2783
Should I copy the code there?

ortem:
@dem1tris Yes, it'd be nice!

PR #2789, opened 2018-08-23
GRAM: remove old reserved keywords from lexer

### LINKED ISSUE #2620: 'proc' is no longer a reserved keyword
As of Rust 1.27.0, ```proc``` is no longer a reserved keyword and may be used as an identifier. Currently, the plugin doesn't recognize this and sees it as an error. Despite this, projects will compile correctly, as expected.

Undin:
Fixes #2620


Undin:
bors retry

PR #2790, opened 2018-08-24
T: Introduce @ProjectDescriptor annotation

vlad20012:
It allows to set project descriptor (like
WithStdRustProjectDescriptor) for a specific
test method. Previously it was possible to set
project descriptor for entire test class only.


src/test/kotlin/org/rust/lang/RsTestBase.kt
```diff
@@ -57,8 +47,14 @@ abstract class RsTestBase : LightPlatformCodeInsightFixtureTestCase(), RsTestCas
...
         super.setUp()
-        val testMethod = javaClass.getMethod(name)
-        val version = testMethod.getAnnotation(MockRustcVersion::class.java)?.rustcVersion ?: return
+
+        (projectDescriptor as? RustProjectDescriptorBase)?.setUp(myFixture)
+
+        setupRustcVersionMock()
+    }
+
+    private fun setupRustcVersionMock() {
```
Undin: Maybe `setupMockRustcVersion` is better?

src/test/kotlin/org/rust/lang/ProjectDescriptor.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Example values:
+ * - [WithStdlibRustProjectDescriptor]
+ * - [WithStdlibAndDependencyRustProjectDescriptor]
+ *
+ * @see RsTestBase.getProjectDescriptor
+ */
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ProjectDescriptor(val descriptor: KClass<out LightProjectDescriptor>)
```
Undin: If we want to use only object classes as descriptors, maybe we pass object instance here directly instead of `KClass`?

vlad20012: It isn't possible for java/kotlin annotations =(

vlad20012:
bors r=undin

mchernyavsky:
bors retry

Undin:
Let's try again
bors r+

PR #2804, opened 2018-08-28
INSP: Do not highlight comments and attributes for impl...

mchernyavsky:
...by `Different impl member order from the trait` inspection.

The problem:
screen shot 2018-08-28 at 18 06 41


Undin:
What about `unsafe` and `default` keywords before `impl`?
I think we should include them into text range of annotation

Undin:
bors retry

PR #2806, opened 2018-08-29
HINT: RsGenericParameterInfoHandler

dem1tris:
Implemented type hints for generic type parameters.

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
@@ -8,48 +8,46 @@ package org.rust.ide.hints
 import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.lang.parameterInfo.*
 import com.intellij.openapi.util.TextRange
-import com.intellij.psi.PsiElement
+import com.twelvemonkeys.lang.StringUtil
 import org.rust.ide.utils.CallInfo
-import org.rust.lang.core.psi.RsCallExpr
-import org.rust.lang.core.psi.RsMethodCall
-import org.rust.lang.core.psi.RsTypeArgumentList
-import org.rust.lang.core.psi.RsValueArgumentList
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsGenericDeclaration
 import org.rust.lang.core.psi.ext.ancestorStrict
 import org.rust.lang.core.types.type
 import org.rust.stdext.buildList
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import sun.swing.StringUIClientPropertyKey
+import kotlin.jvm.javaClass
 
-// RsArgumentsDescription from RsParameterInfoHandler
-class RsGenericParameterInfoHandler : ParameterInfoHandler<PsiElement, RsTypesDescription> {
+class RsGenericParameterInfoHandler : ParameterInfoHandler<RsTypeArgumentList, RsTypeParameter> {
+    fun newLogger(): Logger = LoggerFactory.getLogger(javaClass)
+    val log = newLogger()
 
     var hintText: String = ""
 
-    override fun showParameterInfo(element: PsiElement, context: CreateParameterInfoContext) {
-        if (element !is RsTypeArgumentList) return
-        val valueArgList = element.parent.parent.nextSibling as? RsValueArgumentList ?: return
-        val argsDescr = RsTypesDescription.findDescriptionInExpr(valueArgList) ?: return
-        context.itemsToShow = arrayOf(argsDescr) //arrayOf(RsArgumentsDescription(arrayOf("Dummy hint")))
-        context.showHint(element, element.textRange.startOffset, this);
+    // done, todo: check offset
+    override fun showParameterInfo(element: RsTypeArgumentList, context: CreateParameterInfoContext) {
+        context.showHint(element, element.textRange.startOffset, this)
     }
 
-    override fun updateParameterInfo(parameterOwner: PsiElement, context: UpdateParameterInfoContext) {
+    //
+    override fun updateParameterInfo(parameterOwner: RsTypeArgumentList, context: UpdateParameterInfoContext) {
+        // val argIndex = findArgumentIndex
         TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
     }
 
-    override fun updateUI(p: RsTypesDescription?, context: ParameterInfoUIContext) {
+    override fun updateUI(p: RsTypeParameter, context: ParameterInfoUIContext) {
+        System.err.println("${p.name}")
+        System.err.println("${p.outerAttrList.map { it -> it.text }}")
+        System.err.println("${p.typeParamBounds?.text}") //polyboundList?.map {it -> it.bound.text}}")
+        hintText = p.name + p.outerAttrList.map { it -> it.text }
         // Bulb appears not strictly above the cursor sometimes
-        if (p == null) {
-            context.isUIComponentEnabled = false
-            return
-        }
-        val range = p.getArgumentRange(context.currentParameterIndex)
-        hintText = p.presentText
-        System.err.println(range) // dbg
-        System.err.println(range.startOffset.toString() + "," + range.endOffset) // dbg
-        p.arguments.asIterable().forEach({System.err.println(it)}) // dbg
+        val highlightEndOffset = hintText.length
         context.setupUIComponentPresentation(
             hintText,
-            range.startOffset,
-            range.endOffset,
+            0,
+            highlightEndOffset,
             false, // grayed  // !context.isUIComponentEnabled, ???
             false,
             false, // Define grayed part of args before highlight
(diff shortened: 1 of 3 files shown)


ortem:
@dem1tris Could you squash these commits?

dem1tris:
@ortem done

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+            context.parameterOwner == null -> context.parameterOwner = parameterOwner
+
+            // occurs e.g. in case of up-down cursor moving
+            context.parameterOwner != parameterOwner -> context.removeHint()
+        }
+    }
+
+    override fun updateUI(p: RsGenericPresentation, context: ParameterInfoUIContext) {
+        hintText = p.presentText
+        val (startOffset, endOffset) =
```
fedochet: Please, be consistent with `if` braces; it is ok not to use them only when `if expression` can be put on one line. In This case you should add braces like 
```kt
            if (hintText.startsWith(WHERE_PREFIX)) {
                0 to WHERE_PREFIX.length 
            } else {
                p.getRange(curParam).startOffset to p.getRange(curParam).endOffset
            }
```

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+            if (hintText.startsWith(WHERE_PREFIX))
+                0 to WHERE_PREFIX.length
+            else
+                p.getRange(curParam).startOffset to p.getRange(curParam).endOffset
+        context.setupUIComponentPresentation(
+            hintText,
+            startOffset,
+            endOffset,
+            false, // grayed
```
fedochet: Consider using named params (like `grayed = false` instead of just `false`) to increase readability

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+        val genericDeclaration = when (parent) {
+            is RsMethodCall,
+            is RsPath -> parent.reference?.resolve()
+            else -> return null
+        } as? RsGenericDeclaration ?: return null
+        val typesWithBounds = genericDeclaration.typeParameters.nullize() ?: return null
+        context.itemsToShow = listOfNotNull(
+            RsGenericPresentation(typesWithBounds, false),
+            RsGenericPresentation(typesWithBounds, true))
+            .filterNot { it.presentText == "" }
```
fedochet: Use `isEmpty` instead if possible

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+    override fun findElementForParameterInfo(context: CreateParameterInfoContext): RsTypeArgumentList? {
+        val parameterList = findExceptColonColon(context) ?: return null
+        val parent = parameterList.parent
+        val genericDeclaration = when (parent) {
+            is RsMethodCall,
+            is RsPath -> parent.reference?.resolve()
+            else -> return null
+        } as? RsGenericDeclaration ?: return null
+        val typesWithBounds = genericDeclaration.typeParameters.nullize() ?: return null
+        context.itemsToShow = listOfNotNull(
```
fedochet: I guess you can just use `listOf` instead

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    override fun couldShowInLookup() = false
+
+    override fun findElementForUpdatingParameterInfo(context: UpdateParameterInfoContext): RsTypeArgumentList? =
+        findExceptColonColon(context)
+
+    override fun findElementForParameterInfo(context: CreateParameterInfoContext): RsTypeArgumentList? {
+        val parameterList = findExceptColonColon(context) ?: return null
+        val parent = parameterList.parent
+        val genericDeclaration = when (parent) {
```
fedochet: `when` with only one condition is often better to be rewritten with simple `if`
```kt
      val genericDeclaration = if (parent is RsMethodCall || parent is RsPath) {
            parent.reference?.resolve() as? RsGenericDeclaration ?: return null
      } else {
            return null
      }
```

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+            .filterNot { it.presentText == "" }
+            .toTypedArray()
+        return parameterList
+    }
+
+    // to avoid hint on :: before <>
+    private fun findExceptColonColon(context: ParameterInfoContext?): RsTypeArgumentList? {
+        val element = context?.file?.findElementAt(context.editor.caretModel.offset) ?: return null
+        if (element.elementType == RsElementTypes.COLONCOLON) return null
+        return element.ancestorStrict() ?: return null
```
fedochet: You do not need to add `return null` in the last nullable return; `return element.ancestorStrict()` is enough

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+        wherePreds.map { it.text }
+    }
+
+    val presentText = toText.nullize()?.joinToString(
+        prefix = if (!unrecognizedPartOfWhere) "" else WHERE_PREFIX
+    ) ?: ""
+
+    fun getRange(index: Int): TextRange {
+        return if (index < 0 || index >= ranges.size)
```
fedochet: Use `index !in 0 until ranges.size` instead

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+                params.contains((it.typeReference?.typeElement as? RsBaseType)?.path?.reference?.resolve())
+            }
+        wherePreds.map { it.text }
+    }
+
+    val presentText = toText.nullize()?.joinToString(
+        prefix = if (!unrecognizedPartOfWhere) "" else WHERE_PREFIX
+    ) ?: ""
+
+    fun getRange(index: Int): TextRange {
```
fedochet: This is single expression function, use `=` instead of braced body

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+        return if (index < 0 || index >= ranges.size)
+            TextRange.EMPTY_RANGE
+        else
```
fedochet: Use braces or put `if expr` on single line

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Stores the text representation and ranges for parameters
+ */
+
+class RsGenericPresentation(
+    private val params: List<RsTypeParameter>,
+    unrecognizedPartOfWhere: Boolean
+) {
+    val toText = if (!unrecognizedPartOfWhere) {
```
fedochet: This is rather complicated code; consider splitting it into several functions

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+        params.map { param ->
+            param.name ?: return@map ""
+            val QSizedBound =
+                if (!param.isSized)
```
fedochet: Same `If expression` issue

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+ */
+
+class RsGenericPresentation(
+    private val params: List<RsTypeParameter>,
+    unrecognizedPartOfWhere: Boolean
+) {
+    val toText = if (!unrecognizedPartOfWhere) {
+        params.map { param ->
+            param.name ?: return@map ""
+            val QSizedBound =
```
fedochet: Strange naming, why capital letter first?

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
vlad20012: It isn't possible to specify param names if the method defined in Java code. So a comment is ok, I guess.

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+    var curParam = -1
+        private set
+    var hintText = ""
+        private set
+
+    override fun showParameterInfo(element: RsTypeArgumentList, context: CreateParameterInfoContext) {
+        context.highlightedElement = null
+        context.showHint(element, element.textRange.startOffset, this)
+    }
+
```
vlad20012: Forgotten 4 spaces on this line. I'd recommend enabling the feature to auto remove redundant tabulation :)

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+        val element = context?.file?.findElementAt(context.editor.caretModel.offset) ?: return null
+        if (element.elementType == RsElementTypes.COLONCOLON) return null
+        return element.ancestorStrict() ?: return null
+    }
+}
+
+/**
+ * Stores the text representation and ranges for parameters
+ */
+
```
vlad20012: Redundant linebreak

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
vlad20012: Can be private.
And let's specify types explicitly for each property.
Also, the name `toText` looks like a function name because it's a verb (_convert_ to text). Even more confusing is the fact that this is a **field**, not a kotlin property. Maybe really convert it to a function and compute the value lazily?

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+) {
+    val toText = if (!unrecognizedPartOfWhere) {
+        params.map { param ->
+            param.name ?: return@map ""
+            val QSizedBound =
+                if (!param.isSized)
+                    listOf("?Sized")
+                else
+                    emptyList()
+            val declaredBounds =
```
vlad20012: I think this style looks better:
```
val declaredBounds = param.bounds
    // `?Sized`, if needed, in separate val, `Sized` shouldn't be shown
    .filter { it.bound.traitRef?.resolveToBoundTrait?.element?.isSizedTrait == false }
    .mapNotNull { it.bound.traitRef?.path?.text }
```

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
vlad20012: These hack with true/false params and filters looks like a very bad design.

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.intellij.openapi.util.TextRange
+import com.intellij.util.containers.nullize
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+
+private const val WHERE_PREFIX = "where "
+
+class RsGenericParameterInfoHandler : ParameterInfoHandler<RsTypeArgumentList, RsGenericPresentation> {
+
+    var curParam = -1
```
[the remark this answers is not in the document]
dem1tris: @vlad20012 It's not clear for me, how to use it...
```
@VisibleForTesting
private var curParam = -1
```
has no effect: cannot access from `checkByText()`
In addition, http://plugins.jetbrains.com/plugin/9952-visiblefortesting tells this annotation is for methods. But generating explicit getter with annotation has no effect too.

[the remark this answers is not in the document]
dem1tris: @vlad20012 oh, it just means "I made it visible only for testing"
I thought it is some magic to ignore `private` while testing))

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,140 @@
...
+    val presentText: String,
+    val ranges: List<TextRange>
+) {
+    fun getRange(index: Int) = if (index !in 0 until ranges.size) TextRange.EMPTY_RANGE else ranges[index]
+}
+
+/**
+ * Calculates the text representation and ranges for parameters
+ */
+class RsGenericPresenter(
```
vlad20012: This class isn't needed now, I guess. You can replace it with 2 functions: firstLine(params) and secondLine(params)

dem1tris: @vlad20012 done

src/test/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandlerTest.kt
```diff
@@ -0,0 +1,268 @@
...
+
+    private fun checkByText(@Language("Rust") code: String, hint: String, where: String, index: Int) {
+        myFixture.configureByText("main.rs", replaceCaretMarker(code))
+        val handler = RsGenericParameterInfoHandler()
+        val createContext = MockCreateParameterInfoContext(myFixture.editor, myFixture.file)
+
+        // Check hint
+        val elt = handler.findElementForParameterInfo(createContext)
+        if (hint.isNotEmpty()) {
+            elt ?: throw AssertionFailedError("Hint not found")
```
vlad20012: You can replace `throw AssertionFailedError(...)` to Kotlin's function `error(...)`

src/main/kotlin/org/rust/ide/hints/RsGenericParameterInfoHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+    fun getRange(index: Int) = if (index !in 0 until ranges.size) TextRange.EMPTY_RANGE else ranges[index]
+}
+
+/**
+ * Calculates the text representation and ranges for parameters
+ */
+private fun firstLine(params: List<RsTypeParameter>): HintLine {
+    val splited = params.map { param ->
+        param.name ?: return@map ""
+        val QSizedBound = if (!param.isSized) listOf("?Sized") else emptyList()
```
vlad20012: Local variable names should start with lowercase letter.

vlad20012:
Ok, looks good to me now. Just squash these commits again =)

PR #2807, opened 2018-08-29
add 'iter' postfix template for iterable expressions

Voronchikhin:

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
@@ -11,8 +11,15 @@ import com.intellij.codeInsight.template.postfix.templates.StringBasedPostfixTem
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.RsBinaryExpr
 import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.ext.EqualityOp
 import org.rust.lang.core.psi.ext.operatorType
+import org.rust.lang.core.psi.ext.withSubst
+import org.rust.lang.core.resolve.ImplLookup
+import org.rust.lang.core.resolve.StdKnownItems
+import org.rust.lang.core.types.TraitRef
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.type
 
 abstract class AssertPostfixTemplateBase(
     name: String,
@@ -60,4 +67,27 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
             addVariable("PAT", TextExpression("_"), true)
         }
     }
+}
+
+class IterPostfixTemplate(provider: RsPostfixTemplateProvider) :
+    StringBasedPostfixTemplate("iter", "for x in expr",
+        RsTopMostInScopeSelector { it.implementsIntoIter || it.implementsIter }, provider) {
+    override fun getTemplateString(element: PsiElement): String =
+        "for x in ${element.text} {\n\$END$\n}"
+
+    override fun getElementToRemove(expr: PsiElement): PsiElement = expr
+}
+
+
+private val RsExpr.implementsIntoIter: Boolean
+    get() = isImplementsTrait(this, "iter::IntoIterator")
+
+private val RsExpr.implementsIter: Boolean
+    get() = isImplementsTrait(this, "iter::Iterator")
+
+private fun isImplementsTrait(expr: RsExpr, traitName: String, vararg subst: Ty): Boolean {
+    val items = StdKnownItems.relativeTo(expr)
+    val implLookup = ImplLookup(expr.project, items)
+    val trait = items.findCoreItem(traitName) as? RsTraitItem ?: return false
+    return implLookup.canSelect(TraitRef(expr.type, trait.withSubst(*subst)))
 }
--- a/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
@@ -26,7 +26,8 @@ class RsPostfixTemplateProvider : PostfixTemplateProvider {
         ParenPostfixTemplate(),
         LambdaPostfixTemplate(this),
         NotPostfixTemplate(this),
-        LetPostfixTemplate(this)
+        LetPostfixTemplate(this),
+        IterPostfixTemplate(this)
     )
 
     override fun getTemplates(): Set<PostfixTemplate> = templates
--- a/src/main/kotlin/org/rust/lang/core/resolve/StdKnownItems.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/StdKnownItems.kt
@@ -44,6 +44,9 @@ class StdKnownItems private constructor(
     fun findIteratorTrait(): RsTraitItem? =
         findCoreItem("iter::Iterator") as? RsTraitItem
 
+    fun findIntoIteratorTrait(): RsTraitItem? =
+        findCoreItem("iter::IntoIterator") as? RsTraitItem
+
     fun findVecForElementTy(elementTy: Ty): Ty {
         val ty = findStdTy("alloc", "vec::Vec")
 
--- a/src/main/resources/postfixTemplates/IterPostfixTemplate/description.html
+++ b/src/main/resources/postfixTemplates/IterPostfixTemplate/description.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+Create 'for' loop for collection expression
+</body>
+</html>
--- a/src/test/kotlin/org/rust/ide/template/postfix/IterPostFixTemplateTest.kt
+++ b/src/test/kotlin/org/rust/ide/template/postfix/IterPostFixTemplateTest.kt
@@ -1,0 +1,47 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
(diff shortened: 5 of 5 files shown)


src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -61,3 +67,27 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
...
+    StringBasedPostfixTemplate("iter","for x in expr", RsTopMostInScopeSelector {
+        val items = StdKnownItems.relativeTo(it)
+        val iterTrait = items.findIteratorTrait()?: return@RsTopMostInScopeSelector false
+        val intoIterTrait = items.findCoreItem("iter::IntoIterator") as?RsTraitItem?: return@RsTopMostInScopeSelector false
+        val implLookup = ImplLookup.relativeTo(it)
+        implLookup.canSelect(TraitRef(it.type, iterTrait.withDefaultSubst()))
+        || implLookup.canSelect(TraitRef(it.type, intoIterTrait.withDefaultSubst()))
+    }, provider) {
+    override fun getTemplateString(element: PsiElement): String {
+        if(element !is RsExpr) return element.text
```
ortem: You should run `Reformat code` action on each changed file.

Voronchikhin:
I just reformatted code, and get build error with unclear log. It seems, that error was accidental.

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -61,3 +67,29 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
         }
     }
 }
+
+class IterPostfixTemplate(provider: RsPostfixTemplateProvider) :
```
ortem: IMO it's better to split it into two classes like this. @vlad20012 what do you think?
```
abstract class ForPostfixTemplate(provider: RsPostfixTemplateProvider, pred: (RsExpr) -> Boolean) :
    StringBasedPostfixTemplate("iter", "for x in expr", RsTopMostInScopeSelector(pred), provider) {

}

private val RsExpr.implementsIter: Boolean
    get() {
        val items = StdKnownItems.relativeTo(this)
        val implLookup = ImplLookup(project, items)
        val iterTrait = items.findIteratorTrait() ?: return false
        return implLookup.canSelect(TraitRef(type, iterTrait.withDefaultSubst()))
    }

private val RsExpr.implementsIntoIter: Boolean
    get() {
        val items = StdKnownItems.relativeTo(this)
        val implLookup = ImplLookup(project, items)
        val intoIterTrait = items.findIntoIteratorTrait() ?: return false
        return implLookup.canSelect(TraitRef(type, intoIterTrait.withDefaultSubst()))
    }

class IterPostfixTemplate(provider: RsPostfixTemplateProvider)
    : ForPostfixTemplate(provider, RsExpr::implementsIter) {

    override fun getTemplateString(element: PsiElement): String =
        "for x in ${element.text} {\n\$END$\n}"
}

class IntoIterPostfixTemplate(provider: RsPostfixTemplateProvider)
    : ForPostfixTemplate(provider, RsExpr::implementsIntoIter) {

    override fun getTemplateString(element: PsiElement): String =
        "for x in ${element.text}.into_iter() {\n\$END$\n}"
}
```

src/test/kotlin/org/rust/ide/template/postfix/IntoIterPostfixTemplateTest.kt
```diff
@@ -0,0 +1,36 @@
...
+package org.rust.ide.template.postfix
+
+import com.intellij.testFramework.LightProjectDescriptor
+import org.rust.WithStdlibRustProjectDescriptor
+
+
+class IntoIterPostFixTemplateTest : PostfixTemplateTest(IntoIterPostfixTemplate(RsPostfixTemplateProvider())) {
+    override fun getProjectDescriptor(): LightProjectDescriptor = WithStdlibRustProjectDescriptor
+
+    fun testNonIterable() = doTestNotApplicable(
```
Undin: We usually use the following code stuly for tests
```kotlin
fun `test name of test case`() = someIdentifier("""
    code before
""", """
    code after (if applicable)
""")
```

Yeah, there are test cases with old naming but we will convert them eventually

src/test/kotlin/org/rust/ide/template/postfix/IntoIterPostfixTemplateTest.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.rust.ide.template.postfix
+
+import com.intellij.testFramework.LightProjectDescriptor
+import org.rust.WithStdlibRustProjectDescriptor
+
+
+class IntoIterPostFixTemplateTest : PostfixTemplateTest(IntoIterPostfixTemplate(RsPostfixTemplateProvider())) {
+    override fun getProjectDescriptor(): LightProjectDescriptor = WithStdlibRustProjectDescriptor
```
Undin: Let's use `ProjectDescriptor` annotation to specify project descriptor
See example [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt#L12)

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -61,3 +66,39 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
...
+        return implLookup.canSelect(TraitRef(type, iterTrait.withDefaultSubst()))
+    }
+
+private val RsExpr.implementsIntoIter: Boolean
+    get() {
+        val items = StdKnownItems.relativeTo(this)
+        val implLookup = ImplLookup(project, items)
+        val intoIterTrait = items.findIntoIteratorTrait() ?: return false
+        return implLookup.canSelect(TraitRef(type, intoIterTrait.withDefaultSubst()))
+    }
```
Undin: `implementsIter` and `implementsIntoIter` look very similar
Maybe extract common code?

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -61,3 +68,37 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
...
+}
+
+
+private val RsExpr.implementsIntoIter: Boolean
+    get() = isImplementsTrait(this, "iter::IntoIterator")
+
+private val RsExpr.implementsIter: Boolean
+    get() = isImplementsTrait(this, "iter::Iterator")
+
+fun isImplementsTrait(expr: RsExpr, traitName: String, vararg subst: Ty): Boolean {
```
Undin: Should be private

Undin: You never pass `subst` so I think we can drop this param

src/test/kotlin/org/rust/ide/template/postfix/IntoIterPostfixTemplateTest.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package org.rust.ide.template.postfix
+
+import org.rust.ProjectDescriptor
+import org.rust.WithStdlibRustProjectDescriptor
+
+@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
+class IntoIterPostFixTemplateTest : PostfixTemplateTest(IntoIterPostfixTemplate(RsPostfixTemplateProvider())) {
+    fun `test non intoIterable`() = doTestNotApplicable(
+        """
```
Undin: Still wrong formatting
See #2833 as example

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -61,3 +68,37 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
...
+
+    override fun getTemplateString(element: PsiElement): String =
+        "for x in ${element.text} {\n\$END$\n}"
+}
+
+class IntoIterPostfixTemplate(provider: RsPostfixTemplateProvider)
+    : ForPostfixTemplate(provider, RsExpr::implementsIntoIter, "for x in expr.into_iter()") {
+
+    override fun getTemplateString(element: PsiElement): String =
+        "for x in ${element.text}.into_iter() {\n\$END$\n}"
```
Undin: Why do we need this explicit call of `into_iter` method?
Rust can call it [implicitly](https://doc.rust-lang.org/std/iter/index.html#for-loops-and-intoiterator)

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
Undin: It's not good idea because it leads to such situation
screen shot 2018-09-09 at 12 59 19
And due to [this](https://github.com/intellij-rust/intellij-rust/pull/2807#discussion_r216153042) I think it's worth to merge them again with the same template string

Undin:
LGTM!
Let's squash commits before merge

Undin:
Further improvements:
* Allow a user to modify loop variable like it's done in other similar postfix templates

* Support more stdlib types:
  * Arrays. Array doesn't implement `Iterator` and `IntoIterator`, so you should call `iter()` directly or use reference in loop (see [documentation](https://doc.rust-lang.org/std/primitive.array.html#examples))
  * Ranges. They implement `Iterator` trait but current implementation doesn't work for them. Probably, it's a bug in type inference

PR #2809, opened 2018-08-30
HINT: Inlay parameter type hints in struct literals implemented

dem1tris:

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
@@ -13,10 +13,8 @@ import com.intellij.psi.PsiElement
 import org.rust.ide.presentation.shortPresentableText
 import org.rust.ide.utils.CallInfo
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsFieldsOwner
 import org.rust.lang.core.psi.ext.descendantsOfType
 import org.rust.lang.core.psi.ext.elementType
-import org.rust.lang.core.psi.ext.namedFields
 import org.rust.lang.core.types.declaration
 import org.rust.lang.core.types.infer.substitute
 import org.rust.lang.core.types.ty.TyFunction
@@ -123,17 +121,17 @@ enum class HintType(desc: String, enabled: Boolean) {
     STRUCT_LITERAL_TYPE_HINT("Show type hints for fields in struct literal", true) {
         override fun provideHints(elem: PsiElement): List<InlayInfo> {
             val element = elem as? RsStructLiteral ?: return emptyList()
-            val declaration = element.path.reference.resolve() as? RsFieldsOwner ?: return emptyList()
             val subst = element.type.typeParameterValues
             val hints = element.structLiteralBody.structLiteralFieldList.mapNotNull { field ->
-                (declaration.namedFields
-                    .find { it.name.equals(field.referenceName) }
+                val text = (field.reference.resolve() as? RsFieldDecl)
                     ?.typeReference?.type?.substitute(subst)
                     ?.shortPresentableText
-                    ?: return@mapNotNull null) to (
-                    // in case of incomplete field, COLON will be the sibling, not the child
-                    field.colon ?: if (field.nextSibling.elementType == RsElementTypes.COLON) field.nextSibling
+                    ?: return@mapNotNull null
+                // in case of incomplete field, COLON will be the sibling, not the child
+                val offset =
+                    (field.colon ?: if (field.nextSibling.elementType == RsElementTypes.COLON) field.nextSibling
                     else field.identifier).textRange.endOffset
+                text to offset
             }
             return hints.map { InlayInfo(it.first, it.second) }
         }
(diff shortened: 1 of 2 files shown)


vlad20012:
Do we really need this? I think these hints are mostly redundant. We don't, for example, specially show function argument types in hints. In case of struct, argument names are already provided in syntax, and I think it's enough. 
image

dem1tris:
@vlad20012 I thought about this hint while creating literal of one dummy struct with couple of dummy fields with types I actually don't remember. And it seemed like "type something here, I won't tell what exactly, but I will highlight type mismatch, If you make a mistake". I suggested that we need an easier way than "go to declaration". By the way, it can be disabled, if redundant

vlad20012:
Compare it with a case of function call. We don't show types too. But you can ctrl+P on it and see popup with function signature.
image

dem1tris:
@vlad20012 Yes, but on struct literal we can't use Ctrl+P) Maybe it's worth turning it off by default? Or make it disappear, when correct value typed?

vlad20012:
Maybe it's better to make ctrl+P work with struct literals? =)

sirgl:
We can make an option to make user decide if he wants to see it or not with false default value.

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -112,6 +116,29 @@ enum class HintType(desc: String, enabled: Boolean) {
...
+    /* In case of fields without any expressions
+       PSI is destructed in this place, so only hints before error element will be shown */
+    STRUCT_LITERAL_TYPE_HINT("Show type hints for fields in struct literal", true) {
+        override fun provideHints(elem: PsiElement): List<InlayInfo> {
+            val element = elem as? RsStructLiteral ?: return emptyList()
+            val declaration = element.path.reference.resolve() as? RsFieldsOwner ?: return emptyList()
+            val subst = element.type.typeParameterValues
+            val hints = element.structLiteralBody.structLiteralFieldList.mapNotNull { field ->
+                (declaration.namedFields
+                    .find { it.name.equals(field.referenceName) }
```
vlad20012: You can use `it.reference.resolve() as? RsFieldDecl` instead

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -112,6 +116,29 @@ enum class HintType(desc: String, enabled: Boolean) {
...
 
         override fun isApplicable(elem: PsiElement): Boolean = elem is RsForExpr
+    },
+
+    /* In case of fields without any expressions
+       PSI is destructed in this place, so only hints before error element will be shown */
+    STRUCT_LITERAL_TYPE_HINT("Show type hints for fields in struct literal", true) {
+        override fun provideHints(elem: PsiElement): List<InlayInfo> {
+            val element = elem as? RsStructLiteral ?: return emptyList()
+            val declaration = element.path.reference.resolve() as? RsFieldsOwner ?: return emptyList()
```
vlad20012: ... and so `declaration` var will become unused

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -112,6 +116,29 @@ enum class HintType(desc: String, enabled: Boolean) {
...
+            val subst = element.type.typeParameterValues
+            val hints = element.structLiteralBody.structLiteralFieldList.mapNotNull { field ->
+                (declaration.namedFields
+                    .find { it.name.equals(field.referenceName) }
+                    ?.typeReference?.type?.substitute(subst)
+                    ?.shortPresentableText
+                    ?: return@mapNotNull null) to (
```
vlad20012: it is very unobvious construction. Let's extract left and right parts of `to` expression to variables.

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -112,6 +114,29 @@ enum class HintType(desc: String, enabled: Boolean) {
...
+            val hints = element.structLiteralBody.structLiteralFieldList.mapNotNull { field ->
+                val text = (field.reference.resolve() as? RsFieldDecl)
+                    ?.typeReference?.type?.substitute(subst)
+                    ?.shortPresentableText
+                    ?: return@mapNotNull null
+                // in case of incomplete field, COLON will be the sibling, not the child
+                val offset =
+                    (field.colon ?: if (field.nextSibling.elementType == RsElementTypes.COLON) field.nextSibling
+                    else field.identifier).textRange.endOffset
+                text to offset
```
vlad20012: There was the second comment. You can (I guess) write here directly InlayInfo(text, offset), and then just return `hints` without additional mapping

vlad20012:
I finally decided that we do not need it. We show parameter name hints at function calls, but we don't show their types. But if you want you can see parameter types via ctrl+P. In the case of struct fields, we have their names in syntax already, and I don't see much difference between function parameters and struct fields here. If names are already "shown", I think we don't have to show type hints in this special case. To be uniform we should provide ctrl+P on struct literals.

PR #2829, opened 2018-09-07
Cleanup

mchernyavsky:


Undin:
Also, some changes are rather controversial - they don't make code more readable or cleaner

vlad20012:
Let's do it a little later (after a personal discussion of contentious moments) and in a more incremental way, maybe starting with a simple reformatting.

PR #2832, opened 2018-09-08
Implement snapshotable collections

mchernyavsky:
SnapshotMap generalization; implementation of SnapshotList and SnapshotSet.
Use cases can be found in #2828.

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/snapshot/SnapshotList.kt
+++ b/src/main/kotlin/org/rust/lang/utils/snapshot/SnapshotList.kt
@@ -1,0 +1,27 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.utils.snapshot
+
+import org.rust.stdext.removeLast
+
+class SnapshotList<E> private constructor(
+    private val inner: MutableList<E>
+) : Snapshotable(), Collection<E> by inner {
+
+    constructor() : this(mutableListOf())
+
+    fun add(element: E): Boolean {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/rust/lang/utils/snapshot/SnapshotList.kt
```diff
@@ -0,0 +1,27 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.utils.snapshot
+
+import org.rust.stdext.removeLast
+
+class SnapshotList<E> private constructor(
+    private val inner: MutableList<E>
+) : Snapshotable(), Collection<E> by inner {
```
vlad20012: There is a leak. E.g if you remove an element from a list, this action will not be logged. Consider not implementing any standard interfaces an just write methods you need by yourself. Also make `inner` really "inner" without possibility to set it externally (like in SnapshotMap).
Same with SnapshotSet.

src/main/kotlin/org/rust/lang/utils/snapshot/SnapshotSet.kt
```diff
@@ -0,0 +1,25 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.utils.snapshot
+
+class SnapshotSet<E> private constructor(
+    private val inner: MutableSet<E>
+) : Snapshotable(), Collection<E> by inner {
+
+    constructor() : this(hashSetOf())
```
vlad20012: `hashSetOf()` returns `LinkedHashSet` which isn't needed here. Use `HashSet()` directly

PR #2840, opened 2018-09-11
Allow choosing which cargo project features are enabled

kumbayo:
Allow choosing per project which cargo project features are enabled:
- All features (`--all-features`)
- Default features + custom features
- No default features (`--no-default-features`) + custom features

Custom features are passed via `--features xxx`

The default is changed from `All features` to `Default features`.

This currently only influences calls to `cargo metadata` which influences crate resolution and thus available crates.

In a next step this information about features can be used for analysis/evaluation of `#[cfg(feature="foo")]` attributes (#2166)

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
+++ b/src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
@@ -9,12 +9,17 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.ComboBox
 import com.intellij.ui.EnumComboBoxModel
 import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.components.JBTextField
+import com.intellij.ui.components.Label
+import org.rust.cargo.project.settings.RustProjectSettingsService.FeaturesSetting
 import org.rust.cargo.toolchain.ExternalLinter
 import org.rust.cargo.util.CargoCommandLineEditor
 import org.rust.ide.ui.layout
 import org.rust.openapiext.CheckboxDelegate
 import org.rust.openapiext.ComboBoxDelegate
-import javax.swing.JComponent
+import java.awt.Component
+import javax.swing.*
+import kotlin.reflect.KProperty
 
 class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
 
@@ -34,6 +39,24 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
     private var compileAllTargets: Boolean by CheckboxDelegate(compileAllTargetsCheckBox)
 
     private lateinit var externalLinterArguments: CargoCommandLineEditor
+
+    private val cargoFeaturesComboBox = ComboBox<FeaturesSetting>().apply {
+        FeaturesSetting.values().forEach { addItem(it) }
+        setRenderer(object : DefaultListCellRenderer() {
+            override fun getListCellRendererComponent(list: JList<*>?, value: Any?, index: Int, isSelected: Boolean, cellHasFocus: Boolean): Component {
+                val value2 = when (value as FeaturesSetting) {
+                    FeaturesSetting.All -> "All"
+                    FeaturesSetting.Default -> "Default"
+                    FeaturesSetting.NoDefault -> "No default"
+                }
+                return super.getListCellRendererComponent(list, value2, index, isSelected, cellHasFocus)
+            }
+        })
+    }
+    private var cargoFeatures: FeaturesSetting by FeaturesComboDelegate()
+
+    private val cargoFeaturesAdditionalField = JBTextField()
+    private var cargoFeaturesAdditional: List<String> by FeaturesTextFieldDelegate()
 
     override fun getDisplayName(): String = "Cargo"
 
@@ -62,6 +85,20 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
                 Can be CPU-consuming.
             """)
         }
+        cargoFeaturesAdditionalField.toolTipText = "Additional features to activate. Space-separated."
+        cargoFeaturesAdditionalField.emptyText.text = "Additional features to activate. Space-separated."
+
+        val p = JPanel().apply {
+            layout = BoxLayout(this, BoxLayout.X_AXIS)
+            add(cargoFeaturesComboBox)
+            add(Label("+"))
+            add(cargoFeaturesAdditionalField)
+        }
+
+        row("Cargo metadata features:", p, """
+            Cargo features to enable.
+            Pass --all-features, --no-default-features and --features to `cargo metadata`.
+        """)
     }
 
     override fun isModified(): Boolean =
@@ -69,6 +106,8 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
             || externalLinter != settings.externalLinter
             || runExternalLinterOnTheFly != settings.runExternalLinterOnTheFly
             || compileAllTargets != settings.compileAllTargets
+            || cargoFeatures != settings.cargoFeatures
+            || cargoFeaturesAdditional != settings.cargoFeaturesAdditional
             || useOffline != settings.useOffline
             || externalLinterArguments.text != settings.externalLinterArguments
 
@@ -79,6 +118,8 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
             it.runExternalLinterOnTheFly = runExternalLinterOnTheFly
             it.externalLinterArguments = externalLinterArguments.text
             it.compileAllTargets = compileAllTargets
+            it.cargoFeatures = cargoFeatures
+            it.cargoFeaturesAdditional = cargoFeaturesAdditional
             it.useOffline = useOffline
         }
     }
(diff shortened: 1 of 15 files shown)


src/test/kotlin/org/rust/cargo/RustWithToolchainTestBase.kt
```diff
@@ -15,7 +15,7 @@ import org.rust.cargo.project.settings.rustSettings
 import org.rust.cargo.toolchain.RustToolchain
 
 // This class allows to execute real Cargo during the tests.
-// Unlike `RustTestCaseBase` it does not use in-memory temporary VFS
+// Unlike `RsTestBase` it does not use in-memory temporary VFS
 // and instead copies real files.
 abstract class RustWithToolchainTestBase : CodeInsightFixtureTestCase<ModuleFixtureBuilder<*>>() {
```
kumbayo: Would it be useful to also rename `RustWithToolchainTestBase` -> `RsWithToolchainTestBase` to have similar naming?

kumbayo:
Does this kind of preparation MR make sense or should i rather couple it with an actual feature like displaying the feature information in the Cargo view?

The current WIP state for that can be seen in #2166
image

kumbayo:
This MR is also somewhat related to #2951.
It currently does not influence run configurations but only `cargo metadata`.

kumbayo:
## Displaying features
The only other way i can imagine for displaying features on demand would be in a separate tab of the `Cargo` tool window similar to the `Project` viewer.

(This would allow more tabs for other features like duplicate crates, reverse dependencies, ...)
image

Views like `Gradle` also mix different things like `Source Sets` and `Tasks` so i think we can always display the `features` in the `Cargo` view for now.

## Icons/text
Without icons or without text, i find it a bit difficult to quickly see if a features is enabled or disabled.
So i am not sure which one to prefer :-/
This is how it looks like:
image

image


## Combo box
With the text field next to the combo box it looks like this:
image
(Not pushed yet as the combo box UI is not fully finished. I still need to figure out a nice way to display tooltips for each option)

amakeev:
@kumbayo 

>The only other way i can imagine for displaying features on demand would be in a separate tab of the Cargo tool window similar to the Project viewer.

That might be overkill; I though about a Quick doc for Cargo view, but it might require some work.

>Views like Gradle also mix different things like Source Sets and Tasks so i think we can always display the features in the Cargo view for now.

To me the difference is that in Gradle view a user interacts with tasks (executes them).
Anyways, for starters it's OK to show them, and collect feedback. (Maybe make them switchable)

>Without icons or without text, i find it a bit difficult to quickly see if a features is enabled or disabled.

Try graying the disabled features out (and probably showing the disabled ones in the end).

>(Not pushed yet as the combo box UI is not fully finished. I still need to figure out a nice way to display tooltips for each option

I mistakenly assumed that the the combo box value would make it cleat what the field is. Doesn't seem to be so; I guess it needs to be named after all. The common pattern is the following (indent the sub-control):
```
Cargo metadata features: [combobox]
   Additional features:  [text field]
```

kumbayo:
@amakeev 
I grayed out disabled features and sorted enabled features before disabled ones.
image

I left the text field in the same line as the combo box but placed a + in between the 2.
I would like to leave it this way for now. It can be changed later anyway if it turns out to be confusing.
image

I also added a gray explanation text when the field is initially empty.
image

amakeev:
> I grayed out disabled features and sorted enabled features before disabled ones.

Looks rather distinguishable to me now, what do you think?

> left the text field in the same line as the combo box but placed a + in between the 2.
> I also added a gray explanation text when the field is initially empty.

I like the idea!


Anyways, as the extended Cargo view is a separate feature from project settings, and the question still remains, if it makes sense to show this info to all users, could you please split the PR to separate parts? Mostly likely, the part about features could be useful for the ongoing tests run configurations support by @mchernyavsky

kumbayo:
> Looks rather distinguishable to me now, what do you think?
Yeah it looks good.

> Anyways, as the extended Cargo view is a separate feature from project settings, and the question still remains, if it makes sense to show this info to all users, could you please split the PR to separate parts? Mostly likely, the part about features could be useful for the ongoing tests run configurations support by @mchernyavsky (#3021)

Yes this was my intention.
I have already implemented most of the code necessary for this, but have waited with publishing since this one is not yet merged.

For the cargo view part i have created #3031.

kumbayo:
@amakeev @Undin Is there anything else to do for this MR?

amakeev:
@Undin I remember you had some thought/concerns about the UI

amakeev:
@Undin I remember you had some thoughts/concerns.

kumbayo:
@amakeev @Undin any remaining concerns?

Since the last version, I added tests to check if the project setting flags are passed to `cargo metadata`.
I also use `@MinRustcVersion("1.26.0")` on the tests as this is the rust version required to extract feature information.

Tests are part of #3031 since that adds a visualization/information dumping for the features.

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
```diff
@@ -31,6 +37,24 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
 
     private lateinit var cargoCheckArguments: CargoCommandLineEditor
 
+    private val cargoFeaturesComboBox = ComboBox<FeaturesSetting>().apply {
```
Undin: Hm, I'm not sure that project settings are a good place for such properties. When you have several separate cargo projects, you may want to have a separate feature set for each cargo project
I think it's ok as initial stage but eventually, we should provide per cargo project settings
What do you think about it?

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -92,6 +94,17 @@ interface CargoWorkspace {
         LIB, BIN, TEST, EXAMPLE, BENCH, UNKNOWN
     }
 
+    class Feature(
+        val name: String,
+        val state: FeatureState
```
Undin: Do we really need a state of feature?
Maybe it's enough to keep only enabled features?

src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -220,11 +245,30 @@ object CargoMetadata {
...
+                val resolveNode = project.resolve.nodes.find { node -> node.id == it.id }
+                if (resolveNode == null) {
+                    LOG.error("Could not find package with `id` '${it.id}' in `resolve` section of the `cargo metadata` output.")
+                }
+
+                val enabledFeatures = resolveNode?.features
+                val features = it.features.keys.map { feature ->
+                    val state = when {
+                        enabledFeatures == null -> CargoWorkspace.FeatureState.Unknown
+                        enabledFeatures.contains(feature) -> CargoWorkspace.FeatureState.Enabled
```
Undin: I'm not sure that it's a real issue but the size of `resolveNode.features` can be pretty big (for example, ~300 features in winapi 0.3). And in such cases `enabledFeatures.contains(feature)` for each feature can be relatively long if `enabledFeatures` is stored in list

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
kumbayo: Yeah, i did not think about the fact that we can attach more than 1 cargo project.
This makes the `additional features` field kind of unuseable when using multiple projects as passing a non-existing feature causes an error.

Do you have an idea where we could put a per Cargo project setting?

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
kumbayo: The initial implementation had just that.
But Rust versions below 1.26 do not give us feature information from `cargo metadata`.
So they currently display Unknown as state.
Not sure if treating all Unknown as Disabled is a good idea.

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
Undin: @kumbayo currently, I've come up with three solutions:
* open some dialog on some action (for example, double click) on cargo project node in cargo tool window. But looks like it's absolutely undiscoverable
* add special action to cargo tool window to open a dialog where we can show panel with all cargo projects and the corresponding feature settings for each cargo project
* add panel from the point above to cargo settings (`Preferences | Languages & Frameworks | Rust | Cargo`)

I'd prefer the second variant

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
Undin: We use 1.26 as the oldest rust version in our tests
Also, the latest rust [survey](https://blog.rust-lang.org/2018/11/27/Rust-survey-2018.html#rust-toolchain) says that almost all users use the latest stable or even nightly build
So, I think we can consider that 1.26 is the minimal supported version by the plugin at this moment

elichai:
Is anyone still working on this? 
It's quite hard to work on a project where I always need to uncomment a `#[cfg(...)]` just so it won't gray out the code.

Kobzol:
@elichai this PR is superseded by https://github.com/intellij-rust/intellij-rust/pull/5189. Speaking of which, @ortem what's the current status of #5189?

ortem:
@Kobzol Thank you for mentioning, #5189 is still under progress and was postponed due to other important tasks, but we're definitely going to finish it soon.

@elichai Just to clarify, do you mean **feature** attributes like `#[cfg(feature = my_feature)`? Note that cfg-attributes with non-feature predicates like `#[cfg(unix)]` should work fine.

elichai:
yes. more specifically `#[cfg(all(feature = "feature1", not(feature = "feature2")))]`

PR #2851, opened 2018-09-14
Template `iter`: allow a user to modify loop variable

Voronchikhin:

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
@@ -62,7 +62,7 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
 
     override fun getElementToRemove(expr: PsiElement): PsiElement = expr
 
-    override fun setVariables(template: Template, element: PsiElement): Unit {
+    override fun setVariables(template: Template, element: PsiElement) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -73,7 +76,12 @@ class IterPostfixTemplate(provider: RsPostfixTemplateProvider) :
     StringBasedPostfixTemplate("iter", "for x in expr",
         RsTopMostInScopeSelector { it.implementsIntoIter || it.implementsIter }, provider) {
     override fun getTemplateString(element: PsiElement): String =
-        "for x in ${element.text} {\n\$END$\n}"
+        "for \$name$ in ${element.text} {\n     \$END$\n}"
+
+    override fun setVariables(template: Template, element: PsiElement) {
+        val name = MacroCallNode(SuggestVariableNameMacro())
```
Undin: Unfortunately, we can't just copy kotlin implementation because `SuggestVariableNameMacro` is part of java support. So if IDE doesn't have java module (for example, CLion or PyCharm) it will fail with `NoClassDefFoundError` exception

Undin:
Cool!
Let's squash commits and merge it

Undin:
Thanks!
bors r+

Undin:
bors retry

PR #2859, opened 2018-09-16
INSP: don't search initial items for auto import in transitive dependencies

Undin:
* Introduce cargo project search scope
* Use this scope to exclude transitive dependencies' items from auto import candidates

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoPackageIndex.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoPackageIndex.kt
@@ -1,0 +1,63 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.project.model.impl
+
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Disposer
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.util.Consumer
+import com.intellij.util.indexing.LightDirectoryIndex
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -27,6 +31,9 @@ interface CargoWorkspace {
 
     val packages: Collection<Package>
     fun findPackage(name: String): Package? = packages.find { it.name == name }
+    fun findPackageForFile(file: VirtualFile): Package?
+    /** for internal usage only */
+    fun resetPackageIndex()
```
Undin: I don't like this method in public API but I can't come up with a better way to allow `CargoProjectsServiceImpl` call index reset

Undin:
@vlad20012 Done

src/main/kotlin/org/rust/cargo/project/model/impl/CargoPackageIndex.kt
```diff
@@ -0,0 +1,63 @@
...
+                }
+            })
+        }
+        indexDisposable = disposable
+    }
+
+    fun findPackageForFile(file: VirtualFile): Package? {
+        checkReadAccessAllowed()
+        val cargoProject = service.findProjectForFile(file) ?: return null
+        return indices[cargoProject]?.getInfoForFile(file)?.orElse(null)
```
vlad20012: Can we omit `?.orElse(null)` here?

Undin: But we need to extract `Package` object from `Optional` and looks like it's the simplest way to do it

PR #2862, opened 2018-09-17
RES: resolve references to extern crate root in 2018 edition

### LINKED ISSUE #2547: Type information for crates referenced in use without an extern crate is missing
```rust
// main.rs
use parking_lot::Mutex;

fn main {
  let lock = Mutex::new(());
}
```

IntelliJ can't find the `Mutex` type. If you add `extern crate parking_lot`, IntelliJ can find the type again. Note that in nightly (and stable very soon as of Rust 2018) `extern crate` is linted against and is superflous.

Undin:
Closes #2547

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsPath.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsPath.kt
@@ -9,18 +9,43 @@ import com.intellij.lang.ASTNode
 import com.intellij.psi.PsiElement
 import com.intellij.psi.stubs.IStubElementType
 import org.rust.lang.core.macros.RsExpandedElement
+import org.rust.lang.core.psi.RsElementTypes.*
 import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.tokenSetOf
 import org.rust.lang.core.resolve.ref.RsPathReference
 import org.rust.lang.core.resolve.ref.RsPathReferenceImpl
 import org.rust.lang.core.stubs.RsPathStub
 
+private val RS_PATH_KINDS = tokenSetOf(IDENTIFIER, SELF, SUPER, CSELF, CRATE)
+
 val RsPath.hasColonColon: Boolean get() = stub?.hasColonColon ?: (coloncolon != null)
-val RsPath.hasCself: Boolean get() = stub?.hasCself ?: (cself != null)
+val RsPath.hasCself: Boolean get() = kind == PathKind.CSELF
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -305,6 +305,10 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             val superMod = containingMod.`super`
             if (superMod != null && processor("super", superMod)) return true
             if (crateRoot != null && processor("crate", crateRoot)) return true
+            if (path.kind == PathKind.IDENTIFIER &&
+                path.containingCargoTarget?.edition == CargoWorkspace.Edition.EDITION_2018) {
```
matklad: We already have `target` in scope?

src/main/kotlin/org/rust/lang/core/psi/ext/RsPath.kt
```diff
@@ -9,20 +9,45 @@ import com.intellij.lang.ASTNode
...
+    val child = node.findChildByType(RS_PATH_KINDS)
+    return when (child?.elementType) {
+        IDENTIFIER -> PathKind.IDENTIFIER
+        SELF -> PathKind.SELF
+        SUPER -> PathKind.SUPER
+        CSELF -> PathKind.CSELF
+        CRATE -> PathKind.CRATE
+        else -> error("Malformed RsPath element: `$text`")
+    }
+}
```
matklad: I don't remember exactly, but if `<Type as Trait>::AssocType` thingies are also paths, this might explode?

Undin: Current grammar rule for this case is `Path ::= (Path '::' | '::' | TypeQual)? PathIdent PathTypeArguments?`

So, in this case `AssocType` is an identifie, `<Type as Trait>` is a `TypeQual` and this code returns `PathKind.IDENTIFIER`

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
matklad: Nvm, that target is completely unrelated 
https://github.com/intellij-rust/intellij-rust/pull/2862/files#diff-a2b72fdf4730d3704af99c0f7c4d6103R175 :)

Undin: Okay 😄

PR #2874, opened 2018-09-25
Fix completion for private items

### LINKED ISSUE #644: Don't suggest autocomplete for private functions
In the following code:

``` rust
mod my {
    // Items in modules default to private visibility.
    fn private_function() {
        println!("called `my::private_function()`");
    }
}

fn main() {
    my::
}
```

`intellij-rust` suggests `private_function` for autocomplete, but if a user were to select this, they get a compile error:

`error: function `private_function` is private`

I would think it would be best to omit private functions from the autocomplete.

There is also probably a corollary to `pub` vs. private `struct`s and `struct` members.

ortem:
Fixes #644

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsVisibility.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsVisibility.kt
@@ -6,7 +6,7 @@
 package org.rust.lang.core.psi.ext
 
 import org.rust.ide.icons.addVisibilityIcon
-import org.rust.lang.core.psi.RsVis
+import org.rust.lang.core.psi.*
 import javax.swing.Icon
 
 interface RsVisible : RsElement {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsVisibility.kt
```diff
@@ -22,3 +22,22 @@ fun RsVisibilityOwner.iconWithVisibility(flags: Int, icon: Icon): Icon =
         icon
     else
         icon.addVisibilityIcon(isPublic)
+
+fun RsVisible.isVisibleFrom(mod: RsMod): Boolean {
+    if (isPublic) return true
+
+    val elementMod = (if (this is RsMod) this.`super` else this.contextStrict()) ?: return true
+    // We have access to any item in any super module of `oMod`
```
vlad20012: `oMod` is renamed to `mod` here. Please, fix comments

src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
```diff
@@ -30,7 +30,13 @@ class RsPathReferenceImpl(
     override val RsPath.referenceAnchor: PsiElement get() = referenceNameElement
 
     override fun getVariants(): Array<out Any> =
-        collectCompletionVariants { processPathResolveVariants(ImplLookup.relativeTo(element), element, true, it) }
+        collectCompletionVariants {
+            processPathResolveVariants(ImplLookup.relativeTo(element),
```
vlad20012: Move `ImplLookup.relativeTo(element)` to a next line

src/main/kotlin/org/rust/lang/core/resolve/ref/RsMethodCallReferenceImpl.kt
```diff
@@ -31,7 +31,12 @@ class RsMethodCallReferenceImpl(
         val lookup = ImplLookup.relativeTo(element)
         val receiver = element.receiver.type
         return collectCompletionVariants {
-            processMethodCallExprResolveVariants(lookup, receiver, filterMethodCompletionVariants(it, lookup, receiver))
+            processMethodCallExprResolveVariants(lookup, receiver,
+                filterCompletionVariantsByVisibility(
+                    filterMethodCompletionVariants(it, lookup, receiver),
```
vlad20012: Maybe now we should rename `filterMethodCompletionVariants` to something like `filterMethodCompletionVariantsByTraitBounds` ?

Undin:
Also, it would be great to have tests when we can't resolve some item or when we don't get some item to completion list because it's private

Issue #1908, opened 2017-10-23
Handle include!(concat!(env!("OUT_DIR"), "/bindings.rs"));


jrmuizel:
The bindgen user manual suggests using:
`include!(concat!(env!("OUT_DIR"), "/bindings.rs"));`
as the way to use the generated code. IntelliJ seems to not follow this and thus gives no information/completion for the generated rust code.

Although `include` can take arbitrary compile-time args (or almost), the vast majority of real usages of `include` macro contains only:
* [x] string literals #3935
* [x] `concat` macro #4194
* [ ] `env!("OUT_DIR")` #4542

jrmuizel:
@matklad is this easy to fix? If it is I'd like to give it shot if I get some guidance on how to do it.

matklad:
That partial fix here should be more or less straightforward. Here's the place where we handle macros: https://github.com/intellij-rust/intellij-rust/blob/631a4e3a81b6fd5780a5cf0687605ce9c1319ed9/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt#L581. I think we need to hard code include here: 

1) calculate the path, assuming that OUTPUT_DIR is just target 
2) find the RsFile using findFileByIoFile function 
3) process all names from it

SOF3:
OUT_DIR is not just target. It is something that looks like `target\debug\build\cube-engine-abc40c94fb8a8059\out/protocol.rs`

micfan:
add use case in https://github.com/tower-rs/tower-grpc/issues/186

Riateche:
What's the state of this issue? It would be very nice to have auto-completion for generated code. It's one of the major pain points right now.

Undin:
@Riateche plugin already supports `include` with string literals like `include!("bindings.rs")` and also will support `concat` macro as argument (`include!(concat!("foo", "bindings.rs"))`) since next release.
`env("OUT_DIR")` is not supported yet

onelson:
I'm working with tonic (which I'm guessing is very closely related to tower-rs/tower-grpc#186 in terms of macro usage) and this feature doesn't seem to be working for me in v0.2.113.2150-193 in CLion 2019.3.1.

Wondering if this is a regression, or if I am just missing some config step. My project has several crates in it, but I don't see completions for the codegen module with or without a workspace defined at the project root.

Undin:
[...]

PR #2875, opened 2018-09-27
Expand the builtin include macro (WIP)

SimonAdameit:
I'd like to get code completion support for generated sources.

This is related to the issue: https://github.com/intellij-rust/intellij-rust/issues/1908

The approach taken here is to expand the builtin `include!` macro. When something similar is done for `concat!` and `env!` that should provide code completion for the usual code generation case.

Is this the right approach?

--
I'm new to Kotlin and the IntelliJ internals, so any hints with regards to code style etc. are also welcome.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/Include.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/Include.kt
@@ -1,0 +1,39 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.macros
+
+import org.rust.lang.core.psi.RsLitExpr
+import org.rust.lang.core.psi.RsMacroCall
+import org.rust.lang.core.psi.RsMacroExpr
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.ext.stringLiteralValue
+import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
+import java.nio.ByteBuffer
+
+fun expandInclude(call: RsMacroCall): List<RsExpandedElement>? {
+    val psiFactory = RsPsiFactory(call.project)
+    val args = call.includeMacroArgument?.expr ?: return null
+    val path = expandStringArgument(args) ?: return null
+    val file = call.containingFile.virtualFile.parent.findFileByRelativePath(path) ?: return null
+    val content = file.charset.decode(ByteBuffer.wrap(file.contentsToByteArray()))
+    return psiFactory.parseExpandedTextWithContext(call, content)
+}
+
+private tailrec fun expandStringArgument(args: RsExpandedElement, depth: Int = 0): String? {
+    if (depth > DEFAULT_RECURSION_LIMIT) return null
+    return when (args) {
+        is RsLitExpr -> args.stringLiteralValue
+        is RsMacroExpr -> {
+            val expansion = org.rust.lang.core.macros.expandMacro(args.macroCall).value.orEmpty()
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/macros/Include.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.ext.stringLiteralValue
+import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
+import java.nio.ByteBuffer
+
+fun expandInclude(call: RsMacroCall): List<RsExpandedElement>? {
+    val psiFactory = RsPsiFactory(call.project)
+    val args = call.includeMacroArgument?.expr ?: return null
+    val path = expandStringArgument(args) ?: return null
+    val file = call.containingFile.virtualFile.parent.findFileByRelativePath(path) ?: return null
```
vlad20012: I think we can access PSI directly without re-parsing: `file.toPsiFile(call.project)`.
(don't sure about it)

src/main/kotlin/org/rust/lang/core/macros/Include.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.rust.lang.core.psi.RsMacroCall
+import org.rust.lang.core.psi.RsMacroExpr
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.ext.stringLiteralValue
+import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
+import java.nio.ByteBuffer
+
+fun expandInclude(call: RsMacroCall): List<RsExpandedElement>? {
+    val psiFactory = RsPsiFactory(call.project)
+    val args = call.includeMacroArgument?.expr ?: return null
```
vlad20012: The tests are really failed on this line. It's because `RsIncludeMacroArgument` isn't stubbed. You can read about stubs in out [docs](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md#stubs). It's not too hard to force some element to be stubbed. Just follow some commit that do the same thing for another element, e.g. [this](https://github.com/intellij-rust/intellij-rust/commit/8513a547cdce4b898613d667094f2c4df6b8d3b1). The element that have to be stubbed - `IncludeMacroArgument` (in `RustParser.bnf`).

src/main/kotlin/org/rust/lang/core/macros/Include.kt
```diff
@@ -0,0 +1,39 @@
...
+    val content = file.charset.decode(ByteBuffer.wrap(file.contentsToByteArray()))
+    return psiFactory.parseExpandedTextWithContext(call, content)
+}
+
+private tailrec fun expandStringArgument(args: RsExpandedElement, depth: Int = 0): String? {
+    if (depth > DEFAULT_RECURSION_LIMIT) return null
+    return when (args) {
+        is RsLitExpr -> args.stringLiteralValue
+        is RsMacroExpr -> {
+            val expansion = org.rust.lang.core.macros.expandMacro(args.macroCall).value.orEmpty()
```
vlad20012: Hmm, what macros expected to be expanded here? Are we support `concat!()`?

src/main/kotlin/org/rust/lang/core/macros/Include.kt
vlad20012: Hmm, my proposal will work only if the following conditions are met:
1. context is not RsMacroExpr/RsPatMacro/RsMacroType
2. included file has ".rs" extension

So you solution seems to be more full

PR #2880, opened 2018-09-29
RES: exclude non-lib dependency targets from index

vlad20012:
Rust crates contain test/bench/example targets. If they are indexed
they will be shown in completion/resolve. By this commit we
exclude non-lib targets of project dependencies from index and
also exclude "tests" and "benches" dir from stdlib crates by name.
It would be nice to exclude it by target, but we currently
know nothing about stdlib targets.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/RsAdditionalLibraryRootsProvider.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/RsAdditionalLibraryRootsProvider.kt
@@ -20,11 +20,14 @@ import javax.swing.Icon
  * IDEA side of Cargo package from crates.io
  */
 class CargoLibrary(
-    val root: VirtualFile,
     private val name: String,
+    private val root: VirtualFile,
+    private val excluded: Set<VirtualFile>,
     private val isStd: Boolean
 ) : SyntheticLibrary(), ItemPresentation {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/project/workspace/RsAdditionalLibraryRootsProvider.kt
```diff
@@ -25,6 +25,9 @@ class CargoLibrary(
     private val isStd: Boolean
 ) : SyntheticLibrary(), ItemPresentation {
     override fun getSourceRoots(): Collection<VirtualFile> = listOf(root)
+    override fun getExcludedRoots(): Set<VirtualFile> =
+        if (isStd) root.findChild("tests")?.let { setOf(it) } ?: emptySet() else emptySet()
```
Undin: Some std crates contain also `benches` folder with benchmarks. Looks like that we should exclude them as well

Undin: Also, it would be great to mark out a content of each stdlib crate manually and pass such info into a package
But I think it's ok for now

vlad20012:
Done. Also I've excluded non-lib targets (just crate roots for now)

PR #2881, opened 2018-09-29
COMP: sort completion variants

vlad20012:
I've used this order:
1. Keywords
2. Enum variants
3. Struct fields
4. Inherited members (impls without traits)
5. Everything other ...
6. Macros


src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -20,18 +20,33 @@ import org.rust.lang.core.types.ty.TyUnknown
...
+            when {
+                owner is RsAbstractableOwner.Impl && owner.isInherent -> INHERENT_IMPL_MEMBER_PRIORITY
+                else -> DEFAULT_PRIORITY
+            }
+        }
+        else -> DEFAULT_PRIORITY
+    }
 
-    return base
+    return if (priority == DEFAULT_PRIORITY) base else base.withPriority(priority)
```
ortem: Is this check really necessary? If so, it could be moved into `LookupElementBuilder.withPriority`. Otherwise, use just `return base.withPriority(priority)`.

src/test/kotlin/org/rust/lang/core/completion/RsCompletionSortingTest.kt
```diff
@@ -0,0 +1,84 @@
...
+        for ((actual, e) in elements.zip(expected)) {
+            val (klass, name) = e
+            check(klass.isInstance(actual)) {
+                "Expected a ${klass.java.name}, found ${actual.javaClass}"
+            }
+            val actualName = actual.name
+            check(name == actualName) { "Expected $name got $actualName" }
+        }
+    }
+
```
ortem: Redundant line

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -20,18 +20,33 @@ import org.rust.lang.core.types.ty.TyUnknown
...
 const val KEYWORD_PRIORITY = 10.0
+private const val ENUM_VARIANT_PRIORITY = 4.0
+private const val FIELD_DECL_PRIORITY = 3.0
+private const val INHERENT_IMPL_MEMBER_PRIORITY = 2.0
+private const val DEFAULT_PRIORITY = 0.0
 private const val MACRO_PRIORITY = -0.1
 
 fun createLookupElement(element: RsElement, scopeName: String): LookupElement {
     val base = element.getLookupElementBuilder(scopeName)
         .withInsertHandler { context: InsertionContext, _ -> getInsertHandler(element, scopeName, context) }
```
ortem: Maybe `context: InsertionContext, _` -> `context, _`?

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
vlad20012: This check allows to get rid of unnecessary allocation if we don't change priority. I moved it to `withPriority`

vlad20012:
Done

PR #2882, opened 2018-09-29
GRD: build the rust lexer in addition to parser by IDEA run config

vlad20012:
We had problems with the fact that the lexer did not re-compile
by "Generate Parser" IDEA run configuration.

### DIFF
--- a/.idea/runConfigurations/Generate_Parser.xml
+++ b/.idea/runConfigurations/Generate_Parser.xml
@@ -10,6 +10,7 @@
       </option>
       <option name="taskNames">
         <list>
+          <option value=":generateRustLexer" />
           <option value=":generateRustParser" />
         </list>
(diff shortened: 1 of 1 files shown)


.idea/runConfigurations/Generate_Parser.xml
```diff
@@ -10,11 +10,11 @@
       </option>
       <option name="taskNames">
         <list>
-          <option value=":generateRustParser" />
+          <option value=":generateRustLexerAndParser" />
```
Undin: Maybe it's better to add `generateRustLexer` task into `taskNames` list to launch both gradle tasks instead of new task creation?

vlad20012: :+1: Done

PR #2896, opened 2018-10-04
Unify types of shorthand struct fields

msmorgan:
I need some help with this. I believe I have found the correct place to perform this type
inference, but I'm not sure how to proceed. I believe what needs to happen is to look up
a declaration that matches the struct field's name, and then if it matches, tell the type
inference system that the type of the variable is coercible to the field's type, but I
could not figure out how to do this.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -970,10 +970,19 @@ class RsFnInferenceContext(
     }
 
     private fun inferStructTypeArguments(literal: RsStructLiteral, typeParameters: Substitution) {
-        literal.structLiteralBody.structLiteralFieldList.mapNotNull { field ->
-            field.expr?.let { expr ->
-                val fieldType = field.type
-                expr.inferTypeCoercableTo(fieldType.substitute(typeParameters))
+        literal.structLiteralBody.structLiteralFieldList.filterNotNull().forEach { field ->
+            val fieldType = field.type.substitute(typeParameters)
+            val expr = field.expr
+
+            if (expr != null) {
+                expr.inferTypeCoercableTo(fieldType)
+            } else {
+                // Handle struct field shorthand by looking up the matching declaration in scope.
+                RsCodeFragmentFactory(field.project).createPath(field.referenceName, field)?.let { path ->
+                    val local = resolvePath(path, lookup).singleOrNull()?.element
+                    val ty = (local as? RsPatBinding)?.let { ctx.getBindingType(it) } ?: TyUnknown
+                    tryCoerce(ty, fieldType)
+                }
             }
         }
     }
--- a/src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
@@ -334,6 +334,16 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
         }
     """)
 
+    fun `test struct field shorthand`() = testExpr("""
+        enum E<T> { A, B(T) }
+        struct S<T> { item: E<T> }
+        fn foo() -> S<u8> {
+            let item = E::A;
+                     //^ E<u8>
+            S { item }
+        }
+    """)
+
     fun `test struct expr with 2 fields of same type 1`() = testExpr("""
         struct X;
         struct S<T> { a: T, b: T }


src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
```diff
@@ -334,6 +334,27 @@ class RsGenericExpressionTypeInferenceTest : RsTypificationTestBase() {
...
+        enum E<T> { A, B(T) }
+        struct S<T> { item: E<T> }
+        fn foo() -> S<u8> {
+            let item = E::A;
+                     //^ E<u8>
+            S { item }
+        }
+    """)
+
+    fun `test struct field longhand`() = testExpr("""
```
vlad20012: The test isn't needed

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -971,9 +971,22 @@ class RsFnInferenceContext(
 
     private fun inferStructTypeArguments(literal: RsStructLiteral, typeParameters: Substitution) {
         literal.structLiteralBody.structLiteralFieldList.mapNotNull { field ->
-            field.expr?.let { expr ->
-                val fieldType = field.type
-                expr.inferTypeCoercableTo(fieldType.substitute(typeParameters))
+            val fieldType = field.type.substitute(typeParameters)
+
+            if (field.expr != null) {
+                field.expr!!.inferTypeCoercableTo(fieldType)
```
vlad20012: Please, avoid `!!` in kotlin code. You can extract `field.expr` to a variable and then check `expr != null`

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -971,9 +971,22 @@ class RsFnInferenceContext(
...
+                // TODO: Unify the types of the struct field and the matching declaration
+                //   At this point, the type inference system is looking at a struct
+                //   field that does not have a colon and expr, meaning the shorthand
+                //   form was used. This means we must attempt to find an in-scope match
+                //   for the struct field's name, then infer that its declaration
+                //   is coercible to the field's type.
+
+                // How can I look up a variable that's in scope and shares a name with
+                // the struct field (the identifier below)?
+                val ident = field.identifier
```
vlad20012: I can imagine this solution. Note that I didn't tried to even compile it =)
```kotlin
val path = RsCodeFragmentFactory(project).createPath(field.referenceName, literal)
val ty = (resolvePath(path, lookup) as? RsPatBinding)?.let { bindings[it] } ?: TyUnknown
tryCoerce(ty, fieldType)
```

src/test/kotlin/org/rust/lang/core/type/RsGenericExpressionTypeInferenceTest.kt
msmorgan: Removed.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
msmorgan: Done.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
msmorgan: Thanks for the input! That general approach worked great.

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -970,10 +971,20 @@ class RsFnInferenceContext(
...
+        literal.structLiteralBody.structLiteralFieldList.filterNotNull().forEach { field ->
+            val fieldType = field.type.substitute(typeParameters)
+            val expr = field.expr
+
+            if (expr != null) {
+                expr.inferTypeCoercableTo(fieldType)
+            } else {
+                // Handle struct field shorthand by looking up the matching declaration in scope.
+                RsCodeFragmentFactory(field.project).createPath(field.referenceName, field)?.let { path ->
+                    resolvePath(path, lookup).forEach { bound ->
```
vlad20012: A path is *resolved* only if list of resolved element contains single element. 
```kotlin
val local = resolvePath(path, lookup).singleOrNull()?.element
```

msmorgan: Okay, I'll change that. If it contains multiple elements, what does that mean?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -102,7 +102,7 @@ class RsInferenceContext(
     val items: StdKnownItems
 ) : RsInferenceData {
     val fulfill: FulfillmentContext = FulfillmentContext(this, lookup)
-    private val bindings: MutableMap<RsPatBinding, Ty> = HashMap()
+    val bindings: MutableMap<RsPatBinding, Ty> = HashMap()
```
vlad20012: There is already a function `getBindingType`, so it's not necessary to make this field public...

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -592,6 +592,7 @@ class RsFnInferenceContext(
     private val lookup get() = ctx.lookup
     private val items get() = ctx.items
     private val fulfill get() = ctx.fulfill
+    private val bindings get() = ctx.bindings as Map<RsPatBinding, Ty>
```
vlad20012: ...and add this extra getter...

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -970,10 +971,19 @@ class RsFnInferenceContext(
...
+            val fieldType = field.type.substitute(typeParameters)
+            val expr = field.expr
+
+            if (expr != null) {
+                expr.inferTypeCoercableTo(fieldType)
+            } else {
+                // Handle struct field shorthand by looking up the matching declaration in scope.
+                RsCodeFragmentFactory(field.project).createPath(field.referenceName, field)?.let { path ->
+                    val local = resolvePath(path, lookup).singleOrNull()?.element
+                    val ty = (local as? RsPatBinding)?.let { bindings[it] } ?: TyUnknown
```
vlad20012: ...because `ctx.getBindingType(it)` can be used here directly

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
vlad20012: It means that we resolved to multiple declarations, that in general means that the path is NOT resolved and our (rust) code is invalid. E.g.
```rust
struct S;
struct S;
type T = S; // multi-resolve here. 
```
In the case of local variables the case of multi-resolve is mostly impossible (b/c of name shadowing), but let's follow general style and treat multi-resolve as it were not resolved

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
msmorgan: I didn't see that function before. Fixed!

vlad20012:
Looks good for me now! Thanks!
bors r+

msmorgan:
@vlad20012 Thank you for your patience and thorough review!

PR #2910, opened 2018-10-08
COMP: filter associated types from completion where they aren't allowed

vlad20012:
```rust
trait Tr { type Item; }
impl<T> Tr for T { type Item = (); }
struct S;
fn main() { let a: S::/*caret*/; }
```
In this case associated types aren't allowed ([E0223](https://doc.rust-lang.org/error-index.html#E0223)). I think we should remove them from completion, but resolve them as previously. It allows to simply implement E0223 analysis in future

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/completion/RsTypeAwareCompletionTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/completion/RsTypeAwareCompletionTest.kt
@@ -203,4 +203,16 @@ class RsTypeAwareCompletionTest : RsCompletionTestBase() {
         struct S;
         fn main() { let a: S::/*caret*/; }
     """)
+
+    fun `test associated type in explicit UFCS form`() = doSingleCompletion("""
+        trait Tr { type Item; }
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsTypeAwareCompletionTest.kt
```diff
@@ -195,4 +195,13 @@ class RsTypeAwareCompletionTest : RsCompletionTestBase() {
         struct S<A>(A);
         impl<B: Tr> S<B> { fn foo(self) -> <B as Tr>::Item/*caret*/ }
     """)
+
+
+    // https://doc.rust-lang.org/error-index.html#E0223
+    fun `test no completion for associated type of type other then type parameter`() = checkNoCompletion("""
```
vlad20012: > type of type other then type
😞😞😞

ortem:
bors rertry

Undin:
bors r=ortem

PR #2911, opened 2018-10-08
MACRO: fixed $crate metavar

vlad20012:


src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -205,6 +204,33 @@ class MacroExpander(val project: Project) {
...
+        val prefix = when (relativeDepth) {
+            null -> "::"
+            0 -> "self::"
+            else -> "super::".repeat(relativeDepth)
+        }
+        return prefix + name
+    }
+
+    private fun getRelativeCrateNameAndDepth(context: RsElement, externCrateMod: RsMod): Pair<String, Int?>? {
+        val externCrateWithDepth = context.containingMod.superMods.withIndex().mapNotNull { (index, superMod) ->
```
vlad20012: I copied this code from autoimport. I can't generalize somehow because autoimports version is too specialized

src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -205,6 +204,33 @@ class MacroExpander(val project: Project) {
                 )
             }
         }
+
+    /** Returns path from [context] to [def]'s crate */
+    private fun getPathToCrate(context: RsElement, def: RsMacro): String {
+        val defCrate = def.crateRoot ?: return ""
+        if (defCrate == context.crateRoot) return ""
+        val (name, relativeDepth) = getRelativeCrateNameAndDepth(context, defCrate) ?: return ""
+        val prefix = when (relativeDepth) {
```
Undin: As I understand `extern crate` item with `#[macro_use]` can be only in the crate root (https://doc.rust-lang.org/error-index.html#E0468)
So I think we can drop this complex logic with relative depth and significantly simplify `getPathToCrate` method

vlad20012: Fixed

PR #2913, opened 2018-10-09
Improved completion filtering by visibility

vlad20012:
1. Filter private fields
2. Filter `#[doc(hidden)]` items

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
@@ -150,8 +150,13 @@ fun processAllWithSubst(
 
 fun filterCompletionVariantsByVisibility(processor: RsResolveProcessor, mod: RsMod): RsResolveProcessor {
     return fun(it: ScopeEntry): Boolean {
-        val visible = it.element as? RsVisible ?: return processor(it)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -150,8 +150,10 @@ fun processAllWithSubst(
 
 fun filterCompletionVariantsByVisibility(processor: RsResolveProcessor, mod: RsMod): RsResolveProcessor {
     return fun(it: ScopeEntry): Boolean {
-        val visible = it.element as? RsVisible ?: return processor(it)
-        if (visible.isVisibleFrom(mod)) return processor(it)
-        return false
+        val element = it.element
+        if (element is RsVisible && !element.isVisibleFrom(mod)) return false
+        if (element is RsOuterAttributeOwner && element.queryAttributes.isDocHidden) return false
```
Undin: Are you sure that we should filter such items if they are from our crate?

vlad20012: Umm, yep, I remembered that in the case of macro we also check that it's in another module. I should add the same check here

vlad20012: Fixed

PR #2920, opened 2018-10-10
WIP InternalIterator API proposal

vlad20012:
Currently our name resolution API designed with internal iterator pattern. It means that any name resolution function looks like `forEach` function - it accepts consumer closure as a parameter and calls it for each resolved element. 
For example, `processMethodCallExprResolveVariants` function accepts several arguments, and the last of them is `(MethodResolveVariant) -> Boolean`. A closure passed as this parameter will be called for each resolved method.

The problem of this approach comes up when we start to use a complex logic around resolve results. For example, if we want to just add all elements to a list, we can write something like:
```kotiln
val list = mutableListOf()
processMethodCallExprResolveVariants(..., list::add)
```
But we want to filter results before adding them to list, then we should create closure:
```kotiln
processMethodCallExprResolveVariants(...) {
    if (it.isGoodForOurCase(someCapturedVar)) list.add(it) else false
}
```
After a while, we come into situation where we have a lot of filters and collectors and composing them. It results in a very bad code, e.g:
```kotlin
collectCompletionVariants {
    processMethodCallExprResolveVariants(lookup, receiver,
        filterCompletionVariantsByVisibility(
            filterMethodCompletionVariantsByTraitBounds(it, lookup, receiver),
            element.containingMod
        )
    )
}
```
What is `collectCompletionVariants`? It's a function that accepts something like `forEach` function and collects elements to a list (something like we wrote previously). `filterCompletionVariantsByVisibility` is a function that accepts a function and returns a function (sic!) that performs filtering. Same with `filterMethodCompletionVariantsByTraitBounds`. This code is very hard to read, but if you understand it once, it will be clear in the future. Furthermore, this code is very efficient.

Nevertheless, I think it's reasonable to make it more readable somehow, because last time this code was actively edited. It can imagine multiple solutions:
1. Rewrite Name Resolution engine (very long and complex task)
2. First collect elements to a list and then do any complex logic around Sequence (inefficient)
3. Make better function composition API, like platform's InternalIterator (still hard to read)
4. Make InternalIterator that looks pretty like Sequence, but iterates internally

This PR is a scratch implementation of 4th option. I didn't removed anything, just write new (more readable imo) code next to the old one in few places, so you can compare them. For example, previous awful listing is now looks like:
```kotlin
return InternalIterator.from<MethodResolveVariant> { processMethodCallExprResolveVariants(lookup, receiver, it) }
    .filterCompletionVariantsByVisibility(element.containingMod)
    .filterMethodCompletionVariantsByTraitBounds(lookup, receiver)
    .collectCompletionVariants()
```

I think it's REALLY much more readable.

Please, look at this and leave your comments.


matklad:
Historical note:

I think current internal-iterator based API is the third revision. At this point, I don' remember how the first version looked like, but the second one was build entirely on top of the `Sequence<T>` type. The code looked rather pretty, but was rather difficult to debug and profile: with `Sequence<T>` stack-traces  are pretty unreadable.

This stack traces problem was the motivation to switch to internal iteration API.

matklad:
Some thoughts (take with a pinch of salt, I haven't touched the code for some time):

## Though 1: writing ifs

```
processMethodCallExprResolveVariants(...) {
    if (it.isGoodForOurCase(someCapturedVar)) list.add(it)
    false
}
```

Doesn't look too bad actually, to my taste. Do we really *need* to extract the logic as a high-order-function of some kind?

In particular

```
collectCompletionVariants {
    processMethodCallExprResolveVariants(lookup, receiver,
        filterCompletionVariantsByVisibility(
            filterMethodCompletionVariantsByTraitBounds(it, lookup, receiver),
            element.containingMod
        )
    )
}
```

seems to be roughly the same amount of code as

```
collectCompletionVariants {
    processMethodCallExprResolveVariants(lookup, receiver,
        if isVisibleCompletionVariant(element) && satisifesMethodTraitBounds(element) {
            ...
        }
    )
}
```

That is, on the first blush it seems that it is hard to abstract over control flow but, at the same time, just writing out the control flow is not that bad, actually? Abstracting over logic just seems to work?

## Thought 2, stack traces concern

The internal iteration API looks nice, but it will make stack traces a bit harder to read. This will still be much easier then `Sequence` though, because the iterator will be pushing items to the closure, instead of being pulled via a `.next` method. OTOH, it now occurred to me that `Sequence` itself should support internal iteration: it has `forEach` method. Might be useful to compare stack traces of plain sequence, plain internal iterators, combinator-based internal iterators and sequence, driven by internal-iteratoion consumer methods, like `forEach`. 

## Thought 3, de-lazying

Long term, I would consider doing some combination of 1 & 2. Specifically, I think it might make sense to introduce some explicit scopes/descriptors objects, which store a list of names that are visible in a module, on the type, etc. Such descriptors are cached once per declaration, as opposed to current caching once per reference. These descriptors form "compiler API" (so are unit-tested) and don't directly correspond to any IDE features.

Completion & IDE goto definition are layered on top of the descriptors then, and use lazy Sequnece API.

PR #2928, opened 2018-10-11
Add bundled pretty-printers for LLDB

ortem:
Current version of [LLDB pretty-printer for Rust](https://github.com/rust-lang/rust/blob/master/src/etc/lldb_rust_formatters.py) is very primitive. For example, it doesn't support children-based representation of data (synthetic children).

I started to develop a new LLDB pretty-printer for Rust. This PR adds:
* `Vec`: size as summary and elements as synthetic children;
* `String` and `&str`: value as summary, also supports UTF-8;
* Regular structures, arrays, and slices: LLDB already supports it.

screen shot 2018-10-31 at 14 28 36
screen shot 2018-10-31 at 14 28 50

screen shot 2018-10-31 at 14 09 52

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
@@ -1,0 +1,37 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger
+
+import org.rust.openapiext.plugin
+
+
+const val LLDB_PP_LOOKUP: String = "lookup"
+val LLDB_PP_PATH: String get() = plugin().path.resolve("prettyPrinters/$LLDB_PP_LOOKUP.py").path
+
+enum class LLDBRenderers(private val description: String) {
+    NONE("No renderers"),
+    COMPILER("Rust compiler's renderers"),
+    BUNDLED("Bundled renderers");
+
+    override fun toString(): String = description
+
+    companion object {
+        val DEFAULT: LLDBRenderers = BUNDLED
+        fun fromIndex(index: Int): LLDBRenderers = LLDBRenderers.values().getOrElse(index) { DEFAULT }
+    }
+}
+
+enum class GDBRenderers(private val description: String) {
+    NONE("No renderers"),
+    COMPILER("Rust compiler's renderers");
+
+    override fun toString(): String = description
+
+    companion object {
+        val DEFAULT: GDBRenderers = COMPILER
+        fun fromIndex(index: Int): GDBRenderers = GDBRenderers.values().getOrElse(index) { DEFAULT }
+    }
+}
--- a/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettingsConfigurable.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettingsConfigurable.kt
@@ -6,33 +6,45 @@
 package org.rust.debugger.settings
 
 import com.intellij.openapi.options.SearchableConfigurable
-import com.intellij.ui.components.JBCheckBox
+import com.intellij.openapi.ui.ComboBox
 import com.intellij.ui.layout.panel
-import org.rust.openapiext.CheckboxDelegate
+import com.intellij.util.ui.UIUtil.ComponentStyle.SMALL
+import org.rust.debugger.GDBRenderers
+import org.rust.debugger.LLDBRenderers
 import javax.swing.JComponent
 
 class RsDebuggerSettingsConfigurable(
     private val settings: RsDebuggerSettings
 ) : SearchableConfigurable {
 
-    private val isRendersEnabledCheckbox: JBCheckBox = JBCheckBox("Enable Rust library renders")
-    private var isRendersEnabled: Boolean by CheckboxDelegate(isRendersEnabledCheckbox)
+    private val lldbRenderers = ComboBox<LLDBRenderers>().apply {
+        LLDBRenderers.values().forEach { addItem(it) }
+    }
+
+    private val gdbRenderers = ComboBox<GDBRenderers>().apply {
+        GDBRenderers.values().forEach { addItem(it) }
+    }
 
     override fun getId(): String = "Debugger.Rust"
     override fun getDisplayName(): String = DISPLAY_NAME
 
     override fun createComponent(): JComponent = panel {
-        row { isRendersEnabledCheckbox() }
+        row("LLDB renderers:") { lldbRenderers() }
+        row("GDB renderers:") { gdbRenderers() }
+        row { label("Changing these options will affect next debug session", gapLeft = 4, style = SMALL) }
     }
 
-    override fun isModified(): Boolean = isRendersEnabled != settings.isRendersEnabled
+    override fun isModified(): Boolean =
+        lldbRenderers.selectedIndex != settings.lldbRenderers.ordinal || gdbRenderers.selectedIndex != settings.gdbRenderers.ordinal
 
     override fun apply() {
-        settings.isRendersEnabled = isRendersEnabled
+        settings.lldbRenderers = LLDBRenderers.fromIndex(lldbRenderers.selectedIndex)
+        settings.gdbRenderers = GDBRenderers.fromIndex(gdbRenderers.selectedIndex)
     }
 
     override fun reset() {
-        isRendersEnabled = settings.isRendersEnabled
+        lldbRenderers.selectedIndex = settings.lldbRenderers.ordinal
+        gdbRenderers.selectedIndex = settings.gdbRenderers.ordinal
     }
 
     companion object {
--- a/prettyPrinters/providers.py
+++ b/prettyPrinters/providers.py
@@ -1,0 +1,142 @@
+from lldb import SBValue, SBType, SBError
+from lldb.formatters import Logger
+
+
+#################################################################################################################
+# This file contains two kinds of pretty-printers: summary and synthetic.
+#
+# Important classes from LLDB module:
+#   SBValue: the value of a variable, a register, or an expression
+#   SBType:  the data type; each SBValue has a corresponding SBType
+#
+# Summary provider is a function with the type `(SBValue, dict) -> str`.
+#   The first parameter is the object encapsulating the actual variable being displayed;
+#   The second parameter is an internal support parameter used by LLDB, and you should not touch it.
+#
+# Synthetic children is the way to provide a children-based user-friendly representation of the object's value.
+# Synthetic provider is a class that implements the following interface:
+#
+#     class SyntheticChildrenProvider:
+#         def __init__(self, SBValue, dict)
+#         def num_children(self)
+#         def get_child_index(self, str)
+#         def get_child_at_index(self, int)
+#         def update(self)
+#         def has_children(self)
+#         def get_value(self)
+#
+#
+# You can find more information and examples here:
+#   1. https://lldb.llvm.org/varformats.html
+#   2. https://lldb.llvm.org/python-reference.html
+#   3. https://lldb.llvm.org/python_reference/lldb.formatters.cpp.libcxx-pysrc.html
+#   4. https://github.com/llvm-mirror/lldb/tree/master/examples/summaries/cocoa
+################################################################################################################
+
+
+class DefaultSynthteticProvider:
+    def __init__(self, valobj, dict):
+        # type: (SBValue, dict) -> StdVecSyntheticProvider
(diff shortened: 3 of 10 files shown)


debugger/src/main/kotlin/org/rust/debugger/prettyPrinters/lldb_pretty_printer.py
```diff
@@ -0,0 +1,67 @@
...
+        except:
+            return -1
+
+    def get_child_at_index(self, index):
+        element_type_size = self.element_type.GetByteSize()
+
+        start_address = self.data_ptr.GetValueAsUnsigned()
+
+        address = start_address + index * element_type_size
+        element_val = self.data_ptr.CreateValueFromAddress("vec" + ("[%s]" % index), address, self.element_type)
```
Undin: As I understand here we create child item with name `vec[index]`
I think we can omit `vec` and use only `[index]` as element name as other debuggers do

debugger/src/main/kotlin/org/rust/debugger/prettyPrinters/lldb_pretty_printer.py
```diff
@@ -0,0 +1,67 @@
...
+        self.element_type = self.data_ptr.GetType().GetPointeeType()
+        self.element_type_size = self.element_type.GetByteSize()
+
+    def num_children(self):
+        return self.length
+
+    def get_child_index(self, name):
+        try:
+            return int(name.lstrip('[').rstrip(']'))
+        except:
```
matklad: catchall usually is considered a bad style. I thin it’s better to use name.isdigit check?

debugger/src/main/kotlin/org/rust/debugger/prettyPrinters/lldb_pretty_printer.py
matklad: And, if we use Vec, we don’t need +: “vec[%s]”%index

debugger/src/main/kotlin/org/rust/debugger/prettyPrinters/lldb_pretty_printer.py
```diff
@@ -0,0 +1,67 @@
...
+        start_address = self.data_ptr.GetValueAsUnsigned()
+
+        address = start_address + index * element_type_size
+        element_val = self.data_ptr.CreateValueFromAddress("vec" + ("[%s]" % index), address, self.element_type)
+        return element_val
+
+    def update(self):
+        self.element_type = self.data_ptr.GetPointeeType()
+        self.element_type_size = self.element_type.GetByteSize()
+        pass
```
matklad: No need for pass here

mchernyavsky:
Flake8's output:
```
lldb_pretty_printer.py:18:80: E501 line too long (113 > 79 characters)
lldb_pretty_printer.py:19:80: E501 line too long (98 > 79 characters)
lldb_pretty_printer.py:30:80: E501 line too long (94 > 79 characters)
lldb_pretty_printer.py:38:80: E501 line too long (89 > 79 characters)
lldb_pretty_printer.py:49:9: E722 do not use bare except'
lldb_pretty_printer.py:58:80: E501 line too long (112 > 79 characters)
```

Undin:
@matklad yep, we are going to do it

Undin:
@mchernyavsky I think the warning about 80 symbols per line doesn't really make sense. Looks like we are not going to modify this code via terminal so we can easily use restriction in 120 symbols

vlad20012:
Let's move lldb_pretty_printer.py from src/main/kotiln to src/main/resources. Kotlin directory should contains only kotlin sources

debugger/src/main/kotlin/org/rust/debugger/prettyPrinters/lldb_pretty_printer.py
ortem: Yes, and thank you for review! Now the PR is “work in progress”, so it would be really great if you could review it once again after it is ready.

debugger/src/main/kotlin/org/rust/debugger/UpdatePrettyPrintersComponent.kt
```diff
@@ -25,7 +25,7 @@ class UpdatePrettyPrintersComponent : ApplicationComponent, Disposable {
         val application = ApplicationManager.getApplication()
         if (application.isUnitTestMode) return
         application.executeOnPooledThread {
-            checkPrettyPrinters()
+            checkPrettyPrinters(true)
```
matklad: I personally find it more readable, (especially on GitHub, where there are no parameter hints), when `true` and `false` arguments use named parameter syntax: `checkPretyPrinters(clearOld = true)` ;)

ortem:
@Undin thank you for the review! Now it should work fine again.

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettingsConfigurable.kt
```diff
@@ -6,33 +6,48 @@
...
+import org.rust.debugger.DataFormatters
 import javax.swing.JComponent
 
 class RsDebuggerSettingsConfigurable(
     private val settings: RsDebuggerSettings
 ) : SearchableConfigurable {
 
-    private val isRendersEnabledCheckbox: JBCheckBox = JBCheckBox("Enable Rust library renders")
-    private var isRendersEnabled: Boolean by CheckboxDelegate(isRendersEnabledCheckbox)
+    private val isLLDB = CPPToolchains.getInstance().defaultToolchain?.debuggerKind?.isLLDB() == true
```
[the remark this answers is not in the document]
amakeev: Using `CPPToolchainsConfigurable.TOPIC` is a rather tricky business, because of the lifecycle of the configurables. I'd suggest not to go that way. 
If you absolutely have to separate GDB/LLDB settings, it will be much easier to show both settings groups (for GDB and for LLDB)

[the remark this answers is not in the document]
Undin: @amakeev Oh, as you are much more aware of CLion related things, let's do it in this way!
> If you absolutely have to separate GDB/LLDB settings

Maybe there is a better solution, but the current one allows us to release own pretty printers partially. Also, we still use formatters from rustup distribution as defaut variant for now.
We'll merge these settings and make bundled printers default when we implement all basic features available with formatters from rustup distribution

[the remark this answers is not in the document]
amakeev: Just in case: you're comment about `A user can change default toolchain` is totally valid. I meant that making the correct listener between two configurables is not straightforward.

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettingsConfigurable.kt
```diff
@@ -6,33 +6,47 @@
...
 
 class RsDebuggerSettingsConfigurable(
     private val settings: RsDebuggerSettings
 ) : SearchableConfigurable {
 
-    private val isRendersEnabledCheckbox: JBCheckBox = JBCheckBox("Enable Rust library renders")
-    private var isRendersEnabled: Boolean by CheckboxDelegate(isRendersEnabledCheckbox)
+    private val lldbDataFormatters = ComboBox<DataFormatters>().apply {
+        DataFormatters.values()
+            .sortedBy { it.ordinal }
```
Undin: Do we really need to sort them?

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettingsConfigurable.kt
```diff
@@ -6,33 +6,47 @@
...
+            .forEach { addItem(it) }
+    }
 
     override fun getId(): String = "Debugger.Rust"
     override fun getDisplayName(): String = DISPLAY_NAME
 
     override fun createComponent(): JComponent = panel {
-        row { isRendersEnabledCheckbox() }
+        row("LLDB data formatters") { lldbDataFormatters() }
+        row("GDB data formatters") { gdbDataFormatters() }
```
Undin: We usually add ":" after labels

debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.rust.openapiext.PLUGIN
+
+
+const val LLDB_PP_LOOKUP: String = "lookup"
+val LLDB_PP_PATH: String get() = PLUGIN.path.resolve("prettyPrinters/$LLDB_PP_LOOKUP.py").path
+
+enum class DataFormatters(private val description: String) {
+    NONE("No data formatters"),
+    COMPILER("Rust compiler's data formatters"),
+    BUNDLE("Bundled data formatters");
```
Undin: Looks like `BUNDLED` is more suitable here

debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.rust.openapiext.PLUGIN
+
+
+const val LLDB_PP_LOOKUP: String = "lookup"
+val LLDB_PP_PATH: String get() = PLUGIN.path.resolve("prettyPrinters/$LLDB_PP_LOOKUP.py").path
+
+enum class DataFormatters(private val description: String) {
+    NONE("No data formatters"),
+    COMPILER("Rust compiler's data formatters"),
+    BUNDLED("Bundled data formatters");
```
Undin: Hm, should we rename `formatters` to `renders`?

debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    override fun toString(): String = description
+
+    companion object {
+        val default: DataFormatters = COMPILER
+
+        fun fromIndex(index: Int): DataFormatters =
+            DataFormatters.values().find { it.ordinal == index } ?: default
```
Undin: `DataFormatters.values()[index]` or `DataFormatters.values().getOrElse(index) { default }`?

prettyPrinters/providers.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def StdStringSummaryProvider(valobj, dict):
+    # type: (SBValue, dict) -> str
+    assert valobj.GetNumChildren() == 1
+
+    vec = valobj.GetChildAtIndex(0)
+    length = vec.GetNumChildren()
+    chars = [chr(vec.GetChildAtIndex(i).GetValueAsUnsigned()) for i in range(length)]
+    return "".join(chars)
```
Undin: Now we have different reperesentation for `&str` and `String`
screen shot 2018-10-31 at 00 05 41
So I think we should wrap this value into quotes

src/main/kotlin/org/rust/openapiext/utils.kt
```diff
@@ -125,3 +128,7 @@ inline fun <T> UserDataHolderEx.getOrPutSoft(key: Key<SoftReference<T>>, default
         val value = defaultValue()
         putUserDataIfAbsent(key, SoftReference(value)).get() ?: value
     }
+
+const val PLUGIN_ID: String = "org.rust.lang"
+
+val PLUGIN: IdeaPluginDescriptor get() = PluginManager.getPlugin(PluginId.getId(PLUGIN_ID))!!
```
Undin: It still looks like a constant
I'd prefer something like `fun plugin(): IdeaPluginDescriptor = PluginManager.getPlugin(PluginId.getId(PLUGIN_ID))!!`

debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
Undin: If you rename it, don't forget to rename labels in settings

ortem:
@Undin thank you for attention! Now scalar types render correctly.

debugger/src/main/kotlin/org/rust/debugger/PrettyPrinters.kt
```diff
@@ -0,0 +1,37 @@
...
+package org.rust.debugger
+
+import org.rust.openapiext.plugin
+
+
+const val LLDB_PP_LOOKUP: String = "lookup"
+val LLDB_PP_PATH: String get() = plugin().path.resolve("prettyPrinters/$LLDB_PP_LOOKUP.py").path
+
+enum class LLDBRenderers(private val description: String) {
+    NONE("No renders"),
```
Undin: Looks like it should be `No renderers`

PR #2932, opened 2018-10-12
MACRO: some fixes in macro expansion

vlad20012:
These fixes allows to almost completely expand `table!` macro from diesel.

Note that this PR brings extremely awful hacks around $crate metavar expansion.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -120,7 +120,7 @@ class MacroExpander(val project: Project) {
             subst,
             WithParent(
                 MacroSubstitution(
-                    singletonMap("crate", getPathToCrate(call, def)),
+                    singletonMap("crate", expandDollarCrateVar(call, def)),
                     emptyList()
                 ),
                 null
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -204,19 +209,39 @@ class MacroExpander(val project: Project) {
...
-        if (externCrateMod == contextCrate) return ""
-        val externCrateItem = contextCrate.childrenOfType<RsExternCrateItem>()
-            .find { it.reference.resolve() == externCrateMod } ?: return ""
-        val name = externCrateItem.nameWithAlias
-        return "::$name"
-    }
+/**
+ * Returns (synthetic) path from [call] to [def]'s crate
+ * We can't just expand `$crate` to something like `::crate_name` because
+ * rustc always threat result of this expansion as a single identifier. E.g.
```
Undin: Let's explain it more detailed.
It's not clear that we can pass `$crate` as an identifier but it's the main reason why you have to do such hacks.

Undin: Also, maybe use a better verb than `threat` here?

src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -184,15 +184,20 @@ class MacroExpander(val project: Project) {
...
-                    sb.append(child.text)
-                    sb.append(" ")
+                    sb.safeAppend(child.text)
                 }
             }
         }
         return true
     }
 
+    private fun StringBuilder.safeAppend(str: String) {
```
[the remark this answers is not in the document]
Undin: `safeAppend` name is not worse than `appendWrapped` or `appendWrappedWithWhitespace` because all these variants don't say that this method does.
Due to I don't believe that there is a better name I just suggest to add some short explanation in comment or javadoc

PR #2935, opened 2018-10-12
INSP: Fixed unresolved trait method if trait is aliased

### LINKED ISSUE #2863: Invalid error: unresolved reference
## Environment

* **Intellij-Rust plugin version: 0.2.0.2422-182-nightly**
* **Rust toolchain version: rustc 1.29.0 (aa3ca1994 2018-09-11)**
* **IDE name and version: IntelliJ IDEA Ultimate 2018.2.3**
* **Operating system: Ubuntu 16.04 LTS**

## Problem description
Invalid error for method invocation of a trait bound to a new name.

## Steps to reproduce
```rust
mod x {
    pub trait X {
        fn x(&self);
    }
    pub struct XS {}
    impl X for XS {
        fn x(&self) {}
    }
}
use x::XS;
use x::X as _X;
impl XS {
    fn y(&self) {
        self.x()
    }
}
fn main() {
}
```
Unresolved reference: `x`

vlad20012:
Fixes #2863

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -20,7 +20,6 @@ import org.rust.cargo.util.AutoInjectedCrates
 import org.rust.ide.search.RsCargoProjectScope
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
-import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.resolve.TraitImplSource
 import org.rust.lang.core.resolve.ref.MethodResolveVariant
 import org.rust.lang.core.resolve.ref.deepResolve
@@ -217,16 +216,13 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
             methodCall: RsMethodCall,
             resolveResults: List<MethodResolveVariant>
         ): List<RsTraitItem>? {
-            val lookup = ImplLookup.relativeTo(methodCall)
-
-            val traitsToImport = mutableListOf<RsTraitItem>()
-            loop@ for (variant in resolveResults) {
+            val traits = resolveResults.mapNotNull { variant ->
                 val source = variant.source
                 val trait = when (source) {
                     is TraitImplSource.ExplicitImpl -> {
                         val impl = source.value
                         if (impl.traitRef == null) return null
-                        impl.traitRef?.resolveToTrait ?: continue@loop
+                        impl.traitRef?.resolveToTrait ?: return@mapNotNull null
                     }
                     is TraitImplSource.Derived -> source.value
                     is TraitImplSource.Collapsed -> source.value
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsItemElement.kt
```diff
@@ -6,5 +6,10 @@
 package org.rust.lang.core.psi.ext
 
 import org.rust.lang.core.macros.RsExpandedElement
+import org.rust.lang.core.resolve.TYPES
+import org.rust.lang.core.resolve.processNestedScopesUpwards
 
 interface RsItemElement : RsVisibilityOwner, RsOuterAttributeOwner, RsExpandedElement
+
+fun RsItemElement.isInScope(scope: RsElement): Boolean =
```
Undin: In both case (in type inference and in auto import quick fix) we have the same situation: we have a list of traits and should skip all element that is not in scope. And we call `processNestedScopesUpwards` on each trait. But for all items, we have the same scope so all results of `processNestedScopesUpwards` are the same as well.
Maybe we modify `isInScope` method to call it on list of items, invoke `processNestedScopesUpwards` only one time and just check that elements are in the list? Looks like it's more efficient that is rather important in type inference case

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -236,10 +232,8 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
                     is TraitImplSource.Object -> return null
                 }
 
-                if (lookup.isTraitVisibleFrom(trait, methodCall)) return null
-                traitsToImport += trait
-            }
-            return traitsToImport
+                trait
+            }.filterInScope(methodCall)
```
Undin: Here is the main wrong place
You need something like the following:
```kotlin
val traits = resolveResults.mapNotNull { variant -> ... }
return if (traits.filterInScope(methodCall).isNotEmpty()) null else traits
```
If there is at least one trait in scope, we should return `null` because it means that method can be resolved by rustc, otherwise return all traits as initial import candidates

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -871,6 +871,8 @@ private fun processNestedScopesUpwards(
         if (processLexicalDeclarations(scope, cameFrom, ns, withPlainExternCrateItems, shadowingProcessor)) return@walkUp true
         prevScope.addAll(currScope)
         false
+    }) {
+        return false
```
Undin: Hm, looks like we should return `true` here because a successful result of `if (walkUp())` means that `processor` found suitable item and whole call of `processNestedScopesUpwards` is successful.
Otherwise, all callee would consider that they should continue to search suitable items

src/main/kotlin/org/rust/stdext/Collections.kt
```diff
@@ -6,6 +6,7 @@
 package org.rust.stdext
 
 import java.util.*
+import kotlin.collections.ArrayList
```
Undin: Looks like redundant import

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1124,12 +1124,20 @@ class RsFnInferenceContext(
...
+            val inherent = mutableListOf<MethodResolveVariant>()
+            for (callee in list) {
+                val trait = callee.source.impl?.traitRef?.resolveToTrait
+                if (trait != null) {
+                    traitToCallee.getOrPut(trait) { mutableListOf() }.add(callee)
+                } else {
+                    inherent.add(callee)
+                }
+            }
+            inherent + traitToCallee.keys.filterInScope(methodCall).map(traitToCallee::getValue).flatten()
```
Undin: It seems the following implementation is little more efficient
```kotlin
traitToCallee.keys.filterInScope(methodCall).forEach {
    inherent += traitToCallee.getValue(it)
}
inherent
```
But of course, in this case we should rename `inherent` list

PR #2941, opened 2018-10-14
INSP: Adjust auto import to edition 2018

Undin:
* fix extern crate name resolution in relative paths in edition 2018
* adjust auto import to edition 2018
  * don't add `extern crate` item while importing items from other crates
  * add `crate::` prefix for local absolute paths


vlad20012:
:+1: 
bors r+

PR #2948, opened 2018-10-15
MISC: Excluding radio buttons for "New Project"-Template #2885

### LINKED ISSUE #2885: Excluding radio buttons for "New Project"-Template
When creating a new Rust project there is currently a checkbox:

> Use a binary (application) template: [✓]

If one does uncheck this checkbox, the library template is used. This behaviour is not communicated by the application right now. I just hoped it would work that way. It did and that is a good thing ^^

Nevertheless I think it would be more intuitive to have excluding radio buttons instead:

> ⦿ Use a binary (application) template.
> ◯ Use a library template.

This way it is clear which result is achieved when selecting an option. If in the future there is another template option, one could easily append the new template option. Because people are used to excluding radio buttons, it would be also clear that chosing a template option excludes all other template options.

Thankyou for an already amazing experience =D I am currently using CLion.

amakeev:
Suggestions for better controls naming are welcome.

screenshot 2018-10-15 15 33 26

Fixes #2885

### DIFF
--- a/src/main/kotlin/org/rust/ide/newProject/ui/RsNewProjectPanel.kt
+++ b/src/main/kotlin/org/rust/ide/newProject/ui/RsNewProjectPanel.kt
@@ -8,11 +8,13 @@ package org.rust.ide.newProject.ui
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.options.ConfigurationException
 import com.intellij.openapi.util.Disposer
-import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.components.JBRadioButton
 import org.rust.cargo.project.settings.ui.RustProjectSettingsPanel
 import org.rust.ide.newProject.ConfigurationData
 import org.rust.ide.ui.RsLayoutBuilder
+import java.awt.event.ItemEvent
 import java.awt.event.ItemListener
+import javax.swing.ButtonGroup
 
 class RsNewProjectPanel(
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/newProject/ui/RsNewProjectPanel.kt
```diff
@@ -7,31 +7,38 @@ package org.rust.ide.newProject.ui
...
     private val showProjectTypeCheckbox: Boolean,
     updateListener: (() -> Unit)? = null
 ) : Disposable {
 
     private val rustProjectSettings = RustProjectSettingsPanel(updateListener = updateListener)
-    private val createBinaryCheckbox = JBCheckBox(null, true)
-    private val checkboxListener = ItemListener { updateListener?.invoke() }
+    private val templateButtons = listOf(JBRadioButton("Binary (application)", true), JBRadioButton("Library"))
+    private val checkboxListener = ItemListener {
+        if (it.stateChange == java.awt.event.ItemEvent.SELECTED) updateListener?.invoke()
```
vlad20012: Hmm, why not import `java.awt.event.ItemEvent`?

amakeev: No reason, just failed 'shorten FQNs feature'

PR #2950, opened 2018-10-15
MISC: Cargo.toml does not properly open as project in IDEA #2835

### LINKED ISSUE #2835: Cargo.toml does not properly open as project in IDEA
## Environment

* **Rust toolchain version:** cargo 1.28.0
* **IDE name and version:** IntelliJ IDEA 2018.3 EAP
* **Operating system:** Linux

## Problem description

When I try to open project from file `Cargo.toml`, a new project is really created, but it shows only this file (`Cargo.toml`) instead of the complete file structure.

After closing the project and reopening it again, IDEA reports that the project does not exist; and it is probably right, because the path stored contains `Cargo.toml` instead of the last path element.

### LINKED ISSUE #2949: After opening a cargo project, it should be configured automatically
Currently, when I open an existing project (pointing to a directory; as opposed to a Cargo.toml #2835), source folders, libraries etc. are not configured until I explicitly open Rust files or Cargo project view.

Expected: upon opening, the project is configured by the IDE automatically

amakeev:
Fixes #2835

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/CargoProjectOpenProcessor.kt
+++ b/src/main/kotlin/org/rust/cargo/project/CargoProjectOpenProcessor.kt
@@ -1,0 +1,32 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.project
+
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.startup.StartupManager
+import com.intellij.openapi.util.io.FileUtil
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.platform.PlatformProjectOpenProcessor
+import com.intellij.projectImport.ProjectOpenProcessor
+import org.rust.cargo.CargoConstants
+import org.rust.cargo.icons.CargoIcons
+import org.rust.cargo.project.model.guessAndSetupRustProject
+import javax.swing.Icon
+
+class CargoProjectOpenProcessor : ProjectOpenProcessor() {
+    override fun getIcon(): Icon? = CargoIcons.ICON
+    override fun getName(): String = "Cargo"
+
+    override fun canOpenProject(file: VirtualFile?): Boolean = FileUtil.namesEqual(file?.name, CargoConstants.MANIFEST_FILE)
+
+    override fun doOpenProject(virtualFile: VirtualFile, projectToClose: Project?, forceNewFrame: Boolean): Project? {
+        val basedir = if (virtualFile.isDirectory) virtualFile else virtualFile.parent
+
+        return PlatformProjectOpenProcessor.getInstance().doOpenProject(basedir, projectToClose, forceNewFrame)?.also {
+            StartupManager.getInstance(it).runWhenProjectIsInitialized { guessAndSetupRustProject(it) }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


vlad20012:
@Undin Should we do the same with Xargo.toml?

src/main/kotlin/org/rust/cargo/project/CargoProjectOpenProcessor.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.rust.cargo.icons.CargoIcons
+import org.rust.cargo.project.model.guessAndSetupRustProject
+import javax.swing.Icon
+
+class CargoProjectOpenProcessor : ProjectOpenProcessor() {
+    override fun getIcon(): Icon? = CargoIcons.ICON
+    override fun getName(): String = "Cargo"
+
+    override fun canOpenProject(file: VirtualFile?): Boolean = isCargoToml(file)
+    private fun isCargoToml(file: VirtualFile?) = FileUtil.namesEqual(file?.name, CargoConstants.MANIFEST_FILE)
```
Undin: Maybe inline this method? Currently, we use it only in one place

Undin: But if you think that it really simplify code, let's specify return type of `isCargoToml` explicitly

src/main/kotlin/org/rust/cargo/project/CargoProjectOpenProcessor.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    override fun doOpenProject(virtualFile: VirtualFile, projectToClose: Project?, forceNewFrame: Boolean): Project? {
+        val basedir = if (virtualFile.isDirectory) virtualFile else virtualFile.parent
+
+        val project = PlatformProjectOpenProcessor.getInstance().doOpenProject(basedir, projectToClose, forceNewFrame)
+        project?.let { StartupManager.getInstance(it).runWhenProjectIsInitialized { guessAndSetupRustProject(it) } }
+
+        return project
```
Undin: Maybe it's a little readable?
```kotlin
val project = PlatformProjectOpenProcessor.getInstance()
    .doOpenProject(basedir, projectToClose, forceNewFrame) ?: return null
StartupManager.getInstance(project).runWhenProjectIsInitialized { guessAndSetupRustProject(project) }

return project
```

Undin:
About `Xargo.toml` - as I understand `Xargo.toml` doesn't replace `Cargo.toml`, it just adds some specific properties for `xargo`, so I think we can take into account only `Cargo.toml` for now and implement xargo specific behavior in the future if we really need it

src/main/kotlin/org/rust/cargo/project/CargoProjectOpenProcessor.kt
amakeev: inlined

src/main/kotlin/org/rust/cargo/project/CargoProjectOpenProcessor.kt
amakeev: simplified

Undin:
bors r-
Let's merge it after release

Undin:
bors r=amakeev

PR #2952, opened 2018-10-16
Fix name suggestions (freshen default name)

### LINKED ISSUE #2919: "Extract variable" refactoring renames usages of an accidentally name-conflicting variable
## Environment

Latest git (b1ce136d7332a2b031688e00443625ce076ed904)
* **Rust toolchain version:**
rustc 1.31.0-nightly (2bd5993ca 2018-10-02)
* **IDE name and version:**
Idea Community 2018.2.4 (IC-182.4505.22)
* **Operating system:**
Linux x86-64

## Problem description
When selecting an expression and executing "Extract variable" refactoring, a new variable with the generic name, e.g. "x" is introduced and its name is highlighted in such way, that editing it also changes its name at the point of usage, where there has been the extracted expression. 

But if a variable with such name have already existed, any typing attempt changes all its usages too, not just at the place of extracted expression.  

The refactoring apparently tries to avoid that by choosing "i" or "x", whichever is currently unoccupied, but if both are present, it just gives up and uses "x", with ensuing incorrect behavior.

## Steps to reproduce

1. Take this simple program.
1. Select "3".
1. Execute the refactoring.
1. Type to rename the new variable to "blah".

```rust
fn main() {
    let i = 1; 
    let x = 2;

    let z = 3 + 4;
    let w = x + 5;
}
```
### Expected result
```rust
fn main() {
    let i = 1; 
    let x = 2;
    let blah = 3;

    let z = blah + 4;
    let w = x + 5;
}
```
### Actual result
```rust
fn main() {
    let i = 1; 
    let x = 2;
    let blah = 3;

    let z = blah + 4;
    let w = blah + 5;
}
```

Note how `let w = x + 5` became `let w = blah + 5`, even though it had nothing to do with any of it.

ortem:
Fixes #2919

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/introduceVariable/RsNameSuggestions.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/introduceVariable/RsNameSuggestions.kt
@@ -7,25 +7,28 @@ package org.rust.lang.refactoring.introduceVariable
 
 import com.intellij.psi.PsiElement
 import com.intellij.psi.codeStyle.NameUtil
+import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.inspections.toSnakeCase
 import org.rust.ide.utils.CallInfo
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.ancestorOrSelf
 import org.rust.lang.core.psi.ext.ancestorStrict
-import org.rust.lang.core.psi.ext.descendantsOfType
-import org.rust.lang.core.types.ty.TyInteger
 import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.ty.TyInteger
 import org.rust.lang.core.types.ty.TyTraitObject
 import org.rust.lang.core.types.ty.TyTypeParameter
 import org.rust.lang.core.types.type
 import org.rust.lang.refactoring.isValidRustVariableIdentifier
 import org.rust.openapiext.hitOnFalse
+import org.rust.stdext.mapNotNullToSet
 
 
 class SuggestedNames(
     val default: String,
     val all: LinkedHashSet<String>
 )
+
+const val FRESHEN_LIMIT = 1000
 
 /**
  * This suggests names for an expression about to be bound to a local variable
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/refactoring/introduceVariable/RsNameSuggestions.kt
```diff
@@ -91,9 +104,9 @@ private fun nameForCall(expr: RsCallExpr): List<String> {
     return listOf(pathElement.text)
 }
 
-private fun findNamesInLocalScope(expr: PsiElement): List<String> {
+private fun findNamesInLocalScope(expr: PsiElement): HashSet<String> {
     val blockScope = expr.ancestorOrSelf<RsBlock>()
     val letDecls = blockScope?.descendantsOfType<RsLetDecl>().orEmpty()
 
-    return letDecls.mapNotNull { it.pat?.text }
+    return letDecls.asSequence().mapNotNull { it.pat?.text }.toHashSet()
```
vlad20012: Hahahaha

vlad20012: let (i, x) = (1, 2);

src/main/kotlin/org/rust/lang/refactoring/introduceVariable/RsNameSuggestions.kt
```diff
@@ -91,9 +104,9 @@ private fun nameForCall(expr: RsCallExpr): List<String> {
     return listOf(pathElement.text)
 }
 
-private fun findNamesInLocalScope(expr: PsiElement): List<String> {
+private fun findNamesInLocalScope(expr: PsiElement): HashSet<String> {
```
vlad20012: `Set` can be used?

src/main/kotlin/org/rust/lang/refactoring/introduceVariable/RsNameSuggestions.kt
Undin: I still think that in such cases we should use `mapNotNullTo`

ortem:
bors retry

PR #2954, opened 2018-10-17
Make (Std)KnownItems CargoProject-related cached value

vlad20012:
Also rename StdKnownItems -> KnownItems (we may want to lookup
not only std items)


src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
```diff
@@ -0,0 +1,127 @@
...
+
+val RsElement.knownItems: KnownItems
+    get() = cargoProject?.knownItems ?: DUMMY_KNOWN_ITEMS
+
+/**
+ * [CargoProject]-related object that allows to lookup rust items in project dependencies,
+ * including the standard library.
+ */
+@Suppress("PropertyName")
+class KnownItems(
```
Undin: I'm still not sure that it's the best name but I can't come up with a better one so I think it's ok for now

PR #2975, opened 2018-10-22
Support symlinks better

### LINKED ISSUE #2466: "Cannot declare a new module at this location" on all external crates with symlinked `~/.cargo`
Presumably related to #652, same symptoms as #637. I had my `.cargo/` symlinked to another drive because I have a small SSD, and was seeing `cannot declare a new module at this location` on `mod` declarations in `lib.rs` in all external crates (examined through `External Libraries`). The only exceptions were the stdlib crates found in `rust-src` (`std`, `core`, `alloc`, `rustc_*san`, etc.), which have a different icon and seem to correctly detect `lib.rs` (in purple). 
ij-rust-crates.

This resulted in basically all IDE features breaking: syntax highlighting, autocomplete, goto-def, etc.

ij-rust-no-highlight

The problem immediately resolved itself when I replaced my `.cargo` symlink with the actual directory. It'd be really nice to get a fix for this case since I'm pretty tight on SSD space (and I imagine there are others in the same situation)!

I encountered this issue on IntelliJ 181.4445.78, plugin version 0.2.0.2092-181 on Windows 10

### LINKED ISSUE #2633: Some std references aren't resolved
I also see the #2220  error, but I'm not sure if it's related.

```
use std::ffi::CString; // resolves
use std::path::Path; // resolves
use std::ptr::NonNull; // doesn't resolve
use std::slice; // doesn't resolve
```

Looks like libstd resolves, but not libcore? 

Intellij-rust is pointing to an unzipped `wget https://static.rust-lang.org/dist/rustc-1.27.0-src.tar.gz`, ie. not installed via rustup. The issue is fixed when I grab a new copy of the sources, or by moving the existing, seemingly broken copy somewhere else. 

Something fishy is going on.. how would I go about debugging why only some of the refs weren't being resolved? I thought it might be some internal corporate issue, but would expect that to cause intellij-rust to not work with the stdlib at all.

As a side note, the Standard Library config continues to use the previous path if the new path does not contain sources.

Undin:
Resolve all symlinks in packages and targets.

Also, resolve symlinks before any call of `LightDirectoryIndex` to use the same virtual files as we use while index creation

Should fix #2466, but I don't know how to create good tests for this case
Fixes #2633

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
     - OLD_RUST_VERSION=1.25.0
     - CURRENT_RUST_VERSION=1.29.0
     - NIGHLY_RUST_VERSION=nightly-2018-10-21
+    - RUST_SRC_WITH_SYMLINK=$HOME/.rust-src
   matrix:
     # 182 platform
     - RUST_VERSION=$NIGHLY_RUST_VERSION ORG_GRADLE_PROJECT_platformVersion=182 ORG_GRADLE_PROJECT_sinceBuild=182 ORG_GRADLE_PROJECT_ideaVersion=182.4892.20 # modified by script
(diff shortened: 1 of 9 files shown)


Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Undin:
bors try

Issue #2471, opened 2018-04-14
Intention to introduce getter for a field


matklad:
for 

```rust
struct Spam {
    eggs: u64
}

impl Spam {

}
```

the intention should generate

```rust
struct Spam {
    eggs: u64
}

impl Spam {
    fn eggs(&self) -> u64 {
        self.eggs
    }
}
```

Ideally, it should be able to fix all call-site as well!

Some things to consider:

* check if a getter already exists
* which impl block to add method to?
* fixup return type (Foo -> &Foo, u64 -> u64, String -> &str)

PR #2980, opened 2018-10-24
WIP: INT: Generate accessors.

ttaomae:
My attempt at #2471. 

This is my first attempt at implementing an intention. This is still a work in progress, but hopefully I'm at least on the right track. Any feedback is welcome.

I also had some questions.

> Ideally, it should be able to fix all call-site as well!

I don't quite understand what this means. Does this mean to replace field access with calls to the getter? e.g. `spam.eggs` becomes `spam.eggs()`

> * which impl block to add method to?

Currently I just check that it is an `impl` block that is not implementing a trait. Where else might you want to put getters?

> * fixup return type (Foo -> &Foo, u64 -> u64, String -> &str)

I currently return primitive types as-is and return a reference to non-primitive types. Since `String` is non-primitive the getter returns `&String`. Is it considered idiomatic Rust to return `&str` instead?

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/GenerateAccessorIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/GenerateAccessorIntention.kt
@@ -1,0 +1,63 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.ty.TyPrimitive
+import org.rust.lang.core.types.type
+
+class GenerateAccessorIntention : RsElementBaseIntentionAction<GenerateAccessorIntention.Context>() {
+    data class Context(val structItem: RsStructItem, val implItem: RsImplItem)
+
+    override fun getText() = "Generate accessor for fields struct"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        // Must be in members of an `impl` block
+        if (element.parent !is RsMembers) return null
+        val implItem = element.ancestorStrict<RsImplItem>() ?: return null
+
+        // Must be implementing a struct.
+        val implType = (implItem.typeReference?.type) as? TyAdt ?: return null
+        val structItem = implType.item as? RsStructItem ?: return null
+
+        // Must not be implementing a trait.
+        return if (implItem.traitRef == null) Context(structItem, implItem) else null
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        val functionNames = ctx.implItem.members?.functionList
+            ?.map { it.identifier.text }
+            ?: listOf()
+
+        // Get fields in struct.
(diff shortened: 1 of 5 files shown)


ttaomae:
Added description for intention. 

I also noticed that in the tests, it sees `String` as an unknown type and therefore does not generate the method correctly. It seems to work when the plugin is actually running in IntelliJ though (I guess I missed this because I quickly modified one of the tests after testing it in the IDE and I didn't re-run the tests before pushing). Is there a way to fix this?

Undin:
@ttaomae I haven't looked at your code yet but looks like that your tests know nothing about stdlib and that's why type of `String` is `TyUnknown`. Try to add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` annotation on test functions which require stdlib or on whole test class if each test require stdlib

amakeev:
>>Ideally, it should be able to fix all call-site as well!
>I don't quite understand what this means. Does this mean to replace field access with calls to the getter? e.g. spam.eggs becomes spam.eggs()

I believe it means that it could encapsulate the direct uses of the fields with accessors (as you shown on the example); though, both 'intention' and 'generate' features in IntelliJ tend to be basic and not to to extra work. For encapsulation a separate refactoring is preferable.

ttaomae:
@Undin, thanks for the tip. That seems to have fixed it.

@amakeev, if I understand correctly, you are basically saying that this intention should not make any changes to call-sites, and that that should be a separate feature. Is that correct?

I've also made a change to skip `pub` fields since my understanding is that public fields could just be accessed directly. This also raises another question. Should the generated methods be `pub`? My understanding is that getters are intended to give public access to non-`pub` fields without allowing the value to be modified directly. If the field and getter are both non-`pub` is there any benefit to accessing the field through the method?

amakeev:
@ttaomae 
>you are basically saying that this intention should not make any changes to call-sites, and that that should be a separate feature. Is that correct?

yes

> If the field and getter are both non-pub is there any benefit to accessing the field through the method?

Not sure if a use case is really useful (have a public field, make an accessor, then probably make the field private); but I'd go for consistency anyways (that is, make the intention available regardless the visibility) as it lowers the mental burden.

Kobzol:
@ttaomae is this PR still in active development?

src/main/kotlin/org/rust/ide/intentions/GenerateAccessorIntention.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        // Generate accessors for each field if there is no method name conflict.
+        fields.filter { it.identifier.text !in functionNames }.forEach { field ->
+            // Skip public fields since those can be accessed directly.
+            if (field.isPublic) return@forEach
+
+            val fieldName = field.identifier.text
+            val fieldType = field.typeReference?.type ?: return@forEach
+
+            val borrow = if (fieldType is TyPrimitive) "" else "&"
```
Kobzol: Maybe it would be better to return the attribute by value if it implements the `Copy` trait?

ttaomae:
@Kobzol, to be honest, I've completely forgotten about it. I didn't have time to work on it then it eventually slipped my mind. Sorry to all involved for leaving this open for so long.

I'm not sure what the protocol is here, but if someone would like to take over or to just decline and close this pull request, that's fine with me, since I probably won't be able to work on this for a while.

Kobzol:
I'd like to take a shot at finishing this. @Undin @amakeev what's left to do?

Kobzol:
I updated this to the current master and added support for Copy types (https://github.com/Kobzol/intellij-rust/tree/issue-2471). Maybe it would be nice to show a dialog listing the available fields, so that user can select which getters/setters to implement? Same as how it works in other IntelliJ IDEs.

Kobzol:
@mchernyavsky I'd like to try to push this intention forward. Should I create a new PR?

PR #2985, opened 2018-10-26
MACRO: resolve macros in lexical order

vlad20012:


vlad20012:
> Some cases when current implementation returns wrong resolve elements

Thanks! I forget to test such simple cases =D These was a little problem that we did't stopped iteration when a processor returned true.

PR #2988, opened 2018-10-26
MACRO: expand log macros

vlad20012:
Currently log macros are hardcoded in our grammar and we
can't really expand them. The problem is that std contains
macro with name `debug` that is not "log" macro, and we should
be able to expand it


Undin:
@vlad20012 maybe add some test(s)?

vlad20012:
@Undin It's something very weird that macro with name `debug` can't be expanded just because of its name. And specially test this case... This anyway will be covered when we will start to expand std macros

Undin:
@vlad20012 it's not weird, it's just part of our implementation. I see it as a way to commit desired behavior. It will allow changing the implementation in future without fear to break something

vlad20012:
Okaay, why not.
r? @Undin

Issue #2259, opened 2018-02-04
Import Optimizer


Undin:
Meta issue to support import optimizer:

* [x] Sort use speck in use item
* [x] Sort different import items
* [x] Sort `extren crate` items
* [ ] Collapse imports when possible
* [ ] Remove unused imports
* [x] Group imports by semantics (std, crates, inner mods, etc.)

Related issues:
#2158
~~#2248~~ 
~~#2276~~
~~#2519~~
~~#2577~~
~~#2929~~

farodin91:
Possible other tasks?
* grouping into semantic imports (std, crates, inner mods) (no trivial)
* sort _mod_ and _extern crate_

Undin:
I'm not sure about mod sorting because it's not about items import
Other suggestions sound good)

farodin91:
For extern crate: https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#reorder_extern_crates (possible also in groups)
For grouping of imports: https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#reorder_imports_in_group
I thought of 
```rust
mod a;
mod b;
mod c;
```

Undin:
Yeah, I've understood your idea
I think mod reordering is not very good idea for `Optimize Import` action because it's quite unexpected thing for user while import optimizing
It can be part of other action (for example, formatting) or separate action or something else

farodin91:
Okay

farodin91:
#2277

Undin:
@farodin91 looks like that it's formatter bug - import optimizer works correctly in this case

farodin91:
I wanted to link #2276.

przygienda:
minor but annoying: 

in extern create X as Y 

auto import 

does use X::v 

instead of 

use Y::v

Undin:
@przygienda It's not related to import optimizer
I've created separate issue #2444

U007D:
IntelliJ Rust gets better with each release, thank you everyone!

I've noticed that `Optimize Imports` places `use` statements first followed by `mod` statements.  I would like the inverse order--`mods` then `use` (as my `use` path statements often depend on the `mod`.  As a reviewer, this is easier to follow when it's not in 'forward-declaration format').  Is that configurable somewhere or is it hard-coded?

BillOTei:
installed it on Clion and Idea ultimate, it does not seem to remove unused imports, is it normal?

[...]

PR #2990, opened 2018-10-27
ACT: add ability to group imports by semantics for import optimizer

isamborskiy:
Fulfilled #2259 improvement (the last checkbox).

Add the ability to group imports by semantics (std, crates, inner mods, etc.)
Groups order: [here](https://github.com/isamborskiy/intellij-rust/blob/ce89a6cb41477c480edf0dfdb2d9285fc394084e/src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt#L113)
Examples: [here](https://github.com/isamborskiy/intellij-rust/blob/ce89a6cb41477c480edf0dfdb2d9285fc394084e/src/test/kotlin/org/rust/lang/refactoring/RsImportOptimizerTest.kt#L293) and [here](https://github.com/isamborskiy/intellij-rust/blob/ce89a6cb41477c480edf0dfdb2d9285fc394084e/src/test/kotlin/org/rust/lang/refactoring/RsImportOptimizerTest.kt#L317)

### DIFF
--- a/src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
+++ b/src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
@@ -6,13 +6,13 @@
 package org.rust.lang.refactoring
 
 import com.intellij.lang.ImportOptimizer
+import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiWhiteSpace
+import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.formatter.processors.asTrivial
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsAttr
-import org.rust.lang.core.psi.ext.RsElement
-import org.rust.lang.core.psi.ext.RsMod
-import org.rust.lang.core.psi.ext.childrenOfType
+import org.rust.lang.core.psi.ext.*
 
 class RsImportOptimizer : ImportOptimizer {
     override fun supports(file: PsiFile?): Boolean = file is RsFile
@@ -72,18 +72,63 @@ class RsImportOptimizer : ImportOptimizer {
             val first = mod.childrenOfType<RsElement>()
                 .firstOrNull { it.textOffset >= offset && it !is RsExternCrateItem && it !is RsAttr } ?: return
             val psiFactory = RsPsiFactory(mod.project)
-            val sortedUses = uses
-                .sortedBy { it.useSpeck?.pathText }
-                .mapNotNull { it.copy() as? RsUseItem }
-            sortedUses
-                .mapNotNull { it.useSpeck }
-                .forEach { optimizeUseSpeck(psiFactory, it) }
-            for (importPath in sortedUses) {
-                mod.addBefore(importPath, first)
+            val sortedUsesGroups = uses
+                .map { UseItemWrapper(it) }
+                .groupBy({ it.packageGroupLevel }, { it })
+                .map { (groupLevel, useItemWrapper) ->
+                    val useItems = useItemWrapper
+                        .sortedBy { it.useSpeckText }
+                        .mapNotNull { it.useItem.copy() as? RsUseItem }
+                    groupLevel to useItems
+                }
+                .sortedBy { it.first }
+            sortedUsesGroups.forEach { (_, group) ->
+                group
+                    .mapNotNull { it.useSpeck }
+                    .forEach { optimizeUseSpeck(psiFactory, it) }
             }
-            uses.forEach { it.delete() }
+
+            for ((_, sortedUses) in sortedUsesGroups) {
+                var lastAddedUseItem: PsiElement? = null
+                for (importPath in sortedUses) {
+                    lastAddedUseItem = mod.addBefore(importPath, first)
+                    mod.addAfter(psiFactory.createNewline(), lastAddedUseItem)
+                }
+                mod.addAfter(psiFactory.createNewline(), lastAddedUseItem)
+            }
+            uses.forEach {
+                (it.nextSibling as? PsiWhiteSpace)?.delete()
+                it.delete()
+            }
         }
     }
 }
 
 private val RsUseSpeck.pathText get() = path?.text?.toLowerCase()
+
+private class UseItemWrapper(val useItem: RsUseItem) {
+    private val basePath: RsPath? = useItem.useSpeck?.path?.basePath()
+
+    val useSpeckText: String? = useItem.useSpeck?.pathText
+
+    // `use` order:
+    // 1. Standard library (stdlib)
+    // 2. Related third party (extern crate)
+    // 3. Local
+    //    - otherwise
+    //    - crate::
+    //    - super::
+    //    - self::
+    val packageGroupLevel: Int = when {
+        basePath?.self != null -> 6
+        basePath?.`super` != null -> 5
+        basePath?.crate != null -> 4
+        else -> when (basePath?.reference?.resolve()?.containingCargoPackage?.origin) {
+            PackageOrigin.WORKSPACE -> 3
+            PackageOrigin.TRANSITIVE_DEPENDENCY -> 2
+            PackageOrigin.DEPENDENCY -> 2
+            PackageOrigin.STDLIB -> 1
+            else -> 3
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
+                        .sortedBy { it.useSpeckText }
+                        .mapNotNull { it.useItem.copy() as? RsUseItem }
+                }
+                .filter { it.second.isNotEmpty() }
+                .sortedBy { it.first }
+            sortedUsesGroups.forEach { sortedUses ->
+                sortedUses.second
+                    .mapNotNull { it.useSpeck }
+                    .forEach { optimizeUseSpeck(psiFactory, it) }
+            }
```
Undin: Maybe use desctructuring here too?
```kotlin
sortedUsesGroups.forEach { (_, group) ->
    group
        .mapNotNull { it.useSpeck }
        .forEach { optimizeUseSpeck(psiFactory, it) }
}
```

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
+
+            for ((_, sortedUses) in sortedUsesGroups) {
+                val lastUseItem = sortedUses.map { importPath ->
```
Undin: You don't need `map` here, it can be replaced by `for` loop and mutable variable.
It allows avoiding unnecessary allocation

src/test/kotlin/org/rust/lang/refactoring/RsImportOptimizerTest.kt
```diff
@@ -232,7 +289,78 @@ class RsImportOptimizerTest: RsTestBase() {
...
+        use aaa::bbb;
+        use bbb::{eee, fff};
+        use sss::aaa;
+
+        use crate::aaa::bbb;
+        use crate::bbb::ccc;
+
+        use super::ccc::bbb;
+
+        use self::aaa;
```
Undin: Not sure that we should add empty line between `self`, `crate`, `super` and other local imports
@matklad what do you think?

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
+                    val addedUseItem = mod.addBefore(importPath, first)
+                    mod.addAfter(psiFactory.createNewline(), addedUseItem)
+                    addedUseItem
+                }.last()
+                mod.addAfter(psiFactory.createNewline(), lastUseItem)
+            }
+            uses.forEach {
+                (it.nextSibling as? PsiWhiteSpace)?.delete()
```
Undin: Do we really need it? If yes, can we add the corresponding case in test?

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
+            sortedUsesGroups.forEach { sortedUses ->
+                sortedUses.second
+                    .mapNotNull { it.useSpeck }
+                    .forEach { optimizeUseSpeck(psiFactory, it) }
+            }
+
+            for ((_, sortedUses) in sortedUsesGroups) {
+                val lastUseItem = sortedUses.map { importPath ->
+                    val addedUseItem = mod.addBefore(importPath, first)
+                    mod.addAfter(psiFactory.createNewline(), addedUseItem)
```
Undin: Interesting... it doesn't affect next item in the same group but without it, group separator group doesn't appear in the editor
But looks like we can drop it and insert two newlines after each group to have the same effect

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
-                mod.addBefore(importPath, first)
+            val sortedUsesGroups = uses
+                .map { UseItemWrapper(it) }
+                .groupBy({ it.packageGroupLevel }, { it })
+                .map { (groupLevel, useItemWrapper) ->
+                    groupLevel to useItemWrapper
+                        .sortedBy { it.useSpeckText }
+                        .mapNotNull { it.useItem.copy() as? RsUseItem }
+                }
+                .filter { it.second.isNotEmpty() }
```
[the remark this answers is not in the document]
isamborskiy: I think in practice, this filter operator is redundant; however, in theory, we can obtain empty list after `mapNotNull` above.

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
-                .sortedBy { it.useSpeck?.pathText }
-                .mapNotNull { it.copy() as? RsUseItem }
-            sortedUses
-                .mapNotNull { it.useSpeck }
-                .forEach { optimizeUseSpeck(psiFactory, it) }
-            for (importPath in sortedUses) {
-                mod.addBefore(importPath, first)
+            val sortedUsesGroups = uses
+                .map { UseItemWrapper(it) }
+                .groupBy({ it.packageGroupLevel }, { it })
```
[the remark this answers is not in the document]
isamborskiy: I consider that my way to write `groupBy` is more readable, but I am okay to change it

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
isamborskiy: Yes, this is true.
However, I find the current version more logical. 
You do not think so, do you?

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
isamborskiy: Okay, done

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
isamborskiy: Done

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
isamborskiy: Yes, we need it
I have added corresponding [test](https://github.com/isamborskiy/intellij-rust/blob/af3a3a6607d5ed36994fba8bc82174a51c200677/src/test/kotlin/org/rust/lang/refactoring/RsImportOptimizerTest.kt#L364)

src/test/kotlin/org/rust/lang/refactoring/RsImportOptimizerTest.kt
matklad: I don't have a strong opinion and mildly in favor: `self` and `super` imports are pretty rare in practice, and when they are used, it is ok to slightly emphasize them.

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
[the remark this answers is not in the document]
Undin: But if `it.second.isEmpty()` then it means that `it.useItem.copy()` (some lines above) returned `null`. Looks like it never happens

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
Undin: I don't insist, it's a matter of taste

src/main/kotlin/org/rust/lang/refactoring/RsImportOptimizer.kt
```diff
@@ -72,18 +71,63 @@ class RsImportOptimizer : ImportOptimizer {
...
+                .map { (groupLevel, useItemWrapper) ->
+                    groupLevel to useItemWrapper
+                        .sortedBy { it.useSpeckText }
+                        .mapNotNull { it.useItem.copy() as? RsUseItem }
```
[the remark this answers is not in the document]
isamborskiy: Fixed

singalen:
`std::collections::HashMap` was auto-imported, and the import was below `crate::x`.
Is it a regression?

PR #2992, opened 2018-10-27
Support 'crate in path' in 2015 edition

Undin:
* Add `updateCompilerFeatures` task to update compiler features
* Update compiler features not to annotate `crate` keyword in paths since rust 1.30


Undin:
bors r=vlad20012

PR #2995, opened 2018-10-28
RES: Support paths start with crate name for 2015 edition

### LINKED ISSUE #2987: Support paths start with crate name for 2015 edition
Since Rust 1.30 it's possible to  start non-global paths from crate name. e.g.
```rust
extern crate rand;

fn main() {
    foo::bar();
}

mod foo {
//    mod rand {
//        pub fn random<T>() {}
//    }

    pub fn bar() {
        rand::random::<i32>();
    }
}
```
In this case, path `rand::random` will be resolved to `rand` crate. But if uncomment `mod rand`, the path will be resolved to this module.

Also it's legal to omit `extern crate` declaration.

vlad20012:
Fixes #2987

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
@@ -32,10 +32,19 @@ interface ScopeEntry {
  * to the resolve processor
  */
 enum class ScopeEvent : ScopeEntry {
-    // Communicate to the resolve processor that we are about
-    // to process wildecard imports. This is basically a hack
-    // to make winapi 0.2 work in a reasonable amount of time.
-    STAR_IMPORTS;
+    /**
+     * Communicate to the resolve processor that we are about to process wildecard imports.
+     * This is basically a hack to make winapi 0.2 work in a reasonable amount of time.
+     */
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -32,10 +32,20 @@ interface ScopeEntry {
...
-    STAR_IMPORTS;
+    /**
+     * Communicate to the resolve processor that we are about to process wildecard imports.
+     * This is basically a hack to make winapi 0.2 work in a reasonable amount of time.
+     */
+    STAR_IMPORTS,
+    /**
+     * Communicate to the resolve processor that we are about to process dependency crates
+     * (regardless of `extern crate` declarations). It's used to support 2018 edition
+     * (and Rust 1.30) paths, where non-global path can start with a crate name, but
```
Undin: `since Rust 1.30`

Undin: I mean that we should slightly rewrite this part of the comment to clarify that such resolve should work in edition 2018 and in 2015 edition since rust 1.30

PR #2956, opened 2018-10-17
RUN: Bench runner

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/2488.

Adds a gutter icon to launch benchmarks in the same way as tests.

Edit Configuration Menu

Can be merged after https://github.com/intellij-rust/intellij-rust/pull/2963.

PR #2999, opened 2018-10-29
RUN: Run all tests or benches in directory

mchernyavsky:
Allows you to run:
* all tests from a Cargo project root
* all tests from a Cargo package root
* all tests in directory under the "tests" source root
* all benches in directory under the "benches" source root

Examples:

Edit Configuration Menu Edit Configuration Menu

Can be merged after https://github.com/intellij-rust/intellij-rust/pull/2998 and https://github.com/intellij-rust/intellij-rust/pull/2956.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoBenchRunConfigurationProducer.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoBenchRunConfigurationProducer.kt
@@ -5,8 +5,8 @@
 
 package org.rust.cargo.runconfig.test
 
+import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.RsFunction
-import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.RsMod
 import org.rust.lang.core.psi.ext.isBench
 import org.rust.lang.core.psi.ext.processExpandedItemsExceptImpls
@@ -14,11 +14,11 @@ import org.rust.lang.core.psi.ext.processExpandedItemsExceptImpls
 class CargoBenchRunConfigurationProducer : CargoTestRunConfigurationProducerBase() {
     override val commandName: String = "bench"
 
-    override fun isSuitable(element: RsElement): Boolean =
+    override fun isSuitable(element: PsiElement): Boolean =
         when (element) {
             is RsMod -> hasBenchFunction(element)
             is RsFunction -> element.isBench
-            else -> error("expected RsMod or RsFunction")
+            else -> false
         }
 
     companion object {
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
@@ -8,26 +8,42 @@ package org.rust.cargo.runconfig.test
 import com.intellij.execution.actions.ConfigurationContext
 import com.intellij.openapi.actionSystem.LangDataKeys
 import com.intellij.openapi.util.Ref
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.psi.PsiDirectory
 import com.intellij.psi.PsiElement
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.cargo.project.model.impl.allTargets
 import org.rust.cargo.project.workspace.CargoWorkspace
+import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
 import org.rust.cargo.runconfig.command.CargoRunConfigurationProducer
 import org.rust.cargo.runconfig.mergeWithDefault
 import org.rust.cargo.toolchain.CargoCommandLine
+import org.rust.lang.core.psi.RsFile
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.ext.*
+import org.rust.openapiext.pathAsPath
+import org.rust.openapiext.toPsiFile
+
+private typealias TestConfigProvider = (List<PsiElement>, Boolean) -> TestConfig?
 
 abstract class CargoTestRunConfigurationProducerBase : CargoRunConfigurationProducer() {
-    abstract val commandName: String
+    protected abstract val commandName: String
+    private val testConfigProviders: MutableList<TestConfigProvider> = mutableListOf()
 
-    abstract fun isSuitable(element: RsElement): Boolean
+    init {
+        registerConfigProvider { elements, climbUp -> createConfigFor<RsFunction>(elements, climbUp) }
+        registerConfigProvider { elements, climbUp -> createConfigFor<RsMod>(elements, climbUp) }
+        registerConfigProvider { elements, climbUp -> createConfigForMultipleFiles(elements, climbUp) }
+        registerDirectoryConfigProvider { dir -> createConfigForDirectory(dir) }
+    }
 
     override fun isConfigurationFromContext(
         configuration: CargoCommandConfiguration,
         context: ConfigurationContext
     ): Boolean {
-        val existingConfig = findConfig(context) ?: return false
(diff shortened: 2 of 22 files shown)


src/test/kotlin/org/rust/cargo/runconfig/producers/TestRunConfigurationProducerTest.kt
```diff
@@ -11,11 +11,13 @@ import org.rust.cargo.runconfig.test.CargoTestRunConfigurationProducer
...
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.ext.RsMod
+import org.rust.lang.core.psi.ext.cargoPackage
+import org.rust.openapiext.toPsiDirectory
 
 class TestRunConfigurationProducerTest : RunConfigurationProducerTestBase() {
     fun `test test producer works for annotated functions`() {
         testProject {
-            lib("foo", "src/lib.rs", """
+            lib("foo", "tests/lib.rs", """
```
Undin: Do we really need this change (and similar in other tests)?

vlad20012:
:thinking: 
bors r-

src/test/kotlin/org/rust/cargo/runconfig/producers/TestRunConfigurationProducerTest.kt
mchernyavsky: Not really. Fixed.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
```diff
@@ -8,79 +8,119 @@ package org.rust.cargo.runconfig.test
...
+        val configPath = context.crateRelativePath.configPath() ?: return null
+        val target = context.containingCargoTarget ?: return null
+        return SingleItemTestConfig(commandName, configPath, target, context)
     }
 
-    private fun createMultipleFileConfiguration(modules: Array<RsMod>): TestConfig? {
-        val modulesWithBenches = modules.filter { isSuitable(it) && it.containingCargoTarget != null }
+    private fun createConfigForMultipleFiles(psi: Array<PsiElement>, climbUp: Boolean): TestConfig? {
+        val context = psi.mapNotNull { findElement<RsMod>(it, climbUp) }
+        val modsWithTests = context.filter { isTestModOrFunction(it) && it.containingCargoTarget != null }
```
Undin: Hm, looks like here we want to create test config for multiple files if we can create the corresponding config at least for one atom (function or module) element inside of them.
Can we encapsulate the logic if atom element is suitable for test config creation to some objects and reuse this object for complex elements like a list of psi element or a directory?

mchernyavsky: Not sure I understood, but I have moved common part from `createConfigFor, `createConfigForMultipleFiles` and `createConfigForDirectory` into a separate method + class.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoBenchRunConfigurationProducer.kt
```diff
@@ -5,20 +5,20 @@
...
-import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.RsMod
 import org.rust.lang.core.psi.ext.isBench
 import org.rust.lang.core.psi.ext.processExpandedItemsExceptImpls
 
 class CargoBenchRunConfigurationProducer : CargoTestRunConfigurationProducerBase() {
     override val commandName: String = "bench"
 
-    override fun isSuitable(element: RsElement): Boolean =
+    override fun isTestModOrFunction(element: PsiElement): Boolean =
```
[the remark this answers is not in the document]
mchernyavsky: Actually, cargo `test` command also runs benchmarks, but ok.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
mchernyavsky: Seems to have done.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
```diff
@@ -8,98 +8,144 @@ package org.rust.cargo.runconfig.test
...
-    abstract fun isSuitable(element: RsElement): Boolean
+abstract class CargoTestRunConfigurationProducerBase : CargoRunConfigurationProducer() {
+    protected abstract val commandName: String
+    private val testConfigProviders: MutableList<TestConfigProvider> = mutableListOf()
+
+    init {
+        registerConfigProvider { elements, climbUp -> createConfigFor<RsFunction>(elements, climbUp) }
+        registerConfigProvider { elements, climbUp -> createConfigFor<RsMod>(elements, climbUp) }
+        registerConfigProvider { elements, climbUp -> createConfigForMultipleFiles(elements, climbUp) }
+        registerConfigProvider { elements, _ -> createConfigForDirectory(elements) }
```
Undin: Maybe it's better to add more specific api for single directory?
Now we have three similar methods with the following code:
```kotlin
private fun createConfigFor*(elements: List<PsiElement>): TestConfig? {
    val dir = elements.singleOrNull() as? PsiDirectory ?: return null
    // some other code
}
```
We can create `registerDirectoryConfigProvider` method where we check that we have only one psi element is in context and it's a directory. It allows us to pass only single `PsiDirectory`  element into `createConfigFor*` methods, drop the same code in these methods and in general provide clearer API

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
```diff
@@ -8,98 +8,144 @@ package org.rust.cargo.runconfig.test
...
+        val elements = LangDataKeys.PSI_ELEMENT_ARRAY.getData(context.dataContext)?.toList()
+            ?: context.location?.psiElement?.let { listOf(it) }
+        return elements?.let { findTestConfig(it) }
     }
 
-    private fun createMultipleFileConfiguration(modules: Array<RsMod>): TestConfig? {
-        val modulesWithBenches = modules.filter { isSuitable(it) && it.containingCargoTarget != null }
+    private fun createConfigForDirectory(elements: List<PsiElement>): TestConfig? {
+        val dir = elements.singleOrNull() as? PsiDirectory ?: return null
+        val filesConfig = createConfigForMultipleFiles(dir.rustFiles, climbUp = false) ?: return null
```
Undin: `rustFiles` is a very expensive operation because it reads the whole project
Can we avoid using it? It seems now you need these files only to find target(s). But looks like you can find the necessary target using only `dir` itself

mchernyavsky:
@Undin  Fixed.

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestRunConfigurationProducerBase.kt
```diff
@@ -8,98 +8,159 @@ package org.rust.cargo.runconfig.test
...
     companion object {
+        private val PsiDirectory.targets: List<RsFile>
+            get() {
+                val rootPath = virtualFile.pathAsPath
+                return project
+                    .cargoProjects
+                    .allProjects
+                    .asSequence()
+                    .mapNotNull { it.workspace }
+                    .flatMap { it.packages.asSequence() }
```
Undin: We should process only workspace packages
Otherwise, we will provide run configuration for lib dependencies too
image
And they will fail as expected
```
cargo test --color=always --no-run --package serde
error: package `serde` cannot be tested because it requires dev-dependencies and is not a member of the workspace
```

mchernyavsky: Fixed (here and in `createConfigForCargoPackage`).

PR #3001, opened 2018-10-30
Make functions with external linkage unsafe

### LINKED ISSUE #2918: Extern "C" functions should be implicitly unsafe but not
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2103-182
* **Rust toolchain version:** 1.29.0 (nightly-2018-07-18-x86_64-unknown-linux-gnu)
* **IDE name and version:** CLion 182.4129.15
* **Operating system:** Ubuntu Linux 16.04

## Problem description

All functions under `extern "C" {}` block should be implicitly unsafe, but IntelliJ-Rust doesn't know that.

## Steps to reproduce

+ Write `extern "C" { pub fn bla(); }`
+ Write `fn oh() { bla(); }`
+ No error message, compilation error
+ Write `fn oh() { unsafe { bla(); } }`
+ Compile good

garyttierney:
Fixes #2918.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -120,6 +120,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
 
     private fun checkMethodCallExpr(holder: AnnotationHolder, o: RsMethodCall) {
         val fn = o.reference.resolve() as? RsFunction ?: return
+
         if (fn.isUnsafe) {
             checkUnsafeCall(holder, o.parentDotExpr)
         }
@@ -128,6 +129,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
     private fun checkCallExpr(holder: AnnotationHolder, o: RsCallExpr) {
         val path = (o.expr as? RsPathExpr)?.path ?: return
         val fn = path.reference.resolve() as? RsFunction ?: return
+
         if (fn.isUnsafe) {
             checkUnsafeCall(holder, o)
         }
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -90,6 +90,21 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
         checkReferenceIsPublic(field, o, holder)
     }
 
+    private fun checkStaticUsageIsSafe(ref: RsReferenceElement, holder: AnnotationHolder) {
+        val element = ref.reference.resolve() as? RsConstant ?: return
+        val constantType = when {
+            element.kind == RsConstantKind.MUT_STATIC -> "mutable"
+            element.kind == RsConstantKind.STATIC && element.parent is RsForeignModItem -> "extern"
+            else -> return
+        }
+
+        val expr = ref.ancestorOrSelf<RsExpr>() ?: return
+        if (expr.isInUnsafeBlockOrFn()) return
+
+        RsDiagnostic.UnsafeError(expr, "Use of $constantType static is unsafe and requires unsafe function or block")
+            .addToHolder(holder)
+    }
+
     private fun checkReferenceIsPublic(ref: RsReferenceElement, o: PsiElement, holder: AnnotationHolder) {
         val element = ref.reference.resolve() as? RsVisible ?: return
         val oMod = o.contextStrict<RsMod>() ?: return
@@ -215,6 +230,7 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
             }
         }
 
+        checkStaticUsageIsSafe(path, holder)
         checkReferenceIsPublic(path, path, holder)
     }
 
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -33,7 +33,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
         // Bump this number if Stub structure changes
-        override fun getStubVersion(): Int = 144
+        override fun getStubVersion(): Int = 145
 
         override fun getBuilder(): StubBuilder = object : DefaultStubBuilder() {
             override fun createStubForFile(file: PsiFile): StubElement<*> = RsFileStub(file as RsFile)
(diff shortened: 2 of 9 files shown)


garyttierney:
On second look, I think it might be better to code this into the `isUnsafe` accessor so those error annotations can be computed completely from stub information.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -12,6 +12,7 @@ import com.intellij.lang.annotation.Annotator
 import com.intellij.openapi.util.Key
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.util.parentOfType
```
Undin: Unused import

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -120,15 +121,19 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
 
     private fun checkMethodCallExpr(holder: AnnotationHolder, o: RsMethodCall) {
         val fn = o.reference.resolve() as? RsFunction ?: return
-        if (fn.isUnsafe) {
+        val isExternalDecl = fn.parent as? RsForeignModItem != null
```
Undin: `fn.parent as? RsForeignModItem != null` -> `fn.parent is RsForeignModItem`

src/test/kotlin/org/rust/ide/annotator/RsExpressionAnnotatorTest.kt
```diff
@@ -149,6 +149,14 @@ class RsExpressionAnnotatorTest : RsAnnotationTestBase() {
         }
     """)
 
+    fun `test external ABI is unsafe`() = checkErrors("""
+        extern "C" { fn foo(); }
+
+        fn main() {
+            <error>foo()</error>;
```
Undin: Let's check the description of annotation here too. Just add `descr="some text"` attr into `error` tag

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -129,7 +129,7 @@ class RsErrorAnnotatorTest : RsAnnotationTestBase() {
             fn variadic_2(p1: u32, p2: u32, ...);
         }
 
-        fn main() {
+        unsafe fn main() {
```
Undin: It seems you want to avoid unnecessary error annotation that you added in this PR.
But `main` can't be unsafe, so let's rename it somehow

src/test/kotlin/org/rust/ide/annotator/RsExpressionAnnotatorTest.kt
Undin: Also, it would be great to ensure that `Surround with unsafe block` quick fix works in such case too.
You can find similar tests in `org.rust.ide.annotator.fixes.AddUnsafeTest`

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -120,15 +121,19 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
         }
     }
 
     private fun checkCallExpr(holder: AnnotationHolder, o: RsCallExpr) {
         val path = (o.expr as? RsPathExpr)?.path ?: return
         val fn = path.reference.resolve() as? RsFunction ?: return
-        if (fn.isUnsafe) {
+        val isExternalDecl = fn.parent as? RsForeignModItem != null
+
+        if (fn.isUnsafe || isExternalDecl) {
```
Undin: Looks like a code duplication
But if you move this logic into `isUnsafe` property, it will be fixed automatically

garyttierney:
@Undin could you provide a code example of a reference to a static declared within an `unsafe` block? I'm not clear on how that would look.  In the mean time I've also added unsafety checks to `static mut` references.

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -555,7 +555,7 @@ class RsFunctionStub(
             flags = BitUtil.set(flags, CONST_MASK, psi.isConst)
             flags = BitUtil.set(flags, UNSAFE_MASK, psi.isUnsafe)
             flags = BitUtil.set(flags, EXTERN_MASK, psi.isExtern)
-            flags = BitUtil.set(flags, VARIADIC_MASK, psi.isExtern)
+            flags = BitUtil.set(flags, VARIADIC_MASK, psi.isVariadic)
```
Undin: Good catch! 👍 

But we should increase stub version (see `org.rust.lang.core.stubs.RsFileStub.Type#getStubVersion`) because of this and changes in `isUnsafe` function to recalculate stubs and avoid potential incorrect annotations using old stubs

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -90,6 +90,24 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
     }
 
+    private fun checkConstantUsageIsSafe(ref: RsReferenceElement, holder: AnnotationHolder) {
+        val element = ref.reference.resolve() as? RsConstant ?: return
+        val expr = ref.ancestorOrSelf<RsExpr>() ?: return
+
+        val isInUnsafeCx = expr.isInUnsafeBlockOrFn()
+        if (isInUnsafeCx) {
+            return
+        }
```
Undin: Looks like `if (expr.isInUnsafeBlockOrFn()) return` is not less readable but shorter

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -90,6 +90,24 @@ class RsErrorAnnotator : Annotator, HighlightRangeExtension {
...
+
+        if (element.kind == RsConstantKind.MUT_STATIC) {
+            RsDiagnostic.UnsafeError(expr, "Use of mutable static is unsafe and requires unsafe function or block")
+                .addToHolder(holder)
+        } else if (element.kind == RsConstantKind.STATIC && element.parent is RsForeignModItem) {
+            RsDiagnostic.UnsafeError(expr, "Use of extern static is unsafe and requires unsafe function or block")
+                .addToHolder(holder)
+        }
```
Undin: Both blocks look very similar - only one difference in error message
What do you think about moving out all same code?
Something like this:
```kotlin
val constantType = when {
    element.kind == RsConstantKind.MUT_STATIC -> "mutable"
    element.kind == RsConstantKind.STATIC && element.parent is RsForeignModItem -> "extern"
    else -> return
}
RsDiagnostic.UnsafeError(expr, "Use of $constantType static is unsafe and requires unsafe function or block")
    .addToHolder(holder)
```

garyttierney: That's a good idea.  I've addressed this plus your other comment (and moved the `constantType` check above the unsafe block, to avoid the PSI traversal if the constant type isn't relevant) and rebased the tree into a few separate logical changes.

Undin:
Great!
bors r+

Issue #2979, opened 2018-10-23
--nocapture default removed from Cargo test


jmmk:
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2109-182

## Problem description

Prior to the latest version, output from `println!` calls in tests would be shown in the console (See https://github.com/intellij-rust/intellij-rust/issues/1086)

The latest version merged #2901 which removes the `--nocapture` argument and no longer displays test output. Maybe it's intended, but it caught me off guard

I also see there's an open issue https://github.com/rust-lang/rust/issues/54669 to perhaps address this

## Steps to reproduce

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn it_works() {
        println!("invisible output");
    }
}
```

mchernyavsky:
Hello.

Oh, I didn't think that changing the default options in this particular case can cause trouble.

At the moment we can't support the `--nocapture` option (https://github.com/rust-lang/rust/issues/54669 describes the problem). For the present, we're waiting for an answer from libtest developers.

chinedufn:
@mchernyavsky are there any settings that we can tweak in the IDE to disable `--format=json` and instead enable `--nocapture` for the time being?

mchernyavsky:
@chinedufn https://github.com/intellij-rust/intellij-rust/pull/3004 will partially solve the problem.

PR #2901, opened 2018-10-05
RUN: Support test window

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/1485.

This PR is mostly based on https://github.com/intellij-rust/intellij-rust/pull/2337 (a huge thanks to @farodin91 and @mkaput).

screen shot 2018-10-09 at 22 13 58

I had to remove the `--nocapture` option because of https://github.com/rust-lang/rust/issues/54669.

TODO:
- [x] Provide execution time of each test
- [x] Provide location hint and proper line marker for test suites
- [ ] Add more tests

What else can be implemented/improved?
- [ ] Add more filters
- [ ] Support "Run only tests that failed/crashed after last run"

PR #3004, opened 2018-10-30
RUN: Support `--nocapture` option

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/2979.

Restores a checkbox in the `Edit Configurations` that has been removed in [this PR](https://github.com/intellij-rust/intellij-rust/pull/2901).

If this checkbox is selected, the `--nocapture` option will be used in the tests. The checkbox not selected by default.

At the moment, the test window can't work with `--nocapture` (see https://github.com/rust-lang/rust/issues/54669), so when the checkbox is activated, the test results will be displayed in the old-style console.

Edit Configuration Menu


vlad20012:
I think it's reasonable to add some warning for this option (in parentheses) about the fact that it leads to old-style-console view.

And really we have to solve this problem in the rust's test framework (because who if not us?)

PR #2354, opened 2018-03-07
WIP: TYPE: Region folding

anton-okolelov:
fixes #2316

PR #3005, opened 2018-10-30
Add support for custom folding regions

### LINKED ISSUE #2316: Region folding not supported
regionfoldinggif

#### Region Folding
Region folding is a wonderful code-organization feature standard in Intellij/Webstorm/Intellij-Scala etc. where blocks of code enclosed by region tags may be folded.
```
 //region * some name * 
* some code *
//endregion
```
Creating regions can be done manually or by making use of the standard 'Surround With' -> Region functionality.

#### Problem
Intellij/Jetbrains region folding is not currently supported. There's an option for it in Settings under Code Folding but it has no effect: 
img

Currently we can fold functions, mods, extern crates, use etc. which is a good start. The inability to do properly nested chunks of foldable code (with custom labels!) limits code navigation speed and organization for those proficient at it. 

A benefit of proper region folding over function/body folding is that it can reduce cognitive effort required when navigating code with long/tall(many lines even when 'folded') type signatures.

I spent a night trying to implement it (going so far as adding a parsing token for it and getting '//region' recognized as something similar to a block comment)...but my efforts fell short due to lack of Kotlin/codegeneration knowledge. Memory's a little fuzzy but I remember there being something in another custom language plugin that just used a standard com.intellij.something to add the support without much todo (Maybe don't need to monkey around with the parser to get it working?).

garyttierney:
Re-uses the CustomFoldingBuilder implementation to provide support for folding
custom regions defined by `//region` comments.

Fixes #2316.


amakeev:
@garyttierney there is an existing pending PR to support custom region folding (#2354). Could you please check that all cases are covered/addressed here?

garyttierney:
@amakeev ah, I completely missed that one.  I will take a look.

adrianperreault:
@garyttierney Thanks a ton for this! This fixes  #2316.

garyttierney:
The case where a line-comment is moved into the adjacent function body still isn't handled, but I believe this is where the magic should be happening: https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/lang/folding/CustomFoldingBuilder.java#L204-L214.

The Java implementation, for example, does this: https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/codeInsight/folding/impl/JavaFoldingBuilderBase.java#L785-L793.

It looks like we can fix the same case for Rust by checking the element type we're looking at in here (cc'ing @anton-okolelov who originally started this work).

garyttierney:
I've pushed a patch that handles custom regions correctly when the line comment is attached to a function, but it needs more test cases to make sure it still works correctly under patterns, structs, etc.

anton-okolelov:
Thanks a lot!

garyttierney:
I think this probably has sufficient coverage for it to be working in most cases now.

Undin:
Excellent!
Thanks a lot!
bors r+

PR #3008, opened 2018-11-01
Bundled pretty-printer for LLDB: tuple and enum support

ortem:
This PR adds  `tuple` and `enum` support to LLDB pretty-printers.

screen shot 2018-11-01 at 19 31 07
screen shot 2018-11-01 at 19 31 21

This implementation still doesn't work in some cases due to compilers' (debuggers'?) bugs. I've opened the corresponding [issue](https://github.com/rust-lang/rust/issues/55586).

### DIFF
--- a/prettyPrinters/lookup.py
+++ b/prettyPrinters/lookup.py
@@ -5,22 +5,47 @@ from providers import *
 
 
 class RustType:
-    OTHER = 0
-    STD_VEC = 1
-    STD_STRING = 2
-    STD_STR = 3
+    OTHER = "Other"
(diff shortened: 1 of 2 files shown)


amakeev:
Cool!
Would it be possible to show names for structure/tuple-like variants as well? (`C` and `D`)

ortem:
@amakeev Now I don't know, I'm working on it :)

Undin:
LGTM
bors r+

ice1000:
When can we have this in gdb?

PR #3012, opened 2018-11-05
Feature/ij rust unwrap to match intention

sohich:
Base case - transform this:
```rust
let a = a.unwrap/*caret*/()
```
to this:
```rust
let a = match a {
    Ok(x) => x,
    Err(_) => unimplemented!(),
};
```

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/UnwrapToMatchIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/UnwrapToMatchIntention.kt
@@ -1,0 +1,68 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.ide.intentions.UnwrapToMatchIntention.ReceiverType.*
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorOrSelf
+import org.rust.lang.core.psi.ext.parentDotExpr
+import org.rust.lang.core.psi.ext.receiver
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.type
+
+class UnwrapToMatchIntention: RsElementBaseIntentionAction<UnwrapToMatchIntention.Context>() {
+    override fun getText() = "Replace .unwrap() with match"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val methodCall = element.ancestorOrSelf<RsMethodCall>() ?: return null
+        val itemType = (methodCall.receiver.type as? TyAdt)?.item as? RsEnumItem ?: return null
+        val enumType = getReceiverType(itemType) ?: return null
+
+        if (methodCall.referenceName == "unwrap" &&
+            methodCall.typeArgumentList == null &&
+            methodCall.valueArgumentList.exprList.isEmpty()) {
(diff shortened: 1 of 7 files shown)


sohich:
My tests fail and I need some help with these ones.
The point is that `findApplicableContext` method checks `.unwrap()` method call receiver's type and requires it to be `Option` or `Result`. 
It works fine while debugging in IDE but in tests even the simpliest cases' types are not inferred at all. However, I checked `FillMatchArmsIntention` tests that proceed type checks too and it passes as well. `What can I do with this?

Undin:
@sohich your tests require setup stdlib because `Option` and `Result` types are part of it.
Try to add @ProjectDescriptor(WithStdlibRustProjectDescriptor::class) annotation on test methods which require stdlib or on whole test class if each test require stdlib

sohich:
@Undin Worked for me, thanks a lot!

src/main/kotlin/org/rust/ide/intentions/UnwrapToMatchIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+            append(if (isOption) "None" else "Err(_)")
+            append(" => unimplemented!(),")
+            append("}")
+        }
+
+        val matchExpression = RsPsiFactory(project).createExpression(generatedCode) as RsMatchExpr
+        methodCall.parentDotExpr.replace(matchExpression)
+    }
+
+    private fun isStdOptionOrResult(item: RsEnumItem): Boolean? {
```
vlad20012: I'd recommend to crate a private enum like `ResultKind` with 2 variants and return `ResultKind?` here. Also it's possible to add properties like `okVariant` (`Some`/`Ok`) and `errorVariant` to that enum.

sohich:
@Undin, I need your help again. Your solution with `@ProjectDescriptor` has worked well.
But when I launch tests by myself, I have problems with 100% of tests that are marked with this annotation.
I get following runtime exception every time:
```
ERROR: Stub index points to a file without PSI: 
com.intellij.openapi.fileTypes.UnknownFileType@6ca61c60, used scope Project and Libraries

java.lang.Throwable: Stub index points to a file without PSI:
com.intellij.openapi.fileTypes.UnknownFileType@6ca61c60, used scope Project and Libraries
```

I have invalidated caches, then I installed last version of IDEA, reinstalled `rustup` and `cargo`. 
But all these steps haven't done the job.

Issue reproduces not only in my tests (checked  tests `FillMatchWithArmsIntentionTest` that also use this annotation.
Is it known problem? What can I do? CI tests worked as well.

src/main/kotlin/org/rust/ide/intentions/UnwrapToMatchIntention.kt
```diff
@@ -0,0 +1,68 @@
...
+            else -> null
+        }
+    }
+
+    data class Context (
+        val methodCall: RsMethodCall,
+        val receiverType: ReceiverType
+    )
+
+    enum class ReceiverType(val valueBranch: String, val nonValueMatcher: String) {
```
vlad20012: class can be private?

sohich: @vlad no it can't because data class `Context` returns its instance. `Context` instance is returned by intention's public method.

vlad20012: Okay (

sohich:
> Could you please squash all the commits and use rebase instead of merge? Then we'll merge it. 

I'm not sure if I have understood you completely. I have squashed all my commits to one. But what do you mean by using rebase instad of merge? I had rebased my branch onto `master` before but hadn't used any merges

vlad20012:
> but hadn't used any merges

Oh, I'm sorry, maybe I mixed up it with some other PR.

Thanks!
bors r+

vlad20012:
bors retry

PR #3017, opened 2018-11-06
Bundled pretty-printer for LLDB: VecDeque support

ortem:
This PR adds `VecDeque` support to LLDB pretty-printers.

vd1
vd2

Merge after #3008.

### DIFF
--- a/prettyPrinters/lookup.py
+++ b/prettyPrinters/lookup.py
@@ -17,14 +17,16 @@ class RustType:
     COMPRESSED_ENUM = "CompressedEnum"
     REGULAR_UNION = "RegularUnion"
 
-    STD_VEC = "StdVec"
     STD_STRING = "StdString"
     STD_STR = "StdStr"
+    STD_VEC = "StdVec"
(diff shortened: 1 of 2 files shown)


Undin:
I haven't reviewed it yet but at first glance, it seems that you add all code related to `Rc` and `Arc` in this PR.
Could you please leave changes related to `VecDec` only?

ortem:
@Undin Done.

Undin:
The current implementation doesn't take into account that `VecDeque` uses internal vector as ring buffer, i.e. `head` can be less than `tail` and vice versa.

If we run this code:
```rust
use std::collections::VecDeque;

fn main() {
    let mut deque = VecDeque::new();
    for i in 1..8 {
        deque.push_back(i)
    }
    deque.pop_front();
    deque.push_back(8);
    println!("{:?}", deque);
}
```
we get the following
image

ortem:
@Undin Thanks for this note! I just cloned the original implementation, so I opened the corresponding [issue](https://github.com/rust-lang/rust/issues/55944).

PR #3024, opened 2018-11-07
Bundled pretty-printer for LLDB: Rc, Arc support

ortem:
This PR adds `Rc` and `Arc` support to LLDB pretty-printers.

rc1
rc2

Merge after #3017.

### DIFF
--- a/prettyPrinters/providers.py
+++ b/prettyPrinters/providers.py
@@ -1,4 +1,4 @@
-from lldb import SBValue, SBError
+from lldb import SBValue, SBData, SBError, eBasicTypeLong, eBasicTypeUnsignedLong
 from lldb.formatters import Logger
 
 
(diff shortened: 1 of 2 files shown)


prettyPrinters/providers.py
```diff
@@ -266,3 +266,83 @@ def update(self):
...
+        if name == "weak":
+            return 1
+        if name == "value":
+            return 2
+        return -1
+
+    def get_child_at_index(self, index):
+        # type: (int) -> SBValue
+        if index == 0:
+            count = self.strong.GetValueAsSigned()
```
Undin: Looks like `GetValueAsUnsigned` should be called here

prettyPrinters/providers.py
```diff
@@ -266,3 +266,83 @@ def update(self):
...
+            data = SBData.CreateDataFromUInt64Array(self.endianness, self.pointer_size, [count])
+            return self.valobj.CreateValueFromData("strong", data, self.count_type)
+        if index == 1:
+            count = self.weak.GetValueAsSigned()
+            data = SBData.CreateDataFromUInt64Array(self.endianness, self.pointer_size, [count])
+            return self.valobj.CreateValueFromData("weak", data, self.count_type)
```
Undin: It seems we should move code about `usize` to `SBValue` conversion into a separate function because we already have two similar pieces of code in this PR. Also, there is another one in #3037

Undin: Probably, it should be a method of some new class because we also need to store process properties too (`endianness` and `pointer_size`)

prettyPrinters/providers.py
```diff
@@ -266,3 +266,83 @@ def update(self):
...
+
+    def num_children(self):
+        # type: () -> int
+        return 3
+
+    def get_child_index(self, name):
+        # type: (str) -> int
+        if name == "strong":
+            return 0
+        if name == "weak":
```
Undin: Do we really need `strong` and `weak` children at all? For example, formatter for C++ `shared_ptr` doesn't provide such children, only add them into summary

prettyPrinters/providers.py
```diff
@@ -266,3 +266,83 @@ def update(self):
...
+        if index == 0:
+            count = self.strong.GetValueAsSigned()
+            data = SBData.CreateDataFromUInt64Array(self.endianness, self.pointer_size, [count])
+            return self.valobj.CreateValueFromData("strong", data, self.count_type)
+        if index == 1:
+            count = self.weak.GetValueAsSigned()
+            data = SBData.CreateDataFromUInt64Array(self.endianness, self.pointer_size, [count])
+            return self.valobj.CreateValueFromData("weak", data, self.count_type)
+        if index == 2:
+            return self.value
```
Undin: If we keep `strong` and `weak` children, maybe make `value` child the first because it contains the most important info?

prettyPrinters/providers.py
ortem: Yes, C++ pretty-printer uses the interesting trick:
```
def num_children(self):
    return 1

def get_child_at_index(self, index):
    if index == 0: ...
    if index == 1: ...
    if index == 2: ...
```
So the synthtetic provider doesn't show the 1st and 2nd children, but summary provider can access (and show) these.

I will use it too.

prettyPrinters/providers.py
ortem: Good idea! I've created `ValueBuilder` for such needs.

PR #3031, opened 2018-11-11
Show a list of cargo features and their state in the Cargo tool window

kumbayo:
Display a list of features and their enabled/disabled state for each package in the Cargo tool window.

This depends on #2840, which provides the necessary feature information.


kumbayo:
This is how it currently looks like:
image

PR #3032, opened 2018-11-12
[WIP] T: Tests for Test Runner

mchernyavsky:
TODO:
- [ ] Tests for multi-target configuration

### DIFF
--- a/src/test/kotlin/org/rust/cargo/runconfig/TestRunnerTest.kt
+++ b/src/test/kotlin/org/rust/cargo/runconfig/TestRunnerTest.kt
@@ -1,0 +1,370 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.runconfig
+
+import com.intellij.execution.actions.ConfigurationContext
+import com.intellij.execution.configurations.RunConfiguration
+import com.intellij.execution.executors.DefaultRunExecutor
+import com.intellij.execution.runners.ExecutionEnvironmentBuilder
+import com.intellij.execution.testframework.sm.runner.SMTestProxy
+import com.intellij.execution.testframework.sm.runner.ui.SMTRunnerConsoleView
+import com.intellij.ide.DataManager
+import com.intellij.openapi.application.impl.LaterInvocator
+import com.intellij.openapi.util.Disposer
+import org.rust.cargo.RustWithToolchainTestBase
+import org.rust.cargo.runconfig.command.CargoCommandConfiguration
+import org.rust.cargo.runconfig.test.CargoTestRunConfigurationProducer
+import org.rust.stdext.removeLast
+
+class TestRunnerTest : RustWithToolchainTestBase() {
+
+    fun `test not executed tests are not shown`() {
+        val testProject = buildProject {
+            toml("Cargo.toml", """
+                [package]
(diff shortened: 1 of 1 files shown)


Undin:
bors try

mchernyavsky:
bors try

mchernyavsky:
bors try

src/test/kotlin/org/rust/cargo/runconfig/TestRunnerTest.kt
```diff
@@ -0,0 +1,370 @@
...
+        assert(test1.duration!! > 1000)
+
+        val test2 = findTestByName("tests::test2", root)!!
+        assert(test2.duration!! < 100)
+
+        val mod = findTestByName("tests", root)!!
+        assert(mod.duration!! == test1.duration!! + test2.duration!!)
+    }
+
+    private fun createTestRunConfigurationFromContext(): CargoCommandConfiguration {
```
Undin: Looks very similar with `org.rust.cargo.runconfig.RunConfigurationTestCase#createRunConfigurationFromContext`. Can we reuse existing code somehow?

src/test/kotlin/org/rust/cargo/runconfig/TestRunnerTest.kt
```diff
@@ -0,0 +1,370 @@
...
+            ?: error("Can't create run configuration")
+    }
+
+    private fun runConfigurationAndGetTestRoot(configuration: RunConfiguration): SMTestProxy.SMRootTestProxy {
+        val executor = DefaultRunExecutor.getRunExecutorInstance()
+        val state = ExecutionEnvironmentBuilder
+            .create(executor, configuration)
+            .build()
+            .state!!
+        val result = state.execute(executor, RsRunner())!!
```
Undin: The first part of this method is almost the same as in `org.rust.cargo.runconfig.RunConfigurationTestCase#execute`. Let's reuse it somehow

src/test/kotlin/org/rust/cargo/runconfig/TestRunnerTest.kt
```diff
@@ -0,0 +1,370 @@
...
+
+        private fun findTestByName(testFullName: String, root: SMTestProxy.SMRootTestProxy): SMTestProxy? {
+            for (child in root.children) {
+                val result = findTestByName(testFullName, child)
+                if (result != null) return result
+            }
+            return null
+        }
+
+        private fun findTestByName(
```
Undin: Personally, I usually prefer to encapsulate recursive part of similar functions in inner function
I mean something like
```kotlin
private fun findTestByName(testFullName: String, root: SMTestProxy.SMRootTestProxy): SMTestProxy? {
    val fullNameBuffer = mutableListOf<String>()

    fun find(test: SMTestProxy): SMTestProxy? {
        if (test !is TestProxyRoot) {
            fullNameBuffer.add(test.name)
        }
        if (testFullName == fullNameBuffer.joinToString("::")) return test
        for (child in test.children) {
            val result = find(child)
            if (result != null) return result
        }
        if (test !is TestProxyRoot) {
            fullNameBuffer.removeLast()
        }
        return null
    }

    return find(root)
}
```

Just a comment, I don't insist 😄

mchernyavsky:
I don't see any reason why this test failed, so
bors retry

PR #3036, opened 2018-11-12
Support raw identifiers

Undin:
The corresponding [RFC](https://github.com/rust-lang/rfcs/blob/master/text/2151-raw-identifiers.md)
* Name resolution takes into account raw identifiers
* Correctly rename module files
* Adjust spell checker and naming inspections
* Add `r#` prefix while completion if needed
* Provide `-Xjvm-default=enable` option to kotlin compiler so now we can use `@JvmDefault` annotation

TODO:
* Support renaming of any named item to keyword with the corresponding escaping. Although our tests work in this case, in real IDE rename dialog doesn't allow such renaming because it calls `RsNamesValidator` before all actions and rejects keywords

### DIFF
--- a/src/main/kotlin/org/rust/ide/spelling/RsNameIdentifierOwnerTokenizer.kt
+++ b/src/main/kotlin/org/rust/ide/spelling/RsNameIdentifierOwnerTokenizer.kt
@@ -10,8 +10,9 @@ import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.spellchecker.inspections.IdentifierSplitter
 import com.intellij.spellchecker.tokenizer.TokenConsumer
 import com.intellij.spellchecker.tokenizer.Tokenizer
+import org.rust.lang.core.psi.RS_RAW_PREFIX
 import org.rust.lang.core.psi.ext.RsNameIdentifierOwner
-import org.rust.lang.core.psi.ext.unescapeIdentifier
+import org.rust.lang.core.psi.unescapeIdentifier
 
 // Inspired by `PsiIdentifierOwnerTokenizer`
 object RsNameIdentifierOwnerTokenizer : Tokenizer<RsNameIdentifierOwner>() {
@@ -31,7 +32,7 @@ object RsNameIdentifierOwnerTokenizer : Tokenizer<RsNameIdentifierOwner>() {
         }
         val text = identifier.text
         val unescapedText = text.unescapeIdentifier()
-        if (text.startsWith("r#")) {
+        if (text.startsWith(RS_RAW_PREFIX)) {
             offset += 2
         }
         consumer.consumeToken(parent, unescapedText, true, offset, TextRange.allOf(unescapedText), IdentifierSplitter.getInstance())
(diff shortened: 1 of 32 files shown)


src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
```diff
@@ -159,8 +159,8 @@ class RenameTest : RsTestBase() {
         pub enum Spam { Quux, Eggs }
     """, """
     //- main.rs
-        use foo::Spam;
-        mod foo;
+        use r#mod::Spam;
+        mod r#mod;
```
Undin: Not sure about this change but at least new behavior produces compilable code

src/main/kotlin/org/rust/lang/core/psi/RsRawIdentifiers.kt
```diff
@@ -0,0 +1,22 @@
...
+const val RS_RAW_PREFIX = "r#"
+val CAN_NOT_BE_ESCAPED = listOf("self", "super", "crate", "Self")
+
+fun String.unescapeIdentifier(): String = removePrefix(RS_RAW_PREFIX)
+fun String.escapeIdentifierIfNeeded(): String =
+    if (isValidRustVariableIdentifier(this) || this in CAN_NOT_BE_ESCAPED) this else "$RS_RAW_PREFIX$this"
+
+val PsiElement.unescapedText: String get() {
+    val text = text ?: return ""
+    return if (this is LeafPsiElement) text.unescapeIdentifier() else text
```
vlad20012: Maybe it's better to check that elementType is IDENTIFIER?

src/main/kotlin/org/rust/ide/spelling/RsNameIdentifierOwnerTokenizer.kt
```diff
@@ -0,0 +1,40 @@
...
+
+// Inspired by `PsiIdentifierOwnerTokenizer`
+object RsNameIdentifierOwnerTokenizer : Tokenizer<RsNameIdentifierOwner>() {
+
+    override fun tokenize(element: RsNameIdentifierOwner, consumer: TokenConsumer) {
+        val identifier = element.nameIdentifier ?: return
+        val range = identifier.textRange
+        if (range.isEmpty) return
+
+        var offset = range.startOffset - element.textRange.startOffset
```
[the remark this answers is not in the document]
Undin: @ortem maybe it's better to be closer to the original code (I mean the code of `PsiIdentifierOwnerTokenizer`)? It can make a future comparison of the original code and our easier for example in case of changes in original code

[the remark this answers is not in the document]
ortem: OK, it makes sense.

src/main/kotlin/org/rust/lang/core/psi/RsRawIdentifiers.kt
Undin: done

Issue #3030, opened 2018-11-11
Mimic cargo test standard behavior for stdout and error


la10736:
## Environment

* **Intellij-Rust plugin version:** v0.2.0.2110-182
* **Rust toolchain version:** 1.30
* **IDE name and version:** Intellij 2018.2.5
* **Operating system:** Linux

## Problem description

Test runner should show stderr and stdout for failed test like `cargo test` command do.

## Steps to reproduce
```rust
#[test]
fn it_works() {
    println!("it_works stdout");
    eprintln!("it_works stderr");
    assert_eq!(2 + 2, 4);
}

#[test]
fn it_fail() {
    println!("it_fail stdout");
    eprintln!("it_fail stderr");
    assert_eq!(2 + 2, 5);
}
```
Create a test runner and run it. Select `it_fail` test should show stdout and stderr like `cargo test` do.

Now, cargo json output contains the stdout and stderr:

```none

michele@DartVader:~/learning/rust/cargo_test_json_out$ cargo test --color=always --no-fail-fast --all-features -- -Z unstable-options --format=json
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s                                                                                                                         
     Running target/debug/deps/cargo_test_json_out-4851566aeccea52d
{ "type": "suite", "event": "started", "test_count": 2 }
{ "type": "test", "event": "started", "name": "it_fail" }
{ "type": "test", "event": "started", "name": "it_works" }
{ "type": "test", "name": "it_works", "event": "ok" }
[...]

PR #3038, opened 2018-11-13
RUN: Show stdout for failed tests

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3030.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
@@ -111,13 +111,17 @@ class CargoTestEventsConverter(
                 messages.add(createTestStartedMessage(testMessage.name))
             }
             "ok" -> {
-                messages.add(createTestFinishedMessage(testMessage.name, getTestDuration(testMessage.name)))
+                val duration = getTestDuration(testMessage.name)
+                messages.add(createTestFinishedMessage(testMessage.name, duration))
                 recordSuiteChildFinished(testMessage.name)
                 processFinishedSuites(messages)
             }
             "failed" -> {
-                messages.add(createTestFailedMessage(testMessage.name, getTestResult(testMessage.stdout)))
-                messages.add(createTestFinishedMessage(testMessage.name, getTestDuration(testMessage.name)))
+                val duration = getTestDuration(testMessage.name)
+                val (stdout, failedMessage) = parseFailedTestOutput(testMessage.stdout ?: "")
+                if (stdout.isNotEmpty()) messages.add(createTestStdOutMessage(testMessage.name, stdout + '\n'))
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
```diff
@@ -257,10 +258,22 @@ class CargoTestEventsConverter(
...
+                .addAttribute("out", stdout)
+
         private fun createTestCountMessage(testCount: String): ServiceMessageBuilder =
             ServiceMessageBuilder("testCount")
                 .addAttribute("count", testCount)
 
+        private fun parseFailedTestOutput(output: String): FailedTestOutput {
+            val partitionPredicate: (String) -> Boolean = { !it.trimStart().startsWith("thread", true) }
+            val stdout = output.lineSequence().takeWhile(partitionPredicate).joinToString("\n")
+            val failedMessage = output.lineSequence().dropWhile(partitionPredicate).joinToString("\n")
```
Undin: Looks like you can use `Sequence.partition` here

mchernyavsky: This name is deceptive :)
> Splits the original sequence into pair of lists, where first list contains elements for which predicate yielded `true`, while second list contains elements for which predicate yielded `false`.

matklad:
Hm, I think I disagree about this change in behavior :-) Original divergence from Cargo's defaults was very intentional. 

Cargo's interface is tailored for a command line usage, and, on the command line, you typically execute tests as `cargo test`. This runs all of the tests, without any UI affordances for a fancier output that a "wall of text". In this situation, hiding stdout/stderr by default can be reasonable (although I personally still find it more confusing than helpful). 

In IDE, the situation is quite different. *First*, the user typically runs a subset of the tests, or a single test only. Because of this, the total amount of information is much smaller. *Second*, we can smartly group tests by module, crate, etc, which makes it easy to slice the "wall of text" into manageable pieces. *Third*, test output view is a rather non-central panel in IDE (unlike a typical full-screen terminal), so flooding it with lots of messages it not a problem in the first place.


OTOH, *not* showing the output is quite confusing, especially for beginners. So, I think that it makes sense to stick with `RUST_LOG=short` `--no-capture` *by default*, unless many users explicitly demand the silent behavior :)

matklad:
Or maybe I am misunderstanding what this PR does? To reforumulate, if I run

```
#[test]
fn test_stdout() {
    println!("hello, World")
}
```

in IntelliJ Rust today, I don't see "Hello, World" anywhere, and it seems wrong to me

mchernyavsky:
@matklad Without `--nocapture` flag, cargo shows stdout/stderr for failed tests only. We don't use `--nocapture` because of https://github.com/rust-lang/rust/issues/54669. It will be fixed in https://github.com/intellij-rust/intellij-rust/pull/4139.

Issue #2630, opened 2018-07-02
Debugging is not possible - More than one binary produced. Please specify `--bin`, `--lib`, `--test` or `--example` flag explicitly.


brianjob:
I've seen others with this same error, but none of their solutions seem to work for me.

This occurs when I try to debug tests in CLion. I can run the tests fine, but debugging causes this error.

rust toolchain: nightly-2018-05-16-x86_64-unknown-linux-gnu

I have tried with both LLDB and GDB.

![screenshot from 2018-07-02 16-42-44](https://user-images.githubusercontent.com/1531826/42185476-09e35d90-7e17-11e8-9993-d763ff0abbc9.png)

nlinker:
For me also this happens only with _tests_.
Debugging `main.rs` works fine.

brianjob:
@nlinker Yes, same with me. I can debug main.rs just fine. This only occurs when debugging tests.

Undin:
Could you provide command from your run configuration and intellij-rust plugin version?
<img width="749" alt="screen shot 2018-07-05 at 10 17 04" src="https://user-images.githubusercontent.com/2539310/42307944-c0b311ce-803c-11e8-8b43-aabd863d3cce.png">

brianjob:
run configuration command: `test --package payments --test lib unit::scheduler_helper::is_leap_year -- --exact`

intellij-rust plugin version: `0.2.0.2099-181`

Undin:
@brianjob 
Hm, interesting
Could you also provide result of `cargo build --package payments --test lib --message-format=json` command?

jendrikw:
Where do I have to add the flag?

Undin:
@jendrikw what flag do you mean?

jendrikw:
The ones from the error message (--bin, --lib, --test or --example).

brianjob:
Output from `cargo test --no-run --package payments --test lib --message-format=json`:

[output.txt](https://github.com/intellij-rust/intellij-rust/files/2176954/output.txt)

AsafMah:
I have the same problem in windows, this is the command line:
run --package advent2017 --bin advent2017
My toolchain is:
nightly-x86_64-pc-windows-gnu

ivanovaleksey:
@Undin I try to debug configuration added automatically by the plugin and get the same error.

nlinker:
Reproducible on 
**plugin version:** Version: 0.2.0.2101-182
[...]

PR #3042, opened 2018-11-15
RUN: Fix run tests in debug

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/2630.

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
@@ -67,10 +67,11 @@ class RsDebugRunner : AsyncProgramRunner<RunnerSettings>() {
         val cmd = state.commandLine
         val (commandArguments, executableArguments) = CargoArgsParser.parseArgs(cmd.command, cmd.additionalArguments)
 
-        val buildCommand = if (cmd.command == "run") {
-            cmd.copy(command = "build", additionalArguments = commandArguments)
-        } else {
+        val isTestRun = cmd.command == "test"
+        val buildCommand = if (isTestRun) {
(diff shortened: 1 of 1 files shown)


debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
```diff
@@ -185,8 +191,12 @@ private fun buildProjectAndGetBinaryArtifactPath(project: Project, command: Carg
                                     || kind == TargetKind.TEST
                                     || (kind == TargetKind.LIB && profile.test)
                             }
-                            .flatMap { it.filenames.filter { !it.endsWith(".dSYM") } } // FIXME: correctly launch debug binaries for macos
-                            .let(DebugBuildResult::Binaries)
+                            .filter { (_, profile) -> !testsOnly || profile.test }
```
Undin: Maybe merge this `filter` with previous? Looks like both conditions are about similar things

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
```diff
@@ -167,7 +173,7 @@ private fun buildProjectAndGetBinaryArtifactPath(project: Project, command: Carg
                         }
 
                         val parser = JsonParser()
-                        result = output.stdoutLines
+                        result = output.stdoutLines.asSequence()
```
Undin: Maybe `asSequence` doesn't need here? Looks like it's not a hot place in our code where we can spend a lot of time only for intermediate allocations but using sequences always make debugging significantly less convenient
Instead of this, we can collapse some sequential calls of `mapNotNull` and `filter`

Issue #2530, opened 2018-05-14
False positive E0277 "trait bound `Sized` not satisfied"


rrevenantt:
False positive E0277 in rust standard library in iterator.rs:

![image](https://user-images.githubusercontent.com/8453142/40023777-d6e9ffd8-57d4-11e8-8b50-c659080398d1.png)
\
\
intellij-rust from master
Intellij IDEA 2018.1.1
rust 1.26 stable
Manjaro Linux

ArtsiomCh:
Same issue for mutex.rs as well:
![image](https://user-images.githubusercontent.com/14268320/54303842-47006880-459a-11e9-8c67-d9dd9f13f13b.png)

IntelliJ IDEA 2019.1 EAP (Community Edition)
Build #IC-191.5849.21, built on February 27, 2019
JRE: 1.8.0_202-release-1483-b31 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Linux 4.4.0-142-generic
Rust plugin v0.2.93.2117-191

CAD97:
cc #2783, the apparent cause

minimal cases:

- impl header, where bound

```rust
struct Wrapper<T: ?Sized>(T);

impl<T: ?Sized> Wrapper<T> {
    fn into_inner(self) -> T
    where T: Sized
    { self.0 }
}
```

- fn header, fn header

```rust
fn foo<T: ?Sized + Sized>(t: T) -> T { t }
```

PR #3049, opened 2018-11-17
fixed issue 2530

vlad9486:
The fix is bit ugly, but it works and tests are passed. I left a comment in the code describing a possible problem. Hope, soon will be available a proper fix, but this is also good for me.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
@@ -72,22 +72,16 @@ fun RsTraitItem.searchForImplementations(): Query<RsImplItem> {
 }
 
 private val RsTraitItem.superTraits: Sequence<BoundElement<RsTraitItem>> get() {
-    val bounds = typeParamBounds?.polyboundList.orEmpty().asSequence()
-    return bounds.mapNotNull { it.bound.traitRef?.resolveToBoundTrait }
-}
-
-val RsTraitItem.isSized: Boolean get() {
-    val stub = stub
-    if (stub != null) return implementedTrait?.flattenHierarchy.orEmpty().any { it.element.isSizedTrait }
-
     val whereBounds =
         whereClause?.wherePredList.orEmpty()
             .filter { (it.typeReference?.typeElement as? RsBaseType)?.path?.hasCself == true }
             .flatMap { it.typeParamBounds?.polyboundList.orEmpty() }
     val bounds = typeParamBounds?.polyboundList.orEmpty() + whereBounds
-    // ugly solution, the trait might be refer as std::marker::Sized, or in other way
-    // TODO: parse it properly
-    return bounds.any { it.text.contains(Regex("Sized")) && it.q == null }
+    return bounds.asSequence().mapNotNull { it.bound.traitRef?.resolveToBoundTrait }
+}
+
+val RsTraitItem.isSized: Boolean get() {
+    return implementedTrait?.flattenHierarchy.orEmpty().any { it.element.isSizedTrait }
 }
 
 fun RsTraitItem.withSubst(vararg subst: Ty): BoundElement<RsTraitItem> {
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
@@ -82,7 +82,7 @@ val RsTraitItem.isSized: Boolean get() {
 
     val whereBounds =
         whereClause?.wherePredList.orEmpty()
-            .filter { (it.typeReference?.typeElement as? RsBaseType)?.name == "Self" }
+            .filter { (it.typeReference?.typeElement as? RsBaseType)?.path?.hasCself == true }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
```diff
@@ -76,6 +76,20 @@ private val RsTraitItem.superTraits: Sequence<BoundElement<RsTraitItem>> get() {
     return bounds.mapNotNull { it.bound.traitRef?.resolveToBoundTrait }
 }
 
+val RsTraitItem.isSized: Boolean get() {
+    val stub = stub
+    if (stub != null) return implementedTrait?.flattenHierarchy.orEmpty().any { it.element.isSizedTrait }
```
vlad20012: Hmm. Obviously your solution will work only if the file with the trait is opened in an editor. Really you will not need this special case if write the code without accessing non-stubbed elements and a text of elements. (you can read more about stubs in ARCHITECTURE.md).

src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
```diff
@@ -76,6 +76,20 @@ private val RsTraitItem.superTraits: Sequence<BoundElement<RsTraitItem>> get() {
...
+
+    val whereBounds =
+        whereClause?.wherePredList.orEmpty()
+            .filter { (it.typeReference?.typeElement as? RsBaseType)?.name == "Self" }
```
vlad20012: .path.isCself

src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitItem.kt
```diff
@@ -76,6 +76,20 @@ private val RsTraitItem.superTraits: Sequence<BoundElement<RsTraitItem>> get() {
...
+            .flatMap { it.typeParamBounds?.polyboundList.orEmpty() }
+    val bounds = typeParamBounds?.polyboundList.orEmpty() + whereBounds
+    // ugly solution, the trait might be refer as std::marker::Sized, or in other way
+    // TODO: parse it properly
+    return bounds.any { it.text.contains(Regex("Sized")) && it.q == null }
```
vlad20012: 1. `it.bound.traitRef.resolveToTrait.isSizedTrait`
2. Hmm, `it.q == null` is a more complex stuff. We have to create new stub implementation for `RsPolybound` and write boolean flag `hasQ` to it. Example commits that add stubs ([1](https://github.com/intellij-rust/intellij-rust/commit/8513a547cdce4b898613d667094f2c4df6b8d3b1), [2](https://github.com/intellij-rust/intellij-rust/commit/625c54dfcb3bdbbbea3c1cb2e3783c3744db5ffd), [3](https://github.com/intellij-rust/intellij-rust/commit/00abf615f7c395e408b401902c7903859b4a5211))

vlad9486: The question mark stub is important, but more the important thing is the RsTraitRef.resolveToBoundTrait method does not resolve KnownItems such as Sized or Clone, etc., I cannot found the cause. Looks like it worked when the check it.element.isSizedTrait was added.

Debugging leads me [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt#L61)... in our case the element.parent is not RsPathExpr, so it returns nothing. I think the resolving should be fixed.

We can omit the question mark check for now, I think it is not relevant for traits.

vlad20012: If the problem occurs in tests only, you probably have to add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` annotation to test method

vlad9486: The is the root cause of the issue 2530, I fixed it using `it.text.contains(Regex("Sized"))` check, but it is wrong. I will try to find the commit that brokes resolution. It might take a while.

vlad20012: Please, provide an exact test that shows where name resolution is broken. In the case of test contained in this PR it's enough to add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` to fix it.

vlad9486: The hook is not taking into account types constructed from `Self`, for example `(Self, Self)` considered as not sized.

vlad20012:
Could you please sign [the CLA](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#cla) and write you name to CONTRIBUTORS.txt?

As I see, the PR is almost done (except tests) and I can finish it for you.
Again, I think it's enough to add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` annotation to the single failed tests and everything will start to work

vlad20012:
Without  significant changes, I've included it into #3243

But it is not a fix for #2530. It's a bit harder to take into account fn-level Self bounds because it changes how we think about Self. Currently we have an ugly implementation with much differences from the rust compiler

Issue #3016, opened 2018-11-06
Run/Debug output inconsistencies


mchernyavsky:
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2464-182-nightly
* **Rust toolchain version:** 1.31.0-nightly
* **IDE name and version:** CLion 2018.2.5
* **Operating system:** macOS Mojave 10.14.1

## Problem description

When debugging a bin:
* Jumping to next/previous compilation errors (`⌥ + ⌘ + ↓/↑`) doesn't work.
* Some hyperlinks are missing. Example (run / debug):

<img width="300" alt="Run" src="https://user-images.githubusercontent.com/6079006/48068281-74e83180-e1e3-11e8-877b-32fe9b675616.png"> <img width="300" alt="Debug" src="https://user-images.githubusercontent.com/6079006/48068311-89c4c500-e1e3-11e8-814b-6d01533ffb49.png">

* Stack backtrace is not shown. Example (run / debug):

<img width="300" alt="Run" src="https://user-images.githubusercontent.com/6079006/48069008-5a16bc80-e1e5-11e8-8048-792fec14283a.png"> <img width="300" alt="Debug" src="https://user-images.githubusercontent.com/6079006/48069076-82062000-e1e5-11e8-8d27-089fa58bf0a4.png">

* ???

When debugging tests:
* All of the above
* [Test Runner](https://www.jetbrains.com/help/idea/test-runner-tab.html) is not used.
* The `--no-fail-fast` option is not passed to the Cargo.

PR #3052, opened 2018-11-18
RUN: Fix debug output and use test runner in debug mode

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3016.

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
@@ -34,14 +34,13 @@ import org.jetbrains.concurrency.AsyncPromise
 import org.jetbrains.concurrency.Promise
 import org.rust.cargo.project.workspace.CargoWorkspace.CrateType
 import org.rust.cargo.project.workspace.CargoWorkspace.TargetKind
-import org.rust.cargo.runconfig.CargoRunState
+import org.rust.cargo.runconfig.CargoRunStateBase
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
+import org.rust.cargo.toolchain.Cargo
 import org.rust.cargo.toolchain.CargoCommandLine
 import org.rust.cargo.toolchain.impl.CargoMetadata
 import org.rust.cargo.util.CargoArgsParser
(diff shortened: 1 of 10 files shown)


debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunner.kt
```diff
@@ -74,20 +73,22 @@ class RsDebugRunner : AsyncProgramRunner<RunnerSettings>() {
             cmd.copy(command = "build", additionalArguments = commandArguments)
         }
 
-        val runCommand = { executablePath: Path ->
-            GeneralCommandLine(executablePath)
-                .withWorkDirectory(cmd.workingDirectory)
-                .withParameters(executableArguments)
-                .withEnvironment(cmd.environmentVariables.envs)
+        val createGeneralCommandLine = { executablePath: Path ->
```
vlad20012: `runCommand` was more meaningful IMO. These was `buildCommand` that builds executable, and `runCommand` that execute produced binary (accepting a path to it). Now `createGeneralCommandLine` tell me only about the type it produces, but not about its meaning

src/main/kotlin/org/rust/cargo/runconfig/CargoTestRunState.kt
```diff
@@ -13,31 +13,29 @@ import com.intellij.execution.configurations.RunConfiguration
...
+import org.rust.cargo.runconfig.console.CargoTestConsoleBuilder
 import org.rust.cargo.toolchain.CargoCommandLine
 
 class CargoTestRunState(
     environment: ExecutionEnvironment,
     config: CargoCommandConfiguration.CleanConfiguration.Ok
 ) : CargoRunState(environment, config) {
 
+    init {
+        consoleBuilder = CargoTestConsoleBuilder(environment.runProfile as RunConfiguration, environment.executor)
```
vlad20012: Hmm, we creating `CargoConsoleBuilder` inside `CargoRunState` constructor and adding filters to it, and then replacing `consoleBuilder` with `CargoTestConsoleBuilder` (and again adding filters to it). Maybe we can make it a bit clearer, for example create abstract method `creaeConsoleBuilder` and so call it once.

PR #3053, opened 2018-11-18
PERF: don't switch to AST during completion

vlad20012:


src/main/kotlin/org/rust/ide/presentation/RsPsiRendering.kt
```diff
@@ -0,0 +1,172 @@
...
+            sb.append(patText ?: "_")
+            sb.append(": ")
+            typeReference?.let { sb.appendTypeReference(it) }
+        }
+        append(")")
+    }
+}
+
+private fun renderTypeReference(ref: RsTypeReference): String =
+    StringBuilder().apply { appendTypeReference(ref) }.toString()
```
Undin: Maybe `buildString { appendTypeReference(ref) }`?

src/main/kotlin/org/rust/ide/presentation/RsPsiRendering.kt
```diff
@@ -0,0 +1,172 @@
...
+
+        is RsFnPointerType -> {
+            append("fn")
+            appendValueParameterListTypes(type.valueParameterList.valueParameterList)
+            appendRetType(type.retType)
+        }
+
+        is RsTraitType -> {
+            append(if (type.isImpl) "impl " else "dyn ")
+            type.polyboundList.joinToWithBuffer(this, "+") { sb ->
```
Undin: Looks like we need use `" + "` here.
image

src/main/kotlin/org/rust/ide/presentation/RsPsiRendering.kt
```diff
@@ -0,0 +1,172 @@
...
+                }
+            }
+        }
+    }
+}
+
+private fun StringBuilder.appendPath(path: RsPath) {
+    append(path.referenceName)
+    val inAngles = path.typeArgumentList
+    val fnSugar = path.valueParameterList
```
Undin: Let's use more common names like `typeArguments` and `valueParameters` instead of `inAngles` and `fnSugar ` respectively - now it takes some time to understand what they mean especially when I read the end of method

src/main/kotlin/org/rust/ide/presentation/RsPsiRendering.kt
```diff
@@ -0,0 +1,172 @@
...
+
+package org.rust.ide.presentation
+
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.ext.RsBaseTypeKind.*
+import org.rust.stdext.joinToWithBuffer
+
+/** Return text of the element without switching to AST (loses non-stubbed parts of PSI) */
+val RsTypeReference.stubSafeText: String
```
Undin: Don't like this name. Maybe "presentation" is better?

src/main/kotlin/org/rust/lang/core/psi/ext/RsBaseType.kt
```diff
@@ -6,14 +6,43 @@
...
+        RsBaseTypeKindEnum.NEVER -> RsBaseTypeKind.BtNever
+        RsBaseTypeKindEnum.UNDERSCORE -> RsBaseTypeKind.BtUnderscore
+        RsBaseTypeKindEnum.PATH -> RsBaseTypeKind.BtPath(path ?: error("Malformed RsBaseType element: `$text`"))
+    }
+
+sealed class RsBaseTypeKind {
+    object BtUnit : RsBaseTypeKind()
+    object BtNever : RsBaseTypeKind()
+    object BtUnderscore : RsBaseTypeKind()
+    data class BtPath(val path: RsPath) : RsBaseTypeKind()
```
Undin: Do we really need `Bt` prefix? Looks like it doesn't add any additional info because it's not obvious that `Bt` means `BaseType`. I'd prefer to use this types like `RsBaseTypeKind.Unit` - it's more informative although longer

src/main/kotlin/org/rust/lang/core/psi/ext/RsBaseType.kt
```diff
@@ -6,14 +6,43 @@
...
+    return when (child?.elementType) {
+        LPAREN -> RsBaseTypeKindEnum.UNIT
+        EXCL -> RsBaseTypeKindEnum.NEVER
+        UNDERSCORE -> RsBaseTypeKindEnum.UNDERSCORE
+        PATH -> RsBaseTypeKindEnum.PATH
+        else -> error("Malformed RsBaseType element: `$text`")
+    }
+}
+
+enum class RsBaseTypeKindEnum {
```
Undin: Awful name 😆 
Maybe use some other synonym of `kind`? Like `form` or something else (I can't come up with good synonym)? But I don't insist - anyway we usually shouldn't use anywhere except stubs

src/main/kotlin/org/rust/lang/core/psi/ext/RsBaseType.kt
vlad20012: There is `kotlin.Unit` type... :thinking:

src/main/kotlin/org/rust/lang/core/psi/ext/RsBaseType.kt
vlad20012: Maybe RsBaseTypeStubKind?

src/main/kotlin/org/rust/ide/presentation/RsPsiRendering.kt
vlad20012: It's same as `impl Trait` [ty rendering](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt#L112)

vlad20012:
> Interesting question: can we reuse methods from RsPsiRendering in documentation provider

Hmm, docs don't have restriction for stub-only access? (b/c it should read doc comments at least)

src/main/kotlin/org/rust/ide/presentation/RsPsiRendering.kt
Undin: So we find two places where we need to add spaces around `+` 😄 
Now it looks inconsistently with other separators like `:`, `,` and `->` where we use spaces

src/main/kotlin/org/rust/lang/core/psi/ext/RsBaseType.kt
Undin: That's why I suggest using these classes/objects like `RsBaseTypeKind.Unit` instead of `Unit`

src/main/kotlin/org/rust/lang/core/psi/ext/RsBaseType.kt
Undin: looks ok

PR #3015, opened 2018-11-06
Index macro expansions

vlad20012:
Now most macros are expanded once (together with indexing) and stored persistently. This opens the way to index macro expansions, and thus support impls/modules declared with macros, and also make auto-import, find symbol/usages (and other index-driven features) to work with macros. 

Also it should significantly increases performance.

Closes #2458

TODO:
- [x] support macro call insertion (bug)
- [x] replace java-serialization with something more meaningful
- [ ] compress expansion-file urls
- [x] fix spawning tons of background tasks
- [x] use write-priority retryable read actions
- [x] support modules defined with a macro (improve module index)
- [x] add possibility for auto-import to import items from macros
- [ ] split expansion task into several batches (?)
- [x] validate expanded macros sequentially (don't re-expand macros that may be invalidated by other expansions in same task)
- [ ] disable inspections/annotators while expanding macros
- [x] think about project root / search scopes
- [x] re-validate only workspace macros on local change
- [x] clear dangling expansions (check consistency) on startup
- [x] fix expansion of not-item macros
- [x] fix performance tests
- [ ] add completion tests
- [ ] clear macros directory if not accessed for a long time

PR #3055, opened 2018-11-18
TY: Infer type of macros like cfg!() column!(), module_path!(), ...

kumbayo:
This macros are built into the compiler and the return types
are not visible from the macro definition in stdlib.
Thus we have to hardcode the return type.

For include_bytes!() we infer the byte array size to be unknown.
If necessary we could check the actual file size, as we can already
resolve the path to the file.

The following built-in macros are not yet typified as they are more complicated to typify:
- compile_error
- include
- concat_idents


vlad20012:
LGTM, thanks!
Really I think I'll make honest expansion for such macros right after landing #3015
bors r+

PR #3074, opened 2018-11-22
RUN: Add test tool window switch

mchernyavsky:


src/main/kotlin/org/rust/cargo/project/configurable/RsProjectConfigurable.kt
```diff
@@ -45,6 +48,10 @@ class RsProjectConfigurable(
             Allow plugin to process declarative macro invocations
             to extract information for name resolution and type inference.
         """)
+        row("Show test tool window:", showTestToolWindowCheckbox, """
+            The Test Runner tab opens in the Run tool window when a
+            testing session begins, and features the same toolbar buttons.
```
Undin: I know that this phrase is the same as in [help](https://www.jetbrains.com/help/idea/test-runner-tab.html) for run tool window
But the original documentation is about `Test Runner` while this tooltip is only about turning on/off test tool window. Also, it's quite difficult to understand it without the corresponding highlighting and links. Maybe the following is better
"Show test results in run tool window when testing session begins instead of raw console"?

PR #3075, opened 2018-11-22
Bundled pretty-printer for LLDB: OsString

ortem:
Add `OsString` support.

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -9,8 +9,8 @@
     ]]></description>
 
     <change-notes><![CDATA[
-        <a href="https://intellij-rust.github.io/2018/12/25/changelog-89.html">
-                 https://intellij-rust.github.io/2018/12/25/changelog-89.html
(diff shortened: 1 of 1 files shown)


prettyPrinters/providers.py
```diff
@@ -118,19 +118,28 @@ def SizeSummaryProvider(valobj, dict):
...
 
+def StdOsStringSummaryProvider(valobj, dict):
+    # type: (SBValue, dict) -> str
+    logger = Logger.Logger()
+    logger >> "[StdOsStringSummaryProvider] for " + str(valobj.GetName())
+    buf = valobj.GetChildAtIndex(0).GetChildAtIndex(0)
+    is_windows = "Wtf8Buf" in buf.type.name
+    vec = buf.GetChildAtIndex(0) if is_windows else buf
+    length = vec.GetNumChildren()
+    chars = [chr(vec.GetChildAtIndex(i).GetValueAsUnsigned()) for i in range(length)]
```
mchernyavsky: We need to take into account py version, as is the case with `STD_STRING`.

PR #3086, opened 2018-11-27
Bundled pretty-printer for LLDB: fix Ref provider

ortem:
Fix `StdRefSyntheticProvider`: `value` was computed incorrectly for `Ref` and `RefMut`.

The problem was that `valobj.GetChildAtIndex(0)` doesn't always correspond to `valobj.Dereference()`. For example, it was working for `Ref<i32>`, but not for `Ref<&str>`.

### DIFF
--- a/prettyPrinters/providers.py
+++ b/prettyPrinters/providers.py
@@ -460,10 +460,16 @@ class StdRefSyntheticProvider:
     def __init__(self, valobj, dict, is_cell=False):
         # type: (SBValue, dict, bool) -> StdRefSyntheticProvider
         self.valobj = valobj
-        self.borrow = valobj.GetChildMemberWithName("borrow").GetChildAtIndex(0).GetChildAtIndex(0)
-        if not is_cell:
-            self.borrow = self.borrow.GetChildAtIndex(0)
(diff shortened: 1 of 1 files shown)


prettyPrinters/providers.py
```diff
@@ -457,10 +457,16 @@ class StdRefSyntheticProvider:
...
+
+        if is_cell:
+            self.borrow = valobj.GetChildMemberWithName("borrow").GetChildMemberWithName(
+                "value").GetChildMemberWithName("value")
+            self.value = valobj.GetChildMemberWithName("value").GetChildMemberWithName("value")
+        else:
+            self.borrow = valobj.GetChildMemberWithName("borrow").GetChildMemberWithName(
+                "borrow").GetChildMemberWithName("value").GetChildMemberWithName("value")
+            self.value = valobj.GetChildMemberWithName("value").GetChildWithName("value").GetChildWithName(
+                "value").Dereference()
```
Undin: Looks like it doesn't work, at least for me
screen shot 2018-12-01 at 02 45 15

ortem: Oh, I'm sorry, I've forgotten to push the last changes. Fixed.

screenshot 2018-12-01 at 20 39 51

Issue #2524, opened 2018-05-08
Copyright plugin does not work for rust files 


marhkb:
I dont't know if this is an issue of intellij-rust or the copyright plugin.
I have configured `File | Settings | Editor | Copyright` to insert a copyright into my rust files.
But this is impossible, because the option to insert a copyright is not displayed at all for rust files.

Undin:
Note, Copyright plugin is not available in CLion (see the corresponding [issue](https://youtrack.jetbrains.com/issue/CPP-10423))

PR #3089, opened 2018-11-28
Copyright plugin support

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/2524.

Also provides copyright formatting settings for Rust files.

Note, Copyright plugin is not available in CLion (see the corresponding [issue](https://youtrack.jetbrains.com/issue/CPP-10423))

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -132,7 +132,7 @@ project(":") {
     intellij {
         pluginName = "intellij-rust"
 //        alternativeIdePath = "deps/clion-$clionVersion"
-        setPlugins(project(":intellij-toml"), "IntelliLang")
+        setPlugins(project(":intellij-toml"), "IntelliLang", "copyright")
     }
 
     repositories {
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -21,6 +21,7 @@
     <depends optional="true" config-file="idea-only.xml">com.intellij.modules.java</depends>
     <depends optional="true" config-file="clion-only.xml">com.intellij.modules.clion</depends>
     <depends optional="true" config-file="toml-only.xml">org.toml.lang</depends>
+    <depends optional="true" config-file="plugin-copyright.xml">com.intellij.copyright</depends>
```
Undin: Let's use the same style for all optional dependency files and name it "copyright-only.xml"

mchernyavsky: Done.

Undin:
@mchernyavsky Can we fold copyright comment by default and move caret to line after it? See kotlin file creation in our project as example
I don't insist that it should be done in this PR especially if it's hard or/and unclear

build.gradle.kts
```diff
@@ -132,7 +132,7 @@ project(":") {
     intellij {
         pluginName = "intellij-rust"
 //        alternativeIdePath = "deps/clion-$clionVersion"
-        setPlugins(project(":intellij-toml"), "IntelliLang")
+        setPlugins(project(":intellij-toml"), "IntelliLang", "Copyright")
```
Undin: It should be `copyright`. It works on mac because HFS+ is case insensitive

PR #3094, opened 2018-11-30
TY: Fix isMovesByDefault on copyable type parameter

### LINKED ISSUE #3093: False positive on "Cannot move" inspection with Copy trait bound
## Environment

* Intellij-Rust plugin version: v0.2.0.2111-183
* Rust toolchain version: rustup 1.15.0 (f0a3d9a84 2018-11-27)
* cargo 1.32.0-nightly (b3d0b2e54 2018-11-15)
* IDE name and version: CLion 2018.3 Build #CL-183.4284.156
* Operating system: macOS 10.14.1 (18B75)

## Problem description

I get an incorrect inspection on the following code, saying "Cannot move" (borrow checking).

rust-fn-1

rust-fn-2

## Steps to reproduce

The function from above:

```rust
/// Returns a tuple `(index, v)` such that `v` is the maximum.
pub fn argmax<V>(vs: &[V]) -> (usize, V) where V: Copy + PartialOrd {
    assert_ne!(vs.len(), 0);
    let mut max_i = 0;
    let mut max_v = vs[0];
    for (i, &v) in vs.iter().enumerate().skip(1) {
        if max_v < v {
            max_i = i;
            max_v = v;
        }
    }
    (max_i, max_v)
}
```

Cargo.toml:

```toml
[package]
name = "foo"
version = "0.1.0"
authors = ["First Last <person@place.org>"]
edition = "2018"

[dependencies]
```

ortem:
Fixes #3093

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
@@ -160,8 +160,8 @@ fun Ty.isMovesByDefault(lookup: ImplLookup): Boolean =
     when (this) {
         is TyUnknown, is TyReference, is TyPointer, is TyFunction -> false
         is TyTuple -> types.any { it.isMovesByDefault(lookup) }
-        is TyTypeParameter -> parameter != TyTypeParameter.Self
-        else -> lookup.isCopy(this).not()
+        is TyTypeParameter -> !(parameter == TyTypeParameter.Self || lookup.isCopy(this))
+        else -> !lookup.isCopy(this)
     }
 
 val Ty.isBox: Boolean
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/types/ty/Ty.kt
```diff
@@ -160,7 +160,7 @@ fun Ty.isMovesByDefault(lookup: ImplLookup): Boolean =
     when (this) {
         is TyUnknown, is TyReference, is TyPointer, is TyFunction -> false
         is TyTuple -> types.any { it.isMovesByDefault(lookup) }
-        is TyTypeParameter -> parameter != TyTypeParameter.Self
+        is TyTypeParameter -> parameter != TyTypeParameter.Self && lookup.isCopy(this).not()
```
Undin: `not()` method for boolean value looks quite weird

ortem: I find `.not()` more visible than `!`. But if it looks weird, I'll replace it.

xpe: An outsider's take here -- in most languages I've used, the `!` prefix is idiomatic. People read it as "not <expression>" -- it works out fine.

Undin:
bors retry

PR #3096, opened 2018-11-30
QDOC: Highlight code in quick documentation

### LINKED ISSUE #2501: Add syntax highlighting to code snippets on tooltips
image

Undin:
Implement code highlighting in rust quick documentation.
In fact, it uses rust lexer to wrap some code tokens into `span` tag with the same styles as `RsHighlighter` provides. So, if you change color settings, it will also change code highlighting in quick documentation window.

image

Note, it's available since 2018.3 because the corresponding platform code appeared only in 2018.3

Closes #2501

### DIFF
--- a/src/183/main/kotlin/org/rust/lang/doc/RsCodeFenceProvider.kt
+++ b/src/183/main/kotlin/org/rust/lang/doc/RsCodeFenceProvider.kt
@@ -1,0 +1,69 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.doc
+
+import com.intellij.codeEditor.printing.HTMLTextPainter
+import com.intellij.psi.PsiElement
+import org.intellij.markdown.MarkdownTokenTypes
+import org.intellij.markdown.ast.ASTNode
+import org.intellij.markdown.ast.getTextInNode
+import org.intellij.markdown.html.GeneratingProvider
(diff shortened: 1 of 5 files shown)


src/183/main/kotlin/org/rust/lang/doc/RsCodeFenceProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        for (child in childrenToConsider) {
+            if (isContentStarted && child.type in listOf(MarkdownTokenTypes.CODE_FENCE_CONTENT, MarkdownTokenTypes.EOL)) {
+                if (skipNextEOL && child.type == MarkdownTokenTypes.EOL) {
+                    skipNextEOL = false
+                    continue
+                }
+                val codeLine = HtmlGenerator.trimIndents(child.getTextInNode(text), indentBefore)
+                // `# ` prefix is used to mark lines that should be skipped in rendered documentation.
+                // See https://doc.rust-lang.org/book/first-edition/documentation.html#documentation-as-tests
```
vlad20012: Looks like this link points to some wrong place. Also we can link this https://github.com/rust-lang/rust/blob/5182cc1ca65d05f16ee5e1529707ac6f63233ca9/src/librustdoc/html/markdown.rs#L114

Undin: They dropped first-edition book with 1.31 release and don't add the corresponding section into new doc so I've added only link to rustc code

PR #3099, opened 2018-12-01
REF: Auto-import unresolved types during implement members [WIP]

garyttierney:
Extracts out the code used for the Auto-Import fix to make it reusable by other intentions. 

I've added a visitor that finds all type references in the selected members to implement, and finds import candidates for each.  I'm not sure if this is the best approach, but it seems to work insofar as my testing.  I'll make sure the tests are passing and add a few more that verifies this functionality works.  

Is there anything else that needs covered in https://github.com/intellij-rust/intellij-rust/commit/f26633f0b3fb9388ad177b82d5dffef6344ee7f9#diff-5347552d7d8c8a383193d5925797806fR86 or a better way to do this?

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
@@ -11,10 +11,16 @@ import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
+import org.rust.ide.inspections.import.ImportCandidate
 import org.rust.lang.core.macros.expandedFromRecursively
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.BoundElement
+import org.rust.lang.core.types.canBeImported
+import org.rust.lang.core.types.importItem
+import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.ty.TyTraitObject
+import org.rust.lang.core.types.type
 import org.rust.openapiext.checkReadAccessAllowed
 import org.rust.openapiext.checkWriteAccessAllowed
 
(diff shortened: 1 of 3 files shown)


garyttierney:
Looks like the behaviour for Kotlin and Java is to prompt for which types the user wants to import. I'll update this tree to add that.

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
```diff
@@ -49,6 +55,59 @@ private fun insertNewTraitMembers(selected: Collection<RsAbstractable>, existing
...
+            val candidate = QualifiedNamedItem.ExplicitItem(item)
+                .withModuleReexports(mod.project)
+                .mapNotNull { ImportCandidate(it, it.canBeImported(superMods) ?: return@mapNotNull null) }
+                .firstOrNull() ?: return
+
+            importCandidates.add(candidate)
+        }
+    }
+
+    // Visit the existing trait functions to determine the types that need to be imported.
```
mchernyavsky: Another example:

```rust
// lib.rs
use std::collections::BTreeSet;

pub struct S;

pub trait T {
    fn foo(s: BTreeSet<i32>);
}

// main.rs before using `Implement Members`
use sandbox::T;

struct Q;
impl/*caret*/ T for Q {
}

// main.rs after using `Implement Members`
use sandbox::T;
use alloc::collections::btree_set::set::BTreeSet;

struct Q;
impl T for Q {
    fn foo(s: BTreeSet<i32> /*<- still unresolved*/) {
        unimplemented!()
    }
}
```

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
```diff
@@ -49,6 +55,59 @@ private fun insertNewTraitMembers(selected: Collection<RsAbstractable>, existing
     if (selected.isEmpty()) return
 
     val templateImpl = RsPsiFactory(existingMembers.project).createMembers(selected, trait.subst)
+    val mod = existingMembers.containingMod
+    val superMods = LinkedHashSet(mod.superMods)
+
+    val importCandidates = mutableListOf<ImportCandidate>()
+    val importCandidateVisitor = object : RsVisitor() {
+        override fun visitFunction(o: RsFunction) {
```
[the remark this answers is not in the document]
garyttierney: Ah, that's much better!

garyttierney:
> Hello! I am working on another [PR](https://github.com/intellij-rust/intellij-rust/pull/3145) and I reuse part of your PR that extracts out the code used for the auto-import quickfix (see [2869e95](https://github.com/intellij-rust/intellij-rust/commit/2869e956c80c976e024bb980d1d4135b0a1355c0), I signed you as an author). I didn't change anything except the location of the code - I moved most of the code from `Extensions.kt` back to `AutoImportFix.kt`, because It seemed to me that `Extensions.kt` is too general for such things.

Agreed on the move from `Extensions.kt` to `AutoImportFix.kt`, that seems like a more appropriate place for the code to live. I mainly moved it for some logical separation while I was figuring out what parts were reusable. Thanks for updating it!

I have some other local changes to add the "import chooser" and will have some more to address the comments you've left, so I'll apply my work on top of your more up-to-date tree and keep an eye out for it being merged back to master.

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
garyttierney: Nice catch and good reminder for me to update/add some more test-cases to https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/refactoring/implementMembers/ImplementMembersHandlerTest.kt

garyttierney:
I've had some trouble getting the import chooser working here. I'm calling it like this with a non-empty list of `filteredItems`:

```kotlin
    DataManager.getInstance().dataContextFromFocusAsync.onSuccess { ctx ->
        chooseItemAndImport(mod.project, ctx, filteredItems, mod)
    }
```

I figure this might be a platform issue related to where I'm attempting to get the data context (inside a write action where a chooser has already been opened for trait members), but thinking a little more about this I'm wondering if we want the import chooser to pop-up. If yes, does anyone have any ideas on resolving my issue with the pop-up not appearing?

Undin:
@garyttierney Could you please push your latest changes? It should help to understand what is the problem

mchernyavsky:
@garyttierney Thanks for your work! I'll finish this PR in https://github.com/intellij-rust/intellij-rust/pull/3712.

Issue #2853, opened 2018-09-15
<struct literal field> expected, got '0' in valid Rust code


kryptan:
## Environment

* **Intellij-Rust plugin version:**  0.2.0.2105-181

## Problem description

In the following code IntelliJ highlights non-existing error:

```Rust
struct A(i32);

fn a() {
    A {
        0: 0
    };
}
```

![image](https://user-images.githubusercontent.com/11528855/45587492-02f41a00-b90f-11e8-9079-3fbf242a5b42.png)

PR #3103, opened 2018-12-02
GRAM&RES: Allow use positional fields in struct literals

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/2853.


src/main/grammars/RustParser.bnf
```diff
@@ -999,11 +999,7 @@ StructLiteral ::= <<checkStructAllowed>> OuterAttr* PathGenericArgsWithColonsNoT
 
 StructLiteralBody ::= '{' StructLiteralField_with_recover* ('..'  AnyExpr)? '}' { pin = 1 }
 
-StructLiteralField ::= OuterAttr* identifier [ ':' AnyExpr ] {
-  pin = 2
-  implements = [ "org.rust.lang.core.psi.ext.RsReferenceElement" ]
-  mixin = "org.rust.lang.core.psi.ext.RsStructLiteralFieldImplMixin"
-}
+StructLiteralField ::= OuterAttr* ((FieldName | INTEGER_LITERAL) ':' AnyExpr | PathExpr)
```
vlad20012: I guess we can implement it without making PSI more complicated. I.e. write here `identifier | INTEGER_LITERAL` and make `INTEGER_LITERAL` `referenceNameElement` as well.

PR #3109, opened 2018-12-03
COMP: fix lifetime completion

Dimonchik0036:


src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -99,7 +99,7 @@ private fun getFieldsOwnerTailText(owner: RsFieldsOwner): String = when {
 
 private fun getInsertHandler(element: RsElement, scopeName: String, context: InsertionContext) {
     val curUseItem = context.getElementOfType<RsUseItem>()
-    if (element is RsNameIdentifierOwner && !isValidRustVariableIdentifier(scopeName) && scopeName !in CAN_NOT_BE_ESCAPED) {
+    if (element !is RsLifetimeParameter && element is RsNameIdentifierOwner && !isValidRustVariableIdentifier(scopeName) && scopeName !in CAN_NOT_BE_ESCAPED) {
```
Undin: Maybe it's better to create new method instead of `isValidRustVariableIdentifier` to take into account quote identifiers too? Like `isValidRustIdentifier`
Looks like the best place is "static" method of `RsNamesValidator`. It already has necessary code in `RsNamesValidator#isIdentifier`. We can extract it into companion object function and use it here and in `RsNamesValidator#isIdentifier`

Dimonchik0036: Done

Undin:
Good catch 👍 
Thanks!
bors r+

Issue #2616, opened 2018-06-27
Usages not highlighted in "initialization shorthand"


dvtomas:
See the image. Usages are not highlighted in struct's "initialization shorthand" syntax. I often rely on the highlight visualization to quickly inspect where a variable is used, this piece is missing.

![missing highlight in initialization shorthand](https://user-images.githubusercontent.com/534351/41966755-455e42b0-7a00-11e8-92ab-525f18def434.png)

PR #3117, opened 2018-12-06
RES: Multi resolve for init shorthand in struct literal

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/2616.


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -115,10 +115,20 @@ fun processFieldExprResolveVariants(
...
+    field: RsStructLiteralField,
+    isCompletion: Boolean,
+    processor: RsResolveProcessor
+): Boolean {
     val resolved = field.parentStructLiteral.path.reference.deepResolve()
     val structOrEnumVariant = resolved as? RsFieldsOwner ?: return false
-    return processFieldDeclarations(structOrEnumVariant, processor)
+    if (processFieldDeclarations(structOrEnumVariant, processor)) return true
+    if (!isCompletion && field.expr == null) {
+        processLocalVariables(field) { patBinding ->
```
vlad20012: Not only local variables are allowed here:
```rust
fn foo() {}

struct S { foo: fn() }

fn main() {
    let a = S { foo };
}
```

vlad20012:
What about type inference and rename?

PR #3121, opened 2018-12-07
INT: add "Put parameters on separate lines" intention

### LINKED ISSUE #3113: Add "Put parameters on separate lines" intention
The intention should transform
```rust
fn foo(parameter_1: i32, parameter_2: i32, parameter_3: i32) -> i32 {
```
into
```rust
fn foo(
    parameter_1: i32,
    parameter_2: i32,
    parameter_3: i32,
) -> i32 {
```

[Here](https://github.com/JetBrains/kotlin/blob/ba6da7c40a6cc502508faf6e04fa105b96bc7777/idea/src/org/jetbrains/kotlin/idea/intentions/ChopParameterListIntention.kt) is the implementation of similar intention in Kotlin.

t-kameyama:
fixes #3113

### DIFF
--- a/src/test/kotlin/org/rust/ide/intentions/ChopParameterListIntentionTest.kt
+++ b/src/test/kotlin/org/rust/ide/intentions/ChopParameterListIntentionTest.kt
@@ -1,0 +1,91 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+class ChopParameterListIntentionTest : RsIntentionTestBase(ChopParameterListIntention()) {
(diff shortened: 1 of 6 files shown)


src/test/kotlin/org/rust/ide/intentions/ChopParameterListIntentionTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    """, """
+        fn foo(
+            p1: i32, /*
+                comment
+            */
+            p2: i32,
+            p3: i32
+        ) {}
+    """)
+}
```
mchernyavsky: Let's add a test for a single-line comment.

mchernyavsky:
Thanks a lot!
bors r+

PR #3133, opened 2018-12-09
Import macro with use item & new macro call syntax (rust 1.30)

### LINKED ISSUE #2986: Support Rust 1.30 features
[Rust 1.30 blogpost](https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html)

- [ ] import macros by `use` item without `#[macro_use]` (use_extern_macros)
- [ ] new syntax for macros `foo::bar!()`
- [x] paths start with crate name (2015 edition) #2987
- [x] paths start with `crate` keyword (2015 edition)
- [x] raw identifiers (2015 edition)
- [x] `vis` macro fragment specifier

vlad20012:
- Allows to import macros by `use` item without `#[macro_use]`
  (the most of changes done by the first commit are about performance, so this is supported for free)
- Support new syntax for macros `foo::bar!()`

Closes #2986


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -734,52 +765,59 @@ private fun reexportedMacros(item: RsExternCrateItem): List<RsMacro>? {
...
     return buildList {
         val root = useItem.useSpeck ?: return@buildList
 
-        fun forEachRootUseSpeck(useSpeck: RsUseSpeck, consumer: (RsUseSpeck, RsPath) -> Unit) {
-            val path = useSpeck.path
-            if (path != null) {
-                consumer(useSpeck, path)
+        for ((crateName, macroName) in collect2segmentPaths(root)) {
+            val crateRoot = exportingMacrosCrates[crateName] as? RsFile
+                ?: findDependencyCrateByName(useItem, crateName)
```
Undin: Do I understand correctly that `findDependencyCrateByName` is called when we don't have `extern crate` item?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -528,6 +564,10 @@ private class MacroResolver private constructor(
             return true
         }
 
+        (startElement.contextOrSelf<RsElement>()?.crateRoot as? RsFile)?.let { crateRoot ->
```
Undin: Let's extract `startElement.contextOrSelf<RsElement>()?.crateRoot as? RsFile` into local variable and check for null with simple `if`
I think it will be a little bit readable (at least for me)

Undin: This code is related to `org.rust.lang.core.resolve.RsMacroResolveTest#test macro_export macro is visible in the same crate without macro_use` test, isn't it? If yes, maybe it's worth to add special testmark for this case?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -248,6 +265,10 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
         else -> TYPES_N_VALUES
     }
 
+    if (isCompletion && parent is RsPathExpr && qualifier?.path == null) {
```
Undin: Let's add a comment with an explanation that without `!` we parse `crate::path_prefix` as `RsPathExpr`

src/main/kotlin/org/rust/lang/core/completion/RsMacroCompletionProvider.kt
```diff
@@ -9,34 +9,63 @@ import com.intellij.codeInsight.completion.CompletionParameters
...
+    ) {
+        val position = parameters.position
+        val rsElement = position.ancestorStrict<RsElement>() ?: return
+        val mod = rsElement.ancestorOrSelf<RsMod>()
+
+        val leftSiblings = position.leftSiblings
+            .filter { it !is PsiWhiteSpace && it !is PsiComment && it !is PsiErrorElement }
+            .take(3).toList()
+        val is2segmentPath = leftSiblings.getOrNull(0)?.elementType == COLONCOLON &&
+            leftSiblings.getOrNull(1)?.elementType == IDENTIFIER
```
Undin: Looks like `&&` is missing here

src/main/kotlin/org/rust/lang/core/completion/RsMacroCompletionProvider.kt
```diff
@@ -9,34 +9,63 @@ import com.intellij.codeInsight.completion.CompletionParameters
...
+            leftSiblings.getOrNull(2)?.elementType != COLONCOLON
```
Undin: Hm, why do we need this condition?
It seems `::std::thread_local! { }` is valid

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -734,52 +765,59 @@ private fun reexportedMacros(item: RsExternCrateItem): List<RsMacro>? {
...
+    val result = mutableListOf<Pair<String, String?>>()
+
+    fun go(speck: RsUseSpeck, parentPath: RsPath? = null) {
+        testAssert { parentPath?.path == null }
+        val path = speck.path
+        val qualifier = path?.path
+        val group = speck.useGroup
+        if (path != null && qualifier != null && (qualifier.path != null || parentPath != null)) return
+        val firstSegment = parentPath ?: qualifier ?: path
+        val lastSegment = path ?: qualifier ?: parentPath
```
Undin: This looks quite complicated
I feel it can be simplified with using `RsPath#qualifier` extension method which works with `UseSpeck` too

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -217,6 +214,24 @@ fun processExternCrateResolveVariants(element: RsElement, isCompletion: Boolean,
...
+private fun findDependencyCrateByName(context: RsElement, name: String): RsFile? {
+    var found: RsFile? = null
+    processExternCrateResolveVariants(context, false) {
+        if (it.name == name) {
+            if (found != null) {
+                found = null
+                true
+            } else {
+                found = it.element as? RsFile
+                false
```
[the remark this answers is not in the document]
vlad20012: if we find an appropriate crate, we store it to `found` var and continue to iterate. if we find a second crate with the same name, we store null and stop to iterate (multiresolve). It seems to be correct behavior, but maybe we should not care about multiresolve and just return the first found crate

src/test/kotlin/org/rust/lang/core/psi/RsAttributeTest.kt
```diff
@@ -208,7 +208,7 @@ class RsAttributeTest : RsTestBase() {
 
         #[outer]
         makro!();
-        //^
+           //^
```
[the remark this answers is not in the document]
Undin: okay

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
vlad20012: +

src/main/kotlin/org/rust/lang/core/completion/RsMacroCompletionProvider.kt
vlad20012: We want to forbid a::b::c. More complex condition is needed, I guess.

src/test/kotlin/org/rust/lang/core/resolve/RsPackageLibraryResolveTest.kt
```diff
@@ -217,6 +207,106 @@ class RsPackageLibraryResolveTest : RsResolveTestBase() {
         }
     """, NameResolutionTestmarks.missingMacroUse)
 
+    fun `test import macro by use item`() = stubOnlyResolve("""
```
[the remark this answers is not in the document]
vlad20012: But the name is better :)
Ok, I'll remove the previous test

src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
```diff
@@ -120,9 +122,7 @@ private fun partToHighlight(element: RsElement): TextRange? {
     }
 
     if (element is RsMacroCall) {
-        var range = element.referenceNameElement.textRange ?: return null
-        range = range.union(element.excl.textRange)
-        return range
+        return element.excl.textRange
```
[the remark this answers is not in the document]
vlad20012: I split highlighting to 2 annotations because:
1. There is a bug: if multiple tokens are highlighted with a single annotation, this highlighting isn't visible in language injections
2. It's simpler to implement

It there an example of similar highlighting in other languages?

Undin:
@vlad20012 Could you please fix failed tests after your rebase onto master?

vlad20012:
Done

PR #3141, opened 2018-12-11
Add support for spacing between assoc type bindings.

Jezza:
Added some basic spacing support for associated type bindings.
While useful, the bigger part is definitely that spacings can now be added.
Granted, having seen the spacing builder... I don't think I wanna do that all at once.

So, this is, hopefully, the start of the migration towards giving the user the option to change the spacing settings.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -40,6 +40,7 @@ HybridEidolon
 idubrov
 isamborskiy
 JakubAdamWieczorek
+Jezza
 johnthagen
 jonas-schievink
 Keats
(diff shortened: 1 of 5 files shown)


Undin:
Let's merge it in this state
I'll add necessary tests myself
Thanks!
And welcome to intellij-rust contributors!

Jezza:
Sorry about that; I was planning on coming back to it, but holidays.
Entirely my b.

I'll see about adding some tests and I'll submit a PR.

Undin:
@Jezza I've added the corresponding test in #3202
Also, I've introduced new [method](https://github.com/intellij-rust/intellij-rust/blob/09aeba8988dac7cd6e177fb132009d9663d28b14/src/test/kotlin/org/rust/ide/formatter/RsFormatterTestBase.kt#L34-L49) to check both formatting variants when some option (now only boolean option) enabled and disabled in a single test.
It should make writing of formatting tests simpler. 

We will be glad to check your next PRs!

Jezza:
That does make things a lot easier, thanks!

PR #3144, opened 2018-12-12
LI: inject Rust language to doctests

### LINKED ISSUE #618: Highlight code fences in doc comments
Highlight code fences with Rust source code in doc comments (lexer highlighting is enough), like @wilfred does in Emacs ([tweet](https://twitter.com/_wilfredh/status/767457636820234240)):

cqaoeblw8ai8rj0 png large

I'm not 100% sure, but separating token types for code fence contents ~~and then layering Rust highlighting lexer over them should do~~. Remember that code fences can have language specifier.

The other solution could be to use language injection (@matklad's idea) - actually might be better.

Related source files:
- Documentation comments [lexer](https://github.com/intellij-rust/intellij-rust/tree/master/src/main/kotlin/org/rust/lang/doc/lexer), [token types](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/java/org/rust/lang/doc/psi/RustDocElementTypes.java)
- [Highlighter](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/highlight/RustHighlighter.kt)

vlad20012:
We now thinking about code inside Rust doctests as if it is a valid Rust code, with all IDE intelligence like highlighting, completion, intentions, linting, etc.
image

Unresolved questions:
- postprocess reformatting is not work in injections. It means that after the most of intentions/quickfixes the code will not be formatted and so will looks ugly. Should we disable intentions/quickfixes inside injections because of this? (I think we can ignore it for now, and then fix something in the platform)
- some intentions/quickfixes can edit PSI outside of injection that leads to crashes. But I think it is smaller part of them, and we can disable specific actions if they will not work.
- Typing enter inside injections does not insert an indent at the next line. 

About testing: This is mostly manually tested. There are tests for reference resolution and auto-import inside injection. Injection ranges are not tested (looks too complex to test). RsEnterHandler is not tested too (fixture API ignores injections while typing).

Closes #618

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsDoctestAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsDoctestAnnotator.kt
@@ -1,0 +1,41 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator
+
+import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.editor.colors.EditorColors
+import com.intellij.psi.PsiElement
+import com.intellij.psi.impl.source.tree.injected.InjectionBackgroundSuppressor
+import org.rust.cargo.project.settings.rustSettings
+import org.rust.ide.injected.RsDoctestLanguageInjector
+import org.rust.ide.injected.findDoctestInjectableRanges
+import org.rust.lang.core.psi.RsDocCommentImpl
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.containingCargoTarget
+
+/**
+ * Adds missing background for injections from [RsDoctestLanguageInjector].
+ * Background is disabled by [InjectionBackgroundSuppressor] marker implemented for [RsDocCommentImpl].
+ *
+ * We have to do it this way because we want to highlight fully range inside ```backticks```
+ * but a real injections is shifted by 1 character and empty lines are skipped.
+ */
+class RsDoctestAnnotator : RsAnnotatorBase() {
+    override fun annotateInternal(element: PsiElement, holder: AnnotationHolder) {
+        if (element !is RsDocCommentImpl) return
+        if (!element.project.rustSettings.doctestInjectionEnabled) return
+        // only library targets can have doctests
+        if (element.ancestorStrict<RsElement>()?.containingCargoTarget?.isLib != true) return
+
+        val startOffset = element.startOffset
+        findDoctestInjectableRanges(element).flatten().forEach {
+            holder.createAnnotation(HighlightSeverity.INFORMATION, it.shiftRight(startOffset), null)
+                .textAttributes = EditorColors.INJECTED_LANGUAGE_FRAGMENT
+        }
+    }
+}
--- a/src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
+++ b/src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
@@ -1,0 +1,182 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.injected
+
+import com.intellij.injected.editor.VirtualFileWindow
+import com.intellij.lang.injection.MultiHostInjector
+import com.intellij.lang.injection.MultiHostRegistrar
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.PsiElement
+import com.intellij.psi.tree.IElementType
+import com.intellij.util.text.CharArrayUtil
+import org.rust.cargo.project.settings.rustSettings
+import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.cargo.util.AutoInjectedCrates
+import org.rust.lang.RsLanguage
+import org.rust.lang.core.psi.RS_DOC_COMMENTS
+import org.rust.lang.core.psi.RsDocCommentImpl
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.doc.psi.RsDocKind
+import org.rust.openapiext.toPsiFile
+import org.rust.stdext.nextOrNull
+import java.util.regex.Pattern
+
+// See https://github.com/rust-lang/rust/blob/5182cc1c/src/librustdoc/html/markdown.rs#L646
+private val LANG_SPLIT_REGEX = Pattern.compile("[^\\w-]+", Pattern.UNICODE_CHARACTER_CLASS)
+private val RUST_LANG_ALIASES = listOf(
+    "rust",
+    "allow_fail",
+    "should_panic",
+    "no_run",
+    "test_harness",
+//    "compile_fail", // don't highlight code that is expected to contain errors
+    "edition2018",
+    "edition2015"
+)
+
+class RsDoctestLanguageInjector : MultiHostInjector {
+    private data class CodeRange(val start: Int, val end: Int, val codeStart: Int) {
+        fun isCodeNotEmpty(): Boolean = codeStart + 1 < end
+
+        val indent: Int = codeStart - start
+
+        fun offsetIndent(indent: Int): CodeRange? =
+            if (start + indent < end) CodeRange(start + indent, end, codeStart) else null
+    }
+
+    override fun elementsToInjectIn(): List<Class<out PsiElement>> =
+        listOf(RsDocCommentImpl::class.java)
+
+    override fun getLanguagesToInject(registrar: MultiHostRegistrar, context: PsiElement) {
+        if (context !is RsDocCommentImpl) return
+        if (!context.isValidHost || context.elementType !in RS_DOC_COMMENTS) return
+        if (!context.project.rustSettings.doctestInjectionEnabled) return
+
+        val rsElement = context.ancestorStrict<RsElement>() ?: return
+        val cargoTarget = rsElement.containingCargoTarget ?: return
+        if (!cargoTarget.isLib) return // only library targets can have doctests
+        val crateName = cargoTarget.normName
+        val text = context.text
+
+        findDoctestInjectableRanges(text, context.elementType).map { ranges ->
+            ranges.map {
+                CodeRange(
+                    it.startOffset,
+                    it.endOffset,
+                    CharArrayUtil.shiftForward(text, it.startOffset, it.endOffset, " \t")
+                )
+            }
+        }.map { ranges ->
+            val commonIndent = ranges.filter { it.isCodeNotEmpty() }.map { it.indent }.min()
+            val indentedRanges = if (commonIndent != null) ranges.mapNotNull { it.offsetIndent(commonIndent) } else ranges
+
+            indentedRanges.map { (start, end, codeStart) ->
+                // `cargo doc` has special rules for code lines which start with `#`:
+                //   * `# ` prefix is used to mark lines that should be skipped in rendered documentation.
+                //   * `##` prefix is converted to `#`
+                // See https://github.com/rust-lang/rust/blob/5182cc1c/src/librustdoc/html/markdown.rs#L114
+                when {
+                    text.startsWith("##", codeStart) -> TextRange(codeStart + 1, end)
+                    text.startsWith("# ", codeStart) -> TextRange(codeStart + 2, end)
+                    else -> TextRange(start, end)
+                }
+            }
+        }.forEach { ranges ->
+            val inj = registrar.startInjecting(RsLanguage)
+
+            ranges.forEachIndexed { index, range ->
+                val isFirstIteration = index == 0
+                val isLastIteration = index == ranges.size - 1
+
+                val prefix = if (isFirstIteration) {
+                    buildString {
+                        // Yes, we want to skip the only "std" crate. Not core/alloc/etc, the "std" only
+                        val isStdCrate = crateName == AutoInjectedCrates.STD &&
+                            cargoTarget.pkg.origin == PackageOrigin.STDLIB
+                        if (!isStdCrate) {
+                            append("extern crate ")
+                            append(crateName)
+                            append("; ")
+                        }
+                        append("fn main() {")
+                    }
+                } else {
+                    null
+                }
+                val suffix = if (isLastIteration) "}" else null
+
+                inj.addPlace(prefix, suffix, context, range)
+            }
+
+            inj.doneInjecting()
+        }
+    }
+}
+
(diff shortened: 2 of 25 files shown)


src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,152 @@
...
+                    generateSequence(codeBlockStart) { text.indexOf("\n", it) + 1 }
+                        .takeWhile { it != 0 && it < codeBlockEnd }
+                        .zipWithNext()
+                        .iterator()
+                }
+
+                // ```rust, should_panic, edition2018
+                //     ^ this text
+                val lang = lines.nextOrNull()?.let { text.substring(it.first, it.second - 1) } ?: return@mapNotNull null
+                if (!lang.isEmpty()) {
```
ortem: `isNotEmpty`

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,152 @@
...
+                    val parts = lang.split(LANG_SPLIT_REGEX).filter { it.isNotBlank() }
+                    if (parts.any { it !in RUST_LANG_ALIASES }) return@mapNotNull null
+                }
+
+                // skip doc comment infix (`///`, `//!` or ` * `)
+                val ranges = lines.asSequence().mapNotNull { (lineStart, lineEnd) ->
+                    val index = text.indexOf(infix, lineStart)
+                    if (index != -1 && index < lineEnd) {
+                        val start = index + infix.length
+                        Triple(start, lineEnd, CharArrayUtil.shiftForward(text, start, lineEnd, " \t"))
```
ortem: Maybe introduce a new class instead of triples? Something like
```
class Range(val start: Int, val end: Int, val codeStart: Int) {
    fun isCodeNotEmpty(): Boolean = codeStart + 1 < end

    val codeLength: Int = end - codeStart

    fun withIndent(indent: Int): Range? =
        if (indent < end) Range(codeStart, start + indent, end) else null
}
```
And then use
```
val commonIndent = ranges.filter { it.isCodeNotEmpty() }.map { it.codeLength }.min()
val newRanges = if (commonIndent != null) ranges.mapNotNull { it.withIndent(commonIndent) } else ranges
```

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,152 @@
...
+                            append("fn main() {")
+                        }
+                    } else {
+                        null
+                    }
+                    val suffix = if (last) "}" else null
+                    return inj.addPlace(prefix, suffix, context, TextRange(start, end))
+                }
+                ranges.forEachIndexed { index, (start, end) ->
+                    val first = index == 0
```
ortem: Also, it might be useful to introduce `CodeRange` instead of `Pair` and put the logic inside the class.
IMO it is difficult to read large `map`, `forEach`, and so forth.

vlad20012: I've refactored this place a bit. Should be simpler now.

kumbayo:
Nice!
In a previous version of this change i have seen the doc/test code having a green background.
(just like other injected code)
In this version, the green background is gone which makes source and test/doc code a bit difficult to differentiate IMO.

Was this an intentional change?

Previous:
image


Current:
image

vlad20012:
@kumbayo
> Was this an intentional change?

Yep, it is done by implementing [this marker interface](https://github.com/intellij-rust/intellij-rust/pull/3144/files#diff-f095f06fbfa248bb95aaff47ad74bc20R22). But this is a debatable question. Background brings some problems. There are a lot of places with a LARGE number of doctests (e.g. most of the std). With background a code start to look striped... This can be very annoying. Furthermore, we can't highlight with background continuous text range inside code fence because it often contains blank lines where nothing to inject. So editor starts to looks more and more striped... Of course, such background is created for simpler distinguish injected code. But I am optimistic that an injection can be distinguished by three slashes at the start of line.

kumbayo:
@vlad20012 Thanks for the info.
I assume, just extending the injected region to the left to contain the space character is not an option for some reason? 
In my tests, Rust seems to accept code immediately after /// without a space.

I think that with the dark theme, the difference can be seen much better because of the green doc comment characters /// on the left side.
But in the light theme with the gray /// provides little visual distinction.
Especially for longer code blocks like `Vec::try_reserve`

Have you tried using the light theme (IntellJ) with this change?
I wonder if you feel the same with that one.

vlad20012:
@kumbayo 
> I assume, just extending the injected region to the left to contain the space character is not an option for some reason?

Yes, if we extend the region, IDE will think that your code (inside injections) is shifted by 1 character from the start of the line, and, for example, you tab key will insert only 3 spaces.

I just tried light theme and I perhaps I should agree with you :)
I'll try to think of something.

kumbayo:
@vlad20012 I think that looks quite nice. Thanks!

Undin:
@vlad20012 Could you rebase onto master?

src/main/kotlin/org/rust/lang/core/RsAstFactory.kt
```diff
@@ -0,0 +1,20 @@
...
+package org.rust.lang.core
+
+import com.intellij.core.CoreASTFactory
+import com.intellij.psi.impl.source.tree.LeafElement
+import com.intellij.psi.impl.source.tree.PsiCommentImpl
+import com.intellij.psi.tree.IElementType
+import org.rust.lang.core.psi.RS_DOC_COMMENTS
+import org.rust.lang.core.psi.RsDocCommentImpl
+
+class RsAstFactory : CoreASTFactory() {
```
Undin: Let's use a common naming convention for such classes and rename it to `RsASTFactory ` as all other inheritances of `ASTFactory` do

src/main/kotlin/org/rust/lang/core/psi/RsDocCommentImpl.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.intellij.psi.tree.IElementType
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.util.text.CharArrayUtil
+import org.rust.ide.annotator.RsDoctestAnnotator
+import org.rust.ide.injected.RsDoctestLanguageInjector
+import org.rust.ide.injected.RsSimpleMultiLineEscaper
+import org.rust.lang.doc.psi.RsDocKind
+
+/**
+ * Psi element for rust documentation comments. The class is needed to provide specific behavior for
```
Undin: Looks like we can simplify "The class is needed to provide" to "Provides"

src/main/kotlin/org/rust/lang/core/psi/RsDocCommentImpl.kt
```diff
@@ -0,0 +1,83 @@
...
+ * [InjectionBackgroundSuppressor] is used to disable builtin background highlighting for injected.
+ * We create such background manually by [RsDoctestAnnotator] (see class docs)
+ */
+class RsDocCommentImpl(type: IElementType, text: CharSequence) : PsiCoreCommentImpl(type, text),
+                                                                 PsiLanguageInjectionHost,
+                                                                 InjectionBackgroundSuppressor {
+    override fun isValidHost(): Boolean = true
+
+    /**
+     * This method is used to handle changes in PSI injected to the comment (see [RsDoctestLanguageInjector]).
```
Undin: "This method is used to handle" -> "Handles"

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -65,6 +68,7 @@ class RsFile(
 
     override val isCrateRoot: Boolean
         get() {
+            if (originalFile.isDoctestInjection) return true
```
Undin: Let's add a comment why we need this - looks like it can be not obvious in future reading

src/main/kotlin/org/rust/ide/annotator/RsDoctestAnnotator.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.editor.colors.EditorColors
+import com.intellij.psi.PsiElement
+import com.intellij.psi.impl.source.tree.injected.InjectionBackgroundSuppressor
+import org.rust.ide.injected.RsDoctestLanguageInjector
+import org.rust.ide.injected.findDoctestInjectableRanges
+import org.rust.lang.core.psi.RsDocCommentImpl
+
+/**
+ * Add missing background for injections from [RsDoctestLanguageInjector].
```
Undin: "Add" -> "Adds"

Undin:
I think we should add an option to turn off this feature. If can be useful if a user wants to get previous behavior (no matter because of personal reasons or our possible bugs)

amakeev:
>I think we should add an option to turn off this feature. If can be useful if a user wants to get previous behavior (no matter because of personal reasons or our possible bugs)

just in case: not sure it's the right place, but there is dedicated `Editor | Language Injections` settings page

Undin:
> just in case: not sure it's the right place, but there is dedicated `Editor | Language Injections` settings page

Looks like it's a perfect place for this option. Thanks for the suggestion!

vlad20012:
> just in case: not sure it's the right place, but there is dedicated `Editor | Language Injections` settings page

Unfortunately it's a page of IntellijLang plugin, so even if we can use it, it will not work in all supported IDEs

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -490,7 +491,15 @@ fun RsMod.importItem(candidate: ImportCandidate) {
         0 -> "self::"
         else -> "super::".repeat(relativeDepth)
     }
-    insertUseItem(psiFactory, "$prefix${info.usePath}")
+
+    val insertionScope = if (context.isDoctestInjection) {
```
Undin: Could you add a comment about different cases and add short examples (with doc test with and without module)?

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -447,23 +448,23 @@ private fun RsPath.namespaceFilter(isCompletion: Boolean): (RsQualifiedNamedElem
 }
 
 /**
- * Inserts an use declaration to the receiver mod for importing the selected `candidate`.
+ * Inserts a use declaration to the mod where [context] located for importing the selected candidate ([this]).
  * This action requires write access.
  */
-fun RsMod.importItem(candidate: ImportCandidate) {
+fun ImportCandidate.importItem(context: RsElement) {
```
Undin: It seems that now `import` is more suitable name for this method

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -504,13 +513,13 @@ private fun RsMod.insertExternCrateItem(psiFactory: RsPsiFactory, crateName: Str
...
 
-private fun RsMod.insertUseItem(psiFactory: RsPsiFactory, usePath: String) {
+private fun RsItemsOwner.insertUseItem(psiFactory: RsPsiFactory, usePath: String) {
     val useItem = psiFactory.createUseItem(usePath)
     if (tryGroupWithOtherUseItems(psiFactory, useItem)) return
     val anchor = childrenOfType<RsUseItem>().lastElement ?: childrenOfType<RsExternCrateItem>().lastElement
     if (anchor != null) {
         val insertedUseItem = addAfter(useItem, anchor)
-        if (anchor is RsExternCrateItem) {
+        if (anchor is RsExternCrateItem || isDoctestInjection) {
```
Undin: Add a comment that we have to insert newline manually because autoformatting doesn't work in injected code

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
Undin: Or you can create testmark for each case and "link" the corresponding tests with this logic
Looks like it allows you not to write any examples because you've already added them in tests

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,176 @@
...
+        val text = context.text
+
+        findDoctestInjectableRanges(text, context.elementType).map { ranges1 ->
+            val ranges = ranges1.map {
+                CodeRange(
+                    it.startOffset,
+                    it.endOffset,
+                    CharArrayUtil.shiftForward(text, it.startOffset, it.endOffset, " \t")
+                )
+            }
```
Undin: Let's wrap `range1` to `range` transformation into separate map

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,176 @@
...
+                    null
+                }
+                val suffix = if (isLast) "}" else null
+
+                inj.addPlace(prefix, suffix, context, range)
+            }
+
+            inj.doneInjecting()
+        }
+    }
```
Undin: Maybe it worth to extract lambda from `forEach` and `map` into separate functions?
Looks like it can simplify reading

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,176 @@
...
+        }.forEach { ranges ->
+            val inj = registrar.startInjecting(RsLanguage)
+
+            ranges.forEachIndexed { index, range ->
+                val isFirst = index == 0
+                val isLast = index == ranges.size - 1
+
+                val prefix = if (isFirst) {
+                    buildString {
+                        if (crateName != "std" || cargoTarget.pkg.origin != PackageOrigin.STDLIB) {
```
Undin: `"std"` -> `AutoInjectedCrates.STD`

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -0,0 +1,176 @@
...
+            inj.doneInjecting()
+        }
+    }
+}
+
+fun findDoctestInjectableRanges(comment: RsDocCommentImpl): Sequence<List<TextRange>> =
+    findDoctestInjectableRanges(comment.text, comment.elementType)
+
+private fun findDoctestInjectableRanges(text: String, elementType: IElementType): Sequence<List<TextRange>> {
+    val tripleBacktickIndices = text.indicesOf("```").toList()
```
Undin: Let's add TODO to use markdown parser here

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
vlad20012: Too lot of local variables are used there (

vlad20012:
bors retry

Issue #789, opened 2016-11-23
Autocompletion, Auto Import for unqualified types


the8472:
A feature I'm missing from java development is the ability to just type `"Hash" Ctrl+Space` and have it offer `std::collections::Hash{Map, Set}` as autocompletions and automatically add a use declaration too, assuming it does not conflict or add the fully qualified type otherwise. In the latter case maybe an intention to create a type alias would help too.

vlad20012:
Moved to #1737

PR #3145, opened 2018-12-12
[WIP] RES: Add completion variants from index

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/789.

TODO:
- [ ] Get rid of code duplication
- [ ] Do not insert use item for items that already imported
- [ ] Add tests
- [ ] What about perfomance?


Undin:
@mchernyavsky I haven't looked at this yet but if you suppose that new completion can slow down full completion you can enable it only when it's "extended completion" (see `CompletionParameters#isExtendedCompletion`), i.e. user should press `Ctrl + Space` twice to enter into "extended completion" mode.
I understand it's less discoverable and less convenient for users but it allows us merge this feature without precise performance tests (and introduce new API for other features) and don't break anything. And if we find out that this completion is fast enough, we will drop extended mode condition

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -472,7 +352,159 @@ private val RsPath.namespaceFilter: (RsQualifiedNamedElement) -> Boolean get() =
...
+        if (parentMod == lca) return ImportInfo.LocalImportInfo(crateRelativePath)
+        if (!isPublic) return null
+        ourSuperMods.takeWhile { it != lca }.dropLast(1) to ImportInfo.LocalImportInfo(crateRelativePath)
+    }
+    return if (shouldBePublicMods.all { it.isPublic }) return importInfo else null
+}
+
+fun RsMod.importItem(candidate: ImportCandidate) = importItem(project, candidate, this)
+
+fun importItem(
```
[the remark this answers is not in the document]
mchernyavsky: It looks like we really don't need wrap it in two of three cases.

p.s. - I have converted `importItem` into extension for `RsMod`.

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -437,7 +385,7 @@ sealed class ImportInfo {
 
 data class ImportCandidate(val qualifiedNamedItem: QualifiedNamedItem, val info: ImportInfo)
 
-private val RsPath.namespaceFilter: (RsQualifiedNamedElement) -> Boolean get() = when (context) {
+val RsPath.namespaceFilter: (RsQualifiedNamedElement) -> Boolean get() = when (context) {
```
Undin: Can be private

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -52,10 +53,10 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
         }
 
         if (candidates.size == 1) {
-            importItem(project, candidates.first(), element)
+            element.containingMod.importItem(candidates.first())
```
Undin: Hm, here we assume that `AutoImportFix` starts in write action and it's true at this moment and that's why this call is not wrapped into write action
But it's wrong and we should avoid running the whole action in write action because it can do a lot of things in `findApplicableContext`
So let's add `override fun startInWriteAction(): Boolean = false` and wrap this call into write action

src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
```diff
@@ -156,3 +230,26 @@ private fun filterMethodCompletionVariantsByTraitBounds(
...
+
+private fun addImport(context: InsertionContext, candidate: ImportCandidate, mod: RsMod) {
+    val project = context.project
+    val file = context.file
+
+    val manager = PsiDocumentManager.getInstance(project)
+    val document = manager.getDocument(file)
+    if (document != null) {
+        manager.commitDocument(document)
+    }
```
Undin: Is it really needed here?
Looks like it's called in `com.intellij.codeInsight.completion.CodeCompletionHandlerBase#insertItem` where all documents are commited before insert handler invocation

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
mchernyavsky: Some tests will fail :(

Undin: Ok, I'll take a look after this PR

Undin:
Looks awesome!
bors r+

PR #3156, opened 2018-12-17
COMP: Provide basic completion via completion contributor

mchernyavsky:
This is needed for another [PR](https://github.com/intellij-rust/intellij-rust/pull/3145).
It should also improve the completion response time, because now completion variants will appear in batches.

TODO:
* Add completion-based performance tests

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
@@ -1,0 +1,158 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionProvider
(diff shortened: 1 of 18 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsBasicCompletionProvider.kt
```diff
@@ -0,0 +1,158 @@
...
+import org.rust.lang.core.psiElement
+import org.rust.lang.core.resolve.*
+import org.rust.lang.core.resolve.ref.MethodResolveVariant
+import org.rust.lang.core.types.infer.containsTyOfClass
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyTypeParameter
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.type
+
+object RsBasicCompletionProvider : CompletionProvider<CompletionParameters>() {
```
vlad20012: Maybe "RsCommonCompletionProvider"

vlad20012:
I think we can merge it as is and ring performance tests later

mchernyavsky:
Done.

Issue #3179, opened 2018-12-24
` await!(async { /* ... */ }); ` situation didn't get covered


imkow:
nightly didn't cover  ` await!(async { /* ... */ }); ` situation.

imkow:
it still prompt "`async` is reserved keyword in Edition 2018" error at ` await!(async { /* ... */ }); `

vlad20012:
Hmm, looks like the fix is still not in the nightly, bit it's a part of today's release
![image](https://user-images.githubusercontent.com/3221931/50440218-3a0e9a00-0906-11e9-970a-dd86490a46e7.png)

(release is made manually from the master branch, but nightly is time-based auto build, so release can rarely be more actual then nightly)

PR #3182, opened 2018-12-25
GRAM: Parse `await!` argument as expression

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/3179.


vlad20012:
Let's also add type inference test to ensure that types inside `await!()` are inferred

Issue #3183, opened 2018-12-25
Add fix to create lifetime from usage


sirgl:
Add fix to create lifetime parameter from usage.
Example:
```rust
struct S {
  r: &'a ref
}
```
result:
```rust
struct S<'a> {
  r: &'a ref
}
```
Now only error message is shown, that lifetime unresolved.

Related error (where fix should be registered):
`org.rust.lang.utils.RsDiagnostic.UndeclaredLifetimeError`

Also, if there are more than one place to add lifetime, chooser must be shown.
Chooser can be done using `IntroduceTargetChooser.showChooser`.

sirgl:
I'll do it myself. Even today I tried to use this non existing fix about 10 times.

PR #3159, opened 2018-12-19
ANN: improve annotator highlighting

Undin:
* correctly highlight module declarations, module and crate references
* provide API to register custom severity to simplify highlighting tests
* provide API to test only particular annotator
* use `info` severity to highlight using of unsafe code by unsafe annotator
* reorder annotator declarations in `plugin.xml` to launch faster annotators first

PR #3187, opened 2018-12-25
ACT: add quickfix to create lifetime parameter from usage

sirgl:
Related issue: #3183

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/CreateLifetimeParameterFromUsageFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/CreateLifetimeParameterFromUsageFix.kt
@@ -1,0 +1,70 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.command.WriteCommandAction
+import com.intellij.openapi.editor.Editor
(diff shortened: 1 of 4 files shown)


vlad20012:
Hah, your gif example shows incorrect behavior =)
Lifetimes inside inner structs are independent from parent declaration, so this example will not be compiled
```rust
fn foo<'a>() {
    struct S { f: &'a i32 }
}
```

vlad20012:
(and this is also a bug in our lifetime resolution ):

sirgl:
That's happens when you trust IDE too much :)

src/main/kotlin/org/rust/ide/annotator/fixes/CreateLifetimeParameterFromUsageFix.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    override fun getText(): String = familyName
+
+    override fun invoke(project: Project, file: PsiFile, editor: Editor?, startElement: PsiElement, endElement: PsiElement) {
+        val context = gatherContext(startElement) ?: return
+        insertLifetime(context.declaration, project, context.lifetime)
+    }
+
+    private fun insertLifetime(declaration: RsGenericDeclaration, project: Project, startElement: RsLifetime) {
```
vlad20012: Looks like `startElement` is a temporal name and should be renamed to `lifetime` or something

vlad20012:
Some tests are failed. Looks like it's related to recent changes in annotator tests #3159

Undin:
@sirgl Thanks!
I've fixed tests and merged this feature in #3385

PR #3190, opened 2018-12-26
add mut fix

### LINKED ISSUE #3184: No add mut fix suggested
* **Intellij-Rust plugin version:** v0.2.0.2112-183
* **Rust toolchain version:** 1.31
* **IDE name and version:** Clion
* **Operating system:** Mac

## Problem description

Plugin has fix to make variable mutable, but in this case it is not suggested.
```rust
struct A {}

impl A {
    fn foo(&mut self) {  }
}

struct V {
    a: A
}

impl V {
    fn foo(&self) {
        self.a.foo() // highlighted `Cannot borrow immutable local variable `self.a` as mutable`, but no fix suggested
    }
}
```

E. g. for given code fix is suggested:
```rust
struct A {}

impl A {
    fn foo(&mut self) {  }
    
    fn bar (&self) {
        self.foo(); // make 'self' mutable
    }
}
```

gfreezy:
fixed #3184

This test failed on my computer which is macos 10.14.1 with rust stable 1.31.0 .

```
org.rustSlowTests.CargoProjectResolveTest > test resolve feature gated crate FAILED
    java.lang.IllegalStateException: Failed to resolve the reference `libc` in `src/main.rs`.
        at org.rustSlowTests.CargoProjectResolveTest.test resolve feature gated crate(CargoProjectResolveTest.kt:323)
```

### DIFF
--- a/src/test/kotlin/org/rust/ide/annotator/fixes/AddMutableFixText.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/fixes/AddMutableFixText.kt
@@ -1,0 +1,68 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+
+import org.rust.ide.inspections.RsBorrowCheckerInspection
+import org.rust.ide.inspections.RsInspectionsTestBase
+
+class AddMutableFixText : RsInspectionsTestBase(RsBorrowCheckerInspection()) {
+    fun `test add mutable fix to self`() = checkFixByText("Make `self` mutable", """
+        struct A {}
+
+        impl A {
+            fn foo(&mut self) {  }
+
+            fn bar (&self) {
+                <error>self/*caret*/</error>.foo();
+            }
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/ide/annotator/fixes/AddMutableFixText.kt
```diff
@@ -0,0 +1,68 @@
...
+
+class AddMutableFixText : RsInspectionsTestBase(RsBorrowCheckerInspection()) {
+    fun `test add mutable fix to self`() = checkFixByText("Make `self` mutable", """
+        struct A {}
+
+        impl A {
+            fn foo(&mut self) {  }
+
+            fn bar (&self) {
+                <error>se/*caret*/lf</error>.foo();
```
vlad20012: Let's put /*caret* marker before or after `self`. It's hard to read an exploded identifier.

ortem:
Welcome to IntelliJ Rust contributors! Also, tests for `AddMutableFix`were really needed, thanks.
bors r+

PR #3200, opened 2018-12-28
Provide AddMutableFix inside RsAssignToImmutableInspection

ortem:
The main goal of this PR is to provide `AddMutableFix` inside `RsAssignToImmutableInspection` if it is possible.

mut_fix

To test new functionality, it was needed to use three inpections in `AddMutableFixText` at the same time, so I introduced `RsMultipleInspectionsTestBase`.


src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
```diff
@@ -47,4 +50,12 @@ class RsAssignToImmutableInspection : RsLocalInspectionTool() {
     private fun registerProblem(holder: ProblemsHolder, expr: RsExpr, message: String) {
         RsDiagnostic.CannotAssignToImmutable(expr, message).addToHolder(holder)
     }
+
+    private fun registerProblemWithFix(holder: ProblemsHolder, expr: RsExpr, nameExpr: RsExpr?, message: String) {
```
Undin: Can we merge `registerProblem` and `registerProblemWithFix` into one method where `nameExpr` is null by default?
And maybe `assigneeExpr` in better than `nameExpr`?

src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
```diff
@@ -31,20 +33,25 @@ class RsAssignToImmutableInspection : RsLocalInspectionTool() {
...
-            is TyReference -> registerProblem(holder, expr, "immutable borrowed content")
-            is TyPointer -> registerProblem(holder, expr, "immutable dereference of raw pointer")
+            is TyReference -> registerProblem(holder, expr, message = "immutable borrowed content")
+            is TyPointer -> registerProblem(holder, expr, message = "immutable dereference of raw pointer")
         }
     }
 
-    private fun registerProblem(holder: ProblemsHolder, expr: RsExpr, message: String) {
-        RsDiagnostic.CannotAssignToImmutable(expr, message).addToHolder(holder)
+    private fun registerProblem(holder: ProblemsHolder, expr: RsExpr, assigneeExpr: RsExpr? = null, message: String) {
```
Undin: Let's swap `assigneeExpr` and `message` params not to force using named parameter for `message`
If you think that it's important (I have light feel that it's prettier) to keep `expr` and `assigneeExpr` together, you can move `message` param before `expr` one

src/main/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspection.kt
```diff
@@ -31,20 +33,25 @@ class RsAssignToImmutableInspection : RsLocalInspectionTool() {
...
+        if (assigneeExpr == null || assigneeExpr.type is TyReference) {
+            RsDiagnostic.CannotAssignToImmutable(expr, message).addToHolder(holder)
```
Undin: You can return `null` fix here and move diagnostic creation below if/else expr
But it is up to you

PR #3204, opened 2018-12-29
INT: Add "Put parameters on one line" intention

mchernyavsky:
Based on https://github.com/intellij-rust/intellij-rust/pull/3121.

[Kotlin's implementation](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/intentions/JoinParameterListIntention.kt).

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RsJoinRawLinesHandler.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsJoinRawLinesHandler.kt
@@ -50,7 +50,7 @@ class RsJoinRawLinesHandler : JoinRawLinesHandlerDelegate {
                         if (grandpa is RsMatchArm && grandpa.lastChild?.elementType != COMMA) {
                             grandpa.add(psiFactory.createComma())
                         }
-                        newExpr.textRange.startOffset
+                        newExpr.startOffset
                     }
                 }
 
@@ -58,7 +58,7 @@ class RsJoinRawLinesHandler : JoinRawLinesHandlerDelegate {
 
             is RsIfExpr, is RsElseBranch -> {
                 val newBlock = psiFactory.createBlockExpr(expr.text).block
-                return block.replace(newBlock).textRange.startOffset
+                return block.replace(newBlock).startOffset
             }
         }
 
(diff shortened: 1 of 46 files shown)


src/main/kotlin/org/rust/ide/intentions/JoinParameterListIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsValueParameter
+import org.rust.lang.core.psi.RsValueParameterList
+import org.rust.lang.core.psi.ext.RsElement
+
+abstract class JoinListIntention<TList : RsElement, TElement : RsElement>(
+    listClass: Class<TList>,
+    elementClass: Class<TElement>,
+    intentionText: String
+) : ChopListIntention<TList, TElement>(listClass, elementClass, intentionText) {
```
Undin: I find this inheritance weird (I know that kotlin has similar class hierarchy)
I'd prefer to create some parent class for both ("join" and "chop") subhierarchy, move all base properties and methods to it and make own `invoke` implementation for each subhierarchy instead of overriding
I.e.
```
                SomeCoolNameForParentClassListIntention
                       /               \
           ChopListIntention       JoinListIntention
                     /                    \ 
    ChopParameterListIntention    JoinParameterListIntention

File: src/main/kotlin/org/rust/ide/intentions/JoinParameterListIntention.kt [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+            return null
+        }
+        return list
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: TList) {
+        val document = editor.document
+        val elements = ctx.elements
+
+        nextBreak(elements.last())?.let { document.deleteString(it.textRange.startOffset, it.textRange.endOffset) }
```
Undin: Kotlin has extension properties `PsiElement#startOffset` and `PsiElement#endOffset`
Maybe we want to have them too? Looks like they are rather useful

src/main/kotlin/org/rust/ide/intentions/AbstractListIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import com.intellij.psi.PsiComment
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.leftSiblings
+import org.rust.lang.core.psi.ext.rightSiblings
+
+abstract class AbstractListIntention<TList : RsElement, TElement : RsElement>(
```
Undin: We usually use `SomeSemanticNameBase` name convention for abstract classes

mchernyavsky: Fixed.

Undin:
bors retry

PR #3217, opened 2019-01-02
TYPE: show inlay type hints for variables in `if let` and `while let` patterns

### LINKED ISSUE #3215: Cannot infer type of variable by tuple-unpacking in a pattern match statement
## Environment

* **Intellij-Rust plugin version:** v0.2.0.2113-183
* **Rust toolchain version:** x86_64-pc-windows-msvc
* **IDE name and version:** Clion 2018.3.2
* **Operating system:** Win10 64bit

## Problem description
default
The types of `x` and `y` are not inferred.

## Steps to reproduce

``` rust
let result: Result<(i32, i32), ()> = Ok((1, 2));
if let Ok((x, y)) = result {
    println!("{} {}", x, y);
}
```

Undin:
screen shot 2019-01-02 at 23 10 05

Fixes #3215


Undin:
Hm, looks like we don't provide type hints in patterns inside `match` expression and I think we should do it because it's just another way to introduce new local variables in scope. I'm doing to support in in this PR too.
But I'm not sure about settings for them.
Currently, we have separate settings for `let` declaration bindings, function params, `for` loop and lambda params. 
From one hand, `let`, `if/while let` and bindings in `match` expression are about the same - local variables creation, so we can just implement type hints for each `match` arm and have single option for these cases. From the other hand, options for each case allow a user to tune type hints for his needs. Also, we already have a special option for binding in `for` loop although it creates a new local variable like all other patterns
@vlad20012 @ortem @mchernyavsky @matklad @kumbayo what do you think?

matklad:
I’d be more of a fan of fine-grained options. I think the best case is when most of the users just ignore options because defaults are great, but a few can tweak stuff to a fine detail. That is, using options should be a rare use-case tweaked primarily for power, not for ease of use.

Slightly off topic, but I highly recommend fish shell design doc: 
https://fishshell.com/docs/current/design.htm

It works really well for IDEs also!

kumbayo:
I like the idea of showing the type hints also in this additional cases!

I noticed a problem when using `Smart hints` then matching `None` or other enums without values in them, it sometimes shows unexpected type hints.
image

Code:
```rust
enum MyEnum {
    A,
    B,
    C(i32),
}

fn main() {
    use MyEnum::A;

    let my = MyEnum::A;
    match my {
        A => 1, // A is an enum variant so the type hint is not expected
        MyEnum::B => 2,
        MyEnum::C(c) => c,
        z => 4, // z is a binding so here the type hint is expected
    };

    match Some(5i32) {
        Some(n) => n, // There should be a type hint for n
        None => 2,
    };

    match &Some(5i32) {
        Some(n) => *n,
        None => 2,// None is an enum variant so the type hint is not expected
    };
}
```

If we get more fine grained options, maybe some short examples next to each settings (or in the tooltip) would be a good idea.
And also maybe group "type hints" vs "other hints" together and unify the name/descriptions to consistently say "type hint" either at the start or the end.

I also noticed that we do not show type hints for the shorthand version of destructuring.
Not sure if it belongs to this MR or to a separate one.
image

Undin:
> I noticed a problem when using Smart hints then matching None or other enums without values in them, it sometimes shows unexpected type hints.

@kumbayo looks like now type hints should work as expected

> I also noticed that we do not show type hints for the shorthand version of destructuring.
Not sure if it belongs to this MR or to a separate one.

We will fix it in separate PR

PR #3224, opened 2019-01-04
CARGO&RES: Support extern crate renaming via Cargo.toml

### LINKED ISSUE #3044: Cannot resolve symbols anymore if cargo-features=["rename-dependency"] is used
Now that rename-dependency has been stabilized and uplifted into beta, it would be nice if it was supported.

`Cargo.toml`:

```toml
cargo-features = ["rename-dependency"]

[dependencies]
foo = { package = "bar", version = "0.1" }
```

Somewhere in the codebase...

```rust
use foo::Foo;  // <-- can no longer resolve it, any uses of Foo are highlighted red
```

(Versions: default toolchain is nightly via rustup; latest nightly compiler build; latest IntelliJ Rust plugin version)

Undin:
Closes #3044


src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -83,6 +85,11 @@ interface CargoWorkspace {
         val edition: Edition
     }
 
+    interface Dependency {
+        val pkg: Package
+        val name: String?
```
matklad: This should be non-null and set in constructor from `pkg` if an epxlicit name is missing?

src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -170,9 +171,29 @@ object CargoMetadata {
...
+    data class Dep(
+        /**
+         * Id of dependent package
+         */
+        val pkg: PackageId,
+
+        /**
+         * Dependency name that should be used in code as extern crate name
+         */
+        val name: String?
```
matklad: I think this is guaranteed to be non-null

Undin: Hm, [original structure](https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_output_metadata.rs#L107-L110) in cargo also holds dependency name as `Option<String>`

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
Undin: Good idea!

src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
matklad: Oh, you are right... The reason for that is horrible: it's possible to depend on a carte which contains only binary target. So, while dependency is present, it is sort-of a no-op dependency.

Undin:
bors r=matklad

PR #3228, opened 2019-01-08
INSP: Denote infinite loops with 'loop { ... }'

dimonchik0036:
peek 2019-01-08 18-28

### DIFF
--- a/src/main/resources/inspectionDescriptions/RsWhileTrueLoop.html
+++ b/src/main/resources/inspectionDescriptions/RsWhileTrueLoop.html
@@ -1,0 +1,1 @@
+Detects <code>while</code> loops which can be replaced with <code>loop</code>.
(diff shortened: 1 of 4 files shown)


src/main/resources/inspectionDescriptions/RsWhileTrueLoop.html
```diff
@@ -0,0 +1 @@
+Finds usages of the <code>while true</code> expression which can be replaced with <code>loop</code>.
```
Undin: I think `Detects <code>while</code> loops which can be replaced with <code>loop</code>.` is a little bit better because:
* `Usage` is more about usage of some language item: function, struct, enum, etc so it can be confusing
* usually we don't use `while` loop as an expression
* in future we can make more precise condition analysis and handle more cases when `condition` is always true

Undin:
Excellent!
bors r+

Undin:
bors retry

PR #3232, opened 2019-01-09
Fix LLDB printers due to removed NonZero wrapper

ortem:
The `NonZero` wrapper [was removed](https://github.com/rust-lang/rust/commit/7a0911528058e87d22ea305695f4047572c5e067) from `Unique` and `NonNull`, so current LLDB pretty-printers (for `String`, `Vec`, `VecDeque`, `Rc`, `Arc`) don't work on `rustc 1.33.0-nightly`.

### DIFF
--- a/prettyPrinters/providers.py
+++ b/prettyPrinters/providers.py
@@ -55,6 +55,12 @@ class ValueBuilder:
         return self.valobj.CreateValueFromData(name, data, type)
 
 
+def unwrap_unique_or_non_null(unique_or_nonnull):
+    # BACKCOMPAT: rust 1.32 https://github.com/rust-lang/rust/commit/7a0911528058e87d22ea305695f4047572c5e067
+    ptr = unique_or_nonnull.GetChildMemberWithName("pointer")
+    return ptr if ptr.TypeIsPointerType() else ptr.GetChildAtIndex(0)
+
+
 class DefaultSynthteticProvider:
     def __init__(self, valobj, dict):
         # type: (SBValue, dict) -> DefaultSynthteticProvider
@@ -249,7 +255,8 @@ class StdVecSyntheticProvider:
 
     struct Vec<T> { buf: RawVec<T>, len: usize }
     struct RawVec<T> { ptr: Unique<T>, cap: usize, ... }
-    struct Unique<T: ?Sized> { pointer: NonZero<*const T>, ... }
+    rust 1.31.1: struct Unique<T: ?Sized> { pointer: NonZero<*const T>, ... }
+    rust 1.33.0: struct Unique<T: ?Sized> { pointer: *const T, ... }
     struct NonZero<T>(T)
     """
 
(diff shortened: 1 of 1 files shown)


prettyPrinters/providers.py
```diff
@@ -351,25 +363,28 @@ class StdRcSyntheticProvider:
...
     def __init__(self, valobj, dict, is_atomic=False):
         # type: (SBValue, dict, bool) -> StdRcSyntheticProvider
         self.valobj = valobj
 
-        self.ptr = self.valobj.GetChildMemberWithName("ptr").GetChildMemberWithName("pointer").GetChildAtIndex(0)
+        self.ptr = self.valobj.GetChildMemberWithName("ptr").GetChildMemberWithName("pointer")
+        # https://github.com/rust-lang/rust/commit/7a0911528058e87d22ea305695f4047572c5e067
+        # TODO: get rid of it after new rustc stable release
+        if not self.ptr.TypeIsPointerType():
+            self.ptr = self.ptr.GetChildAtIndex(0)
```
Undin: Can we extract this code into a separate method?
At this moment we have 3 copy of it

prettyPrinters/providers.py
```diff
@@ -249,8 +249,8 @@ class StdVecSyntheticProvider:
 
     struct Vec<T> { buf: RawVec<T>, len: usize }
     struct RawVec<T> { ptr: Unique<T>, cap: usize, ... }
-    struct Unique<T: ?Sized> { pointer: NonZero<*const T>, ... }
-    struct NonZero<T>(T)
```
Undin: Looks like it's still useful info about, so let's keep it

prettyPrinters/providers.py
```diff
@@ -283,7 +283,13 @@ def update(self):
         # type: () -> None
         self.length = self.valobj.GetChildMemberWithName("len").GetValueAsUnsigned()
         self.buf = self.valobj.GetChildMemberWithName("buf")
-        self.data_ptr = self.buf.GetChildAtIndex(0).GetChildAtIndex(0).GetChildAtIndex(0)
+
+        self.data_ptr = self.buf.GetChildMemberWithName("ptr").GetChildAtIndex(0)
+        # https://github.com/rust-lang/rust/commit/7a0911528058e87d22ea305695f4047572c5e067
+        # TODO: get rid of it after new rustc stable release
```
Undin: We can't drop it because we will break debugging with rustc below 1.33
Let's better add `# BACKCOMPAT: rust 1.32 some text` comment to remove this when we will drop rust 1.32 support

prettyPrinters/providers.py
```diff
@@ -55,6 +55,12 @@ def from_uint(self, name, value):
         return self.valobj.CreateValueFromData(name, data, type)
 
 
+def unwrap_Unique_or_NonNull(unique_or_nonnull):
```
Undin: Let's use python name convention and use lower case here

PR #3233, opened 2019-01-09
Add println postfix template

SomeoneToIgnore:
I've decided to add a postfix template for creating a `println` macro for the expression applicable.
I've kept the implementation as generic as possible so that similar macros as `eprintln`, `format` would be easy to add, but this can be done later in a separate PR, after the current one is considered to be normal.

There are two minor issues I've faced when writing the tests:
1. One is the commented out test with a  `fixme` commend in it - looks like type resolution does not work as intended in this case hence macro template is not applicable.
The case seems to be rather minor though and can be fixed separately.

2. I've failed to use `Parameterized` test, so I had to leave the tests as they are now. Does not seem to be a big issue also, since all the tests are still there, but I have no idea how to fix it.

Here's the snippet: https://gist.github.com/SomeoneToIgnore/65d2328c7a3d65ba9a5fc5b0f9e917a4
When I run it, I receive the following error:
`ERROR: Read access is allowed from event dispatch thread or inside read-action only (see com.intellij.openapi.application.Application.runReadAction())`

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
@@ -1,0 +1,142 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.template.postfix
+
+import com.intellij.codeInsight.template.postfix.templates.PostfixTemplateWithExpressionSelector
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import com.intellij.psi.codeStyle.CodeStyleManager
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.ImplLookup
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.TraitRef
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.ty.TyStr
+import org.rust.lang.core.types.ty.TyUnit
+import org.rust.lang.core.types.type
+
+class PrintlnPostfixTemplate(provider: RsPostfixTemplateProvider, private val macroName: String = "println") :
+    PostfixTemplateWithExpressionSelector(null, macroName, "$macroName!(\"{:?}\", expr);",
+        RsTopMostInScopeSelector { it.isNotIgnored && (it.implementsDebug || it.implementsDisplay) }, provider) {
+
+    private enum class Fmt(val text: String) {
+        None(""),
+        Display("{}"),
+        Debug("{:?}");
+
+        companion object {
+            fun fromExpr(expr: RsExpr): Fmt = when {
+                (expr.type as? TyReference)?.referenced is TyStr -> Fmt.None
+                expr.implementsDebug -> Fmt.Debug
+                else -> Fmt.Display
+            }
+        }
+    }
+
+    private class MacroCreator(private val editor: Editor, private val psiFactory: RsPsiFactory,
+                               macroName: String, private val fmt: Fmt) {
+        private val macroStart = "$macroName!("
+
+        fun createMacro(expressionText: String, addTrailingSemicolon: Boolean = true): RsExpr {
+            val macroExpression = if (fmt == Fmt.None) {
+                psiFactory.createExpression("$macroStart$expressionText)")
+            } else {
+                psiFactory.createExpression("$macroStart\"${fmt.text}\", $expressionText)")
+            }
+            if (addTrailingSemicolon) {
+                macroExpression.add(psiFactory.createSemicolon())
+            }
+            return macroExpression
+        }
+
+        fun replaceWithMacro(expression: PsiElement, addTrailingComma: Boolean = false) {
+            val newElement = expression.replace(createMacro(expression.text, expression.parent is RsBlock))
+            if (addTrailingComma) {
+                newElement.add(psiFactory.createComma())
+            }
+            editor.caretModel.moveToOffset(newElement.endOffset)
+        }
+    }
+
+    override fun expandForChooseExpression(expression: PsiElement, editor: Editor) {
+        if (expression !is RsExpr) return
+
+        val psiFactory = RsPsiFactory(expression.project)
+        val macroCreator = MacroCreator(editor, psiFactory, macroName, Fmt.fromExpr(expression))
+
+        val parent = expression.parent
+        when (parent) {
+            is RsLetDecl -> {
+                val expressionText = parent.pat?.text ?: return
+                addMacroNextToElement(macroCreator.createMacro(expressionText), parent, psiFactory, editor)
+                if (parent.semicolon == null) {
+                    parent.add(psiFactory.createSemicolon())
+                }
+            }
+            is RsMatchArm -> {
+                val matchBody = parent.ancestorStrict<RsMatchBody>() ?: return
+                if (matchBody.containsUnitArm) {
+                    macroCreator.replaceWithMacro(expression, parent.comma == null)
+                } else {
+                    surroundWithBlockExpression(macroCreator.createMacro(expression.text), expression, psiFactory, editor)
+                }
+            }
+            is RsBinaryExpr -> addMacroNextToElement(macroCreator.createMacro(expression.text), parent, psiFactory, editor)
+            else -> macroCreator.replaceWithMacro(expression)
+        }
+    }
+
+    private fun addMacroNextToElement(macro: RsExpr, element: PsiElement, psiFactory: RsPsiFactory, editor: Editor) {
+        val childToContainer = element.closestContainerAncestor ?: return
+        val container = childToContainer.second
+        if (container is RsMatchArm) {
+            surroundWithBlockExpression(macro, childToContainer.first, psiFactory, editor)
+        } else {
+            val newElement = container.addAfter(macro, childToContainer.first.getNextNonCommentSibling()?.prevSibling)
+            editor.caretModel.moveToOffset(newElement.endOffset)
+            newElement.add(psiFactory.createNewline())
+            CodeStyleManager.getInstance(newElement.project).reformat(container.ancestorStrict<RsBlock>() ?: container)
+        }
+    }
+
+    private fun surroundWithBlockExpression(macro: RsExpr, originalExpression: PsiElement, psiFactory: RsPsiFactory, editor: Editor) {
+        val newBlock = psiFactory.createBlockExpr("${macro.text}\n${originalExpression.text}")
+        val newElement = originalExpression.replace(newBlock) as RsBlockExpr
+        editor.caretModel.moveToOffset(newElement.block.children[0].endOffset)
+    }
+}
+
+private val PsiElement.closestContainerAncestor: Pair<PsiElement, PsiElement>?
+    get() = ancestorPairs.dropWhile { it.second !is RsBlock && it.second !is RsMatchArm }.firstOrNull()
+
+private val RsMatchBody.containsUnitArm: Boolean
+    get() = matchArmList.any { it.expr?.type is TyUnit }
+
+private val RsExpr.isNotIgnored: Boolean
+    get() {
+        val parent = parent ?: return false
+        return when (parent) {
+            is RsLetDecl -> parent.pat !is RsPatWild
+            is RsMatchArm, is RsStmt, is RsBlock, is RsBinaryExpr -> true
+            else -> false
+        }
+    }
+
+private val RsExpr.implementsDebug: Boolean
+    get() = isImplementsTrait(this, "std::fmt::Debug")
+
+private val RsExpr.implementsDisplay: Boolean
+    get() = isImplementsTrait(this, "std::fmt::Display")
+
+private fun isImplementsTrait(expr: RsExpr, traitName: String, vararg subst: Ty): Boolean {
+    val items = expr.knownItems
+    val implLookup = ImplLookup(expr.project, items)
+    val trait = items.findItem(traitName) as? RsTraitItem ?: return false
+    // Debug and Display are implemented for references: https://doc.rust-lang.org/src/core/fmt/mod.rs.html#1907-1924
+    return implLookup.canSelectWithDeref(TraitRef(expr.type, trait.withSubst(*subst)))
+}
--- a/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
@@ -27,7 +27,8 @@ class RsPostfixTemplateProvider : PostfixTemplateProvider {
         LambdaPostfixTemplate(this),
         NotPostfixTemplate(this),
         LetPostfixTemplate(this),
-        IterPostfixTemplate(this)
+        IterPostfixTemplate(this),
+        PrintlnPostfixTemplate(this)
     )
 
     override fun getTemplates(): Set<PostfixTemplate> = templates
--- a/src/main/resources/postfixTemplates/PrintlnPostfixTemplate/after.rs.template
+++ b/src/main/resources/postfixTemplates/PrintlnPostfixTemplate/after.rs.template
@@ -1,0 +1,3 @@
+pub fn main() {
+    println!("{:?}", 44);
+}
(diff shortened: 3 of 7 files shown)


ortem:
Thanks for this feature, it looks useful!
1. You're right, there was a bug in type inference related to copied files (the copy appears due to `PostfixLiveTemplate.isApplicableTemplate` call inside `org.rust.ide.template.postfix.RsPostfixTemplateTest#checkApplicability`). Now the bug seems to be [fixed](https://github.com/intellij-rust/intellij-rust/pull/3236), so just wait a bit and rebase onto master.
2. I don't have much experience with JUnit 4 and don't know why it doesn't work. Maybe you really should use `runReadAction` somewhere. But the current implementation looks OK too.

SomeoneToIgnore:
1. Great, thanks for the fix, have rebased onto the fresh master and uncommented the test.
2. The only non-invasive place I could think of is `doAction`, but wapping it with that method did not help, so it's something else. It's a bit weird, since relatively similar `CargoTomlCompletionEditionTest` test had neither name nor edition issue while using relatively same JUnit api.
But I'm also fine with the current state of the tests, so decided not to investigate further.

If everything's fine here, would it make sense create a follow-up with another PR containing similar macros?
`print`, `eprint`, `eprintln` and `format` behave the same way, so it's a matter of providing a different macro name to the class and tests. Not sure if all of them are needed though.

ortem:
> The only non-invasive place I could think of is doAction, but wrapping it with that method did not help, so it's something else

Did you try something like this? Seems like it helps.
```
    private fun checkApplicability(testCase: String, isApplicable: Boolean) {
        val provider = LanguagePostfixTemplate.LANG_EP.allForLanguage(RsLanguage)
            .first { descriptor ->
                descriptor.templates.any { template ->
                    template.javaClass == this.postfixTemplate.javaClass
                }
            }

        val result = runReadAction {
            PostfixLiveTemplate.isApplicableTemplate(
                provider,
                postfixTemplate.key,
                myFixture.file,
                myFixture.editor
            )
        }

        check(result == isApplicable) {
            "postfixTemplate ${if (isApplicable) "should" else "shouldn't"} be applicable to given case:\n\n$testCase"
        }
    }
```

With regard to other macro templates, I think we shouldn't provide too many of them, but `.format` looks useful to me. @Undin what do you think?

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    override fun expandForChooseExpression(expression: PsiElement, editor: Editor) {
+        if (expression !is RsExpr) return
+
+        val psiFactory = RsPsiFactory(expression.project)
+
+        val type = expression.type
+        if (type is TyReference && type.referenced is TyStr) {
+            val printlnExpression = psiFactory.createExpression("$macroName!(${expression.text})")
```
ortem: Let's use `MacroCreator` here, maybe something like:
```
        if (type is TyReference && type.referenced is TyStr) {
            val macroCreator = MacroCreator(editor, psiFactory, macroName)
            val printlnExpression = macroCreator.createMacro(expression.text, addTrailingSemicolon = true)
            val newElement = expression.replace(printlnExpression)
            macroCreator.updateCaret(newElement)
```

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,126 @@
...
+            if (addTrailingSemicolon) {
+                printlnExpression.add(psiFactory.createSemicolon())
+            }
+            return printlnExpression
+        }
+
+        fun updateSelection(macroElement: PsiElement) {
+            val selectionStart = macroElement.startOffset + macroStart.length
+            val selectionEnd = selectionStart + formatter.length
+            editor.selectionModel.setSelection(selectionStart, selectionEnd)
```
ortem: I'm not sure whether it should set selection. Is it really useful? In my case, once I've got generated `println!`, I want to press `Enter` and continue writing code. But `setSelection` force me to unselect it to continue.
If you think it is useful, just keep it. Otherwise, consider something like this:
```
        fun updateCaret(macroElement: PsiElement) {
            val offset = if (fmt == Fmt.None) {
                macroElement.endOffset
            } else {
                macroElement.startOffset + macroStart.length + formatter.length
            }
            editor.caretModel.moveToOffset(offset)
        }
```

SomeoneToIgnore:
> Did you try something like this? Seems like it helps.

It did help indeed, thanks a lot. Now all the tests are parameterized which reduced the code amount.

> I think we shouldn't provide too many of them, but .format looks useful to me.

Had exactly the same thoughts, good. Will do it after one is dealt with.

> I'm not sure that the behavior in case of RsLetDecl is convenient:

That is a very good find, thanks for noticing.
Although I was not able to get the expression selection dialogue on my machine — it just takes the whole expression, so I wonder how did you achieve the behavior.
For the same reason I've created only one test.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
SomeoneToIgnore: I guess, I'm too used to the awesome `Shift + Enter` shortcut hence did not consider as a problem.
After some thinking I tend to agree with you and it's way more confusing this way, removed the selection completely.

I see a small contradiction between the phrase
> I want to press Enter and continue writing code

and the proposed way to place a caret somewhere around the formatter
> macroElement.startOffset + macroStart.length + formatter.length

so I just place the cared in the end of the println expression in every case.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.rust.lang.core.types.ty.TyStr
+import org.rust.lang.core.types.ty.TyUnit
+import org.rust.lang.core.types.type
+
+class PrintlnPostfixTemplate(provider: RsPostfixTemplateProvider, private val macroName: String = "println") :
+    PostfixTemplateWithExpressionSelector(null, macroName, "$macroName!(\"{:?}\", expr);",
+        RsTopMostInScopeSelector { it.isNotIgnored && (it.implementsDebug || it.implementsDisplay) }, provider) {
+
+    private class MacroCreator(private val editor: Editor, private val psiFactory: RsPsiFactory,
+                               macroName: String, isDebug: Boolean) {
```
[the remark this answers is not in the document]
SomeoneToIgnore: Thanks, nice point.
I've slightly modified the example, main points:
* require `Fmt` always and add the `&str` case into `fromExpr` so that it can return an empty formatter also. This allows to throw away the first `if` branch completely and make the code way smaller.

* I removed the trailing `\n` from your snippet, since it seems like they get trimmed from the string and there is no trace of them in the actual expression

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,126 @@
...
+
+private fun RsBlock.addContents(psiFactory: RsPsiFactory, contents: List<PsiElement>) {
+    addBefore(psiFactory.createNewline(), rbrace)
+    for (element in contents) {
+        addBefore(element, rbrace)
+        addBefore(psiFactory.createNewline(), rbrace)
+    }
+}
+
+private val RsExpr.isNotIgnored: Boolean
```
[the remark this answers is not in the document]
SomeoneToIgnore: Nice, had applied a slightly refactored suggestion.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,138 @@
...
+                newElement.add(psiFactory.createNewline())
+                if (parent.semicolon == null) {
+                    parent.add(psiFactory.createSemicolon())
+                }
+            }
+            is RsMatchArm -> {
+                if (isUnitMatchExpressionType(parent)) {
+                    macroCreator.replaceWithMacro(expression, parent.comma == null)
+                } else {
+                    val newBlock = psiFactory.createBlockExpr("")
```
ortem: You can get rid of `addContents`:
```Kotlin
val macro = macroCreator.createMacro(expression.text)
val newBlock = psiFactory.createBlockExpr("${macro.text}\n${expression.text}")
val newElement = expression.replace(newBlock) as RsBlockExpr
editor.caretModel.moveToOffset(newElement.block.children[0].endOffset)
```

ortem:
> Although I was not able to get the expression selection dialogue on my machine — it just takes the whole expression, so I wonder how did you achieve the behavior.

Just try `let b = a - 42.println * 3`.

IMO, the functionality looks good now. You can improve it later (e.g. handle other cases in addition to `match` and `let`) if you'd like to.

With regard to tests, I've just looked at them and it seems like you wanted to test the type inference too 😄 .  I mean, there is no need to create so many test cases for different ways of implementing `Debug` and `Display` traits: you can just rely on our type inference here. Please try to reduce the complexity and the number of such tests (e.g. `FormatterTraitStructAndEnumPostfixPrintlnTests` and `MultipleFormatterTraitStructAndEnumPostfixPrintlnTests`), becasue we already have a lot of tests for type inference and resolution.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    private enum class Fmt(val text: String) {
+        None(""),
+        Display("{}"),
+        Debug("{:?}");
+
+        companion object {
+            fun fromExpr(expr: RsExpr): Fmt {
+                val type = expr.type
+                return if (type is TyReference && type.referenced is TyStr) {
```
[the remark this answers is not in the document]
SomeoneToIgnore: Indeed, thanks for noticing.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,138 @@
...
+        if (expression !is RsExpr) return
+
+        val psiFactory = RsPsiFactory(expression.project)
+        val macroCreator = MacroCreator(editor, psiFactory, macroName, Fmt.fromExpr(expression))
+
+        val parent = expression.parent
+        when (parent) {
+            is RsLetDecl -> {
+                // RsExpr.isNotIgnored condition in the selector guarantees pattern presence
+                val printlnExpression = macroCreator.createMacro(parent.pat!!.text)
```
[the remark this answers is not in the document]
SomeoneToIgnore: I would avoid using empty pattern with the current approach to the cursor placement:
image

so I've user `return` instead.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,138 @@
...
+                    newBlock.block.addContents(psiFactory, listOf(macroCreator.createMacro(expression.text), expression))
+                    val newElement = expression.replace(newBlock) as RsBlockExpr
+                    editor.caretModel.moveToOffset(newElement.block.children[0].endOffset)
+                }
+            }
+            else -> macroCreator.replaceWithMacro(expression)
+        }
+    }
+
+    private fun isUnitMatchExpressionType(parent: PsiElement) =
```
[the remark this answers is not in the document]
SomeoneToIgnore: Done, thanks for the proposition.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
SomeoneToIgnore: Great, haven't thought of it, thanks.

SomeoneToIgnore:
> Just try let b = a - 42.println * 3.

Have covered this case, thanks.
I have an issue with it though — whenever there's an expression select dialogue (as in `let b = a - 42 * 3.println` case), the element insertion breaks the formatting:
image

If there's no dialogue (as in `let b = a - 42.println * 3` case), everything's fine
(it's a bit weird that there's no selection dialogue in this case, by the way).

I use the same code for adding that line: `addMacroNextToElement` and I am currently clueless on how to fix it, so would appreciate any help.

> IMO, the functionality looks good now. You can improve it later

Thanks. If you get any issues related with the macro, please ping me and I'll do my best to fix those.

> becasue we already have a lot of tests for type inference and resolution.

Makes sense, I presume explicit `canSelectWithDeref` usage requirement and a small bug from the OP you've fixed later made me slightly paranoid hence all the redundant tests.
In this case, all the parameterized tests can go away for now, but might get useful if `format` postfix template will be introduced.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,140 @@
...
+            val newElement = childToContainer.second.addAfter(macro, childToContainer.first.getNextNonCommentSibling()?.prevSibling)
+            editor.caretModel.moveToOffset(newElement.endOffset)
+            newElement.add(psiFactory.createNewline())
+        }
+    }
+
+    private fun surroundWithBlockExpression(macro: RsExpr, originalExpression: PsiElement, psiFactory: RsPsiFactory, editor: Editor) {
+        val newBlock = psiFactory.createBlockExpr("${macro.text}\n${originalExpression.text}")
+        val newElement = originalExpression.replace(newBlock) as RsBlockExpr
+        editor.caretModel.moveToOffset(newElement.block.children[0].endOffset)
```
ortem: Looks like you should reformat the new block, for example, for such case:
```
let x = match 42 {
    0..10 => 5 + 7.println * 2
};
```
I think you can just put
```Kotlin
CodeStyleManager.getInstance(newElement.project).reformat(newElement.block.ancestorStrict<RsBlock>()?: newElement.block)
```
at the end of this function.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+private val PsiElement.closestContainerAncestor: Pair<PsiElement, PsiElement>?
+    get() = ancestorPairs.dropWhile { it.second !is RsBlock && it.second !is RsMatchArm }.firstOrNull()
+
+
```
ortem: Redundant line

ortem:
> In this case, all the parameterized tests can go away for now, but might get useful if format postfix template will be introduced.

OK, I agree with you!

Great job, thanks a lot! Please squash your commits (and fix the last review comments), and then I'll merge it.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private fun addMacroNextToElement(macro: RsExpr, element: PsiElement, psiFactory: RsPsiFactory, editor: Editor) {
+        val childToContainer = element.closestContainerAncestor ?: return
+        if (childToContainer.second is RsMatchArm) {
+            surroundWithBlockExpression(macro, childToContainer.first, psiFactory, editor)
+        } else {
+            val newElement = childToContainer.second.addAfter(macro, childToContainer.first.getNextNonCommentSibling()?.prevSibling)
+            editor.caretModel.moveToOffset(newElement.endOffset)
+            newElement.add(psiFactory.createNewline())
```
ortem:
```
newElement.add(psiFactory.createNewline())
newElement.ancestorStrict<RsBlock>()?.let {
    CodeStyleManager.getInstance(newElement.project).reformat(it)
}
```
Seems like it solves your problem with `let b = a - 42 * 3.println`. If no, just leave it the way it is, and I'll try to fix this issue later.

SomeoneToIgnore: Great, works like a charm.
Had to use `trimIndent` in a few `let` tests because of this.

src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
SomeoneToIgnore: Removed.

SomeoneToIgnore:
Had squashed all the commits and committed the fixes requested.
Thanks for your time spent on this one.

PR #3234, opened 2019-01-09
Run clippy for with project's channel

### LINKED ISSUE #3135: Cargo sidebar only runs Clippy as `+nightly`
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2497-182-nightly
* **Rust toolchain version:** 1.3.1 stable-x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2018.2.6 (Ultimate Edition) Build #IU-182.5107.16
* **Operating system:** macOS 10.14.1

## Problem description

Clippy is run by the Cargo side panel using `+nightly` when a stable toolchain is set to be default, leading to the error:

```
/Users/James/.cargo/bin/cargo +nightly clippy --color=always
error: 'cargo-clippy' is not installed for the toolchain 'nightly-x86_64-apple-darwin'
To install, run `rustup component add clippy`

Process finished with exit code 1
```

## Steps to reproduce

1. `rustup default stable`
1. Install Clippy
   ```
   $ rustup component add clippy
   $ rustup component list | grep clippy
   clippy-x86_64-apple-darwin (installed)
   ```
3. `cargo new --bin clippy-nightly`
4. Open the project in IDEA
5. Open the side bar for Cargo and click the Clippy button. Error above appears.

SomeoneToIgnore:
Closes #3135

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
@@ -7,7 +7,6 @@ package org.rust.cargo.runconfig.command
 
 import com.intellij.openapi.actionSystem.AnActionEvent
 import org.rust.cargo.icons.CargoIcons
-import org.rust.cargo.project.settings.toolchain
 import org.rust.cargo.runconfig.getAppropriateCargoProject
 import org.rust.cargo.toolchain.CargoCommandLine
 import org.rust.cargo.toolchain.RustChannel
@@ -17,9 +16,8 @@ class RunClippyAction : RunCargoCommandActionBase(CargoIcons.CLIPPY) {
 
     override fun actionPerformed(e: AnActionEvent) {
         val project = e.project ?: return
-        val toolchain = project.toolchain ?: return
         val cargoProject = getAppropriateCargoProject(e.dataContext) ?: return
-        val channel = if (toolchain.isRustupAvailable) RustChannel.NIGHTLY else RustChannel.DEFAULT
+        val channel = cargoProject.rustcInfo?.version?.channel ?: RustChannel.DEFAULT
         CargoCommandLine.forProject(cargoProject, "clippy", channel = channel).run(project, cargoProject)
     }
 }


src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
```diff
@@ -17,9 +16,8 @@ class RunClippyAction : RunCargoCommandActionBase(CargoIcons.CLIPPY) {
 
     override fun actionPerformed(e: AnActionEvent) {
         val project = e.project ?: return
-        val toolchain = project.toolchain ?: return
         val cargoProject = getAppropriateCargoProject(e.dataContext) ?: return
-        val channel = if (toolchain.isRustupAvailable) RustChannel.NIGHTLY else RustChannel.DEFAULT
+        val channel = cargoProject.rustcInfo?.version?.channel ?: RustChannel.DEFAULT
```
Undin: Looks like we don't need set specific channel at all because it can lead to the following situation:
We have `1.31.0` and `stable` toolchains. `1.31.0` is default.
Also, we've installed clippy only for `1.31.0`.
And when we invoke this action, we'll launch `cargo +stable clippy` command and get `error: 'cargo-clippy' is not installed for the toolchain 'stable-...'` message

SomeoneToIgnore: Thank you for the suggestion, had pushed the changes suggested.

Could you explain what is the meaning of the value stored in `cargoProject.rustcInfo?.version` then?
I was under the impression that this is the version of the toolchain used in the current cargo project (and in this context selecting the clippy tool from a different toolchain sounded like a bad idea), but apparently I'm wrong.

Undin: > Could you explain what is the meaning of the value stored in cargoProject.rustcInfo?.version then?

It's a version of rustc for current project stored after the latest project model refresh

> In this context selecting the clippy tool from a different (default) toolchain sounded like a bad idea, but apparently I'm wrong and this version is something else.

But your latest change is exactly about using default toolchain for the current project instead of nightly toolchain i.e. the same toolchain version as user uses for project compilation.

Undin: Looks like `default` word is misleading here
In our case `default` channel means that we **don't** add any channel argument to the command line, i.e. we use **active** toolchain for this project

Undin:
Thanks!
bors r+

src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
SomeoneToIgnore: Thanks for the explanations, now it's clearer.

PR #3240, opened 2019-01-11
RES: fix Self-qualified path resolve

### LINKED ISSUE #3029: False-positive annotation when returning Self::Type
## Environment

* **Intellij-Rust plugin version**: 0.2.0.2103-182
* **Rust toolchain version**: rustc 1.28.0 (9634041f0 2018-07-30)
* **IDE name and version**: IntelliJ IDEA 2018.2 (Ultimate Edition) Build #IU-182.3684.101, built on July 24, 2018
* **Operating system**: OS X 10.12.6

## Problem description

Syntax highligh error: image

May be related to #1549

## Steps to reproduce

```rust
trait Y {
  type Result;
  fn handle() -> Result
}

impl Y for X {
    type Result = ();

    fn handle() -> Self::Result {
      debug!("Foo");
    }
}
```

### LINKED ISSUE #3125: Missing `mismatched types` error if one of the types is an alias
## Environment

* **Intellij-Rust plugin version:** 0.2.0.2492-183-nightly
* **Rust toolchain version:** 1.32.0-nightly
* **IDE name and version:** IC-183.4284.148
* **Operating system:** macOS Mojave 10.14.1

## Problem description

The editor does not report the `mismatched types` error if one of the types is a type alias.

```rust
struct S;

trait T {
    type Item;
    fn foo() -> Self::Item;
}

impl T for S {
    type Item = &'static str;
    fn foo() -> Self::Item {
        let x: Self::Item = 42;
        42
    }
}
```

I expect to see the `mismatched types` error in `foo` implementation.

## Steps to reproduce

* Copy-paste code from problem description.

vlad20012:
Fixes #3125
Fixes #3029

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -362,8 +362,32 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             if (processor("self", base)) return true
         }
         if (processItemOrEnumVariantDeclarations(base, ns, processor, withPrivateImports = isSuperChain)) return true
-        if (base is RsTypeDeclarationElement && parent !is RsUseSpeck) {
-            // Foo::<Bar>::baz
+        if (base is RsTypeDeclarationElement && parent !is RsUseSpeck) { // Foo::<Bar>::baz
+            val shadowingProcessor = run {
+                // Self-qualified type paths `Self::Item` inside impl items are restricted to resolve
+                // to only members of the current impl or implemented trait or its parent traits
+                if (Namespace.Types in ns && base is RsImplItem && qualifier.hasCself) {
+                    NameResolutionTestmarks.selfRelatedTypeSpecialCase.hit()
+                    val traits = base.implementedTrait?.flattenHierarchy
+                        ?.map { it.foldTyTypeParameterWith { TyInfer.TyVar(it) } }
+                        ?: return@run processor
+
+                    fun(e: AssocItemScopeEntry): Boolean {
+                        if (e.element !is RsTypeAlias) return processor(e)
+
+                        val implementedTrait = e.source.value.implementedTrait
+                            ?.foldTyTypeParameterWith { TyInfer.TyVar(it) }
+                            ?: return processor(e)
+
+                        val isAppropriateTrait = traits.any {
+                            lookup.ctx.probe { lookup.ctx.combineBoundElements(it, implementedTrait) }
+                        }
+                        return if (isAppropriateTrait) processor(e) else false
+                    }
+                } else {
+                    processor
+                }
+            }
             val selfTy = if (base is RsImplItem && qualifier.hasCself) {
                 // impl S { fn foo() { Self::bar() } }
                 base.typeReference?.type ?: TyUnknown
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -362,8 +362,32 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             if (processor("self", base)) return true
         }
         if (processItemOrEnumVariantDeclarations(base, ns, processor, withPrivateImports = isSuperChain)) return true
-        if (base is RsTypeDeclarationElement && parent !is RsUseSpeck) {
-            // Foo::<Bar>::baz
+        if (base is RsTypeDeclarationElement && parent !is RsUseSpeck) { // Foo::<Bar>::baz
+            val shadowingProcessor = run {
```
Undin: Maybe we want to extract this into separate function? Because path resolution is already too big

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -362,8 +362,32 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
...
+                // Self-qualified type paths `Self::Item` inside impl items are restricted to resolve
+                // to only members of the current impl or implemented trait or its parent traits
+                if (ns == TYPES && base is RsImplItem && qualifier.hasCself) {
```
Undin: Are you sure that we want to check `ns == TYPES` instead of `TYPES in ns` here?

vlad20012: Oh, it should work without it, I just forget to remove it

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
vlad20012: I don't sure this peace of code is good for extraction. We can extract large branches (in a separate PR) instead: where `if (qualifier != null)`, `if (typeQual != null)` and remained

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -362,8 +362,32 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
...
+                            ?.foldTyTypeParameterWith { TyInfer.TyVar(it) }
+                            ?: return processor(e)
+
+                        val isAppropriateTrait = traits.any {
+                            lookup.ctx.probe { lookup.ctx.combineBoundElements(it, implementedTrait) }
+                        }
+                        return if (isAppropriateTrait) processor(e) else false
+                    }
+                } else {
+                    { e: AssocItemScopeEntry -> processor(e) }
```
mchernyavsky: Looks like we can just return `processor` here (and above), because `(ScopeEntry) -> Boolean` is contravariant to `(AssocItemScopeEntry) -> Boolean`.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -362,8 +362,32 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
             if (processor("self", base)) return true
         }
         if (processItemOrEnumVariantDeclarations(base, ns, processor, withPrivateImports = isSuperChain)) return true
-        if (base is RsTypeDeclarationElement && parent !is RsUseSpeck) {
-            // Foo::<Bar>::baz
+        if (base is RsTypeDeclarationElement && parent !is RsUseSpeck) { // Foo::<Bar>::baz
+            val shadowingProcessor = run {
+                // Self-qualified type paths `Self::Item` inside impl items are restricted to resolve
+                // to only members of the current impl or implemented trait or its parent traits
+                if (Namespace.Types in ns && base is RsImplItem && qualifier.hasCself) {
```
mchernyavsky: We can write method implementations in traits, not only in impls:
```rust
trait Trait {
    type Item;
    fn foo() -> Self::Item;
    fn bar() -> Self::Item { Self::foo() } // ok
    fn baz() -> Self::Item { 42 } // error
}
```

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
vlad20012: I'd do that if it were typechecked :(

PR #3243, opened 2019-01-13
TY&RES: Take into account super traits defined with where clause

vlad20012:
Co-Authored-By: vlad9486 <vlad9486@gmail.com> (#3049)

```rust
trait Bar {
    fn bar(&self);
     //X
}
trait Foo where Self : Bar {
    fn foo(&self) { self.bar(); }
}                      //^
```

This is NOT a fix for #2530 b/c we still ignore fn-level Self bounds


mchernyavsky:
bors retry

Issue #3245, opened 2019-01-13
No type annotation in destructuring structure


ark0f:
## Environment

* **Intellij-Rust plugin version:** v0.2.0.2113-183
* **IDE name and version:** CLion 2018.3.2

## Problem description
Invalid syntax highlighting

## Steps to reproduce
![](https://i.imgur.com/sXcyKc9.png)

ark0f:
Sorry, line `b: String,` is correct, `: String` means variable name.

ark0f:
Here is correct image:

![](https://i.imgur.com/6LqM7GL.png)

PR #3250, opened 2019-01-15
TYPE: Provide type hints for shorthands in pat literals

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/3245.

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
@@ -22,31 +22,23 @@ import org.rust.lang.core.types.type
 import org.rust.stdext.buildList
 
 private val RsPat.inlayInfo: List<InlayInfo>
-    get() {
-        val idents = if (this is RsPatIdent) {
-            listOf(this)
-        } else {
-            descendantsOfType<RsPatIdent>()
-        }
-        return idents.asSequence()
-            .filter {
-                val patBinding = it.patBinding
-                when {
-                    // "ignored" bindings like `let _a = 123;`
-                    patBinding.referenceName.startsWith("_") -> false
-                    // match foo {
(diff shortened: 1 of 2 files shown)


Undin:
@mchernyavsky Maybe merge #3217 before it?
It changes the same code and introduces a little bit more convenient API for testing

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -23,14 +23,13 @@ import org.rust.stdext.buildList
 
 private val RsPat.inlayInfo: List<InlayInfo>
     get() {
-        val idents = if (this is RsPatIdent) {
-            listOf(this)
+        val patBindings = if (this is RsPatBinding) {
```
Undin: Looks like it's always false because `RsPatBinding` isn't `RsPat`

src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
```diff
@@ -23,14 +23,13 @@ import org.rust.stdext.buildList
...
     get() {
-        val idents = if (this is RsPatIdent) {
-            listOf(this)
-        } else {
-            descendantsOfType<RsPatIdent>()
+        val patBindings = when {
+            this is RsPatIdent -> listOf(patBinding)
+            this is RsPatField && patBinding != null -> listOfNotNull(patBinding)
+            else -> descendantsOfType<RsPatBinding>()
         }
```
Undin: It seems we can just use `descendantsOfType<RsPatBinding>()` here for all cases

mchernyavsky: Fixed. But we could have done so in the original code too (`descendantsOfType` isn't strict), so I thought it wasn't made for performance reasons.

PR #3256, opened 2019-01-16
Fix expressions replacing inside introduce refactorings

ortem:
1. Replacing `RsExpr` with `PsiIdentifier` leads to problems with parsing result, so now it uses `createExpression`.
2. `MemberInplaceRenamer` may not work correctly when the user starts typing right after introducing a parameter. So it was replaced with `RustInPlaceVariableIntroducer` (also providing all suggested names).

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
@@ -133,7 +133,7 @@ private fun findAnchor(expr: PsiElement): PsiElement? {
     return anchor
 }
 
-private class RustInPlaceVariableIntroducer(
+class RustInPlaceVariableIntroducer(
     elementToRename: PsiNamedElement,
     editor: Editor,
     project: Project,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
```diff
@@ -134,7 +133,7 @@ private fun findAnchor(expr: PsiElement): PsiElement? {
     return anchor
 }
 
-private class RustInPlaceVariableIntroducer(
+class RustInPlaceVariableIntroducer(
```
vlad20012: Maybe move the class to some common pace?

ortem: Done, created `introduceUtils.kt` for such needs.

src/main/kotlin/org/rust/ide/refactoring/introduceUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+
+package org.rust.ide.refactoring
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiNamedElement
+import com.intellij.refactoring.introduce.inplace.InplaceVariableIntroducer
+
+class RustInPlaceVariableIntroducer(
```
Undin: Let's use `Rs` prefix for such names to be consistent with all other code
And maybe it's worth to rename the file to `RsInPlaceVariableIntroducer`? It contains only this class

Undin:
bors retry

Undin:
bors r=vlad20012

PR #3264, opened 2019-01-19
TY&RES: hardcode `serde` and `failure` custom derives

vlad20012:
Now derive for Serialize, Deserialize and Fail work as well as builtin derives.

I think this makes sense even if we'll support procedural macros. Just as an optimization - serde custom derives are very widespread

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
@@ -128,3 +128,25 @@ private class RealKnownItemsLookup(
         }.orElse(null)
     }
 }
+
+enum class KnownDerivableTrait(
+    private val resolver: (KnownItems) -> RsTraitItem?,
+    val dependencies: Array<KnownDerivableTrait> = emptyArray()
+) {
+    Clone(KnownItems::Clone),
+    Copy(KnownItems::Copy, arrayOf(Clone)),
+    Debug(KnownItems::Debug),
+    Default(KnownItems::Default),
+    Hash(KnownItems::Hash),
+    PartialEq(KnownItems::PartialEq),
+    Eq(KnownItems::Eq, arrayOf(PartialEq)),
+    PartialOrd(KnownItems::PartialOrd, arrayOf(PartialEq)),
+    Ord(KnownItems::Ord, arrayOf(PartialOrd, Eq, PartialEq)),
+    ;
+
+    fun findTrait(items: KnownItems): RsTraitItem? = resolver(items)
+}
+
+val KnownDerivableTrait.withDependencies: List<KnownDerivableTrait> get() = listOf(this, *dependencies)
+
+val KNOWN_DERIVABLE_TRAITS: Map<String, KnownDerivableTrait> = KnownDerivableTrait.values().associate { it.name to it }
--- a/src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
@@ -142,6 +142,12 @@ enum class KnownDerivableTrait(
     Eq(KnownItems::Eq, arrayOf(PartialEq)),
     PartialOrd(KnownItems::PartialOrd, arrayOf(PartialEq)),
     Ord(KnownItems::Ord, arrayOf(PartialOrd, Eq, PartialEq)),
+
+    Serialize({ it.findTrait("serde::Serialize") }),
+    Deserialize({ it.findTrait("serde::Deserialize") }),
+
+    // TODO Fail also derives `Display`. Ignore it for now
+    Fail({ it.findTrait("failure::Fail") }, arrayOf(Debug)),
     ;
 
     fun findTrait(items: KnownItems): RsTraitItem? = resolver(items)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
```diff
@@ -125,3 +125,31 @@ private class RealKnownItemsLookup(
...
+    PartialEq({ it.PartialEq }),
+    Eq({ it.Eq }, arrayOf(PartialEq)),
+    PartialOrd({ it.PartialOrd }, arrayOf(PartialEq)),
+    Ord({ it.Ord }, arrayOf(PartialOrd, Eq, PartialEq)),
+
+    Serialize({ it.findTrait("serde::Serialize") }),
+    Deserialize({ it.findTrait("serde::Deserialize") }),
+
+    // TODO Fail also derives `Display`. Ignore it for now
+    Fail({ it.findTrait("failure::Fail") }),
```
Undin: Looks like doesn't work with `Fail`

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
```diff
@@ -125,3 +125,31 @@ private class RealKnownItemsLookup(
...
+    Hash({ it.Hash }),
+    PartialEq({ it.PartialEq }),
+    Eq({ it.Eq }, arrayOf(PartialEq)),
+    PartialOrd({ it.PartialOrd }, arrayOf(PartialEq)),
+    Ord({ it.Ord }, arrayOf(PartialOrd, Eq, PartialEq)),
+
+    Serialize({ it.findTrait("serde::Serialize") }),
+    Deserialize({ it.findTrait("serde::Deserialize") }),
+
+    // TODO Fail also derives `Display`. Ignore it for now
```
Undin: It also requires `Debug` so we can add the corresponding dependency

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
```diff
@@ -125,3 +125,31 @@ private class RealKnownItemsLookup(
...
+    Copy({ it.Copy }, arrayOf(Clone)),
+    Debug({ it.Debug }),
+    Default({ it.Default }),
+    Hash({ it.Hash }),
+    PartialEq({ it.PartialEq }),
+    Eq({ it.Eq }, arrayOf(PartialEq)),
+    PartialOrd({ it.PartialOrd }, arrayOf(PartialEq)),
+    Ord({ it.Ord }, arrayOf(PartialOrd, Eq, PartialEq)),
+
+    Serialize({ it.findTrait("serde::Serialize") }),
```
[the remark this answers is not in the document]
vlad20012: KnownItems just resolve that path. The trait is re-exported as `serde::Serialize`, so we can resolve it. I think we should use `serde::Serialize` path b/c it is public API, internals can be changed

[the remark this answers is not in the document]
Undin: Oh, I've understood
> I think we should use serde::Serialize path b/c it is public API, internals can be changed

Agree

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
```diff
@@ -1126,4 +1126,18 @@ class RsResolveTest : RsResolveTestBase() {
...
+    fun `test derive serde Serialize`() = checkByCode("""
+    //- main.rs
+        #[lang = "serde::Serialize"]
+        trait Serialize { fn serialize(&self); }
+                           //X
+        #[derive(Serialize)]
+        struct Foo;
+
+        fn bar(foo: Foo) {
+            foo.serialize();
```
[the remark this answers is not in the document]
vlad20012: The `Reference.resolve()` works as specified in test. The "go-to-definition" action does not exactly follow the results of resolve. For derived trait it always point to a derive attribute. It works for serde as well as for standard derives

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
vlad20012: Hmm, but I tested it =/

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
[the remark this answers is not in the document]
Undin: Indeed!

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
vlad20012: Works for me
peek 2019-01-22 19-17

vlad20012:
Fixed

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
Undin: Resolve works as expected, but completion doesn't work for `Fail`.
It's interesting because everything is ok for `Serialize`/`Deserialize` traits

vlad20012: FIxed

src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
```diff
@@ -37,7 +37,7 @@ class RsDeriveCompletionProviderTest : RsCompletionTestBase() {
     """)
 
     fun `test complete with dependencies`() {
-        StdDerivableTrait.values()
+        KnownDerivableTrait.values()
```
Undin: This change brokes test because we can't find `Serialize` and `Fail` trait in index

PR #3266, opened 2019-01-20
Do not add extra lifetimes for the methods implemented 

SomeoneToIgnore:
Currently the plugin adds extra lifetimes when processing the functions with reference types with no parameters.

Example: [`Debug` trait, `fmt` method](https://doc.rust-lang.org/std/fmt/trait.Debug.html#tymethod.fmt): `fn fmt(&self, f: &mut Formatter) -> Result<(), Error>`

When the implementation is generated by the plugin, the following code is created:
```
fn fmt(&self, f: &mut Formatter<'a>) -> Result<(), Error> {
	unimplemented!()
}
```

It seems like the generation should not add lifetimes at all, if they are not defined in the original method, that's what I've tried to achieve with the PR.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -82,7 +82,7 @@ class RsPsiFactory(private val project: Project) {
     }
 
     fun createStructLiteral(name: String): RsStructLiteral =
-        createExpressionOfType<RsStructLiteral>("$name { }")
+        createExpressionOfType("$name { }")
 
     fun createStructLiteralField(name: String, value: RsExpr? = null): RsStructLiteralField {
         val structLiteralField = createExpressionOfType<RsStructLiteral>("S { $name: () }")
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore:
Idea had automatically reformatted a few lines in the file, I can revert those changes back, if necessary.

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -368,8 +371,10 @@ private fun RsFunction.getSignatureText(subst: Substitution): String? {
 
 private fun String.iff(cond: Boolean) = if (cond) this + " " else " "
 
-private fun RsTypeReference.substAndGetText(subst: Substitution): String =
-    type.substitute(subst).insertionSafeTextWithLifetimes
+private fun RsTypeReference.substAndGetText(subst: Substitution): String {
+    val substitutedType = type.substitute(subst)
+    return if (this.refLikeType?.lifetime != null) substitutedType.insertionSafeTextWithLifetimes else substitutedType.insertionSafeText
```
ortem: This line is too long, consider introducing `val hasLifetime = refLikeType?.lifetime != null`
Although there are many too long lines in the project, we try to avoid new ones 😄

ortem:
> Idea had automatically reformatted a few lines in the file, I can revert those changes back, if necessary.

It's OK, just split the changes into two commits like `Do not add extra lifetimes...` and `Cleanup`.

SomeoneToIgnore:
Had split the commit.

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
SomeoneToIgnore: Sure thing, done.

ortem:
Thanks!
bors r+

PR #3268, opened 2019-01-20
TY&RES: fix impls selection for impl Trait

vlad20012:


src/main/kotlin/org/rust/lang/core/resolve/indexes/RsImplIndex.kt
```diff
@@ -39,6 +39,13 @@ class RsImplIndex : AbstractStubIndex<TyFingerprint, RsImplItem>() {
             return (impls.asSequence() + freeImpls.asSequence()).filter { it.crateRoot != null }
         }
 
+        /** return impls for generic type `impl<T> Trait for T {}` */
+        fun findFreeImpls(project: Project): Sequence<RsImplItem> {
```
mchernyavsky: Can we use this function inside `findPotentialImpls`?

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -628,10 +639,10 @@ class ImplLookup(
         return when (selfTy) {
             is TyTypeParameter -> lookupAssocTypeInBounds(selfTy.getTraitBoundsTransitively(), res.impl, assocType)
             is TyTraitObject -> selfTy.trait.assoc[assocType]
-            is TyAnon -> lookupAssocTypeInBounds(selfTy.getTraitBoundsTransitively(), res.impl, assocType)
             else -> {
                 lookupAssocTypeInSelection(res, assocType)
                     ?: lookupAssocTypeInBounds(getHardcodedImpls(selfTy), res.impl, assocType)
+                    ?: (selfTy as? TyAnon)?.let { lookupAssocTypeInBounds(it.getTraitBoundsTransitively(), res.impl, assocType) }
```
mchernyavsky: Can you cover this case with a test?

vlad20012: I sure that it is covered

PR #3278, opened 2019-01-21
TY: Do not implement the trait `Sized` for `str`

kumbayo:
This allows us to to correctly report an error `E0277` for the following code:
```rust
    fn foo(a: str) { }
```


Undin:
Thanks!
bors r+

Undin:
bors retry

PR #3280, opened 2019-01-22
REF: allow extract function from single expression

vlad20012:


src/main/kotlin/org/rust/ide/utils/SearchByOffset.kt
```diff
@@ -62,8 +63,31 @@ fun findExpressionInRange(file: PsiFile, startOffset: Int, endOffset: Int): RsEx
...
+
+/**
+ * Finds statements (mostly [RsStmt]s, [PsiComment]s and [RsExpr]
+ * as return expression, doesn't allow attrs and items) or a single expr within selected range.
+ */
+fun findStatementsOrExprInRange(file: PsiFile, startOffset: Int, endOffset: Int): Array<out PsiElement> {
+    val (element1, element2) = file.getElementRange(startOffset, endOffset) ?: return emptyArray()
+    val parent = PsiTreeUtil.findCommonParent(element1, element2) ?: return emptyArray()
+
+    val mostDistantParent = parent.ancestors
```
ortem: Maybe `val mostDistantParent = parent.ancestors.last { it.textRange == parent.textRange }`?
`lastOrNull` is redundant here because there is at least one such ancestor — the `parent` itself.

src/main/kotlin/org/rust/ide/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -99,11 +99,21 @@ class RsExtractFunctionConfig private constructor(
...
             append(whereClausesText)
-            val stmts = elements.map { it.text }.toMutableList()
-            if (returnValue?.expression != null) {
-                stmts.add(returnValue.expression)
+        }
+
+    val functionText: String
+        get() = buildString {
+            append(signature)
+            val body = if (elements.size == 1 && elements.single() is RsBlockExpr) {
```
ortem: Maybe
```
val single = elements.singleOrNull()
val body = if (single is RsBlockExpr) {
    single.block.text
} else {
```

vlad20012:
Done

PR #3292, opened 2019-01-26
T: run test with clion

Undin:
* Download CLion via gradle-intellij-plugin
* Add ability to change base IDE for plugin compilation/test running. Of course, some modules, for example, `idea` and `debugger` can be compiled only with IDEA and CLion respectively because they use features of concrete IDE
* Add Travis build to compile and run tests with CLion sdk


Undin:
@ortem Could you take a look at changes related to pretty printers tests?

pretty_printers_tests/src/main.rs
```diff
@@ -14,10 +16,32 @@ const SETTINGS: &str = "Settings_macos.toml";
...
+    print_lldb_stdout: bool,
+    lldb_native_rust: bool,
+}
 
+/// Expects `lldb_python` path as the first argument
 fn main() -> Result<(), ()> {
     let settings = fs::read_to_string(SETTINGS).expect(&format!("Cannot read {}", SETTINGS));
-    let lldb_config: LLDBConfig = toml::from_str(&settings).expect(&format!("Invalid {}", SETTINGS));
+    let settings: Settings = toml::from_str(&settings).expect(&format!("Invalid {}", SETTINGS));
+    let lldb_python = get_lldb_python_path();
```
ortem: Maybe place it at the beginning of the `main`? IMO there it's better to check argument before reading a file.

Also, it can be inlined like this (but it is a matter of taste):
```Rust
fn main() -> Result<(), ()> {
    let args: Vec<String> = std::env::args().collect();
    let lldb_python = args.get(1).expect("You need to pass a path to `lldb_python`").clone();
    let settings = fs::read_to_string(SETTINGS).expect(&format!("Cannot read {}", SETTINGS));
```

PR #3293, opened 2019-01-26
TY: Type inference for generators

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/2203. Relates to https://github.com/intellij-rust/intellij-rust/issues/2622.

Issue #2622, opened 2018-06-29
Support Rust 2018 edition


Undin:
[Edition guide](https://rust-lang-nursery.github.io/edition-guide/2018)

* [x] `impl Trait`
* [ ] `dyn Trait`
  * [x] parser and type inference support
  * [ ] inspection and quick fix to convert bare trait object type to `dyn Trait`
* [ ] module system
  * [ ] resolution without `extern crate` item
  * [ ] absolute path started from crate name or `crate`
  * [ ] `crate` visibility modifier
  * [ ] module hierarchy without `mod.rs` file
  * [ ] (re)export macro via `use` item
* [ ] async/await
* [ ] ownership and lifetimes
  * [ ] anonymous lifetime
  * [ ] in-band lifetimes
  * [ ] lifetime elision in impl
  * [ ] default match bindings
* [ ] slice patterns
* [ ] raw identifiers
* [ ] keywords
  * [ ] `async`, `await` and `try` should be keyword in 2018 edition
  * [ ] `proc` is not keyword anymore

jannik17:
I think [this cargo feature to rename a crate](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#cargo-features) is missing in the list.

Undin:
@jannik17 Yep, you are absolutely right
I've added the corresponding point

SomeoneToIgnore:
It looks like `edition` TOML field support is not mentioned here also.
I've created a PR to add the missing functionality, hope it will be useful: https://github.com/intellij-rust/intellij-rust/pull/3214

brokenthorn:
Seems like there's no type inference or autocomplete for all `pub async fn`.

For example, the `pub async fn text(self)` in reqwest's `Response` struct:

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let response_future = reqwest::get("https://httpbin.org/ip");
    let response = response_future.await?;
    let text_future = response.text();
    let text = text_future.await?;
    println!("{:#?}", text);
    Ok(())
}
```

![image](https://user-images.githubusercontent.com/2119004/70816416-2048ba00-1dd8-11ea-9dc1-968b8ff4807c.png)

Cargo dependencies:
```toml
[dependencies]
reqwest = "0.10.0-alpha.2"
tokio = "0.2.0-alpha.6"
futures = "0.3.1"
chrono = "0.4.10"
```

[...]

PR #3294, opened 2019-01-26
TY: Type inference for async/await

mchernyavsky:
Depends on https://github.com/intellij-rust/intellij-rust/pull/3293. Relates to https://github.com/intellij-rust/intellij-rust/issues/2622.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -657,17 +657,28 @@ class RsFnInferenceContext(
             tryTy = expected ?: TyInfer.TyVar()
             val resultTy = tryTy ?: TyUnknown
             val okTy = blockExpr.block.inferType()
-            registerTryProjection(blockExpr.knownItems, resultTy, "Ok", okTy)
+            registerTryProjection(resultTy, "Ok", okTy)
             return resultTy
         } finally {
             tryTy = oldTryTy
         }
     }
 
-    private fun inferAsyncBlockExprType(blockExpr: RsBlockExpr): TyUnknown {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -648,10 +648,22 @@ class RsFnInferenceContext(
         }
     }
 
-    private fun inferAsyncBlockExprType(blockExpr: RsBlockExpr): TyUnknown {
+    private fun inferAsyncBlockExprType(blockExpr: RsBlockExpr, expected: Ty? = null): Ty {
         require(blockExpr.isAsync)
-        blockExpr.block.inferType()
-        return TyUnknown
+        val expectedOutputTy = lookupFutureOutputTy(expected, blockExpr.knownItems)
```
vlad20012: > blockExpr.knownItems
inference context contains KnownItems already. `items` or `ctx.items`

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -648,10 +648,22 @@ class RsFnInferenceContext(
...
+        val outputTy = blockExpr.block.inferTypeCoercableTo(expectedOutputTy)
```
vlad20012: Is it really should be `inferTypeCoercableTo(expectedOutputTy)`? Or `inferType(expectedOutputTy)` or `inferType(null)` or a something else? I don't know the correct answer; it should be looked up in rustc

vlad20012: Hmm, looks like you should just ignore an expectation just b/c you can't provide any valid expectation - async block have `impl Trait` type that can by unified with itself only. Sooo simple `blockExpr.block.inferType()` should be used here

PR #3297, opened 2019-01-27
INT: extract from nested block

sohich:
Intention to extract contents of nested block to outer block. 
Made it in addition to inline method refactoring (finishing it) - methods are inlined in separate block to provide scope for valid parameters borrowing.

Simple example. Before:
```rust
fn foo() {
    let mut x = 0;
    
    {
        x = bar();
    }
    
    x = 0;
}
```
After:
```rust
fn foo() {
    let mut x = 0;

    x = bar();

    x = 0;
}
```


sohich:
> @sohich Could you clean up your commits? At this moment only one commit is about new feature

Have done it. Had some issues with this

src/main/kotlin/org/rust/ide/intentions/ExtractNestedBlockIntention.kt
```diff
@@ -0,0 +1,48 @@
...
+            .drop(1)
+            .toMutableList()
+            .dropLast(1)
+
+        for (elem in elements) {
+            parent.addBefore(elem, holder)
+        }
+
+        holder.delete()
+        parent.let {
```
vlad20012: This can format a wide range of statements. Furthermore, I don't expect this intention to re-format anything. Is this re-format added to fix indents? It's possible to fix indentation without re-format. See UnindentSelectionAction for example.

vlad20012:
Also it should not be available for unsafe/async/try blocks. See UnindentSelectionAction

vlad20012:
Close because of inactivity

PR #3299, opened 2019-01-28
REF: Make member inplace rename available for identifier owners

mchernyavsky:
See [comment](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000014084/comments/360000019204) for explanation.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRefactoringSupportProvider.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRefactoringSupportProvider.kt
@@ -12,9 +12,14 @@ import org.rust.ide.refactoring.extractFunction.RsExtractFunctionHandler
 import org.rust.ide.refactoring.introduceParameter.RsIntroduceParameterHandler
 import org.rust.ide.refactoring.introduceVariable.RsIntroduceVariableHandler
 import org.rust.lang.core.psi.RsPatBinding
+import org.rust.lang.core.psi.ext.RsNameIdentifierOwner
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/refactoring/RsRefactoringSupportProvider.kt
```diff
@@ -12,9 +12,14 @@ import org.rust.ide.refactoring.extractFunction.RsExtractFunctionHandler
...
 import org.rust.lang.core.psi.RsPatBinding
+import org.rust.lang.core.psi.ext.RsReferenceElement
 
 class RsRefactoringSupportProvider : RefactoringSupportProvider() {
-    override fun isInplaceRenameAvailable(element: PsiElement, context: PsiElement?): Boolean = element is RsPatBinding
+    override fun isInplaceRenameAvailable(element: PsiElement, context: PsiElement?): Boolean =
+        element is RsPatBinding
+
+    override fun isMemberInplaceRenameAvailable(element: PsiElement, context: PsiElement?): Boolean =
+        element is RsReferenceElement
```
Undin: In what cases does it should work?
Looks like element is always not `RsReferenceElement` even we call rename on reference element

mchernyavsky: That's a really good point!

mchernyavsky: Fixed.

PR #3300, opened 2019-01-28
[WIP] INSP: Constant condition

dimonchik0036:
TODO:
- [ ]  `break` and `continue` in loops.
- [ ]  Evaluate `if` and `match` expressions.

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
+++ b/src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
@@ -7,12 +7,9 @@ package org.rust.ide.utils
 
 import com.intellij.openapi.project.Project
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.ArithmeticOp
-import org.rust.lang.core.psi.ext.LogicOp
+import org.rust.lang.core.psi.ext.*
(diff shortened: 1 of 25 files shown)


src/main/kotlin/org/rust/ide/utils/ExprUtils.kt
```diff
@@ -235,10 +232,6 @@ fun RsExpr.skipParenExprUp(): RsExpr {
...
  * @return a child expression without parentheses.
  */
-fun RsCondition.skipParenExprDown(): RsExpr {
-    var child = this.expr
-    while (child is RsParenExpr) {
-        child = child.expr
-    }
-    return child
-}
+fun RsCondition.skipParenExprDown(): RsExpr = this.expr.skipParenExprDown()
```
vlad20012: Whyyyyyyyyyyy did you replaced a loop with a recursion? use `tailrec` modifier at least

dimonchik0036: It isn't a recursive call, because `RsCondition` isn't `RsExpr`.

dimonchik0036:
@ortem thanks!

At the moment this case doesn't work, because the value for `y` isn't evaluated.
```rust
 let y = if x == 0 { x / 0 }; // false negative
```

PR #3303, opened 2019-01-28
RES: duplicated `#[macro_export]` macros now unresolved

vlad20012:
We had a problem that if we have multiple `#[macro_export]`
with the same name we resolved to some of them randomly
(really randomly!). Ideally it should be multiresolve.


vlad20012:
bors retry

PR #3319, opened 2019-02-01
GRD: add optional `-Xmx` argument for runIde task

vlad20012:


build.gradle.kts
```diff
@@ -252,6 +253,12 @@ project(":") {
     tasks.withType<Jar> {
         dependsOn(copyXmls)
     }
+
+    tasks.withType<RunIdeTask> {
+        jvmArgs("-Xmx768m") // Default value for IDEA installation
```
amakeev: Mind, that CLion by default uses custom heap and stack sizes. Something like `-Xss2m -Xms256m -Xmx2000m`

PR #3330, opened 2019-02-02
INSP: make auto import works for path with type params

### LINKED ISSUE #3269: No autoimport for path with generics
```rust
let a = HashMap::<isize, isize>::new();
```

Undin:
The main reason why it didn't work before is wrong intermediate path creation.
`AutoImportFix#canBeResolvedToSuitableItem` didn't take into account namespace of path
and always tried to create intermediate path via `fn foo(t: $pathText) {}`.
But for paths like `Foo::<i32>::bar` it doesn't work because of syntax error.

Now, it's possible to pass namespace to `RsPsiFactory#tryCreatePath` to correctly create necessary path element

Fixes #3269

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -90,11 +90,11 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
                 return Context(basePath, emptyList())
             }
 
-            val candidates = getImportCandidates(path.toImportContext(project), basePath.referenceName, path.text) {
+            val candidates = getImportCandidates(ImportContext.from(project, path), basePath.referenceName, path.text) {
                 path != basePath || !(it.item is RsMod || it.item is RsModDeclItem || it.item.parent is RsMembers)
-            }
+            }.toList()
 
-            return Context(basePath, candidates.toList())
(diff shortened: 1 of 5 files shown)


Undin:
@mchernyavsky fixed

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -492,25 +491,34 @@ private fun RsMod.insertUseItem(psiFactory: RsPsiFactory, usePath: String) {
...
+            ns = path.pathNamespace,
+            attributes = path.stdlibAttributes,
+            namespaceFilter = path.namespaceFilter
+        )
+    }
+}
 
+private val RsPath.pathNamespace: RsPsiFactory.PathNamespace get() = when (context) {
+    is RsPathExpr,
+    is RsStructLiteral -> RsPsiFactory.PathNamespace.VALUES
```
mchernyavsky: And also `RsPatStruct` and `RsPatTupleStruct`:
```rust
//use crate::qwe::{S, TS};

mod qwe {
    pub struct S<T> { pub x: T }
    pub struct TS<T>(pub T);
}

fn main() {
    let x = S::<i32> { x: 42 };
    let S::/*caret*/ <i32> { x } = x;

    let x = TS::<i32>(42);
    let TS::/*caret*/ <i32>(x) = x;
}
```

Issue #1541, opened 2017-07-30
Cargo check indicator


zummenix:
It's not clear whether `cargo check` is in progress. I often find myself typing extra space somewhere to force rechecking.

Is it possible to show some kind of indication that the command is running?

PR #3332, opened 2019-02-03
ANN: Add indicator for Cargo Check

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/1541.


src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -122,9 +127,26 @@ private fun checkProject(info: CargoCheckAnnotationInfo): CargoCheckAnnotationRe
...
+        override fun onThrowable(error: Throwable) {
+            future.completeExceptionally(error)
+        }
+    }.queue()
+
     val output = try {
-        info.toolchain.cargoOrWrapper(info.projectPath).checkProject(info.module.project, info.module, info.projectPath)
-    } catch (e: ExecutionException) {
+        future.join()
+    } catch (e: CancellationException) {
```
vlad20012: Exceptions should be re-throwed?

PR #3338, opened 2019-02-04
ACT: Allow a user to install Clippy from IDE

mchernyavsky:


src/main/kotlin/org/rust/ide/actions/RustfmtFileAction.kt
```diff
@@ -55,7 +50,7 @@ class RustfmtFileAction : DumbAwareAction() {
             // #1131 - Check if we get a "'rustfmt' is not installed" and let the user know to install fmt
             if ("'rustfmt' is not installed" in message) {
                 val projectDir = project.cargoProjects.findProjectForFile(file)?.manifest?.parent
-                val action = if (projectDir != null) InstallRustfmtAction(projectDir) else null
+                val action = if (projectDir != null) InstallComponentAction(projectDir, "rustfmt") else null
```
Undin: Let's use `rustfmt-preview` here to support old toolchains. I was surprised but `rustup add component rustfmt-preview` works correctly with the latest `rustup` version

src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
```diff
@@ -5,17 +5,35 @@
...
 class RunClippyAction : RunCargoCommandActionBase(CargoIcons.CLIPPY) {
-
     override fun actionPerformed(e: AnActionEvent) {
         val project = e.project ?: return
         val cargoProject = getAppropriateCargoProject(e.dataContext) ?: return
+        val projectDirectory = cargoProject.manifest.parent ?: return
+
+        val isClippyInstalled = run {
+            val rustup = project.toolchain?.rustup(projectDirectory) ?: return@run true
+            val (_, isClippyInstalled) = rustup.listComponents()
```
Undin: I think we should start external process under progress

src/main/kotlin/org/rust/ide/actions/RustfmtFileAction.kt
mchernyavsky: Fixed. Also have replaced `clippy` with `clippy-preview`.

src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
```diff
@@ -5,17 +5,38 @@
...
+import org.rust.ide.notifications.showBalloon
 
 class RunClippyAction : RunCargoCommandActionBase(CargoIcons.CLIPPY) {
-
     override fun actionPerformed(e: AnActionEvent) {
         val project = e.project ?: return
         val cargoProject = getAppropriateCargoProject(e.dataContext) ?: return
+        val projectDirectory = cargoProject.manifest.parent ?: return
+
+        val isClippyInstalled = ProgressManager.getInstance().runProcess(Computable {
```
Undin: But `ProgressManager#runProcess ` doesn't solve the initial problem. It runs a new process from the current thread, in our case in EDT. Let's better use `ProgressManager#runProcessWithProgressSynchronously`. Yes, it can show progress dialog, but as I understand it doesn't do it when execution is short

Undin: Maybe `needInstallClippy` is better here? Looks like it's pretty tricky to check case when a user doesn't have `rustup` and `clippy` at the same time. And at this moment `isClippyInstalled` is `true` in this case to avoid unexpected notification with a suggestion to install `clippy` via `rustup`. So, this name is rather misleading

src/main/kotlin/org/rust/cargo/runconfig/command/RunClippyAction.kt
```diff
@@ -5,17 +5,38 @@
...
+            val rustup = project.toolchain?.rustup(projectDirectory) ?: return@Computable true
```
Undin: Let's add some comment that we don't handle (and probably don't want) cases when a user doesn't have `rustup` or has too old `rustup` (without `clippy` component)

mchernyavsky:
@Undin fixed.

PR #3346, opened 2019-02-04
ANN: Save all documents before `cargo check`

mchernyavsky:
Depends on https://github.com/intellij-rust/intellij-rust/pull/3332.


src/main/kotlin/org/rust/ide/annotator/RsCargoCheckAnnotator.kt
```diff
@@ -111,15 +109,7 @@ class RsCargoCheckAnnotator : ExternalAnnotator<CargoCheckAnnotationInfo, CargoC
 // NB: executed asynchronously off EDT, so care must be taken not to access disposed objects
 private fun checkProject(info: CargoCheckAnnotationInfo): CargoCheckAnnotationResult? {
     val indicator = WriteAction.computeAndWait<ProgressIndicator, Throwable> {
-        // We have to save the file to disk to give cargo a chance to check fresh file content.
```
vlad20012: Don't remove the comment!

PR #3352, opened 2019-02-06
MACRO: expand macros in stmt context

vlad20012:
Note: this is not about name resolution or type inference. We can only expand (e.g. show expansion) such macros.

Now we can expand statement-like macros. E.g.
```rust
macro_rules! foo {
    ($ i:ident, $ j:ident) => {
        struct $ i;
        let $ j = 0;
        ($ i, $ j)
    }
}
fn main() {
    foo!(S, a);
}
```
I've done it without touching the grammar, but I don't sure in it, may be it's better to create MacroStmt or a something.

Also introduced class MacroExpansion that abstracts macro expansion made in different contexts.

This is also a preparation for #3015

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -20,7 +20,6 @@ import com.intellij.psi.impl.source.DummyHolderFactory
 import com.intellij.psi.tree.TokenSet
 import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
-import com.intellij.psi.util.PsiModificationTracker
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.lang.RsLanguage
 import org.rust.lang.core.parser.RustParser
@@ -101,7 +100,7 @@ private val STD_MACRO_WHITELIST = setOf("write", "writeln")
 class MacroExpander(val project: Project) {
     private val psiFactory = RsPsiFactory(project)
 
-    fun expandMacro(def: RsMacro, call: RsMacroCall): List<RsExpandedElement>? {
+    fun expandMacro(def: RsMacro, call: RsMacroCall): MacroExpansion? {
         // All std macros contain the only `impl`s which are not supported for now, so ignoring them
         if (def.containingCargoTarget?.pkg?.origin == PackageOrigin.STDLIB && def.name !in STD_MACRO_WHITELIST) {
             return null
@@ -109,7 +108,7 @@ class MacroExpander(val project: Project) {
 
         val expandedText = expandMacroAsText(def, call) ?: return null
 
-        return psiFactory.parseExpandedTextWithContext(call, expandedText)
+        return parseExpandedTextWithContext(call.expansionContext, psiFactory, expandedText)
     }
 
     private fun expandMacroAsText(def: RsMacro, call: RsMacroCall): CharSequence? {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -20,7 +20,6 @@ import com.intellij.psi.impl.source.DummyHolderFactory
 import com.intellij.psi.tree.TokenSet
```
mchernyavsky: `RsPsiFactory.parseExpandedTextWithContext` is never used now.

src/main/kotlin/org/rust/lang/core/macros/MacroExpansion.kt
```diff
@@ -8,37 +8,122 @@ package org.rust.lang.core.macros
...
             CachedValueProvider.Result.create(result, project.rustStructureModificationTracker)
         }
     }
 }
+
+enum class MacroExpansionContext {
+    EXPR, PAT, TYPE, STMT, ITEM
+}
+
+val RsMacroCall.expansionContext: MacroExpansionContext
```
mchernyavsky:
```kotlin
val RsMacroCall.expansionContext: MacroExpansionContext
    get() = when (val context = context) {
        is RsMacroExpr -> when {
            context.context is RsExprStmt -> MacroExpansionContext.STMT
            else -> MacroExpansionContext.EXPR
        }
        is RsPatMacro -> MacroExpansionContext.PAT
        is RsMacroType -> MacroExpansionContext.TYPE
        else -> MacroExpansionContext.ITEM
    }
```

src/main/kotlin/org/rust/lang/core/macros/MacroExpansion.kt
```diff
@@ -8,37 +8,122 @@ package org.rust.lang.core.macros
...
 
-private val NULL_RESULT: CachedValueProvider.Result<List<RsExpandedElement>?> =
+private val NULL_RESULT: CachedValueProvider.Result<MacroExpansion?> =
     CachedValueProvider.Result.create(null, PsiModificationTracker.MODIFICATION_COUNT)
 
-fun expandMacro(call: RsMacroCall): CachedValueProvider.Result<List<RsExpandedElement>?> {
+fun expandMacro(call: RsMacroCall): CachedValueProvider.Result<MacroExpansion?> {
     val context = call.context as? RsElement ?: return NULL_RESULT
     return when {
         call.macroName == "lazy_static" -> {
```
mchernyavsky: Let's cover this branch with test... and replace `when` with `if`.

vlad20012: It is covered by RsMacroExpansionResolveTest

vlad20012:
Done

src/main/kotlin/org/rust/lang/core/macros/MacroExpansion.kt
```diff
@@ -8,37 +8,126 @@ package org.rust.lang.core.macros
...
+import org.rust.lang.core.psi.ext.descendantOfTypeStrict
 import org.rust.lang.core.psi.ext.macroName
-import org.rust.lang.core.psi.rustStructureModificationTracker
 
-private val NULL_RESULT: CachedValueProvider.Result<List<RsExpandedElement>?> =
+private val NULL_RESULT: CachedValueProvider.Result<MacroExpansion?> =
     CachedValueProvider.Result.create(null, PsiModificationTracker.MODIFICATION_COUNT)
 
-fun expandMacro(call: RsMacroCall): CachedValueProvider.Result<List<RsExpandedElement>?> {
+fun expandMacro(call: RsMacroCall): CachedValueProvider.Result<MacroExpansion?> {
```
mchernyavsky: Maybe convert it into an extension function of `RsMacroCall` not to provide additional name into global scope? Same comment about other public functions in `MacroExpansion.kt` and `MacroExpansionViewUtils.kt`.

vlad20012: This will be highly refactored with #3015, so I don't think it is not necessary

PR #3359, opened 2019-02-07
PERF: use original element during completion if possible

vlad20012:


mchernyavsky:
```rust
fn main() {
    let/*caret*/
}
```

Stacktrace:
```
java.lang.ClassCastException: org.rust.lang.core.psi.impl.RsLetDeclImpl cannot be cast to org.rust.lang.core.psi.ext.RsReferenceElement
	at org.rust.lang.core.completion.RsCommonCompletionProvider.addCompletions(RsCommonCompletionProvider.kt:45)
	at com.intellij.codeInsight.completion.CompletionProvider.addCompletionVariants(CompletionProvider.java:36)
	at com.intellij.codeInsight.completion.CompletionContributor.fillCompletionVariants(CompletionContributor.java:141)
	at com.intellij.codeInsight.completion.CompletionService.getVariantsFromContributors(CompletionService.java:87)
        ...
```

src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
```diff
@@ -39,8 +39,11 @@ object RsCommonCompletionProvider : CompletionProvider<CompletionParameters>() {
         context: ProcessingContext,
         result: CompletionResultSet
     ) {
-        val element = parameters.position.parent as RsReferenceElement
-        if (parameters.position !== element.referenceNameElement) return
+        // Use original position if possible to re-use caches of the real file
+        val position = CompletionUtil.getOriginalOrSelf(parameters.position)
```
Undin: Looks like we should accurately handle cases when `parameters.position` and `position` have different token types to avoid exceptions described [here](https://github.com/intellij-rust/intellij-rust/pull/3359#issuecomment-461341968) and still provide correct completion
I think it's worth to add test(s) for such case to commit right behavior

vlad20012: Positions can't have different types b/c they are always leaf elements (and the same type is guaranteed by `CompletionUtil.getOriginalOrSelf`), BUT they can be children of different types. And [here](https://github.com/intellij-rust/intellij-rust/pull/3359#issuecomment-461341968) we failed on the parent element cast. Maybe we should rather check that all ancestors of both elements have the same type?

Undin: > Positions can't have different types b/c they are always leaf elements

Yes, but they can have different **token types**, for example, `identifier` and `let` keyword => psi structures of original file and completion copy one differ

> Maybe we should rather check that all ancestors of both elements have the same type?

I think it's enough to just check that parents have the same type or that `CompletionUtil#getOriginalOrSelf` and `parameters.position` have the same token type (looks like these conditions always have the same value)

src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
```diff
@@ -39,8 +39,12 @@ object RsCommonCompletionProvider : CompletionProvider<CompletionParameters>() {
         context: ProcessingContext,
         result: CompletionResultSet
     ) {
-        val element = parameters.position.parent as RsReferenceElement
-        if (parameters.position !== element.referenceNameElement) return
+        // Use original position if possible to re-use caches of the real file
+        val position = CompletionUtil.getOriginalElement(parameters.position)
+            ?.takeIf { isAncestorTypesEquals(it, parameters.position) }
```
Undin: I still think it's enough to check only parents instead of all ancestors but looks like it doesn't affect completion performance so it's ok

PR #3364, opened 2019-02-08
COMP: add `dbg` postfix completion

Undin:

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
@@ -82,5 +82,13 @@ class IterPostfixTemplate(provider: RsPostfixTemplateProvider) :
     override fun getElementToRemove(expr: PsiElement): PsiElement = expr
 }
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -82,5 +82,13 @@ class IterPostfixTemplate(provider: RsPostfixTemplateProvider) :
     override fun getElementToRemove(expr: PsiElement): PsiElement = expr
 }
 
+class DbgPostfixTemplate(provider: RsPostfixTemplateProvider) :
+    StringBasedPostfixTemplate("dbg", "dbg!(expr)", RsTopMostInScopeSelector(), provider) {
+
+    override fun getTemplateString(element: PsiElement): String? = "dbg!(${element.text})"
```
ortem:
```suggestion
    override fun getTemplateString(element: PsiElement): String = "dbg!(${element.text})"
```

Undin: Fixed

PR #3365, opened 2019-02-08
TY: more precise trait pick in type inference

### LINKED ISSUE #3344: Duplicate trait available for import
```rust
use std::path::PathBuf;
mod a {
    use std::path::Path;
    use std::path::PathBuf;
    pub trait X {
        fn do_x(&self);
    }
    impl<T: AsRef<Path>> X for T {
        fn do_x(&self) {}
    }
}
fn main() {
    let p = PathBuf::new();
    p.do_x();
}
```
When I press Alt+Enter on `do_x` in `main`, the IDE suggests two imports, but they refer to the same trait (`X`). This is probably caused by the fact that `p` is `PathBuf` and can be coerced to `&Path` and both of these types implement the trait. If I replace the trait impl with `impl X for PathBuf { ... }`, the duplicate disappears. The expected behavior is to suggest each trait only once, so a disambiguation menu shouldn't be needed in this case.

Undin:
In case, when there isn't any trait in scope, we add all potential variants as resolve variants. But sometimes we can filter out some obviously wrong variants. 

These changes don't affect type inference/name resolution for correct code but should improve name resolution/auto import for incorrect code (when all traits are out of scope)

Fixes #3344

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -1212,7 +1212,12 @@ class RsFnInferenceContext(
             traitToCallee.keys.filterInScope(methodCall).forEach {
                 filtered += traitToCallee.getValue(it)
             }
-            filtered
+            if (filtered.isNotEmpty()) {
+                filtered
+            } else {
+                TypeInferenceMarks.methodPickTraitsOutOfScope.hit()
+                list
+            }
         }.singleOrFilter { callee ->
             // 2. Filter methods by trait bounds (try to select all obligations for each impl)
             TypeInferenceMarks.methodPickCheckBounds.hit()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1212,7 +1213,7 @@ class RsFnInferenceContext(
             traitToCallee.keys.filterInScope(methodCall).forEach {
                 filtered += traitToCallee.getValue(it)
             }
-            filtered
+            if (TypeInferenceMarks.methodPickTraitsOutOfScope.hitOnFalse(filtered.isNotEmpty())) filtered else list
```
vlad20012: Let's use simple if/else. This pretty simple condition starts to look very complex with this mark ^)

Undin: Done

PR #3369, opened 2019-02-09
T: fix completion performance test

vlad20012:
Now it works at least

### DIFF
--- a/src/test/kotlin/org/rustSlowTests/RsHighlightingPerformanceTest.kt
+++ b/src/test/kotlin/org/rustSlowTests/RsHighlightingPerformanceTest.kt
@@ -94,12 +94,13 @@ class RsHighlightingPerformanceTest : RsWithToolchainTestBase() {
             .mapNotNull { it.block?.stmtList?.lastOrNull() }
             .forEach { stmt ->
                 myFixture.editor.caretModel.moveToOffset(stmt.textOffset)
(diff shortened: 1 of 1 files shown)


src/test/kotlin/org/rustSlowTests/RsHighlightingPerformanceTest.kt
```diff
@@ -89,14 +90,24 @@ class RsHighlightingPerformanceTest : RsWithToolchainTestBase() {
...
 
+        val caretsToAdd = mutableListOf<VisualPosition>()
+
         myFixture.file.descendantsOfType<RsFunction>()
             .asSequence()
             .mapNotNull { it.block?.stmtList?.lastOrNull() }
             .forEach { stmt ->
                 myFixture.editor.caretModel.moveToOffset(stmt.textOffset)
-                myFixture.type("a<caret>;")
+                myFixture.type("aaa;")
```
mchernyavsky: Let's replace "aaa;" with "Hash;" to know exactly what variants will be processed.

And also, we can test "a;" - for this you need to increase the limit of completion variants. This can be done inside `repeatTest` with this function:
```kotlin
private fun setCompletionVariantLimit(value: Int) =
    Registry.get("ide.completion.variant.limit").setValue(value, testRootDisposable)
```

PR #3373, opened 2019-02-09
COMP: support completion from index for enums

### LINKED ISSUE #3329: Non-imported enums are not suggested in autocompletion list
Follow-up for #2721. Importing structs works like a charm now, but importing enums is not possible this way. For example:
```rust
mod a {
    pub struct MyAbc;
    pub enum MyDef {
        MyDefVar1,
        MyDefVar2,
    }
}
mod c {
    use crate::a::MyDef::MyDefVar1;

    fn x() {
        [cursor here]

    }
}
```
At the cursor position, the plugin suggests `MyDefVar1` and `MyDefVar2`, so I can import the variant by choosing the suggested item. But most often I want to import the enum itself, and `MyDef` is absent in the suggestion list. `MyAbc` is suggested and can be imported properly.

Undin:
Fixes #3329

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -101,7 +101,7 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
                 return Context(basePath, emptyList())
             }
 
-            val candidates = getImportCandidates(ImportContext.from(project, path), basePath.referenceName, path.text) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -397,6 +397,7 @@ private val RsPath.namespaceFilter: (RsQualifiedNamedElement) -> Boolean get() =
     }
     is RsPathExpr -> { e ->
         when (e) {
+            is RsEnumItem -> isCompletion
```
Undin: Not sure that it's only one place where we should do such change for completion but can't come up with other cases

PR #3375, opened 2019-02-10
INSP: Trait implementation inspection

### LINKED ISSUE #3325: "Implement members" intention doesn't trigger inside the impl block
When I create a trait impl, I have this situation:
```rust
impl Default for A {
     [cursor here]
}
```
The plugin suggests to implement members but I have to move cursor up to `impl Default for A` to use the suggestion. It would be much more convenient to trigger it from where the cursor is. Note that the suggestion doesn't appear until I type `impl Default for A {` and press Enter, so I *always* have to go back.

Undin:
Extract some error annotations about trait implementation into separate inspection

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsTraitImplementationInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsTraitImplementationInspection.kt
@@ -1,0 +1,71 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElementVisitor
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.TraitImplementationInfo
+import org.rust.lang.core.psi.ext.resolveToTrait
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/inspections/RsTraitImplementationInspection.kt
```diff
@@ -0,0 +1,77 @@
...
+
+            val implInfo = TraitImplementationInfo.create(trait, impl) ?: return
+
+            if (implInfo.missingImplementations.isNotEmpty()) {
+                val missing = implInfo.missingImplementations.mapNotNull { it.name?.let { "`$it`" } }.joinToString(", ")
+                val endElement = impl.typeReference ?: impl.impl
+                RsDiagnostic.TraitItemsMissingImplError(impl.impl, endElement, missing, impl, Severity.ERROR)
+                    .addToHolder(holder)
+                endElement.nextSibling?.also {
+                    RsDiagnostic.TraitItemsMissingImplError(it, impl.lastChild, missing, impl, Severity.INFO)
```
Undin: Interesting thing: if you change severity of this inspection via settings, this annotation won't change it

vlad20012:
I have 2 similar fixes if place the caret after the type:
image

Also I don't sure that I want to see this fix inside all descendants if the impl. Maybe just in direct children of RsMembres?
image

Undin:
@vlad20012 I've dropped changes related to additional annotation

Undin:
bors retry

PR #3378, opened 2019-02-10
ANN: Non't show E0583 "File not found for module" if syntax error in declaration

vlad20012:
When typing `mod foo {}` this construction can be incomplete and
so will be parsed as module declaration `mod foo` with missed
semicolon. In this case it tries to find "foo.rs" file and
shows E0583 "File not found for module" because it doesn't exist.

I think we should not show this error if semicolon is missed.
We should let user to fix the syntax error or complete inline
module declaration first.


Undin:
Cool!
bors r+

Issue #2080, opened 2017-11-27
Add ability to pass arguments to cargo check for check run config/annotator


mkaput:
Use case:

1. I have cross-compiled project using Xargo, which I compile to custom target `x86_64-foo`.
2. I have per-target dependencies set up in `Xargo.toml`.
3. I need to pass `--target x86_64-foo` to arguments of `xargo check` for the command to pass.
  *Invoking just `project/$ xargo check` won't execute as code doesn't compile on my host arch.*

IMHO the best way to solve this is to add text box with custom arguments for cargo check here:
![obraz](https://user-images.githubusercontent.com/3450050/33281993-75b4adf0-d3a7-11e7-8a85-df9e0cda067f.png)

Or if one could pass custom run configuration here (though this should be rather platform feature).

matklad:
Hm, I wonder if we should tie this into run confgurations framework somehow...

bgourlie:
This would also be nice for WASM, being able to specify `--target=wasm32-unknown-unknown`.  See https://github.com/koute/stdweb/issues/106.

PR #3380, opened 2019-02-11
ANN: Ability to pass arguments to cargo check annotator

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/2080.

image


src/main/kotlin/org/rust/cargo/util/CargoCommandLineEditor.kt
```diff
@@ -52,12 +55,16 @@ class CargoCommandLineEditor(
         label.labelFor = textField
     }
 
-    private fun createTextField(value: String): TextFieldWithCompletion =
+    private fun createTextField(value: String): TextFieldWithCompletion = runReadAction {
```
mchernyavsky: I'm not sure about this `runReadAction`.

PR #3382, opened 2019-02-11
ACT: add an action to enable/disable cargo check annotator

vlad20012:
(made similar to ToggleInlineHintsAction action)

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/ToggleCargoCheckAction.kt
+++ b/src/main/kotlin/org/rust/ide/actions/ToggleCargoCheckAction.kt
@@ -1,0 +1,37 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.actions
+
+import com.intellij.openapi.actionSystem.AnAction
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/actions/ToggleCargoCheckAction.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    override fun update(e: AnActionEvent) {
+        val project = e.project
+        if (project == null) {
+            e.presentation.isEnabledAndVisible = false
+            return
+        }
+
+        val isHintsShownNow = project.rustSettings.useCargoCheckAnnotator
+        e.presentation.text = if (isHintsShownNow) disableText else enableText
```
Undin: Are we sure about dynamic presentation text?
At this moment it's rather hard to find the action via `find action` action because it can be found by "Disable Cargo Check" or "Enable Cargo Check". At the same time when I type "Toggle Cargo check" (or part of it) in the corresponding dialog, I expect to find the same text in suggested items but because of different dynamic presentation text I fail to do it
image
Maybe it's better add toggle button as some other actions do?

PR #3385, opened 2019-02-11
ANN: add quickfix to create lifetime parameter from usage

### LINKED ISSUE #3183: Add fix to create lifetime from usage

Undin:
Fixed version of #3187

Closes #3183
Closes #3187


Undin:
bors r=vlad20012

PR #3387, opened 2019-02-12
Fix extra line markers next to syntax errors

### LINKED ISSUE #3386: A lot of wrong line markers next to syntax errors
image

```rust
fn foo bar<T>(t: T) {}
#[test] // test function is necessary to reproduce
fn has_icon() { assert(true) }
```

vlad20012:
Fixes #3386


mchernyavsky:
bors retry

PR #3394, opened 2019-02-13
TY: Fix type inference in for/while let/if let exprs with complex patterns

vlad20012:


src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1414,7 +1424,16 @@ class RsFnInferenceContext(
...
+        val condition = expr.condition
+        if (condition != null) {
+            val pat = condition.pat
+            if (pat != null) {
+                // if let Some(a) = ... {}
+                pat.extractBindings(resolveTypeVarsWithObligations(condition.expr.inferType()))
+            } else {
+                condition.expr.inferType(TyBool)
+            }
+        }
```
Undin: Hm, this piece of code is exactly the same as in `inferWhileExprType`
Maybe extract it into a separate method?

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1414,7 +1424,16 @@ class RsFnInferenceContext(
         TyReference(expr.inferType((expected as? TyReference)?.referenced), mutable) // TODO infer the actual lifetime
 
     private fun inferIfExprType(expr: RsIfExpr, expected: Ty?): Ty {
-        expr.condition?.let { it.pat?.extractBindings(it.expr.inferType()) ?: it.expr.inferType(TyBool) }
+        val condition = expr.condition
+        if (condition != null) {
+            val pat = condition.pat
+            if (pat != null) {
+                // if let Some(a) = ... {}
+                pat.extractBindings(resolveTypeVarsWithObligations(condition.expr.inferType()))
```
Undin: Do we need to pass expected type into `inferType()` call?

Undin: Oh, sorry, we shouldn't because we don't know this type

Issue #3129, opened 2018-12-08
False-positive "Call to unsafe function requires unsafe block"


vlad20012:
![image](https://user-images.githubusercontent.com/3221931/49679753-c7e12d00-fa9e-11e8-8faf-c264d556beca.png)
Currently we think that functions in any `extern` block are unsafe. It isn't truth. For example, functions inside `extern "rust-intrinsic" {}` block are safe. Also function inside extern block followed with `#[wasm-bindgen]` proc macro are safe too (but it is rather procedural macro magic).

jsmnbom:
Is there any way to turn off this inspection in the meantime?

ortem:
@jsmnbom try [this](https://www.jetbrains.com/help/idea/disabling-and-enabling-inspections.html)

vlad20012:
@ortem it's annotator, not an inspection. It isn't possible to disable it

vlad20012:
I found that most of functions inside `extern "rust-intrinsic"` blocks are unsafe, but there are some that are safe =/ I guess it is hardcoded somewhere in the compiler

MSxDOS:
Any way to disable this annotation? Getting swamped with false positives when actively using `likely`\ `unlikely` intrinsics.

DzenanJupic:
`core::intrinsics::unlikely` is another example:

![image](https://user-images.githubusercontent.com/56133904/126049812-d5298ab5-f3ea-4b8a-b3c8-2c63a183a543.png)

Kxuan:
> I found that most of functions inside `extern "rust-intrinsic"` blocks are unsafe, but there are some that are safe =/ I guess it is hardcoded somewhere in the compiler

This is the hardcoeded list. These names should be considered as "safe".

Link: [pub fn intrinsic_operation_unsafety(intrinsic: Symbol) -> hir::Unsafety {](https://github.com/rust-lang/rust/blob/fc323035ac49b8d834fc2a0f580af8fd45e90b30/compiler/rustc_typeck/src/check/intrinsic.rs#L66)

insinfo:
It may be off topic, but does anyone know how to disable the hightlight of unsafe function calls in RustRover?

![image](https://github.com/intellij-rust/intellij-rust/assets/12227024/60c87481-2521-4b03-9881-4d329bd66266)

PR #3395, opened 2019-02-13
ANN: don't treat fns in wasm_bindgen extern block as unsafe

vlad20012:
Part of #3129

Also it will be useful if we will start to expand macros in
extern blocks (macros_in_extern)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
@@ -80,6 +80,27 @@ val RsFunction.returnType: Ty get() {
 
 val RsFunction.abi: RsExternAbi? get() = externAbi ?: (parent as? RsForeignModItem)?.externAbi
 
+/**
+ * A function is unsafe if defined with `unsafe` modifier or if defined inside a certain `extern`
+ * block. But [RsFunction.isUnsafe] takes into account only `unsafe` modifier. [isActuallyUnsafe]
+ * takes into account both cases.
+ */
+val RsFunction.isActuallyUnsafe: Boolean
+    get() {
+        if (isUnsafe) return true
+        val context = context
+        return if (context is RsForeignModItem) {
+            // functions inside `extern` block are unsafe in most cases
+            //
+            // #[wasm_bindgen] is a procedural macro that removes the following
+            // extern block, so all function inside it become safe.
+            // See https://github.com/rustwasm/wasm-bindgen
+            !context.queryAttributes.hasAttribute("wasm_bindgen")
+        } else {
+            false
+        }
+    }
+
 abstract class RsFunctionImplMixin : RsStubbedNamedElementImpl<RsFunctionStub>, RsFunction, RsModificationTrackerOwner {
 
     constructor(node: ASTNode) : super(node)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -77,6 +77,13 @@ val RsFunction.returnType: Ty get() {
 
 val RsFunction.abi: RsExternAbi? get() = externAbi ?: (parent as? RsForeignModItem)?.externAbi
 
+val RsFunction.isActuallyUnsafe: Boolean
```
vlad20012: I'm not sure about this name

src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -241,7 +241,7 @@ private val RsItemElement.declarationModifiers: List<String> get() {
             if (isConst) {
                 modifiers += "const"
             }
-            if (isUnsafe) {
+            if (isActuallyUnsafe) {
```
Undin: Oh, it's a bug that we used `isUnsafe` in the old meaning because we should add `unsafe` modifier only in a case when it is written in a function declaration. 
So, let's keep old code here and fix related tests

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
Undin: I can't come up with a better name, so let's add javadocs for this property with some details when it's not the same with `RsFunction.isUnsafe`

Undin: Should we introduce the same property for constants?

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -77,6 +77,13 @@ val RsFunction.returnType: Ty get() {
...
+    get() {
+        if (isUnsafe) return true
+        val extern = context as? RsForeignModItem ?: return false
+        return !extern.queryAttributes.hasAttribute("wasm_bindgen")
```
Undin: Let's add some comment with a link to the corresponding crate and short explanation why functions inside extern block with `wasm_bindgen` are not `unsafe`

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
vlad20012: > Should we introduce the same property for constants?

Constants can't be unsafe

Undin: Yeah, we can't use const in extent blocks but we can use static variables
```rust
extern {
    static foo: i32;
}
```
And psi element for constant and static variables is the same - `RsConstant`
Is it possible that using of static variable is safe in some cases?

vlad20012: Mmm, let's make it in another PR.

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -80,6 +80,27 @@ val RsFunction.returnType: Ty get() {
 
 val RsFunction.abi: RsExternAbi? get() = externAbi ?: (parent as? RsForeignModItem)?.externAbi
 
+/**
+ * A function is unsafe if defined with `unsafe` modifier or if defined inside `extern` block.
```
Undin: It's a little bit misleading because not all function from extern block are unsafe

vlad20012:
bors retry

PR #3398, opened 2019-02-13
RUN: Apply `Offline mode` option not only to `cargo build` but also to `cargo check`

mchernyavsky:


src/test/kotlin/org/rust/cargo/project/RustProjectSettingsServiceTest.kt
```diff
@@ -31,7 +31,7 @@ class RustProjectSettingsServiceTest : LightPlatformTestCase() {
               <option name="toolchainHomeDirectory" value="/" />
               <option name="useCargoCheckAnnotator" value="true" />
               <option name="useCargoCheckForBuild" value="false" />
-              <option name="useOfflineForCargoCheck" value="true" />
+              <option name="useOffline" value="true" />
```
kumbayo: @mchernyavsky I assume renaming option will cause its setting to be lost on update?
Is this OK/expected or should we try to stay compatible with previous settings?

mchernyavsky: I think it's OK since this option is not widely used.
And before this commit, `useOfflineForCargoCheck` had no effect on `cargo check` :)

Issue #2754, opened 2018-08-08
Add "Invert if condition" intention


matklad:
The intention should negate the condition and swap the if's branches. 

Here's a similar intentino in Kotlin: https://github.com/JetBrains/kotlin/blob/efec82c0ebcf6ff26b88f6e411785c3df5510f7e/idea/src/org/jetbrains/kotlin/idea/intentions/InvertIfConditionIntention.kt. We don't need to implement **all** the features in one go, but it might be a good idea to look at it to see what's possible.

Here's a related intention in IntelliJ Rust:
https://github.com/intellij-rust/intellij-rust/blob/3fe98015358240e48a45150554875903d93e1256/src/main/kotlin/org/rust/ide/intentions/AddElseIntention.kt
https://github.com/intellij-rust/intellij-rust/blob/3fe98015358240e48a45150554875903d93e1256/src/test/kotlin/org/rust/ide/intentions/AddElseIntentionTest.kt

zinoviy23:
I want to try implement this feature

PR #3401, opened 2019-02-14
INT: Invert If Intention

zinoviy23:
Implementing #2754

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt
@@ -1,0 +1,53 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.utils.negate
+
+class InvertIfIntention : RsElementBaseIntentionAction<InvertIfIntention.Context>() {
+    data class Context(
+        val ifExpr: RsIfExpr,
+        val condition: RsCondition,
+        val thenBlock: RsBlock,
+        val elseBlock: RsBlock
+    )
+
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val ifExpr = element.ancestorStrict<RsIfExpr>() ?: return null
+        val condition = getSuitableCondition(ifExpr) ?: return null
+        val thenBlock = ifExpr.block ?: return null
+        val elseBlock = ifExpr.elseBranch?.block ?: return null
+
+        return Context(ifExpr, condition, thenBlock, elseBlock)
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        val negatedCondition = ctx.condition.negate() as? RsExpr ?: return
+
+        val newIfExpr = RsPsiFactory(project).createIfElseExpression(negatedCondition, ctx.elseBlock, ctx.thenBlock)
+        val replacedIfExpr = ctx.ifExpr.replace(newIfExpr) as RsIfExpr
+
+        val newCondition = replacedIfExpr.condition!!.expr
+
+        if (newCondition is RsUnaryExpr && newCondition.excl != null) {
+            val binaryExpression = (newCondition.expr as? RsParenExpr)?.expr as? RsBinaryExpr ?: return
+            DemorgansLawIntention().invoke(project, editor, binaryExpression.binaryOp)
+        }
+    }
+
+    private fun getSuitableCondition(ifExpr: RsIfExpr): RsCondition? =
+        ifExpr.condition?.takeIf { it.let == null }
+
+    override fun getFamilyName(): String = text
+
+    override fun getText() = "Invert if condition"
+}
--- a/src/test/kotlin/org/rust/ide/intentions/InvertIfIntentionTest.kt
+++ b/src/test/kotlin/org/rust/ide/intentions/InvertIfIntentionTest.kt
@@ -1,0 +1,98 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+class InvertIfIntentionTest : RsIntentionTestBase(InvertIfIntention()) {
+    fun `test let not available`() = doUnavailableTest(""""
+        fn foo(a: Option<i32>) {
+            if let Some(x) /*caret*/= a {} else {}
+        }
+    """)
+
+    fun `test if without condition unavailable`() = doUnavailableTest("""
+        fn foo() { if /*caret*/ {} else {}}
+    """)
+
+    fun `test if without else branch unavailable`() = doUnavailableTest(""""
+        fn foo(a: i32) {
+            if a == 10 /*caret*/ {}
+        }
+    """)
+
+    fun `test if without then branch`() = doUnavailableTest("""
+        fn foo(a: i32) {if a == 10/*caret*/ else {}}
+    """)
+
+    fun `test simple inversion`() = doAvailableTest("""
+        fn foo() {
+            if 2 =/*caret*/= 2 {
+                Ok(())
+            } else {
+                Err(())
+            }
+        }
+    """, """
+        fn foo() {
+            if !(2 == 2) {
+                Err(())
+            } else {
+                Ok(())
+            }
+        }
+    """)
+
+    fun `test simple inversion on one line`() = doAvailableTest("""
+        fn foo() {
+            if 2 =/*caret*/= 2 { Ok(()) } else { Err(()) }
+        }
+    """, """
+        fn foo() {
+            if !(2 == 2) { Err(()) } else { Ok(()) }
+        }
+    """)
+
+    fun `test bigger condition to simplification`() = doAvailableTest("""
+        fn foo() {
+            if 2 =/*caret*/= 2 && 3 == 3 { Ok(()) } else { Err(()) }
+        }
+    """, """
+        fn foo() {
+            if 2 != 2 || 3 != 3 { Err(()) } else { Ok(()) }
+        }
+    """)
+
+    fun `test simple inversion strange formatting`() = doAvailableTest("""
+        fn foo() {
+            if 2 =/*caret*/= 2 {
+                Ok(())
+            } else { Err(()) }
+        }
+    """, """
+        fn foo() {
+            if !(2 == 2) { Err(()) } else {
+                Ok(())
+            }
+        }
+    """)
+
+    fun `test very simple condition`() = doAvailableTest("""
+        fn foo(cond: bool) {
+            if co/*caret*/nd {
+                Ok(())
+            } else {
+                Err(())
+            }
+        }
+    """, """
+        fn foo(cond: bool) {
+            if !(cond) {
+                Err(())
+            } else {
+                Ok(())
+            }
+        }
+    """)
+}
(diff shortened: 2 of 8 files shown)


vlad20012:
It looks like a very bad practice to call one intention from another. If you _really_ need to share some logic, it's better to extract it to some function and then call it from both intentions

zinoviy23:
Ooh, sorry. I took this solution from Kotlin Plugin. Ok, I'll fix it. And at which package can I place new file with logic?

vlad20012:
It does not matter much. E.g. somewhere in `org.rust.ide.utils` package. There is already `ExprUtils.kt` file

zinoviy23:
Any other comments or suggestions?

src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt
```diff
@@ -0,0 +1,53 @@
...
+class InvertIfIntention : RsElementBaseIntentionAction<RsIfExpr>() {
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsIfExpr? {
+        val ifStatement = element.ancestorStrict<RsIfExpr>() ?: return null
+
+        return if (ifStatement.elseBranch?.block != null &&
+            ifStatement.block != null &&
+            checkIsSuitableCondition(ifStatement)) ifStatement else null
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: RsIfExpr) {
```
ortem: IMO now it's better to call `DemorgansLawIntention` directly like this (note that `ifExpr` can be replaced before `DemorgansLawIntention` call):
```Kotlin
        val negatedCondition = ctx.condition.expr.negate() as? RsExpr ?: return

        val newIf = RsPsiFactory(project).createIfElseExpression(negatedCondition, ctx.elseBlock, ctx.thenBlock)
        ctx.ifExpr.replace(newIf)

        val demorgansLawIntention = DemorgansLawIntention()

        if (negatedCondition is RsUnaryExpr && negatedCondition.excl != null) {
            val binaryExpr = (negatedCondition.expr as? RsParenExpr)?.expr as? RsBinaryExpr ?: return
            val binaryOp = binaryExpr.binaryOp

            if (demorgansLawIntention.isAvailable(project, editor, binaryOp)) {
                demorgansLawIntention.invoke(
                    project,
                    editor,
                    DemorgansLawIntention.Context(binaryExpr, binaryOp.operatorType)
                )
            }
        }
    }
```
Currently `DemorgansLawIntention` is too complicated to be extracted: just look at `simplifyNegatedExpression`, `convertLeafExpression`, `convertConjunctionExpression` inside `ExprUtils.kt`. So it should be refactored and extracted in the future.

ortem:
@zinoviy23 Hi and welcome to IJ Rust contributors!

@vlad20012 Extracting De Morgan's law was a good idea, but now it's clear that this logic is not ready for extracting now. Such a solution works in Koltin, so it should work in our case too.

src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt
zinoviy23: Ooh. OK. But can I copypaste or something like this ```convertLeafExpression```? Or is it fine to leave ```!(2 == 2)``` not simplified?

ortem:
@zinoviy23 also, please add your name to `CONTRIBUTORS.txt`

zinoviy23:
In this branch?

src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt
ortem: Ah, you're right, it doesn't work in this case. The problem relates to `DemorgansLawIntention` itself, espessially with this line inside `findApplicableContext`:
`val binExpr = element.ancestorStrict<RsBinaryExpr>() ?: return null`

It searches strict ancestor with type `RsBinaryExpr`, so you have to pass something inside binary expr to the intention).

You can use code like this here (`isAvailable` call is redundant is you call `invoke(project: Project, editor: Editor, element: PsiElement)`):

```Kotlin
    val negatedCondition = ctx.condition.expr.negate() as? RsExpr ?: return

    val ifExpr = RsPsiFactory(project).createIfElseExpression(negatedCondition, ctx.elseBlock, ctx.thenBlock)
    val replacedIfExpr = ctx.ifExpr.replace(ifExpr) as RsIfExpr
    val newCondition = replacedIfExpr.condition!!.expr

    if (newCondition is RsUnaryExpr && newCondition.excl != null) {
        val binaryExpr = (newCondition.expr as? RsParenExpr)?.expr as? RsBinaryExpr ?: return
        DemorgansLawIntention().invoke(project, editor, binaryExpr.binaryOp)
    }
}
```

zinoviy23: Will it work with 2 == 2? As I understand, DemorganLaw looks for && or ||.

ortem: You're right, it works only for binary expressions with logical operators at top-level (like `2 == 2 && 3 == 3`, but not for `2 == 2`). And it checks inside `DemorgansLawIntention.findApplicableContext` when you call `invoke`.

zinoviy23: And do I leave !(2 == 2) without changes now?

ortem: Seems like it can be fixed via improving `PsiElement.negate()` (and the problem is not about your intention only, but affects other code insight related to boolean expressions). I think you can leave it as is for now and focus on the intention. I'll fix it later.

zinoviy23: Yes!. I found it in Kotlin plugin and want to write about it too.

src/test/kotlin/org/rust/ide/intentions/InvertIfIntentionTest.kt
```diff
@@ -0,0 +1,113 @@
...
+package org.rust.ide.intentions
+
+class InvertIfIntentionTest : RsIntentionTestBase(InvertIfIntention()) {
+    fun `test let not available`() = doUnavailableTest(""""
+        fn foo(a: Option<i32>) {
+            if let Some(x) /*caret*/= a {} else {}
+        }"""
+    )
+
+    fun `test if without condition unavailable`() = doUnavailableTest(
```
ortem: Please format all the tests like this:
```Kotlin
fun `test`() = doAvailableTest("""
    <code>
""", """
   <code>
""")
```
or
```Kotlin
fun `test`() = doUnavailableTest("""
    <code>
""")

Comment (zinoviy23) [reply src/test/kotlin/org/rust/ide/intentions/InvertIfIntentionTest.kt]: Ok!

File: src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt (hunk 1 above)
Comment (zinoviy23) [reply src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt]: There are no ```createIfElseExpression```, only IfExpression with one branch. Can I add it?

Comment (ortem) [reply src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt]: Yes, something like
```Kotlin
    fun createIfElseExpression(condition: RsExpr, thenBlock: RsBlock, elseBlock: RsBlock): RsIfExpr {
        val result = createExpressionOfType<RsIfExpr>("if ${condition.text} { () } else { () }")
        result.block!!.replace(thenBlock)
        result.elseBranch!!.block!!.replace(elseBlock)
        return result
    }
```

zinoviy23: Is there ```!!``` operator instead of ```?.``` for throwing exception if something is broken in ```createExpressionOfType``` function?

ortem: We have valid `condition`, so `createExpressionOfType<RsIfExpr>("if ${condition.text} { () } else { () }")` must be OK.

zinoviy23: So, it doesn't matter ```?.``` or ```!!``` in this case?

zinoviy23: I'm just learning Kotlin, so I think about difference of this operators in this case. In my opinion, it is a idealogic problem, based on what we want. If we want this function is working in all cases, also in case, when someone broke ```createExpressionOfType```. Or we want to throw exception, when someone broke this function.

zinoviy23: OK! I got it.

zinoviy23:
Ooh. Is it fine, that there is merge conflict? Who have to fix it?

ortem:
You have to. Just rebase onto master and resolve conflicts.

zinoviy23:
Oh, it shows merge conflict in adding function, what to do? I did steps from [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md)

ortem:
Now this PR adds 94 commits. You can squash your commits, rebase onto master and push.

zinoviy23:
OK. I'll try! Thanks for patience!

zinoviy23:
All commits? Or commits, which created after trying to resolve merge conflict?

zinoviy23:
It is merge conflict again. I'm squashed and rebased.

ortem:
Looks like you squashed not only your commits, but also a lot of commits from your local master. Did you really pull `upstream` (or `origin`, based on your remote configuration) before rebasing? Anyway, now you can reset your branch to the last commit before yours, then add only your changes, rebase the branch onto pulled master, and push it.

zinoviy23:
Merge conflict is fixed!!
Now I have question.
Is it necessary now to implement inverting if without else branch?
E.g. 
```rust
if a == 0 {
    return a;
} 
return b
```
converts to
```rust
if a != 0 {
   return b;
}

return a;
```

ortem:
@zinoviy23 I don't think it's necessary right now. You (and anyone who wants) can improve the intention later.

src/main/kotlin/org/rust/ide/intentions/InvertIfIntention.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        val newCondition = replacedIfExpr.condition!!.expr
+
+        if (newCondition is RsUnaryExpr && newCondition.excl != null) {
+            val binaryExpression = (newCondition.expr as? RsParenExpr)?.expr as? RsBinaryExpr ?: return
+            DemorgansLawIntention().invoke(project, editor, binaryExpression.binaryOp)
+        }
+    }
+
+    private fun getSuitableCondition(ifExpr: RsIfExpr): RsCondition? {
```
ortem: In Kotlin, such code can be simplified this way:
```Kotlin
private fun getSuitableCondition(ifExpr: RsIfExpr): RsCondition? =
    ifExpr.condition?.takeIf { it.let == null }
```

src/test/kotlin/org/rust/ide/intentions/InvertIfIntentionTest.kt
```diff
@@ -0,0 +1,98 @@
...
+            if a == 10 /*caret*/ {}
+        }
+    """)
+
+    fun `test if without then branch`() = doUnavailableTest("""
+        fn foo(a: i32) {if a == 10/*caret*/ else {}}
+    """)
+
+    fun `test simple inversion`() = doAvailableTest("""
+            fn foo() {
```
ortem: Too long indent (here and below)

zinoviy23:
Oh. Check is canceled:/

Undin:
@zinoviy23 I've canceled it manually because we want to make release today and we need to merge some important fixes as soon as possible. Additional travis tasks prevent this a little bit 😄 
I'll rerun build when we will finish

ortem:
It's OK. We are preparing for the new release now, so your PR will be merged later (don't forget to split your commits into two, like `Update CONTRIBUTORS.txt` and `INT: Invert if intention`). Thanks for contributing!

zinoviy23:
OK! I'll split!

Issue #3403, opened 2019-02-15
Imports should be grouped by default


mkpankov:
Automatic import now inserts a new `use` line for every import: so `std::collections::HashMap` and `std::collections::BTreeMap` are imported in potentially different places.

I'd like to see automatic imports grouping at least by last component, so `std::collections::HashMap` becomes `std::collections::{HashMap, BTreeMap}` when making second import.

Grouping all imports from same library in one `use` statement should be optional.

PR #3405, opened 2019-02-15
ACT: Try to group auto-inserted use item with existing ones by last component

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3403.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -493,6 +493,7 @@ private fun RsMod.insertExternCrateItem(psiFactory: RsPsiFactory, crateName: Str
 
 private fun RsMod.insertUseItem(psiFactory: RsPsiFactory, usePath: String) {
     val useItem = psiFactory.createUseItem(usePath)
+    if (tryGroupWithOtherUseItems(psiFactory, useItem)) return
     val anchor = childrenOfType<RsUseItem>().lastElement ?: childrenOfType<RsExternCrateItem>().lastElement
     if (anchor != null) {
         val insertedUseItem = addAfter(useItem, anchor)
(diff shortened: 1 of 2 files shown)


vlad20012:
_Really_ is should be done via "Optimize imports" action.

Also, should it be optional?

Undin:
> _Really_ is should be done via "Optimize imports" action.

Looks like no. "Optimize import" action usually reorders imports, removes unused, etc. I think it's unexpected during completion. For example, kotlin implementation of completion from index/auto import just insert new `import` item without additional actions

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixTest.kt
```diff
@@ -1598,4 +1604,160 @@ class AutoImportFixTest : AutoImportFixTestBase() {
...
+            pub struct Foo;
+            pub struct Bar;
+        }
+
+        fn main() {
+            let f = <error descr="Unresolved reference: `Bar`">Bar/*caret*/</error>;
+        }
+    """, """
+        use foo::Foo as F;
+        use foo::Bar;
```
Undin: Are you sure that we don't want to merge use items in this case?
`use foo::{Foo as F, Bar}` is valid and semantically the same

mchernyavsky: Fixed.

vlad20012:
Comments above use items should be preserved:
peek 2019-02-23 22-56

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -505,6 +506,49 @@ private fun RsMod.insertUseItem(psiFactory: RsPsiFactory, usePath: String) {
...
 }
 
+private fun RsMod.tryMergeWithOtherUseItems(psiFactory: RsPsiFactory, newUseItem: RsUseItem): Boolean =
+    childrenOfType<RsUseItem>().any { newUseItem.tryMergeWith(psiFactory, it) }
+
+private fun RsUseItem.tryMergeWith(psiFactory: RsPsiFactory, other: RsUseItem): Boolean {
+    if (other.vis != null || other.outerAttrList.isNotEmpty() || other.useSpeck?.isStarImport == true) return false
+
+    val parentPath = parentPath ?: return false
+    val importingName = importingNames?.singleOrNull() ?: return false
```
vlad20012: parentPath and importingName is computed multiple times for "this" when could be computed once

mchernyavsky:
Fixed.

PR #3412, opened 2019-02-17
Replace PsiTreeChangeEvent with more typesafe alternative

vlad20012:


src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
```diff
@@ -0,0 +1,225 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiTreeChangeEvent
+import com.intellij.psi.PsiTreeChangeListener
+import com.intellij.psi.impl.PsiTreeChangeEventImpl
+import java.util.*
+
+sealed class RsPsiTreeChangeEvent {
+    /**
+     * Event may be related to changes in a file system, e.g. file create/delete/movement.
```
Undin: `file create/delete/` -> `file creation/deletion/`

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
```diff
@@ -0,0 +1,225 @@
...
+     * In this case the property is set to null.
+     */
+    open val file: PsiFile? get() = null
+
+    sealed class ChildAdd(
+        override val file: PsiFile?,
+        val parent: PsiElement,
+        val child: PsiElement
```
Undin: Looks like should be nullable

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
```diff
@@ -0,0 +1,225 @@
...
+import java.util.*
+
+sealed class RsPsiTreeChangeEvent {
+    /**
+     * Event may be related to changes in a file system, e.g. file create/delete/movement.
+     * In this case the property is set to null.
+     */
+    open val file: PsiFile? get() = null
+
+    sealed class ChildAdd(
```
Undin: Maybe `ChildAdded` name is better?

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    sealed class ChildrenChange(
+        override val file: PsiFile?,
+        val parent: PsiElement,
+        /**
+         * "generic change" event means that "something changed inside an element" and
+         * sends after all events for concrete PSI changes in the element.
+         */
+        val isGenericChange: Boolean
```
Undin: Do we really need this property?
Currently, we always skip an event if `isGenericChange` is true. Maybe it's better not to create `ChildrenChange` events in such cases at all? It should reduce the count of allocated objects

src/main/kotlin/org/rust/lang/core/psi/RsPsiManager.kt
```diff
@@ -60,30 +60,28 @@ class RsPsiManagerImpl(val project: Project) : ProjectComponent, RsPsiManager {
...
-        override fun propertyChanged(event: PsiTreeChangeEvent) = onPsiChange(event)
-
-        private fun onPsiChange(event: PsiTreeChangeEvent, child: PsiElement? = event.child) {
-            // `GenericChange` event means that "something changed in the file" and sends
-            // after all events for concrete PSI changes in a file.
-            // We handle more concrete events and so should ignore generic event
-            if (event is PsiTreeChangeEventImpl && event.isGenericChange) return
+    inner class CacheInvalidator : RsPsiTreeChangeAdapter() {
+        override fun handleEvent(event: RsPsiTreeChangeEvent) {
+            val child = when (event) {
```
Undin: It seems that `element` name or something similar is more suitable here because sometimes we assign `event.parent` to `child` variable and it can be misleading

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
```diff
@@ -0,0 +1,225 @@
...
+            (event as? PsiTreeChangeEventImpl)?.isGenericChange == true
+        ))
+    }
+}
+
+private val PsiElement.safeText
+    get() = try {
+        text
+    } catch (e: Exception) {
+        "<exception>"
```
Undin: Hm, is exception message always useless? Is not, let's add it into returning value

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
vlad20012: `ChildAdded.Before` looks weird, I guess.

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
vlad20012: For debug purposes, I don't want to skip any events

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
vlad20012: I've added a comment to explain why is it needed. If element is invalid, in general, we can't call getText(). But sometimes it works, so for debug purposes we can use it. But if not, exception message is irrelevant

src/main/kotlin/org/rust/lang/core/psi/RsPsiTreeChangeListener.kt
Undin: For me, `ChildAdded.Before` and `ChildAdd.Before` are at the same level of weirdness
Nevertheless, I read `ChildAdd` as a command to add a child while `ChildAdded` looks like as some event on child addition. That's why I'd prefer the second variant

Issue #773, opened 2016-11-11
Compile errors highlighting


neuronsguy:
Hi,

Would it be possible to get rustc compilation errors to show up in the editor view as errors (i.e. red underlines) or warnings?

At the moment, the only way to get compile errors (I believe) is to use the 'Run cargo command' facility to run 'cargo build'.  Then errors show up in the output and the file/line is linked.

However it would be more useful if the errors showed in the editor the way syntax or other errors do, so that things like "jump to next error" would work.

I would be willing to help with implementing this (I know enough Java to do this, but I know exactly nothing about how IntelliJ plugins work, so I would likely need some pointers)

matklad:
The idea is to implement this thing on top of the RLS, because in theory it should be capable to produce the errors faster than plain `cargo build`. 

At the moment, you can use <kbd>Cltr+Alt+Down</kbd> shortcut to navigate between errors.

Techcable:
What about doing it on top of `cargo check`. Atom's [rust-linter](https://github.com/AtomLinter/linter-rust) does this and it's actually relatively fast. I think this would be a good stopgap solution while we get traditional IDE-based error checking in.

matklad:
The stop-gap solution won't that easy to implement unfortunately, because of the different models in atom and Intellij.

In Intellij, there's no "save" button, if you want to show errors inline in the editor, you should be prepared to work with in-memory representation of files. rls should be able to do that eventually (currently it is a quite a bit wobbly about in-memory files), but `cargo check` external command and a future build-in `cargo check` command won't handle this use case.

Techcable:
@matklad 
Do you mean getting the errors from the [Rust Language Service](https://github.com/jonathandturner/rls)?

matklad:
@Techcable yep, I mean getting errors from the RLS. `cargo check`'s main aim was actually to provide a programmatic API for using Cargo from within RLS: https://github.com/jonathandturner/rls/commit/aaeec565f9c414bf9b7eadb5941c7fe71bb5d047

Maintaining build queue and syncing files on the IDE side is of course possible, but this is a non trivial amount of work, so I'd prefer to build the solution on top of the RLS.

sanmai-NL:
[...]

PR #3415, opened 2019-02-18
INSP & ACT: Add cargo check inspection and run cargo check action

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/773.
Depends on https://github.com/intellij-rust/intellij-rust/pull/3432 and https://github.com/intellij-rust/intellij-rust/pull/3466.

image


vlad20012:
There is a (known) bug: if a user changes a file when the inspection is running, the user will see annotations at incorrect ranges after inspection finish. But today it's better with such inspection than without it and I hope we will fix it in the future.

bors r+

PR #3419, opened 2019-02-19
T: test our analysis on real rust projects

vlad20012:
This PR adds new test `RsRealProjectAnalysisTest` for testing our inspections and error annotators on real rust projects. It allows to easy find possible false positives produced by some new analysis. Currently, we have a LOT of false positives with existing analysis. These false positives should be fixed eventually. 

These tests are disabled on CI because they are very slow and because they are always fails (due to existing false-positives). Use it manually when adding some new analysis or changing behavior of an existing one.

### DIFF
--- a/src/test/kotlin/org/rustSlowTests/RsRealProjectAnalysisTest.kt
+++ b/src/test/kotlin/org/rustSlowTests/RsRealProjectAnalysisTest.kt
@@ -1,0 +1,89 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rustSlowTests
+
+import com.intellij.codeInspection.ex.InspectionToolRegistrar
+import org.junit.ComparisonFailure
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rustSlowTests/RsRealProjectAnalysisTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.rust.ide.inspections.RsLocalInspectionTool
+import org.rust.lang.RsFileType
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.ext.cargoWorkspace
+import org.rust.openapiext.toPsiFile
+import java.lang.reflect.Field
+
+class RsRealProjectAnalysisTest : RsRealProjectTestBase() {
+
+    /** Do't run it on Rustc! It's a kind of stress-test */
```
ortem: `Don't`

vlad20012:
Fixed

PR #3420, opened 2019-02-19
ACT: add `Reformat File with Rustfmt` to cargo toolwindow

Undin:


src/main/resources/META-INF/plugin.xml
```diff
@@ -817,7 +817,8 @@
         <action id="Rust.CargoFmtFile"
                 class="org.rust.ide.actions.RustfmtFileAction"
                 text="Reformat File with Rustfmt"
-                description="Reformat current file with rustfmt">
+                description="Reformat current file with rustfmt"
+                icon="AllIcons.Actions.RealIntentionBulb">
```
Undin: I think we need some other icon but it's ok for now

PR #3423, opened 2019-02-20
PERF&MACRO&RES: Fix PSI switch to AST during macro resolve

vlad20012:
We switched to AST during resolve of a macro call
form expansion of another macro

bors r+


mchernyavsky:
bors retry

Issue #3400, opened 2019-02-14
Add `.ref` postifx template


matklad:
Like `foo.not` -> `!foo`, but `foo.ref` -> `&foo`

matklad:
and `refm` for `&mut foo`

Undin:
Implementation of `not` postfix template can be found [here](https://github.com/intellij-rust/intellij-rust/blob/34bd4c53bccdf5b933205058a093381043406189/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt#L46-L52)

matklad:
Issues with links to code should have E-mentor as well, right?

PR #3425, opened 2019-02-21
COMP: add `.ref` and `.refm` postifx template

t-kameyama:
fixes https://github.com/intellij-rust/intellij-rust/issues/3400


Undin:
@t-kameyama Thanks!

PR #3431, opened 2019-02-21
ANN: Fix saving documents in cargo check annotator

mchernyavsky:


vlad20012:
With this fix Trailing Spaces Stripper totally stops working if cargo check is enabled

mchernyavsky:
Fixed.

src/test/kotlin/org/rustSlowTests/RsCargoCheckAnnotatorTest.kt
```diff
@@ -131,6 +133,25 @@ class RsCargoCheckAnnotatorTest : RsWithToolchainTestBase() {
         }
     }
 
+    fun `test strip document later reflection`() {
```
Undin: If I understand correctly, this test checks that reflection doesn't throw exceptions
If yes, let's add comment about it

PR #3424, opened 2019-02-20
COMP: automatically add trait import while trait method completion

Undin:
Also, enable default insert handler for index-based completion

Fixes #3406
Related to #2174

PR #3441, opened 2019-02-25
COMP: Add setting for "Suggest out of scope items" while completion

mchernyavsky:


src/main/kotlin/org/rust/ide/settings/RsCodeCompletionOptions.kt
```diff
@@ -0,0 +1,34 @@
...
+package org.rust.ide.settings
+
+import com.intellij.application.options.CodeCompletionOptionsCustomSection
+import com.intellij.ui.IdeBorderFactory
+import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.layout.panel
+import org.rust.openapiext.CheckboxDelegate
+import javax.swing.JPanel
+
+class RsCodeCompletionConfigurable : CodeCompletionOptionsCustomSection {
```
Undin: Looks like `CodeCompletionOptionsCustomSection` is available since `183.4886 (IDEA 2018.3.2)`
Let's set `sinceBuild` to `183.4886` in `gradle-183.properties` and increase oldest IDEA version from `2018.3` to `2018.3.2` in `.travis.yml`

mchernyavsky:
Fixed.

src/main/kotlin/org/rust/ide/settings/RsCodeCompletionOptions.kt
vlad20012: Mmm, can we just drop support of these versions?

Undin: Yep. The following actions 
> Let's set `sinceBuild` to `183.4886` in `gradle-183.properties` and increase oldest IDEA version from `2018.3` to `2018.3.2` in `.travis.yml`

just disable the plugin for all IDE's below `2018.3.2`

Undin:
bors retry

Undin:
@mchernyavsky Could you rebase onto master?

mchernyavsky:
Fixed.

amakeev:
Shouldn't the new setting be in Auto Import section?

Undin:
It's an interesting question
From the one hand,  "Suggest out of scope items" feature is not about auto import because it just adds out of scope items into completion result. So I think it's more about completion than auto import
From the other hand, we've also introduced "Add trait imports while trait methods completion" option (see related changes [here](https://github.com/intellij-rust/intellij-rust/pull/3424/files#diff-5e3fd698b4dc8af31907bdcd46a1da3f) for trait method completion. Probably, it makes sense to generalize this option and check it when we apply completion variant with out of scope item (i.e. where auto import is really used)
@mchernyavsky @amakeev What do you think?

amakeev:
Oh, right, it's indeed not quite clear where such 'import on completion' option belong to.
In which cases would I need to disable out of scope items from completion?

Undin:
@amakeev the main case is to allow a user to turn off this feature if we have a bug in our implementation (or a performance issue), i.e. return the plugin behavior to "before feature" state

amakeev:
@Undin I see; in this case the setting could be a registry option, so that we could easily remove it when feature works as expected.

Issue #3277, opened 2019-01-21
Provide AddMutableFix for assignments with immutable references


ortem:
`RsAssignToImmutableInspection` [provides](https://github.com/intellij-rust/intellij-rust/pull/3200)  `AddMutableFix` for values, but not for references. It would be useful to provide such fix for function parameters with reference type, e.g.:
```Rust
struct S { a: i32 }
fn foo(s: &S) {
    s.a = 42; // Cannot assign to field of immutable binding [E0594]
}
```
becomes
```Rust
struct S { a: i32 }
fn foo(s: &mut S) {
    s.a = 42; // OK
}
```

snpefk:
Hello. 

I would like to implement this feature, but I can't figure out the purpose of `RsPatBinding`.

ortem:
Hi @snpefk,

`RsPatBinding` is a part of `PatIdent` (which relates to [IdentifierPattern](https://doc.rust-lang.org/reference/patterns.html#identifier-patterns) in Rust Reference). It consists of an optional binding mode (`mut`, `ref` or `ref mut`) and a identifier.

You can use `PsiViewer` to discover Rust grammar.
<img width="769" alt="screenshot 2019-01-28 at 17 46 48" src="https://user-images.githubusercontent.com/4854600/51843790-cd335900-2324-11e9-9658-c0cf3679daca.png">

snpefk:
Shouldn't it be closed now?

Undin:
@snpefk Yep!
BTW if you add something like `Closes #issue-number` in PR message, Github will automatically close the issue when PR is merged
See [here](https://help.github.com/en/articles/closing-issues-using-keywords) for details

PR #3442, opened 2019-02-25
INSP: Provide AddMutableFix for assignment with immutable ref for RsAssignToImmutableInspection 

snpefk:
If I understand issue correctly this PR should work for #3277.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
@@ -10,14 +10,14 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import org.rust.ide.annotator.isMut
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.psi.ext.RsBindingModeKind.BindByReference
 import org.rust.lang.core.psi.ext.RsBindingModeKind.BindByValue
-import org.rust.lang.core.psi.ext.RsNamedElement
-import org.rust.lang.core.psi.ext.ancestorStrict
-import org.rust.lang.core.psi.ext.kind
-import org.rust.lang.core.psi.ext.typeElement
 import org.rust.lang.core.types.declaration
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.type
 
 class AddMutableFix(binding: RsNamedElement) : LocalQuickFixAndIntentionActionOnPsiElement(binding) {
     private val _text = "Make `${binding.name}` mutable"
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
```diff
@@ -32,10 +32,18 @@ class AddMutableFix(binding: RsNamedElement) : LocalQuickFixAndIntentionActionOn
...
-            return if (declaration is RsSelfParameter || (declaration is RsPatBinding && declaration.kind is BindByValue)) {
-                AddMutableFix(declaration as RsNamedElement)
-            } else {
-                null
+
+            return when {
+                declaration is RsSelfParameter -> {
+                    AddMutableFix(declaration as RsNamedElement)
+                }
+                expr.type is TyReference && declaration is RsPatBinding && declaration.isArg && !declaration.isMut -> {
```
ortem: `isMut` is checked before the creation of the quick fix, so you shouldn't check it here. Also, I found the problem. Consider such code:
```Rust
fn foo(ref s: &S) {
    s.a = 42;
}
```
Converting the argument to `ref s: &mut S` doesn't help. rustc also doesn't suggest such fix to this code because it doesn't work.

 So you should check `declaration.kind is BindByValue` here, and then the resulted arms will be:
`declaration is RsPatBinding && declaration.kind is BindByValue && declaration.isArg`
`declaration is RsPatBinding && declaration.kind is BindByValue && expr.type !is TyReference`

Also, I'm not sure that these arms are completely correct for us, so it'd be really great if you test this fix on different cases of Rust code (and add such tests to `AddMutableFixTest`).

src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
```diff
@@ -32,10 +32,18 @@ class AddMutableFix(binding: RsNamedElement) : LocalQuickFixAndIntentionActionOn
     companion object {
         fun createIfCompatible(expr: RsExpr): AddMutableFix? {
             val declaration = expr.declaration
```
ortem: Maybe perform cast to `RsNamedElement` here? Like `val declaration = expr.declaration as? RsNamedElement ?: return null`

ortem:
@snpefk Welcome to IJ Rust contributors! Please add your name to `CONTRIBUTORS.txt` and rebase your branch onto master.

ortem:
Don't forget to rebase onto master and squash your commits into 2 (one about contributors.txt, another one with code changes)

src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
snpefk: I'll try to figure out something

snpefk:
Yeah, I'll do it, but later (rebase is hard).

snpefk:
Okay, I'm done with rebase

src/main/kotlin/org/rust/ide/annotator/fixes/AddMutableFix.kt
ortem: OK, you're welcome to create another PR if you find something is missed.

ortem:
Looks good, thanks! I'll merge it soon.

PR #3447, opened 2019-02-26
Support multiple patterns in `if let` and `while let` expressions

Undin:
This feature will be stabilized in Rust 1.33
See https://github.com/rust-lang/rust/pull/57532 for more details

Note, some existing features are not supported in this PR:
* control flow graph (@ortem)
* memory categorization (@ortem)
* irrefutable patterns (@mchernyavsky) (not sure that we should do something with it)
`RsCondition.pat` is introduced exactly to keep existing behavior for all cases that we supported before this PR


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -1025,13 +1025,22 @@ private fun processLexicalDeclarations(
         return processAll(boundNames, processor)
     }
 
-    fun processCondition(condition: RsCondition?, processor: RsResolveProcessor): Boolean {
-        if (condition == null || condition == cameFrom) return false
-        val pat = condition.pat
+    fun processPatList(patList: List<RsPat>, processor: RsResolveProcessor): Boolean {
```
ortem: Due to the current name, someone might apply this function to any `patList` (e.g. `RsPatTup.patList`). Maybe rename it somehow to note that there is only one bind? Something like `processPatListBindOnce`.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -1025,13 +1025,22 @@ private fun processLexicalDeclarations(
...
+        if (Namespace.Values !in ns) return false
```
ortem: Looks like this line should be placed under `is RsMatchArm -> {` for uniformity.

Undin: But it will force us to add a similar line into `processCondition`

ortem: This line is executed before each call of `processCondition`

Undin: Dropped redundant namespace checks

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -1438,12 +1438,15 @@ class RsFnInferenceContext(
...
+        val patList = patList
+        if (patList.isNotEmpty()) {
             // if let Some(a) = ... {}
+            // if let V1(a) | V2(a) = ... {}
             // or
             // while let Some(a) = ... {}
-            pat.extractBindings(resolveTypeVarsWithObligations(expr.inferType()))
+            // while let V1(a) | V2(a) = ... {}
+            val exprTy = expr.inferType()
+            patList.forEach { it.extractBindings(resolveTypeVarsWithObligations(exprTy)) }
```
ortem: The same logic is used inside `inferMatchExprType`. So IMO it should be extracted like this (also `resolveTypeVarsWithObligations(exprTy)` will be computed once):
```Kotlin
private fun RsOrPats.inferTypes(matchingExprTy: Ty) {
    patList.forEach { it.extractBindings(matchingExprTy) }
}

private fun inferMatchExprType(expr: RsMatchExpr, expected: Ty?): Ty {
    // ...
        arm.orPats.inferTypes(matchingExprTy)
    // ...
}

private fun RsCondition.inferTypes() {
    val orPats = orPats
    if (orPats != null) {
        val matchingExprTy = resolveTypeVarsWithObligations(expr.inferType())
        orPats.inferTypes(matchingExprTy)
    } else {
        expr.inferType(TyBool)
    }
}
```

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -332,10 +343,15 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
     }
 
     private fun checkCondition(holder: AnnotationHolder, element: RsCondition) {
-        val pat = element.pat ?: return
-        if (pat.isIrrefutable) {
+        val pat = element.pat
+        if (pat != null && pat.isIrrefutable) {
             checkFeature(holder, pat, IRREFUTABLE_LET_PATTERNS, "irrefutable let pattern")
```
ortem: Does this feature should be checked in case `if let <refutable pattern> | <irrefutable pattern>` and `if let <irrefutable pattern> | <refutable pattern>`? If such cases are possible, let's add corresponding tests to `RsErrorAnnotatorTest.kt`

Undin: It's an open question - I did investigate it (see comment in PR description)
I've just keep previous behavior for all possible cases before stabilization of the feature. So it can produce false positive annotations with new syntax

Issue #3163, opened 2018-12-20
Liveness analysis


ortem:
Implement [liveness analysis](https://github.com/rust-lang/rust/blob/master/src/librustc/middle/liveness.rs) including:
- Unused variable
- Dead assignment
- Unused function arguments

Kobzol:
I wonder how the `Remove variable` fix should work.
Changing `let x = 5;` into `5;` is easy, but often the user would also like to remove the expression. We could detect if the expression has side effects (= pretty much if it's a function call), or simply remove the whole declaration, or introduce two intentions?

`Remove variable` -> `let x = 5;` changes to `(empty line)`
`Remove variable (keep expression)` -> `let x = 5;` changes to `5;`

fujiapple852:
@ortem the unused variable change seems to be causing false positives for variable used only within declarative macros.  Shall I raise an issue?

Undin:
@fujiapple852 it would be great

fujiapple852:
Done: https://github.com/intellij-rust/intellij-rust/issues/4995

PR #3448, opened 2019-02-27
WIP: Add liveness analysis (unused parameter/variable)

ortem:
Relates to #3163


ortem:
Reopen from `origin` instead of `ortem` #4710

PR #3485, opened 2019-03-04
Provide binary compatible API for Rust project settings

Undin:
Before these changes, there was only one way to change almost all properties of `RustProjectSettingsService`:
get `data`, create new one with necessary changes and set new instance to the settings.
Because of `RustProjectSettingsService.Data` is immutable, you have to create new instance via constructor or `copy` method.
But any changes in class lead to `java.lang.NoSuchMethodError` in external code
because signatures of constructor and `copy` method are changing accordingly

These changes provide setter for each property (i.e. make each property mutable)
It should be enough to change any property and do not touch `RustProjectSettingsService.Data` class

### DIFF
--- a/idea/src/main/kotlin/org/rust/ide/idea/CargoConfigurationWizardStep.kt
+++ b/idea/src/main/kotlin/org/rust/ide/idea/CargoConfigurationWizardStep.kt
@@ -61,10 +61,10 @@ class CargoConfigurationWizardStep private constructor(
         override fun update(module: Module, rootModel: ModifiableRootModel) {
             val data = data
             if (data != null) {
-                module.project.rustSettings.data = module.project.rustSettings.data.copy(
-                    toolchain = data.toolchain,
-                    explicitPathToStdlib = data.explicitPathToStdlib
-                )
+                module.project.rustSettings.modify {
+                    it.toolchain = data.toolchain
+                    it.explicitPathToStdlib = data.explicitPathToStdlib
+                }
             }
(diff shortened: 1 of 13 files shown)


matklad:
The original motivation for `Data` pattern was to enable transnational changes, where the state as a whole moves from `X` to `Y`. If this is in fact not necessary, I suggest getting rid of `Data` altogether: there's a similar `State` inside. 

If transnational semantics is required (I think it might be useful to avoid publishing several notifications in a row), we could invent some kind of closure-based transactional API:

```
settings.modify {
    toolchain = foo
    ...
}
```

That is, a `modify` would accept the callback which acts on a "mutable" object which records changes. Upon exit from `modify`, we apply changes transactionaly.

matklad:
Here's the bit that handles transactionality: https://github.com/intellij-rust/intellij-rust/blob/37467de8d2e914e77df10cb7c096f291411a1cc7/src/main/kotlin/org/rust/cargo/project/settings/impl/RustProjectSettingsServiceImpl.kt#L83-L86

Undin:
@matklad Sounds good!

matklad:
A specific code smell in this particular implementation is that there are both `var Data` and individual setters. Such non-orthogonality is usually a sign that something is wrong :)

vlad20012:
Rebase it, please

Undin:
@vlad20012 done

vlad20012:
compilation fail

Undin:
bors r=vlad20012

PR #3472, opened 2019-03-03
ANN & RUN: Use "external linter" as a frontend for both cargo check and clippy

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/1583.

Depends on https://github.com/intellij-rust/intellij-rust/pull/3434.

PR #3494, opened 2019-03-07
[WIP] Notify user if external annotator is disabled

mchernyavsky:
Depends on https://github.com/intellij-rust/intellij-rust/pull/3472.


src/main/kotlin/org/rust/ide/notifications/EnableExternalLinterNotificationProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        return EditorNotificationPanel().apply {
+            setText("External linter is disabled")
+            for (externalLinter in ExternalLinter.values()) {
+                createActionLabel("Use ${externalLinter.title}") {
+                    project.rustSettings.modify {
+                        it.externalLinter = externalLinter
+                        it.runExternalLinterOnTheFly = true
+                    }
```
amakeev: Might be a good idea to hide existing notifications right away.

PR #3490, opened 2019-03-06
RUN: Run rustfmt on the document before saving

mchernyavsky:
`Preferences > Languages & Frameworks > Rust > Rustfmt > Run rustfmt on Save`

Also, now rustfmt uses the `rustfmt.toml` from the corresponding cargo project root.

Relates to https://github.com/intellij-rust/intellij-rust/issues/1087.

PR #3499, opened 2019-03-08
QDOC: improvements in macro quick documentation

Undin:
* fix external doc URLs generation for macros defined in non crate root (like `assert_eq` or `write`)
* fix documentation link resolution when link refers to a macro defined in non crate root

Note, `OLD_RUST_VERSION` was increased to 1.28 because:
* before 1.28 stdlib uses special attribute to reexport macros. It fails the corresponding test
* we also need 1.28 as minimal rustc version because of #3490


src/main/kotlin/org/rust/ide/docs/RsDocumentationProvider.kt
```diff
@@ -130,9 +131,8 @@ class RsDocumentationProvider : AbstractDocumentationProvider() {
         }
 
         // macros without #[macro_export] are not public and don't have external documentation
```
vlad20012: Is this comment outdated now?

PR #3500, opened 2019-03-09
GRAM: fix macro_rules pin

vlad20012:
Correctly parse
```rust
let mut macro_rules = 0;
macro_rules += 1;
```


vlad20012:
bors retry

Issue #2577, opened 2018-06-09
Intention to nest/flatten imports


matklad:
```rust
use std::error::Error;
use std::io::Write;
use std::path::PathBuf;
```

<-->

```rust
use std::{
    error::Error,
    io::Write,
    path::PathBuf,
};
```

Flavsditz:
I'd like to take a shot at this one. If I need some guidance I will contact you @matklad. Is that ok?

dvtomas:
In an ideal world, there would be a configurable import optimizer (see also #2259), that would be able to group imports by std/cargo crates/workspace crates/local imports, sorting imports in each group, nesting/flattening them according to the settings, removing curly braces for single imports etc. Unused imports should be automatically removed. When doing a move or copy refactoring to another file, relevant imports should be copied as well. All reformatting and removing unused imports should happen just on pressing Ctrl-Alt-O, one should almost never have to touch the imports section manually, that is a job for a machine, not for human. This is how it works in the Scala plugin, I've used that for a couple of years so now I see just how much time is wasted by unnecessary manual import maintenance. I'm really looking forward this area being worked on. That said, even the "Import" action for unknown symbols has been a huge help, thank you for that!

Sorry for the digress. IMO this should be a configuration setting in an automatic import optimizer, not an intention, at least in the long run.

majecty:
I impelmented NestUseStatementsIntention.
You can find it at https://github.com/intellij-rust/intellij-rust/pull/2938

When implementing the feature, I met some issues.

1. Should we nest use statements full depth?

full depth
```rust
use std::{
    io::Write,
    sync::{
      Arc,
      Mutex
    },
};
```

only 1 depth
```rust
use std::{
    io::Write,
    sync::Arc,
    sync::Mutex,
};;
```

2. Should we preserve statements order or sort them by dictionary order?

If someone has an opinion, please say it freely.

majecty:
I implemented nest intention to only nest 1 depth and preserve original order.

majecty:
@vlad20012 
This issue's subject is "nest/flatten imports".
An intention that flattens imports is not implemented yet.
I think this issue should be reopened.

vlad20012:
Oh, I'm sorry

fan-tom:
Are we need to add an inspection to show programmer that nesting is possible? Because I even didn't know about this feature until I saw that issue

fan-tom:
There is bug when nesting use statements like this
```rust
use foo;
[...]

PR #3503, opened 2019-03-09
INT: handle conversion of module to self in nest use intention

Kobzol:
This PR fixes an error in the `Nest use statements` intention mentioned here: https://github.com/intellij-rust/intellij-rust/issues/2577#issuecomment-469150465.

Before the intention would change this code:
```
use foo;
use foo::bar;
```
into
```
use foo::{foo, bar}`;
```
which is incorrect.

Now it checks if the grouped path is the same as the base path and converts it to `self` if necessary.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -47,6 +47,7 @@ johnthagen
 jonas-schievink
 Keats
 king6cong
+Kobzol
 kumbayo
 la10736
 LiamClark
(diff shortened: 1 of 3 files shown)


vlad20012:
Thanks! 
Please, sing the CLA and add youself to CONTRIBUTORS.md as described [here](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#cla), I'll merge it right after that.

vlad20012:
Thanks!
bors r+

PR #3511, opened 2019-03-09
QDOC: provide quick documentation for primitive types

Undin:


vlad20012:
bors r+

Btw links to primitive types don't work. Try `use std::i32;`, open docs for that `i32` and try to click on link to primitive type

PR #3517, opened 2019-03-11
Fixed formatting issues

sntdevco:
The issue template does not give the expected outcome and I've edited it so that it looks the way we want it to


ortem:
Hi @sntdevco! Could you please explain, what's the problem with the current template? It seems to work fine.

sntdevco:
There's no problem! It was a problem with my markdown parser (experimental - written in rust)! Sorry for that! `closing!()`

PR #3518, opened 2019-03-11
COMP: provide primitive type completion

Undin:


src/main/kotlin/org/rust/lang/core/completion/RsCompletionContributor.kt
```diff
@@ -18,6 +18,7 @@ class RsCompletionContributor : CompletionContributor() {
         extend(CompletionType.BASIC, RsDeriveCompletionProvider.elementPattern, RsDeriveCompletionProvider)
         extend(CompletionType.BASIC, AttributeCompletionProvider.elementPattern, AttributeCompletionProvider)
         extend(CompletionType.BASIC, RsMacroCompletionProvider.elementPattern, RsMacroCompletionProvider)
+        extend(CompletionType.BASIC, RsPrimitiveTypeCompletionProvider.elementPattern, RsPrimitiveTypeCompletionProvider)
```
mchernyavsky: Maybe it would be better to use this provider before `RsCommonCompletionProvider`?

src/main/kotlin/org/rust/lang/core/completion/RsPrimitiveTypeCompletionProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+        add(TyStr.name)
+        add(TyChar.name)
+    }
+
+    val elementPattern: ElementPattern<PsiElement> get() =
+        PlatformPatterns.psiElement().withSuperParent<RsTypeReference>(3)
+
+    override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
+        primitives.forEach {
+            result.addElement(LookupElementBuilder.create(it).bold())
```
mchernyavsky: Let's set a higher priority on lookup elements to group all primitive types and avoid this situation:
example

Undin:
@mchernyavsky Done

Issue #2944, opened 2018-10-15
Implement `New Cargo Project` action


Undin:
It would be great to have `New > Cargo Project` action in the context menu like `New > Rust File` action that we already have.

amakeev:
`File | New Project` seem to already have 'Rust' project type in IntelliJ IDEA and CLion.
Do you mean an action that would create a Cargo.toml?

vlad20012:
> Do you mean an action that would create a Cargo.toml?

Yep. Well-formatted Cargo.toml and `src` directory. It's usual case to crate a nested crate.

Kobzol:
Fixed by https://github.com/intellij-rust/intellij-rust/pull/3519.

PR #3519, opened 2019-03-11
ACT: add context menu action to create a new Cargo (sub)crate

Kobzol:
Hi, I tried implementing an action that would create a new Cargo crate in a project (as described here: https://github.com/intellij-rust/intellij-rust/issues/2944).

It's my first time making a non-trivial contribution to IntelliJ, so please let me know if I made any obvious mistakes.

- [x] Create new cargo project using `cargo init`
- [x] Tests
~~Workspace modification - it would be nice to add the new crate to an existing Cargo workspace, if there is one in the project~~
```toml
[workspace]

members = [
    "main",
    "new-crate" # this would get added automatically by the action
]
```

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/RsCreateCrateAction.kt
+++ b/src/main/kotlin/org/rust/ide/actions/RsCreateCrateAction.kt
@@ -1,0 +1,63 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.actions
+
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.CommonDataKeys
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.application.runWriteAction
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import org.rust.cargo.icons.CargoIcons
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.runconfig.command.RunCargoCommandActionBase
+import org.rust.cargo.toolchain.RustToolchain
+import org.rust.ide.actions.ui.showCargoNewCrateUI
+import org.rust.openapiext.pathAsPath
+
+
+class RsCreateCrateAction : RunCargoCommandActionBase(CargoIcons.ICON) {
+    override fun actionPerformed(e: AnActionEvent) {
+        val dataContext = e.dataContext
+        val project = CommonDataKeys.PROJECT.getData(dataContext) ?: return
+        val root = getRootFolder(dataContext) ?: return
+        val toolchain = project.toolchain ?: return
+
+        val ui = showCargoNewCrateUI(project, root)
+        ui.selectCargoCrateSettings()?.let {
+            createProject(project, toolchain, root, it.crateName, it.binary)
+        }
+    }
+
+    private fun getRootFolder(dataContext: DataContext): VirtualFile? {
+        val file = CommonDataKeys.VIRTUAL_FILE.getData(dataContext) ?: return null
+        return if (!file.isDirectory) {
+            file.parent
+        } else file
+    }
+
+    private fun createProject(
+        project: Project,
+        toolchain: RustToolchain,
+        root: VirtualFile,
+        name: String,
+        binary: Boolean
+    ) {
+        val cargo = toolchain.cargoOrWrapper(
+            project.cargoProjects.findProjectForFile(root)?.workspaceRootDir?.pathAsPath)
+
+        val targetDir = runWriteAction {
+            root.createChildDirectory(this, name)
+        }
+        cargo.init(project, targetDir, binary, "none")
+
+        val manifest = targetDir.findChild(RustToolchain.CARGO_TOML)
+        manifest?.let {
+            project.cargoProjects.attachCargoProject(it.pathAsPath)
+        }
+    }
+}
--- a/src/main/kotlin/org/rust/ide/actions/ui/CargoNewCrateDialog.kt
+++ b/src/main/kotlin/org/rust/ide/actions/ui/CargoNewCrateDialog.kt
@@ -1,0 +1,90 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.actions.ui
+
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.ui.ComboBox
+import com.intellij.openapi.ui.DialogWrapper
+import com.intellij.openapi.ui.ValidationInfo
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.ui.components.JBTextField
+import org.jetbrains.annotations.TestOnly
+import org.rust.ide.newProject.RsPackageNameValidator
+import org.rust.ide.ui.layout
+import org.rust.openapiext.isUnitTestMode
+import javax.swing.JComponent
+
+data class CargoNewCrateSettings(val binary: Boolean, val crateName: String)
+
+interface CargoNewCrateUI {
+    fun selectCargoCrateSettings(): CargoNewCrateSettings?
+}
+
+private var MOCK: CargoNewCrateUI? = null
+
+fun showCargoNewCrateUI(
+    project: Project,
+    root: VirtualFile
+): CargoNewCrateUI {
+    return if (isUnitTestMode) {
+        MOCK ?: error("You should set mock ui via `withMockCargoNewCrateUi`")
+    } else {
+        CargoNewCrateDialog(project, root)
+    }
+}
+
+@TestOnly
+fun withMockCargoNewCrateUi(mockUi: CargoNewCrateUI, action: () -> Unit) {
+    MOCK = mockUi
+    try {
+        action()
+    } finally {
+        MOCK = null
+    }
+}
+
+class CargoNewCrateDialog(project: Project, private val root: VirtualFile) : DialogWrapper(project), CargoNewCrateUI {
+    private val typeCombobox = ComboBox<String>(arrayOf("Binary", "Library"))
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/ide/actions/RsCreateCrateAction.kt
```diff
@@ -0,0 +1,52 @@
...
+    override fun actionPerformed(e: AnActionEvent) {
+        val dataContext = e.dataContext
+        val project = CommonDataKeys.PROJECT.getData(dataContext) ?: return
+        val root = CommonDataKeys.VIRTUAL_FILE.getData(dataContext) ?: return
+        val toolchain = project.toolchain ?: return
+
+        val dialog = CargoNewCrateDialog(project, root)
+        val ok = dialog.showAndGet()
+        if (ok) {
+            this.createProject(project, toolchain, root, dialog.crateName, dialog.binary)
```
Undin: `this` is redundant here

src/main/kotlin/org/rust/ide/actions/RsCreateCrateAction.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.runconfig.command.RunCargoCommandActionBase
+import org.rust.cargo.toolchain.RustToolchain
+import org.rust.ide.actions.ui.CargoNewCrateDialog
+import org.rust.ide.icons.RsIcons
+import org.rust.openapiext.pathAsPath
+
+
+class RsCreateCrateAction : RunCargoCommandActionBase(RsIcons.CARGO),
+                            WriteActionAware {
```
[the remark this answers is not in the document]
Kobzol: Thanks, I forgot it there. I used it initially before I added the dialog and then forgot to remove it.

src/main/kotlin/org/rust/ide/actions/ui/CargoNewCrateDialog.kt
```diff
@@ -0,0 +1,63 @@
...
+class CargoNewCrateDialog(project: Project, private val root: VirtualFile) : DialogWrapper(project) {
+    private val buttons = listOf(
+        JBRadioButton("Binary", true),
+        JBRadioButton("Library")
+    )
+    private val buttonGroup = ButtonGroup()
+    private val name = JBTextField(20)
+
+    val binary
+        get() = this.buttons[0].isSelected
```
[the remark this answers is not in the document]
Kobzol: Ok. I don't consider it redundant (explicit is better IMO), but a quick look tells me that this is not the style used in this project. I will remove it, thanks.

Undin:
> Tests (I don't know how an action with filesystem side effects should be properly tested).

We usually use `RsWithToolchainTestBase` as base class for such tests. In this case, it's supposred that you create file in local file system. And in such environment cargo should work as expected.
You can see example how to invoke action in tests [here](https://github.com/intellij-rust/intellij-rust/blob/a711259520f3369f7c71242bfb364d14e46d666a/src/test/kotlin/org/rust/cargo/commands/RustfmtTest.kt#L289)

> it would be nice to add the new crate to an existing Cargo workspace, if there is one in the project

At this moment, it's quite complicated bacause we need psi for toml but we user can have only IntellJ Rust without Toml plugin. Probably, we should move this action into `toml` module where we keep code related to Rust and Toml plugin integration. But we don't sure what way is better in this case so I suggest not to implement this point in this PR

src/main/resources/META-INF/plugin.xml
```diff
@@ -800,6 +800,11 @@
             <add-to-group group-id="NewGroup" anchor="before" relative-to-action="NewFile"/>
         </action>
 
+        <action id="Rust.NewCargoCrate" class="org.rust.ide.actions.RsCreateCrateAction"
+                text="Cargo crate" description="Create new Cargo crate.">
```
amakeev: Stylistics: 
Action names should be headline capitalized (`Cargo Crate`)
Descriptions should not end with `.`

Kobzol: Fixed, thanks.

Kobzol:
@Undin 
> We usually use RsWithToolchainTestBase as base class for such tests. In this case, it's supposred that you create file in local file system. And in such environment cargo should work as expected.
You can see example how to invoke action in tests here

The problem is with the UI dialog that the action creates though. I haven't found any tests for the 
`RsCreateFileAction`, which is the most similar action.

> At this moment, it's quite complicated bacause we need psi for toml but we user can have only IntellJ Rust without Toml plugin. Probably, we should move this action into toml module where we keep code related to Rust and Toml plugin integration. But we don't sure what way is better in this case so I suggest not to implement this point in this PR

Ok, let's keep it for a different PR.

Kobzol:
@Undin is there a way to test the UI dialog properly?

Undin:
@Kobzol Oh, sorry for so long response. I've missed your previous message somehow

> @Undin is there a way to test the UI dialog properly?

Currently, we don't have UI tests. We usually mock dialogs manually and provide some hardcoded user data via these mocks in tests. In simple cases, it's enough use `com.intellij.openapi.ui.Messages#setTestDialog` method. But your dialog is more complicated, so you should create mock manually.
As an example, you can look at auto import action [ui](https://github.com/intellij-rust/intellij-rust/blob/218adb4196334dad6490f6785b5cd18ba465c09b/src/main/kotlin/org/rust/ide/inspections/import/ui.kt). Interesting places are [here](https://github.com/intellij-rust/intellij-rust/blob/218adb4196334dad6490f6785b5cd18ba465c09b/src/main/kotlin/org/rust/ide/inspections/import/ui.kt#L29) and [here](https://github.com/intellij-rust/intellij-rust/blob/218adb4196334dad6490f6785b5cd18ba465c09b/src/main/kotlin/org/rust/ide/inspections/import/ui.kt#L44)

Kobzol:
@Undin Thank you! I added test for creating binary and library crates and also for creating a crate from a file context (as opposed to a directory context). I didn't like the global statick `MOCK` solution, so I inject the UI creator directly into the action.

src/main/kotlin/org/rust/ide/actions/RsCreateCrateAction.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.rust.cargo.toolchain.RustToolchain
+import org.rust.ide.actions.ui.CargoNewCrateDialog
+import org.rust.openapiext.pathAsPath
+
+
+class RsCreateCrateAction : RunCargoCommandActionBase(CargoIcons.ICON) {
+    override fun actionPerformed(e: AnActionEvent) {
+        val dataContext = e.dataContext
+        val project = CommonDataKeys.PROJECT.getData(dataContext) ?: return
+        val root = CommonDataKeys.VIRTUAL_FILE.getData(dataContext) ?: return
```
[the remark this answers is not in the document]
Kobzol: Since the parent already enables the action only if there is a Cargo project and I handle running the action on both files and directories, it should be available always, or not?

src/test/kotlin/org/rust/ide/actions/RsCreateCrateActionTest.kt
```diff
@@ -0,0 +1,89 @@
...
+            "Failed to run `${RsCreateCrateAction::class.java.simpleName}` action"
+        }
+
+        action.actionPerformed(e)
+    }
+
+    private fun checkProject(binary: Boolean, name: String) {
+        val src = if (binary) "main" else "lib"
+
+        val main = cargoProjectDirectory.findFileByRelativePath("$name/src/$src.rs") ?: return Assert.fail()
```
Undin: We usually use kotlin `error()` instead of `fail()`

src/test/kotlin/org/rust/ide/actions/RsCreateCrateActionTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        val pkg = project.cargoProjects.findPackageForFile(main) ?: return Assert.fail()
+        val target = pkg.targets.first()
+        Assert.assertEquals(target.isBin, binary)
+        Assert.assertEquals(target.name, name)
```
Undin: `Assert` is useless here and above

src/test/kotlin/org/rust/ide/actions/RsCreateCrateActionTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.fileTree
+import org.rust.ide.actions.ui.CargoNewCrateSettings
+import org.rust.ide.actions.ui.CargoNewCrateUI
+
+class RsCreateCrateActionTest : RsWithToolchainTestBase() {
+    fun `test create with file context`() {
+        createProjectScaffold()
+
+        val main = cargoProjectDirectory.findFileByRelativePath("Cargo.toml")!!
```
Undin: Let's use the same style to handle unexpected null results
Currently, you use `!!` and `Assert.fail()`

src/test/kotlin/org/rust/ide/actions/RsCreateCrateActionTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        createProject(main, true, "crate1")
```
Undin: Maybe it's better to make `binary` flag to the end of arguments and pass it in named format?
```suggestion
        createProject(main, "crate1", binary = true)
```
It should improve readability

src/test/kotlin/org/rust/ide/actions/RsCreateCrateActionTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        checkProject(true, "crate1")
```
Undin: Hm, you always pass the same `name` and `binary` args to `createProject` and `checkProject`
Maybe it's better to merge them?

Kobzol:
@Undin fixed code style according to your comments.

Kobzol:
Done.

Undin:
@Kobzol Thanks a lot!
bors r+

Issue #2311, opened 2018-02-20
Improve AddStructFieldsFix


vlad20012:
Recent improvements were done by #1811 PR.

How `AddStructFieldsFix` can be improved further:
1. Better support &ref types; i.e. in struct like `struct S { f: &i32 }` insert `&` op before literal: `S { f: &0 }`
2. Support more container types: `Box`, `Rc`, `Arc`, `Cell`, `RefCell`, `UnsafeCell`, `Cow`, `Mutex` (etc?)
3. When filling struct recursively we may want to check that all the fields are visible (or else we can't instantiate the struct) and maybe try to find `new` method. Or this intention works pretty funny with something like `HashMap` :)
4. Take into account struct init shorthand syntax ([RFC 1682](https://github.com/rust-lang/rfcs/blob/master/text/1682-field-init-shorthand.md)) and maybe don't force default value for a field if there is a local variable with the same name (and type?)

matklad:
@vlad20012 wanna write some mentoring instructions for 4? Shouldn't be that hard to implement, and I miss the feature a lot. 

5. Like 4, but adjust types with autoreferences. I.e, if you have `struct Foo<'a> { quux: &'a mut Bar }` and you have a `quux: Bar` in scope, generate `quux: &mut quux`.

Issue #3495, opened 2019-03-07
Automatically insert imports for `Add remaining patterns`


ortem:
```Rust
// lib.rs
pub enum E { A, B }
pub struct S { pub e: E }
```

```Rust
// main.rs
use lib::S;

fn foo(s: S) {
    match s.e {
    }
}
```
should be transformed (via `Add remaining patterns`) into
```Rust
// main.rs
use lib::{S, E};

fn foo(s: S) {
    match s.e {
        E::A => {},
        E::B => {}
    }
}
```

Issue #3326, opened 2019-02-02
"Add missing fields" should not suggest arbitrary values


Riateche:
Currently, the "add missing fields" suggestion tries to put whatever it can into the initializer. The most annoying case is enums where it just picks the first variant, but zeros as default integer values are also pretty bad. 
```rust
enum X { X1, X2 }
struct A {
    i: i32,
    x: X,  
    s: String,
}
... 
        // filled by intellij-rust:
        A {
            i: 0,
            x: X::X1,
            s: "".to_string()
        }
```
This is very error-prone and against Rust's spirit. I can probably tolerate `None` for `Option` but it's also not great because I usually intend to initialize it in some other way and the plugin currently allows me to easily forget to do that.

I would very much prefer to not have anything filled at all:
```rust
        A {
            i: [cursor here],
            x: ,
            s: ,
        }
```
Or, if you always want to generate valid syntax, just use `unimplemented!()`:
```rust
        A {
            i: unimplemented!(),
            x: unimplemented!(),
            s: unimplemented!(),
        }
```

vlad20012:
Hmm, makes sense. We can leave current behavior for "Recursively fill struct fields" and simplify "fill struct fields" to just inset some placeholder like `unimplemented!()`

Anovs:
Honestly I like the current behavior. While I agree it's a bit more prone to mistakes, I feel that having to remove `unimplemented!()` to write the actual code for each field would be very tedious. Plus, but maybe it's specific to my use case, I regularly have to fill large structures where I need the default value for a lot of fields (usually `0`, `None` or `vec![]`), so in this case the current behavior behavior is really time saver.

If really you want to force the user to fill all the fields, then I think the "leave the fields empty" suggestion is better than the `unimplemented!()` one. But I still prefer the current behavior.

Riateche:
I would prefer leaving fields empty as well. I think it would make sense to make the behavior configurable (leave empty vs unimplemented vs default values) or even make a separate quick fix option for each variant (users can disable those they don't want).

Riateche:
Can we reopen this? The IDE still generates arbitrary values for new structs, and that effectively introduces hard to spot bugs in my code. It doesn't seem like the PR mentioned above is fixing this issue in any way.

Issue #3223, opened 2019-01-04
When types have the same name, full name should be printed


zlepper:
This somewhat a bug report, but probably mostly a feature request. 

When you have a type named e.g. `Error` and a function can return something that is also named `Error`, then you can get the very "helpful" message 'expected `Error`, found `Error`'. In these cases i think it would be very preferable to have the tooltip contain the entire path, so the error instead would be 'expected `some_lib::Error`, found `std::io::Error`'. 

It's only really a problem if the names are the exact same.

PR #3520, opened 2019-03-11
INT: add support for smart pointers, Default implementation, struct init shorthand and local variable to Add struct fields fix

Kobzol:
Hi,
I implemented support for smart pointers, Default impl and shorthand init syntax using local variables with same name and type for the Add struct fields fix (https://github.com/intellij-rust/intellij-rust/issues/2311), .

It would be also nice to auto-import the smart pointers if they are not in scope (if we prefill `Rc::new(0)` and `Rc` is not in scope, the user will have to use autoimport after the fix). Any hint on how could I do that? I guess I could run the autoimport action after the fix, but that seems like a heavyweight solution to me.

- [x] smart pointers
- [x] ? use Default impl
- [x] use local variables for initialization
- [ ] automatically import missing types for filled fields
- [ ] ? make the whole analysis dumber/simpler

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
@@ -12,11 +12,14 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import org.rust.ide.annotator.calculateMissingFields
 import org.rust.lang.core.psi.RsDefaultValueBuilder
+import org.rust.lang.core.psi.RsPatBinding
 import org.rust.lang.core.psi.RsPsiFactory
 import org.rust.lang.core.psi.RsStructLiteral
+import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.RsFieldsOwner
 import org.rust.lang.core.psi.ext.fields
 import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.resolve.processLocalVariables
 import org.rust.lang.core.resolve.ref.deepResolve
 import org.rust.openapiext.buildAndRunTemplate
 import org.rust.openapiext.createSmartPointer
@@ -50,7 +53,13 @@ class AddStructFieldsFix(
         val body = structLiteral.structLiteralBody
         val fieldsToAdd = calculateMissingFields(body, decl)
         val defaultValueBuilder = RsDefaultValueBuilder(decl.knownItems, body.containingMod, RsPsiFactory(project), recursive)
-        val addedFields = defaultValueBuilder.fillStruct(body, decl.fields, fieldsToAdd)
+
+        val addedFields = defaultValueBuilder.fillStruct(
+            body,
+            decl.fields,
+            fieldsToAdd,
+            RsDefaultValueBuilder.getVisibleBindings(startElement)
+        )
         editor?.buildAndRunTemplate(body, addedFields.mapNotNull { it.expr?.createSmartPointer() })
     }
 }
--- a/src/main/kotlin/org/rust/ide/inspections/RsFieldInitShorthandInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsFieldInitShorthandInspection.kt
@@ -28,12 +28,18 @@ class RsFieldInitShorthandInspection : RsLocalInspectionTool() {
                     override fun getFamilyName(): String = "Use initialization shorthand"
 
                     override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
-                        val field = descriptor.psiElement as RsStructLiteralField
-                        field.expr?.delete()
-                        field.colon?.delete()
+                        applyShorthandInit(descriptor.psiElement as RsStructLiteralField)
                     }
                 }
             )
+        }
+    }
+
+    companion object {
+        fun applyShorthandInit(field: RsStructLiteralField)
+        {
+            field.expr?.delete()
+            field.colon?.delete()
         }
     }
 }
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/InitializeWithDefaultValueFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/InitializeWithDefaultValueFix.kt
@@ -14,6 +14,7 @@ import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.ancestorOrSelf
 import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.resolve.processLocalVariables
 import org.rust.lang.core.types.declaration
 import org.rust.lang.core.types.type
 import org.rust.openapiext.buildAndRunTemplate
@@ -29,7 +30,8 @@ class InitializeWithDefaultValueFix(element: RsElement) : LocalQuickFixAndIntent
         val declaration = patBinding.ancestorOrSelf<RsLetDecl>() ?: return
         val semicolon = declaration.semicolon ?: return
         val psiFactory = RsPsiFactory(project)
-        val initExpr = RsDefaultValueBuilder(declaration.knownItems, declaration.containingMod, psiFactory).buildFor(patBinding.type)
+        val initExpr = RsDefaultValueBuilder(declaration.knownItems, declaration.containingMod, psiFactory, true)
+            .buildFor(patBinding.type, RsDefaultValueBuilder.getVisibleBindings(startElement as RsElement))
 
         if (declaration.eq == null) {
             declaration.addBefore(psiFactory.createEq(), semicolon)
--- a/src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
@@ -5,8 +5,10 @@
 
 package org.rust.lang.core.psi
 
+import org.rust.ide.inspections.RsFieldInitShorthandInspection
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.KnownItems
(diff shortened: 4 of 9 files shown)


Kobzol:
Added the possibility to initialize with an existing local variable.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -531,7 +531,7 @@ fun processLocalVariables(place: RsElement, processor: (RsPatBinding) -> Unit) {
         processLexicalDeclarations(scope, cameFrom, VALUES) { v ->
             val el = v.element
             if (el is RsPatBinding) processor(el)
-            true
+            false
```
Kobzol: I think that this should be false, since with true it only returns the first local variable.

src/main/kotlin/org/rust/ide/inspections/fixes/InitializeWithDefaultValueFix.kt
```diff
@@ -47,6 +49,14 @@ class InitializeWithDefaultValueFix(element: RsElement) : LocalQuickFixAndIntent
         }
     }
 
+    private fun getVisibleBindings(place: RsElement) : HashSet<RsPatBinding> {
```
Kobzol: This is duplicated in AddStructFieldsFix. Where is a good place to put it to avoid the duplication?

src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
```diff
@@ -19,32 +20,63 @@ class RsDefaultValueBuilder(
...
+                )
+
                 val item = ty.item
+                if (item in smartPointers) {
+                    return buildForSmartPtr(ty, bindings)
+                }
+
+                var default = this.defaultValue
+                if (item.implLookup.isDefault(ty)) {
+                    default = psiFactory.createAssocFunctionCall("Default", "default", listOf())
```
vlad20012: I don't think it's reasonable to insert `Default::default()` because it's implemented for most of the types but rarely used manually (mostly generic trait magic like `<T: default> T::default()` in some generic function).

vlad20012:
> Also changed "".to_string() to "".to_owned() (https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441).
Looks like you read the only first message in the thread, that is outdated now. Please, read [this, actual answer](https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/5)

src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
Kobzol: It seemed to me like a good default (no pun intended :slightly_smiling_face: ). I agree that it's probably not seen so often in user code, but I'd rather insert `Default::default()` than `()`, which will cause an error. So maybe use the default only as a last resort instead of `()` if possible, but otherwise prefer per-field initialization?

For me this fix would be mostly used in constructors (such as `T::new()`) and there I expect the local variable/parameter initialization will be most useful.

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
```diff
@@ -109,7 +109,7 @@ class KnownItems(
     val Debug: RsTraitItem? get() = findLangItem("debug_trait")
     val Box: RsStructOrEnumItemElement? get() = findLangItem("owned_box", "alloc")
     val Rc: RsStructOrEnumItemElement? get() = findItem("alloc::rc::Rc")
-    val Arc: RsStructOrEnumItemElement? get() = findItem("alloc::sync::Arc")
+    val Arc: RsStructOrEnumItemElement? get() = findItem("std::sync::Arc")
```
vlad20012: Hmm, does some test failed on this? It can be a problem in a project with `#![no_std]` and `extern crate alloc;`

Kobzol:
> > Also changed "".to_string() to "".to_owned() (https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441).
> > Looks like you read the only first message in the thread, that is outdated now. Please, read [this, actual answer](https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/5)

Ah, I thought that they want to solve it with `specialization`, which didn't land yet AFAIK. But they just overriden the trait method impl, now I get it :) So should I revert it back?

vlad20012:
> So should I revert it back?

Yep :)

src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
vlad20012: Maybe it's ok... anyway I'm going to add a separate quickfix to only add missing fields without initializers #3326

src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt
vlad20012: Yeah, `findItem("alloc::sync::Arc") ?: findItem("std::arc::Arc")` looks ok

src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
Kobzol: That's a good idea, personally I find the Add/Recursively add terminology a bit confusing. It could be changed to Add (just add field names and let the user fill them by moving the caret to each field successively) and Add + fill, which would work like the current recursive approach.

Kobzol:
Ok, I reverted the `to_owned() change` and used the `Arc` backup for older Rust. The autoimport of `Rc`, `Arc` etc. that I mentioned should be easier after https://github.com/intellij-rust/intellij-rust/pull/3099 lands.

src/main/kotlin/org/rust/ide/inspections/RsFieldInitShorthandInspection.kt
```diff
@@ -28,12 +28,16 @@ class RsFieldInitShorthandInspection : RsLocalInspectionTool() {
...
-                        field.expr?.delete()
-                        field.colon?.delete()
+                        applyFix(descriptor.psiElement as RsStructLiteralField)
                     }
                 }
             )
         }
     }
+
+    fun applyFix(field: RsStructLiteralField)
```
vlad20012: Seems to simple to extract. It's ok to copy-paste these 2 lines. Or at least let's move it to static method (with better name) instead of object method

Kobzol:
> Some tests are failed.
> Also please reduce number of commits by squashing "fixes" into commits where "bug" was made

Could you please point me to the failed tests? I see the CI green at the latest commit (https://github.com/intellij-rust/intellij-rust/commit/fdb0ae36097244e88a8cc03b293bc4c49a0a35c3). I planned to squash the whole PR into a single commit after it's done.

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
```diff
@@ -51,12 +54,21 @@ class AddStructFieldsFix(
...
         )
 
         if (editor != null && firstAdded != null) {
             val expr = firstAdded.expr
             if (expr != null) editor.caretModel.moveToOffset(expr.textOffset)
         }
     }
+
+    private fun getVisibleBindings(place: RsElement) : HashSet<RsPatBinding> {
+        val bindings = HashSet<RsPatBinding>()
```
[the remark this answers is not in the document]
Kobzol: Actually I just need to iterate though this, so I might as well convert it to a `List`. I wanted to ask whether I should extract it to a separate function (it's used both here and in the initialize with default value fix). If it's extracted into some utility space (I don't know where), then it would make more sense to use the `Map` as you say :)

src/main/kotlin/org/rust/ide/inspections/RsFieldInitShorthandInspection.kt
Kobzol: Moved it to a companion object and renamed to `applyShorthandInit`.

vlad20012:
> Could you please point me to the failed tests?
Ok, I started to write that message yesterday, when there really was some failed tests. I'm sorry!

src/main/kotlin/org/rust/ide/annotator/fixes/AddStructFieldsFix.kt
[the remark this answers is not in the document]
vlad20012: @ortem I remember something similar in a one of your quickfies. Could you, please, point that place?

src/main/kotlin/org/rust/lang/core/psi/RsDefaultValueBuilder.kt
Kobzol: So should this PR wait for #3326?

Kobzol:
@vlad20012 @mchernyavsky so is this now obsolete because of https://github.com/intellij-rust/intellij-rust/pull/3627? Or should it be kept as a separate fix (fill with manually entered values x fill with default values)?

mchernyavsky:
> so is this now obsolete because of #3627?

@Kobzol No, because by default placeholders from #3627 are filled with default values.

> automatically import missing types for filled fields

I think I will do it myself as part of https://github.com/intellij-rust/intellij-rust/pull/3099. Now I am working on https://github.com/intellij-rust/intellij-rust/issues/3586, https://github.com/intellij-rust/intellij-rust/issues/3495 and https://github.com/intellij-rust/intellij-rust/issues/3223.

Kobzol:
Ah sorry, I didn't realize that I actually DO index the variable set by name :slightly_smiling_face: I changed it to `Map` and moved the implementation to `RsDefaultValueBuilder` so that it is not duplicated. Squashed all commits into a6f2bb3 and rebased onto the current master.

Kobzol:
It conflicted with some changes in master, I rebased it again to resolve it.

mchernyavsky:
bors r=vlad20012

PR #3529, opened 2019-03-13
Add CLion's CPU profiler support

ortem:
Add [CLion's CPU profiler](https://www.jetbrains.com/help/clion/cpu-profiler.html) support (perf on Linux and DTrace on MacOS).

Huge thanks to @KirillTim for help!

### Perf
Screenshot 2019-03-14 at 11 41 04
Screenshot 2019-03-14 at 11 42 38

### DTrace
Screenshot 2019-03-19 at 19 09 08
Screenshot 2019-03-19 at 19 10 03

PR #3540, opened 2019-03-15
Add CLion's Valgrind support

ortem:
Add [CLion's Valgrind Memcheck](https://www.jetbrains.com/help/clion/memory-profiling-with-valgrind.html) support. Works on Linux and on MacOS.

I don't know how to properly install Valgrind on MacOS Mojave, but you can try [this solution](https://stackoverflow.com/a/54818930/10445191).

Merge after #3529 

Screenshot 2019-03-14 at 19 20 07

### DIFF
--- a/clion/src/191/main/kotlin/org/rust/clion/valgrind/RsValgrindConfigurationExtension.kt
+++ b/clion/src/191/main/kotlin/org/rust/clion/valgrind/RsValgrindConfigurationExtension.kt
@@ -1,0 +1,198 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.clion.valgrind
+
+import com.intellij.execution.ExecutionException
+import com.intellij.execution.configurations.CommandLineState
+import com.intellij.execution.configurations.GeneralCommandLine
+import com.intellij.execution.configurations.RunnerSettings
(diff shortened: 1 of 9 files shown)


Undin:
@ortem could you rebase onto master?

clion/src/191/main/kotlin/org/rust/clion/valgrind/RsValgrindConfigurationExtension.kt
```diff
@@ -0,0 +1,194 @@
...
+
+private val LOG = Logger.getInstance(RsValgrindConfigurationExtension::class.java.name)
+
+class RsValgrindConfigurationExtension : CargoCommandConfigurationExtension() {
+    override fun isApplicableFor(configuration: CargoCommandConfiguration): Boolean {
+        return true
+    }
+
+    override fun isEnabledFor(applicableConfiguration: CargoCommandConfiguration, runnerSettings: RunnerSettings?): Boolean =
+        SystemInfo.isLinux || SystemInfo.isMac
```
mchernyavsky: Maybe we want to enable it only on Linux?
CC @Undin

Undin: It's enabled for macOS for C++ so I think it's ok for us too even it works not perfect

PR #3543, opened 2019-03-16
add references from pattern fields,fix pattern parsing

rrevenantt:
This is fix for some minor problems I have found while working on another issue  
- added references/highlighting of full structure pattern fields(see added test)
- Fixed parser error for tuple being destructured as block struct. Here is valid example that resulted in error in plugin:
```rust
struct A(i32,i32);
fn main(){
    let A{0:a,1:b} = A(0,0);
}
  ```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsPatFieldFull.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsPatFieldFull.kt
@@ -1,0 +1,37 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import com.intellij.lang.ASTNode
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsPatFieldFull
+import org.rust.lang.core.psi.RsPatStruct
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/main/kotlin/org/rust/lang/core/psi/ext/RsPatFieldFull.kt [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+import org.rust.lang.core.psi.RsPatStruct
+import org.rust.lang.core.resolve.collectResolveVariants
+import org.rust.lang.core.resolve.processStructPatternFieldResolveVariants
+import org.rust.lang.core.resolve.ref.ResolveCacheDependency
+import org.rust.lang.core.resolve.ref.RsReference
+import org.rust.lang.core.resolve.ref.RsReferenceCached
+
+
+val RsPatFieldFull.parentStructPattern: RsPatStruct
+    get() = this.ancestors.filterIsInstance<RsPatStruct>().first()
```
mchernyavsky: `this` is redundant here. And I think this can be implemented using `ancestorStrict` (see `RsFieldDecl.parentStruct`).

rrevenantt:
Done.

mchernyavsky:
Thanks!
bors r+

PR #3550, opened 2019-03-18
INSP: add deprecated attribute inspection

isamborskiy:
Fulfilled #3564 improvement (the first and second checkboxes):

* Implement [deprecated attribute](https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md) inspection (for both `#[deprecated]` and `#[rustc_deprecated]` attributes)
* Support `#[allow(deprecated)]` attribute to suppress the corresponding warnings

Example:
deprecated-std-fun
deprecated-fun

**Note:** update `PathExpr` in RustParser by adding implementation of `RsOuterAttributeOwner` interface

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -1029,7 +1029,8 @@ private StructLiteralField_with_recover ::= StructLiteralField (',' | &'}') {
 private StructLiteralField_recover ::= !(identifier | INTEGER_LITERAL | ',' | '}' | '..' | '#')
 
 PathExpr ::= OuterAttr* PathGenericArgsWithColons {
- elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
+  implements = [ "org.rust.lang.core.psi.ext.RsOuterAttributeOwner" ]
+  elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
 WhileExpr ::= OuterAttr* LabelDecl? while Condition SimpleBlock {
--- a/src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
@@ -1,0 +1,82 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiElementVisitor
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+
+class RsDeprecationInspection : RsLocalInspectionTool() {
+    override fun getDisplayName() = "Deprecated item"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor = object : RsVisitor() {
+        override fun visitElement(ref: RsElement) {
+            // item is non-inline module declaration or not reference element
+            if (ref is RsModDeclItem || ref !is RsWeakReferenceElement) return
+
+            var original = ref.reference?.resolve() ?: return
+
+            val identifier = ref.referenceNameElement ?: ref
+            original = when (original) {
+                is RsFile -> original.declaration
+                is RsAbstractable -> original.superItem
+                else -> original
+            } ?: original
+
+            if (original is RsOuterAttributeOwner &&
+                checkAndRegisterAsDeprecated(identifier, original, holder)) return
+        }
+    }
+
+    private fun checkAndRegisterAsDeprecated(identifier: PsiElement, original: RsOuterAttributeOwner, holder: ProblemsHolder): Boolean =
+        original.outerAttrList
+            .mapNotNull { DeprecatedAttribute.from(it) }
+            .firstOrNull()
+            ?.let { deprecatedAttr ->
+                holder.registerProblem(identifier, deprecatedAttr.getMessage(identifier.text), ProblemHighlightType.LIKE_DEPRECATED)
+                true
+            } ?: false
+}
+
+private class DeprecatedAttribute(val note: String?, val since: String?) {
+
+    fun getMessage(item: String): String = buildString {
+        append("'$item' is marked as deprecated")
+        if (since != null) append(" since version $since")
+        if (note != null) append(" ($note)")
+    }
+
+    companion object {
+        private const val DEPRECATED_ATTR_NAME: String = "deprecated"
+        private const val RUSTC_DEPRECATED_ATTR_NAME: String = "rustc_deprecated"
+
+        private const val SINCE_PARAM_NAME: String = "since"
+        private const val NOTE_PARAM_NAME: String = "note"
+        private const val REASON_PARAM_NAME: String = "reason"
+
+        fun from(attr: RsOuterAttr): DeprecatedAttribute? {
+            fun List<RsMetaItem>.getByName(name: String): String? =
+                firstOrNull { name == it.identifier?.text }?.litExpr?.stringLiteralValue
+
+            fun RsOuterAttr.extract(noteParamName: String, sinceParamName: String): DeprecatedAttribute {
+                val params = metaItem.metaItemArgs?.metaItemList
+                val note = params?.getByName(noteParamName)
+                val since = params?.getByName(sinceParamName)
+                return DeprecatedAttribute(note, since)
+            }
+
+            val identifier = attr.metaItem.identifier?.text
+            return when (identifier) {
+                DEPRECATED_ATTR_NAME -> attr.extract(NOTE_PARAM_NAME, SINCE_PARAM_NAME)
+                RUSTC_DEPRECATED_ATTR_NAME -> attr.extract(REASON_PARAM_NAME, SINCE_PARAM_NAME)
+                else -> null
+            }
+        }
+    }
+}
--- a/src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
@@ -5,11 +5,15 @@
 
 package org.rust.ide.inspections
 
-import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemHighlightType.LIKE_DEPRECATED
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiElementVisitor
-import org.rust.lang.core.psi.*
+import com.intellij.psi.util.parents
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.RsMetaItem
+import org.rust.lang.core.psi.RsModDeclItem
+import org.rust.lang.core.psi.RsVisitor
 import org.rust.lang.core.psi.ext.*
 
 class RsDeprecationInspection : RsLocalInspectionTool() {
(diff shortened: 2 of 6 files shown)


src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.impl.RsModDeclItemImpl
+
+class RsDeprecatedAttributeInspection : RsLocalInspectionTool() {
+    override fun getDisplayName() = "Deprecated item"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor = object : RsVisitor() {
+        override fun visitElement(ref: RsElement) {
+            // non-inline module declaration
+            if (ref is RsModDeclItemImpl) return
```
Undin: You shouldn't use impl psi classes

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+
+            ref.reference?.resolve()?.let {
+                val identifier = if (ref is RsReferenceElement) ref.referenceNameElement else ref
```
Undin: Let's check that ref is `RsReferenceElement` before name resolution to avoid redundant `resolve` call

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+
+class RsDeprecatedAttributeInspection : RsLocalInspectionTool() {
+    override fun getDisplayName() = "Deprecated item"
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor = object : RsVisitor() {
+        override fun visitElement(ref: RsElement) {
+            // non-inline module declaration
+            if (ref is RsModDeclItemImpl) return
+
+            ref.reference?.resolve()?.let {
```
Undin: I'd prefer to unwrap this `let` into
```kotlin
val element = ref.reference?.resolve() ?: return
// do something with resolved element
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+                val identifier = if (ref is RsReferenceElement) ref.referenceNameElement else ref
+                val original = if (it is RsFile) it.declaration ?: it else it
```
Undin: Maybe it's better to get item where you want to search `deprecated` attribute first
and only after it call `checkAndRegisterAsDeprecated `
Note, you can't add `deprecated` attr in trait impl block
```rust
trait Foo {
    type Bar;
    const QWE: i32;
}

struct Baz;

impl Foo for Baz {
    type Bar = ();
    #[deprecated] // compilation error
    const QWE: i32 = 123;
}
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiElementVisitor
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+
+class RsDeprecatedAttributeInspection : RsLocalInspectionTool() {
```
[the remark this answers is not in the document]
isamborskiy: Renamed

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
isamborskiy: Oops... that was a typo :)
Fixed

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
isamborskiy: Fixed

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
isamborskiy: Okay, done

src/main/kotlin/org/rust/ide/inspections/RsDeprecatedAttributeInspection.kt
isamborskiy: Agree, fixed

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+    companion object {
+        private const val ALLOW_ATTR_NAME: String = "allow"
+        private const val DEPRECATED_PARAM_NAME: String = "deprecated"
+    }
+}
+
+private class DeprecatedAttribute(val note: String?, val since: String?) {
+
+    fun getMessage(item: String): String = buildString {
+        append("'$item' is marked as deprecated")
```
Undin: Similar annotations for Java and Kotlin write just `'$item' is deprecated`
image

Undin: And we usually use ` in such cases
image

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+        if (since != null) append(" since version $since")
```
Undin: `version` looks redundant here

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+        if (note != null) append(" ($note)")
```
Undin:
```suggestion
        if (note != null) append(": $note")
```
Let's use the same style as rustdoc uses: 
image

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+    private fun checkAndRegisterAsDeprecated(identifier: PsiElement, original: RsOuterAttributeOwner, holder: ProblemsHolder) {
+        original.outerAttrList
+            .mapNotNull { DeprecatedAttribute.from(it) }
+            .firstOrNull()
+            ?.let {
+                holder.registerProblem(identifier, it.getMessage(identifier.text), ProblemHighlightType.LIKE_DEPRECATED)
+            }
+    }
+
+    private fun PsiElement.hasAllowDeprecatedAttribute(): Boolean = this is RsOuterAttributeOwner &&
```
Undin: Let's move this code into `org.rust.lang.core.psi.ext.QueryAttributes`. This class contains some usefull methods

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+                firstOrNull { name == it.identifier?.text }?.litExpr?.stringLiteralValue
+
+            fun RsOuterAttr.extract(noteParamName: String, sinceParamName: String): DeprecatedAttribute {
+                val params = metaItem.metaItemArgs?.metaItemList
+                val note = params?.getByName(noteParamName)
+                val since = params?.getByName(sinceParamName)
+                return DeprecatedAttribute(note, since)
+            }
+
+            val identifier = attr.metaItem.identifier?.text
```
Undin: Let's move all search-related code into `org.rust.lang.core.psi.ext.QueryAttributes`
Something like this:
```kotlin
fun findDeprecatedAttribute(): RsMetaItem? {
    // some code
}
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+        override fun visitElement(ref: RsElement) {
+            // item is non-inline module declaration or not reference element
+            if (ref is RsModDeclItem || ref !is RsWeakReferenceElement) return
+
+            var original = ref.reference?.resolve() ?: return
+
+            val identifier = ref.referenceNameElement ?: ref
+            original = when (original) {
+                is RsFile -> original.declaration
+                is RsAbstractable -> original.superItem
```
Undin: We should call `superItem` only when `RsAbstractable.owner.isImplOrTrait`

src/test/kotlin/org/rust/ide/inspections/RsDeprecationInspectionTest.kt
```diff
@@ -0,0 +1,225 @@
...
+        pub fn foo() {
+        }
+
+        fn main() {
+            #[allow(deprecated)]
+            foo();
+        }
+    """)
+
+    fun `test allow deprecated item in module`() = checkByText("""
```
Undin: Let's add test when deprecation warnings suppressed in the whole file
In this case, you should add `#![allow(deprecated)]` at the beginning of file

src/main/grammars/RustParser.bnf
```diff
@@ -1029,7 +1029,8 @@ private StructLiteralField_with_recover ::= StructLiteralField (',' | &'}') {
 private StructLiteralField_recover ::= !(identifier | INTEGER_LITERAL | ',' | '}' | '..' | '#')
 
 PathExpr ::= OuterAttr* PathGenericArgsWithColons {
- elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
+  implements = [ "org.rust.lang.core.psi.ext.RsOuterAttributeOwner" ]
```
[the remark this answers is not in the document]
isamborskiy: Agree, let's make it as separate PR

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
isamborskiy: I just replaced this method with a general analog the method from `org.rust.lang.core.psi.ext.QueryAttributes`

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
isamborskiy: Done, but we need fallback like this: 
```kotlin
... original.owner.isImplOrTrait -> original.superItem ?: original
```

Otherwise, this case doesn't work:
```rust
struct Foo;

impl Foo {
    #[deprecated]
    fn new() -> Foo {
        Foo { }
    }
}

fn main() {
    let foo = Foo::<warning descr="`new` is deprecated">new</warning>();
}
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+        private const val DEPRECATED_ATTR_NAME: String = "deprecated"
+
+        private const val SINCE_PARAM_NAME: String = "since"
+        private const val NOTE_PARAM_NAME: String = "note"
+        private const val REASON_PARAM_NAME: String = "reason"
+
+        // #[deprecated] or #[rustc_deprecated]
+        fun from(deprecatedMetaItem: RsMetaItem): DeprecatedAttribute {
+            fun List<RsMetaItem>.getByName(name: String): String? =
+                firstOrNull { name == it.identifier?.text }?.litExpr?.stringLiteralValue
```
Undin: Let's use `RsMetaItem.name` and `RsMetaItem.value` extension properties here instead of `it.identifier?.text` and `litExpr?.stringLiteralValue` not only that we already have such code but also because they use stubs if it's possible, i.e. they don't reparse not opened files
Also, it would be cool to add a test to check that we don't load ast for all files
You need to create a multifile test where a deprecated item is located not in a file where you are using and just call `org.rust.RsTestBase#checkAstNotLoaded` before inspection run

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+
+            if (ref.parents().none { it.hasAllowDeprecatedAttribute() }) {
+                checkAndRegisterAsDeprecated(identifier, targetElement, holder)
+            }
+        }
+    }
+
+    private fun checkAndRegisterAsDeprecated(identifier: PsiElement, original: PsiElement, holder: ProblemsHolder) {
+        if (original is RsOuterAttributeOwner) {
+            QueryAttributes(original)
```
Undin:
```suggestion
            original.queryAttributes
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+                .deprecatedAttribute
+                ?.let { DeprecatedAttribute.from(it) }
```
Undin: Let's avoid `?.let`. It makes code much harder to read.
You can write something like
```kotin
if (original is RsOuterAttributeOwner) {
    val attr = QueryAttributes(original).deprecatedAttribute ?: return
    holder.registerProblem(identifier, DeprecatedAttribute.from(attr).getMessage(identifier.text), LIKE_DEPRECATED)
}
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+    private fun PsiElement.hasAllowDeprecatedAttribute(): Boolean = this is RsDocAndAttributeOwner &&
+        QueryAttributes(this).hasAttributeWithArg(ALLOW_ATTR_NAME, DEPRECATED_ARG_NAME)
+
+    companion object {
+        private const val ALLOW_ATTR_NAME: String = "allow"
+        private const val DEPRECATED_ARG_NAME: String = "deprecated"
+    }
+}
+
+private class DeprecatedAttribute(val note: String?, val since: String?) {
```
Undin: Why do we need this class at all?
Looks like we can just create a private function for message generation

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+        if (original is RsOuterAttributeOwner) {
+            QueryAttributes(original)
+                .deprecatedAttribute
+                ?.let { DeprecatedAttribute.from(it) }
+                ?.let { holder.registerProblem(identifier, it.getMessage(identifier.text), LIKE_DEPRECATED) }
+        }
+    }
+
+    private fun PsiElement.hasAllowDeprecatedAttribute(): Boolean = this is RsDocAndAttributeOwner &&
+        QueryAttributes(this).hasAttributeWithArg(ALLOW_ATTR_NAME, DEPRECATED_ARG_NAME)
```
Undin:
```suggestion
        original.queryAttributes.hasAttributeWithArg(ALLOW_ATTR_NAME, DEPRECATED_ARG_NAME)
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+        override fun visitElement(ref: RsElement) {
+            // item is non-inline module declaration or not reference element
+            if (ref is RsModDeclItem || ref !is RsWeakReferenceElement) return
+
+            val original = ref.reference?.resolve() ?: return
+            val identifier = ref.referenceNameElement ?: return
+
+            val targetElement = when {
+                original is RsFile -> original.declaration
+                original is RsAbstractable && original.owner.isImplOrTrait -> original.superItem ?: original
```
Undin: oops, sorry. My comment was misleading
You need to use `isTraitImpl` instead of `isImplOrTrait`
So, you can write it in the following way
```kotlin
val targetElement = when (original) {
    is RsFile -> original.declaration
    is RsAbstractable -> if (original.owner.isTraitImpl) original.superItem else original
    else -> original
} ?: return
```

src/main/kotlin/org/rust/ide/inspections/RsDeprecationInspection.kt
Undin: Answered in https://github.com/intellij-rust/intellij-rust/pull/3550#discussion_r268511363

src/test/kotlin/org/rust/ide/inspections/RsDeprecationInspectionTest.kt
```diff
@@ -0,0 +1,255 @@
...
+            println!("TEST")
+        }
+
+        fn main() {
+            foo();
+        }
+    """)
+
+    fun `test AST is not loaded when deprecated item in another file`() {
+        checkAstNotLoaded(VirtualFileFilter { "foo.rs" == it.name })
```
Undin: Unfortunately, it doesn't work for highlighting checking (including inspection tests) because inside such tests `VirtualFileFilter` is changed by test code itself. So at this moment, it tests nothing
Let's replace it with temporary hack
```suggestion
        (myFixture as CodeInsightTestFixtureImpl).setVirtualFileFilter { !it.path.endsWith("main.rs") }
```

Undin:
I'll fix test myself

PR #3556, opened 2019-03-18
TY&RES: Fix UFCS-like type-qualified paths (`<T as Foo>::bar`)

vlad20012:
```rust
struct S;
impl S {
    fn foo() {}
}    //X
fn main () {
    <S>::foo;
}      //^
```

(the first commit is only refactoring)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -7,6 +7,7 @@
 
 package org.rust.lang.core.resolve
 
+import com.intellij.codeInsight.completion.CompletionUtil
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -339,90 +337,112 @@ fun processPathResolveVariants(lookup: ImplLookup, path: RsPath, isCompletion: B
...
 
+/** `<T as Trait>::Item` or `<T>::Item` */
+private fun processExplicitTypeQualifiedPathResolveVariants(
+    lookup: ImplLookup,
+    processor: RsResolveProcessor,
+    ns: Set<Namespace>,
+    typeQual: RsTypeQual
+): Boolean {
+    val trait = typeQual.traitRef?.resolveToBoundTrait
+        // TODO this it a hotfix for completion test `test associated type in explicit UFCS form`.
```
Undin: I think it's more a little hack than hotfix

vlad20012:
bors r=undin

vlad20012:
bors r=undin

PR #3567, opened 2019-03-20
RES: Fix non-absolute std-qualified path resolution

### LINKED ISSUE #3421: Rust 2018 path resolution doesn't work in sub modules
## Environment

* **Intellij-Rust plugin version:** v0.2.92.2116-183
* **Rust toolchain version:** rustc 1.33.0-nightly (bf669d1e3 2019-01-25)
* **IDE name and version:** IntelliJ IDEA 2018.3.4 Build #IU-183.5429.30, built on January 29, 2019
* **Operating system:** MacOS Sierra

## Problem description

Rust 2018 edition path resolution doesn't work in sub-modules:

image

However it works if I...
- write `::std::ops::Deref`, or
- write `use std::ops::Deref` and `impl Deref for ..`, or
- Move it into the crate root (`main.rs` or `lib.rs`).

Notice that this additionally causes `Deref` inferring not working:

image

## Steps to reproduce

`main.rs`:

```rust
mod x;

fn main() {
    let bar = x::Outer::new(10);
    println!("{}", bar.field1);
}
```

`x.rs`:

```rust
pub struct Inner {
    pub field1: i64,
}

pub struct Outer(Inner);

impl std::ops::Deref for Outer {
    type Target = Inner;

    fn deref(&self) -> &Inner {
        &self.0
    }
}

impl Outer {
    pub fn new(v: i64) -> Outer {
        Outer(Inner { field1: v })
    }
}
```

`Cargo.toml`:

```toml
[package]
name = "my_test"
version = "0.1.0"
edition = "2018"
```

vlad20012:
Related to "Extern Prelude" and 2018 edition (but should work in 2015 edition as well)
Fixes #3421

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
@@ -92,31 +92,33 @@ fun processItemDeclarations(
         }
     }
 
-    if (Namespace.Types in ns) {
-        if (scope is RsFile && scope.isCrateRoot && withPrivateImports) {
-            // Rust injects implicit `extern crate std` in every crate root module unless it is
-            // a `#![no_std]` crate, in which case `extern crate core` is injected. However, if
-            // there is a (unstable?) `#![no_core]` attribute, nothing is injected.
-            //
-            // https://doc.rust-lang.org/book/using-rust-without-the-standard-library.html
-            // The stdlib lib itself is `#![no_std]`, and the core is `#![no_core]`
-            when (scope.attributes) {
-                RsFile.Attributes.NONE ->
-                    if (processor.lazy(STD) { scope.findDependencyCrateRoot(STD) }) return true
-
-                RsFile.Attributes.NO_STD ->
-                    if (processor.lazy(CORE) { scope.findDependencyCrateRoot(CORE) }) return true
-
-                RsFile.Attributes.NO_CORE -> Unit
-            }
-        }
-    }
-
     for (speck in itemImports) {
         check(speck.useGroup == null)
         val path = speck.path ?: continue
         val name = speck.nameInScope ?: continue
         if (processMultiResolveWithNs(name, ns, path.reference, processor)) return true
+    }
+
+    if (Namespace.Types in ns && scope is RsMod && withPrivateImports) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
```diff
@@ -86,33 +86,32 @@ fun processItemDeclarations(
...
-
     for (speck in itemImports) {
         check(speck.useGroup == null)
         val path = speck.path ?: continue
         val name = speck.nameInScope ?: continue
         if (processMultiResolveWithNs(name, ns, path.reference, processor)) return true
     }
 
+    if (Namespace.Types in ns && scope is RsMod && withPrivateImports) {
+        // Rust injects implicit `extern crate std` in every module unless it is
```
Undin: Change "in every crate root module" to "in every module" is wrong
The corresponding documentation:

image

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
```diff
@@ -86,33 +86,32 @@ fun processItemDeclarations(
...
+        // a `#![no_std]` crate, in which case `extern crate core` is injected. However, if
+        // there is a (unstable?) `#![no_core]` attribute, nothing is injected.
+        //
+        // https://doc.rust-lang.org/book/using-rust-without-the-standard-library.html
+        // The stdlib lib itself is `#![no_std]`, and the core is `#![no_core]`
+        if (!scope.isCrateRoot && processor(ScopeEvent.IMPLICIT_CRATES)) return true
```
Undin: Hm, could you remind why we handle stdlib implicit crates here but all other code related to extern crates is located in [NameResolution](https://github.com/intellij-rust/intellij-rust/blob/31e7b1303e79637a9e95b36f7b04aa82c13f2582/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt#L470-L484)?
Maybe it possible to handle them in one place?

vlad20012: Yeah, we should move them here. I don't sure I should do it in a separate PR or not

vlad20012:
I fixed the comment. Lets merge it, and then I'll try to fix "unified paths"

PR #3569, opened 2019-03-21
PRJ: Implement project settings migration via XML modification

mchernyavsky:


src/main/kotlin/org/rust/cargo/project/settings/impl/SerializationUtils.kt
```diff
@@ -0,0 +1,89 @@
...
+const val USE_SKIP_CHILDREN: String = "useSkipChildren"
+
+// Legacy properties needed for migration
+const val USE_CARGO_CHECK_ANNOTATOR: String = "useCargoCheckAnnotator"
+const val CARGO_CHECK_ARGUMENTS: String = "cargoCheckArguments"
+const val EXPAND_MACROS: String = "expandMacros"
+
+fun Element.updateToCurrentVersion() {
+    updateToVersionIfNeeded(2) {
+        renameOption(USE_CARGO_CHECK_ANNOTATOR, RUN_EXTERNAL_LINTER_ON_THE_FLY)
```
Undin: Hm, we already can have users with current weird migration scheme so we can't just rename option
A user can already have `RUN_EXTERNAL_LINTER_ON_THE_FLY` field and this renaming creates a new option with the same name

mchernyavsky: But a user cannot have `USE_CARGO_CHECK_ANNOTATOR` and `RUN_EXTERNAL_LINTER_ON_THE_FLY` at the same time because we set the `useCargoCheckAnnotator` to default:
https://github.com/intellij-rust/intellij-rust/blob/5b9068d6f7d8aa4fb4edc13af4fed1862116dfcd/src/main/kotlin/org/rust/cargo/project/settings/impl/RustProjectSettingsServiceImpl.kt#L43-L46

Undin: Oh, yeah. You are right

src/test/kotlin/org/rust/cargo/project/RustProjectSettingsServiceTest.kt
```diff
@@ -55,4 +58,41 @@ class RustProjectSettingsServiceTest : LightPlatformTestCase() {
...
+              <option name="useCargoCheckAnnotator" value="true" />
+            </RustProjectSettings>
+        """.trimIndent()
+        service.loadState(elementFromXmlString(text).deserialize())
+
+        @Language("XML")
+        val expected = """
+            <RustProjectSettings>
+              <option name="autoUpdateEnabled" value="true" />
+              <option name="compileAllTargets" value="true" />
```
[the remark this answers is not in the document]
mchernyavsky: Hmm, it is already fixed (in this PR).

PR #3575, opened 2019-03-22
TY: Fix bindings extraction for pat struct

mchernyavsky:
@ortem I didn't update the borrow checker, but I think it should also take into account [this](https://github.com/intellij-rust/intellij-rust/pull/3543) syntax.


mchernyavsky:
bors retry

PR #3576, opened 2019-03-22
Add "impl" & "implt" live templates

vlad20012:
Peek 2019-03-22 19-03


Undin:
Oh, forgot about tests
Let's add tests for new live templates

vlad20012:
Added tests

PR #3595, opened 2019-03-25
ANN: fix highlighting of keywords in paths (`crate`/`Self`)

vlad20012:
Before:
image
After:
image


ice1000:
Why these keywords cannot be distinguished in the lexer (so we can highlight them in the `SyntaxHighlighter` impl)?

vlad20012:
@ice1000 they are. But there is a bug that we override such lexer highlighting by the annotator, and this PR fixes it

PR #3598, opened 2019-03-25
MACRO: navigate to macro call during all nav actions

vlad20012:
(including go-to-super and go-to-implementation)
I've used GeneratedSourcesFilter extension which also
allows to refine navigation target


vlad20012:
Oh, it works only when a macro expansion is a physical file, so it doesn't works #3015

vlad20012:
Moved to #3015

PR #3602, opened 2019-03-26
Fix possible performance regression in type inference

vlad20012:
The bug was introduced in #3556
(the removed code was a kind of legacy that we forget to remove)


vlad20012:
bors retry

PR #3612, opened 2019-03-28
INSP: use stub in inspections if possible

Undin:
Before these changes, `Unresolved reference` and `Different impl member order from the trait` inspections always used AST instead of stubs that led to reparsing all files required for analysis
Now, they use stub in these cases.
Also, file tree based annotation tests check that inspections don't load files (except opened one, of course) while analysis by default


Undin:
bors retry

PR #3615, opened 2019-03-28
Don't affect psi of other languages while renaming

### LINKED ISSUE #3614: RsRenameProcessor shouldn't accept elements from other languages
* **Intellij-Rust plugin version:** v0.2.95.2120-191
* **IDE name and version**: Intellij IDEA Ultimate 2019.1

RsRenameProcessor shouldn't accept elements from other languages [here](https://github.com/intellij-rust/intellij-rust/blob/1409910b8f28133bc1d03c0ca2bfda7ea9a6e48a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt#L18):
```
override fun canProcessElement(element: PsiElement): Boolean = element !is RsTupleFieldDecl
```

RenameProcessors should return true from `canProcessElement` only for elements that they can rename, otherwise chaos ensues. The platform picks the first available processor in `com.intellij.refactoring.rename.RenamePsiElementProcessor#forElement`.
Other processors can have language-specific logic like collecting implementors/inheriors for their language-specific elements.

Currently, having the rust plugin enabled causes invalid rename in JS because `com.intellij.lang.javascript.refactoring.JSDefaultRenameProcessor` has custom logic related to JS properties:

```js
//JS
const userId = 'john'; //rename 'userId here'
const response = { userId };
```
Expected:
```js
//JS
const johnId = 'john';
const response = { userId: johnId };
```

A user found that the problem is caused by the Rust plugin: https://jbs.zendesk.com/agent/tickets/2035133
A similar issue is not reproducible for me with Java-specific rename features, probably because processors from intellij-community are somehow ordered before plugins, but order between different plugins is not determined.

ortem:
Fixes #3614

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
@@ -10,12 +10,15 @@ import com.intellij.openapi.util.Pass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.search.SearchScope
 import com.intellij.refactoring.rename.RenamePsiElementProcessor
-import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsLabel
+import org.rust.lang.core.psi.RsLabelDecl
+import org.rust.lang.core.psi.RsLifetime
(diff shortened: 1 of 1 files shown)


mchernyavsky:
bors retry

PR #3636, opened 2019-04-02
DBG: Add HashMap/HashSet formatter for lldb

ortem:
`HashMap`

Screenshot 2019-04-02 at 18 40 48

`HashSet`

Screenshot 2019-04-02 at 18 39 50


ortem:
> What minimal version of rust should I use to have correct hashmap/hashset rendering in debugger?

I think it's `1.33.0`

Issue #3628, opened 2019-04-01
Enable new macro expansion system by default


vlad20012:
We decided to merge the new macro expansion system (introduced in #3015) but not enable it by default. There is an option to enable it:
![image](https://user-images.githubusercontent.com/2539310/58436688-4f633c80-80cf-11e9-9bcf-0f81bf30f46b.png)

Why new expansion engine is better?
* it makes possible to support `impl` blocks and `mod` declarations generated by declarative macros, i.e. if some methods are inside `impl` block generated by macro call, the plugin makes name resolution and completions for these methods as expected. Ability to make proper name resolution for `mod` declarations provides proper name resolution and completion for crates like tokio, async-std, reqwest, etc.
* it significantly decreases memory consumption that should speed up name resolution/completion operations on average
* if an item is declared in some macro call, the plugin can navigate you inside macro call on `Go to definition` action
* find usages works for macro calls arguments
* the plugin provides better completion inside macro calls
* the plugin provides [semantic highlighting](https://github.com/intellij-rust/intellij-rust/pull/4640) inside macro calls

To enable by default we have to handle:
- [x] UI freezes & memory consumption during expansion (IDEA's default 768mb should be enough) #3653
- [ ] UI blinking during expansion (dumb/smart mode switches)
- [x] Internal files with expansions are still visible sometimes (e.g. ctrl+shift+B)
- [x] Make a better progress bar filling #3945
- [x] Disable inspections/annotators while expanding macros #3660
- [x] https://github.com/intellij-rust/intellij-rust/issues/3692
- [x] Compress expansion files #3968
- [x] Speed up indexing #4355
- [ ] Long scanning phase (`O(project file count)`) after creation of new expansion file
- [ ] Long validation of macro expansions after project re-open or git branch switch or cargo deps change etc

After enabling it by default, these issues should be closed: #955 #2128 #2383 #2615 #2625 #2787 #2902 #3006 #3087 #3155 #3238 #3259 #3468 #3487 #3560 #3587 #3590 #3628 #3859 #4063 #4153 #4224 #4310 #4314 #4328 #4340 #4389 #4519 #4585 #4614 #4627 #4826 #5088 #5127 #5154 #5246 #5248 #5263 #5278 #5300 #5346 #5461 #5686 ...TODO

panstromek:
There are still some hiccups that seems to be related to the new expansion system, even though I am not sure.

[...]

Issue #2898, opened 2018-10-05
Support "Go to Declaration" inside macros


HighCommander4:
It would be very useful if "Go to Declaration" worked inside macro invocations.

For example, in the following macro invocation:

```rust
error_chain! {
    foreign_links {
        Json(serde_json::Error)
    }
}
````

if you would be nice if "Go to Declaration" worked on `serde_json::Error`.

Eclipse CDT supports this for C/C++ macros. Their general implementation strategy is:

* When expanding macros, keep track of a mapping between tokens in the source text, and where in the macro expansion they end up.
* When performing "Go to Declaration" on a token inside a macro invocation, use the mapping to look up the corresponding token in the (final) expansion. (If there are multiple such tokens in the final expansion, can try each one, or pick the first, or something.)
* Perform "Go to Declaration" as if it had been invoked on the token in the expansion.

vlad20012:
Yep, I'm going to try this approach. When will be free time =)

vlad20012:
Yeah, this approach works great! I've just made a prototype and it works as expected. I'll finish it right after #3015 landing.

vlad20012:
Will be very soon. All preparations are done #3640

PR #3640, opened 2019-04-03
MACRO: add range mappings between macro call and expansion

vlad20012:
Now we know which element in macro call is expanded to which (real) element in expansion.
Currently it is used only to provide more precise navigation target when navigating to something expanded from a macro:
Peek 2019-04-03 23-36

But soon I'll use it to support go-to-declaration from macro call #2898 (and for other navigation actions), and then this will be used to support hygiene (and so type inference for custom expression-like macros), and in the future this will allow to support highlighting and completion inside macro calls!

This works only if new macro expansion engine is enabled (see #3628 for details).

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -664,7 +664,7 @@ private fun calcStubIndex(psi: StubBasedPsiElement<*>): Int {
 private val VirtualFile.fileId: Int
     get() = (this as VirtualFileWithId).id
 
-// Use WeakReference it's relatively easy to load RangeMap again
+/** We use [WeakReference] because uncached [loadRangeMap] is quite cheap */
 private val MACRO_RANGE_MAP_CACHE_KEY: Key<WeakReference<RangeMap>> = Key.create("MACRO_RANGE_MAP_CACHE_KEY")
 private val RANGE_MAP_ATTRIBUTE = FileAttribute(
     "org.rust.macro.RangeMap",
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -10,9 +10,11 @@ import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Key
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapi.vfs.VirtualFileWithId
+import com.intellij.openapi.vfs.newvfs.FileAttribute
 import com.intellij.psi.PsiAnchor
 import com.intellij.psi.PsiDocumentManager
 import com.intellij.psi.PsiManager
@@ -37,6 +39,7 @@ import org.rust.stdext.HashCode
 import java.io.DataInputStream
 import java.io.DataOutputStream
 import java.io.FileNotFoundException
+import java.lang.ref.WeakReference
 import java.nio.file.Files
 import java.nio.file.Path
 import java.util.zip.DeflaterOutputStream
@@ -110,7 +113,8 @@ class ExpandedMacroStorage(val project: Project) {
         oldInfo: ExpandedMacroInfo,
         callHash: HashCode?,
         defHash: HashCode?,
-        expansionFile: VirtualFile?
+        expansionFile: VirtualFile?,
+        ranges: RangeMap?
     ): ExpandedMacroInfo {
         checkWriteAccessAllowed()
 
@@ -135,6 +139,10 @@ class ExpandedMacroStorage(val project: Project) {
         }
 
         newInfo.expansionFile?.let { getOrCreateSourceFile(it) }
+
+        if (newInfo.expansionFile != null && ranges != null) {
+            newInfo.expansionFile.writeRangeMap(ranges)
+        }
 
         return newInfo
     }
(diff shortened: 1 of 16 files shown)


vlad20012:
Ready for review

src/main/kotlin/org/rust/lang/core/macros/RangeMap.kt
```diff
@@ -0,0 +1,177 @@
...
+import org.rust.lang.core.psi.ext.expansion
+import org.rust.lang.core.psi.ext.startOffset
+import org.rust.stdext.optimizeList
+import org.rust.stdext.readVarInt
+import org.rust.stdext.writeVarInt
+import java.io.DataInputStream
+import java.io.DataOutputStream
+
+// TODO optimize it using sorted array and binary search
+class RangeMap private constructor(private val ranges: List<Pair<TextRange, TextRange>>) {
```
Undin: Let's add javadoc about `ranges` because it's not clear what these ranges mean without reading their usages

src/main/kotlin/org/rust/lang/core/macros/RangeMap.kt
```diff
@@ -0,0 +1,177 @@
...
+            offset >= s.startOffset && offset < s.endOffset
+        }?.let { (d, s) ->
+            d.startOffset + (offset - s.startOffset)
+        }
+    }
+
+    fun mapOffsetFromCallBodyToExpansion(offset: Int): List<Int> {
+        return ranges.filter { (s, _) ->
+            offset >= s.startOffset && offset < s.endOffset
+        }.map { (s, d) ->
```
Undin: Let's use more meaningful names for ranges instead of `s` and `d` and use the same name in `mapOffsetFromExpansionToCallBody` and `mapOffsetFromCallBodyToExpansion`

src/main/kotlin/org/rust/lang/core/macros/RangeMap.kt
```diff
@@ -0,0 +1,177 @@
...
+            d.startOffset + (offset - s.startOffset)
+        }
+    }
+
+    companion object {
+        val EMPTY: RangeMap = CompactRangeMap.EMPTY.unpack()
```
Undin: Why not just `val EMPTY: RangeMap = RangeMap(emptyList())` ?

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -58,13 +56,31 @@ val RsExpandedElement.expandedFromRecursively: RsMacroCall?
...
+/**
+ * If receiver element is inside a macro expansion, returns the element inside the macro call
+ * we should navigate to (or the macro call itself if there isn't such element inside a macro call).
+ * Returns null if the element isn't inside a macro expansion
+ */
+fun PsiElement.findNavigationTargetIfMacroExpansion(): PsiElement? =
+    /** @see RsNamedElementImpl.getTextOffset */
+    ((this as? RsNameIdentifierOwner)?.nameIdentifier ?: this).findElementExpandedFrom()
+        // fallback to macro call
+        ?: this.findMacroCallExpandedFrom()?.path
```
Undin:
```suggestion
        ?: findMacroCallExpandedFrom()?.path
```

src/main/kotlin/org/rust/lang/core/macros/RangeMap.kt
```diff
@@ -0,0 +1,177 @@
...
+        check(length > 0) { "length should be grater than 0; got: $length" }
+    }
+}
+
+/**
+ * If receiver element is inside a macro expansion, returns the leaf element inside the macro call
+ * from which the first token of this element is expanded. Always returns an element inside a root
+ * macro call, i.e. outside of any expansion.
+ */
+fun PsiElement.findElementExpandedFrom(): PsiElement? {
```
Undin: I'd prefer to move all these helper functions and extension methods to `RsExpandedElement.kt` to group all related methods in one place

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -292,7 +292,7 @@ fun processExternCrateResolveVariants(
 
 private fun findDependencyCrateByName(context: RsElement, name: String): RsFile? {
     val refinedName = when {
-        name.startsWith(MACRO_CRATE_IDENTIFIER_PREFIX) && context.findMacroCallExpandedFrom() != null -> {
+          name.startsWith(MACRO_CRATE_IDENTIFIER_PREFIX) && context.isExpandedFromMacro -> {
```
Undin: indent

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
Undin: Maybe rewrite it somehow? For example, in the following way 
```kotlin
fun PsiElement.findNavigationTargetIfMacroExpansion(): PsiElement? {
    /** @see RsNamedElementImpl.getTextOffset */
    val element = (this as? RsNameIdentifierOwner)?.nameIdentifier ?: this
    return element.findElementExpandedFrom() ?: findMacroCallExpandedFrom()?.path // fallback to macro call
}
```
It seems it's more readable

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -758,14 +758,23 @@ private fun expandMacroOld(call: RsMacroCall): MacroExpansion? {
     if (call.containingCargoTarget?.pkg?.origin == PackageOrigin.STDLIB) {
         return null
     }
-    return expandMacroToMemoryFile(call)
+    return expandMacroToMemoryFile(
+        call,
+        // Old macros already consume too much memory, don't force them to consume more by range maps
+        storeRangeMap = isUnitTestMode // false
```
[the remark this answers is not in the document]
vlad20012: I'd prefer to leave it as is. Old macro expansion engine works much faster in tests and it also simpler to use in tests

src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
```diff
@@ -652,3 +663,34 @@ private fun calcStubIndex(psi: StubBasedPsiElement<*>): Int {
 
 private val VirtualFile.fileId: Int
     get() = (this as VirtualFileWithId).id
+
+// Use WeakReference it's relatively easy to load RangeMap again
```
Undin: Do you mean that `VirtualFile#loadRangeMap` is quite cheap?
If yes, let's rewrite it somehow to make it clearer

vlad20012: Yes. Done

vlad20012:
bors r=undin

PR #3653, opened 2019-04-05
PERF: optimize VFS batch (macro expansion)

vlad20012:
1. Optimize VFS batch
That was an original implementation of VFS batch which I removed because of API difference between 183 and 191 (and 192). But this implementation works faster. So if there will be any problems (including porting on 192), we can roll back to RefreshBasedVfsBatch

2. Split expansion writing into several batches
Now we produce fewer UI freezes

Part of #3628

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
@@ -21,6 +21,7 @@ import org.rust.lang.core.psi.ext.macroBody
 import org.rust.lang.core.psi.ext.resolveToMacro
 import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
 import org.rust.openapiext.*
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
```diff
@@ -121,18 +122,18 @@ abstract class MacroExpansionTaskBase(
             realTaskIndicator.text2 = "Writing expansion results"
 
             if (stages2.isNotEmpty()) {
-                val fs = LocalFsMacroExpansionVfsBatch(realFsExpansionContentRoot)
-                // TODO support cancellation here
-                val stages3 = stages2.map { stage2 ->
-                    val result = stage2.writeExpansionToFs(fs)
-                    doneStages.incrementAndGet()
-                    result
+                val stages3fs = stages2.chunked(100).map { stages2c ->
```
Undin: Let's extract this const and add javadoc for it

PR #3656, opened 2019-04-05
DBG: Load Rust stdlib sources

ortem:
Now it's possible to jump into Rust stdlib sources while debugging.

debug_stdlib

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
@@ -33,6 +33,20 @@ class RsLocalDebugProcess(
         }
     }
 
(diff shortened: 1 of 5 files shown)


debugger/src/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
```diff
@@ -33,6 +35,24 @@ class RsLocalDebugProcess(
         }
     }
 
+    fun loadRustcSources(sysroot: String, exePath: String) {
+        postCommand { driver ->
+            // TODO: find more convenient way to get the SHA-1
```
mchernyavsky: Can we parse commit hash in `scrapeRustcVersion` and add it to the `RustcVersion`?

ortem:
@mchernyavsky I've tested it on macOS, Linux, and Windows; works fine.

PR #3658, opened 2019-04-06
INSP: support integer const expr evaluation in `Match Check` inspection

### LINKED ISSUE #3577: False positive `Unreachable pattern` on const in range
```Rust
const MAX: usize = 10;

fn main() {
    let v = vec![1, 2, 3];
    match v.len() {
        1 => { println ! ("1"); }
        2...MAX => { println ! ("2...MAX"); } // "Unreachable pattern"
        _ => { println ! ("other"); }
    }
}
```

Undin:
Fixes #3577


src/main/kotlin/org/rust/ide/inspections/checkMatch/CheckMatchUtils.kt
```diff
@@ -29,23 +29,28 @@ val Matrix.type: Ty
...
-                is RsLiteralKind.Integer -> Constant.Integer(kind.value ?: 0)
-                is RsLiteralKind.Float -> Constant.Float(kind.value ?: 0.0)
-                is RsLiteralKind.String -> Constant.String(kind.value ?: "")
-                is RsLiteralKind.Char -> Constant.Char(kind.value ?: "")
-                null -> Constant.Unknown
+private val RsExpr.value: Constant get() {
+    return if (this is RsLitExpr) {
+        val kind = kind
+        when (kind) {
+            is RsLiteralKind.Boolean -> Constant.Boolean(kind.value)
```
ortem: Maybe move this logic into "RsConstExprEvaluator"?

src/main/kotlin/org/rust/lang/utils/evaluation/ExprValue.kt
```diff
@@ -0,0 +1,41 @@
...
+    object Unknown : ExprValue()
+
+    operator fun compareTo(other: ExprValue): Int {
+        return when {
+            this is Bool && other is Bool -> value.compareTo(other.value)
+            this is Integer && other is Integer -> value.compareTo(other.value)
+            this is Float && other is Float -> value.compareTo(other.value)
+            this is Str && other is Str -> value.compareTo(other.value)
+            this is Char && other is Char -> value.compareTo(other.value)
+            else -> -1
```
ortem: `-1` violates the comparison contract (`sgn(x.compareTo(y)) == -sgn(y.compareTo(x))`). This leads to errors.

For example, now inside `org.rust.ide.inspections.checkMatch.Constructor.ConstantValue#coveredByRange` the range `[x, Unknown]` covers any value `y` such that `y > x`. It can be exploited to get false positive `Unreachable pattern`:

```Kotlin
    fun `test unresolved constant`() = checkByText("""
        fn main() {
            match 42 {
                0..UNRESOLVED => {}
                20..50 => {} // false positive `Unreachable pattern`
                _ => {}
            }
        }
    """)
```
This bug is inherited from the original `Constant.kt`, and it'd be great if it will be fixed in this PR. You can throw `CheckMatchException("Unresolved constant")` inside `RsPat.kind` method when the evaluated expression is `Unknown` (or just `null` instead of `Unknown`?)

Undin: In case, if we can't evaluate some pattern, I throw `CheckMatchException` so it fixes your case.
But we should be more precise here in future and just exclude unevaluated patterns from matrix

ortem: We should not just exclude them because it leads to false positive non-exhaustiveness

Undin: At least, we can check usefulness on truncated matrix

ortem: Also, maybe throw an exception instead of `-1` in case of invalid types?

Undin: It's unexpected that `<` throws an exception.
Also, it's unclear what exception we should throw here because `ExprValue` is not only `Match Check` inspection api.
In general, I think it's wrong that we provide the ability to compare values of different types.
So I'd prefer not to change it in this PR and allow comparing only values of the same classes in future PR

Issue #3585, opened 2019-03-24
Show local variables or functions in top of the completion list


SomeoneToIgnore:
When there's a local variable or method that starts with the same char sequence as some other entities (keywords, structs from different crates), it would be great if those local variables are proposed first in the completion list, as the more specific for the context.

A few examples:

<img width="530" alt="image" src="https://user-images.githubusercontent.com/2690773/54885359-fc2cfd80-4e83-11e9-81de-58be27726e0d.png">
<img width="563" alt="image" src="https://user-images.githubusercontent.com/2690773/54885360-fdf6c100-4e83-11e9-9444-a92b487fb3e0.png">
<img width="562" alt="image" src="https://user-images.githubusercontent.com/2690773/54885361-ffc08480-4e83-11e9-9eec-34aa81bdb257.png">

PR #3663, opened 2019-04-08
COMP: Show local definitions in top of the completion list

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3585.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
@@ -72,8 +72,8 @@ val RsElement.containingCargoTarget: CargoWorkspace.Target?
 
 val RsElement.containingCargoPackage: CargoWorkspace.Package? get() = containingCargoTarget?.pkg
 
-val PsiElement.isEdition2018: Boolean get() =
-    contextOrSelf<RsElement>()?.containingCargoTarget?.edition == CargoWorkspace.Edition.EDITION_2018
+val PsiElement.isEdition2018: Boolean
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
```diff
@@ -145,3 +146,10 @@ fun RsElement.findInScope(name: String): PsiElement? {
     }
     return resolved
 }
+
+
+val RsElement.isLocal: Boolean
+    get() {
+        val context = PsiTreeUtil.getContextOfType(this, true, RsItemElement::class.java, RsFile::class.java)
+        return context != null && context !is RsMod
```
vlad20012: Is it expected that context can be RsMembers and so methods will be classified as local?

vlad20012: And enum item too

mchernyavsky: > Is it expected that context can be trait or impl and so methods themselves will be classified as locals?

I think yes. Examples from Kotlin:
image

image

src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
```diff
@@ -145,3 +143,10 @@ fun RsElement.findInScope(name: String): PsiElement? {
     }
     return resolved
 }
+
+
+val RsElement.isLocal: Boolean
```
vlad20012: Looks like a something very specific to completion sorting. Maybe move it to LookupElements.kt and make private?

Issue #3672, opened 2019-04-09
quickcheck macro is not detected as test function


fredpointzero:
## Environment

* **Intellij-Rust plugin version:** v0.2.95.2120-191
* **Rust toolchain version:** nightly-x86_64-pc-windows-gnu
* **IDE name and version:**
CLion 2019.1
Build #CL-191.6183.77, built on March 25, 2019
JRE: 1.8.0_202-release-1483-b39 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o

* **Operating system:** Windows 10 10.0

## Problem description

Hi,

When using quickcheck, the test functions are not recognized.

## Steps to reproduce

STR:
1. Create new library project
2. Add quickcheck in `Cargo.toml`:
```
[dev-dependencies]
quickcheck = "^0.8.0"
quickcheck_macros = "^0.8.0"
```
3. Add extern crate statements in `lib.rs`:
```
#[cfg(test)]
#[macro_use]
extern crate quickcheck;
#[cfg(test)]
#[macro_use(quickcheck)]
extern crate quickcheck_macros;
```
4. Add sample test in `lib.rs`:
```
#[cfg(test)]
mod tests {
    fn reverse<T: Clone>(xs: &[T]) -> Vec<T> {
        let mut rev = vec!();
        for x in xs {
            rev.insert(0, x.clone())
        }
        rev
    }

    #[quickcheck]
    fn double_reversal_is_identity(xs: Vec<isize>) -> bool {
        xs == reverse(&reverse(&xs))
    }
}
```
5. Expected: A clickable green arrow on `mod tests` and `fn double_reversal_is_identity(xs: Vec<isize>) -> bool`, to run respectively, tests of the module, and the quickcheck test.
(Note: cargo test will work)

PR #3673, opened 2019-04-09
RUN: Line marker for test with `quickcheck` attribute

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3672.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
@@ -25,7 +25,7 @@ val RsFunction.isAssocFn: Boolean get() = selfParameter == null && owner.isImplO
 
 val RsFunction.isTest: Boolean get() {
     val stub = stub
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -25,7 +25,7 @@ val RsFunction.isAssocFn: Boolean get() = selfParameter == null && owner.isImplO
 
 val RsFunction.isTest: Boolean get() {
     val stub = stub
-    return stub?.isTest ?: queryAttributes.hasAtomAttribute("test")
+    return stub?.isTest ?: queryAttributes.hasAtomAttribute("test") || queryAttributes.hasAtomAttribute("quickcheck")
```
vlad20012:
```suggestion
    return stub?.isTest ?: (queryAttributes.hasAtomAttribute("test") || queryAttributes.hasAtomAttribute("quickcheck"))
```

mchernyavsky: Fixed.

PR #3678, opened 2019-04-10
TY: fix TyProjection folding

vlad20012:
bors r+

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/BoundElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/BoundElement.kt
@@ -52,7 +52,7 @@ data class BoundElement<out E : RsElement>(
         )
 
     override fun superVisitWith(visitor: TypeVisitor): Boolean = assoc.values.any { visitor.visitTy(it) } ||
-        subst.types.any { visitor.visitTy(it) } && subst.regions.any { visitor.visitRegion(it) }
+        subst.types.any { it.visitWith(visitor) } || subst.regions.any { it.visitWith(visitor) }
 
 }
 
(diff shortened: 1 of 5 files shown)


vlad20012:
fixed tests

src/main/kotlin/org/rust/lang/core/types/BoundElement.kt
```diff
@@ -52,7 +52,7 @@ data class BoundElement<out E : RsElement>(
         )
 
     override fun superVisitWith(visitor: TypeVisitor): Boolean = assoc.values.any { visitor.visitTy(it) } ||
```
mchernyavsky: Shouldn't we use `visitWith` instead of `visitTy` and `visitRegion` here?

vlad20012: Fixed

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -52,13 +56,21 @@ sealed class RsDiagnostic(
         element: PsiElement,
         private val expectedTy: Ty,
         private val actualTy: Ty
-    ) : RsDiagnostic(element, inspectionClass = RsTypeCheckInspection::class.java) {
+    ) : RsDiagnostic(element, inspectionClass = RsTypeCheckInspection::class.java), TypeFoldable<TypeError> {
+        private val description: String? = if (expectedTy.hasTyInfer || actualTy.hasTyInfer) {
+            // if types contain infer types, they will be replaced with more specific (e.g. `{integer}` with `i32`)
+            // so we capture string representation of types right here
+            expectedFound(expectedTy, actualTy)
```
mchernyavsky: Can you write a test for this case?

vlad20012: Hmm, this should be covered by multiple tests, but we don't have any tests that check such error description (of any annotation), so if we want to start testing it, maybe we can do it in a separate PR?

PR #3688, opened 2019-04-11
Fix macro expansion progress bar hung

vlad20012:
... if switching between projects during macro expansion

bors r+


vlad20012:
bors retry

PR #3696, opened 2019-04-13
RES&TY: fix Self in struct/enum

### LINKED ISSUE #3457: `Self` in struct
## Environment

* **Intellij-Rust plugin version:** image
* **Rust toolchain version:** Not related
* **IDE name and version:** image
* **Operating system:** Windows 10 but I believe this is not related

## Problem description
Obtain code like this:
```rust
enum Example { Member(Box<Self>) }
fn f(e: Example) { match e { Example::Member(mem) => {}, } }
```

## Steps to reproduce
Look at the type hint:
image

It's not properly inferred and completion is not available.

vlad20012:
Fixes #3457


ice1000:
Cool！！！

PR #3711, opened 2019-04-15
INSP: improve inspection suppression

### LINKED ISSUE #2308: Suppress for item should insert #![allow]
Currently <kbd>ALT</kbd>-<kbd>ENTER</kbd> -> Suppress Inspection for Item does the following:

```rust
fn BadStyle() -> {}

//noinspection RsFunctionNaming
fn BadStyle() -> {}
```

And the inspection is still highlighted. Manually typing the Rust-correct allow:

```rust
#[allow(non_snake_case)]
fn BadStyle() -> {}
```

correctly suppresses the inspection.

Undin:
These changes provide `RsLintInspection` - base class for all inspections that should be suppressed by `allow` attribute. This class uses the correct API to suppress warning annotations. Also, it provides suppression quick fixes to add `allow(lint)` attribute to the chosen item. 

Now naming convention and deprecated inspections extend `RsLintInspection` and provide suppression quick fixes.

Note, our parser doesn't correctly parse expressions with outer attributes like
```rust
#[allow(lint)]
foo();
```
so, statement suppression quick fix is not provided in some cases. Should be fixed from grammar side

Fixes #2308

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -938,6 +938,7 @@ StmtModeExpr ::= <<structLiterals 'ON' <<stmtMode 'ON' Expr>> >> { elementType =
 StmtModeExprOff ::= <<stmtMode 'OFF' Expr>> { elementType = Expr }
 
 BlockExpr ::= OuterAttr* [ (unsafe | asyncBlock move? | try) &'{' ] SimpleBlock {
+  implements = [ "org.rust.lang.core.psi.ext.RsOuterAttributeOwner" ]
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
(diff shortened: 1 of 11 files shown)


src/main/grammars/RustParser.bnf
```diff
@@ -1356,6 +1374,7 @@ fake ExprStmt ::= AnyExpr ';'? { extends = Stmt }
 
 LetDecl ::= OuterAttr* let Pat TypeAscription? [ '=' AnyExpr ] ';' {
   extends = Stmt
+  implements = [ "org.rust.lang.core.psi.ext.RsOuterAttributeOwner" ]
```
mchernyavsky: Maybe `StructLiteralField` should also implement this interface?

src/main/kotlin/org/rust/ide/inspections/RsLintInspection.kt
```diff
@@ -0,0 +1,106 @@
...
+            if (action != null) {
+                fixes += action
+            }
+        }
+        return fixes.toTypedArray()
+    }
+}
+
+private class RsSuppressQuickFix(
+    private val suppressAt: RsDocAndAttributeOwner,
```
vlad20012: Possible memory leak here (quick fixes shouldn't store references to PSI)

Undin: Oh, thanks!
Will fix it asap
Good thing: now I know that DevKit plugin has inspection to detect such mistakes and it's great!

Issue #3586, opened 2019-03-24
Add imports when a trait method gets autoimplemented


SomeoneToIgnore:
Consider a `Debug` trait implementation in the test https://github.com/intellij-rust/intellij-rust/blob/678344c8861bfcfcdd5831fb02295548184410de/src/test/kotlin/org/rust/ide/refactoring/implementMembers/ImplementMembersHandlerTest.kt#L665

After the trait gets implemented, `Formatter` and `Error` entities are not imported after the automatic generation.
Would be good to have every trait used in the generation imported.

PR #3712, opened 2019-04-15
INT: Auto-import unresolved types in generated code

mchernyavsky:
Auto-import unresolved types in intentions:
  - Implement members
  - Specify type explicitly
  - Fill match arms
  - Add remaining patterns
  - Use destructuring declaration

This PR doesn't affect:
  - Add missing fields / Recursively add missing fields
  - Initialize with a default value

because `RsDefaultValueBuilder.buildFor` can't use unresolved types.

Based on https://github.com/intellij-rust/intellij-rust/pull/3099.
Closes https://github.com/intellij-rust/intellij-rust/issues/3586. Closes https://github.com/intellij-rust/intellij-rust/issues/3495.
Relates to https://github.com/intellij-rust/intellij-rust/pull/3520.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -56,7 +56,7 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
         val element = startElement as? RsElement ?: return
         val (_, candidates) = when (element) {
             is RsPath -> findApplicableContext(project, element) ?: return
-            is RsMethodCall -> findApplicableContext(project, element)  ?: return
+            is RsMethodCall -> findApplicableContext(project, element) ?: return
             else -> return
         }
 
@@ -389,10 +389,10 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
                 candidate to attributes
             }
 
-            val condition: (RsFile.Attributes) -> Boolean = if (hasImportWithSameAttributes) {
-                attributes -> attributes == fileAttributes
-            } else {
-                attributes -> attributes < fileAttributes
+            val condition: (RsFile.Attributes) -> Boolean = if (hasImportWithSameAttributes) { attributes ->
+                attributes == fileAttributes
+            } else { attributes ->
+                attributes < fileAttributes
             }
             return candidateToAttributes.mapNotNull { (candidate, attributes) ->
                 if (condition(attributes)) candidate else null
@@ -442,7 +442,7 @@ sealed class ImportInfo {
 
     abstract val usePath: String
 
-    class LocalImportInfo(override val usePath: String): ImportInfo()
+    class LocalImportInfo(override val usePath: String) : ImportInfo()
 
     class ExternCrateImportInfo(
         val target: CargoWorkspace.Target,
@@ -674,17 +674,18 @@ data class ImportContext private constructor(
     }
 }
 
-private val RsPath.pathParsingMode: PathParsingMode get() = when (parent) {
-    is RsPathExpr,
-    is RsStructLiteral,
-    is RsPatStruct,
-    is RsPatTupleStruct -> PathParsingMode.COLONS
-    else -> PathParsingMode.NO_COLONS
-}
+private val RsPath.pathParsingMode: PathParsingMode
+    get() = when (parent) {
+        is RsPathExpr,
+        is RsStructLiteral,
+        is RsPatStruct,
+        is RsPatTupleStruct -> PathParsingMode.COLONS
+        else -> PathParsingMode.NO_COLONS
+    }
 private val RsElement.stdlibAttributes: RsFile.Attributes
     get() = (crateRoot?.containingFile as? RsFile)?.attributes ?: RsFile.Attributes.NONE
 private val RsItemsOwner.firstItem: RsElement get() = itemsAndMacros.first { it !is RsAttr && it !is RsVis }
-private val <T: RsElement> List<T>.lastElement: T? get() = maxBy { it.textOffset }
+private val <T : RsElement> List<T>.lastElement: T? get() = maxBy { it.textOffset }
 
 private val CargoWorkspace.Target.isStd: Boolean
     get() = pkg.origin == PackageOrigin.STDLIB && normName == AutoInjectedCrates.STD
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -671,6 +671,16 @@ data class ImportContext private constructor(
             attributes = path.stdlibAttributes,
             namespaceFilter = path.namespaceFilter(isCompletion)
         )
+
+        fun from(project: Project, element: RsElement): ImportContext = ImportContext(
+            project = project,
+            mod = element.containingMod,
+            superMods = LinkedHashSet(element.containingMod.superMods),
+            scope = RsWithMacrosScope(project, GlobalSearchScope.allScope(project)),
+            pathParsingMode = PathParsingMode.NO_COLONS,
+            attributes = element.stdlibAttributes,
+            namespaceFilter = { true }
+        )
     }
 }
 
--- a/src/main/kotlin/org/rust/ide/intentions/DestructureIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/DestructureIntention.kt
@@ -8,6 +8,7 @@ package org.rust.ide.intentions
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import org.rust.ide.inspections.import.RsImportHelper.importTypeReferencesFromTy
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.ty.TyAdt
@@ -53,7 +54,11 @@ class DestructureIntention : RsElementBaseIntentionAction<DestructureIntention.C
         } else {
             factory.createPatStruct(struct)
         }
-        val newPatStruct = patIdent.replace(patStruct) as? RsPatTupleStruct ?: return
+        val newPatStruct = patIdent.replace(patStruct) as? RsPat ?: return
+
+        importTypeReferencesFromTy(newPatStruct, struct.declaredType)
+
+        if (newPatStruct !is RsPatTupleStruct) return
         if (struct.positionalFields.isNotEmpty()) {
             replaceFieldPatsWithPlaceholders(editor, newPatStruct)
         }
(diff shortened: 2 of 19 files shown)


Undin:
@mchernyavsky Could you rebase it onto master, please?

mchernyavsky:
@Undin Done.

Undin:
bors try

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
-    is RsPatStruct,
-    is RsPatTupleStruct -> PathParsingMode.COLONS
-    else -> PathParsingMode.NO_COLONS
+/**
+ * @param toImport  Set of unresolved items that can be imported
+ * @param toQualify Set of unresolved items that can't be imported
+ */
+data class TypeReferencesInfo(
+    val toImport: Set<RsQualifiedNamedElement>,
+    val toQualify: Set<RsQualifiedNamedElement>
```
Undin: Why do we need this list?
Looks like we don't use it anywhere

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
+                is TyAdt ->
+                    result.add(ty.item)
+                is TyAnon ->
+                    result.addAll(ty.traits.map { it.element })
+                is TyTraitObject ->
+                    result.add(ty.trait.element)
+                is TyProjection -> {
+                    result.add(ty.trait.element)
+                    result.add(ty.target)
+                }
```
Undin: I think it can be reformatted into 
```kotlin
is TyAdt -> result.add(ty.item)
is TyAnon -> result.addAll(ty.traits.map { it.element })
is TyTraitObject -> result.add(ty.trait.element)
is TyProjection -> {
    result.add(ty.trait.element)
    result.add(ty.target)
}
```

Undin: And maybe it's worth to use `+=` instead of `add` as more kotlin way.
But it's a matter of taste so I don't insist

src/main/kotlin/org/rust/ide/intentions/DestructureIntention.kt
```diff
@@ -47,6 +49,9 @@ class DestructureIntention : RsElementBaseIntentionAction<DestructureIntention.C
     }
 
     private fun handleStruct(project: Project, editor: Editor, patIdent: RsPatIdent, struct: RsStructItem) {
+        val (toImport, _) = patIdent.getTypeReferencesInfoFromTy(struct.declaredType)
+        patIdent.importElements(toImport)
```
Undin: Hm, looks like we always do the same actions: call `getTypeReferencesInfoFrom*` and call `importElements`
Maybe we can create a single public function to make both actions inside it?

Undin: Also, I don't like that current functions are extention functions of `RsElement` because we use element only as context.
So I'd prefer pass element as explicit argument.
If you don't want to create new top-level function(s), let's create some object class like `RsImportHelper` and move them into it

Undin: And maybe it's worth to call `import` method after main action?
Imagine the following case: our new awesome import feature throws an exception. And in this case, main action won't be executed and user doesn't see any change. Otherwise, we make all necessary actions and just miss imports that user can import manually in case of potential exception

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
 }
 
-private val RsPath.pathParsingMode: PathParsingMode get() = when (parent) {
-    is RsPathExpr,
-    is RsStructLiteral,
-    is RsPatStruct,
-    is RsPatTupleStruct -> PathParsingMode.COLONS
-    else -> PathParsingMode.NO_COLONS
+/**
+ * @param toImport  Set of unresolved items that can be imported
```
Undin:
```suggestion
 * @param toImport  Set of unresolved items that should be imported
```

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
+ * - Items that can't be imported
+ */
+private fun processRawImportSubjects(
+    context: RsElement,
+    rawImportSubjects: Set<RsQualifiedNamedElement>
+): TypeReferencesInfo {
+    val importSubjects = hashMapOf<String, MutableSet<RsQualifiedNamedElement>>()
+    for (element in rawImportSubjects) {
+        val name = element.name ?: continue
+        val group = importSubjects.getOrPut(name) { hashSetOf() }
```
Undin:
```suggestion
        importSubjects.getOrPut(name, ::hashSetOf) += element
```

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
+ * Takes `rawImportSubjects` and filters items that are unresolved in the current `context`.
+ * Then splits this items into two sets:
+ * - Items that can be imported
+ * - Items that can't be imported
+ */
+private fun processRawImportSubjects(
+    context: RsElement,
+    rawImportSubjects: Set<RsQualifiedNamedElement>
+): TypeReferencesInfo {
+    val importSubjects = hashMapOf<String, MutableSet<RsQualifiedNamedElement>>()
```
Undin: Why do we need this map at all?
Can't we just process all elements from `rawImportSubjects` directly?

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
+    for (element in rawImportSubjects) {
+        val name = element.name ?: continue
+        val group = importSubjects.getOrPut(name) { hashSetOf() }
+        group.add(element)
+    }
+
+    val toQualifiedName = hashSetOf<RsQualifiedNamedElement>()
+    processNestedScopesUpwards(context, TYPES) { entry ->
```
Undin: If `importSubjects` is not necessary, `org.rust.lang.core.psi.ext.RsItemElementKt#filterInScope` maybe useful here

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,146 @@ data class ImportContext private constructor(
...
+private fun collectImportSubjectsFromTy(
+    ty: Ty,
+    subst: Substitution,
+    result: MutableSet<RsQualifiedNamedElement>
+) {
+    ty.substitute(subst).visitWith(object : TypeVisitor {
+        override fun visitTy(ty: Ty): Boolean {
+            when (ty) {
+                is TyAdt ->
+                    result.add(ty.item)
```
Undin: This code doesn't take into account type aliases. It should do it at least for `SpecifyTypeExplicitlyIntention` because I've just merged #4403
For example, 
```rust
use crate::foo::bar_baz;

mod foo {
    pub struct Bar<T>(T);
    pub struct Baz;
    pub type BarBaz = Bar<Baz>;

    pub fn bar_baz() -> BarBaz { Bar(Baz) }
}

fn main() {
    let bar/*caret*/ = bar_baz();
}
```

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -142,6 +142,12 @@ class RsPsiFactory(
         return structLiteralField
     }
 
+    fun createStructLiteralFieldFull(name: String, value: String): RsStructLiteralField {
```
Undin: It seems it's unused

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,170 @@ data class ImportContext private constructor(
...
     }
 }
 
-private val RsPath.pathParsingMode: PathParsingMode get() = when (parent) {
-    is RsPathExpr,
-    is RsStructLiteral,
-    is RsPatStruct,
-    is RsPatTupleStruct -> PathParsingMode.COLONS
-    else -> PathParsingMode.NO_COLONS
+object RsImportHelper {
```
Undin: Maybe move this class to separate file?

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -576,6 +526,60 @@ fun ImportCandidate.import(context: RsElement) {
     insertionScope.insertUseItem(psiFactory, "$prefix${info.usePath}")
 }
 
+// Semantic signature of method is `ImportItem.canBeImported(mod: RsMod)`
+// but in our case `mod` is always same and `mod` needs only to get set of its super mods
+// so we pass `superMods` instead of `mod` for optimization
+fun QualifiedNamedItem.canBeImported(superMods: LinkedHashSet<RsMod>): ImportInfo? {
```
Undin: Why did you move it and make public?

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,170 @@ data class ImportContext private constructor(
...
+        }
+
+        return TypeReferencesInfo(importSubjects.flatMap { it.value }.toSet(), toQualifiedName)
+    }
+
+    /**
+     * @param toImport  Set of unresolved items that should be imported
+     * @param toQualify Set of unresolved items that can't be imported
+     */
+    data class TypeReferencesInfo(
```
Undin: Can be private

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -671,20 +675,170 @@ data class ImportContext private constructor(
...
+     */
+    private fun processRawImportSubjects(
+        context: RsElement,
+        rawImportSubjects: Set<RsQualifiedNamedElement>
+    ): TypeReferencesInfo {
+        val importSubjects = hashMapOf<String, MutableSet<RsQualifiedNamedElement>>()
+        for (element in rawImportSubjects) {
+            val name = element.name ?: continue
+            importSubjects.getOrPut(name, ::hashSetOf) += element
+        }
```
Undin: Could you come up with case when value set has size > 1?

mchernyavsky:
```
use crate::a::foo;

mod a {
    mod b { pub struct S; }
    mod c { pub struct S; }
    pub fn foo() -> (b::S, c::S) { unimplemented!() }
}

fn main() {
    let x/*caret*/ = foo();
}
```

mchernyavsky:
@Undin Fixed.

PR #3714, opened 2019-04-15
STUB&RES: fix scoped resolve inside stubbed fn body

vlad20012:
```rust
//- main.rs
    mod foo;
    fn main() {
        foo::S.baz();
    }        //^ foo.rs
//- foo.rs
    pub struct S;

    fn foobar() {
        if true {
            impl S {
                pub fn baz(&self) {}
            }
        } else {
            struct S;
        }
    }
```

Works when `foo.rs` is stubbed

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -10,6 +10,7 @@ import com.intellij.psi.PsiFile
 import com.intellij.psi.StubBuilder
 import com.intellij.psi.stubs.*
 import com.intellij.psi.tree.IStubFileElementType
+import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.util.BitUtil
 import com.intellij.util.io.DataInputOutputUtil.readNullable
 import com.intellij.util.io.DataInputOutputUtil.writeNullable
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rust/lang/core/resolve/RsStubOnlyResolveTest.kt
```diff
@@ -561,4 +561,24 @@ class RsStubOnlyResolveTest : RsResolveTestBase() {
     //- main/foo.rs
         pub struct Bar;
     """, NameResolutionTestmarks.crateRootModule)
+
+    fun `test scoped resolve inside stubbed function body`() = stubOnlyResolve("""
```
Undin: What does this test case check?
Looks like it already works with the current nightly plugin

vlad20012: I will not work if `foo.rs` is stubbed

Undin: Indeed!

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1108,6 +1109,26 @@ class RsBinaryOpStub(
     }
 }
 
+class RsBlockStub(
+    parent: StubElement<*>?, elementType: IStubElementType<*, *>
+) : StubBase<RsBinaryOp>(parent, elementType) {
```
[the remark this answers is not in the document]
vlad20012: Sounds reasonable. Done

Issue #3000, opened 2018-10-29
Debugger pretty printers


Undin:
This issue is just an overview of current known issues related to debugger pretty printers for `gdb` and `lldb`

* [ ] scalar types
  * [ ] show correct names of scalar types: `u32`, `i16`, etc instead of C/C++ names like `unsigned int`, `short`, etc
  * [ ] lldb considers that `i8` and `u8` are `char` and `unsigned char` respectevely and converts values of these types to symbols
  * [ ] `char` type now is shows as `void` without any value
* [ ] arrays and slices
  * [ ] correct type names
  * [ ] only size in summary
* [ ] tuples
  * [ ] don't show fields values in summary
  * [ ] show `i` instead of  `__i` as field name
* [ ] structs
  * [ ] don't show fields values in summary
* [ ] enums/enum variants
  * [ ] don't show fields values in summary
  * [ ] correctly determine enum variant
  * [ ] provide only related to current enum variant fields
* [ ] references
  * [ ] show correct type names like `&i32` instead of `int *`
  * [ ] show summary of dereferenced value of reference
* [ ] pointers
  * [ ] show correct type names like `*i32` instead of `int *`
* [ ] `Vec`
  * [ ] show only size of vector instead of `vec![item1, item2, ...]` string in summary
  * [ ] show indexed children instead of `buf` and `len` fields
* [ ] `Rc` and `Arc`
  * [ ] show `strong`, `week` and summary of dereferenced value
* [ ] Box
  * [ ] show correct type (`Box<i32>` instead of `int *`)
  * [ ] show summary of dereferenced value

DezzK:
I think, debugger issues are the most annoying at this moment.

Undin:
@DezzK what do you mean?

DezzK:
@Undin I mean, the plugin works pretty good for me (excepting debugging, macro support and several not so serious issues). But still the most annoying of issues is debugging support. Despite it works much better than several months ago, it is still needed to add dbg!(...) macro or other logging just to print variable value in part of cases because not watching, nor expression evaluation can help at these cases. Perhaps it is specific to our project, because we activelly use enums and enum variants with data in them, which are not supported yet by the plugin (I mean that printers for enum variants are not implemented yet).

BTW, using enum variants is the Rust way, it's not a kind of rarely using features. I will be appreciated, if you implement a support of enum variants in the debugger soon. I think, showing variable value in debugger is much more meaningful than showing correct type names and so on. :)

[...]

PR #3045, opened 2018-11-16
Add bundled GDB pretty-printers

ortem:
Add bundled GDB pretty-printers.

Now it supports:
* struct, tuple
* `str`, `String`, `OsString` (with UTF-8 support)
* `Vec`, `VecDeque`
* `BTreeSet`, `BTreeMap` (thanks, @tromey!)
* `Rc`, `Arc`
* `Cell`, `Ref`, `RefMut`, `RefCell`

Fixes #3357 

There are some problems with enums because of wrong DWARF codes in rustc's codegen (it affects default rustc pretty-printers too); I'm going to fix it after the next rustc release.

screenshot 2018-11-19 at 11 54 56
screenshot 2018-11-19 at 11 54 16

screenshot 2018-11-27 at 19 04 44
screenshot 2018-11-27 at 19 00 47

screenshot 2018-11-29 at 18 56 55

PR #3715, opened 2019-04-15
DBG: add support for printing Enums directly in debugger

Kobzol:
Hi, using the latest versions of Rust and the bundled GDB (Rust 1.34 and GDB 8.2), enums are visualized in a weird way in the synthetic debugger view (it basically prints the discriminant and then all of the possible variants as children). I changed the output so that only the correct enum variant is printed.

The name of the variant is outputted as a child field, since to change the type name of the enum variable itself I would have to change its type and I don't know how to do that (if it's even possible in the GDB API, doesn't seem like so).

Code:
enum-gdb

Original output:
enum-gdb-vis-orig

New output:
enum-gdb-vis

So far I only added support for GDB and there are no tests, I'd like to know what you think about this change and if I implemented it correctly before I go forward. From the code it looked like it already tried supporting enums, but it seems that the code involving `is_variant` in structs and tuples was not working in the current Rust/GDB combination, the variant versions of struct and enum were not created at all.

TODO:
- [ ] LLDB support
- [x] Tests

This falls under pretty printer improvements umbrella issue (https://github.com/intellij-rust/intellij-rust/issues/3000).

### DIFF
--- a/prettyPrinters/gdb_lookup.py
+++ b/prettyPrinters/gdb_lookup.py
@@ -23,12 +23,10 @@ def lookup(valobj):
 
     if rust_type == RustType.STRUCT:
         return StructProvider(valobj)
-    if rust_type == RustType.STRUCT_VARIANT:
-        return StructProvider(valobj, is_variant=True)
     if rust_type == RustType.TUPLE:
         return TupleProvider(valobj)
-    if rust_type == RustType.TUPLE_VARIANT:
-        return TupleProvider(valobj, is_variant=True)
+    if rust_type == RustType.ENUM:
+        return EnumProvider(valobj)
 
     if rust_type == RustType.STD_STRING:
         return StdStringProvider(valobj)
--- a/prettyPrinters/gdb_providers.py
+++ b/prettyPrinters/gdb_providers.py
@@ -15,23 +15,22 @@ def unwrap_unique_or_non_null(unique_or_nonnull):
 
 
 class StructProvider:
-    def __init__(self, valobj, is_variant=False):
+    def __init__(self, valobj):
         self.valobj = valobj
-        self.is_variant = is_variant
-        fields = self.valobj.type.fields()
-        self.fields = fields[1:] if is_variant else fields
+        self.fields = self.valobj.type.fields()
+
+    def to_string(self):
+        return self.valobj.type.name
 
     def children(self):
-        for i, field in enumerate(self.fields):
-            yield ((field.name, self.valobj[field.name]))
+        for field in self.fields:
+            yield (field.name, self.valobj[field.name])
 
 
 class TupleProvider:
-    def __init__(self, valobj, is_variant=False):
+    def __init__(self, valobj):
         self.valobj = valobj
-        self.is_variant = is_variant
-        fields = self.valobj.type.fields()
-        self.fields = fields[1:] if is_variant else fields
+        self.fields = self.valobj.type.fields()
 
     def to_string(self):
         return "size={}".format(len(self.fields))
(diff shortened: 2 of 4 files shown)


ortem:
Hi @Kobzol, thanks for this PR! Here are some of my thoughts on this topic.

## gdb
I was thinking about such formatting of enums when I was working on [this PR](https://github.com/intellij-rust/intellij-rust/pull/3045). @tromey explained that current formatting relates to [this issue in gdb](https://sourceware.org/bugzilla/show_bug.cgi?id=20163). As I understand, enum formatting will become fine when this gdb issue will be fixed. But for now, I think your solution is OK and we really can extract only one child from all enum's children.
> I would have to change its type and I don't know how to do that

I was trying to do it and I couldn't find the way too :(

> it seems that the code involving is_variant in structs and tuples was not working in the current Rust/GDB combination

You're right, this code was written for old Rust versions and seems like it not applicable now.

## lldb
AFAIK, it's impossible to print enums in lldb due to [this issue in rustc](https://github.com/rust-lang/rust/issues/58492). It would be great if you find the solution, but I'm not sure that it can be done without some changes in rustc itself.

Kobzol:
I missed that PR, that explains a lot :-) I tried to run an interactive LLDB shell and examined the enum variable. All `fields`, `members` and `enum_members` properties of the Enum type return an empty array, so you're probably right, in the current state this has to be solved either on the LLDB or Rustc level.

ortem:
@Kobzol any progress? I'm going to merge it when you add tests for gdb.

Kobzol:
I'm having a bit of trouble with the tests. I added a new file (`enum.rs`) to `pretty_printers_tests/tests`, however when I run the tests (`cargo run gdb`), it doesn't seem to use my Enum provider (it prints the Enum in a wrong way). When I launch CLion with the plugin containing my changes, run a GDB debug session and use the same command (`print <variable`), it prints the enum correctly. Is there anything else that has to be done to register the changes for the tests?

ortem:
@Kobzol Yes, they use default `gdb` if you didn't specify a path to the bundled gdb. See `runPrettyPrintersTests` task in `build.gradle.kts`. Also, you can just push your tests and check the result of CI.

Kobzol:
@ortem Thanks, `runPrettyPrintersTests` was exactly what I needed. The test I added uses nested enums and includes two special cases (Enum with only one variant and enum with no variants). If you think more tests are necessary please let me know.

ortem:
Thank you!
bors r+

ice1000:
I am getting some errors with gdb when debugging with the current nightly plugin.
The gdb console prints:

```

Traceback (most recent call last):
  File "C:/Users/ice10/AppData/Local/JetBrains/Toolbox/apps/CLion/ch-0/192.4787.12/bin/gdb/renderers\default\printers.py", line 127, in to_string
    return '%s' % self.val.cast(gdb.lookup_type('void *'))
  File "C:/Users/ice10/AppData/Local/JetBrains/Toolbox/apps/CLion/ch-0/192.4787.12/bin/gdb/renderers\default\__init__.py", line 33, in lookup_type
    raise e
  File "C:/Users/ice10/AppData/Local/JetBrains/Toolbox/apps/CLion/ch-0/192.4787.12/bin/gdb/renderers\default\__init__.py", line 28, in lookup_type
    return lookup_type_orig(type_str)
gdb.error: No type named void *.
Traceback (most recent call last):
  File "C:\Users\ice10\.CLion2019.2\config\plugins\intellij-rust\prettyPrinters\gdb_lookup.py", line 29, in lookup
    return EnumProvider(valobj)
  File "C:\Users\ice10\.CLion2019.2\config\plugins\intellij-rust\prettyPrinters\gdb_providers.py", line 59, in __init__
    self.active_variant = content[fields[discriminant]]
IndexError: list index out of range
```

This prevents the pretty printer from showing anything for my variable.

Kobzol:
Could you please post a short example of code that you're debugging? Or some variant of it that causes the crash if the code is private. By nightly you mean nightly version IntelliJ plugin or nightly Rust?

ice1000:
Nightly plugin. The code is pretty large and I'll try to mock some smaller reproducible examples tomorrow.

Kobzol:
Thank you. Please create a new issue for it and maybe link it here in a comment.

PR #3723, opened 2019-04-16
RES: Use `context` instead of `parent`

ortem:


mchernyavsky:
bors ping

vlad20012:
I think these changes aren't necessary. Let's wait for specific reproducible bugs first

Issue #3701, opened 2019-04-13
cargo-expand for procedural macros


XBagon:
I wonder if using cargo-expand for the expansion of procedural macros already was discussed. I think it would make for a nice QoL feature. Instead of `cargo expand path::to::module`, having the ability to just right-click -> expand, would be very neat. I'm not sure how "good" the code output is, so I don't know if this has any use for code completion.

PR #3736, opened 2019-04-19
INT: Expand mod with `cargo expand`

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3701.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
@@ -15,7 +15,6 @@ import com.intellij.execution.process.ProcessOutput
 import com.intellij.notification.NotificationType
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
-import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.util.execution.ParametersListUtil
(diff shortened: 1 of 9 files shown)


vlad20012:
Maybe hint that it is "cargo expand" to make it searchable, e.g. rename it to "Show the result of macro expansion (cargo expand)"?

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -48,6 +49,32 @@ private val LOG = Logger.getInstance(Cargo::class.java)
  * because the user can always just `rm ~/.cargo/bin -rf`.
  */
 class Cargo(private val cargoExecutable: Path) {
+
+    data class Package(val name: String, val version: SemVer? = null) {
```
vlad20012: This can be confused with `CargoWorkspace.Package`, consider rename it to something

mchernyavsky:
Fixed.

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
mchernyavsky: Fixed.

vlad20012: Hmm, `crate` is too confusing to... Maybe BinaryPackage?

mchernyavsky: Now I use `BinaryCrate` since this name is used in help info (`cargo install --help`):
> This command manages Cargo's local set of installed binary crates.

Issue #3180, opened 2018-12-24
Add "convert to block/typle struct" intention


vlad20012:
Add 2 intentions to convert  structs between block/tuple fields:

```rust
struct Foo(i32, i32);
```
<=>
```rust
struct Foo {
    _0: i32,
    _1: i32,
}
```

rrevenantt:
I would like to work on this issue, and I am not sure if these intentions should also convert all usages accordingly.

vlad20012:
I still think that alt+enter intentions are needed, b/c it's hard to find refactorings if you don't know about them

PR #3738, opened 2019-04-20
REF/INT:add refactoring to convert struct and its usages between block and tuple representation

rrevenantt:
This is implementation for #3180. I wasn't sure what exactly i was expected to do because no one has answered me in the issue, so I have implemented it in the way i thought would be best. I think usages conversion should not harm but I have added checkbox to disable it. For now i have done only conversion to block struct. And if you like my implementation, I will add other way conversion later.

When I thought that it might be better to implement it as refactoring, I found that there was already a draft implementation in `RsConvertToNamedFieldsAction.kt`. So I just extended it.


vlad20012:
Hello. I'm sorry for the long response.

Heh, looks like my original issue was implemented already and I didn't know about it :)
However, your work is awesome, now we really can call it "refactoring".

Looks like some tests are failed. It's because you use `SearchUtils` that is not available on CLion. Please, use `ReferenceSearch` class directly

vlad20012:
Thanks!
bors r+

Issue #3677, opened 2019-04-10
[feature suggestion]: auto-delete corresponding parenthesis


DjebbZ:
Hello,

I'm learning Rust using intellij-rust and found a minor annoyance that could be solved. This is a feature suggestion, not even a feature request, I'm humbly asking for your time and don't have the skills to contribute (Rust + Kotin + Intellij etc. are all foreign to me).

### Description

*Legend: `|` is the cursor.
When I type a `(` intellij-rust auto-insert the corresponding closing paren `)` so I end up with `(|)`. But if my cursor is inside empty parens `(|)` and I press Backspace to delete the opening parens, I end with a dangling closing parens `|)`.

### Proposed solution

Deleting an opening parens should automatically delete the corresponding closing parens:

`(|)` => `Backspace` => `|`

I've just checked: it works natively with IntellJ Ultimade when editing Java and JavaScript code. I don't know of IntelliJ Community but I suppose it's the same. I suppose it's possible to implement the feature in intellij-rust since it works for other languages.

Thanks in advance!

ortem:
Hi @DjebbZ,

Does your problem occurs only in case of brackets `[]` somewhere before parentheses `()`? For example, in such code:
```Rust
fn foo() {
    let a = [];
    let b = 123;
    (|);
    let c = 42;
}
```

DjebbZ:
Not sure I understand what you mean, but in the example you've just shown,
pressing BACKSPACE delete only the opening parenthesis, not both.

ortem:
@DjebbZ IntelliJ Rust already supports brace matching and auto-deletion corresponding parentheses. So this issue relates to an existing bug, but not to new feature suggestion. This #3740 PR fixes your problem in case of brackets before parentheses (e.g. it fixes the case I showed above). The question is does it fixes your problem completely? I mean are there some other cases except brackets before parentheses?

DjebbZ:
Sorry, "brackets before parentheses" is totally unclear to me. For me it means `[]()` but this is not what I'm talking about.

Basically what I'm saying is that for brackets, parentheses and the like I prefer the IDE to never allow me to be in an inconsistent state. Other examples:

``` rust
foo()|; // cursor between closing paren and column
```

Pressing BACKSPACE should not delete the closing paren but only move the cursor one character to the left, inside the empty parentheses. Pressing BACKSPACE a second time should delete both parentheses at the same time.

Is it clearer ?

[...]

Issue #2715, opened 2018-07-27
Automatically add closing parenthesis on meta tags


nicholastmosher:
Hello :)

Currently when typing a meta tag (e.g. `#[derive(Debug)]`), upon typing `#[` IntelliJ will automatically add a closing bracket to make the text into `#[^]`(where I'm using `^` to denote the cursor position. However, if you continue on to type `#[derive(^]` then IntelliJ does _not_ add a closing parenthesis, where the expected behavior (to me) would be to automatically add the closing parenthesis (as in, `#[derive(^)]`.

I'd be willing to take a stab at fixing this if somebody could point me to where bracket completions happen. Or if it's more work than it's worth to walk me through it that's fine too.

I left the environment section in just in case I'm accidentally running on old version or something and this is actually fixed already :)

## Environment

* **Intellij-Rust plugin version:** 0.2.0.2100-181
* **Rust toolchain version:** stable-x86_64-apple-darwin
* **IDE name and version:**
IntelliJ IDEA 2018.1.6 (Ultimate Edition)
Build #IU-181.5540.7, built on July 11, 2018
* **Operating system:** macOS 10.13.5

Undin:
@nicholastmosher Hi!
As a temporary workaround, you can invoke completion while you're typing `derive` word and plugin will complete not only the word but also add parenthesis
But of course, it would be great to have parenthesis completion an in this case too.

> I'd be willing to take a stab at fixing this if somebody could point me to where bracket completions happen.

Now, we don't have code about parenthesis completion at all because in almost all cases platform completes them itself. But we have code for [angle brackes completion](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/typing/RsAngleBraceHandlers.kt). Looks like it should be a good example.

> Or if it's more work than it's worth to walk me through it that's fine too.

I don't think that it's a hard task. But some suggestions:
* We need parenthesis completion for all attributes not only for `derive`
* Don't forget to register new type handler in [plugin.xml](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/resources/META-INF/plugin.xml#L204)
* Don't forget about case when user types both `(` and `)` to avoid the following situation: you type `(`, platform adds `)` and then you type yet another `)` manually and get `())` sequence.

nicholastmosher:
Thanks for the reply @Undin! I'll try to look at this in the next few days and see if I can work anything out :)

Undin:
[...]

PR #3740, opened 2019-04-21
TYPE:fix matching brace deletion on backspace

rrevenantt:
Deletion of matching parenthesis on backspace wasn't working after brackets and until the end of file. It is very likely to be related to #3677. You can look at added tests:
before this fix `test pair parenthesis deletion after bracket` was failing, while `test pair parenthesis deletion simple` successfully passing.

The problem was that `RsAttributeParenthesisBackspaceHandler` handles backspace even if it happens not in attribute. So i have changed it to fall back to default handler which uses `RsBraceMatcher`. This is least invasive change that passes all tests. But I have looked into related code a bit more and i don't understand why do we need `RsAttributeParenthesisHandlers`? When I completely removed them, all tests were still passing. Which means `RsBraceMatcher` is enough.
So I think removing them would be a better solution unless I have missed something(which is quite likely because there should be a reason why these handlers were added). If you agree i am ready to change this PR.


ortem:
Hi @rrevenantt,

`RsAttributeParenthesisHandlers` was added due to #2715. Looks like it is really useless now (maybe because of some changes in IntelliJ Platform). So let's try to get rid of it in this PR!

rrevenantt:
Done

ortem:
Thank you!
bors r+

PR #3747, opened 2019-04-23
MACRO: fix file attributes in VFS batch

vlad20012:
The platform VFS uses more precise last-modified file attribute
than java default. We should use `FileSystemUtil.getAttributes()`
or after refresh we will think that expansion files are changed


Undin:
bors retry

Undin:
bors r=vlad20012

mchernyavsky:
bors r=vlad20012

Issue #3611, opened 2019-03-27
Auto-complete does not work with `bitflags!` generated structs


aloucks:
## Environment

* **Intellij-Rust plugin version:** Rust (0.2.95.2120-191)
* **Rust toolchain version:** rustc 1.33.0 (2aa4c46cf 2019-02-28)
* **IDE name and version:** IntelliJ IDEA 2019.1 Build #IC-191.6183.87 March 27, 2019
* **Operating system:** Windows 10 Pro 64bit

## Problem description

Auto-completion does not work for the associated constants of structs created with `bitflags!`.

In addition, Goto Declaration fails when structs are imported with **Rust 2018** edition syntax:

```rust
use bitflags::bitflags;
```

Using **Rust 2015** edition syntax enables Goto Declaration to work, but the auto-completion still does not.

```rust
#[macro_use]
extern crate bitflags;
```

## Steps to reproduce

Cargo.toml

```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2018"

[dependencies]
bitflags = "1.0.4"
```

main.rs (2018 edition syntax)

```rust
use bitflags::bitflags;

bitflags! {
    struct Foo: u32 {
        const A = 0b00000000;
        const B = 0b10000000;
    }
}

fn main() {
    Fo<auto-complete-here-does-NOT-find-the-struct>
    Foo::<auto-complete-here-does-NOT-find-A-or-B>
}
```

main.rs (2015 edition syntax)

```rust
#[macro_use]
extern crate bitflags;

bitflags! {
    struct Foo: u32 {
        const A = 0b00000000;
        const B = 0b10000000;
    }
}

fn main() {
    Fo<auto-complete-finds-the-struct>
    Foo::<auto-complete-here-does-NOT-find-A-or-B>
}
```

It's understandable that general support for macro generated structs may be a long way off (or impossible?), but it would be nice if support for `bitflags!` could be baked in, considering it's the 3rd most downloaded crate on [crates.io](https://crates.io/).

vlad20012:
There are two problems:
1. Autoimport from macro expansions. Will be automatically fixed with #3015
2. `bitflags` with 2018 edition syntax. Looks like we should support attribute [`#[macro_export(local_inner_macros)]`](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-using-local_inner_macros).

aloucks:
[...]

PR #3750, opened 2019-04-23
Support `#[macro_export(local_inner_macros)]`

vlad20012:
Enables support for [`#[macro_export(local_inner_macros)]`](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-using-local_inner_macros). Works only with the new macro expansion engine (see #3628).
Part of #3611


aloucks:
Is this blocked on anything?

vlad20012:
@aloucks Looks like we just forget about it :(

aloucks:
Still forgotten? :)

PR #3753, opened 2019-04-24
PSI: Introduce RsCodeFragment

ortem:
This PR introduces `RsCodeFragment` to be used in `Add format string argument` intention, `Evaluate expression`, and other future stuff.

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -87,6 +87,8 @@
     elementType(".+BinExpr") = BinaryExpr
     elementType(".+BinOp") = BinaryOp
 
+    elementTypeFactory("ExprCodeFragment") = "org.rust.lang.core.psi.RsCodeFragmentElementTypeBaseKt.factory"
+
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -27,22 +27,32 @@ import org.rust.lang.core.types.ty.Mutability.IMMUTABLE
...
+) {
     fun createFile(text: CharSequence): RsFile =
         PsiFileFactory.getInstance(project)
             .createFileFromText(
                 "DUMMY.rs",
                 RsFileType,
                 text,
                 /*modificationStamp =*/ LocalTimeCounter.currentTime(), // default value
-                /*eventSystemEnabled =*/ false, // default value
+                /*eventSystemEnabled =*/ eventSystemEnabled, // default value
```
vlad20012: This change isn't needed?

PR #3760, opened 2019-04-25
GRD: General improvements

ice1000:
Signed-off-by: ice1000 <ice1000kotlin@foxmail.com>

This PR:
+ Fixed a deprecated issue
+ Use `main` instead of `getByName("main")` because it's now added to gradle.kts API back

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,7 +57,7 @@ allprojects {
 
     repositories {
         mavenCentral()
-        maven { setUrl("https://dl.bintray.com/jetbrains/markdown") }
+        maven("https://dl.bintray.com/jetbrains/markdown")
     }
 
     idea {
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -35,7 +35,7 @@ plugins {
     idea
     kotlin("jvm") version "1.3.30"
     id("org.jetbrains.intellij") version "0.4.8"
-    id("org.jetbrains.grammarkit") version "2018.2.2"
+    id("org.jetbrains.grammarkit") version "2019.1"
```
Undin: It doesn't work with CLion because of https://github.com/JetBrains/gradle-grammar-kit-plugin/issues/17
It's already fixed but a new version with the fix is not published yet

build.gradle.kts
```diff
@@ -100,12 +100,12 @@ allprojects {
     }
 
     sourceSets {
-        getByName("main").apply {
+        main {
```
Undin: 👍

build.gradle.kts
ice1000: Oops. I'll revert it.

ice1000: Reverted

Undin:
Thanks!
bors r+

PR #3761, opened 2019-04-25
GRD: Try to find local idea/clion installed by toolbox

ice1000:
Signed-off-by: ice1000 <ice1000kotlin@foxmail.com>

So in most cases people won't need to wait for gradle download a giant dependency.

macOS is not supported because I don't have a mac and I have no idea where toolbox will install your IDE.
I can add macOS if anyone can tell me where the default installation path is.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,6 +25,9 @@ val platformVersion = prop("platformVersion")
 val baseIDE = prop("baseIDE")
 val ideaVersion = prop("ideaVersion")
 val clionVersion = prop("clionVersion")
+val ideaLocalPath = sequenceOf("IDEA-JDK11", "IDEA-C-JDK11", "IDEA-U", "IDEA-C")
+    .map { findInToolbox(it, ideaVersion) }.filterNotNull().firstOrNull()?.absolutePath
+val clionLocalPath = findInToolbox("CLion", clionVersion)?.absolutePath
 val baseVersion = when (baseIDE) {
     "idea" -> ideaVersion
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -266,6 +270,7 @@ project(":idea") {
 project(":clion") {
     intellij {
         version = clionVersion
+        alternativeIdePath = clionLocalPath
```
Undin: It affects nothing because this property makes sense only for root module

ice1000: Ah, I didn't know this. I thought that was for the `runIde` action.

build.gradle.kts
```diff
@@ -526,3 +532,25 @@ fun List<String>.execute(wd: String? = null, ignoreExitCode: Boolean = false): S
...
+        os.startsWith("Windows") -> "C:\\Users\\$user\\AppData\\Local\\JetBrains\\Toolbox\\apps"
+        os == "Linux" -> "/home/$user/.local/share/JetBrains/Toolbox/apps"
+        else -> return null
+    }
+    return file(root)
+        .resolve(ide)
+        .resolve("ch-0")
+        .listFiles()
+        .orEmpty()
+        .filter { it.isDirectory && it.name.startsWith(expectedMajorVersion) }
```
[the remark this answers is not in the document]
ice1000: That parameter is supposed to be assigned as `183`, `191`, etc

ice1000:
@Undin the reason looks quite reasonable to me. Thanks for your explanation!

PR #3768, opened 2019-04-28
RES: finally support unified paths (2018 edition)

### LINKED ISSUE #3199: Imports incorrectly resolve to crate root modules instead of dependencies
## Environment

* **Intellij-Rust plugin version:** v0.2.0.2511-183-nightly
* **Rust toolchain version:** stable (rustc 1.31.1)
* **IDE name and version:** CLion 2018.3.2
* **Operating system:** Linux

## Problem description
(using 2018 edition)

When using an import like `use lib1::fn1;`, if there is a `mod lib1` in the current crate root, the plugin thinks that this module is being imported. But this import actually refers to the `lib1` dependency crate. If there is no such dependency, this is a compilation error (at least on stable), so this import can't refer to `mod lib1` in either case.

## Steps to reproduce
[minimal_example.zip](https://github.com/intellij-rust/intellij-rust/files/2714949/minimal_example.zip)

Open `lib2` project in the IDE. `fn1();` invokation is (incorrectly) displayed in red color, but the code actually compiles. If `mod lib1 {}` is commented out, the red highlight disappears. The problem persists if the import is not in the crate root.

If `pub fn fn1() {}` is added inside `mod lib1 {}`, the red highlight also disappears, and the plugin thinks that `fn1()` is imported from `mod lib1`. But the compiler actually rejects the "ambiguous" import.

The expected behavior is that the plugin accepts the correct version (see attached file) and doesn't show red highlight.

### LINKED ISSUE #3801: Cannot find declaration if declared with `pub use`
## Environment

* **Intellij-Rust plugin version:** v0.2.97.2123-191
* **Rust toolchain version:** 1.34.0
* **IDE name and version:** IntelliJ IDEA 2019.1.1 (Community Edition)
Build #IC-191.6707.61, built on April 16, 2019
JRE: 1.8.0_202-release-1483-b44 x86_64
* **Operating system:** macOS 10.14

## Problem description
A function declared with `pub use` cannot be resolved from outside the module:

Bildschirmfoto 2019-05-06 um 16 17 58

## Steps to reproduce
`main.rs`:
```
mod foo {
    pub use bar::baz;

    mod bar {
        pub fn baz() {}
    }
}

fn main() {
    crate::foo::baz();
}
```

vlad20012:
Fixes #3199, fixes #3801

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -464,8 +464,9 @@ private fun processUnqualifiedPathResolveVariants(
 
     val containingMod = path.containingMod
     val crateRoot = path.crateRoot
-    val isAbsolute = path.hasColonColon
-    if (!isAbsolute) {
+    /** Path starts with `::` */
+    val hasColonColon = path.hasColonColon
+    if (!hasColonColon) {
         run {
             // hacks around $crate macro metavar. See `expandDollarCrateVar` function docs
             val referenceName = path.referenceName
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -1220,30 +1221,34 @@ private fun processLexicalDeclarations(
...
     }) {
         return true
     }
 
     // Prelude is injected via implicit star import `use std::prelude::v1::*;`
     if (processor(ScopeEvent.STAR_IMPORTS)) return false
 
+    if (Namespace.Types in ns) {
+        val result = processWithShadowing(prevScope, processor) { shadowingProcessor ->
+            processExternCrateResolveVariants(scopeStart, isCompletion, shadowingProcessor)
```
Undin: Hm, interesting
The following code
```rust
fn main() {
    extern_crate_name::some_function_name();
}
```
really works with 2015 edition (with the current stable compiler) even without `extern crate` item. When did it become possible?

vlad20012: With Rust 1.30, see https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html#module-system-improvements

(and we support it already, I just moved the code from another place to here)

src/test/kotlin/org/rust/lang/core/resolve/RsStdlibResolveTest.kt
```diff
@@ -41,6 +42,15 @@ class RsStdlibResolveTest : RsResolveTestBase() {
         fn main() {}
     """)
 
+    // TODO handle 2018 edition in std itself
+    fun `test BTreeMap`() = expect<IllegalStateException> {
+    stubOnlyResolve("""
```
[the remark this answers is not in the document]
vlad20012: It's OK for "ignored" tests starting with `expect<IllegalStateException> {`. We will have less changed lines in git patch when the test will become live.

vlad20012:
Fixed. I also added a comment about "extern_prelude" and rebased to the master

PR #3770, opened 2019-04-28
PERF: use green stubs everywhere

vlad20012:
"Green stub" can co-exists with AST. I think we should mostly use
green stubs (just like Java does) because almost all operations
are faster on stubs then on AST


mchernyavsky:
bors ping

PR #3772, opened 2019-04-28
COMP: add `deref` postfix template

mibac138:


src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -67,6 +67,14 @@ class RefmPostfixTemplate(provider: RsPostfixTemplateProvider) :
     override fun getElementToRemove(expr: PsiElement): PsiElement = expr
 }
 
+class DerefPostfixTemplate(provider: RsPostfixTemplateProvider) :
+    StringBasedPostfixTemplate("deref", "*expr", RsTopMostInScopeSelector(), provider) {
```
Undin: Now, it's available for any expr. But it makes no sense in some cases. For example, for integers.
So, I think it's better to provide `deref` postfix template only for pointers and types which implement `Deref` trait. You can look at `IterPostfixTemplate` - it's available only for types that implement `IntoIterator` trait

mibac138: That's a good point, thanks.

PR #3779, opened 2019-05-02
try to fetch stdlib from sysroot even without rustup

matklad:
See

https://internals.rust-lang.org/t/new-style-chocolatey-packages-windows/9890/6

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
+++ b/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
@@ -98,7 +98,7 @@ class RustProjectSettingsPanel(
                 val toolchain = RustToolchain(Paths.get(pathToToolchain))
                 val rustcVersion = toolchain.queryVersions().rustc?.semver
                 val rustup = toolchain.rustup
-                val stdlibLocation = rustup?.getStdlibFromSysroot()?.presentableUrl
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
```diff
@@ -98,7 +98,7 @@ class RustProjectSettingsPanel(
                 val toolchain = RustToolchain(Paths.get(pathToToolchain))
                 val rustcVersion = toolchain.queryVersions().rustc?.semver
                 val rustup = toolchain.rustup
-                val stdlibLocation = rustup?.getStdlibFromSysroot()?.presentableUrl
+                val stdlibLocation = toolchain.getStdlibFromSysroot(cargoProjectDir)?.presentableUrl
```
Undin: Looks like it's not enough to just fetch stdlib path here
You also need to set this path to `pathToStdlibField`. Now we set the corresponding path only when `rustup != null`

matklad: Hm, yeah, looks like `if (rustup == null)` logic is pretty ingrained everywhere :-( `CargoProjectImpl.refreshStdlib` does this as well. 

I **might** try to fix this properly if I have time to spare, but this is pretty unlikely: it seems like we might want to do a medium-sized refactoring of this logic, and the relevant info is out of cache for me :)

PR #3785, opened 2019-05-04
CARGO: Support crate version and version requirements parsing

mibac138:
2nd out of 4 (#3783) groundwork PRs thanks to which there will be in turn added support for newer crate version intention and an error annotator detecting common mistakes in dependencies. The PRs are divided so that it's 
easier to review them.

Final effect once the groundwork and 3 feature (new version intention, error checking, warn on future deprecation) PRs will be merged:
version-404
name-404

update

Edit: I've moved `CrateVersion` and `CrateVersionReq` from the `toml` module to `intellij-toml` as I need these classes in a Rust inspection
Edit 2: Fixed a small parsing bug and properly implemented `compareTo` for `CrateVersion`
Edit 3: Added `tryParse` convenience function
Edit 4: Rebased

PR #3783, opened 2019-05-03
PSI: Add TomlPsiFactory

mibac138:
1st out of 4 groundwork PRs thanks to which there will be in turn added support for newer crate version intention and an error annotator detecting common mistakes in dependencies. The PRs are divided so that it's 
easier to review them.

Final effect once the groundwork and 2 feature (new version intention, error checking) PRs will be merged:
version-404
name-404

update

PR #3786, opened 2019-05-04
CARGO: Improve accessing creates.io API

mibac138:
3rd out of 4 groundwork PRs (#3783, #3785) thanks to which there will be in turn added support for newer crate version intention and an error annotator detecting common mistakes in dependencies. The PRs are divided so that it's easier to review them. *This PR does not depend on the other ones*

Final effect once the groundwork and 2 feature (new version intention, error checking) PRs will be merged:
version-404
name-404

update

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/CratesIoApi.kt
+++ b/toml/src/main/kotlin/org/rust/toml/CratesIoApi.kt
@@ -22,8 +22,6 @@ import java.io.IOException
 
 data class SearchResult(val crates: List<CrateDescription>)
 
-data class CrateInfoResult(val crate: CrateDescription)
-
 data class CrateDescription(
     val name: String,
     @SerializedName("max_version")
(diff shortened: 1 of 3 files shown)


toml/src/main/kotlin/org/rust/toml/CratesIoApi.kt
```diff
@@ -36,21 +34,36 @@ val CrateDescription.dependencyLine: String
 fun searchCrate(key: TomlKey): Collection<CrateDescription> {
     if (isUnitTestMode) return MOCK!!
 
-    val name = CompletionUtil.getOriginalElement(key)?.text ?: ""
+    // Stuff like [dependencies.'crate'] can contain single quotes which we need to remove or otherwise the crates API
```
ortem: What kind of strings can be inside `''`? This function will not work for strings containing whitespaces because they cannot be passed as `q=$name`

toml/src/main/kotlin/org/rust/toml/CratesIoApi.kt
```diff
@@ -36,21 +34,36 @@ val CrateDescription.dependencyLine: String
...
+)
+
+data class CrateVersionDescription(val id: Int, val num: String, val yanked: Boolean)
+
+private class CratesIoApiResponse(val crate: CrateDescription, val versions: List<CrateVersionDescription>)
+
+fun getCrateFullDescription(crateName: TomlKey): CrateFullDescription? =
+// Stuff like [dependencies.'crate'] can contain single quotes which we need to remove or otherwise the crates API
+    // request would not work
+    getCrateFullDescription(crateName, crateName.text.removeSurrounding("'"))
```
ortem: Maybe extract this code to avoid duplication (if such simple removal really works)?

toml/src/main/kotlin/org/rust/toml/CratesIoApi.kt
mibac138: Switched to using `Url` instead of `String` to properly handle this and other special chars

toml/src/test/kotlin/org/rust/toml/CargoTomlDependenciesCompletionTest.kt
```diff
@@ -165,6 +165,16 @@ class CargoTomlDependenciesCompletionTest : CargoTomlCompletionTestBase() {
         }
     }
 
+    private fun doCompletionTest(
+        @Language("TOML") before: String,
+        @Language("TOML") after: String,
+        crate: Pair<String, String>
+    ) {
+        withMockedFullCrateDescription(CrateFullDescription(crate.first, crate.second, emptyList())) {
```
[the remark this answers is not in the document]
mibac138: > Why emptyList()?

Because the current production code leverages `maxVersion` instead of the versions list. I have however replaced this with a proper version now.

Kobzol:
Superseded by https://github.com/intellij-rust/intellij-rust/pull/6390. I will try to port the rest of your Cargo.toml PRs over the local registry index eventually.

PR #3788, opened 2019-05-04
INSP: Fix false positive for RsDropRefInspection

### LINKED ISSUE #3606: False positive with "Dropping reference" inspection
## Environment

* **Intellij-Rust plugin version:** v0.2.95.2120-183
* **Rust toolchain version:** 1.35.0-nightly
* **IDE name and version:** CLion 2018.3.4
* **Operating system:** Ubuntu 18.10

## Problem description

The plugin appears to confuse `std::mem::ManuallyDrop::drop` with `std::mem::drop` and reports a false positive when using the former.

## Steps to reproduce

Minimal example:

```rust
let mut drop = ManuallyDrop::new("nodrop");
unsafe { ManuallyDrop::drop(&mut drop) };
```

mibac138:
Fixes #3606

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsDropRefInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsDropRefInspection.kt
@@ -11,6 +11,9 @@ import org.rust.lang.core.psi.RsCallExpr
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.RsPathExpr
 import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.RsAbstractableOwner
+import org.rust.lang.core.psi.ext.owner
+import org.rust.lang.core.psi.ext.qualifiedName
 import org.rust.lang.core.types.ty.TyReference
 import org.rust.lang.core.types.type
 
@@ -30,7 +33,9 @@ class RsDropRefInspection : RsLocalInspectionTool() {
         val pathExpr = expr.expr as? RsPathExpr ?: return
 
         val resEl = pathExpr.path.reference.resolve()
-        if (resEl !is RsFunction || resEl.crateRelativePath != "::mem::drop") return
+        if (resEl !is RsFunction ||
+            resEl.qualifiedName != "core::mem::drop" ||
+            resEl.owner != RsAbstractableOwner.Free) return
 
         val args = expr.valueArgumentList.exprList
         if (args.size != 1) return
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/inspections/RsDropRefInspection.kt
```diff
@@ -30,7 +33,9 @@ class RsDropRefInspection : RsLocalInspectionTool() {
         val pathExpr = expr.expr as? RsPathExpr ?: return
 
         val resEl = pathExpr.path.reference.resolve()
-        if (resEl !is RsFunction || resEl.crateRelativePath != "::mem::drop") return
+        if (resEl !is RsFunction ||
```
vlad20012: The power of [KnownItems](https://github.com/intellij-rust/intellij-rust/blob/3f96f1509a46d4a61bc933df4e070c1c7fdb8a4a/src/main/kotlin/org/rust/lang/core/resolve/KnownItems.kt#L42) can be used here, so this condition simplifies to something like `resEl == pathExpr.knownItems.drop`

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveRefFix.kt
```diff
@@ -31,7 +36,26 @@ class RemoveRefFix(
         if (argEl.text != null && argEl.text[0] == '&') {
             val offset = argEl.startOffset
             val document = PsiDocumentManager.getInstance(project).getDocument(argEl.containingFile)
-            document?.deleteString(offset, offset + 1)
+
+            // 1. If the reference is a mut one (&mut X) we want to remove the `mut` part too as `mut X` is either not
```
vlad20012: This code is too complex because the author of the original quickfix decided to work with a text instead of PSI. If work with PSI, the implementation of `invoke` may looks like this
```rust
if (argEl is RsUnaryExpr && argEl.operatorType in listOf(UnaryOperator.REF, UnaryOperator.REF_MUT)) {
    argEl.replace(argEl.expr)
}
// yes, this is entire impl of `invoke` =/
```

src/main/kotlin/org/rust/ide/inspections/RsDropRefInspection.kt
mibac138: I'm not sure how can I use it. It doesn't contain a `drop` property and most of its properties are traits while I am checking if the element is a free drop function. Am I missing something here?

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveRefFix.kt
mibac138: Thanks! Totally forgot I could just do that

src/main/kotlin/org/rust/ide/inspections/RsDropRefInspection.kt
vlad20012: Hmm, I thought that it's easy to add a `drop `property to KnownItems, but now I see that not =/
It's designed to resolve only resolve `type` paths, so function will not be resolved... So ok, lets leave it as is for now

PR #3789, opened 2019-05-04
ANN: Detect E0040: Explicit destructor call

mibac138:
Related: #886

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -428,6 +428,21 @@ sealed class RsDiagnostic(
         }
     }
 
+    class ExplicitCallToDrop(
+        element: PsiElement
+    ) : RsDiagnostic(element) {
+        override fun prepare() = PreparedAnnotation(
+            ERROR,
+            E0040,
+            errorText(),
+            fixes = listOf(ReplaceWithStdMemDropFix(element.parent)) // e.parent: fn's name -> RsMethodCall or RsCallExpr
+        )
+
+        private fun errorText(): String {
+            return "Explicit calls to `drop` are forbidden. Use `std::mem::drop` instead"
+        }
+    }
+
     class TraitParamCountMismatchError(
         element: PsiElement,
         private val fn: RsFunction,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -428,6 +428,25 @@ sealed class RsDiagnostic(
         }
     }
 
+    class ExplicitCallToDrop(
+        element: PsiElement
+    ) : RsDiagnostic(element) {
```
Undin: It's better to pass only identifier of method call/call expr not to highlight argument list because it can interfere with some other highlighting (for example, possible error annotation about incorrect parameters)

src/test/kotlin/org/rust/ide/annotator/fixes/ReplaceWithStdMemDropFixTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+package org.rust.ide.annotator.fixes
+
+import org.rust.ProjectDescriptor
+import org.rust.WithStdlibAndDependencyRustProjectDescriptor
+import org.rust.ide.annotator.RsAnnotatorTestBase
+import org.rust.ide.annotator.RsErrorAnnotator
+
+// The Drop trait (which must be recognized for this to work) isn't recognized without stdlib.
+@ProjectDescriptor(WithStdlibAndDependencyRustProjectDescriptor::class)
```
Undin: You don't need this descriptor here because it adds some additional dependency crates besides stdlib.
Use `WithStdlibRustProjectDescriptor` if you want to add only stdlib or (that even better) you can mock `Drop` trait here and don't add stdlib in these tests at all.
Just add
```rust
#[lang = "drop"]
pub trait Drop { 
    fn drop(&mut self); 
}
```
to each test case. Note, it will work only if you apply suggestions about qualified name checking in the inspection

mibac138:
Thanks for your review, @Undin. I found your comments to be helpful and insightful. I believe I have now addressed all of them.

Undin:
@mibac138 Could you rebase your branch onto master to resolve conflicts?
> Edit: I'm not sure how this failing test is related to my code. Is this a spurious failure?

Not sure about your local tests but CI tests passed
Maybe @vlad20012 knows more about it

mibac138:
@Undin Rebased

> Not sure about your local tests but CI tests passed

That's really weird as I have just copied that error from Travis-CI, on `clion-191/nightly-rust`.

vlad20012:
> test vfs file change

Yep, it fails randomly sometimes =(

PR #3790, opened 2019-05-05
INSP: Using a function that is going to be deprecated should warn

### LINKED ISSUE #3681: Items shown as deprecated even when deprecated only in a future version.
## Environment

* **Intellij-Rust plugin version:** v0.2.96.2122-191
* **Rust toolchain version:** 1.33.0
* **IDE name and version:** CLion 2019.1

## Problem description

Functions are shown as deprecated even when they are only scheduled to be deprecated in a future version. This is a problem since that often means that there are no alternatives yet, especially when using stable Rust.

## Steps to reproduce

The following code demonstrate the issue.

```rust
fn main() {
    let _i: u8 = mem::zeroed();
}
```

The `mem::zeroed` function is declared with

```rust
#[rustc_deprecated(since = "2.0.0", reason = "use `mem::MaybeUninit::zeroed` instead")]
```

However, `mem::MaybeUninit` is not available on stable.

Thank you

mibac138:
Fixes: #3681

changelog: Deprecated item inspection now correctly handles items that will be deprecated in a future version.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -333,7 +333,7 @@ project(":") {
             exclude(module = "kotlin-stdlib")
             exclude(module = "kotlin-stdlib-common")
         }
-        implementation("com.vdurmont:semver4j:3.1.0")
+        api("com.vdurmont:semver4j:3.1.0")
         testImplementation(project(":common", "testOutput"))
         testImplementation("com.squareup.okhttp3:mockwebserver:4.9.0")
     }
@@ -446,7 +446,6 @@ project(":toml") {
     }
     dependencies {
         implementation("org.eclipse.jgit:org.eclipse.jgit:5.9.0.202009080501-r") { exclude("org.slf4j") }
-        implementation("com.vdurmont:semver4j:3.1.0")
 
         implementation(project(":"))
         implementation(project(":common"))
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -333,6 +333,7 @@ project(":") {
             exclude(module = "kotlin-stdlib")
             exclude(module = "kotlin-stdlib-common")
         }
+        implementation("com.vdurmont:semver4j:3.1.0")
         testImplementation(project(":common", "testOutput"))
         testImplementation("com.squareup.okhttp3:mockwebserver:4.9.0")
     }
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsDeprecationInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsDeprecationInspection.kt
@@ -5,7 +5,11 @@
 
 package org.rust.ide.inspections.lints
 
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemHighlightType.*
 import com.intellij.psi.PsiElement
+import com.vdurmont.semver4j.Semver
+import com.vdurmont.semver4j.SemverException
 import org.rust.ide.inspections.RsProblemsHolder
 import org.rust.lang.core.psi.RsElementTypes.CSELF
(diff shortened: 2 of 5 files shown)


Kobzol:
Rebased over master. The inspection now needs to parse crate versions, so I included the semver parsing dependency into the root module.

src/main/kotlin/org/rust/ide/inspections/lints/RsDeprecationInspection.kt
```diff
@@ -43,24 +47,41 @@ class RsDeprecationInspection : RsLintInspection() {
     private fun checkAndRegisterAsDeprecated(identifier: PsiElement, original: PsiElement, holder: RsProblemsHolder) {
         if (original is RsOuterAttributeOwner) {
             val attr = original.queryAttributes.deprecatedAttribute ?: return
-            holder.registerLintProblem(identifier, attr.extractDeprecatedMessage(identifier.text))
+            val (message, highlightType) = attr.extractDeprecatedMessage(identifier.text)
+            holder.registerProblem(identifier, message, highlightType)
```
Undin, line 51: Note, `registerLintProblem` takes into account the current [lint level](https://doc.rust-lang.org/rustc/lints/levels.html) of the corresponding lint but new implementation doesn't do it, as I can see

Kobzol: Ah, I was wondering if there's any problem with not using `registerLintProblem`. The current lint API is slightly cumbersome if the lint has several highlight types for various situations.

It would be nice to have the option to send a custom object to `registerLintProblem` that would dynamically allow selecting the highlight type. I tried to implement it, let me know what you think.

vlad20012: Interesting that rustc shows the same warnings regardless of the version. Maybe we shouldn't use different highlighting too

vlad20012: Another option is introducing another lint with the same `id`. I'm not sure whether it is ok or not

build.gradle.kts
```diff
@@ -333,6 +333,7 @@ project(":") {
             exclude(module = "kotlin-stdlib")
             exclude(module = "kotlin-stdlib-common")
         }
+        implementation("com.vdurmont:semver4j:3.1.0")
```
vlad20012, line 336:
```suggestion
        api("com.vdurmont:semver4j:3.1.0")
```
And then it's possible to remove this dependency from `toml` module

src/main/kotlin/org/rust/ide/inspections/lints/RsDeprecationInspection.kt
Kobzol: > Another option is introducing another lint with the same id. I'm not sure whether it is ok or not

That sounds.. a bit confusing :)

> Interesting that rustc shows the same warnings regardless of the version. Maybe we shouldn't use different highlighting too.

I think that we can do better than rustc here. I would also show a different highlight type, since sometimes you might not care about future versions if your version is fixed, so it might be annoying. If it's not deprecated currently, it should be OK to use, so weak warning seems right to me.

Undin (approved):
Works fine but I think the implementation is a bit more complicated than needed. I will try to simplify it in separate PR.

Also, I've found that rustc supports special value for `since` argument - `TBD` to solve the same issue for `rustc_deprecated` attribute. It was introduced by https://github.com/rust-lang/rust/pull/79877 that emits `deprecated_in_future` warning (allow by default) in such cases.
I suppose we can emit weak warning in such cases as well

Thanks @mibac138 and @Kobzol 
bors r+

PR #3793, opened 2019-05-05
ACT: Added self parameter support to type declaration provider

### LINKED ISSUE #3665: Ctrl-Click on 'self' in function args doesn't go anywhere
I can click on 'Self' and it goes to the definition. It would be really great if I could click on self, &self and &mut self in a function's arguments and they take me to the definition as well.
(In many functions, self is far more likely to be part of the sig than Self.)

Keep up the awesome work you're doing! It's going from strength to strength!

mibac138:
<details><summary>Fixed issue. Thanks to @vlad20012 and his PR #3804</summary>Currently, for reasons I believe are unrelated to my changes, it only works with the leftmost element of the self parameter (see gif below). I tried to figure out the reason for this, and so far I've discovered that the `getSymbolTypeDeclaration` function is not even invoked for elements other than the leftmost one. I'd appreciate some help on this.

goto-is-weird-2
</details>

This closes #3665.


vlad20012:
:+1: Thanks!
bors r+

PR #3794, opened 2019-05-05
ANN: Detect E0081: Duplicate enum discriminant

mibac138:
Related to #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -44,7 +44,7 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
             override fun visitConstant(o: RsConstant) = checkConstant(holder, o)
             override fun visitValueArgumentList(o: RsValueArgumentList) = checkValueArgumentList(holder, o)
             override fun visitStructItem(o: RsStructItem) = checkDuplicates(holder, o)
-            override fun visitEnumItem(o: RsEnumItem) = checkDuplicates(holder, o)
+            override fun visitEnumItem(o: RsEnumItem) = checkEnumItem(holder, o)
             override fun visitEnumVariant(o: RsEnumVariant) = checkDuplicates(holder, o)
             override fun visitFunction(o: RsFunction) = checkFunction(holder, o)
             override fun visitImplItem(o: RsImplItem) = checkImpl(holder, o)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -82,6 +82,39 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
...
     }
 
+    private fun checkEnumItem(holder: AnnotationHolder, o: RsEnumItem) {
+        checkDuplicates(holder, o)
+        o.enumBody?.let { checkDuplicateEnumVariants(holder, it) }
+    }
+
+    private fun checkDuplicateEnumVariants(holder: AnnotationHolder, o: RsEnumBody) {
+        var discrCounter = 0
+        val indexToVariantMap = mutableMapOf<Int, Pair<RsEnumVariant, /* already reported */Boolean>>()
```
mchernyavsky: Let's replace `Pair<RsEnumVariant, Boolean>` with some data class, for example:
```kotlin
private data class VariantInfo(val variant: RsEnumVariant, val alreadyReported: Boolean)
```
After some refactoring:
```kotlin
private fun checkDuplicateEnumVariants(holder: AnnotationHolder, o: RsEnumBody) {
    var discrCounter = 0
    val indexToVariantMap = hashMapOf<Int, VariantInfo>()
    for (variant in o.enumVariantList) {
        val literal = variant.variantDiscriminant?.expr as? RsLitExpr
        val int = literal?.integerLiteralValue?.toIntOrNull()
        val idx = int ?: discrCounter
        discrCounter = idx + 1

        val previous = indexToVariantMap[idx]
        if (previous != null) {
            RsDiagnostic.DuplicateEnumDiscriminant(variant, idx).addToHolder(holder)
            if (!previous.alreadyReported) {
                RsDiagnostic.DuplicateEnumDiscriminant(previous.variant, idx).addToHolder(holder)
                indexToVariantMap[idx] = previous.copy(alreadyReported = true)
            }
        } else {
            indexToVariantMap[idx] = VariantInfo(variant, alreadyReported = false)
        }
    }
}
```

mibac138: Thanks! It does look better now

mchernyavsky:
👍 bors r+

mibac138:
@mchernyavsky I don't think bors understood that

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -82,6 +82,35 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
...
+    }
+
+    private fun checkDuplicateEnumVariants(holder: AnnotationHolder, o: RsEnumBody) {
+        data class VariantInfo(val variant: RsEnumVariant, val alreadyReported: Boolean)
+
+        var discrCounter = 0
+        val indexToVariantMap = hashMapOf<Int, VariantInfo>()
+        for (variant in o.enumVariantList) {
+            val literal = variant.variantDiscriminant?.expr as? RsLitExpr
+            val int = literal?.integerLiteralValue?.toIntOrNull()
```
vlad20012: Should we use Long instead?

Undin: Will be fixed in #3796

Undin: Also, it will automatically support binary, octal and hex literals with a new API for literal values

mibac138: @Undin Actually, won't BigInteger be better? With `#![feature(repr128)]` you can have the discriminant be over 64 bits long. I'm not sure if that's justification enough though?

PR #3795, opened 2019-05-05
ANN: Detect E0120: impl Drop for Trait is illegal

mibac138:
Related to #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -26,10 +26,7 @@ import org.rust.lang.core.resolve.knownItems
 import org.rust.lang.core.resolve.namespaces
 import org.rust.lang.core.resolve.ref.deepResolve
 import org.rust.lang.core.types.inference
-import org.rust.lang.core.types.ty.Ty
-import org.rust.lang.core.types.ty.TyUnit
-import org.rust.lang.core.types.ty.isSelf
-import org.rust.lang.core.types.ty.isSized
+import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
 import org.rust.lang.utils.RsDiagnostic
 import org.rust.lang.utils.RsErrorCode
@@ -330,6 +327,7 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
     private fun checkImpl(holder: AnnotationHolder, impl: RsImplItem) {
         val traitRef = impl.traitRef ?: return
         val trait = traitRef.resolveToTrait ?: return
+        checkImplDropForNonAdtError(holder, impl, traitRef, trait)
         val traitName = trait.name ?: return
 
         fun mayDangleOnTypeOrLifetimeParameters(impl: RsImplItem): Boolean {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -297,6 +298,13 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    private fun checkImplDropForTrait(holder: AnnotationHolder, impl: RsImplItem, traitRef: RsTraitRef, trait: RsTraitItem) {
+        if (impl.typeReference?.baseType?.path?.reference?.resolve() !is RsTraitItem) return
+        if (trait.qualifiedName == "core::ops::drop::Drop") {
```
Undin: See https://github.com/intellij-rust/intellij-rust/pull/3789#discussion_r281040984

src/test/kotlin/org/rust/ide/annotator/RsStdlibErrorAnnotatorTest.kt
```diff
@@ -24,4 +24,11 @@ class RsStdlibErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class.j
 
         <error descr="Can't find crate for `litarvan` [E0463]">extern crate litarvan;</error>
     """)
+
+    fun `test impl Drop for Trait E0120`() = checkErrors("""
```
Undin: See https://github.com/intellij-rust/intellij-rust/pull/3789#discussion_r281042654 to move this test to `RsErrorAnnotatorTest.kt`

mibac138:
@Undin I've addressed your comments

Undin:
@mibac138 This branch, as well as #3789, has conflicts with master 
But looks like it will have conflicts with master when #3789 will be merged even you rebase onto master now
So, let's rebase this branch onto master only when #3789 will be merged to avoid additional work

mibac138:
Rebased onto #3789. I think that should avoid creating conflicts once #3789 is merged?

vlad20012:
Hmm, your check is unreasonably very specific. `Drop` can be implemented only for structs and enums, so E0120 can be also showed when implementing `Drop` for reference, pointers, arrays, primitives, etc. Looks like it's easy to detect attempt to impl `Drop` for NOT struct or enum and show the error in this case

mibac138:
@vlad20012 That's a fair point, I must admit I got carried away by the trait part and forgot about other types. I have fixed that now and added tests for the types that have been added.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -353,6 +351,15 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    // E0120: Drop can be only implemented by structs and enums
+    private fun checkImplDropForNonAdtError(holder: AnnotationHolder, impl: RsImplItem, traitRef: RsTraitRef, trait: RsTraitItem) {
+        if (trait != trait.knownItems.Drop) return
+
+        if (impl.typeReference?.type is TyAdt) return
```
vlad20012: Looks dangerous for the case of Drop impl with missed type reference, i.e.
```rust
impl Drop for {}
```

Let's add a test that show that we don't have an error in this case.

mibac138:
Added more tests

vlad20012:
:+1: 
bors r+

PR #3796, opened 2019-05-05
Literal stubs & const expr evaluation

Undin:
* Save values of literals into stubs when needed. It's useful not only for const expr evaluation but for `include` macro support in future
* Provide common API for const expr evaluation
* Drop confusing `compareTo` method from `ExprValue` public API

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsLitExpr.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsLitExpr.kt
@@ -25,26 +25,29 @@ import org.rust.lang.core.psi.impl.RsExprImpl
 import org.rust.lang.core.psi.kind
 import org.rust.lang.core.stubs.RsLitExprStub
 import org.rust.lang.core.stubs.RsPlaceholderStub
-import org.rust.lang.core.stubs.RsStubLiteralType
+import org.rust.lang.core.stubs.RsStubLiteralKind
 import org.rust.lang.core.types.ty.TyFloat
 import org.rust.lang.core.types.ty.TyInteger
 
-val RsLitExpr.stubType: RsStubLiteralType? get() {
-    val stub = (greenStub as? RsLitExprStub)
-    if (stub != null) return stub.type
-    val kind = kind
-    return when (kind) {
-        is RsLiteralKind.Boolean ->  RsStubLiteralType.Boolean
-        is RsLiteralKind.Char -> RsStubLiteralType.Char(kind.isByte)
-        is RsLiteralKind.String -> RsStubLiteralType.String(kind.value?.length?.toLong(), kind.isByte)
-        is RsLiteralKind.Integer -> RsStubLiteralType.Integer(TyInteger.fromSuffixedLiteral(integerLiteral!!))
-        is RsLiteralKind.Float -> RsStubLiteralType.Float(TyFloat.fromSuffixedLiteral(floatLiteral!!))
-        else -> null
+val RsLitExpr.stubKind: RsStubLiteralKind?
+    get() {
+        val stub = (greenStub as? RsLitExprStub)
+        if (stub != null) return stub.kind
+        return when (val kind = kind) {
+            is RsLiteralKind.Boolean -> RsStubLiteralKind.Boolean(kind.value)
+            is RsLiteralKind.Char -> RsStubLiteralKind.Char(kind.value, kind.isByte)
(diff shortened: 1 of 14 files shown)


mibac138:
Do you think you could make it so that if I try to evaluate, say `RsUnaryExpr(MINUS, RsLitExpr(1))` with `expectedTy` `u8` I would know that it underflowed? My use case for this is detecting [`E0370`](https://doc.rust-lang.org/error-index.html#E0370).

Undin:
@mibac138 It's possible - you can see the corresponding `TODO` to check overflow. But I think it should be done in another PR

src/main/kotlin/org/rust/lang/core/psi/ext/RsLitExpr.kt
```diff
@@ -22,26 +25,53 @@ import org.rust.lang.core.psi.impl.RsExprImpl
...
+        is RsLiteralKind.Float -> RsStubLiteralKind.Float(kind.value, TyFloat.fromSuffixedLiteral(floatLiteral!!))
         else -> null
     }
 }
 
-val RsLitExpr.integerLiteralValue: String? get() =
-    (stub as? RsLitExprStub)?.integerLiteralValue ?: integerLiteral?.text
+val RsLitExpr.booleanValue: Boolean?
+    get() {
+        val stubKind = stubKind ?: return (kind as? RsLiteralKind.Boolean)?.value
```
ortem: Looks like if `stubKind` is `null`, then `kind` cannot be `RsLiteralKind.Boolean`. AFAIU, `stubKind` is `null` iff `kind` is `null` or `kind` is not `RsLiteralKind.<...>`.

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1063,30 +1065,24 @@ class RsInnerAttrStub(
...
     parent: StubElement<*>?, elementType: IStubElementType<*, *>,
     val name: String?,
-    val hasEq: Boolean,
-    val value: String?
+    val hasEq: Boolean
 ) : StubBase<RsMetaItem>(parent, elementType) {
     object Type : RsStubElementType<RsMetaItemStub, RsMetaItem>("META_ITEM") {
         override fun createStub(psi: RsMetaItem, parentStub: StubElement<*>?): RsMetaItemStub =
-            RsMetaItemStub(parentStub, this, psi.name, psi.eq != null, psi.litExpr?.stringLiteralValue)
+            RsMetaItemStub(parentStub, this, psi.name, psi.eq != null)
```
ortem: Did you intentionally remove `value` and not add `RsStubLiteralKind`?

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1183,45 +1177,54 @@ class RsUnaryExprStub(
...
+    class String(val value: kotlin.String?, val isByte: kotlin.Boolean) : RsStubLiteralKind(2)
+    class Integer(val value: Long?, val ty: TyInteger?) : RsStubLiteralKind(3)
+    class Float(val value: Double?, val ty: TyFloat?) : RsStubLiteralKind(4)
 
     companion object {
-        fun deserialize(dataStream: StubInputStream): RsStubLiteralType? {
-            with (dataStream) {
+        fun deserialize(dataStream: StubInputStream): RsStubLiteralKind? {
+            with(dataStream) {
                 val ordinal = readByte().toInt()
```
ortem: `ordinal` can be moved inside `when`

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
Undin: Yep
Now, literal expressions keep their values in stubs so no need to duplicate them - we can just get value from the corresponding literal expr

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
Undin: If move it to `when` it will be marked as unused because it is really not used in `when` branches. 
So we can even remove `ordinal` variable at all. But I'd prefer to keep this variable to simplify reading and debugging. So let's keep it as is to avoid unnecessary `unused variable` warning.
Btw IDEA doesn't suggest to move `ordinal` to `when`

Undin:
@ortem fixed

PR #3797, opened 2019-05-06
QDOC/LI: fixed block doc comments without infix in quickdoc and doctest

rrevenantt:
Block documentation comments without `*` infix, e.g.
``` rust
/*!
foo
bar
*/
```
were incorrectly handled by quickdoc and doctest injection. See added tests for particular cases that failed before.

### DIFF
--- a/src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
+++ b/src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
@@ -129,6 +129,7 @@ private fun findDoctestInjectableRanges(text: String, elementType: IElementType)
     if (tripleBacktickIndices.size < 2) return emptySequence() // no code blocks in the comment
 
     val infix = RsDocKind.of(elementType).infix
+    val isBlockInfixUsed = text.count { it == '\n' } == "\\n[\\s]*\\*".toRegex().findAll(text).count()
 
     return tripleBacktickIndices.asSequence().chunked(2).mapNotNull { idx ->
         // Contains code lines inside backticks including `///` at the start and `\n` at the end.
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -129,6 +129,7 @@ private fun findDoctestInjectableRanges(text: String, elementType: IElementType)
     if (tripleBacktickIndices.size < 2) return emptySequence() // no code blocks in the comment
 
     val infix = RsDocKind.of(elementType).infix
+    val isBlockInfixUsed = text.count { it == '\n' } == "\\n[\\s]*\\*".toRegex().findAll(text).count()
```
vlad20012: Looks like rustdoc uses other rules to check it. I don't sure but looks like it checks only first line of the comment body. E.g. this is comment is infixed:
image

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -153,6 +154,7 @@ private fun findDoctestInjectableRanges(text: String, elementType: IElementType)
         // skip doc comment infix (`///`, `//!` or ` * `)
         val ranges = lines.asSequence().mapNotNull { (lineStart, lineEnd) ->
             val index = text.indexOf(infix, lineStart)
+            if (!isBlockInfixUsed && RsDocKind.of(elementType).isBlock) return@mapNotNull TextRange(lineStart, lineEnd)
```
vlad20012: This can be moved out of `mapNotNull`, isn't it?

src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
rrevenantt: Are you sure? On my machine this comment 
``` rust
/*! 
* description header
* bar
foo
* bar
*/
```
results in following documentation
image

and your example is broken as well
image


which doesn't looks like rustdoc tried to remove infix

vlad20012: Hmm, really, you're right

rrevenantt:
Done.

Aren't commits squashed automatically during merge?

vlad20012:
> Aren't commits squashed automatically during merge?

Unfortunately, bors can't squash. See https://github.com/bors-ng/bors-ng/issues/194

vlad20012:
bors retry

Issue #3379, opened 2019-02-10
Parser/highlighter is fooled by line comment with extra "slash" (/) character


detro:
## Environment

* **Intellij-Rust plugin version:** v0.2.91.2115-183 
* **Rust toolchain version:**
* **IDE name and version:** IntelliJ IDEA 2018.3.4 (Ultimate Edition) (Build #IU-183.5429.30, built on January 29, 2019)
* **JRE:** 1.8.0_152-release-1343-b26 amd64
* **JVM:** OpenJDK 64-Bit Server VM by JetBrains s.r.o
* **Operating system:** Linux 4.15.0-45-generic
* **NOTE:** I was experiencing the same issue also on other OS (Win10 and MacOSX), with previous versions of the IntelliJ-Rust, as well as previous version of IntelliJ.

## Problem description

A rust-style line-comment that, by mistake, contains more than 3 "/" characters,
throws the parser off and highlight and code-completion for the parsed file doesn't work.

I discovered it when using the crate [trust-dns-proto](https://docs.rs/trust-dns-proto/0.7.1/trust_dns_proto/), file [record_data.rs](https://github.com/bluejekyll/trust-dns/master/crates/proto/src/rr/record_data.rs).

A screenshot sometimes helps more than words:
![screenshot_20190210_221318](https://user-images.githubusercontent.com/114508/52540480-c45d7100-2d81-11e9-9169-6e9f11f0f6a9.png)

As you can see, after the value enum variant `CNAME` in the screenshot, the highlighting stops (and it will stop for the rest of the file).

Of course the code is valid, so I think the parser just needs to be made a tiny bit smarter.

## Steps to reproduce

* Create a project
* Include "trust-dns-proto" as dependency
* Go at the above mentioned file

You can reproduce this even with a simple comment like:

```rust
///
//// this will be highlighted as error
///
```

PR #3800, opened 2019-05-06
GRAM:fix 4 slashes doc comment error

rrevenantt:
fix for #3379


src/main/grammars/RustLexer.flex
```diff
@@ -228,7 +228,10 @@ OUTER_EOL_DOC = ({EOL_DOC_LINE}{EOL_WS})*{EOL_DOC_LINE}
   "/*"                            { yybegin(IN_BLOCK_COMMENT); yypushback(2); }
 
   {INNER_EOL_DOC}                 { return INNER_EOL_DOC_COMMENT; }
-  {OUTER_EOL_DOC}                 { return OUTER_EOL_DOC_COMMENT; }
+  {OUTER_EOL_DOC} / [^\/]         { //fix lookahead on EOF
```
mchernyavsky: Is `\` needed here?

src/main/grammars/RustLexer.flex
```diff
@@ -228,7 +228,10 @@ OUTER_EOL_DOC = ({EOL_DOC_LINE}{EOL_WS})*{EOL_DOC_LINE}
...
+                                    if (zzMarkedPos + 1 == zzBuffer.length())
```
mchernyavsky: Can you explain what you are trying to fix here?
Now `///<EOF>` is not considered part of the doc comment:

rrevenantt: Problem I am trying to fix is that you can't add <EOF> as accepted lookahead symbol which causes last symbol not to be included into doc comment token. So I just explicitly add last symbol to token. But apparently it is not enough. Then I suppose I need to write new lexer state for line doc comments.

rrevenantt:
I have rewritten it as a separate lexer state. It should handle `<EOF>` better now.

Undin:
bors r=mchernyavsky

mchernyavsky:
bors retry

PR #3802, opened 2019-05-06
ANN: Detect E0184: Cannot implement both Copy and Drop

mibac138:
Related to #886.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -25,6 +25,7 @@ import org.rust.lang.core.resolve.Namespace
 import org.rust.lang.core.resolve.knownItems
 import org.rust.lang.core.resolve.namespaces
 import org.rust.lang.core.resolve.ref.deepResolve
+import org.rust.lang.core.types.TraitRef
 import org.rust.lang.core.types.inference
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
@@ -75,6 +76,7 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
             override fun visitBlockExpr(o: RsBlockExpr) = checkBlockExpr(holder, o)
             override fun visitBreakExpr(o: RsBreakExpr) = checkBreakExpr(holder, o)
             override fun visitContExpr(o: RsContExpr) = checkContExpr(holder, o)
+            override fun visitAttr(o: RsAttr) = checkAttr(holder, o)
         }
 
         element.accept(visitor)
@@ -328,6 +330,7 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         val traitRef = impl.traitRef ?: return
         val trait = traitRef.resolveToTrait ?: return
         checkImplDropForNonAdtError(holder, impl, traitRef, trait)
+        checkImplBothCopyAndDrop(holder, impl, trait)
         val traitName = trait.name ?: return
 
         fun mayDangleOnTypeOrLifetimeParameters(impl: RsImplItem): Boolean {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -354,13 +355,31 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
...
             RsDiagnostic.ImplDropForTraitError(traitRef, impl.typeReference?.text).addToHolder(holder)
         }
     }
 
+    // E0184: Cannot implement both Copy and Drop
+    private fun checkImplBothCopyAndDrop(holder: AnnotationHolder, impl: RsImplItem, traitRef: RsTraitRef, trait: RsTraitItem) {
+        // Only execute when the impl trait is Drop to avoid error duplication (two same errors on the same element)
+        // Also, Drop can be only implemented through impl Drop, contrary to Copy which is somewhat harder to detect
+        if (trait != trait.knownItems.Drop) return
+        val target = impl.typeReference?.baseType?.path?.reference?.resolve() as? RsStructOrEnumItemElement ?: return
```
vlad20012: Use `ImplLookup.canSelect()` and `impl.typeReference.type`

mibac138: I'm not sure it's supposed to look like. I've came up with this code
```kotlin
val self = impl.typeReference?.type ?: return
val copyImpl = (trait.implLookup.select(TraitRef(self, trait.knownItems.Copy?.withSubst() ?: return))
    .ok()?.impl as? RsImplItem)?.traitRef ?: return
```
But it doesn't pass the `derive(Copy)` test. Should I add `target.derivedTraitsToMetaItems[Copy]`?

vlad20012: > But it doesn't pass the derive(Copy) test.

Hmm, it should. I'll check it. And it's wrong to add anotation to `copyImpl` here: the plugin will be crashed if such impl is located in a different file

mibac138: @vlad20012 I believe it doesn't pass that test due to `.ok()?.impl as? RsImplItem)?.traitRef ?: return`. I'm currently working on the code so that it's working

vlad20012: Yeah, your test doesn't pass because you assume result of selection is `impl`, but it is `derive`, so it fails on `.ok()?.impl`. But the real problem is in the general approach - you shouldn't annotate some another place even if you sure that error should be there (Mostly because it can be another file).  I suggest to check&annotate only Drop impl here:
```rust
val type = impl.typeReference?.type ?: return
        val isCopyImplemented = trait.implLookup.canSelect(TraitRef(type, trait.knownItems.Copy?.withSubst() ?: return))

        if (isCopyImplemented) {
            RsDiagnostic.ImplBothCopyAndDropError(traitRef).addToHolder(holder)
        }
```

And then do a reverse thing: in the place of `Copy` impl/derive check that `Drop` isn't implemented for the same type. Yeah, it seems to hacky to check derive and impl in a separate way, but I can't think of something more clear for now. To locate derive meta item you can use `RsPsiPattern.derivedTraitMetaItem`

mibac138: @vlad20012 I've modified the code and checked that it passes the tests. Does the current implementation seem reasonable to you?

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -1654,4 +1654,28 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class.java) {
...
+        impl <error descr="Cannot implement both Copy and Drop [E0184]">Drop</error> for Foo {}
+    """)
+
+    fun `test impl Drop and impl Copy E0184`() = checkErrors("""
+        #[lang = "drop"]
+        trait Drop {}
+        #[lang = "copy"]
+        trait Copy {}
+
+        struct Foo;
```
vlad20012: Let's add a test for generic struct `struct Foo<T>(T);`. I don't sure how `implLookup.canSelect` behaves with unfolded generic types

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
vlad20012: :+1:

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
mibac138: @vlad20012 Done

vlad20012:
Could you rebase it to the master please?

mibac138:
@vlad20012 Done

PR #3803, opened 2019-05-06
 ANN: Detect E0518: Inline attribute is allowed only on functions 

mibac138:
Related to #886.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -456,6 +456,7 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
 
     private fun checkAttr(holder: AnnotationHolder, attr: RsAttr) {
         checkImplBothCopyAndDrop(holder, attr)
+        checkInlineAttr(holder, attr)
         checkReprForEmptyEnum(holder, attr)
     }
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/RemoveAttrFix.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.intellij.codeInspection.LocalQuickFixOnPsiElement
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.ext.RsAttr
+import org.rust.lang.core.psi.ext.name
+
+class RemoveAttrFix(attr: RsAttr) : LocalQuickFixOnPsiElement(attr) {
+    private val _text = "Remove attribute" + when (val name = attr.metaItem.name) {
```
mchernyavsky: You can use `run` here:
```kotlin
private val _text: String = run {
    val suffix = attr.metaItem.name?.let { " `$it`" } ?: ""
    "Remove attribute$suffix"
}
```
Also you can move this code to `getText`:
```kotlin
override fun getText(): String {
    val attr = startElement as? RsAttr
    val suffix = attr?.metaItem?.name?.let { " `$it`" } ?: ""
    return "Remove attribute$suffix"
}
```

mchernyavsky:
Could you please squash/rebase your commits?

mibac138:
@mchernyavsky done

mibac138:
@mchernyavsky done

PR #3806, opened 2019-05-07
Automatically detect "stdlib" when working with rust-lang/rust project

vlad20012:
The `https://github.com/rust-lang/rust` project (aka "rustc",
aka "rust-lang/rust") contains stdlib sources inside the project.
Now we locate such sources and do not link external stdlib to
the project. This fixes a lot of bugs when working with
rust-lang/rust project.


src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -323,6 +323,14 @@ data class CargoProjectImpl(
     }
 
     private fun refreshStdlib(): CompletableFuture<CargoProjectImpl> {
+        if (isProjectLooksLikeRustLangRust()) {
```
Undin: Maybe `isProjectLooksLikeRustc` better?

vlad20012:
> Unresolved references inspection produces a lot of false positive error annotations in doc comments with the current implementation

It's another bug, fixed in #3858

PR #3808, opened 2019-05-07
COMP: Index proc_macro_derive functions and provide completion for derive

### LINKED ISSUE #3690: Completion for derive macros
It would be nice if intellij-rust could assist in writing `#[derive(...)]` with custom macros. For example:

- Suggest derive macros that are in scope;
- Suggest derive macros that can be imported; import them automatically if the completion is confirmed;
- Highlight in red color macros that can be imported and provide an intent to import them.

Looks like the IDE understands `Serialize` and `Deserialize`, but for example it doesn't understand `structopt_derive::StructOpt`.

mibac138:
Closes #3690. This is a *work in progress* PR for indexing `proc_macro_derive` functions and adding completions for `#[derive]` attribute. At the moment there are 3 unresolved issues - type resolving during indexing, how to index macro expansions and `LookupElementBuilder`'s `InsertionHandler` not working (I've looked at `RsCommonCompletionContributor` and then `AutoImportFix` for help but as far as I understand it just happily modifies the PSI without any problems). I'd appreciate any help with resolving them.

Currently, the completion works correctly but some tests are missing, and auto-importing the crates, as well as, just adding `#[macro_use]` if the import is present, is not yet working. Also, does not yet work for macro generates `proc_macro_derive` functions.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -506,15 +506,16 @@ fun ImportCandidate.import(context: RsElement) {
     insertionScope.insertUseItem(psiFactory, "$prefix${info.usePath}")
 }
 
-private fun RsMod.insertExternCrateItem(psiFactory: RsPsiFactory, crateName: String) {
+fun RsMod.insertExternCrateItem(psiFactory: RsPsiFactory, crateName: String): RsExternCrateItem {
     val externCrateItem = psiFactory.createExternCrateItem(crateName)
     val lastExternCrateItem = childrenOfType<RsExternCrateItem>().lastElement
-    if (lastExternCrateItem != null) {
+    return if (lastExternCrateItem != null) {
         addAfter(externCrateItem, lastExternCrateItem)
     } else {
-        addBefore(externCrateItem, firstItem)
-        addAfter(psiFactory.createNewline(), firstItem)
-    }
+        addBefore(externCrateItem, firstItem).also {
+            addAfter(psiFactory.createNewline(), firstItem)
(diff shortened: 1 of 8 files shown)


vlad20012:
Oh, I should say that proc macro name resolution is partially implemented already (but not published anywhere). As far as I know, @fedochet is working on it now. I've [attached a patch](https://github.com/intellij-rust/intellij-rust/files/3156357/RES__resolve_derive_macros.patch.zip) with changes that I have. It can help.

A few notes about this PR:
1. You use old (2015 edition style) proc macro imports (with #[macro_use]). New imports are made by `use` items.
2. Note that proc macros can be re-exported (with `pub use`). It used, for example, in `serde`
3. Invoking resolve during indexing is fundamentally impossible, you should perform all additional checks (if needed) after retrieving something from an index.
4. Index macro expansions is easy now, you just should use `WithMacroProjectScope(project)` instead of `GlobalSearchScope.allScope(project)`. Note that new macro expansion engine should be enabled to make it works (see #3628)

mibac138:
@vlad20012 I had no idea. Is there a place I can check if someone is working on something to avoid situations like this in the future?
@fedochet Should we perhaps join forces? Are you working on the changes in a public repo or somewhere I could access? If you'd prefer to work on it alone though, I don't have a problem with that at all.

fedochet:
@mibac138 hi! My work is not classified, but not yet published :)

I think that I would probably make a branch in my fork of this repository in a couple of days where I will publish my changes; I will notify here you so you can have a look if you're interested

mibac138:
@fedochet Yeah, that'd be great! Thanks :)

vlad20012:
Closing in favor of auto-import rewriting #7558

PR #3813, opened 2019-05-08
ACT: Go to Cargo.toml when invoking "Go to super" in a crate root

### LINKED ISSUE #2781: Navigation from lib.rs to Cargo.toml
It would be convenient if the "go to the parent module" from lib.rs or main.rs by Ctrl-U / Cmd-U was continued to the corresponding Cargo.toml and furthermore in the same way to the parent Cargo.toml of the workspace.

mibac138:
Fixes #2781. I'd like to add more tests but I'm not sure how can I do that. Any tips?

### DIFF
--- a/src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
+++ b/src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
@@ -8,15 +8,15 @@ package org.rust.ide.navigation.goto
 import com.intellij.lang.LanguageCodeInsightActionHandler
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.LocalFileSystem
 import com.intellij.psi.NavigatablePsiElement
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiManager
 import com.intellij.psi.util.PsiTreeUtil
 import org.rust.lang.core.psi.RsFile
-import org.rust.lang.core.psi.ext.RsAbstractable
-import org.rust.lang.core.psi.ext.RsMod
-import org.rust.lang.core.psi.ext.owner
-import org.rust.lang.core.psi.ext.superItem
+import org.rust.lang.core.psi.ext.*
+import java.nio.file.Path
 
 class RsGotoSuperHandler : LanguageCodeInsightActionHandler {
     override fun startInWriteAction() = false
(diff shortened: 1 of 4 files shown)


vlad20012:
Note that `rust` and `toml` are different plugins and `rust` doesn't require `toml` to work, so it shouldn't crash if `toml` isn't installed

mibac138:
Thanks, fixed that.

mibac138:
@vlad20012 Should be fixed. Also, is the Travis failure spurious? It timed out so I'm not sure what should I do about it (didn't find anything like a test failure or something to point me in some direction).

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
```diff
@@ -32,22 +34,38 @@ class RsGotoSuperHandler : LanguageCodeInsightActionHandler {
...
-        is RsFile -> mod.declaration
-        else -> mod.`super`
-    }
+    return null
+}
+
+fun Path.toCargoTomlPsiFile(project: Project): PsiFile? {
+    val virtualFile = LocalFileSystem.getInstance().findFileByPath("$this") ?: return null
+    val psiFile = PsiManager.getInstance(project).findFile(virtualFile) ?: return null
+    if (psiFile.fileType != TomlFileType) return null
```
vlad20012: Looks like this will crash the plugin if no `toml` plugin installed

vlad20012:
> Also, is the Travis failure spurious?

Looks like. You can re-check it by pushing a something

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
mibac138: Oops, missed that one. Fixed now

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandler.kt
```diff
@@ -32,22 +32,37 @@ class RsGotoSuperHandler : LanguageCodeInsightActionHandler {
...
-        is RsFile -> mod.declaration
-        else -> mod.`super`
-    }
+    return null
+}
+
+fun Path.toCargoTomlPsiFile(project: Project): PsiFile? {
+    val virtualFile = LocalFileSystem.getInstance().findFileByPath("$this") ?: return null
+    val psiFile = PsiManager.getInstance(project).findFile(virtualFile) ?: return null
+    if (psiFile.fileType.name != "TOML") return null
```
vlad20012: // maybe this check isn't needed

src/test/kotlin/org/rust/ide/navigation/goto/RsGotoSuperHandlerTest.kt
```diff
@@ -139,6 +139,19 @@ class RsGotoSuperHandlerTest : RsTestBase() {
             mod bar;
     """)
 
+    // Tried to mock Cargo this way but it didn't work (it does work when running IntellJ). Any tips?
```
vlad20012: Hmm, but this test passes on CI. Chat's wring here?

mibac138: @vlad20012 This test didn't pass - it just wasn't run because of `-` at the beginning of the method name

vlad20012:
Sorry for a long response. I tried to figure out how to write tests for it, and it turned out to be very non-trivial. I found that I can't place tests to `RsGotoSuperHandlerTest.kt` because these tests are running without TOML plugin, so I made `CargoTomlGotoSuperHandlerTest` inside `toml` project. Then I ran into a few more problems, including the one solved in #4326... To be honest, I spend few days to write these 2 tests =D

I opened a new fixed PR #4327 as a replacement for this.

Issue #1453, opened 2017-07-08
Refactoring module name removes `.rs` extension


torkleyy:
Using latest plugin release, `0.1.0.2061`, IntelliJ IDEA `2017.1.5`.

So I renamed the module declaration in my `main.rs` file from `mod foo;` to `mod bar;`. This refactoring renamed `foo.rs` to `bar` (without extension).

matklad:
Files are now renamed correctly, but looks like directories (`foo/mod.rs`) don't work really great yet.

Kobzol:
@ortem I checked both of these situations:
1) `a/b.rs` -> renamed `a::b` to `a::c` -> `a/c.rs`
2) `a/mod.rs` -> renamed `a` to `b` -> `b/mod.rs`
and both of them seem to work in `master`.

So I think that this issue can be closed.

ortem:
@Kobzol OK, thank you!

PR #3817, opened 2019-05-09
REF: Improve renaming of mod statements

Kobzol:
Currently, there are at least two problems with renaming `mod` statements.

1) When renaming a mod that has no reference to a file (happens if there is an directory without `mod.rs` used in resolving a module), the directory does not get renamed.
File tree:
```
a/b.rs
main.rs
```
main.rs:
```rust
mod a {
    mod b;
}
```
Renaming `mod a` does not change the name of the `a` directory, therefore the code doesn't compile after the refactoring.

2) When renaming a mod that references a `mod.rs` file, the file itself gets renamed, instead of its containing directory (https://github.com/intellij-rust/intellij-rust/issues/1453#issuecomment-319659558).
File tree:
```
a/mod.rs
main.rs
```
main.rs:
```rust
mod a;
```
Renaming `mod a` to `b` changes `mod.rs` to `b.rs` instead of chaning the `a` directory to `b`.

Ideally in this PR I'd like to solve both problems.

The current solution should solve problem 1). If a module without a file reference gets renamed, a directory matching the name of the module will be attempted to be renamed too. In theory it's possible that someone just defined an artificial module in the file or uses a direct module file (`a.rs`) and there may an unrelated directory with the same name, but that sounds like a rare occasion to me.

Problem 2) is a bit more complex. If the `mod` statement has a reference to a `mod.rs` file, the refactoring will directly try to rename the file, without first attempting to rename the `RsModDeclItem`. This can be checked in the code (I provide a commented naive attempt), however I don't know how to distinguish this from the scenario when the user actually wants to rename the actual `mod.rs` file.

Could you provide me with a hint on how to deal with this? Basically the problem as I understand it is that `mod` statements auto-resolve to the `mod.rs` file during refactoring if they have a reference to it.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
@@ -10,11 +10,9 @@ import com.intellij.openapi.util.Pass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.search.SearchScope
 import com.intellij.refactoring.rename.RenamePsiElementProcessor
-import org.rust.lang.core.psi.RsLabel
-import org.rust.lang.core.psi.RsLabelDecl
-import org.rust.lang.core.psi.RsLifetime
-import org.rust.lang.core.psi.RsLifetimeParameter
+import org.rust.lang.core.psi.*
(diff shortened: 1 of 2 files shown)


Kobzol:
Closing as both of the issues are already fixed.

PR #3820, opened 2019-05-09
ANN: Detect E0497: Stable attribute used outside of stdlib

mibac138:
Depends on #3806, related to #886.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -11,6 +11,7 @@ import com.intellij.lang.annotation.AnnotationSession
 import com.intellij.openapi.util.Key
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import org.rust.cargo.project.model.doesProjectLookLikeRustc
 import org.rust.cargo.project.workspace.CargoWorkspace.Edition
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.annotator.fixes.AddModuleFileFix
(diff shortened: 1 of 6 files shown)


vlad20012:
Why is it depends on #3806? And I see you made some changes in my commit... you can just review that PR

mibac138:
@vlad20012 It depends on #3806 to avoid false positives when working on `rust-lang/rust`. 

> you can just review that PR

I'm not sure what you mean by that?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -451,6 +452,14 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
 
     private fun checkAttr(holder: AnnotationHolder, attr: RsAttr) {
         checkImplBothCopyAndDrop(holder, attr)
+        checkAttrStableUsed(holder, attr)
+    }
+
+    // E0497: Stable attribute can only be used in the standard library
+    private fun checkAttrStableUsed(holder: AnnotationHolder, attr: RsAttr) {
+        if (attr.metaItem.name == "stable" && attr.cargoProject?.doesLookLikeRustLangRust() != true) {
```
mchernyavsky: I think use can just look at the origin of the `attr`'s package:
```kotlin
if (attr.metaItem.name == "stable" && attr.containingCargoPackage?.origin != PackageOrigin.STDLIB) {
```
So you can drop https://github.com/intellij-rust/intellij-rust/pull/3806 dependency. @vlad20012 Am I right?

Undin: @mchernyavsky Looks like if a project is rustc itself, it will produce false positive error annotation because in this case, `stdlib` crates are part of the project workspace

mibac138:
@vlad20012 Should I change anything in this PR? I'm not sure what's the status here

vlad20012:
I don't think that we need this kind of analysis. It is unlikely that the user will make a mistake like using `stable` attribute. Or they will make this mistake only once in a lifetime :)

Currently, this PR gives false-positives on a usual stdlib... Also in `attr.cargoProject?.doesProjectLookLikeRustc() != true` a `attr.cargoProject` can legitimately be null, so so this will give false-positives in this case as well.

PR #3821, opened 2019-05-09
RUN: rename expected/actual to left/right in comparison tests failure message

### LINKED ISSUE #3814: assert_eq! diff can lead to confusion
First of all, thanks for creating this plugin! It makes working with Rust a breeze.

The PR #3465 introduced nice error messages for showing the result of unit tests. The problem is that it assumes that the left argument as "expected" and the right one as "actual" (in my opinion correctly, in accordance to the most popular convention).

Unfortunately, not only such convention is not mentioned in the Rust docs, but the same rust docs consistently reverse the arguments 🤦‍♂ 

In order to avoid confusion, I think it would make sense to change "expected" and "actual" to "left" and "right". Highlighting the assert_eq! call site next to the message with the value of the arguments should also help users in their heroic pursuit to determine what is what.

vlad20012:
Fixes #3814

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
@@ -242,8 +242,8 @@ class CargoTestEventsConverter(
                 .addAttribute("details", failedMessage)
             val (left, right) = extractDiffResult(failedMessage) ?: return builder
             return builder
-                .addAttribute("expected", left)
-                .addAttribute("actual", right)
+                .addAttribute("left", left)
+                .addAttribute("right", right)
         }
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
```diff
@@ -242,8 +242,8 @@ class CargoTestEventsConverter(
                 .addAttribute("details", failedMessage)
             val (left, right) = extractDiffResult(failedMessage) ?: return builder
             return builder
-                .addAttribute("expected", left)
-                .addAttribute("actual", right)
+                .addAttribute("left", left)
+                .addAttribute("right", right)
```
Undin: These constants are not related to UI labels. They are names of concrete fields of a test event
So we should use `expected` and `actual` here

PR #3825, opened 2019-05-09
ANN: Detect E0118: Can impl only `struct`s, `enum`s and `union`s

mibac138:
Related #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -332,6 +332,7 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
     }
 
     private fun checkImpl(holder: AnnotationHolder, impl: RsImplItem) {
+        checkImplForNonAdtError(holder, impl)
         val traitRef = impl.traitRef ?: return
         val trait = traitRef.resolveToTrait() ?: return
         checkImplDropForNonAdtError(holder, impl, traitRef, trait)
@@ -358,6 +359,21 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
                 RsDiagnostic.TraitMissingUnsafeImplAttributeError(traitRef, attrRequiringUnsafeImpl).addToHolder(holder)
         }
     }
+
+    // E0118: Can impl only `struct`s, `enum`s and `union`s (when not implementing a trait)
+    private fun checkImplForNonAdtError(holder: AnnotationHolder, impl: RsImplItem) {
+        if (impl.`for` != null) return
+        val typeRef = impl.typeReference ?: return
+        if (typeRef.traitType != null) return
+        val type = typeRef.type
+        if (type is TyPrimitive && impl.queryAttributes.langAttribute == type.name) {
+            return // `#[lang = "u8"] impl u8 {}` is a valid impl
+        }
+        if (type !is TyAdt && type !is TyTraitObject && type != TyUnknown) {
+            RsDiagnostic.ImplForNonAdtError(typeRef).addToHolder(holder)
+        }
+    }
+
 
     // E0120: Drop can be only implemented by structs and enums
     private fun checkImplDropForNonAdtError(holder: AnnotationHolder, impl: RsImplItem, traitRef: RsTraitRef, trait: RsTraitItem) {
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -879,6 +879,16 @@ sealed class RsDiagnostic(
         }
     }
 
+    class ImplForNonAdtError(
+        element: PsiElement
+    ) : RsDiagnostic(element) {
+        override fun prepare() = PreparedAnnotation(
+            ERROR,
+            E0118,
+            "Can impl only `struct`s, `enum`s, `union`s and trait objects"
+        )
+    }
+
     class CannotAssignToImmutable(
         element: PsiElement,
         private val message: String,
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -864,6 +864,16 @@ sealed class RsDiagnostic(
         }
     }
 
+    class ImplForNonAdtError(
+        element: PsiElement
+    ) : RsDiagnostic(element) {
+        override fun prepare() = PreparedAnnotation(
+            ERROR,
+            E0118,
+            "Can impl only `struct`s, `enum`s and `union`s"
```
vlad20012: The message differs from the compiler one: `error[E0118]: no base type found for inherent implementation`

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -354,6 +355,18 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    // E0118: Can impl only `struct`s, `enum`s and `union`s (when not implementing a trait)
+    private fun checkImplForNonAdtError(holder: AnnotationHolder, impl: RsImplItem) {
+        if (impl.`for` != null) return
+        val typeRef = impl.typeReference ?: return
+        if (typeRef.traitType != null) return
+        val type = typeRef.type
+        if (type !is TyAdt && type != TyUnknown) {
```
vlad20012: There is an exception for primitive types: it's possible to make a single impl for primitive type with corresponding lang item, e.g. `#[lang = "u8"] impl u8 {}`. Such impls exist in `libcore`, but it's possible to make them in project in `#![no_core]` mode.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
mibac138: I thought this message might be somewhat simpler to understand than the compiler's one and seeing how other error messages don't follow compiler error messages that closely I thought it'd be okay. I can change it to the compiler's message though no problem.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
mibac138: Ooh, that's a good catch! I didn't really know about it. Thanks! :)

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
vlad20012: Okaay, sounds reasonable

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
vlad20012: I just found that it's also legal to implement trait objects:
```rust
trait Bar {}
impl dyn Bar {
    fn foo(&self) {}
}
```

mibac138: I believe that's already covered and won't report an error due to this bit of code: `if (typeRef.traitType != null) return`. I will double check just to be sure though

vlad20012: `typeRef.traitType != null` is about
```rust
impl Foo for Bar {}
    //^ this trait
```
but not for the case of trait object impl

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -864,6 +866,23 @@ sealed class RsDiagnostic(
...
 
+    class ImplForNonAdtError(
+        element: PsiElement,
+        private val impl: RsImplItem,
+        private val type: Ty
+    ) : RsDiagnostic(element) {
+        override fun prepare() = PreparedAnnotation(
+            ERROR,
+            E0118,
+            "Can impl only `struct`s, `enum`s and `union`s (and primitives in `no_core` crates)",
```
vlad20012: I don't think it's reasonable to mention primitives here, it isn't useful for must users. Mention trait objects looks much more reasonable

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -864,6 +866,23 @@ sealed class RsDiagnostic(
...
+            fixes = if (type is TyPrimitive && NO_CORE.availability(element) == FeatureAvailability.CAN_BE_ADDED) {
+                listOf(NO_CORE.addFeatureFix(element))
```
vlad20012: No-core is not only feature, but crate attribute `#[no_core]`, and I also think that this fix isn't reasonable

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -354,6 +355,21 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    // E0118: Can impl only `struct`s, `enum`s and `union`s (when not implementing a trait)
+    private fun checkImplForNonAdtError(holder: AnnotationHolder, impl: RsImplItem) {
+        if (impl.`for` != null) return
+        val typeRef = impl.typeReference ?: return
+        if (typeRef.traitType != null) return
+        val type = typeRef.type
+        if (type is TyPrimitive && impl.queryAttributes.langAttribute == type.name) {
```
vlad20012: :+1:

vlad20012:
Please, rebase it to the master

mibac138:
@vlad20012 done

mchernyavsky:
Thanks!
bors r+

Issue #1675, opened 2017-08-30
Code analysis not working when file is not referenced


Alex-PK:
When editing a file in a project, if the file is not referenced from main (or lib), syntax highlighting works, but code analysis does not.

For example, having

src/mymod/mod.rs:
```
mod one;
// missing "mod two;"
```

src/mymod/two.rs
```
use std::fs;

fn foo() -> Result<String, ()> {
  fs::read_dir(...).{cursor_here}
}
```

When you are in {cursor_here} and press ctrl-space, nothing gets suggested as auto-complete; Pressing ctrl-b over `read_dir` returns "declaration not found"; Type hints are not shown for function results; No error is detected for the wrong `Result<>` signature, etc.

The desired effect would be to at least have code analysis for crates (both std and the ones in Cargo.toml), as I understand it's hard to have info about the rest of the application, not knowing where the file is going to end up in the nemaspace.

But also, as a quick fix, having a yellow bar on the top saying "Code analysis is disabled for this file as it's not linked in the project" (or something similar) would be very useful, as it would also remind the developer to add it. `cargo check` ignores the file, and I was puzzled for a while yesterday when it happened: I had no errors or warning from either side. :)

matklad:
That sort of expected, because `use std::fs;` says "go to the crate root and fetch `std` name from there", but if the file is unlinked to the rest of the project, then we can't find a crate root!

>But also, as a quick fix, having a yellow bar on the top saying "Code analysis is disabled for this file as it's not linked in the project" (or something similar) would be very useful, as it would also remind the developer to add it. cargo check ignores the file, and I was puzzled for a while yesterday when it happened: I had no errors or warning from either side. :)

Seems like a useful thing to add!

Kobzol:
@matklad This yellow bar stating that the current file is not analysed is shown in CLion (if there's no CMakeLists.txt entry for that file). CLion is not open source though - could you provide a hint how to add something like this in the Rust plugin?

vlad20012:
@Kobzol We have the same, but for some reason it isn't showed if there is linked `Cargo.toml` in some ancestor directory.

![image](https://user-images.githubusercontent.com/3221931/57465744-eafb4d00-7287-11e9-9068-e10637c90d87.png)

Kobzol:
Oh, you're right, I could swear that I saw that bar before :) I'll try to take a look why it disappears when there's `Cargo.toml`.

Kobzol:
[...]

Issue #3426, opened 2019-02-21
Automatically add module declaration


sirgl:
It would be good to automatically add declaration of module to corresponding location during file creation.
Example:
```
foo.rs  <-- create file
...
lib.rs  <-- add mod foo here
```

vlad20012:
Currently, it works in a different order. When you type `mod foo;` in some file, the plugin shows you suggestion to create a file with such name:
![peek 2019-02-21 12-45](https://user-images.githubusercontent.com/3221931/53159255-8ad50300-35d6-11e9-9cc5-9304d2bdbdaf.gif)

sirgl:
I know, but it would be more convenient to have such feature.
Thw other solution is to add action to create module or make chooser with options to create empty file or module

vlad20012:
I think we can consider it as closed by #3826

PR #3826, opened 2019-05-10
TYPE: display notification panel when a file is not included in the module tree

Kobzol:
This PR adds a notification panel that warns about files which are not included in any module tree.
Currently the file has to be re-opened for the notification bar to update.

Corresponding issue: https://github.com/intellij-rust/intellij-rust/issues/1675

### DIFF
--- a/src/main/kotlin/org/rust/ide/notifications/MissingToolchainNotificationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/notifications/MissingToolchainNotificationProvider.kt
@@ -13,6 +13,8 @@ import com.intellij.openapi.fileEditor.FileEditor
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Key
 import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
 import com.intellij.ui.EditorNotificationPanel
 import com.intellij.ui.EditorNotifications
 import org.rust.cargo.project.model.CargoProject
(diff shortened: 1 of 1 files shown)


vlad20012:
> Currently the file has to be re-opened for the notification bar to update.

Just add this code to `init` block:
```kotlin
subscribe(RUST_STRUCTURE_CHANGE_TOPIC, object : RustStructureChangeListener {
                override fun rustStructureChanged(file: PsiFile?, changedElement: PsiElement?) {
                    notifications.updateAllNotifications()
                }
            })
```

vlad20012:
Btw #3426 can be solved by some action (see `createActionLabel`)

Kobzol:
> Just add this code to `init` block:

Thank you, it works nicely :)

Considering https://github.com/intellij-rust/intellij-rust/issues/3426, what should the action do? Let's say there's a file `a.rs` in directory `b`. If I wanted to include this file in the tree, in theory it could be included either to `b/mod.rs`, `b.rs`, or `lib.rs`/`main.rs`/whatever the root file is called. So I need to get from the containing directory of the said file to a module root file. Is there an API for that? I tried using `toPsiDirectory` for the parent of the file, but I don't know how to get from a directory to anything Rust related (e.g. `RsElement`).

vlad20012:
> what should the action do? Let's say there's a file a.rs in directory b. If I wanted to include this file in the tree, in theory it could be included either to b/mod.rs, b.rs, or lib.rs/main.rs/whatever the root file is called.

... or anywhere else via `#[path]` attribute or `include!()` macro, or it shouldn't included anywhere at all and it have '.rs' by accident, or it is code sample/test (like in IntelliJ Rust tests) etc etc etc

Yes, this is not trivial =) And I can't suggest anything for now.

I think this PR is finished. Thanks!
bors r+

Kobzol:
Speaking of `include!`, this will probably show false positives for files included with it, at least until https://github.com/intellij-rust/intellij-rust/issues/1908 and similar are resolved, but there's probably not much that can be done about that. Thank you.

vlad20012:
I don't think we should call it "false positive". This bar says that "analysis is not available", and it actually is not available :)

Kobzol:
Indeed, the false positive part is actually `File is not included in module tree` :)

PR #3828, opened 2019-05-10
 RUN: fix backtrace hyperlinks on 2018 edition

vlad20012:
Fixes such hyperlinks in 2018 edition crates or crates containing "-" in the name
image

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -27,7 +27,7 @@ interface CargoWorkspace {
     val workspaceRootPath: Path?
 
     val packages: Collection<Package>
-    fun findPackage(name: String): Package? = packages.find { it.name == name }
+    fun findPackage(name: String): Package? = packages.find { it.name == name || it.normName == name }
 
     fun findTargetByCrateRoot(root: VirtualFile): Target?
     fun isCrateRoot(root: VirtualFile) = findTargetByCrateRoot(root) != null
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -28,6 +28,7 @@ interface CargoWorkspace {
 
     val packages: Collection<Package>
     fun findPackage(name: String): Package? = packages.find { it.name == name }
+    fun findPackageByNormName(normName: String): Package? = packages.find { it.normName == normName }
```
Undin: Maybe it's better to upgrade `findPackage` method (i.e. search package by name and norm name at the same time) instead of new rather specific method in API?

src/test/kotlin/org/rust/RsTestBase.kt
```diff
@@ -99,6 +100,24 @@ abstract class RsTestBase : LightPlatformCodeInsightFixtureTestCase(), RsTestCas
             System.err.println("SKIP $name: $reason")
             return
         }
+
+        if (findAnnotationInstance<BothEditions>() != null &&
+            javaClass.getMethod(name).getAnnotation(MockEdition::class.java) == null) {
```
Undin: What if I mark the whole test class with `MockEdition` annotation?
Maybe it's better to always ignore one of these annotations when we find both ones?

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
vlad20012: :thinking: :thinking: :thinking:

src/test/kotlin/org/rust/RsTestBase.kt
vlad20012: There is a case in RsStdResolveTest where they are mixed. But :thinking: really we can just extract it to another file (like RsStdlibResolveTestEdition2015). And in this case, it looks better to forbid to implement both `BothEditions` and `MockEdition` (i.e. throw exception in this case).

Undin: About `RsStdlibResolveTest`: in #3864 I drop code to launch tests with both editions to have exactly the same environment as the current stdlib

vlad20012:
Fixed

PR #3829, opened 2019-05-10
INT: add support for generic types to unelide lifetimes intention

Kobzol:
This PR generalizes the Unelide lifetimes intention to work with generic types.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/769

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/UnElideLifetimesIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/UnElideLifetimesIntention.kt
@@ -10,6 +10,11 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.types.infer.hasReEarlyBounds
+import org.rust.lang.core.types.regions.Region
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.type
 
 class UnElideLifetimesIntention : RsElementBaseIntentionAction<RsFunction>() {
     override fun getText() = "Un-elide lifetimes"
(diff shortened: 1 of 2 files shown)


Kobzol:
What about nested liftimes?

For example with this fn:
`fn make_s<'a>(x: &'a S<'a>) { }`
elision is not offered by the plugin, does that mean that for
`fn make_s(x: &S) { }`
it should not offer to unelide?
What should be the general rule, that there is only one lifetime and no nested ones?

Kobzol:
I pushed an initial version of unelision which tries to be more general. Requesting feedback :-)

This now works:
```rust
struct S<'a, T, X> { x: &'a T, y: X }
fn make_s<X>(x: S<u32, X>) -> S<u32, X> { unimplemented!() }
fn make_s<'a, X>(x: S<'a, u32, X>) -> S<'a, u32, X> { unimplemented!() }
```

Howver, it is not recursive yet, i.e. it does not handle nested types.
Should this also work or should we stop trying at some point?
```rust
trait S<'a, T>
{
    type N;
    fn foo(&self, a: &'a u32, b: T, c: Self::N);
}
fn make_s(x: Box<dyn S<u32, N=u32>>) -> Box<dyn S<u32, N=u32>> { unimplemented!() }
fn make_s<'a>(x: Box<dyn S<'a, u32, N=u32>>) -> Box<dyn S<'a, u32, N=u32>> { unimplemented!() }
```

Kobzol:
I updated the PR and rebased it onto master. I will handle impl block unelision in a different PR. The intention is still not recursive, as mentioned in the previous comments, but I'm not sure if that's a good idea. What do you think?

Kobzol:
@mchernyavsky ping

Kobzol:
Rebased to fix merge conflict.

mchernyavsky (commented):
> I updated the PR and rebased it onto master. I will handle impl block unelision in a different PR. The intention is still not recursive, as mentioned in the previous comments, but I'm not sure if that's a good idea. What do you think?

I agree, but what about ADTs with multiple lifetimes? I think we shouldn't close https://github.com/intellij-rust/intellij-rust/issues/769 until this case is supported.

Kobzol:
I added support for multiple lifetime arguments for structs in parameters. I'm not sure what should be the semantics for the output type though.
```rust
struct S<'a, 'b> { x: &'a u32, y: &'b u32 }
fn make_s<'a, 'b>(x:/*caret*/ &'a i32) -> S<'a, 'b> { unimplemented!() }
```
I'm not even sure if this qualifies as (un)elision.

mchernyavsky:
> I added support for multiple lifetime arguments for structs in parameters. I'm not sure what should be the semantics for the output type though.

I think the intention shouldn't be available in such cases. If I'm not mistaken, in all other cases the result can be reverted with the `Elide lifetimes` fix.

Kobzol:
> I think the intention shouldn't be available in such cases. If I'm not mistaken, in all other cases the result can be reverted with the `Elide lifetimes` fix.

Good point. To mirror `RsNeedlessLifetimesInspection`, I now do not offer the intention if the output has multiple lifetimes.

mchernyavsky (approved):
Thanks!

bors r+

PR #3831, opened 2019-05-11
PERF: significantly optimize "find usages" (in the broad sense)

vlad20012:
1. First, by tuning `Rs*ReferenceImpl.isReferenceTo` we say: "this reference is definitely can't be resolve to this element" without resolving the reference. This optimizes find usages of any elements because now we should resolve less references.
2. Optimize "find usages" of type/lifetime parameters by using local scopes. I'd say it was a bug that local scope wasn't used before because of obvious local nature of such parameters.
3. Take into account item visibility. This significantly speeds up "find usages" of private items (**UPD**: only inside modules without child modules for now) and items declared in local scope (i.e. in function bodies, etc)

I don't know how to write tests for it, and I tested performance rather than correctness, so it should be reviewed intently.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiImplUtil.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiImplUtil.kt
@@ -5,9 +5,9 @@
 
 package org.rust.lang.core.psi
 
-import org.rust.lang.core.psi.ext.RsMod
-import org.rust.lang.core.psi.ext.RsNamedElement
-import org.rust.lang.core.psi.ext.superMods
+import com.intellij.psi.search.LocalSearchScope
+import com.intellij.psi.search.SearchScope
+import org.rust.lang.core.psi.ext.*
 
 /**
  * Mixin methods to implement PSI interfaces without copy pasting and
@@ -27,6 +27,17 @@ object RsPsiImplUtil {
         }
         if (segments.isEmpty()) return ""
         return "::" + segments.joinToString("::")
+    }
+
+    /**
+     * Used by [RsTypeParameter] and [RsLifetimeParameter].
+     * @return null if default scope should be used
+     */
+    fun getParameterUseScope(element: RsElement): SearchScope? {
+        val owner = element.contextStrict<RsGenericDeclaration>()
+        if (owner != null) return LocalSearchScope(owner)
+
+        return null
     }
 }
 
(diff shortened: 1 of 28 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsPsiImplUtil.kt
```diff
@@ -28,5 +30,86 @@ object RsPsiImplUtil {
...
+
+        if (!restrictedMod.hasChildModules()) return LocalSearchScope(containingMod)
+
+        val virtualFile = restrictedMod.contextualFile.virtualFile ?: return null
+
+        // Use parent directory (with subdirectories) as a scope. We assume that all child modules are located in
+        // the same directory. The assumption is generally wrong: a file from any directory can be included to
+        // the module structure via `#[path]` attribute or `include!()` macro. But it works in the most cases.
+        // Maybe we should calculate more precise scope by traversing module tree.
+        return GlobalSearchScopesCore.DirectoryScope(restrictedMod.project, virtualFile.parent, true)
```
Undin: We don't support `include` macro at this moment so it's ok
But `#[path]` attribute is not so rare so we can break some code that works now. How much does it affect `find usages` performance if we use default search scope when `restrictedMod` has child modules?

Looks like there are three main cases:
1. item is public and we should provide default scope (it's already supported)
2. item is private and containing mod doesn't have child modules
3. item is crate private

`1` will not be changed
`2` will not be changed
`3` not sure how much it affects performance

vlad20012: There is also 4th item: item is private and containing mod have child modules. This is enough usual case, I guess, and we optimize it significantly.

About 3th item, it it reasonable when searching usages of some item inside a dependency crate, or in project with mane crates, like `rustc`.

> But #[path] attribute is not so rare so we can break some code that works now.

`#[path]` attribute - yes, but `#[path]` attribute that refer to a file from parent directory - I'm not sure.

vlad20012: Oh, I found a lot of usages of `path` attribute like `#[path = ".../something"]` in stdlib! Ok, looks like it's not a rare case. For now I've replaced it with `return null` (use default scope). I future I'll try to traverse module structure to calculate more precise scope

PR #3832, opened 2019-05-11
ANN: Detect E0132: Invalid `#[start]` function

mibac138:
Related: #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -490,6 +490,43 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         checkImplBothCopyAndDrop(holder, attr)
         checkInlineAttr(holder, attr)
         checkReprForEmptyEnum(holder, attr)
+        checkStartAttribute(holder, attr)
+    }
+
+    // E0132: Invalid `start` attribute
+    private fun checkStartAttribute(holder: AnnotationHolder, attr: RsAttr) {
+        if (attr.metaItem.name != "start") return
+
+        START.check(holder, attr.metaItem, "#[start] function")
+
+        when (val owner = attr.owner) {
+            is RsFunction -> {
+                // Check if signature matches `fn(isize, *const *const u8) -> isize`
+                val params = owner.valueParameters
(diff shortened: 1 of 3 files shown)


mibac138:
I must admit I'm rather quizzed about the failing tests. I ran them locally and discovered that when ran one by one they do pass, but when running all tests of `RsErrorAnnotator` they fail. I'm not sure what should I do?

src/test/kotlin/org/rust/ide/annotator/RsStdlibErrorAnnotatorTest.kt
```diff
@@ -24,4 +24,9 @@ class RsStdlibErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class.j
 
         <error descr="Can't find crate for `litarvan` [E0463]">extern crate litarvan;</error>
     """)
+
+    fun `test missing feature E0132`() = checkErrors("""
```
mchernyavsky: Hmm, why is this test not in `RsErrorAnnotatorTest.kt`? `RsErrorAnnotatorTest` already has a few tests for E0658.

mibac138: Without stdlib `START.availbility` returns unknown. That means that without stdlib the tests would pass too. I could've left `#[feature]` out in the non stdlib tests but I added them just in case

mchernyavsky: > Without stdlib START.availbility returns unknown.

Just add `@MockRustcVersion` annotation:
```kotlin
@MockRustcVersion("1.0.0")
fun `test start feature E0658 1`() = checkErrors("""
    #[<error descr="#[start] function is experimental [E0658]">start</error>]
    fn start(_argc: isize, _argv: *const *const u8) -> isize { 0 }
""")

@MockRustcVersion("1.0.0")
fun `test start feature E0658 2`() = checkErrors("""
    #![feature(start)]
    #[<error descr="#[start] function is experimental [E0658]">start</error>]
    fn start(_argc: isize, _argv: *const *const u8) -> isize { 0 }
""")

@MockRustcVersion("1.0.0-nightly")
fun `test start feature E0658 3`() = checkErrors("""
    #![feature(start)]
    #[start]
    fn start(_argc: isize, _argv: *const *const u8) -> isize { 0 }
""")
```

Issue #3844, opened 2019-05-15
inspection RsNeedlessLifetimes does not respect allow(clippy::needless_lifetimes)


CAD97:
## Environment

* **Intellij-Rust plugin version:** v0.2.97.2123-191
* **Rust toolchain version:**  
rustc 1.35.0-beta.7 (2bafaaf66 2019-05-14)  
cargo 1.35.0-beta (6f3e9c367 2019-04-04)

* **IDE name and version:**
IntelliJ IDEA 2019.1.2 (Ultimate Edition)  
Build #IU-191.7141.44, built on May 7, 2019  
JRE: 1.8.0_202-release-1483-b49 amd64  
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o  
Windows 10 10.0

## Steps to reproduce

```rust
trait T {
    #[allow(clippy::needless_lifetimes)]
    fn f<'a>(this: &'a ());
}
```

![image](https://user-images.githubusercontent.com/5992217/57805888-ecb38d80-772b-11e9-8798-ad7162692a16.png)

```rust
#[allow(clippy::needless_lifetimes)]
fn f<'a>(this: &'a ()) {}
```

![image](https://user-images.githubusercontent.com/5992217/57805962-1e2c5900-772c-11e9-813a-e172611b8571.png)

Strangely enough, I am able to suppress the warning for trait methods with `//noinspection RsNeedlessLifetimes`, but not for free fn.

PR #3855, opened 2019-05-18
INSP: Provide suppress quick fixes for needless_lifetimes inspection

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/3844.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsLint.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsLint.kt
@@ -13,14 +13,15 @@ import org.rust.lang.core.psi.ext.*
  */
 enum class RsLint(
     val id: String,
-    val groupId: String? = null,
+    val groupIds: List<String> = emptyList(),
     val defaultLevel: RsLintLevel = RsLintLevel.WARN
 ) {
-    NonSnakeCase("non_snake_case", "bad_style"),
-    NonCamelCaseTypes("non_camel_case_types", "bad_style"),
-    NonUpperCaseGlobals("non_upper_case_globals", "bad_style"),
+    NonSnakeCase("non_snake_case", listOf("bad_style")),
+    NonCamelCaseTypes("non_camel_case_types", listOf("bad_style")),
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/inspections/RsLint.kt
```diff
@@ -30,7 +31,8 @@ enum class RsLint(
     private fun explicitLevel(el: PsiElement): RsLintLevel? = el.ancestors
         .filterIsInstance<RsDocAndAttributeOwner>()
         .flatMap { it.queryAttributes.metaItems }
-        .filter { it.metaItemArgs?.metaItemList.orEmpty().any { it.name == id || it.name == groupId } }
+        .filter { it.metaItemArgs?.metaItemList.orEmpty()
+            .any { it.text == id || it.text in groupIds } }
```
Undin, line 35: Maybe it's better to use `path?.text` instead of whole text here?
And maybe it's worth to extract `metaItem.path?.text` into some extension property

mchernyavsky:
@Undin Fixed

> But I doubt if it's ok that we suggest user to suppress the corresponding inspection with external clippy::needless_lifetimes attribute?

Yes, why not?

> Maybe we should use them only to suppress our inspections but don't use in suppressing quick fixes?

How can we suppress an inspection, but don't suppress the corresponding quick fix?

Undin:
> Yes, why not?

Because `clippy` is an external tool for us and user may not use it at all (or even may not know about clippy). Ideally, we should have own attributes for custom inspections

> How can we suppress an inspection, but don't suppress the corresponding quick fix?

I mean to check external suppress attributes but not provide suppress quick fixes that insert these external attributes (i.e. don't provide them from `org.rust.ide.inspections.RsLintInspection#getBatchSuppressActions`). To achieve it, we need to mark the corresponding lint objects as external and slightly modify code in `org.rust.ide.inspections.RsLintInspection#getBatchSuppressActions`

mchernyavsky:
@Undin Tests are fixed now.

PR #3861, opened 2019-05-20
LI: disable doctest injection for not doctestable libs

vlad20012:
1. CLEAN: change TargetKind to follow [cargo internal structure](https://github.com/rust-lang/cargo/blob/5a0c31d81/src/cargo/core/manifest.rs#L157)
2. LI: disable doctest injection for not doctestable libs (fixes false-positives inside rust-lang/rust)


src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -159,6 +162,20 @@ object CargoMetadata {
             }
     }
 
+    enum class TargetKind {
```
vlad20012: Moved here from CargoWorkspace

PR #3865, opened 2019-05-20
TYPE: improve heuristic for quote handler

rrevenantt:
Current heuristic works very bad when code contains a lot of format strings, which often results to reanalysis of whole file and disabled code insight for dozens of seconds on big files. I could just have added `{` to the list of invalid tokens, but I think it is much easier to create full list of tokens that are valid after string. 

But it is still not very reliable solution, and since consequences of the unmatched quote can be quite annoying maybe it would be better to have more reliable implementation?
I can suggest couple of options:
- check couple of the next string literals for valid subsequent token, which should drastically decrease possibility of false positive result while still being quite fast.
- calculate number of quotes until the end of file. Although, I am not sure if it will be fast enough for smooth typing experience.

### DIFF
--- a/src/main/kotlin/org/rust/ide/typing/RsQuoteHandler.kt
+++ b/src/main/kotlin/org/rust/ide/typing/RsQuoteHandler.kt
@@ -57,8 +57,8 @@ class RsQuoteHandler : SimpleTokenSetQuoteHandler(
         if (super.isNonClosedLiteral(iterator, chars)) return true
         // Rust allows multiline literals, so an unclosed quote will
         // match with an opening quote of the next literal.
-        // Let's employ heuristics to find out if it is the case!
-        // Specifically, check if the next token can't appear
+        // Let's employ heuristics to find out if it is not the case!
+        // Specifically, check if the next token can appear
         // in valid Rust code.
         val nextChar = chars.getOrElse(iterator.end) { '"' }
         if (nextChar == '"') return true
@@ -68,9 +68,9 @@ class RsQuoteHandler : SimpleTokenSetQuoteHandler(
             iterator.advance()
         }
         when (iterator.tokenType) {
-            IDENTIFIER, INTEGER_LITERAL, FLOAT_LITERAL -> return true
+            COMMA, DOT, RBRACE, RBRACK, RPAREN, SEMICOLON -> return false
         }
-        return false
+        return true
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vlad20012:
Hi!
Previously I never touched quote handlers, so since in any case I would have to review this, I decided to sort it out a bit. I came to the conclusion that both solutions are far from ideal. They depend on string literal content, which can be whatever by definition. So much the worse that we have macros, so we have no guarantees about tokens that will (not) be after a string literal. So I decided to try something similar to your last option - "calculate the number of quotes until the end of file". There should not be performance problems because `HighlighterIterator` consists of an array of pre-lexed tokens of the entire file, so we just iterate an array. I implemented it in #3972

vlad20012:
@rrevenantt Could you please check that my solution works for your cases?

rrevenantt:
@vlad20012 Your solution is fine for me. But i am still not sure about performance.
I thought that when you type quote, file needs to be re-lexed because all literal tokens will be kind of "reversed" and pre-lexed array will not be relevant.

vlad20012:
> all literal tokens will be kind of "reversed"

Yes, but such "reversed" tokens are lexed __before__ the platform calls to `RsQuoteHandler.isNonClosedLiteral`, and all lexemes are stored in the array inside LexerEditorHighlighter (`HighlighterIteratorImpl`). When you call `HighlighterIterator.advance` you literally just increment the counter! Just look at this:
image
If you don't believe, you can just debug it:
image
`myData` is an array of tokes (in a compact form)

So this shouldn't affect performance anyhow

rrevenantt:
I agree that your changes by itself do not affect performance, but platform still re-lexes almost whole file two times(for each quote). So performance might have been affected from the beginning. If for example we could get `Editor` before first quote is typed(maybe from custom `TypedHandlerDelegate`), we would be able to get last token with something like `editor.highlighter.createIterator(editor.document.textLength).tokenType` in constant time. Thus if necessary we can insert two quotes together and platform will not need to do a full lexer pass. I am just not sure if performance benefit worth it.

vlad20012:
Looks like the entire document text is re-lexed without any incremental support every time you type a single symbol. And it's not an issue because lexing is extremely fast.

rrevenantt:
I think you are partially wrong here. `LexerEditorHighlighter.documentChanged()` does incremental re-lexing automatically on any document change event. Here is comment from it:
```
        Highlighting lexer is expected to periodically return to its "initial state" and
        so to denote valid starting points for incremental highlighting.

        If this requirement is unfulfiled, document has to be always re-analyzed from the beginning
        up to the point of modification,  which can hog CPU and make typing / editing very sluggish,
        especially at large offsets (with at least O(n) time complexity).
```
and its second part makes me think that there might be problems. But i have tried to notice any delay when typing quote on 200KB  file - with no success. So it looks like lexing is fast enough to not cause any delay and quote typing is not particularly often event to cause significant CPU load. Although I still think that my proposal might be a good optimization, looks like it is not necessary. So feel free to merge your implementation.

vlad20012:
Hmm, when I debugged it I didn't mentioned anything incremental... Maybe there are some issues with our lexer or a something. Anyway I don't think it's major issue.

Thank you for this PR and pointing out the problem in the quote handler. Then I merge my version.

PR #3866, opened 2019-05-20
TYPE:fix angle brace matching with macro call inside

rrevenantt:
See added test.


ortem:
Thanks!
bors r+

mchernyavsky:
bors retry

Issue #2256, opened 2018-02-04
Format macros support


Undin:
A lot of languages have possibility to create/print strings with some formatting. For example [Formatter](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html) class in java or [printf](http://www.cplusplus.com/reference/cstdio/printf/) function in C/C++.
Rust has [formatting macros](https://doc.rust-lang.org/std/fmt/) for it. 

It would be great to have the support of format macros in intellij-rust like in other languages, i.e.:
* [ ] Highlight special constructions: `{}`, `{:?}`, etc. Looks like that good solution is to create special parser for format strings according to this [syntax](https://doc.rust-lang.org/std/fmt/#syntax)
* [ ] Check number of parameters and their types and show warnings if format strings and arguments don't match #4089

Kobzol:
Fixed by https://github.com/intellij-rust/intellij-rust/pull/4969.

PR #3869, opened 2019-05-21
ANN:Format macros syntax support

rrevenantt:
Closes first part of #2256

Do i need to add more tests?

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
@@ -1,0 +1,134 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator
+
+import com.intellij.ide.annotator.AnnotatorBase
+import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.util.TextRange
+import com.intellij.openapiext.isUnitTestMode
+import com.intellij.psi.PsiElement
+import org.intellij.lang.annotations.Language
+import org.rust.ide.colors.RsColor
+import org.rust.lang.core.psi.RsLitExpr
+import org.rust.lang.core.psi.RsLiteralKind
+import org.rust.lang.core.psi.RsMacroCall
+import org.rust.lang.core.psi.ext.descendantOfTypeStrict
+import org.rust.lang.core.psi.ext.macroName
+import org.rust.lang.core.psi.ext.startOffset
+import org.rust.lang.core.psi.kind
+import org.rust.lang.utils.parseRustStringCharacters
+
+class RsFormatMacroAnnotator : AnnotatorBase() {
+
+    override fun annotateInternal(element: PsiElement, holder: AnnotationHolder) {
+        val formatMacro = element as? RsMacroCall ?: return
+
+        val formatStr = formatMacro
+            .formatMacroArgument
+            ?.formatMacroArgList
+            ?.getOrNull(macroToFormatPos(formatMacro.macroName))
+            ?.descendantOfTypeStrict<RsLitExpr>()
+            ?: return
+
+        val literalKind = (formatStr.kind as? RsLiteralKind.String) ?: return
+        val rawTextRange = literalKind.offsets.value ?: return
+        val (unescapedText, sourceMap, _) = parseRustStringCharacters(rawTextRange.substring(literalKind.node.text))
+
+        val arguments = formatParser.findAll(unescapedText)
+        var currentOffset = 0
+
+        fun IntRange.toSourceRange(additionalOffset: Int = currentOffset): TextRange =
+            TextRange(sourceMap[first + additionalOffset], sourceMap[last + additionalOffset] + 1)
+                .shiftRight(formatStr.startOffset + rawTextRange.startOffset)
+
+        fun AnnotationHolder.highlightArgument(range: IntRange?, color: RsColor = RsColor.IDENTIFIER) {
+            if (range != null && !range.isEmpty()) {
+                createAnnotation(
+                    HighlightSeverity.INFORMATION,
+                    range.toSourceRange(),
+                    null
+                ).textAttributes = color.textAttributesKey
+            }
+        }
+
+        val key = RsColor.FORMAT_SPECIFIER
+        val highlightSeverity = if (isUnitTestMode) key.testSeverity else HighlightSeverity.INFORMATION
+        for (arg in arguments) {
+            currentOffset = 0
+            val textRange = arg.range.toSourceRange()
+            holder.createAnnotation(highlightSeverity, textRange, null)
+                .textAttributes = key.textAttributesKey
+
+            if (arg.groups[3] != null) {
+                holder.createErrorAnnotation(
+                    arg.range.toSourceRange(),
+                    "Invalid format string: unmatched '}'"
+                )
+            }
+            if (arg.groups[1] != null) {
+
+                val inside = arg.groups[2] ?: error(" should not be null because can match empty string")
+                currentOffset = inside.range.first
+                val matchResult = formatArgParser.find(inside.value)
+                    ?: error(" should be not null because can match empty string")
+
+                if (!arg.groups[1]!!.value.endsWith("}")) {
+                    val possibleEnd = matchResult.value.length - 1
+                    holder.createErrorAnnotation(
+                        (possibleEnd..possibleEnd).toSourceRange(),
+                        "Invalid format string: } expected.\nIf you intended to print `{` symbol, you can escape it using `{{`"
+                    )
+                    continue
+                }
+
+                val validParsedEnd = matchResult.range.last + 1
+                if (validParsedEnd != inside.value.length) {
+                    holder.createErrorAnnotation(
+                        (validParsedEnd until inside.value.length).toSourceRange(),
+                        "Invalid format string"
+                    )
+                } else {
+                    holder.highlightArgument(matchResult.groups["id"]?.range)
+                    holder.highlightArgument(matchResult.groups["width"]?.range)
+                    holder.highlightArgument(matchResult.groups["precision"]?.range)
+                    holder.highlightArgument(matchResult.groups["type"]?.range, RsColor.FUNCTION)
+                    //TODO: argument type/count/other checks
+                }
+            }
+        }
+    }
+
+    private fun macroToFormatPos(macro: String): Int = when (macro) {
+        "println" -> 0
+        "print" -> 0
+        "eprintln" -> 0
+        "eprint" -> 0
+        "format" -> 0
+        "format_args" -> 0
+        "format_args_nl" -> 0
+        "write" -> 1
+        "writeln" -> 1
+        else -> -1
+    }
+
+    companion object {
+        private val formatParser = Regex("""\{\{|}}|(\{([^}]*)}?)|(})""")
+
+        @Language("Regexp")
+        private const val argument = """([a-zA-Z_][\w+]*|\d+)"""
+        private val formatArgParser = Regex("""(?x) # enable comments
+^(?<id>$argument)?
+(:
+    (.?[\^<>])?[+\-]?\#?
+    0?(?!\$) # negative lookahead to parse 0$ as width and 00$ as zero padding followed by width
+    (?<width>$argument\$|\d+)?
+    (\.(?<precision>$argument\$|\d+|\*))?
+    (?<type>\w?\??)?
+)?\s*""")
+
+    }
+}
--- a/src/main/kotlin/org/rust/ide/colors/RsColors.kt
+++ b/src/main/kotlin/org/rust/ide/colors/RsColors.kt
@@ -52,6 +52,7 @@ enum class RsColor(humanName: String, default: TextAttributesKey? = null) {
     STRING("Literals//Strings//String", Default.STRING),
     VALID_STRING_ESCAPE("Literals//Strings//Escape sequence//Valid", Default.VALID_STRING_ESCAPE),
     INVALID_STRING_ESCAPE("Literals//Strings//Escape sequence//Invalid", Default.INVALID_STRING_ESCAPE),
+    FORMAT_SPECIFIER("Literals//Strings//Format specifier in string argument", Default.VALID_STRING_ESCAPE),
 
     BLOCK_COMMENT("Comments//Block comment", Default.BLOCK_COMMENT),
     EOL_COMMENT("Comments//Line comment", Default.LINE_COMMENT),
--- a/src/main/resources/META-INF/rust-core.xml
+++ b/src/main/resources/META-INF/rust-core.xml
@@ -169,6 +169,7 @@
         <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsSyntaxErrorsAnnotator"/>
         <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsEdition2018KeywordsAnnotator"/>
         <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsLiteralAnnotator"/>
+        <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsFormatMacroAnnotator"/>
         <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsHighlightingAnnotator"/>
         <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsHighlightingMutableAnnotator"/>
         <annotator language="Rust" implementationClass="org.rust.ide.annotator.RsExpressionAnnotator"/>
--- a/src/test/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotatorTest.kt
@@ -1,0 +1,165 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator
+
+import org.rust.ide.colors.RsColor
+
+class RsFormatMacroAnnotatorTest : RsAnnotatorTestBase(RsFormatMacroAnnotator::class) {
+    override fun setUp() {
+        super.setUp()
+        super.annotationFixture.registerSeverities(RsColor.values().map(RsColor::testSeverity))
+    }
+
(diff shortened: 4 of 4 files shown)


ice1000:
May I ask for the status of this PR?

rrevenantt:
@ice1000 it is waiting for review

src/main/kotlin/org/rust/lang/format/RsFormatMacroInjector.kt
```diff
@@ -0,0 +1,44 @@
...
+                    "writeln",
+                    "write",
+                    "eprintln",
+                    "eprint",
+                    "format_args"
+                ))
+                return
+
+            val rangeInHost = (context.kind!! as RsLiteralKind.String).offsets.value ?: return
+            (context as? RsLitExprImpl)?.overridenEscaper = RsSimpleMultiLineEscaper(context)
```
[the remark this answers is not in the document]
rrevenantt: But then there will be no escapes highlighting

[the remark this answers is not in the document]
vlad20012: This is a discussion question. I think that we should separate rust escape codes from format pattern lexing. I have these arguments:
1. It's a good division of responsibility. Imagine that Rust escape codes are changed. In this case, you have to implement changes twice. Bugs also will be duplicated. Or there will be different bugs (as I see, you missed to implement `\x` escape code, for example).
2. If it is a language injection, it does not have to know about Rust (another language) escape codes. Imagine a project where "Rust Format pattern" used inside some TOML file, for example, for i18n purposes. TOML has different escape codes than Rust, so the user will not be able to inject "Rust Format pattern" lang into such TOML literals.
3. Nobody uses injections this way across JetBrains infrastructure. This is a Kotlin: 
image
The first line is RegExp injection into Kotlin string literal, and the second line is a literal without injection. There are no escape codes highlighting in RegExp injection! `\n` is specifically lexed (see `RegExpTT.CTRL_CHARACTER`), but the lexer doesn't know written as "\n" in the origin literal. It knows only that it is LF char (which can be written just as a linebreak in a literal).

So are we. This is Rust:
image
The first line is RegExp injection into Rust string literal, and the second line is a literal without injection. No escape codes highlighting inside injected code.

CLion highlighting is very interesting:
image
It's not a language injection because I see the available intention to injection something:
image
(It should be unavailable if it is already injected). Soo looks like in CLion this feature was made without language injection. I'll try to investigate it more deeply and assess whether we should do the same.

But anyway, if it is language injection, I think such hacks should not be used and we should come to terms with the fact that escape codes aren't highlighted in injections.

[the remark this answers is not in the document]
rrevenantt: Also i have just checked and following code successfully works
```
println!(" \u{007b}}",1)
```
So looks like you are right and it is a bad idea.

I can't look into clion myself, so i will wait for you to check it out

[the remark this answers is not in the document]
vlad20012: I found that CLion uses regular `Annotator` to highlight format pattern specifiers and it parses them with, umm, regexps, so I don't think this is a good role model =/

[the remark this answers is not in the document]
rrevenantt: > so I don't think this is a good role model =/

By '"this" do you mean language injection or annotator approach?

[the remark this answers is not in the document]
vlad20012: I meant regexp approach =D
It's possible to use a lexer (or a parser) inside an annotator to locate highlightable ranges. Such approach __can__ be used, but it's necessary to understand trade-offs... which are not clear to me.
Looks like we anyway will be forced to use some kind of pattern parser in annotator to highlight errors (e.g. parameters quantity `println!("{}", 1, 2)`). I don't know what option is better... I'll try to think of something a bit later.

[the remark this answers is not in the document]
vlad20012: I discussed with @undin and found that original issue #2256 wasn't supposed to be solved with language injections =/ Now I think it's reasonable to use Annotator approach, but I think that string escaping should not be re-implement. I see a rough plan so:

1. Make something like `RsFormatPatternAnnotator`
2. Locate string literals that are format patterns
3. Use `RsLitExpr.kind` to take offsets and literal text
4. Use `parseRustStringCharacters` to take unescaped text and offsets
5. Parse text with our parser
6. Locate elements that should be highlighted in the parsed PSI
7. Map start and end offset of annotation with an array we got from `parseRustStringCharacters`

Then the same (single-time) parsed pattern could be used to highlight illegal format macro usage (too many or too little arguments, wrong types, etc). This is unachievable with language injections. In this case we should use a name like `RsFormatMacroAnnotator` instead of `RsFormatPatternAnnotator`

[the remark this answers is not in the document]
rrevenantt: It looks like you are suggesting to use standalone PSI parser and lexer. But to me it seems that there is no reason to use PSI parser and lexer with handful of files instead of one line regex to parse such simple syntax especially considering that we are not going to use any PSI infrastructure. Also it will create a lot of useless objects. Also I will have to manually handle parser errors.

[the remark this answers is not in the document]
vlad20012: I'm not much familiar with regexes, so for me it's simpler to write a standalone parser =))0)
Note that it would be cool to provide a full analysis of format macros in the future (), so I think it would be handy to have PSI for format pattern to simplify analyzing of complex patterns like `println!("{number:>0width$}", number=1, width=6);`. Maybe regexes are enough, I don't sure. Maybe I just don't like regexes

rrevenantt:
@vlad20012 you can check out my implementation via annotanor and regex. Although I am not sure about current highlighting, because new rust user might misunderstand it as part of rust syntax like kotlin templates. Maybe mix highlighting color with string one in some proportion? Or background like language injection might be good enough?

vlad20012:
I'm sorry for so long response... For some reason, I wanted to avoid this thing.

Just launched it and found this:
image

rrevenantt:
Hm, I had copied tests from rustc specifically to not miss any test cases, apparently rustc somehow does not have this basic test(i have checked). clearly i should not have relied on compiler test set that much

I fixed it.

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+        fun AnnotationHolder.highlightArgument(range: IntRange?, color: RsColor = RsColor.IDENTIFIER) {
+            if (range != null && !range.isEmpty())
+                createAnnotation(
+                    ARG_HIGHLIGHTING,
+                    range.toSourceRange(),
+                    null)
+                    .textAttributes = color.textAttributesKey
+        }
+
+        val key = RsColor.VALID_STRING_ESCAPE
```
vlad20012:
```suggestion
        val color = RsColor.VALID_STRING_ESCAPE
```

vlad20012: Let's use a special color for this just like CLion does for C/C++. Add this to `RsColor`: `FORMAT_SPECIFIER("Literals//Strings//Format specifier in string argument", Default.VALID_STRING_ESCAPE),`

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+        else -> -1
+    }
+
+    companion object {
+        private val ARG_HIGHLIGHTING = HighlightSeverity(
+            "ARG_HIGHLIGHTING",
+            HighlightSeverity.INFORMATION.myVal + 1
+        )
+        val formatParser = Regex("""\{\{|}}|(\{([^}]*)}?)|(})""")
+        const val argument = """([a-zA-Z_][\w+]*|\d+)"""
```
vlad20012:
```suggestion
        @Language("regexp") private const val argument = """([a-zA-Z_][\w+]*|\d+)"""
```

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+        "writeln" -> 1
+        else -> -1
+    }
+
+    companion object {
+        private val ARG_HIGHLIGHTING = HighlightSeverity(
+            "ARG_HIGHLIGHTING",
+            HighlightSeverity.INFORMATION.myVal + 1
+        )
+        val formatParser = Regex("""\{\{|}}|(\{([^}]*)}?)|(})""")
```
vlad20012: private

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+        const val argument = """([a-zA-Z_][\w+]*|\d+)"""
+        val formatArgParser = Regex("""
```
vlad20012: private

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+        "eprintln" -> 0
+        "eprint" -> 0
+        "format" -> 0
+        "write" -> 1
+        "writeln" -> 1
+        else -> -1
+    }
+
+    companion object {
+        private val ARG_HIGHLIGHTING = HighlightSeverity(
```
vlad20012: Why is it needed? Looks like everything works with simple `HighlightSeverity.INFORMATION`. Next annotations take precedence over the previous ones.

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+            "ARG_HIGHLIGHTING",
+            HighlightSeverity.INFORMATION.myVal + 1
+        )
+        val formatParser = Regex("""\{\{|}}|(\{([^}]*)}?)|(})""")
+        const val argument = """([a-zA-Z_][\w+]*|\d+)"""
+        val formatArgParser = Regex("""
+^(?<id>$argument)?
+(:
+(.?[\^<>])?[+\-]?\#?0?
```
vlad20012: Redundant character escape? 
image

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+
+        val arguments = formatParser.findAll(unescapedText.toString())
+        var currentOffset = 0
+
+        fun IntRange.toSourceRange(additionalOffset: Int = currentOffset): TextRange =
+            TextRange(sourceMap[first + additionalOffset], sourceMap[last + additionalOffset] + 1)
+                .shiftRight(formatStr.startOffset + rawTextRange.startOffset)
+
+        fun AnnotationHolder.highlightArgument(range: IntRange?, color: RsColor = RsColor.IDENTIFIER) {
+            if (range != null && !range.isEmpty())
```
vlad20012: We usually add braces to multiline branches

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+                createAnnotation(
+                    ARG_HIGHLIGHTING,
+                    range.toSourceRange(),
+                    null)
+                    .textAttributes = color.textAttributesKey
```
vlad20012:
```
    null
).textAttributes = color.textAttributesKey
```

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+        for (arg in arguments) {
+            currentOffset = 0
+            val textRange = arg.range.toSourceRange()
+            holder.createAnnotation(highlightSeverity, textRange, null)
+                .textAttributes = key.textAttributesKey
+
+            if (arg.groups[3] != null) {
+                holder.createErrorAnnotation(
+                    arg.range.toSourceRange(),
+                    "Invalid format string: unmatched '}'")
```
vlad20012: We usually add a newline before `)` in multiline function calls.

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+
+                val inside = arg.groups[2]!!
+                currentOffset = inside.range.first
+                val matchResult = formatArgParser.find(inside.value)!!
+
+                if (!arg.groups[1]!!.value.endsWith("}")) {
+                    val possibleEnd = matchResult.value.length - 1
+                    holder.createErrorAnnotation(
+                        (possibleEnd..possibleEnd).toSourceRange(),
+                        "Invalid format string: } expected.\nIf you intended to print `{` symbol, you can escape it using `{{`")
```
vlad20012: Same here `)`

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+                    continue
+                }
+
+                val validParsedEnd = matchResult.range.last + 1
+                if (validParsedEnd != inside.value.length) {
+                    holder.createErrorAnnotation(
+                        (validParsedEnd until inside.value.length).toSourceRange(),
+                        "Invalid format string")
```
vlad20012: And here `)`

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+            if (arg.groups[3] != null) {
+                holder.createErrorAnnotation(
+                    arg.range.toSourceRange(),
+                    "Invalid format string: unmatched '}'")
+            }
+            if (arg.groups[1] != null) {
+
+                val inside = arg.groups[2]!!
+                currentOffset = inside.range.first
+                val matchResult = formatArgParser.find(inside.value)!!
```
vlad20012: Hmm, let's add a comment about why this can't be null. Or even better pattern is `?: error("This can't be null because ...")`

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+                val validParsedEnd = matchResult.range.last + 1
+                if (validParsedEnd != inside.value.length) {
+                    holder.createErrorAnnotation(
+                        (validParsedEnd until inside.value.length).toSourceRange(),
+                        "Invalid format string")
+                } else {
+                    holder.highlightArgument(matchResult.groups["id"]?.range)
+                    holder.highlightArgument(matchResult.groups["width"]?.range)
+                    holder.highlightArgument(matchResult.groups["precision"]?.range)
+                    holder.highlightArgument(matchResult.groups["type"]?.range, RsColor.FUNCTION)
```
vlad20012: Do I understand correctly that this is FUNCTION because it refers to a `fmt` function of some format trait?

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -0,0 +1,127 @@
...
+            ?.formatMacroArgList
+            ?.getOrNull(macroToFormatPos(formatMacro.macroName))
+            ?.descendantOfTypeStrict<RsLitExpr>()
+            ?: return
+
+        val literalKind = (formatStr.kind as? RsLiteralKind.String) ?: return
+        val rawTextRange = literalKind.offsets.value ?: return
+        val (unescapedText, sourceMap, _) = parseRustStringCharacters(rawTextRange.substring(literalKind.node.text))
+
+        val arguments = formatParser.findAll(unescapedText.toString())
```
vlad20012:
```suggestion
        val arguments = formatParser.findAll(unescapedText)
```

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
rrevenantt: AFAIK, it is false positive warning, but i will check this again to be sure.

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
rrevenantt: Kind of. I didn't thought about it exactly like this, but it is also valid reason. We can even add go to definition to corresponding format trait

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
rrevenantt: Sorry, for the amount of code style issues. Rustfmt is very aggressive in formatting(sometimes even too aggressive), Kotlin formatter is usually also enough, so usually don't even think about it.

rrevenantt:
`println!("123 {{ 321 }} 123"); // should not be highlighted?` 
not sure what do you expect here, highlighting looks fine to me.

vlad20012:
> not sure what do you expect here, highlighting looks fine to me.
Oh, yes, `{{` means "{", so it's something like "escape" sequence. Ok, I'm sorry

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
rrevenantt: Yep, it is false positive when comments are enabled in regexp.

rrevenantt:
I have also added `format_args` macro. And reformatted regexp a bit because I found out how to make language injector to know that I have enabled comments in regexp.

BTW, new related rfc was approved since then. https://github.com/rust-lang/rfcs/pull/2795

vlad20012:
:+1: 
bors r+

vlad20012:
@rrevenantt FYI :)
https://www.reddit.com/r/rust/comments/fcdql8/intellij_rust_changelog_117/fja9wfv/

PR #3871, opened 2019-05-21
RES: resolve extern crate items with alias in 2018 edition

Undin:


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -1244,6 +1244,22 @@ fun processNestedScopesUpwards(
     if (processor(ScopeEvent.STAR_IMPORTS)) return false
 
     if (Namespace.Types in ns) {
+        if (scopeStart.isEdition2018 && scopeStart.contextOrSelf<RsMod>()?.isCrateRoot != true) {
```
vlad20012:
```suggestion
        if (scopeStart.isEdition2018 && scopeStart.containingMod?.isCrateRoot != true) {
```

Issue #3868, opened 2019-05-21
`yield` syntax is experimental error with multiple features


Seeker14491:
This code shows no errors in the editor, as expected:

```rust
#![feature(generators)]

fn main() {
    let _generator = || {
        yield 1_u32;
    };
}
```

However, if you add more features next to the `generators` feature line, then an erroneous error is shown, as in the following:

```rust
#![feature(generators, box_syntax)] // <- another feature added to the same line

fn main() {
    let _generator = || {
        yield 1_u32; // `yield` syntax is experimental error
    };
}
```

## Environment

* **Intellij-Rust plugin version:** 0.2.97.2123-191
* **Rust toolchain version:** rustc 1.36.0-nightly (7d5aa4332 2019-05-16)
* **IDE name and version:** CLion 2019.1.2
* **Operating system:** Windows 10 x64

PR #3873, opened 2019-05-21
Fix feature indexing

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/3868.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/index/RsFeatureIndex.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/index/RsFeatureIndex.kt
@@ -28,8 +28,11 @@ class RsFeatureIndex : StringStubIndexExtension<RsInnerAttr>() {
         fun index(stub: RsInnerAttrStub, sink: IndexSink) {
             val metaItem = stub.psi.metaItem
             if (metaItem.name == "feature") {
-                val featureName = metaItem.metaItemArgs?.metaItemList?.singleOrNull()?.name ?: return
-                sink.occurrence(KEY, featureName)
+                val features = metaItem.metaItemArgs?.metaItemList.orEmpty()
+                for (feature in features) {
+                    val featureName = feature.name ?: continue
+                    sink.occurrence(KEY, featureName)
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/stubs/index/RsFeatureIndex.kt
```diff
@@ -28,8 +28,11 @@ class RsFeatureIndex : StringStubIndexExtension<RsInnerAttr>() {
         fun index(stub: RsInnerAttrStub, sink: IndexSink) {
             val metaItem = stub.psi.metaItem
             if (metaItem.name == "feature") {
-                val featureName = metaItem.metaItemArgs?.metaItemList?.singleOrNull()?.name ?: return
-                sink.occurrence(KEY, featureName)
+                val features = metaItem.metaItemArgs?.metaItemList.orEmpty()
+                for (feature in features) {
+                    val featureName = feature.name ?: continue
+                    sink.occurrence(KEY, featureName)
```
Undin: Hm, looks like we need to bump `org.rust.lang.core.stubs.RsFileStub.Type#getStubVersion` to launch reindexing and avoid old behaviour with new plugin version

mchernyavsky: Fixed.

vlad20012: We also can bump version for this index only:
```
override fun getVersion(): Int = RsFileStub.Type.stubVersion + 1
```

Undin: @vlad20012 Looks like it improves reindexing only for users with nightly plugin because `RsFileStub.Type.stubVersion` was already increased after prev release.
So, is it really worth it?

vlad20012: @Undin ok, makes sense

mchernyavsky:
Fixed.

PR #3880, opened 2019-05-22
Provide a User Agent to crates.io

sgrif:
As per [crates.io's crawler policy](https://crates.io/policies#crawlers), all bots must provide a user agent identifying themselves. While sending "Java/1.8.0_202-release" is technically following the policy, it makes it difficult for us to identify the impact this plugin's traffic has on our service, or contact you if we have potential improvements that could be made or changes we need you to make.


sgrif:
Done

PR #3895, opened 2019-05-24
COMP: Add `await` to completion list if receiver's type is `impl Future`

mchernyavsky:


src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -176,7 +199,26 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
             .inside(psiElement<RsFunction>())
     }
 
+    private fun postfixAwaitPattern(): PsiElementPattern.Capture<PsiElement> {
+        val parent = psiElement<RsFieldLookup>()
+            .with(object : PatternCondition<RsFieldLookup>("RsPostfixAwait") {
+                override fun accepts(t: RsFieldLookup, context: ProcessingContext?): Boolean {
+                    if (context == null || !t.isEdition2018) return false
+                    val receiver = CompletionUtil.getOriginalOrSelf(t.receiver)
```
vlad20012: Use `safeGetOriginalOrSelf()` here

mchernyavsky: Fixed.

PR #3898, opened 2019-05-24
TYPE:fixed exit points highlighting

rrevenantt:
fixed exit points highlighting for :
 - try blocks
 - async blocks
 - caret at `?`

### DIFF
--- a/src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
+++ b/src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
@@ -10,27 +10,31 @@ import com.intellij.codeInsight.highlighting.HighlightUsagesHandlerFactoryBase
 import com.intellij.openapi.editor.Editor
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.util.Consumer
 import org.rust.lang.core.cfg.ExitPoint
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.Q
 import org.rust.lang.core.psi.RsElementTypes.RETURN
-import org.rust.lang.core.psi.RsFile
-import org.rust.lang.core.psi.RsFunction
-import org.rust.lang.core.psi.RsLambdaExpr
-import org.rust.lang.core.psi.RsTryExpr
 import org.rust.lang.core.psi.ext.ancestors
 import org.rust.lang.core.psi.ext.elementType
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/cfg/ControlFlowGraph.kt
```diff
@@ -134,34 +131,45 @@ sealed class ExitPoint {
...
+        fun process(fn: PsiElement?, sink: (ExitPoint) -> Unit) {
+            fn?.acceptChildren(ExitPointVisitor(sink))
         }
     }
 }
 
 private class ExitPointVisitor(
     private val sink: (ExitPoint) -> Unit
 ) : RsVisitor() {
+    var inTry = false
```
ortem: I think you need a counter instead because of nested `try`:
```Rust
fn foo() -> i32 {
    let a: Result<i32, ()> = try {
        let tmp: Result<i32, ()> = try { Err(())? };
        tmp? + Ok(42)? // wrong highlighting currently
    };
    return/*caret*/ 1;
}
```
Please add such tests

src/main/kotlin/org/rust/ide/highlight/RsHighlightExitPointsHandlerFactory.kt
```diff
@@ -51,11 +55,18 @@ private class RsHighlightExitPointsHandler(editor: Editor, file: PsiFile, var ta
...
-        when (fnOrLambda) {
-            is RsFunction -> ExitPoint.process(fnOrLambda, sink)
-            is RsLambdaExpr -> ExitPoint.process(fnOrLambda, sink)
+        loop@ for (it in target.ancestors) {
+            when {
+                it is RsBlockExpr
+                    && it.isTry
+                    && target.elementType == Q -> Unit
+                it is RsBlockExpr
+                    && it.isAsync -> ExitPoint.process(it.block, sink)
```
ortem: Currently, there are no tests for this case (`return` highlighting inside `async { ... }`). Please add it. 

Also, the code can be rewritten in a more readable and coverage-friendly way (and then code coverage will show that `async` case is uncovered), like this:
```Rust
for (ancestor in target.ancestors) {
    if (ancestor is RsBlockExpr && ancestor.isTry && target.elementType == Q) {
        break
    } else if (ancestor is RsBlockExpr && ancestor.isAsync) {
        ExitPoint.process(ancestor.block, sink)
        break
    } else if (ancestor is RsFunction) {
        ExitPoint.process(ancestor.block, sink)
        break
    } else if (ancestor is RsLambdaExpr) {
        ExitPoint.process(ancestor.expr, sink)
        break
    }
}
```

rrevenantt:
Done.

ortem:
Thank you!
bors r+

PR #3833, opened 2019-05-11
ANN: Make public fix

shevtsiv:

PR #3903, opened 2019-05-25
Add external package with the same workspace

shevtsiv:
There is already `dep-lib` for separation code into different crates,but it has PackageOrigin.DEPENDENCY, thus it is treated like external dependency which we cannot modify. However, sometimes it is useful to create tests that rely on different crates, but the same workspace. For example, there can be some fixes that cannot modify external dependencies, but have to modify any code that is located within one workspace, even in different crates.

As a real world example, I suggest you to consider [my another commit](https://github.com/intellij-rust/intellij-rust/pull/3833/commits/a948c816792899f5cba23290ba83c314aad23da8) that has to be tested using this feature: There can be multiple crates within one workspace, so `Make public` fix has to be able to modify each of them, since they are not external dependencies. Unfortunately, I cannot test this behavior because there is no possibility to create test environment with multiple crates within one workspace.


shevtsiv:
I am sorry, but as it turned out, we can reach the same behavior using `//- main.rs` and `//- lib.rs`:
https://github.com/intellij-rust/intellij-rust/pull/3833/commits/13b0bddb734a70da26db9ce7286ffd589989b521

PR #3905, opened 2019-05-25
TY&RES: disambiguate trait associated items by scope & trait bounds

### LINKED ISSUE #3689: Impl function fails to resolve with new macro expansion engine
First off, THANK YOU for the new macro expansion engine announced in https://intellij-rust.github.io/2019/04/10/changelog-96.html. This bug report is for a minor issue with the new engine, but overall it has significantly improved my development experience!

## Environment

* **Intellij-Rust plugin version:** 0.2.96.2122-191
* **Rust toolchain version:** cargo 1.34.0 (6789d8a0a 2019-04-01)
* **IDE name and version:** IntelliJ 2019.1 Community Edition
* **Operating system:** MacOS 10.14.1

## Problem description

After enabling the new macro expansion engine, IntelliJ was, for the first time, able to navigate to the macros where many of my functions are defined

For some reason, one of my function calls still fails to resolve: https://github.com/ramosbugs/oauth2-rs/blob/0d306c115f62c87453404d448540f8080f6165a9/tests/lib.rs#L80

The call site is as follows:
```rust
#[test]
fn test_code_verifier_challenge() {
    // Example from https://tools.ietf.org/html/rfc7636#appendix-B
    let code_verifier =
        PkceCodeVerifierS256::new("dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk".to_string());
    assert!(
        code_verifier.code_challenge().as_str() == "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM"
    );
}
```

For the call to `code_verifier.code_challenge()`, IntelliJ isn't able to find the definition of `code_challenge`, which is defined in the macro expansion at: https://github.com/ramosbugs/oauth2-rs/blob/0d306c115f62c87453404d448540f8080f6165a9/src/lib.rs#L617

Note that IntelliJ is able to resolve `PkceCodeVerifierS256::new()`, which is part of the same macro expansion. It seems to be able to find all of the functions in that expansion except for `code_challenge()` and `authorize_url_params()`.

vlad20012:
Fixes #3689

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
@@ -81,6 +81,23 @@ fun collectResolveVariants(referenceName: String, f: (RsResolveProcessor) -> Uni
     return result
 }
 
+fun <T : ScopeEntry> collectResolveVariantsAsScopeEntries(referenceName: String, f: ((T) -> Boolean) -> Unit): List<T> {
+    val result = mutableListOf<T>()
+    f { e ->
+        if ((e == ScopeEvent.STAR_IMPORTS) && result.isNotEmpty()) {
+            return@f true
+        }
+
+        if (e.name == referenceName) {
+            // de-lazying. See `RsResolveProcessor.lazy`
+            e.element ?: return@f false
+            result += e
+        }
+        false
+    }
+    return result
+}
+
 fun pickFirstResolveVariant(referenceName: String, f: (RsResolveProcessor) -> Unit): RsElement? {
     var result: RsElement? = null
     f { e ->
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -81,6 +81,22 @@ fun collectResolveVariants(referenceName: String, f: (RsResolveProcessor) -> Uni
     return result
 }
 
+fun <T : ScopeEntry> collectResolveVariantsAsSEs(referenceName: String, f: ((T) -> Boolean) -> Unit): List<T> {
```
Undin: Maybe rename it somehow? because at this moment the name is rather unclear
Also, it would be great to have some docs about why when it can be used

src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -81,6 +81,22 @@ fun collectResolveVariants(referenceName: String, f: (RsResolveProcessor) -> Uni
...
+    val result = mutableListOf<T>()
+    f { e ->
+        if ((e == ScopeEvent.STAR_IMPORTS) && result.isNotEmpty()) {
+            return@f true
+        }
+
+        if (e.name == referenceName) {
+            e.element ?: return@f false // de-lazying
```
Undin: Let's add a link to `org.rust.lang.core.resolve.Processors#lazy` method and add a note to method's doc about it because it looks quite important

PR #3922, opened 2019-05-30
INSP: Don't check mutability of expr with unknown type

### LINKED ISSUE #3782: False positive E0594 in for loop.
## Environment

* **Intellij-Rust plugin version:** 0.2.97.2123-191
* **Rust toolchain version:**  rustc 1.36.0-nightly
* **IDE name and version:** Clion 2019.01
* **Operating system:** macOS 10.14.4

## Problem description

The Rust plugin gives a false positive E0594 on the `x.f = 1`; assignment in this code:

```rust

#[derive(Default, Copy, Clone)]
struct S {
    f: u32
}

fn main() {
    let mut v = [S::default(); 5];
    
    for x in &mut v {
        x.f = 1;
    }
}
```
Screen Shot 2019-05-03 at 14 15 44

However, compiling this code with Rust works just fine:

```
/Users/rb/.cargo/bin/cargo run --color=always --release
    Finished release [optimized] target(s) in 0.02s
     Running `target/release/mapgen`

```

### LINKED ISSUE #3914: False-positive "Cannot borrow immutable local variable as mutable"
The following code compiles, but shows an error in the plugin (shown in the comment)

```rust
fn main() {
    f(|x| &mut x.0);
            // ^^^ Cannot borrow immutable local variable `x.0` as mutable
}

fn f(_g: impl Fn(&mut (u8,)) -> &mut u8) {}
```

**Plugin version:** 0.2.98.2124-191

ortem:
Fixes #3914 
Fixes #3782

### DIFF
--- a/src/test/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspectionTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspectionTest.kt
@@ -5,6 +5,9 @@
 
 package org.rust.ide.inspections
 
+import org.rust.ProjectDescriptor
+import org.rust.WithStdlibRustProjectDescriptor
+
 class RsAssignToImmutableInspectionTest : RsInspectionsTestBase(RsAssignToImmutableInspection()) {
 
     fun `test E0594 assign to immutable borrowed content`() = checkByText("""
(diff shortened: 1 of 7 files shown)


src/test/kotlin/org/rust/ide/inspections/RsAssignToImmutableInspectionTest.kt
```diff
@@ -66,6 +70,7 @@ class RsAssignToImmutableInspectionTest : RsInspectionsTestBase(RsAssignToImmuta
         }
     """)
 
+    @ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
```
Undin: Do we really need stdlib in this test?

ortem: @Undin Yes, because the type of `a[1]` is `<unknown>` otherwise

Undin: Ok

PR #3929, opened 2019-05-31
Improve "fill match arms" intention

vlad20012:
1. Support match ergonomics (`match &Some(1) {}`)
2. Support custom enums with variant names in the scope. Previously only Option/Result was supported this way.
```rust
mod foo {
    enum E { A, B }
}
use foo::E;
use foo::E::A;
fn foo(x: &E) {
    match x/*caret*/ {}
}
```
=>
```rust
match x {
    A => {},
    E::B => {},
}
```


mchernyavsky:
bors retry

mchernyavsky:
bors r=vlad20012

Issue #1078, opened 2017-03-20
Find Usages does not work if a file uses include!


raphaelcohn:
When a file is included using include! and contains a struct, find usages only finds usages within the file that was included. It should find usages across the project.

See also #771.

Undin:
It works if the plugin knows where including file is located.
So I think it can be closed now - #1908 is enough to track the status of `include` support

Issue #771, opened 2016-11-10
Autocompletion does not work when files use include!


raphaelcohn:
I'm working with a very large code base that uses `include!` to break up modules that otherwise would be excessively large. We also use it to make sure our code is well organized, with a file-per-major item pattern (a struct, a pub fn, a pub type, that sort of thing). It make finding things a snap - find, TextMate cmd-T, etc. And it works with intellij-rust too - Cmd-N is wonderful.

What doesn't work is autocompletion. Autocompletion works perfectly for things in the stdlib. And within a file referencing other items. But not across files. Likewise, find usages does not work, either, and we're not confident refactor change name will. For us, this is a major reason to use an IDE.

(We realise we're not doing things "The Rust Way", but the "Rust Way" for modules doesn't scale. It's an anti pattern. It's more code, and it creates more collisions that need not occur when many developers are at work on a code base. And the the file system does not represent the contents. A million and one mod X / pub use X aren't appropriate, along with all the duplicated use X  statements that creates).

DGolubets:
While I don't agree with using include this way, I stumbled across a library that used it (and it has just 6 modules overall.. a joke, right?) and it's totally of the grid for autocomplete.

Techcable:
More importantly, this allows using generated code like [bindgen ffi bindings](https://github.com/servo/rust-bindgen). Normally I can sacrifice features like macros for the autocomplete productivity gains, but bindgen is so useful I can't go without it. It'd be awesome if this worked for this use case.

Ekranos:
Is there any update on this? Since I got the same problem and this issue is open since Nov 2016.

Also I have to agree that the Rust module way just doesn't really scale. I am working on a project with ~150 structs in one module (and yes, they belong this tightly together...) and it would be a blessing to be able to split them into multiple files while retaining autocomplete.

And including generated bindings which get completed would be really great too.

raphaelcohn:
@Techcable You might like my tool [bindgen-wrapper](https://github.com/lemonrock/bindgen-wrapper) that makes extensive use of include! to make FFI bindings intelligible... An example in use are the bindings for ucx-sys in [ucx](https://github.com/lemonrock/ucx).

dweidenfeld:
Well this is also a problem for other *gen tools like: https://github.com/jsgf/rust-xdr/tree/master/xdrgen

raphaelcohn:
@dweidenfeld Cool - I didn't know of xdrgen. Thanks!

[...]

PR #3935, opened 2019-06-01
MACRO&RES: initial version of include macro support

### LINKED ISSUE #3563: Cannot find declaration (large project: Servo)
Hi everyone, thank you so much for developing this plugin over at JetBrains. Here's my issue:

## Environment

* **Intellij-Rust plugin version:** `0.2.0.2114-182`
* **Rust toolchain version:** `rustup 1.17.0 (069c88ed6 2019-03-05)`
* **IDE name and version:** IntelliJ Ultimate 2018.2.2
* **Operating system:** `x86_64 Linux 5.0.2-arch1-1-ARCH`

## Problem description
When navigating files in subdirectory [servo/servo](https://github.com/servo/servo/tree/cc2d20315149887ee1f125deac1a1807722a50ee/ports/servo), it appears that IntelliJ is not able to detect crate/root information correctly, and will give `Cannot find declaration to go to` for basically all modules being used in this subdirectory.

Here is a screenshot of what I am seeing, specifically:

Screenshot_20190319_150809

Furthermore, closing the project and restarting it does not seem to help here. I think my issue is similar/the same as what occurred for #1346, except that I have imported Servo as a project proper (i.e. from its root).

## Steps to reproduce

1. Open project
2. Open file
3. Attempt go-to definition

Let me know if some more information/debugging on my end would be of use if this is a legitimate bug.

Undin:
Add initial support of `include!` macro. It brings all common code insight features (like name resolution, code completion, etc.) in both including and included files. Also, it should fix false positive editor notification for included files.

Restrictions:
* only string literals are supported as arguments
* `include` macro call should be located only in a module. At this moment, if `include` macro is called in a function, it won't be indexed and as a result, won't be found via the corresponding index. So the current implementation works only with instances of `RsMod`

Fixes #3563
Fixes problems described in #771 and #1078 in certain cases
The initial step of #1908

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -1325,7 +1325,11 @@ LogMacroArgument ::=
 
 private IncludeMacro ::= &(("include" | "include_str" | "include_bytes") '!') PathWithoutTypes '!' IncludeMacroArgument
 
-IncludeMacroArgument ::= <<any_braces (AnyExpr ','?)>>
+IncludeMacroArgument ::= <<any_braces (AnyExpr ','?)>> {
+  extends = "org.rust.lang.core.psi.ext.RsStubbedElementImpl<?>"
+  stubClass = "org.rust.lang.core.stubs.RsPlaceholderStub"
+  elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
+}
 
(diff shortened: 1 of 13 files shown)


vlad20012:
:+1: :+1: :+1: 
bors r+

Issue #3928, opened 2019-05-31
Add intention to expand `S { .. }` in patterns


vlad20012:
In rust it's possible to skip some fields in patterns by adding `..`:
```rust
struct S { a: i32, b: i32, c:i32 }
fn foo(s: S) {
    let S { a, .. } = s;
}
```
And for tuple struct case:
```rust
struct S(i32, i32, i32);
fn foo(s: S) {
    let S(a, ..) = s;
}
```

It would be nice to have an intention to replace such `..` with the actual fields.

shevtsiv:
If no objection, I would like to work on this issue.

vlad20012:
Maybe it's reasonable to implement analysis for E0027 & E0023 first

Undin:
Fixed in #3996

PR #3936, opened 2019-06-01
MACRO: Add expansion of compiler built-in macros

afdw:
Adds support for expanding `env!`, `option_env!`, `concat!`, `include!` macros:
image
Currently some of the caching of macro expansions is commented out.
Eventually should fix https://github.com/intellij-rust/intellij-rust/issues/1908.
The current problem is getting the value of the `OUT_DIR` environment variable, as `cargo metadata` does not report it. It can only have a meaningful value once the project is built at least once, and can change after subsequent builds. Any thoughts of what is possible to do here?

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -7,6 +7,7 @@ package org.rust.lang.core.macros
 
 import com.intellij.lang.PsiBuilder
 import com.intellij.lang.PsiBuilderUtil
+import com.intellij.lang.injection.InjectedLanguageManager
 import com.intellij.lang.parser.GeneratedParserUtilBase
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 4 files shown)


vlad20012:
Umm, we are currently work on `include` macro in #3935 ...
And we're going to support soon other builtin macros that you mentioned, soo...

afdw:
So this was a bit unfortunate, as I have started working on this PR before https://github.com/intellij-rust/intellij-rust/pull/3935 was created. So I think this PR can be closed then?
Please also check your implementation on code like on the screenshot after you are done, Rust actually allows you to use custom macros inside the arguments of the built-in macros.

Undin:
> So this was a bit unfortunate, as I have started working on this PR before #3935 was created.

To avoid such situations, issues that supposed to be fixed by us usually have an assignee (including #1908). Also, a comment (like [this](https://github.com/intellij-rust/intellij-rust/issues/3928#issuecomment-497788352)) in the issue that you are going to fix/implement can inform other contributors and prevent simultaneous development the same feature 

> So I think this PR can be closed then?

I think yes. Your current implementation copies the whole included file and creates its copy in macro expansion folder. It should work good enough if such file is located in dependency (however "find usages" won't work as expected). But it prevents providing proper code insight features: name resolution, completion, etc., in included file. It's important when such file is part of project code base (see #771).

> Please also check your implementation on code like on the screenshot after you are done, Rust actually allows you to use custom macros inside the arguments of the built-in macros.

It's a good point. Thanks!
I'm not sure that we will support arbitrary arguments inside `include` macro invocation because it complicates to use some platform API (for instance, indexing). In the short term, we are going to support arguments similar to `concat!(env!("OUT_DIR"), "/bindings.rs")` and string literals. It should cover most cases

PR #3951, opened 2019-06-04
ANN: fix trait members visibility (false-positive E0603)

vlad20012:


src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -1171,6 +1171,27 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class.java) {
         use self::foo as bar;
     """)
 
+    fun `test no E0603 for trait with impl ni a child mod`() = checkErrors("""
```
mibac138: typo: `ni` -> `in`

vlad20012: Fixed

PR #3961, opened 2019-06-04
UI: introduce detached notification provider

Undin:
Extract "File is not included in module tree, ..." notification introduced in #3826 to separate notification provider and allow to disable such notifications (https://github.com/intellij-rust/intellij-rust/issues/771#issuecomment-496451670)

### DIFF
--- a/src/main/kotlin/org/rust/ide/notifications/RsNotificationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/notifications/RsNotificationProvider.kt
@@ -6,6 +6,7 @@
 package org.rust.ide.notifications
 
 import com.intellij.ide.util.PropertiesComponent
+import com.intellij.openapi.fileEditor.FileEditor
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.ui.EditorNotificationPanel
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/notifications/RsNotificationProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.ui.EditorNotificationPanel
+import com.intellij.ui.EditorNotifications
+
+abstract class RsNotificationProvider(
+    protected val project: Project
+) : EditorNotifications.Provider<EditorNotificationPanel>() {
+
+    protected abstract val disablingKey: String
```
vlad20012: Consider moving it to the constructor. (and remove redundant `const`s)

Undin: I've rewritten implementation as we had discussed in person.
Now it's possible to provide `disablingKey` per file. So I'd prefer not to pass any constant into the constructor to keep base implementation the most flexible

vlad20012:
bors r_

mchernyavsky:
bors retry

PR #3977, opened 2019-06-06
COMP: Add completion for path-based keys in Cargo.toml

mibac138:
path-completion

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/CargoTomlPathCompletion.kt
+++ b/toml/src/main/kotlin/org/rust/toml/CargoTomlPathCompletion.kt
@@ -1,0 +1,71 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionProvider
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.completion.CompletionUtil
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.icons.AllIcons
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.util.ProcessingContext
+import org.rust.lang.core.completion.DEFAULT_PRIORITY
+import org.rust.lang.core.completion.withPriority
+import org.rust.lang.core.psi.ext.ancestorOrSelf
+import org.toml.lang.psi.TomlValue
+
+class CargoTomlPathCompletion private constructor(
+    private val target: Target
+) : CompletionProvider<CompletionParameters>() {
+
+    private enum class Target {
+        File, Directory, Any;
+
+        fun priorityOf(file: VirtualFile): Double = when (this) {
+            File -> if (file.isDirectory) DEFAULT_PRIORITY else 1.0
+            Directory -> if (file.isDirectory) 1.0 else DEFAULT_PRIORITY
+            Any -> DEFAULT_PRIORITY
+        }
+    }
+
+    override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
+        val tomlValue = parameters.position.ancestorOrSelf<TomlValue>() ?: return
+        val containingFile = tomlValue.containingFile.originalFile.virtualFile ?: return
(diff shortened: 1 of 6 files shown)


vlad20012:
Hi!
Take a look it the implementation of include macro path reference. 
Peek 2019-06-06 21-35

See RsLitExprReferenceContributor/RsFileReferenceProvider/RsLiteralFileReferenceSet. It is more idiomatic and also provides go-to-declaration, rename, find usages (and really should provide error reporting if file is undefined, but it is not work for include macro for some reason), etc. I want to implement it this way a lot time ago, but it's rather hard (it lest, if make it correctly).

The most complex part is that strings can have [escape codes](https://github.com/toml-lang/toml#string) and we should be able to both (!) escape and unescape it.
A brief explanation of how to make it work (mostly I quote how it is implemented for Rust):
1. Implement a lexer for TOML escape sequences (see [TOML escape codes](https://github.com/toml-lang/toml#string) ). Rust analogue - RustEscapesLexer. It can be used directly for highlighting (see RsHighlightingLexer). Also see tests RsEscapesLexingTestCase
2. Implement string escape/unescape. See RsEscapeUtils.kt. Also see tests RsEscapesUtilsTest.kt. This can also be used in spell checking. See RsSpellcheckingStrategy.
3. Implement LiteralTextEscaper. See RsNormalStringLiteralEscaper. This is not useful itself, so should be done in a couple with the next options.
4. Implement PsiLanguageInjectionHost interface for TomlLiteral. There should be used LiteralTextEscaper from the prev option and also updateText should be implemented with text escaping. See RsLitExprMixin (but as I see updateText implemented incorrectly there). At this point it's possible to provide language injections to TOML string literals (see RsRegExpInjector). But I don't think it's useful
5. Implement ElementManipulator for TOML literals. See RsStringLiteralManipulator (again, as I see it's implemented incorrectly b/c is doesn't escape the text). `getRangeInElement` method is the most necessary. Finally.
6. Also it can be useful (for all the above) to make something like RsLiteralKind for TOML. (offsets are the most useful)

Note that ElementManipulator and PsiLanguageInjectionHost (and so LiteralTextEscaper) implementation are needed to make RsFileReferenceProvider/FileReferenceSet work.

Please let me know it you want to work on it.
Also let me know it you want this PR to be merged (I'll just leave few comments). It's not ideal, but it's better than nothing. And thank you!

toml/src/main/kotlin/org/rust/toml/CargoTomlPathCompletion.kt
```diff
@@ -0,0 +1,71 @@
...
+            File -> if (file.isDirectory) DEFAULT_PRIORITY else 1.0
+            Directory -> if (file.isDirectory) 1.0 else DEFAULT_PRIORITY
+            Any -> DEFAULT_PRIORITY
+        }
+    }
+
+    override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
+        val keyValue = getClosestKeyValueAncestor(parameters.position) ?: return
+        val containingFile = keyValue.containingFile.originalFile.virtualFile ?: return
+        val path = keyValue.value?.text
```
vlad20012:
```suggestion
        val path = CompletionUtil.getOriginalElement(keyValue.value)?.text
```

After this change you will be able to remove `?.substringBefore("IntellijIdeaRulezzz")` :)

mibac138: Thank you

mibac138:
@vlad20012 After some consideration I'd prefer to have this PR merged and then do a follow-up PR with the changes you described. This way there will be at least some completion in paths and when I'll have more time and once I implement what you described I'll submit another PR enhancing this one.

Undin:
> After some consideration I'd prefer to have this PR merge

@mibac138 sorry but I don't think it's a good idea
Temporary things tend to become permanent.
It will take some of our time to review this PR. Including the fact that the current implementation is not a subset of the desired one and should be replaced in future, I think it's better not to merge this version at all

mibac138:
@Undin sure, I totally get that. I'll open a new PR once I implement this as `@vlad20012` described.

vlad20012:
@mibac138 It may not be useless to write somewhere ([gitter](https://gitter.im/intellij-rust/intellij-rust or open an issue)) about what you're going to work (if it's not trivial) to prevent such situations...

PR #3979, opened 2019-06-06
ANN: Detect E0267, E0268: break/continue used outside of loop

mibac138:
Related to #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -663,10 +663,12 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
 
     private fun checkBreakExpr(holder: AnnotationHolder, expr: RsBreakExpr) {
         checkLabelReferenceOwner(holder, expr)
+        checkLabelRefOwnerPlacementCorrectness(holder, expr)
     }
 
     private fun checkContExpr(holder: AnnotationHolder, expr: RsContExpr) {
         checkLabelReferenceOwner(holder, expr)
+        checkLabelRefOwnerPlacementCorrectness(holder, expr)
     }
 
     private fun checkLabelReferenceOwner(holder: AnnotationHolder, expr: RsLabelReferenceOwner) {
@@ -683,6 +685,23 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    // Detect E0267, E0268: break/continue used outside of loop
+    private fun checkLabelRefOwnerPlacementCorrectness(holder: AnnotationHolder, expr: RsLabelReferenceOwner) {
+        for (ancestor in expr.ancestors) {
+            // We are inside a loop, all is good
+            if (ancestor is RsLooplikeExpr) return
+            // let x = 'foo: { break 'foo: 1; }; is allowed (notice `'foo` label, without it would be invalid)
+            if (ancestor is RsBlockExpr && ancestor.labelDecl != null) return
+            // Reached the function definition - can't be in a loop
+            if (ancestor is RsFunction) break
+            if (ancestor is RsLambdaExpr) {
+                RsDiagnostic.LoopOnlyKeywordUsedInClosureError(expr.operator).addToHolder(holder)
+                return
+            }
+        }
+        // If we got here, we aren't inside a loop expr so emit an error
+        RsDiagnostic.LoopOnlyKeywordUsedOutsideOfLoopError(expr.operator).addToHolder(holder)
+    }
 
     private fun isInTraitImpl(o: RsVis): Boolean {
         val impl = o.parent?.parent?.parent
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsLooplikeExpr.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsLooplikeExpr.kt
@@ -1,0 +1,18 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.psi.RsBreakExpr
+import org.rust.lang.core.psi.RsContExpr
+
+/**
+ * An expression which can contain a [RsLabelReferenceOwner]. It does not however mean that a [RsLabelReferenceOwner]
+ * is invalid without it. E.g.: [RsContExpr] (`continue`) can only be valid inside a [RsLooplikeExpr], however a
+ * [RsBreakExpr] (`break`) can be valid outside of [RsLooplikeExpr] as well. In general, a [RsLabelReferenceOwner]
+ * is expected to be valid inside a [RsLooplikeExpr] if it doesn't have a label.
+ */
+interface RsLooplikeExpr : RsExpr, RsLabeledExpression, RsOuterAttributeOwner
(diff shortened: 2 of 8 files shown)


vlad20012:
Nice! Is it works with `label_break_value` feature?
```rust
#![feature(label_break_value)]

fn main() {
    let a = 'foo: {
        break 'foo 1;
    };
    println!("{}", a);
}
```

mibac138:
@vlad20012 it sure does now

src/main/kotlin/org/rust/lang/core/psi/ext/RsLoopOnlyExpr.kt
```diff
@@ -0,0 +1,14 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsExpr
+
+// Expr that can only be used inside a `loop` or `while` block. E.g. `break`, `continue`
+interface RsLoopOnlyExpr : RsExpr, RsLabelReferenceOwner, RsOuterAttributeOwner {
```
Undin: Name and comment are confusing because `break` can be used inside blocks with `#![feature(label_break_value)]` feature

src/main/kotlin/org/rust/lang/core/psi/ext/RsLoopOnlyExpr.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsExpr
+
+// Expr that can only be used inside a `loop` or `while` block. E.g. `break`, `continue`
```
Undin: Let's convert it into javadoc to have proper documentation

src/main/kotlin/org/rust/lang/core/psi/ext/RsLoopOnlyExpr.kt
```diff
@@ -0,0 +1,14 @@
...
+interface RsLoopOnlyExpr : RsExpr, RsLabelReferenceOwner, RsOuterAttributeOwner {
+    val element: PsiElement
```
Undin: It's quite difficult to understand that is this element without proper name and documentation

src/main/kotlin/org/rust/lang/core/psi/ext/RsLooplikeExpr.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsExpr
+
+// `loop`, `while` and `for`
```
Undin: Javadoc

src/main/kotlin/org/rust/lang/core/psi/ext/PsiElement.kt
```diff
@@ -245,3 +246,9 @@ fun PsiWhiteSpace.isMultiLine(): Boolean = getLineCount() > 1
 @Suppress("UNCHECKED_CAST")
 inline val <T: StubElement<*>> StubBasedPsiElement<T>.greenStub: T?
     get() = (this as? StubBasedPsiElementBase<T>)?.greenStub
+
+inline fun <reified T : PsiElement> PsiElement.parentOfType(): T? = parentOfType(T::class)
```
Undin: We already have `ancestorStrict` extension method

src/main/kotlin/org/rust/lang/core/psi/ext/RsLoopOnlyExpr.kt
mibac138: I agree with you that the name is somewhat misleading. I'm not sure what names should I use though. I have a couple ideas: `RsLoopNeedingExpr`, `RsLoopLikingExpr`, `RsLoopContainedExpr`. I'm leaning towards `RsLoopLikingExpr` as I believe it's the most truthful name but it doesn't really reflect the meaning. Do you have any suggestions? Alternatively, I could just remove this interface and make `RsErrorAnnotator` use type checks.

Undin: Actually, we already have interface for `break`/`continue` expressions: `RsLabelReferenceOwner`. Yeah, its name is only about labels but it was created to make common interface only for `break`/`continue`.
So let's reuse it.
About naming:
* I can come up with only `RsControlFlowExpr` as an interface name.
* As a better name for `element` property, I can suggest `operator` or something similar

src/main/kotlin/org/rust/lang/core/psi/ext/RsLooplikeExpr.kt
```diff
@@ -0,0 +1,14 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.psi.ext
+
+import org.rust.lang.core.psi.RsExpr
+
+/**
+ * An expression in which [RsLoopOnlyExpr] are valid. Currently the following expressions are [RsLooplikeExpr]: `loop`,
+ * `for`, and `while`.
```
Undin: It's not true because `break` expr can be used in blocks with `label_break_value` feature

src/main/kotlin/org/rust/lang/core/psi/ext/RsLoopOnlyExpr.kt
mibac138: Thanks, I have been completely oblivious of `RsLabelReferenceOwner`. I guess creating the `RsLoopOnlyExpr` was completely unnecessary. I have now removed it and replaced it with label reference owner. Additionally, I've added an `operator` property to it so that errors can be displayed on the operator only and not the whole expression. (`break` highlight instead of `break 'foo`).

src/main/kotlin/org/rust/lang/core/psi/ext/RsLooplikeExpr.kt
mibac138: Yeah, I meant to phrase that differently :/
I have now reworded and expanded the documentation comment.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -640,6 +642,20 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    // Detect E0267, E0268: break/continue used outside of loop
+    private fun checkParentIsLooplike(holder: AnnotationHolder, expr: RsLabelReferenceOwner) {
+        if (expr.ancestorStrict<RsLooplikeExpr>() == null) {
+            val containingBlock = expr.ancestorStrict<RsBlockExpr>()
+            // let x = 'foo: { break 'foo: 1; }; is allowed (notice `'foo` label, without it would be invalid)
+            if (containingBlock?.labelDecl != null) return
```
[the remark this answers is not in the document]
Undin: Also, true negative 
```rust
#![feature(label_break_value)]

fn main() {
    for i in 1..10 {
        let a = |x: i32| break;
        1
    };
}
```

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -640,6 +642,20 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    // Detect E0267, E0268: break/continue used outside of loop
+    private fun checkParentIsLooplike(holder: AnnotationHolder, expr: RsLabelReferenceOwner) {
```
Undin: Let's also rename method somehow because it's not about loop, it's about incorrect place of `break`/`continue` expressions. And at least `break` can be placed inside block expression

Undin:
Also, could you rebase onto master, please?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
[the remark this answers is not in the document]
mibac138: Thanks for discovering those :)

mibac138:
I've now rebased and pushed a fixed version.

Undin:
Great!
Thanks!
bors r+

PR #4001, opened 2019-06-10
COMP: Add completion in partial macro call arguments

ortem:
This PR adds new completion provider for incomplete macro calls. The provider suggests variants based on possible kinds (`expr`, `ty` and so on) of the fragment under the cursor. If there are several possible kinds, all suitable completions are suggested.

macrocall_comp

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -87,7 +87,7 @@
     elementType(".+BinExpr") = BinaryExpr
     elementType(".+BinOp") = BinaryOp
 
-    elementTypeFactory("ExprCodeFragment") = "org.rust.lang.core.psi.RsCodeFragmentElementTypeKt.factory"
+    elementTypeFactory("(Expr|Stmt|TypeRef)CodeFragment") = "org.rust.lang.core.psi.RsCodeFragmentElementTypeKt.factory"
 
     extends("Pat(Wild|Ref|Tup|Slice|Macro|Struct|TupleStruct|Ident|Range|Box|Const)") = Pat
 
@@ -1444,6 +1444,10 @@ private never ::= !()
 private meta comma_separated_list ::= <<param>> ( ',' <<param>> )* ','?
 private meta list_element ::= <<param>> !'+' (',' | &'>') { pin = 2 }
 
-// This parsing rule is used directly from the code
+// These parsing rules are used directly from the code
 //noinspection BnfUnusedRule
 ExprCodeFragment ::= Expr?
+//noinspection BnfUnusedRule
+StmtCodeFragment ::= Stmt?
+//noinspection BnfUnusedRule
+TypeRefCodeFragment ::= TypeReference?
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
@@ -7,17 +7,14 @@ package org.rust.lang.core.macros
 
 import com.intellij.lang.PsiBuilder
 import com.intellij.lang.PsiBuilderUtil
-import com.intellij.lang.parser.GeneratedParserUtilBase
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
-import com.intellij.psi.tree.TokenSet
 import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
 import com.intellij.util.SmartList
-import org.rust.lang.core.parser.RustParser
 import org.rust.lang.core.parser.RustParserUtil.collapsedTokenType
 import org.rust.lang.core.parser.createRustPsiBuilder
 import org.rust.lang.core.psi.*
@@ -30,7 +27,7 @@ import org.rust.stdext.mapNotNullToSet
 import java.util.*
 import java.util.Collections.singletonMap
 
-private data class MacroSubstitution(
+data class MacroSubstitution(
     /**
      * Maps meta variable names with the actual values, e.g. for this macro
      * ```rust
@@ -75,7 +72,7 @@ private data class MacroSubstitution(
     val groups: List<MacroGroup>
 )
 
-private data class MacroGroup(
+data class MacroGroup(
     val definition: RsMacroBindingGroup,
     val substs: List<MacroSubstitution>
 )
@@ -91,9 +88,9 @@ private data class WithParent(
         subst.variables[name] ?: parent?.getVar(name)
 }
 
-private data class MetaVarValue(
+data class MetaVarValue(
     val value: String,
-    val type: String,
+    val kind: FragmentKind?,
     val offsetInCallBody: Int
 )
 
@@ -106,7 +103,7 @@ class MacroExpander(val project: Project) {
             subst,
             WithParent(
                 MacroSubstitution(
-                    singletonMap("crate", MetaVarValue(expandDollarCrateVar(call, def), "", -1)),
+                    singletonMap("crate", MetaVarValue(expandDollarCrateVar(call, def), null, -1)),
                     emptyList()
                 ),
                 null
@@ -158,7 +155,7 @@ class MacroExpander(val project: Project) {
                     if (!substituteMacro(sb, ranges, child, subst)) return false
                 is RsMacroReference -> {
                     val value = subst.getVar(child.referenceName) ?: return false
-                    val parensNeeded = value.type == "expr"
+                    val parensNeeded = value.kind == FragmentKind.Expr
                     if (parensNeeded) {
                         sb.append("(")
                         sb.append(value.value)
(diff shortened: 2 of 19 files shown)


vlad20012:
// Note that (much more powerful) completion based on macro expansion will be added later, after #3640 landing. Then this provider will be used for not expanded macros (in incomplete code)

src/main/grammars/RustParser.bnf
```diff
@@ -1409,6 +1411,10 @@ private never ::= !()
 private meta comma_separated_list ::= <<param>> ( ',' <<param>> )* ','?
 private meta list_element ::= <<param>> !'+' (',' | &'>') { pin = 2 }
 
-// This parsing rule is used directly from the code
+// These parsing rules is used directly from the code
```
Undin:
```suggestion
// These parsing rules are used directly from the code
```

src/main/kotlin/org/rust/lang/core/completion/RsMacroArgumentCompletionProvider.kt
```diff
@@ -0,0 +1,75 @@
...
+
+            RsCommonCompletionProvider.processElement(reference, isSimplePath, null, hashSetOf(), parameters, result)
+
+            if (RsPrimitiveTypeCompletionProvider.elementPattern.accepts(reference.firstChild)) {
+                RsPrimitiveTypeCompletionProvider.addCompletionVariants(parameters, context, result)
+            }
+        }
+
+        if (Expr in kinds || Path in kinds) {
+            val fragment = RsExpressionCodeFragment(project, fragmentText, macroCall)
```
vlad20012: Let's inline these 3 `fragment` vals

src/main/kotlin/org/rust/lang/core/completion/RsMacroArgumentCompletionProvider.kt
```diff
@@ -0,0 +1,75 @@
...
+            val fragment = RsStatementCodeFragment(project, fragmentText, macroCall)
+            addCompletions(fragment)
+        }
+        if (Ty in kinds) {
+            val fragment = RsTypeReferenceCodeFragment(project, fragmentText, macroCall, TYPES)
+            addCompletions(fragment)
+        }
+    }
+
+    private fun RsCodeFragment.nearestReferenceElement(offset: Int): RsReferenceElement? =
```
vlad20012: `psi.findReferenceAt()` ??

src/main/kotlin/org/rust/lang/core/macros/MacroDiagram.kt
```diff
@@ -0,0 +1,115 @@
...
+            "pat" -> Pat
+            "stmt" -> Stmt
+            "block" -> Block
+            "item" -> Item
+            "meta" -> Meta
+            "vis" -> Vis
+            "tt" -> Tt
+            "lifetime" -> Lifetime
+            "literal" -> Literal
+            else -> error("Invalid fragment kind")
```
vlad20012: `s` comes here from user code, so it can be any string in general. Choose either return null or add specific `unknown` FragmentKind.

src/main/kotlin/org/rust/lang/core/macros/MacroDiagram.kt
```diff
@@ -0,0 +1,115 @@
...
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.impl.source.tree.LeafPsiElement
+import org.rust.lang.core.psi.RsMacro
+import org.rust.lang.core.psi.RsMacroBinding
+import org.rust.lang.core.psi.RsMacroBindingGroup
+import org.rust.lang.core.psi.ext.fragmentSpecifier
+
+enum class FragmentKind {
+    Ident,
```
vlad20012: Enum items should be in CAPS as far as I know...

src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -407,6 +354,61 @@ private class MacroPattern private constructor(
             generateSequence(firstChild) { it.nextSibling }
                 .filter { it !is PsiWhiteSpace && it !is PsiComment }
 
+        fun parse(builder: PsiBuilder, kind: FragmentKind): Boolean {
```
vlad20012: Consider moving it to a better place and maybe turning it into extension property of `FragmentKind`.

src/main/kotlin/org/rust/lang/core/resolve/Namespace.kt
```diff
@@ -44,6 +44,8 @@ val RsNamedElement.namespaces: Set<Namespace> get() = when (this) {
 
     is RsMacro -> MACROS
 
+    is RsTypeReferenceCodeFragment -> namespaces
```
vlad20012: Why is it needed?

src/main/kotlin/org/rust/lang/core/macros/MacroDiagram.kt
ortem: > For enum constants, it's OK to use either uppercase underscore-separated names (enum class Color { RED, GREEN }) or regular camel-humps names starting with an uppercase letter, depending on the usage.

https://kotlinlang.org/docs/reference/coding-conventions.html

src/main/kotlin/org/rust/lang/core/macros/MacroDiagram.kt
ortem: @vlad20012 or maybe use `Tt` for unknown kinds as the most general kind?

vlad20012: I think a pattern with unknown fragment specifier should never be matched successfully, so it looks like a bad idea. I'd prefer returning null.

ortem: OK, done

src/main/kotlin/org/rust/lang/core/macros/MacroDiagram.kt
```diff
@@ -0,0 +1,289 @@
...
+import org.rust.lang.core.psi.ext.fragmentSpecifier
+import java.util.*
+
+sealed class Matcher : Comparable<Matcher> {
+    data class Literal(val value: String) : Matcher() {
+        override fun compareTo(other: Matcher): Int =
+            if (other is Literal) {
+                this.value.compareTo(other.value)
+            } else {
+                this.ordinal.compareTo(other.ordinal)
```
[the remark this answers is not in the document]
vlad20012: Also consider using comparator:
```kotlin
object MatcherComparator: Comparator<Matcher> {
    override fun compare(m1: Matcher, m2: Matcher): Int = when {
        m1 is Matcher.Literal && m2 is Matcher.Literal -> m1.value.compareTo(m2.value)
        m1 is Matcher.Fragment && m2 is Matcher.Fragment -> m1.kind.compareTo(m2.kind)
        m1 is Matcher.Optional && m2 is Matcher.Optional -> compare(m1.matcher, m2.matcher)
        m1 is Matcher.Choice && m2 is Matcher.Choice -> compare(m1.matchers, m2.matchers)
        m1 is Matcher.Sequence && m2 is Matcher.Sequence -> compare(m1.matchers, m2.matchers)
        m1 is Matcher.Repeat && m2 is Matcher.Repeat -> compare(m1.matchers, m2.matchers) // TODO separator
        else -> m1.ordinal.compareTo(m2.ordinal)
    }

    private fun compare(list: List<Matcher>, other: List<Matcher>): Int = when {
        list.size > other.size -> 1
        list.size < other.size -> -1
        else -> list.zip(other).map { compare(it.first, it.second) }.firstOrNull { it != 0 } ?: 0
    }
}
```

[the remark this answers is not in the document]
ortem: Got rid of matcher folding for now.

src/main/kotlin/org/rust/lang/core/completion/RsMacroArgumentCompletionProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+
+            if (RsPrimitiveTypeCompletionProvider.elementPattern.accepts(referenceElement.firstChild)) {
+                RsPrimitiveTypeCompletionProvider.addCompletionVariants(parameters, context, result)
+            }
+        }
+
+        val project = parameters.originalFile.project
+        val position = parameters.position
+        val macroCall = position.ancestorStrict<RsMacroCall>() ?: return
+        val tokenTree = macroCall.macroArgument?.compactTT ?: return
```
vlad20012: After rebasing to the master here should be used `macroCall.bodyTextRange.startOffset`

src/main/kotlin/org/rust/lang/core/completion/RsMacroArgumentCompletionProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+            RsCommonCompletionProvider.processElement(referenceElement, isSimplePath, null, hashSetOf(), parameters, result)
+
+            if (RsPrimitiveTypeCompletionProvider.elementPattern.accepts(referenceElement.firstChild)) {
+                RsPrimitiveTypeCompletionProvider.addCompletionVariants(parameters, context, result)
+            }
+        }
+
+        val project = parameters.originalFile.project
+        val position = parameters.position
+        val macroCall = position.ancestorStrict<RsMacroCall>() ?: return
```
vlad20012:
```suggestion
        val macroCall = position.ancestorStrict<RsMacroArgument>()?.ancestorStrict<RsMacroCall>() ?: return
```
Use ti to ensure that we're inside a macro argument

src/main/kotlin/org/rust/lang/core/macros/MacroGraphWalker.kt
```diff
@@ -0,0 +1,125 @@
...
+        val descriptor: FragmentDescriptor?
+    )
+
+    private fun rollbackToState(state: State) {
+        position = state.position
+        state.marker?.rollbackTo()
+        descriptor = state.descriptor
+    }
+
+    private val builder = project.createRustPsiBuilder(callBody)
```
vlad20012: Run `builder.eof()` before the first usage of the builder. Will be needed after rebasing to the master

PR #4002, opened 2019-06-10
INSP: support of `Duplicated code fragment` inspection for Rust

Undin:
`Duplicated code fragment` inspection is not available in community IDEs like IDEA Community or PyCharm Community. It leads that the inspection support should be an optional dependency. That's why its implementation is located in a separate module. Also, now we use IDEA Ultimate as a dependency instead of IDEA Community


src/test/kotlin/org/rust/ide/inspections/RsInspectionsTestBase.kt
```diff
@@ -5,18 +5,29 @@
...
 package org.rust.ide.inspections
 
+import com.intellij.codeInspection.InspectionProfileEntry
+import com.intellij.testFramework.InspectionTestUtil
 import org.rust.TestProject
 import org.rust.ide.annotator.RsAnnotationTestBase
 
 abstract class RsInspectionsTestBase(
-    val inspection: RsLocalInspectionTool
+    private val inspectionClass: Class<out InspectionProfileEntry>
```
vlad20012: Consider using `KClass`.
And maybe extract it to another PR? 72 changed files looks scary 😨

Undin: Okay. I'll extract these changes into separate PR and merge it before this one if you think it's better

duplicates/src/192/main/kotlin/org/rust/ide/clones/RsDuplicateScope.kt
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * Used as a DuplicateScope id in settings serialization
+     */
+    override val languageName: String = "Rust"
+
+    /**
+     * @return true if [file] should be analyzed
+     */
+    override fun acceptsFile(project: Project, file: VirtualFile): Boolean =
+        ProjectFileIndex.getInstance(project).isInSourceContent(file)
```
vlad20012: let's check that the file is linked to a module structure, i.e. crate root exists.
Also maybe check that package origin is workspace instead of `isInSourceContent`?

Undin: Unfortunately, we can't get crate root here because it is used while indexing and we get `IndexNotReadyException` because `RsFile.crateRoot` relies on our indices, so we can get rather restricted info in this place

Undin: done

Undin:
bors r=vlad20012

wezm:
[The release notes for this feature](https://intellij-rust.github.io/2019/07/08/changelog-101.html) say

> Support Duplicated code fragment inspection for Rust gif. The inspection can detect duplicates even with different names of local variables, functions and fields, and values of literals. By default, only anonymization of local variable names is enabled. You may change default settings via Preferences | Editor | Duplicates | Rust. Note, this inspection is not available in free IDEs like IDEA Community or PyCharm Community

I have a paid up CLion subscription and have updated to `v0.2.101.2129-191` of the Rust plugin but I do not see a `Preferences | Editor | Duplicates | Rust` section in settings. Is that due to, "Also, now we use IDEA Ultimate as a dependency instead of IDEA Community", meaning it only works on Ultimate?

Undin:
> it only works on Ultimate?

@wezm Oh, no. It means that I forgot to add a note that the inspection is available since 2019.2. I've already updated release notes. Thanks!

wezm:
Great, I’ll eagerly await the stable release of 2019.2 :+1:

PR #4003, opened 2019-06-11
T: correctly instantiate inspection instance in tests

Undin:
Now, `RsInspectionsTestBase` takes inspection class instead of inspection instance. Turned out that inspection should have additional initialization sometimes. `com.intellij.testFramework.InspectionTestUtil#instantiateTools` does it

### DIFF
--- a/src/test/kotlin/org/rust/ide/annotator/fixes/AddAsTyFixTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/fixes/AddAsTyFixTest.kt
@@ -8,7 +8,7 @@ package org.rust.ide.annotator.fixes
 import org.rust.ide.inspections.RsInspectionsTestBase
 import org.rust.ide.inspections.RsTypeCheckInspection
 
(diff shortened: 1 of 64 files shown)


src/test/kotlin/org/rust/ide/annotator/fixes/AddAsTyFixTest.kt
```diff
@@ -8,7 +8,7 @@ package org.rust.ide.annotator.fixes
 import org.rust.ide.inspections.RsInspectionsTestBase
 import org.rust.ide.inspections.RsTypeCheckInspection
 
-class AddAsTyFixTest : RsInspectionsTestBase(RsTypeCheckInspection()) {
+class AddAsTyFixTest : RsInspectionsTestBase(RsTypeCheckInspection::class.java) {
```
vlad20012: Again, maybe it's better to use `RsTypeCheckInspection::class` (and so `KClass`) instead?

PR #4012, opened 2019-06-11
REF: improved rename for mod.rs and directories

### LINKED ISSUE #1859: Renaming directory module renames the `mod.rs`
Given a module `foo`, renaming foo in the `mod foo;` declaration will rename the `mod.rs` instead of the directory.

rrevenantt:
fixes #1859

also as side effect find usages for mod.rs modules started to work as well.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsDirectoryRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsDirectoryRenameProcessor.kt
@@ -1,0 +1,42 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.refactoring
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiDirectory
+import com.intellij.psi.PsiDirectoryContainer
+import com.intellij.psi.PsiElement
+import com.intellij.refactoring.RefactoringSettings
+import com.intellij.refactoring.rename.RenameDialog
+import com.intellij.refactoring.rename.RenamePsiFileProcessor
+import org.rust.lang.RsConstants
+
+class RsDirectoryRenameProcessor : RenamePsiFileProcessor() {
+
+    override fun createRenameDialog(project: Project, element: PsiElement, nameSuggestionContext: PsiElement?, editor: Editor?): RenameDialog {
+        return super.createRenameDialog(project, element.toDir(), nameSuggestionContext, editor)
+    }
+
+    override fun canProcessElement(element: PsiElement): Boolean {
+        return element is PsiDirectory || element is PsiDirectoryContainer
+    }
+
+    override fun prepareRenaming(element: PsiElement, newName: String, allRenames: MutableMap<PsiElement, String>) {
+        super.prepareRenaming(element, newName, allRenames)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/refactoring/RsDirectoryRenameProcessor.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        val modrs = element.toDir()
+            .run {
+                findFile(RsConstants.MOD_RS_FILE)
+                    ?: parentDirectory?.findFile("$name.rs")
+            } ?: return
+        allRenames[modrs] = newName
+    }
+
+    private fun PsiElement.toDir(): PsiDirectory = (this as? PsiDirectoryContainer)?.directories?.first()
```
mchernyavsky: Why are you using `first`?

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -59,6 +59,7 @@ class RsFile(
         get() = superMods.lastOrNull()?.takeIf { it.isCrateRoot }
 
     override fun setName(name: String): PsiElement {
+        if (this.name == RsConstants.MOD_RS_FILE) return this
```
mchernyavsky: How can I rename `mod.rs` now?

src/main/kotlin/org/rust/ide/refactoring/RsDirectoryRenameProcessor.kt
rrevenantt: Because Idea by default wraps every `PsiDirectory` in `PsiPackage`, and since intellij-rust itself does not use PsiPackage, `PsiPackage.directories` will contain exactly one `PsiDirectory`.

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
rrevenantt: Maybe this is not the best place to not rename `mod.rs` file, but anyway how I would know if user wants to rename file instead of module?

rrevenantt:
@mchernyavsky Done.

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
mchernyavsky: Ok. https://github.com/intellij-rust/intellij-rust/pull/4077 has a similar issue. I will think of a solution that doesn't use a mutable state.

mchernyavsky:
Thanks!
bors r+

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
rrevenantt: @mchernyavsky I am quite sure it is not related. #4077's problem is that we don't know which reference target to choose. Here, the problem is that we don't know users intention. No matter where user invokes module renaming and what are intentions, `RsRenameProcessor` end up dealing with mod.rs file. 

As for possible solutions, I have come up to two options:
- I think if user invokes rename action from the Project View we can safely assume that he wants to rename only mod.rs file. So looks like if we create custom `RenameHandler` only for mod.rs files, there is a way to know where action was invoked from. I have tried it and `LangDataKeys.IDE_VIEW` is exactly what we need.
- We can make module references resolve to `RsModDeclItem`, while keep corresponding file as navigation target. Thus files could be renamed independently. 

If mod.rs would still be the only solution for module creation, I personally would prefer second option. But since using mod.rs is currently an unfavorable way, I think the first option is just enough.

PR #4023, opened 2019-06-13
RES: fix duplicating definitions for `use` with bang proc macros

fedochet:
Clicking `Ctrl+B` on bang proc macro `use` item resulted in duplicating definitions like this: 

image

This is due to the fact that bang proc macro definitions are counted as macro imports (since #3455), and at the same time the definition itself is a function and seen as a function by resolve 

Proc macro crates cannot import anything other than proc macro definitions, so I've added extra for crate type to prevent dupliaction (but I may be wrong about the way to fix this)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -391,6 +391,13 @@ private fun processQualifiedPathResolveVariants(
             if (processAll(exportedMacros(base), processor)) return true
         }
     }
+
+    // Proc macro crates are not allowed to export anything but procedural macros,
+    // and all possible macro exports are collected above.
+    if (base.containingCargoTarget?.isProcMacro == true) {
+        return false
+    }
+
     if (parent is RsUseSpeck && path.path == null) {
         selfInGroup.hit()
         if (processor("self", base)) return true
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -390,12 +390,14 @@ private fun processQualifiedPathResolveVariants(
             containingMod is RsFile && containingMod.isCrateRoot) {
             if (processAll(exportedMacros(base), processor)) return true
         }
-    }
 
-    // Proc macro crates are not allowed to export anything but procedural macros,
-    // and all possible macro exports are collected above.
-    if (base.containingCargoTarget?.isProcMacro == true) {
-        return false
+        // Proc macro crates are not allowed to export anything but procedural macros,
+        // and all possible macro exports are collected above. However, when resolve
+        // happens inside proc macro crate itself, all items are allowed
+        val resolveBetweenDifferentTargets = base.containingCargoTarget != path.containingCargoTarget
+        if (resolveBetweenDifferentTargets && base.containingCargoTarget?.isProcMacro == true) {
+            return false
+        }
     }
 
     if (parent is RsUseSpeck && path.path == null) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -391,6 +391,13 @@ private fun processQualifiedPathResolveVariants(
             if (processAll(exportedMacros(base), processor)) return true
         }
     }
+
+    // Proc macro crates are not allowed to export anything but procedural macros,
+    // and all possible macro exports are collected above.
+    if (base.containingCargoTarget?.isProcMacro == true) {
```
vlad20012: Looks like we can move it into `if (base is RsMod)` block?

src/test/kotlin/org/rust/lang/core/resolve/RsPackageLibraryResolveTest.kt
```diff
@@ -557,7 +557,28 @@ class RsPackageLibraryResolveTest : RsResolveTestBase() {
...
     @MockEdition(CargoWorkspace.Edition.EDITION_2018)
-    fun `test resolve bang proc macro with use item`() = stubOnlyResolve("""
+    fun `test resolve bang proc definition in macro crate itself`() = stubOnlyResolve("""
+    //- dep-proc-macro/lib.rs
+        #[proc_macro]
+        pub fn example_proc_macro_1(item: TokenStream) -> TokenStream { item }
+
+        #[proc_macro]
+        pub fn example_proc_macro_2(item: TokenStream) -> TokenStream { example_proc_macro_1(item) }
+                                                                        //^ dep-proc-macro/lib.rs
```
vlad20012: Let's add a test with mod inside proc macro crate:
```rust
mod foo { pub fn bar() {} }
use foo::bar;
       //^
```
(looks like this test fails with your solution)

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
vlad20012: Also looks like you should allow a case when `base.containingCargoTarget == path.containingCargoTarget`

fedochet: Do you mean that when we are resolving use items from crate A to crate A we should be able to see all functions?

src/test/kotlin/org/rust/lang/core/resolve/RsPackageLibraryResolveTest.kt
fedochet: Done

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
fedochet: I've added extra check for it and additional test

vlad20012:
Thanks!
bors r+

Undin:
bors retry

src/test/kotlin/org/rust/lang/core/resolve/RsPackageLibraryResolveTest.kt
```diff
@@ -557,7 +557,51 @@ class RsPackageLibraryResolveTest : RsResolveTestBase() {
...
+    @MockEdition(CargoWorkspace.Edition.EDITION_2018)
+    fun `test proc macro crate cannot export anything but proc macros`() = expect<IllegalStateException> {
+        stubOnlyResolve("""
+        //- dep-proc-macro/lib.rs
+            #[proc_macro]
+            pub fn example_proc_macro(item: TokenStream) -> TokenStream { item }
+            pub fn unseen_function() {}
+        //- lib.rs
+            use dep_proc_macro::unseen_function;
+                                    //^ dep-proc-macro/lib.rs
```
Undin: Looks like we can use `//^ unresolved` comment here and get rid of `expect<IllegalStateException>`

fedochet: I am going to add more tests to that file in the next few PRs, so I can also fix it there

PR #4029, opened 2019-06-14
ANN: Fix E0118 false positive

### LINKED ISSUE #4017: False positive E0118 in stdlib
## Environment

* **Intellij-Rust plugin version:** 0.2.100.2656-192-nightly
* **Rust toolchain version:** rustc 1.37.0-nightly (02564de47 2019-06-10)
* **IDE name and version:** IU-192.4787.16

## Problem description
Primitive types can have free impls is stdlib. But the plugin always produces error annotation.
For example, open `libcore/slice/mod.rs` and find impl for `[T]`
image

The bug was introduced in #3825

mibac138:
Fixes #4017

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -373,8 +373,10 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         val typeRef = impl.typeReference ?: return
         if (typeRef.traitType != null) return
         val type = typeRef.type
-        if (type is TyPrimitive && impl.queryAttributes.langAttribute == type.name) {
-            return // `#[lang = "u8"] impl u8 {}` is a valid impl
+        if (impl.queryAttributes.langAttribute != null) {
+            // There are some special rules for #[lang] items, see:
+            // https://doc.rust-lang.org/unstable-book/language-features/lang-items.html)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -373,8 +373,13 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         val typeRef = impl.typeReference ?: return
         if (typeRef.traitType != null) return
         val type = typeRef.type
-        if (type is TyPrimitive && impl.queryAttributes.langAttribute == type.name) {
-            return // `#[lang = "u8"] impl u8 {}` is a valid impl
+        val langAttr = impl.queryAttributes.langAttribute
+        if ((type is TyPrimitive && langAttr == type.name) ||
+            (type is TyPointer && (langAttr == "const_ptr" || langAttr == "mut_ptr") ||
+                (type is TySlice && langAttr?.startsWith("slice") == true))) {
```
Undin: Hm, this condition looks quite complicated
Maybe we can simplify it and check only existence of `lang` attribute? It's a bit less precise but greatly simplify logic and hardly work worse in the real world

mibac138: @Undin sure thing, will update later today

Undin:
bors retry

PR #4042, opened 2019-06-18
Fix synchronous execution on EDT in Rustfmt

ortem:


src/main/kotlin/org/rust/cargo/RustfmtSavingVetoer.kt
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Checks rustfmt is installed when `Run rustfmt on save` is enabled.
+ * Always permits to save the document and only shows the warning balloon (in [Rustup.checkNeedInstallRustfmt]).
+ */
+class RustfmtSavingVetoer : FileDocumentSynchronizationVetoer() {
+    override fun maySaveDocument(document: Document, isSaveExplicit: Boolean): Boolean {
+        val file = document.virtualFile ?: return true
+        val project = ProjectLocator.getInstance().guessProjectForFile(file) ?: return true
+        val cargoProject = project.cargoProjects.findProjectForFile(file) ?: return true
```
vlad20012: Maybe make public and move to some utility place  `findCargoProject` method and use it here?

src/main/kotlin/org/rust/cargo/RustfmtWatcher.kt
```diff
@@ -85,8 +84,13 @@ class RustfmtWatcher : BaseComponent {
             val project = cargoProject.project
             if (!project.rustSettings.runRustfmtOnSave) return
             val rustfmt = project.toolchain?.rustfmt() ?: return
-            if (checkNeedInstallRustfmt(cargoProject.project, cargoProject.workingDirectory)) return
-            documents.forEach { rustfmt.reformatDocument(cargoProject, it) }
+
+            ApplicationManager.getApplication().invokeLater {
+                for (document in documents) {
+                    rustfmt.reformatDocument(cargoProject, document)
+                    FileDocumentManager.getInstance().saveDocument(document)
```
vlad20012: With this solution the document will be saved twice. I don't sure it's a good idea

vlad20012:
Will be fixed differently on 2019.3

Issue #3995, opened 2019-06-09
Create new module do nothing if no corresponding directory exists


sirgl:
## Environment

* **Intellij-Rust plugin version:** v0.2.99.2126-191
* **Rust toolchain version:** 1.18.3
* **IDE name and version:** 2019.1.4
* **Operating system:** mac os

## Problem description

Create module from usage intention do nothing, if no folder exists.

Example:
![image](https://user-images.githubusercontent.com/11944514/59155305-96c6d100-8ab0-11e9-95d5-d33f710d9eb2.png)
Here intention silently fails.

PR #4056, opened 2019-06-21
INT: Fix creation of new module if no corresponding directory exists

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/3995.


Undin:
bors retry

PR #4059, opened 2019-06-22
ANN: Detect E0586: inclusive ranges must be bounded at the end

mibac138:
Related: #886. Implements checking for [E0586](https://doc.rust-lang.org/error-index.html#E0586).
I modified the grammar to accept inclusive range expressions without end expression, such as `1..=` and `5...`. This change has been made so that the parser doesn't create somewhat unhelpful errors and they can instead be created inside the error annotator. 
Additionally, using `...` ranges now creates an error. I have not found since when does this create an error or if the error can be disabled with a feature so I've made it so that it just always errors when encountering them.


mibac138:
@mchernyavsky Forgot about that, thanks. Fixed now.

PR #4060, opened 2019-06-23
COMP&MACRO: advanced completion inside macro call bodies

vlad20012:
If a macro is successfully expanded we follow element under the caret to the macro expansion and invoke completion there. It allows performing completion taking into account the context of expansion. If a macro isn't expanded (e.g. there is a very incomplete code) completion from #4001 will be used.

image


Undin:
bors retry

PR #4064, opened 2019-06-26
ANN: add "Create field" quick fix for "No such field" error

### LINKED ISSUE #2363: Inspections to create fields on initialize
For example:
```rust
struct S {}
impl S {
    fn new(foo: String) -> S {
        S {
            foo/*caret*/
        }
    }
}
```
after

```rust
struct S {
    foo: String
}
impl S {
    fn new(foo: String) -> S {
        S {
            foo
        }
    }
}
```

t-kameyama:
Fixes #2363.

demo

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/CreateStructFieldFromConstructorFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/CreateStructFieldFromConstructorFix.kt
@@ -1,0 +1,91 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsVisibility
+import org.rust.lang.core.psi.ext.parentStructLiteral
+import org.rust.lang.core.psi.ext.resolveToBinding
+import org.rust.lang.core.types.infer.TypeVisitor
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/CreateStructFieldFromConstructorFix.kt
```diff
@@ -0,0 +1,82 @@
...
+                if (struct.identifier == null) return null
+            }
+            return CreateStructFieldFromConstructorFix(field)
+        }
+
+        private fun RsStructLiteralField.structItem(): RsStructItem? {
+            return parentStructLiteral.path.reference.resolve() as? RsStructItem
+        }
+
+        private fun RsStructLiteralField.nameAndType(): Pair<String, RsTypeReference>? {
```
mchernyavsky: I think we should use `Ty` instead of `RsTypeReference`. This will help us support a more complex case, like this:

```rust
fn foo((x, y): (i32, i32)) {
    S { x }
}
```
You can simply get the field's type:

```rust
private fun RsStructLiteralField.inferFieldTy(): Ty {
    val type = if (colon == null) resolveToBinding()?.type else expr?.type
    return type ?: TyUnknown
}
```

And then check if we can use it in the context of struct item:

```rust
fun canUse(ty: Ty): Boolean {
    val result = ty.visitWith(object : TypeVisitor {
        override fun visitTy(ty: Ty): Boolean = when (ty) {
            is TyUnknown, is TyTypeParameter, is TyAnon, is TyInfer, is TyProjection -> true
            else -> ty.superVisitWith(this)
        }

        override fun visitRegion(region: Region): Boolean = region !is ReStatic
    })
    return !result
}
```

This function returns `false` in these cases:

```rust
struct St {}
trait Tr {}

fn foo<T>(x: T, t: impl Tr, z: &str) {
    St { x, y, z }
}
```

p.s. You can use `insertionSafeTextWithLifetimes` in `createBlockFields` to render `Ty`.

mchernyavsky:
@t-kameyama could you please squash your commits?

PR #4067, opened 2019-06-27
Fix special grammar rules for macros

### LINKED ISSUE #4066: Syntax error while parsing macro calls
## Environment
* **Intellij-Rust plugin version:** 0.2.101.2671-192-nightly
* **Rust toolchain version:** rustc 1.35.0 (3c235d560 2019-05-20)
* **IDE name and version:** IU-192.5281.24

## Problem description
If macro call name in a list of names with a special grammar rule but its argument cannot be parsed by this rule, we get syntax error instead of fallback to common parsing rule.
image
The reason: `pin` attribute in such rules forces parser make a decision right after `!` token. We solved similar [issue](https://github.com/intellij-rust/intellij-rust/issues/738) for log macros.

A real-world example can be found in `src/vector.rs` of [mint](https://crates.io/crates/mint) crate. Part of the problem described in #3396
## Steps to reproduce
```rust
vec!(Foo[]);
```

Undin:
These changes drop `pin` attribute from special macro rules because it prevents parser fallback to common parsing rules

Also, type inference for macro expressions was adjusted to take into account special macro call arguments

Fixes #4066


vlad20012:
:+1 bors r+

vlad20012:
:+1: 
bors r+

PR #4070, opened 2019-06-28
ANN: improve autoimport with UFCS & associated items

vlad20012:


src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -87,8 +88,9 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
 
         fun findApplicableContext(project: Project, path: RsPath): Context<RsPath>? {
             val basePath = path.basePath()
-            if (TyPrimitive.fromPath(basePath) != null) return null
-            if (basePath.reference.multiResolve().isNotEmpty()) return null
+            if (TyPrimitive.fromPath(basePath) != null || basePath.reference.multiResolve().isNotEmpty()) {
+                return findApplicableContextForAssocItemPath(path, project)
```
Undin: Let's add some comment why we want to import if it can be resolved
It's not obvious that we resolve some incorrect (in compiler sense) paths

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixStdTest.kt
```diff
@@ -449,4 +449,23 @@ class AutoImportFixStdTest : AutoImportFixTestBase() {
             let a = HashMap::new/*caret*/();
         }
     """)
+
+    // TODO should be imported from std
```
Undin: After rebasing on master it should work as expected

src/main/kotlin/org/rust/lang/core/types/infer/MemoryCategorization.kt
```diff
@@ -260,7 +260,7 @@ class MemoryCategorizationContext(val lookup: ImplLookup, val inference: RsInfer
     private fun processPathExpr(pathExpr: RsPathExpr): Cmt {
         val type = inference.getExprType(pathExpr)
         // TODO: infcx.getResolvedPaths
```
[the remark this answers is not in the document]
ortem: No, it really should be dropped now.

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -87,8 +88,18 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
...
+
+            val isBasePathResolved = TyPrimitive.fromPath(basePath) != null ||
+                basePath.reference.multiResolve().isNotEmpty()
+
+            if (isBasePathResolved) {
+                // Despite the fact that path is (multi)resolved by our resolve engine, it can be unresolved from
+                // the view of the rust compiler. Specifically we resolve associated items even if corresponding
+                // trait is not in the scope, so here we suggest importing such traits
+
+                if (basePath == path) return null
```
Undin: It doesn't work with  `<i32>::foo` case

PR #4072, opened 2019-06-29
ANN: fix autoimport for type-related paths

vlad20012:
Also introduce `RsPathCodeFragment`.

There was an issue in `RsCodeFragmentFactory.createPath`, so you
wasn't able to select namespaces (types/value) in which path should
be resolved. In most cases in was resolved in the both namespaces,
so it mostly worked as expected, but sometimes it was resolved in
VALUES namespace only (see the test). I made `RsPathCodeFragment`
to precisely select allowed namespaces for generated paths.
Now during autoimport we always use both types&values namespaces
(but it also can be an issue if there are 2 defs with the same name
and in the same module but in different namespaces; I'm going to fix
it later).


vlad20012:
I do not know what I smoked yesterday, but it seems the test passes on the current master without any changes

PR #4077, opened 2019-06-29
REF: implement correct renaming of init value in shorthand struct literal

### LINKED ISSUE #3965: Invalid struct literal after init value rename
## Environment

v0.2.98.2125-191
* **Rust toolchain version:**
* **IDE name and version:**
PyCharm 2019.1.2 (Community Edition)
Build #PC-191.7141.48, built on May 7, 2019
JRE: 11.0.2+9-b159.56 x86_64
* **Operating system:**
macOS 10.14.4

## Problem description
When we create struct literal with named fields we can use short syntax when field name is the same as field init variable, i.e
```rust
struct Foo {
    bar: u32
}

let bar = 1;
let foo = Foo {bar};
```
However, if we rename local `bar` to say `baz`, Foo init will not be affected: `Foo{baz}` instead of `Foo{bar: baz}`

fan-tom:
Fixes #3965

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
@@ -39,7 +39,7 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
                 }
                 .forEach {
                     val newPatField = RsPsiFactory(element.project)
-                        .createPatFieldFull(element.text, newName)
+                        .createStructLiteralField(element.text, newName)
                     it.element!!.replace(newPatField)
                 }
         }
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
@@ -31,6 +31,19 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
     override fun canProcessElement(element: PsiElement): Boolean = element is RsNamedElement
 
     override fun renameElement(element: PsiElement, newName: String, usages: Array<out UsageInfo>, listener: RefactoringElementListener?) {
+        if (element is RsPatBinding) {
+            usages
+                .filter {
+                    val usageElement = it.element
+                    usageElement is RsStructLiteralField && usageElement.colon == null
+                }
+                .forEach {
+                    val newPatField = RsPsiFactory(element.project)
+                        .createPatFieldFull(element.text, newName)
+                    it.element!!.replace(newPatField)
+                }
+        }
+
         val newRenameElement = if (element is RsPatBinding && element.parent.parent is RsPatStruct) {
             val newPatField = RsPsiFactory(element.project)
                 .createPatFieldFull(element.identifier.text, element.text)
--- a/src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
@@ -112,6 +112,24 @@ class RenameTest : RsTestBase() {
     """)
 
 
+    fun `test field initialization shorthand`() = doTest("spam", """
+        struct S { foo: i32 }
+
+        fn main() {
+            let /*caret*/foo = 92;
+            let x = S { foo };
+            println!("{}", x.foo);
+        }
+    """, """
+        struct S { foo: i32 }
+
+        fn main() {
+            let spam = 92;
+            let x = S { foo: spam };
+            println!("{}", x.foo);
+        }
+    """)
+
     fun `test rename lifetime`() = doTest("'bar", """
         fn foo<'foo>(a: &/*caret*/'foo u32) {}
     """, """
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -134,6 +134,12 @@ class RsPsiFactory(
     fun createStructLiteral(name: String): RsStructLiteral =
         createExpressionOfType("$name { }")
 
+    fun createStructLiteralField(name: String, value: String): RsStructLiteralField {
+        return createExpressionOfType<RsStructLiteral>("S { $name: $value }")
+            .structLiteralBody
+            .structLiteralFieldList[0]
+    }
+
     fun createStructLiteralField(name: String, value: RsExpr? = null): RsStructLiteralField {
         val structLiteralField = createExpressionOfType<RsStructLiteral>("S { $name: () }")
             .structLiteralBody


src/main/kotlin/org/rust/lang/core/resolve/ref/RsStructExprFieldReferenceImpl.kt
```diff
@@ -36,12 +38,25 @@ class RsStructLiteralFieldReferenceImpl(
             val psiFactory = RsPsiFactory(element.project)
             val newIdent = psiFactory.createIdentifier(newName)
             val colon = psiFactory.createColon()
-            val initExpression = psiFactory.createExpression(identifier.text)
 
-            identifier.replace(newIdent)
-            element.add(colon)
-            element.add(initExpression)
+            when(renameSourceType) {
```
Undin:
```suggestion
            when (renameSourceType) {
```

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -44,4 +46,42 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
+                ref.rangeInElement.startOffset,
+                ref.rangeInElement.endOffset,
+                element,
+                ref.resolve() == null && ref.multiResolve(true).isEmpty(),
+                renameSourceType
+            )
+        }
+    }
+
+    class RenameSourceTypeTrackingUsageInfo(element: PsiElement,
```
Undin: [formatting](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting)

src/main/kotlin/org/rust/lang/core/resolve/ref/RsStructExprFieldReferenceImpl.kt
```diff
@@ -18,6 +18,8 @@ class RsStructLiteralFieldReferenceImpl(
 ) : RsReferenceCached<RsStructLiteralField>(field),
     RsReference {
 
+    var renameSourceType: RenameSourceType? = null
```
Undin: Mutable state in reference looks rather suspiciously

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -44,4 +46,42 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
+                else -> null
+            }
+
+            return RenameSourceTypeTrackingUsageInfo(
+                referenceElement,
+                ref,
+                ref.rangeInElement.startOffset,
+                ref.rangeInElement.endOffset,
+                element,
+                ref.resolve() == null && ref.multiResolve(true).isEmpty(),
```
Undin: Looks like it's enough to call only `ref.multiResolve(true).isEmpty()` here because if `ref.multiResolve(true).isEmpty()` then `ref.resolve()` is always null
Also, I'd prefer to extract this calculation into separate local variable

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -44,4 +46,42 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
+                is RsPatBinding -> RenameSourceType.PatBinding
+                is RsNamedFieldDecl -> RenameSourceType.Field
+                else -> null
+            }
+
+            return RenameSourceTypeTrackingUsageInfo(
+                referenceElement,
+                ref,
+                ref.rangeInElement.startOffset,
+                ref.rangeInElement.endOffset,
```
Undin: Why do you need these params?
Looks like it's enough to pass only `ref`

fan-tom: I just pass them to `MoveRenameUsageInfo` constructor the same way superclass `RenamePsiElementProcessor` does it. If I don't provide offsets, they will be set to -1, I don't know whether it is ok

src/main/kotlin/org/rust/lang/core/resolve/ref/RsStructExprFieldReferenceImpl.kt
fan-tom: If I can I would add it as constructor argument, but I cannot pass that value where `RsStructLiteralFieldReferenceImpl` object is created (`org.rust.lang.core.psi.ext.RsStructLiteralFieldImplMixin#getReference`), that's why I considered to add that mutable property

src/main/kotlin/org/rust/lang/core/resolve/ref/RsStructExprFieldReferenceImpl.kt
fan-tom: +

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
fan-tom: ok

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
fan-tom: ok

fan-tom:
Are Travis tests broken? I created that PR and all tests succeeded. Then I have pushed changes to my PR and one of builds fails (https://travis-ci.org/intellij-rust/intellij-rust/builds/552369475). Then I pushed empty --amend commit and two builds fail (https://travis-ci.org/intellij-rust/intellij-rust/builds/552460614). Are these fails related to my code changes or they are broken CI process' consequences?

src/main/kotlin/org/rust/lang/core/resolve/ref/RsStructExprFieldReferenceImpl.kt
fan-tom: Can you suggest how can I pass info about rename source (field or binding) to handleRename without carrying it in mutable property?

src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
```diff
@@ -83,6 +83,24 @@ class RenameTest : RsTestBase() {
         }
     """)
 
+    fun `test field initialization shorthand`() = doTest("spam", """
```
mchernyavsky: Hmm, looks like it doesn't work with inplace renaming. Try the test in the IDE.

fan-tom: Oh, yes, you are right. Seems that handleElementRename is not called on inplace renaming. What then I should do to handle this?

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
fan-tom: Ah, I understand, I'll fix

src/main/kotlin/org/rust/lang/core/resolve/ref/RsStructExprFieldReferenceImpl.kt
rrevenantt: I would suggest moving all rename logic from `RsStructLiteralFieldReferenceImpl.handleElementRename()` to `RsRenameProcessor.renameElement()`, where it is possible to get access to rename source and then rename `RsStructLiteralField`s accordingly.

src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
rrevenantt: I have fixed this issue during #4127. It has been merged recently so rebasing to latest master should solve it.

fan-tom:
Pushed new version, moved all logic to `RsRenameProcessor.renameElement()` as @rrevenantt suggested

src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
fan-tom: Yes, rebase fixed that issue

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -31,6 +31,19 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
     override fun renameElement(element: PsiElement, newName: String, usages: Array<out UsageInfo>, listener: RefactoringElementListener?) {
+        if (element is RsPatBinding) {
+            usages
+                .filter {
+                    val usageElement = it.element
+                    usageElement is RsStructLiteralField && usageElement.colon == null
+                }
+                .forEach {
+                    val newPatField = RsPsiFactory(element.project)
+                        .createPatFieldFull(element.text, newName)
```
mchernyavsky: You have to replace the `element` with a new `RsStructLiteralField`, not a `RsPatFieldFull`.

fan-tom: Fixed

PR #4082, opened 2019-06-30
PERF:make lexer incremental

rrevenantt:
Make RustLexer support restarting to utilize platform's incremental re-lexing.

@vlad20012 after #3865 I have checked and you were right - our lexer was actually not incremental. So I have fixed it

Here is execution times for highlighting pass on ~200kb file before and after my changes:
image

### DIFF
--- a/src/main/grammars/RustLexer.flex
+++ b/src/main/grammars/RustLexer.flex
@@ -36,7 +36,7 @@ import static com.intellij.psi.TokenType.*;
 %{
   IElementType imbueBlockComment() {
       assert(zzNestedCommentLevel == 0);
-      yybegin(INITIAL);
+      yybegin(YYINITIAL);
 
       zzStartRead = zzPostponedMarkedPos;
       zzPostponedMarkedPos = -1;
@@ -53,7 +53,7 @@ import static com.intellij.psi.TokenType.*;
   }
 
   IElementType imbueRawLiteral() {
-      yybegin(INITIAL);
+      yybegin(YYINITIAL);
 
       zzStartRead = zzPostponedMarkedPos;
       zzShaStride = -1;
@@ -63,7 +63,7 @@ import static com.intellij.psi.TokenType.*;
   }
 
   IElementType imbueOuterEolComment(){
-      yybegin(INITIAL);
+      yybegin(YYINITIAL);
 
       zzStartRead = zzPostponedMarkedPos;
       zzPostponedMarkedPos = -1;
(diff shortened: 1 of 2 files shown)


vlad20012:
Wow, cool! Thank you!

Unfortunately, I'm not very good at lexers. @mchernyavsky would be a suitable reviewer.

Also, may I ask how it restores a state of mutable fields of the lexer (except the "state"), e.g. `zzNestedCommentLevel`?

rrevenantt:
It retreats until initial `YYINITIAL` state and restarts from that point. AFAIK mutable fields of the lexer do not matter for `YYINITIAL` state.

src/test/kotlin/org/rust/lang/core/lexer/RsRestartLexingTestCase.kt
```diff
@@ -0,0 +1,41 @@
...
+        editor = createEditor("copy.rs", editor.document.text)
+        val fullLexer = editor.createLexer(0)!!
+
+        while (!restartedLexer.atEnd() && !fullLexer.atEnd()) {
+            assertEquals(restartedLexer.tokenType, fullLexer.tokenType)
+            assertEquals(restartedLexer.start, fullLexer.start)
+            restartedLexer.advance()
+            fullLexer.advance()
+        }
+        assert(restartedLexer.atEnd() && fullLexer.atEnd())
```
Undin: Let's better use `assertTrue` or `assertEquals` instead of `assert`

src/main/grammars/RustLexer.flex
```diff
@@ -267,9 +265,18 @@ EOL_DOC_LINE  = {LINE_WS}*!(!("///".*)|("////".*))
                                     zzPostponedMarkedPos = zzStartRead;
                                     zzShaStride          = yylength() - 2; }
 
+//nessesary to allow proper restart of lexer when there is an error in raw literal, see RsRestartLexingTestCase
```
mchernyavsky: Hmm, it looks like the tests from the `RsRestartLexingTestCase` pass even if you comment out this code.

rrevenantt: Weird... for me second test does successfully fail. Are you sure you updated generated lexer? Because running tests from gutter icon does not automatically update it.

mchernyavsky: Yes, I tried `Generate parser` configuration and `:verification:test` gradle task.
@vlad20012 can you check these tests?

vlad20012: For me it works as @rrevenantt expects: `test lexer restart in invalid raw literal` fails if comment these lines

mchernyavsky:
LGTM

PR #4083, opened 2019-06-30
ANN: allow to autoimport traits in UFCS calls

vlad20012:


src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixTest.kt
```diff
@@ -549,7 +549,8 @@ class AutoImportFixTest : AutoImportFixTestBase() {
         }
     """)
 
-    fun `test don't import trait method`() = checkAutoImportFixIsUnavailable("""
+    fun `test don't import trait assoc function if its import is useless`() = expect<IllegalStateException> {
```
vlad20012: Initially, I want to disallow such imports by this check:
```kotlin
/**
 * If function is defined in a trait
 * ```rust
 * trait Trait {
 *     fn foo() {}
 * }
 * ```
 * it potentially can be accessed by the trait name `Trait::foo` only if there are `self` parameter or
 * `Self` type in the function signature
 */
val RsFunction.canBeAccessedByTraitName: Boolean
    get() {
        check(owner is RsAbstractableOwner.Trait)
        return selfParameter != null || typeOfValue.visitWith(object : TypeVisitor {
            override fun visitTy(ty: Ty): Boolean =
                if (ty is TyTypeParameter && ty.parameter is TyTypeParameter.Self) true else ty.superVisitWith(this)
        })
    }
```

but looks like this solution doesn't work with default methods accessed by struct name (I can't be sure without #4070) and the case we want forbid is very rare (why use trait functions without `Self` types and why?) so I this we can allow it for now

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -325,7 +325,11 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
                 .createPathInTmpMod(importingPathName, context.mod, context.ns, info.usePath, externCrateName) ?: return false
             val element = path.reference.deepResolve() as? RsQualifiedNamedElement ?: return false
             if (!context.namespaceFilter(element)) return false
-            return !(element.parent is RsMembers && element.ancestorStrict<RsTraitItem>() != null)
+            // Looks like it's useless to access trait associated types or constants directly
+            // (i.e. `Trait::CONSTANT`), but associated functions can be used in UFCS
+            val isTraitAssocTypeOrConstant =
```
Undin: Hm, it's interesting
There is at least one example when I can use assoc constant with `Trait::CONSTANT` syntax
```rust
struct Foo;

impl Bar for Foo {
    const C: Self = Foo;
}

pub trait Bar {
    const C: Self;
}

fn main() {
    let f: Foo = Bar::C;
}
```

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -325,7 +325,11 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
                 .createPathInTmpMod(importingPathName, context.mod, context.ns, info.usePath, externCrateName) ?: return false
             val element = path.reference.deepResolve() as? RsQualifiedNamedElement ?: return false
             if (!context.namespaceFilter(element)) return false
-            return !(element.parent is RsMembers && element.ancestorStrict<RsTraitItem>() != null)
+            // Looks like it's useless to access trait associated types or constants directly
+            // (i.e. `Trait::CONSTANT`), but associated functions can be used in UFCS
```
Undin:
```suggestion
            // (i.e. `Trait::CONSTANT`), but associated functions and methods can be used in UFCS
```

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixTest.kt
Undin: > I can't be sure without #4070

You can check it now

Undin:
@vlad20012 ping

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
vlad20012: Wow, nice case. Fixed

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixTest.kt
vlad20012: > You can check it now

Checked, and just rewrote it to work with default methods

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -331,7 +336,12 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
             ) ?: return false
             val element = path.reference.deepResolve() as? RsQualifiedNamedElement ?: return false
             if (!context.namespaceFilter(element)) return false
-            return !(element.parent is RsMembers && element.ancestorStrict<RsTraitItem>() != null)
+
+            // Looks like it's useless to access trait associated types directly (i.e. `Trait::Type`),
+            // but methods can be used in UFCS and associated functions or constants can be accessed
+            // it they have `Self` type in a signature
+            return element !is RsAbstractable || element.owner !is RsAbstractableOwner.Trait ||
+                element.canBeAccessedByTraitName || path.qualifier?.reference?.deepResolve() !is RsTraitItem
```
Undin: What do you think about splitting out this condition into several ones (for example, `element !is RsAbstractable || element.owner !is RsAbstractableOwner.Trait`, `element.canBeAccessedByTraitName` and `path.qualifier?.reference?.deepResolve() !is RsTraitItem`)?
It should simplify debugging and separate light logic checks (`element !is RsAbstractable || element.owner !is RsAbstractableOwner.Trait`) from heavy ones: type inference (`element.canBeAccessedByTraitName`) and name resolution (`path.qualifier?.reference?.deepResolve() !is RsTraitItem`). Of course, the last reason is just a matter of taste

PR #4086, opened 2019-07-01
DOC: docs for `findExpansionElements()`

vlad20012:


src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -95,6 +95,36 @@ private fun mapOffsetFromExpansionToCallBody(call: RsMacroCall, offset: Int): In
         ?.fromBodyRelativeOffset(call)
 }
 
+/**
+ * If [this] element is inside a macro call body and this macro is successfully expanded, returns
+ * a leaf element inside the macro expansion that is expanded from [this] element. Returns a
+ * list of elements because an element inside a macro call body can be placed in a macro expansion
+ * multiple times. Returns null if [this] element is not inside a macro call body, or the macro
+ * expanded erroneously.
```
Undin: Maybe "or the macro expansion failed" is better?

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -95,6 +95,36 @@ private fun mapOffsetFromExpansionToCallBody(call: RsMacroCall, offset: Int): In
...
+ *
+ * # Examples
+ *
+ * Return an empty list if [this] element is not placed to an expansion:
```
Undin:
```suggestion
 * Returns an empty list if [this] element is not placed to an expansion:
```

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -95,6 +95,36 @@ private fun mapOffsetFromExpansionToCallBody(call: RsMacroCall, offset: Int): In
...
+ *
+ * ```rust
+ * macro_rules foo { (bar) => { fn baz(){} } }
+ * foo!(bar); // This `bar` is matched with the `bar` in the pattern and is not placed to the expansion
+ * ```
+ *
+ * Return a list of single element if [this] element is placed into the expansion single time:
```
Undin:
```suggestion
 * Returns a single-element list if [this] element is placed into the expansion only once:
```

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -95,6 +95,36 @@ private fun mapOffsetFromExpansionToCallBody(call: RsMacroCall, offset: Int): In
...
+ *
+ * ```rust
+ * macro_rules foo { ($i:ident) => { fn $i(){} } }
+ * foo!(bar); // This `bar` is placed to the expansion as a function name
+ * ```
+ *
+ * Return a list of multiple elements if [this] element is placed into the expansion multiple times:
```
Undin:
```suggestion
 * Returns a list of multiple elements if [this] element is placed into the expansion multiple times:
```

vlad20012:
Fixed

PR #4095, opened 2019-07-04
REF: WIP fix field usages on destructuring intention performed

### LINKED ISSUE #3684: No usage fix on destructuring declaration
## Environment

v0.2.96.2122-191
* **IDE name and version:**
PyCharm 2019.1.1 (Community Edition)
Build #PC-191.6605.12, built on April 3, 2019
JRE: 11.0.2+9-b159.34 x86_64
macOS 10.14.4

## Problem description
When we have a function with tuple struct as argument
изображение

and use `Use destructuring declaration` intent plugin doesn't rename that argument usages
изображение

fan-tom:
WIP, fixes #3684

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/DestructureIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/DestructureIntention.kt
@@ -8,12 +8,15 @@ package org.rust.ide.intentions
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiReference
+import com.intellij.psi.SmartPsiElementPointer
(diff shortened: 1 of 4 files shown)


fan-tom:
Currently it correctly works only on structs with named fields as usages rename is strightforward. But with tuples and tuple structs require user to input names of introduced bindings. It is performed by TemplateBuilder, however, it has no ability to rename usages. How can I start renaming of multiple just-introduced bindings like it is performed now but with usages handling?

Kobzol:
@fan-tom You can try to use `RsInPlaceVariableIntroducer` (see [here](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt#L89).

fan-tom:
Close as work was finished at #5650

PR #4097, opened 2019-07-05
PERF: micro-optimize macro expansion: traverse AST directly

vlad20012:
`substituteMacro` now works 2x faster, but it doesn't bring measurable changes to the overall expansion time.

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/ast.kt
+++ b/src/main/kotlin/org/rust/openapiext/ast.kt
@@ -1,0 +1,18 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.openapiext
+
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/openapiext/ast.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/** Iterates all children of [this] node and invokes [action] for each one */
+inline fun ASTNode.forEachChild(action: (ASTNode) -> Unit) {
+    var treeChild: ASTNode? = firstChildNode
+
+    while (treeChild != null) {
+        action(treeChild)
+        treeChild = treeChild.treeNext
+    }
+}
```
ice1000: Why not `generateSequence(firstChildNode) { it.treeNext }`?

PR #4108, opened 2019-07-07
INT: intention to convert impl Trait to type parameter

rrevenantt:


src/main/kotlin/org/rust/ide/intentions/ImplTraitToTypeParamIntention.kt
```diff
@@ -0,0 +1,84 @@
...
+            .performInplaceRefactoring(LinkedHashSet())
+    }
+
+    data class Context(
+        val argType: RsTraitType,
+        val fnSignature: RsFunction
+    )
+}
+
+val outerImplTestMark = Testmark("called on outer impl Trait")
```
mchernyavsky: Let's move it into a companion object so it's not shown in the completion list.

src/main/kotlin/org/rust/ide/intentions/ImplTraitToTypeParamIntention.kt
```diff
@@ -0,0 +1,84 @@
...
+        val typeParameterList = fnSignature
+            .typeParameterList
+            ?.apply {
+                addBefore(psiFactory.createComma(), gt)
+            }
+            ?: fnSignature
+                .addAfter(psiFactory.createTypeParameterList(""), fnSignature.identifier)
+                as RsTypeParameterList
+
+        val typeParameterName = (1..Int.MAX_VALUE)
```
mchernyavsky:
```rust
trait T1 {}
fn foo(x: impl T1) {}
```
===>

```rust
trait T1 {}
fn foo<T1: T1>(x: T1) {}
```

mchernyavsky: I think you can use `com.intellij.codeInsight.template.Template` instead of `RsInPlaceVariableIntroducer`. See `Template::addPrimaryVariable` and `Template::addVariable`. Also we have `org.rust.openapiext.Template`.

mchernyavsky:
@rrevenantt Please rebase, should fix the test.

Issue #4080, opened 2019-06-30
Running build throws exception


ice1000:
## Environment

* **Intellij-Rust plugin version:** nightly last night
* **Rust toolchain version:** i believe it's unrelated
* **IDE name and version:** CL-192.5438.15
* **Operating system:** Windows 10 1809

## Problem description

```
java.lang.Throwable: Runner Run is not registered
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:145)
	at com.intellij.execution.ui.RunContentManagerImpl.getContentManagerForRunner(RunContentManagerImpl.java:478)
	at com.intellij.execution.ui.RunContentManagerImpl.getReuseContent(RunContentManagerImpl.java:369)
	at com.intellij.execution.impl.ExecutionManagerKtImpl.startRunProfile(ExecutionManagerKtImpl.kt:34)
	at com.intellij.execution.runners.GenericProgramRunnerKt.startRunProfile(GenericProgramRunner.kt:50)
	at com.intellij.execution.runners.GenericProgramRunner.execute(GenericProgramRunner.kt:17)
	at com.intellij.execution.runners.BaseProgramRunner.execute(BaseProgramRunner.java:46)
	at com.intellij.execution.runners.BaseProgramRunner.execute(BaseProgramRunner.java:35)
	at com.intellij.execution.ProgramRunnerUtil.executeConfigurationAsync(ProgramRunnerUtil.java:92)
	at com.intellij.execution.ProgramRunnerUtil.executeConfiguration(ProgramRunnerUtil.java:40)
	at com.intellij.execution.ProgramRunnerUtil.executeConfiguration(ProgramRunnerUtil.java:170)
	at org.rust.cargo.runconfig.UtilsKt.buildProject(Utils.kt:63)
	at org.rust.cargo.runconfig.RsProjectTasksRunner.run(RsProjectTasksRunner.kt:16)
	at com.intellij.task.impl.ProjectTaskManagerImpl.lambda$run$7(ProjectTaskManagerImpl.java:189)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:294)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #4115, opened 2019-07-08
RUN: Manually initialize run content manager in `buildProject`

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4080.


Undin:
@mchernyavsky try again after #4130

Issue #4058, opened 2019-06-22
Cannot fold doc comments


ice1000:
https://github.com/intellij-rust/intellij-rust/blob/3e128b19196a8dccda2e41e740b4b1c681e6eae2/src/main/kotlin/org/rust/ide/folding/RsFoldingBuilder.kt#L102-L105

Why isn't `OUTER_BLOCK_DOC_COMMENT` folded?

PR #4119, opened 2019-07-09
TYPE: Fix folding of doc comments

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4058.

### DIFF
--- a/src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
+++ b/src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
@@ -145,7 +145,7 @@ class RsDoctestLanguageInjector : MultiHostInjector {
                 }
                 if (index == cratesEndIndex && !alreadyHasMain) {
                     if (prefix == null) prefix = StringBuilder()
-                    prefix.append("fn main() {")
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/injected/RsDoctestLanguageInjector.kt
```diff
@@ -145,7 +145,7 @@ class RsDoctestLanguageInjector : MultiHostInjector {
                 }
                 if (index == cratesEndIndex && !alreadyHasMain) {
                     if (prefix == null) prefix = StringBuilder()
-                    prefix.append("fn main() {")
+                    prefix.append("fn IntellijIdeaRulezzz() {")
```
Undin: Is it intentional change?

mchernyavsky: Yes, it disables folding in this case:
image

vlad20012: Consider using another name to not confuse it with the one from completion, and also extract it to some constant.

mchernyavsky: @vlad20012 Fixed.

Issue #4117, opened 2019-07-09
False positive for E0081 (duplicated enum discriminants)


lovasoa:
## Environment

* **IntelliJ Rust plugin version:** 0.2.101.2129-191
* **Rust toolchain version:** 1.36.0 (a53f9df32 2019-07-03) x86_64-apple-darwin
* **IDE name and version:** IC-191.7479.19
* **Operating system:** macOS 10.14.5

## Problem description
When an enum variant has a discriminant that is not a literal, intellij-rust always thinks that the discriminant is 0, and thus may warn that a discriminant is used twice ([E0081](https://doc.rust-lang.org/error-index.html#E0081)).

## Steps to reproduce
```rust
#[repr(u32)]
enum Dx {
    LEFT = std::u32::MAX - 1,
    MIDDLE = 0,
    RIGHT = 1,
}
```      
[playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=590bdd397528e559b22c0e1559ce70b1)

lovasoa:
The same thing seems to happen when using negative literals : 
![image](https://user-images.githubusercontent.com/552629/60889025-5e094b80-a258-11e9-8173-51841f13cb84.png)

Undin:
Thanks for the report!
By the way, did you use `Create New Issue` action to open this issue or fill environment info manually?

lovasoa:
Yes, I used the new *Create New Issue* feature. It is very handy !

PR #4120, opened 2019-07-09
ANN: Evaluate const enum discriminant

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4117.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -34,6 +34,8 @@ import org.rust.lang.core.types.type
 import org.rust.lang.utils.RsDiagnostic
 import org.rust.lang.utils.RsErrorCode
 import org.rust.lang.utils.addToHolder
+import org.rust.lang.utils.evaluation.ExprValue
+import org.rust.lang.utils.evaluation.RsConstExprEvaluator
 
 class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
     override fun isForceHighlightParents(file: PsiFile): Boolean = file is RsFile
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -1747,6 +1747,37 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class.java) {
...
+        #[repr(i8)]
+        enum Bad {
+            <error descr="Discriminant value `-1` already exists [E0081]">X = -1</error>,
+            <error descr="Discriminant value `-1` already exists [E0081]">Y = -1</error>
+        }
+    """)
+
+    fun `test duplicate enum discriminant repr type invalid range E0081`() = checkErrors("""
+        #[repr(u8)]
+        enum Good {
```
lovasoa: Why is this enum called `Good` when it is in fact invalid ?

Undin: Maybe the name can be better.
But anyway, we don't have API to check overflows like here currently so we prefer don't show error annotations when we fail to evaluate an expression to avoid false-positive annotations.
We are going to improve our constant evaluator to handle such cases too

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -119,18 +121,30 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
...
+                null
+            }
             val idx = int ?: discrCounter
             discrCounter = idx + 1
 
             val previous = indexToVariantMap[idx]
             if (previous != null) {
-                RsDiagnostic.DuplicateEnumDiscriminant(variant, idx).addToHolder(holder)
+                RsDiagnostic.DuplicateEnumDiscriminant(
+                    variant.variantDiscriminant ?: variant,
```
Undin: I think we want to use expression from `variant.variantDiscriminant` not to highlight `=`

mchernyavsky: Fixed.

PR #4128, opened 2019-07-10
Feature/union color

### LINKED ISSUE #4116: Allow customizing unions color
Please add a color scheme option for union names. Currently, they're colored the same as structs.

fan-tom:
Fixes #4116


src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
```diff
@@ -107,7 +107,10 @@ private fun colorFor(element: RsElement): RsColor? = when (element) {
     is RsPatBinding -> {
         if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else null
     }
-    is RsStructItem -> RsColor.STRUCT
+    is RsStructItem -> when(element.kind) {
```
Undin:
```suggestion
    is RsStructItem -> when (element.kind) {
```

src/main/resources/org/rust/ide/colors/highlighterDemoText.rs
```diff
@@ -23,6 +23,11 @@ struct <STRUCT>Object</STRUCT><<TYPE_PARAMETER>T</TYPE_PARAMETER>> {
     <FIELD>fields</FIELD>: <STRUCT>HashMap</STRUCT><<TYPE_PARAMETER>T</TYPE_PARAMETER>, <PRIMITIVE_TYPE>u64</PRIMITIVE_TYPE>>
 }
 
+union <UNION>MyUnion</UNION> {
```
Undin:
```suggestion
<KEYWORD>union</KEYWORD> <UNION>MyUnion</UNION> {
```
`union` is a contextual keyword in Rust. It cannot be highlighted via lexer so it isn't hightlighted automatically in setting highlighting demo. Let's add an explicit color tag for `union` here to get proper highlighting for it

Undin:
@fan-tom Great!
Could you please squash your commits (not necessary into the single one) to avoid commits that fix review comments in history?

PR #4134, opened 2019-07-11
GRAM: add trait aliases parsing support

lundibundi:
This adds basic parsing support for trait aliases (https://github.com/rust-lang/rfcs/blob/master/text/1733-trait-alias.md).

Though it doesn't integrate type inference/type resolution/method resolution/import support and etc for them. I have wanted to implement those but based on the code I've seen (`RsTraitOrImpl`, `TyTypeParameter`, `TraitImplSource`) it is tightly bound to the `RsTraitItem` type which has its own Stub and is therefore not an easy target to substitute with `RsTraitAlias` and I'm not sure which abstraction should be added for that purpose.

Another way would be to modify `RsTraitItem`'s BNF to also cover Trait aliases though that may be harder to maintain as they have some strict distinction (i.e. you cannot write `impl` blocks for Trait aliases, they cannot have Associated types)


vlad20012:
:+1: LGTM
Just one thing: please, write yourself to contributors.txt and sign our CLA as described in [contributors.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#cla)

lundibundi:
@vlad20012 done.

mchernyavsky:
bors retry

mchernyavsky:
bors r=vlad20012

Issue #3994, opened 2019-06-09
Add an option to show stderr/stdout in test toolwindow always


sirgl:
## Environment

* **Intellij-Rust plugin version:** v0.2.99.2126-191
* **Rust toolchain version:** 1.18.3
* **IDE name and version:** 2019.1.4
* **Operating system:** mac os

## Problem description
Now if test successfully passed, no output shown. I think, it is wrong, because there may be logging, `dbg!` values and so on, which can be helpful even for passed tests - you may want to compare correct value from passed test and incorrect value from failed one (for java e. g. all output shows always).

Undin:
As I understand it's because of https://github.com/rust-lang/rust/issues/54669
@mchernyavsky am I right?

PR #4139, opened 2019-07-12
RUN: enable --show-output when executing tests

emmericp:
Fixes https://github.com/intellij-rust/intellij-rust/issues/3994

This requires this patch for rust (hopefully in 1.38?): https://github.com/rust-lang/rust/pull/62600

This is work in progress because it doesn't work yet, I'm looking for some help here:
The problem is you can't run `rustVersion()` in the UI thread and I'd really like to have this behavior as default if the compiler is new enough.
What's the best way to achieve this? Fallback would be to make this an option for the run configuration, but that would be a little bit annoying since virtually everyone would want to enable this?

Edit: changed the code to hardcode 1.38 as Rust version passed to `patchArgs` because I'm using this branch ;) Looking for a good way to get the Rust version here.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/CargoTestRunState.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/CargoTestRunState.kt
@@ -13,16 +13,25 @@ import com.intellij.execution.configurations.RunConfiguration
 import com.intellij.execution.runners.ExecutionEnvironment
 import com.intellij.execution.runners.ProgramRunner
 import com.intellij.execution.testframework.autotest.ToggleAutoTestAction
+import com.intellij.util.text.SemVer
 import org.rust.cargo.runconfig.buildtool.CargoPatch
 import org.rust.cargo.runconfig.command.CargoCommandConfiguration
 import org.rust.cargo.runconfig.console.CargoTestConsoleBuilder
 import org.rust.cargo.toolchain.CargoCommandLine
+import org.rust.cargo.toolchain.RustChannel
+import org.rust.cargo.toolchain.RustcVersion
+import java.time.LocalDate
 
 class CargoTestRunState(
     environment: ExecutionEnvironment,
     runConfiguration: CargoCommandConfiguration,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/cargo/runconfig/CargoTestRunState.kt
```diff
@@ -75,6 +76,9 @@ class CargoTestRunState(
             } else {
                 libtestArgs.add(noFormatJson)
             }
+            if (rustcVer != null && rustcVer >= SemVer("1.38.0", 1, 38, 0) && !libtestArgs.contains("--show-output")) {
```
mchernyavsky: Maybe we also need to check current channel?

emmericp:
The feature has been merged and is available in Rust 1.39 nightly builds with a build date of 28-08-2019 or later. I've updated this PR accordingly and fixed the merge conflict

src/main/kotlin/org/rust/cargo/runconfig/CargoTestRunState.kt
```diff
@@ -60,7 +66,24 @@ class CargoTestRunState(
...
+            if (checkShowOutputSupport(rustcVer)) {
+                post.add("--show-output")
+            }
+
             return if (post.isEmpty()) pre else pre + "--" + post
         }
+
+        private fun checkShowOutputSupport(ver: RustcVersion?): Boolean {
+            if (ver == null) return false
+            // --show-output is supported since 1.39.0-nightly/dev with a build date later than 2019-08-27
```
Undin: Can we check that this flag is supported without version checking? Like we do with some other flags

mchernyavsky: I think no. The `--show-output` is only mentioned in `cargo test -- --help` output, but this command may trigger compilation.

emmericp: i don't think it's possible to detect this without compiling tests as it's a feature in libtest that's only in the resulting test binary

PR #4150, opened 2019-07-16
Bugfix/arbitrary self types#4138

### LINKED ISSUE #4138: "Implement members" ignores arbitrary self types.
## Environment

* **IntelliJ Rust plugin version:** 0.2.102.2682-192-nightly

## Problem description

`self` is inserted as `Self` instead of one of a few new self types stabilized recently on the way to arbitrary self types.

## Steps to reproduce

Code:
```rust
trait X {
    fn x(self: Pin<&mut Self>);
    fn y(self: *const Self);
}

struct A{}

impl X for A {
}
```
Use the "implement members" quick fix.

Screenshot_20190712_164425

fan-tom:
Fixes #4138

### DIFF
--- a/src/test/kotlin/org/rust/ide/refactoring/implementMembers/ImplementMembersHandlerTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/implementMembers/ImplementMembersHandlerTest.kt
@@ -8,7 +8,10 @@ package org.rust.ide.refactoring.implementMembers
 import com.intellij.openapi.actionSystem.ex.ActionManagerEx
 import junit.framework.TestCase
 import org.intellij.lang.annotations.Language
+import org.rust.MinRustcVersion
+import org.rust.ProjectDescriptor
 import org.rust.RsTestBase
+import org.rust.WithStdlibRustProjectDescriptor
 import org.rust.ide.inspections.RsTraitImplementationInspection
 
 class ImplementMembersHandlerTest : RsTestBase() {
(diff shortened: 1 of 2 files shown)


fan-tom:
How do I limit tests to not run on old_rust? I tried MinRustcVersion annotation but it seems it doesn't work

Undin:
>  I tried MinRustcVersion annotation but it seems it doesn't work

And it's expected because this annotation is supported only by `RsWithToolchainTestBase` (as described in javadoc)

But indeed, there are cases when this annotation may be useful in `RsTestBase` too. So after #4154 it will be possible

src/test/kotlin/org/rust/ide/refactoring/implementMembers/ImplementMembersHandlerTest.kt
```diff
@@ -8,9 +8,13 @@ package org.rust.ide.refactoring.implementMembers
 import com.intellij.openapi.actionSystem.ex.ActionManagerEx
 import junit.framework.TestCase
 import org.intellij.lang.annotations.Language
+import org.rust.ProjectDescriptor
 import org.rust.RsTestBase
+import org.rust.WithStdlibRustProjectDescriptor
+import org.rust.cargo.MinRustcVersion
 import org.rust.ide.inspections.RsTraitImplementationInspection
 
+@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
```
Undin: Let's avoid setting up `WithStdlibRustProjectDescriptor` for whole test class if not all test cases require stdlib. It slows down tests without real need

vlad20012:
Could you, please, squash commits?

PR #4154, opened 2019-07-17
T: improve `MinRustcVersion` annotation

Undin:
Now it can be used in inheritors of `RsTestBase`. Previously, it was supported only by `RsWithToolchainTestBase`


Undin:
bors retry

Issue #4155, opened 2019-07-18
GDB pretty-printers crash on Option-like enums


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.2.102.2685-192-nightly
* **Rust toolchain version:** 1.36.0
* **IDE name and version:** CLion 2019.2
* **Operating system:** Ubuntu

## Problem description

The GDB pretty-printer introduced in #3715 crashes on some enums, specifically on `Option`-like enums with a non-primitive value type, because of the invalid discriminant. Maybe this is because of wrong DWARF generation in rustc, though console GDB works well on such enums. So probably there should be a way to read the valid discriminant somehow.

Without pretty-printers:
<img width="956" alt="Screenshot 2019-07-18 at 11 53 10" src="https://user-images.githubusercontent.com/4854600/61443833-eff90e80-a952-11e9-91e4-b1a3f484a951.png">

With bundled pretty-printers:
<img width="779" alt="Screenshot 2019-07-18 at 11 57 08" src="https://user-images.githubusercontent.com/4854600/61444192-96ddaa80-a953-11e9-8c5a-a80183431c87.png">

Console GDB:
<img width="726" alt="Screenshot 2019-07-18 at 12 12 32" src="https://user-images.githubusercontent.com/4854600/61445351-eb822500-a955-11e9-8315-2134085aa0fa.png">

Kobzol:
Thanks, I didn't have code that could reproduce it, now that I have it, I'll try to take a look at it.

Kobzol:
So for this code:
```rust
enum MyEnum1<T> { A(T), B }

fn main()
{
    let a1 = MyEnum1::A(String::from("111"));
    let a2 = MyEnum1::B;
}
```
Discriminant in a1 is `94689684138817` (should be `0`).
Discriminant in a2 is `0` (should be `1`).

If the enum is inverted, i.e. `enum MyEnum2<T> { B, A(T) }`, discriminant for `MyEnum2::B` is still 0 (which is in this case correct), discriminant for `MyEnum2::A` is garbage, as in `MyEnum1`.

This doesn't happen with simple types. I think that it has something to do with some zero-sized optimization. If I use the enum for example with a type that contains a reference (i.e. 0 is not a valid value for that type), it also causes this error.

I suspect that this may be a Rustc DWARF generation issue (the enum DWARF generation was changed a few months ago). I'll try to investigate, but it will probably take me some time.

Kobzol:
[...]

PR #4156, opened 2019-07-18
Pretty-printers for GDB: Disable EnumProvider in case of invalid discriminant

ortem:
Hotfix of #4155. Just disables the `EnumProvider` in case of the invalid discriminant.

@Kobzol I haven't found the proper way to print such enums yet, so I would be really grateful if you would investigate this problem too!


mchernyavsky:
bors retry

Undin:
bors r=ortem

PR #4167, opened 2019-07-20
Fix warnings about action name capitalization

Undin:


src/main/resources/META-INF/plugin.xml
```diff
@@ -985,12 +985,12 @@
         </group>
         <action id="Rust.ToggleExternalLinterOnTheFlyAction"
                 class="org.rust.ide.actions.ToggleExternalLinterOnTheFlyAction"
-                text="Run external linter (cargo check / clippy) on the fly">
+                text="Run External Linter on the Fly">
```
mchernyavsky: Without `cargo check / clippy` the option is not searchable via `Find Action`.

Undin:
bors retry

PR #4172, opened 2019-07-22
INSP: lift inspection for return expression

Undin:
TODO:
* [ ] Find out expression size threshold
* [x] Check with real projects


src/test/kotlin/org/rust/ide/inspections/RsLiftInspectionTest.kt
```diff
@@ -0,0 +1,168 @@
...
+                    3
+                } else {
+                    {
+                        4
+                    }
+                }
+            }
+        }
+    """, checkWeakWarn = true)
+}
```
ortem: Maybe add such tests as well: 
* inspection is unavailable when there is a `return` without expression
* comments after `return` doesn't affect inspection availability

Undin:
bors r=ortem

Issue #4174, opened 2019-07-23
'Self' in generic arguments is wrongly accused of being a trait object


burjui:
## Environment

* **IntelliJ Rust plugin version:** 0.2.102.2130-191
* **Rust toolchain version:** rustc 1.36.0 (a53f9df32 2019-07-03)
* **IDE name and version:** IntelliJ IDEA 2019.1.3 (Community Edition)
* **Operating system:** Arch Linux

## Problem
In the following snippet
```rust
// 2018 edition
trait X {
    type A;
}
trait Y<T> {}
trait Z: X + Y<<Self as X>::A> {}
```
`Self` is highlighted with orange background and shows the following hint: `Trait objects without an explicit 'dyn' are deprecated`. Of course, on every commit an "are you sure" dialog pops up, because this is counted as a warning, even though it doesn't come from the compiler.

vlad20012:
Hmm, they should be `TyTypeParameter.self()`

PR #4175, opened 2019-07-23
INSP: Don't warn about missing 'dyn' for Self

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4174.

### DIFF
--- a/src/test/kotlin/org/rust/ide/inspections/RsImplicitTraitObjectInspectionTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/RsImplicitTraitObjectInspectionTest.kt
@@ -45,6 +45,10 @@ class RsImplicitTraitObjectInspectionTest : RsInspectionsTestBase(RsImplicitTrai
         fn test<X>(a: X) where X : Trait {}
     """)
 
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/rust/ide/inspections/RsImplicitTraitObjectInspectionTest.kt
```diff
@@ -45,6 +45,12 @@ class RsImplicitTraitObjectInspectionTest : RsInspectionsTestBase(RsImplicitTrai
         fn test<X>(a: X) where X : Trait {}
     """)
 
+    fun `test no warning on Self`() = checkByText("""
+        trait X { type A; }
+        trait Y<T> {}
+        trait Z: X + Y<<Self as X>::A> {}
```
vlad20012: The example is too complex. Works for `trait T where Self: Foo {}`

mchernyavsky:
bors r=vlad20012

vlad20012:
Oh, I'm wrong, we don't look at types here, so the type of `Self` is irrelevant

PR #4180, opened 2019-07-25
REF: introduce RsPathCodeFragment, use it in RsCodeFragmentFactory

vlad20012:
Simplifies path handling. Also, now it's possible to pass namespaces
to `RsCodeFragmentFactory.createPath`

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsCodeFragment.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsCodeFragment.kt
@@ -17,10 +17,13 @@ import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.testFramework.LightVirtualFile
 import org.rust.lang.RsFileType
 import org.rust.lang.RsLanguage
+import org.rust.lang.core.parser.RustParserUtil
+import org.rust.lang.core.parser.RustParserUtil.PathParsingMode
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.RsInferenceContextOwner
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsCodeFragment.kt
```diff
@@ -111,3 +127,29 @@ class RsTypeReferenceCodeFragment(project: Project, text: CharSequence, context:
...
+        ns: Set<Namespace>
+    ) : this(createFileViewProvider(project, text, eventSystemEnabled), context, mode, ns)
+
+    val path: RsPath? get() = PsiTreeUtil.getChildOfType(this, RsPath::class.java)
+
+    companion object {
+        @JvmStatic
+        private fun RustParserUtil.PathParsingMode.elementType() = when (this) {
+            RustParserUtil.PathParsingMode.NO_COLONS -> RsCodeFragmentElementType.PATH_WITHOUT_COLONS
+            else -> RsCodeFragmentElementType.PATH_WITH_COLONS
```
Undin: Maybe we want to add path code fragment type for `RustParserUtil.PathParsingMode.NO_TYPES` too?
Because now it's possible to pass `NO_TYPES` as `mode` and for this case we will use unexpected parsing rool

Undin: @vlad20012 I still can pass `NO_TYPES` into `RsPathCodeFragment` constructor

vlad20012: fixed

Undin:
bors retry

PR #4181, opened 2019-07-25
PERF: optimize very hot `getExpandedFrom` method

vlad20012:
(micro-optimization)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
@@ -620,9 +620,10 @@ private class MacroExpansionServiceImplInner(
 
     fun getExpandedFrom(element: RsExpandedElement): RsMacroCall? {
         checkReadAccessAllowed()
-        if (element.stubParent is RsFile /*TODO*/) {
-            val file = element.containingFile.virtualFile ?: return null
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -620,9 +620,10 @@ private class MacroExpansionServiceImplInner(
 
     fun getExpandedFrom(element: RsExpandedElement): RsMacroCall? {
         checkReadAccessAllowed()
-        if (element.stubParent is RsFile /*TODO*/) {
-            val file = element.containingFile.virtualFile ?: return null
-            if (isExpansionFile(file)) {
+        val parent = element.stubParent
+        if (parent is RsFile) {
+            val file = parent.virtualFile ?: return null
```
Undin:
```suggestion
            val file = parent.virtualFile as? VirtualFileWithId ?: return null
```

vlad20012: This doesn't work because we don't want to cast the variable to `VirtualFileWithId`, we want just check that it is `VirtualFileWithId` and use it as `VirtualFile`

Undin: Indeed!

PR #4182, opened 2019-07-25
PERF: introduce `RsCachedImplItem`

vlad20012:
I found this place during the recent performance investigation. This speeds up type inference up to 30% (!!!) (the baseline is before #4179)

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -238,9 +238,8 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
             val traits = sources.mapNotNull { source ->
                 val trait = when (source) {
                     is TraitImplSource.ExplicitImpl -> {
-                        val impl = source.value
-                        val traitRef = impl.traitRef ?: return null
-                        traitRef.resolveToTrait() ?: return@mapNotNull null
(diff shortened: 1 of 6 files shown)


vlad20012:
Fixed

src/main/kotlin/org/rust/lang/core/resolve/ref/RsResolveCache.kt
```diff
@@ -122,7 +122,7 @@ class RsResolveCache(messageBus: MessageBus) {
     private fun getCacheFor(element: PsiElement, dep: ResolveCacheDependency): ConcurrentMap<PsiElement, Any?> {
         return when (dep) {
             ResolveCacheDependency.LOCAL, ResolveCacheDependency.LOCAL_AND_RUST_STRUCTURE -> {
-                val owner = element.findModificationTrackerOwner()
+                val owner = element.findModificationTrackerOwner(strict = false)
```
Undin: As I understand, you need this change to return `impl` item as modification tracker owner in case of `RsCachedImplItem`
And looks like we always get null here because `RsImplItem` doesn't implement `RsModificationTrackerOwner`.
And as a result we always return `rustStructureDependentCache`.
Maybe it's better to use `RUST_STRUCTURE` dependency for `RsCachedImplItem`?

vlad20012: Wow, nice catch! You're right!

vlad20012:
Fixed

PR #4183, opened 2019-07-25
PERF&STUB: stub flag that function has a self parameter (i.e. is method)

vlad20012:


src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -22,8 +22,8 @@ import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
 import javax.swing.Icon
 
-val RsFunction.isAssocFn: Boolean get() = selfParameter == null && owner.isImplOrTrait
-val RsFunction.isMethod: Boolean get() = selfParameter != null && owner.isImplOrTrait
+val RsFunction.isAssocFn: Boolean get() = !hasSelfParameters && owner.isImplOrTrait
+val RsFunction.isMethod: Boolean get() = hasSelfParameters
```
Undin: Why did you drop `owner.isImplOrTrait` condition here?
It can be important while incorrect code analyzing

vlad20012: We check it in `RsSyntaxErrorsAnnotator`. In other places, I think, we can treat any function that has `self` parameter as a method

Undin: > We check it in RsSyntaxErrorsAnnotator

And how does it prevent incorrect analysis in case of function with `&self` parameter in other inspections/annotators?

>  I think, we can treat any function that has self parameter as a method

Of course, no. A user (don't forget about beginners) can add `&self` parameter to any function and it potentially may get some irrelevant annotations about methods although it's function.
Let's provide a clear API to avoid an unexpected combination of highlighting annotations, simplify review (otherwise, you always will have to remember about this case while reviewing), etc.

vlad20012: Okaaay, fixed.
I leaved the same behavior for `isMethod` and used `hasSelfParameters` in the resolve (with a comment about why is it used)

PR #4189, opened 2019-07-26
PERF: micro-optimize name resolution (mostly local variables)

vlad20012:


src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -37,8 +37,19 @@ enum class ScopeEvent : ScopeEntry {
     /**
      * Communicate to the resolve processor that we are about to process wildecard imports.
      * This is basically a hack to make winapi 0.2 work in a reasonable amount of time.
+     *
+     * This event brings (very rare) false negatives: an item can be resolved when it
+     * should be multiresolved (i.e. unresolved from the rustc view)
```
vlad20012: Just a doc, i.e. this fact wasn't changed by this PR

vlad20012:
The performance impact is too little, and it can bring new bugs into the name resolution, so I'll close it

PR #4190, opened 2019-07-26
REF&PERF: better cache cargo target for files

vlad20012:
The idea is that we have 4 things:
- cargo project
- cargo workspace
- cargo target
- crate root

that can be cheaply retrieved and cached all together.

This speeds up highlighting up to 5% (the baseline is before #4179)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
@@ -6,7 +6,9 @@
 package org.rust.lang.core.psi
 
 import com.intellij.extapi.psi.PsiFileBase
+import com.intellij.injected.editor.VirtualFileWindow
 import com.intellij.openapi.fileTypes.FileType
+import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Key
 import com.intellij.openapi.util.io.FileUtil
 import com.intellij.openapi.vfs.VirtualFile
@@ -16,6 +18,9 @@ import com.intellij.psi.PsiFile
 import com.intellij.psi.util.CachedValue
 import com.intellij.psi.util.CachedValueProvider
 import com.intellij.psi.util.CachedValuesManager
+import org.rust.cargo.project.model.CargoProject
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.cargo.project.workspace.CargoWorkspace
 import org.rust.ide.injected.isDoctestInjection
 import org.rust.lang.RsConstants
 import org.rust.lang.RsFileType
@@ -27,6 +32,7 @@ import org.rust.lang.core.resolve.ref.RsReference
 import org.rust.lang.core.stubs.RsFileStub
 import org.rust.lang.core.stubs.index.RsIncludeMacroIndex
 import org.rust.lang.core.stubs.index.RsModulesIndex
+import org.rust.openapiext.toPsiFile
 
 /**
  * This class was added in order to fix [RsCodeFragment] copying inside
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -56,17 +57,57 @@ class RsFile(
...
+
+        return if (cargoTarget != null) {
+            // `possibleCrateRoot` is a "real" crate root only if we're able to find a target for it
+            CachedData(cargoProject, workspace, cargoTarget, possibleCrateRoot)
+        } else {
+            CachedData(cargoProject, workspace)
+        }
+    }
+
+    data class CachedData(
```
Undin:
```suggestion
    private data class CachedData(
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
```diff
@@ -61,23 +57,15 @@ fun PsiFileSystemItem.findCargoPackage(): CargoWorkspace.Package? {
...
+    get() = (contextualFile.originalFile as? RsFile)?.cargoTarget
 
 val RsElement.containingCargoPackage: CargoWorkspace.Package? get() = containingCargoTarget?.pkg
 
 val PsiElement.isEdition2018: Boolean
     get() = contextOrSelf<RsElement>()?.containingCargoTarget?.edition == CargoWorkspace.Edition.EDITION_2018
 
 /** @see org.rust.ide.injected.RsDoctestLanguageInjector */
-private fun VirtualFile.getInjectedFromIfDoctestInjection(project: Project): RsFile? {
+fun VirtualFile.getInjectedFromIfDoctestInjection(project: Project): RsFile? {
```
Undin: Maybe rename it somehow?
When it was a private method, it was localized in single file and everything was ok. But now it's public, so I'd prefer some more readable name. Аnd probably more detailed javadoc because I'm not sure that we will come up with a self-explained name.
Maybe `getDelegateRsFile` with something like 
```
/**
 * Returns real rust psi file in when the current one is injected.
 * For example, in case if language injection in rust doc tests
 *
 * @see org.rust.ide.injected.RsDoctestLanguageInjector
 */
```
as java doc

Undin: Also, I feel some inconsistency in method signature- the receiver is `VirtualFile`, returning value is `PsiFile`.
Looks like we can safely convert it into
`fun PsiFileSystemItem.getInjectedFromIfDoctestInjection(project: Project): RsFile?`
because in all places when we have the corresponding psi element.
Also, it restricts area when this method can be called.
At the same time, in all calling places we need virtual file as well.
Just my thoughts

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -56,17 +57,49 @@ class RsFile(
...
             return CachedValuesManager.getCachedValue(this, key) {
-                CachedValueProvider.Result(doCrateRoot(), rustStructureOrAnyPsiModificationTracker)
+                CachedValueProvider.Result(doGetCachedData(), rustStructureOrAnyPsiModificationTracker)
             }
         }
 
-    private fun doCrateRoot(): RsMod? =
-        superMods.lastOrNull()?.takeIf { it.isCrateRoot }
+    private fun doGetCachedData(): CachedData {
+        val possibleCrateRoot = (superMods.lastOrNull() as? PsiFile)?.originalFile as? RsFile
```
Undin: Hm, now we extract `originalFile` but previously we didn't do it.
Can it affect something when we are working with a copy of the file (completion?) ?

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -56,17 +57,49 @@ class RsFile(
...
+            ?: return CachedData()
```
Undin: Maybe create a single object for empty cached data?

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -56,17 +57,49 @@ class RsFile(
...
+
+        val cargoProject = possibleCrateRoot.findCargoProject() ?: return CachedData()
+        val workspace = cargoProject.workspace ?: return CachedData(cargoProject)
+        val cargoTarget = workspace.findTargetByCrateRoot(crateRootVFile)
+
+        return if (cargoTarget != null) {
+            // `possibleCrateRoot` is a "real" crate root only if we're able to find a target for it
+            CachedData(cargoProject, workspace, cargoTarget, possibleCrateRoot)
+        } else {
+            CachedData(cargoProject, workspace)
```
Undin: Can you add a comment when this case is possible?

src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
vlad20012: I removed its usages from `RsElement.kt` (I found that they're redundant), moved to `RsFile.kt` and made it private again :)

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
vlad20012: Really I think that this shouldn't affect anything and it's always safe to use an original element of RsFile, but I agree that I shouldn't change this logic in this PR. Anyway, this change isn't meaningful. So I brought back the old behavior

PR #4193, opened 2019-07-28
CLEAN: fix compiler warnings

vlad20012:


src/main/kotlin/org/rust/ide/inspections/import/ui.kt
```diff
@@ -71,7 +71,7 @@ private class PopupImportItemUi(private val project: Project, private val dataCo
             override fun getTextFor(value: ImportCandidate): String = value.info.usePath
             override fun getIconFor(value: ImportCandidate): Icon? = value.qualifiedNamedItem.item.getIcon(0)
         }
-        val popup = object : ListPopupImpl(step) {
+        val popup = object : ListPopupImpl(project, step) {
```
Undin: Shouldn't be changed (just suppressed) because it cannot be compiled with 2019.1

PR #4197, opened 2019-07-29
DBG: Load pretty-printers and rustc sources via RsDebugProcessConfigurator

ortem:
Now Rust pretty-printers and rustc sources will be loaded even when C/C++ project was launched. So debugging Rust code inside a cross-language project becomes as comfortable as debugging pure-Rust projects.


debugger/src/192/main/kotlin/org/rust/debugger/runconfig/RsDebugProcessConfigurator.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.rust.debugger.*
+import org.rust.debugger.settings.RsDebuggerSettings
+import org.rust.ide.notifications.showBalloon
+import java.nio.file.InvalidPathException
+
+class RsDebugProcessConfigurator : CidrDebugProcessConfigurator {
+    override fun configure(process: CidrDebugProcess) {
+        val project = process.project
+        val toolchain = project.toolchain
+        val cargoProject = project.cargoProjects.allProjects.firstOrNull()
```
Undin: If we have more than one cargo project we may get wrong one here.
Let's better use `CargoCommandConfiguration#findCargoProject` here.
If you need to pass additional params to do it, you can pass them via `RsDebugRunParameters` and extract them via `CidrDebugProcess#getRunParameters`

debugger/src/192/main/kotlin/org/rust/debugger/runconfig/RsDebugProcessConfigurator.kt
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    }
+
+    companion object {
+        private val LOG: Logger = Logger.getInstance(RsDebugProcessConfigurator::class.java)
+    }
+}
+
+private class Loader(
```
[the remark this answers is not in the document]
ortem: How's about `RsExtraDebugDataLoader`?

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsExtraDebugDataLoader.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.jetbrains.cidr.execution.debugger.backend.lldb.LLDBDriver
+import org.rust.cargo.project.model.CargoProject
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.runconfig.command.workingDirectory
+import org.rust.debugger.*
+import org.rust.debugger.settings.RsDebuggerSettings
+import org.rust.ide.notifications.showBalloon
+import java.nio.file.InvalidPathException
+
+class RsExtraDebugDataLoader(private val process: CidrDebugProcess, cargoProject: CargoProject?) {
```
Undin: Maybe `RsDebugProcessConfigurationHelper`?

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsExtraDebugDataLoader.kt
```diff
@@ -0,0 +1,137 @@
...
+    private val threadId = process.currentThreadId
+    private val frameIndex = process.currentFrameIndex
+
+    private val commitHash = cargoProject?.rustcInfo?.version?.commitHash
+
+    private val sysroot: String? by lazy {
+        cargoProject?.workingDirectory?.let { toolchain?.getSysroot(it) }
+    }
+
+    fun load() {
```
Undin: Maybe `configure` or `setUp` in better?
If final class name contains `Configuration`, `configure` will be preferable

debugger/src/191/main/kotlin/org/rust/debugger/runconfig/RsLocalDebugProcess.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.rust.cargo.runconfig.CargoRunStateBase
+
+class RsLocalDebugProcess(
+    parameters: RunParameters,
+    debugSession: XDebugSession,
+    consoleBuilder: TextConsoleBuilder
+) : CidrLocalDebugProcess(parameters, debugSession, consoleBuilder) {
+
+    fun setupDebugSession(state: CargoRunStateBase) {
+        RsExtraDebugDataLoader(this, state.cargoProject).load()
```
Undin: Should be renamed as well to compile with 2019.1

ortem:
@Undin Oh, really. Now in case of cross-language projects Rust lldb pretty-printers are enabled only for specific std types, but disabled for arbitrary struct/enum.

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugProcessConfigurationHelper.kt
```diff
@@ -0,0 +1,167 @@
...
+        return sysroot
+    }
+
+    private fun String.systemDependentAndEscaped(): String =
+        StringUtil.escapeStringCharacters(FileUtil.toSystemDependentName(this))
+
+    companion object {
+        private val LOG: Logger = Logger.getInstance(RsDebugProcessConfigurationHelper::class.java)
+
+        private val RUST_STD_TYPES: List<String> = listOf(
```
Undin: Let's add a comment here and in `rust_types.py` that these regexes should be synchronized

PR #4199, opened 2019-07-30
[WIP] GRAM: pin path

vlad20012:
Better parser recovery for incomplete paths


vlad20012:
Oh, `RsPath.referenceNameElement` should be nullable now...

src/main/kotlin/org/rust/lang/core/psi/ext/RsPath.kt
```diff
@@ -91,7 +91,7 @@ abstract class RsPathImplMixin : RsStubbedElementImpl<RsPathStub>,
 
     override val referenceNameElement: PsiElement
         get() = checkNotNull(identifier ?: self ?: `super` ?: cself ?: crate) {
-            "Path must contain identifier: $this ${this.text} at ${this.containingFile.virtualFile.path}"
+            "Path must contain identifier: $this ${this.text} at ${this.containingFile.virtualFile?.path}"
```
ortem: Now `IllegalStateException` is thrown for every incomplete path (e.g. `Foo::`) in the editor.

vlad20012: :+1: 
This is the reason why it's WIP

vlad20012:
I'll try to fix it later

PR #4202, opened 2019-07-30
RES: Take into account namespaces when resolving const generics

mchernyavsky:
bors r+


Undin:
@mchernyavsky It would be great to have test(s) for this fix

PR #4203, opened 2019-07-31
Usages tooltip scope

### LINKED ISSUE #4200: Usages tooltip doesn't show function name
## Problem description
When I Cmd+Click on struct field, I get small window with usages of that field
изображение

But there are no function and struct names, like in PyCharm
изображение

fan-tom:
Closes #4200.

Adds function/method name to usages tooltip.

изображение

### DIFF
--- a/src/main/kotlin/org/rust/ide/search/RsFunctionGroupingRuleProvider.kt
+++ b/src/main/kotlin/org/rust/ide/search/RsFunctionGroupingRuleProvider.kt
@@ -1,0 +1,33 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.search
+
+import com.intellij.openapi.project.Project
+import com.intellij.usages.PsiNamedElementUsageGroupBase
+import com.intellij.usages.Usage
+import com.intellij.usages.UsageGroup
+import com.intellij.usages.UsageTarget
+import com.intellij.usages.impl.FileStructureGroupRuleProvider
+import com.intellij.usages.rules.PsiElementUsage
+import com.intellij.usages.rules.SingleParentUsageGroupingRule
+import com.intellij.usages.rules.UsageGroupingRule
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsImplItem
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/search/RsFunctionGroupingRuleProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.intellij.usages.rules.UsageGroupingRule
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsImplItem
+
+class RsFunctionGroupingRuleProvider : FileStructureGroupRuleProvider {
+    override fun getUsageGroupingRule(project: Project): UsageGroupingRule? {
+        return RsFunctionGroupingRule()
+    }
+
+    private class RsFunctionGroupingRule : SingleParentUsageGroupingRule() {
```
mchernyavsky: Make it staitc (move to `companion object { ... }`).

src/main/kotlin/org/rust/ide/search/RsFunctionGroupingRuleProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+        override fun getParentGroupFor(usage: Usage, targets: Array<UsageTarget>): UsageGroup? {
+            if (usage !is PsiElementUsage) return null
+            val psiElement = usage.element
+            val rsFunction = PsiTreeUtil.getParentOfType(
```
mchernyavsky: You can use `ancestorOrSelf`:
```kotlin
override fun getParentGroupFor(usage: Usage, targets: Array<UsageTarget>): UsageGroup? {
    if (usage !is PsiElementUsage) return null
    val rsFunction = usage.element.ancestorOrSelf<RsFunction>(RsImplItem::class.java) ?: return null
    return PsiNamedElementUsageGroupBase(rsFunction)
}
```

src/main/kotlin/org/rust/ide/search/RsFunctionGroupingRuleProvider.kt
Undin: @mchernyavsky Are you about `RsFunctionGroupingRule` class?
If yes then it's already static

Undin:
bors retry

PR #4217, opened 2019-08-02
TYPE: New interactive inlay type hints support

ortem:
type_hints

### DIFF
--- a/src/192/main/kotlin/org/rust/ide/hints/RsInlayTypeHintsProvider.kt
+++ b/src/192/main/kotlin/org/rust/ide/hints/RsInlayTypeHintsProvider.kt
@@ -1,0 +1,192 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.hints
+
+import com.intellij.codeInsight.hints.*
+import com.intellij.codeInsight.hints.presentation.InlayPresentation
+import com.intellij.codeInsight.hints.presentation.InsetPresentation
+import com.intellij.codeInsight.hints.presentation.MenuOnClickPresentation
+import com.intellij.openapi.components.service
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.DumbService
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.ui.components.CheckBox
+import com.intellij.ui.layout.panel
+import com.intellij.util.ui.JBUI
+import org.rust.lang.RsLanguage
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.types.declaration
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.type
+import javax.swing.JPanel
+
+class RsInlayTypeHintsProvider : InlayHintsProvider<RsInlayTypeHintsProvider.Settings> {
+    override val key: SettingsKey<Settings> get() = KEY
+
+    override val name: String get() = "Type hints"
+
+    override val previewText: String
+        get() = """
+            struct Foo<T1, T2, T3> { x: T1, y: T2, z: T3 }
+            
+            fn main() {
+                let foo = Foo { x: 1, y: "abc", z: true };
+            }
+            """.trimIndent()
+
+    override fun createConfigurable(settings: Settings): ImmediateConfigurable = object : ImmediateConfigurable {
+        val showForVariables = "Show for variables"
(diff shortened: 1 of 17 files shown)


panstromek:
Wow, this is cool. Can you somehow use this without mouse? (because it seems you cannot move caret to typehint to do `ctrl+b`...)

ortem:
@panstromek AFAIK it's impossible for now. I hope IntelliJ Platform will support such a feature someday!

src/192/main/kotlin/org/rust/ide/hints/RsPlainTypeHint.kt
```diff
@@ -0,0 +1,20 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.hints
+
+import com.intellij.codeInsight.hints.InlayInfo
+import com.intellij.codeInsight.hints.Option
+import com.intellij.psi.PsiElement
+
+/** New interactive type hints are used instead */
```
mchernyavsky: Maybe add an (registry) option so that the user can switch between old and new type hints?

src/192/main/kotlin/org/rust/ide/hints/RsInlayTypeHintsProvider.kt
```diff
@@ -0,0 +1,181 @@
...
+            }
+            """.trimIndent()
+
+    override fun createConfigurable(settings: Settings): ImmediateConfigurable = object : ImmediateConfigurable {
+        val showForVariables = "Show for variables"
+        val showForLambdas = "Show for closures"
+        val showForIterators = "Show for iterators"
+        val showObviousTypes = "Show obvious types"
+
+        private val varField = CheckBox(showForVariables)
```
mchernyavsky: Maybe move these settings to a separate PR?

src/192/main/kotlin/org/rust/ide/hints/RsTypeHintsPresentationFactory.kt
```diff
@@ -0,0 +1,259 @@
...
+            listOf(traitPresentation, traitTypesPresentation).join()
+        }
+    }
+
+    private fun checkSize(level: Int, elementsCount: Int): Boolean =
+        level > FOLDING_LEVEL || elementsCount > FOLDING_ELEMENTS_COUNT
+
+    private fun isDefaultTypeParameter(argument: Ty, parameter: RsTypeParameter): Boolean {
+        val defaultType = parameter.typeReference?.type ?: return false
+        return argument.shortPresentableText == defaultType.shortPresentableText
```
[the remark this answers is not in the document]
ortem: Sure, but I think we can ignore such cases not to use real name resolution here because type hints should be rendered as fast as possible.

src/192/main/kotlin/org/rust/ide/hints/RsInlayTypeHintsProvider.kt
```diff
@@ -0,0 +1,181 @@
...
+    override fun createSettings(): Settings = Settings()
+
+    override fun getCollectorFor(file: PsiFile, editor: Editor, settings: Settings, sink: InlayHintsSink): InlayHintsCollector? =
+        object : FactoryInlayHintsCollector(editor) {
+
+            val typeHintsFactory = RsTypeHintsPresentationFactory(factory, settings.showObviousTypes)
+
+            override fun collect(element: PsiElement, editor: Editor, sink: InlayHintsSink): Boolean {
+                if (file.project.service<DumbService>().isDumb) return false
+                if (element !is RsElement) return false
```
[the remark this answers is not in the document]
ortem: I'm not sure there is a big difference because AFAIK this method only is invoked when `isDumb` is false (there are checks at call sites) and when the file's language is `RsLanguage`. I've just copied the sequence from another implementation of this method.

src/192/main/kotlin/org/rust/ide/hints/RsTypeHintsPresentationFactory.kt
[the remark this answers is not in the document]
ortem: Get rid of it. This heuristics will be introduced in later PR.

src/192/main/kotlin/org/rust/ide/hints/RsPlainTypeHint.kt
ortem: Let's make sure that new type hints cover all the functionality of the old ones instead.

src/192/main/kotlin/org/rust/ide/hints/RsInlayTypeHintsProvider.kt
ortem: These settings are copied from old hints so I think we should leave them in this PR

mchernyavsky:
bors retry

PR #4225, opened 2019-08-05
pretty printer: add core::ptr::NonNull

jacobrosenthal:
I dont have a simplified example, but this a no_std heapless datastructure that includes a core::ptr::NonNull and doesn't print much useful

```
(gdb) p data
$1 = heapless::pool::singleton::Box<nrf52_hal_common::uarte::DMAPool, heapless::pool::Init> {_pool: core::marker::PhantomData<nrf52_hal_common::uarte::
DMAPool>, inner: heapless::pool::Box<[u8; 16], heapless::pool::Init> {_state: core::marker::PhantomData<heapless::pool::Init>, node: core::ptr::NonNull
<heapless::pool::Node<[u8; 16]>> {pointer: 0x2000008c <interrupted::initLocals::new::MEMORY+140>}}}
```
with core::ptr::NonNull dereference its much better
```
(gdb) p data
$1 = heapless::pool::singleton::Box<nrf52_hal_common::uarte::DMAPool, heapless::pool::Init> = {
  _pool = core::marker::PhantomData<nrf52_hal_common::uarte::DMAPool>, inner = heapless::pool::Box<[u8; 16], heapless::pool::Init> = {
    _state = core::marker::PhantomData<heapless::pool::Init>, node = core::ptr::NonNull<heapless::pool::Node<[u8; 16]>> = {
      *pointer = heapless::pool::Node<[u8; 16]> = {data = core::cell::UnsafeCell<[u8; 16]> = {value = [97, 115, 100, 102, 97, 115, 100, 102, 97, 115, 100, 102, 97, 115, 100, 102]}, next = 0x20000078 <interrupted::initLocals::new::MEMORY+120>}}}}
```
Im not sure about lldb yet, but im sure could be implemented there if interested.


ortem:
Hi @jacobrosenthal,

I'd like to find out why you don't like the current presentation of `NonNull`. I find it useful because it shows the pointer address, not only the dereferenced value. 

For example, currently:
old

With `NonNull` pretty-printer:
new

IMO even in the case of console gdb, it is important to show the pointer address. For example, [C++ unique_ptr pretty-printer](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/python/libstdcxx/v6/printers.py#L179) shows the address and is used this way:
```
(gdb) p my_unique 
$1 = std::unique_ptr<S> = {get() = 0x555555768e70}
(gdb) p *my_unique 
$2 = {x = 0}
```

jacobrosenthal:
In my case I was looking for the lowest hanging fruit to be able to see the data* in this complex data structure. Without the dereference I dont get to see it 

*[97, 115, 100, 102, 97, 115, 100, 102, 97, 115, 100, 102, 97, 115, 100, 102]

jacobrosenthal:
Its .. alot but I can newtype it too.. again.. probably off topic, but always feels like a papercut
```


pub struct MyUarteDMAPoolNode {
    obj: UarteDMAPoolNode,
}


impl serde_cbor::ser::Write for MyUarteDMAPoolNode {
    type Error = fmt::Error;

    fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
        if buf.len() > UarteDMAPoolNode::MAX_SIZE {
            Err(fmt::Error)
        } else {
            self.obj.write_checked(buf);
            Ok(())
        }
    }
}

impl From<UarteDMAPoolNode> for MyUarteDMAPoolNode {
    fn from(obj: UarteDMAPoolNode) -> Self {
        MyUarteDMAPoolNode { obj }
    }
}

impl Into<UarteDMAPoolNode> for MyUarteDMAPoolNode {
    fn into(self) -> UarteDMAPoolNode {
        self.obj
    }
}




        let thing: MyUarteDMAPoolNode = UarteDMAPoolNode::new().into();
        let mut ser = Serializer::new(thing);
        &"aabasdf".serialize(&mut ser).unwrap();
        let b = UarteDMAPool::alloc().unwrap().init(ser.into_inner().into());
        resources.PRODUCER.enqueue(b).unwrap();
        rtfm::pend(interrupt::UARTE0_UART0);

```

ortem:
@jacobrosenthal I understand your problem, but I think your pretty-printer cannot be used as a default printer for `NonNull` in CLion because it doesn't show the address of the pointer and isn't applicable in cases when you actually need the address. But your printer is still useful in some cases like yours. Unfortunately, CLion doesn't support [adding custom pretty-printers](https://youtrack.jetbrains.com/issue/CPP-1798), though you can load your printers in GDB console manually.

jacobrosenthal:
Sounds good. Thanks for the look, and I look forward to you continuing to try to upstream your work back to Rust.

ortem:
@jacobrosenthal Thank you! Feel free to report problems related and contribute your ideas to improve Rust debugging.

Issue #1191, opened 2017-05-02
Handle `cfg` blocks in resolve


matklad:
When there are two items with different `cfgs`, like

```Rust
#[cfg(test)]
fn foo() {}

#[cfg(not(test))]
fn foo() {}

#[cfg(unix)]
fn bar() {}

#[cfg(not(unix))]
fn bar() {}
```

we should pick the non-test one matching the current platform. 

This should solve [this problem](https://github.com/intellij-rust/intellij-rust/blob/96fc0f81bd06f55604783e120500379509d7fcbe/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt#L438) in a principle way. Then this can be extended to fix https://github.com/intellij-rust/intellij-rust/issues/1189.

Pzixel:
@xIGBClutchIx #2166 is up to grab - this is an update, unfortunately...

Noah-Kennedy:
Any updates on this?

Undin:
@Noah-Kennedy the latest plugin version supports target-based cfg options listed in #4488 and features declared in `Cargo.toml` of project dependencies.
Support of features declared in workspace `Cargo.toml` is in progress right now.
Initial support of cfg options generated by `build.rs` is in progress too (see #4734).

But I think you are interested in intellisense features in `async-std` library. Name resolution, completion and so on don't work **not** because of `cfg`-related things. They don't work because plugin fails to expand macro calls in files that declared via module declaration inside another macro call, i.e. in the following case
`a.rs`
```rust
cfg_default! {
    mod b;
}
```
`b.rs`
```rust
cfg_default! {
    // some code
}
```
plugin fails to expand macro calls inside `b.rs`.
Of course, after macro expansion plugin should process `cfg` options in right way but looks like it's not a problem now.
The relevant issue about it is #4627.

Noah-Kennedy:
Thanks @Undin!

Sherlock-Holo:
the new engine is the default option, so is this problem solved too?

Undin:
@Sherlock-Holo no, they are completely different problems

Kobzol:
@Undin is there any ongoing work on cfg_attr or is it up for grabs?

Undin:
@Kobzol I don't know about any ongoing work on `cfg_attr` but let's ask @ortem as well
[...]

PR #4232, opened 2019-08-09
RES: Initial support of cfg attributes

ortem:
This PR is based on #2166 (but does not include #2840 and #3031) and adds [cfg attributes](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute) support to our name resolution. 

Part of #1191

@kumbayo thank you so much for your contribution! We are really sorry that we didn't merge your original PRs at a reasonable time. We've decided to merge the initial part of cfg support (related only to name resolution, without any UI changes) via this PR. So reviewing other cfg-related changes will become more convenient after this PR is merged.

Co-authored-by: Peter Oberndorfer <kumbayo84@arcor.de>

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
@@ -17,6 +17,8 @@ import com.intellij.psi.PsiFile
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.ancestorOrSelf
 import org.rust.lang.core.psi.ext.isTupleStruct
+import org.rust.lang.core.psi.ext.namedFields
+import org.rust.lang.core.psi.ext.positionalFields
 import org.rust.openapiext.checkWriteAccessAllowed
 
 class GenerateConstructorAction : CodeInsightAction() {
@@ -93,9 +95,9 @@ data class ConstructorArgument(
     companion object {
         fun fromStruct(structItem: RsStructItem): List<ConstructorArgument> {
             return if (structItem.isTupleStruct) {
-                fromTupleList(structItem.tupleFields?.tupleFieldDeclList.orEmpty())
+                fromTupleList(structItem.positionalFields)
             } else {
-                fromFieldList(structItem.blockFields?.namedFieldDeclList.orEmpty())
+                fromFieldList(structItem.namedFields)
             }
         }
 
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -5,9 +5,12 @@
 
 package org.rust.lang.core.psi.ext
 
+import com.intellij.openapi.util.registry.Registry
 import com.intellij.psi.NavigatablePsiElement
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.*
+import org.rust.lang.utils.evaluation.CfgEvaluator
+import org.rust.openapiext.isUnitTestMode
 
 interface RsDocAndAttributeOwner : RsElement, NavigatablePsiElement
 
@@ -89,6 +92,8 @@ class QueryAttributes(
             val stub = psi.greenStub
             if (stub != null) return stub.isCfg
         }
+        // TODO: We probably want this optimization also for other items that we query regularly
+
         return hasAttribute("cfg")
     }
 
@@ -152,8 +157,12 @@ class QueryAttributes(
     val deprecatedAttribute: RsMetaItem?
         get() = (attrsByName("deprecated") + attrsByName("rustc_deprecated")).firstOrNull()
 
+    val cfgAttributes: Sequence<RsMetaItem>
+        get() = attrsByName("cfg")
+
     // `#[attributeName = "Xxx"]`
-    private fun getStringAttributes(attributeName: String): Sequence<String?> = attrsByName(attributeName).map { it.value }
+    private fun getStringAttributes(attributeName: String): Sequence<String?> =
+        attrsByName(attributeName).map { it.value }
 
     val metaItems: Sequence<RsMetaItem>
         get() = attributes.mapNotNull { it.metaItem }
@@ -167,4 +176,27 @@ class QueryAttributes(
         "QueryAttributes(${attributes.joinToString { it.text }})"
 }
 
+/**
+ * Checks if there are any #[cfg()] attributes that disable this element
+ *
+ * HACK: do not check on [RsFile] as [RsFile.queryAttributes] would access the PSI
+ */
+val RsDocAndAttributeOwner.isEnabledByCfg: Boolean
+    get() {
+        // TODO: get rid of this registry key when cfg support is done
+        if (!isUnitTestMode && !Registry.`is`("rust.cfg.attributes.enabled")) return true
 
+        // TODO: add cfg to RsFile's stub and remove this line
+        if (this is RsFile) return true
+
+        if (!queryAttributes.hasCfgAttr()) return true
+        val cfgAttributes = queryAttributes.cfgAttributes.takeIf { it.any() } ?: return true
+
+        // TODO: When we open both cargo projects for an application and a library,
+        // this will return the library as containing package.
+        // When the application now requests certain features, which are not enabled by default in the library
+        // we will evaluate features wrongly here
+        val options = containingCargoPackage?.cfgOptions ?: return true
+
+        return CfgEvaluator(options).evaluate(cfgAttributes)
+    }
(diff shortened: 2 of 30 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsMetaItem.kt
```diff
@@ -17,10 +17,7 @@ import org.rust.lang.core.resolve.ref.RsReference
 import org.rust.lang.core.resolve.ref.deriveReference
 import org.rust.lang.core.stubs.RsMetaItemStub
 
-val RsMetaItem.name: String? get() {
-    val stub = greenStub
-    return if (stub != null) stub.name else identifier?.unescapedText
-}
+val RsMetaItem.name: String? get() = greenStub?.name ?: identifier?.unescapedText
```
vlad20012: This change causes test failures. Due to `RsMetaItemStub.name` is nullable, these constructions are not equivalent.

vlad20012:
I see a very significant problem that in my opinion prevents this feature to be released as-is.
Let's look at this example:
```rust
#[cfg(feature = "baz")]
fn foo() {}

#[cfg(feature = "baz")]
fn bar() {
    foo();
}
```

It's obvious that during the function `bar` analysis the feature `baz` is enabled. But with the current design, if `baz` feature is disabled globally, we will anyway analyze `bar` function but `foo` function call will be unresolved! 

Ideally, IDE should understand that inside the function `bar` the feature `baz` is enabled and provide an analysis (and name resolution) depending on this fact. But I can't imagine how it can be possible with our (or, really, any) caches. I.e. features should be global.

So I suggest this as a long-term solution:
1. Disable analysis for `bar`
2. Highlight code as *light-gray*, like CLion does
3. Provide an intention to enable missing features. I.e. alt+enter inside `bar` fn should be able to enable analysis for it. Of course, there should be UI to see/enable features.

But I don't want to postpone this PR merging for another year (like it was with the original PR), so I vote to merge this with described limitations BUT disable cfg evaluation for __everyone__ by default by some Registry key (for example). Then we have to decide on the features UI.

vlad20012:
Another problem that I see in the current design is that now it's not obvious whether we work with a raw PSI or with one remained after `cfg` evaluation. I mean, why `RsFieldsOwner.namedFields` / `RsFieldsOwner.positionalFields` returns fields filtered by `cfg` attributes? It's not obvious from its name or interface. Maybe it's time to introduce lowering/HIR which always will be after macro expansion & cfg evaluation.

src/main/kotlin/org/rust/lang/core/psi/ext/RsMetaItem.kt
ortem: Oh really, fixed

vlad20012:
Some tests are failder

src/test/kotlin/org/rust/RsTestCase.kt
```diff
@@ -15,6 +15,24 @@ interface RsTestCase {
...
+            "target_has_atomic=\"ptr\"",
+            "target_arch=\"x86_64\"",
+            "target_endian=\"little\"",
+            "target_env=\"gnu\"",
+            "target_family=\"unix\"",
+            "target_os=\"linux\"",
+            "target_pointer_width=\"64\"",
+            "unix",
+            "feature=\"use_std\"" /*Not enabled normally*/)
+        val testDefaultCfg = testDefaultCfgStdLib.toMutableList() + "test"
```
vlad20012: Why "test" by default?

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
```diff
@@ -62,6 +62,9 @@ fun processItemDeclarations(
     val cachedItems = scope.expandedItemsCached
 
     loop@ for (item in cachedItems.rest) {
+        // Ignore items (and their children) deactivated by #[cfg(...)]
+        if (!item.isEnabledByCfg) continue@loop
```
vlad20012: This isn't needed because of the same check inside `RsElement.processItem`

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -89,6 +92,8 @@ class QueryAttributes(
             val stub = psi.greenStub
             if (stub != null) return stub.isCfg
         }
+        // TODO: We probably want this optimization also for other items that we query regularly
```
vlad20012: Seems reasonable:
image

Looks like these items are queried most often:
image

Also, let's add the same flag to `RsImplItem` stub, it will be used later in `RsImplIndex`, and `RsModDecl`

src/test/kotlin/org/rust/lang/core/type/RsStdlibExpressionTypeInferenceTest.kt
```diff
@@ -770,4 +770,19 @@ class RsStdlibExpressionTypeInferenceTest : RsTypificationTestBase() {
             c;
         } //^ String
     """)
+
+    // During tests we set some hardcoded cfg values like unix
```
vlad20012: Maybe introduce some annotation instead of hardcode?

src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
```diff
@@ -1330,4 +1332,54 @@ class RsResolveTest : RsResolveTestBase() {
                      //^
         }
     """)
+
+    // During our tests, #[cfg(test)] evaluates to true
+    fun `test fn cfg`() = checkByCode("""
```
vlad20012: Let's move these tests to a separate class like `RsCfgAttrResolveTest`.
Also, let's add tests with `cfg`ed mod decl, inline mod, use item, extern crate (with alias), macro call that defines some items

src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
```diff
@@ -114,10 +114,15 @@ private fun RsItemsOwner.processExpandedItemsInternal(processor: (RsItemElement)
     return itemsAndMacros.any { it.processItem(processor) }
 }
 
-private fun RsElement.processItem(processor: (RsItemElement) -> Boolean) = when (this) {
-    is RsMacroCall -> processExpansionRecursively { it is RsItemElement && processor(it) }
-    is RsItemElement -> processor(this)
-    else -> false
+private fun RsElement.processItem(processor: (RsItemElement) -> Boolean): Boolean {
+    // TODO Is this necessary?
```
vlad20012: It is

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -152,19 +157,47 @@ class QueryAttributes(
...
+        attrsByName(attributeName).map { it.value }
 
     val metaItems: Sequence<RsMetaItem>
         get() = attributes.mapNotNull { it.metaItem }
 
     /**
      * Get a sequence of all attributes named [name]
      */
-    private fun attrsByName(name: String): Sequence<RsMetaItem> = metaItems.filter { it.name == name }
+    private fun attrsByName(name: String): Sequence<RsMetaItem> = metaItems.filter { it.name == name } // TODO: Use HashMap instead?
```
vlad20012: Remove TODO

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -152,19 +157,47 @@ class QueryAttributes(
...
 
     override fun toString(): String =
         "QueryAttributes(${attributes.joinToString { it.text }})"
 }
 
+/**
+ * Checks if there are any #[cfg()] attributes that disable this element
+ *
+ * HACK: do not check on [RsFile] as [RsFile.queryAttributes] would access the PSI
+ * TODO: cache?
```
vlad20012: Remove TODO. Should be cached (already) on the upper level (resolve cache)

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -93,6 +93,7 @@ data class ConstructorArgument(
     companion object {
         fun fromStruct(structItem: RsStructItem): List<ConstructorArgument> {
             return if (structItem.isTupleStruct) {
+                //TODO #[cfg()] on tupleFieldDeclList items
```
vlad20012: Why not implement it now?

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
```diff
@@ -221,6 +221,7 @@ private fun processMultiResolveWithNs(name: String, ns: Set<Namespace>, ref: RsR
 
     // XXX: there are two `cfg`ed `boxed` modules in liballoc, so
     // we apply "first in the namespace wins" heuristic.
+    // TODO: How to change this now that we handle #[cfg(...)] attributes?
```
vlad20012:
```suggestion
    // TODO fix after enabling #[cfg(...)] evaluation
```

vlad20012:
LGTM now
bors r+

Note that these changes are very internal and don't affect anything outside of our unit tests

vlad20012:
bors retry

PR #4235, opened 2019-08-10
ANN: Help the user fixing too big source file problems

### LINKED ISSUE #2297: Getting E0463 for stm32f103xx
I am currently experimenting with embedded arm and intellij cannot resolve the crate stm32f103xx / https://crates.io/crates/stm32f103xx while other crates like cortex_m, cortex_m_rt and cortex_m_semihosting work fine.

For stm32f103xx I get: https://doc.rust-lang.org/error-index.html#E0463

I am using the cortex quickstart project (https://github.com/japaric/cortex-m-quickstart) and
added the dependency 
```
[dependencies.stm32f103xx]
version = "0.8.0"
```
and "extern crate stm32f103xx;".

The project compiles and works on the arm processor (although one needs to compile it with xargo like "rustup run nightly xargo build").
Uploading the whole project would be clunky, but if you need it, I can upload it for you.

mibac138:
Resolves #2297. I didn't want to make the error message too long so I just added a link to a [solution](https://github.com/intellij-rust/intellij-rust/issues/2297#issuecomment-366326827) to the problem. I hope that's okay.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -326,6 +326,24 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
             }
         }
 
+        if (path.parent is RsUseSpeck && // For now check just `use` statements
+            path.reference.multiResolve().isEmpty() &&
+            path.containingCargoPackage?.origin == PackageOrigin.WORKSPACE) {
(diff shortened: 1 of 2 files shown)


Kobzol:
@vlad20012 would you accept this feature? If yes, I'd try to finish/rebase it.

vlad20012:
The feature is acceptable, but the implementation is too naive. Also, it works only for unresolved crate roots.
I'd say, the reason for the fact that some reference is not resolved should be recorded during name resolution (somewhere in `ModCollector`/`DefCollector`) and then be somehow extracted in `RsUnresolvedReferenceInspection`.
This really doesn't look trivial. I can ask @dima74 to think of something

PR #4238, opened 2019-08-11
GRD: PsiViewer plugin dependency

mchernyavsky:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,8 @@ val baseVersion = when (baseIDE) {
     else -> error("Unexpected IDE name: `$baseIDE`")
 }
 
+val psiViewerPluginVersion = prop("psiViewerPluginVersion")
+
 val isAtLeast192 = platformVersion.toInt() >= 192
 
 plugins {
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -158,7 +158,11 @@ project(":plugin") {
     version = "0.2.$patchVersion.${prop("buildNumber")}$versionSuffix"
     intellij {
         pluginName = "intellij-rust"
-        val plugins = mutableListOf(project(":intellij-toml"), "IntelliLang")
+        val plugins = mutableListOf(
+            project(":intellij-toml"),
+            "IntelliLang",
+            "PsiViewer:$platformVersion-SNAPSHOT"
```
Undin: Let's avoid relying on some accident relation between platform and plugin versions.
For example, the latest version for 191 is `191.4212` but not `191-SNAPSHOT`
I suggest adding `psiViewerPluginVersion` property to `gradle-%platformVersion%.properties` and use `prop("psiViewerPluginVersion")`

mchernyavsky: Done.

PR #4252, opened 2019-08-13
COMP: keyword completion in trait and impl items

### LINKED ISSUE #4249: Keyword completion in trait and impl items
## Environment

* **IntelliJ Rust plugin version:** 0.2.103.2707-192-nightly
* **Rust toolchain version:** 1.38.0-nightly (d3f8a0b5d 2019-08-04) x86_64-apple-darwin
* **IDE name and version:** IU-192.5728.98

## Problem description
Currently, the plugin provides keyword completion like `fn`, `const`, `type`, etc at top level of a module. But if I want to add some item inside trait or impl block, there isn't similar completion. It would be great to implement it as well.

t-kameyama:
Fixes #4249

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
@@ -15,20 +15,19 @@ import com.intellij.patterns.PlatformPatterns.psiElement
 import com.intellij.patterns.PsiElementPattern
 import com.intellij.patterns.StandardPatterns.or
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiErrorElement
 import com.intellij.psi.PsiWhiteSpace
 import com.intellij.util.ProcessingContext
-import org.rust.lang.core.RsPsiPattern
+import org.rust.lang.core.*
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.*
 import org.rust.lang.core.psi.ext.*
-import org.rust.lang.core.psiElement
 import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.infer.lookupFutureOutputTy
 import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
-import org.rust.lang.core.withPrevSiblingSkipping
-import org.rust.lang.core.withSuperParent
+import kotlin.or
 
 /**
  * Completes Rust keywords
@@ -56,6 +55,8 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
             RsKeywordCompletionProvider("where"))
         extend(CompletionType.BASIC, constParameterBeginningPattern(),
             RsKeywordCompletionProvider("const"))
+        extend(CompletionType.BASIC, traitOrImplDeclarationPattern(),
+            RsKeywordCompletionProvider(*COMPLETION_KEYWORDS_IN_TRAIT_OR_IMPL))
 
         extend(CompletionType.BASIC, elsePattern(), object : CompletionProvider<CompletionParameters>() {
             override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -142,7 +142,12 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
 
     private fun baseDeclarationPattern(): PsiElementPattern.Capture<PsiElement> =
         psiElement()
-            .withParent(or(psiElement<RsPath>(), psiElement<RsModItem>(), psiElement<RsFile>()))
+            .withParent(or(
+                psiElement<RsPath>(),
+                psiElement<RsModItem>(),
+                psiElement<RsMembers>().withParent(psiElement<RsTraitOrImpl>()),
```
Undin: I think we shouldn't try to change `baseDeclarationPattern` because it covers all necessary cases for top-level declarations.
We need a separate completion provider for `trait` and `impl` blocks because only several items can be placed inside them. For example, we can't create `enum` inside `trait` block.
The current implementation can suggest invalid keywords
image

src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
```diff
@@ -589,6 +589,30 @@ class RsKeywordCompletionContributorTest : RsCompletionTestBase() {
         "fn foo<T /*caret*/>() {}"
     """)
 
+    fun `test const inside trait`() = checkCompletion("const", """
```
Undin: Let's add a test to check that we don't suggest `enum`, `trait`, etc keywords inside `trait` & `impl` blocks

src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
```diff
@@ -589,6 +589,30 @@ class RsKeywordCompletionContributorTest : RsCompletionTestBase() {
...
+        pub trait Bar {
+            /*caret*/
+        }
+    ""","""
+        pub trait Bar {
+            const /*caret*/
+        }
+    """)
+
+    fun `test const inside impl`() = checkCompletion("const", """
```
Undin: Could you please add tests for other keywords (not only for `const`)?

Undin: `org.rust.lang.core.completion.RsKeywordCompletionContributorTest#checkCompletion(java.util.List<java.lang.String>, java.lang.String, java.lang.String)` can be usefull for that

src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -238,6 +240,10 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
         return psiElement(IDENTIFIER).withParent(parent)
     }
 
+    private fun traitOrImplPattern(): PsiElementPattern.Capture<PsiElement> {
+        return psiElement().inside(RsTraitOrImpl::class.java).and(statementBeginningPattern())
```
Undin: `inside` iterates over all parents of psi element. It leads to:
* pattern accepts more places than desired
* pattern make much more work than really needed
Let's make a more precise pattern. Looks like we can just check that our parent is `RsMembers`

src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -238,6 +240,10 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
         return psiElement(IDENTIFIER).withParent(parent)
     }
 
+    private fun traitOrImplPattern(): PsiElementPattern.Capture<PsiElement> {
```
Undin: I think `trairOrImplDeclarationPattern` is better name

src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
```diff
@@ -589,6 +589,82 @@ class RsKeywordCompletionContributorTest : RsCompletionTestBase() {
         "fn foo<T /*caret*/>() {}"
     """)
 
+    fun `test inside trait`() = checkCompletion(listOf("const", "fn", "type"), """
```
Undin: Let's extract this list into companion object

src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -238,9 +239,19 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
         return psiElement(IDENTIFIER).withParent(parent)
     }
 
+    private fun traitOrImplDeclarationPattern(): PsiElementPattern.Capture<PsiElement> {
+        return psiElement().withParent(
+            or(
+                psiElement<RsMembers>(),
+                psiElement<PsiErrorElement>().withParent(psiElement<RsMembers>())
```
t-kameyama: For [test inside trait after statement](https://github.com/t-kameyama/intellij-rust/blob/d96e4c3d19ec3e76659a1179167a3fa359450520/src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt#L605), [test inside impl after statement](https://github.com/t-kameyama/intellij-rust/blob/d96e4c3d19ec3e76659a1179167a3fa359450520/src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt#L645)

Issue #4256, opened 2019-08-14
IllegalStateException: Too few arguments for option `--package`


lancelote:
## Environment

* **IntelliJ Rust plugin version:** master build at ec7591c283e14d5db307f5fa6c8d996060865d1a
* **Rust toolchain version:** rustc 1.36.0 (a53f9df32 2019-07-03)
* **IDE name and version:** CLion 2019.2 192.5728.100
* **Operating system:** macOS 10.14.6

## Problem description

Exception is raised when I edit a package name in a run configuration.

## Steps to reproduce

1. Have some run configuration for binary package, e.g.

<img width="1083" alt="config" src="https://user-images.githubusercontent.com/2500320/63017041-e30e0300-be9d-11e9-95bc-2a48345724d0.png">

2. Edit the run configuration by deleting the package name and retyping it again (e.g. due to the package rename the configuration is invalid)
3. Exception is raised

```
java.lang.IllegalStateException: Too few arguments for option `--package`
	at org.rust.cargo.util.CargoArgsParser.getActualOptionArgsCount(CargoArgsParser.kt:50)
	at org.rust.cargo.util.CargoArgsParser.splitArgs(CargoArgsParser.kt:26)
	at org.rust.cargo.util.CargoArgsParser.access$splitArgs(CargoArgsParser.kt:8)
	at org.rust.cargo.util.CargoArgsParser$Companion.parseRunArgs(CargoArgsParser.kt:64)
	at org.rust.cargo.util.CargoArgsParser$Companion.parseArgs(CargoArgsParser.kt:57)
	at org.rust.cargo.runconfig.buildtool.CargoBuildManager.getBuildConfiguration(CargoBuildManager.kt:203)
	at org.rust.cargo.runconfig.legacy.RsAsyncRunner.canRun(RsAsyncRunner.kt:58)
	at com.intellij.execution.runners.ProgramRunner.getRunner(ProgramRunner.java:49)
	at com.intellij.execution.impl.SingleConfigurationConfigurable.getValidationResult(SingleConfigurationConfigurable.java:188)
	at com.intellij.execution.impl.SingleConfigurationConfigurable.access$1400(SingleConfigurationConfigurable.java:48)
	at com.intellij.execution.impl.SingleConfigurationConfigurable$MyValidatableComponent.getValidationResult(SingleConfigurationConfigurable.java:458)
[...]

PR #4259, opened 2019-08-14
RUN: Fix exception in "Run/Debug Configurations" menu

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4256.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/util/CargoArgsParser.kt
+++ b/src/main/kotlin/org/rust/cargo/util/CargoArgsParser.kt
@@ -41,14 +41,12 @@ class CargoArgsParser private constructor(
     private fun getActualOptionArgsCount(optionIdx: Int): Int {
         val option = cargoArgs[optionIdx]
         // Unknown options are assumed to be the flagging options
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/util/CargoArgsParser.kt
```diff
@@ -47,11 +49,15 @@ class CargoArgsParser private constructor(
             .takeWhile { !OPTION_NAME_RE.matches(it) }
             .take(maxArgsCount)
             .count()
-        check(optionArgsCount >= minArgsCount) { "Too few arguments for option `$option`" }
+        if (optionArgsCount < minArgsCount) {
+            LOG.warn("Too few arguments for option `$option`")
```
ortem: `getActualOptionArgsCount` is called from `getBuildConfiguration`, which is called from `canRun` methods of our runners (e.g. `org.rust.cargo.runconfig.RsExecutableRunner#canRun`). So this warning will be printed so many times.

Also, it's not a very good idea to do such computations during `canRun` which is called every moment.

PR #4263, opened 2019-08-15
PERF: use `strict = false` in `RsResolveCache`

vlad20012:
Previous (default) `strict = true` value really isn't needed, and just slows down performance in the case of `RsModDecl` reference cache


mchernyavsky:
bors retry

PR #4291, opened 2019-08-21
Make "Navigate Class/Symbol" works with items expanded from macros

vlad20012:
Screenshot from 2019-08-21 13-54-28

Note that looks like I already enabled search for items expanded from macros in the case of "All places" scope in #4275. But this PR enables it correctly for all scopes (e.g. project/library files)

### DIFF
--- a/src/main/kotlin/org/rust/ide/navigation/goto/RsNavigationContributorBase.kt
+++ b/src/main/kotlin/org/rust/ide/navigation/goto/RsNavigationContributorBase.kt
@@ -8,7 +8,10 @@ package org.rust.ide.navigation.goto
 import com.intellij.navigation.ChooseByNameContributorEx
 import com.intellij.navigation.GotoClassContributor
 import com.intellij.navigation.NavigationItem
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
+import com.intellij.psi.search.EverythingGlobalScope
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/navigation/goto/RsNavigationContributorBase.kt
```diff
@@ -28,19 +32,32 @@ abstract class RsNavigationContributorBase<T> protected constructor(
     GotoClassContributor where T : NavigationItem, T : RsNamedElement {
 
     override fun processNames(processor: Processor<String>, scope: GlobalSearchScope, filter: IdFilter?) {
-        StubIndex.getInstance().processAllKeys(indexKey, processor, scope, filter)
+        StubIndex.getInstance().processAllKeys(
```
ortem: Let's add something like
```Kotlin
if (ApplicationManager.getApplication().isInternal) {
    check(filter == null) { "IdFilter is supposed to be null" }
}
```

lancelote:
Hm, I was sure I did comment on this 🤔 Anyway, @vlad20012 is goto declaration expected to work in the following case:

```rust
macro_rules! foo {
    () => {
        struct FooBarBaz;
    };
}

fn main() {
    foo!();
}
```

Or is it too much?

vlad20012:
@lancelote too much :)
Macros inside function bodies are not indexed. In theory, we can simply enable it, but I expect a significant performance regression in this case.

Issue #4292, opened 2019-08-21
Unable to debug with command line arguments when also specifying which features to use in Run/Debug Configurations.


Delamare2112:
## Environment

* **IntelliJ Rust plugin version:**  0.2.103.2131-192
* **Rust toolchain version:** 1.34.0
* **IDE name and version:** CLion 2019.2
* **Operating system:** Linux

## Problem description
It is not possible to debug applications with command line parameters when the "Command" Run/Debug Configurations field contains "--features".  Running them works fine.

Example:
I am working through ash Vulkan tutorials and wanted to implement Vulkan validation as a feature.  I also want to pass in the path to the shaders directory as an optional command line argument.  I also have `-c` as another optional command line switch.
So for example: `run --package ash_tut --bin ash_tut --features vk-validation  -- -s ./shaders -c`  
But when I use this as the "Command" field for a Cargo Run/Debug Configuration, I get the following:
```
/usr/bin/cargo build --color=always --package ash_tut --bin ash_tut --features vk-validation -- -s -c
error: Found argument '-s' which wasn't expected, or isn't valid in this context

USAGE:
    cargo build --bin <NAME>... --color <WHEN> --features <FEATURES> --package <SPEC>...

For more information try --help
```
If I remove `--features vk-validation` or my command line arguments, debug works fine.  
It also appears that my shader path has been removed from the command sent to cargo for some reason.  

Cargo.toml:
```
...
[features]
vk-validation = []
...
```

PR #4296, opened 2019-08-22
RUN: Fix parsing option arguments

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4292.


src/main/kotlin/org/rust/cargo/util/CargoArgsParser.kt
```diff
@@ -44,7 +44,7 @@ class CargoArgsParser private constructor(
         val (_, maxArgsCount) = optionsToArgsCountRange.getOrDefault(option, OptionArgsCountRange.ZERO)
         return cargoArgs.asSequence()
             .drop(optionIdx + 1)
-            .takeWhile { !OPTION_NAME_RE.matches(it) }
+            .takeWhile { it != "--" && !OPTION_NAME_RE.matches(it) }
```
Undin: @mchernyavsky maybe it's worth to add test for this case?

PR #4301, opened 2019-08-23
INT: add "Put fields on separate lines/one line" intentions

### LINKED ISSUE #4282: "Put fields on separate lines/one line" intentions
It would be great to have such intentions to convert
```Rust
struct S { x: i32, y: i32 }
```
into
```Rust
struct S {
    x: i32,
    y: i32
}
```
as vice versa.

For example, similar feature #3279

t-kameyama:
Fixes #4282

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/JoinListIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/JoinListIntention.kt
@@ -5,13 +5,12 @@
 
 package org.rust.ide.intentions
 
+import com.intellij.openapi.editor.Document
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
-import org.rust.lang.core.psi.RsExpr
-import org.rust.lang.core.psi.RsValueArgumentList
-import org.rust.lang.core.psi.RsValueParameter
-import org.rust.lang.core.psi.RsValueParameterList
+import com.intellij.psi.PsiWhiteSpace
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.endOffset
 import org.rust.lang.core.psi.ext.startOffset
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/rust/ide/intentions/JoinListIntention.kt
```diff
@@ -33,12 +33,18 @@ abstract class JoinListIntentionBase<TList : RsElement, TElement : RsElement>(
...
+    }
 
-        nextBreak(elements.last())?.let { document.deleteString(it.startOffset, it.endOffset) }
-        elements.dropLast(1).asReversed().forEach {
-            nextBreak(it)?.let { document.replaceString(it.startOffset, it.endOffset, " ") }
+    private fun PsiWhiteSpace.replace(replaceString: String, document: Document) {
+        if (replaceString.isEmpty()) {
+            document.deleteString(startOffset, endOffset)
+        } else {
+            document.replaceString(startOffset, endOffset, " ")
```
ortem:
```suggestion
            document.replaceString(startOffset, endOffset, replaceString)
```

src/main/kotlin/org/rust/ide/intentions/JoinListIntention.kt
```diff
@@ -5,21 +5,21 @@
...
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.endOffset
 import org.rust.lang.core.psi.ext.startOffset
 
 abstract class JoinListIntentionBase<TList : RsElement, TElement : RsElement>(
     listClass: Class<TList>,
     elementClass: Class<TElement>,
-    intentionText: String
+    intentionText: String,
+    private val isNeedSpaceWithinParen: Boolean = false
```
ortem: I'm not sure how clear this name is. Maybe use two parameters
```Kotlin
    private val prefix: String = "",
    private val suffix: String = ""
```
instead? Or maybe rename it somehow to make clear that it is only about leading and trailing spaces.

ortem:
Thank you!

bors r+

PR #4305, opened 2019-08-24
ANN: cleanup missing fields annotations

Undin:


Undin:
@shevtsiv I've dropped declaration text from annotation text to avoid the following
image
I think we can just add name of missing fields like the compiler does
```
error[E0027]: pattern does not mention fields `x`, `y`
 --> src/main.rs:7:9
  |
7 |     let Point {} = p1;
  |         ^^^^^^^^ missing fields `x`, `y`
```

PR #3744, opened 2019-04-22
PERF: slightly optimize name resolution (better cache)

vlad20012:
According to our tests, this increases highlighting performance up to 15%

PR #4309, opened 2019-08-26
RES: Fix random unresolved refs; drop per-mod name cache

vlad20012:
I ran into a very weird bug when some name is resolved but then randomly unresolved after some typing. After a long investigation (it was hard to even reduce it to some test) I found that that bug is caused by the cache introduced in #3744. The cache is wrong in general b/c with the current name resolution engine names visible in scope are caller-dependent because of infinite recursions possible with star imports (prevented by a recursion guard). 

Good news is that performance is about 5%. When the cache was introduced it was about 15%. I think this difference is caused by our recent optimizations, so this cache is no longer needed anyway.

(Btw #4262 brings 5% optimization, so if we put them together into one release, we will not have regression :)


vlad20012:
Hmm, looks like dropping this cache doesn't fix all my issues. BUT downgrading to 2019.1 fixes them all. I think I should investigate it more

PR #4313, opened 2019-08-27
COMP: provide fragment specifiers completion

### LINKED ISSUE #4284: Complete fragment-specifiers in macro
IDE should suggest fragment-specifiers in completion list. 
https://doc.rust-lang.org/reference/macros-by-example.html#metavariables
image

## Environment

* **Rust toolchain version:**
* **IDE name and version:**
* **Operating system:**

t-kameyama:
Fixes #4284

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/FragmentKind.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/FragmentKind.kt
@@ -13,20 +13,20 @@ import org.rust.lang.core.psi.RS_KEYWORDS
 import org.rust.lang.core.psi.RsElementTypes
 import org.rust.lang.core.psi.tokenSetOf
 
-enum class FragmentKind {
-    Ident,
-    Path,
-    Expr,
-    Ty,
-    Pat,
-    Stmt,
-    Block,
-    Item,
-    Meta,
-    Tt,
-    Vis,
-    Literal,
-    Lifetime;
+enum class FragmentKind(private val kind: String) {
+    Ident("ident"),
+    Path("path"),
+    Expr("expr"),
+    Ty("ty"),
+    Pat("pat"),
+    Stmt("stmt"),
+    Block("block"),
+    Item("item"),
+    Meta("meta"),
+    Tt("tt"),
+    Vis("vis"),
+    Literal("literal"),
+    Lifetime("lifetime");
 
     fun parse(builder: PsiBuilder): Boolean {
         return if (this == Ident) {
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsFragmentSpecifierCompletionTest.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.completion
+
+import org.intellij.lang.annotations.Language
+import org.rust.lang.core.macros.FragmentKind
+
+class RsFragmentSpecifierCompletionTest : RsCompletionTestBase() {
```
vlad20012: Let's add a test that checks that we don't complete specifiers right after `$` sign, i.e. `$/*caret*/`

src/test/kotlin/org/rust/lang/core/completion/RsFragmentSpecifierCompletionTest.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun `test fragment specifiers`() = doTest("""
+        macro_rules! foo {
+            (${"$"}l:/*caret*/) => {}
```
vlad20012:
```suggestion
            ($ l:/*caret*/) => {}
```
(note space after `$`)

src/main/kotlin/org/rust/lang/core/macros/FragmentKind.kt
```diff
@@ -84,22 +84,11 @@ enum class FragmentKind {
...
-            "block" -> Block
-            "item" -> Item
-            "meta" -> Meta
-            "vis" -> Vis
-            "tt" -> Tt
-            "lifetime" -> Lifetime
-            "literal" -> Literal
-            else -> null
-        }
+        private val fragmentKinds = values().associateBy { it.kind }
```
ortem:
```suggestion
        private val fragmentKinds: Map<String, FragmentKind> = values().associateBy { it.kind }
```

src/test/kotlin/org/rust/lang/core/completion/RsFragmentSpecifierCompletionTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    fun `test fragment specifiers after $ sign`() = checkNotContains("""
+        macro_rules! foo {
+            ($/*caret*/) => {}
+        }
+    """)
+
+    @Suppress("SameParameterValue")
+    private fun checkContains(@Language("Rust") text: String) {
+        for (fragmentKind in FragmentKind.kinds) {
```
vlad20012: Oh. Is it not possible to implement it without such loops? With it, you invoke parsing and completion `FragmentKind.kinds.size` times instead of one.

(I know there is the same code in `RsPartialMacroArgumentCompletionTest`, but there are a very few numbers of elements, mostly 1)

t-kameyama: I referred to:
https://github.com/intellij-rust/intellij-rust/blob/d640c47d97ac9bbad2cb1d3876e74104a38e9a3c/src/test/kotlin/org/rust/lang/core/completion/RsPrimitiveTypeCompletionTest.kt#L23-L28

t-kameyama: Should I fix?

ortem: @t-kameyama OK, I think we can merge it as is and try to fix both tests later!

PR #4315, opened 2019-08-27
ANN: Improve error messages for E0023 and E0027

shevtsiv:
For more info see https://github.com/intellij-rust/intellij-rust/pull/4305#issuecomment-524560326

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -13,6 +13,7 @@ import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.openapi.util.TextRange
+import com.intellij.openapi.util.text.StringUtil.pluralize
 import com.intellij.psi.PsiElement
 import com.intellij.xml.util.XmlStringUtil.escapeString
 import org.rust.ide.annotator.RsErrorAnnotator
@@ -1105,27 +1106,35 @@ sealed class RsDiagnostic(
             )
     }
 
-    class MissingFieldsInTuplePattern(pat: RsPat, private val declaration: RsFieldsOwner) : RsDiagnostic(pat) {
+    class MissingFieldsInTuplePattern(
+        pat: RsPat,
+        private val declaration: RsFieldsOwner,
+        private val expectedAmount: Int,
+        private val actualAmount: Int
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Tuple struct"
             return PreparedAnnotation(
                 ERROR,
                 E0023,
-                // TODO: provide number of expected/actual fields
-                "$itemType pattern does not correspond to its declaration",
+                "$itemType pattern does not correspond to its declaration: expected $expectedAmount ${pluralize("field", expectedAmount)}, found $actualAmount",
                 fixes = listOf(AddStructFieldsPatFix(element))
             )
         }
     }
 
-    class MissingFieldsInStructPattern(pat: RsPat, private val declaration: RsFieldsOwner) : RsDiagnostic(pat) {
+    class MissingFieldsInStructPattern(
+        pat: RsPat,
+        private val declaration: RsFieldsOwner,
+        private val missingFields: List<RsFieldDecl>
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Struct"
+            val missingFieldNames = escapeString(missingFields.joinToString(", ") { "`${it.name!!}`" })
             return PreparedAnnotation(
                 ERROR,
                 E0027,
-                // TODO: provide name of missing fields
-                "$itemType pattern does not correspond to its declaration",
+                "$itemType pattern does not mention ${pluralize("field", missingFields.size)} $missingFieldNames",
                 fixes = listOf(AddStructFieldsPatFix(element))
             )
         }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1105,27 +1106,37 @@ sealed class RsDiagnostic(
...
-    class MissingFieldsInTuplePattern(pat: RsPat, private val declaration: RsFieldsOwner) : RsDiagnostic(pat) {
+    class MissingFieldsInTuplePattern(
+        pat: RsPat,
+        private val declaration: RsFieldsOwner,
+        private val expectedAmount: Int,
+        private val actualAmount: Int
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Tuple struct"
+            val fieldOrFields = if (expectedAmount == 1) "field" else "fields"
```
Undin: `com.intellij.openapi.util.text.StringUtil#pluralize(java.lang.String)` can be usefull here

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1105,27 +1106,37 @@ sealed class RsDiagnostic(
...
-    class MissingFieldsInStructPattern(pat: RsPat, private val declaration: RsFieldsOwner) : RsDiagnostic(pat) {
+    class MissingFieldsInStructPattern(
+        pat: RsPat,
+        private val declaration: RsFieldsOwner,
+        private val missingFields: List<RsFieldDecl>
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Struct"
+            val missingFieldNames = missingFields.map { "`${it.name!!}`" }.stream().collect(Collectors.joining(", "))
+            val fieldOrFields = if (missingFieldNames.length == 1) "field" else "fields"
```
Undin: And here too

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
Undin: And also I think it's worth to inline this expression to avoid weird variable name

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1105,27 +1106,37 @@ sealed class RsDiagnostic(
...
 
-    class MissingFieldsInStructPattern(pat: RsPat, private val declaration: RsFieldsOwner) : RsDiagnostic(pat) {
+    class MissingFieldsInStructPattern(
+        pat: RsPat,
+        private val declaration: RsFieldsOwner,
+        private val missingFields: List<RsFieldDecl>
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Struct"
+            val missingFieldNames = missingFields.map { "`${it.name!!}`" }.stream().collect(Collectors.joining(", "))
```
Undin:
```suggestion
            val missingFieldNames = missingFields.joinToString(", ") { "`${it.name!!}`" }
```

Undin: Also, could you check that `it.name!!` doesn't fail with incomplete code (the corresponding test will be helpful here as well)

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1105,27 +1106,37 @@ sealed class RsDiagnostic(
...
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Tuple struct"
+            val fieldOrFields = if (expectedAmount == 1) "field" else "fields"
             return PreparedAnnotation(
                 ERROR,
                 E0023,
-                // TODO: provide number of expected/actual fields
-                "$itemType pattern does not correspond to its declaration",
+                "$itemType pattern does not correspond to its declaration: Expected $expectedAmount $fieldOrFields, found $actualAmount",
```
Undin:
```suggestion
                "$itemType pattern does not correspond to its declaration: expected $expectedAmount $fieldOrFields, found $actualAmount",
```

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
shevtsiv: > Also, could you check that `it.name!!` doesn't fail with incomplete code (the corresponding test will be helpful here as well)?

I have added a test for error analysis on invalid syntax but I'm not sure if it is correct solution so could you recheck it, please?
https://github.com/intellij-rust/intellij-rust/pull/4315/files#diff-4c22a3080d6a06f7e6d2b6a38f801ab0R2899-R2905

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1105,27 +1106,35 @@ sealed class RsDiagnostic(
...
+    ) : RsDiagnostic(pat) {
         override fun prepare(): PreparedAnnotation {
             val itemType = if (declaration is RsEnumVariant) "Enum variant" else "Struct"
+            val missingFieldNames = escapeString(missingFields.joinToString(", ") { "`${it.name!!}`" })
             return PreparedAnnotation(
                 ERROR,
                 E0027,
-                // TODO: provide name of missing fields
-                "$itemType pattern does not correspond to its declaration",
+                "$itemType pattern does not mention following ${pluralize("field", missingFields.size)}: $missingFieldNames",
```
Undin:
```suggestion
                "$itemType pattern does not mention the following ${pluralize("field", missingFields.size)}: $missingFieldNames",
```

Undin: Or drop `following` at all. It will simplify error message and will be the same as the compiler message

PR #4324, opened 2019-08-29
RES: enable all cfg features by default

vlad20012:
See #4232
Since cargo features should be additive, we can just enable them all for now. At least it should work better than disabling them all.


vlad20012:
bors r=ortem

vlad20012:
bors retry

Issue #4319, opened 2019-08-28
`VirtualFileManagerImpl.findFileByUrl` without read action causes exception


vlad20012:
## Environment

* **IntelliJ Rust plugin version:** 0.2.104.2718-192-nightly
* **Rust toolchain version:** 1.38.0-nightly (8a58268b5 2019-07-31) x86_64-unknown-linux-gnu
* **IDE name and version:** IU-192.6262.58
* **Operating system:** Linux 5.0.0-21-generic

## Problem description

Looks like we shouldn't use `findFileByUrl` without the read action because it can lead to an exception:

```
com.intellij.openapi.vfs.InvalidVirtualFileAccessException: Accessing invalid virtual file: file:///home/loomaclin/.IdeaIC2019.2/system/rust_expanded_macros/hRKfsZeW/expansions/4; original:80561; found:80561; File.exists()=false
  at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.doFindChild(VirtualDirectoryImpl.java:109)
  at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:75)
  at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:451)
  at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:45)
  at com.intellij.openapi.vfs.newvfs.VfsImplUtil.findFileByPath(VfsImplUtil.java:52)
  at com.intellij.openapi.vfs.impl.local.LocalFileSystemBase.findFileByPath(LocalFileSystemBase.java:46)
  at com.intellij.openapi.vfs.impl.VirtualFileManagerImpl.findFileByUrl(VirtualFileManagerImpl.java:157)
  at org.rust.lang.core.macros.SerializedExpandedMacroInfo.toExpandedMacroInfo(ExpandedMacroStorage.kt:590)
  at org.rust.lang.core.macros.SourceFile$Companion.readFrom(ExpandedMacroStorage.kt:514)
  at org.rust.lang.core.macros.ExpandedMacroStorage$Companion.load(ExpandedMacroStorage.kt:216)
  at org.rust.lang.core.macros.MacroExpansionServiceImplInner$Companion.build(MacroExpansionManager.kt:248)
[...]

PR #4325, opened 2019-08-29
Remove VFS usage from `CargoMetadata`

vlad20012:
VFS should be accessed in the read action (the same issue as in #4319). In this case it's simpler to remove VFS usages

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
@@ -7,8 +7,8 @@ package org.rust.cargo.toolchain.impl
 
 import com.google.gson.Gson
 import com.google.gson.JsonObject
+import com.intellij.openapi.util.io.FileUtil
 import com.intellij.openapi.vfs.LocalFileSystem
-import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.util.PathUtil
 import org.rust.cargo.project.workspace.CargoWorkspace
 import org.rust.cargo.project.workspace.CargoWorkspace.Edition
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -256,32 +256,42 @@ object CargoMetadata {
...
-                name,
-                makeTargetKind(cleanKind, cleanCrateTypes),
-                edition.cleanEdition(),
-                doctest = doctest ?: true
-            )
+    private fun Target.clean(rootPath: String): CargoWorkspaceData.Target? {
+        val crateRootPath = if (FileUtil.isAbsolute(src_path)) {
+            src_path
+        } else {
+            rootPath + File.separator + src_path
```
Undin: `File.separator` differs on different OS. At the same time, VFS uses only Unix file separator.
So it seems that it breaks `org.rust.cargo.project.workspace.WorkspaceImpl#findTargetByCrateRoot` method

src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -256,32 +256,42 @@ object CargoMetadata {
...
+        }
+        val crateRootUrl = LocalFileSystem.PROTOCOL_PREFIX + toCanonicalPathSafe(crateRootPath)
```
Undin: Same about separators from `toCanonicalPathSafe`

vlad20012:
Looks like problems here are very unlikely, also [recent changes](https://github.com/JetBrains/intellij-community/commit/501c21024d431fb586c3393d36021207c2d25c3e) in the platform fixes them. It's anyway could be nice to remove VFS usage here, but I don't want to deal with Windows now...

Issue #4322, opened 2019-08-29
"Put fields on separate lines/one line" intentions for expressions


ortem:
It would be great to have such intentions to convert struct expressions (`RsStructLiteral`)

```Rust
MyStruct { x: 10, y: 20 };
```

into

```Rust
MyStruct {
    x: 10,
    y: 20
};
```

as vice versa.

For example, a similar feature for struct declarations #4282

PR #4332, opened 2019-08-30
INT: add "Put struct literal fields on separate lines/one line" intention

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/4322
I basically copy-pasted the tests from the struct fields and changed them to work with struct literal fields.

Note:
Test with a single line comment produces this formatting on a struct:
```rust
struct S {
    x: i32,
    // comment x
    y: i32,
    z: i32 // comment z
}
```
and this formatting with a literal:
```rust
struct S {  x: i32, y: i32, z: i32 }
fn foo() {
    S {
        x: 0, // comment x
        y: 0,
        z: 0 // comment z
    };
}
```
Not sure if that's important. Should I investigate?
Also, should I add the same thing for Enum literals?

### DIFF
--- a/src/test/kotlin/org/rust/ide/intentions/ChopLiteralFieldListIntentionTest.kt
+++ b/src/test/kotlin/org/rust/ide/intentions/ChopLiteralFieldListIntentionTest.kt
@@ -1,0 +1,185 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+class ChopLiteralFieldListIntentionTest : RsIntentionTestBase(ChopLiteralFieldListIntention()) {
+    fun `test one parameter`() = doUnavailableTest("""
+        struct S { x: i32 }
+        fn foo {
+            S { /*caret*/x: 0 };
+        }
+    """)
+
+    fun `test two parameters`() = doAvailableTest("""
+        struct S { x: i32, y: i32 }
+        fn foo() {
+            S { /*caret*/x: 0, y: 0 };
+        }
+    """, """
+        struct S { x: i32, y: i32 }
+        fn foo() {
+            S {
+                x: 0,
+                y: 0
+            };
+        }
+    """)
+
+    fun `test has all line breaks`() = doUnavailableTest("""
+        struct S { x: i32, y: i32, z: i32 }
+        fn foo() {
+            S {
+                /*caret*/x: 0,
+                y: 0,
+                z: 0
+            };
+        }
+    """)
+
+    fun `test has some line breaks`() = doAvailableTest("""
+        struct S { x: i32, y: i32, z: i32 }
+        fn foo() {
+            S { x: 0, /*caret*/y: 0,
+                z: 0
+            };
+        }
+    """, """
+        struct S { x: i32, y: i32, z: i32 }
+        fn foo() {
+            S {
+                x: 0,
+                y: 0,
+                z: 0
+            };
+        }
+    """)
+
+    fun `test has some line breaks 2`() = doAvailableTest("""
+        struct S { x: i32, y: i32, z: i32 }
+        fn foo() {
+            S {
+                x: 0, y: 0, z: 0/*caret*/
+            };
+        }
+    """, """
+        struct S { x: i32, y: i32, z: i32 }
+        fn foo() {
+            S {
+                x: 0,
+                y: 0,
+                z: 0
+            };
+        }
+    """)
+
+    fun `test has comment`() = doUnavailableTest("""
+        struct S {  x: i32, y: i32, z: i32 }
+        fn foo() {
+            S { 
+                /*caret*/x: 0, /* comment */ 
+                y: 0,
+                z: 0
+            };
+        }
+    """)
+
+    fun `test has comment 2`() = doAvailableTest("""
+        struct S {  x: i32, y: i32, z: i32 }
+        fn foo() {
+            S { 
+                /*caret*/x: 0, /*
+                    comment
+                */y: 0,
+                z: 0
+            };
+        }
+    """, """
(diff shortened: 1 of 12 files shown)


src/test/kotlin/org/rust/ide/intentions/ChopLiteralFieldListIntentionTest.kt
```diff
@@ -0,0 +1,130 @@
...
+    fun `test two parameter`() = doAvailableTest("""
+        struct S { x: i32, y: i32 }
+        fn foo() {
+            S { /*caret*/x: i32, y: i32 };
+        }
+    """, """
+        struct S { x: i32, y: i32 }
+        fn foo() {
+            S {
+                x: i32,
```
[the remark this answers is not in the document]
Kobzol: Uhh, this was a really bad case of copy-paste blindness :-) Thanks.

Kobzol:
@ortem The case with `S { x: 0, ..a }` is a bit complicated. Right now the chop/join intentions are implemented using an abstract class that assumes that the individual items can be described with two types (the container element and the child elements). However `..` is a separate attribute of `RsStructLiteralBody`, so it doesn't fit into the current implementation, and it also belongs to the `expr` of the `RsStructLiteralBody` element.

Another problem (which is not handled by the current solution) is what happens if the elements have a whitespace inside of them, i.e.:
```rust
S { /*caret*/x:

i32,
y:

i32 };
```
or
```rust
 S { x: 1, ..
  s };
```
in this case the current solution doesn't work.

So I'd suggest to:
1) Add explicit support for "paired" elements (`a: 5`, `..s`) and normalize them before running the intention, i.e. remove whitespace inside of them. Not sure if this can be done generally if there are comments?
2) Refactor the intentions so that they can work with `..s`. In general they need to a) treat `..s` as a single element, check linefeed only after `s` and check linefeed only before `..`.

Undin:
@ortem ping

ortem:
@Kobzol sorry for the late response,

IMO this PR should be only about literal fields support, but not about things related to formatting and whitespaces handling. If you have some ideas on "Put <something> on separate/one line" intentions improvement -- you're welcome to open another PR (or issue)!

With regard to enum literals, looks like you've already done because enum literals like `MyEnum::Variant { x: 1, y: 2 }` are actually based on `RsStructLiteralBody` and  `RsStructLiteralField` too.

So the main challenge now is the `..s` support, and it will be great if you try to find a way to implement it in this PR.

Kobzol:
@ortem I added basic support for `..` for the chop intention, but I had to make several modifications to `ListIntentionBase`. Let me know if this is OK or if I should find a different solution. 

Also right now `..` is only formatted correctly if it is the last item in the field literal. This is the only case when it compiles (it has to be last for it to compile), should I also make the formatting work if it's not the last element?

ortem:
@Kobzol 
>  I added basic support for .. for the chop intention, but I had to make several modifications to ListIntentionBase

I think it's generally OK, so let's adapt the join intention too!

> should I also make the formatting work if it's not the last element?

Invalid struct expressions like `S { x: 1, .. s, z: 2 }` lead to parsing error, so you shouldn't handle them.

Also, please don't forget to enable `JoinLiteralFieldListIntention` in `src/main/resources/META-INF/core.xml` and add some tests for enums.

Kobzol:
I added `JoinLiteralFieldListIntention` to `core.xml`, added Enum tests, added support for the join intention, rebased onto master and squashed the commits.

ortem:
@Kobzol Yes, you should add `description.html`, `before.rs.template` and `after.rs.template`.

Kobzol:
Thank you, I added the documentation.

src/test/kotlin/org/rust/ide/intentions/ChopLiteralFieldListIntentionTest.kt
```diff
@@ -0,0 +1,200 @@
...
+        struct S { x: i32, y: i32 }
+        fn foo() {
+            S {
+                x: 0,
+                y: 0
+            };
+        }
+    """)
+
+    fun `test two parameters`() = doAvailableTest("""
```
ortem: Looks like you forgot to delete this invalid test

src/test/kotlin/org/rust/ide/intentions/ChopLiteralFieldListIntentionTest.kt
```diff
@@ -0,0 +1,200 @@
...
+
+class ChopLiteralFieldListIntentionTest : RsIntentionTestBase(ChopLiteralFieldListIntention()) {
+    fun `test one parameter`() = doUnavailableTest("""
+        struct S { x: i32 }
+        fn foo {
+            S { /*caret*/x: i32 };
+        }
+    """)
+
+    fun `test two parameter`() = doAvailableTest("""
```
ortem: parameter**s**

src/test/kotlin/org/rust/ide/intentions/JoinLiteralFieldListIntentionTest.kt
```diff
@@ -0,0 +1,148 @@
...
+                /*caret*/x: 0
+            };
+        }
+    """, """
+        struct S { x: i32 }
+        fn foo() {
+            S { x: 0 };
+        }""")
+
+    fun `test two parameter`() = doAvailableTest("""
```
ortem: parameter**s**

ortem:
Also, it'd be better to squash commits into one (`INT: ` prefix would be great) because the first revision is a bit incorrect without templates from the second.

Kobzol:
Thanks for the review, I usually squash the commits right before the final merge, so that I don't have to do it after every change :) The tests failed, but it seems like it's something not relevant to my changes.

ortem:
@Kobzol OK, great work, thank you!

bors r+

Undin:
bors retry

PR #4346, opened 2019-09-02
Fix switch to AST when resolving nested macro with the new engine

vlad20012:
The bug was introduced in #3750 (not released yet)


vlad20012:
bors retry

PR #4352, opened 2019-09-04
PERF&GRAM: micro-optimize parsing

vlad20012:
A set of parser micro-optimizations. In total, they give about 10% of the gain.


src/main/grammars/RustParser.bnf
```diff
@@ -314,9 +314,10 @@ fake ValueParameter ::= OuterAttr* Pat? TypeReference? {
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-FnParameter ::= !(Variadic) OuterAttr* [ Pat ':' ] TypeReference
-private FnParameter_with_recover ::= FnParameter (',' | &')') {
-  pin = 1
+FnParameter ::= OuterAttr* [ Pat ':' ] (VariadicUpper | TypeReference)
```
vlad20012: Fixed double-parsing of `Pat` (a complex rule with a lot of `or`s). Upgrade to `Variadic` with `VariadicUpper`

src/main/grammars/RustParser.bnf
```diff
@@ -340,7 +341,7 @@ fake ValueParameterList ::= SelfParameter? ValueParameter* Variadic? {
 }
 
 FnParameters          ::= '(' !',' [ SelfParameter (',' | &')') ]
-                                   FnParameter_with_recover* [Variadic ','? ]
+                                   FnParameter_with_recover*
```
vlad20012: (Moved to `FnParameter`)

src/main/grammars/RustParser.bnf
```diff
@@ -800,31 +801,31 @@ private TypeAscription ::= ':' TypeReference
...
                                | RefLikeType
                                | FnPointerType
-                               | MacroType
                                // External rule to find the difference between `(T)` and `(T,)`
                                | <<tupleOrParenType TypeReferenceInner TupleType_upper>>
                                | ExplicitTraitType
                                | TrivialBaseType
                                // External rule to find the difference between `T` and `T+T`
                                | <<baseOrTraitType PathGenericArgsWithoutColons ImplicitTraitType TraitType_upper>>
+                               | MacroType
```
vlad20012: Re-ordered. `MacroType` and `BasePath` both start with a path, so this is a double-parsing, but `MacroType` is a very rare case. `baseOrTraitType` is updated to reject `MacroType`.

src/main/grammars/RustParser.bnf
```diff
@@ -800,31 +801,31 @@ private TypeAscription ::= ':' TypeReference
...
                                | ForInType // FIXME O(2^n) for types like `for<'a> A<for<'b> B<for<'c> C<...>>>`
 
 private TypeReferenceNoImplicitTraitTypeInner ::= ArrayType
                                | RefLikeType
                                | FnPointerType
-                               | MacroType
                                // External rule to find the difference between `(T)` and `(T,)`
                                | <<tupleOrParenType TypeReferenceInner TupleType_upper>>
                                | ExplicitTraitType
+                               | MacroType
```
vlad20012: The same, but can't move after `MacroType` b/c it will not be rejected.

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -244,8 +244,8 @@ object RustParserUtil : GeneratedParserUtilBase() {
         val bound = enter_section_(b)
         val traitRef = enter_section_(b)
 
-        if (!pathP.parse(b, level + 1)) {
-            // May be it is lifetime `'a` or `for<'a>`
+        if (!pathP.parse(b, level + 1) || nextTokenIs(b, EXCL)) {
```
vlad20012: Reject `!` after a path. Needed for changes in `TypeReferenceInner` rule

src/test/resources/org/rust/lang/core/parser/fixtures/complete/fn.rs
```diff
@@ -18,3 +18,5 @@ fn diverging() -> ! { panic("! is a type") }
 
 unsafe extern "C" fn ext_fn1(a: bool, ...) {}
 unsafe extern "C" fn ext_fn2(a: bool, args: ...) {}
+unsafe extern "C" fn ext_fn3(a: bool, ...,) {}
```
vlad20012: Just ensures that trailing comma is parsed after a variadic parameter.

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
Undin: Maybe it makes sense to swap these conditions because `nextTokenIs(b, EXCL)` should be much faster?

vlad20012: Order is important b/c `pathP.parse` is not pure (moves a lexer position to the end of a path, if true returned)

Undin:
bors retry

Issue #4355, opened 2019-09-05
Make function bodies lazy parseable & reparseable


vlad20012:
By this I want to achieve the following goals:
1. Speed up indexation. I measured that 40% of indexation is parsing. I also measured that parsing function bodies take more than 50% of parsing. So we can slightly speed up indexation. 
2. Do not re-parse the entire document when typing in a function body. Instead, re-parse the only function in which user is typing. I.e. make parsing incremental.

The plan:
- [x] Make macro def/call bodies lazy parseable/reparseable at first because it's relatively easy and shows the general approach. #4357 ~~(**UPD**: macro **call** bodies are only lazy parseable, but not reparseable for now due to problems with builtin (special case) macros)~~ #4877
- [x] Make function bodies lazy parseable & reparseable, BUT parse them during indexation. By this, we achieve the second goal - incremental parsing. At this point, we must parse function bodies during indexation because they can contain definitions that should be indexed (modules, functions, etc) or even global-visible definitions (`impl`s inside function bodies). #4862
- [x] Make a heuristic to detect function bodies that should be parsed during indexation. E.g. parse it if it contains item definition keywords (`fn`, `impl`, `mod`, etc). #4925

PR #4357, opened 2019-09-05
PERF&GRAM: make macros lazy parseable & reparseable

vlad20012:
PoC implementation of lazy parsing, Most likely this doesn't affect users anyhow.
Part of #4355


vlad20012:
Made macro call bodies non-reparseable because of problems with builtin macros

vlad20012:
bors retry

PR #4398, opened 2019-09-16
Macro type provider and extend selection

vlad20012:
Now `Extend Selection` and `Type Info` actions should work properly for code inside macro calls

Extend selection:
Peek 2019-09-30 18-03

Type provider:
Peek 2019-09-30 18-03_1

(note that this works only with the new macro expansion engine enabled)

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
@@ -6,12 +6,18 @@
 package org.rust.ide.hints
 
 import com.intellij.lang.ExpressionTypeProvider
+import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
+import com.intellij.psi.impl.FakePsiElement
+import org.rust.lang.core.macros.findExpansionElementOrSelf
+import org.rust.lang.core.macros.findMacroCallExpandedFromNonRecursive
+import org.rust.lang.core.macros.mapRangeFromExpansionToCallBodyStrict
 import org.rust.lang.core.psi.RsExpr
 import org.rust.lang.core.psi.RsPat
 import org.rust.lang.core.psi.RsPatField
 import org.rust.lang.core.psi.ext.RsItemElement
-import org.rust.lang.core.psi.ext.ancestors
+import org.rust.lang.core.psi.ext.contexts
+import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.type
 import org.rust.openapiext.escaped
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -233,6 +239,48 @@ private fun Int.fromBodyRelativeOffset(call: RsMacroCall): Int? {
...
+fun RsMacroCall.mapRangeFromExpansionToCallBodyStrict(range: TextRange): TextRange? {
+    return mapRangeFromExpansionToCallBody(range).singleOrNull()?.takeIf { it.length == range.length }
+}
+
+private fun RsMacroCall.mapRangeFromExpansionToCallBody(range: TextRange): List<TextRange> {
+    val expansion = expansion ?: return emptyList()
+    return mapRangeFromExpansionToCallBody(expansion, this, range)
+}
+
+private fun mapRangeFromExpansionToCallBody(
```
Undin: Do we really need all `mapRangeFromExpansionToCallBody*` functions?
It seems we need only `mapRangeFromExpansionToCallBodyStrict` because it public and `mapRangeFromExpansionToCallBody(org.rust.lang.core.macros.MacroExpansion, org.rust.lang.core.psi.RsMacroCall, org.rust.lang.core.macros.MappedTextRange)` because it's recursive.

Maybe it's better to inline other ones? They look quite small and add only additional overloads of `mapRangeFromExpansionToCallBody` that can be confusing. Also, such inlining allows us to rename `mapRangeFromExpansionToCallBodyStrict` to `mapRangeFromExpansionToCallBody` - looks like `strict` doesn't add additional info because we don't have public non strict version

src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
```diff
@@ -20,18 +26,43 @@ class RsExpressionTypeProvider : ExpressionTypeProvider<PsiElement>() {
     override fun getErrorHint(): String = "Select an expression!"
 
     override fun getExpressionsAt(pivot: PsiElement): List<PsiElement> =
-        pivot.ancestors
+        pivot.findExpansionElementOrSelf().contexts
```
Undin:
```suggestion
        pivot.findExpansionElementOrSelf()
            .contexts
```

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -233,6 +239,48 @@ private fun Int.fromBodyRelativeOffset(call: RsMacroCall): Int? {
     return elementOffset
 }
 
+private fun MappedTextRange.fromBodyRelativeRange(call: RsMacroCall): MappedTextRange? {
+    val newSrcOffset = srcOffset.fromBodyRelativeOffset(call) ?: return null
+    return MappedTextRange(newSrcOffset, dstOffset, length)
+}
+
+fun RsMacroCall.mapRangeFromExpansionToCallBodyStrict(range: TextRange): TextRange? {
```
Undin: As we discussed in person, let's convert it from extension function into a common one with `RsMacroCall` as the first argument. Looks like it has more sense
To avoid another one weird function in public scope, let's create util class and move all functions related to expansion and range mapping into it
Don't insist to do it in this PR

src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
```diff
@@ -20,18 +26,43 @@ class RsExpressionTypeProvider : ExpressionTypeProvider<PsiElement>() {
...
+    private fun getType(element: PsiElement): Ty = when (element) {
+        is MyFakePsiElement -> getType(element.elementInMacroExpansion)
+        is RsExpr -> element.type
+        is RsPat -> element.type
+        is RsPatField -> element.type
+        else -> error("Unexpected element type: $element")
+    }
+}
+
+private fun PsiElement.wrapIfInMacroExpansion(): PsiElement? {
```
Undin: Could we come up with better name?
For example, `wrapExpandedElements`?
Also, I'd prefer to add a short comment why we need it at all

src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
vlad20012: Non-strict version is needed for the (ongoing) highlighting. Another one overload replaces `MappedTextRange` to `TextRange`.

PR #4411, opened 2019-09-20
Fix incompatibilities with latest 2019.3 EAPs

Undin:
Fixes #4406
Fixes #4413

PR #4414, opened 2019-09-20
Fix incompatible return type

ice1000:
Something goes wrong in PR #4411, I'm fixing it


Undin:
Already fixed in #4411

PR #4416, opened 2019-09-21
Make "missing toolchain" notification dumb-aware

vlad20012:
If it is not dumb-aware, the notification persists until the end
of indexation. This is a source of confusion

bors r+


vlad20012:
bors retry

Issue #3508, opened 2019-03-09
Plugin throws exception and CLion core is blamed


ice1000:
## Environment

* **Intellij-Rust plugin version:** ![image](https://user-images.githubusercontent.com/16398479/54075470-aba38680-426d-11e9-9dd3-46a6ea0e535c.png)
* **IDE name and version:** ![image](https://user-images.githubusercontent.com/16398479/54075468-a0e8f180-426d-11e9-9ef6-bbe24493b423.png)

* **Operating system:** Windows 10

## Problem description

```
com.intellij.util.IncorrectOperationException: Rebind cannot be performed for class org.rust.lang.core.resolve.ref.RsModReferenceImpl
	at com.intellij.psi.PsiReferenceBase.bindToElement(PsiReferenceBase.java:124)
	at com.intellij.refactoring.move.moveFilesOrDirectories.MoveFilesOrDirectoriesProcessor.retargetUsages(MoveFilesOrDirectoriesProcessor.java:287)
	at com.intellij.refactoring.move.moveFilesOrDirectories.MoveFilesOrDirectoriesProcessor.performRefactoring(MoveFilesOrDirectoriesProcessor.java:195)
	at com.intellij.refactoring.BaseRefactoringProcessor.lambda$doRefactoring$9(BaseRefactoringProcessor.java:481)
	at com.intellij.refactoring.BaseRefactoringProcessor.lambda$doRefactoring$10(BaseRefactoringProcessor.java:492)
	at com.intellij.openapi.application.impl.ApplicationImpl.lambda$null$15(ApplicationImpl.java:1012)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:164)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:586)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:532)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:86)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:151)
	at com.intellij.openapi.progress.util.PotemkinProgress.runInSwingThread(PotemkinProgress.java:184)
[...]

PR #4423, opened 2019-09-22
REF: Fix mod declaration rebinding

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/3508.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsModReferenceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsModReferenceImpl.kt
@@ -24,4 +24,6 @@ class RsModReferenceImpl(
 
     override fun isReferenceTo(element: PsiElement): Boolean =
         element is RsFile && super.isReferenceTo(element)
+
+    override fun bindToElement(element: PsiElement): PsiElement? = null
 }


ice1000:
Nice!

ice1000:
bors try

ice1000:
(please ignore the above comment -- it's made by a mistake)

src/main/kotlin/org/rust/lang/core/resolve/ref/RsModReferenceImpl.kt
```diff
@@ -24,4 +24,6 @@ class RsModReferenceImpl(
 
     override fun isReferenceTo(element: PsiElement): Boolean =
         element is RsFile && super.isReferenceTo(element)
+
+    override fun bindToElement(element: PsiElement): PsiElement? = null
```
Undin: Why does this method return null?
I can imagine a situation when it can be rebound:
```rust
//- main.rs
mod foo;
mod bar;

//- foo/mod.rs
mod xxx;

//- foo/xxx.rs
// some code

//- bar/mod.rs
mod xxx;
```

And we move `foo/xxx.rs` to `bar` directory

Undin: Also, there is a magic interface `BindablePsiReference` that we probably should implement.
It's used somewhere in rename refactoring

PR #4427, opened 2019-09-23
Fix `Hash` trait autoimport on Rust nightly

vlad20012:
bors r+


src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -307,7 +308,7 @@ class RsPsiFactory(
                 blockFields != null -> " { .. }"
                 else -> ""
             }
-            val prefix = if (context.findInScope(variantName) != variant) "$enumName::" else ""
+            val prefix = if (context.findInScope(variantName, VALUES) != variant) "$enumName::" else ""
```
vlad20012: Btw I made namespaces explicit and now this looks like a bug :thinking:

PR #4434, opened 2019-09-25
CI: init github actions

Undin:
The main goal is to try github actions with our project.

For now, it launches only on linux and the corresponding checks are not included to bors mandatory checklist


Undin:
bors try

ice1000:
Can we have artifacts access from CI?

ice1000:
Waiting for a nightly build is kinda patience consuming.

PR #4464, opened 2019-09-27
PERF: fix AST switch in the hygiene

vlad20012:


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -1378,8 +1378,8 @@ private fun makeHygieneFilter(anchor: PsiElement): (RsPatBinding) -> Boolean {
         // This branch is needed to prevent AST access
         anchor.containingFile
     } else {
-        val nameIdentifier = if (anchor is RsReferenceElement) anchor.referenceNameElement else anchor
-        (nameIdentifier.findElementExpandedFrom(strict = false) ?: nameIdentifier).containingFile
+        val startOffset = (anchor as? RsPath)?.greenStub?.startOffset ?: anchor.startOffset
+        (findMacroCallAndOffsetExpandedFromUnchecked(anchor, startOffset)?.call ?: anchor).containingFile
```
Undin: I don't like `findMacroCallAndOffsetExpandedFromUnchecked ` function is public because it's rather hard to understand what it does.
So I'd prefer not to make it a part of our public API.
As I can see, it is only one call outside of `RsExpandedElement.kt` file.
And looks like we use it there only to pass stubbed offset for `RsPath`.
Maybe we can move `val startOffset = (anchor as? RsPath)?.greenStub?.startOffset ?: anchor.startOffset` into `org.rust.lang.core.macros.RsExpandedElementKt#findElementExpandedFrom` and call `findElementExpandedFrom` here as we did before these changes?

Undin: If it's possible than `MacroCallAndOffset` can be private as well

vlad20012: Not really. We can move hacks around RsPath into `findElementExpandedFrom`, but `findElementExpandedFrom` retrieves element inside a macro expansion by an offset, which requires AST access

Undin: As I understand current master implementation calls
```kotlin
    return anchor.containingFile.findElementAt(offset)
        ?.takeIf { it.startOffset == offset }
```
here because it invokes `org.rust.lang.core.macros.RsExpandedElementKt#findElementExpandedFromUnchecked` eventually but this one doesn't do it. Is it intentional change?
If so maybe we can add some parameter to `findElementExpandedFrom` not to invoke this piece of code?

Undin: Discussed in person that we can make a new public method instead of `findMacroCallAndOffsetExpandedFromUnchecked` to encapsulate all offset-related things in single place (`RsExpandedElement`) and do not touch them in `NameResolution`

PR #4471, opened 2019-09-29
Experimental features

Undin:
The main goal is to add an easy way to enable some features in nightly builds by default and disable them in stable ones because sometimes a feature is not production-ready but we want to get some feedback from real users.

Also, I've converted existing registry keys to experimental features because experimental features dialog has about 10 items, registry dialog has much more items. At the same time, registry knows about experimental features and adds all experimental features in registry dialog.

image
image


vlad20012:
Let's not enable `cfg` attributes for now because the implementation has major issues

Issue #4442, opened 2019-09-25
Add `Analyze backtrace` tool window


SomeoneToIgnore:
## Environment

* **IntelliJ Rust plugin version:** 0.2.106.2134-193
* **Rust toolchain version:** 1.37.0 (eae3437df 2019-08-13) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.3519.24)
* **Operating system:** macOS 10.14.6

## Problem description

Intellij Idea has an instrument to analyze the stacktraces: https://blog.jetbrains.com/idea/2006/08/analyzing-external-stack-traces/

When launched, it inserts the formatted clipboard contents into the tool window and, when applied, allows to navigate to the places in the source code for each frame.

Would be great to have similar functionality for Rust.

SomeoneToIgnore:
How can I find and run this action?

Searching everywhere does not help:
<img width="1126" alt="image" src="https://user-images.githubusercontent.com/2690773/68425757-65614700-01af-11ea-9ffd-1f0304148514.png">

mchernyavsky:
@SomeoneToIgnore
![image](https://user-images.githubusercontent.com/6079006/68431496-6ef4aa00-01c3-11ea-9b09-36cfc401ec11.png)

SomeoneToIgnore:
Ok, looks like if I enter `analyze stac` it appears, but not a letter before, that's odd.

What's also strange is that it's disabled (notice how gray it is):
<img width="663" alt="image" src="https://user-images.githubusercontent.com/2690773/68433553-7914a980-01bf-11ea-950a-e07366b53c94.png">

If I select it, nothing happens.

Any idea what should I do to make it work?

ortem:
@SomeoneToIgnore Unfortunately, this action is currently disabled in CLion (even with Rust plugin). We'll try to fix this problem on the CLion side and let you know when it is ready. However, you can use `Analyze Stack Trace or Thread Dump...` action in IDEA.

ortem:
Fixed in CLion 2020.1 EAP (coming soon).

PR #4479, opened 2019-10-01
ACT: Support `Analyze Stack Trace` action

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/4442.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/filters/RsBacktraceFilter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/filters/RsBacktraceFilter.kt
@@ -12,6 +12,7 @@ import com.intellij.openapi.editor.colors.TextAttributesKey
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.psi.PsiDocumentManager
+import org.rust.cargo.project.model.cargoProjects
 import org.rust.cargo.project.workspace.CargoWorkspace
 import org.rust.cargo.project.workspace.PackageOrigin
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/cargo/runconfig/filters/RsBacktraceFilter.kt
```diff
@@ -27,16 +28,42 @@ import java.util.regex.Pattern
  */
 class RsBacktraceFilter(
     project: Project,
-    cargoProjectDir: VirtualFile,
+    cargoProjectDir: VirtualFile?,
     workspace: CargoWorkspace?
 ) : Filter {
+    private val backtraceItemFilters: List<RsBacktraceItemFilter> = run {
```
Undin: Unfortunately, we cannot do such things because IDE store object instantiated for exception point so this filter always "remembers" state that project had at first `Analyze Stack Trace` action invocation.
Also, it leads to memory leak because we store link to `Workspace` and `VirtualFile` objects

mchernyavsky: Fixed.

mchernyavsky:
Now I use the new constructor in `RsBacktraceFilterTest`.

Undin:
> Now I use the new constructor in RsBacktraceFilterTest.

@mchernyavsky I meant that we need **new** test(s) to check new `RsBacktraceFilter` usage instead of changing old tests

mchernyavsky:
Fixed.

Issue #4438, opened 2019-09-25
Add `#![feature(zzz)]` attribute automatically when unstable feature is used


SomeoneToIgnore:
## Environment

* **IntelliJ Rust plugin version:** 0.2.106.2134-193
* **Rust toolchain version:** 1.39.0-nightly (6ef275e6c 2019-09-24) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.3519.24)
* **Operating system:** macOS 10.14.6

## Problem description

When I use the nightly toolchain and complete the unstable feature, I have to manually enable it, if haven't done before.

It would be great to add the necessary `#![feature(` attribute in this case.
Example:

```rust
fn main() {
    let test = Some(Some(22));
    test.fla/*caret*/
}
```

Expected:

```rust
#![feature(option_flattening)]
fn main() {
    let test = Some(Some(22));
    test.flatten()
}
```

Actual:

```rust
fn main() {
    let test = Some(Some(22));
    test.flatten()
}
```

PR #4482, opened 2019-10-02
ANN: Support `unstable` attribute for items

mchernyavsky:
Relates to https://github.com/intellij-rust/intellij-rust/issues/4438.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -24,6 +24,7 @@ import org.rust.lang.core.*
 import org.rust.lang.core.FeatureAvailability.CAN_BE_ADDED
 import org.rust.lang.core.FeatureAvailability.NOT_AVAILABLE
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsElementTypes.IDENTIFIER
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.Namespace
 import org.rust.lang.core.resolve.knownItems
@@ -236,6 +237,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
     private fun checkDotExpr(holder: RsAnnotationHolder, o: RsDotExpr) {
         val field = o.fieldLookup ?: o.methodCall ?: return
         checkReferenceIsPublic(field, o, holder)
+        checkUnstableAttribute(field, holder)
         if (field is RsMethodCall) {
             checkNotCallingDrop(field, holder)
         }
@@ -302,6 +304,21 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         error.addToHolder(holder)
     }
 
+    private fun checkUnstableAttribute(ref: RsReferenceElement, holder: RsAnnotationHolder) {
+        val startElement = ref.referenceNameElement.takeIf { it.elementType == IDENTIFIER } ?: return
+        if (ref.containingCargoPackage?.origin == PackageOrigin.STDLIB) return
+        val element = ref.reference.resolve() as? RsOuterAttributeOwner ?: return
+        for (attr in element.queryAttributes.unstableAttributes) {
+            val metaItems = attr.metaItemArgs?.metaItemList ?: continue
+            val featureName = metaItems.singleOrNull { it.name == "feature" }?.value ?: continue
+            val reason = metaItems.singleOrNull { it.name == "reason" }?.value
+            val reasonSuffix = if (reason != null) ": $reason" else ""
+            val feature = CompilerFeature.find(featureName)
+                ?: CompilerFeature(featureName, FeatureState.ACTIVE, null, cache = false)
+            feature.check(holder, startElement, null, "`$featureName` is unstable$reasonSuffix")
+        }
+    }
+
     private fun checkBaseType(holder: RsAnnotationHolder, type: RsBaseType) {
         if (type.underscore == null) return
         val owner = type.owner.parent
@@ -358,6 +375,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         }
 
         checkReferenceIsPublic(path, path, holder)
+        checkUnstableAttribute(path, holder)
     }
 
     private fun checkConstParameter(holder: RsAnnotationHolder, constParameter: RsConstParameter) {
--- a/src/main/kotlin/org/rust/lang/core/CompilerFeature.kt
+++ b/src/main/kotlin/org/rust/lang/core/CompilerFeature.kt
@@ -21,14 +21,31 @@ import org.rust.lang.core.stubs.index.RsFeatureIndex
 import org.rust.lang.utils.RsDiagnostic
 import org.rust.lang.utils.addToHolder
 
-data class CompilerFeature(val name: String, val state: FeatureState, val since: SemVer) {
-    constructor(name: String, state: FeatureState, since: String) : this(name, state, SemVer.parseFromText(since)!!)
+class CompilerFeature(
+    val name: String,
+    val state: FeatureState,
+    val since: SemVer?,
+    cache: Boolean = true
+) {
+
+    constructor(
+        name: String,
+        state: FeatureState,
+        since: String,
+        cache: Boolean = true
+    ) : this(name, state, SemVer.parseFromText(since)!!, cache)
+
+    init {
+        if (cache) {
+            knownFeatures[name] = this
+        }
+    }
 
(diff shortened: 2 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -299,6 +300,21 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
...
+        val element = ref.reference.resolve() as? RsOuterAttributeOwner ?: return
+        for (outerAttr in element.outerAttrList) {
+            val metaItem = outerAttr.metaItem
+            if (metaItem.name != "unstable") continue
+            val args = metaItem.metaItemArgs ?: continue
+            for (innerMetaItem in args.metaItemList) {
+                if (innerMetaItem.name != "feature") continue
+                val value = innerMetaItem.litExpr?.stringValue ?: continue
+                val feature = CompilerFeature(value, FeatureState.ACTIVE)
+                feature.check(holder, o, "`${feature.name}`")
```
Undin: `unstable` attribute frequently has `reason` value. Let's extract it to (don't forget that it's Rust string literal that can have escapes symbols) and use it as message

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -299,6 +300,21 @@ class RsErrorAnnotator : RsAnnotatorBase(), HighlightRangeExtension {
         error.addToHolder(holder)
     }
 
+    private fun checkUnstableAttribute(ref: RsReferenceElement, o: RsElement, holder: RsAnnotationHolder) {
+        val element = ref.reference.resolve() as? RsOuterAttributeOwner ?: return
+        for (outerAttr in element.outerAttrList) {
```
Undin: Maybe use `org.rust.lang.core.psi.ext.RsDocAndAttributeOwnerKt#getQueryAttributes` instead?

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -302,6 +304,17 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
     }
 
+    private fun checkUnstableAttribute(ref: RsReferenceElement, o: RsElement, holder: RsAnnotationHolder) {
+        val element = ref.reference.resolve() as? RsOuterAttributeOwner ?: return
+        for (attr in element.queryAttributes.unstableAttributes) {
+            val metaItems = attr.metaItemArgs?.metaItemList ?: continue
+            val featureName = metaItems.singleOrNull { it.name == "feature" }?.value ?: continue
+            val reason = metaItems.singleOrNull { it.name == "reason" }?.value?.let(::unescapeStringCharacters)
+            val feature = CompilerFeature.find(featureName) ?: CompilerFeature(featureName, FeatureState.ACTIVE)
+            feature.check(holder, o, null, "`$featureName` is unstable${if (reason != null) ": $reason" else ""}")
```
Undin: nit: I'd would extract `${if (reason != null) ": $reason" else ""}` to `reasonSuffix` variable. Looks like it's a bit more readable

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/lang/core/CompilerFeature.kt
```diff
@@ -21,14 +21,31 @@ import org.rust.lang.core.stubs.index.RsFeatureIndex
 import org.rust.lang.utils.RsDiagnostic
 import org.rust.lang.utils.addToHolder
 
-data class CompilerFeature(val name: String, val state: FeatureState, val since: SemVer) {
-    constructor(name: String, state: FeatureState, since: String) : this(name, state, SemVer.parseFromText(since)!!)
+class CompilerFeature(
+    val name: String,
+    val state: FeatureState,
+    val since: SemVer?,
+    cache: Boolean = false
```
Undin: Why default value is `false`?
You have more than 400 usages with non default value and only one with default one.

mchernyavsky: @Undin Fixed.

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -302,6 +303,20 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         error.addToHolder(holder)
     }
 
+    private fun checkUnstableAttribute(ref: RsReferenceElement, o: RsElement, holder: RsAnnotationHolder) {
+        val element = ref.reference.resolve() as? RsOuterAttributeOwner ?: return
+        if (element.containingCargoPackage?.origin == PackageOrigin.STDLIB) return
```
Undin: This line makes whole inspection useless because `unstable` attribute can be used only in stdlib. So `element.containingCargoPackage?.origin == PackageOrigin.STDLIB` should be always `true `

Undin:
@mchernyavsky please, check your changes yourself before push

mchernyavsky:
@Undin Fixed and tested.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -302,6 +303,20 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+        if (o.containingCargoPackage?.origin == PackageOrigin.STDLIB) return
+        val element = ref.reference.resolve() as? RsOuterAttributeOwner ?: return
+        for (attr in element.queryAttributes.unstableAttributes) {
+            val metaItems = attr.metaItemArgs?.metaItemList ?: continue
+            val featureName = metaItems.singleOrNull { it.name == "feature" }?.value ?: continue
+            val reason = metaItems.singleOrNull { it.name == "reason" }?.value
+            val reasonSuffix = if (reason != null) ": $reason" else ""
+            val feature = CompilerFeature.find(featureName)
+                ?: CompilerFeature(featureName, FeatureState.ACTIVE, null, cache = false)
+            feature.check(holder, o, null, "`$featureName` is unstable$reasonSuffix")
```
Undin: Let's highlight item more precisely - only the corresponding identifier instead of the whole path/dot expression.

| Actual | Expected |
|-|-|
| Screen Shot 2019-12-09 at 16 32 48 | Screen Shot 2019-12-09 at 16 29 58 |

Don't forget to add tests to check compound paths

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -302,6 +303,20 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         error.addToHolder(holder)
     }
 
+    private fun checkUnstableAttribute(ref: RsReferenceElement, o: RsElement, holder: RsAnnotationHolder) {
```
Undin: Hm, why do you need `ref` and `o` here?
Looks like `ref` is always enough

mchernyavsky:
@Undin Fixed.

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -2910,4 +2910,59 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
...
+        impl <error descr="`aaa` is unstable [E0658]">S</error> {
+            #[unstable(feature = "ccc", reason = "bar \
+                baz")]
+            fn foo(self) -> Self {}
+        }
+        
+        fn main() {
+            let x = <error descr="`aaa` is unstable [E0658]">S</error> { field: 0 };
+            x.<error descr="`bbb` is unstable: foo [E0658]">field</error>;
+            x.<error descr="`ccc` is unstable: bar baz [E0658]">foo()</error>;
```
Undin: You still highlight too wide range because it contains all method params
Should be 
```
x.<error descr="`ccc` is unstable: bar baz [E0658]">foo</error>();
```

`org.rust.lang.core.psi.ext.RsReferenceElement#getReferenceNameElement` may be helpful here but looks like we don't want to highlight `Self` paths so probably, you need to check element type of `getReferenceNameElement` is identifier

mchernyavsky: Fixed.

PR #4483, opened 2019-10-02
RES: Use three-valued logic in cfg evaluation

ortem:
Related to #1191. Some cfg options cannot be evaluated to `True` or `False` in IDE (e.g. `test` option), so now they are evaluated to `Unknown` according to [three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic).

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
+++ b/src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
@@ -9,29 +9,105 @@ import org.rust.cargo.CfgOptions
 import org.rust.lang.core.psi.RsMetaItem
 import org.rust.lang.core.psi.ext.name
 import org.rust.lang.core.psi.ext.value
+import org.rust.lang.utils.evaluation.ThreeValuedLogic.*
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
```diff
@@ -9,29 +9,103 @@ import org.rust.cargo.CfgOptions
...
         }
+
         return result
     }
 
-    private fun evaluatePredicate(predicate: CfgPredicate): Boolean = when (predicate) {
-        is CfgPredicate.All -> predicate.list.all(::evaluatePredicate)
-        is CfgPredicate.Any -> predicate.list.any(::evaluatePredicate)
+    private fun evaluatePredicate(predicate: CfgPredicate): ThreeValuedLogic = when (predicate) {
+        is CfgPredicate.All -> predicate.list.fold(Unknown) { acc, pred -> acc and evaluatePredicate(pred) }
```
vlad20012:
```suggestion
        is CfgPredicate.All -> predicate.list.fold(True) { acc, pred -> acc and evaluatePredicate(pred) }
```

PR #4485, opened 2019-10-02
TYPE & STUB: Render RsValueParameter without mutability

### LINKED ISSUE #4393: Confusing type info with mut
## Environment

* **IntelliJ Rust plugin version:** `0.2.105.2133-192`
* **Rust toolchain version:** `rustc 1.39.0-nightly (ca3766e2e 2019-09-14)`
* **IDE name and version:** CLion 2019.2.2

## Problem description

If the type signature of a function's parameter has `mut`, then intellij-rust will show that as a part of the type info. However this information is useless, if not confusing, from the point of view of  a caller.

## Steps to reproduce

Given the following snippet,

```rust
use std::io::{Read, Write};

struct Foo {
    f: std::fs::File,
}

impl Foo {
    fn write(&mut self, buf: &[u8; 4096]) -> std::io::Result<()> {
        self.f.write_all(buf)
    }
}
```

`intellij-rust` will show the following type info:

intellij-rust-screenshot

Note the type info is shown as `mut buf`, but IMHO the `mut` part should be hidden to the caller.

ortem:
Fixes #4393

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsValueParameter.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsValueParameter.kt
@@ -5,10 +5,29 @@
 
 package org.rust.lang.core.psi.ext
 
-import org.rust.lang.core.psi.RsValueParameter
+import org.rust.lang.core.psi.*
 
 
-val RsValueParameter.patText: String? get() {
-    val stub = greenStub
-    return if (stub != null) stub.patText else pat?.text
-}
+val RsValueParameter.patText: String?
+    get() {
+        val stub = greenStub
+        return if (stub != null) stub.patText else pat?.presentableText
+    }
+
+private val RsPat.presentableText: String
+    get() = when (this) {
+        is RsPatTup -> patList.joinToString(separator = ", ", prefix = "(", postfix = ")") { it.presentableText }
+
+        is RsPatSlice -> patList.joinToString(separator = ", ", prefix = "[", postfix = "]") { it.presentableText }
+
+        is RsPatStruct -> patFieldList.joinToString(separator = ", ", prefix = "${path.text} {", postfix = "}") { it.presentableText }
+
+        is RsPatTupleStruct -> patList.joinToString(separator = ", ", prefix = "${path.text}(", postfix = ")") { it.presentableText }
+
+        is RsPatIdent -> patBinding.identifier.text
+
+        else -> text
+    }
+
+private val RsPatField.presentableText: String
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsValueParameter.kt
```diff
@@ -5,10 +5,29 @@
...
-    val stub = greenStub
-    return if (stub != null) stub.patText else pat?.text
-}
+val RsValueParameter.patText: String?
+    get() {
+        val stub = greenStub
+        return if (stub != null) stub.patText else pat?.textWithoutMutability
+    }
+
+private val RsPat.textWithoutMutability: String
```
Undin: Maybe `presentableText` is better?
The current name is too specific

src/test/kotlin/org/rust/ide/hints/RsInlayParameterHintsProviderTest.kt
```diff
@@ -118,6 +118,16 @@ class RsInlayParameterHintsProviderTest : RsPlainInlayHintsProviderTestBase() {
...
+    fun `test fn arg with mut ident`() = checkByText("""
+        fn foo(mut arg: u32) {}
+        fn main() { foo(/*hint text="arg:"*/0); }
+    """, enabledHints = RsPlainParameterHint.PARAMETER_HINT)
+
+    fun `test fn arg with mut array`() = checkByText("""
+        fn foo([mut x, y]: [i32; 2]) {}
+        fn main() { foo(/*hint text="[x, y]:"*/0); }
+    """, enabledHints = RsPlainParameterHint.PARAMETER_HINT)
+
```
Undin: Let's add tests for all cases (`RsPatStruct` and `RsPatTupleStruct` cases)

src/main/kotlin/org/rust/lang/core/psi/ext/RsValueParameter.kt
```diff
@@ -5,10 +5,29 @@
...
+        is RsPatStruct -> patFieldList.joinToString(separator = ", ", prefix = "${path.text} {", postfix = "}") { it.textWithoutMutability }
+
+        is RsPatTupleStruct -> patList.joinToString(separator = ", ", prefix = "${path.text}(", postfix = ")") { it.presentableText }
+
+        is RsPatIdent -> patBinding.identifier.text
+
+        else -> text
+    }
+
+private val RsPatField.textWithoutMutability: String
```
Undin:
```suggestion
private val RsPatField. presentableText: String
```

ortem:
bors r=Undin

Undin:
bors retry

PR #4487, opened 2019-10-03
PERF: fix index access during PSI event processing

vlad20012:
... that forces to rebuild stubs synchronously (in the EDT!) and affects editor responsiveness

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
@@ -24,11 +24,11 @@ interface RsExpandedElement : RsElement {
     override fun getContext(): PsiElement?
 
     companion object {
-        fun getContextImpl(psi: RsExpandedElement): PsiElement? {
+        fun getContextImpl(psi: RsExpandedElement, isIndexAccessForbidden: Boolean = false): PsiElement? {
             psi.expandedFrom?.let { return it.context }
             psi.getUserData(RS_EXPANSION_CONTEXT)?.let { return it }
             val parent = psi.stubParent
-            if (parent is RsFile) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/macros/RsExpandedElement.kt
```diff
@@ -33,6 +33,12 @@ interface RsExpandedElement : RsElement {
             }
             return parent
         }
+
+        fun getContextImplWithoutIndexAccess(psi: RsExpandedElement): PsiElement? {
```
Undin: Maybe it's better to add `isIndexAccessForbidden: Boolean = false` to `getContextImpl` and avoid code duplication?

src/main/kotlin/org/rust/lang/core/psi/ext/RsModificationTrackerOwner.kt
```diff
@@ -33,11 +34,26 @@ interface RsModificationTrackerOwner : RsElement {
...
     ) as? RsModificationTrackerOwner
 }
+
+// We have to process contexts without index access because accessing indices during PSI event processing is slow.
+private val PsiElement.contextWithoutIndexAccess: PsiElement?
+    get() = if (this is RsExpandedElement) RsExpandedElement.getContextImplWithoutIndexAccess(this) else stubParent
+
+@Suppress("UNCHECKED_CAST")
+private fun <T : PsiElement> PsiElement.findContextOfTypeWithoutIndexAccess(strict: Boolean, vararg classes: KClass<out T>): T? {
+    var element: PsiElement? = if (strict) this.contextWithoutIndexAccess else this
```
Undin:
```suggestion
    var element = if (strict) contextWithoutIndexAccess else this
```

vlad20012:
Fixed

PR #4489, opened 2019-10-03
Assert we do not access indices during document committing

vlad20012:
Accessing indices during document committing (PSI event processing) forces to rebuild stubs synchronously (in the EDT!) and affects editor responsiveness, so let's assert that nobody does that.

Previous index access was fixed in #4487

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/utils.kt
+++ b/src/main/kotlin/org/rust/openapiext/utils.kt
@@ -39,6 +39,7 @@ import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapi.vfs.VirtualFileWithId
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.*
+import com.intellij.psi.impl.PsiDocumentManagerBase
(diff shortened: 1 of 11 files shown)


vlad20012:
Hmm, why this isn't merged yet

bors r+

src/main/kotlin/org/rust/openapiext/utils.kt
```diff
@@ -101,6 +102,15 @@ fun checkIsSmartMode(project: Project) {
     if (DumbService.getInstance(project).isDumb) throw IndexNotReadyException.create()
 }
 
+fun checkCommitIsNotInProgress(project: Project) {
+    val app = ApplicationManager.getApplication()
+    if ((app.isUnitTestMode || app.isInternal) && app.isDispatchThread) {
+        if ((PsiDocumentManager.getInstance(project) as PsiDocumentManagerBase).isCommitInProgress) {
+            error("Accessing indices during PSI event processing can lead to typing performance issues")
```
Undin, line 109: Probably, `com.intellij.openapi.diagnostic.Logger#error` is better because it throws an error in tests and always write in log even in non-internal mode

PR #4491, opened 2019-10-03
GRAM&REF&PERF: rework macro calls parsing

vlad20012:
- refactor special macros parsing (now it looks better and works faster)
- recover on missed semicolon at the and of a macro (`foo!();`)


src/main/grammars/RustParser.bnf
```diff
@@ -1319,41 +1320,39 @@ Macro2 ::= AttrsAndVis MACRO_KW identifier ( Macro2FunctionLikeBody | Macro2Matc
...
-"assert" | "debug_assert" | "assert_eq" | "assert_ne" | "debug_assert_eq" | "debug_assert_ne" |
-"trace" | "log" | "warn" | "debug" | "error" | "info" | "include_str" | "include_bytes" | "concat"
+BlockElementMacroCall ::= <<parseMacroCall 'MC_BLOCK'>> {
+  elementType = MacroCall
+  hooks = [ leftBinder = "ADJACENT_LINE_COMMENTS" ]
+  name = "macro call"
+}
 
-private MacroHead ::= AttrsAndVis &(PathWithoutTypes '!') (SpecialMacro | CommonMacro)
+ExprLikeMacroCall ::= <<parseMacroCall 'MC_EXPR'>> {
```
Undin: Hm, as we discussed in person, we need `MC` prefix only to avoid duplicate names in the generated parser.
I personally don't like this prefix because it makes me think about what it means. Of course, it means `macro call` but this knowledge is quite unusable because when I'm reading this rule I'm already in macro call context.
Maybe it's better just to replace it with 
```suggestion
ExprLikeMacroCall ::= <<parseMacroCall 'MacroCallParsingMode.EXPR'>> {
```
and drop prefix at all?

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -36,6 +37,11 @@ object RustParserUtil : GeneratedParserUtilBase() {
         NO_TYPES
     }
     enum class BinaryMode { ON, OFF }
+    enum class MacroCallParsingMode(val semicolon: Boolean, val pin: Boolean, val forbidExprSpecialMacros: Boolean) {
```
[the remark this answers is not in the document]
vlad20012: But it's about any _expr_ special macro rules...

src/main/grammars/RustParser.bnf
```diff
@@ -1319,41 +1320,39 @@ Macro2 ::= AttrsAndVis MACRO_KW identifier ( Macro2FunctionLikeBody | Macro2Matc
 private Macro2FunctionLikeBody ::= '(' MacroPatternContents ')' '{' MacroExpansionContents '}'
 private Macro2MatchLikeBody ::= '{' (MacroCase ','?)* '}'
 
-MacroCall ::= MacroHead <<macroSemicolon>> {
+fake MacroCall ::= AttrsAndVis PathWithoutTypes '!' identifier? (
+    MacroArgument | ExprMacroArgument | FormatMacroArgument | AssertMacroArgument |
+      VecMacroArgument | LogMacroArgument | IncludeMacroArgument | ConcatMacroArgument
```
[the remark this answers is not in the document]
vlad20012: I don't sure how to extract it because it is a  `fake` rule

src/main/grammars/RustParser.bnf
```diff
@@ -1319,41 +1319,42 @@ Macro2 ::= AttrsAndVis MACRO_KW identifier ( Macro2FunctionLikeBody | Macro2Matc
...
-  hooks = [ leftBinder = "ADJACENT_LINE_COMMENTS" ]
   mixin = "org.rust.lang.core.psi.ext.RsMacroCallImplMixin"
   stubClass = "org.rust.lang.core.stubs.RsMacroCallStub"
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-MacroCallNoSemicolons ::= MacroHead {
-  pin = 2
+// Needs semicolon `foo();`
+ItemLikeMacroCall ::= <<parseMacroCall 'MC_ITEM'>> {
```
Undin:
```suggestion
ItemLikeMacroCall ::= <<parseMacroCall ''MacroCallParsingMode.ITEM''>> {
```

src/main/grammars/RustParser.bnf
```diff
@@ -1319,41 +1319,42 @@ Macro2 ::= AttrsAndVis MACRO_KW identifier ( Macro2FunctionLikeBody | Macro2Matc
...
-  pin = 2
-  hooks = [ leftBinder = "ADJACENT_LINE_COMMENTS" ]
   mixin = "org.rust.lang.core.psi.ext.RsMacroCallImplMixin"
   stubClass = "org.rust.lang.core.stubs.RsMacroCallStub"
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-MacroCallNoSemicolons ::= MacroHead {
-  pin = 2
+// Needs semicolon `foo();`
```
Undin:
```suggestion
// Parses top-level macro calls with semicolon like `foo();`
```

src/main/grammars/RustParser.bnf
```diff
@@ -1319,41 +1319,42 @@ Macro2 ::= AttrsAndVis MACRO_KW identifier ( Macro2FunctionLikeBody | Macro2Matc
...
-// We parse some macros (like `println`) as a specific syntax constructions. This
-// is a list of such specific macros that should be always parsed as `MacroExpr`.
-// E.g. `println!();` should be parsed as `ExprStmt(MacroExpr(MacroCall(...)))`, but
-// any other `foo!();` as a simple `MacroCall(...)`
-private SpecialExprContextMacroNames ::= "try" | "await" | "dbg" | "format" | "format_args" | "write" |
-"writeln" | "print" | "println" | "eprint" | "eprintln" | "panic" | "unimplemented" | "unreachable" |
-"assert" | "debug_assert" | "assert_eq" | "assert_ne" | "debug_assert_eq" | "debug_assert_ne" |
-"trace" | "log" | "warn" | "debug" | "error" | "info" | "include_str" | "include_bytes" | "concat"
+// Needs semicolon `foo();`. Differs from `ItemLikeMacroCall` in that it does not recoved on a missed semicolon.
+// It is needed to parse expressions like `foo!() + 1`.
```
Undin:
```suggestion
// Parses macro calls with semicolon like `foo();`. Unlike `ItemLikeMacroCall` rule, it does not recover on a missing semicolon.
// This rule is required to distinguish statements from expressions like `foo!() + 1` inside blocks.
```

src/main/grammars/RustParser.bnf
```diff
@@ -1319,41 +1319,42 @@ Macro2 ::= AttrsAndVis MACRO_KW identifier ( Macro2FunctionLikeBody | Macro2Matc
...
+BlockElementMacroCall ::= <<parseMacroCall 'MacroCallParsingMode.BLOCK'>> {
+  elementType = MacroCall
+  hooks = [ leftBinder = "ADJACENT_LINE_COMMENTS" ]
+  name = "macro call"
+}
 
-private MacroHead ::= AttrsAndVis &(PathWithoutTypes '!') (SpecialMacro | CommonMacro)
+ExprLikeMacroCall ::= <<parseMacroCall 'MacroCallParsingMode.EXPR'>> {
```
Undin: Add doc for this rule too

PR #4493, opened 2019-10-03
Use performance optimizations provided by FileTypeDetector API not to slow down IDEs using TOML plugin

valich:
I suspect TOML plugin to be responsible for numerous IDE freezes due to our poor file type matching API. Namely, all FileTypeDetectors which do not implement `getDetectedFileTypes` take part in all `isFileOfType` checks by eating loading file content, which is loaded lazily.

This PR fixes the behaviour.

### DIFF
--- a/intellij-toml/src/main/kotlin/org/toml/lang/TomlFileTypeDetector.kt
+++ b/intellij-toml/src/main/kotlin/org/toml/lang/TomlFileTypeDetector.kt
@@ -12,8 +12,12 @@ import com.intellij.openapi.vfs.VirtualFile
 import org.toml.lang.psi.TomlFileType
 
 class TomlFileTypeDetector : FileTypeRegistry.FileTypeDetector {
-    override fun getVersion(): Int = 1
+    override fun getVersion() = 1
 
     override fun detect(file: VirtualFile, firstBytes: ByteSequence, firstCharsIfText: CharSequence?): FileType? =
         if (file.name == "config" && file.parent?.name == ".cargo") TomlFileType else null
+
+    override fun getDetectedFileTypes() = listOf(TomlFileType)
+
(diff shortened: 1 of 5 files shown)


intellij-toml/src/main/kotlin/org/toml/lang/TomlFileTypeDetector.kt
```diff
@@ -12,8 +12,12 @@ import com.intellij.openapi.vfs.VirtualFile
...
 class TomlFileTypeDetector : FileTypeRegistry.FileTypeDetector {
-    override fun getVersion(): Int = 1
+    override fun getVersion() = 1
 
     override fun detect(file: VirtualFile, firstBytes: ByteSequence, firstCharsIfText: CharSequence?): FileType? =
         if (file.name == "config" && file.parent?.name == ".cargo") TomlFileType else null
+
+    override fun getDetectedFileTypes() = listOf(TomlFileType)
+
+    override fun getDesiredContentPrefixLength() = 0
```
Undin: Specify return type explicitly here and above, please

valich: Done. I am curious what is rationale behind such decision? https://kotlinlang.org/docs/reference/coding-conventions.html has examples with the opposite.

Undin: I personally prefer explicit returning types for functions because it simplifies readability and prevents unexpected errors when you change the expression type and as a result function returning type is also changed.
Of course, in case of overridden method, the last argument not relevant so it's just for consistency here

Also, Rust uses the same rules :)

> https://kotlinlang.org/docs/reference/coding-conventions.html

Interesting. But I didn't find explicit conventions about returning type - only examples for other things

Issue #4510, opened 2019-10-09
async missing when implementing missing fn


krojew:
## Environment

* **IntelliJ Rust plugin version:** 0.2.107.2137-192
* **Rust toolchain version:** 1.40 nightly
* **IDE name and version:** CLion 2019.2.3
* **Operating system:** Windows 10

## Problem description

When using the auto generation of missing fn in trait implementations, the async keyword is ignored.

## Steps to reproduce

```rust
use async_trait::async_trait;

#[async_trait]
trait Widget {
    async fn foo() {}
}

struct Gadget;

#[async_trait]
impl Widget for Gadget {
    // use "implement methods" here and choose "foo" - the generated code will not contain async
}
```

PR #4513, opened 2019-10-09
ACT: Take into account async keyword when implementing missing fn

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4510.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -36,7 +36,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
         // Bump this number if Stub structure changes
-        override fun getStubVersion(): Int = 181
+        override fun getStubVersion(): Int = 182
 
         override fun getBuilder(): StubBuilder = object : DefaultStubBuilder() {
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/ide/refactoring/implementMembers/ImplementMembersHandlerTest.kt
```diff
@@ -29,14 +29,14 @@ class ImplementMembersHandlerTest : RsTestBase() {
 
     private fun invokeVia(actionInvoker: () -> Unit) {
         checkByText("""
-            trait T { fn f1(); }
+            trait T { async unsafe fn f1(); }
```
Undin: Maybe it's better to create separate test for this case?

mchernyavsky: Fixed.

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -36,7 +36,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
         // Bump this number if Stub structure changes
-        override fun getStubVersion(): Int = 180
+        override fun getStubVersion(): Int = 181
```
Undin: Update it please according to the latest value in master

mchernyavsky: Fixed.

PR #4522, opened 2019-10-11
MACRO: rework macro-to-expansion linkage

vlad20012:
The problem: there is `RsMacroCall` and `ExpandedMacroInfo`.
We want reliably link them to each other.

Previous solution:
- `ExpandedMacroInfo` holds stub index (aka `ID`) of a macro call
  and a `SoftReference` to the macro call
- there is `rebind` function that fills in such soft references
- we use such soft references in most cases
- if we detect that references was lost (GCed), we do `rebind` again
- *sometimes* we dump stub indices (aka `ID`s) from macro calls
  (accessed trough sort references)

Issues with that solution:
- it's not clear when to call `rebind`. `SoftReference`s can
  be GCed at any time. We should detect it and call `rebind`
  in this case. It's difficult to make it reliable.
- it's not clear when we should dump stub indices (IDs).
  We should do it before PSI unloads (soft references GC).
  We did it while saving a document, but PSI may already be unloaded
  at that moment.

In short, the solution was unreliable: we lost macro
expansions sometimes.

New solution:
- `ExpandedMacroInfo` holds stub index (aka `ID`) of a macro call
  and a *strong* reference to the macro call (that is `null`
  most of the time)
- we use *stub indices* most of the time
- before PSI modification we store *strong* references and then,
  after the end of PSI modification, we switching back to stub
  references and releasing strong refs.

It solves both mentioned problems:
- no soft references no problems
- no soft references no problems

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -5,6 +5,7 @@
 
 package org.rust.lang.core.macros
 
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.WriteAction
 import com.intellij.openapi.application.runReadAction
 import com.intellij.openapi.diagnostic.Logger
@@ -21,7 +22,6 @@ import com.intellij.psi.PsiDocumentManager
 import com.intellij.psi.PsiManager
 import com.intellij.psi.StubBasedPsiElement
 import com.intellij.psi.impl.source.StubbedSpine
-import com.intellij.reference.SoftReference
 import com.intellij.util.indexing.FileBasedIndexScanRunnableCollector
 import gnu.trove.TIntObjectHashMap
 import org.rust.cargo.project.workspace.PackageOrigin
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/openapiext/Testmark.kt
```diff
@@ -88,3 +98,13 @@ fun Testmark.hitOnFalse(b: Boolean): Boolean {
     if (!b) hit()
     return b
 }
+
+interface TestmarkPred {
```
Undin: it looks quite suspicious

src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
```diff
@@ -245,21 +249,66 @@ class ExpandedMacroStorage(val project: Project) {
...
 }
 
+/**
+ * A [SourceFile] object is associated to any Rust file that contains macro calls. The main purpose
+ * of [SourceFile] is to map between macro calls and expansions ([RsMacroCall] <-> [ExpandedMacroInfoImpl]).
+ *
+ * We use 3 basic mechanisms:
+ * - **Stub indices**. Each stub has an index (aka *id*, incremental integer given in
+ *   DFS order; see [calcStubIndex]). A PSI element can be retrieved by such index if the PSI
+ *   file has not been changed since we got a stub index. We check that in is not changed
```
Undin:
```suggestion
 *   file has not been changed since we got a stub index. We check that it is not changed
```

src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
```diff
@@ -245,21 +249,66 @@ class ExpandedMacroStorage(val project: Project) {
...
+ *   by the modification stamp, see [getActualModStamp]
+ * - **Strong PSI refs**. If we know that some PSI file will be modified, we store a strong ref to each
+ *   known [RsMacroCall] in the file. Then, after the end of PSI modification, we recover stub indices
+ *   from stored refs. See [switchToStrongRefsTemporary]
+ * - **Recover**. If we lost refs, we can recover them. Each [ExpandedMacroInfoImpl] stores hashes of
```
Undin: It would be great to describe when this is possible

PR #4542, opened 2019-10-18
MACRO: support `env!("OUT_DIR")` in include macro calls

### LINKED ISSUE #1908: Handle include!(concat!(env!("OUT_DIR"), "/bindings.rs"));

Undin:
Under the hood, it uses output of `cargo build --build-plan` command to get value of `OUT_DIR` variable

Fixes #1908


Undin:
I've filed a separate issue #4546 related to the corresponding user notification about the absence of expected generated files

src/main/kotlin/org/rust/lang/core/psi/ext/RsMacroCall.kt
```diff
@@ -105,7 +111,8 @@ fun RsMacroCall.findIncludingFile(): RsFile? {
     val path = includeMacroArgument?.expr?.value ?: return null
     // TODO: it doesn't work if `include!()` macro call comes from other macro
     val file = containingFile?.originalFile?.virtualFile ?: return null
-    return file.parent?.findFileByRelativePath(path)?.toPsiFile(project)?.rustFile
+    // It's important to refresh before find because file can be created externally (for example, while compilation)
+    return file.parent?.refreshAndFindFileByMaybeRelativePath(path)?.toPsiFile(project)?.rustFile
```
vlad20012: `refreshAndFindFileByPath` can be invoked outside of the read action or from EDT, otherwise, it can lead to deadlock. But looks like this code can be invoked from background thread in the read action

Undin: Nice catch!
Thanks

Undin:
@vlad20012 I've fixed all discussed issues

CAD97:
Though I should drop this here:

I've [requested a way to get `OUT_DIR` from `cargo metadata`](https://github.com/rust-lang/cargo/issues/7546) because [`cargo build --build-plan` always causes a full rebuild after](https://github.com/rust-lang/cargo/issues/5579#issuecomment-546760770).

src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -94,12 +100,49 @@ class Cargo(private val cargoExecutable: Path) {
...
+
+    private fun fetchBuildPlan(
+        owner: Project,
+        projectDirectory: Path,
+        listener: ProcessListener?
+    ): CargoBuildPlan? {
+        val additionalArgs = mutableListOf("-Z", "unstable-options", "--all-targets", "--build-plan")
+        // Hack to make cargo think that current channel is nightly because we need unstable `--build-plan` option here
+        val envs = EnvironmentVariablesData.create(mapOf(
+            RUSTC_BOOTSTRAP to "1"
```
CAD97: 😨

bjorn3: This means that this functionality can break at any moment.

vlad20012: We are closely monitoring changes in the rust toolchain and ready for breaking changes in the build-plan or other nightly features that we use. We'd want a more explicit way to unlock all nightly command-line options on the stable toolchain but now there is only "RUSTC_BOOTSTRAP" magic.

vlad20012:
@CAD97 thanks for the comment! Now we're going to disable the feature because of this.

bjorn3:
@vlad20012 Maybe enable it only when the user is actually using a nightly compiler.

vlad20012:
@bjorn3 Umm, I'm about the issue that build-plan invocation forces cargo to make full rebuild on the next build. I don't see differences between stable or nightly compiler in this case. But we'll make some option to enable this feature back

Issue #4144, opened 2019-07-14
Add an advanced support for paths in Cargo.toml [feature request]


alexander-irbis:
Add the validation, completion and navigation for dependencies paths in Cargo.toml.

As it is common to use relative paths in dependencies as it is common to mistake in such paths.
The error messages in such cases are also not so informative to easily fix path errors.
On the other side, it should be easy for IDE to suggest the list of right paths for a cursor position, or to mark the part of the path which was not found.

Also it may be very useful to navigate from the current path segment to the `1. Project` pane when it is available there.

```toml
[dependencies]
wip-crate = { path = "../../my-fork|s/wip-crate" }
```

PR #4569, opened 2019-10-27
TOML: highlight escape symbols

Undin:
Initial preparation to implement #4144

| Before | After |
|-|-|
| image | image |

It also extracts common code between Rust and Toml plugins into new `common` module


build.gradle.kts
```diff
@@ -430,6 +453,14 @@ project(":intellij-toml") {
     }
 }
 
+project(":utils") {
```
Undin: Probably `utils` is not the best name

Undin:
bors r=vlad20012

PR #4573, opened 2019-10-28
MACRO: disable name resolution for include with env("OUT_DIR") by default

Undin:
Recently supported name resolution for included files uses `cargo build --build-plan` to fetch value of `OUT_DIR` environment variable (see #4542).
Found out that `cargo build --build-plan` force next `build` command to recompile whole project (including dependencies).
See https://github.com/rust-lang/cargo/issues/5579#issuecomment-546760770.
At this moment it is part of refresh, i.e. it is invoked on each `Cargo.toml` change, project reopening, etc. So this command is invoked quite frequently.
It can be rather unexpected for large project to make full compilation after each project reopening

So we decided to disable it by default until it will be fixed from cargo side or we'll find better way to get `OUT_DIR` value

### DIFF
--- a/src/192/main/kotlin/org/rust/openapiext/compatUtils.kt
+++ b/src/192/main/kotlin/org/rust/openapiext/compatUtils.kt
@@ -8,3 +8,4 @@ package org.rust.openapiext
 import com.intellij.openapi.application.Experiments
 
 fun isFeatureEnabled(featureId: String): Boolean = Experiments.isFeatureEnabled(featureId)
+fun setFeatureEnabled(featureId: String, enabled: Boolean) = Experiments.setFeatureEnabled(featureId, enabled)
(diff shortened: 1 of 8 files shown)


onelson:
For anyone wondering about how to access the experimental features dialog, it's briefly mentioned here:

https://www.jetbrains.com/clion/features/run-and-debug.html

It's accessed via the _maintenance dialog_ which is summoned with ctrl+alt+shift+/

Undin:
Also, you can find almost any IDE action, setting, etc. including `Experimental Features` dialog via `Help | Find Action` (`Ctrl+Shift+A`) action

panstromek:
Or `shift + shift` to look for even more ;)

elrob:
https://github.com/Nercury/rust-and-opengl-lessons/issues/11#issuecomment-876741170

PR #4575, opened 2019-10-29
ANN: Make cfg-disabled edition 2018 keywords bold

ortem:
Screenshot 2019-10-29 at 13 31 26

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsEdition2018KeywordsAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsEdition2018KeywordsAnnotator.kt
@@ -6,6 +6,8 @@
 package org.rust.ide.annotator
 
 import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.openapi.editor.colors.EditorColorsManager
+import com.intellij.openapi.editor.markup.TextAttributes
 import com.intellij.psi.PsiElement
 import org.rust.ide.colors.RsColor
 import org.rust.ide.utils.isEnabledByCfg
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/annotator/RsEdition2018KeywordsAnnotator.kt
```diff
@@ -16,15 +16,20 @@ import org.rust.lang.core.psi.ext.*
...
         when {
             isEdition2018 && isIdentifier && isNameIdentifier(element) ->
                 holder.createErrorAnnotation(element, "`${element.text}` is reserved keyword in Edition 2018")
-            isEdition2018 && !isIdentifier ->
+
+            isEdition2018 && !isIdentifier && isEnabledByCfg ->
                 holder.createInfoAnnotation(element, null).textAttributes = RsColor.KEYWORD.textAttributesKey
+
+            isEdition2018 && !isIdentifier && !isEnabledByCfg ->
+                holder.createInfoAnnotation(element, null).textAttributes = RsColor.CFG_DISABLED_KEYWORD.textAttributesKey
```
Undin: It's rather weird that we introduce new color for this case.
Now we have separate `CFG_DISABLED_KEYWORD` color that changes color only for `async` and `try`.
I think we can merge `RsColor.KEYWORD` and `RsColor. CFG_DISABLED_CODE` manually and set the result via `com.intellij.lang.annotation.Annotation#setEnforcedTextAttributes`
Only question is how to get correct `TextAttributes` from `RsColor`

ortem: @Undin I've just realized that `CFG_DISABLED_KEYWORD` should not be bold for Dracula theme. So if we don't introduce the new color, we have to take into account the boldness with each `annotateInternal` call. I've pushed a possible solution, but now the attribute is cloned every time (because a user can switch editor theme at any time).

src/main/kotlin/org/rust/ide/annotator/RsEdition2018KeywordsAnnotator.kt
```diff
@@ -16,15 +17,27 @@ import org.rust.lang.core.psi.ext.*
...
         when {
             isEdition2018 && isIdentifier && isNameIdentifier(element) ->
                 holder.createErrorAnnotation(element, "`${element.text}` is reserved keyword in Edition 2018")
-            isEdition2018 && !isIdentifier ->
+
+            isEdition2018 && !isIdentifier && isEnabledByCfg ->
                 holder.createInfoAnnotation(element, null).textAttributes = RsColor.KEYWORD.textAttributesKey
+
+            isEdition2018 && !isIdentifier && !isEnabledByCfg -> {
+                val globalScheme = EditorColorsManager.getInstance().globalScheme
```
Undin: Looks like we need to get color scheme for current editor instead of global one

src/main/kotlin/org/rust/ide/annotator/RsEdition2018KeywordsAnnotator.kt
```diff
@@ -16,15 +17,27 @@ import org.rust.lang.core.psi.ext.*
...
+                val keywordTextAttributes = globalScheme.getAttributes(RsColor.KEYWORD.textAttributesKey)
+                val cfgDisabledKeywordTextAttributes = globalScheme.getAttributes(RsColor.CFG_DISABLED_CODE.textAttributesKey)
+                    .clone()
+                    .apply { fontType = keywordTextAttributes.fontType }
```
Undin: I think you need `com.intellij.openapi.editor.markup.TextAttributes#merge` here

src/main/kotlin/org/rust/ide/annotator/RsEdition2018KeywordsAnnotator.kt
```diff
@@ -16,15 +20,32 @@ import org.rust.lang.core.psi.ext.*
...
         when {
             isEdition2018 && isIdentifier && isNameIdentifier(element) ->
                 holder.createErrorAnnotation(element, "`${element.text}` is reserved keyword in Edition 2018")
-            isEdition2018 && !isIdentifier ->
+
+            isEdition2018 && !isIdentifier && isEnabledByCfg ->
                 holder.createInfoAnnotation(element, null).textAttributes = RsColor.KEYWORD.textAttributesKey
+
+            isEdition2018 && !isIdentifier && !isEnabledByCfg -> {
+                val fileEditor = FileEditorManager.getInstance(element.project).selectedEditor
```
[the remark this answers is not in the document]
ortem: Unfortunately, it requires EDT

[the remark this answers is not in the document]
Undin: Looks like `FileEditorManagerImpl` doesn't have the corresponding assertion by accident
Let's drop `FileEditorManager` usage and use `EditorColorsManager.getInstance().globalScheme` for now

PR #4581, opened 2019-10-31
RUN: Enable Code Coverage for CLion (+193)

mchernyavsky:


clion/src/193/main/resources/META-INF/platform-clion-only.xml
```diff
@@ -1,2 +1,3 @@
-<idea-plugin>
+<idea-plugin xmlns:xi="http://www.w3.org/2001/XInclude" allow-bundled-update="true">
+    <xi:include href="/META-INF/coverage-only.xml" xpointer="xpointer(/idea-plugin/*)"/>
```
Undin: Let's avoid such unexpected mixes of plugin xmls if possible
Looks like it's enough to change that `coverage-only.xml` depends on.
At this moment we use `Coverage` id to add an optional dependency on coverage plugin in root `plugin.xml` but if you convert it to module dependency (`com.intellij.modules.coverage` id), everything will work as expected
Also, don't forget to update the corresponding dependency in `plugin.xml` in `coverage/src/test/resources/META-INF`

gradle-193.properties
```diff
@@ -1,8 +1,8 @@
-ideaVersion=IU-193.4778-EAP-CANDIDATE-SNAPSHOT
-clionVersion=CL-193.4778-EAP-CANDIDATE-SNAPSHOT
+ideaVersion=IU-193.4932.9-EAP-SNAPSHOT
+clionVersion=CL-193.4932.12-EAP-SNAPSHOT
```
Undin: Let's better use `EAP-CANDIDATE-SNAPSHOT` artifacts because sometimes there are some problems with sources code downloading for concrete EAP version

gradle-193.properties
```diff
@@ -1,8 +1,8 @@
...
 
 psiViewerPluginVersion=193-SNAPSHOT
 
 # please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description
-sinceBuild=193.4386
+sinceBuild=193.4932
```
Undin: Will be redundant if use module dependency for coverage integration

mchernyavsky:
Fixed.

Issue #2186, opened 2018-01-10
Improve "extract function" dialog


Undin:
Current "extract function" dialog is not ideal and can be improved:

* Don't allow press "OK" button if function name is empty it leads to exception
* Show simple text field if we can't suggest function name instead of empty drop down list 
* Allow to manually select arguments
* Reformat code in "Signature" field (and maybe drop function body at all)

Good example is kotlin "Extract function" dialog
<img width="537" alt="screen shot 2018-01-11 at 00 14 55" src="https://user-images.githubusercontent.com/2539310/34795786-8054f438-f664-11e7-84ed-2cfcd47596d8.png">

Kobzol:
Should we also support the last item on the list, `Allow changing argument's type`?

For example, the Kotlin plugin is pretty restrictive and doesn't allow type changes apart from subtyping. It also disallows removing the parameters.

We already allow removing the parameter though, so maybe we should also allow changing the type. It shouldn't be hard to implement, we could just allow the user to type in the new type, although it's pretty easy to break the extracted code with that. I will try to implement it if you think that it's a valid option to offer to the users.

PR #4619, opened 2019-11-11
Allow to manually select arguments

Stzx:
:)

Edited: relates to https://github.com/intellij-rust/intellij-rust/issues/2186

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
@@ -8,6 +8,8 @@ package org.rust.ide.refactoring.extractFunction
 import com.intellij.openapi.project.Project
 import com.intellij.refactoring.util.AbstractParameterTablePanel
 import com.intellij.refactoring.util.AbstractVariableData
+import com.intellij.ui.BooleanTableCellEditor
+import com.intellij.ui.BooleanTableCellRenderer
 import com.intellij.util.ui.ColumnInfo
 
 class ParameterDataHolder(val parameter: Parameter, val onChange: () -> Unit) : AbstractVariableData() {
@@ -15,6 +17,20 @@ class ParameterDataHolder(val parameter: Parameter, val onChange: () -> Unit) : 
         parameter.name = name
         onChange()
     }
+}
+
+class ChooseColumn : ColumnInfo<ParameterDataHolder, Boolean>(null) {
+
+    override fun valueOf(item: ParameterDataHolder): Boolean = item.parameter.isSelected
+
+    override fun setValue(item: ParameterDataHolder, value: Boolean) {
+        item.parameter.isSelected = value
+    }
+
+    override fun getColumnClass(): Class<*> = Boolean::class.java
+
+    override fun isCellEditable(item: ParameterDataHolder?): Boolean = true
+
 }
 
 class NameColumn(private val nameValidator: (String) -> Boolean) : ColumnInfo<ParameterDataHolder, String>("Name") {
@@ -44,10 +60,16 @@ class ExtractFunctionParameterTablePanel(nameValidator: (String) -> Boolean,
                                          private val config: RsExtractFunctionConfig,
                                          private val onChange: () -> Unit)
     : AbstractParameterTablePanel<ParameterDataHolder>(
+    ChooseColumn(),
     NameColumn(nameValidator),
     TypeColumn(project)
 ) {
     init {
+        myTable.setDefaultRenderer(Boolean::class.java, BooleanTableCellRenderer())
+        myTable.setDefaultEditor(Boolean::class.java, BooleanTableCellEditor())
+        myTable.columnModel.getColumn(0).preferredWidth = WIDTH
+        myTable.columnModel.getColumn(0).maxWidth = WIDTH
+
         init(config.parameters.map {
             ParameterDataHolder(it) {
                 updateSignature()
@@ -60,5 +82,9 @@ class ExtractFunctionParameterTablePanel(nameValidator: (String) -> Boolean,
     override fun updateSignature() {
         config.parameters = variableData.map { it.parameter }
         onChange()
+    }
+
+    companion object {
+        private const val WIDTH = 40
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/refactoring/extractFunction/RsExtractFunctionConfig.kt
```diff
@@ -127,7 +128,10 @@ class RsExtractFunctionConfig private constructor(
             val body = if (single is RsBlockExpr) {
                 single.block.text
             } else {
-                val stmts = elements.map { it.text }.toMutableList()
+                val stmts = parameters.filter { !it.isSelected }.map { "let ${it.name}: ${it.type} = ;" }.toMutableList()
```
Stzx: Maybe `${it.type}::default()` is better?

src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
```diff
@@ -17,6 +23,57 @@ class ParameterDataHolder(val parameter: Parameter, val onChange: () -> Unit) :
...
+            item.onChange()
+        }
+    }
+
+    override fun getColumnClass(): Class<*> = JBCheckBox::class.java
+
+    override fun isCellEditable(item: ParameterDataHolder?): Boolean = true
+}
+
+class ChooseRenderer : TableCellRenderer {
```
ortem: Looks like you can use `BooleanTableCellRenderer` instead of introducing your own renderer

src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
```diff
@@ -17,6 +23,57 @@ class ParameterDataHolder(val parameter: Parameter, val onChange: () -> Unit) :
...
+
+        return (value as JBCheckBox).apply { background = table.background }
+    }
+
+    companion object{
+        private const val WIDTH = 40
+    }
+}
+
+class ChooseEditor(private var isSelected: Boolean = false) : AbstractCellEditor(), TableCellEditor {
```
ortem: Looks like you can use `BooleanTableCellEditor` instead of introducing your own editor

src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
```diff
@@ -17,6 +23,57 @@ class ParameterDataHolder(val parameter: Parameter, val onChange: () -> Unit) :
     }
 }
 
+class ChooseColumn : ColumnInfo<ParameterDataHolder, JBCheckBox>(null) {
+
+    override fun valueOf(item: ParameterDataHolder): JBCheckBox? = JBCheckBox(null, item.parameter.isSelected).apply {
```
ortem: Probably it should look like this:

```Kotlin
override fun valueOf(item: ParameterDataHolder): Boolean = item.parameter.isSelected

override fun setValue(item: ParameterDataHolder, value: Boolean) {
    item.parameter.isSelected = value
}
```

Please try this way and let me know if there are any problems.

src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
```diff
@@ -17,6 +23,57 @@ class ParameterDataHolder(val parameter: Parameter, val onChange: () -> Unit) :
     }
 }
 
+class ChooseColumn : ColumnInfo<ParameterDataHolder, JBCheckBox>(null) {
```
ortem: Maybe use `Boolean` instead of `JBCheckBox`? Please check `com.intellij.util.ui.ColumnInfo` usages and find a proper way.

src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
```diff
@@ -44,10 +101,14 @@ class ExtractFunctionParameterTablePanel(nameValidator: (String) -> Boolean,
                                          private val config: RsExtractFunctionConfig,
                                          private val onChange: () -> Unit)
     : AbstractParameterTablePanel<ParameterDataHolder>(
+    ChooseColumn(),
     NameColumn(nameValidator),
     TypeColumn(project)
 ) {
     init {
+        myTable.setDefaultRenderer(JBCheckBox::class.java, ChooseRenderer())
```
ortem: Maybe override `org.rust.ide.refactoring.extractFunction.ChooseColumn#getRenderer` instead of call `setDefaultRenderer`? As well as editor.

src/main/kotlin/org/rust/ide/refactoring/extractFunction/RsExtractFunctionConfig.kt
ortem: IMO the type should not be inserted explicitly. And also check `InitializeWithDefaultValueFix`, maybe you can reuse it somehow.

src/main/kotlin/org/rust/ide/refactoring/extractFunction/ExtractFunctionParameterTablePanel.kt
Stzx: I tried override the `org.rust.ide.refactoring.extractFunction.ChooseColumn#getRenderer` method at first, but it didn't work.

Am I missing something?

ortem: You're right, it doesn't work for some reason. Well, I think this code could look like this:
```Kotlin
myTable.setDefaultRenderer(Boolean::class.java, BooleanTableCellRenderer())
myTable.setDefaultEditor(Boolean::class.java, BooleanTableCellEditor())
myTable.columnModel.getColumn(0).preferredWidth = WIDTH
```

src/main/kotlin/org/rust/ide/refactoring/extractFunction/RsExtractFunctionConfig.kt
Stzx: Sorry for so long.

I don't seem to find a more convenient way to reuse the contents of `InitializeWithDefaultValueFix`, maybe there are other ways?

ortem: OK, let's just use `let ${it.name}: ${it.type};`, without `=`. Then a user can apply `InitializeWithDefaultValueFix` when the caret is located under any usage of the variable.

ortem:
Please squash the commits into one

ortem:
Thank you!
bors r+

Issue #4636, opened 2019-11-14
Compile error location is not shown in Build window. Possible error parse problem.


extremegf:
## Environment

* **IntelliJ Rust plugin version:**
Plugin built from master@ 3258d9187 (today 2019-11-14 12:00 CET)
* **Rust toolchain version:**
stable-x86_64-unknown-linux-gnu (default)
rustc 1.39.0 (4560ea788 2019-11-04)
* **IDE name and version:**
IntelliJ IDEA 2019.2.4 (Community Edition)
Build #IC-192.7142.36, built on October 29, 2019
Runtime version: 11.0.4+10-b304.77 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.
* **Operating system:**
Linux 4.15.0-66-generic (Mint)

## Problem description
Compile error location is not shown in Build window. Also the error is not highlighted by Cargo Check Linter. 

![Screenshot from 2019-11-14 11-57-20](https://user-images.githubusercontent.com/1788593/68851467-4cd4bd80-06d6-11ea-85e2-daa7c4f976c4.png)
## Steps to reproduce
1. Create new rust library project.
2. Paste following code in lib.rs:
```rust
enum Enum {
    Variant,
}

fn test_root_get() {
    assert_eq!(Enum::Variant, Enum::Variant);
}
```
3. Run cargo test operation.
4. Observe in the build window, that the error location isn't properly recognized.

PR #4638, opened 2019-11-14
RUN & INSP: Fix error span in rustc messages if the error comes from macro expansion

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4636.


mchernyavsky:
bors retry

PR #4640, opened 2019-11-15
MACRO&ANN: enable advanced highlighting inside macro call bodies

### LINKED ISSUE #3706: Rust code in macros is not syntax highlighted
## Environment

* **Intellij-Rust plugin version:** v0.2.96.2122-191
* **Rust toolchain version:** rustc 1.33.0 (2aa4c46cf 2019-02-28)
* **IDE name and version:** CLion 2019.1 Build #CL-191.6183.77, built on March 25, 2019

## Problem description
Passing a Rust code to declarative macro results in missing syntax highlighting and code completion. 

However when the code passed to the macro has syntax error in it (eg. unbalanced parentheses), the highlighting is back and it seems that the IDE has once again clue what the types are.

## Steps to reproduce

1. Create new project
2. Add https://docs.rs/compact_arena/0.2.1/compact_arena/ crate to project.
3. Write example code. 
4. `assert_eq` should be blue but actually is white.

image

5. We create a syntax error and `assert_eq` is suddenly blue.

image

vlad20012:
Closes #3706

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
@@ -1,0 +1,161 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator
+
+import com.intellij.codeHighlighting.DirtyScopeTrackingHighlightingPassFactory
+import com.intellij.codeHighlighting.TextEditorHighlightingPass
+import com.intellij.codeHighlighting.TextEditorHighlightingPassRegistrar
+import com.intellij.codeInsight.daemon.impl.*
+import com.intellij.lang.annotation.Annotation
+import com.intellij.lang.annotation.AnnotationSession
+import com.intellij.lang.annotation.Annotator
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.progress.ProgressIndicator
+import com.intellij.openapi.progress.ProgressManager
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.ide.utils.isEnabledByCfg
+import org.rust.lang.core.macros.*
+import org.rust.lang.core.psi.RsMacroCall
+import org.rust.lang.core.psi.ext.descendantsOfType
+import org.rust.lang.core.psi.ext.expansion
+import org.rust.stdext.removeLast
+
+class RsMacroExpansionHighlightingPassFactory(
+    val project: Project,
+    registrar: TextEditorHighlightingPassRegistrar
+) : DirtyScopeTrackingHighlightingPassFactory {
+    private val myPassId: Int = registrar.registerTextEditorHighlightingPass(
+        this,
+        null,
+        null,
+        false,
+        -1
+    )
+
+    override fun createHighlightingPass(file: PsiFile, editor: Editor): TextEditorHighlightingPass? {
+        if (project.macroExpansionManager.macroExpansionMode !is MacroExpansionMode.New) return null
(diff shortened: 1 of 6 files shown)


nczhu:
Will there be more work on this? This is a super relevant issue still
thanks!

src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
```diff
@@ -0,0 +1,153 @@
...
+            for (element in macro.elementsForHighlighting) {
+                for (ann in annotators) {
+                    ProgressManager.checkCanceled()
+                    ann.annotate(element, holder)
+                }
+                if (element is RsMacroCall) {
+                    macros += element.prepare() ?: continue
+                }
+            }
+            mapAndCollectAnnotations(macro, holder)
```
Undin, line 85: It seems we can improve performance from `O(annotations * ranges)` to `O(annotations * log(annotations) + ranges * log(ranges) + annotations + ranges)` if we sort annotations and expansion ranges and when make linear pass through two lists
But it would be great to measure somehow how mich time current implementation takes and decide if we need to rewrite our naive mapping

src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
```diff
@@ -0,0 +1,153 @@
...
+            mapAndCollectAnnotation(macro, ann)
+        }
+    }
+
+    private fun mapAndCollectAnnotation(macro: PreparedMacroCall, ann: Annotation) {
+        val originRange = TextRange(ann.startOffset, ann.endOffset)
+        val originInfo = HighlightInfo.fromAnnotation(ann)
+        val mappedRanges = mapRangeFromExpansionToCallBody(macro.expansion, macro.macroCall, originRange)
+        for (mappedRange in mappedRanges) {
+            results += originInfo.copyWithRange(mappedRange)
```
Undin, line 101: Can we check here that there are two ranges with different attributes for the same offset? I think we should exclude them to avoid weird highlighting when the same token is annotated with different colors

src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
```diff
@@ -0,0 +1,159 @@
...
+    private val file: PsiFile,
+    private val restrictedRange: TextRange,
+    document: Document
+) : TextEditorHighlightingPass(file.project, document) {
+    private val results = mutableListOf<HighlightInfo>()
+
+    private fun createAnnotators(): List<Annotator> = listOf(
+        RsEdition2018KeywordsAnnotator(),
+        RsHighlightingAnnotator(),
+        RsHighlightingMutableAnnotator()
```
Undin, line 61: Let's also add `RsCfgDisabledCodeAnnotator` in this list

src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
vlad20012: This requires complex data structures to be solved (or I need more time to think about it), to let's fox it later. I'll open an issue about it

src/test/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPassTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import org.intellij.lang.annotations.Language
+import org.rust.ExpandMacros
+
+// More tests are located in `RsHighlightingAnnotatorTest` (most of those tests are executed
+// in both plain and macro context)
+@ExpandMacros
+class RsMacroExpansionHighlightingPassTest : RsAnnotationTestBase() {
+
+    fun `test attributes inside macro call`() = checkHighlighting("""
```
Undin, line 16: Maybe also add test when some code is disabled with cfg attribute to ensure that we don't forget about `RsCfgDisabledCodeAnnotator`?

PR #4642, opened 2019-11-15
MACRO: fix re-highlighting after a macro call change (new engine)

vlad20012:
Previously, a file was sometimes not re-highlighted after a macro
call change (until caret move or some other action)


vlad20012:
bors retry

PR #4643, opened 2019-11-15
TOML: provide name resolution and completion for paths in Cargo.toml

### LINKED ISSUE #3551: Path-based dependencies do not show up in completion
If a dependency is specified using a path (e.g. `lib1 = { path = "../lib1" }`), it's not suggested in autocompletion, i.e. when I type "li" in the code, "lib1" is not suggested. But when I type "lib1::", the content of `lib1` is suggested, so the IDE knows about the dependency. The issue happens in use statements as well (i.e. "use li" is not auto-completed). Crates.io-based dependencies (e.g. 
`rand = "0.6.5"`) work as expected.

Undin:
Fixes #3551
Partially implement #4144 (navigation and completion)


intellij-toml/src/main/kotlin/org/toml/lang/psi/TomlPsiFactory.kt
```diff
@@ -30,6 +30,8 @@ class TomlPsiFactory(private val project: Project, private val markGenerated: Bo
     private inline fun <reified T : PsiElement> PsiElement.descendantOfTypeStrict(): T? =
         PsiTreeUtil.findChildOfType(this, T::class.java, /* strict */ true)
 
+    fun createLiteral(value: String): TomlLiteral = createFromText("dummy = $value") ?: error("")
```
vlad20012: What's the difference from `createValue`?

Undin: Return type. But they indeed make the same things.
Maybe just drop `TomlPsiFactory#createValue`? We use it only in tests and have methods for all other `TomlValue`s: `TomlArray` and `TomlInlineTable`

toml/src/main/kotlin/org/rust/toml/CargoTomlFileReferenceProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+            WORKSPACE -> true to false
+            GENERAL -> {
+                val table = element.ancestorStrict<TomlHeaderOwner>()
+                val completeRustFiles = table != null && table.header.names.singleOrNull()?.text in TARGET_TABLE_NAMES
+                true to completeRustFiles
+            }
+            BUILD -> false to true
+        }
+
+        return if (kind is TomlLiteralKind.String) CargoTomlFileReferenceSet(
```
vlad20012: braces?

Undin:
bors r=vlad20012

PR #4659, opened 2019-11-19
GRAM&MACRO: recover parser on macros with unbalanced braces

vlad20012:
Works for `foo! { foo(a }` or other kinds of unbalanced macro bodies


vlad20012:
bors r=undin

PR #4671, opened 2019-11-22
INT: Fixed move type constraints intentions for default type parameters

rrevenantt:

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
@@ -10,6 +10,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.RsBaseType
 import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.RsTypeParameter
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/intentions/MoveTypeConstraintToParameterListIntention.kt
```diff
@@ -55,6 +56,10 @@ class MoveTypeConstraintToParameterListIntention : RsElementBaseIntentionAction<
             if (bs.isNotEmpty()) {
                 append(bs.joinToString(separator = "+", prefix = ":") { it.text })
             }
+            if (param is RsTypeParameter && param.typeReference != null) {
+                val default = (param as? RsTypeParameter)?.typeReference?.text
+                if (default != null) append("=$default")
+            }
```
mchernyavsky: You don't need the `if`. You can replace
```kotlin
if (param is RsTypeParameter && param.typeReference != null) {
    val default = (param as? RsTypeParameter)?.typeReference?.text
    if (default != null) append("=$default")
}
```
with just
```kotlin
val default = (param as? RsTypeParameter)?.typeReference?.text
if (default != null) append("=$default")
```

rrevenantt: My bad. Fixed.

mchernyavsky:
Thanks!

bors r+

Issue #3985, opened 2019-06-07
Support const generics


akofke:
Now that const generics (rust-lang/rust#44580) is usable in nightly, it would be great if it could be supported in intellij-rust, at least the syntax if not completion etc. Right now they don't seem to parse correctly and produce errors including the two shown below as well as breaking indentation in struct declarations. This code compiles on recent nightlies (2019-06-05).

<img width="328" alt="Screen Shot 2019-06-07 at 7 59 56 AM" src="https://user-images.githubusercontent.com/6622795/59102828-3c384400-88fb-11e9-8c64-460d784c64e3.png">
<img width="333" alt="Screen Shot 2019-06-07 at 8 04 30 AM" src="https://user-images.githubusercontent.com/6622795/59102835-3fcbcb00-88fb-11e9-9101-b742d8ffa8a7.png">

vlad20012:
Ooooooooooooohhhhhhhhhhhh noooooooooooooooo 😞😞😞

vlad20012:
@mchernyavsky would you like to implement it ?)?)?) :):):)

kmaork:
Hey, what's the status with this issue?

kmaork:
Hey @mchernyavsky @vlad20012 , what's the status of this issue? Can I offer help in order to resolve it?

Undin:
https://blog.rust-lang.org/inside-rust/2021/09/06/Splitting-const-generics.html

PR #4677, opened 2019-11-22
TY: Check type of const generic argument

mchernyavsky:
Relates to https://github.com/intellij-rust/intellij-rust/issues/3985.


src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
```diff
@@ -334,6 +334,7 @@ class RsTypeInferenceWalker(
                         subst = subst,
                         source = TraitImplSource.Collapsed((resolved.owner as RsAbstractableOwner.Trait).trait)
                     )
+                    inferConstArgumentTypes(resolved.constParameters, path.constArguments)
```
vlad20012: Let's move it into `instantiatePath`

PR #4687, opened 2019-11-25
Fix control flow of some elements

ortem:
This allows borrow checker to find move errors in some new cases. Also, these fixes will be necessary for liveness analysis.

Merge after #4686.

- [x] Run `RsRealProjectAnalysisTest` before merge


ortem:
bors retry

PR #4710, opened 2019-12-02
INSP: Add liveness analysis

### LINKED ISSUE #4795: False positive: "Cannot move" error when dereferencing raw pointers
## Environment

* **IntelliJ Rust plugin version:** 0.2.112.2148-193
* **Rust toolchain version:** 1.40.0 (73528e339 2019-12-16) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2019.3 (CL-193.5233.103)
* **Operating system:** Linux 5.4.2-1-MANJARO

## Problem description

Reading or writing through a raw pointer results in a "Cannot move" error, which is a false positive in this case.

## Steps to reproduce

```rust
struct Node<T> {
    data: T,
    next: *mut Self,
}

impl<T> Node<T> {
    unsafe fn next(node: *mut Self) -> *mut Self {
        (*node).next // <-- false error here
    }

    unsafe fn set_next(node: *mut Self, next: *mut Self) {
        (*node).next = next; // <-- false error here
    }
}
```

ortem:
Related to #3163
Fixes #4795

- [x] Run RsRealProjectAnalysisTest before merge

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsLivenessInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsLivenessInspection.kt
@@ -1,0 +1,71 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.psi.PsiElement
+import org.rust.ide.injected.isDoctestInjection
+import org.rust.ide.inspections.fixes.RenameFix
+import org.rust.ide.utils.isCfgUnknown
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.descendantsWithMacrosOfType
+import org.rust.lang.core.psi.ext.expansion
+import org.rust.lang.core.types.DeclarationKind
+import org.rust.lang.core.types.DeclarationKind.Parameter
+import org.rust.lang.core.types.DeclarationKind.Variable
+import org.rust.lang.core.types.liveness
+
+class RsLivenessInspection : RsLintInspection() {
+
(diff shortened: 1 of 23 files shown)


src/main/kotlin/org/rust/ide/inspections/RsLivenessInspection.kt
```diff
@@ -0,0 +1,92 @@
...
+        binding: RsPatBinding,
+        name: String,
+        kind: DeclarationKind
+    ) {
+        if (!binding.isPhysical) return
+
+        if (binding.isCfgUnknown) return
+
+        // TODO: remove this check when multi-resolve for `RsOrPats` is implemented
+        if (binding.ancestorStrict<RsOrPats>() != null) return
```
Undin: Do we have test for it?

ortem:
@Undin thank you for noticing that the analysis doesn't handle such control flow properly. There was a severe mistake in my implementation that I made when dealing with false-positives on real projects. Now it's fixed, so the analysis works in such cases.

src/main/kotlin/org/rust/ide/inspections/RsLivenessInspection.kt
```diff
@@ -0,0 +1,71 @@
...
+        RsLint.UnusedVariables
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitFunction(func: RsFunction) {
+                // Disable inside doc tests
+                if (func.isDoctestInjection) return
+
+                // Don't analyze functions with unresolved macro calls
+                if (func.descendantsWithMacrosOfType<RsMacroCall>().any { it.expansion == null }) return
```
Undin: @vlad20012 looks like you had some questions about this

vlad20012: I rethought it. Now I think everything is fine here

Undin:
bors retry

PR #4737, opened 2019-12-07
Enable running the Rust toolchain through WSL

akainth015:
This pull request will allow the Rust toolchain to be run through the Windows Subsystem for Linux. It does this by abstracting the `RustToolchain` and implementing it in the `NativeRustToolchain` (Windows & Unix) and `WslRustToolchain` (Linux on Windows) classes.


akainth015:
One architectural change I made is that the "binary classes" (e.g. `Cargo`, `Rustup`) now perform their work through the `RustToolchain` class to reduce the repetition of various logics. The next step is converting Windows paths to WSL paths, which will likely be done by implementing a `WslCargo` subclass of `Cargo`, which will override each path-sensitive method and then call it as follows.
~~~kotlin
override fun init(..., directory: VirtualFile, ...): GeneratedFilesHolder {
  val mapped = "bash.exe -c wslpath directory.pathAsPath()"
  super.init(..., mapped, ...)
}
~~~

akainth015:
Decided not to proceed with the `WslCargo` architecture; it would have been a lot of overriding. Instead, the `createGeneralCommandLine` method will convert any Windows paths (`C:\Users\`) to their WSL counterpart (`/mnt/c/Users/`). Additionally, the `processOutput` method will convert WSL paths to their respective Windows counterparts.

akainth015:
Here is a picture of the default Rust project compiled and executed with WSL!

image

Issue #4746, opened 2019-12-12
Detect #[tokio::test] attribute as testable function


fujiapple852:
## Feature request

Tokio 0.2.x provides the [tokio::test](https://docs.rs/tokio/0.2.4/tokio/attr.test.html) attribute for defining async tests.  The plugin does not detect these as a testable functions, i.e. no green "run" arrow is shown in the IDE (tested in CLion).

The `#[tokio::test]` attribute is implemented as a procedural-macro which ultimately outputs a standard `#[test]` attribute (see [here](https://docs.rs/tokio-macros/0.2.0/src/tokio_macros/lib.rs.html#201)).  Given the plugin does not currently expand procedural macros it makes sense that it would not detect these.

Assuming support for procedural macro expansion is not imminent, it may be helpful to expand the detection logic for testable functions to cover attributes matching the pattern `#[.*::test]`.

Undin:
@fujiapple852 Thanks for the report and suggestion!
It's a really good idea. I think the pattern should be a little bit less restricted, something like `#[.*test]` to handle `#[cargo_test]` and similar attributes

PR #4758, opened 2019-12-16
RUN: match test attributes more loosely to support custom attributes

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/4746

I tried to loosen up the matching of `#[test]` attributes, as described in https://github.com/intellij-rust/intellij-rust/issues/4746.

To support `tokio`'s `#[tokio::test(threaded_scheduler)]` (https://github.com/tokio-rs/tokio/blob/master/tokio-macros/src/lib.rs#L137), the check now tries to find any occurence of `test` inside the function attribute.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -105,8 +105,8 @@ class QueryAttributes(
         return attrs.any()
     }
 
-    fun hasMatchingAttribute(regex: Regex): Boolean {
-        val attrs = attrsByRegex(regex)
+    fun hasAttribute(attributeNameRe: Regex): Boolean {
+        val attrs = attrsByText(attributeNameRe)
         return attrs.any()
     }
 
@@ -183,9 +183,12 @@ class QueryAttributes(
     private fun attrsByName(name: String): Sequence<RsMetaItem> = metaItems.filter { it.name == name }
 
     /**
-     * Get a sequence of all attributes matching the given [regex]
+     * Get a sequence of all attributes that match the given [regex]
      */
-    private fun attrsByRegex(regex: Regex): Sequence<RsMetaItem> = metaItems.filter { it.name?.let { name -> regex.containsMatchIn(name) } ?: false }
+    private fun attrsByText(regex: Regex): Sequence<RsMetaItem> = metaItems.filter {
+        val name = it.text ?: return@filter false
+        name.matches(regex)
+    }
 
     override fun toString(): String =
         "QueryAttributes(${attributes.joinToString { it.text }})"
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -105,8 +105,8 @@ class QueryAttributes(
         return attrs.any()
     }
 
-    fun hasAttribute(attributeNameRe: Regex): Boolean {
-        val attrs = attrsByText(attributeNameRe)
+    fun hasAttribute(regex: Regex): Boolean {
+        val attrs = attrsByText(regex)
         return attrs.any()
     }
 
@@ -186,8 +186,8 @@ class QueryAttributes(
      * Get a sequence of all attributes that match the given [regex]
      */
     private fun attrsByText(regex: Regex): Sequence<RsMetaItem> = metaItems.filter {
-        val name = it.text ?: return@filter false
-        name.matches(regex)
+        val text = it.text ?: return@filter false
+        text.matches(regex)
     }
 
     override fun toString(): String =
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -105,6 +105,11 @@ class QueryAttributes(
         return attrs.any()
     }
 
+    fun hasMatchingAttribute(regex: Regex): Boolean {
+        val attrs = attrsByRegex(regex)
+        return attrs.any()
+    }
+
     fun hasAnyOfOuterAttributes(vararg attributes: String): Boolean {
         val outerAttrList = (psi as? RsOuterAttributeOwner)?.outerAttrList ?: return false
         return outerAttrList.any { it.metaItem.name in attributes }
@@ -176,6 +181,11 @@ class QueryAttributes(
      * Get a sequence of all attributes named [name]
      */
     private fun attrsByName(name: String): Sequence<RsMetaItem> = metaItems.filter { it.name == name }
+
+    /**
+     * Get a sequence of all attributes matching the given [regex]
+     */
+    private fun attrsByRegex(regex: Regex): Sequence<RsMetaItem> = metaItems.filter { it.name?.let { name -> regex.containsMatchIn(name) } ?: false }
 
     override fun toString(): String =
         "QueryAttributes(${attributes.joinToString { it.text }})"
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -177,6 +182,11 @@ class QueryAttributes(
      */
     private fun attrsByName(name: String): Sequence<RsMetaItem> = metaItems.filter { it.name == name }
 
+    /**
+     * Get a sequence of all attributes matching the given [regex]
+     */
+    private fun attrsByRegex(regex: Regex): Sequence<RsMetaItem> = metaItems.filter { it.name?.let { name -> regex.containsMatchIn(name) } ?: false }
```
Undin: I'd prefer something like
```kotlin
private fun attrsByName(regex: Regex): Sequence<RsMetaItem> = metaItems.filter {
    val name = it.name ?: return@filter false
    regex.matches(name)
}
```
because:
* we still match only name
* we don't want to match some substring inside attribute `name`
* it's more readable

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -105,6 +105,11 @@ class QueryAttributes(
         return attrs.any()
     }
 
+    fun hasMatchingAttribute(regex: Regex): Boolean {
```
Undin:
```suggestion
    fun hasAttribute(antributeNameRe: Regex): Boolean {
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -29,7 +29,7 @@ val RsFunction.isTest: Boolean
     get() {
         val stub = greenStub
         return stub?.isTest
-            ?: (queryAttributes.hasAtomAttribute("test") || queryAttributes.hasAtomAttribute("quickcheck"))
+            ?: (queryAttributes.hasMatchingAttribute(Regex("test")) || queryAttributes.hasAtomAttribute("quickcheck"))
```
Undin: Let's extract `Regex("test")` into top-level property to avoid pattern compilation on each call

Undin: According to suggested changes in `attrsByRegex` and the fact that we don't want to accept any attribute with `test` substring in its name, let's change the pattern to `.*test`

Undin: This change modifies property that used in stub index creation.
So we need to rebuild indices to take into account new implementation.
To do it, you need to increase `org.rust.lang.core.stubs.RsFileStub.Type#getStubVersion`.
Don't forget to rebase this branch before making the corresponding change to get the actual value of stub version

src/test/kotlin/org/rust/ide/lineMarkers/CargoTestRunLineMarkerContributorTest.kt
```diff
@@ -201,6 +201,11 @@ class CargoTestRunLineMarkerContributorTest : RsLineMarkerProviderTestBase() {
         fn has_icon() { assert(true) } // - Test has_icon
     """)
 
+    fun `test custom test attribute`() = doTestByText("""
+        #[tokio::test(threaded_scheduler)]
```
Undin: Let's add more tests to check different custom attributes like `tokio::test`, `custom_test`

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
Undin: BTW, it doesn't work with `tokio::test` attribute because its name is null.
What's why the corresponding test fails

Kobzol:
Thanks for the commends @Undin. I'm not sure how to extract the path/name from `tokio::test`. Everything that I tried to access is null with this meta item, except for `RsCompactTT`, but I have no idea how to work with that. I could regex the `.text` attribute of the `RsMetaItem` directly, but that's probably not a good solution.

Undin:
@Kobzol Sorry for so long response

I understand your problem and it's because of our parsing:
* in case `#[foo]`, `foo` is parsed as identifier
* in case `#[foo::bar]`, `foo::bar` is parsed as list of tokens

I'm going to improve our parser rules to parse `foo::bar` as a path. it should solve your current problem and some other issues (for example, #4850). For now, I think it's ok to use the text of attribute. As you can see, we calculate it only when we don't have stub object, i.e. when we already have AST of whole file, so text extraction shouldn't be heavy

Kobzol:
OK, I used text to match the attribute. However when I use text, I get the full string, so I check for the `test` regex to be anywhere inside that string, not just at the end (to match both `#[tokio::test]` and `#[tokio::test(threaded_scheduler)]`).

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -22,14 +22,16 @@ import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
 import javax.swing.Icon
 
+val FUNCTION_TEST_REGEX = Regex(".*test.*")
```
Undin: Let's make it private

Undin: Unfortunately, it's too general and makes plugin provide test gutter icon for non-test attributes like https://github.com/tokio-rs/tokio/blob/71c47fabf4a8a450c3b41d58de304a6a49fb4061/tokio-macros/src/lib.rs#L59
Let's restrict it a little bit. For example, `^[\w:]*test.*` or something similar.
And let's add test(s) to check that we don't provide false run icon for some basic cases like I've pointed above

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -177,6 +182,14 @@ class QueryAttributes(
      */
     private fun attrsByName(name: String): Sequence<RsMetaItem> = metaItems.filter { it.name == name }
 
+    /**
+     * Get a sequence of all attributes that match the given [regex]
+     */
+    private fun attrsByText(regex: Regex): Sequence<RsMetaItem> = metaItems.filter {
+        val name = it.text ?: return@filter false
```
Undin:
```suggestion
        val text = it.text ?: return@filter false
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -105,6 +105,11 @@ class QueryAttributes(
         return attrs.any()
     }
 
+    fun hasAttribute(attributeNameRe: Regex): Boolean {
```
Undin:
```suggestion
    fun hasAttribute(regex: Regex): Boolean {
```

Kobzol: Ha, this parameter name was originally suggested by you xD https://github.com/intellij-rust/intellij-rust/pull/4758#discussion_r360304980

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
Kobzol: Good catch.

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
Undin: Oh, it's indeed funny 😆 
But as they say two hardest things in programming are cache invalidation and naming things.
In this case, it makes sense because now we work with attribute text instead of its name. When we improve attribute parsing, I'll rename it in a proper way

PR #4765, opened 2019-12-18
Ignore underscore names during duplicates detection

### LINKED ISSUE #5067: A second item with name '_' imported
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.2170-193
* **Rust toolchain version:** 1.41.1 (f3e1a954d 2020-02-24) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3.4 (CL-193.6494.38)
* **Operating system:** macOS 10.12.6

## Problem description

Two `as _` imports show the error:
scc1

## Steps to reproduce

1. VCS > Get from version control... > https://github.com/instrumentisto/medea
2. Let the project being indexed.
3. Open `src/api/control/mod.rs` file and see 14 line.

ortem:
This fixes false-positive duplicate annotation when there are multiple underscore imports, e.g.
```Rust
use std::fmt::Debug as _;
use std::hash::Hash as _; // False-positive: A second item with name '_' imported 
```

Note, there was a similar problem with `RsConstant`  after implementing underscore constants parsing: https://github.com/intellij-rust/intellij-rust/pull/4724. But that problem was solved in a quite different way – by introducing `nameLikeElement` to distinguish the actual name from the underscores.

However, in this case, looks like it's enough to just ignore the underscore names during duplicates detection in order to keep untouched all the existing logic in the name resolution and auto import. Thanks to @dima74 for finding this out!

Fixes #5067

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -1318,7 +1318,10 @@ private fun AnnotationSession.duplicatesByNamespace(
         (owner.namedChildren(recursively, stopAt = RsFnPointerType::class.java)
             + importedNames)
             .filter { it !is RsExternCrateItem } // extern crates can have aliases.
-            .filter { it.nameOrImportedName() != null }
+            .filter {
+                val name = it.nameOrImportedName()
+                name != null && name != "_"
+            }
             .filter { it.isEnabledByCfg && !it.isCfgUnknown }
             .flatMap { it.namespaced() }
             .groupBy { it.first }       // Group by namespace
--- a/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
@@ -4068,4 +4068,15 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
         impl ForeignTrait for Box<dyn LocalTrait> {}
         impl <error descr="Only traits defined in the current crate can be implemented for arbitrary types [E0117]">ForeignTrait</error> for Box<dyn ForeignTrait0> {}
     """)
+
+    fun `test no E0252 multiple underscore aliases`() = checkErrors("""
+        mod foo {
+            pub trait T1 {}
+        }
+        mod bar {
+            pub trait T2 {}
+        }
+        use foo::T1 as _;
+        use bar::T2 as _;
+    """)
 }


ortem:
@Undin I've changed the implementation: now there is `RsAlias.nameLikeText: String` instead of `RsAlias.nameLikeElement: RsElement`. If this solution looks OK for you, I'll replace `nameLikeElement` by `nameLikeText` in commit message and PR name before merge.

src/main/kotlin/org/rust/lang/core/psi/ext/RsAlias.kt
```diff
@@ -5,20 +5,11 @@
...
-    constructor(stub: RsAliasStub, elementType: IStubElementType<*, *>) : super(stub, elementType)
-
-    override fun getNameIdentifier(): PsiElement? {
-        // `use Foo as _;` really should be unnamed, but "_" is not a valid name in rust, so I think it's ok
-        return identifier ?: underscore
-    }
-}
+val RsAlias.nameLikeText: String
+    get() = greenStub?.nameLikeText
+        ?: node.findChildByType(RsElementTypes.IDENTIFIER)?.text
```
Undin, line 13: Why not just `identifier` instead of `node.findChildByType(RsElementTypes.IDENTIFIER)`?

src/main/kotlin/org/rust/lang/core/psi/ext/RsAlias.kt
```diff
@@ -5,20 +5,11 @@
...
+        ?: node.findChildByType(RsElementTypes.UNDERSCORE)?.text
```
Undin, line 14: `underscore?.text`?

src/main/kotlin/org/rust/lang/core/psi/ext/RsUseSpeck.kt
```diff
@@ -22,7 +22,7 @@ val RsUseSpeck.pathOrQualifier: RsPath? get() = path ?: qualifier
 
 val RsUseSpeck.nameInScope: String? get() {
     if (useGroup != null) return null
-    alias?.name?.let { return it }
+    alias?.nameLikeText?.let { return it }
```
Undin, line 25: Should we provide any value here in case of `_` as alias name?

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -3021,4 +3021,9 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
         const _: i32 = 1;
         const _: i32 = 1;
     """)
+
+    fun `test no E0252 multiple underscore aliases`() = checkErrors("""
```
Undin, line 3025: Unfortunately, this test does nothing because all items in imports are from stdlib and cannot be resolved because you don't add stdlib dependency here
Try to add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` and it will fail.

Also, let's avoid using stdlib here not to slow down tests. You can declare inline module with some items and use them in imports

src/main/kotlin/org/rust/lang/core/psi/ext/RsUseSpeck.kt
Undin: Looks like exactly this thing is the cause of false positive `E0252` error

src/main/kotlin/org/rust/lang/core/psi/ext/RsExternCrateItem.kt
```diff
@@ -17,7 +17,7 @@ import org.rust.lang.core.resolve.ref.RsExternCrateReferenceImpl
 import org.rust.lang.core.resolve.ref.RsReference
 import org.rust.lang.core.stubs.RsExternCrateItemStub
 
-val RsExternCrateItem.nameWithAlias: String get() = alias?.name ?: referenceName
+val RsExternCrateItem.nameWithAlias: String get() = alias?.nameLikeText ?: referenceName
```
Undin, line 20: Looks like this change introduces the following bug:
```rust
extern crate objc as _;

fn foo(cl: Class/*caret*/) {} // I expect error annotation here and `Import` quick fix
                              // but plugin does nothing here
```

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -698,25 +698,30 @@ class RsNamedFieldDeclStub(
 
 class RsAliasStub(
     parent: StubElement<*>?, elementType: IStubElementType<*, *>,
-    override val name: String?
+    override val name: String?,
+    val nameLikeText: String
```
Undin, line 702: After all comments above, it seems we don't need `nameLikeText` in stub at all

src/main/kotlin/org/rust/lang/core/psi/ext/RsAlias.kt
Undin: Btw it doesn't support raw identifiers. You should use `unescapedText`

src/main/kotlin/org/rust/lang/core/psi/ext/RsExternCrateItem.kt
Undin: Note, there is the following case for name resolution (it's valid Rust code):

```rust
extern crate objc as _;
extern crate objc;

use objc::runtime::Class;

fn foo(cl: Class) {}

fn main() {}
```

Undin:
@ortem ping

WaffleLapkin:
@ortem ping. do you still plan to work on this?

ortem:
@WaffleLapkin Yes, thank you for reminding me, I'm going to finish this PR in the coming days

ortem:
@Undin I reimplemented this fix from scratch using the simple approach proposed by @dima74. Probably it's not worth introducing `RsAlias.nameLikeElement` after all. What do you think?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1319,6 +1319,7 @@ private fun AnnotationSession.duplicatesByNamespace(
             + importedNames)
             .filter { it !is RsExternCrateItem } // extern crates can have aliases.
             .filter { it.nameOrImportedName() != null }
+            .filter { it.nameOrImportedName() != "_" }
```
dima74, line 1322: Maybe unite this and previous filter (`it.nameOrImportedName() != null`)

Undin: Agree with @dima74

Undin (approved):
Looks ok to fix #5067 but probably we still need some API to handle `_` properly in other places and avoid bugs like #6850 but let's do it in a separate PR

ortem:
@Undin Wow, that's a really annoying bug. I faced a problem with code completion too: https://github.com/intellij-rust/intellij-rust/issues/6851

I agree with you. Last week, we discussed in person with @dima74 that underscore aliases should be treated in a special way. I was considering changing `org.rust.lang.core.psi.ext.RsUseSpeckKt#getNameInScope`, but that approach was too rough. The proper approach should be accurate enough in order not to break the existing logic (e.g. `RsCachedItems`).

AFAIU currently our name resolution distinguishes two kinds of imports: _named imports_ and _star imports_ (`RsCachedItems`), and process them differently. Probably, there might be three kinds including _underscore imports_, but I'm not pretty sure how to process them properly

PR #4769, opened 2019-12-19
RES: fix pattern namespaces

vlad20012:
E.g. int this case `Foo` was incorrectly multi-resolved:
```rust
struct Foo { f: i32 }
     //X
fn Foo() {}
fn main() {
    let Foo { f } = Foo { f: 1 };
}     //^
```

bors r+


Undin:
bors retry

PR #4772, opened 2019-12-20
INT: select unimplemented after Implement members fix

### LINKED ISSUE #4750: Select "unimplemented" after "Implement members" invocation
## Environment

* **IntelliJ Rust plugin version:** 0.2.113.2810-193-nightly
* **Rust toolchain version:** 1.39.0 (4560ea788 2019-11-04) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3.1 (CL-193.5233.144)
* **Operating system:** macOS 10.15

## Problem description

I found it quite annoying to manually select `unimplemented!()` after calling "Implement members" to be able to replace it with an actual code. Consider how it is done in PyCharm, here I invoked "Implement abstract methods" on `Bar` class. `pass` is selected so I can start typing the method code right away.

Screen Shot 2019-12-14 at 14 42 34

## Steps to reproduce

Consider

```rust
struct WesternUnion {
    name: String,
    verification: u8,
}

impl CreditCharge for WesternUnion {
}

trait CreditCharge {
    fn charge_with_id(&self, id: u32) -> bool;
}
```

Invoke "Implement members" on `impl CreditCharge for WesternUnion`

Screen Shot 2019-12-14 at 16 54 20

expected

Screen Shot 2019-12-14 at 16 54 27

Kobzol:
Fixes #4750

This PR adds support for selecting the content of implemented trait functions so that it can be quickly modified. I adapted it from the source code of PyCharm.

I have two questions:
1. ~~I'm pretty sure that IntelliJ can create a "selection chain" (I don't know the correct terminology), where it selects a first item, after you press Enter, it selects another one and so on, until the user goes through all of them and/or presses Escape. Which API can I use for this? The current implementation only selects the last implemented function, which is far from ideal. Also now if I press Enter after the selection, the content is replaced by a newline, in the "chain" mode it should be just deselected.~~ EDIT: I found the API (`buildAndRunTemplate`) and used it.
2. How to test this, or at least where are similar tests so that I can be inspired how to test it :)

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
@@ -8,6 +8,7 @@ package org.rust.ide.refactoring.implementMembers
 import com.intellij.codeInsight.hint.HintManager
 import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.ScrollType
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
 import org.rust.ide.inspections.import.RsImportHelper.importTypeReferencesFromElements
@@ -32,7 +33,7 @@ fun generateTraitMembers(impl: RsImplItem, editor: Editor?) {
     if (chosen.isEmpty()) return
     runWriteAction {
         // Non-null was checked by `findMembersToImplement`.
-        insertNewTraitMembers(chosen, impl.members!!, trait)
+        insertNewTraitMembers(chosen, impl.members!!, trait, editor)
     }
 }
 
@@ -48,7 +49,8 @@ private fun findMembersToImplement(impl: RsImplItem): Pair<TraitImplementationIn
 private fun insertNewTraitMembers(
     selected: Collection<RsAbstractable>,
     existingMembers: RsMembers,
-    trait: BoundElement<RsTraitItem>
+    trait: BoundElement<RsTraitItem>,
+    editor: Editor?
 ) {
     checkWriteAccessAllowed()
     if (selected.isEmpty()) return
@@ -67,6 +69,7 @@ private fun insertNewTraitMembers(
     }.toMutableList()
     val existingMembersOrder = existingMembersWithPosInTrait.map { it.second }
     val areExistingMembersInTheRightOrder = existingMembersOrder == existingMembersOrder.sorted()
+    var needToSelect: RsElement? = null
 
     for ((index, newMember) in newMembers.withIndex()) {
         val posInTrait = traitMembers.indexOfFirst {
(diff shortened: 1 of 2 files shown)


Undin:
I think it can be annoying so I'd prefer the solution that @mchernyavsky described above

Kobzol:
Good catch, I didn't think of that. However, if I don't use the template builder and just select the text `unimplemented!` with `editor.selectionModel.setSelection(start, statement.textRange.endOffset)`, when the user presses Enter, it will remove the `unimplemented!` line, which is incorrect. Should I use the template builder, but only for the first method or is there a simple way of making the previous line stay there after the user presses Enter even without using the template builder?

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
```diff
@@ -104,9 +105,15 @@ private fun insertNewTraitMembers(
...
         }
+
+        if (addedMember is RsFunction && templateMembers.isEmpty())
+        {
+            templateMembers.add(addedMember)
+        }
     }
 
     importTypeReferencesFromElements(existingMembers, selected, trait.subst)
+    editor?.buildAndRunTemplate(existingMembers, templateMembers.mapNotNull { it.block?.expr?.createSmartPointer() })
```
Undin: Running template builder here looks redundant.
At least, I don't expect it after using support for other languages (kotlin, java) where only body of the first method is selected (without template)
So let's make the same behavior as implemented in other languages

Undin: Also, it can be easily tested

Kobzol: Can you point me to an API that I can use to select the body of the first method? In the first commit of this PR I simply selected the `unimplemented!` body of the first method. However, if I simply select it and the user presses Enter, it will delete the `unimplemented!` text, instead of leaving it in place, which is wrong.

Undin: @Kobzol why do you think that it's wrong?
It's exactly the same behavior that as all other languages provide

Kobzol: Ok.. it seemed a bit unintuitive to me, since with the template builder Enter keeps the original text there :) In that case I'll just revert to the first commit and add tests.

Kobzol:
@Undin Only now I realized that the Implement members fix can also implement Types and Constants. They are not highlighted after being implemented, which would be nice.

So what about if I used the template builder, but for everything together (types + constants + first method)? Types and constants are single-line, so they should be a good fit for the template builder. It could possibly be ordered in such a way that the method would always be last in the template builder.

Undin:
> Types and constants are single-line

Constant can be multiline string

> It could possibly be ordered in such a way that the method would always be last in the template builder.

In case of non-usual order in trait, it will navigate user caret quite weird

In general, template builder is more about something short (like variable name) and ordered (like in live templates). "Override members" can produce a lot of code and a user may want to implement them (or not at all) in arbitrary order.
So I don't think it's a good idea to use template builder here at all

Undin:
Maybe it's worth to select the content of the first **item** instead of method?

Kobzol:
That sounds like a reasonable compromise :-) I've implemented it and added tests.

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
```diff
@@ -104,9 +107,32 @@ private fun insertNewTraitMembers(
             val whitespaces = createExtraWhitespacesAroundFunction(addedMember, next)
             existingMembers.addAfter(whitespaces, addedMember)
         }
+
+        if (selectedElement == null)
+        {
```
Undin: Could you, please, format curly braces (in all changed code) according to [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting)?

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
```diff
@@ -67,6 +69,7 @@ private fun insertNewTraitMembers(
     }.toMutableList()
     val existingMembersOrder = existingMembersWithPosInTrait.map { it.second }
     val areExistingMembersInTheRightOrder = existingMembersOrder == existingMembersOrder.sorted()
+    var selectedElement: RsElement? = null
```
Undin: Maybe `needToSelect` is better?

Kobzol:
Ok, I edited the style and squashed the changes.

PR #4774, opened 2019-12-22
Evcxr REPL integration improvements

dima74:
This is continuation of #4615

* Add actions: stop execution, toggle soft wrap, print console content
* Replace command history with standard ConsoleHistoryController (which additionally provides action to see all executed commands)
* Fix that commands with errors were added to context for completion (based on https://github.com/google/evcxr/pull/82)
* Initial support for multiline commands (Shift+Enter)
* Add variables view

PR #4784, opened 2019-12-25
PSI: Rename code fragments consistently

ortem:
There were two classes named `RsReplCodeFragment`, and that's why plugin build fails sometimes. So now the generated classes will be `*CodeFragmentElement.java`.

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -87,7 +87,7 @@
     elementType(".+BinExpr") = BinaryExpr
     elementType(".+BinOp") = BinaryOp
 
-    extraRoot(".*CodeFragment")=true
+    extraRoot(".*CodeFragmentElement")=true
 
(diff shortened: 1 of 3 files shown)


Undin:
@ortem let's come up with another way how to solve this problem and avoid "Fake" suffix

ortem:
@Undin OK, another option is moving this class to another package. It was made by @dima74 [here](https://github.com/intellij-rust/intellij-rust/pull/4774/files#diff-2ddd161008549720180441626203f1db). What do you think?

ortem:
We've discussed in person that renaming is more general and, therefore, preferable to moving. But `*CodeFragmentElement.java` name is better than `*CodeFragmentFake.java`.

ortem:
@Undin Done. Also, note that
* `TYPE_PATH_CODE_FRAGMENT` renamed to `TYPE_PATH` and `VALUE_PATH_CODE_FRAGMENT` renamed to `VALUE_PATH` just like other `RsCodeFragmentElementType` companion entities
* Code fragments declarations order is now the same in `RustParser.bnf`, `RsCodeFragment.kt`, and `RsCodeFragmentElementType.kt`

PR #4787, opened 2019-12-26
Get rid of deprecated `InjectedFileViewProvider` usage

ortem:


src/main/kotlin/org/rust/lang/core/parser/RustParserDefinition.kt
```diff
@@ -32,26 +31,25 @@ class RustParserDefinition : ParserDefinition {
     override fun createFile(viewProvider: FileViewProvider): PsiFile {
         val default = { RsFile(viewProvider) }
 
-        if (viewProvider is InjectedFileViewProvider) {
-            val project = viewProvider.manager.project
-            val host = InjectedLanguageManager.getInstance(project).getInjectionHost(viewProvider)
+        val project = viewProvider.manager.project
+        val injectionHost = InjectedLanguageManager.getInstance(project).getInjectionHost(viewProvider)
```
Undin: Looks like `InjectedLanguageManager` is not registered in all parsing tests. 
So I think you have to register it yourself in `setUp`, i.e. do something like 
`project.registerService(InjectedLanguageManager::class.java, CoreInjectedLanguageManager::class.java)`

ortem: Thank you, fixed.

PR #4788, opened 2019-12-26
Fix spelling and extend project dictionary

ortem:


Undin:
@ortem you should add `internal` label only if you want to add the corresponding change into release changelog.
See https://github.com/intellij-rust/intellij-rust/blob/master/MAINTAINING.md#accepting-pull-requests

PR #4792, opened 2019-12-27
COMP: not add parentheses when completing a function passed as a param

### LINKED ISSUE #4450: Do not add parenthesis when completing methods in closures
## Environment

* **IntelliJ Rust plugin version:** 0.2.106.2134-193
* **Rust toolchain version:** 1.37.0 (eae3437df 2019-08-13) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.3519.24)
* **Operating system:** macOS 10.14.6

## Problem description

Rust allows omitting closure parameters and use functions in closures directly:

```rust
fn multiply_by_two(input: i32) -> i32 {
    input * 2
}

fn main() {
    let _ = Some(5).map(multiply_by_two);
}
```

When I try co compete such method, I get extra parenthesis added which breaks compilation.

```rust
fn multiply_by_two(input: i32) -> i32 {
    input * 2
}

fn main() {
    let _ = Some(5).map(mul/*caret*/);
}
```

Then I press tab and expect to get the example above.
Instead I get: 

```rust
fn multiply_by_two(input: i32) -> i32 {
    input * 2
}

fn main() {
    let _ = Some(5).map(multiply_by_two(/*caret*/));
}
```

pt2121:
Fix #4450 

I'm new to the code base and not sure if this is the best approach.
If there is a better way to fix this, please let me know. Thanks!

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -244,9 +244,18 @@ open class RsDefaultInsertHandler : InsertHandler<LookupElement> {
                 if (curUseItem != null) {
                     appendSemicolon(context, curUseItem)
                 } else {
-                    val isMethodCall = context.getElementOfType<RsMethodOrField>() != null
+                    val methodOrField = context.getElementOfType<RsMethodOrField>()
+                    val isMethodCall = methodOrField != null
                     if (!context.alreadyHasCallParens) {
-                        document.insertString(context.selectionEndOffset, "()")
+                        val rsMethodCall = context.getElementOfType<RsMethodCall>()
+                        if (rsMethodCall == null) {
+                            document.insertString(context.selectionEndOffset, "()")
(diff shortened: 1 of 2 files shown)


vlad20012:
Hello! Thanks for your contribution.

I think passing a function pointer as argument is more rare case than just calling other function in argument list of other function call:
```rust
fn foo() -> i32 {0}
fn bar(i: i32) {}

fn main() {
    bar(foo());
}
```
Your solution breaks `()` completion in this case.

I think this can be solved by looking to expression expected type (`RsCompletionContext.expectedTy`). But it can be tricky with type parameter types with Fn* bounds

pt2121:
Thanks for the pointer. I'll look into it.

pt2121:
I'm not sure how to reference `RsCompletionContext` in `handleInsert` and can't find a way to get a type of an enclosing function.
fixed the `bar(foo());` issue tho it's not as elegant as type check.

PR #4799, opened 2019-12-28
TYPE: improve extend selection for Rust and Toml string literals

Undin:
Improve `Extend Selection` action for Rust and Toml literals:
* select whole string literal without quotes
* select escapes symbols

| Before | After |
|-|-|
|  |  |
|  |  |


Undin:
@vlad20012 thanks for `com.intellij.codeInsight.editorActions.SelectWordUtil#addWordHonoringEscapeSequences`

PR #4767, opened 2019-12-19
TY: fix deref adjustments

vlad20012:
This shouldn't affect our users anyhow. Will be used by next PRs

PR #4804, opened 2019-12-29
Revert "fix deref adjustments"

Undin:
Revert changes made in #4767.
They lead to new false positive `Cannot assign to indexed content of immutable binding` error annotation in the following case:
```rust
fn swap<T>(index_a: usize, index_b: usize, arr: &mut [T]) where T: Copy {
    let temp = arr[index_a];
    arr[index_a] = arr[index_b];
    arr[index_b] = temp;
}
```
image

I've added this code as test


Undin:
@adamiao thank you for code example. Although, looks like you've faced with another problem

Undin:
@vlad20012 I think it's interesting for you

PR #4812, opened 2020-01-01
COMP: guess proper type of braces while macro call completion

Undin:
Analyze documentation of macro definition and extract the most commonly used type of braces from it.
For example, it allows the plugin to insert `{}` while `thread_local` completion instead of `()`

It was originally implemented in https://github.com/rust-analyzer/rust-analyzer/pull/2039

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsMacro.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsMacro.kt
@@ -16,12 +16,11 @@ import com.intellij.psi.util.CachedValuesManager
 import org.rust.ide.icons.RsIcons
 import org.rust.lang.core.macros.MacroGraph
 import org.rust.lang.core.macros.MacroGraphBuilder
-import org.rust.lang.core.psi.RsElementTypes
-import org.rust.lang.core.psi.RsMacro
-import org.rust.lang.core.psi.RsMacroBody
-import org.rust.lang.core.psi.RsPsiFactory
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsMacro.kt
```diff
@@ -73,6 +72,26 @@ val RsMacro.macroBodyStubbed: RsMacroBody?
...
+    get() = stub?.preferredBraces ?: guessPreferredBraces()
+
+/**
+ * Analyses documentation of macro definition to determine what kind of brackets usually used
+ */
+private fun RsMacro.guessPreferredBraces(): MacroBraces {
+    val documentation = documentation()
+    if (documentation.isNullOrEmpty()) return MacroBraces.PARENS
+
+    val macroCallRegex = """(^|[^\p{Alnum}_])$name\s*!\s*(?<brace>[({\[])""".toRegex()
```
vlad20012: Let's move it to static b/c regex compilation is heavy stuff

Undin: I've extracted it into static field
Note, now the regex matches **all** macro calls because we have only one regex (previously, we substituted macro name into it). All irrelevant matches are filtered manually

Undin:
@vlad20012 thanks for the review!
I've found typo into counting of brace types.
Also, provided support of calls with raw identifiers

Issue #3839, opened 2019-05-13
Hide Cargo tool window button for non-Rust/non-Cargo projects


amakeev:
Currently, 'Cargo' tool window is always shown, that is not optimal for the user experience
Example for a Gradle Java project 
<img width="218" alt="image" src="https://user-images.githubusercontent.com/238583/57651969-e065ee00-75ce-11e9-9f3b-f6a6de56b0fa.png">

PR #4813, opened 2020-01-02
Adds a conditionClass to the Cargo tool window

samdfonseca:
- Only displays tool window if a Cargo.toml file is found in any of the
content roots

I've found myself repeatedly removing the Cargo tool window from the sidebar, since I mostly work on non-Rust projects in IntelliJ. This change makes displaying the tool window conditional on finding a `Cargo.toml` file in any of the content roots.

My only real concern relates to `CargoToolWindowFactory` implementing `DumbAware` and how inspecting the content roots behaves during indexing, however I tested it and it appears to work as expected.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -87,6 +87,7 @@ pavel-v-chernykh
 pcholakov
 pocket7878
 rrevenantt
+samdfonseca
 sanmai-NL
 sgift
 sgrif
--- a/src/main/kotlin/org/rust/cargo/project/toolwindow/CargoToolWindow.kt
+++ b/src/main/kotlin/org/rust/cargo/project/toolwindow/CargoToolWindow.kt
@@ -9,9 +9,12 @@ import com.intellij.ide.DefaultTreeExpander
 import com.intellij.ide.TreeExpander
 import com.intellij.openapi.actionSystem.*
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.DumbAware
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.roots.ModuleRootManager
 import com.intellij.openapi.ui.SimpleToolWindowPanel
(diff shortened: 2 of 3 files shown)


samdfonseca:
Apologies if opening this without any communication beforehand is not following the correct steps. I didn't see anything in the `CONTRIBUTING.md` file about that and it was simple enough to put together so I figured it couldn't hurt.

I'm not sure if the `if Cargo.toml exists then display Cargo tool window` condition logic is 100% correct or if it's the UX you're looking for, so I'd be curious to hear thoughts on that. I considered adding a `Hide Cargo tool window if ...` flag to the application level settings, but I'd rather get some feedback before doing that since it's a more involved change. Unconditionally displayed tool windows seems to be a problem with a lot of plugins (Elm, Typescript, Flutter, ReasonML do it as well) so it'd be nice to have some sort of control over when it appears.

samdfonseca:
Ah, just now noticing https://github.com/intellij-rust/intellij-rust/pull/4102, but don't see any explanation for why it was closed. Looks like the issue https://github.com/intellij-rust/intellij-rust/issues/3839 is still open though. @mchernyavsky was there a reason to close the PR without merging?

Undin:
@samdfonseca Hi!
Sorry for long response
> Ah, just now noticing #4102, but don't see any explanation for why it was closed. Looks like the issue #3839 is still open though. @mchernyavsky was there a reason to close the PR without merging?

We closed that issue because that change it makes attaching new cargo project more difficult due to Cargo tool window provides `Attach` action.
But I've improved `Attach` action in #4836. After the PR will be merged it would be possible to attach new cargo project from Project View and from editor notification of `Cargo.toml` files. So I think we can provide condition for Cargo tool window

CC @mchernyavsky

mchernyavsky:
@samdfonseca Hello! I'm closing this in favor of #4102 because #4102 has an additional small optimization (`hasAtLeastOneValidProject` check) and reuses code from `discoverAndRefresh` (see `suggestManifests`). But anyway, thanks for the PR! Without this, we probably wouldn't have remembered the issue.

Issue #3453, opened 2019-02-28
When will intellij idea support rust debugger ?


99yun:
When will intellij idea support rust debugger ?

Or never ? !

amakeev:
There is no plan to support Rust debugging in IntelliJ IDEA Community; Though, eventually the debugging support might come to IntelliJ IDEA Ultimate  (no ETA on that).

Currently, CLion provides the most complete Rust support, including debugging; we also plan to add features such as Rust profiling support to CLion in future.

Allenyn:
Actually, what I just want to know, is whether will debugging support come to IntelliJ IDEA Ultimate ? 

I am done, some one can close it now.

Undin:
Update:
There is a plan to integrate with [Native Debug](https://plugins.jetbrains.com/plugin/12775-native-debugging-support/) plugin to support Rust debugging in IntelliJ Ultimate

Undin:
What we need to do:
* [ ] use [nativeDebug](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin as a dependency of `debugger` module instead of CLion
* [ ] provide a way to download LLDB with the correspnding UI
* [ ] implement `LLDBDriverConfiguration` for non-CLion environment

PR #4823, opened 2020-01-09
Remove CLion dependencies from `debugger` module

Undin:
Now, `debugger` module contains code related to native debugger only.
Previously, it used some CLion only code. Now all such code moved into `clion` module.
To make sure that `debugger` module doesn't depend on CLion, it is compiled with [nativeDebug](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin dependency instead of CLion when base IDE is IDEA.

Also, duplicate code from `DebugRunner` and `DebugRunnerLegacy` was moved into a single place

Part of #3453


Undin:
@mchernyavsky I didn't change depenency condition of `debugger-only.xml` (i.e. it still uses `com.intellij.modules.clion` module) because it provides some debugger UI that's useless without the corresponding runners

Undin:
@lancelote I've added it into test project just to check that everything works as before, i.e. debugger works in CLion and IDEA Ultimate doesn't have Rust related settings about debugger even with native debug plugin

PR #4829, opened 2020-01-09
Fix compatibility with `Grazie` plugin

ortem:
Explicitly overridden implementations are necessary to support the next version of `Grazie` plugin, because of new jvm-default parameters used in `Grazie`.

Сс @TanVD


Undin:
@TanVD is change with `@JvmDefault` annotation supposed to be included to plugin version compatible with 2019.2/2019.3? Or it will be only in 2020.1 `Grazie` plugin?

TanVD:
We've found the way to accomplish our task without binary incompatiable changes. So PR can be closed without merge :) Thank you!

Undin:
@TanVD Thanks!

PR #4839, opened 2020-01-15
INSP: Improve RsUnresolvedReferenceInspection

Exidex:
#### Before PR (`ignoreWithoutQuickFix` set to`true`) - default 
Screenshot from 2020-01-15 20-23-26

#### Before PR (`ignoreWithoutQuickFix` set to `false`)
Screenshot from 2020-01-15 20-23-54

#### After PR (`ignoreWithoutQuickFix` set to`true`) 
Screenshot from 2020-01-22 23-21-52

#### After PR  (`ignoreWithoutQuickFix` set to `false`)
Screenshot from 2020-01-15 20-26-01

This PR is tested with experimental macro expansion engine enabled,
~~so this PR would probably make more sense to merge after making new engine the default~~ with `ignoreWithoutQuickFix` it could be merged now

~~I removed `ignoreWithoutQuickFix` checkbox, it is my main issue with current implementation,
as it defeats one of the main purposes of the inspection which is to show references that cannot be resolved, for reasons more than just 'not imported'~~

Left `ignoreWithoutQuickFix` as is until experimental macro expansion engine will be enabled by default

Also previously this inspection could register problem for same PsiElement multiple times, this is fixed in PR

~~Due to `ignoreWithoutQuickFix` being removed several tests that assumed this variable was set
to true do not really make sense now, some were corrected.~~

This PR is WIP because it broke some tests in `AutoImportFixTest` and `AutoImportFixStdTest` and I'd like some feedback on the PR before fixing

There are also some failed tests in `RsBuildActionTest` and `CargoBuildManagerTest` not sure whether this is my fault

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsUnresolvedReferenceInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsUnresolvedReferenceInspection.kt
@@ -20,6 +20,7 @@ import org.rust.lang.core.psi.RsVisitor
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.endOffsetInParent
 import org.rust.lang.core.psi.ext.getPrevNonCommentSibling
+import org.rust.lang.core.psi.ext.qualifier
 import javax.swing.JComponent
 
 class RsUnresolvedReferenceInspection : RsLocalInspectionTool() {
@@ -31,28 +32,46 @@ class RsUnresolvedReferenceInspection : RsLocalInspectionTool() {
     override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor =
         object : RsVisitor() {
             override fun visitPath(path: RsPath) {
-                if (path.parent is RsPath) return
-                val (basePath, candidates) = AutoImportFix.findApplicableContext(holder.project, path) ?: return
+                // here we check if this path is resolved
+                // if it is we return
+                // if not than check whether qualifier of this path is resolved
+                // if it is we register problem
+                // if not then we assume that unresolved parent paths of this path
+                // were or will be handled
+                // because all paths will eventually be processed by visitor
+                val (problemPath, candidates) = AutoImportFix.findApplicableContext(holder.project, path)
+                    ?: return
+
                 if (candidates.isEmpty() && ignoreWithoutQuickFix) return
 
-                val referenceName = basePath.identifier?.text
-                // Don't highlight generic parameters
+                path.qualifier
+                    ?.let { AutoImportFix.findApplicableContext(holder.project, it) }
+                    ?.let { return }
+
+                val referenceName = problemPath.identifier?.text
                 val range = TextRange(
-                    0,
-                    path.typeArgumentList?.getPrevNonCommentSibling()?.endOffsetInParent ?: path.textLength
+                    // Don't highlight parent identifiers
+                    problemPath.identifier?.startOffsetInParent ?: 0,
+                    // Don't highlight generic parameters
+                    problemPath.typeArgumentList?.getPrevNonCommentSibling()?.endOffsetInParent
+                        ?: problemPath.textLength
                 )
-                holder.registerProblem(path, candidates, referenceName, range)
+
+                holder.registerProblem(problemPath, candidates, referenceName, range)
             }
 
             override fun visitMethodCall(methodCall: RsMethodCall) {
-                val (_, candidates) = AutoImportFix.findApplicableContext(holder.project, methodCall) ?: return
+                val (_, candidates) = AutoImportFix.findApplicableContext(holder.project, methodCall)
+                    ?: return
 
                 if (candidates.isEmpty() && ignoreWithoutQuickFix) return
 
-                val identifier = methodCall.identifier
-                val referenceName = identifier.text
-                val range = TextRange(0, identifier.textLength)
-                holder.registerProblem(methodCall, candidates, referenceName, range)
+                val range = TextRange(
+                    0,
+                    methodCall.identifier.textLength
+                )
+
+                holder.registerProblem(methodCall, candidates, methodCall.identifier.text, range)
             }
         }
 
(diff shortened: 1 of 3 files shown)


Exidex:
You are right. It makes more sense leaving `ignoreWithoutQuickFix` as is until experimental macro expansion engine will become default. I updated PR and description

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -106,23 +104,27 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
...
             }
 
-            val candidates = getImportCandidates(ImportContext.from(project, path, false), basePath.referenceName, path.text) {
-                path != basePath || !(it.item is RsMod || it.item is RsModDeclItem || it.item.parent is RsMembers)
+            val candidates = getImportCandidates(
+                ImportContext.from(project, path, false),
+                path.referenceName,
+                path.text
+            ) {
+                true
```
Exidex: Not super sure about this one. From what I gathered it returns correct candidates but don't know if this is proper way to do it

src/main/kotlin/org/rust/ide/inspections/RsUnresolvedReferenceInspection.kt
```diff
@@ -31,28 +32,46 @@ class RsUnresolvedReferenceInspection : RsLocalInspectionTool() {
...
-                if (path.parent is RsPath) return
-                val (basePath, candidates) = AutoImportFix.findApplicableContext(holder.project, path) ?: return
+                // here we check if this path is resolved
+                // if it is we return
+                // if not than check whether qualifier of this path is resolved
+                // if it is we register problem
+                // if not then we assume that unresolved parent paths of this path
+                // were or will be handled
+                // because all paths will eventually be processed by visitor
+                val (problemPath, candidates) = AutoImportFix.findApplicableContext(holder.project, path)
```
Undin: Hm, it doesn't make sense to look for auto import candidates for each path because in path `foo::bar::baz` we can import only `foo` part, i.e. only base path. If `bar` cannot be resolved and `foo` can, we definitely should highlight `bar` as unresolved but without auto-import quick fix.
I think we should use `AutoImportFix.findApplicableContext` only when `path == path.baseBath`.

It seems we should do the following things:
* check that reference cannot be resolved
* check that its child path can be resolved (it's cheap because resolve result is already in the cache). If no, just do nothing. Otherwise, highlight it as unresolved
* if path is the same as base bath, try to find import candidates and do all related stuff as before

Probably, some code from `AutoImportFix.findApplicableContext` should be extraced to `RsUnresolvedReferenceInspection` itself.

src/main/kotlin/org/rust/ide/inspections/RsUnresolvedReferenceInspection.kt
```diff
@@ -31,28 +32,46 @@ class RsUnresolvedReferenceInspection : RsLocalInspectionTool() {
...
+
+                val referenceName = problemPath.identifier?.text
                 val range = TextRange(
-                    0,
-                    path.typeArgumentList?.getPrevNonCommentSibling()?.endOffsetInParent ?: path.textLength
+                    // Don't highlight parent identifiers
+                    problemPath.identifier?.startOffsetInParent ?: 0,
+                    // Don't highlight generic parameters
+                    problemPath.typeArgumentList?.getPrevNonCommentSibling()?.endOffsetInParent
+                        ?: problemPath.textLength
```
Undin: Let's extract start and end offsets of range into local variables to improve readability and avoid weird formatting

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
Undin: I think you don't need to change anything here

src/main/kotlin/org/rust/ide/inspections/RsUnresolvedReferenceInspection.kt
Exidex: So, if I correctly understand, you ask me to separate checking if path is resolved and actually getting import candidates. I don't fully know intellij api and code is spread out so it's a little hard for me to fully wrap my head around this. One thing preventing me from easily splitting resolve and lookup is that lookup is done in 2 different places:
https://github.com/intellij-rust/intellij-rust/blob/9ccbb03af96b5804126b4ff555280635afbfc503/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt#L103
and
https://github.com/intellij-rust/intellij-rust/blob/9ccbb03af96b5804126b4ff555280635afbfc503/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt#L121
Any pointers here?

Undin: > So, if I correctly understand, you ask me to separate checking if path is resolved and actually getting import candidates.

Not quite. You don't need to invoke `AutoImportFix.getApplicableContext()` where it doesn't make sense.

You need to check that path cannot be resolve to single item (`path.reference.resolve() == null`)
In this case, you need to check if this path has qualifier (`path.qualifier`):
* if `path.qualifier == null` (i.e. `path == path.basePath()`) you need to check if `AutoImportFix` can be applied to it (i.e. make the same actions that currently in master)
* if `path.qualifier != null`, check if it can be resolved to single item. And only of it can be resolved, you need to add error annotation about unresolved reference for `path`. Also, you don't need to add `AutoImportFix` here because it doesn't make sense

Looks like all checks described above don't require changes in `AutoImportFix`

Exidex: Wouldn't `path.reference.resolve() == null` cause false positives/negatives? Like primitives for example would resolve to null. `AutoImportFix.findApplicableContext` already handles this and couple other cases. That is why I was asking separating that method.

Undin: Yeah, I've forgotten about primitive types. But looks like it's only one additional thing that you need to check

Exidex: There is a comment that says that some traits that would be resolved by plugin that wouldn't be resolved by the rustc. Which, if not handled, would cause false negatives

https://github.com/intellij-rust/intellij-rust/blob/9ccbb03af96b5804126b4ff555280635afbfc503/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt#L99

Exidex:
Cool, thanks for finishing the work on it.

Undin:
@Exidex Thanks for the contribution!
Found out that the desired improvements are more complicated than I had expected.
Looks like I've implemented all proposed improvements from this PR.
If you find something that I've missed - fill free to ping me about it.

PR #4847, opened 2020-01-20
ANN&REF: add annotations to leafs when highlighting leafs

vlad20012:
Now, when we want e.g. to highlight a function name:
```rust
fn foo() {}
// ~~~ highlight this
```
we do it when Annotator.annotate is called with leaf (identifier) element argument, not the function element.

This is more proper usage of annotators API. Particularly, it fixes highlights blinking (most viewable on a dark theme).

Also:
1. fixed incorrectly highlighted "primitive" in this case:
  ```rust
  let mut i32 = 1;
  i32 = 2;
  ```
2. fixed literal highlighting in attributes: `#[foo("bar")]`

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
@@ -9,49 +9,102 @@ import com.intellij.ide.annotator.AnnotatorBase
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.openapi.util.Key
-import com.intellij.openapi.util.TextRange
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
```diff
@@ -9,18 +9,31 @@ import com.intellij.ide.annotator.AnnotatorBase
...
+import org.rust.openapiext.getOrPut
+
+`import com.intellij.psi.impl.source.tree.LeafPsiElement
+import com.intellij.psi.search.PsiTodoSearchHelper
+import org.rust.ide.colors.RsColor
+import org.rust.ide.highlight.RsHighlighter
+import org.rust.ide.todo.isTodoPattern
+import org.rust.ide.utils.isEnabledByCfg
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsElementTypes.*
```
Undin: Clean up imports to make it compile

Issue #4208, opened 2019-08-01
Jump from test results to doctest leads to unexpected position


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.103.2698-192-nightly
* **Rust toolchain version:** rustc 1.36.0 (a53f9df32 2019-07-03)
* **IDE name and version:** CLion 2019.3 Nightly 193.1846
* **Operating system:** macOS 10.14.4

## Problem description

I am running some doctests and trying to jump from the test results back to doctest declaration, unfortunately the resulting goto line is not quite correct.

## Steps to reproduce

- Consider the following code from a library package

```rust
/// Reversed Words
/// https://www.codewars.com/kata/reversed-words
///
/// Complete the solution so that it reverses all of the words within the string
/// passed in.
///
/// # Examples
///
/// ```
/// # use rust_codewars::kyu8::reversed_words;
///
/// let words = "The greatest victory is that which requires no battle";
/// let result = reversed_words::reverse_words(words);
///
/// assert_eq!(result, "battle no requires which that is victory greatest The");
/// ```
///
pub fn reverse_words(str: &str) -> String {
    str.split_whitespace()
        .rev()
        .collect::<Vec<&str>>()
        .join(" ")
}
```

- run `cargo test` run configuration
- try to jump from the test result tree to a doctest
- line 17 is focused, while I expect 8

<img width="806" alt="tests" src="https://user-images.githubusercontent.com/2500320/62275384-7d0b8f80-b44a-11e9-9aaa-09568e7148de.png">

Undin:
Combination of the current psi for doc comments (single psi node) and API for navigation from test result tree to the corresponding element makes desired navigation impossible.
But looks like after creation full ast tree according to markdown grammar rules it will be possible

CC @mchernyavsky

mchernyavsky:
@Undin What can we do in this situation?

<img width="300" alt="screen" src="https://user-images.githubusercontent.com/6079006/62333545-8f2e1200-b4cc-11e9-8ec6-55af7afa905a.png">

Undin:
I don't see a lot of variants in such case. I suppose we will navigate to wrong line.
BTW we already have a similar situation when you change name or location of test function or module after test running.
Don't think that it's an important issue because I believe you change test code and navigate to it from prev test result quite rarely

PR #4848, opened 2020-01-20
T: Fix jump from test results to doctest

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4208.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
@@ -8,6 +8,7 @@ package org.rust.cargo.runconfig.test
 import com.intellij.execution.Location
 import com.intellij.execution.PsiLocation
 import com.intellij.execution.testframework.sm.runner.SMTestLocator
+import com.intellij.openapi.fileEditor.OpenFileDescriptor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.search.GlobalSearchScope
@@ -31,9 +32,9 @@ object CargoTestLocator : SMTestLocator {
         path: String,
         project: Project,
         scope: GlobalSearchScope
-    ): List<Location<PsiElement>> {
+    ): List<Location<out PsiElement>> {
         if (protocol != TEST_PROTOCOL) return emptyList()
-        val qualifiedName = toQualifiedName(path)
+        val (qualifiedName, lineNum) = toQualifiedName(path)
 
         // `RsQualifiedNamedElement.qualifiedName` starts with Cargo target name, so if the `qualifiedName` doesn't
         // contain the name separator, we are looking for a target with the corresponding name
@@ -42,7 +43,7 @@ object CargoTestLocator : SMTestLocator {
                 .filter { it.pkg.origin == PackageOrigin.WORKSPACE }
                 .filter { it.normName == qualifiedName }
                 .mapNotNull { it.crateRoot?.toPsiFile(project) }
-                .map { PsiLocation.fromPsiElement<PsiElement>(it) }
+                .map { getLocation(project, it, lineNum) }
                 .toList()
         }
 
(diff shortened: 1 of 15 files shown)


src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestLocator.kt
```diff
@@ -55,21 +57,40 @@ object CargoTestLocator : SMTestLocator {
...
 
-    private fun toQualifiedName(path: String): String {
+    private fun toQualifiedName(path: String): Pair<String, Int?> {
         val targetName = path.substringBefore(NAME_SEPARATOR).substringBeforeLast("-")
-        if (!path.contains(NAME_SEPARATOR)) return targetName
-        val qualifiedName = path.substringAfter(NAME_SEPARATOR)
-        return "$targetName$NAME_SEPARATOR$qualifiedName"
+        if (NAME_SEPARATOR !in path) return targetName to null
+        val qualifiedName = path.substringAfter(NAME_SEPARATOR).substringBefore("#")
+        val lineNum = if ("#" in path) path.substringAfterLast("#").toInt() - 1 else null
```
Undin, line 93: Let's use `toIntOrNull` to avoid potential exception in future - as I understand cargo doesn't guarantee that output of doctest will remain the same

Undin: Or, you can provide more strict rules while creation of location url. For example, create regexp (something like `\(line (\d+)\)`) that will extract line number from test output

Undin: I'd personally prefer the second suggestion

mchernyavsky:
@Undin Done.

src/test/kotlin/org/rustSlowTests/cargo/runconfig/test/CargoTestRunnerTest.kt
```diff
@@ -699,21 +700,27 @@ class CargoTestRunnerTest : CargoTestRunnerTestBase() {
         val sourceElement = cargoProjectDirectory.toPsiDirectory(project)!!
         val configuration = createTestRunConfigurationFromContext(PsiLocation.fromPsiElement(sourceElement))
         val root = executeAndGetTestRoot(configuration)
+        val scope = GlobalSearchScope.allScope(project)
 
         val test = root.findTestByName("sandbox::test")
-        assertTrue("cargo:test://sandbox-[a-f0-9]{16}::test".toRegex().matches(test.locationUrl ?: ""))
+        val testDesc = test.getLocation(project, scope)?.openFileDescriptor!!
+        assertEquals(71, testDesc.offset)
```
Undin, line 707: Actually, I don't like that we check raw offsets because it's quite hard to understand if it's correct or not.
I suggest making real navigation and compare actual editor state with expected
Something like
```kotlin
myFixture.configureFromExistingVirtualFile(cargoProjectDirectory.findFileByRelativePath("src/lib.rs")!!)
desc.navigateIn(myFixture.editor)
myFixture.checkResult(expectedText)
```

Of course, it will make this test much bigger but at the same time much clearer
What do you think?
If you don't like a lot of expected code in the test, you can place expected text in test data

mchernyavsky: @Undin Done.

src/test/kotlin/org/rustSlowTests/cargo/runconfig/test/CargoTestRunnerTest.kt
```diff
@@ -770,6 +835,22 @@ class CargoTestRunnerTest : CargoTestRunnerTestBase() {
...
+        val configuration = createTestRunConfigurationFromContext(PsiLocation.fromPsiElement(sourceElement))
+        val root = executeAndGetTestRoot(configuration)
+        val test = root.findTestByName(testName)
+        val location = test.getLocation(project, GlobalSearchScope.allScope(project))
+        val desc = location?.openFileDescriptor!!
+        desc.navigateIn(editor)
+
+        val newCaret = myFixture.caretOffset
+        assertEquals(oldCaret, newCaret)
+    }
```
Undin, line 852: The main problem of the current tests that you setup expected state to editor at the beginning.
So, if you drop `desc.navigateIn(editor)` line, all tests won't fail.
Also, we still compare raw offsets and if a test fails, the comparison result won't be human readable
I may suggest something like

```suggestion
    private fun checkTestLocation(testName: String, testProject: TestProject) {
        val expectedText = (testProject.psiFile(testProject.fileWithCaret) as PsiFile).text

        myFixture.configureFromTempProjectFile(testProject.fileWithCaret)
        myFixture.editor.caretModel.moveToOffset(0)

        val sourceElement = cargoProjectDirectory.toPsiDirectory(project)!!
        val configuration = createTestRunConfigurationFromContext(PsiLocation.fromPsiElement(sourceElement))
        val root = executeAndGetTestRoot(configuration)
        val test = root.findTestByName(testName)
        val location = test.getLocation(project, GlobalSearchScope.allScope(project))
        val desc = location?.openFileDescriptor!!
        desc.navigateIn(editor)

        myFixture.checkResult(expectedText)
    }
```

mchernyavsky:
bors r=Undin

PR #4860, opened 2020-01-22
Fix PSI produced by PrintlnPostfixTemplate

vlad20012:
The template incorrectly adds a comma to expression instead to match arm.
(this is something internal/most likely does not affect our users)


vlad20012:
bors retry

PR #4862, opened 2020-01-23
PERF&GRAM: lazy parseable & reparseable code block

vlad20012:
This fulfills the second bullet-point of #4355. Now we do not re-parse the entire document when typing in a function body. Instead, we re-parse the only function in which the user is typing. I.e. parsing is now incremental. This can slightly speedup completion in large documents.

Merge after the end of 2019.2 support


Undin:
bors retry

PR #4870, opened 2020-01-23
Fix rustc/cargo messages being split incorrectly.

Jezza:
For some reason, if/when all of the planets align, it's entirely possible that the text event we get lands directly in the middle of an explanation, which tripped the code to drain the text buffer, and process the event.

IntelliJ Console View:
IntelliJ Console View

I printed out the event's text as we received it.
IntelliJ Log (The second line " - }"):
IntelliJ Log

The fix is a bit weird, but I couldn't really think of another way.
I count all of the quotation marks, and if it's even, assume we have a complete message.

I could only think of one more fix which didn't require changes to cargo, which is to manually construct a JSONParser, and hook it up to the incoming events, and only when the parser spits out a complete Object, drain that part of the buffer, and continue.

It would require much more work, but I'm happy to implement that too.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildAdapter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildAdapter.kt
@@ -110,6 +110,12 @@ class CargoBuildAdapter(
         // otherwise the line contains only part of the message.
         if (concat.contains("{\"reason\"") && !concat.endsWith("}\n")) return
 
+        // Bit of a weird hack, but it's entirely possible that the error message contains explanations.
+        // Those explanations can contain "}\n", and even more bizarre, it's entirely possible that
+        // we get the text event that splits the message exactly on that boundary.
+        val count = concat.count { it == '\"' }
+        if ((count and 1) != 0) return
+
         val text = concat.replace(BUILD_PROGRESS_FULL_RE) { it.value.trimEnd(' ', '\r', '\n') + "\n" }
         textBuffer.clear()
 


src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildAdapter.kt
```diff
@@ -110,6 +110,12 @@ class CargoBuildAdapter(
         // otherwise the line contains only part of the message.
         if (concat.contains("{\"reason\"") && !concat.endsWith("}\n")) return
 
+        // Bit of a weird hack, but it's entirely possible that the error message contains explanations.
+        // Those explanations can contain "}\n", and even more bizarre, it's entirely possible that
+        // we get the text event that splits the message exactly on that boundary.
+        val count = concat.count { it == '\"' }
+        if ((count and 1) != 0) return
```
mchernyavsky: It doesn't work for the following code:
```rust
fn main() {
    // "
    0
}
```

Jezza: Yeah, I knew it was fragile, but short of changing cargo, by requiring a NUL byte or something, we're out of luck.

Jezza:
Yeah, we need a better pipe with Cargo.

Windows 10 supports ASCII control seqs, so we should be able to use them for colour now.

Cargo support for a cleaner message pipe would go a long way.
I can close this, and open up a ticket if you want.

mchernyavsky:
Yes, create a new issue, please. And it will be great if we could reproduce this bug.

PR #4872, opened 2020-01-24
PERF&MACRO: Introduce MacroExpansionVFS

### LINKED ISSUE #3968: Expanded macros occupy a lot of space
## Environment

* **Intellij-Rust plugin version:** v0.2.98.2125-191
* **Rust toolchain version:** 1.37.0-nightly (5d8f59f4b 2019-06-04) x86_64-pc-windows-msvc 
* **IDE name and version:** IntelliJ IDEA 2019.1.3
* **Operating system:** Windows 7 x64

## Problem description
I've opened a few of my projects that have `winapi` and `ntapi` crates among their dependencies using the new macro expansion engine and noticed that `.IdeaIC2019.1\system\rust_expanded_macros` already contains **64 478** files and occupies 266 MB of disk space due to the minimum file size on disk being equal to the cluster size (4 KB) with the actual folder size of 53,8 MB.

I really think we should group expanded macros somehow (anyhow) to both reduce the occupied space and improve performance because every request to open and read or map a file is a separate system call for each operation.

Related: https://github.com/intellij-rust/intellij-rust/issues/3685

vlad20012:
Fixes #3968

Expanded macros now are not stored on a disk as individual physical files. Instead, they are stored in a snapshot only (snapshot is an IntelliJ platform concept that means a copy of a part of filesystem stored in embedded DB in IntelliJ directory) with new `MacroExpansionVFS`.

Now macro expansion should occupy much less disk space. Also, this should speed up macro expansion because of fewer system calls (filesystem interaction).

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystem.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystem.kt
@@ -1,0 +1,455 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.macros
+
+import com.intellij.openapi.util.io.BufferExposingByteArrayInputStream
+import com.intellij.openapi.util.io.FileAttributes
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.openapi.vfs.VirtualFileManager
+import com.intellij.openapi.vfs.impl.local.LocalFileSystemBase
+import com.intellij.util.ArrayUtil
+import com.intellij.util.IncorrectOperationException
+import com.intellij.util.PathUtilRt
+import org.rust.lang.core.macros.MacroExpansionFileSystem.Companion.readFSItem
+import org.rust.lang.core.macros.MacroExpansionFileSystem.Companion.writeFSItem
+import org.rust.lang.core.macros.MacroExpansionFileSystem.FSItem
+import org.rust.lang.core.macros.MacroExpansionFileSystem.FSItem.FSDir
+import org.rust.lang.core.macros.MacroExpansionFileSystem.FSItem.FSFile
+import java.io.*
+
+/**
+ * An implementation of [com.intellij.openapi.vfs.VirtualFileSystem] used to store macro expansions.
+ *
+ * The problem is that we need to store tens of thousands of small files (with macro expansions).
+ * A real filesystem is slow, and a lot small files consumes too much of diskspace.
+ * [com.intellij.openapi.vfs.ex.temp.TempFileSystem] can't be used because these files should
+ * persist between IDE restarts, and because `TempFileSystem` stores all files in the RAM.
+ *
+ * [MacroExpansionFileSystem] is a "snapshot-only VFS", i.e. it doesn't have any "backend" (like
+ * a real FS in [com.intellij.openapi.vfs.LocalFileSystem] or RAM in
+ * [com.intellij.openapi.vfs.ex.temp.TempFileSystem]) and file contents are stored only in the
+ * snapshot ([com.intellij.openapi.vfs.newvfs.persistent.PersistentFS]).
+ *
+ * ## Under the hood
+ * The platform VFS implementation ([com.intellij.openapi.vfs.newvfs.persistent.PersistentFS])
+ * sometimes (during [refresh]) invokes methods [getTimeStamp] and [getLength] from a backend
+ * filesystem (i.e. [MacroExpansionFileSystem]) to check that files does not change (and reloads
+ * changed files). To satisfy this mechanism, [MacroExpansionFileSystem] maintain file tree
+ * structure with `last modified` and `length` properties, but without file contents (to be honest,
+ * there is a "temporary" file content [FSItem.FSFile.tempContent] that is cleared once loaded
+ * to the snapshot). [MacroExpansionFileSystem] stores such tree in-memory, but since it should
+ * persist between the IDE restarts, there are methods that allow loading/storing parts of the file
+ * tree externally (see [getDirectory], [setDirectory], [readFSItem], [writeFSItem])
+ *
+ * Since VFS is global and exists regardless of open projects, [MacroExpansionFileSystem] should
+ * either store the file tree in memory for all ever existing projects, or load subtrees on-demand
+ * for opened projects. Storing the file tree in memory for all ever existing projects can be too
+ * expensive, so on-demand solution is preferred.
+ *
+ * For macro expansions we use such directory layout:
+ * "/rust_expanded_macros/<random_project_id>/a/b/<random_file_name>.rs", where "<random_project_id>" is a
+ * project root that should be loaded/unloaded on-demand. Loading is trivial - it's just a combination of
+ * [readFSItem] and [setDirectory] methods. More interesting is unloading. There is a problem with it:
+ * [MacroExpansionFileSystem] still should say `last modified` and `length` to the platform for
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -218,32 +230,113 @@ class MacroExpansionManagerImpl(
...
 
 private class MacroExpansionServiceImplInner(
     private val project: Project,
     val dirs: Dirs,
-    private val storage: ExpandedMacroStorage
+    private val storage: ExpandedMacroStorage,
+    var expansionsDirVi: VirtualFile?
 ) {
+    companion object {
+        fun build(project: Project, dirs: Dirs): () -> MacroExpansionServiceImplInner {
```
Undin: I still don't like it
We have a lot of indirection in macro expansion manager initialization
Can we split this method into two ones where the first one will make some preparation without read action and the second one will create expansion service after it in read action from data prepared by the first method?

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -218,32 +230,113 @@ class MacroExpansionManagerImpl(
...
+                fun(): MacroExpansionServiceImplInner {
+                    val storage = serStorage.deserializeInReadAction(project)
+                    return MacroExpansionServiceImplInner(project, dirs, storage, dir)
+                }
+            } else {
+                MACRO_LOG.debug("Using fresh ExpandedMacroStorage")
+                vfs.createDirectoryIfNotExistsOrDummy(dirs.expansionDirPath)
+                val dir = vfs.findFileByPath(dirs.expansionDirPath) // Nullable!
+                val mes = MacroExpansionServiceImplInner(project, dirs, ExpandedMacroStorage(project), dir)
+                fun(): MacroExpansionServiceImplInner = mes
```
Undin: Both branches look quite similar. Can we unify them?

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionVFS.kt
```diff
@@ -0,0 +1,388 @@
...
+            override var parent: Dir?,
+            override var name: String,
+            @get:Synchronized
+            @set:Synchronized
+            override var timestamp: Long = System.currentTimeMillis()
+        ) : FSItem() {
+            protected open val children: MutableList<FSItem> = mutableListOf()
+
+            @Synchronized
+            fun copyChildren(): List<FSItem> = children.map { it }
```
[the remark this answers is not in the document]
Undin: Maybe it's just better to call `ArrayList` constructor directly from whole `children` list?

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -218,32 +224,114 @@ class MacroExpansionManagerImpl(
...
+            } catch (e: java.nio.file.NoSuchFileException) {
+                null
+            } catch (e: Exception) {
+                MACRO_LOG.warn(e)
+                null
+            }
+        }
+
+        /** See [MacroExpansionFileSystem] docs for explanation of what happens here */
+        private fun loadProjectDirs() {
```
Undin: Hm, what if we call this method from two different projects at the same time?
It seems we need some synchronization here to avoid `File already exists` exception

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystem.kt
```diff
@@ -0,0 +1,448 @@
...
+ * and answer "yes, this directory still exists and unmodified, do not traverse it!" to the platform.
+ * But subtrees "/rust_expanded_macros/<random_project_id>/..." can be omitted. There is a special
+ * class for such fake, "dummy" directories - [FSItem.FSDir.DummyDir], and a method that replaces
+ * a real directory to dummy - [makeDummy].
+ *
+ * So,
+ * 1. prior to any operations with [MacroExpansionFileSystem], a list of all ever created
+ *   "/rust_expanded_macros/<random_project_id>" directories should be loaded.
+ *   This is done by [MacroExpansionServiceImplInner.loadProjectDirs];
+ * 2. when some project "foo" is closed, "/rust_expanded_macros/foo" should be replaced wich
```
Undin:
```suggestion
 * 2. when some project "foo" is closed, "/rust_expanded_macros/foo" should be replaced with
```

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystem.kt
```diff
@@ -0,0 +1,448 @@
...
+ * [readFSItem] and [setDirectory] methods. More interesting is unloading. There is a problem with it:
+ * [MacroExpansionFileSystem] still should say `last modified` and `length` to the platform for
+ * _all_ files, so we can't just remove a project subtree. But with a knowledge of how [refresh]
+ * works we made a trick.
+ *
+ * Quick explanation of the [refresh] process works. First, a file or directory should be marked as dirty
+ * with [com.intellij.openapi.vfs.newvfs.NewVirtualFile.markDirty]. This method also marks dirty all
+ * ancestor directories of the file, i.e. if we mark a file "/foo/bar/baz.rs" as dirty, "bar", "foo"
+ * and "/" (root) directories will be also marked dirty. Then [refresh] machinery, starting from the root
+ * directory, traverse dirty or modified ([getTimeStamp]) directories.
```
Undin:
```suggestion
 * directory, traverses dirty or modified ([getTimeStamp]) directories.
```

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionVfsBatch.kt
```diff
@@ -69,18 +63,18 @@ class LocalFsMacroExpansionVfsBatch(
     }
 
     private sealed class PathImpl : MacroExpansionVfsBatch.Path {
-        abstract fun toPath(): Path
+        abstract fun toPath(): String
```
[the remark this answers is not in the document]
vlad20012: The idea of this is that we don't expose the implementation of path to the interface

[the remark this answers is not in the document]
Undin: But you use explicit cast to use this method anyway. So looks like this hiding of implementation details is quite useless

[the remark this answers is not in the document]
vlad20012: I means exposing outside of `MacroExpansionVfsBatch.kt` (implementation is private), i.e. to MacroExpansionTaskBase

Undin:
bors retry

Undin:
@vlad20012 some tests on win fails

src/test/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystemTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+        vfs.deleteFile("/foo")
+        fullyRefreshDirectory(vfs.root)
+        assertNull(vfs.findFileByPath("/foo/baz.txt"))
+        assertNull(vfs.findFileByPath("/foo"))
+    }
+}
+
+private fun VirtualFileSystem.findNonNullFileByPath(path: String): VirtualFile =
+    findFileByPath(path) ?: error("Fil not found: $path")
```
Undin:
```suggestion
    findFileByPath(path) ?: error("File not found: $path")
```

vlad20012:
bors try

vlad20012:
bors try

vlad20012:
bors r=undin

vlad20012:
bors retry

Issue #2757, opened 2018-08-09
Switch to local processing when suggestion completions for dependencies in Cargo.toml


matklad:
I am not entirely sure about this, but looks like today, when completing dependence names and versions in Cargo.toml, IntelliJ issues requrests to the crates.io server. This unfortunately makes UX clunky: completion is slow, often freezes the UI thread, and often fails with a loud message "unable to reach crates.io".

I think it should be possible to avoid network and use local processing for completion. For example, here's how you can find available crates:

```
~/.cargo/registry/index/github.com-1ecc6299db9ec823 master*
λ git ls-tree --full-tree  -r remotes/origin/master | sed 's:.*/::' | rg ripgrep -A 10 -B 10
```

With some more legwork, it should be possible to extract version information as well.

Jezza:
I've poked the cargo team, so we'll see if they'll give us access to the index directly.
Otherwise, we're probably going to forgo the index, and just rely on git.

## What I've discovered so far:

#### The index is stored at: "$CARGO_HOME/registry/index/github.com-1ecc6299db9ec823"

The "registry/index" is [hardcoded](https://github.com/rust-lang/cargo/blob/a6a395c6909d8417a1b48d4fd3a23e073108da62/src/cargo/util/config/mod.rs#L276).

The last part is from the [SourceId](https://github.com/rust-lang/cargo/blob/0b653a43ba49ebd67d04dc23cb4071d569a8671e/src/cargo/sources/registry/mod.rs#L387-L391).
It's siphashed, and then appended to the host url.

I doubt it's going to change anytime soon, so we can probably hardcode them ourselves.

### Querying the git repo

Querying the repo should be straight forward enough.
Annoyingly, [git ls-tree doesn't support pathspec](http://git.661346.n2.nabble.com/ls-tree-and-wildcards-td7553581.html), which means we need to dump and then parse the entire output.
It shouldn't be _too_ much of a problem, as we can cache it, but we have to be careful about that.

Once the user starts typing, we need to act based on how many characters there are, because as the [cargo index documentation states](https://github.com/rust-lang/cargo/blob/0b653a43ba49ebd67d04dc23cb4071d569a8671e/src/cargo/sources/registry/mod.rs#L90-L93): They're split based on their length.

* Only 1 character.
* Only 2 characters.
* Only 3 characters.
* The rest.

[...]

PR #4873, opened 2020-01-24
[WIP] Implement a basic parsing and handling for cargo's local index

Jezza:
Basic implementation for #2757

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/completion/CargoRegistry.kt
+++ b/toml/src/main/kotlin/org/rust/toml/completion/CargoRegistry.kt
@@ -1,0 +1,59 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.completion
+
+import com.intellij.codeInsight.completion.CompletionUtil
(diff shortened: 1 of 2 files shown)


Kobzol:
Just for some inspiration, this is a VSCode plugin that does something similar: https://github.com/serayuzgur/crates/pull/77/files. A similar approach is also in this crate https://github.com/LucianoBestia/cargo_reg_local

vlad20012:
I am very wary of such solutions. 
1. Uses `git` binary. Git is not mandatory for our plugin. 
2. Relies on cargo index internal structure. 

Ideally, Cargo should just provide a subcommand with for us.

Kobzol:
What about using crates.io as a fallback and only use the `git` + internal index poking strategy as a best effort strategy if `git` is available? (at least until Cargo provides a proper way to query the DB).
I refactored the crate resolving slightly in https://github.com/intellij-rust/intellij-rust/pull/5439, it could be used to add another (git) resolver. We can also add some caching for the API resolver.

Kobzol:
Superseded by https://github.com/intellij-rust/intellij-rust/pull/6390.

PR #4874, opened 2020-01-25
Fix LLDB pretty-printers tests and Rc/Arc rendering on 2020.1

### LINKED ISSUE #4851: Enable pretty printer tests with 2020.1
## Problem description

Pretty printer tests fail with 2020.1. They were temporarily [disabled on CI](https://github.com/intellij-rust/intellij-rust/blob/ee8133abcdde5f2e57762a8fd1c5cb7140aa50c6/.travis.yml#L54). It would be great to find out the reason and fix it.
Note, it looks like everything is fine in production

ortem:
* Since 2020.1 our LLDB emits exact variable names instead of indexes (e.g. `(i32) x = 42` instead of `(i32) $0 = 42`).

* Since 2020.1 our LLDB contains Rust type system so `SBType.GetBasicType()` fails when called on Rust type supposedly because `eBasicTypeLong` is C type, which may be incompatible with Rust `SBType` somehow. In contrast, `SBTarget.GetBasicType()` uses only `ClangASTContext` to compute the type and doesn't depend on Rust `SBType`, so it works fine in that case. In practice, it fixes the wrong rendering of `Rc` and `Arc` structs on 2020.1

Fixes #4851


pretty_printers_tests/tests/rc_arc.rs
```diff
@@ -1,13 +1,16 @@
 // min-version: 1.30.0
+// max-version: 1.39.0
```
Undin: What's the problem with `Rc`/`Arc` pretty printers on 1.40?
Previously we launched tests with 1.40 and they passed

ortem: They fail on 201 platform with 1.40 for some reason. I'm investigating the problem.

ortem: Should be fixed now

PR #4885, opened 2020-01-27
Fix potential memory leak in suppress quick fix

Undin:


Undin:
bors retry

Undin:
bors retry

PR #4890, opened 2020-01-28
Get rid of `SyntaxHighlighterFactory`s

vlad20012:
They are not needed on 2019.3.
Syntax highlighter can be registered directly with `lang.syntaxHighlighter` extension point


Undin:
bors r=vlad20012

PR #4913, opened 2020-02-03
RES: resolve/completion for trait object inherent impl

### LINKED ISSUE #3781: Support resolving functions from `impl for dyn Trait`
Currently IntelliJ Rust does not resolve functions defined directly on impl on a trait (and as a result, type inference breaks at these calls):

```rust
trait Greeter {
    // Function defined on a trait -- works fine!
    fn greet(&self, name: &str) -> String;
}

impl dyn Greeter + '_ {
    // This function should be callable on any `&dyn Greeter`, but IntelliJ cannot resolve it!
    fn anonymous(&self) -> String {
        self.greet("anonymous")
    }
}

fn invoke(greeter: &dyn Greeter) -> String {
    // This function does not get resolved!
    greeter.anonymous()
}
```

### LINKED ISSUE #4732: Rust plugin does not resolve constant from dyn trait impl block
## Environment

* **IntelliJ Rust plugin version: 0.2.111.2147-193**
* **IDE name and version: CLion 2019.3**

## Problem description

In the following example, reference resolution doesn't work for `Foo::FOO` (cannot navigate), but works fine for `Bar::BAR`:

```rust
trait Foo {}
impl dyn Foo {
  const FOO: &'static str = "http://example.com";
}

struct Bar {}
impl Bar {
  const BAR: &'static str = "http://example.com";
}

fn main() {
  // Doesn't resolve :(
  let x = Foo::FOO;
  // Works fine!
  let y = Bar::BAR;
}
```

P.S. Could be similar to #3781 ?

### LINKED ISSUE #4943: fmt::Debug is not inferred for thread::Result
## Environment

* **IntelliJ Rust plugin version:** 0.2.115.2157-193
* **Rust toolchain version:** cargo 1.41.0 (626f0f40e 2019-12-03)
* **IDE name and version:** CLion 2019.3.3 (CL-193.6015.37)
* **Operating system:** Linux 5.0.0-37-generic

## Problem description
`unwrap` is not offered in code completion for Result returned from the `std::thread::JoinHandle::join` method. This is probably caused by the fact that `fmt::Debug` is not inferred for the Error type parameter of the returned Result type, which is `Box<dyn Any + Send + 'static>`.

## Steps to reproduce
```rust
let t = std::thread::spawn(|| ());
t.join()./* unwrap is not offered here */
```
When I use unwrap, it compiles fine, but code completion doesn't offer it.

rrevenantt:
Fixes #3781
Fixes #4732
Fixes #4943

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -242,8 +242,10 @@ class ImplLookup(
     private fun rawFindImplsAndTraits(ty: Ty): List<TraitImplSource> {
         val implsAndTraits = mutableListOf<TraitImplSource>()
         when (ty) {
-            is TyTraitObject ->
+            is TyTraitObject -> {
                 ty.trait.flattenHierarchy.mapTo(implsAndTraits) { TraitImplSource.Object(it.element) }
+                findExplicitImpls(ty) { implsAndTraits += TraitImplSource.ExplicitImpl(it); false }
+            }
             is TyFunction -> {
                 findExplicitImpls(ty) { implsAndTraits += TraitImplSource.ExplicitImpl(it); false }
                 implsAndTraits += fnTraits.map { TraitImplSource.Object(it) }
@@ -398,7 +400,11 @@ class ImplLookup(
             val isAppropriateImpl = ctx.canCombineTypes(formalSelfTy, selfTy) &&
                 // Check that trait is resolved if it's not an inherent impl; checking it after types because
                 // we assume that unresolved trait is a rare case
-                (cachedImpl.isInherent || cachedImpl.implementedTrait != null)
+                (cachedImpl.isInherent || cachedImpl.implementedTrait != null) &&
+                // ignore blanket implementations for trait objects for now because
+                // both trait objects and trait bounds are represented here as TyTraitObject
+                // but they should be treated differently in terms of applicable impls
+                (selfTy !is TyTraitObject || type !is TyTypeParameter)
             isAppropriateImpl && processor(cachedImpl)
         }
     }
--- a/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
@@ -715,6 +715,32 @@ class RsTypeAwareResolveTest : RsResolveTestBase() {
         }
     """)
 
+    fun `test trait object inherent impl`() = checkByCode("""
+        trait Test{}
+        impl dyn Test{
+            fn test(&self){}
+              //X
+        }
+        fn main(){
+            let a:&dyn Test = unimplemented!();
+            a.test()
+             //^
+        }
+    """)
+
+    fun `test generic trait object inherent impl`() = checkByCode("""
+        trait Test<T>{}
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
```diff
@@ -242,8 +242,14 @@ class ImplLookup(
     private fun rawFindImplsAndTraits(ty: Ty): List<TraitImplSource> {
         val implsAndTraits = mutableListOf<TraitImplSource>()
         when (ty) {
-            is TyTraitObject ->
+            is TyTraitObject -> {
                 ty.trait.flattenHierarchy.mapTo(implsAndTraits) { TraitImplSource.Object(it.element) }
+                findExplicitImpls(ty) {
+                    // skip blanket implementations
+                    if (ty == it.impl.typeReference?.type) implsAndTraits += TraitImplSource.ExplicitImpl(it);
```
vlad20012: Why is it needed? I don't understand the comment. Should be already filtered in `findExplicitImpls`. 

Btw `==` should not be used for type comparison in type inference, use `ctx.canCombineTypes()` instead

src/test/kotlin/org/rust/lang/core/resolve/RsTypeAwareResolveTest.kt
```diff
@@ -715,6 +715,19 @@ class RsTypeAwareResolveTest : RsResolveTestBase() {
         }
     """)
 
+    fun `test trait object inherent impl`() = checkByCode("""
```
vlad20012: Please, add a test with `impl<T> dyn Foo<T>`

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
rrevenantt: The problem is that when resolving associate items for trait bounds, `selfTy` is also a trait object, and in that case `findExplicitImpls` finds also blanket implementations which are inapplicable to trait bound.
I have tried to fix it, but there is too much code that relies on traits being trait object declarations.

I know that blanket impls can be applicable to actual trait objects, but I also thought that partial implementation is better that nothing, so I have added this condition as a workaround.

vlad20012:
:+1: 
bors r+

src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
vlad20012: I fixed it in #4953

vlad20012:
Btw I found this test passed on this branch. Interesting. Looks like it works because `impl for T` resolved for `&<any>` type
```
trait Foo { fn foo(&self); }
trait Bar { fn bar(&self); }
impl<T> Bar for T {
    fn bar(&self) { }
}     //X
fn foo(a: &dyn Foo) {
    a.bar()
}   //^
```

vlad20012:
This is the correct test for blanket impls:
```rust
trait Foo {}
trait Bar { fn bar(&self); }
impl<T: ?Sized> Bar for T {
    fn bar(&self) {}
}    //X
fn foo(a: &dyn Foo) {
    (*a).bar()
}      //^
```
(note that it works only with `?Sized` bound)

rrevenantt:
So should i add this test and remove workaround after #4953 merged?

vlad20012:
Hmm, I think I'll do it myself.

Issue #4714, opened 2019-12-02
Error recognizing modules behind a macro


krojew:
## Environment

* **IntelliJ Rust plugin version:** 0.2.111.2147-193
* **Rust toolchain version:** 1.40 nightly
* **IDE name and version:** CLion 2019.3
* **Operating system:** Windows 10

## Problem description

One of the most popular crates, tokio, has modules which cannot be parsed by the plugin. Examples include the tokio::sync::mpsc module, which is imported inside a macro:

```rust
cfg_sync! {
    mod barrier;
    pub use barrier::{Barrier, BarrierWaitResult};

    pub mod mpsc;

    mod mutex;
    pub use mutex::{Mutex, MutexGuard};

    pub mod oneshot;

    pub(crate) mod semaphore;

    mod task;
    pub(crate) use task::AtomicWaker;

    pub mod watch;
}
```

mpsc is not recognized by the plugins at all.

## Steps to reproduce

Use tokio, try to include tokio::sync::mpsc::Whatever, see error.

tdyas:
I am also seeing this issue with the tokio crate. My project is at https://github.com/tdyas/tftp-rs. Load it into IntelliJ and you will see that references to `UdpSocket` (`tokio::net::UdpSocket`) are red because the Rust plugin cannot find a definition.

stjepangolemac:
Yeah it can't detect the tokio TcpListener too, just keeps importing the std::net one.

lights0123:
This issue is also demonstrated in the [uom](https://crates.io/crates/uom) crate, which has no (useful) assistance by this plugin as all (useful) modules are imported by macros.

SpyMachine:
reqwest 0.10 now has this problem too.

Undin:
It's fixed in #4914 and will be landed in the next release (I hope).
I close this issue as a duplicate of #4627 because they have the same reason

Issue #4706, opened 2019-11-29
Syntax highlighting/code completion not working for (some) tokio modules


YenForYang:
## Environment

* **IntelliJ Rust plugin version:**
0.2.110.2150-193 (also tried 192)
* **Rust toolchain version:**
1.39.0
* **IDE name and version:**
Clion 2019.3, also tested 2019.2.5
* **Operating system:**
Windows 10 Version 1909, also tested MacOS 10.14.6

## Problem description

This problem is possibly very similar to #4627 but with `tokio` instead of `async_std`...something to do with macro resolution probably.
The main reason I'm posting this is because I've found a (rather obvious) *very fragile* **unsafe** workaround to mitigate the pain of
not having proper code analysis (in CLion) temporarily. 

The issue is described best by some code:

```rust
use tokio::net::TcpStream; // 
              //  ^^^^^^ Lack of color here, can't Cmd/Ctrl-B here.

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
  let tcp = TcpStream::connect("420.4.2.0:4204").await?;
         // ^^^^^^^^^^^^^^^^^^ Unresolved reference: `TcpStream`
```

## Steps to reproduce

In `Cargo.toml`, add:

```toml
[dependencies.tokio]
version = "0.2.2"
features = [ "full" ]
```
(I added "full" to enable all possible features, just to make sure the issue isn't caused by a feature not being enabled)

## Steps to hack around

What we do here is basically trade a real eye for 2 fake ones:

From `tokio-0.2.2/src/net/mod.rs`, note lines 28-32:
```rust
cfg_tcp! {  
    pub mod tcp;  
    pub use tcp::listener::TcpListener;  
    pub use tcp::stream::TcpStream;  
}
```
Well, `cfg_tcp!` causes the issue...so I deemed it necessary to hardcode:
```rust
// cfg_tcp! {  
    pub mod tcp;  
    pub use tcp::listener::TcpListener;  
    pub use tcp::stream::TcpStream;  
// }
```

Maybe the plugin could do something like this behind the scenes as a temporary hack to get things working; just throwing out ideas because I've heard that this issue has been around for quite awhile.

ZOXEXIVO:
Same problem with 

`use async_std::net::{TcpListener, TcpStream};`

![image](https://user-images.githubusercontent.com/14014921/70095769-9f026200-1636-11ea-9c18-404338219b7a.png)

[...]

Issue #4679, opened 2019-11-23
Code completion does not work for a crate - "File is not included in module tree, analysis is not available" - Reproducible


YenForYang:
## Environment

* **IntelliJ Rust plugin version:**
0.2.110.2150-192
* **Rust toolchain version:**
1.39.0
* **IDE name and version:**
Clion 2019.2.5
* **Operating system:**
Windows 10 Version 1909

## Problem description
Should be easy to reproduce.

Code completion for `reqwest` does not work when using `reqwest = "0.10.0-alpha.2"` under `[dependencies]` in `Cargo.toml` (the latest version of reqwest).

Whats interesting is that it doesn't work for version "0.10.0-alpha.1" either, but it **does** work for "0.9.22" (the last 0.9 release). Unfortunately I can't use 0.9.22 as there are breaking changes in 0.10+

![Demo](https://user-images.githubusercontent.com/16462860/69485426-ea319f00-0e04-11ea-8a31-16d4c5645ff6.gif)

Note in the above GIF at the start (where the cursor is). `new()` in `ClientBuilder::new()` isn't blue.
## Steps to reproduce
Create a new Rust project in CLion, and

`Cargo.toml` add these dependencies:

```
...
[dependencies]
reqwest = "0.10.0-alpha.2"
tokio = "0.2.0-alpha.6"
```

in `main.rs` have something like:
```rust
use reqwest::ClientBuilder;

#[tokio::main]
pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let client = ClientBuilder::new()
        .http2_prior_knowledge()
...
```

YenForYang:
Seems to be the same issue described in https://github.com/intellij-rust/intellij-rust/issues/4627#

Undin:
@YenForYang thanks for the report!
The reason is module declaration is generated by macro call.
Try to enable experimental macro expansion engine (see #3628). It should solve your issue

> Seems to be the same issue described in #4627

The global reason is the same, yes. But in this case experimental engine can process it but in case of #4627 - not yet because `mod` declaration is under several macro calls that is not supported yet

Undin:
It's fixed in #4914 and will be landed in the next release (I hope).
I close this issue as a duplicate of #4627 because they have the same reason

Issue #4627, opened 2019-11-13
Name resolution and completion don't work with async_std/tokio/reqwest crates


hhggit:
## Environment

* **IntelliJ Rust plugin version:** 0.2.109.2149-192
* **Rust toolchain version:** 1.39.0
* **IDE name and version:** Clion 2019.2.1
* **Operating system:** Ubuntu 19.10

## Problem description

`#[cfg(feature = "...")]` in macros could not be found.

https://github.com/async-rs/async-std/blob/f611ceccc87aaf6b6cfb5bdee8bfffaffee5498f/src/utils.rs#L134-L144
https://github.com/async-rs/async-std/blob/f611ceccc87aaf6b6cfb5bdee8bfffaffee5498f/src/lib.rs#L203-L211

``` rust
/// Declares std items.
#[allow(unused_macros)]
#[doc(hidden)]
macro_rules! cfg_std {
    ($($item:item)*) => {
        $(
            #[cfg(feature = "std")]
            $item
        )*
    }
}

cfg_std! {
    // ...
    pub mod task;
}
```

## Steps to reproduce
The plugin cannot find the declaration of `block_on` to go. But it cound find after changing `cfg_std! {   pub mod task;} `to `#[cfg(feature = "std")]pub mod task;` .
``` rust
use async_std::task;

fn main() {
    task::block_on(async {
        println!("Hello, world!");
    })
}
```

Undin:
> #[cfg(feature = "...")] in macros could not be found.

`cfg` attribute in macro call does not affect anything in this case.

@vlad20012 Looks like the plugin fails to resolve macros from `utils.rs` (except in `lib.rs` where the corresponding module is declared) so all modules declared inside the corresponding macro calls cannot be resolved

vlad20012:
The issue has been known for a long time. The new macro expansion engine (#3628) doesn't solve it for now. Recently we discussed some possible solutions, so I'll try to prototype something.

swarmer:
This is a hard blocker for me because it's common to have large chunks of code in macros like this, and intellij degrades into a very dumb editor (navigation/search fails) in these cases.

caibirdme:
any update for this issue?

s97712:
> any update for this issue?

+1

dobrakmato:
Was looking to make a workaround by using `tokio` but it apparently also uses `cfg` inside marcos.

UkonnRa:
[...]

PR #4914, opened 2020-02-03
MACRO: process modules declared inside macro call better

Undin:
The previous version of macro expansion engine processed all project files on the first step.
It led to a situation when a module was added to the module structure via macro call and at the same time, it had own macro calls.
It such cases we failed to expand module's macro calls because name resolution for macro calls didn't work properly.

These changes introduce pending files - project files that are not belong to the module tree.
Such files are postponed to expand their macro calls until they belong to module tree.
It should solve "File is not included into module tree" issue for some popular crates like `async-std` and `tokio`

Fixes the following issues when new macro expansion engine is enabled: 
* #4627
* #4679
* #4706
* #4714
* #4929

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -66,7 +66,7 @@ class ExpandedMacroStorage(val project: Project) {
     private fun deserialize(sfs: List<SourceFile>) {
         for (sf in sfs) {
             sourceFiles.put(sf.fileId, sf)
-            getOrPutStagedList(sf).add(sf)
+            getOrPutStagedList(sf.depthNotSync).add(sf)
             sf.forEachInfoNotSync { if (it.fileId > 0) expandedFileToInfo.put(it.fileId, it) }
         }
     }
@@ -320,6 +320,9 @@ class SourceFile(
     private val fileUrl: String get() = file.url
     val fileId: Int get() = file.fileId
     val project: Project get() = storage.project
+
+    // Used only for deserialization
+    val depthNotSync: Int get() = _depth
 
     var depth: Int
         get() {
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -219,7 +219,7 @@ class ExpandedMacroStorage(val project: Project) {
 
     companion object {
         private val LOG = Logger.getInstance(ExpandedMacroStorage::class.java)
-        private const val STORAGE_VERSION = 8
+        private const val STORAGE_VERSION = 9
         const val RANGE_MAP_ATTRIBUTE_VERSION = 2
 
         fun load(project: Project, dataFile: Path): ExpandedMacroStorage? {
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -49,6 +49,8 @@ import java.nio.file.Path
 import java.util.concurrent.locks.ReentrantLock
 import java.util.zip.DeflaterOutputStream
 import java.util.zip.InflaterInputStream
+import kotlin.collections.ArrayList
+import kotlin.collections.HashMap
 import kotlin.concurrent.withLock
 
 class ExpandedMacroStorage(val project: Project) {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -224,7 +233,7 @@ private fun updateDirs0(projectDirName: String): Dirs {
     )
 }
 
-private data class Dirs(
+data class Dirs(
```
Undin: Looks like it's redundant change

src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
```diff
@@ -111,16 +117,26 @@ class ExpandedMacroStorage(val project: Project) {
...
-                }
-            } ?: emptyList()
+    fun removeSourceFilesFromStep(files: Collection<SourceFile>, step: Int) {
+        checkWriteAccessAllowed()
+        testAssert {
+            files.all { it.depth == step }
         }
+
+        stepped[step]?.removeAll(files)
+        files.forEach { it.depth = -1 }
```
vlad20012: I've got such crash
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 64
  at org.rust.lang.core.macros.ExpandedMacroStorage.getOrPutStagedList(ExpandedMacroStorage.kt:92)
  at org.rust.lang.core.macros.ExpandedMacroStorage.deserialize(ExpandedMacroStorage.kt:69)
  at org.rust.lang.core.macros.ExpandedMacroStorage.access$deserialize(ExpandedMacroStorage.kt:56)
  at org.rust.lang.core.macros.ExpandedMacroStorage$Companion.load(ExpandedMacroStorage.kt:257)
  at org.rust.lang.core.macros.MacroExpansionServiceImplInner$Companion.build(MacroExpansionManager.kt:441)
  at org.rust.lang.core.macros.MacroExpansionManagerImpl$projectOpened$1.call(MacroExpansionManager.kt:135)
  at org.rust.lang.core.macros.MacroExpansionManagerImpl$projectOpened$1.call(MacroExpansionManager.kt:106)
  at com.intellij.openapi.application.impl.ApplicationImpl$2.call(ApplicationImpl.java:274)
  at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
  at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
  at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
  at java.base/java.lang.Thread.run(Thread.java:834)
```

b/c I restarted the IDE during macro expansion and it saved (and then loaded) `-1` as a depth of some SourceFile. Let's try to merge `removeSourceFilesFromStep` and `addSourceFilesToStep` into one write action in order to avoid any inconsistent state of the storage

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
```diff
@@ -86,12 +90,12 @@ abstract class MacroExpansionTaskBase(
         try {
             submitExpansionTask()
 
-            MACRO_LOG.trace("Awaiting")
-            val millis = measureNanoTime {
+            MACRO_LOG.info("Awaiting")
```
vlad20012: Let's change log levels in other PR. (I don't think it's reasonable to log "Awaiting" as INFO, and other messages should be also improved)

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
```diff
@@ -139,7 +146,9 @@ abstract class MacroExpansionTaskBase(
                 var extractableList: List<Extractable>?
                 do {
                     extractableList = expansionSteps.nextOrNull()
-                    currentStep.incrementAndGet()
+                    val step = currentStep.incrementAndGet()
+                    MACRO_LOG.debug("Expansion step: $step")
```
vlad20012: Also, let's remove this log. It's very verbose. Or maybe use TRACE level

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
```diff
@@ -256,6 +286,37 @@ abstract class MacroExpansionTaskBase(
...
 
+    private fun addPendingFilesToStep(pendingFiles: List<Extractable>, step: Int) {
+        val duration = measureNanoTime {
+            if (pendingFiles.isEmpty()) return@measureNanoTime
+            // TODO: split [pendingFiles] into chunks to reduce UI freezes (if needed)
+            WriteAction.runAndWait<Nothing> {
+                storage.addSourceFilesToStep(pendingFiles.mapToSet { it.sf }, step)
+            }
+        }
+        MACRO_LOG.debug("Addition of pending files to step $step: ${duration / 1000} μs")
```
vlad20012: Same here

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionTask.kt
Undin: But why is it bothering you?
I've increased level of important messages so now you don't need to set debug level at all to see almost everything that previously required debug level

src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
Undin: Looks ok now
Check it again, please

vlad20012:
:+1:
bors r+

Issue #4865, opened 2020-01-23
No way to "unhide" cargo tool window


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.115.SNAPSHOT-193-dev
* **Rust toolchain version:** 1.39.0 (4560ea788 2019-11-04) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3.2 (CL-193.5662.56)
* **Operating system:** macOS 10.15

## Problem description

With #4102 the cargo tool window is now hidden for non-Rust project, but there's no way to enable it back if the project *is* the Rust one but the plugin doesn't understand that.

## Steps to reproduce

Consider the following project structure

```
project
└── crate1
    ├── Cargo.lock
    ├── Cargo.toml
    └── src
        └── main.rs
```

open `project` as a project root. CLion doesn't recognize it as a Rust one and I can't find a way to open cargo tool window.

#4836 helps somewhat, but you have to restart IDE after attaching cargo project for tool window to reappear.

PR #4919, opened 2020-02-04
ACT: Show cargo tool window after attaching a cargo project

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4865


src/main/kotlin/org/rust/ide/notifications/NoCargoProjectNotificationProvider.kt
```diff
@@ -22,6 +23,7 @@ class NoCargoProjectNotificationProvider(project: Project) : RsNotificationProvi
             subscribe(CargoProjectsService.CARGO_PROJECTS_TOPIC, object : CargoProjectsService.CargoProjectsListener {
                 override fun cargoProjectsUpdated(projects: Collection<CargoProject>) {
                     updateAllNotifications()
+                    initializeToolWindowIfNeeded(project)
```
Undin: Why it's located here?
Maybe we have a better place for it?

PR #4920, opened 2020-02-04
T: REPL integration

dima74:
Add tests for REPL integration:
* Communication with Evcxr
* Completion in REPL console

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -42,6 +42,7 @@ jobs:
               run: |
                 rustc --version
                 rustup component list --installed
+                if [ ! "${{ matrix.rust-version }}" \< "1.40.0" ]; then cargo install evcxr_repl; fi # BACKCOMPAT: Rust 1.39.0
 
             - name: Check license
               run: ./check-license.sh
--- a/src/test/kotlin/org/rust/ide/console/RsConsoleCompletionTestBase.kt
+++ b/src/test/kotlin/org/rust/ide/console/RsConsoleCompletionTestBase.kt
@@ -1,0 +1,75 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.console
+
+import org.rust.ProjectDescriptor
+import org.rust.RsTestBase
+import org.rust.WithStdlibRustProjectDescriptor
+import org.rust.lang.core.completion.RsCompletionTestFixture
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.RsReplCodeFragment
+
+@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
+abstract class RsConsoleCompletionTestBase : RsTestBase() {
+
+    private lateinit var completionFixture: RsCompletionTestFixture
+
+    override fun setUp() {
+        super.setUp()
+        completionFixture = RsCompletionTestFixture(myFixture)
+        completionFixture.setUp()
+    }
+
+    override fun tearDown() {
+        completionFixture.tearDown()
+        super.tearDown()
+    }
+
+    protected fun checkContainsCompletion(previousLines: String, lastLine: String, variants: Array<String>) {
+        val previousLinesTrimmed = previousLines.trimIndent()
+        val lastLineTrimmed = lastLine.trimIndent()
+
+        doCheckContainsCompletion(previousLinesTrimmed, lastLineTrimmed, variants)
+        if (previousLines.isBlank()) return
+
+        val lastLines = previousLinesTrimmed + lastLineTrimmed
+        doCheckContainsCompletion("", lastLines, variants)
+        doCheckContainsCompletion("", "if true { $lastLines }", variants)
+        doCheckContainsCompletion("", "fn main() { $lastLines }", variants)
+    }
+
+    private fun doCheckContainsCompletion(previousLines: String, lastLine: String, variants: Array<String>) {
+        inlineReplCodeFragment(lastLine, previousLines)
+
+        val lookups = myFixture.completeBasic()
+        checkNotNull(lookups) {
+            "Expected completions that contain all ${variants.contentToString()}, but no completions found"
+        }
+        for (variant in variants) {
+            if (lookups.all { it.lookupString != variant }) {
+                error("Expected completions that contain `$variant`, but got ${lookups.map { it.lookupString }}")
+            }
+        }
+    }
+
+    protected fun checkSingleCompletion(previousLines: String, lastLineBefore: String, lastLineAfter: String) {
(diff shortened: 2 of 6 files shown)


src/test/kotlin/org/rustSlowTests/console/RsEvcxrTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        // success command
+        writer.println("1 + 2")
+        writer.flush()
+        assertEquals(scanner.nextLine(), "3")
+        expectPrompt(scanner, true)
+
+        // fail command
+        writer.println("let foo = bar;")
+        writer.flush()
+        assert(scanner.nextLine().contains("not found in this scope"))
```
Undin: Let's use `assertTrue` or `check` instead of `assert`

src/test/kotlin/org/rustSlowTests/console/RsEvcxrTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        assert(scanner.nextLine().contains("cannot find value `bar` in this scope"))
+        expectPrompt(scanner, false)
+
+        process.destroy()
```
Undin: Let's put it into `finally` block to destroy process even if some check above throws exception

src/test/kotlin/org/rustSlowTests/console/RsEvcxrTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import org.rust.cargo.RsWithToolchainTestBase
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.toolchain.Cargo
+import java.io.PrintWriter
+import java.util.*
+
+class RsEvcxrTest : RsWithToolchainTestBase() {
+
+    fun testProcessOutputFormat() {
```
Undin:
```suggestion
    fun `test process output format`() {
```
We use names with spaces enclosed in backticks for tests

src/test/kotlin/org/rust/ide/console/RsConsoleCompletionTestBase.kt
```diff
@@ -0,0 +1,58 @@
...
+        doCheckReplCompletion("", lastLines, variants)
+        doCheckReplCompletion("", "if true { $lastLines }", variants)
+        doCheckReplCompletion("", "fn main() { $lastLines }", variants)
+    }
+
+    private fun doCheckReplCompletion(previousLines: String, lastLine: String, variants: Array<String>) {
+        inlineReplCodeFragment(lastLine, previousLines)
+
+        val lookups = myFixture.completeBasic()
+        checkNotNull(lookups) {
```
Undin: Note, `myFixture.completeBasic()` can return `null` when there is only one completion variant

dima74: Thank you, I replaced all tests which expect one completion variant to use `RsCompletionTestFixture::executeSoloCompletion`

src/test/kotlin/org/rustSlowTests/console/RsEvcxrTest.kt
```diff
@@ -0,0 +1,73 @@
...
+            assertTrue(scanner.nextLine().contains("not found in this scope"))
+            assertTrue(scanner.nextLine().contains("cannot find value `bar` in this scope"))
+            expectPrompt(scanner, false)
+        } finally {
+            process.destroy()
+        }
+    }
+
+    private fun createProcess(): Process {
+        val commandLine = project.toolchain!!.evcxr()!!.createCommandLine(null)
```
Undin, line 45: `RsWithToolchainTestBase` uses real local file system so here you can pass `cargoProjectDirectory.pathAsPath.toFile()` as working directory and don't change `org.rust.cargo.toolchain.Evcxr#createCommandLine`

src/test/kotlin/org/rust/ide/console/RsConsoleCompletionTestBase.kt
```diff
@@ -0,0 +1,75 @@
...
+    private fun inlineReplCodeFragment(lastLine: String, previousLines: String) {
+        // create main.rs file, which will be used as crate root
+        InlineFile("")
+
+        InlineFile(lastLine, RsConsoleView.VIRTUAL_FILE_NAME).withCaret()
+
+        val codeFragment = myFixture.file as RsReplCodeFragment
+        val crateRoot = codeFragment.crateRoot as RsFile?
+        codeFragment.context = RsConsoleCodeFragmentContext.createContext(project, crateRoot, previousLines)
+        assert(codeFragment.crateRoot != null)
```
Undin, line 73: Use `check` or `assertNotNull` instead of `assert` because purpose of `assert` is to check code invariants sometimes (asserts can be disabled in VM)

src/test/kotlin/org/rust/ide/console/RsConsoleCompletionTestBase.kt
```diff
@@ -0,0 +1,75 @@
...
+        completionFixture = RsCompletionTestFixture(myFixture)
+        completionFixture.setUp()
+    }
+
+    override fun tearDown() {
+        completionFixture.tearDown()
+        super.tearDown()
+    }
+
+    protected fun checkContainsCompletion(previousLines: String, lastLine: String, variants: Array<String>) {
```
Undin, line 31: Maybe it's worth to replace `variants: Array<String>` to `vararg variants: String`?
It allows to avoid `arrayOf` in tests

Undin:
@dima74 Sorry for so long response

dima74:
@Undin thanks for comments! I updated code

Regarding Windows: `expectPrompt` does not work for Windows (on Evcxr master), so I think it is better to enable tests on Windows in another PR (in order to not block #5218)

.github/workflows/check.yml
```diff
@@ -42,6 +42,7 @@ jobs:
               run: |
                 rustc --version
                 rustup component list --installed
+                if [ "${{ matrix.rust-version }}" != "1.32.0" ]; then cargo install evcxr_repl; fi # BACKCOMPAT: Rust 1.32.0
```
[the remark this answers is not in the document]
dima74: It is necessary because evcxr supports 1.40 and above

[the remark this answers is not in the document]
dima74: Good idea, updated

Undin:
bors retry

dima74:
Looks like rebase is needed

dima74:
@Undin looks like any PR which modifies github workflows should be merged manually: https://forum.bors.tech/t/resource-not-accessible-by-integration/408/12

Undin:
Hm, it's weird. #5106 was merged by bors although PR changed workflow file

dima74:
Moving to #5232

PR #4925, opened 2020-02-06
PERF: don't parse function bodies during indexing

### LINKED ISSUE #4355: Make function bodies lazy parseable & reparseable

vlad20012:
Now during indexation, we only parse function bodies that can contain local items (that should be indexed). I measured that for most projects now only about 3% function bodies should be parsed, i.e. other 97% should not. This speeds up indexation by about 20%.
Closes #4355

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -8,9 +8,13 @@ package org.rust.lang.core.stubs
 import com.intellij.lang.*
 import com.intellij.lang.parser.GeneratedParserUtilBase
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Key
 import com.intellij.psi.PsiFile
 import com.intellij.psi.StubBuilder
 import com.intellij.psi.impl.source.tree.LazyParseableElement
+import com.intellij.psi.impl.source.tree.RecursiveTreeElementWalkingVisitor
+import com.intellij.psi.impl.source.tree.TreeElement
+import com.intellij.psi.impl.source.tree.TreeUtil
 import com.intellij.psi.stubs.*
 import com.intellij.psi.tree.*
 import com.intellij.util.BitUtil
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -82,6 +95,70 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
...
+        // See `RsBlockStubType.reuseCollapsedTokens`
+        val b = PsiBuilderFactory.getInstance().createBuilder(node.psi.project, node, null, RsLanguage, node.chars)
+        var prevToken: IElementType? = null
+        while (true) {
+            val token = b.tokenType ?: break
+            val looksLikeStubElement = token in ITEM_DEF_KWS
+                || token == CONST && prevToken != MUL     // `const` but not `*const`
+                || token == SHA && prevToken == EXCL      // `#!`
+                || token == IDENTIFIER && prevToken == FN // `fn foo` (but not `fn()`)
+                || token == IDENTIFIER && b.tokenText == "union" && b.lookAhead(1) == IDENTIFIER
```
Undin: Maybe it's worth to add a test case (maybe even with some specific testmark) for each specific rule in this condition?

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -82,6 +95,70 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
...
+            return cachedHasItemsOrAttrs
+        }
+        return computeBlockCanHaveStubs(node)
+    }
+
+    private fun assertIsBlock(node: ASTNode) {
+        check(node.elementType == BLOCK) { "Expected block, found: ${node.elementType}, text: `${node.text}`" }
+    }
+
+    private fun computeBlockCanHaveStubs(node: ASTNode): Boolean {
```
Undin:
```suggestion
    private fun computeBlockMayHaveStubs(node: ASTNode): Boolean {
```

src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -82,6 +95,70 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
...
+ *
+ * There are two functions that do mostly the same but differs in how they use the cache.
+ * - [computeAndCache] computes the value and then puts it in the cache. Used in [RsFunctionStub.Type.createStub]
+ * - [getAndClearCached] retrieves and removes the value from the cache or compute the value if it does not
+ *   present. Expected to be called after [computeAndCache] during indexing (because a function is a parent of a
+ *   block, so it is processed earlier)
+ *
+ * See tests in `RsLazyCodeBlockIsNotExpandedDuringStubBuildingTest`
+ */
+private object BlockCanHaveStubsHeuristic {
```
Undin: Maybe `BlockStubUtils` name is better here?
Just to avoid some complex name and pointing to implementation details

vlad20012:
bors r=undin

Issue #2310, opened 2018-02-20
Operators on references of primitives cannot be resolved


sticnarf:
```rust
let x: i32 = 1;
let y: i32 = 2;
let rx: &i32 = &x;
let ry: &i32 = &y;

let z = x + y;  // OK
let z = x + ry;  // OK
let z = rx + y;  // `Add` cannot be resolved, nor is the type of `z`
let z = rx + ry;  // Same as the previous line
```

If the left operand is a reference to a primitive, the operator cannot be properly resolved.

I find something else weird. If I omit the explicit type of `x`:

```rust
let x = 1;  // Type of x is correctly resolved to be `i32`
let y: i32 = 2;

let z = x + y;  // Trait `Add` is resolved, but the type of `z` remains `<unknown>`
let z = x + ry;  // Same as the previous line

let v = vec![z]; // `v` is resolved to be a Vec<<unknown>>
```

Although the type of `x` is properly resolved, the type of `z` cannot be resolved. (However it works in the previous example with the type of `x` explicit)

vlad20012:
Already fixed

PR #4931, opened 2020-02-06
Failing test for cannot resolve operators 

Dmitry-Borodin:
operators cannot be resolved for reference primitives, looks like we already have issue for that #2310 

let a = 4 - &5; is a valid expression in rust for primitive types, implemented by src/libcore/internal_macros.rs:21


vlad20012:
I'm sorry for the long answer.

Unfortunately, this test is not correct. Operator overloading for references doesn't work automagically. There are special impls for references to primitives in Rust stdlib. Also, #2310 is fixed now.

Issue #4366, opened 2019-09-08
Type info action doesn't take into account type aliases


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.105.2731-192-nightly
* **Rust toolchain version:** 1.39.0-nightly (9eae1fc0e 2019-08-23) x86_64-apple-darwin
* **IDE name and version:** IU-192.6603.8
* **Operating system:** macOS 10.13.6

## Problem description
After #4243 we have info about type alias for structs and enums. But `Type Info` intention doesn't use it and always insert base type instead of a type alias.

![2019-09-08 14 32 38](https://user-images.githubusercontent.com/2539310/64487619-8526ce00-d245-11e9-8172-da67019368d6.gif)

## Steps to reproduce
```rust
use std::cell::RefCell;

type BoxedCell<T> = Box<RefCell<T>>;

fn foo<T>(c: &BoxedCell<T>) {
    let b = c;
}

```

Kobzol:
I just noticed this, it seems like it might be caused by the same problem, so I'll just attach it here as a test case:
```rust
```

Undin:
@Kobzol No, your issue has other reason (see #2778).
In general, it's because index (that the plugin uses while completion for items from `impl` blocks) doesn't know relation between `Wrapped` and `Direct`.
Probably, we can support such case as well but it seems such feature will cause a significant performance regression

PR #4932, opened 2020-02-06
INT: use text with aliases in Type Info intention

Kobzol:
I replaced `toString()` with `insertionSafeTextWithAliases` to show aliases in the `Type Info` hint. I didn't find any tests for this, so I'm not sure if this is enough. @Undin thoughts?

Fixes https://github.com/intellij-rust/intellij-rust/issues/4366

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
@@ -9,6 +9,7 @@ import com.intellij.lang.ExpressionTypeProvider
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
 import com.intellij.psi.impl.FakePsiElement
+import org.rust.ide.presentation.textWithAliasNames
 import org.rust.lang.core.macros.findExpansionElementOrSelf
 import org.rust.lang.core.macros.findMacroCallExpandedFromNonRecursive
 import org.rust.lang.core.macros.mapRangeFromExpansionToCallBodyStrict
@@ -35,7 +36,7 @@ class RsExpressionTypeProvider : ExpressionTypeProvider<PsiElement>() {
 
     override fun getInformationHint(element: PsiElement): String {
         val type = getType(element)
-        return type.toString().escaped
+        return type.textWithAliasNames.escaped
     }
 
     private fun getType(element: PsiElement): Ty = when (element) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/hints/RsExpressionTypeProvider.kt
```diff
@@ -35,7 +36,7 @@ class RsExpressionTypeProvider : ExpressionTypeProvider<PsiElement>() {
 
     override fun getInformationHint(element: PsiElement): String {
         val type = getType(element)
-        return type.toString().escaped
+        return type.insertionSafeTextWithAliases.escaped
```
Undin: You should use `textWithAliasNames` instead of `insertionSafeTextWithAliases ` (I know that it's marked a `TestOnly`. It's time to drop this annotation) not to change the previous style of how we render unknown types. `insertionSafeTextWithAliases` is used as text for PSI element creation in intentions/quick fixes/refactorings/etc.

Kobzol:
I added a basic test. The provider returned escaped text, so I introduced a method which returns the raw text for easier testing. Let me know if the test structures looks OK. In that case I will add more tests.

src/test/kotlin/org/rust/ide/hints/RsExpressionTypeProviderTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    private fun doTest(@Language("Rust") code: String, type: String) {
+        InlineFile(code).withCaret()
+
+        val element = myFixture.elementAtCaret
+
+        val provider = RsExpressionTypeProvider()
+        val expressions = provider.getExpressionsAt(element)
+        assertEquals(type, expressions.joinToString(", ") { provider.getInformationHintUnescaped(it) })
```
Undin: Let's better use `getInformationHint` to ensure that we don't forget to unescape something

Kobzol: Ok. I didn't want to use it because then if you have a test failure you see something like "Expected: & lt;Something;& gt". But it's not a big deal.

Kobzol:
Added more tests.

src/test/kotlin/org/rust/ide/hints/RsExpressionTypeProviderTest.kt
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package org.rust.ide.hints
+
+import org.intellij.lang.annotations.Language
+import org.rust.ProjectDescriptor
+import org.rust.RsTestBase
+import org.rust.WithStdlibRustProjectDescriptor
+import org.rust.openapiext.escaped
+import kotlin.test.assertEquals
```
Undin: Could you remove unused imports?

Kobzol:
Removed unused imports and squashed commits.

Issue #3082, opened 2018-11-26
'Implement Methods' functionality expands type alias


theduke:
## Problem description

When using the `Implement Methods` functionality to autogenerate method stubs for a trait implementation,
type aliases in the implemented method signatures are resolved to the original type, rather than using the type alias.

This has always bugged me. I would expect the type alias to be used, not the original type.

## Steps to reproduce

```rust
type Alias = u32;

trait SomeTrait {
    fn t() -> Alias;
}

impl SomeTrait for Alias {
  // EXPECTED expansion:
  fn t() -> Alias { ... }

  // ACTUAL expansion:
  fn t() -> u32 { ... }
}
```

PR #4938, opened 2020-02-07
INT: add support for type aliases in ImplementMembersFix

Kobzol:
This PR changes `RsTypeReference.substAndGetText` to support both lifetimes and aliases. It fixes alias generation in ImplementMembersFix, let's see what CI says about the rest of tests :)

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3082


mchernyavsky:
Thanks!
bors r+

Issue #4905, opened 2020-01-31
Coverage doesn't work with 2020.1


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.116.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.42.0-nightly (212b2c7da 2020-01-30) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 EAP (CL-201.4515.29)
* **Operating system:** macOS 10.13.6

## Problem description
After launch anything with coverage on 2020.1 EAP, IDE shows empty coverage report.
![image](https://user-images.githubusercontent.com/2539310/73541512-b5187d00-4443-11ea-9f7d-1b735b822c3b.png)

Note, everything is fine with 2019.3

## Steps to reproduce
```rust
fn main() {
    foo()
}

fn foo() {
    println!("Hello!")
}
```

PR #4939, opened 2020-02-07
RUN: Fix code coverage runner on 2020.1

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/4905.


Undin:
@mchernyavsky it would be great to implement tests for coverage to prevent such regressions

Issue #4937, opened 2020-02-07
Implement "Generate constructor" (the `new` method)


dvtomas:
Oftentimes I write code like this:

```rust
struct S {
  foo: Foo,
  bar: Bar,
  baz: Baz
}

impl S {
  pub fn new(foo: Foo, bar: Bar, baz: Baz) -> Self {
    S { foo, bar, baz }
  }
}
```

It would be nice if we could have the `new` method auto-generated, e.g. by using `Alt-Insert` when on the struct or inside the impl block.

Undin:
@dvtomas thanks for the suggestion!
Do you know that almost the same feature is already implemented in #2815?
Just invoke `Code | Generate` action and choose `Generate Constructor` option. Note, it works only when you caret inside the corresponding struct.
![2020-02-09 02 00 15](https://user-images.githubusercontent.com/2539310/74093197-eea25580-4adf-11ea-8a3d-4bd62e0b506b.gif)

And, indeed, it would be great if the plugin provides this action inside the corresponding `impl` block

dvtomas:
Aha, I didn't know about that. That's exactly what I had on mind. I guess the problem was i was executing the "Generate" pop-up inside the `impl` block, not on the struct itself, and inside the `impl` block it doesn't offer this action. Maybe it would be nice to offer it in the `impl` context as well.

Anyway, thanks!

PR #2815, opened 2018-08-31
ACT: add 'Rust.GenerateConstructor' action

Voronchikhin:

PR #4944, opened 2020-02-08
INT: AddConstructor intention

Kobzol:
This PR adds support for generating a constructor (a `new` associated function) for structs. The intention is not complete yet, it doesn't create lifetime/type parameters on the impl block (amongst other things).

I wanted to ask if this is the right direction to go and whether you want to add this intention at all.
Some questions to resolve:
- Offer this intention only on the Struct definition or also in an existing struct impl block?
- Detect if there is already a `new` associated function or a `Default` impl present and do not offer this intention in such a case?
- Detect if there is already an impl block present in the same file to avoid creating a new one?
    - Is there an intention to merge impl blocks?
    - The AddImplIntention does no such check.
- Should this be a simple intention or should it be integrated into the `Alt+Insert` generate dialog? Possibly by extending RsTraitImplementationInspection?
    - It would probably make more sense to be in the dialog, as is common in other languages (Generate constructor/getter/setter etc.)

Fixes: #4937

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
@@ -18,6 +18,9 @@ import org.rust.lang.core.psi.ext.ancestorOrSelf
 import org.rust.lang.core.psi.ext.isTupleStruct
 import org.rust.lang.core.psi.ext.namedFields
 import org.rust.lang.core.psi.ext.positionalFields
+import org.rust.lang.core.types.Substitution
+import org.rust.lang.core.types.emptySubstitution
+import org.rust.lang.core.types.type
 import org.rust.openapiext.checkWriteAccessAllowed
 import org.rust.openapiext.checkWriteAccessNotAllowed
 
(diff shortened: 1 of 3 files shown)


Undin:
@Kobzol this feature is already implemented in #2815.
See `GenerateConstructorHandler` class.
But we can improve it. Some improvements are already listed in your questions

> Offer this intention only on the Struct definition or also in an existing struct impl block?

The first context is already supported and it would be great to provide the action in `impl` block as well

> Detect if there is already a new associated function

Probably, it's too expensive to check if `new` function already exists but we can do it after invocation of the action. See the similar action for constructor generation in java classes.

> a Default impl present and do not offer this intention in such a case

But `Default` is about creation of default instances. I may want to create struct with arbitrary field values. So we should provide this action even the corresponding struct has implementation of `Default` trait

> Is there an intention to merge impl blocks?

As I remember we don't have such code

> Should this be a simple intention or should it be integrated into the Alt+Insert generate dialog?

It's already implemented as action in `Generate` dialog. Not sure that we want to provide this action as intention. On the other hand, it would increase its discoverability

Kobzol:
Oh, I didn't know about it, since it wasn't available on the impl block. I'll try to add it there then.

Kobzol:
Added support for running the refactoring from impl blocks.

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -31,37 +31,59 @@ class GenerateConstructorAction : CodeInsightAction() {
...
-            insertNewConstructor(structItem, chosenFields, editor)
+            insertNewConstructor(context.struct, context.implBlock, chosenFields, editor)
         }
     }
 
-    private fun insertNewConstructor(structItem: RsStructItem, selectedFields: List<ConstructorArgument>, editor: Editor) {
+    private fun isImplBlockValid(impl: RsImplItem): Boolean {
+        if (impl.traitRef != null) return false
+
+        return impl.members?.functionList?.find { it.name == "new" } == null
```
Undin: It doesn't work because you can have several `impl` blocks for the same item.
You need to check that you don't have `new` method for struct in whole project.

@vlad20012 do we have some API to check method existence for some type without taking into account type params? Or we should use `org.rust.lang.core.resolve.indexes.RsImplIndex.Companion#findPotentialImpls` directly?
Note, it may be quite heavy operation so you need to move these check from `GenerateConstructorHandler#isValidFor` to `GenerateConstructorHandler#invoke` and abort action if needed

src/test/kotlin/org/rust/ide/refactoring/GenerateConstructorTest.kt
```diff
@@ -166,4 +217,18 @@ class GenerateConstructorTest : RsTestBase() {
             }
         }
     }
+
+    private fun checkNotAvailable(@Language("Rust") code: String) {
+        InlineFile(code)
+
+        withMockStructMemberChooserUi(object : StructMemberChooserUi {
+            override fun selectMembers(project: Project, all: List<RsStructMemberChooserObject>): List<RsStructMemberChooserObject>? {
+                assert(false) // this should not be reached
```
Undin: Let's replace `assert` with `error`

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -31,37 +31,59 @@ class GenerateConstructorAction : CodeInsightAction() {
...
+    }
+
+    private fun insertNewConstructor(structItem: RsStructItem, implBlock: RsImplItem?,
+                                     selectedFields: List<ConstructorArgument>, editor: Editor) {
         checkWriteAccessAllowed()
         val project = editor.project ?: return
         val structName = structItem.name ?: return
         val psiFactory = RsPsiFactory(project)
-        val impl = psiFactory.createInherentImplItem(structName, structItem.typeParameterList, structItem.whereClause)
+        val impl = implBlock ?: let {
```
Undin: Let's replace `let` with simple `if/else`. In this case `let` makes code less readable

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -31,37 +31,59 @@ class GenerateConstructorAction : CodeInsightAction() {
...
+            val impl = psiFactory.createInherentImplItem(structName, structItem.typeParameterList, structItem.whereClause)
+            structItem.parent.addAfter(impl, structItem) as RsImplItem
+        }
         val anchor = impl.lastChild.lastChild
         val constructor = getFunction(structItem, selectedFields, psiFactory)
```
Undin: Note, this code doesn't take into account type arguments because it wasn't needed in previous implementation, i.e. for the following code
```rust
struct S<T>(T);

impl S<i32> {
    /*caret*/
}
```
you generate invalid code
```rust
impl S<i32> {
    pub fn new(field0: T) -> Self {
        S(field0)
    }
}
```

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
Kobzol: You're right, but before this PR the intention also didn't check if another constructor exists. So it's an added (best-effort) check that shouldn't introduce regressions. Therefore I think that we could keep it even if we don't use a global lookup of impls of some kind. A global lookup would be of course nicer :-)

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
Kobzol: Thanks, I didn't think of that.

Kobzol:
Added support for type substitution from the impl block.

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -90,30 +126,30 @@ data class ConstructorArgument(
     val type: RsTypeReference?
```
Undin: Looks like it's unused now

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -90,30 +126,30 @@ data class ConstructorArgument(
...
 ) {
 
-    val dialogRepresentation: String get() = "$argumentIdentifier: ${type?.text ?: "()"}"
+    val dialogRepresentation: String get() = "$argumentIdentifier: ${typeReference ?: "()"}"
```
Undin:
```suggestion
    val dialogRepresentation: String get() = "$argumentIdentifier: $typeReference"
```

Also, let's rename `typeReference` to `typeReferenceText` to avoid possible misleading

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -31,52 +34,85 @@ class GenerateConstructorAction : CodeInsightAction() {
...
-        val insertedImpl = structItem.parent.addAfter(impl, structItem) as RsImplItem
-        editor.caretModel.moveToOffset(insertedImpl.textOffset + insertedImpl.textLength - 1)
+        editor.caretModel.moveToOffset(impl.textOffset + impl.textLength - 1)
     }
 
-    private fun getFunction(structItem: RsStructItem, selectedFields: List<ConstructorArgument>, psiFactory: RsPsiFactory): RsFunction {
+    private fun getFunction(structItem: RsStructItem,
+                            selectedFields: List<ConstructorArgument>,
+                            psiFactory: RsPsiFactory,
+                            substitution: Substitution): RsFunction {
```
Undin: It should look like the following:

```suggestion
    private fun getFunction(
        structItem: RsStructItem,
        selectedFields: List<ConstructorArgument>,
        psiFactory: RsPsiFactory,
        substitution: Substitution
    ): RsFunction {
```
according to kotlin [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting)

`Put parameter on one line` and `Put parameter on separate lines` intentions may help you.
Also, don't forget to check all changed function signature

src/test/kotlin/org/rust/ide/refactoring/GenerateConstructorTest.kt
```diff
@@ -156,7 +239,7 @@ class GenerateConstructorTest : RsTestBase() {
     ) {
         withMockStructMemberChooserUi(object : StructMemberChooserUi {
             override fun selectMembers(project: Project, all: List<RsStructMemberChooserObject>): List<RsStructMemberChooserObject>? {
-                assertEquals(all.map { it.text }, chooser.map { it.member })
+                assertEquals(chooser.map { it.member }, all.map { it.text })
```
Undin: 👍

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -31,52 +34,85 @@ class GenerateConstructorAction : CodeInsightAction() {
...
-        val constructor = getFunction(structItem, selectedFields, psiFactory)
+        val constructor = getFunction(structItem, selectedFields, psiFactory, substitution)
         impl.lastChild.addBefore(constructor, anchor)
-        val insertedImpl = structItem.parent.addAfter(impl, structItem) as RsImplItem
-        editor.caretModel.moveToOffset(insertedImpl.textOffset + insertedImpl.textLength - 1)
+        editor.caretModel.moveToOffset(impl.textOffset + impl.textLength - 1)
     }
 
-    private fun getFunction(structItem: RsStructItem, selectedFields: List<ConstructorArgument>, psiFactory: RsPsiFactory): RsFunction {
+    private fun getFunction(structItem: RsStructItem,
```
Undin: `createFunction` or `createConstructorFunction` or something similar looks better here

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
Kobzol: Aah, this intention was what I was looking for! Thanks.

Kobzol:
Refactored and renamed tests to `GenerateConstructorHandlerTest` so that it's named `<action>Test` to make it easier to find.

If you can point me to an API to use for global impl lookup, I'll try to add it here.

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -31,52 +34,90 @@ class GenerateConstructorAction : CodeInsightAction() {
         generateConstructorHandler.isValidFor(editor, file)
 }
 
+data class Context(val struct: RsStructItem, val implBlock: RsImplItem? = null)
```
Undin: Let's make it private not to add it in global namespace

src/main/kotlin/org/rust/ide/refactoring/generateConstructor/GenerateConstructorHandler.kt
```diff
@@ -86,37 +127,38 @@ class GenerateConstructorHandler : LanguageCodeInsightActionHandler {
...
-                    it.typeReference?.text ?: "()",
-                    it.typeReference
+                    it.typeReference?.substAndGetText(substitution) ?: "()"
                 )
             }
         }
     }
 }
+
+val RsImplItem.isSuitableForConstructor: Boolean
```
Undin:
```suggestion
private val RsImplItem.isSuitableForConstructor: Boolean
```

src/test/kotlin/org/rust/ide/refactoring/GenerateConstructorHandlerTest.kt
```diff
@@ -5,14 +5,15 @@
...
+import com.intellij.openapi.actionSystem.ex.ActionManagerEx
 import com.intellij.openapi.project.Project
 import org.intellij.lang.annotations.Language
 import org.rust.RsTestBase
 import org.rust.ide.refactoring.generateConstructor.RsStructMemberChooserObject
 import org.rust.ide.refactoring.generateConstructor.StructMemberChooserUi
 import org.rust.ide.refactoring.generateConstructor.withMockStructMemberChooserUi
 
-class GenerateConstructorTest : RsTestBase() {
+class GenerateConstructorHandlerTest : RsTestBase() {
```
Undin: Maybe `GenerateConstructorActionTest` is better?

src/test/kotlin/org/rust/ide/refactoring/GenerateConstructorHandlerTest.kt
```diff
@@ -156,12 +239,26 @@ class GenerateConstructorTest : RsTestBase() {
...
         }
     }
+
+    private fun checkNotAvailable(@Language("Rust") code: String) {
+        InlineFile(code)
+
+        withMockStructMemberChooserUi(object : StructMemberChooserUi {
+            override fun selectMembers(project: Project, all: List<RsStructMemberChooserObject>): List<RsStructMemberChooserObject>? {
+                error("unreachable")
+                return null
```
Undin: Unreachable

Undin:
> If you can point me to an API to use for global impl lookup, I'll try to add it here.

@Kobzol we've discussed it with @vlad20012 and decided that current implementation is ok because there are case when two `new` function is allowed. For example,
```rust
struct S<T>(T);

impl S<i32> {
    pub fn new(field0: i32) -> Self {
        S(field0)
    }
}

impl S<u32> {
    pub fn new(field0: u32) -> Self {
        S(field0)
    }
}
```

And it's rather hard to check it. In combination with the fact that several inherent `impl` blocks for the same struct is not a common case, it's not worth to waste time to it

Kobzol:
Good point, it would be difficult to look for other constructors when type parameters are involved.

PR #4953, opened 2020-02-10
RES: Don't use ImplLookup for trait-relative paths

vlad20012:
1. Better resolve for `Self`-related paths in traits:
```rust
trait Foo {
    fn foo() {}
}    //X
trait Bar {
    fn bar() where Self: Foo {
        Self::foo();
    }        //^
}
```
2. Better type inference for trait-relative paths with type parameters (e.g. `Foo::<Bar>::baz()` where `Foo` is a trait)
3. Take into account blanket `impl<T> for T` impls when looking for trait object methods. Continuation of #4913

The PR is a result of [the discussion](https://github.com/intellij-rust/intellij-rust/pull/4913#discussion_r376583843).

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -252,6 +252,7 @@ class AutoImportFix(element: RsElement) : LocalQuickFixOnPsiElement(element), Hi
 
                     is TraitImplSource.TraitBound -> return null
                     is TraitImplSource.Object -> return null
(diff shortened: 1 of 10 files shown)


vlad20012:
bors delegate+ rrevenantt
@rrevenantt, could you, please, take a look?

vlad20012:
bors delegate=rrevenantt

rrevenantt:
@vlad20012 looks good, it is also compatible with #4729, which contains my another fix in this area.

BTW, could you add reviewing it to your plans? seems like it got missed because of a random CI failure.

bors r+

PR #4958, opened 2020-02-12
TYPE: render type arguments of type aliases in inlay hints in proper way

### LINKED ISSUE #4871: Type hint is incorrect when using alias for generic type
Given the following code:

```rust
struct Wrapper<T> {
    wrapped: T,
}

type WrappedString = Wrapper<String>;

fn foo() -> WrappedString {
    WrappedString {
        wrapped: String::new(),
    }
}

fn main() {
    let var = foo();
}
```

The type hint shown for `var` is `WrappedString<String>`. This is incorrect; it should be either `WrappedString`, or `Wrapper<String>`.

Undin:
Fixes #4871


vlad20012:
Let's add a test from the issue: an alias without type parameters for a type with type parameters

Issue #4089, opened 2019-07-03
Warn about missing agrument for a format string


lancelote:
## Environment

* **Intellij-Rust plugin version:** 0.2.100.2128-192
* **Rust toolchain version:** rustc 1.35.0 (3c235d560 2019-05-20)
* **IDE name and version:** CLion 2019.2 Nightly 192.5596
* **Operating system:** macOS 10.14.4

## Problem description

The following code

```rust
println!("You guessed: {}");
```

results in `cargo check` complaint:

```
error: 1 positional argument in format string, but no arguments were given
  --> .../sample.rs:12:28
   |
12 |     println!("You guessed: {}")
   |                            ^^

error: aborting due to previous error
```

Would be great if the plugin will show a warning in the editor. At the moment there is none.

PR #4969, opened 2020-02-14
ANN: check existence and type of parameters in format macros

Kobzol:
This PR builds on top of https://github.com/intellij-rust/intellij-rust/pull/3869 and adds additional checks which look for missing or invalid parameters in format macros. After this check exists we can possibly add quick fixes to add missing parameters.

Questions:
- I need to check if a `Ty` implements a given trait. I tried to use `isTraitImplemented` from `ImplLookup`, but I had to make it public and it also doesn't seem to work in tests, I'm probably doing something wrong.
- Should this check be performed every time? Now I only do the check if there are not large errors in the format (for example missing `}`). If there is such a fundamental error, it will probably produce a lot of (false positive) parameter errors, so the parameter check should probably be skipped.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4089
Umbrella issue: https://github.com/intellij-rust/intellij-rust/issues/2256

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
@@ -13,124 +13,149 @@ import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
 import org.intellij.lang.annotations.Language
 import org.rust.ide.colors.RsColor
-import org.rust.lang.core.psi.RsLitExpr
-import org.rust.lang.core.psi.RsLiteralKind
-import org.rust.lang.core.psi.RsMacroCall
+import org.rust.ide.presentation.render
+import org.rust.lang.core.macros.MacroExpansionMode
+import org.rust.lang.core.macros.macroExpansionManager
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.descendantOfTypeStrict
 import org.rust.lang.core.psi.ext.macroName
 import org.rust.lang.core.psi.ext.startOffset
-import org.rust.lang.core.psi.kind
+import org.rust.lang.core.psi.ext.withSubst
+import org.rust.lang.core.resolve.KnownItems
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.TraitRef
+import org.rust.lang.core.types.implLookup
+import org.rust.lang.core.types.ty.TyInteger
+import org.rust.lang.core.types.ty.stripReferences
+import org.rust.lang.core.types.type
 import org.rust.lang.utils.parseRustStringCharacters
 
 class RsFormatMacroAnnotator : AnnotatorBase() {
-
     override fun annotateInternal(element: PsiElement, holder: AnnotationHolder) {
         val formatMacro = element as? RsMacroCall ?: return
 
-        val formatStr = formatMacro
-            .formatMacroArgument
-            ?.formatMacroArgList
-            ?.getOrNull(macroToFormatPos(formatMacro.macroName))
+        val macroPos = macroToFormatPos(formatMacro.macroName) ?: return
+        val macroArgs = formatMacro.formatMacroArgument?.formatMacroArgList ?: return
+        val formatStr = macroArgs
+            .getOrNull(macroPos)
             ?.descendantOfTypeStrict<RsLitExpr>()
             ?: return
 
-        val literalKind = (formatStr.kind as? RsLiteralKind.String) ?: return
-        val rawTextRange = literalKind.offsets.value ?: return
-        val (unescapedText, sourceMap, _) = parseRustStringCharacters(rawTextRange.substring(literalKind.node.text))
+        val parseCtx = parseParameters(formatStr) ?: return
 
-        val arguments = formatParser.findAll(unescapedText)
-        var currentOffset = 0
-
-        fun IntRange.toSourceRange(additionalOffset: Int = currentOffset): TextRange =
-            TextRange(sourceMap[first + additionalOffset], sourceMap[last + additionalOffset] + 1)
-                .shiftRight(formatStr.startOffset + rawTextRange.startOffset)
-
-        fun AnnotationHolder.highlightArgument(range: IntRange?, color: RsColor = RsColor.IDENTIFIER) {
-            if (range != null && !range.isEmpty()) {
-                // BACKCOMPAT: 2019.3
-                @Suppress("DEPRECATION")
-                createAnnotation(
-                    HighlightSeverity.INFORMATION,
-                    range.toSourceRange(),
-                    null
-                ).textAttributes = color.textAttributesKey
-            }
-        }
-
-        val key = RsColor.FORMAT_SPECIFIER
-        val highlightSeverity = if (isUnitTestMode) key.testSeverity else HighlightSeverity.INFORMATION
-        for (arg in arguments) {
-            currentOffset = 0
-            val textRange = arg.range.toSourceRange()
+        val errors = checkSyntaxErrors(parseCtx)
+        for (error in errors) {
             // BACKCOMPAT: 2019.3
             @Suppress("DEPRECATION")
-            holder.createAnnotation(highlightSeverity, textRange, null)
-                .textAttributes = key.textAttributesKey
+            holder.createErrorAnnotation(error.range, error.error)
+        }
 
-            if (arg.groups[3] != null) {
-                // BACKCOMPAT: 2019.3
-                @Suppress("DEPRECATION")
-                holder.createErrorAnnotation(
-                    arg.range.toSourceRange(),
-                    "Invalid format string: unmatched '}'"
-                )
-            }
-            if (arg.groups[1] != null) {
+        highlightParametersOutside(parseCtx, holder)
 
-                val inside = arg.groups[2] ?: error(" should not be null because can match empty string")
-                currentOffset = inside.range.first
-                val matchResult = formatArgParser.find(inside.value)
-                    ?: error(" should be not null because can match empty string")
+        // skip advanced checks and highlighting if there are syntax errors
+        if (errors.isNotEmpty()) {
+            return
+        }
 
-                if (!arg.groups[1]!!.value.endsWith("}")) {
-                    val possibleEnd = matchResult.value.length - 1
-                    // BACKCOMPAT: 2019.3
-                    @Suppress("DEPRECATION")
-                    holder.createErrorAnnotation(
-                        (possibleEnd..possibleEnd).toSourceRange(),
-                        "Invalid format string: } expected.\nIf you intended to print `{` symbol, you can escape it using `{{`"
-                    )
-                    continue
-                }
+        highlightParametersInside(parseCtx, holder)
 
-                val validParsedEnd = matchResult.range.last + 1
-                if (validParsedEnd != inside.value.length) {
-                    // BACKCOMPAT: 2019.3
-                    @Suppress("DEPRECATION")
-                    holder.createErrorAnnotation(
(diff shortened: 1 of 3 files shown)


vlad20012:
1.1 Use `ImplLookup.canSelect` that is used in `isTraitImplemented` implementation. It has a more complex signature but helps to avoid bugs with generic traits (it's not your case, but anyway).

1.2 The test failed because these traits are implemented for primitives via macros. Impls from macros are work in tests only with `@ExpandMacros(MacroExpansionScope.ALL, "std")` annotation. I'd recommend to use custom types and implement format traits manually instead of primitives.

2. Seems right. And anyway it looks like it's simple to change such behavior in the future.

Kobzol:
Thanks, now it works fine. The tests are a bit too verbose to my liking, I need to put `<FORMAT_SPECIFIER>` everywhere. I managed to "turn that off" using `ignoreExtraHighlighting = true`, but then it also ignores errors in the output, which is bad. But it's not a big deal.

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -13,29 +13,49 @@ import com.intellij.openapiext.isUnitTestMode
...
 import org.rust.lang.core.psi.ext.startOffset
-import org.rust.lang.core.psi.kind
+import org.rust.lang.core.psi.ext.withSubst
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.TraitRef
+import org.rust.lang.core.types.implLookup
+import org.rust.lang.core.types.type
 import org.rust.lang.utils.parseRustStringCharacters
 
+sealed class FormatArgument(val range: TextRange, val match: MatchResult, val type: String) {
```
vlad20012: MatchResult is not used?

vlad20012:
About verbose tests: we usually create different annotators for highlighting and errors to deal with it, but in this case, it is undesirable because a format pattern would be parsed twice. I found a way to ignore a specific HighlightSeverity in a test, but it looks too hacky to me, so let's leave with these verbose tests for now...

Undin:
@vlad20012 Maybe it can be solved by extracting tests with semantic checking into separate class? In new class, we may omit `annotationFixture.registerSeverities(RsColor.values().map(RsColor::testSeverity))` line that seems should fix issue

vlad20012:
@Undin No, it doesn't help. The solution is much hackier :(

Kobzol:
Regarding 4: good idea. Two questions:

```rust
#[derive(Debug)]
struct S;

println!("{0} {0:?}", S);
```
- should this highlight `S`? It would be good for consistency, but we maybe want to add a link to the offending parameter in the error.

Also when something like this happens:
```rust
println!("{}");
```
There's no argument to highlight, so should we highlight the closing brace of the `println!` call?

Kobzol:
Do we want to supports things like `log::info!`? It's a very common case, so it would be nice. But it would also push the current heuristic for checking if a macro is a formatting macro to its limit. Can we instead somehow expand the macro and see if it's parameters are used in `::core::fmt::Arguments::new_v1` or something like that?

vlad20012:
@Kobzol I'd follow Rustc's messages style. In the case of `println!("{}");` it places annotation to `{}`. In other cases - to an argument.

vlad20012:
Regarding logging... I hope in the future these special cases will not be needed because of stuff like #4640. These macros expand to std's `format_args!` and so eventually should be supported without any additional effort.

Kobzol:
> @Kobzol I'd follow Rustc's messages style. In the case of println!("{}"); it places annotation to {}. In other cases - to an argument.

Right, I'm now cross-checking the errors with rustc to make them as similar as possible. So far I have unified `{0}` and `{}` though, Rust has special error messages for the latter, but it seems to me that it would bring unnecessary complexity to the annotator. I'll write as many tests as I can come up with now and we'll see.

Kobzol:
Added more tests and refactored the annotator slightly.

Kobzol:
Rebased onto master to fix conflict.

Kobzol:
@vlad20012 Let me know if there are any other changes that I should make :)

vlad20012:
Looks like this implementation doesn't take into account width/precision arguments. So there are a lot of false-positives in e.g. "liballoc/fmt.rs" (`std::fmt`) examples:
image

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -13,124 +13,123 @@ import com.intellij.openapiext.isUnitTestMode
...
+            "X" -> knownItems.UpperHex
+            "p" -> knownItems.Pointer
+            "b" -> knownItems.Binary
+            "e" -> knownItems.LowerExp
+            "E" -> knownItems.UpperExp
+            else -> return null
+        }
     }
+}
+private class NamedParameter(matchInfo: ParameterMatchInfo, val name: String) : FormatParameter(matchInfo)
```
vlad20012, line 99: We usually put sealed class children into the parent sealed class. I.e. `FormatParameter.Named` and `FormatParameter.Positional`

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -13,124 +13,123 @@ import com.intellij.openapiext.isUnitTestMode
...
+private class PositionalParameter(matchInfo: ParameterMatchInfo, val position: Int) : FormatParameter(matchInfo)
+
+private data class FormatContext(val parameters: List<FormatParameter>,
+                                 val arguments: List<RsFormatMacroArg>,
+                                 val macro: RsMacroCall) {
+    val namedParameters: Set<NamedParameter> = parameters.filterIsInstance<NamedParameter>().toSet()
+    val positionalParameters: List<PositionalParameter> = parameters.filterIsInstance<PositionalParameter>().toList()
+
+    val namedArguments: Set<RsFormatMacroArg> = arguments.filter { it.name() != null }.toSet()
```
vlad20012, line 108: `Map<String, RsFormatMacroArg>` ?

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -13,124 +13,123 @@ import com.intellij.openapiext.isUnitTestMode
...
+
+    val knownItems: KnownItems = macro.knownItems
+}
+
+private data class ParsedParameter(val completeMatch: MatchResult,
+                                   val innerContentMatch: MatchResult? = null) {
+    val innerContent = completeMatch.groups[2]
+    val range: IntRange = completeMatch.range
```
vlad20012, line 116: Looks like `range` is mapped in all usages (`ctx.toSourceRange(parameter.range)`), so maybe it's possible to map it `toSourceRange` once?

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -140,5 +139,244 @@ class RsFormatMacroAnnotator : AnnotatorBase() {
...
+    val requiredTrait = parameter.requiredTrait(argument.knownItems) ?: return null
+
+    val expr = argument.expr
+    if (!expr.implLookup.canSelectWithDeref(TraitRef(expr.type, requiredTrait.withSubst()))) {
+        return ErrorAnnotation(argument.textRange, "`${expr.type.textWithAliasNames}` doesn't implement `${requiredTrait.name}` (required by ${parameter.matchInfo.match.value})")
     }
+    return null
 }
+
+private fun checkArgument(argument: RsFormatMacroArg,
```
vlad20012, line 326: formatting

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -140,5 +139,244 @@ class RsFormatMacroAnnotator : AnnotatorBase() {
...
+        val type = match.groups["type"]
+
+        if (match.value !in ignored) {
+            val typeStr = type?.value ?: ""
+            val matchInfo = ParameterMatchInfo(textRange, param.completeMatch, typeStr)
+            if (id != null) {
+                val identifier = id.value.toIntOrNull()
+                if (identifier == null) {
+                    NamedParameter(matchInfo, id.value)
+                } else PositionalParameter(matchInfo, identifier)
```
vlad20012, line 254: formatting

src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
Kobzol: It turned out that once I added the new width/precision specifier parsing, this method became more useful, so I kept it like this.

Kobzol:
Wow, I had no idea something like that is even in the `std::fmt` spec :) I added support for it, but here I chose to use a different error message than rustc, because its error is not very nice:
image

I added an error that says `Width specifier must be of type usize`/`Precision specifier must be of type usize`. There is however a slight type inference issue. Rust requires that the type of the width/precision specifier is `usize`.

This works in Rust:
```rust
println!("Hello {:1$}!", "x", 5);
println!("Hello {:1$}!", "x", 5usize);
```
but this sadly does not:
```rust
println!("Hello {:1$}!", "x", 5i32);
```

The problem is that type inference parses `5` as `i32`, so it fails in the plugin. I added a workaround which first checks if the type is `usize` and if not, it still allows integer literals without suffix. It doesn't work for something like `&5` though, which is allowed by Rustc, but that seems like an edge case. I can add support for unwrapping unary expressions (like `&`) or solving constant cases like `5+5`, but it seems like needless complexity at this point, for not a lot gain.

This seems to me like an issue that should be solved directly in Rust, I see no reason not to allow for example `u32` to be used instead of `usize`, I doubt that anyone would ever need more than 2^32 space characters or more than 2^32 places of precision. I'll try to send a PR to Rustc.

I also made error tracking more precise:
image
Now only if there is an implicit `{}` parameter the error will span the whole `{..}` range, otherwise only the specific named/positional parameter will be highlighted.

Kobzol:
Hmm, that's weird, build in CI failed, but it's fine locally. I'll try to rebase on master.

vlad20012:
Hmm, there is really an issue with type inference, and it's hard to fix... I don't like false-positives like this:
```rust
let a = 5;
println!("Hello {:1$}!", "x", a);
```
I suggest to just allow i32 (only) along with usize as a width/precision. 

> Hmm, that's weird, build in CI failed, but it's fine locally. I'll try to rebase on master.

Yeah, CI sometimes do rebase before tests run, so try to rebase and fix compilation

Kobzol:
Now both `usize` and `i32` are allowed and format trait errors are more precise. I noticed that when I mark the same text span as a format specifier and as an error, it will not highlight the span:
image
Is there any workaround for that or is it intended?

vlad20012:
> Is there any workaround for that or is it intended?

Hmm, I'd say it's not intended, but I can't suggest a workaround for now =/ I think we can fix it later.

Still, there are false-positives in "liballoc/fmt.rs":

image

Kobzol:
Thanks for pointing that out, I missed it. This is a very ugly part of `std::fmt`. The documentation about the precision asterisk is plainly wrong, but sadly it has not been fixed yet (https://github.com/rust-lang/rust/issues/32391). I have implemented it not according to the documentation but according to how Rustc actually behaves (each asterisk precision consumes an unnamed positional argument). It makes the logic convoluted, but hopefully it's the last bit of logic that was missing. There is one remaining false positive in `liballoc/fmt.rs`, but it seems to be related to type inference in doctests, as it works when the same snippet is used in non-documentation code.

vlad20012:
Ok, let's merge it! Please, squash commits that you think require squashing.

PR #4972, opened 2020-02-16
INT: don't generate useless match arm for irrefutable patterns in IfLetToMatchIntention

Kobzol:
This PR modifies `IfLetToMatchIntention` to avoid generating useless `_ => { ... }` arms if the match arm is already irrefutable. Before the generated code would right away contain a warning that the arms after the irrefutable arm can be removed.

It also removes the code from a potential else branch if the pattern is irrefutable, since the else was basically dead code.

### DIFF
--- a/src/test/kotlin/org/rust/ide/intentions/IfLetToMatchIntentionTest.kt
+++ b/src/test/kotlin/org/rust/ide/intentions/IfLetToMatchIntentionTest.kt
@@ -636,4 +636,50 @@ class IfLetToMatchIntentionTest : RsIntentionTestBase(IfLetToMatchIntention()) {
             }
         }
     """)
+
+    fun `test irrefutable pattern`() = doAvailableTest("""
+        struct Id(u32);
+        struct S { a: Id, b: u32 }
+        fn foo(s: S) {
+            if let S { a: Id(ref name), .. } = s/*caret*/ {
+                let _x = name;
+            }
+        }
+    """, """
+        struct Id(u32);
+        struct S { a: Id, b: u32 }
+        fn foo(s: S) {
+            match s {
+                S { a: Id(ref name), .. } => {
+                    let _x = name;
+                }
+            }
+        }
+    """)
+
+    fun `test irrefutable pattern with else`() = doAvailableTest("""
+        struct Id(u32);
+        struct S { a: Id, b: u32 }
+        fn foo(s: S) {
+            if let S { a: Id(ref name), .. } = s/*caret*/ {
+                let _x = name;
+            }
+            else {
+                let _y = 0;
+            }
+        }
+    """, """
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/rust/ide/intentions/IfLetToMatchIntentionTest.kt
```diff
@@ -636,4 +636,47 @@ class IfLetToMatchIntentionTest : RsIntentionTestBase(IfLetToMatchIntention()) {
...
+        fn foo(s: S) {
+            match s {
+                S { a: Id(ref name), .. } => {
+                    let _x = a;
+                }
+            }
+        }
+    """)
+
+    fun `test irrefutable pattern with else`() = doAvailableTest("""
```
mchernyavsky: Hmmm, I think we should keep else-branch here. @Undin what do you think?

src/test/kotlin/org/rust/ide/intentions/IfLetToMatchIntentionTest.kt
```diff
@@ -636,4 +636,47 @@ class IfLetToMatchIntentionTest : RsIntentionTestBase(IfLetToMatchIntention()) {
             }
         }
     """)
+
+    fun `test irrefutable pattern`() = doAvailableTest("""
+        struct Id(u32);
+        struct S { a: Id, b: u32 }
+        fn foo(s: S) {
+            if let S { a: Id(ref name), .. } = s/*caret*/ {
+                let _x = a;
```
Undin:
```suggestion
                let _x = name;
```

Otherwise, this code cannot be compiled because `a` here is not binding
Fix it please in other tests as well

src/test/kotlin/org/rust/ide/intentions/IfLetToMatchIntentionTest.kt
Undin: I agree that we should keep else branch if a user already has it

Undin: BTW, it would be great to have an inspection to annotate such redundant `else`

Kobzol:
Ok, I changed the `else` branch behaviour and rebased over master. Should I create (in a separate PR) a new `RsLocalInspectionTool` similar to `RsMissingElseInspection` (possibly with a quick fix to remove the `else` branch) or put the inspection elsewhere?

Undin:
> Should I create (in a separate PR) a new RsLocalInspectionTool similar to RsMissingElseInspection (possibly with a quick fix to remove the else branch) or put the inspection elsewhere?

@Kobzol I'd prefer to make it a separate inspection

Undin:
bors retry

PR #4975, opened 2020-02-16
PSI: introduce `escapedName` property

Undin:
It should simplify raw identifiers handling in code that creates new psi from text


Undin:
CC @Kobzol

Issue #4246, opened 2019-08-12
Extract constant


alexander-irbis:
This is a request for the "extract constant" refactoring feature.
The feature may include:

* [ ] "Extract literal as a local constant" action, similar to "extract variable" action.
* [ ] "Extract literal as a global constant" action.
* [ ] (?)"Replace all entries of the literal in the scope" option during extracting.
* [ ] Suggest to extract constant on literals.
* [ ] Suggest (with warning underline of the literal) to replace a literal with the corresponding constant if it is in the scope.
* [ ] (?)Suggest (with warning underline of the literal) to replace a literal with the corresponding constant, and to autoimport last if it may be imported

Kobzol:
https://github.com/intellij-rust/intellij-rust/pull/4985 has now been merged. @alexander-irbis can you check it out if it fulfills your needs? :)

PR #4985, opened 2020-02-17
REF: introduce constant refactoring

Kobzol:
In this PR I'd like to tackle the Introduce constant refactoring. The current state is just a sketch, because I have a few questions:

- Should this work on arbitrary expressions (and types) or just literals?
- Where should the corresponding constant be generated? Maybe we should show the user a short dialog (similar to the Introduce parameter refactoring) to let him select? For example here:

```rust
//x
mod x {
    //x
    fn foo() {
        //x
        fn bar() {
            //x
            let x = 5;
        }   
    }
}
```
We could offer these 4 locations (inner function, outer function, inside module, outside module).
- Should literals with the same type and value in the given context (function/block) be extracted to use the newly created constant (same as with the Introduce Variable refactoring)?

Related issue: https://github.com/intellij-rust/intellij-rust/issues/4246

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
@@ -1,0 +1,99 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.refactoring.introduceConstant
+
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiParserFacade
+import com.intellij.refactoring.RefactoringActionHandler
+import com.intellij.refactoring.RefactoringBundle
+import com.intellij.refactoring.util.CommonRefactoringUtil
+import org.rust.ide.inspections.import.RsImportHelper
+import org.rust.ide.refactoring.*
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.openapiext.runWriteCommandAction
+
+class RsIntroduceConstantHandler : RefactoringActionHandler {
+    override fun invoke(project: Project, editor: Editor, file: PsiFile, dataContext: DataContext) {
+        if (file !is RsFile) return
+        val exprs = findCandidateExpressionsToExtract(editor, file).filter { it.isExtractable() }
+
+        when (exprs.size) {
+            0 -> {
+                val message = RefactoringBundle.message(if (editor.selectionModel.hasSelection())
+                    "selected.block.should.represent.an.expression"
+                else
+                    "refactoring.introduce.selection.error"
+                )
+                val title = RefactoringBundle.message("introduce.constant.title")
+                val helpId = "refactoring.extractConstant"
+                CommonRefactoringUtil.showErrorHint(project, editor, message, title, helpId)
+            }
+            1 -> extractExpression(editor, exprs.single())
+            else -> {
+                showExpressionChooser(editor, exprs) {
+                    extractExpression(editor, it)
+                }
+            }
+        }
+    }
+
+    override fun invoke(project: Project, elements: Array<out PsiElement>, dataContext: DataContext?) {
+        //this doesn't get called from the editor.
+    }
+}
+
+private fun RsExpr.isExtractable(): Boolean {
+    return when (this) {
+        is RsLitExpr -> true
+        is RsBinaryExpr -> this.left.isExtractable() && (this.right?.isExtractable() ?: true)
+        else -> false
+    }
+}
+
+private fun replaceWithConstant(expr: RsExpr, occurrences: List<RsExpr>, candidate: InsertionCandidate, editor: Editor) {
+    val project = expr.project
+    val factory = RsPsiFactory(project)
+    val suggestedNames = expr.suggestedNames()
+    val name = suggestedNames.default.toUpperCase()
+    val const = factory.createConstant(name, expr)
+
+    val (insertedConstant, replaced) = project.runWriteCommandAction {
+        val newline = PsiParserFacade.SERVICE.getInstance(project).createWhiteSpaceFromText("\n")
+        val context = candidate.parent
+        val inserted = context.addBefore(const, candidate.anchor) as RsConstant
+        context.addAfter(newline, inserted)
+        val replaced = occurrences.map {
+            val created = factory.createExpression(name)
+            val element = it.replace(created) as RsElement
+            RsImportHelper.importElements(element, setOf(inserted))
+            element
+        }
+        Pair(inserted, replaced.toList())
+    }
+
+    editor.caretModel.moveToOffset(insertedConstant.identifier?.textRange?.startOffset
+        ?: error("Impossible because we just created a constant with a name"))
+
+    PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)
+    RsInPlaceVariableIntroducer(insertedConstant, editor, project, "Choose a constant name", replaced)
+        .performInplaceRefactoring(LinkedHashSet(suggestedNames.all.map { it.toUpperCase() }))
+}
+
+private fun extractExpression(editor: Editor, expr: RsExpr) {
+    if (!expr.isValid) return
+    val occurrences = findOccurrences(expr)
+    showOccurrencesChooser(editor, expr, occurrences) { occurrencesToReplace ->
+        showInsertionChooser(editor, expr) {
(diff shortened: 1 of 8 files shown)


vlad20012:
1. I think we can follow Java behavior. It work for expressions if they can be extracted to a constent. Otherwise it shows error message: 
image
2. Java implementation by default uses class level, but if there are multiple classes it shows choose dialog 
image
But I like how scope selection works in kotlin's `extract veriable`:
image
In addition to a chooser, it highlights the selection scope. Maybe it is overkill for constant extraction b/c scopes are usually large.
3. Java implementation offers a nice dialog with options =)
image

Kobzol:
Added expression selection, auto import and interactive renaming (those features should be independent of highlighting).
https://gifyu.com/image/76pa

For some reason this works in Clion, but not in tests:
```rust
let x = /*caret*/5 + 5;

const i: i32 = 5 + 5; /// IDE
const X1: _ = 5 + 5; /// tests
```
In tests the `BinaryExpr` is not inferred to be of type `i32` (it's `TyUnknown` instead).

Kobzol:
Rebased over master.

Kobzol:
I fixed the tests, they needed the stdlib `ProjectDescriptor`.

Here's how the refactoring currently looks like: https://gifyu.com/image/76pa. @vlad20012 let me know if there are any other changes that I should make.

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    val (inserted, replaced) = project.runWriteCommandAction {
+        val newline = PsiParserFacade.SERVICE.getInstance(project).createWhiteSpaceFromText("\n")
+        val context = candidate.parent
+        val inserted = context.addBefore(const, candidate.anchor)
+        context.addAfter(newline, inserted)
+        val replaced = occurrences.map {
+            val created = factory.createExpression(name)
+            val element = it.replace(created) as RsElement
+            AutoImportFix(element, AutoImportFix.Type.GENERAL_PATH).invoke(project)
```
vlad20012, line 78: Try to use `RsImportHelper.importElements`

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
```diff
@@ -0,0 +1,99 @@
...
+        val exprs = findCandidateExpressionsToExtract(editor, file).filter { it.isExtractable() }
+
+        when (exprs.size) {
+            0 -> {
+                val message = RefactoringBundle.message(if (editor.selectionModel.hasSelection())
+                    "selected.block.should.represent.an.expression"
+                else
+                    "refactoring.introduce.selection.error"
+                )
+                val title = RefactoringBundle.message("introduce.variable.title")
```
vlad20012, line 37: Looks like a wrong message

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
```diff
@@ -0,0 +1,99 @@
...
+}
+
+private fun replaceWithConstant(expr: RsExpr, occurrences: List<RsExpr>, candidate: InsertionCandidate, editor: Editor) {
+    val project = expr.project
+    val factory = RsPsiFactory(project)
+    val suggestedNames = expr.suggestedNames()
+    val name = suggestedNames.default.toUpperCase()
+    val const = factory.createConstant(name, expr)
+
+    val (inserted, replaced) = project.runWriteCommandAction {
```
vlad20012, line 70:
```suggestion
    val (insertedConstent, replaced) = project.runWriteCommandAction {
```
(just to make the code a bit easier to understand)

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
```diff
@@ -0,0 +1,99 @@
...
+        val replaced = occurrences.map {
+            val created = factory.createExpression(name)
+            val element = it.replace(created) as RsElement
+            AutoImportFix(element, AutoImportFix.Type.GENERAL_PATH).invoke(project)
+            element
+        }
+        Pair(inserted as RsConstant, replaced.toList())
+    }
+
+    editor.caretModel.moveToOffset(inserted.identifier?.textRange?.startOffset ?: 0)
```
vlad20012, line 84: Looks like ` ?: 0` is impossible, but it can leads to fun bugs if it will become possible somehow in the future. Maybe it's better to fail explicitly `?: error("Impossible because we just created a constant with a name")`

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/ui.kt
```diff
@@ -0,0 +1,157 @@
...
+    private val factory = RsPsiFactory(project)
+    private var highlighter: RangeHighlighter? = null
+    private var comment: PsiElement? = null
+    private val attributes = EditorColorsManager.getInstance().globalScheme.getAttributes(EditorColors.SEARCH_RESULT_ATTRIBUTES)
+
+    fun onSelect(candidate: InsertionCandidate) {
+        dropHighlighter()
+        val markupModel: MarkupModel = editor.markupModel
+
+//        val newComment = factory.createComment("<here>")
```
vlad20012, line 38: Well, let's use the usual IntelliJ UI for this for now and drop these comments.

Highlight a position with comments is an interesting idea. Something like inlay hints maybe even better :thinking:

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/ui.kt
```diff
@@ -0,0 +1,157 @@
...
+                    val candidate = value as InsertionCandidate
+                    val text = candidate.description()
+                    return super.getListCellRendererComponent(list, text, index, isSelected, cellHasFocus)
+                }
+            })
+            .setItemSelectedCallback { value: InsertionCandidate? ->
+                if (value == null) return@setItemSelectedCallback
+                highlighter.onSelect(value)
+            }
+            .setTitle("Insert ${expr.text} into:")
```
vlad20012, line 93: Maybe "Choose scope to introduce constant" (more close to java's version)
image

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/ui.kt
```diff
@@ -0,0 +1,157 @@
...
+        while (found.parent != parent) {
+            found = found.parent
+        }
+        return found
+    }
+
+    while (parent !is RsFile) {
+        parent = parent.parent
+        when (parent) {
+            is RsFunction -> {
```
vlad20012, line 133: There is a one corner case where this doesn't work: a module can be declared inside a function:
```rust
fn bar() {
    // <- constant should not be suggested here
    mod foo {
        fn foo() {
            let a = 2+2;
        }
    }
}
```
but if I declare a constant inside a function `bar`, I can't use it inside module `foo` :shrug:

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -360,6 +364,9 @@ class RsPsiFactory(
     fun createWhitespace(ws: String): PsiElement =
         PsiParserFacade.SERVICE.getInstance(project).createWhiteSpaceFromText(ws)
 
+    fun createComment(comment: String): PsiComment =
```
vlad20012, line 367: Not needed now

src/main/kotlin/org/rust/ide/refactoring/introduceConstant/RsIntroduceConstantHandler.kt
Kobzol: What should it say instead of "Selected block should represent an expression"? "Selected block should contain a constant expression"?

vlad20012: "Selected block should represent an expression" looks ok

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -56,16 +56,16 @@ class AutoImportFix(element: RsElement, private val type: Type) : LocalQuickFixO
     fun invoke(project: Project) {
         val element = startElement as? RsElement ?: return
         val (_, candidates) = when (type) {
-            GENERAL_PATH -> findApplicableContext(project, element as RsPath)
-            ASSOC_ITEM_PATH -> findApplicableContextForAssocItemPath(project, element as RsPath)
+            GENERAL_PATH -> findApplicableContext(project, getPath(element))
```
vlad20012, line 59: These changes are not needed after using RsImportHelper?

Kobzol: Right, thanks.

vlad20012:
Could you please squash these commits?

vlad20012:
:+1: Thanks! 
bors r+

PR #4993, opened 2020-02-18
Place the caret near a type when navigating to an impl

vlad20012:
I.e. place the caret near T in `impl T {}` when navigating to this impl (e.g. from `Self`)

```rust
impl /*caret_after*/Foo { // <- navigates here
    fn foo() -> /*caret_before*/Self {} // Ctrl + B
}
```

Note that this also works for "Go To Implementation" navigation from a trait:

```rust
trait T/*caret_before*/ { // Ctrl + Alt + B
    fn test(&self);
}
impl T for /*caret_after*/() { // <- navigates here
    fn test(&self) {}
}
```


mchernyavsky:
@vlad20012 please fix tests

PR #4994, opened 2020-02-19
GRAM: improve syntax error messages

vlad20012:
This improves syntax error messages like `expected <foo>, found <bar>`. Now they should be more short and readable


vlad20012:
Hmm, I'll also rename "pat" to "pattern"
bors r-

PR #5009, opened 2020-02-23
GRAM: fixed parsing of lifetime bound on Fn bound when it returns reference

rrevenantt:
In this case:
```rust
fn test<'b, F, Z: 'b>() where F: Fn() -> &'b Z + 'static {}
```
`'static` was parsed incorrectly as part of return type, while it should be part of `Fn` bound


src/main/grammars/RustParser.bnf
```diff
@@ -857,7 +857,9 @@ ArrayType ::= '[' TypeReference [';' AnyExpr] ']' {
   elementTypeFactory = "org.rust.lang.core.stubs.StubImplementationsKt.factory"
 }
 
-RefLikeType ::= ( '&' Lifetime? mut? | '*' [ const | mut ] ) TypeReference {
+private RefLikeInner ::= ( '&' Lifetime? mut? | '*' [ const | mut ] )
```
vlad20012: Let's rename to `RefLikeStart` like `PathStart`

Undin:
@rrevenantt in #5065 we've introduce `org.rust.lang.core.parser.RustParserDefinition#PARSER_VERSION` that's supposed to be increased after each change in grammar to drop the corresponding indices and avoid possible bugs because of it
Could you please bump it?

Undin:
bors r=vlad20012

Undin:
bors retry

PR #5010, opened 2020-02-23
REF: Introduce field refactoring

Kobzol:
This PR adds support for the IntroduceField refactoring.
The current version is just an initial prototype. First I want to get an idea how exactly should the refactoring work.

Questions:
- If the struct is empty like this: ```struct S;```, should we not offer the refactoring or offer it while letting the user choose whether he wants a block/tuple struct or offer it and choose some default (block/tuple)? Right now I decided not to offer it. It's simple for the user to add two characters (`{}` or `()`), while the possibility to choose would complicate the implementation quite a bit.
- Should the refactoring try to respect whether existing block fields are on separate lines or on a single line? Currently I use a simple heuristic (no fields => use newlines, last field has a newline => use newlines, otherwise no newlines).
- The UI will have to let the user choose the type of the fields. I haven't found a similar dialog in the plugin (the closest is Extract function, but there the type is read-only). Should we just let the user type the type (no pun intended) as a string and then try to parse it as `Ty/TypeReference` and infer it? It would probably be annoying to introduce new lifetimes/types like this, but that's maybe too much for this refactoring.
- Should the usage sites of the struct be changed to add the new fields with a default value?

TODO:
- [ ] import types from the added fields
- [ ] add lifetimes/type parameters to struct from the added fields?
- [ ] change usage sites of the struct
- [ ] create UI dialog

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRefactoringSupportProvider.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRefactoringSupportProvider.kt
@@ -9,6 +9,7 @@ import com.intellij.lang.refactoring.RefactoringSupportProvider
 import com.intellij.psi.PsiElement
 import com.intellij.refactoring.RefactoringActionHandler
 import org.rust.ide.refactoring.extractFunction.RsExtractFunctionHandler
+import org.rust.ide.refactoring.introduceField.RsIntroduceFieldHandler
 import org.rust.ide.refactoring.introduceParameter.RsIntroduceParameterHandler
 import org.rust.ide.refactoring.introduceVariable.RsIntroduceVariableHandler
 import org.rust.lang.core.macros.isExpandedFromMacro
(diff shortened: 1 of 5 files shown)


Kobzol:
@vlad20012 I started to implement this refactoring with the functionality of simply adding new fields to a struct:
```rust
struct S { a: u32 }/*caret*/ // user selects name, type, visibility of added fields
/// vvv
struct S { a: u32, b: u32 } // constructor and other usage sites are modified, b gets a default value
```

Now I realized that in other IntelliJ IDEs it actually works similarly as variable/constant extraction, i.e. an expression is extracted and added to a struct. In Rust it could work like this:
```rust
struct S { a: u32 }
impl S {
   fn new(a: u32) -> Self { S { a } }
   fn foo(&self) {
      let x = <selection>5</selection>;
   }
}
/// vvv
struct S { a: u32, x: u32 }
impl S {
   fn new(a: u32) -> Self { S { a, x: 5 } }
   fn foo(&self) {
      let x = self.x;
   }
}
```
Personally I'm not sure that I ever needed such transformation in Rust. It might be complicated to implement if the expression needs special lifetimes and/or type parameters. And it's also quite problematic if the expression is not self contained (i.e. needs other variables). In languages like Python this can be circumvented by creating the field directly inside the method, but that obviously does not work in Rust.

Now I'm not sure if this refactoring is even useful for Rust at all :) What do you think?

vlad20012:
Well, until now, I did not even know about the existence of such refactoring.
It looks like we really don't need this in Rust. May be useful for languages with classes and a more relaxed type systems...

Kobzol:
Right, I agree. Closing for now.

PR #5033, opened 2020-02-29
REF: use initialization shorthand after rename

### LINKED ISSUE #5028: Use shorthand initialisation after rename
Rename doesn't handle shorthands yet.

```rust
struct Foo {
    value: value
}

fn bar(foo: u32) -> Foo {
    Foo {
        value: foo
    }
}
```
`foo` gets renamed to `value`
```rust
struct Foo {
    value: value
}

fn bar(value: u32) -> Foo {
    Foo {
        value: value
    }
}
```
For sure not terrible, but may be better!
```rust
fn bar(value: u32) -> Foo {
    Foo {
        value
    }
}
```

t-kameyama:
Fixes #5028

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
@@ -10,6 +10,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Pass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.search.SearchScope
+import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.refactoring.listeners.RefactoringElementListener
 import com.intellij.refactoring.rename.RenameDialog
 import com.intellij.refactoring.rename.RenamePsiElementProcessor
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -32,16 +33,19 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
     override fun renameElement(element: PsiElement, newName: String, usages: Array<out UsageInfo>, listener: RefactoringElementListener?) {
         if (element is RsPatBinding) {
-            usages
-                .filter {
-                    val usageElement = it.element
-                    usageElement is RsStructLiteralField && usageElement.colon == null
+            usages.forEach {
+                val usageElement = it.element
+                if (usageElement is RsStructLiteralField && usageElement.colon == null) {
+                    val newPatField = RsPsiFactory(element.project).createStructLiteralField(element.text, newName)
```
Undin: Let's create single psi factory object before loop

src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt
```diff
@@ -519,6 +519,46 @@ class RenameTest : RsTestBase() {
         foo! { type T = Spam; }
     """)
 
+    fun `test use initialization shorthand after rename`() = doTest("value", """
+        struct Foo {
+            value: u32
+        }
+        fn bar(/*caret*/foo: u32) -> Foo {
+            Foo {
+                value: foo
```
Undin:
```suggestion
                value: /*comment*/ foo
```
To check that we keep comment here

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -32,16 +33,19 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
+                    usageElement.replace(newPatField)
+                    return@forEach
                 }
-                .forEach {
-                    val newPatField = RsPsiFactory(element.project)
-                        .createStructLiteralField(element.text, newName)
-                    it.element!!.replace(newPatField)
+                val field = PsiTreeUtil.getParentOfType(usageElement, RsStructLiteralField::class.java)
```
Undin: We usually use `ancestorStrict` or `ancestorOrSelf` instead of direct `PsiTreeUtil#getParentOfType`

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -32,16 +33,19 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
+                if (field?.identifier?.text == newName) {
```
Undin:
```suggestion
                if (field?.referenceName == newName) {
```

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
Undin: Unfortunately, it doesn't work in the following case:
```rust
struct Foo {
    value: u32,
}

fn bar(/*caret*/ bar: u32) -> Foo {
    Foo {
        value: foo(bar),
    }
}

fn foo(value: u32) -> u32 { unimplemented!() }
```
If you rename `bar` to `value`, you change produces 

```rust
fn bar(value: u32) -> Foo {
    Foo { value }
}
```
instead of

```rust
fn bar(value: u32) -> Foo {
    Foo { value: foo(value) }
}
```

We need to check if only a particular ancestor is `RsStructLiteralField` instead of searching any ancestor of this type

src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
```diff
@@ -31,22 +32,25 @@ class RsRenameProcessor : RenamePsiElementProcessor() {
...
-                .forEach {
-                    val newPatField = RsPsiFactory(element.project)
-                        .createStructLiteralField(element.text, newName)
-                    it.element!!.replace(newPatField)
+                val field = usageElement?.ancestorStrict<RsStructLiteralField>()
+                if (field?.expr is RsPathExpr && field.referenceName == newName ) {
+                    field.expr?.delete()
+                    field.colon?.delete()
                 }
+            }
```
Undin: I think this code may be written in the following way
```kotlin
val field = it.element?.ancestorOrSelf<RsStructLiteralField>(RsBlock::class.java) ?: return@forEach
when {
    field.colon == null -> {
        val newPatField = psiFactory.createStructLiteralField(element.text, newName)
        field.replace(newPatField)
    }
    field.referenceName == newName && field.expr is RsPathExpr -> {
        field.expr?.delete()
        field.colon?.delete()
    }
}
```

PR #5050, opened 2020-03-05
GRAM: better error recovery in some cases

### LINKED ISSUE #2435: Bad parser recovery for reference type
image

Probably a `pin` is needed on `*` and `&`: https://github.com/intellij-rust/intellij-rust/blob/a562faf65c08349ecdd6e289061ae4b3df57a395/src/main/grammars/RustParser.bnf#L743

vlad20012:
1. Get rid of `TupleOrParenExpr` PSI node. Now incomplete expressions like `let a = (` are parsed as an (incomplete) parentheses expr.
2. Get rid of `ExprStmtOrLastExpr` PSI node (really it was never created).
3. Parse a statement without a semicolon as a statement (if not last in a block). Previously it was wrongly parsed as an expr.
```rust
2 + 2 // now it's a statement with missing semicolon
let ;
```
3. More pins for types and bounds. E.g. now type of parameter `a` in `fn foo(a: &)` is parsed as a "reference type". Fixes #2435
4. Parse type arguments in any order (e.g. lifetime after a type: `foo::<i32, 'a>::bar()`). This provides betters error messages (e.g. "Lifetime arguments must be declared prior to type arguments").
5. Parse any items inside trait or impl bodies and `extern {}` blocks. This provides betters error messages (e.g. "Struct is not allowed inside trait") and also unifies our items grammar (helps to avoid bugs like #3366)


src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -217,6 +252,40 @@ private fun checkTypeParameterList(holder: AnnotationHolder, element: RsTypePara
...
+    val argumentNameCapitalized: String
+        get() = StringUtil.capitalize(argumentName)
+
+    fun canStandAfter(prevArgument: TypeArgumentKind): Boolean =
+        ordinal >= prevArgument.ordinal
+
+    data class Result(val newState: TypeArgumentKind, val allowed: Boolean)
+
+    companion object {
+        private val VALUES = values()
```
Undin: Why do we need this constant?

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -217,6 +252,40 @@ private fun checkTypeParameterList(holder: AnnotationHolder, element: RsTypePara
...
+    CONST(RsExpr::class, "const arguments"),
+    ASSOC(RsAssocTypeBinding::class, "associated type bindings");
+
+    val argumentNameCapitalized: String
+        get() = StringUtil.capitalize(argumentName)
+
+    fun canStandAfter(prevArgument: TypeArgumentKind): Boolean =
+        ordinal >= prevArgument.ordinal
+
+    data class Result(val newState: TypeArgumentKind, val allowed: Boolean)
```
Undin: looks like unused

src/main/kotlin/org/rust/lang/core/psi/ext/RsItemElement.kt
```diff
@@ -23,3 +26,50 @@ fun <T : RsItemElement> Iterable<T>.filterInScope(scope: RsElement): List<T> {
...
+        is RsTypeAlias -> "type alias"
+        is RsImplItem -> "impl"
+        is RsUseItem -> "use item"
+        is RsForeignModItem -> "foreign module"
+        is RsExternCrateItem -> "extern crate"
+        is RsMacro2 -> "macro"
+        else -> "item"
+    }
+
+val RsItemElement.itemKindNameCapitalized: String
```
Undin: Maybe it's better to use  `itemKindName.capitalize()` instead of new quite specific property?

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
vlad20012: `values()` allocates a copy of the array on each invocation...

PR #5053, opened 2020-03-06
REF: Fix introduce variable when first expression is inside nested block

dima74:
Fix that when extracting `7` to variable from this code:
```rust
let x1 = if true { 7 } else { 0 };
let x2 = 7;
```

it was transformed to 
```rust
let x1 = if true {
    let i = 7;
    i
} else { 0 };
let x2 = i;
```

while correct result would be
```rust
let i = 7;
let x1 = if true { i } else { 0 };
let x2 = i;
```

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
@@ -10,6 +10,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiDocumentManager
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiParserFacade
+import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.refactoring.*
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.ancestorOrSelf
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
```diff
@@ -127,3 +129,13 @@ private fun findAnchor(expr: PsiElement): PsiElement? {
...
 }
+
+private fun findAnchor(exprs: List<PsiElement>, chosenExpr: RsExpr): PsiElement? {
+    var anchor = findAnchor(exprs.minBy { it.startOffset } ?: chosenExpr)
+        ?: return null
+
+    while (exprs.any { !PsiTreeUtil.isAncestor(anchor.parent, it, true) }) {
+        anchor = anchor.parent
+    }
+    return anchor
```
Undin: Looks like the following version does the same but faster
```kotlin
val parent = PsiTreeUtil.findCommonParent(anchor.parent, *exprs.toTypedArray())
while (anchor.parent != parent) {
    anchor = anchor.parent
}
```

Also, maybe it's worth to extract `while` loop into separate function because we have the same code in `findAnchor` above

dima74: Yes, thank you, updated

Undin:
@mchernyavsky Looks like it's bors bug. The same was with #5057
@dima74 Could you rebase onto master, please? Last time it helped

dima74:
@Undin yes, rebased

PR #5064, opened 2020-03-08
PERF&GRAM: parser optimizations

vlad20012:
These optimizations speed up parsing by 20%


src/main/grammars/RustParser.bnf
```diff
@@ -688,9 +697,8 @@ private TupleStructTail ::= TupleFields WhereClause? ';' { pin = 1 }
 private UnitStructTail ::= WhereClause? ';'
 private BlockStructTail ::= WhereClause? BlockFields
 
-EnumItem ::= AttrsAndVis enum identifier TypeParameterList? WhereClause? EnumBody {
+upper EnumItem ::= enum identifier TypeParameterList? WhereClause? EnumBody {
```
Undin: Let's add `name` field here as well

PR #5065, opened 2020-03-08
Provide missing versions for indexers

Undin:
It should update the corresponding indices after lexer or parser changes.
Also, introduce separate lexer, parser and stub versions


vlad20012:
:+1: 
bors r+

Issue #5058, opened 2020-03-07
Rust REPL output contains "garbage"


CDirkx:
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.2170-193
* **Rust toolchain version:** Rust 1.43 nightly
* **IDE name and version:** CLion 2019.3.4 193.6494.38
* **Operating system:** Windows 10

## Problem description
Using evcxr.exe (Rust REPL) with `--disable-readline` causes the output to include "garbage".

```
Welcome to evcxr. For help, type :help
╔?[33m>> ?[0m>> 1 + 2
3
╔?[33m>> ?[0m
>>
```

```
Welcome to evcxr. For help, type :help
╔?[33m>> ?[0m>> a
   ?[91m^?[0m ?[94mnot found in this scope?[0m
?[91mcannot find value `a` in this scope?[0m
╗?[33m>> ?[0m
>>
```

## Steps to reproduce
The problem does not happen when using evcxr.exe standalone without `--disable-readline`.

The problem does happen when using evcxr.exe standalone with `--disable-readline` and in the editor Rust REPL window, as there `--disable-readline` is [always set](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/cargo/toolchain/Evcxr.kt).

## Resolution
I could not find what the `--disable-readline` option does exactly, and could thus not determine if this is:

- a bug in evcxr.exe itself
  - raise bug report at evcxr
  - Rust REPL `--disable-readline` default option is correct
- intended behaviour
  - Rust REPL `--disable-readline` default option is not correct (at least on my system/platform/configuration)

Undin:
@dima74 Looks like the problem is related to console escape symbols in output that windows console can't process

dima74:
By default evcxr uses [rustyline](https://github.com/kkawakam/rustyline) to read input, but rustyline uses escape sequences to turn terminal into special interactive mode, so `--disable-readline` [was added](https://github.com/google/evcxr/pull/68) to read input directly

Evcxr uses colored crate, and there is [similar report](https://github.com/mackwic/colored/issues/59), which suggests using `colored::control::set_virtual_terminal(true);`. I will check evcxr with this line added

Issue #5046, opened 2020-03-04
Rust REPL - Previous Command is still running


smj389:
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.2156-201
* **Rust toolchain version:** 1.43.0-nightly (4ad624882 2020-03-03) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2020.1 EAP (CL-201.5616.31)
* **Operating system:** Windows 10 10.0

## Problem description
In the Rust REPL, after you enter one command in the RUST Repl, you cannot enter any other commands as "Previous Command is still running. Please wait or press Ctrl+C in console to interrupt."

## Steps to reproduce
1. Start REPL
2. Type :help
3. After the help text has appeared and the prompt reappears enter :help again and then press enter.
4. Error Message appears.

            
![image](https://user-images.githubusercontent.com/12642064/75886590-5ff3d080-5e20-11ea-8670-1d3d812a0f67.png)

Issue #4883, opened 2020-01-27
Rust console evcxr on Windows failing


CGMossa:
The issue is, the console only works with one single statement, and then it gets
busy.

![image](https://user-images.githubusercontent.com/1063624/73169495-7d849a80-40fc-11ea-953a-9ddbf65e1934.png)

## Environment

* **IntelliJ Rust plugin version:** Latest; 0.2.114.2151-193
* **Rust toolchain version:** Both tested on nightly and 
* **IDE name and version:** IntelliJ IDEA 2019.3.2 (Community Edition)
Build #IC-193.6015.39, built on January 21, 2020
Runtime version: 11.0.5+10-b520.30 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Windows 10 10.0
GC: ParNew, ConcurrentMarkSweep
Memory: 1998M
Cores: 6
Registry: 
Non-Bundled Plugins: com.jetbrains.ChooseRuntime, izhangzhihao.rainbow.brackets, org.toml.lang, org.rust.lang
* **Operating system:** Windows 10, 1803

## Steps to reproduce
Turn on Rust Console. Enter one expr/statement. Enter another. It does not work.

PR #5066, opened 2020-03-08
REPL: Fix for windows (use new escape characters)

dima74:
There are two problems with rust console on windows:

1. Colors not displayed correctly (#5058, #5046, #4883). It is related to `colored` crate (used by Evcxr) and will be fixed by adding `colored::control::set_virtual_terminal(true);` to Evcxr code
2. It is possible to run only one command (#5046, #4883). This is because we use `\u0001` and `\u0002` characters as special markers to determine whether the last command was successful. For some reason under windows these characters changes into some other characters. So we will use new marker characters `\u0091` and `\u0092`

### DIFF
--- a/src/main/kotlin/org/rust/ide/console/RsConsoleCommunication.kt
+++ b/src/main/kotlin/org/rust/ide/console/RsConsoleCommunication.kt
@@ -54,7 +54,16 @@ class RsConsoleCommunication(private val consoleView: RsConsoleView) {
     }
 
     companion object {
-        const val SUCCESS_EXECUTION_MARKER: String = "\u0001"
-        const val FAILED_EXECUTION_MARKER: String = "\u0002"
+        /**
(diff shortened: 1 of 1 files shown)


Undin:
@dima74 Do I understand correctly, that you are going to make the corresponding fix in `Evcxr`?

dima74:
@Undin Yes, corresponding commit is [here](https://github.com/dima74/evcxr/commit/062a87da90f4ac0bd6503af9a325966a24ac042f). I will make PR to Evcxr after next IntelliJ Rust release (otherwise user may update evcxr and rust console stop working)

Undin:
@dima74 Is it true that the corresponding PR to evcxr is not created yet?

dima74:
@Undin Yes, PR to evcxr is not created yet

dima74:
Here is PR to evcxr: https://github.com/google/evcxr/pull/111

Issue #5070, opened 2020-03-09
Incorrect terminal size


ser-vasilich:
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.2170-193
* **Rust toolchain version:** 1.43.0-nightly (a1912f2e8 2020-02-12) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2019.3.4 (CL-193.6494.38)
* **Operating system:** Linux 5.3.0-40-generic

## Problem description

Clion terminal have not correct dimentions.
Next code show size of terminal.
It work with different terminals, but not with JetBrains.
I use two crates. They work same.

## Steps to reproduce

```rust
use term_size;
use terminal_size::{Width, Height, terminal_size};

fn main() {

    if let Some((w, h)) = term_size::dimensions() {
        println!("Width: {}\nHeight: {}", w, h);
    } else {
        println!("Unable to get term size :(")
    }


    let size = terminal_size();
    if let Some((Width(w), Height(h))) = size {
        println!("Your terminal is {} cols wide and {} lines tall", w, h);
     } else {
         println!("Unable to get terminal size");
     }

}
```

ser-vasilich:
With any size Clion terminal return one dimension:  2498 25 
I think issue early closed.

ser-vasilich:
For debugging run another terminal. With that terminal, I can't get dimension.

PR #5076, opened 2020-03-10
RUN: Set TERM variable and terminal's width if terminal emulation is enabled

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5070.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
@@ -19,6 +19,7 @@ import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Cargo.kt
```diff
@@ -360,7 +361,12 @@ class Cargo(private val cargoExecutable: Path) {
             }
             environmentVariables.configureCommandLine(cmdLine, true)
             return if (emulateTerminal) {
-                PtyCommandLine(cmdLine).withConsoleMode(false)
+                if (!SystemInfo.isWindows) {
+                    cmdLine.environment["TERM"] = "xterm-256color"
+                }
+                PtyCommandLine(cmdLine)
+                    .withConsoleMode(true)
```
Undin, line 368: Looks like it's default option
Maybe we drop this line?

mchernyavsky:
bors r=Undin

PR #5082, opened 2020-03-11
MACRO: Process nested macros in `descendantsWithMacrosOfType`

### LINKED ISSUE #4995: False positive "Variable is never used"
## Environment

* **IntelliJ Rust plugin version:** 0.2.116.2163-193
* **Rust toolchain version:** Rust 1.41.0
* **IDE name and version:** CLion 2019.3.4
* **Operating system:** OSX

## Problem description
False positive "Variable is never used" warning with declarative macros.
## Steps to reproduce
The IDE will give the warning and grey out `bar` in the following code:
```rust
fn foo() -> TokenStream {
    let bar = "";
    quote!(bar)
}
```

ortem:
Before now, the visitor did not go down the nested macro calls due to the implementation of `PsiRecursiveElementWalkingVisitor`, which processes new elements only after finish processing older ones (see `com.intellij.util.WalkingState#elementStarted`).

Also, I've got rid of abstract `RsWithMacroExpansionsRecursiveElementWalkingVisitor` and extract one of the `visitElement` implementations into the new private method to increase readability.

Fixes #4995 issue because now liveness analysis knows about unexpanded nested macros.


vlad20012:
:+1: 
bors r+

PR #4936, opened 2020-02-07
RES: support impls for aliased types

vlad20012:
Fixes #2778

```rust
struct X;
type Y = X;
impl Y { fn foo(&self) {} }
          //X
fn main() {
    X.foo();
}   //^
```

PR #5085, opened 2020-03-11
PERF&STUB: store `cfg` existence flags to stubs

vlad20012:
This speeds up `cfg` attributes processing, which in turn speeds up name resolution & type inference
Required for #4936 to avoid performance regression

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -208,9 +208,6 @@ val RsDocAndAttributeOwner.isCfgUnknown: Boolean
     get() = evaluateCfg() == ThreeValuedLogic.Unknown
 
 private fun RsDocAndAttributeOwner.evaluateCfg(): ThreeValuedLogic {
-    // TODO: get rid of this registry key when cfg support is done
-    if (!isUnitTestMode && !isFeatureEnabled(RsExperiments.CFG_ATTRIBUTES_SUPPORT)) return ThreeValuedLogic.True
-
     // TODO: add cfg to RsFile's stub and remove this line
     if (this is RsFile) return ThreeValuedLogic.True
 
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -5,14 +5,13 @@
 
 package org.rust.lang.core.psi.ext
 
-import com.intellij.openapiext.isUnitTestMode
+import com.intellij.extapi.psi.StubBasedPsiElementBase
 import com.intellij.psi.NavigatablePsiElement
 import com.intellij.psi.PsiElement
-import org.rust.ide.experiments.RsExperiments
 import org.rust.lang.core.psi.*
+import org.rust.lang.core.stubs.RsAttributeOwnerStub
 import org.rust.lang.utils.evaluation.CfgEvaluator
 import org.rust.lang.utils.evaluation.ThreeValuedLogic
-import org.rust.openapiext.isFeatureEnabled
 
 interface RsDocAndAttributeOwner : RsElement, NavigatablePsiElement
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/rust/lang/core/stubs/StubInterfaces.kt
```diff
@@ -5,6 +5,55 @@
...
+        fun extractFlags(element: RsDocAndAttributeOwner): Int =
+            extractFlags(element.queryAttributes)
+
+        fun extractFlags(attrs: QueryAttributes): Int {
+            var hasAttrs = false
+            var hasCfg = false
+            var hasMacroUse = false
+            for (meta in attrs.metaItems) {
+                hasAttrs = true
+                when (meta.name) {
```
Undin: Let's add some `TODO` here to check `cfg_attr` in future as well here

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -70,36 +69,38 @@ data class Attribute(val name: String, val argText: String? = null) {
     val text: String get() = if (argText == null) name else "$name($argText)"
 }
 
+inline val RsDocAndAttributeOwner.docAndAttrStub: RsDocAndAttributeOwnerStub?
```
Undin:
```suggestion
inline val RsDocAndAttributeOwner.attributeStub: RsDocAndAttributeOwnerStub?
```

src/main/kotlin/org/rust/lang/core/stubs/StubInterfaces.kt
```diff
@@ -5,6 +5,55 @@
...
 }
+
+interface RsDocAndAttributeOwnerStub {
+    // These properties are stored in stubs for performance reasons: it's much cheaper to check
+    // a flag in a stub then traverse a PSI
+
+    val hasAttrs: Boolean
+
+    // #[cfg()]
+    val hasCfg: Boolean
```
Undin: Probably, we need a slightly better name because of `cfg_attr` to mark these properties as approximate because we can't evaluate `cfg_attr` while indexing. Or, we need some documentation about it

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -135,7 +135,10 @@ val RsFunction.isCustomDeriveProcMacroDef: Boolean
     get() = queryAttributes.hasAttribute("proc_macro_derive")
 
 val RsFunction.isProcMacroDef: Boolean
-    get() = queryAttributes.hasAnyOfOuterAttributes(
+    get() = greenStub?.isProcMacroDef ?: (queryAttributes.isProcMacroDef)
+
+val QueryAttributes.isProcMacroDef
```
Undin: Do we really need this as separate public property?
Maybe it's worth to unify it with other similar properties and inline it into `RsFunction.isProcMacroDef`?

src/main/kotlin/org/rust/lang/core/stubs/StubInterfaces.kt
```diff
@@ -5,6 +5,55 @@
...
+import org.rust.lang.core.psi.ext.queryAttributes
+import org.rust.stdext.makeBitMask
+
 interface RsNamedStub {
     val name: String?
 }
+
+interface RsDocAndAttributeOwnerStub {
+    // These properties are stored in stubs for performance reasons: it's much cheaper to check
+    // a flag in a stub then traverse a PSI
```
Undin: Let's convert it into doc comment

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
vlad20012: This and similar properties specially separated from RsFunction because QueryAttributes is constructed differently during stub building

src/main/kotlin/org/rust/lang/core/stubs/StubInterfaces.kt
vlad20012: I think we can care about during `cfg_attr` implementation. There are at least 2 different ways how we can support then and chosen way can affect names & documentation

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
vlad20012: This and similar properties specially separated from RsFunction because QueryAttributes is constructed differently during stub building

src/main/kotlin/org/rust/lang/core/stubs/StubInterfaces.kt
Undin: okay

Undin:
bors retry

Issue #5027, opened 2020-02-28
Replace or add return type fix


sirgl:
Plugin already has typechecker, but it still doesn't suggest to fix return type. I think that quick fix should be created, as it is one of the most used quick fix in other IDEs.

Example:
```rust
fn bar() -> u32 {
    return "asdasd"
}
```
After applying fix:
```rust
fn bar() -> &'static str {
    return "asdasd"
}

PR #5087, opened 2020-03-12
INT: Add "Change return type" fix

mchernyavsky:
Relates to https://github.com/intellij-rust/intellij-rust/issues/5027.

Applicable to functions, non-inherited methods, and closures.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -352,11 +352,13 @@ class RsPsiFactory(
         createFromText<RsFunction>("unsafe fn foo(){}")?.unsafe
             ?: error("Failed to create unsafe element")
 
-    fun createFunction(
-        text: String
-    ): RsFunction =
+    fun createFunction(text: String): RsFunction =
         createFromText(text)
-            ?: error("Failed to create function element: text")
+            ?: error("Failed to create function element: $text")
+
+    fun createRetType(ty: String): RsRetType =
+        createFromText("fn foo() -> $ty {}")
+            ?: error("Failed to create function return type: $ty")
 
     fun createImpl(name: String, functions: List<RsFunction>): RsImplItem =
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rust/ide/inspections/typecheck/ChangeReturnTypeFixTest.kt
```diff
@@ -0,0 +1,88 @@
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.typecheck
+
+import org.rust.ide.inspections.RsInspectionsTestBase
+import org.rust.ide.inspections.RsTypeCheckInspection
+
+class ChangeReturnTypeFixTest : RsInspectionsTestBase(RsTypeCheckInspection::class) {
```
Undin: Could you please add tests with type alias to ensure that we support them

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -352,11 +352,13 @@ class RsPsiFactory(
...
 
-    fun createFunction(
-        text: String
-    ): RsFunction =
+    fun createFunction(text: String): RsFunction =
         createFromText(text)
-            ?: error("Failed to create function element: text")
+            ?: error("Failed to create function element: $text")
+
+    fun createRetType(ty: Ty): RsRetType =
```
Undin: Maybe it's better to make this API more generic and use `String` instead of `Ty`?

mchernyavsky:
Fixed.

mchernyavsky:
@Undin 
> Complex returning type

Maybe it's better to fix this in a separate PR? It also can improve error messages and `Type Info` action. I can create an issue.

Undin:
> Maybe it's better to fix this in a separate PR? It also can improve error messages and Type Info action. I can create an issue.

👍

mchernyavsky:
Fixed.

mchernyavsky:
Fixed.

PR #5096, opened 2020-03-13
Fix possible crash during components save

vlad20012:
The bug was introduced in #4872

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
@@ -323,12 +323,11 @@ private object MacroExpansionFileSystemRootsLoader {
     }
 
     fun saveProjectDirs() {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
```diff
@@ -380,7 +379,11 @@ private class MacroExpansionServiceImplInner(
         Files.createDirectories(dataFile.parent)
         dataFile.newDeflaterDataOutputStream().use { data ->
             ExpandedMacroStorage.saveStorage(storage, data)
-            MacroExpansionFileSystem.writeFSItem(data, MacroExpansionFileSystem.getInstance().getDirectory(dirs.expansionDirPath))
```
vlad20012: `FileNotFoundException` was thrown here. 
I don't know the root cause of this... I hope we'll find it later

PR #5099, opened 2020-03-16
GRAM: parse path with semicolon in attributes as RsPath

### LINKED ISSUE #4481: Goto Definition doesn't work for path attribute macros
## Environment

* **IntelliJ Rust plugin:**
0.2.106.2135-192 (Sep 30, 2019)
* **Rust toolchain:**
beta-x86_64-pc-windows-gnu (default)
rustc 1.39.0-beta.1 (968967007 2019-09-24)
* **IDE:**
IntelliJ IDEA 2019.2.2 (Ultimate Edition)
Build #IU-192.6603.28, built on September 6, 2019
Runtime version: 11.0.3+12-b304.56 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Windows 10 10.0
Memory: 1971M
Cores: 8
Registry: 
Non-Bundled Plugins: [...], org.toml.lang, org.rust.lang

## Problem description

Attribute macros using a path (e.g. `#[lib::foo]`) do not work with Goto Definition. Goto Definition works fine when the macro is imported and used without qualification (e.g. `#[foo]`).

## Steps to reproduce

```rust
// name = "foo"
// lib.proc-macro = true
extern crate proc_macro;
use proc_macro::TokenStream;
#[proc_macro_attribute]
pub fn foo(attr: TokenStream, item: TokenStream) -> TokenStream { item }
```
```rust
// name = "consumer"
// dependencies.foo.path = "path/to/foo"
use lib; use lib::foo;
#[lib::foo] fn does_not_work() {}
#[foo] fn this_does_work() {}
```

### LINKED ISSUE #4850: False positive on "Cannot move" inspection on a type with derived Copy
## Environment

* **Rust toolchain version:**
* **IDE name and version:**
* **Operating system:**

## Problem description

From https://github.com/intellij-rust/intellij-rust/issues/3101#issuecomment-576413621:

> I just noticed this still happens when you refer to derive-trait by a fully specified name. It's enough that one of the types is in its full form and then none work.

Minimal example (which, to be honest, is a little ridiculous):
```rust
#[derive(Copy, std::marker::Clone)]
struct X;
```
and a slightly more real-world example:
```rust
#[derive(prost::Message, Copy, Clone)]
struct Message {
  payload: u64,
}
```

Undin:
Previously, we always parse `#[foo::bar]` as token stream. As a result, if we have a complex path in an attribute like `#[derive(prost::Message, Copy, Clone)]`, we parse all items inside `derive` as token stream and don't know about `Copy` and `Clone`. Also, it prevented us to resolve complex paths to the corresponding procedural macros

Now it's parsed as `RsPath`, i.e. we have proper PSI to provide name resolution for described cases

Fixes #4850
Partially solves #4481
Need for #3855


src/main/kotlin/org/rust/lang/core/resolve/ref/RsAttributeProcMacroReferenceImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiElementResolveResult
+import com.intellij.psi.ResolveResult
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.resolve.collectResolveVariants
+import org.rust.lang.core.resolve.processAttributeProcMacroResolveVariants
+
+class RsAttributeProcMacroReferenceImpl(element: RsPath) : RsPathReferenceBase(element) {
```
vlad20012: Why we can't use `RsReferenceCached` here? Looks like even with `referenceAnchor` and `advancedResolve` implemented twice it brings less copy-pasted code.

I made it differently with `RsMacroPathReferenceImpl` because I needed to cache a custom resolve result.

src/main/kotlin/org/rust/lang/core/psi/ext/RsReferenceElement.kt
```diff
@@ -23,14 +23,14 @@ interface RsReferenceElementBase : RsElement {
...
-interface RsWeakReferenceElement : RsReferenceElementBase {
+interface RsReferenceElement : RsReferenceElementBase {
     override fun getReference(): RsReference?
 }
 
 /**
  * Marks an element that has a reference.
  */
-interface RsReferenceElement : RsWeakReferenceElement {
+interface RsMandatoryReferenceElement : RsReferenceElement {
```
vlad20012: 👏👏👏

src/main/kotlin/org/rust/lang/core/resolve/ref/RsAttributeProcMacroReferenceImpl.kt
Undin: Oh, thanks!
I've completely forgotten about it.
I've also made default implementation of `referenceAnchor` in `RsReferenceBase` so now `RsAttributeProcMacroReferenceImpl` impl is pretty minimalistic

src/main/kotlin/org/rust/lang/core/psi/ext/RsMetaItem.kt
```diff
@@ -19,7 +19,13 @@ import org.rust.lang.core.stubs.RsMetaItemStub
 
 val RsMetaItem.name: String? get() {
     val stub = greenStub
-    return if (stub != null) stub.name else identifier?.unescapedText
+    return if (stub != null) {
+        stub.name
```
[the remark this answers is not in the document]
Undin: Just artifact of the previous implementation

vlad20012:
:+1: 
bors r+

PR #5101, opened 2020-03-16
PERF&MACRO: use refresh-based VFS batch

vlad20012:
Previously, we didn't use the platform's refresh because of too large UI freezes. Since 2019.3 a part of the refresh process was extracted from EDT and now freezes duration seems acceptable. This speeds up macro expansion a bit

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionVfsBatch.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionVfsBatch.kt
@@ -5,15 +5,11 @@
 
 package org.rust.lang.core.macros
 
-import com.intellij.openapi.util.io.FileSystemUtil
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.openapi.vfs.VfsUtil
 import com.intellij.openapi.vfs.VirtualFile
-import com.intellij.openapi.vfs.VirtualFileManager
-import com.intellij.openapi.vfs.ex.VirtualFileManagerEx
-import com.intellij.openapi.vfs.newvfs.events.*
-import com.intellij.openapi.vfs.newvfs.persistent.PersistentFS
-import org.rust.openapiext.checkWriteAccessAllowed
+import com.intellij.openapi.vfs.newvfs.RefreshQueue
 import org.rust.stdext.randomLowercaseAlphabetic
-import java.util.*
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionVfsBatch.kt
```diff
@@ -41,154 +37,71 @@ class MacroExpansionVfsBatchImpl(rootDirName: String) : MacroExpansionVfsBatch {
...
-                        iter.remove()
-                        events.add(event)
-                    }
-                }
-                check(events.isNotEmpty())
-                PersistentFS.getInstance().processEvents(events)
-                events.clear()
+        for (path in pathsToMarkDirty) {
+            var file = root
+            for (segment in StringUtil.split(path, "/")) {
```
Undin:
```suggestion
            for (segment in path.split("/")) {
```

src/test/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystemTest.kt
```diff
@@ -9,46 +9,55 @@ import com.intellij.openapi.vfs.VfsUtil
...
+    override fun tearDown() {
+        super.tearDown()
+        checkMacroExpansionFileSystemAfterTest()
+    }
 }
 
 private fun VirtualFileSystem.findNonNullFileByPath(path: String): VirtualFile =
     findFileByPath(path) ?: error("File not found: $path")
 
 private val VirtualFileSystem.root: VirtualFile get() = findNonNullFileByPath("/")
```
Undin: Unused now?

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionVfsBatch.kt
vlad20012: Hmm, looks like `path.split('/')` and `StringUtil.split(path, "/")` works differently: `path.split('/')` produces empty strings.

Undin: Oh, it's interesting difference
Ok, let's revert the corresponding changes then

vlad20012:
bors r=undin

Issue #5110, opened 2020-03-17
Exception after project opening while Cargo toolwindow initialization


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.119.SNAPSHOT-193-dev
* **Rust toolchain version:** 1.31.0 (abe02cefd 2018-12-04) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2019.3.3 Ultimate Edition (IU-193.6494.35)
* **Operating system:** macOS 10.13.6

## Problem description

<details>

```
failed to init toolwindow org.rust.cargo.project.toolwindow.CargoToolWindowFactory

java.lang.IllegalArgumentException: window with id="Cargo" is already registered
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.registerToolWindow(ToolWindowManagerImpl.java:1091)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.initToolWindow(ToolWindowManagerImpl.java:474)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl$6.run(ToolWindowManagerImpl.java:453)
	at com.intellij.openapi.wm.impl.CommandProcessor.runNext(CommandProcessor.java:78)
	at com.intellij.openapi.wm.impl.CommandProcessor.run(CommandProcessor.java:59)
	at com.intellij.openapi.wm.impl.commands.FinalizableCommand.finish(FinalizableCommand.java:36)
	at com.intellij.openapi.wm.impl.commands.ApplyWindowInfoCmd.run(ApplyWindowInfoCmd.java:50)
	at com.intellij.openapi.wm.impl.CommandProcessor.runNext(CommandProcessor.java:78)
	at com.intellij.openapi.wm.impl.CommandProcessor.run(CommandProcessor.java:59)
	at com.intellij.openapi.wm.impl.commands.FinalizableCommand.finish(FinalizableCommand.java:36)
	at com.intellij.openapi.wm.impl.ToolWindowsPane$AddDockedComponentCmd.run(ToolWindowsPane.java:750)
	at com.intellij.openapi.wm.impl.CommandProcessor.runNext(CommandProcessor.java:78)
	at com.intellij.openapi.wm.impl.CommandProcessor.run(CommandProcessor.java:59)
	at com.intellij.openapi.wm.impl.commands.FinalizableCommand.finish(FinalizableCommand.java:36)
[...]

PR #5112, opened 2020-03-18
ACT: Don't try to initialize Cargo Tool Window if it already initialized

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5110.


vlad20012:
Works for me!
bors r+

Issue #4917, opened 2020-02-04
Support subslice patterns


Undin:
[Subslice patterns](https://github.com/rust-lang/rfcs/blob/master/text/2359-subslice-pattern-syntax.md) will be stabilized in 1.42. It would be great to support them in the plugin.
Some technical details can be found in https://github.com/rust-lang/rust/pull/67712

* [x] syntax #4778
* [ ] type inference
* [ ] annotations and inspections
  * [ ] experimental feature check
  * [ ] exhaustiveness

fujiapple852:
I see from the above that type inference is not yet available for subslice patterns and therefore presumably I shouldn't expect the type of `xs` to be inferred correctly at the moment?

```
fn foo() {
    fn bar(_: &[&str]) {}
    match vec![""].as_slice() {
        [] => {},
        [_, xs @ ..] => bar(xs)
    }
}
```

Currently this gives a type inference error in the IDE:
```
mismatched types [E0308] expected `&[&str]`, found `&&str`
```
If not let me know and I'll raise a bug.

Undin:
@fujiapple852 type inference is implemented in #5226
You can try it in tomorrow nightly build. See [instruction](https://intellij-rust.github.io/install) how to install nightly plugin version

PR #5114, opened 2020-03-19
ANN: Annotate slice patterns as experimental

mchernyavsky:
Relates to https://github.com/intellij-rust/intellij-rust/issues/4917.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -63,6 +63,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             override fun visitPatBox(o: RsPatBox) = checkPatBox(holder, o)
             override fun visitPatField(o: RsPatField) = checkPatField(holder, o)
             override fun visitPatBinding(o: RsPatBinding) = checkPatBinding(holder, o)
+            override fun visitPatRest(o: RsPatRest): Unit = checkPatRest(holder, o)
             override fun visitPath(o: RsPath) = checkPath(holder, o)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -354,6 +355,11 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         binding.ancestorStrict<RsValueParameterList>()?.let { checkDuplicates(holder, binding, it, recursively = true) }
     }
 
+    private fun checkPatIdent(holder: RsAnnotationHolder, patIdent: RsPatIdent) {
+        if (patIdent.pat !is RsPatRest) return
```
Undin: Whats' about patterns without bindings?
```rust
fn main() {
    let [x, ..] = [1, 2, 3];
}
```

Undin: Don't forget to check that you don't have false positive for tuple patterns `let (x, .., y)  = (1, 2, 3);` and all other places where we use `RsPatRest`

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -354,6 +355,11 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+        SLICE_PATTERNS.check(holder, patIdent, "slice patterns")
```
Undin: Let's highlight only `..` because:
* only `..` pattern was experimental
* you have to do it in case of rest pattern without binding
* in combination with type hints we get quite weird highlighting
image

mchernyavsky:
Fixed.

PR #5118, opened 2020-03-20
GRAM&PSI&STUB: get rid of TypeReference node

vlad20012:
I think, f3aa1ca9012747bc979c5a9fb9be66b35941d078 (#1422) was a mistake. It introduced an intermediate TypeReference node that wraps any all types in PSI.
image

This node was needed to put parentheses into it (if a type is wrapped into parentheses):
image

By this PR I remove TypeReference intermediate node from PSI tree. Now type tree is flatter
image

I solved the problem with parentheses by introducing ParenType. It appears in PSI only when there is a type in parentheses:
image

By this PR I want to reduce PSI/Stub tree complexity, reduce memory consumption and speedup tree traversing. I assume that a type in parens is a rare case comparing to types without parens.


matklad:
Agree that #1422 was a bad idea!

PR #5119, opened 2020-03-20
GRD: try jcenter instead of maven central

vlad20012:
Maven Central servers are sometimes unavailable from Travis CI (see [issue](https://issues.sonatype.org/browse/MVNCENTRAL-4985)). I think it's reasonable to add [jcenter](https://bintray.com/bintray/jcenter) mirror as a fallback.

(at first, let's try to use jcenter *instead* of maven central to check that it ever works)


vlad20012:
JCenter works!

Issue #3013, opened 2018-11-05
Highlight unsafe operations inside `unsafe` blocks


matklad:
Only some operations are actually unsafe inside an unsafe block (calls to usafe fn's, raw pointer dereference). It would be cool to highlight such operations!

See https://github.com/rust-analyzer/rust-analyzer/issues/190 for details.

I personally think this would be a rather important thing to add for folks who work with unsafe code a lot.

gnzlbg:
There are only 4 things that `unsafe {}` blocks allow one to do ([TRPL 2nd Ed 19.1 Unsafe Superpowers](https://doc.rust-lang.org/book/second-edition/ch19-01-unsafe-rust.html#unsafe-superpowers)): 

> * Dereference a raw pointer
> * Call an unsafe function or method
> * Access or modify a mutable static variable
> * Implement an unsafe trait

Only the first three are relevant for this issue because implementing an unsafe trait requires an `unsafe impl` independently of whether this `impl` is added inside an `unsafe {}` block or not.

garyttierney:
The commit referenced above (https://github.com/garyttierney/intellij-rust/commit/dc17f4fd99dc510f98ea89b1c486b98f5c6a091f) is a WIP implementation of this.  It re-uses the same style as Kotlin smart casts, though I'm not convinced this is the best visual representation of "unsafety".  

![screenshot from 2018-11-20 21-50-30](https://user-images.githubusercontent.com/1180094/48805140-7f8be600-ed0e-11e8-96ec-16405dee258c.png)

gnzlbg:
This is so cool.

vlad20012:
@garyttierney Maybe we shouldn't highlight a full function call and highlight the only name instead?
Also let's discuss possible highlighting style. I'm going to use smartcast-like highlighting for a stuff like #1815. But maybe it's not a problem to use the same style for unsafe

garyttierney:
That's a cool feature!  And yes, that makes sense.  That lets us disambiguate in the following case (which my implementation doesn't do at present):

```rust
extern "C" {
    static TEST: i32;
}

unsafe fn test(_: i32, _: i32) {}
fn test2() -> i32 { 0 }

fn main() {
    test(test2(), TEST);
    ---- -------  ^^^^ unsafe 
    |       |
} unsafe  safe
```

[...]

PR #5130, opened 2020-03-24
ANN: Fix union pattern match error

### LINKED ISSUE #5071: Confusing an union for a struct when using `match`
## Environment

0.2.117.2170-193

* **Rust toolchain version:**
rustc 1.41.1 (f3e1a954d 2020-02-24)
binary: rustc
commit-hash: f3e1a954d2ead4e2fc197c7da7d71e6c61bad196
commit-date: 2020-02-24
host: x86_64-pc-windows-msvc
release: 1.41.1
LLVM version: 9.0

* **IDE name and version:**
Build #IC-193.6494.35 (IntelliJ IDEA 2019.3.3 Community Edition)

* **Operating system:**
Windows 10

## Problem description
The plugin seem to confuse an union for a struct when using `match`, thus
presenting the error "Struct pattern does not match field 'f'" where f is not supposed to be matched at the same line.

## Steps to reproduce
Defining the next union:
```rust
union TestUnion {
    i: i32,
    f: f32
}
```

and matching with the next code:
```rust
fn test_function(tu: TestUnion) {
 unsafe {
   match tu {
            TestUnion { i: 42 } => { println!(" The meaning of.. la la la la la "); }
            TestUnion { f } = > { println!("Farewell and thanks for the fish, {} ", f); }
   } // match
 } // unsafe
} // fn
```

The code however compiles without issues.

Stzx:
Fixes #5071 

During the fix process, the following problems were found:

- Enter `u`, `union` keyword without prompt
- ~There are still problems with `union` automatic fix~
- `Union` does not detect if it exists in an `unsafe` block

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -59,7 +59,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             override fun visitLifetime(o: RsLifetime) = checkLifetime(holder, o)
             override fun visitModDeclItem(o: RsModDeclItem) = checkModDecl(holder, o)
             override fun visitModItem(o: RsModItem) = checkDuplicates(holder, o)
-            override fun visitUseSpeck(o: RsUseSpeck) = checkUseSpeck(holder,o)
+            override fun visitUseSpeck(o: RsUseSpeck) = checkUseSpeck(holder, o)
             override fun visitPatBox(o: RsPatBox) = checkPatBox(holder, o)
             override fun visitPatField(o: RsPatField) = checkPatField(holder, o)
             override fun visitPatBinding(o: RsPatBinding) = checkPatBinding(holder, o)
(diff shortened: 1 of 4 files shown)


ortem:
Hi @Stzx, thank you for PR and for exploring other problems related to unions. I've opened a separate issue #5138 for the first problem you described, but I didn't quite understand the second and third problems. Could you please describe them in more detail?

Stzx:
> * There are still problems with `union` automatic fix

See here: [unions - pattern-matching-on-unions](https://doc.rust-lang.org/reference/items/unions.html#pattern-matching-on-unions)

When the following code appears:
```rust
union TestUnion { i: i32, f: f32 }

fn main() {
    let tu = TestUnion { i: 1 };

    unsafe {
        match tu {
            <error descr="Struct pattern does not mention field `f` [E0027]">TestUnion { i }</error> => {}
            // after: TestUnion { i, f } => {}
            // This fix is wrong
        }
    }
}

```

----

> * `Union` does not detect if it exists in an `unsafe` block

See here: [unions - reading-and-writing-union-fields](https://doc.rust-lang.org/reference/items/unions.html#reading-and-writing-union-fields)
```rust
union TestUnion {  i: i32, }

fn main() {
    let tu = TestUnion { i: 1 };

    if tu.i == 1 {
        println!("Hello!");
    }
}
```
The read for `union` is not in the `unsafe` block, but the IDE does not prompt an error. Write operation has the same problem.

ortem:
@Stzx thanks for noticing, I've updated our issue on unsafe #3013.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1185,6 +1185,26 @@ sealed class RsDiagnostic(
             )
         }
     }
+
+    class MissingFieldsInUnionPattern(pat: RsPat) : RsDiagnostic(pat) {
+        override fun prepare(): PreparedAnnotation {
+            return PreparedAnnotation(
+                ERROR,
+                NONE,
```
ortem: JFYI: I think it's strange that there is no error code for this case. So I've opened the corresponding [issue](https://github.com/rust-lang/rust/issues/70620)

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1195,7 +1215,8 @@ enum class RsErrorCode {
     E0403, E0404, E0407, E0415, E0424, E0426, E0428, E0433, E0449, E0451, E0463,
     E0518, E0562, E0569, E0583, E0586, E0594,
     E0603, E0614, E0616, E0618, E0624, E0658, E0666, E0667, E0688, E0695,
-    E0704, E0732;
+    E0704, E0732,
+    NONE;
```
ortem: Let's use `null` instead of `RsErrorCode.NONE`

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -115,13 +115,27 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
-            RsDiagnostic.ExtraFieldInStructPattern(it).addToHolder(holder)
-        }
-        if (missingFields.isNotEmpty() && patStruct.patRest == null) {
-            RsDiagnostic.MissingFieldsInStructPattern(patStruct, declaration, missingFields).addToHolder(holder)
+
+        if (declaration is RsStructItem && declaration.kind == RsStructKind.UNION) {
+            extraFields.forEach { RsDiagnostic.ExtraFieldInStructPattern(it, "union").addToHolder(holder) }
+
+            val extraFieldNames = extraFields.map { it.kind.fieldName }
+            val bodyFieldNames = bodyFields.filter { it.kind.fieldName !in extraFieldNames }
```
ortem: Consider this case from your tests:
```Rust
union TestUnion { i: i32, f: f32 }

fn test_fun(tu: TestUnion) {
    unsafe {
        match tu {
            TestUnion { x } => {}
        }
    }
}
```

Rustc omits only one error: `union U does not have a field named a`.
I think it's enough and we should not omit two errors in this case, and then you can just use `bodyFields` instead of `bodyFieldNames`. What do you think?

ortem:
@Stzx Thank you!

bors r+

PR #5135, opened 2020-03-27
Do not add directory as source/excluded entries if it doesn't exist

Undin:


Undin:
СС @SergeevPavel

SergeevPavel:
Do it will be added if I will create this folders later?

Undin:
They will be added after next cargo refresh.
You can do in manually via the corresponding action in `Cargo` tool window or if `Watch Cargo.toml` option (I know it's not the best name) in `Preferences | Languages & Frameworks | Rust | Cargo` is enabled (it's enabled by default), the plugin invokes refresh automatically after you create any rust file in the corresponding dir

lancelote:
Did it cause any trouble? Btw are source directories (e.g. `tests`) expected to be removed from `iml` file once deleted?

Undin:
> Did it cause any trouble?

Not for generals users. But IDEA project has a test that checks all content roots exists. And since the project has several Rust modules, it fails this test

> Btw are source directories (e.g. tests) expected to be removed from iml file once deleted?

It would be great but it's not implemented yet and I wasn't going to implement it in this PR. But I will do it eventually

Issue #5136, opened 2020-03-27
`Extract enum variant` intention doesn't replace usages in struct literals


mchernyavsky:
## Environment

* **IntelliJ Rust plugin version:** 0.2.119.3026-201-nightly
* **Rust toolchain version:** 1.43.0-nightly (3dbade652 2020-03-09) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 Beta (CL-201.6668.10)
* **Operating system:** macOS 10.15.4

## Problem description

title

## Steps to reproduce
```rust
enum E {
    /*caret*/V1 { x: i32 },
    V2
}

fn main() {
    let v = E::V1 { x: 42 };
    if let E::V1 { x } = v {
        println!("{}", x)
    };
}
```

mchernyavsky:
cc @Kobzol

Kobzol:
Right, I only handled struct patterns, but not struct constructors. I'll send a PR.

PR #5142, opened 2020-03-27
INT: replace usages of struct literals and tuple constructors in ExtractEnumVariantIntention

Kobzol:
To preserve original comments and other things, I reuse the text of the original struct/tuple literals.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5136


mchernyavsky:
I will fix it by myself.

Thanks!
bors r+

PR #5143, opened 2020-03-28
COMP: Insert trailing comma in match after "Complete current statement"

### LINKED ISSUE #5140: Insert trailing comma in match after "Complete current statement"
## Environment

* **IntelliJ Rust plugin version:** 0.2.119.3025-201-nightly
* **Rust toolchain version:** 1.42.0 (b8cedc004 2020-03-09) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 Beta (CL-201.6668.10)
* **Operating system:** macOS 10.15.3

## Problem description

I'm used to "Complete current statement" action but it doesn't do the best job in `match`, I guess it is safe to insert a trailing `,` there.

## Steps to reproduce

```rust
fn main() {
    let a = 1;

    match a {
        1 => println!(1)<caret>
    };
}
```

Invoke "Complete current statement" (<kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>Return</kbd>):

```rust
fn main() {
    let a = 1;

    match a {
        1 => println!(1)
        <caret>
    };
}
```

Expected:

```rust
fn main() {
    let a = 1;

    match a {
        1 => println!(1),
        <caret>
    };
}
```

Stzx:
Closed #5140

### DIFF
--- a/src/main/kotlin/org/rust/ide/typing/assist/CommaFixer.kt
+++ b/src/main/kotlin/org/rust/ide/typing/assist/CommaFixer.kt
@@ -1,0 +1,23 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.typing.assist
+
+import com.intellij.lang.SmartEnterProcessorWithFixers
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsBlockExpr
+import org.rust.lang.core.psi.RsMatchArm
+
+/**
+ * Fixer that adds missing comma at the end of statements.
+ */
+class CommaFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
+    override fun apply(editor: Editor, processor: RsSmartEnterProcessor, element: PsiElement) {
+        if (element is RsMatchArm && element.expr !is RsBlockExpr && element.comma == null) {
+            editor.document.insertString(element.textRange.endOffset, ",")
+        }
+    }
+}
--- a/src/main/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessor.kt
@@ -13,6 +13,7 @@ import com.intellij.psi.PsiWhiteSpace
 import org.rust.lang.core.psi.RsBlock
 import org.rust.lang.core.psi.RsElementTypes.LBRACE
 import org.rust.lang.core.psi.RsElementTypes.RBRACE
+import org.rust.lang.core.psi.RsMatchArm
 import org.rust.lang.core.psi.ext.ancestors
 
 /**
@@ -23,10 +24,13 @@ class RsSmartEnterProcessor : SmartEnterProcessorWithFixers() {
     init {
         addFixers(
             MethodCallFixer(),
-            SemicolonFixer())
+            SemicolonFixer(),
+            CommaFixer()
+        )
 
         addEnterProcessors(
-            PlainEnterProcessor())
+            PlainEnterProcessor()
+        )
     }
 
     override fun getStatementAtCaret(editor: Editor, psiFile: PsiFile): PsiElement? {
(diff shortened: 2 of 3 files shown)


src/main/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessor.kt
```diff
@@ -36,6 +39,10 @@ class RsSmartEnterProcessor : SmartEnterProcessorWithFixers() {
             val elementType = each.node.elementType
             when {
                 elementType == LBRACE || elementType == RBRACE -> continue@loop
+                each is RsMatchArm && each.parent is RsMatchBody -> {
+                    val psiFactory = RsPsiFactory(psiFile.project)
+                    each.add(psiFactory.createComma())
```
Undin: It's not supposed to modify PSI here
To do it, create an implementation of `SmartEnterProcessorWithFixers.Fixer` and register it in constructor. See `org.rust.ide.typing.assist.MethodCallFixer` as example

src/main/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessor.kt
```diff
@@ -36,6 +39,10 @@ class RsSmartEnterProcessor : SmartEnterProcessorWithFixers() {
             val elementType = each.node.elementType
             when {
                 elementType == LBRACE || elementType == RBRACE -> continue@loop
+                each is RsMatchArm && each.parent is RsMatchBody -> {
```
Undin: `each.parent is RsMatchBody` it's always true according to our current grammar so I think we can omit it

src/main/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessor.kt
Undin: If match arm already has comma, the current implementation adds a new one

Undin: Also, probably it's better to insert text via `editor.document` not to care about psi correctness
See `org.rust.ide.typing.assist.MethodCallFixer` as exmaple

src/main/kotlin/org/rust/ide/typing/assist/CommaFixer.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.rust.lang.core.psi.RsMatchArm
+
+/**
+ * Fixer that adds missing comma at the end of statements.
+ */
+class CommaFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
+    override fun apply(editor: Editor, processor: RsSmartEnterProcessor, element: PsiElement) {
+        if (element is RsMatchArm) {
+            val lastElement = element.lastChild
+            if ((lastElement is RsCallExpr || lastElement is RsMacroExpr) && element.lastChild.text != ",")
```
Undin: Why do we need `(lastElement is RsCallExpr || lastElement is RsMacroExpr)` condition?
What if I write some expression like `1` or `"foo".as_bytes()`?

Stzx: Is to prevent this from happening:

```rust
fn main() {
    let a = 1;

    match a {
        1 => <caret>
    };
}
```

I don't think this situation should add commas. What do you think? @Undin

src/main/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessor.kt
```diff
@@ -36,7 +40,7 @@ class RsSmartEnterProcessor : SmartEnterProcessorWithFixers() {
             val elementType = each.node.elementType
             when {
                 elementType == LBRACE || elementType == RBRACE -> continue@loop
-                each.parent is RsBlock -> return each
+                each.parent is RsMatchBody || each.parent is RsBlock -> return each
```
Undin, line 43: Let's check that `each` is `RsMatchArm` directly

src/main/kotlin/org/rust/ide/typing/assist/CommaFixer.kt
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Fixer that adds missing comma at the end of statements.
+ */
+class CommaFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
+    override fun apply(editor: Editor, processor: RsSmartEnterProcessor, element: PsiElement) {
+        if (element is RsMatchArm) {
+            if (element.expr != null && element.lastChild.text != ",")
+                editor.document.insertString(element.textRange.endOffset, ",")
+        }
```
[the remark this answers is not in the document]
Stzx: I think it should be `element.comma == null`.

Undin:
bors try

Stzx:
Thanks for your review.

Issue #5117, opened 2020-03-20
Postfix completion doesn't work well to print variables of type string


adriafarres:
## Environment

* **IntelliJ Rust plugin version**: 0.2.118.2171-193
* **Rust toolchain version**: 1.42.0
* **IDE name and version**: IntelliJ IDEA 2019.3.4
* **Operating system**: Linux

# Problem and explanation
Postfix completion works well for strings when doing the following:
```rust
"Some string".println
```

In such case, the result ends up being
```rust
println!("Some string")
```
However, in cases where we have a variable, such as

```rust
let s = "Some string";
s.println
```
the result ends up being
```rust
println!(s)
```
when it should be 
```rust
println!("{}, s)
```

PR #5149, opened 2020-03-29
COMP: fix println postfix template for string variables

Kobzol:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5117

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/PrintlnPostfixTemplate.kt
@@ -33,10 +33,13 @@ class PrintlnPostfixTemplate(provider: RsPostfixTemplateProvider, private val ma
         Debug("{:?}");
 
         companion object {
-            fun fromExpr(expr: RsExpr): Fmt = when {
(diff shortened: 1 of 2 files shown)


Kobzol:
Added tests for strings, byte strings, raw strings, raw byte strings (both literals and variables), refactored the function and renamed the commit.

This is probably not relevant for this PR, but wouldn't it make more sense to prefer `Display` before `Debug` and not the other way around? From my experience, more types implement `Debug` than `Display` and when a type implements `Display`, it also usually implements `Debug` (but not necessarily the other way around). So I think that we should prefer using `Display` and only if that is missing use `Debug`.

Undin:
> This is probably not relevant for this PR, but wouldn't it make more sense to prefer Display before Debug and not the other way around? From my experience, more types implement Debug than Display and when a type implements Display, it also usually implements Debug (but not necessarily the other way around). So if this is more common (I can only speak from my experience), I think that we should prefer using Display and only if that is missing use Debug.

We've just discussed this thing. And came to the same decision.
Would you like to fix it (after this PR will be merged)?

Kobzol:
Sure :)

PR #5150, opened 2020-03-29
ANN: Fix false positive E0451 when constructing enum

dima74:
Fix false positive "Field is private" error for such code (introduced in #5120):

```rust
mod some_module {
    pub enum Foo {
        Foo1 { field: i32 },
    }
}

fn main() {
    let foo = some_module::Foo::Foo1 { field: 1 };
}
```


mchernyavsky:
bors retry

Undin:
bors retry

Undin:
bors r=mchernyavsky

PR #5155, opened 2020-03-30
Fix move analysis of struct literals with base expr

### LINKED ISSUE #5013: "Use of moved value" false positive
## Environment

* **IntelliJ Rust plugin version: 0.2.116.2163-193**
* **Rust toolchain version: v1.42.0**
* **IDE name and version: IntelliJ IDEA 2019.3.3**
* **Operating system: Ubuntu 19.10**

## Problem description
The main() function below compiles but is incorrectly flagged as a "Use of a moved value" on the "Destructure the point using a `let` binding". 
The snippet comes from https://doc.rust-lang.org/rust-by-example/custom_types/structs.html

## Steps to reproduce

```rust
// A struct with two fields
struct Point {
    x: f32,
    y: f32,
}

fn main() {

    // Instantiate a `Point`
    let point: Point = Point { x: 10.3, y: 0.4 };

    // Access the fields of the point
    println!("point coordinates: ({}, {})", point.x, point.y);

    // Make a new point by using struct update syntax to use the fields of our
    // other one
    let bottom_right = Point { x: 5.2, ..point };

    // Destructure the point using a `let` binding
    let Point { x: top_edge, y: left_edge } = point;
}
```

### LINKED ISSUE #5151: False positive: Use of moved value
## Environment

* **IntelliJ Rust plugin version:** 0.2.118.2171-193
* **Rust toolchain version:** cargo 1.44.0-nightly (7019b3ed3 2020-03-17)
* **IDE name and version:** IDEA 2019.3.4

## Problem description

Using let expression to deconstruct object, but tip error(`Use of moved value`)

But the compilation can pass.

## Steps to reproduce

Use the following code:
```rust
#[derive(Debug)]
struct Vector2f {
    x: f32,
    y: f32,
}

fn main() {
    let vector = Vector2f { x: 1.0, y: 3.0 };
    let new_vector = Vector2f { x: 2.0, ..vector };

    println!("{:?}", new_vector);

    let Vector2f { x: old_x, y: old_y } = vector;
    // Tips `Use of moved value`

    println!("{:?}", old_x);
    println!("{:?}", old_y);
}
```

ortem:
Fixes #5013
Fixes #5151

- [x] Run `RsRealProjectAnalysisTest` before merge

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/borrowck/ExprUseWalker.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/borrowck/ExprUseWalker.kt
@@ -22,10 +22,12 @@ import org.rust.lang.core.types.infer.Categorization.Local
 import org.rust.lang.core.types.infer.Cmt
 import org.rust.lang.core.types.infer.MemoryCategorizationContext
 import org.rust.lang.core.types.infer.MutabilityCategory
+import org.rust.lang.core.types.infer.substituteOrUnknown
 import org.rust.lang.core.types.regions.ReScope
 import org.rust.lang.core.types.regions.Scope
 import org.rust.lang.core.types.ty.TyAdt
 import org.rust.lang.core.types.ty.TyFunction
+import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.ty.isMovesByDefault
 import org.rust.lang.core.types.type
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/types/borrowck/ExprUseWalker.kt
```diff
@@ -421,23 +420,29 @@ class ExprUseWalker(private val delegate: Delegate, private val mc: MemoryCatego
...
-                    val fieldCmt = Cmt(withExpr, interior, withCmt.mutabilityCategory.inherit(), withType)
-                    delegateConsume(withExpr, fieldCmt)
+        // With base expr, e.g. `S { x, ..s }`
+        if (withExpr != null) {
+            val withCmt = mc.processExpr(withExpr)
+            val withType = withCmt.ty
+            if (withType is TyAdt) {
+                val withFields = (withType.item as? RsStructItem)?.namedFields.orEmpty()
+                for (withField in withFields) {
+                    val isMentioned = fields.any { it.identifier?.text == withField.identifier.text }
```
vlad20012, line 430:
```suggestion
                    val isMentioned = fields.any { it.name == withField.name }
```
... to support raw identifiers here.
Can be done in another PR b/c this behavior wasn't changed by the PR.

src/main/kotlin/org/rust/lang/core/types/borrowck/ExprUseWalker.kt
```diff
@@ -421,23 +420,29 @@ class ExprUseWalker(private val delegate: Delegate, private val mc: MemoryCatego
...
+                    // Consume only needed (not mentioned before) fields of `withExpr`
+                    if (!isMentioned) {
+                        val withFieldType = when (val rawType = withField.typeReference?.type) {
```
vlad20012, line 433: `withType.typeParameterValues` should be substituted unconditionally. Otherwise, the type is wrong. E.g. `struct A<T> { x: B<T> }` must be substotuted

ortem: Oh, really. Thank you!

src/main/kotlin/org/rust/lang/core/types/borrowck/ExprUseWalker.kt
ortem: Looks like it should be `val isMentioned = fields.any { it.referenceName == withField.name }` (`referenceName`, not `name`)

PR #5157, opened 2020-03-30
INSP: add RsRedundantElseInspection

Kobzol:
Adds inspection to find redundant else branches, as discussed in https://github.com/intellij-rust/intellij-rust/pull/4972.

Right now it doesn't work for these irrefutable patterns:
```rust
enum E { V1 }
fn main() {
    let a = E::V1;
    if let E::V1 = a {
    } else { // <- this else should be redundant
    }
}
```
because `let E::V1 = a` is `RsPatConst` and `RsPat.isIrrefutable` returns `false` for it. Should I change this?

Should the inspection's warning range cover only `else` or the whole block after the `else`?

It would be also nice if it recognized irrefutable conditions like `if true` or `if 5 == 5`, but we need some basic constant propagation for this to work (is there something like this in the plugin already)?

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsRedundantElseInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsRedundantElseInspection.kt
@@ -1,0 +1,59 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import org.rust.ide.inspections.fixes.SubstituteTextFix
+import org.rust.lang.core.psi.RsCondition
+import org.rust.lang.core.psi.RsElseBranch
+import org.rust.lang.core.psi.RsIfExpr
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.endOffsetInParent
+import org.rust.lang.core.psi.ext.isIrrefutable
+import org.rust.lang.core.psi.ext.leftSiblings
+import org.rust.lang.core.psi.ext.rangeWithPrevSpace
+
+/**
+ * Detects redundant `else` statements preceded by an irrefutable pattern.
+ * Quick fix: Remove `else`
+ */
+class RsRedundantElseInspection : RsLocalInspectionTool() {
+    override fun getDisplayName() = "Redundant else"
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) =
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/inspections/RsRedundantElseInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) =
+        object : RsVisitor() {
+            override fun visitElseBranch(expr: RsElseBranch) {
+                if (!expr.isRedundant()) return
+
+                val elseExpr = expr.`else`
+                holder.registerProblem(
+                    expr,
+                    TextRange(elseExpr.startOffsetInParent, elseExpr.endOffsetInParent),
```
mchernyavsky, line 35:
```suggestion
                    elseExpr.textRangeInParent,
```

src/main/kotlin/org/rust/ide/inspections/RsRedundantElseInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        while (candidate is RsElseBranch || candidate is RsIfExpr) {
+            candidate.leftSiblings.filterIsInstance<RsCondition>().forEach { set.add(it) }
+            candidate = candidate.parent
+        }
+
+        return set.any {
+            it.orPats?.patList?.all { pat -> pat?.isIrrefutable ?: false } ?: false
+        }
+    }
```
mchernyavsky, line 58:
```suggestion
    companion object {
        private val RsElseBranch.isRedundant: Boolean
            get() {
                val set = mutableSetOf<RsCondition>()
                var candidate: PsiElement = this

                while (candidate is RsElseBranch || candidate is RsIfExpr) {
                    candidate.leftSiblings.filterIsInstance<RsCondition>().forEach { set.add(it) }
                    candidate = candidate.parent
                }

                return set.any {
                    it.orPats?.patList?.all { pat -> pat?.isIrrefutable ?: false } ?: false
                }
            }
    }
```

Kobzol:
I created a new PR for `RsPatConst` here: https://github.com/intellij-rust/intellij-rust/pull/5199. I'll wait for that PR to be merged and also try to add some basic constant evaluation support.

Kobzol:
I rebased onto master with #5199 and added basic constant evaluation. I noticed that the evaluator is handling types in a peculiar way, if I have an expression `1 == 1`, it's of the `bool` type, so the evaluator is trying to search for `&&` or `||`, it doesn't find it so it fails. This could probably be improved in a different PR.

mchernyavsky:
bors retry

PR #5177, opened 2020-04-02
REF: Support binding RsPath to RsMod

dima74:
It will be used in move refactoring (#5181)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
@@ -32,6 +32,27 @@ val RsQualifiedNamedElement.qualifiedName: String?
         return "$cargoTarget$inCratePath"
     }
 
+fun RsQualifiedNamedElement.qualifiedNameRelativeTo(mod: RsMod): String? {
+    val crateRelativePath = crateRelativePath
+    if (mod.crateRoot != crateRoot || crateRelativePath == null) {
+        return qualifiedName
+    }
+
+    check(crateRelativePath.isEmpty() || crateRelativePath.startsWith("::"))
+    val absolutePath = "crate$crateRelativePath"
+    if (!containingMod.superMods.contains(mod)) return absolutePath
+
+    val modPathPrefix = mod.crateRelativePath?.let { "$it::" } ?: return absolutePath
+    if (!crateRelativePath.startsWith(modPathPrefix)) return absolutePath
+    val relativePath = crateRelativePath.removePrefix(modPathPrefix)
+
+    val cargoWorkspace = cargoWorkspace ?: return relativePath
+    if (cargoWorkspace.packages.any { relativePath.startsWith("${it.normName}::") }) {
+        return "self::$relativePath"
+    }
+    return relativePath
+}
+
 @Suppress("DataClassPrivateConstructor")
 data class RsQualifiedName private constructor(
     val crateName: String,
--- a/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt
@@ -96,6 +96,39 @@ class RsPathReferenceImpl(
             .map { it.foldTyInferWith { if (it is TyInfer.TyVar) TyInfer.TyVar(it.origin) else it } }
     }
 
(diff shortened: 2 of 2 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
```diff
@@ -32,6 +32,27 @@ val RsQualifiedNamedElement.qualifiedName: String?
...
+        && modPathPrefix != null
+        && crateRelativePath.startsWith(modPathPrefix)
+    ) {
+        var result = crateRelativePath.removePrefix(modPathPrefix)
+        if (cargoWorkspace?.packages?.any { result.startsWith("${it.normName}::") } == true) {
+            result = "self::$result"
+        }
+        result
+    } else {
+        "crate$crateRelativePath"
```
ortem, line 52: Do I understand correctly that `crateRelativePath` **always** starts with `::`?

src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
```diff
@@ -32,6 +32,27 @@ val RsQualifiedNamedElement.qualifiedName: String?
...
+        return qualifiedName
+    }
+
+    val modPathPrefix = mod.crateRelativePath?.let { "$it::" }
+    return if (containingMod.superMods.contains(mod)
+        && modPathPrefix != null
+        && crateRelativePath.startsWith(modPathPrefix)
+    ) {
+        var result = crateRelativePath.removePrefix(modPathPrefix)
+        if (cargoWorkspace?.packages?.any { result.startsWith("${it.normName}::") } == true) {
```
ortem, line 47: Let's introduce `cargoWorkspace` not to deal with nullability and `== true`. And probably it can be `val cargoWorkspace = ... ?: return ...`, but you can choose any way of doing this

src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
```diff
@@ -32,6 +32,27 @@ val RsQualifiedNamedElement.qualifiedName: String?
...
     }
 
+fun RsQualifiedNamedElement.qualifiedNameRelativeTo(mod: RsMod): String? {
+    val crateRelativePath = crateRelativePath
+    if (mod.crateRoot != crateRoot || crateRelativePath == null) {
+        return qualifiedName
+    }
+
+    val modPathPrefix = mod.crateRelativePath?.let { "$it::" }
+    return if (containingMod.superMods.contains(mod)
```
ortem, line 42: IMO such a complex condition is difficult to read and understand. As a possible solution, you can introduce something, for example `val modPathPrefix = ... ?: return ...`

src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
```diff
@@ -32,6 +32,27 @@ val RsQualifiedNamedElement.qualifiedName: String?
...
+        && modPathPrefix != null
+        && crateRelativePath.startsWith(modPathPrefix)
+    ) {
+        var result = crateRelativePath.removePrefix(modPathPrefix)
```
ortem, line 46: Maybe add two return cases instead of `var`?

src/main/kotlin/org/rust/lang/core/psi/ext/RsQualifiedNamedElement.kt
dima74: `crateRelativePath` is empty for crate root (so path would be `crate`, which is correct). Otherwise it always starts with `;;`, because all implementations call [`RsPsiImplUtil::crateRelativePath`](https://github.com/intellij-rust/intellij-rust/blob/eeec35c110b601b6640fab0acf2a859bc08d9305/src/main/kotlin/org/rust/lang/core/psi/RsPsiImplUtil.kt#L21) method, which uses [`RsPsiImplUtil::modCrateRelativePath`](https://github.com/intellij-rust/intellij-rust/blob/eeec35c110b601b6640fab0acf2a859bc08d9305/src/main/kotlin/org/rust/lang/core/psi/RsPsiImplUtil.kt#L27) method, which can return only `null`, empty string, or string prefixed with `::`

I added `check(crateRelativePath.isEmpty() || crateRelativePath.startsWith("::"))`

dima74:
bors r=ortem

PR #5182, opened 2020-04-02
INT: Fix and enable "Extract enum variant" intention

### LINKED ISSUE #5024: Extract enum variant to struct
It would be cool if there were an intention (or better refactoring) to extract enum variant to separate struct.

Example:
```rust 
enum A {
    Variant1 { name: String, id: usize, }
}
```
extract into something like:
```rust
enum A {
     Variant1(Variant1)
}

struct Variant1 { 
    name: String, 
    id: usize, 
}
```
As a refactoring it could also replace usages, but intention is already great.

mchernyavsky:
Bunch of fixes:
* Enable the intention.
* Get rid of `extractVisibility` and related staff since visibility qualifiers for enum variant's fields are not permitted ([E0449](https://doc.rust-lang.org/error-index.html#E0449)).
* Now generated struct has the same visibility qualifier as parent enum has.
* Since enum's fields have public visibility by default, add `pub` qualifier to all fields in generated struct if the corresponding field in the enum variant has no visibility qualifier.
* After replacing usages of the original enum variant, import the generated struct if needed.
* Fix struct renaming if an element with the same name already exists in the scope.

There is one more problem. `RsInPlaceVariableIntroducer` doesn't work if the element to rename has non-local use scope.

Closes #5024

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/ExtractEnumVariantIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/ExtractEnumVariantIntention.kt
@@ -10,6 +10,8 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiDocumentManager
 import com.intellij.psi.PsiElement
 import com.intellij.psi.search.searches.ReferencesSearch
+import com.intellij.psi.util.PsiTreeUtil
+import org.rust.ide.inspections.import.RsImportHelper.importElements
 import org.rust.ide.refactoring.RsInPlaceVariableIntroducer
 import org.rust.lang.core.ARBITRARY_ENUM_DISCRIMINANT
 import org.rust.lang.core.FeatureAvailability
@@ -46,8 +48,8 @@ class ExtractEnumVariantIntention : RsElementBaseIntentionAction<ExtractEnumVari
     }
 
     override fun invoke(project: Project, editor: Editor, ctx: Context) {
-        val element = createElement(ctx.variant)
         val factory = RsPsiFactory(project)
+        val element = createElement(ctx.variant, factory)
 
         val enum = ctx.variant.parentEnum
         val parameters = filterTypeParameters(element.typeReferences, enum.typeParameterList)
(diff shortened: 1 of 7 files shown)


Kobzol:
Oh wow, I can't believe that I didn't actually test whether the visibility make sense :D Good job :+1:

src/main/kotlin/org/rust/ide/intentions/ExtractEnumVariantIntention.kt
```diff
@@ -314,7 +369,7 @@ private data class CollectLifetimesVisitor(
     }
 }
 
-fun gatherLifetimes(
+private fun gatherLifetimes(
```
Undin, line 372: Let's make the same with `gatherTypeParameters`

src/main/kotlin/org/rust/ide/intentions/ExtractEnumVariantIntention.kt
```diff
@@ -179,92 +197,129 @@ class ExtractEnumVariantIntention : RsElementBaseIntentionAction<ExtractEnumVari
...
 
-        for (pat in parent.patList) {
-            val comma = pat.nextSibling
-            pat.delete()
-            if (comma.text == ",") {
-                comma.delete()
+    override fun createStruct(vis: String?, name: String, typeParameters: String, whereClause: String): RsStructItem {
+        val formattedVis = if (vis == null) "" else "$vis "
+        val fields = formatFields()
+        return factory.createStruct("${formattedVis}struct $name$typeParameters$fields$whereClause;")
```
Undin, line 246: `org.rust.ide.intentions.TupleVariant#createStruct` and `org.rust.ide.intentions.StructVariant#createStruct` looks very similar
Maybe it's better not to override it in each inheritor, move all code in base class and make `when` for struct text?

mchernyavsky:
@Undin

> Some examples when the current implementation doesn't work properly:

Fixed, except [2](https://github.com/intellij-rust/intellij-rust/issues/5265) and 3.

mchernyavsky:
@Undin Fixed

Undin:
@mchernyavsky don't hesitate to merge it yourself after approve of PR
bors r+

Issue #4693, opened 2019-11-26
Support features declared in workspace Cargo.toml


Undin:
After #4635 the plugin can handle [features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) declared in dependencies.
But features declared in your workspace are still unsupported.

So it would be great to support them as well

PR #5189, opened 2020-04-03
CARGO: Support configuring workspace features in `Cargo.toml`

### LINKED ISSUE #2821: Incorrectly warning that there are duplicate definitions with the same name
## Environment
* **Intellij-Rust plugin version:** 0.2.0.2104-182
* **Rust toolchain version:** 1.27.0
* **IDE name and version:** IntelliJ IDEA 2018.2.2 (Community Edition),Build #IC-182.4129.33
* **Operating system:** OSX 10.12.6

## Problem description

IntelliJ rust incorrectly says that there are duplicate definitions with the same name citing error [[0201]](https://doc.rust-lang.org/error-index.html#E0201)

## Steps to reproduce
```rust 
pub struct Foo {
    bar: i32,
    #[cfg(feature = "debug")]
    foo: i32,
}

impl Foo {
    fn new() -> Foo {
        Foo { bar: 0 }
    }

    #[cfg(feature = "debug")]
    fn new() -> Foo {
        Foo { bar: 0, foo: 1 }
    }
}
```

### LINKED ISSUE #4075: Code completion and syntax coloring with fns from some mods (e.g. compute example in gfx-rs project)
## Environment
* all the recent nightly intellij-rust plugin
* all rust toolchain
* CLion
* all OS

## Problem description
Code compleation and syntax coloring with fns from some mods (e.g. compute example in gfx-rs project):
https://github.com/gfx-rs/gfx/blob/master/examples/compute/main.rs#L156
from the line, all command_buffer.xxx() invocations are no code completion and syntax coloring.

Thx.

### LINKED ISSUE #6215: Incorrectly reports duplicate imports when using compile time conditions.
## Environment

* **IntelliJ Rust plugin version:** 0.3.132.3395-202
* **Rust toolchain version:** 1.46.0 (04488afe3 2020-08-24) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.2.3 (CL-202.7319.72)
* **Operating system:** Linux 5.8.11-200.fc32.x86_64

## Problem description

Reports duplicate imports, even when only one is enabled through a conditional `cfg`.

## Steps to reproduce

~~~rust
#[cfg(feature = "stm32f4xx")]
use stm32f4xx_hal::stm32::Peripherals as DevicePeripherals;

#[cfg(feature = "stm32f7xx")]
use stm32f7xx_hal::pac::Peripherals as DevicePeripherals;
~~~

Results in:

image

and:

image

However, this is wrong. As in the current configuration only one feature is enabled. I would expect the editor to follow the external linter configuration.

ortem:
Related to #4693.

This PR is logically based on the original @kumbayo's PRs (thank you very much for your work!) but provides a different UI/UX for viewing and enabling/disabling features.
Closes #2166
Closes #2840
Closes #3031
Closes #2821
Closes #4075
Closes #6215

Now user can enable or disable any specific workspace feature in the `Cargo.toml` file, and our name resolution and other code insight will take them into account.

features

Also supports features in workspaces with multiple packages and relations between them:

features2

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
@@ -17,8 +17,11 @@ import com.intellij.openapi.vfs.VfsUtil
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.util.messages.Topic
 import org.rust.cargo.CargoConstants
+import org.rust.cargo.project.model.impl.UserDisabledFeatures
 import org.rust.cargo.project.settings.rustSettings
 import org.rust.cargo.project.workspace.CargoWorkspace
+import org.rust.cargo.project.workspace.FeatureState
+import org.rust.cargo.project.workspace.PackageFeature
 import org.rust.cargo.toolchain.RsToolchain
 import org.rust.cargo.toolchain.impl.RustcVersion
 import org.rust.cargo.toolchain.isRustupAvailable
@@ -50,6 +53,8 @@ interface CargoProjectsService {
     fun refreshAllProjects(): CompletableFuture<out List<CargoProject>>
     fun discoverAndRefresh(): CompletableFuture<out List<CargoProject>>
     fun suggestManifests(): Sequence<VirtualFile>
+
+    fun modifyFeatures(cargoProject: CargoProject, features: Set<PackageFeature>, newState: FeatureState)
 
     companion object {
         val CARGO_PROJECTS_TOPIC: Topic<CargoProjectsListener> = Topic(
@@ -93,9 +98,12 @@ interface CargoProject : UserDataHolderEx {
     val stdlibStatus: UpdateStatus
     val rustcInfoStatus: UpdateStatus
 
-    val mergedStatus: UpdateStatus get() = workspaceStatus
-        .merge(stdlibStatus)
-        .merge(rustcInfoStatus)
+    val mergedStatus: UpdateStatus
+        get() = workspaceStatus
+            .merge(stdlibStatus)
+            .merge(rustcInfoStatus)
+
+    val userDisabledFeatures: UserDisabledFeatures
 
     sealed class UpdateStatus(private val priority: Int) {
         object UpToDate : UpdateStatus(0)
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
@@ -5,15 +5,13 @@
 
 package org.rust.cargo.project.model.impl
 
+import com.intellij.codeInsight.daemon.DaemonCodeAnalyzer
 import com.intellij.notification.NotificationType
 import com.intellij.openapi.application.ModalityState
 import com.intellij.openapi.application.invokeAndWaitIfNeeded
 import com.intellij.openapi.application.invokeLater
 import com.intellij.openapi.application.runWriteAction
-import com.intellij.openapi.components.PersistentStateComponent
-import com.intellij.openapi.components.State
-import com.intellij.openapi.components.Storage
-import com.intellij.openapi.components.StoragePathMacros
+import com.intellij.openapi.components.*
 import com.intellij.openapi.fileTypes.FileTypeManager
 import com.intellij.openapi.module.Module
 import com.intellij.openapi.module.ModuleUtilCore
@@ -49,22 +47,15 @@ import org.rust.cargo.project.settings.RustProjectSettingsService.RustSettingsCh
 import org.rust.cargo.project.settings.RustProjectSettingsService.RustSettingsListener
 import org.rust.cargo.project.settings.rustSettings
 import org.rust.cargo.project.toolwindow.CargoToolWindow.Companion.initializeToolWindow
-import org.rust.cargo.project.workspace.CargoWorkspace
-import org.rust.cargo.project.workspace.PackageOrigin
-import org.rust.cargo.project.workspace.StandardLibrary
-import org.rust.cargo.project.workspace.additionalRoots
+import org.rust.cargo.project.workspace.*
 import org.rust.cargo.runconfig.command.workingDirectory
 import org.rust.cargo.toolchain.RsToolchain
 import org.rust.cargo.util.AutoInjectedCrates
 import org.rust.ide.notifications.showBalloon
 import org.rust.lang.RsFileType
 import org.rust.lang.core.macros.macroExpansionManager
-import org.rust.openapiext.CachedVirtualFile
-import org.rust.openapiext.TaskResult
-import org.rust.openapiext.modules
-import org.rust.openapiext.pathAsPath
-import org.rust.stdext.AsyncValue
-import org.rust.stdext.applyWithSymlink
+import org.rust.openapiext.*
+import org.rust.stdext.*
 import org.rust.taskQueue
 import java.nio.file.Path
 import java.nio.file.Paths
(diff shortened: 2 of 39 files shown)


Exidex:
I am curious, how does it work with multi-package projects? E.g. how code will be highlighted if one workspace member has feature enabled and other hasn't.

ortem:
@Exidex That's an interesting question, I've added another gif to the description.

Exidex:
@ortem In gif you show two different files
(`workspace_test/one/src/main.rs` and `workspace_test/two/src/main.rs`) 
What if that feature influences same file from shared dependency (e.g. workspace member or crate)?

ortem:
@Exidex Cargo features should be additive (https://github.com/rust-lang/cargo/issues/4328), so a feature considered enabled if it is enabled in one of the packages.

Exidex:
@ortem Cool, thanks for the information

matklad:
Огонище!

src/main/kotlin/org/rust/cargo/project/workspace/FeatureGraph.kt
```diff
@@ -0,0 +1,125 @@
...
+            val node = featureToNode[feature] ?: return
+            disableFeatureTransitively(node)
+        }
+
+        private fun enableFeatureTransitively(node: FeatureNode) {
+            state[node.data] = FeatureState.Enabled
+
+            for (edge in graph.incomingEdges(node)) {
+                val dependency = edge.source
+                enableFeatureTransitively(dependency)
```
matklad, line 77: It might be a good idea to check if the feature is not already enabled:

* to avoid 2^n perf for patahlogical cases
* to avoid infinite loops for cyclic features (I think cargo either already allows cyclic feature, or there's an issue which is accepted for impl)

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -212,6 +213,87 @@ open class CargoProjectsServiceImpl(
...
+                 * - [u] foobar ["foo", "bar"]
+                 * - [u] baz = ["foo", "bar", "foobar"]
+                 *
+                 * When the user disables the feature `foobar`, the feature `baz` will be disabled transitively.
+                 * But we also should disable `bar` because the user did not enable it. So we finished with this:
+                 * - [u] foo = []
+                 * - [ ] bar = []
+                 * - [ ] foobar ["foo", "bar"]
+                 * - [ ] baz = ["foo", "bar", "foobar"]
+                 */
```
matklad, line 259: Suuuuper drive by comment, but the shape of code here feels wrong to me (which quite probably means that I just don't know the whole context).

Naively, I'd expect something like this

```Kotlin
class Features(
    enabledByTheUser: HashSet<String>,
) {
   val enabled: HashSet<String> get() {
     // computes transitive closure
   }
}
```

That is, it seems like enabling or disabling feature should *just* add it to the set of explicitelly checked features. 

Separately, there should be a thing that computes the full set of features which are activated, based on the set of features which are explicitelly selected.

Kobzol: I like this more explicit solution, but it would probably require 2 checkboxes in the UI. One read-only which says whether the feature is enabled [transitively] or not, and one writable checkbox that selects whether the user explicitly wants this feature to be enabled. I think that the solution with 2 checkboxes would generate more predictable behaviour, but it might also confuse users.

matklad: https://jetbrains.design/intellij/controls/checkbox/#14 ?

Kobzol: The states are selected, unselected but enabled and disabled. That seems like a good usecase for a three-state checkbox indeed. When the feature is enabled but not selected, clicking on it should select it and not disable it along with all super-features IMO. It could also simplify the implementation. It depends on who likes which UI design :)

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -273,32 +330,89 @@ private class WorkspaceImpl(
...
+                // Enable remained workspace features (transitively)
+                if (pkg.origin == WORKSPACE || pkg.origin == STDLIB) {
+                    for (feature in pkg.features) {
+                        if (workspaceFeatureState[feature] == FeatureState.Enabled) {
+                            enable(feature)
+                        }
+                    }
+                }
+
+                for (dependency in pkg.dependencies) {
```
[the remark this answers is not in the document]
vlad20012: Because we looking at _dependencies_, not just packages! There is some additional info in such graph edges.

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -371,6 +506,14 @@ private class PackageImpl(
 
     override val outDir: VirtualFile? by CachedVirtualFile(outDirUrl)
 
+    override val features: Set<PackageFeature> = rawFeatures.keys.mapToSet { findFeature(it) }
+
+    override val featureState: Map<String, FeatureState>
+        get() = workspace.featuresState[rootDirectory] ?: emptyMap()
+
+    override fun findFeature(name: String): PackageFeature =
```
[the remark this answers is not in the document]
vlad20012: Me too =D

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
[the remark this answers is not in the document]
ortem: Got it, thank you! IMO it'd be great to explain the logic of this code in a comment

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -198,11 +214,51 @@ private class WorkspaceImpl(
...
             pkg.cfgOptions,
             pkg.features,
+            pkg.enabledFeatures,
             pkg.env,
             pkg.outDirUrl
         )
     }
 
+    override val features: FeatureGraph by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        // TODO deduplicate PackageFeature
```
ortem, line 224: Is this comment still relevant? If so, some additional info may be provided

src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
```diff
@@ -102,10 +100,26 @@ object CargoMetadata {
...
+         */
+        val dependencies: List<RawDependency>
+    )
+
+    data class RawDependency(
+        /** A `package` name (non-normalized) of the dependency */
+        val name: String,
+        val kind: String?,
+        val target: String?,
+        val optional: Boolean,
```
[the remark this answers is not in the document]
vlad20012: The class is used for `cargo metadata` JSON parsing, we must follow its layout. In `cargo metadata` it's just `optional`.

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -216,6 +207,88 @@ open class CargoProjectsServiceImpl(
...
+                        }
+                    }
+                }
+            }.exhaustive
+        }
+    }
+
+    private fun modifyProjectFeatures(
+        cargoProject: CargoProject,
+        action: (CargoProject, CargoWorkspace, userDisabledFeatures: MutableUserDisabledFeatures) -> Unit
```
Undin, line 271: This action looks overcomplicated. You don't use `CargoProject` of `action` anywhere
Personally, I'd change the type of `action` to `(CargoProject) -> UserDisabledFeatures`

src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
```diff
@@ -92,9 +95,20 @@ interface CargoProject : UserDataHolderEx {
...
+
+    val userDisabledFeatures: UserDisabledFeatures
+
+    /**
+     * Modifies [userDisabledFeatures] that eventually affects [CargoWorkspace.Package.featureState].
+     * Note that [CargoProject] is immutable object, so this instance is not mutated by [modifyFeatures]
+     * invocation. Instead, new [CargoProject] instance is created and replaces this instance in
+     * [CargoProjectsService.allProjects].
+     */
+    fun modifyFeatures(features: Set<PackageFeature>, newState: FeatureState)
```
Undin, line 111: Do we really need it?
Maybe it's better just use `org.rust.cargo.project.model.impl.CargoProjectsServiceImpl#modifyFeatures` directly? (of course, it requires moving `modifyFeatures` into `CargoProjectsService` intefrace)

Undin (commented):
Two additional comments/questions:
* could you check that new regressions in regression checks are expected?
* it would be great to have tests for complex cases with optional dependencies, several cargo projects and some crates in workspace

src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
Undin: I don't like it because it creates the impression that it's possible to modify cargo project. And you even had to write the corresponding doc comment because of it. At the same time, it just delegates all logic to `CargoProjectsServiceImpl` so I'd prefer not to introduce potentially confusing API

vlad20012:
In tokio there are no real regressions, there are cases 2500 fixes and 10 cases like this:

Before:
image

After:
image

In amethyst there are conflicting features that should be configured correctly before testing.

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
vlad20012: Oh, let's leave `action: (CargoWorkspace, MutableUserDisabledFeatures) -> Unit`

vlad20012 (commented):
bors r+

vlad20012:
### Final feature checkboxes behavior

1. Initially all features from `workspace` packages are **enabled**
2. The state of features in external dependency packages is a _pure function_ of `workspace` features and dependency relations. A feature in external dependency packages tends to be disabled if nothing forces it to be enabled.
3. A user can disable any features in `workspace` packages. If there are other `workspace` features that depend on the feature being disabled, they are disabled too.
4. A user can enable `workspace` features again. If the feature being enabled has dependency features in `workspace` packages, they will be enabled too.

See `CargoFeaturesModificationTest` for examples.

elichai:
Thank You!

vlad20012:
Thanks to @kumbayo for the original PRs #2166 #2840 #3031 !

kumbayo:
My thanks to everybody involved in resurrecting my PRs and investing the work to bring them into reality with more functionality!
:-)

Issue #5167, opened 2020-04-01
Support blacklists in parameter name hints


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.2.120.3030-201-nightly
* **Rust toolchain version:** 1.41.0 (5e1a79984 2020-01-27) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.1 Beta Ultimate Edition (IU-201.6668.13)
* **Operating system:** macOS 10.15.3

## Problem description
In many languages (Java, Kotlin, C++, ...), parameter name hints can be disabled for specific functions using blacklists. But this feature doesn't work for Rust. Also, the blacklist information should be written for Rust (default is for Java, on the second screenshot).

To fix these problems, some methods need to be implemented in `org.rust.ide.hints.RsInlayParameterHintsProvider` (e.g. `getHintInfo`, `getBlacklistExplanationHTML`).

<img width="1065" alt="Screenshot 2020-04-01 at 12 03 46" src="https://user-images.githubusercontent.com/4854600/78120261-47bca480-7412-11ea-8229-414ca4777155.png">
<img width="547" alt="Screenshot 2020-04-01 at 12 03 51" src="https://user-images.githubusercontent.com/4854600/78120335-691d9080-7412-11ea-9dfd-eb2e7a4bdbe2.png">

PR #5195, opened 2020-04-04
Add support for blacklisting to RsInlayParameterHintsProvider

Kobzol:
Adds support for blacklisting to parameter hints. The explanation HTML was largely adapted from the Java example shown by @ortem, with some adaptations for Rust (`std` module and `_` patterns in parameters).

Questions:
1) The type hint provider has a method which shows an example window in the settings. `InlayParameterHintsProvider` has `getSettingsPreview`, but even if I implement it it doesn't show any preview in the settings. Is this the correct method to implement?
2) Is there a way to test this?
3) Should we provide some special syntax for the local crate? For example we could put `$crate::module` syntax into the blacklist and then use this path for crate-local functions.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5167

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsInlayParameterHintsProvider.kt
@@ -10,6 +10,15 @@ import com.intellij.codeInsight.hints.InlayInfo
 import com.intellij.codeInsight.hints.InlayParameterHintsProvider
 import com.intellij.codeInsight.hints.Option
 import com.intellij.psi.PsiElement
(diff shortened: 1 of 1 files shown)


ortem:
Hey @Kobzol,

1. Seems like this method is not used anywhere so far and its implementation doesn't make sense.
2. There are some tests on blacklists in the IntelliJ Platform, but I don't know any good way to test them in the plugin.
3. I'm not sure is it really necessary and how it should be implemented. But you can try doing this in another PR if you find such a feature important.

For now, I think the current state is generally OK. Thank you!

bors r+

PR #5205, opened 2020-04-07
MACRO: use common task queue with cargo sync

vlad20012:
This makes macro expansion cancellation more fast in the case of cargo refresh (e.g. Cargo.toml editing)

Also:
1. macro expansion is now cancelled immediately if the new engine is disabled in the settings;
2. progress bar of macro expansion task is not shown until the end of the indexing

### DIFF
--- a/src/main/kotlin/org/rust/RsProjectTaskQueueService.kt
+++ b/src/main/kotlin/org/rust/RsProjectTaskQueueService.kt
@@ -1,0 +1,254 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust
+
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.application.ModalityState
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.progress.*
+import com.intellij.openapi.progress.impl.BackgroundableProcessIndicator
+import com.intellij.openapi.progress.impl.ProgressManagerImpl
+import com.intellij.openapi.project.DumbService
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Condition
+import com.intellij.openapiext.isHeadlessEnvironment
+import com.intellij.openapiext.isUnitTestMode
+import com.intellij.util.PairConsumer
+import com.intellij.util.concurrency.QueueProcessor
+import org.rust.RsTask.TaskType.*
+import org.rust.openapiext.DelayedBackgroundableProcessIndicator
+import org.rust.openapiext.checkIsDispatchThread
+import org.rust.stdext.exhaustive
+import java.util.function.BiConsumer
+
+/**
+ * A common queue for cargo and macro expansion tasks that should be executed sequentially.
+ * Can run any [Task.Backgroundable], but provides additional features for tasks that implement [RsTask].
+ * The most important feature is that newly submitted tasks can cancel a currently running task or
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/RsProjectTaskQueueService.kt
```diff
@@ -0,0 +1,223 @@
...
+    val runSyncInUnitTests: Boolean
+        get() = false
+
+    /**
+     * Higher position in the enum means higher priority; Newly submitted tasks with higher or equal
+     * priority eats (cancels) other tasks with lower or equal priority if [canBeEaten] == true.
+     * E.g. [CARGO_SYNC] eats [MACROS_UNPROCESSED] but not [MACROS_CLEAR] or itself.
+     * [MACROS_UNPROCESSED] eats itself, [MACROS_FULL] and subsequent.
+     */
+    enum class EatStrategy(val canBeEaten: Boolean = true) {
```
Undin, line 71: Looks like it's not a strategy, it's a task type.
Also, maybe rename `canBeEaten` to `canBeCosumed`?

vlad20012:
I found that the current implementation leads to the fact that some macro expansion tasks are submitted to the queue before cargo sync tasks. The reason is that MacroExpansionManage is a component (not a service) and is loaded prior to CargoProjectService. 

Let's postpone it until MacroExpansionManage becomes a service

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -558,11 +548,10 @@ private fun fetchCargoWorkspace(
 
 private fun fetchRustcInfo(
     project: Project,
-    queue: BackgroundTaskQueue,
     toolchain: RustToolchain,
     projectDirectory: Path
 ): CompletableFuture<TaskResult<RustcInfo>> {
-    return runAsyncTask(project, queue, "Getting toolchain version") {
+    return runAsyncTask(project, project.rsTaskQueue::run, "Getting toolchain version") {
```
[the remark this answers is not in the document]
vlad20012: I hope, we'll introduce `CargoSyncTask` in future and will remove `runAsyncTask` at all

src/main/kotlin/org/rust/RsProjectTaskQueueService.kt
```diff
@@ -0,0 +1,254 @@
...
+                    return
+                }
+                State.Canceled -> {
+                    continuation.run()
+                    return
+                }
+                is State.Running -> error("Trying to re-run already running task")
+            }
+
+            if (task is RsTask && task.waitForSmartMode && DumbService.getInstance(task.project).isDumb) {
```
Undin, line 202:
```suggestion
            if (task is RsTask && task.waitForSmartMode && DumbService.isDumb(task.project)) {
```

src/main/kotlin/org/rust/openapiext/AsyncTasks.kt
```diff
@@ -23,10 +23,14 @@ interface AsyncTaskCtx<T> {
     fun ok(value: T) = TaskResult.Ok(value)
 }
 
-fun <T> runAsyncTask(project: Project, queue: BackgroundTaskQueue, title: String,
-                     task: AsyncTaskCtx<T>.() -> TaskResult<T>): CompletableFuture<TaskResult<T>> {
+fun <T> runAsyncTask(
+    project: Project,
+    queue: (Task.Backgroundable) -> Unit,
```
Undin, line 28: Maybe `enqueue` is better name now due to the change in argument type?

vlad20012:
bors r=undin

PR #5218, opened 2020-04-09
REPL: Fix imports

### LINKED ISSUE #5214: Imports not working in REPL console
## Environment

* **IntelliJ Rust plugin version:** 0.2.119.2172-193
* **Rust toolchain version:** 1.42.0
* **IDE name and version:** IntelliJ IDEA 2019.3.3
* **Operating system:** Linux

## Problem description
When execute import command (`use ...;`) in REPL console as separate command, it doesn't affect completion

image

## Steps to reproduce
* Execute command `use std::fs;` in REPL console
* Type `fs::` and invoke completion

dima74:
Fix #5214

Previously only `RsNamedElement` and `RsStmt` were added to context, so imports were ignored

### DIFF
--- a/src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
+++ b/src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
@@ -8,44 +8,101 @@ package org.rust.ide.console
 import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.project.DumbService
 import com.intellij.openapi.project.Project
+import com.intellij.util.containers.ContainerUtil
 import org.rust.cargo.project.model.cargoProjects
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.descendantOfTypeStrict
+import org.rust.lang.core.psi.ext.block
+import org.rust.lang.core.psi.ext.childOfType
+import org.rust.lang.core.resolve.TYPES
+import org.rust.lang.core.resolve.findPrelude
+import org.rust.lang.core.resolve.processNestedScopesUpwards
 import org.rust.openapiext.toPsiFile
 
-class RsConsoleCodeFragmentContext {
+class RsConsoleCodeFragmentContext(codeFragment: RsReplCodeFragment?) {
 
-    private val topLevelElements: MutableMap<String, String> = mutableMapOf()
-    private val allCommandsText: StringBuilder = StringBuilder()
+    // concurrent set is needed because accessed both in `addItemsNamesFromPrelude` and `addToContext`
+    private val itemsNames: MutableSet<String> = ContainerUtil.newConcurrentSet()
+    @Volatile
+    private var hasAddedNamesFromPrelude = false
 
-    fun addToContext(lastCommandContext: RsConsoleOneCommandContext) {
-        topLevelElements.putAll(lastCommandContext.topLevelElements)
-        allCommandsText.append(lastCommandContext.statementsText)
-    }
+    private val commands: MutableList<String> = mutableListOf()
 
-    fun updateContext(project: Project, codeFragment: RsReplCodeFragment) {
-        val allCommandsText = getAllCommandsText()
-
-        DumbService.getInstance(project).smartInvokeLater {
-            runWriteAction {
-                codeFragment.context = createContext(project, codeFragment.crateRoot as RsFile?, allCommandsText)
+    init {
+        if (codeFragment != null) {
+            DumbService.getInstance(codeFragment.project).runWhenSmart {
+                addItemsNamesFromPrelude(codeFragment)
             }
         }
     }
 
+    // see org.rust.ide.console.RsConsoleCompletionTest.`test redefine struct from prelude`
+    private fun addItemsNamesFromPrelude(codeFragment: RsReplCodeFragment) {
+        val prelude = findPrelude(codeFragment) ?: return
+
+        val preludeItemsNames = mutableListOf<String>()
+        processNestedScopesUpwards(prelude, TYPES) {
+            preludeItemsNames += it.name
+            false
+        }
+        itemsNames += preludeItemsNames
+        hasAddedNamesFromPrelude = true
+    }
+
+    fun addToContext(lastCommandContext: RsConsoleOneCommandContext) {
+        if (commands.isEmpty()
+            || lastCommandContext.itemsNames.any(itemsNames::contains)
+            || lastCommandContext.containsUseDirective
+            || !hasAddedNamesFromPrelude
+        ) {
+            commands.add(lastCommandContext.command)
+        } else {
+            commands[commands.size - 1] = commands.last() + "\n" + lastCommandContext.command
+        }
+        itemsNames += lastCommandContext.itemsNames
+    }
+
+    fun updateContextAsync(project: Project, codeFragment: RsReplCodeFragment) {
+        DumbService.getInstance(project).smartInvokeLater {
+            runWriteAction {
+                updateContext(project, codeFragment)
+            }
+        }
+    }
+
+    fun updateContext(project: Project, codeFragment: RsReplCodeFragment) {
+        codeFragment.context = createContext(project, codeFragment.crateRoot as RsFile?, commands)
+    }
+
     fun getAllCommandsText(): String {
-        val topLevelElementsText = topLevelElements.values.joinToString("\n")
-        return topLevelElementsText + "\n" + allCommandsText
+        return commands.joinToString("\n")
     }
 
     companion object {
-        fun createContext(project: Project, originalCrateRoot: RsFile?, allCommandsText: String = ""): RsBlock {
-            val rsFile = RsPsiFactory(project).createFile("fn main() { $allCommandsText }")
+        fun createContext(project: Project, originalCrateRoot: RsFile?, commands: List<String> = listOf("")): RsBlock {
+            // command may contain functions/structs with same names as in previous commands
+            // therefore we should put such commands in separate scope
+            // we do this like so:
+            // ```
+            // command1;
+            // {
+            //     command2;
+            //     {
+            //         command3;
+            //     }
+            // }
+            // And `RsBlock` surrounding `command3` will become context of codeFragment
+            // ```
+            val functionBody = commands.joinToString("\n{\n") + "\n}".repeat(commands.size - 1)
+            val rsFile = RsPsiFactory(project).createFile("fn main() { $functionBody }")
 
             val crateRoot = originalCrateRoot ?: findAnyCrateRoot(project)
             crateRoot?.let { rsFile.originalFile = crateRoot }
 
-            return rsFile.descendantOfTypeStrict()!!
+            val functionBlock = rsFile.childOfType<RsFunction>()!!.block!!
+            val blocks = generateSequence(functionBlock) {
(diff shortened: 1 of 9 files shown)


dima74:
Added some tests (and rebased on master)

src/main/kotlin/org/rust/lang/core/psi/RsCodeFragment.kt
```diff
@@ -162,4 +162,5 @@ class RsReplCodeFragment(fileViewProvider: FileViewProvider, override var contex
     val tailExpr: RsExpr?
```
ortem, line 162: I'd suggest go the `org.rust.lang.core.psi.ext.RsBlockKt#getExpandedStmtsAndTailExpr` way here. It can be done even without macro expansion usage (but would be nice to use expanded macros in REPL someday), like this:
```Kotlin
val expandedStmtsAndTailExpr: Pair<List<RsExpandedElement>, RsExpr?>
    get() {
        val expandedElements = childrenOfType<RsExpandedElement>()
        val tailExpr = expandedElements.lastOrNull()?.let { it as? RsExpr }
        return when (tailExpr) {
            null -> expandedElements
            else -> expandedElements.subList(0, expandedElements.size - 1)
        } to tailExpr
    }
```
Then you can use `getExpandedStmtsAndTailExpr` instead of `stmts`, `tailExpr`, and `items` in type inference/name resolution and `org.rust.ide.console.RsConsoleOneCommandContext`.

Or maybe you will come up with another way of doing this, it'd be OK too.

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -57,15 +57,21 @@ class RsConsoleCodeFragmentContext {
 }
 
 class RsConsoleOneCommandContext(codeFragment: RsReplCodeFragment) {
-    val topLevelElements: MutableMap<String, String> = mutableMapOf()
+    val topLevelItems: MutableMap<String, String> = mutableMapOf()
```
ortem, line 60: Do we really want to split all elements into "items" and "statements"?
Probably it may lead to bugs because order matters:
```Rust
use foo;
stmt_1;
use bar;
stmt_2;
```
can be semantically different from
```Rust
use foo;
use bar;
stmt_1;
stmt_2;
```
But maybe I'm wrong and it is OK, what do you think?

dima74: Originally we split elements into items and statements, because we want to do something with items which have same name (they are not allowed in rust functions, but are allowed in evcxr repl). This approach can indeed lead to some bugs (please see new tests which check redefining function/struct). I remove that splitting and now we use nested scopes for next commands, like so:

```rust
fn main() {
    command1;
    {
        command2;
        {
            command3;
        }
    }
}
```
And `RsBlock` surrounding last command (`command3`) will become context of codeFragment

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -57,15 +57,21 @@ class RsConsoleCodeFragmentContext {
...
     val statementsText: String
 
     init {
-        for (namedElement in codeFragment.namedElements) {
-            val name = namedElement.name ?: continue
-            topLevelElements[name] = namedElement.text
+        val namedItems = codeFragment.items
```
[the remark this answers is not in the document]
dima74: Code was partially rewritten, and `items` is no longer used

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -10,42 +10,61 @@ import com.intellij.openapi.project.DumbService
...
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.descendantOfTypeStrict
+import org.rust.lang.core.psi.ext.*
 import org.rust.openapiext.toPsiFile
 
 class RsConsoleCodeFragmentContext {
 
-    private val topLevelElements: MutableMap<String, String> = mutableMapOf()
-    private val allCommandsText: StringBuilder = StringBuilder()
+    private val itemsNames: Set<String> = mutableSetOf()
```
ortem, line 18: Looks like `itemsNames` is always empty, so `commands` contains only one element and nested scopes are not used later

dima74: Thanks! Should be fixed now. Also I updated tests, so now they behave closer to real REPL

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -10,42 +10,87 @@ import com.intellij.openapi.project.DumbService
...
+            // ```
+            val functionBody = commands.joinToString("\n{\n") + "\n}".repeat(commands.size - 1)
+            val rsFile = RsPsiFactory(project).createFile("fn main() { $functionBody }")
 
             val crateRoot = originalCrateRoot ?: findAnyCrateRoot(project)
             crateRoot?.let { rsFile.originalFile = crateRoot }
 
-            return rsFile.descendantOfTypeStrict()!!
+            var block: RsBlock = rsFile.childOfType<RsFunction>()!!.block!!
+            repeat(commands.size - 1) {
```
ortem, line 89: Consider using `generateSequence` instead

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -57,15 +102,17 @@ class RsConsoleCodeFragmentContext {
...
 
     init {
-        for (namedElement in codeFragment.namedElements) {
-            val name = namedElement.name ?: continue
-            topLevelElements[name] = namedElement.text
-        }
+        val elements = codeFragment.expandedStmtsAndTailExpr.first
+        command = elements.joinToString("\n") { it.text }
+
+        itemsNames = elements.filterIsInstance<RsNamedElement>()
```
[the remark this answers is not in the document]
dima74: I think we don't need uniqueness here, code in `RsConsoleCodeFragmentContext` would work correctly if there are duplicates. Also if there are items with same name, then command will not be executed successfully, so `RsConsoleOneCommandContext` will not be used

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -57,15 +102,17 @@ class RsConsoleCodeFragmentContext {
 }
 
 class RsConsoleOneCommandContext(codeFragment: RsReplCodeFragment) {
-    val topLevelElements: MutableMap<String, String> = mutableMapOf()
-    val statementsText: String
+    val command: String
+    val itemsNames: List<String>
```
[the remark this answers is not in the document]
dima74: I am not sure that we need `Set` here:
* we don't need fast lookups (because we don't do lookups)
* we don't need uniqueness (actually if command is executed successfully then all names should already be unique)

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -10,42 +10,98 @@ import com.intellij.openapi.project.DumbService
...
+import org.rust.lang.core.resolve.findPrelude
+import org.rust.lang.core.resolve.processNestedScopesUpwards
 import org.rust.openapiext.toPsiFile
 
-class RsConsoleCodeFragmentContext {
+class RsConsoleCodeFragmentContext(codeFragment: RsReplCodeFragment?) {
 
-    private val topLevelElements: MutableMap<String, String> = mutableMapOf()
-    private val allCommandsText: StringBuilder = StringBuilder()
+    private val itemsNames: MutableSet<String> = mutableSetOf()
```
ortem, line 23: Maybe use `ContainerUtil.newConcurrentSet()` and get rid of synchronized `addItemsNames`?

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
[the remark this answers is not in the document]
ortem: It just looks a bit confusing that there are two cases when all named elements are needed, but they are collected in different ways. Probably there could be something like:
```Kotlin
val namedElementsUnique: Map<String, RsNamedElement>
    get() = expandedStmtsAndTailExpr.first  // to take expanded named elements into account when you implement processing macros in `expandedStmtsAndTailExpr`
        .filterIsInstance<RsNamedElement>()
        .filter { it.name != null }
        .associateBy { it.name!! }
```
So this property can be used when named elements or just their names are needed. But this can be done later.

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -10,42 +10,98 @@ import com.intellij.openapi.project.DumbService
...
+    private val commands: MutableList<String> = mutableListOf()
 
-    fun addToContext(lastCommandContext: RsConsoleOneCommandContext) {
-        topLevelElements.putAll(lastCommandContext.topLevelElements)
-        allCommandsText.append(lastCommandContext.statementsText)
+    init {
+        if (codeFragment != null) {
+            DumbService.getInstance(codeFragment.project).runWhenSmart {
```
ortem, line 28: Looks like the user can redefine some items from prelude before this code is invoked. I mean completion may be invalid in such case:
```Rust
>> let foo = Some(1);
>> struct Some(i32, i32);
// indexing ends, `addItemsNamesFromPrelude` is invoked
>> foo./*caret*/
```

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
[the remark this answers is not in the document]
dima74: Ok, I agree that it is confusing, so updated code as you proposed

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
dima74: Thanks, fixed by forcing creating new scope if names for items from prelude are not yet added

src/main/kotlin/org/rust/ide/console/RsConsoleCodeFragmentContext.kt
```diff
@@ -8,44 +8,100 @@ package org.rust.ide.console
...
 import org.rust.openapiext.toPsiFile
 
-class RsConsoleCodeFragmentContext {
+class RsConsoleCodeFragmentContext(codeFragment: RsReplCodeFragment?) {
 
-    private val topLevelElements: MutableMap<String, String> = mutableMapOf()
-    private val allCommandsText: StringBuilder = StringBuilder()
+    // concurrent set is needed because accessed both in `addItemsNamesFromPrelude` and `addToContext`
+    private val itemsNames: MutableSet<String> = ContainerUtil.newConcurrentSet()
+    private var hasAddedNamesFromPrelude = false
```
ortem, line 25: Don't forget about `@Volatile`

dima74:
bors r=ortem

PR #5223, opened 2020-04-10
REPL: Fix parsing top-level statements

### LINKED ISSUE #5216: REPL cannot parse top-level statements
## Environment

* **IntelliJ Rust plugin version:** 0.2.120.3031-201-nightly
* **Rust toolchain version:** 1.42.0-nightly (f43c34a13 2020-02-02) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.1 Beta Ultimate Edition (IU-201.6668.13)
* **Operating system:** macOS 10.15.3

## Problem description
REPL cannot parse top-level statements and marks them as a syntax error.

## Steps to reproduce
Enter `1;` or any simple statement

Screenshot 2020-04-09 at 15 17 30

dima74:
Fix #5216

### DIFF
--- a/src/main/grammars/RustParser.bnf
+++ b/src/main/grammars/RustParser.bnf
@@ -126,7 +126,7 @@ StatementCodeFragmentElement ::= Stmt?
 TypeReferenceCodeFragmentElement ::= TypeReference?
 ValuePathCodeFragmentElement ::= ValuePathGenericArgs?
 TypePathCodeFragmentElement ::= TypePathGenericArgs?
-ReplCodeFragmentElement ::= (Item | Stmt)* Expr?
+ReplCodeFragmentElement ::= ReplBlockElement*
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 // Attributes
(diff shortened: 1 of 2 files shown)


dima74:
Added some tests (and rebased on master)

src/main/grammars/RustParser.bnf
```diff
@@ -1524,6 +1524,20 @@ LetDecl ::= OuterAttr* let Pat TypeAscription? [ '=' AnyExpr ] ';' {
 
 EmptyStmt ::= ';' { extends = Stmt }
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// REPL
+///////////////////////////////////////////////////////////////////////////////////////////////////
+
+// copy of BlockElement with '}' replaced to <<eof>>
```
ortem, line 1531: I'm not really sure why `<<eof>>` is needed here, but maybe it is. @dima74 @vlad20012 what do you think?

src/test/kotlin/org/rust/ide/console/RsConsoleCompletionTest.kt
```diff
@@ -93,4 +93,34 @@ class RsConsoleCompletionTest : RsConsoleCompletionTestBase() {
...
+    fun `test nested`() = checkContainsCompletion("""
+        fn nop() {}
+        let var1 = 1;
+        let var2 = 2;
+    """, """
+        nop();
+        fn foo() {
+            if true {
+                for i in 0..10 {
+                    let x = var/*caret*/
```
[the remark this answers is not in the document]
dima74: I think Evcxr is correct here, because function has its own scope and [can't access variables from outer function](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a72256e6bd3a4bbe97a3eb6f4388d165)

src/main/grammars/RustParser.bnf
dima74: Looks like `<<eof>>` is not needed in `ReplBlockElement`. But it is needed in `ReplExprStmtOrLastExpr` (otherwise tail expression will not be parsed correctly)

vlad20012: It's ok for `ReplExprStmtOrLastExpr`

PR #5228, opened 2020-04-11
GRD: Update to release version

Stzx:
Upgrade `EAP` version to `release` version


Undin:
We will update IDE dependencies to release versions when all of them release. Right now CLion 2020.1 is not released yet

PR #5245, opened 2020-04-16
MACRO: turn MacroExpansionManager into a service

vlad20012:
Also:
- `MacroExpansionManager` is not created for non-rust projects. Still will be created if you open `.rs` file
- Simplified `MacroExpansionManager` loading. No more Futures and awaits.
- Fix `RsMacroExpansionCachingToolchainTest` test on 2020.1 and on java 11.

Part of #4832

PR #5251, opened 2020-04-17
MACRO: save MacroExpansionManager asynchronously

### LINKED ISSUE #5198: Clion freezes when saving file if experimental macro expansion is enabled
## Environment

* **IntelliJ Rust plugin version:** 0.2.119.2172-193
* **Rust toolchain version:** 1.42.0
* **IDE name and version:** Clion 2019.3.5
* **Operating system:** MacOS 10.15.3

## Problem description

Whenever I save a file (either via Cmd-S or via a menu item), IDE freezes for a few hundred milliseconds. It becomes completely unresponsive, ignoring all keystrokes, mouse clicks and scrolling. When it comes back to normal, it processes all keystrokes that've been sent to it. I.e. if I press Cmd-S and hit Enter for a few times, I'll see newlines added to a file, but with some delay.

The issue appears in any file with any extension, even in scratch txt files, even in non-rust projects.

The issue only appears if experimental macro expansion is enabled. Other IDE and Rust plugin settings does not seem to affect it (I've tried disabling safe write, running clippy on the fly, running rustfmt on save, watching cargo.toml).

## Steps to reproduce

There are no particular steps to reproduce it, the issue happens all the time I save a file.

I've captured a [CPU snapshot], in which I edit a scratch file. I hit Cmd-S and then start hitting Enter until IDE unfreezes.

[CPU snapshot]: https://drive.google.com/file/d/1qwjd1eiwaOpbRY2zmH4jo7L515n8ADp5/view?usp=sharing

vlad20012:
Fixes UI freezes during project save. 
Fixes #5198

Depends on #5245

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/MacroExpansionManager.kt
@@ -37,10 +37,10 @@ import com.intellij.util.concurrency.QueueProcessor
 import com.intellij.util.concurrency.QueueProcessor.ThreadToUse
 import com.intellij.util.indexing.FileBasedIndex
 import com.intellij.util.indexing.IndexableFileSet
-import com.intellij.util.io.createDirectories
-import com.intellij.util.io.createFile
-import com.intellij.util.io.delete
-import com.intellij.util.io.exists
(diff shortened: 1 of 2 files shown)


vlad20012:
bors r=undin

Issue #5190, opened 2020-04-03
Feature request: chain hints


p0lunin:
Add hints to code which contains some chains.
Example:
```rust
let foo: Result<Option<i32>, &str> = Ok(Some(10i32));
foo
  .map_err(|_| 0_u32) [Result<Option<i32>, u32>]
  .unwrap() [Option<i32>]
  .and_then(|_| Some("foo")) [Option<&str>]
  .unwrap();
```
In square brackets placed type hints which plugin can infer.
Same functionality provides rust-analyzer.

PR #5276, opened 2020-04-23
HINT: add hints for chained method calls

Kobzol:
This PR adds support for chained method hints.

image

Questions:
1) ~I copied the new (201) inlay tests from `RsInlayTypeHintsProviderTestBase`, should I also add support for 193?~ Turns out that CI won't pass without it, so I added it. Also I'm not sure that I understand the syntax for writing hint tests :-)
2) The preview text in settings does not seem to be affected by the hints. Maybe it misses the std types, can they be added somehow or do I have to prepare a standalone preview without std?
3) Should I add some settings?
4) I'm a bit confused about the difference between `RsPlainHint` and the various `InlayHintsProvider`s. Should I add support for this hint to `RsPlainHint`?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5190

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
@@ -1,0 +1,138 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+@file:Suppress("UnstableApiUsage")
+
+package org.rust.ide.hints
+
+import com.intellij.codeInsight.hints.*
+import com.intellij.codeInsight.hints.presentation.InlayPresentation
+import com.intellij.codeInsight.hints.presentation.InsetPresentation
+import com.intellij.codeInsight.hints.presentation.MenuOnClickPresentation
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.DumbService
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiComment
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.ui.components.CheckBox
+import com.intellij.ui.layout.panel
+import com.intellij.util.ui.JBUI
+import org.rust.ide.utils.isEnabledByCfg
+import org.rust.lang.RsLanguage
+import org.rust.lang.core.psi.RsDotExpr
+import org.rust.lang.core.psi.RsMethodCall
+import org.rust.lang.core.psi.ext.childOfType
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.lang.core.psi.ext.parentDotExpr
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.type
+import javax.swing.JPanel
+
+class RsChainMethodTypeHintsProvider : InlayHintsProvider<RsChainMethodTypeHintsProvider.Settings> {
+    override val key: SettingsKey<Settings> get() = KEY
+
+    override val name: String get() = "Chain method hints"
+
+    override val previewText: String? = null
+
+    override fun createConfigurable(settings: Settings): ImmediateConfigurable = object : ImmediateConfigurable {
+        private val consecutiveField = CheckBox("Show same consecutive types")
+
+        override fun createComponent(listener: ChangeListener): JPanel {
+            consecutiveField.isSelected = settings.showSameConsecutiveTypes
+            consecutiveField.addItemListener { handleChange(listener) }
+
+            val panel = panel {
+                row { consecutiveField(pushX) }
+            }
+            panel.border = JBUI.Borders.empty(5)
+            return panel
+        }
+
+        private fun handleChange(listener: ChangeListener) {
+            settings.showSameConsecutiveTypes = consecutiveField.isSelected
+            listener.settingsChanged()
+        }
+    }
+
+    override fun createSettings(): Settings = Settings()
+
+    override fun getCollectorFor(file: PsiFile, editor: Editor, settings: Settings, sink: InlayHintsSink): InlayHintsCollector? =
+        object : FactoryInlayHintsCollector(editor) {
+
+            val typeHintsFactory = RsTypeHintsPresentationFactory(factory, true)
+
+            override fun collect(element: PsiElement, editor: Editor, sink: InlayHintsSink): Boolean {
+                if (DumbService.isDumb(element.project)) return true
+                if (element !is RsMethodCall) return true
+                if (!element.isLastInChain) return true
+                if (!element.isEnabledByCfg) return true
+
+                val chain = collectChain(element)
+                var lastType: Ty? = null
+                for (call in chain.dropLast(1)) {
+                    if (call.type != TyUnknown && call.isLastOnLine) {
+                        if (settings.showSameConsecutiveTypes || call.type != lastType) {
+                            presentTypeForMethodCall(call)
+                        }
+                        lastType = call.type
+                    }
+                }
+
+                return true
+            }
+
+            private fun presentTypeForMethodCall(call: RsMethodCall) {
+                val project = call.project
+                val type = call.type
+                val presentation = typeHintsFactory.typeHint(type)
+                val finalPresentation = presentation.withDisableAction(project)
+                sink.addInlineElement(call.endOffset, true, finalPresentation)
+            }
+        }
+
+    private fun InlayPresentation.withDisableAction(project: Project): InsetPresentation = InsetPresentation(
+        MenuOnClickPresentation(this, project) {
+            listOf(InlayProviderDisablingAction(name, RsLanguage, project, key))
+        }, left = 1
+    )
+
+    data class Settings(var showSameConsecutiveTypes: Boolean = true)
+
+    companion object {
+        val KEY: SettingsKey<Settings> = SettingsKey("chain-method.hints")
+    }
+}
+
+private val RsMethodCall.isLastInChain: Boolean
+    get() = parentDotExpr.parent !is RsDotExpr && parentDotExpr.expr.childOfType<RsMethodCall>() != null
+
+private val RsMethodCall.isLastOnLine: Boolean
+    get() = this.parentDotExpr.isLastOnLine
+
+private val PsiElement.isLastOnLine: Boolean
+    get() {
+        return when (val sibling = this.nextSibling) {
+            is PsiWhiteSpace -> sibling.textContains('\n') || sibling.isLastOnLine
+            is PsiComment -> sibling.isLastOnLine
+            else -> false
+        }
+    }
+
+private val RsMethodCall.type: Ty
+    get() = parentDotExpr.type
+
+
+private fun collectChain(call: RsMethodCall): List<RsMethodCall> {
+    val chain = mutableListOf<RsMethodCall>()
+    var current = call
+    while (true) {
+        chain.add(current)
+        current = current.parentDotExpr.expr.childOfType() ?: break
+    }
+    return chain.reversed()
+}
--- a/src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
+++ b/src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
@@ -1,0 +1,128 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.hints
+
+import com.intellij.codeInsight.hints.InlayHintsSettings
+import org.intellij.lang.annotations.Language
+import org.rust.ProjectDescriptor
+import org.rust.WithStdlibRustProjectDescriptor
+import org.rust.lang.RsLanguage
+
+class RsChainMethodTypeHintsProviderTest : RsInlayTypeHintsTestBase() {
+    private val types = """
+        struct A;
+        struct B;
+
+        impl A {
+            fn clone(&self) -> A { A }
+            fn change(&self) -> B { B }
+        }
+
+        impl B {
+            fn clone(&self) -> B { B }
+            fn change(&self) -> A { A }
+        }
+    """
+
+    @ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
+    fun `test result and option`() = doTest("""
+        fn main() {
+            let foo: Result<Option<i32>, &str> = Ok(Some(10i32));
+            foo
+              .map_err(|_| 0_u32)/*hint text="[:  [Result [< [[Option [< i32 >]] ,  u32] >]]]"*/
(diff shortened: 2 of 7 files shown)


sirgl:
2. Any cross-file functionality not working in preview. In such cases you can just return null.

From my side I can say that chain hints are not very useful when they are the same across all the chain. You may check https://github.com/JetBrains/intellij-community/blob/master/java/java-impl/src/com/intellij/codeInsight/hints/MethodChainsInlayProvider.kt on how it is done for Java.

src/193/test/kotlin/org/rust/ide/hints/RsChainMethodHintsProviderTest.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.intellij.codeInsight.hints.presentation.PresentationRenderer
+
+class RsChainMethodHintsProviderTest : RsChainMethodHintsProviderTestBase() {
+    override fun checkInlays() {
+        myFixture.testInlays(
+            { (it.renderer as PresentationRenderer).presentation.toString() },
+            { it.renderer is PresentationRenderer }
+        )
+    }
```
Undin, line 16: Let's make some refactoring in hint tests to avoid code duplication
For example, you can create the following hierarchy:
`RsInlayTypeHintsTestPlatformBase` with single protected method `checkInlays`. Of course, you need two versions of it because of slightly different API in 2019.3 and 2020.
`RsInlayTypeHintsTestBase : RsInlayTypeHintsTestPlatformBase` with protected method `checkByText`. Right now we have two identical versions of it in `RsInlayTypeHintsProviderTestBase` and `RsChainMethodHintsProviderTestBase `
`RsInlayTypeHintsProviderTest : RsInlayTypeHintsTestBase` with all tests related to type hints (right now the corresponding class named `RsInlayTypeHintsTestBase`)
`RsChainMethodHintsProviderTest : RsInlayTypeHintsTestBase` with tests related to method chain hints
If you have ideas how to make it simpler and suggestions about better names - welcome

Kobzol:
For me the preview does not seem to be working at all, even if I use custom types, and even the preview for the normal type hints stopped working for me for some reason once I added the chain hints. Also (probably) my commits somehow broke a test in the normal type provider, although I did not modify that at all..

I rewrote the hinting logic to find the last element in a call chain, then traverse the chain and add hints if a type change happens. Currently the last method in the chain does not have hints (as they could be duplicated with hints in a `let` declaration). Maybe an option to enable the hint for the last method in a chain could be added into the settings?

Multiple calls on a single line look like this:
image
I like it better than the potential alternative, which would be something like
```rust
foo.map_err().unwrap() :Result<Option<i32>, u32>> :Option<i32>
```

I refactored the 193/201 specific code.

Kobzol:
For some reason, the preview cannot handle code that is otherwise type inferred and hinted correctly in a normal file.
```rust
struct A;
impl A {
    fn clone(self) -> A {
        self
    }
    fn change(self) -> B {
        B
    }
}

struct B;
impl B {
    fn clone(self) -> B {
        self
    }
}

fn foo(value: A) -> B {
    value.clone() // :A
        .clone()
        .change() // :B
        .clone()
}
```
This code should be completely standalone. In a normal file, it shows the hints, in preview it only shows `:?`. Type inference seems to be broken in the preview for some reason.

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,115 @@
...
+    override fun createSettings(): Settings = Settings()
+
+    override fun getCollectorFor(file: PsiFile, editor: Editor, settings: Settings, sink: InlayHintsSink): InlayHintsCollector? =
+        object : FactoryInlayHintsCollector(editor) {
+
+            val typeHintsFactory = RsTypeHintsPresentationFactory(factory, true)
+
+            override fun collect(element: PsiElement, editor: Editor, sink: InlayHintsSink): Boolean {
+                if (DumbService.getInstance(file.project).isDumb) return true
+                if (element !is RsElement) return true
```
Undin, line 60:
```suggestion
                if (element !is RsMethodCall) return true
```
You need only method calls so you can check the necessary condition directly

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,115 @@
...
+    private fun InlayPresentation.withDisableAction(project: Project): InsetPresentation = InsetPresentation(
+        MenuOnClickPresentation(this, project) {
+            listOf(InlayProviderDisablingAction(name, RsLanguage, project, key))
+        }, left = 1
+    )
+
+    class Settings
+
+    companion object {
+        private val KEY: SettingsKey<Settings> = SettingsKey("rust.chain-method.hints")
```
Undin, line 95: `rust` prefix looks redundant - `SettingsKey` add it itself. See SettingsKey#getFullId`

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,115 @@
...
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.type
+import javax.swing.JComponent
+
+class RsChainMethodTypeHintsProvider : InlayHintsProvider<RsChainMethodTypeHintsProvider.Settings> {
+    override val key: SettingsKey<Settings> get() = KEY
+
+    override val name: String get() = "Chain method hints"
+
+    override val previewText: String
```
Undin, line 35: Preview doesn't work because our type inference doesn't work properly for dummy files.
I think we can omit preview for now, make some changes in type inference and after that enable it

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+
+    private fun InlayPresentation.withDisableAction(project: Project): InsetPresentation = InsetPresentation(
+        MenuOnClickPresentation(this, project) {
+            listOf(InlayProviderDisablingAction(name, RsLanguage, project, key))
+        }, left = 1
+    )
+
+    class Settings
```
Undin, line 92: Let's add settings similar to `Minimal unique type count to show hints` Java setting

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,115 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.hints
```
Undin, line 6:
```suggestion
@file:Suppress("UnstableApiUsage")

package org.rust.ide.hints
```

Warnings about unstable API are quite annoying

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
Kobzol: `RsInlayTypeHintsProvider` uses `"rust.type.hints"`, that's why I have included it. Should I remove the prefix?

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
Kobzol: I'm not sure if we should copy the specific behaviour of the Java setting, IMO it behaves in a weird way, at least for Rust. For example with minimal unique types = 1 (which is the lowest value that you can set it to):
```java
a
.b()
.c();
```
This doesn't show any hints. However if `b()` returned some complex type, I think that it would still be nice to show it in Rust. In Java it's mostly for builders where the types are usually obvious, while in Rust I feel like the types are often pretty complicated and you can't even name them by yourself, so you do things like assigning them into variables of a different type just to see what the real type is from the compiler error.

Also when a type repeats it suddently stops showing types:
```java
a
 .a()
 .b()
 .b()
 .c();
```

We already ignore repeated types, so some of the noise will be reduced. Maybe we could include a setting to only hint a specific type once in a specific chain? Like this:
```rust
// include_once=false
a
.a() // hint
.b() // hint
.a() // hint
.b() // hint

// include_once=true
a
.a() // hint
.b() // hint
.a()
.b()
```

Kobzol: Maybe we can also ignore obvious stuff like `.clone()`? But that would be pretty brittle I think.

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
Undin: Yep, it was a mistake but if you remove prefix in `RsInlayTypeHintsProvider ` it will affect all users who changed the corresponding settings so I think we shouldn't do it

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
Undin: But I understood that we should have another option. Right now our implementation doesn't show type of method call if it's the same as previous type.
At the same time, we want to hide type hints for unknown types. Since our name resolution and type inference are not ideal (unfortunately), sometimes we can get a situation when it's almost impossible from the user side to distinguish "same as previous type" and "unknown type" cases.

So let's add an option to show consecutive same method types and enable it by default - it will help us to collect a lot of feedback about our type inference. After we implemented type hints, we started to get much more reports about problems in our name resolution/type inference

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    class Settings
+
+    companion object {
+        private val KEY: SettingsKey<Settings> = SettingsKey("rust.chain-method.hints")
+    }
+}
+
+private val RsMethodCall.isLastInChain
```
Undin, line 91: Could you add explicit returning type for all new methods and properties?

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,109 @@
...
+            override fun collect(element: PsiElement, editor: Editor, sink: InlayHintsSink): Boolean {
+                if (DumbService.getInstance(file.project).isDumb) return true
+                if (element !is RsMethodCall) return true
+                if (!element.isLastInChain) return true
+                if (!element.isEnabledByCfg) return true
+
+                val chain = collectChain(element)
+                var lastType: Ty? = null
+                for (call in chain.dropLast(1)) {
+                    if (call.type != lastType && call.isLastOnLine) {
```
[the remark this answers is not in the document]
vlad20012: I think `!=` is ok in this case

Kobzol:
I added an option to show consecutive same types. Is there some way to control the hint's settings in unit tests? Or should I add some mock that will be used in unit tests to test the setting both on and off?

Comments and whitespace at the end of the line is now handled too. Is [this](https://github.com/intellij-rust/intellij-rust/pull/5276/commits/308cf387464669dcf43fe50bcf2c2ca3868cf64e#diff-a0f94161207caf4204b7980e6a4a1d90R119) safe? I assume that there are no cycles between the siblings.

Undin:
> Is there some way to control the hint's settings in unit tests?

I suppose you may use `com.intellij.codeInsight.hints.InlayHintsSettings#storeSettings` to set desired settings. But don't forget to restore default settings on the end of test not to affect other tests

Kobzol:
Thanks, I used this API to test both variants of the settings. I also rebased onto master to fix merge conflicts.

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,140 @@
...
+
+class RsChainMethodTypeHintsProvider : InlayHintsProvider<RsChainMethodTypeHintsProvider.Settings> {
+    override val key: SettingsKey<Settings> get() = KEY
+
+    override val name: String get() = "Chain method hints"
+
+    override val previewText: String? = null
+
+    override fun createConfigurable(settings: Settings): ImmediateConfigurable = object : ImmediateConfigurable {
+        val showSameConsecutiveTypes = "Show same consecutive types"
```
Undin, line 43: Let's inline this label

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    override fun createSettings(): Settings = Settings()
+
+    override fun getCollectorFor(file: PsiFile, editor: Editor, settings: Settings, sink: InlayHintsSink): InlayHintsCollector? =
+        object : FactoryInlayHintsCollector(editor) {
+
+            val typeHintsFactory = RsTypeHintsPresentationFactory(factory, true)
+
+            override fun collect(element: PsiElement, editor: Editor, sink: InlayHintsSink): Boolean {
+                if (DumbService.getInstance(file.project).isDumb) return true
```
Undin, line 72:
```suggestion
                if (DumbService.isDumb(element.project)) return true
```

src/main/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProvider.kt
```diff
@@ -0,0 +1,140 @@
...
+                if (element !is RsMethodCall) return true
+                if (!element.isLastInChain) return true
+                if (!element.isEnabledByCfg) return true
+
+                val chain = collectChain(element)
+                var lastType: Ty? = null
+                for (call in chain.dropLast(1)) {
+                    if (call.type !is TyUnknown && call.isLastOnLine) {
```
Undin, line 80:
```suggestion
                    if (call.type != TyUnknown && call.isLastOnLine) {
```

src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    private fun doTest(@Language("Rust") code: String, showSameConsecutiveTypes: Boolean = true) {
+        val service = InlayHintsSettings.instance()
+        val key = RsChainMethodTypeHintsProvider.KEY
+        val settings = RsChainMethodTypeHintsProvider.Settings(showSameConsecutiveTypes = showSameConsecutiveTypes)
+        val originalSettings = service.findSettings(key, RsLanguage) { settings }
+        try {
+            service.storeSettings(key, RsLanguage, settings)
+            checkByText(code)
+        }
+        finally {
```
Undin, line 125: nit: [formatting](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting-control-flow-statements)

Issue #5284, opened 2020-04-26
Convert `Extract enum variant` intention to refactoring


Undin:
`Extract enum variant` intention introduced in #5037 and #5182 is more refactoring than typical intention because it makes global code transformation (replacement of the corresponding usages).
It's great but it can apply changes too long for typical intention because of working with usages. It's not nice both from the user side - the corresponding actions usually show progress in such cases, and platform side - all code (reference search and replacement) is invoked under write action that may lead to IDE freezes.
It would be great to refactor the implementation and use proper approaches from platform. More details:
* Use `com.intellij.refactoring.BaseRefactoringProcessor` to move search of usages into read action and provide the corresponding progress
* Provide the corresponding action in `Refactor | Extract` menu (see `org.rust.ide.refactoring.RsBaseEditorRefactoringAction`)
* Use `org.rust.ide.intentions.RsRefactoringAdaptorIntention` to provide the same refactoring action as intention or even drop existing intention and leave only refactoring

Also, I think we need to move all the corresponding code to `org.rust.ide.refactoring` package to be consistent with other refactorings

See `org.rust.ide.refactoring.convertStruct.RsConvertToNamedFieldsAction` as example

Undin:
CC @Kobzol @mchernyavsky

Kobzol:
If it causes IDE freezes or similar problems, then it definitely should be changed. I'm a bit worried about discoverability though. Usually if I want to perform something, I check Alt+Enter and if it's offered, I run it directly. Going through the refactoring menus is a bit slower process. Things like introducing a field or extracting a method is known by many users from other IntelliJ IDEs and other languages, but this refactoring would be Rust specific, so other people may not expect it to live in the refactoring menu. Is there a way to somehow increase it's visibility even after converting it to a refactoring action?

Undin:
We may keep it as intention but still use refactoring infrastructure under the hood. See `org.rust.ide.intentions.ConvertToStructIntention` as an example

Kobzol:
OK, that would be ideal from my point of view. I'll try to refactor it and send a PR.

PR #5291, opened 2020-04-27
REF: change ExtractEnumVariant from an intention to a refactoring

Kobzol:
I don't have a lot of experiences with the whole refactoring machinery, so let me know if I could do something in a better way. @Undin 

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5284


src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
```diff
@@ -0,0 +1,76 @@
...
+                                    private val ctx: RsEnumVariant) : BaseRefactoringProcessor(project) {
+    override fun findUsages(): Array<UsageInfo> {
+        return ReferencesSearch.search(ctx).map { UsageInfo(it) }.toTypedArray()
+    }
+
+    override fun performRefactoring(usages: Array<out UsageInfo>) {
+        val project = ctx.project
+        val factory = RsPsiFactory(project)
+        val element = createElement(ctx, factory)
+        val references = ReferencesSearch.search(ctx)
```
Undin, line 32: One of the main purpose of `BaseRefactoringProcessor` is split reading phase (usage searching) from writing phase (usage updating). Reading phase is already done here via invocation of `findUsages` method in `BaseRefactoringProcessor` implementation and the result is passed as `usages`.
So you don't need to search them again. Just use `usages `

Kobzol: Oh, sorry, this was actually a copy paste error. Thanks.

src/main/resources/META-INF/rust-core.xml
```diff
@@ -985,6 +985,13 @@
             <add-to-group group-id="RefactoringMenu"/>
         </action>
 
+        <action id="Rust.RsExtractEnumVariant"
+                class="org.rust.ide.refactoring.extractEnumVariant.RsExtractEnumVariantAction"
+                text="Extract Enum Variant"
+                description="Extract a single enum variant into a separate struct">
+            <add-to-group group-id="RefactoringMenu"/>
```
Undin, line 992: I think it should be located in `Refactor | Extract`. `IntroduceActionsGroup` is the corresponding id

src/test/kotlin/org/rust/ide/refactoring/RsExtractEnumVariantTest.kt
```diff
@@ -476,9 +479,20 @@ class ExtractEnumVariantIntentionTest : RsIntentionTestBase(ExtractEnumVariantIn
...
         fn main() {
             let _ = E::V1(V1 { x: 0, y: 1 });
         }
     """)
+
+    private fun doAvailableTest(@Language("Rust") before: String, @Language("Rust") after: String) {
+        InlineFile(before.trimIndent()).withCaret()
+        myFixture.testAction(RsExtractEnumVariantAction())
+        myFixture.checkResult(replaceCaretMarker(after.trimIndent()))
+    }
```
Undin, line 492: Can be simplified to

```suggestion
    private fun doAvailableTest(@Language("Rust") before: String, @Language("Rust") after: String) {
        checkEditorAction(before, after, "Rust.RsExtractEnumVariant")
    }
```

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.intellij.usageView.UsageViewDescriptor
+import org.rust.ide.inspections.import.RsImportHelper
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsReferenceElement
+import org.rust.lang.core.psi.ext.descendantOfTypeStrict
+import org.rust.lang.core.psi.ext.parentEnum
+
+class RsExtractEnumVariantProcessor(project: Project,
+                                    private val editor: Editor,
+                                    private val ctx: RsEnumVariant) : BaseRefactoringProcessor(project) {
```
Undin, line 23: Let's reformat it according to kotlin [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting)

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/impl.kt
```diff
@@ -0,0 +1,357 @@
...
+) : TypeVisitor {
+    override fun visitTy(ty: Ty): Boolean =
+        when {
+            ty is TyTypeParameter -> ty.name in parameters
+            ty.hasTyTypeParameters -> ty.superVisitWith(this)
+            else -> false
+        }
+}
+
+fun findApplicableContext(editor: Editor, file: PsiFile): RsEnumVariant? {
```
Undin, line 233: Should be moved to `RsExtractEnumVariantAction` as private method

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/impl.kt
```diff
@@ -0,0 +1,357 @@
...
+import org.rust.lang.core.types.consts.CtConstParameter
+import org.rust.lang.core.types.infer.TypeVisitor
+import org.rust.lang.core.types.infer.hasCtConstParameters
+import org.rust.lang.core.types.infer.hasTyTypeParameters
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyTypeParameter
+import org.rust.lang.core.types.type
+
+
+sealed class VariantElement(val toBeReplaced: PsiElement, val factory: RsPsiFactory) {
```
Undin, line 29: Looks like almost everything in this file (except `findApplicableContext`) is implementation details of `RsExtractEnumVariantProcessor` and I suggest keeping them into single file and make everything private (if possible). We don't use implementation details anywhere (including tests) so no need to provide some items to global scope

src/main/resources/META-INF/rust-core.xml
Kobzol: Thank you, I wanted to put it under Extract, but I didn't know how to do it, because the only other refactoring in this group (Extract Function) is built-in via `RsRefactoringSupplier` and I didn't find it in the XML. I also added `anchor="after" relative-to-action="ExtractMethod"`, otherwise there was a needless horizontal line between those two refactorings.

PR #5293, opened 2020-04-27
REF: don't rename non rust directories

Undin:
Otherwise, it may affect rename provided by other languages like renaming of Java packages.

Fixes https://youtrack.jetbrains.com/issue/IDEA-238958


Undin:
bors retry

Undin:
bors r=mchernyavsky

Issue #5283, opened 2020-04-26
Extract enum variant refactoring should respect parent enum attributes


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.122.3056-201-nightly
* **Rust toolchain version:** 1.42.0 (b8cedc004 2020-03-09) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.1 Ultimate Edition (IU-201.6668.121)
* **Operating system:** macOS 10.13.6

## Problem description
`Extract enum variant` refactoring introduced in #5037 and #5182 may produce invalid code when parent enum has attributes. For example, if enum has `#[derive(Debug)]`, extracted struct should have it too. Otherwise, it produces compiler error.

Looks like it's quite hard to support custom attributes because they can have arbitrary semantics, but it would be great to make this refactoring work with compiler attributes like `derive`

## Steps to reproduce
Before:
```rust
#[derive(Debug)]
enum IpAddr {
    V4/*caret*/(u8, u8, u8, u8),
    V6(String),
}

fn main() {
    let home = IpAddr::V4(127, 0, 0, 1);
    let loopback = IpAddr::V6(String::from("::1"));
    println!("{:?} {:?}", home, loopback)
}

```

After:
```rust
struct V4(pub u8, pub u8, pub u8, pub u8);

#[derive(Debug)]
enum IpAddr {
    V4(V4), // error[E0277]: `V4` doesn't implement `std::fmt::Debug`
    V6(String),
}

fn main() {
    let home = IpAddr::V4(V4(127, 0, 0, 1));
    let loopback = IpAddr::V6(String::from("::1"));
    println!("{:?} {:?}", home, loopback)
}
```

Undin:
CC @mchernyavsky @Kobzol

Kobzol:
I agree that copying `derive` would be very useful for this intention. Maybe also things like `repr(C)` could be copied? I see that the intfrastructure for querying attribute is in place, so it shouldn't be hard. Should I do it @mchernyavsky ?

Undin:
@Kobzol We may convert this issue into meta issue, collect all possible stdlib attributes that can be affected by the refactoring and support them one by one. From my point of view, `derive` should be supported first as the most frequent attribute

Kobzol:
Probably should wait for https://github.com/intellij-rust/intellij-rust/pull/5291 to be merged. I'll then try to work on it.

PR #5304, opened 2020-04-29
Copy enum attributes in RsExtractEnumVariant

Kobzol:
This PR adds support for copying (a supported subset of) attributes from the parent enum to the generated struct in `RsExtractEnumVariant`. I added `derive` and `repr` to the set for now.

@Undin The generated code puts an empty line between the attributes and the struct, like this:
```rust
#[derive(Debug, Clone)]
#[repr(C)]

pub struct V1 { pub x: i32, pub y: i32 }
```

I'm not putting it there, it seems like it's done automatically by the formatter. How can I avoid the empty line?

Blocked by https://github.com/intellij-rust/intellij-rust/pull/5291
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5283

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
@@ -54,9 +54,17 @@ class RsExtractEnumVariantProcessor(
         val parameters = filterTypeParameters(element.typeReferences, enum.typeParameterList)
         val typeParametersText = parameters.format()
         val whereClause = buildWhereClause(enum.whereClause, parameters)
+        val attributes = findTransitiveAttributes(enum, TRANSITIVE_ATTRIBUTES)
 
-        val struct = element.createStruct(enum.vis?.text, name, typeParametersText, whereClause)
+        val struct = element.createStruct(
+            enum.vis?.text,
+            name,
+            typeParametersText,
+            whereClause,
+            attributes.map { it.text }
+        )
         val inserted = enum.parent.addBefore(struct, enum) as RsStructItem
+
         for (occurrence in occurrences) {
             if (occurrence.reference?.resolve() == null) {
                 RsImportHelper.importElements(occurrence, setOf(inserted))
@@ -85,20 +93,31 @@ class RsExtractEnumVariantProcessor(
     override fun getRefactoringId(): String? {
         return "refactoring.extractEnumVariant"
     }
+
+    companion object {
+        val TRANSITIVE_ATTRIBUTES = setOf("derive", "repr")
+    }
 }
 
 
 private sealed class VariantElement(val toBeReplaced: PsiElement, val factory: RsPsiFactory) {
     abstract val typeReferences: List<RsTypeReference>
 
-    fun createStruct(vis: String?, name: String, typeParameters: String, whereClause: String): RsStructItem {
+    fun createStruct(
+        vis: String?,
+        name: String,
+        typeParameters: String,
+        whereClause: String,
+        attributes: List<String>
+    ): RsStructItem {
         val formattedVis = if (vis == null) "" else "$vis "
         val fieldsText = addPubToFields(toBeReplaced, factory).text
         val text = when (this) {
             is TupleVariant -> "${formattedVis}struct $name$typeParameters$fieldsText$whereClause;"
             is StructVariant -> "${formattedVis}struct $name$typeParameters$whereClause$fieldsText"
         }
-        return factory.createStruct(text)
+        val textWithAttributes = "${attributes.joinToString(separator = "\n")}\n$text"
+        return factory.createStruct(textWithAttributes)
     }
 
     abstract fun replaceUsage(element: PsiElement, name: String): PsiElement?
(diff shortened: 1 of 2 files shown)


Undin:
@Kobzol could you please rebase it onto master since #5291 is already merged?

Kobzol:
Done. There is still a problem with the whitespace after the added attribute, is there an easy way of removing it?

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
```diff
@@ -54,9 +54,16 @@ class RsExtractEnumVariantProcessor(
...
         val typeParametersText = parameters.format()
         val whereClause = buildWhereClause(enum.whereClause, parameters)
+        val attributes = findTransitiveAttributes(ctx, TRANSITIVE_ATTRIBUTES)
 
         val struct = element.createStruct(enum.vis?.text, name, typeParametersText, whereClause)
         val inserted = enum.parent.addBefore(struct, enum) as RsStructItem
+
+        for (attribute in attributes) {
+            val copiedAttribute = factory.createOuterAttr(attribute.metaItem.text)
+            inserted.parent.addBefore(copiedAttribute, inserted)
```
Undin, line 64: The reason of extra whitespace is here - you insert new attributes *before* struct itself although outer attributes should be part of struct item. As a result, you have attributes without parent struct and struct without attributes that leads to whitespace between them in editor after formatting

The existing code can be fixed in the following way `inserted.addBefore(copiedAttribute, inserted.firstChild)` but I'd recommend avoiding this approach here because each call of `factor.create*` creates a temporary file, fully parses it, and extracts necessary PSI nodes.

We already create struct item via psi factory. Let's add text of necessary attributes into initial struct text and create fully "configured" struct node by `element.createStruct` call, i.e. single `factor.create*` invocations

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
```diff
@@ -404,3 +415,6 @@ private fun createLifetimePredicate(
 
 private fun hasTypeParameter(ref: RsTypeReference, map: Map<String, RsTypeParameter>): Boolean =
     HasTypeParameterVisitor(map, ref).visitTy(ref.type)
+
+private fun findTransitiveAttributes(variant: RsEnumVariant, supportedAttributes: Set<String>): List<RsOuterAttr> =
+    variant.parentEnum.outerAttrList.filter { it.metaItem.name in supportedAttributes }.toList()
```
Undin, line 420:
```suggestion
    variant.parentEnum.outerAttrList.filter { it.metaItem.name in supportedAttributes }
```

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
```diff
@@ -54,9 +54,16 @@ class RsExtractEnumVariantProcessor(
         val parameters = filterTypeParameters(element.typeReferences, enum.typeParameterList)
         val typeParametersText = parameters.format()
         val whereClause = buildWhereClause(enum.whereClause, parameters)
+        val attributes = findTransitiveAttributes(ctx, TRANSITIVE_ATTRIBUTES)
```
Undin, line 57: Maybe it's better to pass `enum` instead of `ctx` here?
Sorry for a lot of single comments

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
Kobzol: Thanks, now it makes sense.

Kobzol:
Thanks, creating the attributes directly in the struct text is much easier.

Two tests that should be unrelated (`test import generated struct if needed` and `test don't import generated struct if its name already in scope`) started failing after the rebase, I checked and they also fail on `master` locally. Yet in the previous CI run it seems that they succeeded, let's see what CI says now.

src/main/kotlin/org/rust/ide/refactoring/extractEnumVariant/RsExtractEnumVariantProcessor.kt
```diff
@@ -85,20 +93,31 @@ class RsExtractEnumVariantProcessor(
...
+        attributes: List<String>
+    ): RsStructItem {
         val formattedVis = if (vis == null) "" else "$vis "
         val fieldsText = addPubToFields(toBeReplaced, factory).text
         val text = when (this) {
             is TupleVariant -> "${formattedVis}struct $name$typeParameters$fieldsText$whereClause;"
             is StructVariant -> "${formattedVis}struct $name$typeParameters$whereClause$fieldsText"
         }
-        return factory.createStruct(text)
+        val textWithAttributes = "${attributes.joinToString(separator = "\n", postfix = "\n")}${text}".trim()
```
Undin, line 119:
```suggestion
        val textWithAttributes = "${attributes.joinToString(separator = "\n")}\n$text"
```

PR #5306, opened 2020-04-29
INSP: add AddTypeArguments quick fix to RsWrongTypeArgumentsNumberInspection

Kobzol:
This PR adds a quick fix that adds missing type arguments to generic argument lists that are missing some required (non-defaulted) parameters. In a follow-up PR I'll improve the quick fix to also include associated types.

add

- Should the parameters be filled with `_` or the type parameter names from the resolved type (i.e. `K`, `V` in a hash map)? The first one should play better with inference, the latter provides more guidance to the user on what to use for the actual type arguments (although the type parameter names are often very short and lack context). Right now the implementation uses the latter.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
@@ -1,0 +1,89 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.ide.inspections.getTypeArgumentsAndDeclaration
+import org.rust.lang.core.parser.RustParserUtil
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsElementTypes.COMMA
+import org.rust.lang.core.psi.RsElementTypes.LT
+import org.rust.lang.core.psi.ext.*
+import org.rust.openapiext.buildAndRunTemplate
+import org.rust.openapiext.createSmartPointer
+
+class AddTypeArguments(element: RsElement) : LocalQuickFixAndIntentionActionOnPsiElement(element) {
+    override fun getText(): String = "Add missing type arguments"
+    override fun getFamilyName() = text
+
+    override fun invoke(
+        project: Project,
+        file: PsiFile,
+        editor: Editor?,
+        startElement: PsiElement,
+        endElement: PsiElement
+    ) {
+        val element = startElement as? RsElement ?: return
+        val (typeArguments, declaration) = getTypeArgumentsAndDeclaration(element) ?: return
+
+        val requiredParameters = declaration.requiredGenericParameters
+        if (requiredParameters.isEmpty()) return
+
+        val argumentCount = typeArguments?.typeReferenceList?.size ?: 0
+        if (argumentCount >= requiredParameters.size) return
+
+        val factory = RsPsiFactory(project)
+        val missingTypes = requiredParameters.drop(argumentCount).map { it.name ?: "_" }
+
+        val replaced = if (typeArguments != null) {
+            var anchor = with(typeArguments) {
+                typeReferenceList.lastOrNull() ?: lifetimeList.lastOrNull() ?: lt
+            }
+            val nextSibling = anchor.getNextNonCommentSibling()
+            val addCommaAfter = nextSibling?.isComma == true
+            if (addCommaAfter && nextSibling != null) {
+                anchor = nextSibling
+            }
+
+            for (type in missingTypes) {
+                if (anchor.elementType != LT && !anchor.isComma) {
+                    anchor = typeArguments.addAfter(factory.createComma(), anchor)
+                }
+                anchor = typeArguments.addAfter(factory.createType(type), anchor)
+            }
+
+            if (addCommaAfter) {
+                typeArguments.addAfter(factory.createComma(), anchor)
+            }
+
+            typeArguments
+        } else {
+            val newArgumentList = factory.createTypeArgumentList(missingTypes)
+
+            // this can only happen for type references (base types/trait refs)
+            val path = getPath(element) ?: return
+            path.addAfter(newArgumentList, path.identifier) as RsTypeArgumentList
+        }
+
+        editor?.buildAndRunTemplate(element, replaced.typeReferenceList.drop(argumentCount).map { it.createSmartPointer() })
+    }
+}
+
+private val RsGenericDeclaration.requiredGenericParameters: List<RsTypeParameter>
+    get() = typeParameters.filter { it.typeReference == null }
+
+private fun getPath(element: PsiElement): RsPath? = when (element) {
+    is RsBaseType -> element.path
+    is RsTraitRef -> element.path
+    else -> null
+}
+
+private val PsiElement.isComma: Boolean
+    get() = elementType == COMMA
--- a/src/test/kotlin/org/rust/ide/inspections/RsWrongTypeArgumentsNumberInspectionTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/RsWrongTypeArgumentsNumberInspectionTest.kt
@@ -209,4 +209,186 @@ class RsWrongTypeArgumentsNumberInspectionTest : RsInspectionsTestBase(RsWrongTy
         trait Trait<A> {}
         fn foo(_: impl <error descr="Wrong number of type arguments: expected 1, found 0 [E0107]">Trait<></error>) {}
     """)
+
+    fun `test add arguments missing arguments`() = checkFixByText("Add missing type arguments", """
+        struct S<T> { t: T }
+
+        fn main() {
+            let x: <error descr="Wrong number of type arguments: expected 1, found 0 [E0107]">S/*caret*/</error>;
+        }
+    """, """
+        struct S<T> { t: T }
+
+        fn main() {
+            let x: S<T>;
+        }
+    """)
+
+    fun `test add arguments empty arguments`() = checkFixByText("Add missing type arguments", """
+        struct S<T> { t: T }
+
+        fn main() {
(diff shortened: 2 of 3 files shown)


Undin:
> Should this be an intention? Or a quick fix associated with RsWrongTypeParametersNumberInspection? Or should it just be part of code completion to provide <...> when tab is pressed?

The original issue was about completion, i.e. just simple insertion of  `<>` while generic type completion and caret movement between braces. I think it's the simplest and efficient way to help users because you don't need to do additional action - just choose completion variant.
Of course, we can also add smarter quick fix/intention (I think quick fix is better but not sure) with more functionality as _additional_ feature instead of replacement of the proposed one
So I suggest to implement the feature from completion side first and only after that do smarter things.
What do you think?

Kobzol:
That sounds like a reasonable first step. It should thus only insert `<>` in code completion, not add something like `<_, _>`, correct? Could you please point me to a file/package where the code completion code lives? I haven't dealt with code completion yet.

Undin:
> It should thus only insert <> in code completion, not add something like <_, _>, correct?

Yep. Exactly as when we insert `()` for function/method calls. But be careful, we need to insert `<>` only when we want to complete type. For example, if you write `let v = Vec::new();` you don't need to insert `<>`

> Could you please point me to a file/package where the code completion code lives? I haven't dealt with code completion yet.

I think this [article](https://www.jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/code_completion.html) can be useful to get a general overview of how completion works in IntelliJ platform.
All completion related code in our project is located in `org.rust.lang.core.completion` package. Completion based on name resolution result is provided by `org.rust.lang.core.completion.RsCommonCompletionProvider`. To insert additional text see `org.rust.lang.core.completion.RsDefaultInsertHandler`

Kobzol:
Thanks, I created https://github.com/intellij-rust/intellij-rust/pull/5308.

Kobzol:
Now that https://github.com/intellij-rust/intellij-rust/pull/5308 is merged, I think that this PR can proceed. `RsWrongTypeParametersNumberInspection` already provides a warning about a wrong number of type parameters, but it cannot fix it currently. Should I modify this intention to serve as a quick fix for `RsWrongTypeParametersNumberInspection`?

Kobzol:
I changed the intention to a quick fix which is used by `RsWrongTypeParametersNumberInspection`.

When the type/function call already has type arguments, I replace them: https://github.com/intellij-rust/intellij-rust/pull/5306/commits/88884f8566edd6fd8cb315105495848e0b08a9c1#diff-7d666fa9e7c6c4b211246e82df2a2cf5R34 However I'm not sure what to do if there are no type arguments yet and I have to fill them from scratch. Should I recreate the `RsPath` of the type/function call with added type arguments using `RsPsiFactory` and then replace the path?

Kobzol:
I also noticed that the inspection is a bit outdated, it does not cover all possible errors on method calls and it uses old error codes (E0087, E0089) instead of the newer E0107. Should I change it or is it there for backwards compatibility?

Undin:
> I also noticed that the inspection is a bit outdated, it does not cover all possible errors on method calls and it uses old error codes (E0087, E0089) instead of the newer E0107. Should I change it or is it there for backwards compatibility?

I don't think it's worth to use old code even if some old compiler versions use them. So it would be great to update error codes. But I'd prefer to make it in separate PR

Kobzol:
Ok, I opened https://github.com/intellij-rust/intellij-rust/pull/5354.

Kobzol:
I rebased the PR onto the recently landed type inspection PRs and added support for some missing cases. I'll add support for adding missing associated types in a separate PR.

> Should the parameters be filled with _ or the type parameter names from the resolved type (i.e. K, V in a hash map)? The first one should play better with inference, the latter provides more guidance to the user on what to use for the actual type arguments (although the type parameter names are often very short and lack context).
 
I chose to prefill the type arguments with the parameter name rather than with `_`, it seems like a better solution to me, what do you think?

add

Kobzol:
@Undin ping

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -294,9 +294,10 @@ class RsPsiFactory(
     }
 
     fun createTypeArgumentList(
-        params: Iterable<String>
+        params: Iterable<String>,
+        useColonColon: Boolean = false
     ): RsTypeArgumentList {
-        val text = params.joinToString(prefix = "<", separator = ", ", postfix = ">")
+        val text = (if (useColonColon) "::" else "") + params.joinToString(prefix = "<", separator = ", ", postfix = ">")
```
Undin, line 300: Looks like we don't have any test when `useColonColon` is true

src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
```diff
@@ -0,0 +1,66 @@
...
+        val argumentLifetimes = typeArguments?.lifetimeList?.map { it.text } ?: listOf()
+        val argumentTypes = typeArguments?.typeReferenceList?.map { it.text } ?: listOf()
+        val argumentAssociatedTypes = typeArguments?.assocTypeBindingList?.map { it.text } ?: listOf()
+
+        val missingTypes = requiredParameters.drop(argumentTypes.size).map { it.name ?: "_" }
+
+        val newArguments = argumentLifetimes + argumentTypes + missingTypes + argumentAssociatedTypes
+        val newArgumentList = factory.createTypeArgumentList(newArguments, typeArguments?.coloncolon != null)
+        val replaced = if (typeArguments != null) {
+            typeArguments.replace(newArgumentList) as RsTypeArgumentList
```
Undin, line 52: In this case, we can drop some additional info in existent code like comments, extra whitespaces, etc.
Can we avoid these changes?

src/test/kotlin/org/rust/ide/inspections/RsWrongTypeArgumentsNumberInspectionTest.kt
```diff
@@ -188,4 +188,72 @@ class RsWrongTypeArgumentsNumberInspectionTest : RsInspectionsTestBase(RsWrongTy
...
+        fn main() {
+            let x: S<u32, B, Item=u32>;
+        }
+    """)
+
+    fun `test add arguments ignore type parameters with a default`() = checkFixIsUnavailable("Add missing type arguments", """
+        struct S<A, B=u32, C=u32>(A, B, C);
+
+        fn main() {
+            let x: S<u32, u32>;
```
Undin, line 256: Hm, all tests of new quick fix use only struct although it should support function and method calls. Let's add necessary tests to be sure that everything works as expected

src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        val requiredParameters = declaration.requiredGenericParameters
+        if (requiredParameters.isEmpty()) return
+
+        if (typeArguments?.typeReferenceList?.size ?: 0 >= requiredParameters.size) return
+
+        val factory = RsPsiFactory(project)
+
+        val argumentLifetimes = typeArguments?.lifetimeList?.map { it.text } ?: listOf()
+        val argumentTypes = typeArguments?.typeReferenceList?.map { it.text } ?: listOf()
```
Undin, line 44: Do we really need `text` here? Maybe `name` is enough?

Kobzol: Hmm, I thought that this would preserve comments, but actually even with `text` comments are lost. In any case, `name` is only available for lifetimes, if I'm not mistaken, type references can be of various types, so `text` has to be used? Or is there a way around it?

src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
Kobzol: I added code that preserves the surrounding whitespace and comments. It's not perfect, because the formatter will reformat the code after it's inserted, but it should preserve most important stuff, mainly comments.

Kobzol:
I added code that preserves whitespace and comments and added basic tests for function and method calls. I rebased the branch, since it was already getting pretty old.

src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
Kobzol: I wanted to do that originally, however I don't know how. If I just use `typeArguments?.typeReferenceList?.add(factory.createType("Foo"))`, it doesn't change PSI/resulting text. Is there a way to insert the items without creating a new `RsTypeArgumentList`?

Undin: Oh, it's expected since `typeArguments?.typeReferenceList` returns `List` collection and `add` method just insert element into it. But this list is not connected to PSI structure so sich insertion doesn't modify PSI.
You need to add children psi objects to `typeArguments` element directly.
As an alternative solution, you can use underlying text abstraction - `Document`. In this case, you don't need to construct PSI yourself, just insert necessary text into document directly. Don't forget to commit your changes via `com.intellij.psi.PsiDocumentManager#commitDocument` before `editor?.buildAndRunTemplate` because you need to get already changed PSI

Kobzol: Thank you, it is much simpler this way.

Undin:
@Kobzol Could you rebase into master, please?

src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        val argumentCount = typeArguments?.typeReferenceList?.size ?: 0
+        if (argumentCount >= requiredParameters.size) return
+
+        val factory = RsPsiFactory(project)
+        val missingTypes = requiredParameters.drop(argumentCount).map { it.name ?: "_" }
+
+        val replaced = if (typeArguments != null) {
+            var anchor = typeArguments.typeReferenceList.lastOrNull() ?: typeArguments.lifetimeList.lastOrNull()
+            ?: typeArguments.lt
```
Undin, line 47: Probably, it's more readable
```suggestion
            var anchor = with(typeArguments) {
                typeReferenceList.lastOrNull() ?: lifetimeList.lastOrNull() ?: lt
            }
```

Kobzol:
bors r=undin

Issue #4508, opened 2019-10-08
Completing generic types could add `<>` automatically


matklad:
In code like

```rust
let xs = (0..10)
    .map(|i| i * 2)
    .collect::<Ve/*cursor*/>()
```

it's pretty clear that, for `Vec` completion variant, `<>` would be needed anyway. It would be cool if IDE inserted them automatically, similarly for `()` in function calls.

see https://github.com/rust-analyzer/rust-analyzer/pull/1970

Kobzol:
I started working on this, however there are multiple ways to proceed. There is currently an inspection called `RsWrongTypeParametersNumberInspection` which warns about wrong number of type parameters and if there are parameters when there should be none, it uses the `RemoveTypeParameter` quick fix to remove them.

I can either (ranging from most to least general solution):
1) Extend the inspection to apply a different quick fix if there are missing type parameters, let's say `AddTypeParameter`, which would add the missing type parameters, and then also use this quick fix for code completion.
2) Create a new intention which would add missing type parameters (independent of `RsWrongTypeParametersNumberInspection`) and reuse this intention during code completion.
3) Just modify code completion to add type parameters and don't change the existing inspection.

I think that 1) would be the best solution, as it would allow to add missing type parameters anywhere where there is a generic type/function call (and it could also be reused for code completion). If you think that it's a good idea, I'll start a PR where we can discuss this further. What do you think?

PR #5308, opened 2020-04-30
COMP: add completion of angle brackets for generic types

Kobzol:
This PR adds completion for angle brackets for generic types, i.e.

```rust
let x: Ve/*caret*/
vvv
let x: Vec</*caret*/>
```

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4508

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -322,8 +322,7 @@ private fun addAngleBrackets(element: RsGenericDeclaration, document: Document, 
     val path = context.getElementOfType<RsPath>()
     if (path == null || path.parent !is RsTypeReference) return
 
-    if (!context.alreadyHasAngleBrackets)
-    {
+    if (!context.alreadyHasAngleBrackets) {
         document.insertString(context.selectionEndOffset, "<>")
     }
     EditorModificationUtil.moveCaretRelatively(context.editor, 1)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -308,6 +314,19 @@ private fun appendSemicolon(context: InsertionContext, curUseItem: RsUseItem?) {
     }
 }
 
+private fun addAngleBrackets(element: RsGenericDeclaration, document: Document, context: InsertionContext) {
+    if (element.typeParameters.isEmpty()) return
+
+    // do not complete angle brackets if not in type place
+    if (context.getElementOfType<RsPathExpr>() != null) return
```
vlad20012, line 321:
```suggestion
    val path = context.getElementOfType<RsPath>()
    if (path == null || path.parent is RsTypeReference) return
```
I'd do a reverse thing: check that we are defenitely inside a type

Kobzol:
Added more tests and also disabled the completion if all type parameters of the target type have a default.

src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
```diff
@@ -308,6 +314,21 @@ private fun appendSemicolon(context: InsertionContext, curUseItem: RsUseItem?) {
...
+private fun addAngleBrackets(element: RsGenericDeclaration, document: Document, context: InsertionContext) {
+    // complete only types that have at least one type parameter without a default
+    if (element.typeParameters.all { it.typeReference != null }) return
+
+    // complete angle brackets only in a type context
+    val path = context.getElementOfType<RsPath>()
+    if (path == null || path.parent !is RsTypeReference) return
+
+    if (!context.alreadyHasAngleBrackets)
+    {
```
Undin, line 326: [formatting](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting)

vlad20012:
:+1: 
bors r+

Issue #1712, opened 2017-09-09
Highlight `!`-typed expressions as exit points of a function


matklad:
The relevant code is `RsHighlightExitPointsHandlerFactory`

CC, @kumbayo you've recently added proper types for unimplemented and unreachable macros :)

PR #5310, opened 2020-04-30
ANN: highlight never type exit points

Kobzol:
Added support for highlighting never types as function exit points.
Also fixes highlighting of macro expressions in tail position.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/1712


ortem:
Thank you!
bors r+

Issue #4444, opened 2019-09-25
Extract function action produces weird output for Maps


SomeoneToIgnore:
## Environment

* **IntelliJ Rust plugin version:** 0.2.106.2134-193
* **Rust toolchain version:** 1.37.0 (eae3437df 2019-08-13) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.3519.24)
* **Operating system:** macOS 10.14.6

## Problem description

Consider the code:

```rust
fn main() {
    let mut test_map = std::collections::HashMap::new();
    test_map.insert(1, 2);
}
```

Note that there are no `use` declarations for the HashMap, it's used with a fully qualified name.
I would like to extract the `test_map.insert(1, 2);` string into a separate method.
I select it, use extract function action, specify the function name and get the method.

Expected:

```rust
fn populate_map(test_map: &mut std::collections::HashMap<i32, i32>) {
    test_map.insert(1, 2);
}
```

Actual:

```rust
fn populate_map(test_map: &mut HashMap<i32, i32, RandomState>) {
    test_map.insert(1, 2);
}
```

Two things that are weird:

1. No fully qualified name is used. Since I've had no `use`, I would expect the extraction to either use the FQN or add a `use` instead and, potentially, remove the first FQN.

2. A `RandomState` generic pops up. Usually I don't care about the generics that have a default value specified so I would expect it to be omitted.

dima74:
@SomeoneToIgnore looks like it is fixed now:
1. Import for `HashMap` is added
2. No `RandomState` generics thanks to #5312

Here is result in version 0.4.154.4085-212:
```rust
use std::collections::HashMap;

fn main() {
    let mut test_map = std::collections::HashMap::new();
    populate(&mut test_map);
}

fn populate(test_map: &mut HashMap<i32, i32>) {
    test_map.insert(1, 2);
}
```

PR #5312, opened 2020-04-30
REF: omit unchanged default type parameters of parameters in ExtractFunction

Kobzol:
This PR modifies the `ExtractFunction` refactoring to omit needless type arguments which are the same as the default for its given type parameter.

~I had to create yet another type renderer to not mess with other places where types are rendered. I wonder if it would be possible to replace things like `Ty.insertionSafeTextWithAliasesAndLifetimes` with simply creating a type renderer in each individual place where it's needed? The callers could then specify exactly the properties for type rendering that they need. The `TypeRenderer` seems very cheap to create, but if it would be a perf. problem then some dynamic caching mechanism could be used.~

I think that skipping the default unchanged type arguments is pretty useful and could be used as a default for type rendering. But maybe there are some places where it could cause problems (I can try to switch it to be used as a default and we can see what the CI says about it). 

Fixes second part of https://github.com/intellij-rust/intellij-rust/issues/4444

### DIFF
--- a/src/test/kotlin/org/rust/ide/refactoring/RsExtractFunctionTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/RsExtractFunctionTest.kt
@@ -1110,6 +1110,50 @@ class RsExtractFunctionTest : RsTestBase() {
         false,
         "foo")
 
+    fun `test extract a function with unset default type parameter`() = doTest("""
+        struct S<R, T=u32>(r: R, t: T);
+
+        fn main() {
+            let s: S<u32> = S(1u32, 2u32);
+            <selection>println!(s)</selection>;
+        }
+    """, """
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/ide/refactoring/RsExtractFunctionTest.kt
```diff
@@ -1110,6 +1110,50 @@ class RsExtractFunctionTest : RsTestBase() {
...
+
+        fn foo(s: S<u32>) {
+            println!(s)
+        }
+    """,
+        false,
+        "foo")
+
+    fun `test extract a function with set default type parameter`() = doTest("""
+        struct S<R, T=u32>(r: R, t: T);
```
vlad20012, line 1136: Same here

PR #5326, opened 2020-05-04
GRD: improve developer workflows

Undin:
These changes should speed up some developer workflows related to IDE launch:
* platform version switch. Now we use different sandbox directories for different platform versions that should keep application data (like indices or expanded macros) when you change platform version. As a result, it eliminates unnecessary indexation when you want to check something with non-default platform version
* debugging. Now gradle script debugging is disabled for `RunIDEA` and `RunCLion` run configurations that speeds up IDE launch

Other changes:
* `Run` run configuration is renamed to `RunIDEA` to be consistent with `RunCLion`
* `configureCLion` task is dropped as redundant because now we provide explicit sandbox name
* `CONTRIBUTING.md` improvements and updates

### DIFF
--- a/.idea/runConfigurations/RunCLion.xml
+++ b/.idea/runConfigurations/RunCLion.xml
@@ -17,7 +17,7 @@
       </option>
       <option name="vmOptions" value="" />
     </ExternalSystemSettings>
-    <GradleScriptDebugEnabled>true</GradleScriptDebugEnabled>
+    <GradleScriptDebugEnabled>false</GradleScriptDebugEnabled>
     <method v="2" />
   </configuration>
 </component>
--- a/.idea/runConfigurations/RunCLion.xml
+++ b/.idea/runConfigurations/RunCLion.xml
@@ -1,6 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="RunCLion" type="GradleRunConfiguration" factoryName="Gradle">
-    <log_file alias="idea.log" path="$PROJECT_DIR$/plugin/build/clion-sandbox/system/log/idea.log" />
+    <log_file alias="idea-201.log" path="$PROJECT_DIR$/plugin/build/clion-sandbox-201/system/log/idea.log" />
+    <log_file alias="idea-193.log" path="$PROJECT_DIR$/plugin/build/clion-sandbox-193/system/log/idea.log" />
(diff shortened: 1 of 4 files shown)


.idea/runConfigurations/RunCLion.xml
```diff
@@ -1,6 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="RunCLion" type="GradleRunConfiguration" factoryName="Gradle">
-    <log_file alias="idea.log" path="$PROJECT_DIR$/plugin/build/clion-sandbox/system/log/idea.log" />
+      <log_file alias="idea-201.log" path="$PROJECT_DIR$/plugin/build-201/clion-sandbox/system/log/idea.log" />
+      <log_file alias="idea-193.log" path="$PROJECT_DIR$/plugin/build-193/clion-sandbox/system/log/idea.log" />
```
Undin, line 4: Unfortunately, I don't know a way how to provide single path for all platform versions because run configurations support ant patterns only for filename

vlad20012:
`.gitignore` should also be updated

.idea/runConfigurations/RunIDEA.xml
```diff
@@ -1,11 +1,12 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="Run" type="GradleRunConfiguration" factoryName="Gradle" singleton="true">
-    <log_file alias="idea.log" path="$PROJECT_DIR$/plugin/build/idea-sandbox/system/log/idea.log" />
+  <configuration default="false" name="RunIDEA" type="GradleRunConfiguration" factoryName="Gradle" singleton="true">
+    <log_file alias="idea-201.log" path="$PROJECT_DIR$/plugin/build/idea-sandbox-201/system/log/idea.log" />
+    <log_file alias="idea-193.log" path="$PROJECT_DIR$/plugin/build/idea-sandbox-201/system/log/idea.log" />
```
vlad20012, line 4:
```suggestion
    <log_file alias="idea-193.log" path="$PROJECT_DIR$/plugin/build/idea-sandbox-193/system/log/idea.log" />
```

Undin: Oh, nice catch!
Thanks!

vlad20012:
:+1: 
bors r+

PR #5329, opened 2020-05-05
CI: use windows runners on github action to run tests

Undin:
Preparation to make github actions default CI for the project


Undin:
Some notes:
* Windows tests are slower than Linux ones. Probably, it's worth to keep them only for PRs as we do currently. I've tried to use [caches](https://github.com/actions/cache) but it doesn't make it faster and even produces some errors sometimes
* Probably, we need to cache `evcxr_repl` binary because it takes around 2 minutes to build it on Linux (19 minutes is the average time of running all tests). 
* Currently, we don't set up a symlink for stdlib that's required for some our tests. See https://github.com/intellij-rust/intellij-rust/blob/master/.travis.yml#L43 and https://github.com/intellij-rust/intellij-rust/blob/master/appveyor.yml#L19 in existing configurations for Travis and Appveyor

lancelote:
bors merge

Issue #4123, opened 2019-07-10
Create implement trait action


sirgl:
It is very tedious write boilerplate code for trait implementations by hand.

E. g. for
```rust
pub struct TreeResultPrettyView<'t, 'r, 'e> {
    result: &'t TreeResult,
    resolver: &'r dyn KindResolver,
    escaper: &'e dyn Escaper,
}
```
it is boring to write something like this with all type and lifetime parameters repeating:
```rust
impl <'t, 'r, 'e>Display for TreeResultPrettyView<'t, 'r, 'e> {
    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
       ...
    }
}
```
So I suggest to add dialog like for goland: 
https://habrastorage.org/webt/io/iy/t3/ioiyt3kh0gn_057wq5_wd66h8iu.gif

## Environment

* **IntelliJ Rust plugin version:** 0.2.101.2129-192
* **Rust toolchain version:** 1.38.0-nightly (481068a70 2019-07-05) x86_64-apple-darwin
* **IDE name and version:** CL-192.5438.15
* **Operating system:** macOS 10.14.5

AndreKR:
This should probably also create `impl` blocks for supertraits.

For example when I want to implement `std::error::Error` for an error type, it needs to create `impl` blocks for `core::fmt::Debug` and `core::fmt::Display`.

PR #5339, opened 2020-05-07
INT: add intention to implement a trait for a struct/enum

Kobzol:
This PR attempts to add a refactoring action which would generate a trait `impl` for a struct/enum. For now, I would only enable this action directly from a struct/enum. Later we can possibly add it anywhere in the top level of a module so that the user can select the struct/enum.

We still have to select the trait and I need some guidance here. I don't know which method should I use (or write) to find all available traits that could be implemented. I looked at `NameResolution.kt`, but didn't find what I wanted.

My intention of what this refactoring should do:
1) User starts the action with `Generate Trait impl` on a struct/enum
2) Text input is shown to the user, she uses it to find the corresponding trait
3) `impl $trait for $struct` is created in the struct's file, the corresponding trait is imported and then `ImplementMembersFix` is executed on the impl to generate the missing methods.

Right now the implementation is just a sketch, as I didn't know how to find the available traits.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4123
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7342

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
@@ -30,48 +30,60 @@ class AddTypeArguments(element: RsElement) : LocalQuickFixAndIntentionActionOnPs
         endElement: PsiElement
     ) {
         val element = startElement as? RsElement ?: return
-        val (typeArguments, declaration) = getTypeArgumentsAndDeclaration(element) ?: return
-
-        val requiredParameters = declaration.requiredGenericParameters
-        if (requiredParameters.isEmpty()) return
-
+        val (typeArguments, _) = getTypeArgumentsAndDeclaration(element) ?: return
         val argumentCount = typeArguments?.typeReferenceList?.size ?: 0
-        if (argumentCount >= requiredParameters.size) return
 
-        val factory = RsPsiFactory(project)
-        val missingTypes = requiredParameters.drop(argumentCount).map { it.name ?: "_" }
+        val replaced = addTypeArguments(element) ?: return
+        editor?.buildAndRunTemplate(element, replaced.typeReferenceList.drop(argumentCount).map { it.createSmartPointer() })
+    }
+}
 
-        val replaced = if (typeArguments != null) {
-            var anchor = with(typeArguments) {
-                typeReferenceList.lastOrNull() ?: lifetimeList.lastOrNull() ?: lt
-            }
-            val nextSibling = anchor.getNextNonCommentSibling()
-            val addCommaAfter = nextSibling?.isComma == true
-            if (addCommaAfter && nextSibling != null) {
-                anchor = nextSibling
-            }
+/**
+ * Adds missing type arguments to the given element.
+ *
+ * Return an instance of the replaced type arguments, if any were added.
+ */
+fun addTypeArguments(element: RsElement): RsTypeArgumentList? {
+    val project = element.project
+    val (typeArguments, declaration) = getTypeArgumentsAndDeclaration(element) ?: return null
 
-            for (type in missingTypes) {
-                if (anchor.elementType != LT && !anchor.isComma) {
-                    anchor = typeArguments.addAfter(factory.createComma(), anchor)
-                }
-                anchor = typeArguments.addAfter(factory.createType(type), anchor)
-            }
+    val requiredParameters = declaration.requiredGenericParameters
+    if (requiredParameters.isEmpty()) return null
 
-            if (addCommaAfter) {
-                typeArguments.addAfter(factory.createComma(), anchor)
-            }
+    val argumentCount = typeArguments?.typeReferenceList?.size ?: 0
+    if (argumentCount >= requiredParameters.size) return null
 
-            typeArguments
-        } else {
-            val newArgumentList = factory.createTypeArgumentList(missingTypes)
+    val factory = RsPsiFactory(project)
+    val missingTypes = requiredParameters.drop(argumentCount).map { it.name ?: "_" }
 
-            // this can only happen for type references (base types/trait refs)
-            val path = getPath(element) ?: return
-            path.addAfter(newArgumentList, path.identifier) as RsTypeArgumentList
+    return if (typeArguments != null) {
+        var anchor = with(typeArguments) {
+            typeReferenceList.lastOrNull() ?: lifetimeList.lastOrNull() ?: lt
+        }
+        val nextSibling = anchor.getNextNonCommentSibling()
+        val addCommaAfter = nextSibling?.isComma == true
+        if (addCommaAfter && nextSibling != null) {
+            anchor = nextSibling
         }
 
-        editor?.buildAndRunTemplate(element, replaced.typeReferenceList.drop(argumentCount).map { it.createSmartPointer() })
+        for (type in missingTypes) {
+            if (anchor.elementType != LT && !anchor.isComma) {
+                anchor = typeArguments.addAfter(factory.createComma(), anchor)
+            }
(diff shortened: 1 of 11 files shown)


vlad20012:
Oh, looks like the PR was missed, I'm sorry. 

If you want to show a dialog, you can try the approach used in AddFmtStringArgumentIntention, but use RsPathCodeFragment. 

Btw the same experience is almost achievable with existing actions:

Peek 2020-06-19 11-19

Kobzol:
You're right, although I still think that it could be valuable, it could for example generate impl blocks for supertraits of the implemented trait and it would be less typing for the user. I'll try the dialog to see if it is viable. Is `NameResolution::processDeriveTraitResolveVariants` the correct API to use to auto-complete available traits based on a value entered in the dialog?

vlad20012:
> You're right, although I still think that it could be valuable, it could for example generate impl blocks for supertraits of the implemented trait and it would be less typing for the user.

Yeah, I don't argue. The idea of supertraits is also cool. 

> Is NameResolution::processDeriveTraitResolveVariants the correct API to use to auto-complete available traits based on a value entered in the dialog?

Umm, no, take a look at AddFmtStringArgumentIntention, it opens a popup with completion for expressions. The same approach can be used to complete trait paths if you use RsPathCodeFragment instead of RsExpressionCodeFragment there.

Kobzol:
I created an intention that shows a popup on a struct/enum and creates an `impl` block for the selected trait and all of its supertraits.

Some questions:
1) It takes a while to generate all of the traits and supertraits, maybe it should be a refactoring instead with an action that triggers it?
2) I refactored out the popup from `AddFmtStringArgumentIntention`, because I needed almost exactly the same thing, just with error validation. If the user enters something that is not a trait, how should I display an error? Some text in the popup? Or an error message dialog?
3) How to use `ImplementMembersFix` programmatically? It runs into errors with write actions, because it doesn't assume that it will be called from another intention. Also it tries to display GUI, should I refactor it to avoid GUI and instead implement all of the available methods/constants/associated types? Or just do not implement the members at all and let the user use the fix afterwards?
4) What about generic traits? I guess that we cannot generally map the struct's type parameters to the trait's type parameters, so should we even include the type parameters of the trait? Maybe we can use template builder to let the user change the trait definitions after the impl blocks are generated.

For example:
```rust
trait Trait<T> {}
struct S<R>(R);

vvv

impl<R> <selection>Trait</selection> for S<R> {}
```
And the user would add `<R>` (or whatever else she wants).

vlad20012:
Oh, I should think a bit about some of these questions =D

3. `startInWriteAction(): Boolean = false` ?
4. You can add `<>` next to the trait reference if the trait is generic. Also (empty) template can appear inside `<>`

Btw consider do almost the same as does `AddImplIntention`, but generate `impl | for Baz {}` template at the trait place.

Kobzol:
`generateTraitMembers` calls `checkWriteAccessNotAllowed` and then `runWriteAction`. If I start the intention without a write action, the first check succeeds, but the write action does not succeed (`Cannot change PSI outside of undo-transparent action..`). If I change `runWriteAction` to `project.runWriteCommandAction`, it works.

However, in tests, if the intention starts without a write action, any test fails with this error:
`com.intellij.util.IncorrectOperationException: Cannot modify a read-only file '/src/main.rs'.`. Even though the write access is wrapped in `runWriteAction` or `runWriteCommandAction`.

Kobzol:
Rebased onto `master` and found a way to run simple tests.
implement-trait

Kobzol:
@vlad20012 ping

Kobzol:
@vlad20012 Is it worth to merge this PR or do you think that the offered functionality is not useful enough to have a separate intention for this?

vlad20012:
Oh

vlad20012:
1. I don't like popup-based UI because it is very unusual for intention actions. I know that we have `AddFmtStringArgumentIntention`, and it's also unusual and weird, but I believe it is rarely used (and will likely become almost useless after `implicit named arguments` stabilization), when `implement trait` pretend to be a common workflow. Btw Intellij-based products seem moving from popup-based refactorings to inline ones, i.e. when only the editor is used. I think it's also related to the fact that it requires some work to support custom UI elements in Code With Me and [Projector](https://lp.jetbrains.com/projector/). 
2. In the current state, the action does a lot of stuff without my permission. I just press "implement me a trait" and it immediately inserts impls for all ancestor traits, then opens many windows... it looks messy. I feel like I'm out of control. This can be fixed by showing a one merged window with all implementing traits and generating members (where I can also unselect some entire traits).
3. trait-is-already-implemented check works wrongly because it's possible to implement one trait multiple times (with different generic arguments). For example, I can't implement `From` trait. Furthermore, the action just silently doesn't work if the trait is already implemented. I think it would better if it generate a duplicate impl.
4. A minor thing: when I type In the popup, autoimport works immediately, so an import remains even if a cancel the action. Furthermore, I can't undo it with `ctrl+Z`.

In general, I think that intellij-based IDEs usually provide more simple and predictable actions/refactorings. A user can achieve any goal by composing these simple refactorings.

> You're right, although I still think that it could be valuable, it could for example generate impl blocks for supertraits of the implemented trait and it would be less typing for the user.

Well, I think this goal is achievable by more simple, "atomic" actions. We just should place the caret to the right place to let the user invoke the next action without typing or even moving the caret. How I imagine it:

1. I press `clt+enter` and Invoke `add trait impl block` intention action that works almost like `add impl block`, but generates `impl |template| for Struct {}` with a template in the trait ref position.
2. I type a trait name. Completion & auto-import works as usual. Then I press `<enter>` to finish the template
3. Then I see the error `E0277 trait bound is not satisfied` that notices me that I should implement a supertrait. 
4. Without moving the caret, I press `clt+enter` and Invoke a quick fix that should solve this error. We don't have such a fix right now, but I guess it's simple to introduce using a code from this PR
5. I repeat (4) for just generated impl if needed
6. Then I press `F2` in order to navigate to `Not all trait items implemented`, press `clt+enter` and invoke `Implement members`
7. I repeat (6) until all errors are fixed

The plan may look complex, but every next step is obvious for a user, and the result is predictable (and all of this can be done lightning-fast using a keyboard).

Summarizing all of the above, I don't think we should land this

Kobzol:
Well :sweat_smile: I can't say that I disagree :) The current implementation indeed does a lot of things and the UI is messy. On the other hand, I think that there may be users that would actually like a single "generate-all-(super)traits-with-all-items" button, but we should try to get close to the functionality of such hypothetical button using atomic actions, as you have described.

I like your proposed workflow, but there's one thing that I would shorten in it, and that is the implementation of all necessary items of a trait, I think that should be done automatically, and without any UI. It seems natural that if I want to implement a trait, the IDE should generate all of the necessary method/type/const "stubs" automatically, without even asking.

So, I propose the following:
1) Improve the completion here (this will be generated by `add impl block`):
```rust
impl /*caret*/ for Struct {}
```
Currently, the completion here allows all paths, which is a bit messy. I propose to change the completion here to only allow traits. In a similar way as was done in https://github.com/intellij-rust/intellij-rust/pull/7451.
After the template finishes here, we could automatically implement all items of the trait, without displaying the member selection UI.
2) Create a quick fix for implementing a missing supertrait. Again, I propose that this quick fix would automatically implement all items of the supertrait in the generated impl block. Maybe we could create another variant that would implement the supertraits recursively.

vlad20012:
> I think that should be done automatically, and without any UI. It seems natural that if I want to implement a trait, the IDE should generate all of the necessary method/type/const "stubs" automatically, without even asking.

Hmm, sounds reasonable

> After the template finishes here, we could automatically implement all items of the trait, without displaying the member selection UI

Let's try it!
I found making template finish callbacks a bit low-level/hacky, so I provided an API for that is #7476.

PR #5341, opened 2020-05-08
CARGO: fetch cargo metadata even with compilation error

### LINKED ISSUE #5016: Plugin fails on servo project
## Environment

* **IntelliJ Rust plugin version:** 0.2.116.2155-201
* **Rust toolchain version:** 1.41.0
* **IDE name and version:** CLion 2020.1 EAP
* **Operating system:** Linux 5.4.19-200.fc31.x86_64

## Problem description

When trying to use CLion and the Rust plugin with https://github.com/servo/servo, the plugin tries to run `~/.cargo/bin/cargo check --message-format json` and fails. That's to be expected, as the command should be something like `./mach check --message-format json` because `mach` does some extra things behind the scene.

There should be a way to specify the command the plugin runs in order to get its input. I might have missed it, but I can't find a way to do it.

## Steps to reproduce

Try to load the https://github.com/servo/servo project.

Undin:
Previously, if `org.rust.cargo.evaluate.build.scripts` feature was enabled and there was compilation error, fetching workspace data failed

Fixes #5016


vlad20012:
:+1: 
bors r+

Issue #1083, opened 2017-03-23
Qualified names for methods should include trait/struct


matklad:

matklad:
The code is in `org.rust.lang.core.psi.ext.RsQualifiedNamedElementKt#getQualifiedName`. This most likely will be surfaced at `org.rust.ide.docs.RsQuickDocumentationTest`. It is possible that the better place to handle this is actually `org.rust.ide.utils.PresentationInfo`

PR #5347, opened 2020-05-09
DOC: add RsQualifiedNameProvider

Kobzol:
This is a revival of a PR originally created by @dimonchik0036: https://github.com/intellij-rust/intellij-rust/pull/2545.

Closes: https://github.com/intellij-rust/intellij-rust/pull/2545, https://github.com/intellij-rust/intellij-rust/pull/5311, https://github.com/intellij-rust/intellij-rust/issues/1083, https://github.com/intellij-rust/intellij-rust/issues/2544


src/main/kotlin/org/rust/ide/docs/RsQualifiedNameProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.intellij.ide.actions.QualifiedNameProvider
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.ext.RsQualifiedNamedElement
+import org.rust.lang.core.psi.ext.qualifiedName
+
+class RsQualifiedNameProvider : QualifiedNameProvider {
+    override fun getQualifiedName(element: PsiElement?): String? = (element as? RsQualifiedNamedElement)?.qualifiedName
+
+    override fun qualifiedNameToElement(fqn: String?, project: Project?): PsiElement? = null
```
Undin, line 17: Just before I created #2544, I had introduced "psi links" feature in [EduTools](https://plugins.jetbrains.com/plugin/10081-edutools/) plugin (plugin for programming language courses in JetBrains IDEs)

In short, an educator can copy reference to psi element, add it to the description of course task, and when you click on it, it navigates you to the corresponding element in the editor.


To get correct PSI element from text `QualifiedNameProvider#qualifiedNameToElement` is used.

Also, there is an [adaptation](https://stepik.org/course/59778) of [rustlings](https://github.com/rust-lang/rustlings) where this feature can be used. So I will think that #2544 is fully implemented when `qualifiedNameToElement` will be supported

But the current implementation of `getQualifiedName` doesn't allow us to do it because it produces the same names for different items sometimes. For example, 

```rust
struct Foo {
    x: i32 // fqn is crate_name::Foo::x
}

impl Foo {
    fn x(&self) -> i32 { self.x }  // fqn is crate_name::Foo::x
}
```

I had an idea to use `org.rust.lang.core.psi.ext.RsQualifiedName` to generate unique strings. 
It already can:
* be created from `RsQualifiedNamedElement` (`org.rust.lang.core.psi.ext.RsQualifiedName.Companion#from(org.rust.lang.core.psi.ext.RsQualifiedNamedElement)`
* provide unique name (see `org.rust.lang.core.psi.ext.RsQualifiedName#toUrlPath`). It's used to generate external documentation URLs so it uses `/` as separator but I think we can introduce similar method with `::` as separator to be more consistent with rust syntax
* be created from string `org.rust.lang.core.psi.ext.RsQualifiedName.Companion#from(java.lang.String)`
* almost find the corresponding psi element `org.rust.lang.core.psi.ext.RsQualifiedName#findPsiElement`. It should be improved because now it requires `context` to get workspace object

src/main/kotlin/org/rust/ide/docs/RsQualifiedNameProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    override fun adjustElementToCopy(element: PsiElement?): PsiElement? = null
```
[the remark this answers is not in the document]
Undin: Looks like we want to implement it as well to support case when `Copy Reference` is invoked for directory/package (in case of IDEA) element

src/main/kotlin/org/rust/ide/docs/RsQualifiedNameProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+
+package org.rust.ide.docs
+
+import com.intellij.ide.actions.QualifiedNameProvider
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.ext.RsQualifiedNamedElement
+import org.rust.lang.core.psi.ext.qualifiedName
+
+class RsQualifiedNameProvider : QualifiedNameProvider {
```
Undin, line 14: I think we need to make custom implementation for `insertQualifiedName` as well because even now (without a suggestion to use `RsQualifiedName`), the combination of `Copy Reference` and `Paste` actions produces invalid reference sometimes, for example

```rust
//- main.rs of `foo` package
struct Foo; // copy reference to this struct

fn main() {
    foo::Foo; // result of paste action. Can be resolved
}
```

src/test/kotlin/org/rust/ide/docs/RsQualifiedNameProviderTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+import com.intellij.ide.actions.CopyReferenceAction
+import org.intellij.lang.annotations.Language
+import org.rust.InlineFile
+import org.rust.RsTestBase
+import org.rust.lang.core.psi.ext.RsReferenceElement
+import org.rust.lang.core.psi.ext.*
+import kotlin.test.assertEquals
+
+class RsQualifiedNameProviderTest : RsTestBase() {
```
Undin, line 16: It would be great to add tests for combination of `Copy Reference` and `Paste` actions

Kobzol:
Ok, I managed to revive the original PR to solve https://github.com/intellij-rust/intellij-rust/issues/1083. Let's resolve https://github.com/intellij-rust/intellij-rust/pull/5311 first.

src/main/kotlin/org/rust/ide/docs/RsQualifiedNameProvider.kt
Kobzol: So if I understand it correctly, the plan is to use `RsQualifiedName` to generate a unique reference path, which could be resolved back to the corresponding element from outside links, and change this unique path to something that can be resolved in Rust when you paste it in the editor?

src/main/kotlin/org/rust/ide/docs/RsQualifiedNameProvider.kt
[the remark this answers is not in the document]
Kobzol: Is is possible to implement `RsQualifiedNamedElement` for a `PsiDirectory`?

src/main/kotlin/org/rust/ide/docs/RsQualifiedNameProvider.kt
Undin: @Kobzol exactly!

Kobzol:
Pasted references are now changed to FQN relative to the place where the reference is pasted and directories containing `mod.rs` are supported. Tests are now checking the result of performing `Copy reference` + `Paste`.

Currently the `QualifiedName` URL is used as the reference string, I'm not sure if we have to use something different? It probably doesn't need to be human readable and if it works after paste, it should be fine. We cannot generate a proper unique FQN for things like struct fields and methods anyway.

Kobzol:
I tried to remove the changes needed for https://github.com/intellij-rust/intellij-rust/pull/5311 (FQN containing owner path for abstractables) to keep quick documentation as it was, but then it gets difficult to use `qualifiedNameRelativeTo`, I would have to copy-paste it and change it to include the owner path in `crateRelativePath` for abstractables or somehow extend it.

Kobzol:
@Kobzol I removed all code related to https://github.com/intellij-rust/intellij-rust/pull/5311 and added more tests.

Kobzol:
Oh, I pinged myself in the last comment :man_facepalming:. @Undin the #5311 is gone now, is there anything else left to do?

Kobzol:
Added support for `mod` decl items. Here's how the `Copy reference` action currently works:
copyref

Kobzol:
@Undin I think that this is now ready for another round of review.

Kobzol:
@Undin ping :innocent:

Kobzol:
It's a bit lonely here :( @Undin ping

Kobzol:
@Undin can we merge this or is there something left to do?

mati865:
What would it take to move it forward?

Kobzol:
@dima74 Could you take a look at this PR?

PR #5352, opened 2020-05-11
INT: generate accessors for struct fields

Kobzol:
This is a revival of a PR originally created by @ttaomae: https://github.com/intellij-rust/intellij-rust/pull/2980

Currently the intention is offered on (non-trait) impl blocks and creates `pub` getters for all non-`pub` fields. `Copy` and primitive types are returned by copy, other types by reference and there is also a special case for `String`, which is returned as `&str`.

The original issue mentioned fixing call sites (presumably by replacing `struct.attr` with `struct.attr()` for fields that receive a getter), however that is probably job for a refactoring and not for an intention?

Some unresolved questions:
- What about setters?
- Should the generated fields be selected with some UI? This might again necessitate a switch to a refactoring. Or, alternatively, we might offer the intention on a specific field, not on an impl block and the getter would only be generated for that specific field, which seems intuitive to me. In this case we would need to find existing (at least at the file local scope) `impl` blocks to avoid generating a new block on each intention execution. Although this might be a bit difficult for generic impl blocks (maybe the user could choose the impl block if there are multiple blocks in the file?).

Closes: https://github.com/intellij-rust/intellij-rust/pull/2980
Closes: https://github.com/intellij-rust/intellij-rust/issues/2471


src/main/resources/intentionDescriptions/GenerateAccessorIntention/after.rs.template
```diff
@@ -0,0 +1,13 @@
+struct S {
+    b: bool,
+    s: String
+}
+
+impl S {
+    <spot></spot>pub fn b(&self) -> bool {
+        self.b
+    }
+    <spot>pub fn s(&self) -> &str {
```
Stzx, line 10: Extra `<spot>`?

Kobzol: Thanks :)

vlad20012 (approved):
Btw in Java it works differently. There is a alt+insert menu with `Getters`, `Setters`, `Getters and Setters` options:
image

In Rust we also have this menu with e.g. `Generate Constructor` action:
image

(see `GenerateConstructorAction`)

Kobzol:
I changed the intention to be a `GenerateGetter` refactoring and refactored the `GenerateConstructorHandler` to factor out common code (which I'll also reuse in a future PR for `GenerateSetter`). I split the changes into several commits to make the review slightly easier.

generate

src/main/kotlin/org/rust/ide/refactoring/generate/getter/GenerateGetterHandler.kt
```diff
@@ -0,0 +1,86 @@
...
+    override fun isFieldValid(member: StructMember, impl: RsImplItem?): Boolean {
+        if (member.field.visibility == RsVisibility.Public) return false
+        return impl?.expandedMembers?.all { it.name != member.argumentIdentifier } ?: true
+    }
+}
+
+private fun getBorrowAndType(type: Ty): Pair<String, Ty> {
+    return when (type) {
+        is TyReference -> Pair("&", type.referenced)
+        is TyPrimitive -> Pair("", type)
```
vlad20012, line 75: Can be a bit better with `type.isMovesByDefault`. Hmm

vlad20012 (approved):
Cool, works as expected. Thanks!

bors r+

Kobzol:
Thanks @ttaomae!

PR #5354, opened 2020-05-11
INSP: simplify and modernize RsWrongTypeArgumentsNumberInspection

Kobzol:
This PR simplifies the implementation of `RsWrongTypeParametersNumberInspection` and renames it to `RsWrongTypeArgumentsNumberInspection`, which is more precise.

Old Rustc errors are removed and now only the unified code `E0107` is used. Tests are cleaned up a bit and now check function/method calls and structs in an unified way.


src/main/kotlin/org/rust/ide/inspections/RsWrongTypeArgumentsNumberInspection.kt
```diff
@@ -30,33 +29,28 @@ class RsWrongTypeParametersNumberInspection : RsLocalInspectionTool() {
...
+        val data = when (o) {
             is RsBaseType -> checkBaseType(nArguments, expectedRequiredParams, expectedTotalParams)
-            is RsMethodCall -> checkMethodCall(nArguments, expectedRequiredParams, expectedTotalParams)
-            is RsCallExpr -> checkCallExpr(nArguments, expectedRequiredParams, expectedTotalParams)
+            is RsMethodCall, is RsCallExpr -> checkFunctionCall(nArguments, expectedRequiredParams, expectedTotalParams)
             else -> null
         } ?: return
 
-        val problemText = "Wrong number of type parameters: expected ${data.expectedText}, found $nArguments [${data.code}]"
+        val problemText = "Wrong number of type arguments: expected ${data.expectedText}, found $nArguments [${data.code}]"
```
Undin, line 51: I haven't made a review yet, but maybe it's worth to use `org.rust.lang.utils.RsDiagnostic` instead of legacy way when we manually inserted error code without link to rust error index?

Kobzol: This is indeed nicer, I switched to `RsDiagnostic`.

Issue #5353, opened 2020-05-11
False positive "Redundant else" inspection when matching on slices


dima74:
## Environment

* **IntelliJ Rust plugin version:** 0.2.122.3086-201-nightly
* **Rust toolchain version:** 1.43.0 (4fb7144ed 2020-04-20) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** Linux 5.6.11-arch1-1

## Problem description
![image](https://user-images.githubusercontent.com/6505554/81600290-d1349e80-93e2-11ea-98c9-118ea04ad3cc.png)

## Steps to reproduce
```rust
fn main() {
    let vec = &[1, 2][..];
    let single = if let &[single] = &vec {
        Some(single)
    } else {
        None
    };
    dbg!(single);
}
```

PR #5358, opened 2020-05-12
INSP: Fix false positive in refutability check

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5353.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsPat.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsPat.kt
@@ -11,7 +11,11 @@ import org.rust.lang.core.resolve.ref.deepResolve
 val RsPat.isIrrefutable: Boolean
     get() = when (val pat = skipUnnecessaryTupDown()) {
         is RsPatSlice ->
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsPat.kt
```diff
@@ -28,7 +26,7 @@ val RsPat.isIrrefutable: Boolean
             pat.patBinding.isIrrefutable
         is RsPatConst ->
             (pat.expr as? RsPathExpr)?.path?.isIrrefutable ?: false
-        is RsPatRange ->
+        is RsPatRange, is RsPatSlice ->
```
Undin, line 29: What's about the following case:
```rust
fn main() {
    if let [a@ ..] = [1, 2, 3] {
        println!("{:?}", a);
    }
}
```

Looks like it's irrefutable slice pattern

mchernyavsky: Fixed.

mchernyavsky:
bors r=Undin

PR #5359, opened 2020-05-12
Use the proper grammar for path fragments in macro_rules!

### LINKED ISSUE #5357: Macros with path fragments fail to expand if the path has a non-turbofish generic
## Environment

* **IntelliJ Rust plugin version:** `0.2.121.3059-201-stable`
* **Rust toolchain version:** `1.45.0-nightly`
* **IDE name and version:** CLion 2020.1
* **Operating system:** NixOS 20.03

## Problem description
When you have a `macro_rules!` macro with a path fragment `$x:path`, the macro fails to expand if you pass a path containing a non-turbofish generic (i.e. `Foo<T>` instead of `Foo::<T>`) to it. I only checked with the experimental macro expander. I do not know how the classic one behaves. 

It seems like the macro expander uses an incorrect grammar for the `path` fragments. [According to the rust reference](https://doc.rust-lang.org/stable/reference/paths.html#paths-in-types) the `::` before generic args should be optional.

## Steps to reproduce
Minimal-ish example:
```rust
trait Foo<T> {}

macro_rules! foo {
    ($p:path) => {
        struct Bar<P: $p> {
            inner: P,
        }
    };
}

foo!(u32); // able to expand
foo!(Foo::<u32>); // able to expand
foo!(Foo<u32>); // ERROR: Failed to expand the macro
```

mrobakowski:
This PR changes the grammar used by the macro expander to match on `path` fragments. The Rust Reference [specifies](https://doc.rust-lang.org/stable/reference/macros-by-example.html#metavariables) that a path fragment should be parsed as a type-like path:
> * `path`: a _[TypePath](https://doc.rust-lang.org/stable/reference/paths.html#paths-in-types)_ style path

Before this PR value-like paths were used, which caused some macros to fail to expand.

Closes #5357

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -75,6 +75,7 @@ mibac138
 mitchhentges
 mkaput
 mrhota
+mrobakowski
 msmorgan
 netvl
 nicholastmosher
--- a/src/main/kotlin/org/rust/lang/core/macros/FragmentKind.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/FragmentKind.kt
@@ -44,7 +44,7 @@ enum class FragmentKind(private val kind: String) {
                 .enter_section_(adaptBuilder, 0, 0, null)
 
             val parsed = when (this) {
-                Path -> RustParser.ValuePathGenericArgs(adaptBuilder, 0)
+                Path -> RustParser.TypePathGenericArgsNoTypeQual(adaptBuilder, 0)
                 Expr -> RustParser.Expr(adaptBuilder, 0, -1)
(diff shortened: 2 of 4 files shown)


vlad20012:
Wow, nice catch! :+1: Thanks!

Please, bump `MacroExpander.EXPANDER_VERSION` number. It is used to track force macro re-expansion if the expansion mechanism is changed.

Undin:
@mrobakowski Welcome to IntelliJ Rust contributors!

mrobakowski:
@Undin glad to be here. This project makes my life so much easier, so I'm happy I could help even a little bit :heart:

PR #5361, opened 2020-05-12
HINT: special case for `impl Iterator` types

### LINKED ISSUE #1803: Display `Iterator<Item=T>` for parameter hints for iterators
Parameter hints are not really useful for iterators right now:

image

I think we can invent some heuristics to decide when we should hide the real type and show something like `Iterator<Item=T>` instead.

vlad20012:
Now any type that implements `Iterator` trait is rendered in inline hints as `impl Iterator<Item=...>` in chained method call hints (#5276). It can be enabled for usual type hints in the future.

image

Fixes #1803

c.c. @Kobzol

### DIFF
--- a/src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
+++ b/src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
@@ -112,6 +112,19 @@ class RsChainMethodTypeHintsProviderTest : RsInlayTypeHintsTestBase(RsChainMetho
         }
     """)
 
+    @ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
+    fun `test iterator special case`() = doTest("""
+        fn main() {
+            vec![1, 2, 3]
+                .into_iter()/*hint text="[:  [impl  [Iterator [< [Item = i32] >]] ]]"*/
+                .map(|x| x as u8)/*hint text="[:  [impl  [Iterator [< [Item = u8] >]] ]]"*/
+                .map(|x| x as u16)/*hint text="[:  [impl  [Iterator [< [Item = u16] >]] ]]"*/
+                .filter(|x| x % 2 == 0)/*hint text="[:  [impl  [Iterator [< [Item = u16] >]] ]]"*/
+                .for_each(|x| {})
+            ;
+        }
+    """)
+
     @Suppress("UnstableApiUsage")
     private fun doTest(@Language("Rust") code: String, showSameConsecutiveTypes: Boolean = true) {
         val service = InlayHintsSettings.instance()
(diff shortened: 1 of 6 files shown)


src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
```diff
@@ -112,6 +112,19 @@ class RsChainMethodTypeHintsProviderTest : RsInlayTypeHintsTestBase() {
         }
     """)
 
+    @ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
+    fun `test iterator special case`() = doTest("""
+        fn main() {
+            vec![1, 2, 3]
+                .into_iter()/*hint text="[:  [impl  [Iterator [< [Item = i32] >]] ]]"*/
+                .map(|x/*hint text="[:  i32]"*/| x as u8)/*hint text="[:  [impl  [Iterator [< [Item = u8] >]] ]]"*/
```
vlad20012, line 120: Hmm, is there a way to disable regular type hints in these tests?

Undin: Looks like there isn't API to do it
I wanted to make this improvement but haven't tried yet

Kobzol: https://github.com/Kobzol/intellij-rust/commit/5af251752cce6c7e3b220f38f7db07730f0705f1 this seems to work. BTW the same situation is in inlay type hints, where I had to change one test to add chain method hints, so if this works it could also be applied there.

Kobzol:
This is great! Another type that might benefit from this is `std::future::Future`. Although it's less common nowadays with `await`, in some cases you still need to map futures and streams, so something like this would be nice:

```rust
futures::future::ready(1)
        .map(|f| f + 1) /* impl Future<Output=i32> */
        .and_then(|f| futures::future::ready(1 + f));
```

But I suppose that each added type will complicate the implementation a lot.

vlad20012:
I wonder why `Future` methods still return concrete types instead of `impl Future` types :thinking: 

In my implementation, I moved the most complicated logic to `selectAllProjectionsStrict`, so ann one more trait seems relatively simple

Kobzol:
It's mostly for practical reasons in libraries I guess. For example inside `tokio` the low-level stuff often has to be implemented using manual `impl Future for X` or `impl Stream for X`. In that case you usually store some inner future inside a struct and you can't do that without boxing if you don't have access to the concrete type.

src/test/kotlin/org/rust/ide/hints/RsChainMethodTypeHintsProviderTest.kt
sirgl: @vlad20012 @Undin Probably you should use here provider specific test case: `com.intellij.testFramework.utils.inlays.InlayHintsProviderTestCase`, it allows to test different settings and only a given provider

vlad20012:
I've limited it to chained method call hints for now.

PR #5362, opened 2020-05-12
PERF&MACRO: slightly optimize macro matching

vlad20012:
Some optimization in `FragmentKind.parse`: do less Grammar Kit stuff, use more lightweight parsing methods.

This slightly speeds up macro expansion


vlad20012:
bors retry

PR #5364, opened 2020-05-12
T: enable only one rust inlay hints provider for test class

Undin:


Undin:
CC @vlad20012 @Kobzol

Issue #1409, opened 2017-06-30
Mutability can aid auto-completion


gilescope:
I'm in two minds on this one.

If you have a non-mutable variable, then the methods that are auto-completed should be subset to only show methods that take a non mutable self.

Alternatively one could show the mutable versions, but if a mutable method was selected one could change the var definition to be a 'mut'. This would be more 'flow', I worry the former may be a gotcha.

Thoughts?

mkaput:
I think instead of hiding we could apply some colouring to mutable methods in immutable context

Issue #5196, opened 2020-04-05
Smart completion


sirgl:
IntelliJ platform has a concept of smart completion (when user invokes it only those elements that exactly match the context are suggested). When you work with Kotlin or Java you get used to it very much. I think having such feature for Rust is very important.

Smart completion works exactly like simple completion, but also filters variants by the type expected in the place of completion.

## Environment

* **IntelliJ Rust plugin version:** 0.2.119.2158-201
* **Rust toolchain version:** 1.44.0-nightly (7ceebd98c 2020-03-17) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 Beta (CL-201.6668.10)
* **Operating system:** macOS 10.14.5

PR #5369, opened 2020-05-13
COMP: Deprioritize completion of &mut self methods for const references and bindings

Kobzol:
This PR attempts to sort methods in completion based on mutability of `self`. If a method call is performed on a const reference or a non-mut binding,`&mut self` methods will be deprioritized.

The implementation needs additional work that I wanted to discuss here. I needed to somehow pass the method call (`RsFieldLookup`) to `ScopedBaseCompletionEntity::getBasePriority` (I didn't know how else to get it from there). `ImplLookup` was already being passed down the completion chain, so I just slapped the element to the lookup, which is not optimal. Is there an easier way of passing down the element? Should I add it to the `RsCompletionContext` and pass it down the completion chain of methods and processors?

The implementation now uses a very simple heuristic for deciding if the method should be deprioritized. Let me know if I should improve it (or how :) ).

Completion could possibly receive a larger overhaul to support use cases like smart completion (https://github.com/intellij-rust/intellij-rust/issues/5196).

Fixes: https://github.com/intellij-rust/intellij-rust/issues/1409

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/LookupElements.kt
@@ -31,10 +31,7 @@ import org.rust.lang.core.types.emptySubstitution
 import org.rust.lang.core.types.implLookup
 import org.rust.lang.core.types.infer.RsInferenceContext
 import org.rust.lang.core.types.infer.TypeFolder
-import org.rust.lang.core.types.ty.Ty
-import org.rust.lang.core.types.ty.TyNever
-import org.rust.lang.core.types.ty.TyTypeParameter
-import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.ty.*
 import org.rust.lang.core.types.type
 
 const val KEYWORD_PRIORITY = 80.0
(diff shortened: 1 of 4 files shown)


mchernyavsky (commented):
> Should I add it to the `RsCompletionContext` and pass it down the completion chain of methods and processors?

Yes, you can pass the element to the completion context instead of `implLookup`:

```rust
data class RsCompletionContext(
    val context: RsElement? = null,
    val expectedTy: Ty? = null,
    val isSimplePath: Boolean = false
) {
    val lookup: ImplLookup? get() = context?.implLookup
}
```

Kobzol:
Ok, I moved it to `RsCompletionContext`.

I think that the lookup should be computed eagerly (as it was before) to avoid recomputing it on every access (i.e. I removed `get()`):
```kotlin
data class RsCompletionContext(
    val context: RsElement? = null,
    val expectedTy: Ty? = null,
    val isSimplePath: Boolean = false
) {
    val lookup: ImplLookup? = context?.implLookup
}
```

mchernyavsky (approved):
Thanks!

bors r+

Issue #5194, opened 2020-04-04
Supertraits inspection


AndreKR:
This is a feature request for an inspection that detects missing supertraits.

## Environment

* **IntelliJ Rust plugin version:** 0.2.119.2172-193
* **Rust toolchain version:** 1.41.1
* **IDE name and version:** CLion 2019.3.5
* **Operating system:** Windows 7

## Problem description
```rust
trait Edible {
    fn eat(&self);
}

trait StillEdible: Edible {}

struct Fruit {}

impl StillEdible for Fruit {
}
```

The compiler correctly shows an error:
> ```text
> error[E0277]: the trait bound `Fruit: Edible` is not satisfied
>   --> src\lib.rs:11:6
>    |
> 11 | impl StillEdible for Fruit {
>    |      ^^^^^^^^^^^ the trait `Edible` is not implemented for `Fruit`
> ```

This error should also show in the IDE.

PR #5375, opened 2020-05-14
INSP: check if supertrait is implemented

Kobzol:
This PR adds inspection that checks if a supertrait is implemented:
```rust
trait A {}
trait C: A {}

struct S;
impl <error descr="the trait bound `[S]: A` is not satisfied [E0277]">C</error> for S {}
```

All supertraits of the trait are checked, but the check is not recursive (i.e. supertraits of supertraits are not checked). If you try to implement the missing trait, the recursive requirements will pop on that `impl`, so that should be fine.

I'm also thinking of adding a quick fix that would add `impl MissingTrait for Type` above the impl with the error and apply `ImplementMembersFix` right away. Do you agree?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5194

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -17,12 +17,15 @@ import com.intellij.psi.util.PsiTreeUtil
 import org.rust.cargo.project.workspace.CargoWorkspace.Edition
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.annotator.fixes.*
+import org.rust.ide.presentation.getStubOnlyText
 import org.rust.ide.refactoring.RsNamesValidator.Companion.RESERVED_LIFETIME_NAMES
 import org.rust.ide.utils.isCfgUnknown
 import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.core.*
 import org.rust.lang.core.FeatureAvailability.CAN_BE_ADDED
 import org.rust.lang.core.FeatureAvailability.NOT_AVAILABLE
+import org.rust.lang.core.macros.MacroExpansionMode
+import org.rust.lang.core.macros.macroExpansionManager
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.IDENTIFIER
 import org.rust.lang.core.psi.ext.*
@@ -32,6 +35,8 @@ import org.rust.lang.core.resolve.namespaces
 import org.rust.lang.core.resolve.ref.deepResolve
 import org.rust.lang.core.types.*
 import org.rust.lang.core.types.consts.asLong
+import org.rust.lang.core.types.infer.containsTyOfClass
+import org.rust.lang.core.types.infer.substitute
 import org.rust.lang.core.types.ty.*
 import org.rust.lang.utils.RsDiagnostic
 import org.rust.lang.utils.RsErrorCode
@@ -524,6 +529,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         val trait = traitRef.resolveToTrait() ?: return
         checkForbiddenImpl(holder, traitRef, trait)
         checkImplDropForNonAdtError(holder, impl, traitRef, trait)
+        checkSuperTraitImplemented(holder, impl, trait)
         checkImplBothCopyAndDrop(holder, impl, trait)
         val traitName = trait.name ?: return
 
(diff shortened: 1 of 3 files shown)


vlad20012:
I'd just say that you work with types incorrectly here. I don't have enough time to explain it now, I'm sorry.

Kobzol:
I improved handling of generic supertraits somewhat, but it's possible that it's still too naive :)

Kobzol:
Btw < > in the annotation is wrongly escaped, if I recall correctly, that's a platform bug?

vlad20012:
Looks like type magic works now! :)

About other questions:
1. Don't use type rendering to display TraitRef. It's not a type! Use `RsTraitRef.getStubOnlyText` - it's PSI-based rendering that also allows to substitute types. This should fix the issue with `dyn`
2. In order to fix an issue with `<>`, use `Ty.escaped` or `escapeTy(String)` functions. Don't use `escapeString` after it! As far as I know, this platform bug should be fixed on the 202 platform, but I haven't been tested it.

Also, why the type is wrapped into `[ ]` brackets? It can be confusing because there is a slice type that is also wrapped in brackets.
image

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -547,6 +551,26 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+
+        val typeRef = impl.typeReference ?: return
+        val traitRef = impl.traitRef ?: return
+        val type = typeRef.type
+        val supertraits = trait.typeParamBounds?.polyboundList?.mapNotNull { it.bound } ?: return
+        val lookup = typeRef.implLookup
+
+        for (bound in supertraits) {
+            val boundTrait = bound.traitRef?.resolveToBoundTrait() ?: continue
+            val localBound = boundTrait.substitute(impl.implementedTrait?.subst ?: emptySubstitution)
```
vlad20012, line 565: Let's extract `impl.implementedTrait?.subst` out of the loop. Also it can be replaced with `traitRef.resolveToBoundTrait()`

vlad20012: Well, let's also suppress this check if implemented trait substitution contains unresolved types:
image

Kobzol:
Hmm, I could swear that rustc showed the type inside `[ ]`, but it doesn't seem to do that. I removed it.

The problem with escaped `< >` is related to the rendered `TraitRef`, not a `Ty`. If I use `escapeTy` on `RsTraitRef.getStubOnlyText`, it escapes the text in a very weird way. With the current implementation it looks OK in the GUI tooltip, but it's escaped wrongly in tests and in the IDE status bar.

Regarding the subtitution, it seems to me that I need to bind the trait twice.
```rust
trait A<T> {}
trait B<R>: A<R> {}

impl B<u32> for u32 {}
```
`resolveToBoundTrait()` gets the bound supertrait `A<R>` from `trait B<R>: A<R>` and then I need to use the local substitution from `impl B<u32> ...` to get `A<u32>`. But maybe it can be simplified further.

vlad20012:
> With the current implementation it looks OK in the GUI tooltip, but it's escaped wrongly in tests and in the IDE status bar.

Well, looks like it works similarly with other type-related inspections :shrug: 

> Regarding the subtitution, it seems to me that I need to bind the trait twice

I'm not sure I understand you, but I see you doing things right. I.e. we should do this stuff on each loop iteration:
```
val boundTrait = requiredTrait.resolveToBoundTrait() ?: continue
val locallyBoundTrait = boundTrait.substitute(substitution)
```

Kobzol:
OK, I added the escaped output to tests.

vlad20012:
Could you rebase and squash the commits please?

Kobzol:
Done.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -548,6 +554,30 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+    private fun checkSuperTraitImplemented(holder: RsAnnotationHolder, impl: RsImplItem, trait: RsTraitItem) {
+        if (impl.project.macroExpansionManager.macroExpansionMode !is MacroExpansionMode.New) return
+
+        val traitRef = impl.traitRef ?: return
+        val typeRef = impl.typeReference ?: return
+        val type = typeRef.type
+        val supertraits = trait.typeParamBounds?.polyboundList?.mapNotNull { it.bound } ?: return
+        val lookup = typeRef.implLookup
+
+        val substitution = impl.implementedTrait?.subst ?: emptySubstitution
```
vlad20012, line 566: I found one more false-positive:
```rust
trait Tr1<A=Self> {}
trait Tr2<A=Self> : Tr1<A> {}

struct Sss;
impl Tr1 for Sss {}
impl Tr2 for Sss {}
```

Should be fixed by this:
```suggestion
        val selfSubst = mapOf(TyTypeParameter.self() to type).toTypeSubst()
        val substitution = (impl.implementedTrait?.subst ?: emptySubstitution).substituteInValues(selfSubst) + selfSubst
```

Kobzol: Thank you!

vlad20012:
We have `RsRealProjectAnalysisTest` - a test that launches `RsErrorAnnotator` and all our inspections on a set of real projects (like Cargo or Tokio) and then prints all errors found in these projects. 

This test doesn't run on CI because it is slow. Also, it's not documented anywhere because it was a temporary solution. We planned to set up testing infrastructure like [crater](https://github.com/rust-lang/crater) to automatically find regressions in our analysis. But we still don't have it. 

So, usually, we run real project test when on PR that changing something in our inspections/annotators to prevent new false-positives. But now there are too many of these PRs, so I decided to ask contributors to run these tests by themselves.

The test is located [here](https://github.com/intellij-rust/intellij-rust/blob/6c7af1ce47cafb314a1d3e3863ef1e2cd5ad0c0c/src/test/kotlin/org/rustPerformanceTests/RsRealProjectAnalysisTest.kt#L21). It can be launched like usual, but it requires some preparations. You need to go to `intellij-rust` repo directory, then go to `testData` directory, then clone some git repos that are listed [here](https://github.com/intellij-rust/intellij-rust/blob/6c7af1ce47cafb314a1d3e3863ef1e2cd5ad0c0c/src/test/kotlin/org/rustPerformanceTests/RsRealProjectTestBase.kt#L81).
```
git clone https://github.com/rust-lang/cargo
git clone https://github.com/blackbeam/mysql_async
...
```
Then comment `test analyze rustc` test. It takes about 30 minutes to run and usually just crashes in the middle. 

Then you can run the test. It WILL fail with a set of false-positives because we HAVE some known false-positives on the stable plugin now. So it's needed to check there are no new false-positives caused y some new inspection. If there are, you should investigate and fix them.

vlad20012:
@Kobzol I found previous false-positives caused by this check using `RsRealProjectAnalysisTest`. You could try to use it when working on inspections/error annotators. It can help to avoid false-positives and speed up review cycles.

Kobzol:
Thanks for the information! I'll try to run it on my other/future inspection/annotator PRs.

Issue #5378, opened 2020-05-14
"<unknown> doesn't implement Display" false positive


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.43.0 (4fb7144ed 2020-04-20) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** macOS 10.15.4

## Problem description

The plugin can't infer type after `par_iter` call (old problem with `rayon`) and complains about unimplemented `Display` (new problem). Could the error be suppressed for such cases?

<img width="659" alt="error" src="https://user-images.githubusercontent.com/2500320/81978623-785c4480-9634-11ea-8e1d-9016234d9870.png">

This is a recent regression, as 0.2.122.3078-201-nightly doesn't complain.

## Steps to reproduce

```toml
rayon = "1.3.0"
```

```rust
use rayon::prelude::*;
use std::thread::sleep;
use std::time::Duration;

fn main() {
    let mut v = Vec::with_capacity(500);
    for i in 0..500 {
        v.push(i);
    }

    let v2: Vec<i32> = (&v)
        .par_iter()
        .map(|&x| {
            sleep(Duration::from_millis(100));
            println!("{}", x);
            x * x
        })
        .collect();

    println!("{:?}", &v2[490..500]);
}
```

PR #5380, opened 2020-05-14
ANN: don't annotate unknown types in format macros

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5378

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
@@ -28,6 +28,7 @@ import org.rust.lang.core.types.TraitRef
 import org.rust.lang.core.types.implLookup
 import org.rust.lang.core.types.infer.containsTyOfClass
 import org.rust.lang.core.types.ty.TyInteger
+import org.rust.lang.core.types.ty.TyNever
 import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.ty.stripReferences
 import org.rust.lang.core.types.type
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -411,7 +412,7 @@ private fun checkParameterTraitMatch(argument: RsFormatMacroArg, parameter: Form
     val requiredTrait = parameter.type?.resolveTrait(argument.knownItems) ?: return null
 
     val expr = argument.expr
-    if (!expr.implLookup.canSelectWithDeref(TraitRef(expr.type, requiredTrait.withSubst()))) {
+    if (expr.type != TyUnknown && !expr.implLookup.canSelectWithDeref(TraitRef(expr.type, requiredTrait.withSubst()))) {
```
vlad20012, line 415:
```suggestion
    if (!expr.type.containsTyOfClass(TyUnknown::class.java) && !expr.implLookup.canSelectWithDeref(TraitRef(expr.type, requiredTrait.withSubst()))) {
```

Kobzol: Thanks.

Kobzol:
Should I also ignore `Never` (!) types? I'm not sure why anyone would want to print a ! type, but currently we show a false positive error for something like `println!("{}", std::process::exit(0));`.

vlad20012:
> Should I also ignore Never (!) types? 

Probably yes. I wonder why it doesn't implement `Display `trait... Or how it works

Kobzol:
Hmm, for some reason in tests `ImplLookup` claims that `!` implements `Display`, but in the IDE it doesn't. I explicitly ignore both `TyUnknown` and `TyNever` now.

Issue #5269, opened 2020-04-21
`Extract enum variant` intention doesn't replace indirect usages


mchernyavsky:
## Environment

* **IntelliJ Rust plugin version:** 0.2.121.3051-201-nightly
* **Rust toolchain version:** 1.42.0 (b8cedc004 2020-03-09) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 (CL-201.6668.126)
* **Operating system:** macOS 10.15.4

## Problem description

Before:
```rust
mod foo {
    pub enum Foo { Bar/*caret*/(i32) }
}

fn main() {
    let ctr = foo::Foo::Bar;
    let f = ctr(123);
}
```

After:
```rust
mod foo {
    pub struct Bar(pub i32);

    pub enum Foo { Bar(Bar) }
}

fn main() {
    let ctr = foo::Foo::Bar;
    let f = ctr(123); // <- nothing changed. Leads to invalid code
}
```

PR #5381, opened 2020-05-15
REF: fix indirect tuple usage replacement in RsExtractEnumVariant

Kobzol:
This PR fixes tuple struct constructor references replacing in `RsExtractEnumVariant`.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5269
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5565


Kobzol:
After I examined https://github.com/intellij-rust/intellij-rust/issues/5565 I realized that the old reference replacing was not correct. I changed the tuple struct replacement to explicitly check only the parent (and not all ancestors) for a call expression to avoid things like `foo(Enum::Variant)` -> `foo(Enum::Variant(Variant))`. References to a tuple constructor (`Enum::Variant` without a function call) are now replaced with a lambda, because they represent a function.

mchernyavsky:
bors retry

Issue #5385, opened 2020-05-16
False-positive "Argument never used" when using backslash in raw string in format macro


dima74:
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.3094-201-nightly
* **Rust toolchain version:** 1.43.0 (4fb7144ed 2020-04-20) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** Linux 5.6.11-arch1-1

## Problem description
![image](https://user-images.githubusercontent.com/6505554/82115130-5b318e00-977a-11ea-89ac-49c2c9734e56.png)

## Steps to reproduce
```rust
fn main() {
    println!(r"\{}", 1);
}
```

PR #5392, opened 2020-05-18
ANN: fix raw format strings in RsFormatMacroAnnotator

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5385


src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
```diff
@@ -188,7 +188,14 @@ private val formatParameterParser = Regex("""(?x) # enable comments
 private fun parseParameters(formatStr: RsLitExpr): ParseContext? {
     val literalKind = (formatStr.kind as? RsLiteralKind.String) ?: return null
     val rawTextRange = literalKind.offsets.value ?: return null
-    val (unescapedText, sourceMap, _) = parseRustStringCharacters(rawTextRange.substring(literalKind.node.text))
+    val (unescapedText, sourceMap) = if (formatStr.rawStringLiteral != null) {
```
Undin, line 191:
```suggestion
    val (unescapedText, sourceMap) = if (literalKind.node.elementType in RS_RAW_LITERALS) {
```

Kobzol: Byte literals are not allowed as format strings though. I wonder if we have a test for that, I'll check.

Kobzol: There was no check, so I added one. Instead of marking the byte string as an error, I just ignore the format macro, because we still get false positives in resolving the macros.

PR #5402, opened 2020-05-19
GRAM: Do not attach EOL comments without docs to items

ortem:
Previously, any EOL comments were attached to items, even if there were no doc comments.
Now only doc and EOL+doc comments are attached.

Extend selection now does not include EOL comment in:
```Rust
trait T {
    fn foo() {}
    
    // comment
    
    fn bar/* <extend selection> */() {}
}
```

Move line down now works in:
```Rust
trait T {
    // /* <move down> */first
    // second
    
    fn foo() {}
}
```


src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -142,10 +142,13 @@ object RustParserUtil : GeneratedParserUtilBase() {
     @JvmField
     val ADJACENT_LINE_COMMENTS = WhitespacesAndCommentsBinder { tokens, _, getter ->
         var candidate = tokens.size
+        var containsDocComment = false
```
Undin, line 145: Since you change parsing, don't forget to increase `org.rust.lang.core.parser.RustParserDefinition#PARSER_VERSION`

ortem:
The suggested behaviour is quite controversial. We discussed it in person with @vlad20012 and finally have decided to close this PR and leave the current behaviour as it is.

Issue #5401, opened 2020-05-19
"Go to declaration" not working for struct field in destructoring when using type alias


dima74:
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.3100-201-nightly
* **Rust toolchain version:** 1.43.0 (4fb7144ed 2020-04-20) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** Linux 5.6.11-arch1-1

## Problem description
Use following code:

```rust
pub struct Foo { field: i32 }
pub type FooAlias = Foo;

fn main() {
    let foo = Foo { field: 0 };
    let FooAlias { /*caret*/field } = foo;
}
```

And invoke "Go to" action (Ctrl+B):
![image](https://user-images.githubusercontent.com/6505554/82292818-f61da880-99c4-11ea-91d7-4e801b874cd6.png)

---
If I remove type alias, then "Go to" works correctly:
```rust
pub struct Foo { field: i32 }

fn main() {
    let foo = Foo { field: 0 };
    let Foo { /*caret*/field } = foo;
}
```

Also "Go to" works correctly when constructing struct using type alias:

```rust
pub struct Foo { field: i32 }
pub type FooAlias = Foo;

fn main() {
    let foo = FooAlias { /*caret*/field: 0 };
}
```

PR #5405, opened 2020-05-19
RES: Fix struct field resolve when using type alias

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5401.


mchernyavsky:
bors retry

Issue #5388, opened 2020-05-17
Uncommenting a doc comment


CDirkx:
## Environment

* **IntelliJ Rust plugin version:** 0.2.121.3059-201-stable
* **Rust toolchain version:** Rust 1.45 nightly
* **IDE name and version:** CLion 2020.1.1
* **Operating system:** Windows 10

## Problem description
Uncommenting (`CTRL+ '/'`) on a doc comment (`///`) results in a single remaining slash:

before:
```rust
/// this is a doc comment
```

after:
```rust
/ this is a doc comment
```

expected:
```rust
this is a doc comment
```

## Solution
The implementation right now is to just remove two '/' characters at the beginning of the line. My proposal would be to extend this to removing two or three characters, depending on how many are written.

vlad20012:
for someone who decides to fix it: it's possible by implementing `SelfManagingCommenter` to `RsCommenter`

Kobzol:
I'll try to implement it. If I understand it correctly, using `SelfManagingCommenter` is a low-level manual comment implementation, therefore if we implement it, we no longer need to implement `Commenter` for `RsCommenter`?

Undin:
@Kobzol we still need to implement `Commenter` because we use `lang.commenter` extension point to register our commenter and it requires to implement `Commenter` interface

vlad20012:
> `SelfManagingCommenter` is a low-level manual comment implementation

Yes. Moreover, there are no open-source implementations of `SelfManagingCommenter` =/

"Default" implementation (used for regular `Commenter`) can be found at `CommentByBlockCommentHandler`/`CommentByLineCommentHandler`. It looks scary. I don't know what parts of them are really needed in the case of Rust.

Also, take a look at `SelfManagingCommenterUtil`.

PR #5416, opened 2020-05-21
TYPE: implement SelfManagingCommenter

Kobzol:
This PR solves https://github.com/intellij-rust/intellij-rust/issues/5388 with a pretty heavyweight solution of implementing `SelfManagingCommenter`. The implementation is pretty basic, but it passes all the original `RsCommenter` tests and some new ones that I have added.

It seems to work for basic use cases, if you can find some counter-examples where it breaks, let me know and I'll add them to the test suite.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5388

### DIFF
--- a/src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
+++ b/src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
@@ -5,6 +5,7 @@
 
 package org.rust.ide.commenter
 
+import com.intellij.application.options.CodeStyle
 import com.intellij.codeInsight.generation.CommenterDataHolder
 import com.intellij.codeInsight.generation.SelfManagingCommenter
 import com.intellij.codeInsight.generation.SelfManagingCommenterUtil
@@ -16,10 +17,8 @@ import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiFile
 import com.intellij.psi.tree.IElementType
 import com.intellij.util.text.CharArrayUtil
+import org.rust.lang.RsLanguage
 import org.rust.lang.core.parser.RustParserDefinition.Companion.BLOCK_COMMENT
-import org.rust.lang.core.parser.RustParserDefinition.Companion.EOL_COMMENT
-import org.rust.lang.core.parser.RustParserDefinition.Companion.INNER_EOL_DOC_COMMENT
-import org.rust.lang.core.parser.RustParserDefinition.Companion.OUTER_EOL_DOC_COMMENT
 import org.rust.lang.core.psi.RS_EOL_COMMENTS
 import org.rust.lang.doc.psi.RsDocKind
 
@@ -64,12 +63,8 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter, SelfManagingComm
         endOffset: Int,
         document: Document,
         data: CommentHolder?
-    ): TextRange {
-        val sequence = document.charsSequence
-        val start = CharArrayUtil.shiftForward(sequence, startOffset, " \t\n")
-        val end = CharArrayUtil.shiftBackward(sequence, endOffset - 1, " \t\n")
-        return SelfManagingCommenterUtil.insertBlockComment(start, end + 1, document, blockCommentPrefix, blockCommentSuffix)
-    }
+    ): TextRange =
+        SelfManagingCommenterUtil.insertBlockComment(startOffset, endOffset, document, blockCommentPrefix, blockCommentSuffix)
 
     override fun uncommentBlockComment(
         startOffset: Int,
@@ -84,25 +79,18 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter, SelfManagingComm
     }
 
     override fun commentLine(line: Int, offset: Int, document: Document, data: CommentHolder) {
-        document.insertString(offset, "// ")
+        val addSpace = CodeStyle.getLanguageSettings(data.file, RsLanguage).LINE_COMMENT_ADD_SPACE
+        document.insertString(offset, "//" + if (addSpace) " " else "")
     }
 
     override fun uncommentLine(line: Int, offset: Int, document: Document, data: CommentHolder) {
-        val element = getStartLineComment(line, document, data.file) ?: return
-        val prefix = getEolCommentPrefix(element.tokenType) ?: return
-        val prefixWithSpace = "$prefix "
-        val length = if (element.text.startsWith(prefixWithSpace)) {
-            prefixWithSpace.length
-        } else {
-            prefix.length
-        }
-        document.deleteString(offset, offset + length)
+        val prefixLen = LINE_PREFIXES.find { CharArrayUtil.regionMatches(document.charsSequence, offset, it) }?.length
+            ?: return
+        val hasSpace = CharArrayUtil.regionMatches(document.charsSequence, offset + prefixLen, " ")
+        document.deleteString(offset, offset + prefixLen + if (hasSpace) 1 else 0)
     }
 
-    override fun getCommentPrefix(line: Int, document: Document, data: CommentHolder): String? {
-        val token = getStartLineComment(line, document, data.file)?.tokenType ?: return null
-        return getEolCommentPrefix(token)
-    }
+    override fun getCommentPrefix(line: Int, document: Document, data: CommentHolder): String? = lineCommentPrefix
 
     override fun createBlockCommentingState(
         selectionStart: Int,
@@ -119,6 +107,10 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter, SelfManagingComm
         file: PsiFile
     ): CommentHolder? =
         CommentHolder(file)
+
+    companion object {
+        private val LINE_PREFIXES = listOf(RsDocKind.OuterEol.prefix, RsDocKind.InnerEol.prefix, "//")
+    }
 }
 
 private fun getStartLineComment(line: Int, document: Document, file: PsiFile): PsiComment? {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
```diff
@@ -27,8 +41,100 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter {
...
+        val prefixWithSpace = "$prefix "
+        val length = if (element.text.startsWith(prefixWithSpace)) {
+            prefixWithSpace.length
+        } else {
+            prefix.length
+        }
+        document.deleteString(offset, offset + length)
+    }
+
+    override fun getCommentPrefix(line: Int, document: Document, data: CommentHolder): String? {
```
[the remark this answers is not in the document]
Kobzol: Thank you. It really wouldn't hurt for these interface methods to have a small comment describing their meaning :D

src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
```diff
@@ -27,8 +41,100 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter {
...
+    override fun insertBlockComment(
+        startOffset: Int,
+        endOffset: Int,
+        document: Document,
+        data: CommentHolder?
+    ): TextRange {
+        val sequence = document.charsSequence
+        val start = CharArrayUtil.shiftForward(sequence, startOffset, " \t\n")
+        val end = CharArrayUtil.shiftBackward(sequence, endOffset - 1, " \t\n")
+        return SelfManagingCommenterUtil.insertBlockComment(start, end + 1, document, blockCommentPrefix, blockCommentSuffix)
```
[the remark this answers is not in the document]
Kobzol: You're right, let's not change the behaviour here.

src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
```diff
@@ -27,8 +41,100 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter {
...
+        data: CommentHolder?
+    ) =
+        SelfManagingCommenterUtil.uncommentBlockComment(startOffset, endOffset, document, blockCommentPrefix, blockCommentSuffix)
+
+    override fun isLineCommented(line: Int, offset: Int, document: Document, data: CommentHolder): Boolean {
+        return getStartLineComment(line, document, data.file)?.isEolComment ?: false
+    }
+
+    override fun commentLine(line: Int, offset: Int, document: Document, data: CommentHolder) {
+        document.insertString(offset, "// ")
```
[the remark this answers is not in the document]
Kobzol: Should the code style be also respected when uncommenting? i.e. when `LINE_COMMENT_ADD_SPACE` is false, do not remove space after `//` even if there is a space?

Kobzol:
Added test for space after line comment settings and fixed the rest. Should `getCommentPrefix` respect the space after line comment settings and return `// ` if it's turned on? It seems to work if I just return `//` everytime.

There's something weird with the `test move caret after commenting empty line` test. Instead of
```rust
//<caret>
fn foo() {}
```
it returns
```rust
//
<caret>fn foo() {}
```
even though when I try it in the IDE, it behaves correctly.

src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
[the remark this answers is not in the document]
vlad20012: The default implementation checks the settings... and yes, I think we should. It may be reasonable for uncommenting an indented block:
```rust
//if foo {
//    bar(); // here the space can be wrongly removed
//}
```

vlad20012:
> Should getCommentPrefix respect the space after line comment settings and return // if it's turned on?

It looks like no, simple `//` is enough

> There's something weird with the `test move caret after commenting empty line` test

This was tricky, but I found the reason. There is a mutable state in `ActionManagerImpl` that outlives a test. It's info about previous run action - `myPrevPerformedActionId`. Then it's used in `CommentByLineCommentHandler.invoke` - see startingNewLineComment condition. I don't know is it a bug or a feature of the test framework. 

I've changed the code of `checkOption` like this and looks like it works as expected now:

```kotlin
private fun checkOption(optionProperty: KMutableProperty0<Boolean>,
                            @Language("Rust") before: String,
                            @Language("Rust") afterOn: String,
                            @Language("Rust") afterOff: String,
                            actionId: String,
                            trimIndent: Boolean = true) {
        val initialValue = optionProperty.get()
        optionProperty.set(true)
        try {
            checkEditorAction(before, afterOn, actionId, trimIndent = trimIndent)
            resetActionManagerState()
            optionProperty.set(false)
            checkEditorAction(before, afterOff, actionId, trimIndent = trimIndent)
            resetActionManagerState()
        } finally {
            optionProperty.set(initialValue)
        }
    }

    private fun settings() = CodeStyle.getSettings(project).getCommonSettings(RsLanguage)

    /**
     * Resets [com.intellij.openapi.actionSystem.impl.ActionManagerImpl.myPrevPerformedActionId].
     * Othervise it affect [com.intellij.codeInsight.generation.CommentByLineCommentHandler.invoke]
     * (see `startingNewLineComment` condition there).
     */
    private fun resetActionManagerState() {
        myFixture.performEditorAction(EMPTY_ACTION_ID)
    }

    override fun setUp() {
        super.setUp()
        ActionManager.getInstance().registerAction(EMPTY_ACTION_ID, EmptyAction.createEmptyAction("empty", null, true))
    }

    override fun tearDown() {
        try {
            super.tearDown()
        } finally {
            ActionManager.getInstance().unregisterAction(EMPTY_ACTION_ID)
        }
    }

    companion object {
        private const val EMPTY_ACTION_ID = "!!!EmptyAction"
    }
```

Kobzol:
Thank you!

vlad20012:
Hmm
bors try

src/test/kotlin/org/rust/ide/commenter/RsCommenterTest.kt
```diff
@@ -147,4 +242,62 @@ class RsCommenterTest : RsTestBase() {
...
+        ActionManager.getInstance().registerAction(EMPTY_ACTION_ID, EmptyAction.createEmptyAction("empty", null, true))
+    }
+
+    override fun tearDown() {
+        try {
+            super.tearDown()
+        } finally {
+            ActionManager.getInstance().unregisterAction(EMPTY_ACTION_ID)
+        }
+    }
```
Undin, line 298:
```suggestion
    override fun tearDown() {
        ActionManager.getInstance().unregisterAction(EMPTY_ACTION_ID)
        super.tearDown()
    }
```
It's recommended way to cleanup our custom state before `super.tearDown()`.

Alternatively, you can drop `tearDown` at all and add the following code in `setUp`:
```kotlin
    override fun setUp() {
        super.setUp()
        val actionManager = ActionManager.getInstance()
        actionManager.registerAction(EMPTY_ACTION_ID, EmptyAction.createEmptyAction("empty", null, true))
        Disposer.register(testRootDisposable, Disposable {
            actionManager.unregisterAction(EMPTY_ACTION_ID)
        })
    }
```

src/test/kotlin/org/rust/ide/commenter/RsCommenterTest.kt
```diff
@@ -147,4 +242,62 @@ class RsCommenterTest : RsTestBase() {
...
+            resetActionManagerState()
+            optionProperty.set(false)
+            checkEditorAction(before, afterOff, actionId, trimIndent = trimIndent)
+            resetActionManagerState()
+        } finally {
+            optionProperty.set(initialValue)
+        }
+    }
+
+    private fun settings() = CodeStyle.getSettings(project).getCommonSettings(RsLanguage)
```
Undin, line 276: Explicit returning type

src/test/kotlin/org/rust/ide/commenter/RsCommenterTest.kt
```diff
@@ -147,4 +242,62 @@ class RsCommenterTest : RsTestBase() {
...
+        //<caret>
+        fn foo() {}
+    """, IdeActions.ACTION_COMMENT_LINE)
+
+    private fun checkOption(optionProperty: KMutableProperty0<Boolean>,
+                            @Language("Rust") before: String,
+                            @Language("Rust") afterOn: String,
+                            @Language("Rust") afterOff: String,
+                            actionId: String,
+                            trimIndent: Boolean = true) {
```
Undin, line 262: nit: [formatting](https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting)

src/test/kotlin/org/rust/ide/commenter/RsCommenterTest.kt
Kobzol: Thanks, let's see if this appeases CI.

vlad20012:
Let's also add this:
```kotlin
override fun checkEditorAction(before: String, after: String, actionId: String, trimIndent: Boolean, testmark: Testmark?) {
        super.checkEditorAction(before, after, actionId, trimIndent, testmark)
        resetActionManagerState()

    }
```

Looks like it fails on CI because of different test order when tests are run by Gradle

vlad20012:
:+1: 
bors r+

src/main/kotlin/org/rust/ide/commenter/RsCommenter.kt
```diff
@@ -27,8 +44,84 @@ class RsCommenter : Commenter, CodeDocumentationAwareCommenter {
...
+    override fun uncommentBlockComment(
+        startOffset: Int,
+        endOffset: Int,
+        document: Document,
+        data: CommentHolder?
+    ) =
+        SelfManagingCommenterUtil.uncommentBlockComment(startOffset, endOffset, document, blockCommentPrefix, blockCommentSuffix)
+
+    override fun isLineCommented(line: Int, offset: Int, document: Document, data: CommentHolder): Boolean {
+        return getStartLineComment(line, document, data.file)?.isEolComment ?: false
```
vlad20012, line 82: Hmm, this can be implemented with `CharArrayUtil.regionMatches` too

Kobzol: I thought that I would have to match three separate strings (`//`, `///`, `//!`), but actually it would be enough to check for `//`, right? I can change it if you want. Now I also realized that the order of `LINE_PREFIXES` is important for uncomment to work, so it maybe warrants a comment.

Issue #4738, opened 2019-12-08
IDE does not infer array size


p0lunin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.111.2147-193
* **Rust toolchain version:** rustc 1.37.0 (eae3437df 2019-08-13)
* **IDE name and version:** PyCharm 2019.3 Community Edition
* **Operating system:**

## Problem description

Editor does not highlight array size when I initialize an array with a size constant BAR, which is equal to the constant FOO + 1.

## Steps to reproduce
```rust
const FOO: usize = 1;
const BAR: usize = FOO + 1;
fn main() {
    let bar = [0; BAR];
}
```

ortem:
@p0lunin thanks for reporting. The issue is related to type inference so I've changed your title to a more specific one.

PR #5436, opened 2020-05-25
ANN: support E0435 error detection

### LINKED ISSUE #5427: Support E0435 error detection
## Environment

* **Rust toolchain version:** 1.43.1 (8d69840ab 2020-05-04) x86_64-pc-windows-msvc
* **IDE name and version:** IntelliJ IDEA 2020.1.1 Ultimate Edition (IU-201.7223.91)

## Problem description

Defining an array with a non-constant length, such as let a: [i32; n] = [0; n];,  is prohibited in Rust, but Intellij Rust does not detect the error. See [E0435](https://doc.rust-lang.org/error-index.html#E0435)
 
## Steps to reproduce
Write a simple program
```rust
fn main() {
    let n = 5;
    let a: [i32; n] = [0; n];
    println!("{}", a[0]);
}
```
IntelliJ Rust does not detect the array initialization error, so it thinks it is a valid program. However, when trying to run it, rustc says that, in let a: [i32; n], n is not allowed because it is a non-constant value.

t-kameyama:
Fixes #5427

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -79,7 +79,8 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             override fun visitExternCrateItem(o: RsExternCrateItem) = checkExternCrate(holder, o)
             override fun visitDotExpr(o: RsDotExpr) = checkDotExpr(holder, o)
             override fun visitYieldExpr(o: RsYieldExpr) = checkYieldExpr(holder, o)
-            override fun visitArrayType(o: RsArrayType) = collectDiagnostics(holder, o)
+            override fun visitArrayType(o: RsArrayType) = checkArrayType(holder, o)
+            override fun visitArrayExpr(o: RsArrayExpr) = checkArrayExpr(holder, o)
             override fun visitVariantDiscriminant(o: RsVariantDiscriminant) = collectDiagnostics(holder, o)
             override fun visitPolybound(o: RsPolybound) = checkPolybound(holder, o)
             override fun visitTraitRef(o: RsTraitRef) = checkTraitRef(holder, o)
@@ -882,6 +883,34 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         }
         // If we got here, we aren't inside a loop expr so emit an error
         RsDiagnostic.LoopOnlyKeywordUsedOutsideOfLoopError(expr.operator).addToHolder(holder)
+    }
+
+    private fun checkArrayType(holder: RsAnnotationHolder, o: RsArrayType) {
+        collectDiagnostics(holder, o)
+        val sizeExpr = o.expr
+        if (sizeExpr != null && o.arraySize == null) {
+            checkArraySizeExpr(holder, sizeExpr)
+        }
+    }
+    private fun checkArrayExpr(holder: RsAnnotationHolder, o: RsArrayExpr) {
+        val sizeExpr = o.sizeExpr
+        if (sizeExpr != null) {
+            checkArraySizeExpr(holder, sizeExpr)
+        }
+    }
+
+    private fun checkArraySizeExpr(holder: RsAnnotationHolder, sizeExpr: RsExpr) {
+        sizeExpr.descendantsOfTypeOrSelf<RsPathExpr>().forEach { pathExpr ->
+            val ref = pathExpr.path.reference?.resolve()
+            if (ref != null && ref !is RsConstant && (ref as? RsFunction)?.isConst != true) {
+                val diagnostic = if (ref is RsPatBinding) {
+                    RsDiagnostic.NonConstantValueInConstantError(pathExpr)
+                } else {
+                    RsDiagnostic.NonConstantCallInConstantError(pathExpr)
+                }
+                diagnostic.addToHolder(holder)
+            }
+        }
     }
 
     private fun isInTraitImpl(o: RsVis): Boolean {
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -1232,14 +1232,30 @@ sealed class RsDiagnostic(
             )
         }
     }
+
+    class NonConstantValueInConstantError(element: PsiElement) : RsDiagnostic(element) {
+        override fun prepare() = PreparedAnnotation(
+            ERROR,
+            E0435,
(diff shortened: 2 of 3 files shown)


vlad20012:
Unfortunately, this will produce a lot of false-positives because we can't compute an array size in complex cases. Maybe we could make is less strict for now and just check that RsBatBinding is not referred in an array size expression?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -884,6 +884,14 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         RsDiagnostic.LoopOnlyKeywordUsedOutsideOfLoopError(expr.operator).addToHolder(holder)
     }
 
+    private fun checkArrayType(holder: RsAnnotationHolder, o: RsArrayType) {
+        collectDiagnostics(holder, o)
+        val expr = o.expr
+        if (expr != null && o.arraySize == null) {
```
Undin, line 890: We can't rely on `arraySize` here. If `arraySize` is null it doesn't exactly mean that a user uses non-constant expression because,  unfortunately, our constant evaluator may fail because of other reasons.
For example, you can call const function as expression but the plugin can't calculate it for now - #4738.
Or, it may fail because of bug (see #4738).
So, we have to check it manually - walk through array size expression and check that whole expression can be evaluated in compile time. i.e. only constants, literals, const functions, etc. are used

Stzx: Yes, but also need to override `visitArrayExpr` to check array expr.

Stzx:
This situation has not been covered by the inspection：

```rust
fn main() {
    let c = 5;
    let _: [i32; c] = [0; c];
}
```

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -884,6 +885,30 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+    }
+    private fun checkArrayExpr(holder: RsAnnotationHolder, o: RsArrayExpr) {
+        val sizeExpr = o.sizeExpr
+        if (sizeExpr != null) {
+            checkArraySizeExpr(holder, sizeExpr)
+        }
+    }
+
+    private fun checkArraySizeExpr(holder: RsAnnotationHolder, sizeExpr: RsExpr) {
+        val nonConstExpr = sizeExpr.descendantsOfTypeOrSelf<RsPathExpr>().firstOrNull {
```
Undin, line 903: Unfortenately, `firstOrNull` may hide some errors from user.

For example, for the following code the compiler produces two errors
```rust
fn main() {
    let b  = 123usize;
    let c = 123;
    let a = [1; b + c];
}
```

```
error[E0435]: attempt to use a non-constant value in a constant
 --> src/main.rs:4:17
  |
4 |     let a = [1; b + c];
  |                 ^ non-constant value

error[E0435]: attempt to use a non-constant value in a constant
 --> src/main.rs:4:21
  |
4 |     let a = [1; b + c];
  |                     ^ non-constant value

error: aborting due to 2 previous errors
```

So, let's better write some recursive function that will collect all top most hon constant expressions in array-size expression and highlight all these expressions with error annotation

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -884,6 +885,29 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+        if (sizeExpr != null) {
+            checkArraySizeExpr(holder, sizeExpr)
+        }
+    }
+
+    private fun checkArraySizeExpr(holder: RsAnnotationHolder, sizeExpr: RsExpr) {
+        sizeExpr.descendantsOfTypeOrSelf<RsPathExpr>().forEach {
+            val ref = it.path.reference?.resolve()
+            if (ref != null && ref !is RsConstant && (ref as? RsFunction)?.isConst != true) {
+                RsDiagnostic.NonConstantValueInConstantError(it).addToHolder(holder)
```
Undin, line 906: BTW compiler produces [E0015](https://doc.rust-lang.org/error-index.html#E0015) when you try to use non-constant value and it's not a local variable.
Looks like it's quite easy to check it here so let's produce the same errors like compiler

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1217,14 +1217,30 @@ sealed class RsDiagnostic(
...
+
+    class NonConstantValueInConstantError(element: PsiElement) : RsDiagnostic(element) {
+        override fun prepare() = PreparedAnnotation(
+            ERROR,
+            E0435,
+            "A non-constant value was used in a constant expression"
+        )
+    }
+
+    class NonConstantFunctionCallInConstantError(element: PsiElement) : RsDiagnostic(element) {
```
Undin, line 1229:
```suggestion
    class NonConstantCallInConstantError(element: PsiElement) : RsDiagnostic(element) {
```

This diagnostic can be applied not only for function calls but also at least for method calls as well. Let's make its name more generic

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -884,6 +885,33 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+
+    private fun checkArraySizeExpr(holder: RsAnnotationHolder, sizeExpr: RsExpr) {
+        sizeExpr.descendantsOfTypeOrSelf<RsPathExpr>().forEach { pathExpr ->
+            val ref = pathExpr.path.reference?.resolve()
+            if (ref != null && ref !is RsConstant && (ref as? RsFunction)?.isConst != true) {
+                if (pathExpr.parent is RsCallExpr) {
+                    RsDiagnostic.NonConstantFunctionCallInConstantError(pathExpr).addToHolder(holder)
+                } else {
+                    RsDiagnostic.NonConstantValueInConstantError(pathExpr).addToHolder(holder)
+                }
```
Undin, line 910: Looks like it's slightly better because allow not to distinguish function calls (`RsCallExpr`) from method calls (`RsMethodCall`)
```suggestion
                val diagnostic = if (ref is RsPatBinding) {
                    RsDiagnostic.NonConstantValueInConstantError(pathExpr)    
                } else {
                    RsDiagnostic.NonConstantFunctionCallInConstantError(pathExpr)
                }
                diagnostic.addToHolder(holder)
```

PR #5439, opened 2020-05-25
CARGO: add error annotator

Kobzol:
This is a revival of PRs that add some infrastructure to the `toml` module and build several useful features on top of that - crate version checking, `Cargo.toml` error annotation, future function deprecation warning (https://github.com/intellij-rust/intellij-rust/pull/3790) and others.
The infrastructure PRs created by @mibac138 are here: https://github.com/intellij-rust/intellij-rust/pull/3785 https://github.com/intellij-rust/intellij-rust/pull/3786 https://github.com/intellij-rust/intellij-rust/pull/3787

It's a bit difficult now to add larger (crate checking) features to the `toml` module, because it lacks some utility functions (parsing crate versions and version requirements, Toml PSI elements visitor, various test base classes), so the first new feature has to bring a lot of stuff with itself. That's why the original PRs were so fragmented. I tried to merge some of them together and build a basic feature on top of them - a very basic error annotator for `Cargo.toml`. I refactored crate resolving to make it easier to mock and also to prepare it for https://github.com/intellij-rust/intellij-rust/pull/4873, so that it can be changed easily in the future if we ditch accessing `crates.io`. It's possible that proper error checking in `Cargo.toml` will be too slow until we switch to local index processing.

If there's too much in this PR, I can try to split it, but I guess that the first set of changes needs to be somewhat large to finally get the infrastructure in place so that we can build new features on top of it.

There are some questions that need to be resolved before I can continue:
1) Semver - the original PRs reimplemented both Semver parsing (`1.2.0-pre`) and Semver requirements parsing (i.e. `>=1.2`). It has been [suggested](https://github.com/intellij-rust/intellij-rust/pull/3785#issuecomment-526904978) that Semver should not be reimplemented, so I used the IntelliJ semver for now and didn't include version requirement parsing. I assume that all Cargo crates use semver (this seems like a safe assumption) and that the IntelliJ semver can parse all Cargo crate versions (not sure about that one). We still need to support crate version requirements, which is not implemented in IntelliJ (AFAIK?). Should I use the original code from the PRs and adapt it to IntelliJ semver? Or try to find some library for this?
2) Some useful infrastructure is missing:
  - I didn't find any `TomlVisitor`, which would be useful for implementing DependencyVisitor (right now its implementation is a bit too manual I think). Since `RsVisitor` is generated, is there an easy way to auto-generate the visitor from the TOML grammar?
  - For testing i.e. error annotations it's nice to have the useful `RsTest*` base classes. They are however hardcoded for Rust (i.e. `main.rs`, `Language("Rust")` etc.). I copied some of them over and transformed them for TOML (`Cargo.toml`, `Language("TOML")`). Is it OK?


Kobzol:
@Undin there are quite a few Cargo.toml improvements (https://github.com/intellij-rust/intellij-rust/issues/5440) that could be make, but it's a bit difficult to work on these without the proper infrastructure in place (TOML annotator, visitor, test base classes, etc.). This PR tries to add some of the infrastructure and it also prepares the code for future offline handling of the Cargo index (for example by parsing `git`). Can this move forward in the current state or should I rework the PR?

toml/src/main/kotlin/org/rust/toml/annotator/CargoTomlErrorAnnotator.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        // Version must be present past this point (it could've been not present because the user didn't type it yet)
+        if (version == null) {
+            return
+        }
+
+        // TODO: use version req
+        val versionValue = version.text.removeSurrounding("\"")
+        val versionInfo = crate.versions.find { it.version.rawVersion == versionValue }
```
Undin, line 40: Unfortunately, it produces false positive
For example, patch number of dependency version can be omitted.
image

Moreover, you can add some requirements (see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) to a dependency version instead of exact version.
image
image

toml/src/main/kotlin/org/rust/toml/annotator/CargoTomlErrorAnnotator.kt
```diff
@@ -0,0 +1,70 @@
...
+            override fun visitDependency(name: TomlKey, version: TomlValue?) {
+                checkDependency(name, version, holder)
+            }
+        }.visitElement(element)
+    }
+
+    private fun checkDependency(dependency: TomlKey, version: TomlValue?, holder: AnnotationHolder) {
+        val crate = getCrate(dependency)
+
+        if (crate == null) {
```
Undin, line 28: If we don't have an internet connection, we can't make request to crates.io. As a result, we get `null` here and provide false-positive annotation
image

toml/src/main/kotlin/org/rust/toml/annotator/CargoTomlErrorAnnotator.kt
```diff
@@ -0,0 +1,70 @@
...
+
+import com.intellij.ide.annotator.AnnotatorBase
+import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.psi.PsiElement
+import org.rust.toml.CargoTomlDependencyVisitor
+import org.rust.toml.crates.getCrate
+import org.toml.lang.psi.TomlKey
+import org.toml.lang.psi.TomlValue
+
+class CargoTomlErrorAnnotator : AnnotatorBase() {
```
Undin, line 16: Let's convert it into inspection to make it possible to disable it

toml/src/main/kotlin/org/rust/toml/annotator/CargoTomlErrorAnnotator.kt
```diff
@@ -0,0 +1,70 @@
...
+                    "Version has been yanked and is no longer available"
+                )
+            }
+        }
+    }
+
+    private fun AnnotationHolder.annotateError(element: PsiElement, message: String) {
+        // BACKCOMPAT: 2019.3
+        @Suppress("DEPRECATION")
+        createErrorAnnotation(
```
Undin, line 65: It seems we can't be sure that we have the latest info about crates, so I suggest to use warning level instead of error one

Undin (changes requested):
In general, this PR is not only about infrastructure, it introduces user-visible feature.
At the same time, I don't like the current implementation of dependency annotator because it relies on network requests. It's rather unstable with a poor internet connection and doesn't work at all in offline. It's reasonably for completion (although, I don't like it as well) that triggered by users, but it's not ok for annotators/inspections

I may suggest several ways what to do with these changes:
* if you need some infrastructure changes from this PR, let's extract them into separate PR (I think it's about resolver API)
* we can merge changes about dependency annotation but disable it by default and wait for more proper implementation of resolver API to enable it. Or we can wait for new resolver implementation, test how it works with it and merge after that

Kobzol:
Thank you, I now see that there are multiple issues. Based on your comments, I think that the resolver is a fundamental feature that will need to work properly before this (or most of the other issues from the linked meta issue) can be advanced. I will therefore first try to create a PR for integrating a hybrid online/offline crate resolver and then base the current PR on that.

Undin:
> I will therefore first try to create a PR for integrating a hybrid online/offline crate resolver and then base the current PR on that.

I'm not sure that you or someone else should spent time for `online` crate resolver. It doesn't look as good solution in any case. At least, I can't come up with case when request to crates.io may provide better result that local one except non-outdated data. It was just simplest way to provide completion for dependency versions but it works far from perfect (#5908)

Kobzol:
Sadly it seems that Cargo will not provide a command for crate lookup anytime soon (https://github.com/rust-lang/cargo/issues/7824), therefore I think that using `git` is the only usable way now. However, using `git` has some disadvantages, as mentioned [here](https://github.com/intellij-rust/intellij-rust/pull/4873#issuecomment-633616183).

I wanted to make an offline `git` index resolver with a fallback to the online resolver if `git` is not available. However, if the online resolver is not a good fallback, we can just disable the crate resolver service (and features that depend on it) in case that `git` is not available. What do you think?

Issue #5395, opened 2020-05-18
Allow 'Run with coverage' with specific nightly toolchain


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.43.0 (4fb7144ed 2020-04-20) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.1.1 Ultimate Edition (IU-201.7223.91)
* **Operating system:** macOS 10.13.6

## Problem description
If your current toolchain is some specific `nightly` version, the plugin still tries to launch coverage with `nightly` channel. It may lead to `error: toolchain 'nightly-x86_64-apple-darwin' is not installed` error if you don't have default `nightly` toolchain

## Steps to reproduce
* `rustup uninstall nightly`
* `rustup default nightly-2020-05-15`
* run any run configuration with coverage

PR #5441, opened 2020-05-25
RUN: Fix environment variables for Coverage runner

mchernyavsky:
Fixes `Run with Coverage` on the latest versions of the plugin.

Also adds tests for the Coverage runner.

NB: For now, `cargo test` with coverage doesn't work on Windows if MSVC toolchain is used (https://github.com/mozilla/grcov/issues/462)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,11 @@ before_script:
   - dig +short myip.opendns.com @resolver1.opendns.com || true
   - curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain $RUST_VERSION -y
   - export PATH=$HOME/.cargo/bin:$PATH
+  - rustup toolchain add nightly
   - rustup component add rust-src
   - rustup component add clippy
   - rustup component add rustfmt
+  - cargo +nightly install grcov
   - if [ ! $RUST_VERSION \< "1.40.0" ]; then cargo install evcxr_repl; fi # BACKCOMPAT: Rust 1.39.0
   - ln -s $(rustc --print sysroot)/lib/rustlib/src/rust/src $RUST_SRC_WITH_SYMLINK
   - ./check-license.sh
--- a/coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
+++ b/coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
@@ -16,7 +16,6 @@ import com.intellij.execution.process.OSProcessHandler
 import com.intellij.execution.process.ProcessAdapter
 import com.intellij.execution.process.ProcessEvent
 import com.intellij.execution.runners.ExecutionEnvironment
-import com.intellij.execution.runners.ProgramRunner
 import com.intellij.execution.ui.RunContentDescriptor
 import com.intellij.openapi.application.WriteAction
 import com.intellij.openapi.diagnostic.Logger
@@ -52,16 +51,13 @@ class GrcovRunner : RsDefaultProgramRunnerBase() {
     }
 
     override fun execute(environment: ExecutionEnvironment) {
-        val state = environment.state as CargoRunStateBase
-
         if (checkNeedInstallGrcov(environment.project)) return
 
-        state.addCommandLinePatch(cargoCoveragePatch)
-        environment.cargoPatches.add(cargoCoveragePatch)
-
+        val state = environment.state as CargoRunStateBase
         val workingDirectory = state.commandLine.workingDirectory.toFile()
         cleanOldCoverageData(workingDirectory)
 
+        environment.cargoPatches += cargoCoveragePatch
         super.execute(environment)
     }
 
--- a/coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
+++ b/coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
@@ -1,0 +1,238 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.coverage
+
+import com.intellij.coverage.CoverageDataManager
+import com.intellij.coverage.CoverageExecutor
+import com.intellij.coverage.CoverageRunnerData
+import com.intellij.execution.ExecutorRegistry
+import com.intellij.execution.configurations.RunConfiguration
+import com.intellij.execution.runners.ExecutionEnvironmentBuilder
+import com.intellij.execution.runners.ProgramRunner
+import com.intellij.execution.ui.RunContentDescriptor
+import com.intellij.openapi.util.SystemInfo
+import com.intellij.openapi.vfs.VfsUtil
+import com.intellij.rt.coverage.data.ClassData
+import com.intellij.rt.coverage.data.LineData
+import com.intellij.rt.coverage.data.ProjectData
+import org.rust.FileTreeBuilder
+import org.rust.lang.core.psi.isRustFile
+import org.rust.openapiext.toPsiFile
+import org.rustSlowTests.cargo.runconfig.RunConfigurationTestBase
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+
+class RsCoverageTest : RunConfigurationTestBase() {
+    private val coverageData: ProjectData?
+        get() = CoverageDataManager.getInstance(project).currentSuitesBundle?.coverageData
+
+    fun `test main`() = doTest {
+        toml("Cargo.toml", """
+            [package]
+            name = "hello"
+            version = "0.1.0"
+            authors = []
+        """)
+
+        dir("src") {
+            rust("main.rs", """
+                fn main() {                     // Hits: 1
+                    println!("Hello, world!");  // Hits: 1
+                }                               // Hits: 1
+            """)
+        }
+    }
+
+    fun `test function call`() = doTest {
+        toml("Cargo.toml", """
+            [package]
+            name = "hello"
+            version = "0.1.0"
+            authors = []
+        """)
+
+        dir("src") {
+            rust("main.rs", """
+                fn foo() {                      // Hits: 1
+                    println!("Hello, world!");  // Hits: 1
+                }                               // Hits: 1
+
+                fn main() {                     // Hits: 1
+                    foo();                      // Hits: 1
+                }                               // Hits: 1
+            """)
+        }
+    }
+
+    fun `test function call from lib`() = doTest {
+        toml("Cargo.toml", """
+            [package]
+            name = "hello"
+            version = "0.1.0"
+            authors = []
+            edition = "2018"
+        """)
+
+        dir("src") {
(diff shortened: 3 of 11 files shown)


coverage/src/main/kotlin/org/rust/coverage/RsCoverageRunner.kt
```diff
@@ -46,6 +48,10 @@ class RsCoverageRunner : CoverageRunner() {
     companion object {
         private val LOG: Logger = Logger.getInstance(RsCoverageRunner::class.java)
 
+        @TestOnly
+        @Volatile
+        var lastProjectData: ProjectData? = null
```
Undin, line 53: Looks like you don't need it at all and you may use `CoverageDataManager.getInstance(project).currentSuitesBundle.coverageData` instead

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+class RsCoverageTest : RunConfigurationTestBase() {
+
+    override fun setUp() {
+        super.setUp()
+        lastProjectData = null
+    }
+
+    override fun tearDown() {
+        super.tearDown()
+        lastProjectData = null
```
Undin, line 30: In this case, it shouldn't affect anything if everything is ok, but you should clear your state before `super.tearDown()`

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+    fun `test for function call`() = doTest {
+        toml("Cargo.toml", """
+            [package]
+            name = "hello"
+            version = "0.1.0"
+            authors = []
+        """)
+
+        dir("src") {
+            rust("lib.rs", """
```
Undin, line 125: Do you mean `main.rs` here? Otherwise, it doesn't work

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+                #[test]
+                fn test2() {                    // Hits: 2
+                    foo();                      // Hits: 1
+                }                               // Hits: 2
+            """)
+        }
+    }
+
+    private fun executeGrcov(configuration: RunConfiguration) {
+        val executor = DefaultRunExecutor.getRunExecutorInstance()
```
Undin, line 167: Why we use default executor? I think we should use `CoverageExecutor`

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+        val environment = ExecutionEnvironmentBuilder.create(executor, configuration).build()
+        val state = environment.state as CargoRunStateBase
+        GrcovRunner().execute(environment, null, state)
```
Undin, line 170: Hm, looks like environment already should contain correct runner (of course, if you use correct executor)

Undin, line 170: And why we use different ways to launched run configuration here and in `org.rustSlowTests.cargo.runconfig.RunConfigurationTestBase#execute`?

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+                Pair(lineNumber, hits)
+            }
+
+    private fun hitsFrom(data: ClassData): List<Pair<Int, Int>> =
+        data.lines
+            .filterIsInstance<LineData>()
+            .map { Pair(it.lineNumber, it.hits) }
+
+    private companion object {
+        const val MARKER: String = "// Hits:"
```
Undin, line 220: private?

coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
```diff
@@ -111,7 +111,7 @@ class GrcovRunner : RsDefaultProgramRunnerBase() {
         private fun startCollectingCoverage(workingDirectory: File, environment: ExecutionEnvironment) {
             val project = environment.project
             val runConfiguration = environment.runProfile as? RunConfigurationBase<*> ?: return
-            val runnerSettings = environment.runnerSettings ?: return
+            val runnerSettings = environment.runnerSettings ?: CoverageRunnerData()
```
Undin, line 114: Can we pass this data into execution environment in tests directly instead of this hack here?

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+            true
+        }
+
+        val configuration = createConfiguration()
+        if (runTests) {
+            configuration.command = "test"
+        }
+
+        executeGrcov(configuration)
+        runWithInvocationEventsDispatching("Failed to fetch coverage data") { lastProjectData != null }
```
Undin, line 190: JFYI: there is `CoverageSuiteListener` and probably we can use `com.intellij.coverage.CoverageSuiteListener#afterSuiteChosen` to be notified that running is finished

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+                val startIndex = it.value.indexOf(MARKER) + MARKER.length
+                val rawHits = it.value.substring(startIndex).trim()
+                val hits = Integer.parseInt(rawHits)
+                Pair(lineNumber, hits)
+            }
+
+    private fun hitsFrom(data: ClassData): List<Pair<Int, Int>> =
+        data.lines
+            .filterIsInstance<LineData>()
+            .map { Pair(it.lineNumber, it.hits) }
```
Undin, line 217: nit: 
```suggestion
            .map { it.lineNumber to it.hits }
```

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+    private fun hitsFrom(text: String): List<Pair<Int, Int>> =
+        text.split('\n')
+            .withIndex()
+            .filter { it.value.contains(MARKER) }
+            .map {
+                val lineNumber = it.index + 1
+                val startIndex = it.value.indexOf(MARKER) + MARKER.length
+                val rawHits = it.value.substring(startIndex).trim()
+                val hits = Integer.parseInt(rawHits)
+                Pair(lineNumber, hits)
```
Undin, line 211: nit:
```suggestion
                lineNumber to hits
```

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
```diff
@@ -0,0 +1,222 @@
...
+        runWithInvocationEventsDispatching("Failed to fetch coverage data") { lastProjectData != null }
+        val projectData = lastProjectData!!
+
+        val actual = hashMapOf<String, List<Pair<Int, Int>>>()
+        for (path in expected.keys) {
+            val data = projectData.getClassData(path)!!
+            actual[path] = hitsFrom(data)
+        }
+
+        assertEquals(expected, actual)
```
Undin, line 199: Let's come up with more readable structure for comparison because currently if actual hits count differs from expected, you get something like
```
expected:<{/private/var/folders/hj/rsx5n85s56g3_mj0fgznct500000gp/T/unitTest__tests/unitTest7730444972665203395/src/lib.rs=[(1, 2), (2, 2), (3, 2), (6, 1), (7, 1), (8, 2), (11, 2), (12, 1), (13, 2)]}> but was:<{/private/var/folders/hj/rsx5n85s56g3_mj0fgznct500000gp/T/unitTest__tests/unitTest7730444972665203395/src/lib.rs=[(1, 2), (2, 2), (3, 2), (6, 2), (7, 1), (8, 2), (11, 2), (12, 1), (13, 2)]}>
```

And it's rather hard to understand what's wrong here

src/test/kotlin/org/rustSlowTests/cargo/runconfig/RunConfigurationTestBase.kt
```diff
@@ -71,13 +67,16 @@ abstract class RunConfigurationTestBase : RsWithToolchainTestBase() {
             ?: error("Can't create run configuration settings")
     }
 
-    protected fun execute(configuration: RunConfiguration): ExecutionResult {
+    protected fun execute(
+        configuration: RunConfiguration,
+        runner: ProgramRunner<*> = CargoCommandRunner()
```
Undin, line 72: Looks like we should pass runner manually
It should be chosen by execution environment

.travis.yml
```diff
@@ -39,6 +39,7 @@ before_script:
   - rustup component add rust-src
   - rustup component add clippy
   - rustup component add rustfmt
+  - cargo install grcov
```
Undin, line 42: We also need to install it by `.github/workflows/check.yml` in `Set up Rust` step to run these tests on github actions as well

Undin: And looks like we also need to install nightly toolchain because we always use `+nightly` channel for coverage
But I'd prefer to teach our runner to use default nightly toolchain if possible - #5395

coverage/src/test/kotlin/org/rust/coverage/RsCoverageTest.kt
mchernyavsky: I use `runWithInvocationEventsDispatching` because it calls `UIUtil.dispatchAllInvocationEvents` periodically.

coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
```diff
@@ -52,16 +51,13 @@ class GrcovRunner : RsDefaultProgramRunnerBase() {
...
         if (checkNeedInstallGrcov(environment.project)) return
 
-        state.addCommandLinePatch(cargoCoveragePatch)
-        environment.cargoPatches.add(cargoCoveragePatch)
-
+        val state = environment.state as CargoRunStateBase
         val workingDirectory = state.commandLine.workingDirectory.toFile()
         cleanOldCoverageData(workingDirectory)
 
+        environment.cargoPatches.add(cargoCoveragePatch)
```
Undin, line 60: It's quite hard to understand from what threads we modify `cargoPatches` list. And even if we modify them from EDT only currently, nothing stops us to start modify them from another thread in future that may lead to unexpected errors.
I'm not sure that even in this case it will lead to errors but I don't want to care about it a all.
So let's convert `MutableList<CargoPatch>` into some immutable structure and create a new object instead of modifying the old one. Note, user data is already thread-safe because it uses immutable map under the hood

Issue #4514, opened 2019-10-10
Improve backtrace filters


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.107.2757-193-nightly
* **Rust toolchain version:** 1.38.0 (625451e37 2019-09-23) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2019.3 EAP Ultimate Edition (IU-193.4099.13)
* **Operating system:** macOS 10.13.6

## Problem description
Quite a lot of lines of Rust backtrace don't have a link to sources although a user has them.
To improve that we should support the following cases:
* [ ] better process fully qualified names to stdlib items (red cases)
* [ ] understand relative paths to stdlib sources (blue cases)
* [ ] understand paths to sdtlib sources start with `/rustc/commit-hash` prefix (green cases)
* [ ] parse `<structPath as traitPath>::method` (violet case). It also may be useful in profiler
* [ ] understand extern function calls that don't have module prefix (oragne case)

<img width="930" alt="Screen Shot 2019-10-10 at 09 57 46" src="https://user-images.githubusercontent.com/2539310/66547599-420f9180-eb48-11e9-9785-1655759412e7.png">

## Steps to reproduce
To get similar backtrace on your machine just run the test from the following code
```rust
#[test]
fn foo() {
   assert_eq!(1, 2);
}
```
```
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:339
[...]

PR #5444, opened 2020-05-25
RUN: resolve stdlib and Cargo source code hyperlinks in backtrace

Kobzol:
This PR adds support for resolving source code links to stdlib and crate files in backtraces. Is there a way to test this properly? The Rust toolchain is `null` in tests and the test file system doesn't see absolute paths from the user's `.cargo` and `.rustup` folders.

Fixes part of https://github.com/intellij-rust/intellij-rust/issues/4514

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
@@ -6,14 +6,15 @@
 package org.rust.cargo.runconfig.filters
 
 import com.intellij.execution.filters.Filter
-import com.intellij.execution.filters.HyperlinkInfo
 import com.intellij.execution.filters.OpenFileHyperlinkInfo
 import com.intellij.openapi.project.DumbAware
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.io.FileUtil
 import com.intellij.openapi.vfs.VirtualFile
 import org.intellij.lang.annotations.Language
-import org.rust.openapiext.findFileByMaybeRelativePath
-import java.io.File
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.cargo.project.settings.rustSettings
+import java.nio.file.Paths
 
 /**
  * Base class for regexp-based output filters that extract
@@ -32,6 +33,9 @@ open class RegexpFileLinkFilter(
         // TODO: named groups when Kotlin supports them
         @Language("RegExp")
         val FILE_POSITION_RE = """((?:\p{Alpha}:)?[0-9 a-z_A-Z\-\\./]+):([0-9]+)(?::([0-9]+))?"""
+
+        @Language("RegExp")
+        private val RUSTC_ABSOLUTE_PATH_RE = Regex("""/rustc/\w+/(.*)""")
     }
 
     init {
@@ -49,10 +53,24 @@ open class RegexpFileLinkFilter(
         val columnNumber = match.groups[3]?.let { zeroBasedNumber(it.value) } ?: 0
 
         val lineStart = entireLength - line.length
+
+        val file = resolveFilePath(fileGroup.value)
+        val link = file?.let { OpenFileHyperlinkInfo(project, file.file, lineNumber, columnNumber) }
+
+        val grayedOut = if (file == null) {
+            false
+        } else {
+            file !is ResolvedPath.Workspace
+        }
+
+        val end = match.groups[3]?.range?.last
+            ?: match.groups[2]?.range?.last
+            ?: fileGroup.range.last
         return Filter.Result(
-            lineStart + fileGroup.range.start,
-            lineStart + fileGroup.range.endInclusive + 1,
-            createOpenFileHyperlink(fileGroup.value, lineNumber, columnNumber)
+            lineStart + fileGroup.range.first,
+            lineStart + end + 1,
+            link,
+            grayedOut
         )
     }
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -65,9 +77,64 @@ open class RegexpFileLinkFilter(
         }
     }
 
-    private fun createOpenFileHyperlink(fileName: String, line: Int, column: Int): HyperlinkInfo? {
-        val platformNeutralName = fileName.replace(File.separatorChar, '/')
-        val file = cargoProjectDirectory.findFileByMaybeRelativePath(platformNeutralName) ?: return null
-        return OpenFileHyperlinkInfo(project, file, line, column)
+    private fun resolveFilePath(fileName: String): ResolvedPath? {
+        val path = fileName.replace(File.separatorChar, '/')
```
Undin, line 81: `FileUtil.toSystemIndependentName`

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -65,9 +77,64 @@ open class RegexpFileLinkFilter(
...
+        if (file != null) return ResolvedPath.Project(file)
+
+        val stdlibFile = resolveStdlibFile(fileName)
+        if (stdlibFile != null) return stdlibFile
+
+        val cargoFile = resolveCargoFile(fileName)
+        if (cargoFile != null) return cargoFile
+
+        // try to resolve absolute path
+        return cargoProjectDirectory.fileSystem.findFileByPath(path)?.let { ResolvedPath.Crate(it) }
```
Undin, line 92: Maybe it's better to introduce something like `Common` or `Unknown` because don't know where the path lead (at least currently)

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -65,9 +77,64 @@ open class RegexpFileLinkFilter(
...
+    private fun resolveFilePath(fileName: String): ResolvedPath? {
+        val path = fileName.replace(File.separatorChar, '/')
+        val file = cargoProjectDirectory.findFileByRelativePath(path)
+        if (file != null) return ResolvedPath.Project(file)
+
+        val stdlibFile = resolveStdlibFile(fileName)
+        if (stdlibFile != null) return stdlibFile
+
+        val cargoFile = resolveCargoFile(fileName)
+        if (cargoFile != null) return cargoFile
```
Undin, line 89: nit:
```suggestion
        val path = resolveStdlibFile(fileName) ?: resolveCargoFile(fileName)
        if (path != null) return path
```

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -65,9 +77,64 @@ open class RegexpFileLinkFilter(
...
+
+        // try to resolve absolute path
+        return cargoProjectDirectory.fileSystem.findFileByPath(path)?.let { ResolvedPath.Crate(it) }
+    }
+
+    private fun resolveCargoFile(path: String): ResolvedPath? {
```
Undin, line 95:
```suggestion
    private fun resolveCargoPath(path: String): ResolvedPath? {
```

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -65,9 +77,64 @@ open class RegexpFileLinkFilter(
...
+        return cargoProjectDirectory.fileSystem.findFileByPath(fullPath)?.let { ResolvedPath.Crate(it) }
+    }
+
+    // /cargo/registry/src/github.com-1ecc6299db9ec823/dependency/... -> dependency/...
+    private fun normalizeCargoPath(path: String): String {
+        val match = CARGO_ABSOLUTE_PATH_RE.matchEntire(path) ?: return path
+        return match.groupValues[1]
+    }
+
+    private fun resolveStdlibFile(path: String): ResolvedPath? {
```
Undin, line 107:
```suggestion
    private fun resolveStdlibPath(path: String): ResolvedPath? {
```

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -65,9 +77,64 @@ open class RegexpFileLinkFilter(
...
+    private fun cargoHome(): String {
+        return System.getenv("CARGO_HOME") ?: "~/.cargo"
+    }
+
+    private fun cargoSourceIndex(): String {
+        return FileUtil.expandUserHome("${cargoHome()}/registry/src/github.com-1ecc6299db9ec823")
+    }
+
+    sealed class ResolvedPath(val file: VirtualFile) {
+        open val ownedByUser: Boolean = false
```
Undin, line 131: It seems yoo don't need this property. Just check that path is `Project`

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -32,6 +33,12 @@ open class RegexpFileLinkFilter(
         // TODO: named groups when Kotlin supports them
         @Language("RegExp")
         val FILE_POSITION_RE = """((?:\p{Alpha}:)?[0-9 a-z_A-Z\-\\./]+):([0-9]+)(?::([0-9]+))?"""
+
+        @Language("RegExp")
+        private val RUSTC_ABSOLUTE_PATH_RE = Regex("""/rustc/\w+/(.*)""")
+
+        @Language("RegExp")
+        private val CARGO_ABSOLUTE_PATH_RE = Regex("""/cargo/registry/src/github.com-1ecc6299db9ec823/(.*)""")
```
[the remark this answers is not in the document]
Kobzol: Good point, thanks.

src/main/kotlin/org/rust/cargo/runconfig/filters/RegexpFileLinkFilter.kt
```diff
@@ -50,10 +57,15 @@ open class RegexpFileLinkFilter(
...
+
+        val file = resolveFilePath(fileGroup.value)
+        val link = file?.let { OpenFileHyperlinkInfo(project, file.file, lineNumber, columnNumber) }
+
         return Filter.Result(
             lineStart + fileGroup.range.start,
             lineStart + fileGroup.range.endInclusive + 1,
-            createOpenFileHyperlink(fileGroup.value, lineNumber, columnNumber)
+            link,
+            file != null && !file.ownedByUser
```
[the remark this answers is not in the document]
Kobzol: You're right, although I'm not sure how to resolve it. If I force gray when `file` is null, it does this:
image

Kobzol:
I tried to write a test, however it fails to finish:
```
[1]: Virtual pointer 'temp:///src/src' hasn't been disposed: --------------Creation trace: 
java.lang.Throwable
	at com.intellij.openapi.util.TraceableDisposable.<init>(TraceableDisposable.java:31)
	at com.intellij.openapi.vfs.impl.VirtualFilePointerImpl.<init>(VirtualFilePointerImpl.java:40)
	at com.intellij.openapi.vfs.impl.VirtualFilePointerManagerImpl.getOrCreate(VirtualFilePointerManagerImpl.java:340)
```

Undin:
> I tried to write a test, however it fails to finish:

@Kobzol it should be fixed by #5950
Moreover, after it you won't have to set up toolchain settings manually, it will be set up by `WithStdlibRustProjectDescriptor` used in the corresponding tests

Kobzol:
Thanks! I rebased and now it works. I don't know how to test cargo paths though. I would have to use some cargo dependency that is 100 % available in the user's cargo root for the filter to find it. Or install the dependency during the test?

Undin:
> I don't know how to test cargo paths though. I would have to use some cargo dependency that is 100 % available in the user's cargo root for the filter to find it. Or install the dependency during the test?

@Kobzol Probably, the simplest way to do it is to use `org.rust.cargo.RsWithToolchainTestBase` as base class for tests. These tests use real filesystem for test projects and launch real `Cargo` so you can specify any dependency in `Cargo.toml` and download it, run compiler or cargo, or any other tool. Of course, these tests are slower than inheritors of `org.rust.RsTestBase`

Kobzol:
Ok, I added a separate class to not slow down the rest of the tests.

Undin (commented):
Let's also highlight line number as we already do for workspace files
image

src/test/kotlin/org/rust/cargo/runconfig/filters/RsBacktraceFilterCargoTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    fun `test resolve cargo crate`() {
+        fileTree {
+            toml("Cargo.toml", """
+                [package]
+                name = "hello"
+                version = "0.1.0"
+                authors = []
+
+                [dependencies]
+                rand = "0.7.3"
```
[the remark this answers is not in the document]
Kobzol: Ok, I switched to `left-pad`. What can go wrong... :D

Kobzol:
I think there was a bug before, because the highlight range was spanning the whole line sometimes. That's why also the line number wasn't highlighted in tests. Now line/column numbers are highlighted explicitly if they are present.

Kobzol:
Hmm, even with a crate that has one source file and no dependencies, it still times out. Maybe it takes too long to download the crates.io index? But I would expect that this is not the only test that needs it.

Undin:
@Kobzol Let's try to increase timeout in `org.rust.cargo.project.model.impl.TestCargoProjectsServiceImpl#discoverAndRefreshSync`

Undin:
@Kobzol I've investigated the problem with timeout. Found out that it's because of `java.lang.AssertionError: File accessed outside allowed roots: file:///usr/share/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/left-pad-1.0.1/src/lib.rs;` somewhere in `org.rust.cargo.project.model.impl.CargoProjectImpl#doRefresh`. Should be fixed by #6133

Kobzol:
It seems that the error is still there:
```
com.intellij.testFramework.LoggedErrorProcessor$TestLoggerAssertionError: File accessed outside allowed roots: file:///usr/share/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/left-pad-1.0.1/src/lib.rs;
```

Undin (commented):
> It seems that the error is still there:

Should be fixed in #6159

Kobzol:
Great, the tests finally passed :) Although there seem to be some false positives in analysis regression tests.

Undin:
>  Although there seem to be some false positives in analysis regression tests.

It's a flaky annotation, unfortunately. It's not related to your changes

Undin:
bors try

Undin (approved):
And finally, all tests pass 🎉 
Thank you!
bors r+

PR #5448, opened 2020-05-26
ANN: don't annotate format-like non-stdlib marcos

### LINKED ISSUE #5425: Can't disable plugin annotations for built-in macros
I want to disable all annotations like "Argument never used" for `print`, `println` and other built-in macros which were added in some plugin version but there's no such option in Editor->Inspections->Rust, the Registry or Experimental Features menus.

How do I do it?

Undin:
Fixes #5425

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotator.kt
@@ -12,16 +12,14 @@ import com.intellij.openapi.util.TextRange
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
 import org.intellij.lang.annotations.Language
+import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.colors.RsColor
 import org.rust.ide.injected.isDoctestInjection
 import org.rust.ide.presentation.render
 import org.rust.lang.core.macros.MacroExpansionMode
 import org.rust.lang.core.macros.macroExpansionManager
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.descendantOfTypeStrict
-import org.rust.lang.core.psi.ext.macroName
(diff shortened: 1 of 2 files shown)


Kobzol:
I thought that users might complain that the annotator applies to too few functions, I wouldn't expect that it would apply to too many of them :sweat_smile: Ultimately, it would be most precise if we could detect that `format_args!` has been used in the macro after macro expansion. LGTM.

MSxDOS:
@Undin I still get the annotations unless the syntax is different from the standard one e.g. 
```Rust
use my_macros::println;
println!("literal", expr);
```
still triggers "Argument never used" but
```Rust
use my_macros::println;
println!(expr, "literal", expr);
```
does not.

Also, going to implementations (Ctrl + Click) forwards to a wrong definition from stdlib, liballoc or another place depending on the macro, so seemingly the plugin makes assumptions with those instead of resolving their definitions properly.

Undin:
@MSxDOS Could you please provide an example where we can reproduce it? It would be very helpful

MSxDOS:
@Undin I think it happens when a macro is imported with a different name via `use`, its origin doesn't matter. Here's a minimal example:
```Rust
#![no_std]

macro_rules! debug {
    ($($anything:tt)*) => {};
}

use debug as println;

fn test() {
    println!("lit", 1); // 'Argument never used'
}
```

vlad20012:
In this case, our poor name resolution just wrongly resolves `println`.
@Undin I think we still should provide registry key to disable it...

Undin:
@vlad20012 I don't like registry options for such purposes because if user turn on such option, it's rather hard to remember that we have such registry key and even harder to found out that something doesn't work because of this option. Moreover, the question: should we introduce such keys for each annotator or for each error type?

I have two other variants:
* fix name resolution for macro aliases (I already have a prototype)
* move error detection into inspection. It will lead to double parsing (one while highlighting, another in inspection) of format string in macro but probably it's ok

What do you think?

Issue #4006, opened 2019-06-11
"Add import for" intention


vlad20012:
```rust
fn main() {
    std::sync::Arc::new(1);
}
```
=>
```rust
use std::sync::Arc;

fn main() {
    Arc::new(1);
}
```

Kotlin's analogue:
![Peek 2019-06-11 11-51](https://user-images.githubusercontent.com/3221931/59257894-8b55e000-8c3f-11e9-91d8-91c8f7ed82d6.gif)

See [`ImportCandidate.import()`](https://github.com/intellij-rust/intellij-rust/blob/90c06317b8956a51a42f8eee81af2bd46b0053a6/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt#L460).
Not that import should not be added if the name is already in the scope.

PR #5451, opened 2020-05-26
INT: add Import for intention

Kobzol:
This PR adds intention that imports a part of a complex path.

import

I'm not very sure about the used import machinery, is it correct? I also looked at `ImportContext`, but this has worked so far. Especially joining parts of the path to put into the use statement seems a bit nasty.

A reverse intention (fully qualify a path) could probably also be useful.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4006


Kobzol:
import

Issue #3531, opened 2019-03-14
Move refactoring


fan-tom:
Does anyone work on "Move ..." refactoring to easily move types/functions/impls from one module to another with automatic imports fixing? Current implementation suggest to move the entire file

vlad20012:
> Does anyone work on

Currently, no

dima74:
Current plan:
* [ ] Move single file (#5181)
* [ ] Move directory / Move file with its child modules
* [ ] Move top level items

dima74:
Move refactoring is mostly completed, so I invite everyone to try it and share feedback! All code is in [`diralik/move`](https://github.com/intellij-rust/intellij-rust/tree/diralik/move) branch, so you can just run IDE from this branch, or, alternatively, use these plugin builds:
* 2020.1: [intellij-rust-0.2.125.SNAPSHOT-201-dev.zip](https://github.com/intellij-rust/intellij-rust/files/4769551/intellij-rust-0.2.125.SNAPSHOT-201-dev.zip)
* 2020.2: [intellij-rust-0.2.125.SNAPSHOT-202-dev.zip](https://github.com/intellij-rust/intellij-rust/files/4769541/intellij-rust-0.2.125.SNAPSHOT-202-dev.zip)

---

There are two types of move refactoring: 
1. Move top level items. To invoke it place cursor at some item in editor (or select multiple items), press F6 and then optionally select additional items to move
![image](https://user-images.githubusercontent.com/6505554/84427701-c28a1300-ac3e-11ea-9934-94e349eb6472.png)

2. Move files or directories. It can be invoked directly from editor (F6 when cursor is at empty line) or from file tree (F6 or using cut&paste or using drag&drop):
![image](https://user-images.githubusercontent.com/6505554/84428547-1517ff00-ac40-11ea-829c-50902591f7e7.png)

---

What is expected to work now:
* Moving multiple files (with child modules) / multiple directories
* Moving to different crate
* Checking visibility conflicts (use of private item/field/method after move)
* Changing references to moved elements
  - including references in imports, references though reexports, self references, etc
* Changing relative references from moved elements
* Adding import for traits which methods are used in moved items

---

What is not expected to work now:
* 2015 edition
* References inside macro calls / macro_rules
* Moving macros / updating references to macros
* Anything related to `cfg` attributes
* Changing paths in attributes (https://github.com/intellij-rust/intellij-rust/issues/5446#issuecomment-634477944)
* References from comments/text
[...]

PR #5455, opened 2020-05-26
REF: Basic version of move top-level items refactoring

dima74:
Initial version of move refactoring for top-level items (#3531). Refactoring is activated by pressing F6, then dialog is shown where user selects items to move:

image

Current implementation just moves items, actual work (e.g. updating references) will be added in following pull requests

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsDialog.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsDialog.kt
@@ -1,0 +1,107 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.refactoring.move
+
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.ui.TextFieldWithBrowseButton
+import com.intellij.openapi.vfs.LocalFileSystem
+import com.intellij.refactoring.RefactoringBundle.message
+import com.intellij.refactoring.ui.RefactoringDialog
+import com.intellij.refactoring.util.CommonRefactoringUtil
+import com.intellij.ui.components.JBTextField
+import com.intellij.ui.layout.panel
+import com.intellij.util.IncorrectOperationException
+import org.rust.lang.core.psi.ext.RsItemElement
+import org.rust.lang.core.psi.ext.RsMod
+import org.rust.openapiext.pathToRsFileTextField
+import org.rust.openapiext.toPsiFile
+import org.rust.stdext.mapToSet
+import java.awt.Dimension
+import java.io.File
+import javax.swing.JComponent
+
+class RsMoveTopLevelItemsDialog(
+    project: Project,
+    private val itemsToMove: Set<RsItemElement>,
+    private val sourceMod: RsMod
+) : RefactoringDialog(project, false) {
+
+    private val sourceFilePath: String = sourceMod.containingFile.virtualFile.path
+    private val sourceFileField: JBTextField = JBTextField(sourceFilePath).apply { isEnabled = false }
+    private val targetFileChooser: TextFieldWithBrowseButton = createTargetFileChooser(project)
+    private val memberPanel: RsMemberSelectionPanel = createMemberInfoPanel()
+
+    private var searchForReferences: Boolean = true
+
+    init {
+        super.init()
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+        val elements = arrayOf(element)
+        if (editor?.selectionModel?.hasSelection() != true && !canMove(elements, null, reference)) return false
+
+        doMove(project, elements)
+        return true
+    }
+
+    private fun doMove(project: Project, elements: Array<out PsiElement>) {
+        val containingMod = PsiTreeUtil.findCommonParent(*elements)?.parentOfType<RsMod>() ?: return
+        val itemsToMove = elements.filterIsInstance<RsItemElement>().toList()
```
dima74, line 66: This is somewhat basic approach to choose initial items. I already implemented approach with better handling of selection, that is if user selects some items in editor, then these items will be selected in dialog (will open another PR for it)

Kobzol:
This is great! I have wanted this refactoring for a long time, it is tedious to manually move a lot of things. I would like to propose one improvement: structs/enums often have a lot of impls (inherent impl(s) and impls for various traits). Often when I move a struct, I want to move all of its impls. In the current GUI dialog it might be a bit slow to find and select all of the corresponding impls. Could there either be a:
1) button that selects all impls for a selected struct/enum
2) hierarchival view (impls for a struct would be located in a tree hierarchy under the struct)

What do you think?

src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsDialog.kt
```diff
@@ -0,0 +1,127 @@
...
+        chooser.textField.caretPosition = sourceFilePath.removeSuffix(".rs").length
+        chooser.textField.moveCaretPosition(sourceFilePath.lastIndexOf('/') + 1)
+
+        return chooser
+    }
+
+    override fun createCenterPanel(): JComponent {
+        val panel = panel {
+            row("From:") {
+                sourceFileField(growX).withLargeLeftGap()
```
Undin, line 63: `withLargeLeftGap` is available only since 2020.1
You can wait a little bit - we are going to drop 193 support very soon.
Or you can try to use our conditional compilation ["magic"](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#platform-versions) and create separate methods for each platform.
I prefer the first variant

dima74:
@Kobzol thanks for the suggestion! I think hierarchical view is better, I will try to implement it in future PRs

src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsDialog.kt
dima74: Yes, thank you, I will wait

Undin: #5479

src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsHandler.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    override fun tryToMove(
+        element: PsiElement,
+        project: Project,
+        dataContext: DataContext?,
+        reference: PsiReference?,
+        editor: Editor?
+    ): Boolean {
+        val elements = arrayOf(element)
+        if (editor?.selectionModel?.hasSelection() != true && !canMove(elements, null, reference)) return false
```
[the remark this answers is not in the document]
dima74: I think we want the following behaviour:
* if something is selected, then most likely user wants to move selected items
* if cursor is at empty line, then we assume that user wants to move entire file
* otherwise we assume that user wants to move item under cursor

So I think `&&` was correct. And thanks for the suggestion to rewrite using `if (everythingOK)`, it is clearer now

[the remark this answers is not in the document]
ortem: OK, thank you for the explanation!

dima74:
bors r=ortem

Issue #5452, opened 2020-05-26
Change return type fix doesn't work with two different types with the same name


Kobzol:
If the `Change return type fix` is invoked and the new return type cannot be imported because of name shadowing, the fix doesn't do anything.

```rust
mod a {
    pub struct Error;
}

pub struct Error;

fn test() -> Error {
    /*caret*/a::Error /* nothing changes */
}
```
It should change the return type to `a::Error`.

Related to https://github.com/intellij-rust/intellij-rust/issues/4979 and https://github.com/intellij-rust/intellij-rust/issues/5400, they are caused by the usage of `ImportHelper`. For this to work, it would have to replace types that cannot be imported by fully qualified paths.

PR #5456, opened 2020-05-26
FIX: render types with a fully qualified name when needed

Kobzol:
This PR enhances `TypeRenderer` to render (fully) qualified names for types when needed. This is basically only needed when two types are displayed together to disambiguate them if their name is the same, but they are different types. Two instances of this scenario that I found is in the type mismatch diagnostic and in `ChangeReturnTypeFix`, so I changed both of them to use the qualified name type renderer in this PR.

Fixes https://github.com/intellij-rust/intellij-rust/issues/3223, fixes https://github.com/intellij-rust/intellij-rust/issues/5452.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeReturnTypeFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeReturnTypeFix.kt
@@ -14,11 +14,14 @@ import com.intellij.psi.util.PsiTreeUtil
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.presentation.render
 import org.rust.ide.presentation.renderInsertionSafe
+import org.rust.ide.utils.import.RsImportHelper.getTypeReferencesInfoFromTys
 import org.rust.ide.utils.import.RsImportHelper.importTypeReferencesFromTy
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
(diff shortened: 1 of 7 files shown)


mchernyavsky (commented):
* It doesn't take into account other names in the scope:

image

* It uses qualified names for all nested types:

image

* In some cases, we can't use the qualified name:

image

I have tried to solve the first two problems using `TypeReferencesInfo.toQualify` from `RsImportHelper`:

[up.patch.zip](https://github.com/intellij-rust/intellij-rust/files/5211743/up.patch.zip)

Kobzol:
Thank you for the patch. For the last item, I suppose that we could look for reexports of `S` (via `RsReexportIndex`), take each  reexport, extract the parent module path (`test_package::A`), join it with the reexported name (`S`), check if that path is visible and take first reexport that passes this check. But I'm not sure if it's worth the effort?

mchernyavsky:
I have updated the previous path. Now it searches for an import candidate and uses its usePath while rendering.

image

The `crate::` prefix is sometimes really necessary:

image

[up2.patch.zip](https://github.com/intellij-rust/intellij-rust/files/5226320/up2.patch.zip)

Kobzol:
Thank you! Passing context to the renderer and using reexports were the missing pieces. Now it seems to behave reasonably in all tested cases.

mchernyavsky (approved):
Thanks!

bors r+

Issue #1095, opened 2017-03-27
Inline variable refactoring


matklad:
We have support for "Introduce a variable" already. It would be nice to be able to do the opposite and inline variables as well. 

The initial version should just inline the expression, without checking any borrowing related things.

mcarlin:
Cool if I grab this one?

matklad:
@mcarlin sure!

vlad20012:
The issue seems still up for grab =)
As sample, "Introduce a variable" refactoring located [here](https://github.com/intellij-rust/intellij-rust/blob/fd96ed4524a6365ea89bfad3e761954a1361edbc/src/main/kotlin/org/rust/lang/refactoring/RsLocalVariableHandler.kt) and tests for it - [here](https://github.com/intellij-rust/intellij-rust/blob/fd96ed4524a6365ea89bfad3e761954a1361edbc/src/test/kotlin/org/rust/lang/refactoring/RsLocalVariableHandlerTest.kt)

matklad:
As an example, a similar feature for Kotlin is implemented [here](https://github.com/JetBrains/kotlin/blob/941e5ae1b2c7a3e16d1c9cdf57d647bd2d3f0a35/idea/src/org/jetbrains/kotlin/idea/refactoring/inline/KotlinInlineValHandler.kt).

snpefk:
I'd like to grab this

snpefk:
Hello @matklad! 

Looks like all methods in `RsRefactoringSupportProvider` overrides methods from `RefactoringSupportProvider` which lies in intellij source code. Should I add changes to Intellij codebase or it's not a big deal and I just add new method to `RsRefactoringSupportProvider`?

PR #5459, opened 2020-05-26
REF: add inline value refactoring

Kobzol:
This PR adds support for inlining variables and constants. Right now the original variable is inlined and removed. If it does not have any usages, an error is reported. In Kotlin, there is an additional dialog that asks if the original expression should be kept or not, should I also include it here?

I wasn't sure which test infrastructure to use as it's not a classic refactoring, so I added some manual test functions, but it's not a lot of code.

~Functions could also be inlined, should that be handled inside this refactoring?~

Fixes: https://github.com/intellij-rust/intellij-rust/issues/1095

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionHandler.kt
@@ -3,7 +3,7 @@
  * found in the LICENSE file.
  */
 
-package org.rust.ide.refactoring.inline
+package org.rust.ide.refactoring.inlineFunction
 
 import com.intellij.codeInsight.TargetElementUtil
 import com.intellij.lang.Language
@@ -22,10 +22,6 @@ import org.rust.lang.core.resolve.ref.RsReference
 
 class RsInlineFunctionHandler : InlineActionHandler() {
     override fun isEnabledOnElement(element: PsiElement): Boolean = canInlineElement(element)
-
-    override fun isEnabledOnElement(element: PsiElement, editor: Editor?): Boolean =
-        isEnabledOnElement(element)
-
     override fun inlineElement(project: Project, editor: Editor, element: PsiElement) {
         val function = element as RsFunction
 
@@ -52,7 +48,7 @@ class RsInlineFunctionHandler : InlineActionHandler() {
         }
 
         if (function.block == null) {
-            errorHint(project, editor,"Cannot inline an empty function")
+            errorHint(project, editor, "Cannot inline an empty function")
             return
         }
 
--- a/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
@@ -3,7 +3,7 @@
  * found in the LICENSE file.
  */
 
-package org.rust.ide.refactoring.inline
+package org.rust.ide.refactoring.inlineFunction
 
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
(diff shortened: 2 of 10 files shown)


src/main/kotlin/org/rust/ide/refactoring/inlineValue/InlineValueHandler.kt
```diff
@@ -0,0 +1,92 @@
...
+        runWriteCommandAction(project) {
+            references.forEach {
+                it.element.replace(context.expr)
+            }
+            context.element.delete()
+        }
+    }
+
+    override fun canInlineElement(element: PsiElement): Boolean {
+        if (element.ancestorOrSelf<RsConstant>() != null) return true
```
vlad20012, line 43: I wonder why Java's implementations like `InlineLocalHandler`, `InlineConstantFieldHandler`, `InlineMethodHandler` just check the type of `element` without checking ancestors

Kobzol: If I just check for the type directly, it is not available for the whole constant/let declaration, which seems unintuitive for me:
```rust
const CONST: u32 = 5/*caret*/; // not available
const CONST: u32/*caret*/ = 5; // not available
const CONST:/*caret*/ u32 = 5; // not available
const CONST/*caret*/: u32 = 5; // available
/*caret*/const CONST: u32 = 5; // not available
```

Kobzol: Actually to be on par with Java, we also need to check if the element is a reference to a constant/let decl, for this to work:
```rust
const CONST: u32 = 1;
let a = CONST/*caret*/ + 2;
```

vlad20012: Clear! But this is exactly how it works in Java. The refactoring is available only when the caret is placed near a constant name or a reference to it.

Kobzol: Okay, so should I change it to behave the same as Java/Kotlin?

vlad20012: > Actually to be on par with Java, we also need to check if the element is a reference to a constant/let decl, for this to work

This works automagically right now =)

> so should I change it to behave the same as Java/Kotlin?

I think, yes

Kobzol:
I rewrote the refactoring to be in line (no pun intended) with the InlineFunction refactoring. The UI dialog was mostly the same, so I refactored it out, along with some other things and tests in InlineFunction. There is now also support for testing the various inlining modes (only this/all references/all references and keep original) for InlineValue. Refactoring from references is now also supported.

Kobzol:
Is there anything left to do here? After this PR I'll add support for inlining of parameters in the inline function refactoring.

LoganDark:
Really missing this feature right now. It's much better than manual copy and paste

Kobzol:
@vlad20012 ping

AqlaSolutions:
Yep, I need this too

LoganDark:
I'd like to know why this PR is still sitting around, without any comments from the maintainers on why it's not being merged.

@vlad20012 Replying with an 👀 emoji just serves to tell us that you see this PR and you're intentionally ignoring it.

Kobzol:
Take it easy, the maintainers are doing a lot of work to keep the plugin stable (fixes, performance improvements, internal refactorings etc.) and adding new features might not always be a top priority. I myself got almost 100 PRs merged in the last 6 months, so there's no problem with the maintainers accepting PRs :) I'm sure the review on this PR will continue in a reasonable timeframe.

LoganDark:
They've looked at it multiple times, all they have to do is hit the merge button. I personally don't see the reasoning for saying that they're doing other things and this might not be a top priority; it's priority enough for them to look at it, and react to comments, and review it and ask for changes.

vlad20012 (approved):
I'm very sorry for the long response(
I also apologize to the people who were waiting for this feature.

Thanks!
bors r+

PR #5470, opened 2020-05-29
CI: improve release actions script

Undin:
* use [click](https://click.palletsprojects.com/) library instead of argparse
* update command syntax. Now you don't need to pass `--command` flag to specify a command
* github token can be provided via `IR_GITHUB_TOKEN` environment variable
* added `--no-update-changelog` option for `stable-release` command to omit changelog link update. Useful for releases with hotfixes

### DIFF
--- a/scripts/release-actions.py
+++ b/scripts/release-actions.py
@@ -1,4 +1,6 @@
-import argparse
+import functools
+
+import click
 import requests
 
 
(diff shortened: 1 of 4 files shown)


scripts/release-actions.py
```diff
@@ -1,9 +1,14 @@
-import argparse
+import functools
```
vlad20012, line 1: Hmm `#!/bin/python3` ?

lancelote: Better `#!/usr/bin/env python3` but why? It is not required if you follow `MAINTAINING.md` instructions.

lancelote:
bors merge

Issue #5431, opened 2020-05-24
Highlight items from `std` part of stdlib with `no_std` attribute


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.3110-201-nightly
* **Rust toolchain version:** 1.45.0-nightly (a74d1862d 2020-05-14) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.1.1 Ultimate Edition (IU-201.7223.91)
* **Operating system:** macOS 10.13.6

## Problem description
`println!` macro is a part of `std` crate. But if there is `#![no_std]` attribute, we can't use it. So usage of `println!` is compilation error. It would be great to highlight usages of items from `std` as compilation error in `no_std` context

## Steps to reproduce
```rust
#![no_std]

fn main() {
    println!("Hello, world!");
}
```

Kobzol:
`#![no_std]` is currently stored in the attributes of `RsFile`, but this attribute is global for the whole crate, so it should probably be first moved to a crate/package.

Undin:
> #![no_std] is currently stored in the attributes of RsFile, but this attribute is global for the whole crate, so it should probably be first moved to a crate/package.

Note, currently, `Package`/`Target`/`CargoWorkspace` objects are immutable (almost) and created during `Cargo refresh` phase. And I don't think we want to pass any info from code to these object directly

> I just realized that you can get the root file of the crate and check if it has the #![no_std] attribute. Is that an OK solution?

Yep. See `org.rust.lang.core.psi.RsFile#attributes` property

PR #5474, opened 2020-05-30
ANN: mark std usage in no_std environments as an error

Kobzol:
This PR adds error annotation for usage of `std` items in a `no_std` environment.

Originally, I wanted to check `RsReferenceElement`s to see if they end up in `std`, but I found two problems with that. First, some `std` references end up being resolved as `null` if `no_std` is present, and second, if I resolve to e.g. `Vec`, it actually resides in `alloc`, not `std`, it's just reexported, and `alloc` is fine. Therefore I use an additional, simpler strategy, where I simply check if the path begins with `std`, and if yes, it is marked as an error.

However, this does not work for some cases, for example `vec!` does not start with `std` and resolves to the `alloc` crate, so it's not marked as an error, but it's not available in `#![no_std]`. Generally, things from the `std` prelude are missing. It's a false negative, so not a big deal, but a better solution would be to ignore the whole `std` prelude when `#![no_std]` is present - is it viable to do that?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5431


vlad20012:
Hmm, this was not assigned to anybody for some reason. 

Consider resolving the first path segment and check that it is `std` crate root.
```
val resolved = ... ... resolve()
val resolvedCrate = resolved.crate.
val isStd = resolvedCrate.isStd && resolvedCrate.rootMod == resolved
```

Kobzol:
That's basically what I have described in the PR's description, with `#![no_std]`, the `std` in `std::vec::Vec` resolves to `null`. I'm not sure if we should change that behaviour just for this one annotation though.

vlad20012:
Oh. Ok, it looks like I read this too fluently =D

What if I'll declare a module with name `std`? I think, `ref.basePath().referenceName == AutoInjectedCrates.STD` should be applied only if the is `ref.basePath()` unresolved

Kobzol:
Good point.

If we check `(resolvedCrate?.isStd ?: false) && resolvedCrate?.rootMod == resolved`, the second part of the condition will not hold for macros like `println!`, so they will not be marked as an error. Macro resolving ignores `#![no_std]`, so `println!` is resolved, but it resolves to the macro, not to the `std` mod.

Undin:
In my opinion, we shouldn't come up with some new error in this specific case. Ideally, items from `std` crate shouldn't be resolved in `no_std` context and plugin should annotate them with `Unresolved reference` annotation as compiler does.
Probably, we can add some specific message to this error message if we can determine more precise reason or some specific quick fix.
My initial issue was about unexpected successful name resolution of `println!` macro with `#![no_std]` attribute.

Kobzol:
Oh, I understand. So let's disable `println` and other macros resolve if `#![no_std]` is present (I'll send a PR), and when unresolved reference becomes shown as error by default, it will also solve what I was trying to do in this PR?

Kobzol:
Created https://github.com/intellij-rust/intellij-rust/pull/6033.

Undin:
> Oh, I understand. So let's disable println and other macros resolve if #![no_std] is present (I'll send a PR), and when unresolved reference becomes shown as error by default, it will also solve what I was trying to do in this PR?

Something like this, yes. But looks like this problem is related not only with std macros. For example, `HashMap` in the following example

```rust
#![no_std]

fn main() {
    let mut map = std::collections::HashMap::new();
    map.insert(1, 1);
}
```

can be successfully resolved by plugin although compiler produces `error[E0433]: failed to resolve: use of undeclared type or module "std"`. So it seems that proper fix should be applied to some parts of our resolve algorithm and can be more complex than it seems

Undin:
@vlad20012 @dima74 let's check that our name resolution 2.0 handle these cases properly

dima74:
> let's check that our new name resolution algorithm (AKA name resolution 2.0) processes these cases properly

Thank you, I found some bugs related to `no_std` in name resolution 2.0 implementation, will fix them

Kobzol:
@Undin should this be closed?

Kobzol:
@dima74 The code provided by @Undin still resolves with name resolution 2.0 on master. Should I create a separate issue for it?

dima74:
@Kobzol thanks, created #7054

dima74:
I guess we can close this PR since now macros and items from std are unresolved with `#![no_std]`

Kobzol:
I'm not sure that it's the same, here we would give an explicit error rather than a general unresolved error. But probably when you use `#[no_std]`, you usually know that using `std` it not a good idea :) Feel free to close.

Issue #4601, opened 2019-11-07
Better inlay hints for generic enum variants


SomeoneToIgnore:
## Environment

* **IntelliJ Rust plugin version:** 0.2.110.2782-193-nightly
* **Rust toolchain version:** 1.38.0 (625451e37 2019-09-23) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.4932.12)
* **Operating system:** macOS 10.15.1

## Problem description

For the code

```rust
fn test() -> Result<(), ()> {
    Ok(())
}
```

with the setting "show only smart hints" set to `false`, I get the following hint currently:

![image](https://user-images.githubusercontent.com/2690773/68371132-affaa900-0146-11ea-93e5-f53b823b2e10.png)

It can be turned into a more useful one if you display a generic name: `Ok(T: ())``

PR #5475, opened 2020-05-30
HINT: add type parameter name to non-smart parameter hints of generic enum variants

Kobzol:
```rust
fn test() -> Result<(), ()> {
    Ok(/*T: */())
}
```

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4601

changelog: Show type parameters in inlay hints for generic enum variants

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/CallInfo.kt
+++ b/src/main/kotlin/org/rust/ide/utils/CallInfo.kt
@@ -6,12 +6,10 @@
 package org.rust.ide.utils
 
 import org.rust.ide.presentation.getStubOnlyText
-import org.rust.lang.core.psi.RsCallExpr
-import org.rust.lang.core.psi.RsFunction
-import org.rust.lang.core.psi.RsMethodCall
-import org.rust.lang.core.psi.RsPathExpr
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.ty.TyFunction
+import org.rust.lang.core.types.ty.TyTypeParameter
 import org.rust.lang.core.types.type
 
 class CallInfo private constructor(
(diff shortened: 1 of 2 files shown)


ortem:
@Kobzol I'm not sure should we enable these hints only when `Show only smart hints` is disabled.  Maybe it's better to show them even with smart hints enabled but also provide a special checkbox (under smart hints checkbox) to disable them. What do you think?

Kobzol:
To be honest I don't usually use these hints, so I'm not sure. It sounds like a "smart" hint, since it's special case for enum variants with a type parameter, so enabling it in smart mode and adding an explicit checkbox sounds OK to me.

ortem:
@Kobzol Are you to add a special checkbox in this PR?

Kobzol:
I added the checkbox, but I don't know how to "nest" it below the smart hints option or how to make it dependent on it. Maybe this should be handled by https://github.com/intellij-rust/intellij-rust/pull/5789? But it would probably have to be migrated to the `InlayHintsProvider` API.

Kobzol:
@ortem ping :)

ortem:
Hey @Kobzol, sorry for the long wait! I looked into the current UX and find it a bit confusing. And I'm not if sure there is a good way to nest these options or something like that. For me, it looks like the initial implementation (that ignores smart mode) was simpler and convenient to use. Sorry for my uncertain review of this PR.

Kobzol:
I agree that the UX isn't great currently. Soo what should I do? :D Remove the checkbox and use this implementation as initially, i.e. only when "show only smart hints" is disabled?

ortem:
Yes, let's remove the checkbox and show such hints only in non-smart mode

ortem:
@Kobzol Note that the new hints are shown even in smart mode in case of multiple type parameters:

Smart mode is **disabled**
Screenshot 2021-01-13 at 22 46 17

Smart mode is **enabled**
Screenshot 2021-01-13 at 22 46 35

I'm not sure if it is desired (as for me, it's generally fine), but anyway it would be great to have tests on this case.

Kobzol:
I agree, I actually had tests like that in the "non-initial" implementation with the checkbox :D Added them again.

ortem (approved):
Thank you!
bors r+

vlad20012:
This should also work for tuple structs:
```rust
struct S(i32);

fn main() {
    S(1);
}
```
image

Kobzol:
So how exactly should it work? Do you want the hints shown for generic parameters of tuple structs? Or just hide the useless `_:`?

PR #5479, opened 2020-06-01
Drop support of 2019.3 platform

Undin:
Closes #5454


Undin:
bors retry

Stzx:
@Undin Can I clean up some APIs marked `@Deprecated`?

Undin:
@Stzx yep. But most of the current deprecations are about new annotation API and I'm not sure that it's very convenient to use. Probably, we need some additional (extension) methods for it.

PR #5481, opened 2020-06-02
ANN: check if the `main` function exists [E0601]

### LINKED ISSUE #5477: Support "`main` function not found" error
## Environment

* **Rust toolchain version:** 1.32.0 (9fda7c223 2019-01-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2 EAP Ultimate Edition (IU-202.4357.23)

## Problem description
Binary crate root should have `main` function. Otherwise, the compiler produces [E0601](https://doc.rust-lang.org/error-index.html#E0601) error. Currently, the plugin says nothing about it.
I think we can add error annotation to `RsFile` if it's is a crate root and the corresponding target is binary and provide quick fix to add `main` function to the file.
Note, there is `#![no_main]` attribute.

Stzx:
Closed #5477

<del>There is currently no way to quickly test `[[bin]]`, so I did not add a test. ;(</del>

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/AddMainFnFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/AddMainFnFix.kt
@@ -1,0 +1,24 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsPsiFactory
+
+class AddMainFnFix(file: PsiElement) : LocalQuickFixAndIntentionActionOnPsiElement(file) {
+
+    override fun getFamilyName(): String = "Add `fn main()`"
+
+    override fun getText(): String = familyName
+
+    override fun invoke(project: Project, file: PsiFile, editor: Editor?, startElement: PsiElement, endElement: PsiElement) {
+        file.add(RsPsiFactory(project).createFunction("fn main() { }"))
+    }
+}
--- a/src/main/resources/META-INF/rust-core.xml
+++ b/src/main/resources/META-INF/rust-core.xml
@@ -535,6 +535,11 @@
                          enabledByDefault="true" level="WARNING"
                          implementationClass="org.rust.ide.inspections.RsLivenessInspection"/>
 
+        <localInspection language="Rust" groupName="Rust"
+                         displayName="Main function not found"
+                         enabledByDefault="true" level="ERROR"
+                         implementationClass="org.rust.ide.inspections.RsMainFunctionNotFoundInspection"/>
+
         <!-- Surrounders -->
 
         <lang.surroundDescriptor language="Rust"
--- a/src/main/resources/inspectionDescriptions/RsMainFunctionNotFound.html
+++ b/src/main/resources/inspectionDescriptions/RsMainFunctionNotFound.html
@@ -1,0 +1,1 @@
+Checks if the <code>main</code> function exists in the binary crates.
--- a/src/test/kotlin/org/rust/RustProjectDescriptors.kt
+++ b/src/test/kotlin/org/rust/RustProjectDescriptors.kt
@@ -27,6 +27,7 @@ import org.rust.cargo.project.workspace.StandardLibrary
 import org.rust.cargo.toolchain.RustToolchain
 import org.rust.cargo.util.DownloadResult
 import java.nio.file.Paths
+import java.util.*
 
 object DefaultDescriptor : RustProjectDescriptorBase()
 
@@ -63,14 +64,18 @@ open class RustProjectDescriptorBase : LightProjectDescriptor() {
             CargoWorkspaceData(packages, emptyMap()), CfgOptions.DEFAULT)
     }
 
-    protected fun testCargoPackage(contentRoot: String, name: String = "test-package") = CargoWorkspaceData.Package(
+    protected fun testCargoPackage(contentRoot: String, name: String = "test-package") = Package(
         id = "$name 0.0.1",
         contentRootUrl = contentRoot,
         name = name,
         version = "0.0.1",
         targets = listOf(
             Target("$contentRoot/main.rs", name, TargetKind.Bin, edition = Edition.EDITION_2015, doctest = true),
-            Target("$contentRoot/lib.rs", name, TargetKind.Lib(LibKind.LIB), edition = Edition.EDITION_2015, doctest = true)
+            Target("$contentRoot/lib.rs", name, TargetKind.Lib(LibKind.LIB), edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/bin/a.rs", name, TargetKind.Bin, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/bench/a.rs", name, TargetKind.Bench, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/example/a.rs", name, TargetKind.ExampleBin, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/example-lib/a.rs", name, TargetKind.ExampleLib(EnumSet.of(LibKind.LIB)), edition = Edition.EDITION_2015, doctest = true)
         ),
         source = null,
         origin = PackageOrigin.WORKSPACE,
@@ -124,10 +129,11 @@ open class WithCustomStdlibRustProjectDescriptor(
         StandardLibrary.fromPath(path)
     }
 
-    override val skipTestReason: String? get() {
-        if (stdlib == null) return "No stdlib"
-        return delegate.skipTestReason
-    }
+    override val skipTestReason: String?
+        get() {
+            if (stdlib == null) return "No stdlib"
+            return delegate.skipTestReason
+        }
 
     override fun testCargoProject(module: Module, contentRoot: String): CargoWorkspace =
         delegate.testCargoProject(module, contentRoot).withStdlib(stdlib!!, CfgOptions.DEFAULT)
--- a/src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
@@ -1,0 +1,108 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import org.rust.ProjectDescriptor
+import org.rust.WithDependencyRustProjectDescriptor
+
+class RsMainFunctionNotFoundInspectionTest : RsInspectionsTestBase(RsMainFunctionNotFoundInspection::class) {
+
+    fun `test does not exist main function`() = checkByFileTree("""
+    //- main.rs
+        <error descr="`main` function not found in crate `test-package` [E0601]"> /*caret*/ </error>
+    """)
+
+    fun `test has nested main function`() = checkByFileTree("""
+    //- main.rs
+        <error descr="`main` function not found in crate `test-package` [E0601]">fn foo() {
+             fn main() {
+                /*caret*/
+             }
+        }</error>
+    """)
+
+    fun `test exists no_main attr`() = checkByFileTree("""
+    //- main.rs
+        #![no_main]/*caret*/
+    """)
+
+    fun `test exists main function`() = checkByFileTree("""
+    //- main.rs
+       fn main() {
+            /*caret*/
+       }
+    """)
+
+    fun `test that the main function does not exist in the custom bin`() = checkByFileTree("""
+    //- bin/a.rs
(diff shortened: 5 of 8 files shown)


src/main/kotlin/org/rust/ide/annotator/RsFileErrorAnnotator.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.rust.cargo.project.model.impl.allTargets
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsInnerAttr
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.descendantsOfType
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.lang.utils.addToHolder
+
+class RsFileErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
```
Undin, line 23: Let's better convert it into inspection (see `RsLocalInspectionTool` as base class).
It preferable for users because they can disable any inspection via settings. It allows to provide easy workaround in case of annoying false positive annotatons

src/main/kotlin/org/rust/ide/annotator/RsFileErrorAnnotator.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun annotateInternal(element: PsiElement, holder: AnnotationHolder) {
+        val visitor = object : RsVisitor() {
+            override fun visitFile(file: PsiFile) { checkRsFile(holder, file as? RsFile ?: return) }
+        }
+
+        element.accept(visitor)
+    }
+
+    private fun checkRsFile(holder: AnnotationHolder, file: PsiFile) {
```
Undin, line 35:
```suggestion
    private fun checkRsFile(holder: AnnotationHolder, file: RsFile) {
```

src/main/kotlin/org/rust/ide/annotator/RsFileErrorAnnotator.kt
```diff
@@ -0,0 +1,43 @@
...
+        val binCargoProject = file.project.cargoProjects.allTargets.lastOrNull { target -> target.isBin && file.virtualFile == target.crateRoot } ?: return
```
Undin, line 36: You can use `org.rust.lang.core.psi.RsFile#isCrateRoot` and `org.rust.lang.core.psi.RsFile#cargoTarget` properties

src/main/kotlin/org/rust/ide/annotator/RsFileErrorAnnotator.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        if (file.descendantsOfType<RsInnerAttr>().count { rsInnerAttr -> "no_main" == rsInnerAttr.metaItem.text } == 1) return
```
Undin, line 38:
```suggestion
        if (file.queryAttributes.hasAttribute("no_main")) return
```

src/main/kotlin/org/rust/ide/annotator/RsFileErrorAnnotator.kt
```diff
@@ -0,0 +1,43 @@
...
+        if (file.descendantsOfType<RsFunction>().lastOrNull { fn -> "main" == fn.name } != null) return
```
Undin, line 39: `descendantsOfType` scans all chindren in AST (not only direct). So if you have any function with `main` name in file, it finds it. But it but that is not what we want

Stzx:
> test u8 in enum variant discriminant repr() FAILED

This seems to have nothing to do with this PR.

src/main/kotlin/org/rust/ide/inspections/fixes/AddMainFnFix.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.RsPsiFactory
+
+class AddMainFnFix(file: RsFile) : LocalQuickFixAndIntentionActionOnPsiElement(file) {
+
+    override fun getFamilyName(): String = "Add `main` fn"
+
+    override fun getText(): String = "Add `fn main()`"
```
Undin, line 20: I don't think it makes sense to have different text in `getFamilyName` and `getText`

src/main/kotlin/org/rust/ide/inspections/RsFileInspection.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    private fun checkMainFn(file: RsFile, manager: InspectionManager, isOnTheFly: Boolean): ProblemDescriptor? {
+        val target = file.cargoTarget ?: return null
+        val cargoName = if (target.isBin && file.isCrateRoot) target.name else return null
+
+        if (file.queryAttributes.hasAttribute("no_main")) return null
+        if (file.childrenOfType<RsFunction>().lastOrNull { fn -> "main" == fn.name } != null) return null
+
+        return manager.createProblemDescriptor(file,
```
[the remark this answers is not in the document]
Stzx: @Undin 

I originally thought about integrating it into `RsDiagnostic`. But there is no convenient way to set `RsProblemsHolder` to file level

I checked the documentation, but the documentation does not say that `buildVisitor` can upgrade the error to the file level.

Maybe I missed something?

Kobzol:
Regarding the inspection name: In this PR https://github.com/intellij-rust/intellij-rust/pull/5490 I'm also creating a file-level inspection which checks a different thing. I think that the inspection in this PR should be named `RsFileContainsMainFunctionInspection` or something like that, and not `RsFileInspection`. The other inspections are also named after a specific check and not after the element that they are checking (i.e. the inspection that checks variable mutability is called`RsVariableMutableInspection`, it's not a specific method inside a `RsPatBindingInspection` - such inspection does not exist). What do you think? :)

src/main/kotlin/org/rust/ide/inspections/RsFileInspection.kt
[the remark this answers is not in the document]
Undin: I don't understand the cause of the problem
Just add new diagnostic class
```kotlin
class DiagnosticName(e: RsFile): RsDiagnostic(e, inspectionClass = RsFileInspection::class.java) {
        return PreparedAnnotation(
            ERROR,
            ErrorCode,
            "Erorr message",
            quickFixObject
        )
    }
}
```

And add it into problems holder in visitor:
```kotlin
    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor? {
        return object : RsVisitor() {
            override fun visitFile(file: PsiFile) {
                if (file is RsFile && necessaryChecks) {
                    RsDiagnostic.DiagnosticName(file).addToHolder(holder)
                }
            }
        }
    }
```

It seems it should work

src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        }</error>
+    """)
+
+    fun `test that no_main attr exists in the custom example bin`() = checkByFileTree("""
+    //- example/a.rs
+        #![no_main]
+        /*caret*/
+    """)
+
+    fun `test main function in custom binTest main function in custom example bin`() = checkByFileTree("""
```
Undin, line 79: Looks like you duplicate test name

src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
```diff
@@ -0,0 +1,85 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+class RsMainFunctionNotFoundInspectionTest : RsInspectionsTestBase(RsMainFunctionNotFoundInspection::class) {
+
+    fun `test does not exist main function`() = checkErrors("""
```
Undin, line 10: Maybe it's better to use `checkByFileTree` for all tests to be consistent and explicitly show file name because it's important here?

src/main/resources/inspectionDescriptions/RsMainFunctionNotFound.html
```diff
@@ -0,0 +1 @@
+Check if the main function exists in the binary crates.
```
Undin, line 1:
```suggestion
Checks if the <code>main</code> function exists in the binary crates.
```

src/test/kotlin/org/rust/RustProjectDescriptors.kt
```diff
@@ -63,14 +63,16 @@ open class RustProjectDescriptorBase : LightProjectDescriptor() {
...
         id = "$name 0.0.1",
         contentRootUrl = contentRoot,
         name = name,
         version = "0.0.1",
         targets = listOf(
             Target("$contentRoot/main.rs", name, TargetKind.Bin, edition = Edition.EDITION_2015, doctest = true),
-            Target("$contentRoot/lib.rs", name, TargetKind.Lib(LibKind.LIB), edition = Edition.EDITION_2015, doctest = true)
+            Target("$contentRoot/lib.rs", name, TargetKind.Lib(LibKind.LIB), edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/bin/a.rs", name, TargetKind.Bin, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/example/a.rs", name, TargetKind.ExampleBin, edition = Edition.EDITION_2015, doctest = true)
```
Undin, line 75: Let's also add test, bench and example lib targets and check that we don't show error in these files as well

src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
```diff
@@ -0,0 +1,85 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+class RsMainFunctionNotFoundInspectionTest : RsInspectionsTestBase(RsMainFunctionNotFoundInspection::class) {
```
Undin, line 8: Let's also add test that we don't highlight anything if file is not create root but it's in binary crate

src/main/resources/META-INF/rust-core.xml
```diff
@@ -535,6 +535,11 @@
                          enabledByDefault="true" level="WARNING"
                          implementationClass="org.rust.ide.inspections.RsLivenessInspection"/>
 
+        <localInspection language="Rust" groupName="Rust"
+                         displayName="Main function checker"
```
Undin, line 539:
```suggestion
                         displayName="Main function not found"
```

src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
Stzx: Yes. It may be better to explicitly indicate the file name.

src/test/kotlin/org/rust/RustProjectDescriptors.kt
Stzx: For the example lib, should I add other types besides `LibKind.LIB` such as `LibKind.STATICLIB`?

Undin: I suppose single example lib with `LibKind.LIB` is enough

src/test/kotlin/org/rust/RustProjectDescriptors.kt
```diff
@@ -13,6 +13,7 @@ import com.intellij.testFramework.LightProjectDescriptor
 import com.intellij.testFramework.VfsTestUtil
 import com.intellij.testFramework.fixtures.CodeInsightTestFixture
 import com.intellij.util.Urls
+import com.jetbrains.rd.util.enumSetOf
```
Undin, line 16: Don't forget to drop this

src/test/kotlin/org/rust/RustProjectDescriptors.kt
```diff
@@ -63,14 +64,18 @@ open class RustProjectDescriptorBase : LightProjectDescriptor() {
...
         name = name,
         version = "0.0.1",
         targets = listOf(
             Target("$contentRoot/main.rs", name, TargetKind.Bin, edition = Edition.EDITION_2015, doctest = true),
-            Target("$contentRoot/lib.rs", name, TargetKind.Lib(LibKind.LIB), edition = Edition.EDITION_2015, doctest = true)
+            Target("$contentRoot/lib.rs", name, TargetKind.Lib(LibKind.LIB), edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/bin/a.rs", name, TargetKind.Bin, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/bench/a.rs", name, TargetKind.Bench, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/example/a.rs", name, TargetKind.ExampleBin, edition = Edition.EDITION_2015, doctest = true),
+            Target("$contentRoot/example-lib/a.rs", name, TargetKind.ExampleLib(enumSetOf(setOf(LibKind.LIB))), edition = Edition.EDITION_2015, doctest = true)
```
Undin, line 78:
```suggestion
Target("$contentRoot/example-lib/a.rs", name, TargetKind.ExampleLib(EnumSet.of(LibKind.LIB)), edition = Edition.EDITION_2015, doctest = true)
```

Stzx:
Thanks for your review ❤️ 

bors r+

PR #5484, opened 2020-06-02
REF: Support reexports in move single file refactoring

dima74:
Improve updating references to moved file when it is under reexport

We use `RsImportHelper` for finding new paths. Please see comment for `RsMovePathHelper` for detailed description.

---

Some background from #5181:

> Currently move refactoring doesn't work well with reexports (examples are in `RsMoveFileReexportTest`). Main issue is that we replace old paths to moved file with absolute path to new moved file location (`crateRelativePath`). It can cause visibility problems if for example new parent module is private, but grandparent module publicly reexports all items from parent module (I think those kinds of reexports are common in libraries). This issue will be solved in future PRs by using `RsImportHelper` for finding new paths.


dima74:
Resolved merge conflicts (after #5688)

PR #5489, opened 2020-06-03
TYPE: Fix run icon display

Stzx:
The run icon is not displayed when `main function` is declared in a nested method.


mchernyavsky:
Thanks!

Issue #5488, opened 2020-06-03
Automatically add file file top module tree or display option to do so.


FelixBrakel:
Currently when you create a new `.rs` file in a project it will give you the following notification:

![Screenshot_20200603_131403](https://user-images.githubusercontent.com/7832280/83630534-3a4fa200-a59c-11ea-8292-d4db23c22392.png)

Which is nice. However I thought it would be nice for this notification to include a button to automatically add the following statement to the corresponding `mod.rs` file (Or whatever the module file is when using the rust 2018 syntax).

```rust
mod scene
```

Such that the file scene.rs is included in the module tree.

Or automatically write this statement when a new `.rs` file is created. I can't really see a scenario where you wouldn't want to add a file to the module tree so I would personally prefer the latter option but I wouldn't presume to know other people's workflow so the former option might be better because it's less intrusive.

PR #5490, opened 2020-06-03
INSP: add attach file to module quick fix

Kobzol:
This PR adds an action to `DetachedFileNotificationProvider` that adds the file to some module so that it will be present in the module tree. The current implementation is not yet complete because I'm not sure about the UI.

image

The directory of the file is checked for `mod.rs`, `lib.rs` or `main.rs` files and if they are valid and exist, they will be used as candidates for insertion. Currently if there is only one available module, I just insert the file into it, if there are more, some UI/dialog could show up - a select box with available modules?

Should it offer modules recursively, i.e. in grandparent modules? That would be pretty complicated if the parent module would already exist, but it would allow to include very deeply nested files (`a/b/c/foo.rs`, where neither `a`, `b` nor `c` are modules yet).

Is it possible to test this? I tried to access the intention action from the notification provider and trigger it, but that seems pretty hacky and I don't have access to an editor in the current notification provider test suite.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5488


Undin:
> Should it offer modules recursively, i.e. in grandparent modules? That would be pretty complicated if the parent module would already exist, but it would allow to include very deeply nested files (a/b/c/foo.rs, where neither a, b nor c are modules yet).

I think no. At least for now.

> Is it possible to test this? I tried to access the intention action from the notification provider and trigger it, but that seems pretty hacky and I don't have access to an editor in the current notification provider test suite.

Let's convert it into inspection that produces file-level warning. See #5481 as example

Kobzol:
Just to be sure, you want me to convert the whole `DetachedFileNotificationProvider` to an inspection, not create an additional inspection just for adding the file to a module, right?

Kobzol:
Updated module search and created a dialog for choosing the module. If there is a single module, the action button will say "Attach file to $modname", otherwise it says "Attach to a module".

Should I convert the whole `RsNotificationProvider` to a file-level inspection for easier testing?

Undin:
> Should I convert the whole RsNotificationProvider to a file-level inspection for easier testing?

@Kobzol I think so

Undin:
@Kobzol Oops, you are about `RsNotificationProvider`. Sorry I didn't notice it. No, let's convert only `DetachedFileNotificationProvider`

Kobzol:
Yeah, that's what I meant, sorry.

Kobzol:
I converted the provider to an inspection with a quick fix, however if I try to use a GUI dialog inside the quickfix, it complaints:
`java.lang.Throwable: AWT events are not allowed inside write action: java.awt.event.InvocationEvent[INVOCATION_DEFAULT,runnable=com.intellij.openapi.application.impl.FlushQueue$FlushNow@6b832df7,notifier=null,catchExceptions=false,when=1591344167928] on sun.awt.X11.XToolkit@152bb963`

How should I handle module selection if there are multiple modules to select?

Kobzol:
Ah-ha! That was the missing piece, thanks. Indeed the exception message was very informative, I just forgot that it's possible to "turn off" write action in quick fixes.

Kobzol:
The inspection doesn't offer any disabling action:
image
But it can still be disabled in Inspection settings. Is that ok or should I add an explicit disabling action somehow?

vlad20012:
> But it can still be disabled in Inspection settings.

Well, yes, but `DetachedFileNotificationProvider` can be disabled for a particular VirtualFile. See `DetachedFileNotificationProvider.disablingKey`

> Is that ok or should I add an explicit disabling action somehow?

I think we should have the ability to disable the inspection in a particular file. It's because we still often can't find a parent module because of macros or something. So, the inspection is useful in general, but in some files, it can be very annoying. Including external library files.
We could leave the same behavior for the inspections - use `PropertiesComponent`, check the property before running the inspection

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -0,0 +1,160 @@
...
+    companion object {
+        fun findAvailableModulesForFile(project: Project, file: RsFile): List<RsMod> {
+            val virtualFile = file.virtualFile ?: return emptyList()
+            val pkg = project.cargoProjects.findPackageForFile(virtualFile) ?: return emptyList()
+
+            val directory = virtualFile.parent ?: return emptyList()
+            val modules = mutableListOf<RsMod>()
+
+            // mod.rs in the same directory
+            modules.addIfNotNull(findModule(file, project, directory.findFileByRelativePath("mod.rs")))
```
vlad20012, line 73: Btw all these heuristics are not applicable if the file is mod.rs itself.

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -0,0 +1,160 @@
...
+            selectModule(rsFile, availableModules)?.let { insertFileToModule(rsFile, it) }
+        }
+    }
+
+    override fun startInWriteAction(): Boolean {
+        return false
+    }
+
+    companion object {
+        fun findAvailableModulesForFile(project: Project, file: RsFile): List<RsMod> {
```
vlad20012, line 65: Let's make it private and add `isAvailable` method

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
Kobzol: Good catch, I'll add support for that.

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
Kobzol: I'm using `findAvailableModulesForFile` in `RsDetachedFileInspection` to choose the label ("Attach to X" or generic "Attach to module") to pass to the quick fix. I did this to avoid recomputing the available modules 1) in the inspection 2) in the quickfix constructor (for it to be available in `getText`) 3) in `invoke`. If it's not a big deal (it's O(target count)), I'll change it.

Kobzol:
Is there a way to modify the notification bar without creating an additional quick fix just for disabling the inspection? I tried overriding `createOptionsPanel`, but that does not seem to do anything. I also looked at `RsSuppressQuickFix`, should I modify/clone it for this inspection?

vlad20012:
Hmm, I'm not sure how to do this better. But it seems that a new quick fix is the easiest way. It can be an inline class btw.

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
vlad20012: Oh, I didn't notice.

Undin:
> I tried overriding createOptionsPanel, but that does not seem to do anything.

@Kobzol JFYI it creates panel for inspection settings. See `Preferences | Editor | Inspections | Rust | Unresolved Reference` as example

vlad20012:
Hmm. Also, maybe we should just navigate a user to the inserted `mod foo;` declaration after applying the fix?

Kobzol:
I wanted to do that, but wasn't sure how to get an `Editor` instance. Just inherit from `LocalQuickFixAndIntentionActionOnPsiElement`?

vlad20012:
Hmm, why do you need an editor instance?

Kobzol:
Good question :D I wanted to do something like `editor.caretModel.moveToOffset(inserted.textOffset)`. But then I realized that the editor of the inspection will be for the wrong file. What API should I use to navigate the user to the inserted `mod`?

vlad20012:
`modItem.navigate(true)`, I guess

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -0,0 +1,189 @@
...
+    WriteCommandAction.runWriteCommandAction(project) {
+        val child = mod.firstChild
+        val inserted = if (child == null) {
+            mod.add(modItem)
+        } else {
+            mod.addBefore(modItem, child)
+        } as RsModDeclItem
+        inserted.navigate(true)
+
+        project.showBalloon("File ${file.name} added to ${mod.containingFile.virtualFile.path}",
```
vlad20012, line 166: Looks like the balloon is not needed now because navigation is enough?

vlad20012:
It's perfect now! Could you please squash the commits?

@Undin, would you like to take a look?

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -0,0 +1,187 @@
...
+                }
+            }
+
+            return modules
+        }
+    }
+}
+
+private fun selectModule(file: RsFile, availableModules: List<RsMod>): RsMod? {
+    if (isUnitTestMode) return MOCK!!(file, availableModules)
```
Undin, line 109: Let's convert `MOCK!!` into `MOCK ?: error ("You should set mock module selector via `withMockModuleAttachSelector`")`
It may help to quickly understand why your test fail

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -0,0 +1,187 @@
...
+        }
+        renderer = SimpleListCellRenderer.create("") {
+            val root = it.containingCargoPackage?.rootDirectory
+            val path = it.containingFile.virtualFile.pathAsPath
+            (root?.relativize(path) ?: path).toString()
+        }
+    }
+
+    val dialog = dialog("Select a module", panel {
+        row { box() }
```
Undin, line 124:
```suggestion
        row { box(CCFlags.growX) }
```

I think 
Screen Shot 2020-06-08 at 10 38 19
is better than

Screen Shot 2020-06-08 at 10 35 35

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -0,0 +1,187 @@
...
+ * foo.rs (not attached)
+ * lib.rs
+ *
+ * After fix:
+ * foo.rs
+ * lib.rs
+ *  mod foo;
+ */
+class AttachFileToModuleFix(file: RsFile,
+                            private val targetModuleName: String? = null) : LocalQuickFixOnPsiElement(file) {
```
Undin, line 48:
```suggestion
class AttachFileToModuleFix(
    file: RsFile,
    private val targetModuleName: String? = null
) : LocalQuickFixOnPsiElement(file) {
```

src/main/resources/META-INF/rust-core.xml
```diff
@@ -540,6 +540,11 @@
                          enabledByDefault="true" level="ERROR"
                          implementationClass="org.rust.ide.inspections.RsMainFunctionNotFoundInspection"/>
 
+        <localInspection language="Rust" groupName="Rust"
+                         displayName="Detached file inspection"
```
Undin, line 544: `Inspection` word is useless here because this name is already shown in inspection settings

src/test/kotlin/org/rust/ide/inspections/RsDetachedFileInspectionTest.kt
```diff
@@ -0,0 +1,183 @@
...
+            /*caret*/mod a;
+
+            fn test() {}
+        //- a/mod.rs
+    """, "lib.rs")
+
+    private fun checkFixWithMultipleModules(fixName: String,
+                                            @Language("Rust") before: String,
+                                            @Language("Rust") after: String,
+                                            moduleName: String) {
```
Undin, line 176:
```suggestion
    private fun checkFixWithMultipleModules(
        fixName: String,
        @Language("Rust") before: String,
        @Language("Rust") after: String,
        moduleName: String
    ) {
```

src/test/kotlin/org/rust/ide/inspections/RsDetachedFileInspectionTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        //- main.rs
+            fn main() {}
+        //- lib.rs
+            /*caret*/mod a;
+
+            fn test() {}
+        //- a/mod.rs
+    """, "lib.rs")
+
+    private fun checkFixWithMultipleModules(fixName: String,
```
Undin, line 173: Looks like you can inline `fixName` parameter because it's always the same

Undin:
> Hmm. Also, maybe we should just navigate a user to the inserted mod foo; declaration after applying the fix?

@vlad20012 @Kobzol I have some doubts about it. Looks like it may be quite annoying. I imagine the following user case: I created a new rust module to write some new code in it but forgot to create mod declaration for it (or did it intentionally because of existence of the corresponding quick fix). I apply quick fix to get proper code insight in the file and IDE moves me into another file. Why? What should a user get from this navigation? I wanted to write new code in new file. But after navigation, I have to move back.
Also, it's inconsistent with other actions. For example, we already have quick fix to make an item public. It may be located in any place in the project (i.e. it does a similar thing - makes some changes in an arbitrary place in the project). But we don't navigate a user to the place where quick fix was applied
What do you think?

Kobzol:
Now that I think about it, the navigation was mainly useful before there was the UI dialog, for the user to see where was the `mod` created.

However, inserting a `mod` is a bit different from the make public quick fix, because that one simply prepends a visibility restriction to an already existing item, while this quick fix creates a new line and in most cases I think that the `mod ...` line will be in a weird location, so the user might want to fix it. We could possibly try to find some existing `mod ...` item in the target file and put the new mod item below it?

Regarding new file creation, maybe we should offer a similar dialog when the user creates a new Rust file, so that it will be added to some module right away?

vlad20012:
@Undin I know a lot of people (mostly newbies in Rust) who asked "what does this message mean" because they don't understand the Rust module system. I think it's necessary to show an inserted mod declaration because otherwise, they will not understand what's happened. 
Another point - there are multiple options for how a module can be introduced, and I think this quick fix doesn't cover all of them, so it can place a module to a wrong place. Navigating a user to the inserted declaration is the best way to let the user check and fix it. Also, it's simple to go back by clicking Ctrl+B, if everything looks ok

vlad20012:
@Undin One more point - I think that this quick fix is mostly for newbies because professional users should write module declarations prior to creating files (then files can be created with another provided quick fix). So this quick fix should be rather obvious and transparent than handy for everyday usage.

Undin:
> I know a lot of people (mostly newbies in Rust) who asked "what does this message mean" ("File is not included in module tree") because they don't understand the Rust module system. I think it's necessary to show an inserted mod declaration because otherwise, they will not understand what's happened.

But what's about a situation when I've already known how the module system works?

> One more point - I think that this quick fix is mostly for newbies because professional users should write module declarations prior to creating files (then files can be created with another provided quick fix)

You don't know because currently, this quick fix doesn't exist and no one can use it in development workflow.
But I don't know too. So let's try current implementation and fix it in future if there are some user complaints

vlad20012:
I just came up with another option: we can show a balloon as before, but with a button that navigates to an inserted module declaration. But ok, let's leave it as is for now

Issue #5282, opened 2020-04-26
Put struct fields on separate lines after "Extract enum variant"


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.122.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.43.0 (4fb7144ed 2020-04-20) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 (CL-201.6668.126)
* **Operating system:** macOS 10.15.3

## Problem description

I guess it makes sense to put struct fields on separate lines in the first place after "Extract enum variant" action (#5182) call as rustfmt enforce.

## Steps to reproduce

```rust
enum A {
    Variant1 { name: String, id: usize, }
}
```

call "Extract enum variant"

```rust
struct Variant1 { pub name: String, pub id: usize, }

enum A {
    Variant1(Variant1)
}
```

after running `cargo fmt`

```rust
struct Variant1 {
    pub name: String,
    pub id: usize,
}

enum A {
    Variant1(Variant1),
}
```

Undin:
Looks like it should be implemented from formatter side (not in the intention) and it will be automatically applied by the platform to all generated code (including described intention)

Kobzol:
@Undin Could you please point me to the corresponding code of the formatter which affects formatting of auto-generated code?

PR #5491, opened 2020-06-03
FMT: add line break after opening brace in struct blocks

Kobzol:
This PR forces a line break after a left opening brace in struct blocks. This should bring struct block formatting closer to rustfmt. This change will probably ruin more tests, let's see what CI says.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5282

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/processors/RsTrailingCommaFormatProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/processors/RsTrailingCommaFormatProcessor.kt
@@ -14,8 +14,10 @@ import com.intellij.psi.impl.source.codeStyle.PostFormatProcessorHelper
 import org.rust.ide.formatter.impl.CommaList
 import org.rust.ide.formatter.rust
 import org.rust.lang.RsLanguage
+import org.rust.lang.core.psi.RsBlockFields
 import org.rust.lang.core.psi.RsElementTypes.COMMA
 import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.RsStructItem
 import org.rust.lang.core.psi.ext.elementType
 import org.rust.lang.core.psi.ext.getNextNonCommentSibling
 import org.rust.lang.core.psi.ext.getPrevNonCommentSibling
@@ -83,7 +85,12 @@ fun CommaList.addTrailingCommaForElement(list: PsiElement): Boolean {
     )
     if (!trailingSpace.contains('\n')) return false
 
-    if (list.firstChild.getNextNonCommentSibling() == lastElement) return false
+    if (list.firstChild.getNextNonCommentSibling() == lastElement) {
+        // allow trailing comma only for struct block fields with a single item
+        if (list !is RsBlockFields || list.parent !is RsStructItem) {
+            return false
+        }
+    }
 
     val comma = RsPsiFactory(list.project).createComma()
     list.addAfter(comma, lastElement)
--- a/src/test/kotlin/org/rust/ide/refactoring/RsExtractEnumVariantTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/RsExtractEnumVariantTest.kt
@@ -42,7 +42,11 @@ class RsExtractEnumVariantTest : RsTestBase() {
             V2
         }
     """, """
-        struct /*caret*/V1 { pub a: i32, pub b: bool, pub c: String }
+        struct /*caret*/V1 {
+            pub a: i32,
+            pub b: bool,
+            pub c: String
+        }
 
         enum A {
             V1(V1),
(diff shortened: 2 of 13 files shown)


Kobzol:
There is a regression with regard to `rustfmt` - the field block should be multiline only in structs, not in struct enum variants - currently my change applies to both of them. Is there a way to do this declaratively or do I have to use `Block.computeSpacing`?

vlad20012:
Looks like `Block.computeSpacing` is the only way, To be honest, I'm not familiar with the formatter and it is mostly unmaintained.

Kobzol:
Well, I tried something like
```kotlin
elementType1 == LBRACE && child2.node?.treeParent?.elementType == BLOCK_FIELDS && psi2.parent?.parent is RsStructItem ->
       return createSpacing(1, 1, 1, true, 0)
```
in `computeSpacing` and it sort of works, but it breaks some subtle conditions (like trailing commas), because the default spacing builder is not used if I override it here. As a last effort, I tried to reuse the output from the spacing builder in `computeSpacing` and just override it's newline count, but even then it still breaks for trailing commas.
```kotlin
val spacing = ctx.spacingBuilder.getSpacing(block, child1, child2)
  when (spacing) {
    is SpacingImpl -> return createSpacing(spacing.minSpaces, spacing.maxSpaces, 1, spacing.shouldKeepLineFeeds(), spacing.keepBlankLines)
    is DependantSpacingImpl -> return createSpacing(spacing.minSpaces, spacing.maxSpaces, 1, spacing.shouldKeepLineFeeds(), spacing.keepBlankLines)
  }
```

I didn't find how I could specify this grandparent relationship in the spacing builder. If it's not possible to do this cleanly, then maybe we should scratch this. Enabling it for enums would not play well with `rustfmt` (although the current implementation also formats differently than `rustfmt`).

Kobzol:
It turned out that the default formatting was quite different from `rustfmt`, so I changed it slightly to match `rustfmt` more. Now trailing commas are also used for all struct block fields.

src/main/kotlin/org/rust/ide/formatter/processors/RsTrailingCommaFormatProcessor.kt
```diff
@@ -84,7 +86,12 @@ fun CommaList.addTrailingCommaForElement(list: PsiElement): Boolean {
     )
     if (!trailingSpace.contains('\n')) return false
 
-    if (list.firstChild.getNextNonCommentSibling() == lastElement) return false
+    if (list.firstChild.getNextNonCommentSibling() == lastElement) {
+        // allow trailing comma only for struct block fields with a single item
```
vlad20012, line 90: What means "with a single item"? I don't get it :(

src/test/kotlin/org/rust/ide/refactoring/RsExtractEnumVariantTest.kt
```diff
@@ -42,7 +42,11 @@ class RsExtractEnumVariantTest : RsTestBase() {
             V2
         }
     """, """
-        struct /*caret*/V1 { pub a: i32, pub b: bool, pub c: String }
+        struct /*caret*/V1 {
+            pub a: i32,
+            pub b: bool,
+            pub c: String
```
vlad20012, line 48: Hmm, where is trailing comma here? Is it the same issue that `RsTrailingCommaFormatProcessor` doesn't invoke for some reason?

src/main/kotlin/org/rust/ide/formatter/processors/RsTrailingCommaFormatProcessor.kt
Kobzol: Sorry for the confusion. `RsTrailingCommaFormatProcessor` does not add trailing commas to comma lists with a single element. So the plugin would format the following struct like this:
```rust
struct S {
    a: u32
}
```
while `rustfmt` formats it like this:
```rust
struct S {
    a: u32,
}
```
so I allowed the trailing comma for a comma list with a single element, but only in structs with named fields.

To make matters more complicated, the actual behaviour of `rustfmt` is this: it adds a trailing comma for fields only if the fields are multi-line. Fields are multi-line if:
1) They are inside struct with named fields ("classic" structs)
2) They are inside tuple struct or inside an enum variant and they are too long (but this happens rarely, the fields basically need to be longer than the maximum line length).

So this condition covers case 1), which is most common.

src/test/kotlin/org/rust/ide/refactoring/RsExtractEnumVariantTest.kt
Kobzol: Yes, I tried it now and the comma processor doesn't get invoked.

vlad20012: Oh

Kobzol:
@vlad20012 do you want to merge this PR? It should bring formatting closer to `rustfmt`, although it is not perfect.

vlad20012:
It looks like `RsTrailingCommaFormatProcessor` is never called after intentions/quick fixes. I guess it's by design.

I am very sorry for such a long responce. For a long time, I was not familiar with the formatting subsystem, and I did not understand how it should work. Now I seem to understand more or less.

bors try

Undin:
bors retry

PR #5493, opened 2020-06-03
GRAM: Fix parsing `crate` at the beginning of macro call as vis modifier

### LINKED ISSUE #5492: False positive "crate visibility modifier is experimental" for macro calls
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.3132-201
* **Rust toolchain version:** 1.43.1 (8d69840ab 2020-05-04) x86_64-unknown-linux-gnu
* **IDE name and version:** IntelliJ IDEA 2020.2 EAP Ultimate Edition (IU-202.4357.23)
* **Operating system:** Linux 5.6.15-arch1-1

## Problem description
image

## Steps to reproduce
```rust
#[macro_export]
macro_rules! hello_macro {
    () => {
        println!("Hello!");
    };
}

fn main() {
    crate::hello_macro!();
}
```

dima74:
Fixes #5492


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -49,7 +49,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
     override fun getType() = Type
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
-        private const val STUB_VERSION = 198
+        private const val STUB_VERSION = 199
```
Undin, line 52: Some time ago we introduced two additional version fields: `org.rust.lang.core.parser.RustParserDefinition#LEXER_VERSION` and `org.rust.lang.core.parser.RustParserDefinition#PARSER_VERSION` for lexer and parser changes respectively. So in this case, you need to change `PARSER_VERSION`. Note, `PARSER_VERSION` depends on `LEXER_VERSION` and `STUB_VERSION` depends on `PARSER_VERSION`, so incrementing `PARSER_VERSION` also rebuilds stubs

dima74: @Undin Thanks!

PR #5495, opened 2020-06-03
Add postfix templates for Result/Option wrapping

### LINKED ISSUE #5494: Add postfix templates for Result/Option wrapping
When you write functions that return Result or Option, it's very common to write out full expression first and then wrap it in `Ok($expr)` and alike. It would be great to have postfix templates for these operations.

Specifically:
 - `$expr.ok<caret>` expand to `Ok($expr)`
 - `$expr.err<caret>` expand to `Err($expr)`
 - `$expr.some<caret>` expand to `Some($expr)`
 - ~~`$expr.none<caret>` expand to `None($expr)`~~

panstromek:
Since all those are trivial, I basically just copy-paste-edited them from `dbg` postfix template, its tests and descriptions. If you think it'd be good to change something or add some specific tests, let me know. This is very lazy job, but it does the job 😄

fixes #5494 

Note: I signed the CLA in the past to contribute to Vue plugin, so I just added my name here.

### DIFF
--- a/src/main/resources/postfixTemplates/ErrPostfixTemplate/after.rs.template
+++ b/src/main/resources/postfixTemplates/ErrPostfixTemplate/after.rs.template
@@ -1,3 +1,3 @@
-fn foo(slice: &[i32]) {
-    let first = Err(slice[0])
+fn foo(number: i32) -> Result<(), i32> {
+    Err(number)
 }
--- a/src/main/resources/postfixTemplates/ErrPostfixTemplate/after.rs.template
+++ b/src/main/resources/postfixTemplates/ErrPostfixTemplate/after.rs.template
@@ -1,0 +1,3 @@
+fn foo(slice: &[i32]) {
+    let first = Err(slice[0])
+}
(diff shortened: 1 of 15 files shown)


src/main/resources/postfixTemplates/ErrPostfixTemplate/after.rs.template
```diff
@@ -0,0 +1,3 @@
+fn foo(slice: &[i32]) {
```
Kobzol, line 1: nit: these templates (and some tests) do not compile. You can change the return type to `Result`/`Option` so that there is only the `Err(...)` expression inside the function.

panstromek: Fair point, let me change that 😉

panstromek: I made them simpler - not using slice, just a number and make the function return. I hope it's fine now.

panstromek:
One check failed in some unrelated test on "Project already disposed".  It seems like it has nothing to do with these changes or did I miss something?

Undin:
@panstromek Thanks!

PR #5496, opened 2020-06-03
MACRO: enable new macro expansion engine in nightly builds

Undin:
It should affect only users with nightly builds

See #3628 for more details about new macro expansion engine


vlad20012:
We need more bug reports!

We still have some unexplained bugs in the macro expansion engine, but we have too few bug reports or exception reports related to it. So I think it's enough stable to enable it by default on the nightly at least. If the number of bug reports increases, it's good anyway because it will help us to fix them. So,

bors r+

PR #5497, opened 2020-06-04
DOC: Add github actions badge to readme

dima74:
Adds badge for GitHub action checks.

Also update links to TeamCity build badges. Actually I am not sure about TeamCity badges, probably we should remove them, because most TeamCity builds are paused now? And probably some other TeamCity build should be paused, e.g. [Upload Nightly 193](https://teamcity.jetbrains.com/viewType.html?buildTypeId=IntellijIdeaPlugins_Rust_Toml_UploadNightly193)


dima74:
@Undin ok, so I removed badge for `check` workflow, and replaced TeamCity badges with badges for `rust release` and `rust nightly` workflows

dima74:
bors r=undin

PR #5500, opened 2020-06-04
MACRO: more consistency checks

vlad20012:
We still have some unexplained bugs in the macro expansion engine. E.g. false-positive "Use of moved value" for i32 - this can be explained by unexplained disappearing of macro expansions =/
I failed to investigate it, so I just add more asserts, more logs and do more attempts to recover things if something got wrong. Maybe we'll catch the reason later.


src/main/kotlin/org/rust/lang/core/macros/MacroExpansionFileSystem.kt
```diff
@@ -236,7 +240,9 @@ class MacroExpansionFileSystem : NewVirtualFileSystem() {
                 if (override) {
                     children.removeIf { it.name == item.name }
                 } else {
-                    check(children.find { it.name == item.name } == null) { "File `${item.name}` already exists" }
+                    if(children.any { it.name == item.name }) {
```
Undin, line 243:
```suggestion
                    if (children.any { it.name == item.name }) {
```

Issue #5301, opened 2020-04-29
Intellij rust plugin does not give give hint to create a function 


Haider8:
## Environment

* **IntelliJ Rust plugin version:** 0.2.121.3059-201-stable
* **Rust toolchain version:** rustc 1.43.0 (4fb7144ed 2020-04-20)
* **IDE name and version:**  CLion-2020.1
* **Operating system:** Linux (16.04.1-Ubuntu)

## Problem description
On some other Intellij IDEs when you call a function that doesn't exist already it gives you a hint to create that function either by pressing `alt+enter` or the yellow bulb appears with the suggestion to create that function. But this is not happening in CLion along with rust plugin.

## Steps to reproduce
Call a function that doesn't exist and then hit `alt+enter`. Nothing will happen, no hint will appear.

Kobzol:
There are already several similar existing issues: https://github.com/intellij-rust/intellij-rust/issues/3581 https://github.com/intellij-rust/intellij-rust/issues/3904. This intention was intentionally not implemented yet because it would present a lot of false positives, at least until the new macro expansion engine is enabled by default (https://github.com/intellij-rust/intellij-rust/issues/3628), as explained here: https://github.com/intellij-rust/intellij-rust/issues/3807#issuecomment-493417580.

@Undin is this still valid, i.e. does this intention wait for the new macro expansion engine? Maybe we can only enable the intention if the new macro expansion engine is active?

Undin:
@Kobzol I think we can implement it as quick fix for unresolved reference inspection even currently.
But right now unresolved inspection is disabled by default for paths that cannot be imported via `Auto import` quick fix to avoid false-positive errors.
So if you want, you can try to implement it (and it would be very cool!). But most of the users won't be able to use it right now.

> Maybe we can only enable the intention if the new macro expansion engine is active?

I think it's more about enabling unresolved reference inspection in the corresponding cases. It requires investigation: how many false-positive errors we get even with new macro expansion engine? Probably, still a lot because of procedural macros but we can take it into account as well.
Anyway, this task doesn't depend on the desired quick fix and can be solved separately

Kobzol:
Okay, then I suggest that we add a new quick fix to unresolved paths that look like a function call. We can also add support for method calls, but for these I think that we would need an option in the plugin settings to turn it off, because I expect that method calls will have more false positives than free-standing functions (because of procedural macros and `derive`s on structs and enums).

[...]

Issue #3904, opened 2019-05-25
Missing 'Create method'


svenschmidt75:
## Environment

* **Intellij-Rust plugin version:** 0.2.98.2125-192
* **Rust toolchain version:** stable-x86_64-unknown-linux-gnu (default)
rustc 1.35.0 (3c235d560 2019-05-20)

* **IDE name and version:** CLion 2019.2 EAP
* **Operating system:** PopOs 18.04

## Problem description
Say you have this code

```rust
fn load_image_file(path: &str) -> Result<Vec<Image>> {
    let mut file = File::open(&Path::new(path))?;
    let mut buffer = Vec::<u8>::new();
    let number = file.read_to_end(&mut buffer);
    let images = parse_image_data(buffer)?;
    Ok(images)
}
```

Place the cursor on parse_image_data on the 2nd last line.
There is no way to create a method stub. I think it would be
helpful if the plugin could generate code
```rust
fn parse_image_data(buffer: &mut Vec<u8>) -> Result<Vec<Image>> { }
```

fan-tom:
I often encounter the same pattern when develop top-to-bottom and just invent method/function on the fly and then want IDE to generate that missing functions for me with signatures based on their calls

Issue #3581, opened 2019-03-24
Add a create function intention


SomeoneToIgnore:
Similar to Kotlin and Java intentions, I would like to be able to define a non-existing function and let the plugin to generate it in the same file with the correct types.

Kotlin example: 
<img width="324" alt="image" src="https://user-images.githubusercontent.com/2690773/54885053-b28ee380-4e80-11e9-94ee-3210b63c8dfc.png">

PR #5505, opened 2020-06-04
INT: add intention to create a function from an unresolved function call

Kobzol:
This PR adds a quick fix to `RsUnresolvedReferenceInspection` which creates a new function for unresolved function call. I took inspiration from a similar quick fix present in the IntelliJ Kotlin plugin.

convert2

It also copies over generic parameters of the call parameters and their bounds:
convert1

Currently, the fix is offered only on function calls (not method calls) to avoid potential false positives for methods generated by procedural macros. The fix is also only offered if the parent module of the path exists (i.e. for `a::b::function`, `a::b` must resolve to an existing module) and is part of the workspace (so that it won't try to create a function in `stdlib`).

There are several things left to resolve:
1) When the function is created in a different module, it will need to be visible from the caller location and also the types used in the function signature will have to be imported (this may not always be possible though, if some of the types are private). Should we support this scenario? The Kotlin plugin doesn't support it.
2) When the function call parameters contain types with generic/const parameters, they will need to be copied to the created function, along with their bounds and `where` clauses. Right now I copy generic and const types present in the parameters along with their `where` clauses, but I do not copy lifetimes. The Kotlin plugin only copies generic parameters and doesn't copy `where` clauses.
I was able to reuse most of the code from `RsExtractEnumVariantProcessor` which basically takes a generic declaration and a set of types (or type references) and returns a subset of generic parameters/const parameters/lifetimes that are present in these types. This subset should be copied over to the new function. The naming here is.. not nice (`GenericParameters`). This code should probably be refactored and extracted into a separate module, as it's needed both here and in `RsExtractEnumVariantProcessor`. Where should I eventually put it?
3) Where to create the function? Right now if the path resolves to the same module as the caller, I find some `RsItemElement` parent of the caller and put the function before it. If it resolves to a different module, I put the function at the beginning of that module. Or should we allow the user to select where to place the function, similar to `RsIntroduceConstantHandler`?
4) Should we also enable the fix for method calls?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3581
Fixes: https://github.com/intellij-rust/intellij-rust/issues/3904
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5301

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
@@ -1,0 +1,127 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.openapiext.isUnitTestMode
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.parentOfType
+import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.ide.presentation.renderInsertionSafe
+import org.rust.ide.utils.GenericConstraints
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.types.type
+import org.rust.openapiext.buildAndRunTemplate
+import org.rust.openapiext.createSmartPointer
+
+class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionIntention.Context>() {
+    override fun getFamilyName() = text
+
+    sealed class Context {
+        data class Function(val callExpr: RsCallExpr, val name: String, val module: RsMod) : Context()
+    }
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val path = element.parentOfType<RsPath>()
+        val functionCall = path?.parentOfType<RsCallExpr>() ?: return null
+        if (!path.isUnresolved) return null
+
+        if (!functionCall.expr.isAncestorOf(path)) return null
+
+        val module = getTargetModuleForFunction(path) ?: return null
+        val name = path.referenceName
+        text = "Create function `$name`"
+        return Context.Function(functionCall, name, module)
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        when (ctx) {
+            is Context.Function -> createFunction(project, editor, ctx)
+        }
+    }
+}
+
+private fun createFunction(project: Project, editor: Editor, ctx: CreateFunctionIntention.Context.Function) {
+    val functionName = ctx.name
+    val callExpr = ctx.callExpr
+
+    val factory = RsPsiFactory(project)
+    val config = getFunctionConfig(callExpr)
+
+    val genericParams = config.genericConstraints.buildTypeParameters()
+    val params = config.parameters.joinToString(", ")
+    val whereClause = config.genericConstraints.buildWhereClause()
+
+    val vis = if (callExpr.containingMod != ctx.module) "pub(crate) " else ""
+    val function = factory.tryCreateFunction("${vis}fn $functionName$genericParams($params)$whereClause {\n    unimplemented!()\n}")
+        ?: return
+
+    val sourceFunction = callExpr.parentOfType<RsFunction>() ?: return
+    val inserted = insertFunction(ctx.module, sourceFunction, function)
+
+    if (ctx.module.containingFile == callExpr.containingFile) {
+        val toBeReplaced = inserted.valueParameters.flatMap { listOfNotNull(it.pat, it.typeReference) } +
+            listOfNotNull(inserted.block?.expr)
+        editor.buildAndRunTemplate(inserted, toBeReplaced.map { it.createSmartPointer() })
+    } else {
+        // template builder cannot be used for a different file
+        inserted.navigate(true)
+    }
+}
+
+private fun getTargetModuleForFunction(path: RsPath): RsMod? {
+    if (path.qualifier != null) {
+        val mod = path.qualifier?.reference?.resolve() as? RsMod
+        if (mod?.containingCargoPackage?.origin != PackageOrigin.WORKSPACE) return null
+        if (!isUnitTestMode && !mod.isWritable) return null
+        return mod
+    }
+    return path.containingMod
+}
+
+private data class FunctionConfig(val parameters: List<String>,
+                                  val genericConstraints: GenericConstraints)
+
+private fun getFunctionConfig(callExpr: RsCallExpr): FunctionConfig {
+    val parameters = callExpr.valueArgumentList.exprList.mapIndexed { index, expr ->
+        "p$index: ${expr.type.renderInsertionSafe(useAliasNames = true)}"
+    }
+
+    val genericConstraints = GenericConstraints.create(callExpr)
+        .filterByTypes(callExpr.valueArgumentList.exprList.map { it.type })
+
+    return FunctionConfig(parameters, genericConstraints)
+}
+
+private fun insertFunction(
+    targetModule: RsMod,
+    sourceFunction: RsFunction,
+    function: RsFunction
+): RsFunction {
+    if (targetModule == sourceFunction.containingMod) {
+        val impl: RsTraitOrImpl? = when (val owner = sourceFunction.owner) {
+            is RsAbstractableOwner.Trait -> owner.trait
+            is RsAbstractableOwner.Impl -> owner.impl
+            else -> null
+        }
+
+        val target = impl ?: sourceFunction
+        return target.parent.addAfter(function, target) as RsFunction
+    }
+
+    // add to the end of module/file
+    return if (targetModule is RsModItem) {
+        targetModule.addBefore(function, targetModule.rbrace)
+    } else {
+        if (targetModule.lastChild == null) {
+            targetModule.add(function)
+        } else {
+            targetModule.addAfter(function, targetModule.lastChild)
+        }
+    } as RsFunction
+}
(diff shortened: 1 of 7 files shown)


Undin:
@Kobzol FYI: "Fixes #issue1 #issue2 ..." will close only issue1. So we have to write "Fixes" word for each issue

Undin:
You can check it in `Linked issues` section of issue

Kobzol:
Thanks, I didn't notice. Actually I'm not sure if this would close the last issue, since it mentions a method and right now these are not supported :)

Kobzol:
I use `Editor.buildAndRunTemplate` to change the parameters of the newly created function. However, if the function is created in a different file (since the fix now supports creating functions in different modules), the template builder fails (it marks text in the file that contains the PSI that I want to modify). Is there a way to use the template builder for a different file/editor or should I just navigate to the function and do not use the template in such a case?

src/main/kotlin/org/rust/ide/inspections/RsUnresolvedReferenceInspection.kt
```diff
@@ -66,39 +65,60 @@ class RsUnresolvedReferenceInspection : RsLocalInspectionTool() {
...
         }
 
     private fun RsProblemsHolder.registerProblem(
         element: RsReferenceElement,
-        context: AutoImportFix.Context?
+        fixes: List<LocalQuickFix>
     ) {
-        val candidates = context?.candidates
-        if (candidates.isNullOrEmpty() && ignoreWithoutQuickFix) return
+        if (fixes.isEmpty() && ignoreWithoutQuickFix) return
```
vlad20012, line 79: `ignoreWithoutQuickFix` was invented to reduce the number of false-positives. We guess that if there is an item that can be auto-imported, then it's not a false-positive. But, these heuristics lead to false-positives anyway... We don't want to make more. 

False-positives caused poor name resolution & macro expansion implementations (we're planning a new name resolution design now).

So, I suggest not to change the current heuristics behavior but make an intention instead of quickfix! An intention applied to an unresolved reference.

Kobzol: So the unresolved function call should not be annotated as an error, and there should instead be an intention that will be offered on unresolved function calls? And the code of the intention that checks if the function call is unresolved will be shared with this inspection?

vlad20012: Yep. But I'm not sure there are a lot of code to share. `path.reference?.multiResolve()?.isEmpty() != false`

Kobzol: Ok, I changed it to an intention. Now that it's an intention, I guess that we can also support method calls?

vlad20012: Yes!

vlad20012:
1. About declaring a function in a different module. Yes, we can do this, but I don't think we should much care about types and visibility. 
2. Somewhere in `org.rust.ide.utils`, I guess. E.g. `ExtractUtil`
3. Hmm, why before the function. Looks like Kotlin places it after. The behavior with `RsItemElement` seems ok except the case with an associated function inside `impl`. In this case, a function should be placed after the impl. In the case of different modules... I dont know =D I don't think it's very important.
4. :+1:

vlad20012:
One more point about different modules - we should check that the modules is writable. We usually check that `containingCargoPackage.origin == PackageOrigin.WORKSPACE`. Also can check that psi.isWritable

Kobzol:
The `test create function in an existing file` test was failing, because it seems to check the contents of the file containing the caret, but I need to check the contents of the second file after performing the intention. It's caused by a different behaviour of the `doAvailableTestWithFileTree` function. I added an additional test function.

Kobzol:
@vlad20012 for creating methods it would be nice if I could find existing `impl` blocks of `structs`/`enums`/`traits`. I know that this was already discussed in some other intentions, is there some API to find existing blocks or does it not exist yet?

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -0,0 +1,128 @@
...
+class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionIntention.Context>() {
+    override fun getFamilyName() = text
+
+    sealed class Context {
+        data class Function(val callExpr: RsCallExpr, val name: String, val module: RsMod) : Context()
+    }
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val path = element.parentOfType<RsPath>()
+        val functionCall = path?.parentOfType<RsCallExpr>()
```
vlad20012, line 31:
```suggestion
        val functionCall = path?.parentOfType<RsCallExpr>() ?: return null
```

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -0,0 +1,128 @@
...
+        if (functionCall != null) {
```
vlad20012, line 32: Let's also check that the path is an actual path of the call (to make it unavailable in a case like `foo(bar::baz/*caret*/))`.
This can be checked by `functionCall.expr.isAncestorOf(path)`. 
Let's also add a test for this case

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -0,0 +1,128 @@
...
+            is RsAbstractableOwner.Trait -> owner.trait
+            is RsAbstractableOwner.Impl -> owner.impl
+            else -> null
+        }
+
+        val target = impl ?: sourceFunction
+        return target.parent.addAfter(function, target) as RsFunction
+    }
+
+    // add to the beginning of module/file
```
vlad20012, line 118: Maybe it's better to insert it to the end? Or skip use declarations at least

src/main/kotlin/org/rust/ide/utils/ExtractUtils.kt
```diff
@@ -0,0 +1,290 @@
...
+
+
+/**
+ * Holds type parameters, lifetimes, const parameters and where clauses.
+ * It serves as a combination of several `RsGenericDeclaration`s.
+ *
+ * Can be filtered by a set of types/type references to only returns parameters/constraints that are needed by these
+ * given types/type references.
+ */
+data class GenericConstraints(
```
vlad20012, line 35: Hmm, cool util :+1:

src/main/kotlin/org/rust/ide/utils/ExtractUtils.kt
```diff
@@ -0,0 +1,290 @@
...
+         * their constraints.
+         */
+        fun create(context: PsiElement): GenericConstraints {
+            val typeParameters = mutableSetOf<RsTypeParameter>()
+            val lifetimes = mutableSetOf<RsLifetimeParameter>()
+            val constParameters = mutableSetOf<RsConstParameter>()
+            val whereClauses = mutableListOf<RsWhereClause>()
+
+            var genericDecl: RsGenericDeclaration? = (context as? RsGenericDeclaration) ?: context.parentOfType()
+            while (genericDecl != null) {
```
vlad20012, line 136: In Rust, parent generics usually can't be used in nested items. For example, this is incorrect code:
```rust
fn foo<T>() {
    fn bar() -> T { unimplemented!() }
}
```
The 2 exceptions are trait end impl. So it's enough to look at `RsAbstrctable.owner`.
But I'm not sure it leads to any bugs. Maybe in a case like this:
```rust
fn foo<T>() where T: Foo {
    fn bar<T>() where T: Bar { unimplemented!() }
}
```

vlad20012 (approved):
LGTM except minor comment.

I'm sorry for the very long response. I hope I didn't hurt you a lot (

vlad20012:
> or creating methods it would be nice if I could find existing impl blocks of structs/enums/traits. I know that this was already discussed in some other intentions, is there some API to find existing blocks or does it not exist yet?

Well, I'm not sure, I didn't review such PRs or I didn't remember that. In theory, this can be done by `ImplLookup.findImplsAndTraits().filterIsInstance<TraitImplSource.ExplicitImpl>().filter { it.isInherent }`. But there can be multiple suitable impls

src/main/kotlin/org/rust/ide/utils/ExtractUtils.kt
Kobzol: Thanky, I added a check and a test for this.

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    // add to the end of module/file
+    return if (targetModule is RsModItem) {
+        targetModule.addBefore(function, targetModule.rbrace)
+    } else {
+        if (targetModule.lastChild == null) {
+            targetModule.add(function)
+        } else {
+            targetModule.addBefore(function, targetModule.lastChild)
```
vlad20012, line 124: Hmm, before last child. This means the function will be the penultimate one, isn't it?

Kobzol: Good catch, thanks. One of the tests was incorrect, so it didn't catch this.

Kobzol:
While going through https://github.com/intellij-rust/intellij-rust/issues/3904, I realized that this intention could also pre-fill the return type, based on the context in which the (unresolved) path was used. For example:
```rust
fn foo() {
   let x: u32 = bar();
}

fn bar() -> u32 {
    unimplemented!()
}
```
We would have to hardcode some common scenarios (`let x = <unresolved>`, `call(<unresolved>)`), because type inference doesn't provide the type if the path is unresolved, but I think that would be fine. I'll send it in another PR, as this one is already complex enough.

vlad20012:
> because type inference doesn't provide the expected type if the path is unresolved, but I think that would be fine.

I think it can. `expected type` functionality just isn't embraced well and has... roughness.

vlad20012:
Thanks!
bors r+

PR #5515, opened 2020-06-05
FIX: disable downgrade module and promote to module refactorings for unavailable files in editor

Kobzol:
Both `RsDowngradeModuleToFile` and `RsPromoteModuleToDirectoryAction` refactorings were available on all Rust PsiElements all of the time, without checking if their corresponding file is valid for the refactoring. Invoking such refactoring could have a nasty effect, for example it could delete a whole (non-empty) source directory.

I modified them to only be available if the corresponding file in the editor is valid for the refactoring. We can also ban them altogether from the editor and only enable it from the Project tree.


vlad20012:
Nice catch! Thanks!
bors r+

Issue #5440, opened 2020-05-25
Cargo.toml improvements


Kobzol:
Meta-issue about various `Cargo.toml`-related improvements.

- [ ] Parse crate information from the local index (https://github.com/intellij-rust/intellij-rust/issues/2757)
- [ ] Provide completion for features of a crate (https://github.com/intellij-rust/intellij-rust/issues/4693 https://github.com/intellij-rust/intellij-rust/issues/4455)
- [ ] Provide completion for dependency attributes (https://github.com/intellij-rust/intellij-rust/issues/4455)
- [ ] Provide completion for local dependency paths (https://github.com/intellij-rust/intellij-rust/issues/4455 https://github.com/intellij-rust/intellij-rust/issues/4144)
- [ ] Check that crates in `[dependencies]` exist and have a valid version (https://github.com/intellij-rust/intellij-rust/issues/4456 https://github.com/intellij-rust/intellij-rust/issues/1778 https://github.com/intellij-rust/intellij-rust/issues/3655)
- [ ] Expand dependency version (https://github.com/intellij-rust/intellij-rust/issues/4869)
- [ ] Avoid macro indexing during benign `Cargo.toml` changes (https://github.com/intellij-rust/intellij-rust/issues/5062)
- [ ] Warn about missing dependencies and add dependency to `Cargo.toml` if it's missing (https://github.com/intellij-rust/intellij-rust/issues/4092)

Kobzol:
The speed is indeed problematic and it's blocking progress a bit I think. https://github.com/intellij-rust/intellij-rust/pull/4873 there are problems with parsing the local index however. I would implement new stuff on top of the API, but make it easy to change the crate resolver in the future. I introduced such abstraction in https://github.com/intellij-rust/intellij-rust/pull/5439.

PR #5517, opened 2020-06-06
TOML: Fix Cargo.toml dependencies docs urls

### LINKED ISSUE #5411: Dash in crate name sometimes causes wrong docs.rs link in Cargo.toml
## Environment

0.2.122.3105-201

* **Rust toolchain version:**
1.42 stable msvc

* **IDE name and version:**
CLion 2020.1.1
Build #CL-201.7223.86, built on April 29, 2020
Non-Bundled Plugins: GLSL, Key Promoter X, com.jetbrains.plugins.ini4idea, com.wakatime.intellij.plugin, org.toml.lang, org.rust.lang

* **Operating system:**
Windows 10

## Problem description

When I click the imgui-winit-support crate link on the left in the Cargo.toml page, it links me to 
`https://docs.rs/imgui-winit-support/0.3.1/imgui-winit-support` which gives 404.

(the right URL is `https://docs.rs/imgui-winit-support/0.3.1/imgui_winit_support/`)

image

However it's important to note that `imgui-wgpu` works.  

I also suggest pointing to `https://docs.rs/imgui-winit-support/0.3.1/` instead, which auto redirects to the right URL.

## Steps to reproduce

Add `imgui-winit-support = "0.3.1"` in any Cargo.toml and try to click the crates icon on the left

### LINKED ISSUE #5467: Cargo.toml dependencies versions for docs urls are different from those Cargo uses
There is a difference between how Cargo and Docs.rs treat versions of dependencies. This is understandable because of the different purposes of version in config file and that one for docs. However, when a user want to see a documentation for specific crate in dependencies, it will be reasonable to show him docs for the version Cargo downloads and then compiler uses.

Two cases that are different from what is expectable:
1.  Version without any operators ⇒ caret requirement
   Docs should be latest semver-compatable version, instead docs.rs treats that as exact. Can be fixed by adding caret to the version in url.
   image
2. Empty version ⇒ wildcard requirement
   Docs should be latest version, instead version is missed in the url. Can be fixed by changing to * (star/wildcard) version in the url.
   image

I found three different implementations of docs.rs url generation for different places in the editor: `CargoCrateDocLineMarkerProvider` for Cargo.toml, `RsDocumentationProvider` for quick help and `RsCrateDocLineMarkerProvider` for `extern crate` expressions.

Source: [Specifying Dependencies - The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

Happy to help and ready to fix this if it is reasonable.

Related to #5440

avrong:
Fixes #5467. Fixes #5411.
Manages version modification to be compatable with docs.rs and fixes urls for crates with dashes.

image

Not sure how incorrect versions should be managed, with other issues of #5440 this could be improved.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/CargoCrateDocLineMarkerProvider.kt
+++ b/toml/src/main/kotlin/org/rust/toml/CargoCrateDocLineMarkerProvider.kt
@@ -49,12 +49,18 @@ class CargoCrateDocLineMarkerProvider : LineMarkerProvider {
     }
 
     private fun genLineMarkerInfo(anchor: PsiElement, name: String, version: String): LineMarkerInfo<PsiElement> {
+        val urlVersion = when {
+            version.isEmpty() -> "*"
+            version.first().isDigit() -> "^${version}"
+            else -> version
+        }
+
         return LineMarkerInfo(
             anchor,
             anchor.textRange,
             RsIcons.DOCS_MARK,
-            { "Open documentation for `$name@$version`" },
-            { _, _ -> BrowserUtil.browse("https://docs.rs/$name/$version/$name") },
+            { "Open documentation for `$name@$urlVersion`" },
+            { _, _ -> BrowserUtil.browse("https://docs.rs/$name/$urlVersion") },
             GutterIconRenderer.Alignment.LEFT)
 
     }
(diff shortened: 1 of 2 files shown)


vlad20012:
Let's add tests for these cases

vlad20012:
Some tests are failed

Issue #5520, opened 2020-06-06
Add unsafe to function quick fix does nothing


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.124.3142-201-nightly
* **Rust toolchain version:** 1.45.0-nightly (a74d1862d 2020-05-14) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2 EAP Ultimate Edition (IU-202.5103.13)
* **Operating system:** macOS 10.13.6

## Problem description
![2020-06-06 20 53 40](https://user-images.githubusercontent.com/2539310/83951075-d21cec80-a837-11ea-80fe-3e4a2f7c8b94.gif)

## Steps to reproduce
```rust
const INITIAL_TEXT: &[u8] = b"Hello, World!";

#[no_mangle]
pub extern "C" fn _start() -> ! {
    let vga_buffer = 0xb8000 as *mut u8;

    for (index, value) in INITIAL_TEXT.iter().enumerate() {
        *vga_buffer.offset(index as isize * 2) = *value;
    }

    loop {}
}
```

PR #5521, opened 2020-06-06
FIX: add unsafe now also works inside loops and conditions

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5520


vlad20012:
:+1: 
bors r+

Issue #5109, opened 2020-03-17
False Negative: Parameter is Never Used when destructuring in parameters


OptimisticPeach:
## Environment

* **IntelliJ Rust plugin version:** 0.2.118.2171-193
* **Rust toolchain version:** rustc 1.43.0-nightly (564758c4c 2020-03-08)
* **IDE name and version:** CLion 2019.3.4
* **Operating system:** Windows

## Problem description
When destructuring in a function's arguments, I get a 'Parameter `ty` is never used'-type message, even though it is used. 

## Steps to reproduce
Here is the minimal example from exactly what my code has:
```rust
union BarVal {
    as_bool: bool
}

enum BarTy {
    Bool,
}

struct Bar {
    ty: BarTy,
    val: BarVal,
}

enum SimpleBar {
    Bool(bool)
}

unsafe fn foo(Bar {ty, val}: Bar) -> SimpleBar {
    match ty {
        BarTy::Bool => SimpleBar::Bool(val.as_bool)
    }
}
```
and here is the minimal example I've found to cause this:
```rust
struct Bar {
    ty: usize,
}

fn foo(Bar {ty}: Bar) -> usize {
    ty
}
```
This does not occur if `Bar` is a tuple struct instead of a named struct.

PR #5528, opened 2020-06-07
FIX: liveness analysis now takes into account destructured struct parameters

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5109

### DIFF
--- a/src/test/kotlin/org/rust/ide/inspections/RsLivenessInspectionTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/RsLivenessInspectionTest.kt
@@ -152,9 +152,9 @@ class RsLivenessInspectionTest : RsInspectionsTestBase(RsLivenessInspection::cla
 
     fun `test live variable match 3`() = checkByText("""
         enum E { A, B(i32) }
-        
+
         fn bar(_x: i32) {}
-        
+
         use E::*;
         fn foo(e: E) {
             let _ = match e {
@@ -164,7 +164,7 @@ class RsLivenessInspectionTest : RsInspectionsTestBase(RsLivenessInspection::cla
                     2
                 }
             };
-            
+
             let _ = match Config::default() {
                 Ok(cfg) => cfg,
                 Err(e) => {
@@ -239,7 +239,7 @@ class RsLivenessInspectionTest : RsInspectionsTestBase(RsLivenessInspection::cla
             let x = 42;
             noreturn(x);
         }
-        
+
         fn noreturn(_a: i32) -> ! { panic!() }
     """)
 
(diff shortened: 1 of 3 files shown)


Undin:
@Kobzol FYI if PR makes some user-visible changes and/or you want that our QA check it (usually if a change is user-visible, you want it), add it into `To test` project

Kobzol:
I think that I don't have permissions to add PRs to projects, there's no button on the Projects section in the right column.

Undin:
@Kobzol it should be fixed now. Could you check it, please?

Kobzol:
I just added the rest of my user-facing PRs to the `To test` project, so it is working :-) Thank you.

ortem:
@Kobzol Changes in our control-flow analysis sometimes lead to false-positives in real projects. So I usually run `RsRealProjectAnalysisTest` (at least `Cargo` and `tokio` but more is better) before merging such PR, like [here](https://github.com/intellij-rust/intellij-rust/pull/5383).

Could you please run some tests from `RsRealProjectAnalysisTest` twice: before your changes and after them, and check there are no new false-positives related to `RsBorrowCheckerInspection` and `RsLivenessInspection`?

Note that `checkWarnings` should be `true` (see `RsRealProjectAnalysisTest.kt:74`). Also, probably you will have to comment `testAssert { implsAndTraits.distinct().size == implsAndTraits.size }` on `ImplLookup.kt:338` because it leads to early termination of some real project tests (BTW @vlad20012 can we deal with it somehow?)

src/test/kotlin/org/rust/ide/inspections/RsLivenessInspectionTest.kt
```diff
@@ -489,9 +489,30 @@ class RsLivenessInspectionTest : RsInspectionsTestBase(RsLivenessInspection::cla
...
+            a: u32
+        }
+
+        fn foo() {
+            let S { a } = S { a: 0 };
+            a;
+        }
+    """)
+
+    fun `test destructured struct parameter`() = checkByText("""
```
ortem, line 509: Please also ensure that destructured struct parameters work fine in move analysis and add such tests to `RsBorrowCheckerMovesTest`

Kobzol:
I checked the real projects (except for `rustc`) and found no regressions. There are some false positives, but they were present even before this PR. And it fixed one false positive in `mysql_async`! :tada: 

This process should really be automated somehow :D Maybe we could write a Python/Bash script that takes two git revisions, runs the real projects tests on both of them and prints a diff of the emitted errors/warnings?

ortem:
@Kobzol Great, thank you!

Having some kind of automation in this process would be very nice. We discussed it for a long time ago but still haven't had time to implement it.

bors r+

PR #5532, opened 2020-06-08
INSP: navigate to created main function

Undin:
Improve quick fix introduced in #5481


Undin:
CC @Stzx @vlad20012

Stzx:
In fact, if the insertion symbol is inside the block, that would be great.
```rust
fn main() { /*caret*/ }
```

For the main function, we actually edit the method body more.

@Undin What do you think?

Undin:
@Stzx good idea. Thanks!

Issue #5518, opened 2020-06-06
Implement initial formatter for toml lang


Undin:
## Environment

* **IntelliJ Toml plugin version:** 0.2.124.168-193-nightly
* **IDE name and version:** IntelliJ IDEA 2020.2 EAP Ultimate Edition (IU-202.5103.13)
* **Operating system:** macOS 10.13.6

## Problem description
Currently, toml plugin doesn't provide any formatting.
Unfortunately, I don't know if the official formatting for language exists. But we can start at least with spacing between tokens. All examples from the official [repo](https://github.com/toml-lang/toml) contain spaces before/after `=`, after `,`, before/after `}`, `]`, `{` and `[` in objects and arrays, no spaces in dotted keys, etc. 

## Steps to reproduce
```toml
foo="bar"
object={a=1,b=2}
array=[1,2,3]
key1 . key2='value'
[ key . "key2" . key3 ]
```

Kobzol:
Maybe we can take some inspiration from [here](https://github.com/jumpscale7/python-pretty-toml) (discussed in the TOML repo: https://github.com/toml-lang/toml/issues/532).

PR #5536, opened 2020-06-09
TOML: initial formatter implementation

Kobzol:
This PR implements an initial formatter for TOML. I copied the Rust formatter and stripped it down to only keep things that were required to implement basic spacing formatting.

The [official repository](https://github.com/toml-lang/toml) does not mandate any formatting style and I think that it is a bit inconsistent (space after `[` in outer array, but not in nested array, while for objects it uses spaces after `{` even for nested objects).
There is https://github.com/jumpscale7/python-pretty-toml/blob/master/sample-prettified.toml, but it's also not very specified.

Right now there is only a very basic spacing implementation. Some things are probably OK to enforce (space after `,`, space before and after `=`), but for controversial topics like space after `[`, we should probably introduce an option in TOML code style settings.

After we resolve that, I think that this implementation is already quite useful and could be merged, but we should deal with indents somehow. Right now only `noneIndent` is used and so this:
```rust
multiline = [
      1,
      2
]
```
turns into
```rust
multiline = [
1,
2
]
```
which is not ideal, the explicit indent should be preserved.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5518

### DIFF
--- a/intellij-toml/src/main/kotlin/org/toml/ide/formatter/settings/TomlCodeStyleSettingsProvider.kt
+++ b/intellij-toml/src/main/kotlin/org/toml/ide/formatter/settings/TomlCodeStyleSettingsProvider.kt
@@ -1,0 +1,34 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml.ide.formatter.settings
+
+import com.intellij.application.options.CodeStyleAbstractConfigurable
+import com.intellij.application.options.TabbedLanguageCodeStylePanel
+import com.intellij.openapi.options.Configurable
+import com.intellij.psi.codeStyle.CodeStyleSettings
+import com.intellij.psi.codeStyle.CodeStyleSettingsProvider
+import com.intellij.psi.codeStyle.CustomCodeStyleSettings
+import org.toml.lang.TomlLanguage
+
+class TomlCodeStyleSettingsProvider : CodeStyleSettingsProvider() {
+    override fun createCustomSettings(settings: CodeStyleSettings): CustomCodeStyleSettings = TomlCodeStyleSettings(settings)
+
+    override fun getConfigurableDisplayName(): String = TomlLanguage.displayName
+
+    override fun createSettingsPage(settings: CodeStyleSettings, originalSettings: CodeStyleSettings): Configurable =
+        object : CodeStyleAbstractConfigurable(settings, originalSettings, configurableDisplayName) {
+            override fun createPanel(settings: CodeStyleSettings) = TomlCodeStyleMainPanel(currentSettings, settings)
+            override fun getHelpTopic() = null
+        }
+
+    private class TomlCodeStyleMainPanel(currentSettings: CodeStyleSettings, settings: CodeStyleSettings) :
+        TabbedLanguageCodeStylePanel(TomlLanguage, currentSettings, settings) {
+
+        override fun initTabs(settings: CodeStyleSettings) {
+            addIndentOptionsTab(settings)
+        }
+    }
+}
--- a/intellij-toml/src/main/kotlin/org/toml/ide/formatter/settings/TomlLanguageCodeStyleSettingsProvider.kt
+++ b/intellij-toml/src/main/kotlin/org/toml/ide/formatter/settings/TomlLanguageCodeStyleSettingsProvider.kt
@@ -1,0 +1,33 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml.ide.formatter.settings
+
+import com.intellij.application.options.IndentOptionsEditor
+import com.intellij.application.options.SmartIndentOptionsEditor
+import com.intellij.lang.Language
+import com.intellij.psi.codeStyle.LanguageCodeStyleSettingsProvider
+import org.toml.lang.TomlLanguage
+
+class TomlLanguageCodeStyleSettingsProvider : LanguageCodeStyleSettingsProvider() {
+    override fun getLanguage(): Language = TomlLanguage
+    override fun getCodeSample(settingsType: SettingsType): String =
+        when (settingsType) {
+            SettingsType.INDENT_SETTINGS -> INDENT_SAMPLE
+            else -> ""
+        }
+
+    override fun getIndentOptionsEditor(): IndentOptionsEditor? = SmartIndentOptionsEditor()
+}
+
+private fun sample(@org.intellij.lang.annotations.Language("TOML") code: String) = code.trim()
+
+private val INDENT_SAMPLE = sample("""
+[config]
+items = [
+    "foo",
+    "bar"
+]
+""")
--- a/intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
+++ b/intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
@@ -1,0 +1,26 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml
+
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import org.intellij.lang.annotations.Language
+
+abstract class TomlTestBase : BasePlatformTestCase() {
+    @Suppress("TestFunctionName")
+    protected fun InlineFile(@Language("TOML") text: String, name: String = "example.toml") {
+        myFixture.configureByText(name, text)
+    }
+
+    protected fun checkByText(
+        @Language("TOML") before: String,
+        @Language("TOML") after: String,
+        action: () -> Unit
+    ) {
+        InlineFile(before)
+        action()
+        myFixture.checkResult(after)
+    }
+}
--- a/intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlAutoIndentTest.kt
+++ b/intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlAutoIndentTest.kt
@@ -1,0 +1,23 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml.ide.formatter
+
+import org.toml.ide.typing.TomlTypingTestBase
+
+class TomlAutoIndentTest : TomlTypingTestBase() {
+    fun `test new array element`() = doTestByText("""
+        [key]
+        foo = [
+            "text",<caret>
+        ]
+    """, """
+        [key]
+        foo = [
+            "text",
+            <caret>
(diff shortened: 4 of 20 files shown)


intellij-toml/src/main/kotlin/org/toml/ide/formatter/settings/TomlLanguageCodeStyleSettingsProvider.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun customizeSettings(consumer: CodeStyleSettingsCustomizable, settingsType: SettingsType) {
+        @Suppress("NON_EXHAUSTIVE_WHEN")
+        when (settingsType) {
+            SettingsType.SPACING_SETTINGS -> {
+                consumer.showStandardOptions(CodeStyleSettingsCustomizable.SpacingOption.SPACE_AFTER_COMMA.name)
+            }
+        }
+    }
+
+    override fun getIndentOptionsEditor(): IndentOptionsEditor? = SmartIndentOptionsEditor()
```
Undin, line 33: If we decide not to provide indent settings in this PR (i.e. `org.toml.ide.formatter.settings.TomlCodeStyleSettingsProvider.TomlCodeStyleMainPanel#initTabs` initializes only space tab), it will be redudant

intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlFormatterTest.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml.ide.formatter
+
+class TomlFormatterTest : TomlFormatterTestBase() {
+    fun `test spacing`() = doTest()
```
Undin, line 9: Let's better inline test texts.
I find it much easier to have code example in test class directly instead of looking for before/after files in test data

Kobzol:
I inlined the formatter tests, added a typing test and a simple base class for `TOML` tests. I also added a simple indent implementation that handles array (and enter being pressed inside an array). I changed the settings to include the basic indent stuff like tabs/spaces and indent size.

Since objects (inline tables) have to be on a single line, I think that arrays are actually the only thing in TOML that require indent handling? In such case I think that the current implementation would be fine as a MVP.

In the TOML reference there are indented sub tables:
```toml
[servers]
  # Indentation (tabs and/or spaces) is allowed but not required
  [servers.alpha]
  ip = "10.0.0.1"
  dc = "eqdc10"

  [servers.beta]
  ip = "10.0.0.2"
  dc = "eqdc10"
```
I haven't seen this in the wild yet, but maybe we can include it as an option in the indent settings.

Should I still turn this as an experimental feature now that basic indenting works?

intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
```diff
@@ -0,0 +1,28 @@
...
+        @Language("Rust") after: String,
+        action: () -> Unit
+    ) {
+        InlineFile(before)
+        action()
+        myFixture.checkResult(replaceCaretMarker(after))
+    }
+}
+
+fun replaceCaretMarker(text: String): String = text.replace("/*caret*/", "<caret>")
```
Undin, line 28: Actually, this function doesn't make sense for TOML. `/*caret*/` is valid block comment in Rust language and it was introduced to avoid syntax errors in injected code.
But TOML langugage doesn't have block comments at all so I don't know how to avoid syntax error when we need caret symbol in TOML code

intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.toml
+
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import org.intellij.lang.annotations.Language
+
+abstract class TomlTestBase: BasePlatformTestCase() {
+    @Suppress("TestFunctionName")
+    protected fun InlineFile(@Language("TOML") text: String, name: String = "Cargo.toml") {
```
Undin, line 13:
```suggestion
    protected fun InlineFile(@Language("TOML") text: String, name: String = "example.toml") {
```

Just non to mislead anyone that these tests somehow related to Rust and Cargo

intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
```diff
@@ -0,0 +1,28 @@
...
+        myFixture.configureByText(name, replaceCaretMarker(text))
+    }
+
+    protected fun checkByText(
+        @Language("Rust") before: String,
+        @Language("Rust") after: String,
```
Undin, line 19: `Rust` -> `TOML`

intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlAutoIndentTest.kt
```diff
@@ -0,0 +1,23 @@
...
+    [key]
+    foo = [
+        "text",/*caret*/
+    ]
+    """, """
+    [key]
+    foo = [
+        "text",
+        /*caret*/
+    ]
```
Undin, line 21:
```suggestion
        [key]
        foo = [
            "text",/*caret*/
        ]
        """, """
        [key]
        foo = [
            "text",
            /*caret*/
        ]
```

intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlFormatterTestBase.kt
```diff
@@ -0,0 +1,25 @@
...
+    protected fun doTest(@Language("TOML") before: String, @Language("TOML") after: String) {
+        InlineFile(before.trim())
+        WriteCommandAction.runWriteCommandAction(project) {
+            val file = myFixture.file
+            CodeStyleManager.getInstance(project)
+                .reformatText(file, file.textRange.startOffset, file.textRange.endOffset)
+        }
+
+        myFixture.checkResult(after.trim())
+    }
```
Undin, line 24: Let's use `checkByText` here

intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlFormatterTest.kt
```diff
@@ -0,0 +1,36 @@
...
+[  config . subconfig ]
+    list  =   [  1,2,     3   ]
+key1 . key2='value'
+object={a=1,b=2}
+    """, """
+[config.subconfig]
+list = [1, 2, 3]
+key1.key2 = 'value'
+object = { a = 1, b = 2 }
+    """)
```
Undin, line 19:
```suggestion
    fun `test spacing`() = doTest("""
        [  config . subconfig ]
            list  =   [  1,2,     3   ]
        key1 . key2='value'
        object={a=1,b=2}
    """, """
        [config.subconfig]
        list = [1, 2, 3]
        key1.key2 = 'value'
        object = { a = 1, b = 2 }
    """)
```

I know that test will not pass without additional changes. Suggest using `trimIndent` for expected/actual texts in some base method. For example, in `checkByText`

intellij-toml/src/test/kotlin/org/toml/ide/typing/RsTypingTestBase.kt
```diff
@@ -0,0 +1,16 @@
...
+* Use of this source code is governed by the MIT license that can be
+* found in the LICENSE file.
+*/
+
+package org.toml.ide.typing
+
+import org.intellij.lang.annotations.Language
+import org.toml.TomlTestBase
+
+abstract class TomlTypingTestBase : TomlTestBase() {
```
Undin, line 11: Let's change file name to `TomlTypingTestBase`

intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
```diff
@@ -0,0 +1,28 @@
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml
+
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import org.intellij.lang.annotations.Language
+
+abstract class TomlTestBase: BasePlatformTestCase() {
```
Undin, line 11: nit: formatting

intellij-toml/src/test/kotlin/org/toml/TomlTestBase.kt
Kobzol: Thanks, that makes sense.

Kobzol:
Ouch, nasty copy paste errors. Thanks.

intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlFormatterTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+package org.toml.ide.formatter
+
+class TomlFormatterTest : TomlFormatterTestBase() {
+    fun `test spacing`() = doTest("""
+        [  config . subconfig ]
+            list  =   [  1,2,     3   ]
+        key1 . key2='value'
+        object={a=1,b=2}
+            """, """
```
Undin, line 14:
```suggestion
    """, """
```

intellij-toml/src/test/kotlin/org/toml/ide/formatter/TomlFormatterTest.kt
```diff
@@ -0,0 +1,36 @@
...
+    """)
+
+    fun `test indent array`() = doTest("""
+        [workspace]
+        members = [
+        "a.rs",
+                "b.rs",
+                    "c.rs"
+        ]
+            """, """
```
Undin, line 28: The same here

intellij-toml/src/main/kotlin/org/toml/ide/formatter/settings/TomlCodeStyleSettingsProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.application.options.CodeStyleAbstractConfigurable
+import com.intellij.application.options.TabbedLanguageCodeStylePanel
+import com.intellij.psi.codeStyle.CodeStyleSettings
+import com.intellij.psi.codeStyle.CodeStyleSettingsProvider
+import org.toml.lang.TomlLanguage
+
+class TomlCodeStyleSettingsProvider : CodeStyleSettingsProvider() {
+    override fun createCustomSettings(settings: CodeStyleSettings) = TomlCodeStyleSettings(settings)
+
+    override fun getConfigurableDisplayName() = TomlLanguage.displayName
```
Undin, line 17: nit: Could you specify return type explicitly?

Issue #5537, opened 2020-06-09
Broken navigation to some items in 2020.2


vlad20012:
- [ ] Navigation from module declaration to a file
```rust
mod foo/*caret*/; // Ctrl + B
```
In 2020.2 Ctrl + B on a module declaration invokes "Find Usages" instead of navigating to the file. But our find usages doesn't work in this case =/

- [ ] Navigation from extern crate declaration to a crate root file
```rust
extern crate foo/*caret*/; // Ctrl + B
```
In 2020.2 Ctrl + B on an extern crate declaration invokes "Find Usages" instead of navigating to the file. But our find usages doesn't work in this case =/

- [x] Navigation to an associated type from its binding #5538
```rust
type T = dyn Iterator<Item/*caret*/=i32>; // Ctrl + B
```
In 2020.2 it does nothing

vlad20012:
There is a discussable question: maybe we really want to do "Find Usages" by Ctrl +B (Ctrl + Click) on module declarations and extern crates. In this case, we could provide navigation to files by "Go To Implementation" (Ctrl + Alt + B) action

Kobzol:
My 2 cents: I personally use Ctrl+click on mod items for navigation, it's a quick way of moving between "neighbour" files. I expect Find usages to be invoked only if there is no target element to navigate to, for example in generic parameters: `struct S<T/*caret*/>(T);`. If there's something to navigate to, I would prefer to keep the old behaviour.

Undin:
@vlad20012 everything works with latest 2020.2 EAP again

PR #5538, opened 2020-06-09
Fix go-to-declaration from assoc type binding on 2020.2

vlad20012:
Fixes "Go To Declaration" (Ctrl + B) action in this case on the 2020.2 platform
```rust
trait Foo { type Item; }
type T = dyn Foo</*caret*/Item = i32>;
``` 
Part of #5537


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -49,7 +49,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
     override fun getType() = Type
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
-        private const val STUB_VERSION = 198
+        private const val STUB_VERSION = 199
```
vlad20012, line 52: I bumped stub version just in case, the stub structure shouldn't change

PR #5539, opened 2020-06-09
INSP: improve AddUnsafeFix to find the closest enclosing block

Kobzol:
https://github.com/intellij-rust/intellij-rust/pull/5521 fixed `AddUnsafeFix` to work with conditions and loops, but it added `unsafe` to the enclosing function in such cases, which is a bit too pessimistic.

This PR changes it to find the closest enclosing block.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/AddUnsafeFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/AddUnsafeFix.kt
@@ -10,26 +10,31 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
-import org.rust.lang.core.psi.RsBlock
+import com.intellij.psi.util.PsiTreeUtil
 import org.rust.lang.core.psi.RsBlockExpr
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.RsPsiFactory
-import org.rust.lang.core.psi.ext.ancestorOrSelf
 
-class AddUnsafeFix(expr: RsBlock) : LocalQuickFixAndIntentionActionOnPsiElement(expr) {
-    private val _text = "Add unsafe to ${if (expr.parent is RsBlockExpr) "block" else "function"}"
+class AddUnsafeFix private constructor(element: PsiElement) : LocalQuickFixAndIntentionActionOnPsiElement(element) {
+    private val _text = "Add unsafe to ${if (element is RsBlockExpr) "block" else "function"}"
     override fun getFamilyName() = text
     override fun getText() = _text
 
-    override fun invoke(project: Project, file: PsiFile, editor: Editor?, expr: PsiElement, endElement: PsiElement) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/AddUnsafeFix.kt
```diff
@@ -32,4 +33,16 @@ class AddUnsafeFix(expr: RsBlock) : LocalQuickFixAndIntentionActionOnPsiElement(
...
+    companion object {
+        fun findApplicableBlock(element: PsiElement): RsBlock? {
+            var block = element.ancestorStrict<RsBlock>()
+
+            while (block != null && block.parent !is RsBlockExpr && block.parent !is RsFunction) {
+                block = block.ancestorStrict()
+            }
+
+            return block
+        }
```
mchernyavsky, line 46:
```suggestion
        fun findApplicableBlock(element: PsiElement): RsBlock? =
            when (val parent = PsiTreeUtil.getParentOfType(element, RsBlockExpr::class.java, RsFunction::class.java)) {
                is RsBlockExpr -> parent.block
                is RsFunction -> parent.block
                else -> null
            }
```

mchernyavsky: And maybe it's better to move this logic into `invoke`?

Kobzol: It's in `findApplicableBlock` because `RsDiagnostic.UnsafeError` uses it to find the corresponding block - it is used to set the fix message (`Add unsafe to block` vs `Add unsafe to function`).

mchernyavsky: Ok, then maybe create some factory method instead of `findApplicableContext`? Something like this:

```kotlin
companion object {
    fun create(element: PsiElement): AddUnsafeFix? {
        val parent = PsiTreeUtil.getParentOfType(element, RsBlockExpr::class.java, RsFunction::class.java) ?: return null
        return AddUnsafeFix(parent)
    }
}
```

```kotlin
class AddUnsafeFix private constructor(expr: PsiElement) : LocalQuickFixAndIntentionActionOnPsiElement(expr) {
```

```kotlin
class UnsafeError(
    element: RsExpr,
    private val message: String
) : RsDiagnostic(element) {
    override fun prepare() = PreparedAnnotation(
        ERROR,
        E0133,
        message,
        fixes = listOfNotNull(SurroundWithUnsafeFix(element as RsExpr), AddUnsafeFix.create(element))
    )
}
```

```kotlin
override fun invoke(project: Project, file: PsiFile, editor: Editor?, parent: PsiElement, endElement: PsiElement) {
    val unsafe = RsPsiFactory(project).createUnsafeKeyword()
    when (parent) {
        is RsBlockExpr -> parent.addBefore(unsafe, parent.block)
        is RsFunction -> parent.addBefore(unsafe, parent.fn)
        else -> error("unreachable")
    }
}
```

Kobzol: Thanks, that is indeed better.

mchernyavsky:
bors retry

Issue #4509, opened 2019-10-09
`use crate_name::crate_name` breaks resolution


AlyoshaVasilieva:
## Environment

* **IntelliJ Rust plugin version:** 0.2.107.2135-193 (previous also tested, I think)
* **Rust toolchain version:** 1.38.0
* **IDE name and version:** IntelliJ IDEA Ultimate 2019.3 193.4099.13 EAP (2019.2.3 also tested)
* **Operating system:** Windows 10 x64

## Problem description
If a user imports a macro (or possibly anything) named the same name as the crate containing it, then navigation/resolution break for that crate. For example:

* `use anyhow::anyhow;`
* `use lazy_static::lazy_static;`

## Steps to reproduce

Cargo.toml:
```toml
[package]
name = "anyhow-test"
version = "0.1.0"
authors = ["Alyosha Vasilieva <1284317+AlyoshaVasilieva@users.noreply.github.com>"]
edition = "2018"

[dependencies]
anyhow = "1.0.8"
```

main.rs (working version):
```rust
use anyhow::{Error, ensure};

fn main() -> Result<(), Error>{
    println!("Hello, world!");
    Ok(())
}
```

Everything is normal in the working version. (There is a compile warning about `ensure` not being used, of course)

main.rs (broken version):
```rust
use anyhow::{Error, anyhow};

fn main() -> Result<(), Error>{
    println!("Hello, world!");
    Ok(())
}
```

In the broken version:
* Ctrl-clicking the use statements gives "Cannot find declaration to go to"
* The `Error` usage is red and reads "Unresolved reference: Error"

However, compilation works fine (again with a warning about unused import).

Auto-completion is broken by this. If you have `use anyhow::{Error, anyh};` then IntelliJ will happily autocomplete the "anyhow" on request and then die and stop giving any autocomplete suggestions for `anyhow`.

fan-tom:
Isn't it the same as #4186?

YvesChan:
Bug still unresolved? Same problem here in goland, it breaks all the resolutions related with anyhow, and make a lot syntax error.

PR #5546, opened 2020-06-10
FIX: crate resolve for `use crate_name::crate_name`

Kobzol:
This PR attempts to fix resolving when `use crate_name::crate_name` is used in a crate with 2018 edition. This situation was previously fixed for `use crate_name;`, `use crate_name::{self};` and other use cases such as re-exports and aliases (https://github.com/intellij-rust/intellij-rust/pull/3850, https://github.com/intellij-rust/intellij-rust/pull/4168).

The name resolution machinery is a complex beast, so I'm not sure if this is the proper change to make, but it fixed the issue and all resolve tests passed locally, so let's see what CI says.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4509 (hopefully)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
@@ -126,7 +126,14 @@ private fun RsElement.processItem(processor: (RsItemElement) -> Boolean): Boolea
 
 private val RsPath.isAtom: Boolean
     get() = when (kind) {
-        PathKind.IDENTIFIER -> qualifier == null
+        PathKind.IDENTIFIER -> {
+            val qual = qualifier
+            if (qual == null) {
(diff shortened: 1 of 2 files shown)


vlad20012:
@Kobzol, I think you haven't fully understood the issue :) 

First, looks like your test is incorrect, this code is not compiled by rustc =D 

We want to bring multiple `crate_name` names with different namespaces into the scope. I.e. first `crate_name` is a crate itself, it is in `Types` scope, and second `crate_name` name in `Values` or `Macros` scope. 

The issue is that a namespace wasn't really taken into account during names shadowing. The bug was known, but I didn’t want to fix this previously because we're going to rewrite the name resolution completely, so there is no need to make cosmetic repairs before the bombing... But looks like we issue is major enough because there are multiple popular crates that suffer of this bug. So, I'm trying to fix it in #5603.

Issue #5541, opened 2020-06-09
RsMainFunctionNotFoundInspection is too eager


Kobzol:
## Environment

* **IntelliJ Rust plugin version:** 0.2.125.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.44.0 (49cae5576 2020-06-01) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** Linux 5.0.0-37-generic

## Problem description
The newly added `RsMainFunctionNotFoundInspection` is a bit too quick to jump in and it appears when an error element appears in the document even if there is a `main` function present. That causes annoying flickering when an error repeatedly appears in the main file (for example when you are typing out function declaration).

## Steps to reproduce
```rust
fn foo( {  // error, inspection pops in
    unimplemented!()
}
fn main() {
    println!("Hello, World");
}
```

I suggest to only show the inspection notification bar if the file does not contain `PsiErrorElement` as a child. In this way the flickering will still be present if errors repeatedly appear, but only in the opposite case - if there is no `main` function present, which should happen less often than when `main` function exists + there is an error.

Kobzol:
@Stzx what do you think?

Stzx:
@Kobzol 

Oh yes. (The IDE license in the project has expired. I installed the dev version in my community version and saw this situation.

It seems we need a code similar to the following to fix
```kotlin
if (file.childrenOfType<PsiErrorElement>().size > 0) return
```

PR #5547, opened 2020-06-10
INSP: do not trigger RsMainFunctionNotFoundInspection for files with a root error

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5541


src/test/kotlin/org/rust/ide/inspections/RsMainFunctionNotFoundInspectionTest.kt
```diff
@@ -15,6 +15,12 @@ class RsMainFunctionNotFoundInspectionTest : RsInspectionsTestBase(RsMainFunctio
         <error descr="`main` function not found in crate `test-package` [E0601]"> /*caret*/ </error>
     """)
 
+    fun `test do not show if an error exists`() = checkByFileTree("""
```
Undin, line 18: Let's also add test where file contains psi error element but as a nondirect child to ensure that we show error in this case

Issue #5166, opened 2020-04-01
No "mismatched types" [E0308] error on empty function


Stzx:
## Environment

* **IntelliJ Rust plugin version:** 0.2.118.2171-193
* **Rust toolchain version:** cargo 1.44.0-nightly (7019b3ed3 2020-03-17)
* **IDE name and version:** IDEA 2019.3.4
* **Operating system:** Windows 10

## Problem description

```rust
fn test_fun() -> ! {
    // For empty function return `()`, no error detected.
    // Actually this will raise [E0308] error.
}
```

## Steps to reproduce

1. Use the following code:
```rust
fn test_fun() -> ! {
}
```
2. Run `cargo check`

ortem:
@Stzx thank you for reporting. Looks like the problem is not only with diverging functions but with any functions with a return type, e.g.
```Rust
fn foo() -> i32 {} // no error highlighted
```

Issue #4833, opened 2020-01-10
Highlight the return type in a signature as an error instead of the whole function body


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.114.2826-193-nightly
* **Rust toolchain version:** 1.39.0 (4560ea788 2019-11-04) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 Snapshot (CL-201.3671)
* **Operating system:** macOS 10.15

## Problem description

I usually write return/collect/... last when implementing a function/method, this way the plugin is constantly complaining about the wrong return type till the function is actually finished. This means the whole function body is underlined in red 99% of the time. Sometimes I even put a mock value in the end while coding just to see more meaningful errors hidden behind wrong type.

I would argue that's too harsh, why not highlighting the return type in the function signature as an error instead? This way it will be easier on eyes and still present.

<img width="808" alt="Screen Shot 2020-01-10 at 19 45 30" src="https://user-images.githubusercontent.com/2500320/72170406-e1f7e800-33e1-11ea-8299-bba5e3b89c4a.png">

ortem:
That's an interesting idea. Another option is more general (but harder to implement right now) [Scala way](https://blog.jetbrains.com/scala/2019/07/24/intellij-scala-plugin-2019-2-functional-code-highlighting-fine-grained-type-diff-case-clause-completion-build-server-protocol-and-more/#functional-code-highlighting).

![image](https://user-images.githubusercontent.com/4854600/72249722-0256bb00-360b-11ea-9d14-96c46455f467.png)

Kobzol:
So, is there consensus whether to implement this or not? :) I think that the trade-off of annotating the return type is worth it. One thing to note: if there are multiple return statements (possibly of different types), we probably shouldn't offer the quick fix on the return type.

Kobzol:
Btw in https://github.com/intellij-rust/intellij-rust/pull/5552 I already annotate the return type for empty functions.

jbis9051:
+1

PR #5552, opened 2020-06-11
ANN: annotate wrong return type for empty functions

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5166

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -898,6 +898,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         collectDiagnostics(holder, fn)
         checkDuplicates(holder, fn)
         checkTypesAreSized(holder, fn)
+        checkEmptyFunctionReturnType(holder, fn)
 
         fn.innerAttrList.forEach { checkStartAttribute(holder, it) }
         fn.outerAttrList.forEach { checkStartAttribute(holder, it) }
@@ -1367,7 +1368,7 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
     val owner = fn.owner
 
     fun isError(ty: Ty): Boolean = !ty.isSized() &&
-        // '?Sized' type paramter types in abstract trait method is not an error
+        // '?Sized' type parameter types in abstract trait method is not an error
         !(owner is RsAbstractableOwner.Trait && fn.isAbstract)
 
     for (arg in arguments) {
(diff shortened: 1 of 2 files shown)


Kobzol:
If we want to do this: https://github.com/intellij-rust/intellij-rust/issues/4833 then it would be consistent to mark the return type (although it might be missing in case of unit type, hmm).. Otherwise I'll change it to annotate the closing brace.

Stzx:
@Kobzol I think you missed the `unimplemented!()`.

Kobzol:
@stzx I'm sorry, I don't understand, what do you mean?

Stzx:
@Kobzol  Oh, sorry. I didn't check it carefully. The original one already contains `unimplemented!()`.

Kobzol:
@mchernyavsky What error type marking should be used? Only the return type, whole function, closing brace?

mchernyavsky:
> @mchernyavsky What error type marking should be used? Only the return type, whole function, closing brace?

I prefer the closing brace. @Undin @vlad20012 What do you think?

vlad20012:
This is the reason why the PR is not merged yet?  Oh...
I would not say that this is important. We can change it later. Let's make it somehow. `cargo check` already highlight a return type, so we won't make things worse if we do that too.
Btw #5552 can also be implemented by `}` brace highlighting

Kobzol:
> This is the reason why the PR is not merged yet? Oh...

Yup :sweat_smile:

The closing brace is probably more general, since we cannot mark the return type if it's missing. Should I change it to the brace?

vlad20012:
> Should I change it to the brace?

Oh :sweat_smile:
Okay, let's chenge

Kobzol:
I changed it to the closing brace for empty functions. The `ChangeReturnTypeFix` is not currently offered for them. It would require non-trivial changes to it to make it work, and I don't think that you usually want to change the return type in situations when the function is empty. If you want the fix to work, let me know.

To solve #4833 with brace highlighting, it would have to only highlight the brace in the case of a trailing expr (and highlight return expressions for `return` statements, as it does now), which might be a bit confusing.

vlad20012 (approved):
bors r+

vlad20012:
Hmm, I think I found a false-positive case:
```rust
trait FooBar {}
impl FooBar for () {}

fn foobar() -> impl FooBar {

}
```
bors r-

Kobzol:
Oh, that's a very... interesting case :D Well, to support cases such as this one, I wonder if maybe the error annotation for empty functions should be moved into `TypeInferenceWalker`? Or should I just change the current implementation to use something like `TypeInference::combineTypes` to see if the result type is compatible with `()`?

vlad20012:
`TypeInferenceWalker` looks more robust solution, but `impl Trait` is poorly supported even there. Let's just ignore `impl Trait`s at all for now

PR #5557, opened 2020-06-12
INSP: add remove variable quick fix to liveness inspection

Kobzol:
This PR adds a quick fix to remove an unused variable.

It is only offered on variables with a simple pattern binding (`RsPatIdent`):
```rust
let a/*caret*/ = 5; // offered
let S { a/*caret*/, b } = ...; // not offered
```
We can also support these more complex cases. For example for struct field bindings we can remove the whole binding if the only field was removed or add `..` to the end if the struct has multiple fields, but I'm not sure if it's worth the complexity.

Inspired by Kotlin, I use a very simple heuristic to check whether to also remove the variable expression or not. It's conservative and considers most things to be side effects, but it allows `BinaryExpr`s and `UnaryExpr`s if their arguments are side effect free. In most cases that should be safe, but if `*` or `+` is overloaded, it might remove some side effects in the operator implementations in theory. Is it fine? We can also probably check if the operator is overloaded.

I also modified the inspection warning text slightly, to refer to complex bindings properly, i.e.
```rust
fn foo(S { <warning>a</warning>, b }: S) {}
```
now says `Binding 'a' is unused`, not `Parameter 'a' is unused`. I can revert it if you don't like it.

Part of https://github.com/intellij-rust/intellij-rust/issues/3163

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsLivenessInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsLivenessInspection.kt
@@ -14,6 +14,7 @@ import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.ancestorStrict
 import org.rust.lang.core.psi.ext.descendantsWithMacrosOfType
 import org.rust.lang.core.psi.ext.expansion
+import org.rust.lang.core.psi.ext.topLevelPattern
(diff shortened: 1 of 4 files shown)


ortem:
The current implementation looks good to me, thank you!

bors r+

Kobzol:
bors retry

PR #5559, opened 2020-06-12
TOML: improve toml file type detection

Undin:
Implement `FileTypeIdentifiableByVirtualFile` instead of `FileTypeRegistry.FileTypeDetector`
to detect file type of `.cargo/config` file. We don't need to read file content so `FileTypeIdentifiableByVirtualFile` is more suitable in this case


Undin:
CC @vlad20012

Undin:
bors retry

PR #5561, opened 2020-06-12
INSP: add remove parameter quick fix to liveness inspection

Kobzol:
This PR adds a quick fix to remove an unused parameter, along with its usages in call sites. The call site arguments are removed without any checks for side effects, same as in the Kotlin plugin.

Should we ignore call sites with an incorrect number of arguments?

This PR is based on https://github.com/intellij-rust/intellij-rust/pull/5557 which introduced some common changes. I opened this one for discussion, but #5557 should be merged first, I'll rebase this one after/if it gets merged.

Part of https://github.com/intellij-rust/intellij-rust/pull/3163

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/RemoveParameterFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/RemoveParameterFix.kt
@@ -1,0 +1,71 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.fixes
+
+import com.intellij.codeInspection.LocalQuickFixOnPsiElement
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.search.searches.ReferencesSearch
+import com.intellij.psi.util.parentOfType
+import com.intellij.psi.util.parentOfTypes
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsElementTypes.COMMA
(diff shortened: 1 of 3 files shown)


Kobzol:
Rebased onto master.

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveParameterFix.kt
```diff
@@ -0,0 +1,65 @@
...
+        val isMethod = function.hasSelfParameters
+        val argumentIndex = when {
+            isMethod && call is RsCallExpr -> parameterIndex + 1 // UFCS
+            else -> parameterIndex
+        }
+        arguments.exprList.getOrNull(argumentIndex)?.deleteElementAndFollowingComma()
+    }
+}
+
+private fun RsElement.deleteElementAndFollowingComma() {
```
ortem, line 59: What about a potentially redundant comma **before** the element? Currently, in this case
```Rust
fn foo(x: i32, y: i32) -> i32 { x }
```
the redundant comma will remain
```Rust
fn foo(x: i32, ) -> i32 { x }
```

I think we can delete them as well if possible

Kobzol: Ah, good catch, thanks!

Kobzol:
I changed comma handling in such a way that only one comma is always removed, either the preceding or the following, based on which one is present. This handles parameter removal at the beginning, middle and end of parameter lists, and it also keeps trailing commas.

```rust
foo(p0, p1/*caret*/) -> foo(p0)
foo(p0, p1, /*caret*/) -> foo(p0, )
```

Kobzol:
bors r=ortem

PR #5570, opened 2020-06-14
Self enum #4980

### LINKED ISSUE #4980: `Self::` prefix for enum breaks navigation
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.2854-193-nightly
* **Rust toolchain version:** 1.41 stable
* **IDE name and version:** IntelliJ IDEA 2019.3.3 Ultimate
* **Operating system:** Windows 10 x64

## Problem description
If referring to an enum variant with a `Self::` prefix, IntelliJ cannot find it.

## Steps to reproduce

main.rs:

```rust
fn main() {
    let _ = Si7021Commands::MeasureTemperatureNoHold.as_bytes();
}

pub trait Command {
    fn as_bytes(&self) -> &[u8];
}

#[non_exhaustive]
enum Si7021Commands {
    MeasureHumidityNoHold,
    MeasureTemperatureNoHold,
}

impl Command for Si7021Commands {
    fn as_bytes(&self) -> &[u8] {
        match *self {
            Si7021Commands::MeasureHumidityNoHold => &[0xF5],
            Self::MeasureTemperatureNoHold => &[0xF3],
        }
    }
}
```

This compiles fine.

Ctrl-clicking `MeasureHumidityNoHold` in the match takes caret to its declaration in the enum.

Ctrl-clicking `MeasureTemperatureNoHold` gives error "Cannot find declaration to go to".

(I was also making a repro for a bug with `#[non_exhaustive]` but it was already reported as #4896)

fan-tom:
Lets `Self` to be resolved into `RsEnumItem` inside impls of enums, so `Self::Var` will resolve to enum variant
Fixes #4980

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
@@ -36,6 +36,13 @@ fun processItemOrEnumVariantDeclarations(
                 if (processAllWithSubst(item.variants, subst, processor)) return true
             }
         }
+        is RsImplItem -> {
+            val (item, subst) = (scope.typeReference?.skipParens() as? RsBaseType)
+                ?.path?.reference?.advancedDeepResolve() ?: return false
+            if (item is RsEnumItem) {
+                if (processAllWithSubst(item.variants, subst, processor)) return true
+            }
+        }
         is RsEnumItem -> {
             if (processAll(scope.variants, processor)) return true
         }
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -57,7 +57,6 @@ import org.rust.lang.core.types.ty.*
 import org.rust.openapiext.testAssert
 import org.rust.openapiext.toPsiFile
 import org.rust.stdext.buildList
-import org.rust.stdext.notEmptyElseNull
 
 // IntelliJ Rust name resolution algorithm.
 // Collapse all methods (`ctrl shift -`) to get a bird's eye view.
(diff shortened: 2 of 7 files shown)


src/main/kotlin/org/rust/stdext/Collections.kt
```diff
@@ -112,6 +112,11 @@ inline fun <T> List<T>.notEmptyOrLet(function: (List<T>) -> List<T>): List<T> =
     else -> function(this)
 }
 
+inline fun <T> List<T>.notEmptyElseNull(): List<T>? = when {
+    isNotEmpty() -> this
+    else -> null
+}
```
mchernyavsky, line 118: You can use the `takeIf` method:

image

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -370,87 +371,89 @@ private fun processQualifiedPathResolveVariants(
...
         }
         return false
     }
-    val isSuperChain = isSuperChain(qualifier)
-    if (base is RsMod) {
-        val s = base.`super`
-        // `super` is allowed only after `self` and `super`
-        // so we add `super` in completion only when it produces valid path
-        if (s != null && (!isCompletion || isSuperChain) && processor("super", s)) return true
+    for ((base, subst) in variants) {
```
vlad20012, line 381: Hmm, why this multiresolve logic is needed to resolve enum items by Self?

fan-tom: If I understand your question right, it is because `Self` was previously resolved to `impl` item, now it is resolved to both `impl` and `enum` items.

vlad20012: Ok, I get it. We probably can make it simpler. Instead of resolution `Self` into 2 places, we can resolve it to `impl` only ane then handle that `impl` in `ItemResolution.kt` `processItemOrEnumVariantDeclarations` near `RsTypeAlias` case

fan-tom: Am I understand it right, we should turn enum varinats into associated items, like static functions?

vlad20012: Umm, yes, something like that. Maybe I'm wrong with `ItemResolution.kt` and really it should be handled in `processQualifiedPathResolveVariants` next to the other `impl`/`Self`-related stuff.

Btw it would be nice to test a corner case when there is an enum variant and an associated function with the same name.

fan-tom: Tried to follow your suggessions, seems working

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
```diff
@@ -36,6 +36,13 @@ fun processItemOrEnumVariantDeclarations(
                 if (processAllWithSubst(item.variants, subst, processor)) return true
             }
         }
+        is RsImplItem -> {
+            val (item, subst) = (scope.typeReference?.skipParens() as? RsBaseType)
+                ?.path?.reference?.advancedDeepResolve() ?: return false
+            if (item is RsEnumItem) {
+                if (processAllWithSubst(item.variants, subst, processor)) return true
```
fan-tom, line 43: Do I really need `subst` here? Maybe better to use `processAll` instead?

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
fan-tom: >Btw it would be nice to test a corner case when there is an enum variant and an associated function with the same name.
Currently in both cases, with enum name and with `Self`, resolves to both enum variant and method with the same name, so I think it should be separate issue and must include some inspection to give overlap error. Interesting that rustc doesn't report it as error yet https://github.com/rust-lang/rust/issues/48758

src/main/kotlin/org/rust/lang/core/resolve/ItemResolution.kt
vlad20012: Looks like yes: rustc shows an error in this case:
```rust
trait Foo {
    fn foo();
}

impl Foo for Option<i32> {
    fn foo() {
        let a: Option<u8> = Self::None;
    }
}
```
that means that rustc takes into account actual generic arguments of the enum

PR #5571, opened 2020-06-15
INSP: Improve lint check

### LINKED ISSUE #5567: Annotation level of lint inspections doesn't respect lint level of lint check attributes
## Environment

## Problem description
The plugin provides some "Lint" inspections. For example, "Trait naming convention", "Liveness analysis", etc. Such inspections are related to compiler inspections and error level of such inspections is controlled by [lint check attributes](https://doc.rust-lang.org/reference/attributes/diagnostics.html#lint-check-attributes).
You can mark some lint check as compiler error with `deny` or `forbid` attributes. But plugin inspections don't take them into account and always use warning level.

image

## Steps to reproduce
```rust
#[error(non_camel_case_types)]

trait foo {}
```

Stzx:
#### Improve
* Associate annotation level with lint level
* Support for `forbid` attribute

Closed #5567

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsLint.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsLint.kt
@@ -5,7 +5,9 @@
 
 package org.rust.ide.inspections
 
+import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.psi.PsiElement
+import org.rust.ide.inspections.RsLintLevel.*
 import org.rust.lang.core.psi.ext.*
 
 /**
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/ide/inspections/RsLint.kt
```diff
@@ -14,14 +17,15 @@ import org.rust.lang.core.psi.ext.*
...
+    val defaultLevel: RsLintLevel = WARN,
+    val highlightType: Map<RsLintLevel, ProblemHighlightType> = mapOf(WARN to WARNING, DENY to GENERIC_ERROR)
 ) {
     NonSnakeCase("non_snake_case", listOf("bad_style")),
     NonCamelCaseTypes("non_camel_case_types", listOf("bad_style")),
     NonUpperCaseGlobals("non_upper_case_globals", listOf("bad_style")),
-    Deprecated("deprecated"),
-    UnusedVariables("unused_variables", listOf("unused")),
-    NeedlessLifetimes("clippy::needless_lifetimes", listOf("clippy::complexity", "clippy::all", "clippy"));
+    Deprecated("deprecated", highlightType = mapOf(WARN to LIKE_DEPRECATED, DENY to GENERIC_ERROR)),
```
Undin, line 26: Personally, I don't like this mapping via `map` because it's quite easy to forget some items in mapping and get wrong highlighting.
I suggest implementing a bit more wordy but less error-prone way. Let's add something like
```
protected open fun toHighlightingType(level: RsLintLevel): ProblemHighlightType {
    return when (level) {
        ALLOW -> INFORMATION
        WARN -> WARNING
        DENY, FORBID -> GENERIC_ERROR
    }
}
```
and override it for each lint variant (if necessary, of course). For example, for `Deprecated`:
```
override fun toHighlightingType(level: RsLintLevel): ProblemHighlightType {
    return if (level == WARN) LIKE_DEPRECATED else super.toHighlightingType(level)
}
```

Of course, you will have to move `getProblemHighlightType` from `RsLintInspection` to `RsLint` but it's ok

src/main/kotlin/org/rust/ide/inspections/RsLint.kt
```diff
@@ -33,7 +37,12 @@ enum class RsLint(
         .flatMap { it.queryAttributes.metaItems }
         .filter { it.metaItemArgs?.metaItemList.orEmpty().any { it.id == id || it.id in groupIds } }
         .mapNotNull { it.name?.let { RsLintLevel.valueForId(it) } }
-        .firstOrNull()
+        .let {
```
Undin, line 40: Do we really need it?

src/test/kotlin/org/rust/ide/inspections/RsLintLevelTest.kt
```diff
@@ -99,6 +99,20 @@ class RsLintLevelStructTest : RsInspectionsTestBase(RsStructNamingInspection::cl
         #[derive(Debug)]
         struct inhabitant;
     """)
+
+    fun testDirrectForbid() = checkByText("""
```
Undin, line 103: Let's use `test dirrect forbid` style for test naming.
`testDirrectForbid` is old style that we don't use anymore.
If you want, you may convert all test names in this file to use new style

src/main/kotlin/org/rust/ide/inspections/RsLint.kt
Stzx: If implemented using the above method, I think we will no longer need it.

Stzx: Yes i think we still need it to avoid the `forbid` declaration at the superior level.

But let me change the code a little bit and return `FORBID` instead of `DENY`.

src/main/kotlin/org/rust/ide/inspections/RsLint.kt
Stzx: > you will have to move `getProblemHighlightType` from `RsLintInspection` to `RsLint` but it's ok

Why do I need to move `getProblemHighlightType`?

src/test/kotlin/org/rust/ide/inspections/RsLintLevelTest.kt
Stzx: Okay, I will convert all test names to the new style.

Kobzol: And fix the `dirrect` typo please, it makes me nervous everytime I look at it! :D

src/main/kotlin/org/rust/ide/inspections/RsLint.kt
Undin: Because `toHighlightingType` is protected and can be used only from `RsLint`. I don't think we should make `toHighlightingType` public because it's an implementation detail of `RsLint`

Stzx: Oh, it seems that I misunderstood, I thought `toHighlightingType` was declared in `RsLintInspection`. 😆

Issue #5327, opened 2020-05-05
replace all occurrences not work


Sherlock-Holo:
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.3059-201-nightly
* **Rust toolchain version:** rustc 1.45.0-nightly (65b448273 2020-05-03)
* **IDE name and version:** clion 2020.1.1
* **Operating system:** Arch Linux

## Problem description
replace all occurrences doesn't work

## Steps to reproduce

1. define a function like `get_data() -> u64`
2. use it in line 2
3. in line 1, type `get_data().let`
4. use the **replace all 2 occurrences**
5. it should create a variable at line 1 and replace line 2 `get_data()` with the variable, but not

Sherlock-Holo:
with plugin version `0.2.125.3154-201-nightly` it still doesn't work, is there any fixing plan for this? 😢

Kobzol:
Could you please post a specific Rust code snippet and tell us what action/refactoring are you trying to run? I don't understand what are you trying to achieve from your description. Thanks.

Sherlock-Holo:
original code
```rust
fn main() {
    let v = Vec::new();

    use_len(len(v))
}

fn len(v: Vec<i32>) -> usize {
    v.len()
}

fn use_len(len: usize) {}
```

insert `len(v)` before `use_len(len(v))` and type `.let` to create a variable for `len(v)`, it will display 
![image](https://user-images.githubusercontent.com/10096425/84590523-0319a480-ae6a-11ea-91cf-636a91f58fde.png)

choose "replace 2", but `use_len(len(v))` doesn't replace

dima74:
Looks likes this was fixed by #8100

PR #5573, opened 2020-06-15
FIX: replace all occurrences of bare expressions in RsIntroduceVariableHandler

Kobzol:
`RsIntroduceVariableHandler` wasn't replacing all occurrences of the selected expression in some cases, I added occurrence replacement to the `inlineLet` method.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5327

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/introduceVariable/impl.kt
@@ -40,8 +40,8 @@ private fun replaceExpression(editor: Editor, chosenExpr: RsExpr, exprs: List<Ps
 
     val replacer = ExpressionReplacer(project, editor, chosenExpr)
     when {
-        anchor == chosenExpr -> replacer.inlineLet(project, editor, chosenExpr, chosenExpr)
-        parent is RsExprStmt -> replacer.inlineLet(project, editor, chosenExpr, chosenExpr.parent)
+        anchor == chosenExpr -> replacer.inlineLet(project, editor, chosenExpr, chosenExpr, exprs)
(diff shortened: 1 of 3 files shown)


Kobzol:
Hmm, this behaviour is actually pretty strange. Here's what Kotlin does:
```kotlin
val a = 1
1/*caret*/
val b = 1

vvv

val i = 1
val a = i
i
val b = i
```
Therefore no *local* variable was introduced, because there's still `i` alone on its line. Maybe for this specific intention (`Introduce *local* variable`), we should disable the option to replace all occurrences?

Or is the difference between `Introduce variable` and `Introduce a local variable` simply that the first is a refactoring and the second is an intention?

mchernyavsky:
Many other plugins (e.g. C++) act in this way. However, Python does this:

```python
a = 1
1
b = 1

vvv

i = 1
a = i
b = i
```

I'm not sure which is better. @Undin What do you think?

Undin:
I think Python behavior quite unexpected because Java/Kotlin/C++ introduce variable refactorings do another transformation.
For the following code:
```rust
let a = 1;
1/*caret*/;
let b = 1;
```
I expected the following result in case of "Replace this occurrence only"
```rust
let a = 1;
let i = 1;
let b = 1;
```
and

```rust
let i = 1;
let a = 1;
i;
let b = 1;
```
in case of "Replace all occurrences"

Kobzol:
@Undin What behaviour should I implement?

mchernyavsky:
cc @Undin

dima74:
@Kobzol sorry for so long response, and looks like this was fixed in #8100

Kobzol:
This PR was at a standstill anyway, so let's close it.

Issue #1410, opened 2017-07-01
Display crate name for lib.rs in multi-crate workspace


kryptan:
When I open mod.rs there is always a name of the module displayed. However, if I open lib.rs/main.rs/Cargo.toml then name of the crate is displayed only if I open multiple files with the same name.

![image](https://user-images.githubusercontent.com/11528855/27761671-6f3c2128-5e69-11e7-9933-219d30cea2ea.png)

PR #5579, opened 2020-06-15
IDE: show parent folder name in editor tab for lib.rs, main.rs and Cargo.toml

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/1410


src/main/kotlin/org/rust/ide/miscExtensions/RsModTabTitleProvider.kt
```diff
@@ -8,18 +8,25 @@ package org.rust.ide.miscExtensions
 import com.intellij.openapi.fileEditor.impl.UniqueNameEditorTabTitleProvider
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
+import org.rust.cargo.CargoConstants
+import org.rust.lang.RsConstants
 import org.rust.lang.core.psi.isRustFile
 import java.io.File
 
 class RsModTabTitleProvider : UniqueNameEditorTabTitleProvider() {
```
mchernyavsky, line 16: Now the class name is a bit confusing :)

Kobzol:
Thanks, this is much better. It can generate very long names in theory, but since it can now be turned off, that should be ok.

PR #5582, opened 2020-06-16
ANN: Enhanced Repr error detection

### LINKED ISSUE #5566: Show error annotation if repr attribute is placed on an unsupported item
## Environment

## Problem description
Currently, you can place `repr` attribute on each rust item the plugin doesn't produce [E0517](https://doc.rust-lang.org/error-index.html#E0517) error as expected.
It would be great to show the corresponding error in the editor when necessary and add `remove` quick fix. As an example, see `org.rust.ide.annotator.RsErrorAnnotator#checkInlineAttr`

## Steps to reproduce
```rust
#[repr(u8)]
fn foo() {}

#[repr(C)]
trait Foo {}
```

Stzx:
Support:
* `E0517`
* `E0552`

Closed #5566

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -748,6 +748,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
     private fun checkAttr(holder: RsAnnotationHolder, attr: RsAttr) {
         checkImplBothCopyAndDrop(holder, attr)
         checkInlineAttr(holder, attr)
+        checkReprAttribute(holder, attr)
         checkReprForEmptyEnum(holder, attr)
 
         if (attr.owner !is RsFunction)
@@ -802,6 +803,42 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         // Not using `enum.variants` to avoid false positive for enum without body
         if (enum.enumBody?.enumVariantList?.isEmpty() == true) {
             RsDiagnostic.ReprForEmptyEnumError(attr).addToHolder(holder)
+        }
+    }
+
+    private fun checkReprAttribute(holder: RsAnnotationHolder, attr: RsAttr) {
+        if (attr.metaItem.name != "repr") return
+
+        val owner = attr.owner ?: return
+
+        val reprArgs = attr.metaItem.metaItemArgs?.metaItemList.orEmpty()
+
+        check@ for (reprArg in reprArgs) {
+            val reprName = reprArg.name ?: continue
+
+            val errorText = when (reprName) {
+                "C", "transparent", "align" -> when (owner) {
+                    is RsStructItem, is RsEnumItem -> continue@check
+                    else -> "$reprName attribute should be applied to struct, enum, or union"
+                }
+
+                in TyInteger.NAMES -> when (owner) {
+                    is RsEnumItem -> continue@check
+                    else -> "$reprName attribute should be applied to enum"
+                }
+
+                "packed", "simd" -> when (owner) {
+                    is RsStructItem -> continue@check
+                    else -> "$reprName attribute should be applied to struct or union"
+                }
+
+                else -> {
+                    RsDiagnostic.UnrecognizedReprAttribute(reprArg, reprName).addToHolder(holder)
+                    continue@check
+                }
+            }
+
+            RsDiagnostic.ReprAttrUnsupportedItem(reprArg, errorText).addToHolder(holder)
         }
     }
 
--- a/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
@@ -2756,6 +2756,7 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
             let x = 1<error descr="`...` syntax is deprecated. Use `..` for an exclusive range or `..=` for an inclusive range">...</error>;;
         }
     """)
+
     fun `test inclusive range with end E0586`() = checkErrors("""
         fn foo() {
             let x = 1..=2;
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -772,6 +773,39 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         }
     }
 
+    private fun checkReprAttribute(holder: RsAnnotationHolder, attr: RsAttr) {
+        if (attr.metaItem.name != "repr") return
+
+        val reprArgs = attr.metaItem.metaItemArgs?.metaItemList.orEmpty()
+
+        reprArgs.forEach {
+            val reprName = it.name ?: return
```
mchernyavsky, line 782: This is non-local `return` directly to the caller of `checkReprAttribute`. Is it intentional or should it be a local return?

mchernyavsky: > or should it be a local return?

In this case, let's use a simple for loop with a label:
```kotlin
label@ for (reprArg in reprArgs) {
    val reprName = reprArg.name ?: continue
    val owner = attr.owner ?: continue

    val errorText = when (reprName) {
        "C", "transparent", "align" -> when (owner) {
            is RsStructItem, is RsEnumItem -> continue@label
            else -> "$reprName attribute should be applied to struct, enum, or union"
        }
        in REPR_PRIMITIVE_REPRESENTATIONS -> when (owner) {
            is RsEnumItem -> continue@label
            else -> "$reprName attribute should be applied to enum"
        }
        "packed" -> when (owner) {
            is RsStructItem -> continue@label
            else -> "packed attribute should be applied to struct or union"
        }
        else -> {
            RsDiagnostic.UnrecognizedReprAttribute(reprArg, reprName).addToHolder(holder)
            continue@label
        }
    }

    RsDiagnostic.ReprAttrUnsupportedItem(reprArg, errorText).addToHolder(holder)
}
```

Stzx: This is intentional, I think this provides some help for `E0587`, `E0692` and other several wrong implementations (probably?

> ```kotlin
> label@ for (reprArg in reprArgs) {
>     val reprName = reprArg.name ?: continue
>     val owner = attr.owner ?: continue
> 
>     val errorText = when (reprName) {
>         "C", "transparent", "align" -> when (owner) {
>             is RsStructItem, is RsEnumItem -> continue@label
>             else -> "$reprName attribute should be applied to struct, enum, or union"
>         }
>         in REPR_PRIMITIVE_REPRESENTATIONS -> when (owner) {
>             is RsEnumItem -> continue@label
>             else -> "$reprName attribute should be applied to enum"
>         }
>         "packed" -> when (owner) {
>             is RsStructItem -> continue@label
>             else -> "packed attribute should be applied to struct or union"
>         }
>         else -> {
>           RsDiagnostic.UnrecognizedReprAttribute(reprArg, reprName).addToHolder(holder)
>             continue@label
>         }
>     }
> 
>     RsDiagnostic.ReprAttrUnsupportedItem(reprArg, errorText).addToHolder(holder)
> }
> ```

But this may be more intuitive, I will modify.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -969,6 +1006,8 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         !(el.reference.resolve() != null || el.reference.multiResolve().size > 1)
 }
 
+private val REPR_PRIMITIVE_REPRESENTATIONS = listOf("u8", "u16", "u32", "u64", "u128", "usize", "i8", "i16", "i32", "i64", "i128", "isize")
```
mchernyavsky, line 1009: You can use `TyInteger.NAMES`.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -805,6 +806,42 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+                    is RsStructItem, is RsEnumItem -> continue@check
+                    else -> "$reprName attribute should be applied to struct, enum, or union"
+                }
+
+                in REPR_PRIMITIVE_REPRESENTATIONS -> when (owner) {
+                    is RsEnumItem -> continue@check
+                    else -> "$reprName attribute should be applied to enum"
+                }
+
+                "packed" -> when (owner) {
```
mchernyavsky, line 830: What about the `simd` attribute?

Stzx: Oh it looks like i forgot it. :D

Undin:
Hm, sorry
bors r=mchernyavsky

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -3389,4 +3390,64 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
             let a: [u8; <error descr="Calls in constants are limited to constant functions, tuple structs and tuple variants [E0015]">foo</error>(1, <error descr="A non-constant value was used in a constant expression [E0435]">bar</error>)];
         }
     """)
+
+    fun `test E0517 placement repr C`() = checkErrors("""
+        #[repr(<error descr="C attribute should be applied to struct, enum, or union [E0517]">C</error>)]
```
Undin, line 3395: Hm, does it make sense to mark only `C` attribute since `repr` attribute can be applied for type item at all?

vlad20012:
did we run real project tests here?

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
Stzx: There is no special meaning, just to detect whether there is an error when placed in `type`.

Stzx:
> did we run real project tests here?

I used it and tested it on a private project. I think I should try the `cargo` project?

mchernyavsky:
> did we run real project tests here?

I tested it on `cargo` and `mysql_async`.

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
Stzx: PR has been merged, is there anything that needs to be fixed here?

Undin: I see two things that we can improve here:
* if builtin `repr` attribute cannot be applied to a particular item (for example, for trait), there is no need to annotate arguments of the attr. Let's better annotate `repr` attribute itself in this case and provide quick fix to remove it. I know that compiler provides error for a particular argument but I don't know why.
* Potentially, you can create your own `repr` proc macro and provide custom rules for it. In this case, we shouldn't annotate anything because we don't know anything about these rules (at least for now). Note, the current implementation provides false positive for this case only when custom `repr` attribute is imported in current scope.
For example:
```rust
//- lib.rs
#[proc_macro_attribute]
pub fn repr(attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}

//- main.rs
use lib::repr;

#[repr(C)] // <- we should provide neither E0517 nor E0552 in this case because compiler provides E0659
type Foo = i32;
```
Of course, it's rather rare case

Stzx: I want to address this situation, the simplest fix method.

```kotlin
if (attr.metaItem.name != "repr" || attr.metaItem.path?.reference?.resolve() != null) return
```

Undin: Yep, looks perfect solution for the second case

Stzx: This issue also affects declarations such as `inline`, `feature`, etc.

Stzx: > if builtin repr attribute cannot be applied to a particular item (for example, for trait), there is no need to annotate arguments of the attr. Let's better annotate repr attribute itself in this case and provide quick fix to remove it. I know that compiler provides error for a particular argument but I don't know why.

Maybe for this situation(Multiple attribute value declarations)
```rust
#[repr(C, i32)]
struct Test(i32);
```
`C` can be placed on a `struct`, but `i32` is not

Undin: > This issue also affects attribute declarations such as inline, feature, etc.

Yep. Ideally, we should check that there isn't custom attr proc macro in all such cases.
But as I mentioned above, it's a rather rare case

Undin: > Maybe for this situation(Multiple attribute value declarations)

In your case, `repr` attribute may make sense for struct in general, but currently, it doesn't make sense for traits, type aliases, functions, etc. with any parameters so I think it's rather natural to suggest to drop the whole attribute instead of only its part.

In the mentioned case, the current implementation works as expected

Issue #5568, opened 2020-06-14
Literal overflow inspection


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.2.125.3158-202-nightly
* **Rust toolchain version:** 1.46.0-nightly (0262de554 2020-06-07) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2 EAP Ultimate Edition (IU-202.5428.22)
* **Operating system:** macOS 10.13.6

## Problem description
Currently, the plugin doesn't check integer overflow if a literal value is out of supported value range of the particular type. It would be nice to have such check as a separate inspection. Note, there is `overflowing_literals` attribute to modify error level for the compiler (default is `deny`).

## Steps to reproduce
```rust
#[repr(u8)]
pub enum Color {
    White = 0xFFF, // overflow
}

fn main() {
    let a: u16 = 100_000_000; // overflow
}
```

Undin:
See `org.rust.ide.inspections.RsLintInspection` as base class of lint inspections and `org.rust.lang.utils.evaluation.ConstExprBuilder` as base class for constant evaluation

PR #5585, opened 2020-06-16
INSP: add RsIntegerOverflowInspection

Kobzol:
This PR adds inspection that checks integer literal overflows.

I tried to reuse `ConstExprEvaluator` as much as I could, but I ended up rewriting a small evaluation function to have better control of the result. There was also an off-by-one error in some of the integer ranges of `ConstExprEvaluator`. 

There are some unresolved things:
1) `isize/usize`: should we query the pointer size of the architecture of the corresponding Rust target? Or just treat it as `i64/u64`? Or ignore this type altogether?
2) Large types (`i64`, `u64`, `i128`, `u128`) cannot be fully represented by `Long`, which is used by the const evaluator. Rustc fully checks `u64` and `i64`, but it gives up at a certain point and when the integer literal is too large, the compilation fails with an error, so it cannot check all of `i128`/`u128`.

Possible solutions of 2):
- Simply ignore the overflow check if the literal is unrepresentable by `Long`. I think that this solution is ok, as users are unlikely to write such large literals.
- Rewrite a part of the const expr evaluator using `BigInteger`.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5568

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsIntegerOverflowInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsIntegerOverflowInspection.kt
@@ -1,0 +1,54 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiElementVisitor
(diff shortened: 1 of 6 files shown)


Kobzol:
Closing as this is superseded by https://github.com/intellij-rust/intellij-rust/pull/5699.

Issue #2506, opened 2018-04-30
Introduce a variable should be an Intention as well


strangemonad:
When programming Java in IntelliJ, introduce variable is one of my most common workflows (and it's been around for a while:
https://blog.jetbrains.com/idea/2010/11/smarter-and-faster-introduce-variable-in-java/

It would be cool to have something similar in intellij-rust

kumbayo:
This should already work.
Can you provide an example where it does not work as expected?
Thanks!

strangemonad:
Sorry if I'm missing something obvious. I don't see any intention I could enable that would give me this behaviour. Even the simplest case provides no indication that a variable can be introduced:

```
fn s() -> &str {}
fn main() {
   s();
}
```

kumbayo:
```rust
fn s() -> &str {}
fn main() {
   |s|(|)|;
}
```
I tried placing the cursor at the any of the `|` and pressing Alt-Shift-V.
I always end up with the expected result below.
This more or less matching the actions documented in the link you provided.
```rust
fn s() -> &str {}
fn main() {
    let s = s();
}
```
What i see now is that pressing Alt-Enter at the `|` does not show a `Introduce local variable` action/popup.
While it does show that action in Kotlin/Java projects.

strangemonad:
oh, I see now. Yeah, I was looking for the Alt-Enter key and the Introduce Variable Intention (what java and kotlin use) rather than the extract variable refactor.

Issue #4926, opened 2020-02-06
Introduce variable


ortem:
Meta-issue about "Introduce variable" (also known as "extract variable") feature.

Related issues:
#2506
#4608
#4246

Kobzol:
https://github.com/intellij-rust/intellij-rust/pull/4985 has now been merged

PR #5586, opened 2020-06-16
INT: add intention to introduce local variable

Kobzol:
This PR makes the `Introduce variable` refactoring available as an action.

Umbrella issue: https://github.com/intellij-rust/intellij-rust/issues/4926
Fixes: https://github.com/intellij-rust/intellij-rust/issues/2506


vlad20012:
:+1:

bors r+

src/main/resources/intentionDescriptions/IntroduceLocalVariableIntention/before.rs.template
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    5 + 5;
```
Undin, line 2: Maybe surround with `<spot>` attribute?

Kobzol: Added, thanks.

Kobzol:
bors retry

Issue #1004, opened 2017-02-13
Variable can be immutable intention


mkaput:
Add intention to detect effectively immutable variables, like:

```rust
struct St(i32);

fn foo(v: &St) {
    unimplemented!();
}

fn foo_mut(v: &mut St) {
    unimplemented!();
}

trait Bar {
    fn bar(&self);
    fn bar_mut(&mut self);
}

impl Bar for St {
    fn bar(&self) {
        unimplemented!();
    }
    
    fn bar_mut(&mut self) {
        unimplemented!();
    }
}

fn main() {
    // This variable is annotated because we do not
    // use its mutability (it's effectively immutable)
    let mut triggers = St(10);
    foo(&triggers);
    triggers.bar();

    // This variable isn't annotated because of one of these calls
    let mut does_not_trigger= St(10);
    foo_mut(&mut does_not_trigger);
    does_not_trigger.bar_mut();
    does_not_trigger= St(20);

    // What about macros??
    // Maybe simple token scanning for matching identifier
    // would work (as far as I remember macro hygiene prohibits
    // implicit variable usages). Or we can just ignore them.
    some_funny_macro!(does_not_trigger);
}
```

farodin91:
I would like to work on it.

Kobzol:
@ortem should this be an additional cached analysis pass, similar to `RsInferenceContextOwner.liveness`? It would scan local usages of bindings in variables, parameters, match bindings, etc., check if they are used within any context that requires mutability and if not, it would mark them. If you agree with this direction, I'd try to implement it.

ortem:
@Kobzol Yes, I think it can be an additional cached analysis pass like liveness. I don't mind if you try to implement it 👍

Kobzol:
Hmm, I found out that there is actually already an inspection for this - `RsVariableMutableInspection`, but it's disabled by default and I didn't know about it, aargh! :D I think that it's a shame that such a useful inspection is disabled. I'll transform it to a lint inspection and enable it by default to gather possible false positives.

ortem:
@Kobzol `RsVariableMutableInspection` [had been created](https://github.com/intellij-rust/intellij-rust/pull/1143) before our control-flow analysis was implemented, so this inspection is very simple and doesn't support lots of real-world cases. So it has many false-positives and then should not be enabled by default.

[...]

PR #5588, opened 2020-06-16
INSP: change RsVariableMutableInspection to a LintInspection

Kobzol:
This PR changes `RsVariableMutableInspection`, introduced in https://github.com/intellij-rust/intellij-rust/pull/1143, to a `RsLintInspection` and enables it by default.

I think that this inspection is incredibly useful. I have been using the Rust plugin for more than two years and I had no clue about its existence, so I think that it would be good if we enable it by default, at least in nightly builds for now, to gather possible false positives.

I'll try to find cases where false positives happen and try to fix them to make this inspection better. If you know about any one reason why this shouldn't be enabled by default except for false positives, let me know.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/1004

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsLint.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsLint.kt
@@ -13,14 +13,15 @@ import org.rust.lang.core.psi.ext.*
  */
 enum class RsLint(
     val id: String,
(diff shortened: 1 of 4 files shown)


Kobzol:
Closing in favour of https://github.com/intellij-rust/intellij-rust/pull/5737.

Issue #4733, opened 2019-12-07
False positive on `impl Trait` declaration inside `impl<...>`


twistedfall:
## Environment

* **IntelliJ Rust plugin version:** 0.2.111.2147-193
* **Rust toolchain version:** 1.39.0 (4560ea788 2019-11-04) x86_64-unknown-linux-gnu
* **IDE name and version:** IntelliJ IDEA 2019.3 Ultimate Edition (IU-193.5233.102)
* **Operating system:** Linux 5.4.2-gentoo

## Steps to reproduce
Hi!

Intellij Rust plugin shows false error "impl Trait not allowed outside of function and inherent method return types [E0562]" on the following [valid](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=154fefcc5c77f6892710170758641e44) code:
```rust
pub trait IteratorExt {
	fn join(&mut self, sep: &str) -> String;
}

impl<T: Iterator<Item=impl AsRef<str>>> IteratorExt for T {
	fn join(&mut self, sep: &str) -> String {
		"".to_string()
	}
}
```
![afbeelding](https://user-images.githubusercontent.com/406037/70371230-01768f00-18d1-11ea-90db-42a4dbf739ac.png)

Thanks!

PR #5599, opened 2020-06-17
ANN: fix false positive of impl Trait in associated type binding

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/4733


vlad20012:
Thanks!
bors r+

Issue #4451, opened 2019-09-25
Add an intention to replace var.clone() with T::clone(&var) for certain types


SomeoneToIgnore:
## Environment

* **IntelliJ Rust plugin version:** 0.2.106.2134-193
* **Rust toolchain version:** 1.39.0-nightly (6ef275e6c 2019-09-24) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.3519.24)
* **Operating system:** macOS 10.14.6

## Problem description

For smart pointers such as `Rc` and `Arc` sometimes it might be a good idea to explicitly indicate that no extra allocation happens by explicitly writing the `Arc::clone` instead of the regular `.clone()` call.

It would be good to have an intention to convert the regular `.clone()` calls into the other version.

Example:

```rust
let test = Arc::new(5);
// I want to turn this
let regular_clone_call = test.clone();
// into this
let explicit_arc_clone = Arc::clone(&test);
```

vlad20012:
What do you think about an intention that converts any method call into UFCS form and vice versa?

SomeoneToIgnore:
Sounds even better.

Kobzol:
Closed by https://github.com/intellij-rust/intellij-rust/pull/5601 and https://github.com/intellij-rust/intellij-rust/pull/5614.

PR #5601, opened 2020-06-18
INT: convert method call to UFCS

Kobzol:
This PR adds an intention that converts a method call to a UFCS call. I'll add the corresponding intention in the opposite direction in a future PR.

ufcs

For trait methods, it seems that using `<dyn Trait>::method(...)` is pretty universal, although it's not super pretty. Should we include some heuristic that checks whether it can be replaced by something simpler, for example `Trait::method(...)` or `T::method(...)` for generic types?

Fixes first part of https://github.com/intellij-rust/intellij-rust/issues/4451.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/ConvertMethodCallToUFCSIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/ConvertMethodCallToUFCSIntention.kt
@@ -1,0 +1,125 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.ide.inspections.import.AutoImportFix
+import org.rust.ide.inspections.import.import
+import org.rust.ide.presentation.renderInsertionSafe
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.TraitImplSource
+import org.rust.lang.core.resolve.ref.MethodResolveVariant
+import org.rust.lang.core.types.inference
+import org.rust.lang.core.types.ty.*
+import org.rust.lang.core.types.type
+
+class ConvertMethodCallToUFCSIntention : RsElementBaseIntentionAction<ConvertMethodCallToUFCSIntention.Context>() {
+    override fun getText() = "Convert to UFCS"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(
+        project: Project,
+        editor: Editor,
+        element: PsiElement
+    ): Context? {
+        val methodCall = element.parent as? RsMethodCall ?: return null
+        val function = methodCall.reference.resolve() as? RsFunction ?: return null
+        val methodVariants = methodCall.inference?.getResolvedMethod(methodCall).orEmpty()
+        if (methodVariants.isEmpty()) return null
+        return Context(methodCall, function, methodVariants)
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/intentions/ConvertMethodCallToUFCSIntention.kt
```diff
@@ -0,0 +1,112 @@
...
+class ConvertMethodCallToUFCSIntention : RsElementBaseIntentionAction<ConvertMethodCallToUFCSIntention.Context>() {
+    override fun getText() = "Convert to UFCS"
+    override fun getFamilyName() = text
+
+    override fun findApplicableContext(
+        project: Project,
+        editor: Editor,
+        element: PsiElement
+    ): Context? {
+        val methodCall = element.parentOfType<RsMethodCall>(true) ?: return null
```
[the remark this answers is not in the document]
vlad20012: Let's also limit the intention scope to a method identifier. I.e. make it available only when the caret is placed next to a method identifier. Otherwise, it can be too annoying in a large method calls like
```rust
foo.bar(
    baz.quux(spam.eggs(0)),
    foobar.quuux(a + b.get_value())
)
```

Kobzol:
I changed `<dyn Trait>` to `Trait`, sort method variants to prefer inherent impls and added more tests.

src/test/kotlin/org/rust/ide/intentions/ConvertMethodCallToUFCSTest.kt
```diff
@@ -0,0 +1,360 @@
...
+        fn foo<T: Trait>(foo: &T) {
+            foo./*caret*/foo();
+        }
+    """, """
+        trait Trait {
+            fn foo(&self);
+        }
+
+        fn foo<T: Trait>(foo: &T) {
+            Trait::foo(foo);
```
vlad20012, line 191: Why not `T::foo`?

src/main/kotlin/org/rust/ide/intentions/ConvertMethodCallToUFCSIntention.kt
```diff
@@ -0,0 +1,119 @@
...
+    val variant = methodVariants.minBy {
+        if (it.source is TraitImplSource.ExplicitImpl) {
+            0
+        } else {
+            1
+        }
+    } ?: error("Method not resolved to any variant")
+
+    val traitName = (variant.source.value as? RsTraitItem)?.name
+    return traitName ?: variant.selfTy.renderInsertionSafe()
```
vlad20012, line 80: What about a test with generic struct? =)

I think type generics should not be rendered too. Compare
```
Arc::<i32>::clone(foo);
Arc::clone(foo);
```

But even if we render type generic, `renderInsertionSafe` produces a type that is not valid in expression context (`Arc<i32>` vs `Arc::<i32>`)

vlad20012:
Oh, let's also add a test with impl for a type like [u8] that should be placed in `<>` when used as path qualifier: `<[u8]>::foo()`

src/test/kotlin/org/rust/ide/intentions/ConvertMethodCallToUFCSTest.kt
Kobzol: I wanted to do that, but didn't know how to get the type parameter elegantly. I switched to rendering the `selfTy` of the variant in most cases.

vlad20012:
Could you please squash these commits?

vlad20012:
:+1: Thanks! I really like this intention :)
bors+

Kobzol:
Looks like it wasn't picked up.
bors r+

Kobzol:
bors retry

PR #5604, opened 2020-06-18
T: add script to run real project tests

Kobzol:
Running the real project tests manually and checking the output is very time consuming. I created a simple Python script that tries to automate some of the process.

Right now it can:
1) Initialize the real project git repositories by cloning them from GitHub
2) Run the real project test suite and store the error/warning output in a normalized form into a file that can be then diffed easily
3) Optionally it can run `git stash` and switch git to a specific revision before running the test

Other things that it could do is filter the executed tests (by temporarily modifying `RsRealProjectAnalysisTest.kt` or by passing environment variables to the test). The diff process could also probably be automated.

@ortem @vlad20012 what do you think? Is it worth to integrate this script or should I keep it to myself? :)

### DIFF
--- a/scripts/real_project_tests.py
+++ b/scripts/real_project_tests.py
@@ -1,0 +1,113 @@
+import os
+import re
+import subprocess
+import time
+from collections import defaultdict
+
+import click
+
+from common import git_command
+
+ROOT_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+TEST_DATA_PATH = os.path.join(ROOT_PATH, "testData")
+
+PROJECTS = [
+    ("cargo", "https://github.com/rust-lang/cargo"),
+    ("mysql_async", "https://github.com/blackbeam/mysql_async"),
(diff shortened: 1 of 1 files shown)


scripts/real_project_tests.py
```diff
@@ -0,0 +1,113 @@
...
+]
+
+
+@click.command("init")
+def init():
+    for (project, git_url) in PROJECTS:
+        directory = os.path.join(TEST_DATA_PATH, project)
+        if not os.path.isdir(directory):
+            print("Cloning {}".format(directory))
+            subprocess.run(["git", "clone", git_url], cwd=TEST_DATA_PATH, check=True)
```
Stzx, line 33: Hey, let's add the `--depth 1` parameter here to reduce the cloning time of the project library just for testing.

Kobzol: Good point. I initially thought that the tests are testing a fixed revision, but then I realized that they are testing master, so this should be fine.

PR #5609, opened 2020-06-19
UI: update icons for Rust items

### LINKED ISSUE #5176: Use CLion-style icons for symbols?
I've noticed that we still use Java-esque `(C)` icon for structs and unions on IntelliJ (hasn't checked CLion). Perhaps it makes sense to move [these icons](https://www.jetbrains.com/help/clion/icons-and-symbols.html#structure-icons) into platform and use them instead?

Undin:
Now, we don't use Java-like icons and use icons with style similar to C/C++.
Also, provide the correct visibility icon. Note, if an item inherits visibility from its owner (trait's and impl's items, enum variants), we don't show icon because it makes a little sense

Example of structure view:
| Before | After |
|-|-|
| image | image |

Closes #5176


Undin:
Code for testing:
```rust
const CONSTANT: i32 = 123;
static STATIC: i32 = 123;
static mut MUT_STATIC: i32 = 123;

struct TupleStruct(i32);
struct Struct {
    field: i32
}
struct Union {
    f1: i32,
    f2: f32
}

enum Enum {
    EnumVariant,
    TupleEnumVariant(i32),
    StructEnumVariant {
        field: i32
    },
}

type TypeAlias = i32;
fn function() {}

mod module {}
mod module_declaration;

pub trait Trait {
    const ABSTRACT_ASSOC_CONSTANT: i32;
    const ASSOC_CONSTANT: i32 = 123;

    type AbstractAssocTypeAlias;
    type AssocTypeAlias = i32;

    fn abstract_method(&self);
    fn method(&self) {}

    fn abstract_assoc_function();
    fn assoc_function() {}
}

impl Trait for Struct {
    const ABSTRACT_ASSOC_CONSTANT: i32 = 0;
    type AbstractAssocTypeAlias = ();

    fn abstract_method(&self) {
        unimplemented!()
    }

    fn abstract_assoc_function() {
        unimplemented!()
    }
}

fn main() {}
```

mchernyavsky:
I think we shouldn't add a visibility icon for fields of enum variants because they can't be private.

Undin:
@mchernyavsky done

Issue #5606, opened 2020-06-18
More precise tracking of missing/extraneous function arguments


Kobzol:
## Environment
* **IntelliJ Rust plugin version:** 0.2.125.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.44.0 (49cae5576 2020-06-01) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** Linux 5.0.0-37-generic

## Problem description
The plugin currently annotates the whole function call argument list with an error if the number of arguments is too low or too high.

```rust
fn foo(a: u32) {}
fn bar() {
    foo<error>()</error>;       // too few arguments
    foo<error>(1, 2, 3)</error>;   // too many arguments
}
```

In Kotlin, it only annotates the extraneous arguments if there are too many, or the closing parenthesis if there are not enough arguments.

```kotlin
fun foo(a: Int) {}
fun bar() {
    foo(<error>)</error>;                         // too few arguments
    foo(1, <error>2</error>, <error>3</error>);   // too many arguments
}
```

Rustc just marks the function name and doesn't give any specific ranges, so I don't think that we have to follow it too closely. If it's a problem that there would be multiple `E0061` errors emitted per function call, we can create a different highlight for the whole function call and then specific error annotations for each extraneous argument.

I think that it would be nicer if we followed Kotlin's behaviour. It would provide more precise highlighting and this quick fix https://github.com/intellij-rust/intellij-rust/issues/3582 would be easier to use.

If you agree, I'll send a PR for the highlighting and after that for the mentioned quick fix.

PR #5612, opened 2020-06-19
ANN: use more precise error annotations for missing/extraneous function call arguments

Kobzol:
This PR implements more precise tracking of missing/extraneous function arguments as proposed in https://github.com/intellij-rust/intellij-rust/issues/5606.

image

I think that this kind of highlighting is also more consistent with arguments that have a wrong type, as in that case also only the single argument is highlighted, and not the whole function call.

On top of this more fine-grained annotation we can add new quick fixes:
- fill missing arguments with defaults
- remove extraneous argument
- add missing parameter to function declaration

Even though all of them could be implemented even with the old, coarse annotation, I think that with this change the fixes would be easier to use for specific arguments.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5606

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -512,7 +512,7 @@ sealed class RsDiagnostic(
         }
     }
 
-    class TooFewParamsError(
+    class IncorrectArgumentCountErrorVariadic(
         element: PsiElement,
         private val expectedCount: Int,
         private val realCount: Int
@@ -530,7 +530,7 @@ sealed class RsDiagnostic(
         }
     }
 
-    class TooManyParamsError(
+    class IncorrectArgumentCountError(
         element: PsiElement,
         private val expectedCount: Int,
         private val realCount: Int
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -512,7 +512,7 @@ sealed class RsDiagnostic(
         }
     }
 
-    class TooFewParamsError(
+    class IncorrectArgumentCountErrorVariadic(
```
mchernyavsky, line 515: Maybe it's better to merge `IncorrectArgumentCountErrorVariadic` with `IncorrectArgumentCountError`? Something like this:
```kotlin
class IncorrectArgumentCountError(
    element: PsiElement,
    private val expectedCount: Int,
    private val realCount: Int,
    private val variadic: Boolean
) : RsDiagnostic(element) {
    override fun prepare() = PreparedAnnotation(
        ERROR,
        if (variadic) E0060 else E0061,
        errorText()
    )

    private fun errorText(): String {
        return "This function takes${if (variadic) " at least" else ""}" +
            " $expectedCount ${pluralise(expectedCount, "parameter", "parameters")}" +
            " but $realCount ${pluralise(realCount, "parameter", "parameters")}" +
            " ${pluralise(realCount, "was", "were")} supplied"
    }
}
```

Issue #2122, opened 2017-12-12
Find trait method implementations


dbrgn:
I hope this wasn't reported before, I didn't find an issue for it.

On a trait, it's possible to list all impls:

![img](https://tmp.dbrgn.ch/screenshots/20171212111004-n_fbog0l.png)

But this isn't available on a trait method:

![img](https://tmp.dbrgn.ch/screenshots/20171212111032-ygychlkd.png)

If I'm not mistaken, IntelliJ supports this for abstract methods in Java. Would be awesome to have this in Rust too :)

dbrgn:
Interesting: The Rust plugin correctly detects that the implementing method implements this trait:

![img](https://tmp.dbrgn.ch/screenshots/20171212111232-vsdca_61.png)

So it's possible to navigate up in the hierarchy, but not back down.

kumbayo:
I am looking into this (already have a prototype)

tchalamarla:
while at it, is it possible to generate code for all functions if 
we type
impl From<u8> for X {
}

can it place dummy functions like
impl From<u8> for X {
    fn from<u8>(d: u8) -> Self {
          unreachable!() /unimplemented!()
    }
}


sorry if this is not the right place for this

kumbayo:
@tchalamarla This feature sort of already exists.
Just write 
```rust
impl From for X {
}
```
The first line will be underlined with a red squiggly line to mark missing items.
Just press Alt-Enter and choose `implement members`

Issue #2849, opened 2018-09-14
Make "Go to implementations" work better


dvtomas:
Traits and types (structs, enums) with impl blocks have a ⓘ↓ icon next to them. After clicking it one can choose to go to a particular implementation. That is cool. However:

1) Even traits and types that don't actually have an implementation display the icon, with tooltip saying "Has implementations". It would be nice not to display the icon for these.

2)  Ctrl + Alt + B (Go to implementations) works only when the cursor is positioned on the trait/type identifier. Example: 
```rust
pub trait FooTrait {
 ...
}
```
Ctrl + Alt + B works only with cursor on FooTrait. It would be great if it worked on the whole line, including the *pub* and *trait* keywords and the opening bracket. I even thought the shortcut was broken before I discovered this behavior.

3) Trait method implementations in impl blocks have a nice ⓘ↑ icons next to them, with tooltip saying "Implements method in FooTrait". That is cool. However, the opposite direction does not work. Individual trait methods don't have any ⓘ↓ icon next to them, and "Go to implementations" is not implemented for them.

Kobzol:
@Undin I think that 1) and 2) cannot be supported because of performance reasons? `RsImplsLineMarkerProvider.kt:{34, 55}`.

vlad20012:
1) Find usages become much faster during the last several years, so I think the point about the performance should be revisited now. But it should be implemented very carefully, i.e. it's not just uncommenting `if (query.isEmptyQuery) return null` line. We should follow Java's approach. I'm going to try...

vlad20012:
2) This point is about `Ctrl + Alt + B`, so the performance doesn't matter. This can be done :thinking:

Kobzol:
Hmm, I looked at 2) and in Kotlin it works the same way as in the Rust plugin. When `Ctrl + Alt + B` is invoked on `trait` (or `interface` in Kotlin), the GoTo query processor is not even invoked. Maybe this should be a platform ticket?

vlad20012:
Hmm, looks like it works for me :thinking: 

![Peek 2020-08-03 14-23](https://user-images.githubusercontent.com/3221931/89177839-f75a7c80-d594-11ea-8fe3-8a840843e7f5.gif)

Kobzol:
Oh, if the interface has no implementations, then it only works on the name, if it has implementations, then it also works on `interface`.

Kobzol:
[...]

PR #5613, opened 2020-06-19
NAV: allow navigation from RsAbstractable to its implementations

Kobzol:
This is a revival of an old PR by @kumbayo (https://github.com/intellij-rust/intellij-rust/pull/2163), which allows navigation from functions/constants/associated types to their corresponding implementations.

The corresponding implementation place is rendered as `<Trait for Type>::function`.

Closes: https://github.com/intellij-rust/intellij-rust/pull/2163
Fixes: https://github.com/intellij-rust/intellij-rust/issues/2122
Part of: https://github.com/intellij-rust/intellij-rust/issues/2849

### DIFF
--- a/src/main/kotlin/org/rust/ide/lineMarkers/RsImplsLineMarkerProvider.kt
+++ b/src/main/kotlin/org/rust/ide/lineMarkers/RsImplsLineMarkerProvider.kt
@@ -8,16 +8,18 @@ package org.rust.ide.lineMarkers
 import com.intellij.codeInsight.daemon.LineMarkerInfo
 import com.intellij.codeInsight.daemon.LineMarkerProvider
 import com.intellij.codeInsight.navigation.NavigationGutterIconBuilder
+import com.intellij.ide.util.DefaultPsiElementCellRenderer
+import com.intellij.openapi.util.Condition
 import com.intellij.openapi.util.NotNullLazyValue
 import com.intellij.psi.PsiElement
 import com.intellij.util.Query
 import org.rust.ide.icons.RsIcons
 import org.rust.lang.core.psi.RsEnumItem
-import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.RsStructItem
 import org.rust.lang.core.psi.RsTraitItem
-import org.rust.lang.core.psi.ext.searchForImplementations
-import org.rust.lang.core.psi.ext.union
+import org.rust.lang.core.psi.ext.*
+import org.rust.openapiext.filterQuery
+import org.rust.openapiext.mapQuery
 
 /**
  * Annotates trait declaration with an icon on the gutter that allows to jump to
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/navigation/goto/RsGotoTargetRendererProvider.kt
```diff
@@ -20,8 +21,11 @@ class RsGotoTargetRendererProvider : GotoTargetRendererProvider {
         if (!gotoData.hasDifferentNames() && targetsInImpl) {
             return GoToImplRenderer()
         }
-        if (element is RsImplItem) return ImplRenderer()
-        return null
+        return when (element) {
+            is RsImplItem,
+            is RsAbstractable -> NamedElementRenderer()
+            else -> null
+        }
```
vlad20012, line 28: This change causes ClassCastException when typing `Ctrl + Alt + B` on a trait.

Also, I don't understand why this is needed because we already have fine rendering for 
`Ctrl + Alt + B` implementation variants
image

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -66,8 +66,27 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
...
+                        append("$traitName for ")
+                    }
+                    append(typeName)
+                    append(">::${psi.name}")
+                    append(typeParameterBounds(o.impl))
+                }
+            } else {
+                psi.name
+            }
+        }
```
vlad20012, line 89: This change causes test failures. And I don't think it's reasonable to change structure view rendering. It's possible to change renderer of implementation variants with `NavigationGutterIconBuilder.setCellRenderer` (in `RsImplsLineMarkerProvider.collectSlowLineMarkers`, see `info` variable)

src/main/kotlin/org/rust/ide/navigation/goto/RsGotoTargetRendererProvider.kt
Kobzol: You're right, this doesn't seem to be needed.

src/main/kotlin/org/rust/ide/presentation/Utils.kt
Kobzol: Thanks, I changed it so that it displays the presentation of the abstractable's owner.

Kobzol:
When you use `Ctrl + Alt + B`, it says `Choose implementations of <abstractable/trait name>`. When you click on Go to, it only shows `Go to implementation`, so when I have a trait like this:

```rust
trait T {              // X0
    fn foo(&self);     // X1
}
```
If you click on X0/X1, it will now show the same set of targets and it might now be obvious on what have you clicked. Maybe we could change the title to says `Go to implementation of T`/`Go to implementation of foo`?

src/main/kotlin/org/rust/ide/lineMarkers/RsImplsLineMarkerProvider.kt
```diff
@@ -42,21 +42,44 @@ class RsImplsLineMarkerProvider : LineMarkerProvider {
...
+                .setCellRenderer(GoToImplRenderer())
                 .createLineMarkerInfo(el)
 
             result.add(info)
         }
     }
 
     companion object {
-        fun implsQuery(psi: PsiElement): Query<RsImplItem>? {
+        private class GoToImplRenderer: DefaultPsiElementCellRenderer() {
```
vlad20012, line 53: Can be moved outside of the companion object =D
(just to RsImplsLineMarkerProvider nested class)

Kobzol:
Fixed tests and added `Go to implementation of <element.text>`. Since the marker is provided for leaf elements (identifiers), I used the text of the identifier as the name.

src/test/kotlin/org/rust/ide/lineMarkers/RsMethodLineSeparatorProviderTest.kt
```diff
@@ -27,14 +27,14 @@ class RsMethodLineSeparatorProviderTest : RsLineMarkerProviderTestBase() {
...
         doTest("""
             trait Foo { // - Has implementations
-                const C1: i32 = 1;
-                fn f1();
-                fn f2();
-                fn f3() { // - null
+                const C1: i32 = 1; // - Has implementations
+                fn f1(); // - null,Has implementations
+                fn f2(); // - null,Has implementations
+                fn f3() { // - Has implementations
```
Kobzol, line 33: I'm not really sure what is happening here and why the null has disappeared. I wonder if the comments can somehow affect the behaviour of `RsMethodLineSeparatorProvider`? Maybe we should strip the `// -` comments in the test?

vlad20012: Hmm, maybe we should strip them. But also, and more reasonable, we should disable other line marker providers in these tests. Let's fix it later

Kobzol:
bors retry

Kobzol:
Yay, my first "resurrection" PR merged 🎉

PR #5614, opened 2020-06-19
INT: convert UFCS to method call

Kobzol:
This PR adds an intention that converts a UFCS to a method call.

The implemented self argument coercion is very basic, but it seems to work. If there is a better API that could support `Deref` implementations and any coercions that are not handled now, let me know.

Related: https://github.com/intellij-rust/intellij-rust/pull/5601
Fixes second part of https://github.com/intellij-rust/intellij-rust/issues/4451.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/ConvertUFCSToMethodCallIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/ConvertUFCSToMethodCallIntention.kt
@@ -1,0 +1,66 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+
+class ConvertUFCSToMethodCallIntention : RsElementBaseIntentionAction<ConvertUFCSToMethodCallIntention.Context>() {
+    override fun getText() = "Convert to method call"
+    override fun getFamilyName() = text
+
+    data class Context(val callExpr: RsCallExpr, val function: RsFunction)
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val call = element.ancestorStrict<RsCallExpr>(RsValueArgumentList::class.java) ?: return null
+        val path = (call.expr as? RsPathExpr)?.path ?: return null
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/intentions/ConvertUFCSToMethodCallIntention.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun getText() = "Convert to method call"
+    override fun getFamilyName() = text
+
+    data class Context(val callExpr: RsCallExpr, val function: RsFunction)
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val call = element.ancestorStrict<RsCallExpr>(RsValueArgumentList::class.java) ?: return null
+        val path = (call.expr as? RsPathExpr)?.path ?: return null
+        val function = path.reference?.resolve() as? RsFunction ?: return null
+        if (!function.hasSelfParameters) return null
```
vlad20012, line 24:
```suggestion
        if (!function.isMethod) return null
```

`isMethod` also check that a function is an `impl`/`trait` member

src/main/kotlin/org/rust/ide/intentions/ConvertUFCSToMethodCallIntention.kt
```diff
@@ -0,0 +1,66 @@
...
+fun normalizeSelf(selfArgument: RsExpr, function: RsFunction): RsExpr {
+    val isRef = function.selfParameter?.isRef ?: false
+    val normalized = when {
+        isRef && selfArgument is RsUnaryExpr && selfArgument.operatorType in REF_OPERATORS -> selfArgument.expr
+        else -> null
+    }
+    return normalized ?: selfArgument
+}
+
+private val REF_OPERATORS = setOf(UnaryOperator.REF, UnaryOperator.REF_MUT)
```
Undin, line 66:
```suggestion
private val REF_OPERATORS = TokenSet#create(UnaryOperator.REF, UnaryOperator.REF_MUT)
```

Kobzol: Maybe I misunderstood your suggested change, but `UnaryOperator.REF` is a Kotlin enum variant, it's not a token (`IElementType`), so it cannot be put into `TokenSet`.

Undin: Oops, you are right
Sorry, my mistake

vlad20012:
LGTM, could you please squash the commits?

vlad20012:
:+1: Thanks!
bors r+

PR #5618, opened 2020-06-20
REF: do refactoring and inplace rename within a single write command

vlad20012:
Now it's enough to press Ctrl+Z single time to undo a refactoring (previously 2 was needed)

Works for `introduce constant`, `introduce parameter` and `introduce variable` refactorings


vlad20012:
> was intentional and it was unnecessary

Yep. looks like it's not needed. I don't know why it was added :shrug:

Issue #5032, opened 2020-02-29
or_patterns support


mvlabat:
This is a feature request to support `#![feature(or_patterns)]`:
https://doc.rust-lang.org/unstable-book/language-features/or-patterns.html

## Environment

* **IntelliJ Rust plugin version:** 0.2.116.2163-193
* **Rust toolchain version:** 1.43.0-nightly (0eb878d2a 2020-02-28)
* **IDE name and version:** CLion 2019.3.4
* **Operating system:** Windows 10

## Problem description
The following pattern match example won't work:
```
Some(Foo::Bar | Foo::Baz) => {
    println!("The value contained `Bar` or `Baz`");
}
```

Kobzol:
@Undin PSA: or_patterns will be stabilized in Rust [1.53](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1530-2021-06-17), next stable release that will be out tomorrow.

Undin:
@Kobzol Thanks!

@ortem I suppose the remaining part of the support is related to your subsystems

PR #5624, opened 2020-06-22
Initial support of or_patterns

### LINKED ISSUE #3454: Support or-patterns in let and if/while let expressions
The corresponding RFC: https://github.com/rust-lang/rfcs/pull/2175
- [x] `if let A(x) | B(x)`
- [x] `while let A(x) | B(x)`
- [ ] `let A(x) | B(x)`

Undin:
* Parse [or_patterns](https://github.com/rust-lang/rfcs/blob/master/text/2535-or-patterns.md)
* Provide name resolution and type inference
* Adapt existing code to new PSI. They should work in the same way as before in all cases supported before this PR
* Provide annotation that it's an experimental feature.
* Provide error about leading `|` in non-top level patterns

I didn't fix control flow graph and related inspections so they can work incorrectly in cases of new patterns

Fixes #3454
Fixes some points in #5032


Undin:
CC @ortem

src/main/grammars/RustParser.bnf
```diff
@@ -504,26 +504,32 @@ upper Function ::= async? const? unsafe? ExternAbi?
...
-      | PatTup
-      | PatSlice
-      | PatIdent
-      | PatMacro
-      | PatStruct
-      | PatTupleStruct
-      | (PatConst !('..' | '...' | '..='))
-      | PatRange
-      | PatBox {
+private TopPat ::= &'|' OrPat | !'!' Pat
```
vlad20012, line 507:
```suggestion
private TopPat ::= &'|' OrPat | !'!' Pat { name = "pattern" }
```
Fixes chanes in `partial/exprs.txt` parsing test

vlad20012: Also, why `!'!'` is needed? Let's add a test in `RsCompleteParsingTestCase` for this case

Undin: > Also, why !'!' is needed?

:man_facepalming:
It should be `!'|'`. But it doesn't need anyway - just artifact of development. Thanks!

Undin:
bors r=vlad20012

PR #5633, opened 2020-06-22
FIX: support tuple struct fields in MakePublicFix

Kobzol:
This PR fixes the `Make public`  quick fix to support tuple struct fields. Previously it was throwing an exception when used on a tuple struct field, because it was unconditionally casting the element to `RsNameIdentifierOwner`.


vlad20012:
:+1: 
bors r+

PR #5637, opened 2020-06-22
INT: fix `Introduce Variable` intention

vlad20012:
I just took a look at Java's `IntroduceVariableIntentionAction` and found that our intention works incorrectly.

The intention should be available in fewer places than the refactoring.

Particularly, it should be available in free exprs like:
`2 + 2;` => `let i = 2 + 2;`.

It should never display expression chooser or work with selection (like the refactoring does)

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/IntroduceLocalVariableIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/IntroduceLocalVariableIntention.kt
@@ -5,31 +5,36 @@
 
 package org.rust.ide.intentions
 
-import com.intellij.openapi.actionSystem.DataContext
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
-import com.intellij.psi.PsiFile
-import org.rust.ide.refactoring.RsBaseEditorRefactoringAction
-import org.rust.ide.refactoring.introduceVariable.RsIntroduceVariableHandler
+import org.rust.ide.refactoring.introduceVariable.extractExpression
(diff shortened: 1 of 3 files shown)


Kobzol:
Oh, seems like we were attempting a similar thing :-D https://github.com/intellij-rust/intellij-rust/pull/5573
My PR probably used a wrong approach (see discussion in that PR), should I close it?

vlad20012:
@Kobzol At first glance, it looks like your PR is about refactoring, but my PR is only about intention action. I'll take a closer look tomorrow.

Kobzol:
I created the intention based on the refactoring without any changes, which probably wasn't ideal. Then I found that the refactoring does not work for standalone expressions (`2 + 2;`), so I changed the refactoring (and because the intention uses it, also the intention) to support that case. But there are some edge cases that are not handled properly now, as discussed in that PR.

Btw in Kotlin the intention does display expression chooser and it works with selections.

vlad20012:
> Btw in Kotlin the intention does display expression chooser and it works with selections.

[Kotlin's version](https://github.com/JetBrains/kotlin/blob/f4e9acb2337a4ab11ac2a5d611faa99e841442ab/idea/src/org/jetbrains/kotlin/idea/intentions/IntroduceVariableIntention.kt#L25) looks very similar to my version :)
And, well, I don't see that it displays expression chooser or it works with selections. I mean not chooser for replacement (multiple occurrences), but a chooser for an extracting expression.

I think my changes are unrelated to #5573

bors r+

Kobzol:
Ah, yes, I was talking about occurrence chooser, not expression chooser.

PR #5641, opened 2020-06-23
ANN: Fix custom macro attr (name conflicts)

Stzx:
https://github.com/intellij-rust/intellij-rust/pull/5582#discussion_r443247691

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -26,11 +26,13 @@ import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.core.*
 import org.rust.lang.core.FeatureAvailability.CAN_BE_ADDED
 import org.rust.lang.core.FeatureAvailability.NOT_AVAILABLE
+import org.rust.lang.core.RsPsiPattern.STD_ATTRIBUTES
 import org.rust.lang.core.macros.MacroExpansionMode
 import org.rust.lang.core.macros.macroExpansionManager
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.IDENTIFIER
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.MACROS
 import org.rust.lang.core.resolve.Namespace
 import org.rust.lang.core.resolve.knownItems
 import org.rust.lang.core.resolve.namespaces
@@ -777,7 +779,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
 
     // E0132: Invalid `start` attribute
     private fun checkStartAttribute(holder: RsAnnotationHolder, attr: RsAttr) {
-        if (attr.metaItem.name != "start") return
+        if (!attr.isBuiltinWithName("start")) return
 
         START.check(holder, attr.metaItem, "#[start] function")
 
@@ -817,7 +819,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
     }
 
     private fun checkReprAttribute(holder: RsAnnotationHolder, attr: RsAttr) {
-        if (attr.metaItem.name != "repr") return
+        if (!attr.isBuiltinWithName("repr")) return
 
         val owner = attr.owner ?: return
 
@@ -861,14 +863,14 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
 
     // E0518: Inline attribute is allowed only on functions
     private fun checkInlineAttr(holder: RsAnnotationHolder, attr: RsAttr) {
-        val metaItem = attr.metaItem
-        if (metaItem.name == "inline") {
-            val parent = attr.owner
+        if (!attr.isBuiltinWithName("inline")) return
 
-            if (parent !is RsFunction && parent !is RsLambdaExpr) {
-                RsDiagnostic.IncorrectlyPlacedInlineAttr(metaItem.path?.referenceNameElement ?: metaItem, attr)
-                    .addToHolder(holder)
-            }
+        val owner = attr.owner
+        if (owner !is RsFunction && owner !is RsLambdaExpr) {
+
+            val metaItem = attr.metaItem
+            RsDiagnostic.IncorrectlyPlacedInlineAttr(metaItem.path?.referenceNameElement ?: metaItem, attr)
+                .addToHolder(holder)
         }
     }
 
@@ -1207,4 +1209,13 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
     if (isError(ty)) {
         RsDiagnostic.SizedTraitIsNotImplemented(typeReference, ty).addToHolder(holder)
     }
+}
+
+private fun RsAttr.isBuiltinWithName(target: String): Boolean {
+    val name = metaItem.name ?: return false
+
+    if (name != target) return false
+    if (name !in STD_ATTRIBUTES) return false
+
+    return !hasInScope(name, MACROS)
 }
--- a/src/main/kotlin/org/rust/lang/core/RsPsiPattern.kt
+++ b/src/main/kotlin/org/rust/lang/core/RsPsiPattern.kt
@@ -25,64 +25,67 @@ object RsPsiPattern {
     private val STATEMENT_BOUNDARIES = TokenSet.create(SEMICOLON, LBRACE, RBRACE)
 
     /**
-     * Source of attributes: [https://doc.rust-lang.org/1.30.0/reference/attributes.html]
+     * Source of attributes: [https://doc.rust-lang.org/1.41.1/reference/attributes.html#built-in-attributes-index]
      */
-    private val STD_ATTRIBUTES: Set<String> = setOf(
-        "crate_name",
-        "crate_type",
-        "no_builtins",
-        "no_main",
-        "no_start",
-        "no_std",
-        "recursion_limit",
-        "windows_subsystem",
+    val STD_ATTRIBUTES: Set<String> = setOf(
+        "cfg",
+        "cfg_attr",
 
-        "no_implicit_prelude",
-        "path",
+        "test",
+        "ignore",
+        "should_panic",
 
-        "link_args",
-        "link",
-        "linked_from",
+        "derive",
 
-        "link_name",
-        "linkage",
-
-        "macro_use",
-        "macro_use",
-        "macro_reexport",
         "macro_export",
-        "no_link",
-
+        "macro_use",
         "proc_macro",
         "proc_macro_derive",
         "proc_macro_attribute",
 
-        "export_name",
-        "global_allocator",
-        "link_section",
-        "no_mangle",
+        "allow",
+        "warn",
+        "deny",
+        "forbid",
 
         "deprecated",
+        "must_use",
+
+        "link",
+        "link_name",
+        "no_link",
+        "repr",
+        "crate_type",
+        "no_main",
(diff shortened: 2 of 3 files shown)


src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -3456,4 +3456,52 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
...
+        use proc_macro::TokenStream;
+
+        #[proc_macro_attribute]
+        pub fn repr(attr: TokenStream, item: TokenStream) -> TokenStream {
+            item
+        }
+    //- main.rs
+        use dep_proc_macro::repr;
+
+        #[repr/*caret*/(C)]
```
Stzx, line 3472: The local test will fail (the error is still reported), am I missing something?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1190,3 +1191,8 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
         RsDiagnostic.SizedTraitIsNotImplemented(typeReference, ty).addToHolder(holder)
     }
 }
+
+// When `resolvePath()` parses the value, it means this is a custom proc macro (It currently looks like this
+private fun RsAttr.checkMetaItemName(target: String): Boolean = (metaItem.name == target && metaItem.resolvePath() == null)
```
Undin, line 1196: Maybe `isBuiltInAttribute` or something similar is better?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1190,3 +1191,8 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
...
+
+private fun RsMetaItem.resolvePath(): RsElement? = path?.reference?.resolve()
```
Undin, line 1198: Do we really need it? Maybe inline?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -844,13 +844,14 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
     private fun checkInlineAttr(holder: RsAnnotationHolder, attr: RsAttr) {
         val metaItem = attr.metaItem
-        if (metaItem.name == "inline") {
-            val parent = attr.owner
 
-            if (parent !is RsFunction && parent !is RsLambdaExpr) {
-                RsDiagnostic.IncorrectlyPlacedInlineAttr(metaItem.path?.referenceNameElement ?: metaItem, attr)
-                    .addToHolder(holder)
-            }
+        if (metaItem.name != "inline") return
```
Undin, line 848: Why do you not use `checkMetaItemName` here?

Stzx: For attributes like `inline`, using 
```rust
//- dep-proc-macro/lib.rs
    use proc_macro::TokenStream;

    #[proc_macro_attribute]
    pub fn inline(attr: TokenStream, item: TokenStream) -> TokenStream {
        item
    }
//- main.rs
    extern crate dep_proc_macro;

    use dep_proc_macro::inline;

    #[inline] // ambiguity
    type Foo = i32;
```
directly will actually lead to `E0659`.

For other explicit statements, 
```rust
//- main.rs
    extern crate dep_proc_macro;

    #[dep_proc_macro::inline] // metaItem.name == null
    type Foo = i32;
```
```rust
//- main.rs
    extern crate dep_proc_macro;

    use dep_proc_macro::inline as test;

    #[test] // metaItem.name == test
    type Foo = i32;
```

the plug-in currently does not resolve the name into `inline`, so only `metaItem.name != "inline"` is used here. I think it is enough.

Stzx:
I re-examined it. In fact, as described by https://github.com/intellij-rust/intellij-rust/pull/5641#discussion_r449950120, other built-in attributes have this phenomenon, so it seems that this problem does not need to be fixed.

@Undin what do you think?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
Undin: > directly will actually lead to E0659.

Yep, I'm aware about it. Moreover, https://github.com/intellij-rust/intellij-rust/pull/5582#discussion_r443247691 says about it directly (see code example). But it's exactly what we want - we shouldn't try to provide any analysis if there is `E0659` for the particular attribute. Yep, currently, we don't handle `E0659` but it will be handled eventually

> the plug-in currently does not resolve the name into inline

Yeah, our name resolution doesn't work properly in this case as well. But it doesn't make your changes worse, i.e. your current changes fix some rare false-positive errors and I think it's ok that we don't handle all similar false-positive errors because of the bug in name resolution. But we don't provide other false-positive errors so it's definitely an improvement

Stzx:
@Undin When I cleaned and moved `RsPsiPattern.STD_ATTRIBUTES`, I found that some of the built-in properties no longer exist in the new version. Do I need to clean it or keep it? (E.g `link_args`...)

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
```diff
@@ -8,15 +8,86 @@ package org.rust.lang.core.psi.ext
...
 import org.rust.lang.core.psi.RsMetaItem
 import org.rust.lang.core.psi.RsOuterAttr
+import org.rust.lang.core.resolve.MACROS
 
 interface RsAttr : RsElement {
     val metaItem: RsMetaItem
 }
 
+/**
+ * Source of attributes: [https://doc.rust-lang.org/1.44.1/reference/attributes.html#built-in-attributes-index]
```
Stzx, line 18: I updated the list from `1.30.0` to `1.44.1` and removed some names that no longer exist in the new version.

Stzx:
ping @Undin

src/main/kotlin/org/rust/lang/core/RsPsiPattern.kt
```diff
@@ -25,64 +25,64 @@ object RsPsiPattern {
...
-    private val STD_ATTRIBUTES: Set<String> = setOf(
-        "crate_name",
-        "crate_type",
-        "no_builtins",
-        "no_main",
-        "no_start",
-        "no_std",
-        "recursion_limit",
-        "windows_subsystem",
+    val STD_ATTRIBUTES: Set<String> = setOf(
```
Stzx, line 30: For built-in attribute that have been removed in the new version (E.g `linked_from`).  keep it, but it is better to provide detection only for toolchains with this attribute?

Undin: I don't think that there is a big difference because we don't have any checks related to such attributes, so we can use any version of this list currently

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
```diff
@@ -20,3 +22,11 @@ val RsAttr.owner: RsDocAndAttributeOwner?
         // Throw exception so that any problems with this property aren't silent and are instead easily findable
         else -> error("Unsupported attribute type: $this")
     }
+
+val RsAttr.isBuiltInAttribute: Boolean
+    get() {
+        val name = metaItem.name ?: return false
+        if (name !in STD_ATTRIBUTES && name != "start") return false
```
Undin, line 29: Can we put `start` in `STD_ATTRIBUTES` list?

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
```diff
@@ -20,3 +22,11 @@ val RsAttr.owner: RsDocAndAttributeOwner?
...
+
+        return !this.hasInScope(name, MACROS)
```
Undin, line 31: `this` is redundant here

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
```diff
@@ -20,3 +22,11 @@ val RsAttr.owner: RsDocAndAttributeOwner?
         // Throw exception so that any problems with this property aren't silent and are instead easily findable
         else -> error("Unsupported attribute type: $this")
     }
+
+val RsAttr.isBuiltInAttribute: Boolean
```
Undin, line 26: I think we should rename it somehow because in case of collision it's still builtin attribute

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1208,3 +1208,5 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
         RsDiagnostic.SizedTraitIsNotImplemented(typeReference, ty).addToHolder(holder)
     }
 }
+
+private fun RsAttr.checkNameAndIsBuiltIn(target: String): Boolean = (metaItem.name == target && isBuiltInAttribute)
```
Undin, line 1212: Personaly, I'd merge it with `isBuiltInAttribute` and name it `isBuiltinWithName` (or something similar)

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
Stzx: Done.

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
Stzx: `start` currently exists in the unstable list, I think maybe it can?

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
Stzx: do you mean `metaItem.name`?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
Stzx: Okay, I will merge it and move it to `RsAttr.kt`

src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
```diff
@@ -20,3 +22,14 @@ val RsAttr.owner: RsDocAndAttributeOwner?
...
+fun RsAttr.isBuiltinWithName(target: String): Boolean {
+    return if (metaItem.name != target) false
+    else {
+        val name = metaItem.name ?: return false
+
+        if (name !in STD_ATTRIBUTES) false
+        else
+            !hasInScope(name, MACROS)
+    }
+}
```
Undin, line 35: nit: I think we can slightly rewrite it to improve readability
```kotlin
fun RsAttr.isBuiltinWithName(target: String): Boolean {
    val name = metaItem.name ?: return false
    if (name != target) return false
    if (name !in STD_ATTRIBUTES) return false
    return !hasInScope(name, MACROS)
}
```

Also, I think we should move it into `RsErrorAnnotator.kt` file and make it private - unlikely we need it somewhere else

Stzx: Done.

Stzx:
bors r=Undin

PR #5644, opened 2020-06-23
ANN: add quick fix to fill missing arguments in function call

Kobzol:
This PR adds a quick fix that fills missing arguments in function calls with default values.

fill

I also wanted to support generic type parameter substitution, like this:
```rust
struct S<T>(T);
impl<R> S<R> {
    fn foo(&self, t: R) {}
}
fn foo(s: S<u32>) {
    s.foo(/*caret*/);
    vvv
    s.foo(0);
}
```
But I don't know how to substitute the type parameters (my naive attempt doesn't work). @vlad20012 is there a straightforward way of doing that?

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
@@ -1,0 +1,107 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.parentOfType
+import org.rust.ide.annotator.expectedParamsCount
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.selfType
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.type
+import org.rust.openapiext.buildAndRunTemplate
+import org.rust.openapiext.createSmartPointer
+
+class FillFunctionArgumentsFix(element: PsiElement) : LocalQuickFixAndIntentionActionOnPsiElement(element) {
+    override fun getText(): String = "Fill missing arguments"
+    override fun getFamilyName(): String = text
+
+    override fun invoke(
+        project: Project,
+        file: PsiFile,
+        editor: Editor?,
+        startElement: PsiElement,
+        endElement: PsiElement
+    ) {
+        val arguments = startElement.parentOfType<RsValueArgumentList>(true) ?: return
+        val parent = arguments.parent as? RsElement ?: return
+
+        val parameters = getParameterTypes(parent) ?: return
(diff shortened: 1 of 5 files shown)


vlad20012:
> But I don't know how to substitute the type parameters (my naive attempt doesn't work). @vlad20012 is there a straightforward way of doing that?

No, unfortunately, there is no good way for this. I'll try to add such info into the inference result.

Kobzol:
Okay. Should we wait for that? I think that this PR can be useful even without such functionality.

vlad20012:
Then no. But some tests are failed 😅

Kobzol:
I reverted the change in default value builder for type parameters, it could introduce type parameter names in contexts where they would not exist or have a different meaning.

Kobzol:
@vlad20012 Maybe we can merge it in the current shape, without support for the advanced case with generic parameters?

vlad20012 (approved):
I have no idea why this is not merged yet. Let's r+ it!

Kobzol:
Great :)
bors r=vlad20012

vlad20012:
bors retry

src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        val insertedArguments = inserted.exprList.drop(actualArgumentCount)
+        editor?.buildAndRunTemplate(inserted, insertedArguments.map { it.createSmartPointer() })
+    }
+}
+
+private fun getParameterTypes(element: PsiElement): List<Ty?>? {
+    return when (element) {
+        is RsCallExpr -> {
+            when (val target = (element.expr as? RsPathExpr)?.path?.reference?.resolve()) {
```
Undin, line 68: Can we just infer type of `element.expr`?
It should have `TyFunction` type with actual types if they can be inferred at all.
It should allow us to:
- support generic cases (#6645)
- support closures cases (#6612)
- support quite rare but still possible cases when "function pointer" is stored in non local variable
```rust
struct S(i32);
const FOO: fn(i32) -> S = S;
fn main() {
    FOO();
}
```

The latest case won't work currently because we don't annotate such cases be the corresponding annotator, but it will automatically be supported when we improved the corresponding diagnostic itself

Kobzol: I tried it and it works fine for function calls and closures. However, for methods, I can't seem to get the proper type substitution.

```rust
// test
struct S<T, R>(T, R);
impl<T, R> S<T, R> {
    fn foo(&self, a: T, b: R) {}
}
fn foo(s: S<u32, ()>) {
    s.foo(/*caret*/);
}
```

```kotlin
// code (getParameterTypes)
is RsMethodCall -> {
    val receiver = element.receiver.type // S<u32, ()>
    val method = element.reference.resolve() as? RsFunction ?: return null
    val fnType = method.type.substitute(receiver.parameterValues)
    // fnType here does not have properly substituted types
}
```
I realize that the substitution fails because the `T` and `R` type parameters are different objects in `receiver` and `method`. Is there a way to get around it?

Undin: > I realize that the substitution fails because the T and R type parameters are different objects in receiver and method. Is there a way to get around it?

Not sure that it's simple. Currently, such inference is performed inside `org.rust.lang.core.types.infer.RsTypeInferenceWalker#inferMethodCallExprType`

I suppose that we can improve the quick-fix for `RsCallExpr` and `RsMethodCall` separately.

Kobzol: Ok, so let's merge https://github.com/intellij-rust/intellij-rust/pull/6613 with improvements for closures and generic function calls only now.

PR #5645, opened 2020-06-24
COMP: Support function and struct complete current statement

### LINKED ISSUE #5587: Feature request: Complete current statement auto add brackets when it's function/struct
Hello,

one of the things I really love when writing Kotlin/Java/Typescript/Js/Python in IntelliJ is the "complete current statement" command. I use it a lot. The default behavior in other languages is that when I declare a function/class and call "complete current statement", it adds a bracket pair and jump to a new line for me. For some reason this is missing in rust IntelliJ. It would be real nice to have that functionality.

Just to make sure it's clear: I would like that when I type:
`fn someFunction`
and then when I hit "complete current statement" it would become
````
fn someFunction(){
 
}
````
with the cursor in the function body, the same for struct. The current behavior is that it just jumps to a new line (without brackets). I don't think that's helpful in the case of function/struct.

### PR DESCRIPTION
**Pull request by Stzx**
Closed #5587

### DIFF
--- a/src/main/kotlin/org/rust/ide/typing/assist/AfterFunctionOrStructEnterProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/typing/assist/AfterFunctionOrStructEnterProcessor.kt
@@ -1,0 +1,40 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.typing.assist
+
+import com.intellij.lang.SmartEnterProcessorWithFixers
+import com.intellij.openapi.actionSystem.IdeActions
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.actionSystem.EditorActionManager
+import com.intellij.openapi.editor.ex.EditorEx
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsStructItem
+import org.rust.lang.core.psi.ext.endOffset
+
+class AfterFunctionOrStructEnterProcessor : SmartEnterProcessorWithFixers.FixEnterProcessor() {
+    override fun doEnter(atCaret: PsiElement, file: PsiFile, editor: Editor, modified: Boolean): Boolean {
+        if (!modified) return false
+
+        return when (val parent = atCaret.parent) {
+            is RsFunction, is RsStructItem -> {
+                val elementEndOffset = parent.endOffset - 2
+                editor.caretModel.moveToOffset(elementEndOffset)
+                plainEnter(editor)
+                modified
+            }
+            else -> false
+        }
+    }
+
+    override fun plainEnter(editor: Editor) {
+        if (editor !is EditorEx) return
+
+        val enterHandler = EditorActionManager.getInstance().getActionHandler(IdeActions.ACTION_EDITOR_ENTER)
+        enterHandler.execute(editor, editor.caretModel.currentCaret, editor.dataContext)
+    }
+}
--- a/src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt
+++ b/src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt
@@ -1,0 +1,60 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.typing.assist
+
+import com.intellij.lang.SmartEnterProcessorWithFixers
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsElementTypes.*
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsStructItem
+import org.rust.lang.core.psi.ext.body
+import org.rust.lang.core.psi.ext.elementType
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.lang.core.psi.ext.returnType
+import org.rust.lang.core.types.ty.TyUnknown
+
+class FunctionOrStructFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
+
+    override fun apply(editor: Editor, processor: RsSmartEnterProcessor, element: PsiElement) {
+        if (element.elementType !in APPROVED_TYPES) return
+
+        val parent = element.parent
+        val prefix = when {
+            parent is RsFunction ->
+                when {
+                    parent.body != null -> null
+
+                    /**
+                     * fn foo/*caret*/
+                     */
+                    parent.valueParameterList == null -> "()"
+
+                    /**
+                     * fn foo(a: i32, b: i32/*caret*/)
+                     * fn foo() -> i32/*caret*/
+                     */
+                    parent.returnType !is TyUnknown -> ""
+
+                    else -> null
+                }
+
+            parent.isStructOrUnionAndIdentifier() -> ""
+
+            else -> null
+        } ?: return
+
+        editor.document.insertString(parent.endOffset, "$prefix { }")
+    }
+
+    // struct Foo/*caret*/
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+import com.intellij.lang.SmartEnterProcessorWithFixers
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.RsElementTypes.IDENTIFIER
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsStructItem
+import org.rust.lang.core.psi.ext.body
+import org.rust.lang.core.psi.ext.elementType
+
+class FunctionOrStructFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
```
Stzx, line 17: Is this name ok?

src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt
```diff
@@ -0,0 +1,38 @@
...
+        if (element.elementType != IDENTIFIER) return
+
+        val parent = element.parent
+
+        val content = when {
+            parent is RsFunction && parent.valueParameterList == null && parent.body == null ->
+                "() {}"
+
+            parent is RsStructItem && parent.tupleFields == null && parent.blockFields == null ->
+                " {}"
```
mchernyavsky, line 29: For example, in Kotlin, smart enter adds newline:
```kotlin
// Before
class C/*caret*/

//////////////////////

// After
class C {
    /*caret*/
}


```

src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class FunctionOrStructFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
+
+    override fun apply(editor: Editor, processor: RsSmartEnterProcessor, element: PsiElement) {
+        if (element.elementType != IDENTIFIER) return
+
+        val parent = element.parent
+
+        val content = when {
+            parent is RsFunction && parent.valueParameterList == null && parent.body == null ->
```
mchernyavsky, line 25: Why it shouldn't work in case of `parent.valueParameterList != null`? For example, I would expect something like this:

```rust
// Before
fn foo(a: i32, b: i32/*caret*/)

//////////////////////

// After
fn foo(a: i32, b: i32) {
    /*caret*/
}
```

src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt
```diff
@@ -0,0 +1,38 @@
...
+                "() {}"
+
+            parent is RsStructItem && parent.tupleFields == null && parent.blockFields == null ->
+                " {}"
+
+            else -> return
+        }
+
+        val offset = parent.textRange.endOffset
```
mchernyavsky, line 34:
```suggestion
        val offset = parent.endOffset
```

src/main/kotlin/org/rust/ide/typing/assist/AfterFunctionOrStructEnterProcessor.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsStructItem
+
+class AfterFunctionOrStructEnterProcessor : SmartEnterProcessorWithFixers.FixEnterProcessor() {
+    override fun doEnter(atCaret: PsiElement, file: PsiFile, editor: Editor, modified: Boolean): Boolean {
+        return when (val parent = atCaret.parent) {
+            is RsFunction, is RsStructItem -> {
+                val elementEndOffset = parent.textRange.endOffset - 1
```
mchernyavsky, line 19:
```suggestion
                val elementEndOffset = parent.endOffset - 1
```

Stzx:
Several hours have passed and the last test is still running 😅

Undin:
> Several hours have passed and the last test is still running 😅

@Stzx For some reason windows tests on Github Actions sometimes hangs. I haven't investigated why it happens yet. Note, we don't use GitHub Action CI to merge PRs yet. So you can look only at travis check for now

src/main/kotlin/org/rust/ide/typing/assist/FunctionOrStructFixer.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.rust.lang.core.psi.ext.endOffset
+
+class FunctionOrStructFixer : SmartEnterProcessorWithFixers.Fixer<RsSmartEnterProcessor>() {
+
+    override fun apply(editor: Editor, processor: RsSmartEnterProcessor, element: PsiElement) {
+        if (element.elementType !in APPROVED_TYPES) return
+
+        val parent = element.parent
+        val content = when {
+            parent.isFnAndIdentifier() -> "() {\n\n}"
```
Stzx, line 25: Maybe adding tabs is better? But `\t` may not be suitable.

mchernyavsky: I think `\n` is ok.

src/test/kotlin/org/rust/ide/typing/assist/RsSmartEnterProcessorTest.kt
```diff
@@ -300,6 +300,54 @@ class RsSmartEnterProcessorTest : RsTestBase() {
...
+        }
+    """)
+
+    fun `test fix union`() = doTest("""
+        union u/*caret*/
+    """, """
+        union u {
+        /*caret*/
+        }
+    """)
```
mchernyavsky, line 349: Some other cases:

```rust
// Before
fun foo(/*caret*/)

// After
fn foo() {
    /*caret*/
}
```

```rust
// Before
fn foo(/*caret*/

// After
fn foo() {
    /*caret*/
}
```

```rust
// Before
fn foo() {
    fn bar(/*caret*/)
}

// After
fn foo() {
    fn bar() {
        /*caret*/
    }
}
```

mchernyavsky: Also, you need to reformat function/struct body. Kotlin does it in `KotlinSmartEnterHandler.moveCaretInsideBracesIfAny`:
https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/editor/KotlinSmartEnterHandler.kt#L111

Stzx: > Also, you need to reformat function/struct body. Kotlin does it in `KotlinSmartEnterHandler.moveCaretInsideBracesIfAny`:
> https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/editor/KotlinSmartEnterHandler.kt#L111

After calling `reformat`, it will become like this.

```rust
// Before
fn foo(/*caret*/)

// After
fn foo() {/*caret*/}
```

Stzx: @mchernyavsky As far as the current parsing results are concerned, I suggest not to implement `fn foo(/*caret*/` `Complete the current statement`.

For example, the following expressions will generate some "inconvenient to parse" tree.
```rust
fn main() {
    fn bar(/*caret*/

    println!("Hello, world!");
}
```

```rust
fn foo(/*caret*/

fn main() {
    println!("Hello, world!");
}
```

```rust
fn foo() {
    fn bar(/*caret*/

    return;
}
```

Stzx: ping @mchernyavsky

Stzx:
Done.

Thanks for the guidance.

Issue #3720, opened 2019-04-16
Completion for lint names in attributes


Riateche:
It would be nice to add completion for lint names in attributes like `#[allow(...)]`. For example, `#[allow(de` should expand to `#[allow(dead_code)]` if the user requests completion. Rustc and clippy lints should be supported.

PR #5646, opened 2020-06-24
COMP: add completion of lint attributes

Kobzol:
This PR adds completion of lint attributes in `allow`, `deny`, `warn` and `forbid` meta items. The lints are filtered based on their path prefix, i.e. if you write `clippy::`, it will only offer you lints from `clippy`.

lint-completion

Right now on the root level it offers all lints (i.e. both `unused` and `clippy::xxx`), maybe we can only offer `clippy::` at the root level and only after completing that we can start completing `clippy` lints?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3720


src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,545 @@
...
+        is Lint.RustcGroup -> RUSTC_GROUP_PRIORITY
+        is Lint.Clippy -> CLIPPY_PRIORITY
+    }
+
+    private fun getPathPrefix(path: RsPath): String {
+        val qualifier = path.qualifier ?: return ""
+        return "${getPathPrefix(qualifier)}${qualifier.referenceName}::"
+    }
+
+    private val LINTS = setOf(
```
Undin, line 77: I'd personally move this list into a separate file just to split data and logic of completion provider.
Also, it may be useful when we implement the update

src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,545 @@
...
+        Lint.Clippy("zero_divided_by_zero"),
+        Lint.Clippy("zero_prefixed_literal"),
+        Lint.Clippy("zero_ptr"),
+        Lint.Clippy("zero_width_space"),
+        Lint.Clippy("zst_offset")
+    )
+}
+
+private sealed class Lint(val name: String) {
+    class RustcSpecific(name: String) : Lint(name)
```
Undin, line 542:
```suggestion
    class Rustc(name: String) : Lint(name)
```
nit: I think `Group` suffix is enough to distinguish them

src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,545 @@
...
+        get() {
+            return PlatformPatterns.psiElement()
+                .withLanguage(RsLanguage)
+                .withParent(RsPath::class.java)
+                .inside(
+                    psiElement<RsMetaItem>().withSuperParent(2, RsPsiPattern.lintAttributeMetaItem)
+                )
+        }
+
+    private fun getIcon(lint: Lint): Icon = when (lint) {
```
Undin, line 61: If we decide to split the current completion provider into two ones, it may be convenient to move this method into `Lint` itself

src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,545 @@
...
+        is Lint.RustcGroup -> GROUP_ICON
+        else -> RsIcons.ATTRIBUTE
+    }
+
+    private fun getPriority(lint: Lint): Double = when (lint) {
```
Undin, line 66: The same about moving method into `Lint`

src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,545 @@
...
+                                context: ProcessingContext,
+                                result: CompletionResultSet) {
+        val path = parameters.position.parentOfType<RsPath>() ?: return
+        val prefix = getPathPrefix(path)
+
+        LINTS.forEach {
+            val name = it.name
+            if (!name.startsWith(prefix)) return@forEach
+
+            val offered = name.removePrefix(prefix)
```
Undin, line 43: > Right now on the root level it offers all lints (i.e. both unused and clippy::xxx), maybe we can only offer clippy:: at the root level and only after completing that we can start completing clippy lints?

It looks quite hacky. I think it's definitely better to suggest complex lints like `clippy::xxx` in two steps.
Probably, it makes sense to split this provider into two: one for `rustc` lints and another one for `clippy`

src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,545 @@
...
+object RsLintCompletionProvider : RsCompletionProvider() {
+    private const val RUSTC_SPECIFIC_PRIORITY = 5.0
+    private const val RUSTC_GROUP_PRIORITY = 4.0
+    private const val CLIPPY_PRIORITY = 3.0
+
+    private val GROUP_ICON = RsIcons.ATTRIBUTE.multiple()
+
+    override fun addCompletions(parameters: CompletionParameters,
+                                context: ProcessingContext,
+                                result: CompletionResultSet) {
```
Undin, line 35: nit: formatting

src/main/kotlin/org/rust/lang/core/RsPsiPattern.kt
```diff
@@ -169,12 +176,17 @@ object RsPsiPattern {
             .withSuperParent(2, RsOuterAttributeOwner::class.java)
             .with("nonStdAttributeCondition") { e -> e.name !in STD_ATTRIBUTES }
 
+    val lintAttributeMetaItem: PsiElementPattern.Capture<RsMetaItem> =
```
Undin, line 179: JFYI there is `org.rust.lang.core.completion.RsPsiPatternTest` to check that patterns works expected.
We usually don't use unit test and use integration instead but since patterns can be quite complex, these tests can be helpful

Undin (commented):
In general looks great!

Could you rebase onto master, please?

src/main/kotlin/org/rust/lang/core/completion/RsLintCompletionProvider.kt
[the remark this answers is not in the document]
Undin: Note, I don't insist to do it in this review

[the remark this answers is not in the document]
Kobzol: Ok, I will do it in a separate PR.

Kobzol:
Actually, I decided to handle the automatic lint setup in this PR, it would just be a needless change if I messed with the lint data structure afterwards. I created a gradle task that parses Rustc lints from `rustc -W help`, however for Clippy it's a bit more complicated. There is a website with all Clippy lints, but sadly it requires Javascript to load. So possible options that I see are:
1) Use some headless browser to parse the lints (probably too heavyweight)
2) Parse the lints manually from the GitHub Clippy repository, either by HTTP requests to the individual files or by cloning the Clippy git repository.
3) Parse the lints by cloning the Clippy repository and running their Python helper script that can already parse them.

What should I use? I would prefer option 3), by creating a Python script in the `scripts` directory, which would clone the Clippy repo and use their script to get the lints.

Undin:
@Kobzol at the first glance, option `3` looks the simplest so I'd prefer this option too

Kobzol:
Ok. My notebook's disk died and I didn't backup the WIP code that I had (which was using gradle `kts` script to download the things), so I'll rewrite it from scratch in Python.

Kobzol:
Rewrote it from scratch.

build.gradle.kts
```diff
@@ -522,6 +523,50 @@ task("updateCargoOptions") {
     }
 }
 
+task("updateLints") {
+    doLast {
+        val lints = JsonSlurper().parseText("python3 fetch_lints.py".execute("scripts", print = false)) as List<Map<String, *>>
+
+        fun writeLints(path: String, lints: List<Map<String, *>>, variableName: String, prefix: String = "") {
+            val file = File(path)
+            val items = lints.sortedWith(compareBy({ !(it["group"] as Boolean) }, { it["name"] as String})).joinToString(
```
Undin, line 532: Maybe we can extract `it["group"] as Boolean` into local function? Something like
```
fun Map<String, *>.isGroup(): Boolean = get("group") as Boolean
```

And the same for `name`
Just to make code a little bit clearer and avoid code duplication - we use such casts in two places

build.gradle.kts
```diff
@@ -522,6 +523,50 @@ task("updateCargoOptions") {
...
+                separator = ",\n    "
+            ) {
+                val name = it["name"] as String
+                val isGroup = it["group"] as Boolean
+                val activePrefix = if (prefix.isNotBlank()) ", \"$prefix\"" else ""
+                "Lint(\"$name\", ${if (isGroup) "true" else "false"}$activePrefix)"
```
Undin, line 538:
```suggestion
                "Lint(\"$name\", $isGroup$activePrefix)"
```

src/main/kotlin/org/rust/lang/core/completion/lint/RsLintCompletionProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.ext.qualifier
+import org.rust.lang.core.psiElement
+import javax.swing.Icon
+
+abstract class RsLintCompletionProvider : RsCompletionProvider() {
+    private val LINT_PRIORITY = 5.0
+    protected val GROUP_PRIORITY = 4.0
+
+    protected val GROUP_ICON = RsIcons.ATTRIBUTE.multiple()
```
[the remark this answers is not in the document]
Kobzol: I had that originally, but then I couldn't access these in the child class without adding `@JvmStatic`. Is it worth it?

build.gradle.kts
```diff
@@ -522,6 +523,50 @@ task("updateCargoOptions") {
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.completion.lint
+
+val $variableName: List<Lint> = listOf(
+    $items
+)
+""".trimIndent())
```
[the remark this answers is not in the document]
Kobzol: I'll keep `trim()` to avoid an empty line at the top of the file.

src/main/kotlin/org/rust/lang/core/completion/lint/RsLintCompletionProvider.kt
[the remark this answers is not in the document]
Undin: I think yes

Kobzol:
Thanks for the review! It seems that I forgot to remove some things that weren't needed. I put the changes into a separate commit to make additional review easier.

Undin (approved):
LGTM
Thank you! It's great!
Could you do the following things before the merge, please?:
- update gif image in the description since it's a little bit outdated regarding clippy lints
- squash commits

Kobzol:
Done and done. Btw, I think that similar completion could be useful for features (both language and compiler ones), what do you think?

Undin (approved):
bors r+

Undin:
> Btw, I think that similar completion could be useful for features (both language and compiler ones), what do you think?

Yep, it may be handy for some users. We even already have a list of compiler features in the plugin and a way to update them.
But it may be harder than completion for lints because:
- it should be available only with nightly compiler otherwise you get https://doc.rust-lang.org/error-index.html#E0554. BTW this error is not handled by the plugin
- there is little sense to complete already released features because it produces compiler warning (which not handled by the plugin as well)
- we shouldn't suggest features that are already used in the project since it a https://doc.rust-lang.org/error-index.html#E0636 (not supported by the plugin too)

Kobzol:
Good point, it should probably be introduced gradually (by solving some of the substeps first).

PR #5647, opened 2020-06-24
ANN: add quick fix to change local variable type

Kobzol:
This PR adds a very simple quick fix that changes the type of a `let` declaration.

change

Umbrella issue: https://github.com/intellij-rust/intellij-rust/issues/1730

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/ConvertLetDeclTypeFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/ConvertLetDeclTypeFix.kt
@@ -1,0 +1,42 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/ConvertLetDeclTypeFix.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+class ConvertLetDeclTypeFix(
+    decl: RsLetDecl,
+    private val fixText: String,
+    private val ty: Ty
+) : LocalQuickFixAndIntentionActionOnPsiElement(decl) {
+    override fun getFamilyName(): String = "Convert type of local variable"
+    override fun getText(): String = fixText
+
+    override fun invoke(
```
mchernyavsky, line 29: I think it shouldn't be available for `let x: wrong_type = y` if `Specify type explicitly` intention can't be applied for `let x = y`. For example:

image

`let y: &impl T` will not compile.

Kobzol: Thanks. I added check for `TyAnon` and changed the text to refer only to the name of the variable and not to the binding mode (`mut`/`ref`).

PR #5648, opened 2020-06-24
RUN: add custom text for assert test diff

Kobzol:
This PR uses `printExpectedActualHeader` to use `Left` and `Right` strings for test assertion differences, instead of `Actual` and `Expected`. The order and text is still incorrect when the diff is opened, but at least now the initial text is correct.

image

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3814

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestConsoleProperties.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestConsoleProperties.kt
@@ -7,11 +7,13 @@ package org.rust.cargo.runconfig.test
 
 import com.intellij.execution.Executor
 import com.intellij.execution.configurations.RunConfiguration
+import com.intellij.execution.testframework.Printer
 import com.intellij.execution.testframework.TestConsoleProperties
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestConsoleProperties.kt
```diff
@@ -29,6 +31,14 @@ class CargoTestConsoleProperties(
         consoleProperties: TestConsoleProperties
     ): OutputToGeneralTestEventsConverter = CargoTestEventsConverter(testFrameworkName, consoleProperties)
 
+    override fun printExpectedActualHeader(printer: Printer, expected: String, actual: String) {
+        printer.print("\n", ConsoleViewContentType.ERROR_OUTPUT)
+        printer.print("Left: ", ConsoleViewContentType.SYSTEM_OUTPUT)
```
Undin, line 36:
```suggestion
        printer.print("Left:  ", ConsoleViewContentType.SYSTEM_OUTPUT)
```

PR #5649, opened 2020-06-24
CI: release nightly toml from github actions

Undin:
The current changes:
* publish nightly builds of toml plugin together with rust plugin
* check if there are changes in repo that can affect particular plugin not to publish the same builds
* disable `buildSearchableOptions` tasks for all non-root gradle projects to speed up publishing. It doesn't affect final builds because result of `buildSearchableOptions` task of non-root gradle project is not included in final build


Undin:
After this PR I'm going to do the same thing for beta/stable releases

lancelote:
bors merge

PR #5651, opened 2020-06-25
INSP: Supports `E0368` and `E0369` and provide fix

### LINKED ISSUE #5569: No E0368/E0369 for binary operations
## Environment

## Problem description
If the compiler can't find a proper implementation of trait responsible for the particular binary operator, it produces an error. Currently, the plugin says nothing in such cases.

## Steps to reproduce
```rust
struct Foo(u8);

fn add(lhs: Foo, rhs: Foo) {
    lhs + rhs; // error[E0369]: cannot add `Foo` to `Foo`
}

fn munis_assign(mut lhs: Foo, rhs: Foo) {
    lhs -= rhs; // error[E0368]: binary assignment operation `-=` cannot be applied to type `Foo`
}
```

Stzx:
Currently only `TyAdt` is supported.

Closed #5569

changelog: Detect [E0368](https://doc.rust-lang.org/error-index.html#E0368) and [E0369](https://doc.rust-lang.org/error-index.html#E0369) compiler errors


src/test/kotlin/org/rust/ide/inspections/RsBinaryOpOrBinaryOpAssignInspectionTest.kt
```diff
@@ -0,0 +1,243 @@
...
+        fn foo(a: Test, b: Test) {
+            <error descr="binary assignment operation `^=` cannot be applied to type `Test` [E0368]">a ^= b</error>;
+        }
+    """)
+
+    fun `test shl op assign`() = checkErrors("""
+        struct Test(i32);
+
+        fn foo(a: Test, b: Test) {
+            <error descr="binary assignment operation `<<=` cannot be applied to type `Test` [E0368]">a <<= b</error>;
```
Stzx, line 41: Bug:
* desc truncated to 
```
binary assignment operation `E0368]
```

src/test/kotlin/org/rust/ide/inspections/RsBinaryOpOrBinaryOpAssignInspectionTest.kt
```diff
@@ -0,0 +1,243 @@
...
+        fn foo(a: Test, b: Test) {
+            <error descr="binary operation `^` cannot be applied to type `Test` [E0369]">a ^ b</error>;
+        }
+    """)
+
+    fun `test shl op`() = checkErrors("""
+        struct Test(i32);
+
+        fn foo(a: Test, b: Test) {
+            <error descr="binary operation `<<` cannot be applied to type `Test` [E0369]">a << b</error>;
```
Stzx, line 113: Same as above.

src/test/kotlin/org/rust/ide/inspections/RsBinaryOpOrBinaryOpAssignInspectionTest.kt
Stzx: @Undin Can you look at this bug?

Stzx: It seems that I found the problem.

https://github.com/JetBrains/intellij-community/blob/3ab71c1c582869742dd2da068c1834963798a7b1/platform/lang-impl/src/com/intellij/codeInsight/daemon/impl/LocalInspectionsPass.java#L546

The problem appears on the regular expression of this line.

Stzx: Found it:
#2082
and
https://youtrack.jetbrains.com/issue/IDEA-186991

Stzx:
Open problem:
* <del>Unit tests have bug (descriptions are truncated abnormally</del>
    * #2082
* <del>How about we call `generateTraitMembers` automatically?</del>
* <del>I did not implement part of the error description, but replaced it with a unified description (is it okay?</del>
    * cannot add `Test` to `Test`
    * ^ Similar to this description
* <del>For repairing the generated content location, maybe it can be improved.</del>

vlad20012:
Please, rebase it to the latest master. Real project test now then should be run on CI.

dima74:
Rebased on master and supported some more cases.
@ortem could you please review?

afetisov:
Looks like the fix is provided only when the types of both operands are the same. The operator traits don't have that restriction. I believe it shouldn't be hard to support autofix for different types.

The restriction to `TyAdt` also seems excessive, it's not unusual to e.g. multiply objects and primitive numbers. Custom arithmetics for arrays can also be defined, provided that one of the operands is not an array.

If that restriction is lifted, then `lhsType.item.containingCrate == holder.currentAnnotationSession.currentCrate()` is also excessive, since we need only one of the types to be local. It would be better to ask the trait system whether the impl of the required form is allowed (since one of the types can be `#[fundamental]`).

It also seems to me that the quickfix impl will be malformed if the types contain any kind of generics (lifetime, type, const). I would expect that the quickfix would correctly infer unbound type parameters in the types of operands, and add them as impl parameters.

Finally, it would be great if the error message was more descriptive than just `"Cannot add `$rhsType` to `$lhsType`"`. It is very easy to miss some subtle difference in implementing and actual types, like a `&/&mut` on a type, some unsupported parameter value, or a bound on lifetime or type parameters. It would be great if the IDE provided some insight into the reason the impl wasn't found. This will also help to debug errors if the IDE can't find some impl due to e.g. a botched macro expansion or a missing generated file.

dima74:
> Looks like the fix is provided only when the types of both operands are the same. The operator traits don't have that restriction. I believe it shouldn't be hard to support autofix for different types.

Fixed

> The restriction to `TyAdt` also seems excessive, it's not unusual to e.g. multiply objects and primitive numbers. Custom arithmetics for arrays can also be defined, provided that one of the operands is not an array.

Fixed

> If that restriction is lifted, then `lhsType.item.containingCrate == holder.currentAnnotationSession.currentCrate()` is also excessive, since we need only one of the types to be local. It would be better to ask the trait system whether the impl of the required form is allowed (since one of the types can be `#[fundamental]`).

Fixed

> It also seems to me that the quickfix impl will be malformed if the types contain any kind of generics (lifetime, type, const). I would expect that the quickfix would correctly infer unbound type parameters in the types of operands, and add them as impl parameters.

Disabled quick-fix when there are unbound generics

> Finally, it would be great if the error message was more descriptive than just `"Cannot add `$rhsType`to`$lhsType`"`. It is very easy to miss some subtle difference in implementing and actual types, like a `&/&mut` on a type, some unsupported parameter value, or a bound on lifetime or type parameters. It would be great if the IDE provided some insight into the reason the impl wasn't found. This will also help to debug errors if the IDE can't find some impl due to e.g. a botched macro expansion or a missing generated file.

Good idea, we may add this in future PRs

afetisov:
>  || ty is TyTraitObject && (ty.traits.isEmpty() || isSameCrate(ty.traits.first().element))

This part is bugged. There is no guarantee on the order of traits within the definition of trait objects (i.e. `dyn Foo + Send` and `dyn Send + Foo` are both valid and equivalent). Github doesn't allow to put review on this line :-/

There can be only one non-auto trait in the list of traits, and all auto traits currently live in libcore. Thus outside of libcore we can just select the unique non-auto trait and do the check for it. libcore itself can't have dependencies, so I think that's the always correct algorithm.

dima74:
@afetisov fixed in #9416

Issue #3630, opened 2019-04-01
Local variable rename should detect name conflicts


vlad20012:
If I rename some variable with a name that is already in the scope, I should be notified about it with dialog with "continue renaming" and "cancel" buttons
 
![Peek 2019-04-01 17-29](https://user-images.githubusercontent.com/3221931/55335428-ba174380-54a3-11e9-89bc-342dd24de082.gif)

1tgr:
I started seeing this popup after a recent plugin update (I'm now on version 0.3.134.3455-202).

Is there a way to turn it off, and accept the rename anyway? I'm comfortable to introduce shadowed identifiers into my Rust code, I would not consider if a conflict to have two variables in the same scope with the same name.

cc @Kobzol

Kobzol:
@vlad20012 should we introduce an option to turn this off?

1tgr:
Or just a "Don't ask me again" box

PR #5657, opened 2020-06-25
REF: handle name conflicts after renaming bindings

Kobzol:
This PR adds support for detecting conflicts that can arise when bindings are renamed. The behaviour is inspired by the Kotlin/Java plugin, however I also added support for conflicts during renaming parameters (this doesn't happen in Kotlin/Java). I'm not sure if it's a good idea though.

The implementation uses `processLocalVariables` to find local bindings, so it only "sees" existing names that are available at the place of the binding which is being renamed.

```rust
fn test() {
    let c/*rename to a*/ = 1;  // no conflict
    let a = 1;
    let b/*rename to a*/ = 1;  // conflict
}
```
I'm not sure what's the best behaviour here. Renaming `c` to `a` should be fine, but only if there is no usage of `c` that follows the declaration of the previous `a` binding. To support this scenario we would have to scan the usages of local variables and only include the conflict if `X` is renamed to `Y` and a usage of `X` "crosses" a declaration of another binding with name `Y`. Or we can do it like Kotlin and simply process all local bindings in the scope (i.e. go both up and down), which might be simpler. It would also be easier for parameter handling, currently I scan the function from the end for local variables, but it does not take bindings that are the last child of a block/function into account because they do not "see" themselves when `processLocalVariables` is used. 

The line `error("Conflicts $conflicts missing")` in tests is not practically reachable, because if there are no conflicts the refactoring will proceed and the test will fail with a `ComparisonFailure`, which I did not know how to handle/catch.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3630

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsRenameProcessor.kt
@@ -10,17 +10,23 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Pass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.search.SearchScope
-import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.psi.util.parentOfType
 import com.intellij.refactoring.listeners.RefactoringElementListener
 import com.intellij.refactoring.rename.RenameDialog
 import com.intellij.refactoring.rename.RenamePsiElementProcessor
 import com.intellij.usageView.UsageInfo
+import com.intellij.util.containers.MultiMap
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.processLocalVariables
 
 class RsRenameProcessor : RenamePsiElementProcessor() {
-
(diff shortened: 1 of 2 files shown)


vlad20012:
:+1: 
bors r+

> Or we can do it like Kotlin and simply process all local bindings in the scope (i.e. go both up and down), which might be simpler. 

Well, why not? :shrug: Sounds better than current behavior. It can be a bit tricky to distinguish scopes where the renaming binding is reachable.
```rust
fn foo() {
    {
        let a = 1; // not reachable
        let b/*caret*/ = 1; // rename to a
        {
            let a = 1; // conflict
        }
    }
    let a = 1; // not reachable
}
```

Looks like we can collect binding in `b`'s parent scope and then filter binding below `b` using text offsets =D

vlad20012:
bors retry

Kobzol:
I'm a bit confused, so should I change the current behaviour? :D

vlad20012:
> I'm a bit confused, so should I change the current behaviour? :D

1. This merged PR is much better than old behavior (thanks!).
2. Looks like it's possible to improve it a bit more.

So, you _may_, if you wish. As usual 😃

Issue #4448, opened 2019-09-25
Improve struct fields' completion


SomeoneToIgnore:
## Environment

* **IntelliJ Rust plugin version:** 0.2.106.2134-193
* **Rust toolchain version:** 1.37.0 (eae3437df 2019-08-13) x86_64-apple-darwin
* **IDE name and version:** CLion 2019.3 EAP (CL-193.3519.24)
* **Operating system:** macOS 10.14.6

## Problem description

- [x] 1. Odd options for field completion:

```rust
pub struct MouseButtonDown {
    timestamp: u32,
    window_id: u32,
    which: u32,
    mouse_btn: MouseButton,
    clicks: u8,
    x: i32,
    y: i32
}

fn main() {
    let test = MouseButtonDown {/*caret*/}
}
```

Expected: only field names are available for completion, maybe a bunch of local variables.
Actual: `let` and `return` are present:
<img width="436" alt="image" src="https://user-images.githubusercontent.com/2690773/65610436-027a6e80-dfba-11e9-9ed9-01cb8d401627.png">

- [ ] 2. Field completion can be improved.

Same example as in 1., try to complete any of the fields.

Expected: a semicolon is added if no matching local variables are in scope:

```rust
fn main() {
    let test = MouseButtonDown {timestamp: /*caret*/}
}
```

Actual: I have to add it myself

```rust
fn main() {
    let test = MouseButtonDown {timestamp/*caret*/}
}
```

- [x] 3. No proper completion when destructuring the variable:

```rust
pub struct MouseButtonDown {
    timestamp: u32,
    window_id: u32,
    which: u32,
    mouse_btn: MouseButton,
    clicks: u8,
    x: i32,
    y: i32,
}

fn get_event() -> MouseButtonDown {
    unimplemented!()
}

fn main() {
    match get_event() {
        MouseButtonDown { /*caret*/ }
    }
}
```

Expected: only struct field names and `..` are available
Actual: ![image](https://user-images.githubusercontent.com/2690773/65610768-79176c00-dfba-11e9-8bef-4a1984385297.png)

Undin:
Reopened because we reverted #5659

Kobzol:
@SomeoneToIgnore I implemented 2) in https://github.com/intellij-rust/intellij-rust/pull/5659, but it maybe it's not such a good idea (https://github.com/intellij-rust/intellij-rust/issues/5875#issuecomment-669428733). Do you still want that functionality or can this issue be closed?

[...]

PR #5658, opened 2020-06-25
COMP: do not offer let and return inside struct literals and pats

Kobzol:
This PR changes keyword completion so that `return` and `let` are not offered directly inside struct literals and struct pats.

Fixes first part of https://github.com/intellij-rust/intellij-rust/issues/4448

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
@@ -854,6 +854,38 @@ class RsKeywordCompletionContributorTest : RsCompletionTestBase() {
         }
     """)
 
+    fun `test no return in struct literal`() = checkNotContainsCompletion("return", """
+        struct S { a: u32, b: u32 }
+        fn foo() {
+            let s = S { /*caret*/ };
+        }
+    """)
+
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
```diff
@@ -854,6 +854,22 @@ class RsKeywordCompletionContributorTest : RsCompletionTestBase() {
         }
     """)
 
+    fun `test no return in struct literal`() = checkNotContainsCompletion("return", """
```
Undin, line 857: Let's check both `let` and `return` keyword here

src/test/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributorTest.kt
```diff
@@ -854,6 +854,22 @@ class RsKeywordCompletionContributorTest : RsCompletionTestBase() {
...
+        struct S { a: u32, b: u32 }
+        fn foo() {
+            let s = S { /*caret*/ };
+        }
+    """)
+
+    fun `test no return in struct pat`() = checkNotContainsCompletion("return", """
```
Undin, line 864: And in this test as well

PR #5659, opened 2020-06-25
COMP: add colon after struct literal field if no matching binding is found

Kobzol:
This PR adds `: ` after completed struct field literal if no matching (same name and same type) local binding is found.

complete

There were 2 failing tests (`RsPartialParseCompletionTest#test struct field 1` and `RsPartialParseCompletionTest#test struct field 2`). If I used `foobar:` or `foobar: ` as the target completion, they still failed. I changed them to use `doSingleCompletion`.

Fixes second part of https://github.com/intellij-rust/intellij-rust/issues/4448


lancelote:
tldr Was reverted due to #5875

PR #5660, opened 2020-06-25
COMP: offer only struct fields in struct pat

Kobzol:
This PR filters out unnecessary completions offered in struct pats.
It also adds `..` to completion, but I'm not sure if that's the right way to do it, so it's in a separate commit.

Fixes third part of https://github.com/intellij-rust/intellij-rust/issues/4448


Kobzol:
OK, I'll create a new completion provider. I suggest to leave that for a separate PR and only merge 2024db21b81009e1292e54e689b53b12660ecc1e in this PR, is that OK?

mchernyavsky:
> OK, I'll create a new completion provider. I suggest to leave that for a separate PR and only merge 2024db2 in this PR, is that OK?

👍

Issue #3436, opened 2019-02-23
Bogus error 'Method is provate'


svenschmidt75:
## Environment

* **Intellij-Rust plugin version:** 0.2.92.2116-191
* **Rust toolchain version:** rustc 1.32.0 (9fda7c223 2019-01-16)
* **IDE name and version:** CLion 2019.EAP1
* **Operating system:** Ubuntu 18.04

## Problem description

Get https://github.com/svenschmidt75/Rust/tree/master/NeuralNetwork, commit 8194bd9307840b6186f60e69845d3dae73ab084a.
Open file fc_layer.rs and see red squigglies for methods being private. Those methods are
trait methods implemented in the derived type.
The code compiles just fine.
![screenshot from 2019-02-23 07-34-23](https://user-images.githubusercontent.com/17226522/53287725-8414d000-373d-11e9-9f33-e7adddc17274.png)

Kobzol:
This is a smaller reproduction:
```rust
use foo:: {Foo, Trait};

mod foo {
    pub struct Foo;
    impl Foo {
        fn get(&self) { println!("struct"); }
    }
    
    pub trait Trait {
        fn get(&self);
    }
    impl Trait for Foo {
        fn get(&self) { println!("trait"); }
    }
}

fn main() {
    let f = foo::Foo;
    f.get();
}
```
`f.get()` gets resolved to `Foo::get` instead of `Trait::get`. Even though inherent `impl`s have large priority than trait `impl`s, this is not true if the inherent method is not accessible (here because of visibility). If the inherent method was public, it would be used, but because it is private, the method from the trait is used.

This assumption is therefore not always valid: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt#L1218. @vlad20012 can we use visibility information in name resolution here?

PR #5661, opened 2020-06-25
COMP: prefer trait methods before private inherent methods

Kobzol:
Name resolution filters items by their visibility, however in the case of the mentioned issue, it is too late. Trait methods are filtered too deeply if an inherent method exists with the same name.

This PR changes it so that inherent methods are not preferred if they are not visible, thus fixing the previously given false positive.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3436


vlad20012:
Oh, I'm sorry, I completely missed this PR. Let's apply this patch, it should improve the performance a bit:

[patch.zip](https://github.com/intellij-rust/intellij-rust/files/5011983/unnamed.zip)

Kobzol:
Thanks, I have applied the patch.

vlad20012:
:+1: 
bors r+

PR #5665, opened 2020-06-26
T: Base class for move items refactoring

dima74:


src/test/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsTestBase.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        RsMoveTopLevelItemsProcessor(
+            project,
+            itemsToMove,
+            targetMod,
+            searchForReferences = true
+        ).run()
+    }
+
+    private fun <T> searchElementInAllFiles(root: PsiDirectory, searcher: (PsiFile) -> T?): T? {
```
ortem, line 49: Probably it's better to rewrite this method to something like
```Kotlin
fun <T> searchElementInAllFiles(root: VirtualFile, searcher: (PsiFile) -> T?): T?
```
using `VfsUtilCore.visitChildrenRecursively` (see usages in our other tests to find examples)

What do you think?

dima74: Thank you, replaced with `visitChildrenRecursively`

PR #5667, opened 2020-06-27
CI: release beta and stable toml builds from GitHub workflow

Undin:
Did the same thing as I implemented in #5649 but for beta and stable builds

Also, fix the case when the latest meaningful commit modifies github workflow. Previously, it leads to fail of `save-commit` job because token generated for github workflow run by default doesn't have permissions to modify workflow files (including creation of tag for existent commit with workflow modification).
To avoid errors in such cases, our script should use custom github token with the corresponding (workflow) permissions

### DIFF
--- a/.github/workflows/release-branch.yml
+++ b/.github/workflows/release-branch.yml
@@ -21,4 +21,4 @@ jobs:
                   git config --local user.name "GitHub Action"
 
             - name: Make release branch
-              run: python scripts/make_release_branch.py --token ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


Undin:
@lancelote let's merge it only after release 125 - looks like it may lead to some unexpected errors in release tasks on CI if master and release branch contains different code that can affect workflow tasks
Also, note, that you need to provide `WORKFLOW_GITHUB_TOKEN` secret with github access token as value that should have `workflow` scope for testing

lancelote:
bors merge

PR #5671, opened 2020-06-29
Fix cyclic stdlib dependencies

vlad20012:
Previously we had cyclic dependencies between stdlib crates like `std -> core -> std -> core`.

This was a bug in our internal project model. Looks like it was not affected our users anyhow.

I fix it because stdlib dependencies stop us from doing topological sort of crates that in turn is needed for the new name resolution engine.

Also, I removed ancient stdlib parts.


vlad20012:
reverted from release-126 branch because it breaks doctests in `liballoc`. In release-127 the bug will be fixed by #5690

PR #5672, opened 2020-06-29
Add `custom-build` cargo target kind

vlad20012:
`custom-build` is a target kind for `build.rs`


src/main/kotlin/org/rust/cargo/toolchain/CargoCommandLine.kt
```diff
@@ -66,6 +66,7 @@ data class CargoCommandLine(
                         listOf("--example", target.name)
                     CargoWorkspace.TargetKind.Bench -> listOf("--bench", target.name)
                     is CargoWorkspace.TargetKind.Lib -> listOf("--lib")
+                    CargoWorkspace.TargetKind.CustomBuild -> emptyList()
```
Undin, line 69:
```suggestion
                    CargoWorkspace.TargetKind.CustomBuild,
```

PR #5673, opened 2020-06-29
Parse `DepKind` from `cargo metadata`

vlad20012:
`DepKind` is needed to distinguish `[dependencies]`, `[dev-dependencies]` (for tests) and `[build-dependencies]` (for `build.rs`)


vlad20012:
reverted from release-126 branch

PR #5688, opened 2020-07-02
REF: Keep existing reference style in move refactoring

dima74:
Currently when updating references to moved file (in move single file refactoring), we firstly update use directives, and then update remaining non-resolved references as following:
* If there are many references, then we add import
* Otherwise replace each reference with absolute one

There are some problems with this approach:
* Threshold for "many references" was chosen kind of randomly (currently it equals 3)
* User may want to always add import, even if there are few usages
* User may want to use more qualified path even if there are many usages (e.g. using `io::Result` instead of `Result`)

So I think it would be better to keep existing reference style when changing references:
* If path is absolute, then replace it with absolute path
* If path is "one segment", then keep it as is and add import for it:
    ```rust
    // consider moving mod `foo` from `mod1` to `mod2`
    // before:
    use mod1::foo;
    foo::func();
    // after:
    use mod2::foo;
    foo::func();
    ```
* If path is "two segment", then change it and add import for parent:
    ```rust
    // consider moving mod `foo` from `mod1` to `mod2`
    // before:
    use crate::mod1;
    mod1::foo::func();
    // after:
    use crate::mod2;
    mod2::foo::func();
    ```
* And so on


src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.rust.lang.core.psi.RsUseItem
+import org.rust.lang.core.psi.ext.*
+
+class RsMoveReferenceInfo(
+    val pathOld: RsPath,
+    val pathNew: RsPath?,
+    // == `pathOld.reference.resolve()`
+    val target: RsQualifiedNamedElement
+) {
+    val isInsideUseDirective: Boolean get() = pathOld.parentOfType<RsUseItem>() != null
```
ortem, line 22: Consider using `pathOld.ancestorStrict<RsUseItem>()` instead (here and below)

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveRetargetReferencesProcessor.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    // "keep existing style" means that
+    // - if `pathOld` is `foo` then we keep it and add import for `foo`
+    // - if `pathOld` is `mod1::foo`, then we change it to `mod2::foo` and add import for `mod2`
+    // - etc for `outer1::mod1::foo`
+    private fun tryRetargetReferenceKeepExistingStyle(reference: RsMoveReferenceInfo): Boolean {
+        val pathOld = reference.pathOld
+        val pathNew = reference.pathNew ?: return false
+
+        val pathOldSegments = pathOld.text.split("::")
```
ortem, line 52: @vlad20012 Could you please tell whether it's OK to use `RsPath.text` here and below (e.g. in `replacePathOld` method)? Or maybe `RsPath.referenceName` should be used instead?

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -0,0 +1,62 @@
...
+    if (text.startsWith("::")) return true
+    if (startsWithSuper()) return false
+
+    val basePathTarget = basePath().reference?.resolve() as? RsMod ?: return false
+    return basePathTarget.isCrateRoot
+}
+
+fun RsPath.startsWithSuper(): Boolean = basePath().text == "super"
+
+fun RsPath?.resolvesToAndAccessible(target: RsQualifiedNamedElement): Boolean {
```
ortem, line 35: Is nullability needed here?

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveRetargetReferencesProcessor.kt
vlad20012: It's almost always NOT ok because of possible whitespace/comments between pats segments: `foo::/*foobar*/   bar :: baz`.

Fun fact is that we use it anyway in AutoImportFix:
```
private val RsUseItem.pathAsList: List<String>?
    get() = useSpeck?.path?.text?.split("::")
```

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
dima74: Not needed, removed

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveRetargetReferencesProcessor.kt
dima74: Ok, thanks, so I added new method `textNormalized`, which excludes comments and whitespaces

dima74:
@ortem thanks for the comments! I added test for code related to `containingModHasSameNameInScope` (and also updated code to cover more cases), and removed code related to deletion of useless (one-segment) import, because it is not used now (will be used by move items refactoring)

src/main/kotlin/org/rust/ide/refactoring/move/RsMoveFilesOrDirectoriesProcessor.kt
```diff
@@ -182,49 +184,14 @@ class RsMoveFilesOrDirectoriesProcessor(
         newParentMod.insertModDecl(psiFactory, newModDeclaration)
     }
 
-    private fun retargetOtherUsages(otherUsages: MutableList<UsageInfo>) {
-        val movedFileName = movedFile.modName
-        if (movedFileName == null) {
-            super.retargetUsages(otherUsages.toTypedArray(), emptyMap())
-            return
+    fun retargetUsages(usages: List<UsageInfo>) {
```
ortem, line 187: `private`?

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -0,0 +1,75 @@
...
+    val pathOld: RsPath,
+    val pathNew: RsPath?,
+    // == `pathOld.reference.resolve()`
+    val target: RsQualifiedNamedElement
+) {
+    val isInsideUseDirective: Boolean get() = pathOld.ancestorStrict<RsUseItem>() != null
+}
+
+fun RsPath.isAbsolute(): Boolean {
+    if (text.startsWith("::")) return true
```
vlad20012, line 28:
```suggestion
    if (hasColonColon) return true
```

dima74: Sorry, I don't understand this change. Consider path `foo::bar`, it has `hasColonColon == true`, but if `foo` is not crate root, then this path is not absolute

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -0,0 +1,75 @@
...
+
+fun RsPath.startsWithSuper(): Boolean = basePath().referenceName == "super"
+
+// like `text`, but without spaces and comments
+val RsPath.textNormalized: String
+    get() {
+        val copy = copy()
+        for (child in copy.childrenOfType<PsiElement>()) {
+            if (child is PsiWhiteSpace || child is PsiComment) {
+                child.delete()
```
ortem, line 43: Deleting PSI children while iterating looks dangerous.

Also, `childrenOfType` returns only children, but not all descendants. We usually use our methods from `org.rust.lang.core.psi.ext.PsiElement.kt` as they have more precise names. You could use `PsiElement.descendantsOfType()` or even `PsiTreeUtil.findChildrenOfAnyType(copy, PsiWhiteSpace::class.java, PsiComment::class.java)` here. But probably it would be even better to build a string from a sequence of nested `RsPath.path` instead of iterating+deleting PSI

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
vlad20012: For given path `foo::bar` `hasColonColon == false` :woman_shrugging: 
`hasColonColon` is almost equivalent to `text.startsWith("::")`

dima74: I thought that `hasColonColon` is equivalent to `text.contains("::")`, isn't it?

Maybe I do something wrong, but on the screenshot `hasColonColon` is true for path `mod1::foo`:
image

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
dima74: Thank you, changed to building a string from parts

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
vlad20012: Oh, I'm you're right, I'm sorry. The, `basePath().hasColonColon` should work

dima74: Ok, so replaced with `basePath().hasColonColon`, thank you

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -0,0 +1,67 @@
...
+}
+
+fun RsPath.startsWithSuper(): Boolean = basePath().referenceName == "super"
+
+// like `text`, but without spaces and comments
+val RsPath.textNormalized: String
+    get() {
+        val parts = listOfNotNull(path?.textNormalized, coloncolon?.text, referenceName, typeArgumentList?.text)
+        return parts.joinToString("")
+    }
```
vlad20012, line 39: Hmm, previously I assumed that this method is used for `use`-item paths only. Is it the right assumption? If yes then `typeArgumentList` is impossible there. If not, then splitting by "::" is wrong becase a path can contain sub-paths in type arguments: `foo::bar::<baz::quux>`, as well as type qualifier `<foo::Bar>::baz`

dima74: Currently this method is used for paths which resolve to moved file (these paths may be not in `use`-item). As far as I know it is not possible to have path resolved to `RsMod` which contains type qualifiers or type arguments. In future (in move items refactoring) we will have paths resolved to `RsItemElement`, and these paths may contain type arguments, but they will be handled separately. So lets remove `typeArgumentList` from `textNormalized` for now

dima74:
bors r=ortem

Issue #5594, opened 2020-06-17
Support for non-Cargo based projects


maxim-banaev:
## Feature description
The Rust plugin currently [only supports](https://intellij-rust.github.io/docs/troubleshooting.html) projects using Cargo.toml files. For large codebase using a centralized build system, such as Android or Fuchsia, this approach is not practical.

The [rust-analyzer](https://rust-analyzer.github.io/) project provides an alternative to declare the dependencies (https://rust-analyzer.github.io/manual.html#non-cargo-based-projects). This de facto standard is already generated by [Fuchsia](https://gn.googlesource.com/gn/+/refs/heads/master/src/gn/rust_project_writer.h) and [Android](https://android.googlesource.com/platform/build/soong/+/refs/heads/master/rust/project_json.go).

Would it be possible to add an option to parse this format and automatically generate the dependencies within CLion?

tweksteen:
cc @matklad @benbrittain @woody77

benbrittain:
also cc/ @dtolnay who might be using rust-project.json now as well at FB

TomaszWaszczyk:
Cargo in Idea++;

n1c00o:
Any update or improvements on this?

Being able to use Rust without Cargo is essential in some code bases that use central build systems.

n1c00o:
A great starting point could be to support Rust-Analyzer's JSON schema?

nikhil-marathe-skydio:
@maxim-banaev any idea of the priority for this?

mgeisler:
I can confirm that [Android (AOSP) is also supporting the `rust-project.json` format](https://source.android.com/docs/setup/build/rust/building-rust-modules/rust-ide-setup).

PR #5690, opened 2020-07-02
Introduce `Crate`, `CrateGraphService`

vlad20012:
*Crate graph* is a simplified project model (comparing to complex [cargo project model](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md#project-model)) described in terms of *crates* (mostly from `rustc` point of view).

Crate is usually a `Target` In terms of Cargo project model

This PR solves these problems:
1. It's just simpler to work with Crate graph then with Cargo project model. Some complex code was simplified
2. Sometimes we want to have a (virtual, dummy, fake) crate where there are no cargo targets, e.g. in the case of `doctest`s
3. Now we have API that allows us to get direct/transitive/reverse dependencies of some crate
4. Now we have topological sorting of crates and persistent crate `id`s, both needed for the new name resolution 2.0
5. In theory, this opens a way to support non cargo-based projects (like #5594)

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -17,6 +17,7 @@ import org.rust.cargo.util.AutoInjectedCrates.STD
 import org.rust.cargo.util.StdLibType
 import org.rust.openapiext.CachedVirtualFile
 import org.rust.stdext.applyWithSymlink
+import org.rust.stdext.mapToSet
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.*
@@ -68,7 +69,8 @@ interface CargoWorkspace {
         val origin: PackageOrigin
 
         val targets: Collection<Target>
-        val libTarget: Target? get() = targets.find { it.isLib }
+        val libTarget: Target? get() = targets.find { it.kind.isLib }
+        val customBuildTarget: Target? get() = targets.find { it.kind == TargetKind.CustomBuild }
 
         val dependencies: Collection<Dependency>
 
@@ -91,20 +93,12 @@ interface CargoWorkspace {
     /** See docs for [CargoProjectsService] */
     interface Target {
         val name: String
+
         // target name must be a valid Rust identifier, so normalize it by mapping `-` to `_`
         // https://github.com/rust-lang/cargo/blob/ece4e963a3054cdd078a46449ef0270b88f74d45/src/cargo/core/manifest.rs#L299
         val normName: String get() = name.replace('-', '_')
 
         val kind: TargetKind
-
-        val isLib: Boolean get() = kind is TargetKind.Lib
-        val isBin: Boolean get() = kind == TargetKind.Bin
-        val isExampleBin: Boolean get() = kind == TargetKind.ExampleBin
-        val isProcMacro: Boolean
-            get() {
-                val kind = kind
-                return kind is TargetKind.Lib && kind.kinds.contains(LibKind.PROC_MACRO)
-            }
 
         val crateRoot: VirtualFile?
 
@@ -127,8 +121,10 @@ interface CargoWorkspace {
     )
 
     enum class DepKind {
-        // Stdlib
-        All,
+        // For old Cargo versions prior to `1.41.0`
+        Unclassified,
+
+        Stdlib,
         // [dependencies]
         Normal,
         // [dev-dependencies]
@@ -149,6 +145,12 @@ interface CargoWorkspace {
         object Bench : TargetKind("bench")
         object CustomBuild : TargetKind("custom-build")
         object Unknown : TargetKind("unknown")
+
+        val isLib: Boolean get() = this is Lib
+        val isBin: Boolean get() = this == Bin
+        val isExampleBin: Boolean get() = this == ExampleBin
+        val isProcMacro: Boolean
+            get() = this is Lib && this.kinds.contains(LibKind.PROC_MACRO)
     }
 
     enum class LibKind {
@@ -159,7 +161,7 @@ interface CargoWorkspace {
         EDITION_2015("2015"), EDITION_2018("2018")
     }
 
-    class Feature(
+    data class Feature(
         val name: String,
         val state: FeatureState
     )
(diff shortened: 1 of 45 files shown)


src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -211,14 +209,30 @@ private class WorkspaceImpl(
         // so we have to rebuild the whole workspace from scratch instead of
         // *just* adding in the stdlib.
 
-        val stdAll = stdlib.crates.associateBy { it.id }
-        val stdGated = stdlib.crates.filter { it.type == StdLibType.FEATURE_GATED }.map { it.id }.toSet()
-        val stdRoots = stdlib.crates.filter { it.type == StdLibType.ROOT }.map { it.id }.toSet()
+        val (newPackagesData, stdCrates) = if (!stdlib.isPartOfCargoProject) {
```
vlad20012, line 212: These changes are about better `rustc` project support: now we don't duplicate stdlib packages (because they are already a part of the project)

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -234,9 +248,8 @@ private class WorkspaceImpl(
                         val dependencyPackage = newIdToPackage[pkg.id] ?: return@mapNotNull null
                         DependencyImpl(dependencyPackage, name, depKinds)
                     })
-                    pkg.dependencies.addAll(stdlibDependencies.filter { it.pkg.id in stdRoots })
-                    val explicitDeps = pkg.dependencies.map { it.pkg.id }.toSet()
-                    pkg.dependencies.addAll(stdlibDependencies.filter { it.pkg.id in stdGated && it.pkg.id !in explicitDeps })
+                    val explicitDeps = pkg.dependencies.map { it.name }.toSet()
+                    pkg.dependencies.addAll(stdlibDependencies.filter { it.name !in explicitDeps })
```
vlad20012, line 252: Here I allowed overriding any stdlib dependency. E.g. in `rustc` it's common to write
```
[dependencies]
core = { path = "../libcore" }
```
So, we should not add one more "core" in this case

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -159,7 +162,11 @@ class RsFile(
         get() = getOwnedDirectory() != null
 
     override val isCrateRoot: Boolean
-        get() = originalFile == crateRoot
+        get() {
+            val file = originalFile.virtualFile ?: return false
+            return file is VirtualFileWithId && project.crateGraph.findCrateByRootModule(file) != null
+                || file.isDoctestInjection(project)
+        }
```
vlad20012, line 169: This change is only about the performance

src/main/kotlin/org/rust/lang/core/psi/RsFile.kt
```diff
@@ -191,8 +198,9 @@ class RsFile(
             MOD_DECL_KEY to ModificationTracker.NEVER_CHANGED
         }
         return CachedValuesManager.getCachedValue(originalFile, key) {
+            val decl = if (isCrateRoot) emptyList() else RsModulesIndex.getDeclarationsFor(originalFile)
```
vlad20012, line 201: This change is only about the performance

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -264,32 +264,25 @@ fun processExternCrateResolveVariants(
     withSelf: Boolean,
     processor: RsResolveProcessor
 ): Boolean {
-    val isDoctestInjection = element.isDoctestInjection
-    val target = element.containingCargoTarget ?: return false
-    val pkg = target.pkg
+    val crate = element.containingCrate ?: return false
```
vlad20012, line 267: Btw this code now looks much clearer with the new Crate API!

src/test/kotlin/org/rustSlowTests/lang/resolve/CargoProjectResolveTest.kt
```diff
@@ -545,4 +547,189 @@ class CargoProjectResolveTest : RsWithToolchainTestBase() {
...
+        }
+    }.run {
+        CrateGraphTestmarks.cyclicDevDependency.checkHit {
+            checkReferenceIsResolved<RsPath>("tests/main.rs")
+            checkReferenceIsResolved<RsPath>("src/lib.rs")
+        }
+    }
+
+    @MinRustcVersion("1.41.0") // In this version Cargo starts providing `[build-dependencies]` info
+    fun `test build dependency is resolved in 'build rs' and not resolved in 'main rs'`() = buildProject {
```
Undin, line 676: Let's add similar tests when a common dependency is not resolved from `build.rs`

src/main/kotlin/org/rust/lang/core/crate/Crate.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.intellij.openapi.vfs.VirtualFile
+import org.rust.cargo.CfgOptions
+import org.rust.cargo.project.model.CargoProject
+import org.rust.cargo.project.workspace.CargoWorkspace
+import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.lang.core.psi.RsFile
+import java.util.*
+
+/**
+ * Immutable objects describes a *crate* from the *rustc* point of view.
```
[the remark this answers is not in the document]
vlad20012: > Do we really need Immutable object part?

Yes, it is a very-very important property of the implementors. It means that if you have a reference to a `Crate` instance, none of its properties can be changed. Yes, we can't guarantee it programmically, so we guarantee it by the documentation :)

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspaceData.kt
```diff
@@ -47,6 +47,6 @@ data class CargoWorkspaceData(
     data class Dependency(
         val id: PackageId,
         val name: String? = null,
-        val depKinds: List<CargoWorkspace.DepKindInfo> = listOf(CargoWorkspace.DepKindInfo(CargoWorkspace.DepKind.Normal))
+        val depKinds: List<CargoWorkspace.DepKindInfo> = listOf(CargoWorkspace.DepKindInfo(CargoWorkspace.DepKind.All))
```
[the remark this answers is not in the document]
vlad20012: Yes, tests and old cargo

src/main/kotlin/org/rust/lang/core/crate/Crate.kt
```diff
@@ -0,0 +1,91 @@
...
+ * In Cargo-based project this is usually a wrapper around [CargoWorkspace.Target]
```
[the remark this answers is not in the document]
vlad20012: doctests :)

src/main/kotlin/org/rust/lang/core/crate/impl/CargoBasedCrate.kt
```diff
@@ -0,0 +1,71 @@
...
+    override val id: CratePersistentId? = rootModuleFile?.fileId
+
+    var cyclicDevDeps: List<Crate.Dependency> = emptyList()
+
+    override val dependenciesWithCyclic: Collection<Crate.Dependency>
+        get() = dependencies + cyclicDevDeps
+
+    init {
+        for (dependency in dependencies) {
+            (dependency.crate as CargoBasedCrate).reverseDependencies += this
```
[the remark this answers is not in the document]
vlad20012: Well, I haven't found the way.

src/main/kotlin/org/rust/lang/core/crate/impl/CrateGraphServiceImpl.kt
```diff
@@ -0,0 +1,400 @@
...
+
+        for (dependency in this) {
+            val visitedDepKinds = enumSetOf<CargoWorkspace.DepKind>()
+
+            for (depKind in dependency.depKinds) {
+                // `cargo metadata` sometimes duplicate `depKinds` for some reason
+                if (!visitedDepKinds.add(depKind.kind)) continue
+
+                when (depKind.kind) {
+                    CargoWorkspace.DepKind.All -> {
```
Undin, line 223: As I understand, `All` is only possible to stdlib dependencies? if yes, maybe it's worth to rename to `Stdlib`?

src/main/kotlin/org/rust/lang/core/crate/impl/CrateGraphServiceImpl.kt
```diff
@@ -0,0 +1,400 @@
...
+                        normal += dependency
+                        build += dependency
```
Undin, line 225: Let's add a comment, why we don't add it to `dev` dependencies as well

vlad20012:
bors r=undin

vlad20012:
bors r=undin

Undin:
bors retry

PR #5693, opened 2020-07-03
Notify user about unsupported toolchain version

vlad20012:
We have a minimal supported Rust toolchain version. Today it is `1.32.0` (but I want to bump it to `1.39.0` in #5690). An older toolchain can lead to random bugs, or the plugin can be totally unusable. For now, we don't notify a user about outdated toolchain. I propose such notification:

image


vlad20012:
image

vlad20012:
bors r=undin

PR #5699, opened 2020-07-05
Detect integer overflow and division by zero errors

### LINKED ISSUE #5568: Literal overflow inspection

Undin:
These changes:
* use `BigInteger` instead of `Long` to cover all valid values for large types like `u128`
* detect literal and arithmetic integer overflow in const expressions
* detect unconditional division by zero

Note, these changes do not highlight errors like `-1u8`

Part of #5888

Closes #5568
Closes #5585

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -39,13 +39,14 @@ import org.rust.lang.core.resolve.knownItems
 import org.rust.lang.core.resolve.namespaces
 import org.rust.lang.core.resolve.ref.deepResolve
 import org.rust.lang.core.types.*
-import org.rust.lang.core.types.consts.asLong
+import org.rust.lang.core.types.consts.asInteger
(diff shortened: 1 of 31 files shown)


Undin:
@Kobzol While review of your PR I understood that I'm not sure how it's better to handle errors while constant evaluation and eventually I made some own implementation 😄

Kobzol:
This looks great! For now I would simply assume that isize/usize is 64-bit and we can change it in a later PR to take the target pointer width into account.

Kobzol (approved):

vlad20012:
This will be covered by the proper MIR-based const evaluation

Issue #1087, opened 2017-03-25
Using rustfmt should be more embraced


fulara:
I think the RustFmt should be preferred way of formatting the code over building the formatter inside the plugin. This has a lot of advantages. Biggest one is probably that you would no longer have to maintain the formatter engine.  
The most drastic measure could be just to by default block the plugin's formatter and by default scream message to install the rustfmt. :). The block could be configurable in the options - novice users would probably prefer to use rustfmt than the Plugin's fmt, and seasoned users if they like the built in formatter could switch that in the options.  
Also, right now its unintuitive to switch to rustfmt over builtin plugin

Since more attention would be paid to integrating of rustfmt then some handling of its stdout would be welcome.  for example:
a) When rustfmt stdouts nothing - that means it formatter correctly a pop-up should be shown that formatting went ok
b) When rustfmt does not know how to split the lines, it outputs file and line number where the problem occured, this should be clickable

matklad:
I would say that we'll need to do both :) Let me explain this a little bit!

IntelliJ and rustfmt use different approaches to formatting. rustfmt will completely rewrite your code to fit the Rust style. IntelliJ will just try to enforce certain rules (like, "there should be a space after comma"), but it will leave the general structure of the code intact. That is, we don't try to maintain a line length limit and let the user control breaking complex expressions into several lines.

This means that IntelliJ formater can be just "less strict" then `rustfmt`: code formatter by `rustfmt` won't be changed by IntelliJ, but not necessary vice verse. And this should be relatively easy to achieve as soon as official style guide is finished, [our implementation](https://github.com/intellij-rust/intellij-rust/blob/b467a49b08ae011863c7e264be5f69d58a3b5df2/src/main/kotlin/org/rust/ide/formatter/impl/spacing.kt#L32-L138) is much shorter than the one in rustfmt. 

And we do need our own formatter because it is used for more than just formatting. For example, when you press `Enter` formatting model is used to determine if the caret should be indented or not. Formatter is also crucial for quick fixes, intentions and refactoring: it's next to impossible to generate good looking code manually, and formatter saves the day.

Improving `rustfmt` UI in the plugin is definitely something we should do though!

fulara:
I am just stating my opinion here - apologies If I am too aggressive with these. :)  

[...]

PR #5704, opened 2020-07-06
FMT: Add an option to use rustfmt instead of build-in formatter

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/5191. c.c. #1087

In this implementation, `Rustfmt` can be used **only** when formatting the **whole file** by: 
1. `Code | Reformat Code` (`Ctrl + Alt + L`) action **without text selection**
2. `Code | Reformat File ...` (`Ctrl + Alt + Shift + L`) action with `Whole file` option

E.g. `Rustfmt` can't be used when committing to git with `Reformat code Before Commit` option (it doesn't clear how to achieve this).

By default, the built-in formatter is still ALWAYS used. `Rustfmt` can be enabled in `Settings | Languages & Frameworks | Rust | Rustfmt`.

NB: rustfmt has an unstable option to restrict reformatting to specific sets of lines (https://github.com/rust-lang/rustfmt/issues/3397).

### DIFF
--- a/src/main/kotlin/org/rust/cargo/RustfmtWatcher.kt
+++ b/src/main/kotlin/org/rust/cargo/RustfmtWatcher.kt
@@ -109,7 +109,7 @@ class RustfmtWatcher {
         private fun Rustfmt.reformatDocument(cargoProject: CargoProject, document: Document) {
             checkIsDispatchThread()
             if (!document.isWritable) return
-            val formattedText = reformatDocumentText(cargoProject, document) ?: return
+            val formattedText = reformatDocumentTextOrNull(cargoProject, document) ?: return
             DocumentUtil.writeInRunUndoTransparentAction { document.setText(formattedText) }
         }
     }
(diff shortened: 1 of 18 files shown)


mchernyavsky:
@Undin Done.

Kobzol:
Just wondering, isn't the option "reformat with rustfmt on save" a bit weird after this change? Maybe it could be changed to "reformat on save" to be ortogonal with selecting the used formatter?

vlad20012:
> Just wondering, isn't the option "reformat with rustfmt on save" a bit weird after this change? Maybe it could be changed to "reformat on save" to be ortogonal with selecting the used formatter?

I think the option should be eventually removed in favor of something like [save action plugin](https://plugins.jetbrains.com/plugin/7642-save-actions)

vlad20012 (approved):
🎉🎉🎉
I updated the PR description

bors r+

vlad20012:
bors retry

PR #5717, opened 2020-07-10
Fix `go to type declaration` for macro-generated items

vlad20012:
Fix `go to type declaration` action (`Ctrl + Shift + B`) for macro-generated items.

Works for this case:
```rust
macro_rules! as_is { ($($ t:tt)*) => { $($ t)* }; }
as_is! { struct /*caret_after*/S; }
fn main() {
    let /*caret_before*/a = S;
}
```

I used PsiElement.getNavigationElement() because it specially designed to get some "real" PSI from "fake" (macro-generated) PSI


vlad20012:
bors retry

PR #5723, opened 2020-07-11
STUB: improve block stub creation

Undin:
* handle `raw const` operator  and not to create stubs in this case
* extract common code of the corresponding tests into `RsLazyBlockStubCreationTestBase`


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -119,7 +119,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
  *   block, so it is processed earlier). The value is then removed from the cache because it is no longer needed
  *   and should be invalidated after each PSI change.
  *
- * See tests in `RsLazyCodeBlockIsNotExpandedDuringStubBuildingTest`
+ * See tests in `RsCodeBlockStubCreationTest`
```
vlad20012, line 122: We could also mention `RsCompilerSourcesLazyBlockStubCreationTest` ...

PR #5725, opened 2020-07-11
CI: improve github workflows

Undin:
* declare env variables via workflow syntax if possible
* check oldest IDE versions


Undin:
bors retry

Issue #5731, opened 2020-07-12
Support `Try` trait while type inference of `?` operator


Sherlock-Holo:
## Environment

* **IntelliJ Rust plugin version:** 0.3.127.3214-201-nightly
* **Rust toolchain version:** rustup 1.22.1 (2020-07-08)
* **IDE name and version:** clion
* **Operating system:** Arch Linux

## Problem description
Poll<Result<T, E>> implement `Try` trait but when use `?`, the return value doesn't have type hint.

## Steps to reproduce

```rust
let x = poll_sth(cx)?;
```

x doesn't have type hint

Undin:
@Sherlock-Holo could you provide a more complete code example?

Sherlock-Holo:
```rust
fn ddd() -> Poll<Result<(), String>> {
    Poll::Pending
}

fn dd() -> Poll<Result<(), String>> {
    let x = ddd()?;

    Poll::Pending
}
```

the x doesn't have type hint

![image](https://user-images.githubusercontent.com/10096425/87250381-eedfbc00-c496-11ea-9597-32ecc3d26231.png)

Undin:
Thanks! I can reproduce it.
Looks like the reason is the absence of common support of `Try` trait. Only hardcoded versions for `Result` and `Option` enums

PR #5733, opened 2020-07-13
TY: infer Try impl for try expressions

Kobzol:
This PR adds support for inferring try expressions (`<expr>?`) where the type of `<expr>` is not `Result` nor `Option`, but it implements the `core::ops::try::Try` trait. The fast path for results and options is kept in the code.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5731

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
@@ -1039,8 +1039,11 @@ class RsTypeInferenceWalker(
 
     private fun inferTryExprType(expr: RsTryExpr): Ty {
         val base = expr.expr.inferType() as? TyAdt ?: return TyUnknown
-        // TODO: make it work with generic `std::ops::Try` trait
-        if (base.item != items.Result && base.item != items.Option) return TyUnknown
+        if (base.item != items.Result && base.item != items.Option) {
+            val tryItem = items.Try ?: return TyUnknown
+            val okType = tryItem.findAssociatedType("Ok") ?: return TyUnknown
+            return ctx.normalizeAssociatedTypesIn(TyProjection.valueOf(base, BoundElement(tryItem), okType)).value
+        }
         TypeInferenceMarks.questionOperator.hit()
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
```diff
@@ -1039,8 +1039,12 @@ class RsTypeInferenceWalker(
 
     private fun inferTryExprType(expr: RsTryExpr): Ty {
         val base = expr.expr.inferType() as? TyAdt ?: return TyUnknown
-        // TODO: make it work with generic `std::ops::Try` trait
-        if (base.item != items.Result && base.item != items.Option) return TyUnknown
+        if (base.item != items.Result && base.item != items.Option) {
+            val (implLookup, knownItems) = expr.implLookupAndKnownItems
```
vlad20012, line 1043: `implLookup` and `knownItems` should already be in the scope (TypeInferenceWalker propertives)

src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
```diff
@@ -1039,8 +1039,12 @@ class RsTypeInferenceWalker(
...
+            val tryItem = knownItems.Try ?: return TyUnknown
+            val result = implLookup.select(TraitRef(base, BoundElement(tryItem))).ok() ?: return TyUnknown
+            return result.impl.associatedTypesTransitively.find { it.name == "Ok" }?.declaredType?.substitute(result.subst) ?: TyUnknown
```
vlad20012, line 1046: Let's do `ctx.normalizeAssociatedTypesIn(TyProjection.valueOf(base, tryItem, tryItem.findAssociatedType("Ok")))` instead. This is better in the case of partially inferred types

Kobzol: Ok! Thanks, this also looks cleaner.

vlad20012:
:+1: 
bors r+

PR #5737, opened 2020-07-14
INSP: change RsVariableMutableInpsection to be flow-sensitive

Kobzol:
This PR changes the `Variable can be immutable` inspection to be based on control flow. I took `RsLivenessInspection` and refactored out the common bits (both of the inspection and of the analysis pass), because the mutability inspection is very similar to it.

Right now the case when a variable without an initialization is assigned only once is not checked (not by `RsLivenessInspection` nor by `RsVariableMutableInspection`) - this is required to remove some false negatives and it will also probably be a basis for a dead assignment analysis. Maybe we should keep this for a separate PR and only enable `RsVariableMutableInspection` by default after it works?

Before changing that I wanted to restructure the intention so that it is control-flow based and resolve some questions. Is this separation OK, i.e. should it be two independent analysis passes? Maybe we should separate and cache the delegate implementation, right now it is invoked twice (for liveness and mutability), which might be needlessly costly.


ortem:
Hi @Kobzol, I'm sorry again that I've been inactive regarding this PR for so long. First of all, thank you for such a great job! I will try to focus on control- and data-flow analysis for a while and give my feedback on this PR soon.

Kobzol:
Heh, I was just thinking about this PR! :laughing: No problem, I'll be glad for any feedback :)

It was a bit difficult for me to undertand the current control/data flow infrastructure, I think that there are some changes that I didn't do yet and that will need done for this PR to work.

ortem:
Hi @Kobzol,

> Right now the case when a variable without an initialization is assigned only once is not checked

Well, a half of this case is already covered by our "borrow checker" (basically, move analysis) in `org.rust.lang.core.types.borrowck.BorrowCheckContext#reportUseOfMovedValue`. I mean the check that such variable is assigned *at least once*.
For example, IJ Rust shows an error for the following code:
```Rust
fn main() {
    let x: i32;
    x; // error: Use of possibly uninitialized variable
}
```
But you're absolutely right that it would be great to check that such a variable is assigned *no more than once* as well. I'll think about that soon and suggest something.

As for refactoring, the idea looks good to me. Moreover, I suppose we should extract all the code related to data-flow analysis to a separate package to make it more clear and easy to maintain. So I've opened https://github.com/intellij-rust/intellij-rust/pull/6718. Feel free to suggest any ideas you have!

src/main/kotlin/org/rust/lang/core/types/analysis/mutability/UnusedMutability.kt
```diff
@@ -0,0 +1,186 @@
...
+}
+
+object MutabilityDataFlowOperator : DataFlowOperator {
+    override fun join(succ: Int, pred: Int): Int = succ or pred     // liveness from both preds are in scope
+    override val initialValue: Boolean get() = false                // does not need mutability by default
+}
+
+private typealias MutabilityDataFlow = DataFlowContext<MutabilityDataFlowOperator>
+
+private class FlowedMutabilityData(
```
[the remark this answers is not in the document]
ortem: But there is also another option which might be better. Note that `UnusedMutability` analysis is very similar to `Liveness` analysis. The idea is to merge this analysis right into the liveness analysis. Basically, when I was implementing liveness analysis, I was going to extend it with dead assignment analysis someday (see https://github.com/intellij-rust/intellij-rust/issues/3163) using the same class.


For example, let's compare the current implementation of `org.rust.lang.core.types.FlowedLivenessData.Companion#buildFor` with the `org.rust.lang.core.types.borrowck.FlowedMoveData.Companion#buildFor`.

The `FlowedMoveData` was initially designed to analyse the data-flow for two different things:
* moves of the variable, e.g. `let another_var = var` or `consume_value(var)`
* assignments to the variable, e.g. `var = another_var`

Though `FlowedLivenessData` currently cares only about _live paths_, this implementation may be extended by  _live assignments_ and probably _live mutabilities_.

src/main/kotlin/org/rust/lang/core/types/analysis/mutability/UnusedMutability.kt
```diff
@@ -0,0 +1,186 @@
...
+    val binOp = parent as? RsBinaryExpr
+    // element = ...
+    if (binOp?.isAssignBinaryExpr == true && binOp.left == element) {
+        return true
+    }
+    // &mut <element>
+    if ((parent as? RsUnaryExpr)?.operatorType == UnaryOperator.REF_MUT) {
+        return true
+    }
+    // lambda calls
```
ortem, line 113: [minor] Not only closures, but normal function calls too

src/main/kotlin/org/rust/lang/core/types/analysis/mutability/UnusedMutability.kt
```diff
@@ -0,0 +1,186 @@
...
+
+            mutabilityData.addGenKills(dfcxLivePaths)
+            dfcxLivePaths.propagate()
+
+            return FlowedMutabilityData(ctx, mutabilityData, dfcxLivePaths)
+        }
+    }
+}
+
+private fun isMutableContext(element: RsElement): Boolean {
```
ortem, line 102: Some of the logic looks duplicated because of `ExprUseWalker`. For example, the `RsCondition` case may be covered by `ExprUseWalker#walkCondition` which calls `ExprUseWalker#walkPat`. Please consider reusing `ExprUseWalker#walkPat`: this method is aimed to be general-purpose to analyze any pattern bindings.

src/main/kotlin/org/rust/lang/core/types/analysis/mutability/UnusedMutability.kt
```diff
@@ -0,0 +1,186 @@
...
+        is RsPatRef -> mut != null
+        is RsPatTupleStruct -> patList.any { it.isMutable() }
+        is RsPatTup -> patList.any { it.isMutable() }
+        is RsPatField -> patFieldFull?.pat.isMutable() || patBinding.isMutable()
+        is RsPatIdent -> pat.isMutable() || patBinding.isMutable()
+        else -> false
+    }
+}
+
+private fun RsPatBinding?.isMutable(): Boolean {
```
ortem, line 164: Consider using `org.rust.lang.core.psi.ext.RsPatBindingKt#getMutability` (`this.mutability.isMut`)

ortem:
As I wrote, it would be great to check that an immutable variable is assigned no more than once. Note that the presence of initialization right inside the declaration does not matter: we just want to ensure that the variable is assigned only once.

However, such inspection should not be lint (since it's a compile error) and therefore should be very separate from "unused mutability" inspection. There is a very related `RsReassignImmutableInspection` which is currently not flow-sensitive. Another way of implementing this is to use our borrow checker and get rid of `RsReassignImmutableInspection` (I haven't thought about that much, but it looks like a good solution).

As I mentioned in the review, it's better to reuse `ExprUseWalker` instead of duplicating the patterns processing in`isMutableContext`. You can also reuse the parts of our "borrow checker": `GatherLoans` may work for collecting mutable loans (e.g. `foo(&mut bar)`). It would be useful to look into the old rust'c implementation https://github.com/rust-lang/rust/commit/d1985c9dd06379b8a173239e90007db926ec052e and maybe the [current one](https://github.com/rust-lang/rust/blob/master/compiler/rustc_mir/src/borrow_check/mod.rs#L408). A good idea for all such inspections is to look into rustc's implementation and use it as a reference (e.g. [liveness.rs](https://github.com/rust-lang/rust/blob/master/compiler/rustc_passes/src/liveness.rs) as well).

Kobzol:
Okay to be honest all this seems like too much information at once, combined with the fact that I haven't touched this PR for 6 months :sweat_smile: Could we split this into smaller tasks? If so, what should I begin with?

Also, should we wait for https://github.com/intellij-rust/intellij-rust/pull/6718 to be merged to avoid monstrous merge conflicts?

Kobzol:
@ortem Should I try to combine the unused mutability information inside `Liveness` by putting more information inside `FlowedLivenessData`, instead of creating a separate data-flow context and analysis for unused mutability?

Issue #5743, opened 2020-07-15
Parameter specified as non-null is null: method org.rust.clion.valgrind.RsValgrindConfigurationExtension


maxim-banaev:
## Environment

* **IntelliJ Rust plugin version:** 0.3.126.3220-202
* **Rust toolchain version:** 1.27.1
* **IDE name and version:** CLion 2020.2 Beta Build #CL-202.6397.12
* **Operating system:** OS X 10.15.5

## Problem description

Exception while running helloWorld application  with valgrind memcheck:

```
java.lang.reflect.InvocationTargetException
	at jdk.internal.reflect.GeneratedMethodAccessor56.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.intellij.execution.process.ProcessHandler$2.invoke(ProcessHandler.java:246)
	at com.sun.proxy.$Proxy18.startNotified(Unknown Source)
	at com.intellij.execution.process.ProcessHandler.startNotify(ProcessHandler.java:57)
	at com.intellij.execution.process.BaseOSProcessHandler.startNotify(BaseOSProcessHandler.java:106)
	at com.intellij.execution.impl.ExecutionManagerImpl$doStartRunProfile$startRunnable$1$2$1.run(ExecutionManagerImpl.kt:227)
	at com.intellij.ui.AppUIUtil.invokeLaterIfProjectAlive(AppUIUtil.java:189)
	at com.intellij.execution.impl.ExecutionManagerImpl$doStartRunProfile$startRunnable$1$2.accept(ExecutionManagerImpl.kt:210)
	at com.intellij.execution.impl.ExecutionManagerImpl$doStartRunProfile$startRunnable$1$2.accept(ExecutionManagerImpl.kt:60)
	at org.jetbrains.concurrency.AsyncPromise$onSuccess$1.accept(AsyncPromise.kt:84)
	at org.jetbrains.concurrency.AsyncPromise$onSuccess$1.accept(AsyncPromise.kt:16)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:859)
	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:837)
[...]

PR #5744, opened 2020-07-15
FIX: set expired Valgrind callback parameter to nullable

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5743


ortem:
Thank you!
bors r+

PR #5747, opened 2020-07-15
RUN: fold external stack traces

Kobzol:
This PR adds folding for non-project related stacktraces:
image

The implementation needs some refactoring, I first wanted to do a sanity check if you think that this approach is valid.
The console folder is a non-dynamic extension, is there a dynamic alternative?

Related issue: https://github.com/intellij-rust/intellij-rust/issues/4514

### DIFF
--- a/src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
+++ b/src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
@@ -1,0 +1,58 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.console
+
+import com.intellij.execution.ConsoleFolding
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.io.FileUtil
+import com.intellij.openapi.util.text.StringUtil.pluralize
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.cargo.runconfig.filters.FilterUtils
+import org.rust.cargo.runconfig.filters.RegexpFileLinkFilter
+import org.rust.cargo.runconfig.filters.RsBacktraceFilter
+import org.rust.cargo.runconfig.filters.RsBacktraceItemFilter
+import org.rust.lang.core.resolve.resolveStringPath
+
+/**
+ * Folds backtrace items (function names and source code locations) that do not belong to the
+ * user's workspace.
+ */
+class RsConsoleFolding : ConsoleFolding() {
+    override fun getPlaceholderText(project: Project, lines: List<String>): String? {
+        // We assume that each stacktrace record has two lines (function name and source code location).
+        // Single line folds also cannot be re-folded after they are opened as of 2020.2
+        if (lines.size < 2) return null
+
+        val count = lines.size / 2
+        val callText = pluralize("call", count)
+        return "<$count internal $callText>"
+    }
+
+    override fun shouldFoldLine(project: Project, line: String): Boolean {
+        val functionNameFound = project.cargoProjects.allProjects.any {
+            val functionName = RsBacktraceItemFilter.parseBacktraceRecord(line)?.functionName ?: return@any false
+            val func = FilterUtils.normalizeFunctionPath(functionName)
+            val workspace = it.workspace ?: return@any false
+            val (_, pkg) = resolveStringPath(func, workspace, project) ?: return@any true
+
+            pkg.origin != PackageOrigin.WORKSPACE
+        }
+        if (functionNameFound) return true
+
+        return project.cargoProjects.allProjects.any {
+            val dir = (it.workspaceRootDir ?: it.rootDir) ?: return@any false
+            val filter = RegexpFileLinkFilter(project, dir, RsBacktraceFilter.LINE_REGEX)
+
+            val filePath = filter.matchLine(line)?.groups?.get(1)?.value ?: return@any false
+            val systemIndependentPath = FileUtil.toSystemIndependentName(filePath)
+
+            dir.findFileByRelativePath(systemIndependentPath) == null
+        }
+    }
+
+    override fun shouldBeAttachedToThePreviousLine(): Boolean = false
+}
--- a/src/test/kotlin/org/rust/ide/console/RsConsoleFoldingTestBase.kt
+++ b/src/test/kotlin/org/rust/ide/console/RsConsoleFoldingTestBase.kt
@@ -1,0 +1,108 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.console
+
+import com.intellij.execution.impl.ConsoleViewImpl
+import com.intellij.execution.process.NopProcessHandler
+import com.intellij.execution.process.ProcessHandler
+import com.intellij.execution.ui.ConsoleViewContentType
+import com.intellij.openapi.util.Disposer
+import org.rust.RsTestBase
+import org.rust.fileTreeFromText
+
+abstract class RsConsoleFoldingTestBase : RsTestBase() {
+    private lateinit var console: ConsoleViewImpl
+
+    public override fun setUp() {
+        super.setUp()
+
+        console = createConsole()
+        console.setUpdateFoldingsEnabled(true)
+    }
+
+    public override fun tearDown() {
+        try {
+            Disposer.dispose(console)
+        } catch (e: Throwable) {
+            addSuppressedException(e)
+        } finally {
+            super.tearDown()
+        }
+    }
+
+    protected fun doFoldingTest(tree: String, text: String) {
+        fileTreeFromText(tree).create()
+
+        val (input, expectedFolds) = createConsoleFoldInput(text)
+
+        val console = console
+        console.print(input, ConsoleViewContentType.NORMAL_OUTPUT)
+        console.flushDeferredText()
+
+        val editor = console.editor
+        val regions = editor.foldingModel.allFoldRegions
+
+        val expectedText = renderTextWithFolds(input, expectedFolds)
+        val actualText = renderTextWithFolds(input,
+            regions.map { Fold(it.placeholderText, it.startOffset, it.endOffset) })
+        assertEquals(expectedText, actualText)
+    }
+
+    private fun createConsole(): ConsoleViewImpl {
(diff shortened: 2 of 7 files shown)


src/main/kotlin/org/rust/ide/console/RsConsoleFolder.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.cargo.runconfig.filters.RegexpFileLinkFilter
+import org.rust.cargo.runconfig.filters.RsBacktraceFilter
+import org.rust.cargo.runconfig.filters.RsBacktraceItemFilter
+import org.rust.lang.core.resolve.resolveStringPath
+import java.io.File
+
+class RsConsoleFolder : ConsoleFolding() {
+    override fun getPlaceholderText(project: Project, lines: MutableList<String>): String? {
+        return "<" + lines.size + " internal calls>"
```
mchernyavsky, line 15: 1 call = 2 lines (not sure it's always true).

image

Kobzol:
> It is dynamic since 2020.2.

Nice!

Should I refactor the output parsing functionality into a separate file/class? Also, is there a way to test this?

mchernyavsky:
> Should I refactor the output parsing functionality into a separate file/class?

Yes, it would be nice.

> Also, is there a way to test this?

I don't see any test for console foldings in the platform, so most likely the answer is no :(

Undin:
About tests: we can try to use the same approach as [here](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-tests/testSrc/com/intellij/execution/impl/ConsoleViewImplTest.java#L626)

src/main/kotlin/org/rust/ide/console/RsConsoleFolder.kt
```diff
@@ -0,0 +1,45 @@
+package org.rust.ide.console;
```
mchernyavsky, line 1: `check-license.sh` failed on CI.

Kobzol:
Refactored the code slightly and added basic tests.

src/main/kotlin/org/rust/cargo/runconfig/filters/Utiils.kt
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Normalizes function path:
+ * - Removes angle brackets from the element path, including enclosed contents when necessary.
+ * - Removes closure markers.
+ * Examples:
+ * - <core::option::Option<T>>::unwrap -> core::option::Option::unwrap
+ * - std::panicking::default_hook::{{closure}} -> std::panicking::default_hook
+ */
+fun normalizeFunctionPath(function: String): String {
```
mchernyavsky, line 16: Let's move it, for example, to `FilterUtils` object class so that the `normalizeFunctionPath` function doesn't show up in the completion list.

mchernyavsky: `Utiils.kt` - typo in the file name.

src/main/kotlin/org/rust/ide/console/RsConsoleFolder.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.rust.cargo.runconfig.filters.RegexpFileLinkFilter
+import org.rust.cargo.runconfig.filters.RsBacktraceFilter
+import org.rust.cargo.runconfig.filters.RsBacktraceItemFilter
+import org.rust.cargo.runconfig.filters.normalizeFunctionPath
+import org.rust.lang.core.resolve.resolveStringPath
+import java.io.File
+import kotlin.math.max
+
+class RsConsoleFolder : ConsoleFolding() {
+    override fun getPlaceholderText(project: Project, lines: MutableList<String>): String? {
```
mchernyavsky, line 23:
```suggestion
    override fun getPlaceholderText(project: Project, lines: List<String>): String {
```

src/test/kotlin/org/rust/ide/console/RsConsoleFoldingTestBase.kt
```diff
@@ -0,0 +1,112 @@
...
+    private var myConsole: ConsoleViewImpl? = null
+
+    public override fun setUp() {
+        super.setUp()
+
+        val console = createConsole()
+        console.setUpdateFoldingsEnabled(true)
+        myConsole = console
+
+        val point: ExtensionPoint<ConsoleFolding> = ConsoleFolding.EP_NAME.point
```
mchernyavsky, line 28: The `point` is available since 2020.2.

Kobzol: Is there an alternative in 2020.1, or should I disable the tests for 2020.1?

src/test/kotlin/org/rust/ide/console/RsConsoleFoldingTestBase.kt
```diff
@@ -0,0 +1,112 @@
...
+                var currentText: String? = null
+                var currentStartOffset: Int? = null
+                for (line in text.trimIndent().lineSequence()) {
+                    when {
+                        line.startsWith("//foldstart") -> {
+                            currentText = line.removePrefix("//foldstart ").trim()
+                            currentStartOffset = offset
+                        }
+                        line.startsWith("//foldend") -> {
+                            assertNotNull(currentText)
```
Undin, line 93: Do you really need this check? You use `!!` on next line anyway. Note, you don't do it for `currentStartOffset` variable. If you still want to do it, let's better use `check(currentText  != null)`. It uses kotlin contracts under the hood and it allows compiler to make smart cast to not null value after the check

Kobzol: It's just a sanity check if the test input is correct. I used `check`, as you hinted.

src/test/kotlin/org/rust/ide/console/RsConsoleFoldingTestBase.kt
Kobzol: I couldn't get the folding in tests to work, so I tried a lot of things :D In the end it turned out to be caused by missing `setUpdateFoldingsEnabled`, but then I forgot to remove the extension point.

Kobzol:
Right, it won't solve the problem completely (for example the `<14 internal calls>` in your example will still contain the firstl ine of the split record), but it's a good idea.

src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        val locationFilters = project.cargoProjects.allProjects.mapNotNull {
+            val dir = it.workspaceRootDir ?: it.rootDir
+            RegexpFileLinkFilter(project, dir ?: return@mapNotNull null, RsBacktraceFilter.LINE_REGEX)
+        }
+        return locationFilters.any {
+            val fileName = it.matchLine(line)?.groups?.get(1)?.value ?: return@any false
+            var platformNeutralName = fileName.replace(File.separatorChar, '/')
+            if (isUnitTestMode) {
```
Undin, line 51: Let's drop this condition and adjust test cases instead.
Our default test workspace has custom layout for symplicity.
Consider that you have the following Cargo.toml:
```toml
[lib]
name = "test-package"
path = "lib.rs"

[[bin]]
name = "test-package"
path = "main.rs"
```

src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
```diff
@@ -0,0 +1,60 @@
...
+                // rootDir in unit tests points to a wrong directory
+                platformNeutralName = platformNeutralName.removePrefix("src/")
+            }
+            it.cargoProjectDirectory.findFileByRelativePath(platformNeutralName) == null
```
Undin, line 55: Don't like that we made `cargoProjectDirectory` public only for this line.
You already have `dir` value in `mapNotNull`.
You can collapse `mapNotNull` and `any` calls into single `any` and use `dir`

src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
```diff
@@ -0,0 +1,60 @@
...
+        val callText = pluralize("call", count)
+        return "<$count internal $callText>"
+    }
+
+    override fun shouldFoldLine(project: Project, line: String): Boolean {
+        val backtraceFilters = project.cargoProjects.allProjects.map {
+            RsBacktraceItemFilter(project, it.workspace)
+        }
+        if (backtraceFilters.any {
+                val functionName = it.parseBacktraceRecord(line)?.functionName ?: return@any false
```
Undin, line 34: Hm, looks like you can move `parseBacktraceRecord` into companion object and avoid creation of `RsBacktraceItemFilter`

src/main/kotlin/org/rust/cargo/runconfig/filters/Utils.kt
```diff
@@ -0,0 +1,58 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.runconfig.filters
+
+object FilterUtils {
```
Undin, line 8: Let's rename file into `FilterUtils`

src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
Kobzol: Ooh, now it makes sense. Thanks.

src/main/kotlin/org/rust/ide/console/RsConsoleFolding.kt
```diff
@@ -0,0 +1,58 @@
...
+            pkg.origin != PackageOrigin.WORKSPACE
+        }
+        if (functionNameFound) return true
+
+        return project.cargoProjects.allProjects.any {
+            val dir = (it.workspaceRootDir ?: it.rootDir) ?: return@any false
+            val filter = RegexpFileLinkFilter(project, dir, RsBacktraceFilter.LINE_REGEX)
+
+            val filePath = filter.matchLine(line)?.groups?.get(1)?.value ?: return@any false
+            val systemIndependentPath = filePath.replace(File.separatorChar, '/')
```
Undin, line 51:
```suggestion
            val systemIndependentPath = FileUtil.toSystemIndependentName(filePath)
```

Kobzol:
bors r=undin

PR #5771, opened 2020-07-21
INT: add intention to change item visibility

Kobzol:
This PR adds a simple intention that toggles the visibility of some item. I reused the code of `MakePublicFix`, maybe the shared code could go somewhere else? Now I moved it into the intention.

visibility


matklad:
I think it makes sense to default to `pub(crate)` visibility here -- that one is the most annoying to type, typing just `pub` is not that bad.

Kobzol:
In Kotlin/Java it offers three intentions for private items (`Make public/internal/protected`). Maybe we should also offer `Make pub` and `Make pub(crate)`?

Kobzol:
Rebased onto `master` and changed `pub` to `pub(crate)`.

src/main/kotlin/org/rust/ide/intentions/ChangeVisibilityIntention.kt
```diff
@@ -0,0 +1,90 @@
...
+            // insert `pub` element before them.
+            while (true) {
+                val prevNonCommentSibling = anchor.getPrevNonCommentSibling()
+                if (!(prevNonCommentSibling is RsOuterAttr || prevNonCommentSibling == null)) {
+                    anchor = prevNonCommentSibling
+                } else {
+                    break
+                }
+            }
+            if (!withinOneCrate) {
```
vlad20012, line 69: I think, inverting this condition could make the meaning of `withinOneCrate` more clear :)

src/main/kotlin/org/rust/ide/intentions/ChangeVisibilityIntention.kt
```diff
@@ -0,0 +1,90 @@
...
+        text = "Make$name $newVisibility"
+
+        return Context(visibleElement)
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: Context) {
+        if (ctx.element.isPublic) {
+            makePrivate(ctx.element)
+        } else {
+            makePublic(ctx.element, true)
```
vlad20012, line 37: Let's add an explicit parameter name before `true`. Currently, it's not clear what this `true` means in this context

vlad20012 (commented):
Btw I agree that two intentions would be ok. Just like in Kotlin! :)

Kobzol:
I rebased the original PR and included an additional commit that uses separate intentions for private/pub/pub(crate) visibility. Let me know if you like it, if not, I can remove the commit and merge it as it was.

src/main/kotlin/org/rust/ide/intentions/visibility/ChangeVisibilityIntention.kt
```diff
@@ -0,0 +1,102 @@
...
+abstract class ChangeVisibilityIntention : RsElementBaseIntentionAction<ChangeVisibilityIntention.Context>() {
+    override fun getFamilyName(): String = "Change item visibility"
+
+    abstract val visibility: String
+    abstract fun isApplicable(element: RsVisibilityOwner): Boolean
+
+    data class Context(val element: RsVisibilityOwner)
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        val visibleElement = element.ancestorStrict<RsVisibilityOwner>() ?: return null
```
vlad20012, line 25: Oh, I forget one more comment. Let's restrict the intention range to function modifiers (just like in Kotlin) and use `checkAvailableInSelectionOnly` to test it

Kobzol:
Added check to only enable the intentions at the item keyword and "to the left" of it.

vlad20012 (approved):
:+1: Thanks!
bors r+

Kobzol:
Some tests were not updated, so bors would fail.

bors r=vlad20012

PR #5785, opened 2020-07-23
GRD: update native debugger plugin

Undin:
Previous versions of native debugger plugin contain wrong versions of lldb binaries


Undin:
@lancelote Could you check that debugging works properly in the latest IDEA Ultimate 2020.2 with the latest Native Debugging Support plugin, please?

Issue #5759, opened 2020-07-19
Detect E0364/E0365 errors


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.127.3225-202-nightly
* **Rust toolchain version:** 1.45.0 (5c1f21c3b 2020-07-13) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2 Beta Ultimate Edition (IU-202.6397.20)
* **Operating system:** macOS 10.13.6

## Problem description
Currently, [E0364](https://doc.rust-lang.org/error-index.html#E0364) and [E0365](https://doc.rust-lang.org/error-index.html#E0365) are not detected by the plugin as compilation errors. It would be great to highlight cases when a reexported item is not public and provide quick fix to add `pub` modified

## Steps to reproduce
```rust
mod foo {
    mod bar {}
    pub use bar as baz;  // <--- error[E0365]: `bar` is private, and cannot be re-exported
    struct S;
    pub use S as T; // <--- error[E0364]: `S` is private, and cannot be re-exported
}
```

PR #5791, opened 2020-07-24
INSP: add inspection for E0364/E0365 errors

Kobzol:
The implementation differs from rustc in that the error annotation doesn't include the reexport alias, if there is any:
```rust
mod foo {
    mod bar {}
    pub use <error>bar as baz</error>; // rustc annotation
    pub use <error>bar</error> as baz; // plugin annotation
}
```

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5759

TODO:
- [x] run real project analysis test

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -133,10 +133,62 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
 
     private fun checkUseSpeck(holder: RsAnnotationHolder, useSpeck: RsUseSpeck) {
         if (useSpeck.isStarImport || useSpeck.useGroup != null) return
+
+        checkDuplicateImport(holder, useSpeck)
+        checkReexports(holder, useSpeck)
+    }
+
+    private fun checkDuplicateImport(holder: RsAnnotationHolder, useSpeck: RsUseSpeck) {
         val duplicates = holder.currentAnnotationSession.duplicatesByNamespace(useSpeck.containingMod, false)
         if (useSpeck.namespaces.any { useSpeck in duplicates[it].orEmpty() }) {
             val identifier = PsiTreeUtil.getDeepestLast(useSpeck)
             RsDiagnostic.DuplicateImportError(identifier).addToHolder(holder)
+        }
+    }
+
+    private fun checkReexports(holder: RsAnnotationHolder, useSpeck: RsUseSpeck) {
+        val item = useSpeck.ancestorStrict<RsUseItem>() ?: return
+        if (item.visibility == RsVisibility.Private) return
+
+        val path = useSpeck.path ?: return
+        val targets = path.reference?.multiResolve()?.filterIsInstance<RsItemElement>() ?: emptyList()
+
+        // targets not accessible from the use element will be handled by E0603
+        val visibleTargets = targets.filter { it.isVisibleFrom(useSpeck.containingMod) }
+        val invalidTargets = visibleTargets.filter {
+            !canBeReexported(item.visibility, it)
+        }
+
+        // if at least one item can be re-exported or nothing is invalid, allow the reexport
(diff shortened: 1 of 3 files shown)


Kobzol:
Rebased over `master` to run regression check on CI.

src/test/kotlin/org/rust/ide/annotator/RsInvalidReexportErrorAnnotatorTest.kt
```diff
@@ -0,0 +1,156 @@
...
+
+            pub use A as B;
+        }
+        mod bar {
+            struct A { a: u32 }
+            pub const A: u32 = 0;
+
+            pub use A as B;
+        }
+    """)
```
Undin, line 77: Let's add a similar test but when items have the same namespace. It should produce only [E0428]

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -133,13 +133,59 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+        val targets = path.reference?.multiResolve()?.filterIsInstance<RsItemElement>() ?: emptyList()
+
+        // targets not accessible from the use element will be handled by E0603
+        val visibleTargets = targets.filter { it.isVisibleFrom(useSpeck.containingMod) }
+        val invalidTargets = visibleTargets.filter {
+            !canBeReexported(item.visibility, it.visibility)
+        }
+
+        // if at least one item can be re-exported, allow the reexport
+        if (invalidTargets.size < targets.size) return
```
Undin, line 163: nit: maybe it's worth to add `|| invalidTargets.isEmpty()` condition to avoid `firstOrNull()?.let` below?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -133,13 +133,59 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+
+        invalidTargets.firstOrNull()?.let {
```
Undin, line 165: nit: I'd prefer to get rid of `let` here

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -133,13 +133,59 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+            val name = if (path.referenceName == "self" && useSpeck.context is RsUseGroup) {
```
Undin, line 166: nit: let's extract `useSpeck.context` into variable here

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -133,13 +133,64 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+
+        RsDiagnostic.InvalidReexport(path.referenceNameElement, name, invalidTarget).addToHolder(holder)
+    }
+
+    private fun canBeReexported(reexportVisibility: RsVisibility, item: RsItemElement): Boolean {
+        return when (val itemVisibility = item.visibility) {
+            is RsVisibility.Private ->
+                when (reexportVisibility) {
+                    is RsVisibility.Private -> true
+                    is RsVisibility.Restricted -> reexportVisibility.inMod == item.containingMod
```
Undin, line 181: I support it still a cause of false-positives
```rust
mod foo {
    mod bar {}
    mod baz {
        pub(in crate::foo::baz) use crate::foo::bar as qwe;
    }
}
```


image

Kobzol: Take #2 :D Now it should be hopefully general. I understood private visibility wrong before.

Undin (approved):
Great!
Thank you!
bors r+

PR #5801, opened 2020-07-26
Fix navigation to derive items in 2020.2

### LINKED ISSUE #5531: Goto definition action doesn't navigate to the derive items
## Environment

* **IntelliJ Rust plugin version:** 0.2.124.SNAPSHOT-202-dev
* **IDE name and version:** CLion 2020.2 EAP (CL-202.5103.16)

## Problem description
Before 2020.2 `Goto definition` action navigates you to meta item in `derive` attribute if it provides generation of the corresponding item (function, method, etc).
But it doesn't work since 2020.2.

## Steps to reproduce
```rust
#[derive(Default)]
struct Foo(i32);

fn main() {
    Foo::default();
}
```

vlad20012:
Fixes #5531

The old hack doesn't work anymore, I invented a new one

### DIFF
--- a/src/main/resources/META-INF/rust-core.xml
+++ b/src/main/resources/META-INF/rust-core.xml
@@ -982,6 +982,11 @@
                         dynamic="true"/>
     </extensionPoints>
 
+    <applicationListeners>
+        <listener class="org.rust.ide.navigation.goto.RsGoToDeclarationRunningService$Listener"
+                  topic="com.intellij.openapi.actionSystem.ex.AnActionListener" />
+    </applicationListeners>
+
     <projectListeners>
         <listener class="org.rust.lang.core.macros.MacroExpansionManagerWaker"
                   topic="org.rust.cargo.project.model.CargoProjectsService$CargoProjectsListener" />
--- a/src/main/kotlin/org/rust/ide/navigation/goto/RsGoToDeclarationRunningService.kt
+++ b/src/main/kotlin/org/rust/ide/navigation/goto/RsGoToDeclarationRunningService.kt
@@ -1,0 +1,52 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.navigation.goto
+
+import com.intellij.codeInsight.navigation.actions.GotoDeclarationAction
+import com.intellij.codeInsight.navigation.actions.GotoDeclarationOnlyAction
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.actionSystem.ex.AnActionListener
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+import com.intellij.openapi.progress.ProgressManager
+import com.intellij.openapi.progress.util.ProgressWindow
+import com.intellij.openapiext.isDispatchThread
+
+/**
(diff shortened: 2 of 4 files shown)


src/main/kotlin/org/rust/ide/navigation/goto/RsCurrentActionService.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.intellij.codeInsight.navigation.actions.GotoDeclarationOnlyAction
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.actionSystem.ex.AnActionListener
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+
+@Service
+class RsCurrentActionService {
```
Undin, line 18: Let's rename it somehow (for example, to `RsGoToDeclarationService`) or move out from `navigation.goto` package

src/main/kotlin/org/rust/ide/navigation/goto/RsCurrentActionService.kt
```diff
@@ -0,0 +1,39 @@
...
+    var isGoToDeclarationAction: Boolean = false
+        private set
+
+    private class Listener : AnActionListener {
+        override fun beforeActionPerformed(action: AnAction, dataContext: DataContext, event: AnActionEvent) {
+            if (action is GotoDeclarationAction || action is GotoDeclarationOnlyAction) {
```
Undin, line 24: Let's extract this condition into property/function not to duplicate logic

Something like

```kotlin
    private val AnAction.isGoToDeclarationAction: Boolean get() = 
        this is GotoDeclarationAction || this is GotoDeclarationOnlyAction
```

src/main/kotlin/org/rust/ide/navigation/goto/RsCurrentActionService.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.actionSystem.ex.AnActionListener
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+
+@Service
+class RsCurrentActionService {
+    var isGoToDeclarationAction: Boolean = false
```
Undin, line 19: Let's add some synchronization for this property. As I can see, we use it from different threads and call setter not under write action

src/main/kotlin/org/rust/ide/navigation/goto/RsCurrentActionService.kt
```diff
@@ -0,0 +1,39 @@
...
+        private set
+
+    private class Listener : AnActionListener {
```
Undin, line 22: Let's add `@Suppress("unused")` to avoid false positive `unused` annotation

src/main/resources/META-INF/rust-core.xml
```diff
@@ -983,6 +983,11 @@
                   topic="org.rust.cargo.project.model.CargoProjectsService$CargoProjectsListener" />
     </projectListeners>
 
+    <applicationListeners>
```
Undin, line 986: nit: I'd personally move `applicationListeners` before `projectListeners` because they have a "wider" scope. But it's up to you

vlad20012:
bors r=undin

PR #5816, opened 2020-07-27
Refactoring of auto-import related code

Undin:
These changes extract auto-import related code from `AutoImportFix` into `org.rust.ide.utils.import` package since this code widely used in different IDE features.
Also, this code is split into several separate files

As a result, a lot of code doesn't depend on `AutoImportFix`


Undin:
@mchernyavsky I changed nothing in implementation logic, just code reordering. But I want to be verified by someone. Probably, you have other suggestions

Undin:
CC @dima74

Issue #5814, opened 2020-07-27
Extract function doesn't preserve unsafety


panstromek:
## Problem description
Extracting  block of unsafe code from unsafe function doesn't preserve its unsafety (ie. the code is not wrapped in unsafe block and the extracted function is also  not unsafe.

## Steps to reproduce
Before:
```rust
unsafe fn this_is_unsafe (ptr : *const u32) -> u32 {
    <span>*ptr<span>
}
```
After
```rust
// this code doesn't compile anymore 
unsafe fn this_is_unsafe (ptr : *const u32) -> u32 {
    extracted(ptr)
}

fn extracted(ptr: *const u32) -> u32 {
    *ptr
}
```

PR #5819, opened 2020-07-28
REF: add unsafe modifier to functions extracted from unsafe functions

Kobzol:
This PR adds `unsafe` to functions extracted from `unsafe` functions. Currently the `unsafe` is added unconditionally, should I improve that?

The extracted statements would have to be scanned for unsafe operations that are not inside an `unsafe` block. Since pretty much the same code is already in `RsUnsafeExpressionAnnotator`, it should be refactored somehow to share the logic.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5814


panstromek:
I think the correct behaviour is to make the extracted function always unsafe - if it's extracted from unsafe block or unsafe function (which is what this PR does).

Unsafe function doesn't even need to contain any unsafe operation to be considered unsafe (typical  example is `Vec::set_len`), so determining if the extracted code should be considered unsafe is not possible in general, so I would expect the IDE to be extra cautious and make the function always unsafe and let the user decide if it's safe to remove the modifier.

btw. thx for such a quick fix 😉

Issue #4274, opened 2019-08-18
Disable snake case lint for `#[no_mangle]` functions


athre0z:
## Environment

* **IntelliJ Rust plugin version:** 0.2.103.2131-192
* **Rust toolchain version:** rustc 1.38.0-nightly (60960a260 2019-08-12)
* **IDE name and version:** CLion 2019.2 Build #CL-192.5728.100, built on July 23, 2019
* **Operating system:** 10.14.6 (18G84) 

## Problem description
rustc automatically disables the `non_snake_case` lint for functions with a `#[no_mangle]` attribute. The IJ plugin still displays it.

## Steps to reproduce

```rust
#[no_mangle]
pub unsafe extern "C" fn Meow() {}
//                       ~~~~ warning here
```

PR #5820, opened 2020-07-28
INSP: suppress snake_case naming lint for #[no_mangle] functions

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/4274

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsNamingInspection.kt
@@ -10,10 +10,7 @@ import com.intellij.psi.PsiNamedElement
 import com.intellij.psi.util.PsiTreeUtil
 import org.rust.ide.inspections.fixes.RenameFix
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsAbstractableOwner
-import org.rust.lang.core.psi.ext.RsConstantKind
-import org.rust.lang.core.psi.ext.kind
-import org.rust.lang.core.psi.ext.owner
+import org.rust.lang.core.psi.ext.*
 
 /**
  * Base class for naming inspections. Implements the core logic of checking names
@@ -206,6 +203,13 @@ class RsFunctionNamingInspection : RsSnakeCaseNamingInspection("Function") {
                 }
             }
         }
+
+    override fun isSuppressedFor(element: PsiElement): Boolean {
+        if (super.isSuppressedFor(element)) return true
+
+        val function = element.parent as? RsFunction ?: return false
+        return function.isExtern && function.findOuterAttr("no_mangle") != null
+    }
 }
 
 class RsMethodNamingInspection : RsSnakeCaseNamingInspection("Method") {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/inspections/RsNamingInspection.kt
```diff
@@ -41,6 +39,8 @@ abstract class RsNamingInspection(
             *fixes)
     }
 
+    open fun isLintSuppressedFor(element: PsiElement): Boolean = false
```
Undin, line 42: Note, there is `com.intellij.codeInspection.InspectionProfileEntry#isSuppressedFor(com.intellij.psi.PsiElement)`
I think we want to implement this meatho instead of introducing new API

Kobzol: Oops, thanks.

src/main/kotlin/org/rust/ide/inspections/RsNamingInspection.kt
```diff
@@ -100,6 +97,11 @@ open class RsSnakeCaseNamingInspection(elementType: String) : RsNamingInspection
         }
         return Pair(false, if (str.isEmpty()) "snake_case" else name.toSnakeCase(false))
     }
+
+    override fun isSuppressedFor(element: PsiElement): Boolean {
+        val function = element.parent as? RsFunction ?: return false
+        return function.findOuterAttr("no_mangle") != null
```
Undin, line 103: Hm, looks like we should suppress annotation only if a function has `extern` modifier. Otherwise, the compiler still produces a warning

Kobzol: You're right, I changed it.

src/main/kotlin/org/rust/ide/inspections/RsNamingInspection.kt
```diff
@@ -100,6 +97,11 @@ open class RsSnakeCaseNamingInspection(elementType: String) : RsNamingInspection
         }
         return Pair(false, if (str.isEmpty()) "snake_case" else name.toSnakeCase(false))
     }
+
+    override fun isSuppressedFor(element: PsiElement): Boolean {
+        val function = element.parent as? RsFunction ?: return false
```
Undin, line 102: Let's invoke `super.isSuppressedFor(element)` to check if this inspection is already suppressed for this element like we do in `org.rust.ide.inspections.RsLintInspection#isSuppressedFor`. Otherwise, it prevents suppressing, for example, via comment:

```rust
//noinspection RsFunctionNaming
fn Bar() {}
```

Undin: BTW maybe move this check into `RsFunctionNamingInspection`? Because it makes no sense for other inheritors

Kobzol: I didn't notice that a function-specific inspection exists, thanks.

Issue #5433, opened 2020-05-24
Iterator type hints noisy when type isn't changing


AlyoshaVasilieva:
The type hints for iterator chains are generally useful. However, sometimes I find a chain like this is useful:

```rust
vec
    .iter()
    .filter(...)
    .filter(...)
    .filter(...)
    .map(...)
    .collect();
```

The `impl Iterator<Item=&SomeType>` hints on the `filter` calls aren't very useful, since the type is the same after all of them. It would be nice if there was an option to only show hints when type has changed; in this case, they'd be shown on `iter` and `map` calls.

Kobzol:
This option is already implemented, it's just turned off by default. If you go to `Editor -> Inlay Hints -> Rust -> Chain method hints` and turn off `Show same consecutive types`, you should get the behaviour that you want.

vlad20012:
> Maybe we could compare the types based on the rendered string output instead?

Oh, it looks like we should do it this way. Except for the fact that inlay presentation is not a string... Let's summon @ortem here

ortem:
Yes, inlay presentations are not strings. A possible solution here: create a special comparator for two `Ty` objects such that `TyAdt` with `impl Iterator` are compared using their `TyAnon` representation (from `org.rust.ide.hints.RsTypeHintsPresentationFactory#tryIteratorHint`). @vlad20012 what do you think?

vlad20012:
Hmm, maybe it's easier to prepare types with `TypeFolder` - replace iterator types with `impl Iterator` and then compare them as usual

PR #5828, opened 2020-07-29
HINT: allow comparison of consecutive iterator special case types in chain method hints

Kobzol:
This PR changes the special case for iterators in `RsChainMethodTypeHintsProvider`. Now the top-level type of something that implements `Iterator` is changed to `impl Iterator` so that consecutive special.cased iterator types can be folded. Before only the rendering was changed, so consecutive types were not properly folded if they implemented the same `Iterator<Item=X>`.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5433

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/type/RsChainMethodTypeHintsProvider.kt
+++ b/src/main/kotlin/org/rust/ide/hints/type/RsChainMethodTypeHintsProvider.kt
@@ -23,12 +23,18 @@ import com.intellij.util.ui.JBUI
 import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.RsLanguage
 import org.rust.lang.core.psi.RsDotExpr
+import org.rust.lang.core.psi.RsFile
 import org.rust.lang.core.psi.RsMethodCall
-import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.RsTraitItem
 import org.rust.lang.core.psi.ext.childOfType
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/hints/type/RsChainMethodTypeHintsProvider.kt
```diff
@@ -69,40 +74,48 @@ class RsChainMethodTypeHintsProvider : InlayHintsProvider<RsChainMethodTypeHints
...
-            private fun presentTypeForMethodCall(call: RsMethodCall) {
+            private fun presentTypeForMethodCall(call: RsMethodCall, type: Ty) {
                 val project = call.project
-                val type = call.type
                 val presentation = typeHintsFactory.typeHint(type)
                 val finalPresentation = presentation.withDisableAction(project)
                 sink.addInlineElement(call.endOffset, true, finalPresentation)
             }
+
+            private fun normalizeType(type: Ty, lookup: ImplLookup, iteratorTrait: BoundElement<RsTraitItem>?): Ty {
```
vlad20012, line 110: Let's add a comment about what's happening here, like this:
/** Returns fake `impl Iterator<Item=...>` type if [type] implements `Iterator` trait */

Also, looks like you missed to check `settings.iteratorSpecialCase` somewhere

Kobzol:
Sorry for the confusion :sweat_smile:

vlad20012:
:+1: 
bors r+

PR #5837, opened 2020-07-30
TY: fix `?` try expr for complex types

vlad20012:
Previously, `foo?` expr was sometimes inferred to `<unknown>` type even if `foo` is fully inferred as `Result<...>` or `Option<...>`.

 This works for
 ```rust
 let a = vec![Some(42)].into_iter().next()??;
 ```


Undin:
bors retry

PR #5849, opened 2020-07-31
[CARGO] Include RUSTC in environment

### LINKED ISSUE #5843: toolchain location not included in PATH passed to cargo
Hello! We are big fans of this plugin in [Fuchsia](https://fuchsia.dev).

In Fuchsia, we use self-contained rust toolchains, which means that our `cargo`, `rustc`, etc. are not usually in our PATH. This seems to be a problem for this plugin because `cargo` is executed with intellij's PATH, and is then unable to find `rustc`.

Would it be possible to execute cargo with the toolchain location (specified in the plugin's settings) appended to PATH?

cc @brunowonka

tamird:
Fixes #5843.

### DIFF
--- a/src/test/kotlin/org/rust/cargo/toolchain/CargoTest.kt
+++ b/src/test/kotlin/org/rust/cargo/toolchain/CargoTest.kt
@@ -18,19 +18,19 @@ class CargoTest : RsTestBase() {
     fun `test run arguments preserved`() = checkCommandLine(
         cargo.toColoredCommandLine(project, CargoCommandLine("run", wd, listOf("--bin", "parity", "--", "--prune", "archive"))), """
         cmd: /usr/bin/cargo run --color=always --bin parity -- --prune archive
-        env: RUST_BACKTRACE=short, TERM=ansi
+        env: RUSTC=/usr/bin/rustc, RUST_BACKTRACE=short, TERM=ansi
         """, """
         cmd: C:/usr/bin/cargo.exe run --color=always --bin parity -- --prune archive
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/cargo/toolchain/CargoTest.kt
```diff
@@ -18,19 +18,19 @@ class CargoTest : RsTestBase() {
     fun `test run arguments preserved`() = checkCommandLine(
         cargo.toColoredCommandLine(project, CargoCommandLine("run", wd, listOf("--bin", "parity", "--", "--prune", "archive"))), """
         cmd: /usr/bin/cargo run --color=always --bin parity -- --prune archive
-        env: RUST_BACKTRACE=short, TERM=ansi
+        env: RUSTC=/usr/bin/rustc, RUST_BACKTRACE=short, TERM=ansi
         """, """
         cmd: C:/usr/bin/cargo.exe run --color=always --bin parity -- --prune archive
-        env: RUST_BACKTRACE=short, TERM=ansi
+        env: RUSTC=/usr/bin/rustc, RUST_BACKTRACE=short, TERM=ansi
```
mchernyavsky, line 24: Here and in the other tests:

```suggestion
        env: RUSTC=C:/usr/bin/rustc.exe, RUST_BACKTRACE=short, TERM=ansi
```

tamird: Done, thanks!

PR #5864, opened 2020-08-04
RES: Macro from local import should shadow macro from crate root

dima74:
Changes order in which macros are processed: macros from imports first, then macros from crate root scope


dima74:
bors retry

dima74:
bors retry

Issue #5868, opened 2020-08-04
`Change return type` fix should not be available on passed closure


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.128.3262-202-nightly
* **Rust toolchain version:** 1.46.0-nightly (8aa18cbdc 2020-07-08) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2 Ultimate Edition (IU-202.6397.94)
* **Operating system:** macOS 10.15.3

## Problem description
This fix doesn't make sense in case of wrong type closure passed to a function as an argument

![change_type](https://user-images.githubusercontent.com/4854600/89303574-0e709b80-d675-11ea-9dc2-fd0e100efb29.gif)

PR #5870, opened 2020-08-04
INSP: do not offer ChangeReturnTypeFix in lambdas without an explicit return type

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5868

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeReturnTypeFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeReturnTypeFix.kt
@@ -15,16 +15,16 @@ import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.presentation.render
 import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.utils.import.RsImportHelper.importTypeReferencesFromTy
-import org.rust.lang.core.psi.RsFunction
-import org.rust.lang.core.psi.RsLambdaExpr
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/ChangeReturnTypeFix.kt
```diff
@@ -84,6 +84,11 @@ class ChangeReturnTypeFix(element: RsElement, private val actualTy: Ty) : LocalQ
             val isOverriddenFn = owner is RsFunction && owner.superItem != null
             if (isOverriddenFn) return null // TODO: Support overridden items
 
+            // do not offer in lambdas passed to functions
```
mchernyavsky, line 87: It is still available in the following cases:

image

image

I'm not sure how to fix this properly. Maybe we shouldn't offer the fix if we're inside a lambda and it doesn't have a return value?

Kobzol: That seems like a better solution. If a lambda does not have an explicit return type, the return type has to be inferred from the environment, otherwise the plugin wouldn't show the error. In such case it indeed does not make sense to offer the fix, because the return type is "fixed" by the inference.

If the lambda has a return type specified, but it's wrong with respect to the surrounding context, it still won't compile after the fix is performed, but it will fix consistency of the lambda itself.

Kobzol:
bors retry

PR #5871, opened 2020-08-04
INSP: check if struct fields are not bound multiple times in struct pats [E0025]

Kobzol:
This PR inspects struct pats to check if there are not fields bound multiple times.

Related: https://github.com/intellij-rust/intellij-rust/issues/886

TODO:
- [x] Run `RsRealProjectAnalysisTest`

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -163,11 +163,34 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             if (missingFields.isNotEmpty() && patStruct.patRest == null) {
                 RsDiagnostic.MissingFieldsInStructPattern(patStruct, declaration, missingFields).addToHolder(holder)
             }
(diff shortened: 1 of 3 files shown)


vlad20012:
Please, rebase it to the latest master. Real project test now then should be run on CI.

Kobzol:
Looks like regressions on Cargo are OK.

mchernyavsky (commented):
It doesn't work for nested patterns:

```rust
struct Foo { a: Bar, b: Bar }
struct Bar { a: i32, b: i32 }

fn foo(x: Foo) {
    let Foo { a: Bar { a, b }, b: Bar { a, b } } = x;
}
```

Kobzol:
@mchernyavsky that is actually a different error (`E0416`), I'm planning to add support  for that one yet.
The difference is:
- `E0025` - you use the exact same field more than once (this is what I handle in this PR)
- `E0416` - you use the same identifier for more than one field in the whole binding (I will do that in another PR)

mchernyavsky (approved):
Thanks!

bors r+

mchernyavsky:
bors retry

Issue #5511, opened 2020-06-05
Inline function refactoring improvements


vlad20012:
#4928

- [ ] Inline functions with never-type caused exit points. E.g. this function could be inlined:
```rust
fn foo() -> i32 {
    if (true) {
        return 1;
    } else {
        panic!()
    }
}
```
- [ ] Substitute parameter values. Looks like now it works with `self` parameter. Note that if an argument is a non-trivial expression and the parameter is used multiple time, it can't be substituted directly: an additional variable should be created. See how Kotlin behaves in this case
```kotlin
fun foo(i: Int): Int = i + bar() + i
fun bar(): Int = 5
val a = foo(bar())
      //^ inline foo
```

Kobzol:
Normally if there's a `return` statement, either there are more return points (and the refactoring is disabled), or it's at the end of the function, i.e. it's not inside any condition/loop block. This assumption is not valid if there are any diverging expressions that follow after that `return`:
```rust
fn foo() {
    let x = test(1);
}
fn /*caret*/test(x: u32) -> i32 {
    if (x > 2) {
        return 1;
    }
    panic!();
}
```
turns into
```rust
fn foo() {
    if (x > 2) {
        ;
    }
    panic!();
    let x = 1;
}
```
If I just ignore the diverging return types.

I tried to use any (potential) blocks containing the `return` as the inlined return value and move all items after the first `return` after the inlined return value: 
```rust
fn foo() {
    let x = if (x > 2) {
        1
    }
    panic!();
}
```
Which is better, but the problem with that is that the code here is now missing `else` and therefore it does not compile. Adding `else { .. rest of code after return }` will not help, since there may be nested conditions in side the block containing `return`.

panstromek:
Today I found out that inlining function from a different module doesn't delete corresponding import, so if the function was removed, the code no longer compiles after the refactoring. That's also another improvement to the list ;)

SaarYogev:
@vlad20012 I created the Kotlin equivilent of the first example:
```kotlin
private fun caller() {
    var num = checkPanicReturnInline()
    print(num)
}

private fun checkPanicReturnInline(): Int {
    if (true) {
        return 1
    } else {
        throw Exception()
}
```

Trying to inline this throws an exception: `Cannot Perform refacotoring. Inline Function is not supported for functions with return statements not at the end of the body.`
I assume you mean suppoting the equivilent of this, instead. Here there's a more definetive exit point, and the inline works:
```kotlin
[...]

PR #5883, opened 2020-08-07
Inline function refactoring improvements

SaarYogev:
Adding inline refactoring imnprovements, as asked in #5511 .

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/inline/RsInlineFunctionProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inline/RsInlineFunctionProcessor.kt
@@ -136,14 +136,14 @@ class RsInlineFunctionProcessor(
 
     companion object {
         fun doesFunctionHaveMultipleReturns(fn: RsFunction): Boolean {
-            var entryCount = 0
+            val entryPoints =  mutableListOf<ExitPoint>()
             val sink: (ExitPoint) -> Unit = {
-                if (it !is ExitPoint.TryExpr && it !is ExitPoint.DivergingExpr) {
-                    ++entryCount
+                if (it !is ExitPoint.TryExpr) {
+                    entryPoints.add(it)
                 }
             }
             ExitPoint.process(fn, sink)
-            return entryCount > 1
+            return entryPoints.count() > 1 && entryPoints.dropLast(1).any { it is ExitPoint.Return }
         }
 
         fun isFunctionRecursive(fn: RsFunction): Boolean {
--- a/src/main/kotlin/org/rust/ide/refactoring/inline/RsInlineFunctionProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inline/RsInlineFunctionProcessor.kt
@@ -11,6 +11,7 @@ import com.intellij.openapi.util.Ref
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiReference
 import com.intellij.psi.codeStyle.CodeStyleManager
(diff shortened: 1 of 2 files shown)


Kobzol:
The inlining of parameters needs to be slightly more complex to handle arguments with side effects and complex patterns. I suggest that we leave it for another PR and this PR would focus only on the divergent return types, to make it easier for review.

```rust
fn test1() {
    let a = foo(bar()); // inline foo
    let a = bar() + bar() + bar(); // produced by PR
    let x = bar();
    let a = x + bar() + x; // should be produced if bar() has side effects (which it does, it's a function call)
}
fn test2() {
    let a = foo(&1 + 2); // inline foo
    let a = 1 + 2 + bar() + 1 + 2; // produced by PR
    let a = &1 + 2 + bar() + &1 + 2; // should be produced
}
fn foo(i: u32) -> u32 {
    i + bar() + i
}
fn bar() -> u32 {
    5
}

fn test3() {
    let x = foo2((1, 1)); // inline foo2
    let x = a + b; // produced by PR
    let (a, b) = (1, 1);
    let x = a + b; // should be produced (if a/b already existed in scope, their name should be changed)
}
fn foo2((a, b): (u32, u32)) -> u32 {
    println!("{} {}", a, b);
    a + b
}
```

SaarYogev:
What exactly do you want in a separate PR? only the part about inlining function parameters correctly, or also the part about removing use statement when inlining a function from a different module?

Kobzol:
Those are three separate things (exit point improvements, parameter inlining and `use` statement removal), so I think that they should be in 3 PRs. The `use` statement removal also needs to be more complex, for example currently it doesn't work if there are no callers of the method or if the `use` path of the function uses whitespace or is nested:
```rust
mod a {
  pub fn foo() {} /*inline*/
}

use a::foo; // not removed
```

```rust
mod a {
    pub fn foo() {} /*inline*/
    fn bar() {
        foo();
    }
}

use a:: foo; // not removed
use a::{foo]; // not removed
```


The `use`s of the inlined function should probably be found using the `ReferencesSearch` API (that's done already, but currently it only looks for calls to the function, it should also include `use`s).

SaarYogev:
Will split, already have it split commit-wise for this case.  About using ReferenceSearch to remove the `use` statements:
I actually don't use `ReferenceSearch`, I compared them using text like so (`RsInlineFunctionProcessor:228`):
```rust
private fun removeFunctionUseStatements(function: RsFunction) {
    ReferencesSearch.search(function)
    caller.containingFile.children.filterIsInstance<RsUseItem>().forEach { useStatement: RsUseItem ->
        if (function.crateRelativePath?.trimStart(':') == useStatement.useSpeck?.text) {
            useStatement.delete()
        }
    }
}
```
ReferenceSearch seems like a better idea, but for some reason it doesn't work here:
```rust
usages.addAll(ReferencesSearch.search(function, projectScope).findAll())

usagesAsReference = usages.filter {
    it.element.ancestorOrSelf<RsCallExpr>() == null
        && it.element.ancestorOrSelf<RsMethodCall>() == null
}

usages.removeAll(usagesAsReference)
```
the `useSpeck` is the parent of one of the usages (debugged to check), but it doesn't get deleted, no idea why.

Kobzol:
`usagesAsReference` is filtering only method/function calls, it needs to also include the `useSpeck` usages. Also there could be some postprocessing, for example `use a::{foo};` would be turned into `use a::{}`, but deleting it entirely is probably even nicer. Let's continue this discussion on the new PR.

SaarYogev:
Opened #6003 #6004 #6005 for review, we'll continue dicussing there. I'll also add what's missing in each PR later today.

Issue #5888, opened 2020-08-07
Compiler lints support


Undin:
Meta issue about builtin compiler inspections ([lint](https://doc.rust-lang.org/rustc/lints/index.html)) support

From development point of view, each lint is a separate inspection and `RsLintInspection` as parent class.
`RsLintInspection` already implements the following:
* Annotations are automatically suppressed if there is  `allow` attribute with the corresponding lint id/lint [group](https://doc.rust-lang.org/rustc/lints/groups.html) id
* Suppression quick fixes to add `#[allow(lint_id)]` attribute
* Highlighting type adjustment depending on current lint [level](https://doc.rust-lang.org/rustc/lints/levels.html) in an editor, i.e. lint attributes change highlighting type of the corresponding annotations 

* [ ] Create separate inspection group in inspection settings

The whole list of compiler lints can be obtained via `rustc -W help` command
| name | groups | compiler severity | description | note |
| - | - | - | - | - |
| <ul><li>[ ] absolute-paths-not-starting-with-crate</li></ul> | future-incompatible, rust-2018-compatibility | allow | fully qualified paths that start with a module name instead of `crate`, `self`, or an extern crate name |  |
| <ul><li>[ ] anonymous-parameters</li></ul> | future-incompatible, rust-2018-compatibility | allow | detects anonymous parameters |  |
| <ul><li>[ ] box-pointers</li></ul> |  | allow | use of owned (Box type) heap memory |  |
| <ul><li>[ ] confusable-idents</li></ul> |  | allow | detects visually confusable pairs between identifiers |  |
| <ul><li>[ ] deprecated-in-future</li></ul> |  | allow | detects use of items that will be deprecated in a future version |  |
| <ul><li>[ ] elided-lifetimes-in-paths</li></ul> | rust-2018-idioms | allow | hidden lifetime parameters in types are deprecated |  |
| <ul><li>[ ] explicit-outlives-requirements</li></ul> | rust-2018-idioms | allow | outlives requirements can be inferred |  |
| <ul><li>[ ] indirect-structural-match</li></ul> | future-incompatible | allow | pattern with const indirectly referencing non-structural-match type |  |
| <ul><li>[ ] keyword-idents</li></ul> | future-incompatible, rust-2018-compatibility | allow | detects edition keywords being used as an identifier |  |
[...]

PR #5890, opened 2020-08-07
INSP: convert while true inspection into lint one

Undin:
After this change, the plugin will be able to provide all features described in #5888 for "while true" inspection, i.e. automatic suppression by `#[allow(while_true)]` and suppression quick fixes

Depends on #5889

Part of #5888


Undin:
bors r=mchernyavsky

PR #5906, opened 2020-08-10
Rename `RsDocAndAttributeOwner.isEnabledByCfg`

dima74:
There are two methods with name `isEnabledByCfg`:
* [`PsiElement.isEnabledByCfg`](https://github.com/intellij-rust/intellij-rust/blob/7dec3d603b1569a2285c2e5e36b1965e100db800/src/main/kotlin/org/rust/ide/utils/CfgUtils.kt#L14)
* [`RsDocAndAttributeOwner.isEnabledByCfg`](https://github.com/intellij-rust/intellij-rust/blob/b36c3a591cce1182674b037e64e47ede2ae021e3/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt#L208)

If I understand correctly, first method checks cfg-attributes of item and all its parent modules, and thus is preffered to use. And second method only checks cfg-attributes of item, thus should be used in specific situations (expanded items, etc). I think it would be better if these methods would have different names to prevent accidental misuse


ortem:
bors retry

PR #5910, opened 2020-08-10
PRJ: Create default run configuration after project generation

avrong:
Makes default run configurations for projects when generating from built-in templates.

Uses these commands:
* `cargo run` for binary
* `cargo test` for library
* `wasm-pack build` for WebAssembly

### DIFF
--- a/src/main/kotlin/org/rust/ide/newProject/RsProjectTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/newProject/RsProjectTemplate.kt
@@ -8,18 +8,20 @@ package org.rust.ide.newProject
 import org.rust.ide.icons.RsIcons
 import javax.swing.Icon
 
-sealed class RsProjectTemplate(val name: String, val isBinary: Boolean) {
-    abstract val icon: Icon
-
+sealed class RsProjectTemplate(val name: String, val isBinary: Boolean, val icon: Icon) {
     fun validateProjectName(crateName: String): String? = RsPackageNameValidator.validate(crateName, isBinary)
 }
 
-class RsGenericTemplate(name: String, isBinary: Boolean) : RsProjectTemplate(name, isBinary) {
-    override val icon: Icon = RsIcons.RUST
+sealed class RsGenericTemplate(name: String, isBinary: Boolean) : RsProjectTemplate(name, isBinary, RsIcons.RUST) {
+    object CargoBinaryTemplate : RsGenericTemplate("Binary (application)", true)
+    object CargoLibraryTemplate : RsGenericTemplate("Library", false)
 }
 
-class RsCustomTemplate(name: String, val link: String, isBinary: Boolean = true) : RsProjectTemplate(name, isBinary) {
-    override val icon: Icon = RsIcons.CARGO_GENERATE
+open class RsCustomTemplate(
+    name: String, val url: String
+) : RsProjectTemplate(name, false, RsIcons.CARGO_GENERATE) {
+    object WasmPackTemplate : RsCustomTemplate("WebAssembly Lib", "https://github.com/rustwasm/wasm-pack-template")
+
     val shortLink: String
-        get() = link.substringAfter("//")
+        get() = url.substringAfter("//")
 }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/newProject/RsProjectTemplate.kt
```diff
@@ -14,12 +14,18 @@ sealed class RsProjectTemplate(val name: String, val isBinary: Boolean) {
     fun validateProjectName(crateName: String): String? = RsPackageNameValidator.validate(crateName, isBinary)
 }
 
-class RsGenericTemplate(name: String, isBinary: Boolean) : RsProjectTemplate(name, isBinary) {
+sealed class RsGenericTemplate(name: String, isBinary: Boolean) : RsProjectTemplate(name, isBinary) {
```
ortem, line 17:
```Kotlin
sealed class RsGenericTemplate(name: String, isBinary: Boolean) : RsProjectTemplate(name, isBinary, RsIcons.RUST) {
    object CargoBinaryTemplate : RsGenericTemplate("Binary (application)", true)
    object CargoLibraryTemplate : RsGenericTemplate("Library", false)
}
```

src/main/kotlin/org/rust/ide/newProject/RsProjectTemplate.kt
```diff
@@ -14,12 +14,18 @@ sealed class RsProjectTemplate(val name: String, val isBinary: Boolean) {
...
     override val icon: Icon = RsIcons.RUST
 }
 
-class RsCustomTemplate(name: String, val link: String, isBinary: Boolean = true) : RsProjectTemplate(name, isBinary) {
+sealed class RsCustomTemplate(name: String, val link: String, isBinary: Boolean = true) : RsProjectTemplate(name, isBinary) {
```
ortem, line 21: Let's move objects inside and get rid of `isBinary` parameter and assume that `RsCustomTemplate` is always **not binary** to avoid false-negative name validation errors

src/main/kotlin/org/rust/ide/newProject/Utils.kt
```diff
@@ -24,12 +41,37 @@ fun Project.openFiles(files: GeneratedFilesHolder) = invokeLater {
...
-    is RsCustomTemplate -> generate(project, module, baseDir, template.link)
+fun Project.makeDefaultRunConfiguration(template: RsProjectTemplate) {
+    val runManager = RunManager.getInstance(this)
+    val configurationFactory = DefaultRunConfigurationFactory(runManager, this)
+
+    val configuration = when (template) {
+        is CargoBinaryTemplate -> configurationFactory.createCargoRunConfiguration()
+        is CargoLibraryTemplate -> configurationFactory.createCargoTestConfiguration()
+        is WasmPackTemplate -> configurationFactory.createWasmPackBuildConfiguration()
+        else -> null
```
ortem, line 52: It is redundant, isn't it?

src/main/kotlin/org/rust/ide/newProject/Utils.kt
```diff
@@ -24,12 +41,37 @@ fun Project.openFiles(files: GeneratedFilesHolder) = invokeLater {
...
-    is RsCustomTemplate -> generate(project, module, baseDir, template.link)
+fun Project.makeDefaultRunConfiguration(template: RsProjectTemplate) {
+    val runManager = RunManager.getInstance(this)
+    val configurationFactory = DefaultRunConfigurationFactory(runManager, this)
+
+    val configuration = when (template) {
+        is RsGenericTemplate.CargoBinaryTemplate -> configurationFactory.createCargoRunConfiguration()
+        is RsGenericTemplate.CargoLibraryTemplate -> configurationFactory.createCargoTestConfiguration()
+        is RsCustomTemplate.WasmPackTemplate -> configurationFactory.createWasmPackBuildConfiguration()
+        else -> null
```
ortem, line 52:
```suggestion
        is RsCustomTemplate -> return
```

PR #5911, opened 2020-08-11
ANN: Check if the custom attr override is used by `findInScope`

Stzx:
- Check if the custom attr override is used by `findInScope`

- Add new error code `E0774`, which should be part of #886

changelog: Fix detection of  [E0774](https://doc.rust-lang.org/error-index.html#E0774) compiler error

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -1858,7 +1858,9 @@ private fun RsAttr.isBuiltinWithName(target: String): Boolean {
     if (name != target) return false
     if (name !in RS_BUILTIN_ATTRIBUTES) return false
 
-    return !hasInScope(name, MACROS)
+    val procMacro = findInScope(name, MACROS) as? RsFunction ?: return true
+
+    return !procMacro.isAttributeProcMacroDef
 }
 
 private val RsPat.isTopLevel: Boolean
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -1370,7 +1370,7 @@ sealed class RsDiagnostic(
     class DeriveAttrUnsupportedItem(element: RsAttr) : RsDiagnostic(element) {
         override fun prepare(): PreparedAnnotation = PreparedAnnotation(
             ERROR,
-            null,
+            E0774,
             "`derive` may only be applied to structs, enums and unions",
             fixes = listOf(RemoveAttrFix(element as RsAttr))
         )
@@ -1580,7 +1580,7 @@ enum class RsErrorCode {
     E0403, E0404, E0407, E0415, E0416, E0424, E0426, E0428, E0429, E0430, E0431, E0433, E0434, E0435, E0449, E0451, E0463,
     E0517, E0518, E0537, E0552, E0554, E0562, E0569, E0583, E0586, E0594,
     E0601, E0603, E0614, E0616, E0618, E0624, E0658, E0666, E0667, E0688, E0695,
-    E0703, E0704, E0728, E0732, E0733, E0741, E0742, E0747;
+    E0703, E0704, E0728, E0732, E0733, E0741, E0742, E0747, E0774;
 
     val code: String
         get() = toString()
--- a/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
@@ -4264,7 +4264,7 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
     """)
 
     fun `test use derive attr on unsupported items`() = checkErrors("""
-        <error descr="`derive` may only be applied to structs, enums and unions">#[derive(Debug)]</error>
+        <error descr="`derive` may only be applied to structs, enums and unions [E0774]">#[derive(Debug)]</error>
         type Test = i32;
     """)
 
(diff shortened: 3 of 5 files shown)


Stzx:
Change the original wrong test method name.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1220,11 +1220,13 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
...
+private fun RsAttr.isBuiltinAttrWithName(target: String): Boolean {
     val name = metaItem.name ?: return false
 
     if (name != target) return false
     if (name !in STD_ATTRIBUTES) return false
 
-    return !hasInScope(name, MACROS)
+    val fn = findInScope(name, MACROS) as? RsFunction
+
+    return !(fn != null && fn.isAttributeProcMacroDef)
```
Undin, line 1231: Consider to use something like
```suggestion
    val fn = findInScope(name, MACROS) as? RsFunction ?: return true

    return !fn.isAttributeProcMacroDef
```
or
```suggestion
    val fn = findInScope(name, MACROS) as? RsFunction

    return fn?.isAttributeProcMacroDef != true
```

src/main/kotlin/org/rust/lang/core/RsPsiPattern.kt
```diff
@@ -84,6 +85,9 @@ object RsPsiPattern {
 
         "non_exhaustive",
 
+        // inner attr
+        "feature",
```
Undin, line 89: I don't think it makes sense to add these comments because you can use, for example, lint attributes (`allow`, `warn`, etc) as inner and outer attributes

Stzx: Oh, this should be part of the realization of another question.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1220,11 +1220,13 @@ private fun checkTypesAreSized(holder: RsAnnotationHolder, fn: RsFunction) {
     }
 }
 
-private fun RsAttr.isBuiltinWithName(target: String): Boolean {
+private fun RsAttr.isBuiltinAttrWithName(target: String): Boolean {
```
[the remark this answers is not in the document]
Stzx: With this name I just want to distinguish `proc_macro_attribute` from other types.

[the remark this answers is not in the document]
Stzx: Maybe there are other better names.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1858,7 +1858,7 @@ private fun RsAttr.isBuiltinWithName(target: String): Boolean {
     if (name != target) return false
     if (name !in RS_BUILTIN_ATTRIBUTES) return false
 
-    return !hasInScope(name, MACROS)
+    return (findInScope(name, MACROS) as? RsFunction)?.isAttributeProcMacroDef?.not() ?: true
```
dima74, line 1861:
```suggestion
    val procMacro = findInScope(name, MACROS) as? RsFunction ?: return true
    return !procMacro.isAttributeProcMacroDef
```

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1858,7 +1858,7 @@ private fun RsAttr.isBuiltinWithName(target: String): Boolean {
     if (name != target) return false
     if (name !in RS_BUILTIN_ATTRIBUTES) return false
 
-    return !hasInScope(name, MACROS)
```
dima74, line 1861: Lets remove `hasInScope` function since it is now unused

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -4264,8 +4264,10 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
     """)
 
     fun `test use derive attr on unsupported items`() = checkErrors("""
-        <error descr="`derive` may only be applied to structs, enums and unions">#[derive(Debug)]</error>
-        type Test = i32;
+        trait Foo {
+            <error descr="`derive` may only be applied to `struct`s, `enum`s and `union`s [E0774]">#[derive(Debug)]</error>
+            type Bar;
+        }
     """)
```
dima74, line 4271: Lets keep the original test, and add new one with associated type if you want

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -4278,6 +4280,28 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
...
+        use proc_macro::TokenStream;
+
+        #[proc_macro_attribute]
+        pub fn derive(attr: TokenStream, item: TokenStream) -> TokenStream {
+            item
+        }
+    //- lib.rs
+        #[dep_proc_macro::derive]
+        fn main() {
+        }
```
dima74, line 4295:
```suggestion
        fn main() {}
```

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -4278,6 +4280,28 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
...
+
+        mod foo {
+            use dep_proc_macro::derive;
+
+            #[derive/*caret*/]
```
dima74, line 4300:
```suggestion
            #[derive]
```

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1370,8 +1370,8 @@ sealed class RsDiagnostic(
     class DeriveAttrUnsupportedItem(element: RsAttr) : RsDiagnostic(element) {
         override fun prepare(): PreparedAnnotation = PreparedAnnotation(
             ERROR,
-            null,
-            "`derive` may only be applied to structs, enums and unions",
+            E0774,
+            "`derive` may only be applied to `struct`s, `enum`s and `union`s",
```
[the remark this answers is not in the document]
Stzx: Error messages from rust, 😆

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
Stzx: `checkByFileTree` requires caret marker, or did I put it in the wrong place?

dima74: Oh, indeed. Lets put it in the beginning of the file
```
    //- lib.rs
        /*caret*/
        #[dep_proc_macro::derive]
        fn main() {}
        ...
```

dima74:
Please also fix test `RemoveElementFixTest#test derive on function`

dima74 (approved):
bors r+

Issue #5909, opened 2020-08-10
Support "in place rename"


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.129.3272-202-nightly
* **Rust toolchain version:** 1.47.0-nightly (6c8927b0c 2020-07-26) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.2 (CL-202.6397.106)
* **Operating system:** macOS 10.15.4

## Problem description

PyCharm 2020.2 [supports](https://youtrack.jetbrains.com/issue/PY-42285) new platform feature of "in-place rename". Would love to see it in Rust plugin as I either too lazy or too forgetful to invoke rename beforehand in 9 times of 10.

How it looks like

![2020-08-10 20 23 54](https://user-images.githubusercontent.com/2500320/89812986-a8d75000-db49-11ea-8a43-84ac091d00e3.gif)

PR #5913, opened 2020-08-11
REF: add suggested rename refactoring

Kobzol:
This PR adds support for the new suggested refactoring. Since change signature refactoring is not implemented for Rust yet, I only added basic support for renaming local bindings.

rename

I copy-pasted the test infrastructure from the Kotlin plugin, because I couldn't get `com.intellij.refactoring.suggested.BaseSuggestedRefactoringTest` to work, as it couldn't autoload some Java `PsiElementFactory` class during runtime. However, even after copying the tests, they do not work. When I perform a change in the test file text and commit it, the refactoring suggestion intention is not offered. Yet when I do the exact same change in the IDE, it works fine. Any hints on what I could be missing?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5909

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringSupport.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringSupport.kt
@@ -1,0 +1,43 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.refactoring.suggested
+
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.refactoring.suggested.*
+import org.rust.lang.core.psi.RsPatBinding
+import org.rust.lang.core.psi.RsPatIdent
+import org.rust.lang.core.psi.ext.RsNameIdentifierOwner
+import org.rust.lang.core.psi.ext.isReferenceToConstant
+
+class RsSuggestedRefactoringSupport : SuggestedRefactoringSupport {
+    override val availability: SuggestedRefactoringAvailability
+        get() = SuggestedRefactoringAvailability.RenameOnly(this)
+    override val execution: SuggestedRefactoringExecution
+        get() = SuggestedRefactoringExecution.RenameOnly(this)
+    override val stateChanges: SuggestedRefactoringStateChanges
+        get() = SuggestedRefactoringStateChanges.RenameOnly(this)
+    override val ui: SuggestedRefactoringUI
+        get() = SuggestedRefactoringUI.RenameOnly
+
+    override fun importsRange(psiFile: PsiFile): TextRange? = null
+
+    override fun isDeclaration(psiElement: PsiElement): Boolean = when (psiElement) {
+        is RsPatBinding -> psiElement.parent is RsPatIdent && !psiElement.isReferenceToConstant
+        is RsNameIdentifierOwner -> true
+        else -> false
+    }
+
+    override fun isIdentifierPart(c: Char): Boolean = Character.isUnicodeIdentifierStart(c)
+    override fun isIdentifierStart(c: Char): Boolean = Character.isUnicodeIdentifierPart(c)
+
+    override fun nameRange(declaration: PsiElement): TextRange? = getRange(declaration)
+    override fun signatureRange(declaration: PsiElement): TextRange? = getRange(declaration)
+
+    private fun getRange(declaration: PsiElement): TextRange? =
+        (declaration as? RsNameIdentifierOwner)?.nameIdentifier?.textRange
+}
--- a/src/test/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringTest.kt
@@ -1,0 +1,103 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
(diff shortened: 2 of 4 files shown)


src/main/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringSupport.kt
```diff
@@ -0,0 +1,36 @@
...
+    override val ui: SuggestedRefactoringUI
+        get() = SuggestedRefactoringUI.RenameOnly
+
+    override fun importsRange(psiFile: PsiFile): TextRange? = null
+
+    override fun isDeclaration(psiElement: PsiElement): Boolean =
+        psiElement is RsPatBinding && psiElement.parent is RsPatIdent
+
+    override fun isIdentifierPart(c: Char): Boolean = isValidRustVariableIdentifier(c.toString())
+    override fun isIdentifierStart(c: Char): Boolean = isValidRustVariableIdentifier(c.toString())
```
Kobzol, line 32: This seems a bit weird, I didn't understand how exactly is an identifier defined in the lexer - does it include numeric literals? But this implementation seems to work, i.e. the suggestion is not offered for `5`, but is offered for `a5` or `a`, which seems correct.

Kobzol:
Hmm, for some reason `myFixture.testHighlighting(false, false, false, myFixture.file.virtualFile)` has to be run in the test for the suggested refactoring to be offered :man_shrugging:

Kobzol:
@Undin any comments? :)

src/main/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringSupport.kt
vlad20012: If I understand Java's std API right (I would not rely on myself!), `Character.isUnicodeIdentifierStart()`/`Character.isUnicodeIdentifierPart()` should works here. If it actually works, let's use it because it just faster then the lexer.

The fun fact is that both options don't work with raw identifiers (because the API sucks). But I think we can live without raw identifiers support in such case

src/main/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringSupport.kt
```diff
@@ -0,0 +1,43 @@
...
+        get() = SuggestedRefactoringExecution.RenameOnly(this)
+    override val stateChanges: SuggestedRefactoringStateChanges
+        get() = SuggestedRefactoringStateChanges.RenameOnly(this)
+    override val ui: SuggestedRefactoringUI
+        get() = SuggestedRefactoringUI.RenameOnly
+
+    override fun importsRange(psiFile: PsiFile): TextRange? = null
+
+    override fun isDeclaration(psiElement: PsiElement): Boolean = when (psiElement) {
+        is RsPatBinding -> psiElement.parent is RsPatIdent
```
vlad20012, line 30:
```suggestion
        is RsPatBinding -> psiElement.parent is RsPatIdent && !psiElement.isReferenceToConstant
```

src/test/kotlin/org/rust/ide/refactoring/suggested/RsSuggestedRefactoringTest.kt
```diff
@@ -0,0 +1,103 @@
...
+
+import com.intellij.codeInsight.intention.IntentionAction
+import com.intellij.openapi.application.runWriteAction
+import com.intellij.openapi.command.executeCommand
+import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.impl.source.PostprocessReformattingAspect
+import com.intellij.refactoring.RefactoringBundle
+import org.rust.RsTestBase
+
+abstract class RsSuggestedRefactoringTest : RsTestBase() {
```
vlad20012, line 16:
```suggestion
abstract class RsSuggestedRefactoringTestBase : RsTestBase() {
```

vlad20012:
:+1: 
bors r+

Kobzol:
bors retry

Kobzol:
bors retry

vlad20012:
Ok, this looks like a real failure. Something was broken it the test infrastructure that leads to test failures outside of RsSuggestedRefactoringTestBase

Kobzol:
The failed test works locally after rebasing on master. Let's try one more time. If it's still broken, something will probably have to be fixed in master.

bors retry

vlad20012:
The failed workflow was a test against the new name resolution engine, so you can reproduce it only with corresponding environment variables. But I didn't check that. Also, we experience some problems with Github Actions right now, so most of bulds are failing by timeout

Kobzol:
I tried it with the new resolve engine :shrug: But who knows what happens on CI right now, it's weird.

Kobzol:
@vlad20012 I'm a bit confused now, do you think that it's spurious issue or should it be fixed from my side? It works locally (even with the new resolution engine).

vlad20012:
I think I should investigate it by myself

dima74:
I can't reproduce error locally and [on CI](https://github.com/dima74/intellij-rust/runs/1621480234?check_suite_focus=true), so let's try again

bors r=vlad20012

lancelote:
Out of interest is it expected to work when I rename the usage instead of the declaration? E.g. if I rename second `foo` in the following example rename refactoring is not suggested

```rust
let foo = 0;
let b = foo;
```

Kobzol:
It's not expected by my implementation, and I also don't think that it's expected by the whole Suggested refactoring machinery (at least I didn't find any mention of this inside the API). When you change an usage it's not possible in general to tell if you want to change the referenced element or if you just want to use a different element.

lancelote:
Makes sense, thank you for the clarification.

PR #5918, opened 2020-08-12
RES: Skip empty attributes during cfg evaluation

ortem:


ortem:
Discussed in person that it has a negative impact on performance so should be closed.

Issue #1940, opened 2017-10-27
Highlight & inspect inline asm


adiba:
OMG how low priority.
But that would make it safer ;)
[Inline Assembly Docs](https://doc.rust-lang.org/1.1.0/book/inline-assembly.html)

vlad20012:
According to the [new RFC](https://github.com/rust-lang/rfcs/pull/2873), it now looks more reasonable to highlight  instruction arguments like we do in `println!` patterns #2256

Kobzol:
The new asm! syntax is relatively simple, so it could be parsed with a regex, but I suppose that writing a grammar for it is a more robust solution?

vlad20012:
I don't think it's more reasonable then write a grammar for format patterns.

PR #5928, opened 2020-08-13
GRAM: add initial parsing of the new asm! macro

Kobzol:
This PR adds initial support of parsing the new unstable `asm!` macro (https://github.com/Amanieu/rfcs/blob/inline-asm/text/0000-inline-asm.md).

The macro itself has a relatively complicated grammar (https://github.com/Amanieu/rfcs/blob/inline-asm/text/0000-inline-asm.md#reference-level-explanation). It includes nested expressions (`<expr> => <expr>`, `const <expr>`), paths (`sym <path>`) etc. Therefore I thought that it would be best if the parser could do most of the heavy lifting.
@vlad20012 I know that you mentioned [here](https://github.com/intellij-rust/intellij-rust/issues/1940#issuecomment-654161427) that grammar for this is probably not necessary, but to me it seems that `asm!` is much more complex to parse than the formatting macros.

I left some things as identifiers (for example the individual option values), so that they could be checked in the annotator and not by the parser, but I had trouble applying the same approach to the register direction, because one of the possible values (`in`) is a keyword, so it couldn't be parsed as an identifier. I resolved it by enumerating all of the possible register directions directly in the parser. Is there any existing lexer/parser rule that says "anything that looks like an identifier, but it could also be a keyword"?

There is a slight edge case - if someone defines his own `asm!` macro (which is very unlikely IMO), the plugin will try to parse it with this complicated grammar. I'm not sure what's the alternative - parse things like `<expr> => <expr>` as opaque macro tokens (`MacroArgumentTT`?) and then try to extract the expressions in annotator/inference code?

I tried to use the name `AsmMacroRegExpr` instead of `AsmMacroRegTarget`, but for some reason if the rule name ended with `Expr`, the parsing failed.

Related: https://github.com/intellij-rust/intellij-rust/issues/1940


vlad20012:
> @vlad20012 I know that you mentioned here that grammar for this is probably not necessary, but to me it seems that `asm!` is much more complex to parse than the formatting macros.

I don't think my advice was too deeply thought out =D
If I remember it correctly, I mostly thought about asm instructions that are in string literals inside `asm!` macro. So, yes, this is a reasonable design, I apologize for misleading.

> Is there any existing lexer/parser rule that says "anything that looks like an identifier, but it could also be a keyword"?

Well, `rule ::= IN | identifier` ? If you want to accept any keyword, `rule ::= <<macroIdentifier>>` is what you need.

> There is a slight edge case - if someone defines his own asm! macro

We have the same issue with other special macros like `vec` :shrug:. I don't know a good way to solve it for now, maybe we'll think of something in the future. 

> but for some reason if the rule name ended with Expr, the parsing failed.

Oh, well, welcome to Grammar Kit =D
We have `extends("(.+Expr|StructLiteral)") = Expr` rule that is applied to all rules whose names match the regexp (`Ctrl + B` works on regexps... or on rules - you can check that the rule is matched over some regexp).

vlad20012 (commented):
I didn't check the grammar against the RFC, I think the tests should be better in this. I just looked at Grammar Kit & stuff around - it looks good. Just 2 things: `RustParserDefinition.PARSER_VERSION` should be bumped, and `GRAM:` prefix instead of `BNF: `

Kobzol:
I suppose that later when annotations, highlighting and other things are implemented, the grammar might change a bit to make it more flexible to work with, but as a first version on which we can build it should be fine.

src/main/grammars/RustParser.bnf
```diff
@@ -1446,6 +1446,24 @@ AssertMacroArgument ::= <<any_braces (AnyExpr [ ',' <<comma_separated_list Forma
...
 //noinspection BnfUnusedRule
+AsmMacroArgument ::= <<any_braces [ <<comma_separated_list AsmMacroArg>> ] >>
+//noinspection BnfUnusedRule
+AsmMacroArg ::= AsmMacroOperand | STRING_LITERAL | AsmMacroOptions
+//noinspection BnfUnusedRule
+AsmMacroOperand ::= [ identifier '=' ] ( CONST AnyExpr | 'sym' PathWithoutTypeArgs | AsmMacroRegOperand )
+//noinspection BnfUnusedRule
+AsmMacroRegOperand ::= AsmMacroRegDir '(' AsmMacroRegSpec ')' AsmMacroRegTarget [ FAT_ARROW AsmMacroRegTarget ]
+//noinspection BnfUnusedRule
+AsmMacroRegDir ::= MetaVarIdentifier
```
vlad20012, line 1457: Let's not use `MetaVarIdentifier` here. It brings unreasonable PSI tree nesting. You can use `<<macroIdentifier>>` directly

Kobzol: I tried it, but the rule then had no attributes. But I suppose it's not needed, I will just read `text` directly anyway.

vlad20012: > I tried it, but the rule then had no attributes

They can be added with extension methods if needed

vlad20012:
:+1: Thanks!

bors r+

Issue #5922, opened 2020-08-12
Use project name instead of root folder's name when generating project


avrong:
Seems like for a long time there's been an issue in IDEA when the project name is not the same as the folder's name for a new project. Currently, this results in generating a cargo project named with a folder's name.

It also affects the custom project generator (`Cargo.generate`), which fails to generate a project in this case (mistake of comparing `project.name` and `path.fileName.toString()` and expecting them to be the same).

<img width="588" src="https://user-images.githubusercontent.com/6342851/90065258-ab29dd80-dcf4-11ea-8da1-6f4f659173e9.png">

<img width="676" src="https://user-images.githubusercontent.com/6342851/90064395-62255980-dcf3-11ea-9fdf-d95d86c9bca8.png">

Additionally, there are places where I'm not quite sure what the right usage will be: in the **project file tree** and in the **cargo projects tree**. They both currently use the root folder's name instead of the cargo project name specified in the manifest.

<img width="1049" src="https://user-images.githubusercontent.com/6342851/90068019-a6672880-dcf8-11ea-8cd6-0d8e36392c85.png">

PR #5929, opened 2020-08-13
PRJ: Use project name instead of root dir in IDEA when generating project

avrong:
Related to #5922

Fixes the first part of the issue. Now cargo package name for a new project will be actually the project name and not the root dir.

Also, fixes the problem of generating projects with custom templates when the project name and the root dir are different.

### DIFF
--- a/idea/src/main/kotlin/org/rust/ide/idea/RsModuleBuilder.kt
+++ b/idea/src/main/kotlin/org/rust/ide/idea/RsModuleBuilder.kt
@@ -52,11 +52,14 @@ class RsModuleBuilder : ModuleBuilder() {
                 val template = configurationData?.template ?: return
                 val cargo = toolchain.rawCargo()
                 val project = modifiableRootModel.project
+                val name = project.name.replace(' ', '_')
 
                 val generatedFiles = cargo.makeProject(
                     project,
                     modifiableRootModel.module,
-                    root, template
+                    root,
+                    name,
+                    template
                 ) ?: return
 
                 project.makeDefaultRunConfiguration(template)
--- a/src/main/kotlin/org/rust/ide/newProject/Utils.kt
+++ b/src/main/kotlin/org/rust/ide/newProject/Utils.kt
@@ -25,10 +25,11 @@ fun Cargo.makeProject(
     project: Project,
     module: Module,
     baseDir: VirtualFile,
+    name: String,
     template: RsProjectTemplate
 ): GeneratedFilesHolder? = when (template) {
-    is RsGenericTemplate -> init(project, module, baseDir, template.isBinary)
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/ide/newProject/Utils.kt
```diff
@@ -24,12 +24,16 @@ fun Project.openFiles(files: GeneratedFilesHolder) = invokeLater {
     }
 }
 
+fun Project.cargoProjectName() =
```
ortem, line 27: Let's inline this method because it is simple but the name is not obvious

idea/src/main/kotlin/org/rust/ide/idea/RsModuleBuilder.kt
```diff
@@ -51,11 +52,12 @@ class RsModuleBuilder : ModuleBuilder() {
                 val template = configurationData?.template ?: return
                 val cargo = toolchain.rawCargo()
                 val project = modifiableRootModel.project
+                val name = project.cargoProjectName()
 
                 val generatedFiles = cargo.makeProject(
                     project,
                     modifiableRootModel.module,
-                    root, template
+                    root, name, template
```
ortem, line 60: Let's put arguments on the separate lines

ortem (commented):
Please rebase onto master

src/main/kotlin/org/rust/ide/newProject/Utils.kt
```diff
@@ -25,10 +25,11 @@ fun Cargo.makeProject(
     project: Project,
     module: Module,
     baseDir: VirtualFile,
+    name: String,
     template: RsProjectTemplate
 ): GeneratedFilesHolder? = when (template) {
-    is RsGenericTemplate -> init(project, module, baseDir, template.isBinary)
-    is RsCustomTemplate -> generate(project, module, baseDir, template.url)
+    is RsGenericTemplate -> init(project, module, baseDir, template.isBinary, name)
+    is RsCustomTemplate -> generate(project, module, baseDir, name, template.url)
```
ortem, line 32: Let's pass the arguments in the same order (`name` after `template...`)

ortem (changes requested):

src/main/kotlin/org/rust/ide/newProject/Utils.kt
avrong: Maybe after `baseDir` then? To keep first 4 parameters the same in both functions. To make it look like this:
```
    init(project, module, baseDir, name, template.isBinary)
generate(project, module, baseDir, name, template.url)
```

ortem (approved):
bors r+

PR #5932, opened 2020-08-14
CI: initial regression testing

Undin:
Currently, each developer can launch `org.rustPerformanceTests.RsRealProjectAnalysisTest` to analyze if they introduce some new false-positive errors. It works somehow but has a lot of cons:
- a developer may forget to launch regression tests
- these tests should be launched locally that consumes CPU and a lot of time
- there no way to get diff between the previous set of errors and the current one. It's important since the plugin produces some false positives (unfortunately) and you have to compare previous results with new ones manually

This PR introduces new `regressions` GitHub workflow that should automate the launch of regression tests and help us to identify regressions earlier. This implementation includes:
- new `CustomRealProjectAnalysisTest`. It can be configured via set of environment variables to run the plugin analysis on custom rust project. To implement it I've made some refactoring in `RsRealProjectAnalysisTest`
- `regressions` workflow itself. It's triggered automatically in each PR in the repository (not sure that it's ok but let's try). Also, you can trigger it with desired branch manually via GitHub UI. What workflow does:
    * checkout all real rust projects specified in the workflow. Currently, the workflow contains only one project - Cargo. The list will be extended in future
    * launches `org.rustPerformanceTests.RsRealProjectAnalysisTest` twice for each real project: with and without changes of the corresponding branch/PR. All produced error annotations are stored into json format
    * calculate diff between two sets of changes to determine what errors were fixed and what were introduced. If there is at least one new error annotation, workflow task fails.
    * all results saved as an artifact of the workflow (artifact per real project)

Further possible improvements:
* extend list of real projects
* better error handling. Currently, our analysis may fail on some projects and produce an exception. Currently, it fails the corresponding test. We should collect such errors as well.
* support not only error annotations to have an ability to check regressions for any inspection/annotator
* reduce total time of execution. Currently, the workflow compiles plugin project for each real project twice. Probably, we can move out project compilation, save the result and use it for each real project

How to test (possible variant):
* push these changes to main branch of your fork of this repo
* create a new branch and make some changes that lead to new error. For example, you can apply this [patch](https://github.com/intellij-rust/intellij-rust/files/5075343/regression.patch.zip)
* push created branch and run `regressions` workflow for it (manually or via PR creation). It should fail and you can check the corresponding results in artifacts

### DIFF
--- a/.github/workflows/regressions.yml
+++ b/.github/workflows/regressions.yml
@@ -1,0 +1,101 @@
+name: regressions
+
+on: workflow_dispatch
+
+jobs:
+    calculate-base-commit:
+        runs-on: ubuntu-latest
+        outputs:
+            base-commit: ${{ steps.base-commit.outputs.base-commit }}
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
+
+            - name: Calculate base commit
+              id: base-commit
+              run: echo "::set-output name=base-commit::$(git merge-base origin/master ${{ github.sha }})"
+
+            - name: Show commits
+              run: |
+                  echo "current commit:"
+                  git log ${{ github.sha }} -n 1 --pretty=short
+                  echo "base commit:"
+                  git log ${{ steps.base-commit.outputs.base-commit }} -n 1 --pretty=short
+
+    check:
+        needs: [ calculate-base-commit ]
+        name: ${{ matrix.project.name }}
+        runs-on: ubuntu-latest
+        strategy:
+            fail-fast: false
+            matrix:
+                project:
+                    - name: cargo
+                      path: cargo
+                      repository: rust-lang/cargo
+                      exclude_paths: ""
+        env:
+            PROJECT_PATH: ${{ matrix.project.path }}
+            PROJECT_URL: https://github.com/${{ matrix.project.url }}
+            PROJECT_EXCLUDE_PATHS: ${{ matrix.project.exclude_paths }}
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
+
+            - uses: actions/checkout@v2
+              with:
+                  repository: ${{ matrix.project.repository }}
+                  ref: master
+                  path: testData/${{ matrix.project.name }}
+
+            - name: Set up JDK 1.8
+              uses: actions/setup-java@v1
+              with:
+                  java-version: 1.8
+
+            - name: Set up Python
+              uses: actions/setup-python@v1
+              with:
+                  python-version: 3.7
+
+            - name: Set up Rust
+              uses: actions-rs/toolchain@v1
+              with:
+                  profile: minimal
+                  toolchain: stable
+                  components: rust-src
+                  default: true
+
+            - name: Download
+              uses: eskatos/gradle-command-action@v1
+              with:
+                  arguments: ":resolveDependencies -Pkotlin.incremental=false"
+
+            - name: Check with changes
+              uses: eskatos/gradle-command-action@v1
+              env:
+                  PROJECT_NAME: ${{ matrix.project.name }}_with_changes
+              with:
+                  arguments: "clean :test --tests \"org.rustPerformanceTests.CustomRealProjectAnalysisTest.test\""
+
+            - name: Checkout base version
+              run: git checkout ${{ needs.calculate-base-commit.outputs.base-commit }}
+
+            - name: Check without changes
+              uses: eskatos/gradle-command-action@v1
+              env:
+                  PROJECT_NAME: ${{ matrix.project.name }}_without_changes
+              with:
+                  arguments: "clean :test --tests \"org.rustPerformanceTests.CustomRealProjectAnalysisTest.test\""
+
+            - name: Calculate regressions
+              run: python scripts/calculate_regressions.py --name ${{ matrix.project.name }}
+
+            - name: Upload results
+              if: ${{ always() }}
+              uses: actions/upload-artifact@v2
+              with:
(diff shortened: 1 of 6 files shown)


Kobzol:
Hehe, this is awesome! Closing https://github.com/intellij-rust/intellij-rust/pull/5604.

scripts/calculate_regressions.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __str__(self) -> str:
+        if self.inspectionToolId is not None:
+            suffix = f" by {self.inspectionToolId}"
+        else:
+            suffix = ""
+        return f"{self.filePath}:{self.line}:{self.column} '{self.highlightedText}' ({self.error}){suffix}"
+
+    @staticmethod
+    def from_dict(raw_dict):
```
lancelote, line 25: `def from_dict(raw_dict: Dict[str]) -> "Annotation":`?

scripts/calculate_regressions.py
```diff
@@ -0,0 +1,72 @@
...
+                          raw_dict["highlightedText"],
+                          raw_dict["error"],
+                          raw_dict["inspectionToolId"])
+
+
+def read_data(path: str) -> List[Annotation]:
+    with open(path) as json_file:
+        data = json.load(json_file)
+
+    return list(map(lambda value: Annotation.from_dict(value), data))
```
lancelote, line 38: List comprehension is such much readable in my opinion 😄

```python
[Annotation.from_dict(value) for value in data]
```

scripts/calculate_regressions.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def dump_as_json(annotations: Iterable[Annotation], path: str):
```
lancelote, line 41: Should have `-> None` by convention ¯\_(ツ)_/¯

scripts/calculate_regressions.py
```diff
@@ -0,0 +1,72 @@
...
+    json_array = list(map(lambda a: dataclasses.asdict(a), annotations))
```
lancelote, line 42: List comprehensions are life 💜

[dataclasses.asdict(a) for a in annotations]

scripts/calculate_regressions.py
Undin: Cool!
It would be great to have an intention in python plugin to make such conversion automatically

scripts/calculate_regressions.py
Undin: 🤦 
It's very consistent syntax 😃

vlad20012:
:+1: 
bors r+

Undin:
bors retry

Undin:
bors=vlad20012

Undin:
bors r=vlad20012

.github/workflows/regressions.yml
```diff
@@ -0,0 +1,116 @@
...
+
+            - name: Set up JDK 1.8
+              uses: actions/setup-java@v1
+              with:
+                  java-version: 1.8
+
+            - name: Set up Python
+              uses: actions/setup-python@v1
+              with:
+                  python-version: 3.7
```
matklad, line 76: I'd personally write this in Kotlin/Rust. Python is undoubtedly the right tool for the job here, but, imo, adding a new technology has high cost. Even if Kotlin is less prettier to run, I *think* it will be easier to maintain long-term.

matklad: But, obviously, no strong opinion here :D

.github/workflows/regressions.yml
```diff
@@ -0,0 +1,116 @@
...
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
+
+            - uses: actions/checkout@v2
+              with:
+                  repository: ${{ matrix.project.repository }}
+                  ref: master
+                  path: testData/${{ matrix.project.name }}
```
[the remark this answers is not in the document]
Undin: It's `1` by default according its [documentation](https://github.com/actions/checkout#usage)

.github/workflows/regressions.yml
Undin: We already used some other python scripts in other workflows so it's not new technology for project.
Unfortunately, there is an issue with jvm-based language here (even if we forget that gradle is insanely slow). If you try to run any kotlin script via gradle in this project, gradle downloads IDEA & CLion dependencies despite we don't run anything related to plugin itself because gradle intellij plugin still downloads IDE dependencies in configuration phase. The [corresponding](https://github.com/JetBrains/gradle-intellij-plugin/issues/123) bug is fixed only recently and is not released yet. So any gradle run leads to downloading about 2GB and it takes more than 2 minutes.
Of course, in this concrete case we need to download these dependencies anyway but it's quite weird to use different approaches in different CI tasks
In combination with slowness of gradle itself, necessity of compilation and small probability of complex python scripts (in my opinion), it makes run/debug workflow quite slow for me (even with fact that I don't know python enough).
Anyway, if we'll understand that maintaining become hard, we will rewrite these scripts

matklad: Hm, that's a bummer indeed... And, apparently, there's no easy way to write Kotlin scripts without gradle? It'd be really nice if one could just `curl` a minimal Kotlin runtime with popular libraries, just for bashesque scripting. Might be a fun project for hackathone.

In rust-analyzer, we write all our bash [in Rust](https://github.com/rust-analyzer/rust-analyzer/tree/master/xtask), and it works so much better than gradle in practice :-) 

I guess, 

>The right tool for the job is often the tool you are already using —adding new tools has a higher cost than many people appreciate. 

has an additional condition -- existing tool is Rust :D

Issue #3066, opened 2018-11-21
WebAssembly


NeverGivinUp:
Are you planning to support or already supporting compiling to WASM? Where would I find instructional documentation? I think there already is some support for compiling Kotlin to WASM, so that could be an starting point for the WASM side. WASM requires Javascript support, to really do stuff in the browser. So this might be a compile target suited specifically for WebStorm-integration. Also: support for Yew would be great.

---

## META TODO

Feature support list:
- [x] WebAssembly projects creation in `New project` dialog
  - [x] `cargo-generate` templates
  - [ ] Native Cargo templates https://github.com/rust-lang/rfcs/pull/2922
- [x] `wasm-pack` run configurations
  - [x] Commands completion
  - [x] Build tool window for `build` and `test` commands
- [x] `wasm-pack` commands in `Run Anything`
- [ ] `wasm-bindgen` tests #3007
- [ ] Rust <-> JS interop
  - [x] Rust->JS navigation
  - [ ] JS->Rust navigation
  - [ ] Cross-language refactorings support
- [ ] npm integration
  - [ ] Set up a web page using `npm init wasm-app` and `npm install`
  - [ ] Provide run configuration `npm run start`
- [ ] Debugging of WebAssembly projects
  - [ ] Plain Rust+Wasm web projects
  - [ ] Rust+Wasm web projects with wasm-bindgen (blocked by https://github.com/rustwasm/wasm-bindgen/issues/2389)
  - [ ] Debugging when running in standalone runtime (e.g. `wasmtime`)
- [ ] Auto refresh/rebuild on changes (`watch` and `serve` subcommands) https://github.com/rustwasm/rfcs/pull/10
- [ ] Documentation
  - [ ] `cargo-generate`
  - [ ] `wasm-pack`

vlad20012:
Hmm, what kind of support do you expect from IDE? "Compiling" is a responsibility of, well, the compiler =) You can find an appropriate instruction about compiling Rust to wasm in [wasm-pack](https://github.com/rustwasm/wasm-pack) or stdweb projects. If you know some specific wasm-related features you expect from our plugin, feel free to list them =)

Btw IntelliJ-Rust works in WebStorm, and IntelliJ IDE ultimate also supports JavaScript.

NeverGivinUp:
> "Compiling" is a responsibility of, well, the compiler =)

[...]

PR #5933, opened 2020-08-14
Introduce line marker for wasm_bindgen attributes in WebAssembly projects

avrong:
Expected workflow: navigating from Rust statement to JS one. Then having the ability to navigate to JS parts related to it.

### How it works

Related to #3066

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -182,7 +182,8 @@ project(":plugin") {
             project(":intellij-toml"),
             "IntelliLang",
             graziePlugin,
-            psiViewerPlugin
+            psiViewerPlugin,
+            "JavaScriptLanguage"
         )
         if (baseIDE == "idea") {
             plugins += listOf(
@@ -205,6 +206,7 @@ project(":plugin") {
         implementation(project(":intelliLang"))
         implementation(project(":duplicates"))
         implementation(project(":grazie"))
+        implementation(project(":js"))
     }
 
     tasks {
@@ -403,6 +405,18 @@ project(":coverage") {
 project(":grazie") {
     intellij {
         setPlugins(graziePlugin)
+    }
+    dependencies {
+        implementation(project(":"))
+        implementation(project(":common"))
+        testImplementation(project(":", "testOutput"))
+        testImplementation(project(":common", "testOutput"))
+    }
+}
+
+project(":js") {
+    intellij {
+        setPlugins("JavaScriptLanguage")
     }
     dependencies {
         implementation(project(":"))
--- a/js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
+++ b/js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
@@ -1,0 +1,103 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.js
+
+import com.intellij.codeInsight.daemon.DefaultGutterIconNavigationHandler
+import com.intellij.codeInsight.daemon.RelatedItemLineMarkerInfo
+import com.intellij.codeInsight.daemon.RelatedItemLineMarkerProvider
+import com.intellij.lang.javascript.psi.JSElement
+import com.intellij.lang.javascript.psi.ecma6.TypeScriptClass
+import com.intellij.lang.javascript.psi.ecma6.TypeScriptEnum
+import com.intellij.lang.javascript.psi.ecma6.TypeScriptFunction
+import com.intellij.navigation.GotoRelatedItem
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.editor.markup.GutterIconRenderer
+import com.intellij.openapi.vfs.LocalFileSystem
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiManager
+import com.intellij.psi.util.findDescendantOfType
+import icons.JavaScriptPsiIcons
+import org.rust.lang.core.psi.RsEnumItem
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsImplItem
+import org.rust.lang.core.psi.RsStructItem
+import org.rust.lang.core.psi.ext.RsOuterAttributeOwner
+import org.rust.lang.core.psi.ext.RsVisibilityOwner
+import org.rust.lang.core.psi.ext.containingCargoPackage
+import org.rust.lang.core.psi.ext.findOuterAttr
+import org.rust.lang.core.psi.impl.RsBaseTypeImpl
+import java.nio.file.InvalidPathException
+import java.nio.file.Paths
+
+@Suppress("NAME_SHADOWING")
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        val element = element as? RsOuterAttributeOwner ?: return null
+        if (element !is RsFunction && element !is RsEnumItem && element !is RsStructItem && element !is RsImplItem) {
+            return null
+        }
+
+        // Check the package has wasm-bindgen as dependency
+        val cargoPackage = element.containingCargoPackage ?: return null
+        if (cargoPackage.dependencies.none { it.pkg.name == "wasm-bindgen" }) {
+            return null
+        }
+
+        // Only public elements are exposed in bindings
+        if (element is RsVisibilityOwner && element !is RsImplItem && !element.isPublic) {
+            return null
+        }
+
+        val attr = element.findOuterAttr("wasm_bindgen") ?: return null
+
+        val basePath = cargoPackage.contentRoot?.path ?: return null
+        val expectedName = cargoPackage.normName
+
+        // Expecting to find related generated typescript declarations in pkg/ folder of current package
+        val expectedPath = try {
+            Paths.get(basePath, "pkg/$expectedName.d.ts").toString()
+        } catch (e: InvalidPathException) {
+            LOG.error(e)
+            return null
+        }
+        val tsFile = LocalFileSystem.getInstance().findFileByPath(expectedPath) ?: return null
+        val tsPsiFile = PsiManager.getInstance(element.project).findFile(tsFile) ?: return null
+
+        // Trying to find the correlating element with the same name
+        val destination = when (element) {
+            // If it is impl, then resolving to the related struct and using its name
+            is RsImplItem -> {
+                val reference = (element.typeReference as? RsBaseTypeImpl) ?: return null
+                val struct = (reference.path?.reference?.resolve() as? RsStructItem) ?: return null
+                findRelatedTsElement(struct, tsPsiFile)
+            }
+            else -> findRelatedTsElement(element, tsPsiFile)
+        } ?: return null
+
+        return RelatedItemLineMarkerInfo(
+            attr,
+            attr.textRange,
+            JavaScriptPsiIcons.FileTypes.TypeScriptFile,
+            { "Go to generated declaration" },
+            DefaultGutterIconNavigationHandler(listOf(destination), "Generated declarations"),
+            GutterIconRenderer.Alignment.RIGHT,
+            { listOf(GotoRelatedItem(destination)) }
+        )
+    }
+
+    private fun findRelatedTsElement(element: RsOuterAttributeOwner, tsPsiFile: PsiFile): JSElement? =
+        when (element) {
+            is RsFunction -> tsPsiFile.findDescendantOfType<TypeScriptFunction> { it.name == element.name }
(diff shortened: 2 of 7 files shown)


build.gradle.kts
```diff
@@ -407,6 +410,19 @@ project(":grazie") {
     }
 }
 
+project(":js") {
```
Undin, line 413: Don't forget to add the corresponding description in [`CONTRIBUTING.md`](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#project-structure)

build.gradle.kts
```diff
@@ -407,6 +410,19 @@ project(":grazie") {
...
+    intellij {
+        version = ideaVersion
```
Undin, line 415: Do we really need IDEA here? As I remember, JavaScript plugin is bundled into CLion as well

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.rust.lang.core.psi.ext.RsOuterAttributeOwner
+import org.rust.lang.core.psi.ext.RsVisibilityOwner
+import org.rust.lang.core.psi.ext.containingCargoPackage
+import org.rust.lang.core.psi.ext.findOuterAttr
+import org.rust.lang.core.psi.impl.RsBaseTypeImpl
+import java.nio.file.Paths
+
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        if (element.project.crateGraph.topSortedCrates.none { it.presentableName == "wasm-bindgen" }) return null
```
Undin, line 40: I think we shouldn't use `presentableName` in non UI code. Let's better use `normName` instead

build.gradle.kts
```diff
@@ -314,6 +316,7 @@ project(":idea") {
 project(":clion") {
     intellij {
         version = clionVersion
+        setPlugins("JavaScriptLanguage")
```
Undin, line 319: BTW why we need JS plugin here?

ortem: Oops, accidental change

avrong:
There is an option to show JS file icon like PyCharm does for Flask templates. But I'm not sure if it is convenient to use a file icon here.

We can use something similar to the C++ icon, then I think we should make it distinctive (but not standing out) from others as wasm_bindgen is a third-party thing to the language itself and its IDE features.

image

A similar feature for extern functions would be cool. I'm not aware of the details but will try to find a way to keep these features in one area for the future.

build.gradle.kts
ortem: Is such description OK?
```
* `:js` - interop with JavaScript language
```

build.gradle.kts
ortem: Thanks, `baseVersion` should be used instead, right?

Undin: Yep but you shouldn't specify it explicitly here since `baseVersion` is set as a default `version` [here](https://github.com/intellij-rust/intellij-rust/blob/953cd3d052383e4eb10ab048a10ad992cd91fc08/build.gradle.kts#L75)

build.gradle.kts
Undin: 👍

Undin:
@avrong personally, I like an approach with file icon since it doesn't require any additional icons, quite extensible and it's already used in other JB products. But we should use TS icon instead of JS in this PR, shouldn't we?
@ortem what do you think?

ortem:
@Undin I think JS/TS icons is a clear and simple solution 👍

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.rust.lang.core.psi.ext.RsVisibilityOwner
+import org.rust.lang.core.psi.ext.containingCargoPackage
+import org.rust.lang.core.psi.ext.findOuterAttr
+import org.rust.lang.core.psi.impl.RsBaseTypeImpl
+import java.nio.file.Paths
+
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        // Line marker will be available only if there is wasm-bindgen dependency in the project,
+        // current element is public and can have outer attribute
```
ortem, line 41: AFAIU currently we only support `RsFunction`, `RsEnumItem`, `RsStructItem`, and `RsImplItem` as `element`. So maybe it's better to restrict the scope here?
I mean check that element is `RsFunction` or ... or `RsImplItem` instead of checking `element !is RsOuterAttributeOwner`, `element is RsVisibilityOwner`, and so forth. It might be more explicit and easy to understand.

ortem, line 41: Also, why we don't provide a line marker in case of private struct/enum?

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        // BACKCOMPAT: 2020.1
+        @Suppress("DEPRECATION")
+        return RelatedItemLineMarkerInfo(
+            attr,
+            attr.textRange,
+            JavaScriptPsiIcons.FileTypes.TypeScriptFile,
+            Pass.LINE_MARKERS,
+            ConstantFunction<PsiElement, String>("Related generated declaration"),
+            DefaultGutterIconNavigationHandler(listOf(destination), "Related generated declaration"),
```
ortem, line 78:
```suggestion
            DefaultGutterIconNavigationHandler(listOf(destination), "Generated declarations"),
```
AFAIU this title will be used in case of multiple targets (declarations). Currently, there is always only one target due to `findDescendantOfType`, but maybe there will be several ones later

ortem (changes requested):

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
avrong: AFAIK the function won't be exported to JS if it is private

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.rust.lang.core.psi.ext.findOuterAttr
+import org.rust.lang.core.psi.impl.RsBaseTypeImpl
+import java.nio.file.InvalidPathException
+import java.nio.file.Paths
+
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        // Line marker will be available only if there is wasm-bindgen dependency in the project,
+        // current element is public and can have outer attribute
+        if (element.project.crateGraph.topSortedCrates.none { it.normName == "wasm_bindgen" }) return null
```
ortem, line 41: Let's use `element.containingCargoPackage.dependencies` (package deps) instead of `element.project.crateGraph.topSortedCrates` (all crates of all cargo projects in `project`)

ortem (changes requested):

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+        } ?: return null
+
+        // BACKCOMPAT: 2020.1
+        @Suppress("DEPRECATION")
+        return RelatedItemLineMarkerInfo(
+            attr,
+            attr.textRange,
+            JavaScriptPsiIcons.FileTypes.TypeScriptFile,
+            Pass.LINE_MARKERS,
+            ConstantFunction<PsiElement, String>("Related generated declaration"),
```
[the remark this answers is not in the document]
ortem: Looks like there is still `Related generated declaration` instead of `Go to generated declaration`. I think `Go to` is better because it describes what happens when you click the button

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
ortem: OK

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+import java.nio.file.InvalidPathException
+import java.nio.file.Paths
+
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        // Line marker will be available only if there is wasm-bindgen dependency in the project,
+        // current element is supported and is public
+        if (element !is RsFunction && element !is RsEnumItem && element !is RsStructItem && element !is RsImplItem) return null
+        if (element is RsElement && element.containingCargoPackage?.dependencies?.none { it.name == "wasm-bindgen" } == true) return null
+        if (element !is RsOuterAttributeOwner && element is RsVisibilityOwner && element !is RsImplItem && !element.isPublic) return null
```
ortem, line 39: Is `element !is RsOuterAttributeOwner` condition reasonable? On this step, we're sure that `element` is a function/enum/struct/impl, and all of them are basically `RsOuterAttributeOwner`

ortem: I'd recommend to rewrite these checks because now they look a bit confusing
(discussed in person)

ortem (changes requested):

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,102 @@
...
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        val element = element as? RsOuterAttributeOwner ?: return null
+        if (element !is RsFunction && element !is RsEnumItem && element !is RsStructItem && element !is RsImplItem) {
+            return null
+        }
+
+        // Check the package has wasm-bindgen as dependency
+        val cargoPackage = element.containingCargoPackage ?: return null
+        if (cargoPackage.dependencies.none { it.name == "wasm-bindgen" }) {
```
ortem, line 45: Unfortunately, this does not work because `dependency.name` refers to the name inside `Cargo.toml`, potentially aliased. Consider such a case:
```
[dependencies]
my_wasm_bindgen_alias = { package = "wasm-bindgen", version = "0.2.63" }
```
In this case, `dependency.name` is `"my_wasm_bindgen_alias"`, but `dependency.pkg.name` is `"wasm-bindgen"`. So looks like you should use `pkg.name` here

ortem (changes requested):

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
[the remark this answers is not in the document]
ortem: @avrong Did you forget to update this code?

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.rust.lang.core.psi.ext.RsVisibilityOwner
+import org.rust.lang.core.psi.ext.containingCargoPackage
+import org.rust.lang.core.psi.ext.findOuterAttr
+import org.rust.lang.core.psi.impl.RsBaseTypeImpl
+import java.nio.file.InvalidPathException
+import java.nio.file.Paths
+
+class RsWasmBindgenLineMarkerProvider : RelatedItemLineMarkerProvider() {
+    override fun getLineMarkerInfo(element: PsiElement): RelatedItemLineMarkerInfo<*>? {
+        val element = element as? RsOuterAttributeOwner ?: return null
```
ortem, line 38: Let's add `@Suppress("NAME_SHADOWING")`

js/src/main/kotlin/org/rust/js/RsWasmBindgenLineMarkerProvider.kt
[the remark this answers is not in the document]
avrong: Oh, sorry, I think I missed your last comment 🤦‍♂️

ortem (approved):

Undin (commented):
What about tests?

ortem:
@Undin We are going to add them later, in a separate PR

Issue #5934, opened 2020-08-14
Attach to lib.rs does not notice attributes in lib.rs.


smj389:
## Environment

* **IntelliJ Rust plugin version:** 0.3.128.3278-202
* **Rust toolchain version:** 1.47.0-nightly (81dc88f88 2020-08-13) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2020.2 (CL-202.6397.106)
* **Operating system:** Windows 10 10.0

## Steps to reproduce

1. Imagine a lib.rs file like this:

```rust
#![allow(dead_code)]
#![feature(async_closure)]
mod one;
mod two;

```
2. Create a new file in the crate, e.g. three.rs
3. Plugin suggests "attach to lib.rs"
4. Get following result:
```rust
mod three;

#![allow(dead_code)]
#![feature(async_closure)]
mod one;
mod two;

```

5. Problem is, of course, that the attributes need to be at the top of the file.
6. 😢

Undin:
@Kobzol I think it can be fixed together with #5903

Note, we already have similar logic in auto-import quick fix. Probably, we can reuse it

Undin:
@smj389 Thanks for the report!

Issue #5903, opened 2020-08-10
Add mod declaration after the last availaible in the file when performing "Attach file to main.rs" action


avrong:
## Environment

* **IntelliJ Rust plugin version:** 0.3.127.3249-202
* **Rust toolchain version:** 1.45.2 (d3fb005a3 2020-07-31) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.2 (CL-202.6397.106)
* **Operating system:** macOS 10.15.6

## Problem description

I believe that the better option will be to add the new `mod` declaration right after the last one available if it is present. It will reduce the following manual editing.

<img width="772" src="https://user-images.githubusercontent.com/6342851/89767968-9c300900-db03-11ea-85da-db26138cb3ba.png">

<img width="772" src="https://user-images.githubusercontent.com/6342851/89768843-d6e67100-db04-11ea-9d9c-d2b09efc0d39.png">

`Optimize Imports` action doesn't solve the problem completely.

<img width="772" src="https://user-images.githubusercontent.com/6342851/89768856-dc43bb80-db04-11ea-9d8f-86c86c6d0e2b.png">

## Expected result

```rust
use std::path::Path;

use module::whatever;

mod module;
mod new_module;

fn main() {...}
```

Kobzol:
I wanted to do this in the original PR, but didn't get to it. I think that it's a great idea, thanks! :)

PR #5937, opened 2020-08-15
INSP: select a better location for attaching file to modules

Kobzol:
This PR adds a simple heuristic to `AttachFileToModuleFix` that tries to find a good place where to insert the `mod` item. If it finds an existing mod items in the file, it will append the new item after the last one. Otherwise, it will skip comments and attributes at the beginning of the module.

If there was a mod item before an attribute in the target module, it could still fail to compile after the fix, but in such case it would also fail to compile before the fix, so I think that is an edge case that we don't have to deal with.

I copied the `firstItem` implementation and changed it slightly so that it also works for empty modules. I can also check if the module is empty and do not use `firstItem` in such case.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5903
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5934

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
@@ -27,8 +27,7 @@ import org.rust.lang.RsConstants
 import org.rust.lang.core.psi.RsFile
 import org.rust.lang.core.psi.RsModDeclItem
 import org.rust.lang.core.psi.RsPsiFactory
-import org.rust.lang.core.psi.ext.RsMod
-import org.rust.lang.core.psi.ext.containingCargoPackage
+import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.psi.rustFile
 import org.rust.openapiext.pathAsPath
 import org.rust.openapiext.toPsiFile
@@ -69,12 +68,12 @@ class AttachFileToModuleFix(
     }
 
     companion object {
-        fun findAvailableModulesForFile(project: Project, file: RsFile): List<RsMod> {
+        fun findAvailableModulesForFile(project: Project, file: RsFile): List<RsFile> {
             val virtualFile = file.virtualFile ?: return emptyList()
             val pkg = project.cargoProjects.findPackageForFile(virtualFile) ?: return emptyList()
 
             val directory = virtualFile.parent ?: return emptyList()
-            val modules = mutableListOf<RsMod>()
+            val modules = mutableListOf<RsFile>()
 
             if (file.isModuleFile) {
                 // module file in parent directory
@@ -110,13 +109,13 @@ class AttachFileToModuleFix(
     }
 }
 
-private fun selectModule(file: RsFile, availableModules: List<RsMod>): RsMod? {
+private fun selectModule(file: RsFile, availableModules: List<RsFile>): RsFile? {
     if (isUnitTestMode) {
         val mock = MOCK ?: error("You should set mock module selector via withMockModuleAttachSelector")
         return mock(file, availableModules)
     }
 
-    val box = ComboBox<RsMod>()
+    val box = ComboBox<RsFile>()
     with(box) {
         for (module in availableModules) {
             addItem(module)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -164,16 +160,28 @@ private fun insertFileToModule(file: RsFile, mod: RsMod) {
...
     }
 }
 
+private fun insertModItem(item: RsModDeclItem, module: RsMod): RsModDeclItem {
+    val anchor = module.firstItem
+    val existingMod = generateSequence(anchor) {
+        it.getNextNonCommentSibling() as? RsElement
+    }
+        .filterIsInstance<RsModDeclItem>()
+        .lastOrNull()
```
Undin, line 173:
```suggestion
    val existingMod = anchor?.rightSiblings
        ?.filterIsInstance<RsModDeclItem>()
        ?.lastOrNull()
```

I suppose it does the same thing

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -164,16 +160,28 @@ private fun insertFileToModule(file: RsFile, mod: RsMod) {
...
+
+    val inserted = if (existingMod != null) {
+        module.addAfter(item, existingMod)
+    } else if (anchor != null) {
+        module.addBefore(item, anchor)
+    } else {
+        module.add(item)
+    }
```
Undin, line 181: Let's use more idiomatic Kotlin style and replace `if`s with `when`
```suggestion
    val inserted = when {
        existingMod != null -> module.addAfter(item, existingMod)
        anchor != null -> module.addBefore(item, anchor)
        else -> module.add(item)
    }
```

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -193,3 +201,5 @@ fun withMockModuleAttachSelector(
         MOCK = null
     }
 }
+
+private val RsMod.firstItem: RsElement? get() = itemsAndMacros.firstOrNull { it !is RsAttr && it !is RsVis }
```
Undin, line 205: Do I understand correctly, that `RsMod` is always `RsFile` in this quick fix?
If yes then `it !is RsVis` looks redundant because `RsFile` cannot contain visibility element as a direct child. In auto-import implementation, it makes sense since we can insert `use` item into inline module

Undin: Also, if we use only `RsFile`s, it would be great to use them instead of `RsMod` to provide more info about the context to the reader

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -164,16 +160,28 @@ private fun insertFileToModule(file: RsFile, mod: RsMod) {
...
+    }
+        .filterIsInstance<RsModDeclItem>()
+        .lastOrNull()
+
+    val inserted = if (existingMod != null) {
+        module.addAfter(item, existingMod)
+    } else if (anchor != null) {
+        module.addBefore(item, anchor)
+    } else {
+        module.add(item)
```
Undin, line 180: Let's add a test when the module contains only doc comments. I expect that we insert mod declaration after comments (and we really do it!). Just to commit behavior in this case.

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
Kobzol: `RsMod` is indeed always `RsFile`, I changed it to be more concrete, thanks.

src/main/kotlin/org/rust/ide/inspections/fixes/AttachFileToModuleFix.kt
```diff
@@ -164,20 +163,28 @@ private fun insertFileToModule(file: RsFile, mod: RsMod) {
...
-            mod.addBefore(modItem, child)
-        } as RsModDeclItem
-        inserted.navigate(true)
+        insertModItem(modItem, targetFile).navigate(true)
     }
 }
 
+private fun insertModItem(item: RsModDeclItem, module: RsFile): RsModDeclItem {
+    val anchor = module.firstItem
+    val existingMod = (anchor as? RsModDeclItem) ?: anchor?.rightSiblings
```
Undin, line 172: If you have more than one mod declarations in target file, this code inserts new mod declaration after the first one instead of the latest one. Consider using something like `module.childrenOfType<RsModDeclItem>().lastOrNull()`. Yep, it creates an intermediate list object and it's not lazy. But laziness doesn't make sense here because we iterate over all elements anyway and suggested version is more readable, in my opinition
Let's add test with several existent mod declarations to check it

Kobzol: Damn, I'm surprised that none of the tests covered this. The cast was in a bad place, it should have been 
```
val existingMod = anchor?.rightSiblings
        ?.filterIsInstance<RsModDeclItem>()
        ?.lastOrNull()
        ?: anchor as? RsModDeclItem
```
My solution skips `mod` items before attributes, but that's an error anyway, so I used yours, it's more readable.

PR #5939, opened 2020-08-15
RUN: Add build task for wasm-pack configurations

avrong:
Adds `Before launch` task for `wasm-pack` run configurations. Emulates the first part of wasm-pack process — executing `cargo build` with specific arguments like `--target wasm32-unknown-unknown`.

Provides a way to show cargo build tool window before `wasm-pack` packing instead of raw printing in the terminal.

The implementation is mostly similar to `CargoBuildTaskProvider`.

Related to #3066

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
@@ -1,0 +1,75 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.runconfig.wasmpack
+
+import com.intellij.execution.configurations.RunConfiguration
+import com.intellij.execution.runners.ExecutionEnvironment
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.util.Key
+import com.intellij.util.execution.ParametersListUtil
+import org.rust.cargo.runconfig.buildtool.RsBuildTaskProvider
+import org.rust.cargo.runconfig.command.CargoCommandConfiguration
+import org.rust.cargo.runconfig.command.CargoCommandConfigurationType
+import org.rust.cargo.util.splitOnDoubleDash
+import org.rust.openapiext.project
+import org.rust.stdext.buildList
+
+class WasmPackBuildTaskProvider : RsBuildTaskProvider<WasmPackBuildTaskProvider.BuildTask>() {
+    override fun getId(): Key<BuildTask> = ID
+
+    override fun createTask(runConfiguration: RunConfiguration): BuildTask? =
+        if (runConfiguration is WasmPackCommandConfiguration) BuildTask() else null
+
+    override fun executeTask(
+        context: DataContext,
+        configuration: RunConfiguration,
+        environment: ExecutionEnvironment,
+        task: BuildTask
+    ): Boolean {
+        if (configuration !is WasmPackCommandConfiguration) return false
+
+        val project = context.project ?: return false
+        val configurationArgs = ParametersListUtil.parse(configuration.command)
+        val (preArgs, postArgs) = splitOnDoubleDash(configurationArgs)
+        val configurationCommand = configurationArgs.firstOrNull() ?: return false
+
+        val parameters = buildList<String> {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+            if (!preArgs.contains("--dev")) {
+                addIfNotPresent("--release")
+            }
+
+            if (!this.contains("--target")) {
+                addAll(listOf("--target", WASM_TARGET))
+            }
+        })
+
+        val buildConfiguration = CargoCommandConfiguration(
```
ortem, line 59: Probably it's better to use `RunManager.getInstance(project).createConfiguration(configuration.name, CargoCommandConfigurationType::class.java)`. Also, do we need a name here, or maybe just pass an empty string as a name?

src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    override fun executeTask(context: DataContext, configuration: RunConfiguration, environment: ExecutionEnvironment, task: BuildTask): Boolean {
+        if (configuration !is WasmPackCommandConfiguration) return false
+
+        val project = context.project ?: return false
+        val configurationArgs = ParametersListUtil.parse(configuration.command)
+        val (preArgs, postArgs) = splitOnDoubleDash(configurationArgs)
+        val configurationCommand = configurationArgs.firstOrNull() ?: return false
+
+        val buildCommand = ParametersListUtil.join(mutableListOf("build").apply {
```
ortem, line 43: I'd recommend extracting a variable to have something like
```Kotlin
val parameters = ...
val buildCommand = ParametersListUtil.join(parameters)
```

src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+            addAll(postArgs)
+
+            if (configurationCommand == "test") {
+                addIfNotPresent("--tests")
```
ortem, line 47: Maybe replace `linkedSetOf` with `mutableListOf` (and add `.toList()` at the end) to use `add` instead?

ortem (changes requested):

src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
ortem: I'm not sure actually, @mchernyavsky maybe you have some comments?

ortem: @avrong Let's finally use `RunManager` here and leave the name if it is displayed in the UI somehow

src/main/kotlin/org/rust/cargo/runconfig/wasmpack/WasmPackBuildTaskProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.rust.cargo.runconfig.buildtool.CargoBuildConfiguration
+import org.rust.cargo.runconfig.buildtool.CargoBuildManager
+import org.rust.cargo.runconfig.command.CargoCommandConfiguration
+import org.rust.cargo.runconfig.command.CargoCommandConfigurationType
+import org.rust.cargo.util.splitOnDoubleDash
+import org.rust.openapiext.project
+import java.util.concurrent.CompletableFuture
+import javax.swing.Icon
+
+class WasmPackBuildTaskProvider : BeforeRunTaskProvider<WasmPackBuildTaskProvider.BuildTask>() {
```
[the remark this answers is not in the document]
mchernyavsky: [Introduce_`RsBuildTaskProvider`.patch.zip](https://github.com/intellij-rust/intellij-rust/files/5504545/Introduce_.RsBuildTaskProvider.patch.zip)

PR #5946, opened 2020-08-16
CI: create a release branch on a schedule from CI

Undin:
These changes add cron trigger for `release-branch.yml` to create release branch on a schedule.
The implementation looks at next milestone each day and if there are less than 7 days before new release, it creates new release branch from master

### DIFF
--- a/scripts/check_release_branch_date.py
+++ b/scripts/check_release_branch_date.py
@@ -1,0 +1,22 @@
+import json
+from datetime import datetime, date, timedelta
+from urllib import request
+
+from common import get_patch_version, env
+
+if __name__ == '__main__':
+    repo = env("GITHUB_REPOSITORY")
+    response = request.urlopen(f"https://api.github.com/repos/{repo}/milestones")
+    milestones = json.load(response)
+
+    milestone_version = f"v{get_patch_version()}"
+    milestone = next(milestone for milestone in milestones if milestone["title"] == milestone_version)
+    # TODO: find out more correct way to parse data
+    release_date = datetime.strptime(milestone["due_on"], "%Y-%m-%dT%H:%M:%SZ").date()
+
+    today = date.today()
+    delta = release_date - today
+    week = timedelta(7)
+
+    # Should be synchronized with .github/workflows/release-branch.yml
+    print(json.dumps({"create_release_branch": delta < week}))
--- a/.github/workflows/release-branch.yml
+++ b/.github/workflows/release-branch.yml
@@ -1,6 +1,9 @@
 name: release branch
 
-on: workflow_dispatch
+on:
+    workflow_dispatch:
+    schedule:
+        - cron: '0 1 * * *' # try everyday
 
 jobs:
     make-release-branch:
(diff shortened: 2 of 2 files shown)


scripts/check_release_branch_date.py
```diff
@@ -0,0 +1,27 @@
+import json
+from datetime import datetime, date, timedelta
+from urllib import request
+
+from common import get_patch_version, env
+
+if __name__ == '__main__':
+    repo = env("GITHUB_REPOSITORY")
+    response = request.urlopen(f"https://api.github.com/repos/{repo}/milestones")
```
Undin, line 9: I don't use `request` or another more convenient library on purpose not to complicate python setup on CI for single HTTP request

lancelote: Fair enough 👍

scripts/check_release_branch_date.py
```diff
@@ -0,0 +1,27 @@
...
+    delta = release_date - today
+    week = timedelta(7)
+
+    if delta < week:
+        value = "true"  # need to create new release branch
+    else:
+        value = "false"  # too early to create release branch yet
+
+    # Should be synchronized with .github/workflows/release-branch.yml
+    print(f"{{ \"create_release_branch\": {value} }}")
```
Kobzol, line 27: `print(json.dumps({ "create_release_branch": delta < week }))` is maybe more readable? And it doesn't require an additional library.

scripts/check_release_branch_date.py
```diff
@@ -0,0 +1,27 @@
...
+
+if __name__ == '__main__':
+    repo = env("GITHUB_REPOSITORY")
+    response = request.urlopen(f"https://api.github.com/repos/{repo}/milestones")
+    milestones = json.load(response)
+
+    milestone_version = f"v{get_patch_version()}"
+    milestone = next(milestone for milestone in milestones if milestone["title"] == milestone_version)
+    # TODO: find out more correct way to parse data
+    release_date = datetime.strptime(milestone["due_on"], "%Y-%m-%dT%H:%M:%SZ").date()
```
avrong, line 15: Maybe we can use [datetime.date.fromisoformat](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) (Python 3.7+) here to parse data

Undin: I tried it but it doesn't work. I think it's because it supports only `YYYY-MM-DD` format

scripts/check_release_branch_date.py
Undin: Thanks! It's definitely much clearer

scripts/check_release_branch_date.py
avrong: Oops, wrong one, [datetime.datetime.fromisoformat](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat).
It fails to work with timezone (`Z` at the end), but without it is ok.

scripts/check_release_branch_date.py
```diff
@@ -0,0 +1,22 @@
...
+
+from common import get_patch_version, env
+
+if __name__ == '__main__':
+    repo = env("GITHUB_REPOSITORY")
+    response = request.urlopen(f"https://api.github.com/repos/{repo}/milestones")
+    milestones = json.load(response)
+
+    milestone_version = f"v{get_patch_version()}"
+    milestone = next(milestone for milestone in milestones if milestone["title"] == milestone_version)
```
lancelote, line 13: Some proper error handling here can be helpful, if no milestones were found it will fail with `StopIteration` exception. The step itself will "succeed" and the next step will fail with some cryptic "The template is not valid" error. See [my example](https://github.com/lancelote/intellij-rust/runs/994205961?check_suite_focus=true) when I accidentally run the workflow against my fork with no milestones available. Though it is not crucial.

lancelote:
As we need the changes by tomorrow ideally and the comments are not show stoppers let's merge.

lancelote:
bors merge

Issue #5768, opened 2020-07-20
Incorrect handling of backslashes in "cargo test" output


johnw42:
## Environment

* **IntelliJ Rust plugin version:** 0.3.126.3220-201
* **Rust toolchain version:** 1.45.0
* **IDE name and version:** IntelliJ IDEA 2020.1.3 (Community Edition) Build #IC-201.8538.31, built on July 7, 2020
* **Operating system:** Windows 10

## Problem description

In the output from "cargo test", backslashes are processed according to C/JSON conventions instead of being shown literally. This is a big problem on Windows, since file paths contain backslashes.

## Steps to reproduce

Execute a "cargo test" command that has backslashes in the output. For instance, `println!("foo\\bar");` gets rendered as `foar` because the `\b` is interpreted as a backspace. I tried removing the call to `unescape` in `org.rust.cargo.runconfig.test.CargoTestEventsConverter.handleTestMessage`, and it seemed to fix the problem.

Issue #5703, opened 2020-07-06
Backslashes not shown in test output


mahkoh:
I'm using CLion 2020.1.2 with IR 0.2.125.3191-201

## Problem description

Title

## Steps to reproduce

```rust
#[test]
fn t() {
    eprintln!("a\\\\b");
    eprintln!("a\\b");
    eprintln!("ab");
}
```

`cargo test -- --nocapture` output:

```
a\\b
a\b
ab
```

IR output:

```
a\b

ab
```

I believe this worked before so possibly caused by #5611

Issue #5924, opened 2020-08-13
Do not format string literals in "cargo test" output


maxim-banaev:
## Environment

* **IntelliJ Rust plugin version:** 
* **Rust toolchain version:**
* **IDE name and version:** WS-202.6397.88
* **Operating system:** Windows 10(amd64) v10.0

## Problem description

Cargo test output formats strings in assertion which may hard to understand where is difference when running test in terminal behaves differently.

Running in IDE:
![image1](https://user-images.githubusercontent.com/37836437/90109406-6fbefb80-dd54-11ea-8f3d-3cf16212914b.png)

Running in terminal:
![image](https://user-images.githubusercontent.com/37836437/90109411-7188bf00-dd54-11ea-9c01-6138914dcc2e.png)

Is it possible to disable formatting somehow?

Submitted to WebStorm/CLion bug tracker https://youtrack.jetbrains.com/issue/CPP-21722

Undin:
JFTR I personally don't think that we should show exactly the same output as Cargo produces - IDE is not console tool.
To check difference between expected and actual results, a user can check diff via `<Click to see difference>` link or `Show Diff` action (`Ctrl+D` or `Cmd+D`) with selected test items in test tree. In my opinion, it's much more convinient than manual comparison in console

![image](https://user-images.githubusercontent.com/2539310/91278059-5ffbda00-e78c-11ea-8d2a-2e3a75223795.png)

So the only problem that I see that we don't show anything useful when expected and actual results contain different line separators.
For example, for the following Kotlin code
```kotlin
fun test() {
    assertEquals("qwe\nqwe", "qwe\r\nqwe")
}
```
I get the following diff
![image](https://user-images.githubusercontent.com/2539310/91278293-b537eb80-e78c-11ea-8fb6-919f789590b4.png)

But for Rust analogue
```rust
#[test]
fn foo() {
    assert_eq!("qwe\nqwe", "qwe\r\nqwe")
}
```
we don't show message about line separators
![image](https://user-images.githubusercontent.com/2539310/91278435-e3b5c680-e78c-11ea-901b-6c435dcdd96e.png)

mati865:
I believe this issue has been fixed at some point, using test case from comment above:
![image](https://user-images.githubusercontent.com/1174646/206794251-a1905bfd-d81c-44fd-be32-12b27dadf557.png)

PR #5949, opened 2020-08-17
T: Don't unescape test output

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5768.
Fixes https://github.com/intellij-rust/intellij-rust/issues/5703.
Fixes https://github.com/intellij-rust/intellij-rust/issues/5974.
Fixes https://github.com/intellij-rust/intellij-rust/issues/5985.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
@@ -51,7 +51,7 @@ class CargoTestEventsConverter(
         handleStartMessage(text)
 
         val jsonObject = try {
-            val escapedText = text.replace(DOCTEST_PATH_RE) { it.value.replace("\\", "\\\\") }
+            val escapedText = text.replace(NAME_RE) { it.value.replace("\\", "\\\\") }
             val reader = JsonReader(StringReader(escapedText)).apply { isLenient = true }
             JsonParser.parseReader(reader).takeIf { it.isJsonObject }?.asJsonObject
         } catch (e: JsonSyntaxException) {
@@ -119,8 +119,7 @@ class CargoTestEventsConverter(
                 // Parse `rustdoc` test name:
                 // src/lib.rs - qualifiedName (line #i) -> qualifiedName (line #i)
                 val qualifiedName = it.name.substringAfter(" - ")
-                val stdout = it.stdout?.let(::unescape)
-                it.copy(name = "$target::$qualifiedName", stdout = stdout)
+                it.copy(name = "$target::$qualifiedName")
             } ?: return false
         val messages = createServiceMessagesFor(testMessage) ?: return false
         for (message in messages) {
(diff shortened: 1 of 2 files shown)


Undin (commented):
@mchernyavsky Do I understand correctly that we can't distinguish `\` as general symbol (for example, as path separator on Windows) and as escape symbol in Cargo messages?

mchernyavsky:
@Undin Yes. This PR reverts https://github.com/intellij-rust/intellij-rust/pull/5611 (1) and closes https://github.com/intellij-rust/intellij-rust/issues/5924 (2). Should I move the second part to a separate PR?

src/test/kotlin/org/rustSlowTests/cargo/runconfig/test/CargoTestNodeInfoTest.kt
```diff
@@ -52,26 +52,14 @@ class CargoTestNodeInfoTest : CargoTestRunnerTestBase() {
        assert_ne!(123, 123, "123 == 123");
     """, "123 == 123")
 
+    fun `test don't unescape error messages`() = checkErrors("""
+       assert_eq!("a\\\\b", "a\\b", "`a\\\\b` != `a\\b`");
+    """, "`a\\\\b` != `a\\b`")
+
     @MinRustcVersion("1.39.0")
-    fun `test successful output`() = checkOutput("""
```
Undin, line 56: Let's keep this test and change expected output just to commit current behavior and not to write these tests again when we will fix problem with escaping

Undin (approved):
LGTM
Don't forget to file the corresponding issue in cargo repo

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
```diff
@@ -51,7 +51,7 @@ class CargoTestEventsConverter(
         handleStartMessage(text)
 
         val jsonObject = try {
-            val escapedText = text.replace(DOCTEST_PATH_RE) { it.value.replace("\\", "\\\\") }
+            val escapedText = text.replace(NAME_RE) { it.value.replace("\\", "\\\\") }
```
Undin, line 54: Let's create an issue in rust or cargo repo to fix this bug and add link to is here

Undin: BTW looks like it doesn't work with properly escaped backslashes - it will double them. Can we do nothing with correct escaped backslashes?

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
```diff
@@ -261,7 +260,7 @@ class CargoTestEventsConverter(
         private const val NAME_SEPARATOR: String = "::"
         private const val DOCTESTS_SUFFIX = "doctests"
 
-        private val DOCTEST_PATH_RE: Regex = """"[0-9 a-z_A-Z\-\\.]+ - """.toRegex()
+        private val NAME_RE: Regex = """"name":\s+"(?<name>[^"]+)"""".toRegex()
```
Undin, line 263: Spaces are not necessary after `:` in json
```suggestion
        private val NAME_RE: Regex = """"name":\s*"(?<name>[^"]+)"""".toRegex()
```

Undin (approved):
LGTM except comments

src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
mchernyavsky: https://github.com/rust-lang/rust/issues/76151

mchernyavsky: > BTW looks like it doesn't work with properly escaped backslashes - it will double them. Can we do nothing with correct escaped backslashes?

It's not possible to open a file/directory with backslash in name from IDE.

mchernyavsky (commented):
bors r+

PR #5955, opened 2020-08-18
RES: Add name to RsResolveProcessor

dima74:
This will be used in new resolve to improve performance (instead of iterating all items in scope, only item with specified name will be processed).

Potentially it can be used in other places, e.g. in [`processModDeclResolveVariants`](https://github.com/intellij-rust/intellij-rust/blob/e2d8902505475453db2ddc9629b3cd2ec9b68ee9/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt#L226)


vlad20012:
:+1: 
bors r+

PR #5956, opened 2020-08-18
COMP: Add `dbgr` postfix completion

mgrachev:
Hi there!

I think it would be useful to convert a value to a reference right away when using `dbg` postfix completion .
Otherwise, we won't be able to use the value in the next line.

```rust
// Old behaviour:
fn foo(slice: &[i32]) {
    let first = slice[0];
    dbg!(first);

   do_something(first); // It won't work
}

// New behaviour:
fn foo(slice: &[i32]) {
    let first = slice[0];
    dbg!(&first);

   do_something(first); // It will work
}
```

### DIFF
--- a/src/main/resources/postfixTemplates/DbgrPostfixTemplate/description.html
+++ b/src/main/resources/postfixTemplates/DbgrPostfixTemplate/description.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+Creates <code>dbg!</code> macro invocation from expression.
+</body>
+</html>
--- a/src/main/resources/postfixTemplates/DbgrPostfixTemplate/description.html
+++ b/src/main/resources/postfixTemplates/DbgrPostfixTemplate/description.html
@@ -1,5 +1,5 @@
 <html>
 <body>
-Creates <code>dbg!</code> macro invocation from expression.
+Creates <code>dbg!</code> macro invocation from expression reference.
 </body>
 </html>
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -71,6 +71,7 @@ mbStavola
 mcarlin
 mchernyavsky
 mfarrugi
+mgrachev
 mhaessig
 mibac138
 mitchhentges
--- a/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
@@ -34,6 +34,7 @@ class RsPostfixTemplateProvider : PostfixTemplateProvider {
         IterPostfixTemplate("for", this),
         PrintlnPostfixTemplate(this),
         DbgPostfixTemplate(this),
+        DbgrPostfixTemplate(this),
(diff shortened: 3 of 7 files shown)


Kobzol:
What's wrong with `let first = dbg!(slice[0]);`?

mgrachev:
> What's wrong with `let first = dbg!(slice[0]);`?

Everything is fine, but there is just a simple example.
At work, I use the dbg macro a lot and I prefer to write `dbg! (&var)` than the longer version of `let var = dbg!(var)`.

Kobzol:
I think that `dbg!` was mainly meant to be used as a wrapper about existing expressions (see RFC https://rust-lang.github.io/rfcs/2361-dbg-macro.html#returning-the-value-that-was-given), not to be introduced as a new statement. However, for parameters and mutable variables, it is also useful to use it as @mgrachev has described.

Maybe we could add another postfix, for example `dbgr`, that would generate the `dbg!(&value)` variant?

mgrachev:
> Maybe we could add another postfix, for example `dbgr`, that would generate the `dbg!(&value)` variant?

I like that idea 👍

Undin:
@mgrachev Sory for delay

> I like that idea 👍 Should I implement it?

If it helps you and if you would like - then yes 😸 . But it would be great to ensure that new name in completion (currently, I mean `dbgr`) won't break existing user behaviors. For example, if I previously typed `dbg` or just `db` and press `Enter` to invoke `dbg` postfix completion, it should work the same with new postfix completion template. It seems it will work in the described way because of alphabet sorting of completion items but It would be great to check it

mgrachev:
@Undin, @Kobzol Done.

> But it would be great to ensure that new name in completion (currently, I mean `dbgr`) won't break existing user behaviors. For example, if I previously typed `dbg` or just `db` and press `Enter` to invoke `dbg` postfix completion, it should work the same with new postfix completion template. It seems it will work in the described way because of alphabet sorting of completion items but It would be great to check it.

Usually the first item is always the `dbg` postfix completion, but sometimes it is `dbgr` 🤔

dbgr

src/main/resources/postfixTemplates/DbgrPostfixTemplate/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+Creates <code>dbg!</code> macro invocation from expression.
```
Undin, line 3: Let's rewrite description. Currently, it's exact copy of `dbg` postfix template description and it doesn't describes what `dbgr` does

mgrachev: What would you suggest to write here? 🤔

Undin: For example, `Creates <code>dbg!</code> macro invocation from expression reference.` or something limilar

mgrachev: Thank you! Done.

Undin (approved):
Great!
Thank you!
Welcome to IntelliJ Rust contributors 🎉 
bors r+

Issue #5958, opened 2020-08-18
Applying lint hint can lead to code destruction


krojew:
## Environment

* **IntelliJ Rust plugin version:** 0.3.128.3278-202
* **Rust toolchain version:** 1.47 nightly
* **IDE name and version:** CLion 2020.2
* **Operating system:** Windows 10

## Problem description

The plugin can apply clippy suggestions to fix warnings. clippy sometimes uses ellipsis (...) to omit chunks of code in suggested fixes and the plugin happily changes the actual code to ... when applying the suggestion. This can be somewhat annoying when the changes have not been committed yet and your OS suddenly shuts down, therefore loosing hours of work.

## Steps to reproduce

Apply a lint fix when ... is present.

mchernyavsky:
@krojew Do you use external linter on the fly and apply suggestions in the editor?

![image](https://user-images.githubusercontent.com/6079006/91339491-3d46e100-e7df-11ea-8cde-66b4f91cae8a.png)

Or you use `Run External Linter` action and apply suggestions in the `Inspection Results` window?

![image](https://user-images.githubusercontent.com/6079006/91339602-64051780-e7df-11ea-8792-d410acbfb42d.png)

krojew:
I'm using `Run External Linter` and applying suggestions from the results window.

PR #5967, opened 2020-08-20
ANN: Don't suggest rustc's quickfixes with placeholders or unspecified applicability in batch mode

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/5958.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsExternalLinterPass.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsExternalLinterPass.kt
@@ -108,7 +108,7 @@ class RsExternalLinterPass(
         if (file !is RsFile || !file.isValid) return
         try {
             annotationHolder.runAnnotatorWithContext(file) { _, holder ->
-                holder.createAnnotationsForFile(file, annotationResult)
+                holder.createAnnotationsForFile(file, annotationResult, isOnTheFly = true)
             }
         } catch (t: Throwable) {
             if (t is ProcessCanceledException) throw t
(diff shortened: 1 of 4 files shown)


mchernyavsky:
@Undin Done.

Undin (commented):
Hm, in both tests quick fixes from compiler are quite useful
I'd prefer to keep them.
Can we come up with more smart way to filter out incorrect quick fixes?
Is the original issue related to clippy only?

mchernyavsky:
@Undin 

> Can we come up with more smart way to filter out incorrect quick fixes?

According to the [docs](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/enum.Applicability.html#variants) we should filter at least fixes with `HasPlaceholders` applicability.

> Is the original issue related to clippy only?

No. Clippy runs `cargo check` under the hood and shows its messages.

mchernyavsky:
@Undin I have updated the PR. Now it filters quick fixes only if the inspection was ran in batch mode (i.e. via `Run External Linter` action or `Analyze | Inspect Code`).

mchernyavsky:
Not sure how to test this because it looks like the tests require both `RsWithToolchainTestBase` and `RsInspectionsTestBase` classes.

Undin:
@mchernyavsky ideally, we should reuse `RsAnnotationTestFixture` here since it was created exactly to be used in different classes (for example, `RsInspectionsTestBase` and `RsMultipleInspectionsTestBase`) and avoid code duplication but I'm not sure that it has all necessary things to check global inspection.
If it's hard to use it, I think we can create custom test class in this case and just copy necessary parts from `RsInspectionsTestBase`/`RsAnnotationTestFixture`

Also, looks like we don't have tests for global inspections yet. You can check how they are tested in IDEA repo

mchernyavsky:
Closed by https://github.com/intellij-rust/intellij-rust/pull/8436.

PR #5979, opened 2020-08-23
REF: Check visibility conflicts in move top-level items refactoring

dima74:
Checks that target of any reference remains accessible after move (in [move items refactoring](https://github.com/intellij-rust/intellij-rust/issues/3531#issuecomment-643142894)) and otherwise shows warning dialog. These cases are checked:
- `RsPath`
- struct/enum field (plain access / struct literal / destructuring)
- struct/trait method call

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveCommonProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveCommonProcessor.kt
@@ -7,14 +7,19 @@ package org.rust.ide.refactoring.move.common
 
 import com.intellij.openapi.application.runReadAction
 import com.intellij.openapi.progress.ProcessCanceledException
+import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Key
+import com.intellij.openapiext.isUnitTestMode
+import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiReference
 import com.intellij.psi.impl.source.DummyHolder
 import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.psi.search.searches.ReferencesSearch
 import com.intellij.refactoring.RefactoringBundle.message
 import com.intellij.usageView.UsageInfo
+import com.intellij.util.IncorrectOperationException
+import com.intellij.util.containers.MultiMap
 import org.rust.ide.refactoring.move.common.RsMoveUtil.LOG
 import org.rust.ide.refactoring.move.common.RsMoveUtil.containingModStrict
 import org.rust.ide.refactoring.move.common.RsMoveUtil.isAbsolute
@@ -29,6 +34,7 @@ import org.rust.ide.utils.import.RsImportHelper
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.openapiext.computeWithCancelableProgress
+import org.rust.openapiext.runWithCancelableProgress
 
 class ItemToMove(val item: RsItemElement) : ElementToMove()
 class ModToMove(val mod: RsMod) : ElementToMove()
@@ -98,12 +104,20 @@ class RsMoveCommonProcessor(
     private val codeFragmentFactory: RsCodeFragmentFactory = RsCodeFragmentFactory(project)
 
     private val sourceMod: RsMod = elementsToMove
+        .also { if (it.isEmpty()) throw IncorrectOperationException("No items to move") }
         .map { it.element.containingModStrict }
         .distinct()
         .singleOrNull()
         ?: error("Elements to move must belong to single parent mod")
 
     private val pathHelper: RsMovePathHelper = RsMovePathHelper(project, targetMod)
+    private lateinit var conflictsDetector: RsMoveConflictsDetector
+
+    init {
+        if (targetMod == sourceMod) {
+            throw IncorrectOperationException("Source and destination modules should be different")
+        }
+    }
 
     fun findUsages(): Array<RsMoveUsageInfo> {
         return elementsToMove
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
```diff
@@ -0,0 +1,159 @@
...
+                    val fields = struct.tupleFields?.tupleFieldDeclList ?: continue@loop
+                    for (field in fields) {
+                        checkVisibility(element, field)
+                    }
+                }
+                is RsPath -> {
+                    // conflicts for simple paths are handled using `pathNewAccessible`/`pathNewFallback` machinery
+                    val isInsideSimplePath = element.parentsWithSelf
+                        .takeWhile { it is RsPath }
+                        .any { isSimplePath(it as RsPath) }
```
ortem, line 138: Not sure I undestand this code right. Does `.any` should be really used, and not `.all`? Or maybe even `isSimplePath(element.rootPath())`?

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
```diff
@@ -0,0 +1,159 @@
...
+        val elementsToCheck = sourceFile.descendantsOfType<RsElement>() - movedElementsShallowDescendants
+        detectPrivateFieldOrMethodReferences(elementsToCheck.asSequence(), this::checkVisibilityForInsideReference)
+    }
+
+    /** We create temp child mod of [targetMod] and copy [target] to this temp mod */
+    private fun checkVisibilityForInsideReference(referenceElement: RsElement, target: RsVisible) {
+        if (target.visibility == RsVisibility.Public) return
+        // can't use `containingMod` directly,
+        // because if `target` is expanded by macros, then it will be inside special file
+        val targetContainingMod = target.ancestorStrict<RsMod>() ?: return
```
ortem, line 69: @vlad20012 is this a proper way to get potentially expanded containing mod? I supposed `containingMod` is quite enough

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveCommonProcessor.kt
```diff
@@ -98,12 +104,20 @@ class RsMoveCommonProcessor(
...
 
     private val sourceMod: RsMod = elementsToMove
+        .also { if (it.isEmpty()) throw IncorrectOperationException("No items to move") }
         .map { it.element.containingModStrict }
         .distinct()
         .singleOrNull()
         ?: error("Elements to move must belong to single parent mod")
 
     private val pathHelper: RsMovePathHelper = RsMovePathHelper(project, targetMod)
+    private lateinit var conflictsDetector: RsMoveConflictsDetector
```
ortem, line 114: Will this field be used later outside `preprocessUsages`, or why it is not a local variable?

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    companion object {
+        val RS_ELEMENT_FOR_CHECK_INSIDE_REFERENCES_VISIBILITY: Key<RsElement> =
+            Key("RS_ELEMENT_FOR_CHECK_INSIDE_REFERENCES_VISIBILITY")
+    }
+}
+
+fun addVisibilityConflict(conflicts: MultiMap<PsiElement, String>, reference: RsElement, target: RsElement) {
```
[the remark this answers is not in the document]
dima74: Yes, it will be used in [`RsMoveTraitMethodsProcessor`](https://github.com/intellij-rust/intellij-rust/blob/92b32a2758fab80c36abe8d69d97bb478c8b4a0e/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveTraitMethodsProcessor.kt#L71) :)

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveCommonProcessor.kt
dima74: Yes, to [make moved items public if necessary](https://github.com/intellij-rust/intellij-rust/blob/92b32a2758fab80c36abe8d69d97bb478c8b4a0e/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveCommonProcessor.kt#L410)

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
dima74: We are not interested in expanded elements here, and `getTopmostParentInside` will fail if we used `containingMod` when `target` is expanded by macros. Also please see updated comment

ortem: Oh, I misunderstood the old version of this comment, thank you for the explanation :)

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
dima74: Any outermost path has exactly one simple subpath (please see updated comment for `isSimplePath`), so we can't use `all` because it will almost always return `false`.

---

Consider this example: `mod1::mod2::Struct1::func();`. Here the only simple path is `mod1::mod2::Struct1`. It is not necessary to check visibility for path `mod1::mod2`, because we already checked visibility for `mod1`, `mod2` (and `Struct1`) when searched `pathNewAccessbile` using `RsImportHelper`. But we need to check visibility for path `mod1::mod2::Struct1::func` (that is for `func`)

ortem: Thank you, I got it. Probably `isSimplePath` is not the best name, but I don't have any good variants right now. Please add a test (if there is no such) that will fail when `.all` or `isSimplePath(element.rootPath())` is used instead to protect this code from such breaking changes

ortem (approved):
LGTM except for the comment about a test

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
dima74: Ok, thank you, I added tests

dima74:
bors r=ortem

PR #5980, opened 2020-08-24
CARGO: Use package name for cargo project if possible

### LINKED ISSUE #5922: Use project name instead of root folder's name when generating project

avrong:
With #5929, fixes #5922.

When the root is not a package (e.g. with [virtual manifest](https://doc.rust-lang.org/cargo/reference/workspaces.html)), falls back to root directory name.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
@@ -338,8 +338,11 @@ data class CargoProjectImpl(
         rawWorkspace.withStdlib(stdlib, rawWorkspace.cfgOptions, rustcInfo)
     }
 
-    override val presentableName: String
-        get() = workingDirectory.fileName.toString()
+    override val presentableName: String by lazy {
+        workspace?.packages?.singleOrNull {
+            it.origin == PackageOrigin.WORKSPACE && it.rootDirectory == workingDirectory
+        }?.name ?: workingDirectory.fileName.toString()
+    }
 
     private val rootDirCache = AtomicReference<VirtualFile>()
     override val rootDir: VirtualFile?


src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -341,7 +341,9 @@ data class CargoProjectImpl(
     }
 
     override val presentableName: String
-        get() = workingDirectory.fileName.toString()
+        get() = projectService.allPackages.firstOrNull {
```
ortem, line 344: Am I understand right that package name should be only used in case of a single-package project? If so, `singleOrNull` should be used instead, or even `.singleOrNull()?.takeIf { ... }`

avrong: As I understand, here I need to find a package with the same directory as the project one.

There is a mistake, instead of just trying first I think I actually should filter one from all by working directory.

ortem: `.find` and `.firstOrNull` are equivalent and IMO `.firstOrNull` is preferable because of very descriptive name. Look at the definition in Kotlin stdlib:

```Kotlin
public inline fun <T> Iterable<T>.find(predicate: (T) -> Boolean): T? {
    return firstOrNull(predicate)
}
```

But I'm not sure that we should use the name of **some** package as a cargo project name in case of **multiple** packages

avrong: I think I can't imagine how multiple packages can have the same working directory.

If we have a Cargo.toml with a name inside it, then we can find it in the packages list by working directory and show its name instead of its directory. Otherwise, when there is no packages **with the same working directory** (so project directory doesn't contain any root package), we fall back to working directory name.

Undin: I think current approach in general

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
```diff
@@ -339,7 +339,9 @@ data class CargoProjectImpl(
     }
 
     override val presentableName: String
-        get() = workingDirectory.fileName.toString()
+        get() = projectService.allPackages.singleOrNull {
```
Undin, line 342: Let's use `lazy` here to avoid iteration over all packages on each call.
It makes cargo project sorting much slower (`O(n^2logn)` instead of `O(nlogn)`). Of course, it will be noticeable only with a lot of package but anyway

Undin, line 342: Also, you don't need to iterate all available packages in project. Only packages in `workspace` of currect project with `origin == STDLIB`

Undin (approved):
Looks ok except comments
It would be great to have tests for this but it seems we don't have necessary test infrastructure to do it, so let's postpone it. Probably, I will write them myself

src/main/kotlin/org/rust/cargo/project/model/impl/CargoProjectImpl.kt
avrong: I guess you meant `origin == WORKSPACE`, which I implemented now

Undin: Oops. Yep, sorry :sweat_smile:

Undin (approved):
bors r+

avrong:
bors r=Undin

PR #5986, opened 2020-08-24
RES: fix mod path attributes to another dirs

vlad20012:
Fixes name resolution of nested modules in modules
liked to a project with path attributes from
another directories, e.g.
```rust
#[path = "../foo/foo.rs"]
mod foo;
```

Closes #5917

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsMod.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsMod.kt
@@ -53,10 +53,16 @@ interface RsMod : RsQualifiedNamedElement, RsItemsOwner, RsVisible {
         if (this is RsFile && name == RsConstants.MOD_RS_FILE || isCrateRoot) return contextualFile.originalFile.parent
 
         val explicitPath = pathAttribute
+        val superMod = `super`
+        val superModDir = { superMod?.getOwnedDirectory(createIfNotExists) }
         val (parentDirectory, path) = if (explicitPath != null) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsMod.kt
```diff
@@ -54,7 +55,11 @@ interface RsMod : RsQualifiedNamedElement, RsItemsOwner, RsVisible {
 
         val explicitPath = pathAttribute
         val (parentDirectory, path) = if (explicitPath != null) {
-            contextualFile.originalFile.parent to explicitPath
+            if (this is RsFile) {
+                declaration?.contextualFile?.parent to PathUtil.getParentPath(explicitPath)
```
dima74, line 59: If I understand correctly it will not work when `declaration` is inside inline block, e.g.:
```kotlin
fun `test module path to different directory 2`() = stubOnlyResolve("""
//- main.rs
    fn main() {
        inner::foo::bar::func();
    }                  //^ inner/src2/bar.rs

    mod inner {
        #[path = "src2/foo.rs"]
        pub mod foo;
    }
//- inner/src2/foo.rs
    pub mod bar;
//- inner/src2/bar.rs
    pub fn func() {}
""")
```

---

Maybe we can just use parent directory of file when file declaration has path attribute?
```suggestion
            if (this is RsFile) return contextualFile.originalFile.parent
```

dima74:
bors retry

vlad20012:
bors r=dima74

PR #5988, opened 2020-08-24
CARGO: add git crate resolver

Kobzol:
It seems that a lot of possible `Cargo.toml` improvements are blocked on having a better way of searching crates and their versions. In an effort to improve the status quo, in this PR I implement a crate resolver that uses the local Cargo `git` index to search crates. I think that `git` is the only proper way to locally index the crates at this moment, as it seems that Cargo will not provide a command for this (https://github.com/rust-lang/cargo/issues/7824). Even though the user might not have `git` installed, I think that it is very common (especially for `Rust` users) to have it installed, and if it's missing, the resolver will simply stop working (so it is best effort). We can also fallback to the original network-based crates.io resolver if you think that it is a good idea. Another option would be to use something like `JGit` to use `git` programatically and not via the command line.

This PR introduces a `CrateResolverService` with a `git`-based implementation and a mock implementation for tests. The current implementation works as follows:
1) In the constructor of the service, `git` availability is checked and local crates are loaded in-memory into an index which maps the crate name to it's directory path in the `git` index. This may be invoked inside a read action, which doesn't play well with running external commands. Is there a better time to initialize the service or get around the read action restriction?
2) When a crate is searched, the index is traversed to look for all crates matching the name query. Currently there are ~45k crates and the linear search seems fast enough, but maybe we could employ some fuzzy/fulltext search index? Also since the git index doesn't allow to query crate version without parsing the index file of each crate, I removed version from the search results and now only crate name is returned. This makes crate key completion preview contain slightly less information, but otherwise shouldn't be a big deal I hope.
3) When a specific crate is accessed, a cache of crates is checked whether it was already looked up or no. If not, the crate index file is parsed from git, cached and returned. I suppose that this cache should be protected with some lock? Although I'm not familiar with the IntelliJ threading model, so I'm not sure what API should I use here.

I also tried to simply preload the whole index with all crates and their versions at the beginning into memory, but it seems to take too much time.

@Undin What do you think?

Closes: https://github.com/intellij-rust/intellij-rust/pull/3786
Closes: https://github.com/intellij-rust/intellij-rust/pull/4873 (this PR is based on @Jezza's work)


Kobzol:
@Undin ping

Kobzol:
@avrong There are multiple completion providers (and in the future probably also inspections, error annotators, etc.) that need to use the crate index, not just `DependencyCratesLocalCompletionProvider`, which was introduced in your PR. Originally I thought that we could create a common crate resolver interface, implement it both by the local index and by crates.io lookup, and use it in all places where it's required.

But now that I see how many issues were solved by the local index, I'm not sure if it's practical. If the remote index was just slower, it would be fine, but it lacks certain functionality that the local index provides (like better handling of `-`/`_` and searching proper crate names). Therefore the remote index would behave differently than the local one and the interface would need to be accustomed to it (to find the "least common denominator").

Maybe it would be better to throw away the remote index completely, once the local one stabilizes?

avrong:
@Kobzol There is a chance of removing the current Crates.io API, however, I'm still trying to find ways to improve it.

The other thing I see possible in the foreseeable future is Cargo's getting rid of local index fetch in [RFC: Serve crates-io registry over HTTP as static files](https://github.com/rust-lang/rfcs/pull/2789). This can become a problem we can resolve in several ways, e.g. downloading the index ourselves or using the new index HTTP API.

I agree, the idea of crate resolver is useful, for now, it's just not clear yet whether we can and there is a need to create such one for a pair of implementations we can have in the future.

Eh2406:
Hi, one of the Cargo Team, and I've done some work on dependency resolution. I don't know the details about this use case, but if you want to chat about how Cargo does things I am available. We tend to chat at https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo.

PR #5989, opened 2020-08-24
PERF: Don't expand cfg-disabled macro calls

dima74:
Disables expansion of cfg-disabled macro calls during full expansion (e.g. when new project is opened). In Cargo project it results in 50% reduction of total expanded calls (from 76973 to 34145) and about 35% reduction of total expansion time. I suppose main improvement in Cargo project comes from `winapi` crate ([example](https://github.com/retep998/winapi-rs/blob/0.3/src/um/mod.rs)).

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
+++ b/src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
@@ -26,6 +26,7 @@ import com.intellij.psi.StubBasedPsiElement
 import com.intellij.psi.impl.source.StubbedSpine
 import gnu.trove.TIntObjectHashMap
 import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.RsFileType
 import org.rust.lang.core.macros.MacroExpansionManagerImpl.Testmarks
 import org.rust.lang.core.psi.RsFile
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
```diff
@@ -510,19 +511,26 @@ class SourceFile(
         val isExpansionFile = MacroExpansionManager.isExpansionFile(file)
         val isIndexedFile = file.isValid && (isExpansionFile || shouldIndexFile(project, file))
 
-        val stages1 = if (!isIndexedFile) {
+        val invalidateFileInfos = {
             // The file is now outside of the project, so we should not access it.
             // All infos of this file should be invalidated
```
vlad20012, line 516: Let's move this comment into `if (!isIndexedFile) {` body, it's mostly related to the condition itself

src/main/kotlin/org/rust/lang/core/macros/ExpandedMacroStorage.kt
```diff
@@ -510,19 +511,26 @@ class SourceFile(
...
                 info.makeInvalidationPipeline()
             }
-        } else {
+        }
+
+        val stages1 = if (!isIndexedFile) {
+            invalidateFileInfos()
+        } else run {
             // If project file doesn't have crate root
             // we shouldn't try to expand its macro calls
```
vlad20012, line 527: And also move this comment near `if (psiFile.crateRoot == null)` condition

vlad20012:
bors retry

PR #5999, opened 2020-08-25
RUN: Don't try to initialize Cargo Tool Window before project opened

mchernyavsky:


Kobzol:
Ooh, this bug has been plagueing me for most of the last year! :smile: I haven't experienced it in the last ~2 months or so though, hopefully this fixes it.

Kobzol:
.. aand just as I wrote the last comment it occurred now :D I guess it somehow depends on which project is opened, as it only happens on a single project. Let's hope that this fixed it for good.

PR #6001, opened 2020-08-25
PRJ: Add Procedural Macro project template

vlad20012:
(via cargo-generatete template https://github.com/eupn/rust-procmacro-quickstart-template)

image

### DIFF
--- a/src/main/kotlin/org/rust/ide/newProject/RsProjectTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/newProject/RsProjectTemplate.kt
@@ -20,6 +20,7 @@ sealed class RsGenericTemplate(name: String, isBinary: Boolean) : RsProjectTempl
 open class RsCustomTemplate(
     name: String, val url: String
 ) : RsProjectTemplate(name, false, RsIcons.CARGO_GENERATE) {
+    object ProcMacroTemplate : RsCustomTemplate("Procedural Macro", "https://github.com/intellij-rust/rust-procmacro-quickstart-template")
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/newProject/Utils.kt
```diff
@@ -48,6 +48,7 @@ fun Project.makeDefaultRunConfiguration(template: RsProjectTemplate) {
     val configuration = when (template) {
         is RsGenericTemplate.CargoBinaryTemplate -> configurationFactory.createCargoRunConfiguration()
         is RsGenericTemplate.CargoLibraryTemplate -> configurationFactory.createCargoTestConfiguration()
+        is RsCustomTemplate.ProcMacroTemplate -> configurationFactory.createCargoTestConfiguration()
```
ortem, line 51: Note that a project generated with `ProcMacroTemplate` doesn't contain any tests so this run configuration will be useful only after a user adds `tests` module. So I'm not sure if we really should create a run configuration here

vlad20012: Well, ok, let's remove it for now

ortem (approved):

vlad20012:
Discussed in private with @Undin that it would be more secure to fork the template repo and merge changes from the original repo manually (it is rarely updated so it shouldn't be too annoying). So I forked it to https://github.com/intellij-rust/rust-procmacro-quickstart-template and updateв the link in the PR

vlad20012:
bors r=ortem

PR #6003, opened 2020-08-26
Feature inline diverging function

SaarYogev:
- [x] Allow refactoring if all but the last exit point aren't `return` statements

- [x] Assure the `return` statement is the last statement in the function, and isn't nested in any other expression. Meaning it is at the end of the body.

changelog: Allow inlining functions (using `Inline Function` refactoring) with diverging exit points (like `panic!()`).

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
@@ -132,14 +132,14 @@ class RsInlineFunctionProcessor(
 
     companion object {
         fun doesFunctionHaveMultipleReturns(fn: RsFunction): Boolean {
-            var entryCount = 0
+            val entryPoints =  mutableListOf<ExitPoint>()
             val sink: (ExitPoint) -> Unit = {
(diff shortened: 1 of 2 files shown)


SaarYogev:
This is PR ready @Kobzol, I just didn't realise it earlier.

From what I could tell there's currently no way to put a `return` statement where it won't be the only way to return, still have the code compile with `rustc`, and not fail the refactoring multiple `return` statements check.

- `if` statements need to have one of these options, or they won't compile:
  1. No branches with `return` statements -> irrelevant for our check
  2. All branches with `return` statements -> will fail our check due to multiple returns
  3. Have a diverging statement proceeding them
      - If it's a return statement, it has to be the only one or the check will fail
      - If it's isn't, then we're back to previous cases, `1.` will still pass and `2.` will still fail
- loops diverge on the first iteration or not at all, unless they have `if` statements.

SaarYogev:
@vlad20012 PR is ready for review, squashed the commits too 😄.

SaarYogev:
@Kobzol @vlad20012 Ready for review here too, with a squashed commit. 😄

SaarYogev:
This is ready to be merged, anything missing? @vlad20012

vlad20012:
:+1: Thanks!

bors r+

PR #6004, opened 2020-08-26
Inline function arguments correctly

SaarYogev:
Currently it replaces function parameters with the function arguments.  Still need to make sure arguments will be resolved in the same order were they not inlined by creating variables for them before the main inline. Here's a Kotlin example:
That's the original file. We'll inline `foo()`
```kotlin
var count = 1
private fun main() {
    val a = foo(bar())
}

private fun foo(number: Int): Int {
    return number * bar() + number
}

private fun bar(): Int {
    return count++
}
```
The original code should get `2 * 3 + 2` inside foo.
My code currently inlines the values directly, like so:
```kotlin
var count = 1
private fun main() {
    val a = bar() * bar() + bar()
}

    private fun bar(): Int {
    return count++
}
```
Which leads the calculation to be `2 * 3 + 4`, because the function will be called an additional time.
The correct inline (as shown by Kotlin) is:
```kotlin
var count = 1
private fun main() {
    val number = bar()
    val a = number * bar() + number
}

    private fun bar(): Int {
    return count++
}
```
Which makes sure number has a single value, assined before any calls for `bar()` happening inside `foo()`.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
@@ -19,11 +19,13 @@ import com.intellij.usageView.UsageInfo
 import com.intellij.usageView.UsageViewDescriptor
 import com.intellij.util.IncorrectOperationException
 import com.intellij.util.containers.MultiMap
+import org.jetbrains.annotations.NotNull
 import org.rust.ide.refactoring.RsInlineUsageViewDescriptor
 import org.rust.ide.surroundWith.addStatements
 import org.rust.lang.core.cfg.ExitPoint
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.impl.RsCallExprImpl
 import org.rust.lang.core.resolve.ref.RsReference
 
 class RsInlineFunctionProcessor(
@@ -70,7 +72,7 @@ class RsInlineFunctionProcessor(
                     return@forEach
                 }
                 usagesAsReference.contains(usage.reference) && removeDefinition -> {
-                    conflicts.putValue(usage.element, "Usage with function pointer inline is't currently supported")
+                    conflicts.putValue(usage.element, "Usage with function pointer inline isn't currently supported")
                     return@forEach
                 }
                 caller == null -> {
@@ -89,7 +91,7 @@ class RsInlineFunctionProcessor(
 
     private fun checkCallerConflicts(function: RsFunction, caller: PsiElement): String? {
         val funcArguments = (function.copy() as RsFunction).valueParameters
-        val callArguments = when (caller) {
+        var callArguments = when (caller) {
             is RsCallExpr -> caller.valueArgumentList.exprList
             is RsDotExpr -> {
                 val methodCall = caller.methodCall
@@ -99,11 +101,17 @@ class RsInlineFunctionProcessor(
             else -> return "Unknown caller expression type"
         }
 
+        if (isDoubleSemicolonMethodCall(function, caller)) {
+            callArguments = callArguments.drop(1)
+        }
         if (funcArguments.size != callArguments.size) {
             return "Cannot inline function to references with mismatching arguments"
         }
         return null
     }
+
+    private fun isDoubleSemicolonMethodCall(function: RsFunction, caller: PsiElement) =
+        function.selfParameter != null && caller is RsCallExpr && caller.firstChild is RsPathExpr
 
     override fun performRefactoring(usages: Array<out UsageInfo>) {
         usages.asIterable().forEach loop@{
@@ -168,12 +176,15 @@ class RsInlineFunctionProcessor(
             }
 
             val caller = ref.element.ancestors.filter { it is RsCallExpr || it is RsDotExpr }.firstOrNull()
-                ?: throw IncorrectOperationException("Usage without caller expression parent bypassed preprocessing: ${ref.element.text}")
+                ?: throw IncorrectOperationException(
+                    "Usage without caller expression parent bypassed preprocessing: ${ref.element.text}")
 
             val funcScope = LocalSearchScope(body)
 
             val selfParam = functionDup.selfParameter
             replaceSelfParamWithExpr(selfParam, caller, funcScope)
+
+            replaceParameterUsages(functionDup, caller, funcScope)
 
             val retExprInside = body.descendantsOfType<RsRetExpr>().firstOrNull()?.expr
             if (retExprInside != null && body.stmtList.size == 1) {
@@ -190,7 +201,13 @@ class RsInlineFunctionProcessor(
                         enclosingStatement.addLeftSibling(it)
                     }
                 } else {
-                    addFunctionBodyToCallerWithCorrectSpacing(leftSibling, rightSibling, blockParent, body, childContainingEnclosingStatement)
+                    addFunctionBodyToCallerWithCorrectSpacing(
+                        leftSibling,
+                        rightSibling,
+                        blockParent,
+                        body,
+                        childContainingEnclosingStatement
+                    )
                 }
 
                 blockParent.children.filter { it.text == ";" }.forEach { it.delete() }
@@ -209,7 +226,91 @@ class RsInlineFunctionProcessor(
         }
     }
 
-    private fun addFunctionBodyToCallerWithCorrectSpacing(leftSibling: PsiElement?, rightSibling: PsiElement?, blockParent: RsBlock, body: RsBlock, childContainingEnclosingStatement: PsiElement?) {
+    private fun replaceParameterUsages(functionDup: RsFunction, caller: PsiElement, funcScope: LocalSearchScope) {
+        val parameterBindingsList = functionDup.valueParameters
+            .filter { it !is RsSelfParameter }
+            .map { it.pat }
+            .filterIsInstance(RsPat::class.java)
+
+        val valueArgumentList = when (caller) {
+            is RsCallExpr -> caller.valueArgumentList
+            is RsDotExpr -> caller.methodCall?.valueArgumentList ?: return
+            else -> return
+        }
+
+        val arguments = if (isDoubleSemicolonMethodCall(functionDup, caller)) {
+            valueArgumentList.exprList.drop(1)
+        } else {
+            valueArgumentList.exprList
+        }
+
+        val paramUsagesToReplace = mutableMapOf<RsExpr?, PsiElement>()
+
+        for ((index, parameter) in parameterBindingsList.withIndex()) {
+            paramUsagesToReplace.putAll(replaceParameterUsage(arguments, index, parameter, caller, funcScope))
+        }
+        paramUsagesToReplace.forEach { it.key?.replace(it.value) }
+    }
+
+    private fun replaceParameterUsage(arguments: List<RsExpr>, index: Int, parameter: RsPat, caller: PsiElement,
+                                      funcScope: LocalSearchScope): Map<RsExpr?, @NotNull PsiElement> {
+        if (arguments.size <= index) {
+            return emptyMap()
+        }
+
+        val argument = arguments[index]
+        var name: String = parameter.text.removePrefix("mut ")
+
+        if (name != argument.text) {
+            if (argument.childrenWithLeaves.any { it.text == name }) {
+                name += "_argument"
+            }
+            val bindingMutability = when (parameter) {
+                is RsPatIdent -> parameter.patBinding.mutability.isMut
+                else -> false
+            }
+            val letDeclaration = factory.createLetDeclaration(name, argument, bindingMutability)
+            addLetDeclarationNearCaller(caller, letDeclaration)
+            return getParamUsagesToReplace(parameter, funcScope, letDeclaration)
+        }
+
+        return emptyMap()
+    }
+
+    private fun addLetDeclarationNearCaller(caller: PsiElement, letDeclaration: RsLetDecl) {
+        val callerParentStmt = caller.ancestorOrSelf<RsStmt>()
+        if (callerParentStmt != null) {
+            callerParentStmt.addLeftSibling(letDeclaration)
+        } else {
+            caller.addLeftSibling(letDeclaration)
+        }
+    }
+
+    private fun getParamUsagesToReplace(parameter: RsPat, funcScope: LocalSearchScope, letDeclaration: RsLetDecl):
+        MutableMap<RsExpr?, PsiElement> {
+        val paramUsagesToReplace = mutableMapOf<RsExpr?, PsiElement>()
+        val paramNavigationElement = (parameter as? RsPatIdent)?.patBinding?.navigationElement
+            ?: parameter.navigationElement
+        ReferencesSearch.search(paramNavigationElement, funcScope).findAll().forEach {
+            val declarationID = when (val declarationVarPat = letDeclaration.pat) {
+                is RsPatIdent -> declarationVarPat.patBinding.identifier
+                else -> declarationVarPat
+            }
+            if (declarationID != null) {
+                val declarationElementExpr: RsExpr? = it.element.ancestorOrSelf()
+                paramUsagesToReplace[declarationElementExpr] = declarationID
+            }
+        }
+        return paramUsagesToReplace
+    }
+
+    private fun addFunctionBodyToCallerWithCorrectSpacing(
+        leftSibling: PsiElement?,
+        rightSibling: PsiElement?,
+        blockParent: RsBlock,
+        body: RsBlock,
+        childContainingEnclosingStatement: PsiElement?
+    ) {
         if (leftSibling != null) {
             val isThereSpaceBeforeFunctionCall = leftSibling.nextSibling.text.startsWith("\n\n")
             val isThereSpaceAfterFunctionCall = rightSibling?.prevSibling?.text?.startsWith("\n\n") == true
@@ -268,7 +369,13 @@ class RsInlineFunctionProcessor(
             val selfExprText = buildString {
                 if (selfParam.mut != null) append("mut")
                 append(" ")
-                append((caller as RsDotExpr).expr.text)
+                val callerText = when {
+                    caller is RsDotExpr -> caller.expr.text
+                    caller is RsCallExpr && caller.expr is RsPathExpr ->
+                        (caller as RsCallExprImpl).valueArgumentList.exprList[0].text
+                    else -> caller.text
+                }
+                append(callerText)
             }
 
             val selfExpr = factory.tryCreateExpression(selfExprText)!!
--- a/src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
@@ -6,7 +6,6 @@
 package org.rust.ide.refactoring
 
 import org.intellij.lang.annotations.Language
-import org.junit.Assert
 import org.rust.RsTestBase
 
 class RsInlineFunctionTest : RsTestBase() {
@@ -123,7 +122,8 @@ class RsInlineFunctionTest : RsTestBase() {
     """, """
         fn main() {
             let mut vec = vec![1, 2, 3];
-            vec.push(1);
+            let vec_argument = &mut vec;
+            vec_argument.push(1);
         }
 
     """)
@@ -369,6 +369,31 @@ class RsInlineFunctionTest : RsTestBase() {
             fn test(self) {
                 println!("bla");
             }
+        }
+    """)
+
+    fun `test inline method with struct self parameter and another parameter`() = doTest("""
+        struct S { a: u32, b: u32 }
+
+        impl S {
+            fn /*caret*/foo(&self, a: u32) {
+                println!("{} {} {}", self.a, self.b, a);
+            }
+        }
+
+        fn main() {
+            let s = S { a: 0, b: 0 };
+            S::foo(&s, 1);
+        }
+    """, """
+        struct S { a: u32, b: u32 }
+
+        impl S {}
+
+        fn main() {
+            let s = S { a: 0, b: 0 };
+            let a = 1;
+            println!("{} {} {}", &s.a, &s.b, a);
         }
     """)
 
@@ -653,9 +678,9 @@ class RsInlineFunctionTest : RsTestBase() {
     """)
 
     fun `test inline function with passing by &`() = doTest("""
-        fn f(a: &Vec<i32>) {}
+        fn f(_a: &Vec<i32>) {}
 
-        fn foo() {
+        fn main() {
             let vec = vec![1, 2, 3];
             let vec2 = vec![1, 2, 3];
             let vec3 = vec![1, 2, 3];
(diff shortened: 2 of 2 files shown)


src/main/kotlin/org/rust/ide/refactoring/inline/RsInlineFunctionProcessor.kt
```diff
@@ -222,6 +224,22 @@ class RsInlineFunctionProcessor(
         }
     }
 
+    private fun replaceParametersWithArgumentValues(functionDup: RsFunction, caller: PsiElement, funcScope: LocalSearchScope) {
+        val parameterBindingsList = functionDup.valueParameters
+            .filter { it !is RsSelfParameter }
+            .map { it.pat }
+            .filterIsInstance(RsPatIdent::class.java)
```
vlad20012, line 231:
```suggestion
            .filterIsInstance<RsPatIdent>()
```

vlad20012 (commented):
This wouldn't work with complex patterns in parameters arguments, e.g.
```rust
fn foo((a, b): (i32, i32), c: i32) {}
```
Let's `zip` `functionDup.valueParameters` with `valueArgumentList` directly, and then (for now) just skip parameters with such complex patterns. And let's add a test for this.

Also, I'm not sure how this will work for methods (with `self`!) invoked with UFCS syntax, e.g.
```rust
impl S {
    fn foo(&self, b: i32) {}
}
...
S::foo(S, 42);
```

Let's also add a test for this.

src/main/kotlin/org/rust/ide/refactoring/inline/RsInlineFunctionProcessor.kt
```diff
@@ -276,6 +294,18 @@ class RsInlineFunctionProcessor(
         }
     }
 
+    private fun replaceParamWithExpr(param: RsPatBinding, argument: RsExpr, funcScope: LocalSearchScope) {
+        val argumentText = buildString {
+            append(argument.text.removePrefix("mut ").removePrefix("&"))
```
vlad20012, line 299: Let's add a test where these `removePrefix` are needed

Kobzol:
@SaarYogev If you're going to handle arguments with side expressions in this PR (i.e. creating variables with `let ...` before the function call), then I think you can also handle complex patterns right away, because the solution for them is the same:
```rust
fn bar() {
    foo((1, 1));
    // turns into
    let (a, b) = (1, 1);
    println!("{}", a);
}
fn foo((a, b): (i32, i32)) {
    println!("{}", a);
}
```
For both cases when you introduce the `let` binding you should check if it doesn't conflict with existing variables in the function where the inlining was invoked (and use a different name in that case?).

SaarYogev:
Now I am creating let declarations for arguments with no constant value (not primitives or variables alone).
the only faling test is failing because of a mutability check. This function inside `test inline function with differently named reference and mutable function parameters`

```rust
fn foo(i: &i32, j: &mut i32) -> i32 {
    return i + j + bar();
}
```

has j, which is a mutable reference. However, `parameterBinding.mutability.isMut` returns `false` for him. It seems like a bug, as I wasn't able to find another place where it shown as mutable.

Kobzol:
It doesn't seem like a bug, j is not a mutable binding. It is an immutable binding of type "mutable referenxe to i32".

SaarYogev:
So I guess I need to get the actual mutability from the type data?

Kobzol:
Hmm, why do you even need any special handling of `mut` bindings? If the parameter was `mut`, the created let declaration should be `mut` (and vice versa). Is there any problem with this?

SaarYogev:
@Kobzol No, I just don't know how to get to the parameter instead of the binding.

Kobzol:
Sorry, I don't understand what do you mean by that. Could you show me an example?

SaarYogev:
@Kobzol You told me that I should check if the parameters is *mutable* or not (which I assumed will be easy, something like `function.parameters[0].isMut`), but I only found how to see the properties of the parameterBindings, which as you said won't help me figure it out (using `function.valueParameters`). I would like to know how can I check the mutibility of function parameters, given an `RsFcuntion`.

Kobzol:
Ok, so we had a slight misunderstanding I think. In the test that you have mentioned, the `j` binding should NOT be `mut`, because the corresponding parameter is also not `mut`. In other words, I think that checking only `parameterBinding.mutability.isMut` is enough. If it returns true, create the local binding with `mut`. If not, create an immutable binding.

```rust
fn main() {
    let i = bar();
    let j = qux(); // this should NOT be a mutable binding
    let a = i + j + bar();
}
fn bar() -> &i32 {
    let baz = 5;
    return &baz;
}
fn qux() -> &mut i32 {
    let mut quux = 5;
    return &mut quux;
}
```
However, this test does not compile (with rustc), please try to change it so that it does.

SaarYogev:
@Kobzol I don't understand why, which parameter do you refer to in 

> the corresponding parameter

I see that `qux()` does return a mutable reference.
`j` doesn't have to be mutable to be added to `a`, but I assume we don't intend to scan extra code in the scope where we inline the function into, unless for very basic purposes (like avoiding duplicate variable names).
So I can't tell how would we know that this code isn't possible:
```rust
fn main() {
    let i = bar();
    let mut j = qux(); // making j mut
    j = j + 1; // using j's mutability to increment its value
    let a = i + j + bar();
}
fn bar() -> &i32 {
    let baz = 5;
    return &baz;
}
fn qux() -> &mut i32 {
    let mut quux = 5;
    return &mut quux;
}
```

Kobzol:
`j = ...` is only possible if `j` was declared with `mut` in the original function that is being inlined. We just assume that the original function was correct. If it was correct, it should remain correct after inlining. If it was incorrect, we shouldn't "fix" the code when inlining it.

To sum up: I think that it is valid to create the new binding with `mut` ONLY if the original parameter binding had `mut`. In fact, the binding of the local variable should be exactly the same as the binding of the parameter.

Maybe there is some case where this wouldn't work, but I think that as a first approach it is a valid and simple heuristic.

SaarYogev:
I agree that we should leave the function as is and not try to fix it, but the original function from the failing test has a mutable variable as far as I can tell: 
We inline `foo` like so: 
```rust
fn /*caret*/foo(i: &i32, j: &mut i32) -> i32 {
    return i + j + bar();
}
```

`j` is declared as `&mut`, so I do not understand what's the difference between the type and the binding that makes it appear immutable. I assumed that `j: &mut i32` should convert to `let mut j`, though I can see that reference is gone, don't know what to do about that.

Kobzol:
The mutability of the binding and the type are two separate things. For example:
- `i: &u32` is an immutable binding with type "immutable reference to `u32`"
- `i: &mut u32` is an immutable binding with type "mutable reference to `u32`"
- `mut i: &u32` is a mutable binding with type "immutable reference to `u32`"
- `mut i: &mut u32` is a mutable binding with type "mutable reference to `u32`"

Therefore, `j: &mut i32` is an immutable binding, therefore it should be converted to `let j` and not `let mut j`.

SaarYogev:
I think I do understand now, it's like changing the contents of the object (mutable object) versus changing the reference to a different object (mutable object reference) in an object-oriented language. I'll try to test that now, thanks 😄

Kobzol:
Exactly, that is a good comparison! :)

SaarYogev:
@Kobzol All done, was a pretty easy fix. Now the test is checking both immutable binding of a mutable var, or a mutable binding of an immutable var.

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -866,11 +867,17 @@ class RsInlineFunctionTest : RsTestBase() {
                 let mut quux = 5;
                 return &mut quux;
             }
+
+            fn corge() -> &mut i32 {
```
Kobzol, line 871: Could you also please make the test code compilable with rustc? It could be a bit hard with `&mut` references (although you could use &'static refs and some unsafe code), maybe choose some simpler code snippet?

SaarYogev: I can make `corge()` return a plain `i32` that will have a mutable binding later and be done with that part, but I do need to handle the errors in `qux()`, where a mutable reference is the point. 
I'll try to handle the lifetimes as the compiler suggests, but it might take me some time.

SaarYogev: Well, I have a piece of test code that passes rustc. However, it doesn't pass our own checks, saying `Cannot move` under `*i` and `*j`
```rust
fn main() {
    let _a = foo(&bar(), &mut qux(), corge());
}

#[allow(unused_mut)]
fn /*caret*/foo(i: &i32, j: &mut i32, mut k: i32) -> i32 {
    return *i + *j + k + bar();
}

fn bar() -> i32 {
    let baz = 5;
    return baz;
}

fn qux() -> i32 {
    let quux = 5;
    return quux;
}

fn corge() -> i32 {
    let grault = 5;
    return grault;
}
```

Kobzol: Try to add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` above the test function. I think that it needs stdlib to see macros that implement `Copy` for primitive types.

SaarYogev: It doesn't help unfortunately, both because I'm using *Inject Language* so it already knows it's *Rust*, and because I tried it all in a scratch file first, so there it's pure *Rust*. I guess that it's a bug in the plugin.

Kobzol: I checked out your branch and the test passes, so it's OK. This code also works in the plugin if you use it inside some project.

SaarYogev: Did you use in the tests what I commited or the code I supplied above? because I only committed it now, the test doesn't pass on my machine, and if it didn't work in a scratch file it shouldn't work in a project either (unless cargo related).

Kobzol: Sorry, didn't see the last commit. But still, the code works inside a project. It needs access to a project and to stdlib to infer `Copy` properly. Inside injected Rust strings and scratch files it does not have access to a project, that's why it shows the `Cannot move` false positive.

image

SaarYogev: That's good, means the code itself, doesn't need change any more.
I also see from searching that `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` is the standard way to introduce `stdlib` inside, any idea why it's not working?

Kobzol: If the problem that you're trying the solve is that there are red annotations inside the injected Rust test code, then the project descriptor will not solve it. But it's not a problem by itself, such annotations are present in a lot of tests, it's completely fine. Do you have any other problems with this test?

SaarYogev: No, I just assumed there is a way to solve it 😅

SaarYogev:
@Kobzol All fixed, now the test both passes `rustc` and the test itself, and the change also revealed a gap in my type check regarding unary expressions, which now I strip of their unary operator, as should be, and then checking the inner expression.

SaarYogev:
@Kobzol Anything left for me to do here? If not I'll start working on #6003 #6005 in the meantime😄.

Kobzol:
Could you add tests for some more complex patterns? For example
```
fn foo((a, b): (u32, u32)) { ... }
```

SaarYogev:
I wish I could do that, each implementation of `RsPat` might require a different way of getting variable name, mutibility, etc.
I thought of handling this using a strategy for each special `RsPat` case (and one for the default case), but it seems overkill and not going well with the general code style in the repo.

Kobzol:
Hmm, it doesn't seem that bad to me to create a `when` statement for all of them. Actually if we won't check existing variable names in this PR, you can probably just copy the `text` of the parameter binding?

@vlad20012 WDYT?

SaarYogev:
@Kobzol There's still the issue of literal/non-literal values and, and I made a different function for each. The reason is that literal values (e.g. `foo(6);`, or `let a = 6; foo(a);`) don't need to be extracted to a *let declaration*, and it looks silly if they do (and makes our inline less concise):
image

Kobzol:
I suppose that we could check if the binding of the local variable is exactly the same as the binding of the parameter (they can differ in slight details, for example mutability) and remove the new `let` declaration in such a case.

But I don't see this as a big problem, first the refactoring should support all possible cases, and then we can start focusing on generating prettier code.

SaarYogev:
It would also happen in the first case that I described:
```rust
fn main() {
    foo(6);
}

fn /*caret*/foo(bar: i32) {
    println!("{}", bar);
}
```
our expected would be `println~("{}", 6);`, but actual will be `let bar = 6; println~("{}", bar);`, which isn't needed as `6` can't change value like a function call.

As for not choosing names as best as possible to make it more generic, it's not exactly cosmetic, as I use this to split into 2 variables in the tuple deconstruction. However, it is using `parameter.text`, which is what you suggested for making it generic. The issue is that I don't know if that would work for every single type.

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -968,6 +968,60 @@ class RsInlineFunctionTest : RsTestBase() {
...
+            }
+
+            fn bar() -> S {
+                S{ a:5, b:5 }
+            }
+
+            struct S { a: i32, b: i32 }
+            """, """
+            fn main() {
+                let _ = bar();
```
Kobzol, line 1012: IMO this should create
```rust
 fn main() {
      let S { a, b } = bar();
      let _c = a + b + bar().a;
}
```

SaarYogev: agree, fixed.

SaarYogev:
@Kobzol Fixed most things, taking the "variable name" in the parameter as the variable name in the new *let* declaration works surprisingly well 😄 . 
I do have one problematic thing left, related to not adding unnecessary `let` declarations - 
`test inline function with differently named function parameters with inlined references`
Eexpected:
```rust
fn main() {
    let b = 6;
    let a = b + bar(&b);
}
```
Actual:
```rust
fn main() {
    let b = 6;
    let i = b;
    let a = i + bar(&i);
}
```
It adds a new let declaration (`let i`) because he doesn't if `6` is a unambiguous expression, or one that can change with each call.
The previous solution was this if statement:
`if (argument is RsPathExpr || argument is RsLitExpr || argument is RsUnaryExpr || argument is RsPatConst)`
But it wasn't generic, so I removed it as agreed.
Is there a generic way (either property/mehtod based or type-based) to know if an expression is unambiguous (won't change if referenced to twice, by reference or by just copying)?

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
```diff
@@ -209,7 +217,50 @@ class RsInlineFunctionProcessor(
...
+            .filterIsInstance(RsPat::class.java)
+
+        val valueArgumentList = when (caller) {
+            is RsCallExpr -> caller.valueArgumentList
+            is RsDotExpr -> caller.methodCall?.valueArgumentList ?: return
+            else -> return
+        }
+        for ((index, parameter) in parameterBindingsList.withIndex()) {
+            var argument = valueArgumentList.exprList[index]
+            if (argument is RsUnaryExpr) {
```
Kobzol, line 233: Why is this here? The following case produced wrong output:
```rust
fn foo(x: bool) {
    println!("{}", x);
}
fn main() {
    foo/*caret*/(!true);
}
```

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
```diff
@@ -209,7 +217,50 @@ class RsInlineFunctionProcessor(
...
+                argument = argument.expr
+            }
+            replaceParameterWithLetDeclaration(parameter, argument, caller)
+        }
+    }
+
+    private fun replaceParameterWithLetDeclaration(parameter: RsPat, argument: RsExpr, caller: PsiElement) {
+        val name: String = parameter.text.removePrefix("mut ")
+        if (name != argument.text) {
```
Kobzol, line 242: What does this condition do? Seems that it's not required. Argument probably won't be equal to e.g. `a: i32`.

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
```diff
@@ -263,6 +314,26 @@ class RsInlineFunctionProcessor(
         }
     }
 
+    private fun replaceParamWithExpr(param: RsPat, argument: RsExpr, funcScope: LocalSearchScope) {
```
Kobzol, line 317: It seems that this functions is unused?

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
```diff
@@ -209,7 +217,50 @@ class RsInlineFunctionProcessor(
...
+            .map { it.pat }
+            .filterIsInstance(RsPat::class.java)
+
+        val valueArgumentList = when (caller) {
+            is RsCallExpr -> caller.valueArgumentList
+            is RsDotExpr -> caller.methodCall?.valueArgumentList ?: return
+            else -> return
+        }
+        for ((index, parameter) in parameterBindingsList.withIndex()) {
+            var argument = valueArgumentList.exprList[index]
```
Kobzol, line 232: If there is a function call without enough arguments (e.g. `fn foo(a: u32)` and you call with `foo()`), the refactoring shows a warning (conflict dialog). The user can proceed with the warning however and in that case this code would crash. Assume that there may not be enough arguments and do not use `[]` indexing here.

Kobzol (changes requested):
I would keep that for future PR, let's just always create the `let` declaration in this PR.

Kobzol:
Btw `self` replacement does not seem to work for some cases:
```rust
struct S { a: u32, b: u32 }

impl S {
    fn foo(&self, a: u32) {
        println!("{} {}", self.a, a);
    }
}

fn main() {
    let s = S { a: 0, b: 0 };
    S::foo(&s, 1);
}
```
Maybe we could remove it and use your code for replacing both normal argument and `&self`?

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: It is equal in about half the tests, that *if* statement prevents us from turning this
```rust
fn main() {
    let bar = 10i32;
    foo(bar);
}
```
into this
```rust
fn main() {
    let bar = 10i32;
    let bar = bar;
    println!("{}", bar);
}
```
and creating a *let* declaration where the exact same one exists.

Kobzol: Sorry, I didn't notice that it compares only to the name of the parameter. Let's keep it then and see if it works in the wild.

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: It isn't used, since we decided to always make a *let* declaration, and will be used again if we won't create it when unecessary besides the obvious `name != argument.text` ([explained here](https://github.com/intellij-rust/intellij-rust/pull/6004#discussion_r512216550)) that's already covered. I can remove it for now too if you want.

Kobzol: Let's remove it for this PR then :)

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: Didn't see a safe way to return (`[]`, `get()`, and `elementAt()` all throw exceptions), so I created an *if* statement that returns if `index` surpasses the list's size.

Kobzol: You can use [this](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-or-null.html) :)

SaarYogev: Yeah, I know about this and `getOrElse()`, they're just delaying the problem a single line, to then *return* when we see a *null*.
I prefer to *return* just as I see that  there aren't enough arguments.

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: Done

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
Kobzol: You could use something like `var argument = valueArgumentList.exprList.getOrNull(index) ?: return` or just return before the cycle, since the sizes are known beforehand, but that's a detail :)

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: It's supposed to remove `mut` and `&` (mut is handled seperately if needed, & is automatically dereferenced). I'll fix the if statement accordigly. It helps about 10 tests reach the state where `name == argument.text` and there's not *let* decleration creation.

SaarYogev: Fixed, now it it will only shorten the argument if all leafs are `&` or `mut`.

Kobzol: It still breaks basics scenarios:
```rust
fn foo(x: &i32) {
    println!("{}", *x);
}
fn main() {
    let a = 5;
    foo/*caret*/(&a);
}
```

You can't just arbitrarily remove unary expressions, therefore I suggest to remove this code entirely :)

SaarYogev:
@Kobzol Checked it now, it happens when trying to use `::` syntax with a method (with `self` parameter). It doesn't realise the first argument is the same as `self` supplied from a `foo.bar()` syntax. Working on  a fix now.

SaarYogev:
@Kobzol Fixed every requested change you had + your comment with the `::`, all is now tested and passes.
The last thing I pushed was handling *let declarations* with only *ref*, like you said [here](https://github.com/intellij-rust/intellij-rust/pull/6004#discussion_r513264458). 
For that I created code that renames the created *let declaration*,as we can't have stuff like `let vec = &vec;`, using the same name twice. The fixing code is quite messy currently, as renaming had to occur in the end, forcing me to inline all my previous changes this PR into a single function. I'll work on refactoring that back on somehing sensible now, while you can still see if the logic is good to merge.

vlad20012:
Cool, it works! I'd refactored the code a bit, but it works fine and especially much better than before, so let's just merge it as is. Thanks!

@SaarYogev could you, please, squash these commits?

vlad20012:
:+1: 
bors r+

vlad20012:
bors retry

PR #6005, opened 2020-08-26
inline function from different module

SaarYogev:
Currently it searches for imports that are children of the file, and deletes them if they match the assumed use speck textuallty. It should instead delete based on `usageReference`, which doesn't work right now.

changelog: Fix inlining of functions that reside in a different module.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
@@ -14,6 +14,7 @@ import com.intellij.psi.codeStyle.CodeStyleManager
 import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.psi.search.LocalSearchScope
 import com.intellij.psi.search.searches.ReferencesSearch
+import com.intellij.psi.util.anyDescendantOfType
 import com.intellij.refactoring.BaseRefactoringProcessor
 import com.intellij.usageView.UsageInfo
 import com.intellij.usageView.UsageViewDescriptor
@@ -183,6 +184,9 @@ class RsInlineFunctionProcessor(
 
             val selfParam = functionDup.selfParameter
             replaceSelfParamWithExpr(selfParam, caller, funcScope)
+            if (removeDefinition) {
+                removeFunctionUseStatements()
+            }
 
             replaceParameterUsages(functionDup, caller, funcScope)
 
@@ -224,6 +228,40 @@ class RsInlineFunctionProcessor(
         } catch (e: IncorrectOperationException) {
             logger.error(e)
         }
+    }
+
+    private fun removeFunctionUseStatements() {
+        usagesAsReference.filter {
+            it.element.ancestorOrSelf<RsUseSpeck>() != null
+        }.forEach {
+            val useGroup = getLowestGroupWithMultipleChildren(it)
+            val useGroupChildrenSize = useGroup?.useSpeckList?.size
+
+            if (useGroupChildrenSize != null && useGroupChildrenSize > 1) {
+                val useSpeck = it.element.ancestorOrSelf<RsUseSpeck>()
+                val extendedSpeck = getExtendedUseSpeck(useGroup, useSpeck)
+                extendedSpeck?.deleteWithSurroundingComma()
+            } else {
+                it.element.ancestorOrSelf<RsUseItem>()?.delete()
+            }
+        }
+    }
+
+    private fun getExtendedUseSpeck(element: PsiElement, useSpeck: RsUseSpeck?): RsUseSpeck? {
+        return when (element) {
+            is RsUseGroup -> element.children.map { getExtendedUseSpeck(it, useSpeck) }
+                .firstOrNull { it != null }
+            is RsUseSpeck -> if (element == useSpeck || element.anyDescendantOfType<RsUseSpeck> { it == useSpeck }) {
+                element
+            } else {
+                null
+            }
+            else -> null
+        }
+    }
+
+    private fun getLowestGroupWithMultipleChildren(it: PsiReference): RsUseGroup? {
+        return it.element.ancestors.filterIsInstance<RsUseGroup>().firstOrNull { it.children.size > 1 }
     }
 
     private fun replaceParameterUsages(functionDup: RsFunction, caller: PsiElement, funcScope: LocalSearchScope) {
--- a/src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
+++ b/src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
@@ -835,241 +835,311 @@ class RsInlineFunctionTest : RsTestBase() {
     """)
 
     fun `test inline function with differently named function parameters`() = doTest("""
-            fn main() {
-                let a = foo(bar());
-            }
-            fn /*caret*/foo(i: i32) -> i32 {
-                return i + bar();
-            }
+        fn main() {
+            let a = foo(bar());
+        }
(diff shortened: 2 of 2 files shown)


Kobzol:
Currently, usages that are not part of a function/method call are removed from the main usages array and put inside `usagesAsReferences`. If `removeDefinition` is true, you should go through these usages in `performRefactoring` and if you find a usage inside a use speck, delete it.

Btw in the tests you can modify `doTest` like this (add `trimIndent`):
```kotlin
    private fun doTest(@Language("Rust") code: String,
                       @Language("Rust") excepted: String) {
        checkByText(code.trimIndent(), excepted.trimIndent()) {
            myFixture.performEditorAction("Inline")
        }
    }
```
and then remove the trailing spaces from the previous tests, so that indentation works properly. I've already done this change here: https://github.com/intellij-rust/intellij-rust/pull/5459/files#diff-011c6f51c6636b754ab56d51140e85a5

SaarYogev:
@Kobzol Seems ready to me, I handle both single use statements and simple use groups. Used `usagesAsReferences` as you suggested 😄.

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -1071,6 +1071,72 @@ class RsInlineFunctionTest : RsTestBase() {
 
             """)
 
+    @Language("Rust")
```
Kobzol, line 1074: `@Language("Rust")` is not needed here (same in the second test).

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -1071,6 +1071,72 @@ class RsInlineFunctionTest : RsTestBase() {
...
+    fun `test inline function from different module with two single use statements`() = doTest("""
+            use foo::bar::test;
+            use foo::bar::test2;
+            fn main() {
+                test();
+                test2();
+            }
+            mod foo {
+                mod bar {
```
Kobzol, line 1083: Consider using `pub mod bar` so that the example compiles.

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
```diff
@@ -226,6 +229,32 @@ class RsInlineFunctionProcessor(
         }
     }
 
+    private fun removeFunctionUseStatements(caller: PsiElement, reference: RsFunction) {
+        usagesAsReference.filter {
+            it.element.ancestorOrSelf<RsUseSpeck>() != null }.forEach {
+            val useGroup = it.element.ancestorOrSelf<RsUseGroup>()
+            val useGroupChildrenSize = useGroup?.useSpeckList?.size
+
+            if (useGroupChildrenSize != null && useGroupChildrenSize > 1) {
```
Kobzol, line 238: Just checking the nearest group is not enough, there may be other use groups above it and in that case they would be deleted. Consider something like this:
```rust
use foo::{bar::{test}, bar::test2};
fn main() {
    test();
}
mod foo {
    mod bar {
        pub fn /*caret*/test() {
            println!("test");
        }
        pub fn test2() {
            println!("test2");
        }
    }
}
```
It will remove too many imports. You should also probably check if the group is a sole child of a use item before deleting the whole use item.

Kobzol (changes requested):

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: Fixed. Quoting the extended solution from my commit message:

> Added handling of multiple levels of use groups.
> 
> Now stopping on the first one that has multiple uses inside it, and deleting from it. Only if all groups have a single use ?> statement inside it will delete the whole use statement.

SaarYogev:
@Kobzol All requsted changes fixed 😄 .

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
```diff
@@ -226,6 +230,51 @@ class RsInlineFunctionProcessor(
...
+                val speckIndexInGroupIncludingLeaves = childrenWithLeaves.indexOf(extendedSpeck)
+
+                // Remove comma if needed
+                if (speckIndexInGroup != 0) {
+                    childrenWithLeaves.take(speckIndexInGroupIncludingLeaves).last { groupChild -> groupChild.text == "," }.delete()
+                } else {
+                    childrenWithLeaves.drop(speckIndexInGroupIncludingLeaves).first { groupChild -> groupChild.text == "," }.delete()
+                }
+
+                extendedSpeck?.delete()
```
Kobzol, line 254: I think that you can use `extendedSpeck?.deleteWithSurroundingComma()` instead.

Kobzol (approved):
LGTM. @vlad20012 could you take a look?

src/main/kotlin/org/rust/ide/refactoring/inlineFunction/RsInlineFunctionProcessor.kt
SaarYogev: Totally works, didn't know about this method. Cleaned up accrodingly 😄 .

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -1071,6 +1071,72 @@ class RsInlineFunctionTest : RsTestBase() {
 
             """)
 
+    fun `test inline function from different module with two single use statements`() = doTest("""
+            use foo::bar::test;
```
vlad20012, line 1075: Let's keep 4 spaces style for the test code. Like
fun `test foo`() = doTest("""
    use ...;
""")

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -1071,6 +1071,72 @@ class RsInlineFunctionTest : RsTestBase() {
...
+            mod foo {
+                pub mod bar {
+                    pub fn /*caret*/test() {
+                        println!("test");
+                    }
+                    pub fn test2() {
+                        println!("test2");
+                    }
+                }
+            }""", """
```
vlad20012, line 1090: Let's move quotes on the new line (just like in other tests)
```suggestion
        }
    """, """
```

src/test/kotlin/org/rust/ide/refactoring/RsInlineFunctionTest.kt
```diff
@@ -1071,6 +1071,72 @@ class RsInlineFunctionTest : RsTestBase() {
...
+                println!("test");
+                test2();
+            }
+            mod foo {
+                pub mod bar {
+                    pub fn test2() {
+                        println!("test2");
+                    }
+                }
+            }""")
```
vlad20012, line 1102: Let's move quotes on the new line (just like in other tests)
```suggestion
        }
    """)
```

vlad20012 (approved):
LGTM. Please, squash your commits (if needed)

SaarYogev:
@vlad20012 Done, fixed indentations etc. like you asked then squashed.

Kobzol:
@SaarYogev Just a last fix, please reword the commit message of the squashed commit (it retained the message of the last commit).

SaarYogev:
@Kobzol Fixed commit message, also added the tag prefix as recommended in CONTRIBUTING.md.

Kobzol:
Great, thank you!

bors r=vlad20012

PR #6013, opened 2020-08-27
Don't run inspections on files outside a valid project

ortem:
Currently, our inspections run on any Rust files, including cases when Rust toolchain is not configured, or invalid or when a file does not belong to a Cargo project.

Now most of the inspections analyze only files that:
1. Have configured and valid toolchain
2. Belong to Cargo project and with workspace
3. Are included in the module tree

Special cases:
* Lightweight inspections (like syntax-only or naming lints) does not require (1)-(3)
* `RsDetachedFileInspection` does not require (3)

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsDetachedFileInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsDetachedFileInspection.kt
@@ -62,6 +62,8 @@ class RsDetachedFileInspection : RsLocalInspectionTool() {
         return null
     }
 
+    override val isSyntaxOnly: Boolean = true
+
     private fun isInspectionEnabled(project: Project, file: VirtualFile): Boolean =
         !PropertiesComponent.getInstance(project).getBoolean(file.disablingKey, false)
 
--- a/src/main/kotlin/org/rust/ide/inspections/RsLocalInspectionTool.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsLocalInspectionTool.kt
@@ -12,12 +12,13 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiElementVisitor
 import com.intellij.psi.PsiFile
 import org.rust.ide.utils.isEnabledByCfg
+import org.rust.lang.core.psi.RsVisitor
 
 abstract class RsLocalInspectionTool : LocalInspectionTool() {
     final override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor =
         buildVisitor(RsProblemsHolder(holder), isOnTheFly) ?: super.buildVisitor(holder, isOnTheFly)
 
-    open fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor? = null
+    open fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): RsVisitor? = null
 }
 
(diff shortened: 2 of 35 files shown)


Undin (commented):
I'm not sure that we want apply this change for all our inspections
It make sense when inspection interacts with project structure, stdlib, etc.
But there is a lot of very local inspections. For example, `RsWhileTrueLoopInspection` just checks `while (true)` constructions and suggest to replace it with `loop`

ortem:
@Undin I've changed the implementation to be more flexible and generic, what do you think?

src/main/kotlin/org/rust/ide/inspections/RsDetachedFileInspection.kt
```diff
@@ -62,6 +62,12 @@ class RsDetachedFileInspection : RsLocalInspectionTool() {
         return null
     }
 
+    override val requirements: Set<InspectionRequirement> = setOf(
+        InspectionRequirement.Toolchain,
```
Kobzol, line 66: Is it needed to distinguish between project and workspace here, i.e. is there any inspection that needs that distinction?

ortem: That's a good question, I'm not sure about this separation too and just left the most general implementation for this review. Let's wait for a review from @vlad20012

Undin: I agree with @Kobzol and think that we should merge `Project` and `Workspace` because I hardly can come up with situation when these requirements may be different

vlad20012:
Hmm, looks like requirements are not additive but inherent. I.e. a fale can't have a crate root without a workspace, can't have workspace without a cargo project, and can't have cargo project without a toolchain...

src/main/kotlin/org/rust/ide/inspections/RsDetachedFileInspection.kt
ortem: Thanks, I got rid of `Project` requirement

ortem:
@vlad20012 @Undin  Do you think there should be just one requirement per inspection, but not a set of them?

ortem:
@Undin Should I continue working on this PR and change something? Do you think it is useful or rather unnecessary?

Undin:
@ortem I suppose this API is useful but looks like we don't need so complicated requirements. Maybe it's worth just using binary differentiation: inspections that require a fully configured project model (toolchain + workspace + crate root) or just syntax

ortem:
@Undin I agree with you. Please review the new implementation based on the binary differentiation

src/main/kotlin/org/rust/ide/inspections/RsLocalInspectionTool.kt
```diff
@@ -8,16 +8,52 @@ package org.rust.ide.inspections
...
+     * - are included in module tree, i.e. have a crate root
+     * - belong to a project with a configured and valid Rust toolchain
+     */
+    private fun isApplicableTo(file: RsFile): Boolean {
+        if (isUnitTestMode) return true
+        if (isSyntaxOnly) return true
+
+        return file.cargoWorkspace != null
+            && file.crateRoot != null
+            && file.project.toolchain?.looksLikeValidToolchain() == true
```
vlad20012, line 55:
```suggestion
        return file.crate != null
            && file.project.toolchain?.looksLikeValidToolchain() == true
```

vlad20012 (approved):
bors r+

PR #6017, opened 2020-08-28
GRD: use kotlin 1.4

Undin:
After these changes it will be possible to use new kotlin language features, like [SAM conversions for kotlin interface](https://kotlinlang.org/docs/reference/whatsnew14.html#sam-conversions-for-kotlin-interfaces) or [Using break and continue inside when expressions included in loops](https://kotlinlang.org/docs/reference/whatsnew14.html#using-break-and-continue-inside-when-expressions-included-in-loops).
See full list of language improvements in https://kotlinlang.org/docs/reference/whatsnew14.html#language-features-and-improvements

Note, we still can't use new stdlib API because the plugin should be compatible with kotlin runtime bundled in 2020.1

Also, update gradle to 6.6.1


Undin:
I had to slightly modify some our code because of [regression](https://youtrack.jetbrains.com/issue/KT-41492) type inference of kotlin 1.4

vlad20012:
I'd want to highlight some new features that are useful for us (in addition to features listed in the PR description):
- [Mixing named and positional arguments](https://kotlinlang.org/docs/reference/whatsnew14.html#mixing-named-and-positional-arguments)
- [Trailing commas](https://kotlinlang.org/docs/reference/whatsnew14.html#trailing-comma) in argument and parameter lists, when entries, and components of destructuring declarations
- [Coroutine Debugger](https://kotlinlang.org/docs/reference/whatsnew14.html#coroutine-debugger) - well, we have one suspendable function, so the debugger can be useful (at least for me)

Kobzol:
Is it just me or could this regress compilation times somehow? After I change a single line of code, it now takes almost a minute to recompile and run a test, before it was ~10 seconds.

vlad20012:
@Kobzol I run into this sometimes, but that was before 1.4. Downgrading Kotlin plugin version in build.gradle.kts usually help (`kotlin("jvm") version`). But some time later the problem appears again, then **upgrading** (sic!) the Kotlin plugin back helps again. :woman_shrugging:

PR #6018, opened 2020-08-28
Enable commit message inspections

ortem:

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -3,6 +3,8 @@
   <component name="CommitMessageInspectionProfile">
     <profile version="1.0">
       <inspection_tool class="GrazieCommit" enabled="true" level="TYPO" enabled_by_default="true" />
+      <inspection_tool class="SubjectBodySeparation" enabled="true" level="WARNING" enabled_by_default="true" />
+      <inspection_tool class="SubjectLimit" enabled="true" level="WARNING" enabled_by_default="true"/>
     </profile>
   </component>
   <component name="IssueNavigationConfiguration">
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -170,7 +170,7 @@ change. Common tags are:
   * DOC for documentation
 </details>
 
-Try to keep the summary line of a commit message under 50 characters.
+Try to keep the summary line of a commit message under 72 characters.
 
 # Project structure
 


.idea/vcs.xml
```diff
@@ -2,7 +2,10 @@
 <project version="4">
   <component name="CommitMessageInspectionProfile">
     <profile version="1.0">
+      <inspection_tool class="BodyLimit" enabled="true" level="WARNING" enabled_by_default="true" />
       <inspection_tool class="GrazieCommit" enabled="true" level="TYPO" enabled_by_default="true" />
+      <inspection_tool class="SubjectBodySeparation" enabled="true" level="WARNING" enabled_by_default="true" />
+      <inspection_tool class="SubjectLimit" enabled="true" level="WARNING" enabled_by_default="true" />
```
ortem, line 8: We can also set 50 characters subject limit by adding `<option name="RIGHT_MARGIN" value="50" />` here, but it is quite restrictive IMO, and GitHub uses 72 characters limit AFAIK

Undin (commented):
I personally think that we shouldn't apply such strict inspections. In fact, everything was ok for me even without any commit message inspection

@vlad20012 @mchernyavsky @dima74 @avrong @Kobzol what do you think?

Kobzol:
I personally write relatively short commit messages (just the header basically). Even though the commit message might contain useful information, I take the corresponding GitHub PRs as the "source-of-truth", because their discussion sometimes leads to a different conclusion than the original message and it would be annoying for me to modify it, especially if it was long.

I wouldn't mind using this check, but I also didn't notice any problems with commit style/size in this repo, so I don't consider it necessary.

vlad20012:
I think 72 symbols header is reasonable because Github does something weird when a header is longer

avrong:
In the web interface, GitHub shows a tip to keep the commit subject under 50 characters. Also, it will automatically truncate commit if it will be longer than 72. For me, 72 characters warning seems ok to match GitHub logic.



I usually ignore commit body (idk if it is a bad practice), but I don't see a reason to enforce line wrap at 72/80 characters. I don't think that there is need to limit the line because of the historical terminal size.

Undin (commented):
Looks like that majority is agreed with 72 symbols restriction for commit header text. At the same time, I don't see strong pros for body restriction. So let's drop `BodyLimit` inspection from this PR and merge it

Undin (approved):
bors r+

PR #6019, opened 2020-08-28
INT: add support for guessing return type in CreateFunctionIntention

Kobzol:
This PR improves the `CreateFunctionIntention` introduced in https://github.com/intellij-rust/intellij-rust/pull/5505 to guess the expected return type of the created function. A lot of edge cases aren't handled (otherwise this could get quite complex), but I think that's fine, as long as the most obvious situations are working.

This functionality was mentioned in https://github.com/intellij-rust/intellij-rust/issues/3904.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
@@ -15,6 +15,10 @@ import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.utils.GenericConstraints
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.types.expectedType
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyUnit
+import org.rust.lang.core.types.ty.TyUnknown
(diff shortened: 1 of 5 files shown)


Undin:
@Kobzol we don't use `improvements` label for PRs. We use only `feature`, `fix`, `performance` and `internal` ones to put the corresponding item in changelog in correct section. See the last paragraph in [release section](https://github.com/intellij-rust/intellij-rust/blob/master/MAINTAINING.md#Releases) for more details

Kobzol:
Sorry, my bad.

vlad20012:
I see you're afraid to touch the type inference code 😅
Let's just enable `expected type` inference for call expressions, see the patch:

[expected_type.patch.zip](https://github.com/intellij-rust/intellij-rust/files/5145119/expected_type.zip)

It looks simpler, less hacky, and more powerful. 
One test now fails: `do not guess return type self parameter`, but I think the test is not needed now because a return type for `self` works well.

Kobzol:
:D Generally I tend to use the existing core infrastructure without modifying it, but here I see that the fix was rather simple. I'll try it next time! Thank you for the patch.

vlad20012:
:+1: 
bors r+

vlad20012:
bors retry

PR #6024, opened 2020-08-28
INSP: don't apply RsStaticConstNamingInspection for extern statics

### LINKED ISSUE #6023: RsStaticConstNaming shouldn't inspect extern statics
## Problem description
I don't believe this is intentional, but there's an inspection on the naming of extern statics. You don't exactly get to decide their name (especially if it's C) while this implies that you should change it as if that wouldn't prevent the linker from making it work, I don't believe you should be the one getting the warning.

## Example
image

Undin:
Fixes #6023


Kobzol (approved):
LGTM

Undin:
bors r=Kobzol

vlad20012:
Should associated constants be checked? Looks like this PR exclude them

Kobzol:
AFAIK this is only about statics, and these cannot be associated. Constants are checked by a different lint, because they use a different lint message.

Undin:
> Should associated constants be checked? Looks like this PR exclude them

Yep, it only about `static`s

Issue #6030, opened 2020-08-30
"Substitute associated type" intention works for regular type aliases


vlad20012:
## Environment

* **IntelliJ Rust plugin version:** 0.3.130.3315-202-nightly
* **Rust toolchain version:** 1.46.0 (04488afe3 2020-08-24) x86_64-unknown-linux-gnu
* **IDE name and version:** IntelliJ IDEA 2020.2.1 Ultimate Edition (IU-202.6948.69)
* **Operating system:** Linux 5.4.0-42-generic

## Steps to reproduce
```rust
type T = i32;

fn main() {
    let a: T = 1;
}
```

![image](https://user-images.githubusercontent.com/3221931/91660007-0ffa7b80-eadc-11ea-87ed-e85bb627347e.png)

Possible fixes:
1. Restrict it to associated types only
2. Rename it to something like `Substitute type alias` because the intention is also good in this.

Found the bug during @Kobzol's [blog post](https://kobzol.github.io/rust/intellij/2020/08/25/contributing-2-subst-assoc-type-int.html) reading.

Kobzol:
After I finished writing the post, I realized this and also that the intention is pretty broken for type aliases :D I tried to fix it, but it would require support of aliases for non-adt types to work properly. So for now I'll just disable it for type aliases.

Kobzol:
@vlad20012 btw, how difficult would it be to support type aliases properly also for non-ADT types? Could I try to with some mentoring or is too complex?

vlad20012:
Well, it would be complex only for numeric types (i.e. `type T = i32;`), but only because of our singleton-based design of numeric types. For other types it looks simple: `TyAdt.withAlias` and `TyAdt.aliasedBy` should just be propagated to the base `Ty` class (with default implementations returning `null` and `self` respectively) and then some boilerplate code should be added each `Ty*` class (`override val aliasedBy` and `withAlias` returning `copy(aliasedBy = aliasedBy)`). Sounds relatively simple... if we ignore numeric types 😅

Kobzol:
@vlad20012 The problem with type singletons could be solved by inverting the API:
```rust
type Alias = u32;
```
(`u32.aliasedBy: Alias` would change to `Alias.aliases: u32`). But I suppose that would require manual check of the `aliases` property in pretty much all places that use types?

vlad20012:
I think I don't understand your proposal... Could you please explain in more detail?

[...]

PR #6032, opened 2020-08-30
INT: disable SubstituteAssociatedTypeIntention for type aliases

Kobzol:
`SubstituteAssociatedTypeIntention` has also worked for type aliases. It is OK in basic cases, but for generic type aliases (which cannot occur currently for associated types on stable - this is the infamous GAT feature) it doesn't work properly. There are some issues with propagating type arguments and also with handling the alias itself - it's only properly supported for ADTs currently. So for now I think that we should disable this intention for type aliases. 

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6030

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/SubstituteAssociatedTypeIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/SubstituteAssociatedTypeIntention.kt
@@ -16,7 +16,9 @@ import org.rust.lang.core.psi.RsPath
 import org.rust.lang.core.psi.RsPsiFactory
 import org.rust.lang.core.psi.RsTypeAlias
 import org.rust.lang.core.psi.RsTypeReference
+import org.rust.lang.core.psi.ext.RsAbstractableOwner
 import org.rust.lang.core.psi.ext.endOffsetInParent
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/intentions/SubstituteAssociatedTypeIntention.kt
```diff
@@ -29,6 +31,7 @@ class SubstituteAssociatedTypeIntention : RsElementBaseIntentionAction<Substitut
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
         val path = element.parentOfType<RsPath>() ?: return null
         val typeAlias = path.reference?.resolve() as? RsTypeAlias ?: return null
+        if (typeAlias.owner.javaClass !in ALLOWED_OWNERS) return null
```
vlad20012, line 34: Maybe just
```suggestion
        if (typeAlias.owner.isImplOrTrait) return null
```
😅

Kobzol: *facepalm*. Thanks :)

PR #6033, opened 2020-08-31
RES: do not resolve std macros when no_std is present

Kobzol:
Previously, stdlib macros like `println` were successfully resolved even if `#![no_std]` was present, this PR changes that.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5431

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -501,6 +501,7 @@ private fun processUnqualifiedPathResolveVariants(
 
     val containingMod = path.containingMod
     val crateRoot = containingMod.crateRoot
+
     /** Path starts with `::` */
     val hasColonColon = path.hasColonColon
     if (!hasColonColon) {
@@ -831,7 +832,9 @@ fun processMacroCallVariantsInScope(context: PsiElement, processor: RsResolvePro
     val result = MacroResolver.processMacrosInLexicalOrderUpward(context, processor)
     if (result) return true
 
-    val prelude = context.contextOrSelf<RsElement>()?.findDependencyCrateRoot(STD) ?: return false
+    val element = context.contextOrSelf<RsElement>() ?: return false
+    val crateRoot = element.crateRoot as? RsFile ?: return false
+    val prelude = implicitStdlibCrate(crateRoot)?.crateRoot ?: return false
     return processAllScopeEntries(exportedMacrosAsScopeEntries(prelude), processor)
 }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -826,7 +827,10 @@ fun processMacroCallVariantsInScope(context: PsiElement, processor: RsResolvePro
     val result = MacroResolver.processMacrosInLexicalOrderUpward(context, processor)
     if (result) return true
 
-    val prelude = context.contextOrSelf<RsElement>()?.findDependencyCrateRoot(STD) ?: return false
+    val element = context.contextOrSelf<RsElement>() ?: return false
+    if (element.crateRoot?.stdlibAttributes == NO_STD) return false
```
Undin, line 831: What's about macros from `core` crate of stdlib?

Kobzol: Right, the code didn't handle `core` prelude before, so I thought that it has no macros, but they are actually all re-exported from `std`. Now if `#![no_std]` is present, `core` prelude should be looked up instead.

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -30,6 +30,7 @@ import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.cargo.util.AutoInjectedCrates.CORE
 import org.rust.cargo.util.AutoInjectedCrates.STD
 import org.rust.ide.injected.isDoctestInjection
+import org.rust.ide.utils.import.stdlibAttributes
```
Undin, line 33: nit: unused

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -826,7 +827,9 @@ fun processMacroCallVariantsInScope(context: PsiElement, processor: RsResolvePro
     val result = MacroResolver.processMacrosInLexicalOrderUpward(context, processor)
     if (result) return true
 
-    val prelude = context.contextOrSelf<RsElement>()?.findDependencyCrateRoot(STD) ?: return false
+    val element = context.contextOrSelf<RsElement>() ?: return false
+    val crateRoot = element.crateRoot as? RsFile ?: return false
+    val prelude = implicitStdlibCrate(crateRoot)?.crateRoot ?: return false
```
Undin, line 832: Looks like we don't need `element.crateRoot` here. You can pass any `RsFile` to `implicitStdlibCrate` directly. So I think we need something like 
```suggestion
    val file = context.contextualFile as? RsFile ?: return false
    val prelude = implicitStdlibCrate(file)?.crateRoot ?: return false
```

Undin (approved):
LGTM except minor comments
Thanks!

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
Kobzol: Hmm, I'm not sure if that would work. If `lib.rs` contains `#![no_std]`, we want to stop resolving macros and `std` in all files imported by `lib.rs`, right? Using `contextualFile` would only stop `std` macro resolving if the file where the macro is used has `#![no_std]`.

Undin: Yep. You are right.
Interesting fact that `implicitStdlibCrate` was written by me :sweat_smile:

Undin (approved):
bors r+

Kobzol:
bors retry

PR #6037, opened 2020-08-31
REF: generate setter action

Kobzol:
This PR adds an action to generate a setter for a struct, mirroring the recently added generate getter action (https://github.com/intellij-rust/intellij-rust/pull/5352).

setter

It also introduces several improvements like better copy type handling, support for type aliases, finding existing impl blocks and navigation to the generated method. Since these are directly applicable to both getters and setters, I included them in this PR, but if you want, I can split them into a separate PR.

Currently, if the action is invoked inside an `impl`, I filter out attributes that would conflict with an existing method:
```rust
struct S { a: u32 }
impl S {
   fn a(&self) {}
   /*caret*/ <- a is not offered, because `a` method exists
}
```
However, if the action is invoked on the struct, all (non-pub) fields are offered, and then if it finds an existing `impl` block with a method with the same name, it will generate the getter/setter and the names will conflict. I think that this behaviour is fine, just wanted to mention it.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/generate/BaseGenerateHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/generate/BaseGenerateHandler.kt
@@ -13,7 +13,6 @@ import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiFile
-import org.rust.ide.refactoring.generate.getter.GenerateGetterHandler
 import org.rust.lang.core.psi.RsBaseType
 import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.RsPsiFactory
--- a/src/main/kotlin/org/rust/ide/refactoring/generate/GenerateAccessorHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/generate/GenerateAccessorHandler.kt
@@ -5,11 +5,14 @@
 
 package org.rust.ide.refactoring.generate
 
+import com.intellij.openapi.editor.Editor
+import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.RsStructItem
 import org.rust.lang.core.psi.ext.RsVisibility
 import org.rust.lang.core.psi.ext.expandedMembers
 import org.rust.lang.core.psi.ext.isTupleStruct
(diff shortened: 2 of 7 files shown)


src/main/kotlin/org/rust/ide/refactoring/generate/BaseGenerateHandler.kt
```diff
@@ -95,8 +97,16 @@ abstract class BaseGenerateHandler : LanguageCodeInsightActionHandler {
         struct: RsStructItem
     ): RsImplItem {
         return if (implBlock == null) {
-            val impl = psiFactory.createInherentImplItem(structName, struct.typeParameterList, struct.whereClause)
-            struct.parent.addAfter(impl, struct) as RsImplItem
+            val existingImpl = struct.implLookup.findImplsAndTraits(struct.declaredType)
+                .filterIsInstance<TraitImplSource.ExplicitImpl>()
+                .filter { it.isInherent }
+                .firstOrNull()
```
vlad20012, line 103: This `firstOrNull` looks very suspicious. A struct can have multiple inherent impls. Or impls with different type bounds... There can be one another dialog that allows to choose the impl (or choose to create a new one). But a new dialog overloads the UI, so such a solution is also very discussable. I'd prefer to generate a new impl over this design.

vlad20012 (commented):
LGTM except one commit

src/main/kotlin/org/rust/ide/refactoring/generate/BaseGenerateHandler.kt
Kobzol: Oh right, I didn't think of impls with different type bounds. Ok, let's ditch finding existing impls. Should I also remove it in https://github.com/intellij-rust/intellij-rust/pull/6039?

vlad20012: Yep, I just reviewed #6039 and was just about to write a comment about it =D

vlad20012:
:+1: 
bors r+

vlad20012:
Oh, you forgot to remove `test find existing impl block` test.
bors r-

Kobzol:
bors r=vlad20012

vlad20012:
I'll cherry-pick it into `release-130` because it mostly consists of fixes and improvements for #5352 that is a part of `release-130`

MarinaKalashina:
Docs done (mentioned as an example for the Generate menu) https://plugins.jetbrains.com/plugin/8182-rust/docs/rust-code-generation.html#generate-menu

PR #6038, opened 2020-08-31
REF: handle struct field shorthand init in RsInlineValue

Kobzol:
This PR fixes `RsInlineValue` so that it properly supports shorthand field init. Inspired by https://github.com/rust-analyzer/rust-analyzer/pull/5888.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/inlineValue/RsInlineValueProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/inlineValue/RsInlineValueProcessor.kt
@@ -13,6 +13,8 @@ import com.intellij.refactoring.BaseRefactoringProcessor
 import com.intellij.usageView.UsageInfo
 import com.intellij.usageView.UsageViewDescriptor
 import org.rust.ide.refactoring.RsInlineUsageViewDescriptor
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.RsStructLiteralField
 import org.rust.lang.core.resolve.ref.RsReference
 
 class RsInlineValueProcessor(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/refactoring/inlineValue/RsInlineValueProcessor.kt
```diff
@@ -33,9 +35,20 @@ class RsInlineValueProcessor(
...
 
     override fun performRefactoring(usages: Array<out UsageInfo>) {
+        val factory = RsPsiFactory(project)
         usages.asIterable().forEach loop@{
             val reference = it.reference as? RsReference ?: return@loop
-            reference.element.replace(context.expr)
+            when (val element = reference.element) {
+                is RsStructLiteralField -> {
+                    if (element.colon == null) {
+                        element.addAfter(factory.createColon(), element.identifier)
```
vlad20012, line 44: `element.identifier` can be null in the case of tuple struct fat initialization `S { 0: 0, 1: 1}`, how it would behave in this case? I suggest this:
```suggestion
                        element.addAfter(factory.createColon(), element.referenceNameElement)
```

vlad20012 (approved):

src/main/kotlin/org/rust/ide/refactoring/inlineValue/RsInlineValueProcessor.kt
Kobzol: Oh, I had no idea you can initialize tuple structs like that. I added a test for it, but it works for the case that you have mentioned, because in that case the reference element is not `RsStructLiteralField`, but `RsPath`, so it works. And you cannot have shorthand initializer for tuple structs, so it should be fine.

Kobzol:
bors r=vlad20012

PR #6039, opened 2020-08-31
INT: add support for creating methods in CreateFunctionIntention

Kobzol:
This PR adds support for creatng methods in `CreateFunctionIntention`. Also I previously forget to register the intention and add a description, since it was originally a quick fix, so this PR also fixes that.

create-method

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
@@ -15,10 +15,10 @@ import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.utils.GenericConstraints
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.TraitImplSource
 import org.rust.lang.core.types.expectedType
-import org.rust.lang.core.types.ty.Ty
-import org.rust.lang.core.types.ty.TyUnit
-import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.implLookup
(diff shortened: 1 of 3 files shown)


vlad20012:
> Also I previously forget to register the intention and add a description, since it was originally a quick fix, so this PR also fixes that.

😄 we must add some check to `RsIntentionTestBase`

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -15,123 +15,198 @@ import org.rust.ide.presentation.renderInsertionSafe
...
+            if (!functionCall.expr.isAncestorOf(path)) return null
+
+            val module = getTargetModuleForFunction(path) ?: return null
+            val name = path.referenceName
+            text = "Create function `$name`"
+            return Context.Function(functionCall, name, module)
+        }
+        val methodCall = element.parentOfType<RsMethodCall>()
+        if (methodCall != null) {
+            if (methodCall.reference.resolve() != null) return null
```
vlad20012, line 69: Same as in function call case, let's limit possible caret position to the method identifier. `element == methodCall.identifier` should works

Kobzol:
>  we must add some check to `RsIntentionTestBase`

Ooh, I actually thought that it exists, I even wrote so on my blog :D But it only exists for inspections. I'll add it for intentions too.

Undin (changes requested):
@Kobzol Let's split this PR into two: the first to fix intention registration and the second one to add new feature. It allows us to cherry-pick only fix into v130 release, put new feature into v131 (or later if something will happen) and don't change our usual workflow

vlad20012 (approved):
:+1: 
Rebase it to the latest master, please

PR #6047, opened 2020-09-01
TY: relax `expectedType` fallback

vlad20012:
This should not affect users anyhow, it's an internal change,
it will be used later in completion etc.

Now we fallback `expectedType` to its origin `TyTypeParameter`
instead of `TyUnknown`. It works for this:
```rust
fn foo<T>(t: T) {}
fn main() {
    foo(foobar/*expectedType = T*/); // previously it was TyUnknown
}
```


vlad20012:
bors retry

Issue #5973, opened 2020-08-20
Run tests with coverage results in "the linked panic runtime `panic_unwind` is not compiled with this crate's panic strategy `abort`" error


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.128.3278-202
* **Rust toolchain version:** 1.47.0-nightly (792c645ca 2020-08-17) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.2 (CL-202.6397.106)
* **Operating system:** macOS 10.15.6

## Problem description

Running tests with basic Run works fine, but Run with coverage fails with the following error

```
Testing started at 22:11 ...
/Users/jetbrains/.cargo/bin/cargo +nightly test --color=always --all-features --no-fail-fast -- -Z unstable-options --format=json --show-output
   Compiling proc-macro2 v1.0.19
   Compiling unicode-xid v0.2.1
   Compiling syn v1.0.38
   Compiling quote v1.0.7
   Compiling flow_impl_derive v0.10.0 (/Users/jetbrains/git/flow/flow_impl_derive)
error: the linked panic runtime `panic_unwind` is not compiled with this crate's panic strategy `abort`

error: aborting due to previous error

error: could not compile `flow_impl_derive`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed

Process finished with exit code 101
```

## Steps to reproduce

1. Open https://github.com/andrewdavidmackenzie/flow/tree/master/flow_impl_derive
2. Run `test --all-features` cargo command with coverage

lancelote:
Based on a user report in support.

andrewdavidmackenzie:
I doubt it maters but just FYI - in my original report it was found in 
```
IntelliJ IDEA 2020.2 (Ultimate Edition)
Build #IU-202.6397.94, built on July 27, 2020
Licensed to Red Hat UK Limited / Andrew Mackenzie
Subscription is active until July 25, 2021
Runtime version: 11.0.7+10-b944.20 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 10.15.3
GC: ParNew, ConcurrentMarkSweep
Memory: 2014M
```

but it seems you were able to reproduce it in CLion!

mchernyavsky:
[...]

PR #6048, opened 2020-09-01
RUN: Don't use `-Cpanic=abort` when `Run with Coverage`

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5973.

### DIFF
--- a/coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
+++ b/coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
@@ -86,8 +86,7 @@ class GrcovRunner : RsDefaultProgramRunnerBase() {
             val environmentVariables = EnvironmentVariablesData.create(
                 oldVariables.envs + mapOf(
                     "CARGO_INCREMENTAL" to "0",
-                    "RUSTFLAGS" to "-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort",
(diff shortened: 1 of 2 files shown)


coverage/src/main/kotlin/org/rust/coverage/GrcovRunner.kt
```diff
@@ -86,8 +86,7 @@ class GrcovRunner : RsDefaultProgramRunnerBase() {
             val environmentVariables = EnvironmentVariablesData.create(
                 oldVariables.envs + mapOf(
                     "CARGO_INCREMENTAL" to "0",
-                    "RUSTFLAGS" to "-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort",
-                    "RUSTDOCFLAGS" to "-Cpanic=abort"
+                    "RUSTFLAGS" to "-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off"
```
Undin, line 89: How does it affect users except fix proc-macro lib tests?

mchernyavsky: Yes, now user can catch panic. Also, previously we incorrectly showed coverage in some cases:

* before:
 

* after:

Undin: Thanks!

Undin (approved):
bors r+

PR #6051, opened 2020-09-02
CI: set PR milestone automatically

Undin:
We use milestones to mark pull requests that are included into the corresponding release.
We use it:
- for changelog template creation
- to get our users full list of release changes
- while regression investigation in published release builds

But we have to set milestones manually. As a result, a developer can forget to do it (and it actually happens).
These changes provides new GitHub workflow to set current milestone to each merged PR automatically if it doesn't have it yet

### DIFF
--- a/scripts/github.py
+++ b/scripts/github.py
@@ -1,0 +1,21 @@
+# TODO: think about using library for GitHub API
+import json
+from typing import Dict
+from urllib.request import Request, urlopen
+
+from common import get_patch_version
+
+
+def get_current_milestone(repo: str) -> Dict:
+    response = urlopen(f"https://api.github.com/repos/{repo}/milestones")
+    milestones = json.load(response)
+    milestone_version = f"v{get_patch_version()}"
+    return next(milestone for milestone in milestones if milestone["title"] == milestone_version)
+
+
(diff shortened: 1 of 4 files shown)


scripts/github.py
```diff
@@ -0,0 +1,23 @@
+# TODO: think about using library for GitHub API
+import json
+from urllib.request import Request, urlopen
+
+from common import get_patch_version
+
+
+def get_current_milestone(repo: str):
```
Undin, line 8: @lancelote what type should I use here? `Any`?

Undin: Debugger says that it's `Dict`

lancelote:
bors merge

Undin:
Found out that it doesn't work for external pull requests because the corresponding github token doesn't have [`write` permissions](https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token). At the same time, we can't pass custom token via secrets because they [are not passed](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) into external pull requests

PR #6053, opened 2020-09-02
ANN: Support suppress file level inspection annotations

### LINKED ISSUE #5976: Provide a way to suppress file level inspection annotations
## Environment

* **IntelliJ Rust plugin version:** 0.3.130.3298-202-nightly
* **Rust toolchain version:** 1.47.0-nightly (9b88e0a86 2020-08-15) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2.1 Preview Ultimate Edition (IU-202.6948.5)

## Problem description
Currently, if a user wants to suppress a false-positive file-level inspection annotation, the plugin doesn't allow them to do it.
Yeah, if a caret points on some item in code, annotation provides a suppress quick fix to suppress it with comment. But it doesn't work because the quick fix add a comment to the pointed item and it doesn't affect inspection itself since a comment belongs to the item, not to file itself

image
image

## Steps to reproduce
Code example is taken from [rocket](https://github.com/SergioBenitez/Rocket/blob/02b79e28f71ccbd1fce0e369469c92caedb38cd8/examples/hello_world/src/main.rs) repo

Stzx:
Closed #5976

- [x] Support suppress file level inspection annotations
- [ ] Support quick fix


Stzx:
After some attempts, I couldn't find a way to directly provide quick fix(Must rely on existing elements to provide). Can anyone provide some help?

PR #6054, opened 2020-09-02
RUN: Add target completion for cargo command

avrong:
Adds a `--target` suggestion to cargo command run configuration and run anything. After the option argument itself, suggests a list of installed target triples from `rustup target list`.


avrong:
Closing this in favor of #6071

PR #6055, opened 2020-09-02
RUN: Add all-targets flag in cargo command completion

### LINKED ISSUE #6034: Missing flag for cargo check
## Environment

* **IntelliJ Rust plugin version:** 0.3.129.3308-202
* **Rust toolchain version:** 1.46
* **IDE name and version:** CLion 2020.2

## Problem description

When specifying command flags when editing run configuration for `cargo check`,  the `--all-targets` flag is missing from autocompletion:

image

Also, it would be nice to have a dedicated `check` button in toolbar, if possible.

## Steps to reproduce

Add a `check` command and try to add given flag.

avrong:
Fixes #6034

### DIFF
--- a/src/main/kotlin/org/rust/cargo/util/CargoCommandCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/cargo/util/CargoCommandCompletionProvider.kt
@@ -160,6 +160,8 @@ class CargoCommandCompletionProvider(
 
             targetBench()
             flag("bench")
+
+            flag("all-targets")
         }
 
         fun pkg() = opt("package") { ctx ->


src/main/kotlin/org/rust/cargo/util/CargoCommandCompletionProvider.kt
```diff
@@ -160,6 +160,8 @@ class CargoCommandCompletionProvider(
 
             targetBench()
             flag("bench")
+
+            flag("all-targets")
```
mchernyavsky, line 164: The `targetAll` is used in `doc` and `install` commands, which don't have `--all-targets` flag.

PR #6056, opened 2020-09-02
RUN: Fix benches flag in cargo command completion

avrong:
According to the [cargo docs](https://doc.rust-lang.org/cargo/commands/cargo-check.html) it should have a `--bench` option, which we have with `targetBench()` to use a specific bench target and `--benches` to use all bench targets.


mchernyavsky (approved):
bors r+

PR #6057, opened 2020-09-03
INT: introduce quick fix to add `..` to struct pattern that has missing fields

### LINKED ISSUE #6022: Suggest subslice pattern quick-fix for non-exaustive pattern matching expressions
[Subslice patterns](https://rust-lang.github.io/rfcs/2707-dotdot-patterns.html) which are [stabilized in Rust 1.42](https://github.com/rust-lang/rust/pull/67712) are a good way to fix the pattern matchings where not all the fields are specified by user.

From a user's point of view, it would be useful to write the ones that are needed and then, apply a quick-fix to ignore the others as an option instead of introducing new variables via `Add missing fields` and dealing with them being unused.

t-kameyama:
Fixes #6022


mchernyavsky (approved):
Thanks!

bors r+

PR #6059, opened 2020-09-03
PRJ: Use Task.Modal and ProcessAdapter in component downloading modals

### LINKED ISSUE #5847: Invoke Task.Modal instead of Task.Backgroundable when downloading rust-src in new project wizard
Similar to cargo-generate installation (#5715) there is no need to have the ability to run the task in the background.

Also, despite the task is getting finished quite fast, it would be good to have additional information about the process and the error pop-up when the installation fails.

Comparison below.

avrong:
Fixes #5847

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
+++ b/src/main/kotlin/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.kt
@@ -5,10 +5,13 @@
 
 package org.rust.cargo.project.settings.ui
 
+import com.intellij.execution.process.ProcessAdapter
+import com.intellij.execution.process.ProcessEvent
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.options.ConfigurationException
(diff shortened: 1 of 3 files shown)


ortem:
Note that the link to download `rust-src` component is available not only in `New Project` dialog but in `Preferences | Languages & Frameworks | Rust` as well. 

Before PR:
download_src_old

After PR:
download_src_new

Is this the desired behavior?

avrong:
With its speed, I reckon it is ok for now. `Task.Modal` will avoid the modal going into nowhere with `Background` action in the new project wizard. Also, as I understand, when the plugin loads a Cargo project, it automatically downloads `rust-src` anyway.

Otherwise, it would be needed to maintain two implementations, one `Backgroundable` and another just `Modal`.

In the future, if we could have several toolchains available, maybe with `add component` UI something like in IDEA SDKs could be used. It automatically becomes background when the window is closed:

ortem (changes requested):
Please rebase onto master

ortem (approved):
bors r+

PR #6065, opened 2020-09-04
T: highlight all unresolved references in real project tests

Undin:
Previously, real project tests checked only unresolved reference annotations visible for user by default.
It should help us to detect errors in name resolution/type inference


Undin:
bors retry

PR #6069, opened 2020-09-04
CI: cache evcxr in check workflow

Kobzol:
This PR caches `evcxr` so that it is not recompiled in every `check` workflow. This should hopefully reduce the runtime of this workflow on Ubuntu by about 2 minutes (~ 10% of the total runtime).

I also tried to cache the `Download` task, but I'm not sure what does it download. When I cached the `deps` folder, it still took a long time to run the download task (it probably also downloads some gradle stuff?).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -63,6 +63,17 @@ jobs:
                   components: rust-src, rustfmt, clippy
                   default: true
 
+            - name: Cache evcxr
+              uses: actions/cache@v2
+              with:
+                  path: |
+                      ~/.cargo/bin/evcxr
+                      ~/.cargo/.crates.toml
+                      ~/.cargo/.crates2.json
+                  key: ${{ runner.os }}-cache-evcxr-${{ matrix.rust-version }}-${{ github.run_id }} # https://github.com/actions/cache/issues/342#issuecomment-673371329
+                  restore-keys: |
+                      ${{ runner.os }}-cache-evcxr-${{ matrix.rust-version }}-
+
             - name: Install evcxr
               if: matrix.os != 'windows-latest' && matrix.rust-version >= '1.40.0' # BACKCOMPAT: Rust 1.39.0
               uses: actions-rs/cargo@v1


.github/workflows/check.yml
```diff
@@ -63,8 +63,15 @@ jobs:
                   components: rust-src, rustfmt, clippy
                   default: true
 
+            - name: Cache evcxr
+              id: cache-evcxr
+              uses: actions/cache@v2
+              with:
+                  path: ~/.cargo/bin/evcxr
+                  key: ${{ runner.os }}-cache-evcxr-${{ matrix.rust-version }}
```
Undin, line 71: Looks like it can prevent updating `evcxr` to new version. Currently, we don't care about it and uses latest version. But if we cache some version of `evcxr`, CI will use cache and potentially will never user a new version.
Probably, we should specify version of `evcxr` for cargo (via `--version` flag according to [documentation](https://doc.rust-lang.org/cargo/commands/cargo-install.html)) and use it in key

Undin (commented):
> This PR caches evcxr so that it is not recompiled in every check workflow. This should hopefully reduce the runtime of this workflow on Ubuntu by about 2 minutes (~ 10% of the total runtime).

JFTR it wasn't done before because `check` run on Windows takes more time even without `evcxr` compilation so total time of all checks will be the same. But of course, if we have several pull requests at the same time, it can speed up at least one of thier check

> I also tried to cache the Download task, but I'm not sure what does it download. When I cached the deps folder, it still took a long time to run the download task (it probably also downloads some gradle stuff?).

I tried to do it to and it didn't work as expected. Firstly, total cache size is too small (5 GB according to [docs](https://github.com/actions/cache#cache-limits)) to cache both IDEA and CLion dependencies for at least two different platforms (201 and 202 currently) for both operating systems (Linux and Windows). Secondly, decompression time took similar time as downloading. So for now, IDE dependencies caching doesn't look promising, unfortunately

.github/workflows/check.yml
Kobzol: The cache depends on the `rustc` version, so it will be updated from time to time. But it's true that we could specify it explicitly, although I'm thinking that it will get relatively complex for a change that doesn't speed up the bottleneck (windows).

.github/workflows/check.yml
```diff
@@ -63,8 +63,15 @@ jobs:
                   components: rust-src, rustfmt, clippy
                   default: true
 
+            - name: Cache evcxr
+              id: cache-evcxr
+              uses: actions/cache@v2
+              with:
+                  path: ~/.cargo/bin/evcxr
```
dima74, line 70: Probably we should cache whole `~/.cargo`, or, at least `~/.cargo/.crates.toml` and `~/.cargo/.crates2.json` - because we use `cargo install --list` command to determine whether evcxr is installed

Kobzol: I tried to cache `~/.cargo` originally, but for some reason it got overwritten or it didn't work at all (the loaded cache was empty). But I didn't investigate further, I can try.

.github/workflows/check.yml
dima74: I think the best approach would be:
* Restore evcxr from cache
* Always run `cargo --install evcxr` (if `evcxr` from cache has latest version, it will not be recompiled)
* Always save cache (that is override existing cache)

Unfortunatelly it is not possible now to [update cache](https://github.com/actions/cache/issues/342). We can try to use [a workaround](https://github.com/actions/cache/issues/342#issuecomment-673371329)

Kobzol: I tried your suggestion, it seems to work.

Kobzol:
By the way, can the plugin tests run in parallel? (or do they already run in parallel?)

Undin:
> By the way, can the plugin tests run in parallel? (or do they already run in parallel?)

@Kobzol do you mean something like [this](https://guides.gradle.org/performance/#parallel_test_execution) or just split tests into several bunches and launch them on different agents?
If you mean the first option, we didn't try it (as far as I know) but looks like it shouldn't improve something because GitHub action agent has only 2 hardware cores. If you mean the second option then, unfortunately, we can't because there is GitHub limit in 20 parallel execution and currently we have 17 parallel checks only for `check` workflow
So we can try `maxParallelForks` option but I doubt that it decreases execution time

Kobzol:
The second option would IMO be too complex to be worth it, even if we didn't hit GitHub limits. I meant the first one, I'll try `maxParallelForks=2`.

Undin (approved):
Great! Thanks!
bors r+

Issue #6058, opened 2020-09-03
Add an Gradle task to update cargo options in `Run anything` and cargo command completion provider


mchernyavsky:

avrong:
I was thinking about this recently.

What if we could make a full-featured DSL for the commands and options description, and then regenerate it from time to time using [cargo commands docs](https://github.com/rust-lang/cargo/tree/master/src/doc/man). It already has some kind of grouping using templates and special tags for the options.

Docs also have full multi-sentence descriptions, but for most of the cases, I think we could just take the first one for the short description in `Run Anything`.

For the special cases like `--bin`, `--test` or `--target`  where we can also provide a suggestion of option depending on the context, there can be another custom completion that will work simultaneously with the generated one.

PR #6071, opened 2020-09-05
Add `updateCargoOptions` task to update cargo options automatically

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/6058.
Fixes https://github.com/intellij-rust/intellij-rust/issues/6034.
Fixes https://github.com/intellij-rust/intellij-rust/issues/6096.

Closes https://github.com/intellij-rust/intellij-rust/pull/6055.


Kobzol:
Btw, it's not strictly relevant to this PR, but a similar task for downloading `rustc` features (`updateCompilerFeatures`) seems to be broken, because the URLs are no longer invalid.

Undin:
@Kobzol thanks!
It was broken by recent changes in rustc project structure
Will be fixed in #6077

Undin (commented):
As we discussed in person, let's consider retrieving info from `cargo` itself instead of site. It looks like a more stable solution. Also, the corresponding parsing can be potentially used to generate the same info dynamically in the plugin itlsef

Undin (approved):
Awesome!

Possible future improvements: 
- currently, we understand that some option is already written and don't suggest it. For example,  for `cargo check --jobs 3` we don't suggest `--jobs` option. But if we use short variant of this option like `-j`, our completion doesn't understand it and suggests `--jobs`. Looks like we can retrieve such info as well
- fetch options of `cargo` itself

Undin:
Note, turned out that parsing of `cargo` output is quite hard for now so we decided to use info from `https://doc.rust-lang.org/cargo/commands` page

mchernyavsky (commented):
bors r+

PR #6075, opened 2020-09-06
Fix project under symlinked directory

### LINKED ISSUE #5861: Possible regression in release 127 - rust edition not detected, file not include in module tree
## Environment

* **IntelliJ Rust plugin version:** Release 127 -- also tried 0.3.128.3262-202-nightly
* **Rust toolchain version:** 1.45
* **IDE name and version:** IntelliJ Community 2020.2 (was also the case on 2020.1)
* **Operating system:** Kubuntu Linux 19.10

## Problem description

After updating the plugin from 126 -> 127 my project had lots of various errors. I have tried reimporting the project, re-loading the Cargo.toml from the "Cargo" menu. In Release 126, I was using the "experimental macro engine" and I didn't have any errors across my project. 

I don't really understand which of the following is a root problem, so I will just post some of the issues to give a picture of what I'm experiencing.

## Edition is not being correctly detected and `async` keyword is disallowed 
"This feature is only available in Edition 2018"

image

## Source files are not being included in the module tree
"File is not included in module tree"

image

## Modules declared at the top level are not found
"File not found for module"

__see first screenshot__

The one thing that makes this project different is that it is using rust-protobuf to do codegen and is including the generated code at the top of main. This means the project not only has a `build.rs`, but also an `include!` macro in `main.rs`.

I downgraded the plugin to 126 in 2020.1 and the project started building properly again. I can't use plugin 126 in 2020.2. **Is there a way I can get a build (or build myself) a version of 126 that I can use in 2020.2 while we work this out?**

I did try suggestions in #3628 and #4907, but have not been able to find a resolution.

Thanks for your help, and thanks for an awesome plugin. :) 

## Steps to reproduce

Source code for project is not open source, so this may be difficult.

### LINKED ISSUE #5962: 'File is not included in module tree' - again
## Environment

* **IntelliJ Rust plugin version:** latest stable && nightly
* **Rust toolchain version:** 1.47.0-nightly
* **IDE name and version:** CLion 2020.2
* **Operating system:** Linux

## Problem description
In a large project with a workspace and multiple crates, I get the warning
`File is not included in module tree, analysis is not available` for every file in the project.

I tried invalidating the cache and restarting etc. but that did not help. I figured out that the latest `0.2` version just before the jump `0.3` works and analyzes the files correctly. However, the `0.2` version doesn't happen to have the fix that's required for the current nightly version, as paths changed. So either I'm missing the `stdlib` sysroot with the old verison, or have non-analyzed files with the `latest stable` or `nightly` plugin version.

## Steps to reproduce
`git clone https://github.com/tokio-rs/tokio`
open the project in CLion.

vlad20012:
The bug was introduced in #5690 release-127

Fixes #5861, fixes #5962

_Probably we need a test for symlinked project_

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/crate/impl/CrateGraphServiceImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/crate/impl/CrateGraphServiceImpl.kt
@@ -27,6 +27,7 @@ import org.rust.lang.core.crate.CrateGraphService
 import org.rust.lang.core.crate.CratePersistentId
 import org.rust.openapiext.CachedValueDelegate
 import org.rust.openapiext.checkReadAccessAllowed
+import org.rust.stdext.applyWithSymlink
 import org.rust.stdext.enumSetOf
 import org.rust.stdext.exhaustive
 import java.nio.file.Path
@@ -61,7 +62,7 @@ class CrateGraphServiceImpl(val project: Project) : CrateGraphService {
 
     override fun findCrateByRootMod(rootModFile: VirtualFile): Crate? {
         checkReadAccessAllowed()
-        return if (rootModFile is VirtualFileWithId) findCrateById(rootModFile.id) else null
+        return rootModFile.applyWithSymlink { if (it is VirtualFileWithId) findCrateById(it.id) else null }
     }
 
 }


vlad20012:
bors retry

Issue #4000, opened 2019-06-10
Auto-import adds private module


mchernyavsky:
## Environment

* **Intellij-Rust plugin version:** v0.2.100.2656-191-nightly
* **Rust toolchain version:** nightly-x86_64-apple-darwin
* **IDE name and version:** CL-191.7479.33
* **Operating system:** macOS 10.14.5

## Problem description

Before auto-import:
```rust
fn main() {
    N/*caret*/;
}
```

After auto-import:
```rust
use core::unicode::tables::general_category::N; // Module `core::unicode::tables` is private [E0603]

fn main() {
    N/*caret*/;
}
```

Kobzol:
@Undin this specific automatic import now works correctly (it imports `use core::unicode::N;`). Should this issue be closed or is it more general?

Undin:
@Kobzol I suppose it's because of changes in stdlib not because of some changes in auto import.
But with latest nightly, it shouldn't suggest anything to import since `core::unicode::N` has `pub (crate)` visibility.
So I've updated the issue description and want to keep it open

PR #6078, opened 2020-09-07
REF: Add explicit visibility check in move refactoring

dima74:
[`RsImportHelper` can find private path](https://github.com/intellij-rust/intellij-rust/issues/4000), so we have to explicitly check accessibility of such paths


src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -217,6 +221,7 @@ fun <T : RsElement> movedElementsDeepDescendantsOfType(
  */
 fun RsVisRestriction.updateScopeIfNecessary(psiFactory: RsPsiFactory, newParent: RsMod) {
     if (crateRoot == newParent.crateRoot) {
+        if (path.textNormalized == "self") return  // `pub(self)`
```
ortem, line 224: Maybe use `PathKind.SELF` instead of string literal?

ortem (approved):
LGTM except one comment

dima74:
bors r=ortem

dima74:
bors ping

dima74:
bors retry

PR #6080, opened 2020-09-07
REF: Use same processor for move files and move items refactoring

dima74:
This pull request is expected to not have any user-visible changes, but is needed to:
* Easily implement move for multiple files (common processor is already able to process multiple items)
* Any future change to common processor will affect both move files and move items refactoring (so no code duplication, etc)


ortem (approved):
LGTM. Probably it'd be better to extract all changes in tests to a separate commit if it's just a cleanup of the tests

dima74:
I would say changes to tests are behaviour changes (in particular now [`keep existing reference style`](https://github.com/intellij-rust/intellij-rust/pull/5688) works in move file refactoring), so lets keep them in one commit.

bors r=ortem

Issue #6098, opened 2020-09-10
CreateFunctionIntention is missing title


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.131.3342-202-nightly
* **Rust toolchain version:** 1.45.0-nightly (a74d1862d 2020-05-14) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3 Snapshot (CL-203.3482)
* **Operating system:** macOS 10.15.6

## Problem description

Spotted during #6050 testing. `CreateFunctionIntention` is missing title in the intention list:

<img width="990" alt="Screen Shot 2020-09-10 at 15 57 06" src="https://user-images.githubusercontent.com/2500320/92731919-64281a00-f37e-11ea-85ad-4aec31196b9a.png">

PR #6102, opened 2020-09-12
INT: add missing title and fix typo in CreateFunctionIntention

Kobzol:
I also noticed that the template for this intention had a typo. I think that it's a small enough change that we can bolt it onto this PR?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6098


vlad20012:
:+1: 
bors r+

PR #5736, opened 2020-07-13
REF: Update references in move items refactoring

dima74:
Update references in [move items refactoring](https://github.com/intellij-rust/intellij-rust/issues/3531#issuecomment-643142894):
* Inside references (to moved items - both from old mod and from other mods)
* Outside references (from moved items - both to old mod and to other mods)

---

Please see comment for `RsMoveCommonProcessor` for detailed description

I will add tests later (after merging #5665), but anyway all existing tests can be found [here](https://github.com/intellij-rust/intellij-rust/blob/diralik/move/src/test/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsTest.kt)

PR #6107, opened 2020-09-14
REF: Support move multiple files

dima74:
[Move refactoring](https://github.com/intellij-rust/intellij-rust/issues/3531#issuecomment-643142894) improvements:
* Support move multiple files/directories
* Support move file to another crate

This pull requests mostly changes only move handler (it finds which exactly files should be moved). Actual implementation was added in #5736 and #6080

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/move/RsMoveFilesOrDirectoriesDialog.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/move/RsMoveFilesOrDirectoriesDialog.kt
@@ -5,6 +5,7 @@
 
 package org.rust.ide.refactoring.move
 
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.DialogWrapper
 import com.intellij.openapi.ui.Messages
@@ -22,6 +23,7 @@ import com.intellij.refactoring.util.CommonRefactoringUtil
 import com.intellij.util.IncorrectOperationException
 import org.rust.cargo.project.workspace.CargoWorkspace
 import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.ide.refactoring.move.RsMoveFilesOrDirectoriesHandler.Companion.adjustForMove
 import org.rust.lang.RsConstants
 import org.rust.lang.core.psi.RsFile
 import org.rust.lang.core.psi.ext.RsMod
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/ide/refactoring/move/RsMoveFilesOrDirectoriesDialog.kt
```diff
@@ -32,29 +34,37 @@ import org.rust.openapiext.toPsiFile
...
     override fun performMove(targetDirectory: PsiDirectory) {
         if (!CommonRefactoringUtil.checkReadOnlyStatus(project, targetDirectory)) return
-        if (!CommonRefactoringUtil.checkReadOnlyStatus(project, elementsToMove.toList(), true)) return
+        if (!CommonRefactoringUtil.checkReadOnlyStatus(project, filesOrDirectoriesToMove.toList(), true)) return
 
         try {
-            for (element in elementsToMove) {
-                if (element is RsFile) {
+            for (element in filesOrDirectoriesToMove) {
+                if (element is PsiFile) {
```
ortem, line 52: Why not `RsFile`? AFAIU we use this dialog only in case of moving `.rs` files

src/main/kotlin/org/rust/ide/refactoring/move/RsMoveFilesOrDirectoriesHandler.kt
```diff
@@ -77,19 +104,17 @@ class RsMoveFilesOrDirectoriesHandler : MoveFilesOrDirectoriesHandler() {
...
+private fun RsFile.canBeMoved(): Boolean =
+    modName != null
         && crateRoot != null
         && crateRelativePath != null
         && !isCrateRoot
         && isEdition2018  // TODO: support 2015 edition
-}
+        && pathAttribute == null  // TODO: support path attribute on mod declaration
+
+fun PsiElement.adjustForMove(): RsFile? =
```
ortem, line 115: It is not clear that **public** method `PsiElement.adjustForMove()` is only for files moving (but not just `PsiElement`s moving). Maybe we can rename this method or put it inside a companion object of some class related to files moving only (like `RsMoveFilesOrDirectoriesHandler`)?

src/test/kotlin/org/rust/ide/refactoring/move/RsMoveMultipleFilesTest.kt
```diff
@@ -0,0 +1,104 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.refactoring.move
+
+class RsMoveMultipleFilesTest : RsMoveFileTestBase() {
```
ortem, line 8: Please add a test that we do not handle moving non-Rust files along with Rust ones, like moving `foo.rs` and `Cargo.toml` together

ortem (changes requested):

src/test/kotlin/org/rust/ide/refactoring/move/RsMoveMultipleFilesTest.kt
dima74: Thanks, added (for `.txt` files)

ortem (approved):

PR #6122, opened 2020-09-16
Get rid of `Don't reformat child modules` option

vlad20012:
This option does nothing for a long time. Finally, get rid of it.

Screenshot from 2020-09-16 22-50-46


mchernyavsky (approved):
bors r+

PR #6126, opened 2020-09-17
Fix exception `PsiElementBase.notNullChild` after variable inlining

vlad20012:
RsPath can't be replaced with `RsExpr`. Instead, parent `RsPathExpr` should be replaced.

Typical stacktrace ends with this:
```
java.lang.Throwable: 10
 parent=let b = 10;
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:143)
	at com.intellij.psi.impl.PsiElementBase.notNullChild(PsiElementBase.java:284)
	at org.rust.lang.core.psi.impl.RsPathExprImpl.getPath(RsPathExprImpl.java:43)
```

Also, I enhanced tests in order to catch such bugs in the future.

c.c. @Kobzol


vlad20012:
(Thanks @lancelote for crash reports)

PR #6129, opened 2020-09-17
PSI: Make RsForeignModItem implement RsItemsOwner

dima74:
`RsForeignModItem` is naturally an item's owner, so it would be reasonable for it to implement `RsItemsOwner` interface


dima74:
@vlad20012 pointed out that `RsItemsOwner` is sometimes used in the meaning "`RsMod` or `RsBlock`" (e.g. in [`RunCargoExpandIntention`](https://github.com/intellij-rust/intellij-rust/blob/b0047b0ac3996edf22d82ce1aae988fcf0be0082/src/main/kotlin/org/rust/ide/intentions/RunCargoExpandIntention.kt#L29)), and also it is not clear why `RsForeignModItem` should be `RsItemsOwner`, and `RsMembers` not. So lets close this PR for now

PR #6143, opened 2020-09-21
GRAM: properly parse special macro calls with fqn

### LINKED ISSUE #6140: include! macro support not working with std::include! or inner std::concat!
## Environment

* **IntelliJ Rust plugin version:** 0.3.130.3337-202
* **Rust toolchain version:** 1.45.1
* **IDE name and version:** IntelliJ IDEA 2020.2.2 (Community Edition)

## Problem description
In a [blog post on Dec 4, 2019, support for the `include!` macro](https://blog.jetbrains.com/clion/2019/12/recent-updates-in-rust/#include_macro_support) including an inner `concat!` was announced. In the meanwhile also support for `include!` macro calls generated by another macro call (#5336) was added. Thanks a lot for these helpful features!

Unfortunately, the `include!` macro support (in the form or autocompletion) does not work when the macro calls are fully qualified, that is, when `std::include!` or `std::concat!` are used. In some situations, such full qualification is necessary. For example, when defining exported macros with `#[macro_export(local_inner_macros)]`.

## Steps to reproduce

hello.rs
```rust
pub fn message() -> &'static str {
    "Hello World!"
}
```

main.rs
```rust
std::include!("hello.rs"); // autocompletion does not work
// include!(std::concat!("hello.rs")); // autocompletion does not work
// include!("hello.rs"); // autocompletion works
// include!(concat!("hello.rs")); // autocompletion works
// include!(concat!("hello", ".rs")); // autocompletion works with warning "File is not included in module tree, analysis is not available" on hello.rs

fn main() {
    println!("{}", message());
}
```

Undin:
It allows us to provide the same features like highlighting of format macro, `include` macro support, etc. for macro calls with fully qualified name like `std::include`

Fixes #6140


Kobzol:
Can this help with recognizing e.g. `log::info!` so that it could be annotated with `RsFormatMacroAnnotator`?

Undin:
> Can this help with recognizing e.g. log::info! so that it could be annotated with RsFormatMacroAnnotator?

@Kobzol if the problem is in parsing then yes

src/test/resources/org/rust/lang/core/parser/fixtures/complete/macros.rs
```diff
@@ -145,5 +159,7 @@ fn foo() {
     asm!("nop", a = const A + 1);
     asm!("nop", in(reg) x => y, out("eax") _);
     asm!("nop", a = const 5, b = sym foo::bar, c = in(reg) _, d = out(reg) a => _);
+    sdt::asm!("nop"); // fully qualified macro call
```
vlad20012, line 162: `sDt`? =D

Undin: Oops)
Just typo. Although it doesn't matter here

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -434,15 +434,28 @@ object RustParserUtil : GeneratedParserUtilBase() {
...
+            val elementType = b.rawLookup(steps) ?: return null
+
+            if (!isWhitespaceOrComment(b, elementType)) {
+                meaningfulSteps--
             }
-            if (nextTokenIsExcl) b.tokenText else null
+        }
+
+        return if (b.rawLookup(steps) == IDENTIFIER) {
+            PsiBuilderUtil.rawTokenText(b, steps).toString()
```
vlad20012, line 458: Hmm, we have an extension `PsiBuilder.rawLookupText()` that does the same as `PsiBuilderUtil.rawTokenText` :thinking:

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -434,15 +434,28 @@ object RustParserUtil : GeneratedParserUtilBase() {
...
+
+        while (meaningfulSteps > nameTokenIndex) {
+            steps--
+            val elementType = b.rawLookup(steps) ?: return null
+
+            if (!isWhitespaceOrComment(b, elementType)) {
+                meaningfulSteps--
             }
-            if (nextTokenIsExcl) b.tokenText else null
+        }
```
vlad20012, line 455: Let's extract it to extension `lookBehind` (similar to existing `lookAhead`) in `org/rust/lang/core/parser/util.kt`. This method would look more clear.

Undin: Note, `lookAhead` method returns `IElementType` but in this code we need to know raw index as well to get token text. So we can't create something like `fun PsiBuilder.lookBehind(steps: Int): IElementType?` only `fun PsiBuilder.lookBehind(steps: Int): Pair<Int, IElementType>?`. Do you think it's worth it? I don't like such inconsistency

vlad20012: Indeed...

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
Undin: Let's use own `PsiBuilder.rawLookupText()` since it checks corner cases

Undin: Also, we have the same extention in toml plugin so I will move it into `common` module

Undin:
bors r=vlad20012

PR #2738, opened 2018-08-02
GRD: update kotlin and grammar kit plugin

Undin:

PR #6151, opened 2020-09-23
Add compile only dependency on kotlin stdlib

dima74:
It allows viewing kotlin stdlib source code with comments.

It was already [added before](https://github.com/intellij-rust/intellij-rust/pull/2312/files), and then [removed](https://github.com/intellij-rust/intellij-rust/pull/2738/files#diff-392475fdf2bc320d17762ed97109a121L135) because it caused problems when running tests using JPS (instead of Gradle). I don't find any problems now when using JPS with this compile only dependency, so let's try add it again.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -122,6 +122,10 @@ allprojects {
         }
     }
 
+    dependencies {
+        compileOnly(kotlin("stdlib-jdk8"))
+    }
+
     afterEvaluate {
         tasks.withType<AbstractTestTask> {
             testLogging {


dima74:
Not sure why `regressions` workflow fails - I restarted it manually and it [succeeds](https://github.com/intellij-rust/intellij-rust/actions/runs/268843651)

Undin:
@dima74 unfortunately, some of our anotations are flaky so it can happen sometimes

build.gradle.kts
```diff
@@ -248,6 +248,7 @@ project(":") {
     val testOutput = configurations.create("testOutput")
 
     dependencies {
+        compileOnly(kotlin("stdlib-jdk8"))
```
Undin, line 251: Currently, you add this dependency only for core module. As a result, you can't navigate to sources of kotlin stdlib in, for example, `clion` module
Let's move it into `allprojects` section to add this dependency for all our modules (including future ones)

Undin (changes requested):

Undin (approved):
bors r+

Undin:
bors retry

Issue #6146, opened 2020-09-21
Configure plugin via environment variables


nmattia:
# Problem description

I would like to be able to use environment variables to configure the plugin; in particular `RUSTC`, `CARGO` and `RUST_SRC_PATH`.

I am working on a developer environment for 50+ engineers and we are not using rustup. Engineers using intellij do not want to have to specify the project settings manually. The discovery mechanism somewhat works for find the rust toolchain, but requires a workaround: creating a dummy "toolchain" directory containing `rustc` and `cargo`, and adding that directory to `$PATH`.

For rust's stdlib, I can't figure out how to get this to work. Without rustup, the plugin with look in `./lib/rustlib/src/rust`. I've tried setting `CI=true` and `RUST_SRC_WITH_SYMLINK=<path to actual src lib>` but that doesn't work.

Am I missing something re. the stdlib source directory?

vlad20012:
Hello!
To avoid [XY problem](https://en.wikipedia.org/wiki/XY_problem) I'd want to ask why the env-based configuration is simpler than GUI-based? Are these env variables used anywhere else?

Also, your case looks very interesting to us. Could you please explain in more detail, why so many people are working without rustup? 


Some random notes:

> creating a dummy "toolchain" directory containing rustc and cargo, and adding that directory to $PATH

Well, this directory can be `/bin` or a something like this.

> Without rustup, the plugin with look in ./lib/rustlib/src/rust

`$(rustc --print sysroot)/lib/rustlib/src/rust`. This is the usual location of Rust stdlib sources (even without rustup). Where you going to store these sources?

> CI, RUST_SRC_WITH_SYMLINK

These envs are used in test code only. Not in the production

nmattia:
Thanks for the quick reply!

> To avoid XY problem I'd want to ask why the env-based configuration is simpler than GUI-based? Are these env variables used anywhere else?
> Also, your case looks very interesting to us. Could you please explain in more detail, why so many people are working without rustup?

We have various compiler flags and dependencies injected through environment variables. Those variables change regularly and are pushed to all the machines (ok, pulled by). The environment variables are also used on CI, and when running e.g. `cargo` on the command line. We have ~1/5 of engineers using intellij, and they mostly use it for edition and then run `cargo` in their terminal to ensure everything works. 

> Well, this directory can be /bin or a something like this. This is pretty usual for UNIX-like systems.

[...]

PR #6156, opened 2020-09-24
Follow `rustc --print sysroot` stdlib path without Rustup

vlad20012:
Improves #3779 & #3856, related to #6146.

1. If there is no `Rustup` in the toolchain, use `$(rustc --print sysroot)/lib/rustlib/src/rust` path to stdlib even if explicit path (in the plugin settings) is set to null
2. In the plugin settings UI, when saving the settings, set explicit stdlib path to null if it equals to `$(rustc --print sysroot)/lib/rustlib/src/rust` (or if there is `Rustup` in the toolchain, as before)

So, if a user doesn't have `Rustup` and doesn't change the stdlib path in the plugin settings, the plugin will **follow** `$(rustc --print sysroot)/lib/rustlib/src/rust` path.

User-visible change: Cargo project structure loading (aka Cargo Refresh) with project opened via `Open` action successfully sets up stdlib path when `rustup` doesn't exist but `$(rustc --print sysroot)/lib/rustlib/src/rust` exists. Previously, error notification with `Cargo project update failed: no explicit stdlib or rustup found` text was shown to a user and stdlib wasn't set up

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
@@ -164,6 +164,7 @@ private fun fetchStdlib(
     val rustup = context.toolchain.rustup(workingDirectory)
     if (rustup == null) {
         val explicitPath = context.project.rustSettings.explicitPathToStdlib
+            ?: context.toolchain.getStdlibFromSysroot(workingDirectory)?.path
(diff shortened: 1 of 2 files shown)


Undin (approved):
LGTM
Let's squash commits into single one

vlad20012:
bors r=undin

vlad20012:
bors retry

PR #6157, opened 2020-09-24
STUB: Add helper methods for new resolve

dima74:
Added helper methods for stubs which will be used in new resolve.

Also:
* Added `hasColonColon` flag to `RsUseSpeckStub` to distinguish `use {aaa, bbb};` and `use ::{aaa, bbb};`
* Added `bodyHash` to macro and macro call stubs as an optimization

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -52,7 +52,7 @@ class RsFileStub : PsiFileStubImpl<RsFile> {
     override fun getType() = Type
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
-        private const val STUB_VERSION = 201
+        private const val STUB_VERSION = 202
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -1337,10 +1416,14 @@ class RsMacro2Stub(
 class RsMacroCallStub(
     parent: StubElement<*>?, elementType: IStubElementType<*, *>,
     val macroBody: String?,
+    val bodyHash: HashCode?,
     val bodyStartOffset: Int,
     override val flags: Int
 ) : RsAttributeOwnerStubBase<RsMacroCall>(parent, elementType) {
 
+    val path: RsPathStub
+        get() = findChildStubByType(RsPathStub.Type)!!
```
vlad20012, line 1425:
```suggestion
        get() = findChildStubByType(RsPathStub.Type)!! // guaranteed to be non-null by the grammar
```

vlad20012:
bors retry

PR #6163, opened 2020-09-26
INSP: improve integration with Grazie plugin

Undin:
These changes:
* Fix incorrect proofreading annotations on raw and binary string literals. The bug can be reproduced only in 2020.3
image
* Properly support domain options (`String literals`, `Comments` and `Documentation`) in `Preferences | Editor | Proofreading | Grammar`
* Provide source code of Grazie plugin while development when `baseIDE` in `gradle.properties` is `idea`. Previously, IDE always shows only class files


mchernyavsky (approved):
bors r+

PR #6166, opened 2020-09-26
UI: Use platform UI DSL

mchernyavsky:
https://github.com/intellij-rust/intellij-rust/pull/5529#discussion_r489290697

I haven't found a way to provide tooltip text, but I'm not sure we should use it, since it's quite undiscoverable. As an alternative, we can use comments, but I don't like it either:

image


avrong:
Maybe it will help if we will organize checkbox labels to the right according to the platform UI guide. [Checkbox](https://jetbrains.github.io/ui/controls/checkbox/) and [Inline help text](https://jetbrains.github.io/ui/components/inline_help_text/) sections can help.

Undin (commented):
> I haven't found a way to provide tooltip text, but I'm not sure we should use it, since it's quite undiscoverable. As an alternative, we can use comments, but I don't like it either:

You can manually create label component (i.e replace `row("Some text")` with `row(Label("Some text"))`) and set tooltip to this label component

Undin:
> Maybe it will help if we will organize checkbox labels to the right according to the platform UI guide

Yep, it's a good idea. We've just discussed similar things with @vlad20012 that we should use the recommended way to render checkbox label.

I'm not sure about inline help text because I can't imagine if it will look good or not but we definitely may try to find out it

mchernyavsky:
> You can manually create label component (i.e replace `row("Some text")` with `row(Label("Some text"))`) and set tooltip to this label component

@Undin Done.

Undin (commented):
Text fields are cropped
image

mchernyavsky:
> Text fields are cropped

@Undin I don't know the right way to fix it, but it works now.

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
```diff
@@ -39,31 +40,33 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
...
-            row("Additional arguments:", externalLinterArguments, """
+            """) { externalLinterComboBox(growX, pushX) }
+            row("Additional arguments:", """
                 Additional arguments to pass to `cargo check` or `cargo clippy` command.
-            """)
-            row("Run external linter to analyze code on the fly:", runExternalLinterOnTheFlyCheckbox, """
-                Enable external linter to add code highlighting based on the used linter result.
+            """) { externalLinterArguments(growX, pushX) }
+            row("Run external linter to analyze code on the fly:", """
+                Enable external linter to add code highlighting based on the used linter result.
```
Undin, line 67: nit: redundant space at the end of line

Undin (approved):
LGTM

mchernyavsky (commented):
bors r+

Issue #6169, opened 2020-09-28
Run with Coverage doesn't work with latest nightly


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.132.3360-202-nightly
* **Rust toolchain version:** 1.48.0-nightly (d006f5734 2020-08-28) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2.2 Ultimate Edition (IU-202.7319.50)
* **Operating system:** macOS 10.15.6
* **Grcov version:** 0.5.15

## Problem description
After execution of run with coverage run configuration, there is no user-visible result except empty coverage toolwindow
![image](https://user-images.githubusercontent.com/2539310/94409335-b2ab2600-017e-11eb-88ff-8481626145e8.png)

Note, everything works fine with `nightly-2020-07-10`

## Steps to reproduce
```rust
fn main() {
    foo();
}

fn foo() {
    println!("Hello!")
}
```

PR #6170, opened 2020-09-28
RUN: Fix `Run with Coverage` on nightly channel

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6169.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/Grcov.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/Grcov.kt
@@ -18,6 +18,7 @@ class Grcov(private val grcovExecutable: Path) {
             .withParameters(
                 ProjectLayout.target,
                 "-t", "lcov",
+                "--llvm",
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/toolchain/Grcov.kt
```diff
@@ -18,6 +18,7 @@ class Grcov(private val grcovExecutable: Path) {
             .withParameters(
                 ProjectLayout.target,
                 "-t", "lcov",
+                "--llvm",
```
Undin, line 21: Is this option a new one, i.e. does `0.4.3` version contain it? If no then we should update our minimal version in the corresponding notification

mchernyavsky: Yes, `0.4.3` contains it.

Undin (approved):
bors r+

PR #6172, opened 2020-09-28
UI: use recommended guidelines for settings UI

Undin:
These changes:
* use embedded checkbox label
* use inline text help (comments) instead of Swing tooltips
* use more features of platform UI DSl that allows dropping mutable variables to keep intermediate state, implement `reset`, `isModified` and `apply` methods almost automatically

Also, these changes introduce mutable `RustProjectSettingsService.settingsState` property to make usage of binding related UI DSL methods convenient

| Before | After |
| - | - |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
+++ b/src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
@@ -28,9 +28,8 @@ class CargoConfigurable(project: Project) : RsConfigurableBase(project, "Cargo")
         return panel {
             row {
                 checkBox(
-                    "Watch Cargo.toml",
-                    state::autoUpdateEnabled,
-                    comment = "Update project automatically if `Cargo.toml` changes"
+                    "Update project automatically if Cargo.toml changes",
+                    state::autoUpdateEnabled
                 )
             }
             row {
--- a/src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
+++ b/src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
@@ -6,95 +6,77 @@
 package org.rust.cargo.project.configurable
 
 import com.intellij.openapi.project.Project
-import com.intellij.openapi.ui.ComboBox
+import com.intellij.openapi.ui.DialogPanel
 import com.intellij.ui.EnumComboBoxModel
-import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.layout.CCFlags
 import com.intellij.ui.layout.panel
+import com.intellij.ui.layout.toBinding
 import org.rust.cargo.project.model.cargoProjects
 import org.rust.cargo.toolchain.ExternalLinter
 import org.rust.cargo.util.CargoCommandCompletionProvider
 import org.rust.cargo.util.RsCommandLineEditor
-import org.rust.openapiext.CheckboxDelegate
-import org.rust.openapiext.ComboBoxDelegate
-import org.rust.openapiext.row
-import javax.swing.JComponent
 
-class CargoConfigurable(project: Project) : RsConfigurableBase(project) {
-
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
```diff
@@ -6,95 +6,82 @@
...
+                checkBox(
+                    "Update project automatically if Cargo.toml changes",
+                    state::autoUpdateEnabled
+                )
+            }
+            row {
+                checkBox(
+                    "Compile all project targets if possible",
+                    state::compileAllTargets,
+                    comment = "Pass <b>--target-all</b> option to Сargo <b>build</b>/<b>check</b> command"
```
avrong, line 40: What if we specify that this would be applied implicitly when such commands are invoked?
Like `Pass --target-all when invoking Сargo build and check commands`

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
```diff
@@ -6,95 +6,82 @@
...
+                )
+            }
+            row {
+                checkBox(
+                    "Offline mode",
+                    state::useOffline,
+                    comment = "Pass <b>--offline</b> option to Cargo not to perform network requests"
```
avrong, line 47: As I understand reading [Cargo --offline](https://www.ncameron.org/blog/cargo-offline/) blog post, the `--offline` option only disables crates index update. Maybe clarify it here?

Like `Pass --offline option to force Cargo not to update the crates registry`

avrong:
Not sure if it is related to current PR, but would be great if such inconsistencies will be resolved during the UI migration.

image

image

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
```diff
@@ -6,95 +6,82 @@
...
-
-    override fun reset() {
-        autoUpdateEnabled = settings.autoUpdateEnabled
-        externalLinter = settings.externalLinter
-        runExternalLinterOnTheFly = settings.runExternalLinterOnTheFly
-        externalLinterArguments.text = settings.externalLinterArguments
-        compileAllTargets = settings.compileAllTargets
-        useOffline = settings.useOffline
+        super.apply()
+        project.rustSettings.settingsState = state
```
mchernyavsky, line 85: Maybe we can move this line to `CargoConfigurable` somehow? We can introduce an abstract method, something like `validateAll` or `doApply`, and call it in `CargoConfigurable`'s `apply` (which should be `final`).

Undin: What problem should this change solve?

mchernyavsky (approved):

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
Undin: I'm not sure that `crates registry` is more understandable for users (especially for newbies in the ecosystem) than `not to perform network requests`.
For example, https://doc.rust-lang.org/cargo/faq.html#how-can-cargo-work-offline says nothing about crates registry as well

src/main/kotlin/org/rust/cargo/project/configurable/CargoConfigurable.kt
Undin: Personally, I don't catch the major difference between original text and your suggestion

Undin:
> Not sure if it is related to current PR, but would be great if such inconsistencies will be resolved during the UI migration.

@avrong yep, this bug is because of platform UI DSL. Don't want to solve it in this PR so I've created a separate issue #6182

mchernyavsky (approved):
bors r+

Issue #6175, opened 2020-09-29
Show test result in the gutter


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.132.3375-203-nightly
* **Rust toolchain version:** 1.45.0-nightly (a74d1862d 2020-05-14) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3 Snapshot (CL-203.4085)
* **Operating system:** macOS 10.15.6

## Problem description

Some IDEs (e.g. WebStorm) have a nice feature of showing test result in the gutter, for example

<img width="747" alt="Screen Shot 2020-09-29 at 16 51 45" src="https://user-images.githubusercontent.com/2500320/94567460-2c710b80-0274-11eb-8e4d-d081cf9baf00.png">

I find it quite useful, would love to have a similar feature in Rust plugin.

PR #6176, opened 2020-09-29
T: Show test result in the gutter

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6175.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/test/CargoTestEventsConverter.kt
@@ -296,10 +296,10 @@ class CargoTestEventsConverter(
                 .addAttribute("nodeId", suite)
 
         private fun createTestStartedMessage(test: NodeId): ServiceMessageBuilder {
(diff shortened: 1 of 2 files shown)


mchernyavsky (commented):
bors r+

Undin (commented):
@mchernyavsky Is it possible to write tests for the corresponding feature?
Note, we already have some tests for line markers

mchernyavsky:
@Undin We already have tests for this feature (in `CargoTestRunLineMarkerContributorTest`), but they don't help really.

Undin:
@mchernyavsky but these tests are about providing some icon to run tests but they don't check a kind of icon

mchernyavsky:
@Undin see
* `ttest show a green mark for a passed test`
* `test show a red mark for a failed test`
* `test show a mark for an ignored test`
* `test show a green mark for a passed mod`
* `test show a red mark for a failed mod`
* `test show a mark for an ignored mod`

Undin:
@mchernyavsky Oh, I see.
But looks like these tests really check nothing because we manually write status into `TestStateStorage`. Do they actually make sense? Maybe it's better to run tests and check statuses after that?

PR #6177, opened 2020-09-29
GRAM: support `tuple.0.0` as tuple indexing

### LINKED ISSUE #6029: Accept tuple.0.0 as tuple indexing
https://github.com/rust-lang/rust/pull/71322
In Rust stable `1.46` it's now possible to index nested tuple fields, for example:
```rust
fn foo(a: ((i32, ), )) {
    let b = a.0.0;
}
```

The problem is that `0.0` in the example is lexed as FLOAT_LITERAL token and then the parser can't parse it. The possible solution is to remove `FLT_LITERAL` rule from the lexer and then teach the parser to construct float literal from parts.

vlad20012:
Fixes #6029

Now we can parse this:
```rust
fn foo(a: ((i32, ), )) {
    let b = a.0.0;
}
```

#### Implementation details:

I removed `FLOAT_LITERAL` rules from the lexer. Now `0.0` is lexed as 3 tokens: `INTEGER_LITERAL` `DOT` `INTEGER_LITERAL`. Then I collapse them into single `FLOAT_LITERAL` during parsing (when needed).


src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
```diff
@@ -226,6 +226,32 @@ object RustParserUtil : GeneratedParserUtilBase() {
     fun isPathMode(b: PsiBuilder, level: Int, mode: PathParsingMode): Boolean =
         mode == getPathMod(b.flags)
 
+    @JvmStatic
+    fun parseFloatLiteral(b: PsiBuilder, level: Int): Boolean {
```
Undin, line 230: Since we collapse tokens by the parser, it breaks proper highlighting of float literals by `RsHighlighter`. More precisely, the color of dot in float literals was blue before your change, now it's black

src/test/resources/org/rust/lang/core/parser/fixtures/partial/exprs.rs
```diff
@@ -9,4 +9,6 @@ fn foo() {
     // Distinguish `for` expr from `for<>` in types
     let _:
     for a in b {}
+
+    let _ = 0.0.0e92;
```
Undin, line 13: Hm, why we parse this as a float literal?
The compiler produces `error[E0610]: `{float}` is a primitive type and therefore doesn't have fields` here, parser in master shows some parser error, parser with these changes doesn't show any error

src/main/kotlin/org/rust/lang/core/parser/RustParserUtil.kt
vlad20012: Oh

src/test/resources/org/rust/lang/core/parser/fixtures/partial/exprs.rs
vlad20012: Fixed

mchernyavsky (approved):
bors r+

kumbayo:
Hi @vlad20012,
for me, this change seems to cause a problem for hexadecimal literals containing `e` or `f`.
They seem to be parsed/recognized as `f64` instead of as `u32` / integer literal.

```rust
const CONST1: u32 = 0xd0;
const CONST2: u32 = 0xe0;
const CONST3: u32 = 0xf0;

const CONST4: u32 = 0x0d;
const CONST5: u32 = 0x0e;
const CONST6: u32 = 0x0f;
```

image

After reverting the PR locally for testing, the issue is gone again.

Undin:
@kumbayo Oh, great catch! Thank you!
I've created a separate issue about it #6255

vlad20012:
@kumbayo Thanks! Luckily we haven't released it yet

PR #6178, opened 2020-09-29
INT: Use public visibility when creating function in other crate

dima74:
Use `pub` visibility instead of `pub(crate)` when creating function or method in other crate using great [CreateFunctionIntention](https://github.com/intellij-rust/intellij-rust/pull/5505)


Kobzol (approved):
LGTM.

dima74:
bors r=Kobzol

PR #6179, opened 2020-09-29
UI: Get rid of scrollbar in project creation dialog in CLion

avrong:
I reckon the scrollbar started to appear after recent UI changes. The scrollbar scrolls to nothing, but taking too much space and focus on the right of the form in default non-resizable CLion wizard window.

This PR makes templates pane height fixed to 125px which allows to show 4 default template items and 2 user ones if they have any without scrollbar in the pane itself.

| Before | After |
| ------- | ----- |
|  |  |

### DIFF
--- a/src/main/kotlin/org/rust/ide/newProject/ui/RsNewProjectPanel.kt
+++ b/src/main/kotlin/org/rust/ide/newProject/ui/RsNewProjectPanel.kt
@@ -22,6 +22,7 @@ import com.intellij.ui.ToolbarDecorator
 import com.intellij.ui.components.JBList
 import com.intellij.ui.components.Link
 import com.intellij.ui.layout.LayoutBuilder
+import com.intellij.util.ui.JBUI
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/ide/newProject/ui/RsNewProjectPanel.kt
```diff
@@ -88,6 +89,7 @@ class RsNewProjectPanel(
 
     val templateToolbar: ToolbarDecorator = ToolbarDecorator.createDecorator(templateList)
         .setToolbarPosition(ActionToolbarPosition.BOTTOM)
+        .setPreferredSize(Dimension(0, 125))
```
Undin, line 92: Let's use `com.intellij.util.ui.JBUI#size(int, int)` instead of raw `Dimension` to support IDE scaling when it's not performed by OS itself

Undin (approved):

avrong:
bors r=Undin

PR #6188, opened 2020-10-02
CI: improve presentation of regression annotations

Undin:
These changes:
- use 1-based indices for line/column in regression annotation to be compatible with editor UI and `Go to Line/Column` action
- sort regression annotations


vlad20012 (approved):
bors r+

PR #6199, opened 2020-10-05
Fix some BACKCOMPATs with 2020.1 and 2019.3

ortem:


Undin (commented):
@ortem some tests fail

Undin:
@ortem tests still fail

src/main/kotlin/org/rust/ide/annotator/RsCfgDisabledCodeAnnotator.kt
```diff
@@ -8,27 +8,23 @@ package org.rust.ide.annotator
...
-    private fun AnnotationHolder.createCfgDisabledAnnotation(textRange: TextRange) {
-        val color = RsColor.CFG_DISABLED_CODE
-        val severity = if (isUnitTestMode) color.testSeverity else HighlightSeverity.INFORMATION
-        // BACKCOMPAT: 2019.3
-        @Suppress("DEPRECATION")
-        createAnnotation(severity, textRange, "Conditionally disabled code").textAttributes = color.textAttributesKey
+            holder.newAnnotation(severity, "Conditionally disabled code")
+                .textAttributes(color.textAttributesKey)
+                .create()
+        }
```
Undin, line 28: These changes are not just using a new API, they lead to different behaviours
Old:
image

New:
image


Is it intentional?

ortem: `rangeWithSurroundingLineBreaks` was used before to handle background highlighting, but then we decided to use the foreground instead (there is a note in the commit body). So this is intentional. But actually, I hadn't thought about the color of comment highlighting. Do you think it is important?

Undin: > Do you think it is important?

No, I don't think so

Undin (approved):
bors r+

PR #6202, opened 2020-10-06
TOML: recover parser on errors in arrays

vlad20012:
Better TOML parser recovery in the case of broken array literals. Needed for completion inside arrays


vlad20012:
bors try

vlad20012:
bors try

PR #6203, opened 2020-10-06
CARGO: Replace paths to symlink source in cargo metadata

dima74:
If project directory is `source` which is a symlink to `target` directory, then `cargo metadata` command inside `source` directory produces output with `target` paths. [`CargoMetadata.Project.workspace_root`](https://github.com/intellij-rust/intellij-rust/blob/12b71cbd08c3c79282cedec71abcdd6dba61d9d5/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt#L56-L59) was already replaced to use `source` paths in #4868. This PR replaces other fields, such as [`CargoMetadata.Package.manifest_path`](https://github.com/intellij-rust/intellij-rust/blob/12b71cbd08c3c79282cedec71abcdd6dba61d9d5/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt#L85-L88), to use `source` paths.

Main purpose of this PR is to have [`Crate.rootModFile`](https://github.com/intellij-rust/intellij-rust/blob/12b71cbd08c3c79282cedec71abcdd6dba61d9d5/src/main/kotlin/org/rust/lang/core/crate/Crate.kt#L61-L65) to use `source` path (`source` paths are better because files inside `source` are indexed, and inside `target` are not).

### DIFF
--- a/src/test/kotlin/org/rustSlowTests/lang/resolve/CargoProjectResolveTest.kt
+++ b/src/test/kotlin/org/rustSlowTests/lang/resolve/CargoProjectResolveTest.kt
@@ -16,6 +16,8 @@ import org.rust.lang.core.crate.impl.CrateGraphTestmarks
 import org.rust.lang.core.psi.RsPath
 import org.rust.lang.core.resolve.NameResolutionTestmarks
 import org.rust.openapiext.pathAsPath
+import org.rustPerformanceTests.fullyRefreshDirectoryInUnitTests
+import java.nio.file.Files
 
 class CargoProjectResolveTest : RsWithToolchainTestBase() {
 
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rustSlowTests/lang/resolve/CargoProjectResolveTest.kt
```diff
@@ -141,6 +142,60 @@ class CargoProjectResolveTest : RsWithToolchainTestBase() {
         checkReferenceIsResolved<RsPath>("src/bar.rs")
     }
 
+    fun `test resolve local package under symlink`() = fileTree {
```
dima74, line 145: Ideally we also need test when project directory is under symlink, but I couldn't get it work yet:
```kotlin
    fun `test module relations when project is under symlink`() = fileTree {
        toml("Cargo.toml", """
            [package]
            name = "mods"
            version = "0.1.0"
            authors = []

            [dependencies]
        """)

        dir("src") {
            rust("lib.rs", """
                mod foo;

                pub struct S;
            """)

            rust("foo.rs", """
                use S;
                  //^
            """)
        }
    }.let { fileTree ->
        if (SystemInfo.isWindows) return@let
        val rootPath = cargoProjectDirectory.pathAsPath
        val targetName = cargoProjectDirectory.name + "_target"
        val targetPath = rootPath.parent.resolve(targetName)
        Files.move(rootPath, targetPath)
        Files.createSymbolicLink(rootPath, targetPath)

        fullyRefreshDirectoryInUnitTests(cargoProjectDirectory.parent)
        val target = cargoProjectDirectory.parent.findChild(targetName)!!
        val project = fileTree.create(project, target)
        refreshWorkspace()
        project.checkReferenceIsResolved<RsPath>("src/foo.rs")
    }
```

dima74: I got test working locally by changing
```kotlin
val project = fileTree.create(project, target)
```
to
```kotlin
val project = fileTree.create(project, cargoProjectDirectory)
```
but unfortunately test still doesn't pass on CI

vlad20012 (approved):

dima74:
bors r=vlad20012

PR #6207, opened 2020-10-06
INT: Don't show "Add wildcard pattern" for empty match

dima74:
Shows "Add wildcard pattern" intention only if `match` is not empty. I think it is not reasonable to use "Add wildcard pattern" intention on empty `match` because resulting code will be similar to `if true { ... }`.

| Before | After |
| ------- | ----- |
| image | image |

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -1090,7 +1090,10 @@ sealed class RsDiagnostic(
             ERROR,
             E0004,
             "Match must be exhaustive",
-            fixes = listOf(AddRemainingArmsFix(matchExpr, patterns), AddWildcardArmFix(matchExpr))
+            fixes = listOfNotNull(
+                AddRemainingArmsFix(matchExpr, patterns),
+                AddWildcardArmFix(matchExpr).takeIf { matchExpr.arms.isNotEmpty() }
+            )
         )
     }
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/ide/intentions/AddWildcardArmIntentionTest.kt
```diff
@@ -10,24 +10,6 @@ package org.rust.ide.intentions
  */
 class AddWildcardArmIntentionTest : RsIntentionTestBase(AddWildcardArmIntention::class) {
 
-    fun `test empty match`() = doAvailableTest("""
```
Undin, line 13: If we decided not to show the corresponding intention in this case, let's convert this test using `doUnavailableTest`

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1090,7 +1090,10 @@ sealed class RsDiagnostic(
             ERROR,
             E0004,
             "Match must be exhaustive",
-            fixes = listOf(AddRemainingArmsFix(matchExpr, patterns), AddWildcardArmFix(matchExpr))
+            fixes = listOfNotNull(
+                AddRemainingArmsFix(matchExpr, patterns),
+                AddWildcardArmFix(matchExpr).takeIf { matchExpr.arms.isNotEmpty() }
```
Undin, line 1095: Let's add the corresponding test

ortem (approved):
bors r+

PR #6212, opened 2020-10-07
MACRO: introduce shared persistent cache for macro expansion

vlad20012:
Introduce persistent (stored on disk, in the real file system) cache for macro expansion text and stubs. The cache is shared between different IDEA projects.

The cache is essential for `name resolution 2.0`, but I measured that it doesn't bring significant performance improvements right now (but uses disk storage), so I decided to disable it by default. It can be enabled via `Registry` option `org.rust.lang.macros.persistentCache`. So, the change doesn't affect users for now.


dima74 (approved):

Undin:
bors retry

PR #6214, opened 2020-10-07
Refactor `toolchain` package

mchernyavsky:

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/components/Cargo.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/components/Cargo.kt
@@ -3,7 +3,7 @@
  * found in the LICENSE file.
  */
 
-package org.rust.cargo.toolchain
+package org.rust.cargo.toolchain.components
 
 import com.google.gson.Gson
 import com.google.gson.JsonParser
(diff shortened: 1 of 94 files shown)


src/main/kotlin/org/rust/cargo/toolchain/components/Cargo.kt
```diff
@@ -48,6 +52,14 @@ import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
 
+fun RsToolchain.rawCargo(): Cargo = Cargo(this)
```
Undin, line 55: Do we really need `raw` prefix here? I have never understood why it's needed. It looks rather inconsistent with other tools and doesn't add additional info (at least for me). Maybe drop it while this refactoring as well?

Undin (commented):
Looks ok in general
But I'm not sure we need separate packages for binaries and components. What's the main difference between them? Maybe it's worth merging `binaries` and `components` into a single package?

mchernyavsky:
bors try

mchernyavsky:
@Undin Done.

mchernyavsky (commented):
bors r=Undin

Undin:
bors retry

PR #6220, opened 2020-10-08
TOML: improve parser recovery on errors

vlad20012:
A continuation of #6202

1. Fix TOML parser tests (was needed to manually register `TomlBraceMatcher` in the parser test)
2. Recover on `key = [invalid, {valid = 2}]` case


intellij-toml/src/main/kotlin/org/toml/ide/TomlBraceMatcher.kt
```diff
@@ -21,6 +22,10 @@ class TomlBraceMatcher : PairedBraceMatcher {
 
     companion object {
         private val PAIRS: Array<BracePair> = arrayOf(
+            // Grammar Kit hack
+            BracePair(IElementType("FAKE_L_BRACE", TomlLanguage), IElementType("FAKE_L_BRACE", TomlLanguage), false),
+
+            BracePair(L_CURLY, R_CURLY, false),
```
Undin, line 28: The line below is the same

PR #6246, opened 2020-10-12
Add an extension point for toolchain path suggestion

mchernyavsky:
This can be useful for integration with a plugin that can suggest toolchain paths. For example, WSL integration would suggest something like this:
```
\\wsl$\Ubuntu\home\misha\.cargo\bin
```

Part of https://github.com/intellij-rust/intellij-rust/pull/5529.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/RsToolchain.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/RsToolchain.kt
@@ -9,22 +9,21 @@ import com.intellij.openapi.util.SystemInfo
 import com.intellij.openapi.util.io.FileUtil
 import com.intellij.util.io.exists
 import com.intellij.util.text.SemVer
+import org.rust.cargo.toolchain.flavors.RsToolchainFlavor
 import org.rust.cargo.toolchain.tools.Cargo
-import org.rust.cargo.toolchain.tools.Rustc
+import org.rust.cargo.util.hasExecutable
+import org.rust.cargo.util.pathToExecutable
+import org.rust.stdext.isExecutable
 import java.io.File
-import java.nio.file.Files
 import java.nio.file.Path
 
 open class RsToolchain(val location: Path) {
     val presentableLocation: String = pathToExecutable(Cargo.NAME).toString()
 
-    fun looksLikeValidToolchain(): Boolean = hasExecutable(Cargo.NAME) && hasExecutable(Rustc.NAME)
+    fun looksLikeValidToolchain(): Boolean = RsToolchainFlavor.getFlavor(location) != null
 
     // for executables from toolchain
-    fun pathToExecutable(toolName: String): Path {
-        val exeName = if (SystemInfo.isWindows) "$toolName.exe" else toolName
-        return location.resolve(exeName).toAbsolutePath()
-    }
+    fun pathToExecutable(toolName: String): Path = location.pathToExecutable(toolName)
 
     // for executables installed using `cargo install`
     fun pathToCargoExecutable(toolName: String): Path {
@@ -39,9 +38,9 @@ open class RsToolchain(val location: Path) {
         return cargoBinPath.resolve(exeName).toAbsolutePath()
     }
 
-    fun hasExecutable(exec: String): Boolean = Files.isExecutable(pathToExecutable(exec))
+    fun hasExecutable(exec: String): Boolean = location.hasExecutable(exec)
 
-    fun hasCargoExecutable(exec: String): Boolean = Files.isExecutable(pathToCargoExecutable(exec))
+    fun hasCargoExecutable(exec: String): Boolean = pathToCargoExecutable(exec).isExecutable()
 
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.util.io.isDirectory
+import org.rust.cargo.toolchain.tools.Cargo
+import org.rust.cargo.toolchain.tools.Rustc
+import org.rust.stdext.isExecutable
+import java.nio.file.Path
+
+fun RsSdkFlavor.suggestHomePaths(): List<Path> =
+    getHomePathCandidates().distinct().filter { isValidSdkPath(it) }
+
+interface RsSdkFlavor {
```
Undin, line 19: I'd personally name it `RsToolchainFlavor`

src/main/kotlin/org/rust/cargo/toolchain/RsToolchain.kt
```diff
@@ -10,15 +10,16 @@ import com.intellij.openapi.util.io.FileUtil
...
+import org.rust.stdext.isExecutable
 import java.io.File
-import java.nio.file.Files
 import java.nio.file.Path
 
 data class RsToolchain(val location: Path) {
     val presentableLocation: String = pathToExecutable(Cargo.NAME).toString()
 
-    fun looksLikeValidToolchain(): Boolean = hasExecutable(Cargo.NAME) && hasExecutable(Rustc.NAME)
+    fun looksLikeValidToolchain(): Boolean = RsSdkFlavor.getFlavor(location) != null
```
Undin, line 22: Looks like it's only one usage of `RsSdkFlavor.getFlavor`. Maybe rename it to `looksLikeValidToolchain` and pass necessary implementation inside it?

src/main/resources/META-INF/rust-core.xml
```diff
@@ -1009,10 +1009,22 @@
 
     </extensions>
 
+    <extensions defaultExtensionNs="org.rust">
+        <!-- SDK Flavors -->
+        <sdkFlavor implementation="org.rust.ide.sdk.flavors.RustupSdkFlavor"/>
```
Undin, line 1014: Let's add id for each flavor. It allows us and external developers properly configure order of suggestions

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+            return getApplicableFlavors().find { flavor -> flavor.isValidSdkPath(sdkPath) }
+        }
+
+        @JvmStatic
+        fun Path.hasExecutable(toolName: String): Boolean = pathToExecutable(toolName).isExecutable()
+
+        private fun Path.pathToExecutable(toolName: String): Path {
+            val exeName = if (SystemInfo.isWindows) "$toolName.exe" else toolName
+            return resolve(exeName).toAbsolutePath()
+        }
```
Undin, line 57: We have the same method in `RsToolchain` class. Can we have only one of them?

src/main/kotlin/org/rust/ide/sdk/flavors/RsMacSdkFlavor.kt
```diff
@@ -0,0 +1,24 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.sdk.flavors
```
Undin, line 6: Maybe it's better to move them into `cargo` package?

src/main/kotlin/org/rust/ide/sdk/flavors/RsMacSdkFlavor.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import com.intellij.openapi.util.SystemInfo
+import com.intellij.util.io.isDirectory
+import org.rust.stdext.toPath
+import java.nio.file.Path
+
+object RsMacSdkFlavor : RsSdkFlavor {
+    override fun isApplicable(): Boolean = SystemInfo.isMac
```
Undin, line 14: Consider to use `os` property while registration extension in plugin xml

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        fun getApplicableFlavors(): List<RsSdkFlavor> = EP_NAME.extensionList.filter { it.isApplicable() }
+
+        fun getFlavor(sdkPath: Path?): RsSdkFlavor? {
+            if (sdkPath == null) return null
+            return getApplicableFlavors().find { flavor -> flavor.isValidSdkPath(sdkPath) }
+        }
+
+        @JvmStatic
+        fun Path.hasExecutable(toolName: String): Boolean = pathToExecutable(toolName).isExecutable()
```
Undin, line 52: We have almost the same method in `RsToolchain`. Can we unify them?

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import com.intellij.openapi.extensions.ExtensionPointName
+import com.intellij.openapi.util.SystemInfo
+import com.intellij.util.io.isDirectory
+import org.rust.cargo.toolchain.tools.Cargo
+import org.rust.cargo.toolchain.tools.Rustc
+import org.rust.stdext.isExecutable
+import java.nio.file.Path
+
+fun RsSdkFlavor.suggestHomePaths(): List<Path> =
```
Undin, line 16: nit: Why this method is extension one and all other (e.g. `getApplicableFlavors`) are in companion object?
Let's place it in companion object as well

src/main/resources/META-INF/rust-core.xml
Undin, line 1014: Also, I'd add `order="first"` here just to be sure that we prefer using `rustup` over all other variants. 
Note, the platform doesn't guarantee implicit order of extensions

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+
+interface RsSdkFlavor {
+
+    /**
+     * Flavor is added to result in [getApplicableFlavors] if this method returns true.
+     * @return whether this flavor is applicable.
+     */
+    fun isApplicable(): Boolean = true
+
+    fun getHomePathCandidates(): List<Path>
```
Undin, line 27: Looks like this method is not a part of public API that we want to provide since it should be use only in combination with `isValidSdkPath`
Maybe it's better to convert `RsSdkFlavor` to abstract class and made this method protected?

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+            && sdkPath.hasExecutable(Rustc.NAME)
+            && sdkPath.hasExecutable(Cargo.NAME)
+    }
+
+    companion object {
+        private val EP_NAME: ExtensionPointName<RsSdkFlavor> = ExtensionPointName.create("org.rust.sdkFlavor")
+
+        fun getApplicableFlavors(): List<RsSdkFlavor> = EP_NAME.extensionList.filter { it.isApplicable() }
+
+        fun getFlavor(sdkPath: Path?): RsSdkFlavor? {
```
Undin, line 46: Do we really expect nullable input here? Current usages (only one) pass not null path

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    fun getHomePathCandidates(): List<Path>
+
+    /**
+     * Checks if the path is the name of a Rust toolchain of this flavor.
+     *
+     * @param sdkPath path to check.
+     * @return true if paths points to a valid home.
+     */
+    fun isValidSdkPath(sdkPath: Path): Boolean {
```
Undin, line 35: The same about non-public API

Undin (changes requested):

Undin (changes requested):
There is a compilation error

src/main/kotlin/org/rust/ide/sdk/flavors/RsSdkFlavor.kt
Undin: I suppose we can extract this method to some utility file (for example, somewhere in `org.rust.cargo.util` package) and reuse it in `org.rust.cargo.toolchain.RsToolchain#pathToExecutable` for `location` property

Undin (approved):

mchernyavsky:
bors r=Undin

mchernyavsky (commented):
bors retry

Issue #6200, opened 2020-10-05
NonZero pretty printers don't work with Rust 1.48


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.131.3366-202
* **Rust toolchain version:** 1.48.0-nightly (ef663a8a4 2020-09-30) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.2.3 (CL-202.7319.72)
* **Operating system:** Linux 4.15.0-117-generic

## Problem description
| debugger | expected | actual |
| - | - | - |
| gdb | ![gdb-nonzero-expected](https://user-images.githubusercontent.com/2539310/95070022-2f567b00-0710-11eb-8718-b1db44a099dd.png) | ![gdb-nonzero-actual](https://user-images.githubusercontent.com/2539310/95070051-38474c80-0710-11eb-9984-ee9a4f4a046c.png) |
| lldb | ![lldb-nonzero-expected](https://user-images.githubusercontent.com/2539310/95070205-717fbc80-0710-11eb-9da3-76d5f911321c.png) | ![lldb-nonzero-actual](https://user-images.githubusercontent.com/2539310/95070251-81979c00-0710-11eb-9e0e-a13182faecaf.png) |

## Steps to reproduce
```rust
use std::num::{NonZeroI32, NonZeroUsize};

fn main() {
    let a = NonZeroI32::new(-1).unwrap();
    let b = NonZeroUsize::new(1024).unwrap();
    print!(""); // #break
}
```

Undin:
CC @Kobzol

ortem:
BTW it would be great to add these printers to rustc (they are located in https://github.com/rust-lang/rust/tree/master/src/etc)

PR #6249, opened 2020-10-13
DBG: make NonZero printers compatible with Rust 1.48+

Kobzol:
The `NonZero` types moved from `core::num` to `core::num::nonzero` in Rust `1.48`, this change makes the regexes compatible with both the old and the new versions.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6200


ortem:
Thank you @Kobzol!
bors r+

PR #6251, opened 2020-10-13
Simplify rustc version parsing

mchernyavsky:
It also fixes stdlib's version string in `External Libraries`:

image

(`beta.2` instead of just `beta`)`


Undin (approved):
Nice!
bors r+

Undin:
bors retry

PR #6256, opened 2020-10-14
Fix integer hexadecimal literal parsing

### LINKED ISSUE #6255: Integer hexadecimal literals are parsed as floats
Changes made in #6177 seem to cause a problem for hexadecimal literals containing `e` or `f`.
They seem to be parsed/recognized as `f64` instead of as `u32` / integer literal.

```rust
const CONST1: u32 = 0xd0;
const CONST2: u32 = 0xe0;
const CONST3: u32 = 0xf0;

const CONST4: u32 = 0x0d;
const CONST5: u32 = 0x0e;
const CONST6: u32 = 0x0f;
```

image

After reverting the PR locally for testing, the issue is gone again.

_Originally posted by @kumbayo in https://github.com/intellij-rust/intellij-rust/pull/6177#issuecomment-708005841_

vlad20012:
Fixes #6255 (the bug was introduced in #6177)


src/test/resources/org/rust/lang/core/parser/fixtures/complete/expr.rs
```diff
@@ -80,6 +80,7 @@ fn test() -> u32 {
     let _ = 1f34;
     let _ = 1.0i98;
     let _ = 0.0.0;
+    let _ = 0x1234567890abcdef;
```
Undin, line 83: Maybe it makes sense to extract literal parsing into separate tests?
Also, it would be great to add binary and octal integer literals as well

vlad20012: Done

Undin (approved):
bors r+

PR #6259, opened 2020-10-14
INSP: check if pattern identifiers are not bound multiple times [E0416]

Kobzol:
Adds support for checking the `E0416` error, which occurs if the same identifier is found more than once in the same pattern.

Related: https://github.com/intellij-rust/intellij-rust/issues/886


mchernyavsky (approved):
Thanks!

bors r+

Issue #5632, opened 2020-06-22
Rust support in MSVC LLDB


ortem:
This meta issue is aimed to track the progress of improving Rust support in [MSVC LLDB](https://blog.jetbrains.com/clion/2019/06/clion-2019-2-eap-msvc-debugger-unused-includes-check-and-more/#msvc_debug).

#### IDE support
- [x] CLion #5535
- [ ] IDEA Ultimate, AppCode, RubyMine #5610

#### Rendering
- [ ] Render enums properly
- [ ] Render primitive types properly
- [ ] Render `&str` properly

#### Other
- [ ] Set up pretty-printers tests on Windows

zhuxiujia:
will support Clion  with MSVC LLDB ???

Undin:
@zhuxiujia MSVC LLDB in CLion works since #5535

harry1453:
When you're using the MSVC rust toolchain and MinGW is configured as the toolchain in CLion, it's not possible to debug without setting Visual Studio as the default toolchain. Is there a way to tell CLion which toolchain to use for rust? If you have multiple CLion projects open, changing the default affects all of them so I want to avoid this.

ortem:
@harry1453 If you have a CMake project and a Rust project and want to use different CLion toolchains for them, you can try to do the following:
- set CLion toolchain for Rust as a default one (top of the [Toolchains list](https://www.jetbrains.com/help/clion/toolchains.html))
- set another CLion toolchain in [CMake profile](https://www.jetbrains.com/help/clion/cmake-profile.html) for your CMake project

Please let us know if it helps.

PR #5200, opened 2020-04-06
DBG: provide debugger support outside of CLion

Undin:
* works starting with 2020.1
* works only for Linux and macOS (Windows support is blocked by msvc toolchain support #4104)
* supported IDEs: IDEA Ultimate, AppCode and RubyMine
* requires [nativeDebug](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin for IDEA Ultimate (advertising of native debug plugin is a separate task #5201)
* only LLDB
* debugger is not bundled and need to be downloaded. Also, debugger binaries should be updated after IDE/native debug plugin update (see #5202 as possible improvement)

idea dbg 2020-04-06 16_54_05

Fixes #2273
Fixes #3453

PR #6267, opened 2020-10-16
DBG: support debugging on Windows outside of CLion

### LINKED ISSUE #5610: Support debugging with msvc rust toolchain on Windows outside of CLion
After #5200, the plugin provides the ability to use debugger outside of CLion. But this feature is available only in Linux and macOS for now. It would be great to support debugging support on Windows as well.

Blocked by #4104

Undin:
Continuation of #5200 and #5535

Restrictions:
* supported IDEs: IDEA Ultimate and RubyMine
* requires [Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin for IDEA Ultimate
* only LLDB
* only Rust MSVC toolchain
* debugger is not bundled and has to be downloaded. Also, debugger binaries should be updated after IDE/[Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin update

See #5632 for details about the status of MSVC LLDB support/restrictions 

Fixes #5610

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/RsDebuggerToolchainService.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/RsDebuggerToolchainService.kt
@@ -40,10 +40,7 @@ class RsDebuggerToolchainService {
         val (frameworkPath, frontendPath) = when {
             SystemInfo.isMac -> "LLDB.framework" to "LLDBFrontend"
             SystemInfo.isUnix -> "lib/liblldb.so" to "bin/LLDBFrontend"
-            SystemInfo.isWindows -> {
-                val binaryDir = "${if (SystemInfo.is32Bit) "x86" else "x64"}/bin"
-                "$binaryDir/liblldb.dll" to "$binaryDir/LLDBFrontend.exe"
-            }
+            SystemInfo.isWindows -> "bin/liblldb.dll" to "bin/LLDBFrontend.exe"
             else -> return LLDBStatus.Unavailable
         }
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ The plugin is compatible with all IntelliJ-based IDEs starting from the version 
 
 \** Requires the
 [Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin. 
-Available on macOS and Linux, LLDB only.
+LLDB only.
 
 ## TOML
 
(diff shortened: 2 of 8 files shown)


Undin:
Main test cases:
- a user doesn't have [Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin in IntelliJ IDEA Ultimate - debugger icon should be enabled and show a dialog to install the plugin
- a user has [Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin - rust plugin should suggest downloading debugger binaries
- a user uses gnu toolchain - an error dialog should be shown with the corresponding message
- a user uses msvc toolchain and debugger binaries are downloaded - a usual debugging process

mchernyavsky (approved):
LGTM

debugger/src/main/kotlin/org/rust/debugger/RsDebuggerToolchainService.kt
```diff
@@ -40,10 +40,7 @@ class RsDebuggerToolchainService {
         val (frameworkPath, frontendPath) = when {
             SystemInfo.isMac -> "LLDB.framework" to "LLDBFrontend"
             SystemInfo.isUnix -> "lib/liblldb.so" to "bin/LLDBFrontend"
-            SystemInfo.isWindows -> {
-                val binaryDir = "${if (SystemInfo.is32Bit) "x86" else "x64"}/bin"
-                "$binaryDir/liblldb.dll" to "$binaryDir/LLDBFrontend.exe"
-            }
+            SystemInfo.isWindows -> "bin/liblldb.dll" to "bin/LLDBFrontend.exe"
```
ortem, line 43: I've noticed that we download LLDB x64 in case of 64-bit Windows and LLDB x86 in case of 32-bit Windows (see `lldbUrls`). Just wondering, what happens if someone uses
1. `stable-i686-pc-windows-msvc` Rust on 64-bit Windows
2. `stable-x86_64-pc-windows-msvc` Rust on 32-bit Windows

I've tried the first case and it works fine

ortem (approved):

debugger/src/main/kotlin/org/rust/debugger/RsDebuggerToolchainService.kt
Undin: I think it's the same situation when someone tries to compile code to non-host platform.
Currently, we do nothing with it on any platform.

> I've tried the first case and it works fine

In general, a user can be not able to run the final binary but in the case of Windows, I suppose it works because 64bit Windows can run 32bit binaries. 
Looks like we should notify users when target platform of binary is not matched with the host target on any platform and in any IDE (including CLion). For example, there is `i686-unknown-linux-gnu` target for Linux.
But I don't want to do it in this PR. It should be part of cross-compilation support in the plugin

Issue #6104, opened 2020-09-13
Rework Cargo project structure loading


Undin:
Cargo project structure loading was designed to be almost invisible to users. In the beginning, it called only `cargo metadata` to retrieve project structure and it's very quick. As a result, automatic project structure reloading was turned on.

But later we found out that we need to run some long operations like compile (part of) project to get compile time info (`OUT_DIR` variable, `cfg` attributes generated by build script, etc.).

Also, the platform support of build tools is also improving constantly and now the current platform contains `Sync` tool window to show result of project importing, common settings of automatic project structure reloading (it may be essential when your reloading operation may compile the whole project), etc.

So we want to rework some part of cargo project structure loading to properly support new needs from one side and new platform features from another side.

This issue is a tracking meta issue to show what we want to change and what we've already done. Note, some items are quite technical and are not visible to users.
* [ ] Create `CargoSyncTask` instead of using `org.rust.openapiext.AsyncTasksKt#runAsyncTask`
* [ ] Use platform read/write lock instead of `org.rust.stdext.AsyncValue` to keep project info
* [ ] Retrieve metadata from stdlib. Currently, we have a hardcoded info about stdlib. It should support proper edition detection in stdlib and get info about dependencies of stdlib
* [ ] Convert `org.rust.cargo.evaluate.build.scripts` experimental feature to user visible setting
* [ ] Take into account project loading settings (see `Preferences | Build, Execution, Deployment | Build Tools`)
* [ ] Show floating button to reload project structure if it's outdated
* [ ] Support platform `Load Changes` action
* [ ] Support `Sync` tool window to show results of project loading (including failed compilation if it's needed) #4699 #4546
* [ ] Don't invoke project reloading on non-important change like the addition of space in `Cargo.toml` #5062
* [ ] Don't invoke project reloading twice after: the first one after `Cargo.toml` change and the second one after `Cargo.lock` change because of previous corresponding changes in `Cargo.toml`

tdrozdowski:
Will this include restoring the code completion of files generated and out into the OUT_DIR?  Its impossible to work with Tonic/Prost (gRPC) with he latest builds - nothing has worked since 126.

Undin:
@tdrozdowski looks like it's an unknown issue and I think it should be fixed separately even without changes described in this issue. Could you file a new issue and attach code example to reproduce it?
[...]

PR #6272, opened 2020-10-19
CARGO: doesn't load Cargo project structure twice

Undin:
Doesn't trigger the loading of Cargo projects structure on the first external change in `Cargo.lock` file after `cargo metadata` call because this change is performed  `cargo metadata` command itself.
It should reduce the loading of cargo projects. It may be essential if/when loading takes a long time

Part of #6104

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoEventService.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoEventService.kt
@@ -1,0 +1,50 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.cargo.project.model.impl
+
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+import com.intellij.openapi.project.Project
+import org.rust.cargo.project.model.CargoProjectsService.CargoProjectsListener
+import org.rust.cargo.project.model.CargoProjectsService.Companion.CARGO_PROJECTS_TOPIC
+import org.rust.cargo.runconfig.command.workingDirectory
+import org.rust.stdext.mapToSet
+import java.nio.file.Path
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+
+/**
+ * The plugin uses `cargo metadata` command to retrieve project structure from Cargo point of view.
+ * It may lead to changes in `Cargo.lock` file.
+ * At the same time, the plugin watches changes in `Cargo.lock` to update project structure
+ * because the content of `Cargo.lock` is taken into account by Cargo while `cargo metadata` command.
+ * As a result, the plugin may call `cargo metadata` twice in a row and the second call always redundant.
+ *
+ * This service keeps timestamps of previous `cargo metadata` invocation for each cargo project
+ * to check changes in `Cargo.lock` should be skipped and avoid unnecessary project loading that can be quite long sometimes
+ */
+@Service
+class CargoEventService(project: Project) {
+
+    private val metadataCallTimestamps: ConcurrentMap<Path, Long> = ConcurrentHashMap()
+
+    init {
+        project.messageBus.connect().subscribe(CARGO_PROJECTS_TOPIC, CargoProjectsListener { _, projects ->
+            val projectDirs = projects.mapToSet { it.workingDirectory }
+            metadataCallTimestamps.keys.retainAll(projectDirs)
+        })
+    }
+
+    fun onMetadataCall(projectDirectory: Path) {
+        metadataCallTimestamps[projectDirectory] = System.currentTimeMillis()
+    }
+
+    fun extractTimestamp(projectDirectory: Path): Long? = metadataCallTimestamps.remove(projectDirectory)
+
+    companion object {
+        fun getInstance(project: Project): CargoEventService = project.service()
+    }
+}
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
@@ -19,6 +19,7 @@ import com.intellij.execution.process.ProcessAdapter
 import com.intellij.execution.process.ProcessEvent
 import com.intellij.icons.AllIcons
 import com.intellij.openapi.actionSystem.AnActionEvent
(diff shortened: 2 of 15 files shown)


src/main/kotlin/org/rust/cargo/project/model/impl/CargoEventService.kt
```diff
@@ -0,0 +1,40 @@
...
+
+@Service
+class CargoEventService(project: Project) {
+
+    val metadataCallTimestamps: ConcurrentMap<Path, Long> = ConcurrentHashMap()
+
+    init {
+        project.messageBus.connect().subscribe(CARGO_PROJECTS_TOPIC, CargoProjectsListener { _, projects ->
+            val projectDirs = projects.mapToSet { it.workingDirectory }
+            metadataCallTimestamps.keys.removeIf { it !in projectDirs }
```
vlad20012, line 27:
```suggestion
            metadataCallTimestamps.keys.retainAll(projectDirs)
```

src/main/kotlin/org/rust/cargo/project/model/impl/CargoEventService.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.rust.cargo.project.model.CargoProjectsService.CargoProjectsListener
+import org.rust.cargo.project.model.CargoProjectsService.Companion.CARGO_PROJECTS_TOPIC
+import org.rust.cargo.runconfig.command.workingDirectory
+import org.rust.stdext.mapToSet
+import java.nio.file.Path
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+
+@Service
+class CargoEventService(project: Project) {
```
vlad20012, line 20: Let's add a short comment about these hacks. 
By the way, maybe just move `metadataCallTimestamps` to CargoProjectsServiceImpl?

src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
```diff
@@ -201,6 +208,14 @@ private fun fetchCargoWorkspace(context: CargoSyncTask.SyncContext): TaskResult<
         val cargo = toolchain.cargoOrWrapper(projectDirectory)
         try {
             val projectDescriptionData = cargo.fullProjectDescription(childContext.project, projectDirectory, object : ProcessAdapter() {
+                override fun startNotified(event: ProcessEvent) {
+                    val processHandler = event.processHandler as? BaseProcessHandler<*> ?: return
+                    val commandLine = processHandler.commandLine
+                    if (" metadata " in commandLine) {
+                        CargoEventService.getInstance(childContext.project).onMetadataCall(projectDirectory)
+                    }
+                }
```
vlad20012, line 217: Why `onMetadataCall` call is in `startNotified` but not just before `cargo.fullProjectDescription`? If it is reasonable, let's add a comment about it

src/main/kotlin/org/rust/cargo/project/model/impl/CargoTomlWatcher.kt
```diff
@@ -57,20 +61,39 @@ class CargoTomlWatcher(
...
+                    val timestamp = CargoEventService.getInstance(project).extractTimestamp(projectDir) ?: 0
+                    if (event.requestor != null) return true
+                    val current = System.currentTimeMillis()
+                    val delayThreshold = Registry.intValue("org.rust.cargo.lock.update.delay.threshold")
+                    val delay = current - timestamp
+                    return if (delay > delayThreshold) {
+                        LOG.info("External change in ${event.path}. Previous Cargo metadata call was $delay ms before")
+                        true
+                    } else {
+                        LOG.info("Skip external change for ${event.path}. Previous Cargo metadata call was $delay ms before")
```
vlad20012, line 78: Maybe `debug`?

src/main/kotlin/org/rust/cargo/project/model/impl/CargoTomlWatcher.kt
```diff
@@ -57,20 +61,39 @@ class CargoTomlWatcher(
...
-                val parent = PathUtil.getParentPath(event.path)
-                val grandParent = PathUtil.getParentPath(parent)
-                IMPLICIT_TARGET_DIRS.any { parent.endsWith(it) || (event.pathEndsWith("main.rs") && grandParent.endsWith(it)) }
+        fun isInterestingEvent(project: Project, event: VFileEvent): Boolean {
+            return when {
+                event.pathEndsWith(CargoConstants.MANIFEST_FILE) -> true
+                event.pathEndsWith(CargoConstants.LOCK_FILE) -> {
+                    val projectDir = Paths.get(event.path).parent
+                    val timestamp = CargoEventService.getInstance(project).extractTimestamp(projectDir) ?: 0
+                    if (event.requestor != null) return true
```
vlad20012, line 70:
```suggestion
                    if (event.requestor != null) return true // Non-null requestor means a change from Intellij
```

vlad20012 (commented):
LGTM in general

src/main/kotlin/org/rust/cargo/project/model/impl/CargoEventService.kt
Undin: > By the way, maybe just move metadataCallTimestamps to CargoProjectsServiceImpl?

It will lead to complication of `CargoProjectsServiceImpl` itself and using `CargoProjectsServiceImpl` in some additional places that I think we should try to avoid. Of course, we can add some method/property into `CargoProjectsService` but I consider that these changes are too specific and I'd prefer to keep them separately from `CargoProjectsService`

src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
Undin: Not, it's not reasonable. Thanks for the suggestion!

src/main/kotlin/org/rust/cargo/project/model/impl/CargoTomlWatcher.kt
Undin: It doesn't look this message will appear in logs too frequently (only on cargo metadata refresh after some change in Cargo.toml or at the first time) so I'd prefer to keep it enabled by default to catch potential problems from logs and not to ask users to enable them

Undin:
@vlad20012 done

PR #6279, opened 2020-10-21
ANN: take into account workspace features in external linter

vlad20012:
Now external linter (`cargo check` or `clippy`) takes into account workspace features configuration (from #5189):

Peek 2020-10-20 20-57

I made some changes in external linter invocation mechanics:

1. Pass `--no-default-features` and `--features` options to `cargo check` invocation
2. In a workspace, instead of invoking `cargo check` in a workspace root with `--package foo` option, now we execute `cargo check` in a **package root** without package option. It's needed because it's not possible to pass `--features` option in a workspace root if the workspace consists of multiple packages.
3. Pass concrete `target` to `cargo check` invocation, e.g. `--bin foo`, `--lib`, `--test foo`. This is not strictly needed, but it makes the code simpler, and also it should speed up the linting.


mchernyavsky (approved):
👍 👍 👍 

bors r+

PR #6285, opened 2020-10-21
GRAM: better parser recovery for incomplete paths (pin path) 

vlad20012:
Works in the case of `foo::`.
Previous attempt - #4199

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
@@ -79,7 +79,7 @@ class AutoImportFix(element: RsElement, private val type: Type) : LocalQuickFixO
             if (path.reference == null) return null
 
             val basePath = path.basePath()
-            if (!basePath.isUnresolved) return null
+            if (basePath.resolveStatus != PathResolveStatus.UNRESOLVED) return null
 
             if (path.ancestorStrict<RsUseSpeck>() != null) {
                 // Don't try to import path in use item
@@ -87,7 +87,9 @@ class AutoImportFix(element: RsElement, private val type: Type) : LocalQuickFixO
                 return null
             }
 
-            val isNameInScope = path.hasInScope(basePath.referenceName, TYPES_N_VALUES)
+            val referenceName = basePath.referenceName ?: return null
+
+            val isNameInScope = path.hasInScope(referenceName, TYPES_N_VALUES)
             if (isNameInScope) {
                 // Don't import names that are already in scope but cannot be resolved
                 // because namespace of psi element prevents correct name resolution.
@@ -99,7 +101,7 @@ class AutoImportFix(element: RsElement, private val type: Type) : LocalQuickFixO
             val superPath = path.rootPath()
             val candidates = ImportCandidatesCollector.getImportCandidates(
                 ImportContext.from(project, path, false),
-                basePath.referenceName,
+                referenceName,
                 superPath.text
             ) {
                 superPath != basePath || !(it.item is RsMod || it.item is RsModDeclItem || it.item.parent is RsMembers)
--- a/src/main/kotlin/org/rust/ide/intentions/AddCurlyBracesIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/AddCurlyBracesIntention.kt
@@ -49,11 +49,11 @@ class AddCurlyBracesIntention : RsElementBaseIntentionAction<AddCurlyBracesInten
     }
 
     override fun invoke(project: Project, editor: Editor, ctx: Context) {
-        val identifier = ctx.path.referenceNameElement
+        val identifier = ctx.path.referenceNameElement?.text ?: ""
 
         // Create a new use item that contains a glob list that we can use.
         // Then extract from it the glob list and the double colon.
-        val newUseSpeck = RsPsiFactory(project).createUseSpeck("dummy::{${identifier.text}}")
+        val newUseSpeck = RsPsiFactory(project).createUseSpeck("dummy::{$identifier}")
         val newGroup = newUseSpeck.useGroup ?: return
         val newColonColon = newUseSpeck.coloncolon ?: return
 
--- a/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
@@ -60,11 +60,11 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
         val path = element.parentOfType<RsPath>()
         val functionCall = path?.parentOfType<RsCallExpr>()
         if (functionCall != null) {
-            if (!path.isUnresolved) return null
+            if (path.resolveStatus != PathResolveStatus.UNRESOLVED) return null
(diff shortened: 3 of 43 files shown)


Undin (commented):
Side effect - for some invalid paths the plugin produces unexpected (at least, for me) `Unresolved reference` errors. Note, if path is correct, the inspection doesn't produce such errors.
Don't forget to disable `Ignore unresolved references without quick fix` option to reproduce this issue.
image

src/main/kotlin/org/rust/lang/core/psi/ext/RsPathReferenceElement.kt
```diff
@@ -12,8 +12,8 @@ import org.rust.lang.core.resolve.ref.RsPathReference
 interface RsPathReferenceElement : RsReferenceElement {
     override fun getReference(): RsPathReference?
 
-    override val referenceNameElement: PsiElement
+    override val referenceNameElement: PsiElement?
 
     @JvmDefault
-    override val referenceName: String get() = referenceNameElement.unescapedText
+    override val referenceName: String get() = referenceNameElement?.unescapedText ?: ""
```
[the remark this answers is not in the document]
vlad20012: I made it nullable!

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -60,11 +60,11 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
         val path = element.parentOfType<RsPath>()
         val functionCall = path?.parentOfType<RsCallExpr>()
         if (functionCall != null) {
-            if (!path.isUnresolved) return null
+            if (path.resolveStatus == PathResolveStatus.RESOLVED) return null
```
Undin, line 63: Does this intention make sense in case of a malformed path?
Looks like we should use `!= PathResolveStatus.UNRESOLVED` here

src/main/kotlin/org/rust/lang/core/psi/ext/RsMetaItem.kt
```diff
@@ -23,7 +23,7 @@ val RsMetaItem.id: String?
         .asIterable()
         .reversed()
         .takeIf { it.isNotEmpty() }
-        ?.joinToString("::") { it.referenceName }
+        ?.joinToString("::") { it.referenceName.orEmpty() }
```
Undin, line 26: Maybe it's better to return null here if any `referenceName` is null?

src/main/kotlin/org/rust/ide/inspections/import/AutoImportFix.kt
```diff
@@ -79,15 +79,17 @@ class AutoImportFix(element: RsElement, private val type: Type) : LocalQuickFixO
             if (path.reference == null) return null
 
             val basePath = path.basePath()
-            if (!basePath.isUnresolved) return null
+            if (basePath.resolveStatus == PathResolveStatus.RESOLVED) return null
```
Undin, line 82: Does this quick fix make sense in case of a malformed path?
Looks like we should use `!= PathResolveStatus.UNRESOLVED` here

src/main/kotlin/org/rust/ide/intentions/AddCurlyBracesIntention.kt
```diff
@@ -49,11 +49,11 @@ class AddCurlyBracesIntention : RsElementBaseIntentionAction<AddCurlyBracesInten
     }
 
     override fun invoke(project: Project, editor: Editor, ctx: Context) {
-        val identifier = ctx.path.referenceNameElement
+        val identifier = ctx.path.referenceNameElement?.text ?: ""
 
         // Create a new use item that contains a glob list that we can use.
         // Then extract from it the glob list and the double colon.
-        val newUseSpeck = RsPsiFactory(project).createUseSpeck("dummy::{${identifier.text}}")
+        val newUseSpeck = RsPsiFactory(project).createUseSpeck("dummy::{$identifier}")
```
Undin, line 56: It would be great to add a test that checks that we provide this intention in the following case: `use foo::/*caret*/`.
It's a definitely improvement of the intention and it would be great to commit such behavior not to break it in the future

src/main/kotlin/org/rust/lang/core/completion/lint/RsLintCompletionProvider.kt
```diff
@@ -80,7 +80,7 @@ abstract class RsLintCompletionProvider : RsCompletionProvider() {
 
     protected fun getPathPrefix(path: RsPath): String {
         val qualifier = path.qualifier ?: return ""
-        return "${getPathPrefix(qualifier)}${qualifier.referenceName}::"
+        return "${getPathPrefix(qualifier)}${qualifier.referenceName.orEmpty()}::"
```
Undin, line 83: Looks like it produces undesired completion for `#[allow(::/*caret*/)]`

src/main/kotlin/org/rust/lang/core/psi/ext/RsMacroCall.kt
```diff
@@ -44,7 +44,7 @@ abstract class RsMacroCallImplMixin : RsStubbedElementImpl<RsMacroCallStub>,
 }
 
 val RsMacroCall.macroName: String
-    get() = path.referenceName
+    get() = path.referenceName.orEmpty()
```
Undin, line 47: It seems we want to return null if `referenceName` is null. Let's Add the corresponding `TODO` here.
I don't want to fix it in this PR - it's already quite large

src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
```diff
@@ -1197,7 +1198,7 @@ private fun collectMacrosImportedWithUseItem(
 
     return buildList {
         for ((crateName, macroName) in twoSegmentPaths) {
-            val crateRoot = exportingMacrosCrates[crateName.referenceName] as? RsFile
+            val crateRoot = exportingMacrosCrates[crateName.referenceName.orEmpty()] as? RsFile
```
Undin, line 1201: Do we really want to do anything when `crateName.referenceName` is null?

src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
```diff
@@ -186,7 +186,7 @@ private fun RsMacroCall.resolveToMacroDefInfo(containingModInfo: RsModInfo): Mac
 
 private val RsMacroCall.pathSegments: List<String>
     get() = generateSequence(path) { it.path }
-        .map { it.referenceName }
+        .map { it.referenceName.orEmpty() }
```
Undin, line 189: Maybe it's better to return `null` if any `referenceName` is null?

src/main/kotlin/org/rust/ide/intentions/AddCurlyBracesIntention.kt
vlad20012: Not related to this PR

src/main/kotlin/org/rust/lang/core/completion/lint/RsLintCompletionProvider.kt
vlad20012: Not related to this PR

Undin (approved):
bors r+

PR #6286, opened 2020-10-22
Show Rust toolchain in status bar

mchernyavsky:
widget

Depends on https://github.com/intellij-rust/intellij-rust/pull/5529.


Kobzol:
@mchernyavsky Will this be available in https://github.com/intellij-rust/intellij-rust/pull/6456?

PR #6204, opened 2020-10-06
RES: resolve cargo features in `Cargo.toml`

vlad20012:
Completion and `find usages` will be done in a separate PRs.

Peek 2020-10-20 11-53

PR #6289, opened 2020-10-24
COMP: complete cargo features in `Cargo.toml`

vlad20012:
Continuation of #6204 - now with features completion in `Cargo.toml` !

Related to #4693

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/Util.kt
+++ b/toml/src/main/kotlin/org/rust/toml/Util.kt
@@ -143,3 +143,33 @@ fun findDependencyTomlFile(element: TomlElement, depName: String): TomlFile? =
     element.findCargoPackageForCargoToml()
         ?.findDependencyByPackageName(depName)
         ?.getPackageTomlFile(element.project)
+
+/**
+ * Consider `Cargo.toml`:
+ *
+ * ```
+ * [dependencies]
+ * foo = { version = "*",                 features = [] }
+ * #X                ^ returns `foo` for this value #^ and for this value
+ * ```
+ *
+ * ```
+ * [dependencies.foo]
+ * features = []
+ *           #^ returns `foo` for this value
+ * ```
+ *
+ * @see [org.rust.toml.resolve.CargoTomlDependencyFeaturesReferenceProvider]
+ */
+val TomlValue.containingDependencyKey: TomlKey?
+    get() {
+        val parentParent = parent?.parent as? TomlElement ?: return null
+        return if (parentParent is TomlTable) {
+            // [dependencies.foo]
+            parentParent.header.names.lastOrNull()
+        } else {
+            // [dependencies]
+            // foo = { ... }
+            (parentParent.parent as? TomlKeyValue)?.key
+        }
+    }
--- a/toml/src/main/kotlin/org/rust/toml/Util.kt
+++ b/toml/src/main/kotlin/org/rust/toml/Util.kt
@@ -92,6 +92,19 @@ class StringValueInsertionHandler(private val keyValue: TomlKeyValue) : InsertHa
     }
 }
 
+/** Wraps inserted string with quotes if needed */
+class StringLiteralInsertionHandler : InsertHandler<LookupElement> {
+    override fun handleInsert(context: InsertionContext, item: LookupElement) {
+        val leaf = context.getElementOfType<PsiElement>() ?: return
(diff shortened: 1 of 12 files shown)


toml/src/main/kotlin/org/rust/toml/Util.kt
```diff
@@ -92,6 +92,19 @@ class StringValueInsertionHandler(private val keyValue: TomlKeyValue) : InsertHa
     }
 }
 
+/** Wraps inserted string with quotes if needed */
+class StringLiteralInsertionHandler : InsertHandler<LookupElement> {
+    override fun handleInsert(context: InsertionContext, item: LookupElement) {
+        val leaf = context.getElementOfType<PsiElement>() ?: return
+        val hasQuotes = leaf.parent is TomlLiteral && leaf.elementType != TomlElementTypes.NUMBER
```
ortem, line 99: Do I understand right that everything inside `TomlLiteral` is either string or number? Or what about other things like booleans? `hasQuotes` looks a bit confusing to me (maybe it works fine, but just looks a bit strange)

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        val pkg = tomlFile.findCargoPackageForCargoToml() ?: return
+        for (dep in pkg.dependencies) {
+            if (dep.pkg.origin == PackageOrigin.STDLIB) continue
+            // TODO avoid AST loading?
+            for (feature in dep.pkg.getPackageTomlFile(tomlFile.project)?.allFeatures().orEmpty()) {
+                result.addElement(
+                    LookupElementBuilder
+                        .createWithSmartPointer("${dep.pkg.name}/${feature.text}", feature)
```
ortem, line 46: Hmm, are you sure about using `dep.pkg.name` name here, not `dep.name`? Looks like this does not work for aliases like `aliased_dep = { package = "dep", version = "1.0" }`

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlDependencyFeaturesCompletionProvider.kt
```diff
@@ -0,0 +1,52 @@
...
+ * features = ["<caret>"]
+ *             #^ Provides completion here
+ * ```
+ *
+ * @see [org.rust.toml.resolve.CargoTomlDependencyFeaturesReferenceProvider]
+ */
+class CargoTomlDependencyFeaturesCompletionProvider : CompletionProvider<CompletionParameters>() {
+    override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
+        val parentParentParent = parameters.position.ancestorOrSelf<TomlArray>()?.parent?.parent as? TomlElement ?: return
+        val pkgName = if (parentParentParent is TomlTable) {
```
ortem, line 37: Code duplication (see `CargoTomlDependencyFeaturesReferenceProvider`)

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/Util.kt
vlad20012: Hmm, this was just copied from somewhere... Yes, I think it's better to write `val hasQuotes = leaf.parent is TomlLiteral && leaf.elementType in TOML_STRING_LITERALS` in this case

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
vlad20012: Oh, it's bad. Looks like I made the same mistake in a lot of places :(

ortem (approved):

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
vlad20012: It's not so simple even in this case. I'll fix this in another PR

vlad20012:
bors r=ortem

Issue #6191, opened 2020-10-02
False positive [E0415] "parameter bound more than once" for trait function


panstromek:
This code is valid but it produces an error "Identifier `x` is bound more than once in this parameter list".

```rust
pub trait Foo {
    fn bar(x: i32, x: i32);
}
```
 Parameter can be bound more than once in trait definitions. Kinda weird, though. I wouldn't expect this to be valid myself.

PR #6291, opened 2020-10-26
INSP: allow duplicate parameter bindings for trait methods without a body [E0415]

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6191

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -51,7 +51,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
     override fun isForceHighlightParents(file: PsiFile): Boolean = file is RsFile
 
     override fun annotateInternal(element: PsiElement, holder: AnnotationHolder) {
-        val rsHolder = RsAnnotationHolder (holder)
+        val rsHolder = RsAnnotationHolder(holder)
         val visitor = object : RsVisitor() {
             override fun visitBaseType(o: RsBaseType) = checkBaseType(rsHolder, o)
             override fun visitCondition(o: RsCondition) = checkCondition(rsHolder, o)
@@ -734,7 +734,8 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
         val coloncolon = args.node.findChildByType(RsElementTypes.COLONCOLON)?.psi ?: return
         // `::` is redundant only in types
         if (PsiTreeUtil.getParentOfType(args, RsTypeReference::class.java, RsTraitRef::class.java) == null) return
-        val annotation = holder.newWeakWarningAnnotation(coloncolon, "Redundant `::`", RemoveElementFix(coloncolon)) ?: return
+        val annotation = holder.newWeakWarningAnnotation(coloncolon, "Redundant `::`", RemoveElementFix(coloncolon))
+            ?: return
         annotation.highlightType(ProblemHighlightType.LIKE_UNUSED_SYMBOL).create()
     }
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1071,7 +1071,13 @@ private fun checkDuplicates(
         element is RsNamedFieldDecl -> RsDiagnostic.DuplicateFieldError(identifier, name)
         element is RsEnumVariant -> RsDiagnostic.DuplicateEnumVariantError(identifier, name)
         element is RsLifetimeParameter -> RsDiagnostic.DuplicateLifetimeError(identifier, name)
-        element is RsPatBinding && owner is RsValueParameterList -> RsDiagnostic.DuplicateBindingError(identifier, name)
+        element is RsPatBinding && owner is RsValueParameterList -> {
+            val parent = owner.parent as? RsFunction
+
+            // allow duplicates in trait methods without a body
+            if (parent?.owner is RsAbstractableOwner.Trait && parent.block == null) return
```
Undin, line 1078:
```suggestion
            if (parent?.owner is RsAbstractableOwner.Trait && parent.isAbstract) return
```

Undin (approved):
Looks good. Thanks!

Probably, we should add a (weak) warning for such cases because it doesn't seem like good style and may lead to unexpected errors - for example, if you just add a default empty implementation, it will lead to a compilation error. Also, our `Implement members` quick fix/action use initial names from trait method that lead to error as well

Kobzol:
Is it OK to just replace `Severity.ERROR` with `Severity.WARN`?

Undin:
> Is it OK to just replace Severity.ERROR with Severity.WARN?

Unfortunately, no.
At least, we shouldn't use error code in this case (as on your screenshot).
But probably it would be better to use inspection for this purpose. It allows a user to:
- change type of highlighting
- disable inspection
- suppress inspection in a concrete place

Kobzol:
I created a new inspection for it. The duplication check is pretty simple, because for abstract methods there can be [no patterns](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6c8508d062906adab0cc4fb57eb259d9) for parameters.

src/main/kotlin/org/rust/ide/inspections/RsDuplicatedTraitMethodBindingInspection.kt
```diff
@@ -0,0 +1,40 @@
...
+                    val name = binding.name ?: return@forEach
+                    if (name in bindings) {
+                        holder.registerProblem(
+                            binding,
+                            "Duplicated parameter name `$name`. Consider renaming it."
+                        )
+                    } else {
+                        bindings.add(name)
+                    }
+                }
```
Undin, line 37: Using approach we don't highlight the first argument. And it's inconsistent with the corresponding error in `RsErrorAnnotator`. Also, it's possible that user should rename the first parameter instead of the second one. Let's better highlight all names with the same names

src/main/kotlin/org/rust/ide/inspections/RsDuplicatedTraitMethodBindingInspection.kt
```diff
@@ -0,0 +1,40 @@
...
+
+                val parameters = function.valueParameterList ?: return
+                val bindings = mutableSetOf<String>()
+                parameters.valueParameterList.forEach {
+                    val binding = it.pat?.findBinding() ?: return@forEach
+                    val name = binding.name ?: return@forEach
+                    if (name in bindings) {
+                        holder.registerProblem(
+                            binding,
+                            "Duplicated parameter name `$name`. Consider renaming it."
```
Undin, line 32:
```suggestion
                            "Duplicated parameter name `$name`. Consider renaming it"
```

src/main/resources/inspectionDescriptions/RsDuplicatedTraitMethodBinding.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+Checks for abstract trait methods that have duplicated parameter names.
```
Undin, line 3:
```suggestion
Checks duplicated parameter names of abstract trait methods.
```

src/main/kotlin/org/rust/ide/inspections/RsDuplicatedTraitMethodBindingInspection.kt
Kobzol: I actually wanted this behaviour, but you're right that it's inconsistent. Btw maybe we should add a quick fix that would select and highlight the parameter for easier renaming? Is there anything existing like that? (I only found quick fixes that rename to a hard-coded string)

Undin: > Btw maybe we should add a quick fix that would select and highlight the parameter for easier renaming?

Rename quick fix will be useful here, indeed. But let's add it in separate PR to merge this PR finally 😄 

> Is there anything existing like that? (I only found quick fixes that rename to a hard-coded string)

Unfortunately, it seems that we don't have such quick fix yet

src/main/kotlin/org/rust/ide/inspections/RsDuplicatedTraitMethodBindingInspection.kt
```diff
@@ -0,0 +1,46 @@
...
+                    val binding = it.pat?.findBinding() ?: return@forEach
+                    val name = binding.name ?: return@forEach
+                    val set = bindings.getOrPut(name) { mutableSetOf() }
+                    set.add(binding)
+                }
+
+                bindings
+                    .filter { it.value.size > 1 }
+                    .forEach { pair ->
+                        pair.value.forEach { binding ->
```
Undin, line 37:
```suggestion
                    .forEach { (_, bindings) ->
                        bindings.forEach { binding ->
```

Undin (approved):
Thanks!

Kobzol:
bors r=undin

PR #6298, opened 2020-10-27
Fix control flow of async blocks

### LINKED ISSUE #6141: Infinite loops in async context causes variables to be marked as unused
## Environment

* **IntelliJ Rust plugin version:** 0.3.130.3337-202
* **Rust toolchain version:** rustc 1.48.0-nightly (f68e08933 2020-09-19)
* **IDE name and version:** IntelliJ Ultimate 2020.2.2 Build #IU-202.7319.50

## Problem description
Variables before an infinite loop in an `async` block are marked as unused, because the IDE thinks that the code after it is unreachable. That would be the case if `await`ed immediately, but when e.g. spawning it on another thread, that is not the case.

## Steps to reproduce
```rust
fn foo() {
    let a = 5;
    let _ = async { loop {} };
    dbg!(a);
}
```
Observe how `a` is grayed out as the IDE thinks that `dbg!(a)` can't be reached.

ortem:
Fixes #6141

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/cfg/CFGBuilder.kt
+++ b/src/main/kotlin/org/rust/lang/core/cfg/CFGBuilder.kt
@@ -339,18 +339,24 @@ class CFGBuilder(
 
     override fun visitBlockExpr(blockExpr: RsBlockExpr) {
         val labelDeclaration = blockExpr.labelDecl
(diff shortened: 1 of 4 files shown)


vlad20012:
:+1: 
bors r+

PR #6305, opened 2020-10-27
COMP: simplify script for fetching Clippy lints

Kobzol:
I noticed in the Rust Analyzer [changelog](https://rust-analyzer.github.io/thisweek/2020/10/26/changelog-48.html) that they also added lint completion a few days ago (what a coincidence! :D ). They use a JSON with all Clippy lints that I didn't know of. It's a much simpler solution than cloning the `clippy` repository, which is what I have done in the [original PR](https://github.com/intellij-rust/intellij-rust/pull/5646).

Theoretically, the Python script could now be moved back into `build.gradle.kts`, but I don't think it's worth it.


scripts/fetch_lints.py
```diff
@@ -54,31 +53,13 @@ def normalize(name):
...
-                       stderr=subprocess.DEVNULL,
-                       stdin=subprocess.DEVNULL,
-                       cwd=os.path.join(DIR, "rust-clippy"),
-                       check=True)
-
-    sys.path.insert(0, os.path.join(DIR, "rust-clippy", "util"))
-    from lintlib import parse_all
-
-    clippy_lints, _ = parse_all(os.path.join(DIR, "rust-clippy", "clippy_lints", "src"))
+    clippy_lints = requests.get("http://rust-lang.github.io/rust-clippy/master/lints.json").json()
```
Undin, line 56: Can we use `urllib` instead of external `requests`?
Otherwise, you have to install `requests` into your system python to run `updateLints` task that unnecessarily complicates updating lints at the first time

Kobzol: I thought that `requests` is already a "required dependency", since it's in the [`Pipfile`](https://github.com/intellij-rust/intellij-rust/blob/master/scripts/Pipfile) and it's already used by [`release-actions.py`](https://github.com/intellij-rust/intellij-rust/blob/master/scripts/release-actions.py). I'll change it to urllib.

Undin: The main difference that `release-actions.py` is supposed to be run directly
In case of `updateLints` task, python script is run by gradle that knows nothing about pipenv currently. It leads using system pythong interpreter

Kobzol: Good point! Didn't think of that :) It's changed now.

Undin (approved):
Thanks!
bors r+

PR #6306, opened 2020-10-27
REF: Support references to generic items in move refactoring

dima74:

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
@@ -31,7 +31,28 @@ class RsPathUsageInfo(
 }
 
 class RsMoveReferenceInfo(
+    /**
+     * [pathOldOriginal] is real path (from user files).
+     * [pathOld] is our helper path (created with [RsCodeFragmentFactory]), which is more convenient to work with.
+     *
+     * In most cases [pathOld] equals to [pathOldOriginal], but there are two corner cases:
+     * 1) Paths with type arguments: `mod1::mod2::Struct1::<T, R>`
+     *                                ^~~~~~~~~~~~~~~~~~~~~~~~~~^ pathOldOriginal
+     *                                ^~~~~~~~~~~~~~~~~~^ pathOld
+     *    We create [pathOld] from [pathOldOriginal] by deleting type arguments
+     * 2) Paths to nullary enum variants in bindings.
+     *    Unfortunately they are parsed as [RsPatIdent]:
+     *        match none_or_some {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveUtil.kt
```diff
@@ -31,7 +31,25 @@ class RsPathUsageInfo(
...
+     *
+     * In most cases [pathOld] equals to [pathOldOriginal], but there are two corner cases:
+     * 1) Paths with type arguments: `mod1::mod2::Struct1::<T, R>`
+     *                                ^~~~~~~~~~~~~~~~~~~~~~~~~~^ pathOldOriginal
+     *                                ^~~~~~~~~~~~~~~~~~^ pathOld
+     * 2) Paths to nullary enum variants in bindings.
+     *    Unfortunately they are parsed as [RsPatIdent]:
+     *        match none_or_some {
+     *            None => ...
+     *            ^~~^ pathOldOriginal, pathOld
```
ortem, line 46: This looks confusing because `pathOld` is `RsPath`, but `None` is not. Am I right that `pathOld` is actually a **new** path created from `None` text?

ortem: IMO, when looking at `RsMoveReferenceInfo` class, it's not obvious why `pathOldOriginal: RsElement` can always be converted to `pathNew: RsPath?`. I mean, why it's impossible to get `RsPatIdent` at the end?

dima74: > Am I right that `pathOld` is actually a **new** path created from `None` text?

Yes

> Why `pathOldOriginal: RsElement` can always be converted to `pathNew: RsPath?`

`pathOldOriginal` is either `RsPath` or `RsPatIdent`:
* If it is `RsPath`, then we either keep it as is, or create new path by deleting type arguments
* If it is `RsPatIdent`, we know that this pat ident is plain identifier, so we create new one-segment `RsPath`

> Why it's impossible to get `RsPatIdent` at the end?

We don't want to get `RsPatIdent` at the end, because all our code expects `RsPath`. So we specially convert `RsPatIdent` to `RsPath`

ortem: OK, thank you! I think it'd be great to add this explanation to the comment because it might be not clear

dima74: Thanks, added

ortem (approved):
bors r+

PR #6307, opened 2020-10-27
REF: Check if target mod has item with same name in move refactoring

dima74:
Show warning in move refactoring if target mod has an item with the same name


ortem (approved):
bors r+

PR #6308, opened 2020-10-28
REF: Check impls in move refactoring when moving to other crate

dima74:
### Rules for inherent impls:
- An implementing type must be defined within the same crate as the original type definition
- https://doc.rust-lang.org/reference/items/implementations.html#inherent-implementations
- https://doc.rust-lang.org/error-index.html#E0116

We should check:
- When moving inherent impl: check that implementing type is also moved
- When moving struct/enum: check that all inherent impls are also moved

### Rules for trait impls:
- Orphan rules: https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules
- https://doc.rust-lang.org/error-index.html#E0117

We should check (denote impls as `impl<P1..=Pn> Trait<T1..=Tn> for T0`):
- When moving trait impl:
    - either implemented trait is in target crate
    - or at least one of the types `T0..=Tn` is a local type
- When moving trait: for each impl of this trait which remains in source crate:
    - at least one of the types `T0..=Tn` is a local type
- Uncovering is not checking, because it is complicated

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
@@ -19,6 +19,11 @@ import org.rust.ide.utils.getTopmostParentInside
 import org.rust.lang.core.macros.setContext
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.type
 
 class RsMoveConflictsDetector(
     private val conflicts: MultiMap<PsiElement, String>,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
```diff
@@ -164,6 +169,94 @@ class RsMoveConflictsDetector(
...
+     * - Uncovering is not checking, because it is complicated
+     */
+    fun checkImpls() {
+        if (sourceMod.crateRoot == targetMod.crateRoot) return
+
+        val structsToMove = movedElementsDeepDescendantsOfType<RsStructOrEnumItemElement>(elementsToMove).toSet()
+        val implsToMove = movedElementsDeepDescendantsOfType<RsImplItem>(elementsToMove)
+        val (inherentImplsToMove, traitImplsToMove) = implsToMove
+            .partition { it.traitRef == null }
+            .run { Pair(first.toSet(), second.toSet()) }
```
ortem, line 199: [minor]
```suggestion
            .run { first.toSet() to second.toSet() }
```

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveConflictsDetector.kt
```diff
@@ -176,3 +269,22 @@ fun addVisibilityConflict(conflicts: MultiMap<PsiElement, String>, reference: Rs
     val message = "$referenceDescription uses $targetDescription which will be inaccessible after move"
     conflicts.putValue(reference, CommonRefactoringUtil.capitalize(message))
 }
+
+private val RsImplItem.implementingType: TyAdt? get() = typeReference?.type as? TyAdt
```
ortem, line 273: What about impls for non-ADT types?

dima74: According to [E0118](https://doc.rust-lang.org/error-index.html#E0118) inherent impl can be defined only for ADT and trait objects. So trait objects indeed are unsupported now. Impls for trait objects are quite rarely used (about 10 such impls for whole `rustc`), so I propose to add corresponding TODO comment for now

ortem (approved):
bors r+

Issue #6311, opened 2020-10-28
Undesired lint completion


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.135.3441-202-nightly
* **Rust toolchain version:** 1.49.0-nightly (beb5ae474 2020-10-04) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2.3 Ultimate Edition (IU-202.7660.26)
* **Operating system:** macOS 10.15.6

## Problem description
Lint completion introduced in #5646 produces undesired completion items for `#[allow(::/*caret*/)]`

![2020-10-28 10 49 21](https://user-images.githubusercontent.com/2539310/97407161-55d09600-190b-11eb-9bc9-587552ef6414.gif)

_Originally posted by @Undin in https://github.com/intellij-rust/intellij-rust/pull/6285#discussion_r513236638_

PR #6313, opened 2020-10-28
COMP: do not complete lints for paths starting with leading ::

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6311


Undin (approved):
Thanks!
bors r+

PR #6314, opened 2020-10-28
ACT: provide proper rustc version by "Create New Issue" action

Undin:
Previously, action produced incorrect version for beta and nightly rustc like `1.47.0-beta.9-beta` because of #6251


Undin:
CC @mchernyavsky

PR #6315, opened 2020-10-28
T: run macro cleaning task synchronously in tests

vlad20012:
Fixes possible false-positive "DummyDir should not be touched" assertion fails in tests


vlad20012:
ищкы к+

PR #6318, opened 2020-10-28
INSP: It is not allowed to cast to a bool. (E0054)

zeroeightysix:
This PR aims to implement an inspection for rust error [E0054](https://doc.rust-lang.org/error-index.html#E0054) *It is not allowed to cast to a bool.*

If this is merged, I am also interested in implementing an intent to "compare with zero instead", as the error index (and compiler) suggests.

Part of #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsCastToBoolInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsCastToBoolInspection.kt
@@ -1,0 +1,32 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import org.rust.lang.core.psi.RsCastExpr
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.types.ty.TyBool
+import org.rust.lang.core.types.ty.TyPrimitive
+import org.rust.lang.core.types.ty.TyUnit
+import org.rust.lang.core.types.type
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.lang.utils.addToHolder
+
+class RsCastToBoolInspection : RsLocalInspectionTool() {
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitCastExpr(castExpr: RsCastExpr) {
+            // It is allowed to cast a bool to a bool, so if the expression's type is of bool, ignore this cast.
+            // Casts from non primitive types (and the unit type) emit another error, so we ignore those as well.
+            val exprType = castExpr.expr.type
+            if (exprType === TyBool || exprType !is TyPrimitive || exprType === TyUnit) return
+
+            if (castExpr.typeReference.type === TyBool) {
+                RsDiagnostic.CastAsBoolError(castExpr).addToHolder(holder)
+            }
+        }
+    }
+
+}
--- a/src/main/resources/META-INF/rust-core.xml
+++ b/src/main/resources/META-INF/rust-core.xml
@@ -524,6 +524,11 @@
                          enabledByDefault="true" level="ERROR"
                          implementationClass="org.rust.ide.inspections.RsTraitImplementationInspection"/>
 
+        <localInspection language="Rust" groupName="Rust"
+                         displayName="Cast to bool"
+                         enabledByDefault="true" level="ERROR"
+                         implementationClass="org.rust.ide.inspections.RsCastToBoolInspection"/>
+
         <localInspection language="Rust" groupPath="Rust" groupName="Lints"
                          displayName="Unreachable patterns"
                          enabledByDefault="true" level="WARNING"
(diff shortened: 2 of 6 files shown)


src/main/resources/META-INF/rust-core.xml
```diff
@@ -524,6 +524,11 @@
                          enabledByDefault="true" level="ERROR"
                          implementationClass="org.rust.ide.inspections.RsTraitImplementationInspection"/>
 
+        <localInspection language="Rust" groupName="Rust"
+                         displayName="Cast to bool"
+                         enabledByDefault="true" level="ERROR"
+                         implementationClass="org.rust.ide.inspections.RsCastToBoolInspection"/>
+
         <localInspection language="Rust" groupPath="Rust" groupName="Lints"
```
zeroeightysix, line 532: Is there an ordering to be followed in this file? I tossed the entry next to the intent for E0050 (as this PR is for E0054, so they line up).

src/test/kotlin/org/rust/ide/inspections/RsCastToBoolInspectionTest.kt
```diff
@@ -0,0 +1,31 @@
...
+        fn main() {
+            <error descr="$errorMessage [E0054]">5 as bool</error>;
+        }
+    """)
+
+    fun testCastBoolToBool() = checkByText("""
+        fn main() {
+            true as bool;
+        }
+    """)
```
zeroeightysix, line 23: I wasn't sure how else to test this, but what is being tested here is that there *aren't* any errors in the provided code. I'd love to know if there's a better way to do it!

src/test/kotlin/org/rust/ide/inspections/RsCastToBoolInspectionTest.kt
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.rust.ide.inspections
+
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.lang.utils.RsDiagnostic.CastAsBoolError.Companion.errorMessage
+
+class RsCastToBoolInspectionTest : RsInspectionsTestBase(RsCastToBoolInspection::class) {
+
+    fun testCastNumToBool() = checkByText("""
```
Undin, line 13:
```suggestion
    fun `test cast num to bool`() = checkByText("""
```
We use names with spaces enclosed in backticks for test methods. [See](https://kotlinlang.org/docs/reference/coding-conventions.html#names-for-test-methods)

src/test/kotlin/org/rust/ide/inspections/RsCastToBoolInspectionTest.kt
Undin: It's exactly how we test absence of errors. If `RsCastToBoolInspection` start produce error in this case, test will fail

src/test/kotlin/org/rust/ide/inspections/RsCastToBoolInspectionTest.kt
zeroeightysix: Oops. I used #713 as reference (where this style was not used), as it was listed in the contributing page [here](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#contributing). Might want to update those to more recent PRs, to prevent cases like these.

src/main/kotlin/org/rust/ide/inspections/RsCastToBoolInspection.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.rust.lang.utils.addToHolder
+
+class RsCastToBoolInspection : RsLocalInspectionTool() {
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitCastExpr(castExpr: RsCastExpr) {
+            // It is allowed to cast a bool to a bool, so if the expression's type is of bool, ignore this cast.
+            // Casts from non primitive types (and the unit type) emit another error, so we ignore those as well.
+            val exprType = castExpr.expr.type
+            if (exprType === TyBool || exprType !is TyPrimitive || exprType is TyUnit) return
```
Undin, line 24:
```suggestion
            if (exprType === TyBool || exprType !is TyPrimitive || exprType == TyUnit) return
```

Undin (commented):
LGTM
Could you squash "fixing" commits in main ones?

zeroeightysix:
> Could you squash "fixing" commits in main ones?

Sure, but I'm not sure what you mean by 'main ones'. Do you want me to group commits *by tag*, so we get something like
```
INSP: Added cast to bool inspection.
T: Added tests for cast to bool inspection
DOC: Added zeroeightysix to CONTRIBUTORS.md
```

Or squash all into one commit (with or without (multiple) tags)?

Undin:
@zeroeightysix we don't have strict rules about number/structure of commits. We just prefer not to have separate commits with review fixes in master because they have a little sense there. Other decisions about commit history are up to you. 

> INSP: Added cast to bool inspection.
> T: Added tests for cast to bool inspection
> DOC: Added zeroeightysix to CONTRIBUTORS.txt

is ok for me

zeroeightysix:
All done :+1:

Undin (approved):
Thank you!

bors r+

Undin:
bors retry

PR #6330, opened 2020-10-30
INT: Compare with zero instead when casting numerical to bool

zeroeightysix:
Builds on #6318.

This PR adds a fix to *Compare with zero instead* when a cast of a numerical type to bool is attempted.

Part of #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/CompareWithZeroFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/CompareWithZeroFix.kt
@@ -1,0 +1,33 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsCastExpr
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.types.ty.TyNumeric
+import org.rust.lang.core.types.type
+
+class CompareWithZeroFix private constructor(expr: RsCastExpr) : LocalQuickFixAndIntentionActionOnPsiElement(expr) {
+    override fun getFamilyName(): String = "Compare with zero"
+
+    override fun getText(): String = familyName
+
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -471,6 +471,9 @@ class RsPsiFactory(
         else -> createExpressionOfType("${expr.text}.$methodNameText()")
     }
 
+    fun createNotEqZero(lhs: RsExpr): RsBinaryExpr =
```
Kobzol, line 474: I think that this is may be too specific to put into `RsPsiFactory`. You can just call `createExpression` and cast the result to `RsBinaryExpr`.

src/main/kotlin/org/rust/ide/annotator/fixes/CompareZeroFix.kt
```diff
@@ -0,0 +1,36 @@
...
+    override fun invoke(project: Project, file: PsiFile, editor: Editor?, startElement: PsiElement, endElement: PsiElement) {
+        if (startElement !is RsCastExpr) return
+        startElement.replace(RsPsiFactory(project).createExpression("${startElement.expr.text} != 0"))
+    }
+
+    companion object {
+        fun createIfCompatible(expression: RsCastExpr): CompareZeroFix? {
+            return if (expression.expr.type is TyNumeric) {
+                CompareZeroFix(expression)
+            } else null
```
Undin, line 33:
```suggestion
            return if (expression.expr.type is TyNumeric) CompareZeroFix(expression) else null
```

src/main/kotlin/org/rust/ide/annotator/fixes/CompareZeroFix.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsCastExpr
+import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.types.ty.TyNumeric
+import org.rust.lang.core.types.type
+
+class CompareZeroFix private constructor(expr: RsExpr) : LocalQuickFixAndIntentionActionOnPsiElement(expr) {
+    override fun getFamilyName() = "Compare with zero instead"
```
Undin, line 20: I think `instead` is redundant in context of quick fix that supposed to modify code by default

Undin (commented):
I general looks ok but comments

Undin (approved):
Great!
Thank you!

bors r+

Issue #6282, opened 2020-10-21
Suggest trait items in impl block while completion


renoth:
## Environment

* 0.3.133.3426-202
* 1.47.0
* IDEA 2020.2.3 Ultimate
* Windows 8

## Problem description

When doing the yew tutorial i tried to get an autocomplete for a method in Component of yew but there was no autocomplete suggestions at all. I was thinking it should  suggest the methods Component is having to implement them in my Class.

![image](https://user-images.githubusercontent.com/4069007/96768141-c5abc180-13dd-11eb-8e59-bcc767a3dff2.png)

For example Component has these two methods which i think should be suggested when autocompleting

![image](https://user-images.githubusercontent.com/4069007/96768388-23400e00-13de-11eb-873d-4f922a5ddb8a.png)

## Steps to reproduce

Create new Project with rust and yew. Try to get autocomplete suggestion.

panstromek:
The issue is on point, but I just wanted to mention that you can invoke `Implement Methods` dialog with `ctrl+i` or `alt+insert | Implement Methods` and plugin will generate boilerplate for the ones you choose.

renoth:
Thanks for the hint, it works well but not as easy as autocomplete suggestion.

PR #6342, opened 2020-11-02
COMP: add completion for impl trait items

Kobzol:
This PR adds completion for types, constants and functions in `impl Trait for Foo` blocks.
complete-impl-items

I'm not sure what should be done with the caret/selection after the completion. Using `RsInPlaceVariableIntroducer` is probably not right in completion? But without it if I select the completed value and the user presses enter, it will be overwritten.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6282

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
@@ -1,0 +1,153 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.patterns.ElementPattern
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.parentOfType
+import com.intellij.util.ProcessingContext
+import org.rust.ide.presentation.ImportingPsiRenderer
+import org.rust.ide.presentation.PsiRenderingOptions
+import org.rust.ide.presentation.renderFunctionSignature
+import org.rust.ide.presentation.renderTypeReference
+import org.rust.ide.utils.import.import
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsAbstractable
+import org.rust.lang.core.psi.ext.block
+import org.rust.lang.core.psi.ext.expandedMembers
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.resolve.ref.pathPsiSubst
+import org.rust.lang.core.types.BoundElement
+import org.rust.lang.core.types.RsPsiSubstitution
+import org.rust.lang.core.types.infer.substitute
+import org.rust.lang.core.types.type
+import org.rust.openapiext.selectElement
+
+object RsImplTraitMemberCompletionProvider : RsCompletionProvider() {
+    override val elementPattern: ElementPattern<PsiElement>
+        get() {
+            val contributor = RsKeywordCompletionContributor()
+            return contributor.traitOrImplDeclarationPattern()
+        }
+
+    override fun addCompletions(
+        parameters: CompletionParameters,
+        context: ProcessingContext,
+        result: CompletionResultSet
+    ) {
+        val element = parameters.position
+        val implBlock = element.parentOfType<RsImplItem>() ?: return
+
+        val traitRef = implBlock.traitRef ?: return
+
+        val trait = implBlock.implementedTrait ?: return
+        val subst = pathPsiSubst(traitRef.path, trait.element)
+
+        val members = trait.element.members ?: return
+
+        val parentItems = members.expandedMembers.toMutableSet()
+        for (item in implBlock.expandedMembers) {
+            parentItems.removeIf { it.javaClass == item.javaClass && it.name == item.name }
+        }
+
+        for (item in parentItems) {
+            val lookup = getCompletion(item, trait, implBlock, subst) ?: continue
+            result.addElement(
+                lookup.withPriority(KEYWORD_PRIORITY + 1)
+            )
+        }
+    }
+}
+
+private fun getCompletion(
+    target: RsAbstractable,
+    trait: BoundElement<RsTraitItem>,
+    impl: RsImplItem,
+    substitution: RsPsiSubstitution
+): LookupElementBuilder? {
+    return when (target) {
+        is RsConstant -> completeConstant(target, trait, impl, substitution)
+        is RsTypeAlias -> completeType(target)
+        is RsFunction -> completeFunction(target, impl, substitution)
+        else -> error("unreachable")
+    }
+}
+
+private fun completeConstant(
+    target: RsConstant,
+    trait: BoundElement<RsTraitItem>,
+    impl: RsImplItem,
+    substitution: RsPsiSubstitution
+): LookupElementBuilder? {
+    val renderer = ImportingPsiRenderer(PsiRenderingOptions(renderGenericsAndWhere = true), substitution, impl)
+    val text = buildString {
+        append("const ")
+        append(target.name)
+        val typeRef = target.typeReference ?: return null
+        append(": ")
+        append(renderer.renderTypeReference(typeRef))
+        val builder = RsDefaultValueBuilder(impl.knownItems, target.containingMod, RsPsiFactory(target.project))
+        append(" = ")
+        val type = typeRef.type.substitute(trait.subst)
+        append(builder.buildFor(type, mapOf()).text)
+        append(';')
+    }
+    return LookupElementBuilder.create(text)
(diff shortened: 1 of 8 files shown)


Kobzol:
I changed the completion so that it now completes the whole abstractable (i.e. `const FOO: u32 = 0;`), just like in Kotlin. It is however only available at the beginning of a statement, I'm not sure what's the proper way to make it work both at P1 and P2:
```rust
impl Trait for () {
   /*P1*/const /*P2*/
}
```

vlad20012 (approved):
Wow, cool! I have not seen this PR until now =/

Could you please rebase & squash (since the second commit removes everything that the first adds)?

I think we should merge this almost as is. Then, if you want to make it complete an item after a keyword (in P2), you can take a look at Kotlin's [OverridesCompletionLookupElementDecorator](https://github.com/JetBrains/kotlin/blob/caa852061a38f8c7f9491c1900ccfce6a64ffc09/idea/idea-frontend-independent/src/org/jetbrains/kotlin/idea/completion/OverridesCompletionLookupElementDecorator.kt#L24)

Kobzol:
> Wow, cool! I have not seen this PR until now =/

Yeah, I still have a lot of surprises [in store](https://github.com/intellij-rust/intellij-rust/pulls/kobzol) :laughing:

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
```diff
@@ -0,0 +1,127 @@
...
+    target: RsFunction,
+    trait: BoundElement<RsTraitItem>,
+    impl: RsImplItem
+): LookupElementBuilder {
+    val substitution = trait.subst
+    val signature = target.getSignatureText(substitution, trait, impl) ?: ""
+    val text = "$signature{\n        unimplemented!()\n    }"
+    return LookupElementBuilder
+        .create(text)
+        .withIcon(RsIcons.FUNCTION)
```
Undin, line 120: Probably, we need to reuse `org.rust.lang.core.psi.ext.RsFunctionImplMixin#getIcon` and its friends in `RsConstantImplMixin` and `RsTypeAliasImplMixin` somehow (I suppose we need to extract the corresponding code into the corresponding util methods) to use the same icon style

Kobzol: Maybe we can keep it simpler, because we know that the completion will only be in a certain situation?
The owner will always be a trait, the abstractable will therefore always be public and non-abstract. Therefore it should be `RsIcons.ASSOC_TYPE_ALIAS` for type aliases, `RsIcons.ASSOC_CONSTANT`  for constants and functions should be either `RsIcons.ASSOC_FUNCTION` or `RsIcons.METHOD`.

But if you want, I can also extract these methods. What should I use as the flags in that case? `Iconable.ICON_FLAG_VISIBILITY | Iconable.ICON_FLAG_READ_STATUS`?

Undin: > Maybe we can keep it simpler, because we know that the completion will only be in a certain situation?
The owner will always be a trait, the abstractable will therefore always be public and non-abstract. 

Ideally, I'd prefer a single place for such logic because if we decide to change something related to icons in the future, it would be great to modify only single place

> But if you want, I can also extract these methods. What should I use as the flags in that case? Iconable.ICON_FLAG_VISIBILITY | Iconable.ICON_FLAG_READ_STATUS?

We use `0` during completion (see `org.rust.lang.core.completion.LookupElementsKt#getLookupElementBuilder`) so I suppose it's ok to use `0` here as well

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
```diff
@@ -0,0 +1,127 @@
...
+        .withInsertHandler { context, _ ->
+            val element = context.getElementOfType<RsFunction>() ?: return@withInsertHandler
+            val body = element.block?.expr ?: return@withInsertHandler
+            selectElement(body, context.editor)
+        }
+        .withPresentableText("$signature { ... }")
```
Undin, line 126: Do we actually want to use raw `$signature` as presentable text?
It may produce quite weird text
image

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
Kobzol: It's straightforward to use the icon of the parent item (https://github.com/intellij-rust/intellij-rust/pull/6342/commits/08aecd6a413fb5fc5957ec1bb44b5c019668d444). I'm not sure if it's better to show the parent or the completed item icon.

But for the latter, to reuse `getIcon` properly, we would probably need to construct a fake PSI element on the completed location and then get it's icon. It sounds hacky. Or we would need to refactor `getIcon` so that it doesn't need an existing PSI element.

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
```diff
@@ -0,0 +1,127 @@
...
+    val signature = target.getSignatureText(substitution, trait, impl) ?: ""
+    val text = "$signature{\n        unimplemented!()\n    }"
+    return LookupElementBuilder
+        .create(text)
+        .withIcon(target.getIcon(0))
+        .withInsertHandler { context, _ ->
+            val element = context.getElementOfType<RsFunction>() ?: return@withInsertHandler
+            val body = element.block?.expr ?: return@withInsertHandler
+            selectElement(body, context.editor)
+        }
```
Undin, line 125: What's about auto import here?
For example, try to complete `take` method of `std::io::Read`
Currently, `Take` is not imported
image
But don't insist to do it in this PR

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
Undin: Yeah, I thought about icon for completed item that's why I wrote about some refactoring. But actually, I don't think it's important enough to make a lot of changes so let's use parent icon for now

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
Kobzol: Good catch. What about ignoring `where` clauses (and maybe also generics, which can also be quite long)? By doing something like this:
```kotlin
fun RsFunction.getSignatureText(
    subst: Substitution,
    trait: BoundElement<RsTraitItem>,
    impl: RsImplItem,
    withWhereClause: Boolean = true,
    withGenerics: Boolean = true
): String? {
```

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
Kobzol: Thanks, I added it in https://github.com/intellij-rust/intellij-rust/pull/6342/commits/3d0807248d7e77a730f2fcf8e5c7393e3d32f2f8.

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
Undin: I've checked Java and Kotlin implementations
Kotlin does nothing with too long signatures, Java always skips generic "section"

Kotlin:
image
Java:
image

Since Rust generics are quite long quite often, let's indeed skip generic and where parts.

Regarding suggesting `withWhereClause` and `withGenerics` - let's use a single flag (at least, for now) because currently we are not going to render only `where` clause or only generic parameters

Kobzol:
Added shorter preview for functions with generic parameters and `where` clauses and squashed commits.

vlad20012:
Let's merge it after #7393 because this conflicts with it

Kobzol:
I changed the implementation so that it uses the new `RsPsiSubstitution`. It's very nice!

I wonder if we could reuse `MembersGenerator` here instead of generating the items manually :thinking:

vlad20012:
> I wonder if we could reuse `MembersGenerator` here instead of generating the items manually :thinking:

Yes, I think it's possible if we extract member text generation to separate methods. But let's finally merge it. 

bors r+

vlad20012:
Oh, stop, one test failed 

bors r-

Kobzol:
@vlad20012 Sorry for the delay, I was on a vacation. The test fails with the old resolve, so I changed it so that the tests use the new resolve engine. Why is the old resolve engine even tested on CI anymore? :thinking:

vlad20012:
:+1: 
bors r+

> Why is the old resolve engine even tested on CI anymore? thinking

Hmm, indeed, I have no idea

Undin:
> Why is the old resolve engine even tested on CI anymore?

Because:
- we still have a user-visible option to disable the new engine
- old name resolution engine is still used in some places, like functions bodies and files without crate root

Kobzol:
bors retry

PR #6317, opened 2020-10-28
Run tests with new resolve on CI

dima74:
Add another job into `check` workflow to run tests with [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217) enabled.

Also add annotation `@IgnoreInNewResolve` to temporary disable few tests if running with new resolve.

Issue #6217, opened 2020-10-08
Name resolution 2.0


dima74:
New name resolution algorithm (aka name resolution 2.0) is an effort to rewrite part of the name resolution code in order to better support complex glob-imports/reexports and [resolve](https://github.com/rust-lang/rfcs/blob/master/text/1561-macro-naming.md) of [macros 2.0](https://github.com/rust-lang/rfcs/blob/master/text/1584-macros.md).

## Name resolution 1.0
Current name resolution algorithm (aka name resolution 1.0) uses [traditional lazy approach](https://rust-analyzer.github.io/blog/2020/07/20/three-architectures-for-responsive-ide.html#map-reduce), which roughly can be described as "going scopes upwards". It works in most cases and is pretty fast. The main downside is that it is some approximation, not necessarily correct by any means.

## Name resolution 2.0
New name resolution algorithm consists of two parts:
* Collect phase, where for each crate we build a special structure called `CrateDefMap`. This structure for each module of the crate stores a set of all items available in this module. It includes explicitly declared items, imported items, and items expanded from macros.
  * To built `CrateDefMap` we first collect all explicitly declared items, imports and macro calls. Then iteratively expand macros and resolve imports, until something changes. This is called fixed-point iteration algorithm which is used in both `rustc` and `rust-analyzer`.
  * Fixed-point iteration is not incremental, so we have to rebuild `CrateDefMap` after any change which might affect resolve (e.g. adding new item)
* Resolve phase, where we use built `CrateDefMap` to do actual resolve.

Please note that new resolve will be used only for top-level items (so resolve of local variables or anything related to type inference is not changed by new resolve).

dima74:
Next steps after #6241 merged:

- [ ] Tests
  - [ ] Add resolve tests which work only with new resolve (#6297)
  - [ ] Run all existing resolve tests on CI (#6317)
  - [ ] Add tests for specific functionality of new resolve (caching/updating `DefMap`s)
- [ ] Memory optimizations
  - [ ] Store `DefMap`s under soft references
  - [ ] Unite `PerNs` and `VisItem` when there is only one item (with namespace) with specific name
  - [ ] Research why half of all `VisItem`s are duplicated (equal, but different objects)
  - [ ] (?) Save `DefMap`s to disk
- [ ] Add option to settings for enabling new resolve
- [ ] Filter/sort private items in completion
- [ ] Propagate expanded macro def to child modules
- [ ] (?) Expand macros at collect phase, if possible

PR #6348, opened 2020-11-03
RES: Support resolve of attribute and bang proc macros in new resolve

dima74:
Support resolve of attribute and bang proc macros in [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217)

Depends on #6317

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
@@ -154,6 +154,17 @@ val RsFunction.isCustomDeriveProcMacroDef: Boolean
 val RsFunction.isProcMacroDef: Boolean
     get() = greenStub?.isProcMacroDef ?: (queryAttributes.isProcMacroDef)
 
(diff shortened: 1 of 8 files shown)


Undin (commented):
@dima74 tests?

dima74:
@Undin there already exists `RsProcMacroResolveTest` (will be enabled in CI for new resolve in #6317). Do you think we need specific proc macro tests for new resolve?

Undin:
@dima74 I usually expect that the feature and its tests are located in the same PR. AFAIK @lancelote also uses our tests to identify what was done and what's need to be tested

dima74:
@Undin ok, so lets first merge #6317 which will add `@IgnoreInNewResolve` annotation to proc macro tests, and then I will update this PR to remove `@IgnoreInNewResolve` annotation from part of proc macro tests, so it will clear which tests are supported by this PR

vlad20012 (approved):
Well, let's remove `@IgnoreInNewResolve` from `RsProcMacroResolveTest`?

dima74:
bors r=vlad20012

PR #6353, opened 2020-11-04
INSP: Constants cannot refer to statics. (E0013)

zeroeightysix:
This PR aims to implement an inspection for rust error [E0013](https://doc.rust-lang.org/error-index.html#E0013) *A const variable cannot refer to a static variable.*

Covering the following [const contexts](https://doc.rust-lang.org/reference/const_eval.html#const-context):
- [X] Constant variables
- [x] Constant size of array
- [x] Constant enum item discriminant

Part of #886

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsConstReferStaticInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsConstReferStaticInspection.kt
@@ -1,0 +1,43 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.RsConstContextKind
+import org.rust.lang.core.psi.ext.classifyConstContext
+import org.rust.lang.core.psi.ext.isConst
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.lang.utils.addToHolder
+
+/**
+ * Inspection that detects the E0013 error.
+ */
+class RsConstReferStaticInspection : RsLocalInspectionTool() {
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitPathExpr(pathExpr: RsPathExpr) {
+            val constContext = pathExpr.classifyConstContext
+            if (constContext != null) {
+                checkPathInConstContext(holder, pathExpr.path, constContext)
+            }
+            super.visitPathExpr(pathExpr)
+        }
+
+        override fun visitBaseType(o: RsBaseType) {
+            val path = o.path
+            if (path != null) {
+                checkPathInConstContext(holder, path, RsConstContextKind.ConstGenericArgument)
+            }
+            super.visitBaseType(o)
+        }
+    }
+
+    private fun checkPathInConstContext(holder: RsProblemsHolder, path: RsPath, constContext: RsConstContextKind) {
+        val ref = path.reference?.resolve() as? RsConstant ?: return
+        if (!ref.isConst) {
+            RsDiagnostic.ConstItemReferToStaticError(path, constContext).addToHolder(holder)
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/inspections/RsConstReferStaticInspection.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.rust.lang.core.psi.*
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.lang.utils.addToHolder
+
+class RsConstReferStaticInspection : RsLocalInspectionTool() {
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitConstant(constant: RsConstant) {
+            // We only want to target `const` constants (thus, excluding `static`) with this inspection.
+            if (constant.const == null) return
```
Undin, line 18:
```suggestion
            if (!constant.isConst) return
```

src/main/kotlin/org/rust/ide/inspections/RsConstReferStaticInspection.kt
```diff
@@ -0,0 +1,51 @@
...
+            }
+        }
+    }
+
+    private fun errorOnStaticRef(expr: RsExpr, holder: RsProblemsHolder, problemFactory: (RsPathExpr) -> RsDiagnostic) {
+        expr.forEachDescendantOfType<RsPathExpr> { pathExpr ->
+            // If the resolved reference is a constant,
+            val ref = pathExpr.path.reference?.resolve() as? RsConstant ?: return@forEachDescendantOfType
+            // and it is declared `static`,
+            if (ref.static != null) {
```
Undin, line 44:
```suggestion
            if (!ref.isConst) {
```

Undin (commented):
Just small suggestions

Note, there are merge conflicts. Could you rebase into master and fix conflicts, please?

mchernyavsky (changes requested):
Also, check this cases:

```rust
#![feature(const_generics)]

static S: usize = 42;

fn foo<const C: usize>() {}

fn main() {
    foo::<{ S }>();
    [0; S];
}
```

src/main/kotlin/org/rust/ide/inspections/RsConstReferStaticInspection.kt
```diff
@@ -0,0 +1,72 @@
...
+            errorOnStaticRef(enumVariant.variantDiscriminant?.expr ?: return, holder) { expr ->
+                RsDiagnostic.ConstEnumDiscriminantReferToStaticError(expr, name, expr.text)
+            }
+        }
+
+        override fun visitCallExpr(call: RsCallExpr) {
+            val typeArgumentList = (call.expr as? RsPathExpr ?: return).path.typeArgumentList ?: return
+            val constParams = ((call.expr as RsPathExpr).path.reference?.resolve() as RsFunction).typeParameterList?.constParameterList
+            typeArgumentList.exprList.forEach {
+                // TODO: Find out which const param this expression targets, and skip if it is not const.
```
zeroeightysix, line 52: I'm not sure how to do this.
Consider the following:
```rust
static X: usize = 5;

fn foo<A, const C: usize>() {}

fn main() {
    foo::<bool, X>();
}
```
in `foo::<bool, X>`, `X` is the second element in `typeArgumentList`. 
However, it matches the first entry in `constParams`. How do I reliably figure out whether or not this specific expression is supposed to be `const`?
Is the order always assured to be the same?

mchernyavsky: Sorry for the long response!

For now, we parse `X` in `foo::<bool, X>()` as a type reference, so without fixing the grammar we cannot handle this case. You can try to support `foo::<bool, { X }>()` / `foo::<{ X }, bool>()` cases - here `{ X }` is parsed as an expression, so it will be the first in the `constArguments` (declared in`RsPath.kt`) in both cases.

Kobzol:
@zeroeightysix Asked for another review round on gitter. @mchernyavsky could you please take a look?

zeroeightysix:
> @zeroeightysix Asked for another review round on gitter. @mchernyavsky could you please take a look?

Mostly because I needed to confirm if the last review is still relevant with all the changes that have been made to const generics in intellij-rust.

mchernyavsky:
@zeroeightysix Hello! Sorry for the long response. We have recently merged some improvements for const generics, so I updated your commits - now the PR handles const arguments that look like type arguments in `RsTypeParameterList` + some minor changes.

vlad20012:
I refactored the inspection and also enabled it for `const fn`

mchernyavsky (approved):
Thanks!

bors r+

PR #6354, opened 2020-11-04
RES: Store CrateDefMap under soft reference

dima74:
Memory optimization for [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217): store `CrateDefMap`s under soft references, so they will be cleared if IDE is low on memory. If `CrateDefMap` is cleared, then it will be rebuild next time it is needed.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve2/DefMapService.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve2/DefMapService.kt
@@ -17,6 +17,8 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.psi.PsiManager
 import com.intellij.psi.PsiTreeChangeEvent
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/lang/core/resolve2/RsDefMapSoftReferenceTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun test() = doTest("""
+        fn foo() {}
+         //X
+        fn main() {
+            foo();
+        } //^
+    """)
+}
+
+private fun forceClearAllSoftReferences() {
```
vlad20012, line 39: A too naive approach, haha! Let's use `GCWatcher` for this. Let's also move the test to `rustSlowTests` directory

vlad20012:
:+1: 
bors r+

PR #6356, opened 2020-11-05
DBG: Update BTreeMap/BTreeSet GDB pretty-printers

### LINKED ISSUE #5722: GDB BTreeMap/BTreeSet pretty printers don't work since 1.44
## Environment

* **IntelliJ Rust plugin version:** 0.2.125.3191-201
* **Rust toolchain version:** 1.44.1 (c7087fe00 2020-06-17) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.2 (CL-201.7846.88)
* **Operating system:** Linux 4.15.0-109-generic

## Problem description

Since rust 1.44.0 debugger shows nothing for `BTreeMap` and `BTreeSet` collections

| 1.43.0 | 1.44.1 |
| - | - |
| Screenshot from 2020-07-11 16-28-47 | Screenshot from 2020-07-11 16-28-01 |
## Steps to reproduce
```rust
use std::collections::{BTreeMap, BTreeSet};

fn main() {
    let mut tree_map = BTreeMap::new();
    let mut tree_set = BTreeSet::new();
    for i in 1..100 {
        tree_map.insert(i.to_string(), i);
        tree_set.insert(i.to_string());
    }
}
```

ortem:
Update `BTreeMap`/`BTreeSet` GDB pretty-printers to the recent rustc's version.

* Fixes #5722

* Handle zero-sized keys or values https://github.com/rust-lang/rust/pull/77788
I also added a fix to the original commit from rustc: `gdb.parse_and_eval("()")` instead of simple `"()"` in `children_of_node` method. This is needed because GDB treats `"()"` as a Rust array of two characters, not as a unit

After merge:
- [ ] Upstream a fix for zero-sized keys or values to rustc

### DIFF
--- a/pretty_printers_tests/tests/btree.rs
+++ b/pretty_printers_tests/tests/btree.rs
@@ -1,5 +1,4 @@
-// min-version: 1.33.0
-// max-version: 1.43.0
+// min-version: 1.34.0
 
 // === GDB TESTS ===================================================================================
 
@@ -11,6 +10,11 @@
 // gdb-command: print btree_map
 // gdbg-check:$2 = size=15 = {[0] = 0, [1] = 1, [2] = 2, [3] = 3, [4] = 4, [5] = 5, [6] = 6, [7] = 7, [8] = 8, [9] = 9, [10] = 10, [11] = 11, [12] = 12, [13] = 13, [14] = 14}
 
+// gdb-command: print zst_btree_map
+// gdbg-check:$3 = size=1 = {[()] = ()}
+
+// gdb-command: print zst_btree_set
+// gdbg-check:$4 = size=1 = {()}
 
 use std::collections::{BTreeMap, BTreeSet};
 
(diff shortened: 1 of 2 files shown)


pretty_printers_tests/tests/btree.rs
```diff
@@ -24,5 +25,9 @@ fn main() {
     for i in 0..15 {
         btree_map.insert(i, i);
     }
+
+    let mut zst_btree_map: BTreeMap<(), ()> = BTreeMap::new();
```
Undin, line 29: Maybe it worth to add the same test for `BTreeSet` as well?

Undin (commented):
The current version of pretty printers doesn't work with rust 1.33. It would be great to make it works with it if possible. Otherwise, let's update `min-version` for btree tests with proper value

pretty_printers_tests/tests/btree.rs
ortem: Added

ortem:
@Undin I've updated the minimum Rust version for the `BTreeMap`/`BTreeSet` pretty-printers tests to 1.34.0. Not sure if it's worth the time finding a way to change the code in order to support such an old version

Undin (approved):
bors r+

Undin:
bors retry

PR #6360, opened 2020-11-06
CI: batch regression tests

Undin:
It should reduce common overhead for downloading/compilation and reduce numbers of CI instances required to run regression tests. I've checked in my own fork and all three tests took less than 35 min (currently, every single run takes about 17 minutes)

Note, now regression tests use 203 platform because I use junit 4 to generate all necessary tests in runtime and 202 platform is not compatible with junit 4 tests.

### DIFF
--- a/.github/workflows/regressions.yml
+++ b/.github/workflows/regressions.yml
@@ -39,35 +39,28 @@ jobs:
 
     check:
         needs: [ calculate-base-commit ]
-        name: ${{ matrix.project.name }}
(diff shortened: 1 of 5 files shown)


Undin:
This [patch](https://github.com/intellij-rust/intellij-rust/files/5503327/Add_regression.patch.zip) can be used to check how it works now when someone adds regressions

lancelote:
bors merge

vlad20012 (approved):

Issue #6361, opened 2020-11-07
Not Recognizing min_const_generics Feature


nevalackin:
Rust plugin: 0.3.134.3455-202 (Latest)
Rust: 1.49.0 nightly (Latest)
IDE: 2020.2.3 IntelliJ CE (Latest)
OS: Win 10 64bit

The rust plugin does not recognize the `#![feature(min_const_generics)]` attribute and still shows errors despite compiling fine.
![Example](https://user-images.githubusercontent.com/59190138/98429292-cc347b80-20f5-11eb-9935-de69ce563f50.png)
It tells me to enable the `const_generics` feature but I already have `min_const_generics` and as such this code is fine.

Use const generics with the 
`#![feature(min_const_generics)]` 
attribute instead of 
`#![feature(const_generics)]`

Example(s) of issue in my codebase:
https://github.com/nevalackin/win_mem/blob/e4d43d48cc4ce57e7e753ef5465b771612880e93/src/utils.rs#L21

PR #6364, opened 2020-11-07
ANN: Support `min_const_generics` feature

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/6361.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -25,8 +25,7 @@ import org.rust.ide.refactoring.findBinding
 import org.rust.ide.utils.isCfgUnknown
 import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.core.*
-import org.rust.lang.core.FeatureAvailability.CAN_BE_ADDED
-import org.rust.lang.core.FeatureAvailability.NOT_AVAILABLE
+import org.rust.lang.core.FeatureAvailability.*
 import org.rust.lang.core.RsPsiPattern.STD_ATTRIBUTES
 import org.rust.lang.core.macros.MacroExpansionMode
 import org.rust.lang.core.macros.macroExpansionManager
@@ -578,7 +577,7 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
     }
 
     private fun checkConstParameter(holder: RsAnnotationHolder, constParameter: RsConstParameter) {
-        CONST_GENERICS.check(holder, constParameter, "const generics")
+        checkConstGenerics(holder, constParameter)
         checkDuplicates(holder, constParameter)
     }
 
@@ -1212,6 +1211,38 @@ private fun checkDuplicates(
         }
     }
     message.addToHolder(holder)
+}
+
+private fun checkConstGenerics(holder: RsAnnotationHolder, constParameter: RsConstParameter) {
+    val constGenericAvailability = CONST_GENERICS.availability(constParameter)
+    if (constGenericAvailability == AVAILABLE) return
+
+    val version = constParameter.cargoProject?.rustcInfo?.version
+    val feature = when (constParameter.typeReference?.type) {
+        is TyInteger, is TyBool, is TyChar -> {
+            val current = version?.semver
+            val since = MIN_CONST_GENERICS.since
+            if (current != null &&
+                since != null &&
+                current.isGreaterOrEqualThan(since.major, since.minor, since.patch)) {
+                MIN_CONST_GENERICS
+            } else {
+                CONST_GENERICS
+            }
+        }
+        else -> CONST_GENERICS
+    }
+
+    if (feature == CONST_GENERICS &&
+        constGenericAvailability == CAN_BE_ADDED &&
+        MIN_CONST_GENERICS.availability(constParameter) == AVAILABLE) {
+        val typeReference = constParameter.typeReference ?: return
+        val message = RsDiagnostic.ForbiddenConstGenericType(typeReference)
+        message.addToHolder(holder)
+        return
+    }
+
+    feature.check(holder, constParameter, "const generics")
 }
 
 private fun checkParamAttrs(holder: RsAnnotationHolder, o: RsOuterAttributeOwner) {
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -8,12 +8,16 @@ package org.rust.lang.utils
 import com.intellij.codeInsight.intention.IntentionAction
 import com.intellij.codeInspection.InspectionManager
 import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
 import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import com.intellij.openapi.util.text.StringUtil.pluralize
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
 import com.intellij.xml.util.XmlStringUtil.escapeString
 import org.rust.ide.annotator.RsAnnotationHolder
 import org.rust.ide.annotator.RsErrorAnnotator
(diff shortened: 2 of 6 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1148,6 +1147,15 @@ private fun checkDuplicates(
...
 }
 
+private fun checkConstGenerics(holder: RsAnnotationHolder, constParameter: RsConstParameter) {
+    if (CONST_GENERICS.availability(constParameter) == AVAILABLE
+        || MIN_CONST_GENERICS.availability(constParameter) == AVAILABLE) return
+    val current = constParameter.cargoProject?.rustcInfo?.version?.semver
+    val since = MIN_CONST_GENERICS.since
+    val feature = if (current != null && since != null && current >= since) MIN_CONST_GENERICS else CONST_GENERICS
+    feature.check(holder, constParameter, "const generics")
+}
```
Undin, line 1157: Note, `min_const_generics` feature is only about integers, `bool` and `char`. More complex options are still under `const_generics`. Also, `min_const_generics` and `const_generics` cannot be used together, so I think in case, when `min_const_generics` exists but code requires `const_generics`, we should replace `min_const_generics` with `const_generics`

```rust
#[derive(Eq, PartialEq)]
struct S(i32);
fn foo<const N: S>() {}
```

we should warn about `const_generics` instead of `min_const_generics`

Undin (changes requested):

mchernyavsky:
@Undin Done.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1148,6 +1152,44 @@ private fun checkDuplicates(
...
+    feature.check(holder, constParameter, "const generics", *listOfNotNull(fix).toTypedArray())
+}
+
+private fun createReplaceFeatureFix(element: RsElement, newFixName: String, oldFixName: String): LocalQuickFix? {
+    val project = element.project
+    val crateRoot = element.crateRoot
+    val oldAttr = RsFeatureIndex.getFeatureAttributes(project, oldFixName)
+        .find { it.crateRoot == crateRoot }
+        ?: return null
+    return object : LocalQuickFixAndIntentionActionOnPsiElement(element) {
```
Undin, line 1182: Can we write test(s) for this quick-fix?

mchernyavsky: 👍

aldanor:
Thanks, neat! Looking forward to this very much, with 1.50 on the horizon there's more and more const generics code popping up all over the place.

Undin (changes requested):
We still have two quick-fixes:
- Add `const_generic` feature
- Replace `min_const_generics` feature with `const_generic` feature

image

that makes users choose what to do. At the same time, the first quick-fix produces a compilation error.

Also, if you have `min_const_generic` feature and try to use some custom type, the compiler produces more concrete error with more precise range:

```rust
#![feature(min_const_generics)]

#[derive(PartialEq, Eq)]
struct Foo;

fn foo<const N: Foo>() {}
```

```
error: `Foo` is forbidden as the type of a const generic parameter
  --> src/main.rs:28:17
   |
28 | fn foo<const N: Foo>() {}
   |                 ^^^
   |
   = note: the only supported types are integers, `bool` and `char`
   = help: more complex types are supported with `#[feature(const_generics)]`
```

So I think we should:
- provide "const generics is experimental [E0658]" error and "Add `min_const_generic` feature"/ "Add `const_generic` feature" quick-fix if there isn't any feature related to const generics and a user tries to use const generics (quick-fixes only on nightly, of course)
- provide "`%TypeName%` is forbidden as the type of a const generic parameter" error annotation if `min_const_generic` is available (via the corresponding feature or when it will be stabilized) with "Add `const_generic` feature" quick fix on nightly that will replace `min_const_generic` if needed

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1318,6 +1325,17 @@ sealed class RsDiagnostic(
...
     }
+
+    class ForbiddenConstGenericType(
+        private val typeReference: RsTypeReference
+    ) : RsDiagnostic(typeReference) {
+        override fun prepare(): PreparedAnnotation = PreparedAnnotation(
+            ERROR,
+            null,
+            "`${typeReference.text}` is forbidden as the type of a const generic parameter",
+            fixes = listOfNotNull(createReplaceFeatureFix(typeReference, CONST_GENERICS.name, MIN_CONST_GENERICS.name))
```
Undin, line 1336: We should provide this quick fix only for nightly compiler

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1455,3 +1473,21 @@ private fun getConflictingNames(element: PsiElement, vararg tys: Ty): Set<RsQual
         emptySet()
     }
 }
+
+private fun createReplaceFeatureFix(element: RsElement, newFixName: String, oldFixName: String): LocalQuickFix? {
+    val project = element.project
+    val crateRoot = element.crateRoot
+    val oldAttr = RsFeatureIndex.getFeatureAttributes(project, oldFixName)
+        .find { it.crateRoot == crateRoot }
+        ?: return null
```
Undin, line 1482: It won't work when `min_const_generics` will be stabilized because there won't be `#![feature(min_const_generics)]` anywhere

Undin: I suppose in such cases we should provide `AddFeatureAttributeFix` fix instead

Undin (changes requested):

mchernyavsky:
@Undin Fixed.

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1330,6 +1337,17 @@ sealed class RsDiagnostic(
             fixes = listOfNotNull(MakePublicFix.createIfCompatible(exportedItem, exportedItem.name, false))
         )
     }
+
+    class ForbiddenConstGenericType(
+        private val typeReference: RsTypeReference
+    ) : RsDiagnostic(typeReference) {
+        override fun prepare(): PreparedAnnotation = PreparedAnnotation(
+            ERROR,
+            null,
```
Undin, line 1346:
```suggestion
            null,
            "the only supported types are integers, `bool` and `char`",
```

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -2829,6 +2829,51 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
...
+    """)
+
+    @MockRustcVersion("1.47.0-nightly")
+    fun `test min const generics E0658 2`() = checkErrors("""
+        #![feature(min_const_generics)]
+        fn f<const C: i32>() {}
+        struct S<const C: i32>(A);
+        trait T<const C: i32> {}
+        enum E<const C: i32> {}
+    """)
```
Undin, line 2847: This test fails now because `since` version for `min_const_generics` was changed after updating compiler features. I suggest not trying to support such false positives for old nightly toolchain. It seems that a proper solution requires retrieving compiler features from the actual compiler that's not possible for now.
Let's just check that for `1.51.0-nightly` we don't show error annotation either with `min_const_generics ` feature or without it

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -1214,6 +1215,32 @@ private fun checkDuplicates(
...
+private fun checkConstGenerics(holder: RsAnnotationHolder, constParameter: RsConstParameter) {
+    val availability = CONST_GENERICS.availability(constParameter)
+    if (availability == AVAILABLE) return
+
+    val version = constParameter.cargoProject?.rustcInfo?.version
+    val feature = when (constParameter.typeReference?.type) {
+        is TyInteger, is TyBool, is TyChar -> {
+            val current = version?.semver
+            val since = MIN_CONST_GENERICS.since
+            if (current != null && since != null && current >= since) MIN_CONST_GENERICS else CONST_GENERICS
```
Undin, line 1227:
```suggestion
            if (current != null && since != null && current.isGreaterOrEqualThan(since.major, since.minor, since.patch)) MIN_CONST_GENERICS else CONST_GENERICS
```

Also, the same fix we need to add into `org.rust.lang.core.CompilerFeature#availability` to support case when you use `1.51.0-nightly` and feature requires at least `1.51.0`. In this case, we shouldn't show any error annotation

Undin (approved):
In general look ok
Let's fix new minor comments and merge it

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
Undin: If you know better fix here, you are welcome

mchernyavsky (commented):
bors r=Undin

Undin:
@mchernyavsky some tests fail

mchernyavsky (commented):
bors r=Undin

mchernyavsky:
bors retry

Issue #6288, opened 2020-10-24
Regex highlighting and intentions not working inside lazy_static! macro


chylex:
## Environment

* **IntelliJ Rust plugin version:** 0.3.133.3426-202
* **Rust toolchain version:** 1.47.0
* **IDE name and version:** CLion 2020.2
* **Operating system:** Windows 10

## Problem description
Putting regexes into a `lazy_static!` macro is a recommended pattern for global regexes, so it'd be nice if they were highlighted and had intentions (such as *Edit RegExp Fragment*) like regexes outside the macro:

![obrazek](https://user-images.githubusercontent.com/3685160/97083909-adbd8300-1613-11eb-85c3-8a6e23101550.png)

## Steps to reproduce
Cargo:
```toml
[dependencies]
regex = "1.4.1"
```
Code:
```rust
lazy_static! {
	static ref FOLDER_NAME_REGEX: Regex = Regex::new(r"(?i)^([a-z]+)([0-9.]+)$").unwrap();
}
	
let folder_name_regex2: Regex = Regex::new(r"(?i)^([a-z]+)([0-9.]+)$").unwrap();
```

panstromek:
Side question: is the regex highlighting for Regex crate supposed to work automatically or do you need to always inject it manually? I always have to inject it..

chylex:
For me, it immediately starts highlighting as I type `Regex::new("")` or `Regex::new(r"")`, I did not have to inject it manually. I checked `Settings - Editor - Language Injections` and there is no built-in injection for Rust, but it somehow works automatically (as long as I'm typing outside the macro).

panstromek:
huh, I've actually found it, it's the path prefix:

`regex::Regex::new(r#",(?:\s+,)+"#)` -> this is not highlighted

`Regex::new(r#",(?:\s+,)+"#)` -> this is highlighted

Undin:
@panstromek I've filed a separate issue for your case #6312

PR #6369, opened 2020-11-08
GRAM: allow language injections in string literals inside macro calls

### LINKED ISSUE #4844: Macros language injection
I want to inject language to a macros string, but there is no such option (https://www.jetbrains.com/help/clion/using-language-injections.html). 
image
How may I do that?

vlad20012:
Enables `inject language or reference` intention action in strings inside macro calls. Fixes #4844. Part of #6288.


abonander:
@vlad20012 This is amazing work! Do you know of any way to modify a macro _definition_ to turn on language injection by default? Currently it only works at a specific invocation, like so: https://cdn.discordapp.com/attachments/665528275556106243/777980521799745536/unknown.png

I've tried all of the following but to no avail (I want language injection on `$query`):

```rust
// language=SQL
macro_rules! query (
    (
        // language=SQL
        $query:expr
    ) => ({
        // language=SQL
        $crate::sqlx_macros::expand_query!(source = $query)
    });
    (
        // language=SQL
        $query:expr,
        $($args:tt)*
    ) => ({
        // language=SQL
        $crate::sqlx_macros::expand_query!(source = $query, args = [$($args)*])
    })
);
```

Issue #6366, opened 2020-11-08
Handle missing type for `const` item error as semantic error


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.135.3464-202-nightly
* **Rust toolchain version:** 1.49.0-nightly (a601302ff 2020-11-06) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.2.3 Ultimate Edition (IU-202.7660.26)
* **Operating system:** macOS 10.15.6

## Problem description
Currently, the plugin Rust grammar requires explicit type for `const` items. And when the type is missing, the plugin shows a parsing error with not quite user-friendly error message. At the same time, the compiler shows a semantic error about missing type and with the proper error message.

It should be great to relax our grammar to parse const items without type and detect this error in [`RsSyntaxErrorsAnnotator`](https://github.com/intellij-rust/intellij-rust/blob/c24e3fb404811087cec9c308e2375df1cfc32928/src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt#L145-L168). Also, the plugin should provide a quick fix to add explicit type if it can infer the type of the corresponding expression.

Plugin error message:
![image](https://user-images.githubusercontent.com/2539310/98464198-46d2c900-21d2-11eb-9bd3-1f5968480262.png)

Compiler error message:
![image](https://user-images.githubusercontent.com/2539310/98464201-4fc39a80-21d2-11eb-9b19-711294f05b37.png)

## Steps to reproduce
```rust
const BUFFER_HEIGHT = 25;
```

PR #6372, opened 2020-11-09
ANN: add check that constants have a type and quick fix that adds it

Kobzol:
add-type

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6366

changelog: Detect the missing type of `const` and `static` items as a semantic error and produce a user-friendly error message. Also, provide a quick-fix to add explicit type in such cases


src/main/grammars/RustParser.bnf
```diff
@@ -604,7 +604,7 @@ PatField ::= PatFieldFull | box? PatBinding
 
 BindingMode ::= ref mut? | mut
 
-upper Constant ::= (static mut? | const) (identifier | '_') TypeAscription [ '=' AnyExpr ] ';' {
+upper Constant ::= (static mut? | const) (identifier | '_') TypeAscription? [ '=' AnyExpr ] ';' {
```
Undin, line 607: Don't forget to increment `org.rust.lang.core.parser.RustParserDefinition#PARSER_VERSION` to update all nesseary indixes after plugin update

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -165,6 +170,27 @@ private fun checkConstant(holder: AnnotationHolder, const: RsConstant) {
...
+        val typeText = if (element.static != null) { "static" } else { "const" }
+        val message = "Missing type for `$typeText` item"
+
+        val annotation = holder.newAnnotation(HighlightSeverity.ERROR, message)
+            .range(element.textRange)
+
+        val expr = element.expr
+        if (expr != null) {
+            val type = expr.type
+            if (!type.containsTyOfClass(TyUnknown::class.java)) {
```
Undin, line 187: Maybe it make sense to suggest this quick fix even final type is not fully known by the plugin?
A user should insert it anyway and in case of partially known type the plugin may insert significant part of the correct type (especially, if type is complex). Also, if it will be completely wrong, a user just replace it since we use template builder
What do you think?

Undin (commented):
In general looks ok!

Just small nitpick: we usually use `GRAM` prefix for changes in grammar instead of `BNF`

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
Kobzol: When the type is `TyUnknown`, it's probably faster to type `:` than to invoke the quick fix and then delete the unknown type. But you're right that the current check might be too strict. So just check if it isn't `TyUnknown` and otherwise offer it? That sounds reasonable.

Undin: > When the type is TyUnknown, it's probably faster to type : than to invoke the quick fix and then delete the unknown type.

Probably, it's indeed faster. But an additional quick fix doesn't interfere with manual code modification at all. Also, I don't know other quick fixes/intentions that can interfere in a situation when constant doesn't have explicit type. So I'd prefer just always add this quick fix (it also simplifies code a little bit) if you don't have strong objections

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -165,6 +169,31 @@ private fun checkConstant(holder: AnnotationHolder, const: RsConstant) {
             require(const.expr, holder, "$title must have a value", const)
         }
     }
+    checkConstantType(holder, const)
+}
+
+private fun checkConstantType(holder: AnnotationHolder, element: RsConstant) {
+    if (element.colon == null && element.typeReference == null) {
+        val typeText = if (element.static != null) {
```
Undin, line 177: Let's use `element.isConst` here

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
Kobzol: Fair enough, it's better to be consistent! I removed the check for `TyUnknown` :+1:

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -43,6 +45,7 @@ class RsSyntaxErrorsAnnotator : AnnotatorBase() {
             is RsValueParameter -> checkValueParameter(holder, element)
             is RsTypeParameterList -> checkTypeParameterList(holder, element)
             is RsTypeArgumentList -> checkTypeArgumentList(holder, element)
+            is RsConstant -> checkConstantType(holder, element)
```
Undin, line 48: It seems this branch is unreachable since `RsConstant` is `RsItemElement`

Undin (approved):
LGTM except comment
Thank you!

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
Kobzol: Indeed, I forgot to remove it. Thank you :)

Kobzol:
bors r=undin

PR #6374, opened 2020-11-09
CI: update nightly rust version

Undin:
Also, update compiler features and lints


Undin:
@avrong could you check why `CratesLocalIndexServiceTest` fails with nightly toolchain, please?

avrong:
Nightly cargo started to load and update the registry index using only `origin/HEAD` and ignoring `origin/master`. It seems like this was introduced in rust-lang/cargo#8364.

Can be fixed by fetching the registry index using the `origin/HEAD` instead of the `origin/master`. It is likely to work with both new and previous cargo versions, as previous versions of cargo are updating both tags

PR #6378, opened 2020-11-10
RES: Add option to settings to enable new resolve

dima74:
Add option to settings to enable [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217)

TODO: Prepare some text for changelog

PR #6382, opened 2020-11-11
RES: Fix name of option in settings to enable new resolve

dima74:
Fix name of [option to enable new resolve](https://github.com/intellij-rust/intellij-rust/pull/6378)

bors r+


dima74:
bors retry

PR #6383, opened 2020-11-11
PERF: Store dollar crate indices in macro expansion cache

dima74:
It slightly optimizes [CrateDefMap](https://github.com/intellij-rust/intellij-rust/issues/6217) building


src/main/kotlin/org/rust/lang/core/macros/MacroExpander.kt
```diff
@@ -346,7 +346,7 @@ class MacroExpander(val project: Project) {
     }
 
     companion object {
-        const val EXPANDER_VERSION = 13
+        const val EXPANDER_VERSION = 14
```
vlad20012, line 349: Let's increment a version in MacroExpansionSharedCache (add `+ 1`) instead of modifying this version that also affects other stuff

vlad20012 (approved):

dima74:
bors r=vlad20012

PR #6387, opened 2020-11-13
Update GitHub pull request template

Undin:
Add note about changelog description in template.
Description is supposed to be used to create changelog template.
See https://github.com/intellij-rust/intellij-rust.github.io/pull/98


dima74 (approved):
Probably it is also worth to add a note that `changelog:` lines must be at the end of PR text

Undin:
> Probably it is also worth to add a note that changelog: lines must be at the end of PR text

@dima74 why? our script can properly extract the corresponding paragraph even `changelog:` is located in the middle of PR description

dima74:
@Undin oh, indeed, thank you

bors r+

PR #6389, opened 2020-11-13
T: improve error messages in `RsMacroExpansionTest`

vlad20012:
`MacroExpander.expandMacroAsText`  returned a nullable value for a long time. I made something like Rust's `Result` and return some kind of `Err(MacroExpansionError(...))` instead of `null`. This simplifies `MacroExpander` debugging (e.g. #6380)


vlad20012:
bors retry

PR #6390, opened 2020-11-13
CARGO: Implement crates local index based on cargo registry index

avrong:
Initial implementation of crates local index inside the plugin.

It gives the ability to retrieve crate names, versions, and features right from the crates index stored on disk and generated from the local cargo registry index. This way code analysis can be performed faster and with better quality and more complete results.

changelog: Added experimental `Cargo.toml` dependency completion based on local cargo registry index. It can be enabled with `org.rust.crates.local.index` experimental flag

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -73,6 +73,7 @@ allprojects {
         mavenCentral()
         jcenter()
         maven("https://dl.bintray.com/jetbrains/markdown")
+        maven("http://download.eclipse.org/jgit/maven")
     }
 
     idea {
@@ -376,6 +377,7 @@ project(":toml") {
     dependencies {
         implementation(project(":"))
         implementation(project(":common"))
+        implementation("org.eclipse.jgit:org.eclipse.jgit:5.9.0.202009080501-r") { exclude("org.slf4j") }
         testImplementation(project(":", "testOutput"))
         testImplementation(project(":common", "testOutput"))
     }
--- a/toml/src/main/kotlin/org/rust/toml/completion/CargoTomlCompletionContributor.kt
+++ b/toml/src/main/kotlin/org/rust/toml/completion/CargoTomlCompletionContributor.kt
@@ -7,6 +7,8 @@ package org.rust.toml.completion
 
 import com.intellij.codeInsight.completion.CompletionContributor
 import com.intellij.codeInsight.completion.CompletionType.BASIC
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
 import org.rust.toml.CargoTomlPsiPattern.inDependencyKeyValue
 import org.rust.toml.CargoTomlPsiPattern.inDependencyPackageFeatureArray
 import org.rust.toml.CargoTomlPsiPattern.inFeatureDependencyArray
@@ -14,18 +16,28 @@ import org.rust.toml.CargoTomlPsiPattern.inKey
 import org.rust.toml.CargoTomlPsiPattern.inSpecificDependencyHeaderKey
 import org.rust.toml.CargoTomlPsiPattern.inSpecificDependencyKeyValue
 import org.rust.toml.CargoTomlPsiPattern.inValueWithKey
+import org.rust.toml.crates.local.completion.DependencyCratesLocalCompletionProvider
 import org.rust.toml.tomlPluginIsAbiCompatible
 
 class CargoTomlCompletionContributor : CompletionContributor() {
     init {
+        val isCratesLocalIndexEnabled = isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)
+
+        val dependencyCompletionProvider = if (isCratesLocalIndexEnabled) {
+            DependencyCratesLocalCompletionProvider()
+        } else {
+            CargoTomlDependencyCompletionProvider()
+        }
+
         if (tomlPluginIsAbiCompatible()) {
             extend(BASIC, inKey, CargoTomlKeysCompletionProvider())
             extend(BASIC, inValueWithKey("edition"), CargoTomlKnownValuesCompletionProvider(listOf("2015", "2018")))
-            extend(BASIC, inDependencyKeyValue, CargoTomlDependencyCompletionProvider())
             extend(BASIC, inSpecificDependencyHeaderKey, CargoTomlSpecificDependencyHeaderCompletionProvider())
             extend(BASIC, inSpecificDependencyKeyValue, CargoTomlSpecificDependencyVersionCompletionProvider())
             extend(BASIC, inFeatureDependencyArray, CargoTomlFeatureDependencyCompletionProvider())
             extend(BASIC, inDependencyPackageFeatureArray, CargoTomlDependencyFeaturesCompletionProvider())
+
+            extend(BASIC, inDependencyKeyValue, dependencyCompletionProvider)
         }
     }
 }
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexCachesInvalidator.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexCachesInvalidator.kt
@@ -1,0 +1,15 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import com.intellij.ide.caches.CachesInvalidator
+import com.intellij.openapi.components.service
+
+class CratesLocalIndexCachesInvalidator : CachesInvalidator() {
+    override fun invalidateCaches() {
+        service<CratesLocalIndexService>().invalidateCaches()
+    }
+}
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
@@ -1,0 +1,274 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import com.google.gson.Gson
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.components.PersistentStateComponent
+import com.intellij.openapi.components.State
+import com.intellij.openapi.components.Storage
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.diagnostic.logger
+import com.intellij.openapi.progress.ProgressIndicator
+import com.intellij.openapi.progress.Task
+import com.intellij.util.io.*
+import org.eclipse.jgit.lib.ObjectId
+import org.eclipse.jgit.lib.Repository
+import org.eclipse.jgit.revwalk.RevTree
+import org.eclipse.jgit.revwalk.RevWalk
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder
+import org.eclipse.jgit.treewalk.TreeWalk
+import org.rust.openapiext.pluginDirInSystem
+import org.rust.stdext.cleanDirectory
+import org.rust.toml.crates.local.CratesLocalIndexService.Companion.CratesLocalIndexState
+import java.io.*
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Crates local index, created from user cargo registry index on host machine.
+ * Used for dependency code insight in project's `Cargo.toml`.
+ * Stores crates info in [crates] persistent hash map and hash for commit which has been used for index load in
+ * persistent state [CratesLocalIndexState].
+ */
+@State(name = "CratesLocalIndexState", storages = [Storage("rust.crateslocalindex.xml")])
+class CratesLocalIndexService : PersistentStateComponent<CratesLocalIndexState>, Disposable {
+    private val userCargoIndexDir: Path
+        get() = Paths.get(System.getProperty("user.home"), CARGO_REGISTRY_INDEX_LOCATION)
+
+    // TODO: handle RepositoryNotFoundException
+    private val repository: Repository = FileRepositoryBuilder()
+        .setGitDir(userCargoIndexDir.toFile())
+        .build()
+
+    private val registryHeadCommitHash: String = repository.resolve("origin/master")?.name ?: run {
+        LOG.error("Git revision string cannot be resolved to any object id")
+        INVALID_COMMIT_HASH
+    }
+
+    private val crates: PersistentHashMap<String, CargoRegistryCrate>? = run {
+        resetIndexIfNeeded()
+
+        val file = baseCratesLocalRegistryDir.resolve("crates-local-index")
+        try {
+            IOUtil.openCleanOrResetBroken({
+                PersistentHashMap(
+                    file,
+                    EnumeratorStringDescriptor.INSTANCE,
+                    CrateExternalizer,
+                    4 * 1024,
+                    CRATES_INDEX_VERSION
+                )
+            }, file)
+        } catch (e: IOException) {
+            LOG.error("Cannot open or create PersistentHashMap in $file", e)
+            null
+        }
+    }
+
+    @Volatile
+    private var state: CratesLocalIndexState = CratesLocalIndexState()
+
+    private val isReady: AtomicBoolean = AtomicBoolean(true)
+
+    override fun getState(): CratesLocalIndexState = state
+    override fun loadState(state: CratesLocalIndexState) {
+        this.state = state
+    }
+
+    fun isReady(): Boolean = isReady.get()
+
+    fun getCrate(crateName: String): CargoRegistryCrate? {
+        updateIfNeeded()
+        if (crates == null) return null
+
+        return try {
+            crates.get(crateName)
+        } catch (e: IOException) {
+            LOG.error("Failed to get crate $crateName", e)
+            null
+        }
+    }
+
+    fun getAllCrateNames(): List<String> {
+        updateIfNeeded()
+        if (crates == null) return emptyList()
+
+        val crateNames = mutableListOf<String>()
+
+        try {
+            crates.processKeys { name ->
+                crateNames.add(name)
+            }
+        } catch (e: IOException) {
+            LOG.error("Failed to get crate names", e)
+        }
+
+        return crateNames
+    }
+
+    private fun updateIfNeeded() {
+        if (state.indexedCommitHash != registryHeadCommitHash && isReady.compareAndSet(true, false)) {
+            CratesLocalIndexUpdateTask().queue()
+        }
+    }
+
+    private fun resetIndexIfNeeded() {
+        if (corruptionMarkerFile.exists()) {
+            baseCratesLocalRegistryDir.cleanDirectory()
+        }
+    }
+
+    fun invalidateCaches() {
+        corruptionMarkerFile.apply {
+            parent?.createDirectories()
+            Files.createFile(this)
+            state = CratesLocalIndexState(INVALID_COMMIT_HASH)
+        }
+    }
+
+    private fun reloadCrates(revTree: RevTree) {
+        TreeWalk(repository).use { treeWalk ->
+            treeWalk.addTree(revTree)
+            treeWalk.isRecursive = true
+            treeWalk.isPostOrderTraversal = false
+
+            while (treeWalk.next()) {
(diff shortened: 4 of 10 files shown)


Kobzol:
Nice! I wonder how it relates to this: https://github.com/intellij-rust/intellij-rust/pull/5988? Maybe we can combine your index cache with my crate resolver? Or is my PR redundant after this?

avrong:
@Kobzol I think our plan is to use both local index and network API depending on the environment and tools the user has. `CrateResolver` in your PR can contribute to it, so in the near future, we will see how we can combine them

build.gradle.kts
```diff
@@ -376,6 +377,8 @@ project(":toml") {
     dependencies {
         implementation(project(":"))
         implementation(project(":common"))
+        implementation("org.eclipse.jgit:org.eclipse.jgit:5.9.0.202009080501-r") { exclude("org.slf4j") }
+        implementation("com.google.code.gson:gson:2.8.6")
```
Undin, line 381: The platform has dependency jars of gson and jackson libraries so I don't think we should add our own version of gson into plugin

toml/src/main/kotlin/org/rust/toml/completion/crates/local/CratesLocalIndexState.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package org.rust.toml.completion.crates.local
+
+import com.intellij.openapi.components.PersistentStateComponent
+import com.intellij.openapi.components.State
+import com.intellij.openapi.components.Storage
+import com.intellij.util.xmlb.XmlSerializerUtil
+
+@State(name = "CratesLocalIndexState", storages = [Storage("rust.crateslocalindex.xml")])
+class CratesLocalIndexState : PersistentStateComponent<CratesLocalIndexState> {
```
ortem, line 14: Probably it's better to make `CratesLocalIndexService : PersistentStateComponent` and put `CratesLocalIndexState` as a data class inside the service. Discussed in person.

toml/src/main/kotlin/org/rust/toml/completion/crates/local/LocalCargoTomlDependencyCompletionProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+class LocalCargoTomlDependencyCompletionProvider : TomlKeyValueCompletionProviderBase() {
+    override fun completeKey(keyValue: TomlKeyValue, result: CompletionResultSet) {
+        val prefix = CompletionUtil.getOriginalElement(keyValue.key)?.text ?: return
+
+        val indexService = service<CratesLocalIndexService>()
+        if (!indexService.isReady) return
+
+        val crateNames = indexService.getAllCrateNames()
+
+        result.withPrefixMatcher(CargoNormalizedNamesPrefixMatcher(prefix)).addAllElements(crateNames.mapNotNull { crateName ->
```
ortem, line 31: Let's extract `crateNames.mapNotNull { crateName -> ... }` to a variable

toml/src/main/kotlin/org/rust/toml/completion/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,212 @@
...
+        get() = Paths.get(
+            System.getProperty("user.home"),
+            ".cargo/registry/index/github.com-1ecc6299db9ec823/.git/"
+        )
+
+    private val repository: Repository = FileRepositoryBuilder()
+        .setGitDir(userCargoIndexDir.toFile())
+        .build()
+
+    private var crates: PersistentHashMap<String, CargoRegistryCrate>
```
ortem, line 48: Looks like mutability is not necessary here. Consider something like
```Kotlin
private val crates: PersistentHashMap<String, CargoRegistryCrate> = run {
    checkInvalidatedStorage()
    PersistentHashMap(
        baseCratesLocalRegistryDir.resolve("crates-local-index"),
        EnumeratorStringDescriptor.INSTANCE,
        CrateExternalizer
    )
}
```

toml/src/main/kotlin/org/rust/toml/completion/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,212 @@
...
+
+    private fun updateIfNeeded() {
+        if (updateIsNeeded) {
+            updateIsNeeded = false
+            isReady = false
+            CratesLocalIndexUpdateTask().queue()
+        }
+    }
+
+    fun checkInvalidatedStorage() {
```
ortem, line 86: Probably something like this would be more clear because it's not just a check but also an invalidation action
```suggestion
    private fun invalidateStorageIfNeeded() {
```

toml/src/main/kotlin/org/rust/toml/completion/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,212 @@
...
+                val loader = repository.open(objectId)
+                val versions = mutableListOf<CargoRegistryCrateVersion>()
+                val reader = BufferedReader(InputStreamReader(loader.openStream()))
+
+                reader.forEachLine { line ->
+                    if (line.isBlank()) return@forEachLine
+
+                    try {
+                        val parsedVersion = gson.fromJson(line, ParsedVersion::class.java)
+                        versions.add(CargoRegistryCrateVersion(parsedVersion.vers, parsedVersion.features.map { it.key }))
```
ortem, line 130: Let's add `fun fromJson(json: String): CargoRegistryCrateVersion` to `CargoRegistryCrateVersion`'s companion object (and incapsulate `ParsedVersion` class inside the companion). Also, looks like `gson` used only once and can be inlined

toml/src/main/kotlin/org/rust/toml/completion/crates/local/LocalCargoTomlDependencyCompletionProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+            val crate = indexService.getCrate(crateName)!!
```
ortem, line 32: `!!` looks dangerous here because `getCrate` potentially changes `PersistentHashMap` contents due to `updateIfNeeded` call inside

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlCompletionContributor.kt
```diff
@@ -7,25 +7,36 @@ package org.rust.toml.completion
...
 class CargoTomlCompletionContributor : CompletionContributor() {
     init {
         if (tomlPluginIsAbiCompatible()) {
             extend(BASIC, inKey, CargoTomlKeysCompletionProvider())
             extend(BASIC, inValueWithKey("edition"), CargoTomlKnownValuesCompletionProvider(listOf("2015", "2018")))
-            extend(BASIC, inDependencyKeyValue, CargoTomlDependencyCompletionProvider())
             extend(BASIC, inSpecificDependencyHeaderKey, CargoTomlSpecificDependencyHeaderCompletionProvider())
             extend(BASIC, inSpecificDependencyKeyValue, CargoTomlSpecificDependencyVersionCompletionProvider())
             extend(BASIC, inFeatureDependencyArray, CargoTomlFeatureDependencyCompletionProvider())
-            extend(BASIC, inDependencyPackageFeatureArray, CargoTomlDependencyFeaturesCompletionProvider())
```
ortem, line 28: Did you accidentally delete this line?

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,233 @@
...
+import java.io.DataInput
+import java.io.DataOutput
+import java.io.InputStreamReader
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.util.concurrent.atomic.AtomicReference
+
+@State(name = "CratesLocalIndexState", storages = [Storage("rust.crateslocalindex.xml")])
+class CratesLocalIndexService : PersistentStateComponent<CratesLocalIndexService.Companion.CratesLocalIndexState>, Disposable {
```
ortem, line 34: Let's add an import for `org.rust.toml.crates.local.CratesLocalIndexService.Companion.CratesLocalIndexState` (there is an intention in case of `CratesLocalIndexService.Companion./*<caret>*/CratesLocalIndexState`)

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,233 @@
...
+
+    private val corruptionMarkerFile: Path
+        get() = baseCratesLocalRegistryDir.resolve(CORRUPTION_MARKER_NAME)
+
+    private val userCargoIndexDir: Path
+        get() = Paths.get(System.getProperty("user.home"), CARGO_REGISTRY_INDEX_LOCATION)
+
+    private val repository: Repository = FileRepositoryBuilder()
+        .setGitDir(userCargoIndexDir.toFile())
+        .build()
```
ortem, line 46: Probably we should handle `RepositoryNotFoundException` somehow

ortem (changes requested):
I haven't reviewed all the code yet, to be continued

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    private inner class CratesLocalIndexUpdateTask : Task.Backgroundable(null, "Updating cargo registry index", false) {
+        override fun run(indicator: ProgressIndicator) {
+            val branch = repository.resolve("origin/master") ?: return
+            val revCommit = RevWalk(repository).parseCommit(ObjectId.fromString(branch.name))
+            reloadCrates(revCommit.tree)
+        }
+
+        override fun onSuccess() {
+            state.indexedCommitHash = repository.resolve("origin/master")?.name ?: return
```
ortem, line 160: There are two problems:
1. `indexedCommitHash` is not volatile, but there are write accesses from different threads
2. Last repository's commit might be fresher than `registryHeadCommitHash`, and so `state.indexedCommitHash != registryHeadCommitHash` might be always true (in `updateIfNeeded`)

ortem: So I'd recommend to use
```suggestion
            state = CratesLocalIndexState(registryHeadCommitHash)
```

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    private fun resetIndexIfNeeded() {
+        if (corruptionMarkerFile.exists()) resetIndex()
+    }
+
+    private fun resetIndex() {
+        baseCratesLocalRegistryDir.cleanDirectory()
+        state.indexedCommitHash = ""
```
ortem, line 91:
```suggestion
state = CratesLocalIndexState()
```
since `indexedCommitHash` field is not volatile (but `state` is volatile)

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,236 @@
...
+    override fun getState(): CratesLocalIndexState = _state.get()
+    override fun loadState(state: CratesLocalIndexState) {
+        _state.set(state)
+    }
+
+    private fun updateIfNeeded() {
+        if (state.indexedCommitHash != registryHeadCommitHash && isReady) {
+            isReady = false
+            CratesLocalIndexUpdateTask().queue()
+        }
```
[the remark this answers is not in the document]
ortem: The logic is "If the indexed commit is outdated and the index is ready, mark the index as not ready and update it". `compareAndSet` is needed because two threads might enter this section simultaneously, but we want to update the index **only once** (consider a situation when two threads checked that `isReady` is `false` and try to update the index twice).

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexCachesInvalidator.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package org.rust.toml.crates.local
+
+import com.intellij.ide.caches.CachesInvalidator
+import com.intellij.openapi.components.service
+
+class CratesLocalIndexCachesInvalidator : CachesInvalidator() {
+    override fun invalidateCaches() {
+        service<CratesLocalIndexService>().invalidateCaches()
```
ortem, line 13: Looks like `serviceIfCreated` should be used instead

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlCompletionContributor.kt
```diff
@@ -7,25 +7,38 @@ package org.rust.toml.completion
...
+            val isCratesLocalIndexEnabled = isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)
+
+            extend(
+                BASIC, inDependencyKeyValue,
+
+                if (isCratesLocalIndexEnabled)
+                    DependencyCratesLocalCompletionProvider()
+                else
+                    CargoTomlDependencyCompletionProvider()
+            )
```
ortem, line 41: IMO something like that looks a bit clearer
```Kotlin
val dependencyCompletionProvider = if (isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
    DependencyCratesLocalCompletionProvider()
} else {
    CargoTomlDependencyCompletionProvider()
}
extend(BASIC, inKey, CargoTomlKeysCompletionProvider())
// ...
extend(BASIC, inDependencyKeyValue, dependencyCompletionProvider)
```

toml/src/main/kotlin/org/rust/toml/crates/local/completion/DependencyCratesLocalCompletionProvider.kt
```diff
@@ -0,0 +1,78 @@
...
+        if (!indexService.isReady) return
+
+        val versions = indexService.getCrate(name)?.versions ?: return
+        val elements = versions.mapIndexed { index, variant ->
+            var lookupElement = LookupElementBuilder.create(variant.version)
+                .withInsertHandler(StringValueInsertionHandler(keyValue))
+
+            if (variant.isYanked) {
+                lookupElement = lookupElement.withTailText(" yanked")
+            }
```
ortem, line 69:
```suggestion
            val lookupElement = LookupElementBuilder.create(variant.version)
                .withInsertHandler(StringValueInsertionHandler(keyValue))
                .withTailText(if (variant.isYanked) " yanked" else null)
```

Looks like this works fine as well

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,236 @@
...
+        get() = baseCratesLocalRegistryDir.resolve(CORRUPTION_MARKER_NAME)
+
+    private val userCargoIndexDir: Path
+        get() = Paths.get(System.getProperty("user.home"), CARGO_REGISTRY_INDEX_LOCATION)
+
+    private val repository: Repository = FileRepositoryBuilder()
+        .setGitDir(userCargoIndexDir.toFile())
+        .build()
+
+    private val registryHeadCommitHash: String? = repository.resolve("origin/master")?.name
```
ortem, line 50: We should handle `null` somehow, probably something like that (at least write to LOG)
```suggestion
    private val registryHeadCommitHash: String = repository.resolve("origin/master")?.name ?: run {
        LOG.error("Git revision string cannot be resolved to any object id")
        INVALID_COMMIT_HASH
    }
```
(not sure)

ortem (changes requested):
(Accidentally created a second review)
Please also try to reorder the `CratesLocalIndexService ` class members a bit. For example, `updateIfNeeded` used only in `getCrate` and `getAllCrateNames`, so probably it should be introduced right after them.

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlCompletionContributor.kt
avrong: I was thinking what if we would have several of them in the future. Then we will have to add a variable with the condition for each of the completion we implement both for API and Crates Local Index until e.g. crate resolver is implemented

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
avrong: I guess `INVALID_COMMIT_HASH` can be just an empty string

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,236 @@
...
+        if (corruptionMarkerFile.exists()) resetIndex()
+    }
+
+    private fun resetIndex() {
+        baseCratesLocalRegistryDir.cleanDirectory()
+        state.indexedCommitHash = ""
+    }
+
+    fun getCrate(crateName: String): CargoRegistryCrate? {
+        updateIfNeeded()
```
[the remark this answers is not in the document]
avrong: Made it `if (isReady() || crates == null) return null` to handle isReady inside the service

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexCachesInvalidator.kt
avrong: Sorry, I didn't really understand why

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
avrong: Maybe in the future here we could throw an exception to handle it inside, for instance, crate resolver

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
ortem: I don't think so since an empty string is already used as a default value for `indexedCommitHash` field in `CratesLocalIndexState`. It might be better to introduce a separate value inside the companion object, probably something like
```Kotlin
private const val INVALID_COMMIT_HASH: String = "<invalid>"
```

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
ortem: OK, let's leave something like `// TODO: handle RepositoryNotFoundException` for now

ortem (approved):
Generally LGTM, but to be improved in the future (discussed in person).
bors r+

Undin (commented):
It would be great to have tests for this

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,274 @@
...
+                )
+            }, file)
+        } catch (e: IOException) {
+            LOG.error("Cannot open or create PersistentHashMap in $file", e)
+            null
+        }
+    }
+
+    @Volatile
+    private var state: CratesLocalIndexState = CratesLocalIndexState()
```
Undin, line 75: Why do we have a mutable variable with mutable state? It makes `@Volatile` useless

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -0,0 +1,274 @@
...
+
+    private val isReady: AtomicBoolean = AtomicBoolean(true)
```
Undin, line 77: Is service intentionally ready after creation?

avrong: Yes, as it doesn't update the index right away. The current behavior is that when the index is not updating, then it is ready

Undin: Maybe use less confusion name for this field then? And/or add some doc comment about it

Issue #2158, opened 2017-12-27
Automatic sort of "use"s and clean-up of unused.


bltnkl:
Hi.
Could you please add the ability to clean-up unused "use"s? (By pressing Alt-Enter, for example).
Also automatic sorting of "use"'s would be really nice too.
Thanks.

Undin:
I think you mean something like `Optimize imports` action. Yeah, it would be great to support it)

farodin91:
Would like start working on it.

jatupan1989:
เยี่ยม

Kobzol:
Annotating and removing unused imports would be really nice, this is probably the most common Rust warning that I encounter on a daily basis. @Undin should this be implemented as a cached analysis pass akin to the liveness inspection? Since it's probably too expensive to search for use items usages in an error annotator without caching.

drewm1980:
"cargo fix" also cannot do this yet; this would be a welcome addition.

colin-grapl:
> "cargo fix" also cannot do this yet; this would be a welcome addition.

I believe `cargo fix` can do this, but IDE support would be idea, especially since `cargo fix` isn't configurable such that *only* the unused import warnings would be fixed; that is, it will fix all warnings, and you can not select which.

Kobzol:
@colin-grapl I also thought that `cargo fix` can remove them, but sadly it can't yet (https://github.com/rust-lang/cargo/issues/6278). But the plugin might be [able to do that](https://github.com/intellij-rust/intellij-rust/pull/7495) in the near future hopefully!

Kobzol:
@dima74 Can the inspection be enabled by default now or is do you want to wait for the missing things (e.g. usage in attribute macros)? Enabling it could help find more bugs, and it shouldn't be too annoying, since it's an inspection that can be easily turned off if needed.

dima74:
@Kobzol we need to be sure that there are no false-positives before enabling the inspection by default. Currently, known false-positives are caused by:
* usages in macro body (not macro call body)
* usages in strings, e.g. [`#[serde(default = "path")]`](https://serde.rs/field-attrs.html#default--path)
* usages in attribute macros - will be implemented after #7194

And yes, we should wait for `org.rust.macros.proc` to be enabled by default

PR #6392, opened 2020-11-14
INSP: add unused imports inspection

Kobzol:
This PR adds basic support for detecting unused `use` imports. I did not know which platform machinery should be used to implement this, so I took a little inspiration from the Kotlin plugin and implemented something that made sense to me to get the ball rolling. If it should be implemented in a different way, please let me know.

unused

~Currently, the inspection does not seem to run after every code change, I'm not sure what should be changed in the caching implementation, any hints?~

There are currently some false negatives (see last commented test), but these should be rare and I think that we can solve these in a future PR. If this gets merged, I'll add a quick fix to remove the unused use speck and then try to integrate this into the import optimizer to automate the process.

Related issues: https://github.com/intellij-rust/intellij-rust/issues/2158 https://github.com/intellij-rust/intellij-rust/issues/2259

changelog: TODO

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
@@ -1,0 +1,116 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.lints
+
+import com.intellij.openapi.util.Key
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.*
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.types.infer.ResolvedPath
+import org.rust.lang.core.types.inference
+import org.rust.openapiext.processElementsWithMacros
+
+data class PathUsageMap(
+    val pathUsages: Map<String, Set<RsElement>>,
+    val traitUsages: Set<RsTraitItem>
+)
+
+private val PATH_USAGE_KEY: Key<CachedValue<PathUsageMap>> = Key.create("PATH_USAGE_KEY")
+
+val RsItemsOwner.pathUsage: PathUsageMap
+    get() = CachedValuesManager.getCachedValue(this, PATH_USAGE_KEY) {
+        val usages = calculatePathUsages(this)
+        CachedValueProvider.Result.create(usages, PsiModificationTracker.MODIFICATION_COUNT)
+    }
+
+private fun calculatePathUsages(owner: RsItemsOwner): PathUsageMap {
+    val directUsages = hashMapOf<String, MutableSet<RsElement>>()
+    val traitUsages = hashSetOf<RsTraitItem>()
+
+    for (child in owner.children) {
+        processElementsWithMacros(child) { element ->
+            handleElement(element, directUsages, traitUsages)
+        }
+    }
+
+    return PathUsageMap(directUsages, traitUsages)
+}
+
+private fun handleElement(
+    element: PsiElement,
+    directUsages: MutableMap<String, MutableSet<RsElement>>,
+    traitUsages: MutableSet<RsTraitItem>
+): Boolean {
+    fun addItem(name: String, item: RsElement) {
+        directUsages.getOrPut(name) { hashSetOf() }.add(item)
+    }
+
+    if (!element.isEnabledByCfg) return false
+
+    return when (element) {
+        is RsModItem -> false
+        is RsPatIdent -> {
+            val name = element.patBinding.referenceName
+            val targets = element.patBinding.reference.multiResolve()
+            targets.forEach {
+                addItem(name, it)
+            }
+            true
+        }
+        is RsPath -> {
+            if (element.qualifier != null) {
+                val requiredTraits = getAssociatedItemRequiredTraits(element).orEmpty()
+                traitUsages.addAll(requiredTraits)
+            } else {
+                val useSpeck = element.parentOfType<RsUseSpeck>()
+                if (useSpeck == null || useSpeck.isTopLevel) {
+                    val name = element.referenceName ?: return true
+                    if (name in IGNORED_USE_PATHS) return true
+                    val targets = element.reference?.multiResolve().orEmpty()
+                    targets.forEach {
+                        addItem(name, it)
+                    }
+                }
+            }
+            true
+        }
+        is RsMethodCall -> {
+            val requiredTraits = getMethodRequiredTraits(element).orEmpty()
+            traitUsages.addAll(requiredTraits)
+            true
+        }
+        else -> true
+    }
+}
+
+private val IGNORED_USE_PATHS = listOf("crate", "self", "super")
+
+private fun getMethodRequiredTraits(call: RsMethodCall): Set<RsTraitItem>? {
+    val result = call.inference?.getResolvedMethod(call) ?: return null
+    return result.mapNotNull {
+        it.source.implementedTrait?.element
+    }.toSet()
+}
+
+private fun getAssociatedItemRequiredTraits(path: RsPath): Set<RsTraitItem>? {
+    val parent = path.parent as? RsPathExpr ?: return null
+    val resolved = path.inference?.getResolvedPath(parent) ?: return null
+    return resolved.mapNotNull {
+        if (it is ResolvedPath.AssocItem) {
+            it.source.implementedTrait?.element
+        } else null
+    }.toSet()
+}
+
+/**
+ * We should collect paths only from relative use specks,
+ * that is top-level use specks without `::`
+ * E.g. we shouldn't collect such paths: `use ::{foo, bar}`
+ */
+private val RsUseSpeck.isTopLevel: Boolean
+    get() = (path != null || coloncolon == null)
+        && parentOfType<RsUseSpeck>()?.isTopLevel != false
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
@@ -1,0 +1,91 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.lints
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.parentOfType
+import org.rust.ide.inspections.RsProblemsHolder
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve2.NamedItem
+import org.rust.lang.core.resolve2.exportedItems
+import org.rust.lang.core.resolve2.isNewResolveEnabled
+
+class RsUnusedImportInspection : RsLintInspection() {
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnusedImports
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitUseSpeck(o: RsUseSpeck) {
+            if (!holder.project.isNewResolveEnabled) return
+            val item = o.ancestorStrict<RsUseItem>() ?: return
+            if (item.isReexport) return
+            if (o.path?.resolveStatus != PathResolveStatus.RESOLVED) return
+
+            // Do not check uses if there is a child mod inside the current mod
+            val parentMod = o.containingMod
+            val hasChildMods = parentMod.expandedItemsExceptImplsAndUses.any {
+                it is RsModItem || it is RsModDeclItem
+            }
+            if (o.parentOfType<RsFunction>() == null && hasChildMods) return
+
+            checkUseSpeck(o, holder)
+        }
+    }
+
+    private fun checkUseSpeck(useSpeck: RsUseSpeck, holder: RsProblemsHolder) {
+        if (!useSpeck.isUsed()) {
+            val element = getHighlightElement(useSpeck)
+            holder.registerLintProblem(
+                element,
+                "Unused import: `${useSpeck.text}`"
+            )
+        }
+    }
+}
+
+private fun getHighlightElement(useSpeck: RsUseSpeck): PsiElement {
+    if (useSpeck.parent is RsUseItem) {
+        return useSpeck.parent
+    }
+    return useSpeck
+}
+
+/**
+ * Returns true if this use speck has at least one usage in its containing owner (module or function).
+ * A usage can be either a path that uses the import of the use speck or a method call/associated item available through
+ * a trait that is imported by this use speck.
+ */
+private fun RsUseSpeck.isUsed(): Boolean {
+    val owner = this.parentOfType<RsUseItem>()?.parent as? RsItemsOwner ?: return true
+    val usage = owner.pathUsage
+    return isUseSpeckUsed(this, usage)
+}
+
+private fun isUseSpeckUsed(useSpeck: RsUseSpeck, usage: PathUsageMap): Boolean {
+    // Use speck with an empty group is always unused
+    val useGroup = useSpeck.useGroup
+    if (useGroup != null && useGroup.useSpeckList.isEmpty()) return false
+
+    val items = if (useSpeck.isStarImport) {
+        val module = useSpeck.path?.reference?.resolve() as? RsMod ?: return true
+        module.exportedItems(useSpeck.containingMod)
+    } else {
+        val path = useSpeck.path ?: return true
+        val items = path.reference?.multiResolve() ?: return true
+        val name = useSpeck.itemName(withAlias = true) ?: return true
+        items.filterIsInstance<RsNamedElement>().map { NamedItem(name, it) }
+    }
+
+    // TODO: remove after macros are supported in RsPathUsageAnalysis.kt
+    if (items.any { (_, item) -> item is RsMacro }) {
+        return true
+    }
+
+    return items.any { (name, item) ->
+        item in usage.pathUsages[name].orEmpty()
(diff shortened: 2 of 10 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsElement.kt
```diff
@@ -138,3 +139,26 @@ fun RsElement.findInScope(name: String, ns: Set<Namespace>): PsiElement? {
...
+        // So we have to invalidate the cached value on any PSI change
+        containingFile.virtualFile is VirtualFileWindow -> CachedValueProvider.Result.create(value, PsiModificationTracker.MODIFICATION_COUNT)
+
+        // Invalidate cached value of code fragment on any PSI change
+        this is RsCodeFragment -> CachedValueProvider.Result.create(value, PsiModificationTracker.MODIFICATION_COUNT)
+
+        // CachedValueProvider.Result can accept a ModificationTracker as a dependency, so the
+        // cached value will be invalidated if the modification counter is incremented.
+        else -> {
+            val modificationTracker = contextOrSelf<RsModificationTrackerOwner>()?.modificationTracker
```
Undin, line 160: > Currently, the inspection does not seem to run after every code change, I'm not sure what should be changed in the caching implementation, any hints?

I haven't read the whole implementation yet but it's likely the reason of the described problem.
`RsModificationTrackerOwner` and `rustStructureModificationTracker` are supposed to be used for computation that don't depend on code inside function body. But "Unused import" functionality depends on it that leads to issue with incorrect updating. Just use `PsiModificationTracker.MODIFICATION_COUNT`

Kobzol: Thank you :)

Kobzol:
I changed the PSI modification tracker. I'm not sure about the performance implications, maybe we could create a new tracker that would only track changes to paths and method calls?

vlad20012:
I didn't look too closely, I just have 2 questions/tips right now:
1. what about child modules that use its parent private imports (like `use super::foobar;`)?
2. What about macros? Where is `RsWithMacrosRecursiveElementWalkingVisitor` btw

Also, please, be very careful with `*` imports. Our poor name resolution can lie a lot about them. Maybe this part could be enabled only if the new name resolution is enabled (but it is still far from ideal)

Kobzol:
Thanks for the tips. Handling usages in child modules is gonna be tricky.

I wonder if we could instead somehow get metadata from name resolution about which imports are actually used? It has to "walk through" them during name resolution after all.

dima74:
Regarding usages in child modules - with new name resolution engine we can relatively easily for each import determine whether or not it was used by some import in child mod (*). However I still see two problems:

---

Usages in child modules inside functions:
```rust
//- mod.rs
use foo::bar::func;
mod child;
//- child.rs
fn qqq() {
    super::func();
}
```
Afaik here `qqq` function body is not even parsed if we don't open `child.rs` file in editor

---

Naive implementation of (*), when encounter glob import from child mod to parent mod, will mark all imports in parent mod as used:
```rust
use foo::func1;
use foo::func2;
mod tests {
    use super::*;  // marks both imports as used, though only `func1` import is actually used
    fn test() {
        func1();
    }
}
```

Kobzol:
The second case is not that problematic, because it's a false negative, but the first one is... bad :( If the function bodies are not even parsed (makes sense for speed), then there's no way to tell if the import is used or not.

I have checked how Kotlin marks e.g. unused classes (which in theory could also only be used from inside function bodies). It seems that it actually uses text search to find potential usages quicker (at the cost of false negatives). Could this approach be also used here? It would probably require all star imports to never be marked as unused (but I would do that anyway, since they are problematic and not so widely used).

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
```diff
@@ -0,0 +1,117 @@
...
+        val name = useSpeck.itemName(withAlias = true) ?: return true
+        val usedItems = usage.pathUsages[name].orEmpty()
+
+        Pair(items.toSet(), usedItems)
+    }
+
+    return candidateItems.intersects(usedItems) || candidateItems.intersects(usage.traitUsages)
+}
+
+private fun RsMod.exportedItems(context: RsElement): List<NamedItem> {
```
dima74, line 94: We would better use new resolve to implement this function, because recursive approach will not work if there are cyclic glob imports. I will write implementation using new resolve

dima74 (commented):
Yes, using text search is a good idea! So, first we will find usages in current module (as you do currently), and then if no usages found, we will use text search to find usages in scope corresponding to import visibility (e.g. `pub(crate)` - scope is all crate). I propose to merge this PR for usual imports (not reexports) inside modules without child modules. And then I would like to implement text search myself, if you don't mind

Kobzol:
Oh, I will gladly leave the text search to you :laughing: I would really like to see this implemented, as it's one of the most common compiler/clippy warnings that I have to fix manually.

So, what exactly should I do now? To be honest I don't remember how this PR worked all that much, since it was a long time ago, I'll have to remind myself a bit.

dima74:
Could you please integrate [implementation of `exportedItems` using new resolve](https://github.com/intellij-rust/intellij-rust/commit/64faa5f385f8ab6adc03b10b29c768c72887fab9), and disable inspection on modules with child modules, and then I will review the code

Kobzol:
Just so we're clear, it should be disabled even in cases like this?
```rust
mod foo {
    pub struct S;
}

use foo::S;
```
I know that it's problematic with child modules, but like this the implementation will be very limited. I'll have to modify the tests to something like this in that case:
```rust
mod foo {
    pub struct S;
}

mod bar {
    use crate::foo::S;
}
```

dima74:
> Just so we're clear, it should be disabled even in cases like this?

Yes, in order to avoid false-positive when there is only usage in child module:
```rust
use source::func;
mod usage {
    fn usage() {
        super::func();
    }
}
mod source {
    pub fn func() {}
}
```

Alternatively we can somehow take into account paths from *inline* modules in `PathUsageMap`, but I don't think that we should do it, because it will be anyway covered by text search

Kobzol:
I disabled the inspection for modules that have a child mod and fixed the tests. Some tests are commented out, either because they don't work with the current implementation or they don't make sense without child modules.

I guess that some parts of my implementation should be removed to make way for text search, I'm not sure yet exactly what parts though.

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
```diff
@@ -0,0 +1,95 @@
...
+        val path = useSpeck.path ?: return true
+        val items = path.reference?.multiResolve() ?: return true
+
+        val name = useSpeck.itemName(withAlias = true) ?: return true
+        val usedItems = usage.pathUsages[name].orEmpty()
+
+        Pair(items.toSet(), usedItems)
+    }
+
+    return candidateItems.intersects(usedItems) || candidateItems.intersects(usage.traitUsages)
```
dima74, line 94: It was a bit hard for me to understand the logic with intersection

Also I am not sure that we should unite `pathUsages[it.name]`, because it may lead to false-negatives (please add such test):

```rust
mod mod1 {
    pub fn foo() {}
}
mod mod2 {
    pub use crate::mod1::foo;
    pub fn bar() {}
}
mod test {
    use crate::mod2::*;
    use crate::mod1::foo as bar;
    fn main() {
        bar();
    }
}
```

I would do something like this:
```kotlin
    val items = if (useSpeck.isStarImport) {
        val module = useSpeck.path?.reference?.resolve() as? RsMod ?: return true
        module.exportedItems(useSpeck.containingMod)
    } else {
        val path = useSpeck.path ?: return true
        val items = path.reference?.multiResolve() ?: return true
        val name = useSpeck.itemName(withAlias = true) ?: return true
        items.filterIsInstance<RsNamedElement>().map { NamedItem(name, it) }
    }

    return items.any { (name, item) ->
        item in usage.pathUsages[name].orEmpty()
            || item in usage.traitUsages
    }
```

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
```diff
@@ -0,0 +1,111 @@
...
+private val PATH_USAGE_KEY: Key<CachedValue<PathUsageMap>> = Key.create("PATH_USAGE_KEY")
+
+val RsItemsOwner.pathUsage: PathUsageMap
+    get() = CachedValuesManager.getCachedValue(this, PATH_USAGE_KEY) {
+        val usages = calculatePathUsages(this)
+        CachedValueProvider.Result.create(usages, PsiModificationTracker.MODIFICATION_COUNT)
+    }
+
+private fun calculatePathUsages(owner: RsItemsOwner): PathUsageMap {
+    val directUsages = mutableMapOf<String, MutableSet<RsElement>>()
```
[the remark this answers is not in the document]
Kobzol: Oh, I had no idea that it uses a different implementation. Thanks.

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
```diff
@@ -0,0 +1,95 @@
...
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitUseSpeck(o: RsUseSpeck) {
+            if (!holder.project.isNewResolveEnabled) return
+            val item = o.ancestorStrict<RsUseItem>() ?: return
+            if (item.isReexport) return
+            if (o.path?.resolveStatus != PathResolveStatus.RESOLVED) return
+
+            // Do not check uses if there is a child mod inside the current mod
+            val parentMod = o.containingMod
+            if (parentMod.expandedItemsExceptImplsAndUses.any { it is RsModItem || it is RsModDeclItem }) return
```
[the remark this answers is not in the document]
Kobzol: Do you mean that I should disable the unused import inspection for use imports inside functions?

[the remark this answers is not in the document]
dima74: Oh, I mean that we can run inspection inside functions even if there is child modules (because such imports can't be used outside function)

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
```diff
@@ -0,0 +1,111 @@
...
+): Boolean {
+    fun addItem(name: String, item: RsElement) {
+        directUsages.getOrPut(name) { mutableSetOf() }.add(item)
+    }
+
+    if (!element.isEnabledByCfg) return false
+
+    return when (element) {
+        is RsModItem -> false
+        is RsUseItem -> {
```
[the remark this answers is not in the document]
Kobzol: It is handled separately, because otherwise this:
```rust
use foo::{S};
```
would be seen as an usage of `S`, although it is not usage, it is just import. But it should be probably handled in a different way, I'll think about it.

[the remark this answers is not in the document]
dima74: We can use `RsPath#qualifier`, which is aware of imports

Kobzol:
I changed path traversal so that use items are not special cased and decided on the following rule:
A path is a use if it
1) isn't inside a use speck
2) is inside a use speck with an alias

However, I'm having trouble with your example code:
```rust
mod bar {
    pub fn foo1() {}
    pub fn foo2() {}
}
mod foo {
    use crate::bar::{foo1, foo2};
    use {foo1 as bar1, foo2 as bar2};

    fn foo() {
        bar1();
        bar2();
    }
}
```
The problem is that in `foo1 as bar1`, `foo1` does not resolve to anything (and therefore not even `bar1` and `bar2` are resolved later). Which seems like a resolve bug?

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
Kobzol: Thanks!

Kobzol:
I used your implementation, it's cleaner and it also works better =D

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
```diff
@@ -0,0 +1,108 @@
...
+            true
+        }
+        is RsPath -> {
+            if (element.qualifier != null) {
+                val requiredTraits = getAssociatedItemRequiredTraits(element).orEmpty()
+                traitUsages.addAll(requiredTraits)
+            } else {
+                val useSpeck = element.parent as? RsUseSpeck
+                // `use S as T;` is also a use of S
+                if (useSpeck == null || useSpeck.alias != null) {
```
dima74, line 74: > I changed path traversal so that use items are not special cased and decided on the following rule:
> A (root path, without a qualifier) path is a use if it
> 
> 1. isn't inside a use speck
> 2. is inside a use speck with an alias

It is a bit more complicated, e.g. there can be absolute paths (`use ::{foo1::foo2, ...}`), and also such test:

```rust
mod inner {
    pub fn func() {}
}
mod test {
    use crate::inner;
    use inner::{func};
    fn main() {
        func();
    }
}
```

I would do something like this:
```suggestion
                val useSpeck = element.parentOfType<RsUseSpeck>()
                if (useSpeck == null || useSpeck.isTopLevel) {


/**
 * We should collect paths only from relative use specks,
 * that is top-level use specks without `::`
 * E.g. we shouldn't collect such paths: `use ::{foo, bar}`
 */
private val RsUseSpeck.isRelative: Boolean
    get() = (path != null || coloncolon == null)
        && parentOfType<RsUseSpeck>()?.isRelative != false

```

dima74, line 74: Looks like we also should ignore `crate::` and `super::` paths here

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
```diff
@@ -0,0 +1,108 @@
...
+        val usages = calculatePathUsages(this)
+        CachedValueProvider.Result.create(usages, PsiModificationTracker.MODIFICATION_COUNT)
+    }
+
+private fun calculatePathUsages(owner: RsItemsOwner): PathUsageMap {
+    val directUsages = hashMapOf<String, MutableSet<RsElement>>()
+    val traitUsages = hashSetOf<RsTraitItem>()
+
+    for (child in owner.children) {
+        processElementsWithMacros(child) { element ->
```
dima74, line 38: We should also process macro paths here

```rust
use lazy_static::lazy_static;

lazy_static! {
    static ref EXAMPLE: u8 = 42;
}
```

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
Kobzol: Oh, I see that my way worked "by accident". Thanks for the test.

Kobzol:
While trying out your suggestions, I think that I found a bug in (the new) name resolution in edition 2018.
```rust
mod foo {
    use ::{std::collections};
    use collections::HashMap;

    fn bar(_: HashMap<u32, u32>) {}
}
```
This compiles in edition 2018, but it's unresolved in the IDE (also with edition 2018).
It seems to work with the old name resolution engine.

Kobzol:
Another weird case is this:
```rust
mod foo {
    #[macro_export]
    macro_rules! mac { () => {} }
}

mod bar {
    fn baz() {
        use crate::mac;
        mac!();
    }
}
```
Compiles in both 2015 & 2018 editions, but the `mac!` call is unresolved in the IDE (it works with the old name resolution).

This leads me to think: what about unresolved references?
Something like this:
```rust
mod bar {
    use super::foo::Baz;

    fn baz(x: Baz) {}
}
```
Currently, I ignore unresolved use items to avoid false positives. But maybe it would be better if we instead collected name uses that are unresolved and match that against the imports? So even though an import is unresolved, it would still be marked as unused, as long as no name that matches the import was used in its scope.

dima74:
> Compiles in both 2015 & 2018 editions, but the `mac!` call is unresolved in the IDE (it works with the old name resolution).

This is because new resolve currently is not used inside functions

> But maybe it would be better if we instead collected name uses that are unresolved and match that against the imports?

This is good idea, but probably it is better to do in another PR (since it is actually false-negative)

Also if import is resolved, but usage is unresolved, we will have false-positive

Kobzol:
> This is because new resolve currently is not used inside functions

I thought that this means that the old resolve will be used inside functions. But currently this code resolves if I disable the new engine, and doesn't resolve if I enable the new engine.

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
Kobzol: I'm not sure how to do this properly because of the unresolved macro problems mentioned in https://github.com/intellij-rust/intellij-rust/pull/6392#issuecomment-853369494.

dima74: Let's then ignore imports which resolve to macros in this PR, and I will think how to best handle macros

Kobzol:
Rebased over master to fix merge conflict and added a check to ignore imports resolved to macros.

dima74:
I run regression testing and it found some false-positives on `cargo` project. Some of them are not directly related to this PR (e.g. `RsWithMacrosRecursiveElementWalkingVisitor` don't handle some base macros like `writeln!`). Because this PR is already quite big, I propose to temporary disalbe inspection by default and merge it, and then I will fix the false-postivies

Kobzol:
I agree. I rebased again and disabled the inpsection by default.

dima74:
Thanks!
bors r+

Issue #6379, opened 2020-11-11
tokio 0.3 code completion has problem


Sherlock-Holo:
## Environment

* **IntelliJ Rust plugin version:** 0.3.136.3472-202-nightly
* **Rust toolchain version:** rustc 1.49.0-nightly (1773f60ea 2020-11-08)
* **IDE name and version:** clion 2020.2
* **Operating system:** Windows with WSL toolchain

## Problem description
the codes
```rust
use tokio::io::AsyncWriteExt;
use tokio::net::TcpStream;

#[tokio::main]
async fn main() {
    let mut stream = TcpStream::connect("www.test.com:80").await.unwrap();
    stream.write_all(b"123").
}
```

when I type `.`  the completion list doesn't have `await` or any about the `WriteAll` method, only have 
![image](https://user-images.githubusercontent.com/10096425/98787761-dc709180-243a-11eb-8fca-676473a68aff.png)

## Steps to reproduce

input this codes and look the completion list

vlad20012:
The problem is legacy-scope `write_impl!` macro defined with other macro:
```rust
cfg_io_util! {
    /// Defines numeric writer
    macro_rules! write_impl {
```

Sherlock-Holo:
> The problem is legacy-scope `write_impl!` macro defined with other macro:
> 
> ```rust
> cfg_io_util! {
>     /// Defines numeric writer
>     macro_rules! write_impl {
> ```
> 
> This should be eventually covered by #6217

Can I experience the new name resolution for now?

vlad20012:
Oh, I'm wrong. The problem is in `pin_project!` macro expansion. New resolve only makes things worst in this case (i.e. also adds `write_impl` resolution failure)

PR #6402, opened 2020-11-17
REF: Support resolve macro call to expanded macro def using PSI

dima74:
Support resolving macro call to expanded macro def in [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217) when using PSI. Resolve without PSI already works, so this change doesn't affect macro expansions.

Fixes https://github.com/intellij-rust/intellij-rust/issues/6379#issuecomment-725364423


vlad20012 (approved):
bors r+

PR #6404, opened 2020-11-17
Support fully-qualified paths in macro calls at item level.

msmorgan:
This change adds '::', 'super', and 'self' tokens to Item_first, so
macro calls can use paths that start with these.

changelog: Add support for qualified paths in item-level macro invocations.


vlad20012:
:+1: 
Let's also change the parser version in `RustParserDefinition`

Undin (commented):
Note, test failed

vlad20012:
@msmorgan ping

msmorgan:
@Undin @vlad20012 I'm not sure how to fix the tests for [RsPartialParsingTestCase](../blob/master/src/test/kotlin/org/rust/lang/core/parser/RsPartialParsingTestCase.kt). I could change the expected output to match the PSI emitted, but it doesn't look right.

The first two lines of [paths.rs](../blob/master/src/test/resources/org/rust/lang/core/parser/fixtures/partial/paths.rs) now produce the following PSI. (The previous output is [here](../blob/master/src/test/resources/org/rust/lang/core/parser/fixtures/partial/paths.txt).)

```
FILE
  RsUseItemImpl(USE_ITEM)
    PsiElement(use)('use')
  PsiErrorElement:'::' or ';' expected, got '::'
    <empty list>
  PsiWhiteSpace(' ')
  PsiElement(::)('::')
  PsiErrorElement:Self, crate, identifier, self or super expected, got ';'
    PsiElement(;)(';')
  PsiWhiteSpace('\n')
  RsUseItemImpl(USE_ITEM)
    PsiElement(use)('use')
  PsiErrorElement:'::' or ';' expected, got '::'
    <empty list>
  PsiWhiteSpace(' ')
  PsiElement(::)('::')
  PsiErrorElement:Self, crate, identifier, self or super expected, got '::'
    <empty list>
  PsiElement(::)('::')
  PsiErrorElement:Self, crate, identifier, self or super expected, got ';'
    PsiElement(;)(';')
  PsiWhiteSpace('\n')
```

Specifically, the line `PsiErrorElement:'::' or ';' expected, got '::'` seems suspect to me, since it appears to say it got the token it was expecting and produced an error anyway.

Frankly, this all a bit over my head and I could use some help.

vlad20012:
> I could change the expected output to match the PSI emitted, but it doesn't look right.

Well, yes, but an ideal fix requires changes in `path` parsing rules (in order to make `::` in `use ::;` parsed as an incomplete path). I'd not suggest trying it because it can be quite tricky. I'll try it later. I think that correct parsing of a correct code is much more important than good error messages in an incorrect one, so let's just replace the expected output (and the parser version in `RustParserDefinition`) and r+ it.

msmorgan:
@vlad20012 Done.

vlad20012:
:+1: 
bors r+

PR #6406, opened 2020-11-18
Fix 2020.3 compatibility

vlad20012:
`FileAttributes` has `caseSensitivity` property on 203.
Default value `CaseSensitivity.UNKNOWN` leads to errors in logs

### DIFF
--- a/src/202/main/kotlin/org/rust/lang/core/macros/compatUtils.kt
+++ b/src/202/main/kotlin/org/rust/lang/core/macros/compatUtils.kt
@@ -6,6 +6,7 @@
 package org.rust.lang.core.macros
 
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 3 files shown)


src/202/main/kotlin/org/rust/lang/core/macros/compatUtils.kt
```diff
@@ -28,3 +29,6 @@ fun newSerializedStubTreeDataExternalizer(
 @Suppress("UnstableApiUsage")
 fun createFileContent(project: Project, file: ReadOnlyLightVirtualFile, fileContent: String): FileContent =
     FileContentImpl(file, fileContent, file.modificationStamp).also { it.project = project }
+
+fun createFileAttributes(isDir: Boolean, length: Long, lastModified: Long) =
```
Undin, line 33: nit: explicit return type

PR #6352, opened 2020-11-04
INSP: Add error for `repr` without parentheses

zeroeightysix:
Adds a new error for when a `repr` attribute does not have `RsMetaItemArgs` (thus, no pair of parentheses)

image

Fixes #6336

changelog: Show error if `repr` attribute doesn't have parentheses and provide quick-fix to add them

PR #6413, opened 2020-11-19
INSP: Detect attributes that require parentheses but have no parentheses.

zeroeightysix:
This PR builds on #6352, making the diagnostic and fix work for all attributes that 'require parentheses', e.g. are not valid as just a name (like `#[test]`).

image

changelog: Detect attributes that require parentheses but have no parentheses, and suggest `Add parentheses` quick-fix


src/main/kotlin/org/rust/ide/inspections/RsAttrWithoutParenInspection.kt
```diff
@@ -0,0 +1,40 @@
...
+        "warn",
+        "forbid",
+        "deny",
+        "proc_macro_derive"
+    )
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitMetaItem(metaItem: RsMetaItem) {
+            val name = metaItem.name ?: return
+            if (name in attributesWithParen && metaItem.metaItemArgs == null) {
```
Undin, line 34: Let's also check that `metaItem.parent` is `RsAttr`. 
Otherwise, we can get false positive erros for custom proc-macros.
For example:
`lib.rs`
```rust
use proc_macro::TokenStream;

#[proc_macro_attribute]
pub fn return_as_is(_attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}
```

`main.rs`
```rust
use crate_name::return_as_is;

#[return_as_is(allow)] // <- there isn't error here
fn foo() {}

fn main() {}
```

Of course, in future we should take into account child attributes of `cfg_attr` but we don't convenient API for it currently

src/main/kotlin/org/rust/ide/inspections/RsAttrWithoutParenInspection.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package org.rust.ide.inspections
+
+import org.rust.lang.core.psi.RsMetaItem
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.name
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.lang.utils.addToHolder
+
+class RsAttrWithoutParenInspection : RsLocalInspectionTool() {
```
Undin, line 14:
```suggestion
class RsAttrWithoutParenthesesInspection : RsLocalInspectionTool() {
```

Undin (changes requested):

src/test/kotlin/org/rust/ide/annotator/fixes/AddAttrParenthesesFixTest.kt
```diff
@@ -5,10 +5,10 @@
 
 package org.rust.ide.annotator.fixes
 
-import org.rust.ide.annotator.RsAnnotatorTestBase
-import org.rust.ide.annotator.RsErrorAnnotator
+import org.rust.ide.inspections.RsAttrWithoutParenthesesInspection
+import org.rust.ide.inspections.RsInspectionsTestBase
 
-class AddAttrParenthesesFixTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
+class AddAttrParenthesesFixTest : RsInspectionsTestBase(RsAttrWithoutParenthesesInspection::class) {
```
Undin, line 11: Let's move all tests from this class to `RsAttrWithoutParenthesesInspectionTest`

src/test/kotlin/org/rust/ide/inspections/RsAttrWithoutParenthesesInspectionTest.kt
```diff
@@ -0,0 +1,26 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections
+
+class RsAttrWithoutParenthesesInspectionTest : RsInspectionsTestBase(RsAttrWithoutParenthesesInspection::class) {
```
Undin, line 8: Let's add at least three additional tests:
- with custom attribute with child attribute from `attributesWithParen` list, something like
```rust
#[custom_attr(repr)]
struct Foo(i32);
```
- something with `cfg_attr` where, ideally, we should detect the same error but don't do it yet
```rust
#[cfg_attr(unix, repr)]
struct Foo(u8);
```
- case with an inner attribute like `#![warn]`

src/main/kotlin/org/rust/ide/inspections/RsAttrWithoutParenthesesInspection.kt
```diff
@@ -0,0 +1,42 @@
...
+            if (metaItem.parent !is RsAttr) return
+            val name = metaItem.name ?: return
+            if (name in attributesWithParen && metaItem.metaItemArgs == null) {
+                RsDiagnostic.NoAttrParentheses(metaItem, name).addToHolder(holder)
+            }
+        }
+    }
+
+    companion object {
+        private val attributesWithParen = setOf(
```
Undin, line 28:
```suggestion
        private val ATTRIBUTES_WITH_PARENTHESES = setOf(
```

Undin (commented):
LGTM 👍 
Could you squash current and next "fix" commits into main ones?

Undin (approved):
Great!
Thank you!
bors r+

PR #6414, opened 2020-11-19
INT: Add intention for replacing multiple `_` wildcards into one `..`

### LINKED ISSUE #6409: Suggest to replace redundant wildcard patterns with `..`
Consider any enum with variants with more than one item:
```rust
enum Enum {
    A(usize, usize),
    B(String, usize, bool)
}
```

```rust
let e = Enum::A(1, 2);

match e {
    Enum::A(_, _) => { // Can be replaced with Enum::A(..) =>
        println!("{}", "hello world!");
    }
    _ => ()
};
```

```rust
if let Enum::B(a, _, _) = e { // Can be replaced with Enum:B(a, ..) =>
    println!("{}, {}!", "hello", a)
}
```

It might be useful to provide an intent (maybe with even a warning, such as the unnecessary parentheses warning) to convert redundant usages of multiple `_` to one `..`.

zeroeightysix:
Closes #6409

changelog: Add intention to replace redundant wildcards with a single `..`

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
@@ -7,6 +7,7 @@ package org.rust.ide.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.ancestorStrict
@@ -14,30 +15,47 @@ import org.rust.lang.core.psi.ext.endOffset
 import org.rust.lang.core.psi.ext.startOffset
 
 class JoinWildcardsIntention : RsElementBaseIntentionAction<List<RsPatWild>>() {
-    override fun getFamilyName(): String = "Join wildcards into `..`"
-    override fun getText(): String = familyName
+    override fun getFamilyName(): String = "Replace successive `_` with `..`"
 
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): List<RsPatWild>? {
-        val patList = when (val pat = element.ancestorStrict<RsPat>() ?: return null) {
-            is RsPatTup -> pat.patList
-            is RsPatTupleStruct -> pat.patList
-            is RsPatSlice -> pat.patList
+        var patUnderCaret = element.ancestorStrict<RsPat>() ?: return null
+        if (patUnderCaret is RsPatWild) {
+            patUnderCaret = patUnderCaret.ancestorStrict() ?: return null
+        }
+        val patList = when (patUnderCaret) {
+            is RsPatTup -> patUnderCaret.patList
+            is RsPatTupleStruct -> patUnderCaret.patList
+            is RsPatSlice -> patUnderCaret.patList
             // note that RsOrPat also has (and allows) multiple _ wildcards.
             // it does, however, not support `..`, so we leave it out for this inspection.
             else -> return null
         }
-        val wildcards = patList.takeLastWhile { it is RsPatWild }.map { it as RsPatWild }
 
-        // There is more than one wildcard, thus it is useful to join to `..`
-        // Replacing one (or zero) wildcards with one `..` is not useful.
-        return if (wildcards.size > 1) {
-            wildcards
-        } else null
+        // Unavailable if `..` is already there
+        if (patList.any { it is RsPatRest }) return null
+
+        val patWildSeq = mutableListOf<RsPatWild>()
+        for (pat in patList.asSequence() + sequenceOf(null)) {
+            if (pat is RsPatWild) {
+                patWildSeq += pat
+            } else {
+                if (patWildSeq.size > 0) {
+                    val patWildSeqRange = TextRange(patWildSeq.first().startOffset, patWildSeq.last().endOffset)
+                    if (element.startOffset in patWildSeqRange) {
+                        text = if (patWildSeq.size == 1) "Replace `_` with `..`" else familyName
+                        return patWildSeq
+                    }
+                }
+                patWildSeq.clear()
+            }
+        }
+        return null
     }
 
     override fun invoke(project: Project, editor: Editor, ctx: List<RsPatWild>) {
         val startOffset = ctx.first().startOffset
         val endOffset = ctx.last().endOffset
         editor.document.replaceString(startOffset, endOffset, "..")
+        editor.caretModel.moveToOffset(startOffset + 2)
     }
 }
--- a/src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
@@ -1,0 +1,43 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.lang.core.psi.ext.startOffset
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.lang.core.psi.ext.startOffset
+
+class JoinWildcardsIntention : RsElementBaseIntentionAction<List<RsPatWild>>() {
+    override fun getFamilyName(): String = "Join wildcards into `..`"
+    override fun getText(): String = familyName
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): List<RsPatWild>? {
+        val structPat = element.ancestorStrict<RsPatTupleStruct>() ?: return null
```
Kobzol, line 22: I think that `RsPatTup`s can be also applicable for this intention (for example `let (a, ..) = (1, 2, 3);`).

zeroeightysix: And apparently slices, as well. I didn't even know you could match slices - TIL.

Kobzol: Yeah I find that very common, working on the plugin gives me a lot of new knowledge about Rust :D

src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
```diff
@@ -0,0 +1,43 @@
...
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): List<RsPatWild>? {
+        val patList = when (val pat = element.ancestorStrict<RsPat>() ?: return null) {
+            is RsPatTup -> pat.patList
+            is RsPatTupleStruct -> pat.patList
+            is RsPatSlice -> pat.patList
+            // note that RsOrPat also has (and allows) multiple _ wildcards.
+            // it does, however, not support `..`, so we leave it out for this inspection.
+            else -> return null
+        }
+        val wildcards = patList.takeLastWhile { it is RsPatWild }.map { it as RsPatWild }
```
Kobzol, line 29: Some remarks:
1) It does not have to be only at the end, i.e. `let (_, _, a) = (1, 2, 3);`. Basically any sequence of 1 or more wild patterns can be replaced with a single `..`.
2) You should also check if `..` is already not present in the pat.
3) The intention should be applicable on a wild pat, not on its whole parent pat, because of 1). Consider e.g. `(_, _, a, _, _) = (1, 2, 3, 4, 5)`. You can only change the first two or the last two underscores to `..`, but not both.

Kobzol (changes requested):
Please add some basic tests (check for example `AddStructFieldsPatIntentionTest` for inspiration).

Kobzol:
Just a quick comment, I think that replacing a single `_` with `..` is also useful. You can do this proactively to ensure that the pattern will stay forward compatible.

vlad20012:
I've pushed some fixes. They're almost fixing @Kobzol's remarks:
- Make it work with leading and middle `_` patterns
- Make it work with single `_`
- Reduce the intention availability range
- Add tests

src/main/resources/intentionDescriptions/JoinWildcardsIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention converts multiple <code>_</code> wildcards to <code>..</code>.
```
Kobzol, line 3: Maybe leave out multiple, since single `_` is now also supported?
```suggestion
This intention converts <code>_</code> wildcard(s) to <code>..</code>.
```

src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.lang.core.psi.ext.startOffset
+
+class JoinWildcardsIntention : RsElementBaseIntentionAction<List<RsPatWild>>() {
+    override fun getText(): String = "Replace `_` sequence with `..`"
```
vlad20012, line 18: I also renamed the intention, But I don't like the new name too. Maybe "Replace repeating `_` with `..`"? But these names are not applicable for single `_`

src/test/kotlin/org/rust/ide/intentions/JoinWildcardsIntentionTest.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    fun `test slice`() = doAvailableTest("""
+        fn main() {
+            let [first, _, _/*caret*/] = [1, 2, 3];
+        }
+    """, """
+        fn main() {
+            let [first, ../*caret*/] = [1, 2, 3];
+        }
+    """)
```
Kobzol, line 83: Maybe also add a test that checks the logic on line 45? If multiple "runs" of wild pats are present.

```kotlin
fun `test multiple wild pat runs`() = doAvailableTest("""
    fn main() {
        let [_, _, x, _/*caret*/, _] = [1, 2, 3, 4, 5];
    }
  """, """
    fn main() {
        let [_, _, x, ../*caret*/] = [1, 2, 3, 4, 5];
    }
  """)
```

Kobzol (approved):

src/main/kotlin/org/rust/ide/intentions/JoinWildcardsIntention.kt
Kobzol: I think that `sequence` is fine, since sequence can also have length 1. But otherwise maybe `successive` instead of `repeating`?

vlad20012: Ok, I made the name different for single or multiple `_` cases

vlad20012:
@zeroeightysix Thanks!
@Kobzol Thanks!

bors r+

src/main/resources/intentionDescriptions/JoinWildcardsIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention converts successive <code>_</code> wildcard(s) to <code>..</code>.
```
vlad20012, line 3:
```suggestion
This intention converts successive <code>_</code> wildcard(s) to <code>..</code>
```
Oh, this dot at the end can be confusing

vlad20012: image

Kobzol: Hmm, I wonder if anyone even reads these intention descriptions :D

vlad20012: (me too, but anyway)

src/main/resources/intentionDescriptions/JoinWildcardsIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention converts successive <code>_</code> wildcard(s) to <code>..</code>
```
Undin, line 3:
```suggestion
Converts successive <code>_</code> wildcard(s) to <code>..</code>
```

src/main/resources/intentionDescriptions/JoinWildcardsIntention/description.html
Undin: JFYI I suppose it will fail on `RsIntentionTestBase.test intention has documentation`

src/main/resources/intentionDescriptions/JoinWildcardsIntention/description.html
vlad20012: Almost all our intention descriptions start with `This intention ...`. I think we should rewrite them all at once

vlad20012:
bors r+
(please)

Issue #6411, opened 2020-11-19
ML code completion for Rust


ortem:
IntelliJ-based IDEs provide a ranking mechanism for code completion based on machine learning. This mechanism [is already supported](https://blog.jetbrains.com/idea/2020/10/intellij-idea-2020-3-eap5/) for Java and some other languages. It would be great to support this for Rust as well.

TODO:
- [ ] Basic support of ML ranking for Rust code completion
- [ ] Provide Rust-specific element features using `com.intellij.codeInsight.completion.ml.ElementFeatureProvider`
  - [ ] Consider if element is a keyword
  - [ ] Consider element's PSI type
  - [ ] Consider element's origin (stdlib or not)
  - [ ] Consider expected argument names based on parameter names
  - [ ] Consider expected argument types based on parameter types
- [ ] Provide Rust-specific context features using `com.intellij.codeInsight.completion.ml.ContextFeatureProvider`
- [ ] Enable ML ranking by default

To be continued.

PR #6419, opened 2020-11-20
Provide ML completion ranking for Rust

ortem:
Part of https://github.com/intellij-rust/intellij-rust/issues/6411

This PR adds a new module `ml-completion` to the project. This module provides the ML-based code completion ranking support for Rust. Currently, it consists of:
* `RsCompletionFeaturesPolicy` enabling Ngram model for future models
* `RsElementFeatureProvider` providing Rust-specific features to train future models, with tests
* `RsMLRankingProvider` providing an initial ranking model (trained without the stuff above, therefore not very intelligent so far)

Thanks to @bibaev for help!

### How to use
The new ranking model can be enabled in `Editor | General | Code Completion | Machine Learning-Assisted Completion` settings (see [help](https://www.jetbrains.com/help/idea/auto-completing-code.html#ml_completion) for more details). However, at the moment, the model is very rough, so it's not really recommended for enabling. 

settings

before

after

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import groovy.json.JsonSlurper
 import org.apache.tools.ant.taskdefs.condition.Os.*
 import org.gradle.api.JavaVersion.VERSION_1_8
 import org.gradle.api.internal.HasConvention
@@ -12,7 +13,6 @@ import org.jetbrains.intellij.tasks.RunIdeTask
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import org.jsoup.Jsoup
-import groovy.json.JsonSlurper
 import java.io.Writer
 import java.net.URL
 import kotlin.concurrent.thread
@@ -45,6 +45,7 @@ val javaPlugin = "java"
 val javaScriptPlugin = "JavaScript"
 // BACKCOMPAT: 2020.2
 val clionPlugins = if (platformVersion < 203) emptyList() else listOf("com.intellij.cidr.base", "com.intellij.clion")
+val mlCompletionPlugin = "com.intellij.completion.ml.ranking"
 
 plugins {
     idea
@@ -194,6 +195,10 @@ project(":plugin") {
             psiViewerPlugin,
             javaScriptPlugin
         )
+        // BACKCOMPAT: 2020.2
+        if (platformVersion >= 203) {
+            plugins += mlCompletionPlugin
+        }
         if (baseIDE == "idea") {
             plugins += listOf(
                 copyrightPlugin,
@@ -216,6 +221,7 @@ project(":plugin") {
         implementation(project(":duplicates"))
         implementation(project(":grazie"))
         implementation(project(":js"))
+        implementation(project(":ml-completion"))
     }
 
     tasks {
@@ -441,6 +447,26 @@ project(":grazie") {
 project(":js") {
     intellij {
         setPlugins(javaScriptPlugin)
+    }
+    dependencies {
+        implementation(project(":"))
+        implementation(project(":common"))
+        testImplementation(project(":", "testOutput"))
+        testImplementation(project(":common", "testOutput"))
+    }
+}
+
+project(":ml-completion") {
+    intellij {
+        // BACKCOMPAT: 2020.2
+        if (platformVersion >= 203) {
+            val plugins = mutableListOf<Any>(mlCompletionPlugin)
+            // TODO: drop it when CLion move `navigation.class.hierarchy` property from c-plugin to CLion resources
+            if (baseIDE == "clion") {
+                plugins += "c-plugin"
+            }
+            setPlugins(*plugins.toTypedArray())
+        }
     }
     dependencies {
         implementation(project(":"))
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -75,6 +75,7 @@ allprojects {
         jcenter()
         maven("https://dl.bintray.com/jetbrains/markdown")
         maven("http://download.eclipse.org/jgit/maven")
+        maven("https://dl.bintray.com/jetbrains/intellij-third-party-dependencies")
     }
 
     idea {
@@ -469,6 +470,7 @@ project(":ml-completion") {
         }
     }
     dependencies {
+        implementation("org.jetbrains.intellij.deps.completion:completion-ranking-rust:0.0.4")
         implementation(project(":"))
         implementation(project(":common"))
         testImplementation(project(":", "testOutput"))
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,6 +198,7 @@ The current Rust plugin modules:
 * `:coverage` - integration with [coverage](https://github.com/JetBrains/intellij-community/tree/master/plugins/coverage-common) plugin
 * `:grazie` - integration with [grazie](https://plugins.jetbrains.com/plugin/12175-grazie) plugin 
 * `:js` - interop with JavaScript language
+* `:ml-completion` - integration with [Machine Learning Code Completion](https://github.com/JetBrains/intellij-community/tree/master/plugins/completion-ml-ranking) plugin
 
 If you want to implement integration with another plugin/IDE, you should create a new gradle module for that.
 
--- a/ml-completion/src/203/main/kotlin/org/rust/ml/RsCompletionFeaturesPolicy.kt
+++ b/ml-completion/src/203/main/kotlin/org/rust/ml/RsCompletionFeaturesPolicy.kt
@@ -1,0 +1,12 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ml
+
+import com.intellij.completion.ml.features.CompletionFeaturesPolicy
+
+class RsCompletionFeaturesPolicy : CompletionFeaturesPolicy {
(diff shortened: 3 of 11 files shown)


build.gradle.kts
```diff
@@ -45,6 +45,8 @@ val javaPlugin = "java"
 val javaScriptPlugin = "JavaScript"
 // BACKCOMPAT: 2020.2
 val clionPlugins = if (platformVersion < 203) emptyList() else listOf("com.intellij.cidr.base", "com.intellij.clion")
+val mlCompletionPlugin = "com.intellij.stats.completion"
+val mlCompletionRankingRustPlugin = "org.jetbrains.intellij.deps.completion:completion-ranking-rust:0.0.4"
```
bibaev, line 49: I'd suggest renaming the value to "mlCompletionModelLibrary"

ml-completion/src/main/kotlin/RustElementFeatureProvider.kt
```diff
@@ -0,0 +1,197 @@
...
+
+@Suppress("UnstableApiUsage")
+class RustElementFeatureProvider : ElementFeatureProvider {
+    override fun getName(): String = "rust"
+
+    override fun calculateFeatures(
+        element: LookupElement,
+        location: CompletionLocation,
+        contextFeatures: ContextFeatures
+    ): MutableMap<String, MLFeatureValue> {
```
bibaev, line 27: This method might return immutable `Map<String, MLFeatureValue>`

ml-completion/src/main/kotlin/RustMLRankingProvider.kt
```diff
@@ -0,0 +1,15 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.ml
+
+import com.intellij.internal.ml.catboost.CatBoostJarCompletionModelProvider
+import com.intellij.lang.Language
+
+@Suppress("UnstableApiUsage")
+class RustMLRankingProvider : CatBoostJarCompletionModelProvider("Rust", "rust_features", "rust_model") {
```
bibaev, line 12: It worth adding a test to check that metadata we ship with the model is fully consistent. See example here -  `com.jetbrains.completion.ml.ranker.MetadataConsistencyTest#testRustMetadata`

ml-completion/src/main/kotlin/RustElementFeatureProvider.kt
```diff
@@ -0,0 +1,197 @@
...
+        if (containingCrate != null) {
+            result["is_from_stdlib"] = MLFeatureValue.binary(containingCrate.isStd)
+        }
+
+        return result
+    }
+}
+
+/** See [org.rust.lang.core.psi.RsTokenType] */
+internal enum class RsKeywordMLKind {
```
Undin, line 59: Let's somehow check that this list contains exactly the same elements as `org.rust.lang.core.psi.RsTokenTypeKt#RS_KEYWORDS` not to forgot to add the corresponding element to the second list when you add a new keyword element

ml-completion/src/main/kotlin/RustCompletionFeaturesPolicy.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ml
+
+import com.intellij.completion.ml.features.CompletionFeaturesPolicy
+
+class RustCompletionFeaturesPolicy : CompletionFeaturesPolicy {
```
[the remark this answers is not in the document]
ortem: Renamed, as well as `RustElementFeatureProvider` -> `RsElementFeatureProvider`

ml-completion/src/main/kotlin/RustElementFeatureProvider.kt
```diff
@@ -0,0 +1,197 @@
...
+
+        /**
+         * Otherwise, if [element] is [RsElement],
+         * store PSI kind as "kind" feature
+         * and store if its origin is stdlib as "is_from_stdlib" feature
+         */
+        val rsElement = psiElement as? RsElement ?: return result
+        val rsPsiElementKind = RsPsiElementMLKind.from(rsElement)
+        if (rsPsiElementKind != null) {
+            result["kind"] = MLFeatureValue.categorical(rsPsiElementKind)
```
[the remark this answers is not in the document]
ortem: Extracted to
```Kotlin
private object Features {
    const val Kind: String = "kind"
    const val IsFromStdlib: String = "is_from_stdlib"
}
```

ml-completion/src/main/kotlin/RustElementFeatureProvider.kt
ortem: Added `test all keywords are covered` test to `RsElementFeatureProviderTest`

ml-completion/src/test/kotlin/RustModelMetadataConsistencyTest.kt
```diff
@@ -0,0 +1,12 @@
...
+import org.rust.ml.RustMLRankingProvider
+
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+class RustModelMetadataConsistencyTest {
+    @Test
+    fun testMetadata() = RustMLRankingProvider().assertModelMetadataConsistent()
```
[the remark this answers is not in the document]
ortem: Changed to `test model metadata consistency`

Undin (approved):
LGTM except minor comments

ml-completion/src/main/kotlin/org/rust/ml/RustMLRankingProvider.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package org.rust.ml
+
+import com.intellij.internal.ml.catboost.CatBoostJarCompletionModelProvider
+import com.intellij.lang.Language
+import org.rust.lang.RsLanguage
+
+@Suppress("UnstableApiUsage")
+class RustMLRankingProvider : CatBoostJarCompletionModelProvider("Rust", "rust_features", "rust_model") {
```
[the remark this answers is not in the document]
ortem: I've chosen the name with "Rust" word because all the `CatBoostJarCompletionModelProvider` subclasses use the full language names instead of the abbreviations (including our experimental model provider for EAP builds `ExperimentRustMLRankingProvider`). So it would be easier to search across all the implementations with `Find external Usages` action for those who unfamiliar with our "Rs" prefix, for example

Undin (changes requested):
Could you rebase into master, please? It will show the problem with `com.intellij.stats.completion` dependency (CI faces it because it actually checks master +  your changes)

ml-completion/src/test/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,5 @@
+<idea-plugin xmlns:xi="http://www.w3.org/2001/XInclude" allow-bundled-update="true">
+    <id>org.rust.ml</id>
+    <xi:include href="/META-INF/rust-core.xml" xpointer="xpointer(/idea-plugin/*)"/>
+    <depends optional="true" config-file="ml-completion-only.xml">com.intellij.stats.completion</depends>
```
Undin, line 4: plugin id is not updated to `com.intellij.completion.ml.ranking`, as a result, test on 2020.3 fail

build.gradle.kts
```diff
@@ -45,6 +45,7 @@ val javaPlugin = "java"
 val javaScriptPlugin = "JavaScript"
 // BACKCOMPAT: 2020.2
 val clionPlugins = if (platformVersion < 203) emptyList() else listOf("com.intellij.cidr.base", "com.intellij.clion")
+val mlCompletionPlugin = "com.intellij.completion.ml.ranking"
```
Undin, line 48: Looks like `com.intellij.completion.ml.ranking` was introduced only in 2020.3 (extracted from `com.intellij.stats.completion`) so current impls cannot be applied to 2020.2

I see 2 variants:
- try to support 2020.2 as well. It leads to a lot of conditions in build scripts and xmls
- support only 2020.3

I prefer the second option (support only 2020.3) because it's simpler (for example, `CatBoostJarCompletionModelProvider` exists only in 2020.3) and we are going to drop 2020.2 support quite soon

Undin (changes requested):

Undin (commented):
Also, there is a merge conflict

Undin (commented):
Note, I've added missed keywords in #6492 (it was merged recently) so we should update `RsKeywordMLKind` and the corresponding tests

ortem:
Rebased onto master (the first force-push) and fixed comments (the second force-push).

Also, I removed `"` symbols from `RsElementFeatureProviderTest` due to `[DANGEROUS_CHARACTERS] Name contains characters which can cause problems on Windows` warning.

build.gradle.kts
ortem: I agree with you, so I moved all the code into `203` folder

build.gradle.kts
```diff
@@ -450,6 +457,22 @@ project(":js") {
     }
 }
 
+project(":ml-completion") {
+    intellij {
+        // BACKCOMPAT: 2020.2
+        if (platformVersion >= 203) {
+            setPlugins(mlCompletionPlugin)
+        }
```
Undin, line 465: I suppose we also have to add dependency in `c-plugin` when we use `clion` as base IDE to avoid weird test fails because of `'navigation.class.hierarchy' is not found` error.
See `:toml` project as example

ortem: Done

Undin (approved):
bors r+

PR #6421, opened 2020-11-22
ANN: improve "Item is private" annotation and "Make public" fix

Undin:
These changes:
- make range of `Item is private` annotation narrower
- provide `Make public` quick-fix available for items with restricted visibility in case of `Item is private` error
- fix `Make public` quick-fix for items with restricted visibility in case of `Item is private, and cannot be re-exported` error
- don't allow providing `Make public` quick-fix for unsupported cases

changelog: Make `Item is private` error highlighting more precisely; Provide `Make public` quick-fix for items with [restricted visibility](https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself) like `pub(crate)`

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/MakePublicFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/MakePublicFix.kt
@@ -14,8 +14,8 @@ import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 
-class MakePublicFix(
-    element: RsVisible,
+class MakePublicFix private constructor(
+    element: RsVisibilityOwner,
     elementName: String?,
     private val withinOneCrate: Boolean
 ) : LocalQuickFixAndIntentionActionOnPsiElement(element) {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/MakePublicFix.kt
```diff
@@ -70,16 +70,15 @@ class MakePublicFix(
...
-                    && visible.containingCrate?.origin == PackageOrigin.WORKSPACE
-                -> MakePublicFix(visible, elementName, withinOneCrate)
-                else -> null
-            }
+        fun createIfCompatible(
+            visible: RsVisibilityOwner,
+            elementName: String?,
+            withinOneCrate: Boolean
+        ): MakePublicFix? {
+            if (visible is RsEnumVariant ||
```
ortem, line 78: Maybe it's worth to return `null` in case `visible.visibility is RsVisibility.Public` as well

ortem (approved):
LGTM except one optional thing

src/main/kotlin/org/rust/ide/annotator/fixes/MakePublicFix.kt
Undin: I don't think it's really needed here.
In my opinion, `visible.visibility is RsVisibility.Public` it's a part of a condition that inspection/annotator should check to provide some annotation at all (even without quick-fix) so the addition of this code here will be just a duplication of inspection/annotator code.
Other conditions in `createIfCompatible ` are more about quick fix itself and its current restrictions of implementation

Undin:
bors r=ortem

PR #6423, opened 2020-11-23
Support `Find Usages` action for Cargo features

vlad20012:
Usage search works for Rust and TOML code. Related to #4693

With `ctrl + B`:
image

With `alt + F7`:
image

Usage highlighting:
Peek 2020-11-23 17-27

There is a **bug** with usage highlighting - now any key in any TOML file is highlighted as some definition when under the caret. But I've checked that YAML plugin has the same bug (or a feature?), so I don't think it's critical. I also found that this bug (feature?) can be fixed with a new `Symbol` API that is available on 2020.3 platform.

Peek 2020-11-23 17-53

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/search/CargoTomlUsageTypeProvider.kt
+++ b/toml/src/main/kotlin/org/rust/toml/search/CargoTomlUsageTypeProvider.kt
@@ -1,0 +1,33 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.search
+
+import com.intellij.psi.PsiElement
+import com.intellij.usages.UsageTarget
+import com.intellij.usages.impl.rules.UsageType
+import com.intellij.usages.impl.rules.UsageTypeProviderEx
+import org.rust.lang.core.RsPsiPattern.anyCfgFeature
+import org.rust.lang.core.psi.RsLitExpr
+import org.rust.toml.CargoTomlPsiPattern.onFeatureDependencyLiteral
+import org.toml.lang.psi.TomlLiteral
+
+class CargoTomlUsageTypeProvider : UsageTypeProviderEx {
+    override fun getUsageType(element: PsiElement?): UsageType? = getUsageType(element, UsageTarget.EMPTY_ARRAY)
+
+    override fun getUsageType(element: PsiElement?, targets: Array<out UsageTarget>): UsageType? {
+        return when (element) {
+            is TomlLiteral -> if (onFeatureDependencyLiteral.accepts(element)) FEATURE_DEPENDENCY else DEPENDENCY_FEATURE
+            is RsLitExpr -> if (anyCfgFeature.accepts(element)) CFG_FEATURE else null
+            else -> null
+        }
+    }
+
+    companion object {
+        private val FEATURE_DEPENDENCY = UsageType { "Cargo feature dependency" }
+        private val DEPENDENCY_FEATURE = UsageType { "Package dependency" }
+        private val CFG_FEATURE = UsageType { "Cfg attribute" }
+    }
+}
--- a/toml/src/test/kotlin/org/rust/toml/search/CargoTomlFindUsagesTest.kt
+++ b/toml/src/test/kotlin/org/rust/toml/search/CargoTomlFindUsagesTest.kt
@@ -1,0 +1,131 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.search
+
+import com.intellij.codeInsight.TargetElementUtil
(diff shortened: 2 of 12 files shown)


toml/src/main/kotlin/org/rust/toml/search/CargoTomlUsageTypeProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.usages.UsageTarget
+import com.intellij.usages.impl.rules.UsageType
+import com.intellij.usages.impl.rules.UsageTypeProviderEx
+import org.rust.lang.core.RsPsiPattern.anyCfgFeature
+import org.rust.lang.core.psi.RsLitExpr
+import org.rust.toml.CargoTomlPsiPattern.onFeatureDependencyLiteral
+import org.toml.lang.psi.TomlLiteral
+
+object CargoTomlUsageTypeProvider : UsageTypeProviderEx {
```
Undin, line 17: Looks like it doesn't make sense to use `object` here since the platform will create another `CargoTomlUsageTypeProvider` instance via reflection anyway

toml/src/main/kotlin/org/rust/toml/search/CargoTomlUsageTypeProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val FEATURE_DEPENDENCY = UsageType { "Cargo feature dependency" }
+    private val DEPENDENCY_FEATURE = UsageType { "Package dependency" }
+    private val CFG_FEATURE = UsageType { "Cfg attribute" }
```
Undin, line 20: Let's move them into companion object after converting `object` to general class

toml/src/main/kotlin/org/rust/toml/search/CargoTomlUsageTypeProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    override fun getUsageType(element: PsiElement?): UsageType? = getUsageType(element, UsageTarget.EMPTY_ARRAY)
+
+    override fun getUsageType(element: PsiElement?, targets: Array<out UsageTarget>): UsageType? {
+        if (element == null) return null // Impossible?
```
Undin, line 25: Do we really need this line?

toml/src/test/kotlin/org/rust/toml/search/CargoTomlFindUsagesTest.kt
```diff
@@ -0,0 +1,132 @@
...
+            val actual = markersActual.remove(file) ?: error("Expected usages in $file, but found 0 usages")
+            assertEquals(expected.joinToString("\n"), actual.joinToString("\n"))
+        }
+
+        if (markersActual.isNotEmpty()) {
+            error("Extra usages in $markersActual")
+        }
+    }
+
+    private fun markersActual(source: PsiElement): List<Pair<Int, String>> =
```
Undin, line 115: unused

toml/src/test/kotlin/org/rust/toml/search/CargoTomlFindUsagesTest.kt
```diff
@@ -0,0 +1,132 @@
...
+            dir("src") {
+                rust("lib.rs", """
+                    #[cfg(feature = "feature_bar")] // - Cfg attribute
+                    #[cfg_attr(feature = "feature_bar", allow(all))] // - Cfg attribute
+                    fn foobar() {}
+                """)
+            }
+        }
+    }.run {
+        assertNotNull(project.cargoProjects.allProjects.single().workspace)
```
Undin, line 65: Maybe it's worth to extract code from `run` into a separate method? Currently, it's quite huge and cannot be reused if needed

toml/src/main/kotlin/org/rust/toml/search/CargoTomlUsageTypeProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+            } else {
+                DEPENDENCY_FEATURE
+            }
+            is RsLitExpr -> if (anyCfgFeature.accepts(element)) {
+                CFG_FEATURE
+            } else {
+                null
+            }
+            else -> null
+        }
```
Undin, line 39: nit: I like the following more but it's up to you

```suggestion
        return when (element) {
            is TomlLiteral -> if (onFeatureDependencyLiteral.accepts(element)) FEATURE_DEPENDENCY else DEPENDENCY_FEATURE
            is RsLitExpr -> if (onAnyCfgFeature.accepts(element)) CFG_FEATURE else null
            else -> null
        }
```

Undin (commented):
Rename doesn't work correctly in case of feature dependencies

Undin (approved):
bors r+

PR #6425, opened 2020-11-23
T: fix empty test fail if the test is `@ExpandMacros`

vlad20012:
bors r+


Undin:
@vlad20012 Thank you!

PR #6323, opened 2020-10-29
Validate perf profiler settings

ortem:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6173.

Now, in case of no `perf` path specified in settings, "Perf executable path is not set" error will be shown with a link to configure the profiler settings (the way how it works in CLion with C/C++).

validate-perf

PR #6427, opened 2020-11-24
Validate DTrace profiler settings

ortem:
See https://github.com/intellij-rust/intellij-rust/pull/6323 for more details.

changelog: Validate DTrace profiler settings when running code with CPU profiler on macOS


Undin (approved):
bors r+

PR #6429, opened 2020-11-24
REF(MOVE): Use unqualified path for references to target mod if possible

dima74:
E.g. when moving `fn foo() { crate::mod2::bar(); }` to `mod2`, we can replace absolute path `crate::mod2::bar` with just `bar`

changelog: Use unqualified path for references to target mod in move items refactoring


ortem (approved):
LGTM
bors r+

PR #6430, opened 2020-11-24
Fix wrong resolve of cargo feature in `cfg_attr`

vlad20012:
This "foo" was resolved as a cargo feature, although it shouldn't:
```
#[cfg_attr(foobar, foobar(feature = "foo"))]
```


Undin (commented):
Let's add a test where we shouldn't resolve features that you described in commit message

vlad20012:
There is a test for PSI pattern! But patter tests was broken before this PR

PR #6435, opened 2020-11-25
RES: Exclude private items from completion when using new resolve

dima74:
Currently private items are excluded from completion in both stable and [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217). And items which are public, but still can't be used (e.g. because they are privately reexported), are excluded in stable resolve only. So this PR excludes such items in new resolve too, for consistency with stable resolve.

changelog: Exclude private items from completion when using new resolve

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt
@@ -934,7 +934,6 @@ class RsCompletionTest : RsCompletionTestBase() {
         }
     """)
 
-    @IgnoreInNewResolve
     @ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
     fun `test private extern crate`() = checkNoCompletion("""
         mod foo { extern crate std; }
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt
```diff
@@ -934,7 +935,28 @@ class RsCompletionTest : RsCompletionTestBase() {
         }
     """)
 
-    @IgnoreInNewResolve
+    @MockEdition(Edition.EDITION_2018)
+    fun `test private reexport of public function`() = checkNoCompletion("""
```
vlad20012, line 939: Let's move these tests to `RsCompletionFilteringTest`

src/test/kotlin/org/rust/lang/core/completion/RsCompletionTest.kt
```diff
@@ -934,7 +935,28 @@ class RsCompletionTest : RsCompletionTestBase() {
...
+
+        fn main() {
+            mod2::b/*caret*/
+        }
+    """)
+
+    // there was error in new resolve when legacy textual macros are always completed
+    @MockEdition(Edition.EDITION_2018)
+    @ProjectDescriptor(WithStdlibRustProjectDescriptor::class)
+    fun `test no completion on empty mod`() = checkNoCompletion("""
```
vlad20012, line 955: Btw it is possible to write it without std. Can be written in addition to this test.

vlad20012 (approved):

dima74:
bors r=vlad20012

PR #6438, opened 2020-11-25
STUB: create stubs for lifetime parameters only if needed

Undin:
Original issue was found in [gimli](https://github.com/gimli-rs/gimli/blob/d36fcf11c878c8ebfb4da4cad300773869b5b5e4/examples/simple.rs#L38-L41) project that now is part of stdlib sources (stdlib dependency).

Also, print file url in `RsLazyBlockStubCreationTestBase` tests. It may help to find file with problem code and even open it in your IDE if it's located in local file system.

Fix the corresponding test in #6374 and #6417


vlad20012 (approved):
bors r+

PR #6443, opened 2020-11-27
REF(MOVE): Change references in use groups

dima74:
E.g. when moving `foo` from `mod1` to `mod2`, changing `use mod1::{foo, bar};` to `use mod2::foo; use mod1::bar;` is now supported.

changelog: Support changing references in use groups in `Move` refactoring (<kbd>F6</kbd>)

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/RsImportOptimizer.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/RsImportOptimizer.kt
@@ -39,7 +39,7 @@ class RsImportOptimizer : ImportOptimizer {
         externCrateItems.forEach { it.delete() }
     }
 
-    private fun executeForUseItem(mod: RsMod) {
+    fun executeForUseItem(mod: RsMod) {
         val uses = mod.childrenOfType<RsUseItem>()
         if (uses.isNotEmpty()) {
             replaceOrderOfUseItems(mod, uses)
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/refactoring/RsImportOptimizer.kt
```diff
@@ -39,7 +39,7 @@ class RsImportOptimizer : ImportOptimizer {
         externCrateItems.forEach { it.delete() }
     }
 
-    private fun executeForUseItem(mod: RsMod) {
+    fun executeForUseItem(mod: RsMod) {
```
ortem, line 42: [minor] Now it may be better to introduce public `executeForUseItem` before private `executeForExternCrate`

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveRetargetReferencesProcessor.kt
```diff
@@ -178,4 +186,64 @@ class RsMoveRetargetReferencesProcessor(
...
+
+    private fun addImport(context: RsElement, usePath: String) {
+        addImport(psiFactory, context, usePath)
+        filesToOptimizeImports.add(context.containingFile as RsFile)
+    }
+
+    fun optimizeImports() {
+        useSpecksToOptimize.forEach { RsImportOptimizer.optimizeUseSpeck(psiFactory, it) }
+
+        filesToOptimizeImports.addAll(useSpecksToOptimize.mapNotNull { it.containingFile as? RsFile })
```
ortem, line 246: I'm not sure why do you use safe case here and unsafe cast in other places (`context.containingFile as RsFile`, `containingMod.containingFile as RsFile`, etc.)? Is this case different from the others?

ortem (approved):
LGTM except for minor comments

src/main/kotlin/org/rust/ide/refactoring/move/common/RsMoveRetargetReferencesProcessor.kt
dima74: Thanks, fixed (it was different because `containingFile` can be `DummyHolder`)

src/main/kotlin/org/rust/ide/refactoring/RsImportOptimizer.kt
dima74: I think it is better to keep as-is, so that `executeForUseItem` and `optimizeUseSpeck` functions will be close to each other

PR #6447, opened 2020-11-28
T: test that Grazie doesn't check Rust code in doctest injections

vlad20012:


Undin (changes requested):
The current change adds regression for 202 platform exactly in case that you added in new test and suppressed for 202 platform for some reason.
`RsGrammarCheckingStrategy#isTypoAccepted` is called only on 202 platform where `Grazie` plugin doesn't take into account injections.
So, let's revert deletion of `isTypoAccepted ` method and drop `@IgnoreInPlatform(202)` annotation for the new test

Undin (approved):
bors r+

vlad20012:
bors retry

PR #6448, opened 2020-11-28
COMP: take into account `cfg_attr` attribute while attribute completion

Undin:
changelog: take into account `cfg_attr` attribute while attribute and lint attribute completion

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsAttributeCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsAttributeCompletionProvider.kt
@@ -6,7 +6,6 @@
 package org.rust.lang.core.completion
 
 import com.intellij.codeInsight.completion.CompletionParameters
-import com.intellij.codeInsight.completion.CompletionProvider
 import com.intellij.codeInsight.completion.CompletionResultSet
 import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.codeInsight.lookup.LookupElementBuilder
@@ -16,8 +15,7 @@ import com.intellij.patterns.PlatformPatterns.psiElement
 import com.intellij.psi.PsiElement
 import com.intellij.util.ProcessingContext
 import org.rust.ide.icons.RsIcons
-import org.rust.lang.RsLanguage
-import org.rust.lang.core.RsPsiPattern
+import org.rust.lang.core.RsPsiPattern.META_ITEM_ATTR
 import org.rust.lang.core.RsPsiPattern.onAnyItem
 import org.rust.lang.core.RsPsiPattern.onCrate
 import org.rust.lang.core.RsPsiPattern.onDropFn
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsAttributeCompletionProvider.kt
```diff
@@ -70,11 +65,12 @@ object RsAttributeCompletionProvider : RsCompletionProvider() {
...
-        val elem = parameters.position.superParent(RsPsiPattern.META_ITEM_IDENTIFIER_DEPTH)
+    override fun addCompletions(
+        parameters: CompletionParameters,
+        context: ProcessingContext,
+        result: CompletionResultSet
+    ) {
+        val elem = context[META_ITEM_ATTR]?.parent ?: return
 
         val suggestions = attributes
             .filter { it.appliesTo.accepts(parameters.position) && elem.attrMetaItems.none { item -> item == it.name } }
```
Undin, line 76: In fact, we can replace `it.appliesTo.accepts(parameters.position)` with `it.appliesTo.accepts(elem)` to reduce number of computation here because all our patterns in `attributes` list provide some restriction only for items but it:
- potentially restricts us from writing patterns that take into account attributes characteristics
- requires to refactor all our `on` patterns and the corresponding tests
so Im't not sure if it's worth it

src/main/kotlin/org/rust/lang/core/completion/RsAttributeCompletionProvider.kt
```diff
@@ -70,11 +65,12 @@ object RsAttributeCompletionProvider : RsCompletionProvider() {
...
-                                context: ProcessingContext,
-                                result: CompletionResultSet) {
-
-        val elem = parameters.position.superParent(RsPsiPattern.META_ITEM_IDENTIFIER_DEPTH)
+    override fun addCompletions(
+        parameters: CompletionParameters,
+        context: ProcessingContext,
+        result: CompletionResultSet
+    ) {
+        val elem = context[META_ITEM_ATTR]?.parent ?: return
```
vlad20012, line 73: Btw, this is incorrect in the case of an inner attribute inside a function body. We probably should use `RsAttr.owner`, but it's also buggy in the case of inline modules (for example) =/

src/main/kotlin/org/rust/lang/core/completion/RsAttributeCompletionProvider.kt
```diff
@@ -84,15 +80,9 @@ object RsAttributeCompletionProvider : RsCompletionProvider() {
...
-            val outerAttrElem = psiElement<RsOuterAttr>()
-            val innerAttrElem = psiElement<RsInnerAttr>()
-            val metaItemElem = psiElement<RsMetaItem>().and(
-                psiElement().withParent(outerAttrElem) or psiElement().withParent(innerAttrElem)
-            )
-
             return psiElement()
                 .withLanguage(RsLanguage)
-                .withParent(psiElement<RsPath>().withParent(metaItemElem))
+                .withParent(psiElement<RsPath>().withParent(rootMetaItem))
```
vlad20012, line 85: Let's (try to) replace the pattern to
```
psiElement(RsElementTypes.IDENTIFIER).withParent(psiElement<RsPath>().withParent(rootMetaItem))
```
just because it's more explicit and restrictive

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/completion/RsAttributeCompletionProvider.kt
vlad20012: Fixed `RsAttr.owner` in #6453

Undin: 👍

Undin:
bors r=vlad20012

Issue #6441, opened 2020-11-26
No support for in_band_lifetimes


CraftSpider:
## Environment

* **IntelliJ Rust plugin version:** 0.3.137.3503-203-nightly
* **Rust toolchain version:** rustc 1.50.0-nightly (a0d664bae 2020-11-22)
* **IDE name and version:** CLion 2020.3 Beta
* **Operating system:** Windows 10

## Problem description

An error is emitted on the following, despite `in_band_lifetime` being enabled:
```rust
impl Foo<'a> {
}
```

## Steps to reproduce

Create a crate, and add `#![feature(in_band_lifetimes)]` to it. Add the code shown above.

PR #6450, opened 2020-11-28
ANN & RES: Fix in-band lifetimes support

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6441.

changelog: Fix inspection and name resolution for in-band lifetimes

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -629,9 +629,13 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
 
         val owner = lifetime.ancestorStrict<RsGenericDeclaration>() ?: return
         val declarationParts = listOfNotNull(
-            owner.typeParameterList,
+            (owner as? RsImplItem)?.traitRef,
+            (owner as? RsImplItem)?.typeReference,
+            (owner as? RsImplItem)?.whereClause,
+
+            (owner as? RsFunction)?.typeParameterList,
             (owner as? RsFunction)?.valueParameterList,
-            owner.whereClause
+            (owner as? RsFunction)?.whereClause
         )
         val inDeclaration = lifetime.ancestors.takeWhile { it != owner }.any { it in declarationParts }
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/lang/core/resolve/RsResolveTest.kt
```diff
@@ -1129,6 +1129,17 @@ class RsResolveTest : RsResolveTestBase() {
...
     """)
 
+    @MockRustcVersion("1.23.0-nightly")
+    fun `test in-band lifetime resolve in impl`() = checkByCode("""
+        #![feature(in_band_lifetimes)]
+        trait T<'a> {}
+        struct S<'a>(&'a str);
+        impl T<'a> 
+              //X 
+        for T<'a> {}
```
Undin, line 1139: Do you mean `S` here instead of `T`?

mchernyavsky: Fixed.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -630,9 +630,11 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
 
         val owner = lifetime.ancestorStrict<RsGenericDeclaration>() ?: return
         val declarationParts = listOfNotNull(
-            owner.typeParameterList,
+            (owner as? RsImplItem)?.traitRef,
+            (owner as? RsImplItem)?.typeReference,
+            (owner as? RsFunction)?.typeParameterList,
             (owner as? RsFunction)?.valueParameterList,
-            owner.whereClause
+            (owner as? RsFunction)?.whereClause
```
Undin, line 637: I suppose we should check `whereClause` of `impl` blocks as well since:
```rust
#![feature(in_band_lifetimes)]
trait T {}
struct S<'a>(&'a str);
impl T for S<'_> where 'a : 'static {}
```
is allowed and 
```rust
#![feature(in_band_lifetimes)]
trait T {}
struct S<'a>(&'a str);
impl T for S where 'a : 'static {}
```
produces only 
```
error[E0726]: implicit elided lifetime not allowed here
 --> src/main.rs:4:12
  |
4 | impl T for S where 'a : 'static {}
  |            ^- help: indicate the anonymous lifetime: `<'_>`
```

Undin (commented):
In general looks ok

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
mchernyavsky: Fixed.

Undin (approved):
bors r+

PR #6453, opened 2020-11-29
Fix trait inner attributes and `RsAttr.owner`

vlad20012:
The change about `RsAttr.owner` most likely doesn't affect users because today it is used exactly in the cases where it works perfectly.

Take into account trait inner attributes like
```rust
trait Foo {
    #![allow(deprecated)]
}
```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
@@ -5,9 +5,7 @@
 
 package org.rust.lang.core.psi.ext
 
-import org.rust.lang.core.psi.RsInnerAttr
-import org.rust.lang.core.psi.RsMetaItem
-import org.rust.lang.core.psi.RsOuterAttr
+import org.rust.lang.core.psi.*
 
 interface RsAttr : RsElement {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsAttr.kt
```diff
@@ -5,18 +5,23 @@
...
-        is RsInnerAttr -> parent?.parent as? RsDocAndAttributeOwner
         is RsOuterAttr -> parent as? RsDocAndAttributeOwner
+        is RsInnerAttr -> when (val parent = parent) {
+            is RsMembers -> parent.parent as? RsDocAndAttributeOwner
+            is RsBlock -> when (val parentParent = parent.parent) {
+                is RsFunction -> parentParent
+                else -> parent as? RsDocAndAttributeOwner
+            }
+            else -> parent as? RsDocAndAttributeOwner
+        }
```
Undin, line 24: I wonder how we didn't notice it previously 😅

Undin (approved):
bors r+

Note, I removed changelog note since this change doesn't affect users currently and as a result, there isn't a reason to add it to changelog

PR #6459, opened 2020-11-30
Add Python typings for LLDB/GDB pretty-printers

ortem:
This PR adds Python [typings](https://docs.python.org/3/library/typing.html) for `lldb` and `gdb` internal modules. Pretty-printers should still work with any supported versions of LLDB and GDB, including versions linked with Python 2.7 (it's CLion's bundled LLDB for macOS [before 2020.3](https://youtrack.jetbrains.com/issue/CPP-19038)) and MSVC LLDB.

This should simplify maintenance of pretty-printers by providing Python code completion, error highlighting, and so on.

Also
* Cleanup both LLDB and GDB pretty-printers after adding these typings
* Disable some inspections on Python code inside the `prettyPrinters` module

To test:
- [x] LLDB on macOS with CLion 2020.3
- [x] LLDB on macOS with CLion 2021.1
- [x] GDB on Linux with CLion 2020.3
- [x] LLDB on Linux with CLion 2020.3
- [x] GDB on Linux with CLion 2021.1
- [x] LLDB on Linux with CLion 2021.1
- [x] MSVC LLDB on Windows

changelog: Add Python typings for our LLDB/GDB pretty-printers to enable code analysis and completion in the `prettyPrinters` module


ortem:
@Kobzol these changes might be interesting for you, by the way 😄

Kobzol:
It sure will be nice to have these typings when working on debugger visualisations. Thank you!

.idea/scopes/prettyPrinters.xml
```diff
@@ -0,0 +1,3 @@
+<component name="DependencyValidationManager">
+    <scope name="prettyPrinters" pattern="file[intellij-rust]:prettyPrinters/*.py*" />
```
Undin, line 2: Looks like it doesn't work properly - I still see weak warnings in `lldb/__init__.pyi`
image

Undin (commented):
LGTM in general

What's about type annotation in `gdb_lookup.py`?

.idea/scopes/prettyPrinters.xml
ortem: Well, I'm not sure why it doesn't work and how to fix it actually. For me, it works fine even when I check out this branch on another machine.

Could you try to open `.idea/inspectionProfiles/intellij_rust.xml` replace
`<scope name="prettyPrinters" level="INFORMATION" enabled="true" />` with
`<scope name="prettyPrinters" level="WEAK WARNING" enabled="false" />`?

prettyPrinters/lldb/__init__.pyi
```diff
@@ -0,0 +1,379 @@
+from typing import Any, List
+
+import formatters
+
+
+# Some useful interfaces from https://github.com/llvm/llvm-project/tree/master/lldb/bindings/interface
```
Undin, line 6: 404 when I try to open the page

Undin (approved):

prettyPrinters/lldb/__init__.pyi
ortem: Thanks for catching that, the branch was renamed [recently](https://foundation.llvm.org/docs/branch-rename/), and now `main` should be used instead

Undin: To prevent such things, you may use concrete commit instead of branch name

PR #6460, opened 2020-11-30
RES&COMP: resolve and complete derive attributes under `cfg_attr`

vlad20012:
Make resolve and complete works for `derive()` inside `cfg_attr`: `#[cfg_attr(foobar, derive(...))]`

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
@@ -5,6 +5,7 @@
 
 package org.rust.lang.core.completion
 
+import org.rust.MockAdditionalCfgOptions
 import org.rust.ProjectDescriptor
 import org.rust.WithStdlibRustProjectDescriptor
 import org.rust.lang.core.resolve.KnownDerivableTrait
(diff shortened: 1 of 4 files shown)


src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
```diff
@@ -128,4 +128,16 @@ class RsDeriveCompletionProviderTest : RsCompletionTestBase() {
...
+        #[cfg_attr(windows, derive(Debu/*caret*/))]
+        struct Test {
+            foo: u8
+        }
+    """, """
+        #[cfg_attr(windows, derive(Debug/*caret*/))]
+        struct Test {
+            foo: u8
+        }
+    """)
```
Undin, line 142: What do you think about filtering derive attributes in case of `cfg_attr`?
For example,
```rust
#[cfg_attr(windows, derive(Debu/*caret*/))]
#[cfg_attr(windows, derive(Debug))]
struct Test {
    foo: u8
}
```

In this case, obviously, we shouldn't suggest `Debug` but in common case I don't think we can take into account all possible combinations of options/features. So I suggest just take into account all impls according to current options/features.
If you agree, let's add such test to catch such case when cfg evaluator will take into account `cfg_attr`

Undin (approved):

src/test/kotlin/org/rust/lang/core/completion/RsDeriveCompletionProviderTest.kt
vlad20012: Added a test

Undin (approved):
bors r+

vlad20012:
bors r=undin

PR #6464, opened 2020-12-01
FIX: do not offer create method intention on foreign types

Kobzol:
The `CreateFunctionIntention` checks that the target module is from the user's workspace, but this check was missing for creating methods. This PR fixes that.

changelog: Create method intention is no longer offered for types from external or std crates.


vlad20012 (approved):
bors r+

PR #6466, opened 2020-12-01
Support function parameters under cfg attributes

### LINKED ISSUE #6148: Function arguments based on `#[cfg(feature = "xyz")]` are handled incorrectly
## Environment

* **IntelliJ Rust plugin version:** 0.3.132.3360-202-nightly
* **Rust toolchain version:** 1.46.0 (04488afe3 2020-08-24) x86_64-unknown-linux-gnu
* **IDE name and version:** IntelliJ IDEA 2020.2.2 Ultimate Edition (IU-202.7319.50)
* **Operating system:** Linux 5.4.0-47-generic

## Problem description

Consider this example:
```rust
fn main() {
    foo("123");
}

fn foo(#[cfg(windows)] a: i32, #[cfg(unix)] b: &str) {
    
}
```

It gives 2 false-positives:
1. mismatched types [E0308] expected `i32`, found `&str` 
2. This function takes 2 parameters but 1 parameter was supplied [E0061]

image

Also, the function is weirdly highlighted:

image
 
Similar to #6147

vlad20012:
Fixes #6148

Correctly handle function parameters under `cfg` attributes in type inference and annotators. Works for:

```rust
fn foo(
    #[cfg(foo)]      a: u8,
    #[cfg(not(foo))] a: i8,
) {}
```

Note that it doesn't works `implement members` quick fix/action because it doesn't support any attributes at all. it should be fixed separately.

`self` parameter can also be under cfg. This case is not covered.

### Implementation details

There is `RsFunction.valueParameters` that resturns function parameter list after `cfg` evaluation and  should be used in type inference/inspections/annotatos/etc, and there is `RsFunction.rawValueParameters` that resturns function parameter list **without** `cfg` evaluation and should be used in PSI manipulations

changelog: Support function parameters under `#[cfg()]` attributes

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/UnElideLifetimesIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/UnElideLifetimesIntention.kt
@@ -8,6 +8,7 @@ package org.rust.ide.intentions
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import org.rust.ide.intentions.UnElideLifetimesIntention.*
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.infer.hasReEarlyBounds
@@ -16,15 +17,15 @@ import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.ty.TyAdt
 import org.rust.lang.core.types.type
 
-class UnElideLifetimesIntention : RsElementBaseIntentionAction<RsFunction>() {
+class UnElideLifetimesIntention : RsElementBaseIntentionAction<LifetimeContext>() {
     override fun getText() = "Un-elide lifetimes"
     override fun getFamilyName(): String = text
 
-    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): RsFunction? {
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): LifetimeContext? {
         if (element is RsDocCommentImpl) return null
         val fn = element.ancestorOrSelf<RsFunction>(stopAt = RsBlock::class.java) ?: return null
 
-        val ctx = getLifetimeContext(fn)
+        val ctx = getLifetimeContext(fn) ?: return null
         val outputLifetimes = ctx.output?.lifetimes
         if (outputLifetimes != null) {
             if (outputLifetimes.any { it != null } || outputLifetimes.size > 1) return null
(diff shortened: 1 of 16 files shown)


Kobzol:
I wonder how should this be handled for https://github.com/intellij-rust/intellij-rust/pull/5944. Probably going to disallow the refactoring on such functions for start.

vlad20012:
In this PR I disabled `Unelide lifetimes` intention in such functions, for example. I think it's ok, taking into account the fact that [arameters under cfg is a very rare case

src/main/kotlin/org/rust/ide/intentions/UnElideLifetimesIntention.kt
```diff
@@ -118,17 +126,13 @@ private fun parsePotentialLifetimeType(ref: RsTypeReference): PotentialLifetimeR
...
-    val inputs: List<PotentialLifetimeRef>,
-    val output: PotentialLifetimeRef?,
-    val self: PotentialLifetimeRef?
-)
-
-private fun getLifetimeContext(fn: RsFunction): LifetimeContext {
-    val inputArgs = fn.valueParameters.mapNotNull { elem -> elem.typeReference?.let { parsePotentialLifetimeType(it) } }
+private fun getLifetimeContext(fn: RsFunction): LifetimeContext? {
+    val valueParameters = fn.rawValueParameters
+    if (valueParameters.any { it.queryAttributes.hasCfgAttr() }) return null
```
ortem, line 131: Let's add a comment about why the intention should be disabled in that case

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -73,7 +73,11 @@ val RsFunction.abiName: String?
         return stub?.abiName ?: abi?.stringLiteral?.text
     }
 
+/** Returns those function parameters that are not disabled by cfg attributes */
```
ortem, line 76: I'd add some description to help contributors understand which one should be used in their case. Probably something like that
```suggestion
/**
 * Those function parameters that are not disabled by cfg attributes.
 *
 * Should be used in code analysis: name resolution, type inference, inspections, annotations, etc.
 */
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -73,7 +73,11 @@ val RsFunction.abiName: String?
...
 val RsFunction.valueParameters: List<RsValueParameter>
+    get() = rawValueParameters.filter { it.isEnabledByCfgSelf }
+
```
ortem, line 79: Similarly, something like that would be useful IMO 

```suggestion

/**
 * All function parameters.
 *
 * Should be used in code (PSI) manipulations: intentions, quick-fixes, refactorings, code generation, etc.
 */
```

ortem (commented):
LGTM except for minor comments.
Also, I'm not sure about the names `valueParameters` and `rawValueParameters`. I'm afraid this may lead to choosing `valueParameters` by mistake as the most simple and obvious option. Probably we should avoid so short `valueParameters` name and use something like `actualValueParameters` (I haven't come up with a really good option yet). @Undin maybe you have some ideas?

ortem (approved):
Although we can choose a better name later (as we don't have a really good one for now), as well as improve all the code documentation about cfg stuff

vlad20012:
> Also, I'm not sure about the names valueParameters and rawValueParameters. I'm afraid this may lead to choosing valueParameters by mistake as the most simple and obvious option. Probably we should avoid so short valueParameters name and use something like actualValueParameters

Same about `RsFieldsOwner.namedFields` and `RsFieldsOwner.positionalFields`. If we come up with naming conventions for cfg-filtered PSI accessors, then let's rename all such things in a separate PR. (I think the real long-term solution is a next abstraction layer around PSI (aka HIR or, well, FIR), that is built after cfg evaluation)

bors r=ortem

PR #6471, opened 2020-12-02
INT: add support for associated functions in CreateFunctionIntention

Kobzol:
This PR enables `CreateFunctionIntention` for associated methods.
associated-method

changelog: Enable `CreateFunctionIntention` also for associated methods.


vlad20012:
I think, in Rust it's usually called `associated function`, not `associated method`. There are multiple usages of this term across the PR

vlad20012 (approved):
bors r+

Issue #6463, opened 2020-11-30
Code analysis with crates local index


avrong:
#### Completion
 - [ ] Features

#### Inspections
 - [ ] Crate is not found
 - [ ] Crate is found but version is not
 - [ ] Crate version is yanked
 - [ ] Version should not be empty
 - [ ] Feature is not found in crate

#### Suggestions
 - [ ] Suggest to bump version it if it is not newest
 - [ ] Suggest to bump version if it is yanked
 - [ ] Suggest to bump version to nearest one if a feature is not present in this one

#### Actions
- [ ] Update all dependencies to the latest major/minor/patch version

Kobzol:
It would be great if versions were also auto-completed inside the complex dependency specification:
```toml
[dependencies]
serde = { version = "/*here version is not completed*/" }
```

PR #6490, opened 2020-12-05
COMP: Add dependency features completion using crates local index

avrong:
Uses crates local index to suggest available features from crates local index if `org.rust.crates.local.index` experimental feature is enabled.

In contrast with current features completion, it doesn't need the project to be updated with new dependencies to perform completion.

Depends on #6390
Part of #6463


ortem:
Please rebase onto master

ortem:
@avrong Please rebase onto master

PR #6497, opened 2020-12-08
Support `cfg_attr` attributes

### LINKED ISSUE #3683: Incorrect inspection on mismatched types in function
## Environment

* **Intellij-Rust plugin version:** v0.2.96.2601-191-nightly
* **Rust toolchain version:** rustc 1.35.0-nightly (96d700f1b 2019-04-10)
* **IDE name and version:** CLion 2019.1

## Problem description
When writing a method to extract a `syn::Type` from another `syn::Type`, the type checker for the extension incorrectly infers the return type as a `&Type` instead of `Type`. The compiler accepts this input but the Intellij Rust plugin does not.

## Steps to reproduce

The following code snippet should result in the plugin displaying an error inspection:

```rust
fn get_type(syn_type: Type) -> Type {
    match syn_type {
        Type::Path(ref type_path) => {
            if type_path.path.segments[0].ident == "Option" {
                match type_path.path.segments[0].arguments {
                    PathArguments::AngleBracketed(ref angle_bracketed) => {
                        match angle_bracketed.args[0] {
                            GenericArgument::Type(ref arg_type) => arg_type.clone(),
                            _ => unreachable!(),
                        }
                    }
                    _ => unreachable!(),
                }
            } else {
                syn_type
            }
        }
        _ => unreachable!(),
    }
}
```

image

### LINKED ISSUE #5458: Unresolved reference for type that is available via module with `cfg_attr` with `path`
## Environment

* **Rust toolchain version:** 1.45.0-nightly (a74d1862d 2020-05-14) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.1.1 Ultimate Edition (IU-201.7223.91)
* **Operating system:** macOS 10.15.4

## Problem description

[This `Color` type](https://docs.rs/env_logger/0.7.1/env_logger/fmt/enum.Color.html) is not found in IntelliJ for code completion, and when used, is shown as red erroring with Unresolved reference.

The type is implemented in `env_logger` in `extern_impl.rs` that's referenced like so:
```rust
#[cfg_attr(feature = "termcolor", path = "extern_impl.rs")]
#[cfg_attr(not(feature = "termcolor"), path = "shim_impl.rs")]
mod imp;
```

fwiw `imp` is also red:
image

## Steps to reproduce

Offending line: https://github.com/apollographql/apollo-cli/blob/ba48c9f6af91a98bcf0b13fea026acfde1d36f0f/cli/src/log.rs#L24

check out that commit in that repo

oddly, the `use` line (line 6) isn't red.
```rust
use env_logger::fmt::{Color, Formatter};
```

image

vlad20012:
Now `cfg_attr` is mostly transparent for the plugin. The support includes (but is not limited to) these attributes:
* `derive` (`#[cfg_attr(windows, derive(Clone))]`, #3683 case)
* `path` (`#[cfg_attr(windows, path="win.rs")] mod foo;`, #5458 case)
* `cfg` (yes, `cfg` under `cfg_attr`: `#[cfg_attr(feature = "foo", cfg(feature = "bar"))]`)
* `proc_macro`, `proc_macro_attribute`, `proc_macro_derive`
* `test`, `bench`
* `no_std`, `no_core`
* `macro_use`, `macro_export`
* `no_main`, `may_dangle`, `non_exhaustive`, `doc(hidden)`, `deprecated`
* `lang`, `repr`, `unstable`

Documentation under `cfg_attr` is not supported for now (e.g. `#[cfg_attr(windows, doc = "...")]`).

### Implementation details

`RsDocAndAttributeOwner.queryAttributes` now yields meta items after `cfg_attr` expansion.

We often store attribute values to item PSI Stubs as an optimization (stub field access is much faster than PSI traversing). For example, we have `hasMacroUse` field that means that there is a `#[macro_use]` attribute on the item. The attribute can also be under cfg_attr: `#[cfg_attr(foobar, macro_use)]`. We can't evaluate `cfg` conditions during Stub building, so now we can't strictly say whether there is a `#[macro_use]` attribute or not. Now, instead of storing strict  `hasMacroUse`, we store a more fuzzy  `mayHaveMacroUse` flag. It's `true` when `#[macro_use]` or `#[cfg_attr(foobar, macro_use)]` is mentioned on the item. 

Fixes #5458, fixes #3683
Related to #1191

changelog: Take into account attributes under `#[cfg_attr]` attribute in type inference, name resolution, and the rest analysis.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
@@ -6,15 +6,19 @@
 package org.rust.lang.core.psi.ext
 
 import com.intellij.extapi.psi.StubBasedPsiElementBase
+import com.intellij.openapi.project.DumbService
 import com.intellij.openapi.util.registry.Registry
 import com.intellij.psi.NavigatablePsiElement
 import com.intellij.psi.PsiElement
+import com.intellij.psi.StubBasedPsiElement
 import com.intellij.psi.stubs.StubElement
 import org.rust.lang.core.crate.Crate
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.stubs.RsAttributeOwnerStub
 import org.rust.lang.utils.evaluation.CfgEvaluator
 import org.rust.lang.utils.evaluation.ThreeValuedLogic
+import org.rust.openapiext.testAssert
+import org.rust.stdext.nextOrNull
 
 interface RsDocAndAttributeOwner : RsElement, NavigatablePsiElement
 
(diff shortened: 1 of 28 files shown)


src/main/kotlin/org/rust/lang/core/resolve2/ModCollectorBase.kt
```diff
@@ -151,12 +151,30 @@ class ModCollectorBase private constructor(
...
-            hasMacroExport = def.hasMacroExport,
-            hasLocalInnerMacros = def.hasMacroExportLocalInnerMacros
+            hasMacroExport = HAS_MACRO_EXPORT_PROP.getByStub(def, crate),
+            hasLocalInnerMacros = HAS_MACRO_EXPORT_LOCAL_INNER_MACROS_PROP.getByStub(def, crate)
         )
         visitor.collectMacroDef(defLight)
     }
 
+    private val RsAttributeOwnerStubBase<out RsDocAndAttributeOwner>.hasMacroUse: Boolean
+        get() = psi.getQueryAttributes(crate, this).hasAttribute("macro_use")
```
dima74, line 161:
```suggestion
    private val RsAttributeOwnerStubBase<out RsDocAndAttributeOwner>.hasMacroUse: Boolean
        get() = mayHaveMacroUse && psi.getQueryAttributes(crate, this).hasAttribute("macro_use")
```

src/main/kotlin/org/rust/lang/core/resolve2/ModCollectorBase.kt
```diff
@@ -151,12 +151,30 @@ class ModCollectorBase private constructor(
...
+
+    private val RsAttributeOwnerStubBase<out RsDocAndAttributeOwner>.pathAttribute: String?
+        get() = psi.getQueryAttributes(crate, this).lookupStringValueForKey("path")
+
+    private val RsFunctionStub.isProcMacroDef: Boolean
+        get() = IS_PROC_MACRO_DEF_PROP.getByStub(this, crate)
+
+    val RsFunctionStub.procMacroName: String?
```
dima74, line 169:
```suggestion
    private val RsFunctionStub.procMacroName: String?
```

dima74 (approved):

src/main/kotlin/org/rust/lang/core/resolve2/ModCollectorBase.kt
```diff
@@ -151,12 +151,30 @@ class ModCollectorBase private constructor(
...
-            hasMacroExport = def.hasMacroExport,
-            hasLocalInnerMacros = def.hasMacroExportLocalInnerMacros
+            hasMacroExport = HAS_MACRO_EXPORT_PROP.getByStub(def, crate),
+            hasLocalInnerMacros = HAS_MACRO_EXPORT_LOCAL_INNER_MACROS_PROP.getByStub(def, crate)
         )
         visitor.collectMacroDef(defLight)
     }
 
+    private val RsAttributeOwnerStubBase<out RsDocAndAttributeOwner>.hasMacroUse: Boolean
+        get() = mayHaveMacroUse && psi.getQueryAttributes(crate, this).hasAttribute("macro_use")
```
dima74, line 161: Potentially we can now introduce `hasMacroUse` for `RsModItemStub`, `RsModDeclItemStub` and `RsExternCrateItemStub` using their `HAS_MACRO_USE_PROP`, but I am not sure if it worth it

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -79,24 +83,120 @@ inline val RsDocAndAttributeOwner.attributeStub: RsAttributeOwnerStub?
...
+/** [explicitCrate] is passed to avoid resolve triggering */
+fun RsDocAndAttributeOwner.getQueryAttributes(
+    explicitCrate: Crate?,
+    stub: RsAttributeOwnerStub? = attributeStub
+): QueryAttributes {
+    testAssert { !DumbService.isDumb(project) }
+    return when {
+        stub?.hasAttrs == false -> QueryAttributes.EMPTY
+        stub?.hasCfgAttr == false -> rawAttributesUnchecked
+        else -> getQueryAttributesNoStub(explicitCrate)
```
ortem, line 97: Maybe add a comment on why we should distinguish between these three cases? I mean, to notice that only in case of `cfg_attr`, we really have to evaluate something

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -79,24 +83,120 @@ inline val RsDocAndAttributeOwner.attributeStub: RsAttributeOwnerStub?
...
+            } else {
+                emptySequence()
+            }
+        } else {
+            sequenceOf(it)
+        }
+    }
+}
+
+fun RsDocAndAttributeOwner.getTraversedRawAttributes(withCfgAttr: Boolean): QueryAttributes {
```
ortem, line 125: [minor] `cfgAttr` name can be treated as "cfg attribute", not "cfg_attr attribute". Maybe use `withCfgAttrAttribute` name or even something else?

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
```diff
@@ -156,6 +251,10 @@ class QueryAttributes(
     val langAttribute: String?
```
ortem, line 251: Do we want to use this field after introducing `langAttributes` (I'm just not sure)?

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -22,32 +22,20 @@ import org.rust.lang.core.types.ty.TyUnknown
...
-        val stub = greenStub
-        return stub?.isTest
-            ?: (queryAttributes.isTest)
-    }
+    get() = queryAttributes.isTest
 
-val QueryAttributes.isTest
-    get() = hasAttribute(FUNCTION_TEST_REGEX) || hasAtomAttribute("quickcheck")
+private val QueryAttributes.isTest
+    get() = metaItems.mapNotNull { it.path?.referenceName }.any { it.endsWith("test") } || hasAtomAttribute("quickcheck")
```
ortem, line 35: Is this correct? `FUNCTION_TEST_REGEX` allowed items like `footestbar`, but this does not. Do we have any tests on this?

ortem:
Discussed in person that https://github.com/intellij-rust/intellij-rust/pull/6549 should be merged right after this because now we might face the following problem (and others):
```
#[cfg_attr(unknown_option, derive(SomethingBreaking))]] // breaks `Foo` usages somehow
struct Foo;
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
ortem: @Kobzol looks like `FUNCTION_TEST_REGEX` was introduced by your commit, maybe you have some observations as well?

Kobzol: `FUNCTION_TEST_REGEX` was introduced to support cases like `#[tokio::test(flavor = "multi_thread", worker_threads = 1)]`. It looks like iterating through the meta items and finding anything that resembles `test` should be fine.

The corresponding test is `test custom test attribute with underscore`.

Undin: `it.endsWith("test")` breaks support for custom test frameworks (see https://doc.rust-lang.org/beta/unstable-book/language-features/custom-test-frameworks.html) that currently use `test_case` attribute. I know that we didn't implement anything special for them (I didn't even know about it until recently) and we don't have the corresponding test(s) but anyway

vlad20012: Let's use `contains("test")`?

src/main/kotlin/org/rust/lang/core/psi/ext/RsDocAndAttributeOwner.kt
vlad20012: Yep, `langAttributes` is used only during stubs building. After cfg_attr evaluation there should exist only one lang item

vlad20012:
bors r=ortem

ortem (approved):

PR #6501, opened 2020-12-11
INSP: make items from std appear first when auto import

gfreezy:
changelog: make items from std appear first when auto import

when have `async-std` as dependency, auto import `Arc` should use `std::sync::Arc` instead of `async_std::sync::Arc`

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/import/ImportCandidatesCollector.kt
+++ b/src/main/kotlin/org/rust/ide/utils/import/ImportCandidatesCollector.kt
@@ -61,6 +61,8 @@ object ImportCandidatesCollector {
             // check that result after import can be resolved and resolved element is suitable
             // if no, don't add it in candidate list
             .filter { canBeResolvedToSuitableItem(importingPathText, importContext, it.info) }
+            // make items from workspace appear first, std seconds.
+            .sorted()
     }
 
     private fun getReexportedItems(
@@ -129,6 +131,8 @@ object ImportCandidatesCollector {
             .flatMap { it.withModuleReexports(project).asSequence() }
             .mapNotNull { importItem -> importItem.toImportCandidate(superMods) }
             .filterImportCandidates(attributes)
+            // make items from workspace appear first, std seconds.
+            .sorted()
     }
 
     private fun QualifiedNamedItem.toImportCandidate(superMods: LinkedHashSet<RsMod>): ImportCandidate? =
@@ -350,7 +354,22 @@ sealed class ImportInfo {
     }
 }
 
-data class ImportCandidate(val qualifiedNamedItem: QualifiedNamedItem, val info: ImportInfo)
+data class ImportCandidate(val qualifiedNamedItem: QualifiedNamedItem, val info: ImportInfo) : Comparable<ImportCandidate> {
+    override fun compareTo(other: ImportCandidate): Int {
+        val origin = qualifiedNamedItem.containingCrate?.origin
+        val otherOrigin = other.qualifiedNamedItem.containingCrate?.origin
+        val usePath = info.usePath
+        val otherUsePath = other.info.usePath
+        return when {
+            origin == otherOrigin -> usePath.compareTo(otherUsePath)
+            origin == PackageOrigin.WORKSPACE -> -1
+            otherOrigin == PackageOrigin.WORKSPACE -> 1
+            origin == PackageOrigin.STDLIB -> -1
+            otherOrigin == PackageOrigin.STDLIB -> 1
+            else -> usePath.compareTo(otherUsePath)
+        }
+    }
+}
 
 /**
  * If function or constant is defined in a trait
--- a/src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixStdTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixStdTest.kt
@@ -692,4 +692,54 @@ class AutoImportFixStdTest : AutoImportFixTestBase() {
 
         fn foo(x: FooBar/*caret*/) {}
     """)
+
+    @MockEdition(CargoWorkspace.Edition.EDITION_2018)
+    fun `test import item from workspace over std and extern crate`() = checkAutoImportFixByFileTreeWithMultipleChoice("""
+        //- dep-lib/lib.rs
+        pub mod foo {
+            pub struct Rc<T> {
+                v: T
+            }
+        }
+
+        //- main.rs
+        mod bar {
+            pub struct Rc<T> {
+                v: T
+            }
+        }
+
+        fn foo(t: <error descr="Unresolved reference: `Rc`">Rc/*caret*/</error><usize>) {}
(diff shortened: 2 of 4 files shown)


gfreezy:
@Undin i have no idea how to fix the `regressions`.

src/main/kotlin/org/rust/ide/utils/import/ImportCandidatesCollector.kt
```diff
@@ -61,6 +61,14 @@ object ImportCandidatesCollector {
             // check that result after import can be resolved and resolved element is suitable
             // if no, don't add it in candidate list
             .filter { canBeResolvedToSuitableItem(importingPathText, importContext, it.info) }
+            // make std appear first
+            .sortedBy {
+                if (it.qualifiedNamedItem.containingCrate?.origin == PackageOrigin.STDLIB) {
```
Undin, line 66: What's about code from your workspace?
If we want to make any sorting here, we want to put items from workspace first

src/main/kotlin/org/rust/ide/utils/import/ImportCandidatesCollector.kt
```diff
@@ -61,6 +61,14 @@ object ImportCandidatesCollector {
...
+                    "0${it.qualifiedNamedItem}"
+                } else {
+                    "1${it.qualifiedNamedItem}"
```
Undin, line 69: The current approach creates a new string object on each comparison. Let's try to avoid it.
You can create a composite comparator to compare `origin` first and only after that compare qualified paths.
`compareBy` and `thenBy` should help you to create the corresponding comparator and `sortedWith` method to use it.
Also, I'd recommend not to use `it.qualifiedNamedItem.toString()` to get qualified name (currently it's used implicitly to construct `"1${it.qualifiedNamedItem}"` string) and use `ImportInfo.usePath` for it

Undin: As an alternative, I can suggest implementing `Comparable` for `ImportCandidate` class

src/main/kotlin/org/rust/ide/utils/import/ImportCandidatesCollector.kt
```diff
@@ -129,6 +137,14 @@ object ImportCandidatesCollector {
...
             .mapNotNull { importItem -> importItem.toImportCandidate(superMods) }
             .filterImportCandidates(attributes)
+            // make std appear first
+            .sortedBy {
+                if (it.qualifiedNamedItem.containingCrate?.origin == PackageOrigin.STDLIB) {
+                    "0${it.qualifiedNamedItem}"
+                } else {
+                    "1${it.qualifiedNamedItem}"
+                }
+            }
```
Undin, line 147: Exactly the same code as above. Let's create single comparator for `ImportCandidate` or implement `Comparable` interface for `ImportCandidate`

Undin (changes requested):
It would be great to provide tests.
See `org.rust.ide.inspections.import.AutoImportFixTestBase` as base class for all auto import tests. I suppose `org.rust.ide.inspections.import.AutoImportFixTestBase#checkAutoImportFixByTextWithMultipleChoice` is most suitable to check order (but it requires to change `Set` to `List` as type of `expectedElements` collection)

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixStdTest.kt
```diff
@@ -692,4 +692,44 @@ class AutoImportFixStdTest : AutoImportFixTestBase() {
...
+            pub struct AtomicUsize;
+        }
+
+        //- main.rs
+        mod bar {
+            pub struct AtomicUsize;
+        }
+
+        fn foo(t: <error descr="Unresolved reference: `AtomicUsize`">AtomicUsize/*caret*/</error>) {}
+    """, listOf("crate::bar::AtomicUsize", "dep_lib_target::foo::AtomicUsize", "std::sync::atomic::AtomicUsize"),
```
gfreezy, line 710:
```
ERROR: expected:<[crate::bar::AtomicUsize, dep_lib_target::foo::AtomicUsize, std::sync::atomic::AtomicUsize]> but was:<[crate::bar::AtomicUsize, dep_lib_target::foo::AtomicUsize]>
```

why `std` is not available?

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixStdTest.kt
```diff
@@ -692,4 +692,44 @@ class AutoImportFixStdTest : AutoImportFixTestBase() {
...
+        "crate::bar::AtomicUsize","""
+        //- main.rs
+        use crate::bar::AtomicUsize;
+
+        fn foo(t: AtomicUsize/*caret*/) {}
+    """,
+        Testmarks.autoInjectedStdCrate)
+
+    @MockEdition(CargoWorkspace.Edition.EDITION_2018)
+    fun `test import item from std over extern crate`() = checkAutoImportFixByFileTreeWithMultipleChoice("""
```
gfreezy, line 720: failed with the same problem.

src/test/kotlin/org/rust/ide/inspections/import/AutoImportFixStdTest.kt
Undin: Because all `Atomic*` structs are generated by macros that are not expanded in these tests. To make it work, we use `ExpandMacros` annotation. But I suggest not to use it because it slows down test execution without actual necessity.
Let's just use another stdlib struct, for example, `std::sync::Arc`

Undin (commented):
@gfreezy Could you adjust failed tests, please?. As I can see, they fail because we started to check the order of suggested items and in these tests order is wrong

gfreezy:
image

@Undin need help to fix the regression

src/main/kotlin/org/rust/ide/utils/import/ImportCandidatesCollector.kt
```diff
@@ -350,7 +354,22 @@ sealed class ImportInfo {
     }
 }
 
-data class ImportCandidate(val qualifiedNamedItem: QualifiedNamedItem, val info: ImportInfo)
+data class ImportCandidate(val qualifiedNamedItem: QualifiedNamedItem, val info: ImportInfo) : Comparable<ImportCandidate> {
+    override fun compareTo(other: ImportCandidate): Int {
+        val origin = qualifiedNamedItem.containingCrate?.origin
+        val otherOrigin = other.qualifiedNamedItem.containingCrate?.origin
+        val userPath = info.usePath
+        val otherUsePath = info.usePath
```
Undin, line 362:
```suggestion
        val otherUsePath = other.info.usePath
```
It's a reason of weird candidates order in tests

Undin (changes requested):
> @Undin need help to fix the regression

Unfortunately, `regression-2` is broken currently for some reason. So, just ignore it.

Could you squash commits into a single one after all necessary fixes, please?

Undin (approved):
Thanks!
bors r+

Issue #5414, opened 2020-05-21
Tests are executed in the wrong working directory when debugging


mahkoh:
## Environment

* **IntelliJ Rust plugin version:** 0.2.122.3105-201
* **Rust toolchain version:** Nightly
* **IDE name and version:** CL-201.7223.86
* **Operating system:** Arch linux

## Problem description

When debugging, tests appear to always run in the working directory of the workspace instead of the working directory of the concrete cargo project. Tests that reference relative paths will fail.

## Steps to reproduce

https://github.com/mahkoh/irtestcase

Works:

```
cargo test --workspace
```

Fails:

Try to debug the test in CLion

PR #6504, opened 2020-12-11
RUN: Fix working directory when run `cargo test` in debug mode

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/5414

(works only if `org.rust.cargo.build.tool.window` is enabled)

changelog: Fix working directory when run `cargo test` in debug mode

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -47,7 +47,8 @@ interface CargoWorkspace {
 
     val featureGraph: FeatureGraph
 
-    fun findPackage(name: String): Package? = packages.find { it.name == name || it.normName == name }
+    fun findPackageById(id: PackageId): Package? = packages.find { it.id == id }
+    fun findPackageByName(name: String): Package? = packages.find { it.name == name || it.normName == name }
 
     fun findTargetByCrateRoot(root: VirtualFile): Target?
     fun isCrateRoot(root: VirtualFile) = findTargetByCrateRoot(root) != null
@@ -67,6 +68,7 @@ interface CargoWorkspace {
         val contentRoot: VirtualFile?
         val rootDirectory: Path
 
+        val id: String
         val name: String
         val normName: String get() = name.replace('-', '_')
 
@@ -157,10 +159,13 @@ interface CargoWorkspace {
         Unclassified(null),
 
         Stdlib("stdlib?"),
+
         // [dependencies]
         Normal(null),
+
         // [dev-dependencies]
         Development("dev"),
+
         // [build-dependencies]
         Build("build")
     }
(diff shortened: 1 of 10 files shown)


mchernyavsky:
CC @sirgl

Undin (commented):
Could you rebase onto master and fix conflicts, please?

mchernyavsky:
@Undin Done.

src/main/kotlin/org/rust/cargo/runconfig/RsExecutableRunner.kt
```diff
@@ -74,9 +78,20 @@ abstract class RsExecutableRunner(
 
         val runCargoCommand = state.prepareCommandLine()
         val (_, executableArguments) = parseArgs(runCargoCommand.command, runCargoCommand.additionalArguments)
+
+        val workingDirectory = if (runCargoCommand.command == "test") {
+            val srcPath = artifact?.target?.src_path
+            val file = srcPath?.let { LocalFileSystem.getInstance().refreshAndFindFileByPath(it) }
```
Undin, line 84: Let's avoid using VFS here because it has some contract how it should be called not to get dead lock. As I can see, artifact json object has `package_id` field (at least at 1.48 and above, i.e. when build view works in the plugin) so we can use this id to find the corresponding package and avoid using VFS at all

src/main/kotlin/org/rust/cargo/runconfig/RsExecutableRunner.kt
```diff
@@ -74,9 +78,20 @@ abstract class RsExecutableRunner(
...
+            val pkg = file?.let { environment.project.cargoProjects.findPackageForFile(it) }
+            val rootDirectory = pkg?.rootDirectory
+            rootDirectory ?: runCargoCommand.workingDirectory
+        } else {
+            runCargoCommand.workingDirectory
+        }
```
Undin, line 90: nit: Let's extract into separate method

Undin (changes requested):
It's a pity that we still don't have debugger tests 😞 
Anyway, works for me

src/main/kotlin/org/rust/cargo/runconfig/RsExecutableRunner.kt
mchernyavsky: Do you mean something like this?

```rust
cargoProjects.allProjects.mapNotNull { it.workspace?.findPackage(packageId) }.singleOrNull()
```

Undin: Yep

mchernyavsky:
@Undin Done.

src/main/kotlin/org/rust/cargo/runconfig/RsExecutableRunner.kt
Undin: I don't insist but just note that getting a proper working directory is still no extracted to separate method

Undin (approved):

src/main/kotlin/org/rust/cargo/runconfig/RsExecutableRunner.kt
mchernyavsky: Oh, I forgot about it. Fixed.

src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
```diff
@@ -67,6 +68,7 @@ interface CargoWorkspace {
         val contentRoot: VirtualFile?
         val rootDirectory: Path
 
+        val id: String
```
Undin, line 71:
```suggestion
        val id: PackageId
```

Undin (approved):
👍
bors r+

PR #6506, opened 2020-12-12
ANN: Provides case-checking for non-ascii name

### LINKED ISSUE #6480: Naming convention checks have weird behavior on Chinese characters
## Environment

* **IntelliJ Rust plugin version:** 0.3.136.3514-203
* **Rust toolchain version:** rustc 1.49.0-nightly (ffa2e7ae8 2020-10-24)
* **IDE name and version:** CLion 2020.3 Build #CL-203.5981.166, built on December 2, 2020
* **Operating system:** Linux (manjaro), kernel 5.9.11-3-MANJARO

## Problem description

Rust suggests using camel case for types and traits, and snake case for functions and variables. For both English and Russian identifiers, both the Rust compiler and intellij-rust give correct hints:

```
warning: type `my_error` should have an upper camel case name
  --> FILENAME:LL:CC
   |
LL | pub struct my_error ();
   |            ^^^^^^^^ help: convert the identifier to upper camel case: `MyError`
   |
   = note: `#[warn(non_camel_case_types)]` on by default

warning: type `ошибка_моя` should have an upper camel case name
  --> FILENAME:LL:CC
   |
LL | pub struct ошибка_моя ();
   |            ^^^^^^^^^^ help: convert the identifier to upper camel case: `ОшибкаМоя`
```

english

russian

But what about Chinese characters? there's no "uppercase" or "lowercase" in Chinese. Rust compiler simply ignores Chinese identifiers, while intellij-rust still produces the following hint:

chinese

Since there's no "uppercase" or "lowercase", the plugin in effect renames nothing.

## Steps to reproduce

Just try several Chinese identifiers. You may copy and paste the following snippet if you don't have a Chinese input method.

```rust
#![feature(non_ascii_idents)] // nightly only

pub struct 我的错误 ();
```

Stzx:
The PR provides detection for non-ASCII naming and suppresses `CamelCase`, `SnakeCase`, `UpperCase` detection if case is not supported.

CamelCase rules from https://github.com/rust-lang/rust/blob/master/compiler/rustc_lint/src/nonstandard_style.rs

Fixed #6480

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
@@ -36,10 +36,16 @@ abstract class RsNamingInspection(
         holder.registerLintProblem(
             id,
             "$elementType `$name` should have $styleName case name such as `$suggestedName`",
-            *fixes)
+            *fixes
+        )
     }
 
     abstract fun checkName(name: String): Pair<Boolean, String?>
+
+    companion object {
+        @JvmStatic
+        protected val OK = Pair(true, null)
+    }
 }
 
 /**
@@ -54,10 +60,11 @@ open class RsCamelCaseNamingInspection(
 
     override fun checkName(name: String): Pair<Boolean, String?> {
         val str = name.trim('_')
-        if (!str.isEmpty() && str[0].canStartWord && '_' !in str) {
-            return Pair(true, null)
+        return if (str.isCamelCase()) {
+            OK
+        } else {
+            Pair(false, if (str.isEmpty()) "CamelCase" else suggestName(name))
         }
-        return Pair(false, if (str.isEmpty()) "CamelCase" else suggestName(name))
     }
 
     private fun suggestName(name: String): String {
@@ -82,6 +89,17 @@ open class RsCamelCaseNamingInspection(
     }
 
     private val Char.canStartWord: Boolean get() = isUpperCase() || isDigit()
+
+    companion object {
+
+        private val Char.hasCase: Boolean get() = isLowerCase() || isUpperCase()
+
+        private fun String.isCamelCase(): Boolean =
+            isNotEmpty()
+                && !first().isLowerCase()
+                && !contains("__")
+                && !zipWithNext().any { (fst, snd) -> (fst.hasCase && snd == '_') || (snd.hasCase && fst == '_') }
+    }
 }
 
 /**
@@ -93,9 +111,11 @@ open class RsSnakeCaseNamingInspection(elementType: String) : RsNamingInspection
 
     override fun checkName(name: String): Pair<Boolean, String?> {
         val str = name.trim('_')
-        if (!str.isEmpty() && str.all { !it.isLetter() || it.isLowerCase() }) {
-            return Pair(true, null)
+        // Some characters don't have case so we can't use `isLowerCase` here
+        if (str.isNotEmpty() && str.all { !it.isUpperCase() }) {
+            return OK
         }
+
         return Pair(false, if (str.isEmpty()) "snake_case" else name.toSnakeCase(false))
     }
 }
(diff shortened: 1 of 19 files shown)


src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
```diff
@@ -36,10 +36,17 @@ abstract class RsNamingInspection(
...
             "$elementType `$name` should have $styleName case name such as `$suggestedName`",
-            *fixes)
+            *fixes
+        )
     }
 
     abstract fun checkName(name: String): Pair<Boolean, String?>
+
+    companion object {
+        internal val OK = Pair(true, null)
```
Undin, line 46:
```suggestion
        @JvmStatic
        protected val OK = Pair(true, null)
```

`internal` modifier doesn't usually make sense in non library code since it's the same as public for current module

src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
```diff
@@ -36,10 +36,17 @@ abstract class RsNamingInspection(
...
+
+        internal val Char.hasCase: Boolean get() = isLowerCase() || isUpperCase()
```
Undin, line 48: Looks like we actually need this property only in `RsCamelCaseNamingInspection`, so let's move it into `RsCamelCaseNamingInspection` and make it private

src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
```diff
@@ -109,8 +127,8 @@ open class RsUpperCaseNamingInspection(elementType: String) : RsNamingInspection
 
     override fun checkName(name: String): Pair<Boolean, String?> {
         val str = name.trim('_')
-        if (!str.isEmpty() && str.all { !it.isLetter() || it.isUpperCase() }) {
-            return Pair(true, null)
+        if (str.isNotEmpty() && str.all { !it.hasCase || !it.isLetter() || it.isUpperCase() }) {
```
Undin, line 130:
```suggestion
        if (str.isNotEmpty() && str.none { it.isLowerCase() }) {
```

I think it does the same but more readable

src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
```diff
@@ -93,9 +110,10 @@ open class RsSnakeCaseNamingInspection(elementType: String) : RsNamingInspection
 
     override fun checkName(name: String): Pair<Boolean, String?> {
         val str = name.trim('_')
-        if (!str.isEmpty() && str.all { !it.isLetter() || it.isLowerCase() }) {
-            return Pair(true, null)
+        if (str.isNotEmpty() && str.all { !it.hasCase || !it.isLetter() || it.isLowerCase() }) {
```
Undin, line 113:
```suggestion
        // Some characters don't have case so we can't use `isLowerCase` here
        if (str.isNotEmpty() && str.all { !it.isUpperCase() }) {
```

src/main/kotlin/org/rust/ide/inspections/lints/RsNamingInspection.kt
```diff
@@ -82,6 +90,15 @@ open class RsCamelCaseNamingInspection(
...
 
     private val Char.canStartWord: Boolean get() = isUpperCase() || isDigit()
+
+    companion object {
+
+        private fun String.isCamelCase(): Boolean =
+            isNotEmpty()
+                && !this[0].isLowerCase()
+                && !this.contains("__")
+                && !this.zipWithNext().any { (fst, snd) -> (fst.hasCase && snd == '_') || (snd.hasCase && fst == '_') }
```
Undin, line 100:
```suggestion
                && !first().isLowerCase()
                && !contains("__")
                && !zipWithNext().any { (fst, snd) -> (fst.hasCase && snd == '_') || (snd.hasCase && fst == '_') }
```

Undin (approved):
Thanks!
bors r+

Undin:
bors retry

PR #6507, opened 2020-12-12
Disable `cfg(test)` in external dependencies if there is `cfg(not(test))`

vlad20012:
Consider a code under `#[cfg(test)]` disabled in external dependencies (I.e. outside of a cargo workspace). Previously it was considered "enabled and disabled simultaneously", i.e. a code under `#[cfg(test)]` and `#[cfg(not(test))]` both was enabled.

The issue is that `#[cfg(test)]` sometimes used to mock something for testing:
```rust
#[cfg(test)]
mod foo {
    // Some mocks here
}
#[cfg(not(test))]
use foobar::foo; // Import the production variant of `foo`
```
So, we need `cfg(test)` to be disabled in order to provide correct name resolution.

But more often `cfg(test)` is used placed on test modules like this:
```rust
#[cfg(test)]
mod test {
    #[test]
    fn some_test() {}
}
```
So, if we disable `cfg(test)`, we make all tests grayed out!

The question "what to do with `cfg(test)`" was discussed for a long time. This PR starts with disabling `cfg(test)` in external dependency if that dependency has `not(test)` cfg condition in any context (like simple `#[cfg(not(test))]` or more complex `#[cfg(and(or(not(test), not(bar)), baz))]`). Yes, it makes it harder to read the test code in such dependencies. But firstly, this affects a minority of crates. Secondly, why usually someone reads tests in external dependencies? I think it's because they want to find examples of some library functionality. And there are doctests for this case, and we support doctests very well! So we decided that disabling `cfg(test)` in external dependencies should not be a real problem for our users.

The question "what to do with `cfg(test)` in workspace packages" is not answered yet. We think there should be some UI for enabling/disabling `cfg(test)` _in workspace packages_ like cargo features, but we don't imagine such UI yet.

changelog: Consider a code under `#[cfg(test)]` disabled in external dependencies (I.e. outside of a cargo workspace) that have `cfg(not(test))` cfg conditions. Previously it was considered "enabled and disabled simultaneously", i.e. a code under `#[cfg(test)]` and `#[cfg(not(test))]` both was enabled.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
+++ b/src/main/kotlin/org/rust/cargo/project/workspace/CargoWorkspace.kt
@@ -5,6 +5,8 @@
 
 package org.rust.cargo.project.workspace
 
+import com.intellij.openapi.util.UserDataHolderBase
+import com.intellij.openapi.util.UserDataHolderEx
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapiext.isUnitTestMode
@@ -61,7 +63,7 @@ interface CargoWorkspace {
     fun withCfgOptions(cfgOptions: CfgOptions): CargoWorkspace
 
     /** See docs for [CargoProjectsService] */
-    interface Package {
+    interface Package : UserDataHolderEx {
         val contentRoot: VirtualFile?
         val rootDirectory: Path
 
@@ -519,7 +521,7 @@ private class PackageImpl(
     val cargoEnabledFeatures: Set<FeatureName>,
     override val env: Map<String, String>,
     val outDirUrl: String?
-) : CargoWorkspace.Package {
+) : UserDataHolderBase(), CargoWorkspace.Package {
     override val targets = targetsData.map {
         TargetImpl(
             this,
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -49,7 +49,7 @@ class RsFileStub(
     override fun getType() = Type
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
-        private const val STUB_VERSION = 205
+        private const val STUB_VERSION = 206
 
         // Bump this number if Stub structure changes
         override fun getStubVersion(): Int = RustParserDefinition.PARSER_VERSION + STUB_VERSION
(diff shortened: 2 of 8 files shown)


Kobzol:
What about disabling the code (for name resolution etc.), but not graying it out, i.e. making an exception in the highlighter? Might be a better temporary fix than graying external test code until a more general UI is prepared.

vlad20012:
Oh, I had to clarify that UI is for workspace only. `#[cfg(test)]` will be disabled in dependencies anyway. The reason is that a user should have the ability to write tests in their project, but during writing tests they dontt need `cfg(test)` enabled in dependencies.

Kobzol:
Right. Still, I think that reading tests (even in external crates) might be quite common? So maybe an exception in the highlighter might be worth it.

vlad20012:
Maybe. But I afraid that our name resolution (both old and new) may work pretty unpredictably in cfg-disabled modules...

vlad20012:
Just pushed one more idea: we can find `#[cfg(not(test))]` conditions in a package and enable `test` if none. And disable if they are.

This means that nothing changes for packages that use `test` like `#[cfg(test)]` only, i.e. `test` will be enabled. And `test` will be disabled for packages that has `#[cfg(not(test))]` conditions.

Kobzol:
It might be a bit unexpected for the user that having `cfg(not(test))` in an external crate disables test highlighting, but it's probably a much better tradeoff than highlighting them always with potential buggy name resolution. And I guess it's not that common to cause many problems. I like this solution :)

I wonder if it's enough to do this on a per-crate level? What if there's something like
```rust
// crate_a
#[cfg(test)]
mod foo {
    // Some mocks here
}
#[cfg(not(test))]
use foobar::foo; // Import the production variant of `foo`

// crate_b (does not have cfg(not(test)))
#[cfg(test)]
mod foo {
    use crate_a::foo; // wrong name resolution?
}
```
But that's probably a niche use case.

vlad20012:
@Kobzol your example shows exactly how Cargo works, haha. `cfg(test)` is set only for the crate being `cargo test`ed. Any its dependencies are compiled without `--cfg test`.

> It might be a bit unexpected for the user that having cfg(not(test)) in an external crate disables test highlighting.

I can't decide what's better: be clear but unhandy or handy-in-most-cases but confusing. With the heuristic, it will really look like a bug that in _some_ crates there are no `#[cfg(test)]` highlighting. However, we can show some tooltip about why a cfg is disabled for a code under `cfg(test)`

vlad20012:
Ok, let's merge this heuristic for now. I updated the PR description.

bors r+

Issue #6513, opened 2020-12-15
"Remove parameter" quick fix deletes function reference


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3544-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3 Ultimate Edition (IU-203.5981.155)
* **Operating system:** macOS 10.15.7
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description
"Remove parameter" quick fix improperly deletes function references.

![remove_parameter_bug](https://user-images.githubusercontent.com/4854600/102206173-3bfc0a00-3edd-11eb-924b-e7cf7c8f8830.gif)

## Steps to reproduce
```Rust
fn foo(x: i32) {} // invoke "Remove parameter x" quick fix

fn bar<F>(f: F) where F: Fn((i32)) -> () {}

fn main() {
    bar(foo); // `foo` will be deleted after that
}
```

PR #6515, opened 2020-12-15
INSP: do not remove function references in remove parameter quick fix

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6513

changelog: RemoveParameter quick fix no longer removes some references of the modified function.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
@@ -8,9 +8,12 @@ package org.rust.lang.core.psi.ext
 import com.intellij.lang.ASTNode
 import com.intellij.openapi.util.SimpleModificationTracker
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiReference
 import com.intellij.psi.search.SearchScope
+import com.intellij.psi.search.searches.ReferencesSearch
 import com.intellij.psi.stubs.IStubElementType
 import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.util.Query
 import org.rust.ide.icons.RsIcons
 import org.rust.ide.icons.addTestMark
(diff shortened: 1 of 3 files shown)


src/test/kotlin/org/rust/ide/inspections/lints/RsLivenessInspectionTest.kt
```diff
@@ -739,6 +739,23 @@ class RsLivenessInspectionTest : RsInspectionsTestBase(RsLivenessInspection::cla
         }
     """)
 
+    // https://github.com/intellij-rust/intellij-rust/issues/6513
+    fun `test remove function argument when function is used as an argument`() = checkFixByText("Remove parameter `x`", """
+        fn foo(<warning>x/*caret*/</warning>: i32) {}
+        fn bar<F>(<warning>f</warning>: F) where F: Fn((i32)) -> () {}
```
dima74, line 745: Here refactoring would break code because `bar` expects function with one argument (probably would be nice to show some warning in such cases). Maybe it is better to use `id` here

```suggestion
        fn id<T>(t: T) { t }
```

src/main/kotlin/org/rust/lang/core/psi/ext/RsFunction.kt
```diff
@@ -185,6 +188,31 @@ val QueryAttributes.isProcMacroDef
...
+}
+
+fun RsFunction.findMethodCalls(scope: SearchScope? = null): Sequence<RsMethodCall> {
+    return createQuery(scope)
+        .asSequence()
+        .map { it.element }
+        .filterIsInstance<RsMethodCall>()
+}
+
+private fun RsElement.createQuery(scope: SearchScope? = null): Query<PsiReference> {
```
dima74, line 208: Maybe we should name it `searchReferences`

dima74 (approved):

Kobzol:
bors r=dima74

PR #6530, opened 2020-12-17
RUN: Enable Build Tool Window by default

### LINKED ISSUE #5972: The actual command of the "Build" button doesn't change according to cargo command
For example,
I have three binary cargo crates in the same directory as follows:
```text
project
| -- c1
| -- c2
| -- c3
```
And I have three commands in Run/Debug configuration
- Run c1 `run --package c1 --bin c1`
- Run c2 `run --package c2 --bin c2`
- Run c3 `run --package c3 --bin c3`

I choose `Run c1` and click on Build button, then c1 compiles. It's what is supposed to happen.
However, I choose `Run c2` and click on Build button, then c1, instead of c2, compiles. It should compile c2 instead.
So do c3.

Moreover, I cannot change the arrangement of crates in Crate Panel

mchernyavsky:
Fixes #5972

### DIFF
--- a/src/main/resources-stable/META-INF/experiments.xml
+++ b/src/main/resources-stable/META-INF/experiments.xml
@@ -1,6 +1,6 @@
 <idea-plugin>
     <extensions defaultExtensionNs="com.intellij">
-        <experimentalFeature id="org.rust.cargo.build.tool.window" percentOfUsers="0">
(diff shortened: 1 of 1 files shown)


Undin (approved):
Don't forget to write `changelog` description

bors r+

vlad20012:
Finally

PR #6526, opened 2020-12-17
PERF: fix name resolution performance in large modules

vlad20012:
Fixes #6522

TODO measure performance impact

changelog: TODO

PR #6534, opened 2020-12-19
CI: provide regression testing for stdlib

Undin:
Note, I've created a new batch in regression tests workflow because analysis is quite slow currently for `core_arch` module in `std`.
But #6526 should significantly improve the performance (from 17 to 3 minutes on my machine) of stdlib test, so it's probably worth to merge both batches into a single one after #6526

changelog: provide regression testing for stdlib packages on CI

### DIFF
--- a/.github/workflows/regressions.yml
+++ b/.github/workflows/regressions.yml
@@ -53,6 +53,9 @@ jobs:
                             repository: tokio-rs/tokio
                           - name: amethyst
                             repository: amethyst/amethyst
+                    - name: regressions-2
+                      projects:
+                          - name: stdlib
         env:
             ORG_GRADLE_PROJECT_showStandardStreams: true
             PROJECTS: ${{ toJSON(matrix.batch.projects) }}
--- a/scripts/fetch_projects.py
+++ b/scripts/fetch_projects.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+import subprocess
 from typing import List, Dict
 
 from common import git_command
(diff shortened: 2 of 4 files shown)


Undin:
I've rebased onto master and it became significantly faster with changes #6526 (@vlad20012 thanks!).
But since `regression-1` batch takes more time than any `check` even without `stdlib` project, I keep the second regression batch

src/test/kotlin/org/rustPerformanceTests/RsRealProjectTestBase.kt
```diff
@@ -92,13 +92,19 @@ abstract class RsRealProjectTestBase : RsWithToolchainTestBase() {
...
-fun VirtualFile.findDescendants(filter: (VirtualFile) -> Boolean): ArrayList<VirtualFile> {
+fun VirtualFile.findDescendants(exclude: List<String>, filter: (VirtualFile) -> Boolean): ArrayList<VirtualFile> {
     val result = ArrayList<VirtualFile>()
     VfsUtilCore.visitChildrenRecursively(this,
         object : VirtualFileVisitor<ArrayList<VirtualFile>>() {
             override fun visitFile(file: VirtualFile): Boolean {
-                if (!file.isDirectory && filter(file)) result.add(file)
-                return true
+                return if (file.isDirectory) {
+                    file.name !in exclude
```
vlad20012, line 101: Currently `exclude`  matches relative path prefix (e.g. `src/llvm`) but not a directory name on any level. Also, I don't like changing `findDescendants` to provide such kind of filtering. You can just use `filter` to match a prefix

Undin: The corresponding change was introduced to filter out `core_arch` module from stdlib. But after #6526 we don't want to do it so I just dropped the change

vlad20012 (approved):
bors r+

PR #6535, opened 2020-12-20
INSP&COMP: Improve cyclic feature dependencies handling

### LINKED ISSUE #6523: Warn about cyclic feature dependencies
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3546-203-nightly
* **Rust toolchain version:** 1.50.0-nightly (e792288df 2020-12-05) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.1 Snapshot (CL-211.3241)

## Problem description

```toml
[features]
f2 = ["f2"]
```

The example is silly but if you have long and similar feature names it can become more realistic. Cyclic feature dependency is not reported as an error, there is also a questionable completion for `f2`, which ideally should be omitted.

feature

avrong:
Closes #6523

<image width="600" src="https://user-images.githubusercontent.com/6342851/103595146-cac3db80-4f0b-11eb-8bad-79f696e6cdb4.png">

changelog: Add inspection to check whether a feature has a cyclic dependency on itself

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
+++ b/toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
@@ -9,6 +9,7 @@ import com.intellij.codeInsight.completion.CompletionParameters
 import com.intellij.codeInsight.completion.CompletionProvider
 import com.intellij.codeInsight.completion.CompletionResultSet
 import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.psi.util.parentOfType
 import com.intellij.util.ProcessingContext
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.toml.StringLiteralInsertionHandler
@@ -16,6 +17,7 @@ import org.rust.toml.findCargoPackageForCargoToml
 import org.rust.toml.getPackageTomlFile
 import org.rust.toml.resolve.allFeatures
 import org.toml.lang.psi.TomlFile
+import org.toml.lang.psi.impl.TomlKeyValueImpl
 
 /**
  *  * Consider `Cargo.toml`:
@@ -32,7 +34,10 @@ class CargoTomlFeatureDependencyCompletionProvider : CompletionProvider<Completi
     override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
         val element = parameters.position
         val tomlFile = element.containingFile as? TomlFile ?: return
+
+        val parentFeature = element.parentOfType<TomlKeyValueImpl>()?.key ?: return
         for (feature in tomlFile.allFeatures()) {
+            if (feature == parentFeature) continue
             result.addElement(lookupElementForFeature(feature))
         }
 
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
@@ -1,0 +1,51 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.inspections
+
+import com.intellij.codeInspection.LocalInspectionTool
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElementVisitor
+import org.rust.cargo.CargoConstants
+import org.rust.toml.isFeatureListHeader
+import org.toml.lang.psi.*
+import org.toml.lang.psi.ext.TomlLiteralKind
+import org.toml.lang.psi.ext.kind
+
+/**
+ *  * Consider `Cargo.toml`:
+ * ```
+ * [features]
+ * foo = ["foo"]
+ *       #^ Shows error that "foo" feature depends on itself
+ * ```
+ */
+class CargoTomlCyclicFeatureInspection : LocalInspectionTool() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor =
+        buildVisitor(holder) ?: super.buildVisitor(holder, isOnTheFly)
+
+    private fun buildVisitor(holder: ProblemsHolder): PsiElementVisitor? {
+        if (holder.file.name != CargoConstants.MANIFEST_FILE) return null
+
+        return object : TomlVisitor() {
+            override fun visitLiteral(element: TomlLiteral) {
+                val parentArray = element.parent as? TomlArray ?: return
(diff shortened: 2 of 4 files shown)


toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
```diff
@@ -32,7 +34,10 @@ class CargoTomlFeatureDependencyCompletionProvider : CompletionProvider<Completi
     override fun addCompletions(parameters: CompletionParameters, context: ProcessingContext, result: CompletionResultSet) {
         val element = parameters.position
         val tomlFile = element.containingFile as? TomlFile ?: return
+
+        val parentFeature = element.parentOfType<TomlKeyValueImpl>()?.key ?: return
         for (feature in tomlFile.allFeatures()) {
+            if (feature == parentFeature) continue
```
vlad20012, line 40: Features depend on each other like a tree, so just checking a parent is not enough. Let's build a feature graph and check cycles based on it

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+ *       #^ Shows error that "foo" feature depends on itself
+ * ```
+ */
+class CargoTomlCyclicFeatureInspection : LocalInspectionTool() {
+    override fun checkFile(file: PsiFile, manager: InspectionManager, isOnTheFly: Boolean): Array<ProblemDescriptor?>? {
+        if (file.name != CargoConstants.MANIFEST_FILE) return null
+
+        val problems = mutableListOf<ProblemDescriptor>()
+        val featuresTable = file.findDescendantOfType<TomlTable> {
+            it.header.names.size == 1 && it.header.names[0].text == "features"
```
vlad20012, line 31:
```suggestion
            it.header.isFeatureListHeader
```

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+        } ?: return emptyArray()
+
+        for (keyValue in featuresTable.entries) {
+            val parentFeatureName = keyValue.key.cleanTextOrNull ?: continue
+
+            val featureArray = keyValue.value as? TomlArray ?: continue
+            for (feature in featureArray.elements) {
+                val featureName = feature.cleanTextOrNull ?: continue
```
vlad20012, line 39:
```suggestion
                val featureName = ((feature as? TomlLiteral)?.kind as? TomlLiteralKind.String)?.value ?: continue
```

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun checkFile(file: PsiFile, manager: InspectionManager, isOnTheFly: Boolean): Array<ProblemDescriptor?>? {
+        if (file.name != CargoConstants.MANIFEST_FILE) return null
+
+        val problems = mutableListOf<ProblemDescriptor>()
+        val featuresTable = file.findDescendantOfType<TomlTable> {
+            it.header.names.size == 1 && it.header.names[0].text == "features"
+        } ?: return emptyArray()
+
+        for (keyValue in featuresTable.entries) {
+            val parentFeatureName = keyValue.key.cleanTextOrNull ?: continue
```
vlad20012, line 35:
```suggestion
            val parentFeatureName = keyValue.key.text ?: continue
```

vlad20012 (changes requested):

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *  * Consider `Cargo.toml`:
+ * ```
+ * [features]
+ * foo = ["foo"]
+ *       #^ Shows error that "foo" feature depends on itself
+ * ```
+ */
+class CargoTomlCyclicFeatureInspection : LocalInspectionTool() {
+    override fun checkFile(file: PsiFile, manager: InspectionManager, isOnTheFly: Boolean): Array<ProblemDescriptor?>? {
```
Kobzol, line 26: With the local crate index, there will probably be multiple new inspections for `Cargo.toml` files. I wonder if it would be better to create a Toml PSI visitor so we could implement inspections using the `buildVisitor` method? Assuming it has benefits versus `checkFile` (such as better performance), otherwise it probably doesn't matter.

vlad20012: Yep, this definitely affects inspection latency

toml/src/main/kotlin/org/rust/toml/completion/CargoTomlFeatureDependencyCompletionProvider.kt
ortem: IIUC Cargo features can be cyclic (https://github.com/rust-lang/cargo/pull/4473), but only optional dependencies, as well as related features, cannot be

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
Kobzol: I tried to create a visitor in https://github.com/intellij-rust/intellij-rust/pull/6578.

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
```diff
@@ -0,0 +1,58 @@
...
+        return object : TomlVisitor() {
+            override fun visitTable(element: TomlTable) {
+                if (!element.header.isFeatureListHeader) return
+
+                for (keyValue in element.entries) {
+                    val parentFeatureName = keyValue.key.text ?: continue
+
+                    val featureArray = keyValue.value as? TomlArray ?: continue
+                    for (feature in featureArray.elements) {
+                        val featureName = ((feature as? TomlLiteral)?.kind as? TomlLiteralKind.String)?.value
```
vlad20012, line 44: In IntelliJ it's always better to attach annotations directly to the `element` that is visited now. it's a bit long story... And we often don't adhere to this rule. But let's try it in this case since there is a visitor and it looks pretty easy. Let's override `visitLiteral` and then check that it is a string in TomlArray in TomlKeyValue in TomlTable with a header that is `isFeatureListHeader`.

Undin:
@avrong tests?

avrong:
@Undin I'm working on this

PR #6541, opened 2020-12-22
GRAM: parse `::` as an incomplete path

vlad20012:
As discussed in https://github.com/intellij-rust/intellij-rust/pull/6404#issuecomment-748450945. Can be considered as a continuation of #6285.

Provides better syntax error messages in the case of an incomplete path like this:
```rust
use ::;
type T = <Foo as Bar>::;
```

c.c. @msmorgan

changelog: Improve parser recovery and syntax error messages in the case of incomplete paths


vlad20012:
bors r+

The regression tests failure is OK because there is a replacement of one syntax error with another

PR #6543, opened 2020-12-22
RES: Proper resolve textual scoped macros in new resolve

dima74:
Support multiple macros with same name in same mod when using [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217)

changelog: Proper resolve textual scoped macros in new resolve

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt
@@ -33,6 +33,7 @@ import org.rust.ide.injected.isDoctestInjection
 import org.rust.lang.RsConstants
 import org.rust.lang.core.FeatureAvailability
 import org.rust.lang.core.IN_BAND_LIFETIMES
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
```diff
@@ -100,6 +103,7 @@ val RsItemsOwner.expandedItemsCached: RsCachedItems
                     }
                 }
             }
+            if (it.hasMacroIndex()) ++localMacroIndex
```
Undin, line 106:
```suggestion
            if (it.hasMacroIndex()) {
                ++localMacroIndex
            }
```

we omit block only in two cases:
- "exit" statements like `return`, `error`, `throw`, etc.
- `if expr2 else expr2` where both expressions are quite simple

vlad20012 (approved):
It's very cool! 🎉🎉🎉
bors r+

PR #6549, opened 2020-12-23
Evaluate unknown cfg options to false if no `build.rs` in a package

vlad20012:
For a long time, we did consider _unknown_ `cfg` options and unknown cargo features as enabled and disabled at the same time.
For example, both `foo` were enabled here:
```rust
#[cfg(unknown_cfg_option)]
fn foo() {}
#[cfg(not(unknown_cfg_option))]
fn foo() {}
```

What means "unknown"? Well, we had a list of known `cfg` options. It contained `debug_assertions`, `unix`, `windows` named options and `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_os`, `target_pointer_width`, `target_vendor` key-value options.

In the case of cargo features, we considered as "unknown" features that are not mentioned in `[features]` table of `Cargo.toml`.

Why? Why not just consider these `cfg` options as disabled (just like `rustc` does)?

There are two reasons:
1. We use `rustc --print cfg` command in order to retrieve actual compiler cfg options. But a stable compiler doesn't print unstable options that are used in the stdlib. 
2. Custom build scripts (`build.rs`) can produce custom `cfg` options (`#[cfg(foobar)]`), as well as features that are not mentioned in `Cargo-toml`: `#[cfg(feature = "foobar")]`.

This PR fights the first problem by unsing `RUSTC_BOOTSTRAP=1` environment variable when invoking `rustc --print cfg`. This magic variable makes the compiler think that it is nightly and so print unstable `cfg` options.

The second problem should be eventually solved by honest buildscript evaluation (see #6104), but now we can just apply "honest" behavior for packages without `build.rs` and leave old behavior for packages with build.rs. Also, the "honest" behavior is applied to all packages if experimental buildscript evaluation is enabled.

So finally if a package doesn't have a custom build script (`build.rs`), (almost) all `cfg` conditions are evaluated correctly. But there are two exception:
1. `test` cfg option is still considered enabled and disabled at the same time (`#[cfg(test)]` and `#[cfg(not(test))]` both enabled). #6507 fixes it for external dependencies, but not for workspace packages.
2. An unstable [`panic`](https://doc.rust-lang.org/nightly/unstable-book/language-features/cfg-panic.html) cfg option (`#[cfg(panic = "unwind")]`) is still considered enabled and disabled at the same time, just because I haven't found a way to determine panic strategy for a crate.

changelog: Honestly Evaluate unknown `cfg` options to false if there is no custom build script (`build.rs`) in a package

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/crate/Crate.kt
+++ b/src/main/kotlin/org/rust/lang/core/crate/Crate.kt
@@ -37,6 +37,13 @@ interface Crate {
     val cfgOptions: CfgOptions
     val features: Map<String, FeatureState>
 
+    /**
+     * `true` if there isn't a custom build script (`build.rs`) in the package or if the build script is
+     * successfully evaluated (hence [cfgOptions] is filled). The value affects `#[cfg()]` and `#[cfg_attr()]`
+     * attributes evaluation.
+     */
+    val evaluateUnknownCfgToFalse: Boolean
+
     /** A map of compile-time environment variables, needed for `env!("FOO")` macros expansion */
     val env: Map<String, String>
 
--- a/src/main/kotlin/org/rust/lang/core/crate/impl/CargoBasedCrate.kt
+++ b/src/main/kotlin/org/rust/lang/core/crate/impl/CargoBasedCrate.kt
@@ -47,7 +47,12 @@ class CargoBasedCrate(
     override val cargoWorkspace: CargoWorkspace get() = cargoTarget.pkg.workspace
     override val kind: CargoWorkspace.TargetKind get() = cargoTarget.kind
 
-    override val cfgOptions: CfgOptions get() = cargoTarget.pkg.cfgOptions
+    override val cfgOptions: CfgOptions get() = cargoTarget.cfgOptions
+
+    override val evaluateUnknownCfgToFalse: Boolean
+        get() = cargoTarget.pkg.cfgOptions != null   // `true` if there is `build.rs` and it is evaluated
+            || !cargoTarget.pkg.hasCustomBuildScript // `true` if there isn't `build.rs`
+
     override val env: Map<String, String> get() = cargoTarget.pkg.env
     override val outDir: VirtualFile? get() = cargoTarget.pkg.outDir
 
--- a/src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
+++ b/src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
@@ -6,7 +6,6 @@
 package org.rust.lang.utils.evaluation
 
 import com.intellij.openapiext.Testmark
-import com.intellij.openapiext.isUnitTestMode
 import org.rust.cargo.CfgOptions
 import org.rust.cargo.project.workspace.FeatureState
 import org.rust.cargo.project.workspace.PackageOrigin
@@ -60,14 +59,14 @@ operator fun ThreeValuedLogic.not(): ThreeValuedLogic = when (this) {
     Unknown -> Unknown
 }
 
-// Note, [options] and [packageOptions] can contain the same option values,
+// Note, [options] can contain conflicting option values,
 // i.e. it's possible to have both `unix` and `windows` values at the same time.
 // See https://doc.rust-lang.org/reference/conditional-compilation.html#set-configuration-options
 class CfgEvaluator(
-    val options: CfgOptions,
-    val packageOptions: CfgOptions,
-    val features: Map<String, FeatureState>,
-    val origin: PackageOrigin
+    private val options: CfgOptions,
+    private val features: Map<String, FeatureState>,
+    private val origin: PackageOrigin,
+    private val evaluateUnknownCfgToFalse: Boolean
 ) {
     fun evaluate(cfgAttributes: Sequence<RsMetaItem>): ThreeValuedLogic {
         return evaluate(CfgPredicate.fromCfgAttributes(cfgAttributes))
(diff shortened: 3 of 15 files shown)


src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
```diff
@@ -87,24 +86,29 @@ class CfgEvaluator(
...
-        name == CfgOptions.TEST && isUnitTestMode -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
-        else -> Unknown
+        // https://doc.rust-lang.org/nightly/unstable-book/language-features/cfg-panic.html
+        name == "cfg_panic" -> Unknown
+
+        name == "test" && origin != PackageOrigin.STDLIB -> Unknown
+        strict -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
+        else -> when {
+            name in SUPPORTED_NAME_OPTIONS -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
+            (name == "test" || name == "bootstrap" || name == "doc") && origin == PackageOrigin.STDLIB -> False
```
ortem, line 100: Note that `test` and `doc` options are in `SUPPORTED_NAME_OPTIONS` so they are processed by the first match arm. Consider something like
```Kotlin
when {
    name in STDLIB_FALSE_NAME_OPTIONS && origin == PackageOrigin.STDLIB -> False
    name in SUPPORTED_NAME_OPTIONS -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
    else -> Unknown
}
```

src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
```diff
@@ -87,24 +86,29 @@ class CfgEvaluator(
...
-        // TODO: convert whitelist to blacklist and merge options with packageOption
-        name in SUPPORTED_NAME_OPTIONS -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
-        (name == "test" || name == "bootstrap" || name == "doc") && origin == PackageOrigin.STDLIB -> False
-        name == CfgOptions.TEST && isUnitTestMode -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
-        else -> Unknown
+        // https://doc.rust-lang.org/nightly/unstable-book/language-features/cfg-panic.html
+        name == "cfg_panic" -> Unknown
+
+        name == "test" && origin != PackageOrigin.STDLIB -> Unknown
+        strict -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
```
ortem, line 97: Note that currently the logic is a bit duplicated, e.g. `ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))` here and below in case of `name in SUPPORTED_NAME_OPTIONS`. I suppose strict evaluation can be performed in an easier way. AFAIU all we want to do is to replace `Unknown` with `False` in case of strict evaluation. All the evaluations like `options.isNameEnabled` or `options.isNameValueEnabled` are cheap, so we can perform them at first (like we did before), and then just replace `Unknown` with `False`.

What do you think about introducing such a method
```Kotlin
enum class ThreeValuedLogic {
    True, False, Unknown;

    // Not sure about the name, maybe `confirm` is better or even something else
    fun refine(strict: Boolean): ThreeValuedLogic = when (this) {
        True -> True
        False -> False
        Unknown -> if (strict) False else Unknown
    }
//...
}
```

and using it this way?
```Kotlin
private fun evaluateName(name: String): ThreeValuedLogic = when {
    //...
    name in SUPPORTED_NAME_OPTIONS -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
    else -> Unknown
}.refine(strict)
```

src/test/kotlin/org/rust/lang/core/resolve/RsCfgAttrResolveTest.kt
```diff
@@ -752,4 +752,30 @@ class RsCfgAttrResolveTest : RsResolveTestBase() {
...
+        fn foo() {}
+        #[cfg(not(feature = "unknown_cargo_feature"))]
+        fn foo() {}
+         //X
+
+        fn main() {
+            foo();
+          //^
+        }
+    """)
```
ortem, line 780: Let's add more tests of using unknown options and features in complex predicates.
Also, some tests for stdlib features (like `bootstrap` and `test`) would be very helpful as well. However, I'm not sure if such tests are easy to implement, what do you think?

ortem (changes requested):

src/main/kotlin/org/rust/lang/core/crate/Crate.kt
```diff
@@ -37,6 +37,8 @@ interface Crate {
     val cfgOptions: CfgOptions
     val features: Map<String, FeatureState>
 
+    val isBuildscriptEvaluated: Boolean
```
Undin, line 40: Let's rename it somehow because the current name is misleading: I suppose it should be false when we don't evaluate build scripts at all but actually, it's true when crate's package doesn't have build script at all.
Something like `canHaveUnknownCfgOptions`

src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
vlad20012: I think this variant is much less explicit... and it saves only one line of code: `strict -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))`. I don't this wirth it.

src/test/kotlin/org/rust/lang/core/resolve/RsCfgAttrResolveTest.kt
vlad20012: Oh, I don't know how to check evaluation in the stdlib...

src/main/kotlin/org/rust/lang/utils/evaluation/CfgEvaluator.kt
```diff
@@ -95,24 +94,32 @@ class CfgEvaluator(
...
-        (name == "test" || name == "bootstrap" || name == "doc") && origin == PackageOrigin.STDLIB -> False
-        name == CfgOptions.TEST && isUnitTestMode -> ThreeValuedLogic.fromBoolean(options.isNameEnabled(name))
-        else -> Unknown
+        // https://doc.rust-lang.org/nightly/unstable-book/language-features/cfg-panic.html
+        name == "cfg_panic" -> Unknown
+
+        name == "test" && origin != PackageOrigin.STDLIB -> Unknown
+
+        // Standard macros in old stdlib (1.32.0)
+        name == "rustdoc" && origin == PackageOrigin.STDLIB -> Unknown
```
ortem, line 107: Maybe add `// BACKCOMPAT: rust 1.32`?

src/main/kotlin/org/rust/lang/core/crate/impl/CargoBasedCrate.kt
```diff
@@ -47,7 +47,9 @@ class CargoBasedCrate(
     override val cargoWorkspace: CargoWorkspace get() = cargoTarget.pkg.workspace
     override val kind: CargoWorkspace.TargetKind get() = cargoTarget.kind
 
-    override val cfgOptions: CfgOptions get() = cargoTarget.pkg.cfgOptions
+    override val cfgOptions: CfgOptions get() = cargoTarget.cfgOptions
+    override val evaluateUnknownCfgToFalse: Boolean
+        get() = cargoTarget.pkg.cfgOptions != null || cargoTarget.pkg.customBuildTarget == null
```
ortem, line 52: This condition is not really obvious. Probably it should be separated into three cases, e.g. “no build.rs”, “evaluated build.rs”, and “ignored build.rs"

ortem (approved):
Generally LGTM

vlad20012:
bors r=ortem

vlad20012:
bors retry

PR #6558, opened 2020-12-25
RES: Fix propagation expanded macro def to grandparent mod

dima74:
Fix `StackOverflowError` in case when both parent and grandparent mod of expanded macro def has `macro_use` attribute (Introduced in #6543)


vlad20012 (approved):
bors r+

PR #6563, opened 2020-12-27
ACT: introduce `Share in Playground` action

Undin:
These changes:
* Introduce `Share in Playground` action to share your code in https://play.rust-lang.org/. The action correctly handles selected text in editor (i.e. share only selected text), current edition and toolchain channel.
* Add `Rust` action group to have a single item in context menus for util Rust actions. `Reformat File with Rustfmt`, `Reformat Cargo Project with Rustfmt`, `Rust REPL`, `Share in Playground` are placed into this group as well as `Show expanded Macro` actions. Also, the group is added to `Tools` menu.
* Provide `MockServerFixture` to help setting up mock web server in test to check network interactions

https://user-images.githubusercontent.com/2539310/103181788-265fea80-48b6-11eb-87f7-2ba813cfb596.mov

changelog: Introduce `Share in Playground` action to share your code in https://play.rust-lang.org/. You can invoke it via `Tools | Rust | Share in Playground` or via context menu


ortem:
Probably it's worth showing a confirmation modal window before sending the code to the server. I'm afraid someone might _accidentally_ invoke this action, and some potentially _proprietary_ code will immediately go to the server. In the modal window, we can also provide some information like "entire file `filename.rs` / selected code fragment will be sent to the server"

Undin:
@ortem done

build.gradle.kts
```diff
@@ -286,6 +286,7 @@ project(":") {
             exclude(module = "kotlin-stdlib")
         }
         testImplementation(project(":common", "testOutput"))
+        testImplementation("com.squareup.okhttp3:mockwebserver:4.9.0")
```
vlad20012, line 289: Wow, it's cool

vlad20012 (approved):

src/test/kotlin/org/rust/MockServerFixture.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val mockWebServer = MockWebServer().apply {
+        dispatcher = object : Dispatcher() {
+            override fun dispatch(request: RecordedRequest): MockResponse {
+                Assert.assertEquals(USER_AGENT, request.getHeader("User-Agent"))
+                return handler?.invoke(request) ?: MockResponse().setResponseCode(404)
+            }
+        }
+    }
+
+    val baseUrl: String get() = mockWebServer.url("/").toString()
```
ortem, line 31: [minor] not sure if it really can change, `get()` might be redundant

src/main/kotlin/org/rust/ide/actions/ShareInPlaygroundAction.kt
```diff
@@ -0,0 +1,161 @@
...
+            }
+        }.queue()
+    }
+
+    private fun confirmShare(file: RsFile, hasSelection: Boolean): Boolean {
+        val showConfirmation = PropertiesComponent.getInstance().getBoolean(SHOW_SHARE_IN_PLAYGROUND_CONFIRMATION, true)
+        if (!showConfirmation) {
+            return true
+        }
+        val doNotAskOption: DialogWrapper.DoNotAskOption.Adapter = object : DialogWrapper.DoNotAskOption.Adapter() {
```
ortem, line 110:
```suggestion
        val doNotAskOption = object : DialogWrapper.DoNotAskOption.Adapter() {
```

src/main/kotlin/org/rust/ide/actions/ShareInPlaygroundAction.kt
```diff
@@ -0,0 +1,161 @@
...
+
+            @Volatile
+            private var gistId: String? = null
+
+            override fun shouldStartInBackground(): Boolean = true
+            override fun run(indicator: ProgressIndicator) {
+                val json = Gson().toJson(PlaygroundCode(text))
+                val response = HttpRequests.post("$playgroundHost/meta/gist/", "application/json")
+                    .userAgent(USER_AGENT)
+                    .connect {
```
ortem, line 70: Do you intentionally ignore `IOException` here and use `parseJsonObject` instead of `tryParseJsonObject` below?

ortem (approved):
LGTM except for several minor comments

src/main/kotlin/org/rust/ide/actions/ShareInPlaygroundAction.kt
Undin: Yep. In case of exception, platform call `onThrowable` and we log exception and show error notification to user

src/test/kotlin/org/rust/MockServerFixture.kt
Undin: If we drop `get` here, then mock web server will be started on each creation of `MockServerFixture`. It doesn't matter with current usages, but we don't do it in general

Undin:
bors r=ortem

src/test/kotlin/org/rust/MockServerFixture.kt
ortem: Probably `lazy` delegate would be a bit more clear in this case

KseniaStar:
Changes reflected in docs: https://plugins.jetbrains.com/plugin/8182-rust/docs/rust-quick-start.html#share-in-playground

Issue #6367, opened 2020-11-08
Adding support for custom syntax introduced by proc macros


hamza1311:
Libraries/frameworks use proc macros to introduce custom syntax for their specialized use case. Take `html!` macro from [Yew](https://github.com/yewstack/yew/) for example (multiple examples of syntax introduced by it can be found on [their website](https://yew.rs/docs/en/concepts/html)). There is no input from IDE for *anything* written inside the macro and (I assume) it is impractical for intellij-rust to add support for this. Is there any way to add support for such macros on top of intellij-rust (such as with a plugin)? If not, can this feature be introduced? Currently, when working with frameworks like Yew and not having support for such a crucial macro's syntax is annoying to deal with and hurts the developer experience.

hamza1311:
Ref: https://github.com/rust-lang/rust-analyzer/issues/7402

vlad20012:
With #9729 landed there is *some* highlighting & completion in proc macros with very custom syntax like `yew`:
![image](https://user-images.githubusercontent.com/3221931/201992389-02923f7a-48d2-46de-abfd-f709a2b71a2e.png)

So I think this issue should be clarified now. What particularly meant by "support" of custom syntax? For example, I'm also working on completion in proc macros like `yew` in #9711 (I'll write the success story about completion in `yew` a bit later). Is it in the scope of this issue?

hamza1311:
@vlad20012 The scope of this issue is really to allow writing the kind of plugin that you mentioned in https://github.com/yewstack/yew/pull/2972#issuecomment-1324830781

At the moment, we have to no way to hook into IntelliJ Rust and tell it about the specifics of a proc macro

vlad20012:
If we talk about IntelliJ plugins, then, to be honest, it was always possible to write a plugin for IntelliJ Rust that does things like that. And we need almost nothing from IntelliJ-Rust side (well, maybe just a bit) because IntelliJ platform itself is very plugin-centric, so a one can adjust anything with a plugin. For example, there was a plugin for LALRPOP: https://github.com/lalrpop/lalrpop

hamza1311:
I tried to look into it a while back but couldn't find any documentation on where to get started. LALRPOP plugin seems to be something else though. It adds language support for .lalrapop files, not a specific macro.

I imagine the part IntelliJ Rust plays is tell the plugin how to parse the Rust file and where to look for the macro. Not sure how else the plugin will know what input it is working with. 

[...]

PR #6564, opened 2020-12-28
Expand function-like (bang) procedural macros

vlad20012:
Support function-like procedural macros like `foo!()`. The support includes most of the features that work with declarative macros (aka macro by example) - name resolution, highlighting, code completion, etc.

## User guide

This is an experimental feature, hence bugs are expected. The support is limited to function-like procedural macros (like `foo!()`).
If you want to try it, you should enable `org.rust.macros.proc` and `org.rust.cargo.evaluate.build.script` experimental features (`Ctrl` + `Shift` + `A` -> type "experimental features") and then refresh the cargo project (`Ctrl` + `Shift` + `A` -> type "Refresh Cargo Projects"). If a procedural macro is a part of your project workspace, you should refresh the cargo project (manually) every time you change you proc macro code. 

## Historical notes and implementation details

Proc macro support possibilities were initially researched during JetBrains
internship in 2018. At that time we finalized such design:

```
[IDE] ---------> [Proc macro expander (server)] ----> [proc_macro.so]
       ^ some protocol (e.g. JSON based)         ^ dynamic linkage
```

Basic idea is that the Rust compiler produces dynamic libraries (.dll, .so, .dylib) for proc macro crates.
Then we can link such dynamic libraries and request them to expand a macro that we need. In theory.
But in practice, in 2018 these libraries used unstable Rust ABI and a lot of internal compilers
data structures that made this approach rather impossible.

But some time later some groundwork was done in the Rust compiler, and proc macro libraries started exporting
stable C ABI. Also, there was a compact abstraction layer in `libproc_macro`, so using
`#![feature(proc_macro_internals)]` it became possible to link proc macro libraries and call functions from them.

This approach was implemented in this repo: https://github.com/fedochet/rust-proc-macro-expander, but
it hasn't been integrated into intellij-rust because our macro expansion infrastructure was not ready
for proc macros and because we prioritized other directions. 

Then, rust-analyzer developers used this repo as a base for their proc macro expander! They improved it in several ways and
[published it on crate.io](https://crates.io/crates/ra_ap_proc_macro_srv). Now we can just use `ra_ap_proc_macro_srv` crate from crates.io.

So, I've pushed a simple rust program that embeds `ra_ap_proc_macro_srv` in #6658, then @Undin set up its builds on CI in #6713. Then, `proc-macro-expander` was renamed to `intellij-rust-native-helper` in #6735 because the binary started to be responsible for other stuff, not only proc macro expansion.

## TODO
Add tests for completion, `extend selection`, `type info`


vlad20012:
It's ready for review. I've updated the PR description.

In this PR we only support function-like procedural macros like `foo!()`. The support includes most of the features that work with declarative macros (aka macro by example) - name resolution, highlighting, code completion, etc. If you want to try it, you should enable `org.rust.macros.proc` and `org.rust.cargo.evaluate.build.script` experimental features. Other kinds of procedural macros (custom derive and attribute macros) will be supported in other PRs

hamza1311:
> The support includes most of the features that work with declarative macros (aka macro by example) - name resolution, highlighting, code completion, etc

How does highlighting and code completion work with macros that don't take syntactically correct rust as their input? I don't see anything related to such macros from a quick look at test cases. Are they not supported yet? I'm assuming the support is only for their output, correct? If so, is there any plans to allow adding support custom syntax provided by proc macros (like I described in #6367)?

vlad20012:
@hamza1311, Completion for procedural macros works similar to completion for declarative macros - we expand the macro (if we can) and then invoke the completion on the macro expansion (with caret position mapping). This is expected to work worse for procedural macros because they are (usually) not resilient to incorrect rust syntax (this should be solved on the procedural macro side, i.e. in `syn` crate)
There are ideas on how we could support very custom syntax (again, this should be done on the proc macro side), see https://github.com/rust-analyzer/rust-analyzer/issues/7402#issuecomment-770196608

I can't test `yew` for now because `html` macro is defined using `proc_macro_hack` that is an attribute macro which are not covered by this PR, but I'd expect a lot of code completion working there, at least in lambdas (because they are just arbitrary rust code) and in custom component names (because they are just a rust struct names)

#6367 is a valid issue anyway

hamza1311:
> we expand the macro (if we can) and then invoke the completion on the macro expansion (with caret position mapping).

I'm not sure how much of this is done on IntelliJ's side so correct me if I'm wrong but would it be possible to provide a way to set the environment where the expanded code is invoked? For example, the [`sqlx::query!`](https://docs.rs/sqlx/0.5.1/sqlx/macro.query.html) family of macros make calls to database so I think ability to force offline mode (where it reads from a pre-generated file instead of calling the database) in those macros is important.

> I can't test yew for now because html macro is defined using proc_macro_hack that is an attribute macro which are not covered by this PR, but I'd expect a lot of code completion working there in the future, at least in lambdas (because they are just arbitrary rust code) and in custom component names (because they are just a rust struct names)

`Yew` doesn't use `proc_macro_hack` anymore. That change isn't released so if you wish to test that macro, you can depend from master. You can find the macro on master [here](https://github.com/yewstack/yew/blob/master/packages/yew-macro/src/lib.rs#L112-L116).

vlad20012:
> You can find the macro on master here

Oh, thanks! Yep, the completion works as expected:

https://user-images.githubusercontent.com/3221931/107040448-49908a80-67d0-11eb-8137-a214d44547e6.mp4

> I'm not sure how much of this is done on IntelliJ's side so correct me if I'm wrong but would it be possible to provide a way to set the environment where the expanded code is invoked? For example, the sqlx::query! family of macros make calls to database so I think ability to force offline mode (where it reads from a pre-generated file instead of calling the database) in those macros is important.

We could tell a proc macro that it is launched from IDE (In this PR I set env var `INTELLIJ_RUST=true`), but a proc macro should take it into account.

hamza1311:
> Yep, the completion works as expected:

Incredible! Looking forward to the release.

src/main/kotlin/org/rust/lang/core/macros/decl/DeclMacroExpander.kt
```diff
@@ -0,0 +1,518 @@
+/*
```
dima74, line 1: Maybe split files move into separate commit

dima74 (approved):
🚀🚀🚀

src/main/kotlin/org/rust/lang/core/macros/RangeMap.kt
```diff
@@ -122,3 +123,20 @@ private fun MappedTextRange.dstIntersection(range: TextRange): MappedTextRange?
         null
     }
 }
+
+fun MutableList<MappedTextRange>.mergeAdd(range: MappedTextRange) {
```
Undin, line 127: Actually, I don't fully understand why we need this. Could you explain, please (probably, it's worth adding a doc comment with an example)?

src/main/kotlin/org/rust/ide/wordSelection/RsMacroCallSelectionHandler.kt
```diff
@@ -54,15 +54,19 @@ class RsMacroCallSelectionHandler : ExtendWordSelectionHandlerBase() {
         val expansion = macroCall.expansion ?: return null // impossible?
         val expansionText = expansion.file.text
 
+        if (offsetInExpansion > expansionText.length) return null
```
Undin, line 57: Is it really possible?

Undin (changes requested):
In general, looks awesome! 🚀 

It would be great to write more tests for existing features like completion, highlighting, extend  selection, etc. but I don't insist to do it in this PR - it's already quite huge

Also, some tests fail with these changes

src/main/kotlin/org/rust/lang/core/macros/proc/Request.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.rust.lang.core.macros.proc
+
+import com.google.gson.*
+import com.google.gson.annotations.SerializedName
+import org.rust.lang.core.macros.tt.TokenTree
+import java.lang.reflect.Type
+
+sealed class Request {
```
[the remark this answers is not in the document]
vlad20012: There is a `ListMacros` request in the protocol. I'm not going to use it, but I'd want to specify that there are other request kinds

src/main/kotlin/org/rust/lang/core/macros/RangeMap.kt
vlad20012: It's an optimization to reduce the list size. The method replaces the last list element with a united range if they intersect

dima74 (approved):

src/main/kotlin/org/rust/lang/core/macros/RsMacroDataWithHash.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsMacro
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve2.DeclMacroDefInfo
+import org.rust.lang.core.resolve2.MacroDefInfo
+import org.rust.lang.core.resolve2.ProcMacroDefInfo
+import org.rust.stdext.HashCode
+
+class RsMacroDataWithHash<T: RsMacroData>(
```
Undin, line 16:
```suggestion
class RsMacroDataWithHash<T : RsMacroData>(
```

src/main/kotlin/org/rust/lang/core/macros/RsMacroDataWithHash.kt
```diff
@@ -0,0 +1,54 @@
...
+    val data: T,
+    val bodyHash: HashCode?
+) {
+    fun mixHash(call: RsMacroCallDataWithHash): HashCode? {
+        val callHash = when (data as RsMacroData) {
```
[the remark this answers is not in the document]
Undin: Actually, it's a kotlin plugin false positive. Let's just add a suppression here

Undin (commented):
* The current implementation cannot expand https://github.com/yewstack/yew/blob/f93fa3c8b2106dca4eb768ae005ef530310b7721/examples/boids/src/main.rs#L118-L166
* expand selection can't select `Msg::RestartSimulation` in https://github.com/yewstack/yew/blob/f93fa3c8b2106dca4eb768ae005ef530310b7721/examples/boids/src/main.rs#L92
* Type info doesn't work any expr in https://github.com/yewstack/yew/blob/f93fa3c8b2106dca4eb768ae005ef530310b7721/examples/boids/src/main.rs#L92

src/main/kotlin/org/rust/lang/core/macros/RsMacroCallDataWithHash.kt
```diff
@@ -0,0 +1,24 @@
...
+    val bodyHash: HashCode?
+) {
+    fun hashWithEnv(): HashCode? {
+        val bodyHash = bodyHash ?: return null
+        val env = data.packageEnv
+            .entries
+            .toList()
+            .sortedBy { it.key }
+            .joinToString(separator = ";") { it.key + "=" + it.value }
+        val envHash = HashCode.compute(env)
```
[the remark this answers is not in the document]
vlad20012: I don't think it's ok, but let's please fix it later :)

vlad20012:
> The current implementation cannot expand https://github.com/yewstack/yew/blob/f93fa3c8b2106dca4eb768ae005ef530310b7721/examples/boids/src/main.rs#L118-L166

Fixed. It was a bug in our TokenTree parser.

> expand selection can't select Msg::RestartSimulation in https://github.com/yewstack/yew/blob/f93fa3c8b2106dca4eb768ae005ef530310b7721/examples/boids/src/main.rs#L92
Type info doesn't work for any expr in https://github.com/yewstack/yew/blob/f93fa3c8b2106dca4eb768ae005ef530310b7721/examples/boids/src/main.rs#L92

I'll fix it in another PR. The same issues can be also applied to (some specific) declarative macros, I guess. I'll tweak Extend Selection / Type Info later

src/main/kotlin/org/rust/lang/core/macros/MacroExpansionError.kt
```diff
@@ -8,44 +8,88 @@ package org.rust.lang.core.macros
...
-    class Nesting(macroCallBody: PsiBuilder, val variableName: String): MacroMatchingError(macroCallBody) {
+
+    class Nesting(macroCallBody: PsiBuilder, val variableName: String) : MacroMatchingError(macroCallBody) {
         override fun toString(): String = "Nesting($variableName)"
     }
 }
+
+sealed class ProcMacroExpansionError : MacroExpansionError() {
+    data class ServerSideError(val message: String) : ProcMacroExpansionError() {
+        override fun toString(): String = "ProcMacroExpansionError.Expansion(message = \"$message\")"
```
Undin, line 69:
```suggestion
        override fun toString(): String = "ProcMacroExpansionError.ServerSideError(message = \"$message\")"
```

Undin: nit: Do you actually need `ProcMacroExpansionError.` prefix? If not, you can use `javaClass.simpleName` not to forget updating `toString`. If you need it, probably we can use something like `.javaClass.canonicalName.removePrefix("${javaClass.packageName}.")`.

Note, in fact you use `simpleName` for `DeclMacroExpansionError` errors.
Moreover, you can provide basic implementation of `toString` in this way and avoid `override fun toString(): String = "ProcMacroExpansionError.Timeout"` and friends

src/test/kotlin/org/rust/lang/core/macros/proc/RsProcMacroExpansionTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        }
+        val pkg = project.cargoProjects.singleWorkspace().packages
+            .find { it.name == "my_proc_macro" }!!
+        val lib = pkg.procMacroArtifact?.path?.toString()
+            ?: error("Procedural macro artifact is not found. This most likely means a compilation failure")
+        val server = ProcMacroServerPool.tryCreate(testRootDisposable)
+            ?: error("native-helper is not available")
+        val expander = ProcMacroExpander(project, server)
+
+        expander.apply {
```
Undin, line 120: nit:
```suggestion
        with(expander) {
```
looks `with` is a little bit more suitable here

Undin (approved):
LGTM except minor comments

vlad20012:
bors r=undin

PR #6568, opened 2020-12-28
RES: Fix restricted visibility in new resolve

dima74:
Fix handling restricted visibility with path from crate root in [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217). E.g.:
```rust
mod inner1 {
    mod inner2 {
        mod inner3 {
            pub(in crate::inner1) fn foo1() {}
            pub(in crate::inner1::inner2) fn foo2() {}
        }
    }
}
```

Previously items with such visibility were considered as `pub(crate)`

changelog: Fix handling items with restricted visibility with absolute path in new resolve

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve2/ModCollector.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve2/ModCollector.kt
@@ -320,7 +320,9 @@ private class ModCollector(
             VisibilityLight.Public -> Visibility.Public
             VisibilityLight.RestrictedCrate -> crateRoot.visibilityInSelf
             VisibilityLight.Private -> modData.visibilityInSelf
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/resolve2/ModCollector.kt
```diff
@@ -373,25 +375,18 @@ fun CrateDefMap.importExternCrateMacros(externCrateName: String) {
...
+        parents.getOrNull(path.size)
+            ?.takeIf { path.contentEquals(it.path.segments) }
+            ?.visibilityInSelf
     }
-    val pathTarget = path
-        .fold(initialModData) { modData, segment ->
-            val nextModData = when (segment) {
-                "self" -> modData
-                "super" -> modData.parent
-                else -> modData.childModules[segment]
```
dima74, line 390: Bug was here - `modData.childModules` don't contain any of parent modules here

PR #6575, opened 2020-12-29
Adjust multiline functions code style

dima74:
Some time ago intention "Put parameters on separate lines" converted this function:
```kt
fun function1(parameter1: Int, parameter2: Int, parameter3: Int) {}
```
into this:
```kt
fun function1(
    parameter1: Int,
    parameter2: Int,
    parameter3: Int
) {}
```

Recently this intention started to instead convert function to
```kt
fun function1(parameter1: Int,
              parameter2: Int,
              parameter3: Int) {}
```

Since all our code use first variant, I guess it makes sense to adjust code style settings to continue using it (for consistency purpose)


vlad20012:
Indeed. Thanks!
bors r+

Issue #6136, opened 2020-09-19
xml-rs macros not expanded properly


AndreKR:
## Environment

* **IntelliJ Rust plugin version:** 0.3.132.3360-202-nightly
* **Rust toolchain version:** nightly-x86_64-pc-windows-gnu
* **IDE name and version:** CLion 2020.2.1
* **Operating system:** Windows 7 Pro

## Problem description
The macro-generated methods of `xml-rs`'s `EmitterConfig` are not autocompleted/recognized.

## Steps to reproduce
```yaml
[dependencies]
xml-rs = "*"
```

```rust
use xml::EmitterConfig;

fn main() {
	let mut v = Vec::new();
	let mut xmlwriter = EmitterConfig::new().perform_indent(true).create_writer(&mut v);
}
```

You see that there is a method `perform_indent()`. This method is created by a [macro](https://github.com/netvl/xml-rs/blob/master/src/macros.rs) called [here](https://github.com/netvl/xml-rs/blob/master/src/writer/config.rs#L147-L157).

## Expected result

`perform_indent()` should be autocompleted.

## Actual result:

Only the field is autocompleted:

![image](https://user-images.githubusercontent.com/1188538/93663448-6c601380-fa68-11ea-991d-62906b66eb32.png)

Also:

![image](https://user-images.githubusercontent.com/1188538/93663521-ebede280-fa68-11ea-852f-a13301707161.png)

(But maybe that's expected because the methods don't really exist.)

vlad20012:
It's probably because of [`#![macro_use]`](https://github.com/netvl/xml-rs/blob/master/src/macros.rs#L1) inside the module file instead of typical `#[macro_use]` on the module declaration. 

I guess this case will be supported with `name resolution 2.0`

dima74:
Duplicate of #6486

Issue #6486, opened 2020-12-05
Plugin does not understand #![macro_use]


KarelPeeters:
* **IntelliJ Rust plugin version:** 0.3.136.3514-203
* **IDE name and version:** CLion 2020.3, Build #CL-203.5981.166

The plugin is not able to resolve macros that should be visible because of #![macro_use] instead of the more common #[macro_use].

To reproduce the issue, create a new empty binary project, write in `main.rs`:

```rust
mod module;

my_macro!();

fn main() {}
```

and in a second file `module.rs`:

```rust
#![macro_use]

macro_rules! my_macro { () => {} }
```

This code compiles fine with rustc. In the IDE, executing the "go to declaration" action on the macro call in main doesn't work, showing that the plugin didn't resolve the macro. If the macro is changed to actually declare items those items are also not resolved. The issue goes away when `#![macro_use]` in `module.rs` is replaced by `#[macro_use]` on the module declaration in `main.rs`.

I don't fully understand what `#![macro_use]` even means, but this came up in an existing project I'm working on and I reduced it to this minimal example.

dima74:
Fixed in [latest nightly](https://intellij-rust.github.io/install.html#nightly-builds) with [new name resolution](https://github.com/intellij-rust/intellij-rust/issues/6217) engine [enabled](https://github.com/intellij-rust/intellij-rust/pull/6378)

PR #6576, opened 2020-12-29
RES: Support file level macro_use attribute in new resolve

dima74:
Fixes  #6136 and #6486 with [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217)

changelog: Support `#![macro_use]` attribute at file level when using new name resolution engine


dima74:
bors retry

PR #6578, opened 2020-12-29
TOML: add TomlVisitor

Kobzol:
This PR adds a simple visitor for TOML elements. I think that it is useful enough to warrant a PR even without a specific usage.
It can be used e.g. in https://github.com/intellij-rust/intellij-rust/pull/6457, https://github.com/intellij-rust/intellij-rust/pull/3787, https://github.com/intellij-rust/intellij-rust/pull/6535 and if it's merged, I suppose that it will be used in many future Cargo.toml inspections, now that the local index is finally working.


Undin (approved):
Thanks!
bors r+

Issue #6577, opened 2020-12-29
ACT  | CLion | Weird `make pub` suggestion at the end of Code blocks


DzenanJupic:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3572-203
* **Rust toolchain version:** 1.51.0-nightly (0b644e419 2020-12-26) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2020.3 (CL-203.5981.166)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description
When the cursor is at the end of a `fn`, `impl`, or `trait` block, there are two new context actions `make pub` and `make pub(crate)`:
![image](https://user-images.githubusercontent.com/56133904/103304373-c037ac80-4a08-11eb-8b6c-d9d929f49f91.png)

Which, when selected, append a `pub` or `pub(crate)` to the end of the code block:
![image](https://user-images.githubusercontent.com/56133904/103304497-1278cd80-4a09-11eb-8820-bd41447af726.png)

I think the intended behaviour is to make the function public, like when you place the cursor at the beginning of the function:
![image](https://user-images.githubusercontent.com/56133904/103304645-77342800-4a09-11eb-867b-7f170929ce13.png)

Even though in this case, this also makes no sense, since `deserialize` is a trait function and therefore cannot have a visibility modifier.

PR #6580, opened 2020-12-29
INT: do not offer ChangeVisibilityIntention in trait impl blocks

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6577

changelog: Fix `ChangeVisibilityIntention` so that it is not offered in trait impl blocks.


vlad20012:
:+1: Thanks!
bors r+

PR #6581, opened 2020-12-29
IDE: add support for navbar breadcrumbs

Kobzol:
While working on something else, I noticed that IDEA has these cool new breadcrumbs directly inside the navigation bar. So I looked at the Kotlin to find inspiration and implemented it also for the Rust plugin, by reusing the implementation of the breadcrumbs provider.

navbar
navbar2

The Kotlin plugin [disables](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/codeInsight/BreadcrumbsProviderCompatBase.kt#L14) classic breadcrumbs when the navigation bar model is enabled, but I suppose it doesn't hurt to show both?

changelog: Introduce navigation bar breadcrumbs.

### DIFF
--- a/src/main/kotlin/org/rust/ide/navigationToolbar/RsNavBarModelExtension.kt
+++ b/src/main/kotlin/org/rust/ide/navigationToolbar/RsNavBarModelExtension.kt
@@ -1,0 +1,30 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.navigationToolbar
+
+import com.intellij.ide.navigationToolbar.StructureAwareNavBarModelExtension
+import com.intellij.lang.Language
+import org.rust.ide.miscExtensions.RsBreadcrumbsInfoProvider
+import org.rust.lang.RsLanguage
+import org.rust.lang.core.psi.RsFile
+import org.rust.lang.core.psi.ext.RsElement
+
+class RsNavBarModelExtension : StructureAwareNavBarModelExtension() {
+    override val language: Language = RsLanguage
+
+    override fun getPresentableText(item: Any?): String? {
+        val element = item as? RsElement ?: return null
+        if (element is RsFile) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/navigationToolbar/RsNavBarModelExtension.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
```
Undin, line 4: Wrong license comment

Kobzol: Oops :sweat_smile:. That's what I get for copy pasting too much :)

src/main/kotlin/org/rust/ide/navigationToolbar/RsNavBarModelExtension.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    override fun getPresentableText(item: Any?): String? {
+        val element = item as? RsElement ?: return null
+        if (element is RsFile) {
+            return element.name
+        }
+
+        val provider = RsBreadcrumbsInfoProvider()
+        if (provider.acceptElement(element)) {
+            return RsBreadcrumbsInfoProvider().getElementInfo(element)
```
mchernyavsky, line 26:
```suggestion
            return provider.getElementInfo(element)
```

mchernyavsky (approved):
LGTM

Kobzol:
bors r=mchernyavsky

mchernyavsky:
bors retry

Undin:
It would be great to have tests for this feature to ensure that we don't have things like #7153

Issue #6583, opened 2020-12-29
Support structured build output tool window for cargo check run configurations


Mcat12:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3572-203
* **Rust toolchain version:** 1.48.0
* **IDE name and version:** CLion 2020.3
* **Operating system:** Fedora 33

## Problem description
The structured build tool window shows up for `build`/`run`/`test`, but does not work for `check`. Running a check build is usually much faster than even a debug build, so it is often used to check for errors. Having the structured build output window for check builds would improve the user experience of fixing errors.

## Steps to reproduce
Create a `cargo check` run configuration and see that it opens a text-only build output tool window.
Create a `cargo build` run configuration and see that it opens in a structured build output tool window.

sabbyX:
maybe for clippy too 🙂

PR #6590, opened 2020-12-30
RUN: Support Build tool window for cargo check and clippy

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/6583.

changelog: Support `Build` tool window for `cargo check` and `clippy`

### DIFF
--- a/src/test/kotlin/org/rustSlowTests/cargo/runconfig/RunConfigurationTestBase.kt
+++ b/src/test/kotlin/org/rustSlowTests/cargo/runconfig/RunConfigurationTestBase.kt
@@ -5,11 +5,7 @@
 
 package org.rustSlowTests.cargo.runconfig
 
-import com.intellij.execution.ExecutionResult
-import com.intellij.execution.ExecutorRegistry
-import com.intellij.execution.Location
-import com.intellij.execution.RunnerAndConfigurationSettings
-import com.intellij.execution.RunManager
+import com.intellij.execution.*
 import com.intellij.execution.actions.ConfigurationContext
 import com.intellij.execution.actions.RunConfigurationProducer
 import com.intellij.execution.configurations.RunConfiguration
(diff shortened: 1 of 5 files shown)


src/test/kotlin/org/rustSlowTests/cargo/runconfig/RunConfigurationTestBase.kt
```diff
@@ -97,6 +93,12 @@ abstract class RunConfigurationTestBase : RsWithToolchainTestBase() {
         return CargoBuildManager.build(buildConfiguration).get()
     }
 
+    protected fun checkProject(useClippy: Boolean = false): CargoBuildResult {
+        val checkConfiguration = createBuildConfiguration()
+        checkConfiguration.configuration.command = if (useClippy) "clippy" else "check"
+        return CargoBuildManager.build(checkConfiguration).get()
+    }
```
Undin, line 100: This method looks very similar with `buildProject`. Can we unify them? For example, pass command name to `buildProject` method with some default value (`run` for example)

src/test/kotlin/org/rustSlowTests/cargo/runconfig/RunConfigurationTestBase.kt
```diff
@@ -97,6 +93,12 @@ abstract class RunConfigurationTestBase : RsWithToolchainTestBase() {
         return CargoBuildManager.build(buildConfiguration).get()
     }
 
+    protected fun checkProject(useClippy: Boolean = false): CargoBuildResult {
+        val checkConfiguration = createBuildConfiguration()
+        checkConfiguration.configuration.command = if (useClippy) "clippy" else "check"
```
Undin, line 98: Let's avoid such hacks. Here you create build configuration for default cargo command (`run` I suppose) and only after that switch actual cargo command. As a result, you didn't check your changes in `CargoBuildManager`. Try to revert them and run the corresponding tests, they will pass.

I suppose the right way to do it is to create `CargoCommandConfiguration` with proper command, for example, pass desired command to `org.rustSlowTests.cargo.runconfig.RunConfigurationTestBase#createConfiguration` (it may be also useful in `RsCoverageTest`)

Undin (changes requested):
In general looks ok 👍

src/test/kotlin/org/rustSlowTests/cargo/runconfig/RunConfigurationTestBase.kt
Undin: BTW currently we don't have tests for `build` command itself, only for `run`

mchernyavsky:
@Undin Fixed.

Undin (approved):
bors r+

aldanor:
Great, thanks @mchernyavsky, was looking forward to it so much, hoping to see it in the next release!

PR #6591, opened 2020-12-30
INSP: add inspection that checks non-existent crates

Kobzol:
This PR add more basic infrastructure to TOML, based on the new local crate index. The first commit introduces a simple inspection which checks crates in `Cargo.toml` that do not exist (=were not found in the index). It uses the new TOML visitor (https://github.com/intellij-rust/intellij-rust/pull/6578) and also the dependency visitor from https://github.com/intellij-rust/intellij-rust/pull/3787. After another inspection is added, this dependency visitor should be extracted into a separate file.

The second commit then introduces tests for the inspection, which however require a lot of new infrastructure. Mainly it adds a service interface and a test implementation for the local crate index (taken from https://github.com/intellij-rust/intellij-rust/pull/5988).

Closes: https://github.com/intellij-rust/intellij-rust/pull/3787
Closes: https://github.com/intellij-rust/intellij-rust/pull/5988
Related: https://github.com/intellij-rust/intellij-rust/issues/6463

changelog: Add inspection that checks non-existent crates in Cargo.toml. Enabled only with the experimental local crate index.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
@@ -1,0 +1,20 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import com.intellij.openapi.components.service
+
+interface CratesLocalIndexService {
+    fun getCrate(crateName: String): CargoRegistryCrate?
+    fun getAllCrateNames(): List<String>
+
+    companion object {
+        fun getInstance(): CratesLocalIndexService = service()
+    }
+}
+
+data class CargoRegistryCrate(val versions: List<CargoRegistryCrateVersion>)
+data class CargoRegistryCrateVersion(val version: String, val isYanked: Boolean, val features: List<String>)
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/TestCrateResolverService.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/TestCrateResolverService.kt
@@ -1,0 +1,27 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import org.jetbrains.annotations.TestOnly
+
+class TestCratesLocalIndexServiceImpl : CratesLocalIndexService {
+    var testCrates: Map<String, CargoRegistryCrate> = emptyMap()
+
+    override fun getCrate(crateName: String): CargoRegistryCrate? = testCrates[crateName]
+    override fun getAllCrateNames(): List<String> = testCrates.keys.toList()
+}
+
+@TestOnly
+fun withMockedCrates(crates: Map<String, CargoRegistryCrate>, action: () -> Unit) {
+    val resolver = CratesLocalIndexService.getInstance() as TestCratesLocalIndexServiceImpl
+    val orgCrates = resolver.testCrates
+    try {
+        resolver.testCrates = crates
+        action()
+    } finally {
+        resolver.testCrates = orgCrates
+    }
+}
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
@@ -1,0 +1,112 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.inspections
+
+import com.intellij.codeInspection.*
+import com.intellij.psi.PsiElementVisitor
+import org.rust.cargo.CargoConstants
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
+import org.rust.toml.crates.local.CratesLocalIndexService
+import org.rust.toml.isDependencyKey
+import org.toml.lang.psi.*
+import org.toml.lang.psi.ext.TomlLiteralKind
+import org.toml.lang.psi.ext.kind
+import org.toml.lang.psi.ext.name
+
+class CrateNotFoundInspection : LocalInspectionTool() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor =
+        buildVisitor(holder) ?: super.buildVisitor(holder, isOnTheFly)
+
+    private fun buildVisitor(holder: ProblemsHolder): PsiElementVisitor? {
+        if (!isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) return null
+        if (holder.file.name != CargoConstants.MANIFEST_FILE) return null
+
+        return object : TomlVisitor() {
+            override fun visitKeyValue(element: TomlKeyValue) {
+                val table = element.parent as? TomlTable ?: return
+                val depTable = DependencyTable.fromTomlTable(table) ?: return
+                if (depTable !is DependencyTable.General) return
+
+                val segment = element.key.segments.getOrNull(0) ?: return
+                val name = segment.name ?: return
+                val value = element.value ?: return
+                if (value is TomlLiteral && value.kind is TomlLiteralKind.String) {
+                    handleDependency(DependencyCrate(name, segment, mapOf("version" to value)), holder)
+                } else if (value is TomlInlineTable) {
+                    handleDependency(DependencyCrate(name, segment, collectProperties(value)), holder)
+                }
+            }
+
+            override fun visitTable(element: TomlTable) {
+                val depTable = DependencyTable.fromTomlTable(element) ?: return
+                if (depTable !is DependencyTable.Specific) return
+
+                handleDependency(
+                    DependencyCrate(depTable.crateName, depTable.crateNameElement, collectProperties(element)), holder
+                )
+            }
+        }
+    }
+
+    private fun handleDependency(dependency: DependencyCrate, holder: ProblemsHolder) {
+        // Do not check local, custom or git crates
+        for (property in IGNORED_PROPERTIES) {
+            if (property in dependency.properties) return
+        }
+
+        if (CratesLocalIndexService.getInstance().getCrate(dependency.crateName) == null) {
+            holder.registerProblem(dependency.crateNameElement, "Crate ${dependency.crateName} not found")
+        }
+    }
+
+    companion object {
+        private val IGNORED_PROPERTIES = listOf("git", "path", "registry")
+    }
+}
+
+private fun collectProperties(owner: TomlKeyValueOwner): Map<String, TomlValue> {
+    return owner.entries.mapNotNull {
+        val name = it.key.name ?: return@mapNotNull null
+        val value = it.value ?: return@mapNotNull null
+        name to value
+    }.toMap()
+}
+
+private data class DependencyCrate(
+    val crateName: String,
+    val crateNameElement: TomlKeySegment,
+    val properties: Map<String, TomlValue>
+)
+
+private sealed class DependencyTable {
+    object General : DependencyTable()
+    data class Specific(val crateName: String, val crateNameElement: TomlKeySegment) : DependencyTable()
+
+    companion object {
+        fun fromTomlTable(table: TomlTable): DependencyTable? {
+            val key = table.header.key ?: return null
+            val segments = key.segments
+            val dependencyNameIndex = segments.indexOfFirst { it.isDependencyKey }
+
+            return when {
+                // [dependencies], [x86.dev-dependencies], etc.
+                dependencyNameIndex == segments.lastIndex -> General
+                // [dependencies.crate]
+                dependencyNameIndex != -1 -> {
+                    val crate = segments.getOrNull(dependencyNameIndex + 1)
+                    val crateName = crate?.name
+                    if (crate != null && crateName != null) {
+                        Specific(crateName, crate)
+                    } else {
+                        null
+                    }
+                }
+                else -> null
+            }
+        }
+    }
+}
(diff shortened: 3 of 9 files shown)


Kobzol:
We may need to add some method to the crate resolver interface for checking if the crate resolver is ready. Otherwise this inspection could result in false positives if `getCrate` returns null.

Btw, when running CLion locally with this PR, the local crate index sometimes does not have any entry e.g. for `serde` or `log`, and `getAllCrateNames` returns only 64 crate names, IDK why.

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -337,3 +312,20 @@ private object CrateExternalizer : DataExternalizer<CargoRegistryCrate> {
         return CargoRegistryCrate(versions)
     }
 }
+
+private data class ParsedVersion(
```
avrong, line 316: Can we move it to CargoRegistryCrateVersion's companion as `ParsedVersion` has no use outside of the `crateFromJson` function?

avrong (changes requested):

toml/src/test/kotlin/org/rust/toml/inspections/CrateNotFoundInspectionTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.rust.ide.experiments.RsExperiments
+import org.rust.ide.inspections.RsInspectionsTestBase
+import org.rust.openapiext.runWithEnabledFeature
+import org.rust.toml.crates.local.CargoRegistryCrate
+import org.rust.toml.crates.local.CargoRegistryCrateVersion
+import org.rust.toml.crates.local.withMockedCrates
+
+class CrateNotFoundInspectionTest : RsInspectionsTestBase(CrateNotFoundInspection::class) {
+    fun `test missing crate in dependencies`() = doTest(
+        """
```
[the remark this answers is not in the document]
Kobzol: Uh, no idea why it was formatted like that. Thanks.

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
Kobzol: I suppose that it should be in this file, as it's specific to the format of the local index? `CargoRegistryCrateVersion` should be a general interface describing a crate version. For example, if we later port the crates.io resolver to this service interface, it should use `CargoRegistryCrateVersion`, but it shouldn't have to know about `ParsedVersion`.

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+
+                // [dependencies], [x86.dev-dependencies], etc.
+                if (dependencyNameIndex == names.lastIndex) {
+                    element.entries.forEach {
+                        val name = it.key.name ?: return@forEach
+                        val value = it.value ?: return@forEach
+                        if (value is TomlLiteral && value.kind is TomlLiteralKind.String) {
+                            handleDependency(Dependency(name, it.key, mapOf("version" to value)), holder)
+                        }
+                        else if (value is TomlInlineTable) {
```
avrong, line 41:
```suggestion
                        } else if (value is TomlInlineTable) {
```

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+                            handleDependency(Dependency(name, it.key, collectProperties(value)), holder)
+                        }
+                    }
+                }
+                else if (dependencyNameIndex != -1) {
```
avrong, line 46:
```suggestion
                } else if (dependencyNameIndex != -1) {
```

toml/src/main/kotlin/org/rust/toml/crates/local/TestCrateResolverService.kt
```diff
@@ -0,0 +1,28 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import org.jetbrains.annotations.TestOnly
+
+@TestOnly
```
ortem, line 10:
```suggestion
```
This class is registered in XML so it should not be excluded. Otherwise, the compilation of other plugins depending on IJ Rust would be broken. See `TestCargoProjectsServiceImpl` for example

toml/src/main/resources/inspectionDescriptions/CrateNotFound.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+Detects unknown dependency crates in Cargo.toml.
```
ortem, line 3:
```suggestion
Detects unknown dependency crates in Cargo.toml
```

toml/src/test/kotlin/org/rust/toml/inspections/CrateNotFoundInspectionTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    """)
+
+    fun `test missing crate in specific dependency`() = doTest("""
+        [dependencies.<warning descr="Crate foo not found">foo</warning>]
+    """)
+
+    fun `test existing crate in dependencies`() = doTest("""
+        [dependencies]
+        foo = "1"
+    """, crate("foo", "1"))
```
ortem, line 40: Let's add more tests with several crates provided, both positive (with `<warning>`) and negative

toml/src/main/kotlin/org/rust/toml/crates/local/CrateResolverService.kt
```diff
@@ -0,0 +1,20 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import com.intellij.openapi.components.service
+
+interface CrateResolverService {
```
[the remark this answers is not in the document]
Kobzol: Originally I thought that we also might implement this interface with the crates.io resolver. Is that futile?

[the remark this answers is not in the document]
ortem: Frankly, I'm not sure if this would be very useful in general since using crates.io API is slow and isn't suitable enough for real-time inspections in IDE. But if you know cases when such implementation would be nice to have, please feel free to describe them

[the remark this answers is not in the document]
Kobzol: I only wanted to do that to make the inspections work if the local (git) is not available for some reason. But I agree that its functionality (and latency) leaves a lot to be desired. I'll rename the service :)

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private fun handleDependency(dependency: Dependency, holder: ProblemsHolder) {
+        if (CratesLocalIndexService.getInstance().getCrate(dependency.name) == null) {
+            holder.registerProblem(dependency.nameElement, "Crate ${dependency.name} not found")
+        }
+    }
+}
+
+data class Dependency(val name: String, val nameElement: TomlKeySegment, val properties: Map<String, TomlValue>)
```
ortem, line 60: > Potential memory leak: don't hold PsiElement, use SmartPsiElementPointer instead 

Let's ultimately get rid of this class and pass its values separately. Also, `properties` seems to be unused

ortem (changes requested):
Currently, there are lots of fake git changes, so it's difficult to look into the real changes. To avoid these fake changes, let's separate `CratesLocalIndexService` refactoring into two commits:
* Extract `CratesLocalIndexServiceImpl` (`CratesLocalIndexService` class renaming)
* Add CratesLocalIndexService interface (new `CratesLocalIndexService` interface introducing)

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
Kobzol: I was already preparing a bit for other inspections, the Dependency code from this file will be probably refactored into a separate visitor to be used by multiple other inspections that might require the properties etc. But you are right that currently it's used only for this one inspection.

toml/src/main/resources/META-INF/toml-only.xml
```diff
@@ -27,8 +27,15 @@
                          displayName="Cyclic feature dependency"
                          enabledByDefault="true" level="ERROR"
                          implementationClass="org.rust.toml.inspections.CargoTomlCyclicFeatureInspection"/>
+        <localInspection groupName="Rust"
+                         displayName="Crate not found"
+                         enabledByDefault="true" level="WARNING"
+                         implementationClass="org.rust.toml.inspections.CrateNotFoundInspection"/>
 
-        <applicationService serviceImplementation="org.rust.toml.crates.local.CratesLocalIndexService"/>
+        <applicationService
```
ortem, line 35: Let's use the same formatting as for `CargoProjectsService`:
```
<applicationService serviceInterface="org.rust.toml.crates.local.CratesLocalIndexService"
                    serviceImplementation="org.rust.toml.crates.local.CratesLocalIndexServiceImpl"
                    testServiceImplementation="org.rust.toml.crates.local.TestCratesLocalIndexServiceImpl"/>
```

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+        else -> null
+    }
+}
+
+private sealed class DependencyTableBase {
+    object DependencyTable : DependencyTableBase()
+    data class SpecificDependencyTable(val crate: TomlKeySegment, val crateName: String) : DependencyTableBase()
+}
```
ortem, line 84: What do you think about this naming and fields order?
```suggestion
private sealed class DependencyTable {
    object General : DependencyTable()
    data class Specific(val crateName: String, val crateElement: TomlKeySegment) : DependencyTable()
}
```

toml/src/main/resources/META-INF/toml-only.xml
```diff
@@ -27,8 +27,15 @@
                          displayName="Cyclic feature dependency"
                          enabledByDefault="true" level="ERROR"
                          implementationClass="org.rust.toml.inspections.CargoTomlCyclicFeatureInspection"/>
+        <localInspection groupName="Rust"
```
ortem, line 30:
```suggestion
        <localInspection language="TOML" groupName="Rust"
```
it's just missed, I suppose? See `CargoTomlCyclicFeatureInspection` declaration above

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.rust.cargo.CargoConstants
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
+import org.rust.toml.crates.local.CratesLocalIndexService
+import org.rust.toml.isDependencyKey
+import org.toml.lang.psi.*
+import org.toml.lang.psi.ext.TomlLiteralKind
+import org.toml.lang.psi.ext.kind
+
+class CrateNotFoundInspection : LocalInspectionTool() {
```
ortem, line 19: I think this inspection should ignore dependencies with `path` or `git` specified? Seems like currently false-positive warning is shown instead. I mean something like `my_local_crate = { version = "1.0", path = "../my_local_crate" }` ([docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations))

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
ortem: Also, looks like `TomlInlineTable` case is not covered by tests

Kobzol: Well, I guess then it's back to having the `Dependency` class and parsing its properties? :sweat_smile: Otherwise I won't be able to check if it's a local/git dependency.

ortem: Hm, probably 😄 
However please try to avoid holding `PsiElement`s. If it is necessary, `org.rust.openapiext.UtilsKt#createSmartPointer` might help to store a smart pointer instead

Kobzol: Is it a problem in this case? I'm basically implementing a visitor which calls `visitDependency(object containing several PSI elements)`. It seems to me that it is the same situation as in normal PSI visitors.

ortem: Oh, sorry, I looked at the code very briefly and didn't figure the problem out!
Looks like this warning is false-positive. I'll report this case to IntelliJ team a bit later. For now, you can either add `@Suppress("StatefulEp")` for `Dependency` class or rename it to something else (seems like it solves the warning), e.g. `DependencyCrate`

ortem: BTW `DependencyCrate` and `DependencyTable` looks similar. Let's put them nearby and use the same naming for them, e.g. `val crateName: String, val crateElement: TomlKeySegment`.

`getDependencyTable` function could also be a factory method of `DependencyTable` named `from` or something like that, as far a I see.

ortem (approved):
Thank you!

bors r+

Kobzol:
bors retry

Kobzol:
bors retry

Kobzol:
Thanks to @mibac138 for the groundwork!

Kobzol:
@dima74 Maybe this shouldn't be put into v141 until https://github.com/intellij-rust/intellij-rust/issues/6812 is resolved.

avrong:
@Kobzol discussed in private with @ortem. Decided that it's ok to have it now in the release as it under the experimental feature flag anyway. We will look into #6812 in details soon

Kobzol:
Oh, I forgot that it's experimental. Then it's ok :)

PR #6592, opened 2020-12-30
Make `RsInferenceResult` immutable

Undin:
Use `Map` instead of `MutableMap` that was used by accident and not necessary here


Undin:
CC @vlad20012 @ortem

PR #6599, opened 2021-01-01
CARGO: Use semver to sort crate versions

avrong:
Gives the ability to compare crate versions using semantic versioning and sort them this way instead of the default in index version uploading date order. That's useful for completion and inspections in the future.

Related: #6463

changelog: Use semantic versioning to improve version sorting in dependencies when using experimental local crates index

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProvider.kt
+++ b/toml/src/main/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProvider.kt
@@ -7,9 +7,13 @@ package org.rust.toml.completion
 
 import com.intellij.codeInsight.AutoPopupController
 import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.completion.CompletionSorter
 import com.intellij.codeInsight.completion.CompletionUtil
 import com.intellij.codeInsight.completion.PrioritizedLookupElement
+import com.intellij.codeInsight.completion.impl.RealPrefixMatchingWeigher
+import com.intellij.codeInsight.lookup.LookupElement
 import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.codeInsight.lookup.LookupElementWeigher
 import com.intellij.icons.AllIcons
 import com.intellij.openapi.editor.EditorModificationUtil
 import org.rust.lang.core.completion.nextCharIs
@@ -54,7 +58,7 @@ class LocalCargoTomlDependencyCompletionProvider : TomlKeyValueCompletionProvide
 
         val indexService = CratesLocalIndexService.getInstance()
 
-        val versions = indexService.getCrate(name)?.versions ?: return
+        val versions = indexService.getCrate(name)?.sortedVersions ?: return
         val elements = versions.mapIndexed { index, variant ->
             val lookupElement = LookupElementBuilder.create(variant.version)
                 .withInsertHandler(StringValueInsertionHandler(keyValue))
@@ -65,6 +69,13 @@ class LocalCargoTomlDependencyCompletionProvider : TomlKeyValueCompletionProvide
                 index.toDouble()
             )
         }
-        result.addAllElements(elements)
+        val sorter = CompletionSorter.emptySorter()
+            .weigh(RealPrefixMatchingWeigher())
+            .weigh(object : LookupElementWeigher("priority", true, false) {
+                override fun weigh(element: LookupElement): Double = (element as PrioritizedLookupElement<*>).priority
+            })
+
+        result.withRelevanceSorter(sorter).addAllElements(elements)
     }
 }
+
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
@@ -6,6 +6,8 @@
 package org.rust.toml.crates.local
 
 import com.intellij.openapi.components.service
+import com.vdurmont.semver4j.Semver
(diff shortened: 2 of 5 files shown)


Kobzol:
Oh, I see that we chose the same library for semver, so it should be good :D I started working on semver requirements parsing in https://github.com/Kobzol/intellij-rust/commit/7cb437341b3e74243314519670d74fee0c854e8e, but it needs several other PRs (probably including this one) to be merged first.

toml/src/main/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProvider.kt
```diff
@@ -54,16 +53,13 @@ class LocalCargoTomlDependencyCompletionProvider : TomlKeyValueCompletionProvide
...
                 .withInsertHandler(StringValueInsertionHandler(keyValue))
                 .withTailText(if (variant.isYanked) " yanked" else null)
 
-            PrioritizedLookupElement.withPriority(
-                lookupElement,
-                index.toDouble()
-            )
+            PrioritizedLookupElement.withExplicitProximity(lookupElement, index)
         }
         result.addAllElements(elements)
```
ortem, line 64: Discussed in person the following problem with sorting:
image
Here, `0.2.1` should go after `0.2.10`, `0.2.9`, and so on. The problem is due to the `liftShorter` weigher: looks like it has a higher priority than others (see `com.intellij.codeInsight.completion.BaseCompletionService#defaultSorter` for more detail). To avoid this situation, let's try using an empty sorter:
```
result.withRelevanceSorter(CompletionSorter.emptySorter()).addAllElements(elements)
```

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -279,8 +281,32 @@ class CratesLocalIndexService : PersistentStateComponent<CratesLocalIndexState>,
...
+    val version: Semver? by lazy {
+        try {
+            Semver(versionString, Semver.SemverType.NPM)
+        } catch (e: SemverException) {
+            null
+        }
+    }
+
+    override fun compareTo(other: CargoRegistryCrateVersion): Int {
+        return this.version?.compareTo(other.version) ?: return 0
```
ortem, line 307: Note that currently object with `version == null` is considered **equal** to any other object. But I guess `null` should be considered **less** than any other objects. Also, calling `com.vdurmont.semver4j.Semver#compareTo` with `null` parameter value might be dangerous.

However, probably we should just ignore `CargoRegistryCrateVersion` with `version == null`. What do you think?
We can implement such logic by making a private constructor
```Kotlin
data class CargoRegistryCrateVersion private constructor(
    val version: Semver,
    val isYanked: Boolean,
    val features: List<String>
)
```
with a factory method to instantiate `CargoRegistryCrateVersion` from `versionString: String`

toml/src/main/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProvider.kt
```diff
@@ -65,6 +68,8 @@ class LocalCargoTomlDependencyCompletionProvider : TomlKeyValueCompletionProvide
                 index.toDouble()
             )
         }
-        result.addAllElements(elements)
+        result.withRelevanceSorter(CompletionSorter.emptySorter().weigh(object : LookupElementWeigher("priorityWeigher", true, false) {
```
ortem, line 71: WDYT about using "priority" instead of "priorityWeigher"? What's the reason for using "priorityWeigher"?

ortem, line 71: I'd reformat this code to improve readability, maybe like this:
```Kotlin
val weigher = CompletionSorter.emptySorter()
    .weigh(object : LookupElementWeigher("priority", true, false) {
        override fun weigh(element: LookupElement): Double = (element as PrioritizedLookupElement<*>).priority
    })

result.withRelevanceSorter(weigher).addAllElements(elements)
```

ortem: And maybe we should add `.weigh(RealPrefixMatchingWeigher())` before our weigher
Screenshot 2021-03-26 at 14 38 05

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -18,6 +20,10 @@ interface CratesLocalIndexService {
 }
 
 data class CargoRegistryCrate(val versions: List<CargoRegistryCrateVersion>) {
+    val sortedVersions: List<CargoRegistryCrateVersion> by lazy {
```
ortem, line 23: Is it really useful to keep this list as a field? AFAIU `getCrate` invokes `read` which instantiates a new  `CargoRegistryCrate` every time. So a new `CargoRegistryCrate` object is created on every code completion session.

Related to `semanticVersion` as well

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
avrong: As far as I tested, it seems that PersistentHashMap keeps the structures in the memory and doesn't reinitialize them if there is no need for that. At least the breakpoints on the constructors are being called only the first time I call completion

ortem (approved):
LGTM

bors r+

PR #6602, opened 2021-01-02
ANN: gray out disabled cfg_attr attributes

vlad20012:
A little improvement for #6497: highlight disabled `cfg_attr`s as conditionally disabled code:

image

Nested `cfg_attr`s (e.g. `#[cfg_attr(foo, cfg_attr(bar, foobar))]`) are not highlighted (and why someone needs to write them?)

changelog: Gray out disabled `#[cfg_attr()]` attributes

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsCfgDisabledCodeAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsCfgDisabledCodeAnnotator.kt
@@ -11,20 +11,38 @@ import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
 import org.rust.ide.colors.RsColor
+import org.rust.ide.utils.isDisabledCfgAttrAttribute
+import org.rust.lang.core.psi.ext.RsAttr
 import org.rust.lang.core.psi.ext.RsDocAndAttributeOwner
 import org.rust.lang.core.psi.ext.isEnabledByCfgSelf
+import org.rust.lang.core.psi.ext.owner
 
 class RsCfgDisabledCodeAnnotator : AnnotatorBase() {
     override fun annotateInternal(element: PsiElement, holder: AnnotationHolder) {
         if (holder.isBatchMode) return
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/rust/ide/annotator/RsCfgDisabledCodeAnnotator.kt
```diff
@@ -11,20 +11,38 @@ import com.intellij.lang.annotation.HighlightSeverity
...
         if (element is RsDocAndAttributeOwner && !element.isEnabledByCfgSelf) {
-            val color = RsColor.CFG_DISABLED_CODE
-            val severity = if (isUnitTestMode) color.testSeverity else HighlightSeverity.INFORMATION
+            holder.createCondDisabledAnnotation()
+        }
 
-            holder.newAnnotation(severity, "Conditionally disabled code")
-                .textAttributes(color.textAttributesKey)
-                .create()
+        if (element is RsAttr && element.isDisabledCfgAttrAttribute && element.owner?.isEnabledByCfgSelf == true) {
```
Undin, line 28: What's about nested `cfg_attr`s?

vlad20012: From PR descriontion:

> Nested cfg_attrs (e.g. #[cfg_attr(foo, cfg_attr(bar, foobar))]) are not highlighted (and why someone needs to write them?)

Undin: Oh, sorry, I completely missed this part of PR description.

> why someone needs to write them

I don't know why but it's possible and looks like it's quite easy to support this case too from our side. But since it's a quite rare case (I failed to find such cases in top 2000 crates from crates.io), I don't insist so it's up to you

vlad20012: I'd not say that it is easy. I mean, technically it's easy, but I don't understand how such a highlighting should look like. There are several options and I don't like all of them

Issue #6612, opened 2021-01-03
"Fill missing parameters"  do nothing for closures


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.139.3583-203-nightly
* **Rust toolchain version:** 1.50.0-nightly (f76ecd066 2020-12-15) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3.1 (CL-203.6682.181)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description

"Fill missing parameters" intention action doesn't work for closures.

## Steps to reproduce

Consider

```rust
fn func(param: i32) {}

fn main() {
    let closure = |x| (x);

    func();
    closure();
}
```

... place care between  parentheses  in `func()` call and invoke "Fill missing parameters":

```rust
func(0);
```

... now do the same for `closure()` call. Nothing will happen.

PR #6613, opened 2021-01-04
ANN: improve fill function arguments fix

Kobzol:
This PR adds basic support for closures to `FillFunctionArgumentsFix` (introduced in https://github.com/intellij-rust/intellij-rust/pull/5644) and also resolves function types with substituted types.

The substitution doesn't yet work properly for method calls (see https://github.com/intellij-rust/intellij-rust/pull/5644#discussion_r555034800). 

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6612
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6645

changelog: Enable the new "Fill function arguments" quick-fix for closures and take generic arguments into consideration when filling default values.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
@@ -15,9 +15,8 @@ import org.rust.ide.annotator.expectedParamsCount
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.knownItems
-import org.rust.lang.core.types.selfType
 import org.rust.lang.core.types.ty.Ty
-import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.ty.TyFunction
 import org.rust.lang.core.types.type
 import org.rust.openapiext.buildAndRunTemplate
 import org.rust.openapiext.createSmartPointer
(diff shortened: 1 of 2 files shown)


Undin (commented):
See https://github.com/intellij-rust/intellij-rust/pull/5644#r554568911. If I didn't miss anything, we can use a more generic approach to support closures

Kobzol:
I tried using the expression as `TyFunction` at first, but I think that I hit some problem. But maybe it was unrelated, I'll try it.

src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
```diff
@@ -64,44 +63,10 @@ class FillFunctionArgumentsFix(element: PsiElement) : LocalQuickFixAndIntentionA
...
-                    }
-                    else {
-                        valueParameters
-                    }
-                }
-                is RsFieldsOwner -> target.tupleFields?.tupleFieldDeclList?.map { it.typeReference.type }
-                else -> null
-            }
-        }
+        is RsCallExpr -> (element.expr.type as? TyFunction)?.paramTypes
```
Undin, line 66: I like such fixes very much when you mostly remove code and something becomes work better 👍

Undin (approved):
LGTM
Thanks!
bors r+

src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
Kobzol: Me too :) I think that I have tried using the `type` directly for the case of a temporary lambda and it didn't work, but I didn't realize that it's because the plugin doesn't support these lambda cases, but overall the function `type` inference works fine.

Undin:
bors retry

Issue #6620, opened 2021-01-04
Broken remove redundant type arguments


PizzaCrust:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3572-203
* **Rust toolchain version:** 1.51.0-nightly (fde692739 2021-01-02) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2020.3.1 (CL-203.6682.181)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description
```rust
struct B<'a, T>;

struct C<'a> {
    a: B<'a, u32, u32>
}
```
It fails to remove redundant type arguments

## Steps to reproduce
1. Place aforementioned code in editor
2. Attempt to `remove redundant type arguments`

PR #6621, opened 2021-01-04
ANN: allow removing redundant type arguments from types with lifetimes

Kobzol:
The `RemoveTypeArguments` fix was introduced in https://github.com/intellij-rust/intellij-rust/pull/1305 and for some reason it ignored types with lifetimes (I hadn't found why in the PR description. @farodin91 do you recall? :D ). I suppose that it shouldn't be a problem, so I removed the condition to fix https://github.com/intellij-rust/intellij-rust/issues/6620. If you are aware of any issues that might arise from this, let me know.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6620

changelog: Allow removing redundant type arguments from types with lifetimes

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/RemoveTypeArguments.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/RemoveTypeArguments.kt
@@ -10,6 +10,8 @@ import com.intellij.codeInspection.ProblemDescriptor
 import com.intellij.openapi.project.Project
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.deleteWithSurroundingComma
+import org.rust.lang.core.psi.ext.getNextNonCommentSibling
 
 class RemoveTypeArguments(
     private val startIndex: Int,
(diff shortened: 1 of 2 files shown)


farodin91:
> @farodin91 do you recall? 

Sry, i can't remember.

Kobzol:
Well, then I suppose we can remove it and see if it breaks anything :D

src/main/kotlin/org/rust/ide/inspections/fixes/RemoveTypeArguments.kt
```diff
@@ -32,8 +32,6 @@ class RemoveTypeArguments(
     }
 
     private fun RsTypeArgumentList.removeTypeParameters() {
-        if (lifetimeList.size > 0) return
```
Undin, line 35: Unfortunately, without this condition, next line breaks code when the declaration doesn't have type parameter.

Undin: Also, don't forget about const generics.

the quick-fix probably shouldn't be suggested at all because it's not clear what type argument it should remove.

But it seems, all our checks and quick fixes don't take into account const arguments so I suppose it can be supported in separate PR

Undin (changes requested):

Kobzol:
Thanks, I fixed it. The fix also fixes the same bug when associated types or const arguments are present (but only when for situations when the last type argument was removed).

Undin (approved):
Thanks!
bors r+

PR #6622, opened 2021-01-04
ANN: add quick fix to remove redundant function call arguments

Kobzol:
This PR adds a quick fix that lets the user quickly remove a redundant function argument. This fix is inspired by a similar one that exists for Kotlin (and probably also for other languages).

remove-argument

I wanted to implement this right after https://github.com/intellij-rust/intellij-rust/pull/5612, but I kind of forgot about it. Another option (instead of removing the argument) is to add a new parameter, but that is blocked on https://github.com/intellij-rust/intellij-rust/pull/5944.

I didn't inherit from `RemoveElementFix`, because I would have to override all of its functions anyway. Currently, whitespace/comments next to the removed argument are preserved, not removed, I'm not sure which is better.

changelog: Add a quick fix to remove redundant function call arguments.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -869,7 +869,9 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             )).addToHolder(holder)
         } else if (!functionType.variadic && realCount > expectedCount) {
             args.exprList.drop(expectedCount).forEach {
-                RsDiagnostic.IncorrectFunctionArgumentCountError(it, expectedCount, realCount, functionType).addToHolder(holder)
+                RsDiagnostic.IncorrectFunctionArgumentCountError(it, expectedCount, realCount, functionType, listOf(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/RemoveArgumentFix.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.deleteWithSurroundingComma
+
+class RemoveArgumentFix(element: RsElement) : LocalQuickFixAndIntentionActionOnPsiElement(element) {
```
dima74, line 16: Maybe rename to `RemoveFunctionArgumentFix`

dima74 (approved):

Kobzol:
bors r=dima74

PR #6623, opened 2021-01-05
TOML: unify toml key parsing

Undin:
Previously, we parsed [dotted keys](https://toml.io/en/v1.0.0-rc.3#keys) in `key1.key2=value` and `[key1.key2]` differently that was quite confusing. Now they are parsed in the same way

changelog: unify toml key parsing

### DIFF
--- a/intellij-toml/src/main/grammars/TomlParser.bnf
+++ b/intellij-toml/src/main/grammars/TomlParser.bnf
@@ -42,16 +42,14 @@ private FileForm_recover ::=
   !<<atNewLine (BARE_KEY | BARE_KEY_OR_NUMBER | BARE_KEY_OR_DATE | BASIC_STRING | LITERAL_STRING | '[')>>
 private EatUntilNextLine_recover ::= !<<atNewLine <<any>>>>
 
-KeyValue ::= LongKey '=' <<atSameLine Value>> { pin = 1 }
+KeyValue ::= Key '=' <<atSameLine Value>> { pin = 1 }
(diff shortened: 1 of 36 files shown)


intellij-toml/src/main/grammars/TomlParser.bnf
```diff
@@ -42,16 +42,14 @@ private FileForm_recover ::=
...
 NewLineKeyValue ::= <<atNewLine KeyValue>> {
     elementType = KeyValue
     recoverWhile = EatUntilNextLine_recover
 }
 
-fake Key ::= BasicKey ('.' BasicKey)*
-LongKey ::= BasicKey ('.' BasicKey)* { elementType = Key }
-ShortKey ::= BasicKey { elementType = Key }
-private BasicKey ::= BareKey | BASIC_STRING | LITERAL_STRING
+Key ::= SimpleKey ('.' SimpleKey)*
```
vlad20012, line 51: I don't like `SimpleKey` name. Maybe `KeySegment`?
Also, there is an option to parse keys with left recursion. We parser Rust paths this way. In this case, SimpleKey will not be needed. Pros/cons?

Undin: Discussed in person. As a result:
- rename `SimpleKey` to `KeySegment`
- we currently don't know the actual cons of "parse keys with left recursion". Also, original TOML grammar parse keys also as segments list so we decided to keep current implementation

Undin:
bors retry

Undin:
bors r=vlad20012

PR #6625, opened 2021-01-05
Drop using old API

Undin:
Remove usage of `FileBasedIndex.registerIndexableSet` and `FileBasedIndex.removeIndexableSet`.
We don't actually use it anymore (it was used before 2020.2 because of bug in the platform) and these methods are dropped in 2021.1


Undin:
CC @vlad20012

PR #6643, opened 2021-01-10
TY: infer type of default type parameters for known stdlib items

Undin:
Recent implementations of `Box` (since 1.49) and `Vec` (latest nightly) provide new allocator API.
So now these items contain additional type parameter for allocator type with some default value.

Previously, we didn't infer type of allocator type parameter for `box` patterns and `vec!` macro.
Now it works as expected

Note, I haven't added any new test since all necessary cases are covered with existing tests. So this PR fixes some failed test with new Rust versions (see #6604)

changelog: Adjust type inference to infer type of allocator type parameter for `Box` and `Vec` items. See the corresponding [RFC](https://rust-lang.github.io/rfcs/1398-kinds-of-allocators.html) about allocator API

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
@@ -1438,19 +1438,9 @@ val RsFunction.type: TyFunction
 private fun Sequence<Ty>.infiniteWithTyUnknown(): Sequence<Ty> =
     this + generateSequence { TyUnknown }
 
-private fun KnownItems.findVecForElementTy(elementTy: Ty): Ty {
-    val ty = Vec?.declaredType ?: TyUnknown
+private fun KnownItems.findVecForElementTy(elementTy: Ty): Ty = Vec.makeTy("T", elementTy)
 
-    val typeParameter = ty.getTypeParameter("T") ?: return ty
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
```diff
@@ -1461,10 +1451,36 @@ private fun KnownItems.findRangeTy(rangeName: String, indexType: Ty?): Ty {
...
-    return boxTy.substitute(mapOf(boxTy.typeArguments[0] as TyTypeParameter to innerTy).toTypeSubst())
+private fun KnownItems.makeBox(innerTy: Ty): Ty = make("T", innerTy) { Box }
+
+private inline fun <T> KnownItems.make(
+    typeParamName: String,
+    innerTy: Ty,
+    itemSupplier: KnownItems.() -> T?
+): Ty where T : RsTypeDeclarationElement,
+            T : RsGenericDeclaration {
+    val item = itemSupplier() ?: return TyUnknown
```
vlad20012, line 1462: Why we need a function parameter that we call immediately? We could pass `T?` as a parameter.

Undin: Actually, we don't need it. I've simplified this function
Thanks!

ortem (approved):
LGTM

PR #6650, opened 2021-01-11
COMP: show all re-exports of the same item in completion

### LINKED ISSUE #5415: bug report: code complete doesn't display std Future
## Environment

* **IntelliJ Rust plugin version:** 0.2.123.3108-201-nightly
* **Rust toolchain version:** rustc 1.45.0-nightly (a74d1862d 2020-05-14)
* **IDE name and version:** clion 2020.1.1

## Problem description
when `async-std` or `tokio` has used in Cargo.toml, code auto-completion list will show `async-std` `Future` or `tokio` `Future`, but there is not std `Future` unless we only type `Future` directly then press `alt+enter`, the list shows `std Future` now.

image

## Steps to reproduce
1. import `async-std` or `tokio` in Cargo.toml
2. type `Future` and wait for the auto-completion list
3. read all entry in the list, there is no `std Future`

Undin:
Previously, we filter out secondary import candidates for the same item from completion suggestions, as a result, if a library (e.g. async-std) re-exports some item from another library/stdlib, completion shows only single suggestion per item.

Now, we show all suggestions that can be imported - the same items as in `Auto Import` quick-fix

Fixes #5415

changelog: Show all re-exports of the same item in completion list. Previously, the plugin shows only one suggestion per item that hid alternative ways to import the item.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
@@ -11,6 +11,7 @@ import com.intellij.codeInsight.completion.CompletionResultSet
 import com.intellij.codeInsight.completion.CompletionUtil
 import com.intellij.codeInsight.completion.InsertionContext
 import com.intellij.codeInsight.lookup.LookupElement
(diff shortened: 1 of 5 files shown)


mchernyavsky (approved):
bors r+

Issue #6646, opened 2021-01-10
can't set environments for a test run config


Sherlock-Holo:
## Environment

* **IntelliJ Rust plugin version:** 0.3.139.3589-203-nightly
* **Rust toolchain version:** rustc 1.51.0-nightly (26438b473 2021-01-08)
* **IDE name and version:** clion 2020.3.1
* **Operating system:** Archlinux

## Problem description
can't set environments for an existing test run config, if set environments, press ctrl+shift+f10 will create a new config named 'xxx(1)', xxx is the origin config name

## Steps to reproduce

1. press ctrl+shift+f10 for a test case, it will create a new run config
2. edit the config environments
3. press ctrl+shift+f10 again, it will create a new config named xxx(1), not reuse the existing config with the new environments

Sherlock-Holo:
I've met this problem in Goland https://youtrack.jetbrains.com/issue/GO-10254
will it be the same problem?

Undin:
@Sherlock-Holo I don't think it's the same problem. But maybe they are related

PR #6654, opened 2021-01-12
RUN: Ignore environment variables when comparing Cargo run configurations

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6646.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/command/CompositeCargoRunConfigurationProducer.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/command/CompositeCargoRunConfigurationProducer.kt
@@ -113,7 +113,6 @@ class CompositeCargoRunConfigurationProducer : CargoRunConfigurationProducer() {
             command != other.command -> false
             backtrace != other.backtrace -> false
             workingDirectory != other.workingDirectory -> false
-            env != other.env -> false
             else -> true
         }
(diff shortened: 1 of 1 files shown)


ortem (approved):
The fix is probably OK. But I've just looked a bit into the `CompositeCargoRunConfigurationProducer` class and found out that `findOrCreateConfigurationFromContext` overloading is a bit strange and probably redundant. For example, `isSame` checked in `findExistingConfiguration` and then rechecked in `findOrCreateConfigurationFromContext`:
```Kotlin
val settings = findExistingConfiguration(context)
if (preferredConfig.configuration.isSame(settings?.configuration)) { ... }
```
In addition, `findOrCreateConfigurationFromContext` and `findExistingConfiguration` methods are rarely overloaded across IDEs and plugins. In contrast, `isConfigurationFromContext` is usually used to implement "similarity" checks.

I'm not sure it is important, but anyway, @mchernyavsky what do you think?

ortem:
@mchernyavsky ping

ortem:
Anyway, this should be merged of course.

bors r+

Undin:
@mchernyavsky can we add tests for this? Something like: create a run configuration from some context, change some parameters, try to create a new run configuration from the same context and check if new configuration is really new or old one

PR #6658, opened 2021-01-13
Add proc-macro-expander source

vlad20012:
I've extracted the Rust source from #6564 in order to make it possible to setup CI pipelines before #6564 is merged. A very internal change


vlad20012:
Oh, hello, 615 lines or Cargo.lock

PR #6659, opened 2021-01-13
PERF: split TyFingerprint for slices and arrays

vlad20012:


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -49,7 +49,7 @@ class RsFileStub(
     override fun getType() = Type
 
     object Type : IStubFileElementType<RsFileStub>(RsLanguage) {
-        private const val STUB_VERSION = 206
+        private const val STUB_VERSION = 207
```
Undin, line 52:
```suggestion
        private const val STUB_VERSION = 208
```

The current version in master is already 207

vlad20012:
ищкы к+

Issue #4308, opened 2019-08-26
Substitute type parameters with actual types in Parameter Info


lancelote:
##  Environment

* **IntelliJ Rust plugin version:** Build from source `release-104` b3e124a8bec678541004e69cb7aba51c247cf714
* **Rust toolchain version:** rustc 1.36.0 (a53f9df32 2019-07-03)
* **IDE name and version:** 2019.2 CLion 192.5728.100
* **Operating system:** macOS 10.14.6

## Problem description

With #4207 in place the Plugin now substitute generic types in completion variants

<img width="781" alt="completion" src="https://user-images.githubusercontent.com/2500320/63672287-50992800-c7ea-11e9-94f1-2a54ef5545ca.png">

Should not the same be done for `Parameter Info` (`cmd + P`) as well? The same example

<img width="782" alt="parameter_hit" src="https://user-images.githubusercontent.com/2500320/63672325-61e23480-c7ea-11e9-8566-d50a6bbba4e2.png">

The implementation is located in `org.rust.ide.hints.parameter.RsParameterInfoHandler`

## Steps to reproduce

```rust
struct S<T1>();

impl <T2> S<T2> {
    fn foo(&self, x: T2) -> T2 { x }
}

fn main() {
    let s = S::<i32>();
    s.foo()
}
```

Kobzol:
I started to work on this some time ago and I'm not sure about `E-easy` :D In any case, I'll revive my branch and try to finish it.

ortem:
@Kobzol OK, I was not pretty sure if this is "easy" or "medium" actually, so I've changed the label 😄

> I'll revive my branch and try to finish it

Great, thank you!

PR #6664, opened 2021-01-13
HINT: Substitute generic parameters in Parameter info

Kobzol:
I'm not sure about using types vs type references for function/method calls.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4308

changelog: Substitute generic parameters in `Parameter Info`.

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/CallInfo.kt
+++ b/src/main/kotlin/org/rust/ide/utils/CallInfo.kt
@@ -5,11 +5,14 @@
 
 package org.rust.ide.utils
 
-import org.rust.ide.presentation.getStubOnlyText
+import org.rust.ide.presentation.render
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.types.emptySubstitution
+import org.rust.lang.core.types.inference
+import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.ty.TyFunction
-import org.rust.lang.core.types.ty.TyTypeParameter
+import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
 
 class CallInfo private constructor(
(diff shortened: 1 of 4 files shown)


ortem:
@Kobzol Could you please rebase onto master?

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
```diff
@@ -22,14 +30,33 @@ class CallInfo private constructor(
...
             val fn = (call.expr as? RsPathExpr)?.path?.reference?.resolve() ?: return null
-            if (fn is RsFunction) return CallInfo(fn)
-
             val ty = call.expr.type
-            if (ty is TyFunction) {
-                val parameterNames = getParameterNames(fn) ?: List(ty.paramTypes.size) { "_" }
-                return CallInfo(ty, parameterNames)
+
+            return when {
+                fn is RsFunction -> {
```
ortem, line 36: Consider extracting branches bodies into private functions. Maybe something like this (if I understand the whole logic right):
```Kotlin
return when {
    fn is RsFunction -> handleFunction(fn, ty)
    ty is TyFunction -> handleEnum(fn, ty)
    else -> null
}
```

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
```diff
@@ -38,11 +65,25 @@ class CallInfo private constructor(
         }
 
         fun resolve(methodCall: RsMethodCall): CallInfo? {
-            return (methodCall.reference.resolve() as? RsFunction)?.let { CallInfo(it) }
+            val function = (methodCall.reference.resolve() as? RsFunction) ?: return null
+            val substitution = let {
```
ortem, line 69: Probably I'm missing something, could you please explain why we cannot use something like `(methodCall.parentDotExpr.type as? TyFunction)?.paramTypes` as the actual types instead of these substitutions?

src/test/kotlin/org/rust/ide/hints/parameter/RsParameterInfoHandlerTest.kt
```diff
@@ -166,4 +169,25 @@ class RsParameterInfoHandlerTest
...
+
+        impl <T2> S<T2> {
+            fn foo(&self, x: T2) -> T2 { x }
+        }
+
+        fn main() {
+            let s = S::<i32>();
+            s.foo(/*caret*/);
+        }
+    """, "x: i32", 0)
```
ortem, line 192: Please add more tests with unknown types, lifetimes, etc.

ortem (changes requested):
I think it's better to show a generic type name in case of unknown actual types.

Before:
Screenshot 2021-04-08 at 20 10 22
After:
Screenshot 2021-04-08 at 20 10 08

The same with lifetimes. BTW do you think lifetimes should be shown? Or maybe just ignore them?

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
Kobzol: `methodCall.parentDotExpr.type` is the type of the return value of the method call, not of the method (function) itself. I haven't found any easier way how to get the actual substituted method type with generics resolved.

Kobzol:
> I think it's better to show a generic type name in case of unknown actual types.

I don't like that too much, as a user I would probably want to see that I'm trying to pass an unresolved/invalid type to the function, and showing `<unknown>` is way more explicit in that regard than showing the original generic parameter name.

Also, I'm not sure how to properly implement it, since e.g. for functions I get the resolved type parameters, where the generic type is already resolved to `TyUnknown`. Maybe zip the resolved types with the original signature and take the original text of the  parameter if the resolved type is unknown? Seems a bit hacky, but it might work.

I would show lifetimes.

Kobzol:
@ortem Rebased on master. I added lifetime rendering and more tests.

ortem:
> as a user I would probably want to see that I'm trying to pass an unresolved/invalid type to the function, and showing <unknown> is way more explicit in that regard

@Kobzol That's right, I agree with you about being explicit when the types are unresolved or invalid. But I also think in some cases a generic type name might be more informative than `<unknown>`, especially in incomplete code.

Let's check another example.
```Rust
fn add_element<T>(xs: &mut Vec<T>, x: T) { xs.push(x) }

fn main() {
    // ...
    add_element(/*caret*/); // user just started thinking about what to pass in this function
}
```
Before:
Screenshot 2021-07-06 at 15 33 40
After:
Screenshot 2021-07-06 at 15 29 42

Here generic type names indicate that elements of `xs` have the same type as `x`. But this fact is not so clear with `<unknown>` in place. 

What do you think about these cases?

Kobzol:
Hmm. Your example convinced me that showing the parameter might be a good idea :) So let's show the actual type if it's not `TyUnknown`, and otherwise show the type from the original parameter declaration?

Kobzol:
I have reworked the implementation to be more unified and added proper support for closures. If the type is unknown, the original type parameter name will be used.

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
```diff
@@ -22,27 +30,79 @@ class CallInfo private constructor(
...
-            return (methodCall.reference.resolve() as? RsFunction)?.let { CallInfo(it) }
+            val function = (methodCall.reference.resolve() as? RsFunction) ?: return null
+            val type = methodCall.inference?.getResolvedMethodType(methodCall) ?: return null
+            return buildFunctionParameters(function, type)
+        }
+
+        private fun buildFunctionParameters(function: RsFunction, ty: Ty): CallInfo {
+            val types = let {
+                val types = (ty as? TyFunction)?.paramTypes
+                    ?: function.valueParameters.mapNotNull { it.typeReference?.type }
```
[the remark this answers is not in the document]
Kobzol: `self` is actually filled in another place (https://github.com/intellij-rust/intellij-rust/pull/6664/files#diff-59b6929890a0d4ca48aad7ebe026cff696f8b855e7a3ef42076b819648b1caf8R107), so it works (there are several tests that work with `self`, even with UFCS).

The parameters handle this kind of automatically, since we access `valueParameters`, which doesn't contain `self`. But `TyFunction.paramTypes` does (check e.g. `RsParameterInfoHandlerTest::test method`), so we need to drop the self type from it. It wasn't needed before this PR, because the actual parameter types were not used, just `valueParameters` were used before.

But what is wrong is that we mix `paramTypes` with `valueParameters` here and then drop the first one, but if `valueParameters` would be used, that would indeed result in a skipped parameter. Since `ty` is always `TyFunction`, we can solve that by just checking that `ty` is `TyFunction` soon (in `resolve` and then this problem should go away). I pushed the corresponding change.

Kobzol:
Hmm, now there's another problem though. The `CallInfo` is used for both normal parameter hints and for inlay hints. With these changes, inlay hints no longer get the name of the corresponding parameter, since the type is known.

The problem is that (for displaying generic parameter types), normal hints use the `type` attribute of the parameter, while inlay hints use the `name` attribute of the parameter. Maybe we should change `CallInfo` so that the parameter contains
1) Pattern
2) Type
3) "Name" (of a generic parameter), which would be used as a hint for inlay parameters and also as a placeholder for unknown types in normal parameter hints?

ortem:
@Kobzol Do you mean such hints for enum variants? Seems like only tests about enum variants are failed in `RsInlayParameterHintsProviderTest`.

Kobzol:
Well, it should also be the same for tuple structs I guess, but yeah, the enum variants tests are failed, because they use the `CallInfo` in a different way than normal hints.

ortem:
OK, this really should be fixed somehow.

Kobzol:
I changed how `CallInfo` works. Now inlay hints get the same behaviour as before and the parameter provider should also work as intended. Also now if there is no pattern for the parameter (happens for enum variants or tuple structs), the pattern will not be rendered at all (not even as `_`), it just produced visual clutter.

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
```diff
@@ -5,44 +5,105 @@
...
+            return CallInfo(buildParameters(ty.paramTypes, parameters))
+        }
+
+        private fun getFunctionLikeParameters(element: RsElement): List<Pair<String?, RsTypeReference?>>? {
+            return when {
+                element is RsEnumVariant -> element.positionalFields.map { null to it.typeReference }
+                element is RsStructItem && element.isTupleStruct ->
+                    element.tupleFields?.tupleFieldDeclList?.map { null to it.typeReference }
+                element is RsPatBinding -> {
+                    val decl = element.parentOfType<RsLetDecl>() ?: return null
```
ortem, line 64: [minor]
```suggestion
                    val decl = element.ancestorStrict<RsLetDecl>() ?: return null
```

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
```diff
@@ -5,44 +5,105 @@
...
+                    val lambda = decl.expr as? RsLambdaExpr ?: return null
+                    lambda.valueParameters.map { (it.patText ?: "_") to it.typeReference }
+                }
+                else -> null
+            }
         }
 
         fun resolve(methodCall: RsMethodCall): CallInfo? {
-            return (methodCall.reference.resolve() as? RsFunction)?.let { CallInfo(it) }
+            val function = (methodCall.reference.resolve() as? RsFunction) ?: return null
```
ortem, line 73: [minor] Let's move this method up to group two public `resolve` methods together for better readability

src/main/kotlin/org/rust/ide/utils/CallInfo.kt
```diff
@@ -5,44 +5,105 @@
...
+            val type = methodCall.inference?.getResolvedMethodType(methodCall) ?: return null
+            return buildFunctionParameters(function, type)
+        }
+
+        private fun buildFunctionParameters(function: RsFunction, ty: TyFunction): CallInfo {
+            val types = let {
```
ortem, line 79: [minor] Executing a block of statements is usually implemented using `run`
```suggestion
            val types = run {
```

ortem (approved):
LGTM except for minor comments. @Kobzol Thank you!

Kobzol:
bors r=ortem

Kobzol:
bors retry

PR #6667, opened 2021-01-13
CI: set milestone properly from CI

Undin:
#6538 changed event when milestone workflow triggered. Most important part of this change is that now we use base commit of each PR to determine milestone.
It allows us to launch the workflow for any PR (even from forks) but at the same time it works incorrectly because now the workflow look on old version of `gradle.properties` files to get proper milestone version.

These changes get actual content of `gradle.properties` via downloading it via http instead of using potentially outdated local version


Undin:
@lancelote I want to merge it ASAP not to set a lot of wrong milestone numbers - it's rather hard to fix them manually.
But a review is still welcome

Issue #6524, opened 2020-12-17
The "continue" key is not prompted


kyssion:
## Environment

* **IntelliJ Rust plugin version:** 0.3.137.3543-2 
* **Rust toolchain version:** 1.48
* **IDE name and version:** clion
* **Operating system:** linux mint 20.04

## Problem description

The "continue" key is not prompted

![image](https://user-images.githubusercontent.com/12844517/102434637-1f341380-4050-11eb-9bfd-b50e50adb362.png)

In this scene he should prompt the "continue" keyword, but now it does not。

This problem not only exists with the "continue" keyword, but also with the "break" keyword

## Steps to reproduce

code ：
```rust
pub fn test1() {
    println!("Guess the number!");
    let secret_number = rand::thread_rng().gen_range(1, 101);
    loop {
        println!("Please input your guess.");

        let mut guess = String::new();

        io::stdin()
            .read_line(&mut guess)
            .expect("Failed to read line");

        let guess: u32 = match guess.trim().parse() {
            Ok(num) => num,
            Err(_) => continu,
        };

        println!("You guessed: {}", guess);

        match guess.cmp(&secret_number) {
            Ordering::Less => println!("Too small!"),
            Ordering::Greater => println!("Too big!"),
            Ordering::Equal => {
                println!("You win!");
                break;
            }
        }
    }
}
```

PR #6673, opened 2021-01-14
COMP: complete continue and break in expression context inside loops

Kobzol:
Both `continue` and `break` are expressions, so they should be completed in expression contexts. I'm not too sure about the "expression location" pattern that I have used, but it seems to work.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6524

changelog: Complete `continue` and `break` statements inside expression contexts in loops.


mchernyavsky (approved):
Thanks!

bors r+

Undin:
bors r=mchernyavsky

PR #6675, opened 2021-01-15
IDE: basic implementation of SSR (Struct, ConstantItem, MetaItem, Types)

Kobzol:
I noticed the SSR feature in IntelliJ and tried to take a shot at implementing a (very basic) version of it for Rust. I took heavy inspiration from the Kotlin SSR plugin (https://github.com/JetBrains/intellij-structural-search-for-kotlin).

Firstly I wanted to ask if you think that it makes sense and if I should continue implementing it :) @Undin

changelog: Add support for structural-search-and-replace.


matklad:
FWIW, I feel we should have this feature at the current maturity level. It's not something one is expected to use daily, but:

* when you *do* use it, it saves a boatload of time
* it's such a cool feature to brag on twitter :)

afetisov:
Could you rebase it on master?

Kobzol:
Ok now it's rebased and can be actually compiled :D And the tests should also pass.

mati865:
Failed for 223 platform with: `e: /home/runner/work/intellij-rust/intellij-rust/src/main/kotlin/org/rust/ide/ssr/RsStructuralSearchProvider.kt: (27, 5): 'getLexicalNodesFilter' overrides nothing`.

yopox:
Hi! I will work on the SSR implementation one day per week.
The current plan is to release a first incomplete version when struct declaration matching and replacing (yet to be done) are good enough, then gradually complete SSR coverage.

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    override fun visitMetaItemArgs(o: RsMetaItemArgs) {
+        val metaItemArgs = getElement<RsMetaItemArgs>() ?: return
+        myMatchingVisitor.result = myMatchingVisitor.matchInAnyOrder(o.metaItemList, metaItemArgs.metaItemList)
+    }
+
+    override fun visitPathType(o: RsPathType) {
+        val path = getElement<RsPathType>() ?: return
+
+        // Match the last path component
```
dima74, line 202: What is motivation for matching paths this way? Personally I think path pattern should match path if they both resolve to the same item. Note that it can be done without `ImportCandidatesCollector` like this:
```kotlin
private fun resolvesToSameItem(templatePath: RsPath, treePath: RsPath): Boolean {
    val path = RsCodeFragmentFactory(treePath.project)
        .createPath(templatePath.text, treePath, treePath.pathParsingMode, treePath.allowedNamespaces()) ?: return false
    return path.reference?.resolve() == treePath.reference?.resolve()
}
```

---

Here are a few examples to think of and add tests:

1. Path pattern is fully-qualified

Pattern:
```rust
const $name$: crate::mod1::Foo = 0;
```
Code:
```rust
const x1: mod1::Foo = 0;
const x2: crate::mod1::Foo = 0;
mod inner {
    use crate::mod1::Foo;
    const x3: Foo = 0;
}
mod mod1 {
    pub struct Foo;
}
```

2. Type aliases

Pattern:
```rust
const $name$: i32 = 0;
```
Code:
```rust
const x: T = 0;
type T = i32;
```

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
```diff
@@ -0,0 +1,258 @@
...
+package org.rust.ide.ssr
+
+import com.intellij.psi.PsiElement
+import com.intellij.structuralsearch.impl.matcher.GlobalMatchingVisitor
+import com.intellij.structuralsearch.impl.matcher.handlers.SubstitutionHandler
+import org.rust.ide.utils.import.ImportCandidatesCollector
+import org.rust.ide.utils.import.ImportContext
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.ext.stringValue
```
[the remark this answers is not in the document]
yopox: The first 3 are used, last one is unused indeed but IDEA was showing it as used for some reason :(

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    override fun visitConstParameter(o: RsConstParameter) {
+        val const = getElement<RsConstParameter>() ?: return
+        myMatchingVisitor.result = match(o.typeReference, const.typeReference)
+            && matchIdentifier(o.identifier, const.identifier)
+    }
+
+    override fun visitConstant(o: RsConstant) {
+        val constant = getElement<RsConstant>() ?: return
+        myMatchingVisitor.result = matchIdentifier(o.identifier, constant.identifier) &&
```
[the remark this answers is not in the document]
yopox: For now I have added the underscore match for constants. When we match the other constructs later I will try not to forget the underscore case 👍

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    override fun visitTypeParameterList(o: RsTypeParameterList) {
+        val parameters = getElement<RsTypeParameterList>() ?: return
+        myMatchingVisitor.result =
+            myMatchingVisitor.matchSequentially(o.typeParameterList, parameters.typeParameterList) &&
+                myMatchingVisitor.matchSequentially(o.lifetimeParameterList, parameters.lifetimeParameterList) &&
+                myMatchingVisitor.matchSequentially(o.constParameterList, parameters.constParameterList)
+    }
+
+    override fun visitWhereClause(o: RsWhereClause) {
```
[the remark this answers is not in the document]
yopox: I think it is ok not to match equivalent constructs and wait for a feature request.
In a custom inspection we can add another template to match both constructs.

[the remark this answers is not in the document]
Undin: I also think we should not merge such cases (at least by default) because SSR is about syntax, not about semantics.
For example, you may want to have a custom inspection to annotate all usages of `struct Foo1<T: Sized>(T);` style and a quick-fix to convert them to `struct Foo2<T>(T) where T: Sized;` style

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
Undin: > I think path pattern should match path if they both resolve to the same item

Do we really want it? As I understand the concept of SSR, it's a way how to write patterns similar to regular expressions but for wider scope of languages (not only regular languages). I.e. it should [mostly] use syntax rules. @yopox am I right? If so, I'd like to keep Rust declarative macros model in mind during matching

But name resolution is a semantic rule.
Two more points not to use name resolution:
- it may cause performance issues because of more work
- it doesn't allow creating syntax only rules. Of course, otherwise, it won't be possible to create semantic rules but not sure that we really need them since they may be too complex for users

Undin: Maybe semantic rules should be expressed as modifiers

afetisov: I would definitely want it to be able to match different expressions which resolve to the same one. There are too many ways to write equivalent expressions in Rust. A method could be called as `foo.bar()`, `(&foo).bar()`, `Foo::bar(&foo)`, `Trait::bar(&foo)`, `path::Trait::bar(&foo)`, `a::b::ReExport::bar(&foo)` etc. If SSR is able to do only literal matches, it would be useless.  Particularly for SSR-based lints.

On the other hand, sometimes the literal tokens are exactly what I want to match, so there should be an option to do either.

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
```diff
@@ -0,0 +1,258 @@
...
+            match(o.metaItemArgs, metaItem.metaItemArgs) &&
+            match(o.path, metaItem.path)
+    }
+
+    override fun visitMetaItemArgs(o: RsMetaItemArgs) {
+        val metaItemArgs = getElement<RsMetaItemArgs>() ?: return
+        myMatchingVisitor.result = myMatchingVisitor.matchInAnyOrder(o.metaItemList, metaItemArgs.metaItemList)
+    }
+
+    override fun visitPathType(o: RsPathType) {
```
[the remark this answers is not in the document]
dima74: Why treat types specially here? Probably it will be unintuitive that paths in types and other positions are matched differently. E.g. consider this example with `RsPathType` and `RsPathExpr`:
```rust
struct Struct(i32);

let x: Struct = Struct(1);
       ~~~~~~   ~~~~~~
       type     pathExpr
```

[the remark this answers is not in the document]
yopox: I moved the custom logic to `visitPath`, it can indeed apply to all paths.

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
yopox: @dima74 We can not resolve the path from the template to an item :(

Consider this template: `struct $S$(bevy::$x$::HashMap)`, or simply `struct $S1$(bevy::utils::$S2$)`, to get these matching we have to go through the matching visitor machinery.

A working approach I have tried is:

1. Matching the last part of the path
2. Check if the rest is matching
3. If 1 is OK but not 2:
    - Collect import paths for the code path
    - Build `RsPath` with the fully qualified import path
    - Check if any is matching

We must use resolve because it is absurd not to match `crate::foo::Foo` with `Foo` in the code (where it is the same reference). It will look if we can have a language specific option in the SSR dialog to disable the use of resolve. If not we will be able to do `crate::foo::$X$` with "Don't resolve paths" modifier on `$X$`.

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
```diff
@@ -0,0 +1,245 @@
...
+        matchingVisitor.result = match(o.path, path.path)
+    }
+
+    override fun visitPath(o: RsPath) {
+        val path = getElement<RsPath>() ?: return
+
+        matchingVisitor.result = matchIdentifier(o.identifier, path.identifier) &&
+            match(o.typeArgumentList, path.typeArgumentList) &&
+            match(o.valueParameterList, path.valueParameterList) &&
+            match(o.path, path.path)
```
dima74, line 216: Let's also match `o.coloncolon` to distinguish between `foo::bar` and `::foo::bar`

dima74 (approved):
LGTM
Could you please squash to small number of commits (1-5). Separate commits were super useful for review so please keep using them in future PR, and before merging we usually squash commits

src/main/kotlin/org/rust/ide/ssr/RsMatchingVisitor.kt
yopox: This causes problems (`Foo` doesn't match `x::Foo` anymore), let's postpone and include this change in the revamp of path matching.

neonaot:
Hi all!
Since a lot of people took part in this functionality, I write here and my remarks are open to discussions :) 

1. Cosmetic: All other languages are named like `%languagename%`, but Rust named `%languagename% files`
image

2. More urgent: Can we add some label/notifications which object types are supported? it's very sad when you write something you think is correct, but it's not found in project files, and you think you are stupid or it's a bug or something else... (real story, need to reed more carefully 😢)  
image

3. Targets:
For Java, I specified that target is`$a2$` and it shows `my_int2` as I expected:
image

For Rust, I did the same for consts and got `const A` instead of `const A2` in the left window and it highlights the whole match in the right window and editor: 
image

vlad20012:
> More urgent: Can we add some label/notifications which object types are supported? it's very sad when you write something you think is correct, but it's not found in project files, and you think you are stupid or it's a bug or something else... (real story, need to read more carefully cry)

Perhaps we should hide the rust SSR under an experimental feature or registry flag until it is ready to search most of PSI parts

yopox:
@neonaot Thank you!

1. I have also spotted the issue and fixed it in the platform last week. Language description was used instead of the language name. (we get the fix in 2023.2)
2. It is possible to show an error in the structural search dialog if the template contains an unsupported element, good idea.
3. This issue was my main focus this week :)

Hiding SSR behind a flag for now seems reasonable. Should issues be opened to address these points? I am not familiar with the processes here.

Issue #6677, opened 2021-01-15
`Extract function` refactoring extracts more statements then selected


vlad20012:
## Environment

* **IntelliJ Rust plugin version:** 0.3.140.3605-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-unknown-linux-gnu
* **IDE name and version:** IntelliJ IDEA 2020.3.1 Ultimate Edition (IU-203.6682.168)
* **Operating system:** Linux 5.8.0-36-generic
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
When I extract multiple statements from a code block (not a function body directly), the plugin extracts the entire code block instead.

## Steps to reproduce
```rust
fn main() {
    {
        let a = 1;
        let b = 2;

        let c = a + b;
    };
}
```

1. Select `let a = 1; let b = 2;` statements
2. Perform `Extract function` refactoring (`Ctrl` + `Alt` + `M`)
3. The plugin extracts `let c = a + b;` and replaces the whole code block

PR #6678, opened 2021-01-15
REF: correctly extract selected statements inside blocks into a function

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6677

changelog: Fix extracting selected statements into a function inside blocks.


vlad20012:
Wow. It was fast. Thanks!

bors r+

PR #6684, opened 2021-01-16
REF: correctly extract selected statements inside exprs into a function

Kobzol:
Continuation of https://github.com/intellij-rust/intellij-rust/pull/6678.

I could special case it for `RsBlockExpr`/`RsExprStmt`, since blocks are probably the only place where this could happen, but I realized that the actually important thing is the containment within an expr/stmt. So I removed the explicit check for `RsBlockExpr`.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6683


vlad20012:
:+1: Thanks!

bors r+

PR #6685, opened 2021-01-16
RES: prefer trait methods before private inherent methods

vlad20012:
Improves the #5661 fix: now resolve it to visible trait impl instead of private inherent impl:

```rust
use foo::{Foo, Trait};

mod foo {
    pub struct Foo;
    impl Foo {
        // Private
        fn get(&self) { println!("struct"); }
    }

    pub trait Trait {
        fn get(&self);
    }
    impl Trait for Foo {
        fn get(&self) { println!("trait"); }
    }    //X
}

fn main() {
    let f = foo::Foo;
    f.get();
    //^
}
```

c.c. @Kobzol

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
@@ -670,7 +670,8 @@ class RsTypeInferenceWalker(
         return methodType.retType
     }
 
-    private fun <T : AssocItemScopeEntryBase<E>, E> filterAssocItems(variants: List<T>, context: RsElement): List<T> {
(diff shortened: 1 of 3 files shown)


Kobzol (approved):
Even better :) Maybe we could keep the old test with
```rust
impl Foo {
    pub fn get(&self) { println("struct"); }
}
```
?

vlad20012:
@Kobzol Hmm yep, Let's keep it

vlad20012:
@Kobzol Oh, actually there isn't a multiresolve because inherent impl wins

Kobzol:
Oh, ok, even better :)

vlad20012:
@Undin the PR awaiting the regression testing repairs :)

Undin:
@vlad20012 as we discussed in person, `regression-2` is only about stdlib and yeah, it's broken currently for some reason. All other regression tests work as expected so looks like you don't need to wait

Kobzol:
bors retry

PR #6688, opened 2021-01-17
TY&ANN: fix false-positive E0277

### LINKED ISSUE #6331: [Bug] False positive for [E0277]
## Environment

Plugin Version: 0.3.133.3426-202
IDE:  CLion Commercial 2020.2.4 
Rust Toolchain: 1.47.0 stable-x86_64-unknown-linux-gnu
OS: Fedora 33 KDE

## Problem description / Steps to reproduce
False positive for [E0277] 
The code below, which can be compiled successfully, demonstrates the issue:
```rust
trait Foo {}

trait Bar: Foo {}

trait Baz: Foo {}

trait Bif: Baz + Bar {}

impl<T> Baz for T where T: Foo {}

impl<T> Bif for T where T: Bar {}
```
In this example `Bif` will be incorrectly highlighted by CLion to say [E0277] because the trait bound Baz is not satisfied.
This is incorrect. 
The reason being is as follows: 
- In `impl Bif` `T` is already bound by `Bar`, and any implementation of `Bar` must implement `Foo`
- `Baz` is is automatically implemented for any `T` where `T` implements `Foo`

Because of these two things it is guaranteed that when a type `T` is bound by the trait `Bar` it must have an implementation of `Baz`and thus the trait bound of `Baz` is satisfied.

The problem can be 'solved' by changing the last line to this:
```
impl<T> Bif for T where T: Bar + Baz {}
```
however this is not needed for rustc to compile the code.

vlad20012:
Fixes #6331

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -655,7 +655,7 @@ class ImplLookup(
                     .filter { ctx.probe { ctx.combineBoundElements(it, ref.trait) } }
                     .map { SelectionCandidate.TypeParameter(it) }
                     .forEach(::add)
-                if (ref.selfTy is TyTypeParameter) return@buildList
+
                 if (ref.selfTy is TyProjection) {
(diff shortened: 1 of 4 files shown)


vlad20012:
The regression is OK - a type aws previously not inferred. Now it's inferred and triggers an unrelated false-positive.

bors r+

PR #6690, opened 2021-01-18
Initialize crates local index service on startup

avrong:
After opening a project, `CratesLocalIndexStartupActivity` tries to get an instance of the service. This initializes the service's `Cargo.toml` watcher and PersistentHashMap before performing any inspection or completion.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
@@ -1,0 +1,28 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.startup.StartupActivity
+import com.intellij.openapiext.isUnitTestMode
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
+
+/**
+ * [CratesLocalIndexService] initializer.
+ * Only tries to get service instance, without any updates to its crates index.
+ */
+class CratesLocalIndexStartupActivity : StartupActivity.Background {
+    override fun runActivity(project: Project) {
+        if (isUnitTestMode) return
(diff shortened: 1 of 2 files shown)


toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.startup.StartupActivity
+import com.intellij.openapiext.isUnitTestMode
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
+
+class CratesLocalIndexStartupActivity : StartupActivity {
+    override fun runActivity(project: Project) {
+        if (!isUnitTestMode && isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
+            CratesLocalIndexService.getInstance()
```
ortem, line 17: AFAIU this call only initialize but does not actually **build** the index. To ultimately build it, `CratesLocalIndexService#updateIfNeeded` should be invoked.

Also, if something heavyweight like `CratesLocalIndexService#updateIfNeeded` is called, we may consider `StartupActivity.Background` instead of `StartupActivity`

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.startup.StartupActivity
+import com.intellij.openapiext.isUnitTestMode
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
+
+class CratesLocalIndexStartupActivity : StartupActivity {
+    override fun runActivity(project: Project) {
+        if (!isUnitTestMode && isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
```
ortem, line 16: Note that currently `CratesLocalIndexService.getInstance()` is invoked even if the user opens **any** IDEA project, potentially not related to Rust anyhow. I suppose we should invoke it only if `project.cargoProjects.hasAtLeastOneValidProject` is true

ortem (changes requested):

ortem (approved):
bors r+

Issue #6670, opened 2021-01-14
Not suggesting assert_eq!


krojew:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3572-203
* **Rust toolchain version:** 1.49.0 (e1884a8e3 2020-12-29) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2020.3.1 (CL-203.6682.181)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

Since the new engine was introduced, the `assert_eq!` macro is no longer suggested. 

## Steps to reproduce

Try to type `assert_eq!` and notice it's not suggested.

dima74:
@krojew thanks for the report!

Do I understand correctly that with new name resolution engine:

* Completion doesn't work when cursor is in the middle of macro path:

![image](https://user-images.githubusercontent.com/6505554/104947294-9a307700-59cc-11eb-9edb-ad085d5143b9.png)

* Completion works when cursor is at the end of macro path:

![image](https://user-images.githubusercontent.com/6505554/104947367-b92f0900-59cc-11eb-897c-ef6c5d40b3c6.png)

?

krojew:
To be honest - this is very non-deterministic. Sometimes it works when writing, but mostly, the only suggestions are:

![image](https://user-images.githubusercontent.com/1268573/104947889-b4faef80-59bc-11eb-858a-f9fb07c583a9.png)

I've enabled the machine learning suggestions, if that makes a difference.

dima74:
@krojew could you please full code from the screenshot?

krojew:
I can't, since it's proprietary. I can tell you that it happens in all test modules. It may be influenced by the float_eq import: `use float_eq::*;`, since those seem to be the only suggestions.

krojew:
Just made an observation that suggestions do not appear when typing, but they do appear when pressing the completion shortcut.

dima74:
@krojew could you please install [nightly build](https://www.jetbrains.com/rust/download/) and check whether issue was fixed

krojew:
Well, now the opposite happens - assert_* macros are immediately suggested, but the ones which were there before disappeared and require manually pressing ctrl+space.

dima74:
> but the ones which were there before disappeared and require manually pressing ctrl+space.

[...]

PR #6692, opened 2021-01-18
RES: Fix completion of unqualified macro calls in new resolve

dima74:
Fixes part of #6670, when cursor is located in the middle of macro path:
```
asse/*cursor*/rt_!();
```

changelog: Fix completion of unqualified macro calls in new resolve in some cases

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
@@ -112,24 +112,19 @@ fun processMacros(
     }
     if (runBeforeResolve != null && runBeforeResolve()) return true
 
-    return modData.processMacros(scope, macroPath, processor, defMap, project)
+    return modData.processMacros(macroPath, processor, defMap, project)
 }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
```diff
@@ -112,24 +112,19 @@ fun processMacros(
...
     processor: RsResolveProcessor,
     defMap: CrateDefMap,
     project: Project,
 ): Boolean {
-    val isCompletion = processor.name == null
-
     val isQualified = macroPath == null || macroPath is RsPath && macroPath.qualifier != null
     if (!isQualified) {
         check(macroPath != null)
-        val exactScopeVisibilityFilter = if (isCompletion) { mod: RsMod -> mod == scope } else { _ -> true }
```
dima74, line 132: Error was here: `mod == scope`, instead we should use `mod == scope.originalElement`. But turns out that `mod` and `scope` are equal anyway when completing unqualified macro calls, so we can remove this visibility filter

PR #6702, opened 2021-01-20
RUN: Allow to build different projects simultaneously

mchernyavsky:
Currently, if you are using the new Build View and try to build a second project while building the first project, the second build will hang with the "Waiting for another build to finish" message.

changelog: Allow building different projects simultaneously

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildTaskRunner.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildTaskRunner.kt
@@ -40,7 +40,6 @@ import java.util.concurrent.*
 private val LOG: Logger = Logger.getInstance(CargoBuildTaskRunner::class.java)
 
 class CargoBuildTaskRunner : ProjectTaskRunner() {
-    private val buildSessionsQueue: BackgroundTaskQueue = BackgroundTaskQueue(null, "Building...")
 
     override fun run(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildTaskRunner.kt
```diff
@@ -70,7 +70,9 @@ class CargoBuildTaskRunner : ProjectTaskRunner() {
             WaitingTask(project, waitingIndicator, queuedTask.executionStarted).queue()
         }
 
-        buildSessionsQueue.run(queuedTask, null, EmptyProgressIndicator())
+        project.service<CargoBuildSessionsQueueManager>()
```
Undin, line 73: nit: I usually prefer creating `getInstance(project: Project)` static method in service class because:
- it's a common pattern in the platform
- it's clearer how to get the instance of this class (at least for me)

Undin (approved):
Works for me

src/main/kotlin/org/rust/cargo/runconfig/buildtool/CargoBuildTaskRunner.kt
mchernyavsky: Fixed.

mchernyavsky (commented):
bors r=Undin

PR #6704, opened 2021-01-21
Support for get cargo path in `CARGO_HOME` env

Stzx:
This is helpful if you do not use the default installation path locally, but need to do tests.

### DIFF
--- a/src/test/kotlin/org/rust/cargo/RustupTestFixture.kt
+++ b/src/test/kotlin/org/rust/cargo/RustupTestFixture.kt
@@ -45,7 +45,9 @@ open class RustupTestFixture(
     }
 
     private fun addCargoHomeToAllowedRoots() {
-        val cargoHome = Paths.get(FileUtil.expandUserHome("~/.cargo"))
+        val cargoHome = Paths.get(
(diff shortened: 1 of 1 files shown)


Stzx:
@Kobzol I would like to ask if this PR will be merged? I think the change is minimal.

vlad20012:
What was the purpose of this change? As far as I know, `addCargoHomeToAllowedRoots` (now renamed to `setUpAllowedRoots`) is only needed on CI. It does not set up the cargo location for tests.

Stzx:
> What was the purpose of this change?

If when the toolchain is not installed in the default location `~/.cargo` (e.g. the toolchain is installed via `scoop`.) . If you want to run the relevant test locally, the test will complain that the toolchain does not exist.

> As far as I know, `addCargoHomeToAllowedRoots` (now renamed to `setUpAllowedRoots`) is only needed on CI.

Yes, it's been a bit long. This PR is no longer suitable for new file.

> It does not set up the cargo location for tests.

I think it should be valid at the time I send this PR.

vlad20012:
Let's close it for now. I'm not familiar with toolchain discovery in tests, unfortunately. If you'll fix this again (and you will be sure the fix works for you), please ping me there.

PR #6705, opened 2021-01-21
L10N: Support partial display of content as `zh`

Stzx:
Problem:
* Probably need to force the `Locale` of the test environment to be `US` (Otherwise the presence of the corresponding translation file will cause the current part of the test to fail.
* Whether the user should be allowed to manually select the language. (Shouldn't actually be implemented in the plugin.


Stzx:
Please let me know if there is an error.

Undin (commented):
Thanks for these improvements
But we have our own workflow to localize IDE/plugins via language pack plugins like https://plugins.jetbrains.com/plugin/13710-chinese-simplified-language-pack----

PR #6715, opened 2021-01-24
RUN: Suggest to install wasm32 target in wasm-pack configuration

### LINKED ISSUE #6714: Building of wasm-pack configuration fails if wasm32 target is not installed
image

avrong:
Fixes #6714

In wasm-pack configuration before launch Build step checks if `wasm32-unknown-unknown` target is installed and suggests installing it if It is not.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
@@ -13,9 +13,11 @@ import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.runconfig.wasmpack.WasmPackBuildTaskProvider.Companion.WASM_TARGET
 import org.rust.cargo.toolchain.RsToolchain
 import org.rust.cargo.util.DownloadResult
 import org.rust.ide.actions.InstallComponentAction
+import org.rust.ide.actions.InstallTargetAction
 import org.rust.ide.notifications.showBalloon
 import org.rust.openapiext.execute
 import org.rust.openapiext.fullyRefreshDirectory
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
```diff
@@ -130,5 +173,24 @@ class Rustup(toolchain: RsToolchain, private val projectDirectory: Path) : RsToo
 
             return needInstall
         }
+
+        private fun checkNeedInstallTarget(
```
ortem, line 177: Let's add `@Suppress("SameParameterValue")` above this method

src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
```diff
@@ -43,12 +45,28 @@ class Rustup(toolchain: RsToolchain, private val projectDirectory: Path) : RsToo
...
+        }
+    }
+
     fun listComponents(): List<Component> =
         createBaseCommandLine(
             "component", "list",
             workingDirectory = projectDirectory
         ).execute()?.stdoutLines?.map { Component.from(it) }.orEmpty()
 
+    fun listTargets(): List<Target> =
```
ortem, line 64:
```suggestion
    private fun listTargets(): List<Target> =
```

src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
```diff
@@ -43,12 +45,28 @@ class Rustup(toolchain: RsToolchain, private val projectDirectory: Path) : RsToo
...
+        companion object {
+            fun from(line: String): Target {
+                val name = line.substringBefore(' ')
+                val isInstalled = line.substringAfter(' ') == "(installed)"
+                return Target(name, isInstalled)
+            }
+        }
+    }
+
     fun listComponents(): List<Component> =
```
ortem, line 58:
```suggestion
    private fun listComponents(): List<Component> =
```

src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
```diff
@@ -43,12 +45,28 @@ class Rustup(toolchain: RsToolchain, private val projectDirectory: Path) : RsToo
         }
     }
 
+    data class Target(val name: String, val isInstalled: Boolean) {
+        companion object {
+            fun from(line: String): Target {
+                val name = line.substringBefore(' ')
+                val isInstalled = line.substringAfter(' ') == "(installed)"
```
ortem, line 52: Hmm, I'm not sure why we check `"(default)"` case in `Component#from`, and I haven't managed to simulate such output from `rustup component list` and `rustup target list` (but such output is definetely possible for `rustup toolchain list`).

@mchernyavsky Could you please explain is it really necessary?

src/main/kotlin/org/rust/ide/actions/InstallTargetAction.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.intellij.openapi.project.DumbAwareAction
+import org.rust.cargo.project.settings.toolchain
+import org.rust.cargo.toolchain.tools.rustup
+import org.rust.cargo.util.DownloadResult
+import org.rust.ide.notifications.showBalloon
+import java.nio.file.Path
+
+class InstallTargetAction(
+    val projectDirectory: Path,
+    val targetName: String
```
ortem, line 22:
```suggestion
    private val projectDirectory: Path,
    private val targetName: String
```

ortem (changes requested):

src/main/kotlin/org/rust/cargo/toolchain/tools/Rustup.kt
mchernyavsky: I think this is needed on older rustup versions

(see the output from this issue - https://github.com/rust-lang/rustup/issues/1544)

ortem: @mchernyavsky Thank you! So we added `"(default)"` case for targets as well just in case

ortem (approved):
LGTM
bors r+

PR #6716, opened 2021-01-25
INT: Support async function by create function intention

### LINKED ISSUE #6712: Create function intention should create async function if necessary
## Environment

* **IntelliJ Rust plugin version:** 0.3.139.3615-203
* **Rust toolchain version:** 1.51.0-nightly (4253153db 2021-01-17) x86_64-unknown-linux-gnu
* **Operating system:** Linux 5.10.9-arch1-1

## Problem description
"Create function intention" should create async function if we call `.await` directly on function call result:

Peek 2021-01-24 18-26

## Steps to reproduce
```rust
async fn foo() {
    /*cursor*/bar().await;
}
```

Stzx:
Fixed #6712

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
@@ -10,6 +10,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
 import com.intellij.psi.util.parentOfType
+import com.intellij.psi.util.parentsOfType
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.utils.GenericConstraints
@@ -37,16 +38,19 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
 
     sealed class Context(val name: String, val callElement: PsiElement) {
         abstract val visibility: String
+        open val isAsync: Boolean = callElement.isEdition2018
         abstract val arguments: RsValueArgumentList
         abstract val returnType: Ty?
         open val implItem: RsImplItem? = null
 
-        open class Function(val callExpr: RsCallExpr, name: String, val module: RsMod) : Context(name, callExpr) {
+        open class Function(callExpr: RsCallExpr, name: String, val module: RsMod) : Context(name, callExpr) {
             override val visibility: String = when {
                 callExpr.containingCrate != module.containingCrate -> "pub "
                 callExpr.containingMod != module -> "pub(crate) "
                 else -> ""
             }
+            override val isAsync: Boolean = super.isAsync
+                && (callExpr.parent as? RsDotExpr)?.fieldLookup?.isAsync == true
             override val arguments: RsValueArgumentList = callExpr.valueArgumentList
             override val returnType: Ty? = callExpr.expectedType
         }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -37,16 +38,25 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
...
 
-        open class Function(val callExpr: RsCallExpr, name: String, val module: RsMod) : Context(name, callExpr) {
+        open class Function(callExpr: RsCallExpr, name: String, val module: RsMod) : Context(name, callExpr) {
             override val visibility: String = when {
                 callExpr.containingCrate != module.containingCrate -> "pub "
                 callExpr.containingMod != module -> "pub(crate) "
                 else -> ""
             }
+            override val isAsync: Boolean = super.isAsync
+                && callExpr.parentOfType<RsDotExpr>()?.fieldLookup?.isAsync == true
```
Kobzol, line 59: This check should be more precise, to only check the direct parent.
Otherwise e.g. this:
```rust
async fn foo() {
    baz(/*caret*/bar()).await;
}
async fn baz(a: u32) {}
```
will produce
```rust
async fn foo() {
    baz(bar()).await;
}
async fn bar() -> u32 {
    unimplemented!()
}
async fn baz(a: u32) {}
```
Which is probably not what the user wanted.

Kobzol (changes requested):

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -73,6 +83,8 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
                         else -> "pub(crate)"
                     }
                 }
+            override val isAsync: Boolean = super.isAsync
+                && methodCall.parentDotExpr.parentOfType<RsDotExpr>()?.fieldLookup?.isAsync == true
```
Kobzol, line 87: Same things applies to methods:
```rust
struct S;

async fn foo(s: S) {
    baz(s./*caret*/bar()).await;
}
async fn baz(a: u32) {}
```

Kobzol (changes requested):

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
Stzx: Sorry, I forgot this.

Stzx:
@Kobzol This is a question:
```rust
struct S;

impl S {
    async fn foo(&self) {
        self.foz(self./*caret*/bar(1, 2).await);
    }

    fn foz(&self, p0: u32) {
        unimplemented!()
    }
}
```
Actual:
```rust
struct S;

impl S {
    async fn foo(&self) {
        self.foz(self.bar(1, 2).await);
    }

    fn foz(&self, p0: u32) {
        unimplemented!()
    }

    async fn bar(&self, p0: u32, p1: u32) {
        unimplemented!()
    }
}
```

Kobzol:
Hmm, maybe it could ask for the expected type of the parent `await` if it is present, but I think that it's okay.

Kobzol (approved):

src/main/kotlin/org/rust/ide/intentions/CreateFunctionIntention.kt
```diff
@@ -37,16 +38,25 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
 
     sealed class Context(val name: String, val callElement: PsiElement) {
         abstract val visibility: String
+        open val isAsync: Boolean = callElement.parentsOfType<RsBlock>().firstOrNull().let { block ->
```
dima74, line 41: Do we really want to check if containing function is `async`? Code like this:
```rust
fn foo() {
    bar().await;
}
```
is invalid, but if user already has `await`, maybe user also will make function `foo` be `async` later?

Kobzol: FWIW I was thinking the same, but I didn't feel strongly either way. If there is no check for async, there should probably at least be a check for edition 2018.

Stzx: Well, let me modify it to only detect edition 2018.

dima74 (approved):
Thanks!
bors r+

Issue #6454, opened 2020-11-29
False positive for 'Type arguments must be declared prior to const arguments'


zeroeightysix:
Consider the following code:

```rust
#![feature(const_generics)]

fn foo<const C: usize, A>() {}

fn main() {
    foo::<5, bool>();
}
```

The plugin will insist that the order of type and const parameters is wrong (at error level), even though the code compiles correctly:
![image](https://user-images.githubusercontent.com/27009727/100552832-eb07d700-3289-11eb-8da8-38fd34556ec5.png)

Using rustc 1.49.0-nightly (ffa2e7ae8 2020-10-24).

PR #6717, opened 2021-01-25
ANN: Take into account const generics in inspections about generic parameters/arguments ordering

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6454.
Relates to https://github.com/intellij-rust/intellij-rust/issues/3985.

changelog: Take into account const generics in inspections about generic parameters/arguments ordering

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
@@ -313,7 +313,7 @@ private enum class TypeArgumentKind(private val elementClass: KClass<*>, val arg
         get() = StringUtil.capitalize(argumentName)
 
     fun canStandAfter(prevArgument: TypeArgumentKind): Boolean =
-        ordinal >= prevArgument.ordinal
+        this !== LIFETIME || prevArgument === LIFETIME
 
     companion object {
         private val VALUES = values()
--- a/src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
@@ -277,14 +277,36 @@ private fun checkTypeParameterList(holder: AnnotationHolder, element: RsTypePara
                     .range(it).create()
             }
     }
-    val lifetimeParams = element.lifetimeParameterList
-    if (lifetimeParams.isEmpty()) return
-    val startOfTypeParams = element.typeParameterList.firstOrNull()?.textOffset ?: return
-    for (e in lifetimeParams) {
-        if (e.textOffset > startOfTypeParams) {
-            holder.newAnnotation(HighlightSeverity.ERROR, "Lifetime parameters must be declared prior to type parameters")
-                .range(e).create()
+
+    var kind = TypeParameterKind.LIFETIME
+    element.forEachChild { child ->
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -277,17 +277,39 @@ private fun checkTypeParameterList(holder: AnnotationHolder, element: RsTypePara
...
+
+    var kind = TypeParameterKind.LIFETIME
+    element.forEachChild { child ->
+        val newKind = TypeParameterKind.forType(child) ?: return@forEachChild
+        if (newKind.canStandAfter(kind)) {
+            kind = newKind
+        } else {
+            val newStateName = newKind.parameterNameCapitalized
+
+            holder.newAnnotation(HighlightSeverity.ERROR, "$newStateName must be declared prior to ${kind.parameterName}")
```
Stzx, line 289: Perhaps similar content from PR can be extracted into the localization file now to ease the work later.

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
```diff
@@ -277,17 +277,39 @@ private fun checkTypeParameterList(holder: AnnotationHolder, element: RsTypePara
...
+    fun canStandAfter(prevArgument: TypeParameterKind): Boolean =
+        this !== LIFETIME || prevArgument === LIFETIME
+
+    companion object {
+        private val VALUES = values()
+        fun forType(seekingElement: PsiElement): TypeParameterKind? =
+            VALUES.find { it.elementClass.isInstance(seekingElement) }
+    }
+}
+
```
Undin, line 312: Do we actually want to have two almost the same enum and two versions of almost the same code to check the order?
Looks like we can merge `TypeParameterKind` and `TypeArgumentKind` enums into one enum with `LIFETIME`, `TYPE` and `CONST` variants and some class with `name` and `kind` fields.
And after that it will be quite simple unify checking methods into single one

Undin (commented):
Consider creating a quick fix for wrong lifetime order and wrong assoc type bindings like compiler suggest
For example, for the following code:
```rust
fn foo<const A: i32, 'a>() {}
```
compiler suggest possible reordering in help section
```
error: lifetime parameters must be declared prior to const parameters
 --> src/main.rs:7:22
  |
7 | fn foo<const A: i32, 'a>() {}
  |       ---------------^^- help: reorder the parameters: lifetimes, then consts and types: `<'a, const A: i32>`
```

src/main/kotlin/org/rust/ide/annotator/RsSyntaxErrorsAnnotator.kt
Undin: Not sure that we should do it in this PR
I suppose that it should be done in separate PRs with extracting all string literals for a particular subsystem into property files and declaring rules not to add new hardcoded literals in code.
But I still don't fully understand how we will support translations in an actual state

LoganDark:
Poke. Const generics are making it into stable and IntelliJ-Rust currently marks completely valid code as an error. The inspection can be disabled with a comment, but that stinks

mchernyavsky:
@Undin fixed.

Undin (approved):
LGTM except minor comments

mchernyavsky (commented):
bors r=Undin

PR #6718, opened 2021-01-25
Extract code related to data-flow analysis into `dfa` package

ortem:
Previously, this code was located in `org.rust.lang.core.types` (which is not very relevant) and `org.rust.lang.core.cfg` (which might be confused with `cfg` attributes) packages. And now there is a separate `org.rust.lang.core.dfa` package.


ortem:
@Kobzol I ultimately agree with you! Lots of code and naming were actually derived from the original  rustc's implementation (which has been partially replaced after non-lexical lifetimes). This really needs to be documented and I try to do that soon.

PR #6719, opened 2021-01-25
Fix passing item name through RsResolveProcessor

dima74:
Please see #5955 for details why `name` is needed

changelog: Fix performance regression in name resolution for large modules


Undin (approved):
👍 
bors r+

PR #6720, opened 2021-01-25
RES: Fix resolve inside cfg-disabled inline mod in new resolve

dima74:
E.g.
```rust
#[cfg(false)]
mod my {
    fn bar() {}
     //X
    fn t() {
        bar();
    } //^
}
```

changelog: Fix resolve inside cfg-disabled inline mod in new resolve


vlad20012 (approved):
bors r+

PR #6721, opened 2021-01-25
RES: Support basic multiresolve of inline modules in new resolve

dima74:
E.g.
```rust
mod m {}
mod m {}
fn main() {
    m::func();
} //^
```
changelog: Support basic multiresolve of inline modules in new resolve


vlad20012 (approved):
:+1: 
bors r+

dima74:
bors retry

PR #6724, opened 2021-01-26
REF: Fix types auto import in change signature dialog with new resolve

dima74:
E.g. try changing function return type to `FooStruct`:
```rust
mod mod1 {
    pub struct FooStruct {}
}

fn foo() {}
```

changelog: Fix auto-import in change signature refactoring dialog when using new resolve


Kobzol:
What's the issue? When I run your code on `master`, no import is offered, since the struct is private (that seems expected to me). If I make it public, the import seems to work correctly.

Kobzol (approved):

dima74:
bors r=Kobzol

dima74:
bors retry

PR #6730, opened 2021-01-27
Support file level cfg attributes in new resolve

dima74:
E.g. now such macro will not be expanded in new resolve (we don't expand cfg-disabled macros):
```rust
#![cfg(false)]
...
some_macro!();
```

Potentially this should slightly speed up [winapi crate](https://github.com/retep998/winapi-rs/blob/2f76bdea3a79817ccfab496fbd1786d5a697387b/src/lib.rs#L6)

changelog: Support file level cfg attributes in new resolve


vlad20012 (approved):
bors r+

Issue #6236, opened 2020-10-10
`Token` macro not found in syn crate


mahkoh:
## Environment

* **IntelliJ Rust plugin version: 0.3.132.3395-202**
* **Rust toolchain version: 1.47.0**
* **IDE name and version: CLion 2020.2.4**
* **Operating system: Arch Linux**

## Problem description

```
[dependencies]
syn = "1"
```

```
use syn::Token;

fn main() {
    Token![where] // Unresolved reference: `Token`
				  // syn::token::Token? Alt+Enter
}
```

Replacing `syn::Token` by `syn::token::Token` leads to a compilation error. Macros must be imported from the crate root:

```
error: cannot find macro `Token` in this scope
 --> src/main.rs:5:5
  |
5 |     Token![where]
  |     ^^^^^
```

The macro not being found is acceptable. What's not acceptable is permanently suggesting I import `syn::token::Token`. As a workaround, it's possible to add `#[allow(unused_imports)] use syn::token::Token;`.

## Steps to reproduce

See above.

Juici:
A workaround for this is to disable the lint, and just rely on the `cargo check` lints to cover unresolved references.

dima74:
Fixed with [new name resolution](https://github.com/intellij-rust/intellij-rust/issues/6217) engine [enabled](https://github.com/intellij-rust/intellij-rust/pull/6378)

sabbyX:
> Fixed with [new name resolution](https://github.com/intellij-rust/intellij-rust/issues/6217) engine [enabled](https://github.com/intellij-rust/intellij-rust/pull/6378)

It's still not fixed, already enabled `name resolution 2.0`

**Plugin version**:  0.3.139.3589-203-nightly
**IDE**: Clion 2020.3.1
**Operation system**: Windows 10
**Rust**: rustc 1.51.0-nightly (2987785df 2020-12-28)

---

![image](https://user-images.githubusercontent.com/61574665/104103559-bb0c1480-52c8-11eb-9273-d7cfb2cd19b4.png)

![image](https://user-images.githubusercontent.com/61574665/104103649-3b327a00-52c9-11eb-8f30-ea4c746d516c.png)

Juici:
It is fixed, the import should be `syn::Token`, I was using it earlier
today.

sabbyX:
> It is fixed, the import should be `syn::Token`, I was using it earlier today.

[...]

PR #6734, opened 2021-01-27
RES: Fix resolve macro calls inside included file when using new resolve

dima74:
E.g.:
```rust
// main.rs
macro_rules! foo {
    () => {};
}
include!("foo.rs");
// foo.rs
foo!();
//^
```

Related: #6236 - [`Token` macro in `syn` crate](https://github.com/dtolnay/syn/blob/133a053fe67d5f360c0ad24370216e0052624dfc/src/token.rs#L882)

changelog: Fix resolve macro calls inside file included by `include!` macro when using new resolve


vlad20012 (approved):
bors r+

PR #6735, opened 2021-01-27
Compile only buildscripts and proc macros when fetching build scripts

vlad20012:
Trying suggestion from https://github.com/rust-lang/cargo/pull/9071#issuecomment-760281548


src/main/kotlin/org/rust/cargo/toolchain/tools/Cargo.kt
```diff
@@ -171,7 +172,14 @@ open class Cargo(toolchain: RsToolchain, useWrapper: Boolean = false)
     ): BuildScriptsInfo? {
         if (!isFeatureEnabled(RsExperiments.EVALUATE_BUILD_SCRIPTS)) return null
         val additionalArgs = listOf("--message-format", "json")
-        val commandLine = CargoCommandLine("check", projectDirectory, additionalArgs)
+        val nativeHelper = RsPathManager.procNativeHelper()
+        val envs = if (nativeHelper != null) {
```
Undin, line 176: It would be great to have the ability to disable using native helper in case of any bug in it.
Otherwise, we can break existing feature (yep, it's experimental but anyway) without any workaround

src/main/kotlin/org/rust/cargo/toolchain/tools/Cargo.kt
```diff
@@ -171,7 +172,14 @@ open class Cargo(toolchain: RsToolchain, useWrapper: Boolean = false)
...
+            EnvironmentVariablesData.create(mapOf(RUSTC_WRAPPER to nativeHelper.toString()), true)
+        } else {
+            EnvironmentVariablesData.DEFAULT
+        }
+
+        val commandLine = CargoCommandLine("check", projectDirectory, additionalArgs, environmentVariables = envs)
```
Undin, line 182: Can we write a test to ensure that this command with our own `RUSTC_WRAPPER` doesn't actually compile target code? For example, check that we don't receive warning/error messages from the compiler, can extract build script info even there is a compilation error in some target code or anything else

src/main/kotlin/org/rust/cargo/toolchain/tools/Cargo.kt
vlad20012: I've added a registry key `org.rust.cargo.evaluate.build.scripts.wrapper` (true by default)

src/main/kotlin/org/rust/cargo/toolchain/tools/Cargo.kt
vlad20012: Added a test with a compilation error

Undin (approved):
LGTM

intellij-rust-native-helper/src/rustc_wrapper.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub struct ExitCode(pub Option<i32>);
+
+pub fn run_rustc_skipping_cargo_checking(
+    rustc_executable: OsString,
+    args: Vec<OsString>,
+) -> io::Result<ExitCode> {
+    let is_cargo_check = args.iter().any(|a| {
+        let a = a.to_string_lossy();
+        a.starts_with("--emit=") && a.contains("metadata") && !a.contains("link")
```
Undin, line 13: Let's also add a comment that explains this condition and add this https://doc.rust-lang.org/rustc/command-line-arguments.html#--emit-specifies-the-types-of-output-files-to-generate url

vlad20012:
bors r=undin

Issue #6292, opened 2020-10-26
Test hierarchy tree is not shown on the client when using "Code with me" plugin


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.134.3437-203-nightly
* **Rust toolchain version:** 1.48.0-nightly-nightly (8876ffc92 2020-10-02) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.1 Snapshot (CL-211.284)
* **Operating system:** macOS 10.15.6

## Problem description

Test hierarchy is not show on the client while using "Code with me" plugin

<img width="2304" alt="Screen Shot 2020-10-21 at 22 44 03" src="https://user-images.githubusercontent.com/2500320/97171971-5ea45900-179f-11eb-9ae7-3c768f2c83aa.png">

- YouTrack ticket https://youtrack.jetbrains.com/issue/CWM-1208
- Similar PyCharm ticket https://youtrack.jetbrains.com/issue/CWM-653

lancelote:
@mchernyavsky Thank you for taking care of it 🙇🏻‍♂️

PR #6738, opened 2021-01-27
RUN: Show test hierarchy tree on the client when using "Code with me" plugin

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6292.

changelog: Show test hierarchy tree on the client when using "Code with me" plugin


Undin (approved):
👍 
bors r+

Issue #3582, opened 2019-03-24
Introduce add parameter intention


SomeoneToIgnore:
Similar to Kotlin and Java intentions, I would like to add a parameter into the function invocation and, if the function is editable, the plugin should be able to add an extra parameter with the correct type into the function definition.

Kotlin intention example:

<img width="1148" alt="image" src="https://user-images.githubusercontent.com/2690773/54885103-5f696080-4e81-11e9-97e5-cb6c3c2da9cd.png">

Issue #6744, opened 2021-01-27
Improve Change Signature refactoring


Kobzol:
Improve the Change Signature refactoring introduced in #5944.

- [ ] Allow specifying default values for parameters.
- [ ] Add support for suggested refactoring (introduced for local variables in https://github.com/intellij-rust/intellij-rust/pull/5913).
- [ ] Add function parameter from function call.
- Support `extern`?
- Allow changing type parameters and `where` clauses?

Undin:
> Sometimes it would be super handy if @intellijrust had a refactoring to add a new type or lifetime parameter to a struct (an make all the necessary changes to surrounding code).
 
From https://twitter.com/wezm/status/1355047090087288832.

Looks like as part of Change Signature refactoring

Kobzol:
I suppose that change signature is only for functions, or not? For structs we probably need a separate type of refactoring.

sassman:
Would be very handy to have it for Structs as well.

PR #6745, opened 2021-01-27
ANN: add quick fix to add function parameter from function call

Kobzol:
This PR adds a quick fix that lets the user add new function parameters to a function from a function call, using the new Change signature refactoring. So far it only works for function and method calls and not tuple struct constructors. The behaviour of the fix is heavily inspired by a similar fix in the Kotlin plugin.
add-parameters

There are two TODOs in the code, which I'm not sure how to resolve:
- How to name parameters? `p0`, `p1`, etc.? Or continue counting from the existing parameter count? Also the fix should probably make sure that it doesn't generate collisions in parameter binding names.
- How to get the correct type reference from the argument expression type? That is especially important for import to work. I tried to use a type reference fragment, that sort-of works, but it imports the type with a wrong path. I need to somehow get from `Ty` to `RsTypeReference` that exists in the context of the target function.

Related issue: https://github.com/intellij-rust/intellij-rust/issues/6744
Fixes: https://github.com/intellij-rust/intellij-rust/issues/3582

changelog: Add quick fix to add a parameter to a function from a function call.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
@@ -1,0 +1,252 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInsight.intention.PriorityAction
+import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.ide.annotator.getFunctionCallContext
+import org.rust.ide.presentation.renderInsertionSafe
+import org.rust.ide.refactoring.changeSignature.*
+import org.rust.ide.refactoring.suggestedNames
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.isMethod
+import org.rust.lang.core.psi.ext.valueParameters
+import org.rust.lang.core.types.implLookup
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyUnknown
+import org.rust.lang.core.types.type
+
+/**
+ * Requires that the number of arguments is larger than the number of parameters.
+ */
+class ChangeFunctionSignatureFix private constructor(
+    arguments: RsValueArgumentList,
+    function: RsFunction,
+    private val direction: ArgumentScanDirection,
+) : LocalQuickFixAndIntentionActionOnPsiElement(arguments), PriorityAction {
+    private val fixText: String = run {
+        val signature = calculateSignature(function, arguments, direction)
+        val argumentInsertions = signature.withIndex().filter { (_, item) -> item is SignatureMember.Argument }
+
+        val callableType = if (function.isMethod) "method" else "function"
+        val name = function.name
+        if (argumentInsertions.size == 1) {
+            val index = argumentInsertions[0].index + 1
+            val argument = argumentInsertions[0].value as SignatureMember.Argument
+
+            val ordinal = "${index}${suffix(index)}"
+            "Add `${renderType(argument.expr.type)}` as `$ordinal` parameter to $callableType `$name`"
+        } else {
+            val signatureText = signature.joinToString(", ") {
+                when (it) {
+                    is SignatureMember.Argument -> "<b>${renderType(it.expr.type)}</b>"
+                    is SignatureMember.Parameter -> it.parameter.typeReference?.text.orEmpty()
+                }
+            }
+
+            "<html>Change signature of $name($signatureText)</html>"
+        }
+    }
+
+    override fun getText(): String = fixText
+    override fun getFamilyName(): String = "Add parameter to function"
+
+    override fun startInWriteAction(): Boolean = false
+
+    override fun getPriority(): PriorityAction.Priority = when (direction) {
+        ArgumentScanDirection.Forward -> PriorityAction.Priority.HIGH
+        ArgumentScanDirection.Backward -> PriorityAction.Priority.NORMAL
+    }
+
+    override fun invoke(
+        project: Project,
+        file: PsiFile,
+        editor: Editor?,
+        startElement: PsiElement,
+        endElement: PsiElement
+    ) {
+        val args = startElement as? RsValueArgumentList ?: return
+        val context = when (val parent = args.parent) {
+            is RsCallExpr -> parent.getFunctionCallContext()
+            is RsMethodCall -> parent.getFunctionCallContext()
+            else -> null
+        } ?: return
+        val function = context.function ?: return
+
+        val usedNames = mutableSetOf<String>()
+        getExistingParameterNames(usedNames, function)
+
+        val config = RsChangeFunctionSignatureConfig.create(function)
+        val signature = calculateSignature(function, args, direction)
+        val factory = RsPsiFactory(project)
+
+        for ((index, item) in signature.withIndex()) {
+            if (item is SignatureMember.Argument) {
+                val expr = item.expr
+                val typeText = renderType(expr.type)
+                val fragment = RsTypeReferenceCodeFragment(project, typeText, context = args)
+                val typeReference = fragment.typeReference ?: factory.createType("()")
+                val name = generateName(suggestName(expr), usedNames)
+
+                config.parameters.add(index, Parameter(factory, name, ParameterProperty.fromItem(typeReference)))
+                config.additionalTypesToImport.add(expr.type)
+            }
+        }
+
+        runChangeSignatureRefactoring(config)
+    }
+
+    companion object {
+        fun createIfCompatible(
+            arguments: RsValueArgumentList,
+            function: RsFunction,
+            direction: ArgumentScanDirection
+        ): ChangeFunctionSignatureFix? {
+            if (function.containingCrate?.origin != PackageOrigin.WORKSPACE) return null
+
+            val signature = calculateSignature(function, arguments, direction)
+            val parameterInsertions = signature.filterIsInstance<SignatureMember.Parameter>()
+
+            // There are parameters that were not matched by any argument
+            if (function.valueParameters.size < parameterInsertions.size) {
+                return null
+            }
+
+            return ChangeFunctionSignatureFix(arguments, function, direction)
+        }
+    }
+
+    sealed class ArgumentScanDirection {
+        open fun <T> map(iterable: Iterable<T>): Iterable<T> = iterable
+
+        object Forward : ArgumentScanDirection()
+        object Backward : ArgumentScanDirection() {
+            override fun <T> map(iterable: Iterable<T>): Iterable<T> = iterable.reversed()
+        }
+    }
+}
+
+private fun suggestName(expr: RsExpr): String {
+    if (expr is RsPathExpr) {
+        val reference = expr.path.reference?.resolve()
+        val name = (reference as? RsPatBinding)?.name
+
+        if (name != null) {
+            return name
+        }
+    }
+
+    return expr.suggestedNames().default
+}
+
+private fun generateName(defaultName: String, usedNames: MutableSet<String>): String {
+    var name = defaultName
+    var index = 0
+    while (name in usedNames) {
+        name = "${defaultName}$index"
+        index += 1
+    }
+    usedNames.add(name)
+    return name
+}
+
+/**
+ * Calculates a possible new signature in the specified direction.
+ * Assumes that there are more arguments than parameters.
+ */
+private fun calculateSignature(
+    function: RsFunction,
+    argumentList: RsValueArgumentList,
+    direction: ChangeFunctionSignatureFix.ArgumentScanDirection
+): List<SignatureMember> {
+    val parameters = function.valueParameters
+    val arguments = argumentList.exprList
+
+    if (parameters.isEmpty()) {
+        return arguments.map { SignatureMember.Argument(it) }
+    }
+
+    val ctx = function.implLookup.ctx
+
+    fun <T> createIterator(items: List<T>): PeekableIterator<T> = PeekableIterator(direction.map(items).iterator())
+
+    val parameterIterator = createIterator(parameters)
+    val argumentIterator = createIterator(arguments)
+
+    val insertions = mutableListOf<SignatureMember>()
+    while (true) {
+        val parameter = parameterIterator.value
+        val argument = argumentIterator.value ?: break
+
+        if (parameter != null && ctx.combineTypes(parameter.typeReference?.type ?: TyUnknown, argument.type).isOk) {
+            insertions.add(SignatureMember.Parameter(parameter))
+            parameterIterator.advance()
+            argumentIterator.advance()
+        } else {
+            insertions.add(SignatureMember.Argument(argument))
+            argumentIterator.advance()
+        }
+    }
+
+    return direction.map(insertions).toList()
+}
+
+private class PeekableIterator<T>(private val iterator: Iterator<T>) {
+    var value: T? = null
+
+    init {
+        if (iterator.hasNext()) {
+            value = iterator.next()
+        }
+    }
+
+    fun advance() {
+        value = if (iterator.hasNext()) {
+            iterator.next()
+        } else {
+            null
+        }
+    }
+}
+
+private sealed class SignatureMember {
+    data class Parameter(val parameter: RsValueParameter) : SignatureMember()
+    data class Argument(val expr: RsExpr) : SignatureMember()
+}
+
+private fun suffix(number: Int): String {
+    if ((number % 100) in 11..13) {
+        return "th"
+    }
+    return when (number % 10) {
+        1 -> "st"
+        2 -> "nd"
+        3 -> "rd"
+        else -> "th"
+    }
+}
+
+private fun renderType(ty: Ty): String =
+    ty.renderInsertionSafe(
+        skipUnchangedDefaultTypeArguments = true,
+        useAliasNames = true,
+        includeLifetimeArguments = true
+    )
+
+private fun getExistingParameterNames(usedNames: MutableSet<String>, function: RsFunction) {
+    val visitor = object : RsRecursiveVisitor() {
+        override fun visitPatBinding(o: RsPatBinding) {
+            val name = o.identifier.text
+            usedNames.add(name)
+        }
+    }
+    function.valueParameterList?.acceptChildren(visitor)
+}
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -6,16 +6,20 @@
 package org.rust.ide.annotator
 
 import com.intellij.codeInsight.daemon.impl.HighlightRangeExtension
+import com.intellij.codeInspection.LocalQuickFix
 import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.ide.annotator.AnnotatorBase
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.AnnotationSession
+import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.editor.colors.TextAttributesKey
 import com.intellij.openapi.util.Key
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.psi.util.parentOfType
+import com.intellij.util.containers.addIfNotNull
 import org.rust.cargo.project.workspace.CargoWorkspace.Edition
 import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.annotator.fixes.*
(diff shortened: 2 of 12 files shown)


dima74:
What do you think about adding parameters to start/middle of the function? E.g. Java and TypeScript support this:

image

Kobzol:
Interesting. I never used it personally, but it seems useful. I'll check how it's implemented for Java, it seems that if the target function had multiple parameters already, there could be quite a large number of possible combinations where to insert the new parameters.

Kobzol:
It seems that in Java it iterates through the parameters and arguments in both directions (from left and from right) to generate the two variants, so the algorithm is quite simple actually. If exactly one argument is extraneous, it does the same, but uses a different quick fix text (`Add <type> as n-th parameter to function/method <name>` instead of `Change signature...`).

I'll try to implement it.

Kobzol:
I implemented it and therefore changed the quickfix to `ChangeFunctionSignatureFix`.

Maybe we can keep a separate quick fix that would only add a single parameter, since we already have precise error tracking for each argument? Because now (when we mirror the behaviour of the Java plugin), if you have N additional arguments, the fix will add all of them to the signature.

Also we should offer a quick fix to change the parameter type if the argument count is correct (I already have it partially prepared). Should it be joined with this quick fix somehow?

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,200 @@
...
+        val argumentInsertions = signature.withIndex().filter { (_, item) -> item is SignatureMember.Argument }
+
+        val callableType = if (function.isMethod) "method" else "function"
+        val name = function.name
+        if (argumentInsertions.size == 1) {
+            val index = argumentInsertions[0].index + 1
+            val argument = argumentInsertions[0].value as SignatureMember.Argument
+
+            val ordinal = "${index}${suffix(index)}"
+            "Add `${argument.expr.type.renderInsertionSafe()}` as `$ordinal` parameter to $callableType `$name`"
```
dima74, line 48: `skipUnchangedDefaultTypeArguments = true` ?

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,200 @@
...
+        } else {
+            val signatureText = signature.joinToString(", ") {
+                when (it) {
+                    is SignatureMember.Argument -> "<b>${it.expr.type.renderInsertionSafe()}</b>"
```
dima74, line 52: For some reason `<b>` does not work

image

It is interesting how Java plugin does this

dima74 (commented):
Sorry for the long delay,

> * How to name parameters? `p0`, `p1`, etc.? Or continue counting from the existing parameter count?

Kotlin plugin tries to create parameter name based on argument expression. E.g.:

```rust
struct Foo { field: i32 }
fn test1(foo: Foo) {
    func(foo);          // parameter name is "foo"
    func(foo.field);    // parameter name is "field"
}
fn func() {}
```

For parameters where such detection is not possible (e.g. integer literal), we can use `p0`, `p1`, etc. Also I think counting from zero or from existing parameter count is not important because user will change argument name anyway

> Maybe we can keep a separate quick fix that would only add a single parameter, since we already have precise error tracking for each argument? Because now (when we mirror the behaviour of the Java plugin), if you have N additional arguments, the fix will add all of them to the signature.

I am not sure that we need it, do you know any examples in other languages? I mean why user would want to add only one parameter is he/she can add all parameters at once?

Btw currently when we add parameter to the beginning/middle of function, inspection is available only for last parameter, which is a bit unintuitive. In java there is single inspection covering all function arguments. What do you think about using single inspection for all arguments when argument count is incorrect, and using precise argument tracking when argument count is correct?

> Also we should offer a quick fix to change the parameter type if the argument count is correct (I already have it partially prepared). Should it be joined with this quick fix somehow?
Actually I don't know how they can be joined since they should be offered in different cases (correct vs incorrect argument count). But we may try to reuse some basic functions

> * How to get the correct type reference from the argument expression type? That is especially important for import to work. I tried to use a type reference fragment, that sort-of works, but it imports the type with a wrong path. I need to somehow get from `Ty` to `RsTypeReference` that exists in the context of the target function.
Why do we need to get from `Ty` to `RsTypeReference`, because change signature refactoring works with `RsTypeReference`? I propose to generalize `Parameter` or `ParameterType` class so that it can store either `Ty` or `RsTypeReference`, and in case of `Ty` we will directly use `RsImportHelper#importTypeReferencesFromTy`

Kobzol:
> Also I think counting from zero or from existing parameter count is not important because user will change argument name anyway.

That will be indeed true most of the time, but I still think that the refactoring should produce correct code, e.g. the generated names should not conflict with existing parameter bindings. I'll improve the name generation to use `RsExpr.suggestedNames` and add some sanity check to avoid conflicts.

> I am not sure that we need it, do you know any examples in other languages? I mean why user would want to add only one parameter is he/she can add all parameters at once?

Okay, good point, we probably don't need it.

> What do you think about using single inspection for all arguments when argument count is incorrect, and using precise argument tracking when argument count is correct?

I'm not opposed to it, but what about https://github.com/intellij-rust/intellij-rust/pull/6622? It seems useful. Can two inspections "overlap"? I.e. could we have one inspection that covers all arguments and then another ones that cover individual arguments? I recall that there were some problems with this, but I'm not sure.

> Why do we need to get from Ty to RsTypeReference, because change signature refactoring works with RsTypeReference? I propose to generalize Parameter or ParameterType class so that it can store either Ty or RsTypeReference, and in case of Ty we will directly use RsImportHelper#importTypeReferencesFromTy

Okay, sounds reasonable, I'll do it.

dima74:
> Can two inspections "overlap"? I.e. could we have one inspection that covers all arguments and then another ones that cover individual arguments? I recall that there were some problems with this, but I'm not sure.

I am not aware of problems related to it. E.g. we have overlapping inspections in [`RsSimplifyBooleanExpressionInspectionTest`](https://github.com/intellij-rust/intellij-rust/blob/2f423617ad2fe6d122243dfd6123f956b11776a4/src/test/kotlin/org/rust/ide/inspections/RsSimplifyBooleanExpressionInspectionTest.kt#L11)

> I'm not opposed to it, but what about #6622? It seems useful.

Good question, maybe convert #6622 to single inspection covering all arguments (so that all extra arguments will be removed at once)?

Kobzol:
> Good question, maybe convert #6622 to single inspection covering all arguments (so that all extra arguments will be removed at once)?

Hmm, I introduced the precise argument error tracking especially for the reason so that you could quickly see what's wrong with the arguments. If we remove it and mark the whole argument list with an error, it will not be immediately obvious what is wrong.

What about keeping the precise error tracking for extra arguments and allowing them to be deleted/added to the target function individually, but also offering a separate inspection spanning the whole argument list that would either remove all extra arguments or add all of them to the target function? It could have an "invisible" or e.g. a transparent grey underline and it would be available in the whole argument list.

Kobzol:
That sounds like a nice compromise! So basically we would mark only the erroneous arguments, but the quick fixes would be available on the whole argument list and they would offer to either remove all extra arguments or add all extra arguments to the target function, right?

dima74:
> That sounds like a nice compromise! So basically we would mark only the erroneous arguments, but the quick fixes would be available on the whole argument list and they would offer to either remove all extra arguments or add all extra arguments to the target function, right?

Yes, exactly!

Kobzol:
Ok! I'll change it in this PR.

Kobzol:
Btw, Kotlin uses two quick fixes (one to remove a specific argument and another one to add all extra arguments to the target function).

Kobzol:
> Why do we need to get from Ty to RsTypeReference, because change signature refactoring works with RsTypeReference? I propose to generalize Parameter or ParameterType class so that it can store either Ty or RsTypeReference, and in case of Ty we will directly use RsImportHelper#importTypeReferencesFromTy

I started to change it and it complicates the code a lot. Basically everything (change signature config, change signature dialog) requires TypeReference, the Ty would only be used for this fix. Also I would probably have to reimplement `RsPsiFactory::createValueParameter` to work with a string (a rendered type) instead of a type reference.

Is there really no other way how to get the type reference directly from the argument? :( For example using `render(useQualifiedName = ...)` and create the qualified element set from the type's ADT item?

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,200 @@
...
+ * Requires that the number of arguments is larger than the number of parameters.
+ */
+class ChangeFunctionSignatureFix private constructor(
+    arguments: RsValueArgumentList,
+    function: RsFunction,
+    private val direction: ArgumentScanDirection
+) : LocalQuickFixAndIntentionActionOnPsiElement(arguments) {
+    private val fixText = run {
+        val signature = calculateSignature(function, arguments, direction)
+        val argumentInsertions = signature.withIndex().filter { (_, item) -> item is SignatureMember.Argument }
```
dima74, line 39:
```suggestion
        val argumentInsertions = signature.filterIsInstance<SignatureMember.Argument>().withIndex()
```

dima74:
> Is there really no other way how to get the type reference directly from the argument? :(

I am not aware of such way

> For example using `render(useQualifiedName = ...)` and create the qualified element set from the type's ADT item?

I do not quite understand how it will work, if we use `useQualifiedName`, then all paths will be absolute?

---

Actually [`importTypeReferencesFromTy`](https://github.com/intellij-rust/intellij-rust/blob/71d3224bcaa279d1b0cb02048873fedcd29d9c29/src/main/kotlin/org/rust/ide/utils/import/RsImportHelper.kt#L32-L35) is implemented in the way, that we first collect all items to import, and then import them. What if we collect all items to import based on `Ty` (before calling `runChangeSignatureRefactoring`), and pass these items as additional field in `RsChangeFunctionSignatureConfig`?

Kobzol:
> I do not quite understand how it will work, if we use useQualifiedName, then all paths will be absolute?

I wanted to use qualified paths when rendering the type into a code fragment, so that the fragment could resolve the type reference of the type from any location. But I'm not even sure if it would work. I like your idea with passing additional types for import, I will try it.

Kobzol:
I pushed a new version, please try it interactively, the tests are currently broken.
I kept the old fix to remove a single argument (same as it works in Kotlin), I like it a bit better, let me know what you think.

The annotation spanning the whole argument range is currently using the `INFO` severity, so that it doesn't mark the range with a red underline. However, this also causes the quick fix name to be displayed in the intention list, which is not optimal. I tried playing with highlight severity and problem highlight, but couldn!t find a combination that would both:
1) Keep the quick fix name in the "error" category after Alt + Enter is pressed
2) Do not display a red underline

dima74:
> The annotation spanning the whole argument range is currently using the `INFO` severity, so that it doesn't mark the range with a red underline. However, this also causes the quick fix name to be displayed in the intention list, which is not optimal. I tried playing with highlight severity and problem highlight, but couldn!t find a combination that would both:

What do you think about using `WARNING` or `WEAK_WARNING` severity for whole-range annotation? It will place quick fixes at the top of the list. Although it looks a bit strange:

image
image

Kobzol:
I personally wouldn't use warning with the yellow background, it looks weird and it's an error, not really just a warning. The weak warning looks a bit better, but it hides the extra arguments with the yellow squiggle. Ideally I would want something like "invisible error", it behaves as an error, but it has no visible marking.

dima74:
> Ideally I would want something like "invisible error", it behaves as an error, but it has no visible marking.

That would be ideal. Sadly I can't find a way to achieve this. There is `AnnotationHolder#newSilentAnnotation`, but looks like it doesn't do what we want.

So if we can't get whole-range annotations to work properly, looks like the only solution is to remove whole-range annotations and keep only individual arguments annotations

Kobzol:
What about this (to be refactored)? Severity is set to error, but text attributes are reset to default.
```kotlin
var builder = holder.newAnnotation(args, HighlightSeverity.ERROR, "Wrong arguments") ?: return
for (fix in fixes) {
    builder = builder.withFix(fix)
}
builder.textAttributes(TextAttributesKey.createTextAttributesKey("DEFAULT_TEXT_ATTRIBUTES"))
builder.create()
```
ann-error

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
Kobzol: This actually broke the logic, since the indices need to be collected before the filtering :sweat_smile:.

dima74:
This is exactly as we wanted! Lets do it that way

Kobzol:
Great! I already refactored it and modified the tests to use the new format, I just need to resolve a few details.

I think that with this type of annotation, we might as well go back to your idea of having a single fix for removing all extra arguments - but at the same time marking the individual arguments with separate red squiggles. WDYT?

@vlad20012 In the quick fix, I need to check if two types are "compatible", more specifically if an argument expression has a valid type for a given parameter. I tried this: https://github.com/intellij-rust/intellij-rust/pull/6745/files#diff-328479af1d56df7e76cd6d3e73246e1f737064829322ebdc5e40bd0f3f241032R164, but it returns false e.g. for `u32` and `i32` in this case:
```rust
fn foo(a: u32) {}
fn bar() {
    foo(1); // should be compatible, even though default type of 1 is i32
}
```
Is there a better way?

dima74:
> I think that with this type of annotation, we might as well go back to your idea of having a single fix for removing all extra arguments - but at the same time marking the individual arguments with separate red squiggles. WDYT?

So do I understand correctly that totally there will be three quick-fixes: add all arguments to the target function and remove one/all argument from function call?

vlad20012:
> Is there a better way?

I don't know a better way. `{integer}` fell back to `i32` at the end of type inference of the function, and the information about it is lost.

Kobzol:
> I don't know a better way. {integer} fell back to i32 at the end of type inference of the function, and the information about it is lost.

Thanks, knowing that this way is OK is good enough for me.

> So do I understand correctly that totally there will be three quick-fixes: add all arguments to the target function and remove one/all argument from function call?

I meant to do it the Java way, so only have a fix to remove ALL extra arguments, but after all, why not having both fixes. Sometimes you might want to remove a single one, most of the times you might want to remove all, so we could have both fixes and give a higher priority to the one that removes all extra arguments.

Kobzol:
I pushed the current version.

dima74:
> I meant to do it the Java way, so only have a fix to remove ALL extra arguments, but after all, why not having both fixes. Sometimes you might want to remove a single one, most of the times you might want to remove all, so we could have both fixes and give a higher priority to the one that removes all extra arguments.
> 
> (You can see that I can't decide :D So having both of them is a compromise).

I personally would like to have only quick-fix to remove all extra arguments. But if you want both, then I am not opposed

dima74:
* Can we change order of quick-fixes (adding parameters to begin/end of the function), so that adding to the end of function will be first quick-fix (since looks like it will be used more often)

* We probably should change existing parameters types if needed, e.g. we have `fn func(s: &str) {}` and `func(1, 2);`. After refactoring we will have `fn func(i: i32, i0: i32, s: &str) {}`, though expected result is `fn func(i: i32, i0: i32) {}`

Kobzol:
> I personally would like to have only quick-fix to remove all extra arguments. But if you want both, then I am not opposed

Let's keep it simple then. I created a single fix to remove all extra arguments :+1:

> Can we change order of quick-fixes (adding parameters to begin/end of the function), so that adding to the end of function will be first quick-fix (since looks like it will be used more often)

I already swapped the order based on your previous suggestion. Is it enough to change the order of the fix in the list of fixes? Or does this have to be solved with some priority?

> We probably should change existing parameters types if needed, e.g. we have fn func(s: &str) {} and func(1, 2);. After refactoring we will have fn func(i: i32, i0: i32, s: &str) {}, though expected result is fn func(i: i32, i0: i32) {}

This will probably be more complicated and it could generate a lot of variants of the target signature that could be created (unless we decide that we will only allow to change the type of the parameter in the same index as the argument). I suggest to create a new PR that will be focused on changing parameters types (I already mentioned it [here](https://github.com/intellij-rust/intellij-rust/issues/6744)), as this PR is already quite complex.

Btw in Java, here:
```java
private void foo(int a) {}
private void bar() {
    foo("a", 1)
}
```
There is only a single fix offered, add String as a first parameter. If I remove the `1` argument, it will offer to change the type of `a` to String.

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Requires that the number of arguments is larger than the number of parameters.
+ */
+class ChangeFunctionSignatureFix private constructor(
+    arguments: RsValueArgumentList,
+    function: RsFunction,
+    private val direction: ArgumentScanDirection
+) : LocalQuickFixAndIntentionActionOnPsiElement(arguments) {
+    private val fixText = run {
```
dima74, line 35:
```suggestion
    private val fixText: String = run {
```

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    override fun invoke(
+        project: Project,
+        file: PsiFile,
+        editor: Editor?,
+        startElement: PsiElement,
+        endElement: PsiElement
+    ) {
+        val args = startElement as? RsValueArgumentList ?: return
+        val ctx = when (val parent = args.parent) {
```
dima74, line 72:
```suggestion
        val context = when (val parent = args.parent) {
```

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,237 @@
...
+        name = "${defaultName}$index"
+        index += 1
+    }
+    usedNames.add(name)
+    return name
+}
+
+/**
+ * Calculates a possible new signature in the specified direction.
+ * Assumes that there are more arguments that parameters.
```
dima74, line 149:
```suggestion
 * Assumes that there are more arguments than parameters.
```

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -0,0 +1,237 @@
...
+            argumentIterator.advance()
+        } else {
+            insertions.add(SignatureMember.Argument(argument))
+            argumentIterator.advance()
+        }
+    }
+    return direction.map(insertions).toList()
+}
+
+private class PeekableIterator<T>(val iterator: Iterator<T>) {
```
dima74, line 187:
```suggestion
private class PeekableIterator<T>(private val iterator: Iterator<T>) {
```

dima74 (commented):
> I already swapped the order based on your previous suggestion. Is it enough to change the order of the fix in the list of fixes? Or does this have to be solved with some priority?

Looks like it is not enough. We can try `ChangeFunctionSignatureFix` to implement `com.intellij.codeInsight.intention.PriorityAction`

> I suggest to create a new PR that will be focused on changing parameters types (I already mentioned it here), as this PR is already quite complex.

Good idea, so let's disable the quick fix in case when types of function parameters are not compatible with types of arguments (like in that example), in order to not generate incorrect code in this case

Kobzol:
Added priority to the quick fix and disabled it when there are unmatched parameters.

dima74:
Thanks!
bors r+

Kobzol:
@dima74 I started thinking about the follow-up fix to change the type of a parameter and I want to discuss it first before implementing it.

The fix should offer to change the parameter type, if the type of an argument does not match the type of the parameter.
```rust
fn foo(
    a: u32,
    b: u32
) {}

fn main() {
    foo(
       true, // error: Change the type of first parameter of function `foo` to `bool`
       "", // error: Change the type of second parameter of function `foo` to `&str`
    );
}
```
1) Wording of the fix
As with the fix in this PR, we cannot always directly reference the name of the parameter, because the name might be complex (nested patterns etc.). I would suggest to use the name of the parameter if it has a simple binding (`change the type of parameter a ...`) and use its position if not (`change the type of the first parameter ...`).

2) Synergy with other fixes
There are already fixes that can change the type of the argument instead, i.e.:
```rust
fn foo(a: &i32) {}
fn main() {
    foo(1); // Convert to &i32 using & 
}
```
I suggest that we somehow use priority to make sure that these fixes are on top, as it's probably much more common to change the type of the argument instead of changing the signature the function.

3) Behaviour of the change signature fix if there are unmatched parameters
There are two situations that can happen when you have an invalid argument list. Based on them, we currently decide if we offer the change signature fix or not. With the change type fix the situation gets more complicated.

a) Each parameter has a matching argument (i.e. for each parameter, there is some argument with the same type that can be matched with it).
This situation is supported by this PR. Based on the number of extra arguments, either `Add X as N-th parameter` or `Change signature` fixes will be provided. We can also support the change type fix in this case.
```rust
fn foo(a: u32) {}
fn main() {
    foo(true, 0); // insert bool as the first parameter + change the type of first parameter to bool
    foo(true, true, 0); // change signature + change the type of first parameter to bool
}
```
b) There is some parameter that has no matching argument.
Currently this situation is ignored by the curent PR (the corresponding fixes are disabled to avoid generating weird refactoring).
```rust
fn foo(c: u32) {}
fn main() {
     foo(true, true);
     // change signature -> fn foo(a: bool, b: bool, c: u32)? or
     // change signature -> fn foo(a: bool, b: bool)
}
```
Currently, the change signature fix is disabled in this case to avoid generating the first signature. Yet it probably would be useful
to offer the fix also in this case. We should somehow decide whether we want to keep adding things to the signature, or if we want to change the signature altogether. We could base this decision on the presence of an unmatched parameter.

(This point is actually a bit orthogonal from change type fix and could be implemented in a separate PR).

4) Error/annotation range
Currently, the whole argument list is being annotated with an (invisible) error, so that you can invoke change signature or remove all argument fixes from anywhere. This does not play well with the "change type" fix, since that one is specific for a given argument/parameter.

    This depends on how we decide to handle point 3), but I would suggest to not show the "change signature fix" on arguments that will have the "change type fix". It seems natural to me, Kotlin also does this. It would be a bit weird if we showed both I think.

PR #6769, opened 2021-01-31
ANN: fix false positive E0658 in stdlib

Undin:
changelog: Fix false-positive [E0658](https://doc.rust-lang.org/error-index.html#E0658) in stdlib


Undin:
Actually, I don't know a good way to check these changes in our tests. But regression tests show that it fixes quite a lot of false positives in stdlib, so we can use regressions tests as a verifier for these changes

Screen Shot 2021-03-10 at 15 13 41

mchernyavsky (approved):
bors r+

PR #6773, opened 2021-02-01
RES: Use old resolve in test/bench non-workspace crates

dima74:
Because currently we don't build [DefMaps](https://github.com/intellij-rust/intellij-rust/issues/6217) for such crates in new resolve (for performance reasons)

changelog: Use old resolve in test/bench non-workspace crates


vlad20012 (approved):
bors r+

Issue #6781, opened 2021-02-02
Wrong control-flow of break/continue inside block expressions


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.3638-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3.2 Preview Ultimate Edition (IU-203.7148.15)
* **Operating system:** macOS 10.16
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
`break` and `continue` located inside block expressions break control-flow/data-flow analysis, e.g. liveness analysis:
```Rust
fn main() {
    let x = 1; // false-positive: Variable `x` is never used
    loop {
        {
            break;
        }
    }
    println!("{}", x);
}
```

PR #6782, opened 2021-02-02
Fix control flow of break/continue inside block expression

ortem:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6781

changelog: Fix processing of `break`/`continue` inside block expression during control-flow analysis


ortem:
bors retry

Issue #6784, opened 2021-02-02
Control-flow analysis ignores expression `cfg` attributes


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.3638-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3.2 Preview Ultimate Edition (IU-203.7148.15)
* **Operating system:** macOS 10.16
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
Currently, our control-flow analysis does not take into account [expression `cfg` attributes](https://doc.rust-lang.org/reference/expressions.html#expression-attributes). This leads to false-positives and false-negatives in several inspections, e.g. in liveness
```Rust
fn main() {
    let a = 1; // false-negative: `a` should be annotated as unused
    #[cfg(disabled_option)] println!("{}", a);
}
```
and in move analysis
```Rust
struct S;
fn consume(s: S) {}

fn main() {
    let s = S;
    #[cfg(disabled_option)] consume(s);
    let t = s; // false-positive: Use of moved value
}
```

PR #6786, opened 2021-02-02
Take `cfg` attributes into account in control flow

ortem:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6784

changelog: Take `cfg` attributes into account during control-flow analysis

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/dfa/CFGBuilder.kt
+++ b/src/main/kotlin/org/rust/lang/core/dfa/CFGBuilder.kt
@@ -156,6 +156,12 @@ class CFGBuilder(
     fun process(element: RsElement?, pred: CFGNode): CFGNode {
         if (element == null) return pred
 
+        // Conditionally disabled code should be ignored since it does not affect the execution
+        // https://doc.rust-lang.org/reference/expressions.html#expression-attributes
+        if (element is RsOuterAttributeOwner && !element.isEnabledByCfgSelf) {
+            return pred
+        }
+
         result = null
         val oldPredsSize = preds.size
         preds.push(pred)
@@ -247,7 +253,7 @@ class CFGBuilder(
     override fun visitMacroExpr(macroExpr: RsMacroExpr) {
         val macroCallExit = process(macroExpr.macroCall, pred)
 
-        if (macroExpr.type is TyNever) {
+        if (macroCallExit != pred && macroExpr.type is TyNever) {
             finishWithUnreachableNode(macroCallExit)
         } else {
             finishWith(macroCallExit)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/dfa/CFGBuilder.kt
```diff
@@ -216,8 +217,15 @@ class CFGBuilder(
     override fun visitLabelDecl(labelDecl: RsLabelDecl) = finishWith(pred)
 
     override fun visitExprStmt(exprStmt: RsExprStmt) {
-        val exprExit = process(exprStmt.expr, pred)
-        finishWithAstNode(exprStmt, exprExit)
+        val expr = exprStmt.expr
+        // Conditionally disabled code should be ignored since it does not affect the execution
+        // https://doc.rust-lang.org/reference/expressions.html#expression-attributes
+        if (expr is RsOuterAttributeOwner && !expr.isEnabledByCfgSelf) {
```
dima74, line 223: Looks like some our classes, doesn't implement `RsOuterAttributeOwner`, though actually they should, e.g. `RsMacroExpr` and `RsCallExpr`. Because of this we don't have unused variable warning in code like this:
```rust
fn main() {
    let a/*caret*/ = 1;
    #[cfg(windows)] println!("{}", a);
}
```

```rust
fn main() {
    let a/*caret*/ = 1;
    #[cfg(windows)] foo(a);
}

fn foo(a: i32) {}
```

ortem: Thank you for finding these cases!

`RsMacroExpr` is not `RsOuterAttributeOwner` intentionally, see https://github.com/intellij-rust/intellij-rust/commit/0eee09bec8052865d1836a994388a03857bd6d14 (but I don't know the exact reason actually). However there is an easy workaround: check `cfg` attributes right inside `visitMacroExpr`.

But as for `RsCallExpr`, there is a real problem with the parser, so `#[cfg(x)] foo(a)` is parsed as `{#[cfg(x)] foo]} {(a)}`: 
Screenshot 2021-02-03 at 19 05 00
I've opened a separate issue for that https://github.com/intellij-rust/intellij-rust/issues/6797

src/main/kotlin/org/rust/lang/core/dfa/CFGBuilder.kt
```diff
@@ -216,8 +217,15 @@ class CFGBuilder(
     override fun visitLabelDecl(labelDecl: RsLabelDecl) = finishWith(pred)
 
     override fun visitExprStmt(exprStmt: RsExprStmt) {
```
dima74, line 219: Looks like we also need to handle cfg in `visitLetDecl`:
```rust
fn main() {
    let a = 1;
    #[cfg(windows)] let b = a;
}
```

dima74:
What about struct literals:
```rust
#[derive(Debug)]
struct Foo {
    #[cfg(windows)]
    field: i32,
}

fn test1() {
    let unused = 1;
    let foo = Foo {
        #[cfg(windows)]
        field: unused,
    };
    dbg!(foo);
}
```

and function arguments:
```rust
fn test2() {
    let unused = 1;
    foo1(#[cfg(windows)] unused);
}

fn foo1(#[cfg(windows)] a: i32) {}
```

ortem:
@dima74 Oh, I didn't notice these cases initially. Now, it seems like `element is RsOuterAttributeOwner && !element.isEnabledByCfgSelf` filtering should be performed for any processed elements.

Note that there is a special processing checkup `macroCallExit != pred` in `visitMacroExpr` in order not to generate termination edge for macro expressions with conditionally disabled macro calls inside.

dima74 (approved):
👍
bors r+

PR #6787, opened 2021-02-02
T: Introduce CratesLocalIndexService testing

avrong:
Adds tests for CratesLocalIndexService that use real crates.io registry.

Uses [tokio](https://crates.io/crates/tokio) and [code-generation-example](https://crates.io/crates/code-generation-example) crates to test index.

### TODO:
- [x] Decide on what crate we should perform tests
- [ ] ~~Clean resources after performing all the tests~~
   - [ ] ~~Remove index from the disk~~
   - [ ] ~~Unload service~~
 - [x] Get a real service in testing environment


toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -134,7 +136,7 @@ class CratesLocalIndexService : PersistentStateComponent<CratesLocalIndexState>,
         return crateNames
     }
 
-    private fun updateIfNeeded() {
+    fun updateIfNeeded() {
```
Undin, line 139: We usually mark such methods with `com.google.common.annotations.VisibleForTesting` annotation. I believe in the future kotlin plugin will have an inspection to detect usages of marked methods in inappropriate places. Java plugin has such inspection

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -102,6 +102,8 @@ class CratesLocalIndexService : PersistentStateComponent<CratesLocalIndexState>,
         })
     }
 
+    fun isReady(): Boolean = isReady.get()
```
Undin, line 105: nit: probably, property with custom getter is more idiomatic here

Undin (commented):
Consider disabling listener in `CratesLocalIndexService` init block in tests since all tests use the same `Application` instance and this listener will work in all tests and can affect them

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -274,6 +279,33 @@ class CratesLocalIndexServiceImpl
         private const val CRATES_INDEX_VERSION: Int = 0
 
         private val LOG: Logger = logger<CratesLocalIndexServiceImpl>()
+
+        private val _TEST_INSTANCE: CratesLocalIndexService by lazy {
+                CratesLocalIndexServiceImpl().apply {
```
ortem, line 284: Please reformat these lines

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -274,6 +279,33 @@ class CratesLocalIndexServiceImpl
...
+        /**
+         * Simulates the behaviour of real service.
+         * Returns the service object in local [_TEST_INSTANCE], where it is being properly initialized on first access.
+         */
+        @TestOnly
+        fun getTestInstance(): CratesLocalIndexService = _TEST_INSTANCE
+    }
+
+    @TestOnly
+    @Volatile
```
ortem, line 298: Looks like this `@Volatile` is redundant because the variable is accessed from only one thread

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -274,6 +279,33 @@ class CratesLocalIndexServiceImpl
         private const val CRATES_INDEX_VERSION: Int = 0
 
         private val LOG: Logger = logger<CratesLocalIndexServiceImpl>()
+
+        private val _TEST_INSTANCE: CratesLocalIndexService by lazy {
```
ortem, line 283: `Object or top-level property name '_TEST_INSTANCE' should not start with an underscore`
Let's use `TEST_INSTANCE` instead. Looks like there is no conflict with `getTestInstance` because of uppercase

ortem (approved):
LGTM except for minor comments.
@Undin what do you think about the current implementation, should we finally merge this?

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
```diff
@@ -8,6 +8,8 @@ package org.rust.toml.crates.local
 import com.intellij.openapi.components.service
 
 interface CratesLocalIndexService {
+    fun isReady(): Boolean
```
Undin, line 11: Let's avoid this method in public API. As I can see, we need it only for test purposes. Let's add it only in `Impl` class with `TestOnly` annotation if it's really required

toml/src/test/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+class CratesLocalIndexServiceTest : BasePlatformTestCase() {
+    lateinit var cratesService: CratesLocalIndexService
+
+    override fun setUp() {
+        super.setUp()
+        cratesService = CratesLocalIndexServiceImpl.getTestInstance()
+    }
+
+    fun `test service is ready`() {
```
Undin, line 18: Why do we need this test?
As I can see, it checks nothing because `CratesLocalIndexServiceImpl` is ready just after creation

toml/src/test/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceTest.kt
```diff
@@ -0,0 +1,69 @@
...
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.crates.local
+
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
+
+class CratesLocalIndexServiceTest : BasePlatformTestCase() {
+    lateinit var cratesService: CratesLocalIndexService
```
Undin, line 11: Since we use single global instance for testing, `lateinit var` looks redundant

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexService.kt
ortem: We discussed with @avrong that this method is a part of API because it will used later in inspections based on the crates index

toml/src/test/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceTest.kt
avrong: It ensures that the update of the index has been finished as the test gets to run. Test instance from `CratesLocalIndexServiceImpl.getTestInstance()` will update it on first access.

But I'm not sure if it is the right way to check this and if there is a need for this

Undin: Note, currently `isReady` is not touched anywhere during `CratesLocalIndexServiceTest` except explicit call in `test service is ready` because of a way how test instance is initialized.

toml/src/test/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceTest.kt
avrong: What can be the better way to handle this?

If I try to initialize the service right away then the tests fail with
```
Exception in constructor: test tokio features (java.lang.IllegalStateException: ApplicationManager.getApplication() must not be null
	at com.intellij.openapiext.UtilsKt.isUnitTestMode(utils.kt:10)
	at org.rust.toml.crates.local.CratesLocalIndexServiceImpl.<init>(CratesLocalIndexServiceImpl.kt:100)
	at org.rust.toml.crates.local.TestCratesLocalIndexServiceImpl$Companion$REAL_INSTANCE$2.invoke(TestCratesLocalIndexServiceImpl.kt:20)
	at org.rust.toml.crates.local.TestCratesLocalIndexServiceImpl$Companion$REAL_INSTANCE$2.invoke(TestCratesLocalIndexServiceImpl.kt:18)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at org.rust.toml.crates.local.TestCratesLocalIndexServiceImpl$Companion.getREAL_INSTANCE(TestCratesLocalIndexServiceImpl.kt)
	at org.rust.toml.crates.local.TestCratesLocalIndexServiceImpl$Companion.getRealInstance(TestCratesLocalIndexServiceImpl.kt:30)
	at org.rust.toml.crates.local.CratesLocalIndexServiceTest.<init>(CratesLocalIndexServiceTest.kt:11)
```

It happens when the service tries to check for `isUnitTestMode` in the `init` block

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -275,6 +283,19 @@ class CratesLocalIndexServiceImpl
...
+
+    /**
+     * Simulates the behaviour of real service.
+     * Loads crates local index in current thread and updates local [indexedCommitHash].
+     */
+    @TestOnly
+    fun updateCrates() {
+        updateCrates(registryHeadCommitHash, indexedCommitHash)
+        indexedCommitHash = registryHeadCommitHash
+    }
```
Undin, line 298: Why do we need this additional code? Let's better use the existing one. I.e. we already have a place where the service keeps commit hash. Also, we already have a method to update info about crate registry - `updateIfNeeded`. Note, although `updateIfNeeded` uses background task to do it, in tests `com.intellij.openapi.progress.Task#queue` is synchronous so let's better use it instead.
Otherwise, we use a lot of code that is never called in production. That makes our tests less useful than they can be

Undin (changes requested):

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
ortem: > in tests `com.intellij.openapi.progress.Task#queue` is synchronous

Wow, I didn't know about it either, thanks for the clarification!

toml/src/test/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun `test code-generation-example has specific versions`() {
+        assertEquals(
+            cratesService.getCrate("code-generation-example")?.versions?.map { it.version},
+            listOf("0.1.0", "0.2.0")
+        )
+    }
+
+    fun `test code-generation-example has latest version`() {
+        assertEquals(
+            cratesService.getCrate("code-generation-example")?.lastVersion,
```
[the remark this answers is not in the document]
avrong: The idea was to use the crate we own to control its metadata e.g. have a special immutable crate for testing such things

toml/src/main/kotlin/org/rust/toml/crates/local/TestCratesLocalIndexServiceImpl.kt
```diff
@@ -10,8 +10,21 @@ import org.jetbrains.annotations.TestOnly
 class TestCratesLocalIndexServiceImpl : CratesLocalIndexService {
     var testCrates: Map<String, CargoRegistryCrate> = emptyMap()
 
+    private val _realInstance: CratesLocalIndexService by lazy {
+        CratesLocalIndexServiceImpl().apply {
+            updateIfNeeded()
+        }
+    }
+
```
Undin, line 18: Let's move this code to `CratesLocalIndexServiceTest`. Currently, we use single static manually created object of service for testing. It's ok to speed up test execution but we need this instance only in `CratesLocalIndexServiceTest` and don't need it in `TestCratesLocalIndexServiceImpl`

Undin (commented):
LGTM except comments

toml/src/test/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceTest.kt
```diff
@@ -0,0 +1,59 @@
...
+                ?.find { it.version == "1.0.0" }
+                ?.features,
+            listOf("io-util", "process", "macros", "rt", "io-std", "sync", "fs", "rt-multi-thread", "default", "test-util", "time", "net", "signal", "full")
+        )
+    }
+
+    fun `test code-generation-example has specific versions`() {
+        val versions = cratesService.getCrate("code-generation-example")?.versions?.map { it.version }
+
+        kotlin.test.assertNotNull(versions)
```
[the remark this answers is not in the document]
Undin: Oh, I see. I don't mind using this lib but I usually try to avoid using libraries from IDEA since it causes issues from time to time
I think you can just write `assertTrue(versions!!.contains("0.1.0"))`

Undin (approved):
Let's disable listener for `VirtualFileManager.VFS_CHANGES` topic inside `CratesLocalIndexServiceImpl` if `isUnitTestMode` to avoid some weird interactions between tests

Undin (approved):
bors r+

Issue #6772, opened 2021-02-01
`Fill missing arguments` adds redundant argument after introducing new parameter


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.3634-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3.1 (CL-203.6682.181)
* **Operating system:** macOS 10.16
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
![change_signature_bug](https://user-images.githubusercontent.com/4854600/106447407-82f48d80-6492-11eb-8fac-279d9eda1dd8.gif)

Before `Change signature` + `Fill missing arguments`:
```Rust
fn process_person(first_name: &str, age: u8) {
}
fn main() {
    process_person("Alice", 42);
}
```

After:
```Rust
fn process_person(first_name: &str, last_name: &str, age: u8) {
}
fn main() {
    process_person("Alice", "", 0), 42); // expected: `process_person("Alice", "", 42);`
}
```

Kobzol:
Hmm, I guess that fill missing arguments wasn't really intended for situations where the argument is missing, but the comma is already there :D Time to fix it.

Kobzol:
@ortem
This:
```rust
fn foo(a: u32, b: &str, c: u32) {}
fn main() {
    foo(0,/*caret*/ ,1);
}
```
is currently parsed like this:
```
RsExprStmtImpl(EXPR_STMT)
  RsCallExprImpl(CALL_EXPR)
    RsPathExprImpl(PATH_EXPR)
    RsValueArgumentListImpl(VALUE_ARGUMENT_LIST)
      PsiElement(()
      RsLitExprImpl(LIT_EXPR)
      PsiElement(,)
      PsiErrorElement:')' or <expr> expected, got ','
  PsiWhiteSpace
  PsiElement(,)
  RsExprStmtImpl(EXPR_STMT)
    RsLitExprImpl(LIT_EXPR)
    PsiErrorElement:'(', ';', <operator> or '[' expected, got ')'
  PsiElement())
```
And the `exprList` of the call arguments contains just the first argument (`0`). This is a bit problematic for this quick fix.

Without changing the parser, I tried to reconstruct the arguments step-by-step, i.e. put the first missing argument after a comma, then the second etc. It inserts the code correctly, i.e. I get from
[...]

PR #6788, opened 2021-02-02
ANN: correctly fill empty function call arguments

Kobzol:
See https://github.com/intellij-rust/intellij-rust/issues/6772#issuecomment-770784560 for an explanation why it is so complicated :D

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6772

changelog: Correctly fill empty function call arguments.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -930,10 +930,17 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
 
         val realCount = args.exprList.size
         if (realCount < expectedCount) {
-            val target = args.rparen ?: args
+            // Mark only the right parenthesis
+            val rparen = args.rparen
+            if (rparen != null) {
+                holder.newErrorAnnotation(rparen, null)?.create()
+            }
+
+            // But enable the quick fix on the whole argument range
             RsDiagnostic.IncorrectFunctionArgumentCountError(
-                target, expectedCount, realCount, functionType,
-                listOf(FillFunctionArgumentsFix(args))
+                args, expectedCount, realCount, functionType,
+                listOf(FillFunctionArgumentsFix(args)),
+                textAttributes = TextAttributesKey.createTextAttributesKey("DEFAULT_TEXT_ATTRIBUTES")
             ).addToHolder(holder)
         } else if (!functionType.variadic && realCount > expectedCount) {
             args.exprList.drop(expectedCount).forEach {
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/FillFunctionArgumentsFix.kt
@@ -9,11 +9,17 @@ import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiErrorElement
 import com.intellij.psi.PsiFile
 import com.intellij.psi.util.parentOfType
 import org.rust.ide.annotator.getFunctionCallContext
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.RsElementTypes.COMMA
+import org.rust.lang.core.psi.RsElementTypes.RPAREN
+import org.rust.lang.core.psi.ext.RsElement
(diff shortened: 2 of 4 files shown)


ortem:
Hi @Kobzol, I've found the current implementation doesn't work if I invoke `Fill missing arguments` after adding multiple parameters.

Consider adding multiple parameters to `foo`:
```Rust
fn foo() {}
fn main() {
    foo();
}
```
After `Change Signature`, I got this:
```Rust
fn foo(p0: i32, p1: bool, p2: char) {}
fn main() {
    foo(, , ); // invoke `Fill missing arguments`
}
```
After invoking `Fill missing arguments`, I got broken code:
```Rust
fn foo(p0: i32, p1: bool, p2: char) {}
fn main() {
    foo(0, false, ''), , );
}
```

ortem (changes requested):
Hi @Kobzol, I've found the current implementation doesn't work if I invoke `Fill missing arguments` after adding multiple parameters.

Consider adding multiple parameters to `foo`:
```Rust
fn foo() {}
fn main() {
    foo();
}
```
After `Change Signature`, I got this:
```Rust
fn foo(p0: i32, p1: bool, p2: char) {}
fn main() {
    foo(, , ); // invoke `Fill missing arguments`
}
```
After invoking `Fill missing arguments`, I got broken code:
```Rust
fn foo(p0: i32, p1: bool, p2: char) {}
fn main() {
    foo(0, false, ''), , );
}
```

Kobzol:
Hmm, if the call has no arguments and the comma directly follows the opening parenthesis, it parses like this :disappointed:
```rust
foo(,,);
```
image

Maybe it would be easier to change the parser after all?

Kobzol:
@ortem Now that #7019 improved the parsing of function call arguments, I wonder if the error range of the `FillFunctionArgumentsFix` should change.

Currently, if there is not enough arguments, only the right parenthesis is annotated. This was done in https://github.com/intellij-rust/intellij-rust/pull/5612 to make the annotation more precise to let the user see quickly what is wrong:
```rust
fn foo(a: u32, b: u32, c: u32) {}

fn main() {
    foo(1<error>)</error>;
}
```

However, there is another case when this fix is applicable, and that's the situation where there are arguments, but they are "empty" (i.e. just whitespace/comments between commas):
```rust
foo(1 , , );
```
In this case it's probably not ideal to just mark the right parenthesis? Maybe we again go back to either marking the whole argument list or marking the individual commas?

ortem:
@Kobzol I think marking the individual comma would be nice but not obligatory in this PR

Kobzol:
Ok. I tried rebasing onto #7019 to reimplement the quick fix, but the parser is behaving a bit strangely, so I'll probably wait until #7019 is fully resolved and merged.

ortem:
@Kobzol have you tried running `Generate Parser` Gradle task? If there are any problems with #7019 implementation, please mention them in the PR: probably they should be fixed before merging

Kobzol:
I have regenerated the parser just to be sure. To be honest, I'm not sure if it's the parser or if I'm doing something wrong. I pushed a WIP version of my code (a simplified version that build a new argument list instead of modifying the original one).

When I run `FillFunctionArgumentsFixTest::test ufcs self value`, in `FillFunctionArgumentsFix:98` a new argument list is being created out of two expressions (`S` and `0`). However, the newly created argument list will have no expressions and its text simply equals `(`. CC @vlad20012

vlad20012:
Some tests in #7019 are failed. I'll take a look

Kobzol:
Great, after the parser fix it now works perfectly! There are still some implementation questions though. I have basically two implementations working, one is creating a new argument list and then replacing the old one, and another one inserts the missing arguments directly into the old list.

Creating a new argument was already done in the original fix implementation, it has simpler code and it also "normalizes" the argument list, e.g. in the case when the user inserted too many commas:
```rust
fn foo(a: u32, b: u32, c: u32) {}

foo(1,,,3);
v
foo(1, 0, 0);      // replace
foo(1, 0, 0,,3);  // insert
```
But it also removes any potential formatting/comments in the argument list (the original fix did the same). Is that OK?

Also for the error annotation, I would need something similar as in https://github.com/intellij-rust/intellij-rust/pull/6745. Basically I want to enable the quick fix in the whole argument range, but I only want to mark the end parenthesis with an error annotation. But I had some problems with that (https://github.com/intellij-rust/intellij-rust/pull/6745#issuecomment-797706690).

Kobzol:
I rebased over master and changed the highlighting. Now the whole argument range offers the quick fix, but only the right parenthesis gets error annotation to reduce visual clutter. The changes in parsing also add additional error highlight to successive commas, which gets us the best of both worlds I think.

Can you take another look please @ortem? :) If you like it, I'll re-bless tests.
fill-arguments

ortem (approved):
LGTM now. @Kobzol thank you!

Kobzol:
Great! Rebased over master and reblessed tests.

bors r=ortem

PR #6731, opened 2021-01-27
RES: Enable new name resolution engine on nightly by default

dima74:
This allows us to gather more feedback and find potential issues

Should be merged after #6719

changelog: Enable new name resolution engine on the nightly plugin channel by default

PR #6794, opened 2021-02-03
T: Replace `@IgnoreInNewResolve` with `@UseOldResolve`

dima74:
As a consequence of #6731, [new resolve](https://github.com/intellij-rust/intellij-rust/issues/6217) was enabled by default on CI and for local development. There are few tests which are currently not supported by new resolve, so it makes sense to force use old resolve for them (instead of ignoring them in new resolve)

### DIFF
--- a/src/test/kotlin/org/rust/cargo/RsWithToolchainTestBase.kt
+++ b/src/test/kotlin/org/rust/cargo/RsWithToolchainTestBase.kt
@@ -48,7 +48,7 @@ abstract class RsWithToolchainTestBase : CodeInsightFixtureTestCase<ModuleFixtur
     }
 
     override fun runTestRunnable(testRunnable: ThrowableRunnable<Throwable>) {
-        val skipReason = rustupFixture.skipTestReason ?: getIgnoredInNewResolveReason(project)
+        val skipReason = rustupFixture.skipTestReason
         if (skipReason != null) {
             System.err.println("SKIP \"$name\": $skipReason")
             return
@@ -77,6 +77,7 @@ abstract class RsWithToolchainTestBase : CodeInsightFixtureTestCase<ModuleFixtur
         if (disableMissedCacheAssertions) {
             RecursionManager.disableMissedCacheAssertions(testRootDisposable)
         }
+        setupResolveEngine(project, testRootDisposable)
     }
 
     override fun tearDown() {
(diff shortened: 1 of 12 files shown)


src/test/kotlin/org/rust/cargo/RsWithToolchainTestBase.kt
```diff
@@ -77,6 +78,9 @@ abstract class RsWithToolchainTestBase : CodeInsightFixtureTestCase<ModuleFixtur
         if (disableMissedCacheAssertions) {
             RecursionManager.disableMissedCacheAssertions(testRootDisposable)
         }
+        findAnnotationInstance<UseOldResolve>()?.let {
+            DefMapService.setNewResolveEnabled(project, testRootDisposable, false)
+        }
```
Undin, line 83: Why only `UseOldResolve`?
I'd extract code from `RsTestBase` for both annotations to util function and use it here as well

src/test/kotlin/org/rustPerformanceTests/RsBuildDefMapTest.kt
```diff
@@ -31,7 +31,7 @@ class RsBuildDefMapTest : RsRealProjectTestBase() {
     fun `test build juniper`() = doTest(JUNIPER)
 
     private fun doTest(info: RealProjectInfo) {
-        DefMapService.setUseNewResolve(project, testRootDisposable)
+        DefMapService.setNewResolveEnabled(project, testRootDisposable, true)
```
Undin, line 34: Let's use `UseNewResolve` instead not to use two different ways to do something in different places

src/test/kotlin/org/rustPerformanceTests/RsProfileBuildDefMapTest.kt
```diff
@@ -23,7 +23,7 @@ class RsProfileBuildDefMapTest : RsPerformanceTestBase() {
     fun `test juniper`() = doTest(JUNIPER)
 
     private fun doTest(info: RealProjectInfo) {
-        DefMapService.setUseNewResolve(project, testRootDisposable)
+        DefMapService.setNewResolveEnabled(project, testRootDisposable, true)
```
Undin, line 26: The same about `UseNewResolve `

Undin (approved):
👍 
bors r+

Issue #6793, opened 2021-02-03
Change signature refactoring works unexpectedly for function and method calls


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.3645-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3.2 Ultimate Edition (IU-203.7148.57)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
If I invoke `Change Signature` refactoring on function/method calls, I expect that IDE suggests me to change signature of this function/method (at least, other languages like Java and Kotlin use such behavior). But currently, the plugin suggests me to change signature of parent function.

![2021-02-03 12 10 32](https://user-images.githubusercontent.com/2539310/106724289-18ba2500-6619-11eb-8fa8-5dac66233d45.gif)

## Steps to reproduce
```rust
fn main() {
    foo/*caret*/(1);
    println!("hello!")
}



fn foo(x: i32) -> i32 {
    x
}
```

PR #6795, opened 2021-02-03
REF: allow applying the Change Signature refactoring to function calls

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/6793

changelog: Apply `Change Signature` refactoring to the called function when invoked on function and method calls.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureHandler.kt
@@ -16,7 +16,7 @@ import com.intellij.refactoring.RefactoringBundle
 import com.intellij.refactoring.changeSignature.ChangeSignatureHandler
 import com.intellij.refactoring.util.CommonRefactoringUtil
 import org.rust.RsBundle
-import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.openapiext.editor
 import org.rust.openapiext.elementUnderCaretInEditor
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureHandler.kt
```diff
@@ -16,15 +16,26 @@ import com.intellij.refactoring.RefactoringBundle
...
+        val function = element.ancestorOrSelf<RsFunction>(stopAt = RsBlock::class.java)
+        if (function is RsFunction) return function
+
+        return element.ancestors.mapNotNull { el ->
+            when (el) {
+                is RsCallExpr -> (el.expr as? RsPathExpr)?.path?.reference?.resolve() as? RsFunction
+                is RsMethodCall -> el.reference.resolve() as? RsFunction
+                else -> null
+            }
+        }.firstOrNull()
```
Undin, line 37: This code works as expected for valid code when the plugin is able to provide proper name resolution. But if the plugin failed to resolve function/method call, we will try to find parent call that's quite unexpected.

For example,
```rust
fn main() {
    bar(foo/*caret*/(1));
}

fn bar(x: i32) {}
```

I expect that IDE shows me some warning (because `foo` doesn't exist and cannot be refactored) but currently we start to refactor `bar`

Undin: I suppose we can fix this issue and generalize current implementation in the following way:
```kotlin
override fun findTargetMember(element: PsiElement): RsFunction? {
    for (el in element.ancestors) {
        return when (el) {
            is RsFunction -> el
            is RsCallExpr -> (el.expr as? RsPathExpr)?.path?.reference?.resolve() as? RsFunction
            is RsMethodCall -> el.reference.resolve() as? RsFunction
            is RsBlock -> return null
            else -> continue
        }
    }
    return null
}
```

Undin (changes requested):

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureHandler.kt
Kobzol: Good point. I changed the implementation and added a corresponding test.

Undin (approved):
Great!
Thank you!
bors r+

PR #6796, opened 2021-02-03
CI: Retry failed tests

dima74:
Failed test will be retried at most 3 times. This is needed to mitigate flaky tests. If there are more then 20 failed test, no test will be retried. Implemented by [`test-retry-gradle-plugin`](https://github.com/gradle/test-retry-gradle-plugin).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -52,6 +52,7 @@ plugins {
     id("org.jetbrains.intellij") version "0.6.3"
     id("org.jetbrains.grammarkit") version "2020.2.1"
     id("net.saliman.properties") version "1.5.1"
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -113,6 +115,12 @@ allprojects {
 
         test {
             testLogging.showStandardStreams = prop("showStandardStreams").toBoolean()
+            if (CI) {
+                retry {
+                    maxRetries.set(3)
+                    maxFailures.set(20)
```
Undin, line 121: I suppose `5` is enough threshold not to retry. Usually, only one test fail

Undin (approved):
👍

dima74:
bors r=Undin

PR #6799, opened 2021-02-04
COMP: Add completion by subwords for dependencies

avrong:
### TODO:
- [x] Add specific completion tests

Related: #6463

changelog: Complete dependencies by subwords when using experimental local index


toml/src/test/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProviderTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    fun `test subwords completion`() = doSingleCompletion("""
+        [dependencies]
+        f-ba<caret>
+    """, """
+        [dependencies]
+        foo_bar = "1.0.0"<caret>
+    """, crate("foo_bar", "1.0.0"))
+
+    private fun crate(name: String, vararg versions: String): Crate =
```
Kobzol, line 40: This is used also in other tests. I extracted a base test class to avoid duplication here: https://github.com/intellij-rust/intellij-rust/pull/6801/files#diff-e97b3231eba5cb633cf03dfed0b9a1ece02d7403da95fbc3390a92fe7e765615R19. However, it's inspection specific. Maybe we should introduce a kind of a "mixin" test base class which would contain the `crate` function and the `Crate` class and a function that would run the passed callback in
```kotlin
runWithEnabledFeature(RsExperiments.CRATES_LOCAL_INDEX) {
    withMockedCrates(crateMap) {
        ...
    }
}
```
?

avrong: Introduced `LocalCargoTomlCompletionTestBase` as a completion test base for crates local index.

There still will be some code duplication between base-classes and their methods, not sure what to do with this right now

Kobzol: In theory, the mixin base class could be an interface with default method implementations. Or it could be a superclass of both completion, i.e. put into some `RsTomlTestBase`. But I'm not sure if it's worth it.

toml/src/test/kotlin/org/rust/toml/completion/LocalCargoTomlCompletionTestBase.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import org.intellij.lang.annotations.Language
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.runWithEnabledFeature
+import org.rust.toml.crates.local.CargoRegistryCrate
+import org.rust.toml.crates.local.CargoRegistryCrateVersion
+import org.rust.toml.crates.local.withMockedCrates
+
+abstract class LocalCargoTomlCompletionTestBase : CargoTomlCompletionTestBase() {
+    protected fun crate(name: String, vararg versions: String): Crate =
```
ortem, line 16: What do you think about using a static `@TestOnly` method instead of this function here and in `CrateNotFoundInspectionTest`? For example, something like:
```Kotlin
data class CargoRegistryCrate(val versions: List<CargoRegistryCrateVersion>) {
    companion object {
        @TestOnly
        fun of(vararg versions: String): CargoRegistryCrate =
            CargoRegistryCrate(versions.map {
                CargoRegistryCrateVersion(it, false, listOf())
            })
    }
}
```
and use `vararg crates: Pair<String, CargoRegistryCrate>` instead of `vararg crates: Crate`

toml/src/test/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProviderTest.kt
```diff
@@ -0,0 +1,37 @@
...
+    """, crate("foo_bar", "1.0.0"))
+
+    fun `test subwords completion`() = doSingleCompletion("""
+        [dependencies]
+        f-ba<caret>
+    """, """
+        [dependencies]
+        foo_bar = "1.0.0"<caret>
+    """, crate("foo_bar", "1.0.0"))
+}
```
ortem, line 37: Maybe provide some tests with multiple versions of the same crate to check that the latest version is inserted?

ortem (changes requested):

toml/src/test/kotlin/org/rust/toml/completion/LocalCargoTomlDependencyCompletionProviderTest.kt
avrong: Right now it wouldn't be that useful, as without semantic versioning sorting it just takes the last one, but I did add one.

Also, I'm still thinking that maybe we should insert the name only in the name completion to decrease the number of situations when completion leads to invalid code. Then, such tests will be performed in version intention/completion tests

ortem (approved):
LGTM
bors r+

PR #6801, opened 2021-02-04
TOML: add CrateVersionInvalidInspection

Kobzol:
This PR adds a TOML inspection that checks if dependency version requirements match any known versions. It can also possibly detect yanked versions (or it could be in a separate inspection).

Since this is the second "crate-level" inspection (after https://github.com/intellij-rust/intellij-rust/pull/6591), I separated the crate dependency visitor and introduced shared abstractions for Cargo.toml inspections (and their tests) in separate commits.

For parsing the version requirements, I use the same library as in https://github.com/intellij-rust/intellij-rust/pull/6599. I created a lot of basic tests to check that it works correctly. It mostly seems to work, but sadly, there are (at least) two cases where its behaviour differs from Cargo:
```
1.2.*
Cargo does not accept 1.3.0
Semver4J accepts 1.3.0

>=1.2
Cargo does not accept 1.1.2
Semver4J accepts 1.1.2
```
The second case can be "fixed" if we add patch version to the requirement (`>=1.2.0`), I don't know how to fix the first one.

Possible workarounds for this:
1) Implement some workarounds for wildcard and range version requirements to make them work in Semver4J
2) Don't do anything (Semver4J matches more versions than Cargo, so it should not produce false positives in this case).
3) Implement version matching from scratch (:disappointed:).

(Probably) blocked on: https://github.com/intellij-rust/intellij-rust/pull/6599
Closes: https://github.com/intellij-rust/intellij-rust/pull/3785
Related issue: https://github.com/intellij-rust/intellij-rust/issues/6463

changelog: Add inspection that checks crate dependency version requirements in Cargo.toml that do not match any known crate version.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlInspectionToolBase.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlInspectionToolBase.kt
@@ -1,0 +1,29 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.inspections
+
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.PsiElementVisitor
+import org.rust.cargo.CargoConstants
+import org.rust.ide.experiments.RsExperiments
+import org.rust.openapiext.isFeatureEnabled
+
+abstract class CargoTomlInspectionToolBase : TomlLocalInspectionToolBase() {
+    open val requiresLocalCrateIndex: Boolean = false
+
+    abstract fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor?
+
+    override fun buildVisitorInternal(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor? {
+        if (requiresLocalCrateIndex && !isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
+            return super.buildVisitor(holder, isOnTheFly)
+        }
+        return if (holder.file.name != CargoConstants.MANIFEST_FILE) {
+            super.buildVisitor(holder, isOnTheFly)
+        } else {
+            buildCargoTomlVisitor(holder)
+        }
+    }
+}
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlInspectionToolBase.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlInspectionToolBase.kt
@@ -10,20 +10,18 @@ import com.intellij.psi.PsiElementVisitor
 import org.rust.cargo.CargoConstants
 import org.rust.ide.experiments.RsExperiments
 import org.rust.openapiext.isFeatureEnabled
+import org.toml.lang.psi.TomlVisitor
 
 abstract class CargoTomlInspectionToolBase : TomlLocalInspectionToolBase() {
     open val requiresLocalCrateIndex: Boolean = false
 
-    abstract fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor?
+    abstract fun buildCargoTomlVisitor(holder: ProblemsHolder): TomlVisitor
 
     override fun buildVisitorInternal(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor? {
-        if (requiresLocalCrateIndex && !isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
-            return super.buildVisitor(holder, isOnTheFly)
-        }
-        return if (holder.file.name != CargoConstants.MANIFEST_FILE) {
-            super.buildVisitor(holder, isOnTheFly)
-        } else {
-            buildCargoTomlVisitor(holder)
+        return when {
+            requiresLocalCrateIndex && !isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX) -> null
+            holder.file.name != CargoConstants.MANIFEST_FILE -> null
+            else -> buildCargoTomlVisitor(holder)
         }
     }
 }
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
@@ -18,35 +18,32 @@ import org.toml.lang.psi.ext.TomlLiteralKind
 import org.toml.lang.psi.ext.kind
 import org.toml.lang.psi.ext.name
 
-class CrateNotFoundInspection : TomlLocalInspectionToolBase() {
-    override fun buildVisitorInternal(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor? {
-        if (!isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) return null
-        if (holder.file.name != CargoConstants.MANIFEST_FILE) return null
+class CrateNotFoundInspection : CargoTomlInspectionToolBase() {
+    override val requiresLocalCrateIndex: Boolean = true
 
-        return object : TomlVisitor() {
-            override fun visitKeyValue(element: TomlKeyValue) {
-                val table = element.parent as? TomlTable ?: return
-                val depTable = DependencyTable.fromTomlTable(table) ?: return
-                if (depTable !is DependencyTable.General) return
+    override fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor = object : TomlVisitor() {
+        override fun visitKeyValue(element: TomlKeyValue) {
+            val table = element.parent as? TomlTable ?: return
+            val depTable = DependencyTable.fromTomlTable(table) ?: return
+            if (depTable !is DependencyTable.General) return
 
-                val segment = element.key.segments.getOrNull(0) ?: return
-                val name = segment.name ?: return
-                val value = element.value ?: return
-                if (value is TomlLiteral && value.kind is TomlLiteralKind.String) {
-                    handleDependency(DependencyCrate(name, segment, mapOf("version" to value)), holder)
-                } else if (value is TomlInlineTable) {
-                    handleDependency(DependencyCrate(name, segment, collectProperties(value)), holder)
-                }
+            val segment = element.key.segments.getOrNull(0) ?: return
+            val name = segment.name ?: return
+            val value = element.value ?: return
+            if (value is TomlLiteral && value.kind is TomlLiteralKind.String) {
+                handleDependency(DependencyCrate(name, segment, mapOf("version" to value)), holder)
+            } else if (value is TomlInlineTable) {
+                handleDependency(DependencyCrate(name, segment, collectProperties(value)), holder)
             }
+        }
 
-            override fun visitTable(element: TomlTable) {
-                val depTable = DependencyTable.fromTomlTable(element) ?: return
-                if (depTable !is DependencyTable.Specific) return
+        override fun visitTable(element: TomlTable) {
+            val depTable = DependencyTable.fromTomlTable(element) ?: return
+            if (depTable !is DependencyTable.Specific) return
 
-                handleDependency(
-                    DependencyCrate(depTable.crateName, depTable.crateNameElement, collectProperties(element)), holder
-                )
-            }
+            handleDependency(
+                DependencyCrate(depTable.crateName, depTable.crateNameElement, collectProperties(element)), holder
+            )
         }
     }
 
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
@@ -12,101 +12,24 @@ import org.rust.ide.experiments.RsExperiments
 import org.rust.openapiext.isFeatureEnabled
 import org.rust.toml.crates.local.CratesLocalIndexException
 import org.rust.toml.crates.local.CratesLocalIndexService
-import org.rust.toml.isDependencyKey
-import org.toml.lang.psi.*
-import org.toml.lang.psi.ext.TomlLiteralKind
-import org.toml.lang.psi.ext.kind
-import org.toml.lang.psi.ext.name
 
 class CrateNotFoundInspection : CargoTomlInspectionToolBase() {
-    override val requiresLocalCrateIndex: Boolean = true
+    override val requiresLocalCrateIndex = true
 
-    override fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor = object : TomlVisitor() {
-        override fun visitKeyValue(element: TomlKeyValue) {
-            val table = element.parent as? TomlTable ?: return
-            val depTable = DependencyTable.fromTomlTable(table) ?: return
-            if (depTable !is DependencyTable.General) return
+    override fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor {
+        return object : CargoDependencyCrateVisitor() {
+            override fun visitDependency(dependency: DependencyCrate) {
+                if (dependency.isForeign()) return
 
-            val segment = element.key.segments.getOrNull(0) ?: return
-            val name = segment.name ?: return
-            val value = element.value ?: return
-            if (value is TomlLiteral && value.kind is TomlLiteralKind.String) {
-                handleDependency(DependencyCrate(name, segment, mapOf("version" to value)), holder)
-            } else if (value is TomlInlineTable) {
-                handleDependency(DependencyCrate(name, segment, collectProperties(value)), holder)
-            }
-        }
-
-        override fun visitTable(element: TomlTable) {
-            val depTable = DependencyTable.fromTomlTable(element) ?: return
-            if (depTable !is DependencyTable.Specific) return
-
-            handleDependency(
-                DependencyCrate(depTable.crateName, depTable.crateNameElement, collectProperties(element)), holder
-            )
-        }
-    }
-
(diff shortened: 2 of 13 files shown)


avrong:
> 1.2.*
> Cargo does not accept 1.3.0
> Semver4J accepts 1.3.0
> 
> >=1.2
> Cargo does not accept 1.1.2
> Semver4J accepts 1.1.2

There is also a problem with supporting versions like `999999999.999999999.9999999991` and `0.1.20190409163015`. Suddenly, semver4j can't fit them into `Int` which it uses internally.

There is a bigger chance we will need to think about implementing something on our own

Kobzol:
> There is also a problem with supporting versions like 999999999.999999999.9999999991 and 0.1.20190409163015. Suddenly, semver4j can't fit them into Int which it uses internally.

This seems like less of an issue to me :D But you're right, we'll probably have to implement it from scratch, based on Cargo's implementation. At least we have some basic tests now, I'll try to come up with an implementation.

Kobzol:
I wonder, now that the plugin packs a native Rust program, could we leverage it to do the following?
1) Use the https://github.com/steveklabnik/semver crate in the native crate
2) Create a shared library that would expose a few functions of this crate with a C interface
3) Create a Java JNI wrapper that would call these methods.

In this way we would not have to reimplement the functionality and use the exact same implementation as Cargo uses.
WDYT @vlad20012? Is the idea too crazy? :)

vlad20012:
@Kobzol, haha, it's too crazy, I guess =D
 I really like extending our Rust part usage, but usages like _parsing_ are overkill, I guess. Chalk/Polonius is considerable, for example.

Kobzol:
Ok, I understand :) Then I guess we can copy the already existing implementation from https://github.com/intellij-rust/intellij-rust/pull/3785/files.

ortem:
@Kobzol @avrong 

> 1.2.*
> Cargo does not accept 1.3.0
> Semver4J accepts 1.3.0

I've read your implementation briefly and found that you add a caret `^` to the version if the first character is a digit. However, the docs you mentioned tells "Since this string **does not have any operators** in it, it is interpreted the same way as if we had <...> a caret requirement". So I suppose we should not add a caret if there is a wildcard already. Note that comparison works fine without the caret:
Screenshot 2021-03-15 at 12 14 29



> \>=1.2
> Cargo does not accept 1.1.2
> Semver4J accepts 1.1.2
> ... can be "fixed" if we add patch version to the requirement (>=1.2.0)

Absolutely! So does everything work fine if we just normalize any versions by adding omitted minor and patch, like `x --> x.0.0` and `x.y --> x.y.0`? BTW it looks like Semver4J implementation bug (`Requirement.buildNPM` returns `LOOSE` versions sometimes), but anyway not critical for us I suppose:
Screenshot 2021-03-15 at 12 39 15


> There is also a problem with supporting versions like 999999999.999999999.9999999991 and 0.1.20190409163015. Suddenly, semver4j can't fit them into Int which it uses internally.

I found only four crates with such versions, with ~3K downloads all time in total:
* [susu](https://crates.io/crates/susu)
* [susudb](https://crates.io/crates/susudb)
* [test-max-version-example-crate](https://crates.io/crates/test-max-version-example-crate)
* [version-length-checking-is-overrated](https://crates.io/crates/version-length-checking-is-overrated)

Please correct me if there are other (and more popular) crates with so long versions. But I think we can just ignore them for now.

Finally, IMO we can just use Semver4J if there are no other problems.

Kobzol:
I also think that long versions can be ignored.

> I've read your implementation briefly and found that you add a caret ^ to the version if the first character is a digit. However, the docs you mentioned tells "Since this string does not have any operators in it, it is interpreted the same way as if we had <...> a caret requirement". So I suppose we should not add a caret if there is a wildcard already. Note that comparison works fine without the caret:

The problem is that if I don't add the caret, `1.2.3` requirement will not match `1.2.4` version. But maybe we could add `^` only if the version requirement has no wildcard? It seems to solve this specific case.

I rebased the code over master and added a very hacky way of adding `.0` to the end of version requirements. Along with the caret if, it was enough to fix the current tests.

Kobzol:
@ortem Now that https://github.com/intellij-rust/intellij-rust/pull/6599 is being merged, what about this PR? Should we roll our own parser or just hack around semver4j?

avrong:
@Kobzol, we decided to keep `semver4j` and use some workaround in the corner cases.
In #6599, the versions that cannot be parsed are being ignored as only a few of them are in the registry

Kobzol:
Ok, good :) Rebased over master.

toml/src/test/kotlin/org/rust/toml/inspections/CrateVersionInvalidInspectionTest.kt
```diff
@@ -0,0 +1,301 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.inspections
+
+import org.rust.toml.crates.local.CargoRegistryCrate
+import org.rust.toml.crates.local.CargoRegistryCrateVersion
+
+class CrateVersionInvalidInspectionTest : CargoTomlCrateInspectionTestBase(CrateVersionInvalidInspection::class) {
+    fun `test do not warn about missing crate`() = doTest(
```
avrong, line 12: Let's reformat it such way:
```kotlin
fun `test do not warn about missing crate`() = doTest("""
        [dependencies]
        foo = "1"
    """)
```

toml/src/test/kotlin/org/rust/toml/inspections/CrateNotFoundInspectionTest.kt
```diff
@@ -5,40 +5,45 @@
...
-import org.rust.ide.experiments.RsExperiments
-import org.rust.ide.inspections.RsInspectionsTestBase
-import org.rust.openapiext.runWithEnabledFeatures
 import org.rust.toml.crates.local.CargoRegistryCrate
-import org.rust.toml.crates.local.withMockedCrates
 
-class CrateNotFoundInspectionTest : RsInspectionsTestBase(CrateNotFoundInspection::class) {
-    fun `test missing crate in dependencies`() = doTest("""
+class CrateNotFoundInspectionTest : CargoTomlCrateInspectionTestBase(CrateNotFoundInspection::class) {
+    fun `test missing crate in dependencies`() = doTest(
```
avrong, line 11: Let's keep previous formatting in this file

toml/src/main/resources/META-INF/toml-only.xml
```diff
@@ -31,6 +31,10 @@
                          displayName="Crate not found"
                          enabledByDefault="true" level="WARNING"
                          implementationClass="org.rust.toml.inspections.CrateNotFoundInspection"/>
+        <localInspection language="TOML" groupName="Rust"
```
avrong, line 34: Probably it makes sense here to have `Cargo.toml` dependency inspections in a separate group:
```suggestion
        <localInspection language="Rust" groupPath="Rust,Cargo.toml" groupName="Dependencies"
```

toml/src/main/kotlin/org/rust/toml/inspections/CrateVersionInvalidInspection.kt
```diff
@@ -0,0 +1,44 @@
...
+class CrateVersionInvalidInspection : CargoTomlInspectionToolBase() {
+    override val requiresLocalCrateIndex = true
+
+    override fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor {
+        return object : CargoDependencyCrateVisitor() {
+            override fun visitDependency(dependency: DependencyCrate) {
+                if (dependency.isForeign()) return
+
+                val service = CratesLocalIndexService.getInstance()
+                val crate = service.getCrate(dependency.crateName) ?: return
```
avrong, line 24: On `master`, `CratesLocalIndexService::getCrate(crateName: String)` now can throw `CratesLocalIndexException` in cases when index is not available when it either being updated or not available at all. Let's ignore inspections when the exception is caught

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -5,105 +5,22 @@
...
+                if (dependency.isForeign()) return
 
-                val segment = element.key.segments.getOrNull(0) ?: return
-                val name = segment.name ?: return
-                val value = element.value ?: return
-                if (value is TomlLiteral && value.kind is TomlLiteralKind.String) {
-                    handleDependency(DependencyCrate(name, segment, mapOf("version" to value)), holder)
-                } else if (value is TomlInlineTable) {
-                    handleDependency(DependencyCrate(name, segment, collectProperties(value)), holder)
+                if (CratesLocalIndexService.getInstance().getCrate(dependency.crateName) == null) {
```
avrong, line 20: Let's catch `CratesLocalIndexException`

avrong (changes requested):

toml/src/main/resources/META-INF/toml-only.xml
Kobzol: Hmm, if I put `Rust` there, the inspection tests no longer invoke the crate not found/invalid inspections. Maybe it won't run Rust inspections in `.toml` files?

avrong: Oh, sure 😅
```
<localInspection language="TOML" groupPath="Rust,Cargo.toml" groupName="Dependencies"
```

Kobzol:
I have to remember to avoid using `Ctrl + Shift + I` in test files, it breaks formatting terribly :)

toml/src/main/resources/META-INF/toml-only.xml
Undin: Are we sure about `Rust/Cargo.toml/Dependencies` ? Maybe `Rust/Cargo.toml` is enough for now?
We don't have a lot of inspections for Cargo.toml and suggested structure maybe too complicated
image

toml/src/main/kotlin/org/rust/toml/inspections/CrateNotFoundInspection.kt
```diff
@@ -7,109 +7,26 @@ package org.rust.toml.inspections
...
-                dependencyNameIndex != -1 -> {
-                    val crate = segments.getOrNull(dependencyNameIndex + 1)
-                    val crateName = crate?.name
-                    if (crate != null && crateName != null) {
-                        Specific(crateName, crate)
-                    } else {
-                        null
-                    }
+                if (crate == null) {
+                    holder.registerProblem(dependency.crateNameElement, "Crate ${dependency.crateName} not found")
```
ortem, line 28: [minor] Let's introduce `val crateName = dependency.crateName` to replace all 3 `dependency.crateName` occurrences

toml/src/main/resources/META-INF/toml-only.xml
```diff
@@ -27,10 +27,20 @@
                          displayName="Cyclic feature dependency"
                          enabledByDefault="true" level="ERROR"
                          implementationClass="org.rust.toml.inspections.CargoTomlCyclicFeatureInspection"/>
-        <localInspection language="TOML" groupName="Rust"
+        <localInspection language="TOML"
                          displayName="Crate not found"
-                         enabledByDefault="true" level="WARNING"
+                         groupPath="Rust,Cargo.toml"
+                         groupName="Dependencies"
```
ortem, line 33: Looks like https://github.com/intellij-rust/intellij-rust/pull/6801#discussion_r611579058 is still relevant

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlInspectionToolBase.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.rust.openapiext.isFeatureEnabled
+
+abstract class CargoTomlInspectionToolBase : TomlLocalInspectionToolBase() {
+    open val requiresLocalCrateIndex: Boolean = false
+
+    abstract fun buildCargoTomlVisitor(holder: ProblemsHolder): PsiElementVisitor?
+
+    override fun buildVisitorInternal(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor? {
+        if (requiresLocalCrateIndex && !isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
+            return super.buildVisitor(holder, isOnTheFly)
```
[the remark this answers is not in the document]
Kobzol: Thanks!

Kobzol:
It's too much work rebasing these 4 commits one-by-one, so I pushed the changes in a separate commit, it also makes review easier I suppose. I'll rebase it before final merge.

toml/src/main/kotlin/org/rust/toml/crates/local/CrateVersion.kt
```diff
@@ -0,0 +1,51 @@
...
+            if (parsed.size != requirements.size) return null
+
+            return CrateVersionRequirement(parsed)
+        }
+    }
+}
+
+/**
+ * Turns 1.2.3 into ^1.2.3.
+ * See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
```
ortem, line 36:
```suggestion
  *
  * See [Cargo Reference](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)
```
to get a clickable link in the rendered view

toml/src/main/kotlin/org/rust/toml/crates/local/CrateVersion.kt
```diff
@@ -0,0 +1,51 @@
...
+            }
+            if (parsed.size != requirements.size) return null
+
+            return CrateVersionRequirement(parsed)
+        }
+    }
+}
+
+/**
+ * Turns 1.2.3 into ^1.2.3.
```
ortem, line 35: Let's add more examples here, at least about adding `.0`

toml/src/main/kotlin/org/rust/toml/crates/local/CrateVersion.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+private fun normalizeVersion(version: String): String {
+    var normalized = version
+    if (normalized.firstOrNull() in listOf('<', '>', '=')) {
+        while (normalized.count { it == '.' } < 2) {
+            normalized += ".0"
+        }
+    }
+
+    return if (normalized.getOrNull(0)?.isDigit() == true && !normalized.contains("*")) {
```
ortem, line 46: `getOrNull(0)` is equivalent to `firstOrNull()`
BTW empty `version` can be handled separately at the very beginning to avoid nullability checking

toml/src/main/kotlin/org/rust/toml/crates/local/CrateVersion.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+/**
+ * Turns 1.2.3 into ^1.2.3.
+ * See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
+ */
+private fun normalizeVersion(version: String): String {
+    var normalized = version
+    if (normalized.firstOrNull() in listOf('<', '>', '=')) {
```
ortem, line 40: Let's add some comments to describe what's going here and why it is necessary. It'd be great to have two comments, about `.0` and `^` respectively

toml/src/main/kotlin/org/rust/toml/crates/local/CrateVersion.kt
```diff
@@ -0,0 +1,51 @@
+/*
```
ortem, line 1: This file should be renamed since there is no `CrateVersion` class after all

ortem (changes requested):
I found one false-positive in [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/syntax/Cargo.toml#L17). Please fix it in this PR before merging or open a new issue after (so as not to forget fixing it before enabling the index by default)

Kobzol:
What false positive have you found? I tried your example and I don't see any warning.

Kobzol:
Ah, good catch, this was actually a bug present also in `CrateNotFoundInspection`, I didn't handle renamed packages in the dependency visitor.

avrong:
While entering random versions such as `rocket = "0.4.0-rc--3-42=34=-23=-42="` I keep getting NPEs from semver4j. It is better to handle them to prevent IDE from showing them as errors.

```
java.lang.NullPointerException
	at com.vdurmont.semver4j.Semver.<init>(Semver.java:26)
	at com.vdurmont.semver4j.Requirement.hyphenRequirement(Requirement.java:487)
	at com.vdurmont.semver4j.Requirement.evaluateReversePolishNotation(Requirement.java:338)
	at com.vdurmont.semver4j.Requirement.buildWithTokenizer(Requirement.java:138)
	at com.vdurmont.semver4j.Requirement.buildNPM(Requirement.java:112)
	at org.rust.toml.crates.local.CrateVersionRequirement$Companion.build(CrateVersionRequirement.kt:22)
	at org.rust.toml.inspections.CrateVersionInvalidInspection$buildCargoTomlVisitor$1.visitDependency(CrateVersionInvalidInspection.kt:32)
	at org.rust.toml.inspections.CargoDependencyCrateVisitor.visitKeyValue(CargoDependencyCrateVisitor.kt:28)
	at org.toml.lang.psi.impl.TomlKeyValueImpl.accept(Psi.kt:30)

```

Kobzol:
Oh, that's messy, thank you. I wonder if we should show a warning if the version requirement cannot be parsed? It fits into this inspection I guess.

avrong:
Let's do it, as a weak warning. Seems like there are not a lot of such cases

Kobzol:
I was hoping you'd say that :D

avrong (approved):
LGTM
Please squash your commits and we will be ready to merge.

Kobzol:
Is it OK like this or do you want me to squash into a single commit?

avrong:
Thank you!
bors r+

PR #6803, opened 2021-02-05
TOML: check TOML plugin compatibility in CrateNotFoundInspection

Undin:
Otherwise, the inspection may throw exceptions when PSI in TOML plugin is changed

Also, introduce `TomlLocalInspectionToolBase` not to forget to check TOML plugin compatibility.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/CargoTomlCyclicFeatureInspection.kt
@@ -5,12 +5,10 @@
 
 package org.rust.toml.inspections
 
-import com.intellij.codeInspection.LocalInspectionTool
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.psi.PsiElementVisitor
 import org.rust.cargo.CargoConstants
 import org.rust.toml.isFeatureListHeader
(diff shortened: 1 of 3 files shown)


Undin:
CC @avrong @Kobzol @ortem

Kobzol:
Cool, I already created something like this in https://github.com/intellij-rust/intellij-rust/pull/6801 (https://github.com/intellij-rust/intellij-rust/pull/6801/commits/1ec832d9b8792395c160907de2f45cf1a81b7523). If this is urgent, let's merge it and I'll rebase on top of it.

Kobzol (approved):

Undin:
> Cool, I already created something like this in #6801 (1ec832d). If this is urgent, let's merge it and I'll rebase on top of it.

I want to merge it as is now because:
- it should be included in 141 release to prevent unexpected exceptions with old TOML plugin. #6801 may not be merged before release branch creation (next Tuesday)
- `CargoTomlInspection` introduced in #6801 is about inspections in `Cargo.toml` that want to use crate index. But we shouldn't touch TOML psi in any inspection in `toml` module regardless of concrete file or other conditions. For example, potential inspections for `rustfmt.toml` also should do nothing if `tomlPluginIsAbiCompatible` is false

Undin:
bors retry

Issue #6746, opened 2021-01-27
Enums are not properly rendered in CLion 2021.1 with gdb


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.SNAPSHOT-211-dev
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2021.1 EAP (CL-211.4961.21)
* **Operating system:** Linux 4.15.0-132-generic
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description
Expected:
![Screenshot from 2021-01-28 01-37-18](https://user-images.githubusercontent.com/2539310/106064074-4b8c8680-610a-11eb-8d3c-295d1660259e.png)

Actual:
![Screenshot from 2021-01-28 01-42-16](https://user-images.githubusercontent.com/2539310/106064095-547d5800-610a-11eb-8aea-cdeb16e9e218.png)

## Steps to reproduce
```rust
enum EnumA {
    Var1 { a: u32 },
    Var2(u64),
    Var3 { a: u32, b: TestEnumB },
}

enum TestEnumB {
    Var1(u64),
    Var2 { a: u32, b: String, c: EnumC },
}

enum EnumC {
    Var1(u64),
}

enum EnumD {}

fn main() {
    let a = EnumA::Var3 {
        a: 5,
        b: TestEnumB::Var2 {
            a: 5,
            b: "hello".to_owned(),
            c: EnumC::Var1(8),
        },
    };
    let d: EnumD;
    print!(""); // #break
}
```

memoryallocator:
I faced the same issue. The simplier way to reproduce:

```
#[derive(Eq, PartialEq, Debug)]
    enum NestedEnum {
        NestedFirst,
        NestedSecond,
        NestedLast,
    }

    #[derive(Eq, PartialEq, Debug)]
    enum Enum {
        Variant(NestedEnum),
        A,
        B,
        C,
    }

    let q = Enum::A;

    assert_ne!(Enum::A, Enum::C);
    assert_eq!(q, Enum::A);
    assert_eq!(q, Enum::C);
```

![image](https://user-images.githubusercontent.com/40700358/107142312-16b8d480-693f-11eb-88e7-364e721446b3.png)

More than that, on commenting out Enum::B variant, I get
[...]

PR #6804, opened 2021-02-05
Fix enum rendering while debugging with GDB in CLion 2021.1

ortem:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6746

The new enum representation was introduced in https://github.com/bminor/binutils-gdb/commit/9c6a1327ad9a92b8584f0501dd25bf8ba9e84ac6

To test:
- [x] CLion 2020.3 on Linux
- [x] CLion 2021.1 on Linux
- [ ]  CLion 2020.3 with MinGW on Windows
- [ ]  CLion 2021.1 with MinGW on Windows

changelog: Fix enum rendering while debugging with GDB in CLion 2021.1

### DIFF
--- a/prettyPrinters/gdb_lookup.py
+++ b/prettyPrinters/gdb_lookup.py
@@ -12,18 +12,11 @@ def is_hashbrown_hashmap(hash_map):
     return len(hash_map.type.fields()) == 1
 
 
-def classify_rust_type(type):
-    type_class = type.code
-    if type_class == gdb.TYPE_CODE_STRUCT:
-        return classify_struct(type.tag, type.fields())
-    if type_class == gdb.TYPE_CODE_UNION:
-        return classify_union(type.fields())
-
-    return RustType.OTHER
-
-
-def check_enum_discriminant(valobj):
+# Enum representation in gdb <= 9.1
+def is_old_enum(valobj):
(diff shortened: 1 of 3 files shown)


ortem:
I haven't managed to test this on Windows with MinGW-W64 yet. Seems like CLion (even 2020.3) ignores pretty-printed values and renders them using the raw representation, although GDB console shows them fine:
Screenshot 2021-02-05 at 13 34 22

I'll investigate this problem a bit later.

Undin (commented):
Works for me with CLion 2021.1 and CLion 2020.3 with bundled GDB
But pretty-printer tests fail

ortem:
@Undin I've changed the test. It failed because `c = enum::EnumC::Var1 = { ... }` was expected, but now such single-variant enums are treated as regular structs (since they don't contain artificial `<<variant>>` field now and AFAIU it's impossible to distinguish them from regular structs) and therefore the actual string representation is `c = enum::EnumC = { ... }`.

However, this particular thing does not affect enum rendering in IDE anyhow because CLion [ignores](https://youtrack.jetbrains.com/issue/CPP-15217) `to_string` if there are synthetic children (enum variants in that case)

Undin (approved):
👍 
bors r+

PR #6807, opened 2021-02-05
REF: allow choosing default values for new parameters in ChangeSignature

Kobzol:
I started working on suggested refactoring for functions using the new change signature refactoring, but I realized that it's not that useful without default value parameters, so first I'll begin with this PR.

Unresolved questions:
1) What about error checking? Should we somehow check that the default value expression's type is compatible with the parameter type? Or just do a basic sanity check that the expression should be parsed?
2) What to do about importing types from the default value expression? The trick with storing the type reference from the parameter type code fragment no longer works here, because the type reference created in the dialog has its context set to the refactored function. We need a type reference with a context set to each specific function call for it to work, and that is not available during refactoring. Maybe we can just automatically invoke the import quick fix on the inserted argument expression?

Btw @dima74 your refactored function for fixing parameters/arguments was really pleasant to use here, I just had to modify a single line of code to get default values working :+1:.

Related issue: https://github.com/intellij-rust/intellij-rust/issues/6744

changelog: The Change Signature refactoring now allows you to choose a default value for new parameters.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
@@ -28,7 +28,7 @@ import org.rust.ide.refactoring.isValidRustVariableIdentifier
 import org.rust.lang.RsFileType
 import org.rust.lang.core.macros.setContext
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.*
 import org.rust.openapiext.document
 import java.awt.BorderLayout
 import java.awt.GridBagConstraints
@@ -68,7 +68,7 @@ fun withMockChangeFunctionSignature(mock: ChangeFunctionSignatureMock, action: (
 private class SignatureParameter(val factory: RsPsiFactory, val parameter: Parameter) : ParameterInfo {
     override fun getName(): String = parameter.patText
     override fun getOldIndex(): Int = parameter.index
-    override fun getDefaultValue(): String? = null
+    override fun getDefaultValue(): String = parameter.defaultValue.text
     override fun setName(name: String?) {
         if (name != null) {
             parameter.patText = name
@@ -81,8 +81,9 @@ private class SignatureParameter(val factory: RsPsiFactory, val parameter: Param
     override fun setUseAnySingleVariable(b: Boolean) {}
 }
 
-private class SignatureDescriptor(val config: RsChangeFunctionSignatureConfig)
-    : MethodDescriptor<SignatureParameter, String> {
+private class SignatureDescriptor(
+    val config: RsChangeFunctionSignatureConfig
+) : MethodDescriptor<SignatureParameter, String> {
     val function: RsFunction = config.function
 
     override fun getName(): String = config.name
@@ -108,23 +109,29 @@ private class SignatureDescriptor(val config: RsChangeFunctionSignatureConfig)
     override fun canChangeReturnType(): MethodDescriptor.ReadWriteOption = MethodDescriptor.ReadWriteOption.ReadWrite
 }
 
-private class ModelItem(val function: RsFunction, parameter: SignatureParameter)
-    : ParameterTableModelItemBase<SignatureParameter>(
+private class ModelItem(
+    importContext: RsMod,
+    parameter: SignatureParameter
+) : ParameterTableModelItemBase<SignatureParameter>(
     parameter,
-    createTypeCodeFragment(function, parameter.parameter.parseTypeReference()),
-    createTypeCodeFragment(function, parameter.parameter.parseTypeReference()),
+    createTypeCodeFragment(importContext, parameter.parameter.parseTypeReference()),
+    createExprCodeFragment(importContext),
 ) {
     override fun isEllipsisType(): Boolean = false
 }
 
-private class TableModel(val descriptor: SignatureDescriptor, val onUpdate: () -> Unit)
-    : ParameterTableModelBase<SignatureParameter, ModelItem>(
+private class TableModel(
+    val descriptor: SignatureDescriptor,
+    val onUpdate: () -> Unit
+) : ParameterTableModelBase<SignatureParameter, ModelItem>(
     descriptor.function,
     descriptor.function,
     NameColumn<SignatureParameter, ModelItem>(descriptor.function.project, "Pattern:"),
-    SignatureTypeColumn(descriptor)
+    SignatureTypeColumn(descriptor),
+    SignatureDefaultValueColumn(descriptor)
 ) {
     private val factory: RsPsiFactory = RsPsiFactory(descriptor.function.project)
+    private val importContext: RsMod = descriptor.function.createImportContext()
 
     init {
         addTableModelListener {
@@ -139,7 +146,7 @@ private class TableModel(val descriptor: SignatureDescriptor, val onUpdate: () -
             SignatureParameter(factory, newParameter)
         } else parameterInfo
 
-        return ModelItem(descriptor.function, parameter)
+        return ModelItem(importContext, parameter)
     }
 
     override fun removeRow(index: Int) {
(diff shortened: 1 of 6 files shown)


dima74:
> 1. What about error checking? Should we somehow check that the default value expression's type is compatible with the parameter type? Or just do a basic sanity check that the expression should be parsed?

I think basic sanity check is enough, looks like in some other languages there is no check at all

> 2. What to do about importing types from the default value expression? The trick with storing the type reference from the parameter type code fragment no longer works here, because the type reference created in the dialog has its context set to the refactored function. We need a type reference with a context set to each specific function call for it to work, and that is not available during refactoring. Maybe we can just automatically invoke the import quick fix on the inserted argument expression?

We have two problems:
* Actual type is lost during `RsExpressionCodeFragment` -> `defaultValueText` (`String`) → `defaultValue` (`RsExpr`) conversion. Probably we should find a way to directly store `RsExpressionCodeFragment::expr` in `Parameter::defaultValue`
* `RsImportHelper.importTypeReferencesFromElements` imports only types, not references from `RsPath`. Probably we should create `RsImportHelper.importPathReferencesFromElements`

Here is proof of concept with import working: https://github.com/intellij-rust/intellij-rust/commit/781497fd0ac51940611f8dc293c180640c8c1cb2. There both problems were solved in a very hacky way (so probably it should be refactored somehow)

Kobzol:
Ooh, I think that this: https://github.com/intellij-rust/intellij-rust/commit/781497fd0ac51940611f8dc293c180640c8c1cb2#diff-f90f288e15fa514055574bbec4bae978c9cbfcf80aa0366053250f07a964075aR103 is what has been missing. Thanks, I'll refactor it and push a new version.

src/main/kotlin/org/rust/ide/utils/import/RsImportHelper.kt
```diff
@@ -100,6 +98,20 @@ object RsImportHelper {
...
+                if (qualifier == null) {
+                    val item = path.reference?.resolve() as? RsQualifiedNamedElement
+                    if (item != null) {
+                        result += item
+                    }
+                } else {
+                    visitPath(qualifier)
+                }
+                path.typeArgumentList?.acceptChildren(this)
+                path.typeQual?.acceptChildren(this)
```
dima74, line 112:
```suggestion
                }
                super.visitPath(path)
```

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
```diff
@@ -356,16 +371,30 @@ private class ChangeSignatureDialog(project: Project, descriptor: SignatureDescr
...
+        "",
+        context = importTarget,
+        importTarget = importTarget
+    )
+}
+private fun createCodeFragment(
+    context: RsElement,
+    factory: (importTarget: RsItemsOwner) -> RsCodeFragment
 ): PsiCodeFragment {
     val freshMod = RsPsiFactory(context.project).createModItem(TMP_MOD_NAME, "")
```
dima74, line 394: Can we use single `freshMod` for all code fragments (both `type` and `default value`) for all parameters? So that if some type was imported in `type` code fragment, user will not need to import it again in `default value`. E.g. consider some enum `Foo` with enum variant `Foo::Bar`, user enters `Foo` as new parameter type and `Foo::Bar` as this parameter default value

Kobzol: That's a very good point! I'll change it. Also, is there a way for the temporary context module to "see" the same things as the refactored function? Importing works fine, but if I want to use some item that is in the same scope as the function, I currently still have to import it in the type/default value fragment.

dima74: > Also, is there a way for the temporary context module to "see" the same things as the refactored function?

Maybe set `use path::to::function::containing::mod;` as fresh module text

Kobzol: Hmm, it won't help with relative paths (like `super`, but it's not a big deal). I hoped that there would be some better way with setting the context of the module, but if not, this is a good enough fix :) Thanks.

Kobzol:
I used two different contexts (one for all parameters and another one for the return type). Mainly because it was pretty annoying to create a single shared context inside `ChangeSignatureDialog`, because its base class constructor calls an overridden method that needs the import context (`createParametersInfoModel`).

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
```diff
@@ -160,14 +162,22 @@ private class TableModel(val descriptor: SignatureDescriptor, val onUpdate: () -
...
         override fun setValue(item: ModelItem?, value: PsiCodeFragment?) {
             val fragment = value as? RsTypeReferenceCodeFragment ?: return
             if (item != null) {
-                item.parameter.parameter.type = ParameterType.fromText(fragment.typeReference, fragment.text)
+                item.parameter.parameter.type = ParameterProperty.fromText(fragment.typeReference, fragment.text)
+            }
+        }
+    }
+
+    private class SignatureDefaultValueColumn(descriptor: SignatureDescriptor) : DefaultValueColumn<SignatureParameter, ModelItem>(descriptor.function.project, RsFileType) {
```
dima74, line 176: Please split this and other long lines into multiple lines

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
```diff
@@ -399,3 +429,17 @@ private class VisibilityComboBox(project: Project, initialVis: RsVis?, onChange:
 
 private fun createVisibilityHints(initialVis: RsVis?): Array<String> =
     setOf(initialVis?.text.orEmpty(), "", "pub", "pub(crate)", "pub(super)").toTypedArray()
+
+private fun RsElement.createImportContext(): RsMod {
+    val factory = RsPsiFactory(project)
+    val module = factory.createModItem(TMP_MOD_NAME, "")
+    val sourceContext = containingFile as RsFile
+
+    val qualifiedPath = containingMod.qualifiedNameRelativeTo(sourceContext)
```
dima74, line 438: Shouldn't we use absolute path here (`qualifiedNameInCrate`)?

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
```diff
@@ -399,3 +429,17 @@ private class VisibilityComboBox(project: Project, initialVis: RsVis?, onChange:
 
 private fun createVisibilityHints(initialVis: RsVis?): Array<String> =
     setOf(initialVis?.text.orEmpty(), "", "pub", "pub(crate)", "pub(super)").toTypedArray()
+
+private fun RsElement.createImportContext(): RsMod {
+    val factory = RsPsiFactory(project)
+    val module = factory.createModItem(TMP_MOD_NAME, "")
```
dima74, line 435: Maybe pass use item as `modText` here

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
Kobzol: I checked the (auto)imported imports and they are relative to the containing file (probably because of `setContext`). So that's why I used the same import style (I think that absolute imports didn't work). However, now that I think of it, it might have been caused by the usage of edition 2015. I'll test it in both editions to find a solution that works for both.

Kobzol: Thank you, `qualifiedNameInCrate` works in both editions.

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureDialog.kt
```diff
@@ -399,3 +437,17 @@ private class VisibilityComboBox(project: Project, initialVis: RsVis?, onChange:
...
     setOf(initialVis?.text.orEmpty(), "", "pub", "pub(crate)", "pub(super)").toTypedArray()
+
+private fun RsElement.createImportContext(): RsMod {
+    val factory = RsPsiFactory(project)
+    val sourceContext = containingFile as RsFile
+
+    val qualifiedPath = containingMod.qualifiedNameInCrate(this)
+    val defaultUseItem = if (qualifiedPath != null) {
+        "use $qualifiedPath::*;"
+    } else ""
```
dima74, line 448:
```suggestion
    } else {
        ""
    }
```

dima74:
Thanks!
bors r+

PR #6813, opened 2021-02-07
INT: Convert from lambdas to local named functions

### LINKED ISSUE #6711: Convert function to closure (and vice versa) intentions
## Environment

* **IntelliJ Rust plugin version:** 0.3.140.3613-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **Operating system:** macOS 10.16

## Problem description
It would be great to have intentions to convert a local function into a closure and vice versa. For example, to convert
```Rust
fn main() {
    fn foo(x: i32) -> i32 { x + 1 }
}
```
into
```Rust
fn main() {
    let foo = |x: i32| -> i32 { x + 1 };
}
```
and vice versa.

This intention might also be a quick fix for [`E0434`](https://doc.rust-lang.org/error-index.html#E0434) compiler error (which is not implemented yet and tracked [here](https://github.com/intellij-rust/intellij-rust/issues/886)). For example,
```Rust
fn main() {
    let env = 0;
    fn foo_func(x: i32) -> i32 { x + env + 1 } // error[E0434]: can't capture dynamic environment in a fn item; help: use the `|| { ... }` closure form instead
}
```

nschoellhorn:
Since this is my first PR on this project, I'd be very grateful for detailed feedback on how I approached this issue.
One thing I'm still not sure about is the way I'm generating the context in `findApplicableContext()`. The way it currently checks whether the intention is applicable, is pretty broad. The caret could be somewhere in the lambda/function body as well and the intention would still be available. I didn't find any general guidelines on how intention availability is defined, would appreciate directions here. 

Also, regarding tests: I am not sure what I can test here exactly. I added tests for the basic "positive" cases and also a test for the edge case that the lambda is bound to `let _`, which would result in an invalid function name. Any other good test case hints would be very appreciated as well! :)

Closes #6711 

changelog: Added intentions to convert from lambdas to named local functions and vice versa

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
@@ -7,13 +7,17 @@ package org.rust.ide.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
 import com.intellij.psi.SmartPsiElementPointer
-import org.rust.ide.presentation.render
-import org.rust.ide.refactoring.isValidRustVariableIdentifier
+import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.utils.template.buildAndRunTemplate
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.ext.ancestorStrict
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.lang.core.psi.ext.startOffset
+import org.rust.lang.core.psi.ext.valueParameters
+import org.rust.lang.core.types.ty.TyFunction
 import org.rust.lang.core.types.ty.TyUnit
 import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
```diff
@@ -0,0 +1,59 @@
...
+    data class Context(
+        val assignment: RsLetDecl,
+        val lambda: RsLambdaExpr,
+    )
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        // We try to find a let declaration
+        val possibleTarget = element.ancestorStrict<RsLetDecl>() ?: return null
+
+        // we cannot convert lambdas bound to "_"
```
Kobzol, line 30: Maybe a better approach would be to create the function always (even if there is an invalid binding or no binding at all), and let the user change it's name afterwards. If the binding was invalid, the default name of the function could be something autogenerated (like `fn0`).

`RsInPlaceVariableIntroducer` or `buildAndRunTemplate` can be used for the renaming.

src/main/kotlin/org/rust/ide/intentions/ConvertFunctionToClosureIntention.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    override fun getText(): String = "Convert function to closure"
+
+    data class Context(
+        val targetFunction: RsFunction,
+    )
+
+    override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
+        // We try to find a function declaration
+        val possibleTarget = element.ancestorStrict<RsFunction>() ?: return null
```
Kobzol, line 26: I suggest to only make the intention available in the function signature (e.g. its identifier and parameter list). See https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureHandler.kt#L28 for inspiration (basically the same code could work, it you remove the cases for `RsCallExpr` and `RsMethodCall`.

src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
nschoellhorn: I like that idea. I changed it so it builds the template in case an invalid function name would be generated. Please tell me if there's a better way for that than `isValidVariableIdentifier()`. Also, should we change the availability of this intention to only the arguments and maybe the variable name as well? Currently, this intention would also be available anywhere in the closure body.

Kobzol: About the availability: yes, I suggest to only offer it in the signature of functions and parameter list + variable name for closures (https://github.com/intellij-rust/intellij-rust/pull/6813#discussion_r571641678). I think that `isValidVariableIdentifier` is ok.

src/main/kotlin/org/rust/ide/intentions/ConvertFunctionToClosureIntention.kt
nschoellhorn: I've changed this for both intentions as well - thank you for the hints on how to test that! For the closure -> function intention, I had to hack a little bit to make it feel like I thought it should. Please let me know if there's a better way of excluding the part after `}`.

Stzx:
Regarding the generation of capture mode keywords for asynchronous or variable, will it be implemented if possible?

nschoellhorn:
@Stzx I'm not sure what you mean, could you give me an example?

If you're talking about async functions, I'm not sure on how to convert them to closures. As of now, there aren't async closures, just async blocks. Async closures are tracked here: https://github.com/rust-lang/rust/issues/62290 - it'll take some time until this becomes a stable feature. I think implementing this conversion should be deferred until the suggested feature is stable.

Kobzol (changes requested):
I feel like this intention will be ultimately a bit more complicated than it seems :sweat_smile: Would you mind splitting this PR into two? (one for function -> closure, another one for closure -> function).

Also consider the following cases:
- What happens here? There is no type annotation, so a function cannot be simply created out of the value parameters. We can either:
  - Disable the intention on such closure
  - Try to use inferred parameter types (I'm not sure if the plugin can infer them)
  - Create a dummy type (like `()`) and then let the user interactively change all of the missing parameter types using e.g. template builder.
```rust
let x = |a| a + 1;
x(5);
```
- If the closure's return type is not specified, it should be inferred from the lambda expression (`expr.type`) and put inside the new function:
```rust
let x = |a: u32| a + 1;
// should create fn x(a: u32) -> u32 { a + 1 };
```
Also do not forget to add the block `{ ... }` to the function body if the lambda has no block. You can check if the lambda's expr is `RsBlockExpr`.

- What if the function had some generic parameters?
```rust
fn foo<T>(a: T) {}
```
In this case parameter type `T` in the closure will have an invalid type, and Rust doesn't support generic closures. I suggest to remove the type reference from the parameter in such case. I think that the easiest way how to do that would be to generate the closure, then go through the type references of its parameters and if they are invalid (e.g. they contain `TyUnknown`), remove them and hope that Rust's inference will take care of the rest.

Stzx:
@nschoellhorn  Sorry for not expressing my point clearly.

> @Stzx I'm not sure what you mean, could you give me an example?

My example may be a bit extreme, we don't normally write code like this.

For code that needs to use `move` explicitly:
```rust
fn foo(a: f64) -> impl Fn(f64) -> f64 {
    fn to(x: f64) -> f64 {
        a + x
    }
}
```

```rust
fn foo(a: f64) -> impl Fn(f64) -> f64 {
    let to = |x: f64| -> f64 { a + x };
    //      ^ The move keyword should be used explicitly here
    to
}
```

But for function using this conversion, the implicit move can already handle most of the problems.

About asynchrony. I guess if the maintainer agrees, I have no problem.

nschoellhorn:
@Kobzol 
> I feel like this intention will be ultimately a bit more complicated than it seems 😅 Would you mind splitting this PR into two? (one for function -> closure, another one for closure -> function).

Sure, will do! Since there are already some comments about both parts here, how would you like to see the split? Should I just remove the commits from this branch and then force push, and then add them to a different branch? The comments would get lost then, right? Is there a better way? Never did that before :/

Kobzol:
A few comments getting lost isn't a big deal (as long as you remember them :D ). I suggest to create a new PR with one of the intentions, and then remove that one from this PR and force-push. It's OK to force-push as long as the PR is reasonably small.

nschoellhorn:
I've rebased and adressed the issues you mentioned, @Kobzol  
I also added a few more test cases for the changes. Sadly, since switching to my new M1 MacBook, I can't run tests locally anymore due to some incompatible native dependencies (JNA I think?), so let's see what CI says :)

nschoellhorn:
Not sure why the test is failing. Obviously `type` contains `TyUnknown` here and thus the placeholder unit type is added, even though the engine should be able to infer the type. I am pretty sure that this worked in my local tests previously, but now it doesn't work anymore. What's the best play here? Change the test to make it pass or just remove it entirely? I think "fixing" the engine to correctly infer the type here is a bit over the top for this PR, right?

Kobzol:
Sorry for the long response. The problem is that the plugin can't infer binary operations on integers without the stdlib (it contains macros that define the impls for such operations on primitive types). Add `@ProjectDescriptor(WithStdlibRustProjectDescriptor::class)` above the failing test.

Also when the return type is `()`, probably it's better to keep it empty, rather than to insert `-> ()`.

nschoellhorn:
No problemo, we all do this in our spare time, right?! So no worries! Thank you for the hints, I added a test for your case of Unit return type and fixed the code correspondingly. Let's see what CI says.

src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
```diff
@@ -0,0 +1,135 @@
...
+        var isTemplate = if (targetName == null || !isValidRustVariableIdentifier(targetName)) {
+            targetName = "func"
+
+            true
+        } else {
+            false
+        }
+
+        val valueParameters = ctx.lambda.valueParameters
+        val (parametersText, parameterPlaceholders) = this.createParameterList(factory, valueParameters)
```
Kobzol, line 77: Redundant `this`.

src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
```diff
@@ -0,0 +1,135 @@
...
+        if (parameterPlaceholders.isNotEmpty()) {
+            isTemplate = true
+        }
+        val lambdaExpr = ctx.lambda.expr
+        val returnText = if (ctx.lambda.retType != null) {
+            ctx.lambda.retType!!.text
```
Kobzol, line 83:
```suggestion
        val returnType = ctx.lambda.retType
        val returnText = if (returnType != null) {
            returnType.text
```
Just to avoid unnecessary `!!`.

src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
```diff
@@ -0,0 +1,135 @@
...
+                it.type == TyUnit
+            }.map { it.createSmartPointer() })
+                editor.buildAndRunTemplate(replaced, placeholderElements)
+
+        } else {
+            editor.caretModel.moveToOffset(replaced.endOffset)
+        }
+    }
+
+    private fun createParameterList(factory: RsPsiFactory, valueParameters: List<RsValueParameter>): Pair<String, List<RsTypeReference>> {
```
Kobzol, line 117:
```suggestion
private fun createParameterList(
        factory: RsPsiFactory,
        valueParameters: List<RsValueParameter>
    ): Pair<String, List<RsTypeReference>> {
```

src/main/kotlin/org/rust/ide/intentions/ConvertClosureToFunctionIntention.kt
```diff
@@ -0,0 +1,135 @@
...
+
+        val identifier = replaced.identifier
+
+        // in case we auto-generated a function name, we encourage the user to change it by running a template on the replacement
+        if (isTemplate) {
+            val placeholderElements = mutableListOf<SmartPsiElementPointer<PsiElement>>()
+            if (targetName == "func") {
+                placeholderElements.add(identifier.createSmartPointer())
+            }
+            placeholderElements.addAll(replaced.descendantsOfType<RsTypeReference>().filter {
```
Kobzol, line 107: This would also use the template i.e. for unit type reference inside the generated function body.
What about returning a list of parameter indices that need to be replaced from `createParameterList` and then using this index list to select the correct parameters to put into the template builder?

Kobzol (approved):
Looks good to me, please just take a look at the comments and also reformat the `ConvertClosureToFunctionIntention` file using the IntelliJ auto-formatter.

@ortem can you take a look? (Since you created the original issue :D)

vlad20012:
@nschoellhorn thanks!
I've pushed several fixes here.

bors r+

Issue #6812, opened 2021-02-07
"Crate not found" false positive in Cargo.toml after enabling "Local Crates Index"


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.3655-203-nightly
* **Rust toolchain version:** 1.49.0 (e1884a8e3 2020-12-29) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3.2 (CL-203.7148.70)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

While testing #6439 I enabled `org.rust.crates.local.index` experimental feature. It triggers "Crate not found" false positive (the crate is installed and compilation / running works OK). Cache invalidation doesn't help to resolve the problem

<img width="400" alt="not_found" src="https://user-images.githubusercontent.com/2500320/107145609-dd3e9400-6953-11eb-9977-ed2fbd8cb913.png">

## Steps to reproduce

```toml
[dependencies]
xml-rs = "*"
```

My log: [idea.log](https://github.com/intellij-rust/intellij-rust/files/5938924/idea.log).

Kobzol:
@avrong This may be the same situation that I have mentioned in https://github.com/intellij-rust/intellij-rust/pull/6591#issuecomment-752784229, it seems that sometimes (at least some crates) don't end up in the index.

PR #6818, opened 2021-02-08
Fix crate version json parsing

avrong:
Bug introduced in https://github.com/intellij-rust/intellij-rust/pull/6591

As it turns out, both `Gson` and `Jackson` can't work with inside-function classes.

Citation from https://github.com/FasterXML/jackson-module-kotlin/issues/135:
> It sounds like inner class you use would be non-static (as per exception). Such classes can not be deserialized since there is no access to implicit this reference that would be needed, at least on regular Java. I don't know if Kotlin allows some way make inner class static from JVM perspective; if so, that would make things work fine.

`Gson` is returning `null` in this case without any hint of what could go wrong.

This PR fixes the parsing logic by moving `ParsedVersion` out from the function and uses Jackson as it gives more understandable messages and has better Kotlin support. We might also try to use `kotlinx.serialization` instead, but it may need to have some modifications in Gradle plugins.

In my environment, it now works with no one warning about failed-to-parse versions.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
@@ -5,7 +5,10 @@
 
 package org.rust.toml.crates.local
 
-import com.google.gson.Gson
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.json.JsonMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.PersistentStateComponent
@@ -314,15 +317,18 @@ private object CrateExternalizer : DataExternalizer<CargoRegistryCrate> {
     }
 }
 
-private fun crateFromJson(json: String): CargoRegistryCrateVersion {
-    data class ParsedVersion(
-        val name: String,
-        val vers: String,
-        val yanked: Boolean,
-        val features: HashMap<String, List<String>>
-    )
+data class ParsedVersion(
+    val name: String,
+    val vers: String,
+    val yanked: Boolean,
+    val features: HashMap<String, List<String>>
+)
 
-    val parsedVersion = Gson().fromJson(json, ParsedVersion::class.java)
+private fun crateFromJson(json: String): CargoRegistryCrateVersion {
+    val parsedVersion = JsonMapper()
+        .registerKotlinModule()
+        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+        .readValue<ParsedVersion>(json)
 
     return CargoRegistryCrateVersion(
         parsedVersion.vers,


Kobzol:
Heh, it seemed like an innocent change :) Good catch, thanks.

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -314,15 +316,17 @@ private object CrateExternalizer : DataExternalizer<CargoRegistryCrate> {
...
+data class ParsedVersion(
+    @JsonProperty("name") val name: String,
+    @JsonProperty("vers") val vers: String,
+    @JsonProperty("yanked") val yanked: Boolean,
+    @JsonProperty("features") val features: HashMap<String, List<String>>
+)
 
-    val parsedVersion = Gson().fromJson(json, ParsedVersion::class.java)
+private fun crateFromJson(json: String): CargoRegistryCrateVersion {
+    val parsedVersion = ObjectMapper()
```
ortem, line 327: Consider using Kotlin module this way:
```Kotlin
    val parsedVersion = JsonMapper()
        .registerKotlinModule()
        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
        .readValue<ParsedVersion>(json)
```
See `org.rustPerformanceTests.CustomRealProjectAnalysisTest.Companion#data` for example.

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -314,15 +316,17 @@ private object CrateExternalizer : DataExternalizer<CargoRegistryCrate> {
...
 
-private fun crateFromJson(json: String): CargoRegistryCrateVersion {
-    data class ParsedVersion(
-        val name: String,
-        val vers: String,
-        val yanked: Boolean,
-        val features: HashMap<String, List<String>>
-    )
+data class ParsedVersion(
+    @JsonProperty("name") val name: String,
```
ortem, line 320: Looks like these annotations are redundant if the Kotlin module is used (see the comment below)

ortem (changes requested):

ortem:
BTW I haven't managed to reproduce the bug, but I faced a different problem:
bug

@avrong Is this related to the original issue or JSON parsing?

ortem:
> Heh, it seemed like an innocent change :)

@Kobzol Yep, and it's a funny thing that I suggested this breaking change in the review actually, sorry for that 😄

avrong:
@ortem That's interesting. I fixed the problem that the log warnings contained, but now see it is different from the one explained in the issue. Even with the version info parsing problems, there should be no problems with crate names by themselves.

The original problem might be related to @Kobzol case, as he mentioned. But I haven’t managed to reproduce it too.

In your case, version parsing seems to be the problem. You can invalidate caches to force index load from scratch

ortem (approved):
LGTM. Just to sum up: this **does not** fix https://github.com/intellij-rust/intellij-rust/issues/6812, but anyway, this change is necessary to parse JSON properly. As for https://github.com/intellij-rust/intellij-rust/issues/6812, it should be investigated and fixed in a separate PR.

bors r+

avrong:
bors r=ortem

Issue #6630, opened 2021-01-07
`.match` postfix template should generate all reasonable arms


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3572-203
* **Rust toolchain version:** 1.50.0-nightly (a0d664bae 2020-11-22) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3 (CL-203.5981.166)
* **Operating system:** macOS 10.16
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
Currently, `.match` postfix template generates `match` expression with only one wildcard arm. I think it should generate all the arms that `Add remaining patterns` quick-fix suggests instead.
![match](https://user-images.githubusercontent.com/4854600/103884862-3b2b5200-50f0-11eb-8138-a3f1bd41ad6e.gif)

Related to https://github.com/intellij-rust/intellij-rust/issues/3579

## Steps to reproduce
```Rust
fn foo(x: Option<bool>) {
    x.match/*caret*/
}
```

Issue #4727, opened 2019-12-05
`MatchPostfixTemplate` doesn't fill match arms


ortem:
Currently `MatchPostfixTemplate` only adds a wildcard pattern to the created match expression:
```Rust
enum E { A, B }

fn foo(e: E) {
    match e {
        _ => {}
    }
}
```

It'd be improved to fill the match arms instead:
```Rust
enum E { A, B }

fn foo(e: E) {
    match e {
        E::A => {},
        E::B => {},
    }
}
```

It can be done by invoking or reusing `FillMatchArmsIntention`.

PR #6825, opened 2021-02-10
COMP: add all remaining arms match postfix template

Kobzol:
This PR changes `match` postfix template to generate all match arms instead of just a single wildcard.

Currently, no template is used, the caret is just moved after the left brace of the first arm. Should there be some template to rename things from the generated arms? If yes, what should it rename?

Fixes https://github.com/intellij-rust/intellij-rust/issues/6630
Fixes https://github.com/intellij-rust/intellij-rust/issues/4727

changelog: `match` postfix template now generates all missing match arms by default.

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
@@ -1,0 +1,52 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.template.postfix
+
+import com.intellij.codeInsight.template.postfix.templates.PostfixTemplateWithExpressionSelector
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiElement
+import org.rust.ide.inspections.fixes.AddRemainingArmsFix
+import org.rust.ide.inspections.fixes.AddWildcardArmFix
+import org.rust.ide.utils.checkMatch.checkExhaustive
+import org.rust.lang.core.psi.*
+
+class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
+    PostfixTemplateWithExpressionSelector(
+        null,
+        "match",
+        "match expr {...}",
+        RsExprParentsSelector(),
+        provider
+    ) {
+
+    override fun expandForChooseExpression(expression: PsiElement, editor: Editor) {
+        if (expression !is RsExpr) return
+
+        val project = expression.project
+        val factory = RsPsiFactory(project)
+
+        val exprText = if (expression is RsStructLiteral) "(${expression.text})" else expression.text
+        val emptyMatch = factory.createExpression("match $exprText {}")
+        val matchExpr = expression.replace(emptyMatch) as RsMatchExpr
+
+        PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)
+
+        val patterns = matchExpr.checkExhaustive().orEmpty()
+        val fix = if (patterns.isEmpty()) {
+            AddWildcardArmFix(matchExpr)
+        } else {
+            AddRemainingArmsFix(matchExpr, patterns)
+        }
+        fix.invoke(project, matchExpr.containingFile, matchExpr, matchExpr)
+
+        PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)
+
+        val arm = matchExpr.matchBody?.matchArmList?.firstOrNull() ?: return
+        val blockExpr = arm.expr as? RsBlockExpr ?: return
+        editor.caretModel.moveToOffset(blockExpr.block.lbrace.textOffset + 1)
+    }
+}
--- a/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
@@ -76,20 +76,6 @@ class DerefPostfixTemplate(provider: RsPostfixTemplateProvider) :
         }
     )
 
-class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
-    StringBasedPostfixTemplate("match", "match expr {...}", RsExprParentsSelector(), provider) {
-
-    override fun getTemplateString(element: PsiElement): String = "match ${element.text} {\n\$PAT\$ => {\$END\$}\n}"
-
-    override fun getElementToRemove(expr: PsiElement): PsiElement = expr
-
-    override fun setVariables(template: Template, element: PsiElement) {
-        with(template) {
-            addVariable("PAT", TextExpression("_"), true)
-        }
-    }
-}
-
 class IterPostfixTemplate(name: String, provider: RsPostfixTemplateProvider) :
     StringBasedPostfixTemplate(
         name,
--- a/src/main/resources/postfixTemplates/MatchPostfixTemplate/after.rs.template
+++ b/src/main/resources/postfixTemplates/MatchPostfixTemplate/after.rs.template
@@ -4,6 +4,8 @@ enum Foo {
 
 fn foo(x: Foo) {
     match x {
-        <spot>_</spot> => {}
+        Foo::A => {<spot></spot>}
+        Foo::B => {}
+        Foo::C => {}
     }
 }
--- a/src/test/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplateTest.kt
+++ b/src/test/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplateTest.kt
@@ -29,24 +29,43 @@ class MatchPostfixTemplateTest : RsPostfixTemplateTest(MatchPostfixTemplate(RsPo
         fn process_message() {
             let msg = Message::ChangeColor(255, 255, 255);
             match msg {
-                _ => {}
+                Message::Quit => {/*caret*/}
+                Message::ChangeColor(_, _, _) => {}
+                Message::Move { .. } => {}
+                Message::Write(_) => {}
             }
         }
     """)
 
-    fun `test can match constant`() = doTest("""
+    fun `test constant`() = doTest("""
         const THE_ANSWER: i32 = 42;
 
-        fn check(x: i32) -> bool {
+        fn check(x: i32) {
             x.match/*caret*/
         }
     """, """
         const THE_ANSWER: i32 = 42;
 
-        fn check(x: i32) -> bool {
-            match x {
-                _ => {}
-            }
+        fn check(x: i32) {
+            match x { _ => {/*caret*/} }
+        }
+    """)
+
+    fun `test struct literal`() = doTest("""
+        struct S {
+            a: u32
+        }
+
+        fn foo() {
+            S { a: 0 }.match/*caret*/
+        }
+    """, """
+        struct S {
+            a: u32
+        }
+
+        fn foo() {
+            match (S { a: 0 }) { S { .. } => {/*caret*/} }
         }
     """)
 
@@ -59,11 +78,18 @@ class MatchPostfixTemplateTest : RsPostfixTemplateTest(MatchPostfixTemplate(RsPo
     """, """
         fn main() {
             match a {
-                _ => match b {
-                    _ => {/*caret*/}
-                }
+                _ => match b { _ => {/*caret*/} }
             }
         }
     """)
 
+    fun `test subexpression`() = doTest("""
+        fn main() {
+            let x = 1 + 2.match/*caret*/;
+        }
+    """, """
+        fn main() {
+            let x = match 1 + 2 { _ => {/*caret*/} };
+        }
+    """)
 }


src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
```diff
@@ -0,0 +1,48 @@
...
+        RsTopMostInScopeSelector(),
+        provider
+    ) {
+
+    override fun expandForChooseExpression(expression: PsiElement, editor: Editor) {
+        if (expression !is RsExpr) return
+
+        val factory = RsPsiFactory(expression.project)
+
+        val project: Project = expression.getProject()
```
ortem, line 31:
```suggestion
        val project = expression.project
        val factory = RsPsiFactory(project)
```

src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
```diff
@@ -0,0 +1,48 @@
...
+        val newMatchExpr = factory.tryCreateExpression("match ${expression.text} {\n}") as? RsMatchExpr ?: return
```
ortem, line 32: Do you think we should just do nothing if `tryCreateExpression` returns `null`? IMO it might be better to use `createExpression` to throw an exception: this situation should be very rare, with an opportunity to report an exception to us

ortem (changes requested):
> Should there be some template to rename things from the generated arms?

Possible yes. Note, the resulting code should be reformatted after all. The old implementation based on `StringBasedPostfixTemplate` performs reformatting, but the new one does not:
reformat_needed

I see two different ways of performing reformatting here:
1. `PsiDocumentManager#commitDocument` + `CodeStyleManager#reformat/reformatRange/reformatText`
2. Instantiate a `TemplateImpl` and run it using `TemplateManager#startTemplate`

Regarding the contents of such a template, I don't have a strong opinion right now, and even not sure if a template is better than `commitDocument` + `reformat`. You're welcome to experiment with possible ways and share your ideas 😃

src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
Kobzol: Good point, `match <expr>` should work always (unless there is a parser bug or something).

Kobzol:
`TemplateManager#startTemplate` might be problematic, because we don't have the final generated "template" right away, but we get it dynamically via quick fixes. We could probably invoke the quick fix and then get the text back and run the template, but it seems hacky.

Actually now I'm not sure if we want any templates at all. Since we should now always generate proper match arms, in a lot of cases the generated result will be ok on its own. Sometimes there will of course be some `_` bindings, but the problem is that there can be a lot of them. If you run the postfix completion on a complex enum, it can generate e.g. tens of underscores and I'm not sure if it's a good idea to create so many placeholders to be changed?

I'm having some trouble with reformatting too. I changed the match to have proper indentation. When I generate the match by itself, it creates this:
```rust
let x = match 1 + 2 {

};
```
If I invoke the quick fix on this code manually (from the IDE editor), it produces this:
```rust
let x = match 1 + 2 {
   _ => {}
};
```
which is correct.

However, if I invoke the same fix from code, immediately after the match is inserted into the document, it instead produces this:
```rust
let x = match 1 + 2 {

_ => {}};
```
I tried inserting `PsiDocumentManager.commitDocument` and `PsiDocumentManager.doPostponedOperationsAndUnblockDocument` after the match is inserted, but it doesn't help.

If I reformat the whole thing afterward, it produces this:
```rust
    let x = match 1 + 2 {
    _ => {}
};
```

I tried a few options and couldn't find a variant that would produce the correct formatting :D Do you have an idea why the quick fix produces different code when invoked from code vs via the IDE?

ortem:
> Actually now I'm not sure if we want any templates at all. <...> I'm not sure if it's a good idea to create so many placeholders to be changed?

Thank you for analyzing the possible ways of implementation. I agree with you, let's avoid lots of placeholders here.

As for the formatting problems, please check out the corresponding feature in [Scala plugin](https://github.com/JetBrains/intellij-scala) and [its implementation](https://github.com/JetBrains/intellij-scala/blob/idea211.x/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/completion/postfix/templates/ScalaExhaustiveMatchPostfixTemplate.scala). Probably the proper way is to use a template without any placeholders.

Feel free to provide any further details; if the Scala implementation does not help either, I will try to investigate the formatting problem in more detail.

Kobzol:
I tried the same approach as is done by the Scala plugin (copy the text of the expression, delete it and then reinsert it with the template). It seems to work, however it puts match expressions with a single arm on a single line, I'm not sure if it's ideal.

I need to be able to get to the newly inserted/formatted match expression to move the caret. I used a smart pointer for it, it could also be done by storing the start offset and then look up an element on that offset after the template runs.

shocoman:
Hello. I've waited for this feature for quite some time now. Can I somehow 'nudge' it?
With this piece of code in `expandForChooseExpression` everything works fine. Well, at least in my definition of fine
```kotlin
        val project = expr.project
        val factory = RsPsiFactory(project)

        val exprText = if (expr is RsStructLiteral) "(${expr.text})" else expr.text
        val emptyMatch = factory.createExpression("match $exprText {}")
        val match = expr.replace(emptyMatch) as RsMatchExpr
        val patterns = match.checkExhaustive() ?: emptyList()
        PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)
        WriteCommandAction.runWriteCommandAction(project) {
            if (patterns.isNotEmpty())
                AddRemainingArmsFix(match, patterns).invoke(project, match.containingFile, match, match)
            else
                AddWildcardArmFix(match).invoke(project, match.containingFile, match, match)
            match.matchBody?.let {
                it.addAfter(factory.createNewline(), it.lbrace)
                it.addBefore(factory.createNewline(), it.rbrace)
            }

            match.matchBody?.matchArmList?.firstOrNull()?.expr.castSafelyTo<RsBlockExpr>()?.block?.let {
                editor.caretModel.moveToOffset(it.lbrace.textOffset + 1)
            }
        }
```

ezgif-1-06865cb49f1e

Kobzol:
@ortem Can you take another look please? Thanks :)

shocoman:
Duplicate issue #4727 can also be closed if this PR will ever get merged

ortem:
Hi @shocoman, thank you for finding an issue duplicate and providing your implementation. It seems to work fine, and we definitely should handle `RsStructLiteral`. @Kobzol, please add a test with struct literal matching.

However, I'm not sure that additional `WriteCommandAction` is really necessary because `expandForChooseExpression` is called under write action (see `PostfixTemplateWithExpressionSelector#prepareAndExpandForChooseExpression`). Looks like all we have to do is to invoke `doPostponedOperationsAndUnblockDocument` again before moving the caret, as the quick-fix has just modified the document through the PSI. So I suppose the following approach should work as well:
```Kotlin
val exprText = if (expression is RsStructLiteral) "(${expression.text})" else expression.text
val emptyMatch = factory.createExpression("match $exprText {}")
val match = expression.replace(emptyMatch) as RsMatchExpr

PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)

val patterns = match.checkExhaustive() ?: emptyList()
val fix = if (patterns.isNotEmpty()) {
    AddRemainingArmsFix(match, patterns)
} else {
    AddWildcardArmFix(match)
}
fix.invoke(project, match.containingFile, match, match)

PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)

val firstArmBlock = match.matchBody?.matchArmList?.firstOrNull()?.expr as? RsBlockExpr ?: return
editor.caretModel.moveToOffset(firstArmBlock.block.lbrace.textOffset + 1)
```

Kobzol:
I'm still having trouble with the formatting. The case with `1 + 2.match` still has wrong formatting if I add newlines manually (otherwise it would just be on a single line).

When I reformat the whole file after the quick fix, it works, but that's probably not a great solution. I tried to format the nearest containing statement, but that doesn't seem to work. It works if I reformat the parent of the containing expression (which will most likely be the containing function). Is that OK?

```rust
fn foo() {
    let x = 1 + 2.match
}
// v
fn foo() {
    let x = match 1 + 2 {
    _ => {}
};
}
```

shocoman:
With 
```
match.matchBody?.let {
    it.addAfter(factory.createNewline(), it.lbrace)
    it.addBefore(factory.createNewline(), it.rbrace)
}
```
as in my example it works correctly. Though it probably does formatting underneath

Kobzol:
I am actually using your code, but it doesn't work always. Have you tried it on my example with `let x = 1 + 2`? Your code works in the cases that you have posted, and it also works if the match has multiple lines, but for `1 + 2` it generates wrong indentation.

ortem:
@Kobzol

> The case with 1 + 2.match still has wrong formatting if I add newlines manually (otherwise it would just be on a single line)

This behaviour is actually expected. Such formatting is provided by the plugin. So if you run `Reforamat Code` on the following code (without `Use rustfmt instead of built-in formatter` option)
```Rust
fn foobar() {
    let x = 1 + match 2 { _ => {} };
}
```
you will get the same only-line formatting. On the other hand, adding whitespace manually is a [bad practice](https://plugins.jetbrains.com/docs/intellij/modifying-psi.html#whitespaces-and-imports).

ortem:
@Kobzol Yes, the plugin's formatter puts `match 1 + 2 { _ => {} } ` on a single line, as well as `match 2 { _ => {} }`. This behaviour differs from `rust-fmt` indeed. But in order to fix it, the plugin's formatter (located in `org/rust/ide/formatter`) should be changed instead.

Kobzol:
@shocoman Interesting :) Doesn't work for me, maybe it's something with the write action.

Anyway, I agree that we're solving a problem in the wrong place, this should be solved inside the formatter.
I pushed a version without formatting and with added struct literal test (thanks @shocoman).

src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
```diff
@@ -0,0 +1,54 @@
...
+        "match",
+        "match expr {...}",
+        RsExprParentsSelector(),
+        provider
+    ) {
+
+    override fun expandForChooseExpression(expression: PsiElement, editor: Editor) {
+        if (expression !is RsExpr) return
+
+        val project: Project = expression.project
```
ortem, line 30:
```suggestion
        val project = expression.project
```

src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
```diff
@@ -0,0 +1,54 @@
...
+            AddRemainingArmsFix(matchExpr, patterns)
+        }
+        fix.invoke(project, matchExpr.containingFile, matchExpr, matchExpr)
+
+        PsiDocumentManager.getInstance(project).doPostponedOperationsAndUnblockDocument(editor.document)
+
+        val arm = matchExpr.matchBody?.matchArmList?.firstOrNull() ?: return
+        val blockExpr = arm.expr as? RsBlockExpr ?: return
+        val block = blockExpr.block as? RsBlock ?: return
+        editor.caretModel.moveToOffset(block.lbrace.textOffset + 1)
```
ortem, line 52:
```suggestion
        editor.caretModel.moveToOffset(blockExpr.block.lbrace.textOffset + 1)
```

shocoman:
There is also template description in Settings that needs to be changed, if not yet. Though it probably isn't important now
image

ortem:
@shocoman Nice catch, thank you! It should be changed in this PR I think

ortem (approved):
LGTM except for minor comments and changing the description (`after.rs.template`)

Kobzol:
Ok, *now* the template should be correct :sweat: I'll leave bors to you.

Kobzol:
The Qodana unused import warning seems to be a false positive.

Kobzol:
bors retry

Sherlock-Holo:
Is there anyway to disable this feature? sometimes I need to write the arm manually

Kobzol:
No, there isn't currently. Maybe it would be better to create a new template for this, something like "generate bare match"? @ortem

Sherlock-Holo:
> No, there isn't currently. Maybe it would be better to create a new template for this, something like "generate bare match"? @ortem

now if I use the postfix `match`, it will add something like
```rust
match result {
    Ok(_) => {}
    Err(_) => {}
}
```

the cursor is at the first `{}`, now I have to press ← mutli times to rewrite the `_` to something meaningful

Kobzol:
I understand, I also thought about that in this PR. But the situation is not always so easy, for example matching a big enum could produce a lot of placeholders that would all need to be changed, which might be annoying. And also nested bindings will not be generated, so they would need to be changed anyway:

```rust
struct S {
    a: u32,
    b: u32
}

enum Foo {
    V1(u32, u32, u32),
    V2{
        a: S,
        b: (u32, u32)
    }
}

fn foo(x: Foo) {
    match x {
        Foo::V1(_, _, _) => {}  // need to change _, _ and _
        Foo::V2 { .. } => {} // inner bindings need to be updated manually anyway if needed
    }
}
```

Originally, I scrapped this idea because I ran into some technical problems when trying to provide a template builder for the generated `_` patterns, but that could be revisited.

If you think that this is a big problem, I can try to add a template builder that will allow you to change the generated `_` bindings.

Kobzol:
@Sherlock-Holo Would this be ok for you?
video

Sherlock-Holo:
> @Sherlock-Holo Would this be ok for you?
> video
> 
> 

that is better than now

but I  found another problem: if i want to match a String, it will generate something like

match s { String ( ... ) => _}

Kobzol:
Well.. yeah :) String is a just a struct. You can't really match String easily anyway (https://stackoverflow.com/questions/25383488/how-to-match-a-string-against-string-literals). What do you suggest that should be generated for String?

Sherlock-Holo:
> Well.. yeah :) String is a just a struct. You can't really match String easily anyway (https://stackoverflow.com/questions/25383488/how-to-match-a-string-against-string-literals). What do you suggest that should be generated for String?

i am not sure what is the better way, perhaps if user want to match a String, plugin can advise user use .as_str？

Kobzol:
@Sherlock-Holo See https://github.com/intellij-rust/intellij-rust/pull/7346.

Sherlock-Holo:
> @Sherlock-Holo See #7346.

👍 I think that's very excellent

vlad20012:
I must say, I really love this feature.
And I miss a similar feature for Kotlin every day.

Issue #6749, opened 2021-01-28
Incorrectly inferred never type of match expression with diverging arm


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.140.3624-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3.2 Preview Ultimate Edition (IU-203.7148.15)
* **Operating system:** macOS 10.16
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
The inferred type of `num2` in the following example should be `i32`, not `!`
<img width="551" alt="Screenshot 2021-01-28 at 14 22 12" src="https://user-images.githubusercontent.com/4854600/106131830-38ae9c00-6174-11eb-8fc3-477c717a49dc.png">

Might be related to https://github.com/intellij-rust/intellij-rust/issues/1743

## Steps to reproduce
```Rust
fn foo(x: i32) {}

fn main() {
    let num2 = match "42".parse() {
        Ok(v) => v,
        Err(e) => panic!(),
    }; // type of `num2` should be `i32`, not `!`
    foo(num2);
}
```

ortem:
From https://github.com/intellij-rust/intellij-rust/pull/6829#issuecomment-800969813:

[`CoerceMany`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_typeck/check/coercion/struct.CoerceMany.html) should be implemented in order to fix this issue. Currently, IJ Rust uses a simplified and rough approach to handle this (see `RsTypeInferenceWalker#getMoreCompleteType`).

Issue #6797, opened 2021-02-03
Call expressions with outer attributes are parsed incorrectly


ortem:
## Environment

* **IntelliJ Rust plugin version:** 0.3.141.3645-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3.2 Preview Ultimate Edition (IU-203.7148.15)
* **Operating system:** macOS 10.16
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
Call expressions with [outer attributes](https://doc.rust-lang.org/reference/expressions.html#expression-attributes) are parsed incorrectly, with attributes attached to the path, not to the whole call expression.

For example:
```Rust
fn main() {
    #[cfg(unix)] foo(x); // Parsed as `{#[cfg(unix)] foo} {(x)}`. Expected:: `{#[cfg(unix)] foo(x)}`
}
```

<img width="860" alt="Screenshot 2021-02-03 at 19 37 43" src="https://user-images.githubusercontent.com/4854600/106778779-62c1fb80-6657-11eb-85ba-ac7045cd69e6.png">

dima74:
Same problem for `RsTryExpr`:

```rust
fn foo() -> Option<()> {
    let a = Some(1);
    #[cfg(unix)] a?;   // Parsed as `{#[cfg(unix)] a} {?}`. Expected: `{#[cfg(unix)] {a?}}`
    None
}
```
![image](https://user-images.githubusercontent.com/6505554/106861202-81b3a280-66d6-11eb-94be-cb17ed0f828c.png)

Issue #1612, opened 2017-08-11
control flow analysis


farodin91:
Combine multiple analysis part into one analysis per function. Prevent recursive analysis.

Possible parts:
- type inference
- type checking
- borrow checker/ mutable
- unsafe 
- impossible branches

@matklad and @vlad20012  Any suggestions?

vlad20012:
I think we should perform type checking (and possibly all of the above) during type inference and save errors into `RsInferenceContext` (as mapping like `expr => error`). Then extract errors from the context by an error annotator. But I do not think that we can do any type checking right now. Currently I'm working on something called `Expectation`, and after implementing it, it should be possible to perform some limited type checking.

matklad:
@vlad20012 we also need a limited form of control flow analysis in type inference, to deal with `!` type properly: https://github.com/intellij-rust/intellij-rust/issues/1599

PR #6829, opened 2021-02-11
INSP: Implement unreachable code detection

ortem:
This PR introduces `Unreachable code` inspection to detect and remove unreachable code.

unreachable_code

Note that now it might produce some false-positive warnings due to https://github.com/intellij-rust/intellij-rust/issues/6749 and https://github.com/intellij-rust/intellij-rust/issues/6797, so probably it's better to try fixing these issues before merging this. 

Some implementation details:
* Based on the control-flow graph which is already used for several inspections (move analysis, liveness analysis)
* Found unreachable elements are cached by `org.rust.lang.core.types.ExtensionsKt#CONTROL_FLOW_KEY` as well as the  graph itself
* Does not highlight each unreachable element separately, but merges their text ranges instead in order to highlight the widest range

Part of https://github.com/intellij-rust/intellij-rust/issues/1612.

changelog: Add `Unreachable code` inspection and quick-fix

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspection.kt
@@ -1,0 +1,82 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.lints
+
+import com.intellij.openapi.util.Segment
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.refactoring.suggested.stripWhitespace
+import org.rust.ide.injected.isDoctestInjection
+import org.rust.ide.inspections.RsProblemsHolder
+import org.rust.ide.inspections.fixes.SubstituteTextFix
+import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.rangeWithPrevSpace
+import org.rust.lang.core.psi.ext.startOffset
+import org.rust.lang.core.types.controlFlowGraph
+import org.rust.openapiext.document
+import org.rust.stdext.mapToMutableList
+import java.util.*
+
+class RsUnreachableCodeInspection : RsLintInspection() {
+    override fun getDisplayName(): String = "Unreachable code"
+
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnreachableCode
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitFunction(func: RsFunction) {
+            if (func.isDoctestInjection) return
+            val controlFlowGraph = func.controlFlowGraph ?: return
+
+            val elementsToReport = controlFlowGraph
+                .unreachableElements
+                .takeIf { it.isNotEmpty() }
+                ?: return
+
+            // Collect text ranges of unreachable elements and merge them in order to highlight
+            // most enclosing ranges entirely, instead of highlighting each element separately
+            val sortedRanges = elementsToReport
+                .filter { it.isPhysical }
+                .mapToMutableList { it.rangeWithPrevSpace }
+                .apply { sortWith(Segment.BY_START_OFFSET_THEN_END_OFFSET) }
+                .takeIf { it.isNotEmpty() }
+                ?: return
+
+            val mergedRanges = mergeRanges(sortedRanges)
+            for (range in mergedRanges) {
+                registerUnreachableProblem(holder, func, range)
+            }
+        }
+    }
+
+    /** Merges intersecting (including adjacent) text ranges into one */
+    private fun mergeRanges(sortedRanges: List<TextRange>): Collection<TextRange> {
+        val mergedRanges = ArrayDeque<TextRange>()
+        mergedRanges.add(sortedRanges[0])
+        for (range in sortedRanges.drop(1)) {
+            val leftNeighbour = mergedRanges.peek()
+            if (leftNeighbour.intersects(range)) {
+                mergedRanges.pop()
+                mergedRanges.push(leftNeighbour.union(range))
+            } else {
+                mergedRanges.push(range)
+            }
+        }
+        return mergedRanges
+    }
+
+    private fun registerUnreachableProblem(holder: RsProblemsHolder, func: RsFunction, range: TextRange) {
+        val chars = func.containingFile.document?.immutableCharSequence ?: return
+        val strippedRangeInFunction = range.stripWhitespace(chars).shiftLeft(func.startOffset)
+
+        holder.registerLintProblem(
+            func,
+            "Unreachable code",
+            strippedRangeInFunction,
+            SubstituteTextFix.delete("Remove unreachable code", func.containingFile, range)
+        )
+    }
+}
--- a/src/main/kotlin/org/rust/lang/core/dfa/ControlFlowGraph.kt
+++ b/src/main/kotlin/org/rust/lang/core/dfa/ControlFlowGraph.kt
@@ -6,6 +6,7 @@
 package org.rust.lang.core.dfa
 
 import com.intellij.psi.PsiElement
+import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.regions.ScopeTree
@@ -14,6 +15,7 @@ import org.rust.lang.core.types.type
 import org.rust.lang.utils.Node
 import org.rust.lang.utils.PresentableGraph
 import org.rust.lang.utils.PresentableNodeData
+import org.rust.stdext.buildSet
 
 sealed class CFGNodeData(val element: RsElement? = null) : PresentableNodeData {
 
@@ -69,7 +71,8 @@ class ControlFlowGraph private constructor(
     val body: RsBlock,
     val regionScopeTree: ScopeTree,
     val entry: CFGNode,
-    val exit: CFGNode
+    val exit: CFGNode,
+    val unreachableElements: Set<RsElement>
 ) {
     companion object {
         fun buildFor(body: RsBlock, regionScopeTree: ScopeTree): ControlFlowGraph {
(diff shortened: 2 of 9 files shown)


src/test/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspectionTest.kt
```diff
@@ -0,0 +1,218 @@
...
+import org.rust.WithStdlibRustProjectDescriptor
+import org.rust.ide.inspections.RsInspectionsTestBase
+
+class RsUnreachableCodeInspectionTest : RsInspectionsTestBase(RsUnreachableCodeInspection::class) {
+    fun `test straightline with return`() = checkByText("""
+        fn main() {
+            let x = 1;
+            return;<warning descr="Unreachable code">
+            let y = 2;
+            let z = 3;</warning>
```
dima74, line 19: Can we exclude first PsiWhitespace during merging ranges?

src/main/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspection.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.rust.stdext.mapToMutableList
+import java.util.*
+
+class RsUnreachableCodeInspection : RsLintInspection() {
+    override fun getDisplayName(): String = "Unreachable code"
+
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnreachableCode
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
+        override fun visitFunction(func: RsFunction) {
```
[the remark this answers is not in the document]
ortem: Yes, such cases should be supported someday as well. The same about move analysis.
I mention this in https://github.com/intellij-rust/intellij-rust/issues/1612, thank you for clarification

src/main/kotlin/org/rust/lang/core/dfa/ControlFlowGraph.kt
```diff
@@ -84,13 +89,103 @@ class ControlFlowGraph private constructor(
...
+                }
+
+            /**
+             * If [unexecuted] produces termination by itself
+             * (which means its inner expression type is [TyNever], see the corresponding checks below),
+             * it should be treated as unreachable only if the execution does not reaches the beginning of [unexecuted].
+             * The latter means that previous statement is not fully executed.
+             * This is needed in order not to treat the first, basically reachable, `return` as unreachable
+             */
+            fun isUnreachable(unexecuted: RsElement, innerExpr: RsExpr?): Boolean {
```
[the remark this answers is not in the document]
ortem: Well, this is an interesting one. The type of the loop expression is `()`, i.e.:
```Rust
let _: () = loop {
    return;
    break;
};
let _: ! = loop {
    return;
};
```
So my heuristic `Unexecuted element is definitely unreachable if its inner expression does not produce termination` does not work here.

I'm not sure if this situation is common and how we should actually deal with it. I guess it is quite rare and related only to loops, so as a possible solution, we can extend our check from
```Kotlin
if (innerExpr != null && innerExpr.type !is TyNever) { ... }
```
to
```Kotlin
if (innerExpr != null && innerExpr.type !is TyNever && innerExpr !is RsLooplikeExpr) { ... }
```
but this might affect other cases of unreachable `RsLooplikeExpr`. On the other hand, if this situation is very rare, we can just ignore it for now. What do you think?

src/test/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspectionTest.kt
ortem: Yes, I think we can, but I intentionally leave it as is for now.
Do you think it would be better to exclude it? Maybe there are some real cases when it is necessary for good UI/UX?

src/main/kotlin/org/rust/lang/core/dfa/ControlFlowGraph.kt
```diff
@@ -84,13 +89,103 @@ class ControlFlowGraph private constructor(
...
+                graph.forEachNode { node ->
+                    if (node.index !in fullyExecutedNodeIndices) {
+                        val element = node.data.element ?: return@forEachNode
+                        add(element)
+                    }
+                }
+            }
+            val unexecutedStmts = unexecutedElements
+                .filterIsInstance<RsStmt>()
+            val unexecutedTailExprs = unexecutedElements
```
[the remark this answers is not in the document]
ortem: Good catch, thank you!
* Added `override fun visitUnitExpr` to `CFGBuilder` to consider `()` in control flow
* Fixed `unexecutedTailExprs` collecting to handle macro calls

src/test/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspectionTest.kt
dima74: > some real cases when it is necessary for good UI/UX?

When using "next error" (F2) it will navigate to the correct line

src/main/kotlin/org/rust/lang/core/dfa/ControlFlowGraph.kt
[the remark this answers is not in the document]
dima74: I agree that this case is rare so we can ignore it for now

Btw can't we extend heuristic so that any parent of executed element should be not marked as unreachable? Looks like In this example `return` is executed, so `return;` and `loop` will not be marked as unreachable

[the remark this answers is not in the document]
ortem: Do I understand correctly that you are talking about replacing the current `innerExpr` checking with something like this?
```Kotlin
if (unexecuted.children.any { it !in unexecutedElements }) {
    return false
}
```
If so, I'm afraid that would have an unreasonable negative impact on the algorithmic complexity. What do you think?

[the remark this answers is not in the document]
dima74: > If so, I'm afraid that would have an unreasonable negative impact on the algorithmic complexity.

Ok, so lets indeed ignore such cases for now since they are rare

src/test/kotlin/org/rust/ide/inspections/lints/RsUnreachableCodeInspectionTest.kt
ortem: Fixed in a separate temporary commit for more convenience during review. Please let me know if the fix looks good to you, and I will finally rebase the commits

dima74: Yes, fix looks good

dima74 (approved):

ortem:
Note, although no regressions are found with CI (because the regression tests only search for error annotations, not warnings), there are some false positive warnings in real projects, mostly due to #6749, as I mentioned in the PR description. 

Here is what I found:

- **cargo**: no false-positive warnings
- **diesel**: 1 file affected due to unsupported `#[allow(warnings)]` attribute #5888
- **tokio**: 3 files affected due to #6749, 1 file affected due to an unevaluated cfg option
- **rust-analyzer**: 4 files affected due to #6749

As we just briefly investigated with @vlad20012, [`CoerceMany`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_typeck/check/coercion/struct.CoerceMany.html) should be implemented in order to fix #6749. Currently, IJ Rust uses a simplified and rough approach to handle this (see `RsTypeInferenceWalker#getMoreCompleteType`). So #6749 is probably not very easy to fix right now, and I don't know a good and simple workaround to avoid analyzing such problematic `match` expressions since many of them are expanded from macros.

However, I think we can merge this inspection because the number of false positives is not so much. @Undin WDYT?

Undin:
> However, I think we can merge this inspection because the number of false positives is not so much. @Undin WDYT?

As we've discussed in person - let's try to fix type inference first not to introduce new false positive annotations for users. Looks like we can come up with at least some hack to make type inference work in described cases.
As an alternative solution, I may suggest merging the inspection right now but disabled by default. And enable it when type inference will be fixed. It may help to avoid merge conflicts

Undin:
@ortem BTW tests fail

Undin:
@ortem #6749 is fixed so I suppose we can merge this PR

ortem:
@Undin I'd like not to merge this PR right now because some false-positive warnings remained, with #6749 fixed:

- **cargo**: no false-positive warnings
- **diesel**: 1 file affected due to unsupported `#[allow(warnings)]` attribute #5888
- **tokio**: 2 files are still affected due to `tokio::select!` macro, 1 file affected due to an unevaluated cfg option
- **rust-analyzer**: no false-positive warnings

So the most interesting case is `tokio`. I've found a problem with the processing of `break` expanded from a macro:
```Rust
macro_rules! break_macro {
    () => { break };
}

fn main() {
    loop {
        break_macro!();
    }
    1; // false-positive "unreachable"
}
```

I'm investigating the problem. If I don't find a solution soon then I'll disable the inspection by default and merge this PR.

ortem:
I opened an issue for this case: https://github.com/intellij-rust/intellij-rust/issues/7016.
Now it is fixed and there are no false positives with `tokio::select!` macro in `tokio`. So this can be merged I suppose.

Undin:
@ortem don't insist but I'd personally prefer to create a separate PR for #7016. For example, it's convenient for changelog creation

ortem:
bors r=dima74

src/test/kotlin/org/rust/lang/core/dfa/RsControlFlowGraphTest.kt
```diff
@@ -1098,6 +1098,6 @@ class RsControlFlowGraphTest : RsTestBase() {
         val cfg = ControlFlowGraph.buildFor(function.block!!, getRegionScopeTree(function))
         val expected = expectedIndented.trimIndent()
         val actual = cfg.graph.depthFirstTraversalTrace(cfg.entry)
-        check(actual == expected) { throw ComparisonFailure("Comparision failed", expected, actual) }
+        check(actual == expected) { throw ComparisonFailure("Comparison failed", expected, actual) }
```
Undin, line 1101: BTW why we don't use `assertEquals` here?

ortem: I'm not sure, probably I've just copied this code from another test.
Similar code appears in `RsMacroGraphWalkerTest` and `RsMacroGraphBuilderTest` as well, so I can open another PR to change them all.

ortem: Done: https://github.com/intellij-rust/intellij-rust/pull/7027

Issue #2576, opened 2018-06-07
Create struct/method from usage refactoring


JekoTronik:
In my TDD workflow, I start by creating a test in my lib.rs with **tmod** and **tfn**
Which I will complete as follow :
```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn empty_shopping_list() {
        let basket = ShoppingBasket { errands: "" };
        assert_eq!(basket.view(), "Sales Taxes: 0.0 Total: 0.0");
    }
}
```

Then, I would like to extract a struct using some shortcut on **"ShoppingBasket"** word. It would add this at the beginning of my file :
```rust
struct ShoppingBasket {
    errands: ,
}
```
Could add `#[derive(Debug)]` also and suggest errands'type.
Then, I would like to implements a function for the newly created struct using another shortcut on **"basket.view()"**. As a result it would add this after the structure definition : 
```rust
impl ShoppingBasket {
    fn view(&self) -> &str {
        ""
    }
}
```

Is that possible actually ? Or something similar ? No need to create module nor extra file actually.

dima74:
Creating method from usage is now implemented thanks to #6039

PR #5650, opened 2020-06-24
INT: replace usages in DestructureIntention

Kobzol:
This is a revival of an older PR by @fan-tom (https://github.com/intellij-rust/intellij-rust/pull/4095). It adds support for replacing usages of field/tuple bindings after performing `DestructureIntention`. Both field lookup replacement (`s.0 -> _0`) and whole item replacement (`foo(s) -> foo((_0, _1))`) is supported. Method calls on the original binding are ignored, as the user probably doesn't want to transform them to
`S { x, y }.foo()`.

replace

To rename multiple introduced bindings at once, I needed something that combines TemplateBuilder (renames multiple things in a single editing session) and `RsInPlaceVariableIntroducer` (renames occurrences of a renamed item). I created `RsMultipleVariableIntroducer`, which invokes the `RsInPlaceVariableIntroducer` repeatedly. It's probably a horrible hack, but it seems to work (at least on my machine :D ).

I rebased the original PR on master, moved reference search from `findAvailableContext` to `invoke` and used a local search scope.
CC @fan-tom

Closes: https://github.com/intellij-rust/intellij-rust/pull/4095
Fixes: https://github.com/intellij-rust/intellij-rust/issues/3684

PR #6837, opened 2021-02-12
INT: add intention to create a struct

Kobzol:
This PR adds an intention to create a struct from an unresolved struct literal (similar to https://github.com/intellij-rust/intellij-rust/pull/5505).

create-struct

I'm not sure what template to offer the user after the struct is created. In the intention for creating functions, we create a template builder that allows the user to change the name and type of each parameter. However, here a template builder is not good enough, because if the user decides to rename the field, the usage from which the struct was created would suddenly contain an invalid field name. Therefore, I borrowed `RsMultipleVariableRenamer` from https://github.com/intellij-rust/intellij-rust/pull/5650 to propagate the field name changes to the struct literal. But with this, it doesn't allow the user to change the field type.

Should I also add support for creating tuples structs in this PR? It could probably share a lot of code from `CreateFunctionIntention`.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/2576

changelog: Add an intention to create a struct from an unresolved struct literal usage.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateFunctionIntention.kt
@@ -3,14 +3,14 @@
  * found in the LICENSE file.
  */
 
-package org.rust.ide.intentions
+package org.rust.ide.intentions.createFromUsage
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
-import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
 import com.intellij.psi.util.parentOfType
 import org.rust.cargo.project.workspace.PackageOrigin
+import org.rust.ide.intentions.RsElementBaseIntentionAction
 import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.utils.GenericConstraints
 import org.rust.lang.core.psi.*
@@ -36,7 +36,7 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
         }
     }
 
-    sealed class Context(val name: String, val callElement: PsiElement) {
+    sealed class Context(val name: String, val callElement: RsElement) {
         abstract val visibility: String
         open val isAsync: Boolean = callElement.isAtLeastEdition2018
         abstract val arguments: RsValueArgumentList
@@ -44,11 +44,7 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
         open val implItem: RsImplItem? = null
 
         open class Function(callExpr: RsCallExpr, name: String, val module: RsMod) : Context(name, callExpr) {
-            override val visibility: String = when {
-                callExpr.containingCrate != module.containingCrate -> "pub "
-                callExpr.containingMod != module -> "pub(crate) "
-                else -> ""
-            }
+            override val visibility: String = getVisibility(module, callExpr.containingMod)
             override val isAsync: Boolean = super.isAsync
                 && (callExpr.parent as? RsDotExpr)?.fieldLookup?.isAsync == true
             override val arguments: RsValueArgumentList = callExpr.valueArgumentList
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateFunctionIntention.kt
```diff
@@ -267,8 +253,10 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
             }
         }
 
-        val newImpl = RsPsiFactory(item.project).createInherentImplItem(item.name
-            ?: "?", item.typeParameterList, item.whereClause)
+        val newImpl = RsPsiFactory(item.project).createInherentImplItem(
+            item.name
+                ?: "?", item.typeParameterList, item.whereClause
+        )
```
vlad20012, line 259: This is quite a weird formatting. Is it a work of our formatter? =D 
Let's just put parameters on separate lines and leave `item.name ?: "?"` on one line.

src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateStructIntention.kt
```diff
@@ -0,0 +1,96 @@
...
+        val fieldList = ctx.literalElement.structLiteralBody.structLiteralFieldList
+        val bindings = ctx.literalElement.getVisibleBindings()
+        val visibility = getVisibility(ctx.target, ctx.literalElement.containingMod)
+
+        val fieldsJoined = fieldList.joinToString(separator = ",\n") {
+            val name = it.referenceName
+            val expr = it.expr
+            val type = when {
+                expr != null -> expr.type
+                else -> bindings[name]?.type ?: TyUnknown
```
vlad20012, line 72: I improved the name resolution for this case in #7132. Now it's possible to use `RsStructLiteralField.resolveToBinding()` instead of hacks with `bindings` map

src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateStructIntention.kt
```diff
@@ -0,0 +1,96 @@
...
+            fieldsJoined.isEmpty() -> ";"
+            else -> " {\n$fieldsJoined\n}"
+        }
+
+        return factory.tryCreateStruct("${visibility}struct ${ctx.name}$suffix")
+    }
+
+    private fun insertStruct(targetModule: RsMod, struct: RsStructItem, sourceFunction: RsElement): RsStructItem {
+        if (targetModule == sourceFunction.containingMod) {
+            return sourceFunction.parent.addBefore(struct, sourceFunction) as RsStructItem
```
vlad20012, line 87: Why do we add a struct before the function? There is a bit similar `Create class` quickfix in Kotlin and it adds classes after

vlad20012 (approved):
> But with this, it doesn't allow the user to change the field type.

Ideally, we should allow specifying field names and types (like in Create Function intention), but this is ok for start, I guess.

> Should I also add support for creating tuples structs in this PR? It could probably share a lot of code from `CreateFunctionIntention`.

A separate PR is better, I guess. And yes, it sounds like a little `CreateFunctionIntention` extension =)

src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateStructIntention.kt
Kobzol: Great! Thanks.

src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateStructIntention.kt
Kobzol: For me the logical order is structs -> functions that use them. But maybe I'm too affected by one-pass compilers (such as C/C++) :) Do you want to switch and create the struct after the function? It's true that create function intention also creates the target function after the caller.

vlad20012: I feel there is some "unspoken rule" that IDEA tends to generate stuff below the place where such generation is invoked. But `structs -> functions` also sounds reasonable :thinking: 
Well, let's leave it as is for now. Nevermind

Kobzol:
Let's merge after https://github.com/intellij-rust/intellij-rust/pull/7132.

Kobzol:
bors r=vlad20012

PR #6838, opened 2021-02-13
INT: Convert local named function to lambda

### LINKED ISSUE #6711: Convert function to closure (and vice versa) intentions

nschoellhorn:
Related: #6813 handles the other way around. This PR was originally integrated in the other one, but I split it up into two since they will get a little more complex than originally anticipated.

Fixes #6711

changelog: Add intention to convert local named functions to lambdas


nschoellhorn:
Comments copied over from the previously combined PR:

@Stzx 

> My example may be a bit extreme, we don't normally write code like this.
> 
> For code that needs to use move explicitly:
> 
> ```rust
> fn foo(a: f64) -> impl Fn(f64) -> f64 {
>     fn to(x: f64) -> f64 {
>         a + x
>     }
> }
> fn foo(a: f64) -> impl Fn(f64) -> f64 {
>     let to = |x: f64| -> f64 { a + x };
>     //      ^ The move keyword should be used explicitly here
>     to
> }
> ```
> But for function using this conversion, the implicit move can already handle most of the problems.
> 
> About asynchrony. I guess if the maintainer agrees, I have no problem.

@Kobzol 
> * What if the function had some generic parameters?
> 
> ```rust
> fn foo<T>(a: T) {}
> ```
> 
> In this case parameter type `T` in the closure will have an invalid type, and Rust doesn't support generic closures. I suggest to remove the type reference from the parameter in such case. I think that the easiest way how to do that would be to generate the closure, then go through the type references of its parameters and if they are invalid (e.g. they contain `TyUnknown`), remove them and hope that Rust's inference will take care of the rest.

vlad20012:
Thanks!
I made the intention unavailable for generic functions.
bors r+

Issue #5384, opened 2020-05-15
Highlight implicit usages of Copy trait


Exidex:
It would be very useful for beginners to explicitly show usages of Copy trait (I am beginner myself, but I think that is what's happening in code snippet).

For example, here usages of `copy_item` could be highlighted
```
fn main() {
    let copy_item = "test";

    Some(copy_item);
    Some(copy_item);
}
```
 Maybe similar to how Intellij highlights Java variables captured by closure:
![image](https://user-images.githubusercontent.com/16986685/82098729-d2790a80-9705-11ea-8501-ded44190dd0c.png)

kumbayo:
There is a very similar issue #2029 that talks about marking `Copy` bindings / literals.

vlad20012:
Technically it's much similar to #3013 (that is almost done) but even simpler.
Some annotator just should annotate `RsPath` consist of a single segment that is resolved to a local binding of `Copy`able type.

The most complex question is about how such highlighting should look like.

koriwi:
Any progress on this or similar issues?

PR #6841, opened 2021-02-13
ANN: add highlighter for implicit Copy usages

Kobzol:
This PR adds a simple highlighter for implicit usages of Copy. There are two main questions:
1) How should the highlighting look like? I experimented with a few options and either some light background color or **bold** seem reasonable. We can also leave it without a default highlighting and let the users style it themselves.
2) Currently the highlighting annotates all path expressions of a Copy type. A better approach would probably be to only mark usages that would actually require a move if the type is not Copy. @ortem can the existing move analysis/borrow checker infrastructure be used for something like `isMove(RsExpr)`?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5384

changelog: Highlight implicit usages of the Copy trait.


vlad20012:
I thought a little more and realized that my [comment](https://github.com/intellij-rust/intellij-rust/issues/5384#issuecomment-630723543) was not too useful and we really should highlight only such usages that would lead to a compilation error if the type is not Copy (i.e. a "second" usage of some value). I think the only @ortem knows how to achieve this.

Also, not all usages is a copy. For example, borrowing is not copying.
Also, not only local bindings can be copied, but also, for example, struct fields or a boxed value (with `*a`). Move analysis should be aware of all possible cases.

So, well, the issue is much more complex than meets the eye =/

And also, I thought that it might be useful to do the opposite thing - highlight moves of values (when these values can't be used after that move). This sounds a bit simpler, but I'm not sure how to use Move Analysis infrastructure to achieve this.

Kobzol:
Right. I agree that highlighting moves would probably be much more useful than highlighting copies. I think that we have already discussed with @ortem in some issue or PR that there ideally should be some API to find actual moves in code.

vlad20012:
`ExprUseWalker` and `ConsumeMode` slightly look like the API we looking for.

Issue #1323, opened 2017-06-08
Extend SelfConvention


farodin91:
* with_ for builders take no ref
* get_ take ref
* set_ take mut ref

PR #6842, opened 2021-02-13
INSP: extend RsSelfConventionInspection with get, set and with

Kobzol:
This PR adds more cases to `RsSelfConventionInspection`, useful for getters/setters and the builder pattern.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/1323

changelog: Extend RsSelfConvention inspection with new method prefixes (get, set and with).

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSelfConventionInspection.kt
@@ -12,15 +12,13 @@ import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ImplLookup
 import org.rust.lang.core.types.selfType
 import org.rust.lang.core.types.ty.TyUnknown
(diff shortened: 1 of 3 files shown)


mchernyavsky (approved):
Could you rebase onto master, please?

Kobzol:
bors r=mchernyavsky

a1ien:
I think `with_` suggestion was incorrect. Сounterexample `Vec::with_capacity`, `HashMap::with_capacity`, `HashMap::with_hasher`
I think RsSelfConvention inspection `with` must be removed.

Kobzol:
The current inspection is indeed wrong in some cases. We should probably rework it to work in the same way as the corresponding Clippy inspection (https://github.com/intellij-rust/intellij-rust/issues/7475).

a1ien:
Or you incorrect impalement check. In issues says `with_` method as builder. So the must not take `self` at all.

PR #6852, opened 2021-02-15
CI: fix next milestone creation

Undin:
Previously, to get collect the latest milestones, we didn't specify `sort` and `direction` query params.
And GitHub API had might return milestones without the latest one (since GitHub uses pagination).
And as a result, "make next milestone" step on CI had might create next milestone without due_on date and with an incorrect release manager

Now both parameters are specified to make GitHub return the latest milestones by `due_on` property


Undin:
@lancelote haven't assigned it to you since you are on vacation. Just want to let you know about these changes

lancelote:
bors merge

Issue #5165, opened 2020-04-01
False positive "Cannot move" inspection on static variable that has Copy trait


aloucks:
## Environment

* **IntelliJ Rust plugin version:** Rust (0.2.118.2171-193)
* **Rust toolchain version:** rustc 1.42.0 (b8cedc004 2020-03-09)
* **IDE name and version:**  IntelliJ IDEA 2019.3.4 Build #IC-193.6911.18 March 17, 2020
* **Operating system:** Windows 10 amd64

I've noticed this with the `glfw` crate's global `FAIL_ON_ERRORS` static variable.

![image](https://user-images.githubusercontent.com/221559/77493663-180e0b00-6e1a-11ea-809d-8c6e12f7894d.png)

The type is `Copy`.

https://github.com/PistonDevelopers/glfw-rs/blob/c51524b52e2e5cd7411b0e7d1cfe3e4eb110d4d6/src/lib.rs#L428

https://github.com/PistonDevelopers/glfw-rs/blob/c51524b52e2e5cd7411b0e7d1cfe3e4eb110d4d6/src/lib.rs#L377

EDIT: This is in the most recent IntelliJ Rust as of 2020-03-34 on Windows 10.

_Originally posted by @aloucks in https://github.com/intellij-rust/intellij-rust/issues/4850#issuecomment-603596901_

ambye85:
![Screenshot 2020-05-18 at 10 34 10](https://user-images.githubusercontent.com/17502031/82197705-36145b00-98f3-11ea-8de9-d1ea6524da70.png)

Same issue with plugin version 0.2.121.3059-201-stable on:

CLion 2020.1.1
Build #CL-201.7223.86, built on April 29, 2020
Runtime version: 11.0.6+8-b765.40 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
macOS 10.15.4

codedcosmos:
Still an issue with the rust plugin from JetBrains 0.3.139.3615-203

IntelliJ IDEA 2020.3 (Community Edition)
Build #IC-203.5981.155, built on December 1, 2020
Runtime version: 11.0.9+11-b1145.21 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Linux 5.8.0-43-generic
GC: ParNew, ConcurrentMarkSweep
Non-Bundled Plugins: GLSL, org.toml.lang, org.rust.lang

ortem:
[...]

PR #6855, opened 2021-02-16
Provide builtin Clone & Copy impls for TyFunction

ortem:
Previously, `TyFunction` was considered as non-Copy and non-Clone type in terms of our type inference. Although there was a special case in `Ty.isMovesByDefault` to handle `TyFunction` properly during move analysis, it did not work in case of `TyFunction` used as a type parameter.
Now `TyFunction` is considered as `Copy` & `Clone` type so the special case in `Ty.isMovesByDefault` has been removed.

More information in Rust docs: [builtin `Clone` impls](https://doc.rust-lang.org/std/clone/trait.Clone.html#additional-implementors) and [builtin `Copy` impls](https://doc.rust-lang.org/std/marker/trait.Copy.html#additional-implementors).

Fixes https://github.com/intellij-rust/intellij-rust/issues/5165

changelog: Fix false-positive move errors on a function pointer type used as a type parameter

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/ImplLookup.kt
@@ -365,7 +365,7 @@ class ImplLookup(
      * @see <a href="https://doc.rust-lang.org/std/marker/trait.Copy.html#additional-implementors">Copy additional implementors</a>
      */
     private fun getHardcodedImpls(ty: Ty): Collection<BoundElement<RsTraitItem>> {
-        if (ty is TyTuple || ty is TyArray) {
+        if (ty is TyTuple || ty is TyArray || ty is TyFunction) {
             return listOfNotNull(items.Clone, items.Copy).map { BoundElement(it) }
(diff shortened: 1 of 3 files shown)


Undin (commented):
Note, our current implementation of type inference returns `TyFunction` as a type of closures that not always implement `Copy` and `Clone`

vlad20012 (approved):
bors r+

ortem:
@Undin Yes, thank you for clarification.

Regarding move analysis, treating all `TyFunction` as `Clone` & `Copy` leads to some false negatives related to moved closures. But this PR does not change this logic because previously there was a corresponding special case in `Ty.isMovesByDefault` actually.

As for name resolution and code completion, this PR might change the behavior a bit, but I think it's rather insignificant since closure support is still weak in the plugin.

@vlad20012 WDYT?

vlad20012:
> As for name resolution and code completion, this PR might change the behavior a bit, but I think it's rather insignificant since closure support is still weak in the plugin.

I agree with it

PR #6869, opened 2021-02-19
Make native helper binary executable

Undin:
Found out all binaries of native-helper lose executable permissions in two places: during uploading/downloading artifacts on CI and during unzipping plugin archive. As a result, all features that may use native helper (fetching info about items generated by build scripts and future proc macro support #6564) don't work

These changes restore executable permissions manually.

changelog: Fix fetching info about items generated by build scripts. Note, the corresponding feature work only when `org.rust.cargo.evaluate.build.scripts` experimental feature is enabled


Undin:
@lancelote could you check it for all OSs during testing, please?

Issue #6873, opened 2021-02-20
`Extract Function` doesn't handle the default type arguments correctly


mchernyavsky:
## Environment

* **IntelliJ Rust plugin version:** 0.3.142.SNAPSHOT-203-dev
* **Rust toolchain version:** 1.49.0 (e1884a8e3 2020-12-29) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2020.3.2 Ultimate Edition (IU-203.7148.57)
* **Operating system:** macOS 10.15.7
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

<img width="749" alt="Screenshot 2021-02-20 at 13 50 42" src="https://user-images.githubusercontent.com/6079006/108593094-c1d77e00-7382-11eb-9eb0-d780d9a59394.png">

Expected signature: `fn (s: S<u32, i32>)`
Actual signature (with defaults): `fn (s: S<i32, u32>)`

## Steps to reproduce
```rust
struct S<R=u32, T=u32>(R, T);

fn main() {
    let s = S(1u32, 2i32);
    s;
}

```

PR #6874, opened 2021-02-20
REF: Omit unchanged default type argument only if the next type argument is already ommited

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6873.

changelog: TODO


Kobzol (approved):
Oh, that's a very obvious error in retrospect :D I wonder how it wasn't caught earlier. LGTM.

mchernyavsky (commented):
bors r+

PR #6876, opened 2021-02-20
Do not expand builtin macros as "{}"

vlad20012:
Initial groundwork for built-in macros support (like `concat!()`).
Now these macros are not expanded as "{}" (as before)

changelog: INTERNAL: Add initial infrastructure for builtin macro expansion, FIX: Do not expand builtin macros as "{}"


vlad20012:
bors retry

PR #6881, opened 2021-02-21
Simplify bit flags creation

vlad20012:
Internal. Simplifies bitmask creation and makes it less error-prone. 

How we crated bitmasks before:

```kotlin
companion object {
    private val NAMESPACE_TYPES_MASK: Int = makeBitMask(0)
    private val NAMESPACE_VALUES_MASK: Int = makeBitMask(1)
    private val NAMESPACE_MACROS_MASK: Int = makeBitMask(2)
}
```

And how we can create them now:

```kotlin
companion object : BitFlagsBuilder(Limit.BYTE) {
    private val NAMESPACE_TYPES_MASK: Int = nextBitMask()
    private val NAMESPACE_VALUES_MASK: Int = nextBitMask()
    private val NAMESPACE_MACROS_MASK: Int = nextBitMask()
}
```

Also, it's possible to "extend" another bitmask:

```kotlin
object Foo : BitFlagsBuilder(Limit.BYTE) {
    private val ONE: Int = nextBitMask() // 0
    private val TWO: Int = nextBitMask() // 1
}

object Bar : BitFlagsBuilder(Foo, Limit.BYTE) { // Note `Foo` here!
    private val THREE: Int = nextBitMask() // 2
    private val FOUR: Int = nextBitMask() // 3
}
```

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
+++ b/src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
@@ -35,8 +35,10 @@ import org.rust.lang.core.stubs.BlockMayHaveStubsHeuristic.getAndClearCached
 import org.rust.lang.core.types.ty.TyFloat
 import org.rust.lang.core.types.ty.TyInteger
 import org.rust.openapiext.ancestors
+import org.rust.stdext.BitFlagsBuilder
+import org.rust.stdext.BitFlagsBuilder.Limit.BYTE
+import org.rust.stdext.BitFlagsBuilder.Limit.INT
 import org.rust.stdext.HashCode
-import org.rust.stdext.makeBitMask
 import org.rust.stdext.readHashCodeNullable
 import org.rust.stdext.writeHashCodeNullable
 
@@ -120,8 +122,8 @@ class RsFileStub(
 //        }
     }
 
-    companion object {
-        private val MAY_HAVE_STDLIB_ATTRIBUTES_MASK: Int = makeBitMask(RsAttributeOwnerStub.USED_BITS + 0)
+    companion object : BitFlagsBuilder(RsAttributeOwnerStub, BYTE) {
+        private val MAY_HAVE_STDLIB_ATTRIBUTES_MASK: Int = nextBitMask()
     }
 }
 
@@ -402,8 +404,8 @@ class RsUseItemStub(
         }
     }
 
-    companion object {
-        private val HAS_PRELUDE_IMPORT_MASK: Int = makeBitMask(RsAttributeOwnerStub.USED_BITS + 0)
+    companion object : BitFlagsBuilder(RsAttributeOwnerStub, BYTE) {
+        private val HAS_PRELUDE_IMPORT_MASK: Int = nextBitMask()
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/stubs/StubImplementations.kt
```diff
@@ -816,15 +817,15 @@ class RsFunctionStub(
...
-        private val ABSTRACT_MASK: Int =           makeBitMask(RsAttributeOwnerStub.USED_BITS + 0)
-        private val CONST_MASK: Int =              makeBitMask(RsAttributeOwnerStub.USED_BITS + 1)
-        private val UNSAFE_MASK: Int =             makeBitMask(RsAttributeOwnerStub.USED_BITS + 2)
-        private val EXTERN_MASK: Int =             makeBitMask(RsAttributeOwnerStub.USED_BITS + 3)
-        private val VARIADIC_MASK: Int =           makeBitMask(RsAttributeOwnerStub.USED_BITS + 4)
-        private val ASYNC_MASK: Int =              makeBitMask(RsAttributeOwnerStub.USED_BITS + 5)
-        private val HAS_SELF_PARAMETER_MASK: Int = makeBitMask(RsAttributeOwnerStub.USED_BITS + 6)
-        private val IS_PROC_MACRO_DEF: Int =       makeBitMask(RsAttributeOwnerStub.USED_BITS + 7)
+    companion object : BitFlagsBuilder(RsAttributeOwnerStub, INT) {
+        private val ABSTRACT_MASK: Int =           nextBitMask()
```
Undin, line 821: Let's use the same formatting style everywhere. Currently, we have two different ones

Undin, line 821: If you decide to use this style, let's
```suggestion
        private val ABSTRACT_MASK: Int           = nextBitMask()
```

Undin: Also, if you decide to use this style, let's surround this group with `// @formatter:off` and `// @formatter:on` comments to prevent accident formatting of this region

Undin (approved):
In general, looks ok for me.
Only one minor concern: now you can't just reorder two fields/add new field in non-last position in companion object which are initialized with `nextBitMask` because it changes semantic

src/main/kotlin/org/rust/stdext/BitFlagsBuilder.kt
```diff
@@ -0,0 +1,35 @@
...
+ *     val BIT_MASK_0 = nextBitMask() // Equivalent to `1 shl 0`
+ *     val BIT_MASK_1 = nextBitMask() // Equivalent to `1 shl 1`
+ *     val BIT_MASK_2 = nextBitMask() // Equivalent to `1 shl 2`
+ *     // ...etc
+ * }
+ * ```
+ */
+abstract class BitFlagsBuilder private constructor(private val limit: Limit, startFromBit: Int) {
+    protected constructor(limit: Limit) : this(limit, 0)
+    protected constructor(prevBuilder: BitFlagsBuilder, limit: Limit) : this(limit, prevBuilder.counter)
```
dima74, line 22: Since we use `Byte` almost anywhere, maybe set it as default parameter

```suggestion
    protected constructor(limit: Limit = Limit.Byte) : this(limit, 0)
    protected constructor(prevBuilder: BitFlagsBuilder, limit: Limit = Limit.Byte) : this(limit, prevBuilder.counter)
```

dima74 (approved):

src/main/kotlin/org/rust/stdext/BitFlagsBuilder.kt
vlad20012: I'd want it to be more explicit :thinking:

matklad:
Cursed, I like it :+1:!

PR #6888, opened 2021-02-24
Expand `concat!()` builtin macro

vlad20012:
Depends on #6876

Expand `concat!()` builtin macro


vlad20012:
I forget about nested macros here `concat!(foo!(), "bar")`, so let's do it later

PR #6891, opened 2021-02-24
avoid empty test tool window on the Code With Me client (CWM-2044)

akozlova:
changelog: if run configuration returns non-empty `SMTRunnerConsoleProperties`, Code With Me client treats it as tests and reserves test toolwindow for it. The suggested fix ensures that properties are provided only for the case when test toolwindow  is shown on host

https://youtrack.jetbrains.com/issue/CWM-2044

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
@@ -146,7 +146,7 @@ open class CargoCommandConfiguration(
         }
     }
 
-    private fun showTestToolWindow() = command.startsWith("test") &&
+    private fun showTestToolWindow(): Boolean = command.startsWith("test") &&
         isFeatureEnabled(RsExperiments.TEST_TOOL_WINDOW) &&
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
```diff
@@ -134,15 +139,22 @@ open class CargoCommandConfiguration(
...
-            isFeatureEnabled(RsExperiments.TEST_TOOL_WINDOW) &&
-            !command.contains("--nocapture")) {
+        return if (showTestToolWindow()) {
             CargoTestRunState(environment, this, config)
         } else {
             CargoRunState(environment, this, config)
         }
     }
 
+    private fun showTestToolWindow() = command.startsWith("test") &&
```
Undin, line 149: Let's specify returning type explicitly here

Undin (approved):
Great!
Thank you!
And welcome to IntelliJ Rust contributors! 🎉 

Could you add your github nickname to `CONTRIBUTORS.txt`, please?

bors d+

PR #6892, opened 2021-02-25
RES: Add method to RsPathReference to resolve only public items

dima74:
Adds methods `resolveIfVisible` and `multiResolveIfVisible` to `RsPathReference` class (works only with new resolve; with old resolve they are equivalent to `resolve` and `multiResolve`). They can be used in refactorings and intentions, e.g. to fix "A second item with name ... imported" false-positive:

```rust
mod mod1 {
    fn foo() {}
    pub struct foo {}
}
mod mod2 {
    pub fn foo() {}
}

use mod1::foo;  // foo func is private, so no conflict with `mod2::foo`
use mod2::foo;
```

changelog: Fix false-positive [E0252](https://doc.rust-lang.org/error-index.html#E0252) "A second item with name ... imported" error for private items

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
@@ -10,9 +10,11 @@ import com.intellij.util.SmartList
 import org.rust.lang.core.completion.RsCompletionContext
 import org.rust.lang.core.completion.createLookupElement
 import org.rust.lang.core.psi.RsFunction
+import org.rust.lang.core.psi.RsPath
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.resolve.ref.MethodResolveVariant
 import org.rust.lang.core.types.BoundElement
+import org.rust.lang.core.types.BoundElementWithVisibility
 import org.rust.lang.core.types.Substitution
 import org.rust.lang.core.types.emptySubstitution
 import org.rust.lang.core.types.ty.Ty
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReference.kt
```diff
@@ -12,6 +12,10 @@ import org.rust.lang.core.types.BoundElement
 interface RsPathReference : RsReference {
     override fun getElement(): RsPath
 
+    fun resolvePublicOnly(): RsElement? = multiResolvePublicOnly().singleOrNull()
```
Undin, line 15: I think `publicOnly` is a confusing name since as I understand it resolves visible items that can have different visibility

dima74: Good point, what do you think about `resolveCheckVisibility`?

dima74: Replaced to `resolveIfVisible`

src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -94,7 +97,9 @@ fun collectPathResolveVariants(
         if (e.name == referenceName) {
             val element = e.element ?: return@createProcessor false
             if (element !is RsDocAndAttributeOwner || element.isEnabledByCfgSelf) {
-                result += BoundElement(element, e.subst)
+                val boundElement = BoundElement(element, e.subst)
+                val isPublic = e.isVisibleFrom(containingMod)
+                result += BoundElementWithVisibility(boundElement, isPublic)
```
vlad20012, line 102: Let's rename `isPublic` to `isVisible` in the remaining places like this variable and BoundElementWithVisibility.isPublic and ResolvedPath.Item.isPublic

vlad20012 (approved):

dima74:
bors r=vlad20012

dima74:
bors r=vlad20012

PR #6898, opened 2021-02-27
Improve CratesLocalIndexService internals

avrong:
Based on the comments in #6390.

1. Made `CratesLocalIndexState`'s field immutable
2. Clarified `isReady` by a doc comment, not sure what's the better name for it can be


toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -90,6 +90,9 @@ class CratesLocalIndexServiceImpl
     @Volatile
     private var state: CratesLocalIndexState = CratesLocalIndexState()
 
+    /**
+     * [isReady] will be true when index is not updating
+     */
     private val isReady: AtomicBoolean = AtomicBoolean(true)
```
Undin, line 96: I read this code again and name is still quite confusing. Maybe something like `isUpdating` is better here?

PR #6901, opened 2021-02-27
Associate `Cargo.toml.orig` files with TOML file type

### LINKED ISSUE #6884: Feature: Add Cargo.toml.orig to toml file type patterns
Minor feature request: Have `Cargo.toml.orig` added as a pattern in File Types for `TOML file` when installing the plugin in CLion et. al.

Rationale: When uploading crates to crates.io Cargo replaces `Cargo.toml` and creates a `Cargo.toml.orig` as noted in the generated `Cargo.toml` file:

```
# THIS FILE IS AUTOMATICALLY GENERATED BY CARGO
#
# When uploading crates to the registry Cargo will automatically
# "normalize" Cargo.toml files for maximal compatibility
# with all versions of Cargo and also rewrite `path` dependencies
# to registry (e.g., crates.io) dependencies
#
# If you believe there's an error in this file please file an
# issue against the rust-lang/cargo repository. If you're
# editing this file be aware that the upstream Cargo.toml
# will likely look very different (and much more reasonable)
```

As it says, the original `Cargo.toml` is often more readable and so it would seem sensible to support toml syntax highlighting for files matching the `Cargo.toml.orig` pattern when using the Rust plugin.  Currently this has to be added manually.

vlad20012:
Fixes #6884

changelog: Toml syntax highlighting is now available in `Cargo.toml.orig` files

### DIFF
--- a/intellij-toml/src/main/resources/META-INF/plugin.xml
+++ b/intellij-toml/src/main/resources/META-INF/plugin.xml
@@ -24,7 +24,7 @@
                   implementationClass="org.toml.lang.psi.TomlFileType"
                   fieldName="INSTANCE"
                   extensions="toml"
-                  fileNames="Cargo.lock;Gopkg.lock;Pipfile"/>
+                  fileNames="Cargo.lock;Cargo.toml.orig;Gopkg.lock;Pipfile"/>
         <lang.parserDefinition language="TOML" implementationClass="org.toml.lang.parse.TomlParserDefinition"/>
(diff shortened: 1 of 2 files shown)


vlad20012:
Btw we may want to priotitize `Cargo.toml.orig` over `Cargo.toml` when doing navigation

Undin (commented):
Let's add the correponding test in `TomlFileTypeTest`

Undin (approved):
bors r+

Issue #6902, opened 2021-02-27
Support 2021 edition


Undin:
The tracking issue about [2021 edition](https://github.com/rust-lang/rfcs/pull/3085) support

- [ ] introduce enum variant for 2021 edition in the plugin and adjust existing code
- [ ] ...

HTG-YT:
@Undin Rust 2021 Edition is now available in nightly: https://blog.rust-lang.org/2021/07/21/Rust-2021-public-testing.html.

Can probably implement the second checkpoint.

HTG-YT:
Pull Request for Second Checkpoint: #7569

HTG-YT:
The `2021` `edition` value is only available when the `edition2021` feature is enabled for `cargo-features`. I don't see any `cargo-features` autocompletion at the moment, might also need to implement that.

HTG-YT:
> 
> 
> The tracking issue about [2021 edition](https://github.com/rust-lang/rfcs/pull/3085) support
> 
>     * [x]  introduce enum variant for `2021` edition in the plugin and adjust existing code
> 
>     * [ ]  `2021` value completion for `edition` field in Cargo.toml. Should be implemented only when `2021` will be available at least in nightly
> 
>     * [ ]  adjust parsing of `:pat` macro argument specifier to parse [`or_patterns`](https://blog.rust-lang.org/2021/05/11/edition-2021.html#or-patterns-in-macro_rules)
> 
>     * [ ]  ...

@Undin can now check the second task!

HTG-YT:
Rust 2021 Edition is currently planned to land at Rust version **1.56.0**.

The guide for Rust 2021 Edition can be found at: https://doc.rust-lang.org/nightly/edition-guide/rust-2021/

HTG-YT:
The new added preludes are in `../prelude/mod.rs` instead of `../prelude/v1.rs`.

https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/NameResolution.kt#L1713

It is needed to tweak the name resolution to support the added preludes.

HTG-YT:
Update: A issue for the 2021 edition stabilization is now created.
https://github.com/rust-lang/rust/issues/87959

EwoutH:
What's the latest status of this effort?

vlad20012:
@EwoutH Which particular feature are you interested in?

EwoutH:
[...]

PR #6903, opened 2021-02-27
Introduce EDITION_2021

Undin:
Initial preparation for 2021 edition. It should only support parsing `2021` edition from cargo metadata. All user-visible features should be the same as for `2018` edition

Part of #6902 

changelog: Introduce `2021` edition variant in the plugin and adjust the corresponding code


vlad20012 (approved):

Undin:
bors r=vlad20012

Issue #5422, opened 2020-05-22
Support debugging in PyCharm, GoLand, etc


Undin:
After #3453 the plugin support debugging in:
* CLion
* AppCode
* IntelliJ IDEA Ultimate (with [Native Debug](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin)
* RubyMine

It would be great to support debugging capabilities in other paid IDEs like PyCharm Professional, GoLand, etc

cata0309:
I support this idea ! The Pycharm Edu would make such a great IDE for Rust as it is free and also less bloated!

Undin:
@cata0309 Note, this issue only about **paid** Jetbrains IDEs

cata0309:
Ok, then for Pycharm Professional, I own the entire suite so that's not a problem !

ZOXEXIVO:
I would like to see debugging in Rider

jcbritobr:
Debug in community. Make it available full to all ides. Or make a proprietary full and I'll buy. But half feature ide is not good.

DevDevious:
+1 for other paid IDEs like GoLand

cata0309:
> @cata0309 Note, this issue is only about **paid** Jetbrains IDEs

This is a MIT license, it doesn't say that this is only for paid IDEs. Just because you paid for them doesn't mean others cannot benefit from this plugin too.

jcbritobr:
Vscode now can debug rust code very well, and rust analyzer is as good as intellij rust. I will be good intellij rust got lldb as vscode does. It's alredy implemented.

miranquil:
Please, implement Goland's debug soon!

Undin:
FYI debugging support is available in GoLand since the latest EAP (211.5787.16 version). Support is the same as in IntelliJ IDEA Ultimate: only LLDB and requires [Native Debug](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin

Undin:
And now the debugging is available in the latest PyCharm Pro EAP as well. Only LLDB and requires [Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin

harutyundr:
Is there any reason Debugging is not supported in other Pro IDEs? I own PhpStorm license and would like to develop my CLI tools using Rust.

Undin:
FYI debugging support is available in Rider in 2022.1 builds

[...]

PR #6907, opened 2021-03-01
DBG: debugger support for Rider for Unreal Engine 2021.1

Undin:
Current restrictions:
- the support works starting 2021.1 EAPs
- only [Rider for Unreal Engine](https://www.jetbrains.com/lp/rider-unreal/) is supported for now
- only LLDB

Part of #5422

changelog: Provide debugging in [Rider for Unreal Engine](https://www.jetbrains.com/lp/rider-unreal/) 2021.1

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunnerUtils.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunnerUtils.kt
@@ -14,10 +14,9 @@ import com.intellij.openapi.ui.DialogWrapper
 import com.intellij.openapi.ui.Messages
 import com.intellij.openapi.util.SystemInfo
 import com.intellij.xdebugger.XDebugProcess
+import com.intellij.xdebugger.XDebugProcessStarter
 import com.intellij.xdebugger.XDebugSession
 import com.intellij.xdebugger.XDebuggerManager
-import com.intellij.xdebugger.impl.XDebugProcessConfiguratorStarter
-import com.intellij.xdebugger.impl.ui.XDebugSessionData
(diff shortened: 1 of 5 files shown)


Undin:
Important places for testing:
- Rider 4UE uses Native Debugging Support plugin to provide native debugging capabilities. The plugin is bundled and already contains debugger binaries. So Rust plugin **shouldn't** try to download anything
- Also, since we don't need to download debugger binaries, Rust section in `Preferences | Build, Execution, Deployment | Debugger` settings shouldn't be shown

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugRunnerUtils.kt
```diff
@@ -32,17 +31,15 @@ object RsDebugRunnerUtils {
         state: CargoRunStateBase,
         environment: ExecutionEnvironment,
         runExecutable: GeneralCommandLine
-    ): RunContentDescriptor? {
+    ): RunContentDescriptor {
         val runParameters = RsDebugRunParameters(environment.project, runExecutable, state.cargoProject)
         return XDebuggerManager.getInstance(environment.project)
-            .startSession(environment, object : XDebugProcessConfiguratorStarter() {
```
ortem, line 38: > For some reason, Rider debugging integration fails when the platform calls `XDebugProcessConfiguratorStarter#configure`

Sounds strange to me, I hope `XDebugProcessStarter` really solves this and there are no other unexpected problems

ortem (approved):
LGTM. Works fine on my machine, without extra downloads and settings

Undin:
bors r=ortem

PR #6912, opened 2021-03-02
ANN: Enable by default "Simplify boolean expression" inspection

dima74:
There are `RsSimplifyBooleanExpressionInspection` and `SimplifyBooleanExpressionIntention`. Inspection is offered only for pure expressions, and intention is always offered. Inspection was disabled in e0962ec008e9ed9f8031ecd3d12ca57be1b20fbb because of bug which however was fixed in #1001

changelog: Enable by default "Simplify boolean expression" inspection


ortem (approved):
LGTM, at least this inspection does not produce any exceptions or false positive annotations.

bors r+

Issue #6927, opened 2021-03-06
Keyword in lifetime name is falsely allowed


DomWilliams0:
## Environment

* **IntelliJ Rust plugin version:** 0.3.142.3705-203
* **Rust toolchain version:** rustc 1.52.0-nightly (3f5aee2d5 2021-02-12) 
* **IDE name and version:** CLion 2020.3.2
* **Operating system:** Linux

## Problem description

The plugin shows no errors or warnings when using a keyword as a lifetime name, but the code doesn't compile.

## Steps to reproduce

Code:
```rust
pub struct AddressSpaceRef<'p: 'ref, 'ref>(AddressSpace<'p>, PhantomData<&'ref ()>);
```

Compiler error:
```
error: lifetimes cannot use keyword names
  --> kernel/src/memory/address_space.rs:18:32
   |
18 | pub struct AddressSpaceRef<'p: 'ref, 'ref>(AddressSpace<'p>, PhantomData<&'ref ()>);
   |                                ^^^^
```

PR #6929, opened 2021-03-07
ANN: Provide lifetime name check

Stzx:
changelog:

Fixed: #6927


mchernyavsky (approved):
Thanks!

bors r+

Undin (commented):
Maybe it's worth using some code from `org.rust.ide.refactoring.RsNamesValidator`? For example, use `isKeyword` instead of hardcoding keyword names

PR #6934, opened 2021-03-08
TYPE&FMT: auto-indent line when typing <enter> in macro bodies

vlad20012:
Previously there was absolutely no auto-indentation in macro (call&def) bodies. Now there is a simple syntax tree-based auto-indentation (based on a token tree `{}`/`()`/`[]` nesting).

changelog: auto-indent line when typing `<enter>` in macro bodies

### DIFF
--- a/src/main/kotlin/org/rust/ide/formatter/blocks/RsMacroArgFmtBlock.kt
+++ b/src/main/kotlin/org/rust/ide/formatter/blocks/RsMacroArgFmtBlock.kt
@@ -8,7 +8,13 @@ package org.rust.ide.formatter.blocks
 import com.intellij.formatting.*
 import com.intellij.lang.ASTNode
 import com.intellij.openapi.util.TextRange
+import com.intellij.psi.formatter.FormatterUtil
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/formatter/blocks/RsMacroArgFmtBlock.kt
```diff
@@ -23,11 +29,51 @@ class RsMacroArgFmtBlock(
...
+                    node = childNode,
+                    alignment = null,
+                    indent = computeIndentForChild(childNode),
+                    wrap = null,
+                    ctx = ctx)
+            }
+    }
+
+    override fun getSpacing(child1: Block?, child2: Block): Spacing? {
+        return Spacing.getReadOnlySpacing()
```
vlad20012, line 49: Btw this code actually prevents formatter from doing its work, hence formatter does not work in macros. But if we'll decide to enable it, we can just replace this line with `return null`. Formatter may be handy in macros e.g. when copy-pasting a code into macros.

PR #6947, opened 2021-03-10
REF: Support import aliases in Move refactoring

### LINKED ISSUE #5952: Move refactor doesn't move rename imports
## Environment

0.3.128.3278-202
* **Rust toolchain version:**
1.45.2
* **IDE name and version:**
CLion 2020.2
Build #CL-202.6397.106, built on July 28, 2020
Licensed to CLion Evaluator
Expiration date: September 6, 2020
Runtime version: 11.0.7+10-b944.20 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 10.15.5
Memory: 1987M
Cores: 12
Registry: run.processes.with.pty=TRUE
Non-Bundled Plugins: org.toml.lang, org.rust.lang
* **Operating system:**
macOS 10.15.5

## Problem description
Move refactors broke compile because it doesn't factor in rename imports.

I have a rename import like:

```rust
use serde_json::Value as Json;
```

And when I move a struct from one to another, it doesn't move the rename import. It correctly imported the right type (`serde_json::Value`) but didn't respect the rename so it broke my build. 

## Steps to reproduce

Create a struct using a rename import and move to another file and it won't move the rename import.

dima74:
Fixes #5952

changelog: Support import aliases in Move refactoring (<kbd>F6</kbd> or `Refactor | Move`)

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -220,9 +220,11 @@ class RsPsiFactory(
     fun tryCreateModDeclItem(modName: String): RsModDeclItem? =
         createFromText("mod $modName;")
 
-    fun createUseItem(text: String, visibility: String = ""): RsUseItem =
-        createFromText("$visibility use $text;")
+    fun createUseItem(text: String, visibility: String = "", alias: String? = null): RsUseItem {
+        val aliasText = if (!alias.isNullOrEmpty()) " as $alias" else ""
+        return createFromText("$visibility use $text$aliasText;")
             ?: error("Failed to create use item from text: `$text`")
+    }
 
     fun createUseSpeck(text: String): RsUseSpeck =
         createFromText("use $text;")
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -220,9 +220,11 @@ class RsPsiFactory(
     fun tryCreateModDeclItem(modName: String): RsModDeclItem? =
         createFromText("mod $modName;")
 
-    fun createUseItem(text: String, visibility: String = ""): RsUseItem =
-        createFromText("$visibility use $text;")
+    fun createUseItem(text: String, visibility: String = "", alias: String? = null): RsUseItem {
+        val aliasText = if (alias != null) " as $alias" else ""
```
ortem, line 224: [minor]
```suggestion
        val aliasText = if (!alias.isNullOrEmpty()) " as $alias" else ""
```

ortem (approved):
LGTM

dima74:
bors r=ortem

Issue #6943, opened 2021-03-10
Wrong type rendering with const generic parameters


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.144.3720-211-nightly
* **Rust toolchain version:** 1.52.0-nightly (3a5d45f68 2021-03-09) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.1 EAP (CL-211.6305.15)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
`const_generics` feature allows you to write const generic params before type params. But the plugin doesn't take it into account in type rendering and always render type parameters before const parameters. For example, in the example below, I expect `S<3, i32>` type hint but the plugin shows `S<i32, 3>`

UPD:
use #![feature(generic_const_exprs)]
#![feature(const_generics)] - feature has been removed

<img width="459" alt="Screen Shot 2021-03-10 at 13 43 31" src="https://user-images.githubusercontent.com/2539310/110617783-1eba9d00-81a7-11eb-8490-9fadbb1c2844.png">

There is the same issue in quick doc where types are present and in some error annotations

![image](https://user-images.githubusercontent.com/2539310/110618749-1878f080-81a8-11eb-9857-27000c4ec189.png)

## Steps to reproduce
```rust
#![feature(const_generics)]

fn main() {
    let a = [1, 2, 3];
    let s = S(&a);
    foo(s);
}

fn foo(x: i32) {}

struct S<'a, const N: usize, T>(&'a [T; N]);

```

Undin:
Inlay hints rendering still works incorrectly

PR #6948, opened 2021-03-10
Fix type rendering with const generic parameters

mchernyavsky:
Partially fixes https://github.com/intellij-rust/intellij-rust/issues/6943 (except inlay type hints).

changelog: Fix type rendering with const generic parameters

### DIFF
--- a/src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
@@ -7,9 +7,13 @@ package org.rust.ide.presentation
 
 import org.rust.ide.utils.import.ImportCandidatesCollector.findImportCandidate
 import org.rust.ide.utils.import.ImportContext
+import org.rust.lang.core.psi.RsConstParameter
+import org.rust.lang.core.psi.RsLifetimeParameter
 import org.rust.lang.core.psi.RsTraitItem
+import org.rust.lang.core.psi.RsTypeParameter
 import org.rust.lang.core.psi.ext.*
 import org.rust.lang.core.types.BoundElement
+import org.rust.lang.core.types.Substitution
 import org.rust.lang.core.types.consts.Const
 import org.rust.lang.core.types.consts.CtConstParameter
 import org.rust.lang.core.types.consts.CtUnknown
@@ -164,7 +168,7 @@ private data class TypeRenderer(
                     append(formatBoundElement(ty.aliasedBy, render))
                 } else {
                     append(getName(ty.item) ?: return anonymous)
-                    if (includeTypeArguments) append(formatGenerics(ty, render))
+                    if (includeTypeArguments) append(formatAdtGenerics(ty, render))
                 }
             }
             is TyInfer -> when (ty) {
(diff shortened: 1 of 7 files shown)


Undin:
@mchernyavsky tests fail

mchernyavsky:
@Undin Fixed.

src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
```diff
@@ -266,15 +257,30 @@ private data class TypeRenderer(
...
-        }
-        val constSubst = boundElement.element.constParameters.map { render(boundElement.subst[it] ?: CtUnknown) }
-        return regionSubst + tySubst + constSubst
+        var nonDefaultTyFound = false
+        return declaration.genericParameters
+            .asReversed()
+            .filter { parameter ->
+                if (!skipUnchangedDefaultTypeArguments || parameter !is RsTypeParameter) return@filter true
+                nonDefaultTyFound = nonDefaultTyFound || parameter.typeReference?.type != subst[parameter]
+                nonDefaultTyFound
```
Undin, line 273: Unfortunately, it doesn't work as expected.
For example, for the following code (yep, I know, that it's not allowed yet but anyway):
```rust
#![feature(const_generics)]

struct S<T=u32, const B: usize>([T; B]);

fn main() {
    let s = S([1u32]);
    s;
}
```

if you try to extract `s` into function via the corresponding refactoring, you get
```rust
#![feature(const_generics)]

struct S<T=u32, const B: usize>([T; B]);

fn main() {
    let s = S([1u32]);
    foo(s);
}

fn foo(s: S<1>) -> S<1> {
    s
}
```
but I expect
```rust
#![feature(const_generics)]

struct S<T=u32, const B: usize>([T; B]);

fn main() {
    let s = S([1u32]);
    foo(s);
}

fn foo(s: S<u32, 1>) -> S<u32, 1> {
    s
}
```

You need always change `nonDefaultTyFound` (BYW not the best name since we can find const parameter) to `true` if `parameter` is not `RsTypeParameter`

src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
```diff
@@ -266,15 +257,30 @@ private data class TypeRenderer(
...
-            emptyList()
-        }
-        val constSubst = boundElement.element.constParameters.map { render(boundElement.subst[it] ?: CtUnknown) }
-        return regionSubst + tySubst + constSubst
+        var nonDefaultTyFound = false
+        return declaration.genericParameters
+            .asReversed()
+            .filter { parameter ->
+                if (!skipUnchangedDefaultTypeArguments || parameter !is RsTypeParameter) return@filter true
+                nonDefaultTyFound = nonDefaultTyFound || parameter.typeReference?.type != subst[parameter]
```
Undin, line 272: Is it possible to get `null` from `subst[parameter]` and `null` from `parameter.typeReference?.type`?
Maybe it's worth explicitly check that `parameter.typeReference != null`?

Undin (changes requested):
Note, it still doesn't fix type hints because they are rendered by `org.rust.ide.hints.type.RsTypeHintsPresentationFactory#adtTypeHint`. Don't insist to fix it in this PR

src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
Undin: Oh, it's even checked in your tests but the corresponding test is wrong

src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
```diff
@@ -266,15 +257,30 @@ private data class TypeRenderer(
...
-        val tySubst = boundElement.element.typeParameters.map { render(boundElement.subst[it] ?: TyUnknown) }
-        val regionSubst = if (includeLifetimeArguments) {
-            boundElement.element.lifetimeParameters.map { render(boundElement.subst[it] ?: ReUnknown) }
-        } else {
-            emptyList()
-        }
-        val constSubst = boundElement.element.constParameters.map { render(boundElement.subst[it] ?: CtUnknown) }
-        return regionSubst + tySubst + constSubst
+        var nonDefaultTyFound = false
+        return declaration.genericParameters
```
Undin, line 268: nit: I personally prefer not to use `filter`, `map`, etc. methods when they have some state (`nonDefaultTyFound` here) and use simple `for` loop instead. Also, it allows not to create a separate list object by `map`. But it's up to you

mchernyavsky:
@Undin fixed.

Undin (approved):
Great!

mchernyavsky (commented):
bors r=Undin

PR #6957, opened 2021-03-13
REF: Create file in Move refactoring if it doesn't exist

### LINKED ISSUE #5576: "Target file must be Rust file" moving objects to a new file
## Environment

* **IntelliJ Rust plugin version:** 0.2.125.SNAPSHOT-202-dev
* **Rust toolchain version:** 1.44.0-nightly (b2e36e6c2 2020-04-22) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.2 EAP (CL-202.5428.21)
* **Operating system:** macOS 10.15.4

## Problem description

I am testing #3531 and the plugin complains

> Target file must be Rust file

... when I try to move a function to a non-existed file. I believe in other languages (e.g. Python) IDE simply creates a new file. If this is not easy to do we can improve the error message I think.

## Steps to reproduce

- Create a new binary create
- Define `fn foo() {}` in `main.rs`
- Try to move `foo` to `main2.rs` (don't create the file in advance)

Screen Shot 2020-06-15 at 19 03 41

### LINKED ISSUE #6584: Move refactor should create file if it doesn't exist
Now if you try to move to non-existent file refactoring gives an error, while it could create it for me.
I think it is pretty common use case (actually, most of the time I use it to extract parts of module to separate one).

## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3572-203
* **Rust toolchain version:** 1.50.0-nightly (fa4163942 2020-12-14) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.3 (CL-203.5981.166)
* **Operating system:** Linux 5.4.0-58-generic

dima74:
Fixes #5576
Fixes #6584

changelog: Create target file when moving items (<kbd>F6</kbd> or `Refactor | Move`) if it doesn't exist


src/main/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsDialog.kt
```diff
@@ -196,3 +238,59 @@ class RsMoveItemAndImplsInfo(
...
+            val file = virtualFile.toPsiFile(project)?.rustFile ?: return@createNewFile null
+            if (!attachFileToParentMod(file, project, crateRoot)) return@createNewFile null
+            file
+        }
+    }
+}
+
+/** Finds parent mod of [file] and adds mod declaration to it */
+private fun attachFileToParentMod(file: RsFile, project: Project, crateRoot: RsMod?): Boolean {
+    val (parentModOwningDirectory, modName) = if (file.name == RsConstants.MOD_RS_FILE) {
```
Kobzol, line 257: Super drive-by comment: does this work for both 2015 and 2018 editions?

dima74: Looks like the only relevant edition difference is that in 2018 edition [module with submodules doesn't have to be named `foo/mod.rs`](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs), so yes, it should work for both editions

Btw Move refactoring itself currently supports only 2018 edition

ortem:
Is it expected that file creation cannot be undone? Consider two cases:

**Cancelation**
1. Invoke `Move`, select a non-existent file, and press `Refactor`
2. `Problems detected`, press `Cancel`
3. Press `Cancel` in the main `Move` dialog
4. File is created and attached to the original mod

**Undo**
1. Invoke `Move`, select a non-existent file, and press `Refactor`
2. `Problems detected`, press `Continue`
4. Move is done, press `Ctrl`+`Z`
5. File is still there and is still attached to the original mod

BTW it seems like Kotlin plugin and Pycharm are unable to undo move to a non-existent file, aren't they? I mean, there is a dialog `Cannot undo. Unexpected document state` when I try to undo such a refactoring. So our current behaviour might be fine considering it is much better than nothing. @dima74 WDYT?

dima74:
Yes, it is expected - we have to create target file before checking conflicts. We may manually delete created file if refactoring was cancelled, but I would leave it for future PR

Regarding undo in Kotlin - it works a bit strange, I indeed can't undo move if cursor is inside source file, but can if cursor is inside target file

ortem (approved):
OK, I'll create a separate issue about the undoing. LGTM.
bors r+

Issue #6962, opened 2021-03-14
Toolchain is not automatically detected for /opt/homebrew rust installation


auduchinok:
Homebrew 3.0 uses `/opt/homebrew` as the default path on Apple Silicon, and `rust` binaries installed via `brew install rust` are located in the corresponding path:
```
eugene@Eugenes-Air:hello_cargo$ which cargo
/opt/homebrew/bin/cargo
eugene@Eugenes-Air:hello_cargo$ which rustc
/opt/homebrew/bin/rustc
```

This toolchain is not picked up automatically:
<img width="1094" alt="Screenshot 2021-03-14 at 21 04 44" src="https://user-images.githubusercontent.com/3923587/111079014-28c2ff80-8509-11eb-9b02-c2f50166186e.png">

CLion 2021.1
Build #CL-211.6432.9, built on March 10, 2021
Runtime version: 11.0.10+9-b1341.19 aarch64
Non-Bundled Plugins: org.toml.lang (0.2.142.3705-211), org.rust.lang (0.3.142.3705-211)

PR #6965, opened 2021-03-14
Fix automatic detection of brew installed Rust toolchain

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6962.

changelog: Automatically detect Rust toolchain installed via Homebrew 3

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/flavors/RsMacToolchainFlavor.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/flavors/RsMacToolchainFlavor.kt
@@ -11,7 +11,7 @@ import java.nio.file.Path
 
 object RsMacToolchainFlavor : RsToolchainFlavor() {
     override fun getHomePathCandidates(): List<Path> {
-        val path = "/usr/local/Cellar/rust/bin".toPath()
+        val path = "/opt/homebrew/bin".toPath()
         return if (path.isDirectory()) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/toolchain/flavors/RsMacToolchainFlavor.kt
```diff
@@ -11,7 +11,7 @@ import java.nio.file.Path
 
 object RsMacToolchainFlavor : RsToolchainFlavor() {
     override fun getHomePathCandidates(): List<Path> {
-        val path = "/usr/local/Cellar/rust/bin".toPath()
```
ortem, line 14: Shouldn't we check the old path as well to maintain the compatibility?

mchernyavsky: Homebrew <3.0 adds symlinks for installed binaries to `/usr/local/bin`, which is already covered by `RsUnixToolchainFlavor`.

auduchinok: Homebrew 3.0 `arm` and `x86-64` builds can be installed side-to-side on an M1 machine. Will `x86-64` build also add the symlinks?

Issue #6964, opened 2021-03-14
Postfix templates are suggested inside string literals


auduchinok:
<img width="622" alt="Screenshot 2021-03-14 at 21 43 47" src="https://user-images.githubusercontent.com/3923587/111080318-ffa56d80-850e-11eb-886a-692e661c146d.png">

CLion 2021.1
Build #CL-211.6432.9, built on March 10, 2021
Runtime version: 11.0.10+9-b1341.19 aarch64
Non-Bundled Plugins: org.toml.lang (0.2.142.3705-211), org.rust.lang (0.3.142.3705-211)

PR #6968, opened 2021-03-14
COMP: Don't suggest postfix templates inside string literals

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/6964.

changelog: Don't suggest postfix templates inside string literals

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/utils.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/utils.kt
@@ -5,17 +5,16 @@
 
 package org.rust.ide.template.postfix
 
-import com.intellij.codeInsight.template.postfix.templates.PostfixTemplateExpressionSelector
+import com.intellij.codeInsight.template.postfix.templates.PostfixTemplateExpressionSelectorBase
 import com.intellij.codeInsight.template.postfix.templates.PostfixTemplatePsiInfo
 import com.intellij.openapi.editor.Document
+import com.intellij.openapi.util.Condition
 import com.intellij.openapiext.isUnitTestMode
 import com.intellij.psi.PsiElement
-import com.intellij.util.Function
 import org.rust.lang.core.psi.RsBlock
 import org.rust.lang.core.psi.RsExpr
 import org.rust.lang.core.psi.RsPsiFactory
 import org.rust.lang.core.psi.ext.ancestors
-import org.rust.lang.core.psi.ext.endOffset
 import org.rust.lang.core.types.ty.TyBool
 import org.rust.lang.core.types.type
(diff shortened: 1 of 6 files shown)


Undin (commented):
Maybe it's worth just use `PostfixTemplateExpressionSelectorBase` as base class for `RsExprParentsSelectorBase`?
It already has the corresponding filtering by offset (i.e. you don't need to check it yourself) and moreover provides a slightly simplified API that can help not to duplicate logic in `hasExpression` and `getExpressionsInternal`

Note, the current implementation fixes the bug only in string literals but the bug can be reproduced in float literals as well

src/main/kotlin/org/rust/ide/template/postfix/utils.kt
```diff
@@ -28,48 +28,32 @@ internal object RsPostfixTemplatePsiInfo : PostfixTemplatePsiInfo() {
...
-
-    protected abstract fun getExpressionsInternal(context: PsiElement, document: Document, offset: Int): List<PsiElement>
 }
 
 class RsTopMostInScopeSelector(pred: ((RsExpr) -> Boolean) = { true }) : RsExprParentsSelectorBase(pred) {
-    override fun getExpressionsInternal(context: PsiElement, document: Document, offset: Int): List<PsiElement> =
+    override fun getNonFilteredExpressions(context: PsiElement, document: Document, offset: Int): List<PsiElement> =
         context
             .ancestors
             .takeWhile { it !is RsBlock && it.endOffset == context.endOffset }
```
mchernyavsky, line 45: I'm not sure if `it.endOffset == context.endOffset` condition is still needed. All tests in `org.rust.ide.template.postfix` package pass without it.

mchernyavsky:
@Undin fixed.

src/main/kotlin/org/rust/ide/template/postfix/utils.kt
Undin: Oh, it's interesting

Currently, some of our postfix templates behave in the following way:


And it's rather weird in my opinion that the plugin tries to select all parent expressions in a block even if we are inside this expression. The same postfix templates work differently in other languages.

But after your changes, they become to works in a similar way as in other languages. At the same time, `RsTopMostInScopeSelector` and `RsAllParentsSelector` provide the same behaviour now because `PostfixTemplateExpressionSelectorBase` always use `element.getTextRange().getEndOffset() == offset` condition so I suppose we can merge them into single `RsExprParentsSelector`

mchernyavsky: @Undin fixed.

Undin (approved):
bors r+

PR #6969, opened 2021-03-15
COMP: Improve dependency completion when using crates local index

avrong:
changelog: Make dependencies completion in `Cargo.toml` more intuitive and faster when using crates local index. Note, this feature is disabled by default for now. To use it, enable `org.rust.crates.local.index` experimental feature


ortem (approved):
LGTM
bors r+

PR #6970, opened 2021-03-15
Revert "Slightly optimize RsFile.super"

dima74:
Reverts #6861

For some reason it causes up to 2-3x slow down of building CrateDefMaps. Lets revert it for current release

bors r+


dima74:
Will merge fix instead: #6971

PR #6971, opened 2021-03-15
RES: Fix updating file maps in new resolve

dima74:
I incorrectly used method `.remove` while it was expected to use `.removeAll`. It leads to some sort of memory leak after changing type of `fileIdToCrateId` from `Map` to `MultiMap` in #6861


Undin (approved):
bors r+

PR #6972, opened 2021-03-15
T: Fix pretty printer tests on Python 3.8

dima74:
Method `time.clock()` was removed in Python 3.8


Undin (commented):
Let's better make the same fix as in [rust-lang repo](https://github.com/rust-lang/rust/commit/a9484d464741689383a93d7bc2f85c26416fcd57) to support python 2

Undin:
bors try

Undin (approved):
bors r+

Issue #6978, opened 2021-03-16
Ask user to provide function return type when creating a new function if it's result is assigned to a variable


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.144.3732-211-nightly
* **Rust toolchain version:** 1.52.0-nightly (b3e19a221 2021-03-12) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.2 EAP (CL-212.1007)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

If the variable used to store the function return result has no explicit type annotation "create function" intention action doesn't suggest filling the return type. This is a bit inconvenient as I have to go back and declare the return type once I finished filling the live template.

I guess in this case it is safe to assume the function will return something and the user have to specify the type.

## Steps to reproduce

Consider

```rust
fn main() {
    let foo = bar(42i8);
}
```

... invoke "create function" on `bar`

```rust
fn bar(p0: i8) {
    unimplemented!()
}
```

Would prefer

<img width="687" alt="create" src="https://user-images.githubusercontent.com/2500320/111373523-08cc4100-86ad-11eb-9b98-eba4f44aa498.png">

PR #6986, opened 2021-03-17
INT: use template for unknown return type in CreateFunctionIntention

Kobzol:
This PR changes how `CreateFunctionIntention` treats function call expressions with unknown expected return types. Previously, it collapsed them to `TyUnit`, so the created function didn't have any return type. This was problematic in scenarios like this:
```rust
let foo = bar(42i8); // missing return type
```
as noted by @lancelote (https://github.com/intellij-rust/intellij-rust/issues/6978).

After inspecting the issue, I realized that this can happen in more cases:
```rust
let foo: UnknownType = bar(); // let decl with unknown type

fn baz(a: UnknownType) {}
baz(bar());  // argument with unknown type
```

In all these cases, the user will most likely want to enter the return type right away. Therefore I created a general heuristic that tries to guess if the user will want to enter the return type or not. Currently, the heuristic works like this:
1) If the return type is known, do not use any template for entering return type and just render it
2) If the return type is unknown:
    a) If the function call is a statement, treat return type as unit, do not display template.
    b) If the function call is an `await` statement (`foo().await;`), treat return type as unit, do not display template.
    c) Otherwise, let the user choose the return type with a template.

Basically it wants to let the user choose the return type if it is unknown, with the exception of "base" calls (and "bare" await calls), which probably are not supposed to have a return type:
```rust
foo();
// v
fn foo() {}

foo().await;
// v
fn foo() {}

let x = foo();
// v
fn foo() -> _/*caret*/ {}

fn bar(a: UnknownType) {}
bar(foo());
// v
fn foo() -> _/*caret*/ {}
```

Fixes: https://github.com/intellij-rust/intellij-rust/issues/6978

changelog: Allow explicitly stating the function return type in CreateFunctionIntention if the inferred return type was unknown.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateFunctionIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/createFromUsage/CreateFunctionIntention.kt
@@ -36,11 +36,31 @@ class CreateFunctionIntention : RsElementBaseIntentionAction<CreateFunctionInten
         }
     }
 
+    class ReturnType(val type: Ty, val needsTemplate: Boolean) {
+        companion object {
+            fun create(expr: RsExpr): ReturnType {
+                val expected = expr.expectedType ?: TyUnknown
+
+                // Show template if the type is unknown and there is something that expects the return value
+                val parent = expr.parent
+                val needsTemplate = expected is TyUnknown && when (parent) {
+                    is RsExprStmt -> false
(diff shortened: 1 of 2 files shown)


vlad20012 (approved):
:+1: 
bors r+

Kobzol:
bors retry

Kobzol:
bors retry

Kobzol:
bors retry

Issue #4869, opened 2020-01-23
Suggestion - quick action to expand dependency specification in Cargo.toml


dmarcuse:
When adding dependencies to my `Cargo.toml`, it's handy that the latest version is provided as a completion. However, it's still a bit annoying to type out the dependency specifications when the dependency needs to be configured with additional options, like `default_features = false` or `features = [ "serde" ]`. It would be handy to have an action that expands the shorter syntax to the full syntax, placing the cursor before the closing bracket so additional options can be added - replacing `serde = "1.0"` with `serde = { version = "1.0" }`

dmarcuse:
I'd also like to add that while I'm not familiar with developing Jetbrains IDE plugins, I do know Kotlin, so I could take a stab at implementing this and opening a PR if others are interested but nobody else has the bandwidth to work on it.

Undin:
@dmarcuse Hi!
Thanks for the feature request. Interesting that I thought about the same intention recently but did not have time to open an issue. So it would be cool if you implement it - I think it's rather useful.

> I'm not familiar with developing Jetbrains IDE plugins

It's not a big problem.
You can read [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md) that provides a short description of project structure, building, testing and other useful stuff for contributors. If you are interested in our plugin architecture - see [ARCHITECTURE.md](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md) but it looks like only a small part of this article is helpful for this particular feature (for example, paragraph about [PSI](https://github.com/intellij-rust/intellij-rust/blob/master/ARCHITECTURE.md#psi).

[...]

PR #7003, opened 2021-03-22
INT: Introduce extract dependency in a separate table intention

avrong:
Depends on #6960
Part of #4869

Introduces intention to extract dependency specification written as an inline table to a separate table below. 

### TODO
- [x] Fix tests by providing a correct intention test base for TOML

changelog: Add intention to extract dependency specification in a separate table in `Cargo.toml`

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
+++ b/toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
@@ -1,0 +1,64 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiParserFacade
+import com.intellij.psi.util.parentOfType
+import com.intellij.psi.util.parentsOfType
+import org.rust.cargo.CargoConstants
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.toml.isDependencyListHeader
+import org.toml.lang.psi.TomlInlineTable
+import org.toml.lang.psi.TomlKeyValue
+import org.toml.lang.psi.TomlPsiFactory
+import org.toml.lang.psi.TomlTable
+
+class ExtractDependencySpecificationIntention : TomlElementBaseIntentionAction<TomlKeyValue>() {
+    override fun getText() = "Extract dependency specification"
+    override fun getFamilyName(): String = text
+
+    override fun findApplicableContextInternal(project: Project, editor: Editor, element: PsiElement): TomlKeyValue? {
+        if (element.containingFile.name != CargoConstants.MANIFEST_FILE) return null
+
+        val dependencyTable = element.parentOfType<TomlTable>()
+            ?.takeIf { it.header.isDependencyListHeader }
+            ?: return null
+
+        val dependency = element.parentsOfType<TomlKeyValue>()
+            .firstOrNull { it.parent == dependencyTable }
+
+        return dependency?.takeIf { it.value is TomlInlineTable }
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: TomlKeyValue) {
+        val inlineTable = ctx.value as? TomlInlineTable ?: return
+        val table = ctx.parent as? TomlTable ?: return
+        val dependencyTableName = table.header.key?.text.orEmpty()
+        val crateName = ctx.key.text
+        val newTable = TomlPsiFactory(project).createTable("$dependencyTableName.$crateName")
+
+        val psiParserFacade = PsiParserFacade.SERVICE.getInstance(project)
+        for (keyValue in inlineTable.entries) {
+            newTable.add(psiParserFacade.createWhiteSpaceFromText("\n"))
+            newTable.add(keyValue)
+        }
+
+        val newTableOffset = if (table.entries.size == 1) {
+            table.replace(newTable).endOffset
+        } else {
+            ctx.delete()
+            val whitespace = psiParserFacade.createWhiteSpaceFromText("\n\n")
+
+            val addedTable = table.parent.addAfter(newTable, table)
+            table.parent.addAfter(whitespace, table)
+            addedTable.endOffset
+        }
+        editor.caretModel.moveToOffset(newTableOffset)
+    }
+}
--- a/toml/src/main/resources/META-INF/toml-only.xml
+++ b/toml/src/main/resources/META-INF/toml-only.xml
@@ -37,6 +37,11 @@
             <category>Rust/Cargo.toml</category>
         </intentionAction>
 
+        <intentionAction>
+            <className>org.rust.toml.intentions.ExtractDependencySpecificationIntention</className>
+            <category>Rust/Cargo.toml</category>
+        </intentionAction>
+
         <applicationService serviceInterface="org.rust.toml.crates.local.CratesLocalIndexService"
                             serviceImplementation="org.rust.toml.crates.local.CratesLocalIndexServiceImpl"
                             testServiceImplementation="org.rust.toml.crates.local.TestCratesLocalIndexServiceImpl"/>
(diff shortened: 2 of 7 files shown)


toml/src/main/resources/intentionDescriptions/ExtractDependencySpecificationIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention extracts dependency specification to a separate table in Cargo.toml.
```
Undin, line 3: Let's drop `This intention` from the description. There is a [recommendation](https://jetbrains.design/intellij/text/inspections/#descriptions ) for inspection descriptions to omit `This inspection`. I haven't found a similar recommendation for intentions but I suppose we should use the same principle.
And yeah, I know that a lot of intentions (including Rust plugin intentions) use `This intention`.

toml/src/main/resources/META-INF/toml-only.xml
```diff
@@ -32,6 +32,16 @@
                          enabledByDefault="true" level="WARNING"
                          implementationClass="org.rust.toml.inspections.CrateNotFoundInspection"/>
 
+        <intentionAction>
+            <className>org.rust.toml.intentions.ExpandDependencySpecificationIntention</className>
+            <category>Rust</category>
```
Undin, line 37: Let's extract these new intentions into separate subgroup like `Cargo.toml`. Note, I'm note sure about category syntax to create a subgroup but you can check platform inspections

toml/src/main/kotlin/org/rust/toml/intentions/TomlElementBaseIntentionAction.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package org.rust.toml.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.rust.ide.intentions.RsElementBaseIntentionAction
+import org.rust.toml.tomlPluginIsAbiCompatible
+
+abstract class TomlElementBaseIntentionAction<Ctx> : RsElementBaseIntentionAction<Ctx>() {
```
Undin, line 14: It's a little wierd that `RsElementBaseIntentionAction` is base class for `TomlElementBaseIntentionAction`. Probably, it's worth extracting some language independent implementation to `common` module.
Just a thought, not a call to action

toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    override fun invoke(project: Project, editor: Editor, ctx: TomlKeyValue) {
+        val inlineTable = ctx.value as? TomlInlineTable ?: return
+        val table = ctx.parent as? TomlTable ?: return
+        val dependencyTableName = table.header.key?.text.orEmpty()
+        val crateName = ctx.key.text
+        val newTable = TomlPsiFactory(project).createTable("$dependencyTableName.$crateName")
+
+        val rsPsiFactory = RsPsiFactory(project)
```
Undin, line 57: It's suspicious that we use **Rs**PsiFactory to modify toml file

toml/src/main/kotlin/org/rust/toml/intentions/ExpandDependencySpecificationIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.toml.isDependencyListHeader
+import org.toml.lang.psi.TomlKeyValue
+import org.toml.lang.psi.TomlLiteral
+import org.toml.lang.psi.TomlPsiFactory
+import org.toml.lang.psi.TomlTable
+import org.toml.lang.psi.ext.TomlLiteralKind
+import org.toml.lang.psi.ext.kind
+
+class ExpandDependencySpecificationIntention : TomlElementBaseIntentionAction<TomlKeyValue>() {
```
Undin, line 22: Consider creating an opposite intention, i.e. to convert dependency table only with version to key-value form.
Of course, in separate PR

Undin:
Oh, I didn't notice that this PR includes changes from #6960. Most of the comments are valid for #6960 as well

toml/src/main/resources/intentionDescriptions/ExtractDependencySpecificationIntention/after.rs.template
```diff
@@ -0,0 +1,3 @@
+[dependencies.foo]
```
ortem, line 1: That's not true because the intention leaves an empty `[dependencies]` table right now. WDYT about actually removing them?

toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.rust.cargo.CargoConstants
+import org.rust.lang.core.psi.ext.endOffset
+import org.rust.toml.isDependencyListHeader
+import org.toml.lang.psi.TomlInlineTable
+import org.toml.lang.psi.TomlKeyValue
+import org.toml.lang.psi.TomlPsiFactory
+import org.toml.lang.psi.TomlTable
+
+class ExtractDependencySpecificationIntention : TomlElementBaseIntentionAction<TomlKeyValue>() {
+    override fun getText() = "Extract dependency in separate table"
```
ortem, line 22: This differs from `"Extracts dependency specification to a separate table in Cargo.toml"` (`description.html`)

WDYT about `"Extract dependency specification in Cargo.toml"` for consistency with `ExpandDependencySpecificationIntention`?

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
avrong: Then it is better to be `Extract dependency specification`, similar to `ExpandDependencySpecificationIntention` and to not repeat that it is `Cargo.toml` intention in `Cargo.toml` file.

toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        val dependencyTable = element.parentOfType<TomlTable>()
+            ?.takeIf { it.header.isDependencyListHeader }
+            ?: return null
+
+        val keyValue = element.parentsOfType<TomlKeyValue>()
+            .firstOrNull { it.parent == dependencyTable }
+            ?.takeIf { it.value is TomlInlineTable }
+
+        return keyValue
```
ortem, line 37: `Variable used only in following return and can be inlined`

Also, the name `keyValue` is not very descriptive. WDYT about rewriting it in this way, to indicate that we are trying to find the corresponding `dependency` and then use it only if it contains a table?
```
val dependency = element.parentsOfType<TomlKeyValue>()
    .firstOrNull { it.parent == dependencyTable }

return dependency?.takeIf { it.value is TomlInlineTable }
```

ortem (changes requested):

toml/src/main/kotlin/org/rust/toml/intentions/ExtractDependencySpecificationIntention.kt
```diff
@@ -0,0 +1,64 @@
...
+
+        val newTableOffset = if (table.entries.size == 1) {
+            table.replace(newTable).endOffset
+        } else {
+            ctx.delete()
+            val whitespace = psiParserFacade.createWhiteSpaceFromText("\n\n")
+
+            val offset = table.parent.addAfter(newTable, table).endOffset
+            table.parent.addAfter(whitespace, table)
+            offset + whitespace.textLength
```
avrong, line 60: Not sure that this is the best way to calculate offset for cursor

ortem: Consider this approach:
```Kotlin
val addedTable = table.parent.addAfter(newTable, table)
table.parent.addAfter(whitespace, table)
addedTable.endOffset
```

Also, I see only one test covers this piece of code. It would great to cover more cases, for example, with multiple tables 
```Toml
[dependencies]
foo = { version = "0.0.1" }
bar = { version = "0.0.2" }<caret>
baz = { version = "0.0.3" } 

[dependencies.quux]
version = "0.0.4"
```

ortem (changes requested):
https://github.com/intellij-rust/intellij-rust/pull/7003#discussion_r638704983

ortem (approved):
bors r+

PR #7012, opened 2021-03-24
REF: don't import default type parameters by `Extract Function` refactoring

Undin:
changelog: Don't import default type parameters by `Extract Function` refactoring (`Refactor | Extract | Method` or <kbd>Cmd</kbd> + <kbd>Alt</kbd> + <kbd>M</kbd>)


mchernyavsky (approved):
bors r+

PR #7019, opened 2021-03-25
GRAM: improve parser recovery in function arguments

vlad20012:
Better parser recovery on missed function arguments. Works in cases like:

```
foo(1, , 2);
foo(,, 2);
foo(1,,);
```

changelog: Improve parser recovery in function arguments


vlad20012:
bors try

PR #7031, opened 2021-03-28
Fix crates local index state updates

avrong:
Bug introduced by changes in #6898. Using `val` for a state field doesn't make it possible to serialize and save it, so it was being ignored.

changelog: Fix unnecessary reload of crates local index after IDE restart.


toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -267,7 +267,10 @@ class CratesLocalIndexServiceImpl
     }
 
     companion object {
-        data class CratesLocalIndexState(val indexedCommitHash: String = "")
+        data class CratesLocalIndexState(
+            // Should be mutable to enable field serialization and saving
```
ortem, line 271: I'd prefer using a doc comment `/** ... */` above `CratesLocalIndexState` declaration

Undin (approved):
In general, looks ok
Consider using modern API: `SimplePersistentStateComponent` and `BaseState`. I suppose it may slightly simplify implementation by dropping some boilerplate code

bors r+

PR #7035, opened 2021-03-29
Fix fetching registry index branch on nightly

avrong:
Uses `origin/HEAD` instead of `origin/master` to fetch the branch of registry index.

changelog: Fix loading crates local index on nightly rust

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
@@ -64,9 +64,15 @@ class CratesLocalIndexServiceImpl
         .build()
 
     private val registryHeadCommitHash: String
-        get() = repository.resolve("origin/master")?.name ?: run {
-            LOG.error("Git revision `origin/master` cannot be resolved to any object id")
-            INVALID_COMMIT_HASH
+        get() {
+            // BACKCOMPAT: Rust 1.49
+            // Since 1.50 there should always be CARGO_REGISTRY_INDEX_TAG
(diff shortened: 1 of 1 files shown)


toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -64,8 +64,8 @@ class CratesLocalIndexServiceImpl
         .build()
 
     private val registryHeadCommitHash: String
-        get() = repository.resolve("origin/master")?.name ?: run {
-            LOG.error("Git revision `origin/master` cannot be resolved to any object id")
+        get() = repository.resolve("origin/HEAD")?.name ?: run {
```
Undin, line 67: Let's extract `origin/HEAD` into constant

Undin (approved):
Works for me!
Thank you for the quick investigation and fix 👍

Undin (commented):
Looks like tests fail for rust 1.41

avrong:
Cargo in toolchains uses such tags (remotes) for the registry index:
- 1.49 and older — `origin/master`;
- 1.50–1.52 — `origin/master` and `origin/HEAD`;
- 1.53 and future — `origin/HEAD`.

There are two possible solutions I see:
1. Use the toolchain to determine the Rust version and then, decide what tag should be used. But for that, we need to somehow get the project instance. I don't know if it is possible in an application-level service.
  As a solution, the service and state can be separated into project service and application service. It will also be a solution to mapping crates local index update to the cargo project updates.
2. A simpler solution is to just check both tags: if `origin/HEAD` is not found, then try `origin/master`.

Undin:
I like the second option: it's very simple and quite efficient. Also, we usually prefer not to use concrete versions (if it's possible, of course) to support intermediate states like nightly builds

Undin (approved):
bors r+

PR #7046, opened 2021-04-02
Implement stub-only `QueryAttributes` API

vlad20012:
Internal. Now `QueryAttributes` can be used without PSI


dima74 (approved):

vlad20012:
bors r=dima74

PR #7053, opened 2021-04-05
Do not evaluate unknown cfg to false in stdlib crates

Undin:
When `org.rust.cargo.fetch.actual.stdlib.metadata` feature is enabled, we know that stdlib crates have build scripts.

Previously, we tried to evaluate unknown cfg options that led to unresolved references in some cases.
For example, `panic!` macro has two versions with `cfg(bootstrap)` and `cfg(not(bootstrap))` on 1.51.0 and in combination with the current name resolution engine, it cannot be resolved.

Now the plugin doesn't evaluate unknown options (at least temporarily) in stdlib crates.

changelog: Fix name resolution of items in standard library with `cfg` attribute in some cases


Undin:
bors retry

PR #7057, opened 2021-04-05
RES: Filter out cfg-disabled items during resolve and completion in new resolve

dima74:
Filters out cfg-disabled items in completion and ignores them in resolve. Note that it currently doesn't apply to out-of-scope items added to completion using auto-import

changelog: Exclude some cfg-disabled items from completion when using new name resolution engine


dima74:
bors retry

PR #7063, opened 2021-04-07
INSP: improve `Can't find crate` annotations

Undin:
Now the corresponding annotation is provided by `Unresolved reference` inspection instead of annotator and, as a result, it can be suppressed and even turned off if the corresponding inspection is disabled.
Also, only identifier is highlighted not to highlight a lot of other code. Highlighting style was changed to "unknown symbol" to be more consistent with other unresolved references annotations

| Before| After |
| - | - |
| image | image |

changelog: Now `Can't find crate` annotation can be suppressed, highlights only unknown identifier and highlighting style is `unknown symbol` instead of general `error`


Undin:
@avrong it can be interesting for you because `Can't find crate` annotation is the perfect place to try `Add dependency` quick-fix. And now it's a part of `Unresolved reference` inspection that we discussed recently.

mchernyavsky (approved):
bors r+

PR #7065, opened 2021-04-07
DBG: Treat `str*` as `&str` in pretty-printers

### LINKED ISSUE #7331: Debug &str printing is non-deterministic on Windows
## Environment

* **IntelliJ Rust plugin version:** 0.4.148.3911-211
* **Rust toolchain version:** 1.54.0-nightly (dbe459ded 2021-06-02) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2021.1.2 (CL-211.7442.42)

## Problem description

When trying to print &str variables, the result is very non-deterministic. Sometimes the actual string is printed, but often additional characters are added:

image

Note what is the actual length of the string.

## Steps to reproduce

Try to print &str variable on Windows.

ortem:
In LLDB without native Rust support (e.g. MSVC LLDB currently), `&str` type is represented as `str*`.
Both `&str` and `str*` are now treated as string slices by the pretty-printers.

Part of #5632.

Fixes #7331

changelog: Render `&str` string slices content when debugging using MSVC toolchain on Windows

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugProcessConfigurationHelper.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugProcessConfigurationHelper.kt
@@ -159,7 +159,7 @@ class RsDebugProcessConfigurationHelper(
          */
         private val RUST_STD_TYPES: List<String> = listOf(
             "^(alloc::([a-z_]+::)+)String$",
-            "^&str$",
+            "^[&*]?(const |mut )?str\\*?$",
             "^(std::ffi::([a-z_]+::)+)OsString$",
             "^(alloc::([a-z_]+::)+)Vec<.+>$",
             "^(alloc::([a-z_]+::)+)VecDeque<.+>$",
--- a/prettyPrinters/rust_types.py
+++ b/prettyPrinters/rust_types.py
@@ -35,7 +35,8 @@ class RustType(object):
 
 # Should be synchronized with `RsDebugProcessConfigurationHelper.RUST_STD_TYPES`
 STD_STRING_REGEX = re.compile(r"^(alloc::([a-z_]+::)+)String$")
-STD_STR_REGEX = re.compile(r"^&str$")
+# str, mut str, const str*, mut str* (vanilla LLDB); &str, &mut str, *const str, *mut str (Rust-enabled LLDB)
+STD_STR_REGEX = re.compile(r"^[&*]?(const |mut )?str\*?$")
 STD_OS_STRING_REGEX = re.compile(r"^(std::ffi::([a-z_]+::)+)OsString$")
 STD_VEC_REGEX = re.compile(r"^(alloc::([a-z_]+::)+)Vec<.+>$")
 STD_VEC_DEQUE_REGEX = re.compile(r"^(alloc::([a-z_]+::)+)VecDeque<.+>$")
--- a/pretty_printers_tests/tests/string.rs
+++ b/pretty_printers_tests/tests/string.rs
@@ -6,10 +6,22 @@
 // lldbg-check:[...]s1 = "A∆й中" [...]
 // lldb-command:print s2
 // lldbg-check:[...]s2 = "A∆й中" [...]
+// lldb-command:print s3
+// lldbg-check:[...]s3 = "A∆й中" [...]
+// lldb-command:print s4
+// lldbg-check:[...]s4 = "A∆й中" [...]
+// lldb-command:print s5
+// lldbg-check:[...]s5 = "A∆й中" [...]
 // lldb-command:print empty_s1
 // lldbg-check:[...]empty_s1 = "" [...]
 // lldb-command:print empty_s2
 // lldbg-check:[...]empty_s2 = "" [...]
+// lldb-command:print empty_s3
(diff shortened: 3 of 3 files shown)


prettyPrinters/rust_types.py
```diff
@@ -35,7 +35,7 @@ class RustType(object):
 
 # Should be synchronized with `RsDebugProcessConfigurationHelper.RUST_STD_TYPES`
 STD_STRING_REGEX = re.compile(r"^(alloc::([a-z_]+::)+)String$")
-STD_STR_REGEX = re.compile(r"^&str$")
+STD_STR_REGEX = re.compile(r"^&?str\*?$")
```
Undin, line 38: Actually, I cannot get `str*` type for MSVC LLDB
For the following code
```rust
fn main() {
    let mut s = String::from("Hello!");
    let x = s.as_mut();
    let y = x as *mut str;

    let a = "123";
    let b = a as *const str;
    println!("")
}
```
I got
image

i.e. `&str` without Rust support is rendered as `str` (at least for 1.51.0).
So you change works for msvc lldb but only because of `?` after `*`.

Undin, line 38: Also, I've checked the same code
```rust
fn main() {
    let mut s = String::from("Hello!");
    let x = s.as_mut();
    let y = x as *mut str;

    let a = "123";
    let b = a as *const str;
    println!("")
}
```
for unix lldb and found that we don't handle even `&mut str`
image

Maybe it's worth generalizing this PR to support any reference and pointer types for strings (including "c"-style for msvc lldb)?
Note, I'm not sure about gdb where we don't provide a summary (if I remember correctly, we can't do it) but it should work well at least for LLDB where we provide it as a summary and don't hide actual children

WDYT?

ortem: That's interesting. Previously, I saw `str*` on Windows, but now I see `str` too.
I don't know why, but anyway, I generalized the implementation to handle all reference/pointer types

ortem: That's a good idea indeed, thank you for the suggestion!

prettyPrinters/rust_types.py
```diff
@@ -35,7 +35,7 @@ class RustType(object):
 
 # Should be synchronized with `RsDebugProcessConfigurationHelper.RUST_STD_TYPES`
 STD_STRING_REGEX = re.compile(r"^(alloc::([a-z_]+::)+)String$")
-STD_STR_REGEX = re.compile(r"^&str$")
+STD_STR_REGEX = re.compile(r"^[&*]?(const |mut )?str[*]?$")
```
Undin, line 38: Let's add a comment on why we need so complex (and incorrect in rust type terms) regex

Undin (approved):
🔥

Undin:
@ortem Let's rebase it onto master to ensure that pretty-printers tests pass

debugger/src/main/kotlin/org/rust/debugger/runconfig/RsDebugProcessConfigurationHelper.kt
```diff
@@ -165,7 +165,7 @@ class RsDebugProcessConfigurationHelper(
          */
         private val RUST_STD_TYPES: List<String> = listOf(
             "^(alloc::([a-z_]+::)+)String$",
-            "^&str$",
+            "^[&*]?(const |mut )?str[*]?$",
```
Undin, line 168: nit: I suppose `\*` is more common than `[*]`

prettyPrinters/rust_types.py
ortem: I've already described this in the commit message body. The list of supported Rust types is quite big so I decided not to repeat it in a comment.

Do you think a commit message is not enough in such cases? I'd like to get your opinion on it because sometimes I use commit message bodies for descriptions but now I'm not pretty sure if they are convenient for others.

Undin: From my point of view, commit message is less convenient for reader because
- with comment, you don't need to do anything else, just read the code
- if someone makes a change on this line, it will take more time to find the commit that introduced necessary change to read its message

At the same time, comments tend to be outdated so I'd prefer to have both comment to simplify reading in most cases and useful commit message to get the truth if something happens with comment

ortem: Thank you for your suggestion, I added a comment. I will try to use both commit messages and comments for such explanations

pretty_printers_tests/tests/string.rs
```diff
@@ -6,10 +6,22 @@
 // lldbg-check:[...]s1 = "A∆й中" [...]
 // lldb-command:print s2
 // lldbg-check:[...]s2 = "A∆й中" [...]
+// lldb-command:print s3
+// lldbg-check:[...]s3 = "A∆й中" [...]
+// lldb-command:print s4
+// lldbg-check:[...]s4 = "A∆й中" [...]
+// lldb-command:print s5
+// lldbg-check:[...]s2 = "A∆й中" [...]
```
Undin, line 14:
```suggestion
// lldbg-check:[...]s5 = "A∆й中" [...]
```

Undin (approved):

ortem:
bors r=Undin

PR #7067, opened 2021-04-08
Invoke crates local index update on on cargo registry repository changes

avrong:
Part of #6463

changelog: Crates local index for dependencies completion and inspections is now being updated on project changes. Note, currently this feature is experimental and can be enabled with `org.rust.crates.local.index` option in `Experimental Features` dialog.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
@@ -9,7 +9,6 @@ import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.json.JsonMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.fasterxml.jackson.module.kotlin.registerKotlinModule
-import com.google.common.annotations.VisibleForTesting
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.*
@@ -17,11 +16,13 @@ import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.progress.ProgressIndicator
 import com.intellij.openapi.progress.Task
+import com.intellij.openapi.vfs.LocalFileSystem
+import com.intellij.openapi.vfs.VfsUtilCore
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapi.vfs.newvfs.BulkFileListener
+import com.intellij.openapi.vfs.newvfs.RefreshQueue
 import com.intellij.openapi.vfs.newvfs.events.VFileEvent
 import com.intellij.openapi.vfs.newvfs.events.VFilePropertyChangeEvent
-import com.intellij.openapiext.isUnitTestMode
 import com.intellij.util.io.*
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.api.errors.GitAPIException
@@ -34,7 +35,6 @@ import org.eclipse.jgit.treewalk.TreeWalk
 import org.eclipse.jgit.treewalk.filter.OrTreeFilter
 import org.eclipse.jgit.treewalk.filter.PathFilter
 import org.eclipse.jgit.treewalk.filter.TreeFilter
-import org.rust.cargo.CargoConstants
 import org.rust.openapiext.RsPathManager
 import org.rust.stdext.cleanDirectory
 import org.rust.toml.crates.local.CratesLocalIndexServiceImpl.Companion.CratesLocalIndexState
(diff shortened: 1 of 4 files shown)


Undin (commented):
What's about updates of a registry itself? For example, you can trigger registry update from a terminal without any changes in the project and currently we won't know about it

A possible solution is to subscribe to updates of the registry itself. You can use `com.intellij.openapi.vfs.LocalFileSystem#addRootToWatch` to force IDE track changes in registry files and `com.intellij.openapi.vfs.VirtualFileManager#VFS_CHANGES` topic to get the corresponding events.
The suggested solution is quite low-level and requires handling the following cases:
- handle only changes from registry directory
- don't read file content from virtual files related to registry directory not to save a snapshot of registry content in IDE caches
- do nothing if there are no open cargo projects not to affect other projects
- file system may produce a lot of events and probably it doesn't make sense to call `updateIfNeeded` each time (not sure, need to check). But `MergingUpdateQueue` may help to collapse updates

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -5,24 +5,50 @@
...
+            val root = LocalFileSystem.getInstance().refreshAndFindFileByIoFile(File(userCargoRegistryIndexLocation)) ?: return
+            VfsUtilCore.processFilesRecursively(root) { true }
+            RefreshQueue.getInstance().refresh(true, true, null, root)
+
+            ApplicationManager.getApplication().messageBus.connect().subscribe(VirtualFileManager.VFS_CHANGES, object : BulkFileListener {
+                override fun after(events: MutableList<out VFileEvent>) {
+                    if (events.any { it.path.startsWith(userCargoRegistryIndexLocation) }) {
+                        cratesService.updateIfNeeded()
+                    }
+                }
```
Undin, line 48: Looks like we invoke this code on each project opening, i.e if a user opens two rust projects, we will set up two VFS listeners.
Yeah, we won't invoke two update tasks because of `org.rust.toml.crates.local.CratesLocalIndexServiceImpl#updateIfNeeded` implementation but anyway, let's fix it.

The solution is quite simple - just to move all this logic into init block of `CratesLocalIndexServiceImpl`

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -5,24 +5,50 @@
...
         if (isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
-            CratesLocalIndexService.getInstance()
+            val cratesService = CratesLocalIndexService.getInstance()
+
+            val userCargoRegistryIndexLocation = Paths.get(System.getProperty("user.home"), CratesLocalIndexServiceImpl.CARGO_REGISTRY_INDEX_LOCATION).toString()
+            LocalFileSystem.getInstance().addRootToWatch(userCargoRegistryIndexLocation, true)
+
+            val root = LocalFileSystem.getInstance().refreshAndFindFileByIoFile(File(userCargoRegistryIndexLocation)) ?: return
+            VfsUtilCore.processFilesRecursively(root) { true }
+            RefreshQueue.getInstance().refresh(true, true, null, root)
```
Undin, line 41: Looks like we always can work with `Path` and don't convert them into `String` and `File`

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -5,24 +5,50 @@
...
     override fun runActivity(project: Project) {
         if (isUnitTestMode) return
         if (!project.cargoProjects.hasAtLeastOneValidProject) return
 
         if (isFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX)) {
-            CratesLocalIndexService.getInstance()
+            val cratesService = CratesLocalIndexService.getInstance()
+
+            val userCargoRegistryIndexLocation = Paths.get(System.getProperty("user.home"), CratesLocalIndexServiceImpl.CARGO_REGISTRY_INDEX_LOCATION).toString()
+            LocalFileSystem.getInstance().addRootToWatch(userCargoRegistryIndexLocation, true)
```
ortem, line 37: Probably this is enough to check the changes and easier to watch, WDYT?
```suggestion
            LocalFileSystem.getInstance().addRootToWatch("$userCargoRegistryIndexLocation/refs/", true)
```

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -5,24 +5,50 @@
...
+
+            val root = LocalFileSystem.getInstance().refreshAndFindFileByIoFile(File(userCargoRegistryIndexLocation)) ?: return
```
ortem, line 39: It is not obvious why `LocalFileSystem#addRootToWatch` does not work without these additional three lines of code. So I asked IntelliJ Platform team and here is what I learned.

VFS fills up lazily: any directory or file must be explicitly requested by the client to be included in VFS.

Here the `LocalFileSystem#refreshAndFindFileByIoFile` call requests the root directory. After that, the root directory is actually added to VFS (but without subdirectories). Another two lines request all the contents, and finally, we have everything included in VFS.

Another fact: to receive events about new subdirectories (created after our refresh), the client should explicitly request their contents. Fortunately, I suppose, we are not interested in new subdirectories here.

The reason why this routine is so complicated is that the main client of `LocalFileSystem#addRootToWatch` is the project model which maintains everything inside content roots by itself. So any other clients should do the same work as well.

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
avrong: I didn't find a way to pass `Path` in every case, so I made a `Path` and used it's `Srtring` representation everywhere

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -286,7 +289,7 @@ class CratesLocalIndexServiceImpl
             get() = RsPathManager.pluginDirInSystem().resolve("crates-local-index")
 
         // TODO: Determine how path to index is created
-        private const val CARGO_REGISTRY_INDEX_LOCATION: String = ".cargo/registry/index/github.com-1ecc6299db9ec823/.git/"
+        const val CARGO_REGISTRY_INDEX_LOCATION: String = ".cargo/registry/index/github.com-1ecc6299db9ec823/.git/"
```
Undin, line 292: Looks like we don't need this change now

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
```diff
@@ -14,15 +14,15 @@ import org.rust.openapiext.isFeatureEnabled
 
 /**
  * [CratesLocalIndexService] initializer.
- * Only tries to get service instance, without any updates to its crates index.
+ * Tries to get a service instance and binds crates local index update cargo registry index repository changes.
```
Undin, line 17: I suppose we need to update this comment since we moved binding to index repo update into service itself

Undin (approved):
Looks ok except minor comments

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexStartupActivity.kt
avrong: Changed it to `Tries to get and initialize a service instance and updates it if it is needed` as the initializer by itself doesn't bind anything.

avrong:
bors r=Undin

Undin:
bors retry

PR #7070, opened 2021-04-09
ANN: If no `crate` exists, no check is performed

### LINKED ISSUE #7044: This feature is only available in edition 2018 for async/await syntax in README.md
## Environment

* **IntelliJ Rust plugin version:** 0.3.144.3766-203
* **IDE name and version:** CLion 2020.3.3
* **Operating system:** MacOS

## Steps to reproduce
Have a README.md
```
` ``rust
async fn foo(){}
` ``
```
image

Stzx:
In the case of language injection, crate information may not be present in the context.

Fixed #7044


mchernyavsky (approved):
Thanks!

bors r+

vlad20012:
Btw maybe it's better to assume 2018 edition  by default

Undin:
@vlad20012 I think this change is ok - to do nothing if there isn't necessary info.
But we can actually consider the 2018 edition as the default one for highlighting

PR #7071, opened 2021-04-09
INT: provide `Share in Playground` action as intention for selected Rust code

Undin:
It should promote the corresponding feature because `Alt`+`Enter` is one of the most popular shortcuts in IntelliJ IDEs

changelog: provide `Share in Playground` intention to share selected Rust code on [Rust Playground](https://play.rust-lang.org)


Undin:
bors retry

PR #7072, opened 2021-04-09
INT: make `Show the result of macro expansion` intention low priority

Undin:
The change makes IDE place the intention below other general intentions in the intention list.
Also, now the intention isn't launched under write action because it doesn't change code

| Before | After |
| - | - |
|  |  |

changelog: place `Show the result of macro expansion` intention lower in the intention list


mchernyavsky (approved):
bors r+

Undin:
bors retry

Issue #7075, opened 2021-04-09
Adding visibility modifier via "Change Signature" adds it to the function attribute


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.146.3786-211-nightly
* **Rust toolchain version:** 1.52.0-nightly (b3e19a221 2021-03-12) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.1 (CL-211.6693.114)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

Visibility modifier is added to the function attribute when I add it through the "Change Signature" refactoring.

## Steps to reproduce

```rust
#[allow(clippy::needless_lifetimes)]
fn add_with_lifetimes<'a, 'b>(i: &'a i32, j: &'b i32) -> i32 {
    *i + *j
}
```

Call "Change Signature" and add `pub` to the function

```rust
pub #[allow(clippy::needless_lifetimes)]
fn add_with_lifetimes<'a, 'b>(i: &'a i32, j: &'b i32) -> i32 {
    *i + *j
}
```

https://user-images.githubusercontent.com/2500320/114203797-29608180-9961-11eb-96f1-275a91577fb0.mov

PR #7076, opened 2021-04-09
FIX: correctly place visibility in ChangeSignature with attributes

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7075

changelog: Correctly place function visibility after attributes in Change signature refactoring.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/changeSignature/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/changeSignature/impl.kt
@@ -11,6 +11,7 @@ import com.intellij.psi.PsiWhiteSpace
 import com.intellij.refactoring.changeSignature.ParameterInfo
 import com.intellij.refactoring.rename.RenameUtil
 import com.intellij.usageView.UsageInfo
+import org.rust.ide.intentions.visibility.ChangeVisibilityIntention
 import org.rust.ide.refactoring.findBinding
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/refactoring/changeSignature/impl.kt
```diff
@@ -98,7 +98,9 @@ fun changeVisibility(function: RsFunction, config: RsChangeFunctionSignatureConf
 
     val vis = config.visibility
     if (vis != null) {
-        function.addBefore(vis, function.firstChild)
+        val firstChild = function.firstChild
+        val anchor = firstChild.takeIf { it !is RsAttr } ?: firstChild.getNextNonWhitespaceSibling()
```
mchernyavsky, line 102: It doesn't work for this case:

```rust
// qwerty
fn foo() {
    todo!()
}
```

Note: `ChangeVisibilityIntention.makePublic` can be useful.

Kobzol: I looked at the implementation, but thought that handling attributes is enough. Well, it isn't :) Thanks. I reused the metod from `ChangeVisibilityIntention`. Not sure if it should be placed in some other place.

mchernyavsky (approved):
Thanks!

bors r+

PR #7081, opened 2021-04-09
FIX: check namespaces in ChangeSignature name conflict detection

Kobzol:
`ChangeSignature` had a bug where it was incorrectly producing a name conflict between e.g. modules and the refactored function, which is incorrect. I found out about it randomly while testing it.

I was aware of this imperfection while implementing the refactoring, but I thought that because functions and types (like structs) have different naming conventions, it should not be a big deal. Well, turns out that modules kind of break that assumption :) Now it should be working properly.

```rust
mod foo {}
fn foo/*caret*/() {}
```
Running the refactoring on this function should not produce name visibility conflicts after this PR.

changelog: Fix some false positives in name conflict detection in ChangeSignature refactoring.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureUsageProcessor.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureUsageProcessor.kt
@@ -18,6 +18,8 @@ import org.rust.ide.presentation.getPresentation
 import org.rust.lang.core.psi.RsFunction
 import org.rust.lang.core.psi.RsImplItem
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.namespaces
+import org.rust.stdext.intersects
 
 class RsChangeSignatureUsageProcessor : ChangeSignatureUsageProcessor {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureUsageProcessor.kt
```diff
@@ -132,11 +134,14 @@ private fun findNameConflicts(
         if (item == function) continue
         if (!item.isEnabledByCfgSelf) continue
 
-        if (item.name == config.name) {
+        val namedItem = item as? RsNamedElement ?: continue
```
dima74, line 137:
```suggestion
        if (item !is RsNamedElement) continue
```

dima74 (approved):

src/main/kotlin/org/rust/ide/refactoring/changeSignature/RsChangeSignatureUsageProcessor.kt
Kobzol: Actually this is needed, otherwise the `name` method call will be ambiguous.
image

Kobzol:
bors r=dima74

PR #7093, opened 2021-04-12
CARGO: fetch cfg options according to Cargo config

### LINKED ISSUE #5183: Lack of intellisense when target_arch other than current machine's
## Environment

* **IntelliJ Rust plugin version:** 0.2.118.2157-201
* **IDE name and version:**  CLion 2020.1 EAP
Build #CL-201.6251.14, built on March 11, 2020
Experimental macro resolution is enabled!

## Problem description
obraz

I'd like to have code completion for `js_sys::Object`, and other conditionally disabled types somehow. Is there an option to set the `target_arch` somehow to something different than the current machine's?
`js_` ✔️  `js_sys`
`js_sys::Re` ✔️  `js_sys::Reflect`
Even if the above are completed, then on `ctrl+q` they have no documentation, although `ctrl+q` on `js_sys` does show docs

`Object::` :x: empty nothing
`js_sys::Obj` ❌ `js_sys::Object`
`JsValue::fr` ✔️ `JsValue::from_bool` and many more
This is a weird one:
`js_sys::Reflect::se` ✔️ `js_sys::Reflect::set`
`Reflect::se` ❌  `Reflect::set`

## Steps to reproduce

```rust
#[cfg(target_arch = "wasm32")]
use {
	wasm_bindgen::{
		prelude::*,
		JsValue
	},
	js_sys::{
		Object,
		Reflect
	},
}
#[cfg(target_arch = "wasm32")]
#[cfg_attr(target_arch= "wasm32", wasm_bindgen)]
impl SomeStruct {
	#[wasm_bindgen]
	pub fn take(self) -> Object {
		let obj = Object::new();
		
	}
}
```

This has the same effect:
```rust
#[cfg(target_arch = "wasm32")]
#[wasm_bindgen]
impl SomeStruct {
...
}
```

Undin:
Now the plugin use `cargo rustc --print cfg` instead of raw `rustc --print cfg` that allows taking [Cargo config](https://doc.rust-lang.org/cargo/reference/config.html) into account, i.e. fetch cfg options for custom build target, `RUSTFLAGS`, etc.
As a result, evaluation of `cfg` attributes also takes such custom options into account and users can set up them.

Note, `cargo rustc --print cfg` option is available only since Rust 1.52. With older toolchain, the plugin uses host `cfg` options

Fixes #5183
Part of #6104
Related to https://github.com/rust-lang/cargo/issues/9342

Screen Shot 2021-04-12 at 15 12 51

changelog: Now the plugin takes [Cargo config](https://doc.rust-lang.org/cargo/reference/config.html) into account during evaluation of `cfg` conditions. Note, it works since Rust 1.52.0

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
@@ -222,7 +222,13 @@ private fun fetchCargoWorkspace(context: CargoSyncTask.SyncContext): TaskResult<
             })
             val manifestPath = projectDirectory.resolve("Cargo.toml")
 
-            val cfgOptions = toolchain.rustc().getCfgOptions(projectDirectory)
+            val cfgOptions = try {
+                cargo.getCfgOption(childContext.project, projectDirectory)
+            } catch (e: ExecutionException) {
+                // BACKCOMPAT: Rust 1.51. Drop fallback `rustc --print cfg` call
+                toolchain.rustc().getCfgOptions(projectDirectory)
+            }
+
             val ws = CargoWorkspace.deserialize(manifestPath, projectDescriptionData, cfgOptions)
             TaskResult.Ok(ws)
         } catch (e: ExecutionException) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/cargo/project/model/impl/CargoSyncTask.kt
```diff
@@ -222,7 +224,23 @@ private fun fetchCargoWorkspace(context: CargoSyncTask.SyncContext): TaskResult<
             })
             val manifestPath = projectDirectory.resolve("Cargo.toml")
 
-            val cfgOptions = toolchain.rustc().getCfgOptions(projectDirectory)
+            val cfgOptions = try {
+                cargo.getCfgOption(childContext.project, projectDirectory)
+            } catch (e: ExecutionException) {
```
vlad20012, line 229: Let's also log the exception

vlad20012 (approved):

PR #7094, opened 2021-04-12
GRD: enable Gradle Build Cache

vlad20012:
From [Gradle docs](https://docs.gradle.org/current/userguide/build_cache.html):

> The Gradle build cache is a cache mechanism that aims to save time by reusing
> outputs produced by other builds. The build cache works by storing build outputs
> and allowing builds to fetch these outputs from the cache when it is determined
> that inputs have not changed, avoiding the expensive work of regenerating them.

In short, now we have a directory `build-cache` where Gradle puts all
(intermediate) build outputs it ever produces and reuses them if possible, 
so frequent git branch switching or IntelliJ platform version switching 
does not cause long build times.

In order to prevent constant growth of `build-cache` there is a mechanism that
removes cached entries that have not been accessed for 30 days (the value is
configured in `settings.gradle.kts`). My `build-cache` directory size is less than
150 Mb after a week of usage.

Now you can mention these messages in a build log:

```
> Task :compileKotlin FROM-CACHE
> Task :compileJava FROM-CACHE
```

Such `FROM-CACHE` records (instead of the usual `UP-TO-DATE`) means that the task
was not `UP-TO-DATE`, but it was **not** re-launched. Instead, the build output
was fetched from the build cache.

For me, this drastically improves the build time in the case of the platform version
switching.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -60,7 +60,7 @@ plugins {
 idea {
     module {
         // https://github.com/gradle/kotlin-dsl/issues/537/
-        excludeDirs = excludeDirs + file("testData") + file("deps") + file("bin")
+        excludeDirs = excludeDirs + file("testData") + file("deps") + file("bin") + file("build-cache")
     }
 }
 
@@ -380,6 +380,10 @@ project(":") {
                 into("bin")
                 include("**")
             }
+        }
+
+        clean {
+            delete(*(File("${rootDir}/build-cache").listFiles() ?: emptyArray()))
         }
     }
 
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -16,3 +16,13 @@ include(
     "ml-completion",
     "intellij-toml"
 )
+
+// Configure Gradle Build Cache. It is enabled in `gradle.properties` via `org.gradle.caching`.
+// Also, `gradle clean` task is configured to delete `build-cache` directory.
+buildCache {
+    local {
+        isEnabled = System.getenv("CI") == null
+        directory = File(rootDir, "build-cache")
+        removeUnusedEntriesAfterDays = 30
+    }
+}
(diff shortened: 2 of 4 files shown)


settings.gradle.kts
```diff
@@ -16,3 +16,12 @@ include(
     "ml-completion",
     "intellij-toml"
 )
+
+// Configure Gradle Build Cache. It is enabled in `gradle.properties` via `org.gradle.caching`.
+buildCache {
+    local {
+        isEnabled = System.getenv("CI") == null
+        directory = File(rootDir, "build/build-cache")
```
Undin, line 24: It doesn't work with `clean` task properly because `clean` task remove `build` dir and you get
```
Execution failed for task ':common:compileKotlin'.
> Failed to store cache entry for task ':common:compileKotlin'
```

And I don't know how to avoid it because all code in `settings.gradle` executed before any task execution and at the same time it seems caching code doesn't call `mkdirs` for its `directory`

Maybe it's worth just using a separate `build-cache` dir and mark it as excluded dir as we do with `deps` dir?

Undin (changes requested):

settings.gradle.kts
vlad20012: I moved `build-cache` directory to the top level and configured `clean` task to also clean this directory

build.gradle.kts
```diff
@@ -391,6 +391,12 @@ project(":") {
                 .forEach { it.get().resolve() }
         }
     }
+
+    tasks.named("clean") {
+        doLast {
+            delete(*(File("${rootDir}/build-cache").listFiles() ?: emptyArray()))
+        }
+    }
```
Undin, line 399: nit: you can move it into `tasks` section above and write something like
```
clean {
    delete(*file("${rootDir}/build-cache").listFiles().orEmpty())
}
```

Undin (approved):

vlad20012:
bors r=undin

Undin:
bors retry

Issue #2419, opened 2018-03-30
Support declarative macros 2.0


abonander:
The following produces parse errors:

```rust
macro takes_u32_returns_u32($ident:ident) {
// ^ `macro` unexpected    ^      ^     ^ expected `!` got `)`
//                         ^      ^ expected `!` got `:`
//                         ^ expected `!` got `(`         
//                                      
    fn $ident (arg: u32) -> u32 { arg }
}

**Comment by vlad20012** [comment]
What's the current status of macros 2.0 in rust language? I can find the only [RFC 1584](https://github.com/rust-lang/rfcs/blob/master/text/1584-macros.md), that looks like an abstract, but not detailed design. Can you please like more info about it? If there is no even documentation, I think it's too low priority task for us.

**Comment by abonander** [comment]
Most of it is still in-flux and implementation defined. For now it should probably allow anything with the following form:

```rust
macro ident(tt*)  {
  tt*
}
```

PR #7098, opened 2021-04-12
RES: Basic resolve of macros 2.0 with new resolve

dima74:
Support basic resolve of macros 2.0:
```rust
pub macro foo() {
    ...
}
```

Rust 1.51 added two such macros, `addr_of` and `addr_of_mut`.
Note that importing macros 2.0 using `#[macro_use] extern crate ...` is not supported by this PR.

changelog: Support basic resolve of macros 2.0 (`pub macro name() { ... }`) with new name resolution engine

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
@@ -129,7 +129,8 @@ val RsItemsOwner.expandedItemsCached: RsCachedItems
 class RsCachedItems(
     val namedImports: List<CachedNamedImport>,
     val starImports: List<CachedStarImport>,
-    val macros: List<RsMacro>,
+    /** [RsMacro2] are stored in [named] */
+    val legacyMacros: List<RsMacro>,
     val named: Map<String, List<RsItemElement>>,
     val namedCfgDisabled: Map<String, List<RsItemElement>>,
 ) {
--- a/src/main/kotlin/org/rust/lang/core/resolve2/FileModificationTracker.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve2/FileModificationTracker.kt
@@ -66,6 +66,7 @@ private fun calculateModHash(modData: ModDataLight): HashCode {
     data.writeElements(modData.imports)
     data.writeElements(modData.macroCalls)
     data.writeElements(modData.macroDefs)
+    data.writeElements(modData.macro2Defs)
     data.writeByte(modData.attributes?.ordinal ?: RsFile.Attributes.values().size)
 
     return HashCode.fromByteArray(digest.digest())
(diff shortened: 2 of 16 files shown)


src/main/kotlin/org/rust/lang/core/psi/ext/RsItemsOwner.kt
```diff
@@ -61,7 +61,7 @@ val RsItemsOwner.expandedItemsCached: RsCachedItems
     get() = CachedValuesManager.getCachedValue(this, EXPANDED_ITEMS_KEY) {
         val namedImports = SmartList<CachedNamedImport>()
         val starImports = SmartList<CachedStarImport>()
-        val macros = SmartList<RsMacro>()
+        val macros = SmartList<RsMacroDefinitionBase>()
```
vlad20012, line 64: I think we should put macros 2.0 to namedCfgEnabled/namedCfgDisabled because they are just like usual named elements. In this case, they should also work in old name resolution (e.g. if they are defined in a function body). `macros` field could be renamed to `legacyMacros`.

src/main/kotlin/org/rust/lang/core/resolve2/FileModificationTracker.kt
```diff
@@ -77,13 +78,14 @@ private class ModDataLight {
...
     val macroDefs: MutableList<MacroDefLight> = mutableListOf()
+    val macro2Defs: MutableList<Macro2DefLight> = mutableListOf()
     var attributes: RsFile.Attributes? = null  // not null only for crate root
 
     fun sort() {
         items.sortBy { it.name }
         enums.sortBy { it.item.name }
         imports.sortWith(compareBy(Arrays::compare) { it.usePath })
-        // TODO: Smart sort for macro calls & defs
+        macro2Defs.sortBy { it.name }
```
vlad20012, line 88: Does the order matter?

src/test/kotlin/org/rust/lang/core/resolve/RsPackageLibraryResolveTest.kt
```diff
@@ -526,6 +526,28 @@ class RsPackageLibraryResolveTest : RsResolveTestBase() {
         pub use foo_ as foo;
     """)
 
+    @UseNewResolve
+    @MockEdition(CargoWorkspace.Edition.EDITION_2018)
```
vlad20012, line 530: Let's also add tests where a macro is defined:
- in the same mod
- in other mod in the same crate
- locally in a function body
And let's also test that we don't resolve macro 2.0 in lexical scope.

vlad20012 (approved):
LGTM, but one test failed

ortem:
@dima74 WDYT about linking https://github.com/intellij-rust/intellij-rust/issues/2419 and using it as a tracking issue for everything related to macros 2.0 support?

PR #7100, opened 2021-04-12
Associate `poetry.lock` files with TOML file type

abn:
changelog: Consider poetry.lock as TOML file


Undin (approved):
Thanks!
And welcome to contributors!

bors r+

Issue #7092, opened 2021-04-12
Incorrect elide lifetimes suggestion


PizzaCrust:
## Environment

* **IntelliJ Rust plugin version:** 0.3.146.3792-211-nightly
* **Rust toolchain version:** 1.53.0-nightly (07e0e2ec2 2021-03-24) x86_64-pc-windows-msvc
* **IDE name and version:** IntelliJ IDEA 2021.1 Community Edition (IC-211.6693.111)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
The IDE suggests that you elide lifetimes, however when you elide the lifetimes it results in invalid code. (you can't elide lifetimes here)
![image](https://user-images.githubusercontent.com/15954148/114358443-1fc35d80-9b9d-11eb-96ba-ca3a5b37956d.png)
When you elide lifetimes:
![image](https://user-images.githubusercontent.com/15954148/114358498-2f42a680-9b9d-11eb-8876-239ffe0983e5.png)

## Steps to reproduce
```rust
trait A<'b> {}

struct C<T>(T);

impl<T> C<T> {
    fn d<'e>(&'e self) where T: A<'e>   {

    }
}
```

PR #7101, opened 2021-04-12
FIX: better check where clause in RsNeedlessLifetimesInspection

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7092

changelog: Remove false positive in RsNeedlessLifetimesInspection when the lifetime is used in where clause.


mchernyavsky (approved):
Thanks!

bors r+

Issue #7096, opened 2021-04-12
Detect unknown cfg predicates E0537


Undin:
`cfg` attribute may have only `not`, `all` or `any` predicates according to [reference](https://doc.rust-lang.org/reference/conditional-compilation.html). Other ones are treated as [E0537](https://doc.rust-lang.org/error-index.html#E0537) errors by the compiler.

Currently, the plugin doesn't detect such errors, so you can easily write `non` instead of `not` and get quite weird behavior of the plugin. It would be great to highlight wrong predicates.

Part of #886

```rust
#[cfg(target_os = "android")]
fn foo() {}

#[cfg(non(target_os = "android"))] // <- error[E0537]: invalid predicate `non`
fn foo() {}

fn main() {
    foo();
}
```

PR #7102, opened 2021-04-12
ANN: annotate unknown cfg predicates [E0537]

Kobzol:
This PR adds error annotations for unknown `cfg` predicates.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7096

changelog: Annotate unknown cfg attributes. This corresponds to the E0537 compiler error.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -109,9 +109,36 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             override fun visitPatTupleStruct(o: RsPatTupleStruct) = checkRsPatTupleStruct(rsHolder, o)
             override fun visitPatTup(o: RsPatTup) = checkRsPatTup(rsHolder, o)
             override fun visitStructLiteralField(o: RsStructLiteralField) = checkReferenceIsPublic(o, o, rsHolder)
+            override fun visitMetaItem(o: RsMetaItem) = checkMetaItem(rsHolder, o)
         }
 
         element.accept(visitor)
+    }
+
+    private fun checkMetaItem(holder: RsAnnotationHolder, metaItem: RsMetaItem) {
+        val args = metaItem.metaItemArgs
+        val name = metaItem.name
+        if (metaItem.isRootMetaItem() && args != null && name in listOf("cfg", "cfg_attr")) {
+            val item = args.metaItemList.getOrNull(0) ?: return
+            checkCfgPredicate(holder, item)
+        }
+    }
+
+    private fun checkCfgPredicate(holder: RsAnnotationHolder, item: RsMetaItem) {
+        val itemName = item.name ?: return
+        val args = item.metaItemArgs ?: return
+        when (itemName) {
+            "all", "any" -> args.metaItemList.forEach { checkCfgPredicate(holder, it) }
+            "not" -> {
+                val parameter = args.metaItemList.getOrNull(0) ?: return
+                checkCfgPredicate(holder, parameter)
+            }
+            "version" -> { /* version is currently experimental */ }
+            else -> {
+                val path = item.path ?: return
+                RsDiagnostic.UnknownCfgPredicate(path, itemName).addToHolder(holder, checkCfg = false)
+            }
+        }
     }
 
     private fun checkRsPatTup(holder: RsAnnotationHolder, pattern: RsPatTup) {
--- a/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
@@ -4149,4 +4149,64 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
         use foo::T1 as _;
         use bar::T2 as _;
     """)
+
+    fun `test no E0537 valid cfg`() = checkErrors("""
+        #[cfg(any(foo, bar))]
+        #[cfg(all(foo, baz))]
+        #[cfg(not(foo))]
+        #[cfg(all(not(foo)))]
+        #[cfg(not(any(foo)))]
+        #[cfg(foo)]
+        fn foo() {}
+    """)
+
+    fun `test E0537 invalid cfg`() = checkErrors("""
+        #[cfg(<error descr="Invalid predicate `an` [E0537]">an</error>(foo))]
+        #[cfg(<error descr="Invalid predicate `allx` [E0537]">allx</error>(foo))]
+        #[cfg(<error descr="Invalid predicate `non` [E0537]">non</error>(foo))]
+        #[cfg(<error descr="Invalid predicate `non` [E0537]">non</error>(an(foo)))]
+        #[cfg(all(x, <error descr="Invalid predicate `bar` [E0537]">bar</error>()))]
+        #[cfg(all(x, not(y), <error descr="Invalid predicate `baz` [E0537]">baz</error>()))]
+        #[cfg(not(<error descr="Invalid predicate `foo` [E0537]">foo</error>()))]
+        #[cfg(all(not(any(bar, <error descr="Invalid predicate `baz` [E0537]">baz</error>()))))]
+        #[cfg(any(x, not(y), <error descr="Invalid predicate `baz` [E0537]">baz</error>()))]
+        fn foo() {}
+    """)
+
+    fun `test no E0537 valid cfg_attr`() = checkErrors("""
+        #[cfg_attr(any(foo), bar)]
+        #[cfg_attr(all(foo), bar)]
+        #[cfg_attr(not(foo), bar)]
+        #[cfg_attr(all(not(foo)), bar)]
+        #[cfg_attr(not(any(foo)), bar)]
+        #[cfg_attr(non, bar)]
+        fn foo() {}
+    """)
+
+    fun `test E0537 invalid cfg_attr`() = checkErrors("""
+        #[cfg_attr(<error descr="Invalid predicate `an` [E0537]">an</error>(foo), bar)]
+        #[cfg_attr(<error descr="Invalid predicate `allx` [E0537]">allx</error>(foo), bar)]
+        #[cfg_attr(<error descr="Invalid predicate `non` [E0537]">non</error>(foo), bar)]
+        #[cfg_attr(<error descr="Invalid predicate `non` [E0537]">non</error>(an(foo)), bar)]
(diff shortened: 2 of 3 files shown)


src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -109,11 +109,36 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+        val itemName = item.name ?: return
+        val args = item.metaItemArgs
+        if (args != null) {
+            when (itemName) {
+                "all", "any" -> args.metaItemList.forEach { checkCfgPredicate(holder, it) }
+                "not" -> {
+                    val parameter = args.metaItemList.getOrNull(0) ?: return
+                    checkCfgPredicate(holder, parameter)
+                }
+                else -> RsDiagnostic.UnknownCfgPredicate(item, itemName).addToHolder(holder)
```
Undin, line 137: Not a call to action: don't like that compiler highlights the whole attribute instead of the identifier only

vlad20012:
There is also a nightly syntax `#[cfg(version("1.42"))]` that looks like a predicate https://doc.rust-lang.org/nightly/unstable-book/language-features/cfg-version.html

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -109,11 +109,36 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+        if (args != null && name in listOf("cfg", "cfg_attr")) {
+            val item = args.metaItemList.getOrNull(0) ?: return
+            checkCfgPredicate(holder, item)
+        }
+    }
+
+    private fun checkCfgPredicate(holder: RsAnnotationHolder, item: RsMetaItem) {
+        val itemName = item.name ?: return
+        val args = item.metaItemArgs
+        if (args != null) {
```
Undin, line 130: nit: you can use `item.metaItemArgs ?: return` to void `if` and additional nesting

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -109,11 +109,36 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+            override fun visitMetaItem(o: RsMetaItem) = checkMetaItem(rsHolder, o)
         }
 
         element.accept(visitor)
     }
 
+    private fun checkMetaItem(holder: RsAnnotationHolder, metaItem: RsMetaItem) {
+        val args = metaItem.metaItemArgs
+        val name = metaItem.name
+        if (args != null && name in listOf("cfg", "cfg_attr")) {
```
vlad20012, line 121: Let's also check `metaitem.isRootMetaItem()`. And let's add tests for such check:
- a `cfg`/`cfg_attr` that means something non-relevant - `#[foobar(cfg(baz))]`
- a nested `cfg`/`cfg_attr` - `#[cfg_attr(foo, cfg_attr(error_predicate(), baz))]`

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
vlad20012: I'd prefer an item name

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
```diff
@@ -4149,4 +4149,44 @@ class RsErrorAnnotatorTest : RsAnnotatorTestBase(RsErrorAnnotator::class) {
...
+        fn foo() {}
+    """)
+
+    fun `test E0537 invalid cfg`() = checkErrors("""
+        #[cfg(<error descr="Invalid predicate `an` [E0537]">an(foo)</error>)]
+        #[cfg(<error descr="Invalid predicate `allx` [E0537]">allx(foo)</error>)]
+        #[cfg(<error descr="Invalid predicate `non` [E0537]">non(foo)</error>)]
+        #[cfg(<error descr="Invalid predicate `non` [E0537]">non(an(foo))</error>)]
+        #[cfg(not(<error descr="Invalid predicate `foo` [E0537]">foo()</error>))]
+        #[cfg(all(not(any(bar, <error descr="Invalid predicate `baz` [E0537]">baz()</error>))))]
```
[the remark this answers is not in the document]
Kobzol: I noticed that `rustc` ignores (short-circuits) further predicates, even if they are invalid.
```rust
#[cfg(any(not(y), baz()))]
fn foo() {}
```
This does not produce an error with `rustc`. But I suppose that we want to show an error, since it's invalid and probably a user mistake?

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
Kobzol: This mirrors the behavior of `rustc`:
image
But if you prefer to highlight only the attribute name, I will change it. I suppose that it would be more readable for the user.

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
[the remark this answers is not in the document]
Kobzol: Hmm, actually I see that the plugin works similarly.
For this code:
```rust
 #[cfg(all(x, bar()))]
fn foo() {}
```
My annotation marks `bar()`, but the annotation si then not shown (not even in tests), because the whole thing is considered disabled because of short-circuit on `x`.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
Undin: We've discussed it with @vlad20012 and decided that it's ok to have a narrower highlighting range in comparison with the compiler

src/test/kotlin/org/rust/ide/annotator/RsErrorAnnotatorTest.kt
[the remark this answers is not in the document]
Undin: > But I suppose that we want to show an error, since it's invalid and probably a user mistake?

Yep, most likely users want that their code works with any cfg options/features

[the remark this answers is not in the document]
Kobzol: I wonder how to do that though. Because when the rest of the cfg is disabled, it is automatically greyed out and annotations are missing. Some special case in the annotator to ignore cfg-disabling in this case?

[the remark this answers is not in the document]
Undin: We can add `checkCfg` flag to `org.rust.lang.utils.RsDiagnosticKt#addToHolder(org.rust.lang.utils.RsDiagnostic, org.rust.ide.annotator.RsAnnotationHolder)` method with `true` as default value
@vlad20012 WDYT?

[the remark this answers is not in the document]
vlad20012: Yep, sounds good to me.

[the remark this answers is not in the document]
Kobzol: Done.

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -109,11 +109,37 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+            checkCfgPredicate(holder, item)
+        }
+    }
+
+    private fun checkCfgPredicate(holder: RsAnnotationHolder, item: RsMetaItem) {
+        val itemName = item.name ?: return
+        val args = item.metaItemArgs ?: return
+        when (itemName) {
+            "all", "any" -> args.metaItemList.forEach { checkCfgPredicate(holder, it) }
+            "not" -> {
```
Undin, line 132: Let's also take into account version attribute as mentioned in https://github.com/intellij-rust/intellij-rust/pull/7102#issuecomment-820199568.
For example, stable rustc emit `` error[E0658]: `cfg(version)` is experimental and subject to change `` for the following code:
```rust
#[cfg(version("1.42"))]
struct S;

fn main() {}
```

Undin (commented):
Looks ok except comment

src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
```diff
@@ -109,11 +109,42 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
...
+                val parameter = args.metaItemList.getOrNull(0) ?: return
+                checkCfgPredicate(holder, parameter)
+            }
+            "version" -> {
+                val path = item.path ?: return
+                RsDiagnostic.ExperimentalFeature(path, null, "`cfg(version)` is experimental and subject to change",
```
Undin, line 138: Let's just do nothing in the case of `version` attribute. Currently, this code produces false-positive error even if you have `#![feature(cfg_version)]` and nightly toolchain.
image

If you would like to support it as well, let's do it in separate PR

Undin (changes requested):

Undin (approved):
Great!
Thank you!

bors r+

Undin:
@Kobzol would you like to slightly improve this feature by adding a quick fix to use a proper operator?
For example, we can check [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) between all possible operators and actual one and suggest using that operator when the distance is 1.
To calculate the distance you can use `com.intellij.util.text.EditDistance#levenshtein` method

Kobzol:
Sure! What about creating something like this (I'm not sure about the name):
```kotlin
class DidYouMeanFix(element: RsElement, private val name: String) : QuickFix() {
      fun invoke(startElement) {
          // how generally replace any element with the selected text?
         startElement.text = name // not allowed
         startElement.replace(RsPsiFactory(project).createFromText(name)) // maybe would produce parser errors?
      }

     companion object {
          fun createApplicable(element: RsElement, name: String, availableNames: List<String>): List<DidYouMeanFix> {
              val validNames = availableNames.filter(<distance to name is 1>)
              return validNames.map(DidYouMeanFix)
         }
     }
}
```

Undin:
@Kobzol LGTM in general

> // how to replace any element with the selected text in a general way?

I suppose `com.intellij.psi.PsiElement#replace` is ok in this case. I suggest moving responsibility on what will be created for the given name on caller (via providing the corresponding lambda) and writing the corresponding doc comment about it.

PR #7105, opened 2021-04-13
Fix constructor return type and ellipses usage in prettyPrinters Python stubs

lancelote:
1. Constructor return type is `None`, see [PEP 484][1] explanation

> Note that the return type of `__init__` ought to be annotated with `-> None`. The reason for this is subtle. If `__init__` assumed a return annotation of `-> None`, would that mean that an argument-less, un-annotated `__init__` method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that `__init__` ought to have a return annotation; the default behavior is thus the same as for other methods

2. One cannot use ellipses `...` as `Any`. E.g. `SBTypeList = ...` will be reported by `mypy` as

> Variable "lldb.SBTypeList" is not valid as a type

[1]: https://www.python.org/dev/peps/pep-0484/#the-meaning-of-annotations

changelog: Fix some type annotations in python pretty printers


ortem (approved):
Thank you for the explanations!

bors r+

PR #7111, opened 2021-04-14
Support fold `where`

### LINKED ISSUE #7110: Feature request: hiding `where` clauses
Sometimes `where` clauses with generics may take a lot of space, as you can see at the image below:

image

There is currently no way to hide this, which can be annoying when the user wants to quickly scroll up|down.
I think it can be solved in one of the following ways:
1. Button for hiding `where` clauses, like a button for hiding the arguments of a function or the body of a function.
2. Button to hide the function definition and body with arguments, body and generics, leaving only the function name.

Stzx:
Resolved #7110


mchernyavsky (approved):
Thanks!

bors r+

PR #6845, opened 2021-02-14
INT: add intention to toggle feature state from cfg attribute

Kobzol:
This PR adds a simple intention that toggles feature state. It can be invoked from `cfg` feature attributes.

I needed an intention test with a toolchain, so the test inherits from `RsWithToolchainTestBase` and reimplements a small amount of logic from `RsIntentionTestBase`.

Related issue: https://github.com/intellij-rust/intellij-rust/issues/4693

changelog: Add intention that toggles feature state from a cfg attribute

PR #7115, opened 2021-04-15
INT: prioritize `ToggleFeatureIntention` and extend its availability range

vlad20012:
Some improvements to #6845

1. make `ToggleFeatureIntention` a high priority action. Since the intention has a lower availability scope (a feature meta item only) I think it should have a higher priority than intentions with wider scope:
image

2. extend availability range of `ToggleFeatureIntention`
The intention is now available in the entire meta item, not only in the feature name:
```
#[cfg(feature = "foo")]
//    ~~~~~~~~~~~~~~~
```

changelog: should not be mentioned in the changelog if placed in the same release as #6845


Kobzol (approved):
LGTM.

vlad20012:
bors r=Kobzol

PR #7118, opened 2021-04-15
CI: Setup Qodana

mchernyavsky:


mchernyavsky (commented):
bors r+

mchernyavsky:
(Qodana still produces some false-positives and will be enabled when the problem is fixed by Qodana-team)

PR #7119, opened 2021-04-15
RES: resolve qualified attribute proc macro paths

### LINKED ISSUE #4481: Goto Definition doesn't work for path attribute macros

vlad20012:
Now the plugin should correctly resolve this:

```rust
#[foo::bar] // Resolve `foo::bar`
fn baz() {}
```

Fixes #4481

changelog: correctly resolve qualified attribute proc macro paths (like `#[foo::bar]`)

### DIFF
--- a/src/test/kotlin/org/rust/lang/core/resolve/RsProcMacroResolveTest.kt
+++ b/src/test/kotlin/org/rust/lang/core/resolve/RsProcMacroResolveTest.kt
@@ -112,19 +112,31 @@ class RsProcMacroResolveTest : RsResolveTestBase() {
         struct S;
     """)
 
-    // FIXME
-    fun `test resolve attr proc macro from macro call with full path`() = expect<IllegalStateException> {
-        stubOnlyResolve("""
+    fun `test resolve attr proc macro from macro call with full path`() = stubOnlyResolve("""
         //- dep-proc-macro/lib.rs
             #[proc_macro_attribute]
             pub fn example_proc_macro(attr: TokenStream, item: TokenStream) -> TokenStream { item }
         //- lib.rs
-            use dep_proc_macro
             #[dep_proc_macro::example_proc_macro]
                                 //^ dep-proc-macro/lib.rs
             struct S;
     """)
-    }
+
+    fun `test attr proc macro is not resolved to a declarative macro`() = stubOnlyResolve("""
+        //- dep-proc-macro/lib.rs
+            #[proc_macro_attribute]
+            pub fn example_proc_macro(attr: TokenStream, item: TokenStream) -> TokenStream { item }
+        //- lib.rs
+            use dep_proc_macro::example_proc_macro;
+
+            macro_rules! example_proc_macro {
+                () => {};
+            }
+
+            #[example_proc_macro]
+               //^ dep-proc-macro/lib.rs
+            struct S;
+    """)
 
     fun `test resolve custom derive proc macro in use item`() = stubOnlyResolve("""
     //- dep-proc-macro/lib.rs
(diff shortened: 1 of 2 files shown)


Undin (commented):
Note, this change breaks name resolution for `foo` item in the following code:
```rust
#[lib::bar] // Resolve `lib::bar`
fn baz() {}
```

Note, `lib` is resolved in master currently

**UPD**:
Name resolution works only if library has `foo` name for some reason (at least, for me)

```rust
#[foo::bar]
fn baz() {}
```
Moreover, it works with `foo` even if the library is not called `foo`

Undin:
@vlad20012 looks like it's a cache invalidation problem
Steps to reproduce:
- create the following project:

`Cargo.toml`
```toml
[package]
name = "rust_project"
version = "0.1.0"
authors = []
edition = "2018"

[lib]
name = "lib"
proc-macro = true
```

`lib.rs`
```rust
extern crate proc_macro;
use proc_macro::TokenStream;
#[proc_macro_attribute]
pub fn foo(attr: TokenStream, item: TokenStream) -> TokenStream { item }
```

`src/main.rs`
```rust
use lib::foo;

#[lib::foo]
fn baz() {}

fn main() {}
```

* change library name in `Cargo.toml` from `lib` to `libb`, for example, and invoke `Refresh Cargo projects` actions. After that, `Goto definition` for all references in `use lib::foo`  and `lib` in `#[lib::foo]` works as previously. Only structured change (for example, the addition of new function) helps to drop old resolve info.

* make revert change, i.e. change `libb` from `Cargo.toml` to `lib` again and launch `Refresh Cargo projects`. `Goto definition` action works only for `foo` reference in `#[lib::foo]`.


It seems the described bug is not related to your changes but it would be great to fix it as well

src/test/kotlin/org/rust/lang/core/resolve/RsProcMacroResolveTest.kt
```diff
@@ -112,19 +112,33 @@ class RsProcMacroResolveTest : RsResolveTestBase() {
...
             pub fn example_proc_macro(attr: TokenStream, item: TokenStream) -> TokenStream { item }
         //- lib.rs
-            use dep_proc_macro
             #[dep_proc_macro::example_proc_macro]
                                 //^ dep-proc-macro/lib.rs
             struct S;
     """)
-    }
+
+    @MockEdition(Edition.EDITION_2018)
```
dima74, line 126: Lets remove `@MockEdition` from individual methods since we have `@MockEdition` on the whole class

dima74 (approved):

vlad20012:
@Undin's case is unrelated to the PR, let's fix it deparately

bors r=dima74

Undin:
I've created a separate issue #7143

PR #7134, opened 2021-04-17
ANN: Add a quick fix to add missing trait definitions

Sejsel:
This PR adds one of the quick fixes listed under E0407 in #886 (the most useful one that I have missed a couple of times already). Generates trait members with no default implementation, which is the behavior I would expect from this. Maintains the member order if the order of existing trait and impl members matches.

https://user-images.githubusercontent.com/998408/115119582-7370e500-9fa9-11eb-84b8-c4046ab98938.mp4

Currently, the E0437 and E0438 (the `type` and `const` versions of E0407, respectively) rustc errors are also classified as E0407 by the plugin, and this quick fix solves all three of them.

The current implementation duplicates `RsFunction.getSignatureText()` from the "implement members" fix/refactor. I am not sure what is the best place to move this (if it's even a good idea). The duplicated implementation includes a comment about anonymous parameters preventing simply using a substring of the original signature - if I understand that correctly, this should not be an issue for the impl->trait direction (regardless of Rust edition), but I am not 100% sure. I would appreciate some insight from someone who's not touching the codebase for the first time like me :)

**TODO**
- [x] Solve duplication of `RsFunction.getSignatureText()`
- [x] Add tests (mostly adapted from implement members - the opposite action)

changelog: Add a quickfix to generate missing trait member definitions in trait implementations.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitOrImpl.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitOrImpl.kt
@@ -20,3 +20,6 @@ interface RsTraitOrImpl : RsItemElement, RsGenericDeclaration, RsAttrProcMacroOw
 
 val RsTraitOrImpl.expandedMembers: List<RsAbstractable>
     get() = members?.expandedMembers.orEmpty()
+
+val RsTraitOrImpl.explicitMembers: List<RsAbstractable>
+    get() = members?.stubChildrenOfType<RsAbstractable>().orEmpty()
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -106,6 +106,7 @@ rrevenantt
 saaryogev
 sanmai-NL
 SeekingMeaning
+Sejsel
 sgift
 sgrif
 shevtsiv
--- a/src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/RsSortImplTraitMembersInspection.kt
@@ -20,7 +20,7 @@ class RsSortImplTraitMembersInspection : RsLocalInspectionTool() {
         override fun visitImplItem(impl: RsImplItem) {
             val trait = impl.traitRef?.resolveToTrait() ?: return
             val typeRef = impl.typeReference ?: return
-            if (sortedImplItems(impl.items(), trait.items()) == null) return
+            if (sortedImplItems(impl.explicitMembers, trait.explicitMembers) == null) return
             val textRange = TextRange(
                 (impl.vis ?: impl.default ?: impl.unsafe ?: impl.impl).startOffsetInParent,
                 typeRef.startOffsetInParent + typeRef.textLength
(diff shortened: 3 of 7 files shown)


vlad20012:
Hello! Thanks for the PR! I'm sorry for so long response.

I see you're a first-time contributor. Could you, please, sign our `Contributor License Agreement` as mentioned in [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#cla) and add your username to `CONTRIBUTORS.txt`?

dima74:
@Sejsel could you please rebase on master

Sejsel:
@dima74 
I have rebased it on master and hopefully fixed compilation errors due to changes since the PR was made, but I won't have time to properly adjust this and finish it at the very least until the end of this month.

I couldn't even manually test if it works now as I can't get it to compile (`org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find com.jetbrains.intellij.idea:ideaIC:2022.2.3.`) and I sadly don't have time to play with it for now.

If you want, feel free to take this over. Otherwise I might take a look at it again sometime during October. Thanks!

Sejsel:
I also noticed the ImplementMembersFix this is partially based on got reworked a bit since I made this, it's worth taking a look at to finish this PR.

dima74:
Added tests and refactored `RsFunction.getSignatureText()`
@ortem could you please review

src/main/kotlin/org/rust/lang/core/psi/ext/RsTraitOrImpl.kt
```diff
@@ -20,3 +20,6 @@ interface RsTraitOrImpl : RsItemElement, RsGenericDeclaration, RsAttrProcMacroOw
 
 val RsTraitOrImpl.expandedMembers: List<RsAbstractable>
     get() = members?.expandedMembers.orEmpty()
+
+val RsTraitOrImpl.explicitMembers: List<RsAbstractable>
```
ortem, line 24: I think now we should carefully check all existing usages of `RsTraitOrImpl.members` and replace some of them either with `RsTraitOrImpl.expandedMembers` (if the macro-generated ones are suitable there) or `RsTraitOrImpl.explicitMembers` (otherwise).

For example, we could review the usages from the following classes:
* `org.rust.ide.refactoring.extractTrait.RsExtractTraitHandler`
* `org.rust.ide.refactoring.generate.constructor.GenerateConstructorHandler`

Secondly, I am not sure if we really need `RsMembers.expandedMembers` since we already have `RsTraitOrImpl.expandedMembers`. So maybe we could delete this properly to avoid confusion in the future.

@vlad20012 Maybe you also have some thoughts?

vlad20012: Another option is removing `RsTraitOrImpl.expandedMembers` and using `RsMembers.expandedMembers` or newly introduced `RsMembers.explicitMembers` =D

I'm not sure which is better, but I also find `members`/`expandedMembers`/`explicitMembers` confusing. We can also rename `getMembers()` property/function in the `RsTraitOrImpl` and in the grammar as well (and choose a better name to highlight its low-level nature, like `rawMembers`).

vlad20012: But, to be honest, we currently have no naming style that highlight differences between raw and macro-expanded and cfg-checked parts of PSI

ortem: > But, to be honest, we currently have no naming style that highlights differences between raw and macro-expanded and cfg-checked parts of PSI

Right, I think it would be great to introduce the different naming for them to avoid confusion

ortem (approved):

dima74:
bors r=ortem

dima74:
bors retry

PR #7137, opened 2021-04-18
INT: add intention to create a tuple struct

Kobzol:
This PR extends https://github.com/intellij-rust/intellij-rust/pull/6837. It adds a new intention which allows creating tuple structs from unresolved usage.

I decided to use a simple heuristic to offer the intention. It is offered for an unresolved function call if:
1) The target of the call is a module (so associated function calls are ignored).
2) The name of the called function is in CamelCase.
3) The expected return type of the function call is unknown or missing. i.e. the intention is not offered in this case:
```rust
let x: u32 = Foo/*caret*/(1, 2);
```
Because the code would not compile after the tuple struct would be created and it seems like its not what the user wants.
I'm not sure if 3) is needed, we can also skip it.

changelog: Add an intention to create a tuple struct from an unresolved function call.


vlad20012 (approved):
:+1: Thanks!

bors r+

PR #7141, opened 2021-04-19
RES: Fix completion of derive proc macro in imports

### LINKED ISSUE #7062: Derive proc macro is not suggested in completion
## Environment

* **IntelliJ Rust plugin version:** 0.3.145.SNAPSHOT-211-dev
* **Rust toolchain version:** 1.51.0 (2fd73fabe 2021-03-23) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2021.1 Ultimate Edition (IU-211.6693.103)

## Problem description
Related to #7061

When I try to use `Builder` custom derive macro from `derive_builder` lib, `Builder` is not suggested by the plugin in use item.

image

Note, it's successfully resolved if I write complete use item: `use derive_builder::Builder;`

## Steps to reproduce
`Cargo.toml`
```toml
[package]
name = "rust_project"
version = "0.1.0"
authors = []
edition = "2018"

[dependencies]
derive_builder = "0.10.0"
```

`src/main.rs`
```rust
use derive_builder::/*caret*/;

#[derive(Clone, Debug, Builder)]
pub struct MyStruct {
    foo: String
}

fn main() {}
```

dima74:
Fixes #7062

changelog: Fix completion of derive proc macro with `doc(hidden)` attribute in imports

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
+++ b/src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
@@ -593,7 +593,7 @@ fun filterCompletionVariantsByVisibility(context: RsElement, processor: RsResolv
         if (!it.isVisibleFrom(context)) return@wrapWithFilter false
 
         if (element is RsOuterAttributeOwner) {
-            val isHidden = element.shouldHideElementInCompletion(contextMod)
+            val isHidden = element.shouldHideElementInCompletion(context, contextMod)
             if (isHidden) return@wrapWithFilter false
         }
 
@@ -622,12 +622,17 @@ fun filterDeriveProcMacros(processor: RsResolveProcessor): RsResolveProcessor =
         true
     }
 
-fun RsOuterAttributeOwner.shouldHideElementInCompletion(contextMod: RsMod): Boolean {
+fun RsOuterAttributeOwner.shouldHideElementInCompletion(context: RsElement, contextMod: RsMod): Boolean {
     val elementContainingMod = containingMod
     if (elementContainingMod == contextMod) return false
 
     // Hide `#[doc(hidden)]` items
-    if (queryAttributes.isDocHidden) return true
+    if (queryAttributes.isDocHidden) {
+        val isDeriveMacroInImport = context.ancestorStrict<RsUseItem>() != null
+            && this is RsFunction && isCustomDeriveProcMacroDef
+            && containingCrate != contextMod.containingCrate
+        if (!isDeriveMacroInImport) return true
+    }
 
     val rustcChannel = contextMod.cargoProject?.rustcInfo?.version?.channel ?: return false
     val showUnstableItems = rustcChannel != RustChannel.STABLE && rustcChannel != RustChannel.BETA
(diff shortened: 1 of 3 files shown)


Undin (commented):
It seems we have quite poor support of name resolution/completion support for proc macro function/attributes.
Maybe it's worth refactoring the corresponding code not to introduce regressions?

Also small possible improvement:
Let's use different icon when we complete proc-macro attribute. Currently, we use function icon 
Screen Shot 2021-04-21 at 14 05 10
I suggest `org.rust.ide.icons.RsIcons#ATTRIBUTE` in case of attribute completion. Don't insist on doing it in this PR

src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -622,12 +622,15 @@ fun filterDeriveProcMacros(processor: RsResolveProcessor): RsResolveProcessor =
...
+fun RsOuterAttributeOwner.shouldHideElementInCompletion(context: RsElement, contextMod: RsMod): Boolean {
     val elementContainingMod = containingMod
     if (elementContainingMod == contextMod) return false
 
     // Hide `#[doc(hidden)]` items
-    if (queryAttributes.isDocHidden) return true
+    val isDeriveMacroInImport = context.ancestorStrict<RsUseItem>() != null
+        && this is RsFunction && isCustomDeriveProcMacroDef
+        && containingCrate != contextMod.containingCrate
+    if (queryAttributes.isDocHidden && !isDeriveMacroInImport) return true
```
vlad20012, line 633: Let's calculate the second condition more lazy:
```suggestion
    if (queryAttributes.isDocHidden) {
        val isDeriveMacroInImport = context.ancestorStrict<RsUseItem>() != null
            && this is RsFunction && isCustomDeriveProcMacroDef
            && containingCrate != contextMod.containingCrate
        if (!isDeriveMacroInImport) return true
    }
```

src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
```diff
@@ -622,12 +622,15 @@ fun filterDeriveProcMacros(processor: RsResolveProcessor): RsResolveProcessor =
...
-fun RsOuterAttributeOwner.shouldHideElementInCompletion(contextMod: RsMod): Boolean {
+fun RsOuterAttributeOwner.shouldHideElementInCompletion(context: RsElement, contextMod: RsMod): Boolean {
     val elementContainingMod = containingMod
     if (elementContainingMod == contextMod) return false
 
     // Hide `#[doc(hidden)]` items
-    if (queryAttributes.isDocHidden) return true
+    val isDeriveMacroInImport = context.ancestorStrict<RsUseItem>() != null
+        && this is RsFunction && isCustomDeriveProcMacroDef
+        && containingCrate != contextMod.containingCrate
```
vlad20012, line 632: Why do we need the crate check? :thinking: 
If it is the same crate, we should show it anyway (or now show, but it shouldn't depend on `#[doc(hidden)]` value), isn't it?

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/resolve/Processors.kt
dima74: I added crate check because don't want to change any behaviour of completing proc macros within same crate (it is currently working incorrectly - should complete function name and not derive macro name - so it should be fixed separately)

vlad20012 (approved):
bors r+

PR #7142, opened 2021-04-19
CARGO: drop `org.rust.cargo.fetch.out.dir` feature

Undin:
The feature was an initial attempt to support `include` macro with `OUT_DIR` variable.
It has two major issues:
- invocation of Cargo with `--build-plan` option forces Cargo to recompile everything from the scratch next time
- the option is supposed to be [removed](https://github.com/rust-lang/cargo/issues/7614).

Also, there is a more powerful feature - `org.rust.cargo.evaluate.build.scripts` that allows plugin to know about all generated `cfg` options and environment variables (including `OUT_DIR`).

So I think we can drop `--build-plan` support since it has major cons and can be superseded by `org.rust.cargo.evaluate.build.scripts` feature

changelog: Drop support of `org.rust.cargo.evaluate.build.scripts` experimental feature. Use `org.rust.cargo.evaluate.build.scripts` instead if you need to evaluate `env!("OUT_DIR")` in your `include!` macro calls

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt
@@ -303,12 +303,10 @@ object CargoMetadata {
 
     fun clean(
         project: Project,
-        buildMessages: BuildMessages? = null,
-        buildPlan: CargoBuildPlan? = null
+        buildMessages: BuildMessages? = null
     ): CargoWorkspaceData {
         val fs = LocalFileSystem.getInstance()
         val members = project.workspace_members
-        val variables = PackageVariables.from(buildPlan)
         val packageIdToNode = project.resolve.nodes.associateBy { it.id }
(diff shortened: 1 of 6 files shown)


vlad20012:
:+1: 
bors r+

aetelani:
Doesn't work for me. Trying to use bindgen generated file. Nightly plugins and build. Tried to clean index and restart.
```
#![allow(nonstandard_style, non_snake_case, non_camel_case_types, non_upper_case_globals,dead_code,unused_variables,unused_must_use)]
include!(concat!(env!("OUT_DIR"), "/api.rs"));
```

Undin:
@aetelani is `org.rust.cargo.evaluate.build.scripts` [experimental feature](https://plugins.jetbrains.com/plugin/8182-rust/docs/rust-faq.html#experimental-features) enabled?

If no, could you provide a whole reproducible example, please? Otherwise, it's quite hard to say why it doesn't work as expected.

PR #7146, opened 2021-04-19
CI: speed up check workflow

Undin:
These changes:
- prevent running checks before `cancel-previous` job
- ask GitHub to fail the whole `check-plugin` job if at least one step fails during merging. Implemented as check of `staging` and `trying` branches
- prevent running checks merge commits in `master` branch. We assume that they are already checked in `staging` branch

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -18,6 +18,16 @@ jobs:
               with:
                   access_token: ${{ github.token }}
 
+    calculate-git-info:
+        needs: [ cancel-previous ]
+        runs-on: ubuntu-18.04
+        outputs:
+            is_bors_branch: ${{ steps.calculate-git-info.outputs.is_bors_branch }}
+        steps:
+            - name: Calculate git info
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -83,9 +117,10 @@ jobs:
                   path: ${{ matrix.config.artifact_path }}
 
     check-plugin:
-        needs: [ build-native-code ]
+        needs: [ cancel-previous, calculate-git-info, build-native-code ]
         strategy:
-            fail-fast: false
+            # `fromJSON` is used here to convert string output to boolean
+            fail-fast: ${{ fromJSON(needs.calculate-git-info.outputs.is_bors_branch) }}
```
Undin, line 123: Unfortunately, I don't know a better way how to do this conversion

vlad20012 (approved):

Undin:
Merged by accident. Dropped from master. See new attempt in #7162

Issue #7126, opened 2021-04-16
Add detailed view for crate version at Cargo.toml


ShagonRU:
## Environment

* **IntelliJ Rust plugin version:** 0.3.146.3800-211-nightly
* **Rust toolchain version:** 1.53.0-nightly (16bf626a3 2021-04-14) x86_64-pc-windows-msvc
* **IDE name and version:** PyCharm 2021.1 Professional Edition (PY-211.6693.115)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
1) I want to see the versions available, for example the `requirements` plugin for Python, with warning about yanked versions or versions. which is fails to build (if such info is available).
2) And it would be cool if i can go by the link to rustdocs to specific version by clicking on the version string (like in `requirements` plugin).

I understand that this will most likely be possible only with `crates local index` feature, but I would still like to see it.

## Steps to reproduce
https://plugins.jetbrains.com/plugin/10837-requirements

1.
![image](https://user-images.githubusercontent.com/28485306/115084989-5c32e880-9f12-11eb-8807-7554e652cdfb.png)

2. 
![image](https://user-images.githubusercontent.com/28485306/115085275-e9763d00-9f12-11eb-8d7c-b5477c3f6854.png)

Undin:
@ShagonRU 
1. I suppose it will be implemented (at least partially) in #6801
2. You can click on the icon in the gutter of Cargo.toml file near to dependency declaration
![image](https://user-images.githubusercontent.com/2539310/115269011-8cad9900-a143-11eb-8eef-28a4f4bc01dd.png)

Kobzol:
@avrong I have prepared an inspection (based on https://github.com/intellij-rust/intellij-rust/pull/6801), which checks if there's a new crate version available, inspired by https://github.com/mibac138/intellij-rust/commit/c8960b23adeb70deac77aab3d5b6a4edbb7fa700.

I noticed that you are assigned to this issue, should I send a PR with the inspection or are you already working on it?

avrong:
@Kobzol I didn't start work on that. Feel free to open PR :)

PR #7148, opened 2021-04-19
TOML: add inspection that offers an update to a newer dependency version

Kobzol:
This PR adds a TOML inspection that uses a weak warning to notify the user that there might be a more recent version of a dependency available (inspired by https://github.com/mibac138/intellij-rust/commit/c8960b23adeb70deac77aab3d5b6a4edbb7fa700). It also provides a quick fix that replaces the version requirement with the newest crate version.

Currently, the inspection is only offered if there is a non-yanked, stable version of the dependency which is not matched by the provided version requirement. This means that if the maximum version is e.g. `3` and the user requires `4`, the inspection will trigger, which might be a bit weird, but this situation should be very rare. I didn't know how else to implement it, since I didn't find a way how to compare versions with version requirements.

Also, I'm not sure if we should disable the inspection in some situations, like when the version requirement is exact (`=1.2.3`) or when it's empty/wildcard (`*`)?

Related issue: https://github.com/intellij-rust/intellij-rust/issues/7126

changelog: Add an inspection that warns about possibly outdated dependency versions in Cargo.toml.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/inspections/NewCrateVersionAvailableInspection.kt
+++ b/toml/src/main/kotlin/org/rust/toml/inspections/NewCrateVersionAvailableInspection.kt
@@ -1,0 +1,48 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.toml.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.vdurmont.semver4j.Semver
+import org.rust.toml.crates.local.CargoRegistryCrate
+import org.rust.toml.crates.local.CrateVersionRequirement
+import org.rust.toml.stringValue
+import org.toml.lang.psi.TomlValue
+
+class NewCrateVersionAvailableInspection : CrateVersionInspection() {
+    override fun handleCrateVersion(
+        dependency: DependencyCrate,
+        crate: CargoRegistryCrate,
+        versionElement: TomlValue,
+        holder: ProblemsHolder
+    ) {
+        val versionText = versionElement.stringValue ?: return
+        val versionReq = CrateVersionRequirement.build(versionText) ?: return
+        if (versionReq.isPinned) return
+
+        val versions = crate.versions
+            .filter { !it.isYanked }
+            .mapNotNull { it.semanticVersion }
+            .sortedDescending()
+
+        val highestMatchingVersion = versions.firstOrNull { versionReq.matches(it) } ?: return
+        val newerVersion = versions.firstOrNull { it.isRustStable && it > highestMatchingVersion } ?: return
+
+        holder.registerProblem(
+            versionElement,
+            "A newer version is available for crate ${dependency.crateName}: ${newerVersion.value}",
+            ProblemHighlightType.WEAK_WARNING,
+            UpdateCrateVersionFix(versionElement, newerVersion.value)
+        )
+    }
+}
+
+/**
+ * A lot of Rust crates stay at 0.x.y for a long time, so we consider even major versions 0 to be stable.
+ */
+private val Semver.isRustStable: Boolean
+    get() = suffixTokens.isEmpty()
--- a/toml/src/test/kotlin/org/rust/toml/inspections/CargoTomlCrateInspectionTestBase.kt
+++ b/toml/src/test/kotlin/org/rust/toml/inspections/CargoTomlCrateInspectionTestBase.kt
@@ -8,23 +8,47 @@ package org.rust.toml.inspections
 import com.intellij.codeInspection.InspectionProfileEntry
 import org.intellij.lang.annotations.Language
 import org.rust.cargo.CargoConstants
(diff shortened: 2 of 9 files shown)


avrong:
I think it is okay to show inspection on empty and star requirements, as usually, they are being temporary solutions. For exact ones, let's do not touch them for now as they are being marked with equals sigh manually by a user.

What do you think?

Kobzol:
Now that I think of it, there are probably two separate inspections/quick fixes that make sense here:
1) Notify the user that there is a newer version of a dependency which **doesn't match** his version requirement
    - In this case the quick fix should update the version requirement, if possible (might not be so easy to do in a general case).
    - This is not ideal for `"*"` or `""` requirements, but it doesn't hurt to show it I guess.
2) Notify the user that there is a newer version of a dependency, which **matches** his version requirement
    - This is basically only for files with `Cargo.lock` and the corresponding quick fix should run `cargo update` in the background instead.

avrong:
I guess we still have to somehow figure out a major version of the requirement to avoid such situations:

Kobzol:
Thanks. I have reworked the algorithm to solve this case.

toml/src/main/kotlin/org/rust/toml/inspections/NewCrateVersionAvailableInspection.kt
```diff
@@ -0,0 +1,41 @@
...
+        val versionReq = CrateVersionRequirement.build(versionText) ?: return
+        if (versionReq.isPinned) return
+
+        val versions = crate.versions
+            .filter { !it.isYanked }
+            .mapNotNull { it.semanticVersion }
+            .sortedDescending()
+
+        val highestMatchingVersion = versions.firstOrNull { versionReq.matches(it) } ?: return
+        val newerVersion = versions.firstOrNull { it.isStable && it > highestMatchingVersion } ?: return
```
avrong, line 32: What if we will offer to update versions before `1.0` if there are only nightly versions available? It can be useful for crates that are non-stable for a long time

Kobzol: It's true that Rust crate authors have a certain.. "reluctancy" to go above 0.x :D But I'm not sure if offering nightly or (semver) unstable 0.xyz versions.

Do you propose to include a special case and also offer newer versions if they are 0.x.y.z (i.e. unstable by semver, but "stable" by Rust standards)?

avrong: I think it will be good to have such a special case. We have `weak warning` severity for this inspection, so it shouldn't be that annoying to the users

Kobzol: Ok, I added support for versions with major version 0.

toml/src/test/kotlin/org/rust/toml/inspections/CargoTomlCrateInspectionTestBase.kt
```diff
@@ -8,24 +8,48 @@ package org.rust.toml.inspections
...
-        myFixture.configureByText(CargoConstants.MANIFEST_FILE, code)
+    override fun createAnnotationFixture(): RsAnnotationTestFixture<Unit> =
+        RsAnnotationTestFixture(
+            this,
+            myFixture,
+            inspectionClasses = listOf(inspectionClass),
+            baseFileName = CargoConstants.MANIFEST_FILE
+        )
 
+    private fun runTest(action: () -> Unit, vararg crates: Pair<String, CargoRegistryCrate>) {
```
avrong, line 30: Such function declaration can be more concise:
```runTest(crates: List<Pair<String, CargoRegistryCrate>>, action: () -> Unit)```

It moves function type parameter to the end and removes `vararg` as it doesn't get used here

avrong (approved):
LGTM except for minor comment

toml/src/test/kotlin/org/rust/toml/inspections/CargoTomlCrateInspectionTestBase.kt
Kobzol: You're right, thanks :)

Kobzol:
bors r=avrong

lancelote:
Works great but I wonder - why IDE shows a warning for `rand = "0.7.2"` but not for `rand = "0.8.3"`? To not DDOS the user with warnings about "patch"-level version changes?

Kobzol:
Version requirement `"0.8.3"` is the same as `"^0.8.3"`, which in semver still matches version 0.8.4. Which means two things:
1) If you install the crate from scratch (without a lockfile), 0.8.4 will get installed, so there's probably no need to upgrade it.
2) Even if you change the requirement to `"0.8.4"`, if you have a lockfile, the dependency might not change? (I can't test it right now)

This comes down to something that I have discussed in a comment above. Maybe we actually want two actions - one to update the lockfile to the newest version that still matches the specified version requirement, and one that actually bumps the version requirement (that is what this action does).

PR #7150, opened 2021-04-20
Update DTrace profiler integration to CLion 2021.1

ortem:
Fixes https://youtrack.jetbrains.com/issue/CPP-24532
The bundled `osx-starter` lib is located in `bin/profiler/mac` directory [since](https://youtrack.jetbrains.com/issue/CPP-23530) CLion 211.6693.66

changelog: Fix `Can't find process starter` error when running DTrace profiler with CLion 2021.1

### DIFF
--- a/clion/src/main/kotlin/org/rust/clion/profiler/dtrace/RsDTraceConfigurationExtension.kt
+++ b/clion/src/main/kotlin/org/rust/clion/profiler/dtrace/RsDTraceConfigurationExtension.kt
@@ -14,9 +14,11 @@ import com.intellij.execution.process.OSProcessUtil
 import com.intellij.execution.process.ProcessHandler
 import com.intellij.execution.process.UnixProcessManager
 import com.intellij.execution.runners.ExecutionEnvironment
+import com.intellij.openapi.application.ApplicationInfo
 import com.intellij.openapi.application.PathManager
 import com.intellij.openapi.progress.PerformInBackgroundOption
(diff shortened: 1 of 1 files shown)


clion/src/main/kotlin/org/rust/clion/profiler/dtrace/RsDTraceConfigurationExtension.kt
```diff
@@ -82,13 +84,14 @@ class RsDTraceConfigurationExtension : CargoCommandConfigurationExtension() {
...
-            }
-            return File(PathManager.getLibPath(), BUNDLED_STARTER_PATH) //production
+
+        // BACKCOMPAT: 2020.3
+        private val BUILD_211: BuildNumber = BuildNumber.fromString("211")!!
+
+        private fun profilerStarterPath(): File = if (ApplicationInfo.getInstance().build >= BUILD_211) {
+            File(PathManager.getBinPath(), "profiler/mac/libosx-starter.dylib")
+        } else {
+            File(PathManager.getLibPath(), "profiler/macosx/libosx-starter.dylib")
```
Undin, line 94: It would be great not to handle it by ourselves and use CLion method not to worry about it in the future.
Let's made the corresponding change in CLion codebase. Also, it can be done together with dropping `internal` modifier from `validateDTraceSettings` function

Undin (approved):
bors r+

Issue #7153, opened 2021-04-21
Unexpected navigation bar items in Rust files


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.3.147.3812-211-nightly
* **Rust toolchain version:** 1.51.0 (2fd73fabe 2021-03-23) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2021.1 Ultimate Edition (IU-211.6693.111)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description
<img width="562" alt="Screen Shot 2021-04-21 at 19 50 10" src="https://user-images.githubusercontent.com/2539310/115591485-d086d680-a2da-11eb-96d6-227f5ece30da.png">

<img width="580" alt="Screen Shot 2021-04-21 at 19 46 22" src="https://user-images.githubusercontent.com/2539310/115591106-6e2dd600-a2da-11eb-9d31-9ada5775692d.png">

Introduced in #6581

## Steps to reproduce
```rust
struct C {
    a: i32
}

mod bar;

fn main() {}
```

Kobzol:
Oops. Since I reused the breadcrumbs implementation, I kind of assumed that its tests would cover all needed situations. But it is obviously not the case. I'll send a PR with tests right away.

Kobzol:
Hmm, when `StructureAwareNavBarModelExtension::getPresentableText` returns null, it renders it with `toString`. That is unfortunate :disappointed: Should I add support for the missing cases (things like named field decl) only to the navbar or also to the breadcrumbs? Or do we actually want to disable the breadcrumbs in favor of the navbar, as is done e.g. by Kotlin?

PR #7157, opened 2021-04-21
IDE: Add support for navbar breadcrumbs

Kobzol:
This PR adds tests for the new navbar navigation (#6581), and enables it by default.

While writing the tests, I realized that when `StructureAwareNavBarModelExtension::getPresentableText` returns null, the element is rendered with `toString`, which is less than ideal.

I tried to find the missing cases not handled by the breadcrumbs provider, and added them to `RsBreadcrumbsInfoProvider::RsNamedHandler`. I'm not sure if that's the ideal solution though. Maybe I should only add them to the navbar model? Or somehow unify the navbar model with the breadcrumbs provider?

There is also some issue in tests, when a macro 2.0 or an extern crate is rendered by the navbar, the rendered path does not contain the name of the containing file for some reason.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7153

changelog: Introduce navigation bar breadcrumbs

### DIFF
--- a/src/main/kotlin/org/rust/ide/miscExtensions/RsBreadcrumbsInfoProvider.kt
+++ b/src/main/kotlin/org/rust/ide/miscExtensions/RsBreadcrumbsInfoProvider.kt
@@ -38,7 +38,8 @@ class RsBreadcrumbsInfoProvider : BreadcrumbsProvider {
     private object RsNamedHandler : RsElementHandler<RsNamedElement> {
         override fun accepts(e: PsiElement): Boolean =
             e is RsModItem || e is RsStructOrEnumItemElement || e is RsTraitItem
-                || e is RsConstant || e is RsTypeAlias
+            || e is RsConstant || e is RsTypeAlias || e is RsNamedFieldDecl
+            || e is RsEnumVariant || e is RsModDeclItem || e is RsExternCrateItem
 
         override fun elementInfo(e: RsNamedElement): String = e.name.let { "$it" }
     }
@@ -75,10 +76,10 @@ class RsBreadcrumbsInfoProvider : BreadcrumbsProvider {
         }
     }
 
-    private object RsMacroHandler : RsElementHandler<RsMacro> {
-        override fun accepts(e: PsiElement): Boolean = e is RsMacro
+    private object RsMacroHandler : RsElementHandler<RsMacroDefinitionBase> {
+        override fun accepts(e: PsiElement): Boolean = e is RsMacroDefinitionBase
 
-        override fun elementInfo(e: RsMacro): String = e.name.let { "$it!" }
+        override fun elementInfo(e: RsMacroDefinitionBase): String = e.name.let { "$it!" }
     }
 
     private object RsFunctionHandler : RsElementHandler<RsFunction> {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/ide/miscExtensions/RsBreadcrumbsInfoProvider.kt
```diff
@@ -78,10 +76,10 @@ class RsBreadcrumbsInfoProvider : BreadcrumbsProvider {
         }
     }
 
-    private object RsMacroHandler : RsElementHandler<RsMacro> {
-        override fun accepts(e: PsiElement): Boolean = e is RsMacro
+    private object RsMacroHandler : RsElementHandler<RsMacroDefinitionBase> {
+        override fun accepts(e: PsiElement): Boolean = e is RsMacro || e is RsMacro2
```
dima74, line 80:
```suggestion
        override fun accepts(e: PsiElement): Boolean = e is RsMacroDefinitionBase
```

src/main/kotlin/org/rust/ide/miscExtensions/RsBreadcrumbsInfoProvider.kt
```diff
@@ -78,10 +76,10 @@ class RsBreadcrumbsInfoProvider : BreadcrumbsProvider {
...
 
-        override fun elementInfo(e: RsMacro): String = e.name.let { "$it!" }
+        override fun elementInfo(e: RsMacroDefinitionBase): String = (e as RsNamedElement).name.let { "$it!" }
```
dima74, line 82:
```suggestion
        override fun elementInfo(e: RsMacroDefinitionBase): String = e.name.let { "$it!" }
```

src/main/kotlin/org/rust/ide/miscExtensions/RsBreadcrumbsInfoProvider.kt
```diff
@@ -218,6 +216,8 @@ class RsBreadcrumbsInfoProvider : BreadcrumbsProvider {
 
     override fun getElementTooltip(e: PsiElement): String? = null
 
+    fun getBreadcrumb(e: PsiElement): String? = handler(e)?.elementInfo(e as RsElement)
```
dima74, line 219:
```suggestion
    fun getBreadcrumb(e: RsElement): String? = handler(e)?.elementInfo(e)
```

src/main/kotlin/org/rust/ide/navigationToolbar/RsNavBarModelExtension.kt
```diff
@@ -22,8 +22,9 @@ class RsNavBarModelExtension : StructureAwareNavBarModelExtension() {
         }
 
         val provider = RsBreadcrumbsInfoProvider()
-        if (provider.acceptElement(element)) {
-            return provider.getElementInfo(element)
+        val breadcrumb = provider.getBreadcrumb(element)
```
dima74, line 25:
```suggestion
        return provider.getBreadcrumb(element)
```

dima74 (commented):
> There is also some issue in tests, when a macro 2.0 or an extern crate is rendered by the navbar, the rendered path does not contain the name of the containing file for some reason.

Could you please rebase on master, probably tests will be fixed

Kobzol:
Rebased. (Sorry for the delay, I was on a vacation.)

dima74 (approved):
bors r+

dima74:
It will be enabled by this PR since we uncomment `<navbar implementation ...`, isn't it?

Kobzol:
Oh, sorry, I didn't notice it, it was a long time ago :sweat_smile:

PR #7160, opened 2021-04-22
T: write proc macro tests without heavy `RsWithToolchainTestBase`

vlad20012:
I made a crate `testData/test-proc-macros` with all proc macros needed in our tests and a special test project descriptor `WithProcMacroRustProjectDescriptor` that adds this crate as a dependency to test crates. So it's now possible to use proc macros in our tests everywhere without the requirement to extend `RsWithToolchainTestBase` (which is slow and unhandy).


src/test/kotlin/org/rust/RsTestBase.kt
```diff
@@ -140,6 +141,12 @@ abstract class RsTestBase : BasePlatformTestCase(), RsTestCase {
         testCargoProjects.setCargoFeatures(packageFeatures, testRootDisposable)
     }
 
+    private fun setupExperimentalFeatures() {
```
Kobzol, line 144: I was just looking for something like this in TOML tests that need the new local crate index! Great.

PR #7162, opened 2021-04-22
CI: speed up check workflow

Undin:
These changes:
- prevent running checks before `cancel-previous` job
- ask GitHub to fail the whole `check-plugin` job if at least one step fails during merging. Implemented as check of `staging` and `trying` branches
- run checks in `master` branch only once, i.e. don't run `check` workflow for changes merged by bors because they are already checked in `staging` branch


lancelote:
bors merge

PR #7165, opened 2021-04-23
ANN: Add quick fix to correct invalid cfg predicates

Kobzol:
Improves https://github.com/intellij-rust/intellij-rust/pull/7102 by adding a quick fix to correct invalid cfg name.

Note that I couldn't just use `RenameFix`, because this fix has to work on `RsPath` in the case of `cfg` predicates. And in this way the fix is general and it can be used in more places.

changelog: Add quick fix to correct the name of an invalid cfg predicate.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
@@ -1,0 +1,44 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixOnPsiElement
+import com.intellij.util.text.EditDistance
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+
+/**
+ * Changes the text of some element to the suggested name using the provided function.
+ */
+class NameSuggestionFix<T : PsiElement>(
+    element: T,
+    private val newName: String,
+    private val elementFactory: (name: String) -> T
+): LocalQuickFixOnPsiElement(element) {
+    override fun getFamilyName(): String = "Change name of element"
+    override fun getText(): String = "Change to `$newName`"
+
+    override fun invoke(project: Project, file: PsiFile, startElement: PsiElement, endElement: PsiElement) {
+        val newElement = elementFactory(newName)
+        startElement.replace(newElement)
+    }
+
+    companion object {
+        fun <T : PsiElement> createApplicable(
+            element: T,
+            name: String,
+            validNames: List<String>,
+            maxDistance: Int,
+            elementFactory: (name: String) -> T,
+        ): List<NameSuggestionFix<T>> {
+            val applicableNames = validNames.filter {
+                name != it && EditDistance.levenshtein(it, name, true) <= maxDistance
+            }
+            return applicableNames.map { NameSuggestionFix(element, it, elementFactory) }
+        }
+    }
+}
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -135,7 +135,13 @@ class RsErrorAnnotator : AnnotatorBase(), HighlightRangeExtension {
             "version" -> { /* version is currently experimental */ }
             else -> {
                 val path = item.path ?: return
-                RsDiagnostic.UnknownCfgPredicate(path, itemName).addToHolder(holder, checkExistsAfterExpansion = false)
+                val fixes = NameSuggestionFix.createApplicable(path, itemName, listOf("all", "any", "not"), 1) { name ->
+                    RsPsiFactory(path.project).tryCreatePath(name) ?: error("Cannot create path out of $name")
+                }
+                RsDiagnostic.UnknownCfgPredicate(path, itemName, fixes).addToHolder(
+                    holder,
+                    checkExistsAfterExpansion = false
+                )
             }
         }
     }
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -1376,11 +1376,16 @@ sealed class RsDiagnostic(
         )
     }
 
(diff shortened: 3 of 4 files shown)


src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.intellij.psi.PsiFile
+import org.rust.lang.core.psi.RsPath
+
+/**
+ * Changes the text of some element to the suggested name using the provided function.
+ */
+class NameSuggestionFix(
+    element: PsiElement,
+    private val newName: String,
+    private val replaceElementFn: (name: String) -> PsiElement?
```
Undin, line 21: Let's rename it somehow because it actually doesn't replace element, only creates a new one
For example, `elementSupplier` or `elementFactory`

Undin, line 21: Also, I cannot come up with a situation when `null` result is expected value from. So let's drop `?`

Undin, line 21:
```suggestion
class NameSuggestionFix<T : PseElement>(
    element: T,
    private val newName: String,
    private val replaceElementFn: (name: String) -> T
```

Let's try to avoid inconsistency between element type and factory method. Of course, you always can use pure `PsiElement` but anyway

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    companion object {
+        fun createApplicable(
+            element: RsPath,
+            name: String,
+            validNames: List<String>,
+            replaceElementFn: (name: String) -> PsiElement?,
+            maxDistance: Int = 1
+        ): List<NameSuggestionFix> {
```
Undin, line 38: Let's also try to generify this method to have ability to pass not only `RsPath `

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,45 @@
...
+        startElement.replace(newElement)
+    }
+
+    companion object {
+        fun createApplicable(
+            element: RsPath,
+            name: String,
+            validNames: List<String>,
+            replaceElementFn: (name: String) -> PsiElement?,
+            maxDistance: Int = 1
```
Undin, line 37: If you swap `replaceElementFn` and `maxDistance` with each other, you can pass lambda directly outside of the parenthesis on call site. I think it's a little bit more idiomatic

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,45 @@
...
+        val newElement = replaceElementFn(newName) ?: return
+        startElement.replace(newElement)
+    }
+
+    companion object {
+        fun createApplicable(
+            element: RsPath,
+            name: String,
+            validNames: List<String>,
+            replaceElementFn: (name: String) -> PsiElement?,
```
Undin, line 36: The same about name

Undin (commented):
LGTM except minor comments

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
Kobzol: That's true I guess. Can I afford `!!` here in that case?
```kotlin
val nameFactoryFn = { name: String ->
    RsPsiFactory(path.project).tryCreatePath(name)!!
}
```

Undin: I usually prefer `?: error("some usefull message")`

Kobzol:
Rebased & fixed comments.

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,44 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import com.intellij.codeInspection.LocalQuickFixOnPsiElement
+import com.intellij.grazie.speller.utils.Distances.levenshtein
```
Undin, line 9: Oh, I think we shouldn't use this implementation because it's from `grazie-spellchecker` lib and I doubt that we can rely on it because it can be changed, doesn't exist in all IDEs or even excluded from IDEA/CLion libraries in the future. Let's better use `com.intellij.util.text.EditDistance#levenshtein` that a part of common API

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+
+/**
+ * Changes the text of some element to the suggested name using the provided function.
+ */
+class NameSuggestionFix<T: PsiElement>(
```
Undin, line 17:
```suggestion
class NameSuggestionFix<T : PsiElement>(
```

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun getFamilyName(): String = "Change name of element"
+    override fun getText(): String = "Change to `$newName`"
+
+    override fun invoke(project: Project, file: PsiFile, startElement: PsiElement, endElement: PsiElement) {
+        val newElement = elementFactory(newName)
+        startElement.replace(newElement)
+    }
+
+    companion object {
+        fun <T: PsiElement>createApplicable(
```
Undin, line 31:
```suggestion
        fun <T : PsiElement> createApplicable(
```

Undin (approved):
Thank you! 👍

Undin:
And could you add `changelog` section to PR description, please?

src/main/kotlin/org/rust/ide/annotator/fixes/NameSuggestionFix.kt
Kobzol: Oops, wrong autoimport, thanks.

Kobzol:
bors r=Undin

PR #7168, opened 2021-04-23
RES2: Fix resolve crate when importing private module with same name

### LINKED ISSUE #7124: IntelliJ Rust falsely reports that the error_chain module is private but it compiles and works
## Environment

* **IntelliJ Rust plugin version:** 0.3.145.3797-211
* **Rust toolchain version:** null
* **IDE name and version:** IntelliJ IDEA 2021.1 Community Edition (IC-211.6693.111)
* **Operating system:** macOS 10.14.6

## Problem description
So as the title says, the error_chain module is reported that it is private, but it's not true, it can be compiled and most importantly it works.

## Steps to reproduce
Import error chain like this:
```rust
use error_chain::error_chain;
```

image

dima74:
Fixes #7124

changelog: Fix false-positive [E0603](https://doc.rust-lang.org/error-index.html#E0603) "Module is private" when using new name resolution engine


vlad20012:
:+1: 
bors r+

PR #7174, opened 2021-04-25
Cancel external linter (cargo check) on PSI any change

### LINKED ISSUE #4283: Intellij-rust should stop external linter before running
Now if project is big enough and complex, external linter works not that fast.
AFAIU, cargo takes lock on directory when running `cargo check`, so when I run application, I have to wait until `cargo check` completes.
image

## Environment

* **IntelliJ Rust plugin version:** 0.2.103.2131-192
* **Rust toolchain version:** 1.39.0-nightly (29a54035c 2019-08-19) x86_64-apple-darwin
* **IDE name and version:** CL-192.5728.100
* **Operating system:** macOS 10.14.5

vlad20012:
Fixes #4283

I think we got issues with Clippy (https://github.com/rust-lang/rust-clippy/issues/3890) due to a simple reason: sometimes we miss the file descriptor to running Cargo/Clippy process, hence we never kill it and don't read its output. The result is that such missed process hangs on writing to stdout because nobody actually read its output.

The fix is that, well, now we don't miss the process. Also, now we use SIGKILL instead of SIGTERM.

changelog: Cancel external linter (cargo check) immediately when some file is changed

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/CommandLineExt.kt
+++ b/src/main/kotlin/org/rust/openapiext/CommandLineExt.kt
@@ -51,7 +51,8 @@ fun GeneralCommandLine.execute(
     listener: ProcessListener? = null
 ): ProcessOutput {
 
-    val handler = CapturingProcessHandler(this)
+    val handler = CapturingProcessHandler(this) // The OS process is started here
+
     val cargoKiller = Disposable {
         // Don't attempt a graceful termination, Cargo can be SIGKILLed safely.
         // https://github.com/rust-lang/cargo/issues/3566
@@ -68,6 +69,8 @@ fun GeneralCommandLine.execute(
     }
 
     if (alreadyDisposed) {
+        Disposer.dispose(cargoKiller) // Kill the process
+
         // On the one hand, this seems fishy,
         // on the other hand, this is isomorphic
         // to the scenario where cargoKiller triggers.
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/openapiext/CommandLineExt.kt
```diff
@@ -55,7 +55,10 @@ fun GeneralCommandLine.execute(
     val cargoKiller = Disposable {
         // Don't attempt a graceful termination, Cargo can be SIGKILLed safely.
         // https://github.com/rust-lang/cargo/issues/3566
-        handler.destroyProcess()
+        if (!handler.isProcessTerminated) {
+            handler.process.destroyForcibly() // Send SIGKILL
+            handler.destroyProcess()
```
vlad20012, line 60: Btw I'm not sure we need this line at all. Can it be useful if `listener` is present?

mchernyavsky (commented):
It doesn't work for me :(

Screenshot 2021-04-29 at 16 20 43

vlad20012:
The initial considerations were wrong. I've updated the PR description

mchernyavsky (approved):
bors r+

matklad:
> hangs on writing to stdout 

interesting, it seems like this means that there are two bugs here:

* *First*, we don’t sigkill the pid of child process 
* *Second*, we don’t close fd corresponding to child’s stdout


If we close fd, the child won’t hang and die with sigpipe

vlad20012:
@matklad I meant that we totally forget the process and its descriptors, hence we don't kill or close anything.

Issue #7158, opened 2021-04-21
Missing methods introspection for fltk-rs


varomix:
## Environment

* **IntelliJ Rust plugin version:** 0.3.145.3797-211
* **Rust toolchain version:** 1.51.0 (2fd73fabe 2021-03-23) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2021.1 (CL-211.6693.114)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
Hi
I've been using fltk-rs and seems like I'm missing a lot of code completion for that library, I was checking the same with VSCode and in that editor, there's much more information and completion about the functions on the library, I love CLion and I really want to keep using it, would love to have this help since is a huge help specially when you are learning/exploring a new library.

## Steps to reproduce
Code I am using
```toml
[dependencies]
fltk = "1.0.1"
```
```rust
use fltk::{prelude::*, enums::*, *};

fn main() {
    let app = app::App::default();
    let mut win = window::Window::new(100,100, 300, 300, "My window");
    let but = button::Button::new(0,0,200,20,"Push me");
    win.end();
    win.show();

    app.run().unwrap();
```
If you mouse over the **new** function of the **Button**, Clion/Rust show nothing and VSCode has all the information for it.
![image](https://user-images.githubusercontent.com/728059/115616556-8fc49880-a2ad-11eb-8b1d-a09f8cbe1758.png)

Also if you type **but.** VSCode shows all the methods that Button has, while CLion shows generic stuff
![image](https://user-images.githubusercontent.com/728059/115616680-b5ea3880-a2ad-11eb-9fdd-3245090b6c4c.png)
![image](https://user-images.githubusercontent.com/728059/115616743-c8fd0880-a2ad-11eb-9e5f-ebe5b65fa726.png)

Thank you

varomix:
I enabled every feature I found on the settings

dima74:
Root cause is that derive proc macro for `WidgetBase` is unresolved because we don't currently support `#[macro_use] extern crate ...` for proc macros

varomix:
I guess is something is planned to support later, right? 

thanks

dima74:
> I guess is something is planned to support later, right?

Yes, sure. As far as I know @vlad20012 has some initial implementation of `#[macro_use] extern crate ...` support for proc macros

PR #7184, opened 2021-04-27
RES: resolve proc macros through `#[macro_use]`

vlad20012:


src/test/kotlin/org/rust/common.kt
```diff
@@ -119,6 +125,17 @@ private fun checkSearchScope(referenceElement: PsiElement, resolvedTo: PsiElemen
     }
 }
 
+private fun checkProcMacroResolve(reference: PsiReference, resolved: @Nullable PsiElement) {
```
dima74, line 128:
```suggestion
private fun checkProcMacroResolve(reference: PsiReference, resolved: PsiElement) {
```

src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
```diff
@@ -254,6 +261,13 @@ fun RsMetaItem.resolveToProcMacroWithoutPsi(): ProcMacroDefInfo? {
         is RsModInfo -> {
             val (_, defMap, modData) = info
             val macroPath = path?.pathSegmentsAdjustedForAttrMacro?.toTypedArray() ?: return null
+            if (macroPath.size == 1) {
```
dima74, line 264: Lets instead pass `ResolveMode.IMPORT` to `defMap.resolvePathFp`, so that it will not ignore proc macros in legacy scope

src/test/kotlin/org/rust/lang/core/resolve/RsProcMacroResolveTest.kt
```diff
@@ -259,4 +270,160 @@ class RsProcMacroResolveTest : RsResolveTestBase() {
...
+            #[macro_use]
+            extern crate dep_proc_macro;
+
+            use ProcMacroName as ProcMacroNameAlias;
+
+            #[derive(ProcMacroNameAlias)]
+                    //^ dep-proc-macro/lib.rs
+            struct S;
+    """)
 }
```
dima74, line 429: Let's add a test when `dep-proc-macro/lib.rs` has macro with `#[macro_export]` with same name as proc macro

dima74:
This also fixes #7158 👍

src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
vlad20012: They're not equivalent. Here we resolve the last `ProcMacroDefInfo`, but `resolvePathFp` resolves the first (any) macro in the scope

Undin:
@vlad20012 I suppose we want to mention it in changelog, don't I? If you agree with me, add the corresponding label and `changelog` section in the description, please

PR #7187, opened 2021-04-28
Introduce `existsAfterExpansion`, use it instead of `isEnabledByCfg`

vlad20012:
At the moment, `existsAfterExpansion` is equivalent to `isEnabledByCfg`,
so this commit is just a rename. In the future, `existsAfterExpansion`
will return `false` for elements under attribute procedural macros

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
+++ b/src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
@@ -34,8 +34,8 @@ import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.presentation.shortPresentableText
 import org.rust.ide.refactoring.implementMembers.ImplementMembersFix
 import org.rust.ide.utils.checkMatch.Pattern
+import org.rust.ide.utils.existsAfterExpansion
 import org.rust.ide.utils.import.RsImportHelper.getTypeReferencesInfoFromTys
-import org.rust.ide.utils.isEnabledByCfg
 import org.rust.lang.core.CONST_GENERICS
 import org.rust.lang.core.CompilerFeature
 import org.rust.lang.core.MIN_CONST_GENERICS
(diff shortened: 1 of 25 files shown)


src/main/kotlin/org/rust/ide/utils/CfgUtils.kt
```diff
@@ -14,6 +14,20 @@ import org.rust.lang.utils.evaluation.ThreeValuedLogic
...
+ * Returns `true` if it [isEnabledByCfg] and not inside an element under attribute procedural macro.
+ *
+ * A one exception is that it returns `true` for attribute macro itself:
+ *
+ * ```
+ * #[a_macro]  // `true` for the attribute
+ * fn foo() {} // `false` for the function
+ * ```
+ */
+val PsiElement.existsAfterExpansion: Boolean get() = isEnabledByCfg
```
Undin, line 29: Should we also introduce `fun PsiElement. existsAfterExpansion(crate: Crate): Boolean`?

src/main/kotlin/org/rust/ide/presentation/Utils.kt
```diff
@@ -58,7 +58,7 @@ fun getPresentationForStructure(psi: RsElement): ItemPresentation {
     }
     val icon = psi.getIcon(Iconable.ICON_FLAG_VISIBILITY)
     val textAttributes = when {
-        psi is RsDocAndAttributeOwner && !psi.isEnabledByCfgSelf -> RsColor.CFG_DISABLED_CODE.textAttributesKey
+        psi is RsDocAndAttributeOwner && !psi.existsAfterExpansionSelf -> RsColor.CFG_DISABLED_CODE.textAttributesKey
```
Undin, line 61: Not sure that we should use `existsAfterExpansionSelf` here

src/main/kotlin/org/rust/lang/utils/RsDiagnostic.kt
```diff
@@ -1414,7 +1414,7 @@ class PreparedAnnotation(
 )
 
 fun RsDiagnostic.addToHolder(holder: RsAnnotationHolder, checkCfg: Boolean = true) {
```
Undin, line 1416: What about `checkCfg` flag name?

src/main/kotlin/org/rust/ide/utils/CfgUtils.kt
Undin, line 29: Are you sure that it should be located in `CfgUtils`?

vlad20012: Not needed for now

src/main/kotlin/org/rust/ide/presentation/Utils.kt
vlad20012: Ok, I'm not sure how items under attr proc macros should be shown in structure view, but it's sounds reasonable that we should not highlight them as `CFG_DISABLED_CODE`

PR #7189, opened 2021-04-29
Use single stdlib

vlad20012:
Intellij-Rust plugin currently supports only one stdlib at a time. If there are different standard libraries in
different cargo projects, select and use the most recent of them.
A cargo project may have a different stdlib if there is a `rust-toolchain.toml` file,
or if it uses `rustup override`


Undin (approved):
Looks ok but there is a merge conflict

vlad20012:
bors r=Undin

PR #7192, opened 2021-04-30
Initial version of state collectors

Undin:
Provides statistic state collectors to gather info about user toolchains and some general project characteristics.
Note, the corresponding statistics are collected only if `Send usage statistics` option in `Preferences | Appearance & Behavior | System Settings | Data Sharing` settings is enabled. See more about usage statistics politic [here](https://www.jetbrains.com/help/idea/settings-usage-statistics.html)


lancelote:
bors merge

Issue #6908, opened 2021-03-01
Procedural macro support


Undin:
Tracking issue about procedural macro support.

Procedural macro support feature depends on the following experimental features:
- `org.rust.cargo.evaluate.build.scripts` - enables building and collecting build artefacts including proc-macro libraries during importing of project structure
- `org.rust.macros.proc` -  enables expansion of procedural macros

To enable an experimental feature, type `Experimental feature` in the dialog of `Help | Find Action` action and enabled the corresponding item. 

- [x] introduce proc macro server to evaluate procedural macro functions #6658
- [x] compile only "host" code of project: build scripts and proc-macro targets when `org.rust.cargo.evaluate.build.scripts` experimental feature is enabled #6735
- [x] expand [function-like](https://doc.rust-lang.org/reference/procedural-macros.html#function-like-procedural-macros) macros #6564
    Fixes #4707
    Fixes #6367
- [ ] expand [custom derive](https://doc.rust-lang.org/reference/procedural-macros.html#derive-macros) macros
    Fixes #1786
    Fixes #5468
    Fixes #5478
    Fixes #5728
    Fixes #5954
    Fixes #6036
    Fixes #6489
    Fixes #6880
- [ ] expand [attribute](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) macros
    Fixes #5239
    Fixes #5748
    Fixes #5896
    Fixes #5975 (requires changes in the corresponding inspection)
    Fixes #6091
    Fixes #6162 (requires changes in the annotator)
    Fixes #6482
- [ ] enable `org.rust.cargo.evaluate.build.scripts` experimental feature by default
- [ ] enable `org.rust.macros.proc` experimental feature by default

webern:
Does this mean that a subset of derive macro code is currently considered (e.g. for code completion and missing-symbol highlighting?

> take into account all items generated by custom derive macros

For example, the Snafu derive macro produces structs which intellij-rust seems to be unaware of.

Would a minimal use case example of this and tracking issue be of any benefit?

Undin:
@webern 
> Does this mean that a subset of derive macro code is currently considered (e.g. for code completion and missing-symbol highlighting?

Yep. Currently, the plugin takes into account only impl blocks generated by derive proc macro

> Would a minimal use case example of this and tracking issue be of any benefit?

[...]

PR #7194, opened 2021-04-30
MACRO: expand attribute proc macros

vlad20012:
Issue #6908

Despite the fact that this is a draft, it is useable. 

What is done:
- We can distinguish procedural macro attributes from other attributes (including tool attributes, `#![register_attr()]` and `#![register_tool()]`)
- An item under procedural macro attribute is considered `!existsAfterExpansion` (see #7187), so it feels like a cfg-disabled code (or like a regular function-like macro call body)
- When a procedural macro is expanded, IDE analysis works with expanded elements (instead of original elements under a macro attribute)
- There is a concept of hardcoded "not a macro" macros (better names are welcome). The idea is that there are procedural macros that feel like regular attributes and should be simply ignored by the IDE. For example, `tokio::main` and `tokio::test`. We want to allow proc macro authors to specify  "not a macro" status in [`Cargo.toml` metadata](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table)
- Navigation takes into account range mapping
- `Find usages` works in items under procedural macro attributes

TODO (maybe in next PRs):
- [x] Highlighting (`RsMacroExpansionHighlightingPass`)
- [x] Fix `existsAfterExpansion` vs `isEnabledByCfg` - an element can be wrongly considered cfg-disabled when it is a part of procedural macro and vise versa
- [x] Support attribute macros on declarative macro definitions and regular macro calls
- [x] Clean doPrepareCustomDeriveMacroCallBody/doPrepareAttributeProcMacroCallBody
- [x] Support "Expand recursively" action for attribute macros
- [ ] Correctly resolve references inside attribute macros
- [x] Support range mapping for an attribute value
- [ ] (!) Store `CustomAttributes` in defMap, pass it to getProcMacroAttributeRaw
- [x] Fill in the hardcoded macros list
- [ ] Allow specifying not-a-macro status in cargo metadata
- [x] Measure how item text storage impacts memory footprint
- [x] Measure how heavy existsAfterExpansion and ProcMacroAttribute.getProcMacroAttribute implementations impact the performance
- [x] Check `$crate` & hygiene
- [x] Investigate why `find usages` doesn't work in `web-sys` crate
- [x] "not-a-macro" works only for some kinds of items (functions, structs, etc). Doublecheck that `ProcMacroAttribute.getProcMacroAttribute` takes this fact into account
- [ ] Add test for proc macro body preparation (undecoration)
- [ ] Show generated items in the structure view
- [x] Check completion in proc macros

changelog: Provide initial support for [attribute](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) procedural macros. Now the plugin can expand such procedural macro calls and take into account all expanded items in type inference and name resolution. Note, it’s only an initial implementation, so it may work in an unexpected way in some cases. The feature is disabled by default for now. To turn it on, you should enable `Use experimental name resolution engine` option in `Preferences | Languages & Frameworks | Rust` settings and enable `org.rust.cargo.evaluate.build.scripts` and `org.rust.macros.proc` experimental features. Don’t forget to reload a project structure after enabling the corresponding features via `Refresh Cargo Projects` action on Cargo tool window. See [tracking issue](https://github.com/intellij-rust/intellij-rust/issues/6908) for more details and the current status of procedural macros support

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
@@ -21,9 +21,11 @@ import com.intellij.openapi.util.TextRange
 import com.intellij.openapi.util.registry.Registry
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.util.PsiTreeUtil
 import org.rust.lang.core.macros.*
 import org.rust.lang.core.psi.RsMacroCall
-import org.rust.lang.core.psi.ext.descendantsOfType
+import org.rust.lang.core.psi.ext.RsAttrProcMacroOwner
+import org.rust.lang.core.psi.ext.RsPossibleMacroCall
 import org.rust.lang.core.psi.ext.existsAfterExpansion
 import org.rust.lang.core.psi.ext.expansion
 import org.rust.stdext.removeLast
@@ -74,10 +76,21 @@ class RsMacroExpansionHighlightingPass(
     @Suppress("UnstableApiUsage")
     override fun doCollectInformation(progress: ProgressIndicator) {
         val macros = mutableListOf<PreparedMacroCall>()
-        for (macro in file.descendantsOfType<RsMacroCall>()) {
-            if (macro.macroArgument?.textRange?.intersects(restrictedRange) != true) continue
-            macros += macro.prepare() ?: continue
+
+        PsiTreeUtil.processElements(file) {
+            when (it) {
+                is RsMacroCall -> {
+                    if (it.macroArgument?.textRange?.intersects(restrictedRange) != true) return@processElements true
+                    macros += it.prepare() ?: return@processElements true
+                }
+                is RsAttrProcMacroOwner -> {
+                    if (it.textRange?.intersects(restrictedRange) != true) return@processElements true
+                    macros += it.procMacroAttribute.attr?.prepare() ?: return@processElements true
+                }
+            }
+            true // Continue
         }
+
         if (macros.isEmpty()) return
 
         val annotators = createAnnotators()
@@ -90,8 +103,11 @@ class RsMacroExpansionHighlightingPass(
                     ProgressManager.checkCanceled()
                     holder.runAnnotatorWithContext(element, ann)
                 }
+
                 if (element is RsMacroCall) {
                     macros += element.prepare() ?: continue
+                } else if (element is RsAttrProcMacroOwner) {
+                    macros += element.procMacroAttribute.attr?.prepare() ?: continue
                 }
             }
 
@@ -123,14 +139,14 @@ class RsMacroExpansionHighlightingPass(
     }
 }
 
-private fun RsMacroCall.prepare(): PreparedMacroCall? {
-    if (macroArgument == null) return null // special macros should not be highlighted
+private fun RsPossibleMacroCall.prepare(): PreparedMacroCall? {
+    if (this is RsMacroCall && macroArgument == null) return null // special macros should not be highlighted
     if (!existsAfterExpansion) return null
     val expansion = expansion ?: return null
     return PreparedMacroCall(this, expansion)
 }
 
-private data class PreparedMacroCall(val macroCall: RsMacroCall, val expansion: MacroExpansion) {
+private data class PreparedMacroCall(val macroCall: RsPossibleMacroCall, val expansion: MacroExpansion) {
     val elementsForHighlighting: List<PsiElement>
         get() {
             if (expansion.ranges.isEmpty()) return emptyList()
--- a/src/test/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPassTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPassTest.kt
@@ -6,9 +6,10 @@
 package org.rust.ide.annotator
 
 import org.intellij.lang.annotations.Language
-import org.rust.ExpandMacros
-import org.rust.MockAdditionalCfgOptions
+import org.rust.*
+import org.rust.cargo.project.workspace.CargoWorkspace
 import org.rust.ide.colors.RsColor
+import org.rust.ide.experiments.RsExperiments
 
 // More tests are located in `RsHighlightingAnnotatorTest` (most of those tests are executed
 // in both plain and macro context)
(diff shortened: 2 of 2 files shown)


matklad:
cc @jonas-schievink :0)

jmg-duarte:
The "not a macro" could be renamed as "identity" macro, given they technically return the same input they're passed

src/main/kotlin/org/rust/lang/core/resolve2/DefCollector.kt
```diff
@@ -469,20 +470,22 @@ class ProcMacroDefInfo(
...
-    val body: String,
+    val body: MacroCallBody,
     val bodyHash: HashCode?,  // null for `include!` macro
     val depth: Int,
     /**
      * `srcOffset` - [CratePersistentId]
      * `dstOffset` - index of [MACRO_DOLLAR_CRATE_IDENTIFIER] in [body]
      */
     val dollarCrateMap: RangeMap = RangeMap.EMPTY,
+    val originalItem: Pair<VisItem, Set<Namespace>>? = null,
```
Undin, line 488: Let's add doc comment to clarify that `originalItem` is not null only in case of proc macro and only if this proc macro is not expanded by the plugin

src/main/kotlin/org/rust/lang/core/resolve2/FileModificationTracker.kt
```diff
@@ -77,6 +78,7 @@ private class ModDataLight {
     val enums: MutableList<EnumLight> = mutableListOf()
     val imports: MutableList<ImportLight> = mutableListOf()
     val macroCalls: MutableList<MacroCallLight> = mutableListOf()
+    val procMacroCalls: MutableList<ProcMacroCallLight> = mutableListOf()
```
Undin, line 81: Do I understand correctly that we need this list of proc macro calls to understand if something changed in the call and relaunch name resolution for the corresponding crate?

src/main/kotlin/org/rust/lang/core/psi/ProcMacroAttribute.kt
```diff
@@ -0,0 +1,164 @@
...
+            return when (val attr = getProcMacroAttributeRaw(owner, stub, explicitCrate)) {
+                Derive -> Derive
+                None -> None
+                is Attr -> {
+                    val props = attr.attr.resolveToProcMacroWithoutPsiUnchecked(checkIsMacroAttr = false)?.props
+                    if (props != null && props.treatAsBuiltinAttr && RsProcMacroPsiUtil.canFallbackItem(owner)) {
+                        None
+                    } else {
+                        attr
+                    }
```
Undin, line 159: Is it true that we return some `Attr` instance even if we failed to resolve proc macro?

src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
```diff
@@ -279,46 +283,62 @@ private fun <T> RsPossibleMacroCall.resolveToMacroAndThen(
 }
 
 fun RsMetaItem.resolveToProcMacroWithoutPsi(): ProcMacroDefInfo? {
+    val resolvedMacro = resolveToProcMacroWithoutPsiUnchecked(checkIsMacroAttr = true)
+        ?: return null
+    if (resolvedMacro.props.treatAsBuiltinAttr) return null
+    return resolvedMacro
+}
+
+fun RsMetaItem.resolveToProcMacroWithoutPsiUnchecked(checkIsMacroAttr: Boolean): ProcMacroDefInfo? {
```
Undin, line 292: If we find a way to get rid of this method, I will be happy

If we actually don't want to return `Attr` instance if `resolveToProcMacroWithoutPsiUnchecked` returns null in `org.rust.lang.core.psi.ProcMacroAttribute.getProcMacroAttribute`, then we can merge `resolveToProcMacroWithoutPsi` and `resolveToProcMacroWithoutPsiUnchecked`

src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
```diff
@@ -74,10 +76,18 @@ class RsMacroExpansionHighlightingPass(
...
-            macros += macro.prepare() ?: continue
+
+        PsiTreeUtil.processElements(file) {
+            if (it is RsMacroCall) {
+                if (it.macroArgument?.textRange?.intersects(restrictedRange) != true) return@processElements true
+                macros += it.prepare() ?: return@processElements true
+            } else if (it is RsAttrProcMacroOwner) {
+                if (it.textRange?.intersects(restrictedRange) != true) return@processElements true
+                macros += it.procMacroAttribute.attr?.prepare() ?: return@processElements true
+            }
```
Undin, line 87: nit: maybe `when` is more idiomatic here?

src/test/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPassTest.kt
```diff
@@ -30,6 +31,37 @@ class RsMacroExpansionHighlightingPassTest : RsAnnotationTestBase() {
...
+    fun `test attributes under attribute proc macro 2`() = checkHighlighting("""
+        use <CRATE>test_proc_macros</CRATE>::<FUNCTION>attr_as_is</FUNCTION>;
+
+        <ATTRIBUTE>#[<FUNCTION>attr_as_is</FUNCTION>]</ATTRIBUTE>
+        <ATTRIBUTE>#</ATTRIBUTE><ATTRIBUTE>[<FUNCTION>attr_as_is</FUNCTION></ATTRIBUTE><ATTRIBUTE>]</ATTRIBUTE>
+        <ATTRIBUTE>#</ATTRIBUTE><ATTRIBUTE>[allow</ATTRIBUTE><ATTRIBUTE>(foo</ATTRIBUTE><ATTRIBUTE>)]</ATTRIBUTE>
+        fn <FUNCTION>main</FUNCTION>() {
+            <ATTRIBUTE>#!</ATTRIBUTE><ATTRIBUTE>[crate_type = <STRING>"lib"</STRING></ATTRIBUTE><ATTRIBUTE>]</ATTRIBUTE>
+        }
+    """)
```
Undin, line 63: I'd add more tests with non trivial proc macro

Undin (commented):
Sorry, for long review but code is quite hard 😅 

It works in general but some notes and questions:
- Some tests fail
- I get `Low Memory` notification if I run IDE with default `-Xmx768m`. I see that you haven't checked memory consumption yet, so maybe it's worth adding a note in the tracking issue about it?
- I get errors with nightly toolchain (`rustc 1.56.0-nightly (b7404c898 2021-09-03)`) although everything is ok with stable one. Yet another change in API?

Also, it would be great if @dima74 checks changes in name resolution code

Undin (commented):
Still receive `Low Memory` notification on Rocket project with `-Xmx768m`
Note, `-Xmx1024m` is enough for me for this project

src/main/kotlin/org/rust/lang/core/macros/proc/ProcMacroExpander.kt
```diff
@@ -37,20 +38,40 @@ class ProcMacroExpander(
...
-        val env = call.packageEnv
+        val (macroCallBodyText, attrText) = when (val macroBody = call.macroBody) {
+            is MacroCallBody.Attribute -> macroBody.item to macroBody.attr
+            is MacroCallBody.Derive -> MappedText.single(macroBody.item, 0) to null
+            is MacroCallBody.FunctionLike -> MappedText.single(macroBody.text, 0) to null
+        }
         val (macroCallBodyLowered, rangesLowering) = project
-            .createRustPsiBuilder(macroCallBodyText)
+            .createRustPsiBuilder(macroCallBodyText.text)
             .lowerDocCommentsToPsiBuilder(project)
```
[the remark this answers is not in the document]
vlad20012: It's `null` there, haha

src/main/kotlin/org/rust/lang/core/psi/ProcMacroAttribute.kt
vlad20012: Absolutely

src/main/kotlin/org/rust/lang/core/resolve2/FacadeResolve.kt
vlad20012: Ok, now you can be happy :)

src/main/kotlin/org/rust/lang/core/psi/ProcMacroAttribute.kt
```diff
@@ -0,0 +1,164 @@
...
+                if (!stub.mayHaveCustomAttrs) {
+                    return if (stub.mayHaveCustomDerive) Derive else None
+                }
+            }
+
+            val crate = explicitCrate ?: owner.containingCrate ?: return None
+            val customAttributes = explicitCustomAttributes ?: CustomAttributes.fromCrate(crate)
+
+            owner.getQueryAttributes(crate, stub, fromOuterAttrsOnly = true).metaItems.forEachIndexed { index, meta ->
+                if (meta.name == "derive") return Derive
```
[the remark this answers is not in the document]
vlad20012: It's clarified a bit in #7725

src/main/kotlin/org/rust/lang/core/psi/ext/RsPossibleMacroCall.kt
```diff
@@ -83,57 +82,99 @@ val RsPossibleMacroCall.isMacroCall: Boolean
...
-            val owner = kind.meta.owner
-            if (owner is RsStructOrEnumItemElement) {
-                owner.preparedCustomDeriveMacroCallBody
-            } else {
-                null
+            val owner = kind.meta.owner as? RsAttrProcMacroOwner
+            when (val body = owner?.preparedProcMacroCallBody) {
+                is PreparedProcMacroCallBody.Attribute -> body.body.takeIf { body.attr.attr == this }
+                is PreparedProcMacroCallBody.Derive -> body.body.takeIf {
+                    owner is RsStructOrEnumItemElement && RsProcMacroPsiUtil.canBeCustomDerive(kind.meta)
```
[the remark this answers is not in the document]
vlad20012: It's different. `ProcMacroAttribute.getProcMacroAttributeRaw` checks an entire item when here we check a particular attribute

src/main/kotlin/org/rust/lang/core/resolve2/FileModificationTracker.kt
vlad20012: Yes

PR #7207, opened 2021-05-05
Anonymize info collected by `RsProjectUsagesCollector`

Undin:
Anonymize any counts by rounding to next power of two reported by `RsProjectUsagesCollector`


Undin:
CC @lancelote

PR #7211, opened 2021-05-06
COMP: add custom postfix templates

### LINKED ISSUE #6823: Enable Custom Postfix Completion
Simple feature request!  I would like the ability to build my own postfix completions in Rust like you can with other languages.  I'm happy to work on this myself if someone can point me in the correct direction

shocoman:
Short demonstration:
custom_template_1

Basically, it's the same as custom postfix templates in IDEA for Java or in Goland for Go. You can select several types (bool, number, reference, etc) for an expression which the template would be applicable to. You can also enter the type yourself (like "Vec", "String", or with full path "sdl2::event::Event" ("sdl2" is a crate name). Type parameters are ignored so "Vec" type would be the same as "Vec\<String\>" or "Vec\<Vec\<u32\>\>". In addition, if the type list is empty (it is by default), the template will be applicable to an expression of any type.

There is a limited support for user variables ($val$ in the gif) inside the template text. You can simply 'tab' through them but cannot choose the order. 

Unfortunately, there are no such nice things as "Live Template macros" or "TemplateContext" familiar from regular live templates. And I am not really sure we need them now.

Closes #6823

changelog: Allow users to create their own postfix templates

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/editable/RsPostfixTemplateEditor.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/editable/RsPostfixTemplateEditor.kt
@@ -1,0 +1,64 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.template.postfix.editable
+
+import com.intellij.codeInsight.template.impl.TemplateEditorUtil
+import com.intellij.codeInsight.template.postfix.settings.PostfixTemplateEditorBase
+import com.intellij.codeInsight.template.postfix.templates.PostfixTemplate
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.DefaultActionGroup
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.EditorFactory
+import com.intellij.openapi.project.DumbAwareAction
+import com.intellij.openapi.project.ProjectManager
+import com.intellij.openapi.ui.Messages
+import com.intellij.util.ui.FormBuilder
+import com.intellij.util.ui.UIUtil
+import org.rust.ide.template.postfix.RsPostfixTemplateProvider
+import javax.swing.JComponent
+import javax.swing.JPanel
+
+open class RsPostfixTemplateEditor(provider: RsPostfixTemplateProvider) :
+    PostfixTemplateEditorBase<RsPostfixTemplateExpressionCondition>(provider, createEditor(), true) {
+
+    var myPanel: JPanel = FormBuilder.createFormBuilder()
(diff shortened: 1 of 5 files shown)


Kobzol:
This looks awesome! I was slowly starting to work on this, but it seems that adding the `help wanted` tag on issues does wonders :laughing:

src/main/kotlin/org/rust/ide/template/postfix/editable/RsPostfixTemplateEditor.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.intellij.openapi.project.DumbAwareAction
+import com.intellij.openapi.project.ProjectManager
+import com.intellij.openapi.ui.Messages
+import com.intellij.util.ui.FormBuilder
+import com.intellij.util.ui.UIUtil
+import org.rust.ide.template.postfix.RsPostfixTemplateProvider
+import javax.swing.JComponent
+import javax.swing.JPanel
+
+open class RsPostfixTemplateEditor(val provider: RsPostfixTemplateProvider) :
```
vlad20012, line 24: The field could be private

src/main/kotlin/org/rust/ide/template/postfix/editable/RsPostfixTemplateExpressionCondition.kt
```diff
@@ -0,0 +1,112 @@
...
+import com.intellij.codeInsight.template.postfix.templates.editable.PostfixTemplateExpressionCondition
+import com.intellij.openapi.util.text.StringUtil
+import org.jdom.Element
+import org.rust.ide.presentation.renderInsertionSafe
+import org.rust.lang.core.psi.RsExpr
+import org.rust.lang.core.types.ty.*
+import org.rust.lang.core.types.type
+import java.util.*
+
+class RsPostfixTemplateExpressionCondition(val expressionType: Type, val userEnteredTypeName: String = "") : PostfixTemplateExpressionCondition<RsExpr> {
```
vlad20012, line 17: These fields could be private

src/main/kotlin/org/rust/ide/template/postfix/editable/RsPostfixTemplateExpressionCondition.kt
```diff
@@ -0,0 +1,112 @@
...
+    enum class Type {
+        Ref, Slice, Bool, Number, ADT, Array, Str, Tuple, Unit, UserEntered;
+
+        fun asId(): String = toString()
```
vlad20012, line 21:
```suggestion
        val id: String get() = toString()
```

vlad20012:
Please, use rebase instead of merge

shocoman:
Okay, I hope I did it right

vlad20012:
:+1: 

bors r+

PR #7228, opened 2021-05-17
QDOC: Make links that look like Rust paths work correctly in quick documentation

### LINKED ISSUE #7209: When click some links in rust doc, it will open a dialog to open crate link
## Environment

* **IntelliJ Rust plugin version:** 0.3.146.3826-211
* **Rust toolchain version:**  rustc 1.51.0 (2fd73fabe 2021-03-23)
* **IDE name and version:** IDEA 
* **Operating system:** Windos10 Enterprise 19041.928

## Problem description
When click some link in rust doc, it will open a dialog to open crate link
anim

## Steps to reproduce
open `io/mod.rs`  click    `std::fs::read_to_string`  in the doc of   `read_to_string`

shocoman:
changelog: Make links that look like Rust identifiers or paths (e.g., A, A::B, A::B::C) work correctly in quick documentation by prepending `psi_element://` to them. It fixes #7209
Now clickable links inside stdlib docs should work properly. 

0RV6H6R3Km

There is just a few lines of my code, but to write them I had to copy three classes from `org.intellij.markdown.html.LinkGeneratingProvider`. I used `RsPsiFactory.tryCreatePath()` to check if a link is a valid Rust path, but I am not sure how it is performance-wise. Probably I should replace it with some quick and simple check?

Also, I had to change one test, but it was incorrect anyway (the test itself passed, but the link didn't work when I checked it manually in IDE)

### DIFF
--- a/src/main/kotlin/org/rust/lang/doc/RsDocPipeline.kt
+++ b/src/main/kotlin/org/rust/lang/doc/RsDocPipeline.kt
@@ -18,12 +18,12 @@ import org.intellij.markdown.IElementType
 import org.intellij.markdown.MarkdownElementTypes
 import org.intellij.markdown.MarkdownTokenTypes
 import org.intellij.markdown.ast.ASTNode
+import org.intellij.markdown.ast.findChildOfType
 import org.intellij.markdown.ast.getTextInNode
 import org.intellij.markdown.flavours.MarkdownFlavourDescriptor
 import org.intellij.markdown.flavours.gfm.GFMFlavourDescriptor
-import org.intellij.markdown.html.GeneratingProvider
-import org.intellij.markdown.html.HtmlGenerator
-import org.intellij.markdown.html.SimpleTagProvider
+import org.intellij.markdown.html.*
+import org.intellij.markdown.html.entities.EntityConverter
 import org.intellij.markdown.parser.LinkMap
 import org.intellij.markdown.parser.MarkdownParser
 import org.rust.cargo.util.AutoInjectedCrates.STD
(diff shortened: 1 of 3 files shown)


vlad20012:
It looks like it can be done with less code copy-pasting. 
1. We don't need XSS-safe links. XSS safety makes sense only for web pages since there are cookies/local storage/other possibly sensitive data. XSS-safe links can be disabled using `useSafeLinks = false` in GFMFlavourDescriptor constructor (`private )val gfm: MarkdownFlavourDescriptor = GFMFlavourDescriptor(useSafeLinks = false, absolutizeAnchorLinks = true)`
2. Then we could just override `LinkGeneratingProvider.makeAbsoluteUrl` in all existing reference providers (`InlineLinkGeneratingProvider`/`ReferenceLinksGeneratingProvider`).
3. Instead of doing a full path validation `RsPsiFactory.tryCreatePath()` we could just check if a URI contains `/` (absolute or 2+ segment relative URI) or `.` (for 1-segment URI like `index.html)` or `#`

shocoman:
> 
> 
> It looks like it can be done with less code copy-pasting.
> 
>     1. We don't need XSS-safe links. XSS safety makes sense only for web pages since there are cookies/local storage/other possibly sensitive data. XSS-safe links can be disabled using `useSafeLinks = false` in GFMFlavourDescriptor constructor (`private )val gfm: MarkdownFlavourDescriptor = GFMFlavourDescriptor(useSafeLinks = false, absolutizeAnchorLinks = true)`
> 
>     2. Then we could just override `LinkGeneratingProvider.makeAbsoluteUrl` in all existing reference providers (`InlineLinkGeneratingProvider`/`ReferenceLinksGeneratingProvider`).
> 
>     3. Instead of doing a full path validation `RsPsiFactory.tryCreatePath()` we could just check if a URI contains `/` (absolute or 2+ segment relative URI) or `.` (for 1-segment URI like `index.html)` or `#`

Done
2) Method `makeAbsoluteUrl` is not opened, so I've overriden `renderLink` instead. Strangely, Intellij thinks that these classes are final but code compiles without a problem. That's why I reimplemented them in the first place. I guess there's something wrong with my IDE settings
internal

3) Actually, the "local" rust links may contain '#' symbol, here is example. Though I don't know how/if it's possible to jump to fragment part inside QuickDoc popup.
hash_in_link

vlad20012:
> Method makeAbsoluteUrl is not opened,

Oh, indeed. I'm just used to Java where `open` is not needed.

> Strangely, Intellij thinks that these classes are final

My IDE shows these errors too. I guess there's something wrong with our project configuration =D

> Actually, the "local" rust links may contain '#' symbol

Hmm, indeed. So, I guess at some point we may want to prepend `"psi_element://"` to such links in order to resolve them somehow

vlad20012 (approved):
bors r+

Issue #7219, opened 2021-05-11
File system completion for string literals should suggest paths relative to project root


shocoman:
## Environment

* **IntelliJ Rust plugin version:** 0.3.147.SNAPSHOT-211-dev
* **Rust toolchain version:** 1.51.0 (2fd73fabe 2021-03-23) x86_64-pc-windows-msvc
* **IDE name and version:** IntelliJ IDEA 2021.1 Ultimate Edition (IU-211.6693.111)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
Right now file system completion (#5430) displays files in the directory that contains our .rs file. It works correctly for macros like include_str(...) but not for functions like File::open(...) as it expects paths to be relative to project root. 

![before](https://user-images.githubusercontent.com/8329446/117814542-a3509700-b28e-11eb-8e58-4a5c4e0753f3.gif)

It would be pretty nice if the default directory for completion was project root (except the most obvious cases such as "include..." macros). I think it is more common use case. Optionally, we may be able to use a special path prefix ("./") to display paths relative to the current folder as before. Like this:
![after2](https://user-images.githubusercontent.com/8329446/117820728-00e7e200-b295-11eb-88df-75f53ff96507.gif)

Undin:
@Kobzol would you like to handle it since you implemented the corresponding functionality?
If so, I suppose we need to provide proper context in `org.rust.lang.core.resolve.ref.RsLiteralFileReferenceSet#getDefaultContexts` when we resolve/complete relative paths for `pathValueLiteral` pattern

Kobzol:
Sure, I'll implement this. I'm just now sure how it should work yet.

Making a distinction between `"path"` and `"./path"` doesn't seem right to me, I think that it's people just use these interchangeably and I don't think that we should distinguish between these.

Regarding the context directory, should it always be project root? We could also look at the current active Run target and try to take the working directory from it. WDYT?

Undin:
> Making a distinction between "path" and "./path" doesn't seem right to me, I think that it's people just use these interchangeably and I don't think that we should distinguish between these.

Agree

> Regarding the context directory, should it always be project root?

I think the behavior should be consistent with default working dir created for the Cargo run configuration. Also, it may be useful to check default implementation of `com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReferenceSet#getDefaultContexts`

[...]

PR #7236, opened 2021-05-19
COMP: change target of file system completion for path string literals

Kobzol:
This PR changes the context of filesystem path completion for path-like string literals. They will now resolve to the containing workspace root.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7219

changelog: Change the completion context of path string literals to the containing workspace root.


src/main/kotlin/org/rust/lang/core/resolve/ref/RsLitExprReferenceContributor.kt
```diff
@@ -59,7 +59,11 @@ private class RsLiteralFileReferenceSet(
                 val item = parent.ancestorStrict<RsModDeclItem>() ?: parent.ancestorStrict<RsMod>()
                 listOfNotNull(item?.containingMod?.getOwnedDirectory())
             }
-            else -> parentDirectoryContext
+            else -> {
+                val rsElement = element as? RsElement ?: return emptyList()
+                val workspaceRoot = rsElement.containingCargoPackage?.contentRoot ?: return emptyList()
+                return toFileSystemItems(workspaceRoot)
+            }
```
Undin, line 66: Unfortunately, it breaks name resolution and completion in `include` macros and that's why test fails.
For example, the following code is correct
```rust
// -crate_root.txt
/* hello from crate root! */

// -src/sibling.txt
/* hello from src dir! */

// -src/main.rs

use std::fs::File;
use std::io::Read;

fn main() {
    let mut buffer = String::new();
    File::open("crate_root.txt").unwrap().read_to_string(&mut buffer).unwrap();
    println!("{}", buffer);

    let s = include_str!("sibling.txt");
    println!("{}", s);
}
```
and it prints
```
/* hello from crate root! */
/* hello from src dir! */
```
But with these changes, `sibling.txt` cannot be resolved correctly.
So let's use `parentDirectoryContext` for `RsIncludeMacroArgument` as before

Undin (changes requested):
> I need a nudge. I want to set the directory to element.containingCargoPackage.workspace.contentRoot, i.e. the directory of the containing workspace root. However, this information is provided with a Path, while I need a VirtualFile. I cannot just use the filesystem to resolve the path, because it won't work in tests.
>
> Is there a way to get from Path to VirtualFile so that it works in tests?

As I can see, you've already solved this problem

Kobzol:
I actually didn't solve it yet :sweat:
I used `rsElement.containingCargoPackage?.contentRoot`, because it is a `VirtualFile`.
But I want to use `rsElement.containingCargoPackage?.workspace.contentRoot`, but that is a `Path`.
So I managed to get a virtual file, but for the wrong directory.

Undin:
> I actually didn't solve it yet 😓
I used rsElement.containingCargoPackage?.contentRoot, because it is a VirtualFile.
But I want to use rsElement.containingCargoPackage?.workspace.contentRoot, but that is a Path.
So I managed to get a virtual file, but for the wrong directory.

Oh, sorry, I've missed this detail. I've tried to provide a proper API for you in #7254

Kobzol:
Thank you! My head always spins when I'm trying to remember the difference between workspace, package, crate, project, cargo project etc. :laughing: So I wasn't sure if I missed anything.

Kobzol:
I changed it to point to workspace root and fixed the include macro case. I tested it in IDE, it seems to work. To properly test that it resolves to a parent workspace in tests, I guess I would have to create a completion test that would also inherit from `RsWithToolchainTestBase` to get access to a "real" project with a workspace?

Undin:
> To properly test that it resolves to a parent workspace in tests, I guess I would have to create a completion test that would also inherit from RsWithToolchainTestBase to get access to a "real" project with a workspace?

@Kobzol I suppose now it's enough to create a custom `org.rust.RustProjectDescriptorBase` with overridden `testCargoProject`

Kobzol:
I tried something like this:
```kotlin
@ProjectDescriptor(WithWorkspaceProjectDescriptor::class)
fun `test complete paths in path constructor 1`() = doSingleCompletionByFileTree("""
//- Cargo.toml
    [workspace]
    members = [
        "crate-a",
        "crate-b"
    ]
//- crate-a/src/main.rs
    fn main() {
        std::path::Path::new("crate-a/src/mai/*caret*/");
    }
//- foo.rs
    pub struct Foo;
""", """
    fn main() {
        std::path::Path::new("crate-a/src/main.rs/*caret*/");
    }
""")

object WithWorkspaceProjectDescriptor : WithRustup(DefaultDescriptor) {
    override fun testCargoProject(module: Module, contentRoot: String): CargoWorkspace {
        val crateA = testCargoPackage("$contentRoot/crate-a")
        val crateB = testCargoPackage("$contentRoot/crate-b")

        val packages = listOf(
            crateA,
            crateB
        )
        return CargoWorkspace.deserialize(
            Paths.get("${Urls.newFromIdea(contentRoot).path}/Cargo.toml"),
            CargoWorkspaceData(packages, emptyMap(), emptyMap(), contentRoot), CfgOptions.DEFAULT)
    }
}
```
Is it at least heading into the right direction? :)

Undin:
@Kobzol I think it's exactly what we need in the corresponding tests 👍 
Note, since in these "light" tests we don't use actual manifest files at all, we can omit `Cargo.toml` in the tests

Kobzol:
It didn't work for me at first, but I managed to get the stdlib + workspace descriptor to work, thanks!

src/test/kotlin/org/rust/lang/core/completion/RsPathCompletionTest.kt
```diff
@@ -0,0 +1,294 @@
...
+        mod foo;
+    //- bar.rs
+        fn bar() {}
+    """, """
+        #[cfg_attr(unix, path="bar.rs/*caret*/")]
+        mod foo;
+    """)
+}
+
+object WithWorkspaceProjectDescriptor: RustProjectDescriptorBase() {
```
Undin, line 279:
```suggestion
private object WithWorkspaceProjectDescriptor : RustProjectDescriptorBase() {
```

Undin (approved):
Thanks!

Kobzol:
bors r=Undin

Kobzol:
bors retry

Kobzol:
Btw @Undin could you please take a look at https://github.com/intellij-rust/intellij-rust/pull/5347? It's been sitting there for almost a year now :(

PR #7237, opened 2021-05-19
Update Rust ML ranking provider to 0.2.2

ortem:
Part of #6411.

This PR updates the Rust ML ranking provider version to 0.2.2. The new model takes into account Rust-specific features introduced in #6419, and so is more precise than the previous one.

In addition, this PR adds support for decorating the most relevant item with a star icon.

Screenshot 2021-05-19 at 16 22 48

This feature can be enabled via `Mark the most relevant item in the completion popup` option in `Preferences | Editor | General | Code Completion` settings.

changelog: Update experimental code completion ranking for Rust based on machine learning. The ranking model is disabled by default for now. You can enable it via `Editor | General | Code Completion | Machine Learning-Assisted Completion` [settings](https://www.jetbrains.com/help/idea/auto-completing-code.html#ml_completion).


ml-completion/src/main/kotlin/org/rust/ml/RsMLRankingProvider.kt
```diff
@@ -6,10 +6,15 @@
...
 
 import com.intellij.internal.ml.catboost.CatBoostJarCompletionModelProvider
+import com.intellij.internal.ml.completion.DecoratingItemsPolicy
 import com.intellij.lang.Language
 import org.rust.lang.RsLanguage
 
 @Suppress("UnstableApiUsage")
 class RsMLRankingProvider : CatBoostJarCompletionModelProvider("Rust", "rust_features", "rust_model") {
     override fun isLanguageSupported(language: Language): Boolean = language == RsLanguage
+    override fun getDecoratingPolicy(): DecoratingItemsPolicy = DecoratingItemsPolicy.Composite(
```
Undin, line 16: Looks like this API is only available since 211

ml-completion/src/main/kotlin/org/rust/ml/RsMLRankingProvider.kt
```diff
@@ -6,10 +6,15 @@
...
+        DecoratingItemsPolicy.ByAbsoluteThreshold(3.0),
+        DecoratingItemsPolicy.ByRelativeThreshold(2.5)
```
Undin, line 18: How these values were chosen? As I can see, different languages use different values

ortem: The threshold values are chosen separately for each ranking model, such that the probability of selecting the top element is high in a large number of sessions. So the first element is decorated with a star icon if it satisfies these thresholds. We should change these values to the new proper ones (obtained from the corresponding experiment analysis) with every model update.

ml-completion/src/main/kotlin/org/rust/ml/RsMLRankingProvider.kt
ortem: Thank you, fixed!

Undin (approved):
bors r+

Undin:
bors retry

Issue #6644, opened 2021-01-10
Add sanity check for missing input file


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.139.3589-203-nightly
* **Rust toolchain version:** 1.49.0 (e1884a8e3 2020-12-29) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.3.1 (CL-203.6682.181)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description

With #6332 in place you can specify a file to redirect input from. Unfortunately, if the file doesn't exist no relevant error is shown, instead a permanent "Builing..." progress bar is shown when you try to execute the run configuration

<img width="673" alt="building" src="https://user-images.githubusercontent.com/2500320/104122387-c5282480-5355-11eb-9a21-311bb159ad3c.png">

<details>
<summary>Error in the log</summary>

```
2021-01-10 15:05:22,083 [  33177]  ERROR - rains.concurrency.AsyncPromise - Cannot run program "/Users/jetbrains/.cargo/bin/cargo" (in directory "/Users/jetbrains/CLionProjects/rust_6332"): /Users/jetbrains/CLionProjects/rust_6332/input2.txt (No such file or directory) 
com.intellij.execution.process.ProcessNotCreatedException: Cannot run program "/Users/jetbrains/.cargo/bin/cargo" (in directory "/Users/jetbrains/CLionProjects/rust_6332"): /Users/jetbrains/CLionProjects/rust_6332/input2.txt (No such file or directory)
	at com.intellij.execution.configurations.GeneralCommandLine.createProcess(GeneralCommandLine.java:380)
	at com.intellij.execution.process.OSProcessHandler.startProcess(OSProcessHandler.java:101)
	at com.intellij.execution.process.OSProcessHandler.<init>(OSProcessHandler.java:49)
	at com.intellij.execution.process.KillableProcessHandler.<init>(KillableProcessHandler.java:37)
	at org.rust.cargo.runconfig.CargoRunStateBase.startProcess(CargoRunStateBase.kt:71)
	at org.rust.cargo.runconfig.buildtool.CargoBuildManager$build$1.invoke(CargoBuildManager.kt:116)
	at org.rust.cargo.runconfig.buildtool.CargoBuildManager$build$1.invoke(CargoBuildManager.kt:55)
[...]

PR #7239, opened 2021-05-20
RUN: Add sanity check for missing input file

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/6644.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
@@ -9,10 +9,7 @@ import com.intellij.execution.BeforeRunTask
 import com.intellij.execution.Executor
 import com.intellij.execution.InputRedirectAware
 import com.intellij.execution.configuration.EnvironmentVariablesData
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
```diff
@@ -130,6 +129,10 @@ open class CargoCommandConfiguration(
 
     @Throws(RuntimeConfigurationError::class)
     override fun checkConfiguration() {
+        if (isRedirectInput && redirectInputPath?.toPath()?.exists() != true) {
+            throw RuntimeConfigurationWarning("Input file doesn't exist")
```
Undin, line 133: BTW `Throws` annotation is wrong now. `checkConfiguration` can throw `RuntimeConfigurationWarning` as well, not only `RuntimeConfigurationError`
Let's use `@Throws(RuntimeConfigurationException::class)` instead

mchernyavsky: Fixed.

Undin (approved):
bors r+

PR #7240, opened 2021-05-20
Complete "tests" to cfg test mod tests

stepancheg:
Expand `tests` as:

```
#[cfg(test)]
mod tests {
}
```

There's no convention to place tests into `tests` module, but Rust
by Example book uses it:
https://doc.rust-lang.org/stable/rust-by-example/testing/unit_testing.html

changelog: complete `test` to `#[cfg(test)] mod tests {}`

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -112,6 +112,7 @@ snpefk
 sohich
 SomeoneToIgnore
 srjek
+stepancheg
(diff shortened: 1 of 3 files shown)


Undin:
@stepancheg do you know about `tmod` live template?

stepancheg:
No, I did not know it exists. Thanks!

Issue #5397, opened 2020-05-18
Syntax highlighting recognizes everything as variable by default


LastExceed:
## Environment

* **IntelliJ Rust plugin version:** 0.2.121.3059-201-stable
* **Rust toolchain version:** 1.41.0
* **IDE name and version:** IntelliJ Ultimate 2020.1.2 Preview
* **Operating system:** Windows 10 Home x64

## Problem description

Anything typed in the editor is highlighted as `Variables -> Default` unless it is explicitly recognized as something else. This can be very confusing when you're the type of programmer that likes to read information by color (who usually has a different color for variables compared to default text), most notably with syntax errors. Variables should only be highlighted in their respective color when recognized as such, just like everything else. unrecognized text should be displayed in the editor's default text color (or have it's own color).

I am not naming any actual colors since they depend on user configuration.

## Steps to reproduce

* set the highlighting color of `Variables -> Default` to some unique color of your choice
* type any random gibberish in a rust file
* see that it's highlighted in that color even though it's clearly not a variable

vlad20012:
Hmm, looks like our "Variables | Default" really means "any identifier". So, we don't have a color setting for "any variable". For some reason...

PR #7243, opened 2021-05-21
fix highlighting variables//default to actually highlight variables only

LastExceed:
step1 for #5397, can't figure out how to do step2 (implementing the missing option)

changelog: rename syntax highlighting option "variables -> default" to "identifier" since that's what the option has been doing the whole time (the actual option for default variables has yet to be implemented)

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -62,6 +62,7 @@ king6cong
 Kobzol
 kumbayo
 la10736
+LastExceed
 LiamClark
 Litarvan
 lundibundi
--- a/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
@@ -181,7 +181,7 @@ private fun colorFor(element: RsElement): RsColor? = when (element) {
     is RsModDeclItem -> RsColor.MODULE
     is RsMod -> if (element.isCrateRoot) RsColor.CRATE else RsColor.MODULE
     is RsPatBinding -> {
-        if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else RsColor.IDENTIFIER
+        if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else RsColor.VARIABLE
     }
     is RsStructItem -> when (element.kind) {
         RsStructKind.STRUCT -> RsColor.STRUCT
--- a/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsHighlightingAnnotator.kt
@@ -181,7 +181,7 @@ private fun colorFor(element: RsElement): RsColor? = when (element) {
     is RsModDeclItem -> RsColor.MODULE
     is RsMod -> if (element.isCrateRoot) RsColor.CRATE else RsColor.MODULE
     is RsPatBinding -> {
-        if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else null
+        if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else RsColor.IDENTIFIER
     }
     is RsStructItem -> when (element.kind) {
         RsStructKind.STRUCT -> RsColor.STRUCT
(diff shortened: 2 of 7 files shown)


vlad20012:
To be honest, we should totally remove such kind of identifier highlighting (and then somehow add a sane highlighting for variables). I see just one issue with it - `RsColor.IDENTIFIER` is (ab)used in `RsFormatMacroAnnotator`. I think we should introduce a separate `RsColor` (or maybe several ones) for this, but I'm not sure how to call it. Maybe @Kobzol could suggest something

Kobzol:
Right, formatting highlighting should definitely use something specific instead of just an identifier.
In
```rust
format!("{a:<5}", ...);
```
Rust calls `{}` named parameter and `<` and `5` are formatting parameters. Not very helpful.
In `RsFormatMacroAnnotator`, I call `{}` parameter and `<` and `5` specifiers. This makes sense to me.
The current highlighting uses `FORMAT_SPECIFIER` for `{}` and `IDENTIFIER` for `<` and `5`. Naming is hard :)

I suggest to rename `FORMAT_SPECIFIER` to `FORMAT_PARAMETER` and then introduce a new style named `FORMAT_SPECIFIER`, which will be used for the inner specifiers inside parameters.

vlad20012:
@Kobzol the one difference is that in string `"{a:<5}"` we actually highlight `a` and `5`, but not `<` and `5`

Kobzol:
Right. I would probably keep it that way, so that's it not all in the same color.

vlad20012:
@LastExceed, could you, please, rebase it to the latest master and then just remove `RsColor.IDENTIFIER` and all its usages? It should be simple since I removed `RsColor.IDENTIFIER` usages from `RsFormatMacroAnnotator` (in #7362).
I can do it by myself if you're not interested.

LastExceed:
@vlad20012 i can try, but shouldn't we do it the other way round (implement proper variable highlighting first and remove this option afterwards) ? I totally agree that this option doesn't make much sense in it's current form, but it's still better than not being able to color variables *at all*

vlad20012:
Ok, I think this also could be simple enough, so I can explain. 

There is a `RsColor.PARAMETER` color. It's used for function parameters highlighting. But syntactically, function parameters and local variables and the same - it is `RsPatBinding` PSI element. 
So there is a line `if (element.ancestorStrict<RsValueParameter>() != null) RsColor.PARAMETER else null` in `RsHighlightingAnnotator.kt` I guess that if we replace `null` to `RsColor.IDENTIFIER`, it will highlight variables (that are not function parameters) as `RsColor.IDENTIFIER`.

LastExceed:
easy. what should happen with underscores ? should they be highlighted as variable ?

vlad20012:
I'd say, the question is why they was highlighted as variables =D

I think that no, they shouldn't

LastExceed:
done

LastExceed:
gonna need help here. I can't even figure out where the tests are so no idea why they are failing

vlad20012:
Failed test classes:

`org.rust.ide.annotator.RsHighlightingAnnotatorTest`
`org.rust.ide.docs.RsCodeHighlightingInQuickDocumentationTest`
`org.rust.ide.docs.RsCodeHighlightingInQuickDocumentationTest`
`org.rust.ide.docs.RsCodeHighlightingInQuickDocumentationTest`
`org.rust.ide.docs.RsRenderedDocumentationTest`

You can run them locally to see the actual failures.
Tests should be just changed in order to reflect our changes in highlighting

Kobzol:
You can use Ctrl+Shift+N to quickly find files based on their filename. Then just click the arrow next to the Test class to run the corresponding tests, check how the output differs and modify the test so that it matches the current output.

LastExceed:
i was overwhelmed by the CI output, and running the `Test` run configuration in IntelliJ gave me `no tests were found`

with the file names I was able to find the failing tests and run them individually, but honestly I don't understand a thing. I could of course just copypaste the current output and be done with it, but I'm hestitant about changing things that I dont understand

should i really ?

vlad20012:
I've checked each test and it looks like yes, we should just copy-paste the new test output. I made a patch for you:

[Fix_tests.patch.zip](https://github.com/intellij-rust/intellij-rust/files/6685305/Fix_tests.patch.zip)

(btw you can apply it directly from IDEA using `Git | Patch | Apply Patch ...` action)

LastExceed:
done, thanks for your help

PR #7247, opened 2021-05-22
COMP: Display enums with variants in code completion popup when appropriate

shocoman:
Similar feature exists in Intellij for Java/Kotlin.

Demonstration: 
enum_variants

Variants are only suggested if this enum's type is expected
There is already completions for independent variants with autoimport, but I find that feature a bit lacking as it clutters namespace and users may not even know beforehand what variants an enum has. 

changelog: display enums with variants in code completion popup when appropriate

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
@@ -6,10 +6,7 @@
 package org.rust.lang.core.completion
 
 import com.google.common.annotations.VisibleForTesting
-import com.intellij.codeInsight.completion.CompletionParameters
-import com.intellij.codeInsight.completion.CompletionResultSet
-import com.intellij.codeInsight.completion.CompletionUtil
-import com.intellij.codeInsight.completion.InsertionContext
+import com.intellij.codeInsight.completion.*
 import com.intellij.codeInsight.lookup.LookupElement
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
```diff
@@ -411,3 +415,37 @@ private class RsImportLookupElement(
...
+            object : RsDefaultInsertHandler() {
+                override fun handleInsert(element: RsElement, scopeName: String, context: InsertionContext, item: LookupElement) {
+                    // move start offset from enum to its variant to escape it
+                    val enumStartOffset = context.startOffset
+                    val variantStartOffset = enumStartOffset + (enumName.length + 2)
+                    context.offsetMap.addOffset(CompletionInitializationContext.START_OFFSET, variantStartOffset)
+                    super.handleInsert(enumVariant, variantName, context, item)
+
+                    // escape enum name if needed
+                    if (element is RsNameIdentifierOwner && !RsNamesValidator.isIdentifier(enumName) && enumName !in CAN_NOT_BE_ESCAPED)
```
mchernyavsky, line 440: Please add some test(s) for this route.

shocoman: done. I added couple of tests for this method

mchernyavsky (approved):
Thanks!

bors r+

PR #7250, opened 2021-05-23
Update usage collectors

Undin:
General:
- Don't report anything if a project doesn't contain cargo projects

Toolchain usage collector: 
- support wsl toolchains
- use `EventFields.Enum` for `type` event instead of `EventFields.String`

Project usage collector:
- use `crateGraph` to handle the same dependencies in different cargo projects
- collect direct dependency count in addition to all dependency count (including transitive)
- report `0` instead of `1` if the original value is `0`


lancelote:
bors merge

PR #7253, opened 2021-05-24
FIX: AddRemainingArms intention incorrectly prints enum names and variants if they are escaped rust keywords

shocoman:
`AddRemainingArmsIntention` doesn't print enum names and variants correctly when they are escaped reserved Rust keywords. 
Here is example after activating `AddRemainingArmsFix`. And after that goes the same example with my tiny fix

Before: 
image

After:
image

### DIFF
--- a/src/main/kotlin/org/rust/ide/utils/checkMatch/Pattern.kt
+++ b/src/main/kotlin/org/rust/ide/utils/checkMatch/Pattern.kt
@@ -5,9 +5,7 @@
 
 package org.rust.ide.utils.checkMatch
 
-import org.rust.lang.core.psi.RsEnumItem
-import org.rust.lang.core.psi.RsEnumVariant
-import org.rust.lang.core.psi.RsStructItem
+import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.RsElement
 import org.rust.lang.core.psi.ext.RsFieldsOwner
 import org.rust.lang.core.psi.ext.findInScope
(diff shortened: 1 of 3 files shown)


shocoman:
Qodana doesn't like tests :worried:

Undin:
@shocoman yeah, unfortunately
Just ignore `unused` annotations for test methods. We are just setting it up and Qodana team hasn't fixed this false positive yet

src/main/kotlin/org/rust/ide/utils/checkMatch/Pattern.kt
```diff
@@ -23,10 +21,10 @@ data class Pattern(val ty: Ty, val kind: PatternKind) {
             is PatternKind.Binding -> kind.name
 
             is PatternKind.Variant -> {
-                val variantName = kind.variant.name.orEmpty()
-                val itemName = kind.item.name
+                val variantName = kind.variant.name.orEmpty().escapeIdentifierIfNeeded()
+                val itemName = kind.item.name.orEmpty().escapeIdentifierIfNeeded()
 
-                val variantInScope = ctx?.findInScope(variantName, VALUES)
+                val variantInScope = ctx?.findInScope(variantName.unescapeIdentifier(), VALUES)
```
Undin, line 27: Let's avoid unnecessary escaping and unescaping.
I suggest call `.escapeIdentifierIfNeeded()` in the following code below
```kotlin
val name = if (variantInScope != null) {
    variantName
} else {
    "$itemName::$variantName"
}
```

Note, it will require additional tests when all enum variants are imported. Something like
```rust
enum r#type { A, r#if, r#fn }

fn main() {
    use r#type::*;
    
    let a = A;
    match a/*caret*/ {

    }
}
```

src/main/kotlin/org/rust/ide/utils/checkMatch/Pattern.kt
```diff
@@ -45,8 +43,9 @@ data class Pattern(val ty: Ty, val kind: PatternKind) {
                     is TyTuple -> subPatterns.joinToString(", ", "(", ")") { it.text(ctx) }
                     is TyAdt -> {
                         val name = (ty.item as? RsStructItem)?.name ?: (ty.item as? RsEnumVariant)?.name
+                        val escapedName = name.orEmpty().escapeIdentifierIfNeeded()
```
Undin, line 46: Looks like the corresponding test doesn't cover this change
Let's make the following code work
```rust
struct r#type { r#if: bool }

fn foo(s: r#type) {
    match s/*caret*/ {
        r#type { r#if: true } => {}
    }
}
```

Undin (changes requested):

src/main/kotlin/org/rust/ide/utils/checkMatch/Pattern.kt
shocoman: Ok, I fixed it and added couple of tests. I didn't know that in some cases this intention can expand struct fields

Undin (approved):
Great!
Thank you!

Since it will be your first merged PR in this repo, could you move changes in `CONTRIBUTORS.txt` from #7211 to this PR, please?

bors d+

Undin:
@shocoman welcome to IntelliJ contributors!

PR #7255, opened 2021-05-24
REF: add default values to suggested change signature refactoring

Kobzol:
Follow-up to https://github.com/intellij-rust/intellij-rust/pull/7135, adds support for default values.

changelog: Add support for default values to suggested change signature refactoring.


dima74:
Thanks!
bors r+

dima74:
Hmm, why bors not started

bors r+

PR #7265, opened 2021-05-27
RUN: Add Cargo commands logging

ortem:
Now plugin logs every Cargo command executed through Cargo run configurations (including Build, Run, Debug, Test).
This might be useful when investigating a reported problem related to building, running, and debugging Rust code.

Note, to enable this _debug-level_ logging in IDE, `#org.rust.cargo.runconfig` should be entered into `Help | Diagnostic Tools | Debug Log Settings`.


Undin (approved):
Works for me

Interesting fact, that I didn't know before that `#some.prefix` enables all logs with `#some.prefix` prefix. It's probably worth highlighting this fact somewhere

bors r+

PR #7267, opened 2021-05-27
TY: infer expression type of log-like macros

Kobzol:
Previously, log-like macros were hardcoded in type inference to have type `TyUnit`. This PR changes it so that they are inferred properly.

Some notes:
1) I included some code that special cases the inference for `log` macros. However, with the new macro engine, it seems that they are correctly inferred as-is. Is it even worth it to keep the special case?
2) If yes, the commented code is not completely correct, since I use the name of the crate (not the name of the crate as a dependency), so it might not work correctly for renamed `log` dependency. But that should be rare and the macro is inferred properly in any case...
3) I didn't find any (macro) type inference tests that would use toolchain so that I could test this with the `log` crate. It it worth it? I think that the main point is that custom "log-like" macros are resolved. I included a test for that.

changelog: The type of custom macros that have log-like names (like `error!`, `warn!` or `debug!`) is now correctly inferred.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInferenceWalker.kt
@@ -1184,7 +1184,7 @@ class RsTypeInferenceWalker(
 
         inferChildExprsRecursively(macroCall)
         return when {
-            macroCall.assertMacroArgument != null || macroCall.logMacroArgument != null -> TyUnit
+            macroCall.assertMacroArgument != null -> TyUnit
             macroCall.formatMacroArgument != null -> inferFormatMacro(macroCall)
             macroCall.includeMacroArgument != null -> inferIncludeMacro(macroCall)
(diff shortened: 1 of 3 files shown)


vlad20012:
To be honest, I think we can remove all the hacks around log macros. I mean, we can totally remove all special cases from everywhere, starting from the parser. We just should carefully check that everything works fine without these hacks.

Regarding this PR, yes, I think special cases aren't needed and the tests can be removed as well.

vlad20012 (approved):

Kobzol:
This test is failing after removing the hard-coded unit type:
```kotlin
    //From the log crate
    fun `test warn!`() = stubOnlyTypeInfer("""
    //- main.rs
        fn main() {
            let x = warn!("Something went wrong");
            x;
          //^ ()
        }
    """)
```
I suppose that we can remove that test?

vlad20012:
> I suppose that we can remove that test?

Yes, but there is an alternative - you can copy-paste the text of `warn` macro to the test, like we did in `RsErrorChainMacroExpansionTest`

Kobzol:
bors r=vlad20012

PR #7273, opened 2021-05-29
RUN: Fix validation of input file

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/pull/7239.

TODO: Fix sanity check for missing input file in the Run Configuration settings.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
+++ b/src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
@@ -35,7 +35,7 @@ import org.rust.cargo.toolchain.RustChannel
 import org.rust.cargo.toolchain.tools.isRustupAvailable
 import org.rust.ide.experiments.RsExperiments
 import org.rust.openapiext.isFeatureEnabled
-import org.rust.stdext.toPath
+import org.rust.stdext.toPathOrNull
 import java.io.File
 import java.nio.file.Path
 import java.nio.file.Paths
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
```diff
@@ -129,8 +129,13 @@ open class CargoCommandConfiguration(
 
     @Throws(RuntimeConfigurationException::class)
     override fun checkConfiguration() {
-        if (isRedirectInput && redirectInputPath?.toPath()?.exists() != true) {
-            throw RuntimeConfigurationWarning("Input file doesn't exist")
+        if (isRedirectInput) {
+            val path = redirectInputPath?.toPathOrNull()
+            when {
+                path?.exists() != true -> throw RuntimeConfigurationWarning("Input file doesn't exist")
```
Undin, line 135: It seems it works incorrectly if I pass `"` on Windows because in this case `path` will be null and we show `Input file doesn't exist` instead of `Input file is not valid`.
`path` can be null only in two cases: `redirectInputPath` is `null` or `redirectInputPath` cannot be converted to valid path on particular OS. And looks like we should show `Input file is not valid` only in the second case

src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
```diff
@@ -129,8 +129,13 @@ open class CargoCommandConfiguration(
...
+                !path.toFile().isFile -> throw RuntimeConfigurationWarning("Input file is not valid")
```
Undin, line 136: Why we need this complicated check?
Do we want to show `Input file is not valid` in case of directory?

mchernyavsky: Yes.

src/main/kotlin/org/rust/cargo/runconfig/command/CargoCommandConfiguration.kt
Undin: Decided that file with invalid sumbols cannot exist on filesystem as well so `Input file doesn't exist` is valid here as well

Undin (approved):
bors r+

Undin:
@mchernyavsky don't forget to add `changelog` section for the PR

PR #7274, opened 2021-05-29
CI: cancel previous ci runs via native `concurrency` feature

Undin:
Recently, GitHub [announced](https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/) a `concurrency` [feature](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) to run only single workflow/job for provided concurrency group. In combination with the `cancel-in-progress: true` flag, it also allows canceling already running tasks for the same concurrency group via builtin GitHub feature.
As a result, we can replace hacky `styfle/cancel-workflow-action` action that we used before to cancel outdated workflow runs (that actually worked not quite well when all available GitHub runners were busy) with the native GitHub feature.

changelog: Improve cancellation of outdated CI runs

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -8,18 +8,14 @@ on:
             - trying
     pull_request:
 
-jobs:
-    # TODO: drop it when GitHub supports its by itself
-    cancel-previous:
-        runs-on: ubuntu-18.04
-        steps:
-            - name: Cancel Previous Runs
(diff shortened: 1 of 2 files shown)


Undin:
CC @vlad20012 @lancelote

Undin:
bors retry

Issue #7203, opened 2021-05-04
Auto-select WSL based toolchain when opening a project inside 


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.3.147.3837-211-nightly
* **Rust toolchain version:** 1.51.0 (2fd73fabe 2021-03-23) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2021.1.1 (CL-211.7142.21)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

When I open a project which is located directly in WSL FS (using #6456) CLion automatically selects a default Rust toolchain in Windows FS, which is not very convenient.

Would prefer it to select a Rust toolchain inside the WSL.

iamwwc:
`To turn it on, enable org.rust.wsl experimental feature`, Where I can enable this feature :) @lancelote

Undin:
@iamwwc https://plugins.jetbrains.com/plugin/8182-rust/docs/rust-faq.html#experimental-features

PR #7217, opened 2021-05-11
UI: Extend toolchain path field to have an opportunity to choose a suggested path

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/7196.

Screenshot 2021-05-11 at 03 17 59

changelog: Extend toolchain path field to have an opportunity to choose a suggested path

PR #7279, opened 2021-05-31
PRJ: Auto-select WSL based toolchain when opening a project inside WSL FS

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/7203

changelog: Auto-select WSL based toolchain when opening a project inside WSL FS

### DIFF
--- a/src/main/kotlin/org/rust/cargo/toolchain/RsToolchainBase.kt
+++ b/src/main/kotlin/org/rust/cargo/toolchain/RsToolchainBase.kt
@@ -8,6 +8,7 @@ package org.rust.cargo.toolchain
 import com.intellij.execution.configuration.EnvironmentVariablesData
 import com.intellij.execution.configurations.GeneralCommandLine
 import com.intellij.execution.configurations.PtyCommandLine
+import com.intellij.execution.wsl.WslPath
 import com.intellij.openapi.util.SystemInfo
 import com.intellij.util.io.exists
 import com.intellij.util.net.HttpConfigurable
@@ -15,6 +16,7 @@ import com.intellij.util.text.SemVer
 import org.rust.cargo.CargoConstants
 import org.rust.cargo.toolchain.flavors.RsToolchainFlavor
 import org.rust.cargo.toolchain.tools.Cargo
+import org.rust.cargo.toolchain.wsl.getHomePathCandidates
 import org.rust.openapiext.GeneralCommandLine
 import org.rust.openapiext.withWorkDirectory
 import java.io.File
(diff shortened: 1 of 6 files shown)


src/211/main/kotlin/org/rust/cargo/toolchain/wsl/RsWslToolchainFlavor.kt
```diff
@@ -19,42 +20,11 @@ import java.nio.file.Path
...
-                if (!localPath.isDirectory()) continue
-                yield(localPath)
-            }
-
-            for (remotePath in listOf("/usr/local/bin", "/usr/bin")) {
-                val localPath = root.resolve(remotePath)
-                if (!localPath.isDirectory()) continue
-                yield(localPath)
-            }
+            yieldAll(distro.getHomePathCandidates())
```
vlad20012, line 27: Let's remove `sequence {}` and just return `distro.getHomePathCandidates()`

Undin (changes requested):
I've dropped 203 support so the code can be simplified.

Also, it would also be useful to select a proper initial toolchain in project creation dialog (at least when default project location is already known, i.e. in minor IDEs). But I think it should be implemented only after #7217 and this PRs

mchernyavsky:
@Undin Done.

src/main/kotlin/org/rust/cargo/toolchain/RsToolchainBase.kt
```diff
@@ -126,11 +129,24 @@ abstract class RsToolchainBase(val location: Path) {
...
+            val path = projectDir?.toString() ?: return null
+            if (!WslDistributionManager.isWslPath(path)) return null
+
+            val distribution = WslPath.getDistributionByWindowsUncPath(path)
+            val toolchain = distribution
+                ?.getHomePathCandidates()
+                ?.filter { RsToolchainFlavor.getFlavor(it) != null }
+                ?.mapNotNull { RsToolchainProvider.getToolchain(it.toAbsolutePath()) }
+                ?.firstOrNull()
+            if (toolchain != null) return toolchain
```
Undin, line 143: Now, if you pass `null` as `projectDir` or pass non-wsl path, we always return null as suggested toolchain

Undin (changes requested):

src/main/kotlin/org/rust/cargo/toolchain/RsToolchainBase.kt
mchernyavsky: Fixed.

Undin (approved):
bors r+

Issue #7245, opened 2021-05-22
using WSL toolchains got unknown type for Arc


zeyugao:
## Environment

* **IntelliJ Rust plugin version:**: 0.4.148.3878-211-nightly with org.rust.wsl enabled
* **Rust toolchain version:**: 1.51
* **IDE name and version:**: Clion 2021.1
* **Operating system:**: Windows 10 19043 and WSL ArchLinux

## Problem description
wsl support cannot corrently get the type for Arc::new()

## Steps to reproduce
```rust
use std::sync::{Arc, Mutex};
fn main(){
    let r = Arc::new(Mutex::new(Vec::new()));
}
```

When using toolchains on windows, it works correctly, including the autocompletion

![image](https://user-images.githubusercontent.com/6374697/119212929-3343db80-baee-11eb-893f-a9d30335fbdb.png)

But after switching to toolchains in wsl, it gets unknown type.

![image](https://user-images.githubusercontent.com/6374697/119212960-6a19f180-baee-11eb-96d8-3244e9d58e8c.png)

Also, the autocompletion is broken

![image](https://user-images.githubusercontent.com/6374697/119212972-8322a280-baee-11eb-889c-75e57eb2f64f.png)

mchernyavsky:
@zeyugao Can you please check it on 1.54.0-nightly?

Don't forget to refresh Cargo projects:

<img width="270" alt="Screenshot 2021-05-31 at 06 12 14" src="https://user-images.githubusercontent.com/6079006/120134875-bb556f80-c1d7-11eb-8ed4-86f84b3ef61e.png">

zeyugao:
@mchernyavsky Well, it works charmingly.

PR #7284, opened 2021-05-31
RUN: Fix fetching of cfg-options on WSL

mchernyavsky:
Fixes https://github.com/intellij-rust/intellij-rust/issues/7245.

changelog: Fix fetching of cfg-options on WSL


mchernyavsky (commented):
bors r+

Issue #7195, opened 2021-04-30
Setup tests for WSL toolchains on CI


mchernyavsky:
Relates to https://github.com/intellij-rust/intellij-rust/pull/6456.

PR #7285, opened 2021-05-31
CI: Test WSL toolchains on CI

mchernyavsky:
Closes https://github.com/intellij-rust/intellij-rust/issues/7195.

### DIFF
--- a/.github/workflows/check-wsl.yml
+++ b/.github/workflows/check-wsl.yml
@@ -1,0 +1,125 @@
+name: check WSL
+
+on:
+    push:
+        branches:
+            - master
+            - staging
+            - trying
+    pull_request:
+
+# Allow cancelling all previous runs for the same branch
+# See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+jobs:
+    build-native-code:
+        uses: ./.github/workflows/build-native-code.yml
+        with:
+            rust-version: 1.64.0
+            cache: true
+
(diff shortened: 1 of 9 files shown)


ssbarnea:
@mchernyavsky Out of curiosity, what is your current take regarding WSL support on GHA? I am asking this because I need do find a working way to make use of WSL for another project where on WSL would work. Seen a project that correctly makes use of WSL in GHA could of a great help.

ssbarnea (changes requested):
It might me my job but this approach is guaranteed to break as soon github upgrades the runner, and that happens quite often and without notice.

mchernyavsky:
@ssbarnea Looks like the workflow is not hanging since Rust 1.64.0, so I'm not sure that the problem is related to the runner.

ssbarnea:
> @ssbarnea Looks like the workflow is not hanging since Rust 1.64.0, so I'm not sure that the problem is related to the runner.

My only concern is related to use of a full path `C:\hostedtoolcache\windows\Ubuntu\20.4.0\x64\ubuntu.exe` which is high likely to break when Github will refresh their container image. They do this often, updating the hostedtoolcache and that is not really announced. I think that we should determine the path instead to make the pipeline more resilient.

mchernyavsky:
> Looks like the workflow is not hanging since Rust 1.64.0

No, it doesn't work :(

> I think that we should determine the path instead to make the pipeline more resilient.

I haven't tested it, but probably you could do this to change default user:

```
- name: Set up WSL
  uses: Vampire/setup-wsl@v1
  with:
      distribution: Ubuntu-20.04
      wsl-shell-user: test
      wsl-conf: |
          [user]
          default = test
```

Issue #6045, opened 2020-09-01
asm! macro meta issue


Kobzol:
This is a meta issue about support for the new `asm!` macro.

Implement:
- [ ] parsing (https://github.com/intellij-rust/intellij-rust/pull/5928)
- [ ] argument highlight + check
- [ ] completion
- [ ] inference
- [ ] control flow (https://github.com/intellij-rust/intellij-rust/issues/5815)
- [ ] surround with unsafe (https://github.com/intellij-rust/intellij-rust/issues/6044)

Original issue: https://github.com/intellij-rust/intellij-rust/issues/1940

PR #7288, opened 2021-06-01
Take `asm!` macro calls into account in control flow and data flow

### LINKED ISSUE #5815: Overzealous with warnings - does not look into "unsafe" blocks?
## Environment

* **IntelliJ Rust plugin version:** 0.3.127.3249-201
* **Rust toolchain version:** 1.47.0-nightly (d6953df14 2020-07-25) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1.3 (CL-201.8743.17)

## Problem description
I'm getting a false positive:

IntelliJ-Rust_issue

*Yes I know what `#![deny(warnings)]` does, but even the current `clippy` is perfectly OK with this code - so why is this plugin unhappy?*

`Clippy` output:
```
/Users/ur20980/.cargo/bin/cargo clippy --color=always
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s

```
## Steps to reproduce

Try the following Rust code:
```rust
#![feature(asm)]
#![deny(warnings)]
#![deny(clippy::all)]

fn main() {
    let z: i32 = add_new(4, 11);
    println!("4 + 11 = {}", z);
}

fn add_new(a: i32, b: i32) -> i32 {
    let c: i32;
    unsafe {
        asm!(
            "add {0:e}, {1:e}",
            inlateout (reg) a => c,
            in (reg) b,
        );
    }
    c
}
```

### LINKED ISSUE #6937: Having variable as output of asm! macro is not counted as initializing it
## Environment

* **IntelliJ Rust plugin version:** 0.3.142.3705-203
* **Rust toolchain version:** 1.52.0-nightly (caca2121f 2021-03-05) x86_64-unknown-none
* **IDE name and version:** CLion 2020.3.2 (CL-203.7148.70)
* **Operating system:** Linux 5.11.2-arch1-1

## Problem description
I have a function in my code that reads from a register into a variable then returns it. The variable is not initialized before being used inside of the asm! macro and when I try to return in, Clion tells me it is not initialized and underlines it with squiggles as if it was an error.

## Steps to reproduce

```rust
    pub fn read_cr3() -> usize {
        let result: u64;
        unsafe {
            asm!("mov {}, cr3", out(reg) result);
        };
        result as usize
    }
```

With this Clion says there is an error when there isn't

### LINKED ISSUE #7213: Code Analysis Inevitably Gets Into an "Analyzing Loop" with the Use of the "asm" Macro
## Environment

* **IntelliJ Rust plugin version:** 0.3.147.3845-211-nightly
* **Rust toolchain version:** 1.54.0-nightly (676ee1472 2021-05-06) x86_64-pc-windows-msvc
* **IDE name and version:** IntelliJ IDEA 2021.1.1 Community Edition (IC-211.7142.45)

## Problem description

The code analysis seems to inevitably goes into an "analyzing loop" with the use of the [`asm!` macro](https://doc.rust-lang.org/unstable-book/library-features/asm.html). 

It is just stuck here from my local Windows machine:
image

## Steps to reproduce

1. Add `#![feature(asm)]` to your `lib.rs` or `main.rs` (as this is a nightly-only unstable feature right now);
2. Add a line which uses the `asm` macro;
3. Now the code analysis *should* be stuck at "Analyzing..." as shown from the image attached above.

ortem:
This PR aims to fix several problems related to  `asm!()` macro calls processing in control flow and data flow analysis through the following steps:

* Parse `asm!()` macro call as expression. This is needed to wrap an `asm!()` macro call into `RsExprStmt`. As a result, now `expandedStmtsAndTailExpr` contains `asm!()` macro calls
* Process functions with built-in macro calls in liveness analysis. Liveness analysis ignored functions containing built-in macro calls since #6876 because such macro calls do not have `expansion` anymore. This change isolated from the rest might produce new false positives since `asm!` is a built-in macro so it is attached
* Initially handle `asm!()` macro calls during control flow graph building. Previously, `asm!()` macro calls caused `java.lang.IllegalStateException: unreachable` (see #7213). Proper `asm!()` macro calls processing should be added later after more complete `asm!` macro support is implemented by the plugin (see #6045)
* Disable data-flow analysis on functions with `asm!()` macro calls in order not to provide false-positive errors and warnings (see #5815 and #6937)

Fixes #5815
Fixes #6937
Fixes #7213

Part of #6045

changelog: Fix false-positive `Use of possibly uninitialized variable` [E0381](https://doc.rust-lang.org/error-index.html#E0381) errors and infinite analysis when a function contains [`asm!()`](https://github.com/rust-lang/rfcs/blob/master/text/2873-inline-asm.md) macro calls


vlad20012 (approved):
bors r+

Issue #7295, opened 2021-06-02
Rust REPL doesn't work with recent evcxr_repl v0.10.0


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.4.148.3892-211-nightly
* **Rust toolchain version:** 1.53.0-nightly (42816d61e 2021-04-24) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2021.1.2 RC Ultimate Edition (IU-211.7442.27)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
I can pass only one command. The second one cannot be entered and the IDE shows errors popup 

![image](https://user-images.githubusercontent.com/2539310/120469879-974a8780-c3ab-11eb-9a09-364061e7d19d.png)

Note, everything is fine with `evcxr_repl v0.9.0`

wonbyte:
I also am experiencing this problem.

PR #7296, opened 2021-06-02
CI: use evcxr_repl 0.9 in tests

Undin:
Recent 0.10.0 version doesn't work with IDE integration.
https://github.com/intellij-rust/intellij-rust/issues/7295


Undin:
CC @dima74 

Also, note that we still don't have repl tests on Windows

PR #7299, opened 2021-06-02
DBG: Pretty-print HashMap/HashSet with MSVC LLDB

ortem:
Previously, our pretty-printers could not render `HashMap`/`HashSet` with MSVC LLDB (which does not contain Rust patches yet). The new implementation uses type lookup by name when template arguments are not recognized and should work without Rust patches.

Part of #5632.

changelog: Fix debug pretty-printers to render `HashMap`/`HashSet` with MSVC LLDB on Windows


Undin (commented):
@ortem is it possible to set up pretty printers tests for Windows to verify such changes?

ortem:
@Undin Yes, I implemented them: https://github.com/intellij-rust/intellij-rust/pull/7399

Undin (approved):
🔥 
bors r+

Undin:
bors retry

PR #7301, opened 2021-06-03
COMP: deduplicate method completion variants by name and trait

### LINKED ISSUE #7238: Duplicate completion variants for methods
## Environment

* **IntelliJ Rust plugin version:** 0.3.148.3855-211-nightly
* **Rust toolchain version:** 1.53.0-nightly (b0c818c5e 2021-04-16) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2021.1.1 Ultimate Edition (IU-211.7142.45)
* **Operating system:** macOS 11.2.3

## Problem description
When using the **new macro expansion engine**, the plugin suggests the same completion variants for some methods multiple times. For example, completion results for `fmt`  method contain two similar variants, and both variants lead to the same code with `use std::fmt::Debug`.

duplicate_comp

However, there are no such duplicates when using the old macro expansion engine.

## Steps to reproduce

```Rust
fn main() {
    "123".fmt/*caret*/
}
```

vlad20012:
Fixes #7238

changelog: Fix method duplication issue in completion list

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
@@ -115,11 +115,11 @@ object RsCommonCompletionProvider : RsCompletionProvider() {
                     filterAssocTypes(
                         element,
                         filterCompletionVariantsByVisibility(
+                            element,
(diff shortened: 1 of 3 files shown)


ortem (approved):
LGTM
bors r+

PR #7303, opened 2021-06-03
INSP: Implement `unknown_crate_types` lint

avrong:
[The lint in the rustc book](https://doc.rust-lang.org/rustc/lints/listing/deny-by-default.html#unknown-crate-types)

[`Linkage` section in the Rust Reference](https://doc.rust-lang.org/reference/linkage.html)

Part of https://github.com/intellij-rust/intellij-rust/issues/5888

changelog: Detect `unknown_crate_type` compiler lint

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspection.kt
@@ -1,0 +1,41 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.lints
+
+import com.intellij.psi.PsiElement
+import com.intellij.util.ProcessingContext
+import org.rust.ide.inspections.RsProblemsHolder
+import org.rust.lang.core.RsPsiPattern
+import org.rust.lang.core.psi.RsInnerAttr
+import org.rust.lang.core.psi.RsLitExpr
+import org.rust.lang.core.psi.RsMetaItem
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.isRootMetaItem
+import org.rust.lang.core.psi.ext.stringValue
+
+class RsUnknownCrateTypesInspection : RsLintInspection() {
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnknownCrateTypes
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): RsVisitor =
+        object : RsVisitor() {
+            override fun visitLitExpr(o: RsLitExpr) {
+                val parent = o.parent as? RsMetaItem ?: return
+                if (parent.path?.text != "crate_type") return
+
+                val context = ProcessingContext()
+                if (!parent.isRootMetaItem(context)) return
+                if (context.get(RsPsiPattern.META_ITEM_ATTR) !is RsInnerAttr) return
+
+                if (o.stringValue !in KNOWN_CRATE_TYPES) {
+                    holder.registerLintProblem(o, "Invalid `crate_type` value")
+                }
+            }
+        }
+
+    companion object {
+        private val KNOWN_CRATE_TYPES = listOf("bin", "lib", "dylib", "staticlib", "cdylib", "rlib", "proc-macro")
+    }
+}
--- a/src/test/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspectionTest.kt
+++ b/src/test/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspectionTest.kt
@@ -1,0 +1,37 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.lints
+
+import org.rust.ide.inspections.RsInspectionsTestBase
+
+class RsUnknownCrateTypesInspectionTest : RsInspectionsTestBase(RsUnknownCrateTypesInspection::class) {
+    fun `test incorrect crate type`() = checkWarnings("""
+        #![crate_type = <error descr="Invalid `crate_type` value">"foo"</error>]
+    """)
+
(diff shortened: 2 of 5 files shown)


src/main/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspection.kt
```diff
@@ -0,0 +1,33 @@
...
+                if (parent.path?.text != "crate_type") return
+
+                if (o.stringValue !in KNOWN_CRATE_TYPES) {
+                    holder.registerLintProblem(o, "Invalid `crate_type` value")
+                }
+            }
+        }
+
+    companion object {
+        val KNOWN_CRATE_TYPES = listOf("bin", "lib", "dylib", "staticlib", "cdylib", "rlib", "proc-macro")
```
Undin, line 31: private

src/main/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspection.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.stringValue
+
+class RsUnknownCrateTypesInspection : RsLintInspection() {
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnknownCrateTypes
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): RsVisitor =
+        object : RsVisitor() {
+            override fun visitLitExpr(o: RsLitExpr) {
+                val parent = o.parent as? RsMetaItem ?: return
```
Undin, line 21: I think we should check only top level meta items not to produce false-positive errors in such cases
```rust
#[bar(crate_type = "bar")]
fn foo() {}
```
where `bar` is proc macro

Undin, line 21: Also, I suppose, we also should check that we are in inner attribute since the compiler for following code
```rust
#[crate_type = "foo"]
fn main() {}
```
produces only
```
warning: unused attribute
 --> src/main.rs:1:1
  |
1 | #[crate_type = "foo"]
  | ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_attributes)]` on by default

warning: crate-level attribute should be an inner attribute: add an exclamation mark: `#![foo]`
 --> src/main.rs:1:1
  |
1 | #[crate_type = "foo"]
  | ^^^^^^^^^^^^^^^^^^^^^

warning: 2 warnings emitted
```

Undin (changes requested):

src/main/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspection.kt
Undin: See `org.rust.lang.core.psi.ext.RsMetaItemKt#isRootMetaItem`, it should help here

Undin: `org.rust.lang.core.psi.ext.RsMetaItemKt#isRootMetaItem` can also help here. If you pass `ProcessingContext` into it, it put attribute into it by `org.rust.lang.core.RsPsiPattern#META_ITEM_ATTR` key

src/main/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspection.kt
```diff
@@ -0,0 +1,41 @@
...
+                if (parent.path?.text != "crate_type") return
+
+                val context = ProcessingContext()
+                if (!parent.isRootMetaItem(context)) return
+                if (context.get(RsPsiPattern.META_ITEM_ATTR) !is RsInnerAttr) return
```
Undin, line 30: Looks like this case is not covered in tests

src/test/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspectionTest.kt
```diff
@@ -0,0 +1,36 @@
...
+    fun `test incorrect crate type`() = checkWarnings("""
+        #![crate_type = <error descr="Invalid `crate_type` value">"foo"</error>]
+    """)
+
+    fun `test correct crate type`() = checkWarnings("""
+        #![crate_type = "cdylib"]
+    """)
+
+    fun `test ignore if inner`() = checkWarnings("""
+        #[<error descr="identifier expected, got 'crate_type'">c</error>rate_type <error descr="'!' or '::' expected, got '='">=</error> "foo"]
```
Undin, line 20: You need something to put the attribute on it (for example, `main` function). Otherwise, the plugins can't parse attribute correctly.
See `OuterAttr` rule in `RustParser.bnf` and its usages

src/test/kotlin/org/rust/ide/inspections/lints/RsUnknownCrateTypesInspectionTest.kt
```diff
@@ -0,0 +1,37 @@
...
+class RsUnknownCrateTypesInspectionTest : RsInspectionsTestBase(RsUnknownCrateTypesInspection::class) {
+    fun `test incorrect crate type`() = checkWarnings("""
+        #![crate_type = <error descr="Invalid `crate_type` value">"foo"</error>]
+    """)
+
+    fun `test correct crate type`() = checkWarnings("""
+        #![crate_type = "cdylib"]
+    """)
+
+    fun `test ignore if inner`() = checkWarnings("""
```
Undin, line 19:
```suggestion
    fun `test ignore if outer`() = checkWarnings("""
```

Undin (approved):

Undin (approved):
bors r+

Undin:
@avrong Possible improvements of this feature:
- provide a quick fix to use the correct value. It may be useful when a user makes a mistake (like `dlib` instead of `dylib`). `org.rust.ide.annotator.fixes.NameSuggestionFix.Companion#createApplicable` may be helpful here
- provide completion of the corresponding values since we already have a code that can detect necessary place and list of possible values

PR #7307, opened 2021-06-04
CI&MACRO: expand proc macros on Apple Silicon machines

Undin:
Part of #6908

changelog: Support [procedural macro](https://doc.rust-lang.org/reference/procedural-macros.html) expansion on Apple Silicon machines. Note:
- you should use `aarch64-apple-darwin` toolchain to make procedural macro expansion work
- the proc macro expansion is still disabled by default. To turn it on, enable `org.rust.cargo.evaluate.build.scripts` and `org.rust.macros.proc` [experimental features](https://plugins.jetbrains.com/plugin/8182-rust/docs/rust-faq.html#experimental-features)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -64,17 +64,22 @@ jobs:
                       arch: x86-64
                       os: ubuntu-latest
                       target: x86_64-unknown-linux-gnu
-                      artifact_path: native-helper/target/release/intellij-rust-native-helper
+                      artifact_path: native-helper/target/x86_64-unknown-linux-gnu/release/intellij-rust-native-helper
                     - name: windows
                       arch: x86-64
                       os: windows-latest
                       target: x86_64-pc-windows-msvc
-                      artifact_path: native-helper/target/release/intellij-rust-native-helper.exe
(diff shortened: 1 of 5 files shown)


Undin:
@ortem I've added the latest commit temporarily just to save plugin zip archive in artifacts. Could you please check that proc macro expansion works on Apple Silicon machines with it?

ortem (changes requested):
Discussed in person with @Undin that the implementation should be changed because `CpuArch.isArm64()` does not work as expected on Apple M1:  `System.getProperty("os.arch")` is `"x86_64"` in virtualized JVM

Undin:
> Discussed in person with @Undin that the implementation should be changed because CpuArch.isArm64() does not work as expected on Apple M1: System.getProperty("os.arch") is "x86_64" in virtualized JVM

Found out that it's not an actual issue because:
- it works even if we try to use x86 native helper on Apple Silicon machine
- JetBrains IDEs for Apple Silicon uses arm64 version of jdk that correctly returns os architecture.

The actual problem was using `x86_64-apple-darwin` toolchain instead of `aarch64-apple-darwin`. That led to an error in `dlopen`

Issue #6505, opened 2020-12-11
Show test output during test


mahkoh:
## Environment

* **IntelliJ Rust plugin version:** 0.3.138.3533-203-nightly
* **Rust toolchain version:** 1.48.0 (7eac88abb 2020-11-16) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.3 (CL-203.5981.166)
* **Operating system:** Linux 5.9.13-arch1-1
* **Macro expansion engine:** new
* **Name resolution engine:** old

## Problem description

println etc. within tests is shown after the test completes but while the test is running.

## Steps to reproduce

```rust
    #[test]
    fn f() {
        println!("xyz");
        std::thread::sleep_ms(10000);
    }
```

sum-elier:
I think you might want to check this question in SO: https://stackoverflow.com/questions/25106554/why-doesnt-println-work-in-rust-unit-tests

mchernyavsky:
@mahkoh Output of `cargo test` for your example looks something like this:

```bash
$ cargo test -- -Z unstable-options --format=json --show-output
    Compiling foo v0.1.0 (/Users/misha/CLionProjects/foo)
    Finished test [unoptimized + debuginfo] target(s) in 0.07s
     Running target/debug/deps/foo-bed43e71a26e9aea
{ "type": "suite", "event": "started", "test_count": 1 }
{ "type": "test", "event": "started", "name": "f" }
{ "type": "test", "name": "f", "event": "ok", "stdout": "xyz\n" }
{ "type": "suite", "event": "ok", "passed": 1, "failed": 0, "allowed_fail": 0, "ignored": 0, "measured": 0, "filtered_out": 0 }
```

So we can't get the test output until the test is finished.

RazgrizHsu:
Can i set the `--nocapture` to the default option?

jmgilman:
It looks like a merge introduced the ability to add `--nocapture` to a test, however, it's not clear where this setting is or how to enable it. Can someone point me in the right direction?

mchernyavsky:
@jmgilman 

1.
<img width="981" alt="Screenshot 2021-09-15 at 20 46 15" src="https://user-images.githubusercontent.com/6079006/133483783-b7764d30-3c99-4f9a-b236-484972982241.png">

2.
[...]

PR #7308, opened 2021-06-04
RUN: Provide an registry key to add `--nocapture` flag to `cargo test` command

mchernyavsky:
Relates to https://github.com/intellij-rust/intellij-rust/issues/6505#issuecomment-853227324.

changelog: Provide an experimental feature to add `--nocapture` flag to `cargo test` command


Undin (commented):
I don't think it should be an experimental feature. Most likely it's a registry option or advanced setting since 212

mchernyavsky:
@Undin Done.

mchernyavsky:
@Undin Looks like advanced settings required the message bundle for the plugin. And since advanced settings are not available on 211 I think it's better to use a registry option for now and introduce advanced settings instead of current registry options after the drop of 211.

mchernyavsky (commented):
bors r+

PR #7310, opened 2021-06-04
Update 212 dependencies and fix incompatibilities

Undin:


src/test/kotlin/org/rust/ide/refactoring/move/RsMoveFileTest.kt
```diff
@@ -422,6 +424,7 @@ class RsMoveFileTest : RsMoveFileTestBase() {
         fn func() {}
     """)
 
+    @IgnoreInPlatform(212)
```
Undin, line 427: @dima74 I've disabled several tests for move refactoring for 2021.2 plaform because they fail.
Could you check them and make the corresponding fixes, please? Note, I don't want to include these fixes in today release

PR #7330, opened 2021-06-10
ANN: improve Change function signature fix 

Kobzol:
This PR improve the `Change signature` quick fix. It now also allows changing parameter types and removing extraneous parameters at the end of the signature (in addition to adding new parameters, which it could do before. A discussion of this feature is present [here](https://github.com/intellij-rust/intellij-rust/pull/6745#issuecomment-858554947).

change-parameter

changelog: Improve the `Change signature` quick fix. It can now also change parameter types and remove extraneous parameters from the end of the signature.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
@@ -11,7 +11,6 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
-import org.rust.cargo.project.workspace.PackageOrigin
 import org.rust.ide.annotator.getFunctionCallContext
 import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.refactoring.changeSignature.*
@@ -20,51 +19,71 @@ import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.isMethod
 import org.rust.lang.core.psi.ext.valueParameters
 import org.rust.lang.core.types.implLookup
+import org.rust.lang.core.types.inference
 import org.rust.lang.core.types.ty.Ty
 import org.rust.lang.core.types.ty.TyUnknown
 import org.rust.lang.core.types.type
+import org.rust.lang.utils.RsDiagnostic
+import org.rust.stdext.mapToSet
+import org.rust.stdext.numberSuffix
+import kotlin.math.max
 
 /**
- * Requires that the number of arguments is larger than the number of parameters.
+ * This fix can add, remove or change the type of parameters of a function.
  */
 class ChangeFunctionSignatureFix private constructor(
-    arguments: RsValueArgumentList,
+    argumentList: RsValueArgumentList,
     function: RsFunction,
-    private val direction: ArgumentScanDirection,
-) : LocalQuickFixAndIntentionActionOnPsiElement(arguments), PriorityAction {
+    private val signature: Signature,
+    private val priority: PriorityAction.Priority? = null
+) : LocalQuickFixAndIntentionActionOnPsiElement(argumentList), PriorityAction {
     private val fixText: String = run {
-        val signature = calculateSignature(function, arguments, direction)
-        val argumentInsertions = signature.withIndex().filter { (_, item) -> item is SignatureMember.Argument }
-
         val callableType = if (function.isMethod) "method" else "function"
         val name = function.name
-        if (argumentInsertions.size == 1) {
-            val index = argumentInsertions[0].index + 1
-            val argument = argumentInsertions[0].value as SignatureMember.Argument
+        val changes = signature.actions.withIndex().filter { (_, item) -> item !is SignatureAction.KeepParameter }
+        val arguments = getEffectiveArguments(argumentList, function)
 
-            val ordinal = "${index}${suffix(index)}"
-            "Add `${renderType(argument.expr.type)}` as `$ordinal` parameter to $callableType `$name`"
+        if (changes.size == 1) {
+            val (index, action) = changes[0]
+            val indexOffset = index + 1
+            val ordinal = "$indexOffset${numberSuffix(indexOffset)}"
+            val parameterFormat = function.valueParameters.getOrNull(index)?.pat?.formatParameter(index)
+
+            when (action) {
+                is SignatureAction.ChangeParameterType -> {
+                    val argument = arguments[action.argumentIndex]
+                    "Change type of $parameterFormat of $callableType `$name` to `${renderType(argument.type)}`"
+                }
+                is SignatureAction.InsertArgument -> {
+                    val argument = arguments[action.argumentIndex]
+                    "Add `${renderType(argument.type)}` as `$ordinal` parameter to $callableType `$name`"
+                }
+                SignatureAction.RemoveParameter -> "Remove $parameterFormat from $callableType `$name`"
+                is SignatureAction.KeepParameter -> error("unreachable")
+            }
         } else {
-            val signatureText = signature.joinToString(", ") {
-                when (it) {
-                    is SignatureMember.Argument -> "<b>${renderType(it.expr.type)}</b>"
-                    is SignatureMember.Parameter -> it.parameter.typeReference?.text.orEmpty()
+            val actions = signature.actions.filter { it !is SignatureAction.RemoveParameter }
+            val signatureText = actions.joinToString(", ") { action ->
+                when (action) {
+                    is SignatureAction.InsertArgument -> "<b>${renderType(arguments[action.argumentIndex].type)}</b>"
+                    is SignatureAction.KeepParameter -> renderType(
+                        function.valueParameters[action.parameterIndex].typeReference?.type ?: TyUnknown
+                    )
+                    is SignatureAction.ChangeParameterType -> "<b>${renderType(arguments[action.argumentIndex].type)}</b>"
+                    SignatureAction.RemoveParameter -> error("unreachable")
                 }
             }
 
-            "<html>Change signature of $name($signatureText)</html>"
+            "<html>Change signature to $name($signatureText)</html>"
         }
     }
 
     override fun getText(): String = fixText
-    override fun getFamilyName(): String = "Add parameter to function"
+    override fun getFamilyName(): String = "Change function signature"
 
     override fun startInWriteAction(): Boolean = false
 
-    override fun getPriority(): PriorityAction.Priority = when (direction) {
-        ArgumentScanDirection.Forward -> PriorityAction.Priority.HIGH
-        ArgumentScanDirection.Backward -> PriorityAction.Priority.NORMAL
-    }
+    override fun getPriority(): PriorityAction.Priority = priority ?: PriorityAction.Priority.NORMAL
 
     override fun invoke(
         project: Project,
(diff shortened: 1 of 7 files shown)


dima74:
@Kobzol Sorry for the long response

> 1. Wording of the fix
>    As with the fix in this PR, we cannot always directly reference the name of the parameter, because the name might be complex (nested patterns etc.). I would suggest to use the name of the parameter if it has a simple binding (`change the type of parameter a ...`) and use its position if not (`change the type of the first parameter ...`).

👍

> 2\. Synergy with other fixes
>     There are already fixes that can change the type of the argument instead, i.e.:
> 
> ```rust
> fn foo(a: &i32) {}
> fn main() {
>     foo(1); // Convert to &i32 using & 
> }
> ```
> 
> I suggest that we somehow use priority to make sure that these fixes are on top, as it's probably much more common to change the type of the argument instead of changing the signature the function.

I agree that fixes for adding reference or mutability should be on top, but I am not sure about fixes which convert number to string and vice versa. In Kotlin and Java "Change type of parameter" fix has higher priority than e.g. "Add `toString()` call"

> 5\. What to do about complex bindings?
> 
> ```rust
> fn foo((a, b): (u32, u32)) {}
> fn main() {
>     foo(true);
> }
> // v
> fn foo((a, b): bool) {} // a bit weird
> ```

Ideally we should check that type of argument can be combined with type of parameter (e.g. paramter has type `(u32, u32)`, argument has type `(i32, i32)`) and offer fix only in that case. But for now we can e.g. just disable quick fix if parameter has such complex type

---

Regarding 3 & 4, consider code like
```rust
fn foo(a: &str, b: &str) {}
fn main() {
    foo(1, 2);
}
```
In Kotlin there are two fixes for each argument, in Java there is single fix that changes types of both parameters. Personally I prefer Java approach. So that there always will be quick-fix which will change count and types of parameters so that they identically match arguments. Something like this:

* argument count = parameters count
  - `Change signature` fix for changing all parameter types
* argument count > parameters count
  * each parameter has a matching argument
    - `Add X as N-th parameter` or `Change signature` fix for adding new parameters
  * otherwise
    - `Change signature` fix for adding new parameters to the end and changing all existing parameter types
* argument count < parameters count
  - `Change signature` fix for removing extra parameters and changing all remaining parameter types

@Kobzol what do you think?

Kobzol:
> Ideally we should check that type of argument can be combined with type of parameter (e.g. paramter has type (u32, u32), argument has type (i32, i32)) and offer fix only in that case. But for now we can e.g. just disable quick fix if parameter has such complex type

Currently the plugin can't even recognize this situation:
```rust
fn foo((a, b): u32) {} // no error, missing E308
```
So it's probably not easy to check it at this moment. On the other hand, to keep "parity" with change signature refactoring, maybe we should just allow this and let the user change the binding later? Change signature refactoring will also currently let you change e.g. `(a, b): (u32, u32)` into `(a, b): u32`.

> In Kotlin there are two fixes for each argument, in Java there is single fix that changes types of both parameters. Personally I prefer Java approach. So that there always will be quick-fix which will change count and types of parameters so that they identically match arguments. Something like this:

Sounds reasonable. Although I'm not sure how to implement it :D Currently this fix is being invoked in `RsDiagnostics` on E0308, which is created in type inference. With the "global" approach that would change all parameters with a different type, I would need to check if the argument type matches the parameter type manually in `RsErrorAnnotator::checkValueArgumentList`. I'm not sure if it's trivial to reproduce the correct logic from type inference here. Maybe we could instead somehow check if the argument was "marked" by E0308 already?

dima74:
> On the other hand, to keep "parity" with change signature refactoring, maybe we should just allow this and let the user change the binding later? Change signature refactoring will also currently let you change e.g. `(a, b): (u32, u32)` into `(a, b): u32`.

Ok, lets do it that way

> I would need to check if the argument type matches the parameter type manually in `RsErrorAnnotator::checkValueArgumentList`.

Maybe use something like `args.inference.diagnostics.any { it is RsDiagnostic.TypeError && it.element in args.exprList }`

Kobzol:
That seems to work, thanks.

I tried refactoring this fix, but then I realized that for the "one-in-all" functionality that you describe, all of the logic will have to be in a single fix. So I will try to modify `ChangeSignatureFix` to handle all of the described cases.

Kobzol:
I have reworked the fix to now include the previous "smart" insertion of parameters in two directions, and also a simple solution that just offers to change the parameters to exactly match the argument list.

Kobzol:
@dima74 I'm not sure what to do about `RsErrorAnnotatorTest::test don't touch AST in other files`. It's currently failing, because in Change signature fix accesses PSI of a remote file to read the parameter types of a function. How should this be solved? The only thing that comes to mind is adding the parameter types to the function stub, but that seems like overkill.

dima74:
> I'm not sure what to do about `RsErrorAnnotatorTest::test don't touch AST in other files`. It's currently failing, because in Change signature fix accesses PSI of a remote file to read the parameter types of a function. How should this be solved? The only thing that comes to mind is adding the parameter types to the function stub, but that seems like overkill.

This is because we access `RsTypeReference::text`:

image

We can either render type reference using e.g. `RsPsiRenderer` or use type of argument (looks like it should be same as type of parameter since we handle `KeepParameter` case) like in `InsertArgument` and `ChangeParameterType` cases.

---

Also some comments:

* There is something strange with intention priorities. Can we make that "Add parameter" and "Change signature" intentions will always be before "Remove redundant arguments"?

image

* Looks like we show three "Change signature" intentions when parameters and arguments types mismatch. I suppose we want to show only one in this case?

image

* Consider this code:
```rust
fn func(a: i32, b: i32) {}

fn main() {
    func("1", 2, 3);
}
```
Both "Change signature" and "Add ... as 1st parameter" intentions are shown. Looks like we should show only "Add ... as 1st parameter"

image

Kobzol:
For this case:
```rust
fn func(a: i32, b: i32) {}

fn main() {
    func("1", 2, 3);
}
```
the fixes `Add first` and `Change signature` have a different meaning:
```rust
fn func(a: &str, b: i32, i: i32) {} // change signature
// vs
fn func(a0: &str, a: i32, b: i32) {} // add first parameter
```
While the signature types will end up the same, the parameter names will be different (which would be important if they were used within the function).

Currently, I try to merge fixes that produce the exact same signature changes. Do you want to change this logic to merge fixes that will result in the same (type) signature? I suppose that sometimes it might make a difference if you want to

1. Add a new first parameter
or
2. Change the type of the first parameter and add a new parameter at the end

For this case:
```rust
fn func(a: i32, b: i32) {}

fn main() {
    func("1", 2, 3);
}
```
again all of the three fixes have a different meaning. Before I tried to ignore the "change signature to match arguments exactly" fix if the "add new parameter(s)" fix is available, but in some cases it resulted in useful fixes being ignored. Maybe to reduce the cound we can do something like "if there are more than two fixes available, ignore one of the directions for adding new parameters"? Therefore there would be at most two fixes. But it seems a bit arbitrary.

dima74:
> Before I tried to ignore the "change signature to match arguments exactly" fix if the "add new parameter(s)" fix is available, but in some cases it resulted in useful fixes being ignored.

Actually I wanted exactly this, not show "change signature to match arguments exactly" fix if the "add new parameter(s)" fix is available. Could you please show some examples where both fixes are useful? Maybe we can think up some heuristics to show both fixes in such cases

---

Regarding case when three fixes are shown
```rust
fn func(i: i32) {}

fn main() {
    func("1", "2");
}
```

Here I think that we shouldn't offer change signature either to `(i32, &str, &str)` or `(&str, &str, i32)` since in both cases in resulting code arguments and parameters will mismatch (three parameter vs two arguments)

Kobzol:
> Actually I wanted exactly this, not show "change signature to match arguments exactly" fix if the "add new parameter(s)" fix is available. Could you please show some examples where both fixes are useful? Maybe we can think up some heuristics to show both fixes in such cases

For example, in this case:
```rust
fn foo(a: i32, b: i32) {}

fn main() {
    foo(0, false, true);
}
```
We get:
```rust
foo(i32, <b>bool</b>, <b>bool</b>, i32) // insert forward
foo(i32, i32, <b>bool</b>, <b>bool</b>) // insert backward
```
but the (arguably most useful) fix here (`foo(i32, <b>bool</b>, <b>bool</b>)`) would not be offered.

This case could be made more extreme. Any unmatching parameter type will just generate more parameters in the fix:
```rust
fn foo(a: i32, b: i32) {}

fn main() {
    foo(false, false, true);
}
//v
foo(<b>bool</b>, <b>bool</b>, <b>bool</b>, i32, i32)
foo(i32, i32, <b>bool</b>, <b>bool</b>, <b>bool</b>)
```
The `change signature fix to match arguments` is more stable in this way, because it will always match the number of arguments.

> Here I think that we shouldn't offer change signature either to (i32, &str, &str) or (&str, &str, i32) since in both cases in resulting code arguments and parameters will mismatch (three parameter vs two arguments)

Oh, that's right. So that could be one heuristic by which to filter the fixes, the signature has to match (at least the number of arguments) after the fix would be performed?

I *think* that it could solve both of these problems, if we used something like this:
1) Create the `Change signature` fix
2) Try to generate `Add parameter` fixes, but only use them if they would result in a number of parameters that matches the argument count.
3) If any `Add parameter` fix was used, ignore the `Change signature` fix

dima74:
> Oh, that's right. So that could be one heuristic by which to filter the fixes, the signature has to match (at least the number of arguments) after the fix would be performed?

Yes, exactly

> I _think_ that it could solve both of these problems, if we used something like this:
> 
> 1. Create the `Change signature` fix
> 2. Try to generate `Add parameter` fixes, but only use them if they would result in a number of parameters that matches the argument count.
> 3. If any `Add parameter` fix was used, ignore the `Change signature` fix

Yes, something like this. We can ever determine which fix we should show without creating the fixes - looks like we should show "Add parameter" fix only if "argument count > parameters count" and "each parameter has a matching argument"

Kobzol:
Ok. I changed the heuristic. It should remove cases where the `Add parameter(s)` fix would result in a horrible mess.

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -217,25 +331,28 @@ private class PeekableIterator<T>(private val iterator: Iterator<T>) {
...
-    if ((number % 100) in 11..13) {
-        return "th"
-    }
-    return when (number % 10) {
-        1 -> "st"
-        2 -> "nd"
-        3 -> "rd"
-        else -> "th"
-    }
+private sealed class SignatureAction {
```
dima74, line 347: Do I understand correctly that we take desired signature (based on arguments types), create diff using this class by comparing desired signature with current signature, and then apply diff to function? Wouldn't it be easier if we take desired signature and directly modify function signature (without creating diff using `Keep`/`Insert`/`Remove` actions)?

Kobzol: It's a list of actions that can be taken to reach some target signature. Even though a target signature might be the same, you can reach it with different actions. This was also a "problem" in the `Change signature` refactoring, which is used by this fix. In the refactoring I need to be able to identify e.g. if the user moved a parameter or deleted it and then created a new one, and here the situation is the same.

A very simple example:
```
source signature: (u32, u32)
target_signature: (u32, u32, u32)

actions_1: (keep, insert, keep)
actions_2: (keep, keep, insert)
```
Just storing the target signature would lose information about how to reach it. The actions affect the wording of the fix and also the semantics (as we have discussed a few comments above). I need to be able to trace the actions that have to be performed to change the signature, otherwise the `Change signature` refactoring will not work properly.

An action at index `i` says how to build the `i-th` member of the target signature:
1) KeepParameter(source parameter) -> use the parameter from a specified index the source signature
2) InsertArgument(source argument) -> create a new parameter from the specified argument
3) RemoveParameter -> remove the parameter
4) ChangeParameterType(source argument) -> change the parameter type, based on the specified argument

The `RemoveParameter` differs from the others, but there's an invariant that it can only be created at the end of the action list. This makes the implementation simpler, otherwise removing from the middle would complicate stuff a lot, but we do not need that.

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -74,53 +90,94 @@ class ChangeFunctionSignatureFix private constructor(
...
+                    config.parameters.add(index, Parameter(factory, name, ParameterProperty.fromItem(typeReference)))
+                    config.additionalTypesToImport.add(expr.type)
+                    parameterIndex++
+                }
+                is SignatureAction.RemoveParameter -> config.parameters.removeAt(parameterIndex)
+                is SignatureAction.ChangeParameterType -> {
+                    val original = config.parameters[parameterIndex]
+                    val expr = arguments.getOrNull(item.argumentIndex) ?: continue
+                    val typeText = expr.type.renderInsertionSafe()
+                    val fragment = RsTypeReferenceCodeFragment(project, typeText, context = args)
```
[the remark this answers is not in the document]
Kobzol: Oh, you're right, we can. Because there is `config.additionalTypesToImport.add(expr.type)`.

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
dima74: Ok, it is reasonable, thanks for the explanation!

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -74,53 +93,91 @@ class ChangeFunctionSignatureFix private constructor(
...
-        ): ChangeFunctionSignatureFix? {
-            if (function.containingCrate?.origin != PackageOrigin.WORKSPACE) return null
-
-            val signature = calculateSignature(function, arguments, direction)
-            val parameterInsertions = signature.filterIsInstance<SignatureMember.Parameter>()
-
-            // There are parameters that were not matched by any argument
-            if (function.valueParameters.size < parameterInsertions.size) {
-                return null
+        fun createIfCompatible(args: RsValueArgumentList, function: RsFunction): List<ChangeFunctionSignatureFix> {
```
dima74, line 142:
```suggestion
        fun createIfCompatible(arguments: RsValueArgumentList, function: RsFunction): List<ChangeFunctionSignatureFix> {
```

src/main/kotlin/org/rust/ide/annotator/fixes/ChangeFunctionSignatureFix.kt
```diff
@@ -74,53 +93,91 @@ class ChangeFunctionSignatureFix private constructor(
...
+            if (!RsChangeSignatureHandler.isChangeSignatureAvailable(function)) return emptyList()
+            val ctx = args.getFunctionCallContext() ?: return emptyList()
```
dima74, line 144:
```suggestion
            val context = args.getFunctionCallContext() ?: return emptyList()
```

dima74 (approved):
Thanks!
Please also squash commits and we can merge

Kobzol:
bors r=dima74

Kobzol:
bors retry

Issue #7332, opened 2021-06-11
Wrong no named argument error highlighting with `format_args_capture` feature


josem435:
## Environment

* **IntelliJ Rust plugin version:** `0.4.148.3911-211`
* **Rust toolchain version:** `1.54.0-nightly-2021-05-11-x86_64-pc-windows-msvc`
* **IDE name and version:** `PyCharm 2021.1.2 (Community Edition) (PC-211.7442.45)`
* **Operating system:** `Windows 10 10.0.19042 Build 19042`
* **Macro expansion engine:** `new`
* **Name expansion engine:** `new`

## Problem description

`format_args_capture` feature allows you to pass implicit named args into formatting macros. The inspector doesn't recognize this feature.

![image](https://user-images.githubusercontent.com/77453282/121678944-b8982b80-caaf-11eb-90cc-86ac9ea05281.png)

## Steps to reproduce

```rust
#![feature(format_args_capture)]

fn foo() {
    let output_string = String::from("I am a string");
    println!("{output_string}");
}
```

bugproof:
This shouldn't be closed. I still get those warnings.

vault-thirteen:
I am still getting this bug in the latest version of CLion with the latest verison of Rust plug-in.

PR #7334, opened 2021-06-11
ANN: fix false-positive highlighting on implicit named arguments in format macro annotations

Kobzol:
This PR adds basic support for the `format_args_capture` unstable feature, just so that we avoid false positives.

Later we can add proper support for resolving format macro arguments with a reference contributor that would resolve arguments either to named parameters or to local bindings. It would also display unresolved reference errors if no argument is found when `format_args_capture` is enabled.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7332

changelog: Do not warn about missing format macro arguments if `format_args_capture` feature is enabled.


vlad20012 (approved):
LGTM
Rebase is needed

Kobzol:
bors r=vlad20012

PR #7345, opened 2021-06-14
INT: enable SubstituteAssociatedTypeIntention for type aliases

Kobzol:
This PR adds support for substituting type aliases to `SubstituteAssociatedTypeIntention`. This was previously disabled in https://github.com/intellij-rust/intellij-rust/pull/6032 due to problems with generic parameters (https://github.com/intellij-rust/intellij-rust/issues/6030).

I used the approach hinted by @vlad20012 [here](https://github.com/intellij-rust/intellij-rust/issues/6030#issuecomment-685092669), it seems to work fine.

However, there is still a problem with paths using type aliases that are in expression context and that have inferred type arguments (see the commented test). In this situation, the resolved type is just a new type argument placeholder (e.g. `T`), but it's not resolved to the actual type yet (https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/resolve/ref/RsPathReferenceImpl.kt#L234).

I need a way to somehow get the inferred result of the type from type inference. I can get the type of the function call `Foo::new` (`fn() -> S<i32>`), but I need the type of `Foo` (`S<i32>`). Is there some existing way of doing that?

changelog: The intention that enables you to substitute associated types now works on all type aliases.


vlad20012:
I tried to make this work with #7401.
It seems much more tricky than I expected. And, well, to be honest, I'm not sure I completely understand what I am doing. But here is the patch:

[temp.zip](https://github.com/intellij-rust/intellij-rust/files/6710684/temp.zip)

Kobzol:
> And, well, to be honest, I'm not sure I completely understand what I am doing

That's how I always feel when interacting with the type inference machinery :sweat_smile:

Thank you! It *seems*™ to work.

Btw, it's a bit ironic that my main motivation for "more generic type aliases", which was this intention, ends up using the old API (`type.renderInsertionSafe(skipUnchangedDefaultTypeArguments = true)`) instead of the new PSI API :D But it seems to work :man_shrugging:

vlad20012:
> **type.renderInsertionSafe**

Hmm, indeed. But why? I thought we want to perform single-step expansion of an alias. It's possible with `typeRef.getStubOnlyText(subst)`:

[temp2.zip](https://github.com/intellij-rust/intellij-rust/files/6716051/temp2.zip)

But one test is failing with this patch because type quals rendering isn't implemented... I'll add it to #7393, let's wait for it

Kobzol:
It's just because this was implemented before #7393 :) Let's resolve it once that gets merged.

vlad20012:
That API also existed before #7393 =)

Kobzol:
I know, it's just that before I didn't realize that type rendering might be a bad idea when you already have access to a type reference :)

vlad20012:
Well, let's apply the patch, fix the test and merge this?

Kobzol:
bors r=vlad20012

Kobzol:
bors retry

PR #7346, opened 2021-06-14
COMP: add special case for String in MatchPostfixTemplate

Kobzol:
This PR adds a special case for `String` in `MatchPostfixTemplate`,  as mentioned [here](https://github.com/intellij-rust/intellij-rust/pull/6825#issuecomment-858592723), because the default is not very useful. This could possibly be applied to other types too, for example `&str`.

It also adds simple infrastructure for adding custom match generation for selected types.

changelog: Add special case for Strings in MatchPostfixTemplate.

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
@@ -14,6 +14,13 @@ import org.rust.ide.inspections.fixes.AddWildcardArmFix
 import org.rust.ide.utils.checkMatch.checkExhaustive
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.descendantsOfType
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.resolve.knownItems
+import org.rust.lang.core.types.ty.Ty
+import org.rust.lang.core.types.ty.TyAdt
+import org.rust.lang.core.types.ty.TyReference
+import org.rust.lang.core.types.ty.TyStr
+import org.rust.lang.core.types.type
 import org.rust.openapiext.buildAndRunTemplate
 import org.rust.openapiext.createSmartPointer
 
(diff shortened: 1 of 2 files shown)


vlad20012 (approved):

vlad20012:
What do you think about more special cases, for example, for smart pointers? I think it's possible to make a more or less generic implementation for smart pointers by using dereference (`ImplLookup.coercionSequence()`) or finding `AsRef`
 implementations (`isTraitWithTySubstImplForActual`/`ConvertToRefTyFix`).

Kobzol:
I wouldn't do it in a completely generic way, because it's not clear to me when should `AsRef` or something like that be applied. For `String` specifically, it doesn't make a lot of sense to pattern match against it directly, but for your own struct, it might be actually what you want (and calling `AsRef` on it might produce something that doesn't make sense for matching, even if your struct implements it).

I suppose that we could create some heuristic like "if the type is not an enum AND its fields are private, so you cannot pattern match against it directly AND it implements AsRef", etc. But I'm not sure if it's worth it to do it in such a complex way.

But if you meant that we should special-case several "known" types, like `Box`, `Rc`, `Arc`, etc., then that should be fine, and in that case we probably don't even need any inference.

vlad20012:
By "generic implementation" I meant "works for any custom type", but you're right it doesn't work without the heuristic.

I also realized that this doesn't work with containers like `Cell<T>` because they don't implement Deref or AsRef and require special method invocation.

About the heuristic, I don't think we should restructure structs without public fields at all. Why would someone need this? By the way, it should be easy to check that a struct doesn't have visible fields: we already have a similar method `RsFieldsOwner.canBeInstantiatedIn(mod)` that checks that *all* struct fields are visible.

Anyway, these ideas are for further PRs, this one looks complete as is

Kobzol:
I know that it wouldn't be too hard to check if a struct can be used in a match pattern, I just thought that the resulting heuristic might be a bit too complicated.

Btw, I realized that it might be better to generate this:
```rust
match string.as_str() {
    "" => {},
    _ => {}
}
```
instead of just
```rust
match string.as_str() {
    "" => {}
}
```
so that the code compiles right away? And also since the infrastructure is already there, I would add the same special case for `&str`, in addition to `String`.

vlad20012:
> I know that it wouldn't be too hard to check if a struct can be used in a match pattern, I just thought that the resulting heuristic might be a bit too complicated.

Oh, I forget to say that I think you're right and it likely doesn't worth it.

> so that the code compiles right away?

:+1: 

> And also since the infrastructure is already there, I would add the same special case for &str, in addition to String (in this PR, if you don't mind).

:+1: :+1:

Kobzol:
> so that the code compiles right away?

Oh, that was actually already done by the code, I'm thinking ahead :D

Kobzol:
With the addition of `&str`, I made a few other changes, because for `&str` we couldn't generate `.as_str()`, so I created an additional processor. Because of that, I'll not merge it yet so that you can take a look.

src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
```diff
@@ -32,31 +39,83 @@ class MatchPostfixTemplate(provider: RsPostfixTemplateProvider) :
...
+private open class StringLikeMatchProcessor : MatchProcessor() {
+    open fun expressionToText(expression: RsExpr): String = expression.text
+
+    override fun createMatch(factory: RsPsiFactory, expression: RsExpr): RsMatchExpr {
+        val exprText = expressionToText(expression)
+        return factory.createExpression("match $exprText {\n\"\" => {} }") as RsMatchExpr
+    }
+
+    override fun normalizeMatch(matchExpr: RsMatchExpr) {
+        AddWildcardArmFix(matchExpr).invoke(matchExpr.project, matchExpr.containingFile, matchExpr, matchExpr)
```
vlad20012, line 112: Why not just produce a `match` with an additional wildcard arm in `createMatch`?

Kobzol: Hmm, I was a bit worried about formatting, but it works fine. I changed it to generate `_` directly.

PR #7348, opened 2021-06-15
Split TOML plugin into modules

avrong:
Introduces separation into modules:
 - `intellij-toml` with `plugin.xml` and icons
 - `intellij-toml:core` with `toml-core.xml`, toml grammar and basically code insight
 - `intellij-toml:schema` which is not present now, but soon will include all json-schema related stuff and be enabled by platform only if it's dependencies are satisfied

### TODO
- [x] Update `CONTRIBUTING.md` to include info about new modules


Undin (changes requested):
- let's move `intellij-toml/src/main/resources/icons` icons to core module as well - they are used by `TomlIcons`
- looks like `setPlugins(project(":intellij-toml"))` doesn't add all transitive dependencies as a result `:toml` module cannot be compiled. I think it should be fixed on `gradle-intellij-plugin` side but as temporary workaround it seems that we can add `compileOnly(project(":intellij-toml:core"))` and `testCompileOnly(project(":intellij-toml:core"))` into `:toml` module

intellij-toml/src/main/resources/META-INF/plugin.xml
```diff
@@ -18,43 +18,5 @@
 
     <depends>com.intellij.modules.lang</depends>
```
avrong, line 19: I'm not sure if we need to have this dependency here

Undin: We need it. Otherwise, we can't rely on that IDE contains classes required for language support. For example, Code With Me client doesn't have `com.intellij.modules.lang` module. As a result, you can't install Toml plugin on CWM client and it's desired behavior

Undin: Oh, probably I misunderstood you. We actually don't need this `depends` here because we have it in `toml-core.xml`

build.gradle.kts
```diff
@@ -448,12 +448,16 @@ project(":toml") {
...
         implementation("org.eclipse.jgit:org.eclipse.jgit:5.9.0.202009080501-r") { exclude("org.slf4j") }
         implementation("com.vdurmont:semver4j:3.1.0")
+
+        implementation(project(":"))
+        implementation(project(":common"))
         testImplementation(project(":", "testOutput"))
         testImplementation(project(":common", "testOutput"))
+
+        compileOnly(project(":intellij-toml:core"))
+        testCompileOnly(project(":intellij-toml:core"))
```
Undin, line 460: Let's add TODO to drop it when gradle-intellij plugin fix it

Undin (approved):
LGTM
Let's update CONTRIBUTING.md and fix minor comments

intellij-toml/src/main/resources/META-INF/plugin.xml
avrong: Then probably the same can be applied to `plugin.xml` for intellij-rust

Undin: Yep. Would you like to fix it in separate PR?

Undin: Let's drop it not to produce `[gradle-intellij-plugin :intellij-toml:verifyPlugin] Duplicated dependency on 'com.intellij.modules.lang'` warning - we have necessary dependency declaration in core xml

CONTRIBUTING.md
```diff
@@ -198,6 +198,10 @@ The current Rust plugin modules:
 * `:js` - interop with JavaScript language
 * `:ml-completion` - integration with [Machine Learning Code Completion](https://github.com/JetBrains/intellij-community/tree/master/plugins/completion-ml-ranking) plugin
 
+The current Toml plugin modules:
+* `:intellij-toml` - contains plugin meta-information & icons
```
Undin, line 202: Also, this module is used as a root module to build toml plugin. Let's unify description of this module and `:plugin` one because they used for the same purposes

Undin (approved):
👍 
bors r+

PR #6502, opened 2020-12-11
ANN: highlight markdown in rust docs based on PSI

vlad20012:
Replaces old lexer-based highlighting with a new PSI-based highlighting.

The PR doesn't change the highlighting rules.

PR #7349, opened 2021-06-15
Fix rustdoc undecoration

vlad20012:
Fix some corner cases with rustdoc undecoration (removing `/// `, ` * `, etc). I guess that almost nobody will be affected by this change because almost all use default `/// ` doc prefixes that work file already.


vlad20012:
Looks like it conflicts with #6502

bors r-

PR #7355, opened 2021-06-16
Move `WSL_TOOLCHAIN` const into `RsExperiments`

Undin:
Also, check this feature in `RsWslToolchainProvider`

Just a small cleanup


mchernyavsky (approved):
bors r+

Issue #7338, opened 2021-06-11
Support implicit named arguments in formatting macros


Undin:
RFC: https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html

Small example:
```rust
let (person, species, name) = ("Charlie Brown", "dog", "Snoopy");

// implicit named argument `person`
print!("Hello {person}");

// implicit named arguments `species` and `name`
format!("The {species}'s name is {name}.");
```

Note, `format_args_capture` is current feature name in rustc 1.54.0-nightly

* [ ] don't produce false-positives for name arguments #7332
* [ ] name resolution
* [ ] completion
* [ ] take into account in analysis (unused variables, for example)

Kobzol:
I started working on name resolution, I'll send a PR soon.

Kobzol:
@Undin I wanted to provide a reference provider that would resolve named arguments:
```rust
println!("{foo}", foo = 1); // clicking on {foo} would resolve to `foo =`

let bar = 1;
println!("{bar}"); // clicking on {bar} would resolve to `let bar`.
```
I started by refactoring the code from the format macro annotator, which basically parses the formatting string with a regex and then applies highlighting to it. But now that I see all of the things that could be supported (name resolution, completion, dataflow analysis, unresolved reference etc.), I wonder if it would be better to actually parse the formatting string into proper PSI elements, so that we can better provide these features?

So that e.g. in `println!("{foo}", foo = 1);` the argument `foo = 1` would be a named PSI element and `{foo}` would be another PSI element that would resolve to it. Maybe resolving to named arguments is overkill, but even for resolving to local bindings for implicit named arguments it could be easier with proper PSI parsing.

Is that the proper solution? Or should I stick to regex-based handling of the format string?

baadc0de:
if `#![feature(format_args_capture)]` the 1.58.0 compiler now emits a warning:

> warning: the feature `format_args_capture` has been stable since 1.58.0 and no longer requires an attribute to enable

intellij-rust should detect that `#![feature(format_args_capture)]` is not needed anymore.

kolloch:
Is there a workaround for rust `stable` on `1.58.x` to make intellij aware that format_args_capture is allowed? Thank you.

Undin:
Actually, PSI implementation looks more promising because it allows:
- not use custom approaches during any analysis (from name resolution and completion and to `RsLivenessInspection` inspection) and just use psi for it
[...]

PR #7356, opened 2021-06-16
ANN: highlight format templates in custom macros

vlad20012:
Apply highlighting from #3869 and #4969 to custom macro bodies (#4640)

image

changelog: Provide highlighting of [format string literals](https://doc.rust-lang.org/std/fmt/) in any macros that expand to macros like println, write, format, etc

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt
@@ -67,7 +67,8 @@ class RsMacroExpansionHighlightingPass(
         RsEdition2018KeywordsAnnotator(),
         RsHighlightingAnnotator(),
         RsHighlightingMutableAnnotator(),
-        RsCfgDisabledCodeAnnotator()
+        RsCfgDisabledCodeAnnotator(),
+        RsFormatMacroAnnotator(),
     )
 
     @Suppress("UnstableApiUsage")
--- a/src/test/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotatorTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/RsFormatMacroAnnotatorTest.kt
@@ -6,6 +6,7 @@
 package org.rust.ide.annotator
 
 import com.intellij.ide.annotator.BatchMode
+import org.rust.ExpandMacros
 import org.rust.ProjectDescriptor
 import org.rust.WithStdlibRustProjectDescriptor
 import org.rust.ide.colors.RsColor
@@ -512,6 +513,19 @@ If you intended to print `{` symbol, you can escape it using `{{`">{</error>"###
     fun `test panic with single literal`() = checkErrors("""
         fn main() {
             panic!("{}");
+        }
+    """)
+
+    @ExpandMacros
+    fun `test custom macro`() = checkErrors("""
+        $implDisplayI32
+
+        macro_rules! as_is { ($($ t:tt)*) => {$($ t)*}; }
+        fn main() {
+            as_is! {
+                println!("<FORMAT_SPECIFIER>{}</FORMAT_SPECIFIER>", 1);
+                println!("", <error descr="Argument never used">1</error>);
+            }
         }
     """)
 }


Kobzol:
That's pretty cool!

I'm not sure exactly how this works, but I tried to remove the special case for `log` in `RsFormatMacroAnnotator:519` to see if it will work, and it doesn't.
```rust
log::info!("Hello {}", 1);
```
I would expect that this gets formatting annotation, because it calls `format_args!` somewhere deep down (https://github.com/intellij-rust/intellij-rust/pull/7267/files#diff-495620f02decf91610a63dd5254ec4f2f97fa1f4d5e8b2ae8d14a99445588af0R451). The `log!` macro doesn't seem to be resolved by the plugin though.

vlad20012:
About `log::info!`, I checked it before sending the PR, and it looks like it doesn't work because of the special parsing of that macro. See these lines:

https://github.com/intellij-rust/intellij-rust/blob/435e6a71791f2e3da4e70b5bdeff43d3670b0b0b/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt#L126

https://github.com/intellij-rust/intellij-rust/blob/435e6a71791f2e3da4e70b5bdeff43d3670b0b0b/src/main/kotlin/org/rust/ide/annotator/RsMacroExpansionHighlightingPass.kt#L77

We highlight (this way) only macros that have `RsMacroArgument` child (but the log macro has `RsLogMacroArgument` instead).

> I'm not sure exactly how this works

I can explain! It's pretty simple.

`TextEditorHighlightingPass` is the most low-level API for "automatic" highlighting. There is a highlighting pass that runs annotators, a highlighting pass that runs inspections, a highlighting pass that highlights injected code, etc. We use a highlighting pass API for macro highlighting mostly because it runs in parallel with other highlighting passes (when annotators are invoked sequentially in a single thread). 
Ok, so in `RsMacroExpansionHighlightingPass.doCollectInformation` we have a file that we're going to highlight. We collect all macro calls in that file and associate them with their expansion (`PreparedMacroCall`). Then we instantiate some annotators (`createAnnotators()`) and run these annotators on _macro expansions_. Then we _map_ text ranges of the resulting annotations (`mapAndCollectAnnotation()`) and eventually apply them to the editor (`doApplyInformationToEditor`). And we do it recursively for all nested macros.

Kobzol:
Thanks for the explanation! I looked at the codebase to see whether we could actually remove the log macro handling altogether (even from the parser etc.). It seems to me that the only real usage of it is in `RsFormatMacroAnnotator`, which needs access to the identifier (`println!("{foo}", foo = 1);`).

It's probably worth it to keep the special parsing to get the identifier though. Actually, I think that maybe we should parse log/formatting macros even [more](https://github.com/intellij-rust/intellij-rust/issues/7338#issuecomment-859924434) :D To support implicit named arguments.

When I naively tried this:
```kotlin
if ((macro.macroArgument ?: macro.logMacroArgument)?.textRange?.intersects(restrictedRange) != true) continue
...
if (macroArgument == null && logMacroArgument == null) return null
```
the highlighting seems to work, but I have no idea if that doesn't break anything else.

vlad20012:
> It's probably worth it to keep the special parsing to get the identifier though.

Log macro and format macro are different special cases (`RsLogMacroArgument` vs `RsFormatMacroArgument`). We can get rid of a special case for log macros without degrading highlighting (now), isn't it? Since log macros eventually expand to `format_args!` invocation that is parsed as `RsFormatMacroArgument`

Kobzol:
Yes, that's true (should I send a PR?). Then the question becomes if the macro highlighting pass should also support formatting macros :)

Kobzol:
Oh, nevermind, I tested it and it actually works even without the `if (macroArgument == null && formatMacroArgument == null) return null` hack. I think that it's time to say good-bye to `LogMacroArgument` then.

vlad20012:
> should I send a PR?

As you with, as usual :) Otherwise, I can do it.

> Then the question becomes if the macro highlighting pass should also support formatting macros :)

We should at least parse built-in format macros: `format_args!` and `format_args` and `format_args_nl`... But anyway I would not be so radical, since "honest" macros handling is still not so feature-rich as a native PSI. For example, intentions doesn't work in macro bodies (but I have a branch for this!), and inspections too, and maybe some other features, I'm not sure.

Kobzol:
Ok, got it. I think that keeping formatting macros is a good idea, but log macros seem kinda overkill with the new macro expansoin system. I have a branch prepared where special-cased log macros are removed, I will send a PR once I figure out all of the tests that fail because of that :)

PR #7358, opened 2021-06-17
GRAM: remove log macro argument special case

Kobzol:
Due to recent improvements in macro expansion technology (:tada: @vlad20012), log macros can now be expanded without a problem and do not need any special casing any more (hopefully).

https://github.com/intellij-rust/intellij-rust/pull/7267 removed special casing of type inference, this PR removes the log macro special case altogether, even from the parser, so log macros are now parsed as any other macro. Let's see if any tests fail on CI.

I implemented this PR based on discussion [here](https://github.com/intellij-rust/intellij-rust/pull/7356#issuecomment-863181583). I have an additional question: for supporting implicit format arguments (https://github.com/intellij-rust/intellij-rust/issues/7338#issuecomment-859924434), we might need to parse the format string as PSI. If we decide to do that and we somehow parse format strings as PSI, will it be possible to map this PSI back into e.g. log macro format string, in the same way that macro highlighting works?

Depends on: https://github.com/intellij-rust/intellij-rust/pull/7356

changelog: Remove special handling of log macros in the parser. Log macros can now be resolved via the macro expansion machinery.


vlad20012:
> for supporting implicit format arguments (#7338 (comment)), we might need to parse the format string as PSI. If we decide to do that and we somehow parse format strings as PSI, will it be possible to map this PSI back into e.g. log macro format string, in the same way that macro highlighting works?

Yep. Highlighting doesn't require additional work, but if we'll resolve these identifiers to pat bindings, this will require some additional changes since `RsMacroBodyReferenceDelegateImpl` works only for identifiers at the moment.

Btw I'm not sure we need PSI in string literals, a contributed reference should be enough.

Kobzol:
> Btw I'm not sure we need PSI in string literals, a contributed reference should be enough.

I started implementing this, and for basic name resolution this should be indeed enough. But if we want to also support completion, dataflow analysis etc., I'm not sure how practical it would be.

vlad20012:
@Kobzol Regarding dataflow analysis I just came up with one idea. We could _expand_ such macros with implicit format arguments to macros with explicit format arguments (using `BuiltinMacroExpander`). And then somehow use such expansions everywhere in the analysis. Hmm

Kobzol:
That could simplify analysis, if we would have all of the implicit arguments expanded in an unified way. I'm not sure if completion could also work like this though.

vlad20012 (approved):
Don't forget to increment `RustParserDefinition.PARSER_VERSION`

Kobzol:
bors r=vlad20012

src/main/grammars/RustParser.bnf
```diff
@@ -1528,9 +1528,6 @@ AsmMacroOptions ::= 'options' '(' <<comma_separated_list AsmMacroOption>> ')'
 //noinspection BnfUnusedRule
 AsmMacroOption ::= identifier
 //noinspection BnfUnusedRule
-LogMacroArgument ::=
-  <<any_braces (('target' ':' Expr ','?)? (<<comma_separated_list FormatMacroArg>>)?)>>
```
matklad, line 1532: This makes my heart sing 😊

PR #7362, opened 2021-06-18
ANN: introduce a color for `std::fmt` format specifiers

vlad20012:
As discussed in https://github.com/intellij-rust/intellij-rust/pull/7243#issuecomment-863359497, rename `FORMAT_SPECIFIER` color to `FORMAT_PARAMETER` and introduce `FORMAT_SPECIFIER` color

https://user-images.githubusercontent.com/3221931/122535107-e63f2080-d02b-11eb-9554-b525f69b9690.mp4


Kobzol (approved):
LGTM.

The name of a named parameter, e.g. `foo` in `{foo}` is probably not a "specifier" per se, but I don't think that we should highlight it in a different color or introduce a special name for it.
Although, once implicit named format arguments are supported, we could highlight it as a "path/reference to a variable".

vlad20012:
bors r=Kobzol

PR #7366, opened 2021-06-18
CI: properly push changes in repository

Undin:
Previously, we tried to push into repo via `https://{actor}:{token}@github.com/{repo}.git` url. Found out that it doesn't work as expected and token provided in `actions/checkout` is used instead.
Now we:
- set proper token during checkout
- use [intellij-rust-bot](https://github.com/intellij-rust-bot) [machine account](https://docs.github.com/en/github/site-policy/github-terms-of-service#3-account-requirements) to generate PAT and push such changes into repo


lancelote:
bors merge

Issue #5080, opened 2020-03-11
Provide API to get type alias for non TyAdt types


Undin:
The current implementation (introduced in #4243) supports only TyAdt (i.e. structs and enums) and doesn't provide type alias for other types (for example, for primitive types).
It leads to inconsistent behavior for actions that generate some new code when they work with primitive and non-primitive types
* #4967
* #5044

Issue #5044, opened 2020-03-04
"Generator Constructor" doesn't use type alias


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.118.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.41.0 (5e1a79984 2020-01-27) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 EAP (CL-201.5616.31)
* **Operating system:** macOS 10.15.3

## Problem description

"Generator Constructor" action always uses original type instead of the alias. I guess it is #4944 regression.

## Steps to reproduce

```rust
type Coordinates = (u32, u32);

struct System {
    point: Coordinates,
}
```

call "Generator Constructor" on `System`

```rust
impl System {
    pub fn new(point: (u32, u32)) -> Self {
        System { point }
    }
}
```

expected

```rust
impl System {
    pub fn new(point: Coordinates) -> Self {
        System { point }
    }
}
```

Kobzol:
You're right, this is indeed a regression caused by #4944. It is caused by using `substAndGetText` on the type reference instead of just taking its raw `text`. This is needed to substitute bound type parameters on an existing impl block.

There is a problem though, after the substitution we can no longer access the `typeReference` text. And since aliases only work on `TyAdt` so far, it will not work for tuples and other types :-(
With `TyAdt` it should work:

```rust
struct S;
type Coordinates = S;

struct /*caret*/System {
    point: Coordinates,
}
// generates
impl System {
    pub fn new(point: Coordinates) -> Self {
        System { point }
    }
}
```
@Undin what do you think?

Undin:
@Kobzol I think we need to extend current API with aliases from `TyAdt` to `Ty` (or at least for something less specific). And use this API to make proper fix

Issue #4979, opened 2020-02-16
Wrong type is imported when generating trait implementation


Kobzol:
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.SNAPSHOT-193-dev (#680fb11)
* **Rust toolchain version:** 1.41.0 (5e1a79984 2020-01-27) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2019.3.3 (CL-193.6015.37)
* **Operating system:** Linux 5.0.0-37-generic

## Problem description
When you generate a method implementation for the `fmt::Display` trait, it imports a wrong type and the resulting code does not compile.

## Steps to reproduce
```rust
use std::fmt;

struct S;
impl fmt::Display for S {
    /*caret*/
}
```
generates this code after `ImplementMembersFix` is run:
```rust
use std::fmt;
use std::fmt::{Formatter, Error};

struct S;
impl fmt::Display for S {
    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
        unimplemented!()
    }
}
```
which does not compile. `Error` is imported instead of `Result`.

Kobzol:
I looked at `RsImportHelper` and there seem to be two issues:
1) `Result` is put into the `toQualify` set in `TypeReferencesInfo`, so it's not imported
2) `Error` is extracted to be imported from the `Result`, although this probably shouldn't happen, as type aliases should stop the import propagation.

I can try to fix 2), but for 1) the import machinery seems a bit complex to me, I'd need some pointers where to look to fix this.

Kobzol:
@Undin I fixed the second problem in the attached PR and found the cause of the first problem. Type aliases that share a name with something which is in Rust's prelude are not auto-imported correctly, because they are stripped here: https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/ide/inspections/import/RsImportHelper.kt#L146

I'm not sure how to proceed here, as this seems to have deeper implications :)

Kobzol:
@Undin Just a random idea: it seems to me that the current approach of auto importing after introducing some new code via refactorings/intentions does not always work so well and I think that it's partly caused by the fact that the insertion loses information about the types. For example here:
```rust
struct S;
impl fmt::Display for S {
    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
        unimplemented!()
    }
}
```
[...]

Issue #4967, opened 2020-02-14
"Implement Methods" expands primitive type aliases


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.2.117.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.41.0 (5e1a79984 2020-01-27) x86_64-apple-darwin
* **IDE name and version:** CLion 2020.1 EAP (CL-201.4865.10)
* **Operating system:** macOS 10.15.3

## Problem description

Unfortunately, #4938 does not fix #3082 as primitive type aliases are still expanded.

## Steps to reproduce

```rust
pub struct R;
pub type U = i32;
pub trait T {
    fn f() -> (R, U);
}
struct S;
impl T for S {}
``` 

Invoke "Implement Methods"

```rust
pub struct R;
pub type U = i32;
pub trait T {
    fn f() -> (R, U);
}
struct S;
impl T for S {
    fn f() -> (R, i32) {
        unimplemented!()
    }
}
```

PR #7368, opened 2021-06-19
TY: introduce TyAlias to allow aliasing for any type

Kobzol:
This PR attempts to introduce a new type called `TyAlias`, which represents an alias of an arbitrary type. This removes the previous special case that only supported type aliases for ADT types (structs, enums, etc.).

This should (eventually) help resolve several issues with type aliases, mainly in code generation, auto import and type rendering/hinting. Some of them are mentioned in https://github.com/intellij-rust/intellij-rust/issues/5080.

I have **literally** no idea what am I doing, so I would be glad for some feedback, if this is the way to go or if I should go back to the drawing board :) Thanks!

### DIFF
--- a/src/main/kotlin/org/rust/ide/hints/type/RsTypeHintsPresentationFactory.kt
+++ b/src/main/kotlin/org/rust/ide/hints/type/RsTypeHintsPresentationFactory.kt
@@ -31,6 +31,7 @@ class RsTypeHintsPresentationFactory(
     private fun hint(kind: Kind, level: Int): InlayPresentation = when (kind) {
         is TyTuple -> tupleTypeHint(kind, level)
         is TyAdt -> adtTypeHint(kind, level)
+        is TyAlias -> aliasTypeHint(kind, level)
         is TyFunction -> functionTypeHint(kind, level)
         is TyReference -> referenceTypeHint(kind, level)
         is TyPointer -> pointerTypeHint(kind, level)
@@ -87,25 +88,36 @@ class RsTypeHintsPresentationFactory(
         )
 
     private fun adtTypeHint(type: TyAdt, level: Int): InlayPresentation {
-        val aliasedBy = type.aliasedBy
-        val alias = aliasedBy?.element
-
-        val adtName = alias?.name ?: type.item.name
-        val typeDeclaration = alias ?: type.item
+        val adtName = type.item.name
+        val typeDeclaration = type.item
         val typeNamePresentation = factory.psiSingleReference(text(adtName)) { typeDeclaration }
+        val typeArguments = type.typeArguments.zip(type.item.typeParameters)
 
-        val typeArguments = alias?.typeParameters?.map { (aliasedBy.subst[it] ?: TyUnknown) to it }
-            ?: type.typeArguments.zip(type.item.typeParameters)
+        return withGenericsTypeHint(typeNamePresentation, typeArguments, level)
+    }
 
+    private fun aliasTypeHint(type: TyAlias, level: Int): InlayPresentation {
+        val boundElement = type.typeAlias ?: return hint(type.aliases, level)
+        val alias = boundElement.element
+
+        val adtName = alias.name
+        val typeNamePresentation = factory.psiSingleReference(text(adtName)) { alias }
+        val typeArguments = alias.typeParameters.map { (boundElement.subst[it] ?: TyUnknown) to it }
+
+        return withGenericsTypeHint(typeNamePresentation, typeArguments, level)
+    }
+
+    private fun withGenericsTypeHint(
+        typeNamePresentation: InlayPresentation,
+        typeArguments: List<Pair<Ty, RsTypeParameter>>,
+        level: Int
+    ): InlayPresentation {
         val userVisibleKindArguments = mutableListOf<Kind>()
         for ((argument, parameter) in typeArguments) {
             if (!showObviousTypes && isDefaultTypeParameter(argument, parameter)) {
                 // don't show default types
                 continue
             }
-            userVisibleKindArguments.add(argument)
-        }
-        for (argument in type.constArguments) {
             userVisibleKindArguments.add(argument)
         }
 
--- a/src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
+++ b/src/main/kotlin/org/rust/ide/presentation/TypeRendering.kt
@@ -164,11 +164,14 @@ private data class TypeRenderer(
             is TyTraitObject -> ty.traits.joinToString("+", "dyn ") { formatTrait(it, render) }
             is TyAnon -> ty.traits.joinToString("+", "impl ") { formatTrait(it, render) }
             is TyAdt -> buildString {
-                if (useAliasNames && ty.aliasedBy != null) {
-                    append(formatBoundElement(ty.aliasedBy, render))
+                append(getName(ty.item) ?: return anonymous)
+                if (includeTypeArguments) append(formatAdtGenerics(ty, render))
+            }
+            is TyAlias -> buildString {
+                if (useAliasNames && ty.typeAlias != null) {
+                    append(formatBoundElement(ty.typeAlias, render))
                 } else {
-                    append(getName(ty.item) ?: return anonymous)
-                    if (includeTypeArguments) append(formatAdtGenerics(ty, render))
+                    append(render(ty.aliases, level))
                 }
             }
             is TyInfer -> when (ty) {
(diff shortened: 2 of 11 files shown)


vlad20012:
After the years of working with such wrappers (like RsParenType that I mentioned [here](https://github.com/intellij-rust/intellij-rust/issues/6030#issuecomment-685092669), or like **macros** that can be everywhere) I conclude that I don't want to work with these wrappers anymore =D

Regarding PSI, I really want to introduce something like rustc's HIR, i.e. another abstraction layer over PSI, where all macros are expanded, parens are unwrapped, sugar is desugared, etc. Because taking into account such wrappers (and MACROS) everywhere is very tedious and bug-prone.

Regarding `TyAlias`, it mainly affects type unification, type coercion, and all the places where type internals are unwrapped (try to find usages of, for example, `TyAdt.item`). `TyAlias` have to be unwrapped in each place. I guess, this approach is very bug-prone and our tests most likely don't cover these cases since there are not too many tests with aliases. 

I think this is likely a bad design. I'd prefer adding `val aliasedBy` field to each `Ty` class

Kobzol:
I see. My intention was to make TyAlias as transparent as possible. Ideally, it would just behave like the aliased type in most places, automatically (if possible?). And only in a select few cases (like in auto import, type render etc.), you could optionally take a look at the alias and work not with the aliased type, but with the TypeAlias itself. But I'm not sure if it's possible to do this easily.

There are two things that I don't like about `aliasedBy`:
1) It's not recursive. If I have some type `S` referenced by recursive type aliases, I can't go through them, there will always only be the outermost type alias. Something like `type T2 = T1` is unrepresentable in the type system.
    ```
    struct S;
    
    type T1 = S;
    type T2 = T1;
    ```
    `S` will have `aliasedBy = T2`.

    This one is probably less of an issue than I originally thought, since mostly we will care only about the alias that was used to resolve the type.

2) From my (probably quite naive :) ) view, it seems to me that using `aliasedBy` would introduce the same, or possibly larger, amount of boilerplate code, than using something like `TyAlias`. If we look at the places where `aliasedBy` was used for `TyAdt`: [1](https://github.com/intellij-rust/intellij-rust/pull/7368/files#diff-480ac5778f2549cddc688daf253ba235fc347469372d1c02412cdb32626f3aadL31), [2](https://github.com/intellij-rust/intellij-rust/pull/7368/files#diff-7a8aca45b2621a125d734dcbbf5c9594e494643ec6ad83b59ddffc75e49e6168L166), [3](https://github.com/intellij-rust/intellij-rust/pull/7368/files#diff-f90f288e15fa514055574bbec4bae978c9cbfcf80aa0366053250f07a964075aR153) or [4](https://github.com/intellij-rust/intellij-rust/pull/7368/files#diff-c733056d361af5c821c609edc8acc0112b062ae6b1d48bd53a2a276a7dff97e8L64), it seems to me that where before only one type had to work with `aliasedBy`, now this attribute could occur for any type, so either all of the individual methods would have to take it into account (= lot of boilerplate), or before calling these methods you would need to take a look at `aliasedBy` and unwrap it somehow. Which is probably the same amount of boilerplate as with `TyAlias`, but it seems to me that it would be more complicated, because suddenly you would have to deal with `BoundElement<RsTypeAlias>`, while the rest of the code works with `Ty`s.

vlad20012:
I figured out our problem. We think of different abstraction layers. I mostly think about the type inference engine itself, when you're a client of this engine, you think like "this expression has a type Foo, how can I use it?".

So, you can consider my points in the context of type inference engine design. 

I'm absolutely sure that I don't want to see `TyAlias` during type inference. I feel very comfortable with `aliasedBy`.
Maybe we should use different `Ty`s inside and outside of the type inference engine... I'm not sure it worth it.

By the way, `aliasedBy` could be recursive. Is it needed somewhere?

Kobzol:
> By the way, aliasedBy could be recursive. Is it needed somewhere?

Actually, probably not for now, let's ignore that. It seems that substituting type aliases works even without it (https://github.com/intellij-rust/intellij-rust/pull/7345).

As I understand it, supporting `TyAlias` in type inference would be quite annoying. I have looked at the issues that are caused by missing type aliases for non-ADT types (mainly https://github.com/intellij-rust/intellij-rust/issues/4967 and https://github.com/intellij-rust/intellij-rust/issues/5044). In these cases, we don't actually need type inference all that much, because the corresponding items are resolved using name resolution from type references.

If type inference is problematic, would it be possible to transparently ignore/always unwrap `TyAlias` inside of it, and only support it properly in name resolution (with something like `path.resolve(stopAtTypeAlias = true)`)?

I don't particularly like the `TyAlias` approach, but supporting `aliasedBy` in all the places modified in this PR (and probably there will be much more) seems quite error-prone. Although I also believe you that e.g. in type inference the same might be true for `TyAlias` :D

vlad20012:
Why do we ever use types without type inference? If there is a `RsTypeReference` and a substitution, we could use `RsTypeReference.getStubOnlyText(subst)` with actual substitution in order to get its text with substituted generics

Kobzol:
I'm not sure. The problem with wrong type aliases is caused by the import helper (among other things), which uses types for importing. Even if you provide it with a type reference, it first extracts the type from it (AFAIK). I suppose that it does it to unify importing, because sometimes you only get a type from inferring an expression that needs to be imported.

vlad20012:
`RsImportHelper` is just written wrongly, I guess 🤷‍♂️

vlad20012:
Let me try to prototype a typeless solution

Kobzol:
There are situations where you can't get over it though I guess. For example with `CreateFunctionIntention`:

```rust
mod foo {}

pub struct S;
pub type T1 = S;

fn get_t1() -> T1 { ... }

fn bar() {
    foo::function/*caret*/(get_t1());
}
```
needs to create something like this:
```rust
mod foo {
    use crate::T1;

    fn function(p0: T1) {}
}

pub struct S;
pub type T1 = S;

fn get_t1() -> T1 { ... }

fn bar() {
    foo::function/*caret*/(get_t1());
}
```
Here you only have access to a type, not a type reference..

vlad20012:
I tried a typeless PSI generation and now I'm even more sure that this should be done this way.

I have a several additional arguments against `Ty`-based PSi generation:
1. After introducing `TyAlias` you will have to introduce `TyMacro` due to the same reasons
2. After that, you will anyway need PSI-based import&replacement logic because we have type bounds in function signatures
```rust
mod private {
    pub trait Trait;
}
pub mod public {
    pub use super::private::Trait;
}
...
fn foo<T>() where T: private::Trait; // We need to replace `private::Trait` to `Trait` and import it from `public` mod
```
3. The approach "render and then import" is not enough; it doesn't work in the case of name conflict or if we just have a renamed element in the scope (`use foo::A as B;`).

I placed my intermediate results to [genenrate-without-types](https://github.com/intellij-rust/intellij-rust/commits/genenrate-without-types) branch. Generally, it works. There are a bit more changes than I expected, unfortunately(

Kobzol:
Thank you. Now I see that `TyAlias` would be too unwieldy. While it made some things easier, it would also introduce a lot of other problems. I wonder if `aliasedBy` will be even needed on `TyAdt` with your approach?

> The approach "render and then import"

I agree. Actually I think that in general a better way to do auto import (vs the current approach) would be to always generate things using their fully qualified paths and then automatically shorten the paths, instead of generating the short path first and then trying to guess what thing should be imported (https://github.com/intellij-rust/intellij-rust/issues/4979#issuecomment-741626957).

I suppose that we can close this PR, as it's a dead end. But I'm glad that maybe now it's clearer how aliases may be handled in a better way. Will you continue in your branch and create a new PR? Or do you want me to finish it somehow?

vlad20012:
> I wonder if `aliasedBy` will be even needed on `TyAdt` with your approach?
  
Your example with `CreateFunctionIntention` is valid. If we have expression, we can only take its `Ty` (through the type inference machinery, so `TyAlias` is unlikely to be used there). `aliasedBy` is still needed there. Well, yes, it looks like we anyway need both PSI-based and Ty-based rendering:(
Also, don't forget about type hints. We show aliases there instead of expanded types.

> I think that in general a better way to do auto import (vs the current approach) would be to always generate things using their fully qualified paths

Hmm, yes, sounds good. I'd say, I made some groundwork for this in my branch. For example, I covered the case with a renamed item in the local scope. But I didn't think of inserting qualified paths =D

> Will you continue in your branch and create a new PR? Or do you want me to finish it somehow?

I'll finish it. It has almost done. Just some style stuff remains...

> But I'm glad that maybe now it's clearer how aliases may be handled in a better way.

You made me think a lot about this problem =D
Thank you for this!

PR #7377, opened 2021-06-20
CI: create changelog pull request automatically

Undin:
These changes introduce a new `create changelog` workflow that should create a pull request with release changelog template in https://github.com/intellij-rust/intellij-rust.github.io repo and assign it to the corresponding release manager (according to the corresponding milestone). The workflow creates a changelog template by running `changelog.py` script from `intellij-rust.github.io` repo sometime before the release (currently, it's 5 days, i.e. usually on Wednesday morning).
The goals of this change are to remind release managers about changelog creation and automate some of their routine work.

### DIFF
--- a/.github/workflows/create-changelog.yml
+++ b/.github/workflows/create-changelog.yml
@@ -1,0 +1,35 @@
+name: create changelog
+
+on:
+    workflow_dispatch:
+    schedule:
+        - cron: '0 3 * * *' # try everyday
+
+jobs:
+    create-changelog:
+        runs-on: ubuntu-18.04
+        steps:
+            - uses: actions/checkout@v2
+
+            - name: Checkout intellij-rust.github.io repo
+              uses: actions/checkout@v2
+              with:
+                  token: ${{ secrets.WORKFLOW_GITHUB_TOKEN }}
+                  repository: ${{ github.repository_owner }}/intellij-rust.github.io
+                  path: intellij-rust.github.io
+
+            - name: Set up Python
+              uses: actions/setup-python@v1
+              with:
+                  python-version: 3.9
+
+            - name: Install python packages
+              run: pip install -r intellij-rust.github.io/requirements.txt
+
+            - name: Set up git user
+              run: |
+                  (cd intellij-rust.github.io; git config --local user.email "intellij.rust@gmail.com")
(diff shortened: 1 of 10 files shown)


Undin:
@lancelote some additional requirements for testing: fork of intellij-rust.github.io repo and access to the repository for https://github.com/intellij-rust-bot

lancelote:
Thank you 👍 

Some minor notes

- Once Python 3.10 is released we can write `Optional[str]` as `str | None` ([PEP 604][1])
- Since Python 3.9 you are free to write `milestone: Dict` as `milestone: dict` ([PEP 585][3])
  - Also key/value types are welcomed (e.g., `dict[str, str]`)
- I believe `raise Exception` is usually considered bad practice as it is too broad and it is harder to catch this specific exception later in code if needed
- `argparse` is a part of the standard library, hence its import should not be separated by a blank line from other standard library imports
- Type annotation for a return type should not be ommited if the function return nothing, it's `-> None` (or `-> NoReturn` in [some cases][2])

[1]: https://www.python.org/dev/peps/pep-0604/
[2]: https://docs.python.org/3/library/typing.html#typing.NoReturn
[3]: https://www.python.org/dev/peps/pep-0585/

lancelote:
bors merge

Undin:
> Once Python 3.10 is released we can write Optional[str] as str | None

Union types look great!

> Since Python 3.9 you are free to write milestone: Dict as milestone: dict (PEP 585)

👍 
But we need to update all workflows to use python 3.9 for that

> Also key/value types are welcomed (e.g., dict[str, str])

Unfortunately, I usually use `dict` in our python scripts for github API responses where values may be not only strings

> I believe raise Exception is usually considered bad practice as it is too broad and it is harder to catch this specific exception later in code if needed

Is there an analog of Java `IllegalStateException` in python?

> argparse is a part of the standard library, hence its import should not be separated by a blank line from other standard library imports

It's interesting. Python plugin navigated me to some `site-packages` dir on `Go to defition` action for `argparse` module, i.e. it considered the module as external. As a result, the builtin formatter adds a new line for this module. I've dropped `venv` dir and recreated sdk and now it works as expected

> Type annotation for a return type should not be ommited if the function return nothing, it's -> None (or -> NoReturn in some cases)

Quite unexpected after Kotlin where `Unit` return type is usually omitted. I tried to remember this difference. Thanks!

PR #7378, opened 2021-06-20
Specify plugin id for root module tests

avrong:
Based on https://github.com/intellij-rust/intellij-rust/pull/7348#discussion_r652607967 suggestion


avrong:
Tests are failing because of the `PLUGIN_ID` constant in `org/rust/openapiext/utils.kt:256`, which is the same for both test and non-test environments

Undin:
> Tests are failing because of the PLUGIN_ID constant in org/rust/openapiext/utils.kt:256, which is the same for both test and non-test environments

Oh, I see. I totally forgot about it. Sorry
Let's not try to fix it and just close the PR - I don't think that result is worth the effort

PR #7379, opened 2021-06-20
Remove unnecessary platform dependencies

avrong:
PR removes the unnecessary dependencies on `com.intellij.modules.lang` in `plugin.xml`

### DIFF
--- a/plugin/src/main/resources/META-INF/plugin.xml
+++ b/plugin/src/main/resources/META-INF/plugin.xml
@@ -10,19 +10,16 @@
     ]]>
     </change-notes>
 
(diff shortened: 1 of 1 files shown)


Undin (commented):
Let's also remove duplicate `org.toml.lang` dependency.
Currently, we have `<depends>org.toml.lang</depends>` and `<depends optional="true" config-file="toml-only.xml">org.toml.lang</depends>`. Looks like we can replace it with single `<depends config-file="toml-only.xml">org.toml.lang</depends>`
Don't forget to ensure that after these changes, the Rust plugin cannot be installed/enabled without Toml plugin and integration from `toml` module works as expected (I think it's enough to check any one feature from this module)

Undin (approved):
bors r+

Issue #7382, opened 2021-06-21
Convert impl Trait to type parameter works incorrectly with const generics


afetisov:
## Environment

* **IntelliJ Rust plugin version:** 0.4.149.3939-211
* **Rust toolchain version:** 1.53.0 (53cb7b09b 2021-06-17) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2021.1.2 (CL-211.7442.42)
* **Operating system:** Linux 5.4.0-74-generic
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

If a function has const generic parameters and `impl Trait` arguments, then calling the "Convert `impl Trait` to type parameter" refactoring results in incorrect code, since it places the new generic parameter after const generic ones, which is forbidden by the compiler.

## Steps to reproduce

```rust
fn f<const N: usize>(_x: impl /* caret */ Send) {}
```

Call the "Convert `impl Trait` to type parameter" refactoring. Result:

```rust
fn f<const N: usize, T: Send>(_x: T) {}
```

Error: Type parameters must be declared prior to const parameters.

Expected result:

```rust
fn f<T: Send, const N: usize>(_x: T) {}
```

PR #7383, opened 2021-06-21
FIX: support const parameters and trailing commas in ImplTraitToTypeParamIntention

Kobzol:
This PR fixes the behaviour of `ImplTraitToTypeParamIntention` when there are already existing const parameters. It also reworks the insertion logic so that it better supports trailing commas.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7382

changelog: Fix intention to convert an `impl trait` parameter to a generic parameter when the corresponding function already has const parameters.


afetisov:
What's blocking this?

Kobzol:
Probably nothing much except for review bandwidth :) There's a lot of PRs to review, issues to resolve, bugs to fix, features to implement, etc. It might take some time to review a PR, but it will go through eventually.

CC @vlad20012 or @mchernyavsky

mchernyavsky (approved):
Thanks!

bors r+

PR #7384, opened 2021-06-21
Save persistent data of crate local index in workspace.xml

Undin:
Crates local index keeps info related to local user state. It's not supposed to be shared between different machines, so let's use `other.xml` file not to share it

changelog: Don't allow sharing/exporting data of crate local index


Undin:
@avrong I've changed `workspace.xml` to `other.xml` because crate local index is not a project-level service but an application-level one and the default file for non-roamable application data is `other.xml`.
Thank you!

avrong (approved):
It works! The change of state location will force the PersistentHashMap to be fully updated, but as it is an experimental feature, this doesn't seem to be a big problem

Undin:
> The change of state location will force the PersistentHashMap to be fully updated, but as it is an experimental feature, this doesn't seem to be a big problem

Yep. But if we need to change an XML file for a non-experimental feature, there is a way to say IDE to migrate data from an old file to a new one.

Undin:
bors r=avrong

Issue #7376, opened 2021-06-20
"Move Statement Up/Down" doesn't work for statements inside a block expression as parameter


sum-elier:
## Environment

* **IntelliJ Rust plugin version:** 0.4.150.3936-211-nightly
* **Rust toolchain version:** 1.55.0-nightly (ce1d5611a 2021-06-18) x86_64-pc-windows-msvc
* **IDE name and version:** CLion 2021.1.2 (CL-211.7442.42)
* **Operating system:** Windows 10 10.0
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

Trying to move statements inside an expression block being passed as a parameter to a function prevents the command "Move Statement Up/Down" from working in the statements.

## Steps to reproduce

Paste the code snippet below into IDE and try to move any of the `println!` statements up or down using the shortcut mapped to the command "Move Statement Up/Down":

```rust
fn async_receiver<F, T>(future: F)
   where
      F: std::future::Future<Output = T>,
{
}

fn foo() {
   async_receiver(async {
      println!("1");
      println!("2");
   });
}
```
or
```rust
fn foo(a: u32) {}

fn main() {
    foo({
        println!("1");
        println!("2");
        1
    });
}
```

Kobzol:
It's not actually caused by `async` blocks, but by the block being inside an argument list. It also doesn't work here:
```rust
fn foo(a: u32) {}

fn main() {
    foo({
        println!("1");
        println!("2");
        1
    });
}
```

sum-elier:
I'll update the issue title to be accurate and include your example.

PR #7392, opened 2021-06-23
IDE: allow moving up/down in block expressions in comma-separated lists

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7376

changelog: Allow moving up/down in block expressions inside comma-separated lists.

### DIFF
--- a/src/main/kotlin/org/rust/ide/actions/mover/RsLineMover.kt
+++ b/src/main/kotlin/org/rust/ide/actions/mover/RsLineMover.kt
@@ -13,10 +13,8 @@ import com.intellij.openapiext.Testmark
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.psi.PsiWhiteSpace
-import org.rust.lang.core.psi.RsElementTypes
-import org.rust.lang.core.psi.RsFile
-import org.rust.lang.core.psi.RsFunction
-import org.rust.lang.core.psi.RsMatchArm
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.ext.ancestorStrict
 import org.rust.lang.core.psi.ext.elementType
 
 abstract class RsLineMover : LineMover() {
@@ -93,6 +91,12 @@ abstract class RsLineMover : LineMover() {
 
         fun isMovingOutOfMatchArmBlock(sibling: PsiElement, down: Boolean): Boolean =
             isMovingOutOfBraceBlock(sibling, down) && sibling.parent?.parent?.parent is RsMatchArm
+
+        fun isMovingOutOfBlockExprInsideArgList(sibling: PsiElement, down: Boolean): Boolean {
+            if (!isMovingOutOfBraceBlock(sibling, down)) return false
+            val blockExpr = sibling.ancestorStrict<RsBlockExpr>() ?: return false
+            return blockExpr.ancestorStrict<RsValueArgumentList>() != null
+        }
     }
 }
 
--- a/src/test/kotlin/org/rust/ide/actions/mover/RsStatementUpDownMoverTest.kt
+++ b/src/test/kotlin/org/rust/ide/actions/mover/RsStatementUpDownMoverTest.kt
@@ -620,4 +620,79 @@ class RsStatementUpDownMoverTest : RsStatementUpDownMoverTestBase() {
              };
         }
     """, testmark = UpDownMoverTestMarks.moveOutOfBody)
+
+    // https://github.com/intellij-rust/intellij-rust/issues/7376
+    fun `test move down in block inside argument list`() = moveDown("""
+        fn foo(a: u32) {}
+
+        fn main() {
+            foo({
+                println!("1");/*caret*/
+                println!("2");
+                1
+            });
+        }
+    """, """
+        fn foo(a: u32) {}
+
+        fn main() {
+            foo({
+                println!("2");
+                println!("1");/*caret*/
+                1
+            });
+        }
+    """)
+
+    fun `test move up first statement of block expr inside argument list`() = moveUp("""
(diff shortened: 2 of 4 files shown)


src/test/kotlin/org/rust/ide/actions/mover/RsStatementUpDownMoverTest.kt
```diff
@@ -620,4 +620,27 @@ class RsStatementUpDownMoverTest : RsStatementUpDownMoverTestBase() {
...
         }
     """, testmark = UpDownMoverTestMarks.moveOutOfBody)
+
+    // https://github.com/intellij-rust/intellij-rust/issues/7376
+    fun `test move down in block inside argument list`() = moveDown("""
+        fn foo(a: u32) {}
+
+        fn main() {
+            foo({
+                println!("1");/*caret*/
```
ortem, line 630: What about moving up in this case?

Looks like the current implementation does not check boundaries and might produce invalid code, for example
moving up here
```Rust
foo(
    {
        println!("2"); // move up
        println!("1");
        1
    }
);
```
leads to
```Rust
fn main1() {
    foo(
        println!("2");
    {
            println!("1");
            1
        }
    );
}
```

ortem (changes requested):

src/test/kotlin/org/rust/ide/actions/mover/RsStatementUpDownMoverTest.kt
Kobzol: Good catch! The only comma separated lists that accept expressions are argument lists (AFAIK), so I added a check for them. It should be fixed now.

ortem: Great, now the implementation works in that case. But I found another problematic case
```Rust
S {
    a: 42,
    b: foo(|| {
        bar; // move up
    })
};
```
Could you please try to fix it as well and add the corresponding tests?

Kobzol: Thanks, fixed.

ortem: Well, how's about this? 🙂 
```Rust
S {
    a: 42,
    b: foo(&mut || {
        bar; // move up
    })
};
```

You can find such problems too by manually moving some random lines inside block expressions in some real projects (e.g. `testData/rust-analyzer`). BTW probably such fuzz testing could be performed automatically someday

Kobzol: Ok I guess there's no need to be prudent, now any block expr inside argument lists will be considered. Maybe the heuristic could be improved (stop at blocks?), but it seems to me that this is such a rare edge case that it's not worth it.

src/main/kotlin/org/rust/ide/actions/mover/RsLineMover.kt
```diff
@@ -93,6 +91,12 @@ abstract class RsLineMover : LineMover() {
 
         fun isMovingOutOfMatchArmBlock(sibling: PsiElement, down: Boolean): Boolean =
             isMovingOutOfBraceBlock(sibling, down) && sibling.parent?.parent?.parent is RsMatchArm
+
+        fun isMovingOutOfBlockExprInsideArgList(sibling: PsiElement, down: Boolean): Boolean {
+            if (!isMovingOutOfBraceBlock(sibling, down)) return false
+            val blockExpr = sibling.parentOfType<RsBlockExpr>() ?: return false
+            return blockExpr.parentOfType<RsValueArgumentList>() != null
```
ortem, line 98:
```suggestion
            val blockExpr = sibling.ancestorStrict<RsBlockExpr>() ?: return false
            return blockExpr.ancestorStrict<RsValueArgumentList>() != null
```

ortem (approved):
LGTM except for one minor comment. Thank you!

src/test/kotlin/org/rust/ide/actions/mover/RsStatementUpDownMoverTest.kt
ortem: Yes, I think it's mostly fine now 👍

src/main/kotlin/org/rust/ide/actions/mover/RsLineMover.kt
Kobzol: Just out of curiosity, why is this preferred? It seems to call the same function with the same arguments.

Kobzol:
bors r=ortem

Kobzol:
bors retry

src/main/kotlin/org/rust/ide/actions/mover/RsLineMover.kt
ortem: Well, I guess there are several reasons for preferring this method
1. `PsiElement.parentOfType(withSelf: Boolean = true): T?` was introduced only in 201 platform. Before that, there was only `PsiElement.parentOfType(): T?`. So there was no good equivalent for `ancestorOrSelf` for a long time. Now I think it's better to be consistent and use the same methods for the same things
2. `ancestorStrict` and `ancestorOrSelf` have more explicit names and you don't have to remember what that boolean flag means

Kobzol: To be honest, `strict` is also a bit confusing to me, but I agree that `ancestor` is a better name than `parent`. Thanks for the explanation :+1:

Btw CI seems to be stuck at this moment because of download issues.

Undin:
bors r=ortem

Issue #7396, opened 2021-06-23
Incorrect results of "Flatten use statements"


afetisov:
## Environment

* **IntelliJ Rust plugin version:** 0.4.149.3939-211
* **Rust toolchain version:** 1.53.0 (53cb7b09b 2021-06-17) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2021.1.2 (CL-211.7442.42)
* **Operating system:** Linux 5.4.0-74-generic
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

Calling "Flatten use statements" on nested imports results in broken code.

## Steps to reproduce

Consider the following module structure:

```rust
// lib.rs
mod foo {
    pub const FOO: usize = 3;
    pub fn fff() {}
}

mod bar {
    pub fn baz() {}
}
```

Let's assume that this file also has the following import statements:

```rust
// lib.rs
use self::{
    foo::{/* <caret> */FOO},
    bar::baz,
};
```

After calling "Flatten use statements" in the popup menu we get broken imports:

```rust
// lib.rs
use self::{
    foo::foo::FOO,
    foo::bar::baz,
    bar::baz,
};
```

_Expected_:

```rust
// lib.rs
use self::{
    foo::FOO,
    bar::baz,
};
```

Surprisingly, the broken code is produced only when the action is called at the start or end of the use block. I.e. if we consider the following imports:

```rust
use self::{
    foo::{
        FOO,
        /* <caret 1> */ fff /* <caret 2> */,
    },
    bar::baz,
};
```

then calling "Flatten use statements" on caret 1 results in proper code:

```rust
use self::{
    foo::FOO,
    foo::fff,
    bar::baz,
};
```

while calling it on caret 2 results in broken imports:

```rust
use self::{
    foo::foo::FOO,
    foo::foo::fff,
    foo::bar::baz,
    bar::baz,
};
```

PR #7397, opened 2021-06-23
INT: considers siblings near caret in FlattenUseStatementsIntention

Kobzol:
This PR changes `FlattenUseStatementsInspection` so that it works properly on all places inside a use group.

Before, the intention was looking for the use speck to flatten using `ancestorStrict`. That meant that if you have invoked the intention on some element that wasn't a use speck inside a use group, e.g. here:
```rust
use self::{
    foo::{
        a,/*caret*/
        b
    },
    bar::baz,
};
```
the `useSpeckOnCursor` variable was set to `foo::` instead of `a` or `b`. Therefore the intention produced invalid code. Now the intention first looks for siblings located at caret to find the proper use speck to replace.

I suppose that the same thing should be done for `NestUseStatementsIntention`, I can do it in another PR.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7396

changelog: The intention to flatten use imports now works properly on all places inside a use group.


mchernyavsky (approved):
Thanks!

bors r+

PR #6663, opened 2021-01-13
ANN: substitute generic method arguments in FillFunctionArgumentsFix

Kobzol:
Improves https://github.com/intellij-rust/intellij-rust/pull/6613, as discussed [here](https://github.com/intellij-rust/intellij-rust/pull/5644#discussion_r555088815).

Changelog: Correctly resolve generic types when invoking "Fill function arguments fix" on generic methods.

PR #7401, opened 2021-06-24
Add API to get inferred method type and substitution

vlad20012:
Internal. It's now possible to get `Substotution` of `RsMethodCall` or `RsPathExpr`.

I promised that [here](https://github.com/intellij-rust/intellij-rust/pull/5644#issuecomment-683338917) :)

Note that this doesn't work correctly for `refined` paths (see `test infer method arg with multiple impls of the same trait` or `test resolve UFCS method call with multiple impls of the same trait`). I'll fix it later

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
+++ b/src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
@@ -70,7 +70,7 @@ class RsInferenceResult(
     val patFieldTypes: Map<RsPatField, Ty>,
     private val expectedExprTypes: Map<RsExpr, Ty>,
     private val resolvedPaths: Map<RsPathExpr, List<ResolvedPath>>,
-    private val resolvedMethods: Map<RsMethodCall, List<MethodResolveVariant>>,
+    private val resolvedMethods: Map<RsMethodCall, InferredMethodCallInfo>,
     private val resolvedFields: Map<RsFieldLookup, List<RsElement>>,
     private val adjustments: Map<RsExpr, List<Adjustment>>,
     val diagnostics: List<RsDiagnostic>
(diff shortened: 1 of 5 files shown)


Kobzol:
Nice! I was trying to do this with some semi-internal API (https://github.com/intellij-rust/intellij-rust/pull/6663). I suppose that this approach will be much better.

Kobzol (approved):
I tried https://github.com/intellij-rust/intellij-rust/pull/6663 with this PR and it works fine :+1:

vlad20012:
I'm wondering if such an API is also needed for paths (for #7345) and it looks like yes. I'll add it too

Kobzol:
Yeah, that would be great for the substitute type alias intention :+1:

src/main/kotlin/org/rust/lang/core/types/infer/TypeInference.kt
```diff
@@ -270,14 +280,18 @@ class RsInferenceContext(
...
         }
         for ((call, traitRef) in methodRefinements) {
-            val variant = resolvedMethods[call]?.firstOrNull() ?: continue
+            val info = resolvedMethods[call] ?: continue
+            val variant = info.resolveVariants.firstOrNull() ?: continue
             val impl = lookup.select(resolveTypeVarsIfPossible(traitRef)).ok()?.impl as? RsImplItem ?: continue
             val fn = impl.expandedMembers.functions.find { it.name == variant.name } ?: continue
             val source = TraitImplSource.ExplicitImpl(RsCachedImplItem.forImpl(impl))
-            resolvedMethods[call] = listOf(variant.copy(element = fn, source = source))
+            // TODO remap subst
```
vlad20012, line 293: Note that this doesn't work correctly for refined paths (see test infer method arg with multiple impls of the same trait or test resolve UFCS method call with multiple impls of the same trait). I'll fix it later

Issue #7402, opened 2021-06-24
False-positive `Can't resolve` warnings in Cargo.toml


Undin:
## Environment

* **IntelliJ Rust plugin version:** 0.4.149.3939-211
* **Rust toolchain version:** 1.53.0 (53cb7b09b 2021-06-17) x86_64-apple-darwin
* **IDE name and version:** IntelliJ IDEA 2021.1.1 Ultimate Edition (IU-211.7142.45)
* **Operating system:** macOS 10.15.6
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description
If you try to use patterns in `[workspace]` table (i.e. `*` and `?` symbols), you will get false-positive `Can't resolve directory` and `Can't resolve file` warnings

<img width="445" alt="Screen Shot 2021-06-24 at 17 54 57" src="https://user-images.githubusercontent.com/2539310/123299760-87ddda80-d522-11eb-9805-e5a4e6b3bfc4.png">

## Steps to reproduce
```toml
[workspace]
members = [
    "lesson1/*/",
]
exclude = [
    "**/*.yaml"
]
```

Kobzol:
@Undin In theory, we could expand the glob in the inspection, but I worry about possible edge cases and DDOS attacks :D If the glob was nasty. So I propose to just disable the inspection if the path contains a wildcard `*`?

PR #7403, opened 2021-06-24
TOML: ignore path references with wildcards to avoid false positives

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7402

changelog: Ignore paths with wildcards in TOML to avoid false positives.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
+++ b/toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
@@ -6,6 +6,7 @@
 package org.rust.toml.resolve
 
 import com.intellij.openapi.util.Condition
+import com.intellij.openapi.util.TextRange
 import com.intellij.patterns.PsiElementPattern
 import com.intellij.psi.PsiDirectory
 import com.intellij.psi.PsiElement
@@ -13,6 +14,7 @@ import com.intellij.psi.PsiFileSystemItem
 import com.intellij.psi.PsiReferenceProvider
 import com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReference
 import com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReferenceSet
+import com.intellij.psi.util.parentOfType
 import com.intellij.util.ProcessingContext
 import org.rust.lang.core.or
 import org.rust.lang.core.psi.RsFile
@@ -20,6 +22,7 @@ import org.rust.lang.core.psi.ext.ancestorStrict
 import org.rust.toml.CargoTomlPsiPattern
 import org.rust.toml.resolve.PathPatternType.*
 import org.toml.lang.psi.TomlHeaderOwner
+import org.toml.lang.psi.TomlKeyValue
 import org.toml.lang.psi.TomlLiteral
 import org.toml.lang.psi.ext.TomlLiteralKind
 import org.toml.lang.psi.ext.kind
@@ -28,6 +31,8 @@ import org.toml.lang.psi.ext.name
 class CargoTomlFileReferenceProvider(private val patternType: PathPatternType) : PsiReferenceProvider() {
     override fun getReferencesByElement(element: PsiElement, context: ProcessingContext): Array<FileReference> {
         val kind = (element as? TomlLiteral)?.kind ?: return emptyArray()
+        if (kind !is TomlLiteralKind.String) return emptyArray()
+
         val (completeDirs, completeRustFiles) = when (patternType) {
             WORKSPACE -> true to false
             GENERAL -> {
@@ -38,12 +43,21 @@ class CargoTomlFileReferenceProvider(private val patternType: PathPatternType) :
             BUILD -> false to true
         }
 
-        if (kind !is TomlLiteralKind.String) return emptyArray()
-        return CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles).allReferences
+        val ignoreGlobs = patternType == WORKSPACE &&
+            element.parentOfType<TomlKeyValue>()?.key?.segments?.firstOrNull()?.name in KEYS_SUPPORTING_GLOB
+
+        val referenceSet: FileReferenceSet = if (ignoreGlobs) {
+            GlobIgnoringFileReferenceSet(element, completeDirs, completeRustFiles)
+        } else {
+            CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles)
+        }
+
+        return referenceSet.allReferences
     }
 
     companion object {
         private val TARGET_TABLE_NAMES = listOf("lib", "bin", "test", "bench", "example")
+        private val KEYS_SUPPORTING_GLOB = listOf("members", "default-members")
     }
 }
 
@@ -53,7 +67,7 @@ enum class PathPatternType(val pattern: PsiElementPattern.Capture<out PsiElement
     BUILD(CargoTomlPsiPattern.buildPath)
 }
 
-private class CargoTomlFileReferenceSet(
+private open class CargoTomlFileReferenceSet(
     element: TomlLiteral,
     private val completeDirs: Boolean,
     private val completeRustFiles: Boolean
@@ -67,4 +81,30 @@ private class CargoTomlFileReferenceSet(
             }
         }
     }
+}
+
+private class GlobIgnoringFileReferenceSet(
+    element: TomlLiteral,
+    completeDirs: Boolean,
+    completeRustFiles: Boolean
+) : CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles) {
+    private var globPatternFound: Boolean = false
+
+    override fun reparse() {
+        globPatternFound = false
+        super.reparse()
+    }
+
+    override fun createFileReference(range: TextRange, index: Int, text: String): FileReference? {
+        if (!globPatternFound && isGlobPathFragment(text)) {
+            globPatternFound = true
+        }
+        if (globPatternFound) return null
+        return super.createFileReference(range, index, text)
+    }
+}
+
+private fun isGlobPathFragment(text: String?): Boolean {
+    if (text == null) return false
+    return text.contains("?") || text.contains("*") || text.contains("[") || text.contains("]")
 }
--- a/toml/src/test/kotlin/org/rust/toml/inspections/CargoTomlUnresolvedPathReferenceInspectionTest.kt
+++ b/toml/src/test/kotlin/org/rust/toml/inspections/CargoTomlUnresolvedPathReferenceInspectionTest.kt
@@ -100,4 +100,80 @@ class CargoTomlUnresolvedPathReferenceInspectionTest : RsInspectionsTestBase(Tom
         [dependencies]
         foo = { path = "foo.rs" }/*caret*/
     """)
+
+    fun `test wildcard inside path`() = checkByFileTree("""
+        //- foo/Cargo.toml
+        [package]
+        name = "foo"
+
+        //- Cargo.toml
+        [dependencies]
+        foo = { path = "foo/<warning descr="Cannot resolve directory '*'">*</warning>/<warning descr="Cannot resolve file 'main.rs'">main.rs</warning>" }/*caret*/
+    """)
+
+    fun `test wildcard inside workspace members path`() = checkByFileTree("""
+        //- foo/bar.rs
+        fn foo() {}
+
+        //- Cargo.toml
+        [workspace]
+        members = [
+            "foo/*/bar.rs"/*caret*/,
+        ]
+    """)
+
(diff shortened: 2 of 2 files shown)


toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -39,6 +40,10 @@ class CargoTomlFileReferenceProvider(private val patternType: PathPatternType) :
         }
 
         if (kind !is TomlLiteralKind.String) return emptyArray()
+
+        // Ignore paths with wildcards to avoid false positives, since they cannot be easily resolved
+        if (element.stringValue?.contains("*") == true) return emptyArray()
```
Undin, line 45: Unfortunately, it doesn't work as expected because:
- in `"foo/*/"` path I expect that `foo` is resolved to the corresponding directory. Otherwise, it not only breaks "go to definition" action but also breaks refactoring like rename
- it allows `*` in any reference provided by `CargoTomlFileReferenceProvider` although they are allowed only in `workspace` table
- it doesn't cover other special symbols described in https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspace-section

Since the corresponding issue is not critical at all, we can provide a proper solution instead of quick hack to workaround the issue somehow. I suppose we can try to partially parse glob patterns and if the current segment or some previous one contains special symbols just return `null` in `com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReferenceSet#createFileReference`.
I suppose `com.intellij.openapi.vcs.changes.ignore.cache.PatternCache#createRegex` may help here

Undin (changes requested):

Kobzol:
I see. I changed it so that any references after and including the first glob reference are ignored inside workspace `members`.
I'm not sure how to use the `createRegex` function though, it didn't seem to create a regex that would recognize a glob pattern.

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -56,7 +67,7 @@ enum class PathPatternType(val pattern: PsiElementPattern.Capture<out PsiElement
 private class CargoTomlFileReferenceSet(
     element: TomlLiteral,
     private val completeDirs: Boolean,
-    private val completeRustFiles: Boolean
+    private val completeRustFiles: Boolean,
```
Undin, line 70: nit: looks redundant

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -67,4 +78,21 @@ private class CargoTomlFileReferenceSet(
             }
         }
     }
+
+    override fun createFileReference(range: TextRange?, index: Int, text: String?): FileReference? {
+        val reference = super.createFileReference(range, index, text) ?: return null
+
+        return if (isGlobPathFragment(text)) {
+            GlobReference(reference)
```
Undin, line 86: Why do we need `GlobReference` at all? I think we can return `null` if `isGlobPathFragment(text)`

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -67,4 +78,21 @@ private class CargoTomlFileReferenceSet(
             }
         }
     }
+
+    override fun createFileReference(range: TextRange?, index: Int, text: String?): FileReference? {
```
Undin, line 82:
```suggestion
    override fun createFileReference(range: TextRange, index: Int, text: String): FileReference? {
```

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -39,7 +42,15 @@ class CargoTomlFileReferenceProvider(private val patternType: PathPatternType) :
         }
 
         if (kind !is TomlLiteralKind.String) return emptyArray()
-        return CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles).allReferences
+        val references = CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles).allReferences
+
+        val ignoreGlobs = patternType == WORKSPACE &&
+            element.parentOfType<TomlKeyValue>()?.key?.segments?.firstOrNull()?.name == "members"
```
Undin, line 48: Why do we check only `members` ? As I see, glob patterns can be used in all arrays in `workspace` table

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -67,4 +78,21 @@ private class CargoTomlFileReferenceSet(
...
+            GlobReference(reference)
+        } else {
+            reference
+        }
+    }
+}
+
+private fun isGlobPathFragment(text: String?): Boolean {
+    if (text == null) return false
+    return text.contains("?") || text.contains("*") || text.contains("[") || text.contains("]")
```
Undin, line 95: Let's add some tests to check that we actually support not only `*` symbols

Undin (changes requested):
> I see. I changed it so that any references after and including the first glob reference are ignored inside workspace members.
I'm not sure how to use the `createRegex` function though, it didn't seem to create a regex that would recognize a glob pattern.

Sorry for the confusion. I meant that `createRegex` can be used as an example of parsing implementation if we need it. But parsing reduced to several `contains`

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
Kobzol: I did that originally, but if you return null, there will be no way to find out that a glob has occurred "in the middle". Therefore we could not skip path fragments after a glob. Maybe it could be done with some inner state inside the reference set, but this seemed cleaner.

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
Kobzol: The documentation states that it works only in members I think. I tried globs in exclude and they do not seem to work (i.e. exclude seems to consider * to be a normal path character).

Undin: Hm, interesting
Looks like it doesn't work in `exclude` (but I actually was sure that it should) but it works in `default-members` section

For example, for the following project

Screen Shot 2021-06-28 at 19 18 19

with root Cargo.toml
```toml
[workspace]
members = [
    "lesson1/*/",
    "lesson2/*/",
]
default-members = [ "lesson1/*/" ]
```

`cargo build` compiles only `lesson1/task1` and `lesson1/task3`

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
Undin: Oh, I see
But I'm still a bit worried about the current implementation because if `createFileReference` doesn't return null, `FileReferenceSet` saves the reference into inner array and it can be used in other parts.
So I thought about saving index of the first glob pattern into local field and checking it in `createFileReference`.
To restore the default state, we can override `reparse`
Something like
```kotlin
private var firstGlobPatternIndex: Int? = null

override fun reparse() {
    try {
        super.reparse()
    } finally {
        firstGlobPatternIndex = null
    }
}
```

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
Kobzol: Added `default-members` to the supported key list.

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
Kobzol:
```kotlin
override fun reparse() {
    firstGlobPatternIndex = null
    super.reparse()
}
```
This makes more sense to me - the state should be reset before every parse, right? Or maybe I misunderstood.

Undin: Oh, it doesn't matter, they do the same things

toml/src/main/kotlin/org/rust/toml/resolve/CargoTomlFileReferenceProvider.kt
```diff
@@ -39,11 +42,20 @@ class CargoTomlFileReferenceProvider(private val patternType: PathPatternType) :
...
-        return CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles).allReferences
+        val referenceSet = CargoTomlFileReferenceSet(element, completeDirs, completeRustFiles)
+
+        val ignoreGlobs = patternType == WORKSPACE &&
+            element.parentOfType<TomlKeyValue>()?.key?.segments?.firstOrNull()?.name in KEYS_SUPPORTING_GLOB
+        return if (ignoreGlobs) {
+            referenceSet.referencesUntilFirstGlob.toList().toTypedArray()
+        } else {
+            referenceSet.allReferences
+        }
```
Undin, line 53: Maybe it's better just to pass a flag to `CargoTomlFileReferenceSet` not to produce references for glob patterns?
I suppose it should simplify implementation

Kobzol: Ok, although then we need to introduce a new getter for references anyway, since allReferences cannot be used directly (because of the problem with nulls returned from createFileReference).

Undin: > because of the problem with nulls returned from createFileReference

What problem are you talking about?

Kobzol: Sorry for the confusion, I meant this: https://github.com/intellij-rust/intellij-rust/pull/7403#discussion_r659176699.
It seemed to me that it would be fragile to filter the references directly in `createFileReference`, because it would need to become stateful (it would need to return `null` after the first glob reference).

But since now `CargoTomlFileReferenceSet` is stateful anyway, it probably doesn't matter. Would this be ok?

```kotlin
// ignoreGlobReferences is a flag passed to CargoTomlFileReferenceSet

override fun createFileReference(range: TextRange, index: Int, text: String): FileReference? {
        val reference = super.createFileReference(range, index, text) ?: return null
        if (ignoreGlobReferences) {
            if (!globFound && isGlobPathFragment(text)) {
                globFound = true
            }
            if (globFound) return null
        }
        return reference
    }
```

Undin: Yep, looks ok for me and it should work as expected

Kobzol: Ah, dammit. Virtual calls in constructors strike back again :(
`FileReferenceSet` calls reparse in its constructor, therefore the `ignoreGlobReferences` is not set properly when `reparse` is executed :(

We could call reparse manually after the object is fully constructed, but that is a nasty hack.
So I guess let's introduce additional property that will filter the references if the flag is set?

Undin: You can create a child class of `CargoTomlFileReferenceSet` where you can filter glob references unconditionally.
And choose which class you should use in `getReferencesByElement` to avoid this problem

Undin: Interesting fact: there is another constructor `com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReferenceSet#FileReferenceSet(java.lang.String, com.intellij.psi.PsiElement, int, com.intellij.psi.PsiReferenceProvider, boolean, boolean, com.intellij.openapi.fileTypes.FileType[], boolean)` where you can pass `init` flag to control if `reparse` call is needed in  super constructor

Kobzol: Hmm, using a child class is probably the cleanest solution. I changed it :)

Undin (approved):
Thanks!
bors r+

Undin:
bors retry

Issue #7400, opened 2021-06-24
Inline variable fails to take operator precedence into account


afetisov:
## Environment

* **IntelliJ Rust plugin version:** 0.4.149.3939-211
* **Rust toolchain version:** 1.53.0 (53cb7b09b 2021-06-17) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2021.1.2 (CL-211.7442.42)
* **Operating system:** Linux 5.4.0-74-generic
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

Code:

```rust
fn f() -> isize {
    let a = 2 - 3;
    5 - a
}
```

Inline variable `a`:

```rust
fn f() -> isize {
    5 - 2 - 3
}
```

Expected:

```rust
fn f() -> isize {
    5 - (2 - 3)
}
```

PR #7407, opened 2021-06-25
FIX: parenthesize expressions where needed in Inline Value refactoring

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7400

changelog: Parenthesize inlined expressions in Inline Value refactoring if they could produce invalid or semantically different code after performing the refactoring.


vlad20012:
Please also take a look at `UnaryExpr`, `CastExpr`, `TryExpr` and `RangeExpr`, and to struct literals in `ForExpr`

Kobzol:
Added more cases.

vlad20012:
Oh, and also `while let` with a struct literal

Kobzol:
And also `if let`. I wonder what have I missed this time :D

vlad20012 (approved):
bors r+

Issue #5608, opened 2020-06-19
Annotate unnecessarily qualified paths


Kobzol:
## Environment
* **IntelliJ Rust plugin version:** 0.2.124.SNAPSHOT-201-dev
* **Rust toolchain version:** 1.44.0 (49cae5576 2020-06-01) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.1.1 (CL-201.7223.86)
* **Operating system:** Linux 5.0.0-37-generic

The Kotlin plugin marks paths that are unnecessarily qualified, for example:
```kotlin
import ...A.B

object A {
    object B {
        fun foo() {}
    }
}

fun test() {
    <warning>A.</warning>B.foo()
    B.foo()
}
```
We could do the same for Rust paths (https://github.com/intellij-rust/intellij-rust/pull/5451 could be used to remove the nested path).

Kobzol:
Now that https://github.com/intellij-rust/intellij-rust/pull/5451 is merged, it could be used as a quick fix for this annotation.

However, I realized that there is no existing lint for this, if I'm not mistaken (in `check`/`clippy`). So maybe it's not a good idea to display this annotation. @Undin what do you think? :)

Undin:
@Kobzol actually, compiler provides this lint but it has `allow` level by default

For example, for the following code
```rust
#![warn(unused_qualifications)]

mod foo {
    pub fn bar() {}
}

use foo::bar;

fn main() {
    foo::bar();
}


```

rustc produces
```
warning: unnecessary qualification
  --> src/main.rs:10:5
   |
10 |     foo::bar();
   |     ^^^^^^^^
   |
note: the lint level is defined here
  --> src/main.rs:1:9
   |
1  | #![warn(unused_qualifications)]
   |         ^^^^^^^^^^^^^^^^^^^^^
```

It can be useful in some cases but not sure how to provide the ability to disable it because it's already "disabled" by the compiler itself

Kobzol:
Ooh, I didn't know about that! Hmm, I have never seen `#![warn(unused_qualifications)]` in the wild. It seems that some projects use it (https://github.com/search?q=%22%23%21%5Bwarn%28unused_qualifications%29%5D%22&type=Code), but it's probably not very common.

Undin:
> it's probably not very common.

Yeah.
Probably, we can provide weak warning for `allow` lint level  (instead of default "info") but we should be sure that we don't irritate users

[...]

PR #7393, opened 2021-06-23
Fix "Implement Members" action: prefer PSI to `Ty`s when generating PSI

vlad20012:
Fixes #4967, fixes #5044, fixes #4979, fixes #6340, fixes #2982, fixes #5592, fixes #6085

1. Never expand type aliases (#4967, #5044)
2. Don't expand macro types
3. Auto-import trait names from type bounds (#2982)
4. Resolve name conflicts during auto-import (#4979, #6340):
   1. If the target item is in the scope with a different name (`use A as B`), use that name (#5592, #6085)
   2. If the name exists in the scope, but refers to another item, use a qualified path to the target item
   3. If the target can't be imported, use crate-qualified path to the item (produces incorrect code, but it should be simpler to fix)

changelog:

PR #7410, opened 2021-06-25
INSP: add lint inspection that annotates unnecessarily qualified paths

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/5608

Related issue: https://github.com/intellij-rust/intellij-rust/issues/5888

changelog: Add inspection that finds unnecessarily qualified paths. It corresponds to the `unused_qualifications` lint.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/RemovePathPrefixFix.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/RemovePathPrefixFix.kt
@@ -1,0 +1,37 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.fixes
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiFile
+import com.intellij.psi.SmartPointerManager
+import org.rust.openapiext.document
+
+/**
+ * Fix that removes a prefix of a certain path.
+ *
+ * E.g. `foo::bar::baz` -> `baz`.
+ */
+class RemovePathPrefixFix(
+    file: PsiFile,
+    range: TextRange,
+) : LocalQuickFix {
+    private val fileWithRange = SmartPointerManager.getInstance(file.project)
+        .createSmartPsiFileRangePointer(file, range)
+
+    override fun getName(): String = "Remove unnecessary path prefix"
+    override fun getFamilyName(): String = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val file = fileWithRange.containingFile ?: return
+        val range = fileWithRange.range ?: return
+        val document = file.document
+        document?.deleteString(range.startOffset, range.endOffset)
+    }
+}
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
@@ -1,0 +1,85 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.inspections.lints
+
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.util.parentOfType
+import org.rust.ide.inspections.RsProblemsHolder
+import org.rust.ide.inspections.fixes.RemovePathPrefixFix
+import org.rust.lang.core.parser.RustParserUtil
+import org.rust.lang.core.psi.RsPath
+import org.rust.lang.core.psi.RsPathCodeFragment
+import org.rust.lang.core.psi.RsUseItem
+import org.rust.lang.core.psi.RsVisitor
+import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.resolve.TYPES_N_VALUES_N_MACROS
+
+class RsUnnecessaryQualificationsInspection : RsLintInspection() {
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnusedQualifications
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): RsVisitor = object : RsVisitor() {
+        override fun visitPath(path: RsPath) {
+            val root = path.rootPath()
+            val useItem = path.parentOfType<RsUseItem>()
+
+            if (useItem == null && root == path && root.canBeShortened()) {
+                val target = getUnnecessarilyQualifiedPath(path)
+                if (target != null) {
+                    val unnecessaryPart = target.fullQualifier()
+                    val range = TextRange(0, unnecessaryPart.length)
+
+                    val fix = RemovePathPrefixFix(path.containingFile,
+                        TextRange(root.startOffset, root.startOffset + unnecessaryPart.length)
+                    )
+                    holder.registerLintProblem(root, "Unnecessary qualification", range, fix)
+                }
+            }
+            super.visitPath(path)
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+                        TextRange(root.startOffset, root.startOffset + unnecessaryPart.length)
+                    )
+                    holder.registerLintProblem(root, "Unnecessary qualification", range, fix)
+                }
+            }
+            super.visitPath(path)
+        }
+    }
+
+    private fun getUnnecessarilyQualifiedPath(path: RsPath): RsPath? {
```
vlad20012, line 45: Btw this looks a bit similar to my [trySimplifyPath](https://github.com/intellij-rust/intellij-rust/pull/7393/files#diff-6e12710c6434ae7f5debe4dd6682cf237be74e09d0680d12e891263e89bd9cfdR686) from #7393

Kobzol: Indeed it does! It seems that your code also does some renaming? Which is probably not applicable here. I'm not sure if it's worth it to deduplicate these methods, as the intent is a bit different I guess.

src/main/kotlin/org/rust/ide/inspections/fixes/RemovePathPrefixFix.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.intellij.psi.PsiFile
+import com.intellij.psi.SmartPointerManager
+import org.rust.openapiext.document
+
+/**
+ * Fix that removes a prefix of a certain path.
+ *
+ * E.g. `foo::bar::baz` -> `baz`.
+ */
+class RemovePathPrefixFix(
```
dima74, line 21: Why not use `SubstituteTextFix.Companion#delete`?

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
```diff
@@ -0,0 +1,87 @@
...
+            val currentName = currentPath.referenceName ?: break
+            pathText = "$currentName::$pathText"
+        }
+
+        return null
+    }
+}
+
+private fun RsPath.canBeShortened(): Boolean {
+    return path != null || coloncolon != null
```
dima74, line 77: Why not `&&`?

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
```diff
@@ -0,0 +1,87 @@
...
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): RsVisitor = object : RsVisitor() {
+        override fun visitPath(path: RsPath) {
+            val root = path.rootPath()
+            val useItem = path.parentOfType<RsUseItem>()
+
+            if (useItem == null && root == path && root.canBeShortened()) {
+                val target = getUnnecessarilyQualifiedPath(path)
+                if (target != null) {
+                    val unnecessaryPart = target.fullQualifier()
+                    val range = TextRange(0, unnecessaryPart.length)
```
dima74, line 33: This will not work when there are spaces or comments in the middle of the path (please add such test):
```rust
pub mod a {
    pub mod b {
        pub fn func() {}
    }
}

fn main() {
    use a::b::func;
    a :: b::func();
}
```

Consider using `target.referenceNameElement!!.startOffset` here instead

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
```diff
@@ -0,0 +1,87 @@
...
+    private fun getUnnecessarilyQualifiedPath(path: RsPath): RsPath? {
+        if (path.resolveStatus == PathResolveStatus.UNRESOLVED) return null
+
+        val target = path.reference?.resolve() ?: return null
+
+        var pathText = path.referenceName ?: return null
+        var currentPath = path
+        val basePath = path.basePath()
+
+        while (true) {
```
dima74, line 56: I don't like a bit that we use cycle with mutable variables here. Maybe we can analyze not only root paths in visitor, so we will not need to do manual recursion here?

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
Kobzol: We also want to shorten paths like `::serde::Serialize`, so there can only be the double colon and no parent path.

src/main/kotlin/org/rust/ide/inspections/fixes/RemovePathPrefixFix.kt
Kobzol: I wasn't aware that I can modify the fix name in such case. Thanks!

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
Kobzol: Thanks, that's much easier indeed.

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
Kobzol: But then the inspection could be triggered multiple times? For
`a::b::c` I'm walking the path from the right, and once I find some suffix (`c`, `b::c`, `a::b::c`) that can be resolved, I mark everything to the left as unnecessary. If I analyzed `a::b` and `a::b::c` separately, I could mark the `a::` path as unnecessary twice if both `b::c` and `c` were resolvable.

I'm also not sure if we could depend on the visitor on visiting the path in the right order. I can try to refactor the while loop to a more functional implementation though.

dima74: > But then the inspection could be triggered multiple times?

Yes, you are right

> I can try to refactor the while loop to a more functional implementation though.

Let's try. I would use something like this:
```kotlin
val qualifiers = generateSequence(path) { path.qualifier }
for (currentPath in qualifiers) {
    val currentName = currentPath.referenceName ?: break
    val pathText = "$currentName::${path.text.removePrefix(currentPath.text)}"
    ...
}
```

Kobzol:
I changed the implementation to be hopefully a bit readable (first time I have used `runningFold`! :D ). Although there are still some rough mainly around leading `::`. I also added a few more tests and rebased, since I needed changes from `master` to test the changes locally in `CLion`.

dima74:
I tested the inspection on `cargo` and `rust-analyzer` projects, and found only one false-positive, good work 👍

Here is the false-positive:
```rust
fn main() {
    let _ = Option::<()>::None;
}
```

src/main/kotlin/org/rust/ide/inspections/lints/RsUnnecessaryQualificationsInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+
+class RsUnnecessaryQualificationsInspection : RsLintInspection() {
+    override fun getLint(element: PsiElement): RsLint = RsLint.UnusedQualifications
+
+    override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean): RsVisitor = object : RsVisitor() {
+        override fun visitPath(path: RsPath) {
+            val root = path.rootPath()
+            val useItem = path.parentOfType<RsUseItem>()
+
+            if (useItem == null && root == path && root.canBeShortened()) {
```
dima74, line 29: Let's inline root variable here, since it's a bit confusing to have two equal variables `root` and `path` inside `if`-block

dima74 (approved):

Kobzol:
bors r=dima74

Undin:
bors retry

PR #7411, opened 2021-06-25
Take into account CARGO_HOME environment variable in Crates Local Index

avrong:
changelog: Take into account the CARGO_HOME environment variable when using crates local index

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
@@ -109,13 +109,17 @@ class CratesLocalIndexServiceImpl
     private val isUpdating: AtomicBoolean = AtomicBoolean(false)
 
     init {
-        val cargoRegistryIndexRefsLocation = Paths.get(System.getProperty("user.home"), CARGO_REGISTRY_INDEX_LOCATION, "refs/").toString()
+        val cargoRegistryIndexRefsLocation = Paths.get(cargoRegistryIndexPath.toString(), "refs/").toString()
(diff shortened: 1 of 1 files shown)


toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -285,10 +284,20 @@ class CratesLocalIndexServiceImpl
...
-        // TODO: Determine how path to index is created
-        private const val CARGO_REGISTRY_INDEX_LOCATION: String = ".cargo/registry/index/github.com-1ecc6299db9ec823/.git/"
-        private const val CARGO_REGISTRY_INDEX_TAG_PRE_1_50 = "origin/master"
-        private const val CARGO_REGISTRY_INDEX_TAG = "origin/HEAD"
+        private val cargoHome: String
+            get() = System.getenv("CARGO_HOME")
+                ?: Paths.get(System.getProperty("user.home"), ".cargo/").toString()
+
+        // Currently for crates.io only
+        private val cargoRegistryIndexPath
```
ortem, line 292: Please specify the type explicitly

ortem (changes requested):
Please handle the situation when `CARGO_HOME` is invalid (e.g. contains a non-existent path). Currently, `LocalFileSystem.getInstance().refreshAndFindFileByPath(cargoRegistryIndexRefsLocation)!!` call (after rebase onto master) leads to `java.lang.NullPointerException` without any user-friendly description

avrong:
I changed how the `CARGO_HOME` environment variable is being retrieved as macOS doesn't bypass them from the shell, so `System.getEnv()` doesn't return any of those. To handle this, IDE calls shell environment on start to get envars using `printenv` and can return them using `EnviromentUtil`

ortem (approved):
LGTM

PR #7412, opened 2021-06-25
REF: add support for type aliases in getter/setter generation

Kobzol:
changelog: Do not expand type aliases when generating getters and setters.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/generate/getter/GenerateGetterHandler.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/generate/getter/GenerateGetterHandler.kt
@@ -7,15 +7,11 @@ package org.rust.ide.refactoring.generate.getter
 
 
 import com.intellij.openapi.editor.Editor
-import org.rust.ide.presentation.renderInsertionSafe
 import org.rust.ide.refactoring.generate.BaseGenerateAction
 import org.rust.ide.refactoring.generate.BaseGenerateHandler
 import org.rust.ide.refactoring.generate.GenerateAccessorHandler
 import org.rust.ide.refactoring.generate.StructMember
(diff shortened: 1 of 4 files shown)


vlad20012 (approved):
:+1: 
bors r+

lancelote:
Hi @Kobzol! While looking into tests I noticed [you expect](https://github.com/intellij-rust/intellij-rust/blob/0e7668fe21d0b2dee5994f81fd0f7f954bf4b679/src/test/kotlin/org/rust/ide/refactoring/generate/GenerateGetterActionTest.kt#L427) the following getter to be generated

```rust
type Alias = (u32, u32);

struct S {
    a: Alias

impl S {
    pub fn a(&self) -> &Alias {
        &self.a
    }
}
```

... while I got

```rust
type Alias = (u32, u32);

struct S {
    a: Alias
}

impl S {
    pub fn a(&self) -> Alias {
        self.a
    }
}
```

I wonder why there is such a difference and why the tests don't fail? 🤔 

My environment:


* **IntelliJ Rust plugin version:** 0.4.155.4075-212-nightly
* **Rust toolchain version:** 1.56.0-nightly (492723897 2021-07-29) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.3 EAP (CL-213.3008)
* **Operating system:** macOS 11.4

Kobzol:
`Alias` implements `Copy`, so the IDE version is correct. I can't check it right now, but I suppose that in tests `Copy` is not inferred correctly (probably because of missing stdlib descriptor or `@ExpandMacros` annotation) and therefore the test is wrong.

I'll fix it once I return from vacation, in September.

lancelote:
@Kobzol Thank you for the clarification 🙏 I am sorry for pinging you on the vacation 🙇

PR #7418, opened 2021-06-27
CI: don't create changelog PR for closed milestone

Undin:
bors r+


scripts/make_changelog_pull_request.py
```diff
@@ -59,6 +59,7 @@ def main() -> None:
     today = date.today()
     if today >= release_date or milestone["state"] == "closed":
         print(f"Milestone v{release_patch_version} is over")
+        return
```
Undin, line 62: CC @lancelote

PR #7425, opened 2021-06-28
Enable by default crates local index for nightly users

avrong:
changelog: Crates local index feature is now enabled on nightly by default

### DIFF
--- a/src/main/resources-nightly/META-INF/experiments.xml
+++ b/src/main/resources-nightly/META-INF/experiments.xml
@@ -21,7 +21,7 @@
         <experimentalFeature id="org.rust.resolve.new.engine" percentOfUsers="100">
             <description>Enables experimental resolve engine</description>
         </experimentalFeature>
-        <experimentalFeature id="org.rust.crates.local.index" percentOfUsers="0">
+        <experimentalFeature id="org.rust.crates.local.index" percentOfUsers="100">
             <description>Enables crates local index</description>
         </experimentalFeature>
     </extensions>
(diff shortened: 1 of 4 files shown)


toml/src/test/kotlin/org/rust/toml/completion/CargoTomlCompletionTestBase.kt
```diff
@@ -28,12 +31,18 @@ abstract class CargoTomlCompletionTestBase : RsTestBase() {
...
         @Language("TOML") after: String
-    ) = completionFixture.doSingleCompletionByFileTree(fileTree, after)
+    ) {
+        completionFixture.doSingleCompletionByFileTree(fileTree, after)
+    }
 
-    protected fun checkNoCompletion(@Language("TOML") code: String) = completionFixture.checkNoCompletion(code)
+    protected fun checkNoCompletion(@Language("TOML") code: String) {
+        completionFixture.checkNoCompletion(code)
+    }
```
Undin, line 47: Redundant changes

toml/src/test/kotlin/org/rust/toml/completion/CargoTomlCompletionTestBase.kt
```diff
@@ -8,14 +8,17 @@ package org.rust.toml.completion
...
 import org.rust.lang.core.completion.RsCompletionTestFixture
+import org.rust.openapiext.setFeatureEnabled
 
 abstract class CargoTomlCompletionTestBase : RsTestBase() {
 
     protected lateinit var completionFixture: RsCompletionTestFixture
 
     override fun setUp() {
         super.setUp()
+        setFeatureEnabled(RsExperiments.CRATES_LOCAL_INDEX, false)
```
Undin, line 21: Why do we need to disable `CRATES_LOCAL_INDEX` feature in all inheritors of `CargoTomlCompletionTestBase`?
As I can see, all tests pass if we drop this change

Undin, line 21: BTW if we actually need this, we need to restore the default value on test finish

Undin (changes requested):

PR #7427, opened 2021-06-29
Reset crates local index state if index is empty

avrong:
Reset the last commit hash of the index if the index PersistentHashMap size is 0.

This will fix the issues when the index file has been deleted or did not move with the state, e.g. when the latest IDE release copied the state without index map


Undin (changes requested):

Undin (approved):
bors r+

PR #7429, opened 2021-06-29
Introduce JSON schemas support for TOML

avrong:
changelog: TODO

### DIFF
--- a/intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
+++ b/intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
@@ -1,0 +1,125 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.toml.ide.schema
+
+import com.intellij.json.pointer.JsonPointerPosition
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.parentOfType
+import com.intellij.util.ThreeState
+import com.jetbrains.jsonSchema.extension.JsonLikePsiWalker
+import com.jetbrains.jsonSchema.extension.adapters.JsonPropertyAdapter
+import com.jetbrains.jsonSchema.extension.adapters.JsonValueAdapter
+import org.toml.lang.psi.*
+
+class TomlJsonPsiWalker private constructor() : JsonLikePsiWalker {
+    override fun isName(element: PsiElement?): ThreeState {
+        if (element is TomlKey || element is TomlKeyValue) return ThreeState.YES;
+        return ThreeState.NO;
+    }
+
+    override fun isPropertyWithValue(element: PsiElement): Boolean = element is TomlKeyValue && element.value != null
+
+    override fun findElementToCheck(element: PsiElement): PsiElement? {
+        var current: PsiElement? = element
+
+        while (current != null && current !is PsiFile) {
+            if (current is TomlValue || current is TomlKeyValue) return current
+            current = current.parent
+        }
+
+        return null
+    }
+
+    override fun findPosition(element: PsiElement, forceLastTransition: Boolean): JsonPointerPosition? {
+        val pos = JsonPointerPosition()
+        var current = element
+
+        while (current !is PsiFile) {
+            val position = current
+            current = current.parent
+
+            if (current is TomlArray) {
+                var requiredIndex = -1
+
+                for ((index, value) in current.elements.withIndex()) {
+                    if (value == position) {
+                        requiredIndex = index
+                        break
+                    }
+                }
+
+                pos.addPrecedingStep(requiredIndex)
+            } else if (current is TomlKeyValue) {
+                val propertyName = current.key.text
+
+                if (current.parent !is TomlKeyValueOwner) return null // incorrect syntax?
+
+                // if either value or not first in the chain - needed for completion variant
+                if (position != element || forceLastTransition) {
+                    pos.addPrecedingStep(propertyName)
+                }
+            } else if (current is TomlKeyValueOwner && current is TomlHeaderOwner) {
+                // if either value or not first in the chain - needed for completion variant
+//                if (position != element || forceLastTransition) {
+                    val propertyName = current.header.key?.text ?: break
+                    pos.addPrecedingStep(propertyName)
+//                }
+            } else return if (current is PsiFile) {
+                break
+            } else {
+                null // something went wrong
+            }
+        }
+
+        return pos
+    }
+
+    override fun requiresNameQuotes(): Boolean = false
+
+    override fun allowsSingleQuotes(): Boolean = false
+
+    override fun hasMissingCommaAfter(element: PsiElement): Boolean = false
+
+    override fun getPropertyNamesOfParentObject(originalPosition: PsiElement, computedPosition: PsiElement?): Set<String> {
+        val table = originalPosition.parentOfType<TomlTable>() ?: originalPosition.prevSibling as? TomlTable ?: return emptySet()
+        return TomlObjectAdapter(table).propertyList.mapNotNull { it.name }.toSet()
+    }
+
+    override fun getParentPropertyAdapter(element: PsiElement): JsonPropertyAdapter? {
+        val property = element.parentOfType<TomlKeyValue>(true) ?: return null
+        return TomlPropertyAdapter(property)
+    }
+
+    override fun isTopJsonElement(element: PsiElement): Boolean = element is TomlFile
+
+    override fun createValueAdapter(element: PsiElement): JsonValueAdapter? =
+        if (element is TomlElement) TomlPropertyAdapter.createAdapterByType(element) else null
+
+    override fun getRoots(file: PsiFile): Collection<PsiElement> {
+        if (file !is TomlFile) return emptyList()
+
+        val roots = hashSetOf<PsiElement>()
+        roots.addAll(file.children)
+        return roots
+    }
+
+    override fun getPropertyNameElement(property: PsiElement?): PsiElement? = (property as? TomlKeyValue)?.key
+
+    override fun acceptsEmptyRoot(): Boolean = true
+
+    override fun getDefaultArrayValue(): String = ""
+
+    override fun getDefaultObjectValue(): String = ""
+
+    override fun getNodeTextForValidation(element: PsiElement?): String {
+        return super.getNodeTextForValidation(element)
+    }
+
+    companion object {
+        val INSTANCE = TomlJsonPsiWalker()
+    }
+}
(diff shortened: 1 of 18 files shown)


intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonSchemaCompletionContributor.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    companion object {
+        fun doCompletion(parameters: CompletionParameters, result: CompletionResultSet, rootSchema: JsonSchemaObject, stop: Boolean) {
+            val completionPosition = parameters.originalPosition ?: parameters.position
+            val worker = Worker(rootSchema, parameters.position, completionPosition, result)
+
+            worker.work()
+
+            if (stop && worker.variants.isNotEmpty()) {
+                result.stopHere()
```
Undin, line 126: Why do we want to abort all other completion contributors after this one?

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonSchemaCompletionContributor.kt
```diff
@@ -0,0 +1,141 @@
...
+            for (variant in variants) {
+                knownNames.add(variant)
+                addPropertyVariant(variant, schemaProperties[variant])
+            }
+        }
+
+        @Suppress("NAME_SHADOWING")
+        private fun addPropertyVariant(key: String, jsonSchemaObject: JsonSchemaObject?) {
+            val currentVariants = JsonSchemaResolver(project, jsonSchemaObject!!).resolve()
+            val jsonSchemaObject = ObjectUtils.coalesce(ContainerUtil.getFirstItem(currentVariants), jsonSchemaObject)
```
Undin, line 96: I suppose `?:` is more expected here than `ObjectUtils.coalesce`

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        return null
+    }
+
+    override fun findPosition(element: PsiElement, forceLastTransition: Boolean): JsonPointerPosition {
+        val position = JsonPointerPosition()
+        var current = element
+
+        while (current !is PsiFile) {
```
Undin, line 42: Looks like this code doesn't take into account position inside arrays and array table

intellij-toml/core/src/main/resources/META-INF/experiments.xml
```diff
@@ -0,0 +1,7 @@
+<idea-plugin>
+    <extensions defaultExtensionNs="com.intellij">
+        <experimentalFeature id="org.toml.json.schema" percentOfUsers="0">
```
Undin, line 3: Since we are going to move this code into IDEA repo, let's use registry instead

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    override fun createValueAdapter(element: PsiElement): JsonValueAdapter? =
+        if (element is TomlElement) TomlPropertyAdapter.createAdapterByType(element) else null
+
+    override fun getRoots(file: PsiFile): Collection<PsiElement> {
+        if (file !is TomlFile) return emptyList()
+
+        val roots = hashSetOf<PsiElement>()
+        roots.addAll(file.children)
+        return roots
```
Undin, line 93:
```suggestion
        return file.children.toHashSet()
```

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.intellij.util.ThreeState
+import com.jetbrains.jsonSchema.extension.JsonLikePsiWalker
+import com.jetbrains.jsonSchema.extension.adapters.JsonPropertyAdapter
+import com.jetbrains.jsonSchema.extension.adapters.JsonValueAdapter
+import org.toml.lang.psi.*
+
+class TomlJsonPsiWalker private constructor() : JsonLikePsiWalker {
+    override fun isName(element: PsiElement?): ThreeState {
+        if (element is TomlKeySegment) return ThreeState.YES
+        return ThreeState.NO
```
Undin, line 21: nit:
```suggestion
        return if (element is TomlKeySegment) ThreeState.YES else ThreeState.NO
```

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    override fun findElementToCheck(element: PsiElement): PsiElement? {
+        var current: PsiElement? = element
+
+        while (current != null && current !is PsiFile) {
+            if (current is TomlElement) return current
+            current = current.parent
+        }
+
+        return null
```
Undin, line 35:
```suggestion
        return element.parentOfType<TomlElement>()
```

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonPsiWalker.kt
```diff
@@ -0,0 +1,111 @@
...
+        val roots = hashSetOf<PsiElement>()
+        roots.addAll(file.children)
+        return roots
+    }
+
+    override fun getPropertyNameElement(property: PsiElement?): PsiElement? = (property as? TomlKeyValue)?.key
+
+    override fun getNodeTextForValidation(element: PsiElement?): String {
+        return super.getNodeTextForValidation(element)
+    }
```
Undin, line 100: Looks redundant

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonSchemaCompletionContributor.kt
```diff
@@ -0,0 +1,141 @@
...
+
+            worker.work()
+
+            if (stop && worker.variants.isNotEmpty()) {
+                result.stopHere()
+            }
+        }
+
+        @TestOnly
+        fun getCompletionVariants(
```
Undin, line 131: It seems we don't need this code due to suggested changes in test.
If it's so, don't forget drop it and related code and provide proper visibility modifiers for `Worker` class

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlJsonSchemaCompletionContributor.kt
```diff
@@ -0,0 +1,141 @@
...
+            schema: JsonSchemaObject,
+            position: PsiElement,
+            originalPosition: PsiElement
+        ): List<LookupElement> {
+            val result = mutableListOf<LookupElement>()
+            Worker(schema, position, originalPosition) { element: LookupElement? -> result.add(element!!) }.work()
```
Undin, line 137: I understand that we will drop this code but anyway: let's try to avoid `!!` in code
If we really expect not null object somewhere but can't express it via type system, use `error` with an explicit message

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlPropertyAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.jetbrains.jsonSchema.extension.adapters.JsonObjectValueAdapter
+import com.jetbrains.jsonSchema.extension.adapters.JsonPropertyAdapter
+import com.jetbrains.jsonSchema.extension.adapters.JsonValueAdapter
+import org.toml.lang.psi.TomlArray
+import org.toml.lang.psi.TomlElement
+import org.toml.lang.psi.TomlKeyValue
+import org.toml.lang.psi.TomlKeyValueOwner
+
+class TomlPropertyAdapter(private val tomlKeyValue: TomlKeyValue) : JsonPropertyAdapter {
+    override fun getName(): String = tomlKeyValue.key.text
```
Undin, line 18: What's about multi-segment keys?

intellij-toml/schema/src/main/resources/META-INF/schema-only.xml
```diff
@@ -0,0 +1,11 @@
+<idea-plugin>
```
Undin, line 1:
```suggestion
<idea-plugin>

    <depends>com.intellij.modules.json</depends>
```

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlPropertyAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun getParentObject(): JsonObjectValueAdapter? {
+        val parent = tomlKeyValue.parent
+        if (parent !is TomlKeyValueOwner) return null
+
+        return TomlObjectAdapter(parent)
+    }
+
+    companion object {
+        fun createAdapterByType(value: TomlElement) = when (value) {
```
Undin, line 37: Explicit returning type is missing here

intellij-toml/schema/src/main/kotlin/org/toml/ide/schema/TomlPropertyAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.intellij.psi.PsiElement
+import com.jetbrains.jsonSchema.extension.adapters.JsonObjectValueAdapter
+import com.jetbrains.jsonSchema.extension.adapters.JsonPropertyAdapter
+import com.jetbrains.jsonSchema.extension.adapters.JsonValueAdapter
+import org.toml.lang.psi.TomlArray
+import org.toml.lang.psi.TomlElement
+import org.toml.lang.psi.TomlKeyValue
+import org.toml.lang.psi.TomlKeyValueOwner
+
+class TomlPropertyAdapter(private val tomlKeyValue: TomlKeyValue) : JsonPropertyAdapter {
```
Undin, line 17: Maybe it's worth creation some base (sealed?) class like 

```kotlin
sealed class TomlPropertyAdapterBase<T : TomlElement>(protected val element: T) : JsonPropertyAdapter {
    override fun getDelegate(): PsiElement = element

    companion object {
        fun createAdapterByType(value: TomlElement): TomlPropertyAdapterBase<*> = when (value) {
            is TomlKeyValueOwner -> TomlObjectAdapter(value)
            is TomlArray -> TomlArrayAdapter(value)
            else -> TomlGenericValueAdapter(value)
        }
    }
}
```

and inherit all other toml adapters from it?

Undin (changes requested):
Note, I haven't checked properly yet how it works. Just wanted to provide some quick feedback for the most suspicious parts

PR #7432, opened 2021-06-30
Confirm loading untrusted Cargo project

Undin:
Show [untrusted project](https://www.jetbrains.com/help/idea/2021.1/project-security.html) dialog to ensure that project is safe to invoke rust tools like `cargo`/`rustc`/`rustfmt` that can be overridden to execute arbitrary (probably malicious) code and execute build scripts/procedural macros that also can contain arbitrary code

Screen Shot 2021-06-30 at 23 36 41

In the case of "Preview in Safe Mode" choice, the plugin won't load any info about a project by default like `cargo metadata`, `rustc -V`, `rustc --cfg`, etc.
The plugin shows editor notification in Rust and Cargo.toml files
Screen Shot 2021-06-30 at 23 47 43

Also, for some dangerous actions (not all) the plugin now shows the following dialog to continue execution
Screen Shot 2021-06-30 at 23 58 08
Currently, it's implemented only for `Build` and `Refresh Cargo Projects` actions. For other actions it will be implemented in separate PR

See https://github.com/jonas-schievink/mallory as an example of Rust project that can "steal" your private information in various way

changelog: Confirm loading untrusted Cargo project

### DIFF
--- a/src/main/kotlin/org/rust/ide/notifications/RsUntrustedNotificationProvider.kt
+++ b/src/main/kotlin/org/rust/ide/notifications/RsUntrustedNotificationProvider.kt
@@ -1,0 +1,42 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.notifications
+
+import com.intellij.ide.impl.UntrustedProjectEditorNotificationPanel
+import com.intellij.ide.impl.isTrusted
+import com.intellij.ide.impl.setTrusted
+import com.intellij.openapi.fileEditor.FileEditor
+import com.intellij.openapi.project.DumbAware
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.Key
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.ui.EditorNotificationPanel
+import com.intellij.ui.EditorNotifications
+import org.rust.cargo.project.model.cargoProjects
+import org.rust.cargo.project.model.isCargoToml
+import org.rust.lang.core.psi.isRustFile
+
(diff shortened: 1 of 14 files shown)


src/main/kotlin/org/rust/ide/notifications/RsUntrustedNotificationProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+    override fun createNotificationPanel(file: VirtualFile, fileEditor: FileEditor, project: Project): EditorNotificationPanel? {
+        if (project.isTrusted()) return null
+        if (!(file.isRustFile || file.isCargoToml)) return null
+
+        val cargoProjects = project.cargoProjects
+        if (!cargoProjects.initialized) return null
+
+        return UntrustedProjectEditorNotificationPanel(project, fileEditor) {
+            project.setTrusted(true)
+            cargoProjects.refreshAllProjects()
```
vlad20012, line 37: Is it possible to set `setTrusted(true)` from a different place, e.g. from settings? If yes, maybe we should invoke `cargoProjects.refreshAllProjects()` from `TrustChangeNotifier` ?

Undin: It's not possible from settings (at least, I don't know how) but it's possible from similar things in other build systems (if project is complex, it can contain different build systems), so I've extracted the notifier into a separate file and call `cargoProjects.refreshAllProjects()` from it.
Note, it leads to a double call of `refreshAllProjects` if you mark a project as trusted from dialog shown during invocation of `Reload Cargo Projects` action but I think it's ok since it will be only once

vlad20012 (approved):
bors r+

vlad20012:
There is a compilation error on 211

Undin:
bors r=vlad20012

Issue #7409, opened 2021-06-25
Suggest `match to if-let` intention in more cases


avrong:
The current intention behavior: convert match with a single non-empty arm to if let.

The situations below can also be handled:

### 1. One empty arm

```rust
match something {
    Some(_) => {}
}
```

```rust
if let Some(_) = something {

}
```

### 2. Two exhaustive arms

```rust
match something {
    Some(_) => {}
    None => {}
}
```

```rust
if let Some(_) = something {

} else {
    
}
```

#### For example, `if let to match` intention behaves this way, but the opposite can't be applied:

```rust
if let Some(_) = something {

}
```

```rust
match something {
    Some(_) => {}
    None => {}
}
```

Kobzol:
I suppose that this can be done fully general now with irrefutable patterns. Something like
```rust
match e {
    E::V1(_) => 1,
    E::V2(_) => 1 + 2,
    E::V3(_) => 3
}
// v
if let E::V1(_) = e {
   1
} else if let E::V2(_) = e {
   1 + 2
} else {
   3
}
```
I'll send a PR next week.

Undin:
Since #9396 reverted changes introduced in #7435, I reopen this issue

PR #7435, opened 2021-07-01
INT: generalize MatchToIfLetIntention to arbitrary arm count

Kobzol:
This PR generalizes `MatchToIfLetIntention` so that it now works for almost all match expressions.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/7409

changelog: The intention to convert a match expression to a series of if let statements is now offered in more situations.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
@@ -10,7 +10,6 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.ext.ancestorStrict
-import org.rust.lang.core.psi.ext.getNextNonCommentSibling
 
 class MatchToIfLetIntention : RsElementBaseIntentionAction<MatchToIfLetIntention.Context>() {
     override fun getText() = "Convert match statement to if let"
@@ -19,8 +18,7 @@ class MatchToIfLetIntention : RsElementBaseIntentionAction<MatchToIfLetIntention
     data class Context(
         val match: RsMatchExpr,
         val matchTarget: RsExpr,
-        val nonVoidArm: RsMatchArm,
-        val pat: RsPat
+        val matchBody: RsMatchBody
     )
 
     override fun findApplicableContext(project: Project, editor: Editor, element: PsiElement): Context? {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/rust/ide/intentions/MatchToIfLetIntentionTest.kt
```diff
@@ -80,13 +59,44 @@ class MatchToIfLetIntentionTest : RsIntentionTestBase(MatchToIfLetIntention::cla
         fn main() {
             let color = OptionColor::Color(255, 255, 255);
 
-            if let _ = color { print!("No color") };
+            if let OptionColor::Color(255, 255, 255) = color { 1 } else if let OptionColor::Color(_, _, _) = color { 2 } else if let OptionColor::NoColor = color { 3 };
```
dima74, line 62:
```suggestion
    if let OptionColor::Color(255, 255, 255) = color {
        1
    } else if let OptionColor::Color(_, _, _) = color {
        2
    } else if let OptionColor::NoColor = color { 
        3
    };
```

I think we should make if blocks multiline

src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
```diff
@@ -19,38 +18,57 @@ class MatchToIfLetIntention : RsElementBaseIntentionAction<MatchToIfLetIntention
...
+            when {
+                it is RsUnitExpr -> true
+                it is RsBlockExpr && it.block.children.isEmpty() -> true
+                else -> false
+            }
+        } ?: false
+
+        val text = buildString {
+            for ((index, arm) in arms.withIndex()) {
+                var armText = "if let ${arm.pat.text} = ${ctx.matchTarget.text}"
```
dima74, line 49: I think we should use `==` when possible, e.g `if x == 1` instead of `if let 1 = x`:

```rust
fn func(X: i32) {
    match x {
        1 => println!("1"),
        _ => println!("2"),
    };
}
```

Kobzol: Would you mind if we leave this for a follow-up PR? It doesn't seem trivial and it's not strictly related to the generalization done by this PR.

src/test/kotlin/org/rust/ide/intentions/MatchToIfLetIntentionTest.kt
Kobzol: I wonder how to do this cleanly. It is currently done for normal expressions, but if the input expression was already a block expression, it has braces, so we need to somehow take the contents inside the braces out of it. I did it in a very crude way, I hope that it's ok, since match arm braces usually don't contain things like `move`, `unsafe` or attributes on the block.

src/main/kotlin/org/rust/ide/intentions/MatchToIfLetIntention.kt
dima74: Ok, sure

src/test/kotlin/org/rust/ide/intentions/MatchToIfLetIntentionTest.kt
dima74: I think your approach is fine

dima74 (approved):
Thanks!

Undin (commented):
Unfortunately, since the intention produces incorrect code and the release will be quite soon (at 26.09 in accordance with our plan), I want to revert these changes not to introduce bug in new stable build

src/test/kotlin/org/rust/ide/intentions/MatchToIfLetIntentionTest.kt
```diff
@@ -25,39 +25,51 @@ class MatchToIfLetIntentionTest : RsIntentionTestBase(MatchToIfLetIntention::cla
...
+            let color = OptionColor::Color(255, 255, 255);
+
+            if let OptionColor::Color(255, 255, 255) = color {
+                1
+            } else if let OptionColor::Color(_, _, _) = color {
+                2
+            } else if let OptionColor::NoColor = color {
+                3
+            };
         }
```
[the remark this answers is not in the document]
Kobzol: Hmm, I wonder what should we do in this case. Introduce an artificial `else` branch? I don't think that we can just use `else` for the last branch, since the `let` pattern might introduce some variables used in the corresponding match arm.

Issue #7419, opened 2021-06-27
"Create struct" intention generates a struct in another struct impl block


lancelote:
## Environment

* **IntelliJ Rust plugin version:** 0.4.150.3950-212-nightly
* **Rust toolchain version:** 1.52.0-nightly (b3e19a221 2021-03-12) x86_64-apple-darwin
* **IDE name and version:** CLion 2021.3 EAP (CL-213.516)
* **Operating system:** macOS 10.15.7
* **Macro expansion engine:** new
* **Name resolution engine:** new

## Problem description

"Create struct" intention from #6837 generates a struct in another struct `impl` block when called inside a method.

## Steps to reproduce

```rust
struct Foo {
    bar: i32,
}

impl Foo {
    fn foo() {
        Baz { a, b: true }
    }
}

fn main() {
    println!("Hello, world!");
}
```

- Place caret at `Baz`
- Invoke "Create struct Baz"

```rust
struct Foo {
    bar: i32,
}

impl Foo {
    struct Baz {
        a: _,
        b: bool
    }
    fn foo() {
        Baz { a, b: true }
    }
}

fn main() {
    println!("Hello, world!");
}
```

PR #7436, opened 2021-07-01
FIX: create structs outside if invoked from inside an impl block

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7419

changelog: The intention to create structs from unresolved usage now creates the structs outside if invoked from inside an impl block.

### DIFF
--- a/src/main/kotlin/org/rust/ide/intentions/createFromUsage/utils.kt
+++ b/src/main/kotlin/org/rust/ide/intentions/createFromUsage/utils.kt
@@ -67,7 +67,8 @@ fun getTargetItemForFunctionCall(path: RsPath): CallableInsertionTarget? {
 
 fun insertStruct(targetModule: RsMod, struct: RsStructItem, sourceFunction: RsElement): RsStructItem {
     if (targetModule == sourceFunction.containingMod) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/intentions/createFromUsage/utils.kt
```diff
@@ -67,7 +67,8 @@ fun getTargetItemForFunctionCall(path: RsPath): CallableInsertionTarget? {
 
 fun insertStruct(targetModule: RsMod, struct: RsStructItem, sourceFunction: RsElement): RsStructItem {
     if (targetModule == sourceFunction.containingMod) {
-        return sourceFunction.parent.addBefore(struct, sourceFunction) as RsStructItem
+        val anchor = sourceFunction.ancestors.firstOrNull { it in targetModule.children } ?: sourceFunction
```
vlad20012, line 70:
```suggestion
        val anchor = sourceFunction.ancestors.firstOrNull { it.parent == targetModule } ?: sourceFunction
```
?

Kobzol: :man_facepalming: Yes, that's what I was looking for :D Thanks.

PR #7437, opened 2021-07-02
PERF: optimize local crate index parsing

Kobzol:
I noticed that the local crate index takes a while to load, so I tried to profile it to see what's going on. I made two changes in two separate commits:

1) The object mapper for deserializing JSON was being created time and time again. It takes some instructions to initialize it and (most importantly), if it's created from scratch each time, it cannot leverage any proper caching of deserializers. This change reduces the time to run the `test tokio features` test from ~45s to ~22s on my notebook.

Before:
before

After:
after

2) Since loading and deserializing the versions take a lot of time, I tried to parallelize it, by first walking the tree to collect all object IDs, then deserializing them in parallel. This cut the time further from ~22s to ~10s on my notebook with 4 cores (8 hyper-threads). It used the default parallel pool, which uses 7 threads on my notebook (number of virtual cores - 1), if I'm not mistaken.
Without the first commit (with only parallelization), the test takes around 22s.

The parallelization might be a bit overkill and probably isn't really needed, but I think that 1) is a net win and should be changed.

Btw, `VFileEvent.pathEndsWith` seems to be unused.

changelog: Optimize creation of local crate index.

### DIFF
--- a/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
+++ b/toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
@@ -6,6 +6,7 @@
 package org.rust.toml.crates.local
 
 import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.json.JsonMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.fasterxml.jackson.module.kotlin.registerKotlinModule
@@ -22,7 +23,6 @@ import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapi.vfs.newvfs.BulkFileListener
 import com.intellij.openapi.vfs.newvfs.RefreshQueue
 import com.intellij.openapi.vfs.newvfs.events.VFileEvent
-import com.intellij.openapi.vfs.newvfs.events.VFilePropertyChangeEvent
 import com.intellij.util.io.*
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.api.errors.GitAPIException
@@ -37,12 +37,16 @@ import org.eclipse.jgit.treewalk.filter.PathFilter
 import org.eclipse.jgit.treewalk.filter.TreeFilter
 import org.rust.openapiext.RsPathManager
 import org.rust.stdext.cleanDirectory
+import org.rust.stdext.supplyAsync
 import org.rust.toml.crates.local.CratesLocalIndexServiceImpl.Companion.CratesLocalIndexState
 import java.io.*
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
+import java.util.concurrent.Executors
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.stream.Collectors
+
 
 /**
  * Crates local index, created from user cargo registry index on host machine.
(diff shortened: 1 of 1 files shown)


vlad20012:
About parallelization: `parallelStream` implicitly uses the global ForkJoinPool instance. There are several issues with it:
1. It is not intended to be used in IO operations.
2. In the IntelliJ platform it is heavily used in highlighting and other stuff. For example, all inspections run from the global ForkJoinPool. This means that heavy or IO tasks submitted to the pool may freeze the highlighting.
3. Tasks that execute on ForkJoinPool may interfere. This is a complex topic... Let's say that during the execution of a task ForkJoinPool may start to execute another task. Of course, it can't do this at a random point, but it can do this when interacting with ForkJoinPool API, for e.g. `ForkJoinTask.get`. This actually means that your code inside `.map {}` may be executed inside a read action, because the platform actually invokes `ForkJoinTask.get` inside a read action. That, in turn, may hang the IDE because of heavy operation inside a read action without cancellation checks =/

In short, we must not use the global ForkJoinPool instance.
I this case it's better to run a new ForkJoinPool and execute the stream on it.
```
val pool = Executors.newWorkStealingPool()
val future = supplyAsync(pool) {
    objectIds
        .parallelStream()
        .map { ... }
        .forEach { ... }        
}
try {
    future.join()
} finally {
    pool.shutdownNow()
}
```
You can also play with a worker thread count. I expect that performance will not increase after some amount of threads, say, 2 or 3.

Kobzol:
I suspected that the global pool is not a good idea, but I had some trouble getting the local pool to work, I'm not very familiar with the API and I didn't find many examples in Kotlin. I'll check how the performance changes with varying thread counts.

Kobzol:
Work stealing pool:
- 1 thread: ~23s
- 2 threads: ~15s
- 3 threads: ~13s
- 4 threads: ~12s

I also wanted to try the fixed thread pool, since work stealing often has higher overhead than just simple scheduling. But if I use `Executors.newFixedThreadPool(<any number>)`, it seems to run for the same time and use all of the available cores :man_shrugging:

Iterating the git tree takes about 4 seconds and storing the results into the persistent hash map takes about 3 seconds. Therefore the actual file and JSON parsing part is only about a third of the total computation time and therefore there is probably no point in increasing the number of threads further.

I hope that `repository.open(objectId)` is thread-safe :)

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -215,37 +220,57 @@ class CratesLocalIndexServiceImpl
...
+                objectIds.add(objectId to treeWalk.nameString)
+            }
+            val pool = Executors.newWorkStealingPool(2)
+            val future = supplyAsync(pool) {
+                objectIds
+                    .parallelStream()
+                    .map { (objectId, name) ->
+                        val loader = repository.open(objectId)
+                        val versions = mutableListOf<CargoRegistryCrateVersion>()
+                        val fileReader = BufferedReader(InputStreamReader(loader.openStream()))
```
vlad20012, line 248:
```suggestion
                        val fileReader = BufferedReader(InputStreamReader(loader.openStream(), Charsets.UTF_8))
```
I accidentally noticed that we forget to specify `Charsets.UTF_8` charset here(

vlad20012 (approved):
> I also wanted to try the fixed thread pool, since work stealing often has higher overhead than just simple scheduling. But if I use `Executors.newFixedThreadPool(<any number>)`, it seems to run for the same time and use all of the available cores :man_shrugging:

`parallelStream` works only with `ForkJoinPool` (wich is `newWorkStealingPool`). If `parallelStream` is invoked from a thread from another type of pool, `parallelStream` uses the global ForkJoinPool. Java magic :man_shrugging:

toml/src/main/kotlin/org/rust/toml/crates/local/CratesLocalIndexServiceImpl.kt
```diff
@@ -215,37 +220,57 @@ class CratesLocalIndexServiceImpl
...
+                            } catch (e: Exception) {
+                                LOG.warn("Failed to parse JSON from ${treeWalk.pathString}, line ${line}: ${e.message}")
+                            }
+                        }
+                        name to CargoRegistryCrate(versions)
+                    }
+                    .collect(Collectors.toList())
+            }
+            try {
+                val crateList = future.join()
```
vlad20012, line 264: By the way, the `try` block could be closed right after `future.join()` invocation since we doesn't use the pool later

Kobzol:
@avrong Do you have any comments or can this be merged? :)

avrong (approved):
LGTM. Thanks @vlad20012

PR #7438, opened 2021-07-02
COMP: reuse MembersGenerator in RsImplTraitMemberCompletionProvider

Kobzol:
Improves https://github.com/intellij-rust/intellij-rust/pull/6342 by reusing the `MembersGenerator` class.

### DIFF
--- a/src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
+++ b/src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
@@ -8,7 +8,6 @@ package org.rust.ide.refactoring.implementMembers
 import com.intellij.codeInsight.hint.HintManager
 import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.editor.Editor
-import com.intellij.openapi.editor.ScrollType
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
 import org.rust.ide.presentation.ImportingPsiRenderer
@@ -177,7 +176,7 @@ private fun createExtraWhitespacesAroundFunction(left: PsiElement, right: PsiEle
     return RsPsiFactory(left.project).createWhitespace("\n".repeat(extraLineCount))
 }
 
-private class MembersGenerator(
+class MembersGenerator(
     private val factory: RsPsiFactory,
     impl: RsImplItem,
     private val trait: BoundElement<RsTraitItem>,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
```diff
@@ -190,26 +189,38 @@ private class MembersGenerator(
...
+            else ->
+                error("Unknown trait member")
+        }
+    }
+
     private fun RsFunction.renderFunctionSignature(): String = renderer.renderFunctionSignature(this)
     private fun RsTypeReference.renderTypeReference(): String = renderer.renderTypeReference(this)
 }
+
+private fun String.withSpace(): String {
```
vlad20012, line 220: Hmm, when is it possible that `renderFunctionSignature` ends with space?

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
```diff
@@ -128,22 +108,18 @@ private fun completeType(target: RsTypeAlias): LookupElementBuilder {
 private fun completeFunction(
     target: RsFunction,
     impl: RsImplItem,
-    substitution: RsPsiSubstitution
+    substitution: RsPsiSubstitution,
+    memberGenerator: MembersGenerator
 ): LookupElementBuilder {
-    val renderer = ImportingPsiRenderer(PsiRenderingOptions(renderGenericsAndWhere = true), substitution, impl)
     val shortRenderer = ImportingPsiRenderer(PsiRenderingOptions(renderGenericsAndWhere = false), substitution, impl)
```
vlad20012, line 114: By the way, `ImportingPsiRenderer` looks too heavy for the completion presentation. Maybe it's better to use simple `PsiSubstitutingPsiRenderer` (with default `PsiRenderingOptions.shortPaths = true` option)?

vlad20012 (approved):

src/main/kotlin/org/rust/ide/refactoring/implementMembers/impl.kt
Kobzol: Oh, it's not. I was just surprised that the space was added automatically in implement members fix, but not here in completion. Probably in implement members the auto formatter is run automatically.

src/main/kotlin/org/rust/lang/core/completion/RsImplTraitMemberCompletionProvider.kt
Kobzol: Good point, using the importing renderer in presentation is probably not a good idea. I changed it per your suggestion.

Kobzol:
bors r=vlad20012

PR #7450, opened 2021-07-04
TY: improve inference of references to constants in const evaluation

Kobzol:
Fixes: https://github.com/intellij-rust/intellij-rust/issues/4738

changelog: Improve type inference of array sizes.

### DIFF
--- a/src/main/kotlin/org/rust/lang/utils/evaluation/ConstExprBuilder.kt
+++ b/src/main/kotlin/org/rust/lang/utils/evaluation/ConstExprBuilder.kt
@@ -34,6 +34,7 @@ private abstract class ConstExprBuilder<T : Ty, V> {
     protected abstract val resolver: PathExprResolver?
     protected abstract val RsLitExpr.value: V?
     protected abstract fun V.wrap(): ConstExpr<T>
+    protected abstract fun copyWithDefaultResolver(): ConstExprBuilder<T, V>
 
     protected fun makeLeafValue(expr: RsLitExpr): ConstExpr<T>? = expr.value?.wrap()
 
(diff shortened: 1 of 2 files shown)


mchernyavsky (approved):
Thanks!

bors r+

src/main/kotlin/org/rust/lang/utils/evaluation/Utils.kt
```diff
@@ -14,6 +14,6 @@ class PathExprResolver(resolver: (RsPathExpr) -> RsElement?): (RsPathExpr) -> Rs
         val default: PathExprResolver = PathExprResolver { it.path.reference?.resolve() }
 
         fun fromContext(ctx: RsInferenceContext): PathExprResolver =
-            PathExprResolver { ctx.getResolvedPath(it).singleOrNull()?.element }
+            PathExprResolver { ctx.getResolvedPath(it).singleOrNull()?.element ?: default.invoke(it) }
```
Undin, line 17: I don't think we can do it in this way because it may call type inference during type inference

Kobzol: Oh, I see. I was wondering whether it could be that easy :) The main problem is that the expression of the constant is not written into the inference context before the const evaluation takes place.

```rust
const FOO: usize = 2;
const COUNT: usize = FOO;
fn main() {
    let x = [1; COUNT];
    x;
  //^ [i32; 2]
}
```
When `COUNT` is inferred, the `FOO` path is not yet written into the context. It could be done e.g. by inferring the type of the constant expression in `instantiatePath`, when the `COUNT` path is being instantiated, but I'm not sure if that is a good idea.

vlad20012:
I think I fixed this properly.

bors r+

Kobzol:
Thanks for fixing it!

PR #7456, opened 2021-07-05
ANN: Annotate trailing comma in toml inline table

avrong:
changelog: Annotate trailing commas in TOML inline tables as errors

### DIFF
--- a/intellij-toml/core/src/main/kotlin/org/toml/ide/annotator/TomlAnnotator.kt
+++ b/intellij-toml/core/src/main/kotlin/org/toml/ide/annotator/TomlAnnotator.kt
@@ -5,14 +5,21 @@
 
 package org.toml.ide.annotator
 
+import com.intellij.codeInspection.InspectionManager
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.ide.annotator.AnnotatorBase
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
 import com.intellij.psi.SyntaxTraverser
 import org.toml.lang.psi.TomlArray
+import org.toml.lang.psi.TomlElementTypes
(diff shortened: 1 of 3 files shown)


intellij-toml/core/src/main/kotlin/org/toml/ide/annotator/TomlAnnotator.kt
```diff
@@ -24,5 +24,10 @@ class TomlAnnotator : AnnotatorBase() {
                 holder.newAnnotation(HighlightSeverity.ERROR, "Inline tables are intended to appear on a single line").create()
             }
         }
+
+        val parent = element.parent
+        if (element.text == "," && parent is TomlInlineTable && element.textOffset > parent.entries.last().textOffset) {
```
Undin, line 29: Let's better use `lastOrNull` here not to produce exception if there are no entries.
Yeah, looks like it's impossible for now but anyway

vlad20012, line 29: Could you please check the element type instead of its text? `psi.text` looks very suspicious and almost always there is a better way to do things.

intellij-toml/core/src/main/kotlin/org/toml/ide/annotator/TomlAnnotator.kt
```diff
@@ -24,5 +31,24 @@ class TomlAnnotator : AnnotatorBase() {
...
         }
+
+        val parent = element.parent
+        if (element.elementType == TomlElementTypes.COMMA && parent is TomlInlineTable &&
+            element.textOffset > parent.entries.lastOrNull()?.textOffset ?: 0) {
+            val message = "Remove trailing comma"
+
+            val fix = object : LocalQuickFix {
+                override fun getFamilyName(): String = message
+                override fun applyFix(project: Project, descriptor: ProblemDescriptor) = descriptor.psiElement.delete()
```
Undin, line 42:
```suggestion
                override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
                    descriptor.psiElement?.delete()
                }
```
`descriptor.psiElement` can actually return null, so let's avoid NPE here

Undin (approved):
LGTM except comment

avrong:
By the way, should I manually add the change to `<change-notes>` in `plugin.xml`?

avrong:
bors r=Undin

Issue #5025, opened 2020-02-28
pub(crate) should be completed


sirgl:
2 issues:
1. There is no separate completion item `pub(crate)`
2. Inside () `crate` is not suggested

Here suggested `crate::` which is syntactically incorrect.
![image](https://user-images.githubusercontent.com/11944514/75509899-b9708f80-5a1b-11ea-93ff-9d768b5f77f5.png)

Kobzol:
This is not yet fully resolved (not sure why GH closed it, I didn't use the words "closes" or "fixes"). In a follow-up PR, I will also improve completion to contain `pub(crate)` directly.

PR #7457, opened 2021-07-05
COMP: complete possible options inside visibility restriction

Kobzol:
This PR adds completion for things that can be put inside a visibility restriction (`self`, `super`, `crate` or `in`).
complete-vis-restriction

Partially resolves https://github.com/intellij-rust/intellij-rust/issues/5025

changelog: Add completion for items inside visibility restrictions (`pub(<...>)`).

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsVisRestrictionCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsVisRestrictionCompletionProvider.kt
@@ -1,0 +1,57 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.patterns.PlatformPatterns.psiElement
+import com.intellij.patterns.PsiElementPattern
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsVisRestrictionCompletionProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Provides completion inside visibility restriction:
+ * `pub(<here>)`
+ */
+object RsVisRestrictionCompletionProvider : RsCompletionProvider() {
+    override val elementPattern: PsiElementPattern.Capture<PsiElement>
+        get() = psiElement(RsElementTypes.IDENTIFIER)
+            .withParent(
+                psiElement<RsPath>()
+                    .with("hasOneSegment") { item, _ -> item.qualifier == null && item.path == null && !item.hasColonColon }
```
vlad20012, line 32: `item.qualifier` and `item.path` for are the same. Maybe you want `typeQual`?

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/completion/RsVisRestrictionCompletionProvider.kt
Kobzol: Actually, I wanted to check if the path has no `::<sub-path>` on the "right" side, but now I realized that due to the way paths are parsed, this is not relevant here. And `typeQual` was missed in the check, so thanks :D

Kobzol:
bors r=vlad20012

PR #7458, opened 2021-07-05
GRD: provide `testOutput` configuration in all gradle modules

Undin:
`testOutput` contains references to test classes of the corresponding modules. Allows specifying dependency between tests in different gradle modules


Undin:
CC @avrong

PR #7461, opened 2021-07-05
INSP: Fix handling `writeln!` macro in unused import inspection

dima74:
Expand format macros like `writeln` in `RsWithMacrosRecursiveElementWalkingVisitor`, so that unused import inspection can correctly handle needed imports (`std::io::Write`) for them

changelog: Correctly handle `writeln!` macro in unused import inspection

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/psi.kt
+++ b/src/main/kotlin/org/rust/openapiext/psi.kt
@@ -14,6 +14,7 @@ import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.util.containers.ContainerUtil
 import org.rust.lang.core.psi.RsMacroCall
 import org.rust.lang.core.psi.ext.expansion
+import org.rust.lang.core.psi.ext.macroName
 import org.rust.lang.core.resolve.DEFAULT_RECURSION_LIMIT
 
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/rust/openapiext/psi.kt
```diff
@@ -96,14 +98,23 @@ private class RsWithMacrosRecursiveElementWalkingVisitor(
...
+            ?: return element.resolveToMacro()?.hasRustcBuiltinMacro != true
         for (expandedElement in expansion.elements) {
             val visitor = RsWithMacrosRecursiveElementWalkingVisitor(processor, nestedMacroDepth + 1)
             expandedElement.accept(visitor)
         }
+        return true
     }
 }
+
+private val RsMacroCall.macroArgumentAny: RsElement?
```
vlad20012, line 118: This can be confused with `RsMacroCall.macroArgumentElement`. Let's choose another name or use `macroArgumentElement` (if reasonable here)

vlad20012 (approved):
Let's also bump `RustParserDefinition.PARSER_VERSION` because of changed parsing rules

dima74:
bors r=vlad20012

PR #7462, opened 2021-07-05
GRAM: fixed parsing priorities for expressions containing blocks

### LINKED ISSUE #7220: Incorrect operator precedence in mutable assignments
## Environment

* **IntelliJ Rust plugin version:** 0.3.146.3826-211
* **Rust toolchain version:** 1.54.0-nightly (bacf770f2 2021-05-05) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2021.1 (CL-211.6693.114)
* **Operating system:** Linux 5.4.0-72-generic

## Problem description

In mutable assignments, expressions which contain logical && and || operators as well as block expressions (ordinary, try, async or unsafe) are parsed incorrectly. Surprisingly, if a mutable assignment is changed into a fresh let binding, the expression is parsed correctly.

## Steps to reproduce

```rust
    let mut x = false;
    x = { 0 } != 0;
    //  ^^^^^ Error: mismatched types [E0308] expected `bool`, found `i32`
    //           ^^ Error: mismatched types [E0308] expected `()`, found `i32`
    // Parsed as: ( ( x = { 0 } ) != 0 )
    
    x = true && { 0 } == 0;
    //          ^^^^^ Error: mismatched types [E0308] expected `bool`, found `i32` 
    // Parsed as:  ( x = ( ( true && { 0 } ) == 0 ) )
    // Same errors if `==` is changed to `!=`, `>`, `<`, `>=`, `<=`, or with `||` instead of `&&`.
    
    // the following statements are parsed correctly
    x = 0 != { 0 } && true;
    x = 0 != 0;
    let x = { 0 } != 0;
    let x = true && { 0 } == 0;
    let x = 0 != { 0 } && true;
    let x = 0 != 0;
```

afetisov:
Fixes #7220

Fixed the StmtModeExprOff grammar rule which oddly interacts with GrammarKit's Pratt parser generator. Previously the actual rule contents were ignored and the rule was parsed simply as an Expr, which meant that the StmtMode flags were set incorrectly. This produced incorrect parsing behaviour for expressions containing any kind of blocks (simple, async, unsafe etc). I have changed the rule to a sequence of explicit calls to the meta rules which set/reset the parsing flags, as well as an explicit Expr recursive call. Unfortunately I couldn't make the resulting rule interact correctly with the Pratt parser generator, so the rule had to be inlined to work properly.

The commit also includes changes from Expr calls to AnyExpr calls in rules for `yield` and `return`, otherwise expressions with block would also parse incorrectly.

UnaryExpr, LambdaExpr, FullRangeExpr and OpenRangeExpr are also switched to correctly handle inner blocks.

changelog: fixed parsing of expressions and statements containing block subexpressions.


vlad20012:
It's cool!
I thought about such a solution, but I was sure this will also break GK Pratt parser generator. I'm surprised it works!

The only dangerous thing is that `setStmtMode` can be called without a subsequent `resetFlags` call. I've checked all generated setStmtMode/resetFlags pairs and found two very suspicious ones: `FullRangeExpr_0_1_0` and `OpenRangeExpr_1_0`. 
Thay contains code like this:
```java
result_ = result_ && setStmtMode(builder_, level_ + 1, StmtMode.OFF);
result_ = result_ && Expr(builder_, level_ + 1, -1);
result_ = result_ && resetFlags(builder_, level_ + 1);
```
Here, `resetFlags` can be not called if `Expr` call returns `false`. In this case, the rule is rollbacked and the parser continues the parsing in an incorrect state (but I can't make an example where it really affects parsing).

I suggest changing range rules as follows:

```
FullRangeExpr ::= Expr ( ('..' | '...' | '..=')  <<setStmtMode 'StmtMode.OFF'>> (<<checkBraceAllowed>> Expr)? <<resetFlags>>) { elementType = RangeExpr }
OpenRangeExpr ::=      ( ('..' | '...' | '..=')  <<setStmtMode 'StmtMode.OFF'>> (<<checkBraceAllowed>> Expr)? <<resetFlags>>) { elementType = RangeExpr }
```

In this case, we will always call `resetFlags`.

afetisov:
Indeed, the parser generator would break if the rule wasn't inlined. I don't know if there is an option which would tell GK to always inline the rule, it would allow to factor out the common logic.

> The only dangerous thing is that setStmtMode can be called without a subsequent resetFlags call. 

Indeed, it is a problem. It also has a potential to make the stmtFlags stack grow without limits on some very broken Rust files. I believe it is an issue not only in your example but for all affected rules, since the RHS of an expression can be malformed and put the parser in the recovery mode. Unfortunately, I don't see a good solution to that problem. Maybe duplicate the stack rollback logic in the recovery parsers? But how would we know how far to rollback the stack?

On the other hand, maybe we can ignore this edge case. After all, it would trigger only on already broken code, and with a very special kind of errors, and parsing broken code is anyway supported only on the best-effort basis.

I'll take a second look whether I can solve that problem or provide a compelling error example.

I have no idea why the CI tests failed :-( Worked on my machine >_<

vlad20012:
> I believe it is an issue not only in your example but for all affected rules, since the RHS of an expression can be malformed and put the parser in the recovery mode. 

🤨 hmm, well, I don't believe. Most of setStmtMode/resetFlags  calls are in `Expr_0` function. Typically it looks like
```
else if (priority_ < 15 && BoolAndBinExpr_0(builder_, level_ + 1)) {
    result_ = report_error_(builder_, Expr(builder_, level_, 15));
    result_ = resetFlags(builder_, level_ + 1) && result_;
    exit_section_(builder_, level_, marker_, BINARY_EXPR, result_, true, null);
}
```
I.e. we call `resetFlags` if a binary operator is parsed (in this case, `BoolAndBinExpr_0`), but regardless of `RHS` parsing result (`report_error_(builder_, Expr(builder_, level_, 15))` in this the example).

Let's look at `BoolAndBinExpr_0`:
```
...
  result_ = result_ && setStmtMode(builder_, level_ + 1, StmtMode.OFF);
  exit_section_(builder_, marker_, null, result_);
  return result_;
}
```

`setStmtMode` is the last call that can change `result_` value (actually can't because `setStmtMode()` is always `true`). 

So, logic is simple:
1. If a binary operator (`BoolAndBinExpr_0`) is parsed successfully, then `setStmtMode` **is** already called.
2. If a binary operator is parsed successfully, then `resetFlags` is called in `Expr_0` regardless of RHS parsing result.

So, we should take attention to `resetFlags` invocations outside of `Expr_0`. It is `LambdaExpr`, `UnaryExpr`, `FullRangeExpr_0` and `OpenRangeExpr`. I've checked them and it looks like only range parsing rules have such problems.

> I have no idea why the CI tests failed :-( Worked on my machine >_<

They're failing on my machine too. Parsing test output differs in whitespace PSI elements. I can send you a patch with fixes

vlad20012:
Could you, please, also increment `RustParserDefinition.PARSER_VERSION`?

vlad20012:
:+1: Thanks!
bors r+

vlad20012:
I see you're a first-time contributor. Could you, please, sign our `Contributor License Agreement` as mentioned in [CONTRIBUTING.md](https://github.com/intellij-rust/intellij-rust/blob/master/CONTRIBUTING.md#cla) and ad your username to `CONTRIBUTORS.txt`?

PR #7463, opened 2021-07-05
T: Fix pretty printers tests on 2021.2

### LINKED ISSUE #4527: Fix formatter tests for Vec
I've disabled formatter tests for `Vec` in d88a35478e550ad4f8b16065ddad80bb20fa81f0 because these tests fail with latest CLion. At the same time, it looks like it's only test issue because everything ok with real IDE.
Need to investigate the reason, fix tests and enable them again

### LINKED ISSUE #7453: Fix pretty printers tests for strings
I've temporarily disabled pretty printers tests for strings in #7430 because they started failing with the new 212 EAP.
As I can see, everything is fine in production, only tests are affected.
It would be great to investigate why tests started failing and fix them

### LINKED ISSUE #7454: Fix pretty printers tests for non zero structs
I've temporarily disabled pretty printers tests for non zero structs in #7430 because they started failing with the new 212 EAP.
As I can see, everything is fine in production, only tests are affected
It would be great to investigate why tests start failing and fix them

ortem:
Fix pretty-printers tests on 2021.2 platform.

After updating LLDB bundle to version 12 (since CLion 212.4535-EAP), LLDB is unable to resolve a breakpoint location by a file name. Fortunately, breakpoints configured with (full or relative) file paths seem to work fine. So now pretty-printers tests provide a path instead of just a file name.

For example, a previously used command
```
(lldb) breakpoint set --file 'nonzero.rs' --line 27
```
is now replaced with
```
(lldb) breakpoint set --file './tests/nonzero.rs' --line 27
```

Fixes #4527
Fixes #7453
Fixes #7454

### DIFF
--- a/pretty_printers_tests/src/test_runner.rs
+++ b/pretty_printers_tests/src/test_runner.rs
@@ -256,7 +256,7 @@ impl<'test> TestRunner<'test> for LLDBTestRunner<'test> {
         script_str.push_str(&format!("{}\n", ENABLE_RUST));
 
         // Set breakpoints on every line that contains the string "#break"
-        let source_file_name = self.src_path.file_name().unwrap().to_string_lossy();
+        let source_file_name = self.src_path.to_string_lossy();
         for line in &breakpoint_lines {
             script_str.push_str(&format!(
                 "breakpoint set --file '{}' --line {}\n",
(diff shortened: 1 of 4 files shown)


Undin (approved):
👍 
bors r+

PR #7471, opened 2021-07-06
COMP: only complete paths that resolve to traits in trait refs

Kobzol:
This PR makes path completion more precise inside trait refs. The various path filters were nested quite deeply, so I changed the implementation to apply them in a cycle.

changelog: Only complete traits in paths that must resolve to traits (e.g. `impl <...> for type` or `T: <...>`).

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
@@ -104,23 +104,26 @@ object RsCommonCompletionProvider : RsCompletionProvider() {
             is RsMetaItem -> return
             else -> {
                 val lookup = ImplLookup.relativeTo(element)
+                var filtered: RsResolveProcessor = filterPathCompletionVariantsByTraitBounds(
+                    addProcessedPathName(processor, processedPathNames),
+                    lookup
+                )
+                // Filters are applied in reverse order (the last filter is applied first)
+                val filters = listOf(
+                    ::filterCompletionVariantsByVisibility,
+                    ::filterAssocTypes,
+                    ::filterVisRestrictionPaths,
+                    ::filterTraitRefPaths
+                )
+                for (filter in filters) {
+                    filtered = filter(element, filtered)
+                }
+
                 processPathResolveVariants(
                     lookup,
                     element,
                     true,
-                    filterVisRestrictionPaths(
-                        element,
-                        filterAssocTypes(
-                            element,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
```diff
@@ -281,6 +284,28 @@ private fun filterVisRestrictionPaths(
...
+ */
+private fun filterTraitRefPaths(
+    path: RsPath,
+    processor: RsResolveProcessor
+): RsResolveProcessor {
+    val parent = path.parent
+    return if (parent is RsTraitRef) {
+        createProcessor(processor.name) {
+            when (it.element) {
+                !is RsTraitItem -> false
```
vlad20012, line 300: We should also complete modules since someone may want to type a qualified path in a trait ref

src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
```diff
@@ -104,23 +104,26 @@ object RsCommonCompletionProvider : RsCompletionProvider() {
...
+                var filtered: RsResolveProcessor = filterPathCompletionVariantsByTraitBounds(
+                    addProcessedPathName(processor, processedPathNames),
+                    lookup
+                )
+                // Filters are applied in reverse order (the last filter is applied first)
+                val filters = listOf(
+                    ::filterCompletionVariantsByVisibility,
+                    ::filterAssocTypes,
+                    ::filterVisRestrictionPaths,
+                    ::filterTraitRefPaths
```
vlad20012, line 116: Looks like re-inventing #2920 step by step... =D

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
Kobzol: Good point! Thanks.

src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt
Kobzol: Uhh :sweat_smile: I have to say that your proposal seems much more readable :) I have no idea about how "debuggable" it is though, sometimes it must be painful to step through the resolve code.

Kobzol:
bors r=vlad20012

PR #7473, opened 2021-07-07
INT: const generic parameters handling by "Add impl block" intention

### LINKED ISSUE #7465: Missing const generics when trying add impl block for struct
## Environment
* **IntelliJ Rust plugin version:** 0.4.150.3968-211
* **Rust toolchain version:** rustc 1.55.0-nightly (a435b49e8 2021-06-28) x86_64-unknown-linux-gnu
* **IDE name and version:** PhpStorm 2021.1
* **Operating system:** Ubuntu 18.04 LTS

## Problem description
image1
image2

There is also missing syntax highlight for `N`
image3

Without lifetimes:
image4

## Steps to reproduce
Press Alt + Enter on that struct
```rust
struct Cursor<'a, const N: usize>(&'a mut [u8; N]);
```

afetisov:
Closes #7465 

changelog: fixed "Add impl block" intention for structures with const generic parameters

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -261,7 +261,8 @@ class RsPsiFactory(
             ""
         } else {
             val parameterNames = typeParameterList.lifetimeParameterList.map { it.quoteIdentifier.text } +
-                typeParameterList.typeParameterList.map { it.name }
+                typeParameterList.typeParameterList.map { it.name } +
+                typeParameterList.constParameterList.map { it.name }
             parameterNames.joinToString(", ", "<", ">")
         }
 
@@ -423,12 +424,14 @@ class RsPsiFactory(
             ?: error("Failed to create pat element")
 
     fun createPatField(name: String): RsPatField =
-        createFromText("""
+        createFromText(
+            """
             struct Foo { bar: i32 }
             fn baz(foo: Foo) {
                 let Foo { $name } = foo;
             }
-        """) ?: error("Failed to create pat field")
+        """
+        ) ?: error("Failed to create pat field")
 
     fun createPatStruct(name: String, pats: List<RsPatField>, patRest: RsPatRest?): RsPatStruct {
         val pad = if (pats.isEmpty()) "" else " "
--- a/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
+++ b/src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
@@ -424,14 +424,12 @@ class RsPsiFactory(
             ?: error("Failed to create pat element")
 
     fun createPatField(name: String): RsPatField =
-        createFromText(
-            """
+        createFromText("""
             struct Foo { bar: i32 }
             fn baz(foo: Foo) {
                 let Foo { $name } = foo;
             }
-        """
-        ) ?: error("Failed to create pat field")
+        """) ?: error("Failed to create pat field")
 
     fun createPatStruct(name: String, pats: List<RsPatField>, patRest: RsPatRest?): RsPatStruct {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -423,12 +424,14 @@ class RsPsiFactory(
...
-        createFromText("""
+        createFromText(
+            """
             struct Foo { bar: i32 }
             fn baz(foo: Foo) {
                 let Foo { $name } = foo;
             }
-        """) ?: error("Failed to create pat field")
+        """
+        ) ?: error("Failed to create pat field")
```
vlad20012, line 434: it looks like an unintentional change

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
afetisov: Kinda, but this is what the Idea autoformatting does. It is tedious to wrestle with it, so I suggest either changing the formatting rules, if possible, or surrendering to it.

afetisov, line 434:
```suggestion
        """) ?: error("Failed to create pat field")
```

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -423,12 +424,14 @@ class RsPsiFactory(
             ?: error("Failed to create pat element")
 
     fun createPatField(name: String): RsPatField =
-        createFromText("""
+        createFromText(
+            """
```
afetisov, line 428:
```suggestion
        createFromText("""
```

src/main/kotlin/org/rust/lang/core/psi/RsPsiFactory.kt
```diff
@@ -261,7 +261,8 @@ class RsPsiFactory(
             ""
         } else {
             val parameterNames = typeParameterList.lifetimeParameterList.map { it.quoteIdentifier.text } +
-                typeParameterList.typeParameterList.map { it.name }
+                typeParameterList.typeParameterList.map { it.name } +
+                typeParameterList.constParameterList.map { it.name }
```
Undin, line 265: Actually, it doesn't work as expected with `const_generics` feature where const and type parameters can be mixed

vlad20012: I'll take a look after #7479

Undin: Not sure #7479 is actually needed here. We recently with @mchernyavsky  introduced `RsGenericParameter` and `org.rust.lang.core.psi.ext.RsTypeParameterListKt#getGenericParameterList` to get single list of all generic arguments for such cases

afetisov: > Actually, it doesn't work as expected with `const_generics` feature where const and type parameters can be mixed

Do you have a reference for that feature? I have always believed that the restrictions on the order of parameters are not intended to be ever lifted, for readability reasons, same as we are always required to place lifetimes before type arguments.

Undin: > Do you have a reference for that feature?

https://github.com/rust-lang/rust/pull/74953

Just try to compile the following code with nightly compiler
```rust
#![feature(const_generics)]

fn foo<const C: i32, T>() {}

fn main() {}
```

> I have always believed that the restrictions on the order of parameters are not intended to be ever lifted, for readability reasons, same as we are always required to place lifetimes before type arguments.

It's only about the order of const and type parameters. Lifetime parameters still should be placed first

PR #7328, opened 2021-06-10
COMP: use template builder to fill wild pats in MatchPostfixTemplate

Kobzol:
This PR improves https://github.com/intellij-rust/intellij-rust/pull/6825 by letting the user change the generated wild patterns (`_`) with a template builder. Based on a suggestion by @Sherlock-Holo (https://github.com/intellij-rust/intellij-rust/pull/6825#issuecomment-858216675).

changelog: Allow filling pats after generating match arms with `.match` postfix template.

PR #7476, opened 2021-07-07
COMP: always place caret into the first arm in MatchPostfixTemplate

vlad20012:
An improvement to #7328: place the caret to the first arm after finishing the template

https://user-images.githubusercontent.com/3221931/124749393-d630a780-df2c-11eb-8f2d-ee0c3f282a35.mp4

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/MatchPostfixTemplate.kt
@@ -5,6 +5,7 @@
 
 package org.rust.ide.template.postfix
 
+import com.intellij.codeInsight.template.TemplateResultListener
 import com.intellij.codeInsight.template.postfix.templates.PostfixTemplateWithExpressionSelector
 import com.intellij.openapi.editor.Editor
 import com.intellij.psi.PsiDocumentManager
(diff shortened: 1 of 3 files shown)


Kobzol (approved):
This is great! I actually wanted to do something like this in numerous occasions before, but didn't know how to do it (didn't notice that templates contain a listener).

A similar template builder could also be used for the standalone `Fill remaining arms` quick fix.

Btw, the completion (or is it a template maybe?) inside pats seems to be quite wrong (I already stumbled upon it several times), and can severely mess up the template (e.g. `<template>_</template>` -> write `a` -> completion offers `assert!` -> template generates nonsense). But that's for another PR.

vlad20012:
Well, it's possible to use macros in patterns, in theory...

Kobzol:
Well.. :sweat_smile: In theory yes, but I have yet to see a macro being used inside a pattern :) I would rather just disable these kinds of completion here.

Kobzol:
With the template listener, it would be nice to have a proper way of testing the template. Otherwise if template listener is used, it pretty much as to be combined with some `isUnitTestMode` check to fake the entered text in tests in order to test the listener behaviour.

vlad20012:
I found how to make tests for templates. I'll explain it later (when I've done with the new API)

bors r=Kobzol

vlad20012:
> Btw, the completion (or is it a template maybe?) inside pats seems to be quite wrong (I already stumbled upon it several times), and can severely mess up the template (e.g. `<template>_</template>` -> write `a` -> completion offers `assert!` -> template generates nonsense). But that's for another PR.

That was a live template. Fixed in #7547

Issue #7342, opened 2021-06-13
Generate stubs for trait implementation


shravanshetty1:
https://www.jetbrains.com/go/guide/tips/implement-an-interface/

Kobzol:
Would something like this work for you? https://github.com/intellij-rust/intellij-rust/pull/5339#issuecomment-727628728

shravanshetty1:
> Would something like this work for you? [#5339 (comment)](https://github.com/intellij-rust/intellij-rust/pull/5339#issuecomment-727628728)

Yes exactly, this saves alot of time and i find this feature highly useful while using intellij to code other languages.

Kobzol:
@vlad20012 Could you please take another look at the mentioned PR? Thanks.

PR #7480, opened 2021-07-07
INT: add intention to implement a trait for an ADT type

Kobzol:
This PR attempts to implement (a part of) https://github.com/intellij-rust/intellij-rust/pull/5339 in a more "atomic" way. The idea is to create an intention that will generate an impl block for a given type, let the user type a trait name and then automatically implement all members of that trait. Dealing with required supertraits will be left for another quick fix.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4123
Fixes: https://github.com/intellij-rust/intellij-rust/issues/7342

Closes: https://github.com/intellij-rust/intellij-rust/pull/5339

changelog: Add intention to implement a trait for a struct or enum.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/fixes/AddTypeArguments.kt
@@ -16,6 +16,10 @@ import org.rust.lang.core.psi.*
 import org.rust.lang.core.psi.RsElementTypes.COMMA
 import org.rust.lang.core.psi.RsElementTypes.LT
 import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.elementType
+import org.rust.lang.core.psi.ext.getNextNonCommentSibling
+import org.rust.lang.core.psi.ext.requiredGenericParameters
 import org.rust.openapiext.createSmartPointer
 
 class AddTypeArguments(element: RsElement) : LocalQuickFixAndIntentionActionOnPsiElement(element) {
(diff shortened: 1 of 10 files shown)


vlad20012:
Haha, it's possible!

https://user-images.githubusercontent.com/3221931/124938701-300a9d80-e011-11eb-8d91-89a2ecd9fcad.mp4

Kobzol:
Have you changed anything to make this work? It didn't work for me with `AsRef` if I remember correctly.

vlad20012:
I'm afraid I won't have time to finish the API today, so I pushed the intermediate result here
https://github.com/vlad20012/intellij-rust/commit/a0013b6bc26904d0063f5f5ae3a9f89ab0642914

I discovered the template API and found that it is very powerful, but unfortunately very messy:(
I tried to make another abstraction over it (`TemplateBuilderImpl` was a previous abstraction attempt, haha)

> What to use as the original trait name when the block is generated

The template allows replacing some placeholder with an empty text.

> It's not clear to me what to do if the entered text does not resolve to a trait.

Why not just interrupt the template? Do nothing is ok here, I guess.

> Maybe we could just attach an insert listener in completion that will generate

This would be much more complex.


Ok. About trait type arguments. As you can see, I've implemented the variant with placeholder types with subsequent "rename". There is a little issue - these placeholders are highlighted as unresolved (red) if "unresolved reference" inspection is enabled.

Kobzol:
> Why not just interrupt the template? Do nothing is ok here, I guess.

Ok, seems reasonable.

> There is a little issue - these placeholders are highlighted as unresolved (red) if "unresolved reference" inspection is enabled.

This doesn't seem like a big problem to me. The "happy path" will replace these placeholders, so in most cases the user might not even notice. And if the replacing or the template fails for some reason, it will correctly mark them as unresolved, so the user can fix it manually.

> I'm afraid I won't have time to finish the API today, so I pushed the intermediate result here

There's no rush :) Do you want to prepare/merge the new template API first and then I'll use it in this PR? Or do you want to include the template API in this PR directly?

vlad20012:
> Do you want to prepare/merge the new template API first and then I'll use it in this PR? Or do you want to include the template API in this PR directly?

The first option :)

vlad20012:
I've done it in #7512. You can rebase on it and apply this patch:
[tpl.zip](https://github.com/intellij-rust/intellij-rust/files/6802700/tpl.zip)

It's a prototype I made during designing of `RsTemplateBuilder`. I think it rather works, but I spent very little time testing it.

Kobzol:
It works nicely indeed, thank you :) I have two questions:

1) While trying out the template, I noticed that completion behaves in a weird way for un-imported (but importable) traits. Basically it only offers them only if the completion was invoked on some existing path (but not when it was invoked on an "empty" path, which happens with the template).

    completion
    
    I tracked it down to [this line](https://github.com/intellij-rust/intellij-rust/blob/master/src/main/kotlin/org/rust/lang/core/completion/RsCommonCompletionProvider.kt#L177). When the original position is empty, `originalPosition.elementType` is whitespace and its parent is not a path.
    
    It's a bit unfortunate, since implementing an out-of-scope trait would be very useful. I'm not sure how to fix this though, `addCompletionsFromIndex` looks like it really needs to get a `RsPath` from somewhere. Maybe we can fake it somehow, even if the completion location will be initially empty?

2) Now that we can create a more complex template builder, what do you think about letting the user change the contents of the generated trait items using the template builder? I think that we decided sometime ago (in the `Implement trait` quick fix) that it's not a good idea for functions, because these are typically multiline and would be annoying to fill with the template builder, but maybe for constants and types? Or maybe just move caret to the location of the first generated item once the user fully fills in the trait name and the template is finished?

Kobzol:
I pushed a new version using the new template builder API. As an experiment, I "fixed" the problem with empty path completion from the index by using `parameters.position` instead of `parameters.originalPosition`, but it feels like undoing the intent of the original code.

vlad20012:
I sure that was done intentionally because unfiltered completion (without a prefix) would be quite slow. But I didn't consider that after typing a letter the completion list is not updated. It's a bug. I've fixed this in #7516. But the fix is that the completion list will be updated after typing some characters.
The proper fix will be possible after rewriting autoimport using `CrateDefMap` from `Name Resolution 2.0`. It should be lightning fast and also should improve autoimport correctness.

Kobzol:
Integrated https://github.com/intellij-rust/intellij-rust/pull/7516. Now trait completion works fine even with an empty path.

vlad20012:
> Now that we can create a more complex template builder, what do you think about letting the user change the contents of the generated trait items using the template builder? I think that we decided sometime ago (in the Implement trait quick fix) that it's not a good idea for functions, because these are typically multiline and would be annoying to fill with the template builder, but maybe for constants and types? Or maybe just move caret to the location of the first generated item once the user fully fills in the trait name and the template is finished?

Hmm, maybe, as we do it in `implement members` quick fix (there is a template for types at least), but this complicates the workflow if we're going to implement supertraits

vlad20012 (approved):
This looks pretty cool now! Let's finally merge it.

bors r+

vlad20012:
bors retry

Kobzol:
bors retry

PR #7481, opened 2021-07-07
COMP: add visibility completion provider

Kobzol:
This PR improves the completion of visibility items. The changes made are:

1) Previously, completion only offered `pub`. Now it also offers `pub(crate)` and `pub()` (restricted visibility). With `pub()`, the caret is placed inside `()` and completion popup is invoked, so that the user can quickly choose the restricted visibility (thanks to https://github.com/intellij-rust/intellij-rust/pull/7457). In theory, we could only offer `pub` and `pub()`, since `pub(crate)` can be quickly completed using `pub()` + completion popup. However, I think that `pub(crate)` is common enough that it warrants a separate completion item.
2) Visibility completion is now also offered for tuple (positional) and block (named) fields.
3) Items that were completed before after `pub` (such as `fn`), are now also offered after `pub(crate)`, `pub(in ...)`, etc. However, the implementation is quite nasty and far from correct. It is therefore included in a separate commit. I think that a much better solution would be to improve `RsVis` recovery, so that we can just check whether there is a valid visibility modifier before the current completion point.

    I managed to get `RsVisRestriction` recovery working somehow, but for general `RsVisAndAttrs` before items it seems to be more complex. It then starts to generate `RsToBeUpped` nodes and I'm not sure what to do with that, it seems to be a big change for the parser (?).

    Just for clarity, I want for this:
    ```rust
    pub
    ```
    or this:
    ```rust
    pub(crate)
    ```
    to be parsed as `RsVis` (even though there is no item following it yet).

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5025

changelog: Add completion for `pub(crate)` and `pub()` visibility modifers.


vlad20012:
> TODO(parser recovery?): it would be really nice to just say something like element.prevSibling is RsVis

I tried to do some magic with GK, but unfortunately the magic doesn't work today

image

Kobzol:
:( Well, we can also just remove this from the PR and just merge the first commit.

src/main/kotlin/org/rust/lang/core/completion/RsVisibilityCompletionContributor.kt
```diff
@@ -0,0 +1,15 @@
+package org.rust.lang.core.completion;
+
+import com.intellij.codeInsight.completion.CompletionContributor;
+import com.intellij.codeInsight.completion.CompletionType
+import org.rust.lang.core.RsPsiPattern
+import org.rust.lang.core.RsPsiPattern.declarationPattern
+import org.rust.lang.core.RsPsiPattern.inherentImplDeclarationPattern
+import org.rust.lang.core.or
+
+class RsVisibilityCompletionContributor: CompletionContributor() {
```
vlad20012, line 10: Let's not introduce new `CompletionContributor` and use `RsCompletionContributor`. (I'd rather remove `RsKeywordCompletionContributor`)

src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -229,24 +226,36 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
...
+                TokenType.WHITE_SPACE,
+                TokenType.ERROR_ELEMENT
+            ),
+            RS_COMMENTS
+        )
+        val siblings = generateSequence(item) { it.prevSibling }.takeWhile {
+            it is RsPath || it.elementType in allowedTokens
+        }.filter {
+            it !is PsiWhiteSpace && it !is PsiComment
+        }.toList().reversed()
```
vlad20012, line 256:
```suggestion
        }
```

src/main/kotlin/org/rust/lang/core/completion/RsKeywordCompletionContributor.kt
```diff
@@ -229,24 +226,36 @@ class RsKeywordCompletionContributor : CompletionContributor(), DumbAware {
...
 
-    private fun pubInherentImplDeclarationPattern(): PsiElementPattern.Capture<PsiElement> {
-        return baseInherentImplDeclarationPattern().and(statementBeginningPattern("pub"))
+        siblings.firstOrNull()?.elementType == PUB
```
vlad20012, line 258:
```suggestion
        siblings.lastOrNull()?.elementType == PUB
```

vlad20012 (approved):
> Well, we can also just remove this from the PR and just merge the first commit.

I'd not say the hack is so bad that we should remove it... Let's use it since it works. I'll try to improve parser recovery later

Kobzol:
bors r=vlad20012

Issue #3830, opened 2019-05-11
Lambda syntax completion


sirgl:
It would be great, if closure syntax (with pipes) can be completed.
E. g. for:
```rust
pub fn map<O, F: FnOnce(T) -> O>(self, mapper: F) -> Spanned<O> {
   ...
}

fn use(s: Spanned<i32>) {
   s.map(<caret>)
}
```
Here I expect completion item `|i| <caret>`.

Name suggestion already implemented for introduce variable refactoring.
Also, after completion we can suggest change parameter's name.

PR #7482, opened 2021-07-07
COMP: fix RsCompletionTestFixtureBase::checkCompletion

Kobzol:
While working on https://github.com/intellij-rust/intellij-rust/issues/3830, I noticed that the function `RsCompletionTestBase::checkCompletion` does not behave correctly. If no testmark is passed to the function, it basically won't do anything, therefore the test will always succeed.

I changed `checkCompletion` so that it actually does something even without a testmark. It now also throws an error if the passed lookup string is not found.

There were a few tests that used `checkCompletion` without a testmark, so they passed even though they were actually failing. I tried to fix them.


Kobzol:
There were some errors in tests that were hidden in the meantime and were exposed by rebasing on master.

Kobzol:
bors retry

PR #7483, opened 2021-07-07
T: minor provisional changes to RsStructureViewModel tests

afetisov:
Renamed RsStructureViewTest.kt to RsStructureViewModelTest.kt, changed test formatting for improved readability, minor test class refactoring.
This change is intended in preparation for further additions of Filters, Sorters and Groupers to the RsStructureViewModel.

PR #7060, opened 2021-04-06
INT: Generate documentation stub

0x7CA:
changelog: Added an editor intention that generates a documentation stub as demonstrated in: https://doc.rust-lang.org/beta/rust-by-example/meta/doc.html and https://github.com/rust-lang/rust/issues/57525

oKw9t3X

I'm not sure if the current implementation is robust enough, as the RustCommenter lacks the functionality that was used for IntelliJ Javadocs and Pycharm docstring implementations. So this implementation has to manually create the comments by inserting strings.  I was also not able to parse the function context like in intellij to see whether a function was being declared or just being referenced, or already had a doc, so I had to implement it. 
Feedback is very welcome

PR #7486, opened 2021-07-08
LI: fixed intentions and fixes running within doctests

### LINKED ISSUE #6790: Editing in code blocks in doc comments completely broken
## Environment

* **IntelliJ Rust plugin version:** 0.3.140.3644-203
* **Rust toolchain version:** 1.49.0 (e1884a8e3 2020-12-29) x86_64-unknown-linux-gnu
* **IDE name and version:** CLion 2020.3.2 (CL-203.7148.70)
* **Operating system:** Linux 5.9.13-arch1-1

## Problem description

Start here:

```rust
/// ```
/// # use repr_c_impl::layout::{Array, Type};
/// let x = Array {
/// };
/// ```
#[derive(Clone, Debug, Eq, PartialEq)]
pub struct Array<I: Layout> {
    pub element_type: Box<Type<I>>,
    pub num_elements: Option<u64>,
}
```

Execute the "Add missing fields action". Result:

```rust
/// ```
/// # use repr_c_impl::layout::{Array, Type};
/// let x = Array {
/// element_type: ,num_elements: };
/// ```
#[derive(CloBox::new(Type { })ne, Debug, Eq, NonePartialEq)]
pub struct Array<I: Layout> {
    pub element_type: Box<Type<I>>,
    pub num_elements: Option<u64>,
}
```

---

Start here with the cursor at the marked position:

```rust
/// ```
/// # use repr_c_impl::layout::{Array, Type};
/// let x = Array {
/// <cursor>
/// };
/// ```
```

Press tab. Result:

```rust
/// ```
/// struct Array {
    <cursor> /// 
/// }
/// ```
```

afetisov:
Closes #6790

Previously modification of injected doctest code was incorrectly handled, so that applying an fix or intention could result in corrupted code following the doctest. The reason was that modifications within a doctest run in their own injected editor, which wasn't correctly projected back into the containing file editor, but the offsets of changes were calculated relative to the whole file.

changelog: fixed intentions and type fixes applied with documentation tests

### DIFF
--- a/src/main/kotlin/org/rust/openapiext/Template.kt
+++ b/src/main/kotlin/org/rust/openapiext/Template.kt
@@ -10,6 +10,7 @@ import com.intellij.codeInsight.template.TemplateBuilderFactory
 import com.intellij.openapi.editor.Editor
 import com.intellij.psi.PsiElement
 import com.intellij.psi.SmartPsiElementPointer
+import com.intellij.psi.impl.source.tree.injected.InjectedLanguageEditorUtil
 
 fun <T : PsiElement, E : PsiElement> Editor.buildAndRunTemplate(
     owner: T,
@@ -22,5 +23,5 @@ fun <T : PsiElement, E : PsiElement> Editor.buildAndRunTemplate(
         val element = elementPointer.element ?: continue
         templateBuilder.replaceElement(element, element.text)
     }
-    templateBuilder.run(this, true)
+    templateBuilder.run(InjectedLanguageEditorUtil.getTopLevelEditor(this), true)
 }
--- a/src/test/kotlin/org/rust/ide/annotator/fixes/DoctestFixTest.kt
+++ b/src/test/kotlin/org/rust/ide/annotator/fixes/DoctestFixTest.kt
@@ -1,0 +1,101 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.annotator.fixes
+
+import org.intellij.lang.annotations.Language
+import org.rust.ide.annotator.*
+
+class DoctestFixTest : RsAnnotationTestBase() {
+
+    fun `test in doctest fix const`() = doTest("Add type i32", """
+        <error descr="Missing type for `const` item">const CONST/*caret*/ = 1;</error>
+    """, """
+        const CONST:i32/*caret*/ = 1;
+    """)
+
+    fun `test in doctest add missing fields`() = doTest("Add missing fields", """
+        struct Foo {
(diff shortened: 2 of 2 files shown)


afetisov:
This warrants a test, but I couldn't decide how and where to put it. Ideally we would want to run all annotation/fix/intention tests both within ordinary code and within doc comments, but that would probably significantly slow the testing with little gain. After all, there is no reason for injected code to behave differently from the usual one in the general case, and the edge cases are unlikely to be found simply by duplicating tests. I have tried to create a RsDoctestFixTest with a few test cases within org.rust.ide.annotator.fixes, but I couldn't understand how to make it treat the doctests as an injection. The test just treats it as text.

afetisov:
Also please could someone take a look at my other PRs? #7483 #7474

vlad20012:
Take a look at `RsEnterInLineCommentHandlerTest.doDoctestTestByText`. The magic is in `fileName = "lib.rs"` since only libs can have doctests.

Undin:
Let's add tests before merge

afetisov:
@vlad20012 Unfortunately that approach doesn't work. The problem is that when I try to test the code fences, I can only specify the behaviour with respect to the whole test file, but the editor fix tries to run in the injected editor. Specifically, I get the following diff:

```rust
extern crate test_package; fn __main() {
<error descr="Missing type for `const` item">const CONST = 1;</error>
```

vs 

```rust
/// ```
/// <error descr="Missing type for `const` item">const CONST/*caret*/ = 1;</error>
/// ```
struct Foo {
    var: i32,
    bar: i64,
    zar: [bool; 780],
}
```

in the pre-run code highlighting check when trying the expected test method 

```kotlin
checkFixByFileTree(
        "Add type i32",
        """
            //- lib.rs
            /// ```
            /// <error descr="Missing type for `const` item">const CONST/*caret*/ = 1;</error>
            /// ```
            struct Foo {
                var: i32,
                bar: i64,
                zar: [bool; 780],
            }
        """, """
            /// ```
            /// const CONST: i32 = 1;
            /// ```
            struct Foo {
                var: i32,
                bar: i64,
                zar: [bool; 780],
            }
            """,
    )
```

I don't see how I can tell the test fixture to use the same editor on both sides.

Undin:
@afetisov I suppose `myFixture.setCaresAboutInjection(false)` before `checkFixByFileTree ` should solve described issue

vlad20012:
Could you please it to the master instead of merging the master into it? There are rebasing instructions [here](https://github.com/intellij-rust/intellij-rust/pull/7060#issuecomment-823148179)

vlad20012:
@afetisov thanks!

PR #7487, opened 2021-07-08
COMP: complete lambdas in function expression position

Kobzol:
This PR adds completion for lambds in expression positions that expect a function (`&dyn Fn`, `F: Fn`, `impl Fn`, `fn`). The check for `|` already being present is quite hacky, but I didn't know how else to do it, since the `|` in e.g. `foo(|)` is not parsed as an expression nor as an identifier.

closure-completion

Fixes: https://github.com/intellij-rust/intellij-rust/issues/3830

changelog: Add completion for lambdas in expression positions that expect a function.

### DIFF
--- a/src/main/kotlin/org/rust/lang/core/completion/RsLambdaExprCompletionProvider.kt
+++ b/src/main/kotlin/org/rust/lang/core/completion/RsLambdaExprCompletionProvider.kt
@@ -1,0 +1,98 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.lang.core.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.completion.InsertionContext
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.patterns.ElementPattern
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiErrorElement
+import com.intellij.psi.util.descendantsOfType
+import com.intellij.psi.util.parentOfType
+import com.intellij.util.ProcessingContext
+import org.rust.ide.utils.template.newTemplateBuilder
+import org.rust.lang.core.psi.*
+import org.rust.lang.core.psi.RsElementTypes.OR
+import org.rust.lang.core.psi.ext.elementType
+import org.rust.lang.core.psi.ext.getNextNonWhitespaceSibling
+import org.rust.lang.core.psi.ext.leftSiblings
+import org.rust.lang.core.psi.ext.startOffset
+import org.rust.lang.core.psiElement
+import org.rust.lang.core.types.expectedType
+import org.rust.lang.core.types.implLookup
+import org.rust.openapiext.createSmartPointer
+
+object RsLambdaExprCompletionProvider : RsCompletionProvider() {
+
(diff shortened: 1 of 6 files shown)


vlad20012:
Why not to suggest lambda completion even if there is no `|` typed?

src/main/kotlin/org/rust/lang/core/completion/RsLambdaExprCompletionProvider.kt
```diff
@@ -0,0 +1,112 @@
...
+
+        val paramCount = getParameterCount(expr) ?: return
+        val params = (0 until paramCount).joinToString(", ") { "p${it}" }
+
+        val start = if (expr.leftSiblings.filter { it !is PsiErrorElement }.firstOrNull()?.elementType == OR) {
+            ""
+        } else {
+            "|"
+        }
+        val text = "$start$params| {}"
```
vlad20012, line 56: I'd not say we should not complete `{}` here. A user can type it if needed

src/main/kotlin/org/rust/lang/core/completion/RsLambdaExprCompletionProvider.kt
```diff
@@ -0,0 +1,112 @@
...
+    return when (type) {
+        is TyFunction -> type.paramTypes.size
+        is TyAnon -> type.traits.mapNotNull { it.getParameterCount() }.firstOrNull()
+        is TyTraitObject -> type.traits.mapNotNull { it.getParameterCount() }.firstOrNull()
+        is TyTypeParameter -> {
+            val traits = ctx.implLookup.getEnvBoundTransitivelyFor(type)
+            traits.mapNotNull { it.getParameterCount() }.firstOrNull()
+        }
+        else -> null
+    }
```
vlad20012, line 105: `ImplLookup.asTyFunction` ?

vlad20012 (approved):

src/main/kotlin/org/rust/lang/core/completion/RsLambdaExprCompletionProvider.kt
Kobzol: I included the body in the template builder - maybe that's even better?
complete-lambda

vlad20012: :+1: but let's move the caret between `{}` if a user decides to leave them as is

Kobzol:
Sorry for the repeated push spam. After discovering `tyFunction` I could remove a third of the PR :D

Kobzol:
Added a listener that moves the caret inside the block, if it still exists after the template finishes.

vlad20012:
Well, it moves the caret even if I type something in these braces(


https://user-images.githubusercontent.com/3221931/125449907-649fc18b-923b-47d7-80ed-e81ec60e1e4a.mp4

Kobzol:
The intended workflow is either replace the whole expression or press enter and then start typing in the block, you're doing it wrong :D

So, let's put it into the braces only if the block is empty?

vlad20012:
> The intended workflow is either replace the whole expression or press enter and then start typing in the block

But what if I replace the whole expression with an expression in braces? :D
Let's replace it if it's empty, I guess

Kobzol:
> But what if I replace the whole expression with an expression in braces? :D

:open_mouth:

Kobzol:
I changed it so that the caret is moved only if there are no elements inside the block.

vlad20012:
:+1: 
bors r+

Issue #4763, opened 2019-12-17
Feature Request - Break on panic


GrumpyMetalGuy:
It would be extremely helpful to have the debugger break when a panic occurs. Seeing the stack trace and any relevant context at the point of panic would make debugging much easier.

ortem:
You can add the following command to your [.gdbinit/.lldbinit](https://www.jetbrains.com/help/clion/configuring-debugger-options.html#gdbinit-lldbinit):
* `breakpoint set -n rust_panic` for LLDB
* `break rust_panic` for GDB

The execution will stop inside `panic` call
<img width="1312" alt="Screenshot 2020-02-05 at 12 33 01" src="https://user-images.githubusercontent.com/4854600/73829291-c7afff00-4813-11ea-9942-2634a74b5577.png">

Then you can select the desired frame and to look at local variables and so on
<img width="1312" alt="Screenshot 2020-02-05 at 12 33 08" src="https://user-images.githubusercontent.com/4854600/73829309-cd0d4980-4813-11ea-853f-0b6b67e254cb.png">


Does this solution work in your case? If so, we'll consider adding such a setting inside IDE.

GrumpyMetalGuy:
Sorry for the delay in replying, just got around to trying this out now. Probably user error on my part, but not able to make it work. I'm on Windows, have installed MSys2 64 bit, setup the toolchain to point to it, and added a .gdbinit file in my C:\Users\<username> folder with the following contents:

set auto-load local-gdbinit on
add-auto-load-safe-path /
break rust_panic

I also tried adding a .gdbinit in the root of my project folder, adding the local settings as outlined at https://www.jetbrains.com/help/clion/configuring-debugger-options.html#gdbinit-lldbinit, but it still doesn't seem to stop.

I'm guessing I just don't have the files setup correctly, if anyone has an example .gdbinit I could shamelessly copy-pasta, that'd be awesome!

GrumpyMetalGuy:
OK, made some progress. I set a breakpoint early in my program, and when I hit it, switched to the GDB view. One of the first things I see in the output window is:

No symbol table is loaded.  Use the "file" command.
Make breakpoint pending on future shared library load? (y or [n]) [answered N; input not from terminal]
No source file named C:/Development/Projects/rust/AdventOfCode2019/src/bin/day6.rs.
No source file named C:/Development/Projects/rust/AdventOfCode2019/src/intcode.rs.
[...]

PR #7488, opened 2021-07-08
DBG: automatically break on panic inside debugger

Kobzol:
This PR adds the option to break the debugger on any rust panic.

I had to slightly modify the debugger configurable implementation. Previously, only the toolchain options were inside the general debugger settings. I could not add another configurable, because it would have a duplicated `Rust` header, so I had to add the `break on panic` setting to the existing configurable that is used inside the general settings.

To keep the individual settings separate, I split them into separate classes so that they are not all mixed up. I also created a new base class (`RsDebuggerUiComponent`), because I didn't know how to compose `JComponent`s of multiple objects inside a single `panel` in the `getComponent` method in `RsDebuggerGeneralSettingsConfigurableUi`.

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4763

changelog: Add debugger option to break on panic (enabled by default).

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/settings/RsBreakOnPanicConfigurableUi.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/settings/RsBreakOnPanicConfigurableUi.kt
@@ -1,0 +1,31 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger.settings
+
+import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.layout.LayoutBuilder
+
+class RsBreakOnPanicConfigurableUi : RsDebuggerUiComponent() {
+    private val breakOnPanicCheckBox: JBCheckBox
+        = JBCheckBox("Break on panic", RsDebuggerSettings.getInstance().breakOnPanic)
+
+    override fun reset(settings: RsDebuggerSettings) {
+        breakOnPanicCheckBox.isSelected = settings.breakOnPanic
+    }
+
+    override fun isModified(settings: RsDebuggerSettings): Boolean
+        = settings.breakOnPanic != breakOnPanicCheckBox.isSelected
+
+    override fun apply(settings: RsDebuggerSettings) {
+        settings.breakOnPanic = breakOnPanicCheckBox.isSelected
+    }
+
+    override fun buildUi(builder: LayoutBuilder) {
+        with(builder) {
+            row { breakOnPanicCheckBox() }
+        }
+    }
+}
--- a/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerGeneralSettingsConfigurableUi.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerGeneralSettingsConfigurableUi.kt
@@ -1,0 +1,56 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger.settings
+
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.options.ConfigurableUi
+import com.intellij.ui.layout.panel
+import com.intellij.util.PlatformUtils
+import org.rust.debugger.RsDebuggerToolchainService
+import org.rust.debugger.RsDebuggerToolchainService.LLDBStatus
+import javax.swing.JComponent
+
+class RsDebuggerGeneralSettingsConfigurableUi : ConfigurableUi<RsDebuggerSettings>, Disposable {
+    private val needsToolchainSettings: Boolean
+        get() {
+            // CLion has own Toolchain settings
+            if (PlatformUtils.isCLion()) return false
+            val status = RsDebuggerToolchainService.getInstance().getBundledLLDBStatus()
+            // If there is bundled LLDB, no need to show this toolchain settings
+            return status !is LLDBStatus.Binaries
+        }
+
+    private val components: List<RsDebuggerUiComponent> = run {
+        val components = mutableListOf<RsDebuggerUiComponent>()
+        if (needsToolchainSettings) {
+            components.add(RsDebuggerToolchainConfigurableUi())
+        }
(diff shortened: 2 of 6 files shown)


debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettings.kt
```diff
@@ -48,38 +48,28 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
...
-            val status = RsDebuggerToolchainService.getInstance().getBundledLLDBStatus()
-            // If there is bundled LLDB, no need to show this toolchain settings
-            return status !is RsDebuggerToolchainService.LLDBStatus.Binaries
-        }
-
     companion object {
         @JvmStatic
         fun getInstance(): RsDebuggerSettings = getInstance(RsDebuggerSettings::class.java)
 
-        const val TOOLCHAIN_ID: String = "Debugger.Rust.Toolchain"
```
Kobzol, line 82: Maybe this ID needs to be preserved?

debugger/src/main/kotlin/org/rust/debugger/settings/RsBreakOnPanicConfigurableUi.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.intellij.ui.layout.LayoutBuilder
+
+class RsBreakOnPanicConfigurableUi : RsDebuggerUiComponent() {
+    private val breakCheckbox = JBCheckBox("Break on panic", RsDebuggerSettings.getInstance().breakOnPanic)
+
+    override fun reset(settings: RsDebuggerSettings) {
+        breakCheckbox.isSelected = settings.breakOnPanic
+    }
+
+    override fun isModified(settings: RsDebuggerSettings): Boolean = settings.breakOnPanic != breakCheckbox.isSelected
```
ortem, line 18: [minor]: expression body can be placed on the new line for readability

debugger/src/main/kotlin/org/rust/debugger/settings/RsBreakOnPanicConfigurableUi.kt
```diff
@@ -0,0 +1,29 @@
...
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger.settings
+
+import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.layout.LayoutBuilder
+
+class RsBreakOnPanicConfigurableUi : RsDebuggerUiComponent() {
+    private val breakCheckbox = JBCheckBox("Break on panic", RsDebuggerSettings.getInstance().breakOnPanic)
```
ortem, line 12: [minor]
```suggestion
    private val breakOnPanicCheckBox: JBCheckBox =
        JBCheckBox("Break on panic", RsDebuggerSettings.getInstance().breakOnPanic)
```

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerUiComponent.kt
```diff
@@ -0,0 +1,23 @@
...
+
+package org.rust.debugger.settings
+
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.options.ConfigurableUi
+import com.intellij.ui.layout.LayoutBuilder
+import com.intellij.ui.layout.panel
+import javax.swing.JComponent
+
+abstract class RsDebuggerUiComponent: Disposable, ConfigurableUi<RsDebuggerSettings> {
```
ortem, line 14: [minor]
```suggestion
abstract class RsDebuggerUiComponent : ConfigurableUi<RsDebuggerSettings>, Disposable {
```

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerUiComponent.kt
```diff
@@ -0,0 +1,23 @@
...
+    abstract fun buildUi(builder: LayoutBuilder)
+
+    override fun getComponent(): JComponent {
+        return panel {
+            buildUi(this)
+        }
+    }
+    override fun dispose() {}
```
ortem, line 22: [minor]
```suggestion

    override fun dispose() {}
```

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerGeneralSettingsConfigurableUi.kt
```diff
@@ -0,0 +1,55 @@
...
+            if (PlatformUtils.isCLion()) return false
+            val status = RsDebuggerToolchainService.getInstance().getBundledLLDBStatus()
+            // If there is bundled LLDB, no need to show this toolchain settings
+            return status !is LLDBStatus.Binaries
+        }
+
+    private val components: List<RsDebuggerUiComponent> = run {
+        val components = mutableListOf<RsDebuggerUiComponent>(RsBreakOnPanicConfigurableUi())
+        if (needsToolchainSettings) {
+            components.add(RsDebuggerToolchainConfigurableUi())
```
ortem, line 29: Maybe add `RsDebuggerToolchainConfigurableUi` before `RsBreakOnPanicConfigurableUi` as more important?

ortem (commented):
Thank you @Kobzol!

Generally LGTM except for minor comments. Tested on macOS with LLDB (CLion, IJ IDEA Ultimate), on Windows with MSVC LLDB and MinGW GDB.

@Undin Please check the UI changes

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerGeneralSettingsConfigurableUi.kt
Kobzol: Good point :)

ortem (approved):
Important note: this feature should probably be reimplemented as [exception breakpoints](https://www.jetbrains.com/help/clion/using-breakpoints.html#exception-breakpoints) someday (I'll open a new issue for that). But this might require some changes from the CLion side. Anyway, this PR may be considered as an initial implementation. Thank you @Kobzol!

bors r+

Issue #4764, opened 2019-12-17
Feature Request - option to not step into core Rust code while debugging


GrumpyMetalGuy:
It would be super helpful to have a facility for the Rust plugin similar to the one for JS and Python under Build, Execution, Deployment -> Debugger -> Stepping to not step into the Rust core code when debugging. I don't really want to step into collections code or similar, and it would be nice to have the plugin do this for me.

Bonus points if there's also a feature to list crates that can be skipped over too :)

Scordo:
I would really appreciate it, too.

GrumpyMetalGuy:
Looks like, similar to #4763 , you can add something like:

skip -rfu ^(std)|(core)|(alloc)::.*

to .gdbinit to skip over these. Clearly not exhaustive, but a step in the right direction...

PR #7489, opened 2021-07-08
DBG: add option to not step into stdlib sources

Kobzol:
This PR adds an option to not step into stdlib sources. In the future, it could be extended with a table for adding custom paths to be ignored (like what e.g. Python offers).

Currently, it doesn't work inside LLDB, because it does not use full paths for functions (e.g. `Vec::clear` is `example_project'clear<i32,alloc::alloc::Global>` instead of `alloc::vec::Vec::clear`). @ortem Do you have an idea how this could be fixed?

Fixes: https://github.com/intellij-rust/intellij-rust/issues/4764

https://user-images.githubusercontent.com/4854600/216394673-4292c396-1dd3-4531-b608-368cd592c8ad.mov

For QA: this feature is supposed to work
* with GDB on Linux and Windows
* with LLDB on macOS and Linux since CLion 2023.1 EAP 231.6471.1
* with MSVC LLDB on Windows

changelog: Add debugger option to not step into stdlib sources.

### DIFF
--- a/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettings.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettings.kt
@@ -24,6 +24,7 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
     var downloadAutomatically: Boolean = false
 
     var breakOnPanic: Boolean = true
+    var stepSettings: RsStepFilterSettings = RsStepFilterSettings(filterStdlib = false)
 
     var decorateMsvcTypeNames: Boolean = true
 
@@ -37,6 +38,7 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
         val configurable = when (category) {
             DebuggerSettingsCategory.DATA_VIEWS -> createDataViewConfigurable()
             DebuggerSettingsCategory.GENERAL -> createGeneralSettingsConfigurable()
+            DebuggerSettingsCategory.STEPPING -> createSteppingConfigurable()
             else -> null
         }
         return listOfNotNull(configurable)
@@ -60,10 +62,19 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
         )
     }
 
+    private fun createSteppingConfigurable(): Configurable {
+        return SimpleConfigurable.create(
+            STEPPING_ID,
+            RsDebuggerBundle.message("settings.rust.debugger.title"),
+            RsDebuggerSteppingSettingsConfigurableUi::class.java,
+            Companion::getInstance
+        )
+    }
+
     override fun isTargetedToProduct(configurable: Configurable): Boolean {
         if (configurable !is SearchableConfigurable) return false
         return when (configurable.id) {
-            GENERAL_SETTINGS_ID, DATA_VIEW_ID -> true
+            GENERAL_SETTINGS_ID, DATA_VIEW_ID, STEPPING_ID -> true
             else -> false
         }
     }
@@ -74,5 +85,6 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
 
         const val GENERAL_SETTINGS_ID: String = "Debugger.Rust.General"
         const val DATA_VIEW_ID: String = "Debugger.Rust.DataView"
+        const val STEPPING_ID: String = "Debugger.Rust.Stepping"
     }
 }
--- a/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSteppingSettingsConfigurableUi.kt
+++ b/debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSteppingSettingsConfigurableUi.kt
@@ -1,0 +1,37 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger.settings
+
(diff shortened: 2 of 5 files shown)


debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettings.kt
```diff
@@ -60,16 +62,26 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
...
-            GENERAL_SETTINGS_ID, DATA_VIEW_ID -> true
+            GENERAL_SETTINGS_ID, DATA_VIEW_ID, STEPPING_ID -> true
             else -> false
         }
     }
 
+    private fun createSteppingConfigurable(): Configurable {
+        return SimpleConfigurable.create(
+            STEPPING_ID,
+            "Rust",
```
ortem, line 73:
```suggestion
            RsDebuggerBundle.message("settings.rust.debugger.title"),
```

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettings.kt
```diff
@@ -60,16 +62,26 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
     override fun isTargetedToProduct(configurable: Configurable): Boolean {
         if (configurable !is SearchableConfigurable) return false
         return when (configurable.id) {
-            GENERAL_SETTINGS_ID, DATA_VIEW_ID -> true
+            GENERAL_SETTINGS_ID, DATA_VIEW_ID, STEPPING_ID -> true
             else -> false
         }
     }
 
+    private fun createSteppingConfigurable(): Configurable {
```
ortem, line 70: Let's move the method up, just below `createGeneralSettingsConfigurable`

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSteppingSettingsConfigurableUi.kt
```diff
@@ -0,0 +1,36 @@
...
+package org.rust.debugger.settings
+
+import com.intellij.openapi.options.ConfigurableUi
+import com.intellij.ui.components.JBCheckBox
+import com.intellij.ui.layout.panel
+import javax.swing.JComponent
+
+class RsDebuggerSteppingSettingsConfigurableUi : ConfigurableUi<RsDebuggerSettings> {
+    private val filterStdlib = JBCheckBox(
+        "Do not step into stdlib (std, core, alloc)",
```
ortem, line 15: Let's move to to bundle like this
```suggestion
RsDebuggerBundle.message("settings.rust.debugger.do.not.step.into.stdlib.checkbox"),
```

debugger/src/main/kotlin/org/rust/debugger/settings/RsDebuggerSettings.kt
```diff
@@ -23,6 +23,7 @@ class RsDebuggerSettings : XDebuggerSettings<RsDebuggerSettings>("Rust") {
     var downloadAutomatically: Boolean = false
 
     var breakOnPanic: Boolean = true
+    var stepSettings: RsStepFilterSettings = RsStepFilterSettings(filterStdlib = true)
```
ortem, line 26: I think we should leave the default behaviour as is
```suggestion
    var stepSettings: RsStepFilterSettings = RsStepFilterSettings(filterStdlib = false)
```

ortem (changes requested):
Hi @Kobzol, thank you for this PR! I'm sorry that it took so long.
I have a great news to share: since #10065 is fixed in the latest CLion EAP, your implementation now actually works with LLDB!
 
The implementation looks good to me in general, I have just few comments here

Kobzol:
Rebased on master and resolved review comments.

ortem (approved):
LGTM
bors r+

neonaot:
Hi!
I've found one problem - https://github.com/intellij-rust/intellij-rust/issues/10088 (on both 2023.1 and 2022.3 platforms). 

And tested the following pairs, didn't find any issues:
GDB on Linux - OK
GDB on Windows  - OK
LLDB on macOS since CLion 2023.1 EAP 231.6471.1 - OK
LLDB on Linux since CLion 2023.1 EAP 231.6471.1 - OK

cc @ortem

ortem:
@Kobzol Just to clarify, the found problem on Windows is most likely caused by the bundled LLDB itself, and not by the implementation on the Rust plugin side. I am going to look at it later.

ortem:
@Kobzol However #10089 is probably caused by the Rust plugin, so you're welcome to help us investigate it if you like 😄

debugger/src/main/kotlin/org/rust/debugger/settings/RsStepFilterSettings.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.debugger.settings
+
+data class RsStepFilterSettings(
```
ortem, line 8: @Kobzol I guess that `XmlSerializerUtil` does not properly serialize a Kotlin data class out-of-the-box. So it's probably worth just using a primitive `var skipStdlib: Boolean = false` field in the `RsDebuggerSettings` instead of boxing it into `data class RsStepFilterSettings`. Just the same way as we do with `var breakOnPanic: Boolean = true`

KseniaStar:
I documented the change: https://plugins.jetbrains.com/plugin/8182-rust/docs/rust-stepping-through.html

PR #7492, opened 2021-07-09
INSP: add quick fix to remove an unused import

Kobzol:
This PR adds a quick fix to remove an unused import. To make this fix more usable, I first changed the warning annotation range of `RsUnusedImportInspection`. This slightly deviates form the text displayed by `rustc`, but I think that it's more obvious for the user to see what is actually unused. And it will also allow the user to remove multiple unused use specks with a single quick fix.

The annotation is now checked only once per each use item, so some of the checks will not be performed per each use speck. This new annotation works in two steps. First I collect the usage (used/unused) of each use speck, and then I traverse them from the top and try to mark as large ranges as possible (i.e. if a group has two unused children, the group itself will also be marked as unused).

Before:
before

After:
after

In the quickfix it made sense to me to normalize the resulting use speck and flatten it if a group was reduced to a single element after removing an unused child of the group. To avoid duplicating code here, I reused `RemoveCurlyBracesIntention`. But before I had to update it a bit, since it's quite old and was unnecessarily strict. Now it's more general and can be also reused from other parts of the plugin.

The changes were relevant to each other and it's not too much code, but if you want, I can split this into multiple PRs.

changelog: Add a quick fix to remove an unused import.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
@@ -18,31 +18,72 @@ class RsUnusedImportInspection : RsLintInspection() {
     override fun getLint(element: PsiElement): RsLint = RsLint.UnusedImports
 
     override fun buildVisitor(holder: RsProblemsHolder, isOnTheFly: Boolean) = object : RsVisitor() {
-        override fun visitUseSpeck(o: RsUseSpeck) {
+        override fun visitUseItem(item: RsUseItem) {
             if (!holder.project.isNewResolveEnabled) return
-            val item = o.ancestorStrict<RsUseItem>() ?: return
             if (item.isReexport) return
-            if (o.path?.resolveStatus != PathResolveStatus.RESOLVED) return
 
             // Do not check uses if there is a child mod inside the current mod
-            val parentMod = o.containingMod
+            val parentMod = item.containingMod
             val hasChildMods = parentMod.expandedItemsExceptImplsAndUses.any {
                 it is RsModItem || it is RsModDeclItem
             }
-            if (o.parentOfType<RsFunction>() == null && hasChildMods) return
+            if (item.parentOfType<RsFunction>() == null && hasChildMods) return
 
-            checkUseSpeck(o, holder)
+            val owner = item.parent as? RsItemsOwner ?: return
+            val usage = owner.pathUsage
+
+            val speck = item.useSpeck ?: return
+
+            val unusedSpeckSet = HashSet<RsUseSpeck>()
+            storeUseSpeckUsage(speck, usage, unusedSpeckSet)
+            markUnusedSpecks(speck, unusedSpeckSet, holder)
         }
     }
 
-    private fun checkUseSpeck(useSpeck: RsUseSpeck, holder: RsProblemsHolder) {
-        if (!useSpeck.isUsed()) {
-            val element = getHighlightElement(useSpeck)
-            holder.registerLintProblem(
-                element,
-                "Unused import: `${useSpeck.text}`"
-            )
+    /**
+     * Traverse all use specks recursively and store information about their usage.
+     */
+    private fun storeUseSpeckUsage(
+        useSpeck: RsUseSpeck,
+        usage: PathUsageMap,
+        unusedSpeckSet: MutableSet<RsUseSpeck>
+    ) {
+        val group = useSpeck.useGroup
+        val isUsed = if (group == null) {
+            isUseSpeckUsed(useSpeck, usage)
+        } else {
+            group.useSpeckList.forEach {
+                storeUseSpeckUsage(it, usage, unusedSpeckSet)
+            }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
```diff
@@ -18,32 +19,79 @@ class RsUnusedImportInspection : RsLintInspection() {
...
-            if (o.parentOfType<RsFunction>() == null && hasChildMods) return
+            if (item.parentOfType<RsFunction>() == null && hasChildMods) return
 
-            checkUseSpeck(o, holder)
+            val owner = item.parent as? RsItemsOwner ?: return
+            val usage = owner.pathUsage
+
+            val speck = item.useSpeck ?: return
+
+            val isSpeckUsedMap = HashMap<RsUseSpeck, Boolean>()
```
dima74, line 38: Is it intentionally `HashMap` and not `HashSet`?

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
```diff
@@ -18,32 +19,79 @@ class RsUnusedImportInspection : RsLintInspection() {
...
+    ) {
+        val used = isSpeckUsedMap[useSpeck] ?: return
+
+        val group = useSpeck.useGroup
+        if (group == null) {
+            if (!used) {
+                markAsUnused(useSpeck, holder)
+            }
+        } else {
+            if (group.useSpeckList.none { isSpeckUsedMap[it] == true }) {
```
dima74, line 57:
```suggestion
            if (isSpeckUsedMap[useSpeck] == false) {
```

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
```diff
@@ -18,32 +19,79 @@ class RsUnusedImportInspection : RsLintInspection() {
...
 
-            checkUseSpeck(o, holder)
+            val owner = item.parent as? RsItemsOwner ?: return
+            val usage = owner.pathUsage
+
+            val speck = item.useSpeck ?: return
+
+            val isSpeckUsedMap = HashMap<RsUseSpeck, Boolean>()
+            storeUseSpeckUsage(speck, usage, isSpeckUsedMap)
+            markUnusedSpecks(speck, isSpeckUsedMap, holder)
```
dima74, line 40: It's a bit inconsistent that `storeUseSpeckUsage` is outside class and `markUnusedSpecks` is inside (since these methods are highly related)

src/main/kotlin/org/rust/ide/intentions/RemoveCurlyBracesIntention.kt
```diff
@@ -9,74 +9,82 @@ import com.intellij.openapi.editor.Editor
...
  *
  * to this:
  *
  * ```
- * import std::mem;
+ * use std::mem;
  * ```
  */
-// TODO: this really should reuse code from RsSingleImportRemoveBracesFormatProcessor.
 class RemoveCurlyBracesIntention : RsElementBaseIntentionAction<RemoveCurlyBracesIntention.Context>() {
```
dima74, line 27: Lets also support this case:
```rust
struct A;
mod inner {
    pub struct B;
}
mod foo {
    use crate::{A, inner::B};
    fn foo(b: B) {}
}
```

Expected after invoking "Remove unused import": `use crate::inner::B;`
Actual: `use crate::{inner::B};`

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
Kobzol: It was more of a sanity check, to distinguish used/unused/use speck was not found. But if the implementation of `storeUseSpeckUsage` is correct and finds all specks, the third situation should not happen.

src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
Kobzol: Right, `markUnusedSpecks` needed to be inside the class because of `registerLintProblem`. I'll move `storeUseSpeckUsage` inside the class too.

src/main/kotlin/org/rust/ide/intentions/RemoveCurlyBracesIntention.kt
Kobzol: Good catch, thanks.

Kobzol:
I rebased to get latest master, but I still kept the commits since I made some subtle changes in response to your review. Let me know if you want me to squash it (or we can just do it before final merge).

dima74 (approved):
Changes looks good, lets squash the commit (btw separate review commits are very convenient for review, thanks), and we can merge the PR

Kobzol:
bors r=dima74

PR #7494, opened 2021-07-09
T&PERF: improve tests performance

vlad20012:
1. Cache `RustcInfo` and `StandardLibrary` instances in rust project descriptors.
2. Don't change rust edition for stdlib dependencies by `@MockEdition` (this makes sense only for `WithActualStdlibRustProjectDescriptor`)
3. Use a different macro expansion caching key for `WithActualStdlibRustProjectDescriptor` tests. Now macros in stdlib dependencies will not be re-expanded every time.
4. Some optimization here and there

Points 1-3 make `WithActualStdlibRustProjectDescriptor` tests almost as cheap as `WithStdlibRustProjectDescriptor`.

### DIFF
--- a/src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
+++ b/src/main/kotlin/org/rust/cargo/project/model/CargoProjectService.kt
@@ -51,6 +51,7 @@ interface CargoProjectsService {
      * @param manifest a path to `Cargo.toml` file of the project that should be attached
      */
     fun attachCargoProject(manifest: Path): Boolean
+    fun attachCargoProjects(vararg manifests: Path)
     fun detachCargoProject(cargoProject: CargoProject)
     fun refreshAllProjects(): CompletableFuture<out List<CargoProject>>
     fun discoverAndRefresh(): CompletableFuture<out List<CargoProject>>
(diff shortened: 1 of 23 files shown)


Kobzol:
I was recently looking at how could the tests be sped up (actually https://github.com/intellij-rust/intellij-rust/pull/7437 started with that in mind). I looked at the Kotlin plugin and it seems to me that it can actually leverage multiple test runners in parallel (https://github.com/JetBrains/kotlin/blob/48a4d7d4186d0642d8e69bb274e18268af69fcd6/buildSrc/src/main/kotlin/tasks.kt#L229). I wonder if this could also be done for the Rust plugin? Just setting `maxParallelForks` doesn't work, since immediately some files (indices etc.) get corrupted.

vlad20012:
Most likely, you still need to somehow change (`idea.system.path` system variable?), but to be honest I don't know anything special about parallel testing.

Kobzol:
It's not directly related to test performance *per se*, but I noticed that often PRs are sent to bors immediately after a final push. This causes the check workflow to basically run 2x, once in PR and once in bors, which can increase the latency of the final merge (especially if there are more workflow running in parallel), if bors gets into the queue. I'm not sure if it's worth it, but maybe `bors r+` could cancel any running workflows on the PR itself?

vlad20012:
I think this is a question for @Undin.

Sometimes I cancel a PR workflow manually to speed up bors work (btw do you have permission to do this?)

Kobzol:
Oh, I do :smile: Ok, I guess doing it manually from time to time is enough :)

vlad20012:
I've removed the most controversial commit about the unified project descriptor (I think I'll push it later with more improvements).
Unfortunately, these changes do not affect CI execution time too much :(

bors r+

vlad20012:
bors retry

Issue #5774, opened 2020-07-21
Optimize imports intention


fctorial:
There is no optimize imports intention available for rust files. Also, unused imports aren't grayed out in rust files.

Undin:
@fctorial are you about the addition of the corresponding intention (i.e. action available on <kbd>Alt + Enter</kbd>) or `Code | Optimize Import` action? Note, `Optimize Import` action is available for Rust but it doesn't support removing unused imports (since it's not implemented yet), only imports sorting

fctorial:
Yes, that is it.

PR #7495, opened 2021-07-09
IDE: remove unused imports in import optimizer

Kobzol:
This PR modifies import optimizer so that it now removes unused imports.

optimize-imports

First I thought that we might need to introduce an experimental feature for this, but I think that we can simply remove the unused imports that we find with the current implementation and in the future, once the implementation improves, just relax the checks performed by the import optimizer (these correspond to the same checks that are in `RsUnusedImportInspection`, these should be probably merged, it should be easier after https://github.com/intellij-rust/intellij-rust/pull/7492 gets merged).

Fixes: https://github.com/intellij-rust/intellij-rust/issues/5774
Fixes: https://github.com/intellij-rust/intellij-rust/issues/2158

Related issue: https://github.com/intellij-rust/intellij-rust/issues/2259

changelog: The import optimizer will now remove unused imports.

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsUnusedImportInspection.kt
@@ -22,7 +22,6 @@ class RsUnusedImportInspection : RsLintInspection() {
             if (!holder.project.isNewResolveEnabled) return
             val item = o.ancestorStrict<RsUseItem>() ?: return
             if (item.isReexport) return
-            if (o.path?.resolveStatus != PathResolveStatus.RESOLVED) return
 
             // Do not check uses if there is a child mod inside the current mod
             val parentMod = o.containingMod
@@ -58,7 +57,7 @@ private fun getHighlightElement(useSpeck: RsUseSpeck): PsiElement {
  * A usage can be either a path that uses the import of the use speck or a method call/associated item available through
  * a trait that is imported by this use speck.
  */
-private fun RsUseSpeck.isUsed(): Boolean {
+fun RsUseSpeck.isUsed(): Boolean {
     val owner = this.parentOfType<RsUseItem>()?.parent as? RsItemsOwner ?: return true
     val usage = owner.pathUsage
     return isUseSpeckUsed(this, usage)
@@ -68,6 +67,8 @@ private fun isUseSpeckUsed(useSpeck: RsUseSpeck, usage: PathUsageMap): Boolean {
     // Use speck with an empty group is always unused
     val useGroup = useSpeck.useGroup
     if (useGroup != null && useGroup.useSpeckList.isEmpty()) return false
+
+    if (useSpeck.path?.resolveStatus != PathResolveStatus.RESOLVED) return true
 
     val items = if (useSpeck.isStarImport) {
         val module = useSpeck.path?.reference?.resolve() as? RsMod ?: return true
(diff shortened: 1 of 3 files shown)


Kobzol:
@dima74 Some of the tests fail because new resolve is not enabled for them in the CI runs where old resolve is used. Should we just mark the corresponding classes with `UseNewResolve`? Or create separate versions of these tests?

src/test/kotlin/org/rust/ide/refactoring/move/RsMoveTopLevelItemsTest.kt
```diff
@@ -924,6 +924,10 @@ class RsMoveTopLevelItemsTest : RsMoveTopLevelItemsTestBase() {
                 bar::bar_func();
                 let _ = BarStruct {};
             }
+            fn foo2() {
```
dima74, line 927: Lets instead remove imports in `after` code (and mark test with `@UseNewResolve`)

dima74 (commented):
Let's mark necessary tests (functions, not the whole classes) with `@UseNewResolve`. There are a few such tests, so I think it's OK if they will be tested only with new resolve (anyway we plan to remove option to enable old resolve in future)

dima74:
Thanks! 👍
bors r+

Undin:
@dima74 @Kobzol in #6392 we decided to disable the corresponding inspection because it's still incomplete and has false positives. But at the same time, we want to enable this behavior by default in import optimizer. It's quite inconsistent and, as I understand, still may lead to an undesired result. So I suggest disabling this feature by default for now and enable when we will be sure that unused import search is ready for users

dima74:
@Undin yes, this indeed is inconsistent
@Kobzol let's then introduce experimental feature for this

Undin:
@Kobzol I've also removed `Fixes` from PR description to close the corresponding issue when we will enable this feature by default

dima74:
> Maybe we can just disable this behavior in the optimizer if the inspection is currently disabled, instead of introducing an experimental feature?

Good idea, let's do it that way:
```kotlin
val profile = InspectionProjectProfileManager.getInstance(project).currentProfile
profile.isToolEnabled(HighlightDisplayKey.find(RsUnusedImportInspection.SHORT_NAME))
```

> Another question, in the future we would probably want to enable the experimental feature for nightly users. Can we also set the inspection to be enabled by default for them?

I think we can directly enable the future for all users when false-positives will be addressed (since the future scope is relatively small)

Kobzol:
What should be the value of `SHORT_NAME`?

Undin:
Just to clarify: I don't see the problem in inconsistency with availability of inspection and unused import deletion. If both features are disabled by default, almost nobody won't be affected with such inconsistency.
I see the problem in turning on incomplete feature for a users by default when we decided not to do it with related feature.
So I this it's ok just to introduce a separate experimental feature just for import optimizer only

Undin:
> Another question, in the future we would probably want to enable the experimental feature for nightly users. Can we also set the inspection to be enabled by default for them?

@Kobzol yep, it's possible. We can do it in the same way as we enable experimental features in nightly builds - via separate plugin manifest (see https://github.com/intellij-rust/intellij-rust/tree/master/src/main/resources-nightly/META-INF)

dima74:
I personally prefer to enable future in import optimizer if inspection is enabled, but of course introducing experimental future is also OK

> What should be the value of `SHORT_NAME`?

Yes, something like `"RsUnusedImportInspection"`. In this case `getShortName` also should be overridden like in [`RsExternalLinterInspection`](https://github.com/intellij-rust/intellij-rust/blob/9c33738b7e4436f3397ede83c23e885c5f2eab2f/src/main/kotlin/org/rust/ide/inspections/RsExternalLinterInspection.kt#L114)

Kobzol:
Thanks, I added the check and corresponding infrastructure for tests.

doplumi:
Doesn't seem to work on my CLion w/ Rust plugin enabled and import ordering + unused import warning enabled. Is there any other setting I need to configure?

Undin:
@doplumi could you create a separate issue, please? Preferable with some code to reproduce it. `Help | Create new Issue` action may help you

PR #7496, opened 2021-07-09
INSP: Fix handling cfg-disabled code in unused import inspection

dima74:
Adjust visitor usage so that we don't abort processing when encounter cfg-disabled code

changelog: Fix handling cfg-disabled code in unused import inspection


Kobzol (approved):
LGTM. I didn't realize that returning `false` would actually stop the processing altogether, but it makes sense in retrospect.

vlad20012 (approved):
bors r+

dima74:
bors r=vlad20012

PR #7499, opened 2021-07-10
TYPE: added filter for items expanded from macros to Structure View

afetisov:
changelog: added a new filter to the Structure View panel, which toggles whether macro-generated items should be included in the Structure View.

The reason to add this filter is that it is common to produce boilerplate structures and trait implementations using macros. For example, if we introduce a new trait which should be implemented for all primitive numbers, then there will likely be 12 implementations which are derived by a single macro (signed and unsigned with 8, 16, 32, 64, 128, and `size` bits). Similarly, it is common in crates which don't use const generics to implement traits for 32 or more different array sizes separately, and crates which need generically sized tuples (like `itertools` or `warp`) often implement traits for all tuples up to a certain size.

Such boilerplate impls clutter the structure view, making it hard to find the actual types and functions.

### DIFF
--- a/src/main/kotlin/org/rust/ide/structure/RsMacroExpandedFilter.kt
+++ b/src/main/kotlin/org/rust/ide/structure/RsMacroExpandedFilter.kt
@@ -1,0 +1,40 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.structure
+
+import com.intellij.ide.util.treeView.smartTree.ActionPresentation
+import com.intellij.ide.util.treeView.smartTree.ActionPresentationData
+import com.intellij.ide.util.treeView.smartTree.Filter
+import com.intellij.ide.util.treeView.smartTree.TreeElement
+import org.rust.RsBundle
+import org.rust.ide.icons.RsIcons
+import org.rust.lang.core.macros.isExpandedFromMacro
+
+class RsMacroExpandedFilter : Filter {
+    override fun getPresentation(): ActionPresentation = ActionPresentationData(
+        RsBundle.message("structure.view.show.macro.expanded"),
+        null,
+        RsIcons.MACRO_EXPANSION,
+    )
+
+    override fun getName(): String = ID
+
+    override fun isVisible(treeNode: TreeElement?): Boolean {
+        val psi = (treeNode as? RsStructureViewElement)?.value ?: return true
+        return !psi.isExpandedFromMacro
+    }
+
+    override fun isReverted(): Boolean = true
+
+    companion object {
+        const val ID = "STRUCTURE_VIEW_MACRO_EXPANDED_FILTER"
+    }
+}
--- a/src/test/kotlin/org/rust/ide/structure/RsMacroExpandedFilterTest.kt
+++ b/src/test/kotlin/org/rust/ide/structure/RsMacroExpandedFilterTest.kt
@@ -1,0 +1,165 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.structure
+
+import org.intellij.lang.annotations.Language
+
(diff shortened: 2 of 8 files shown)


afetisov:
cc @vlad20012

src/main/kotlin/org/rust/ide/structure/RsMacroExpandedFilter.kt
```diff
@@ -0,0 +1,39 @@
...
+    override fun getPresentation(): ActionPresentation = ActionPresentationData(
+        RsBundle.message("structure.view.show.macro.expanded"),
+        null,
+        RsIcons.MACRO_EXPANSION,
+    )
+
+    override fun getName(): String = ID
+
+    override fun isVisible(treeNode: TreeElement?): Boolean {
+        return !((treeNode as? RsStructureViewElement)?.value?.isExpandedFromMacro ?: true)
```
vlad20012, line 31:
```suggestion
        val psi = treeNode as? RsStructureViewElement)?.value ?: return true
        return !psi.isExpandedFromMacro
```

src/test/kotlin/org/rust/ide/structure/RsMacroExpandedFilterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        |  Two visibility=none
+        | foo visibility=none
+        | -Struct visibility=public
+        |  field: bool visibility=private
+        | -Trait visibility=public
+        |  func() visibility=none
+        | -Trait for Struct visibility=none
+        |  func() visibility=none
+        | Persistent visibility=private
+        """, """
```
vlad20012, line 47:
```suggestion
    """, """
```

src/test/kotlin/org/rust/ide/structure/RsMacroExpandedFilterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        |   Inner visibility=public
+    """, """
+        |-main.rs visibility=none
+        | nested visibility=none
+        | external() visibility=private
+    """)
+
+    fun `test filter on macro expanded type definitions has no effect`() = doTest("""
+        macro_rules! typemacro {
+            (${'$'}t: ty) => {Option<${'$'}t>};
```
vlad20012, line 90:
```suggestion
            ($ t: ty) => {Option<$ t>};
```

src/test/kotlin/org/rust/ide/structure/RsMacroExpandedFilterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        | -Trait visibility=private
+        |  T visibility=none
+        | Foo visibility=private
+        | -Trait for Foo visibility=none
+        |  T visibility=none
+    """)
+
+    fun `test macro expanded expressions are always invisible`() = doTest("""
+        macro_rules! expr {
+            (${'$'}b: expr) => {if ${'$'}b {
```
vlad20012, line 130:
```suggestion
            ($ b: expr) => {if $ b {
```

vlad20012 (approved):
LGTM

vlad20012:
@Undin btw we likely want to design a better icon

image

src/main/kotlin/org/rust/ide/structure/RsMacroExpandedFilter.kt
```diff
@@ -0,0 +1,40 @@
...
+    override fun getPresentation(): ActionPresentation = ActionPresentationData(
+        RsBundle.message("structure.view.show.macro.expanded"),
+        null,
+        RsIcons.MACRO_EXPANSION,
+    )
+
+    override fun getName(): String = ID
+
+    override fun isVisible(treeNode: TreeElement?): Boolean {
+        val psi = (treeNode as? RsStructureViewElement)?.value ?: return false
```
vlad20012, line 31: Why did you decide to return false here? If we'll introduce another tree element type, this filter will start filter such elements

afetisov: Makes sense, I'll change it.

vlad20012:
:+1: Thanks!
bors r+

PR #7503, opened 2021-07-11
TY: convert primitive types from object to class

Kobzol:
This is a naive attempt to reduce the usage of `object` in primitive types and convert them to standard classes instead. This is probably needed as a first step to solve https://github.com/intellij-rust/intellij-rust/issues/5080. I didn't change `TyNever` and `TyUnknown`, because we probably don't need aliases to work for them (?).

The ultimate plan is to move `aliasedBy` from `TyAdt` to `Ty`. Which will probably introduce all kinds of scary breakage and problems (especially the description of this PR https://github.com/intellij-rust/intellij-rust/pull/4243 scares me in relation to comparing types :D ).

Sending PR to see what kinds of tests fail.

### DIFF
--- a/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
+++ b/src/main/kotlin/org/rust/ide/annotator/RsErrorAnnotator.kt
@@ -11,7 +11,6 @@ import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.ide.annotator.AnnotatorBase
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.AnnotationSession
-import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.openapi.editor.colors.TextAttributesKey
 import com.intellij.openapi.util.Key
 import com.intellij.openapiext.isUnitTestMode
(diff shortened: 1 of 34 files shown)


vlad20012 (approved):
LGTM.

The one thing I'd made additionally is turning `name` and `ordinal` fields in `TyInteger` and `TyFloat` classes into abstract properties with getters overridden in descendant classes.

> Which will probably introduce all kinds of scary breakage and problems (especially the description of this PR #4243 scares me in relation to comparing types

Well, `==` just shouldn't be used for `Ty`s. But there are already quite lot of such usages and I have no idea how to find them

> I didn't change TyNever and TyUnknown, because we probably don't need aliases to work for them (?).

Most likely yes.

Kobzol:
Good idea, I changed them.

EDIT: Oh, I thought that `override val name: String = "foo"` will just generate a getter without a field, but it does not seem to be the case :( I'll convert them to getters.

Kobzol:
Now all primitive types implement these properties with getters and there should be no memory wasted on backing fields.

Kobzol:
bors r=vlad20012

Kobzol:
bors retry

Kobzol:
bors retry

Kobzol:
bors retry

Kobzol:
bors r=vlad20012

PR #7507, opened 2021-07-12
INSP: Support declarative macros in unused import inspection

dima74:
Take into account macro paths when traversing modules in `PathUsageAnalysis`. Previously macros were intentionally ignored in inspection because of this

changelog: Support declarative macros in unused import inspection

### DIFF
--- a/src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
+++ b/src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
@@ -9,7 +9,10 @@ import com.intellij.openapi.util.Key
 import com.intellij.psi.PsiElement
 import com.intellij.psi.util.*
 import org.rust.lang.core.psi.*
-import org.rust.lang.core.psi.ext.*
+import org.rust.lang.core.psi.ext.RsElement
+import org.rust.lang.core.psi.ext.RsItemsOwner
+import org.rust.lang.core.psi.ext.isEnabledByCfg
+import org.rust.lang.core.psi.ext.qualifier
 import org.rust.lang.core.types.infer.ResolvedPath
 import org.rust.lang.core.types.inference
 import org.rust.openapiext.processElementsWithMacros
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
```diff
@@ -78,6 +78,10 @@ private fun handleElement(
             }
             true
         }
+        is RsMacroCall -> {
+            handleElement(element.path, directUsages, traitUsages)
+            true
```
Kobzol, line 83:
```suggestion
```
It's probably overkill, but for future-proofing this should maybe return the same thing as the recursive call?

Kobzol (approved):
LGTM :+1:

src/main/kotlin/org/rust/ide/inspections/lints/RsPathUsageAnalysis.kt
dima74: If we return false here, the entire macro expansion will be skipped, and return value of `handleElement` indicated whether or not we should go inside `element.path` subtree. It turns out my implementation doesn't handle qualified paths, I fixed it now, thanks!

Kobzol: Hmm. But I wonder why this change was needed though? In e.g. `foo::bar!`, the root path `foo::bar` shouldn't have a qualifier, so it should be handled by the `else` branch at :71, which should mark `foo` as visited path.

dima74: The root path `foo::bar` has a qualifier `foo`, isn't it?

Kobzol: Oh, nevermind, I again confused the order of nested paths produced by the parser. It confuses me that the qualifier is on the left, but at the same time it is the child of the root path. Thanks for the explanation.

dima74:
bors r=Kobzol

PR #7512, opened 2021-07-12
Introduce new `RsTemplateBuilder` API

vlad20012:
It should be a bit handier than `TemplateBuilder`. See example usage in `ImplTraitToTypeParamIntention.kt `


vlad20012:
To be honest, I don't like the API I've done :laughing:
But at the moment I'm not sure in which direction I should change it. Let's use it more extensively first, then I'll try to rethink it.

Kobzol (approved):
The testing ended up surprisingly simple! It's great that templates can be now properly tested.

The API looks OK, maybe there could be a nicer one that deviates more from the original `TemplateBuilderImpl` API, but as you said, for now I would keep this and then see if there are any problems or annoyances with its usage.

vlad20012:
bors r=Kobzol

PR #7513, opened 2021-07-12
ANN: disable Grazie proofreading for Rust language in 2021.2 IDEs

Undin:
Grazie plugin enables proofreading inspection for all languages by default in 2021.2 release. Since we actually haven't checked it properly yet (Grazie plugin was significantly changed recently).
So we decided to disable it for now, check properly, fix potential bugs and enable it by default again

changelog: Temporarily disable proofreading checks from [Grazie plugin](https://plugins.jetbrains.com/plugin/12175-grazie) for Rust language in 2021.2 IDEs


Undin:
CC @lancelote

lancelote:
Note: will be enabled in #7681.

PR #7516, opened 2021-07-13
COMP: fix autoimport-aware completion after completion at empty place

vlad20012:
The bug was explained @Kobzol in [the comment](https://github.com/intellij-rust/intellij-rust/pull/7480#issuecomment-878633497)

completion

The full completion list isn't shown when completing in an empty place because unfiltered completion (without a prefix) would be quite slow. But the bug is that the full completion list isn't loaded after typing the first letter.


vlad20012:
bors retry

PR #7519, opened 2021-07-13
COMP: add postfix template for wrapping a type with another type

Kobzol:
Recently I needed to wrap several types inside `Mutex` and I thought that a postfix template that wraps a type in another type could be nice.

wrap

To make it practical for repeated wrapping, it would be nice if the caret moved to the end of the wrapped type, instead of moving after its name. But I didn't find how to do this with the `Template` API. I could probably override and completely re-implement `expandForChooseExpression` to use a template builder, but I'm not sure if that's the right way to do it.

For some reason autoimport in the template used by the postfix completion doesn't work in tests. It works in the IDE though.

changelog: Add postfix template named `.wrap` that wraps an existing type with another type.

### DIFF
--- a/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/RsPostfixTemplateProvider.kt
@@ -46,7 +46,8 @@ class RsPostfixTemplateProvider : PostfixTemplateProvider {
         DbgrPostfixTemplate(this),
         OkPostfixTemplate(this),
         SomePostfixTemplate(this),
-        ErrPostfixTemplate(this)
+        ErrPostfixTemplate(this),
+        WrapTypePathPostfixTemplate(this)
     )
 
     override fun getTemplates(): Set<PostfixTemplate> = templates
--- a/src/main/kotlin/org/rust/ide/template/postfix/WrapTypePathPostfixTemplate.kt
+++ b/src/main/kotlin/org/rust/ide/template/postfix/WrapTypePathPostfixTemplate.kt
@@ -1,0 +1,48 @@
+/*
+ * Use of this source code is governed by the MIT license that can be
+ * found in the LICENSE file.
+ */
+
+package org.rust.ide.template.postfix
+
+import com.intellij.codeInsight.template.TemplateResultListener
+import com.intellij.codeInsight.template.impl.MacroCallNode
+import com.intellij.codeInsight.template.macro.CompleteMacro
+import com.intellij.codeInsight.template.postfix.templates.PostfixTemplateWithExpressionSelector
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import org.rust.ide.utils.template.newTemplateBuilder
+import org.rust.lang.core.parser.RustParserUtil
+import org.rust.lang.core.psi.RsBaseType
+import org.rust.lang.core.psi.RsPsiFactory
+import org.rust.lang.core.psi.RsTypeReference
+import org.rust.lang.core.psi.ext.startOffset
+import org.rust.openapiext.createSmartPointer
+
+class WrapTypePathPostfixTemplate(provider: RsPostfixTemplateProvider)
+    : PostfixTemplateWithExpressionSelector(null, "wrap", "\$wrapper$<path>", RsTypeParentsSelector(), provider) {
+    override fun expandForChooseExpression(element: PsiElement, editor: Editor) {
+        val typeRef = element as? RsTypeReference ?: return
+
+        val factory = RsPsiFactory(typeRef.project)
+        val path = factory.tryCreatePath("Wrapper<${typeRef.text}>", RustParserUtil.PathParsingMode.TYPE) ?: return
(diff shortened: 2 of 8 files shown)


vlad20012:
> it would be nice if the caret moved to the end of the wrapped type, instead of moving after its name. But I didn't find how to do this with the Template API.

Unfortunately, there's no way to do it using `Template`

Kobzol:
Now depends on https://github.com/intellij-rust/intellij-rust/pull/7519

completion-wrap

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -100,6 +107,34 @@ class SomePostfixTemplate(provider: RsPostfixTemplateProvider) : SimpleExprPostf
...
+        val path = factory.tryCreatePath("Wrapper<${typeRef.text}>", RustParserUtil.PathParsingMode.TYPE) ?: return
+        val newTypeRef = factory.tryCreateType(path.text) ?: return
+        val inserted = typeRef.replace(newTypeRef) as? RsBaseType ?: return
+        val ptr = inserted.createSmartPointer()
+
+        val template = editor.newTemplateBuilder(inserted) ?: return
+        val name = inserted.path?.referenceNameElement ?: return
+        template.replaceElement(name, MacroCallNode(CompleteMacro()))
+        template.withResultListener {
+            if (it != TemplateResultListener.TemplateResult.Canceled) {
```
vlad20012, line 126: Why did you decide to use `it != TemplateResultListener.TemplateResult.Canceled` instead of usual `== TemplateResult.Finished` (`withFinishResultListener`)? In this case, the template moves the caret if I cancel the template by pressing `Esc`

Kobzol: For some reason, if I press enter in this case in the builder: `Mutex/*caret*/<u32>`, the template ends with `BrokenOff`.

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -100,6 +107,34 @@ class SomePostfixTemplate(provider: RsPostfixTemplateProvider) : SimpleExprPostf
 class OkPostfixTemplate(provider: RsPostfixTemplateProvider) : SimpleExprPostfixTemplate("ok", "Ok(expr)", provider)
 class ErrPostfixTemplate(provider: RsPostfixTemplateProvider) : SimpleExprPostfixTemplate("err", "Err(expr)", provider)
 
+class WrapTypePathPostfixTemplate(provider: RsPostfixTemplateProvider) : StringBasedPostfixTemplate("wrap", "\$wrapper$<path>", RsTypeParentsSelector(), provider) {
```
vlad20012, line 110: `StringBasedPostfixTemplate` is not needed now, we can extend `PostfixTemplateWithExpressionSelector`

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
```diff
@@ -100,6 +107,34 @@ class SomePostfixTemplate(provider: RsPostfixTemplateProvider) : SimpleExprPostf
...
+        val typeRef = element as? RsTypeReference ?: return
+
+        val factory = RsPsiFactory(typeRef.project)
+        val path = factory.tryCreatePath("Wrapper<${typeRef.text}>", RustParserUtil.PathParsingMode.TYPE) ?: return
+        val newTypeRef = factory.tryCreateType(path.text) ?: return
+        val inserted = typeRef.replace(newTypeRef) as? RsBaseType ?: return
+        val ptr = inserted.createSmartPointer()
+
+        val template = editor.newTemplateBuilder(inserted) ?: return
+        val name = inserted.path?.referenceNameElement ?: return
```
vlad20012, line 123:
```suggestion
        val name = ptr.element?.path?.referenceNameElement ?: return
```

src/main/kotlin/org/rust/ide/template/postfix/stringBasedPostfixTemplates.kt
vlad20012: Yes, I've reproduced. The fun fact is that it is called with `BrokenOff` for `Mutex` but not for my custom types! This looks like a bug (somewhere) and I tried to investigate it but with no success :(

Kobzol: I suppose that it's not such a problem. If the user canceled the template, she probably wants to revert the wrapping anyway, so she will invoke Ctrl + Z and then it makes sense for the caret to return to the original location.

vlad20012: I just don't like when something doesn't work as expected and I have no idea why

vlad20012 (approved):
bors r+

Kobzol:
(sorry, I just wanted to push changes made after your last review and include a comment about the BrokenOff problem)

bors r=vlad20012

